PR #2928, opened 2015-03-31
shutdown robustness and test tasks

colinsurprenant:
refactored test tasks and added robustness in shutdown handling and other thread safeties

this issue also relates to
- [x] logstash-plugins/logstash-input-tcp#6
- [x] logstash-plugins/logstash-input-syslog#8
- [x] logstash-plugins/logstash-input-generator#5
- [x] logstash-plugins/logstash-input-file#21
- [x] logstash-plugins/logstash-input-gelf#9
- [x] logstash-plugins/logstash-input-pipe#7
- [x] logstash-plugins/logstash-input-redis#5
- [x] logstash-plugins/logstash-input-relp#7
- [x] logstash-plugins/logstash-input-heartbeat#6
- [x] elastic/logstash-devutils#25

PR #2932, opened 2015-04-01
bump version 1.5.0.rc3

colinsurprenant:
this is to merge in the 1.5 branch when ready to release RC3

PR #2950, opened 2015-04-07
New version bump for testing PR #2928

### LINKED ISSUE #2949: Publish logstash-core SNAPSHOT version
As part of testing for #2928 we need to publish a new version for logstash-core gem. Lets publish a snapshot version before we are ready for a RC3 version

The version should be 1.5.0-rc3.snapshot

suyograo:
Closes #2949

### DIFF
--- a/lib/logstash/version.rb
+++ b/lib/logstash/version.rb
@@ -1,6 +1,6 @@
 # encoding: utf-8
 # The version of logstash.
-LOGSTASH_VERSION = "1.5.0.rc2"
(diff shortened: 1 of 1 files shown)


purbon:
re #2932, we should merge one or another .... but this two PR are basically doing the same thing, In my case I agree on having as much snapshots as need.

purbon:
I close this as in #2958 I included this code plus a number in the snapshot label, so we know the precedence in case of different snapshots releases.

Issue #2931, opened 2015-04-01
logstash-core jruby-httpclient runtime dependency required?


colinsurprenant:
It looks like the jruby-httpclient gem is not used in logstash-core but yet it is listed as a runtime dependency [in the gemspec](https://github.com/elastic/logstash/blob/master/logstash-core.gemspec#L59)

Can we remove this dependency?

This is related to logstash-plugins/logstash-output-elasticsearch_http#2

ph:
I think we can remove this dependency, I do not recall any plugin using this?

purbon:
If is not used in the core, should go to each plugins, +1 on removing stale dependencies.

PR #2968, opened 2015-04-09
Remove dependencies on jruby-httpclient

ph:
This gem isn't used in any logstash plugins but the core still depends on it.
Fixes: #2931


suyograo:
LGTM

Issue #2974, opened 2015-04-09
jars_path is invalid in load_elasticsearch_jars method


Mogztter:
In my plugin I'm using `load_elasticsearch_jars` with LogStash 1.5.0-rc2.

The `load_elasticsearch_jars` in [environment.rb](https://github.com/elastic/logstash/blob/master/lib/logstash/environment.rb#L59) is looking for ElasticSearch jars in `$LOGSTASH_HOME/vendor/elasticsearch/**/*.jar`.

The default [logstash-1.5.0-rc2 package](https://www.elastic.co/downloads/logstash) doesn't contains the `vendor/elasticsearch` directory and the method throws `Could not find Elasticsearch jar files under...`.
Maybe this is intended ? but the jars can be found in ` bundle/jruby/1.9/gems/logstash-output-elasticsearch-0.1.18-java/lib`.

A simple symbolic link `elasticsearch ->  bundle/jruby/1.9/gems/logstash-output-elasticsearch-0.1.18-java` in `$LOGSTASH_HOME/vendor/` can solve the issue but... is this a bug ? Am I doing something wrong ?

Thanks

ph:
You are writing a plugin that uses the `logstash-output-elasticsearch`? If it is the case your gem should define a runtime dependencies on it, by doing that the gem will make the jar available after the installation process. Can I ask why you are building a plugin on top of the elasticsearch output?

Sadly the `load_elasticsearch_jars` is legacy code and should be removed in a future PR.

Mogztter:
> You are writing a plugin that uses the logstash-output-elasticsearch? 

Not exactly, I'm using the [RabbitMQ river plugin](https://github.com/elastic/logstash/blob/v1.4.2/lib/logstash/outputs/elasticsearch_river.rb) from the 1.4.2 branch. I know that rivers are deprecated and will be removed (or are removed ?) from ElasticSearch 1.5 but this plugin is really useful to consume messages from RabbitMQ.

And I'm using LogStash 1.5.0 because we want to replace RabbitMQ with Kafka but while we are transitioning we need to keep RabbitMQ running (with river).

purbon:
For what I understand from your description I would encourage you to use an idea similar to the one proposed in #2980, so by embedding in your gem/plugin them afterward in execution time you are basically done. If you need inspiration to write this code, is basically doing the same thing as it's done right now with the ```load_elasticsearch_jars!```.

Hope this helps you fix your issue.

purbon:
[...]

Issue #2954, opened 2015-04-07
Bundling JAR dependencies with the main gem


purbon:
The way jar dependencies are managed through ruby-maven and jar-dependencies have had many issues, so as noticed in https://github.com/elastic/logstash/issues/2937#issuecomment-90608786 we aim to do to embeed the jars within the gem. 

In other words, we want to download the jars and package it at gem creation time and not at install time

NOTE: This is actually a meta issue to track the necessary changes in the different projects.

The list of plugins that might need to be changed are:

- [x] https://github.com/logstash-plugins/logstash-input-log4j
- [x] https://github.com/logstash-plugins/logstash-output-elasticsearch
- [x] https://github.com/logstash-plugins/logstash-output-elasticsearch-ec2
- [x] https://github.com/logstash-plugins/logstash-input-drupal_dblog

Others that are slightly related with this change as they use jars, or have wrong unexpected dependencies:

- [x] https://github.com/logstash-plugins/logstash-input-elasticsearch
- [x] https://github.com/logstash-plugins/logstash-filter-elasticsearch
- [x] https://github.com/logstash-plugins/logstash-input-jdbc

An special deal goes with 

- [x] https://github.com/logstash-plugins/logstash-input-kafka
- [x] https://github.com/logstash-plugins/logstash-output-kafka

as they use internally jar_dependencies.

This is a WiP list of plugins to be change, so there might be new ones added, but at the end we should be able to remove all references in devutils and the other projects of jar-dependencies, ruby-maven, etc.

This will help to provide a fix to #2937, https://github.com/logstash-plugins/logstash-input-log4j/issues/5 and help to build an offline installation package easy (related issue #2376)

colinsurprenant:
also relates to #2376

purbon:
@colinsurprenant wrote that in the last line of the main comment :-)

wiibaa:
@purbon kafka plugins simply depends on jruby-kafka gem that already bundle all necessary jars, is there really work to be done in this case of transitive dependencies ?

purbon:
@wiibaa with this strategy there is a hope the 3 party dependencies are well managed by the gem, or in case we know this is not the case, then we can handle them the same way as in here. Does this makes sense?

purbon:
By looking at the hole set of plugins, the only ones that make direct usage of jars as in jar-dependencies are:

```
[...]

Issue #2937, opened 2015-04-02
Issues with different version name schemas between rubygems and maven.


purbon:
Moved from the logstash-input-log4j repo to here as is where it belongs:

--
I got on step further, having a successfull bundle install by using specific logstash-core gem version in the gemspec 
````
s.add_runtime_dependency "logstash-core", '1.5.0.rc2'
````
I strongly suspect that ruby-maven is having a hard time because logstash-core gem versioning does not follow [maven convention for qualifier](http://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN401)
So for him, none of the [available version in rubygem ](https://rubygems.org/gems/logstash-core) are following the rule [1.4.0,2.0.0), thus failing

This should be fixed with 1.5.0 release hopefully or if you do an rc3, please try naming it with a **dash** 1.5.0-rc3 as 
  > All versions with a qualifier are older than the same version without a qualifier (release version).
---

Thanks @wiibaa for bringing this.

More details in the old issue (https://github.com/logstash-plugins/logstash-input-log4j/issues/5).

It relates to https://github.com/elastic/logstash/pull/2932

**Comment by ph** [comment]
Make sense!

**Comment by colinsurprenant** [comment]
I don't understand the relation with a gemspec `.add_runtime_dependency` on a gem and ruby-maven, what am I missing?

I think the problem is simply that that using a dependency like
```ruby
s.add_runtime_dependency "logstash-core", '>= 1.4.0', '< 2.0.0'
```
will just try to fetch the latest **released** gem that match this version constrain but `1.5.0.rc2` is a **pre-released** version. Just like when using Rubygems you have to use  `gem install --pre ...` to force installing a pre-released version.

wiibaa:
@colinsurprenant the root issue I strongly suspect 
> If you do not follow Maven versioning standards in your project versioning scheme, then for version comparison, **Maven interprets the entire version as a simple string.**

this would explain why the version range check  in ruby-maven to lookup the logstash-core gem is failing.

Side discussion about pre-release, maven is doing it its own way and enforcing the use of **SNAPSHOT** as a special qualifier for prerelease see [7.2 The SNAPSHOT Qualifier](http://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN402)
Otherwise you are free to set the last part of the version as you like i.e Spring or Hibernate does as following 
[...]

PR #2752, opened 2015-03-03
refactored all gem/bundler code into logstash/bundler.rb and moved patches into logstash/patches

colinsurprenant:

PR #2980, opened 2015-04-10
Add the idea of a jar manager to handle jar_dependencies

purbon:
In #2937 and #2954 we discussed the idea to build a jar manager in order to change the way we're loading the jar dependencies within our code base.

This PR add the concept of a jar manager responsible of finding and loading jars as dependencies.

Collaterally this PR also removed the load_elasticsearch_jars! method as it's a legacy code that was there from the early days when the ES jars were shipped with LS and not in the plugins. 

This new jar manager is in use in

* https://github.com/logstash-plugins/logstash-input-log4j/pull/8
* https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/101
* https://github.com/logstash-plugins/logstash-output-elasticsearch-ec2/pull/6
* https://github.com/logstash-plugins/logstash-input-drupal_dblog/pull/7

Relates to #2974

For now it still need a bit more test to be sure everything is working as expected.


purbon:
This test scenario has been successful for me.

* Install all plugins
* Update the plugins that has been changed in this PR.
* Run the all plugins test.

:green_apple:

lib/logstash/environment.rb
```diff
@@ -45,7 +45,6 @@ module Environment
 
     LOGSTASH_HOME = ::File.expand_path(::File.join(::File.dirname(__FILE__), "..", ".."))
     JAR_DIR = ::File.join(LOGSTASH_HOME, "vendor", "jar")
```
wiibaa: I suppose you could also remove this line, JAR_DIR is not used anymore in 1.5

purbon: makes sense @wiibaa

colinsurprenant:
- why not keep `load_jars!` in the `Envronment` module? this is pretty much the idea of this module...
- why default to a new `dependencies/jar` path? we already have the `vendor` dir for that? just `vendor/jars` no?

purbon:
* I wanted to create something separate in order to let this be loaded without the overhead of the hole LS-Environment, I don't specifically see this as LS env, while loading JARS we don't need stuff like bundle for example. But this was just my point or initial idea. 

* I wanted to made this generic enough  so you can change this if you want, this are just the default parameters I thought interesting. But the important question here is, do we want the jars to live within the lib dir? or within the vendors? I don't have strong opinions, but I always use them throw lib and don't strongly dislike the idea.

colinsurprenant:
Agreed about bundler code not belonging in the `Environment` module and PR #2752 solves that.

On the other hand since the `Environment` module is already loaded in all plugins I don't think we need to created a separate module for a single jar loading method which we basically already had with the `Environment,load_elasticsearch_jars!` method. Also, I this it is specifically about the environment: it is basically just a path convention and checking that we are running JRuby. 

For the jars dir, first, let's make the default relevant to the logstash environment. Once we agree and where we should store the jars, let's change the defaults to that. For the `vendor/` dir, I am actually on the fence about it, since typically you expect the be able to delete the `vendor` dir and recreate it using some rake tasks or something. On the other hand, this is exactly what the `vendor` dir is about, to store external dependencies.  So I think I'd vote for putting jars into `vendor/jar` unless we see a potential problem with that?

purbon:
* About ``Environment``` ok, I like the idea to make it there.
* About the directory, I see the point and the regular usage of the ```vendor``` dir, as you said I'm on the fence, but find the idea interesting ... so lets use vendor too. Is this something we all like?

jsvd:
+1 for vendor/jars

colinsurprenant:
@jsvd did you mean `vendor/jar` ?

purbon:
why not ```vendor/jar-dependencies/runtime``` ?

jsvd:
yes. yes I did.

colinsurprenant:
@purbon `vendor/jar-dependencies/runtime` wat?

purbon:
what happen when we have also other dependencies for test? for example.

purbon:
All related code in here has been adapted to the last review comments, including the dependants PR's on the plugins side.

jsvd:
image

purbon:
@jsvd I agree I should add them, but before I wanted us to agree on the API, although it seems there is no concern with it.

purbon:
@jsvd simple unit test suite added.

lib/logstash/environment.rb
```diff
@@ -114,6 +95,42 @@ def bundler_setup!(options = {})
       ::Bundler.setup
     end
 
+    def runtime_jars_root(dir_name, package)
```
jsvd: In the plugins we do:

```
ROOT_DIR   = File.expand_path(File.join(File.dirname(__FILE__), ".."))
LogStash::Environment.load_runtime_jars! File.join(ROOT_DIR, "vendor")
```

But that's a lot of unecessary code we need on each plugin. At most, we need `LogStash::Environment.load_runtime_jars! File.dirname(__FILE__)` and even that we could somehow find by ourselves.

If we reduce this to just `LogStash::Environment.load_runtime_jars!`, we no longer need this extra file and can just include the require call in the plugin code. thoughts?

purbon: I agree with you in part, however I did it this way in order to provide more flexibility, keep in mind now this file has a ```ROOT_DIR``` just one level below the file that load the jars, but it can happen to be in another location. Or that for some jars, we aim to store them in another place. 

In my opinion, even if It might make sense, I find it too restrictive.

lib/logstash/environment.rb
```diff
@@ -114,6 +95,42 @@ def bundler_setup!(options = {})
...
+      ::File.join(dir_name, package, "runtime-jars")
+    end
+
+    def test_jars_root(dir_name, package)
+      ::File.join(dir_name, package, "test-jars")
+    end
+
+    def load_runtime_jars!(dir_name="vendor", package="jar-dependencies")
```
jsvd: the code of `load_runtime_jars!` and `load_test_jars!` is the same, it should be put in a common method, leaving them as 

```ruby
def load_runtime_jars!(dir_name="vendor", package="jar-dependencies")
  generic_load_jars!(runtime_jars_root(dir_name, package))
end

def load_test_jars!(dir_name="vendor", package="jar-dependencies")
  generic_load_jars!(test_jars_root(dir_name, package))
end
```

lib/logstash/environment.rb
jsvd: I understand the flexibility argument, but don't really see a practical use case for it. The jars should be in a predetermined folder, and all .jar files we find in that folder and its subfolders get loaded.

purbon: Ok, lets do it that way so we can move on and merge, we are always allowed to revert the change if need.

wiibaa:
@purbon there is an empty spec file `spec/util/environment_spec.rb`, maybe this PR could clean it up ?

purbon:
good catch again @wiibaa, cleaning this up too.

lib/logstash/environment.rb
jsvd: After discussing with @purbon, let's go with the current strategy. The amount of code plugins need to load their jars should either be a one liner or it could be detected automatically. However this requires a different strategy, so I've opened an issue to deal with this: https://github.com/elastic/logstash/issues/3005

purbon: Good @jsvd, this is a good discussion :-)

jsvd:
Reviewed, ran specs, tested with logstash-output-elasticsearch ([pr/101](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/101)) using :path and `bin/plugin install --no-verify`

LGTM

lib/logstash/environment.rb
```diff
@@ -1,5 +1,5 @@
 require "logstash/errors"
-require 'logstash/version'
```
ph: double *dragon* quotes +1

lib/logstash/environment.rb
```diff
@@ -114,6 +95,43 @@ def bundler_setup!(options = {})
...
+    def test_jars_root(dir_name, package)
+      ::File.join(dir_name, package, "test-jars")
+    end
+
+    def load_runtime_jars!(dir_name="vendor", package="jar-dependencies")
+      load_jars!("#{runtime_jars_root(dir_name, package)}/*.jar")
+    end
+
+    def load_test_jars!(dir_name="vendor", package="jar-dependencies")
+      load_jars!("#{test_jars_root(dir_name, package)}/*.jar")
```
ph: `/*.jar` this is unix specific no?

purbon: good point @ph, changing that to use a non platform specific constant

lib/logstash/environment.rb
```diff
@@ -114,6 +95,43 @@ def bundler_setup!(options = {})
...
+      jar_files = find_jars(pattern)
+      require_jars! jar_files
+    end
+
+    def find_jars(pattern)
+      require 'java'
+      jar_files = Dir.glob(pattern)
+      raise(LogStash::EnvironmentError, "Could not find jar files under #{pattern}") if jar_files.empty?
+      jar_files
+    end
```
ph: The actual messages should be defined in the locale yml.

spec/core/environment_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    it "find test jars in the default location" do
+      expect(subject).to receive(:find_jars).with(default_test_location).and_return([])
+      subject.load_test_jars!
+    end
+
+    context "when loading a jar file" do
+
+      let(:dummy_jar_file)          { "vendor/jar-dependencies/runtime-jars/elasticsearch.jar" }
```
ph: many spaces?

purbon: probably :-), stupid editors

ph:
So if I understand correctly this code, this will load all the jars in the vendor directory?
Even if we don't actually use the plugin?

purbon:
@ph it load the plugins whenever this functions are used, in the current case from a file within each plugins, so no, it will load the jars whenever the plugins are used for first time.

purbon:
@ph can you make sure the code I took from here related to maven makes sense?

lib/logstash/environment.rb
ph: No need for constant, `File.join` will take care of that.

```
irb(main):001:0> File.join('vendor', '*.jar')
=> "vendor/*.jar"

Comment (purbon) [inline lib/logstash/environment.rb]: ok

Comment (ph) [inline lib/logstash/environment.rb]: `File::SEPARATOR` is the constant for the platform specific stuff, but `File.join` doesnt care if the path exist or not.

**Comment by ph** [comment]
Well, in my current checkout this PR breaks `logstash-*-kafka`

```
~/e/logstash git:master ❯❯❯ bin/plugin install logstash-input-kafka                                                                                                                                                                                                                                                                                                   ⬆ ✱
Validating logstash-input-kafka
Installing logstash-input-kafka
Installation successful
~/e/logstash git:master ❯❯❯ tree vendor/bundle/jruby/1.9/gems/logstash-input-kafka-0.1.12                                                                                                                                                                                                                                                                             ⬆ ✱
vendor/bundle/jruby/1.9/gems/logstash-input-kafka-0.1.12
├── CONTRIBUTORS
├── DEVELOPER.md
├── Gemfile
├── LICENSE
├── README.md
├── Rakefile
├── lib
│   └── logstash
│       └── inputs
│           └── kafka.rb
├── logstash-input-kafka.gemspec
└── spec
    └── inputs
        └── kafka_spec.rb

5 directories, 9 files
```

purbon:
@ph good you have test with kafka now, might be is because i removed the maven code, because I tested before that and it was working ... see thanks for the review man

purbon:
@ph I checked with the change and with master and I see the same number of files for the kafka input, I see no change in this plugin ... I must say there was actually no change with this plugin for this PR

ph:
Did a test with this PR, Jars are in jruby-kafka.

LGTM.

purbon:
ok, merging this in :shipit:  :shipit:  :shipit:  :shipit:  :shipit:  :shipit:

purbon:
test this please

purbon:
can you test this please

codekitchen:
Related to this change, it looks like the documentation describing how to add jar dependencies to a logstash plugin is now out of date: https://www.elastic.co/guide/en/logstash/current/_how_to_write_a_logstash_input_plugin.html#_jar_dependencies

Is there guidance on the new correct way of doing things? I'm trying to update my plugin https://github.com/codekitchen/logstash-input-kinesis to work with logstash 1.5.1, but I'm not clear on how to get my jars properly vendored now. Thanks.

purbon:
@codekitchen you are right, documentation should be updated. Do you mind opening an issue for that? 

On the other side, you can see how to add you own jars if you take a look for example at https://github.com/logstash-plugins/logstash-input-log4j/tree/master/lib , there you will see a final named ```logstash-input-log4j_jars.rb``` who has the code necessary to load the jars.

Jars are then placed into the gem, in the case of the ```logstash-input-log4j``` you can find them at https://github.com/logstash-plugins/logstash-input-log4j/tree/master/vendor/jar-dependencies/runtime-jars . You can apply the same pattern in your plugin.

Makes sense? 

/cheers

codekitchen:
Thanks @purbon I filed https://github.com/elastic/logstash/issues/3455

I very rarely work with the JVM, so it actually took me a little while to figure out a good method for getting the jars I need into `vendor` -- I found it a lot more difficult than the previous `jar-dependencies` based method where everything just seemed to work based on a gemspec declaration. I also found no scripts or clues in the other plugins such as elasticsearch or log4j, which surprised me a bit. How were those jars vendored?

Eventually I ended up getting it working by adding a `pom.xml` file to my project and using `mvn` to vendor the jars, in https://github.com/codekitchen/logstash-input-kinesis/commit/816a10671ae44b3342580263f43f378511bb5322 , but I have no idea if that's the best option.

purbon:
@codekitchen good you figure it out, basically nowadays we just drop all need jars in a folder, even if not the best solution, ```jar-dependencies``` was causing us some random hassle during plugin management and installation. We're working with the maintainer of the gem to address this issues sooner or later and go back to it as it's clear a better way to handle jar dependencies.

In the meanwhile this is a necessary pain point to handle plugins who need jars. In a nutshell you should grab all jars you need and put them in the runtime folder. 

Thanks a lot for filling #3455, this would help our documentation efforts.

Issue #2261, opened 2014-12-17
Logstash1.5 - Removing tag in filter fails with undefined method delete


thehybridtechnician:
Upgraded to 1.5Beta1 to test out functionality.  Attempting to remove a tag fails.  Below is a snippet from my configuration.

Filter configuration
filter {
  if [program] == "services_monitor" and [message] =~ /Current memory used by supervisor/ {
    grok {
      patterns_dir            => "/etc/logstash/patterns/dellPatterns"
      match           => [
        "message", "Current memory used by supervisor %{DATA}\[%{DATA:Process}\] = %{NUMBER:memUsedKB:int}"
      ]
      remove_tag      => ["logs"]
      add_tag         => ["state"]
    }
  }
}

Failure Message
Exception in filterworker {"exception"=>#<NoMethodError: undefined method `delete' for #<Java::JavaUtil::ArrayList:0x6ada79a9>>, "backtrace"=>["/opt/logstash/lib/logstash/filters/base.rb:198:in `filter_matched'", "org/jruby/RubyArray.java:1613:in `each'", "/opt/logstash/lib/logstash/filters/base.rb:193:in `filter_matched'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-filter-grok-0.1.2/lib/logstash/filters/grok.rb:287:in `filter'", "(eval):187175:in `initialize'", "org/jruby/RubyArray.java:1613:in `each'", "org/jruby/RubyEnumerable.java:805:in `flat_map'", "(eval):187171:in `initialize'", "org/jruby/RubyArray.java:1613:in `each'", "org/jruby/RubyEnumerable.java:805:in `flat_map'", "(eval):187168:in `initialize'", "org/jruby/RubyArray.java:1613:in `each'", "org/jruby/RubyEnumerable.java:805:in `flat_map'", "(eval):187129:in `initialize'", "org/jruby/RubyProc.java:271:in `call'", "/opt/logstash/lib/logstash/pipeline.rb:268:in `filter'", "/opt/logstash/lib/logstash/pipeline.rb:206:in `filterworker'", "/opt/logstash/lib/logstash/pipeline.rb:150:in `start_filters'"], :level=>:error}

Thank you!

talevy:
hey @thehybridtechnician! Just checking-in to see if this is still an issue.

suyograo:
I tried to reproduce this too, but could not. Please re-open this if this is still an issue

ruddj:
[...]

PR #2991, opened 2015-04-13
add Java Collection delete support with specs & more java_integration specs

colinsurprenant:
this solves #2261

### DIFF
--- a/lib/logstash/java_integration.rb
+++ b/lib/logstash/java_integration.rb
@@ -6,27 +6,11 @@ require "java"
 # not test for is_a?(Array) or is_a?(Hash) and we do not want to include tests for
 # both classes everywhere. see LogStash::JSon.
 
-class Java::JavaUtil::ArrayList
-  # have ArrayList objects report is_a?(Array) == true
-  def is_a?(clazz)
-    return true if clazz == Array
-    super
-  end
-end
-
-class Java::JavaUtil::LinkedHashMap
-  # have LinkedHashMap objects report is_a?(Array) == true
-  def is_a?(clazz)
-    return true if clazz == Hash
-    super
-  end
-end
-
 class Array
   # enable class equivalence between Array and ArrayList
   # so that ArrayList will work with case o when Array ...
   def self.===(other)
-    return true if other.is_a?(Java::JavaUtil::ArrayList)
+    return true if other.is_a?(Java::JavaUtil::Collection)
     super
   end
 end
@@ -35,7 +19,44 @@ class Hash
   # enable class equivalence between Hash and LinkedHashMap
   # so that LinkedHashMap will work with case o when Hash ...
   def self.===(other)
-    return true if other.is_a?(Java::JavaUtil::LinkedHashMap)
+    return true if other.is_a?(Java::JavaUtil::Map)
     super
+  end
+end
+
+module java::util::Map
+  # have Map objects like LinkedHashMap objects report is_a?(Array) == true
+  def is_a?(clazz)
+    return true if clazz == Hash
+    super
+  end
+end
+
+module java::util::Collection
+  # have Collections objects like ArrayList report is_a?(Array) == true
+  def is_a?(clazz)
+    return true if clazz == Array
+    super
+  end
+
+  # support the Ruby Array delete method on a Java Collection
+  def delete(o)
+    self.removeAll([o]) ? o : block_given? ? yield : nil
+  end
+
+  # support the Ruby intersection method on Java Collection
+  def &(other)
+    # transform self into a LinkedHashSet to remove duplicates and preserve order as defined by the Ruby Array intersection contract
+    duped = Java::JavaUtil::LinkedHashSet.new(self)
+    duped.retainAll(other)
+    duped
+  end
+
+  # support the Ruby union method on Java Collection
+  def |(other)
+    # transform self into a LinkedHashSet to remove duplicates and preserve order as defined by the Ruby Array union contract
+    duped = Java::JavaUtil::LinkedHashSet.new(self)
+    duped.addAll(other)
+    duped
   end
 end
--- a/spec/lib/logstash/java_integration_spec.rb
+++ b/spec/lib/logstash/java_integration_spec.rb
@@ -1,0 +1,169 @@
+# encoding: utf-8
+require "spec_helper"
+require "logstash/java_integration"
+
+describe "Java integration" do
+
+  context "type equivalence" do
+
+    # here we test for both is_a? and case/when usage of the Java types
+    # because these are the specific use-cases in our code and the expected
+    # behaviour.
+
+    context "Java::JavaUtil::ArrayList" do
+
+      it "should report to be a Ruby Array" do
+        expect(Java::JavaUtil::ArrayList.new.is_a?(Array)).to be_true
+      end
+
+      it "should be class equivalent to Ruby Array" do
+        expect do
+          case Java::JavaUtil::ArrayList.new
+          when Array
+            true
+          else
+            raise
+          end
+        end.not_to raise_error
+
+        expect(Array === Java::JavaUtil::ArrayList.new).to be_true
+      end
+    end
+
+    context "Java::JavaUtil::LinkedHashMap" do
+      it "should report to be a Ruby Hash" do
+        expect(Java::JavaUtil::LinkedHashMap.new.is_a?(Hash)).to be_true
+      end
+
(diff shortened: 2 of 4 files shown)


spec/lib/logstash/java_integration_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      expect(result).to be_nil
+      expect(a.size).to eq(3)
+
+      a = Java::JavaUtil::ArrayList.new
+      result = a.delete("baz")
+      expect(result).to be_nil
+      expect(a.size).to eq(0)
+   end
+
+    it "should yeild block when nothing to delete" do
```
jsvd: yield

spec/lib/logstash/java_integration_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      a = Java::JavaUtil::ArrayList.new
+      a += ["foo", "bar", "foo"]
+      result = a.delete("baz") {"boo"}
+      expect(result).to eq("boo")
+      expect(a.size).to eq(3)
+    end
+  end
+
+end
```
ph: @colinsurprenant can we use `subject` for `Java::JavaUtil::ArrayList.new` ?

ph:
@colinsurprenant This fix is working with me, the tests are running.
But I would take some time to tidy up the test suite a bit.

What I mean by that is:

1. Uses subject for the object under evaluation.
2. Uses let statements for the initialization of the array or other useful variables.
3. It is better to have one expect per example, it provide a better separation of errors and context.

spec/lib/logstash/java_integration_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      expect do
+        case Java::JavaUtil::LinkedHashMap.new
+        when Hash
+          true
+        else
+          raise
+        end
+      end.not_to raise_error
+    end
+  end
```
ph: Could we replace this construct with one of the following?

```ruby
expect(subject).to be_a_kind_of(Hash)
expect(subject).to be_a(Hash)

```

ph: I believe `when Klass` is using kind_of?..

purbon:
+1 on @ph proposals for refactoring the test suite a bit :-)

spec/lib/logstash/java_integration_spec.rb
colinsurprenant: no. these 2 tests using `is_a?` and using `case/when` are exactly & specifically what need to be be supported and tested. Here rspec dsl will make thing **not as clear** because the java integration code has been specifically written to enable both the `is_a?` and `case/when` functionalities on the related Java classes.

spec/lib/logstash/java_integration_spec.rb
colinsurprenant: you mean replace with
```ruby
subject { Java::JavaUtil::ArrayList.new }
subject += ["foo", "bar", "foo"]
result = subject.delete("baz") {"boo"}
expect(subject.size).to eq(3)
```

?

ph: This is what I propose and display how subject could be useful
```ruby
context "Java::JavaUtil::Collection" do
  context "when deleting elements" do
    subject { Java::JavaUtil::ArrayList.new("foo", "bar") } #not sure if this is supported.
 
    let(:candidate) { "foo" }
 
    it "returns the object" do
      expect(subject.delete(candidate)).to eq(candidate)
    end
 
    it "updates the size of the array" do
      expect_change { subject.delete(candidate) }.to change { subject.size }.by(1)
    end
 
    it "reorder the items" do
      expect_change { subject.delete(candidate) }.to change { subject.first }.from(candidate).to("bar")
    end
  end
```

spec/lib/logstash/java_integration_spec.rb
ph: I didn't know you have to write specific code to support the `case` behavior, I have always assumed it was relying on `kind_of` to do the actual type checking. This believe could come from the days I was looking through the rubinius code.

Could you point me to it to the actual logic? I would like to learn how it is done.

colinsurprenant: it's all in `java_integration.rb`, see https://github.com/elastic/logstash/blob/master/lib/logstash/java_integration.rb#L9 and https://github.com/elastic/logstash/blob/master/lib/logstash/java_integration.rb#L28
and comments

ph: Ty I though it was outside of logstash.

ph:
Other than my minor comments on the test suite this PR is LGTM.
@colinsurprenant since we want to get this merged ASAP we could create a ticket for the test changes if you want.

spec/lib/logstash/java_integration_spec.rb
colinsurprenant: Ok thanks for the examples, TIL expect_change.
I get the idea but in this specific case, I think this is somewhat over engineering a simple test isn't it?

spec/lib/logstash/java_integration_spec.rb
colinsurprenant: also, `kind_of` might be a correct test but these java integration patches were specifically made to solve the `case` and `is_a?` use-cases hence the specific test on `case` and `is_a?`.

spec/lib/logstash/java_integration_spec.rb
ph: I don't think it is over engineering:

- `expect_change` is actually a better test here than asserting an exact value.
- Running a simple expectation is fast and isolated http://betterspecs.org/#single
- Line of code from your examples and mine are really close similar.
- Using subject and let make test code easier to refactor, move or reuse via shared helpers.

jsvd: Also line one should be describe, not context:

```ruby
describe "Java::JavaUtil::Collection" do
  context "when deleting elements" do
    it "should  ..." do
    end
  end
end
```

spec/lib/logstash/java_integration_spec.rb
ph: Or you could test `===` directly?

colinsurprenant: yes, I could but again, we don't use `===` we use `case` so in that respect, changing `===` is an implementation detail and you normally don't want to test an implementation detail, but the end-result behaviour you want.

lib/logstash/java_integration.rb
```diff
@@ -39,3 +39,9 @@ def self.===(other)
     super
   end
 end
+
+module java::util::Collection
+  def delete(o)
+    self.removeAll([o]) ? o : block_given? ? yield : nil
```
suyograo: why the conversion to single element array? can't we just call `self.remove(o)` ?

colinsurprenant: because `self.remove` removes only the first item equal to `o` in `self` and this is not the Ruby   `Array#delete` contract which should deletes all items from self. `self.removeAll` does that but takes a collection of objects as argument.

colinsurprenant:
thanks for your suggestions for making better specs :P

ph:
LGTM

colinsurprenant:
I decided to refactor the `spec/core/event_spec.rb` to test for the initialization of an `Event` from deserialized json for the case where data is from json, deserialized by JrJackson and contains native Java Collection objects. It showed that the `Array |` (union) operator was missing so I also added it in the java integration module and added corresponding specs.

lib/logstash/java_integration.rb
```diff
@@ -35,7 +19,42 @@ class Hash
...
+    return true if clazz == Array
+    super
+  end
+
+  # support the Ruby Array delete method on a Java Collection
+  def delete(o)
+    self.removeAll([o]) ? o : block_given? ? yield : nil
+  end
+
+  # support the Ruby intersection methos on Java Collection
```
purbon: methos should be method, right?

colinsurprenant: right :P

jsvd:
LGTM

purbon:
LGTM, even if the test suites need a bit more love :-) is out of the scope of this PR, I like the usage of a shared example, well done on this @colinsurprenant.

btw, you need rebase :-)

colinsurprenant:
great! :shipit:

colinsurprenant:
for the record, different union and intersection implementations benchmarks in https://gist.github.com/colinsurprenant/bc3575efde2743f14666

PR #2997, opened 2015-04-13
Fixes Event#include? to properly check for field

### LINKED ISSUE #2977: Bad mutate filter behavior due to nil type checking
So I have a config in which I am trying to rename/delete fields which occasionally have nil values.  When I rename a field with the mutate filters rename parameter, I get a new field, but the event fails to remove the old field.  When I use the remove_field parameter, it simply fails to remove the field.  Event.remove calls the del accessor.  The problem lies with the del accessor in logstash/lib/logstash/util/accessors.rb.

Here is the offending code:
```
 def del(accessor)
      target, key = lookup(accessor)
      unless target.nil?
        target.is_a?(Array) ? target.delete_at(key.to_i) : target.delete(key)
      end
    end
```
unless target.nil? prevents me from deleting and properly renaming fields within logstash (via mutate) when the value of the field (ie target) is nil.

Perhaps this check should be removed.

talevy:
Beforehand, Event#include? would depend on a value of a certain
field to be non-nil as the signal that a field exists. This breaks
when a field's value is nil, even if the field exists. It is now
possible to check for existance of both normal and metadata fields.

Closes #2977.

### DIFF
--- a/lib/logstash/util/accessors.rb
+++ b/lib/logstash/util/accessors.rb
@@ -53,6 +53,11 @@ module LogStash::Util
       end
     end
 
+    def include?(accessor)
+      target, key = lookup_path(accessor)
+      return target.include?(key)
+    end
+
     private
 
     def lookup(accessor)
--- a/lib/logstash/event.rb
+++ b/lib/logstash/event.rb
@@ -176,8 +176,14 @@ class LogStash::Event
   end
 
   public
-  def include?(key)
-    return !self[key].nil?
+  def include?(fieldref)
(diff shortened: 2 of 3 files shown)


suyograo:
LGTM

lib/logstash/util/accessors.rb
```diff
@@ -53,6 +53,11 @@ def del(accessor)
       end
     end
 
+    def include?(accessor)
+      target, key = lookup_path(accessor)
+      return target.include?(key)
```
colinsurprenant: explicit returns are useless in Ruby

PR #3017, opened 2015-04-15
Update the exclude list for the all plugins test

purbon:
There are some plugins being empty that needed to be added to the exclude list.


jsvd:
LGTM. haven't tested

Issue #2656, opened 2015-02-20
Adding an unpublished plugin to logstash-plugins breaks test:install-all


jsvd:
Example with the latest logstash-output-logentries plugin:

```
% rake test:install-all
Resolving dependencies...
Plugin not found, aborting
Error: Bundler::GemNotFound, Could not find gem 'logstash-output-logentries (>= 0) java' in the gems available on this machine.
Installation aborted
rake aborted!
```

The workaround is adding the plugin to the skip list in `rakelib/default_plugins.rake`

purbon:
Yeah! we should make this more resilient, but for now I guess is good if we somehow catch this exception and just log something happened.

I guess @colinsurprenant used the same approach as the old code, fetching all plugins and then filtering the ones we don't want ;-)

purbon:
Actually breaking the jenkins jobs for that with an error like:

```
asticsearch-ec2, logstash-patterns-core, logstash-mixin-aws, logstash-input-couchdb_changes, logstash-filter-metaevent, logstash-filter-zeromq, logstash-codec-s3plain, logstash-codec-cloudfront, logstash-codec-gzip_lines, logstash-codec-avro, logstash-input-heartbeat, logstash-filter-tld, logstash-filter-oui, logstash-output-logentries
Plugin not found, aborting
Installation aborted
rake aborted!

/home/jenkins/workspace/logstash_integration_all_plugins/rakelib/plugin.rake:7:in `install_plugins'
/home/jenkins/workspace/logstash_integration_all_plugins/rakelib/plugin.rake:55:in `(root)'
org/jruby/RubyProc.java:271:in `call'
org/jruby/RubyArray.java:1613:in `each'
org/jruby/RubyArray.java:1613:in `each'
org/jruby/RubyArray.java:1613:in `each'
org/jruby/RubyKernel.java:1087:in `load'
Tasks: TOP => test:install-all => plugin:install-all
(See full trace by running task with --trace)
rake aborted!
no such file to load -- rspec/core/runner
org/jruby/RubyKernel.java:1071:in `require'
/home/jenkins/workspace/logstash_integration_all_plugins/rakelib/test.rake:5:in `run_rspec'
/home/jenkins/workspace/logstash_integration_all_plugins/rakelib/test.rake:19:in `(root)'
org/jruby/RubyProc.java:271:in `call'
org/jruby/RubyArray.java:1613:in `each'
org/jruby/RubyArray.java:1613:in `each'
[...]

PR #3021, opened 2015-04-15
Check for plugins to be released

purbon:
Hi,
  one regular issue with the plugins test pipeline has been that we search for plugins in github, while most of them are released to rubygems and perse usable, some are not. This PR search rubygems and make sure there is a plugin with the name that can be installed.


purbon:
criteria for being sure this pass is being able to run ``` rake test:install-all``` without any exception.

ph:
LGTM, but I havent run the full scenario.

purbon:
Fixes #2656

jsvd:
This means we can remove or empty the ALL_PLUGINS_SKIP_LIST?

purbon:
@jsvd not sure if we want to remove it at all, we might still want to skip something that might be also released. But might be we want to revisit what is there.

PR #3022, opened 2015-04-15
fix java enumerable integration bug

### LINKED ISSUE #3018: failing event spec on deserialized json input
the new event spec introduced in #2997 (@talevy) does not work with deserialized json input as exposed with a new event spec strategy to test both normal hash and hash from deserialized json introduced in #2991 

this need to be figured out.

colinsurprenant:
will fix #3018

### DIFF
--- a/lib/logstash/java_integration.rb
+++ b/lib/logstash/java_integration.rb
@@ -24,6 +24,24 @@ class Hash
   end
 end
 
+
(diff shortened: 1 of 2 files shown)


talevy:
LGTM

ph:
LGTM, minor comment to use subject more, there is a lot of code repetition in the suite.
Since the build is currently broken lets merge it.

PR #3023, opened 2015-04-15
fix include to support nil and array types

talevy:
Closes https://github.com/logstash-plugins/logstash-filter-mutate/issues/21.

### DIFF
--- a/lib/logstash/util/accessors.rb
+++ b/lib/logstash/util/accessors.rb
@@ -55,7 +55,8 @@ module LogStash::Util
 
     def include?(accessor)
       target, key = lookup_path(accessor)
-      return target.include?(key)
+      return false unless target
+      target.is_a?(Array) ? !target[key.to_i].nil? : target.include?(key)
     end
 
     private
--- a/spec/core/event_spec.rb
+++ b/spec/core/event_spec.rb
@@ -120,6 +120,7 @@ describe LogStash::Event do
       it "should include existing fields" do
         expect(subject.include?("c")).to be_true
         expect(subject.include?("[c][d]")).to be_true
+        expect(subject.include?("[j][k4][0][nested]")).to be_true
       end
 
       it "should include field with nil value" do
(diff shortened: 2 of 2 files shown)


talevy:
The last commit with the Java LinkedHashMap stuff is not expected to stay. but I wanted to hear all the pros and cons about it since I am already doing a bunch of type checking. I know this forces a dependency on Jruby within core. I will have a "is in jruby" check, just like the rest of the pipeline does if this commit is acceptable.

talevy:
Can you test this please?

spec/core/event_spec.rb
```diff
@@ -137,6 +137,12 @@
       it "should not include non-existing fields" do
         expect(subject.include?("doesnotexist")).to be_false
         expect(subject.include?("[j][doesnotexist]")).to be_false
+        expect(subject.include?("[tag][0][hello][yes]")).to be_false
+      end
+
+      it "should include within arrays" do
+        expect(subject.include?("[tags][0]")).to be_true
+        expect(subject.include?("[tags][1]")).to be_false
       end
```
ph: Maybe we should add tests to test nested hash?
`"[tags][0][exist]"`

talevy: I have a test with nesting, up above. or do you specifically mean one with nesting of hash within an array?

```
expect(subject.include?("[c][d]")).to be_true
```

ph: yes, I think we need to have test for hash within an array.

lib/logstash/util/accessors.rb
```diff
@@ -55,7 +55,13 @@ def del(accessor)
...
     def include?(accessor)
       target, key = lookup_path(accessor)
-      return target.include?(key)
+      if target.is_a?(Array)
+        !target[key.to_i].nil?
+      elsif target.nil?
+        false
+      else
+        target.include?(key)
+      end
```
colinsurprenant: I prefer to avoid too many branches and be more concise, I'd suggest:
```ruby
return false unless target
target.is_a?(Array) ? !target[key.to_i].nil? : target.include?(key)
```

talevy: done!

spec/core/event_spec.rb
talevy: added!

colinsurprenant:
LGTM

ph:
LGTM test run.

PR #3025, opened 2015-04-15
Fixes 2987

### LINKED ISSUE #2987: add_field adds boolean values as string
When using mutate add_field trying to add fields as boolean, the field is added as string. Can be seen both with rubydebug and json output
Tested on windows in logstash 1.4.2, 1.5.0.rc2 and also latest version pulled from github.
`
logstash-1.5.0.rc2\bin\logstash.bat  agent  -e 'filter{mutate{add_field => {"isthisbool" => false}}} output{stdout {codec => json}}'
fisk
{"message":"fisk\r","@version":"1","@timestamp":"2015-04-10T09:24:40.573Z","type":"stdin","host":"CV-E6420","isthisbool":"false"}
`
In this sample, "isthisbool" should have been false, not "false".

untergeek:
This ensures that:

    add_field => [ "fieldname", true ]

and

    add_field => [ "fieldname", false ]

actually result in true boolean values.

fixes #2987

### DIFF
--- a/lib/logstash/filters/base.rb
+++ b/lib/logstash/filters/base.rb
@@ -168,6 +168,10 @@ class LogStash::Filters::Base < LogStash::Plugin
         if event.include?(field)
           event[field] = [event[field]] if !event[field].is_a?(Array)
           event[field] << v
+        elsif v == 'true'
+          event[field] = true
+        elsif v == 'false'
+          event[field] = false
         else
           event[field] = v
         end
--- a/spec/filters/base_spec.rb
+++ b/spec/filters/base_spec.rb
@@ -47,6 +47,24 @@ describe LogStash::Filters::NOOP do
     end
   end
 
+  describe "add_field with string booleans" do
+    config <<-CONFIG
+    filter {
+      noop {
+        add_field => ["true_field", "true"]
+        add_field => ["false_field", "false"]
+      }
+    }
+    CONFIG
+    sample 'add_field => ["true_field", "true"]' do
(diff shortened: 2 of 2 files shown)


talevy:
I think we should leave this alone, and keep it as a string. as is the expected behavior.

If one wants to then force that "fieldname" to be a bool. they should follow up with an explicit
`convert` action like is described here: https://github.com/logstash-plugins/logstash-filter-mutate/blob/master/lib/logstash/filters/mutate.rb#L258

and we can add a `def convert_bool`

untergeek:
@talevy I think that would work, but ugh.  Why can't it be a single step?

What if we made it harder to make a mistake?

```
add_field => ["field1", "boolean_true"]
add_field => ["field2", "boolean_false"]
```

What about making this the official way to force a `true` or `false` value?  Just change the string in the tests and the filter.rb.

talevy:
I'm a bigger fan of being explicit, rather than raising the implicit difficulty. but, I agree. In practice, that would suffice.

suyograo:
I like explicit too, since we already have `convert => {int, float}` .. extending that paradigm

untergeek:
I'm outvoted here.  I dislike the extra step, but I'll go along with it.

untergeek:
Closing in favor of https://github.com/logstash-plugins/logstash-filter-mutate/issues/20

untergeek:
Fixed by https://github.com/logstash-plugins/logstash-filter-mutate/pull/22

PR #3034, opened 2015-04-16
Remove the if branch the tests still passes..

ph:
When investigating this https://github.com/logstash-plugins/logstash-filter-multiline/issues/10 issue I found out that we always set @metadata to a hash and it cant be nil. The tests was actually checking if the metadata was empty.

### DIFF
--- a/lib/logstash/event.rb
+++ b/lib/logstash/event.rb
@@ -295,11 +295,7 @@ class LogStash::Event
 
   public
   def to_hash_with_metadata
-    if @metadata.nil?
-      to_hash
-    else
-      to_hash.merge(METADATA => @metadata)
-    end
+    @metadata.empty? ? to_hash : to_hash.merge(METADATA => @metadata)
   end
 
   public
--- a/spec/core/event_spec.rb
+++ b/spec/core/event_spec.rb
@@ -409,6 +409,10 @@ describe LogStash::Event do
         it "should still allow normal field access" do
           expect(subject["foo"]).to eq("bar")
         end
+
+        it "should not include the @metadata key" do
+          expect(subject.to_hash_with_metadata).not_to include("@metadata")
+        end
       end
     end
 


suyograo:
@ph tests?

lib/logstash/event.rb
```diff
@@ -295,11 +295,7 @@ def init_timestamp(o)
 
   public
   def to_hash_with_metadata
-    if @metadata.nil?
-      to_hash
-    else
-      to_hash.merge(METADATA => @metadata)
-    end
+    to_hash.merge(METADATA => @metadata)
```
colinsurprenant: since @metadata can never be nil per https://github.com/elastic/logstash/blob/master/lib/logstash/event.rb#L77 I believe the intent is to not include a @metadata field when it is empty, so it should actually be:
```ruby
@metadata.empty? ? to_hash : to_hash.merge(METADATA => @metadata)
```

no?

ph: @colinsurprenant I know how to make it work.

ph: it either the test or the code is wrong, the test was checking for `be_empty`

ph: re: linked code.

Well the intent is to provide a fall back if no metadata key is supplied at object creation time, we never enforce it to be a hash. (maybe we should?)

The only place I could see a benefit to return nil is if we were serializing that field... but we dont serialize metadata.

colinsurprenant: I just verified everywhere:
- it seems that `to_json_with_metadata` is not used anywhere.
- `to_hash_with_metada` is used in rubydebug codec, multiline filter and couchdb_changes input, in all cases it make no difference.

So I'd say we do what's fastest: either checking for `empty?` or doing a hash `merge` I would assume checking for `empty?` is fastest then `merge`...?

ph: Sold

ph:
@colinsurprenant I've testing it locally the conditionals is faster, i'll go ahead and merge it.

colinsurprenant:
LGTM

PR #3037, opened 2015-04-17
bin/bundle shim

colinsurprenant:
add the `bin/bundle` shim. this shim is essentially the same shim as the one supplied with the bundler gem but includes our bundler patches for the `.lock` file naming. 

this allow doing `bin/bundle exec rspec` for example, or any other bundle command. very useful. with that we can get rid of `bin/logstash rspec`.

### DIFF
--- a/bin/bundle
+++ b/bin/bundle
@@ -1,0 +1,24 @@
+#!/usr/bin/env ruby
+
+# This is basically a copy of the original bundler "bundle" shim
+# with the addition of the loading of our Bundler patches that
+# modifies the .lock file naming. Without this, using the original Bundler bundle
+# shim would result in creating or failing on a Gemfile.lock file.
+
+# Exit cleanly from an early interrupt
+Signal.trap("INT") { exit 1 }
+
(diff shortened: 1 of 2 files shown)


bin/bundle
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env ruby
+
+# Exit cleanly from an early interrupt
+Signal.trap("INT") { exit 1 }
+
+$: << File.expand_path(File.join("__FILE__", "..", "lib"))
```
purbon: Even if ```$:``` is well known for many people as for LOAD_PATH, I recommend using ```$LOAD_PATH``` as it's more clear for everyone.

colinsurprenant: ok, done

colinsurprenant:
as discussed I also excluded `bin/bundle` from the packages

purbon:
makes sense, 
 LGTM :shipit: :shipit: :shipit: :shipit:

ph:
LGTM

PR #3040, opened 2015-04-17
Fix relative pathname lookup while installing a local .gem file

purbon:
Fix the Pathname lookup for installing new .gem files.

### DIFF
--- a/lib/logstash/pluginmanager/command.rb
+++ b/lib/logstash/pluginmanager/command.rb
@@ -32,6 +32,7 @@ class LogStash::PluginManager::Command < Clamp::Command
   end
 
   def relative_path(path)
-    Pathname.new(path).relative_path_from(Pathname.new(LogStash::Environment::LOGSTASH_HOME)).to_s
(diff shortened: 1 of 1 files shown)


purbon:
@suyograo @ph review on here?

ph:
@purbon it was failing?  I think we require "pathname" globably ?

purbon:
I tested master and 1.5 and not.

purbon:
and btw, we should not require pathname globally :-)

lib/logstash/pluginmanager/command.rb
```diff
@@ -32,6 +32,7 @@ def remove_unused_locally_installed_gems!
   end
 
   def relative_path(path)
-    Pathname.new(path).relative_path_from(Pathname.new(LogStash::Environment::LOGSTASH_HOME)).to_s
+    require 'pathname'
```
ph: double quotes and merge.

ph:
:shipit: lgtm

PR #3041, opened 2015-04-17
wrong link for the config type `:bytes` documentation

ph:
When fixing https://github.com/logstash-plugins/logstash-codec-multiline/pull/7 I found the link was wrong


talevy:
LGTM

PR #3047, opened 2015-04-20
Add an rspec shim

purbon:
In the spirit of https://github.com/elastic/logstash/pull/3037, this PR adds a new rspec shim so users can run test in the same way as they do with the command provided by rspec.

### DIFF
--- a/bin/rspec
+++ b/bin/rspec
@@ -1,0 +1,14 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH << File.expand_path(File.join("__FILE__", "..", "lib"))
+$LOAD_PATH << File.expand_path(File.join(File.dirname(__FILE__), "../spec"))
+
+require "logstash/environment"
+require "logstash/bundler"
+LogStash::Bundler.setup!
+
+require "rspec/core"
+require "rspec"
+
+status = RSpec::Core::Runner.run(ARGV).to_i
+exit status if status != 0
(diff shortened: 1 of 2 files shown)


purbon:
@colinsurprenant @suyograo review?

colinsurprenant:
is this taken from the original `rspec` shim?

purbon:
somehow yes, the original shim is very minimalistic and actually using autorun (aka https://github.com/rspec/rspec-core/blob/v2.14.8/lib/rspec/core/runner.rb#L17) that end up doing the same I'm doing here.

colinsurprenant:
why don't we use the same shim code? (with autorun etc?)

purbon:
because autorun is something you don't want to use + as soon as we update to rspec3 this is going to be outdated.

colinsurprenant:
I am not sure I understand - why don't we use exactly the same shim logic as the original but simply add our bundler initialization on top?

purbon:
The shim code is different from the 2.x branch to the actual 3.x branch, the only common part is the run method.

colinsurprenant:
I just checked the v3 and its not the same either. since we use v2, let's use the same as v2 and when we upgrade to v3, let's use the same as v3, no?

jsvd:
Since this is an enhancement for developers and wont be included in the releases, I vote for merging as an experiment and see the use we get out of it. LGTM

bin/rspec
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env ruby
+
+spec_path = File.expand_path(File.join(File.dirname(__FILE__), "../spec"))
+
+$LOAD_PATH << File.expand_path(File.join("__FILE__", "..", "lib"))
+$LOAD_PATH << spec_path
```
colinsurprenant: $LOAD_PATH <<  File.expand_path(File.join(File.dirname(__FILE__), "../spec"))
spec_path is useless

colinsurprenant:
LGTM except my minor comment

Issue #3062, opened 2015-04-21
bin/plugin update fails on a fresh packaged release


jsvd:
error was:

```
C:\logstash-1.5.0-rc3.snapshot4>bin\plugin update
ERROR: Update is not supported for manually defined plugins or local .gem plugin
 installations:

C:\logstash-1.5.0-rc3.snapshot4>
```

ph:
Can you add the content of you gemfile ?

ph:
Nevermind, this also fail on macos, updated the title.

ph:
Another error when doing

```
bin/plugin uninstall logstash-input-stdin
bin/plugin update
```
```
stash-output-pipe, logstash-output-rabbitmq, logstash-output-redis, logstash-output-s3, logstash-output-sns, logstash-output-sqs, logstash-output-statsd, logstash-output-stdout, logstash-output-tcp, logstash-output-udp, logstash-output-xmpp, logstash-output-zeromq, logstash-patterns-core
Plugin not found, aborting
ERROR: Updated Aborted, message: Could not find gem 'logstash-input-stdin'.
Did you mean logstash-input-redis?

```

ph:
@jsvd I have updated the PR to fix the uninstall/update behavior.

```
/tmp ❯❯❯ cd logstash-1.5.0-rc3.snapshot4                                                                                                                                                                                                                                                                                                                                ⏎
/t/logstash-1.5.0-rc3.snapshot4 ❯❯❯ patch -p1 < ../3059.diff
patching file lib/logstash/pluginmanager/uninstall.rb
/t/logstash-1.5.0-rc3.snapshot4 ❯❯❯ patch -p1 < ../3063.diff
patching file lib/logstash/pluginmanager/update.rb
/t/logstash-1.5.0-rc3.snapshot4 ❯❯❯ bin/plugin list stdin                                                                                                                                                                                                                                                                                                               ⏎
logstash-input-stdin
/t/logstash-1.5.0-rc3.snapshot4 ❯❯❯ bin/plugin uninstall logstash-input-stdin
Uninstalling logstash-input-stdin

/t/logstash-1.5.0-rc3.snapshot4 ❯❯❯ bin/plugin update
[...]

PR #3063, opened 2015-04-21
Fix an issue with bin/plugin update with no arguments and no locally installed gems

ph:
Fixes: https://github.com/elastic/logstash/issues/3062


jsvd:
Tested, LGTM

PR #3078, opened 2015-04-22
clean 1.5 package generation

colinsurprenant:
the 1.5 package is not clean, here are some fixes:
- the Gemfile.defautls contained reference to the local gemspec for logstash-core. it need to be referring to the gem from the source
- we don't want to include the `.bundle/config`, `Gemfile*.defaults` and the `logstash-core.gemspec` in the package

Note that the `Gemfile` can still use the reference to the gemspec but when creating the `.defaults` files we need to chage the `Gemfile` to
```ruby
source "https://rubygems.org"
gem "logstash-core"
```

 and then:

```sh
rm -rf .bundle Gemfile.jruby-1.9.lock vendor/bundle
rake bootstrap
rake plugin:install-default
```

```sh
cp Gemfile Gemfile.default
cp Gemfile.jruby-1.9.lock  Gemfile.jruby-1.9.lock.defaults
```

and finally do 
```sh
rake artifact:tar
```

### DIFF
--- a/Gemfile.defaults
+++ b/Gemfile.defaults
@@ -2,7 +2,7 @@
 # If you modify this file manually all comments and formatting will be lost.
 
 source "https://rubygems.org"
-gemspec :name => "logstash-core"
+gem "logstash-core"
 gem "logstash-input-heartbeat"
 gem "logstash-output-zeromq"
 gem "logstash-codec-collectd"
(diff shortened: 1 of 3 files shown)


suyograo:
Tested. Created new tar artifact and ran basic tests: `bin/plugin list` `bin/plugin install` and `bin/plugin update`

ph:
photo on 2015-04-22 at 12 21 pm

ph:
Tested scenario includes.

- Manually adding a gem to the gemfile with :path
- Installing a .gem file into logstash

jsvd:
LOOKS GOOD TO MERGE

colinsurprenant:
:shipit:

Issue #3075, opened 2015-04-21
bin/plugin update will make plugin functions unusable


suyograo:
Tried with 1.5.0-rc3.snapshot6. 

1. un-tar and run `bin/plugin list`. Things work as expected

2. update plugins (we know mutate was published to 0.1.7 on RubyGems):

```sh
ubuntu@ip-10-0-1-9:~/logstash-1.5.0-rc3.snapshot6$ bin/plugin list --verbose | grep mutate
logstash-filter-mutate (0.1.6)
```
^^ old version packaged in artifacts

```sh
ubuntu@ip-10-0-1-9:~/logstash-1.5.0-rc3.snapshot6$ bin/plugin update
[...]

PR #3081, opened 2015-04-22
revert bundler setup

colinsurprenant:
after discussing the #3075 issue with @ph we think the root problem is probably this addition of bunler setup introduced in #3063 and that the previous problems were probably related to a bad package, fixed by #3078 

if we merge this, it will replace #3079


colinsurprenant:
cancelling in favour of #3084

PR #3093, opened 2015-04-23
Update hello-search.conf to avoid Version Mismatch

manavkataria:
Update hello-search.conf to avoid ES <-> Logstash Version Mismatch Error

### DIFF
--- a/docs/tutorials/10-minute-walkthrough/hello-search.conf
+++ b/docs/tutorials/10-minute-walkthrough/hello-search.conf
@@ -21,5 +21,7 @@ output {
     # This option below saves you from having to run a separate process just
     # for ElasticSearch, so you can get started quicker!
     embedded => true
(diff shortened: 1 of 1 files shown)


manavkataria:
I've signed the CLA now.

suyograo:
@manavkataria you can force push an update using `git push origin -f <branch_name>`

jordansissel:
I think this file only is used in 1.4 branch, 1.5+ uses asciidoc only and this tutorial was merged into others.

manavkataria:
@jordansissel Is this update required in 1.5 as well?

jordansissel:
@manavkataria the documentation doesn't refer to that anymore. In 1.5 and master we should probably just delete the old content.

manavkataria:
@jordansissel Agreed, that should avoid any confusion.

PR #3107, opened 2015-04-24
Fix #3100 and one other bug

### LINKED ISSUE #3100: [1.5.0-RC3] filters: undefined local variable or method `event`
As reported by @wjimenez5271 using 1.5.0-RC3 while debugging https://github.com/elastic/logstash/issues/3092

```
$ sudo /opt/logstash/bin/logstash -f /etc/logstash.d/lumberjack.conf
Exception in filterworker {"exception"=>#<NameError: undefined local variable or method `event' for #<LogStash::Pipeline:0x51a97faa>>, "backtrace"=>["(eval):111:in `initialize'", "org/jruby/RubyProc.java:271:in `call'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc3-java/lib/logstash/pipeline.rb:309:in `flush_filters'", "org/jruby/RubyArray.java:1613:in `each'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc3-java/lib/logstash/pipeline.rb:308:in `flush_filters'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc3-java/lib/logstash/pipeline.rb:317:in `flush_filters_to_output!'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc3-java/lib/logstash/pipeline.rb:223:in `filterworker'", "org/jruby/ext/thread/Mutex.java:149:in `synchronize'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc3-java/lib/logstash/pipeline.rb:223:in `filterworker'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc3-java/lib/logstash/pipeline.rb:156:in `start_filters'"], :level=>:error}
Logstash startup completed
cx^CSIGINT received. Shutting down the pipeline. {:level=>:warn}
Exception in lumberjack input {:exception=>#<LogStash::ShutdownSignal: LogStash::ShutdownSignal>, :level=>:error}
```

Config: https://gist.github.com/wjimenez5271/d145cf8906752b5a5cbe

jordansissel:
This PR fixes #3100 and another bug.

#3100 is a bug triggered when you have any filter occurring after a previous filter, when that previous filter uses flushing. For example:

```
filter {
  multiline { ... }
  mutate { }
}
```

You would crash with `NameError: undefined local variable or method `event'`

The 2nd bug is that event cancellation is only checked when deciding what to send to the output queue. In this case, you could have `drop { } multiline { }` and the multiline filter would _still_ receive the event even though `drop { }` marked it as cancelled. 

Both bugs should now be fixed in this PR.

### DIFF
--- a/lib/logstash/config/config_ast.rb
+++ b/lib/logstash/config/config_ast.rb
@@ -233,7 +233,14 @@ module LogStash; module Config; module AST
       when "filter"
         return <<-CODE
           #{variable_name}.filter(event) {|new_event| events << new_event }
+
+          # Break early if the event was cancelled
+          return [] if events.all?(&:cancelled?)
         CODE
+        # TODO(sissel): The above break-early is probably not the best
+        # implementation because a filter could possibly emit many events and
+        # only cancel one of them. I don't know of any plugins that do this,
+        # though.
       when "output"
         return "#{variable_name}.handle(event)\n"
       when "codec"
--- a/lib/logstash/config/config_ast.rb
+++ b/lib/logstash/config/config_ast.rb
@@ -155,7 +155,9 @@ module LogStash; module Config; module AST
 
               @logger.debug? && @logger.debug(\"Flushing\", :plugin => #{name}, :events => events)
 
-              #{plugin.compile_starting_here.gsub(/^/, "  ")}
+              events.each do |event|
+                #{plugin.compile_starting_here.gsub(/^/, "  ")}
+              end
 
               events.each{|e| block.call(e)}
             end


jordansissel:
I do apologize that this code change is in the config compiler. It's my least favorite part of Logstash to debug :P

lib/logstash/config/config_ast.rb
```diff
@@ -155,7 +155,9 @@ def compile_initializer
 
               @logger.debug? && @logger.debug(\"Flushing\", :plugin => #{name}, :events => events)
 
-              #{plugin.compile_starting_here.gsub(/^/, "  ")}
+              events.each do |event|
+                #{plugin.compile_starting_here.gsub(/^/, "  ")}
```
jsvd: not sure what this is supposed to do: expand `event.size` times the `plugin.compile_starting_here.gsub(/^/, "  ")`

lib/logstash/config/config_ast.rb
```diff
@@ -231,7 +233,14 @@ def compile
       when "filter"
         return <<-CODE
           #{variable_name}.filter(event) {|new_event| events << new_event }
+
+          # Break early if the event was cancelled
+          return [] if events.all?(&:cancelled?)
```
jsvd: If this is to be merged with #3110,

line 235 will use the multi_filter method, so events will be empty if all events are cancelled, this could be changed to return events if events.empty?

colinsurprenant:
@jsvd this PR is not needed if we merge #3110 - all that is fixed here is also fixed in #3110 - it's more a question of choosing the approach. I think we should go with #3110 :P since this problem is mainly a regression which is reverted by #3110 and the cancellation issue is, I think, better handled in #3110

jordansissel:
Closing in favor of #3110

PR #3084, opened 2015-04-22
remove setup! and add :clean to uninstall

colinsurprenant:
from @jsvd suggestion, remove all setup! and add :clean to uninstall
this replaces #3081 and #3079 
fixes #3075

Issue #2776, opened 2015-03-05
Generated packages include spec files


electrical:
Generated packages currently still include the spec directory. Should not need those.

ph:
We were shipping the `spec/` directory in the logstash 1.4.2 packages (deb, rpm, tar), if we are removing the folder we should also hide the `bin/logstash rspec` command. If we are removing them, we should update the developer workflow to with: *to develop a plugin and run tests you need a fresh copy from master.

purbon:
+1 on not removing test from the page (for now), but we should discuss what is the approach we aim to take. 

In guess source packages use to include the tests, but binaries (also debs, rpms, etc..) should not have them. Might be a good idea to just keep the test when necessary and make packages that will go to deployment (only) smaller.

electrical:
Production packages != development environment. I still strongly feel that anything dev wise should not be in the created packages.
If people want to do development they will need to do a git clone.

jordansissel:
I have always shipped the tests with the releases for as long as I can
remember with logstash.

It's not for development, but for testing. The idea is that if a user's
platform has issues, the test suite could hep find it.

-Jordan

purbon:
agree with you @jordansissel that is how I see test when I compile a project.

jordansissel:
If we decide to remove the tests, then we'll have to strip down lib/logstash/runner during a build, also, since it provides `bin/logstash rspec`, `bin/logstash pry`, etc, which are useful for testing.

ph:
We are currently removing the tests for the plugins but not the tests for the core.

colinsurprenant:
we currently cannot run test/specs in 1.5.rcX as the development dependencies are not included in the package, see #3229. 

Personally I am not convinced we should include tests and dev dependencies in a release package. Has it been really useful in supporting users?

colinsurprenant:
note that if we include the `spec` dir in logstash-core, in a release package we can just do `bin/plugin install --development` and all dev dependencies will be installed. 

maybe the specs could be added as a logstash-core-specs gem as a development dependency? or we just include the `spec/` dir in logstash-core.

purbon:
I see a point of not having the test included in the packages, actually I don't see many people running core test anyway. But then I don't see the benefit of maintaining a logstash-core-spec gem.

jordansissel:
My original intent of including tests with logstash was for two reasons:
1) Folks with bugs could run the main logstash test suite and hopefully more easily find a failing case.
[...]

PR #3111, opened 2015-04-25
use either local core lib dir or logstash-core gem

### LINKED ISSUE #3096: remove logstash-core directories and fix paths to use the logstash-core bundled gem
The 1.5.0-rc3 package includes the `logstash-core` gem and also has the `lib/`, `spec/` and `locale/` directories in the root of the package. This is redundant and we should leverage our `logstash-core` gem.

For this here’s what I suggest:

1- add the `spec/` and `locale/` directories in the `logstash-core` gem.
2- remove the `lib/`, `spec/` and `locale/` directories from the package root when building the artifacts
3- add the concept of both a package-home dir and a core-home dir
4- fix the `bin/` scripts, the `Environment` module and the few places that rely on `__FILE__` to use the proper package-home dir or a core-home dirs
5- make sure these work in the dual dev/package contexts where in dev the `logstash-core` code is local and in the package context the `logstash-core` code is under the `vendor/bundle/…` directory
6- make the `logstash-core` gem a proper plugin updatable by the plugin manager

This is the last mile to our gemification process and will allow us to update the core code without having to necessarily release a whole new package. This is very very very useful for bug fixes.

The changes are not very complicated and I think we should aim for 1.5.0.

colinsurprenant:
will fix #3096

adds the concept of `LOGSTASH_HOME` vs `LOGSTASH_CORE` where `LOGSTASH_HOME` points to the package root while `LOGSTASH_CORE` points to the directory containing the core code `lib/logstash/...`.
Both can be the same, when in development using a repo clone, or different when packaged where `LOGSTASH_CORE` points to the `logstash-core` gem under `vendor/bundle/...`.

It looks like it's working well but I haven't fully tested the package generation because I did not have internet on my flight :P hence the WIP. This is an early release so we can have more eyeballs on the idea. 

I also did a few cleanup of what I thought was dead/unused code.


colinsurprenant:
need to check with @pere or @suyograo about the jenkins environment setup for the build failure.

purbon:
@colinsurprenant best thing to test this might be throw a custom feature branch test job. Let me know what do you exactly need and I get this setup for you :-) ... 

For now it seems that jenkins is still not fully happy.

colinsurprenant:
@purbon yeah, just wanted to pair to sync on how the env setup is done in Jenkins to make sure its good with this strategy.

purbon:
@colinsurprenant can you elaborate more what do you need  ? what are your specific environment requirements to test that this changes work ?

suyograo:
Jenkins test this

colinsurprenant:
I think we can start testing this. there is currently a problem with the installation of the kafka plugins and I did not yet try to solve it. I removed them from the defaults list. the problem is reproducible with `bin/plugins install logstash-input-kafka`, it looks like it might be a jar-dependencies issue.

since all the code is now referenced through the `logstash-core` gem, there are basically 3 ways to reference it in the `Gemfile` for different scenarios:

1- for normal dev from the logstash cloned repo:

```ruby
source "https://rubygems.org"
gemspec :name => "logstash-core"
```

2- to generate a local test package that will rely on the local logstash-core gem code in the cloned repo:

```ruby
source "https://rubygems.org"
gem "logstash-core", :path => "/path/to/your/logstash/repo"
```

```sh
rm -rf vendor/bundle .bundle Gemfile.jruby-1.9.lock 
rake bootstrap
rake plugin:install-default
cp Gemfile Gemfile.defaults
cp Gemfile.jruby-1.9.lock Gemfile.jruby-1.9.lock.defaults
rake artifact:tar
```

3- to generate a "public" package that will rely on the latest published `logstash-core` gem

```ruby
source "https://rubygems.org"
gem "logstash-core"
```

same as (2)

lib/bootstrap.rb
```diff
@@ -0,0 +1,118 @@
+module LogStash
+  module Environment
```
purbon: Why are now bootstrap and environment mix ? as for the main responsibilities of both, I see them in separate files for clarity. I see times where I might need to load the env without having to bootstrap.

lib/bootstrap.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    def logstash_gem_home
+      ::File.join(BUNDLE_DIR, ruby_engine, gem_ruby_version)
+    end
+  end
+
+  module Bootstrap
+    extend self
+
+    def bundler_setup!(options = {})
```
purbon: In order to enforce clarity I see this code belonging to a bundler module, that then can be used from here when necessary.

lib/bootstrap.rb
```diff
@@ -0,0 +1,118 @@
...
+      # This keeps MRI/JRuby/RBX from conflicting over the Gemfile.lock updates
+      ::Bundler::SharedHelpers.module_exec do
+        def default_lockfile
+          ruby = "#{Environment.ruby_engine}-#{Environment.ruby_abi_version}"
+          Pathname.new("#{default_gemfile}.#{ruby}.lock")
+        end
+      end
+    end
+
+    def patch_rubygem!
```
purbon: why not having this code into the patches directory ?, there we've some others patches and helps to keep things easy to locate when have to go away.

purbon:
@colinsurprenant you really clean up a lot man !! :-) good job!. Even if very nice, might be the clean up was a good idea to have it in another pr.

lib/logstash/runner.rb
```diff
@@ -63,15 +44,6 @@ def run(args)
         end
         return LogStash::Agent.run($0, agent_args)
       end,
-      "rspec" => lambda do
```
purbon: We never spoke to clean this option. We added the ```bin/rspec``` as complement, are we sure we want to get rid of this and move to bin/rspec ?

bin/logstash.lib.sh
```diff
@@ -1,4 +1,9 @@
-basedir=$(cd `dirname $0`/..; pwd)
+unset CDPATH
```
purbon: what is CDPATH?

logstash-core.gemspec
```diff
@@ -9,7 +9,7 @@ Gem::Specification.new do |gem|
   gem.homepage      = "http://www.elastic.co/guide/en/logstash/current/index.html"
   gem.license       = "Apache License (2.0)"
 
-  gem.files         = Dir.glob(["lib/**/*.rb", "locales/*"])
+  gem.files         = Dir.glob(["lib/logstash-core.rb", "lib/logstash/**/*.rb", "spec/**/*.rb", "locales/*"])
```
purbon: As far as I know, ```"lib/**/*.rb"``` is equivalent to add ```lib/logstash-core.rb``` and  ```lib/logstash/**/*.rb```, any reason for not using it ?

lib/logstash/runner.rb
colinsurprenant: ah yes, I kind of unilaterally decided that this did not really belong into the runner. we can put it back if that's a problem.

bin/logstash.lib.sh
colinsurprenant: I am not sure, it was like this before and I haven't tested the effect of removing it.

logstash-core.gemspec
colinsurprenant: in the logstash-core gem we specifically **just** want the `lib/logstash-core.rb` and what is **under** `lib/logstash` not anything else in `lib/`.

colinsurprenant:
@purbon about `lib/bootstrap.rb`:
- as a **first step** I'd like to concentrate on any functional problems rather than its structure (separation into multiple files, modules, etc)
- keep in mind that `bootstrap.rb` contains the minimal stuff that is needed to bootstrap the bundler environment without having yet access to the `logstash-core` gem (which contains the rest of `lib/logstash` because until `Bundler` is launched we cannot assume **where** the `logstash-core` gem code is located, it is `Bundler` job to figure that out. in other words when running as a package `lib/bootstrap.rb` and the rest of `lib/logstash` are not in the same path: `lib/logstash` will be somewhere in `vendor/bundle/...` or to a completely different path, depending on how you configure your `Gemfile` for the `logstash-core` gem.
- so, again, before proposing changes in `lib/bootstrap.rb` assume that everything that is there is **all** what you have access too.

We can certainly better structure `lib/bootstrap.rb` and I am definitely open to suggestions but the way I did it was the shortest path to success without having code duplication - so I first would prefer we find bug with it then once its all working we can look at  how to clean it? make sense?

lib/logstash/runner.rb
purbon: Is not a problem to remove it, if we agree is a functionality we don't want to offer anymore. If we remove this, it will actually remove the ability to run test in released packages, I just wander is we're all good for that.

bin/logstash.lib.sh
purbon: You remove it in ```bin/logstash```, if gone from there, why not from here? or the other way around.

logstash-core.gemspec
purbon: I see your point. However I still have question, do we still want/need ```lib\logstash-event.rb``` ? might be we can get rid of this too.

bin/logstash.lib.sh
colinsurprenant: I 'moved' it from `bin/logstash` to `bin/logstash.lib.sh` to have all env initialization in `bin/logstash.lib.sh`

lib/logstash-core.rb
```diff
@@ -0,0 +1,2 @@
+module LogStash
```
purbon: Are we good with changing the name of the main file to logstash-core and getting rid of logstash ? I might make sense to change if we understand the code as a self contained gem

lib/logstash/runner.rb
colinsurprenant: ah good point. let's see if that is the best strategy to run that through runner.

ph: I would leave the rspec subcommand, it can help us debug user issues, see @jordansissel's comment on https://github.com/elastic/logstash/issues/2776#issuecomment-77572492, IIRC we are not shipping the `/bin/rspec` in any packages.

lib/logstash-core.rb
colinsurprenant: `logstash-core` is now a self contain gem, it just happen to also share the packaging stuff in the same repo.

lib/logstash/runner.rb
colinsurprenant: personally I'd prefer to include`bin/rspec` in packages... it seems a better separation oc concerns.

logstash-core.gemspec
colinsurprenant: `logstash-event` is another issue we'll deal with eventually. it is currently a released gem still in use but in its current state we cannot generate a new version because I assume the its gemspec is very probably broken. we can deal with this in another issue. not urgent for now, but yes, we keep it in the repo for now.

lib/logstash/runner.rb
ph: Well if we do that, we need to update the default help message.

```
~/e/logstash git:pr/3111 ❯❯❯ bin/logstash rspec                                                                                                                                                                                                                                                                                                                       ⏎ ✱
No such command "rspec"

Usage: logstash <command> [command args]
Run a command with the --help flag to see the arguments.
For example: logstash agent --help


Available commands:
  agent - runs the logstash agent
  version - emits version info about this logstash
  rspec - runs tests
```

purbon:
Hi,
  thanks for your time clarifying,

   speaking about lib/bootstrap.sh. I agree we should concentrate on having
something functional, but we should not forget to have a bit of code
structure and design, otherwise we might end up with a bit of code mess
that we need to clean up afterward (and we might forget, or fail, to do
it). So +1 on having everything more close to their right place.

If I understand properly your explanation, bootstrap is only need to start
working for bundler, isn't? does it makes sense to name it a bit different
and/or locate it near to bundler, or the environment module then ?
Bootstrap seems like an strange name to me, I would understand something
more like minimum environment, or something else ... bootstrap is more an
action than a name for me.

So your explanation make sense, just want to remember not to hurry up and
actually think about where to put each piece of code (after it works as
expected for sure!).

/cheers

On Tue, Apr 28, 2015 at 1:43 PM, Colin Surprenant <notifications@github.com>
wrote:

> @purbon <https://github.com/purbon> about lib/bootstrap.rb:
>
>    - as a *first step* I'd like to concentrate on any functional problems
>    rather than its structure (separation into multiple files, modules, etc)
>    - keep in mind that bootstrap.rb contains the minimal stuff that is
>    needed to bootstrap the bundler environment without having yet access to
>    the logstash-core gem (which contains the rest of lib/logstash because
>    until Bundler is launched we cannot assume *where* the logstash-core
>    gem code is located, it is Bundler job to figure that out. in other
>    words when running as a package lib/bootstrap.rb and the rest of
>    lib/logstash are not in the same path: lib/logstash will be somewhere
>    in vendor/bundle/... or to a completely different path, depending on
>    how you configure your Gemfile for the logstash-core gem.
>    - so, again, before proposing changes in lib/bootstrap.rb assume that
>    everything that is there is *all* what you have access too.
>
> We can certainly better structure lib/bootstrap.rb and I am definitely
> open to suggestions but the way I did it was the shortest path to success
> without having code duplication - so I first would prefer we find bug with
> it then once its all working we can look at how to clean it? make sense?
>
> —
> <https://github.com/elastic/logstash/pull/3111#issuecomment-97032076>.
>



-- 
Pere Urbon-Bayes
Software Engineer
https://www.elastic.co
https://twitter.com/purbon

ph:
I cannot use the `bin/rspec` to run the test if I dont have bundler installed..?
```
rm -rf vendor
rake test:install-core
[...]

~/e/logstash git:pr/3111 ❯❯❯ bin/rspec                                                                                                                                                                                                                                                                                                                                  ✱
/Users/ph/.rbenv/versions/2.2.0/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- bundler (LoadError)
	from /Users/ph/.rbenv/versions/2.2.0/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /Users/ph/es/logstash/lib/bootstrap.rb:50:in `bundler_setup!'
	from bin/rspec:4:in `<main>'
```

I have bundler installed in `vendor/bundle` from the bootstrap process.

``
ls -l vendor/bundle/jruby/1.9/gems/bundler-1.9.4

purbon:
@ph this might be a bug, I don't think you should need bundler to run the specs, however somehow bundler should help locating the dependencies, isn't? I don't recall this same issue happening in other places.

bin/logstash.lib.sh
purbon: ok, makes sense.

ph:
@purbon I have bundler installed locally.. I'll give another shot with a clean dev env.

lib/logstash/runner.rb
jsvd: bin/logstash rspec works on 1.4.2, let's not remove it :)

colinsurprenant: @jsvd why not? just "because it works on 1.4.2" is not a good enough reason IMO. This is not a major UI change and will not affect our "normal" users. It affects "only" those that want to run specs, and seriously, the developers will understand that change.

colinsurprenant:
@purbon ok, thanks for the feedback. Now, let's please focus first on testing it, finding and fixing the bugs and making sure this all works. I first want to make sure the strategy & mechanics works then we can look at the cosmetics. 

@ph I will verify this but normally the vendorized bundler should be used, it should not be required to have a local bundler I think.

ph:
Same issue on clean env:

```
rbenv install 2.2.0
rm -rf vendor
git reset --hard head
rake test:install-core
~/e/logstash git:pr/3111 ❯❯❯ bin/rspec                                                                                                                                                                                                                                                                                                                                
/Users/ph/.rbenv/versions/2.2.0/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- bundler (LoadError)
	from /Users/ph/.rbenv/versions/2.2.0/lib/ruby/2.2.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	from /Users/ph/es/logstash/lib/bootstrap.rb:50:in `bundler_setup!'
	from bin/rspec:4:in `<main>'
```

The environment is correctly bootstrapped for logstash:
```
~/e/logstash git:pr/3111 ❯❯❯ bin/logstash -e 'input { generator {}} output { stdout {}}'
2015-04-28T14:07:11.210Z sashimi Hello world!
2015-04-28T14:07:11.210Z sashimi Hello world!
2015-04-28T14:07:11.210Z sashimi Hello world!
```

ph:
@colinsurprenant Yeah i think its a minor issue, but running specs was my first test ;)

logstash-core.gemspec
ph: +1 postpone logstash-event.

lib/logstash/runner.rb
ph: The only issue we could have here is a clash with an existing rspec command, if the user add the `logstash/bin` to his $PATH. (naturally we have the same issue with the plugin command)

purbon:
@colinsurprenant we should keep in mind to clean up not too much, or at less note somewhere what we clean up ... I don't want to endup having a flashback, where did functionally went in a few weeks :-)

@ph  I checked with master, this is not happening there, so, this is a bug introduced by this PR .. good catch.

I think we agreed somehow not to remove the rspec part, more takers on what to remove or keep from this PR ?

lib/logstash/runner.rb
jsvd: just trying to reduce the number of changes in UI between minor versions, specially in a timeslot between RC releases. However I agree that most users do not call this.

bin/logstash.lib.sh
timbunce: When CDPATH env var is set `cd` becomes verbose about what directory was changed into. Here's an extract from the bash manual on CentOS 6:

> If a non-empty directory name from CDPATH is used, or if - is the first argument, and the directory change is successful, the absolute pathname of the new working directory is written to the standard output.

So unsetting CDPATH is good practice in shell scripts.

colinsurprenant:
@purbon I don't agree for `bin/logstash rspec` because there was no argument put forward in favor to keeping it. But if the majority vote in favor to keep it then we'll keep it, this is not central to this refactor.

ph:
@colinsurprenant Everything was done under `ruby 2.2.0`

colinsurprenant:
@ph I was able to reproduce when launching `bin/rspec` from MRI. But from JRuby it works. Can you confirm this?

ph:
@colinsurprenant it is working under `jruby-1.7.19`, after investigation bin/rspec never launch the bundled jruby, Maybe a good reason to keep `bin/logstash rspec` for 1.5.

colinsurprenant:
@ph oh, I see the issue, `bin/rspec` is directly a Ruby script and does not relaunch using the vendored JRuby like we do with the `bin/logstash` shell script of the rake tasks. I don't think this **ever** worked to be able to launch `bin/rspec` from MRI. To solve this wewill have to followup the same strategy as `bin/logstash` and `bin/plugin`.

colinsurprenant:
I just confirmed that the Kafka plugins installation is also failing in master, so this issue is probably not related to this refactor. Filed under #3123

colinsurprenant:
I am facing the same issue of uninstalled plugins as fixed in #3084 that was caused by calling Bundler setup before the uninstall. Now the problem is that to retrieve the pluginmanager code which is part of logstash-core, I have to do a Bundler setup, so we are back with the same issue. 

Since the pluginmanager code is not really "core" code, it is more peripheral code, just like the stuff in `bin/`, I will move it out of the `logstash-core` gem and put it part of the package, same idea as with `bootstrap.rb`. This way, Bundler setup will not be required to execute the pluginmanger code.

colinsurprenant:
Jenkins test this

suyograo:
please jenkins, test this.

colinsurprenant:
@jsvd @ph can I get help to make sure the Windows startup scripts are working?

jsvd:
I'll certainly test it tomorrow :+1:

colinsurprenant:
Everything seems to be working correctly for the local dev context and the artifact:tar package. 

At this point the only blockers (see comment with list of issues) are:
- Windows startup scripts
- Agree on the fate of `bin/logstash rspec`

lib/logstash/runner.rb
colinsurprenant: @ph users should definitely not do this anyway and I doubt they do...

bin/logstash.lib.sh
colinsurprenant: @timbunce :thumbsup:

suyograo:
```
source "https://rubygems.org"
gem "logstash-core", :path => "/tmp/logstash"
```
I was able to build `logstash-2.0.0.dev.tar.gz` but got this warning, fyi:
```
Installation successful
/private/tmp/logstash/lib/logstash/version.rb:3 warning: already initialized constant LOGSTASH_VERSION
```

colinsurprenant:
@suyograo trying to reproduce, at which step exactly did the *already initialized constant* warning was printed?

colinsurprenant:
@suyograo I cannot reproduce. Can you gist the your terminal session that show when/at which step this is happening exactly?

colinsurprenant:
Confirming that kafka plugins installation problem is gone after the plugin manager refactor. :metal:

colinsurprenant:
I just pushed a cleanup of `lib/bootstrap`. I split `lib/bootstrap.rb` into `lib/bootstrap/environment.rb`,  `lib/bootstrap/bundler.rb`, `lib/bootstrap/rubygems.rb` - rejoice @purbon :D

colinsurprenant:
please jenkins, test this.

colinsurprenant:
- did a cleanup and a bit of refactoring of the `bin/` startup scripts **much** better now.
- `bin/rspec` now work when started either from MRI or JRuby, using the same launching strategy a with `bin/plugin` or `bin/logstash` - this mean I am now very comfortable in not reintroducing `bin/logstash rspec` /cc @purbon @jsvd @ph 
- documented the supported environment  variables in `bin/logstash`
- added support for DEBUG=1 in `bin/` scripts to trace the exact command line launched

I am now satisfied with the state of this PR and ready to take review comments and/or merge so we can go through the RC4 tests before packaging. /cc @suyograo

lib/logstash/runner.rb
colinsurprenant: `bin/rspec` now works consistently when launched from any ruby and startup time is **much much** better that using `bin/logstash rspec` and I included it in the package. Let me know how you guys feel about that.

suyograo:
Testing update. Tested on mac and ubuntu. basic smoke testing of
* `bin/plugin list`
* `bin/plugin install`
* `bin/plugin uninstall`
* Built local gem and installed on core

`bin/plugin update` throws this error:

```sh
ubuntu@ip-10-0-1-11:~/ws/logstash$ bin/plugin update
ERROR: Update is not supported for manually defined plugins or local .gem plugin installations: logstash-core
```

colinsurprenant:
@suyograo it not a bug, its normal, the message says it. since `bin/plugin update` tries to update all plugins and `logstash-core` is "local", it cannot proceed. If you do a specific update it will work.

suyograo:
ok, specific update works 

```
ubuntu@ip-10-0-1-11:~/ws/logstash$ bin/plugin update logstash-input-irc
Updating logstash-input-irc
```

colinsurprenant:
@suyograo @jsvd @ph I updated the Windows startup scripts.

colinsurprenant:
I fixed the package generation and it is now straightforward and the package is completely self contained, using a locally built & installed `logstash-core` gem. The nice thing is that we can generate test packages without having to push a new version of `logstash-core` to Rubygems. Now the 2nd nice thing is that we can update the `logstash-core` gem using the plugin manager using either bare `logstash-core` that will be fetched from Rubygems or using a local `logstash-core` *dot gem* file.

**New instuctions for building a package**
```sh
rm -rf vendor/bundle .bundle/ Gemfile.jruby-1.9.lock ~/.m2/ vendor/local_gems
rake artifact:tar
```

colinsurprenant:
please jenkins, test this.

colinsurprenant:
please jenkins, test this.

colinsurprenant:
please jenkins, test this.

suyograo:
@colinsurprenant test run on Windows is all good. Created zip package from osx, tested the usual suspects on win, `bin\plugin` `bin\logstash` etc. Details here: https://gist.github.com/suyograo/34f091c3ce72955cf332

jsvd:
Had a problem downgrading a plugin. Context:

Used pr/3111 HEAD, cleaned repo, did `rake artifact:tar`, then:

```
/tmp % tar -zxf logstash-2.0.0.dev.tar.gz
/tmp % cd logstash-2.0.0.dev
/tmp/logstash-2.0.0.dev % bin/plugin uninstall logstash-codec-json
Uninstalling logstash-codec-json
/private/tmp/logstash-2.0.0.dev/vendor/local_gems/fccdd8d6/logstash-core-2.0.0.dev-java/lib/logstash/version.rb:3 warning: already initialized constant LOGSTASH_VERSION
/tmp/logstash-2.0.0.dev % bin/plugin install --version="0.1.6" logstash-codec-json
Validating logstash-codec-json-0.1.6
Installing logstash-codec-json
Plugin not found, aborting
ERROR: Installation Aborted, message: You have requested:
  logstash-codec-json = 0.1.6

The bundle currently has logstash-codec-json locked at 0.1.7.
Try running `bundle update logstash-codec-json`
/tmp/logstash-2.0.0.dev %
```

Gemfile
```diff
@@ -2,4 +2,8 @@
 # If you modify this file manually all comments and formatting will be lost.
 
 source "https://rubygems.org"
-gemspec :name => "logstash-core"
+gem "logstash-core", "2.0.0.dev", :path => "."
+gem "ci_reporter", "1.9.3", :group => :development
+gem "simplecov", :group => :development
+gem "coveralls", :group => :development
+gem "rspec", "~> 2.14.0", :group => :development
```
purbon: This has been usually loaded throw devutils till now, do we want to change that?

Gemfile
```diff
@@ -2,4 +2,8 @@
 # If you modify this file manually all comments and formatting will be lost.
 
 source "https://rubygems.org"
-gemspec :name => "logstash-core"
+gem "logstash-core", "2.0.0.dev", :path => "."
+gem "ci_reporter", "1.9.3", :group => :development
```
purbon: I don't get the context of this change, what is the criteria to have dependencies on the gemspec or in the Gemfile? why only the testing dependencies are here? Do we keep only in the gemspec dependencies that are runtime? 

is there any problem on grouping all the testing dependencies in the Gemfile like:

```
group :developement do 
 blablabla
end
```

Gemfile
colinsurprenant: by «this» you mean `rspec`?

purbon: yes.

ph:
Did a few tests on this PR, so far looking really good, great cleanup!

I am running into a minor issue, I have created a package with `artifact:tar` and ran the few usual commands list, install, uninstall.
When running the install/uninstall command I get this error:

    /t/logstash-2.0.0.dev ❯❯❯ bin/plugin uninstall logstash-output-elasticsearch
    Uninstalling logstash-output-elasticsearch
    /private/tmp/logstash-2.0.0.dev/vendor/local_gems/2e85d104/logstash-core-2.0.0.dev-java/lib/logstash/version.rb:3 warning: already initialized constant LOGSTASH_VERSION

Gemfile
colinsurprenant: the reason is that the testing dependencies do not belong in the `logstash-core` gem. The testing by Jenkins is done at the package level and so the testing dependencies belong in the top level package `Gemfile`. If you think about it, there is no relation with `ci_reporter` and the `logstash-core` gem, nothing in  `logstash-core` requires `ci_reporter`. The reason we never did that before is that it was not structured this way.

I do not think  `ci_reporter` belongs in `logstash-devutils` either, `ci_reporter` is really a package level tool used by Jenkins. Nothing in `logstash-devutils` depends on `ci_reporter`.

For the grouping, yes there is a problem. The `Gemfile` is managed by our `LogStash::Gemfile` class which does not (yet) support the `group` construct, it only support options hashes for the `gem` statements.

purbon: I agree ```ci_reporter``` should be in here. Another questions is what to do with rspecs and the coverage deps, I see then living in the gemspec, because all gem specific dependencies are there. I explain myself, If I want to run test I want test+coverage, but not to use ci_reports.

If finally all this deps are in the Gemfile, devutils have to be here too as it's mostlly all things for testing.

- purbon

colinsurprenant: very good point for `rspec`, but first, for the same reasons as `ci_reporter` I think `coveralls`'s place is in the to Gemfile. Nothing depends on `coveralls` other than the top level package.

As for `rspec`, I do think it belongs in `logstash-devutils` since to run specs in that gem, you need rspec. So in that respect, it is probably redundant to also have it in the top Gemfile. I'll see if we can clean that, but I don't think we should delay merging just for that, but I'll try to see if there is a problem in just leaving it in `logstash-devutils`.

ph:
More context on my previous comment, before generating the package with `rake artifact:tar`, I did rake `test:install-core` and `bin/rspec` than generated the package.

colinsurprenant:
please jenkins, test this.

colinsurprenant:
@ph this problem should be solved with last commit https://github.com/elastic/logstash/commit/ffc0740345ff235f9a780c28da4a08c101c4cf4b - the problem seems related to the gemspec requiring the version file using an absolute path and eventually the environment doing a require on "logstash/version" - because it uses different paths, it will load the file twice and produce that constant redefinition error.

ph:
@colinsurprenant works for me +1

Gemfile
colinsurprenant: hm. the Gemfile does not depend on `logstash-devutils` but does depends on `rspec` for running specs from the top level I think it's good to declare `rspec` as a dependency here. In other words, its not because some gems you depend on depend themselves on `rspec` that you should not declare this dependency.

suyograo:
LGTM to merge. Re-ran tests on OS X, looks good

talevy:
We should give more descriptive warnings on things like this:
```
/talevy-logstash (pr/3111)$ bin/plugin update
ERROR: Update is not supported for manually defined plugins or local .gem plugin installations: logstash-core
```

As a user I should not have to understand all of this / I probably won't.

suyograo:
@talevy Its mostly developers who will install a local gem and update it, not the typical user -- who will get the updated versions from RubyGem. That said, it is a bug we talked about fixing. Agree on re-wording the warning too. I opened https://github.com/elastic/logstash/issues/3157 to track it.

colinsurprenant:
@suyograo @talevy and the public packages will not use a local logstash-core but will use the pubished one so this error will not show up. 

I any case as said in #3157 the strategy will be to just ignore local gems and update the rest so in both case this will go away.

suyograo:
:ship:

PR #3114, opened 2015-04-27
Clean up the all_plugins_skip_list to be up to date

purbon:
The skip list is used during test and documentation generation to be sure not to use plugins that are not supposed to. This PR updates this list to be up to date whitelisting a list of plugins that got content enough to be ready for documentation and testing.

### DIFF
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -2,15 +2,14 @@ namespace "artifact" do
 
   def package_files
     [
-      ".bundle/config",
       "LICENSE",
       "CHANGELOG",
       "CONTRIBUTORS",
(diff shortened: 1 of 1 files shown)


jsvd:
are we also using fetch_all_plugins and is_released? here to select the plugins for test/docgen?

purbon:
yap, it's used in the docs.rake file. The motivation of this PR is that some plugins, like the neo4j ones were missing from our documentation.

colinsurprenant:
`fetch_all_plugins` is used in `rakelib/plugin.rake` for the `plugin:install-all` task which is used by `test:install-all` in `ci/ci_test.sh` and in the `docs:generate` task.
so if the tests are passing, the plugins working and the doc correctly generated I am LGTM.

Issue #3120, opened 2015-04-27
[docs] Asciidoc generation broken 


suyograo:
Tried to use `rake docs:generate` which internally installs all plugins on core and generates docs.

Also tried the command directly -- 

```rb
suyog@machine:~/ws/logstash (master)$ bin/logstash docgen -o asciidoc_generated vendor/bundle/jruby/1.9/gems/logstash-codec-line-0.1.6/lib/logstash/codecs/line.rb
io/console not supported; tty will not be manipulated
NameError: uninitialized constant LogStashConfigAsciiDocGenerator::LogStash::Environment
  const_missing at org/jruby/RubyModule.java:2726
       generate at /Users/suyog/ws/logstash/docs/asciidocgen.rb:174
            run at /Users/suyog/ws/logstash/lib/logstash/runner.rb:95
           each at org/jruby/RubyArray.java:1613
            run at /Users/suyog/ws/logstash/lib/logstash/runner.rb:94
           call at org/jruby/RubyProc.java:271
            run at /Users/suyog/ws/logstash/lib/logstash/runner.rb:131
           call at org/jruby/RubyProc.java:271
     initialize at /Users/suyog/ws/logstash/vendor/bundle/jruby/1.9/gems/stud-0.0.19/lib/stud/task.rb:12
```

suyograo:
Tested, works now

```sh
suyog@machine:~/ws/logstash (pr/3122)$ bin/logstash docgen -o asciidoc_generated vendor/bundle/jruby/1.9/gems/logstash-codec-line-0.1.6/lib/logstash/codecs/line.rb
io/console not supported; tty will not be manipulated
Generating docs for vendor/bundle/jruby/1.9/gems/logstash-codec-line-0.1.6/lib/logstash/codecs/line.rb
```

PR #3122, opened 2015-04-28
Fix the documentation generator process

purbon:
As in #3120 documentation generator is broken. This PR fix this issue. To run the documentation generator you can do it by running the task ```rake docs:generate```.

There is still however one open question for this PR, in the past we had the difference between contrib plugins and supported ? do we still have this ? how do we know it now that they are external ? For this PR, everything that is in the DEFAULT_PLUGINS is names as supported, otherwise is not.

There is also the need to refactor the documentation generator as the code is messy, however this is for another PR.

### DIFF
--- a/docs/asciidocgen.rb
+++ b/docs/asciidocgen.rb
@@ -4,6 +4,7 @@ require "optparse"
 
 $: << Dir.pwd
 $: << File.join(File.dirname(__FILE__), "..", "lib")
+$: << File.join(File.dirname(__FILE__), "..", "rakelib")
 
 require "logstash/config/mixin"
 require "logstash/inputs/base"
@@ -11,6 +12,7 @@ require "logstash/codecs/base"
 require "logstash/filters/base"
 require "logstash/outputs/base"
 require "logstash/version"
+require "rakelib/default_plugins"
 
 class LogStashConfigAsciiDocGenerator
   COMMENT_RE = /^ *#(?: (.*)| *$)/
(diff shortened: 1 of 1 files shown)


purbon:
@suyograo @untergeek can you review this ?

suyograo:
Tested, works now

```sh
suyog@machine:~/ws/logstash (pr/3122)$ bin/logstash docgen -o asciidoc_generated vendor/bundle/jruby/1.9/gems/logstash-codec-line-0.1.6/lib/logstash/codecs/line.rb
io/console not supported; tty will not be manipulated
Generating docs for vendor/bundle/jruby/1.9/gems/logstash-codec-line-0.1.6/lib/logstash/codecs/line.rb
```

jsvd:
Whitespace is quite messed up in this PR, probably the result of automatic indentation by the text editor?

running `rake docs:generate` but this takes a while.
Also, @suyograo and @purbon, this way this ["docs:generate" task will call "plugins:install-all"](https://github.com/elastic/logstash/blob/master/rakelib/docs.rake#L7) but then only generate docs for [default plugins](https://github.com/elastic/logstash/pull/3122/files#diff-c7022f14c4d92f3c75a45e5b64ec5fa9R31)

purbon:
@jsvd I fixed the indentations problems this PR had.

purbon:
test this please

jsvd:
the initialize method is still misaligned :)

kurtado:
LGTM, as far as I can tell, if the plugin is NOT listed here:
https://github.com/elastic/logstash/blob/master/rakelib/default_plugins.rb#L5
it should include the banner. Correct?

purbon:
@kurtado it makes sense to me, so lets go ahead and merge it for now. @suyograo are you ok with this to be merge for now?

suyograo:
Sure LGTM

PR #3128, opened 2015-04-28
slim down the skip list from plugins that are now released and accesible

purbon:
This list was intended to be used to keep away empty plugins, however some of them got now graduated, so are fully released and testable in our infrastructure.  This PR aims to have this plugins up and running as expected for our configuration.

- purbon


purbon:
@suyograo @jsvd can you review that ?

suyograo:
LGTM

PR #3159, opened 2015-05-05
support remote or local gem packaging

colinsurprenant:
rake artifact:tar will either build the local `logstash-core` gem and embed it in the package if the `Gemfile` contains

    gem "logstash-core", :path => "."

or othewise rely on the Rubygems  `logstash-core` if there is no :path specification

### DIFF
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -51,7 +51,20 @@ namespace "artifact" do
     File.open(".bundle/config", "w") { }
   end
 
-  task "prepare" => ["bootstrap", "plugin:install-default", "plugin:install-local-logstash-core-gem", "clean-bundle-config"]
+  # locate the "gem "logstash-core" ..." line in Gemfile, and if the :path => "." option if specified
+  # build and install the local logstash-core gem otherwise just do nothing, bundler will deal with it.
+  task "install-logstash-core" do
+    lines = File.readlines("Gemfile")
+    matches = lines.select{|line| line[/^gem\s+["']logstash-core["']/i]}
+    abort("ERROR: Gemfile format error, need a single logstash-core gem specification") if matches.size != 1
+    if matches.first =~ /:path\s*=>\s*["']\.["']/
+      Rake::Task["plugin:install-local-logstash-core-gem"].invoke
+    else
+      puts("[artifact:install-logstash-core] using logstash-core from Rubygems")
+    end
+  end
+
+  task "prepare" => ["bootstrap", "plugin:install-default", "install-logstash-core", "clean-bundle-config"]
 
   desc "Build a tar.gz of logstash with all dependencies"
   task "tar" => ["prepare"] do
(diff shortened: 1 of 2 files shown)


rakelib/artifacts.rake
```diff
@@ -51,7 +51,20 @@ namespace "artifact" do
...
+  # locate the "gem "logstash-core" ..." line in Gemfile, and if the :path => "." option if specified
+  # build and install the local logstash-core gem otherwise just do nothing, bundler will deal with it.
+  task "install-logstash-core" do
+    lines = File.readlines("Gemfile")
+    matches = lines.select{|line| line[/^gem\s+["']logstash-core["']/i]}
+    abort("ERROR: Gemfile format error, need a single logstash-core gem specification") if matches.size != 1
+    if matches.first =~ /:path\s*=>\s*["']\.["']/
+      Rake::Task["plugin:install-local-logstash-core-gem"].invoke
+    else
+      puts("[artifact:install-logstash-cor] using logstash-core from Rubygems")
```
ph: small typo
[artifact:install-logstash-core]

colinsurprenant: fixed

rakelib/plugin.rake
```diff
@@ -61,6 +61,8 @@ namespace "plugin" do
     Dir["logstash-core*.gem"].each do |gem|
       rm(gem)
     end
```
ph: I know this isnt part of this PR, but can we use a more specific path here?

ph: I am a bit scare of this *globly* path :fearful:

colinsurprenant: I don't think it's worth bothering... but we can open an issue if there is a good reaso to be more specific

colinsurprenant: it will only glob .gem files in the current path.... what is scary? (maybe I'm missing something?)

ph: Well, it will check all the possible path, it will be probably faster if we use a more specific path?

suyograo:
Works for me. Replaced `Gemfile` to point `logstash-core` to RubyGems

```
artifact:install-logstash-cor] using logstash-core from Rubygems
```

And in `Gemfile.jruby-1.9.lock`
```
...
logstash-codec-rubydebug (0.1.7)
      awesome_print
      logstash-core (>= 1.4.0, < 2.0.0)
    logstash-core (1.5.0.rc4.snapshot1-java)
      cabin (~> 0.7.0)
      clamp (~> 0.6.5)
...
```

rakelib/plugin.rake
colinsurprenant: waht do you mean all possible paths??? there is no double-star in the glob `**` ??

ph: Ah right, no **, should have double checked that, nothing to change.

suyograo:
LGTM, tested

ph:
LGTM

colinsurprenant:
:shipit:

PR #3179, opened 2015-05-06
Allow to update all the plugins and skip locally defined gems fixes #3157

### LINKED ISSUE #3157: bin/plugin update fails when there is a local gem installed
1. Build a local `foo-logstash-plugin.gem` and install on logstash core
2. Do `bin/plugin update`

it fails with: 

```sh
ERROR: Update is not supported for manually defined plugins or local .gem plugin installations: foo-logstash-plugin
```

We should: 
1. Ignore this local gem(s) and update remaining gems installed via RubyGems
2. Re-word the warning so it is more developer friendly.

ph:
Replace the Error by a simple warning and skip locally installed gems.
fixes #3157

### DIFF
--- a/lib/pluginmanager/update.rb
+++ b/lib/pluginmanager/update.rb
@@ -3,28 +3,33 @@ require "jar_install_post_install_hook"
 require "file-dependencies/gem"
 
 class LogStash::PluginManager::Update < LogStash::PluginManager::Command
+  REJECTED_OPTIONS = [:path, :git, :github]
+
   parameter "[PLUGIN] ...", "Plugin name(s) to upgrade to latest version", :attribute_name => :plugins_arg
 
   def execute
     local_gems = gemfile.locally_installed_gems
 
-    if update_all? && !local_gems.empty?
-      error_plugin_that_use_path!(local_gems)
-    else
-      plugins_with_path = plugins_arg & local_gems
-      error_plugin_that_use_path!(plugins_with_path) if plugins_with_path.size > 0
+    if local_gems.size > 0
+      if update_all?
+        plugins_with_path = local_gems.map(&:name)
+      else
+        plugins_with_path = plugins_arg & local_gems.map(&:name)
+      end
+
+      warn_local_gems(plugins_with_path)
     end
 
     update_gems!
   end
 
   private
-  def error_plugin_that_use_path!(plugins)
-    signal_error("Update is not supported for manually defined plugins or local .gem plugin installations: #{plugins.collect(&:name).join(",")}")
-  end
-
   def update_all?
     plugins_arg.size == 0
+  end
+
+  def warn_local_gems(plugins_with_path)
+    puts("Update is not supported for manually defined plugins or local .gem plugin installations, skipping: #{plugins_with_path.join(", ")}")
   end
 
   def update_gems!
@@ -34,14 +39,15 @@ class LogStash::PluginManager::Update < LogStash::PluginManager::Command
     # remove any version constrain from the Gemfile so the plugin(s) can be updated to latest version
     # calling update without requiremend will remove any previous requirements
     plugins = plugins_to_update(previous_gem_specs_map)
-    plugins
-      .select { |plugin| gemfile.find(plugin) }
-      .each { |plugin| gemfile.update(plugin) }
+    filtered_plugins = plugins.map { |plugin| gemfile.find(plugin) }
+      .compact
+      .reject { |plugin| REJECTED_OPTIONS.any? { |key| plugin.options.has_key?(key) } }
+      .each { |plugin| gemfile.update(plugin.name) }
 
     # force a disk sync before running bundler
     gemfile.save
 
-    puts("Updating " + plugins.join(", "))
+    puts("Updating #{filtered_plugins.collect(&:name).join(", ")}")
 
     # any errors will be logged to $stderr by invoke!
     # Bundler cannot update and clean gems in one operation so we have to call the CLI twice.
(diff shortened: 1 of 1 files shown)


suyograo:
@purbon review?

lib/pluginmanager/update.rb
```diff
@@ -8,25 +8,28 @@ class LogStash::PluginManager::Update < LogStash::PluginManager::Command
   def execute
     local_gems = gemfile.locally_installed_gems
 
-    if update_all? && !local_gems.empty?
-      error_plugin_that_use_path!(local_gems)
-    else
-      plugins_with_path = plugins_arg & local_gems
-      error_plugin_that_use_path!(plugins_with_path) if plugins_with_path.size > 0
+    if local_gems.size > 0
```
purbon: Why not checking if there are local_gems with ``` if !local_gems.empty? ``` might be more intutive than checking for size > 0. Is there any reason for this change that I might be missing ?

lib/pluginmanager/update.rb
```diff
@@ -34,14 +37,15 @@ def update_gems!
     # remove any version constrain from the Gemfile so the plugin(s) can be updated to latest version
```
purbon: Are we actually locking plugins to versions throw the Gemfile? I don't think so, isn't? is this comment outdated? for now I understand we lock plugin version in the Gemfile.lock, isn't?

lib/pluginmanager/update.rb
```diff
@@ -34,14 +37,15 @@ def update_gems!
...
     # calling update without requiremend will remove any previous requirements
     plugins = plugins_to_update(previous_gem_specs_map)
-    plugins
-      .select { |plugin| gemfile.find(plugin) }
-      .each { |plugin| gemfile.update(plugin) }
+    filtered_plugins = plugins.collect { |plugin| gemfile.find(plugin) }
+      .compact
+      .reject { |plugin| plugin.options.has_key?(:path) }
```
purbon: what about plugins using a :git option ? this is actually nice when we want to build test with remote repos :+1:

lib/pluginmanager/update.rb
```diff
@@ -61,7 +65,9 @@ def plugins_to_update(previous_gem_specs_map)
     if update_all?
       previous_gem_specs_map.values.map{|spec| spec.name}
     else
-      not_installed = plugins_arg.select{|plugin| !previous_gem_specs_map.has_key?(plugin.downcase)}
+      # If the plugins isn't available in the gemspec or in
```
purbon: I would update/add, ```or is in the gemfile dfined with a local path```, as it's now  it took me a while to understand that you don't update the plugins if:

* Not in the gemspec
* Defined with local path in the gemfile.

lib/pluginmanager/update.rb
```diff
@@ -61,7 +65,9 @@ def plugins_to_update(previous_gem_specs_map)
     if update_all?
       previous_gem_specs_map.values.map{|spec| spec.name}
     else
-      not_installed = plugins_arg.select{|plugin| !previous_gem_specs_map.has_key?(plugin.downcase)}
+      # If the plugins isn't available in the gemspec or in 
+      # the gemfile defined with a local path, the plugin is not installed
+      not_installed = plugins_arg.select{|plugin| !previous_gem_specs_map.has_key?(plugin.downcase) && !gemfile.find(plugin) }
```
purbon: is the call to ```!gemfile.find(plugin)``` sure to find plugins not in the gemfile ? are we suposing some state with this function, so we don't need somehow to check for the ```:path``` option ? 

Related to this, is there of any importance here the ones defined with the :git option?

purbon:
I see we don't have a test suite for the plugin manager, it would be a very nice to have. I will add an issue for that.

purbon:
To test this I basically did:

* Build a local plugin.
* Install the plugin throw bin/plugin install.
* call bin/plugin update

the output was the expected:

```
╰─$ bin/plugin update
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: logstash-core, logstash-input-neo4j
Updating logstash-filter-clone, logstash-filter-multiline, logstash-filter-mutate, logstash-input-generator, logstash-input-stdin, logstash-input-tcp, logstash-output-stdout
No plugin updated
```
(note: the local plugin was logstash-input-neo4j)

lib/pluginmanager/update.rb
ph: I usually go with positive conditionals, this is why I've changed it to check for the size instead of using `empty?`
if we had active_support I could go with `local_gems.present?`, ahh rails ;)

lib/pluginmanager/update.rb
ph: If you do `bin/plugin install --version '1.2' logstash-input-lumberjack` if will use a version in the gemfile, to update this plugin we remove the version and update to the latest.

lib/pluginmanager/update.rb
ph: Good catch! I will reject :git, :github and :path.

lib/pluginmanager/update.rb
purbon: active_support, jajajaj! well, rails is another world. at the end is everything mostly the same thing, but done in the ruby side or in the core. 

I asked because for me, size is not that self explanatory, but at the end is personal taste, good to go as it is now :+1:

lib/pluginmanager/update.rb
purbon: makes sense!

lib/pluginmanager/update.rb
ph: +1 did a small change on the comment.

lib/pluginmanager/update.rb
ph: Gemfile#find, only look in the current gemfile it will find normal gem, gems with a path, git or github option.

`LogStash::PluginManager.find_plugins_gem_specs`, look all the available specification.

ph:
Did a few changes from your review @purbon

purbon:
LGTM

PR #3188, opened 2015-05-07
Few minor improvements and pending cleanups

purbon:
* Clean up pending development dependencies from the logstash-core gemspec, including a crossed reference to rspec and a reference to logstash-devutils.
* Add gems as a build dependency in the Gemfile, is used when doing an install-all task, and failed in the ci env (like octokit)
  * Fix this kind of error: http://build-eu-00.elastic.co/view/LS%201.5/job/Logstash_1.5_All_Plugins_test/460/console


ph:
LGTM

tested with the following
```
rake test:install-core
rake test

rake test:install
rake test:plugins
```

PR #3194, opened 2015-05-07
Fix the documentation generation

### LINKED ISSUE #3184: [Docs] Class description is not getting converted to asciidoc
Trying to generate a doc after installing all plugins. First off, `rake docs:generate-docs` is not working again. I then tried manually generating docs; the asciidoc is created without the class description blob. Every other section is generated fine. 

```
bin/logstash docgen -o asciidoc ./vendor/bundle/jruby/1.9/gems/logstash-input-heartbeat-0.1.6/lib/logstash/inputs/heartbeat.rb
```

purbon:
Fix #3184 by fixing the pattern for the class definition line, otherwise sometimes there is no need to have it defined like ```::LogStash:...```, so the first ```::``` should be optional.

This leats the generation process create the class description again.

I generated all the docs and check a few and saw this part always generated.

As a side efect, it cleanup a bit the rake task used to generate the documentation as the pattern used was very long.

### DIFF
--- a/rakelib/docs.rake
+++ b/rakelib/docs.rake
@@ -1,7 +1,6 @@
 namespace "docs" do
 
   task "generate" do
-    Rake::Task['dependency:octokit'].invoke
     Rake::Task['plugin:install-all'].invoke
     Rake::Task['docs:generate-docs'].invoke
     Rake::Task['docs:generate-index'].invoke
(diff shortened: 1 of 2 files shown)


purbon:
@suyograo can you review this ?

rakelib/docs.rake
```diff
@@ -9,9 +8,9 @@ namespace "docs" do
 
   task "generate-docs" do
     require "bootstrap/environment"
-
-    list = Dir.glob("#{LogStash::Environment.logstash_gem_home}/gems/logstash-*/lib/logstash/{input,output,filter,codec}s/*.rb").join(" ")
-    cmd = "bin/logstash docgen -o asciidoc_generated #{list}"
+    pattern = "#{LogStash::Environment.logstash_gem_home}/gems/logstash-*/lib/logstash/{input,output,filter,codec}s/*.rb"
+    list    = Dir.glob(pattern).join(" ")
```
suyograo: Extra spaces?

suyograo:
Tested with `bin/logstash docgen -o asciidoc/new ./vendor/bundle/jruby/1.9/gems/logstash-input-heartbeat-0.1.6/lib/logstash/inputs/heartbeat.rb`. Description is generated now. Thanks @purbon 

Merge after resolving the new spaces introduced

rakelib/docs.rake
purbon: Nope, I use to have all ```=``` alligned. actually the way it was before it's uncommon as a coding style.

PR #3236, opened 2015-05-13
[Old-docs] Update changes to Elastic brand

suyograo:


docs/learn.md
```diff
@@ -28,8 +28,8 @@ The slides are available online here: [slides](http://semicomplete.com/presentat
 
 ## Getting Help
 
-There's [documentation](.) here on this site. If that isn't sufficient, you can
-email the mailing list (logstash-users@googlegroups.com). Further, there is also
+There's [documentation](http://www.elastic.co/guide/en/logstash/current/index.html) here on this site. If that isn't sufficient, you can
```
jordansissel: I don't think we should make this change.  If we do this, then folks reading the docs for 1.4.3 will be directed at the docs for `current` (1.5.x), which is not right.

After 1.5.x, we can probably delete docs/learn.md since it's not used in the new docs site, right?

docs/learn.md
```diff
@@ -38,9 +38,9 @@ for such things, that works for me, too.)
...
 ## Download It
 
-[Download logstash-%VERSION%](https://download.elasticsearch.org/logstash/logstash/logstash-%VERSION%.tar.gz)
+[Download logstash-%VERSION%](https://download.elastic.co/logstash/logstash/logstash-%VERSION%.tar.gz)
 
 ## What's next?
 
-Try this [guide](tutorials/getting-started-with-logstash) for a simple
-real-world example getting started using logstash.
+Try this [guide](http://www.elastic.co/guide/en/logstash/current/getting-started-with-logstash.html) for a simple
```
jordansissel: same comment here; we don't want users reading the 1.4.3 docs to be pointed at the `current` guide docs, since that will confuse them.

docs/learn.md
suyograo: you are right, i will undo this.

suyograo:
@jordansissel updated

jordansissel:
LGTM

PR #3251, opened 2015-05-15
Make untar extract message only when DEBUG=true

purbon:
small clean up improvement.


jsvd:
since this a rake task, maybe we could check if rake was called with -v (verbose)?

purbon:
@jsvd all this options are lost because of the restart with the internal rake, so we can not do that.

rakelib/z_rubycheck.rake
```diff
@@ -1,6 +1,6 @@
 if ENV['USE_RUBY'] != '1'
   if RUBY_ENGINE != "jruby" or Gem.ruby !~ /vendor\/jruby\/bin\/jruby/
-    puts "Restarting myself under Vendored JRuby (currently #{RUBY_ENGINE} #{RUBY_VERSION})"  if $DEBUG
+    puts "Restarting myself under Vendored JRuby (currently #{RUBY_ENGINE} #{RUBY_VERSION})"  if ENV['DEBUG']
```
jsvd: take this opportunity to remove the extra space before the "if" :)

jsvd:
LGTM

PR #3252, opened 2015-05-15
Use windows short path to support launching logstash from a path with spaces

wiibaa:
Without this little `s` and a path with space
````
D:\tmp\logstash\release\logstash-1.5.0 release>bin\logstash
'D:\tmp\logstash\release\logstash-1.5.0' is not recognized as an internal or external command,
operable program or batch file.
The system cannot find the path specified.
````
With this, it will resolve the path to `D:\tmp\logstash\release\LOGSTA~1.0RE` thus avoiding the space issue

Or do you see another solution ?

### DIFF
--- a/bin/logstash.bat
+++ b/bin/logstash.bat
@@ -2,7 +2,9 @@
 
 SETLOCAL
 
-set SCRIPT_DIR=%~dp0
+REM get logstash/bin absolute path: d => drive letter, p => path, s => use short names (no-spaces)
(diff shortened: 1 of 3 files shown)


jsvd:
can you do the same for bin/plugin.bat and add some url that describes the syntax for this stuff? it's quite cryptic

wiibaa:
@jsvd done

jsvd:
awesome, LGTM

PR #3264, opened 2015-05-18
Revert "fix string interpolation error when transforming characters"

purbon:
This reverts commit 65c789a24c5d9c11323c00a2b3554677626ab638 as it fix one thing, but was not fixing the hole ```\n``` issue.

Requested after talking to @jordansissel about the way the config_ast is working and expected to work.


purbon:
@jordansissel can you review so we can get rid of my mistake trying to fix the ```\n``` issue. ? thanks

jordansissel:
LGTM!

PR #3279, opened 2015-05-19
[1.4] Dont allow creating new files outside the document root

ph:
Before this Fix, people were able to use the string interpolation of the event (Event#sprintf)
to generate files outside the defined path. This pull request try to extract a document
root from the configuration and sandbox the creation of new file in that environment.

### DIFF
--- a/lib/logstash/outputs/file.rb
+++ b/lib/logstash/outputs/file.rb
@@ -1,22 +1,27 @@
 # encoding: utf-8
 require "logstash/namespace"
 require "logstash/outputs/base"
+require "logstash/errors"
 require "zlib"
 
 # This output will write events to files on disk. You can use fields
 # from the event as parts of the filename and/or path.
 class LogStash::Outputs::File < LogStash::Outputs::Base
+  FIELD_REF = /%\{[^}]+\}/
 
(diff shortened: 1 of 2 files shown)


suyograo:
This was already reviewed as part of https://github.com/logstash-plugins/logstash-output-file/pull/1

LGTM. I ran tests and they passed

```
suyog@machine:~/ws/logstash (pr/3279)$ bin/logstash rspec spec/outputs/file.rb
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :elasticsearch=>true, :broken=>true, :export_cypher=>true}
..File: The starting part of the path should not be dynamic. {:path=>"/%{name}", :level=>:error}
..File: The starting part of the path should not be dynamic. {:path=>"/%{name}", :level=>:error}
.File: The starting part of the path should not be dynamic. {:path=>"/%{name}", :level=>:error}
.File: The starting part of the path should not be dynamic. {:path=>"/%{name}", :level=>:error}
.File: The starting part of the path should not be dynamic. {:path=>"/%{name}", :level=>:error}
........

Finished in 3.94 seconds
15 examples, 0 failures
```

PR #3281, opened 2015-05-19
Strip line terminators when writing the comment for a branch

### LINKED ISSUE #2850: [regression] Syntax error with "if" on multiple lines
Just tried 1.5.0 RC2 and got a syntax error with my configuration file (worked fine in 1.4.2):
```
SyntaxError: (eval):160: syntax error, unexpected kOR
      or [tag] == "catalina.out"
       ^
        eval at org/jruby/RubyKernel.java:1107
  initialize at /opt/logstash/lib/logstash/pipeline.rb:32
     execute at /opt/logstash/lib/logstash/agent.rb:114
         run at /opt/logstash/lib/logstash/runner.rb:124
        call at org/jruby/RubyProc.java:271
         run at /opt/logstash/lib/logstash/runner.rb:129
        call at org/jruby/RubyProc.java:271
  initialize at /opt/logstash/vendor/bundle/jruby/1.9/gems/stud-0.0.19/lib/stud/task.rb:12
```
Here is the culprit block from the configuration file `filter` section:
```
  if [tag] == "foo"
    or [tag] == "catalina.out"
    or [tag] == "bar" {
```
Workaround: put all "or" conditions on the same line.

jordansissel:
Logstash's config compiler adds a comment to the compiled code, like

    if ..... # if [your] and [conditional]

The idea is to to help aid in reading the compiled logstash config.
However, if a conditional has newlines in it, the `#text_value` of
that conditional will have newlines, and we'll accidentally create
invalid ruby code which will fail with SyntaxError.

Prior to this change, the following Logstash config, under 1.5.0,
would cause a crash on startup:

    if [some]
      or [condition] {
      ...
    }

The cause was that Logstash would compile this to:

    if event("[some]"]) || event("[condition]") # if [some]
    or [condition]
      ...
    end

The 2nd line there `or [condition]` was intended to be on the line
above.

This change strips the line terminators \r and \n, just in case, and
provides a test case to cover.

I verified that this test case _fails_ without the config_ast.rb patch
and _succeeds_ with the patch.

Fixes #2850

### DIFF
--- a/lib/logstash/config/config_ast.rb
+++ b/lib/logstash/config/config_ast.rb
@@ -412,7 +412,7 @@ module LogStash; module Config; module AST
   class If < BranchEntry
     def compile
       children = recursive_inject { |e| e.is_a?(Branch) || e.is_a?(Plugin) }
-      return "if #{condition.compile} # if #{condition.text_value}\n" \
+      return "if #{condition.compile} # if #{condition.text_value.gsub(/[\r\n]/, " ")}\n" \
         << children.collect(&:compile).map { |s| s.split("\n", -1).map { |l| "  " + l }.join("\n") }.join("") << "\n"
     end
   end
--- a/spec/core/config_spec.rb
+++ b/spec/core/config_spec.rb
@@ -48,6 +48,24 @@ describe LogStashConfigParser do
   end
 
   context "#compile" do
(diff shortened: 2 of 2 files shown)


ph:
LGTM, test passes

PR #3290, opened 2015-05-21
Accessors: correct use of @lut and cleanups	

colinsurprenant:
while working on a Java implementation of the `Accessors` class, I realized commit b7ba48d393f9d17a8a4f377d8739af4fdf9c0343 introduced errors with the `@lut` usage resulting in less efficient code and also some unnecessary code complexity. 

the slowdown is really dependant of the type of config used but a config doing multiple lookups on the same field reference will benefit from this fix. 

benchmarks are showing 10-30% better performance. the following test results in a ~28% improvement.

- using fix/accessors at 320607aa04c9459c692e0238045d9001737241a6
```sh
bin/logstash -e 'input { generator { } } filter { mutate { add_field => { "[foo][bar][baz]" => "foo" "[fancy][pants]" => "bar" } } if [foo][bar][baz] == "baz" {} if [fancy][pants] == "baz" {} if [foo][bar][baz] == "baz" {} if [fancy][pants] == "baz" {} if [foo][bar][baz] == "baz" {} if [fancy][pants] == "baz" {} } output { stdout { codec => dots } }' | pv -Wbart > /dev/null
^C.8MiB 0:01:07 [30.7KiB/s] [27.4KiB/s]
```

- using master at 99741e57a2e36e1ca509c3af146efe1581ad452a
```sh
bin/logstash -e 'input { generator { } } filter { mutate { add_field => { "[foo][bar][baz]" => "foo" "[fancy][pants]" => "bar" } } if [foo][bar][baz] == "baz" {} if [fancy][pants] == "baz" {} if [foo][bar][baz] == "baz" {} if [fancy][pants] == "baz" {} if [foo][bar][baz] == "baz" {} if [fancy][pants] == "baz" {} } output { stdout { codec => dots } }' | pv -Wbart > /dev/null
^C48MiB 0:01:11 [21.1KiB/s] [21.3KiB/s]
```

### DIFF
--- a/lib/logstash/util/accessors.rb
+++ b/lib/logstash/util/accessors.rb
@@ -2,90 +2,120 @@
 
 require "logstash/namespace"
 require "logstash/util"
+require "thread_safe"
 
 module LogStash::Util
 
-  # PathCache is a singleton which globally caches a parsed fields path for the path to the
-  # container hash and key in that hash.
+  # PathCache is a singleton which globally caches the relation between a field reference and its
+  # decomposition into a [key, path array] tuple. For example the field reference [foo][bar][baz]
+  # is decomposed into ["baz", ["foo", "bar"]].
   module PathCache
     extend self
 
-    def get(accessor)
-      @cache ||= {}
-      @cache[accessor] ||= parse(accessor)
+    # requiring libraries and defining constants is thread safe in JRuby so
+    # PathCache::CACHE will be corretly initialized, once, when accessors.rb
+    # will be first required
+    CACHE = ThreadSafe::Cache.new
+
+    def get(field_reference)
+      # the "get_or_default(x, nil) || put(x, parse(x))" is ~2x faster than "get || put" because the get call is
+      # proxied through the JRuby JavaProxy op_aref method. the correct idiom here would be to use
+      # "compute_if_absent(x){parse(x)}" but because of the closure creation, it is ~1.5x slower than
+      # "get_or_default || put".
+      # this "get_or_default || put" is obviously non-atomic which is not really important here
+      # since all threads will set the same value and this cache will stabilize very quickly after the first
+      # few events.
+      CACHE.get_or_default(field_reference, nil) || CACHE.put(field_reference, parse(field_reference))
     end
 
-    def parse(accessor)
-      path = accessor.split(/[\[\]]/).select{|s| !s.empty?}
+    def parse(field_reference)
+      path = field_reference.split(/[\[\]]/).select{|s| !s.empty?}
       [path.pop, path]
     end
   end
 
-  # Accessors uses a lookup table to speedup access of an accessor field of the type
+  # Accessors uses a lookup table to speedup access of a field reference of the form
   # "[hello][world]" to the underlying store hash into {"hello" => {"world" => "foo"}}
   class Accessors
 
+    # @param store [Hash] the backing data store field refereces point to
     def initialize(store)
       @store = store
+
+      # @lut is a lookup table between a field reference and a [target, key] tuple
+      # where target is the containing Hash or Array for key in @store.
+      # this allows us to directly access the containing object for key instead of
+      # walking the field reference path into the inner @store objects
       @lut = {}
     end
 
-    def get(accessor)
-      target, key = lookup(accessor)
-      unless target.nil?
-        target.is_a?(Array) ? target[key.to_i] : target[key]
-      end
+    # @param field_reference [String] the field reference
+    # @return [Object] the value in @store for this field reference
+    def get(field_reference)
+      target, key = lookup(field_reference)
+      return nil unless target
+      target.is_a?(Array) ? target[key.to_i] : target[key]
     end
 
-    def set(accessor, value)
-      target, key = store_and_lookup(accessor)
+    # @param field_reference [String] the field reference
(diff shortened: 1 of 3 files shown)


lib/logstash/util/accessors.rb
```diff
@@ -48,44 +47,37 @@ def strict_set(accessor, value)
...
-      unless target.nil?
-        target.is_a?(Array) ? target.delete_at(key.to_i) : target.delete(key)
-      end
+      return nil unless target
+      target.is_a?(Array) ? target.delete_at(key.to_i) : target.delete(key)
     end
 
     def include?(accessor)
-      target, key = lookup_path(accessor)
+      target, key = lookup(accessor)
```
talevy: this eagerness to be `nil` is still something I find difficult to swallow in Ruby, but I get it.

```
x = nil
y, z = nil
y.nil? && z.nil? #true
```

talevy:
while we are updating this file. would you be able to add a little documentation with the roles of each variable. I find myself having to run through all the code to see how things are related each time.

talevy:
nice! my bad I guess for mis-using the cache appropriately before.

lib/logstash/util/accessors.rb
colinsurprenant: Yes, it makes sense once you understand the `a, b, ... = x, y, ...` semantic:
```ruby
irb(main):018:0> a, b = 1, 2
=> [1, 2]
irb(main):019:0> a
=> 1
irb(main):020:0> b
=> 2
```

`a, b = ...` is actually assigning an `Array` result for the RHS, first item to `a`, second item to `b` and so on. So what it on the RHS is automatically converted to an array. Now:

```ruby
irb(main):002:0> a, b = nil
=> [nil]
```
Since the RHS is a `[nil]` `Array` then 

```ruby
rb(main):021:0> [nil][0]
=> nil
irb(main):022:0> [nil][1]
=> nil
```

colinsurprenant:
@talevy absolutely, I will add a bit more documentation and also change a few identifiers. I think renaming the private functions to what they actually do did help a bit but I agree it's tricky to understand what is going on. This would actually make a good blog post I think :P

colinsurprenant:
@talevy added docs, let me know what you think

lib/logstash/util/accessors.rb
```diff
@@ -5,87 +5,111 @@
...
-    def lookup(accessor)
-      target, key = lookup_path(accessor)
-      if target.nil?
-        [target, key]
-      else
-        @lut[accessor] = [target, key]
-      end
+    # retrieve the [target, key] tuple associated with this field reference
+    # @param field_reference [String] the field referece
+    # @returm [[Object, String]] the  [target, key] tuple associated with this field reference
```
talevy: s/returm/return

colinsurprenant: fixed

lib/logstash/util/accessors.rb
```diff
@@ -5,87 +5,111 @@
...
+    def lookup_or_create(field_reference)
+      @lut[field_reference] ||= find_or_create_target(field_reference)
     end
 
-    def lookup_path(accessor)
-      key, path = PathCache.get(accessor)
+    # find the target container object in store for this field reference
+    # @param field_reference [String] the field referece
+    # @returm [Object] the target container object in store associated with this field reference
```
talevy: s/returm/return

spec/util/accessors_spec.rb
```diff
@@ -183,33 +167,35 @@
...
-    end
-
-    it "strinct_set should raise on invalid UTF-8 string encoding in array" do
-      str = "[hello]"
-      data = {}
-      accessors = LogStash::Util::Accessors.new(data)
-      expect { accessors.strict_set(str, ["foo", "bar \xED\xB9\x81\xC3"]) }.to raise_error
-    end
-  end
+  # should we move this in Event#validate_value specs?
```
talevy: is `Event#validate_value` used?

colinsurprenant: yes, for specs, see https://github.com/elastic/logstash-devutils/blob/master/lib/logstash/devutils/rspec/spec_helper.rb#L40

talevy: can we migrate these tests to there in this PR?

colinsurprenant: it's a bit out of scope. I can create an issue to add them.

talevy:
LGTM

ph:
LGTM, run test, Nice documentation improvements!

jsvd:
ran tests and benchmarks, LGTM

this code really is on the critical path, from what I understand, just by creating 2 less arrays on failed lookups gave this performance boost, nice one @colinsurprenant

colinsurprenant:
@jsvd the improvement is more about actually using the `@lut` which is actually the purpose of the `Accessors` class in the first place but this `@lut` cache lookup was "lost" in a bug fix commit.

ph:
@colinsurprenant The only thing I see here is the `@cache` is not thread safe, so multiple workers threads could access the cache table, maybe we could use this `ThreadSafe::Cache` from the thread_safe library? What do you think?

colinsurprenant:
@ph you are right. PathCache is now thread safe :P

ph:
LGTM

Issue #2901, opened 2015-03-26
Add configcheck to init and make sure to check config before trying to start


Mister-X-:
Using the deb package (1.5RC2), when using 'service' to start logstash, it will tell you that it started correctly if you have an invalid configuration. Java did, logstash started but exited immediately because the config is invalid.

So, this patch add an option to check the config and it also makes sure to check the config before starting.

```
diff --git a/pkg/logstash.sysv b/pkg/logstash.sysv
index fddc14d..d9954d0 100755
--- a/pkg/logstash.sysv
+++ b/pkg/logstash.sysv
@@ -115,6 +115,27 @@ force_stop() {
   fi
 }
 
+config_test() {
+  # Check if a config file exists
+  if [ ! "$(ls -A $LS_CONF_DIR/*.conf 2> /dev/null)" ]; then
+    log_failure_msg "There aren't any configuration files in $LS_CONF_DIR"
+    exit 1
+  fi
+
+  JAVA_OPTS=${LS_JAVA_OPTS}
+  HOME=${LS_HOME}
+  export PATH HOME JAVA_OPTS LS_HEAP_SIZE LS_JAVA_OPTS LS_USE_GC_LOGGING
+
+  test_args="-f ${LS_CONF_DIR} --configtest ${LS_OPTS}"
+  if [ "$1" = '1' ]; then
+    $program ${test_args} 2>/dev/null >/dev/null
+  else
+    $program ${test_args}
+  fi
+  [ $? -eq 0 ] && return 0
+  # Program not configured
+  return 6
+}
 
 case "$1" in
   start)
@@ -123,8 +144,14 @@ case "$1" in
     if [ $code -eq 0 ]; then
       echo "$name is already running"
     else
-      start
+      config_test 1
       code=$?
+      if [ $code -ne 0 ]; then
+       echo "Invalid config for $name. Re-run with configtest to see what's wrong"
+      else
+       start
+       code=$?
+      fi
     fi
     exit $code
     ;;
@@ -144,8 +171,12 @@ case "$1" in
     
     stop && start 
     ;;
+  configtest)
+    config_test
+    exit $?
+    ;;
   *)
-    echo "Usage: $SCRIPTNAME {start|stop|force-stop|status|restart}" >&2
+    echo "Usage: $SCRIPTNAME {start|stop|force-stop|status|restart|configtest}" >&2
     exit 3
   ;;
 esac
[...]

PR #3304, opened 2015-05-24
Add configcheck to init and make sure to check config before trying to start

Mister-X-:
Using the deb package (1.5RC2 and most likely RC3 and GA), when using 'service' to start logstash, it will tell you that it started correctly if you have an invalid configuration. Java did, logstash started but exited immediately because the config is invalid.

So, this patch add an option to check the config and it also makes sure to check the config before starting.

See elastic/logstash#2901

On a side note, in the headers, Required-Start doesn't need $syslog

### DIFF
--- a/pkg/logstash.sysv
+++ b/pkg/logstash.sysv
@@ -114,6 +114,27 @@ force_stop() {
   fi
 }
 
+config_test() {
+  # Check if a config file exists
+  if [ ! "$(ls -A $LS_CONF_DIR/*.conf 2> /dev/null)" ]; then
+    log_failure_msg "There aren't any configuration files in $LS_CONF_DIR"
+    exit 1
+  fi
+
+  JAVA_OPTS=${LS_JAVA_OPTS}
+  HOME=${LS_HOME}
+  export PATH HOME JAVA_OPTS LS_HEAP_SIZE LS_JAVA_OPTS LS_USE_GC_LOGGING
+
+  test_args="-f ${LS_CONF_DIR} --configtest ${LS_OPTS}"
+  if [ "$1" = '1' ]; then
+    $program ${test_args} 2>/dev/null >/dev/null
+  else
+    $program ${test_args}
+  fi
+  [ $? -eq 0 ] && return 0
+  # Program not configured
+  return 6
+}
 
 case "$1" in
   start)
@@ -122,8 +143,14 @@ case "$1" in
     if [ $code -eq 0 ]; then
       echo "$name is already running"
     else
-      start
+      config_test 1
       code=$?
+      if [ $code -ne 0 ]; then
+       echo "Invalid config for $name. Re-run with configtest to see what's wrong"
+      else
+       start
+       code=$?
+      fi
     fi
     exit $code
     ;;
@@ -143,8 +170,12 @@ case "$1" in
     
     stop && start 
     ;;
+  configtest)
+    config_test
+    exit $?
+    ;;
   *)
-    echo "Usage: $SCRIPTNAME {start|stop|force-stop|status|restart}" >&2
+    echo "Usage: $SCRIPTNAME {start|stop|force-stop|status|restart|configtest}" >&2
     exit 3
   ;;
 esac


purbon:
@Mister-X- you might need to sign the CLA, more info at https://www.elastic.co/contributor-agreement .. before we can move on. Thanks.

suyograo:
@Mister-X- can you sign the CLA please?

ThomasdOtreppe:
CLA is signed under this account. Both Mister-X- and this one belong to me, so should I redo a PR with this one or can I just confirm with my other account (Mister-X-) that they both belong to me?

Mister-X-:
I own both accounts.

ThomasdOtreppe:
Should I redo a PR with this one?

pkg/logstash.sysv
```diff
@@ -114,6 +114,27 @@ force_stop() {
   fi
 }
 
+config_test() {
+  # Check if a config file exists
+  if [ ! "$(ls -A $LS_CONF_DIR/*.conf 2> /dev/null)" ]; then
```
jordansissel: `LS_CONF_DIR` [defaults to a directory](https://github.com/Mister-X-/logstash/blob/master/pkg/logstash.sysv#L35), not a wildcard. Checking for `*.conf` here will falsely indicate that there are no configuration files if someone has a file named `$LS_CONF_DIR/fancy` (which Logstash normally would see as a config file)

pkg/logstash.sysv
```diff
@@ -114,6 +114,27 @@ force_stop() {
...
+    log_failure_msg "There aren't any configuration files in $LS_CONF_DIR"
```
jordansissel: Shouldn't Logstash deal with this problem?

pkg/logstash.sysv
```diff
@@ -114,6 +114,27 @@ force_stop() {
...
+    exit 1
+  fi
+
+  JAVA_OPTS=${LS_JAVA_OPTS}
+  HOME=${LS_HOME}
+  export PATH HOME JAVA_OPTS LS_HEAP_SIZE LS_JAVA_OPTS LS_USE_GC_LOGGING
+
+  test_args="-f ${LS_CONF_DIR} --configtest ${LS_OPTS}"
+  if [ "$1" = '1' ]; then
```
jordansissel: for quiet operation, I'd prefer a `quiet` function, something like:

```shell
quiet() {
  "$@" > /dev/null 2>&1
}
```

Then later:

```shell
quiet config_test
```

pkg/logstash.sysv
```diff
@@ -114,6 +114,27 @@ force_stop() {
...
+    $program ${test_args} 2>/dev/null >/dev/null
+  else
+    $program ${test_args}
+  fi
+  [ $? -eq 0 ] && return 0
+  # Program not configured
+  return 6
```
jordansissel: What does exit code 6 mean?

pkg/logstash.sysv
```diff
@@ -143,8 +170,12 @@ case "$1" in
     
     stop && start 
     ;;
+  configtest)
+    config_test
```
jordansissel: Should we make these the same names? (both configtest _or_ config_test?)

jordansissel:
@electrical thoughts on adding serverspec/puppet test coverage for this?

jordansissel:
@ThomasdOtreppe @Mister-X- If you edit the commit log and use an email address that you've signed the CLA with, then force push to your branch, this PR should get updated and see the CLA signed.

jordansissel:
I have a concern that this will add a surprising amount of startup delay to the 'start' command because it forces a config check (full java/jruby/logstash startup cost) on all start invocations.

pkg/logstash.sysv
aircrack-ng: We could let it deal with it but it's less than 1 sec check vs a few seconds for logstash to start and notice there are not files.

pkg/logstash.sysv
aircrack-ng: program is not configured

Check out the bottom of the page http://refspecs.linuxbase.org/LSB_3.1.0/LSB-Core-generic/LSB-Core-generic/iniscrptact.html

pkg/logstash.sysv
aircrack-ng: I can do that.

aircrack-ng:
Something I can do is only call config check when restarting where it's critical to check to avoid having an instance of logstash that cannot be started.

pkg/logstash.sysv
aircrack-ng: I missed that. I was used to other programs requiring .conf in order to load the file (and they were skipping everything that didn't contain that specific extension). I'll change it to wildcard.

ThomasdOtreppe:
@jordansissel  Would you prefer to do a check only when doing a restart?

ThomasdOtreppe:
@jordansissel Could you tell me if you prefer to a check on restart only (or not) so that I can update this PR?

jordansissel:
Ahh, sorry I thought I responded. My bad!

+1 on doing the check only on restart. If there's more demand, I'm open to adding it into the 'start' - but I think your idea of doing it by default on restart is a good compromise.

jordansissel:
(fyi, I still get CLA check failures on this PR)

suyograo:
@Mister-X- will try closing and reopening to see if CLA kicks in

ThomasdOtreppe:
@suyograo I will update this PR with this account (which has the CLA signed and linked to the company I work for). The other one hasn't.

Issue #3271, opened 2015-05-19
Update to JRuby 1.7.20


suyograo:
http://jruby.org/2015/05/05/jruby-1-7-20.html

mpasternacki:
This will make Logstash work properly on FreeBSD, as JRuby 1.7.20 includes (or no longer needs) https://github.com/jnr/jnr-ffi/pull/26. 1.7.19 still exhibited this issue:

Vanilla 1.5.0:

```
 $ ./bin/logstash pry
[…]
[1] pry(#<LogStash::Runner>)> JRUBY_VERSION
=> "1.7.19"
[2] pry(#<LogStash::Runner>)> File::Stat.new('/dev/null').dev_major
NotImplementedError: stat.st_dev unsupported or native support failed to load
from org/jruby/RubyFileStat.java:188:in `dev_major'
[3] pry(#<LogStash::Runner>)> 
```

1.5.0 with JRuby 1.7.20 vendored:

```
$ ./bin/logstash pry                                      
[…]
[1] pry(#<LogStash::Runner>)> JRUBY_VERSION
=> "1.7.20"
[2] pry(#<LogStash::Runner>)> File::Stat.new('/dev/null').dev_major
=> 113
[3] pry(#<LogStash::Runner>)> 
```

```
$ uname -a
FreeBSD portinari.local 11.0-CURRENT FreeBSD 11.0-CURRENT #0 r282300M: Sat May  2 00:27:44 CEST 2015     japhy@portinari.local:/usr/obj/usr/src/sys/MINE  amd64
```

suyograo:
Merged

Issue #2817, opened 2015-03-13
logstash throws exception with simple program


ralphbellofatto:
we are attempting to run logstash on a ppc64 machine.

To get around the jruby-complete not supporting ppc64 we obtained the latest git repo and compiled it locally under java 1.8 (It does not compile under java 1.7).

When we run one of the example programs we get:

```
[root@dccxcat jar]# /opt/logstash/bin/logstash -e 'input { stdin { } } output { stdout { codec => rubydebug } }'
goodnight moon

Mar 13, 2015 12:37:53 PM jnr.ffi.provider.jffi.NativeRuntime buildNativeTypeAliases
SEVERE: failed to load type aliases: java.lang.ClassNotFoundException: jnr.ffi.provider.jffi.platform.ppc64.linux.TypeAliases
{
       "message" => "goodnight moon",
      "@version" => "1",
    "@timestamp" => "2015-03-13T16:37:53.279Z",
          "host" => "dccxcat"
}

```
Can someone help out with what we are missing now...

We are running RHEL7 on a power7 ppc64 architecture machine after installing:
logstash-1.4.2-1_2c0f5a1.noarch.rpm

ralphbellofatto:
Per a suggestion on: https://github.com/elastic/logstash/issues/2747

![boogers](https://cloud.githubusercontent.com/assets/4139514/6689632/d2b58f9e-cc91-11e4-9bf3-95946c33c8b0.png)

we did:
```
mkdir /opt/logstash
http://download.elasticsearch.org/logstash/logstash/logstash-1.5.0.rc2.tar.gz
tar xzf logstash-1.5.0.rc2.tar.gz
cd logstash-1.5.0.rc2
cp /home/ralphbel/jruby/jruby-jruby-1_7/lib/jruby.jar ./vendor/jruby/lib/jruby.jar
```

While the above works for the stdin-> stdout script, logstash fails on ppc64 when I try to have digest a file with timestamps in it.

We installed RC2 as follows:
```
mkdir /opt/logstash
http://download.elasticsearch.org/logstash/logstash/logstash-1.5.0.rc2.tar.gz
tar xzf logstash-1.5.0.rc2.tar.gz
cd logstash-1.5.0.rc2
cp /home/ralphbel/jruby/jruby-jruby-1_7/lib/jruby.jar ./vendor/jruby/lib/jruby.jar
```

and then we constructed the following script
```
#!/bin/bash
# simple log stash test wrapper...


[...]

Issue #2747, opened 2015-03-03
logstash running on linux ppc64 crashes on startup.


ralphbellofatto:
When I run logstash on a linux ppc64 machine as follows:
```
 bin/logstash -e 'input { stdin { } } output { stdout {} }'
```
It crashes with
```
	LoadError: Could not load FFI Provider: (NotImplementedError) FFI not available: null
	 See http://jira.codehaus.org/browse/JRUBY-4583
	  require at org/jruby/RubyKernel.java:1085
	  require at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:55
	   (root) at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/ffi/ffi.rb:69
	  require at org/jruby/RubyKernel.java:1085
	   (root) at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:1
	  require at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:55
	  require at org/jruby/RubyKernel.java:1085
	   (root) at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/ffi.rb:1
	   (root) at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:1
	  require at file:/opt/logstash/vendor/jar/jruby-complete-1.7.11.jar!/META-INF/jruby.home/lib/ruby/shared/rubygems/core_ext/kernel_require.rb:55
	  require at org/jruby/RubyKernel.java:1085
	     LibC at /opt/logstash/lib/logstash/util/prctl.rb:4
	   (root) at /opt/logstash/lib/logstash/util/prctl.rb:3
	     main at /opt/logstash/lib/logstash/runner.rb:79
	   (root) at /opt/logstash/lib/logstash/runner.rb:215
```

[...]

PR #3312, opened 2015-05-25
update to jruby-1.7.20 and cleanup legacy patches

colinsurprenant:


colinsurprenant:
solves #3271 and relates to #2747 #2817

purbon:
I wonder if there is a way to test that the patch is no longer need with the new jruby? 

for the rest the patch looks good to me (LGTM).

/cheers

- purbon

colinsurprenant:
I manually ran specs for core and default-plugins. Also, this is related to an older version of jar-dependencies and we don't dynamically install jar dependencies into our plugins. Exception is kafka through jruby-kakfa, and tests/specs passed.

suyograo:
+1.. would like to get this baking on Jenkins

Issue #2598, opened 2015-02-12
DRY lib/logstash/util/filetools.rb and rakelib/fetch.rake


colinsurprenant:
once the `wip/bundler-move` branch is merged

colinsurprenant:
wip/bundler-move branch is merged in #2475

colinsurprenant:
@suyograo @purbon  unless this is in progress and/or almost finished I say we move it to post-rc1

colinsurprenant:
looks like `filetools.rb` could just be removed, it is not used anywhere...

ph:
We can close this, I've look at the code and we removed the files.

PR #3319, opened 2015-05-26
WiP: make minitar a non runtime dependency

purbon:
Minitar is only used for the build process, so it should be listed in the gemfile as build dependency and not included in the ```logstash-core.gemspec```.

This PR try to fix this.

### DIFF
--- a/rakelib/bootstrap.rake
+++ b/rakelib/bootstrap.rake
@@ -1,2 +1,1 @@
-
-task "bootstrap" => [ "vendor:all", "compile:all" ]
+task "bootstrap" => [ "vendor:gems", "vendor:all", "compile:all" ]
--- a/Gemfile
+++ b/Gemfile
@@ -14,3 +14,5 @@ gem "stud", "~> 0.0.19", :group => :build
 gem "fpm", "~> 1.3.3", :group => :build
 gem "rubyzip", "~> 1.1.7", :group => :build
 gem "gems", "~> 0.8.3", :group => :build
+gem "minitar", "~> 0.5.4", :group => :build
+
--- a/Gemfile.jruby-1.9.lock
+++ b/Gemfile.jruby-1.9.lock
@@ -7,7 +7,6 @@ PATH
       filesize (= 0.0.4)
       i18n (= 0.6.9)
       jrjackson (~> 0.2.8)
-      minitar (~> 0.5.4)
       pry (~> 0.10.1)
       stud (~> 0.0.19)
       treetop (< 1.5.0)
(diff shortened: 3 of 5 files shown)


rakelib/bootstrap.rake
```diff
@@ -1,2 +1 @@
-
-task "bootstrap" => [ "vendor:all", "compile:all" ]
+task "bootstrap" => [ "vendor:gems", "vendor:all", "compile:all" ]
```
colinsurprenant: vendor:all includes vendor:gems, so this is useless

purbon: not useless because I removed the all from the gems task. https://github.com/elastic/logstash/pull/3319/files#diff-77ee63fa86421d6bcb23e149dc0f8854L138

purbon: I try to make the gems installation go in the first place. There is still a bit of room for a refactoring to make this cleaner.

colinsurprenant: ah probably because you wanted to force the order before the jruby task. maybe its better to define the gems task as a dependency in the jruby task then?

colinsurprenant: which you did too. so you should not need to specify vendor:gems in the bootstrap task...

purbon: Also there https://github.com/elastic/logstash/pull/3319/files#diff-77ee63fa86421d6bcb23e149dc0f8854R63

colinsurprenant:
did you test it?

jordansissel:
Isn't Minitar used as a runtime dependency of file-dependencies gem?
(Bin/plugin install's vendor file handling)

colinsurprenant:
I see minitar used in `filetools.rb` but `filetools.rb` doesn't look like its being used anywhere in core or in any plugins O_o   ... this was reported in #2598 too

colinsurprenant:
@jordansissel oh yes, minitar is a runtime dependency of file-dependencies ... and file-dependencies is a logstash runtime dependency for the plugin manager.

colinsurprenant:
this http://dracoater.blogspot.ca/2013/10/extracting-files-from-targz-with-ruby.html looks like a good no-dependencies solution to replace minitar and per comments, it is much faster too.

purbon:
I think I will close this PR and clean up minitar throught having this https://github.com/purbon/logstash/blob/feature/plugins_offline/lib/bootstrap/util/compress.rb shared with the rest of the code base, makes sense? I can also update file-dependencies if necessary.

PR #3330, opened 2015-05-27
Settings doesnt persist the values to the bundle/config

ph:
Since we aren't persisting anything to disk the test to check for persisted values was obviously failling


suyograo:
LGTM

PR #3331, opened 2015-05-27
Fix permissions after install (debian package)

### LINKED ISSUE #3305: Debian package creates g+rwx for root on /etc/logstash?
This is probably not a major issue, but a user noted that the permissions on `/etc/logstash` create via our deb package are a bit wonky:

```
$ stat /etc/logstash
  File: ‘/etc/logstash’
  Size: 4096      	Blocks: 8          IO Block: 4096   directory
Device: fc00h/64512d	Inode: 1044922     Links: 3
Access: (0775/drwxrwxr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2015-05-24 17:16:45.946652412 -0700
Modify: 2015-05-20 18:32:16.179115456 -0700
Change: 2015-05-20 18:32:16.179115456 -0700
 Birth: -
```

`/etc/logstash` has root:root permissions and the root group has rwx to it. I'm not sure if this is a security issue (I doubt many admins assign any other accounts to the root group) but it is an odd permission.

untergeek:
/etc/logstash was getting 775 perms in the debian package.

This manually sets 755 to /etc/logstash

fixes #3305

### DIFF
--- a/pkg/debian/after-install.sh
+++ b/pkg/debian/after-install.sh
@@ -3,3 +3,4 @@
 chown -R logstash:logstash /opt/logstash
 chown logstash /var/log/logstash
 chown logstash:logstash /var/lib/logstash
+chmod 755 /etc/logstash


suyograo:
LGTM

PR #3332, opened 2015-05-28
Better implementation of the no .bundle/config that respect the bundler contract in a running process

ph:
The previous fix for disabling the `.bundle/config` wasn't not respecting the bundler contract. 
The failling tests were not wrong, in fact they exposed the issue that the configuration was transient and the underlying hash of bundler was not correctly keeping the updated values.

This patch make sure the hash is updated with the new or deleted value without persisting the change to disk.

### DIFF
--- a/lib/bootstrap/bundler.rb
+++ b/lib/bootstrap/bundler.rb
@@ -16,6 +16,13 @@ module LogStash
       # of the application
       ::Bundler::Settings.module_exec do
         def set_key(key, value, hash, file)
+          key = key_for(key)
+
+          unless hash[key] == value
+            hash[key] = value
+            hash.delete(key) if value.nil?
+          end
+
           value
         end
       end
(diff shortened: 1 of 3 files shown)


ph:
Tested with a generated package on ubuntu 12.04
```
vagrant@vagrant-ubuntu-precise-32:~$ /opt/logstash/bin/logstash version
logstash 2.0.0.dev
vagrant@vagrant-ubuntu-precise-32:~$ sudo /opt/logstash/bin/plugin install logstash-filter-cidr
Validating logstash-filter-cidr
Installing logstash-filter-cidr
Installation successful
vagrant@vagrant-ubuntu-precise-32:~$ /opt/logstash/bin/logstash version
logstash 2.0.0.dev
vagrant@vagrant-ubuntu-precise-32:~$ whoami
vagrant
```

ph:
@suyograo this fix the current build can you review and merge?

suyograo:
@ph LGTM, merging

Issue #3248, opened 2015-05-15
Update documentation to new 1.5 release


purbon:
We should revisit the documentation to be sure it's all naming 1.5 as the current GA, for example https://www.elastic.co/guide/en/logstash/current/package-repositories.html is still referencing to 1.4.x repos.

technosf:
Please check for voracity too. For example, on the getting started page there is this cmdline example:

  `bin/plugin -install lmenezes/elasticsearch-kopf`

Which results in an error:

    ERROR: Unrecognised option '-i'

    See: 'bin/plugin --help

Thanks!

untergeek:
Clearly that is in the [Elasticsearch Plugins (an aside)](https://www.elastic.co/guide/en/logstash/current/getting-started-with-logstash.html#_elasticsearch_plugins_an_aside) section.  Kopf is an Elasticsearch plugin, rather than a Logstash one.

We could be more clear about that, but it is mentioned as an Elasticsearch plugin, rather than a Logstash one.

technosf:
Yup - but the flag is wrong ;)

it's _bin/plugin install foobar_ 

No minus before _install_

untergeek:
Um, I know not what you mean:

```
bin/plugin -install lmenezes/elasticsearch-kopf
-> Installing lmenezes/elasticsearch-kopf...
Trying https://github.com/lmenezes/elasticsearch-kopf/archive/master.zip...
Downloading .....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................DONE
Installed lmenezes/elasticsearch-kopf into /Users/buh/elasticsearch-1.5.2/plugins/kopf
Identified as a _site plugin, moving to _site structure ...
```

It works just fine for me.

technosf:
Ah... You're right. My appologies - Although it is in the _logstash_ doc, it is noted as being for ES as _an aside_. Interesting that ED and LS have the same command but one is _-install_ and one is _install_

PR #3348, opened 2015-05-29
Docs: Remove ES plugin reference because it is confusing to users

suyograo:
The docs had an ES plugins aside, which is very confusing to users because LS has plugins too (plenty of them :) )
`bin/plugin -install lmenezes/elasticsearch-kopf` 

See here: https://github.com/elastic/logstash/issues/3248#issuecomment-104477776


untergeek:
:+1: 

lgtm

PR #3358, opened 2015-06-01
User/group deletion shouldn't be mandatory

briandef:
Currently, the debian prerm maintainer script for logstash demands to remove the logstash user/group from passwd/shadow files, and fails if it cannot do so.

Assuming that the logstash user is in passwd/shadow means that the logstash package will not finish uninstalling/removing/purging for systems with LDAP or any other user account database.

This change simply allows the maintainer script to complete (with error message) even if the user/group deletion requests fail. Then the package can finish uninstalling properly.

### DIFF
--- a/pkg/debian/before-remove.sh
+++ b/pkg/debian/before-remove.sh
@@ -4,10 +4,10 @@ if [ $1 = "remove" ]; then
   service logstash stop >/dev/null 2>&1 || true
 
   if getent passwd logstash >/dev/null ; then
-    userdel logstash
+    userdel logstash || true
   fi
 
   if getent group logstash >/dev/null ; then
-    groupdel logstash
+    groupdel logstash || true
   fi
 fi


jordansissel:
Ignoring the exit code seems like not the right solution. What if removing the user failed _and_ that is an important signal of a failure?

jordansissel:
I'm not sure what the best approach is, here. It's a tricky road dealing with user management. Should we provide separate packages that don't include preinstall and post-remove scripts?

briandef:
It would appear that the debian package maintainer guidelines are unclear on removing users when package is removed/purged (https://wiki.debian.org/AccountHandlingInMaintainerScripts). Leaving the user in place is an acceptable option.

Since there may be logstash-owned artifacts leftover from the installation, including log files, keeping the logstash user makes sense to me.

This solves the problem for all users, including those who auth with /etc/passwd and those who auth with ldap. And saves the project from having to manage multiple debian packages with different post-install scripts.

vincentbernat:
Most packages use `|| true` because it also handles the case where `deluser` or `delgroup` are not available anymore. However, users should only be removed on purge.

jordansissel:
@vincentbernat I"ve never once used 'purge' in my 6 years of using debian-derived systems :\

I'm still not sure what should be done. I believe there is good cause to want these two mutually exclusive things:
- user deletion failure should be an error
- user deletion should be optional

I don't think failing to delete the user should be ignored. I don't know how to make the user deletion optional, though. A new package without this after-remove script, maybe?

vincentbernat:
To not help you, many Debian Developers think that a system user should never be deleted because some files owned by this user may be left on the disk and become readable by another user which may reuse the previous UID. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=390457.

I would suggest to adopt one of the two (widely adopted) behaviours:

 1. Not deleting the system account.
 2. Deleting the system account on purge and ignore any error.

Other way would be "special" to logstash and unexpected. For example, deleting the system account on uninstall would leave log files owned by logstash with no logstash user available. The problem with making user deletion failure an error is that it is difficult to recover from the user point of view. Assume the user uninstalled the package, then noticed that there still is a logstash user, removing it manually, then trying to purge the package. Because the postrm script failed, the package is now in an inconsistent state and dpkg will try to execute the postrm script each time and the user will have to recreate a logstash user to get out of that (or modify `/var/lib/dpkg/info/logstash.postrm` manually).

And I just noticed that I mostly said the same thing than @briandef

jordansissel:
> For example, deleting the system account on uninstall would leave log files owned by logstash with no logstash user available

should delete the log files as well upon package removal?

jordansissel:
> Because the postrm script failed

Indeed. I had a fun experience myself with this with a debian-originated package for collectd that would fail in postrm if collectd wasn't running (like, if you stopped collectd and then tried to remove the package). Part of the problem is that all you get is /bin/sh and shell is a trainwreck of a reliable interface; another part is that there are categories of users with different expectations (leave the user, delete the user, etc).

vincentbernat:
The Debian policy tells to remove log files on purge (but not on removal).

The best you can do is do like most packages in Debian. This doesn't apply for user deletion on removal since we (Debian) didn't decide what is the best thing to do yet.

jordansissel:
> do like most packages in Debian

My experience has been that there exists a Debian policy handbook but many packages deviate or ignore it. 

Let's ignore what Debian policy says. Without quoting the Debian guide, what configurable behaviors would benefit users?

vincentbernat:
Many packages? Unless you only speak about third-party packages, any deviation from the Debian policy is considered a serious bug and makes the package unfit for release. Therefore, IMO, users are usually expecting the behavior described in the Debian policy since this is the behavior of the majority of packages.

As for the question, the behavior that would benefit users is the one that makes the packages fit with other packages. So either remove on purge an ignore errors or don't remove on purge. Debian didn't decide it yet because most users don't care which one it is. However, Debian has ignored errors on user removal since many years ago because users did care about that. We did enforce this behavior through a policy trick and this is detected archive-wide by automatic testing.

I'll vote for "don't remove". Less code to write. Less bugs.

jordansissel:
Is this still an issue?

PR #3362, opened 2015-06-02
Fix the uninstall of plugins

### LINKED ISSUE #3336: Uninstall plugin fails
If I install a plugin using a gem file, and after that I try to remove with the command `bin/plugin uninstall plugin-name` it fails, saying that plugin is not installed.

purbon:
Fixes #3336 by doing a bundler setup call that was not done beforehand, like this bundler, and subsequently ```gem:specification```, is aware of plugins installed when invoking the ```bin/plugin``` command.

For now the way to test this is by building a plugin as a gem, by then installing it and uninstall it, so ```bin/plugin list --verbose``` is not showing it anymore.

### DIFF
--- a/lib/pluginmanager/main.rb
+++ b/lib/pluginmanager/main.rb
@@ -34,6 +34,7 @@ end
 
 if $0 == __FILE__
   begin
+    LogStash::Bundler.setup!({:without => [:build, :development]})
     LogStash::PluginManager::Main.run("bin/plugin", ARGV)
   rescue LogStash::PluginManager::Error => e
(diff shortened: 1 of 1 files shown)


purbon:
@ph @jsvd can you review that please?

purbon:
@ph @suyograo can you guys review this please? I know is hard for now to test and mainly manually, will work on the testing side too, but a good review it would be very nice. thanks.

purbon:
please jenkins, test this.

jsvd:
LGTM

PR #3365, opened 2015-06-02
allow parameters in peeraddr monkey patch

### LINKED ISSUE #3364: On Windows, bugfix_jruby_2558.rb breaks plugins using peeraddr with a parameter
`wrong number of arguments calling 'peeraddr' (1 for 0)`

Following file is the cause:
https://github.com/elastic/logstash/blob/f97c01da0631566e0ec07c9a076286e22a8f43a9/lib/logstash/patches/bugfix_jruby_2558.rb

Seems to be a very old patch that re-implements `peeraddr`, albeit not allowing any parameter. `TCPSocket.peeraddr` and `UDPSocket.peeraddr` have allowed a parameter in MRI since at least 1.9.3 (http://ruby-doc.org/stdlib-1.9.3/libdoc/socket/rdoc/IPSocket.html#method-i-peeraddr), and in JRuby since 1.7.0 (https://github.com/jruby/jruby/commit/d835b3c2adf7f87fbbc8e4cfe8b952c65fd3c3a9). This parameter allows reverse lookup to be disabled.
MRI allows :numeric symbol as well as boolean value, but JRuby only supports the boolean value. 

However, using either when using Logstash breaks the plugin.

jsvd:
fixes #3364

### DIFF
--- a/lib/logstash/patches/bugfix_jruby_2558.rb
+++ b/lib/logstash/patches/bugfix_jruby_2558.rb
@@ -3,8 +3,8 @@ require "logstash/environment"
 if LogStash::Environment.windows? && LogStash::Environment.jruby?
   require "socket"
   module JRubyBug2558SocketPeerAddrBugFix
-    def peeraddr
-      orig_peeraddr.map do |v|
+    def peeraddr(*args)
+      orig_peeraddr(*args).map do |v|
         case v
         when String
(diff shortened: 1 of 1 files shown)


ph:
LGTM

driskell:
:+1:

PR #3366, opened 2015-06-02
Replacing mailing list address with discussion forum address.

ycombinator:
Please let me know if I should cherry-pick this into other branches as well.


colinsurprenant:
LGTM

Issue #3361, opened 2015-06-02
Deadlock in Logstash 1.5 with filter flusher


driskell:
Hello,

I've noticed on one my multiline worker nodes I'm getting very frequent deadlocks in the "|worker" thread and "pipeline" thread. Other nodes are perfectly fine, so seems one of those annoying race conditions with very specific triggers. Also nodes without multiline filter, and therefore with `-w 8` for multiple workers, are fine too.

Investigating the Java stack trace (https://gist.github.com/driskell/2173a91ac3de78e035e9) and trying to locate the equivalent ruby code, it looks like there is a possibility of a race condition in regards to the filter flusher, if there's only one worker. Probably would never happen in MRI Ruby because of green threads. But JRuby native threads I think it could.

If the following happens between the filterworker `pop`ing off a Flush event and processing it in the mutex...
1. A new event is pushed onto the queue to fill it
2. The flush interval triggers and blocks trying to add a flush event to the queue
... then the filter worker will block trying to obtain the lock. However, the lock is held waiting for the queue to free up, which will never happen because the worker reading from the queue is blocked.
**Deadlock**

If I find time I'll try a PR. I think the solution is to push only a single flush event to the queue? And not push another until that one is processed. So just needs a flag inside the Mutex, say, `@filter_flush_pending` - that way if the circumstances above occur, number 2 wouldn't block as it wouldn't try to add another flush event.

ph:
@driskell Thanks we changed a bit that part and that could explain the newer issues, @colinsurprenant can you look at this please?

driskell:
So far my patch is working - the one I've committed above is slightly tweaked from what I'm testing but only on syntax level to make it cleaner so could be a good baseline. I'll apply it directly shortly and see how it runs next couple days (it's been hanging nearly every day since upgraded but not had time to look until now.)

I can raise as PR earlier if you wish.

ph:
@driskell I would still raise a PR if you can, other people using the multiline filter are reporting similar issues. With your report we will be able to reproduce it.

driskell:
@ph it's worth noting this is not related directly to multiline filter. It's purely related to single threaded worker setups, which includes all users of multiline filter but also includes others.

driskell:
@ph The root cause is in Logstash pipeline code it seems

ph:
@jsvd and I were able to reproduce it, we are discussing the need to actually do a synchronize on that part of the code.

driskell:
[...]

PR #3378, opened 2015-06-04
Remove the synchronize on the `flush_filters_to_ouputs!`

ph:
We were synchronizing on the tick and the flush call,
in some situation it could set logstash in a deadlock when the queue was
blocked and the tick occurred. The flush call was already thread safe
since only one worker can start the flushing process.

Fix: https://github.com/elastic/logstash/issues/3361

### DIFF
--- a/lib/logstash/pipeline.rb
+++ b/lib/logstash/pipeline.rb
@@ -115,7 +115,7 @@ class LogStash::Pipeline
   end
 
   def shutdown_filters
-    @flusher_lock.synchronize { @flusher_thread.kill }
+    @flusher_thread.kill
     @input_to_filter.push(LogStash::SHUTDOWN)
   end
 
@@ -156,8 +156,7 @@ class LogStash::Pipeline
       Thread.new { filterworker }
     end
 
-    @flusher_lock = Mutex.new
-    @flusher_thread = Thread.new { Stud.interval(5) { @flusher_lock.synchronize { @input_to_filter.push(LogStash::FLUSH) } } }
+    @flusher_thread = Thread.new { Stud.interval(5) { @input_to_filter.push(LogStash::FLUSH) } }
   end
 
   def start_outputs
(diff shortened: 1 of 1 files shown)


jsvd:
LGTM

colinsurprenant:
LGTM
before merging: @jsvd @ph did you run tests to make sure it does fix the origin issue?

jsvd:
@colinsurprenant  yes I had already experimented with this patch and confirmed in the original scenario LS did no longer hang.

ph:
@colinsurprenant We also tested in the 1.4.2 and we have similar behavior of losing events.

colinsurprenant:
:shipit:

driskell:
Now testing this.

ph:
@driskell thanks!

driskell:
@ph So far so good and no issues. Looks good to me.

jsvd:
:fireworks:

ph:
@driskell Thank you for testing this and your debug effort really appreciated!

bechtoldt:
@ph Can you estimate when 1.5.1 will be released?

ph:
@bechtoldt really soon, can you test with this patch if it solves your issue?

bechtoldt:
@ph that file isn't even available in the RPM. the entire /opt/logstash/lib/logstash dir is missing in a fresh install.

ph:
@bechtoldt Sorry, It's because the layout between the master branch and the release version is a bit different because of our packaging. 

The file you are looking fore is actually this one:
`logstash-1.5.0/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0-java/lib/logstash/pipeline.rb`

bechtoldt:
@ph LS is for 16 hours stable. I think this bug was the root-cause of our problem. Thank you for fixing it!

bradvido:
This PR fixed a major lock-up issue we were having on Logstash 1.5. 
It's been stable with no lockups for 12+ hours now. Previously would lock up every 1-2 hours.

ph:
:heart:  Thanks for the report we will release an official fix!

andreausu:
This PR fixed the issue for me too, thanks guys! <3

PR #3379, opened 2015-06-04
allow forced termination through multiple SIGINT

jsvd:
Create the ability to force logstash to exit in a stalled situation:

```shell
% bin/logstash -e 'input { generator { count => 21 message => "1"} } filter { ruby { code => "sleep 1000" } } output { stdout { }}'
Logstash startup completed
    #  ... after 5 seconds ...
Shutdown still ocurring. Send another ^C to force termination..
Warning: Data loss may occur if shutdown is forced! {:level=>:warn}
^CSIGINT received. Terminating immediately.. {:level=>:fatal}
%                                                 
```

### DIFF
--- a/lib/logstash/agent.rb
+++ b/lib/logstash/agent.rb
@@ -113,8 +113,16 @@ class LogStash::Agent < Clamp::Command
 
     # Make SIGINT shutdown the pipeline.
     sigint_id = Stud::trap("INT") do
-      @logger.warn(I18n.t("logstash.agent.sigint"))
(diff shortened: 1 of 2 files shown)


ph:
We should probably change the message to mention that the loss of `in transit` event could occur?

ph:
Tested locally and code LGTM. Only need to improve the message with the data loss mention and rebase the commits?

jsvd:
updated accordingly :ok_hand:

ph:
LGTM !

PR #3386, opened 2015-06-05
Add support for multiline conditionals with the `else if` statements

ph:
This is a followup of the issues #2850 and #3281.

The following configuration:

```
    if [condition] {

    } else if [condition1]
      or [condition2] {
        ..
    }
```
Was compiled to ruby like this:
```
     elsif condition or condition2 # else if [condition1]
       or [condition2]
```

and making the intepreter fails.


ph:
@jordansissel I found this when debugging a failling job on jenkins can you review it?

jordansissel:
LGTM

PR #3391, opened 2015-06-08
Fix all plugins test check for non released plugins

purbon:
fix the case where a plugin is released and has a similar name to one still not released, fixing failured on the all plugins test.

For example: http://build-eu-00.elastic.co/job/logstash_integration_all_plugins/533/console as it has been there for long time now.

### DIFF
--- a/rakelib/default_plugins.rb
+++ b/rakelib/default_plugins.rb
@@ -152,7 +152,7 @@ module LogStash
 
     def self.is_released?(plugin)
       require 'gems'
-      !Gems.search(plugin).reject{ |h| h['name'] != plugin }.empty?
+      Gems.info(plugin) != "This rubygem could not be found."
     end
   end
 end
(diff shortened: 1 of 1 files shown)


purbon:
@jsvd @ph can you guys please take a look here, silly change, but enable again the all plugins test to be happy in jenkins. Thanks a lot for your time.

purbon:
please jenkins, test this.

rakelib/default_plugins.rb
```diff
@@ -152,7 +152,7 @@ def self.fetch_all_plugins
 
     def self.is_released?(plugin)
       require 'gems'
-      !Gems.search(plugin).empty?
+      !Gems.search(plugin).reject{ |h| h['name'] != plugin }.empty?
```
jsvd: The rubygems gem seems to have an exact search request called [info](https://github.com/rubygems/gems/blob/master/lib/gems/client.rb#L25-L28) so maybe we could use that instead of doing a query and double checking afterwards

purbon: good idea, I will use it. thanks @jsvd

jsvd:
LGTM

PR #3395, opened 2015-06-08
Deprecate the input plugin's base charset setting.

### LINKED ISSUE #3044: Provide guidance on the right way to set charset
We have deprecated the charset option (http://www.elastic.co/guide/en/logstash/current/plugins-inputs-tcp.html#plugins-inputs-tcp-charset), it will be nice to provide the now-recommended alternative along with the deprecation message, eg.

```
codec => plain {
      charset => "your_charset"
}
```

jordansissel:
Refer uses to the codecs instead.

Fixes #3044


suyograo:
great. LGTM

PR #3399, opened 2015-06-09
add the perfmon plugin to the skip list

purbon:
this plugin is released, by the original author, and as we created a repo to be in logstash-plugins before we merge this plugin in, we need to add it to the skip list so the all plugins test is not complaining.


purbon:
@jsvd can you take a fast look at that.

jsvd:
LGTM

PR #3401, opened 2015-06-09
add logstash-output-webhdfs as is released, but still not completely …

purbon:
…integrated with logstash to be run in the all plugins test


jsvd:
LGTM

PR #3405, opened 2015-06-09
[DOC] Update repositories.asciidoc

untergeek:
The wording was unclear.  This clarifies our intent and the verbiage.

### DIFF
--- a/docs/asciidoc/static/repositories.asciidoc
+++ b/docs/asciidoc/static/repositories.asciidoc
@@ -5,9 +5,9 @@ We also have repositories available for APT and YUM based distributions. Note
 that we only provide binary packages, but no source packages, as the packages
 are created as part of the Logstash build.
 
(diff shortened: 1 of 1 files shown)


docs/asciidoc/static/repositories.asciidoc
```diff
@@ -5,9 +5,9 @@ We also have repositories available for APT and YUM based distributions. Note
 that we only provide binary packages, but no source packages, as the packages
 are created as part of the Logstash build.
 
-We have split the major versions in separate urls to avoid accidental upgrades
-across major version. For all 1.5.x releases use 1.5 as version number, for
-1.4.x use 1.4, etc.
+We have split the Logstash package repositories into separate urls to avoid
```
jordansissel: I like this change, how about "We have split the Logstash package repositories by version into ..." (Basically, add "by version" after "repositories)

jordansissel:
LGTM

PR #3407, opened 2015-06-10
Show errors encountered when validating a plugin rather than saying it does not exist

driskell:
Currently, if you attempt to install a plugin and an error occurs, the plugin installer just reports the plugin does not exist. For example, if I turn off my internet connection:

```
$ bin/plugin install logstash-input-courier
Validating logstash-input-courier
Plugin logstash-input-courier does not exist
ERROR: Installation aborted, verification failed for logstash-input-courier 
```

This PR reports the errors if there are any, instead of reporting the plugin does not exist. For example, the above scenario would now report:

```
$ bin/plugin install logstash-input-courier
Validating logstash-input-courier
Unable to download data from https://rubygems.org - SocketError: initialize: name or service not known (https://rubygems.org/latest_specs.4.8.gz)
ERROR: Installation aborted, verification failed for logstash-input-courier 
```

### DIFF
--- a/lib/pluginmanager/util.rb
+++ b/lib/pluginmanager/util.rb
@@ -17,7 +17,10 @@ module LogStash::PluginManager
       end
     else
       dep = Gem::Dependency.new(plugin, version || Gem::Requirement.default)
-      specs, error = Gem::SpecFetcher.fetcher.spec_for_dependency(dep)
+      specs, errors = Gem::SpecFetcher.fetcher.spec_for_dependency(dep)
+
+      # dump errors
+      errors.each { |error| $stderr.puts(error.wordy) }
 
       # depending on version requirements, multiple specs can be returned in which case
       # we will grab the one with the highest version number
@@ -27,7 +30,7 @@ module LogStash::PluginManager
         end
         return valid
       else
-        $stderr.puts("Plugin #{plugin}" + (version ? " version #{version}" : "") + " does not exist")
+        $stderr.puts("Plugin #{plugin}" + (version ? " version #{version}" : "") + " does not exist") if errors.empty?
         return false
       end
     end


suyograo:
nice one @driskell

lib/pluginmanager/util.rb
```diff
@@ -19,6 +19,11 @@ def self.logstash_plugin?(plugin, version = nil)
       dep = Gem::Dependency.new(plugin, version || Gem::Requirement.default)
       specs, error = Gem::SpecFetcher.fetcher.spec_for_dependency(dep)
 
+      # dump errors
+      error.each do |e|
+        $stderr.puts(e.wordy)
+      end
```
ph: Switch to plural form,`errors` instead of `error`?
Since it is a one liner we can use an inline block `{ |e|  .... }`

lib/pluginmanager/util.rb
```diff
@@ -27,7 +32,7 @@ def self.logstash_plugin?(plugin, version = nil)
         end
         return valid
       else
-        $stderr.puts("Plugin #{plugin}" + (version ? " version #{version}" : "") + " does not exist")
+        $stderr.puts("Plugin #{plugin}" + (version ? " version #{version}" : "") + " does not exist") unless error.length != 0
```
ph: Can you replace the conditionals with `unless errors.empty?`

ph:
Thank you @driskell ! Only minor comments and we ready to merge it :)

lib/pluginmanager/util.rb
driskell: Replaced with `if errors.empty?` (to only display the plugin does not exist message if no errors were reported)

lib/pluginmanager/util.rb
driskell: Yep!

driskell:
I've modified the commit - all done! Thanks.

PR #3420, opened 2015-06-11
Update warning message to include explanation

### LINKED ISSUE #3406: Warning about inflight messages when trying to stop Logstash
When you send a ^C signal to Logstash and if it is processing in-flight events, it will not stop immediately. It will first stop the input and wait for existing events in the internal buffer to completely drain before quitting. We should provide feedback to the user that it is waiting for events to drain, so it is clear that LS is not stuck, but this is designed behavior. We should also suggest a force kill option introduced in https://github.com/elastic/logstash/pull/3379, but warn that users will lose data.

suyograo:
Fixes #3406

### DIFF
--- a/locales/en.yml
+++ b/locales/en.yml
@@ -67,8 +67,9 @@ en:
       sigterm: >-
         SIGTERM received. Shutting down the pipeline.
       slow_shutdown: |-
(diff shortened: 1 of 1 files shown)


ph:
LGTM

Issue #3383, opened 2015-06-04
Add warning to bin/plugin update when updating to a new major version of plugin


talevy:
Soon, there are going to be some backwards-incompatible changes bubbling up in plugins, and it would 
be helpful to users to understand that there is a risk in updating to such versions.

maybe it'll look something like this:

```
$ bin/plugin update logstash-output-elasticsearch
You are updating logstash-output-elasticsearch to a new version 5.0, which may not be compatible with 
4.2. are you sure you want to proceed (Y/N)?
```

purbon:
This makes complete sense to me, specially if we follow with the plugins a more or less semver policy, then this warning is informative for users updating.

suyograo:
Closed by https://github.com/elastic/logstash/pull/3423

PR #3423, opened 2015-06-11
Validation & warning when updating plugins between major versions

purbon:
This PR adds a validation phase and a warning if you're updating plugins between major versions.

Fixes: #3383

### DIFF
--- a/lib/pluginmanager/update.rb
+++ b/lib/pluginmanager/update.rb
@@ -42,12 +42,13 @@ class LogStash::PluginManager::Update < LogStash::PluginManager::Command
     filtered_plugins = plugins.map { |plugin| gemfile.find(plugin) }
       .compact
       .reject { |plugin| REJECTED_OPTIONS.any? { |key| plugin.options.has_key?(key) } }
-      .each { |plugin| gemfile.update(plugin.name) }
+      .select { |plugin| validate_major_version(plugin.name) }
+      .each   { |plugin| gemfile.update(plugin.name) }
 
     # force a disk sync before running bundler
     gemfile.save
 
-    puts("Updating #{filtered_plugins.collect(&:name).join(", ")}")
+    puts("Updating #{filtered_plugins.collect(&:name).join(", ")}") unless filtered_plugins.empty?
 
     # any errors will be logged to $stderr by invoke!
     # Bundler cannot update and clean gems in one operation so we have to call the CLI twice.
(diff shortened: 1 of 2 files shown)


purbon:
@suyograo @talevy can one of you kindly review this? Thanks a lot for your time guys!

talevy:
```
(pr/3423)$ bin/plugin update logstash-input-mysql                                                                                 
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: 
You are updating logstash-input-mysql to a new version 1.0.0, which may not be compatible with 0.1.1. are you sure you want to proceed (Y/N)?
y
Updating logstash-input-mysql
Updated logstash-input-mysql 0.1.1 to 1.0.0
```

purbon:
@talevy you see this message because you're testing with master, if you redo this test with 1.5 you're not going to see it. This is because it check for the ```path``` option in the Gemfile and in master there is always one.

lib/pluginmanager/update.rb
```diff
@@ -62,6 +63,20 @@ def update_gems!
     display_bundler_output(output)
   end
 
+  # validate if there is any major version update so then we can ask the user if he is
+  # sure to update or not.
+  def validate_major_version(plugin)
+    require "gems"
+    latest_version  = Gems.versions(plugin)[0]['number'].split(".")
```
talevy: can we add some exception handling here for when a plugin is not found in rubygems?

It doesn't make sense to, but some may run this on a local gem that is not on rubygems.

purbon: This makes complete sense

purbon: @talevy as I understand the code base nowadays, all plugins installed as local gems are skip, so this function will only get executed for plugins that are present in rubygems because they got installed throw them.

In detail, this https://github.com/elastic/logstash/blob/master/lib/pluginmanager/update.rb#L44 line reject plugins that are installed with https://github.com/elastic/logstash/blob/master/lib/pluginmanager/update.rb#L6, so all local fields. This is why we see this message:

```
╰─$ bin/plugin update logstash-input-log4j
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: logstash-input-log4j
No plugin updated
```
if we try to update with a local gem installed. And for now there is no option to update a gem installed from a rubygems with a gem locally as all calls goes throw our beloved Bundler.

Makes sense to you @talevy ?

talevy: makes sense, yeah. sounds good to me

talevy:
LGTM

PR #3426, opened 2015-06-11
add a runtime license test

purbon:
In the spirit of what Kibana https://github.com/elastic/kibana/blob/dc07a456cd146ec6a3deaf412520c66523602920/tasks/config/licenses.js has to check for dependency licenses I created this small test. 

With this we can test core, but also plugin dependencies in regular basis throw the CI environment.

### DIFF
--- a/spec/license_spec.rb
+++ b/spec/license_spec.rb
@@ -1,0 +1,47 @@
+require 'spec_helper'
+require 'rakelib/default_plugins'
+
+describe "Plugin licenses" do
+
+  let(:common_licenses) {
+    Regexp.union([ /mit/,
+                   /apache*/,
+                   /bsd/,
+                   /ruby/])
+  }
+
+  shared_examples "a license test" do
+
+    subject(:gem_name) do |example|
+      example.metadata[:example_group][:parent_example_group][:description]
+    end
+
+    let(:spec) { Gem::Specification.find_all_by_name(gem_name)[0] }
+
+    it "should have an expected licenses" do
+      spec.licenses.each do |license|
+        expect(license.downcase).to match(common_licenses)
+      end
+    end
+
+    it "has runtime dependencies with expected licenses" do
+      spec.runtime_dependencies.map { |dep| dep.to_spec }.each do |runtime_spec|
+        next unless runtime_spec
+        runtime_spec.licenses.each do |license|
+          expect(license.downcase).to match(common_licenses)
+        end
+      end
+    end
+  end
+
+  describe "logstash-core" do
+    it_behaves_like "a license test"
+  end
+
+  installed_plugins.each do |default_plugin|
+    describe default_plugin do
+      it_behaves_like "a license test"
+    end
+  end
+
+end
--- a/spec/license_spec.rb
+++ b/spec/license_spec.rb
@@ -1,16 +1,16 @@
 require 'spec_helper'
 require 'rakelib/default_plugins'
 
-describe "Plugin licenses" do
+describe "Project licenses" do
 
-  let(:common_licenses) {
+  let(:expected_licenses) {
     Regexp.union([ /mit/,
                    /apache*/,
                    /bsd/,
                    /ruby/])
   }
 
-  shared_examples "a license test" do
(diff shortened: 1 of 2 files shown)


purbon:
please jenkins, test this.

purbon:
please jenkins, test this.

suyograo:
@jsvd review?

jsvd:
It would be nice to have this on plugins too, as a task that could be invoked before publish, for example.
This way when you're developing a plugin you have feedback about the licenses you're setting/depending on.

spec/license_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+                   /ruby/])
+  }
+
+  shared_examples "a license test" do
+
+    subject(:gem_name) do |example|
+      example.metadata[:example_group][:parent_example_group][:description]
+    end
+
+    let(:spec) { Gem::Specification.find_all_by_name(gem_name)[0] }
```
jsvd: do `find_by_name` instead of `find_all_by_name[0]`

spec/license_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        next unless runtime_spec
+        runtime_spec.licenses.each do |license|
+          expect(license.downcase).to match(common_licenses)
+        end
+      end
+    end
+  end
+
+  describe "logstash-core" do
+    it_behaves_like "a license test"
```
jsvd: maybe `it_behaves_like "a correctly licensed gem"`?

spec/license_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  shared_examples "a license test" do
+
+    subject(:gem_name) do |example|
+      example.metadata[:example_group][:parent_example_group][:description]
+    end
+
+    let(:spec) { Gem::Specification.find_all_by_name(gem_name)[0] }
+
+    it "should have an expected licenses" do
```
jsvd: maybe `it "should only contain expected licenses"`

spec/license_spec.rb
purbon: ```find_by_name``` does not work for the case of logstash-core.gemspec.

Test:
```
jruby-1.7.19 :003 > Gem::Specification.find_by_name("logstash-codec-json")
 => #<Gem::Specification:0x447a logstash-codec-json-0.1.7> 
jruby-1.7.19 :004 > Gem::Specification.find_by_name("logstash-core")
 => nil 
jruby-1.7.19 :005 > Gem::Specification.find_all_by_name("logstash-core")
 => [#<Gem::Specification:0x447c logstash-core-2.0.0.dev-java>] 
```

So if I want to check for the main project I should use this ugly notation.

purbon:
@jsvd I agree it would be a nice to have for plugins, this might be a good stuff for another PR :+1:

purbon:
@jsvd applied your comments, thanks for your time. Could you re evaluate this PR ? Thanks!

spec/license_spec.rb
jsvd: wow that sucks, does it not match because of -java?

spec/license_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'spec_helper'
+require 'rakelib/default_plugins'
+
+describe "Project licenses" do
+
+  let(:expected_licenses) {
```
jsvd: We should document here (or link to a place) the licenses we're expecting to match with these regular expressions, I'm guessing:

Apache License 2.0
BSD License
MIT license
..?

jsvd:
minor documentation comment, this is a great thing to have :+1:

spec/license_spec.rb
purbon: I think it fails because it uses two different methods in ruby, one is actively looking for specs ```matching_specs``` and the other is just doing ```to_spec```, but I did not more research.

purbon:
@jsvd added the requested comment, can you re evaluate this PR and stamp it if good for you?

jsvd:
LGTM

purbon:
Thank you for your time @jsvd

PR #3434, opened 2015-06-11
Make the changelog markdown format

suyograo:
On a long plane ride from Brazil I sat down to convert this into markdown. It was surprisingly straightforward. I didn't want to move this file to CHANGELOG.md because we have blogs linking to this file all over the place.


ph:
Gormatting LGTM,

I think what we should here is to rename this file to `CHANGELOG.md` and create a `CHANGELOG` with a mention that we moved the to markdown and put a link to. Because a real .md will be rendered on github and will allow us to link to specific section (like this https://github.com/elastic/logstash#logstash-plugins). 

What do you think @jordansissel ?

untergeek:
I agree with the link idea.  We want it rendered.

suyograo:
@ph @untergeek makes sense. Added an actual changelog.md file

ph:
LGTM

PR #3452, opened 2015-06-16
Warn on java version

### LINKED ISSUE #2547: Reject (or warn strongly) if running on a known-buggy JVM
See https://github.com/elasticsearch/logstash/issues/2537 for context.

There are known bugs in the JVM/Java that we'd like to avoid by not using known buggy versions.

Should we abort execution on these JVMs? Just warn strongly? Add a flag that permits running, like `--i-know-my-java-is-broken` ?

andrewvc:
Fixes #2547 . @jordansissel is STDERR.puts the best place to put this? I wasn't sure if we should force a message like this to stdout, or if it should use regular logging mechanisms. Sometimes regular logging is ignored by users.

### DIFF
--- a/lib/logstash/runner.rb
+++ b/lib/logstash/runner.rb
@@ -16,6 +16,7 @@ class LogStash::Runner
 
   def main(args)
     require "logstash/util"
+    require "logstash/util/java_version"
     require "stud/trap"
     require "stud/task"
     @startup_interruption_trap = Stud::trap("INT") { puts "Interrupted"; exit 0 }
@@ -26,6 +27,9 @@ class LogStash::Runner
       $stderr.puts "Ruby 1.9.2 or later is required. (You are running: " + RUBY_VERSION + ")"
       return 1
     end
+
+    # Print a warning to STDERR for bad java versions
+    LogStash::Util::JavaVersion.warn_on_bad_java_version
 
     Stud::untrap("INT", @startup_interruption_trap)
 
--- a/lib/logstash/util/java_version.rb
+++ b/lib/logstash/util/java_version.rb
@@ -1,8 +1,16 @@
+require 'cabin'
+
 module LogStash::Util::JavaVersion
+  def self.logger
+    @logger ||= Cabin::Channel.get(LogStash)
+  end
+
   # Print a warning if we're on a bad version of java
   def self.warn_on_bad_java_version
     if self.bad_java_version?(self.version)
-      STDERR.puts("!!! Please upgrade your java version, the current version '#{self.version}' may cause problems. We recommend a minimum version of 1.7.0_51")
+      msg = "!!! Please upgrade your java version, the current version '#{self.version}' may cause problems. We recommend a minimum version of 1.7.0_51"
+      STDERR.puts(msg)
+      logger.warn(msg)
     end
   end
 
--- a/lib/logstash/util/java_version.rb
+++ b/lib/logstash/util/java_version.rb
@@ -26,7 +26,7 @@ module LogStash::Util::JavaVersion
     match = version_string.match(/\A(\d+)\.(\d+)\.(\d+)(_(\d+))?(-(.+))?\Z/)
     major, minor, patch, ufull, update, bfull, build = match.captures
 
-    return {
+    {
       :full => version_string,
       :major => major.to_i,
       :minor => minor.to_i,
@@ -42,11 +42,11 @@ module LogStash::Util::JavaVersion
     parsed = parse_java_version(version_string)
 
     if parsed[:major] == 1 && parsed[:minor] == 7 && parsed[:patch] == 0 && parsed[:update] < 51
-      return true
+      true
     elsif parsed[:major] == 1 && parsed[:minor] < 7
-      return true
+      true
     else
-      return false
+      false
     end
   end
 end
--- a/lib/logstash/util/java_version.rb
+++ b/lib/logstash/util/java_version.rb
@@ -2,7 +2,7 @@ module LogStash::Util::JavaVersion
   # Print a warning if we're on a bad version of java
   def self.warn_on_bad_java_version
     if self.bad_java_version?(self.version)
-      STDERR.puts("Please upgrade your java version, the current version '#{self.version}' may cause problems. We recommend a minimum version of 1.7.0_51")
+      STDERR.puts("!!! Please upgrade your java version, the current version '#{self.version}' may cause problems. We recommend a minimum version of 1.7.0_51")
     end
   end
 
@@ -40,7 +40,7 @@ module LogStash::Util::JavaVersion
     return nil if version_string.nil?
 
     parsed = parse_java_version(version_string)
-
+    
     if parsed[:major] >= 1 && parsed[:minor] == 7 && parsed[:patch] == 0 && parsed[:update] < 51
       return true
     elsif parsed[:major] >= 1 && parsed[:minor] < 7
--- a/lib/logstash/util/java_version.rb
+++ b/lib/logstash/util/java_version.rb
@@ -9,12 +9,12 @@ module LogStash::Util::JavaVersion
   # Check to see if this is a recommended java version, print a warning to stdout if this is a bad version
   # Returns the current java version
   def self.version
-    return nil if RUBY_ENGINE != "jruby"
+    return nil unless LogStash::Environment.jruby?
 
     require 'java'
     java_import "java.lang.System"
 
-    System.getProperties["java.runtime.version"]
+    System.getProperty("java.runtime.version")
   end
 
   def self.parse_java_version(version_string)
@@ -45,6 +45,8 @@ module LogStash::Util::JavaVersion
       return true
     elsif parsed[:major] == 1 && parsed[:minor] < 7
       return true
+    else
+      return false
     end
   end
 end
--- a/lib/logstash/util/java_version.rb
+++ b/lib/logstash/util/java_version.rb
@@ -17,6 +17,9 @@ module LogStash::Util::JavaVersion
     System.getProperty("java.runtime.version")
   end
 
+  # Takes a string of a java version ex: "1.8.0_24-beta"
+  # and returns a parsed map of the components.
+  # nil inputs will be returned as nil.
   def self.parse_java_version(version_string)
     return nil if version_string.nil?
 
@@ -36,6 +39,9 @@ module LogStash::Util::JavaVersion
     }
   end
 
+  # Determine if the given java version string is a bad version of java
+  # If it is, return true, if it isn't return false.
+  # Accepts nil, returning nil.
   def self.bad_java_version?(version_string)
     return nil if version_string.nil?
 
--- a/lib/logstash/util/java_version.rb
+++ b/lib/logstash/util/java_version.rb
@@ -14,15 +14,10 @@ module LogStash::Util::JavaVersion
     end
   end
 
-  # Check to see if this is a recommended java version, print a warning to stdout if this is a bad version
-  # Returns the current java version
+  # Return the current java version string. Returns nil if this is a non-java platform (e.g. MRI).
   def self.version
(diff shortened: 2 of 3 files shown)


jordansissel:
@andrewvc hmm, this PR only has changes for Gemfile and Gemfile.jruby-19.lock. I don't see any code changes. Something went funny in git?

Gemfile
```diff
@@ -8,9 +8,17 @@ gem "ci_reporter_rspec", "1.0.0", :group => :development
...
 gem "coveralls", :group => :development
 gem "rspec", "~> 3.1.0", :group => :development
-gem "logstash-devutils", "~> 0", :group => :development
+gem "logstash-devutils", ">= 0"
 gem "octokit", "3.8.0", :group => :build
 gem "stud", "~> 0.0.19", :group => :build
 gem "fpm", "~> 1.3.3", :group => :build
 gem "rubyzip", "~> 1.1.7", :group => :build
 gem "gems", "~> 0.8.3", :group => :build
+gem "logstash-filter-clone"
```
purbon: this should not be here, you committed the generated Gemfile (after you run rake test:core) by mistake.

lib/logstash/util/java_version.rb
```diff
@@ -0,0 +1,50 @@
...
+  def self.warn_on_bad_java_version
+    if self.bad_java_version?(self.version)
+      STDERR.puts("!!! Please upgrade your java version, the current version '#{self.version}' may cause problems. We recommend a minimum version of 1.7.0_51")
+    end
+  end
+
+  # Check to see if this is a recommended java version, print a warning to stdout if this is a bad version
+  # Returns the current java version
+  def self.version
+    return nil if RUBY_ENGINE != "jruby"
```
purbon: Here you can use ``Environment.jruby?```

andrewvc: ah, nice.

lib/logstash/util/java_version.rb
```diff
@@ -0,0 +1,50 @@
...
+  def self.bad_java_version?(version_string)
+    return nil if version_string.nil?
+
+    parsed = parse_java_version(version_string)
+    
+    if parsed[:major] >= 1 && parsed[:minor] == 7 && parsed[:patch] == 0 && parsed[:update] < 51
+      return true
+    elsif parsed[:major] >= 1 && parsed[:minor] < 7
+      return true
+    end
```
purbon: what does this return if this two conditions are not met? false? nil? 

on the other side, I see expression possible to be simplified as one.

spec/util/java_version_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'spec_helper'
+require 'logstash/util/java_version'
+
+describe "LogStash::Util::JavaVersion" do
+  let(:mod) { LogStash::Util::JavaVersion }
```
purbon: you should use subject here as is the test guy of this example.

lib/logstash/util/java_version.rb
andrewvc: Yes, an if block will return nil if no conditions are met. I'll make it explicit however.

I chose to use two if branches because I thought it was a little more clear than a big boolean.

spec/util/java_version_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  it "should get the current java version if we're on Java" do
+    if RUBY_ENGINE == "jruby"
```
purbon: you are testing java versions, so it makes no sense to just check here, you should check before running the test.

spec/util/java_version_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      expect(LogStash::Util::JavaVersion.version).to be_a(String)
+    end
+  end
+
+  it "should mark a bad java version as bad" do
+    expect(mod.bad_java_version?("1.7.0_45-beta")).to be_truthy
```
purbon: Here, you have at the end two test cases, one with the beta and another without. This should be two it section, so when debugging we get a nice error message.

spec/util/java_version_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    expect(mod.bad_java_version?("1.8.0-beta")).to be_falsey
+  end
+
+  describe "parsing java versions" do
+    it "should return nil on a nil version" do
+      expect(mod.parse_java_version(nil)).to be_nil
+    end
+
+    it "should parse a plain version" do
+      parsed = mod.parse_java_version("1.3.0")
```
purbon: this should become a subject as noted beforehand.

spec/util/java_version_spec.rb
andrewvc: Fixed!

andrewvc:
@purbon I dryed up the tests. Let me know what you think

lib/logstash/util/java_version.rb
```diff
@@ -0,0 +1,52 @@
+module LogStash::Util::JavaVersion
+  # Print a warning if we're on a bad version of java
+  def self.warn_on_bad_java_version
+    if self.bad_java_version?(self.version)
+      STDERR.puts("!!! Please upgrade your java version, the current version '#{self.version}' may cause problems. We recommend a minimum version of 1.7.0_51")
```
purbon: I would use the logger in here, you can see how to use it for example in the agent file.

spec/util/java_version_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+          expect(parsed[:update]).to eql(update)
+        end
+
+        it "should have the correct build string" do
+          expect(parsed[:build]).to eql(build)
+        end
+      end
+    end
+
+    include_examples("version parsing", "a plain version", "1.3.0", 1, 3, 0, 0, nil)
```
purbon: good job man!

spec/util/java_version_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  subject(:mod) { LogStash::Util::JavaVersion }
+
+  it "should get the current java version if we're on Java" do
+    if LogStash::Environment.jruby?
+      expect(LogStash::Util::JavaVersion.version).to be_a(String)
+    end
+  end
+
+  it "should mark a bad beta version as bad" do
+    expect(mod.bad_java_version?("1.7.0_45-beta")).to be_truthy
```
purbon: What happen if I run the test without being java? you should skip the test if not under jruby?, do something like: https://github.com/elastic/logstash/blob/master/spec/util/json_spec.rb#L33

lib/logstash/util/java_version.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  # Check to see if this is a recommended java version, print a warning to stdout if this is a bad version
+  # Returns the current java version
+  def self.version
+    return nil unless LogStash::Environment.jruby?
```
purbon: you might return an exception here, what does nil means? not parsed properly? is not very clear.

lib/logstash/util/java_version.rb
```diff
@@ -0,0 +1,52 @@
...
+  def self.parse_java_version(version_string)
+    return nil if version_string.nil?
+
+    # Crazy java versioning rules @ http://www.oracle.com/technetwork/java/javase/versioning-naming-139433.html
+    # The regex below parses this all correctly http://rubular.com/r/sInQc3Nc7f
+
+    match = version_string.match(/\A(\d+)\.(\d+)\.(\d+)(_(\d+))?(-(.+))?\Z/)
+    major, minor, patch, ufull, update, bfull, build = match.captures
+
+    return {
```
purbon: this return is not necessary as by default ruby will return the last thing :-)

lib/logstash/util/java_version.rb
```diff
@@ -0,0 +1,52 @@
...
+      :major => major.to_i,
+      :minor => minor.to_i,
+      :patch => patch.to_i,
+      :update => update.to_i, # this is always coerced to an int (a nil will be zero) to make comparisons easier
+      :build => build # not an integer, could be b06 for instance!,
+    }
+  end
+
+  def self.bad_java_version?(version_string)
+    return nil if version_string.nil?
```
purbon: Similar thing here, what does nil means for this method?

purbon:
@andrewvc I left you some more comments, this is starting to looks good! good job on the test man!

lib/logstash/util/java_version.rb
andrewvc: I read in the style guide that explicit returns were preferred. I'm 100% with you on implicit return personally. Is the style guide out of date?

lib/logstash/util/java_version.rb
andrewvc: Nil means its not even a version string (its a nil). This would let someone perhaps chain nils if they so choose (with something like #try in rails).

lib/logstash/util/java_version.rb
andrewvc: nil means we are not running on java. I don't see a need to raise an exception here.

purbon: Please document the nil meaning, so we know :-)

spec/util/java_version_spec.rb
andrewvc: This function is not platform dependent. It just checks java version strings. If you want to check a java version string on MRI I think that's not an issue.

lib/logstash/util/java_version.rb
andrewvc: I think this could go either way on this. void is a concept from languages that aren't ruby, and are used to denote functions with side effects. In ruby a nil means nil, which is a much more expansive concept than void. Raising an exception here is a stylistic choice in my opinion, but I don't see a strong case for it. If getting a nil java version on MRI makes sense to me, because in that case the java version non-existent AKA nil.

lib/logstash/util/java_version.rb
suyograo: can we do both maybe? logger and stderr so we don't miss it?

lib/logstash/util/java_version.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+  end
+
+  # Check to see if this is a recommended java version, print a warning to stdout if this is a bad version
+  # Returns the current java version
+  def self.version
+    return nil unless LogStash::Environment.jruby?
+
+    require 'java'
+    java_import "java.lang.System"
```
jordansissel: java_import isn't necessary, you can just do `java.lang.System.getProperty("java.runtime.version")` directly w/o java_import. 

I think `require "java"` is also not needed as some time ago JRuby requires that by default, I think? At least on JRuby 1.7.19 it seems to:

```
% rvm use 1.7.19
Using /Users/jls/.rvm/gems/jruby-1.7.19
% ruby -e 'puts require "java"'
false
```
(false means it was already required)

lib/logstash/util/java_version.rb
jordansissel: I'm on the fence about this also. Leaning towards wanting an Exception raised, but this is such a small part of the code, just documenting 'nil means we aren't running on Java' might be OK, especially if we do an early return from `warn_on_bad_java_version` if we aren't even running under Java/JRuby. Thoughts?

lib/logstash/runner.rb
```diff
@@ -27,6 +28,9 @@ def main(args)
       return 1
     end
 
+    # Print a warning to STDERR for bad java versions
+    LogStash::Util::JavaVersion.warn_on_bad_java_version
```
jsvd: why not change this into
`LogStash::Util::JavaVersion.warn_bad_java_version if bad_java_version?`
Then this predicate can return false on MRI

jsvd: nevermind my comment. ^

andrewvc:
@jordansissel I've cleaned up all issues. Let me know if we're good to go. Thanks for all the reviews :)

jordansissel:
LGTM

PR #3457, opened 2015-06-17
Add a micro benchmark runner

ph:
Add `rake benchmark:run` task to run micro benchmark on specific part of
the code, it's uses the benchmark-ips gem to generate them.
See https://github.com/evanphx/benchmark-ips for usage

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem "simplecov", :group => :development
 gem "coveralls", :group => :development
 gem "rspec", "~> 3.1.0", :group => :development
 gem "logstash-devutils", "~> 0", :group => :development
+gem "benchmark-ips", :group => :development
 gem "octokit", "3.8.0", :group => :build
 gem "stud", "~> 0.0.19", :group => :build
 gem "fpm", "~> 1.3.3", :group => :build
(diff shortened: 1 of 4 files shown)


andrewvc:
@ph Tested this with the benchmark file below

```ruby
require 'benchmark/ips'

Benchmark.ips do |x|
  x.config(:time => 5, :warmup => 2)

  # These parameters can also be configured this way
  x.time = 5
  x.warmup = 2

  # Typical mode, runs the block as many times as it can
  x.report("addition") { 1 + 2 }
end
```

It worked.

LGTM

purbon:
I like this very much, good idea!!

jordansissel:
Proposal + code LGTM. @andrewvc [tested](https://github.com/elastic/logstash/pull/3457#issuecomment-113655232) it and all looks well.

LGTM :)

ph:
small typo in the commit message i'll update it.

andrewvc:
@ph it might be good to include the example test I posted, just so that people have an idea of how to use this functionality. ¯\_(ツ)_/¯

ph:
yes ^ I'll add my other tests too :)

ph:
I've cleaned the PR a bit, removed the `.gitkeep` added an example test run and fixed the dependencies in the `Gemfile` / `Gemfile*.lock`

Issue #3453, opened 2015-06-17
Add NOTICE.TXT to Logstash


suyograo:
We need to add NOTICE.TXT to Logstash distributables.  We should document copyrights of jars used from Apache Software Foundation. 
We also need to audit third party compiled code used in Logstash and add information to comply with their Licenses

purbon:
@suyograo I think we can part automate this in the same spirit of https://github.com/elastic/logstash/pull/3426

PR #3458, opened 2015-06-17
Add NOTICE.TXT to comply with License

suyograo:
Also changed packaging to include CHANGELOG.md now that we switched over to markdown format.

Ref: #3453

### DIFF
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -5,7 +5,8 @@ namespace "artifact" do
   def package_files
     [
       "LICENSE",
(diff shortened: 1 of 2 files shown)


rakelib/artifacts.rake
```diff
@@ -5,7 +5,8 @@ namespace "artifact" do
   def package_files
     [
       "LICENSE",
-      "CHANGELOG",
+      "CHANGELOG.md",
```
ph: good catch ^

ph:
LGTM

PR #3463, opened 2015-06-18
Better java event inspect

### LINKED ISSUE #3308: Calling #inspect on an Event instance doesn't return the content of an ArrayList.
Calling #inspect on a event which contains an ArrayList will return the instance id instead of returning a string of the actual content.

``
"instances"=>#<Java::JavaUtil::ArrayList:0x205196d9>
``

andrewvc:
Fixes #3308 

@ph can you check this out when you get a chance?

### DIFF
--- a/spec/core/event_spec.rb
+++ b/spec/core/event_spec.rb
@@ -2,7 +2,6 @@
 require "spec_helper"
 
 describe LogStash::Event do
-
   shared_examples "all event tests" do
     context "[]=" do
       it "should raise an exception if you attempt to set @timestamp to a value type other than a Time object" do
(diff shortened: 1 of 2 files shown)


ph:
@andrewvc I think this isn't the right PR for the `inspect` change? :)

andrewvc:
@ph apologies. I just pushed up the right branch.

spec/core/event_spec.rb
```diff
@@ -331,6 +330,17 @@
...
+    # This is a custom method on Event
+    context "inspect" do
+      it "should include the json metadata" do
+        expect(subject.inspect).to include(subject.to_json_with_metadata)
+      end
+
+      it "should include the object id" do
+        expect(subject.inspect).to include(subject.object_id.to_s)
+      end
+    end
```
ph: Should include the `class` ?

andrewvc:
@ph fixed the tests up. Also made it output the object hash as hex rather than object_id (like #inspect normally does)

ph:
I've been thinking this morning and I've made a mistake concerning implementing the fix only inside the `Event` class, I am really sorry. In reality we have this problem everywhere as soon we use the `LogStash::Json.load` to unserialize json stuff and we use it all over the place because it's faster than the stdlib json.

```
irb(main):004:0> LogStash::Json.load(json)
=> {"awe"=>#<Java::JavaUtil::ArrayList:0x7479ca10>}
irb(main):005:0> Event.new(LogStash::Json.load(json))
```

iirc, This problem come from using the `Raw` methods of jrjackson, it is faster since it doesn't convert the structure and we work with native java structure. 

```
# from logstash/json
options[:symbolize_keys] ? JrJackson::Raw.parse_sym(data) : JrJackson::Raw.parse_raw(data)
```

So we could add support for the `ArrayList` in `logstash/java_integration` as a patch?

andrewvc:
@ph I'm killing this PR and moving to this one: https://github.com/elastic/logstash/pull/3474

PR #3465, opened 2015-06-18
Fix the coverage analysis throw simplecov to take care of all files

purbon:
Our actual coverage analysis report were not accurate mostly because of files not being really analyses by simplecov.

While this PR does not solve the source issue, it workaround the situation proving us back with reports in good shape.

This PR also add detailed groups and filters to the coverage configuration so the reports gets more clear, this configuration is logstash-core specific so I added it here.

--- Description of the error source ---

the problem actually life in doing the simplecov require before or after the ```LogStash::Bundler.setup!({:without => [:build]})```, still not clear who might be the offender, but doing the simplecov setup before this call overcome the error for now.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,3 +1,8 @@
+require_relative 'coverage_helper'
+# In order to archive an expected coverage analysis we need to eager load
+# all logstash code base, otherwise it will not get a good analysis.
+CoverageHelper.eager_load if ENV['COVERAGE']
+
 require "logstash/devutils/rspec/spec_helper"
 
 def installed_plugins
--- a/lib/logstash/environment.rb
+++ b/lib/logstash/environment.rb
@@ -77,7 +77,7 @@ module LogStash
     end
 
     def windows?
-      Gem.win_platform?
+      ::Gem.win_platform?
     end
 
     def vendor_path(path)
--- a/rakelib/test.rake
+++ b/rakelib/test.rake
@@ -4,7 +4,13 @@
 require "pluginmanager/util"
 
 namespace "test" do
+
   task "setup" do
+    # Need to be run here as because if run aftewarse (after the bundler.setup task) then the report got wrong
+    # numbers and misses files. There is an issue with our setup! method as this does not happen with the regular
+    # bundler.setup used in regular bundler flows.
+    Rake::Task["test:setup-simplecov"].invoke if ENV['COVERAGE']
+
     require "bootstrap/environment"
     LogStash::Bundler.setup!({:without => [:build]})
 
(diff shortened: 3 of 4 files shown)


purbon:
@jordansissel @suyograo this fix the coverage reports by simplecov, looks like the way we use bundler nowaday made then not to play along together.

purbon:
please jenkins, test this.

purbon:
@suyograo @jordansissel can you guys review this so we can merge and gets good coverage reports in regular basis? thanks!

spec/spec_helper.rb
```diff
@@ -1 +1,23 @@
+if ENV['COVERAGE']
+  # In order to archive an expected coverage analysis we need to eager load
+  # all logstash code base, otherwise it will not get a good analysis.
+
+  require "pluginmanager/main"
+  Dir.glob("lib/logstash/patches/*") do |file|
```
jsvd: for consistency sake would be nice to use *.rb on the three Dir.glob operations

spec/spec_helper.rb
```diff
@@ -1 +1,23 @@
...
+    require file
+  end
+
+  Dir.glob("lib/logstash/*.rb") do |file|
+    require file
+  end
+
+  Dir.glob("lib/logstash/util/*") do |file|
+    next if ["lib/logstash/util/prctl.rb"].include?(file)
```
jsvd: is this file still relevant? maybe we can remove it.

purbon: Actually I don't know, it looks like being a file for MRI and FFI. @jordansissel @ph do you know if this is still necessary? 

```
    if UNAME == "linux"
      require "logstash/util/prctl"
      # prctl PR_SET_NAME allows up to 16 bytes for a process name
      # since MRI 1.9, JRuby, and Rubinius use system threads for this.
      LibC.prctl(PR_SET_NAME, name[0..16], 0, 0, 0)
    end
``` 

^^ is the place where is used in ```lib/logstash/util.rb```.

spec/spec_helper.rb
```diff
@@ -1 +1,23 @@
+if ENV['COVERAGE']
```
jsvd: could all these requires be wrapped into a method and called in setup-simplecov before `require "simplecov"`?

purbon: They can be wrapped in a method, but I the eager loading should be done here as is part of the specs code base, moving this to the rake task feels awkward to me. In the long run we should be able to have an option somewhere that just load all code on demand without this ugly hack.

jsvd:
LGTM

spec/spec_helper.rb
colinsurprenant: this should not be MRI specific but Linux specific so AFAIK it is still relevant.

purbon: ok, makes sense, thanks for reporting @colinsurprenant

spec/spec_helper.rb
```diff
@@ -1 +1,23 @@
...
+    require file
+  end
+
+  Dir.glob("lib/logstash/*.rb") do |file|
+    require file
+  end
+
+  Dir.glob("lib/logstash/util/*.rb") do |file|
+    next if ["lib/logstash/util/prctl.rb"].include?(file)
+    require file
```
colinsurprenant: why are we using an array here to exclude this specific file? why not just test for that file directly?

purbon: because i tend to write like this in case we've more than one file in the feature, I can change this for now and make it just an string comparison.

spec/spec_helper.rb
colinsurprenant: not sure I understand. these requires are specifically needed for simplecov, why not do them in the setup-simplecov task?

purbon: This load are not there "just for simplecov", they are there because if where are not loading them, then we can not have a complete coverage analysis. This is an option to have in the specs configuration, this is not an option to have in a rake task. 

I agree, this should be wrapped in a helper that basically load everything, but still an method to be call in here. At less for my understanding.

spec/spec_helper.rb
colinsurprenant: or better yet, put it in a constant named something like "EXCLUDE_LIST" or something ? then its obvious what this array is etc

spec/spec_helper.rb
colinsurprenant: if its not just for simpleconv then why is it wrapped in a `if ENV['COVERAGE']` ?

purbon: because is for coverage analysis, no matter what tool you use.

spec/spec_helper.rb
jordansissel: Logstash calls prctl to set the thread name in the OS so things like top -H
will show that name instead of "Java" - it helps with debugging and such.

spec/spec_helper.rb
colinsurprenant: ok, so it's not related to actual testing/specs but for coverage analysis so I would suggest to create a specific task for eager loading stuff and have the setup-simplecov task depend on that other eager loading task.

spec/spec_helper.rb
colinsurprenant: ok, so it's not related to actual testing/specs but for coverage analysis so I would suggest to create a specific task for eager loading stuff and have the setup-simplecov task depend on that other eager loading task.

spec/spec_helper.rb
purbon: makes sense, I will wrap this is a more decent method + exclude_list before merging this in. Thanks for your time reviewing @colinsurprenant

PR #3467, opened 2015-06-18
Removing reference of `fieldreference`

ph:
This module was replaced by `lib/logstash/util/accessor.rb`


talevy:
LGTM

colinsurprenant:
should we try to transpose the old fieldeval_spec.rb tests in the accessors specs?

ph:
I'll check if we are not missing some tests in the accessors.

jordansissel:
LGTM + @colinsurprenant's [comment](https://github.com/elastic/logstash/pull/3467#issuecomment-113983172)

ph:
I've checked, we cover the test case in the new accessors class.

Issue #3468, opened 2015-06-18
License report


mrsolo:
See 

https://github.com/elastic/elasticsearch/issues/10684

> We should go through all the third party dependencies that we redistribute and make sure 
> NOTICE.TXT has all the info needed to comply with the license terms of the dependencies.

Similar information is required for logstash project

suyograo:
@mrsolo can i close this for https://github.com/elastic/logstash/issues/3453?

mrsolo:
@suyograo yes please

PR #3469, opened 2015-06-18
Generate gem depdendency information

mrsolo:
Add rake test 'license' to generate yaml output on current gem depedencies

Part of

https://github.com/elastic/logstash/issues/3468

### DIFF
--- a/rakelib/license.rake
+++ b/rakelib/license.rake
@@ -13,30 +13,4 @@ namespace "license" do
 
 end
 
-# def all_installed_gems
-#   Gem::Specification.all = nil
-#   all = Gem::Specification
-#   Gem::Specification.reset
-#   all
-# end
-
-# def generate_license_information
-#   licenses = []
-#   all_installed_gems.select {|y| y.gem_dir.include?('vendor') }.each do |x|
-#     if(x.license) #ah gem has license information
-#       licenses.push ({:name => x.name, :version => x.version.to_s, :license => x.license, :homepage => x.homepage, :email => x.email})
-#     else
-#       license = {:name => x.name, :version => x.version.to_s, :homepage => x.homepage, :email => x.email}
-#       license_file =  Dir.glob(File.join(x.gem_dir,'LICENSE*')).first #see if there is a license file
-#       if(license_file)
-#         license[:license] = File.read(license_file)
-#       else
-#         license = license.merge({:license=> 'unknown', :gem_dir => x.gem_dir, :gem_path => x.files.join("\n")})
-#       end
-#       licenses.push(license.sort{|u, v| v.name <=> u.name })
-#     end
-#   end
-#   puts YAML.dump(licenses)
-# end
-
 task "license" => [ "license:core" ]
--- a/rakelib/license.rake
+++ b/rakelib/license.rake
@@ -1,0 +1,45 @@
+# we need to call exit explicity  in order to set the proper exit code, otherwise
+# most common CI systems can not know whats up with this tests.
+require_relative "default_plugins"
+
+namespace "license" do
+
+  desc "run core specs"
+  #task "core" =>  ["test:setup"] do
+  task "core" do
+    require 'yaml'
+    generate_license_information
+  end
+
+end
+
+def all_installed_gems
+  Gem::Specification.all = nil
+  all = Gem::Specification
+  Gem::Specification.reset
+  all
+end
+
+def generate_license_information
+  licenses = []
+  all_installed_gems.select {|y| y.gem_dir.include?('vendor') }.sort {|v, u| v.name  <=> u.name }.each do |x|
+    if(x.license) #ah gem has license information
+      licenses.push ({:name => x.name, :version => x.version.to_s, :license => x.license, :homepage => x.homepage, :email => x.email})
+    else
+      license = {:name => x.name, :version => x.version.to_s, :homepage => x.homepage, :email => x.email}
+      license_file =  Dir.glob(File.join(x.gem_dir,'LICENSE*')).first #see if there is a license file
+      if(license_file)
+        file = File.open(license_file, 'r')
+        data = file.read
+        file.close
+        license[:license] = data
+      else
+        license = license.merge({:license=> 'unknown', :gem_dir => x.gem_dir, :gem_path => x.files.join("\n")})
+      end
+      licenses.push(license)
+    end
+  end
+  puts YAML.dump(licenses)
+end
+
+task "license" => [ "license:core" ]
--- a/rakelib/license.rake
+++ b/rakelib/license.rake
@@ -1,11 +1,11 @@
 # we need to call exit explicity  in order to set the proper exit code, otherwise
 # most common CI systems can not know whats up with this tests.
 require_relative "default_plugins"
+require_relative "license"
 
 namespace "license" do
 
   desc "run core specs"
-  #task "core" =>  ["test:setup"] do
   task "core" do
     require 'yaml'
     generate_license_information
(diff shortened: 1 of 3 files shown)


purbon:
@mrsolo I was wondering, if you aim to get just the license name or also the description of the license?

purbon:
@mrsolo have you seen what https://github.com/elastic/logstash/blob/master/spec/license_spec.rb#L26 does ? can we integrate this two related task somehow?

mrsolo:
If they can be.  This script and others will be used to produce a spreadsheet report.   However if
https://github.com/elastic/logstash/issues/3453 is done.  The information can be based on that.

The main interested on that spreadsheet is check the license type and compliance on third party libraries that are distributed with logstash distribution.

purbon:
@mrsolo boths things are compatible, one for us to check throw CI and the other to compile the notice.txt file. I was more wondering if we can merge some code here. I will drop you here some small comments on the code, and then I think we can merge and start using it somehow, makes sense?

rakelib/license.rake
```diff
@@ -0,0 +1,45 @@
...
+  desc "run core specs"
+  #task "core" =>  ["test:setup"] do
+  task "core" do
+    require 'yaml'
+    generate_license_information
+  end
+
+end
+
+def all_installed_gems
```
purbon: I would rather move this code to a ```.rb``` file the same way that ```default_plugins.rb``` is used.

rakelib/license.rake
```diff
@@ -0,0 +1,45 @@
+# we need to call exit explicity  in order to set the proper exit code, otherwise
+# most common CI systems can not know whats up with this tests.
+require_relative "default_plugins"
+
+namespace "license" do
+
+  desc "run core specs"
+  #task "core" =>  ["test:setup"] do
```
purbon: This might be a left over from debugging?

rakelib/license.rake
```diff
@@ -0,0 +1,45 @@
...
+def generate_license_information
+  licenses = []
+  all_installed_gems.select {|y| y.gem_dir.include?('vendor') }.sort {|v, u| v.name  <=> u.name }.each do |x|
+    if(x.license) #ah gem has license information
+      licenses.push ({:name => x.name, :version => x.version.to_s, :license => x.license, :homepage => x.homepage, :email => x.email})
+    else
+      license = {:name => x.name, :version => x.version.to_s, :homepage => x.homepage, :email => x.email}
+      license_file =  Dir.glob(File.join(x.gem_dir,'LICENSE*')).first #see if there is a license file
+      if(license_file)
+        file = File.open(license_file, 'r')
```
purbon: To do this you can use the online ```File.read(file_name)``` as an simpler way to handle this.

rakelib/license.rake
```diff
@@ -0,0 +1,45 @@
...
+def all_installed_gems
+  Gem::Specification.all = nil
+  all = Gem::Specification
+  Gem::Specification.reset
+  all
+end
+
+def generate_license_information
+  licenses = []
+  all_installed_gems.select {|y| y.gem_dir.include?('vendor') }.sort {|v, u| v.name  <=> u.name }.each do |x|
```
purbon: This line looks complicated to understand,  I guess braking it or arranging it in a function would make it more easy to follow, what do you think?

rakelib/license.rake
```diff
@@ -0,0 +1,45 @@
...
+  #task "core" =>  ["test:setup"] do
+  task "core" do
+    require 'yaml'
+    generate_license_information
+  end
+
+end
+
+def all_installed_gems
+  Gem::Specification.all = nil
```
purbon: I'm wondering why do you need to do that? Why not just using grasping all runtime dependencies the same way as https://github.com/elastic/logstash/blob/master/spec/license_spec.rb#L33 is doing? can you educate me of the differences?

rakelib/license.rake
```diff
@@ -0,0 +1,45 @@
...
+  all = Gem::Specification
+  Gem::Specification.reset
+  all
+end
+
+def generate_license_information
+  licenses = []
+  all_installed_gems.select {|y| y.gem_dir.include?('vendor') }.sort {|v, u| v.name  <=> u.name }.each do |x|
+    if(x.license) #ah gem has license information
+      licenses.push ({:name => x.name, :version => x.version.to_s, :license => x.license, :homepage => x.homepage, :email => x.email})
```
purbon: I would rather make this a separate function, for example. Like this we can DRY this data points and simplifier changes in the feature.

rakelib/license.rake
mrsolo: Either way is fine with me.  I though rake only pick up *.rake

rakelib/license.rake
mrsolo: Right will remove

rakelib/license.rake
mrsolo: Ditto

rakelib/license.rake
mrsolo: Pretty clear to me. but I wrote it.  Putting into function or breaking into multiline is fine

rakelib/license.rake
mrsolo: I wasn't aware of  the alternatives, will check it out

rakelib/license.rake
mrsolo: Hmm?  I dunno.  The statement is pretty clear. The expected change is pretty small and there is only one consumer.

I am in 'not make it a function' unless one of the requirement is met camp.

rakelib/license.rake
jordansissel: The `.sort` can be written shorter as `.sort_by(:name)` also

jordansissel: Does it need to be sorted?

rakelib/license.rake
jordansissel: I'd rather see some kind of gem_license_info(x) that returns the hash, then:
`licenses.push(gem_license_info(x))` or something similar. 

There's some code duplication in these sections (license hash is hand-coded multiple times `{:name => ...` for each use case. Maybe we should find the license information first, then separately add it to the hash?

```ruby
license_info = { :name => ... }
if x.license
  license_info[:license] = x.license
else
  license_file = Dir.glob(....)...
  license_info[:license] = File.read(license_file)
...
end
```

mrsolo: ditto

rakelib/license.rake
mrsolo: The sort can be delayed till the actual output.  It helps on constructing and auditing the spreadsheet.

rakelib/license.rake
mrsolo: Hmm https://github.com/elastic/logstash/blob/master/spec/license_spec.rb#L24 needs names.  For this analysis, that is essentially whatever will be bundled in the distributions.  Where can one grab that information?

suyograo:
@purbon is this ready to merge?

purbon:
@suyograo @mrsolo reviewing again, sorry it took longer.

rakelib/license.rb
```diff
@@ -0,0 +1,40 @@
...
+require_relative "default_plugins"
+require "rubygems/specification"
+require "bootstrap/environment"
+
+
+def all_installed_gems
+  ENV["GEM_HOME"] = ENV["GEM_PATH"] = LogStash::Environment.logstash_gem_home
+  Gem.use_paths(LogStash::Environment.logstash_gem_home)
+
+  Gem::Specification.all = nil
```
purbon: I don't understand the logic behind this set of method calls. Can you explain them to me? specially using the ```.all``` method recommends in the documentation:

```
Returns all specifications. This method is discouraged from use. You probably want to use one of the Enumerable methods instead.
```

I think is totally safe and more easy to use one of the many enumerated methods available. 

What do you think?

purbon:
@mrsolo why do you point this PR to the 1.5 branch, please point it to master and then we can merge it there and backport it. This would make everything more easy when merging. Thanks

rakelib/license.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+  license
+end
+
+def generate_license_information
+  licenses = []
+  all_installed_gems.select {|y| y.gem_dir.include?('vendor') }.each do |x|
+    licenses.push(gem_license_info(x))
+  end
+  puts YAML.dump(licenses.sort{|u, v| u[:name] <=> v[:name] })
```
purbon: couple of very cosmetic change:

* I would move the puts to the line 10 of the ```license.rake``` task, print to STDOUT should be a call of the user of this method.

* I tend to like more using ```to_yaml```, in some rubies has been playing better to me than dump, but this is more a matter of taste to me.

- purbon

purbon:
@suyograo @mrsolo I left a couple of small improvements, with this being addressed I'm happy this goes in.

rakelib/license.rb
mrsolo: .all is used to get all of the dependencies that is accessible in the current rubygem environments. 

I will look up on enumerable to see what it does exactly

mrsolo:
> why do you point this PR to the 1.5 branch, please point it to master and then we can merge it there
 > and backport it. This would make everything more easy when merging. Thanks

My task is very specific, Ggenerate logstash 1.5 third party license report.  I know this does break how the usually development flow of pr to main and backport.  But my primary reason for this PR is to park my one off script somewhere instead of possibility of losing it and have more people providing input on how to retrieve such information.

After running the license analysis for logstash 1.5, I realized that analysis at the current form is not sufficient.  What is required is to do analysis against the distributed file.  That is given a logstash.tgz file, figuring out all the packaged third parties and licenses.

rakelib/license.rb
purbon: I think a nice and probably better way here to do it would  be by using ```all_names()```  and then iterating over the returned enumeration and asking for details with ```find_by_name(name, *req)```.

purbon:
@mrsolo i see your point, but why not having this to all the current versions? then we can simply execute it in 1.5 or in master, or somewhere else. 

If you want to get all dependency information I recommend you actually install all plugins, and then use ```find_by_name```, using all the names returned by ```all_names()``` to retrieve details for all dependencies. Then you can see if a dependency is runtime/development and fetch their license. Makes sense? basically in the same spirit of the https://github.com/elastic/logstash/blob/master/spec/license_spec.rb file.

Makes sense to you?

mrsolo:
The point of moving to master is well taken.

> If you want to get all dependency information I recommend you actually install all plugins, and then use > find_by_name, using all the names returned by all_names() to retrieve details for all dependencies. 
Right if the api calls yield the same result.  However no plugin should be installed by this activities..other then what's being offered as part of rake.  Even that is flawed...

purbon:
@mrsolo I think I need some clarification? you say:

>  However no plugin should be installed by this activities..other then what's being offered as part of rake.

In the case of LS, how do you plan to get the dependencies if the plugins are not installed? Nowadays there is no other option.

> Right if the api calls yield the same result.

Yes the api I told you provide the same answer, but in a more fashionable way. :+1:

PR #3484, opened 2015-06-22
add a periodic report of inflight events during shutdown

jsvd:
Logstash can be prevented from exiting due to inflight messages that are residing between inputs and outputs.
Once pipeline.shutdown is called there will be a periodic report every 5 seconds showing which queues still have events. For example, starting logstash with the following config:

```
input { generator { count => 1000 } }
output {
  tcp { host => localhost port => 3333 workers => 2 }
  stdout {}
}
```
And then typing Ctrl+C, will report each 5 seconds:

```
*** BEGIN INFLIGHT EVENTS REPORT ****
input_to_filter => 20 events
filter_to_output => 20 events
LogStash::Outputs::Tcp: {"host"=>"localhost", "port"=>3333, "workers"=>2, "type"=>"", "tags"=>[], "exclude_tags"=>[], "codec"=><LogStash::Codecs::JSON charset=>"UTF-8">, "reconnect_interval"=>10, "mode"=>"client"} => 20 events
*** END INFLIGHT EVENTS REPORT **** {:level=>:warn, :file=>"logstash/pipeline.rb", :line=>"338", :method=>"report_inflight_events"} 
```

### DIFF
--- a/lib/logstash/pipeline.rb
+++ b/lib/logstash/pipeline.rb
@@ -8,6 +8,7 @@ require "logstash/config/file"
 require "logstash/filters/base"
 require "logstash/inputs/base"
 require "logstash/outputs/base"
+require "logstash/util/reporter"
 
 class LogStash::Pipeline
 
@@ -252,7 +253,8 @@ class LogStash::Pipeline
   #
   # This method is intended to be called from another thread
   def shutdown
-    Thread.new { report_inflight_events(@input_to_filter, @filter_to_output, @outputs) }
+    InflightEventsReporter.logger = @logger
+    InflightEventsReporter.start(@input_to_filter, @filter_to_output, @outputs)
     @input_threads.each do |thread|
       # Interrupt all inputs
       @logger.info("Sending shutdown signal to input thread", :thread => thread)
@@ -321,22 +323,5 @@ class LogStash::Pipeline
       end
     end
   end # flush_filters_to_output!
-
-  private
-  def report_inflight_events(input_to_filter, filter_to_output, outputs)
-    loop do
-      sleep 5
-      report = []
-      report << "*** BEGIN INFLIGHT EVENTS REPORT ****"
-      report << "input_to_filter => #{input_to_filter.size} events" if input_to_filter.size > 0
-      report << "filter_to_output => #{filter_to_output.size} events" if filter_to_output.size > 0
-      outputs.each do |output|
-        next unless output.worker_queue
-        report << "#{output} => #{output.worker_queue.size} events" if output.worker_queue.size > 0
-      end
-      report << "*** END INFLIGHT EVENTS REPORT ****"
-      @logger.warn report.join("\n")
-    end
-  end
 
 end # class Pipeline
--- a/lib/logstash/pipeline.rb
+++ b/lib/logstash/pipeline.rb
@@ -252,6 +252,7 @@ class LogStash::Pipeline
   #
   # This method is intended to be called from another thread
   def shutdown
+    Thread.new { report_inflight_events(@input_to_filter, @filter_to_output, @outputs) }
     @input_threads.each do |thread|
       # Interrupt all inputs
       @logger.info("Sending shutdown signal to input thread", :thread => thread)
@@ -320,5 +321,22 @@ class LogStash::Pipeline
       end
     end
   end # flush_filters_to_output!
+
+  private
+  def report_inflight_events(input_to_filter, filter_to_output, outputs)
+    loop do
+      sleep 5
+      report = []
+      report << "*** BEGIN INFLIGHT EVENTS REPORT ****"
+      report << "input_to_filter => #{input_to_filter.size} events" if input_to_filter.size > 0
+      report << "filter_to_output => #{filter_to_output.size} events" if filter_to_output.size > 0
+      outputs.each do |output|
+        next unless output.worker_queue
+        report << "#{output} => #{output.worker_queue.size} events" if output.worker_queue.size > 0
+      end
+      report << "*** END INFLIGHT EVENTS REPORT ****"
+      @logger.warn report.join("\n")
+    end
+  end
 
 end # class Pipeline
--- a/lib/logstash/util/reporter.rb
+++ b/lib/logstash/util/reporter.rb
@@ -1,0 +1,27 @@
+class InflightEventsReporter
+  def self.logger=(logger)
+    @logger = logger
+  end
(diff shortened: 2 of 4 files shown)


lib/logstash/pipeline.rb
```diff
@@ -252,6 +252,7 @@ def outputworker
   #
   # This method is intended to be called from another thread
   def shutdown
+    Thread.new { report_inflight_events(@input_to_filter, @filter_to_output, @outputs) }
```
jordansissel: Can we encapsulate this into a separate method somehow? Maybe `report_inflight_events_async(....)` or instead of Thread.new, use Concurrent::Future? I dunno, feels weird to create a thread like this.

lib/logstash/pipeline.rb
```diff
@@ -321,4 +322,21 @@ def flush_filters_to_output!(options = {})
...
+      report = []
+      report << "*** BEGIN INFLIGHT EVENTS REPORT ****"
+      report << "input_to_filter => #{input_to_filter.size} events" if input_to_filter.size > 0
+      report << "filter_to_output => #{filter_to_output.size} events" if filter_to_output.size > 0
+      outputs.each do |output|
+        next unless output.worker_queue
+        report << "#{output} => #{output.worker_queue.size} events" if output.worker_queue.size > 0
+      end
+      report << "*** END INFLIGHT EVENTS REPORT ****"
+      @logger.warn report.join("\n")
```
jordansissel: This may make the report hard to read if users are logging to a file since `Object#inspect` (and in the future, json) is used to write to teh logstash logfile.

lib/logstash/pipeline.rb
jordansissel: Or something like

```ruby
InflightEventsReporter.new(@input_to_filter, ....).start
```

ph: I would vote to encapsulate that logic into his own class.
I suggest something like this, so this factory method would take care of the initializing the class and managing the thread.

```
InflightEventsReporter.watch(@input_to_filter, ....)
```

ph: or Maybe event more generic.

```ruby
SizeQueueReporter.new(:timer => 5)
SizeQueueReporter.watch(:output, @output)
SizeQueueReporter.watch(:input, @input)
SizeQueueReporter.start()
```

jsvd: I see, I like to be very upfront/explicit about creating threads, but I'll think of a better way to design this.

jordansissel: @jsvd +1 on being explicit. Maybe we compromise and name the reporter "SomethingSomethingTask" to indicate it's a separate task/thread? ShutdownReporterTask, or something?

ph: +1 naming ^

jsvd:
UPDATE:
Changed the format to a proper ruby object (array), this way the logger can serialize that at will, to plain text or json, for example. Example of slow filters + failing tcp output healing during shutdown

config:

```
input { generator { count => 30 message => "1"} }
filter {
  ruby { code => "sleep 2" }
}
output {
  tcp { host => "localhost" workers => 2 port => 3333 }
}
```

```
data loss. {:level=>:warn}
["INFLIGHT_EVENTS_REPORT", "2015-07-07T10:20:07+01:00", {"input_to_filter"=>19, "filter_to_output"=>0, "outputs"=>[["<LogStash::Outputs::Tcp host=>\"localhost\", workers=>2, port=>3333, codec=><LogStash::Codecs::JSON charset=>\"UTF-8\">, reconnect_interval=>10, mode=>\"client\">", 12]]}] {:level=>:warn}
["INFLIGHT_EVENTS_REPORT", "2015-07-07T10:20:12+01:00", {"input_to_filter"=>16, "filter_to_output"=>0, "outputs"=>[["<LogStash::Outputs::Tcp host=>\"localhost\", workers=>2, port=>3333, codec=><LogStash::Codecs::JSON charset=>\"UTF-8\">, reconnect_interval=>10, mode=>\"client\">", 15]]}] {:level=>:warn}
["INFLIGHT_EVENTS_REPORT", "2015-07-07T10:20:17+01:00", {"input_to_filter"=>14, "filter_to_output"=>0, "outputs"=>[["<LogStash::Outputs::Tcp host=>\"localhost\", workers=>2, port=>3333, codec=><LogStash::Codecs::JSON charset=>\"UTF-8\">, reconnect_interval=>10, mode=>\"client\">", 17]]}] {:level=>:warn}
["INFLIGHT_EVENTS_REPORT", "2015-07-07T10:20:22+01:00", {"input_to_filter"=>10, "filter_to_output"=>0, "outputs"=>[["<LogStash::Outputs::Tcp host=>\"localhost\", workers=>2, port=>3333, codec=><LogStash::Codecs::JSON charset=>\"UTF-8\">, reconnect_interval=>10, mode=>\"client\">", 20]]}] {:level=>:warn}

 ###  REMOVED BACKPRESSURE BY STARTING "nc -l 3333" ####

["INFLIGHT_EVENTS_REPORT", "2015-07-07T10:20:27+01:00", {"input_to_filter"=>7, "filter_to_output"=>0, "outputs"=>[]}] {:level=>:warn}
["INFLIGHT_EVENTS_REPORT", "2015-07-07T10:20:32+01:00", {"input_to_filter"=>2, "filter_to_output"=>0, "outputs"=>[]}] {:level=>:warn}
Logstash shutdown completed
```

suyograo:
@jordansissel are you good with this?

lib/logstash/util/reporter.rb
```diff
@@ -0,0 +1,27 @@
...
+    report = {
+      "input_to_filter" => input_to_filter.size,
+      "filter_to_output" => filter_to_output.size,
+      "outputs" => []
+    }
+    outputs.each do |output|
+      next unless output.worker_queue && output.worker_queue.size > 0
+      report["outputs"] << [output.inspect, output.worker_queue.size]
+    end
+    @logger.warn ["INFLIGHT_EVENTS_REPORT", Time.now.iso8601, report]
```
jordansissel: I'm not sold on this use of logging an array, especially since logger includes timestamp by default. We can merge as-is and clean it up later.

jordansissel:
Code LGTM; output LGTM for now - we can clean up later. The important part (notification of in-flight events during shutdown) is solved.

suyograo:
Merged to 1.5 manually because of conflicts in .lock file

PR #3485, opened 2015-06-22
Add a ruby style guide plus a matching rubocop file

### LINKED ISSUE #3462: Work on a common coding style guide
As big as the logstash code base is getting, and specially with having many contributors, lots of different coding styles will encounter. In order to simplify the codebase manageability and in order to simplify reviews we should work out a common coding style.

This style guide should help reviews and also be enforced throw tools like rubocop.

/cc @suyograo @jordansissel

purbon:
This PR is a working progress PR with an initial version of the style guide, only roughly edited the first two bullet points, and a matching rubocop file.

Feel free to dump here your particular points of views about the style guide, as soon as we've an initial agreement over the style guide, we'll create matching rubocop checks for that.

Fixes #3462


.rubocop.yml
```diff
@@ -0,0 +1,13 @@
+Style/Encoding:
+  Enabled: false
```
jordansissel: For all cops we change, let's have a comment in the rubocop.yml that explains the reasoning.

.rubocop.yml
```diff
@@ -0,0 +1,13 @@
+Style/Encoding:
```
jordansissel: Why disable this?

.rubocop.yml
```diff
@@ -0,0 +1,13 @@
...
+  Enabled: false
+
+AllCops:
+  Include:
+    - 'lib/**/*'
+    - 'rakelib/**/*'
+  Exclude:
+    - 'vendor/**/*'
+    - 'spec/**/*'
```
jordansissel: Why exclude spec?

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+* [Exceptions](#exceptions)
+* [Collections](#collections)
+* [Strings](#strings)
+* [Regular Expressions](#regular-expressions)
+* [Percent Literals](#percent-literals)
+* [Metaprogramming](#metaprogramming)
+
+## Source Code Layout
+
+  * Use `UTF-8` as the source file encoding. If you're using a ruby version older than 2.0, adding ```# encoding: utf-8``` on top of a ruby file will tell the compiler to use this encoding by default. If using a ruby VM older than 2.0, `UTF-8` is the default encoding.
```
jordansissel: > If using a ruby VM older than 2.0, `UTF-8` is the default encoding

I think "older" should be "newer". In Ruby 2.0+, the default encoding is UTF-8. Prior in 1.8 and 1.9, I think the default encoding was assumed to be based on $LANG but overridable with the `# encoding:xyz` comment at the top of the file.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  ```Ruby
+  # bad
+  class FooError < StandardError
+  end
+
+  # okish
+  class FooError < StandardError; end
+
+  # good
+  FooError = Class.new(StandardError)
```
jordansissel: We may want to explicitly discourage using Class.new (and Module.new). See https://bugs.ruby-lang.org/issues/11119

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  # good
+  e = M * c**2
+  ```
+
+  `{` and `}` deserve a bit of clarification, since they are used for block and hash literals, as well as string interpolation. For hash literals two styles are considered acceptable.
+
+  ```Ruby
+  # good - space after { and before }
+  { one: 1, two: 2 }
```
jordansissel: It's notable that because of my own style practices, Logstash uses hashrocket `{ key => value }` everywhere, not the newer Ruby 1.9+ hash-colon `{ key: value }` syntax.

We _can_ move to the new hash-colon syntax (though it has some caveats such as requiring symbols for all keys) if we determine it valuable, but if we don't, we shouldn't use hash-colon syntax in the style guide.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  # good - no space after { and before }
+  {one: 1, two: 2}
```
jordansissel: We should choose one style for hash spacing, either spaces after { and before }, or no spaces.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    # good
+    1..3
+    'a'...'z'
+    ```
+
+* Indent `when` as deep as `case`. I know that many would disagree with this one, but it's the style established in both "The Ruby Programming Language" and "Programming Ruby".
+
+  ```Ruby
+  # bad
+  case
```
jordansissel: This 'bad' example is actually my preferred way to read and write case statements in ruby. Can someone somehow convince me to do the "good" way instead?

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  result = if some_cond
+    calc_something
+  else
+    calc_something_else
+  end
+
+  # good - it's apparent what's going on
+  kind = case year
+         when 1850..1889 then 'Blues'
```
jordansissel: I don't really like this because it's _so_ much whitespace. My alternative is something like this:

```ruby
kind = case year
  when ...
  when ...
end
```

Indent all the whens by 2 spaces.

CODE_STYLE.md
andrewvc: Good call @jordansissel . Plus,  I like the consistency of `class FooError < StandardError; end`. Doesn't take any special knowledge to understand.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+         when 1890..1909 then 'Ragtime'
+         when 1910..1929 then 'New Orleans Jazz'
+         when 1930..1939 then 'Swing'
+         when 1940..1950 then 'Bebop'
+         else 'Jazz'
+         end
+
+  result = if some_cond
```
jordansissel: same here, so much whitespace. I'm open to being convinced that this is good, but like the wildly-indented-whens above, it gives me some kind of visual dissonance that feels weird when reading and makes it harder to structure if we _also_ limit ourselves to N-character-wide line length.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  # okish - notice that the first ; is required
+  def no_braces_method; body end
+
+  # okish - notice that the second ; is optional
+  def no_braces_method; body; end
+
+  # okish - valid syntax, but no ; makes it kind of hard to read
+  def some_method() body end
+
```
andrewvc: These are the devil! I wouldn't mind getting rid of the `okish` ones. The only time I'd consider this OK would be is if there was some long boilerplate needed for some reason with 40-50 method calls in there (maybe for a shim or something?).

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+         end
+
+  result = if some_cond
+             calc_something
+           else
+             calc_something_else
+           end
+
+  # good (and a bit more width efficient)
+  kind =
```
jordansissel: This is a good compromise for spacing and readability, imo. +1 this one only. ;P

CODE_STYLE.md
andrewvc: I personally find the colon syntax to be more clear. Additionally, it highlights where high-cardinality key names not intended for GC are more clearly since the `""` really sticks out.

CODE_STYLE.md
jordansissel: On the other hand...

If we have such a need as:

```ruby
thing = 
  case ...
  ... many lines
   end
```

then it's likely we should move the case into a separate method:

```
def music_style(year)
  case year
    when 1850..1889 then 'Blues'
    ...
  end
end

... and later ...

kind = music_style(year)
```

CODE_STYLE.md
andrewvc: I think the logic here is that you really can't nest anything inside of case itself other than `when|else`. I don't really care about this portion of the code being consistent, but I write it the second (recommended) way.

CODE_STYLE.md
andrewvc: I prefer the style as given in this guide for the simple reason that RubyMine does it that way already by default.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  result = 1 \
+           - 2
+
+  long_string = 'First part of the long string' \
+                ' and second part of the long string'
+  ```
+
+* Adopt a consistent multi-line method chaining style. There are two popular styles in the Ruby community, both of which are considered good - leading `.` (Option A) and trailing `.` (Option B).
+
+  * **(Option A)** When continuing a chained method invocation on
```
jordansissel: We should only have 1 option documented as our style, here. In this case, my preference is always to have an operator lead a continued-line because we read left-to-right and the start of the continued-line begins with an operator so we can assume it starts from the previous line. This is `Option A` in this example.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    Mailer.deliver(
+        to: 'bob@example.com',
+        from: 'us@example.com',
+        subject: 'Important message',
+        body: source.text)
+  end
+
+  # good
+  def send_mail(source)
+    Mailer.deliver(to: 'bob@example.com',
```
jordansissel: So much whitespace. My preference is to instead do this:

```ruby
args = {
  to: ...,
  from: ...,
  subject: ...,
  ...
}
Mailer.deliver(args)
```

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+                   from: 'us@example.com',
+                   subject: 'Important message',
+                   body: source.text)
+  end
+
+  # good (normal indent)
+  def send_mail(source)
```
jordansissel: This "normal" indent is also OK with me, but it spreads a method call across so many lines that it may hurt readability.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+* Align the elements of array literals spanning multiple lines.
+
+  ```Ruby
+  # bad - single indent
+  menu_item = ['Spam', 'Spam', 'Spam', 'Spam', 'Spam', 'Spam', 'Spam', 'Spam',
+    'Baked beans', 'Spam', 'Spam', 'Spam', 'Spam', 'Spam']
+
+  # good
+  menu_item = [
```
jordansissel: Of the two 'good' options here, my preference is this first one:

```ruby
foo = [
  items .... ,
  more, items ...
]
```

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+* Use empty lines between method definitions and also to break up a method into logical paragraphs internally.
+
+  ```Ruby
+  def some_method
+    data = initialize(options)
+
+    data.manipulate!
+
+    data.result
+  end
```
andrewvc: I think this is excessive whitespace. My personal rule is to group statements based on grain and function. In other words:

```
def some_method
  data = initialize(options)
  data.manipulate!
  data.result
end
``` 

makes sense because they all are operating around the `data` object. This would all change however if you had some second key noun, as in the example below:

```ruby
def some_method
  client = TwitterClient.new("arg")
  client.initialize("arg", "arg")

  data = initialize(client)
  data.manipulate!
  data.result
end
```

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  ```Ruby
+  # bad - how many 0s are there?
+  num = 1000000
+
+  # good - much easier to parse for the human brain
+  num = 1_000_000
+  ```
+
+ * Use RDoc and its conventions for API documentation.  Don't put an empty line between the comment block and the `def`.
```
jordansissel: Actually, I think we should use YARD notation, not RDoc notation. Another alternative is tomdoc which I don't think has much tooling, but focuses on human readability.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  * Limit lines to 80 characters.
```
jordansissel: only 80? I usually use 100 these days. Open to discuss.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  * Avoid trailing whitespace.
```
jordansissel: The only reason 'avoid trailing whitespace' is a good idea is because it seems randomly folks editors have this obsessive-configuration that will strip _all_ trailing whitespace when they write to files. This causes unintentional whitespace changes to litter contributors' pull requests and causes confusion.

To avoid this confusion during PR review and to avoid accidental changes, maybe we enforce no trailing whitespace?

Otherwise, trailing whitespace has literally zero impact on the actual code or its readability, so I'm fine if we don't enforce it.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  * End each file with a newline.
+  * Don't use block comments. They cannot be preceded by whitespace and are not as easy to spot as regular comments.
+
+  ```Ruby
+  # bad
+  =begin
```
jordansissel: Funny story.

Back in college, I was helping code stuff for someone's PhD thesis project, and I spent 2 hours trying to tweak some code and _nothing_ ever changed. Turns out about 200 lines above, there was an `#if 0` and about 100 lines below was an `#endif`. I was editing code that wasn't even seen by the compiler. *facepalm*

This `=begin`...`=end` doesn't really impact us, but it's an amusing story. 

This story is actually why I enabled syntax highlighting. For my first 5 years of coding, I never used syntax coloring. After this story, I now rely on coloring to tell me good things like "You are actually editing text that isn't code!" hehe ;P

CODE_STYLE.md
andrewvc: I'm fine personally with keeping both A and B together. I think they're equally clear, and if it cuts down on effort trying to keep stuff consistent I'm OK with that.

andrewvc: That said I would say you should never mix them in the same expression

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  Favor the ternary operator(`?:`) over `if/then/else/end` constructs.
+  It's more common and obviously more concise.
+<sup>[[link](#ternary-operator)]</sup>
+
+  ```Ruby
+  # bad
+  result = if some_condition then something else something_else end
+
+  # good
+  result = some_condition ? something : something_else
```
jordansissel: I'm on the fence. I tend to avoid ternary in practice, especially in ruby, mostly due to the presence of predicate-style method names (ones ending in `?`). This looks weird to me:

```ruby
foo = something? ? bar : baz
```

And for symbolic values, this also looks weird to me:

```ruby
foo = something ? :bar : :baz
```

I am in favor of not using ternary operator if at all possible.

CODE_STYLE.md
andrewvc: I personally just put the args on the next line to keep the whitespace down, that way you only need to indent by 2

CODE_STYLE.md
andrewvc: +1

CODE_STYLE.md
andrewvc: I agree 100 is better. Esp. given that ruby variables are usually not abbreviated.

CODE_STYLE.md
andrewvc: I agree here. I've had fantastic contributions from people with absolutely bonkers editors that stripped whitespace everywhere and made the PR a mess.

CODE_STYLE.md
colinsurprenant: TIL. Amaze.
+1 I also prefer
`class FooError < StandardError; end`

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+Avoid usage of `unless` over negative conditions (or control flow), it
+can cause issues when having multi term expressions.
+<sup>[[link](#unless-for-negatives)]</sup>
+
+  ```Ruby
+  # bad
+  do_something unless some_condition
+
+  # good
+  do_something if !some_condition
```
jordansissel: I'm actually OK with using `unless` only in one circumstance:

1. an in-line condition `expression unless condition`
2. the condition is only a single boolean check, not multiple.
3. the condition isn't already a negative phrase

Examples:

```ruby
# bad, multiple conditions
do_something unless foo && bar

# bad, already negative term
do_something unless not_done

# good, single condition
do_something unless foo
```

andrewvc: +1 . I hate unless, but I do reach for it in that situation.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  # another good option
+  some_condition || do_something
```
jordansissel: This one is OK only if `do_something` is short.

As an example, conditional logging in Logstash is generally done this way:

```ruby
logger.debug? && logger.debug("some very long line", :something => here, :another => value)
```

The reason for this is that having the `if logger.debug?` at the end of the line makes it harder to audit visually. The reason the conditional is even used is because the contextual logging `logger.debug("message", some => hash, context => here)` sets up a Hash object for the `logger.debug` call, and if `logger.debug` is not enabled, we're creating a Hash that is thrown away immediately. It's an optimization.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  names.map(&:upcase)
+  ```
+
+* <a name="single-line-blocks"></a>
+  Prefer `{...}` over `do...end` for single-line blocks.  Avoid using `{...}`
+  for multi-line blocks (multiline chaining is always ugly). Always use
+  `do...end` for "control flow" and "method definitions" (e.g. in Rakefiles and
+  certain DSLs).  Avoid `do...end` when chaining.
+<sup>[[link](#single-line-blocks)]</sup>
+
```
andrewvc: As a fan of FP I find this advice poor. There is nothing wrong with multiline chaining, and in that case `{}` works much better.

I think we should allow this. Otherwise, people will cram too much onto one line.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    # body omitted
+  end
+
+  until x > 5
+    # body omitted
+  end
+```
+
+* <a name="infinite-loop"></a>
+  Use `Kernel#loop` instead of `while/until` when you need an infinite loop.
```
jordansissel: Style-wise, I agree. In practice, however, `while true` is faster. An experiment I did last year - https://github.com/jordansissel/experiments/blob/master/ruby/jruby-loop/test.rb

`loop do ... end` sets up a block and a new context, and I think that's why it's slower, but that's just guessing. Proof:

* Jruby 1.7.19

```
% ruby test.rb
Rehearsal ------------------------------------------------------------------
loop do                          5.670000   0.260000   5.930000 (  6.030000)
while true                       1.860000   0.050000   1.910000 (  1.895000)
--------------------------------------------------------- total: 7.840000sec

                                     user     system      total        real
loop do                          5.000000   0.110000   5.110000 (  5.591000)
while true                       2.220000   0.060000   2.280000 (  2.658000)
```

* MRI 2.1.6

```
ruby test.rb
Rehearsal ------------------------------------------------------------------
loop do                          7.680000   0.100000   7.780000 (  8.075701)
while true                       2.440000   0.030000   2.470000 (  2.642652)
-------------------------------------------------------- total: 10.250000sec

                                     user     system      total        real
loop do                          7.500000   0.070000   7.570000 (  7.678882)
while true                       2.420000   0.010000   2.430000 (  2.443713)
```

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  # good
+  names.each { |name| puts name }
+
+  # bad
+  names.select do |name|
+    name.start_with?('S')
+  end.map { |name| name.upcase }
+
+  # good
+  names.select { |name| name.start_with?('S') }.map(&:upcase)
```
andrewvc: This would look better as:
```ruby
names.
  select {|name| name.start_with?('S') }.
  map(&:upcase)
```

Addiionally, this makes subsequent git revisions far more clear, as each function is described per line.

CODE_STYLE.md
colinsurprenant: personally I prefer no spaces, no sure why it should be different than with array declaration that don't have spaces?
`[1, 2, 3]`
`{:a => 1, :b => 2}`

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    break unless val < 0
+  end
+  ```
+
+* <a name="no-dsl-parens"></a>
+  Omit parentheses around parameters for methods that are part of an internal
+  DSL (e.g. Rake, Rails, RSpec), methods that have "keyword" status in Ruby
+  (e.g. `attr_reader`, `puts`) and attribute access methods. Use parentheses
+  around the arguments of all other method invocations.
+<sup>[[link](#no-dsl-parens)]</sup>
```
jordansissel: There is a common extension to this style point in rspec where parentheses are specifically omitted in RSpec's `expect().to` call:

```ruby
expect(value).to be(...)
```

My preference is to use parentheses in this situation:

```ruby
# good?
expect(value).to(be....)
```

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  end.map { |name| name.upcase }
+
+  # good
+  names.select { |name| name.start_with?('S') }.map(&:upcase)
+  ```
+
+  Some will argue that multiline chaining would look OK with the use of {...},
+  but they should ask themselves - is this code really readable and can the
+  blocks' contents be extracted into nifty methods?
+
```
andrewvc: There are often cases where this just begets stilly methods. Such as

```ruby
private
  # Return ActionResponse instances for a bulk request grouped in a hash by #bin_for_status
  def group_statuses(actions, statuses)
    statuses.
      each.
      with_index.
      map {|status,i| ActionResponse.new(actions[i], status, bin_for_status(status)) }.
      group_by {|action_response| action_response.bin }
  end
```

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  temperance = Person.new('Temperance', 30)
+  temperance.name
+
+  puts temperance.age
+
+  x = Math.sin(y)
+  array.delete(e)
+
+  bowling.score.should == 0
```
jordansissel: ruby's warning system will tell you this is a void statement, btw:

```
% ruby -w -e 'foo = 4; foo == 3'
-e:1: warning: possibly useless use of == in void context
```

Ruby's parser seems to not remember that the `==` operator is a method call.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+* <a name="single-action-blocks"></a>
+  Use the proc invocation shorthand when the invoked method is the only operation of a block.
+<sup>[[link](#single-action-blocks)]</sup>
+
+  ```Ruby
+  # bad
+  names.map { |name| name.upcase }
+
+  # good
+  names.map(&:upcase)
```
jordansissel: +1 this usage

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  # bad
+  def with_tmp_dir
+    Dir.mktmpdir do |tmp_dir|
+      Dir.chdir(tmp_dir) { |dir| yield dir }  # block just passes arguments
+    end
+  end
+
+  # good
+  def with_tmp_dir(&block)
```
jordansissel: +1 always having an explicit block argument if a method is expected to take a block.

CODE_STYLE.md
colinsurprenant: I also prefer the "good" example otherwise the `end` does not align right and tickles my OCD :P

CODE_STYLE.md
ph: Class.new is bad style. Im in favor of the longer syntax.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  def some_method(some_arr)
+    some_arr.size
+  end
+  ```
+
+* <a name="no-self-unless-required"></a>
+  Avoid `self` where not required. (It is only required when calling a self
+  write accessor.)
+<sup>[[link](#no-self-unless-required)]</sup>
+
```
andrewvc: I think this should be more flexible. There are times where I feel strongly about being more explicit about this being a property of the current object.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    if self.last_reviewed_at > self.last_updated_at
+      self.worker.update(self.content, self.options)
+      self.status = :in_progress
+    end
+    self.status == :verified
+  end
+
+  # good
+  def ready?
+    if last_reviewed_at > last_updated_at
```
jordansissel: This method is a bad example, in a way, because it reads like a predicate method, and I think predicate methods should definitely not have side effects. In this example, `worker.update()` reads, to me, to be a side effect. Further, the next line `self.status = ...` is also a side effect. Bad mojo!

CODE_STYLE.md
colinsurprenant: it *really* dislike the proposed syntax, I do not understand this right flushing. I don't see why it shouldn't be a simple 2 spaces indentation just like everything else.

CODE_STYLE.md
colinsurprenant: agree

CODE_STYLE.md
ph: I prefer `{ `  and ` }`, Find it faster to scan through the code but  I don't have a reasoning.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    end
+
+    # bad
+    def do_something(options = {})
+      unless options[:when] == :later
+        output(self.options[:message])
+      end
+    end
+
+    # good
```
jordansissel: +1 on avoiding shadowing even between scopes (local/method arg of `options` conflicting with the instance method `self.options`). Clarity matters, so let's try to avoid having a local var shadowing a method/etc.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  end
+
+  # bad
+  something = something ? something.downcase : nil
+
+  # ok
+  something = something.downcase if something
+
+  # good
+  something = something && something.downcase
```
jordansissel: Ruby needs a `maybe monad`, eh? hehe

CODE_STYLE.md
ph: +1

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    # body omitted
+  end
+
+  def foo(x)
+    bar(x)
+  end
+
+  # also good
+  def foo(x)
+    bar = ->(y) { ... }
```
jordansissel: We should discuss whether we want to encourage the new arrow syntax for lambdas, or further, if we should even use lambdas directly at all.

CODE_STYLE.md
ph: 100 lines, @colinsurprenant set your defaults :+1:

CODE_STYLE.md
ph: +1

CODE_STYLE.md
colinsurprenant: I  mostly do this "normal" indent or use an argument var as you said above

CODE_STYLE.md
colinsurprenant: +1

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  ```Ruby
+  # bad
+  p = Proc.new { |n| puts n }
+
+  # good
+  p = proc { |n| puts n }
+  ```
+
+* <a name="proc-call"></a>
```
jordansissel: My tendency is to use proc (and sometimes lambda) and then regret it later when I could have written the code differently (method definitions, etc). I wonder - should we actively discourage use of lamdba/procs in this style guide. Thoughts?

CODE_STYLE.md
colinsurprenant: seriously? my Apple ][ had a 80 columns video card in 1984.

CODE_STYLE.md
ph: My rule of thumb is, if the ternary operator keep the expression under 100 chars I am ok with it. 
If goes over I don't use it.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  def something(x)
+    _, used_var = something_else(x)
+    # ...
+  end
+  ```
+
+* <a name="global-stdout"></a>
+  Use `$stdout/$stderr/$stdin` instead of `STDOUT/STDERR/STDIN`.
+  `STDOUT/STDERR/STDIN` are constants, and while you can actually reassign
```
jordansissel: I mean, technically, you can do STDOUT.reopen(some_io), but I'll lean towards the globals here.

This shouldn't really affect Logstash anyway since we, in practice, should be avoiding std{out,err} in favor of logging. Exceptions are the stdout output and stdin input, and maybe some other small outliers.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+* <a name="solid-design"></a>
+  Try to make your classes as
+  [SOLID](http://en.wikipedia.org/wiki/SOLID_\(object-oriented_design\)) as
+  possible.
+<sup>[[link](#solid-design)]</sup>
+
+* <a name="define-to-s"></a>
+  Always supply a proper `to_s` method for classes that represent domain
+  objects.
+<sup>[[link](#define-to-s)]</sup>
```
andrewvc: The likelihood of everyone actually doing this is low because the utility is low. It'd be cool if we all did this, but it isn't free and the benefits are negligible.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+* <a name="global-stdout"></a>
+  Use `$stdout/$stderr/$stdin` instead of `STDOUT/STDERR/STDIN`.
+  `STDOUT/STDERR/STDIN` are constants, and while you can actually reassign
+  (possibly to redirect some stream) constants in Ruby, you'll get an
+  interpreter warning if you do so.
+<sup>[[link](#global-stdout)]</sup>
+
+* <a name="warn"></a>
+  Use `warn` instead of `$stderr.puts`. Apart from being more concise and
```
jordansissel: Recommend never using $stderr.puts or warn. We should use logger calls instead.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  clear, `warn` allows you to suppress warnings if you need to (by setting the
+  warn level to 0 via `-W0`).
+<sup>[[link](#warn)]</sup>
+
+* <a name="sprintf"></a>
+  Favor the use of `sprintf` and its alias `format` over the fairly cryptic
```
jordansissel: We should choose either sprintf or format, not both.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  `String#%` method.
```
jordansissel: Agreed on not using `String#%`

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  ```Ruby
+  # bad
+  paths = [paths] unless paths.is_a? Array
+  paths.each { |path| do_something(path) }
+
+  # good
+  [*paths].each { |path| do_something(path) }
+
+  # good (and a bit more readable)
+  Array(paths).each { |path| do_something(path) }
```
jordansissel: +1 on this one. I've been very inconsistent in the past and usually relied on the 'bad' example instead. Oops :P

CODE_STYLE.md
ph: I use this style a lot. I am in favor of it.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  and purposefully raising an exception).
+<sup>[[link](#fail-method)]</sup>
+
+  ```Ruby
+  begin
+    fail 'Oops'
+  rescue => error
+    raise if error.message != 'Oops'
+  end
+  ```
```
andrewvc: New to me... I like it!

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  # bad
+  [0, 1, 2, 3].each do |item|
+    if item > 1
+      puts item
+    end
+  end
+
+  # good
+  [0, 1, 2, 3].each do |item|
+    next unless item > 1
```
jordansissel: In this specific example, should we use `Enumerable#select` ?

```ruby
[0, 1, 2, 3].select { |x| x > 1 }.each do |item|
   puts item
end
```

CODE_STYLE.md
colinsurprenant: meh, not sure we need to ban ternary operators be cause some combinations feels weird, I personally really like ternary operators and use them regularly for their expressiveness and conciseness.

CODE_STYLE.md
colinsurprenant: +1 and I love unless ;)

CODE_STYLE.md
jordansissel: But I agree, `next unless ...` is a good practice for blocks like this.

CODE_STYLE.md
ph: Not sure I agree here. (I am a bit out of context but)
When I see a lot chaining done on parameters I usually think we are missing a class to act on the data, doing this make naming the methods a bit easier.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    do_something(v)
+    ...
+  end
+
+  # good
+  v = array.grep(/foo/)
+  if v
+    do_something(v)
+    ...
+  end
```
ph: I don't like doing conditionals and assigment on the same line. It always lead problems.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+* <a name="no-END-blocks"></a>
+  Do not use `END` blocks. Use `Kernel#at_exit` instead.
+<sup>[[link](#no-END-blocks)]</sup>
+
+  ```ruby
+  # bad
+  END { puts 'Goodbye!' }
+
+  # good
+  at_exit { puts 'Goodbye!' }
```
jordansissel: we should avoid at_exit.

jordansissel: Rather, we should avoid both END and at_exit. If a library needs to do something on exit, it should provide a teardown method or function.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  When accessing the first or last element from an array, prefer `first` or
+  `last` over `[0]` or `[-1]`.
+<sup>[[link](#first-and-last)]</sup>
+
+* <a name="set-vs-array"></a>
+  Use `Set` instead of `Array` when dealing with unique elements. `Set`
+  implements a collection of unordered values with no duplicates. This is a
+  hybrid of `Array`'s intuitive inter-operation facilities and `Hash`'s fast
+  lookup.
+<sup>[[link](#set-vs-array)]</sup>
```
andrewvc: I disagree with this in the case of very short lists for two reasons:

1. Ruby has no Set literal, so it is bulkier.
2. Array is faster for low cardinality sets.

I generally prefer Set, but I think code like: `[:a,:b].include?(x)` is much better than `Set.new([:a,b]).include?(x)`. Additionally, from a performance standpoint, the Set option is ludicrous in that it sets up a hash map, hashes each element, then has to hash the argument before doing the lookup. Constant factors count.

CODE_STYLE.md
colinsurprenant: agree. let's also not forget that **usually** the computation done inside the loop will probably be orders of magnitude slower then the loop iteration overhead. So in general it probably does not make any difference but it's good to keep that in mind.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  SOME_CONST = 5
+  ```
+
+* <a name="bool-methods-qmark"></a>
+  The names of predicate methods (methods that return a boolean value) should
+  end in a question mark.  (i.e. `Array#empty?`). Methods that don't return a
+  boolean, shouldn't end in a question mark.
+<sup>[[link](#bool-methods-qmark)]</sup>
+
+* <a name="dangerous-method-bang"></a>
```
jordansissel: We should audit to ensure we aren't doing such dangerous things, right? Use of `method!` as dangerous should be a sign that we need to do better.

CODE_STYLE.md
ph: from experience.. I almost never, never use next. I prefer to use an iterator because I can easily factor that behavior out of the current context and reuse it.

If i see one I usually remove it.. :P

CODE_STYLE.md
colinsurprenant: hm, specifically for rspec, I think we should follow the generally accepted rspec style which I believe is 
```ruby
expect(value).to be(...)
```

maybe @purbon can chime in?

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    end
+
+    def flatten_once
+      dup.flatten_once!
+    end
+  end
+  ```
+
+* <a name="reduce-blocks"></a>
+  When using `reduce` with short blocks, name the arguments `|a, e|`
```
jordansissel: We should decide also if we want to use `Enumerable#reduce` or `Enumerable#inject`

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  hash.has_value?(value)
+
+  # good
+  hash.key?(:test)
+  hash.value?(value)
+  ```
+
+* <a name="hash-fetch"></a>
+  Use `Hash#fetch` when dealing with hash keys that should be present.
+<sup>[[link](#hash-fetch)]</sup>
```
andrewvc: Love this. Safe-gets FTW

CODE_STYLE.md
colinsurprenant: I will rebel against this I promise :P

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  # bad
+  counter += 1 # Increments counter by one.
+  ```
+
+* <a name="comment-upkeep"></a>
+  Keep existing comments up-to-date. An outdated comment is worse than no
+  comment at all.
+<sup>[[link](#comment-upkeep)]</sup>
+
+> Good code is like a good joke - it needs no explanation. <br>
```
jordansissel: This quote is not good. Code can be good but missing immediate context and the decisions that produced the code. Not needing an explanation means all readers have the necessary context (background, etc) to understand the joke (or code).

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+> -- Russ Olsen
+
+* <a name="refactor-dont-comment"></a>
+  Avoid writing comments to explain bad code. Refactor the code to make it
```
jordansissel: I'd rather have good comments on bad code than bad code with no comments. Code is a work in progress. Without explanation of context (comments) then bad code stays hard to maintain.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+* <a name="string-interpolation"></a>
+  Prefer string interpolation and string formatting instead of string
+  concatenation:
+<sup>[[link](#string-interpolation)]</sup>
+
+  ```Ruby
+  # bad
+  email_with_name = user.name + ' <' + user.email + '>'
+
+
```
andrewvc: I don't see why we need to be consistent here. Both work, both are clear. There's an imperceptible difference.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+* <a name="indent-annotations"></a>
+  If multiple lines are required to describe the problem, subsequent lines
+  should be indented three spaces after the `#` (one general plus two for
+  indentation purpose).
+<sup>[[link](#indent-annotations)]</sup>
+
+  ```Ruby
+  def bar
+    # FIXME: This has crashed occasionally since v3.2.1. It may
```
jordansissel: I don't have a strong opinion here. I don't see why indentation is helpful in this case. Thoughts?

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+* <a name="rare-eol-annotations"></a>
+  In cases where the problem is so obvious that any documentation would be
+  redundant, annotations may be left at the end of the offending line with no
+  note. This usage should be the exception and not the rule.
+<sup>[[link](#rare-eol-annotations)]</sup>
+
+  ```Ruby
+  def bar
+    sleep 100 # OPTIMIZE
```
jordansissel: seems like this should be a `# TODO: OPTIMIZE` not just `# OPTIMIZE` or maybe I"m not understanding what is being strictly demonstrated as this example beyond just an inline comment.

CODE_STYLE.md
andrewvc: This seems excessive. Why do we care about trimming the whitespace here? If you need the whitespace trimmed, do this, otherwise, it seems nutty.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  end
+  ```
+
+* <a name="todo"></a>
+  Use `TODO` to note missing features or functionality that should be added at
```
jordansissel: Historically (from a habit I picked up at Google many years ago), I've marked `TODO` items with my name `TODO(sissel): ...` as this helps communicate who is authoring the todo. This helps in reviewing code later to know who to ask about the TODO. It's more explicit than using things like `git blame` which can shadow the original author through formatting/styling/etc updates.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  a later date.
+<sup>[[link](#todo)]</sup>
+
+* <a name="fixme"></a>
+  Use `FIXME` to note broken code that needs to be fixed.
+<sup>[[link](#fixme)]</sup>
+
+* <a name="optimize"></a>
+  Use `OPTIMIZE` to note slow or inefficient code that may cause performance
```
jordansissel: `OPTIMIZE` isn't clear enough to me. `TODO: optimize` is much more clear.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+<sup>[[link](#consistent-classes)]</sup>
+
+  ```Ruby
+  class Person
+    # extend and include go first
+    extend SomeModule
+    include AnotherModule
+
+    # inner classes
+    CustomErrorKlass = Class.new(StandardError)
```
jordansissel: We should avoid this specific syntax because `Class.new()` has some bugs (https://bugs.ruby-lang.org/issues/11119) and also has no clear name. `class CustomErrorKlass < StandardError; end` is best because it is named.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+    # initialization goes between class methods and other instance methods
+    def initialize
+    end
+
+    # followed by other public instance methods
+    def some_method
+    end
+
+    # protected and private methods are grouped near the end
```
jordansissel: we shouldn't use protected. I feel it's not meaningful in ruby.

CODE_STYLE.md
andrewvc: YES.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    protected
+
+    def some_protected_method
+    end
+
+    private
```
jordansissel: I propose we actually enforce that `private` is the default and that we explicitly expose public methods:

```ruby
class Greeting
  private  # default everything to private

  def initialize(message)
    @message = message
  end

  def greet(object)
    "#{@message}, #{object}"
  end

  # Explicitly expose the `#greet` method.
  public(:greet)
end

g = Greeting.new("Hola")
puts g.greet("Jordan")
```

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  # good
+
+  # foo.rb
+  class Foo
+    # 30 methods inside
+  end
+
+  # foo/bar.rb
+  class Foo
```
jordansissel: `class Foo::Bar`

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    class Bar
+      # 30 methods inside
+    end
+  end
+
+  # foo/car.rb
+  class Foo
```
jordansissel: `class Foo::Car`

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    class Car
+      # 20 methods inside
+    end
+  end
+  ```
+
+* <a name="modules-vs-classes"></a>
+  Prefer modules to classes with only class methods. Classes should be used
```
jordansissel: +1

CODE_STYLE.md
andrewvc: I love my ternary operator.

`over_my_cold_dead_body ? can_take_away_my_ternary_operator : cannot_take_it_away`

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+      # body omitted
+    end
+
+    def self.some_other_method
+    end
+  end
+
+  # good
+  module SomeModule
+    module_function
```
jordansissel: Should we use `module_function` instead of `extend self` ?

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+      # body omitted
+    end
+
+    def some_other_method
+    end
+  end
+  ```
+
+* <a name="module-function"></a>
+  Favor the use of `module_function` over `extend self` when you want to turn
```
jordansissel: haha, they expected me to ask this question.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+      # do stuff here
+    end
+
+    def other_utility_method(number, string)
+      # do some more stuff
+    end
+  end
+  ```
+
+* <a name="liskov"></a>
```
jordansissel: We should document examples of this

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  When designing class hierarchies make sure that they conform to the [Liskov
+  Substitution
+  Principle](http://en.wikipedia.org/wiki/Liskov_substitution_principle).
+<sup>[[link](#liskov)]</sup>
+
+* <a name="solid-design"></a>
```
jordansissel: We should document examples of this

CODE_STYLE.md
andrewvc: By the way, have we considered using block syntax here to cut down on redundancy? That's how ruby's built in logger works.

`logger.debug { ... `} . Keeps the execution time fast and makes things clear.

CODE_STYLE.md
jordansissel: +1, having a to_s isn't important. Ruby's `#inspect` is quite useful _and_ we can use `pry` to debug with a REPL. I don't see much value in requiring `#to_s` for everything.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  ```Ruby
+  class Person
+    attr_reader :first_name, :last_name
+
+    def initialize(first_name, last_name)
+      @first_name = first_name
+      @last_name = last_name
+    end
+
+    def to_s
```
jordansissel: This example is an interesting counter-point, in my opinion. Not all locales represent a full name as "Firstname Lastname". Japan is "LASTNAME Firstname" isn't it? IMO, this example is a reason for using localization and explicitly avoiding to_s :)

CODE_STYLE.md
andrewvc: Eeeeenteresting Jordan. Still, I agree with colin, I would say we should default to `loop`, and if you can prove you need the speed boost of `while true` put it in a comment explaining why. Most micro-optimizations wind up not doing much.

CODE_STYLE.md
andrewvc: I agree with @colinsurprenant . I like my rspec vanilla.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  Avoid the use of `attr`. Use `attr_reader` and `attr_accessor` instead.
+<sup>[[link](#attr)]</sup>
+
+  ```Ruby
+  # bad - creates a single attribute accessor (deprecated in 1.9)
+  attr :something, true
+  attr :one, :two, :three # behaves as attr_reader
+
+  # good
+  attr_accessor :something
```
jordansissel: Actually, I've been trending towards avoiding mutators (and `attr_accessor`) mostly due to Ruby's typeless system. I want a mutator to validate input.

```ruby
def foo=(value)
  raise InvalidFoo, "The value #{value.inspect} is not a valid foo." unless value.is_a?(Foo)
  @foo = value
end
```

CODE_STYLE.md
andrewvc: My Lisp is showing. That being said. Pulling out that very specific map statement into a class would make it harder to understand IMHO. Additionally, you wind up with a million open `.rb` files as you navigate the project. I think concision here makes sense :shrugs:.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  attr_reader :one, :two, :three
+  ```
+
+* <a name="struct-new"></a>
+  Consider using `Struct.new`, which defines the trivial accessors,
```
jordansissel: NEVER USE THIS.

`Struct.new` returns a new Class instance for each call:

```ruby
[3] pry(main)> Struct.new(:foo)
=> #<Class:0x007f981c376fc0>
[4] pry(main)> Struct.new(:foo)
=> #<Class:0x007f981c3459e8>
```

Different object ids for essentially the same class definition. This can cause problems in a few places, most notably sometimes libraries will require a file twice (by accident, somehow) and stuff like this:


```ruby
class Foo < Struct.new(:whatever)
  ...
end
```

The result is similar to this pry example:

```ruby
[5] pry(main)> class Foo < Struct.new(:whatever); end
=> nil
[6] pry(main)> class Foo < Struct.new(:whatever); end
TypeError: superclass mismatch for class Foo
```
```

File: CODE_STYLE.md (hunk 34 above)
Comment (andrewvc) [inline CODE_STYLE.md]: @jordansissel I agree 100% this is a bad example. The only way I could forgive them here is if it lazy evaluation were a goal, but even then I'd have a separate `update_state` method.

File: CODE_STYLE.md (hunk 37 above)
Comment (andrewvc) [inline CODE_STYLE.md]: I use them so rarely it's not a huge issue for me. I could go either way.

File: CODE_STYLE.md (hunk 72 above)
Comment (jordansissel) [inline CODE_STYLE.md]: Further, Struct.new gives you a basic class w/ just accessors. These accessors won't have validators for your input, so you can get bugs that way. Recommend just thin classes so we can do tests:

```
# bad
Foo = Struct.new(:whatever)

# good
class Foo
  attr_reader :whatever
  def whatever=(value)
    raise Something, "Invalid whatever: #{value}" unless value.is_a?(Numeric)
   @whatever = value
  end
end
```

File: CODE_STYLE.md (hunk 38 above)
Comment (andrewvc) [inline CODE_STYLE.md]: I agree, even as a lisper!

If you use a lambda you should have a reason why. The main place where I'll actually use a lambda is when creating dynamic mappings.

```ruby
handlers = {
  foo: lambda {...},
  bar: lambda {...}
}
```

CODE_STYLE.md
andrewvc: I like it for stuff like `launch_the_nukes`. For instance, when writing billing code `process_payment!` is a good indication that you better invoke that thing carefully. For a project like logstash, that's less risky, but occasions may still arise.

CODE_STYLE.md
andrewvc: I think that `map` and `reduce` are more widely known terms, so I prefer them.

CODE_STYLE.md
andrewvc: Agreed! For tricky code I need to know why something is happening!

CODE_STYLE.md
andrewvc: I agree, this is nutty. Indent as you like I say.

CODE_STYLE.md
andrewvc: +1

CODE_STYLE.md
andrewvc: +1

CODE_STYLE.md
andrewvc: +1

purbon:
This looks amazing, thanks a lot @jordansissel, and others, for this in deep collaboration. I will follow up the comments in the next days, and as soon as we reach consensus, will be consolidating sections. I will do this section by section, so I don't loss track of all the comments.

This is going to be a great addition to Logstash.

CODE_STYLE.md
colinsurprenant: +1 on avoiding the okish

CODE_STYLE.md
colinsurprenant: +1 YARD notation for API documentation. I ---think--- I have mostly be the only one actually documenting methods and I've been doing it YARD style.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+   # bad
+   def some_method_with_parameters param1, param2
+     # body omitted
+   end
+
+   # good
+   def some_method_with_parameters(param1, param2)
+     # body omitted
+   end
+   ```
```
colinsurprenant: I just want to stress that I am strongly in favour of strict compliance to 
```ruby
def some_method
```
and 
```ruby
def some_method(param1, param2)
```
same for invokation
```ruby
some_method
```
and
```ruby
some_method(param1, param2)
```
my only 2 exceptions are `raise` and `return`.

CODE_STYLE.md
colinsurprenant: it is still slower because a closure is created at each debug call

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    end
+  end
+
+  # better
+  Person = Struct.new(:first_name, :last_name) do
+  end
+  ````
+
+* <a name="no-extend-struct-new"></a>
+  Don't extend an instance initialized by `Struct.new`. Extending it introduces
```
jordansissel: Recommend striking this in favor of "Do not use Struct" and "Do not use OpenStruct"

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  ````
+
+* <a name="factory-methods"></a>
+  Consider adding factory methods to provide additional sensible ways to
+  create instances of a particular class.
+<sup>[[link](#factory-methods)]</sup>
+
+  ```Ruby
+  class Person
+    def self.create(options_hash)
```
jordansissel: I don't like options hashes.

If we're going to do factory pattern, I think the builder pattern is nice (Elasticsearch uses it all over the place)

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+      # body omitted
+    end
+  end
+  ```
+
+* <a name="duck-typing"></a>
```
jordansissel: This one impacts the Plugin heirarchy (LogStash::Plugin -> LogStash::Inputs::Base -> LogStash::Inputs::Stdin, for example). We should decide and then refactor to fit whatever model we choose.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  class Dog
+    def speak
+      puts 'Bau! Bau!'
+    end
+  end
+  ```
+
+* <a name="no-class-vars"></a>
+  Avoid the usage of class (`@@`) variables due to their "nasty" behavior in
```
jordansissel: Agreed, though maybe not for the same reasons. Class vars are probably better off as namespaced globals or accessed through a singleton.

CODE_STYLE.md
andrewvc: @colinsurprenant I agree. And even in tight loops, the loop has to be fairly tight. See the benchmarks below:

```
irb(main):065:0* Benchmark.measure { 10_000_000.times { is_level? && puts("hai") } }
=> #<Benchmark::Tms:0x4f0f2942 @stime=0.0, @label="", @cstime=0.0, @real=0.7079999446868896, @total=0.7000000000000028, @cutime=0.0, @utime=0.7000000000000028>
irb(main):066:0> Benchmark.measure { 10_000_000.times { block_taker { "whaatup" } } }
=> #<Benchmark::Tms:0x2657d4dd @stime=0.0, @label="", @cstime=0.0, @real=1.0509998798370361, @total=1.1099999999999994, @cutime=0.0, @utime=1.1099999999999994>
```

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+      def second_method_etc
+        # body omitted
+      end
+    end
+  end
+  ```
+
+* <a name="alias-method-lexically"></a>
+  Prefer `alias` when aliasing methods in lexical class scope as the
```
jordansissel: This point `use alias keyword` seems to be counter or at least in mild conflict with the next point about using `alias_method`

CODE_STYLE.md
colinsurprenant: just want to add that in critical sections, I would stick with `while true` because using `loop` implies calling a closure on every iteration - again, this is really micro-optimization and not too elegant Ruby-wise but I can live with that :P

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  class Sting < Westerner
+    include Mononymous
+  end
+  ```
+
+## Exceptions
+
+* <a name="fail-method"></a>
+  Signal exceptions using the `fail` method. Use `raise` only when catching an
```
jordansissel: I don't see exceptions as a category of failures, but a signal of a boundary condition being exceeded. Failure is a subset of things that exceed boundary conditions, so `raise`, in my opinion, is more correct than `fail` for the general case.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  exception and re-raising it (because here you're not failing, but explicitly
+  and purposefully raising an exception).
+<sup>[[link](#fail-method)]</sup>
+
+  ```Ruby
+  begin
+    fail 'Oops'
+  rescue => error
+    raise if error.message != 'Oops'
```
jordansissel: I would love it if we never used strings for exception conditions. I greatly prefer exception types instead.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  Don't specify `RuntimeError` explicitly in the two argument version of
+  `fail/raise`.
+<sup>[[link](#no-explicit-runtimeerror)]</sup>
+
+  ```Ruby
+  # bad
+  fail RuntimeError, 'message'
+
+  # good - signals a RuntimeError by default
+  fail 'message'
```
jordansissel: Recommend never using a string like this. We should always have an exception class w/ context. Strings are harder to test and manage than classes specifically because you can rescue a class directly -

`rescue SomeSpecificException ...`

vs

`rescue => e; if e.message == "some specific exception" ...`

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  to `fail/raise`, instead of an exception instance.
+<sup>[[link](#exception-class-messages)]</sup>
+
+  ```Ruby
+  # bad
+  fail SomeException.new('message')
+  # Note that there is no way to do `fail SomeException.new('message'), backtrace`.
+
+  # good
+  fail SomeException, 'message'
```
jordansissel: While this is common, it still misses structure and good context.

Like we have `logger.info("message", :context => "goes", :here => true)` we should somehow have an exception that includes both a message (what went wrong) and context (state/data during the exception). Not just a message.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  # good
+  def foo
+    # main logic goes here
+  rescue
+    # failure handling goes here
+  end
+  ```
+
+* <a name="contingency-methods"></a>
+  Mitigate the proliferation of `begin` blocks by using *contingency methods*
```
jordansissel: +1 this, I think it's easier to read and makes it easier to test

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  do_something rescue nil
+  ```
+
+* <a name="no-rescue-modifiers"></a>
+  Avoid using `rescue` in its modifier form.
+<sup>[[link](#no-rescue-modifiers)]</sup>
+
+  ```Ruby
+  # bad - this catches exceptions of StandardError class and its descendant classes
+  read_file rescue handle_error($!)
```
jordansissel: agreed, inline rescue is bad mojo.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  # good - this catches only the exceptions of Errno::ENOENT class and its descendant classes
+  def foo
+    read_file
+  rescue Errno::ENOENT => ex
+    handle_error(ex)
+  end
+  ```
+
+* <a name="no-exceptional-flows"></a>
+  Don't use exceptions for flow of control.
```
jordansissel: +1

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+    # .. handle error
+  ensure
+    f.close if f
+  end
+  ```
+
+* <a name="auto-release-resources"></a>
+Use versions of resource obtaining methods that do automatic
+resource cleanup when possible.
+<sup>[[link](#auto-release-resources)]</sup>
```
jordansissel: But never use things which have finalizers. (Tempfile, etc)

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+  (and you don't need to maintain Ruby 1.9 compatibility). Apply this rule only
+  to arrays with two or more elements.
+<sup>[[link](#percent-i)]</sup>
+
+  ```Ruby
+  # bad
+  STATES = [:draft, :open, :closed]
+
+  # good
+  STATES = %i(draft open closed)
```
jordansissel: eh. do not like this.

CODE_STYLE.md
```diff
@@ -0,0 +1,3293 @@
...
+* <a name="use-hash-blocks"></a>
+  Prefer the use of the block instead of the default value in `Hash#fetch`
+  if the code that has to be evaluated may have side effects or be expensive.
+  <sup>[[link](#use-hash-blocks)]</sup>
+
+  ```Ruby
+  batman = { name: 'Bruce Wayne' }
+
+  # bad - if we use the default value, we eager evaluate it
+  # so it can slow the program down if done multiple times
```
jordansissel: A counter example of this is that if the default value is constant, then it's better to use `batman.fetch(:powers, "some value")` hardcoded especially in a critical latency path because block setup is costly. 

This is related slightly to benchmarks checking the performance of blocks vs function arguments in logger calls: https://github.com/jordansissel/experiments/tree/master/ruby/logger-string-vs-block

Blocks had a higher overhead than function arguments.

CODE_STYLE.md
jordansissel: My personal preference is option B (double quotes everywhere) but my most important feeling for the project is consistency. 

We could adopt this strategy:
* single quotes everywhere
* use of string interpolation should be discouraged  and we should figure out why string interp is needed and find a better way?

jordansissel: I didn't even know this was valid ruby.

jordansissel: Oh this is clever.

jordansissel: We shouldn't use global Regexp.last_match anyway. Always use the MatchData returned from the regexp execution:

```ruby
>> re = /hello (\w+)/
>> data = re.match("hello world")
=> #<MatchData "hello world" 1:"world">
>> data[1]
=> "world"
```

jordansissel: If you need MatchData, use `RegExp#match` not the `=~` operator.

jordansissel: We should probably never use %x or backtick operators. If we need to run a subprocess and capture the output, popen or its friends are superior.

jordansissel: *cough* active_support *cough*

jordansissel: +1 on avoiding method_missing

jordansissel: this++

jordansissel: My opinion is that clamp is what we should use for command lines.

jordansissel: OMG I finally got to the end of this file! I GET A PRIZE!!!!

This is exciting. :)

STYLE.md
```diff
@@ -1,96 +0,0 @@
-# Style Guide
```
jordansissel: Good bye, old friend.

STYLE.md
```diff
@@ -1,96 +0,0 @@
...
-* parentheses on function definitions/calls
-* explicit is better than implicit
-  * implicit returns are forbidden except in the case of a single expression 
-
-The point is consistency and documentation. If you see inconsistencies, let me
-know, and I'll fix them :)
-
-Short example:
-
-      require "something from a gem" # from gem 'thing'
```
jordansissel: I do like this notation, but maybe it's not needed. The intent is to be very clear as to what gem provides a given library.

STYLE.md
```diff
@@ -1,96 +0,0 @@
...
-      # some documentation about this class
-      class Foo < Bar
-        # some documentation about this function
-        def somefunc(arg1, arg2, arg3)
-          # comment
-          puts "Hello"
-          if (some_long_condition \
-              or some_other_condition)
-            puts "World"
-          end # if <very short description>
```
jordansissel: I also still prefer putting a comment on `end` for long blocks to describe context. It's helpful when methods/blocks are either unavoidably long _or_ are simply just long and could be refactored.

CODE_STYLE.md
andrewvc: You're right, and I'm not arguing. I WILL point out however that `=~` is concise, and, little known fact, `$1`, `$2`, etc. are actually thread safe! #perl4evar

CODE_STYLE.md
jordansissel: > not intended for GC

Except in Ruby 2.2 symbols are GC'd. Hehe ;P

I get what you mean, though. 

I am willing to concede and use hash-colon syntax, buuuuuut my general feel of hash-colon is that they're usually for option hashes, and option hashes (in this style guide) are frowned upon.

The one place I like to use an options hash is in logging w/ context:

```ruby
logger.info("Some really nice description of the a thing", the: context, goes: here)
```

All other hash literals should probably be destroyed if we can do so.

CODE_STYLE.md
jordansissel: @colinsurprenant your reasoning on consistency w/ array literals and method calls and boolean groupings - I agree. 

+1 on `{key => value, key2 => value2}` (no space after `{` and no space before `}`)

CODE_STYLE.md
jordansissel: I agree, sometimes I try to write it like this:

```ruby
case; when first_case
  when second_case
  when ...
end
```

Then we dont' have `case` on a line by itself and it avoids the jarring double-unindent from a final `when`/`else` block's `end`

I'm also open to just accepting `case` on a line by itself with perhaps some extra prodding ;)

CODE_STYLE.md
jordansissel: I do feel like this syntax is a kind of funky code smell. If the assignment is so complex as to need a case statement, that case statement belongs in a separate function as said below.

CODE_STYLE.md
jordansissel: > if there was some long boilerplate needed for some reason with 40-50 method calls

If so, we could just metaprogram that nonsense if it came down to it. FTW does this for `FTW::Agent#get` and similar methods: https://github.com/jordansissel/ruby-ftw/blob/master/lib/ftw/agent.rb#L158-L183

+1 on avoiding `okish`

CODE_STYLE.md
jordansissel: I am OK with this also.

CODE_STYLE.md
andrewvc: I generally am against structs, but I think they work in places like this: https://github.com/andrewvc/logstash-output-elasticsearch/blob/reuse-errored-ids/lib/logstash/outputs/elasticsearch.rb#L539

I wanted just a bit more typing and concision.

CODE_STYLE.md
jordansissel: @andrewvc agreed. In as much as code can be grouped into paragraphs, your final example above has two 'paragraphs' and it reads well with a blank line between the two paragraphs.

CODE_STYLE.md
jordansissel: +1 YARD.

I've been documenting API calls sporadically using a style that appears to be a random number generator. I have no idea. haha

CODE_STYLE.md
jordansissel: @colinsurprenant IMPLICIT 100 character width in an off-screen buffer YOU KNOW.

CODE_STYLE.md
jordansissel: @andrewvc exactly. +1 on enforcing trailing-whitespace stripping.

The only functional impact this has today is on Grok which treats trailing whitespace on patterns as part of the pattern. However, I don't think we've had a problem with this yet.

CODE_STYLE.md
jordansissel: Some support for this ternary operator, eh?

CODE_STYLE.md
jordansissel: See https://github.com/jordansissel/experiments/tree/master/ruby/logger-string-vs-block for my prior research on this. 

Running today on my laptop:

```
% ruby rubylogger.rb
         string | jruby/  1.9.3 |     0.50
      string_if | jruby/  1.9.3 |     0.13
     string_and | jruby/  1.9.3 |     0.13
     control_if | jruby/  1.9.3 |     0.12
    control_and | jruby/  1.9.3 |     0.12
          block | jruby/  1.9.3 |     0.26
    doubleblock | jruby/  1.9.3 |     0.28
```

I don't think it tests `logger.info("string") { some_hash_context_here }` but it does test blocks.

Specifically, it tries to test the case where something is _not_ needing to be logged and we want to avoid extra work.

CODE_STYLE.md
jordansissel: I'm actually OK with multiple `do ... end` being chained for readability. 

My rough guide today is:
* single-line block is always `{ ... }`
* multi-line block is always `do ... end`
* chaining any blocks together is ok, such as:

```ruby
# This is OK to me, most of the time
foo.select do |val|
  some fancy thing with val
end.sort.each do |val|
  something with val
end
```

CODE_STYLE.md
jordansissel: @andrewvc I agree, mostly, except that I prefer continuations to start with an operator:

```ruby
names
  .select { ... }
  .map ...
```

jordansissel: More practically, it may read like this to avoid `names` on a line by itself:

```
names.select { ... }
  .map { ... }
  .sort
```

or something

CODE_STYLE.md
jordansissel: I see how it is. Y'all are coming for my `(` and `)` keys...

CODE_STYLE.md
jordansissel: I am OK being flexible. I trend towards not using `self` in my own habits.

CODE_STYLE.md
jordansissel: I think the only big place we use lambda is in the date filter, and that I wrote it with lambdas gives me great regrets (it's hard to read, hard to maintain).

CODE_STYLE.md
jordansissel: Yeah I'm down with using lambdas for dispatch tables (hash w/ lambdas as values). 

However, Ruby also let's you do this with `send` anyway:

```
class Dispatcher
  def handle_foo(...); end
  def handle_bar(...); end
  def handle(thing, ...)
    send("handle_#{thing}", ...)
  end
end
```

Inner methods and anonymous callbacks like this make it a little hard to test (does the foo lambda work?, etc) where as with individual methods (in this specific case) can be tested more unit-test style.

CODE_STYLE.md
jordansissel: check my comment here: https://github.com/elastic/logstash/pull/3485#discussion_r32988440

CODE_STYLE.md
jordansissel: Same. Both C and Go seem to encourage this kind of thing, but i feel most of the time it obfuscates things. Checking a condition _and_ assigning a variable in the same line - too much at once - visually.

CODE_STYLE.md
jordansissel: +1 I don't think we have strong need for Set at this time.

CODE_STYLE.md
jordansissel: Indeed. I feel like if we find an instance where we need to discuss "should we use interpolation or concatenation?" it is because the string being created is _too_ complex for either of the options in this interpolation-concatenation dichotomy. The correct answer, in that case, feels like we should use a text templating tool instead (mustache, `Array#join`, or something else)

jordansissel:
Some of my concerns are less about a style recommendation and more about the examples given. For the examples that are confusing, we can hopefully improve these and contribute the new examples back to the ruby style guide project as PRs.

colinsurprenant:
omg, this is already a monster :P
with the amount of comments and the potential of discussion over a lot of them, I am afraid this will never end. let's try and see.

CODE_STYLE.md
colinsurprenant: I am not sure why we should ban it, it could be useful, provided the right justification, personally I don't recall having used it but that does not mean it could not be used...

CODE_STYLE.md
colinsurprenant: not sure why we should diverge from the widely accepted Ruby structure of having everything public by default and grouping private at the end. it feel very natural for me to read top-down a class knowing everything at the top is public down to the `private` statement...

CODE_STYLE.md
colinsurprenant: no. I'll expand later on the subject.

purbon:
@jordansissel @colinsurprenant We'll take it easy, I will go through  the comments in small batches, so we can reach consensus per steps, then we'll be able to move forward steady. If you guys agree, I prefer to track this as an ongoing task, and as soon as we reach consensus we can applied, but obviously not for next week. Makes sense?

purbon:
closing this in benefit of having a minimum style guide we all should follow.

PR #3505, opened 2015-06-26
Add unicode trimmer utility function

andrewvc:
This is mainly in support of https://github.com/logstash-plugins/logstash-output-sns/issues/7

I looked around the ruby world, and I couldn't find an intelligent unicode byte trimmer. This algorithm is probably a lot more sane than rails' (which just goes byte by byte).

### DIFF
--- a/lib/logstash/util/unicode_trimmer.rb
+++ b/lib/logstash/util/unicode_trimmer.rb
@@ -1,0 +1,80 @@
+module UnicodeTrimmer
+  # The largest possible unicode chars are 4 bytes
+  # http://stackoverflow.com/questions/9533258/what-is-the-maximum-number-of-bytes-for-a-utf-8-encoded-character
+  # http://tools.ietf.org/html/rfc3629
+  MAX_CHAR_BYTES = 4
+
+  # Takes a unicode string and makes sure it fits in a max of `desired_bytes`
+  # This aims to be somewhat efficient about this for the average case and get as close to
+  # O(1) as possible. Given certain distributions of multi-byte characters it'll be slower
+  # It tries to find the point the truncation *should* happen based on the average byte size.
+  # If that snips it in the wrong place it'll try to add or remove chars to get it to the right
+  # spot and preserve as much data as possible.
+  public
+  def self.trim_bytes(orig_str, desired_bytes)
+    return orig_str if orig_str.bytesize <= desired_bytes
+
+    pre_shortened = pre_shorten(orig_str, desired_bytes)
+
+    case pre_shortened.bytesize <=> desired_bytes
+    when 0
+      pre_shortened
+    when 1
+      shrink_bytes(pre_shortened, orig_str, desired_bytes)
+    when -1
+      grow_bytes(pre_shortened, orig_str, desired_bytes)
+    end
+  end
+
+  private
+  # Try to cut the string at the right place based on the avg. byte size
+  def self.pre_shorten(orig_str, desired_bytes)
+    # Compute the average size to get an idea of where should chop
+    orig_len = orig_str.length
+    orig_bs = orig_str.bytesize
+    avg_size = (orig_bs.to_f / orig_len.to_f)
+
+    # Try to do an initial shortening based on the average char size
+    # The goal here is to get us somewhere above or below the boundary quickly
+    orig_extra_bytes = orig_bs - desired_bytes
+    pre_shorten_by = (orig_extra_bytes  / avg_size).to_i
+    orig_str.slice(0, orig_len - pre_shorten_by)
+  end
+
+  private
+  def self.grow_bytes(pre_shortened, orig_str, desired_bytes)
+    res_str = pre_shortened.clone()
+
+    loop do
+      bs = res_str.bytesize
+      deficit = desired_bytes - bs
+      lengthen_by = deficit / MAX_CHAR_BYTES
+      lengthen_by = 1 if lengthen_by < 1
+      append = orig_str.slice(res_str.length, lengthen_by)
+
+      break if (bs + append.bytesize) > desired_bytes
+
+      res_str << append
+    end
+
+    res_str
+  end
+
+  private
+  def self.shrink_bytes(pre_shortened, orig_str, desired_bytes)
+    res_str = pre_shortened.clone()
+
+    loop do
+      bs = res_str.bytesize
+      break if bs <= desired_bytes
+
+      extra = bs - desired_bytes
+      shorten_by = extra / MAX_CHAR_BYTES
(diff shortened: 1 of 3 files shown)


spec/util/unicode_trimmer_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    it "should truncate to exact byte boundaries when possible" do
+      expect(UnicodeTrimmer.trim_bytes(ustr, 21).bytesize).to eql(21)
+    end
+
+    it "should truncate below the bytesize when splitting a byte" do
+      expect(UnicodeTrimmer.trim_bytes(ustr, 52).bytesize).to eql(51)
+    end
+
+    it "should not truncate the string when the bytesize is already OK" do
+      expect(UnicodeTrimmer.trim_bytes(ustr, ustr.bytesize))
```
jordansissel: missing `.to ...` matcher?

andrewvc: Good catch, fixed!

lib/logstash/util/unicode_trimmer.rb
```diff
@@ -0,0 +1,69 @@
...
+    orig_extra_bytes = orig_bs - desired_bytes
+    pre_shorten_by = (orig_extra_bytes  / avg_size).to_i
+    pre_shortened = orig_str.slice(0, orig_len - pre_shorten_by)
+  end
+
+  private
+  def self.grow_bytes(pre_shortened, orig_str, desired_bytes)
+    res_str = pre_shortened.clone()
+
+    while (bs = res_str.bytesize) && (bs <= desired_bytes)
```
jordansissel: not a huge fan of assignment in a condition, especially since it's unlikely `x.bytesize` returns a falsey value (0 is truthy). It reads weird to me.

jordansissel: Should the condition be `bs < desired_bytes` ? Otherwise we'll attempt to lengthen it even after we have a good length, right?

andrewvc: The main reason I did an assignment here was to avoid calling '#bytesize' twice. It might be a needless microoptimization. 

What do you recommend? I could replace `while` with `loop` and use a `break` to move this logic inline. Is that better?

lib/logstash/util/unicode_trimmer.rb
```diff
@@ -0,0 +1,69 @@
...
+    res_str
+  end
+
+  private
+  def self.shrink_bytes(pre_shortened, orig_str, desired_bytes)
+    res_str = pre_shortened.clone()
+
+    while (bs = res_str.bytesize) && (bs > desired_bytes)
+      extra = bs - desired_bytes
+      shorten_by = extra / 6 # UTF-8 char max bytesize
```
jordansissel: Let's make 6 a constant and use it in both shrink and grow methods. Also documenting a citation for that would be useful also :)

lib/logstash/util/unicode_trimmer.rb
andrewvc: Also, good point on it being `<` not `<=`. will fix.

jordansissel: +1 loop + break

lib/logstash/util/unicode_trimmer.rb
```diff
@@ -0,0 +1,74 @@
+module UnicodeTrimmer
+  # The largest possible unicode chars are 4 bytes
+  # http://stackoverflow.com/questions/9533258/what-is-the-maximum-number-of-bytes-for-a-utf-8-encoded-character
+  # http://tools.ietf.org/html/rfc3629
+  MAX_CHAR_BYTES = 4
```
jordansissel: Hah, I thought it was 6 also. The RFC indicates otherwise though. 5- and 6-byte sequences were removed (section 10 Security of the RFC) for security/buffer sizing problems, it seems.

jordansissel:
This patch adds randomization to ensure random strings are within the expected bounds (equal or minus one character):

```diff
diff --git a/Gemfile b/Gemfile
index a6d7d67..aac1d5c 100644
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem "simplecov", :group => :development
 gem "coveralls", :group => :development
 gem "rspec", "~> 3.1.0", :group => :development
 gem "logstash-devutils", "~> 0", :group => :development
+gem "flores", :group => :development
 gem "benchmark-ips", :group => :development
 gem "octokit", "3.8.0", :group => :build
 gem "stud", "~> 0.0.19", :group => :build
diff --git a/spec/util/unicode_trimmer_spec.rb b/spec/util/unicode_trimmer_spec.rb
index 36ff2d5..6d4e72c 100644
--- a/spec/util/unicode_trimmer_spec.rb
+++ b/spec/util/unicode_trimmer_spec.rb
@@ -1,6 +1,12 @@
 # encoding: utf-8
 require "spec_helper"
 require "logstash/util/unicode_trimmer"
+require "flores/rspec"
+require "flores/random"
+
+RSpec.configure do |config|
+  Flores::RSpec.configure(config)
+end
 
 describe "truncating unicode strings correctly" do
   context "with extra bytes before the snip" do
@@ -35,4 +41,15 @@ describe "truncating unicode strings correctly" do
     end
   end
 
+  context "randomized testing" do
+    let(:text) { Flores::Random.text(1..1000) }
+    let(:size) { Flores::Random.integer(1..text.bytesize) }
+    let(:expected_range) { (size - 4)..size }
+
+    stress_it "should be near the boundary of requested size" do
+      #expect(UnicodeTrimmer.trim_bytes(text, size).bytesize).to be(
+      expect(expected_range).to include(UnicodeTrimmer.trim_bytes(text, size).bytesize)
+    end
+  end
+
 end
```

jordansissel:
Results of my patch:

```
% bundle exec rspec spec/util/unicode_trimmer_spec.rb -e 'random'
Run options:
  include {:full_description=>/random/}
  exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integrat
ion=>true, :windows=>true}
..............................................................................................................................................................
..............................................................................................................................................................
..............................................................................................................................................................
..........................................................................................

Finished in 1.13 seconds (files took 1.37 seconds to load)
564 examples, 0 failures
```

andrewvc:
@jordansissel tiiight! It actually works! I'll apply it.

jordansissel:
Code LGTM. Tests pass. Ship it.

jordansissel:
Jarvis is having some trouble merging this into 1.5

PR #3509, opened 2015-06-29
Fix dependency loading for Linux and old releases of JDK7.

purbon:
As seen in CI, JDK7, specially Java HotSpot(TM) 64-Bit Server VM 1.7.0_60-b19 has a different behaviour than expected when loading classes with ```Dir.glob```. This has been addressed in this PR.

CI test can be seen at: http://build-eu-00.elastic.co/job/Logstash_regression_Master_(coverage_fix)/

### DIFF
--- a/lib/pluginmanager/install.rb
+++ b/lib/pluginmanager/install.rb
@@ -1,3 +1,4 @@
+require "pluginmanager/command"
 require "jar-dependencies"
 require "jar_install_post_install_hook"
 require "file-dependencies/gem"
--- a/lib/pluginmanager/list.rb
+++ b/lib/pluginmanager/list.rb
@@ -1,4 +1,5 @@
 require 'rubygems/spec_fetcher'
+require "pluginmanager/command"
 
 class LogStash::PluginManager::List < LogStash::PluginManager::Command
 
--- a/lib/pluginmanager/main.rb
+++ b/lib/pluginmanager/main.rb
@@ -11,7 +11,6 @@ module LogStash
 end
 
 require "clamp"
-require "pluginmanager/command"
 require "pluginmanager/util"
 require "pluginmanager/gemfile"
 require "pluginmanager/install"
--- a/lib/pluginmanager/uninstall.rb
+++ b/lib/pluginmanager/uninstall.rb
@@ -1,4 +1,7 @@
+require "pluginmanager/command"
+
 class LogStash::PluginManager::Uninstall < LogStash::PluginManager::Command
+
   parameter "PLUGIN", "plugin name"
 
   def execute
--- a/lib/pluginmanager/update.rb
+++ b/lib/pluginmanager/update.rb
@@ -1,3 +1,4 @@
+require "pluginmanager/command"
 require "jar-dependencies"
 require "jar_install_post_install_hook"
 require "file-dependencies/gem"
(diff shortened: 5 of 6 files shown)


purbon:
@suyograo reviews welcome.

ph:
@purbon The build mention in this issue is green? What is the difference between the JDK7/8 `Dir.glob`, is that a known jruby issue?

purbon:
@ph if you check the link you will see the build is green (is a public link to everyone). ```Dir.glob``` with regular expressions fetch stuff in different order in linux for JDK7 as in JDK8.

ph:
@purbon I'll be more clear, the build link you linked in this issue is green for both JDK7 and JDK8. After looking at the build history, I think the failing build for this issue would be this one: http://build-eu-00.elastic.co/job/Logstash_regression_Master_(coverage_fix)/jdk=JDK7,label=metal-pool/5/console

``ruby
Dir.glob("lib/**/*.rb")
``

I think you mean using **wildcards** path in glob return files in different order for `JDK7 and JDK8` ? This is a bit unexpected? 

Could it be only a ruby namespace issue not correctly declared? I am saying this because we have a `https://github.com/elastic/logstash/blob/master/lib/logstash/namespace.rb` to fix kind of problem for logstash

purbon:
Yes, http://build-eu-00.elastic.co/job/Logstash_regression_Master_(coverage_fix)/5/ is the job descriving the issue.

Yeah! I meant exactly this, wildcards in Dir.glob works funny with boths versions and linux vs mac. Happy to me other times. 

I don't think is a ruby namespace, if you go to linux machine, and use the same JDK as in Jenkins you will see this. Namespaces where actually not the problem, but:

```
ill not capture all events without --debug flag
rake aborted!
NameError: uninitialized constant LogStash::PluginManager::Command
/home/jenkins/workspace/Logstash_regression_Master_(coverage_fix)/jdk/JDK7/label/metal-pool/lib/pluginmanager/update.rb:5:in `(root)'
```

as you can see from the error message.

let me know if you have more questions.

ph:
@purbon I haven't checked the JDK order issue, but would it be more appropriate to make sure the `pluginmanager` correctly require his dependencies like we do everywhere else? This would make sure that glob ordering behavior would never impact the require ordering. 

In the `update` file we explicitly require the `command` class:
https://github.com/elastic/logstash/blob/master/lib/pluginmanager/update.rb#L5

purbon:
@ph As you might know, nowadays all comes required in the ```main.rb``` file, I can add that if you like to, but the way dependencies are loaded right now, seems reasonable to me as there is no need to require command in every file, as the entry point is always main.rb. 

Let me know what do you want me to do.

- purbon

purbon:
@ph please give me directions to make you happy, so I can fix this and activate coverage again. Thank for your time.

ph:
@purbon The problem I see with this strategy is we will need to maintain it when we encounter errors and I find it less proactive to me. Also if we add testing to the CLI commands, we will still need to require the command class either in the tested class or in the spec file.

purbon:
Makes sense, I will do that as adding test for each command makes sense.

- purbon

purbon:
@ph I just update this with your last comments, test are as before http://build-eu-00.elastic.co/view/LS%20Features/job/Logstash_regression_Master_(coverage_fix)/

let me know if you like it :-)

ph:
LGTM

PR #3516, opened 2015-06-29
allow Logstash program path to be overridden in init script config files

marko-asplund:


jordansissel:
This init script is only intended for use with our rpm and deb packages which install to /opt/logstash/... - What causes you to need to change the path?

jordansissel:
(also, this init script is only recommended for ancient linux distributions (RHEL5/CentOS5/etc) which lack upstart or systemd.

marko-asplund:
Writing and testing init scripts that actually work correctly is tedious and takes some time, so I'd like to be able to re-use the Logstash init script while still using the Logstash tar distribution on a custom path.

jordansissel:
@marko-asplund I think i understand your needs, but I dont' think your patch will fix it. If you want the path to be configurable, you'll need to make this LS_PROG variable overridable externally. Are you using `/etc/{default,sysconfig}/logstash` for this?

pkg/logstash.sysv
```diff
@@ -36,12 +36,12 @@ LS_CONF_DIR=/etc/logstash/conf.d
...
 LS_NICE=19
 LS_OPTS=""
-
+LS_PROG=/opt/logstash/bin/logstash
 
 [ -r /etc/default/$name ] && . /etc/default/$name
 [ -r /etc/sysconfig/$name ] && . /etc/sysconfig/$name
 
-program=/opt/logstash/bin/logstash
+program=$LS_PROG
```
jordansissel: This will need to be quoted:
```shell
program="$LS_PROG"
```

marko-asplund:
@jordansissel yes, I'm planning to use `/etc/{default,sysconfig}/logstash` for overriding program path.
`$LS_PROG` is now quoted in the assignment.

PR #3519, opened 2015-06-29
Do not include rackspace input/output in tests

### LINKED ISSUE #3518: Remove logstash-[in|out]put-rackspace from default_plugins.rb
Until https://github.com/logstash-plugins/logstash-input-rackspace/issues/2 and https://github.com/logstash-plugins/logstash-output-rackspace/issues/3 are resolved, the rackspace input and output plugins cannot be in the default bundle.

untergeek:
Block testing by regex.

fixes #3518


jordansissel:
LGTM

talevy:
LGTM2

PR #3521, opened 2015-06-29
Add changelog 1.5.2

suyograo:


CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
...
+  - Performance improvements: String interpolation is widely used in LS to create keys combining dynamic
+    values from extracted fields. Added a caching mechanism where we compile this template on first use
+    and reuse them subsequently, giving us a good performance gain in configs that do a lot of date 
+    processing, sprintf, and use field reference syntax (#3425).
+  - Added warning when LS is running on a JVM version which has known issues/bugs (#2547).  
+  - Updated AWS based plugins to v2 of AWS ruby SDK. This involves an update to s3-input, s3-output,
+    sqs-input, sns-output.
+
+### input
+  - Lumberjack: This input was not handling backpressure properly from downstream plugins and
```
jordansissel: Let's add that the `max_clients` setting now has no effect because it is not needed due to this improved circuit breaker. Pending https://github.com/logstash-plugins/logstash-input-lumberjack/pull/32

suyograo: Done

ph:
Lgtm maybe mention plugins have their own changelog?

suyograo:
@ph we have a lot of plugins with no changelog content in them so until its available IMO we should keep changes here.

PR #3528, opened 2015-06-30
update Gemfile.lock file with new released plugins and dependencies

purbon:
so we get the last released plugins with fixes :100:


suyograo:
LGTM

PR #3525, opened 2015-06-30
Initial draft of Deploying and Scaling Logstash.

palecur:

PR #3548, opened 2015-07-02
Deploying and Scaling RC commit

palecur:
See former PR, #3525, for historical context and discussion. This PR supersedes 3525.


suyograo:
LGTM, most of the reviews were done in #3548 and addressed by @palecur

PR #3564, opened 2015-07-07
make plugin#inspect show :number config parameters

jsvd:


spec/core/plugin_spec.rb
```diff
@@ -29,6 +29,22 @@ class LogStash::Filters::LadyGaga < LogStash::Filters::Base
     expect(LogStash::Plugin.lookup("filter", "lady_gaga")).to eq(LogStash::Filters::LadyGaga)
   end
 
+  describe "#inspect" do
+    class LogStash::Filters::MyTestFilter < LogStash::Filters::Base
```
purbon: Can we make this part of a helpers file ? or locate it somewhere else. Having the definition of a class here looks strange to me.

spec/core/plugin_spec.rb
```diff
@@ -29,6 +29,22 @@ class LogStash::Filters::LadyGaga < LogStash::Filters::Base
...
+      config_name "param1"
+      config :num, :validate => :number, :default => 20
+      config :str, :validate => :string, :default => "test"
+    end
+    subject { LogStash::Filters::MyTestFilter.new("num" => 1, "str" => "hello") }
+
+    it "should print the class of the filter" do
+      expect(subject.inspect).to match(/LogStash::Filters::MyTestFilter/)
```
purbon: Good use of the match helper. I was wondering if we should make sure things are printed in a given order?

spec/core/plugin_spec.rb
jsvd: can I maybe wrap this into a let?

like `let (:new_plugin_class) { .. }`

lib/logstash/plugin.rb
```diff
@@ -111,9 +111,9 @@ def clear_watchdog
 
   public
   def inspect
-    if !@config.nil?
-      description = @config \
-        .select { |k,v| !v.nil? && (v.respond_to?(:empty?) && !v.empty?) } \
+    if !@params.nil?
+      description = @params \
```
purbon: Is the backslash really need here?

spec/core/plugin_spec.rb
purbon: I like it.

lib/logstash/plugin.rb
jsvd: no, just keeping with the current format, some might like an explicit line break, I have no strong feelings about it

spec/core/plugin_spec.rb
colinsurprenant: not sure I see why it's not a good idea to define a test class here?

lib/logstash/plugin.rb
purbon: I would say clean them :-) they look strange to me, but at the end up to you :smile:

spec/core/plugin_spec.rb
purbon: Is a personal taste of mine, I like to have this classes either wrap in lets or defined in a helper file somewhere else, just to organize stuff. Defining classes like this in test cases tend to pollute the code, at less in my experience.

lib/logstash/plugin.rb
```diff
@@ -111,9 +111,9 @@ def clear_watchdog
...
+        .reject { |k,v| v.nil? || (v.respond_to?(:empty?) && v.empty?) } \
```
colinsurprenant: syntax nitpicking but: `k, v` instead of `k,v` ?

spec/core/plugin_spec.rb
colinsurprenant: this is a class specific for this test, I don't see the pollution and having that in a separate file would make it really less clear. here we can straight away see that a test class is defined and the test is performed on it.

colinsurprenant:
LGTM - did not run tests.

spec/core/plugin_spec.rb
jsvd: I can't seem to properly use a let here to create a class. I can use a 

```
let(:new_filter_class) { Class.new(Filters::Base)  do ... end }
````

but then the new_filter_class will not have a name I can later match against in the spec because it's an anonymous class.
I'd then consider leaving this as is.

purbon: Go ahead and ignore me. I was just doing the statement that we should avoid having class definition in the spec files, this should be created in auxiliary files as they tend to pollute, but this is just my personal taste.

purbon:
LGTM - did not run the tests.

PR #3574, opened 2015-07-08
Port & Protocol are required field

ebuildy:


untergeek:
@ebuildy, thanks for going through our docs and submitting a PR.

Unfortunately, neither `port` nor `protocol` are required.  If no value is specified for `port`, it defaults to 9200.  If no value is specified for `protocol`, it defaults to `node`.

Perhaps there is something we could do to make this more clear.  Would you be willing to add text to that effect?

jordansissel:
@ebuildy any thoughts on @untergeek's comments?

suyograo:
Closing this since 2.0 allows `host:port` setting

Issue #3576, opened 2015-07-08
Support private gem repositories for plugins


purbon:
The current implementation of the command ```bin/plugin [install]``` ignores any gem repo that is not rubygems.

To reproduce this I used https://github.com/geminabox/geminabox, build a plugin myself, that is not released at rubygems, add the new gem source in several ways, like

* gem sources -a [gem_server]
* Adding it below the line ```source 'https:///rubygems.org'``` in the Gemfile. and update the Gemfile.lock
* Make it the only possible source in the Gemfile, also updating the Gemfile.lock.

If we aim to support private gem repositories, we should load all sources. 

Related: #2376

purbon:
Adding the new source manually around https://github.com/elastic/logstash/blob/master/lib/pluginmanager/main.rb#L12 lets the installation to finish.

colinsurprenant:
@purbon not sure I understand your last comment about adding the new source in `main.rb` ?!

In any case, manually adding `source "{source_url}"` in the `Gemfile` *should* work... if it doesn't I think this is where we should focus for a fix.

purbon:
@colinsurprenant my last comment was just a note that adding this, while debugging the problem got the plugin installed, not a proposed solution at all.

manually adding the source to the Gemfile does not work to install plugins that are *only* in the private repo. But I agree adding a new source is the way to fix this, we just need to find a proper way to tell bundler about it.

purbon:
Clarification:

the process that fails is the verification, not the installation perse.

```
─purbon@purbon-elastic  ~/work/logstash  ‹fix/plugin_manager_and_private_repo*› 
╰─$ bin/plugin install logstash-input-log4r 
Validating logstash-input-log4r
Plugin logstash-input-log4r does not exist
ERROR: Installation aborted, verification failed for logstash-input-log4r 
╭─purbon@purbon-elastic  ~/work/logstash  ‹fix/plugin_manager_and_private_repo*› 
╰─$ bin/plugin install --no-verify logstash-input-log4r                                                                                   1 ↵
Installing logstash-input-log4r
Installation successful
```

colinsurprenant:
@purbon AH!!! yes, makes sense... the verification directly talks to `rubygems.org` regardless of the sources... good catch.

purbon:
@colinsurprenant made a proposal to fix this in #3583, feel free to review when you get a minute.

purbon:
wondering if makes sense to enable a couple of notations in our Gemfile, something like:

[...]

Issue #2631, opened 2015-02-17
bin/plugin install --development should be sticky


colinsurprenant:
currently after doing 

``` sh
bin/plugin install --development
```

to install development dependencies, if we do any other plugin installation like

``` sh
bin/plugin install logstash-input-foobar
```

then the development dependencies are not available anymore and any rspec run will fail.

Once `bin/plugin install --development` has been executed development dependencies should always be installed/visible, at least until `bin/plugin install --no-development` is called...

ph:
To see the actual log of the behavior look at this gist: https://gist.github.com/jsvd/983af2dad9a144c602ba

jsvd:
And also https://gist.github.com/jsvd/dde24df1acbb34b56771

ph:
We could keep the state in a file and minic the bundler behavior.
``
.pluginmanager/config ?
```

colinsurprenant:
well, the state is already in the `.bundle/config`with the `BUNDLE_WITHOUT` I think we should try to use that first instead of having a separate state file...

ph:
Because of the smalls issues we had with `bundler/config` maybe it is wiser to define our own state file? 
Also this is not a `bundler` behavior but a pluginmanager behavior.

ph:
Let's use the:
`Bundler.settings['STICKY_DEVELOPMENT"]`

ph:
^ previous command will save this information to the `.bundler/config`, so we can use bundler settings hash to access it.

jordansissel:
I *hate* the sticky feature of Bundler, btw. I even go as far as forcing `rm -rf .bundle` on every bundle call:
https://github.com/jordansissel/dotfiles/blob/master/.zshrc#L51-L56

I think a sticky flag would irrittate me on bin/plugin, too.

ph:
@jordansissel It's only affecting us the developers, 
I  am perfectly OKAY to have to do `bin/plugin install --development` after installing a new plugin? Maybe the fix is to actually document it?

ph:
Let me bring up another related bundler subject.

Our current `plugin install` actually create a `.bundler/config` file, I guess you would be in favor of not persisting anything to disk?

ph:
@jordansissel You were [concerned](https://github.com/elastic/logstash/issues/2631#issuecomment-76284738) to add `sticky development` feature for the dev dependencies. This problem isn't blocking our development but we as the developers we are running into this issue often.

[...]

PR #3583, opened 2015-07-09
Fix plugin_manager install and update commands to work properly with no internet env private gem repos

purbon:
Add all defined sources to rubygems so verification can talk to all the repositories, even the private ones.

Fix the workflow for using a private gem repo and fixed issues with #3576 .

Fixes https://github.com/elastic/logstash/issues/3576


purbon:
@colinsurprenant would you mind reviewing this ? thanks man ! What is your take on simply adding all the sources at the very beginning as done in this PR for the verification ?

It would simplify things like  ```bundler_options[:rubygems_source] = gemfile.gemset.sources ``` done in the install_gems_list method.

jordansissel:
I admit I feel happy that this change seems simple given our historical frictions trying to use Bundler from Ruby ;P

colinsurprenant:
Yay!

@purbon is that testable in any way?

purbon:
Yes, use the procedure described in #3576

colinsurprenant:
Oh, I meant, testable in a spec.

purbon:
No, I can add a test with some mocks that basically check if the verification gets all the sources. This makes sense. I will add it.

purbon:
Feel free test it manually, so you can check it works as expected.  What about my question?

colinsurprenant:
Which question? you mean what I think about this proposal? A priori it looks like a fine solution! and we can followup in #3576 about the other potential sources notations that we could also add in a 2nd phase.

purbon:
@colinsurprenant talking about https://github.com/elastic/logstash/pull/3583#issuecomment-120059028

---- 
What is your take on simply adding all the sources at the very beginning as done in this PR for the verification ? It would simplify things like bundler_options[:rubygems_source] = gemfile.gemset.sources done in the install_gems_list method.
--- 

we're basically doing the same thing twice here now.

lib/pluginmanager/install.rb
```diff
@@ -43,6 +43,7 @@ def validate_cli_options!
   # Check if the specified gems contains
   # the logstash `metadata`
   def verify_remote!(gems)
+    Gem.sources = Gem::SourceList.from(gemfile.gemset.sources)
     gems.each do |plugin, version|
       puts("Validating #{[plugin, version].compact.join("-")}")
       signal_error("Installation aborted, verification failed for #{plugin} #{version}") unless LogStash::PluginManager.logstash_plugin?(plugin, version)
```
guyboertje: This is a v v v long line.  Suggest -
```ruby
gems.each do |plugin, version|
  next if LogStash::PluginManager.logstash_plugin?(plugin, version)
  signal_error("Installation aborted, verification failed for #{plugin} #{version}")
end
```

purbon: makes sense @guyboertje, we should use every PR to format code better.

guyboertje: does signal_error terminate the block?
otherwise it would be possible to validate all then signal all errors e.g.
```ruby
failed = []
gems.each do |plugin, version|
  next if ...
  failed << "verification failed for #{plugin} #{version}"
end
msg = failed.join(', ')
signal_error msg.prepend('Installation aborted,  ')
```

purbon: Yes, it terminates. Why not using another PR, so make this cleaner? I don't like the idea of doing lot of refactoring, changing the scope of the PR. What do you think?

guyboertje: Sure.

purbon: Will apply, the first cleanup that makes a lot of sense, and then use another PR to clean this a bit more. Thanks for your feedback @guyboertje !!

purbon:
@colinsurprenant just add an small spec to check for this kind of things.

purbon:
@colinsurprenant is this ready to be merged?

purbon:
@suyograo would you be so nice to review this and let me know if is in health to be merged? thanks a lot!

lib/pluginmanager/install.rb
```diff
@@ -43,9 +43,11 @@ def validate_cli_options!
   # Check if the specified gems contains
   # the logstash `metadata`
   def verify_remote!(gems)
+    Gem.sources = Gem::SourceList.from(gemfile.gemset.sources)
     gems.each do |plugin, version|
       puts("Validating #{[plugin, version].compact.join("-")}")
-      signal_error("Installation aborted, verification failed for #{plugin} #{version}") unless LogStash::PluginManager.logstash_plugin?(plugin, version)
+      next unless LogStash::PluginManager.logstash_plugin?(plugin, version)
+      signal_error("Installation aborted, verification failed for #{plugin} #{version}")
```
colinsurprenant: I am not sure the logic is correct here. Previously it was:
"signal error if not a valid logstash plugin"

and the new logic is now:
"proceed to next check if not a valid logstash plugin otherwise continue and signal error (if it is a valid plugin)"

was this tested? also, the proposed line breakup of using `next unless ...` is a lot harder to understand and follow the logic.

lib/pluginmanager/update.rb
```diff
@@ -6,6 +6,7 @@ class LogStash::PluginManager::Update < LogStash::PluginManager::Command
   REJECTED_OPTIONS = [:path, :git, :github]
 
   parameter "[PLUGIN] ...", "Plugin name(s) to upgrade to latest version", :attribute_name => :plugins_arg
+  option "--[no-]verify", :flag, "verify plugin validity before installation", :default => true
```
colinsurprenant: Is this new option related the original intent fixing the Gem sources to enable private repos support?

lib/pluginmanager/update.rb
```diff
@@ -52,9 +52,8 @@ def update_gems!
 
     # any errors will be logged to $stderr by invoke!
     # Bundler cannot update and clean gems in one operation so we have to call the CLI twice.
-    output = LogStash::Bundler.invoke!(:update => plugins)
+    output = LogStash::Bundler.invoke!(:update => plugins, :rubygems_source => Gem::SourceList.from(gemfile.gemset.sources))
```
colinsurprenant: I think the `Gem::SourceList.from` should be pushed into the `LogStash::Bundler` class as a separation of concerns. the `LogStash::Bundler.invoke!` method is a logstash method and I think it does not need to know about the SourceList in its api, it will take care of transforming this structure to whatever is required.

lib/pluginmanager/install.rb
```diff
@@ -43,9 +43,11 @@ def validate_cli_options!
   # Check if the specified gems contains
   # the logstash `metadata`
   def verify_remote!(gems)
+    Gem.sources = Gem::SourceList.from(gemfile.gemset.sources)
```
colinsurprenant: I think `Gem.sources` should be setup in `LogStash::PluginManager.logstash_plugin?`  right before the `Gem::SpecFetcher.fetcher.spec_for_dependency` call https://github.com/elastic/logstash/blob/0cf1a98b4a77f3278e814e1a1cb308fd184934ba/lib/pluginmanager/util.rb#L20  no?

purbon: not sure is the right place for it, as what happen if then we've another call that aim to interact with the gem repos?, I guess we should make this happen in some kind of initialize method, run when the plugin manager gets up and running ... what do you think?

colinsurprenant: - first there is no other calls in `lib/pluginmanager/install.rb`, in fact there is **no** calls at all, everything is handled at a lower level.

- I really prefer to explicitly set the sources **where it is needed** instead of globally setting it in an optimistic way hoping it will cover all cases. In the first case it is precise and explicit, you can see the relation between the sources and the call that requires it, in the other case you don't know where it was done before in the code path... there are lots of chances future refactor will break this and hope there is a spec to catch it. 

So, no, I would definitely stick it in the `spec_for_dependency` call.

lib/pluginmanager/update.rb
purbon: No, this option is to skip a validation process done later in the code, you can see it https://github.com/elastic/logstash/blob/master/lib/pluginmanager/update.rb#L69 where there is a warning when updating between major versions. 

The code used there ```rubygems/gems``` only works for rubygems, not for other gem repositories.

purbon:
please jenkins, test this.

purbon:
please jenkins, test this.

lib/bootstrap/bundler.rb
```diff
@@ -79,7 +79,7 @@ def invoke!(options = {})
       LogStash::Bundler.patch!
 
       # force Rubygems sources to our Gemfile sources
-      ::Gem.sources = options[:rubygems_source] if options[:rubygems_source]
+      ::Gem.sources = Gem::SourceList.from(options[:rubygems_source]) if options[:rubygems_source]
```
colinsurprenant: interesting. so this was actually not working before?

lib/pluginmanager/command.rb
```diff
@@ -1,8 +1,16 @@
 class LogStash::PluginManager::Command < Clamp::Command
+
+  def initialize(invocation_path, context = {}, parent_attribute_values = {})
+    initialize_gemfile
+    super(invocation_path, context, parent_attribute_values)
+  end
+
```
colinsurprenant: why is this refactor required?

lib/pluginmanager/command.rb
```diff
@@ -35,4 +43,12 @@ def relative_path(path)
     require "pathname"
     ::Pathname.new(path).relative_path_from(::Pathname.new(LogStash::Environment::LOGSTASH_HOME)).to_s
   end
+
+  private
+  def load_gemfile
+    gemfile = LogStash::Gemfile.new(File.new(LogStash::Environment::GEMFILE_PATH, 'r+')).load
+    Gem.sources = Gem::SourceList.from(gemfile.gemset.sources)
+    gemfile
```
colinsurprenant: any reason why we globally set Gem.sources here?

lib/pluginmanager/update.rb
colinsurprenant: ok, then, ideally this should have been in another PR since this is not related to this PR intent.

lib/pluginmanager/command.rb
colinsurprenant: Ok, now I understand that this refactor was done to accommodate setting `Gem.sources` here. I don't think we should do this here, for the same reasons I gave about not setting `Gem.sources` globally. Setting `Gem.sources` definitely does not belong here. Also this refactor breaks the purpose of having a lazy `@gemfile`.

lib/pluginmanager/update.rb
purbon: Well, my intent with this PR was to fix the plugin manager for private repos, I tried to put a descriptive title:

```
Fix plugin_manager install and update commands to work properly with no internet env private gem repos
```

I apologies for the confusions.

lib/pluginmanager/command.rb
purbon: I did this to show one idea of loading sources once, and actually not having to go to each call that need and load the sources because ```rubygems.org```might not be the only one there. If we aim to support private repos, and offline environments, we've to think there are others repositories possible.

I think we should find a solution where sources are loaded once, i can not understand why sources should not be loaded once for all the plugin manager interactions. I will explore other solutions, but at the end loading sources is something that might be better done once, as it's not going to cause any problem, or slow down, just simplifying Gem configuration.

What do you think?

purbon: I think I know why you aim to have a lazy gemfile, so not to load it until is need. What might be the problems for you to actually load the gemfile once at the "initialization" of the plugin manager?

colinsurprenant: ok 2 things here:
- first, setting `Gem.sources` does not belong in the `Command` class. if you look at how the plugin manager is architectured you will see that there's no direct interaction with the `Gem` module, everything is decoupled through the `LogStash::Bundler` and `LogStash::Rubygems` modules (and in the pluginmanager utils.rb). Second, I do not like the idea of a globally setting `Gem.sources`, this is a setting that should be kept close to the calls that require it so that it is explicit. it will be easier to refactor in the future and/or support options like setting specific sources for specific calls for example.

- refactoring the lazy initialization stuff does not belong in this PR, if we want to revisit why/how this is done here, let's create another PR, AFAIK this is not a problem right now.

purbon: Ok! will revert the commit and implement your idea.

colinsurprenant:
@purbon I see you reverted some changes but now `Gem.sources` is back in `verify_remotes` and my previous comment still applies, it should be in `LogStash::PluginManager.logstash_plugin?` I think.

purbon:
@colinsurprenant I reverted the commits to actually make happen what you requested, this code is landed now. Please let me know if you have any more concerns about it.

purbon:
After giving it more thought, I think we should not be passing a gemfile, we should be passing a settings hash, I will implement this tomorrow if you agree with me @colinsurprenant ?

lib/pluginmanager/util.rb
```diff
@@ -1,12 +1,35 @@
 require "rubygems/package"
 
 module LogStash::PluginManager
+
+  class Settings
```
colinsurprenant: why do we need a settings class here?

colinsurprenant: So to expand a bit on that, I think that if we want to introduce a global settings concept, this is not the right place to do that, we should think and discuss how we want to do that. for this PR purpose here I don't understand why a simple parameter for the sources is no ok?

purbon: I kinda like the idea of having this settings in a way similar of what Bundler.settings does. I find it super useful for later invocations or usages of the same parameters, and for actually keeping the knowledge where it belongs. 

In the current implementation the only place where the PluginManager life, is the PluginManager module, there is no other object who might be holding this plugin manager settings. Do you have a proposal where we can keep this? 

I actually think that either naming this module ```PluginManager``` or the file ```utils.rb``` is wrong. Do we aim to have a plugin manager concept? there would be a nice place to hold this settings idea, and many other concepts useful for this codebase.

But this is just my point of view.

The ```Settings``` idea is removed as you requested.

purbon:
please jenkins, test this.

lib/pluginmanager/update.rb
colinsurprenant: Sorry, I am still confused. is the `--no-verify` option required for offline installation?

purbon: Yes! because we introduced a feature to warn for major versions. Please take a look at the code and you will understand. 

The code used only talks to rubygems.org and not to others private repositories that might not share the same api.

lib/pluginmanager/util.rb
```diff
@@ -17,6 +19,7 @@ def self.logstash_plugin?(plugin, version = nil)
       end
     else
       dep = Gem::Dependency.new(plugin, version || Gem::Requirement.default)
+      Gem.sources = Gem::SourceList.from(options[:sources]) if options[:sources]
```
colinsurprenant: in `LogStash::Bundler#invoke!` https://github.com/purbon/logstash/blob/13da8e2f6612c1a4af3377d9bbb1875fd0a791e6/lib/bootstrap/bundler.rb#L82 we use the `rubygems_sources` option, we should be consistent and either use `rubygems_sources` or `source` for both.

lib/pluginmanager/util.rb
colinsurprenant: ok. you can bring your proposal in a separate issue/PR. also note that this also probably relates to #2631

lib/pluginmanager/update.rb
colinsurprenant: ok. now I get it. the version verification is hard-wired against rubygems.org through the usage of the `gems` gem. this should probably be better documented.

lib/pluginmanager/util.rb
colinsurprenant: I agree with the `LogStash::PluginManager` vs `util.rb` confusion, opened issue #3746 for that.

lib/pluginmanager/util.rb
purbon: Hi,
  I used sources here because using rubygems_sources is confusing in the context of other gem repositories. I will change this here to be in the same page as bundler, as you requested.

colinsurprenant:
@purbon there is still a mismatch in option names: in `invoke!` it is `:rubygems_source` and in `verify_remote!` its `:rubygems_sources` ... let's pick one please. also, I'd be ok if we just change them all to `:sources`.

purbon:
@colinsurprenant I forgot that, I apologies.

purbon:
@colinsurprenant I will do the rename to sources in another PR.

lib/pluginmanager/util.rb
```diff
@@ -1,12 +1,14 @@
...
 
 module LogStash::PluginManager
+
   # check for valid logstash plugin gem name & version or .gem file, logs errors to $stdout
   # uses Rubygems API and will remotely validated agains the current Gem.sources
   # @param plugin [String] plugin name or .gem file path
   # @param version [String] gem version requirement string
   # @return [Boolean] true if valid logstash plugin gem name & version or a .gem file
-  def self.logstash_plugin?(plugin, version = nil)
+  def self.logstash_plugin?(plugin, version = nil, options={})
```
colinsurprenant: it would be nice to keep the method description comments up to date with the new options parameter.

lib/pluginmanager/util.rb
```diff
@@ -1,12 +1,15 @@
 require "rubygems/package"
 
 module LogStash::PluginManager
+
   # check for valid logstash plugin gem name & version or .gem file, logs errors to $stdout
   # uses Rubygems API and will remotely validated agains the current Gem.sources
   # @param plugin [String] plugin name or .gem file path
   # @param version [String] gem version requirement string
+  # @param options [Hash] Configuration options used to setup external components
```
colinsurprenant: actually, the proper way to documents options hash is like this, as described in http://www.rubydoc.info/gems/yard/file/docs/Tags.md#option 

```ruby
# @param [Hash] options the options used to setup external components
# @option options [Array<String>] :rubygems_source Gem sources to lookup for the verification
```

colinsurprenant:
at this point to move this PR forward I would like to see the result of thorough manual tests of the feature and the procedure to reproduce these tests. Also, I would like to make sure this PR passes the current test kitchen integration tests and see if we can add integration tests for this feature too.

purbon:
The way to reproduce this test are easy:

* spin up a local gem repo, for example geminabox.
* add the new source in your Gemfile.
* Remove internet from your machine.
* run the install command (you should be pulling gems out of your local repo)

colinsurprenant:
What about uninstall and update? what about having only the local source and having both rubygems.org and the local source, etc. we need to test all scenarios.

purbon:
@colinsurprenant all this test where tested in my development environment in the description I was using install only as example of a command you can run.

Please feel free to validate this issue yourself, or let me know how can I show you this works in my machine? might be should I make a video and shared here? not sure how to add integration test here as this commands are dependent to have an external gem repo. 

Specs where added to track the sources were actually loaded into the system.

lib/pluginmanager/install.rb
```diff
@@ -44,9 +44,11 @@ def validate_cli_options!
   # Check if the specified gems contains
   # the logstash `metadata`
   def verify_remote!(gems)
+    options = { :rubygems_source => gemfile.gemset.sources }
```
ph: messing with the sources should be hidden inside the bundler class, the install class doesn't need that kind of information.

lib/pluginmanager/update.rb
```diff
@@ -53,9 +53,8 @@ def update_gems!
 
     # any errors will be logged to $stderr by invoke!
     # Bundler cannot update and clean gems in one operation so we have to call the CLI twice.
-    output = LogStash::Bundler.invoke!(:update => plugins)
+    output = LogStash::Bundler.invoke!(:update => plugins, :rubygems_source => gemfile.gemset.sources)
```
ph: Some comment, the install doesn't need to know the concept of the rubygems source.

lib/pluginmanager/install.rb
purbon: This part does not mess with sources :+1: , that it does is pulling out the sources from the gemfile and let later know rubygems. If I do understand you right, you mean I should do what? pass the gemfile to the bundler class? can you elaborate a bit more please?

lib/pluginmanager/util.rb
```diff
@@ -17,6 +21,7 @@ def self.logstash_plugin?(plugin, version = nil)
       end
     else
       dep = Gem::Dependency.new(plugin, version || Gem::Requirement.default)
+      Gem.sources = Gem::SourceList.from(options[:rubygems_source]) if options[:rubygems_source]
```
ph: Do we really need to do this here?
We are already settings up the environment in the `bundler.rb` file? Is the environment correctly configured at that point?

spec/plugin_manager/install_spec.rb
```diff
@@ -0,0 +1,32 @@
+# encoding: utf-8
+require 'spec_helper'
+require 'pluginmanager/main'
+
+describe LogStash::PluginManager::Install do
+
```
ph: remove space?

spec/plugin_manager/install_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:cmd) { LogStash::PluginManager::Install.new("install") }
+
+  before(:each) do
+    expect(cmd).to receive(:validate_cli_options!).and_return(nil)
+  end
+
+
+  context "when validating plugins" do
+
```
ph: remove space?

spec/plugin_manager/update_spec.rb
```diff
@@ -0,0 +1,42 @@
+# encoding: utf-8
+require 'spec_helper'
+require 'pluginmanager/main'
+
+describe LogStash::PluginManager::Update do
+
```
ph: remove space?

spec/plugin_manager/update_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  it "pass all gem sources to the bundle update command" do
+    sources = cmd.gemfile.gemset.sources
+    expect_any_instance_of(LogStash::Bundler).to receive(:invoke!).with(:update => [], :rubygems_source => sources)
+    cmd.execute
+  end
+
+  context "when skipping validation" do
+
```
ph: remove space?

lib/pluginmanager/update.rb
purbon: But bundler it actually does, https://github.com/elastic/logstash/blob/master/lib/bootstrap/bundler.rb#L83, I need more clarification to understand your request :smile:

spec/plugin_manager/util_spec.rb
```diff
@@ -0,0 +1,37 @@
+#encoding: utf-8
+require 'spec_helper'
+require 'pluginmanager/util'
+
+describe LogStash::PluginManager do
+
```
ph: remove space?

spec/plugin_manager/util_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  describe "a logstash_plugin validation" do
+
```
ph: remove space?

spec/plugin_manager/util_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(subject).to receive(:plugin_file?).and_return(false)
+      expect(Gem::Dependency).to receive(:new).and_return(dep)
+      expect(Gem::SpecFetcher).to receive(:fetcher).and_return(fetcher)
+
+      subject.logstash_plugin?(plugin, version, options)
+      expect(Gem.sources.map { |source| source }).to eq(sources)
+    end
+
+  end
+
```
ph: remove space?

spec/plugin_manager/update_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect_any_instance_of(LogStash::Bundler).to receive(:invoke!).with(:update => [plugin], :rubygems_source => sources).and_return(nil)
+    end
+
+    it "skips version verification when ask for it" do
+      cmd.verify = false
+      expect(cmd).to_not receive(:validate_major_versions)
+      cmd.execute
+    end
+  end
+
```
ph: remove space?

spec/plugin_manager/install_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      cmd.verify = true
+    end
+
+    it "should load all the sources defined in the Gemfile" do
+      expect(cmd.gemfile.gemset).to receive(:sources).and_return(sources)
+      expect(LogStash::PluginManager).to receive(:logstash_plugin?).with("dummy", nil, {:rubygems_source => sources}).and_return(true)
+      cmd.execute
+    end
+  end
+
```
ph: remove space?

spec/plugin_manager/install_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+describe LogStash::PluginManager::Install do
+
+  let(:cmd) { LogStash::PluginManager::Install.new("install") }
+
+  before(:each) do
+    expect(cmd).to receive(:validate_cli_options!).and_return(nil)
+  end
+
+
```
ph: remove one space?

lib/pluginmanager/util.rb
```diff
@@ -1,12 +1,16 @@
 require "rubygems/package"
 
 module LogStash::PluginManager
+
```
ph: remove space?

lib/pluginmanager/update.rb
ph: After looking more closely at the previous version of the code and your changes I understand why you pass the sources around, which we didn't to do it.

lib/pluginmanager/util.rb
ph: I really hate class methods, they are a bad patterns.

```
m = Logstash::BundlerManager.new(gemfile)
m.install("logstash-output-elasticsearch", version)
```
```

File: spec/plugin_manager/util_spec.rb [hunk 27]
```diff
@@ -0,0 +1,37 @@
...
+    it "should load all available sources" do
+      expect(subject).to receive(:plugin_file?).and_return(false)
+      expect(Gem::Dependency).to receive(:new).and_return(dep)
+      expect(Gem::SpecFetcher).to receive(:fetcher).and_return(fetcher)
+
+      subject.logstash_plugin?(plugin, version, options)
+      expect(Gem.sources.map { |source| source }).to eq(sources)
+    end
+
```
ph: remove space?

lib/pluginmanager/util.rb
ph: ^ this could wrap the current classes into a more sane api.

ph:
## Tested scenario

### With Internet on.
- Pushed a valid gem to gemfury.com (I have cloned the logstash-input-sqs and changed the name of the gem)
- Adding the source to the gemfile.
- bin/plugin install logstash-input-bleh is failling with `ERROR: Installation aborted, verification failed for logstash-input-bleh` Which I believe should be fine.
- Adding `--no-verify` make the installation work.

ph:
### Tested scenario
#### with internet on
- Pushed a valid gem to gemfury.com  with no dependencies.
Adding the source to the gemfile and removing the default.
bin/plugin install logstash-input-bleh is failling with ERROR: Installation aborted, verification failed for logstash-input-bleh Which I believe should be fine. (unless gemfury changes the metadata)
Adding --no-verify make the installation work.

ph:
### Tested scenario

#### Install With internet on

- Pushed a valid gem and all the dependencies to gemfury.com. Adding the source to the gemfile and removing the default.
- bin/plugin install logstash-input-bleh is failling with ERROR: Installation aborted, verification failed for logstash-input-bleh Which I believe should be fine. (unless gemfury changes the metadata)
- Adding --no-verify make the installation work.

ph:
### Tested scenario

#### Update With internet on
I have pushed a new version of the gem on gemfury and I tried to update my installation.

Doesnt work:

```
~/e/logstash git:pr/3583 ❯❯❯ bin/plugin update                                                                                                                                                                                                                                                                                                                        ✱ ◼
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: logstash-core
ERROR: Updated Aborted, message: undefined method `split' for nil:NilClass
```

```
~/e/logstash git:pr/3583 ❯❯❯ bin/plugin update  logstash-input-bleh                                                                                                                                                                                                                                                                                                 ⏎ ✱ ◼
Update is not supported for manually defined plugins or local .gem plugin installations, skipping:
ERROR: Updated Aborted, message: undefined method `split' for nil:NilClass
~/e/logstash git:pr/3583 ❯❯❯
```


stacktrace:
```
~/e/logstash git:pr/3583 ❯❯❯ DEBUG=1 bin/plugin update  logstash-input-bleh                                                                                                                                                                                                                                                                                         ⏎ ✱ ◼
DEBUG: exec /Users/ph/es/logstash/vendor/jruby/bin/jruby --1.9 -J-XX:+UseParNewGC -J-XX:+UseConcMarkSweepGC -J-Djava.awt.headless=true -J-XX:CMSInitiatingOccupancyFraction=75 -J-XX:+UseCMSInitiatingOccupancyOnly -J-Xmx500m /Users/ph/es/logstash/lib/pluginmanager/main.rb update logstash-input-bleh
Update is not supported for manually defined plugins or local .gem plugin installations, skipping:
NoMethodError: undefined method `split' for nil:NilClass
  validate_major_version at /Users/ph/es/logstash/lib/pluginmanager/update.rb:70
            update_gems! at /Users/ph/es/logstash/lib/pluginmanager/update.rb:46
                  select at org/jruby/RubyArray.java:2468
            update_gems! at /Users/ph/es/logstash/lib/pluginmanager/update.rb:43
                 execute at /Users/ph/es/logstash/lib/pluginmanager/update.rb:24
                     run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67
                 execute at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/subcommand/execution.rb:11
                     run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67
                     run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:132
                  (root) at /Users/ph/es/logstash/lib/pluginmanager/main.rb:36
```

purbon:
@ph interesting, never saw this error myself with internet. BTW, did you add the new source to the gemfile? or are you only relying on rubygems? keep in mind the validate_version uses a rubygems specific api. so you should skip verify there.

Please report how your gemfile looks like, thanks!

- cheers

purbon:
@ph on the other side, your error complains from the part of the code https://github.com/elastic/logstash/blob/master/lib/pluginmanager/util.rb#L58, can you let me know how the version looks like? 

So need to know, your Gemfile + the *.gemspec of your gemfile, to help me debug this. 

Thanks a lot,

ph:
after explicitely used the `--no-verify` flag on the update it worked. But we should still guard agains't a `NIL` error

purbon:
@ph makes sense, because this flag avoid the validation, but I'm interested to know why It failed in your example. Please share:

* How your Gemfile looks?
* The gemspec of your test gem.
* If you are able, what is the outcome of ```Gems.versions(plugin_name)``` vs this server you just used.

Thanks a lot man!

- purbon

ph:
@purbon I saw your comments before no need to repeat it 3 times :)

ph:
The Gemfile https://gist.github.com/6c8c55c49e19cb7943ae before running the bin/plugin update.

ph:
https://gist.github.com/31efb075b8394bef05ef

ph:
# Tested 

Context: Using the `gem server` command that servers all the local gem.

- With verify on
- install worked.
- Update worked

lib/pluginmanager/util.rb
purbon: I like this idea, this would be a nice thing for a feature PR.

ph:
I suggest you go through all the scenarios with gemfury and with the `gem server`, gemfury is easy to setup.

purbon:
@ph I did my test with geminabox, covering all this scenario.  But I introduced the ``` --no-verify``` to avoid this situation when the server is not rubygems. Certainly I did not take into very important issue  people not setting this flag and trying to hit a custom server, but I will provide a better solution for this use case as it makes sense.

purbon:
please jenkins, test this.

purbon:
I think https://github.com/elastic/logstash/pull/3583#issuecomment-137468446 the problem is with the delivery of the gemspec by gemfury, the installation with geminabox (with proper gemspec delivery) works as expected.

```
purbon-elastic% bin/plugin install --version 1.0.5 logstash-output-elasticsearch
Validating logstash-output-elasticsearch-1.0.5
Installing logstash-output-elasticsearch
Installation successful

purbon-elastic% head -n 10 Gemfile
# This is a Logstash generated Gemfile.

source "http://localhost:9292"
gem "logstash-core", "2.0.0.dev", :path => "."
gem "file-dependencies", "0.1.6"
gem "ci_reporter_rspec", "1.0.0", :group => :development
gem "simplecov", :group => :development
```

ph:
Like I have said in my test,  **I believe gemfury is messing with the gemspec..** they are probably not sending the metadata and our verify is based on that.

purbon:
@ph agree! I'm just confirming this as I'm sure my geminabox deliver the gemspec properly :+1:.

ph:
@guyboertje `gem` also takes a source this changes was in master.

guyboertje:
@colin @ph - thats cool.  we are only using one source at a time.  One thought though, is `@gemset.sources` pre-populated with `rubygems.org`? Should we do `@gemset.sources.clear` before the append?

purbon:
@guyboertje you always have rubygems in there, even if not declared/present in the gemfile. This is actually not a source of problems for now. The issue we notice is not having the local source (private gem repo), that for now we add manually in the gemfile and is loaded by the file @colinsurprenant linked to you.

Does it makes sense?

purbon:
@ph with the last changes, https://github.com/elastic/logstash/pull/3583#issuecomment-137481639, should not happen. This was happening because of ```Gems.sources``` failing, I introduced a nice error check for this. 

I was also thinking on skipping the validation automatically when there is no rubygems.org declared as a source, but I went for a more explicit usage of ```--no-verify```  to avoid this check so people are aware of one common pattern of doing things. What do you think?

/cheers

richkettle:
@guyboertje you have posted a private url for sidekiq pro and gemfury, please remove.

guyboertje:
@richkettle - my apologies, removed.

lib/pluginmanager/util.rb
```diff
@@ -46,6 +50,7 @@ def self.fetch_latest_version_info(plugin, options={})
     require "gems"
     exclude_prereleases =  options.fetch(:pre, false)
     versions = Gems.versions(plugin)
+    raise Exception.new("Something went wrong with the validation. You can skip the validation with the --no-verify option") if !versions.is_a?(Array) || versions.empty?
```
ph: Create a custom exception for this.

ph:
jenkins test this please

lib/pluginmanager/install.rb
```diff
@@ -45,9 +45,11 @@ def validate_cli_options!
...
     gems.each do |plugin, version|
       puts("Validating #{[plugin, version].compact.join("-")}")
-      signal_error("Installation aborted, verification failed for #{plugin} #{version}") unless LogStash::PluginManager.logstash_plugin?(plugin, version)
+      next if LogStash::PluginManager.logstash_plugin?(plugin, version, options)
+      signal_error("Installation aborted, verification failed for #{plugin} #{version}")
     end
```
ph: @purbon Can we wrap the call to logstash_plugins? with a begin / rescue clause? 
When I run `bin/plugin install my-plugin` without internet access and without `--no-verify` I get an exception. 

I suppose that if the user run his own gem server he may not have access to the internet.
I think we should provide a better error and not a stacktrace what do you think?


```
/e/logstash git:pr/3583 ❯❯❯ bin/plugin install logstash-mixin-rabbitmq_connection --no-verify                                                                                                                                                                                                                                                                          ⏎ ✱
Validating logstash-mixin-rabbitmq_connection
SocketError: recv: name or service not known
             recv_reply at /Users/ph/es/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:782
                request at /Users/ph/es/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:664
          each_resource at /Users/ph/es/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:512
                 resolv at /Users/ph/es/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:0
                 resolv at /Users/ph/es/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:1034
                 resolv at /Users/ph/es/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:1033
                 resolv at /Users/ph/es/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:1031
          each_resource at /Users/ph/es/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:503
            getresource at /Users/ph/es/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:480
           api_endpoint at /Users/ph/es/logstash/vendor/jruby/lib/ruby/shared/rubygems/remote_fetcher.rb:92
                api_uri at /Users/ph/es/logstash/vendor/jruby/lib/ruby/shared/rubygems/source.rb:46
             load_specs at /Users/ph/es/logstash/vendor/jruby/lib/ruby/shared/rubygems/source.rb:182
             tuples_for at /Users/ph/es/logstash/vendor/jruby/lib/ruby/shared/rubygems/spec_fetcher.rb:261
        available_specs at /Users/ph/es/logstash/vendor/jruby/lib/ruby/shared/rubygems/spec_fetcher.rb:226
            each_source at /Users/ph/es/logstash/vendor/jruby/lib/ruby/shared/rubygems/source_list.rb:97
        available_specs at /Users/ph/es/logstash/vendor/jruby/lib/ruby/shared/rubygems/spec_fetcher.rb:222
  search_for_dependency at /Users/ph/es/logstash/vendor/jruby/lib/ruby/shared/rubygems/spec_fetcher.rb:102
    spec_for_dependency at /Users/ph/es/logstash/vendor/jruby/lib/ruby/shared/rubygems/spec_fetcher.rb:166
       logstash_plugin? at /Users/ph/es/logstash/lib/pluginmanager/util.rb:25
         verify_remote! at /Users/ph/es/logstash/lib/pluginmanager/install.rb:51
         verify_remote! at /Users/ph/es/logstash/lib/pluginmanager/install.rb:49
                execute at /Users/ph/es/logstash/lib/pluginmanager/install.rb:26
                    run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67
                execute at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/subcommand/execution.rb:11
                    run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67
                    run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:132
                 (root) at /Users/ph/es/logstash/lib/pluginmanager/main.rb:37
```

ph:
Added 2 minor comments address them and we can merge this!

lib/pluginmanager/install.rb
purbon: I agree with providing better error handling, however the error you get is not from validation.

(with connection)

```
skywalker% bin/plugin install logstash-mixin-rabbitmq_connection --no-verify
Validating logstash-mixin-rabbitmq_connection
logstash-mixin-rabbitmq_connection is not a Logstash plugin
ERROR: Installation aborted, verification failed for logstash-mixin-rabbitmq_connection 
```

(without internet)

```
skywalker% bin/plugin install logstash-mixin-rabbitmq_connection --no-verify
Validating logstash-mixin-rabbitmq_connection
SocketError: recv: name or service not known
             recv_reply at /Users/purbon/work/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:782
                request at /Users/purbon/work/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:664
          each_resource at /Users/purbon/work/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:512
                 resolv at /Users/purbon/work/logstash/vendor/jruby/lib/ruby/1.9/resolv.rb:0
```

reason is, I guess, the --no-verify appended at the end gets ignored with the mixin name you provide as example.

in the case of passing by a plugin, this is the error you get when installation fail:

```
skywalker% bin/plugin install --no-verify logstash-input-file
Installing logstash-input-file
Error Bundler::HTTPError, retrying 1/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 2/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 3/10
```

passing the right order of argument you get the same error (without internet)

```
skywalker% bin/plugin install --no-verify logstash-mixin-rabbitmq_connection
Installing logstash-mixin-rabbitmq_connection
Error Bundler::HTTPError, retrying 1/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 2/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 3/10
Could not fetch specs from https://rubygems.org/
```

investigating where to actually add the better error handling.

purbon:
@ph added code to amend you two concerns from yesterday, please let me know if you have any other concern before we can merge this :+1: thanks a lot for your review.

ph:
LGTM, tests pass

purbon:
@colinsurprenant after @ph gave his LGTM, what do you think? please let me know your thought about this PR, looking forward to get a review.

lib/pluginmanager/util.rb
colinsurprenant: @ph agree, we are due for a 3rd refactor here. Do we have an issue for this idea?

colinsurprenant:
I believe so-far-so-good, my only concern at this point is that I would like to have a final round of manual tests as they were documented by @ph after this is rebased. Also, does this pass the current test-kitchen tests?

purbon:
@colinsurprenant should I rebase now and you or @ph can do again more tests?

purbon:
@colinsurprenant tests as your requested.

Change the Gemfile source like:

```
-source "https://rubygems.org"
+source "http://localhost:9292/"
 gem "logstash-core", "3.0.0.dev", :path => "."
 gem "file-dependencies", "0.1.6"
 gem "ci_reporter_rspec", "1.0.0", :group => :development
@@ -15,3 +15,4 @@ gem "stud", "~> 0.0.21", :group => :build
 gem "fpm", "~> 1.3.3", :group => :build
```

Internet=ON, connection to rubygems

Install plugin

```
skywalker% bin/plugin install logstash-input-file
Validating logstash-input-file
Installing logstash-input-file
Installation successful
```
update plugin

```
skywalker% bin/plugin install --version 2.1.0 logstash-output-elasticsearch
Validating logstash-output-elasticsearch-2.1.0
Installing logstash-output-elasticsearch
Installation successful
skywalker% bin/plugin update logstash-output-elasticsearch
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: 
Updating logstash-output-elasticsearch
Updated logstash-output-elasticsearch 2.1.0 to 2.1.2
```

Internet=OFF, no connection to rubygems.

Install plugin

```
skywalker% bin/plugin install logstash-input-file
Validating logstash-input-file
Installing logstash-input-file
Installation successful
```

update plugin

```
skywalker% bin/plugin install --version 2.1.0 logstash-output-elasticsearch
Validating logstash-output-elasticsearch-2.1.0
Installing logstash-output-elasticsearch
Installation successful
skywalker% bin/plugin update logstash-output-elasticsearch
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: 
Updating logstash-output-elasticsearch
Updated logstash-output-elasticsearch 2.1.0 to 2.1.2
```

ph:
I ran this branch with https://github.com/elastic/logstash/pull/4166

All green! :heart: 
```
~/e/logstash git:pr/3583 ❯❯❯ rake test:integration:local                                                                                                                                                                                                                                                                                                              ⏎ ✱
[integration_spec] configuring local environment for running test in /Users/ph/es/logstash/integration_run, if you want to change this behavior delete the directory.
[integration_spec] configure environment
[integration_spec] Rsync source code into: /Users/ph/es/logstash/integration_run
[integration_spec] Finish rsync
[integration_spec] Running the test in /Users/ph/es/logstash/integration_run/logstash
[integration_spec] Running specs

update
  update all the plugins
    has executed succesfully
  update a specific plugin
    has executed succesfully

bin/plugin list
  with a specific plugin
    list the plugin with his version
    list the plugin and display the plugin name
  without a specific plugin
    display a list of installed plugins
    list the plugins with their versions
    display a list of plugins

File input to File output
  writes events to file

bin/logstash
  returns the logstash version

bin/plugin uninstall
  when the plugin is installed
    succesfully uninstall it
  when the plugin isn't installed
    fails to uninstall it

bin/plugin install
  when the plugin exist
    allow to install a specific version
    sucessfully install
  when the plugin doesn't exist
    fails to install
  with a local gem
    install the gem succesfully

Finished in 4 minutes 7.7 seconds (files took 3.71 seconds to load)
15 examples, 0 failures

Randomized with seed 7613
```

purbon:
@ph good, good, you see! :green_heart: :green_apple:

colinsurprenant:
LGTM :shipit: 
@ph I suggest you create an issue and dump your refactor thoughts [as discussed](https://github.com/elastic/logstash/pull/3583/files#r44023533).

purbon:
@colinsurprenant thanks for your review, will squash and start the merge process.

antoni12753 (commented):
Mio

Issue #3580, opened 2015-07-09
Add back --pluginpath command line option


suyograo:
In 1.5.0 plugins went through a big revamp, separating it into individual gems. The de facto way of developing custom plugins was to create ruby gems and installing it on LS using the following ways:
1. Publish to RubyGems.org and `bin/plugin install <my_new_plugin>` 
2. Building the gem locally and installing using `bin/plugin install <my_new_plugin.gem>`
3. Changing the `gemspec` to [point](https://github.com/logstash-plugins/logstash-output-elasticsearch#2-running-your-unpublished-plugin-in-logstash) to a github repository remotely or to a cloned copy on disk (using `:git =>` or `:path => plugin_path` ) 

We need to add back `bin/plugin --pluginpath` to allow users to point to ruby files which contain custom plugin code. This will allow for rapid prototyping and experimenting with new plugins

Issues:

- [ ] custom plugins not recognized in RC4 https://github.com/elastic/logstash/issues/3227
- [ ] Support for using custom plugins using --pluginpath https://github.com/elastic/logstash/issues/3553

Discuss:

https://discuss.elastic.co/t/logstash-new-plugin-system-too-restrictive/24247

suyograo:
Related: https://github.com/elastic/logstash/issues/2423

colinsurprenant:
fixed by #3587

PR #3587, opened 2015-07-09
support for bare/ungemified local plugins using --pluginpath option

colinsurprenant:
restore the `--pluginpath` option to support bare/ungemified local plugins

solves #3580

### DIFF
--- a/spec/core/environment_spec.rb
+++ b/spec/core/environment_spec.rb
@@ -38,7 +38,18 @@ describe LogStash::Environment do
         allow(Dir).to receive(:glob).and_return([])
         expect { subject.load_runtime_jars! }.to raise_error
       end
+    end
+  end
 
+  context "add_plugin_path" do
+    let(:path) { "/some/path" }
+
+    before(:each) { expect($LOAD_PATH).to_not include(path) }
+    after(:each) { $LOAD_PATH.delete(path) }
+
+    it "should add the path to $LOAD_PATH" do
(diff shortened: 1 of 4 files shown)


jordansissel:
code LGTM; haven't tested

purbon:
I would add a test for the ```agent. configure_plugin_paths``` in the same fashion as the one already created.

spec/core/environment_spec.rb
```diff
@@ -38,7 +38,16 @@
         allow(Dir).to receive(:glob).and_return([])
         expect { subject.load_runtime_jars! }.to raise_error
       end
+    end
+  end
+
+  context "plugin paths" do
```
purbon: In https://github.com/colinsurprenant/logstash/pull/1 I proposed a way to improve/organize a bit better the code in this test. It does not means this test are wrong, but the way of writing this I think i could be improved. Let me know what do you think?.

colinsurprenant:
@purbon agree, I will. For some reason I thought no spec existed for the `Agent` class but that's not true. In any case, we should always take the opportunity to add specs for all new code we create or bug we fix. +1

colinsurprenant:
ready for more review of added code please.

purbon:
LGTM, tests pass.

colinsurprenant:
thanks! :shipit:

colinsurprenant:
rebasing and merging shortly.

PR #3593, opened 2015-07-10
Fix namespacing for unicode trimmer

andrewvc:
This fixes an issue where the `UnicodeTrimmer` used the root namespace.


suyograo:
LGTM

PR #3603, opened 2015-07-13
Update lock file to use 1.5.3.snapshot2

suyograo:


Gemfile.jruby-1.9.lock
```diff
@@ -17,7 +17,7 @@ GEM
     aws-sdk-v1 (1.64.0)
       json (~> 1.4)
       nokogiri (>= 1.4.4)
-    backports (3.6.4)
+    backports (3.6.5)
```
jordansissel: intentional change? It's probably harmless. Also I love the backports gem. :)

jordansissel:
LGTM

Gemfile.jruby-1.9.lock
suyograo: not sure when this was added. I did clean my local .bundle dir and Gemfile.jruby-1.9.lock

PR #3608, opened 2015-07-14
Add new section for plugin manager CLI

suyograo:


untergeek:
Don't we need an include in the index.asciidoc?

untergeek:
And, does it build?

suyograo:
@untergeek the index is maintained here: https://github.com/elastic/logstash-docs/blob/master/docs/index.asciidoc. Will make a PR there

PR #3614, opened 2015-07-15
Use echo instead of trace or emit.

jordansissel:
This should fix a test failure on CentOS 6 where 'emit' gives 'command
not found' which, as the last command run in stop(), causes stop() to
always return exit code 127.


jordansissel:
Note to self: also need to patch this into pleaserun also

suyograo:
LGTM

suyograo:
merging to fix tests

PR #3629, opened 2015-07-20
Use latest ES version for correct jars

suyograo:


suyograo:
the benchmark jar update is harmless

talevy:
famous last words

talevy:
LGTM

PR #3634, opened 2015-07-21
Restore rack dependency because http input needs it

suyograo:
See https://github.com/logstash-plugins/logstash-input-http/pull/15


jordansissel:
LGTM, didn't test.

PR #3636, opened 2015-07-21
added case insensitive option to regex

nbvehrfr:
Hi

I added case insensitive option to regex, so you can use it as /MLB/i

Tim


jordansissel:
Thanks for helping improve logstash!

Code looks ok. 

I'm not sure how this change fits into our plans for a management api, though, and I am hesitant to add further extensions to the config language until we figure out the path from today's config file to the future's management api.

As a workaround until we figure things out, folks can use the `(?i)` flag within the regexp to have the same effect: `/(?i)this Is Case Insensitve/`

Issue #3657, opened 2015-07-27
Logstash 1.5.3 SSL problems


jordansissel:
Something is funky with Logstash 1.5.3's SSL. I'm not sure what, yet.

Tests -

With a self-signed CN=localhost certificate on java 1.7.0_79
* :white_check_mark: LSF 0.4.0 to LS 1.5.3
* :x: LS 1.5.3 to LS 1.5.3  `certificate verify failure`
* :x: LS 1.5.3 to LS 1.5.2  `certificate verify failure`
* :x: LS 1.5.2 to LS 1.5.2 `server key exchange invalid`
* :x: LS 1.4.3 to LS 1.5.2 `Socket closed`
* :x: openssl s_client to LS 1.5.2: `RSA_EAY_PUBLIC_DECRYPT:data too large for modulus / SSL routines:ssl3_get_key_exchange:bad signature`

Related tickets
* https://github.com/elastic/logstash/issues/3640
* https://github.com/elastic/logstash/issues/3650
* https://github.com/elastic/logstash/issues/3644

jordansissel:
Getting all kinds of weird failures even with Logstash 1.5.2...

ph:
Do you have the same issue with Java 1.8?

jordansissel:
@ph it's on my todo list to try.

jordansissel:
Same problem on Java 1.8.0_60-internal

jordansissel:
@ph can you help me test these scenarios (LSF 0.4.0 to LS 1.5.3, LS 1.5.2 to LS 1.5.3, etc)? I am still trying to figure out what's wrong with my workstation.

jordansissel:
I also tested in an Ubuntu 14.04 docker container. Same errors. This is bizarre. Will try on a different system tomorrow.

timukas:
Here's my setup: java 1.8.0_45, LSF 0.4.0, ubuntu 14.04 server 64bit.
When i run LS 1.5.3 in debug mode i see this warnings:

Starting lumberjack input listener {:address=>"172.xx.xx.5:6xx2", :level=>:info, :file=>"logstash/inputs/lumberjack.rb", :line=>"50", :method=>"register"}
[2015-07-28 09:52:15.967]  WARN -- Concurrent::Condition: [DEPRECATED] Will be replaced with Synchronization::Object in v1.0.
[...]

PR #3667, opened 2015-07-29
Do not force a `verify_mode` with a our stronger ssl settings

ph:
We have discovered that in some cases and some plaftorms
configuring a default `verify_mode` when creating a SSL/TCPServer
could make the certificate verification fail. Ruby default behavior is
to use `NIL` when creating a new ssl context, this revert that change.

keep in mind that all TCP clients using SSL **must** use `VERIFY_PEER`
as their verify mode to prevent man in the middle attack.

Fix: https://github.com/elastic/logstash/issues/3657

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,4 +15,4 @@ gem "stud", "~> 0.0.19", :group => :build
 gem "fpm", "~> 1.3.3", :group => :build
 gem "rubyzip", "~> 1.1.7", :group => :build
 gem "gems", "~> 0.8.3", :group => :build
-gem "flores", "~> 0.0.4", :group => :development
+gem "flores", "~> 0.0.6", :group => :development
--- a/Gemfile.jruby-1.9.lock
+++ b/Gemfile.jruby-1.9.lock
@@ -49,7 +49,7 @@ GEM
     file-dependencies (0.1.6)
       minitar
     filesize (0.0.4)
-    flores (0.0.4)
+    flores (0.0.6)
     fpm (1.3.3)
       arr-pm (~> 0.0.9)
       backports (>= 2.6.2)
@@ -133,7 +133,7 @@ DEPENDENCIES
   ci_reporter_rspec (= 1.0.0)
   coveralls
   file-dependencies (= 0.1.6)
-  flores (~> 0.0.4)
+  flores (~> 0.0.6)
   fpm (~> 1.3.3)
   gems (~> 0.8.3)
   logstash-core (= 2.0.0.dev)!
--- a/lib/logstash/patches/stronger_openssl_defaults.rb
+++ b/lib/logstash/patches/stronger_openssl_defaults.rb
@@ -1,4 +1,3 @@
-
 require "openssl"
 
 # :nodoc:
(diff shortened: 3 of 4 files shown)


ph:
@jordansissel I took your test from `Flores/PKI` since it was really well written, I have added a reference.

ph:
@jsvd @purbon Can you test this PR? Make sure you read the context in https://github.com/elastic/logstash/issues/3657 and you are able to reproduce this problem.

ph:
@suyograo ^

jordansissel:
Code LGTM. Let's get a few folks to validate this tests passes on their systems.

purbon:
@ph tested this changes in my system (os x) by:

* cloning the PR.
* rake bootstrap
* rake test:install-core
* rake test:core

And got this error:

```
Failures:

  1) OpenSSL defaults SSLSocket with self-signed client/server certificate should successfully connect as a client
     Failure/Error: expect { ssl_client.connect }.not_to raise_error
       expected no Exception, got #<OpenSSL::SSL::SSLError: java.io.IOException: Invalid encoding of AuthorityKeyIdentifierExtension.> with backtrace:
         # ./spec/logstash/patches_spec.rb:86:in `(root)'
         # ./spec/logstash/patches_spec.rb:86:in `(root)'
         # ./lib/bootstrap/rspec.rb:10:in `(root)'
     # ./spec/logstash/patches_spec.rb:86:in `(root)'
     # ./lib/bootstrap/rspec.rb:10:in `(root)'

Finished in 3.96 seconds (files took 1.19 seconds to load)
1126 examples, 1 failure

Failed examples:

rspec ./spec/logstash/patches_spec.rb:83 # OpenSSL defaults SSLSocket with self-signed client/server certificate should successfully connect as a client
```

does it makes sense?

ph:
Yeah, forgot to update flores in the `lock` file, same errors on jenkins, I should have been more careful sorry for time wasted.

ph:
@purbon by tested we also need to do an actual LSF test to see that the behavior is correctly fixed.

purbon:
no problem! cool!, what do you mean by actual LSF test? run test in LSF?

ph:
Try to reproduce the bug described in https://github.com/elastic/logstash/issues/3657 first.
Use logstash 1.5.3 and LSF 0.4.0, LSF should not be able to send events to logstash, then try the patched version and it should work.

@jordansissel and I had issues reproducing it.

purbon:
@ph cool! will try. thanks for clarifying. 

@jordansissel @ph we might like to introduce a regression test for this?

ph:
@purbon we have added a test that for that in this PR and I have added https://github.com/elastic/logstash-kitchen/issues/6

purbon:
cool! perfect man! good work!

ph:
Adding 1.4.5 label, we need to backport it.

jsvd:
Does this mean that if we want the ruby-lumberjack server (or logstash-input-lumberjack) to verify client certs this will still not work, right?
I guess the plugin doesn't expose verify mode [in its options](https://github.com/logstash-plugins/logstash-input-lumberjack/blob/master/lib/logstash/inputs/lumberjack.rb#L15-L37)

ph:
@jsvd correct, this is why from your comments on my PR, I am looking at this how complicated to add it to the actual library. We have some Discrepancies between the configuration we require in LSF, output-lumberjack and what we actually validate.

purbon:
tested this manually and LGTM. Patched version works.

my env is:

OS: os x.
JVM: java version "1.7.0_71"

untergeek:
Manually applied the patch to my "production" (read: home use) Logstash by deleting the offending line:

```
    :verify_mode => OpenSSL::SSL::VERIFY_PEER, 
```

LSF immediately connected and began spewing data, so this LGTM™

jordansissel:
I think we have enough successful testers now. +1 on merging. LGTM

ph:
Merged in 1.5, 1.4 and 1.6

Issue #3677, opened 2015-08-02
logstash 1.5.3 - java version checking - how to allow a known good version?


dougmcclure:
I've reviewed issue #2547 and get what's trying to be achieved.  Have we thought of an easy way to allow a java version that's known to be good?  Can this check be disabled, updated with a known pattern for something like IBM Java in my case?

bin/logstash agent -f foo.conf --pluginpath MyPlugins/ --verbose

NoMethodError: undefined method `captures' for nil:NilClass
        parse_java_version at /opt/logstash153/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.3-java/lib/logstash/util/java_version.rb:33
         bad_java_version? at /opt/logstash153/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.3-java/lib/logstash/util/java_version.rb:51
  warn_on_bad_java_version at /opt/logstash153/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.3-java/lib/logstash/util/java_version.rb:10
                      main at /opt/logstash153/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.3-java/lib/logstash/runner.rb:32
                    (root) at /opt/logstash153/lib/bootstrap/environment.rb:48

This is what I get from IBM Java. Maybe we can add a simple "allow IBM Java" if IBM J9 VM is detected or inspect more closely and look for "based on Oracle 7u51-b11".

java -version

java version "1.7.0"
Java(TM) SE Runtime Environment (build pxa6470sr6fp1-20140108_01(SR6 FP1))
IBM J9 VM (build 2.6, JRE 1.7.0 Linux amd64-64 Compressed References 20140106_181350 (JIT enabled, AOT enabled)
J9VM - R26_Java726_SR6_20140106_1601_B181350
JIT  - r11.b05_20131003_47443.02
GC   - R26_Java726_SR6_20140106_1601_B181350_CMPRSS
J9CL - 20140106_181350)
JCL - 20140103_01 based on Oracle 7u51-b11

Tks!  A quick work around would be appreciated so I can test LS153 out in my environments.

suyograo:
@andrewvc maybe we can extract the known Java versions to a separate file that can be updated outside of the release artifact? @jordansissel ideas?

andrewvc:
@suyograo I'm inclined to keep that in code for flexibility. It gets tricky to create a definitions file if there's a single bad release for instance. 

[...]

PR #3685, opened 2015-08-03
Handle non-hotspot javas in version check

andrewvc:
Fixes https://github.com/elastic/logstash/issues/3677


ph:
code lgtm, tests are running.

PR #3686, opened 2015-08-03
Update *.lock and uses 1.5.4.snapshot2

ph:


lib/logstash/version.rb
```diff
@@ -1,6 +1,6 @@
 # encoding: utf-8
 # The version of logstash.
-LOGSTASH_VERSION = "1.5.4.snapshot1"
+LOGSTASH_VERSION = "1.5.4.snapshot2"
```
suyograo: nice, closed quotes :)

suyograo:
LGTM

lib/logstash/version.rb
ph: :ribbon:

PR #3718, opened 2015-08-11
Make sure all lookups using event.sprintf are returned as UTF-8

purbon:
The changes in this PR make sure all lookups done using event.sprintf are returned as UTF-8, before this PR, if the event was not there another encoding was used.

### DIFF
--- a/lib/logstash/string_interpolation.rb
+++ b/lib/logstash/string_interpolation.rb
@@ -17,9 +17,9 @@ module LogStash
       if template.is_a?(Float) && (template < MIN_FLOAT_BEFORE_SCI_NOT || template >= MAX_FLOAT_BEFORE_SCI_NOT)
         return ("%.15f" % template).sub(/0*$/,"")
       end
-      
+
       template = template.to_s
-      
+
       return template if not_cachable?(template)
 
       compiled = CACHE.get_or_default(template, nil) || CACHE.put(template, compile_template(template))
@@ -108,11 +108,11 @@ module LogStash
 
       case value
       when nil
-        "%{#{@key}}"
+        "%{#{@key}}".encode(Encoding::UTF_8)
       when Array
-        value.join(",")
+        value.join(",").encode(Encoding::UTF_8)
       when Hash
-        LogStash::Json.dump(value)
+        LogStash::Json.dump(value).encode(Encoding::UTF_8)
       else
         value
       end
--- a/lib/logstash/string_interpolation.rb
+++ b/lib/logstash/string_interpolation.rb
@@ -110,9 +110,9 @@ module LogStash
       when nil
         "%{#{@key}}".encode(Encoding::UTF_8)
       when Array
-        value.join(",").encode(Encoding::UTF_8)
+        value.join(",")
       when Hash
-        LogStash::Json.dump(value).encode(Encoding::UTF_8)
+        LogStash::Json.dump(value)
       else
         value
       end
--- a/lib/logstash/string_interpolation.rb
+++ b/lib/logstash/string_interpolation.rb
@@ -1,3 +1,5 @@
+# encoding: utf-8
+
 require "thread_safe"
 require "forwardable"
 
@@ -108,7 +110,7 @@ module LogStash
 
       case value
       when nil
-        "%{#{@key}}".encode(Encoding::UTF_8)
+        "%{#{@key}}"
       when Array
         value.join(",")
       when Hash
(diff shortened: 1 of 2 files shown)


purbon:
@colinsurprenant would you be so nice to review the changes in here? Thanks a lot for your time.

purbon:
This changes are the general fix for the issues seen at https://github.com/logstash-plugins/logstash-filter-alter/pull/5 .

A detailed stack trace with the error could be seen at http://build-eu-00.elastic.co/view/LS%20Filters/job/logstash_plugin_filter_alter/866/console

colinsurprenant:
@purbon I do not think that systematically force encoding everything is the right approach here. calling `encode` is costly. I think we need to find & understand in what case we see non UTF-8 strings and correct that instead. 

Is this only spesc/tests failings or did we have a "real" logstash usage problem with this? 

Maybe it is just because the `alter_spec.rb` file does not start with `# encoding: utf-8` ?

purbon:
@colinsurprenant I added the encoding in the ```alter_spec.rb``` and did not solve the issue. I understand your concerns. The problem I found is when I lookup for a non existent pattern. Might be the solution is to encode somehow the key to be UTF-8.

lib/logstash/string_interpolation.rb
```diff
@@ -108,7 +108,7 @@ def evaluate(event)
 
       case value
       when nil
-        "%{#{@key}}"
+        "%{#{@key}}".encode(Encoding::UTF_8)
```
jordansissel: This would only not be UTF-8, I think, if the top of this file doesn't have `# encoding: utf-8`, and I see it does not. That'll be the fix :)

We should ensure all our source code has '# encoding: utf-8` at the top of it.

purbon:
good catch, thanks @jordansissel

colinsurprenant:
:thumbsup:

purbon:
@colinsurprenant does this means I can merge it?

purbon:
@jordansissel will open another PR to add this to the source code. Thanks for your comment.

colinsurprenant:
I just verified and I confirm that the `# encoding: utf-8` are required in both `string_interpolation.rb` and `alter_spec.rb` to fix the alter filter specs. 

LGTM to merge this.

purbon:
cool, thanks @colinsurprenant as spoken I will do another pass to be sure this magic string is everywhere :P

jsvd:
I'd take this opportunity to include the encoding top line on all files, here's a list of files that don't have the encoding as the first line of the file :)

```
lib/bootstrap/bundler.rb,module LogStash
lib/bootstrap/environment.rb,# bootstrap.rb contains the minimal code to be able to launch Bundler to eventually be able
lib/bootstrap/rspec.rb,require_relative "environment"
lib/bootstrap/rubygems.rb,module LogStash
lib/logstash/config/grammar.rb,# Autogenerated from a Treetop grammar. Edits may be lost.
lib/logstash/environment.rb,require "logstash/errors"
lib/logstash/java_integration.rb,require "java"
lib/logstash/patches/bugfix_jruby_2558.rb,require "logstash/environment"
lib/logstash/patches/bundler.rb,# Bundler monkey patches
lib/logstash/patches/cabin.rb,if ENV["PROFILE_BAD_LOG_CALLS"] || ($DEBUGLIST || []).include?("log")
lib/logstash/patches/rubygems.rb,# monkey patch RubyGems to silence ffi warnings:
lib/logstash/patches/stronger_openssl_defaults.rb,require "openssl"
lib/logstash/patches.rb,require "logstash/patches/bugfix_jruby_2558"
lib/logstash/string_interpolation.rb,require "thread_safe"
lib/logstash/util/filetools.rb,require "net/http"
lib/logstash/util/java_version.rb,require 'cabin'
lib/logstash/util/plugin_version.rb,require 'logstash/errors'
lib/logstash/util/reporter.rb,class InflightEventsReporter
lib/logstash/util/retryable.rb,module LogStash
lib/logstash/util/unicode_trimmer.rb,module LogStash::Util::UnicodeTrimmer
lib/logstash-core.rb,module LogStash
lib/pluginmanager/command.rb,class LogStash::PluginManager::Command < Clamp::Command
lib/pluginmanager/gemfile.rb,module LogStash
lib/pluginmanager/install.rb,require "pluginmanager/command"
lib/pluginmanager/list.rb,require 'rubygems/spec_fetcher'
lib/pluginmanager/main.rb,$LOAD_PATH.unshift(File.expand_path(File.join(__FILE__, "..", "..")))
lib/pluginmanager/uninstall.rb,require "pluginmanager/command"
lib/pluginmanager/update.rb,require "pluginmanager/command"
lib/pluginmanager/util.rb,require "rubygems/package"
```

purbon:
@jsvd I will do that in a separate PR.

PR #3723, opened 2015-08-12
Add the missing magic header # encoding: utf-8

purbon:
… so all internal strings are UTF-8 in Ruby < 2.0, this has been found thanks to the work on #3718, we might like to work on internal check for this, probably with the inclusion of rubocop or other kind of changes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,4 @@
+# encoding: utf-8
 
 $: << File.join(File.dirname(__FILE__), "lib")
 
--- a/lib/bootstrap/bundler.rb
+++ b/lib/bootstrap/bundler.rb
@@ -1,3 +1,4 @@
+# encoding: utf-8
 module LogStash
   module Bundler
     extend self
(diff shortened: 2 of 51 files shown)


purbon:
@jsvd or @jordansissel  are you willing to get an eye here, and let me know if this is good to be reviewed ... this is the follow up from #3718 where we discover some missing utf-8 magic headers.

jsvd:
@purbon do you think it's worth modifying the config_ast.rb generator so the file is created with the encoding line on top?

purbon:
We might do it, not sure how to, but I can investigate :panda_face:

purbon:
@jsvd I was looking at the ```config_ast.rb``` and no clue how to make this happen.

jsvd:
The only .rb file without encoding is the generated grammar file `lib/logstash/config/grammar.rb`.
Making treetop generate the encoding header would require monkeypatching treetop or doing a post prepend on the file.
Unless someone sees a big reason to take on this job, we can merge this PR as it is.

This pending question aside, LGTM

PR #3724, opened 2015-08-12
[WIP] Shutdown improvement

jsvd:
This pr takes the code from #3211 and adds a shutdown controller that detects stalled pipeline teardowns and allows forced exits.

- [x] change shutdown sequence to not rely on Thread.raise
- [ ] refactor plugins to accomodate shutdown refactor
- [x] detect stalled teardown and allow for forced exit
- [x] add fourth stage to pipeline to receive dead letter events


jsvd:
The implementation of the shutdown sequence now introduces a DeadLetterPostOffice class that is responsible for handling..dead letters. Dead letters will occur in two scenarios:

1. a plugin decides that an event will never be correct/valid/processed and wants it out of its sight.
2. a stalled pipeline refuses to quit so we need to do something with the inflight events

The API's for this mechanism are:

a) `Plugin#dead_letter(event)` -> all plugin instances can call this to send the event to the post office
b) `DeadLetterPostOffice << event` -> this is the direct way to mail a dead letter

@colinsurprenant could you take a look at this and provide feedback? This isn't merge ready, some of the options needs to be exposed, but both the ShutdownController and DeadLetterPostOffice have specs.

jsvd:
@colinsurprenant considering this PR is starting to grow due to the dead letter feature, it should be split back into your initial PR (refactoring the shutdown sequence) and this one, introducing the dead letter mechanism, thoughts?

purbon:
I think they are two different things, related but different, so splitting might be a good idea.

suyograo:
@jsvd +1 for splitting `dead_letter` out of this...

To summarize from our discussion on hipchat, lets keep in mind that the `dead_letter` work does not become a full-blown feature for 1.6. We should keep it limited to the output stage and to using a file output for now. Basically, this feature should help the shutdown work (in plugins that need it) and showcase usage in ES output (mapping errors) which seem to cause users a lot of pain.

jsvd:
Closing this issue. I'll open 2 new ones, one foccused on shutdown sequence, and the other on the dead_letter mechanism

PR #3731, opened 2015-08-13
fix update plugins list to not include uninstalled but not purged plugins

### LINKED ISSUE #3678: LS > 1.5.0 bin/plugin uninstall not removing gem
```bash
bin/plugin install logstash-output-circonus
bin/plugin uninstall logstash-output-circonus
```

After running these commands, the **logstash-output-circonus** gem is removed from the Gemfile but it is still installed under **vendor/bundle/jruby/1.9/gems**

```
bin/plugin update
```
Plugin not found, aborting
ERROR: Updated Aborted, message: Could not find gem 'logstash-output-circonus'.

When installing uninstalling 2 plugins, it removes the first gem, but not the second one.
```
bin/plugin install logstash-output-circonus
bin/plugin install logstash-output-riemann
bin/plugin uninstall logstash-output-circonus
bin/plugin uninstall logstash-output-riemann
```

After running these commands, both gems are removed from the Gemfile and the **logstash-output-circonus** gem is uninstalled, but the **logstash-output-riemann** gem is still installed under **vendor/bundle/jruby/1.9/gems**.

```
bin/plugin update
```
Plugin not found, aborting
ERROR: Updated Aborted, message: Could not find gem 'logstash-output-riemann'.

purbon:
This PR is the first attempt to provide a closure to #3678 to users are able to clean stale dependencies left behind by the plugin manager.

## Comments
Having a clean command is not the best solution, we should be able to actually let Bundler know we aim to clean from a given environment, actually this only happen to me when another process is kick in so the environment is fresh.

Possible solutions:

- use ```Bundler.with_clean_env``` if possible.
- use new process.
- Remove stale dependencies by code.

After doing some investigation on 1, and 2, the solution was either not working as expected, or having a too complicated setup. I end up implementing 3 + the proposed solution by @colinsurprenant in the review process, so now we've a clean environment after uninstall (my expectation) and update works only with the gems available in the gemfile.

Test has been added to logstash-kitchen project.

Fixes #3678

### DIFF
--- a/lib/pluginmanager/update.rb
+++ b/lib/pluginmanager/update.rb
@@ -107,7 +107,7 @@ class LogStash::PluginManager::Update < LogStash::PluginManager::Command
   # retrieve only the latest spec for all locally installed plugins
   # @return [Hash] result hash {plugin_name.downcase => plugin_spec}
   def find_latest_gem_specs
-    LogStash::PluginManager.find_plugins_gem_specs.inject({}) do |result, spec|
+    LogStash::PluginManager.all_installed_plugins_gem_specs(gemfile).inject({}) do |result, spec|
       previous = result[spec.name.downcase]
       result[spec.name.downcase] = previous ? [previous, spec].max_by{|s| s.version} : spec
       result


colinsurprenant:
I suggest you put the [WIP] state directly at the beginning of the PR subject.

colinsurprenant:
As discussed in #3678 I would prefer we solve the issue instead of providing a manual `clean` command. This will be really confusing for users.

purbon:
Yep, i agree. For now this has been an experiment to show when this works,
nothing else.

/purbon

lib/pluginmanager/clean.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def execute
+    signal_error("File #{LogStash::Environment::GEMFILE_PATH} does not exist or is not writable, aborting") unless File.writable?(LogStash::Environment::GEMFILE_PATH)
+
+    ##
+    # Need to setup the bundler status to enable uninstall of plugins
+    # installed as local_gems, otherwise gem:specification is not
+    # finding the plugins
+    ##
+    LogStash::Bundler.setup!({:without => [:build, :development]})
```
colinsurprenant: this should not be required, the setup is already done in the `invoke!` method. I am guessing you took this from the `uninstall` command, in which case the `setup!` call is required for the `LogStash::PluginManager.installed_plugin?` call.

purbon:
@ph is the best way to add test here to use your logstash-kitchen ?

purbon:
could you also @ph be so kind to review this?

colinsurprenant:
@purbon  before we review, did you verify the case where you uninstall a plugin which is a dependency of another plugin so it does not get physically deleted? I am afraid that the `remove_stale_gems` method might force-delete it as well.

purbon:
@colinsurprenant if the plugin gets throw the uninstall method, actually calling ```gemfile.save``` I understand this is safe to be removed. Does it makes sense? do you have an example I could use to test the behaviour is worrying you?

colinsurprenant:
so, basically a scenario like this: (not tested, but I think that should show it).
- bare logstash install, no plugins
- bin/plugin install logstash-codec-line
- bin/plugin install logstash-input-stdin
- bin/plugin uninstall logstash-codec-line

this last uninstall will trigger the removal of the explicit dependency in the Gemfile, and invoke bundler but the actual gem will never be removed because logstash-codec-line is still an implicit dependency of logstash-input-stdin and bundler knows that through its dependency resolution. 

eventually a `bin/plugin uninstall logstash-input-stdin` should allow the removal of logstash-codec-line. 

I am afraid that your code here will remove logstash-codec-line regardless of its it should **not** be physically removed because it is still a dependency of another gem.

purbon:
@colinsurprenant you're completely right with it, I was expecting bundler to complain here, but is not the case. Thanks for your test case.

colinsurprenant:
@purbon not sure I understand what you mean by «was expecting bundler to complain» but from a bundler perspective, everything is ok, the problem is force-deleting gems which we don't know for-sure if they are still a valid dependency. if we figure how to verify that then we'll be good. 

The lesson here is that we **really** have to be extra careful if we try to directly play with the gems under bundler's control. we should always try to go through bundler, which, granted, is not easy since there is no (clear) programmatic api.

purbon:
Bundler does install, because it has no uninstall command. I somehow expected the bundler implementation we made to detect there is a pending dependency and complain (this was my expectation).

colinsurprenant:
@purbon well, its more a semantic discussion, `bundle install` actually means "process Gemfile", and if some gems have been removed from the Gemfile, bundler will remove them etc etc. 

so the implementation we have, which relies on bundler, actually does detect there is a pending dependency and will not delete any dependancy gems even if we remove them from the Gemfile because bundler knows that with its dependency graph. 

the problem is force-deleting from from the bundler gems directory without being aware of the dependency graph.

colinsurprenant:
To recap:
- this PR now introduces the `clean` command. Personally I am not in favour of introducing this new command. its purpose will be confusing for the users because the real problem here is not that a gem is not properly deleted but because we have a bug in selecting the proper plugins needed for update.
- the `find_plugins_gem_specs` method has been replaced by `PluginManager#all_installed_plugins_gem_specs ` [as discussed](https://github.com/elastic/logstash/issues/3678#issuecomment-130782765) - but as this been **thoroughly tested?** is this passing our plugin manager test suite? we should also add these breaking cases to the suite.

purbon:
@colinsurprenant 
  I can get rid of the ```clean``` command. On the other side, speaking about test, this is a good candidate and a perfect test for the logstash-kitchen test @ph has been working on, don't you think? or you expect to have integration test for the plugin manager build in here? please educate me.

purbon:
@colinsurprenant the clean command has been reverted as you requested.

colinsurprenant:
@purbon yes when I said «plugin manager test suite» I meant the test kitchen test suite for the plugin manager, sorry for the confusion. I also said this should be **thoroughly tested** which mean either installing the test kitchen setup locally and running all tests and adding missing one for this PR condition and/or manually executing all cases to make sure nothing breaks, **before** merging.

purbon:
@ph as we spoke, I need a hand on this from you. Please don't forget me :P

purbon:
A test for this has been added to https://github.com/elastic/logstash-kitchen/pull/11 , this is my first kitchen-ci test, so please let me know if I did all correct there.

colinsurprenant:
changing subject to reflect that this PR will actually fix the update side to correctly build the list of plugins to update and not include uninstalled but not purged plugins.

colinsurprenant:
moved the bundler clean problems into #3767

colinsurprenant:
quick update: on my side, to move this PR forward, the changes has to be manually validated for the different update scenarios and related tests must be added to the test kitchen setup.

purbon:
@colinsurprenant test where added to test-kitchen a few days ago, waiting for validation from @ph

colinsurprenant:
@purbon did you also do thorough manual tests of the different update scenarios with this patch applied?

purbon:
yes I did! I hope I didn't forget any scenario. Please feel free to validate.

colinsurprenant:
@purbon I suggest you copy/paste you test results here then we can see if we think something is missing.

purbon:
ok will do that.

purbon:
here you've the commands requested:

```
purbon-elastic% bin/plugin install logstash-output-circonus
Validating logstash-output-circonus
Installing logstash-output-circonus
Installation successful
purbon-elastic% bin/plugin uninstall logstash-output-circonus
Uninstalling logstash-output-circonus
purbon-elastic% bin/plugin update
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: logstash-core
No plugin updated
```

as noted in #3678 the update command was actually raising an exception, something that does not happen here.

purbon:
Hi, is this ok to be merged?

purbon:
@ph is this PR in a good shape to be merged? thanks a lot or your time man!

ph:
*slip* in the crack, let me test this real quick.

ph:
#### Tested scenario:

```
~/e/logstash git:pr/3731 ❯❯❯ bin/plugin install logstash-output-circonus             ✱
Validating logstash-output-circonus
Installing logstash-output-circonus
Installation successful
~/e/logstash git:pr/3731 ❯❯❯ bin/plugin uninstall logstash-output-circonus           ✱
Uninstalling logstash-output-circonus
~/e/logstash git:pr/3731 ❯❯❯ bin/plugin update                                       ✱
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: logstash-core
Updating logstash-filter-clone, logstash-filter-multiline, logstash-filter-mutate, logstash-input-generator, logstash-input-stdin, logstash-input-tcp, logstash-output-stdout
No plugin updated
~/e/logstash git:pr/3731 ❯❯❯ bin/plugin list circonus                                ✱
ERROR: No plugins found
```

```
~/e/logstash git:pr/3731 ❯❯❯ bin/plugin install --version 1.0.2 logstash-input-lumberjack
Validating logstash-input-lumberjack-1.0.2
Installing logstash-input-lumberjack
Installation successful
~/e/logstash git:pr/3731 ❯❯❯ bin/plugin update                                       ✱
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: logstash-core
Updating logstash-filter-clone, logstash-filter-multiline, logstash-filter-mutate, logstash-input-generator, logstash-input-lumberjack, logstash-input-stdin, logstash-input-tcp, logstash-output-stdout
Updated logstash-input-lumberjack 1.0.2 to 1.0.5
~/e/logstash git:pr/3731 ❯❯❯ bin/plugin install logstash-output-circonus             ✱
Validating logstash-output-circonus
Installing logstash-output-circonus
~/e/logstash git:pr/3731 ❯❯❯ bin/plugin update                                                                                                                                   ✱
Update is not supported for manually defined plugins or local .gem plugin installations, skipping: logstash-core
Updating logstash-filter-clone, logstash-filter-multiline, logstash-filter-mutate, logstash-input-generator, logstash-input-lumberjack, logstash-input-stdin, logstash-input-tcp, logstash-output-circonus, logstash-output-stdout
No plugin updated
~/e/logstash git:pr/3731 ❯❯❯ bin/plugin list --verbose lumberjack                                                                                                                ✱
logstash-input-lumberjack (1.0.5)
```

ph:
Code LGTM, run the core test and did some manual testing. I am happy with it.

The gems are still present in the installation but they don't harm the different use cases since they are not  available in the gemfile or the lock.

I am :+1: to merge this.

PR #3745, opened 2015-08-17
Bring in a minimal rubocop config

purbon:
Hi, 
   there is another PR, #3485 willing to bring in rubocop to logstash, however rubocop is mix there with a ruby style guide, so the chances to bring this PR in soon are low.

In this PR I wanted to include a minimal rubocop configuration, including a rake task, so we can actually get rubocop in sooner. This configuration include:
- The check for the encoding, that would have caught the last UTF-8 encoding issues.
- Checks for complexity and metrics.
- Checks for security.

I find this a minimal configuration, but feel free to bring your ideas to improve it!

/cheers


purbon:
Open questions here are:

* When to run the rubocop? do we aim this test to be mandatory?

for now is just another check you can pass, so no restriction.

- purbon

purbon:
@jordansissel your pair of eyes here are more than appreciate it.

jordansissel:
Will review soon.

Related, I'm not super sure how to enforce this across a PR because ideally we don't mix style changes with functional/behavioral changes. I have work on a tool that will review only the diffs, so maybe we can carry forward with that? I can demo it later this week if I resurrect it ;P

purbon:
Thanks @jordansissel 

In this cause I thought on adding rubocop, with this minimal configuration because it will help us track and detect this kind of issues, like encoding or too complex methods. I know rubocop makes also syntax checks, and also let you write your own cops. I think is a good tool, we can leverage.

How deep we integrate, and how strict are our "make rubocop happy" policies, might vary, but to have rubocop capabilities build in I think give good feedback to developers. What do you think?

I'm open to education regarding your diff verification tool, however I think we're talking about two things here. The only issue I see on introducing rubocop as mandatory thing with PR, is exactly the one you might be thinking with your diff tool. However to have rubocop run be default and report would be a good things, don't you think?

- purbon

PR #3761, opened 2015-08-20
Upgrade lumberjack lib

jordansissel:


jordansissel:
Not backporting. Not a security fix.

PR #3769, opened 2015-08-20
Wrong date and added Elasticsearch upsert support

ph:


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
      - For SSL certificate verification, The client now enforces the `VERIFY_PEER` mode when 
        connecting to the server. ([#4](https://github.com/elastic/ruby-lumberjack/issues/4))
     - Added better handling of congestion scenario on the output by using a buffered send of events ([#7](https://github.com/logstash-plugins/logstash-output-lumberjack/pull/7))
+  - Elasticsearch: Added the ability to update existing ES documents and support of upsert  -- if document doesn't exists, create it.([#116](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/116))
```
talevy: maybe make a nested description, just as above.

```
- Elasticsearch:
  - Added the ability to update existing ES documents and support of upsert  -- if document doesn't exists, create it.([#116](https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/116))
```

ph: We do that when we have multiple items. I dont think its need in that case.

suyograo:
LGTM

PR #3784, opened 2015-08-24
Adds a section on managing multiline events

palecur:
Has details on multiline plugin configuration.

### DIFF
--- a/docs/asciidoc/static/managing-multiline-events.asciidoc
+++ b/docs/asciidoc/static/managing-multiline-events.asciidoc
@@ -1,0 +1,107 @@
+[[multiline]]
+=== Managing Multiline Events
+
+Several use cases generate events that span multiple lines of text. In order to correctly handle these multline events, 
+Logstash needs to know how to tell which lines are part of a single event.
+
+Two different plugins handle recognizing single events that span multiple lines:
+
+* The {logstash}plugins-codecs-multiline.html[multiline codec] plugin combines multiple lines from a single input into 
+a single event.
+* The {logstash}plugins-filters-multiline.html[multiline filter] plugin combines multiple events, from one or more 
+inputs, into a single event.
+
+The most important aspects of configuring either multiline plugin are the following:
+
+* The `pattern` option specifies a regular expression. Lines that match the specified regular expression are considered 
+either continuations of a previous line or the start of a new multiline event. You can use 
+{logstash}plugins-filters-grok.html[grok] regular expression templates with this configuration option.
+* The `what` option takes two values: `previous` or `next`. The `previous` value specifies that lines that match the 
+value in the `pattern` option are part of the previous line. The `next` value specifies that lines that match the value 
+in the `pattern` option are part of the following line.* The `negate` option applies the multiline codec to lines that 
+_do not_ match the regular expression specified in the `pattern` option.
+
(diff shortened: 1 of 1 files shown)


andrewvc:
An important thing to note is that the multiline codec is far, far more preferable to the multiline filter because its scoped to a single input. The general principle is that the earlier in the pipeline multiline happens the better off you are If multiple lines can be assembled into a single event *before* they hit logstash the better off you are. Failing that, using the codec is the next best option. Failing that the filter.

The reason for this is that the farther away from the source you get the greater the chance of streams being mixed. The logstash filter does not actually work with multiple threads for this reason. The codecs can work, but they have issues when using some multithreaded inputs.

The main use cases where multiline makes sense are:

1. As a codec with the 'File' input
1. As a filter when workers will never be greater than 1 (maybe you're forwarding to another aggregating LS instance for heavier processing)
1. As a filter on a multiplexed input (like syslog) with [stream_identity](https://www.elastic.co/guide/en/logstash/current/plugins-filters-multiline.html#plugins-filters-multiline-stream_identity) enabled
1. You could also use [log-courier](https://github.com/driskell/log-courier/blob/master/docs/codecs/Multiline.md) which is a lightweight agent (forked from logstash-forwarder) with multiline 

There are other cases where it could make sense, but those are the most likely.

colinsurprenant:
@palecur thanks for this, it is indeed much needed.

We should add a section explaining why we have both a multiline filter and codec. A few notes on that subject:

- in most situations, the multiline **codec** should be used. We also recently added robustness and fixed potential OOM problem with the codec. 

- IIRC, the multiline **codec** does not work correctly with the lumberjack input which does not demux streams at the codec level so it can mix lines. @ph and/or @jordansissel please confirm?

- the multiline **codec** does not work correctly with the file input when the file input can create events from multiple sources/files see logstash-plugins/logstash-input-file#44

- the multiline **filter** is not thread safe and multiple filter workers should be avoided when using the multiline filter see logstash-plugins/logstash-filter-multiline#12

andrewvc:
@palecur Proposed copy

<blockquote>
The best multiline configuration is none at all. If you can get your events into JSON or some other well framed format that's the best possible logging situation and can save quite a bit of headache. However, we often don't have that choice, and for those situations logstash has a full suite of tools for merging together multiline logs. 

The preferred multiline tool in logstash is the `multiline` codec. This merges together the lines from a single input using a simple set of rules. A fallback if this doesn't work is the `multiline` filter, which does the same work, but at the filter stage of processing where multiple inputs are aggregated. The codec is preferred however because merging across inputs is rarely needed and fragile. There are a number of caveats to these two tools, which we will discuss later.
</blockquote>

suyograo:
@palecur can you reopen this for master?

palecur:
Will do.

palecur:
new PR is up: https://github.com/elastic/logstash/pull/3832

allenmchan:
This statement is mighty scary:
"IIRC, the multiline codec does not work correctly with the lumberjack input which does not demux streams at the codec level so it can mix lines. @ph and/or @jordansissel please confirm?"

https://github.com/logstash-plugins/logstash-input-lumberjack/pull/5 was the PR that added the functionality and your team asked me to use multiline codec with lumberjack.

So can i get a confirmation is it safe or not?

PR #3812, opened 2015-08-28
trigger input plugin shutdown using out of band call

### LINKED ISSUE #3210: [Design] plugins shutdown sequence refactor proposal

jsvd:
Currently, during shutdown, the pipeline loops through input workers
and calls Thread.raise on each input thread. Plugins rescue that
exception to exit the `run` loop. Afterwards, `teardown` is called for
any additional bookkeeping.

This proposal exposes a `stop` call on the input plugin class to alert
an input that it should shutdown. It is the plugin job to frequently
poll an internal `stop?` method to know if it's time to exit and
terminate if so.

The `teardown` method remains to do the additional bookkeeping, and it
will be called by the inputworker when `run` terminates.

resolves #3210 and replaces #3211

### DIFF
--- a/lib/logstash/inputs/base.rb
+++ b/lib/logstash/inputs/base.rb
@@ -67,6 +67,7 @@ class LogStash::Inputs::Base < LogStash::Plugin
   def initialize(params={})
     super
     @threadable = false
+    @stop_called = Concurrent::AtomicBoolean.new(false)
     config_init(params)
     @tags ||= []
 
@@ -99,6 +100,20 @@ class LogStash::Inputs::Base < LogStash::Plugin
   def tag(newtag)
     @tags << newtag
   end # def tag
+
+  # if you override stop, don't forget to call super
+  # as the first action
+  public
+  def stop
+    @logger.debug("stopping", :plugin => self)
+    @stop_called.make_true
+  end
+
+  # stop? should never be overriden
+  private
+  def stop?
+    @stop_called.value
+  end
 
   protected
   def to_event(raw, source)
--- a/lib/logstash/inputs/base.rb
+++ b/lib/logstash/inputs/base.rb
@@ -110,7 +110,7 @@ class LogStash::Inputs::Base < LogStash::Plugin
   end
 
   # stop? should never be overriden
-  private
+  public
   def stop?
     @stop_called.value
   end
--- a/lib/logstash/pipeline.rb
+++ b/lib/logstash/pipeline.rb
@@ -175,9 +175,14 @@ class LogStash::Pipeline
     LogStash::Util::set_thread_name("<#{plugin.class.config_name}")
     begin
       plugin.run(@input_to_filter)
-    rescue LogStash::ShutdownSignal
-      # ignore and quit
     rescue => e
+      # if plugin is stopping, ignore uncatched exceptions and exit worker
+      if plugin.stop?
+        @logger.debug("Ignoring stopping plugin exception", :exception => e, "backtrace" => e.backtrace)
+        return
+      end
+
+      # otherwise, report error and restart
       if @logger.debug?
         @logger.error(I18n.t("logstash.pipeline.worker-error-debug",
                              :plugin => plugin.inspect, :error => e.to_s,
@@ -187,23 +192,11 @@ class LogStash::Pipeline
         @logger.error(I18n.t("logstash.pipeline.worker-error",
                              :plugin => plugin.inspect, :error => e))
       end
-      puts e.backtrace if @logger.debug?
-      # input teardown must be synchronized since is can be called concurrently by
-      # the input worker thread and from the pipeline thread shutdown method.
-      # this means that input teardown methods must support multiple calls.
-      @run_mutex.synchronize{plugin.teardown}
-      sleep 1
+
+      sleep(1)
       retry
-    end
-  ensure
-    begin
-      # input teardown must be synchronized since is can be called concurrently by
-      # the input worker thread and from the pipeline thread shutdown method.
-      # this means that input teardown methods must support multiple calls.
-      @run_mutex.synchronize{plugin.teardown}
-    rescue LogStash::ShutdownSignal
-      # teardown could receive the ShutdownSignal, retry it
-      retry
+    ensure
+      plugin.teardown
     end
   end # def inputworker
 
@@ -242,8 +235,8 @@ class LogStash::Pipeline
       event = @filter_to_output.pop
       break if event == LogStash::SHUTDOWN
       output_func(event)
-    end # while true
-
+    end
+  ensure
     @outputs.each do |output|
       output.worker_plugins.each(&:teardown)
     end
@@ -255,21 +248,6 @@ class LogStash::Pipeline
   def shutdown
     InflightEventsReporter.logger = @logger
     InflightEventsReporter.start(@input_to_filter, @filter_to_output, @outputs)
-    @input_threads.each do |thread|
-      # Interrupt all inputs
-      @logger.info("Sending shutdown signal to input thread", :thread => thread)
-
-      # synchronize both ShutdownSignal and teardown below. by synchronizing both
-      # we will avoid potentially sending a shutdown signal when the inputworker is
-      # executing the teardown method.
-      @run_mutex.synchronize do
-        thread.raise(LogStash::ShutdownSignal)
-        begin
-          thread.wakeup # in case it's in blocked IO or sleeping
-        rescue ThreadError
-        end
-      end
-    end
 
     # sometimes an input is stuck in a blocking I/O so we need to tell it to teardown directly
     @inputs.each do |input|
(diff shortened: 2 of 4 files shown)


logstash-core.gemspec
```diff
@@ -23,6 +23,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency "clamp", "~> 0.6.5" #(MIT license) for command line args/flags
   gem.add_runtime_dependency "filesize", "0.0.4" #(MIT license) for :bytes config validator
   gem.add_runtime_dependency "gems", "~> 0.8.3"  #(MIT license)
+  gem.add_runtime_dependency "concurrent-ruby", "~> 0.9.1"
```
ph: ty for fixing a specific version. +1

jsvd: well actually I'm saying anything 0.9.*

ph: Yeah, I meant that :)

ph: I expect 1.0 to break a lot of stuff.

jordansissel:
Jenkins failure is due to build timeout. Can you investigate this cause?

jordansissel:
I tested this patch (and logstash-devutils#32) using https://github.com/logstash-plugins/logstash-input-pipe/pull/10. The spec for having 'stop' terminate the plugin fails: https://github.com/logstash-plugins/logstash-input-pipe/pull/10#issuecomment-136828599

lib/logstash/pipeline.rb
```diff
@@ -175,9 +175,14 @@ def inputworker(plugin)
     LogStash::Util::set_thread_name("<#{plugin.class.config_name}")
     begin
       plugin.run(@input_to_filter)
-    rescue LogStash::ShutdownSignal
-      # ignore and quit
     rescue => e
+      # if plugin is stopping, ignore uncatched exceptions and exit worker
+      if plugin.stop?
+        @logger.debug("Ignoring stopping plugin exception", :exception => e, "backtrace" => e.backtrace)
```
jordansissel: The wording here is confusing. Something like "Exception thrown while plugin was stopping, ignoring it." ?

lib/logstash/pipeline.rb
```diff
@@ -187,23 +192,11 @@ def inputworker(plugin)
...
-  ensure
-    begin
-      # input teardown must be synchronized since is can be called concurrently by
-      # the input worker thread and from the pipeline thread shutdown method.
-      # this means that input teardown methods must support multiple calls.
-      @run_mutex.synchronize{plugin.teardown}
-    rescue LogStash::ShutdownSignal
-      # teardown could receive the ShutdownSignal, retry it
+
+      sleep(1)
```
jordansissel: I know the previous code's sleep() didn't have a comment, but let's make a comment here explaining why we are sleeping.

jsvd:
@jordansissel it's normal for `rake test:core` to hang since some core tests use plugins that need refactoring, like logstash-input-generator

purbon:
I'm wondering if we should notify somehow from the plugin side the fact that the plugin is ready to work. This could be helpful in some situations like:

* In case a plugin got stalled during setup we could restart it.
* When doing things like in https://github.com/elastic/logstash-devutils/pull/32/files#diff-08b652c52dd968e57b68615d394bfbbdR15 we don't need to use sleep with an arbitrary amount of time, but waiting until the plugin is ready.

This strategy could also work for slow stop process where might be interesting to be be sure the process is inside the stop method, or also in other situations like being sleeping. I think we might be thinking about this the same way as the thread status api, but in our case the plugin status api.

This might simplify the handling of some situations inside the plugins like the wakeup from sleeping.

What do you think?

colinsurprenant:
@purbon I am not sure how we would define the semantic of a kind of a `ready?` method. normally, a plugin is ready once it returns from the `register` method? if not, when is a plugin ready? since we're dealing with async stuff, when are we ready and for what? 

Further:
- I don' t think «a plugin got stalled during setup» is a problem that need fixing. 
- potential «slow stop process» is not related to this `ready?` idea which is related to the plugin startup.
- not sure what you mean by a plugin «being sleeping»? if you mean waiting on a mutex or IO blocked, this is again not related to a `ready?` idea. 

Please open a new issue to submit this idea if you think this is still valuable and we can discuss it there, let's keep the focus on this PR.

lib/logstash/inputs/base.rb
```diff
@@ -100,6 +101,20 @@ def tag(newtag)
     @tags << newtag
   end # def tag
 
+  # if you override stop, don't forget to call super
+  # as the first action
+  public
+  def stop
+    @logger.debug("stopping", :plugin => self)
```
colinsurprenant: suggestion: maybe log "stop called" instead of "stopping" it seems to better represent the reality?

lib/logstash/pipeline.rb
```diff
@@ -277,15 +259,12 @@ def shutdown
         # input teardown must be synchronized since is can be called concurrently by
         # the input worker thread and from the pipeline thread shutdown method.
         # this means that input teardown methods must support multiple calls.
-        @run_mutex.synchronize{input.teardown}
+        @run_mutex.synchronize{input.stop}
       rescue LogStash::ShutdownSignal
```
colinsurprenant: the @run_mutex should not be necessary here?

colinsurprenant:
If we are to break the API any reason to keep `teardown` around? IMO `close` is more appropriate and commonly used for this?

lib/logstash/pipeline.rb
jsvd: is it still necessary anywhere? I was trying to find out why it was there in the first place, but couldn't find a reason.
I know it's supposed to isolate the "start input plugins" phase from a shutdown, but I couldn't see the need for the isolation.

colinsurprenant:
moving discussion about a potential `ready?` method in #3885

lib/logstash/pipeline.rb
colinsurprenant: after re-analyzing the `@run_mutex` related code it is useful for 2 things:
- consistent view of the `@ready` and `@started` properties which need to be consistent across threads since the related [`ready?` and `started?` methods](https://github.com/jsvd/logstash/blob/79d822a03929c5dd30a6b92bcf81a1a8705b18b3/lib/logstash/pipeline.rb#L54-L60) will be typically called from another thread that the one executing the `run` method 
- to avoid calling the shutdown code before the startup is completed

I believe the `ready?` and `started?` methods could be implemented using atomic booleans and for the prevention startup and shutdown race condition it should be refactored in a better way. It really isn't clear now.

I will open another issue for this

colinsurprenant: moved to #3886

colinsurprenant:
this PR is moving to #3895 and is now against the `feature/oob_plugins_stop` shared branch

PR #3404, opened 2015-06-09
Manage plugins to be installed offline

purbon:
The motivation of this PR is to enable users to dump a set of plugins, to later be managed in machines that have no internet connection.

This introduces a set of new commands that enable offline plugin management.

* ```bin/plugin pack```:  let's you create a bundle with all plugins currently installed in a LS instance plus their dependencies. By default this package of plugins is created as zip files in windows and tar.gz in unix machines.
* ```bin/plugin unpack```: From a previously created package this command lets you install it in another logstash instance.
* Add the ```--local``` flag option to the ```bin/plugin install``` command. When the new flag is passed the plugin manager will fetch the plugins from a local file system directory (created when doing the ```bin/plugin unpack```).
* Add the ```--local``` flag option to the ```bin/plugin update``` command. This command has the same behaviour and motivation as the one for the ```bin/plugin install```

## Expected workflow

The expected workflow would be:
*  a user created a package with the necessary plugins from a LS that has internet access.
* with the generated bundle (zip/tgz) the user unpack this file into a LS that has no internet.
* then using the ```--local``` flag when doing install or update he can install plugins from the local filesystem.

### bin/plugin pack:

```
option "--tgz", :flag, "compress package as a tar.gz file", :default => !LogStash::Environment.windows?   option "--zip", :flag, "compress package as a zip file", :default => LogStash::Environment.windows?
option "--[no-]clean", :flag, "clean up the generated dump of plugins", :default => true
option "--overwride", :flag, "Overwrite a previously generated package file", :default => false
```

### bin/plugin unpack

```
+  option "--tgz", :flag, "unpack a packaged tar.gz file", :default => !LogStash::Environment.windows?
+  option "--zip", :flag, "unpack a packaged  zip file", :default => LogStash::Environment.windows?
```

Relates/Fixes to #2376

PR #3826, opened 2015-09-01
Exclude pre releases when checking for update in the plugins manager

### LINKED ISSUE #3818: Plugin manager should not present beta plugins as upgrade options
Rubygems defines a beta as any plugin version with an alphabet character in it.

The code [here](https://github.com/elastic/logstash/blob/028d76497ccef3a8d11a3528cdf6b99e8d10f070/lib/pluginmanager/update.rb#L72) should skip past these versions.

purbon:
In the intention to provide a fix for #3818 this PR introduces some changes to exclude pre release versions from the filter/validation routine that's in place for the update process. This follows the actual bundler (out internal packet manager) of dealing with prereleases.

Bundler will always update (using the update command) to the last non prerelease version, in order to install a prerelease version actually the way is to change the name in the Gemfile + running bundler install. This could be archived in our setup by using ```bin/plugin install``` command with the right version (the pre release one)

We could add this to the update command, but __I would not recommend__ that for some reasons:
 * It breaks the bundler way of doing things, and we know, making bundler unhappy makes us unhappy.
 * I kinda understand the reason behind this, updating to a pre release is actually explicit, but throw uninstall/install.

Let me know what do you think :+1: 

/cheers

Fixes #3818

### DIFF
--- a/lib/pluginmanager/util.rb
+++ b/lib/pluginmanager/util.rb
@@ -37,6 +37,34 @@ module LogStash::PluginManager
     end
   end
 
+  # Fetch latest version information as in rubygems
+  # @param [String] The plugin name
+  # @param [Hash] Set of available options when fetching the information
+  # @option options [Boolean] :pre Include pre release versions in the search (default: false)
+  # @return [Hash] The plugin version information as returned by rubygems
+  def self.fetch_latest_version_info(plugin, options={})
+    require "gems"
+    exclude_prereleases =  options.fetch(:pre, false)
+    versions = Gems.versions(plugin)
+    versions = versions.select { |version| !version["prerelease"] } if !exclude_prereleases
+    versions.first
+  end
+
+  # Let's you decide to update to the last version of a plugin if this is a major version
+  # @param [String] A plugin name
+  # @return [Boolean] True in case the update is moving forward, false otherwise
+  def self.update_to_major_version?(plugin_name)
+    plugin_version  = fetch_latest_version_info(plugin_name)
+    latest_version  = plugin_version['number'].split(".")
+    current_version = Gem::Specification.find_by_name(plugin_name).version.version.split(".")
+    if (latest_version[0].to_i > current_version[0].to_i)
+      ## warn if users want to continue
+      puts("You are updating #{plugin_name} to a new version #{latest_version.join('.')}, which may not be compatible with #{current_version.join('.')}. are you sure you want to proceed (Y/N)?")
+      return ( "y" == STDIN.gets.strip.downcase ? true : false)
+    end
+    true
+  end
+
   # @param spec [Gem::Specification] plugin gem specification
   # @return [Boolean] true if this spec is for an installable logstash plugin
   def self.logstash_plugin_gem_spec?(spec)
--- a/lib/pluginmanager/update.rb
+++ b/lib/pluginmanager/update.rb
@@ -44,7 +44,7 @@ class LogStash::PluginManager::Update < LogStash::PluginManager::Command
     filtered_plugins = plugins.map { |plugin| gemfile.find(plugin) }
       .compact
       .reject { |plugin| REJECTED_OPTIONS.any? { |key| plugin.options.has_key?(key) } }
-      .select { |plugin| validate_major_version(plugin.name) }
+      .select { |plugin| validates_version(plugin.name) }
       .each   { |plugin| gemfile.update(plugin.name) }
 
     # force a disk sync before running bundler
@@ -67,16 +67,8 @@ class LogStash::PluginManager::Update < LogStash::PluginManager::Command
 
   # validate if there is any major version update so then we can ask the user if he is
   # sure to update or not.
-  def validate_major_version(plugin)
-    require "gems"
-    latest_version  = Gems.versions(plugin)[0]['number'].split(".")
-    current_version = Gem::Specification.find_by_name(plugin).version.version.split(".")
-    if (latest_version[0].to_i > current_version[0].to_i)
-      ## warn if users want to continue
-      puts("You are updating #{plugin} to a new version #{latest_version.join('.')}, which may not be compatible with #{current_version.join('.')}. are you sure you want to proceed (Y/N)?")
-      return ( "y" == STDIN.gets.strip.downcase ? true : false)
-    end
-    true
+  def validates_version(plugin)
+    LogStash::PluginManager.update_to_major_version?(plugin)
   end
 
   # create list of plugins to update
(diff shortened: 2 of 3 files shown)


purbon:
@andrewvc @suyograo  thoughts?

lib/pluginmanager/util.rb
```diff
@@ -37,6 +37,36 @@ def self.logstash_plugin?(plugin, version = nil)
...
   end
 
+  # Fetch latest version information as in rubygems
+  # @param [String] The plugin name
+  # @param [Hash] Set of available options when fetching the information
+  # @option options [Boolean] :pre Include pre release versions in the search (default: false)
+  # @return [Hash] The plugin version information as returned by rubygems
+  def self.fetch_latest_version_info(plugin, options={})
+    require "gems"
+    options = { :pre => false }.merge(options)
```
ph: Not necessary, but you can use `#fetch` to get a default value.
```ruby
include_release = `options.fetch(:pre, false)`
``

File: lib/pluginmanager/util.rb [hunk 2]
```diff
@@ -37,6 +37,36 @@ def self.logstash_plugin?(plugin, version = nil)
...
+    versions = Gems.versions(plugin)
+    if !options[:pre]
+      versions.select! { |version| !version["prerelease"] }
```
ph:
```ruby
versions.select! { |version| !version["prerelease"] } unless @options[:pre]
```

lib/pluginmanager/util.rb
```diff
@@ -37,6 +37,36 @@ def self.logstash_plugin?(plugin, version = nil)
...
+    plugin_version  = fetch_latest_version_info(plugin_name)
+    latest_version  = plugin_version['number'].split(".")
+    current_version = Gem::Specification.find_by_name(plugin_name).version.version.split(".")
+    if (latest_version[0].to_i > current_version[0].to_i)
+      ## warn if users want to continue
+      puts("You are updating #{plugin_name} to a new version #{latest_version.join('.')}, which may not be compatible with #{current_version.join('.')}. are you sure you want to proceed (Y/N)?")
+      return ( "y" == STDIN.gets.strip.downcase ? true : false)
+    end
+    true
+  end
```
ph: Might be worth it to use and expend this class `https://github.com/elastic/logstash/blob/master/lib/logstash/util/plugin_version.rb` ?

ph:
Added a few comments.

So this PR make sure we don't install pre-release version when we are doing a `plugin update`, but if I want to explicitly install a pre-release version how do I do that?

`bin/plugin install` and `bin/plugin update` should have some sort of `--pre`, `--prerelease` or a `--beta` flag?

purbon:
@ph thanks for your review, this PR actually what it does it makes sure --pre release versions are not getting into the validate_version method. Bundler is going to skip all pre releases unless you tell them directly, so we are doing the same. Does it makes sense?

ph:
Yes, unless you explicitly specify the pre-release version in the gemfile, bundler will not use them.
It make sense that both `validations` use the same rules.

I haven't tested it yet, but does the `bin/plugin install` command support installing pre-release version?

ph:
it probably work if the user specific the version when invoking the command.

lib/pluginmanager/util.rb
andrewvc: +1 for `#fetch`

lib/pluginmanager/util.rb
andrewvc: This would be cleaner as

`version = versions.select { |version| !version["prerelease"] } unless @options[:pre]`

you'd have to declare `version` as null, but mutating values like this is generally confusing

ph: agree for mutation.

andrewvc:
LGTM (sans the changes I proposed)

I haven't run any tests, but the code looks good.

suyograo:
so, can we install pre releases with this PR? or is it just to avoid them being installed explicitly?

purbon:
we can always install pre releases with logstash, for example from our master:

```
purbon-elastic% bin/plugin install --version 2.0.0.pre.beta2 logstash-output-elasticsearch
Validating logstash-output-elasticsearch-2.0.0.pre.beta2
Installing logstash-output-elasticsearch
Installation successful
purbon-elastic% bin/plugin list
logstash-output-elasticsearch
purbon-elastic% bin/plugin list --verbose
logstash-output-elasticsearch (2.0.0.pre.beta2)
```

what happens, that this PR fix, is when doing the update you won't be ask if you want to update to a pre release (major version validation). Bundler will never update to a pre release, so it makes no sense to pop this messages.

Does it makes sense?

- purbon

purbon:
@ph as you see in https://github.com/elastic/logstash/pull/3826#issuecomment-136846644 it works as I expected.

lib/pluginmanager/util.rb
purbon: That might make sense, however what is still not clear to me is that this class you speak about is under logstash, but when we're in the plugin manager we're not supposed to load stuff under there. At less this is what I understood while working on #3404 (about the pack/unpack commands you can see in the utils module there, located under bootstrap).

I agree we should have this kind of things unified, this is a great idea for a refactoring where we can handle this kind of code only in one place.

What do you think?

suyograo:
> we can always install pre releases with logstash, for example from our master:

when you say this ^^, which version of LS are you using. Does it include this patch? Because from what @andrewvc and I tried on master a couple of days go, installing pre release plugin was not possible.

purbon:
@suyograo installing I did with the LS master, it works out of the box, another question is update. Update from bundler side is never going to pick up a pre release version, you need to install it specifically. What this PR does is to skip the pre releases in the version validation, so boths validations works the same way.

Does it makes sense to you @suyograo ?

lib/pluginmanager/util.rb
ph: I have been bite again bit the chicken/egg problem, you are right its not available in the current scope.

purbon: I would like to get this "fix" in a feature refactoring release :smile:, some of this things are reasonably to be available here.

purbon:
@ph do you see more open questions for this PR? good to go?

ph:
I am currently running a local test, thank for clarifying all my questions :)

purbon:
you're welcome man! Is always a pleasure!

ph:
code lgtm, the tests run and I have tested it.

purbon:
thanks @ph, 
  @suyograo @andrewvc are you ok with me merging this?

andrewvc:
LGTM

Issue #3834, opened 2015-09-01
Jenkins: Build failing in Lumberjack


suyograo:
http://build-eu-00.elastic.co/view/LS%20Master/job/Logstash_Master_Default_Plugins/

jordansissel:
Ahh.

The `try` helper method was added logstash-devutils 0.0.15, but the master
Gemfile.lock is using 0.0.14

suyograo:
Fixed in #3836

PR #3835, opened 2015-09-01
Update ot JRuby 1.7.22

suyograo:
Tests pass locally. One test that failed (lumberjack) fails on master too. #3834


jordansissel:
Didn't test, but LGTM (+1 on upgrading to Jruby 1.7.22)

PR #3836, opened 2015-09-01
Update devutils to 0.0.15 or newer

jordansissel:
This ensures we have a 'try' helper method in rspec

Fixes a failing test in lumberjack output when combining all plugins for a big test.


suyograo:
LGTM

jordansissel:
Checked the 1.5 branch, and it has 0.0.15 in the .lock file so no update is needed on that branch.

PR #3848, opened 2015-09-03
generate lock file for 2.0.0.beta1

suyograo:
Includes beta versions of `logstash-output-elasticsearch` and `logstash-output-kafka`


Gemfile
```diff
@@ -2,7 +2,7 @@
 # If you modify this file manually all comments and formatting will be lost.
 
 source "https://rubygems.org"
-gem "logstash-core", "2.0.0.dev", :path => "."
+gem "logstash-core", "2.0.0.snapshot1"
```
purbon: why not using here something like " ~> 2.0.0.snapshot" then all snapshot + new non pre release version will be into this pattern? I think makes sense. What do you think?

purbon:
@ph @suyograo left you small comment on the pattern used.

Gemfile
ph: @suyograo ^ this would work, IIRC bin/plugin update shouldn't update the core.

I've tested this version guard with:
```
source "https://rubygems.org"
gem "logstash-output-elasticsearch", "~> 2.0.0.beta"
```

`bundle install` command installed the latest version which is beta 2

suyograo: @purbon @ph Should that be `"~> 2.0.0.snapshot"` or `"~> 2.0.0.snapshot1"`

ph: @suyograo I would say "2.0.0.snapshot" since  "~> 2.0.0.beta" worked in my test to get the latest beta1...

I've looked through the rubygems doc this morning and I wasn't able to find any information about specifying version for prerelease gems, it might an unwanted effect? So after a bit of thinking maybe we should fix it with something we know ?

purbon: @suyograo using  `"~> 2.0.0.snapshot"`  will get everything from now on, imagine we do an snapshot two, etc ... but not beta or other labels (the check is for alphabetic order I think, but not sure) 

using  `"~> 2.0.0.snapshot1"` will do the same.

purbon:
LGTM

ph:
lgtm

Issue #3599, opened 2015-07-13
Refactor all usage of def Module::Module::Class style declarations


guyboertje:
There are potential problems with this style of Class declaration in that:
1. While the class is associated with the referenced namespace it is defined in the root scope
2. It is not idiomatic Ruby
   3: inner classes must also be declared with the full namespace reference otherwise they end up being defined on the root namespace

We should make attempt to fix this as we go.

I think we all don't want to see a lot of indentation (the original reason for coding like this).

I think everyone will be happy with...

``` ruby
module X module Y module Z
  def self.foo() puts 'foo'; end
end end end
```

NOTE: in JRuby 1.7 there is no need for semi-colons e.g.

``` ruby
module X; module Y; module Z
end; end; end;
```

guyboertje:
I have begun to make these changes as I fix bugs in the plugins.  However LogStash core classes will need one big update.

guyboertje:
I think people are mostly doing this for new classes. Closing

PR #3870, opened 2015-09-04
Support added for filter_workers default to cpu-cores

guyboertje:
WIP

add basic defaults module and spec.  use defaults in pipeline and agent

### DIFF
--- a/lib/logstash/config/cpu_core_strategy.rb
+++ b/lib/logstash/config/cpu_core_strategy.rb
@@ -1,0 +1,32 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/config/defaults"
+
+module LogStash module Config module CpuCoreStrategy
+
+  extend self
+
+  def maximum
+    LogStash::Config::Defaults.cpu_cores
+  end
+
+  def fifty_percent
+    [1, (maximum * 0.5)].max.floor
+  end
+
+  def seventy_five_percent
+    [1, (maximum * 0.75)].max.floor
+  end
+
+  def twenty_five_percent
+    [1, (maximum * 0.25)].max.floor
+  end
+
+  def max_minus_one
+    [1, (maximum - 1)].max.floor
+  end
+
+  def max_minus_two
+    [1, (maximum - 2)].max.floor
+  end
+end end end
--- a/lib/logstash/config/defaults.rb
+++ b/lib/logstash/config/defaults.rb
@@ -1,0 +1,12 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "concurrent"
+
+module LogStash module Config module Defaults
+
+  extend self
+
+  def cpu_cores
+    Concurrent.processor_count
+  end
+end end end
--- a/lib/logstash/util/defaults_printer.rb
+++ b/lib/logstash/util/defaults_printer.rb
@@ -1,0 +1,31 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/util"
+require "logstash/util/worker_threads_default_printer"
+
+
+# This class exists to format the settings for defaults used
+module LogStash module Util class DefaultsPrinter
+  def self.print(settings)
+    new(settings).print
+  end
+
+  def initialize(settings)
+    @settings = settings
+    @printers = [workers]
+  end
+
+  def print
+    collector = []
+    @printers.each do |printer|
+      printer.visit(collector)
+    end
+    "Default settings used: " + collector.join(', ')
+  end
+
+  private
+
+  def workers
+    WorkerThreadsDefaultPrinter.new(@settings)
+  end
+end end end
--- a/lib/logstash/agent.rb
+++ b/lib/logstash/agent.rb
@@ -2,6 +2,7 @@
 require "clamp" # gem 'clamp'
 require "logstash/environment"
 require "logstash/errors"
+require "logstash/config/cpu_core_strategy"
 require "uri"
 require "net/http"
 LogStash::Environment.load_locale!
@@ -21,7 +22,8 @@ class LogStash::Agent < Clamp::Command
 
   option ["-w", "--filterworkers"], "COUNT",
     I18n.t("logstash.agent.flag.filterworkers"),
-    :attribute_name => :filter_workers, :default => 1, &:to_i
+    :attribute_name => :filter_workers,
+    :default => LogStash::Config::CpuCoreStrategy.fifty_percent, &:to_i
 
   option ["-l", "--log"], "FILE",
     I18n.t("logstash.agent.flag.log"),
--- a/lib/logstash/pipeline.rb
+++ b/lib/logstash/pipeline.rb
@@ -10,6 +10,8 @@ require "logstash/filters/base"
 require "logstash/inputs/base"
 require "logstash/outputs/base"
 require "logstash/util/reporter"
+require "logstash/config/cpu_core_strategy"
+require "logstash/util/defaults_printer"
 
 class LogStash::Pipeline
 
@@ -38,7 +40,7 @@ class LogStash::Pipeline
     # if no filters, pipe inputs directly to outputs
     @filter_to_output = filters? ? SizedQueue.new(20) : @input_to_filter
     @settings = {
-      "filter-workers" => 1,
+      "filter-workers" => LogStash::Config::CpuCoreStrategy.fifty_percent
     }
 
     # @ready requires thread safety since it is typically polled from outside the pipeline thread
@@ -51,10 +53,10 @@ class LogStash::Pipeline
   end
 
   def configure(setting, value)
-    if setting == "filter-workers"
+    if setting == "filter-workers" && value > 1
       # Abort if we have any filters that aren't threadsafe
-      if value > 1 && @filters.any? { |f| !f.threadsafe? }
-        plugins = @filters.select { |f| !f.threadsafe? }.collect { |f| f.class.config_name }
+      plugins = @filters.select { |f| !f.threadsafe? }.collect { |f| f.class.config_name }
+      if !plugins.size.zero?
         raise LogStash::ConfigurationError, "Cannot use more than 1 filter worker because the following plugins don't work with more than one worker: #{plugins.join(", ")}"
       end
     end
@@ -66,6 +68,8 @@ class LogStash::Pipeline
   end
 
   def run
+    @logger.terminal(LogStash::Util::DefaultsPrinter.print(@settings))
+
     begin
       start_inputs
       start_filters if filters?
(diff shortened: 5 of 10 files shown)


guyboertje:
still need to do a performance benchmark

ph:
@andrewvc can you run your benchmark on this pr?

guyboertje:
@ph @andrewvc - have we implemented benchmark specs? i.e. http://chriskottom.com/blog/2015/04/minitest-benchmark-an-introduction/

So in a spec (if in the test env [CI?] cpu-cores > 1) I can verify that `-w 1` benchmark is less than default (cpu-cores).  Overkill? Possibly unreliable?

Although, ideally, in general, for all the defaults changes, we should have some mechanism that expects an improvement when a default value change is committed.  In practice, though, it may only be partially achievable in the spec suite.

ph:
I just remembered that @colinsurprenant had some kind of performance suite @pere can help you with it.  I'll start with it instead of @andrewvc stuff

ph:
https://github.com/elastic/logstash-performance-testing this is the repo

andrewvc:
@ph makes sense. To clarify, my test is only really useful for testing LS -> Elasticsearch end to end. For most stuff including elasticsearch just muddies the results.

suyograo:
@guyboertje what benchmarks were you planning to run for this? One suggestion: we can easily show grok's perf bump with multiple filters here.. as @ph mentioned https://github.com/elastic/logstash-performance-testing can be used here..

guyboertje:
Benchmarking - i will use what ph and suyog mentioned.

purbon:
@guyboertje let me know if https://github.com/elastic/logstash-performance-testing is clear enought to be used or you need me to improve it somehow, need clarification, etc.. All contributions welcome!.

lib/logstash/util/defaults.rb
```diff
@@ -0,0 +1,16 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/util"
+
+module LogStash module Util module Defaults
```
ph: Can we have bit of <CR> here? :)

lib/logstash/util/defaults.rb
```diff
@@ -0,0 +1,16 @@
...
+  def self.filter_workers_count
+    cpu_cores
+  end
+
+  # specifics
+
+  def self.cpu_cores
+    Java::Java.lang.Runtime.getRuntime.availableProcessors
+  end
+end end end
```
ph: Can we have bit of <CR> here? :)

ph: Because of the refactoring of the pipeline we will be using the `concurrent-ruby` gem a lot more and it already provide a method to return the number of cpu.

https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/utility/processor_counter.rb

It uses the same code as your behind the scene, they also provide a `physical_core` method.
I think we can use it directly?

lib/logstash/util/defaults.rb
ph: I think this module should be in `lib/logstash/config/defaults.rb` since its more about default config value than util, what do you think?

ph:
Added minor comments, address them and we will be ready to merge it! Yeah for cores!

lib/logstash/util/defaults.rb
guyboertje: @ph - sure that makes good sense, will change.

ph:
After a discussion on rabbitmq with @electrical,  I think we shouldn't reserve all the machine cores the workers, maybe half the core would a more reasonable default?

purbon:
@ph is cool if you are living with more inhabitants in a machine to let them have CPU's, otherwise your threads are going to in and out.

electrical:
for the sake of argument, lets assume people setup LS dedicated on a box without anything else running :-) they can always change it if that is not the case.

colinsurprenant:
let's not forget that we do not have control over the *affinity* or relation between threads and cores. There is already X number of *base* threads running when starting logstash (JVM threads, logstash threads) and we want to optimize how many more Y filter worker threads we will be adding. We do not really control the relation between these Y filter worker thread and the number of CPU cores (there are already X number of threads running) but we know that in ideal circumstances is should be a somewhat linear relation between both but it is not clear if 1:1 is necessarily the best relation. 

also, while doing some performance tests, I have seems occasions where setting the number of filter worker thread higher than the number of CPU cores (like 2x) did yield a better performance. I would guess this was because those threads were much sleeping/blocking on IO or system calls - and it might have been a particular filters combination (which I don't recall).

purbon:
@colinsurprenant very valid points, setting the number of threads to the number of CPU by default it could be misleading in some situations.

guyboertje:
Given that this is a default setting and as such cannot second guess any particular environment, what is the conclusion for the setting?
- [ ] `cores * 1`
- [ ] `cores * 3 / 4`
- [ ] `cores / 2`
- [ ] `[1, cores - 1].max`
- [ ] `[1, cores - 2].max`
- [ ] `other, please specify algorithm`

guyboertje:
Should I create a rake task that the user can run to show the number of cores JRuby detects?

guyboertje:
Should the -w parameter also support a float e.g -w 0.75 means workers =  75% of cores as an int?
Also negative integer e.g. -w -2 means workers = cores - 2.
This would make parameter less absolute.

guyboertje:
Further, given the comments above, is local (laptop) perf testing meaningful?

ph:
@guyboertje personally I wont change how `-w` work.

@colinsurprenant I seen that in the past two, but theses scenario can be environment and config specific. I still think giving half the cpu core to the workers is a conservative choice.

ph:
@colinsurprenant since you hop in the thread what is your algorithm suggestion? :)

colinsurprenant:
I don't have an algorithm suggestion handy, just wanted to add a bit more confusion. that's ok, don't thank me.

jordansissel:
@colinsurprenant haha

jordansissel:
I'm OK exploring a "use a number of workers relative to the number of detected cpu cores" but we can always add that later in a future revision. Phase/version 1 of this doesn't have to detect or support percentages. Chef/puppet/etc users will already have access to "number of cores" so they can easily forward this (or a calculation based-on) to Logstash for setting worker count.

colinsurprenant:
but seriously, I would probably be pick a conservative default like `cores * 0.5` or `cores * 0.75`. This should be a **better** default in most situations.

jordansissel:
@colinsurprenant +1 on defaulting to something like 0.5 cores or 0.75 cores

ph:
@colinsurprenant @guyboertje +1   => cores * 0.5

Like I said ;)
>maybe half the core would a more reasonable default?

colinsurprenant:
@ph corrected.

colinsurprenant:
... also (I just won't shut up already) - when we have a metrics api, we could also try to auto-adjust (using a cmd line switch?) : from default, add/remove workers and gather performance impact and  find the best number of workers for this particular environment & config and output result to console:

```
Logstash magically figured that the optimal number of filter workers for this environment should be 42
```

ph:
@colinsurprenant old news .. ;) 

see: https://github.com/elastic/logstash/issues/3897
> This would allow to decouple the message from that actual work. This could allow us to do some neat trick like dynamic resources allocation for slower plugins based on the metric data.

lib/logstash/util/defaults.rb
```diff
@@ -0,0 +1,16 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/util"
+
+module LogStash module Util module Defaults
+
+  def self.filter_workers_count
```
colinsurprenant: may I suggest using the `extend self` pattern for "singleton" modules? ex.:
```ruby
module Defaults
  extend self

  def filter_workers_count
    cpu_cores
  end
end

**Comment by ph** [comment]
I will review it today  :D

File: lib/logstash/config/cpu_core_strategy.rb [hunk 4]
```diff
@@ -0,0 +1,29 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/config/defaults"
+
+module LogStash module Config module CpuCoreStrategy
```
andrewvc: is this our new indenting style? I've been making christmas trees. Has this been discussed?

lib/logstash/config/defaults.rb
```diff
@@ -0,0 +1,9 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "concurrent"
+
+module LogStash module Config module Defaults
+  def self.cpu_cores
+    Concurrent.processor_count
```
jordansissel: Does this value reflect real cores? or hyperthreaded cores?

lib/logstash/config/cpu_core_strategy.rb
jordansissel: hah, I think we discussed it a few weeks ago. I don't remember where (email?). I think the goal here was a hybrid between the brevity of `LogStash::Config::CpuCoreStrategy` and the scoping properties of `module Logstash\n module Config\n  module CpuCoreStrategy`

andrewvc:
Great work @guyboertje ! Very clean and well written.

Code almost LGTM, the only thing missing is that we should log how many CPU cores we're actually using. I tried starting up LS on this branch and I didn't see that printed out anywhere, even with `--debug`.

ph:
Lgtm same comment as @andrewvc,  also not sure about defining in-line module

lib/logstash/config/defaults.rb
guyboertje: @jordansissel - Concurrent does provide a physical_processor_count method but on the JVM both methods return the hyperthreads.

guyboertje:
@ph - I am not sure what you are referring to.
@andrewvc  - I will try to add that.

lib/logstash/config/defaults.rb
jordansissel: I suppose we could just note (comment it) what processor count is returned (physical cores vs hyperthreads) and that would satisfy me :)

ph:
@guyboertje 
```ruby
module LogStash module Config module Defaults
```
 Not sure about this style, where/when it was decided, I might have missed that discussion.

guyboertje:
@ph - see jordan's comment. https://github.com/elastic/logstash/pull/3870#discussion_r40034269 and
https://github.com/elastic/logstash/issues/3599
Other that the above, I think we ZOOMed on it, I am not sure who was on the ZOOM.

ph:
@guyboertje Thx for bringing it up!

andrewvc:
I'm fine with the style FYI :)

andrewvc:
LGTM

lib/logstash/util/defaults_printer.rb
```diff
@@ -0,0 +1,31 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/util"
+require "logstash/util/worker_threads_default_printer"
+
+
+# This class exists to quietly wrap a password string so that, when printed or
```
suyograo: @guyboertje can you update the comment :)

guyboertje:
I need to write tests for the added defaults printer classes and benchmark

PR #3872, opened 2015-09-04
Move Clamp away from Agent, Remove Subcommands and Clean CLI

jsvd:
* move all cli parsing to Runner, freeing Agent to solely manage pipelines
* remove all subcommands, focusing on the "agent" behaviour
* ensure logging goes through Cabin as much as possible
* log fatal messages to terminal even --log is enabled
* Agent now starts empty, and 1 pipeline is added and started afterwards
* for now it Agent only supports 1 pipeline, but will easily support more than 1
* remove doc generation subcommand from bin/logstash
* adds "-i irb/pry" flag for interactive shell

Depends on https://github.com/jordansissel/ruby-cabin/pull/37


guyboertje:
This breaks my PR on agent for default filter_workers :-)

lib/logstash/runner.rb
```diff
@@ -3,22 +3,72 @@
...
+  option ["-w", "--filterworkers"], "COUNT",
+    I18n.t("logstash.runner.flag.filterworkers"),
+    :attribute_name => :filter_workers, :default => 1, &:to_i
+
+  option ["-l", "--log"], "FILE",
+    I18n.t("logstash.runner.flag.log"),
+    :attribute_name => :log_file
+
+  # Old support for the '-v' flag'
+  option "-v", :flag,
```
suyograo: do we need this still? we can remove this since we support `--debug`. Than we can get rid of `-V` and call it `-v` for version like all other systems do. Thoughts?

jsvd:
@guyboertje I'll likely lose the race on master, so I'll adapt this, don't worry :D

lib/logstash/runner.rb
jsvd: I like.

jordansissel: I'm open to that, though it may be surprising to users.  If anything, I'd prefer that `-v` went away and did _not_ change to report the version - mostly to avoid surprising users.

suyograo: Historically `-v` has been used to increase log verbosity, so its mostly an advanced user functionality. I would argue that replacing it with version output wouldn't surprise them. Maybe they have already migrated to `--debug`. All in all `-v` is a no-side-affect action, and folks wouldn't have used this in scripts so its less dangerous :)

lib/logstash/runner.rb
```diff
@@ -30,95 +85,245 @@ def main(args)
     # Print a warning to STDERR for bad java versions
     LogStash::Util::JavaVersion.warn_on_bad_java_version
 
-    Stud::untrap("INT", @startup_interruption_trap)
+    if version?
```
andrewvc: It would be awesome if this function was broken up into smaller ones

```ruby
show_version
configure

```

lib/logstash/runner.rb
```diff
@@ -30,95 +85,245 @@ def main(args)
...
+  rescue LoadError => e
+    fail("Configuration problem.")
+  rescue LogStash::ConfigurationError => e
+    @logger.fatal I18n.t("logstash.error", :error => e)
+    if !config_test?
+      @logger.warn I18n.t("logstash.runner.configtest-flag-information")
+    end
+    show_short_help
+    return 1
+  rescue => e
```
andrewvc: Should this be `rescue Exception` now given @guyboertje and @colinsurprenant 's latest research?

lib/logstash/runner.rb
```diff
@@ -30,95 +85,245 @@ def main(args)
...
-          # the next command. Otherwise it's a real error.
-          raise if e.message != "too many arguments"
-          remaining = agent.remaining_arguments
+  def show_version
+    show_version_logstash
+
+    if [:info, :debug].include?(verbosity?) || debug? || verbose?
+      show_version_ruby
+      show_version_java if LogStash::Environment.jruby?
+      show_gems if [:debug].include?(verbosity?) || debug?
```
andrewvc: This is probably copy/paste from the old spot, but this could be cleaned up!

bin/logstash
```diff
@@ -22,11 +22,4 @@ unset CDPATH
...
 
-case $1 in
-  -*)
-    ruby_exec "${LOGSTASH_HOME}/lib/bootstrap/environment.rb" "logstash/runner.rb" "agent" "$@"
-    ;;
-  *)
-    ruby_exec "${LOGSTASH_HOME}/lib/bootstrap/environment.rb" "logstash/runner.rb" "$@"
-    ;;
-esac
+ruby_exec "${LOGSTASH_HOME}/lib/bootstrap/environment.rb" "logstash/runner.rb" "$@"
```
andrewvc: I wonder if it might be nice to still provide support for 'irb' and 'pry' by checking for them in bash?

andrewvc:
On the whole this looks very good. Thanks for the work @jsvd 

I would have preferred that we not deprecate commands unless strictly necessary, but then again, times change.

LGTM.

jsvd:
@andrewvc I have updated this to ensure bwc with the bin/logstash [irb/pry]

bin/logstash
jsvd: done in b4a9a51f71351d4f1734d98b0038a1e87f310fb6

andrewvc:
@jsvd awesome :) I think our users will appreciate it! :+1:

lib/logstash/agent.rb
```diff
@@ -181,167 +55,28 @@ def shutdown(pipeline)
...
-  end # def configure_logging
-
-  # add the given paths for ungemified/bare plugins lookups
-  # @param paths [String, Array<String>] plugins path string or list of path strings to add
-  def configure_plugin_paths(paths)
-    Array(paths).each do |path|
-      fail(I18n.t("logstash.agent.configuration.plugin_path_missing", :path => path)) unless File.directory?(path)
-      LogStash::Environment.add_plugin_path(path)
-    end
+  def shutdown_pipelines
```
ph: Why do we have two `shutdown` methods? (shutdown and shutdown_pipelines)

The shutdown method and the hook was added to fix an issue with the test see
https://github.com/elastic/logstash/pull/4024

lib/logstash/agent.rb
```diff
@@ -181,167 +55,28 @@ def shutdown(pipeline)
...
-        @logger.debug? && @logger.debug("\nThe following is the content of a file", :config_file => file.to_s)
-        @logger.debug? && @logger.debug("\n" + cfg + "\n\n")
+  def trap_sigint
+    Stud::trap("INT") do
+      if @interrupted_once
+        @logger.fatal(I18n.t("logstash.agent.forced_sigint"))
+        exit
+      else
+        @logger.warn(I18n.t("logstash.agent.sigint"))
+        Thread.new(@logger) {|logger| sleep 5; logger.warn(I18n.t("logstash.agent.slow_shutdown")) }
```
ph: For clarity can we use a block in that thread definition?

lib/logstash/runner.rb
ph: +1 for `-v` to return version

ph:
LGTM! really minor comments.

andrewvc:
@jsvd what do we need to do to move this into master? I'd love to merge #4194 in as well as follow up with my work on #4254

jsvd:
@andrewvc just rebasing against master/2.x, will do this today

Issue #3886, opened 2015-09-08
refactor @run_mutex for the new shutdown logic


colinsurprenant:
`@run_mutex` specific refactor discussion moved here from #3812. Copied [comment](https://github.com/elastic/logstash/pull/3812/files#r38965625) 

> after re-analyzing the `@run_mutex` related code it is useful for 2 things:
> - consistent view of the `@ready` and `@started` properties which need to be consistent across threads since the related [`ready?` and `started?` methods](https://github.com/jsvd/logstash/blob/79d822a03929c5dd30a6b92bcf81a1a8705b18b3/lib/logstash/pipeline.rb#L54-L60) will be typically called from another thread that the one executing the `run` method 
> - to avoid calling the shutdown code before the startup is completed

> I believe the `ready?` and `started?` methods could be implemented using atomic booleans and for the prevention startup and shutdown race condition it should be refactored in a better way. It really isn't clear now.

jsvd:
`ready?` and `started?` are only used for plugin specs, similar to what we're discussing in https://github.com/elastic/logstash/issues/3885

And looking at the existing plugins we have:

```
~/all_logstash_plugins % grep -e started? -e ready? . -r
./logstash-input-relp/spec/spec_helper.rb:    sleep 0.1 while !pipeline.ready?
./logstash-output-s3/spec/outputs/s3_spec.rb:        sleep 0.1 while !pipeline.ready?
```

```
~/projects/logstash % grep -e started? -e ready? . -r
./lib/logstash/pipeline.rb:  def ready?
./lib/logstash/pipeline.rb:  def started?
```

Regarding the second point "to avoid calling the shutdown code before the startup is completed", do you mind doing a thought experiment and answer.."why is that a requirement?"

colinsurprenant:
PR for this for review in jsvd/logstash#1

colinsurprenant:
commit for this pushed in #3895

suyograo:
Addressed in https://github.com/elastic/logstash/pull/3895

PR #3895, opened 2015-09-09
pipeline shutdown OOB plugins stop

### LINKED ISSUE #3210: [Design] plugins shutdown sequence refactor proposal

colinsurprenant:
This is a PR reboot of #3812 from a shared branch to facilitate collaboration.

### Original description by @jsvd :

Currently, during shutdown, the pipeline loops through input workers
and calls Thread.raise on each input thread. Plugins rescue that
exception to exit the `run` loop. Afterwards, `teardown` is called for
any additional bookkeeping.

This proposal exposes a `stop` call on the input plugin class to alert
an input that it should shutdown. It is the plugin job to frequently
poll an internal `stop?` method to know if it's time to exit and
terminate if so.

The `teardown` method remains to do the additional bookkeeping, and it
will be called by the inputworker when `run` terminates.

resolves #3210 and replaces #3211

### DIFF
--- a/docs/asciidoc/static/include/pluginbody.asciidoc
+++ b/docs/asciidoc/static/include/pluginbody.asciidoc
@@ -725,24 +725,24 @@ endif::receive_method[]
 
 // Teardown is now in the base class... can be pruned?
 // /////////////////////////////////////////////////////////////////////////////
-// If teardown_method is defined (should only be for input or output plugin page)
+// If close_method is defined (should only be for input or output plugin page)
 // /////////////////////////////////////////////////////////////////////////////
-// ifdef::teardown_method[]
+// ifdef::close_method[]
 // [float]
-// ==== `teardown` Method
+// ==== `close` Method
 // [source,ruby]
 // [subs="attributes"]
 // ----------------------------------
 // public
-// def teardown
+// def close
 //   @udp.close if @udp && !@udp.closed?
 // end
 // ----------------------------------
-// The `teardown` method is not present in all input or output plugins.  It is
+// The `close` method is not present in all input or output plugins.  It is
 // called when a shutdown happens to ensure that sockets, files, connections and
 // threads are all closed down properly.  If your plugin uses these connections,
-// you should include a teardown method.
-// endif::teardown_method[]
+// you should include a close method.
+// endif::close_method[]
 
 ==== Building the Plugin
 
--- a/lib/logstash/codecs/base.rb
+++ b/lib/logstash/codecs/base.rb
@@ -28,7 +28,7 @@ module LogStash::Codecs; class Base < LogStash::Plugin
   end # def encode
 
   public 
-  def teardown; end;
+  def close; end;
 
   # @param block [Proc(event, data)] the callback proc passing the original event and the encoded event
   public
--- a/lib/logstash/filters/base.rb
+++ b/lib/logstash/filters/base.rb
@@ -235,7 +235,7 @@ class LogStash::Filters::Base < LogStash::Plugin
   end
 
   public
-  def teardown
+  def close
     # Nothing to do by default.
   end
 end # class LogStash::Filters::Base
--- a/lib/logstash/inputs/base.rb
+++ b/lib/logstash/inputs/base.rb
@@ -101,12 +101,20 @@ class LogStash::Inputs::Base < LogStash::Plugin
     @tags << newtag
   end # def tag
 
-  # if you override stop, don't forget to call super
-  # as the first action
   public
+  # override stop if you need to do more than do_stop to
+  # enforce the input plugin to return from `run`.
+  # e.g. a tcp plugin might need to close the tcp socket
+  # so blocking read operation aborts
   def stop
+    # override if necessary
+  end
+
+  public
+  def do_stop
     @logger.debug("stopping", :plugin => self)
     @stop_called.make_true
+    stop
   end
 
   # stop? should never be overriden
--- a/lib/logstash/inputs/base.rb
+++ b/lib/logstash/inputs/base.rb
@@ -67,6 +67,7 @@ class LogStash::Inputs::Base < LogStash::Plugin
   def initialize(params={})
     super
     @threadable = false
+    @stop_called = Concurrent::AtomicBoolean.new(false)
     config_init(params)
     @tags ||= []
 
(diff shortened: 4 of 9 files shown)


colinsurprenant:
pushing commits related to `started?`, `ready?` and `@run_mutex cleanups` as discussed in https://github.com/elastic/logstash/pull/3812#discussion_r38965625 and #3886

purbon:
@colinsurprenant @jsvd Added https://github.com/jordansissel/ruby-stud/issues/23 to track work on making ```Stud``` more interrupt friendly.

talevy:
why not get rid of `close`/`teardown` in favor of just using `stop` for both stop signalling and cleanup?

jsvd:
In some plugins stop will not make run abort right away (maybe stop? will
only be checked in the next loop).
Such plugins might still need to perform cleanup/teardown actions

andrewvc:
@jsvd Is there a reason output plugins don't have `stop` and `stop?`. I could use this for the rabbitmq output where it sometimes sleeps if the external rabbitmq server is down, waiting to retry (it could check `stop?` in a loop.

talevy:
@jsvd, do you know what is up with this?

https://github.com/elastic/logstash/blob/feature/oob_plugins_stop/lib/logstash/pipeline.rb#L103-L106
```
  def wait_inputs
    @input_threads.each(&:join)
  rescue Interrupt
    # rbx does weird things during do SIGINT that I haven't debugged
    # so we catch Interrupt here and signal a shutdown. For some reason the
    # signal handler isn't invoked it seems? I dunno, haven't looked much into
    # it.
    shutdown
  end
```

I think we should clean that up for this PR. not sure what it is referencing.

jsvd:
Agreed, looking at that comment we can actually reduce the method to:

```ruby
def wait_inputs
  @input_threads.each(&:join)
end
```

jsvd:
@andrewvc eventually both filters and outputs might also have `stop`. The first implementation of this PR actually added `stop`/`stop?` at the LogStash::Plugin level, but only wired the calls for the input plugins, they weren't called during the shutdown.
At the time I could not find a good reason to justify changing the shutdown sequence further and have those methods also in filters/outputs.
I haven't check the rabbitmq plugins, is it really necessary to have those in order to shutdown?

andrewvc:
@jsvd not necessary, but it would be convenient vs. setting up my own @stopping variable

jordansissel:
LGTM; didn't run the tests, though.

colinsurprenant:
WRT to the new `do_close` and `do_stop`, I am +1 on the change but I would like to suggest that we do not provide an noop implementation of `close` and `stop` in the base class, it is useless IMO and plugin developers do no read base classes, they follow examples and documentation.

instead I suggest we do something like
```ruby
def do_close
  # do whatever should be done prior to call the real plugin defined close
  close if self.respond_to? :close
end
```

same idea with `stop`

Issue #3735, opened 2015-08-14
Bringing back jar-dependencies to logstash, including offline support for it.


purbon:
Hi, 
  in an effort of having a good citizenship  in the gems world, embedding jars the way we do it nowadays in logstash, and has been done for long time in jruby projects is not optimal. To do this work there is a set of tools like https://github.com/mkristian/jar-dependencies mantained by Herr @mkristian.

In LS we removed the usage of jar-dependencies because our way of using it was not providing us with all the necessary features we aim for, we should be sure managed dependencies  are:
- [x] Downloaded and setup during a plugin installation, this could happen in many situations (for example with proxies, etc)
- [x] Enable offline installation of plugins.
- [ ] Add test in logstash to be sure jars are packaged in the plugins when the gem is build/installed/etc..

colinsurprenant:
I think at this point we have to evaluate the benefits vs the costs of trying to solve this. In particular for offline installation, I do not see how it will be possible to package plugins that depends on jars, via jar-dependencies (and Maven, ...), for self-contained offline installation. We also have to keep in perspective that only a handful of plugins depends on jars, so I am not sure if it is really worth investing much time on this at this point.

Also, other than the size issue of including a jar into a gem, which from a user-perspective has zero impact, our plugins systems allows up yo very easily update a jar version if needed and release a new plugin version. 

Am I missing somethings? Thoughts?

purbon:
@colinsurprenant I'm starting to think and work on this, and as I spoke to @suyograo this would be my five cents back to the jruby community. I will keep digging on ways to get this working for us.

I think the idea of having jars into gems is a necessary pain to suffer, so enabling the option to fetch and and work as expected is very nice feature to have in jruby projects.

colinsurprenant:
@purbon sure but what about **offline**? There might be a solution but I am not sure I see a practical/pragmatic one, but that doesn't one does not exits :P

purbon:
I don't know and I need to investigate before making my mind.

mkristian:
I mentioned on an other issue how offline packaging could work with jars as packaging container. they can have the embedded gem and its jars. to use such a jar you just need to require it.

for example jruby-openssl does use jar-dependencies but comes with the jars packed inside the gem. here jar-dependencies just ensures that only one version of jar gets loaded into the JRubyClassLoader (via ```require_jar```). this means the overall application still has control over the which version of jars used or excluded or replaced.

[...]

PR #3926, opened 2015-09-15
New gemspec for 2.0.0-beta1

suyograo:


Gemfile.jruby-1.9.lock
```diff
@@ -94,7 +94,7 @@ GEM
     http_parser.rb (0.6.0-java)
     i18n (0.6.9)
     insist (1.0.0)
-    jar-dependencies (0.1.15)
+    jar-dependencies (0.2.1)
```
ph: This scares me a bit, the CI would probably trap any problems.

suyograo: We don't use it much, only for kafka plugin right?

ph: right!

purbon: Is only jruby-kafka, but not sure it need it at the end of the day also. I think we should experiment on removing it at all and see what happen. At less try an integration test before moving forward.

ph:
LGTM, I have looked at a few gems version and all looked OK.

Gemfile.jruby-1.9.lock
purbon: from https://rubygems.org/gems/jruby-kafka it looks like jruby-kafka have all the jars into it, so I think we even don't use jar dependencies for anything this days.

ph: Agree, might be done in beta 2.

purbon: will make a PR to clean this dependencies tomorrow :+1:

suyograo: +1 to remove it in beta2

ph: @purbon can you do an upstream PR on `jruby-kafka` see https://github.com/joekiller/jruby-kafka/blob/master/jruby-kafka.gemspec#L21

Since it is declared as a runtime dependencies we will have it anyway.

purbon: makes sense, I don't think it should be runtime, at less from my research in #3735 .... I will report there.

PR #3937, opened 2015-09-17
String Interpolation skip the last character

### LINKED ISSUE #3931: Field Expression Causes Last Character to be removed
If a field expression is used in an add_field property at the end of a string, the last character will be dropped in the output i.e. it occupies upto position N-2 in the string, where N is the length of the string indexed from 0.

For example,

mutate {
    	add_field => { "Intel-Foundation" => "|%{fAffiliate}|" }
  	}

In this case, the the final pipe character will be dropped.  This can be solved by adding a character to the end of the string thus moving the expression termination point to N-3 e.g. 

mutate {
    	add_field => { "Intel-Foundation" => "|%{fAffiliate}| " }
  	}

ph:
This PR fix a problem when doing the interpolation with a string that
did not end with a fieldref but with a character. The interpolation was
ignoring the last character.

Example:
```
"%{type}|" => "syslog"
```

Fixes #3931

### DIFF
--- a/lib/logstash/string_interpolation.rb
+++ b/lib/logstash/string_interpolation.rb
@@ -46,7 +46,7 @@ module LogStash
         position = match.offset(0).last
       end
 
-      if position < template.size - 1
+      if position < template.size
         nodes << StaticNode.new(template[position..-1])
       end
 
--- a/spec/core/event_spec.rb
+++ b/spec/core/event_spec.rb
@@ -50,6 +50,10 @@ describe LogStash::Event do
         expect(subject.sprintf("%{+%s}")).to eq("1356998400")
       end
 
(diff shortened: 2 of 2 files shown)


jordansissel:
Code looks good; didn't run the tests, but the new test case covers the bug fix.

guyboertje:
@ph - do you remember why you had the `if position < template.size - 1` in the first place?

ph:
To the time machine.

@guyboertjein at the beginning I didn't have a StaticNode class and the parsing was done differently and was pretty much the same as logstash 1.4.2. I gave another look at it this morning and I found a missing test case that I just added.

guyboertje:
@ph - that is the ticket! I suspected asking that question would result in something good :-)

LGTM

PR #3938, opened 2015-09-18
rFix/cleanup require helper

ph:
This file isn't used in the code at all, It was probably used at some
point but its not the case anymore.


jordansissel:
LGTM

PR #3941, opened 2015-09-18
Remove unused files

ph:
Removed related files for compiling the grammer, we have refactored the
makefile using rake. If you want to generate a new parser you can use

`rake compile:all`


jordansissel:
LGTM.

Jenkins failed due to a network error talking to github. I'm ignoring it.

PR #3950, opened 2015-09-22
Remove deprecated config options for 2.0

suyograo:
Will also make docs nicer. This PR is just to get a feel of what it takes to delete existing deprecated stuff from core plugins. there's a lot of core tests that refer this configuration/variables so it may be hairy to check and remove them. Not sure we can do a blind "global search and delete".

### DIFF
--- a/lib/logstash/filters/base.rb
+++ b/lib/logstash/filters/base.rb
@@ -183,6 +183,12 @@ class LogStash::Filters::Base < LogStash::Plugin
     end
   end # def filter_matched
 
+  protected
+  def filter?(event)
+    # TODO: noop for now, remove this once we delete this call from all plugins
+    true
+  end # def filter?
+
   public
   def close
     # Nothing to do by default.
--- a/lib/logstash/filters/base.rb
+++ b/lib/logstash/filters/base.rb
@@ -11,6 +11,12 @@ class LogStash::Filters::Base < LogStash::Plugin
 
   config_name "filter"
 
+  config :type, :validate => :string, :default => "", :obsolete => "You can achieve this same behavior with the new conditionals, like: `if [type] == \"sometype\" { %PLUGIN% { ... } }`."
+
+  config :tags, :validate => :array, :default => [], :obsolete => "You can achieve similar behavior with the new conditionals, like: `if \"sometag\" in [tags] { %PLUGIN% { ... } }`"
+
+  config :exclude_tags, :validate => :array, :default => [], :obsolete => "You can achieve similar behavior with the new conditionals, like: `if !(\"sometag\" in [tags]) { %PLUGIN% { ... } }`"
+
   # If this filter is successful, add arbitrary tags to the event.
   # Tags can be dynamic and include parts of the event using the `%{field}`
   # syntax.
--- a/lib/logstash/filters/base.rb
+++ b/lib/logstash/filters/base.rb
@@ -183,40 +183,6 @@ class LogStash::Filters::Base < LogStash::Plugin
     end
   end # def filter_matched
 
-  protected
-  def filter?(event)
-    if !@type.empty?
-      if event["type"] != @type
-        @logger.debug? and @logger.debug("filters/#{self.class.name}: Skipping event because type doesn't match",
-                                         :type=> @type, :event => event)
-        return false
-      end
-    end
-
-    if !@tags.empty?
-      # this filter has only works on events with certain tags,
-      # and this event has no tags.
-      return false if !event["tags"]
-
-      # Is @tags a subset of the event's tags? If not, skip it.
-      if (event["tags"] & @tags).size != @tags.size
-        @logger.debug? and @logger.debug("filters/#{self.class.name}: Skipping event because tags don't match",
-                                         :tags => tags, :event => event)
-        return false
-      end
-    end
-
-    if !@exclude_tags.empty? && event["tags"]
-      if (diff_tags = (event["tags"] & @exclude_tags)).size != 0
-        @logger.debug("filters/#{self.class.name}: Skipping event because tags contains excluded tags:",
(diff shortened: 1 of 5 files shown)


suyograo:
We can take a shot at plugins in the next pass

lib/logstash/filters/base.rb
```diff
@@ -200,40 +183,6 @@ def filter_matched(event)
...
-    if !@exclude_tags.empty? && event["tags"]
-      if (diff_tags = (event["tags"] & @exclude_tags)).size != 0
-        @logger.debug("filters/#{self.class.name}: Skipping event because tags contains excluded tags:",
-                      :diff_tags => diff_tags, :exclude_tags => @exclude_tags, :event => event)
-        return false
-      end
-    end
-
-    return true
-  end
```
ph: Create this issue: https://github.com/elastic/logstash/issues/3968
for the affected plugins.

lib/logstash/outputs/base.rb
```diff
@@ -92,33 +78,4 @@ def handle_worker(event)
...
-    if !@exclude_tags.empty? && event["tags"]
-      if (diff_tags = (event["tags"] & @exclude_tags)).size != 0
-        @logger.debug? and @logger.debug("outputs/#{self.class.name}: Dropping event because tags contains excluded tags",
-                                         :diff_tags => diff_tags, :exclude_tags => @exclude_tags, :event => event)
-        return false
-      end
-    end
-
-    return true
-  end
```
ph: Created this https://github.com/elastic/logstash/issues/3969 for tracking impacted plugins.

ph:
@suyograo may I suggest we do this change in two PRs? This one could make the `filter?` and `#output?` noop method, than another PR to actually remove them this would make the changes to the plugin non blocker.

ph:
Changes look good to me, I have applied this patch to the 2.0 branch and I was able to run the test.

lib/logstash/inputs/base.rb
```diff
@@ -70,25 +47,6 @@ def initialize(params={})
...
-    # Backwards compat for the 'format' setting
-    case @format
-      when "plain"; # do nothing
-      when "json"
-        @codec = LogStash::Plugin.lookup("codec", "json").new
-      when "json_event"
-        @codec = LogStash::Plugin.lookup("codec", "oldlogstashjson").new
-    end
-
   end # def initialize
```
ph: At first I thought plugins were needed to be updated for this changes but after verification this is not the case. +1

suyograo:
@ph makes sense to do a 2 phase change. Will add the noop method

ph:
When I run the test I get this exception:
```
Failures:

  1) LogStash::Config::Mixin when validating :password should correctly copy password types
     Failure/Error: clone = subject.class.new(subject.params)
     LogStash::ConfigurationError:
       The setting `type` in plugin `fake` is obsolete and is no longer available. You can achieve this same behavior with the new conditionals, like: `if [type] == "sometype" { fake { ... } }`. If you have any questions about this, you are invited to visit https://discuss.elastic.co/c/logstash and ask.
     # ./lib/logstash/config/mixin.rb:82:in `config_init'
     # ./lib/logstash/config/mixin.rb:66:in `config_init'
     # ./lib/logstash/filters/base.rb:125:in `initialize'
     # ./spec/core/config_mixin_spec.rb:97:in `(root)'
     # ./lib/bootstrap/rspec.rb:11:in `(root)'
```
```ruby
    it "should correctly copy password types" do
      clone = subject.class.new(subject.params)
      expect(clone.password.value).to(be == secret)
    end
```

fast guess is that subject.params return a hash that contains the `type` key.

ph:
@suyograo I have created the fix for the failing test in https://github.com/elastic/logstash/pull/4011 Can you review with your changes?

With my PR you can merge your changes!

lib/logstash/filters/base.rb
purbon: I apologies if this is already public, but I'm kind of struggling finding where this was actually deprecated. Any idea where we announced it? I say it because it implies changes in configurations and would be nice to announced somehow. I understand this is the time to cleanup, just wanted to be sure we deprecated the stuff properly.

purbon: I apologies again, saw https://github.com/elastic/logstash/pull/3950/files#diff-5aea1d7958d8c2f7ff39fd1f8624acabL21 as where it was going to be deprecated! cool and good to go.

purbon:
As noted in https://github.com/logstash-plugins/logstash-filter-environment/pull/6#issuecomment-146910212 we should add very clear documentation of everything that breaks compatiblity.

lib/logstash/filters/base.rb
```diff
@@ -11,22 +11,11 @@ class LogStash::Filters::Base < LogStash::Plugin
...
-
-  # Only handle events with all of these tags.
-  # Optional.
-  config :tags, :validate => :array, :default => [], :deprecated => "You can achieve similar behavior with the new conditionals, like: `if \"sometag\" in [tags] { %PLUGIN% { ... } }`"
+  config :tags, :validate => :array, :default => [], :obsolete => "You can achieve similar behavior with the new conditionals, like: `if \"sometag\" in [tags] { %PLUGIN% { ... } }`"
 
-  # Only handle events without any of these tags.
-  # Optional.
-  config :exclude_tags, :validate => :array, :default => [], :deprecated => "You can achieve similar behavior with the new conditionals, like: `if !(\"sometag\" in [tags]) { %PLUGIN% { ... } }`"
+  config :exclude_tags, :validate => :array, :default => [], :obsolete => "You can achieve similar behavior with the new conditionals, like: `if !(\"sometag\" in [tags]) { %PLUGIN% { ... } }`"
```
jordansissel: There's also a `not in` operator we could recommend here. Thoughts?

ph:
fixed by #4011

PR #3951, opened 2015-09-22
Revert wrong commit in the Gemfile that introduced dev dependencies

purbon:
This reverts commit a6ab581979bafdf32483b7f8c126cc1ade159ee8 that introduced wrong dependencies to be in the Gemfile for master.

This should only be merged with master.

Issue #3948, opened 2015-09-21
Core test should not be dependant in plugins


purbon:
Core test should not be dependant on plugins to work. This days core test need

```

gem "logstash-filter-clone"
gem "logstash-filter-mutate"
gem "logstash-filter-multiline"
gem "logstash-input-generator"
gem "logstash-input-stdin"
gem "logstash-input-tcp"
gem "logstash-output-stdout"
```
to work, but for this test to be actually validating the pipeline behaviour it should be using mocks or internal dummy filters created for this special purpose. Otherwise when a plugin has a blocker update, test in core are not working as expected.

purbon:
First try to implement it in #3954

purbon:
@suyograo let me know if label is resonable here.

suyograo:
Made it 5.1, since 5.0 after beta1 will be bug fixes only.

purbon:
makes sense, thank for feedback Suyog, also there is another problem here and is we have a ton of Dummy plugins and this internal real plugins, my idea is we should move to something that really make our current test _independant_ of plugins, but we still have integrations test for it. 

Like this we are not blocking anyone when a big refactor happen, or any other small issue in plugins used for test. And also it makes more sense in terms of code separation.

- purbon

PR #3954, opened 2015-09-22
Cleanup usage of plugins to test Logstash-core

purbon:
Nowadays we use some plugins to test the behaviour of the logstash-core, while this is an acceptable behaviour it should be required that plugins behave as expected to run the core test.

The current idea is, clean up:
- Current plugin usage for dummy/mock plugins where need.
- Rewrite test to they don't require this plugins to be used.

Like this we'll win isolation in the test, so easy to reproduce problems without the need to think if there is a bug, or an issue introduced by the plugins used in testing.

This is an initial approach to remove this issue, next tasks related to that would be:
- Wrap group knowledge on how each parts of logstash should behave, starting by configuration, validation (config mixing) and pipeline behaviour.
- This would help us write proper test that will work as a reference.
- With the reference we can start tracking a more deep test refactor as they should be.


Gemfile.jruby-1.9.lock
```diff
@@ -46,7 +45,7 @@ GEM
       unf (>= 0.0.5, < 1.0.0)
     faraday (0.9.1)
       multipart-post (>= 1.2, < 3)
-    ffi (1.9.10-java)
+    ffi (1.9.8-java)
```
untergeek: Interesting.  I hadn't heard about a regression.

untergeek:
Why all the version number regressions in the gem versions?

purbon:
@untergeek because this PR include https://github.com/elastic/logstash/pull/3951 basically we merged a wrong Gemfile in for master. This is the reason why you see what you see.

purbon:
jenkins test this

purbon:
please jenkins, test this.

purbon:
@ph happy to get feedback from you man.

ph:
I love the idea looking into it.

purbon:
@ph thanks much appreciate it

colinsurprenant:
is the strategy to actually re-implement all these plugins as mocks?

purbon:
@colinsurprenant for some, like the generator I kinda like the idea, however there might be situations like the canceled events test where using multiline could be done in another way. I'm happy to remove as mocks as possible, feel free to point me out where they make no sense.

colinsurprenant:
re-implementing plugins as mocks feels weird and will create duplicate code and eventually not mean much as these mocks are not the real plugins and these mock could be buggy rendering the spec useless. mocks should typically not contain logic, if it does, they should be tested themselves which becomes completely redundant. 

If we decide to move forward with a strategy of eliminating the need of plugins in the core specs, I think the specs should be refactored to completely not depend on plugins logic to run. the original tests should probably be moved into integration tests. 

For example, the cancel logic instead of being tested using a multiline plugin should be rewritten to only test the cancel logic expectations in itself and completely remove the need of a plugin, not even a multiline mock.

andrewvc:
I agree with @colinsurprenant on the mocking issue. I think we should have a small list of 'golden' plugins for this purpose.

ph:
When I said I like the idea, I haven't look at all of it.
This is what I think similar to @andrewvc and @colinsurprenant 

- Remove the plugins from the core spec.
- If a plugin is needed this should be an integration tests.
- We should rewrite the tests that need plugins, they are obviously testing too much.

andrewvc:
Thinking through it a bit more, I'm thinking that we should either do something a bit closer to what @pere mentioned in terms of dummy plugins, or move some 'golden' plugins to logstash core. 

I'm strongly opposed to these plugins being mocks, I would prefer that they be actual running plugins that do something, just that they live in the same source tree.

Ideally we'd have have 3 plugins, an input (a simple version of the generator plugin), a filter (maybe a super simple one that just adds a tag, and an output (maybe one that just counts received messages?). 

I think that sits at the right abstraction level in terms of being useful and 'real world', without us writing a big chunk of real code that users depend on.

colinsurprenant:
- I agree with the idea that core tests should not depend on plugins.
- all tests that use plugins to test core functionalities like using the multiline plugin to test the cancellation logic should be rewritten to not depend on a plugin. 
- the only place where a dummy plugin should be required is for the plugins discovery/loading code.
- all tests that uses plugins to test core functionalities should be moved as "integration" tests. These tests are still "valid" but don't belong in the core tests.
- as we ramp up better core tests, we will still depend on the "integrations" tests being run to test core.

purbon:
In the benefit of keeping this a sane discussion I would recommend we focus discussions in the issue, check #3948 for that.

Just a matter of context, I start this as an initial approach to have a non plugin dependant tests for core, by using mocks, something that I still believe is not a bad thing. This mocks used here are very simple and easy to understand. Although I do see here some strong opinions about following this path for this solution.

I would rather keep this open, but start also doing work on refactoring the test as requested here, just to be aware:

* this means a lot more work and per se means this problem is staying here longer
* the new tests could be not easy to understand as they testing the same thing by reviewers, requiring then double effort from them.

purbon:
@andrewvc using running plugins makes no sense, the problem is here because we've a dependency with a third party component, what exactly mocks/stubs are for.

purbon:
@ph the plugins are needed because the test were poorly written, but they are not integration test at all. They are more functional tests.

andrewvc:
I don't think I did a good job of communicating clearly.

@purbon what I'm trying to say is that our integration tests shouldn't be on things that are gem dependencies (like real world plugins). But rather lightweight plugins that are part of the core to prevent issues when we bump version numbers. We need a way to test the pipeline without using 'real' plugins, but while using something with a bit more reality than a mock.

I agree that we should use mocks in places where we are testing unit behavior, as @colinsurprenant has pointed out.

purbon:
@andrewvc so what you ask for is to move the code I made for this plugins, named mocks because they mock a real plugin to a gem? then is ok for you? The plugins I wrote here are lightweight plugins that do a simple work, give them another name if mock is not appropriate.

The actual problem here the test uses plugins a lot to do it work, see the sample method. and before going for a full refactor I started this as an initial step. And they are not integration test, they test behaviour, but using plugins (what is wrong, but we could use this lightweight dummy plugins). Is my idea more clear now?

jordansissel:
@purbon +1 on having dummy plugins used for testing core behaviors where needed. I agree 'mock' isn't the right word, but reading https://github.com/purbon/logstash/blob/feature/cleanup_plugin_test_deps/spec/support/plugins_helper.rb makes it clear the purpose (some fixture classes used during testing). +1 :)

purbon:
agree naming is not probably appropriate, my bad on this.

jordansissel:
@purbon hah, no worries. I confess that I never remember the definitions of mocks/doubles/fakes/stubs/etc. :)

andrewvc:
@purbon I really have to apologize. I mixed up this thread with something else and thought I was commenting on a design issue instead of a PR. I've got too many balls in the air doing refactors and have been moving too fast. I apologize for not devoting more focus to reading this PR.

I reviewed all the code. I love the approach, and think it all looks great. I hereby retract all my comments.

ph:
@purbon your comment made me realize this.

This PR improve the situation *right now* which make sure we can run the core test without having external dependency, I am OK with it.

colinsurprenant:
For the record, I still disagree with the approach. It removes/weakens still valid tests which could have been simply moved into integration tests, replaces plugins with dummy/fake/mock plugins which is rather useless and redundant code, does not address the problems properly which would be to rewrite the tests to correctly test the expected core behaviour without a plugin harness.

I am at peace with my disagreement and seems to not be able to have a convincing argument so please feel free to reach consensus without me.

guyboertje:
I think this approach is a good first step on a long journey.  Basically, if creating correct tests for a codebase is hard then that usually means the code needs refactoring.

@purbon - consider stubbing the `Logstash::Plugin.lookup` method to return a suitable Mock rather than rewriting the config strings or perhaps a combination of both.

On the subject of Unit vs Integration tests - imagine unit testing a piece of code that interacts with an external HTTP API, one would mock the external server and code it (with suitable stimuli) to return various 2XX, 3XX, 4XX, 5XX status codes so one can verify that one's code reacts correctly.  Later, for integration testing, if the external server had test endpoints, you would use that.  Stripe has a great test system, if you use certain credit card numbers in your call, their end will respond with an associated error (e.g. Insufficient funds) which allows one to verify that all the edge cases are covered.
Why did I mention Stripe? Because, in LS core, to correctly verify some core behaviours the dependencies albeit real or mocked have to react in a predicable way.

Having the odious task of maintaining both the Plugins and their Mocks should provide ample incentive to refactor for minimum burden.

purbon:
I like this idea @guyboertje,

> consider stubbing the Logstash::Plugin.lookup method to return a suitable Mock rather than rewriting > the config strings or perhaps a combination of both

It would have been awesome to get this ideas in an issue like #3948 but I will try to resume them there for posterity :+1: 

- cheers

purbon:
removing the stalled tag we reach consensus about the approach. Going to rename the dummy plugins to have a more descriptive name.

ph:
@purbon its that ready to review?

ph:
@purbon Not sure of the state of this PR, I need rebasing feel free to reassign it to me went its ready.

suyograo:
Closing for now. We need to revisit this strategy.

PR #3967, opened 2015-09-25
Update Gemfile.lock in the 2.0 branch to use last released plugins

purbon:
Fixes broken http://build-eu-00.elastic.co/view/LS%202.0/job/Logstash_Default_Plugins_20/247/


Gemfile.jruby-1.9.lock
```diff
@@ -724,4 +724,4 @@ DEPENDENCIES
   rspec (~> 3.1.0)
   rubyzip (~> 1.1.7)
   simplecov
-  stud (~> 0.0.19)
+  stud (~> 0.0.21)
```
jsvd: we should force this to be ~> 0.0.22, that's where the new stud stop stuff was added

purbon: makes sense, good point.

jsvd:
LGTM

purbon:
Still not merge it because looks like logstash-output-statsd is failing under the way we run the defaults plugins test.

PR #3978, opened 2015-09-30
New plugin config attribute :obsolete

### LINKED ISSUE #3977: Create an obsolete syntax for config options
We need to create a special `obsolete` syntax/option for `config`. The idea is to use this for config options which have been marked deprecated. When `obsolete` is used on a config, it passes the config check, but LS will fail to start and it will provide a message that it has been deprecated and no longer relevant. This provides better feedback for users of deprecated configs.

Before a major version, config changes which are backward incompatible will be marked `deprecated`. A major version will remove the code usage, remove documentation, but keep the config option with the `obsolete` tag. The next major version will delete the config completely. 

Proposed by @andrewvc during the ES http default work. Ref issue from a user: https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/258

jordansissel:
An obsolete setting is one that will cause a configuration error if it
is used.

The purpose of `:obsolete` is to help inform users when a setting has
been completely removed. The lifecycle of a plugin setting is now 4
phases: available, deprecated, obsolete, deleted.

Available is the default, and deprecated remains the same as it was
(logging a warning). The new obsolete will cause a configuration error
if such a setting is used. Then later, we can finally delete the config
setting after it's been obsolete for some time.

Fixes #3977

### DIFF
--- a/docs/asciidoc/static/include/pluginbody.asciidoc
+++ b/docs/asciidoc/static/include/pluginbody.asciidoc
@@ -439,7 +439,7 @@ endif::encode_method[]
 ==== Configuration Parameters
 [source,ruby]
 ----------------------------------
-  config :variable_name, :validate => :variable_type, :default => "Default value", :required => boolean, :deprecated => boolean
+  config :variable_name, :validate => :variable_type, :default => "Default value", :required => boolean, :deprecated => boolean, :obsolete => string
 ----------------------------------
 The configuration, or `config` section allows you to define as many (or as few)
 parameters as are needed to enable Logstash to process events.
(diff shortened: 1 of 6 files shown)


docs/asciidoc/static/include/pluginbody.asciidoc
```diff
@@ -439,7 +439,7 @@ endif::encode_method[]
 ==== Configuration Parameters
 [source,ruby]
 ----------------------------------
-  config :variable_name, :validate => :variable_type, :default => "Default value", :required => boolean, :deprecated => boolean
+  config :variable_name, :validate => :variable_type, :default => "Default value", :required => boolean, :deprecated => boolean, :obsolete => string
```
untergeek: Is `:obsolete` supposed to be `boolean` like the other options?

untergeek: Nevermind.  I see that you're populating `:obsolete` with a value later.

i-see-what-you-did-there

untergeek:
LGTM

andrewvc:
LGTM , and thanks for implementing this @jordansissel !

docs/asciidoc/static/include/pluginbody.asciidoc
jordansissel: Indeed. Also, I noticed that the docs for `:deprecated` are incorrect. It's not a boolean (although it can be), it's a String. I opted not to fix that in this PR to avoid confusing things, oops ;P

PR #3986, opened 2015-10-02
Fix master after we published all the 2.0 plugins

suyograo:
Today we can't install any 2.0 plugins on master because of the logstash-core constraint in plugin's gemspec

```
s.add_runtime_dependency "logstash-core", "~> 2.0.0.snapshot"
```

This `~` allows only snapshot versions. This is a stop-gap for now so we can get tests running on master. After the beta2 we will be changing the constraint to ` > 2.0.0` and do a mass update, which will allow us to make master's version `3.0.0.dev`. 

This PR will un-hose the master branch

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@
 # If you modify this file manually all comments and formatting will be lost.
 
 source "https://rubygems.org"
-gem "logstash-core", "2.0.0.dev", :path => "."
+gem "logstash-core", "2.0.0.snapshot100", :path => "."
 gem "file-dependencies", "0.1.6"
 gem "ci_reporter_rspec", "1.0.0", :group => :development
 gem "simplecov", :group => :development
--- a/Gemfile.jruby-1.9.lock
+++ b/Gemfile.jruby-1.9.lock
@@ -1,9 +1,10 @@
 PATH
   remote: .
   specs:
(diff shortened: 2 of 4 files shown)


ph:
So the constraints should be `>= 2.0.0` and `< 3.0.0` Correct?

suyograo:
@ph the constraint in the plugins should just be `>= 2.0.0` because master will become `3.0.0` version

ph:
If we don't define a upper bound the plugins will also work in 3.0, which might also require a change in  the api?

suyograo:
@ph `>= 2.0.0 and < 3.0.0` is fine. Master will actually become `3.0.0.dev`

suyograo:
which weirdly works and used to work when the constraint was `'>= 1.4.0', '< 2.0.0'`

ph:
@suyograo  oh well, `~>` on the snapshot version is also weird ;)

I'll update all the dependency of the plugins and mass publish them and fix the version of the master branch to be 3.0.0.dev.

suyograo:
@ph +1. We can close this PR once you do that

ph:
I'll cherry pick the removal of elasticsearch_http

ph:
fix by https://github.com/elastic/logstash/pull/3999#issuecomment-146556992

PR #3990, opened 2015-10-05
Added clearer message when a filter crash

### LINKED ISSUE #3983: When a filter is raising an exception, the filter worker dies.
When there is an exception in a filter worker loop the thread dies and logstash just stop doing anything.
We should make sure we give a more explicitly error in the log.

```
:message=>"Exception in filterworker", "exception"=>#<NoMethodError: undefined method `<' for nil:NilClass>, "backtrace"=>["(eval):3636:in `cond_func_144'", "org/jruby/RubyArray.java:1613:in
```

ph:
When the filter workers crash Logstash will stop consuming any events,
Since we don't have a dead letter queue yet, we can't restart the
workers.

Fixes #3983

### DIFF
--- a/lib/logstash/pipeline.rb
+++ b/lib/logstash/pipeline.rb
@@ -221,11 +221,19 @@ class LogStash::Pipeline
           break
         end
       end
-    rescue => e
-      @logger.error("Exception in filterworker", "exception" => e, "backtrace" => e.backtrace)
+    rescue Exception => e
+      # Plugins authors should manage their own exceptions in the plugin code. 
+      # But if an exception is raised up to the worker thread their are mostly 
+      # fatal and logstash can't recover from this situation.
+      #
+      # Users need to check their configuration or see if there is a bug in the
+      # plugin.
+      @logger.error("Exception in filterworker, the pipeline stopped processing new events, please check your filter configuration and restart Logstash.",
+                    "exception" => e, "backtrace" => e.backtrace)
+      raise
+    ensure
+      @filters.each(&:do_close)
     end
-
-    @filters.each(&:do_close)
   end # def filterworker
 
   def outputworker


ph:
Resume from discussion with @guyboertje  from zoom.

rescue `Exception` and reraise, make a difference in the message for standard error and exception for future debugging.

It's a best effort, if logstash goes OOM we might not be able to log at all.

ph:
I think, we should do something like this, since in all scenarios the worker is dead and we should try to close it gracefully. What do you think?

```ruby
begin
    ##### 
rescue StandardError, Exception => e
  @logger.error("StandardError Exception in filterworker, the pipeline stopped processing new events, please check your filter configuration and restart Logstash.",
                    "exception" => e, "backtrace" => e.backtrace)
   raise
 ensure
   filters.each(&:do_close)
 end
```

guyboertje:
I like the ensure part
However as `StandardError` is a subclass of `Exception` it is not needed.
Consider adding a large comment in the code explaining why Exception is caught and that, in general, most workers will already have a begin ... rescue => e structure in place so the only exceptions that are rescued here pretty much fatal.  WDYT?

ph:
@guyboertje agree, will change and update.

ph:
@guyboertje updated with your comments

guyboertje:
Perfect! LGTM

jsvd:
why not raise a fatal exception? data is lost anyway we might as well exit logstash

ph:
@jsvd I am ok with this `.fatal`, data is lost and logstash can't recover.

jsvd:
though this might be out of scope for this pr. just considering if we should plainly say "logstash died but won't exit. do it yourself please"

guyboertje:
@jsvd - line 233 does a raise

jsvd:
oh dear. I need to take a break. LGTM

ph:
@jsvd :coffee:

PR #4011, opened 2015-10-09
Remove deprecated config options for 2.0

ph:
Two small problems with the `obsolete` field.

1) We can use `Plugin#params` to retrieve the option hash, the params method should not return the `obsolete` options. Lets say that you use `#params` method to create a new instance of the plugin, the new instance validation will fails even if the original option hash did not contain any options marked as obsolete. Since we don't have any context of a `dirty options`, where a user actually use that key, it's easier to not return any obsoletes values.

2) Also we were defining the obsolete instance value with `instance_variable_set`

The `#params` method was only used in the context of a password test.

Ref: #3950

### DIFF
--- a/lib/logstash/config/mixin.rb
+++ b/lib/logstash/config/mixin.rb
@@ -110,6 +110,16 @@ module LogStash::Config::Mixin
         I18n.t("logstash.agent.configuration.invalid_plugin_settings")
     end
 
+    # We remove any config options marked as obsolete,
+    # no code should be associated to them and their values should not bleed
+    # to the plugin context.
+    #
+    # This need to be done after fetching the options from the parents classed
+    params.reject! do |name, value|
+      opts = self.class.get_config[name]
+      opts.include?(:obsolete)
+    end
+
     # set instance variables like '@foo'  for each config value given.
     params.each do |key, value|
       next if key[0, 1] == "@"
--- a/lib/logstash/filters/base.rb
+++ b/lib/logstash/filters/base.rb
@@ -11,22 +11,11 @@ class LogStash::Filters::Base < LogStash::Plugin
 
   config_name "filter"
 
-  # Note that all of the specified routing options (`type`,`tags`,`exclude_tags`,`include_fields`,
-  # `exclude_fields`) must be met in order for the event to be handled by the filter.
+  config :type, :validate => :string, :default => "", :obsolete => "You can achieve this same behavior with the new conditionals, like: `if [type] == \"sometype\" { %PLUGIN% { ... } }`."
 
-  # The type to act on. If a type is given, then this filter will only
-  # act on messages with the same type. See any input plugin's `type`
-  # attribute for more.
-  # Optional.
-  config :type, :validate => :string, :default => "", :deprecated => "You can achieve this same behavior with the new conditionals, like: `if [type] == \"sometype\" { %PLUGIN% { ... } }`."
+  config :tags, :validate => :array, :default => [], :obsolete => "You can achieve similar behavior with the new conditionals, like: `if \"sometag\" in [tags] { %PLUGIN% { ... } }`"
 
-  # Only handle events with all of these tags.
-  # Optional.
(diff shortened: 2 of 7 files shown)


spec/core/plugin_spec.rb
```diff
@@ -120,4 +120,24 @@ class LogStash::Filters::Stromae < LogStash::Filters::Base
       end
     end
   end
+
+  context "when accessing the #params" do
```
jsvd: I'd just change this to `describe "#params"`

lib/logstash/config/mixin.rb
```diff
@@ -85,6 +85,11 @@ def config_init(params)
       end
     end
 
+    params.reject! do |name, value| 
+      opts = self.class.get_config[name]
+      opts.include?(:deprecated)
```
jsvd: shouldn't this be obsolete?

suyograo: This aint gonna work. Check for `obsolete` happens before `params.reject!` is executed. This will bail once params with `obsolete` is encountered before this line. 

I think the correct issue (or place to fix this) is when you do `subject.params`, default params are created. This includes the obsolete config options with default values.

ph: Ahh, last week I really needed more sleep, looking into this.

spec/core/plugin_spec.rb
ph: +1

jsvd:
LGTM

lib/logstash/filters/base.rb
```diff
@@ -11,22 +11,11 @@ class LogStash::Filters::Base < LogStash::Plugin
...
-
-  # Only handle events with all of these tags.
-  # Optional.
-  config :tags, :validate => :array, :default => [], :deprecated => "You can achieve similar behavior with the new conditionals, like: `if \"sometag\" in [tags] { %PLUGIN% { ... } }`"
+  config :tags, :validate => :array, :default => [], :obsolete => "You can achieve similar behavior with the new conditionals, like: `if \"sometag\" in [tags] { %PLUGIN% { ... } }`"
 
-  # Only handle events without any of these tags.
-  # Optional.
-  config :exclude_tags, :validate => :array, :default => [], :deprecated => "You can achieve similar behavior with the new conditionals, like: `if !(\"sometag\" in [tags]) { %PLUGIN% { ... } }`"
+  config :exclude_tags, :validate => :array, :default => [], :obsolete => "You can achieve similar behavior with the new conditionals, like: `if !(\"sometag\" in [tags]) { %PLUGIN% { ... } }`"
```
jordansissel: Let's recommend `"sometag" not in [tags]` instead of `!(.... in ....)`

lib/logstash/outputs/base.rb
```diff
@@ -12,19 +12,11 @@ class LogStash::Outputs::Base < LogStash::Plugin
...
 
-  # Only handle events with all of these tags.
-  # Optional.
-  config :tags, :validate => :array, :default => [], :deprecated => "You can achieve similar behavior with the new conditionals, like: `if \"sometag\" in [tags] { %PLUGIN% { ... } }`"
+  config :tags, :validate => :array, :default => [], :obsolete => "You can achieve similar behavior with the new conditionals, like: `if \"sometag\" in [tags] { %PLUGIN% { ... } }`"
 
-  # Only handle events without any of these tags.
-  # Optional.
-  config :exclude_tags, :validate => :array, :default => [], :deprecated => "You can achieve similar behavior with the new conditionals, like: `if !(\"sometag\" in [tags]) { %PLUGIN% { ... } }`"
+  config :exclude_tags, :validate => :array, :default => [], :obsolete => "You can achieve similar behavior with the new conditionals, like: `if !(\"sometag\" in [tags]) { %PLUGIN% { ... } }`"
```
jordansissel: use "not in"

jordansissel:
Code looking good. Left some notes about the doc text for some obsolete settings.

ph:
@jordansissel I will update thanks!

suyograo:
@ph LGTM, please squash before merging.

PR #4024, opened 2015-10-13
add before_stop shutdown hook and refactor inflight reporter to use it in agent

### LINKED ISSUE #4023: inflight reporter should not run in specs
it is also interfering with specs that mocks `Time.now` where the inflight reporter can do the Time.now call before the spec.

colinsurprenant:
fixes #4023

### DIFF
--- a/lib/logstash/agent.rb
+++ b/lib/logstash/agent.rb
@@ -128,14 +128,14 @@ class LogStash::Agent < Clamp::Command
         @logger.warn(I18n.t("logstash.agent.sigint"))
         Thread.new(@logger) {|logger| sleep 5; logger.warn(I18n.t("logstash.agent.slow_shutdown")) }
         @interrupted_once = true
-        pipeline.shutdown
+        shutdown(pipeline)
       end
     end
 
     # Make SIGTERM shutdown the pipeline.
     sigterm_id = Stud::trap("TERM") do
       @logger.warn(I18n.t("logstash.agent.sigterm"))
-      pipeline.shutdown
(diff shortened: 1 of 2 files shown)


lib/logstash/agent.rb
```diff
@@ -174,6 +174,13 @@ def execute
     Stud::untrap("TERM", sigterm_id) unless sigterm_id.nil?
   end # def execute
 
+  def shutdown(pipeline)
+    pipeline.shutdown do
+      InflightEventsReporter.logger = @logger
+      InflightEventsReporter.start(pipeline.input_to_filter, pipeline.filter_to_output, pipeline.outputs)
+    end
+  end
```
ph: +1

lib/logstash/pipeline.rb
```diff
@@ -222,9 +228,9 @@ def filterworker
         end
       end
     rescue Exception => e
-      # Plugins authors should manage their own exceptions in the plugin code. 
-      # But if an exception is raised up to the worker thread their are mostly 
-      # fatal and logstash can't recover from this situation.
+      # Plugins authors should manage their own exceptions in the plugin code
+      # but if an exception is raised up to the worker thread they are considered
+      # fatal and logstash will not recover from this situation.
```
ph: hehe I saw what you did here.

ph:
LGTM, I am ok to merge this!

@colinsurprenant might be worth it to create a test if no block is given to the shutdown method?
It can be done after, I don't consider this to block the merge because we are testing this path with our devutils helpers.

PR #4027, opened 2015-10-14
Freeze lockfile for JrJackson

suyograo:


ph:
We need to change the gemspec too?

suyograo:
@ph already merged. I had to jump through hoops to get the lock file generated :) https://github.com/elastic/logstash/blob/2.0/logstash-core.gemspec#L42

ph:
oh well LGTM :dancers:

PR #4030, opened 2015-10-14
Adding in link to codec list

### LINKED ISSUE #4029: Link to available codecs from config docs
It'd be useful if we had a link [here](https://github.com/elastic/logstash/blob/master/docs/asciidoc/static/configuration.asciidoc) in the codec section that pointed over to the list of available codes, [here](https://www.elastic.co/guide/en/logstash/current/codec-plugins.html).

markwalkom:
Fixes #4029


suyograo:
LGTM

PR #4036, opened 2015-10-14
Bump version to 2.0 beta2

suyograo:


jsvd:
looking at beta1, the Gemfile should also be updated, no?

suyograo:
@jsvd yep, its a 2 step process. Get this in, then release a logstash-core with this version and then update Gemfile and .lock file

jsvd:
LGTM

PR #4039, opened 2015-10-15
Adds Breaking Changes section.

palecur:


docs/asciidoc/static/breaking-changes.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+== Breaking changes
+
+Version 2.0 of Logstash has some changes that are incompatible with previous versions of Logstash. This section discusses 
+what you need to be aware of when migrating to this version.
+
+[float]
+=== Elasticsearch Output Default
+
+Starting with the 2.0 release of Logstash, the default Logstash output for Elasticsearch is HTTP. To use the `node` or
+`transport` protocols, download the {logstash}plugins-outputs-elasticsearch_java.html[Elasticsearch Java plugin]. The
```
suyograo: can we add instructions on how to install it? See details here: https://www.elastic.co/blog/logstash-2-0-0-beta1-released "Installing java clients for Elasticsearch 1.x:"

docs/asciidoc/static/breaking-changes.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+Starting with the 2.0 release of Logstash, the default Logstash output for Elasticsearch is HTTP. To use the `node` or
+`transport` protocols, download the {logstash}plugins-outputs-elasticsearch_java.html[Elasticsearch Java plugin]. The 
+Logstash HTTP output to Elasticsearch now supports sniffing.
+
+==== Configuration Changes
+
+The Elasticsearch output plugin configuration has the following changes:
+
+* The `host` configuration option is now `hosts`
+* New options: `bind_host`, `bind_port`, `cluster`, `embedded`, `embedded_http_port`, `port`, `sniffing_delay
```
suyograo: Also `port` is gone. hosts string can specify ports with `:` and can be multiple hosts like `myserver:9200`

docs/asciidoc/static/breaking-changes.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+* The `node_name` and `protocol` options have been moved to the `elasticsearch_java` plugin
+
+[float]
+=== Kafka Output Configuration Changes
+
+The 2.0 release of Logstash includes a new version of the Kafka output plugin with significant configuration changes.
+Please compare the documentation pages for the 
+https://www.elastic.co/guide/en/logstash/1.5/plugins-outputs-kafka.html[Logstash 1.5] and
+https://www.elastic.co/guide/en/logstash/2.0/plugins-outputs-kafka.html[Logstash 2.0] versions of the Kafka output plugin 
+and update your configuration files accordingly.
```
suyograo: Also can we add a section on configs that were removed common to all inputs, filters and output. See here: https://github.com/elastic/logstash/blob/master/CHANGELOG.md#input. We can mention they were already deprecated

suyograo: Also need a breaking change section on default workers. 

Now, out of the box, the default value of the `filter_workers` setting will be set to half of the CPU cores of the machine. Increasing the workers provides parallelism in filter execution which is crucial when doing heavier processing like complex grok patterns or the useragent filter.

Note that, as before, this setting can be changed with the `w` flag, e.g. `bin/logstash -f test.config -w 1`

palecur: I though about adding that, but it doesn't seem like a breaking change, so
I decided to reserve it for the release notes. Should it go here instead?

docs/asciidoc/static/breaking-changes.asciidoc
palecur: Isn't it installed just like any other plugin? We've already got docs on
plugin management -- we can link to them from here.

suyograo: It is, but the important thing to highlight here is the version string. `--version 1.x` to be compatible with 1.x ES cluster and `--version 2.x` to be compatible with 2.x ES cluster

docs/asciidoc/static/breaking-changes.asciidoc
suyograo: yeah, its not breaking per se, but the default is changed from prev version, so i thought it be worthwhile to list here

palecur:
Docs have been revised to include feedback.

suyograo:
LGTM

PR #4045, opened 2015-10-15
fix bin/bundle command

colinsurprenant:
invoking `bin/bundle` does not use the correct logstash vendor path. 

with this fix invoking `bin/bundle` will work as intended using the correct `Gemfile`, `Gemfile.jruby-1.9.lock` and `vendor/bundle/...` path


ph:
LGTM

Issue #3451, opened 2015-06-16
stalled outputs will prevent proper shutdown


colinsurprenant:
This is a meta issue to track related shutdown issues.

--------------------------------------------------------
### Problem
Logstash refuses to stop if the output is stalled.

### Details
If the output plugins are stalled and not popping events from the `@filter_to_output` and back pressure propagates, upon shutdown request the `LogStash::SHUTDOWN` event will never be pushed because queues will be full and so plugins shutdown will not happen.  

this relates to shutdown refactor proposals in #3211 and #2477 

### Related issues:

- [ ] exception in output thread silently locks up logstash #2130
- [ ] outputs can crash silently, locking up entire pipeline #2545
- [ ] Logstash stalled/blocked - 1.5.0rc2 #2846
- [ ] 1.5.2 fails to shutdown #3575
- [ ] Service never stop #3520
- [ ] logstash 1.5.0 freeze #3440
- [ ] Logstash wont kill after rabbit problem #3203

jsvd:
This is a duplicate of https://github.com/elastic/logstash/issues/2130

colinsurprenant:
#2130 is really focused on the exception handling and this ticket describes the general condition that leads to preventing proper shutdown where "stalled outputs" can be for any number of reason, one of which is that the output plugin crashed. I think we should keep this ticket open because I believe that the proper fix is not limited to improving exception handling...

purbon:
In #3491 another explanation of this had raised, there are plugins having an infinite retry option, something that is very nice, but in case of having a target system down, it raises this issue as there are no more events being pop of the queues causing shutdown to be impossible.

I add this explanation here as I will close #3491 as is a duplicate of this issue, but knowing infinite retries are also another source of this is important to notice here.

colinsurprenant:
@purbon ok, perfect. yes, this is exactly the general condition described here where the outputs are *stalled*. they can be stalled for different reasons like with timeouts or slowdowns communicating with the external destination or stalled on retries. 

So the general condition of stalled output plugins need to be dealt with to prevent proper shutdown but also, specific plugins doing infinite retries should be fixed to add a limit on the number of retries.

purbon:
@colinsurprenant I added this to the redis-output https://github.com/logstash-plugins/logstash-output-redis/pull/9 but I see @jordansissel having concerns with data loss.

imsaar:
Is there a work around for this problem so get out of this condition before the 1.6 release?

[...]

Issue #3693, opened 2015-08-04
Persistence, Next Steps


andrewvc:
This discusses next steps in persistence, continuing from #2609 .

Summing up a conversation @colinsurprenant and @suyograo had this morning.

Colin's work on [Jruby MMAP Queues](https://github.com/colinsurprenant/jruby-mmap-queues) is going well. I took the time to look at some other queues out there, including deep dives in both [Tape](https://github.com/substack/tape) and [Chronicle](https://github.com/OpenHFT/Java-Chronicle). Both have their issues, and I fully agree with @colinsurprenant 's approach with the custom queues.

I performed my own work on pipeline level refactors to deal with latency and input semantics related to persistence. It was agreed during the meeting that we should introduce pluggable pipelines to test out new concepts and experiment with them. Earlier, pluggable queues had been discussed, but after looking into it the Queue tends to be tightly bound to the pipeline, so that is the logical place to make a pluggable abstraction.

I made a number of notes myself on a new pipeline prototype that focusses on reducing latency while improving throughput. I will provide notes on that below in a comment once I have written out all my thoughts.

andrewvc:
This document builds on the discussion and work started by many others on the logstash team and attempts to tackle some of the pipeline level concerns (vs. storage specific implementations) that impart persistence. This document starts from the premise that **we want to fully process each message at-least-once, or optionally at-most-once**, and proposes a system that can be configurably at-least-once or at-most-once, while maintaining good performance. Additionally, adding in persistence in a future-friendly way opens the door for a number of new features. Remember [you can't have exactly once delivery](http://bravenewgeek.com/you-cannot-have-exactly-once-delivery/).

The key to moving forward is reframing our relationship to Logstash inputs. Let's examine the relationship our inputs have with logstash:

* **Syslog:* Chucks messages at logstash, assumes that LS delivered them and that once its in the TCP buffer on the client everything's good
* **Kafka:* When using the [simple consumer](https://cwiki.apache.org/confluence/display/KAFKA/0.8.0+SimpleConsumer+Example) it can be configured to support [application-level acks](https://kafka.apache.org/08/design.html#semantics) after processing (though logstash doesn't do this today). Additionally, supports linear reads through partitions for in-order processing.

## Why Personality Matters for Persistence

[...]

PR #4051, opened 2015-10-16
add shutdown controller to force exit on stalled shutdown

jsvd:
* start logstash with --force-shutdown to force exit on stallled shutdown
* stall detection kicks in when SIGTERM/SIGINT is received
* check if inflight event count isn't going down and if there are blocked/blocking plugin threads
* abort logstash if a stall is detected and --force-shutdown is enabled

fixes https://github.com/elastic/logstash/issues/3451

NOTE TO REVIEWERS: I'm still working on the tests for the ShutdownController

To experiment with the shutdown controller use pipelines like
`bin/logstash -e "input { generator { count => 10000 } } filter { ruby { code => 'sleep 5' } } output { tcp { port => 3333 host => localhost workers => 2} }" -w 2 --force-shutdown`

or 
`bin/logstash -e 'input { generator { } } filter { ruby { code => "sleep 10000" } } output { stdout { codec => dots } }' -w 1 --force-shutdown`

With and without "--force-shutdown"


jsvd:
@colinsurprenant what do you think about this? I had this branch getting bit rot in my local git

lib/logstash/agent.rb
```diff
@@ -50,6 +50,11 @@ class LogStash::Agent < Clamp::Command
     I18n.t("logstash.agent.flag.configtest"),
     :attribute_name => :config_test
 
+  option "--[no-]force-shutdown", :flag,
```
jordansissel: not surethe flag really indicates the behavior. Something like `--allow-unsafe-shutdown` ?

lib/logstash/dead_letter_post_office.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def self.destination=(destination)
+    logger.info("Dead letter events will be sent to \"#{destination.location}\".")
+    @destination = destination
+  end
+
+  def self.<<(events)
+    events = [events] unless events.is_a?(Array)
+
+    events.each do |event|
```
jordansissel: the << api feels weird to accept an array partly because `Array#<<` appends an object to an array, not concatonating an array to another array (that uses Array#+)

jordansissel: Maybe call `<<` instead by `post` ? ("posting a dead letter")

lib/logstash/dead_letter_post_office.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    class Base
+      def location; end
+      def <<(event); end
+    end
+
+    class File < Base
+
+      START_TIME = Time.now
+      DUMP_PATH = ::File.join("/tmp", "dump.#{START_TIME.strftime("%Y%m%d%H%M%S")}.log")
```
jordansissel: This should probably use Dir.tmpdir? 

If this path is needed to survive reboots, then we shouldn't use the 'temp' directory because on some operating systems (CentOS, for example, I think), /tmp is wiped on reboot.

lib/logstash/dead_letter_post_office.rb
```diff
@@ -0,0 +1,55 @@
...
+  def self.logger=(logger)
+    @logger = logger
+  end
+
+  def self.logger
+    @logger ||= Cabin::Channel.get(LogStash)
+  end
+
+  def self.destination=(destination)
+    logger.info("Dead letter events will be sent to \"#{destination.location}\".")
```
jordansissel: use structured logging:

```
logger.info("Setting dead letter path", :path => destination.location)
```

lib/logstash/dead_letter_post_office.rb
```diff
@@ -0,0 +1,55 @@
...
+      event.cancel
+      @destination << event
+    end
+  end
+
+  module Destination
+
+    class Base
+      def location; end
+      def <<(event); end
```
jordansissel: same concern for << here, kinda, though the semantics seem different

`DeadLetterPostOffice#<<` will accept an array and do `#each ... @destination << object ...` but this one takes a single object and does not iterate over it.

So two things, I suppose:
1) Different parameter handling of the class method `#<<` and the `Destination::Base#<<` feels weird
2) The method `<<` feels like the wrong one to use.

Thoughts?

lib/logstash/dead_letter_post_office.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    class File < Base
+
+      START_TIME = Time.now
```
jordansissel: This time is the time this line of code is parsed in ruby, is that ideal? What's the goal of this constant-that-changes-on-every-run ?

lib/logstash/pipeline.rb
```diff
@@ -309,4 +319,31 @@ def flush_filters_to_output!(options = {})
...
+      data["outputs"] << [output.inspect, output.worker_queue.size]
+    end
+
+    data["total"] = data["input_to_filter"] + data["filter_to_output"] +
+                    data["outputs"].map(&:last).inject(0, :+)
+    data
+  end
+
+  def dump
+    dump = []
```
jordansissel: this method name shadows the name of the method. Can we change one of the names? (method or variable) ?

lib/logstash/plugin.rb
```diff
@@ -59,6 +59,11 @@ def inspect
     end
   end
 
+  def dead_letter(event)
+    return unless event.is_a?(LogStash::Event)
+    LogStash::DeadLetterPostOffice << event
```
jordansissel: I'm not sure this needs to be global. We already provide the 'logger' to each plugin instance, and we could provide a DeadLetter::Destination to each plugin instance at creation-time.

Thoughts? My preference is to try and avoid global singletons like this so that we can more easily test via dependency injection and also perhaps have more than one DeadLetteRPostOffice in a single program.

colinsurprenant:
I did not look into the details of the PR but does having a `--force-shutdown` makes sense conceptually? How will a logstash user decide to use this or not? Or from another angle, does it even make sense to *support* shutdown stall? here by *support* I mean that we acknowledge that this is an acceptable and possible behaviour. Instead shouldn't we just fix the shutdown stall problem and guarantee that when you trigger a shutdown, it will shutdown (whatever that mean but ideally without data loss).

lib/logstash/dead_letter_post_office.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def self.<<(events)
+    events = [events] unless events.is_a?(Array)
+
+    events.each do |event|
+      logger.warn("dead letter received!", :event => event.to_hash)
+      event.tag("_dead_letter")
+      event.cancel
+      @destination << event
```
jordansissel: What's the thread-safety here? Multiple plugins could be doing this simultaneously. It's possible `@file.puts` is not threadsafe (further down in your File destination), so we might want to protect this with some kind of lock.

locales/en.yml
```diff
@@ -187,3 +187,8 @@ en:
         debug: |+
           Most verbose logging. This causes 'debug'
           level logs to be emitted.
+        force_shutdown: |+
+          Force logstash to exit during shutdown even
+          if there are still inflight events in memory.
+          By default, logstash will refuse to quit until all
+          received events have been pushed to the outputs.
```
jordansissel: I like this. If this should cause data loss (dead letter replay isn't available yet, right?), can you add a note indicating the data loss here?

jordansissel:
Regarding [@colinsurprenant's comment](https://github.com/elastic/logstash/pull/4051#issuecomment-149982234), I agree on the concept of this but also sit with Colin here that maybe the flag shouldn't be even available. Ideallyy we'd always have this dead lettering available, so an instant shutdown was always possible due to dead lettering functionality.

Thoughts?

lib/logstash/dead_letter_post_office.rb
jsvd: This dead letter mechanism v0.1 is quite dumb: save "lost" events to a file. I set up this destination to include an approximate start up time of Logstash so there's 1 dead letter file per start up.

Having a constant file name here would make it hard to understand which events were lost during each execution, though I might be overthinking this.

jsvd:
+1 on all of @colinsurprenant's comments. The flag comes from my uncertainty of the final version of this feature. With the upcoming persistent work, I'm not sure how this DeadLetterPostOffice stuff will survive, even though the stall detection should remain the same.
Also, right now we don't have a mechanism to automatically feed the dead letters back into Logstash at startup, it's just a simple "where's how you can discard messages in a way that it's possible to ingest later".
This is why I'd consider making this an experimental feature, disabled by default in one or 2 minor versions, and them eventually remove the setting and make it on by default

lib/logstash/dead_letter_post_office.rb
jsvd: agreed. on rpm/deb based installations I'd suggest placing this somewhere close to the logstash logs.
Otherwise, by default, I've though about something inside of the logstash directory, maybe a tmp/ dir?

colinsurprenant:
@jsvd ok, makes sense. Maybe we could wait a bit so see how persistence shape up and see how we can/will hook that up here? I am definitely not against moving forward with an experimental feature, it will provide us a better comprehension of the problem space, OTOH we know this will tie up with persistence so if we wait a bit for this part it will avoid redundant work. But OTOOH :P I think that having a simple json event dump could still be a good choice regardless of persistence, it is simple and provides an very practical way for users to inspect/manipulate these events. 

On a side note: Isn't a Dead Letter (Queue) normally specific to "poison" or unsuccessful events? In the context of a stalled shutdown, the pending events in the queues are not poison events, just normal events that we need to flush to disk to avoid loosing them. Maybe we should change naming here?

jsvd:
According to wikipedia: Dead letter mail or undeliverable mail is mail that cannot be delivered to the addressee or returned to the sender. 

Futher:
This is usually due to lack of compliance with postal regulations, an incomplete address and return address, or the inability to forward the mail when both correspondents move before the letter can be delivered.

I think it fits :)

colinsurprenant:
:incoming_envelope: :postbox: :mailbox:

lib/logstash/plugin.rb
jsvd: I struggled a bit with this decision to either 
a) pass a DeadLetterDestination or DeadLetterPostOffice to each plugin
b) make the DLPO a singleton

Looking at how logger behaves, it's actually more of a singleton, since [plugin.rb](https://github.com/elastic/logstash/blob/master/lib/logstash/plugin.rb#L28) does `@logger = Cabin::Channel.get(LogStash)` and it's always the same channel identifier (LogStash). An approximate equivalent would be to do `@post_office = LogStash::DeadLetterPostOffice` in plugin.rb which doesn't seem that useful, imo.

The big driver question here really is: how fine grain will we want the dead letter destinations to be:
a) 1 per instance 
b) 1 per pipeline
c) 1 per plugin instance

Option d) is to have only one per instance which is: send dead letters to the  `dead_letter { ... }` section of the pipeline config.
This extra config section can then use conditionals, file output, stdout, etc

lib/logstash/dead_letter_post_office.rb
jsvd: updated by surrounding this code with Thread.exclusive

lib/logstash/dead_letter_post_office.rb
```diff
@@ -0,0 +1,53 @@
...
+    @logger ||= Cabin::Channel.get(LogStash)
+  end
+
+  def self.destination=(destination)
+    logger.info("Setting dead letter path", :path => destination.location)
+    @destination = destination
+  end
+
+  def self.post(event)
+    Thread.exclusive do
```
colinsurprenant: any specific reason for using Thread.exclusive here instead of a specific Mutex?

jsvd: not really, this is a single purpose synchronisation point, where all threads access it in the same manner buy doing a post on a singleton, so I didn't see a reason to create a specific instance, but I can be convinced otherwise.

jsvd:
@jordansissel (and anyone else): I would like some advice on the dead letter destination. My though is that dumping queues should (for now) be forcefully done to a file on disk in json format, which is what this PR does.
That said, either this path is configurable or not..so:
a) if it's configurable how to do it? is it another flag that only works if stalled detection is enabled? what is the default value?
b) if it's not, what should be the path? a tmpfile? a file inside the logstash installation directory? is it always the same file name or varies (e.g. with time)

My thought idea is to change this PR's default to write the file as something like: `logstash/logs/dead_letter.#{time_when_logstash_started}.dump`
This way multiple restarts will write multiple files, otherwise multiple restarts would either overwrite the existing dump or append to it, making it harder to understand which stalled events belonged to each run.

andrewvc:
I'd like to say I agree with @colinsurprenant that we shouldn't send in-flight stuff to the DLQ just because of a shutdown. It should only be for poisonous or permanently failing events.

Once persistence is implemented LogStash should be capable of a crash only shutdown, which tackles that problem in a much cleaner way than having to respool selective events.

WRT the DLQ destination, the one design concern I'd like to see addressed is capping the DLQ size. I'd prefer, given that, to just append to a single file vs. write multiple files for each run. If you regularly restart logstash (say via a cron job), it could be easy to blow past any per-file limit.

I'm not sure how we'd accomplish the file locking here (probably a .dlq-lock file containing the PID?).

jsvd:
@andrewvc agreed with all points in the long term when eyeballing the next major version of logstash with persistence.

However, until then, we need to dump the SizedQueues to some place and force the shutdown. In my opinion a lot of this code will probably be deleted once we're able to safely shutdown logstash due to persistence.
Another thing to note here is that what this PR, for now, will only dump the input_to_filter SizeQueue, which will have at most 20 events, so file size should not be much of a concern here. 

I may have done the wrong thing here and associate this with dead letters, I'm considering finding another name for this.

andrewvc:
@jsvd why will only the input_to_filter queue be dropped? Won't we lose in-flight events infilter_to_output (to say nothing of the internal buffer in elasticsearch.rb).

IMHO if we really want this feature we should do the full thing without caveats. Dropping stuff in the Elasticsearch queue is not acceptable. I think our biggest roadblock right now is the java core rewrite, and if persistence is that important we can tackle it in ruby earlier than later.

I think it'd be faster to just implement a persistent queue along the [guidelines](https://github.com/elastic/logstash/issues/3693) I proposed before in ruby (instead of waiting for a java core rewrite), and let it be a configurable setting for now (for those who don't want to eat the perf cost). In fact, I already implemented a rough PoC of that in ~1 day a few months ago with a not bad performance cost (and that was using Elasticsearch as a persistence backend).

colinsurprenant:
@andrewvc we already have a non core-event java-rewrite based persistent queue implementation and we decided to move on with the core event java rewrite and if we are to move into that direction as we seems to agree we are, as you suggest, we should probably focus on this instead on side-tracking on yet another implementation strategy, no? this sound to me going in full circle...

jsvd:
OK so I'm going to make this PR simpler by removing the persisting logic as it is flawed by two big reasons:
a) to ensure a "data loss free" shutdown it persists the SizedQueues (of size 20), but won't persist other buffers like the elasticsearch output (default size 500)
b) Even if all buffers are dumped to disk, there's no reliable way for a user to write a config file that will feed these dumped files into the different levels of the existing config.

With these ideas in mind, the PR will focus on stall detection during shutdown and ensure the messaging is clear that enabling the abort on stall will lose data but ensure logstash can exit.
This will be disabled by default and should stop being a problem once persistence kicks in.

jsvd:
Current output if --force-shutdown is enabled and the pipeline is refusing to shutdown:

```
% bin/logstash -e 'input { generator { } } filter { ruby { code => "sleep 10000" } } output { stdout { codec => dots } }' -w 1 --force-shutdown
Default settings used: Filter workers: 1
Logstash startup completed
data loss. {:level=>:warn}
 {:level=>:warn, "INFLIGHT_EVENT_COUNT"=>{"input_to_filter"=>20, "total"=>20}, "STALLING_THREADS"=>{["LogStash::Filters::Ruby", {"code"=>"sleep 10000"}]=>[{"thread_id"=>15, "name"=>"|filterworker.0", "current_call"=>"(ruby filter code):1:in `sleep'"}]}}
The shutdown process appears to be stalled due to busy or blocked plugins. Check the logs for more information. {:level=>:error}
 {:level=>:warn, "INFLIGHT_EVENT_COUNT"=>{"input_to_filter"=>20, "total"=>20}, "STALLING_THREADS"=>{["LogStash::Filters::Ruby", {"code"=>"sleep 10000"}]=>[{"thread_id"=>15, "name"=>"|filterworker.0", "current_call"=>"(ruby filter code):1:in `sleep'"}]}}
 {:level=>:warn, "INFLIGHT_EVENT_COUNT"=>{"input_to_filter"=>20, "total"=>20}, "STALLING_THREADS"=>{["LogStash::Filters::Ruby", {"code"=>"sleep 10000"}]=>[{"thread_id"=>15, "name"=>"|filterworker.0", "current_call"=>"(ruby filter code):1:in `sleep'"}]}}
Forcefully quitting logstash.. {:level=>:fatal}
%
```

lib/logstash/agent.rb
andrewvc: +1

I would personally prefer `--unsafe-shutdown`. It makes it clear that it's a risk.

lib/logstash/pipeline.rb
```diff
@@ -175,7 +179,8 @@ def start_input(plugin)
   end
 
   def inputworker(plugin)
-    LogStash::Util::set_thread_name("<#{plugin.class.config_name}")
+    LogStash::Util.set_thread_name("<#{plugin.class.config_name}")
+    LogStash::Util.set_thread_plugin(plugin)
```
andrewvc: Love having this around for troubleshooting!

lib/logstash/agent.rb
ph: +1  for  `--allow-unsafe-shutdown`

lib/logstash/filters/base.rb
```diff
@@ -143,6 +143,7 @@ def filter(event)
   # @return [Array<LogStash::Event] filtered events and any new events generated by the filter
   public
   def multi_filter(events)
+    LogStash::Util.set_thread_plugin(self)
```
ph: +1

lib/logstash/agent.rb
```diff
@@ -175,10 +182,7 @@ def execute
   end # def execute
 
   def shutdown(pipeline)
-    pipeline.shutdown do
-      InflightEventsReporter.logger = @logger
-      InflightEventsReporter.start(pipeline.input_to_filter, pipeline.filter_to_output, pipeline.outputs)
-    end
+    pipeline.shutdown
   end
```
ph: Since we are removing most the content of this method, is there any reason not to call pipeline.shutdown directly and remove the inderection?

lib/logstash/pipeline.rb
```diff
@@ -150,8 +151,11 @@ def start_inputs
   def start_filters
     @filters.each(&:register)
     to_start = @settings["filter-workers"]
-    @filter_threads = to_start.times.collect do
-      Thread.new { filterworker }
+    @filter_threads = to_start.times.collect do |i|
+      Thread.new do
+        LogStash::Util.set_thread_name("|filterworker.#{i}")
```
ph: +1

lib/logstash/util.rb
```diff
@@ -24,6 +24,37 @@ def self.set_thread_name(name)
...
+
+  def self.thread_info(thread)
+    backtrace = thread.backtrace.map do |line|
+      line.gsub(LogStash::Environment::LOGSTASH_HOME, "[...]")
+    end
+    state = case backtrace.first
+            when /in `push'/ then "blocked_on_push"
+            when /(?:pipeline|base).*pop/ then "waiting_for_events"
+            else "running"
+            end
```
ph: Should we check if the thread is actually block/sleeping before inspecting the backtrace?
This code look to bypass the meaning of Thread#status, but maybe the naming is wrong, a `block_on` key might be more appropriate?

lib/logstash/shutdown_controller.rb
```diff
@@ -0,0 +1,98 @@
+# encoding: utf-8
+require 'pp'
```
ph: I dont see where its needed?

lib/logstash/shutdown_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+
+module LogStash
+  class ShutdownController
+
+    CHECK_EVERY = 1 # second
+    REPORT_EVERY = 5 # checks
+    ABORT_AFTER = 3 # stalled reports
+    REPORTS = []
```
ph: I think this should be a instance variable?

lib/logstash/shutdown_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+          if cycle_number == (REPORT_EVERY - 1) # it's report time!
+            logger.warn(REPORTS.last)
+
+            if stalled?
+              logger.error("The shutdown process appears to be stalled due to busy or blocked plugins. Check the logs for more information.") if stalled_count == 0
+              stalled_count += 1
+
+              if self.class.force_shutdown? && abort_threshold == stalled_count
+                logger.fatal("Forcefully quitting logstash..")
+                @pipeline.force_exit()
```
ph: `force_shutdown?` do you think it would be a better name?

ph: I know the implementation details is `exit(-1)`, but the concept is forcing an unsafe shutdown?

lib/logstash/shutdown_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+    def logger
+      @logger ||= Cabin::Channel.get(LogStash)
+    end
+
+    def initialize(pipeline)
+      @pipeline = pipeline
+    end
+
+    def start(cycle_period=CHECK_EVERY, report_every=REPORT_EVERY, abort_threshold=ABORT_AFTER)
+      @thread ||= Thread.new do
```
ph: I am looking at the calling code and I don't see where `start` could be called on the same instance?
I don't think we need to lazy instantiate the @thread variable

lib/logstash/shutdown_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+              stalled_count = 0
+            end
+          end
+          cycle_number = (cycle_number + 1) % report_every
+        end
+      end
+    end
+
+    def stop!
+      @thread.terminate if @thread.is_a?(Thread)
```
ph: Did you mean to check agains't nil here?

lib/logstash/pipeline.rb
```diff
@@ -309,4 +322,54 @@ def flush_filters_to_output!(options = {})
...
+  def stalling_threads
+    plugin_threads.select {|t| t["state"] == "running" }.each {|t| t.delete("state") }
+  end
+
+  def plugin_threads
+    input_threads = @input_threads.select {|t| t.alive? }.map {|t| thread_info(t) }
+    filter_threads = @filter_threads.select {|t| t.alive? }.map {|t| thread_info(t) }
+    output_threads = @output_threads.select {|t| t.alive? }.map {|t| thread_info(t) }
+    output_worker_threads = @outputs.flat_map {|output| output.worker_threads }.map {|t| thread_info(t) }
+    input_threads + filter_threads + output_threads + output_worker_threads
```
ph: What do you think to refactor `{|t| t.alive? }.map {|t| thread_info(t) }` into a proc?

lib/logstash/pipeline.rb
```diff
@@ -309,4 +322,54 @@ def flush_filters_to_output!(options = {})
...
+    info = LogStash::Util.thread_info(thread)
+    info["current_call"] = info["backtrace"].first
+    info.delete("backtrace")
+    info["plugin"] = format_plugin_info(info["plugin"])
+    info
+  end
+
+  def format_plugin_info(plugin)
+    return unless plugin.kind_of?(LogStash::Plugin)
+    [plugin.class.to_s, plugin.original_params]
```
ph: I think this information string could be mode the Plugin's base class?

lib/logstash/shutdown_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+    def stalled?
+      return false unless REPORTS.size == REPORT_EVERY
+      # is stalled if inflight count is either constant or increasing
+      stalled_event_count = REPORTS.each_cons(2).all? do |prev_report, next_report|
+        prev_report["INFLIGHT_EVENT_COUNT"]["total"] <= next_report["INFLIGHT_EVENT_COUNT"]["total"]
+      end
+      if stalled_event_count
+        REPORTS.each_cons(2).all? do |prev_report, next_report|
```
ph: TIL I should use `each_cons` more often +1

locales/en.yml
ph: +1 on the data loss

lib/logstash/shutdown_controller.rb
jsvd: +1 I already had this in a local commit, will push today

lib/logstash/shutdown_controller.rb
jsvd: because it's not. :+1:

lib/logstash/shutdown_controller.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+    end
+
+    def stop!
+      @thread.terminate if @thread.is_a?(Thread)
+      @thread = nil
+    end
+
+    def stalled?
+      return false unless REPORTS.size == REPORT_EVERY
```
ph: use an instance variable here instead of the constant

lib/logstash/util.rb
ph: We need test harness for the inspected string, since this part of the code will change in the upcoming weeks.

lib/logstash/pipeline.rb
ph: Do we expect something else than a plugin here?

ph:
After discussing with @jsvd we agreed to move the ShutdownController initialize outside of the pipeline and the pipeline will be pass to the controller  this will make the code easier to follow and more extensible.

lib/logstash/shutdown_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+        end
+        cycle_number = (cycle_number + 1) % @report_every
+      end
+    end
+
+    # A pipeline shutdown is stalled if
+    # * at least REPORT_EVERY reports have been created
+    # * the inflight event count is in monotonically increasing
+    # * there are worker threads running which aren't blocked on SizedQueue pop/push
+    # * the stalled thread list is constant in the previous REPORT_EVERY reports
```
ph: +1 for the comment

lib/logstash/shutdown_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    def initialize(inflight_count, stalling_threads)
+      @inflight_count = inflight_count
+      @stalling_threads = format_threads_by_plugin(stalling_threads)
+    end
+
+    def to_hash
+      {
+        "INFLIGHT_EVENT_COUNT" => @inflight_count,
+        "STALLING_THREADS" => @stalling_threads
```
ph: You could use the accessor here for theses values.
@stalling_threads => stalling_threads

ph:
code looks code, will do some manual testing.

ph:
LGTM.

Maybe add the `--allow-unsafe-shutdown` to the documentation

Manual tests run successfully.

Stuck pipeline.
```
~/e/logstash git:pr/4051 ❯❯❯ bin/logstash -e 'input { generator { } } filter { ruby { code => "sleep 10000" } } output { stdout { codec => dots } }' -w 1 --allow-unsafe-shutdown                                                                                                                                                                                       ✱
Default settings used: Filter workers: 1
Logstash startup completed
data loss. {:level=>:warn}
 {:level=>:warn, "INFLIGHT_EVENT_COUNT"=>{"input_to_filter"=>20, "total"=>20}, "STALLING_THREADS"=>{["LogStash::Filters::Ruby", {"code"=>"sleep 10000"}]=>[{"thread_id"=>15, "name"=>"|filterworker.0", "current_call"=>"(ruby filter code):1:in `sleep'"}]}}
The shutdown process appears to be stalled due to busy or blocked plugins. Check the logs for more information. {:level=>:error}
 {:level=>:warn, "INFLIGHT_EVENT_COUNT"=>{"input_to_filter"=>20, "total"=>20}, "STALLING_THREADS"=>{["LogStash::Filters::Ruby", {"code"=>"sleep 10000"}]=>[{"thread_id"=>15, "name"=>"|filterworker.0", "current_call"=>"(ruby filter code):1:in `sleep'"}]}}
 {:level=>:warn, "INFLIGHT_EVENT_COUNT"=>{"input_to_filter"=>20, "total"=>20}, "STALLING_THREADS"=>{["LogStash::Filters::Ruby", {"code"=>"sleep 10000"}]=>[{"thread_id"=>15, "name"=>"|filterworker.0", "current_call"=>"(ruby filter code):1:in `sleep'"}]}}
Forcefully quitting logstash.. {:level=>:fatal}
````
Normal healthy pipeline.
````
~/e/logstash git:pr/4051 ❯❯❯ bin/logstash -e 'input { generator { } } output { stdout { codec => dots } }' -w 1 --allow-unsafe-shutdown                                                                                                                                                                                                                               ⏎ ✱
Default settings used: Filter workers: 1
............................................................................................Logstash startup completed
.................................................................................................................................................................SIGINT received. Shutting down the pipeline. {:level=>:warn}

**Comment by ph** [comment]
@jsvd I have removed the LGTM, the tests are failling on this branch.
Just fix them and its LGTM.

**Comment by ph** [comment]
failling tests:
```

  1) LogStash::ShutdownController when unsafe_shutdown is true with a non-stalled pipeline should request more than NUM_REPORTS "inflight_count"
     Failure/Error: LogStash::ShutdownController::REPORTS.clear
     NameError:
       uninitialized constant LogStash::ShutdownController::REPORTS
     # ./spec/core/shutdown_controller_spec.rb:12:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.7/lib/rspec/wait.rb:46:in `(root)'
     # ./lib/bootstrap/rspec.rb:11:in `(root)'

  2) LogStash::ShutdownController when unsafe_shutdown is true with a non-stalled pipeline shouldn't force exit after NUM_REPORTS cycles
     Failure/Error: LogStash::ShutdownController::REPORTS.clear
     NameError:
       uninitialized constant LogStash::ShutdownController::REPORTS
     # ./spec/core/shutdown_controller_spec.rb:12:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.7/lib/rspec/wait.rb:46:in `(root)'
     # ./lib/bootstrap/rspec.rb:11:in `(root)'

  3) LogStash::ShutdownController when unsafe_shutdown is true with a stalled pipeline should force exit after NUM_REPORTS cycles
     Failure/Error: LogStash::ShutdownController::REPORTS.clear
     NameError:
       uninitialized constant LogStash::ShutdownController::REPORTS
     # ./spec/core/shutdown_controller_spec.rb:12:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.7/lib/rspec/wait.rb:46:in `(root)'
     # ./lib/bootstrap/rspec.rb:11:in `(root)'
````

jsvd:
merged to:
2.1: 1412885
2.x: 79b90b1
master: 5db1d28

PR #4052, opened 2015-10-16
Use packages.elastic.co instead

untergeek:
packages.elastic.co uses the same bucket as packages.elasticsearch.org.  We can use this now.


suyograo:
LGTM

PR #4057, opened 2015-10-19
Logstash-input-beats is now part of the default plugins

ph:


ph:
This should be merge in 2.0 and 1.5

PR #4061, opened 2015-10-19
Bump versions to use latest jrjackson and beats

suyograo:


ph:
LGTM

suyograo:
@ph can you take another look after the recent commit? I added beats and bumped version to beta3

ph:
LGTM

PR #4085, opened 2015-10-23
Change how we set the max heap size

### LINKED ISSUE #3134: support consistent env variables between the Linux and Windows startup scripts
the supported environment variables are not consistent between Linux and Windows. 

for example, in Linux we use `LS_HEAP_SIZE` while on it's `LS_MIN_MEM` and `LS_MAX_MEM`. 

I don't know what the history is for this but I think we should make them consistent as much as possible.

ph:
This PR change the name of the variable to set to change the maximum
heap size on window to `LS_HEAP_SIZE` this replace the LS_MIN_MEM and
LS_MAX_MEM option.

Also We change the default size of the heap from 500m to 1g

Fixes #3134

### DIFF
--- a/bin/logstash.lib.sh
+++ b/bin/logstash.lib.sh
@@ -3,7 +3,7 @@ LOGSTASH_HOME=$(cd `dirname $0`/..; pwd)
 export LOGSTASH_HOME
 
 # Defaults you can override with environment variables
-LS_HEAP_SIZE="${LS_HEAP_SIZE:=500m}"
+LS_HEAP_SIZE="${LS_HEAP_SIZE:=1g}"
 
 setup_java() {
   if [ -z "$JAVACMD" ] ; then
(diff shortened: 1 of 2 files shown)


untergeek:
I don't have a windows box to test it on, but LGTM™ code-wise.

jordansissel:
Didn't test, but code LGTM

jsvd:
why change bump the default to 1g on this PR? I'd suggest making this 500mb as linux is https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L6

ph:
@jsvd I was also intrigued by why we had a different settings on windows, but I think to reduce the memory on windows from 1gb to 500mb would be a bit confusing for people. Since we want to give 1gb in 2.1 I though I would normalize both to the upper bound which is 1gb

ph:
@jsvd it wasn't a global change see https://github.com/elastic/logstash/pull/4085/files#diff-623067f911d57d55649dacff126510fdL26 windows used 1gb as their default for some time

jsvd:
LGTM

PR #4089, opened 2015-10-26
Fix doc generation for obsolete settings

suyograo:
We introduced `obsolete` in #3977 which means we need to remove these obsolete configs from docs. Previous fix was broken, this one properly removes configs which are obsolete. See https://www.elastic.co/guide/en/logstash/2.0/plugins-outputs-elasticsearch.html

New docs pushed to elastic.co


jordansissel:
Haven't tested, but the change looks correct. :)

LGTM

Issue #4074, opened 2015-10-21
[meta] codec stream identity


colinsurprenant:
this issue follows up the discussion in logstash-plugins/logstash-codec-multiline#10 about supporting stream identity in codecs to solve the problem of using the multiline codec in conjunction with the file or lumberjack inputs for example.

Since this issue involves many plugins and is an important issue to solve we should keep track of it here.

My [last suggestion](https://github.com/logstash-plugins/logstash-codec-multiline/issues/10#issuecomment-149993542) is to do the following:
- replace the codec instantiation in the config to instead instantiate a new "codec multiplexer" class, which will ultimately expose a **backward compatible** decode method like `decode(data, stream_identity = nil)`
- this new multiplexer `decode` method would take care of doing the bookkeeping of the stream identities vs actual codec instances.

This would allow to not touch/update all plugins where dealing with stream identities in codecs is not important and simply update the ones like file & lumberjack inputs to pass the values which makes sense as `stream_identity` to the `decode` method. for example for file input, the current code:

``` ruby
    @tail.subscribe do |path, line|
      @codec.decode(line) do |event|
```

would become:

``` ruby
    @tail.subscribe do |path, line|
      @codec.decode(line, path) do |event|
```

For the codecs nothing would change and we can all assume a single codec instance will always deal with a single consistent stream.

guyboertje:
@colin, @ph - will each input know the full set of stream_identity values when starting or is there dynamic stream_identity detection and new codec spawn while receiving data?

guyboertje:
As expressed in [this comment](https://github.com/logstash-plugins/logstash-input-file/issues/44#issuecomment-124762024), do I presume that multiple file inputs having different absolute paths will not use any form of codec multiplexing?

jordansissel:
The following question is meant earnestly, and I am on my phone so I will
be brief -

Do we need an abstraction for this? How many plugins would benefit? Only
file and lumberjack? Maybe TCP? Each of these (3) plugins have different
thread models so it may be irritating to make one interface for all - my
counter question is this: can we just implement what each of these plugins
need in each plugin that needs it? For data points, what plugins need this
multiplexing?

colinsurprenant:
[...]

PR #4104, opened 2015-10-28
WIP: Add stream identity support to LS

guyboertje:
This after discussion on #4074.

Notes for reviewers:

You will see commented out lines that use Concurrent::Hash.  They are so because I could not seem to require the correct path to make the Hash class appear in the Concurrent module. The intention is still to use Concurrent::Hash.

There are some inner classes and modules, done so to keep it self contained.

### DIFF
--- a/lib/logstash/codecs/identity_map_codec.rb
+++ b/lib/logstash/codecs/identity_map_codec.rb
@@ -1,0 +1,176 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "concurrent"
+
+# This class is a Codec duck type
+# Using Composition, it maps from a stream identity to
+# a cloned codec instance via the same API as a Codec
+
+module LogStash module Codecs class IdentityMapCodec
+  # subclass of Exception, LS has more than limit (100) active streams
+  class IdentityMapUpperLimitException < Exception; end
+
+  module EightyPercentWarning
+    extend self
+    def visit(imc)
+      current_size, limit = imc.current_size_and_limit
+      return if current_size < (limit * 0.8)
+      imc.logger.warn("IdentityMapCodec has reached 80% capacity",
+        :current_size => current_size, :upper_limit => limit)
+    end
+  end
+
+  module UpperLimitReached
+    extend self
+    def visit(imc)
+      current_size, limit = imc.current_size_and_limit
+      return if current_size < limit
+      current_size, limit = imc.map_cleanup
+      return if current_size < limit
+      imc.logger.error("IdentityMapCodec has reached 100% capacity",
+          :current_size => current_size, :upper_limit => limit)
+      raise IdentityMapUpperLimitException.new
+    end
+  end
+
+  class MapCleaner
+    def initialize(imc, interval)
+      @running = true
+      @imc, @interval = imc, interval
+    end
+
+    def run
+      @thread = Thread.new(@imc) do |imc|
+        loop do
+          sleep @interval
+          break if !@running
+          imc.map_cleanup
+        end
+      end
+      self
+    end
+
+    def stop
+      @running = false
+      @thread.wakeup
+    end
+  end
+
+  MAX_IDENTITIES = 100
+  EVICT_TIMEOUT = 60 * 60 * 4 # 4 hours
+  CLEANER_INTERVAL = 60 * 5 # 5 minutes
+
+  attr_reader :identity_map, :usage_map
+  attr_accessor :base_codec, :logger, :cleaner
+
+  def initialize(codec, logger)
+    @base_codec = codec
+    @base_values = [codec]
+    @identity_map = Hash.new &method(:codec_builder)
+    # @identity_map = Concurrent::Hash.new &method(:codec_builder)
+    @usage_map = Hash.new
+    # @usage_map = Concurrent::Hash.new
+    @logger = logger
+    @max_identities = MAX_IDENTITIES
+    @evict_timeout = EVICT_TIMEOUT
+    @cleaner = MapCleaner.new(self, CLEANER_INTERVAL).run
+  end
+
+  def max_identities(max)
+    @max_identities = max.to_i.abs
+    self
+  end
+
+  def evict_timeout(timeout)
+    @evict_timeout = timeout.to_i.abs
+    self
+  end
+
+  def cleaner_interval(interval)
+    @cleaner.stop
+    @cleaner = MapCleaner.new(self, interval.to_i.abs).run
+    self
+  end
+
+  def stream_codec(identity)
+    return base_codec if identity.nil?
+    track_identity_usage(identity)
+    identity_map[identity]
+  end
+
+  def decode(data, identity = nil, &block)
+    stream_codec(identity).decode(data, &block)
+  end
+
+  alias_method :<<, :decode
+
+  def encode(event, identity = nil)
+    stream_codec(identity).encode(event)
+  end
+
+  # this method should not be called from
+  # the input or the pipeline
+  def flush(&block)
+    map_values.each do |codec|
+      #let ruby do its default args thing
+      block.nil? ? codec.flush : codec.flush(&block)
+    end
+  end
+
+  def close()
+    cleaner.stop
+    map_values.each(&:close)
+  end
+
+  def map_values
+    no_streams? ? @base_values : identity_map.values
+  end
+
+  def max_limit
+    @max_identities
+  end
+
+  def size
+    identity_map.size
+  end
+
+  # support cleaning of stale codecs
+  def map_cleanup
+    cut_off = Time.now.to_i
+    candidates, rest = usage_map.partition {|identity, timeout| timeout <= cut_off }
+    candidates.each do |identity, timeout|
+      identity_map.delete(identity).flush
+      usage_map.delete(identity)
+    end
+    current_size_and_limit
+  end
+
+  def current_size_and_limit
+    [size, max_limit]
+  end
+
+  private
+
+  def track_identity_usage(identity)
+    check_map_limits
+    usage_map.store(identity, eviction_timestamp)
+  end
+
+  def eviction_timestamp
+    Time.now.to_i + @evict_timeout
+  end
+
+  def check_map_limits
+    UpperLimitReached.visit(self)
+    EightyPercentWarning.visit(self)
+  end
+
+  def codec_builder(hash, k)
+    codec = hash.empty? ? @base_codec : @base_codec.clone
+    hash.store(k, codec)
+  end
+
+  def no_streams?
+    identity_map.empty?
+  end
+end end end
--- a/spec/codecs/identity_map_codec_spec.rb
+++ b/spec/codecs/identity_map_codec_spec.rb
@@ -1,0 +1,207 @@
+# encoding: utf-8
+require "spec_helper"
+
+require "logstash/codecs/identity_map_codec"
+
+class IdentityMapCodecTracer
+  def initialize() @tracer = []; end
+  def clone() self.class.new; end
+  def decode(data) @tracer.push [:decode, data]; end
+  def encode(event) @tracer.push [:encode, event]; end
+  def flush(&block) @tracer.push [:flush, true]; end
+  def close() @tracer.push [:close, true]; end
+
+  def trace_for(symbol)
+    params = @tracer.assoc(symbol)
+    params.nil? ? false : params.last
+  end
+end
+
+class LogTracer
+  def initialize() @tracer = []; end
+  def warn(*args) @tracer.push [:warn, args]; end
+  def error(*args) @tracer.push [:error, args]; end
+
+  def trace_for(symbol)
+    params = @tracer.assoc(symbol)
+    params.nil? ? false : params.last
+  end
+end
+
+describe LogStash::Codecs::IdentityMapCodec do
+  let(:codec)   { IdentityMapCodecTracer.new }
+  let(:logger)  { LogTracer.new }
+  let(:demuxer) { described_class.new(codec, logger) }
+  let(:stream1) { "stream-a" }
+  let(:codec1)  { demuxer.stream_codec(stream1) }
+  let(:arg1)    { "data-a" }
+
+  after do
+    codec.close
+  end
+
+  describe "operating without stream identity" do
+    let(:stream1) { nil }
+
+    it "transparently refers to the original codec" do
+      expect(codec).to eql(codec1)
+    end
+  end
+
+  describe "operating with stream identity" do
+
+    before { demuxer.decode(arg1, stream1) }
+
+    it "the first identity refers to the original codec" do
+      expect(codec).to eql(codec1)
+    end
+  end
+
+  describe "#decode" do
+    context "when no identity is used" do
+      let(:stream1) { nil }
+
+      it "calls the method on the original codec" do
+        demuxer.decode(arg1, stream1)
+
+        expect(codec.trace_for(:decode)).to eq(arg1)
+      end
+    end
+
+    context "when multiple identities are used" do
+      let(:stream2) { "stream-b" }
+      let(:codec2) { demuxer.stream_codec(stream2) }
+      let(:arg2)   { "data-b" }
+
+      it "calls the method on the appropriate codec" do
+        demuxer.decode(arg1, stream1)
+        demuxer.decode(arg2, stream2)
+
+        expect(codec1.trace_for(:decode)).to eq(arg1)
+        expect(codec2.trace_for(:decode)).to eq(arg2)
+      end
+    end
+  end
+
+  describe "#encode" do
+    context "when no identity is used" do
+      let(:stream1) { nil }
+      let(:arg1) { LogStash::Event.new({"type" => "file"}) }
+
+      it "calls the method on the original codec" do
+        demuxer.encode(arg1, stream1)
+
+        expect(codec.trace_for(:encode)).to eq(arg1)
+      end
+    end
+
+    context "when multiple identities are used" do
+      let(:stream2) { "stream-b" }
+      let(:codec2) { demuxer.stream_codec(stream2) }
+      let(:arg2)   { LogStash::Event.new({"type" => "file"}) }
+
+      it "calls the method on the appropriate codec" do
+        demuxer.encode(arg1, stream1)
+        demuxer.encode(arg2, stream2)
+
+        expect(codec1.trace_for(:encode)).to eq(arg1)
+        expect(codec2.trace_for(:encode)).to eq(arg2)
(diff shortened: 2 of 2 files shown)


jordansissel:
I'm still cautious of this solution. It's nearly 400 lines of code (including tests) to solve problems that, what, only 3 inputs have, and 2 of which don't have it well-solved?

I agree with the problem (tcp, file, lumberjack inputs w/ codecs and multiple streams), but I feel like the solution for file and lumberjack inputs will be simpler if focused in each plugin. Thoughts?

suyograo:
@guyboertje does this change affect users (config)? If so, can you paste some config examples here of how the file input would be used with multiline?

guyboertje:
@jordansissel - as it is self contained, it could easily be moved/replaced.

@suyograo - file input: no change to config; other inputs: possible change to specify the stream_Identity field.  For lumberjack and tcp, we discussed supporting explicit and implicit modes, explicit is when the other end manages the stream_identity and implicit when this end synthesises the identity from known values inherent in the connection. We did not foresee a config change but a plugin (file, lumberjack, tcp) change would be needed.

TO ANY READERS:  The elephant in the room here has more to do with where this code should live.
Q1: Does this solution, in reality, only apply to the multiline codec? If so, should this solution be in the multiline codec or stay in core? For if we move this to the multiline codec then the solution can be made available to 1.5 users if their upgrade their plugins.
Q2: If we move the solution to the multiline codec, how problematic is it to have the file input version tied to the multiline codec version?

jordansissel:
@guyboertje I like the proposed change. I felt some initial dischord with the change as proposed, but I'm not quite able to articulate it beyond "it feels ..." and certainly nothing as firm or probable, haha. So! I am happy with the proposed implementation idea and will get you some feedback on the code and your other questions :)

jordansissel:
> Does this solution, in reality, only apply to the multiline codec

Any codec which has a non 1:1 relationship between payload and event will have this problem. Examples: multiline, line, json_lines. Other serializations and compression codecs (when and if they exist) will have the same problems.

We have at least one input plugin which have multiple streams, but handle this already by implementing stream identities themselves. For example, the [tcp input clones `@codec` for each connection](https://github.com/logstash-plugins/logstash-input-tcp/blob/master/lib/logstash/inputs/tcp.rb#L129). (Experience with the way tcp input solves this is where my [initial feelings](https://github.com/elastic/logstash/pull/4104#issuecomment-151935636) on this issue came from)

jordansissel:
> how problematic is it to have the file/lumberjack/tcp input version tied to the multiline codec version

I think we can enforce this with plugin dependency versions, possibly. So may not be a problem.

colinsurprenant:
@guyboertje @jordansissel or we could release this as a plugin mixin and just depend on it as we roll out  support in the codecs, no? this solves the problem of avoiding a core change and add a "clean" dependency without depending on a specific codec line multiline.

jordansissel:
@colinsurprenant we could yes. I should articulate things more, sorry about that.

This implementation uses Hash and doesn't seem to have any memory barriers to prevent concurrent access (things like invoking `map.values` _might_ cause an exception if we insert a new identity while `.values` is running).

This implementation also seems to have a cap on the number of simultaneous active streams allowed - do we need a cap? If we do, 100 feels really small - at my past job I had ~8000 logstash-forwarders sending about 5 to 10 files _each_ which means 40,000 to 80,000 active streams).

I am also unsure of the purpose of the time-based eviction. TCP and file inputs both know when a stream is dead (tcp connection close, file hasn't been used in a while), but lumberjack doesn't, so it feels like the time-based eviction is only helping the lumberjack/beats input. We could improve the beats/lumberjack protocol to support stream signalling (new stream, close stream) and then we can evict based on signals other than time.

Basically, this approach feels right, but given the scope (only a few plugins, and each plugin doing work a different way), it seems like this solves some problems that aren't necessary? Hopefully I'm making sense here. Let me know :)

guyboertje:
@jordansissel - in the "notes to reviewers" I state that I intend to use Concurrent::Hash

lib/logstash/codecs/identity_map_codec.rb
```diff
@@ -0,0 +1,176 @@
...
+  end
+
+  alias_method :<<, :decode
+
+  def encode(event, identity = nil)
+    stream_codec(identity).encode(event)
+  end
+
+  # this method should not be called from
+  # the input or the pipeline
```
colinsurprenant: not sure I understand this comment: «should not be called» ? 
currently AFAICT the only place codec flush is called is from the multiline input and this is where codec flush should be called no?

lib/logstash/codecs/identity_map_codec.rb
```diff
@@ -0,0 +1,176 @@
...
+require "logstash/namespace"
+require "concurrent"
+
+# This class is a Codec duck type
+# Using Composition, it maps from a stream identity to
+# a cloned codec instance via the same API as a Codec
+
+module LogStash module Codecs class IdentityMapCodec
+  # subclass of Exception, LS has more than limit (100) active streams
+  class IdentityMapUpperLimitException < Exception; end
```
colinsurprenant: unless there is a specific reason for it, error/exception classes should preferably descend from `StandardError` because when descending from `Exception` a bare rescue clause will not catch them.

lib/logstash/codecs/identity_map_codec.rb
```diff
@@ -0,0 +1,176 @@
...
+  EVICT_TIMEOUT = 60 * 60 * 4 # 4 hours
+  CLEANER_INTERVAL = 60 * 5 # 5 minutes
+
+  attr_reader :identity_map, :usage_map
+  attr_accessor :base_codec, :logger, :cleaner
+
+  def initialize(codec, logger)
+    @base_codec = codec
+    @base_values = [codec]
+    @identity_map = Hash.new &method(:codec_builder)
```
colinsurprenant: I personally prefer having parenthesis for method call parameters, in any case, it should at least be consistent? below is
```
MapCleaner.new(self, CLEANER_INTERVAL).run
```

lib/logstash/codecs/identity_map_codec.rb
```diff
@@ -0,0 +1,176 @@
...
+    @usage_map = Hash.new
+    # @usage_map = Concurrent::Hash.new
+    @logger = logger
+    @max_identities = MAX_IDENTITIES
+    @evict_timeout = EVICT_TIMEOUT
+    @cleaner = MapCleaner.new(self, CLEANER_INTERVAL).run
+  end
+
+  def max_identities(max)
+    @max_identities = max.to_i.abs
```
colinsurprenant: any specific reason to call `.abs` here? I don't understand the purpose

jordansissel:
@guyboertje haha, I see that now- thanks for pointing out what I was missing :)

colinsurprenant:
@jordansissel I am also on the fence about time-based evictions. I will try to formulate a more constructive comment once I wrap my mind around it.

lib/logstash/codecs/identity_map_codec.rb
```diff
@@ -0,0 +1,176 @@
...
+    self
+  end
+
+  def evict_timeout(timeout)
+    @evict_timeout = timeout.to_i.abs
+    self
+  end
+
+  def cleaner_interval(interval)
```
colinsurprenant: what is the purpose of this `cleaner_interval` method? I only see it used in the specs

jordansissel:
I just realized I'm arguing about less code vs more code, and that's not really helpful. Sorry about this :)

I agree with this proposed design to solve the codecs problem.

jordansissel:
@guyboertje Can you tell me the story behind time-based identity eviction? Why do we need it? Why would time-based eviction be chosen instead of explicit eviction (stream closed, tcp connection closed, etc)?

lib/logstash/codecs/identity_map_codec.rb
guyboertje: In the multiline codec the flush method is called from the `do_next` or `do_previous` methods.  That is, its done in a controlled manner, not arbitrarily from the pipeline.

In fact I think there is a bug in LS - filters are added to the `@shutdown_flushers` array and are flushed at shutdown.  The file input, for example, does not flush at shutdown.

Further, it is not possible ATM, to tell, from a plugin, whether a codec has is buffered and whether a buffered codec is empty or not.

lib/logstash/codecs/identity_map_codec.rb
guyboertje: I did this because I feel that when this codec has reached the upper limit (what ever we agree that should be) and no streams can be evicted to make space, it is a fatal error that indicates something very bad has gone wrong.

lib/logstash/codecs/identity_map_codec.rb
guyboertje: I do this because passing a block is usually done outside the parentheses.  It is a syntactic anomaly in Ruby.
```ruby
method_call(42) {|arg| some_code }
method_call(42) do |arg|
  some_code
end
method_call(42, &method(:sym))
method_call 42, &method(:sym)
# whereas these fail
method_call(42) &method(:sym)
method_call 42, {|arg| some_code}
```
But we never write `array.<<(member)` or `60.*(60).*(4)` or `hash.[]=('key', 42)`

lib/logstash/codecs/identity_map_codec.rb
guyboertje: A negative number would not be good.

lib/logstash/codecs/identity_map_codec.rb
guyboertje: It is useful in the tests but a plugin could customise this if necessary

guyboertje:
Regarding time based eviction. After discussions with @ph, I decided that, during normal running, it would be difficult to get the upstream end to signal that it had finished with a stream - here I am more concerned with a connection based input.  Pipeline shutdown is a different matter.  So we decided to **try** using a time based eviction mechanism.  Also, with time based eviction, allows for minimal changes to the plugins that use this code.

If the mapped codec is stateless then there is harm, from the plugins viewpoint, in evicting the stream + codec because the next usage of that stream would create it again. However if it has state, a buffer, then evicting it is more problematic.  We need a way of telling whether the eviction candidate has an empty buffer before we can really evict it. Currently, buffering has been added to the multiline codec, but this is ad hoc, There is no general API in the codec base to ask it if it is currently holding state.

guyboertje:
@jordansissel @colinsurprenant - ^^ some replies

lib/logstash/codecs/identity_map_codec.rb
guyboertje: `max_identities`, `evict_timeout ` and `cleaner_interval ` have a 'fluent' call returning `self`. Intended to be used during construction.

lib/logstash/codecs/identity_map_codec.rb
colinsurprenant: still, I find it a lot more easy to read to group a *defined* methods parameters into parenthesis. I think that regardless of the "anomaly" created between using `{ }` and `do |x| end` vs `&method`, these are actually different declarations, `&method` is more a "reference" to a closure whereas `do |x| end` is "the" closure and in the latter it would be awkward to require to have them declared inside the parenthesis so IMO this is good "sugar" :P

But, hey, as I said, this is a personal preference!

lib/logstash/codecs/identity_map_codec.rb
colinsurprenant: so, the idea is to accept negative numbers as positives? I dont think this is a good idea for many reasons:
- the max parameter is not documented to it is not clear what are the **exact** expected & valid parameters from an API standpoint.
- by reading the method, we can probably infer that a positive integer it certainly expected here
- having the `abs` call  just adds confusion and it actually made me want to understand the purpose and started searching the `max_identities` call in the sources.
- this is an "undocumented and weird" feature that this method also accepts negatives numbers and eventually, someone somewhere may leverage this but since this is not an explicit feature, it may & will certainly change in the future, if for example, there is positive number validation added upstream in the code. 

I would prefer that either this is clearly documented of just removed and either the we leave it for the code to break if an invalid parameter is passed, or we add explicit parameter validation.

lib/logstash/codecs/identity_map_codec.rb
colinsurprenant: ok. AFAIK *fluent* style has not been used in the logstash code base. I am not against it but we should probably discuss if we want to introduce that style. In any way, I suggest you document it is the class definition with probably an example so that eventual plugin authors that want to use this will know how to use this class?

lib/logstash/codecs/identity_map_codec.rb
colinsurprenant: ok, let's have a side discussion on flushing, there is definitely some inconsistencies, we should resolve that first then we can fix & document per our agreed strategy.

lib/logstash/codecs/identity_map_codec.rb
colinsurprenant: ok, I see. but still, generally speaking I think a better strategy is to have a base error class and have all  class specific errors inherit from this base class so that its possible to have a specific class-wide rescue clause. in this case I guess it's debatable if we should consider this a hard or fatal error or an error that could be expected and rescued (for logging)

```ruby
class MyClassBaseError < StandardError; end
class FooError < MyClassBaseError; end

...
rescue MyClassBaseError => e
  # do something specific with all errors from MyClass
end
...
```

lib/logstash/codecs/identity_map_codec.rb
```diff
@@ -0,0 +1,176 @@
...
+  end
+
+  MAX_IDENTITIES = 100
+  EVICT_TIMEOUT = 60 * 60 * 4 # 4 hours
+  CLEANER_INTERVAL = 60 * 5 # 5 minutes
+
+  attr_reader :identity_map, :usage_map
+  attr_accessor :base_codec, :logger, :cleaner
+
+  def initialize(codec, logger)
```
colinsurprenant: I am not sure I like the idea of passing a `logger` here. I don't think its this class concern to log errors? Personally I find having the logging code here clutters the code and make it less reusable (not that we could really reuse it in other contexts) but still, I don't really like the idea of passing around high-level concerns like logging into more specific classes.

lib/logstash/codecs/identity_map_codec.rb
guyboertje: You are correct. This is a public OSS project. I intended that these calls would be for use inside an input only and not part of the external config. If someone submits a PR that makes a change to these values then we can validate by eye. I will remove the `abs` call.

lib/logstash/codecs/identity_map_codec.rb
guyboertje: As the class is not a subclass of `Codecs::Base` it does not have a logger of its own.  I had not considered "borrowing" the logger from the codec until now.  I suppose if the different logger instances in the pipeline and plugins were given specific contexts then the logger from the codec would be more appropriate than the input.

Would you be happy if I used the codec logger?

guyboertje:
Damn.  Concurrent::Hash is only available on v1.0.0 as yet unreleased. (pre4 is though)

lib/logstash/codecs/identity_map_codec.rb
colinsurprenant: it's more a general concern of passing around logger instances - for me there is code where it "makes sense" to do logging, typically where the code is directly related to the "UI flow" I would say, and other code where errors conditions and/or exceptions should be used. So in plugins it makes sense to me to have logging, but classes used in a plugin should not, it makes the code depend on some "logger" thing which it should not be concerned with, etc. So I would revisit the actual need for logging here (but pragmatically I do understand the benefit of having more/better logging).

So in this case it is a bit ambiguous as this actually wraps the codecs. So I think your suggestion to use the wrapped codec logger would be good.

ph:
@guyboertje you can probably use `ThreadSafe::Cache` which exist in 0.9?

guyboertje:
@ph - `ThreadSafe::Hash` is available.  Concurrent merges thread_safe in 1.0.0 but then its called Concurrent::Hash

lib/logstash/codecs/identity_map_codec.rb
guyboertje: Agreed

colinsurprenant:
@ph @guyboertje actually, excellent suggestion, since [«we recommend you consider ThreadSafe::Cache instead of ThreadSafe::Hash for your concurrency-safe hash needs»](https://github.com/ruby-concurrency/thread_safe/tree/fa5a8affaf8d9d5335d4396db84c384318201fcc#usage) when ordering by insertion time is not important.

ph:
@guyboertje yeah, they are making a lot of breaking changes < 1.0 ;)

lib/logstash/codecs/identity_map_codec.rb
guyboertje: For me it was a design choice in the initializer between horrible extra three defaulted arguments or ruby 2.0 named arguments (has jruby nailed this yet?) or the fluent methods.

Given that if we settled on sane defaults for these three values then the likelihood of anyone needing to provide alternative values goes way down - so I chose the fluent option.

I will add some document comments though when the non-WIP of this PR comes around.

guyboertje:
@ph @colinsurprenant - I can use Cache for one hash and Hash for the other because I would like to use the Enumerable method `partition` on the Hash based one.

suyograo:
> 
Q1: Does this solution, in reality, only apply to the multiline codec? If so, should this solution be in the multiline codec or stay in core? For if we move this to the multiline codec then the solution can be made available to 1.5 users if their upgrade their plugins.

Just had a sync with @guyboertje on this. We feel its better to have the code in ML codec so we can iterate quickly (if bugs happen) and publish new versions. Having this in core would tie it to a core release. Also this really does not have any dependency on core code. 

> 
Q2: If we move the solution to the multiline codec, how problematic is it to have the file/lumberjack/tcp input version tied to the multiline codec version?

Should be fine. It can depend on a particular version of ML codec and they should all be backward compatible.

@guyboertje @colinsurprenant @jordansissel re: cache, i see its use where we need to close streams periodically so we don't grow unbounded. I suggested to @guyboertje that we make the default eviction policy the same as `dead_time` (24h) in LSF. 

Also, another point I'd like to make: this is mostly a stop-gap solution until [filebeat implements](https://github.com/elastic/filebeat/issues/89) multiline capabilities so we can push all this logic to the edge node -- this makes it cleaner.

jordansissel:
> We feel its better to have the code in ML codec

+1 for fast iteration.

+1 what @suyograo  said above :)

guyboertje:
OK then I will move it to ML codec.  Closing.

PR #4108, opened 2015-10-29
Revert back to stud 0.0.21 and es-out 1.0.7

suyograo:
I can't see a tag for v1.1.0 for elasticsearch-output, so going back to 1.5.4 version. Also [stud 0.0.22](https://github.com/jordansissel/ruby-stud/commit/4f2128a2be9616bae902feed37a661370dfde638) has the shutdown change which we don't want to backport to 1.5.


colinsurprenant:
LGTM

PR #4112, opened 2015-10-29
Unused constant `CONFIGSORT`

ph:
This was probably used to have some sort of preference when defining multiple options, but I can't find trace of it in the actual code.


jordansissel:
Indeed. It was used way way way back in the day (Logstash 1.0/1.1)

jordansissel:
LGTM

PR #4114, opened 2015-10-29
Deprecated method `to_event` in input.base

ph:
No current plugins uses this and it was deprecated since at least 1.4.2


suyograo:
LGTM

PR #4115, opened 2015-10-29
RESERVED doesn't do anything

ph:
I understand why this constant was created but its not used anywhere.
I think we lived without that concept for quite a while ;)


suyograo:
@jordansissel safe to remove?

jordansissel:
+1 Safe to remove.

This was used back when setting names could be regular expression patterns (See [this example from Logstash v1.1.13](https://github.com/elastic/logstash/blob/v1.1.13/lib/logstash/filters/grok.rb#L155-L168)). It was a good idea at the time, but was a nightmare for debugging ;P

PR #4117, opened 2015-10-29
Replace "Install" with "Update" where appropriate

### LINKED ISSUE #4116: bin/plugin --help shows incorrect information for update
```
Subcommands:
    install                       Install a plugin
    uninstall                     Uninstall a plugin
    update                        Install a plugin
    list                          List all installed plugins
```

I think `update` should show `Update a plugin`.

The problem can be found [here](https://github.com/elastic/logstash/blob/a81f0249b17f84ce2ac8d8d9547b9a548bd70a7a/lib/pluginmanager/main.rb#L28), but I'm not sure when it was added.

untergeek:
fixes #4116


jordansissel:
LGTM

PR #4135, opened 2015-11-03
Fix logrotate permissions after install

### LINKED ISSUE #3490: logrotate config file invalid file mode in RPM
Hey,
This was tested under 1.5.0 and 1.5.1 RPM package and CentOS 7:
```
[vagrant@localhost ~]$ ll /etc/logrotate.d/
total 20
-rw-rw-r--. 1 root root 156 Jun 15 08:33 logstash
-rw-r--r--. 1 root root 136 Jun 10  2014 ppp
-rw-r--r--. 1 root root 210 Mar 26  2014 syslog
-rw-r--r--. 1 root root 100 Jun  9  2014 wpa_supplicant
-rw-r--r--. 1 root root 100 Jun 27  2014 yum
[vagrant@localhost ~]$ sudo logrotate -d /etc/logrotate.d/logstash
Ignoring /etc/logrotate.d/logstash because of bad file mode.

Handling 0 logs
```
fixing file mode:
```
[vagrant@localhost ~]$ sudo chmod 644 /etc/logrotate.d/logstash
[vagrant@localhost ~]$ sudo logrotate -d /etc/logrotate.d/logstash
reading config file /etc/logrotate.d/logstash

Handling 1 logs

rotating pattern: /var/log/logstash/*.log  after 1 days (7 rotations)
empty log files are not rotated, old logs are removed
considering log /var/log/logstash/*.log
  log /var/log/logstash/*.log does not exist -- skipping
```

basically Iogstash's own log file rotation doesn't work out of the box.

untergeek:
In some cases it seems that the file permissions for the logrotate
configuration file are not set properly.  This should address those
cases.

fixes #3490

### DIFF
--- a/pkg/centos/after-install.sh
+++ b/pkg/centos/after-install.sh
@@ -3,3 +3,4 @@
 chown -R logstash:logstash /opt/logstash
 chown logstash /var/log/logstash
 chown logstash:logstash /var/lib/logstash
+chmod 0644 /etc/logrotate.d/logstash


suyograo:
LGTM, not tested

talevy:
LGTM

PR #4136, opened 2015-11-03
Fix/1574

### LINKED ISSUE #1574: logrotation of logstash logs in Debian
an aside: I tried to make an account on logstash.jira.com but Atlassian claims JIRA has no more licenses available.

logstash 1.4.2-1-2c0f5a1 from logstash's Debian repository
cron logrotate error:
error: skipping "/var/log/logstash/logstash.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.
I added
'su logstash adm' (I have logstash added to group adm on my system)
to /etc/logrotate.d/logstash
and the error disappears during execution of /etc/cron.daily/logrotate

untergeek:
In some cases it seems that the file permissions for the `logrotate` configuration file are not set properly. This should address those cases.

As suggested in #1574, rotate the `logstash.err` and `logstash.stdout` files also.

fixed #1574

### DIFF
--- a/pkg/debian/after-install.sh
+++ b/pkg/debian/after-install.sh
@@ -4,3 +4,4 @@ chown -R logstash:logstash /opt/logstash
 chown logstash /var/log/logstash
 chown logstash:logstash /var/lib/logstash
 chmod 755 /etc/logstash
+chmod 0644 /etc/logrotate.d/logstash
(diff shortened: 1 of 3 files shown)


suyograo:
LGTM

talevy:
LGTM

PR #4143, opened 2015-11-05
Update release notes for breaking changes in ES

suyograo:
Ref: https://discuss.elastic.co/t/dots-in-field-names-and-elasticsearch-2-0/33791 and https://discuss.elastic.co/t/elasticseach-2-geoip-problem/33424

### DIFF
--- a/docs/asciidoc/static/upgrading.asciidoc
+++ b/docs/asciidoc/static/upgrading.asciidoc
@@ -31,3 +31,43 @@ This procedure downloads the relevant Logstash binaries directly from Elastic.
 4. Test your configuration file with the `logstash --configtest -f <configuration-file>` command. Configuration options for 
 some Logstash plugins have changed in the 2.0 release.
 5. Restart your Logstash pipeline after updating your configuration file.
+
+=== Upgrading Logstash and Elasticsearch to 2.0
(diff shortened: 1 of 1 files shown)


docs/asciidoc/static/upgrading.asciidoc
```diff
@@ -31,3 +31,43 @@ This procedure downloads the relevant Logstash binaries directly from Elastic.
...
+Note: There is one known issue (removal of https://www.elastic.co/guide/en/elasticsearch/reference/1.4/mapping-object-type.html#_path_3[path]) with using GeoIP filter that needs a manual update to the template.
+
+If you have custom template changes, please make sure to save it and merge any changes. You can 
+get the existing template by running:
+
+```
+curl -XGET localhost:9200/_template/logstash
+```
+
+Add the following option to your Logstash config:
```
untergeek: I'd include something here to the effect of:

"Even if you don't have a custom template, Logstash will not overwrite an existing template by default.  If you are not using a custom template, you can automatically update to whatever is included with Logstash by setting `template_overwrite => true`, like this:"

untergeek:
Other than the one comment, LGTM™

PR #4156, opened 2015-11-09
Add provision for worker threads based on threadsafe workers

### LINKED ISSUE #4130: Reset default worker to 1 if using non thread-safe filter
See background here: https://discuss.elastic.co/t/multiline-issue/33349

LS will fail to startup in 2.0 when using ML filter. It should instead revert to old behavior of setting workers to 1 and start the pipeline. 

Using the config specified in the above discuss link
```
suyog@machine:~/bin/releases/2.0.0/logstash-2.0.0$ bin/logstash -f discuss.conf
Error: Cannot use more than 1 filter worker because the following plugins don't work with more than one worker: multiline, multiline
You may be interested in the '--configtest' flag which you can
use to validate logstash's configuration before you choose
to restart a running system.
```

guyboertje:
**Notes to reviewers:**
- Allow the default to be set as a function of cores without checking thread safety of filters. 
- Allow the command line setting of `-w 6` without checking as above.
- Move the check and alteration of thread_count into the `start_filters` because this allows for dynamic changes to the config. Before, it was only checked at pipeline start.

After PH and Colin initial review:
- override logic from -w command line args refactored, see agent.rb and pipeline.rb

Fixes #4130

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -23,7 +23,7 @@ class LogStash::Agent < Clamp::Command
   option ["-w", "--filterworkers"], "COUNT",
     I18n.t("logstash.agent.flag.filterworkers"),
     :attribute_name => :filter_workers,
-    :default => LogStash::Config::CpuCoreStrategy.fifty_percent, &:to_i
+    :default => 0, &:to_i
 
   option ["-l", "--log"], "FILE",
     I18n.t("logstash.agent.flag.log"),
@@ -143,7 +143,7 @@ class LogStash::Agent < Clamp::Command
       configure_logging(log_file)
     end
 
-    pipeline.configure("filter-workers", filter_workers)
+    pipeline.configure("filter-workers", filter_workers) if filter_workers > 0
 
     # Stop now if we are only asking for a config test.
     if config_test?
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -46,7 +46,7 @@ class LogStash::Pipeline
     @filter_to_output = filters? ? SizedQueue.new(20) : @input_to_filter
 
     @settings = {
-      "filter-workers" => LogStash::Config::CpuCoreStrategy.fifty_percent
+      "default-filter-workers" => LogStash::Config::CpuCoreStrategy.fifty_percent
     }
 
     # @ready requires thread safety since it is typically polled from outside the pipeline thread
@@ -59,14 +59,32 @@ class LogStash::Pipeline
   end
 
   def configure(setting, value)
-    if setting == "filter-workers" && value > 1
-      # Abort if we have any filters that aren't threadsafe
-      plugins = @filters.select { |f| !f.threadsafe? }.collect { |f| f.class.config_name }
-      if !plugins.size.zero?
-        raise LogStash::ConfigurationError, "Cannot use more than 1 filter worker because the following plugins don't work with more than one worker: #{plugins.join(", ")}"
-      end
-    end
     @settings[setting] = value
+  end
+
+  def safe_filter_worker_count
+    default = @settings["default-filter-workers"]
+    thread_count = @settings["filter-workers"] #override from args "-w 8" or config
+    safe_filters, unsafe_filters = @filters.partition(&:threadsafe?)
+    if unsafe_filters.any?
+      plugins = unsafe_filters.collect { |f| f.class.config_name }
+      case thread_count
+      when nil
+        # user did not specify a worker thread count
+        # warn if the default is multiple
+        @logger.warn("Defaulting filter worker threads to 1 because there are some filters that might not work with multiple worker threads",
+          :count_was => default, :filters => plugins) if default > 1
+        1 # can't allow the default value to propagate if there are unsafe filters
+      when 0, 1
+        1
+      else
+        @logger.warn("Warning: Manual override - there are filters that might not work with multiple worker threads",
+          :worker_threads => thread_count, :filters => plugins)
+        thread_count # allow user to force this even if there are unsafe filters
+      end
+    else
+      thread_count || default
+    end
   end
 
   def filters?
@@ -149,7 +167,9 @@ class LogStash::Pipeline
 
   def start_filters
     @filters.each(&:register)
-    to_start = @settings["filter-workers"]
+    # dynamically get thread count based on filter threadsafety
+    # moved this test to here to allow for future config reloading
+    to_start = safe_filter_worker_count
     @filter_threads = to_start.times.collect do
       Thread.new { filterworker }
     end
--- a/logstash-core/lib/logstash/util/defaults_printer.rb
+++ b/logstash-core/lib/logstash/util/defaults_printer.rb
@@ -20,7 +20,7 @@ module LogStash module Util class DefaultsPrinter
     @printers.each do |printer|
       printer.visit(collector)
     end
-    "Default settings used: " + collector.join(', ')
+    "Settings: " + collector.join(', ')
   end
 
   private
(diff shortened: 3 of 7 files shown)


lib/logstash/pipeline.rb
```diff
@@ -59,16 +59,20 @@ def ready?
...
   end
 
+  def safe_filter_worker_count(desired)
+    safe_filters, unsafe_filters = @filters.partition(&:threadsafe?)
+    if unsafe_filters.any?
+      plugins = unsafe_filters.collect { |f| f.class.config_name }
+      @logger.warn("Setting filter threads to 1 because there are some filters that will not work with multiple worker threads", :filters => plugins)
+      1
+    else
+      desired
```
ph: `desired?` Not sure about that naming? Can we be a little more classy :tophat: `workers_count` or `count`?

spec/core/pipeline_spec.rb
```diff
@@ -53,20 +53,105 @@ def close
...
 
-context "close" do
+  describe "defaulting the filter workers based on thread safety" do
+    before(:each) do
+      allow(LogStash::Plugin).to receive(:lookup).with("input", "dummyinput").and_return(DummyInput)
+      allow(LogStash::Plugin).to receive(:lookup).with("codec", "plain").and_return(DummyCodec)
+      allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(DummyOutput)
+      allow(LogStash::Plugin).to receive(:lookup).with("filter", "dummyfilter").and_return(DummyFilter)
+      allow(LogStash::Plugin).to receive(:lookup).with("filter", "dummysafefilter").and_return(DummySafeFilter)
+      allow(LogStash::Config::CpuCoreStrategy).to receive(:fifty_percent).and_return(8)
```
ph: can we extract the number of core in a `let` so we can reuse the same value in the expectation?

spec/core/pipeline_spec.rb
```diff
@@ -53,20 +53,105 @@ def close
...
+        output {
+          dummyoutput {}
+        }
+        eos
+      }
+
+      it "starts eight filter threads" do
+        pipeline = TestPipeline.new(test_config_with_filters)
+        pipeline.run
+        expect(pipeline.filter_threads.size).to eq(8)
```
ph: See previous comment on extracting that value.

lib/logstash/pipeline.rb
```diff
@@ -59,16 +59,20 @@ def ready?
...
-      end
-    end
     @settings[setting] = value
   end
 
+  def safe_filter_worker_count(desired)
+    safe_filters, unsafe_filters = @filters.partition(&:threadsafe?)
+    if unsafe_filters.any?
+      plugins = unsafe_filters.collect { |f| f.class.config_name }
+      @logger.warn("Setting filter threads to 1 because there are some filters that will not work with multiple worker threads", :filters => plugins)
```
ph: > Setting filter *workers* threads with

 We use the workers naming  covention in the doc, I think we should reuse it in the warning.

lib/logstash/pipeline.rb
```diff
@@ -149,7 +153,9 @@ def start_inputs
 
   def start_filters
     @filters.each(&:register)
-    to_start = @settings["filter-workers"]
+    # dynamically get thread count based on filter threadsafety
+    # moved this test to here to allow for future config reloading
+    to_start = safe_filter_worker_count(@settings["filter-workers"])
```
ph: Good catch to move it here.

ph:
LGTM

spec/core/pipeline_spec.rb
guyboertje: Sure - you are correct, it should be in a let block

lib/logstash/pipeline.rb
guyboertje: True - will change. Nice catch.

lib/logstash/pipeline.rb
guyboertje: Well spotted.  I will change to `thread_count`.

colinsurprenant:
does explicitly setting the `-w X` on the command line overrides the count regardless of non threadsafe plugins configured?

guyboertje:
@colinsurprenant - not that I know of.  AFAIK, by the time such a setting hits the pipeline code I can't say where the setting came from.

That said, I suppose I can find where it is in the args processing and do something special with it in terms of the key used in the @settings hash.  However I can reverse this and put the default in a special key instead and process differentially.

colinsurprenant:
so the question is, should we allow users to override the forced 1 worker setting when non threadsafe plugins are discovered (and log a big bold warning)? Personally I think so and I believe this is what we [agreed on](https://github.com/elastic/logstash/issues/4130#issuecomment-153422493).

guyboertje:
yes you are correct.  And also if the user wants to force to 1 when the default could be > 1.

Will fix.  Bigger edit though.

colinsurprenant:
:+1:

ph:
@guyboertje this is ready for a review?

guyboertje:
I was hoping @colinsurprenant would review.

ph:
@guyboertje yes but since the code change a bit since my last review, I will have to go through it again?

lib/logstash/pipeline.rb
```diff
@@ -59,14 +59,32 @@ def ready?
...
-      plugins = @filters.select { |f| !f.threadsafe? }.collect { |f| f.class.config_name }
-      if !plugins.size.zero?
-        raise LogStash::ConfigurationError, "Cannot use more than 1 filter worker because the following plugins don't work with more than one worker: #{plugins.join(", ")}"
+    @settings[setting] = value
+  end
+
+  def safe_filter_worker_count
+    default = @settings["default-filter-workers"]
+    thread_count = @settings["filter-workers"] #override from args "-w 8" or config
+    safe_filters, unsafe_filters = @filters.partition(&:threadsafe?)
```
ph: Since the logic only check for `unstafe_filters`, `#select` would be sufficient?
Maybe this is a personal style.

spec/util/worker_threads_default_printer_spec.rb
```diff
@@ -3,24 +3,41 @@
...
 
 describe LogStash::Util::WorkerThreadsDefaultPrinter do
-  let(:settings) { {} }
+  let(:settings)  { {} }
   let(:collector) { [] }
 
   subject { described_class.new(settings) }
 
+  before { subject.visit(collector) }
+
```
ph: Maybe adding a `context "#visit"` would make the "it" less redundant?

ph:
Tested manually the different use case and it work as describe.

Minor comments, other than that LGTM.

lib/logstash/pipeline.rb
guyboertje: I want to keep this if possible.  They are temporary arrays of pointers to filters instances.  We can perhaps look at the ratio of safe vs unsafe later and partition is an efficient way of splitting them. WDYT?

spec/util/worker_threads_default_printer_spec.rb
guyboertje: Ok I'll go with that.  BTW, I always try to contexts for variants. i.e. the text starts with `when` e.g.
`context 'when the settings hash has only user content' do`
I will use a `describe` block.

lib/logstash/pipeline.rb
ph: +1 dont change it

ph:
I like how you did the reporting of the values, I think we can reuse that flow it in a lot of places.

lib/logstash/pipeline.rb
guyboertje: OK.

guyboertje:
Thanks - re the defaults printer system.  Yes it is a pattern with a lot of mileage.

colinsurprenant:
LGTM :shipit:

suyograo:
@guyboertje rebase and merge please

Issue #4163, opened 2015-11-10
Tins 1.7 dependency from coveralls requires ruby 2.0


ph:
Tins is a dependency from `coveralls` which is only installed when running the tests.

On November 10, Tins 1.7 was released and they now require ruby 2.0 in their gemspec (https://github.com/flori/tins/blob/master/tins.gemspec#L19).

This change is breaking our build, since coveralls only need 1.* I think its safe to pin the tins version on 1.6.

PR #4164, opened 2015-11-10
Fix tins to 1.6

ph:
The 1.6 release is the latest release to support ruby 1.9

Fixes: #4163

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,10 @@ gem "file-dependencies", "0.1.6"
 gem "ci_reporter_rspec", "1.0.0", :group => :development
 gem "simplecov", :group => :development
 gem "coveralls", :group => :development
+# Tins 1.7 requires the ruby 2.0 platform to install,
+# this gem is a dependency of term-ansi-color which is a dependency of coveralls.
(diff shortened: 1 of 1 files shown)


colinsurprenant:
LGTM 
maybe add a comment explaining that tins a requirement of term-ansi-color which is in turn a requirement of coveralls and 1.6 release is required until we switch to a Ruby 2.0 compliant JRuby ?

colinsurprenant:
this is such a straightforward fix I am comfortable with a quick merge, plus I am stumbling on this now :P

ph:
+1 will update the PR with the comment.

ph:
@colinsurprenant concerning the comment where do you want it? In the Gemfile there is risk it will be removed?

`# If you modify this file manually all comments and formatting will be lost.` :)

colinsurprenant:
its not a problem, the comment will be in the repo, it will be lost when executing plugin manager commands but that's not important. what's important is that we have the comment in the source tree repo

ph:
Updated with a comment,  
This need to be merged in master 1.5, 1.5, 2.0, 2.1

Issue #4063, opened 2015-10-19
Add simple integration test infra to Logstash core


suyograo:
We have good system wide integration tests in Jenkins using [logstash-kitchen](https://github.com/elastic/logstash-kitchen) repo. These are end-end tests infrastructure which validate things like installing/uninstalling plugins, smoke tests for ES, file input, Kafka etc. 

We also need a simple integration test bed in Logstash core which tests multiple combinations of input, filters, codec and output using various test data sets. We should aim to keep the inputs and outputs simple, maybe just use the generator input (or file input) and file output. This will help us catch issues like #4048 which could be reproduced using file input, grok filter and json codec.

Suggest using `spec/integration` dir for adding these tests. Should also add `rake tests:integration-plugins` which will install dependencies, run these simple integration tests. 

Ultimately, it should be easy enough to drop in a config file plus a results (expected output) fixture and the rest should be taken care of by the infra.

guyboertje:
At one time I considered storing the config and expectations in ES, but in attempting to be self-contained we could store a list of config and expectations in a (Yaml JSON?) file and have a runner that worked through them with a bit of help from RSpec.

purbon:
+1 on this idea @suyograo, for this business I see, as you said, a collection of rspec tests inside ```spec/integrations```, this test should test the filter integrations, test kitchen should become the external systems integration tests.

On a more detailed stuff we can leverage some of the actual helpers, that with a bit of more love, will let pass a configuration to the pipeline (basically a collection of filters), and get the output. 

This should not be a complicated task to archive.

Speaking about

> Ultimately, it should be easy enough to drop in a config file plus a results (expected output) fixture and the rest should be taken care of by the infra.

is hard for me to see it right now, especially because drop in test are hard then to fetch context, we should clearly define the interfaces and the report outcome for it so the test results are nice to debug and work with.

ph:
+1 on spec/integration and kitchen for isolated command testing.

colinsurprenant:
+1

I guess a lots of tests involving codecs and filters can be done easily using stdin/stdout, we could maybe look at reusing the perf framework and add the ability to compare output for validation?

It would be nice to be able to easily and "quickly" run a good set of integration tests through rake so that we can integrate this in our "tdd" loop. I would suggest we first focus on this and make it as efficient as possible. Then we can add move involved integration tests using test kitchen for example. 

Thoughts?

purbon:
[...]

PR #4166, opened 2015-11-10
New test for integration testing

ph:
Last year I have created `logstash-kitchen` https://github.com/elastic/logstash-kitchen
To have a minimal integration testing on top of the plugin manager and some frequently used plugins.
I was useful at the time when we were messing with the api.

**This implementation had a few problems:**

- It was external to Logstash making it harder to maintain.
- It was slow ~15-17m to run on my machine
- It was hard to run the test in isolations, you had to run the full test suite.
- Using the plugin command is destructive we needed some kind of isolation.
- Reproducing errors and debugging was hard.

**I migrated theses tests inside logstash and changed how it behave, a few of the changes.**

- The tests are inside logstash making them linked to a specific version of the code. (Could be run on pull request)
- Since theses tests use the plugins manager It can affect the developer environment. I use `rsync` to clone the current directory to a temporary directory where the tests are executed. This change has two benefits, It leave the current developer installation untouched but it allow you to see how the logstash installation modifies the files and this operation is fast.
- Using a temporary directory allow you to reproduce the issue manually with the same setup as the test.
- The tests can also be run in the current directory without requiring the need for rsync, this is useful for Jenkins.
- Running the tests takes **3** minutes on my machine (down from 17 minutes)
- bin/rspec works with this new setup, this mean you can run the tests in isolation like this.
`bin/rspec integration_spec/logstash_config/file_input_to_file_output_spec.rb:15` They are normal spec files.
- You will be able to use `longshoreman` to run services with a docker image. (like elasticsearch)
- I have consolidated some expectations in the same example to reduce the number of external process to start.

**Caveats**
- When you write new specs they will need to manage their own dependencies.
- Make multiples `expects` on the same result, this will reduce the number of external calls and make the suite faster to run.
- The temporary directory is inside logstash, I am open to put it somewhere else.

**What is left:**
- The elasticsearch integration test doesn't work, I need help here to setup docker/longshoreman to fix it.
- This setup could possibly work with drip to make it faster, it doesn't currently work on my machine but I didn't investigate it too much.

**Expectations:**
This  is a WIP in progress, I have copied all the tests from kitchen and moved it inside Logstash and fixed them to be able to run in this new environment. I know the test suite could be improved this is not the goal of this PR.  I want feedback on the way I have made it work and to see if people are interested in using it and improving it.

I am aware that some files need cleanup, (ie `logstash_helpers`) this will be done when the WIP tags is removed.

**How to make it run**

On a correctly bootstrapped test environment run this command: `rake test:integration:local` this task will setup the initial environment. 

If you call `bin/rspec` on a suite, it will check if the `integration_run` directory exist, if it exist it will use it as the target of the test.


purbon:
@ph good work here. I wonder why having this under another directory? why not having them under ```spec/integration```? and other others under ```spec/unit```?

suyograo:
@ph @purbon +1 on having it under `spec/integration`. We did discuss this in #4063

ph:
I am ok with spec/integration,  I was expecting a need for a custom gemfile but this isn't the case

integration_spec/spec_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  LOGSTASH_TEST_PATH = File.join(integration_path, "logstash")
+else
+  LOGSTASH_TEST_PATH = File.join(File.dirname(__FILE__), "..")
+end
+puts "[integration_spec] Running the test in #{LOGSTASH_TEST_PATH}"
+puts "[integration_spec] Running specs"
+
+RSpec.configure do |config|
+  config.order = "random"
```
purbon: why not adding here the Logstash test helpers you wrote? so we are just including them as regular helpers?

rakelib/test.rake
```diff
@@ -87,6 +87,28 @@ namespace "test" do
     task.reenable
   end
 
+  task "integration" => ["setup"] do
```
purbon: I wonder if we should add here something like for others test, that they do the bootstrap automatically and you only need to run ```rake test:integration```, what do you think?

purbon:
@ph have you thought to create some custom matchers? they might help to simplify the test semantics and also might make string matching easier.

ph:
+1 for custom matcher but I think its should be done in another PR.

integration_spec/spec_helper.rb
ph: I would prefer refactor that to custom helpers in another PR.

rakelib/test.rake
ph: If you depends on bootstrap task, you will run into `bundler` issues, trust me I didn't learn my lesson the first time and tried to make it work. :cry: 

If you look at the current workflow of the tests they don't bootstrap automatically.

rake bootstrap https://github.com/elastic/logstash/blob/master/ci/ci_setup.sh#L17
rake test:install-core https://github.com/elastic/logstash/blob/master/ci/ci_test.sh#L12
rake test:core https://github.com/elastic/logstash/blob/master/ci/ci_test.sh#L13

If you look at the `test:core` it depends on setup too.
https://github.com/ph/logstash/blob/fix/integration-test/rakelib/test.rake#L23-L25

ph:
Almost done with that only a weird dependency issue on rake with the move. >:)

ph:
@purbon @suyograo Should I move the other test to `spec/unit` ?
So the directory structure would look like this.

```
spec
 - integration
   - plugin_manager
   - config
- unit
 - core
 - logstash
- [....]

**Comment by purbon** [comment]
@ph yes I like it.

**Comment by ph** [comment]
@purbon @suyograo  can you review this?

**Comment by colinsurprenant** [comment]
I don't like the proposed dir structure. integration tests and unit tests are not specs, why have them under spec? if we absolutely want a single root test directory (why? is there a directory shortage?) then having all under `test/` seems to make more sense: 
```
test/
test/spec
test/integration
```

why not just keep root dirs? I don't see any problem with that?? 
```
spec/
integration/
```

**Comment by ph** [comment]
@colin in the current PR I did not implemented the mentioned structure. Since this would have created a lot of conflict with your PRs.

I have decided to do a `spec/integration` and change the `rake test:core` to skip them, following the review I had.

I don't have a strong opinion on where it should be. And if you look at first commit https://github.com/ph/logstash/commit/1aa2539d2f7321d06706ef562585457771c93bb2 it was mostly where you are suggesting.

**Comment by ph** [comment]
@purbon updated with an `integration` folder at the root.

File: ci/ci_integration.sh [hunk 3]
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+rake test:install-all
```
purbon: I wonder if install all or only default plugins? what do you think?

ph: I believe default plugin would be sufficient and will go around some issue with have with uninstalling plugins that has cross dependencies.

purbon: Yes I think I will just change it to the ones we bundle logstash with. makes more sense to me.

integration/logstash_config/file_input_to_file_output_spec.rb
```diff
@@ -0,0 +1,25 @@
+# Encoding: utf-8
+require_relative "../spec_helper"
+require "stud/temporary"
+
+describe "File input to File output" do
+  let(:number_of_events) { IO.readlines(sample_log).size }
+  let(:sample_log) { File.expand_path(File.join(File.dirname(__FILE__), "..", "supports", "sample.log")) }
+  let(:output_file) { Stud::Temporary.file.path }
+  let(:config) { "input { file { path => \"#{sample_log}\" stat_interval => 0 start_position => \"beginning\" sincedb_path => \"/dev/null\" } } output { file { path => \"#{output_file}\" } }" }
```
purbon: can we break this line somehow? makes reading it complicated :smile_cat:

integration/plugin_manager/plugin_install_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+context "bin/plugin install" do
+  context "with a local gem" do
+    let(:gem_name) { "logstash-input-wmi" }
+    let(:local_gem) { gem_fetch(gem_name) }
+
+    it "install the gem succesfully" do
+      result = command("bin/plugin install --no-verify #{local_gem}")
+      expect(result.exit_status).to eq(0)
+      expect(result.stdout).to match(/Installing #{gem_name}/)
```
purbon: I wonder if it makes sense to do this matches all in one line? so we track also the text format? what do you think?

ph: Are you saying creating an example for the `exist_status` and another one for the stdout?

The problem I see with, is that let/subject aren't cached between example, which mean will have the overhead of launching 2 logstash.

ph: Maybe I did not understand your issue?

integration/logstash_config/file_input_to_file_output_spec.rb
ph: +1

integration/plugin_manager/plugin_update_spec.rb
```diff
@@ -0,0 +1,32 @@
+# Encoding: utf-8
+require_relative "../spec_helper"
+
+describe "update" do
+  let(:plugin) { "logstash-input-stdin" }
+  let(:previous_version) { "2.0.1" }
+
+  before do
+    command("bin/plugin install --version #{previous_version} #{plugin}")
+    cmd = command("bin/plugin list --verbose logstash-input-stdin")
```
purbon: we should use also the plugin variable here.

integration/plugin_manager/plugin_update_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    expect(cmd.stdout).to match(/#{plugin} \(#{previous_version}\)/)
+  end
+
+  context "update a specific plugin" do
+    subject { command("bin/plugin update logstash-input-stdin") }
```
purbon: same thing here

integration/plugin_manager/plugin_update_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      expect(subject.stdout).to match(/Updating #{plugin}/)
+    end
+  end
+
+  context "update all the plugins" do
+    subject { command("bin/plugin update") }
+
+    it "has executed succesfully" do
+      expect(subject.exit_status).to eq(0)
+      cmd = command("bin/plugin list --verbose logstash-input-stdin").stdout
```
purbon: and here.

integration/spec_helper.rb
```diff
@@ -0,0 +1,31 @@
+# encoding: utf-8
+require_relative "supports/logstash_test_helpers"
```
purbon: why is this directory call ```supports```? does it makes sense or is a typo?

purbon: also I don't think ```logstash_test_helpers``` would a good name, what about ``logstash_integration_test_helpers```? I mean helpers there are only for integration test, right?

integration/spec_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+source = File.expand_path(File.join(File.dirname(__FILE__), ".."))
+integration_path = File.join(source, "integration_run")
+
+puts "[integration_spec] configure environment"
+
+if Dir.exists?(integration_path)
+  # We copy the current logstash into a temporary directory
+  # since the tests are a bit destructive
+  FileUtils.mkdir_p(integration_path)
+  rsync_cmd = "rsync -a --exclude 'rspec' --exclude '#{File.basename(integration_path)}' --exclude 'integration_spec' --exclude '.git' #{source} #{integration_path}"
```
purbon: this will not work on windows, right? can we do some flag checking somewhere?

integration/spec_helper.rb
ph: What is wrong with `supports?` Remove the s, rename it?

lets drop the logstash and called it: `integration_test_helpers`

integration/spec_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+require_relative "supports/logstash_test_helpers"
+require "fileutils"
+
+# Configure the test environment
+source = File.expand_path(File.join(File.dirname(__FILE__), ".."))
+integration_path = File.join(source, "integration_run")
+
+puts "[integration_spec] configure environment"
+
+if Dir.exists?(integration_path)
```
purbon: I wonder if we should add  warning here and simply break. I imagine running a test, failing and then willing to run them again with the same dir. makes sense?

integration/spec_helper.rb
ph: Agree, we have specific tests for windows.

I can raise an error in the `spec_helper` when we run this on windows?

integration/supports/logstash_test_helpers.rb
```diff
@@ -0,0 +1,102 @@
...
+  @@cached = {}
+
+
+  def self.latest_version(name)
+    get_rubygems_specification(name).first.fetch("number")
+  end
+
+
+  def self.get_rubygems_specification(name)
+    @@cached[name] ||= make_api_call("https://rubygems.org/api/v1/versions/#{name}.json")
```
purbon: we can use gems probably here? does it makes sense not use it? what do you think?

rakelib/test.rake
purbon: @ph sure I know this script, I wrote it :smile:, ok I like your argument.

integration/spec_helper.rb
ph: Not sure I understand here.

Since the `integration_run` should be the same as the root directory because of the rsync.
If you run the test again you they hopefully will get the same results, no?

I have added the check on the `Dir.exist?` to allow people to run their tests with `bin/rspec file:number`, which is more convenient and a lot faster to run a portion of the suite.

I agree that this may be a bit of magic, but I think it give flexibility.

integration/supports/logstash_test_helpers.rb
```diff
@@ -0,0 +1,102 @@
...
+  sleep(30)
+  begin
+    Process.kill("INT", pid)
+  rescue
+  end
+end
+
+module LogStashTestHelpers
+  RANDOM_PREVIOUS_VERSION_LIMIT = 3
+  @@cached = {}
```
colinsurprenant: why use a `@@` class variable in a module that is meant as a singleton?

ph: +1 this shouldnt be the case.

jordansissel:
Finished in 4 minutes 27.1 seconds (files took 4.38 seconds to load)
15 examples, 0 failures

Tests pass for me. Will continue reviewing :)

integration/plugin_manager/plugin_install_spec.rb
purbon: no, no I even If my formal side would prefer it, I see the benefit of what you did here, no questions on it.

My point was on extending the match regular expression, now it looks for a sentence, but it does not care where is it printed, could be that ```Installing gem_name``` is not in the right place. Does it makes sense to check that too? what do you think?

Doing it would be as easy as making the expression a bit more complex, but not more.

integration/spec_helper.rb
purbon: +1 on ```integration_test_helpers```

re ```supports```: I understand it as plural, I think is better to use singular names, in this case ```support```. You know stupid naming :+1:

integration/spec_helper.rb
purbon: yes please! raising an error would be nice.

integration/spec_helper.rb
purbon: ok, makes sense.

integration/spec_helper.rb
ph: +1 for singular

ph:
updated with your comments @purbon

purbon:
@ph LGTM

ph:
@jordansissel can you give it a LGTM?

integration/spec_helper.rb
jordansissel: I think we probably want to use the `--delete` flag on rsync, but I'm OK waiting until we find a need for it first.

jordansissel:
LGTM

integration/spec_helper.rb
ph: I will add it, dont think it will slow it down that much.

Issue #4173, opened 2015-11-11
Fix concurrent-ruby to 0.9.2 in the gemspec


ph:
We have removed the patch in 972e13d52e103dfaf5f8ab5c33ceb852402cde49, but we did not set the version to 0.9.2

ph:
Fixed by

Issue #4141, opened 2015-11-05
concurrent-ruby gem pin is not precise enough


jordansissel:
```
# From a plugin
% bundle exec rspec
RuntimeError: Logstash expects concurrent-ruby version 0.9.1 and version 0.9.2 is installed, please verify this patch: /Users/
jls/.rvm/gems/jruby-1.7.22/gems/logstash-core-2.0.0-java/lib/logstash/patches/silence_concurrent_ruby_warning.rb
```

The problem is that `bundle install` fetched 0.9.2 because our dependency on `~> 0.9.1` and our concurrent ruby warning thingy patch barfs if we are using anything but 0.9.1

jordansissel:
(this only affects master, really)

jordansissel:
I was going to send a patch, but I'm not entirely sure at a glance if we should pin exactly 0.9.1 (in gem dependency) or relax the silence_concurrent_ruby patch to support 0.9.x

jordansissel:
Over to @ph for some insight (based on authorship of b863ace5)

ph:
Oh oh :(

ph:
When we merged the shutdown changes the dependencie for concurrent-ruby was also changed. I think there are really close to 1.0

ph:
I will review if they changed their warnings.

Russell-IO:
@jordansissel - this affects logstash 2.0.0 & 1.5.5 as far as I can tell.
root@ip-10-12-0-62:/var/log/logstash# /opt/logstash/bin/logstash agent -f /opt/logstash-configuration/green/conf.d/v2/output/apache/
RuntimeError: Logstash expects concurrent-ruby version 0.9.1 and version 0.9.2 is installed, please verify this patch: /opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.0.0-java/lib/logstash/patches/silence_concurrent_ruby_warning.rb
   (root) at /opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.0.0-java/lib/logstash/patches/silence_concurrent_ruby_warning.rb:53
  require at org/jruby/RubyKernel.java:1040
   (root) at /opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.0.0-java/lib/logstash/patches.rb:1
  require at org/jruby/RubyKernel.java:1040
[...]

PR #4174, opened 2015-11-11
Strict dependency on concurrent-ruby 0.9.2

ph:
Followup on #4141 and #4173


colinsurprenant:
LGTM, straightforward fix, :shipit:

PR #4182, opened 2015-11-12
Add config file output to --configtest

### LINKED ISSUE #3243: allow outputting resulting config file when merging multiple ones
```
./bin/logstash --show-config -f /my_config_files_dir 
//don't start just show resulting config
input{
//all merged inputs
}
filter{
//all merged filters
}
output{
//all merged outputs
}
```

untergeek:
This only works when the `--debug` flag is also enabled.

fixes #3243

### DIFF
--- a/lib/logstash/agent.rb
+++ b/lib/logstash/agent.rb
@@ -312,18 +312,26 @@ class LogStash::Agent < Clamp::Command
     Dir.glob(path).sort.each do |file|
       next unless File.file?(file)
       if file.match(/~$/)
-        @logger.debug("NOT reading config file because it is a temp file", :file => file)
+        @logger.debug("NOT reading config file because it is a temp file", :config_file => file)
         next
       end
-      @logger.debug("Reading config file", :file => file)
+      @logger.debug("Reading config file", :config_file => file)
       cfg = File.read(file)
       if !cfg.ascii_only? && !cfg.valid_encoding?
         encoding_issue_files << file
       end
       config << cfg + "\n"
+      if config_test?
+        @logger.debug? && @logger.debug("\nThe following is the content of a file", :config_file => file.to_s)
+        @logger.debug? && @logger.debug("\n" + cfg + "\n\n")
+      end
     end
     if (encoding_issue_files.any?)
       fail("The following config files contains non-ascii characters but are not UTF-8 encoded #{encoding_issue_files}")
+    end
+    if config_test?
+      @logger.debug? && @logger.debug("\nThe following is the merged configuration")
+      @logger.debug? && @logger.debug("\n" + config + "\n\n")
     end
     return config
   end # def load_config


untergeek:
For reference, this creates output like this: https://gist.github.com/untergeek/e31bedbe5b34f21d1cba

lib/logstash/agent.rb
```diff
@@ -321,10 +321,20 @@ def local_config(path)
         encoding_issue_files << file
       end
       config << cfg + "\n"
+      if config_test?
+        @logger.debug? && @logger.debug("\n***** Begin configuration file: #{file}")
```
jordansissel: Why the newlines?

Also, any thoughts on making this look more like a valid config, so in this case using a comment `#` symbol to prefix the '***** Begin con....' lines?

jordansissel: Let's use structured logging here, too.

`@logger.debug("whatever...", :file => file)`

untergeek: The newlines make it more readable in debug log output, IMO.

I can definitely add the `#`, that's a great idea.  Allows cut/paste testing (but that was the other reason for the newlines).

untergeek:
> Let's use structured logging here, too.
>
>`@logger.debug("whatever...", :file => file)`

I tried that, but it did not output the file readably.  That's the other reason this _only_ outputs like this for `--configtest` _and_ `--debug` output.  People aren't likely to log this to Logstash.  Just stdout or file, for sending to support or a gist.

untergeek:
@jordansissel This is what I get when I try structured logging:

```
Whatever... {:config=>"input {\n  tcp {\n    port => 9500\n  }\n}\n\ninput {\n  tcp {\n    port => 9601\n  }\n}\n\ninput {\n  tcp {\n    port => 9702\n  }\n}\n\ninput {\n  tcp {\n    port => 9399\n  }\n} \n\nfilter {\n  mutate {\n    add_field => { \"a\" => \"b\" }\n  }\n}\n\n\ninput {\n  tcp {\n    port => 9288\n  }\n} \n\nfilter {\n  mutate {\n    add_field => { \"c\" => \"d\" }\n  }\n}\n\n\noutput {\n  stdout { }\n}\n\n", :level=>:debug, :file=>"logstash/agent.rb", :line=>"338", :method=>"local_config"}
```

That's why I went the other way.

jordansissel:
@untergeek I meant specifically for the header-style lines, where you have roughly "The following is the content of a file named #{file}" - my recmomendation is `"The following is the content of a file", :file => file`

jordansissel:
For the actual config itself, I am OK if that appears in the text, like, this line that you have is OK by me:

```diff
+        @logger.debug? && @logger.debug("\n" + cfg + "\n\n")
```

untergeek:
Ah, that.  Yeah, I can address that.

jordansissel:
Otherwise LGTM. Feel free to merge after the `:file => file` change is done. <3

untergeek:
Interesting.  It appears another code change needs to happen a few lines up in 315:

```ruby
@logger.debug("NOT reading config file because it is a temp file", :file => file)
```

I cannot use `:file` as my symbol, because that gets overridden by `logstash/agent.rb` every time.  I'm using `:config_file` instead.

jordansissel:
@untergeek haha oops, yes. Good catch.

PR #4194, opened 2015-11-16
Pluggable agents

andrewvc:
This builds off @jsvd 's patch #3872 to make the agent a pluggable command line class for greater extensibility. It will need to be either merged into @jsvd 's branch or rebased after his branch is merged in the future. 

## Notes for Reviewers

This should enable the following behaviors

1. Plugin authors can create an agent plugin by creating a gem named `logstash-agent-myagent` and creating a class that looks like
2. Users should be able to choose the agent they desire at startup with a new CLI flag `-a AGENTNAME`. 

```ruby
require 'lib/logstash/agent_plugin_manager'
  class MyLogStashAgent < LogStash::Agent
    LogStash::AgentPluginManager.register(:my_logstash_agent, self)
   # ...
end
```

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -10,38 +10,23 @@ require "stud/trap"
 LogStash::Environment.load_locale!
 
 class LogStash::Agent
-  attr_reader :logger, :runner
+  attr_reader :logger
 
-  def initialize(logger, runner)
+  def initialize(logger)
     @logger = logger
-    @runner = runner
     @pipelines = {}
   end
 
-  def execute
-    add_pipelines
-    run_pipelines
-  end
-
-  def config_valid?
+  def config_valid?(config_str)
     begin
-      add_pipelines
+      # There should be a better way to test this ideally
+      LogStash::Pipeline.new(config_str)
     rescue Exception => e
       e
     end
   end
 
-  def add_pipelines
-    if (runner.config_string.nil? || runner.config_string.empty?) && runner.config_path.nil?
-      fail(I18n.t("logstash.runner.missing-configuration"))
-    end
-
-    config_string = LogStash::Config::Loader.format_config(runner.config_path, runner.config_string)
-
-    add_pipeline("base", config_string, :filter_workers => runner.filter_workers)
-  end
-
-  def run_pipelines
+  def execute
     # Make SIGINT/SIGTERM shutdown the pipeline.
     sigint_id = trap_sigint()
     sigterm_id = trap_sigterm()
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -3,20 +3,45 @@ require "logstash/environment"
 require "logstash/errors"
 require "logstash/config/cpu_core_strategy"
 require "logstash/pipeline"
+require "logstash/config/loader"
 require "uri"
 require "stud/trap"
 
 LogStash::Environment.load_locale!
 
 class LogStash::Agent
+  attr_reader :logger, :runner
 
-  attr_writer :logger
-
-  def initialize
+  def initialize(logger, runner)
+    @logger = logger
+    @runner = runner
     @pipelines = {}
   end
 
   def execute
+    add_pipelines
+    run_pipelines
+  end
+
+  def config_valid?
+    begin
+      add_pipelines
+    rescue Exception => e
+      e
+    end
+  end
+
+  def add_pipelines
+    if (runner.config_string.nil? || runner.config_string.empty?) && runner.config_path.nil?
+      fail(I18n.t("logstash.runner.missing-configuration"))
+    end
+
+    config_string = LogStash::Config::Loader.format_config(runner.config_path, runner.config_string)
+
+    add_pipeline("base", config_string, :filter_workers => runner.filter_workers)
+  end
+
+  def run_pipelines
     # Make SIGINT/SIGTERM shutdown the pipeline.
     sigint_id = trap_sigint()
     sigterm_id = trap_sigterm()
--- a/logstash-core/lib/logstash/agent_plugin_manager.rb
+++ b/logstash-core/lib/logstash/agent_plugin_manager.rb
@@ -1,0 +1,65 @@
+# encoding: utf-8
+require "logstash/agent"
+
+module LogStash module AgentPluginManager
+  DEFAULT_AGENT_NAME = :default
+  class DuplicatePluginNameError < Error; end
+
+  REGISTRY = {}
+
+  # Reset plugin registry to just the default plugin
+  def self.reset!
+    REGISTRY.clear
+    REGISTRY[DEFAULT_AGENT_NAME] = LogStash::Agent
+  end
+
+  reset!
+
+  # Search gems for available plugins and load their libs
+  def self.load_all
+    find_plugins.each do |plugin|
+      name = plugin.name.split('-')[-1]
+      require "logstash/agents/#{name}"
+    end
+  end
+
+  # Return a list of Gem::Specification s that start with logstash-agent-
+  def self.find_plugins
+    Gem::Specification.find_all{|spec| spec.name =~ /logstash-agent-/ }
+  end
+
+  # To be called by a plugin when its class is first loaded
+  # Plugins should call this with the following code:
+  #
+  # require 'lib/logstash/agent_plugin_manager'
+  #
+  # class MyLogStashAgent < LogStash::Agent
+  #   LogStash::AgentPluginManager.register(:my_logstash_agent, self)
+  #
+  #   # ...
+  # end
+  def self.register(name, plugin_class)
+    name_sym = name.to_sym
+
+    if (conflicting_class = registry[name_sym])
+      raise DuplicatePluginNameError, "Could not register plugin '#{plugin_class}' as '#{name}', this name is already taken by '#{conflicting_class}'"
+    end
+
+    registry[name_sym] = plugin_class
+  end
+
+  # A hash of plugin names to plugin classes
+  def self.registry
+    REGISTRY
+  end
+
+  # Get a plugin by name
+  def self.lookup(name)
+    registry[name.to_sym]
+  end
+
+  # List of available plugins
+  def self.available
+    registry.keys
+  end
+end end
--- a/logstash-core/lib/logstash/agent_plugin_registry.rb
+++ b/logstash-core/lib/logstash/agent_plugin_registry.rb
@@ -1,9 +1,9 @@
 # encoding: utf-8
 require "logstash/agent"
 
-module LogStash module AgentPluginManager
+module LogStash module AgentPluginRegistry
   DEFAULT_AGENT_NAME = :default
-  class DuplicatePluginNameError < Error; end
+  class DuplicatePluginError < Error; end
 
   REGISTRY = {}
 
@@ -31,10 +31,10 @@ module LogStash module AgentPluginManager
   # To be called by a plugin when its class is first loaded
   # Plugins should call this with the following code:
   #
-  # require 'lib/logstash/agent_plugin_manager'
+  # require 'lib/logstash/agent_plugin_registry'
   #
   # class MyLogStashAgent < LogStash::Agent
-  #   LogStash::AgentPluginManager.register(:my_logstash_agent, self)
+  #   LogStash::AgentPluginRegistry.register(:my_logstash_agent, self)
   #
   #   # ...
   # end
@@ -42,7 +42,8 @@ module LogStash module AgentPluginManager
     name_sym = name.to_sym
 
     if (conflicting_class = registry[name_sym])
-      raise DuplicatePluginNameError, "Could not register plugin '#{plugin_class}' as '#{name}', this name is already taken by '#{conflicting_class}'"
+      raise DuplicatePluginError, "Could not register plugin '#{plugin_class}'" <<
+        " as '#{name}', this name is already taken by '#{conflicting_class}'"
     end
 
     registry[name_sym] = plugin_class
--- a/logstash-core/lib/logstash/agent_plugin_registry.rb
+++ b/logstash-core/lib/logstash/agent_plugin_registry.rb
@@ -12,7 +12,7 @@ module LogStash module AgentPluginRegistry
     REGISTRY.clear
     REGISTRY[DEFAULT_AGENT_NAME] = LogStash::Agent
   end
-
+  # Just have the defaults when the process starts
   reset!
 
   # Search gems for available plugins and load their libs
--- a/logstash-core/lib/logstash/agent_plugin_registry.rb
+++ b/logstash-core/lib/logstash/agent_plugin_registry.rb
@@ -25,7 +25,7 @@ module LogStash module AgentPluginRegistry
 
   # Return a list of Gem::Specification s that start with logstash-agent-
   def self.find_plugins
-    Gem::Specification.find_all{|spec| spec.name =~ /logstash-agent-/ }
+    ::Gem::Specification.find_all{|spec| spec.name =~ /logstash-agent-/ }
   end
 
   # To be called by a plugin when its class is first loaded
--- a/logstash-core/lib/logstash/config/defaults.rb
+++ b/logstash-core/lib/logstash/config/defaults.rb
@@ -6,6 +6,14 @@ module LogStash module Config module Defaults
 
   extend self
 
+  def input
+    "input { stdin { type => stdin } }"
+  end
+
+  def output
+    "output { stdout { codec => rubydebug } }"
+  end
+
   def cpu_cores
     Concurrent.processor_count
   end
--- a/logstash-core/lib/logstash/config/loader.rb
+++ b/logstash-core/lib/logstash/config/loader.rb
@@ -1,7 +1,16 @@
 require "logstash/config/defaults"
 
-module LogStash; module Config; module Loader
-  def self.format_config(config_path, config_string)
+module LogStash; module Config; class Loader
+  def initialize(logger, is_config_test)
+    @logger = logger
+    @is_config_test = is_config_test
+  end
+
+  def config_test?
+    !!@is_config_test
+  end
+
+  def format_config(config_path, config_string)
     config_string = config_string.to_s
     if config_path
       # Append the config string.
@@ -21,19 +30,19 @@ module LogStash; module Config; module Loader
     config_string
   end
 
-  def self.load_config(path)
+  def load_config(path)
     begin
       uri = URI.parse(path)
 
       case uri.scheme
-        when nil then
-          local_config(path)
-        when /http/ then
-          fetch_config(uri)
-        when "file" then
-          local_config(uri.path)
-        else
-          fail(I18n.t("logstash.runner.configuration.scheme-not-supported", :path => path))
+      when nil then
+        local_config(path)
+      when /http/ then
+        fetch_config(uri)
+      when "file" then
+        local_config(uri.path)
+      else
+        fail(I18n.t("logstash.runner.configuration.scheme-not-supported", :path => path))
       end
     rescue URI::InvalidURIError
       # fallback for windows.
@@ -43,9 +52,9 @@ module LogStash; module Config; module Loader
     end
   end
 
-  def self.local_config(path)
-    path = File.expand_path(path)
-    path = File.join(path, "*") if File.directory?(path)
+  def local_config(path)
+    path = ::File.expand_path(path)
+    path = ::File.join(path, "*") if ::File.directory?(path)
 
     if Dir.glob(path).length == 0
       fail(I18n.t("logstash.runner.configuration.file-not-found", :path => path))
@@ -54,13 +63,13 @@ module LogStash; module Config; module Loader
     config = ""
     encoding_issue_files = []
     Dir.glob(path).sort.each do |file|
-      next unless File.file?(file)
+      next unless ::File.file?(file)
       if file.match(/~$/)
         @logger.debug("NOT reading config file because it is a temp file", :config_file => file)
         next
       end
       @logger.debug("Reading config file", :config_file => file)
-      cfg = File.read(file)
+      cfg = ::File.read(file)
       if !cfg.ascii_only? && !cfg.valid_encoding?
         encoding_issue_files << file
       end
@@ -80,7 +89,7 @@ module LogStash; module Config; module Loader
     return config
   end # def load_config
 
-  def self.fetch_config(uri)
+  def fetch_config(uri)
     begin
       Net::HTTP.get(uri) + "\n"
     rescue Exception => e
--- a/logstash-core/lib/logstash/config/loader.rb
+++ b/logstash-core/lib/logstash/config/loader.rb
@@ -1,0 +1,90 @@
+require "logstash/config/defaults"
+
+module LogStash; module Config; module Loader
+  def self.format_config(config_path, config_string)
+    config_string = config_string.to_s
+    if config_path
+      # Append the config string.
+      # This allows users to provide both -f and -e flags. The combination
+      # is rare, but useful for debugging.
+      config_string = config_string + load_config(config_path)
+    else
+      # include a default stdin input if no inputs given
+      if config_string !~ /input *{/
+        config_string += LogStash::Config::Defaults.input
+      end
+      # include a default stdout output if no outputs given
+      if config_string !~ /output *{/
+        config_string += LogStash::Config::Defaults.output
+      end
+    end
+    config_string
+  end
+
+  def self.load_config(path)
+    begin
+      uri = URI.parse(path)
+
+      case uri.scheme
+        when nil then
+          local_config(path)
+        when /http/ then
+          fetch_config(uri)
+        when "file" then
+          local_config(uri.path)
+        else
+          fail(I18n.t("logstash.runner.configuration.scheme-not-supported", :path => path))
+      end
+    rescue URI::InvalidURIError
+      # fallback for windows.
+      # if the parsing of the file failed we assume we can reach it locally.
+      # some relative path on windows arent parsed correctly (.\logstash.conf)
+      local_config(path)
+    end
+  end
+
+  def self.local_config(path)
+    path = File.expand_path(path)
+    path = File.join(path, "*") if File.directory?(path)
+
+    if Dir.glob(path).length == 0
+      fail(I18n.t("logstash.runner.configuration.file-not-found", :path => path))
+    end
+
+    config = ""
+    encoding_issue_files = []
+    Dir.glob(path).sort.each do |file|
+      next unless File.file?(file)
+      if file.match(/~$/)
+        @logger.debug("NOT reading config file because it is a temp file", :config_file => file)
+        next
+      end
+      @logger.debug("Reading config file", :config_file => file)
+      cfg = File.read(file)
+      if !cfg.ascii_only? && !cfg.valid_encoding?
+        encoding_issue_files << file
+      end
+      config << cfg + "\n"
+      if config_test?
+        @logger.debug? && @logger.debug("\nThe following is the content of a file", :config_file => file.to_s)
+        @logger.debug? && @logger.debug("\n" + cfg + "\n\n")
+      end
+    end
+    if encoding_issue_files.any?
+      fail("The following config files contains non-ascii characters but are not UTF-8 encoded #{encoding_issue_files}")
+    end
+    if config_test?
+      @logger.debug? && @logger.debug("\nThe following is the merged configuration")
+      @logger.debug? && @logger.debug("\n" + config + "\n\n")
+    end
+    return config
+  end # def load_config
+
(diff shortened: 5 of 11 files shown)


logstash-core/lib/logstash/agent_plugin_manager.rb
```diff
@@ -0,0 +1,65 @@
+# encoding: utf-8
+require "logstash/agent"
+
+module LogStash module AgentPluginManager
```
purbon: Not the best one myself about naming, but I think I prefer if we name this class thing like ```AgentPluginRegistry``` what do you think?

logstash-core/lib/logstash/agent_plugin_manager.rb
```diff
@@ -0,0 +1,65 @@
...
+  DEFAULT_AGENT_NAME = :default
+  class DuplicatePluginNameError < Error; end
```
purbon: should this exception be ```DuplicatePluginError``` insted of ```DuplicatePluginNameError```?

logstash-core/lib/logstash/config/defaults.rb
```diff
@@ -6,6 +6,14 @@ module LogStash module Config module Defaults
 
   extend self
 
+  def input
+    "input { stdin { type => stdin } }"
+  end
+
+  def output
+    "output { stdout { codec => rubydebug } }"
```
purbon: If not mistaken, the current default is without the codec. why do you add here the rubydebug codec? I know it provide more clear information but wondering if we should just output plain text without codec. what do you think?

logstash-core/lib/logstash/config/defaults.rb
```diff
@@ -6,6 +6,14 @@ module LogStash module Config module Defaults
 
   extend self
 
+  def input
```
purbon: I think using constants here might be more clear, what do you think?

logstash-core/lib/logstash/config/loader.rb
```diff
@@ -0,0 +1,90 @@
...
+    config_string = config_string.to_s
+    if config_path
+      # Append the config string.
+      # This allows users to provide both -f and -e flags. The combination
+      # is rare, but useful for debugging.
+      config_string = config_string + load_config(config_path)
+    else
+      # include a default stdin input if no inputs given
+      if config_string !~ /input *{/
+        config_string += LogStash::Config::Defaults.input
```
purbon: My expectation would be see this line as something like  ```LogStash::Config::Defaults::INPUT``` does it makes sense to you? why using a method instead of a constant?

logstash-core/lib/logstash/config/loader.rb
```diff
@@ -0,0 +1,90 @@
+require "logstash/config/defaults"
+
+module LogStash; module Config; module Loader
```
purbon: I think the scope of this module is out of the scope of this PR. If we talk about plugable agent, I don't understand the purpose of this loader config class here. Can you explain it to me?

logstash-core/lib/logstash/runner.rb
```diff
@@ -102,18 +103,16 @@ def execute
...
+    @agent = create_agent
 
     if config_test?
-      puts "Configuration OK"
+      config_error = @agent.config_valid?
+      if config_error
+        @logger.fatal I18n.t("logstash.error", :error => config_error)
+        return 1
+      else
+        puts "Configuration OK"
```
purbon: we should not use puts here. Can we use cabin for this with the terminal output for example?

logstash-core/lib/logstash/runner.rb
```diff
@@ -122,10 +121,8 @@ def execute
   rescue LoadError => e
     fail("Configuration problem.")
   rescue LogStash::ConfigurationError => e
+    @logger.warn I18n.t("logstash.runner.configtest-flag-information")
     @logger.fatal I18n.t("logstash.error", :error => e)
-    if !config_test?
-      @logger.warn I18n.t("logstash.runner.configtest-flag-information")
-    end
     show_short_help
     return 1
```
purbon: why not using ```exit(1)``` as you did in another examples?

logstash-core/lib/logstash/runner.rb
```diff
@@ -102,18 +103,16 @@ def execute
...
-
-    @agent.add_pipeline("base", config_string, :filter_workers => filter_workers)
+    @agent = create_agent
 
     if config_test?
-      puts "Configuration OK"
+      config_error = @agent.config_valid?
+      if config_error
+        @logger.fatal I18n.t("logstash.error", :error => config_error)
+        return 1
```
purbon: why not using exit(1)?

spec/core/agent_plugin_manager_spec.rb
```diff
@@ -0,0 +1,36 @@
+# encoding: utf-8
+require "spec_helper"
+require "logstash/runner"
+require "stud/task"
+
+describe LogStash::AgentPluginManager do
+  class TestAgent < LogStash::Agent; end
+  class TestAgent2 < LogStash::Agent; end
+
+  subject { described_class }
```
purbon: this line is not strictly necessary. Here you are basically doing ``` a = a ``` :+1:

logstash-core/spec/logstash/runner_spec.rb
```diff
@@ -33,45 +44,28 @@ def run(args); end
...
-        let(:dummy_config) { 'input {}' }
-
-        before { expect(Net::HTTP).to receive(:get) { dummy_config } }
-        it 'works with http' do
-          expect(subject.load_config(path)).to eq("#{dummy_config}\n")
-        end
-      end
-    end
+  context "--agent" do
+    # class DummyAgent < LogStash::Agent; end
```
purbon: why is this code commented out?

purbon:
@andrewvc I did an initial review. I have an "interesting" concern. What about security? have you thought about implication of loading all the agents as part of being a core component of LS?. I think we should evaluate the security implications of it. What do you think?

spec/core/agent_plugin_manager_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      subject.register(:foo, TestAgent)
+      expect(subject.lookup(:foo)).to eql(TestAgent)
+    end
+
+    it "should not allow two plugins to be registered under the same name" do
+      subject.register(:foo, TestAgent)
+      expect { subject.register(:foo, TestAgent2) }.to raise_error(LogStash::AgentPluginManager::DuplicatePluginNameError)
+    end
+  end
+
```
purbon: I would consider adding test for find_plugins and load_all functions.

spec/core/agent_plugin_manager_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      expect(subject.available.size).to eql(1)
+    end
+
+    it "should be able to register an additional plugin" do
+      subject.register(:foo, TestAgent)
+      expect(subject.lookup(:foo)).to eql(TestAgent)
+    end
+
+    it "should not allow two plugins to be registered under the same name" do
+      subject.register(:foo, TestAgent)
```
purbon: this calls could be DRY out and might grouped into a context?

spec/core/agent_plugin_manager_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  subject { described_class }
+
+  after(:each) do
+    LogStash::AgentPluginManager.reset!
+  end
+
+  describe "the default registry with no registered plugins" do
+    it "should have the default agent" do
+      expect(subject.lookup(LogStash::AgentPluginManager::DEFAULT_AGENT_NAME)).to eql(LogStash::Agent)
```
purbon: in order to make the lines smaller, I would break this one and have the lookup in a separate variable or even a ```let``` statement.

spec/core/agent_plugin_manager_spec.rb
purbon: for the way you use subject, I would recommend actually using the sintax ``` subject(:foo) { Klass.new } ``` like this you don't need to use subject and can use :foo as a more contextualized var name.

logstash-core/lib/logstash/runner.rb
```diff
@@ -184,6 +186,20 @@ def configure_plugin_paths(paths)
     end
   end
 
+  def agent_class
+    return @agent_class if @agent_class # We don't want to load these things twice
+
+    LogStash::AgentPluginManager.load_all
```
purbon: I would probably break this function into more parts. Specially having this ```load_all``` as something done probably in construction time of this class. What do you think?

logstash-core/lib/logstash/agent_plugin_manager.rb
```diff
@@ -0,0 +1,65 @@
...
+  # class MyLogStashAgent < LogStash::Agent
+  #   LogStash::AgentPluginManager.register(:my_logstash_agent, self)
+  #
+  #   # ...
+  # end
+  def self.register(name, plugin_class)
+    name_sym = name.to_sym
+
+    if (conflicting_class = registry[name_sym])
+      raise DuplicatePluginNameError, "Could not register plugin '#{plugin_class}' as '#{name}', this name is already taken by '#{conflicting_class}'"
```
purbon: This line is extra long, can we break it a bit? just to make reading the code easier :+1:

logstash-core/lib/logstash/agent_plugin_manager.rb
```diff
@@ -0,0 +1,65 @@
...
+  #
+  # class MyLogStashAgent < LogStash::Agent
+  #   LogStash::AgentPluginManager.register(:my_logstash_agent, self)
+  #
+  #   # ...
+  # end
+  def self.register(name, plugin_class)
+    name_sym = name.to_sym
+
+    if (conflicting_class = registry[name_sym])
```
purbon: I'm not a fan at all of this kind of syntax, always found it confusing like hell. I would recommend having a clear separation between the boolean condition and the retrieval of the class.

logstash-core/lib/logstash/agent_plugin_manager.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  REGISTRY = {}
+
+  # Reset plugin registry to just the default plugin
+  def self.reset!
+    REGISTRY.clear
+    REGISTRY[DEFAULT_AGENT_NAME] = LogStash::Agent
+  end
+
+  reset!
```
purbon: what is this function call business? why is not enclosed in a method?

logstash-core/lib/logstash/agent_plugin_manager.rb
andrewvc: That's probably more accurate, I'll change the name.

logstash-core/lib/logstash/agent_plugin_manager.rb
andrewvc: Yes!

logstash-core/lib/logstash/agent_plugin_manager.rb
andrewvc: That needs a comment explaining that the purpose is to set the default agent when the class is loaded. That DRYs up the code or I'd have to put it in the `REGISTRY` declaration above.

logstash-core/lib/logstash/agent_plugin_manager.rb
andrewvc: Well, the convention in the ruby style guide is that this is OK so long as its enclosed with parents `()`. Did we ever finish up our style guide?

logstash-core/lib/logstash/agent_plugin_manager.rb
andrewvc: sure thing :)

logstash-core/lib/logstash/config/defaults.rb
andrewvc: Well, I want to follow the convention established by the cpu worker strategy. I think its cleaner to have only methods than to pick and choose based on how dynamic stuff is.

logstash-core/lib/logstash/config/defaults.rb
andrewvc: Hmmm, I got that from here: https://github.com/elastic/logstash/blob/00a99c19e512a6ec0089cf5996168cbc15470330/logstash-core/lib/logstash/runner.rb#L18 

Is that incorrect?

purbon: I was confused, please ignore me :+1:  https://github.com/elastic/logstash/blob/1.5/lib/logstash/agent.rb#L104-L109

logstash-core/lib/logstash/config/defaults.rb
purbon: I understand, still thinks constants are better suited for this kind of data. The difference for me is this are string while the other is coming from a method call. I think we should move to constant as it was previously. But I agree is good to have this methods in here.

logstash-core/lib/logstash/agent_plugin_manager.rb
purbon: nope, so this is tied not to "laws" but to personal feelings. I speak by myself, this kind of constructors are confusing and error prone to me. I would recommend rewriting, but your call at the end of the day.

logstash-core/lib/logstash/agent_plugin_manager.rb
purbon: I strongly think we should not use this, all calls should go into methods/classes. We can not have this kind of uncontrolled changes to status. Please avoid using it.

andrewvc: @purbon well, where would you put this logic? We need to set this somewhere, and this is a default property of this module.

The only alternative I can think of would be rewriting the module as a class and putting it in the initializer. But that seems like excessive ceremony to work around something that could be implemented simply.

I stand by my assertion that an explanatory comment is all that's needed here. Maybe @jordansissel can chime in with a tie breaking opinion here.

logstash-core/lib/logstash/config/defaults.rb
andrewvc: Well, the main reason for constants is performance / memory allocation, and this will never be called more than a couple times per invocation so I don't think the reasoning behind using constants in general applies here.

logstash-core/lib/logstash/config/loader.rb
andrewvc: This moves code that was in the runner into this new separate class.

This is important because new agent plugins may need to invoke this functionality themselves. For that reason the code could no longer go in either the runner or the Agent. Plus it has the happy side effect of make runner address only a single concern.

logstash-core/lib/logstash/config/loader.rb
andrewvc: Addressed in the comment above.

logstash-core/lib/logstash/runner.rb
andrewvc: Good catch, I meant to call exit!

logstash-core/lib/logstash/runner.rb
andrewvc: Hmmm, well, this was in the original code. I'm not sure refactoring that is in scope here.

logstash-core/lib/logstash/agent_plugin_manager.rb
purbon: Lets discuss about it, for sure. But we should definitely not do this kind of things as then errors are very complicated to find out. So yeah, look for another person to remove the conflict, but -1 from my side to this kind of constructs.

logstash-core/spec/logstash/runner_spec.rb
andrewvc: Good call, fixed and pushed!

logstash-core/lib/logstash/runner.rb
andrewvc: Actually, this is not a bug, and is how the rest of this works. I believe that this return code will be used by clamp during exit. Cleverly, this lets us run unit tests without exiting the process. So this should stay :)

logstash-core/lib/logstash/agent_plugin_manager.rb
colinsurprenant: @purbon could you please explain what kind of complicated to find out errors can result from this? I am not sure I understand your concern here. 

This seems to me a fine singleton pattern implementation, `reset!` is only called at module load time. something I am missing?

logstash-core/lib/logstash/runner.rb
suyograo: +1 on using Cabin here instead of `puts` since you are in this code area

logstash-core/lib/logstash/runner.rb
purbon: interesting, looks definitely strange to me :+1: ...

logstash-core/spec/logstash/runner_spec.rb
```diff
@@ -18,14 +18,25 @@ def run(args); end
 
   describe "argument parsing" do
 
+    before do
```
purbon: If the code within is commented out, this block should be removed.

logstash-core/spec/logstash/runner_spec.rb
```diff
@@ -33,44 +44,29 @@ def run(args); end
...
       end
     end
   end
 
-  context "when loading the configuration" do
+  context "--agent" do
+    class DummyAgent < LogStash::Agent; end
+
+    let(:agent_name) { "testagent" }
+    let(:dummy_agent_class) { DummyAgentClass }
```
purbon: I know this is picky, but wording this as dummy it always felt to me strange. Why not using mock or stub? what do you think?

logstash-core/spec/logstash/runner_spec.rb
```diff
@@ -18,14 +18,25 @@ def run(args); end
...
+    end
+
     subject { LogStash::Runner.new("") }
     context "when -e is given" do
 
       let(:args) { ["-e", ""] }
+      let(:agent) { double("agent") }
+      let(:agent_logger) { double("agent logger") }
+
+      before do
```
purbon: I see this pattern shared in most of the test here, I would recommend having explicit before(:each), before(:all), etc statements. I found this syntax more clear to the reader.

logstash-core/lib/logstash/config/defaults.rb
purbon: Well nothing against that, as we have no clear guidelines on that I can not really argument :smile_cat:, It does the work, so no problem on it.

logstash-core/lib/logstash/agent_plugin_registry.rb
```diff
@@ -0,0 +1,66 @@
+# encoding: utf-8
+require "logstash/agent"
+
+module LogStash module AgentPluginRegistry
```
purbon: As we spoke, I wonder why the registry should be a module. It have some state, as seen in the registry lookup. Can you walk me through you thinking?

logstash-core/lib/logstash/agent_plugin_registry.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  REGISTRY = {}
+
+  # Reset plugin registry to just the default plugin
+  def self.reset!
+    REGISTRY.clear
+    REGISTRY[DEFAULT_AGENT_NAME] = LogStash::Agent
+  end
+
+  reset!
```
purbon: @colinsurprenant not able to find your comment now, but will try to explain my reasoning. I find this usage of singleton implementation here strange, the code is basically within a module, but keeping an state that need an initialize method to be run like actually @andrewvc did here. 

My problem goes with having a pattern of knowing an encapsulating there the change of state happen. If we want to follow proper object orientation this should only happen in methods.

I think this should be changed to be a class and not a module. Unless I'm obviously missing why this should be a module.

logstash-core/lib/logstash/runner.rb
andrewvc: Clamp will return the proper exit code. This is really useful! The reason being that if I called exit directly this module wouldn't really be testable.

logstash-core/spec/logstash/runner_spec.rb
andrewvc: Good catch!

logstash-core/spec/logstash/runner_spec.rb
andrewvc: Errr, `before` == `before(:each)`. Shouldn't anyone familiar with rspec know this?

logstash-core/spec/logstash/runner_spec.rb
andrewvc: Hmmm, this line wasn't even being used! So that's an easy answer here :)

andrewvc:
@ph @purbon do I have your LGTM on this? I just pushed up all the changes you've mentioned

logstash-core/lib/logstash/agent_plugin_registry.rb
colinsurprenant: having a singleton pattern implemented in a module is IMO perfectly acceptable, there is no problem in keeping state, it could even be with using an instance variable in the module. it is a correct pattern.

personally I like using `extend self` instead of defining methods with `self.xxx` but both are acceptable.

logstash-core/lib/logstash/runner.rb
colinsurprenant: confirmed for return + clamp and way better for specs.

logstash-core/lib/logstash/runner.rb
```diff
@@ -184,6 +186,22 @@ def configure_plugin_paths(paths)
...
+    return @agent_class if @agent_class # We don't want to load these things twice
+
+    LogStash::AgentPluginRegistry.load_all
+    @agent_class = LogStash::AgentPluginRegistry.lookup(agent_name)
+
+    if !@agent_class
+      @logger.fatal("Could not load specified agent",
+                    :agent_name => agent_name,
+                    :valid_agent_names => LogStash::AgentPluginRegistry.available.map(&:to_s))
+      exit(1)
```
colinsurprenant: should'nt we push down the exit code to clamp and let it do the exit?

logstash-core/spec/logstash/runner_spec.rb
colinsurprenant: I'm always confused with the `before` semantic, I'd vote to use the explicit version...

andrewvc: I think maybe we should punt on this and finish our style guide. A code search shows 8 usages of `before` and 9 usages of `before(:each)` currently.

ph:
LGTM

logstash-core/lib/logstash/agent.rb
```diff
@@ -3,20 +3,45 @@
...
 
 LogStash::Environment.load_locale!
 
 class LogStash::Agent
+  attr_reader :logger, :runner
 
-  attr_writer :logger
-
-  def initialize
+  def initialize(logger, runner)
```
jsvd: I'd expect the agent to be given commands by the runner and not receive it as an argument.

having the agent be smart and interact with the runner makes it less flexible, the "add_pipelines" api has all this hidden meaning of getting the config from the runner.

regardless of the agent implementation, as long as it has the add_pipeline and execute api, if the cli args are `-e ""`,  then the runner should at least add a new pipeline and start the agent

logstash-core/lib/logstash/runner.rb
```diff
@@ -184,6 +190,22 @@ def configure_plugin_paths(paths)
     end
   end
 
+  def agent_class
+    return @agent_class if @agent_class # We don't want to load these things twice
+
+    LogStash::AgentPluginRegistry.load_all
```
jsvd: since we're loading all anyway I don't see why it needs to be hidden under the agent_class getter, I'd suggest making this part of the runner initialize or something similar..this way we can be clearer during the runner#execute:

```ruby
@agent = LogStash::AgentPluginRegistry.lookup(agent_name).new
```

purbon: +1 on something like this. I don't really see he need doing a require for all.

logstash-core/lib/logstash/agent.rb
andrewvc: That's a great point. I'll refactor this to decouple the runner completely.

logstash-core/lib/logstash/runner.rb
andrewvc: Good point :)

andrewvc:
@jsvd @purbon thanks for your latest feedback! I've incorporated those changes and now think its pretty good! Can I get LGTMs from you guys if possible?

logstash-core/lib/logstash/config/loader.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+    end
+    config_string
+  end
+
+  def self.load_config(path)
+    begin
+      uri = URI.parse(path)
+
+      case uri.scheme
```
jsvd: please realign this case statement

logstash-core/lib/logstash/config/loader.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+      # include a default stdout output if no outputs given
+      if config_string !~ /output *{/
+        config_string += LogStash::Config::Defaults.output
+      end
+    end
+    config_string
+  end
+
+  def self.load_config(path)
```
jsvd: I know this was the previous name but can you change it to something more accurate like "fetch_config" and change fetch_config to...`from_url` or something?

Specially now that the loader is an isolated entity, It's strange that `LogStash::Config::Loader.load_config` is a stateless operation that retrieves a config from a path

logstash-core/lib/logstash/runner.rb
```diff
@@ -102,30 +106,35 @@ def execute
...
+        @logger.terminal "Configuration OK"
+      end
     else
+      @agent.add_pipeline("base", config_str, :filter_workers => filter_workers)
       task = Stud::Task.new { @agent.execute }
       return task.wait
     end
 
+  rescue MissingAgentError => e
+    return 1
```
jsvd: does this mean that the exception is swallowed ?
also, this exception can only happen during `@agent = create_agent` so I'd suggest containing the begin rescue to that

logstash-core/lib/logstash/runner.rb
```diff
@@ -184,6 +198,21 @@ def configure_plugin_paths(paths)
     end
   end
 
+  def agent_class
+    return @agent_class if @agent_class # We don't want to load these things twice
```
jsvd: a lookup should be rather harmless so I suggest:

* eliminating this method
* make create agent:

```ruby
def create_agent
  @logger.info("Loading agent", :class => agent_class)
  LogStash::AgentPluginRegistry.lookup(agent_name).new(@logger)
end
```

* make the logic of raising a `MissingAgentError` exception part of the lookup method
* put the @logger.fatal message in the rescue of `@agent = create_agent`

andrewvc: I'd personally rather keep the exception out of `LogStash::AgentPluginRegistry.lookup`. I think a nil value better fits a lot of usage patterns.

andrewvc: I do agree with eliminating the method and I would like to make the exception part of the `create_agent` method.

logstash-core/lib/logstash/runner.rb
andrewvc: No, because before the exception is raised a `@logger.fatal` is called. It is confusing however. I'll try to rework the flow.

andrewvc:
@jsvd I believe my latest pushes have addressed all your concerns. How's it look now?

purbon:
LGTM, shipt it! when possible, for sure!

logstash-core/lib/logstash/runner.rb
```diff
@@ -10,12 +10,12 @@
...
 require "logstash/namespace"
+require "logstash/agent_plugin_registry"
 require "logstash/agent"
+require "logstash/config/defaults"
 
 class LogStash::Runner < Clamp::Command
-
-  DEFAULT_INPUT = "input { stdin { type => stdin } }"
-  DEFAULT_OUTPUT = "output { stdout { codec => rubydebug } }"
+  class MissingAgentError < StandardError; end # Raised when the user asks for an agent plugin that doesn't exist
```
jsvd: this doesn't seem to be used anymore

ph:
@andrewvc can you rebase this?

jsvd:
I vote for not merging this PR and consider respinning in order to have a generic plugin loading mechanism.

If we transform the AgentPluginRegistry into just a PluginRegistry that can handle multiple types of plugins (agents, filters, codecs, pipelines, logstash-events), we can re-use it in multiple places.

In its essence the PluginRegistry is quite simple and we can either depend on gemspecs and their metadata or some other sort of annotation to interact with the Registry. After that, a simple api of `LogStash::PluginRegistry.lookup(type, name)` should yield the desired class.

andrewvc:
+1 on the respin for the generic thing.

suyograo:
Closing this for a re-implementation here: https://github.com/elastic/logstash/pull/4535

PR #4197, opened 2015-11-16
updated yum repo doku - load gpg via https

lhampe:
Updated the docs:
ensure the YUM GPG-Key is imported via **https** to preserve users getting into security problems.

### DIFF
--- a/docs/asciidoc/static/repositories.asciidoc
+++ b/docs/asciidoc/static/repositories.asciidoc
@@ -74,7 +74,7 @@ in a file with a `.repo` suffix, for example `logstash.repo`
 name=Logstash repository for {branch}.x packages
 baseurl=http://packages.elastic.co/logstash/{branch}/centos
 gpgcheck=1
-gpgkey=http://packages.elastic.co/GPG-KEY-elasticsearch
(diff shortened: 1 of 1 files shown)


jordansissel:
I haven't reviewed the change, but the language in the description of this PR is highly inappropriate. I'm deleting the description - please update it with something more accurate?

jordansissel:
Change LGTM, but I haven't tested it. @electrical - thoughts?

jordansissel:
@lhampe also I see that we have no CLA on file for you; in order to proceed, can you follow up on the [contributing guide](https://github.com/elastic/logstash/blob/master/CONTRIBUTING.md#contribution-steps) and look at step 2?

lhampe:
@jordansissel my description should just be a little Joke ;)
... updated it to a politically correct declaration.

For sure you're allowed to distribute this single character.
I don't have time to read and accept this licenses now. Is this really necessary for this single letter?

jordansissel:
@lhampe copyright and intellectual propertly laws are tricky things. I am not a lawyer, so I probably don't have a satisfying answer for you, but we generally require CLA for all source contributions with the goal to further protect all the users of the project.

PR #4199, opened 2015-11-16
Add CLI documentation for #3243

suyograo:
We added new --debug option in #3243, adding it to CLI flags


untergeek:
Shiny! 

LGTM™

PR #4206, opened 2015-11-17
[2.1] Bump version to snapshot1

suyograo:


purbon:
Added this morning #4210 to fix the build, is related to that and I guess yours will fix also the same thing.  let me know what do you think. 

we should do a similar thing for 2.0 branch.

purbon:
LGTM, merging this.

PR #4207, opened 2015-11-17
fix version path, relates to #4123

colinsurprenant:
this fixed the build artifacts bug introduced by #4123


suyograo:
LGTM, merge away

PR #4210, opened 2015-11-17
Update 2.1 lock file with latest changes

purbon:
specially  in plugins and dependencies, fixing the concurrent ruby issue.

Fix http://build-eu-00.elastic.co/view/LS%202.1/job/logstash_regression_21/jdk=JDK7,label=metal-pool/3/console by updating the lock file.

```
Successfully installed bundler-1.9.10
Invoking bundler install...
Plugin version conflict, aborting
rake aborted!
Bundler could not find compatible versions for gem "concurrent-ruby":
  In snapshot (Gemfile.lock):
    concurrent-ruby (= 0.9.1)

  In Gemfile:
    logstash-core (>= 0) java depends on
      concurrent-ruby (= 0.9.2) java

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
``` 

Similar fix should be applied to 2.0 branch, this will come in another pr.

https://github.com/purbon/logstash/commit/3a25f5a99827dd8fed8ceb69f73bbcdadff6301e was added only as a compliment, so the test pass,  but better check #4212 for a general version of it that is the one that will be merged.

relates to #4212 && #4206 , where  3a25f5a99827dd8fed8ceb69f73bbcdadff6301e gets also introduced.

### DIFF
--- a/Gemfile.jruby-1.9.lock
+++ b/Gemfile.jruby-1.9.lock
@@ -8,38 +8,39 @@ GEM
     avl_tree (1.2.1)
       atomic (~> 1.1)
     awesome_print (1.6.1)
-    aws-sdk (2.1.30)
-      aws-sdk-resources (= 2.1.30)
-    aws-sdk-core (2.1.30)
+    aws-sdk (2.1.36)
+      aws-sdk-resources (= 2.1.36)
+    aws-sdk-core (2.1.36)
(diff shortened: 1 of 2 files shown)


purbon:
closing this in favor of #4206

PR #4211, opened 2015-11-17
Fix init script for Redhat.

chrw:
I had some problem using the legacy init script with RHEL 5.10. I fixed some glitches which worked for me.


chrw:
I signed the CLA. :)

PR #4212, opened 2015-11-17
Fix build in version branches by stub the warn_local_gems call in the update command 

purbon:
This is necessary, not for master, but for the version branches where we actually use no local gems, so this call will never be called.

Fixes the broken build in jenkins errors after #4210 is merged.


guyboertje:
LGTM

purbon:
funny thing, is after merging #4206, the build is green again, interesting -_-! something I see ghosts walking around

PR #4215, opened 2015-11-17
Fixes 2.0 build

purbon:
See http://build-eu-00.elastic.co/view/LS%202.0/job/logstash_regression_20/109/jdk=JDK8,label=metal-pool/console

This change fixed the concurrent-ruby version in the lock file, so test can run without conflicts.

Related to #4210


guyboertje:
LGTM

PR #4222, opened 2015-11-17
Fix a string interpolation memory leak

ph:
When the event was serialized to string using a `to_s` call, mostly
happenning when doing logging. Each calls would generate a new template
each template was unique because it was containing the date.

The accumulation of templates was making logstash goes OOM.

We aren't cleaning the templates because they should stabilize to a
finite number.

### DIFF
--- a/logstash-core-event/lib/logstash/event.rb
+++ b/logstash-core-event/lib/logstash/event.rb
@@ -106,7 +106,7 @@ class LogStash::Event
 
   public
   def to_s
-    self.sprintf("#{timestamp.to_iso8601} %{host} %{message}")
+    "#{timestamp.to_iso8601} #{self.sprintf("%{host} %{message}")}"
   end # def to_s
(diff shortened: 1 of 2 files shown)


logstash-core-event/lib/logstash/event.rb
```diff
@@ -106,7 +106,7 @@ def clone
 
   public
   def to_s
-    self.sprintf("#{timestamp.to_iso8601} %{host} %{message}")
+    "#{timestamp.to_iso8601} #{self.sprintf("%{host} %{message}")}"
```
suyograo: wow

ph: We were not hit bad that before since we did not cache the template :(
The bad thing this is probably a logging statement somewhere.

colinsurprenant:
LGTM - nice catch!

PR #4226, opened 2015-11-18
Added the code to set output to json format with stdout {codec=>rub…

zeevgl:
the example in this page shows the output in json format but does not explain how to set the stdout to show json.

output {
 stdout {codec=>rubydebug} 
}

### DIFF
--- a/docs/asciidoc/static/advanced-pipeline.asciidoc
+++ b/docs/asciidoc/static/advanced-pipeline.asciidoc
@@ -116,6 +116,14 @@ filter {
 }
 --------------------------------------------------------------------------------
 
+To set the stdout output to json format:
+[source,json]
+--------------------------------------------------------------------------------
+output {
+    stdout {codec=>rubydebug} 
+}
+--------------------------------------------------------------------------------
+
 After processing, the sample line has the following JSON representation:
 
 [source,json]


zeevgl:
OK, I have signed the CLA and became a member of Elastic

zeevgl:
jenkins test it

docs/asciidoc/static/advanced-pipeline.asciidoc
```diff
@@ -116,6 +116,14 @@ filter {
 }
 --------------------------------------------------------------------------------
 
+To set the stdout output to json format:
```
magnusbaeck: Can we be consistent with the following paragraph and say "JSON" instead of "json"?

docs/asciidoc/static/advanced-pipeline.asciidoc
```diff
@@ -116,6 +116,14 @@ filter {
...
+[source,json]
+--------------------------------------------------------------------------------
+output {
+    stdout {codec=>rubydebug}
```
magnusbaeck: I think you actually mean:

    stdout {
        codec => "json"
    }

zeevgl: nope, the one that worked for me was "rubydebug"
this is what is used in the "Getting Started with Logstash Overview Video
Watch Now"

maybe json works to but I tried rubydebug

magnusbaeck: Well, both work but the preceding paragraph says "To set the stdout output to json format" and then you can't set the output codec to rubydebug because that's not JSON.

zeevgl: OK, we can use the word json. my goal was to explain how to get the result that is shown in the example

jordansissel: I think the issue is that rubydebug is not json, so we should probably not use the word `json`. If we really want to show json, we can use `codec => json_lines` for stdout (json codec doesn't have line terminators).

PR #4229, opened 2015-11-18
clean cache to empty previous entries

colinsurprenant:
this fixed intermittent core specs failures depending if `Event#to_s` was called before the `Event` specs.


ph:
LGTM

suyograo:
LGTM

PR #4247, opened 2015-11-20
better rspec defaults to use documentation formatter in rake tasks and CI

colinsurprenant:
- default to `--format=documentation` for all rspec tasks and when running CI tests

Some advantages of using `--format=documentation` 
- better/more verbose jenkins reports
- visual cue to exactly where a test stalls when it does

### DIFF
--- a/ci/ci_test.sh
+++ b/ci/ci_test.sh
@@ -7,6 +7,8 @@
 
 SELECTED_TEST_SUITE=$1
 
+export RSPEC_OPTIONS="--order rand --format documentation --format CI::Reporter::RSpecFormatter"
+
 if [[ $SELECTED_TEST_SUITE == $"core-fail-fast" ]]; then
   echo "Running core-fail-fast tests"
   rake test:install-core    # Install core dependencies for testing.
--- a/ci/ci_test.sh
+++ b/ci/ci_test.sh
@@ -7,7 +7,7 @@
 
 SELECTED_TEST_SUITE=$1
 
-export RSPEC_OPTIONS="--order rand --format documentation --format CI::Reporter::RSpecFormatter"
+export SPEC_OPTS="--order rand --format documentation --format CI::Reporter::RSpecFormatter"
 
 if [[ $SELECTED_TEST_SUITE == $"core-fail-fast" ]]; then
   echo "Running core-fail-fast tests"
--- a/ci/ci_setup.sh
+++ b/ci/ci_setup.sh
@@ -17,7 +17,4 @@ rm -rf spec/reports # no stale spec reports from previous executions
 rake bootstrap # Bootstrap your logstash instance
 
 
-# Set up some general options for the rspec runner
-#echo "--order rand" > .rspec
-#echo "--format progress" >> .rspec
-#echo "--format CI::Reporter::RSpecFormatter" >> .rspec
+# RSpec options are set in ci_test.sh
--- a/ci/ci_setup.sh
+++ b/ci/ci_setup.sh
@@ -16,7 +16,10 @@ rm -rf spec/reports # no stale spec reports from previous executions
 # Setup the environment
 rake bootstrap # Bootstrap your logstash instance
 
+
 # Set up some general options for the rspec runner
-echo "--order rand" > .rspec
-echo "--format progress" >> .rspec
-echo "--format CI::Reporter::RSpecFormatter" >> .rspec
+#echo "--order rand" > .rspec
+#echo "--format progress" >> .rspec
+#echo "--format CI::Reporter::RSpecFormatter" >> .rspec
+
+export RSPEC_OPTIONS="--order rand --format documentation --format CI::Reporter::RSpecFormatter"
--- a/ci/ci_setup.sh
+++ b/ci/ci_setup.sh
@@ -21,5 +21,3 @@ rake bootstrap # Bootstrap your logstash instance
 #echo "--order rand" > .rspec
 #echo "--format progress" >> .rspec
 #echo "--format CI::Reporter::RSpecFormatter" >> .rspec
-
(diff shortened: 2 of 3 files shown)


jordansissel:
+1 on the idea, code looks fine to me.

jordansissel:
Oh also, keep in mind that with Flores's randomized stress testing, it can make lots of specs like this:

```
  randomized testing
    should be near the boundary of requested size
    should be near the boundary of requested size
    should be near the boundary of requested size
    should be near the boundary of requested size
    should be near the boundary of requested size
...
```

It's not a problem today, but could be in the future. We can improve the Flores RSpec formatter and use that later if desired!

Jenkins output _looks_ good to me (docs formatted spec output). It failed for .. some other reason.

jordansissel:
```
Finished in 8.09 seconds (files took 1.81 seconds to load)
752 examples, 0 failures

Randomized with seed 63599

Recording test results
ERROR: Publisher hudson.tasks.junit.JUnitResultArchiver aborted due to exception
hudson.AbortException: No test report files were found. Configuration error?
	at hudson.tasks.junit.JUnitParser$ParseResultCallable.invoke(JUnitParser.java:116)

```

No test report output now?

colinsurprenant:
no idea what this failure is about :confused:

jordansissel:
jenkins, test this

colinsurprenant:
maybe @purbon or @mrsolo can help with that?

mrsolo:
Jenkins junit post processor expects to find junit.xml files with this file mask '**/spec/reports/*.xml'.

The failure basically means that processor has failed to find any junit result.

colinsurprenant:
weird. this new default option in this PR probably overwrites the format required for CI setup in https://github.com/elastic/logstash/blob/82f2868a9589e3ac2f39e1216c6f383295a760a0/ci/ci_setup.sh#L21-L22

instead of passing rspec options through the creation of a `.rspec` local file per ^^ we could use the new `RSPEC_OPTIONS` env var exposed in this PR? @ph @purbon  thoughts? I think it would be cleaner...

colinsurprenant:
I will add a commit to modify the CI scripts to see if that works. If it does we can decide if these commits should be made into another PR.

ph:
@colinsurprenant I believe this PR is not needed to achieve what you want.
`SPEC_OPTS="--format documentation" rake test:core` will work with the current code base.

If you want this behavior by default in our CI you can just add the SPEC_OPTS in the CI scripts.

ph:
Not sure why we/I created a `.rspec` file in the ci_setup

colinsurprenant:
ok, it is probably because the `ci_setup.sh` and `ci_test.sh` are called from different shells. 
I guess we could simply pass options using and env var directly in `ci_test.sh` ... not sure if there is a specific reason that it was done in the setup script. will try.

colinsurprenant:
@ph no, it's using my PR's env var. We can decide after if it is really needed

colinsurprenant:
ok, now that works. no need for the `.rspec` file for options passing. 

just tried with using `SPEC_OPTS` and it correctly overrides the rake tasks default options so I'll remove the `RSPEC_OPTIONS` env var handling. thanks for the headsup @ph

colinsurprenant:
ok, so that works. 

I updated the PR description for the changes last pushed. If we agree on this, I will squash and merge.

purbon:
@colinsurprenant about ```.rspec``` file I created/used this because I wanted to avoid passing env vars between different scripts/shell commands. But the option to setup env variables is good too.

ci/ci_test.sh
```diff
@@ -7,6 +7,8 @@
 
 SELECTED_TEST_SUITE=$1
 
+export SPEC_OPTS="--order rand --format documentation --format CI::Reporter::RSpecFormatter"
```
purbon: I see how documentation is good, but I wonder if makes sense to have it here? not sure if I will provide too much output for each run of jenkins job. what do you think?

purbon: I don't understand, documentation is the default in ```test.rake``` but is also set here? is this done in any purpose I don't see?

purbon:
In terms of formatters, what about using ```documentation``` when run locally (default) and ```progress``` in CI? I think this would make the best out of each system.

ci/ci_test.sh
colinsurprenant: documentation format: its the only way to know where a test "stalls" if it stalls. with dots you only see dots and there's no way to know which test has stalled.

documentation option: my thinking is that it makes it more explicit and not rely on any default value.

colinsurprenant:
@purbon it makes more sense to move the rspec options setting closer to where the specs are run and this is what ENV vars are for. Also passing arguments using and "defaults settings" files is not really good, you actually create a "permanent state" which in this case happens to be wiped at every run but it might not always be the case. It makes these script not reusable in an environment where the FS/image/container is not wiped at every run.

colinsurprenant:
@purbon for the formatter, as I said above in the code comment, documentation is the only way to know where a test "stalls" if it stalls. with dots you only see dots and there's no way to know which test has stalled.

in general is there a problem in having more information in the test report?

purbon:
@colinsurprenant yeah, no problem at all on that  ... I was only wondering if made sense or not. I agree to detect stalls documentation is more clear as progress is not good at all on this. The only situation where I would really recommend progress is when the number of test is bigger, see the plugins tests. There I think it makes no sense as the information generated will be just too much.

purbon:
In general the change here looks good to me, also core test pass on my side. Only concern is on the default format for plugins test, there I would really not use documentation as default, I'm worried about information overload.

purbon:
As a recap of this PR, I think adding --format documentation makes sense, and I like the changes. I would however, not add it as a default for the ```test:plugins``` task as there the generated report might be to noisy.

ph:
lgtm

PR #4254, opened 2015-11-23
Next Gen Pipeline

andrewvc:
This takes over from #4155 which was pointed toward the next_gen_pipeline branch. This new PR is pointed at master.

NOTE: I have not included configurable values for batch size or batch delay as I'm waiting for https://github.com/elastic/logstash/pull/3872 to merge first which will certainly cause a merge conflict! 

ATTN: This is no longer a POC but production ready code!

This is PR that aims to accomplish the following objectives.

1. Provide an in-memory implementation (no persistence), of the pipeline I described in https://github.com/elastic/logstash/issues/3693
2. Be completely backward compatible with current logstash plugins, nothing should break
3. Allow for complete pipeline snapshotting (even with the Elasticsearch plugin) to address concerns raised by @jsvd in https://github.com/elastic/logstash/pull/4051. This would allow forced shutdowns without losing data. This would require the elasticsearch output to implement `receive_multi` and no longer use its own internal queue
4. Expose a new 'batch' interface for filters and output plugins, allowing them to process events in microbatches for efficiency. This could have very significant potential for speed improvements for filters like the useragent filter, which would only have to lookup the number of unique keys per batch vs. each key individually. This would also obviate the need for a separate  buffer in the elasticsearch output.

I would say that, along with these changes we should contemplate adding a new plugin config option which would by default set the number of workers for the plugin to be == the number of worker threads. That will give the best user experience for this.

**Notes on implementation**

This is 90% the system described in https://github.com/elastic/logstash/issues/3693 , the one exception to this is that output plugins are assumed to be not threadsafe. This was necessary to make it backward compatible with 2.x series plugins. When the # of output workers is equal to the number of worker threads specified with `-w` the behavior should be very similar to what the original behavior was. I think the default worker fix described above might actually be a long term solution here.

**Persistence Examples**
Since this is just a POC, the pipeline shutdown persistence is demonstrated in the writing of two files on shutdown:

**Update: This functionality remains in the `dump_inflight` function, but is currently never called. We may enable this in a future release**
1. **/tmp/ls_current_batches:** The contents of in-memory buffers BEFORE the shutdown sequence is initiated. A 'fast' shutdown (as described in https://github.com/elastic/logstash/pull/4051 could simple write this then `exit` safely).
1. **/tmp/ls_current_batches_post_close:** This is just there for debugging in the POC, this is written after a clean shutdown. It will always be empty (unless there's a bug in the pipeline!).

** Notes For Reviewers **

1. This PR involves a small number of changes to the AST. To enable batch oriented processing for each batch we:
  1. Execute the filters one by one accumulating a new batch of filtered events.
  1. Execute `output_func` once per each function which no longer invokes the output func but rather returns an array of outputs that the event should be sent to.
  1. We use the info from the step above to build separate batches per output as seen in `output_batch` (in the pipeline)
1. This PR adds 2 new CLI options `batch_size` and `batch_poll_wait` to control the new features of the pipeline.
1. PLEASE verify my performance results for yourself. I'm currently using [this config](https://gist.github.com/andrewvc/b25097d19aa50df8e05f) and running it with `time bin/logstash -f that_config.conf`. Then comparing it with Logstash 2.0
1. I set the default batch size to be 125 to match the Elasticsearch output defaults assuming four workers (4*125 = 500). This seemed like a safe default.

**Performance**

I used the following config to test this thing out:
`bin/logstash -w 4 -e "input { generator {} } filter {  mutate { add_field => {'foo' => 'bar'} } } output { http { http_method => 'post' url => 'http://localhost:4567/hi' workers => 2 } }"`

Note that this requires a little web service to hit. I used the following

```ruby
require 'sinatra'

post '/hi' do
  puts "Got #{request.body.read}"
  [200, "Thanks!"]
end
```

**Performance**

Also worth noting, this new pipeline seems, in at least one micro-benchmark, to be as much as 63% faster!

```
~/p/logstash (pipeline_poc) $ bin/logstash -w 4 -e "input { generator {} } filter {  mutate { add_field => {'foo' => 'bar'} } } output { stdout {} }" | pv -l > /dev/null
^C61M 0:01:00 [67.1k/s] [                                                                                                                                                    <=>                         ]
~/p/logstash (pipeline_poc) $ cd ~/Downloads/logstash-2.0.0
~/D/logstash-2.0.0 $ bin/logstash -w 4 -e "input { generator {} } filter {  mutate { add_field => {'foo' => 'bar'} } } output { stdout {} }" | pv -l > /dev/null
^C24M 0:01:00 [  41k/s] [
```

Doing further benchmarking I was able to get the throughput up to 74k/s using a batch size of 40 (the default is 20). Upping the batch size to 200 made it go down to 71k/s, and upping it to 2000 brought it down a lot, to 50k/s. So, batch size has a big impact! Probably due to not being able to use memory/the cache as efficiently. We'll have to do further tests with different messages sizes as well.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -33,7 +33,7 @@ class LogStash::Agent
   end # def execute
 
   def add_pipeline(pipeline_id, config_str, settings = {})
-    @pipelines[pipeline_id] = LogStash::Pipeline.new(config_str, settings)
+    @pipelines[pipeline_id] = LogStash::Pipeline.new(config_str, settings.merge(:pipeline_id => pipeline_id))
   end
 
   private
--- a/logstash-core/lib/logstash/config/config_ast.rb
+++ b/logstash-core/lib/logstash/config/config_ast.rb
@@ -144,7 +144,7 @@ module LogStash; module Config; module AST
 
 
         code << <<-CODE
-#{name} = #{plugin.compile_initializer}
+          #{name} = #{plugin.compile_initializer}
           @#{plugin.plugin_type}s << #{name}
         CODE
 
@@ -152,7 +152,7 @@ module LogStash; module Config; module AST
         if plugin.plugin_type == "filter"
 
           code << <<-CODE
-#{name}_flush = lambda do |options, &block|
+            #{name}_flush = lambda do |options, &block|
               @logger.debug? && @logger.debug(\"Flushing\", :plugin => #{name})
 
               events = #{name}.flush(options)
@@ -232,18 +232,18 @@ module LogStash; module Config; module AST
 
     def compile
       case plugin_type
-        when "input"
-          return "start_input(#{variable_name})"
-        when "filter"
-          return <<-CODE
+      when "input"
+        return "start_input(#{variable_name})"
+      when "filter"
+        return <<-CODE
           events = #{variable_name}.multi_filter(events)
-          CODE
-        when "output"
-          return "targeted_outputs << #{variable_name}\n"
-        when "codec"
-          settings = attributes.recursive_select(Attribute).collect(&:compile).reject(&:empty?)
-          attributes_code = "LogStash::Util.hash_merge_many(#{settings.map { |c| "{ #{c} }" }.join(", ")})"
-          return "plugin(#{plugin_type.inspect}, #{plugin_name.inspect}, #{attributes_code})"
+        CODE
+      when "output"
+        return "targeted_outputs << #{variable_name}\n"
+      when "codec"
+        settings = attributes.recursive_select(Attribute).collect(&:compile).reject(&:empty?)
+        attributes_code = "LogStash::Util.hash_merge_many(#{settings.map { |c| "{ #{c} }" }.join(", ")})"
+        return "plugin(#{plugin_type.inspect}, #{plugin_name.inspect}, #{attributes_code})"
       end
     end
 
@@ -347,7 +347,7 @@ module LogStash; module Config; module AST
                  :column => input.column_of(interval.first),
                  :byte => interval.first + 1,
                  :after => input[0..interval.first]
-          )
+                )
         )
       end
     end
@@ -406,7 +406,7 @@ module LogStash; module Config; module AST
         CODE
       else # Output
         <<-CODE
-#{super}
+          #{super}
           end
         CODE
       end
@@ -527,7 +527,7 @@ class Treetop::Runtime::SyntaxNode
     tv = "...#{tv[-20..-1]}" if tv.size > 20
 
     indent +
-      self.class.to_s.sub(/.*:/,'') +
+    self.class.to_s.sub(/.*:/,'') +
       em.map{|m| "+"+m.to_s.sub(/.*:/,'')}*"" +
       " offset=#{interval.first}" +
       ", #{tv.inspect}" +
@@ -535,11 +535,11 @@ class Treetop::Runtime::SyntaxNode
       (elements && elements.size > 0 ?
         ":" +
           (elements.select { |e| !e.is_a?(LogStash::Config::AST::Whitespace) && e.elements && e.elements.size > 0 }||[]).map{|e|
-            begin
-              "\n"+e.inspect(indent+"  ")
-            rescue  # Defend against inspect not taking a parameter
-              "\n"+indent+" "+e.inspect
-            end
+      begin
+        "\n"+e.inspect(indent+"  ")
+      rescue  # Defend against inspect not taking a parameter
+        "\n"+indent+" "+e.inspect
+      end
           }.join("") :
         ""
       )
--- a/logstash-core/lib/logstash/config/config_ast.rb
+++ b/logstash-core/lib/logstash/config/config_ast.rb
@@ -108,6 +108,7 @@ module LogStash; module Config; module AST
         # defines @filter_func and @output_func
 
         definitions << "def #{type}_func(event)"
+        definitions << "  targeted_outputs = []" if type == "output"
         definitions << "  events = [event]" if type == "filter"
         definitions << "  @logger.debug? && @logger.debug(\"#{type} received\", :event => event.to_hash)"
 
@@ -116,6 +117,7 @@ module LogStash; module Config; module AST
         end
 
         definitions << "  events" if type == "filter"
+        definitions << "  targeted_outputs" if type == "output"
         definitions << "end"
       end
 
@@ -142,7 +144,7 @@ module LogStash; module Config; module AST
 
 
         code << <<-CODE
-          #{name} = #{plugin.compile_initializer}
+#{name} = #{plugin.compile_initializer}
           @#{plugin.plugin_type}s << #{name}
         CODE
 
@@ -150,7 +152,7 @@ module LogStash; module Config; module AST
         if plugin.plugin_type == "filter"
 
           code << <<-CODE
-            #{name}_flush = lambda do |options, &block|
+#{name}_flush = lambda do |options, &block|
               @logger.debug? && @logger.debug(\"Flushing\", :plugin => #{name})
 
               events = #{name}.flush(options)
@@ -230,18 +232,18 @@ module LogStash; module Config; module AST
 
     def compile
       case plugin_type
-      when "input"
-        return "start_input(#{variable_name})"
-      when "filter"
-        return <<-CODE
+        when "input"
+          return "start_input(#{variable_name})"
+        when "filter"
+          return <<-CODE
           events = #{variable_name}.multi_filter(events)
-        CODE
-      when "output"
-        return "#{variable_name}.handle(event)\n"
-      when "codec"
-        settings = attributes.recursive_select(Attribute).collect(&:compile).reject(&:empty?)
-        attributes_code = "LogStash::Util.hash_merge_many(#{settings.map { |c| "{ #{c} }" }.join(", ")})"
-        return "plugin(#{plugin_type.inspect}, #{plugin_name.inspect}, #{attributes_code})"
+          CODE
+        when "output"
+          return "targeted_outputs << #{variable_name}\n"
+        when "codec"
+          settings = attributes.recursive_select(Attribute).collect(&:compile).reject(&:empty?)
+          attributes_code = "LogStash::Util.hash_merge_many(#{settings.map { |c| "{ #{c} }" }.join(", ")})"
+          return "plugin(#{plugin_type.inspect}, #{plugin_name.inspect}, #{attributes_code})"
       end
     end
 
@@ -345,7 +347,7 @@ module LogStash; module Config; module AST
                  :column => input.column_of(interval.first),
                  :byte => interval.first + 1,
                  :after => input[0..interval.first]
-                )
+          )
         )
       end
     end
@@ -402,9 +404,9 @@ module LogStash; module Config; module AST
         <<-CODE
           events = cond_func_#{i}(events)
         CODE
-      else
+      else # Output
         <<-CODE
-          #{super}
+#{super}
           end
         CODE
       end
@@ -525,7 +527,7 @@ class Treetop::Runtime::SyntaxNode
     tv = "...#{tv[-20..-1]}" if tv.size > 20
 
     indent +
-    self.class.to_s.sub(/.*:/,'') +
+      self.class.to_s.sub(/.*:/,'') +
       em.map{|m| "+"+m.to_s.sub(/.*:/,'')}*"" +
       " offset=#{interval.first}" +
       ", #{tv.inspect}" +
@@ -533,11 +535,11 @@ class Treetop::Runtime::SyntaxNode
       (elements && elements.size > 0 ?
         ":" +
           (elements.select { |e| !e.is_a?(LogStash::Config::AST::Whitespace) && e.elements && e.elements.size > 0 }||[]).map{|e|
-      begin
-        "\n"+e.inspect(indent+"  ")
-      rescue  # Defend against inspect not taking a parameter
-        "\n"+indent+" "+e.inspect
-      end
+            begin
+              "\n"+e.inspect(indent+"  ")
+            rescue  # Defend against inspect not taking a parameter
+              "\n"+indent+" "+e.inspect
+            end
           }.join("") :
         ""
       )
--- a/logstash-core/lib/logstash/output_delegator.rb
+++ b/logstash-core/lib/logstash/output_delegator.rb
@@ -10,6 +10,8 @@ require "concurrent/atomic/atomic_fixnum"
 module LogStash; class OutputDelegator
   attr_reader :workers, :config, :worker_count
 
+  # The *args this takes are the same format that a Outputs::Base takes. A list of hashes with parameters in them
+  # Internally these just get merged together into a single hash
   def initialize(logger, klass, *args)
     @logger = logger
     @config = args.reduce({}, :merge)
--- a/logstash-core/lib/logstash/output_delegator.rb
+++ b/logstash-core/lib/logstash/output_delegator.rb
@@ -8,13 +8,13 @@ require "concurrent/atomic/atomic_fixnum"
 #
 # This plugin also records some basic statistics
 module LogStash; class OutputDelegator
-  attr_reader :options, :workers
+  attr_reader :workers, :config, :worker_count
 
   def initialize(logger, klass, *args)
     @logger = logger
-    @options = args.reduce({}, :merge)
+    @config = args.reduce({}, :merge)
     @klass = klass
-    @worker_count = @options["workers"] || 1
+    @worker_count = @config["workers"] || 1
 
     @worker_queue = SizedQueue.new(@worker_count)
 
@@ -26,6 +26,10 @@ module LogStash; class OutputDelegator
     end
 
     @events_received = Concurrent::AtomicFixnum.new(0)
+  end
+
+  def config_name
+    @klass.config_name
   end
 
   def register
@@ -54,5 +58,9 @@ module LogStash; class OutputDelegator
 
   def events_received
     @events_received.value
+  end
+
+  def busy_workers
+    @worker_queue.size
   end
 end end
--- a/logstash-core/lib/logstash/output_delegator.rb
+++ b/logstash-core/lib/logstash/output_delegator.rb
@@ -1,0 +1,58 @@
+# encoding: utf-8
+require "concurrent/atomic/atomic_fixnum"
+
+# This class goes hand in hand with the pipeline to provide a pool of
+# free workers to be used by pipeline worker threads. The pool is
+# internally represented with a SizedQueue set the the size of the number
+# of 'workers' the output plugin is configured with.
+#
+# This plugin also records some basic statistics
+module LogStash; class OutputDelegator
+  attr_reader :options, :workers
+
+  def initialize(logger, klass, *args)
+    @logger = logger
+    @options = args.reduce({}, :merge)
+    @klass = klass
+    @worker_count = @options["workers"] || 1
+
+    @worker_queue = SizedQueue.new(@worker_count)
+
+    @workers = @worker_count.times.map do
+      w = @klass.new(*args)
+      w.register
+      @worker_queue << w
+      w
+    end
+
+    @events_received = Concurrent::AtomicFixnum.new(0)
+  end
+
+  def register
+    @workers.each {|w| w.register}
+  end
+
+  def multi_receive(events)
+    @events_received.increment(events.length)
+
+    worker = @worker_queue.pop
+    begin
+      worker.multi_receive(events)
+    ensure
+      @worker_queue.push(worker)
+    end
+  end
+
+  def do_close
+    @logger.debug("closing output delegator", :klass => self)
+
+    @worker_count.times do
+      worker = @worker_queue.pop
+      worker.do_close
+    end
+  end
+
+  def events_received
+    @events_received.value
+  end
+end end
--- a/logstash-core/lib/logstash/outputs/base.rb
+++ b/logstash-core/lib/logstash/outputs/base.rb
@@ -25,7 +25,7 @@ class LogStash::Outputs::Base < LogStash::Plugin
   # Note that this setting may not be useful for all outputs.
   config :workers, :validate => :number, :default => 1
 
-  attr_reader :worker_plugins, :available_workers, :workers, :single_worker_mutex, :is_multi_worker, :worker_plugins
+  attr_reader :worker_plugins, :available_workers, :workers, :single_worker_mutex, :worker_plugins
 
   public
   def workers_not_supported(message=nil)
@@ -61,12 +61,9 @@ class LogStash::Outputs::Base < LogStash::Plugin
 
   public
   def worker_setup
-    # TODO: Remove this branch, delete this function
     if @workers == 1
-      @is_multi_worker = false
       @worker_plugins = [self]
     else
-      @is_multi_worker = true
       define_singleton_method(:multi_handle, method(:handle_worker))
 
       @worker_plugins = @workers.times.map { self.class.new(@original_params.merge("workers" => 1)) }
@@ -88,9 +85,7 @@ class LogStash::Outputs::Base < LogStash::Plugin
 
   # To be overriden in implementations
   def multi_receive(events)
-    events.each {|event|
-      receive(event)
-    }
+    events.each {|event| receive(event) }
   end
 
   # Not to be overriden by plugin authors!
@@ -108,11 +103,10 @@ class LogStash::Outputs::Base < LogStash::Plugin
   end
 
   def do_close
-    if @worker_plugins
-      @worker_plugins.each do |wp|
-        wp.do_close unless wp === self
-      end
+    @worker_plugins.each do |wp|
+      wp.do_close unless wp === self
     end
+
     super
   end
 
--- a/logstash-core/lib/logstash/outputs/base.rb
+++ b/logstash-core/lib/logstash/outputs/base.rb
@@ -4,6 +4,8 @@ require "logstash/logging"
 require "logstash/plugin"
 require "logstash/namespace"
 require "logstash/config/mixin"
+require "logstash/util/wrapped_synchronous_queue"
+require "concurrent/atomic/atomic_fixnum"
 
 class LogStash::Outputs::Base < LogStash::Plugin
   include LogStash::Config::Mixin
@@ -23,7 +25,7 @@ class LogStash::Outputs::Base < LogStash::Plugin
   # Note that this setting may not be useful for all outputs.
   config :workers, :validate => :number, :default => 1
 
-  attr_reader :worker_plugins, :worker_queue, :worker_threads
+  attr_reader :worker_plugins, :available_workers, :workers, :single_worker_mutex, :is_multi_worker, :worker_plugins
 
   public
   def workers_not_supported(message=nil)
@@ -40,6 +42,11 @@ class LogStash::Outputs::Base < LogStash::Plugin
   def initialize(params={})
     super
     config_init(params)
+
+    # If we're running with a single thread we must enforce single-threaded concurrency by default
+    # Maybe in a future version we'll assume output plugins are threadsafe
+    @single_worker_mutex = Mutex.new
+    worker_setup
   end
 
   public
@@ -54,36 +61,59 @@ class LogStash::Outputs::Base < LogStash::Plugin
 
   public
   def worker_setup
+    # TODO: Remove this branch, delete this function
     if @workers == 1
+      @is_multi_worker = false
       @worker_plugins = [self]
-      @worker_threads = []
     else
-      define_singleton_method(:handle, method(:handle_worker))
-      @worker_queue = SizedQueue.new(20)
+      @is_multi_worker = true
+      define_singleton_method(:multi_handle, method(:handle_worker))
+
       @worker_plugins = @workers.times.map { self.class.new(@original_params.merge("workers" => 1)) }
-      @worker_threads = @worker_plugins.map.with_index do |plugin, i|
-        Thread.new(original_params, @worker_queue) do |params, queue|
-          LogStash::Util.set_thread_name(">#{self.class.config_name}.#{i}")
-          LogStash::Util.set_thread_plugin(self)
-          plugin.register
-          while true
-            event = queue.pop
-            plugin.handle(event)
-          end
-        end
+
+      @available_workers = SizedQueue.new(@worker_plugins.length)
+
+      @worker_plugins.each do |wp|
+        wp.register
+        @available_workers << wp
       end
     end
   end
 
   public
+  # Not to be overriden by plugin authors!
   def handle(event)
-    LogStash::Util.set_thread_plugin(self)
-    receive(event)
+    @single_worker_mutex.synchronize { receive(event) }
   end # def handle
 
-  def handle_worker(event)
-    LogStash::Util.set_thread_plugin(self)
-    @worker_queue.push(event)
+  # To be overriden in implementations
+  def multi_receive(events)
+    events.each {|event|
+      receive(event)
+    }
+  end
+
+  # Not to be overriden by plugin authors!
+  def multi_handle(events)
+    @single_worker_mutex.synchronize { multi_receive(events) }
+  end
+
+  def handle_worker(events)
+    worker = @available_workers.pop
+    begin
+      worker.multi_receive(events)
+    ensure
+      @available_workers.push(worker)
+    end
+  end
+
+  def do_close
+    if @worker_plugins
+      @worker_plugins.each do |wp|
+        wp.do_close unless wp === self
+      end
+    end
+    super
   end
 
   private
--- a/logstash-core/lib/logstash/outputs/base.rb
+++ b/logstash-core/lib/logstash/outputs/base.rb
@@ -25,7 +25,7 @@ class LogStash::Outputs::Base < LogStash::Plugin
   # Note that this setting may not be useful for all outputs.
   config :workers, :validate => :number, :default => 1
 
-  attr_reader :worker_plugins, :available_workers, :workers, :single_worker_mutex, :worker_plugins
+  attr_reader :worker_plugins, :available_workers, :workers, :worker_plugins
 
   public
   def workers_not_supported(message=nil)
@@ -46,7 +46,6 @@ class LogStash::Outputs::Base < LogStash::Plugin
     # If we're running with a single thread we must enforce single-threaded concurrency by default
     # Maybe in a future version we'll assume output plugins are threadsafe
     @single_worker_mutex = Mutex.new
-    worker_setup
   end
 
   public
@@ -60,49 +59,9 @@ class LogStash::Outputs::Base < LogStash::Plugin
   end # def receive
 
   public
-  def worker_setup
-    if @workers == 1
-      @worker_plugins = [self]
-    else
-      define_singleton_method(:multi_handle, method(:handle_worker))
-
-      @worker_plugins = @workers.times.map { self.class.new(@original_params.merge("workers" => 1)) }
-
-      @available_workers = SizedQueue.new(@worker_plugins.length)
-
-      @worker_plugins.each do |wp|
-        wp.register
-        @available_workers << wp
-      end
-    end
-  end
-
-  public
   # To be overriden in implementations
   def multi_receive(events)
     events.each {|event| receive(event) }
-  end
-
-  # Not to be overriden by plugin authors!
-  def multi_handle(events)
-    @single_worker_mutex.synchronize { multi_receive(events) }
-  end
-
-  def handle_worker(events)
-    worker = @available_workers.pop
-    begin
-      worker.multi_receive(events)
-    ensure
-      @available_workers.push(worker)
-    end
-  end
-
-  def do_close
-    @worker_plugins.each do |wp|
-      wp.do_close unless wp === self
-    end
-
-    super
   end
 
   private
--- a/logstash-core/lib/logstash/outputs/base.rb
+++ b/logstash-core/lib/logstash/outputs/base.rb
@@ -78,11 +78,6 @@ class LogStash::Outputs::Base < LogStash::Plugin
   end
 
   public
-  # Not to be overriden by plugin authors!
-  def handle(event)
-    @single_worker_mutex.synchronize { receive(event) }
-  end # def handle
-
   # To be overriden in implementations
   def multi_receive(events)
     events.each {|event| receive(event) }
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -15,6 +15,7 @@ require "logstash/shutdown_controller"
 require "logstash/util/wrapped_synchronous_queue"
 require "logstash/pipeline_reporter"
 require "concurrent/timer_task"
+require "logstash/output_delegator"
 
 module LogStash; class Pipeline
   attr_reader :inputs, :filters, :outputs, :worker_threads, :events_consumed, :events_filtered, :reporter, :pipeline_id
@@ -252,7 +253,7 @@ module LogStash; class Pipeline
     # Now that we have our output to event mapping we can just invoke each output
     # once with its list of events
     outputs_events.each do |output, events|
-      output.multi_handle(events)
+      output.multi_receive(events)
     end
   end
 
@@ -362,8 +363,14 @@ module LogStash; class Pipeline
 
   def plugin(plugin_type, name, *args)
     args << {} if args.empty?
+
     klass = LogStash::Plugin.lookup(plugin_type, name)
-    return klass.new(*args)
+
+    if plugin_type == "output"
+      LogStash::OutputDelegator.new(@logger, klass, *args)
+    else
+      klass.new(*args)
+    end
   end
 
   # for backward compatibility in devutils for the rspec helpers, this method is not used
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -240,14 +240,18 @@ module LogStash; class Pipeline
 
   # Take an array of events and send them to the correct output
   def output_batch(batch)
-    batch.reduce(Hash.new { |h, k| h[k] = [] }) do |outputs_events, event|
+    # Build a mapping of { output_plugin => [events...]}
+    outputs_events = batch.reduce(Hash.new { |h, k| h[k] = [] }) do |acc, event|
       # We ask the AST to tell us which outputs to send each event to
+      # Then, we stick it in the correct bin
       output_func(event).each do |output|
-        outputs_events[output] << event
+        acc[output] << event
       end
-      outputs_events
-    end.each do |output, events|
-      # Once we have a mapping of outputs => [events] we can execute them
+      acc
+    end
+    # Now that we have our output to event mapping we can just invoke each output
+    # once with its list of events
+    outputs_events.each do |output, events|
       output.multi_handle(events)
     end
   end
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -94,8 +94,12 @@ module LogStash; class Pipeline
       when nil
         # user did not specify a worker thread count
         # warn if the default is multiple
-        @logger.warn("Defaulting pipeline worker threads to 1 because there are some filters that might not work with multiple worker threads",
-                     :count_was => default, :filters => plugins) if default > 1
+
+        if default > 1
+          @logger.warn("Defaulting pipeline worker threads to 1 because there are some filters that might not work with multiple worker threads",
+                       :count_was => default, :filters => plugins)
+        end
+
         1 # can't allow the default value to propagate if there are unsafe filters
       when 0, 1
         1
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -14,6 +14,7 @@ require "logstash/util/defaults_printer"
 require "logstash/shutdown_controller"
 require "logstash/util/wrapped_synchronous_queue"
 require "logstash/pipeline_reporter"
+require "concurrent/timer_task"
 
 module LogStash; class Pipeline
   attr_reader :inputs, :filters, :outputs, :worker_threads, :events_consumed, :events_filtered, :reporter, :pipeline_id
@@ -21,7 +22,9 @@ module LogStash; class Pipeline
   DEFAULT_SETTINGS = {
     :default_pipeline_workers => LogStash::Config::CpuCoreStrategy.fifty_percent,
     :pipeline_batch_size => 125,
-    :pipeline_batch_delay => 5 # in milliseconds
+    :pipeline_batch_delay => 5, # in milliseconds
+    :flush_interval => 5, # in seconds
+    :flush_timeout_interval => 60 # in seconds
   }
 
   def initialize(config_str, settings = {})
@@ -65,7 +68,10 @@ module LogStash; class Pipeline
     @settings = DEFAULT_SETTINGS.clone
     # @ready requires thread safety since it is typically polled from outside the pipeline thread
     @ready = Concurrent::AtomicBoolean.new(false)
+    @flushing = Concurrent::AtomicReference.new(false)
     settings.each {|setting, value| configure(setting, value) }
+
+    start_flusher
   end # def initialize
 
   def ready?
@@ -120,6 +126,7 @@ module LogStash; class Pipeline
     wait_inputs
     @logger.info("Input plugins stopped! Will shutdown filter/output workers.")
 
+    shutdown_flusher
     shutdown_workers
 
     @logger.info("Pipeline shutdown complete.")
@@ -167,14 +174,20 @@ module LogStash; class Pipeline
 
     while running
       # To understand the purpose behind this synchronize please read the body of take_batch
-      input_batch, shutdown_received = @input_queue_pop_mutex.synchronize { take_batch(batch_size, batch_delay) }
-      running = false if shutdown_received
+      input_batch, signal = @input_queue_pop_mutex.synchronize { take_batch(batch_size, batch_delay) }
+      running = false if signal == LogStash::SHUTDOWN
+
       @events_consumed.increment(input_batch.size)
 
-      filtered = filter_batch(input_batch)
-      @events_filtered.increment(filtered.size)
+      filtered_batch = filter_batch(input_batch)
 
-      output_batch(filtered)
+      if signal == LogStash::FLUSH
+        flush_filters_to_batch(filtered_batch)
+      end
+
+      @events_filtered.increment(filtered_batch.size)
+
+      output_batch(filtered_batch)
 
       inflight_batches_synchronize { set_current_thread_inflight_batch(nil) }
     end
@@ -186,23 +199,23 @@ module LogStash; class Pipeline
     # guaranteed to be a full batch not a partial batch
     set_current_thread_inflight_batch(batch)
 
-    shutdown_received = false
+    signal = false
     batch_size.times do |t|
       event = t==0 ? @input_queue.take : @input_queue.poll(batch_delay)
       if event.nil?
         next
-      elsif event == LogStash::SHUTDOWN
-        shutdown_received = true
+      elsif event == LogStash::SHUTDOWN || event == LogStash::FLUSH
         # We MUST break here. If a batch consumes two SHUTDOWN events
         # then another worker may have its SHUTDOWN 'stolen', thus blocking
-        # the pipeline
+        # the pipeline. We should stop doing work after flush as well.
+        signal = event
         break
       else
         batch << event
       end
     end
 
-    [batch, shutdown_received]
+    [batch, signal]
   end
 
   def filter_batch(batch)
@@ -380,6 +393,53 @@ module LogStash; class Pipeline
       flusher.call(options, &block)
     end
   end
+
+  class FlusherObserver
+    def initialize(logger)
+      @logger = logger
+    end
+
+    def update(time, result, exception)
+      return unless exception
+      @logger.warn("Error during flush!",
+                   :message => exception.message,
+                   :class => exception.class.name,
+                   :backtrace => exception.backtrace)
+    end
+  end
+
+  def start_flusher
+    @flusher_task = Concurrent::TimerTask.new { flush }
+    @flusher_task.execution_interval = @settings[:flush_interval]
+    @flusher_task.timeout_interval = @settings[:flush_timeout_interval]
+    @flusher_task.add_observer(FlusherObserver.new(@logger))
+    @flusher_task.execute
+  end
+
+  def shutdown_flusher
+    @flusher_task.shutdown
+  end
+
+  def flush
+    if @flushing.compare_and_set(false, true)
+      @logger.debug? && @logger.debug("Pushing flush onto pipeline")
+      @input_queue.push(LogStash::FLUSH)
+    end
+  end
+
+  # perform filters flush into the output queue
+  # @param options [Hash]
+  # @option options [Boolean] :final => true to signal a final shutdown flush
+  def flush_filters_to_batch(batch, options = {})
+    flush_filters(options) do |event|
+      unless event.cancelled?
+        @logger.debug? and @logger.debug("Pushing flushed events", :event => event)
+        batch << event
+      end
+    end
+
+    @flushing.set(false)
+  end # flush_filters_to_output!
 
   def plugin_threads_info
     input_threads = @input_threads.select {|t| t.alive? }
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -160,7 +160,7 @@ module LogStash; class Pipeline
     end
   end
 
-  # Main body of what a worker threda does
+  # Main body of what a worker thread does
   # Repeatedly takes batches off the queu, filters, then outputs them
   def worker_loop(batch_size, batch_delay)
     running = true
@@ -292,7 +292,6 @@ module LogStash; class Pipeline
     begin
       plugin.run(@input_queue)
     rescue => e
-      # if plugin is stop
       if plugin.stop?
         @logger.debug("Input plugin raised exception during shutdown, ignoring it.",
                       :plugin => plugin.class.config_name, :exception => e,
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -18,6 +18,13 @@ require "logstash/pipeline_reporter"
 module LogStash; class Pipeline
   attr_reader :inputs, :filters, :outputs, :worker_threads, :events_consumed, :events_filtered, :reporter, :pipeline_id
 
+  DEFAULT_SETTINGS = {
+    # We don't set pipeline workers defaults here, but rather in safe_pipeline_workers
+    :default_pipeline_workers => LogStash::Config::CpuCoreStrategy.fifty_percent,
+    :pipeline_batch_size => 125,
+    :pipeline_batch_delay => 5 # in milliseconds
+  }
+
   def initialize(config_str, settings = {})
     @pipeline_id = settings[:pipeline_id] || self.object_id
     @logger = Cabin::Channel.get(LogStash)
@@ -58,11 +65,7 @@ module LogStash; class Pipeline
 
     @input_threads = []
 
-    @settings = {
-      # We don't set pipeline workers defaults here, but rather in safe_pipeline_workers
-      :pipeline_batch_size => 125,
-      :pipeline_batch_delay => 50 # in milliseconds
-    }
+    @settings = DEFAULT_SETTINGS.clone
 
     # @ready requires thread safety since it is typically polled from outside the pipeline thread
     @ready = Concurrent::AtomicBoolean.new(false)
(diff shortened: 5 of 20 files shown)


logstash-core/lib/logstash/config/config_ast.rb
```diff
@@ -230,18 +232,18 @@ def compile_initializer
 
     def compile
       case plugin_type
-      when "input"
-        return "start_input(#{variable_name})"
-      when "filter"
-        return <<-CODE
+        when "input"
```
untergeek: Are these cosmetic indentation changes part of _our_ style guidelines?  Or are these general ruby style guidelines?  I've seen the unindented style much more frequently, so I'm curious.

andrewvc: That would be my editor (rubymine) sneaking in with its opinion. Good call, I'll re-indent this stuff.

andrewvc:
@untergeek my latest code changes clean up the whitespace mangling in config_ast . Enjoy!

jordansissel:
Big change! <3

I'm going to review this in several rounds - starting with tests and userinterface, then ending with code review.

Doing testing now.

jordansissel:
I get this when doing `rake test:install-default`

```
Error Bundler::InstallError, retrying 1/10
An error occurred while installing atomic (1.1.99), and Bundler cannot continue.
Make sure that `gem install atomic -v '1.1.99'` succeeds before bundling.
WARNING: SSLSocket#session= is not supported
```

jordansissel:
Ignore my prior comment - Upon trying again, I wasn't able to reproduce it.

jordansissel:
Running the following throughput mini benchmark, I get the same results on my laptop (battery power, 2 cores)

```
% bin/logstash -e 'input { generator { } } filter { mutate { } } output { stdout { codec => dots } }' | pv -War > /dev/null
```

This PR: 62keps. 
Logstash 2.1.0.snapshot3: 62keps

logstash-core/lib/logstash/agent.rb
```diff
@@ -33,7 +33,7 @@ def execute
   end # def execute
 
   def add_pipeline(pipeline_id, config_str, settings = {})
-    @pipelines[pipeline_id] = LogStash::Pipeline.new(config_str, settings)
+    @pipelines[pipeline_id] = LogStash::Pipeline.new(config_str, settings.merge(:pipeline_id => pipeline_id))
```
purbon: does this means this changes allows us to have multiple pipelines?

ph: This change is already in the current master, but yes this is the first step to enable multiple pipelines.

purbon: nice, very interesting.

andrewvc:
@jordansissel your benchmark results are correct, but they don't really measure the salient thing here, since that pipeline is basically a no-op. The new pipeline will run your 'null' pipeline at about the same throughput, but it will run it with less user/system time. I suspect that the bottleneck is stdout. Once real filters/codecs are added in the story changes significantly. 

First, lets inspect the execution profile of the 'null' example using the `time` command. Note that the wallclock time is roughly the same

```
 ~/D/logstash-2.1.0> time bin/logstash -w 5 -e 'input { generator { count => 4000000 } } filter { mutate { } } output { stdout { codec => dots } }' | wc -c
       47.31 real       135.16 user        40.84 sys
 4000000
 ~/p/logstash /ng_pipeline> time bin/logstash -w 5 -e 'input { generator { count => 4000000 } } filter { mutate { } } output { stdout { codec => dots } }' | wc -c
       40.95 real        96.60 user        17.16 sys
 4000000
```

Now, let's take a look at a super-light 'real world' example. The new pipeline has much higher throughput.

```
 ~/D/logstash-2.1.0> time bin/logstash -w 5 -e 'input { generator { count => 4000000 } } filter { mutate { add_field => {"x" => "y"}} mutate { add_field => {"blah" => "bot"}} } output { stdout { codec => json_lines } }' | wc -l
       87.34 real       294.83 user        72.91 sys
 4000000
 ~/p/logstash /ng_pipeline> time bin/logstash -w 5 -e 'input { generator { count => 4000000 } } filter { mutate { add_field => {"x" => "y"}} mutate { add_field => {"blah" => "bot"}} } output { stdout { codec => json_lines } }' | wc -l
       68.04 real       210.77 user        26.39 sys
 4000000
```

jordansissel:
Hmm good points on the cpu time differences. The new architecture (one less message passing point) really shows up on the sys (kernel/contextswitching) area.

jordansissel:
I'll test this on a machine with more cpus than my tiny laptop also ;P

andrewvc:
@jordansissel yep! Thanks for taking a look. Curious to see your benchmarks. It's interesting that on my box the results were more skewed in favor of the new pipeline (maybe my CPU is different?). To be clear I think the performance is a combination of the following three factors.

0. Less context switching (as you say).
1. SynchronousQueue may just be a faster impl. (which I'd expect given that it's a simpler construct)
2. Coarser context switch granularity with the batching approach (maybe more CPU cache friendly).

logstash-core/lib/logstash/agent.rb
andrewvc: One thing to note is I made the pipeline ID part of the thread names so it looks nice in VisualVM!

suyograo:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;B4ZSDeVI&#x2F;1-abc"> abc</a>

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -12,20 +12,32 @@
...
 require "logstash/shutdown_controller"
+require "logstash/util/wrapped_synchronous_queue"
+require "logstash/pipeline_reporter"
 
 module LogStash; class Pipeline
-  attr_reader :inputs, :filters, :outputs, :input_to_filter, :filter_to_output
+  attr_reader :inputs, :filters, :outputs, :worker_threads, :events_consumed, :events_filtered, :reporter, :pipeline_id
+
+  DEFAULT_SETTINGS = {
+    # We don't set pipeline workers defaults here, but rather in safe_pipeline_workers
```
jordansissel: This comment confuses me when I read it. It says we don't set the default pipeline workers, but there's a key called `:default_pipeline_workers` on the next line. Help? :)

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +112,164 @@ def filters?
...
+    # Since this is externally synchronized in `worker_look` wec can guarantee that the visibility of an insight batch
+    # guaranteed to be a full batch not a partial batch
+    set_current_thread_inflight_batch(batch)
+
+    shutdown_received = false
+    batch_size.times do |t|
+      event = t==0 ? @input_queue.take : @input_queue.poll(batch_delay)
+      # Exit early so each thread only gets one copy of this
+      # This is necessary to ensure proper shutdown!
+      next if event.nil?
```
jordansissel: Is the comment above correct? The way I read it, queue.take will never return nil, and queue.poll will return nil only if the queue was empty during the `batch_delay` time. In this case, I read it to mean that there is nothing in the queue so that we should be done with our loop `batch_size.times` loop. 

Would rewriting this block as such make more sense?

```ruby
if event.nil?
  # Break early if there was nothing in the queue.
  break
elsif event == LogStash::SHUTDOWN
  shutdown_received = true
else
  batch << event
end
```

Then delete the `break` below in the SHUTDOWN check (line 205).

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +112,164 @@ def filters?
...
+    [batch, shutdown_received]
+  end
 
-  def wait_inputs
-    @input_threads.each(&:join)
+  def filter_batch(batch)
+    batch.reduce([]) do |acc,e|
+      if e.is_a?(LogStash::Event)
+        filtered = filter_func(e)
+        filtered.each {|fe| acc << fe unless fe.cancelled?}
```
jordansissel: Subjective, but I think this may be more readable as:

```ruby
acc += filtered.reject(&:cancelled?)
```

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +112,164 @@ def filters?
...
-    return 0
-  end # def run
+    [batch, shutdown_received]
+  end
 
-  def wait_inputs
-    @input_threads.each(&:join)
+  def filter_batch(batch)
+    batch.reduce([]) do |acc,e|
+      if e.is_a?(LogStash::Event)
```
jordansissel: I may be forgetting, but why would `e` not be an Event? When? Should we comment/document when this occurs?

logstash-core/lib/logstash/pipeline_reporter.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    def to_hash
+      @data
+    end
+
+    def to_simple_hash
+      {"inflight_count" => inflight_count, "stalling_thread_info" => format_threads_by_plugin}
+    end
+
+    def method_missing(meth)
```
jordansissel: Not a fan of method_missing lately since it often makes things more difficult to debug. Tell me a story about why Snapshot needs this kind of interface?

jordansissel:
Found a stall:

```
% bin/logstash --verbose -e 'input { stdin { } } output { if "one" == "two" { stdout { } } }'
Settings:
Starting pipeline {:id=>"base", :pipeline_workers=>2, :batch_size=>125, :batch_delay=>5, :level=>:info}
Pipeline started {:level=>:info}
Logstash startup completed
Will run till input threads stopped {:level=>:info}
hello
world
fancy
pants
Closing inputs {:level=>:info}
Closed inputs {:level=>:info}
data loss. {:level=>:warn}
```

I haven't looked into the cause, but it is reproducible every time for me.

Basically, stdin -> stdout w/ a conditional preventing anything from going from stdout, send a few events, ^C, and it stalls shutting down.

jordansissel:
Here's debug output:

```
Logstash startup completed
Will run till input threads stopped {:level=>:info, :file=>"logstash/pipeline.rb", :line=>"123", :method=>"run"}
hello
^CSIGINT received. Shutting down the pipeline. {:level=>:warn, :file=>"logstash/agent.rb", :line=>"72", :method=>"trap_sigint"}
Closing inputs {:level=>:info, :file=>"logstash/pipeline.rb", :line=>"342", :method=>"shutdown"}
stopping {:plugin=><LogStash::Inputs::Stdin codec=><LogStash::Codecs::Line charset=>"UTF-8">>, :level=>:debug, :file=>"logstash/inputs/base.rb", :line=>"81", :method=>"do_stop"}
Closed inputs {:level=>:info, :file=>"logstash/pipeline.rb", :line=>"344", :method=>"shutdown"}
filter received {:event=>{"message"=>"hello", "@version"=>"1", "@timestamp"=>"2015-12-03T01:42:52.792Z", "host"=>"localhost.localdomain"}, :level=>:debug, :file=>"(eval)", :line=>"17
", :method=>"filter_func"}
output received {:event=>{"message"=>"hello", "@version"=>"1", "@timestamp"=>"2015-12-03T01:42:52.792Z", "host"=>"localhost.localdomain"}, :level=>:debug, :file=>"(eval)", :line=>"22
", :method=>"output_func"}
data loss. {:level=>:warn, :file=>"logstash/agent.rb", :line=>"73", :method=>"trap_sigint"}
```

logstash-core/lib/logstash/pipeline_reporter.rb
andrewvc: The goal was to provide something like a hash that we could easily decorate with methods like `to_simple_hash`. This is really a Hash decorator. 

I'd love a better solution however!

logstash-core/lib/logstash/pipeline.rb
andrewvc: This is VERY confusing. I agree. The comment should be changed.

That being said, the whole idea of safe pipeline workers is temporary and is only here to not change too much compared to the current pipeline. I plan on submitting a patch that removes the whole idea of 'safe' pipeline workers by just using a mutex around filters that aren't thread safe. I've already opened a PR about it #4272.  That should simplify the whole situation.

logstash-core/lib/logstash/pipeline.rb
andrewvc: I believe the comment was meant to apply to LogStash::SHUTDOWN, not event.nil?

I'm not sure in which case event would be nil? except in the case of a misbehaving plugin (which is why we should keep it). That deserves its own comment.

I believe you're right that we can simplify the logic, but I think your example is incorrect since we should just call `next` not `break` if event is nil. We should break if Shutdown is received however, because we don't expect to keep taking events if a shutdown is Found.

logstash-core/lib/logstash/pipeline.rb
andrewvc: More readable but less performant. `+=` allocates a new array and reassigns the value, `<<` is in-place. Given that this is the pipeline I think it makes sense to err on the side of speed.

andrewvc:
@jordansissel Thanks for the careful review, and good working finding the stall!

Interestingly `bin/logstash --verbose -e 'input { generator { count => 1 } } output { if "one" == "two" { stdout { } } }'` does not exhibit that error. Also interestingly I can unstall the pipeline with stdin by hitting enter once after hitting ctrl-c. So, somehow the stdin is blocking it.

Also interestingly, that same config causes the current pipeline to break in 2.1.0. Luckily, it won't stall, but it will raise an exception!

```
bin/logstash --verbose -e 'input { stdin {} } output { if "one" == "two" { stdout { } } }'
Pipeline started {:level=>:info}
Logstash startup completed
ocreau
oreuc
NoMethodError: private method `select' called for nil:NilClass
    plugin_threads at /Users/andrewvc/Downloads/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/pipeline.rb:373
  stalling_threads at /Users/andrewvc/Downloads/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/pipeline.rb:364
     from_pipeline at /Users/andrewvc/Downloads/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/shutdown_controller.rb:102
             start at /Users/andrewvc/Downloads/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/shutdown_controller.rb:50
              call at org/jruby/RubyProc.java:271
          interval at /Users/andrewvc/Downloads/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/stud-0.0.22/lib/stud/interval.rb:20
             start at /Users/andrewvc/Downloads/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/shutdown_controller.rb:49
             start at /Users/andrewvc/Downloads/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/shutdown_controller.rb:38
~/D/logstash-2.1.0 $
```

jsvd:
@andrewvc this select exception has been fixed in 2.x https://github.com/elastic/logstash/commit/21dfae16523bba8f04f3cb9f703e67e971287339

As for the stall using stdin, hasn't it always been this way? Colin has made a PR some months ago to make stdin interruptible but for now it always requires either a double ctrl+c, ctrl+d, or ctrl+C + enter

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -54,36 +61,59 @@ def receive(event)
 
   public
   def worker_setup
+    # TODO: Remove this branch, delete this function
     if @workers == 1
+      @is_multi_worker = false
```
jsvd: I'm not sure if it's worth defining this variable it's not used anywhere.

In https://github.com/elastic/logstash/pull/4254/files#diff-1646a27fa8dcf9f75d0b65aeaa7c9047R100 it is defined again

logstash-core/lib/logstash/agent.rb
jsvd: @andrewvc this pipeline_id handling is quirky,I see that in pipeline.rb if the id is nil then `object_id` is used. maybe we could do it here instead? because otherwise calling `add_pipeline(nil, "") will create a nil key in @pipelines but the pipeline.id will return `object_id`

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +112,164 @@ def filters?
...
       @ready.make_true
     end
+  end
 
-    @logger.info("Pipeline started")
-    @logger.terminal("Logstash startup completed")
+  # Main body of what a worker threda does
+  # Repeatedly takes batches off the queu, filters, then outputs them
+  def worker_loop(batch_size, batch_delay)
+    running = true
```
jsvd: I'd suggest removing this running and doing a simple

```ruby
loop do
  ....
  break if shutdown_received
end
```

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -42,13 +54,18 @@ def initialize(config_str, settings = {})
...
+    @input_queue = LogStash::Util::WrappedSynchronousQueue.new
+    @events_filtered = Concurrent::AtomicFixnum.new(0)
+    @events_consumed = Concurrent::AtomicFixnum.new(0)
+
+    # We generally only want one thread at a time able to access pop/take/poll operations
+    # from this queue. We also depend on this to be able to block consumers while we snapshot
+    # in-flight buffers
+    @input_queue_pop_mutex = Mutex.new
+
+    @input_threads = []
```
jsvd: this is already done at line 72

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -64,25 +81,26 @@ def configure(setting, value)
...
-        @logger.warn("Defaulting filter worker threads to 1 because there are some filters that might not work with multiple worker threads",
-          :count_was => default, :filters => plugins) if default > 1
-        1 # can't allow the default value to propagate if there are unsafe filters
-      when 0, 1
-        1
-      else
-        @logger.warn("Warning: Manual override - there are filters that might not work with multiple worker threads",
-          :worker_threads => thread_count, :filters => plugins)
-        thread_count # allow user to force this even if there are unsafe filters
+        when nil
```
jsvd: indentation-wise I think we're still using the same level, like:

```ruby
case X
when Y ...
else
end
```

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +112,164 @@ def filters?
...
   def run
+    LogStash::Util.set_thread_name("[#{pipeline_id}]-pipeline-manager")
     @logger.terminal(LogStash::Util::DefaultsPrinter.print(@settings))
 
+    start_workers
+
+    @logger.info("Pipeline started")
+    @logger.terminal("Logstash startup completed")
+
+    @logger.info("Will run till input threads stopped")
```
jsvd: not sure this is useful, any special expectation you're trying to set here for the user?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +112,164 @@ def filters?
...
+
+    # Block until all inputs have stopped
+    # Generally this happens if SIGINT is sent and `shutdown` is called from an external thread
+    wait_inputs
+    @logger.info("Inputs stopped")
```
jsvd: maybe "Input plugins stopped. Terminating filter/output workers.." ?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -167,45 +289,16 @@ def start_inputs
...
   def inputworker(plugin)
-    LogStash::Util.set_thread_name("<#{plugin.class.config_name}")
-    LogStash::Util.set_thread_plugin(plugin)
+    LogStash::Util::set_thread_name("[#{pipeline_id}]<#{plugin.class.config_name}")
     begin
-      plugin.run(@input_to_filter)
+      plugin.run(@input_queue)
     rescue => e
-      # if plugin is stopping, ignore uncatched exceptions and exit worker
+      # if plugin is stop
```
jsvd: somewhat useless comment, I suggest removing it

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -296,9 +339,30 @@ def shutdown(&before_stop)
...
+      @logger.debug("Pushing shutdown", :thread => t)
+      @input_queue.push(LogStash::SHUTDOWN)
+    end
+
+    @worker_threads.each do |t|
+      @logger.debug("Shutdown waiting for worker thread #{t}")
+      t.join
+    end
+
+    @filters.each(&:do_close)
```
jsvd: should we protect these do_close calls with rescues? not sure..

logstash-core/lib/logstash/agent.rb
ph: I agree with @jsvd comment.

But this scenario is also problematic if you are trying to add a new pipeline with an existing key, but I think this can be solve in another PR.

logstash-core/lib/logstash/config/config_ast.rb
```diff
@@ -108,6 +108,7 @@ def compile
         # defines @filter_func and @output_func
 
         definitions << "def #{type}_func(event)"
```
ph: There is an error here when you run multiples pipeline and the way we are generating the code.
the `def` need to be a `define_singleton_method`

I have a PR to fix it in https://github.com/elastic/logstash/pull/4298

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -54,36 +61,59 @@ def receive(event)
 
   public
   def worker_setup
+    # TODO: Remove this branch, delete this function
```
ph: TODO?

logstash-core/lib/logstash/pipeline.rb
andrewvc: I should really fix my rubymine defaults here. Thanks for catching it :)

logstash-core/lib/logstash/pipeline.rb
guyboertje: The comment is confusing, I wrote it, it was meant to be read with the PR diff so the reviewer did not ask 'hey is this a mistake, don't we need this setting?

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -54,36 +61,59 @@ def receive(event)
...
   end # def handle
 
-  def handle_worker(event)
-    LogStash::Util.set_thread_plugin(self)
-    @worker_queue.push(event)
+  # To be overriden in implementations
+  def multi_receive(events)
+    events.each {|event|
+      receive(event)
+    }
```
ph: one line maybe?

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -54,36 +61,59 @@ def receive(event)
...
+    # TODO: Remove this branch, delete this function
     if @workers == 1
+      @is_multi_worker = false
       @worker_plugins = [self]
-      @worker_threads = []
     else
-      define_singleton_method(:handle, method(:handle_worker))
-      @worker_queue = SizedQueue.new(20)
+      @is_multi_worker = true
+      define_singleton_method(:multi_handle, method(:handle_worker))
```
ph: Redefining the method seems a bit extreme to me? 
Can we just set the proc to an instance variable and multi_handle proxy the call?

logstash-core/lib/logstash/pipeline.rb
andrewvc: I'm just leaning toward excising the whole safe_worker  thing at this point, it just makes the code more complicated. I suppose it could be in scope for this PR, and needn't be separate in #4272

ph: @andrewvc +1 make it another PR.

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -54,36 +61,59 @@ def receive(event)
...
+      @available_workers.push(worker)
+    end
+  end
+
+  def do_close
+    if @worker_plugins
+      @worker_plugins.each do |wp|
+        wp.do_close unless wp === self
+      end
+    end
```
ph: From the code `@worker_plugins` will always be an array since we set it in the initialize method, so we dont need to do `if @worker_plugins`

logstash-core/lib/logstash/pipeline_reporter.rb
```diff
@@ -0,0 +1,119 @@
...
+  # this provides a (more or less) consistent snapshot of what's going on in the
+  # pipeline with some extra decoration
+  def snapshot
+    Snapshot.new(self.to_hash)
+  end
+
+  def to_hash
+    pipeline.inflight_batches_synchronize do |batch_map|
+      worker_states_snap = worker_states(batch_map) # We only want to run this once
+      inflight_count = worker_states_snap.map {|s| s[:inflight_count] }.reduce(:+) || 0
```
jsvd: `inflight_count = worker_states_snap.map {|s| s[:inflight_count] }.reduce(0, :+)`

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +112,164 @@ def filters?
...
     ensure
       # it is important to garantee @ready to be true after the startup sequence has been completed
       # to potentially unblock the shutdown method which may be waiting on @ready to proceed
       @ready.make_true
     end
+  end
 
-    @logger.info("Pipeline started")
-    @logger.terminal("Logstash startup completed")
+  # Main body of what a worker threda does
```
ph: worker **thread** does

logstash-core/lib/logstash/pipeline.rb
ph:
```
loop do
end
```
is slower than `while true`  

https://github.com/JuanitoFatas/fast-ruby#loop-vs-while-true-code

logstash-core/lib/logstash/pipeline_reporter.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+  end
+
+  def initialize(logger,pipeline)
+    @logger = logger
+    @pipeline = pipeline
+  end
+
+  def report
+    @logger.warn ["INFLIGHT_EVENTS_REPORT", Time.now.iso8601, report]
```
jsvd: ∞

logstash-core/lib/logstash/pipeline_reporter.rb
jsvd: the easiest alternative is going back to hash lookups, imo

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +112,164 @@ def filters?
...
+      # We ask the AST to tell us which outputs to send each event to
+      output_func(event).each do |output|
+        outputs_events[output] << event
+      end
+      outputs_events
+    end.each do |output, events|
+      # Once we have a mapping of outputs => [events] we can execute them
+      output.multi_handle(events)
+    end
+  end
```
ph: Can we split the iterators a bit, it makes things a bit harder to follow/read

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +112,164 @@ def filters?
...
+  def inflight_batches_synchronize
+    @input_queue_pop_mutex.synchronize do
+      yield(@inflight_batches)
+    end
   end
 
-  def shutdown_outputs
-    # nothing, filters will do this
-    @filter_to_output.push(LogStash::SHUTDOWN)
+  def dump_inflight(file_path)
```
ph: Where this method is called?

logstash-core/lib/logstash/pipeline.rb
ph: @jsvd we were not doing it before... ;)

We could at a rescue block on the method and add a logger?

jsvd: I know, the thing is that we're preparing the agent/pipeline more and more to be restart-proof.
anyway it can be done later

logstash-core/lib/logstash/pipeline_reporter.rb
ph: Use `forwardable` and `def_delegator`.

Example at the end: http://words.steveklabnik.com/beware-subclassing-ruby-core-classes

logstash-core/lib/logstash/shutdown_controller.rb
```diff
@@ -47,10 +47,10 @@ def start
       cycle_number = 0
       stalled_count = 0
       Stud.interval(@cycle_period) do
-        @reports << Report.from_pipeline(@pipeline)
+        @reports << pipeline_report_snapshot
         @reports.delete_at(0) if @reports.size > @report_every # expire old report
         if cycle_number == (@report_every - 1) # it's report time!
-          logger.warn(@reports.last.to_hash)
+          logger.warn(@reports.last.to_simple_hash)
```
ph: You don't really need to explicitly call `to_simple_hash` here, you can implement `to_str` in the snapshot class, when the string conversion occur ruby will call `to_str`

logstash-core/lib/logstash/shutdown_controller.rb
```diff
@@ -78,7 +82,7 @@ def shutdown_stalled?
       return false unless @reports.size == @report_every #
       # is stalled if inflight count is either constant or increasing
       stalled_event_count = @reports.each_cons(2).all? do |prev_report, next_report|
-        prev_report.inflight_count["total"] <= next_report.inflight_count["total"]
+        prev_report.inflight_count <= next_report.inflight_count
```
ph: I think I made this comment before, but we could a `Comparable`?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -0,0 +1,25 @@
+module LogStash; module Util
```
ph: # encoding: utf-8

logstash-core/lib/logstash/shutdown_controller.rb
jsvd: @ph we have thought about this, the problem is that what we're trying to gauge here (is it stalled or not) is not a property between 2 reports, but of an array of reports. for this we test

for an array is condition X true?
if so, for the array, is condition Y true?

This means we need 2 Comparable strategies

logstash-core/lib/logstash/pipeline_reporter.rb
guyboertje: Or, what are the chances of method_missing being invoked? Is Andrew's solution suitable if the chances are very low?
Maybe include a couple of methods for well known keys and leave method_missing for the rest.

logstash-core/lib/logstash/shutdown_controller.rb
ph: thanks for remembering!

logstash-core/lib/logstash/agent.rb
andrewvc: @jsvd @ph yep, great observation. I'll move that logic here.

ph:
Added minor comments to the codec I an currently doing some manual testing.

logstash-core/lib/logstash/agent.rb
andrewvc: @jsvd @ph looking through the code the reason I made the UUID generation automatic inside of pipeline was that for testing this is pretty inconvenient.

If we're thinking of making pipeline_id required it should change the signature of the constructor to `Pipeline.new(pipeline_id, config_str, settings)`. Are you guys OK with this change? It will make the PR a good bit larger as all tests instantiating the pipeline will need to change.

ph: @andrewvc Lets make that into another PR :(

logstash-core/lib/logstash/outputs/base.rb
andrewvc: That TODO is inaccurate and out of date, will remove!

logstash-core/lib/logstash/outputs/base.rb
andrewvc: Good point! Will remove this ivar.

logstash-core/lib/logstash/outputs/base.rb
andrewvc: I assume it was built this way for speed? I'm just following the pattern the base output used before. I'm totally open to just using a Proc, but I'm curious if anyone else has any insight here.

logstash-core/lib/logstash/outputs/base.rb
andrewvc: yep! Should be :)

logstash-core/lib/logstash/outputs/base.rb
andrewvc: good call

logstash-core/lib/logstash/outputs/base.rb
ph: I don't know..  :D

logstash-core/lib/logstash/pipeline.rb
andrewvc: good call!

logstash-core/lib/logstash/pipeline.rb
andrewvc: Normally I'd go for semantics, but in this case given that its the pipeline I don't mind erring on the side of speed.

jsvd: +1 then

```
while !shutdown_received
  ....
end
```

logstash-core/lib/logstash/pipeline.rb
andrewvc: I just rewrote this section and pushed it up. Should be more clear now!

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
andrewvc: This one is always tricky to remember :)

logstash-core/lib/logstash/pipeline_reporter.rb
andrewvc: Nice little trick :)

logstash-core/lib/logstash/pipeline_reporter.rb
andrewvc: Haha, luckily this isn't called anywhere :)

logstash-core/lib/logstash/pipeline.rb
andrewvc: I left this here because one of our goals was the ability to have a safe shutdown that could recover in flight messages. I thought it made sense to leave this in the commit to be exposed by a later patch.

Should I pull it out?

logstash-core/lib/logstash/pipeline.rb
andrewvc: Just because it isn't even a complete sentence and contains less information than the code that follows it doesn't make it totally useless... :wink:

logstash-core/lib/logstash/pipeline_reporter.rb
andrewvc: Having read through these comments I agree with the concerns raised in the general case, but I don't think they apply in this specific case (as @guyboertje mentions).

@ph I like forwardable and def_delegator but I don't see what concrete benefit it provides here.

These methods will be called relatively infrequently, so the perf impact is negligible. I also don't see why any errors here would be hard to debug, because everything gets forwarded to a hash lookup. This is basically a decorated OStruct.

ph:
@jordansissel This stall isn't from the change in this PRs, the current stdin doesn't allow to be interrupted, we have that problem since the birth of Logstash.

@colinsurprenant has created a PR to implement an interruptible stdin implementation for jruby. See this gem https://github.com/colinsurprenant/jruby-stdin-channel, I did some testing on it a few month ago. 

I think we just need to use it.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
andrewvc: Fixed! ^^

andrewvc:
Phew! I think I got through all the suggestions. We still need to discuss the proxying in the Snapshot class, but other than that looking good I think!

logstash-core/lib/logstash/pipeline.rb
ph: I would remove it, since its unused :(

logstash-core/lib/logstash/pipeline_reporter.rb
ph: @guyboertje @andrewvc I agree with your arguments, I still prefer to have a `OpenStruct` like structure than access the hash.. (@jordansissel knows how I like openstruct/struct :D)

ph:
I found an issue with this PR, its doesn't call the periodic flush on the filters.
If you use that config and send non matching data the filter will never flush them to the queues.

```
input {
  stdin {
    }
}

filter {
    multiline {
      pattern => "^2015"
      negate => true
      what => "previous" 
}
}

output {
    stdout { codec => rubydebug }
}
```

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: +1 on thinking about that. exception ca certainly happen in do_close.

logstash-core/lib/logstash/shutdown_controller.rb
colinsurprenant: did you  mean `to_s` ?

colinsurprenant:
@ph @jordansissel we even have a PR for this (interruptible stdin) logstash-plugins/logstash-input-stdin#3

logstash-core/lib/logstash/shutdown_controller.rb
ph: it depends on the behavior you want.. I have been using to_str in another project for some stuff. 
`to_str` tells the ruby language the the instance you are passing has a string like api.

```ruby
class Message
  def initialize(title)
    @title = title
  end

  def to_str
    @title
  end
 alias_method :to_s, :to_str
end

m = Message.new("world")

puts "Hello  #{m}"
puts "Hello " + m
```
Defining both method allow both of this workflow to work, but in that case `to_s` would be fine.

ph: `to_str` makes this work: 
```ruby
puts "Hello " + m
```

logstash-core/lib/logstash/pipeline_reporter.rb
```diff
@@ -0,0 +1,115 @@
...
+  # this provides a (more or less) consistent snapshot of what's going on in the
+  # pipeline with some extra decoration
+  def snapshot
+    Snapshot.new(self.to_hash)
+  end
+
+  def to_hash
+    pipeline.inflight_batches_synchronize do |batch_map|
+      worker_states_snap = worker_states(batch_map) # We only want to run this once
+      inflight_count = worker_states_snap.map {|s| s[:inflight_count] }.reduce(0, :+)
```
guyboertje: Is this not a case for defining the `inflight_count` method on the Snapshot class?
Its used internally in `to_simple_hash` and here.
If the internal hash in an instance of the `Snapshot` class is immutable you could easily memoize the inflight_count hash value in the method. moar speed!

logstash-core/lib/logstash/pipeline.rb
andrewvc: I agree that it should go in a future PR.

Thing is, we'll probably have to wrap and re-raise in the best case. If someone tries to shut down a pipeline and it raises an exception, they're kinda screwed because its now in a permanent bad state.

logstash-core/lib/logstash/pipeline_reporter.rb
andrewvc: I think there be a small mixup here @guyboertje. `to_simple_hash` is actually called on `Snapshot`, which uses the materialized value as proxied through its snapshotted hash. This line is in a different class, `PipelineReporter` which just builds the snapshot hash used by `Snapshot`.

andrewvc:
@ph and I paired on adding flushing back into the pipeline. Luckily it was a pretty simple thing to do :)

logstash-core/lib/logstash/shutdown_controller.rb
colinsurprenant: TIL `to_str` vs `String#+` . found this about it https://gist.github.com/kinopyo/5682347 .
+1 for `to_str` if it makes sense if we consider a report to be String-like.

ph: Ruby has a lot more of function like this. `to_path` `to_int` ...
For example `File.open` will call `to_path`

ph: I learned it from the same source `Confident Ruby` from avdi

colinsurprenant: hm. lesser known Ruby-isms. good to know.

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -56,34 +63,51 @@ def receive(event)
   def worker_setup
     if @workers == 1
       @worker_plugins = [self]
-      @worker_threads = []
     else
-      define_singleton_method(:handle, method(:handle_worker))
-      @worker_queue = SizedQueue.new(20)
+      define_singleton_method(:multi_handle, method(:handle_worker))
```
colinsurprenant: I am not sureI understand why this singleton_method is needed?

andrewvc: Because otherwise it would affect the workers. This is just redifining it in the parent.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,60 +113,170 @@ def filters?
...
+    batch.reduce(Hash.new { |h, k| h[k] = [] }) do |outputs_events, event|
+      # We ask the AST to tell us which outputs to send each event to
+      output_func(event).each do |output|
+        outputs_events[output] << event
+      end
+      outputs_events
+    end.each do |output, events|
+      # Once we have a mapping of outputs => [events] we can execute them
+      output.multi_handle(events)
+    end
```
colinsurprenant: this is very convoluted, I am wondering if it cannot be simplified somehow, for example to avoid calling `output_func` at each batch to create the outputs list to call `multi_handle` on? I am still struggling to wrap my head around this logic here so bare with me.

logstash-core/lib/logstash/outputs/base.rb
andrewvc: I'll just add that it'd be nice to not do this sort of hackery but rather just use ivars for changing this stuff up, however that runs the risk of being a performance regression and is IMHO out of scope for this PR.

logstash-core/lib/logstash/pipeline.rb
andrewvc: I'll clean it up with some intermediate variables to be less confusing.

Unfortunately short of a full rewrite of the AST this was the best solution I could come up with for getting this to work. The good news is its efficient, the bad news is its a little confusing.

logstash-core/lib/logstash/outputs/base.rb
colinsurprenant: but `multi_handle` would never be invoked in workers?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -127,27 +136,138 @@ def run
...
+      output_func(event).each do |output|
+        acc[output] << event
+      end
+      acc
+    end
+    # Now that we have our output to event mapping we can just invoke each output
+    # once with its list of events
+    outputs_events.each do |output, events|
+      output.multi_handle(events)
+    end
```
colinsurprenant: so previous conversation is now lost and my important question regarding the logic still stand. I will bump it up in the main comment thread to avoid loosing the conversation again for a code style update.

colinsurprenant:
I am still wrapping my head around the batching logic so bare with me, but I'd like to know if it would be possible to avoid the per-batch output selection in Pipeline#output_batch at https://github.com/andrewvc/logstash/blob/de66bb2243fbc863f5bd11b1652b2e9a7f5241ca/logstash-core/lib/logstash/pipeline.rb#L242-L257 ?

In particular the `output_func` is called at every batch, but doesn't that always return the same list?

andrewvc:
@colinsurprenant it returns a different list if the user has `if / else` blocks in their code, so its data-dependent.

andrewvc:
@colinsurprenant 

```
output {
  if [foo] == "bar" {
    stdout {}
  }
}
```

is what I'm referring to

logstash-core/lib/logstash/outputs/base.rb
colinsurprenant: oh, I just got it. there is already a `multi_handle` defined in the class which is the one used by the actual workers and the `define_singleton_method(:multi_handle, method(:handle_worker))` redefinition is only for the *top* worker which will pass events downs to the downstream output workers. 

I wonder how we could either name or in some way organize the code or document or whatever that would make it more obvious for someone in the code to understand this mechanic. even knowing about how it works it is really hard to understand. I believe we should take the opportunity of the current refactor to try and make that better?

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -56,34 +63,51 @@ def receive(event)
...
+  def multi_receive(events)
+    events.each {|event| receive(event) }
+  end
+
+  # Not to be overriden by plugin authors!
+  def multi_handle(events)
+    @single_worker_mutex.synchronize { multi_receive(events) }
+  end
+
+  def handle_worker(events)
```
colinsurprenant: I would try use a more descriptive name for this method. this is in reality the top level `multi_handle` when multiple workers are configured. maybe something like `multi_handle_proxy` or `multi_handle_multiplexer ` or something better

colinsurprenant:
@andrewvc Ah yes, the conditionals evaluation. Thanks that makes sense.

colinsurprenant:
A general concern about `outputs/base.rb` is that the original, let's call it the master/workers setup, for handling multiple output workers was already convoluted and hard to understand of what was going on and this is not improved in this evolution. I'd really like for the code to be easier to understand. I know we have to work with the contrains of backward compatibility with current plugins API but I'd really like to see something done to make that code more clear and easier to understand. this should include better documentation and comments in the code and maybe(?) some reorganization of the code?

andrewvc:
@colinsurprenant I agree that the code is confusing as all hell. The general issue with comprehension is the weird overriding of methods, compounded by the fact that the base class is a worker delegator plus a worker implementation. This could be fixed any number of ways, either with conditional statements or perhaps creating a new `WorkerDelegator` class that could be the 'parent' for all the children.

My personal feeling is that the outputs/base cleanup is work for a subsequent PR. If anything the new code simplifies outputs/base.rb by removing threading from it. I'm glad to tackle that work relatively soon, but I don't believe that it belongs in this PR which already has a large delta. I'm not inclined to make PRs too big.

colinsurprenant:
I think the WorkerDelegator class is a good idea. I don't think it is work for subsequent PRs. It is a major refactor, I don't understand why "doing it right" would not be part of it? We aim at improving quality, and proper code structure, readability and understandability is central to that.

It is really great work already and having reached the current state of making it work properly is an achievement by itself, but for something as critical as the pipeline we not only need something solid, we need something that everyone else will be able to understand and work with and possibly improve and fix bugs.

Don't get me wrong, I am not trying to deliberately delay merging, I just don't understand why we need to be in a rush to merge for something this critical. There are also workflow alternative like pushing a feature branch, closing this PR and reopening another one to reset discussions etc.

logstash-core/lib/logstash/pipeline_reporter.rb
guyboertje: Fair enough. I thought it was working with a snapshot instance not a hash.

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -56,34 +63,51 @@ def receive(event)
...
+
+      @worker_plugins.each do |wp|
+        wp.register
+        @available_workers << wp
       end
     end
   end
 
   public
+  # Not to be overriden by plugin authors!
```
jsvd: maybe I'm missing something but this method is no longer used.
I suggest removing it and renaming multi_handle to handle

logstash-core/lib/logstash/pipeline.rb
ph: Well, If we create a supervisor for the plugins we can deal with the error handling there depending on what we do on errors (send somewhere or ignore). This would simplify the pipeline and make it easier to test, I don't think one filter that crash on `do_close` should stop the other from closing.

ph:
I think other improvements should come in new PRs, I am comfortable to merge this and have smaller PRs to decouple the code evaluation part and the delegate of calls to the plugins. 

The complexity of this class and the flow of execution come from 2 problems for me:
- This class mutate, which make it hard to reason about.
- This class has similar named methods, (some with different invocation context).

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -323,61 +385,64 @@ def flush_filters(options = {}, &block)
...
+      @logger = logger
+    end
+
+    def update(time, result, exception)
+      return unless exception
+      @logger.warn("Error during flush!",
+                   :message => exception.message,
+                   :class => exception.class.name,
+                   :backtrace => exception.backtrace)
+    end
```
ph: @andrewvc Should we reset the `@flushing` to false on exception?

ph: The only reason of timeout here, we are stuck on a push for `X TIMEOUT_DELAY`, I have looked at the TimerTask options / code they don't permit to create a task without a timeout. 

So in the worst case is the timer is schedule to run again and try to push the flush to the queue?
Which is similar to be block on the queue forever right? WDYT?

logstash-core/lib/logstash/outputs/base.rb
andrewvc: I'm fine with removing this, but multi_handle is so named to be consistent with `Filter#multi_filter`. I'd rather not change `multi_handle` given that.

logstash-core/lib/logstash/outputs/base.rb
andrewvc: I think the real solution is to move multiplexing code out into its own module as discussed elsewhere. That way no more weird method redefinition.

logstash-core/lib/logstash/pipeline.rb
andrewvc: @ph I think just resetting `@flushing` will be correct.

ph: lets go with that.

andrewvc:
@colinsurprenant @ph @jsvd @suyograo per our discussion friday afternoon I have performed additional work out to extract multithreading capabilities out of outputs/base.rb and into a new class `OutputDelegator`.  Please let me know how it looks, and if possible, provide an LGTM.

ph:
@andrewvc Can you take a look at this https://github.com/elastic/logstash/commit/447292dec9ea7e1fda70b818f914bd98006d832e I believe this spec should pass with the changes we have made to the pipeline but its currently not the case.

If I uncomment the multiline block the events are send, I think the flushes dont occur correctly.

andrewvc:
I have updated this pr to include @ph 's fantastic new test for broken filter flushes, as well as to include pipeline reporter updates for the new `OutputDelegator`.

ph:
@andrewvc we will get somewhere, I think other people outside of us should test it with real configuration.

andrewvc:
@ph yep! I don't think merging this to master should wait on that however.

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -0,0 +1,66 @@
...
+# This class goes hand in hand with the pipeline to provide a pool of
+# free workers to be used by pipeline worker threads. The pool is
+# internally represented with a SizedQueue set the the size of the number
+# of 'workers' the output plugin is configured with.
+#
+# This plugin also records some basic statistics
+module LogStash; class OutputDelegator
+  attr_reader :workers, :config, :worker_count
+
+  def initialize(logger, klass, *args)
```
colinsurprenant: for readability and maintainability and especially when non-obvious, but ideally always, we should document method parameters. In particular here `*args` requires reading code below to understand the purpose.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -64,24 +83,25 @@ def configure(setting, value)
...
     if unsafe_filters.any?
       plugins = unsafe_filters.collect { |f| f.class.config_name }
       case thread_count
       when nil
         # user did not specify a worker thread count
         # warn if the default is multiple
-        @logger.warn("Defaulting filter worker threads to 1 because there are some filters that might not work with multiple worker threads",
-          :count_was => default, :filters => plugins) if default > 1
+        @logger.warn("Defaulting pipeline worker threads to 1 because there are some filters that might not work with multiple worker threads",
+                     :count_was => default, :filters => plugins) if default > 1
```
colinsurprenant: this is only a style comment but personally I think that if we are to break down parameters on multiple line, which I try to avoid, this multiline parameters construct is cleaner and flows better with the indentation.
```ruby
@logger.warn(
  "Defaulting pipeline worker threads to 1 because there are some filters that might not work with multiple worker threads", 
  :count_was => default, 
  :filters => plugins
) if default > 1
```
and if we consider that the `if` becomes non-obvious when placed last, we can always do
```ruby
default > 1 && @logger.warn(
  "Defaulting pipeline worker threads to 1 because there are some filters that might not work with multiple worker threads", 
  :count_was => default, 
  :filters => plugins
)
```

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,31 +114,21 @@ def filters?
...
-      flush_filters_to_output!(:final => true)
-    end
-
-    shutdown_outputs
-    wait_outputs
+    shutdown_flusher
+    shutdown_workers
 
     @logger.info("Pipeline shutdown complete.")
     @logger.terminal("Logstash shutdown completed")
```
colinsurprenant: TIL `@logger.terminal` ... not sure about the purpose vs info though ...

jordansissel: @colinsurprenant it was added kinda to help ensure folks who were using `--log` to log to a file, but still running Logstash in the terminal, would see this stuff. It's kinda a hack but helped solve that issue a bit.

andrewvc:
@colinsurprenant I've incorporated your feedback with my latest commit, how's it look now :)

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -127,27 +137,139 @@ def run
...
+
+  def take_batch(batch_size, batch_delay)
+    batch = []
+    # Since this is externally synchronized in `worker_look` wec can guarantee that the visibility of an insight batch
+    # guaranteed to be a full batch not a partial batch
+    set_current_thread_inflight_batch(batch)
+
+    signal = false
+    batch_size.times do |t|
+      event = t==0 ? @input_queue.take : @input_queue.poll(batch_delay)
```
colinsurprenant: cosmetic comment: `t == 0` ? sometime using the ternary form, grouping in parens makes it easier to read:
```ruby
event = (t == 0) ? ... : ...
```

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -127,27 +137,139 @@ def run
...
+        next
+      elsif event == LogStash::SHUTDOWN || event == LogStash::FLUSH
+        # We MUST break here. If a batch consumes two SHUTDOWN events
+        # then another worker may have its SHUTDOWN 'stolen', thus blocking
+        # the pipeline. We should stop doing work after flush as well.
+        signal = event
+        break
+      else
+        batch << event
+      end
```
colinsurprenant: another cosmetic/styling suggestion, when having if/elsif/else that are all breaking the flow and are at the beginning of a block, I think it's more readable and concise to have these breaking conditions one after the other like this:
```ruby
next if event.nil?

if event == LogStash::SHUTDOWN || event == LogStash::FLUSH
  signal = event
  break
end

batch << event
```

logstash-core/lib/logstash/pipeline.rb
andrewvc: I typically only use parens for more complex operations than a single boolean operation on a ternary operator. So, I politely disagree here.

logstash-core/lib/logstash/pipeline.rb
andrewvc: That was the original structure of this, but after a rewrite with @ph we wound up switching to this form for clarity.  Since only one of these three things could happen, and the operation of this stanza block of code is critical, we felt a little more verbosity in the name of being more explicit was a good thing.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -323,61 +393,70 @@ def flush_filters(options = {}, &block)
     end
   end
 
+  class FlusherObserver
+    def initialize(logger, flushing)
+      @logger = logger
+      @flushing = flushing
+    end
+
+    def update(time, result, exception)
```
colinsurprenant: it's not clear what the `FlusherObserver` class is meant for nor what the `update` method parameter relates to... maybe  add documentation?

logstash-core/lib/logstash/pipeline.rb
andrewvc: I'll add that I generally agree with you here @colinsurprenant , but in this case I wanted to underscore the importance of the SHUTDOWN and FLUSH events in terms of how they interrupt the control flow.

In this case the default case is less interesting than the edge cases.

logstash-core/lib/logstash/pipeline.rb
andrewvc: Well, this is part of the concurrent ruby package. I don't believe that we should document external libraries here. As to what its used for, it is clearly used in the function immediately following its definition. I don't think we need to add a comment explaining that given that fact.

As we transition toward concurrent ruby's timer task and away from stud buffer this will simply become more familiar IMHO, and thus shouldn't be documented any more than we'd document a specific idiom in stud.

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: ah I see, so it's to force terminal output event if logging is redirected. makes sense.

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: my comment was first about adding spaces `t == 0` instead of `t==0`

then my second comment is really just about readability specifically when using an expression (like `x == u` or something similar instead of a bool check like `x.nil?`)  in a ternary form with all the `=` `==` `?`  very close together makes it confusing when looking at it. using parens ease the parsing/comprehension process.

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: I still believe this is clearer than the if/elsif/else - but both are *correct* so it's up to you.

logstash-core/lib/logstash/pipeline.rb
andrewvc: You're right about the spaces, I'll go add them in. I'll just add the parens anyway since they aren't a  huge deal to me.

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: I disagree. I read the code and do not understand quickly the purpose so something is missing to help me make the mental link between this and concurrent ruby and timer task and what not. 

It might be obvious for you but you have to think about other people coming in the code and having to figure what this is about and what it does. Here it is not obvious to understand the purpose of the `FlusherObserver` class. 

As I said, maybe a few lines of comments would help both for the class and method.

colinsurprenant:
Generally speaking this is awesome work and the modifications for adding the `OutputDelegator` is really nice. I also really like the fact that we have more pipeline specs. 

I would like to know what the strategy is in terms of moving forward with merging this in master. Personally I did not do any real-world tests and I wonder if any stress-tests/real world tests have been done so far? 

I think we basically have 2 choices:
- we merge soon and hope for the best as we go and fix bugs as we encounter them? One of the problem I see for this is that since this is so central, it might be harder to find the cause of future bugs, is it because of a modification I introduced of is it because of the new pipeline?

- we take the time to do stress tests and integrations tests and when we reach a point of comfort we merge? 

Thoughts?

andrewvc:
@colinsurprenant great points, and sadly we don't have great answers. I've spent quite a while running various benchmarks with the pipeline and so far have not hit any stability problems. I don't think stress style tests are useful here, but throwing a variety of configs is.

Ideally we'd have a test suite that did macro style testing for LS, but since we don't I think only the first option is viable. 

I'd say that our goal should be to merge into master only and beware of future bugs that may arise.

I'll do a run with the cloud benchmarker this afternoon which is a bit more stressy, but I don't anticipate finding anything there given that it's somewhat similar to previous benchmarks I've run.

suyograo:
I am in favor of merging this to master *only* as-is. We can then backport to 2.x -- which is a release version -- when we feel we are ready. This will provide us the bake time it needs, CI time, and also LS developers start using this code in master as they work on stuff.

andrewvc:
@suyograo +1 for merging this now.

ph:
@suyograo +1 on your points.

purbon:
@suyograo fair points. +1

andrewvc:
@ph @purbon @suyograo are those LGTMS? :shipit: :sunglasses: :tada:

colinsurprenant:
On my side, I will take the required time to actually test it in *real life (tm)* locally with difference configs/plugins, etc, either this weekend or on Monday and present my final thoughts on it once I'm through that.

andrewvc:
Moved to https://github.com/elastic/logstash/pull/4340

PR #4261, opened 2015-11-24
cleanup Event class

colinsurprenant:
cleaned and streamlined the Event class. added a few comments. grouped deprecated methods. all core specs are passing, there should not be any code logic change including the few explicit `return` I removed which do not change logic.


jsvd:
LGTM

jordansissel:
LGTM

Issue #4253, opened 2015-11-23
strategy for implementation specific logging specs 


colinsurprenant:
The `logstash-core-event` `Event` specs contains tests like this to test for specific logging calls

```ruby
      it "should warn for invalid string format" do
        expect(LogStash::Event::LOGGER).to receive(:warn)
        LogStash::Event.new("@timestamp" => "foo")
      end
```

Obviously these tests don't pass the Java `Event` implementation because, first, logging is not yet implemented in the Java Event but these tests, as they are, will be difficult to support in any alternate implementation because they rely on an implementation detail which is the use of a specific method from a specific inner object constant. 

I am not sure what would be the proper approach to correctly test logging in a non implementation specific way.

An idea would be to use some  form of dependency injection by either passing a logger object to the Event constructor or use a class method to set a logger instance in an Event class variable. That way we could use a logging mock object in specs, or something around these lines. 

I will investigate a bit.

colinsurprenant:
implemented and fixed in #4263

colinsurprenant:
#4263 merged, closing.

Issue #4249, opened 2015-11-20
better string interpolation cache api and spec


colinsurprenant:
the fix I introduced in #4229 is actually too implementation specific and fails when using logstash-core-event-java. Obviously the LogStash::StringInterpolation class does not exist. 

```
 1) LogStash::Event#to_s should cache only one template
     Failure/Error: LogStash::StringInterpolation::CACHE.clear
     NameError:
       uninitialized constant LogStash::StringInterpolation
     # ./logstash-core-event/spec/logstash/event_spec.rb:509:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.8/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:54:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:240:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:235:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:179:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:172:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/task.rb:165:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:150:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:115:in `run_with_threads'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:100:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:78:in `run'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.4.2/lib/rake/application.rb:176:in `standard_exception_handling'
[...]

PR #4263, opened 2015-11-25
proper testable Ruby Event logging and Java Event logging support

colinsurprenant:
this PR builds on PR #4261 which cleans the Event class and adds the following:

- replace the Ruby `Event` `LOGGER` constant with an injectable logger object which allows correct testing as discussed in issue #4253 
- adds testable logging support in the Java `Event` and makes the Java `Event` pass all Ruby `Event` specs.

### DIFF
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -316,6 +316,10 @@ describe LogStash::Event do
     context "timestamp initialization" do
       let(:logger_mock) { double("logger") }
 
+      after(:each) do
+        LogStash::Event.logger = LogStash::Event::DEFAULT_LOGGER
+      end
+
       it "should coerce timestamp" do
         t = Time.iso8601("2014-06-12T00:12:17.114Z")
         expect(LogStash::Event.new("@timestamp" => t).timestamp.to_i).to eq(t.to_i)
@@ -346,9 +350,6 @@ describe LogStash::Event do
 
         LogStash::Event.new("@timestamp" => :foo)
         LogStash::Event.new("@timestamp" => 666)
-
-        # reset logger to default
-        LogStash::Event.logger = LogStash::Event::DEFAULT_LOGGER
       end
 
       it "should tag for invalid string format" do
@@ -363,9 +364,6 @@ describe LogStash::Event do
 
         expect(logger_mock).to receive(:warn)
         LogStash::Event.new("@timestamp" => "foo")
-
-        # reset logger to default
-        LogStash::Event.logger = LogStash::Event::DEFAULT_LOGGER
       end
     end
 


logstash-core-event-java/spec/event_spec.rb
```diff
@@ -135,4 +135,54 @@
       expect(event["tags"]).to eq(["foo"])
     end
   end
+
+  module DummyLogger
```
purbon: I would move this to a support location and rename it somehow including the word stub/mock. Moving this to a support location will help reusing this loggers in case of others need it.

logstash-core-event-java/spec/event_spec.rb
```diff
@@ -135,4 +135,54 @@
...
+    end
+  end
+
+  context "logger" do
+
+    let(:logger) { double("Logger") }
+    after(:each) {  LogStash::Event.logger = nil }
+
+    it "should set logger using a module" do
+      LogStash::Event.logger = DummyLogger
```
purbon: I don't understand what is the purpose of this test if the next one uses a rspec mock?

logstash-core-event-java/spec/event_spec.rb
```diff
@@ -135,4 +135,54 @@
...
+
+      # then unset
+      LogStash::Event.logger = nil
+      expect(logger).to receive(:warn).never
+      # this will produce a log line in stdout by the Java Event
+      LogStash::Event.new(TIMESTAMP => "ignore this log")
+    end
+
+
+    it "should warn on parsing errort" do
```
purbon: typo?

purbon:
I only see test for the invalid situations, can we add some test for valid outcomes? so we test both sides of the story.

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -354,8 +359,13 @@
       end
 
       it "should warn for invalid string format" do
-        expect(LogStash::Event::LOGGER).to receive(:warn)
+        logger_mock = double("logger")
```
purbon: I see some duplication in here? I know is not in the style of the spec, but we might introduce this now and make usage of let statements as you do for the core-event-java spec.

purbon:
I put some comments on specs, but in general the idea looks good!

logstash-core-event-java/spec/event_spec.rb
colinsurprenant: this is really specific to this test and the stage the implementation is in now, i.e.: only the warn method implemented. it should stay here.

logstash-core-event-java/spec/event_spec.rb
colinsurprenant: good point, I will add a bit of documentation, but the idea is that passing a real class/module vs a mock uses different objects and the JRuby ext implementation has to be compatible with both.

colinsurprenant:
@purbon for invalid/valid test: the specs in here are typically only for specific cases related to the Java & JRuby ext implementation that either were required while doing the implementation or are really specific for corner cases for only this implementation. Note that all logstash-core-event specs must and are passing this implementation

logstash-core-event-java/spec/event_spec.rb
```diff
@@ -135,4 +135,62 @@
...
+
+    it "should set logger using a module" do
+      LogStash::Event.logger = DummyLogger
+      expect(DummyLogger).to receive(:warn).once
+      LogStash::Event.new(TIMESTAMP => "invalid timestamp")
+    end
+
+    it "should set logger using a mock" do
+      LogStash::Event.logger = logger
+      expect(logger).to receive(:warn)
```
jsvd: since in line 160 you use .once, maybe do it here also?

jsvd:
Minor comment, otherwise LGTM.
Is there an issue to remove the technical debt of having two spec suites
https://github.com/colinsurprenant/logstash/blob/fix/event_logging/logstash-core-event/spec/logstash/event_spec.rb

colinsurprenant:
@jsvd it is a good point. some specs might be similar but the ones in the Java implementation were, for most part, added to validate this specific implementation. At some point it might be a good idea to review this and possibly merge/refactor them but this will be done at a later stage and we'll have to think about general specs vs contract specs, etc. I don't think having "more" specs is necessarily too harmful of tech debt so for me this is not too urgent for now.

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -340,10 +340,15 @@
       end
 
       it "should warn for invalid value" do
-        expect(LogStash::Event::LOGGER).to receive(:warn).twice
+        LogStash::Event.logger = logger_mock
```
jordansissel: Feels like this should be done in a `before` and the resetting `...logger = nil` should be in an `after` ?

colinsurprenant: +1, it feels like that :D

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -28,9 +28,9 @@
     public static final String VERSION = "@version";
     public static final String VERSION_ONE = "1";
 
-    private static final ObjectMapper mapper = new ObjectMapper();
-
-    // TODO: add metadata support
+    private static final Logger DEFAULT_LOGGER = new StdioLogger();
+    private transient final ObjectMapper mapper = new ObjectMapper();
+    private transient static Logger logger = DEFAULT_LOGGER;
```
talevy: if this is being set at runtime, why should this be `static`?

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -274,5 +286,12 @@ public IRubyObject ruby_tag(ThreadContext context, RubyString value)
         public IRubyObject ruby_timestamp(ThreadContext context) throws IOException {
             return new JrubyTimestampExtLibrary.RubyTimestamp(context.getRuntime(), this.event.getTimestamp());
         }
+
+        @JRubyMethod(name = "logger=", required = 1, meta = true)
+        public static IRubyObject ruby_set_logger(ThreadContext context, IRubyObject recv, IRubyObject value)
+        {
+            Event.setLogger(value.isNil() ? null : new ProxyLogger((RubyObject)value));
```
talevy: should we fail this if it is `null`? I understand that the current behavior internally is to default to the default logger (stdio), but technically this operation failed since it could not set a new logger.

logstash-core-event-java/src/main/java/com/logstash/Event.java
colinsurprenant: because its a *class* variable, the same logger for all Event instances.

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: good question. not sure, will think about it.

logstash-core-event-java/spec/event_spec.rb
```diff
@@ -135,4 +135,62 @@
...
+
+    let(:logger) { double("Logger") }
+    after(:each) {  LogStash::Event.logger = nil }
+
+    # the following 2 specs are using both a real module (DummyLogger)
+    # and a mock. both tests are needed to make sure the implementation
+    # supports both types of objects.
+
+    it "should set logger using a module" do
+      LogStash::Event.logger = DummyLogger
```
jordansissel: Why not use the double here?

jordansissel: I suppose my question is more that - why test both module and mock?

logstash-core-event-java/src/main/java/com/logstash/Event.java
jordansissel: Should we use a singleton or factory pattern here instead of a global variable? Feels weird to have a mutable global variable.

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -225,18 +225,16 @@ private Timestamp initTimestamp(Object o) {
             } else if (o instanceof RubySymbol) {
                 return new Timestamp(((RubySymbol) o).asJavaString());
             } else {
-                // TODO: add logging
-                //return Timestamp.now();
-                throw new IllegalArgumentException();
+                Event.logger.warn("Unrecognized " + TIMESTAMP + " value type=" + o.getClass().toString());
```
jordansissel: Can we use structured logging instead of string formatting?

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -250,4 +248,8 @@ public void tag(String tag) {
             tags.add(tag);
         }
     }
+
+    public static void setLogger(Logger logger) {
+        Event.logger = (logger == null) ? DEFAULT_LOGGER : logger;
```
jordansissel: Should we allow null? If DEFAULT_LOGGER is a constant, what if instead there's a setLogger(DEFAULT_LOGGER) or resetLoggerToDefault() ?

logstash-core-event-java/src/main/java/com/logstash/Logger.java
```diff
@@ -0,0 +1,13 @@
...
+// minimalist Logger interface to wire a logger callback in the Event class
+// for now only warn is defined because this is the only method that's required
+// in the Event class.
+// TODO: (colin) generalize this
+
+public interface Logger {
+
+    // TODO: (colin) complete interface beyond warn when needed
+
+    void warn(String message);
```
jordansissel: Can we do structured logging similar to Cabin or log4j's MDC?

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -1,15 +1,10 @@
 package com.logstash.ext;
 
-import com.logstash.Event;
-import com.logstash.PathCache;
-import com.logstash.RubyToJavaConverter;
-import com.logstash.Timestamp;
+import com.logstash.*;
```
jordansissel: Why use wildcard imports? Explicit imports, for me at least, make it easier to read.

logstash-core-event-java/spec/event_spec.rb
colinsurprenant: as written in the comment above the test, this is needed to avoid regression basically. That is because of the fast that this is going through the JRuby Java Ext wrapper which at first did not work with the mock/double. so I had to adjust the code to be more permissive on the type of object it would accept.

maybe the comment is not clear or detailed enough for wht "why"...?

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
jordansissel: I am also in favor of avoiding nil/null where possible.

logstash-core-event/lib/logstash/event.rb
```diff
@@ -245,16 +243,24 @@ def fields
     raise DeprecatedMethod
   end
 
+  def self.logger=(logger)
+    @@logger = logger
```
jordansissel: Are class variables in JRuby thread-safe?

logstash-core-event-java/src/main/java/com/logstash/Event.java
colinsurprenant: that's in line with @talevy comment above for the same thing but in the JRubu wrapper. I need to think about it.

logstash-core-event-java/src/main/java/com/logstash/Logger.java
colinsurprenant: we can, this is the minimal implementation to support what is currently used in the Event. I did not want to go through a complete logging infrastructure work at this point, but basically this is a passthrough/proxy to whatever logging strategy is used in Ruby.

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: agree. this is IntelliJ reorganizing the imports...

logstash-core-event-java/src/main/java/com/logstash/Event.java
colinsurprenant: I am not sure. at this point this is only mutated in the tests/specs. there is not really *currently* any point in changing the logger at runtime. for now this is what make it work with the less overhead, this is certainly something we can improve.

logstash-core-event-java/src/main/java/com/logstash/Event.java
colinsurprenant: yes, I guess we could. I would prefer to create a new PR for this though.

logstash-core-event-java/src/main/java/com/logstash/Logger.java
colinsurprenant: I see your point now related to the other comment above (I was going backward in the comment :P). So yes, I agree, but would prefer to push back that improvement in another PR.

colinsurprenant:
after this first pass of reviews I realize I probably should have done 2 separate PRs, one for the current Ruby stuff and the other for the Java Event because both have different expectations for the reviews:
- the Ruby code need to be reviewed the "normal" way for merging into master

- the Java Event code is still not production quality and does not have the same quality expectations. this is code to get me going to an initial/base version that will pass **all** specs. this code **is** merged in master but is not in any *live* execution path and will not be included in the release package until we have the ready stamp on it. I suggest that any *bigger improvements* be noted and created as issues so I/we can work on them at a later stage.

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
talevy: it is possible to configure intellij to not do this. I think you have to set a threshold class count before wildcarding (and just set it to something crazy).

logstash-core-event/lib/logstash/event.rb
colinsurprenant: I am not sure that thread safety is a *real* problem here:
- `@@logger` is only assigned once normally. We currently do not reassign loggers at runtime.
- that being said, `@@logger` could probably be initialized sooner instead of in the the first `#Event#logger` call
- in JRuby defining or modifying constants and class variables is thread safe

Are there any other concerns I missed?

logstash-core-event-java/spec/event_spec.rb
jordansissel: your comment is clear, I didn't read it somehow. haha - thanks! <3

logstash-core-event-java/src/main/java/com/logstash/Event.java
jordansissel: I'm OK with a global variable here if we comment that it's intended for testing/specs.

logstash-core-event-java/src/main/java/com/logstash/Event.java
jordansissel: +1 on waiting for a new PR for log structuring.

logstash-core-event-java/src/main/java/com/logstash/Logger.java
jordansissel: +1 on your proposal to add structured logging in another PR, not this one.

logstash-core-event/lib/logstash/event.rb
jordansissel: OK let's leave this as-is. I am satisfied ;P

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: yeah, need to do this :P thanks for the heads up

logstash-core-event/lib/logstash/event.rb
colinsurprenant: will add a comment per this discussion though

colinsurprenant:
@ph just a ping since I know you are working on refactoring logging stuff.

logstash-core-event-java/spec/event_spec.rb
colinsurprenant: +1

logstash-core-event-java/src/main/java/com/logstash/Event.java
colinsurprenant: +1 adding comments

logstash-core-event-java/src/main/java/com/logstash/Event.java
colinsurprenant: added issue #4249

logstash-core-event-java/src/main/java/com/logstash/Event.java
colinsurprenant: will expose DEFAULT_LOGGER and remove nil/null support

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: will expose DEFAULT_LOGGER and remove nil/null support

logstash-core-event/spec/logstash/event_spec.rb
colinsurprenant: fixed

colinsurprenant:
@jsvd @talevy @jordansissel fixes following your comments, would like another round of review! Thanks!

guyboertje:
It's a pity that I was not pushy enough on this RSpec matcher for Cabin.  https://github.com/jordansissel/ruby-cabin/pull/36

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -1,47 +1,72 @@
...
                 return new RubyEvent(runtime, rubyClass);
             }
         }, module);
-        clazz.setConstant("LOGGER", runtime.getModule("Cabin").getClass("Channel")
-                .callMethod("get", runtime.getModule("LogStash")));
+
         clazz.setConstant("TIMESTAMP", runtime.newString(Event.TIMESTAMP));
         clazz.setConstant("TIMESTAMP_FAILURE_TAG", runtime.newString(Event.TIMESTAMP_FAILURE_TAG));
         clazz.setConstant("TIMESTAMP_FAILURE_FIELD", runtime.newString(Event.TIMESTAMP_FAILURE_FIELD));
+        clazz.setConstant("DEFAULT_LOGGER", runtime.getModule("Cabin").getClass("Channel").callMethod("get", runtime.getModule("LogStash")));
```
jordansissel: +1 this, good idea.

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -314,7 +314,11 @@
     end
 
     context "timestamp initialization" do
-      let(:logger) { double("logger") }
+      let(:logger_mock) { double("logger") }
+
+      after(:each) do
+        LogStash::Event.logger = LogStash::Event::DEFAULT_LOGGER
```
jordansissel: ++

jordansissel:
Second pass review - LGTM

colinsurprenant:
thanks, will merge!

colinsurprenant:
merged into master and 2.x

Issue #4264, opened 2015-11-25
[meta] plugins specs fixes tracking for logstash-core-event-java


colinsurprenant:
As I work on having all the plugins specs pass using logstash-core-event-java I will update this tracking list of the related issues/PRs. relates to #4191

- [x] filter-grok in logstash-plugins/logstash-filter-grok#62 and #4265
- [x] input-sqs in #4278
- [x] codec-edn #4279
- [x] filter-geoip logstash-plugins/logstash-filter-geoip#56
- [x] filter-kv  logstash-plugins/logstash-filter-kv#20 ogstash-plugins/logstash-filter-kv#21
- [x] filter-mutate logstash-plugins/logstash-filter-mutate#59 logstash-plugins/logstash-filter-mutate#61 #4279
- [x] filter-json logstash-plugins/logstash-filter-json#18
- [x] filter-xml logstash-plugins/logstash-filter-xml#21
- [x] filter-ruby logstash-plugins/logstash-filter-ruby#15
- [x] filter-csv logstash-plugins/logstash-filter-csv#20
- [x] output-file logstash-plugins/logstash-output-file#23
- [x] filter-useragent logstash-plugins/logstash-filter-useragent#20
- [x] output-graphite logstash-plugins/logstash-output-graphite#20
- [x] codec-json_lines logstash-plugins/logstash-codec-json_lines#12
- [x] input-elasticsearch logstash-plugins/logstash-input-elasticsearch#35
- [x] codec-netflow logstash-plugins/logstash-codec-netflow#29
- [x] filter-date logstash-plugins/logstash-filter-date#45
- [x] input-imap logstash-plugins/logstash-input-imap#11

and also but not specific to Java Event

- [x] codec-rubydebug see logstash-plugins/logstash-codec-rubydebug#8

jsvd:
Can you also open issues for the plugins that are failing but don't have a PR yet?

colinsurprenant:
as of now, these are the remaining failures:

```
Finished in 1 minute 37.3 seconds (files took 9.64 seconds to load)
889 examples, 14 failures

Failed examples:

rspec ./vendor/bundle/jruby/1.9/gems/logstash-filter-useragent-2.0.3/spec/filters/useragent_spec.rb:101 # LogStash::Filters::UserAgent Replace source with target "Mozilla/5.0 (X11; Linux i686) AppleWebKit/537.31 (K..." when processed
[...]

Issue #4191, opened 2015-11-13
core Event class Java rewrite


colinsurprenant:
### Overview

Rewrite the `Event` class in pure Java and its supporting classes like `Timestamp` and `Accessors` and keep 100% compatibility with current implementation of other core logstash components and all plugins. 

### Purpose

`Event`is the main object which encapsulates data and provides and API for the plugins to perform processing on the event content. Having `Event` implemented in pure Java will improve performance, make possible **faster serialization** by avoiding costly type conversion between JRuby and Java which in turn will help with an **efficient persistence** implementation.

Also, these critical and costly operations are related to the `Event` where the Java implementation should provide performance improvement:
- **fields reference getters and setters**: All `Event` data manipulation in the config is done using the logstash field reference syntax (ex. `[foo][bar][baz]`). This is implemented in the `Accessors` class. an optimized implementation should improve performance.
- **string interpolation**: (`Event#sprintf()`) The config syntax provides string interpolation in field references. The Ruby implementation was known to be slow and the Java implementation should provides a noticeable improvement. String interpolation is heavily used in all configs.

### Strategy

- [x] Extract the current `Event` implementation from logstash-core and make it a proper self-contained gem which should be drop-in replaceable with an alternate implementation. This is completed in #4123 and the gem is `logstash-core-event`.
- [x] Create the Java implementation gem logstash-core-event-java. This is in progress in #4123. 
- [x] provide benchmarks
- [x] have all core specs pass
- [x] have all core plugins spec pass
- [x] have all default plugins spec pass

colinsurprenant:
All default plugins are now passing specs, all related PRs are tracked in #4264. Merge fest can now start.

colinsurprenant:
quick benchmark update, I tested the config defined in https://gist.github.com/colinsurprenant/8143a0dfd30830d57b03 using 4 workers `-w 4` and default batch size of 125 (not specifying `-b`) on master pre-g-pipeline, ng-pipeline and ng-pipeline+java-event:

TPS|pipeline
--- | --- 
4.86k|pre ng-pipeline
6.29k|ng-pipeline
7.50k|ng-pipeline+java-event

untergeek:
That's an impressive increase!

acchen97:
[...]

PR #4265, opened 2015-11-25
fix nil value field reference assignment in Java Event

colinsurprenant:
this PR builds on top of #4263 

and also add related missing spec in Ruby Event
this relates to #4264 and #4191 and is required to have the filter grok specs pass with the Java Event.

### DIFF
--- a/logstash-core-event-java/spec/event_spec.rb
+++ b/logstash-core-event-java/spec/event_spec.rb
@@ -90,6 +90,12 @@ describe LogStash::Event do
       expect(e["[foo][2]"]).to eq(1.0)
       expect(e["[foo][3]"]).to be_nil
     end
+
+    it "should add key when setting nil value" do
+      e = LogStash::Event.new()
+      e["[foo]"] = nil
+      expect(e.to_hash).to include("foo" => nil)
+    end
   end
 
   context "timestamp" do
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -44,6 +44,11 @@ describe LogStash::Event do
         subject["@metadata"] = { "action" => "index" }
         expect(subject["[@metadata][action]"]).to eq("index")
       end
+
+      it "should add key when setting nil value" do
+        subject["[baz]"] = nil
+        expect(subject.to_hash).to include("baz" => nil)
+      end
     end
 
     context "#sprintf" do
--- a/logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
+++ b/logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
@@ -155,6 +155,8 @@ public class JrubyEventExtLibrary implements Library {
                     this.event.setField(r, RubyToJavaConverter.convertToList((RubyArray) value));
                 } else if (value instanceof RubyHash) {
                     this.event.setField(r, RubyToJavaConverter.convertToMap((RubyHash) value));
+                } else if (value.isNil()) {
+                    this.event.setField(r, null);
                 } else {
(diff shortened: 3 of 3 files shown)


logstash-core-event-java/spec/event_spec.rb
```diff
@@ -90,6 +90,12 @@
       expect(e["[foo][2]"]).to eq(1.0)
       expect(e["[foo][3]"]).to be_nil
     end
+
+    it "should set nil value" do
```
jsvd: suggestion: it "should add a key even if the value is nil"

or
```ruby
context "when the value is nil" do
  it "should create the key" do
  end
end

**Comment by jsvd** [comment]
minor comment about test phrasing, otherwise LGTM

File: logstash-core-event-java/spec/event_spec.rb (hunk 1 above)
Comment (purbon) [inline logstash-core-event-java/spec/event_spec.rb]: +1 on the wording fix, looks to me more clear like joao propose.

File: logstash-core-event-java/spec/event_spec.rb [hunk 2]
```diff
@@ -90,6 +90,12 @@
...
+      e = LogStash::Event.new()
+      e["[foo]"] = nil
+      expect(e.to_hash).to include("foo")
```
purbon: proposal, what about having the condition like ```expect(e.to_hash).to include("foo" => nil)``` ? this will check the existence of the pair in the hole, that the key exist and the value is nil.

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -44,6 +44,11 @@
         subject["@metadata"] = { "action" => "index" }
         expect(subject["[@metadata][action]"]).to eq("index")
       end
+
+      it "should set nil value" do
+        subject["[baz]"] = nil
+        expect(subject.to_hash).to include("baz")
```
purbon: also https://github.com/elastic/logstash/pull/4265/files#r46529798 applied here.

purbon:
make small proposal on rspec matcher syntax usage, other than that looks good.

logstash-core-event-java/spec/event_spec.rb
colinsurprenant: +1

logstash-core-event-java/spec/event_spec.rb
colinsurprenant: +1

logstash-core-event/spec/logstash/event_spec.rb
colinsurprenant: +1

colinsurprenant:
@jsvd @purbon good comments, addressed, you can followup! thanks!

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -44,6 +44,11 @@
         subject["@metadata"] = { "action" => "index" }
         expect(subject["[@metadata][action]"]).to eq("index")
       end
+
+      it "should add key when setting nil value" do
```
jordansissel: Mild curiosity - both core-event-java and core-event have this test, but the implementation is not the same, though it's really close. Should this use a shared example (sometime soon in a later PR?)

jordansissel:
LGTM

logstash-core-event/spec/logstash/event_spec.rb
colinsurprenant: yeah, I will do specs reunification pass at some point. actually, I'll create an issue for it (did I create one for this already? memory #fail) - there is a bit of duplication and what is really generic should be in core-event and java specific only in core-event-java.

colinsurprenant:
ok, thanks, will merge!

colinsurprenant:
merged in master and 2.x

PR #4269, opened 2015-11-25
change version to 2.2.0.dev

jsvd:


purbon:
LGTM

suyograo:
@jsvd We should remove the `dev` part too and just point it to next 2.x release, which is 2.2

jsvd:
@suyograo to be consistent, 2.x branch should be always be 2.x.0.dev, and 2.2 (when we create it) should be 2.2.0

suyograo:
Closing this, as this needs to be 2.3 now

PR #4278, opened 2015-11-25
fix missing methods in Java Timestamp

colinsurprenant:
this PR builds on top of #4265

- add missing `utc`, `gmtime` and `to_f` methods in the Java `Timestamp`
- add missing specs for these methods in the Ruby `Timestamp` specs to prevent future regressions

this relates to #4264 and #4191 and fixes sqs input specs under Java `Event`.

### DIFF
--- a/logstash-core-event/spec/logstash/timestamp_spec.rb
+++ b/logstash-core-event/spec/logstash/timestamp_spec.rb
@@ -81,4 +81,29 @@ describe LogStash::Timestamp do
       expect(t - 10).to eq(current)
     end
   end
+
+  context "identity methods" do
+    subject { LogStash::Timestamp.new }
+
+    it "should support utc" do
+      expect(subject.utc).to eq(subject)
+    end
+
+    it "should support gmtime" do
+      expect(subject.gmtime).to eq(subject)
+    end
+  end
+
+  context "numeric casting methods" do
+    let (:now) {Time.now}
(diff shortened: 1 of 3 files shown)


logstash-core-event/spec/logstash/timestamp_spec.rb
```diff
@@ -81,4 +81,29 @@
       expect(t - 10).to eq(current)
     end
   end
+
+  context "identity methods" do
+    subject { LogStash::Timestamp.new }
+
+    it "should support utc" do
+      expect(subject.utc).to eq(subject)
```
purbon: Is this really checking that the returned method is himself? I understand why you need this test, but find the condition interesting, might be there is another way in rspec to explain this check better.

purbon: As I understand the ruby API, calling UTC converters this value, right? I guess we have test for this already, if so I would recommend grouping this test together for clear understanding.

colinsurprenant: The Timestamp contract is that any time will be converted in UTC so the `utc` and `gmtime` methods are just identity function because a `Timestamp` is always UTC.

jordansissel:
LGTM

colinsurprenant:
merged in master and 2.x

PR #4279, opened 2015-11-26
proper java to ruby convertion and specs

colinsurprenant:
this PR builds on top of #4278 

- added the `Rubyfier` class to deep convert Java to Ruby event data. 
- added specs 

this relates to #4264 and #4191 and fixes edn codec specs under Java Event.


logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -235,26 +229,21 @@ public IRubyObject ruby_to_s(ThreadContext context)
...
         {
-            HashMap<String, Object> dataAndMetadata = new HashMap<String, Object>(this.event.getData());
+            Map data;
+
             if (!this.event.getMetadata().isEmpty()) {
-                dataAndMetadata.put(Event.METADATA, this.event.getMetadata());
+                data = new HashMap<>(this.event.toMap());
+                data.put(Event.METADATA, this.event.getMetadata());
+            } else {
+                data = this.event.toMap();
```
guyboertje: I might be mistaken, but could this else clause be eliminated by, in line 238, setting data to `this.event.toMap()` and removing line 241?

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -140,6 +132,8 @@ public IRubyObject ruby_set_field(ThreadContext context, RubyString reference, I
                     this.event.setField(r, RubyToJavaConverter.convertToList((RubyArray) value));
                 } else if (value instanceof RubyHash) {
                     this.event.setField(r, RubyToJavaConverter.convertToMap((RubyHash) value));
+                } else if (value instanceof RubyBoolean) {
+                    this.event.setField(r, ((RubyBoolean) value).toJava(boolean.class));
```
guyboertje: Do you need to store objects or are java primitives OK? If object, maybe use `Boolean.class`, otherwise you could use
`((RubyBoolean value).isTrue()`

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -140,6 +132,8 @@ public IRubyObject ruby_set_field(ThreadContext context, RubyString reference, I
...
                 } else if (value.isNil()) {
                     this.event.setField(r, null);
                 } else {
```
guyboertje: Consider whether you need to provide for: RubySymbol, RubyBignum and RubyBigDecimal

guyboertje: On line 124 consider RubyFixnum (base class) instead of RubyInteger.

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: absolutely

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: I believe primitives are ok. I like your `is_True()` proposal.

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: yeah, so what value type would you extract that to? for RubyInteger we use `.getLongValue())` for RubyFloat `getDoubleValue()` ... ?

Unless I missed something, I think we need to do each type independently.  

There shouldn't be a case where a Ruby symbol is in the Event data, if it is, its a bug no? but instead of crashing we could transform them to string maybe?

RubyBigDecimal does not exists?

I wonder if we shouldn't just do a Numeric catch all with:
```ruby
if (value instanceof RubyNumeric) {
  this.event.setField(r, ((RubyNumeric) value).toJava())
}
```

colinsurprenant: ah no,  need to set the target class in `toJava(...)` ...

colinsurprenant: actaully, `RubyBignum` and `RubyFixnum` are subclasses of `RubyInteger`, so I believe that 
```ruby
                } else if (value instanceof RubyInteger {
                    this.event.setField(r, ((RubyInteger) value).getLongValue();
```
is the right strategy.

colinsurprenant: I will add BigDecimal conversion.

colinsurprenant:
@guyboertje waiting on your final feedback, when ok I will merge, after which I will reboot a shared feature branch with the java code.

guyboertje:
LGTM

colinsurprenant:
@guyboertje added a last commit to rework the Ruby to Java type conversion. Since this is all Java stuff it's safe to merge also considering I will reboot a new shared branch will all this. Nonetheless I'd like to have your final blessing (or comments :)

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -233,11 +215,7 @@ public IRubyObject ruby_sprintf(ThreadContext context, IRubyObject format) throw
             try {
                 return RubyString.newString(context.runtime, event.sprintf(format.toString()));
             } catch (IOException e) {
-                throw new RaiseException(
-                    getRuntime(),
-                    (RubyClass) getRuntime().getModule("LogStash").getClass("Error"),
-                    "timestamp field is missing", true
-                );
+                throw new RaiseException(getRuntime(), (RubyClass)getRuntime().getModule("LogStash").getClass("Error"), "timestamp field is missing", true);
```
guyboertje: Consider making a LogStashError utility class for all others to use. Along the lines of:
```
import org.jruby.Ruby;
import org.jruby.RubyClass;
import org.jruby.RubyException;
import org.jruby.anno.JRubyClass;
import org.jruby.exceptions.RaiseException;

@JRubyClass(name = "JrJackson::ParseError", parent = "RuntimeError")
public class ParseError {

    public static RaiseException newParseError(Ruby ruby, String message) {
        RubyClass errorClass = ruby.getModule("JrJackson").getClass("ParseError");
        return new RaiseException(RubyException.newException(ruby, errorClass, message), true);
    }
}
```

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
```diff
@@ -0,0 +1,147 @@
...
+            case True: return deep((RubyBoolean.True)o);
+            case False: return deep((RubyBoolean.False)o);
+        }
+
+        // TODO: (colin) temporary trace to spot any unhandled types
+        System.out.println("***** WARN: UNHANDLED IRubyObject full class name=" + o.getClass().getName() + ", simple name=" + o.getClass().getSimpleName() + " java class=" + o.getJavaClass().toString() + " toString=" + o.toString());
+
+        return o.toJava(o.getJavaClass());
+    }
+
```
guyboertje: @colinsurprenant - Just remembered about issue with JrJackson and the twitter input - NullObjects built on top of BasicObject.  They should behave like `o.nil?` is true.

Implications
- consider the case of `o.isNil()` -> true
- consider using o.getMetaClass().getRealClass().getName() -> this returns the Ruby side class name e.g. `MyModule::MyCustomClass` instead of `RubyObject`, especially for your trace println

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: +1 agree

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
colinsurprenant: good points, thanks.

colinsurprenant: yeah, excellent tip on the class name, now  it gives
```
***** WARN: UNHANDLED IRubyObject full class name=Foobar, simple name=RubyObject java class=class org.jruby.RubyObject toString=#<Foobar:0x2b8b2792>
```

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: I will actually create a followup issue to review all exception handling in the Java code

colinsurprenant:
I will go ahead a merge this into master and we can figure the shared branch rebooting.

jordansissel:
@colinsurprenant +1

colinsurprenant:
merged in master and 2.x

Issue #3152, opened 2015-05-04
lock file prevents upgrade of plugins that are needed by others


jsvd:
in a release the .lock file will contain the reference to a plugin's locked version after being uninstalled if other plugins depend on it. Example:

```
/tmp % tar -zxf logstash-2.0.0.dev.tar.gz
/tmp % cd logstash-2.0.0.dev
/tmp/logstash-2.0.0.dev % grep "logstash-codec-json .*0" Gemfile.jruby-1.9.lock
    logstash-codec-json (0.1.7)
/tmp/logstash-2.0.0.dev % bin/plugin uninstall logstash-codec-json
Uninstalling logstash-codec-json
/private/tmp/logstash-2.0.0.dev/vendor/local_gems/fccdd8d6/logstash-core-2.0.0.dev-java/lib/logstash/version.rb:3 warning: already initialized constant LOGSTASH_VERSION
/tmp/logstash-2.0.0.dev % grep "logstash-codec-json .*0" Gemfile.jruby-1.9.lock
    logstash-codec-json (0.1.7)
/tmp/logstash-2.0.0.dev % bin/plugin install --version="0.1.6" logstash-codec-json
Validating logstash-codec-json-0.1.6
Installing logstash-codec-json
Plugin not found, aborting
ERROR: Installation Aborted, message: You have requested:
  logstash-codec-json = 0.1.6

The bundle currently has logstash-codec-json locked at 0.1.7.
Try running `bundle update logstash-codec-json`
/tmp/logstash-2.0.0.dev %
```

jsvd:
Also a simple uninstall of a plugin that others depend on "fails":

```
% bin/plugin list tcp
logstash-input-tcp
logstash-output-tcp
% bin/plugin uninstall logstash-input-tcp
Uninstalling logstash-input-tcp
% bin/plugin list tcp
logstash-input-tcp
logstash-output-tcp
% bin/plugin uninstall logstash-input-tcp
ERROR: Uninstall Aborted, message: This plugin has not been previously installed, aborting
```

purbon:
I confirm this issue.

jsvd:
The uninstalled plugin remains in vendor/bundle:

```
 % find . -name *logstash-input-tcp* 
./vendor/bundle/jruby/1.9/gems/logstash-input-tcp-0.1.5
./vendor/bundle/jruby/1.9/gems/logstash-input-tcp-0.1.5/logstash-input-tcp.gemspec
[...]

PR #4288, opened 2015-11-30
Provide feedback to users when trying to uninstall plugins that are dependencies of others.

purbon:
Add a way to detect dependencies for plugins when uninstalling, this provides feedback to users when hitting #3152. It provides this by building an initial version of a dependency graph from the Gemfile.lock in a similar way and strategy as we deal with the Gemfile.

**_Some notes**_

The dependency graph will let us do a lot of more things, but for now the implementation is general enough, in some way, and simple enough, in another to fit this use case. Keep this in mind when reviewing.

This add a warning to the user when doing uninstall, I chose this path as the less intrusive one from the many options possible.

Let me know what do you think,

### DIFF
--- a/integration/plugin_manager/logstash_spec.rb
+++ b/integration/plugin_manager/logstash_spec.rb
@@ -1,10 +1,10 @@
 # Encoding: utf-8
 require_relative "../spec_helper"
-require_relative "../../lib/logstash/version"
+require_relative "../../logstash-core/lib/logstash/version"
 
 describe "bin/logstash" do
   it "returns the logstash version" do
-    result = command("bin/logstash version")
+    result = command("bin/logstash --version")
     expect(result.exit_status).to eq(0)
     expect(result.stdout).to match(/^logstash\s#{LOGSTASH_VERSION}/)
   end
--- a/integration/plugin_manager/plugin_uninstall_spec.rb
+++ b/integration/plugin_manager/plugin_uninstall_spec.rb
@@ -20,5 +20,12 @@ describe "bin/plugin uninstall" do
       expect(result.stdout).to match(/^Uninstalling logstash-filter-ruby/)
       expect(result.exit_status).to eq(0)
     end
+
+    it "fails if has dependencies" do
+      result  = command("bin/plugin uninstall logstash-input-tcp")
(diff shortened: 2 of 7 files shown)


purbon:
@ph feedback welcome. @colinsurprenant please go ahead as well.

ph:
Before I am reviewing this, I just want to point that the index format contract is subject to change, a lot more than the `Gemfile` format.

ph:
I have read the code and I have a bit of concerns, this code is well done but still complex and subject to break if the underlying structure change which will happen when https://github.com/bundler/bundler/pull/3556 land or any related PR.

Have you investigated if you were able to get the same information you are fetching with the existing bundler methods?

colinsurprenant:
I do not understand what this solves exactly and how it solves it and why this is needed. #3152 has a long comment thread and its not clear what is or not a bug, what should be fixed or improved. 

I am particularly scared by a re-implementation of the `.lock` file dependency graph and why we need this?

ph:
@colinsurprenant I *believe*, @purbon correct me if I am wrong, the goal of this PR is to have an infrastructure in place to give better feedback on errors that can occur from the dependency graph. 

`Plugin A depends on plugins B and I try to uninstall plugin B`

I believe this implementation is fragile and we should investigate other ways first.

purbon:
@ph I would like to know why __I believe this implementation is fragile__, please let me know your concerns so I can fix them.

@colinsurprenant yes, the idea is to provide, and might be in the feature action when such dependency conflicts happen.

purbon:
@ph if ___underlying structure change___ I agree this method of fetching the dependency information, by parsing the ```.lock``` file is going to fail. I would love bundler to provide methods for us to get this information and I can investigate if possible. Do you know where to start looking into bundler?

segiddins:
Why not use bundler's existing lockfile parser and SpecSet classes?

purbon:
@segiddins this makes sense, I must admit was not able to find it, can you please point it to me? this would be super nice of you.

purbon:
@segiddins do you mean https://github.com/bundler/bundler/blob/master/lib/bundler/lockfile_parser.rb ?

segiddins:
Yes

purbon:
@segiddins makes complete sense, thank you very much for the feedback.

PR #4292, opened 2015-12-01
Allow Logstash to be launched from a symlink

### LINKED ISSUE #4291: bin/logstash not symlink friendly
I have logstash in `/opt/logstash` (installed from RPM, 2.0.0-1). The problem is the utility does not add a system-wide executable in $PATH.

When I try to do it myself (for the root user) by symlinking to `/root/bin` folder, it dies with:

    /root/bin/logstash: line 22: /root/bin/logstash.lib.sh: No such file or directory

Looking at the invocation shell script, this is the problem:

    . "$(cd `dirname $0`/..; pwd)/bin/logstash.lib.sh" 

It's relative to the invoked path, not the shell script.

    [root@machine bin]# dirname /root/bin/logstash 
    /root/bin

Should be an easy fix. Thanks.

untergeek:
These changes allow Logstash to be launched from a symlink _only_ when the `readlink`, or suitable version of `stat` binaries are present. An error message will result if Logstash is launched from a symlink and the path cannot be discovered by one of these two methods.  The message will advise the user to use the full path instead of a symlink to launch Logstash.

fixes #4291

### DIFF
--- a/bin/logstash.lib.sh
+++ b/bin/logstash.lib.sh
@@ -1,5 +1,29 @@
 unset CDPATH
-LOGSTASH_HOME=$(cd `dirname $0`/..; pwd)
+# This unwieldy bit of scripting is to try to catch instances where Logstash
+# was launched from a symlink, rather than a full path to the Logstash binary
+if [ -L $0 ]; then
+  # Launched from a symlink
+  # --Test for the readlink binary
+  RL=$(which readlink)
+  if [ $? -eq 0 ]; then
+    # readlink exists
+    SOURCEPATH=$($RL $0)
+  else
+    # readlink not found, attempt to parse the output of stat
+    SOURCEPATH=$(stat -c %N $0 | awk '{print $3}' | sed -e 's/\‘//' -e 's/\’//')
+    if [ $? -ne 0 ]; then
+      # Failed to execute or parse stat
+      echo "Failed to set LOGSTASH_HOME from $(cd `dirname $0`/..; pwd)/bin/logstash.lib.sh"
+      echo "You may need to launch Logstash with a full path instead of a symlink."
+      exit 1
+    fi
+  fi
+else
+  # Not a symlink
+  SOURCEPATH=$0
+fi
+
+LOGSTASH_HOME=$(cd `dirname $SOURCEPATH`/..; pwd)
 export LOGSTASH_HOME
 
 # Defaults you can override with environment variables
--- a/bin/logstash
+++ b/bin/logstash
@@ -19,7 +19,31 @@
 #   DEBUG=1 to output debugging information
 
 unset CDPATH
-. "$(cd `dirname $0`/..; pwd)/bin/logstash.lib.sh"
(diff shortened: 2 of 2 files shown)


andrewvc:
Just tested this and it seems to work! LGTM

```
~ $ ln -s ~/projects/logstash/bin/logstash myls
~ $ ls -l myls
lrwxr-xr-x  1 andrewvc  staff  46 Dec  1 16:38 myls -> /Users/andrewvc/projects/logstash/bin/logstash
~ $ ./myls -e 'input { stdin {} } output { stdout {} }'
Logstash startup completed
ouehtn
2015-12-01T22:38:24.805Z andrewvc-elastic.local ouehtn

2015-12-01T22:38:25.636Z andrewvc-elastic.local
oeuthn
2015-12-01T22:38:26.140Z andrewvc-elastic.local oeuthn
The error reported is:
  private method `select' called for nil:NilClass {:level=>:fatal}
```

bin/logstash.lib.sh
```diff
@@ -1,5 +1,29 @@
 unset CDPATH
-LOGSTASH_HOME=$(cd `dirname $0`/..; pwd)
+# This unwieldy bit of scripting is to try to catch instances where Logstash
```
suyograo: not related to your PR, but why is this code repeated twice?

bin/logstash.lib.sh
```diff
@@ -1,5 +1,29 @@
...
+# was launched from a symlink, rather than a full path to the Logstash binary
+if [ -L $0 ]; then
```
suyograo: Maybe wrap this a shell util lib and reuse?

bin/logstash.lib.sh
untergeek: Because 

```
cd `dirname $0`/..
```

was in both files, and after I fixed the first, it still wouldn't work until the second _also_ had it.

bin/logstash.lib.sh
untergeek: I considered that, but the error messages are slightly different.  Also, inclusion works with bash, but not as simply with bourne shell (which this uses for its universality).

untergeek: Plus, this code is for _finding_ the correct directory, so finding/including the source of a library in order to find the full path to the file is a full loop of pain :frowning:

untergeek: In other words, you'd never be able to find the util lib if you were starting with a symlink, which is the whole reason for this code's existence.  It's self-defeating to use a lib if the lib has to be found to be used :smile:

suyograo: hah, good point :)

suyograo:
LGTM

dkarlovi:
Wow, this was fast, you guys are awesome. :+1:

suyograo:
Merged in https://github.com/elastic/logstash/commit/68bdde886ab0aa8a7609dc8cf88ebdbaf002e10e

PR #4298, opened 2015-12-02
Replace `def #{type}_func` with define_singleton_method

ph:
When you run multiples pipeline and the config code get evaluated, on
every evaluation the class cached is clear, everytime you were calling a
`func` method you had the latest evaluated code. The `filter_func` and
the `output_func` need to be unique for every instance of the pipeline,
this PR replace the `def` with a `define_single_method` call ensuring
the uniqueness of the code on every instance.

This PR is based on #4254

### DIFF
--- a/logstash-core/spec/logstash/pipeline_spec.rb
+++ b/logstash-core/spec/logstash/pipeline_spec.rb
@@ -47,7 +47,7 @@ class DummyOutput < LogStash::Outputs::Base
 
   def register
   end
-
+  
   def receive(event)
     @events << event
   end
@@ -380,6 +380,35 @@ describe LogStash::Pipeline do
         output.events.first["message"].split("\n").count
       end.to eq(number_of_events)
       pipeline.shutdown
+    end
+  end
+
+  context "Multiple pipelines" do
+    before do
+      allow(LogStash::Plugin).to receive(:lookup).with("input", "generator").and_return(LogStash::Inputs::Generator)
+      allow(LogStash::Plugin).to receive(:lookup).with("codec", "plain").and_return(DummyCodec)
+      allow(LogStash::Plugin).to receive(:lookup).with("filter", "dummyfilter").and_return(DummyFilter)
+      allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(DummyOutput)
+    end
+
+    let(:pipeline1) { LogStash::Pipeline.new("input { generator {} } filter { dummyfilter {} } output { dummyoutput {}}") }
+    let(:pipeline2) { LogStash::Pipeline.new("input { generator {} } filter { dummyfilter {} } output { dummyoutput {}}") }
+
+    it "should handle evaluating different config" do
+      # When the functions are compiled from the AST it will generate instance
+      # variables that are unique to the actual config, the intance are pointing
+      # to conditionals/plugins.
+      #
+      # Before the `defined_singleton_method`, the definition of the method was
+      # not unique per class, but the `instance variables` were unique per class.
+      #
+      # So the methods were trying to access instance variables that did not exist
+      # in the current instance and was returning an array containing nil values for
+      # the match.
+      expect(pipeline1.output_func(LogStash::Event.new)).not_to include(nil)
+      expect(pipeline1.filter_func(LogStash::Event.new)).not_to include(nil)
+      expect(pipeline2.output_func(LogStash::Event.new)).not_to include(nil)
+      expect(pipeline1.filter_func(LogStash::Event.new)).not_to include(nil)
     end
   end
 end
--- a/logstash-core/lib/logstash/config/config_ast.rb
+++ b/logstash-core/lib/logstash/config/config_ast.rb
@@ -107,7 +107,10 @@ module LogStash; module Config; module AST
       ["filter", "output"].each do |type|
         # defines @filter_func and @output_func
 
-        definitions << "def #{type}_func(event)"
+        # This need to be defined as a singleton method
+        # so each instance of the pipeline has his own implementation
+        # of the output/filter function
+        definitions << "define_singleton_method :#{type}_func do |event|"
         definitions << "  targeted_outputs = []" if type == "output"
         definitions << "  events = [event]" if type == "filter"
         definitions << "  @logger.debug? && @logger.debug(\"#{type} received\", :event => event.to_hash)"


andrewvc:
LGTM, just waiting for #4254  to merge in

purbon:
@andrewvc @ph I think this change is not just for your #4254, right? I can also go to master now, right? If not, can we point this PR to your PR and merge it there? or also point it to the feature branch? then we don't have to wait. 

what do you say?

purbon:
@ph having @andrewvc commits in this PR it does really make the review more complicated. hard to spot where are your changes for example. I would recommend to simplify the dependency as I shared in my previous comment.

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -15,6 +15,21 @@ def close
   end
 end
 
+class DummyInputGenerator < LogStash::Inputs::Base
```
purbon: we install the generator plugin as part of the core plugins need for test, https://github.com/elastic/logstash/blob/master/rakelib/default_plugins.rb#L113-L121 any reason not to redefined here?   you know I would like our make our test plugin free, only use mocks/stubs/dummy (pick your name), but if we require it for now I think it would be better to stick to one usage. what do you think?

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -53,6 +72,10 @@ def close
   end
 end
 
+class DummyOutputMore < DummyOutput
```
purbon: Can you explain to me probably the need of this dummy/mock plugin definition? ```DummyOutputMore``` ? not much clear to me the purpose as we have to mocks defined for this.

logstash-core/spec/logstash/pipeline_spec.rb
ph: I was thinking to use the generator but I have checked the `logstash-core` gemspec and it doesn't include the generator plugin and the only test in logstash-core that require the `generator` is the `pipeline_reporter_spec`. 

Its not obvious that the generator is available in the context of theses specs :disappointed:

logstash-core/spec/logstash/pipeline_spec.rb
ph: Let me check If I can remove this and still expose the same problem. I wanted to replicate the situation I had locally with my metric branch.

logstash-core/spec/logstash/pipeline_spec.rb
purbon: I know is not obvious, is in the file I shared, :+1: I understand. I think we should have only 1 way if possible, right? what do you think?

logstash-core/spec/logstash/pipeline_spec.rb
purbon: Yeah, most probably is only wording that is not addient here.

ph:
@pere Concerning this PR against master vs the branch, let me check If I can do it agains't master.

ph:
I will remake this PR against master

ph:
@purbon It can be done, but the test would change, so lets keep it for #4254

purbon:
@ph then I would recommend to point to @andrewvc branch/PR/whatever. It makes no sense we wait to incorporate this if is good enoughts. makes sense? @andrewvc what do you think?

logstash-core/spec/logstash/pipeline_spec.rb
ph: Well, lets keep the DummyGenerator, since we don't need a dependency for that and you were proposing to move away. This also in line with the actual dummy plugin in the current `pipeline_spec`

purbon: are we already ok on going away from plugins to be used in specs? I don't think we agreed, even if I love the idea. not sure what to think.

ph:
updated with the review, @andrewvc do you want to cherry-pick this or I can create the PR against a feature branch?

andrewvc:
@ph well, I don't want to reopen the pr for #4254 so maybe we can wait till that gets in master? I think I can address all of the open concerns today. 

PRs against PRs are hard :(

ph:
@andrewvc np I will rebase, when @purbon gives his lgtm.

purbon:
LGTM

andrewvc:
Hey @ph are you planning on rebasing this?

ph:
rebased with master @andrewvc !

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -47,6 +47,10 @@ def initialize(params={})
 
   def register
   end
+  
+  def threadsafe?
+    false
```
andrewvc: I don't think this currently exists as an API. In https://github.com/elastic/logstash/pull/4391/files#diff-3d93d6656c81049b6de7a09dbe141a46R30 I added a new method `Outputs/Base.declare_threadsafe!` that would declare this at the class level.

ph: let me check that, if its needed, I think I had an error without it.

ph: You were right, I think at the time I have created the pr it was required.

ph:
updated with your comment @andrewvc

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -382,4 +382,23 @@ class TestPipeline < LogStash::Pipeline
...
+      allow(LogStash::Plugin).to receive(:lookup).with("codec", "plain").and_return(DummyCodec)
+      allow(LogStash::Plugin).to receive(:lookup).with("filter", "dummyfilter").and_return(DummyFilter)
+      allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(DummyOutput)
+    end
+
+    let(:pipeline1) { LogStash::Pipeline.new("input { generator {} } filter { dummyfilter {} } output { dummyoutput {}}") }
+    let(:pipeline2) { LogStash::Pipeline.new("input { generator {} } filter { dummyfilter {} } output { dummyoutput {}}") }
+
+    it "should handle evaluating different config" do
+      expect(pipeline1.output_func(LogStash::Event.new)).not_to include(nil)
```
andrewvc: This could use a comment explaining why we're checking for it not to be nil.

andrewvc:
Could use an explanatory comment, but after that LGTM!

ph:
Added a comment!

PR #4303, opened 2015-12-04
allow logstash to run in a dir with spaces

### LINKED ISSUE #4283: logstash fails to start if the parent directory contains a space
Reproduction on my MacBook:
 - &lt;Download logstash 2.1.0&gt;
 - mkdir "DIR WITH SPACE"
 - cd DIR\ WITH\ SPACE
 - tar -xvzf ~/Downloads/logstash-2.1.0.tar.gz
 - cd logstash-2.1.0
 - ./bin/logstash

And I get: `Error: Could not find or load main class EL.lostash-2.1.0-heapdump.hprof`.

jsvd:
because HeapDumpPath is set, if it contains spaces the jruby_opts
function will create two strings appended by -J
even if the list of jruby opts is constructed using replacement
shell script will not escape the path correct to the heap dump path
option needs to be added as late as possible

this is a *nix fix only, windows is not affected

fixes #4283

### DIFF
--- a/bin/logstash.lib.sh
+++ b/bin/logstash.lib.sh
@@ -38,7 +38,8 @@ setup_java() {
     # Causes the JVM to dump its heap on OutOfMemory.
     JAVA_OPTS="$JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError"
     # The path to the heap dump location
-    JAVA_OPTS="$JAVA_OPTS -XX:HeapDumpPath=${LOGSTASH_HOME}/heapdump.hprof"
+    # This variable needs to be isolated since it may contain spaces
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM, works fine now

```
suyog@machine:/tmp/test spaces/logstash-3.0.0.dev$ bin/logstash -e 'input { stdin { } } output { stdout { } }'
Logstash startup completed
abc
2015-12-04T21:34:25.580Z Suyogs-MacBook-Pro-2.local abc
```

jasperla:
This breaks when `$JAVA_OPTS` is set, because then `$HEAP_DUMP_PATH` won't be set and `-J ` is passed to jruby/java which doesn't recognize it as a valid option.

jordansissel:
@jasperla good catch. Can you open a new bug about that?

PR #4309, opened 2015-12-06
Docs: Clarify that it's the @timestamp field that's formatted

magnusbaeck:
The `%{+FORMAT}` form always formats the `@timestamp` field but previously that hasn't been obvious from the documentation.

### DIFF
--- a/docs/static/configuration.asciidoc
+++ b/docs/static/configuration.asciidoc
@@ -323,10 +323,10 @@ output {
 }
 ----------------------------------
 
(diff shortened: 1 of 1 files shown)


suyograo:
@magnusbaeck .md files are gone, they were used in 1.4 and below. The docs are now in https://github.com/elastic/logstash/blob/master/docs/static/configuration.asciidoc

magnusbaeck:
Oops. That wasn't obvious. I'll patch in the change to the asciidoc file after the holidays.

magnusbaeck:
Correct file patched—please have another look @suyograo.

jordansissel:
LGTM

PR #4314, opened 2015-12-07
Reorganize tests in a common main location

purbon:
The main motivation of this PR is to reorganize the tests location, as of this days we've test located in many different places, for example the ```acceptance_spec``` directory, the ```test``` directory, the ```spec``` and so on. This is a concern I think we should fix in order to provide a single location, well described location that helps everyone locate old test, place new test and basically have a better testing experience.

@gmoskovicz it would be nice if you can validate the windows part of this test and make sure they work as expected.

### Notes to reviewers

* Please understand this PR as a living thing, and let it serve as a living example.
* The motivation of this PR is to locate test all together, a part from that I have no special concerns about naming or final location.
* Ideas and improvements always welcome.


purbon:
@ph @jordansissel would be awesome to get your feedback in here. this is something I wanted to do for such a long time I finally took some time to.

purbon:
Let me know if there is any doubt or missing bit I can provide to make this more clear.

gmoskovicz:
@purbon looks good for me.. there are no deleted files and i verified that the scripts are there!

rakelib/test.rake
```diff
@@ -24,7 +24,7 @@ namespace "test" do
     # logstash-core-event specs since currently this is the most complete Event and Timestamp specs
     # which actually defines the Event contract and should pass regardless of the actuall underlying
     # implementation.
-    specs = ["spec/**/*_spec.rb", "logstash-core/spec/**/*_spec.rb", "logstash-core-event/spec/**/*_spec.rb"]
+    specs = ["test/unit/**/*_spec.rb"]
```
andrewvc: With the java event rewrite are we sure we don't want to keep the event specs separate still?

rakelib/test.rake
```diff
@@ -87,7 +87,7 @@ namespace "test" do
     require "simplecov"
     SimpleCov.start do
       # Skip non core related directories and files.
-      ["vendor/", "spec/", "bootstrap/rspec", "Gemfile", "gemspec"].each do |pattern|
+      ["vendor/", "test/", "bootstrap/rspec", "Gemfile", "gemspec"].each do |pattern|
```
andrewvc: Why are we now calling our rspec specs 'test'. Isn't that lesson common for rspec?

jordansissel:
Overall I am :+1: on this proposed change. I agree a bit with @andrewvc on the naming of 'test' vs 'spec' but I don't have strong preferences.

andrewvc:
I think that telling people to run `bin/rspec test` is unexpected. I think newcomers will expect us to be using Test::Unit in that case. I'd love to be educated as to why I may be wrong here however.

jordansissel:
@andrewvc it'll still be a bit weird-feeling to run `bin/rspec test/spec` so either way we're having to pass an argument to rspec, right?

andrewvc:
@jordansissel I was hoping to be able to just run `bin/rspec spec` . Is that not possible?

rakelib/test.rake
purbon: good point with the event specs, I guess this makes sense to have it separate. Somehow it starts to get strange to me this days when we get

- logstash
- logstash-core
- logstash-event
- logstash-event-java

where to have specs relevant to one close to the source of them. I guess a good mid point would be to have the event specs closer to the event folder as I guess we will have this as interchangeable  components with the java one, right? 

I think your point of having event separated might makes sense, I would just dislike having test to spread if not necessary as it makes tracking them harder, at less for me.

purbon: @andrewvc after thinking a bit more on this I have a few questions about it, I would really love to understand better:

- logstash-core-event test include event_spec but also timestamp_spec, the first one I understand belongs to the event, but not the timestamp. Do you know about the relation of both?

- I understand we want to extract the java-event, but is still having a logstash-event "gem" something meaningful to us? does it makes sense? what are the benefits we get for it?

rakelib/test.rake
purbon: I changed it because some folks complained to me when I first introduced this idea that specs was misunderstood, but happy to see this also confuses you, so I qill bring back the spec dir as I understand is much clear to "ruby" folks.

purbon:
@andrewvc I think most people might be confused as https://github.com/elastic/logstash/pull/4314#issuecomment-163714985 said, but no strong arguments in favor of spec/test directory names. my pending thought is with the logstash-event separation, would really love to understand pros and cons of having it in a different location, specially in terms of code and test location.

purbon:
@andrewvc @jordansissel ```bin/rspec``` should work as expected now, let me know what do you think.

electrical:
The nodeset paths for acceptance testing will break acceptance testing since it expects it all in a specific path..

purbon:
After a discussion with @colinsurprenant looks like this is not a nice move, so most probably will be ditching this PR. Please share your thoughts in benefit of having a more uniform and unified test locations as we've now or we should discard this idea :+1: open to every option! for a better and cleaner test organization.

colinsurprenant:
Just a quick note to resume my thoughts, we can debate about some more if needed (I need to run now) - I don't see the point of breaking apart specs from their related sources. everything in `logstash-core`, `logstash-core-event` and `logstash-core-event-java` is structured as an independent & self-contained  gem, so everything that concerns that gem/code should stay together. I don't see the problem of having very specific specs with
```
$ bin/rspec spec
$ bin/rspec logstash-core/spec
$ bin/rspec logstash-core-event/spec
$ bin/rspec logstash-core-event-java/spec
```

These typically test different things and we can choose to run one and not the other while we work on specific parts of code.

To group specs run we can use rake tasks like `rake test:core` which basically does the above run.

colinsurprenant:
OTOH I am open on discussing grouping all test-related stuff (integration, unit)  under the respective spec (or other name like test?) directories for each gem, but definitely not moving each core gem specific specs/tests away from their own directories. we can have rake tasks to help running all specs/tests.

I **do** like the fact that I can just run `bin/rspec {...}/spec` and know it will run all specs for that component. It is pretty standard, or even `bin/rspec {...}` with just the base dir without `/spec/`.

purbon:
Deciding to close this after the long explanation and clarifications by @colinsurprenant either way is not that important for now to get this sorted out :smile_cat:  and things works. We can move this discussion to an issue.

PR #4321, opened 2015-12-09
Fix configtest in sysv init script

neumayer:
Replace non-existing log function with echo.
Replace "exit 1" with "return 1" to make quiet function work.

### DIFF
--- a/pkg/logstash.sysv
+++ b/pkg/logstash.sysv
@@ -137,8 +137,8 @@ force_stop() {
 configtest() {
   # Check if a config file exists
   if [ ! "$(ls -A ${LS_CONF_DIR}/* 2> /dev/null)" ]; then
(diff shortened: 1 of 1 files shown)


jordansissel:
Looks like the CLA check failed; have you signed our [CLA](https://www.elastic.co/contributor-agreement/) ?

neumayer:
Three times. I'm struggling with this :-), will try even harder tomorrow.

neumayer:
Can this cla test be retriggered? How does that work? Is there any more I can do apart from signing it and confirm the email address?

jordansissel:
CLA is passing, I just verified. The automated check in this PR only looks for every new commit.

neumayer:
Will force-pushing solve it, i.e. do we need the confirmation here?

PR #4334, opened 2015-12-10
Add first iteration of the webapi

purbon:
Hi,
 this PR add the first iteration of the webapi, it include the next set of features:

* A sinatra webapp to handle the api resources.
* A set of test for the webapp.
* Code to have an embedded puma webserver that will handle the web requests. 

### notes

for now the web server is created in the agent execute, I tried moving this to the new background service but is currently broken with this error:

```
skywalker% bin/logstash -e ""
Concurrent::TimeoutError {:class=>"Concurrent::TimeoutError", :backtrace=>nil, :level=>:warn}
Concurrent::TimeoutError {:class=>"Concurrent::TimeoutError", :backtrace=>nil, :level=>:warn}
```


logstash-core/api/lib/app/root.rb
```diff
@@ -0,0 +1,15 @@
+# encoding: utf-8
+require "app"
+
+module LogStash::Api
+  class Root < BaseApp
+
+    get "/" do
+      content = { "name" => "Logstash API",
+                  "version" => { "number" => "0.1.0" },
+                  "tag" => "You know, for logs" }
```
suyograo: Logstash is more than just logs, so let's remove this :) I think I put it in the design doc, but we can pretty much remove it

jordansissel: well, it's all perspective.

When I say log, I mean (time, data). Most people think of logs as being server logs, so maybe it's fair to try a different term. I dunno.

jordansissel: But yeah I don't think we need it ;P

purbon: yeah, here you the superpowers of copy and paste! :-P I will remove it.

logstash-core/lib/logstash/webservers/puma.rb
```diff
@@ -0,0 +1,67 @@
...
+  private
+
+  def parse_options
+    @config = Puma::Configuration.new(cli_options)
+    @config.load
+    @options = @config.options
+  end
+
+
+
```
ph: So many new lines, time to get a new return key ?:D

purbon: lol. true :+1:

logstash-core/api/lib/app.rb
```diff
@@ -0,0 +1,38 @@
...
+  class BaseApp < ::Sinatra::Application
+
+    attr_reader :factory
+
+    if settings.environment == :production
+      ::Logger.class_eval { alias :write :'<<' }
+      access_log    = ::File.join(::File.dirname(::File.expand_path(__FILE__)),'../../..', 'access.log')
+      access_logger = ::Logger.new(access_log)
+      error_logger  = ::File.new(::File.join(::File.dirname(::File.expand_path(__FILE__)),'../../..','error.log'),"a+")
+      error_logger.sync = true
```
ph: I think we should use our internal logger here (cabin) we could probably ignore the access log unless we run in debug mode?

logstash-core/api/lib/app.rb
```diff
@@ -0,0 +1,38 @@
...
+require "app_helpers"
+require "app/service"
+require "app/command_factory"
+
+module LogStash::Api
+  class BaseApp < ::Sinatra::Application
+
+    attr_reader :factory
+
+    if settings.environment == :production
```
ph: Sinatra start in production mode by default?

logstash-core/api/config.ru
```diff
@@ -0,0 +1,22 @@
+ROOT = File.expand_path(File.dirname(__FILE__))
+$LOAD_PATH.unshift File.join(ROOT, 'lib')
+Dir.glob('lib/**').each{ |d| $LOAD_PATH.unshift(File.join(ROOT, d)) }
+
+require 'sinatra'
+require 'app/root'
+require 'app/pipeline'
+require 'app/system'
+
+env = ENV["RACK_ENV"].to_sym
```
ph: I don't think we need this, for us in reality the `RACK_ENV` is defined by the `--debug` option passed on the command line, no ?

logstash-core/api/lib/app.rb
```diff
@@ -0,0 +1,38 @@
+# encoding: utf-8
+require "json"
```
ph: We should not use the "json" library but instead our `LogStash::Json` which uses jrjackson.

logstash-core/api/lib/app.rb
```diff
@@ -0,0 +1,38 @@
...
+      error_logger  = ::File.new(::File.join(::File.dirname(::File.expand_path(__FILE__)),'../../..','error.log'),"a+")
+      error_logger.sync = true
+
+      configure do
+        use ::Rack::CommonLogger, access_logger
+      end
+
+      before do
+        env["rack.errors"] =  error_logger
+      end
```
ph: Oh rack oh why rack you have a so ugly api sometime :P

logstash-core/api/lib/app/pipeline/stats_command.rb
```diff
@@ -0,0 +1,20 @@
+# encoding: utf-8
+require "app/command"
+
+class LogStash::Api::PipelineStatsCommand < LogStash::Api::Command
+
+  def run
+    #return whatever is comming out of the snapshot event, this obvoiusly
+    #need to be tailored to the right metrics for this command.
+    stats = service.get(:pipeline_stats)
```
ph: I like that very much, If we want to move away from the observer this will not impact this part +1.

logstash-core/api/lib/app/service.rb
```diff
@@ -0,0 +1,31 @@
...
+  def stop
+    LogStash::Instrument::Collector.instance.delete_observer(self)
+  end
+
+  def update(time, snapshot)
+    if snapshot_rotation_mutex.try_lock
+      @snapshot = snapshot
+      @snapshot_rotation_mutex.unlock
+    end
+  end
```
ph: Thanks for bringing this to me @purbon and sorry for taking more time to review it, Sound like I will revisit thread safety :D

Lets use the Concurrent::AromicReference http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/MutexAtomic.html for now.

logstash-core/api/lib/app/service.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def get(key=:service_metrics)
+    @snapshot.to_event.to_hash
+  rescue
+    {}
+  end
+end
```
ph: Might be better to have a check against `NIL` ?

logstash-core/api/lib/app/system/info_command.rb
```diff
@@ -0,0 +1,23 @@
...
+  private
+
+  def type
+    ["input", "output", "filter"][rand(3)]
+  end
+
+  def name
+    ["elasticsearch", "json", "yaml", "translate"][rand(4)]
+  end
+end
```
ph: Love the shim.

logstash-core/api/lib/app_helpers.rb
```diff
@@ -0,0 +1,9 @@
+# encoding: utf-8
+require "json"
+
+module LogStash::Api::AppHelpers
+
+  def respond_with(data)
+    data.to_json
+  end
+end
```
ph: use `LogStash::Json` instead

logstash-core/api/spec/lib/api/pipeline_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    get "/stats"
+    expect(last_response).to be_ok
+  end
+
+  it "respond to the stats resource" do
+    get "/plugins"
+    expect(last_response).to be_ok
+  end
+
+end
```
ph: This will be clean!

logstash-core/lib/logstash/agent.rb
```diff
@@ -34,6 +36,10 @@ def execute
     start_background_services
 
     @pipelines.each { |_, p| Thread.new { p.run } }
+    @web_agent = Thread.new do
+      LogStash::WebServer::Puma.new(@logger).run
+    end
+
```
ph: Move that logic `start_background_services` and make sure we can stop it in `stop_background_services`

logstash-core/lib/logstash/webservers/puma.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def error(str)
+    logger.error(str) if logger.error?
+  end
+
+
+  def write_state
+
+  end
```
ph: Required? what this method do?

ph: Maybe add a comment if required

logstash-core/lib/logstash/webservers/puma.rb
```diff
@@ -0,0 +1,67 @@
...
+require 'puma/binder'
+require 'puma/configuration'
+
+class LogStash::WebServer::Puma < LogStash::WebServer::Base
+
+  extend Forwardable
+
+  attr_reader :runner, :binder, :events
+
+  def_delegator :@runner, :stats, :stats
```
ph: double `stats`?

logstash-core/lib/logstash/webserver.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    begin
+      Signal.trap "SIGTERM" do
+        stop
+      end
+    rescue Exception
+      log "*** SIGTERM not implemented, signal based gracefully stopping unavailable!"
+    end
+  end
```
ph: This is already handler inside the `agent.rb` I don't think we need to have a custom logic in our webserver

logstash-core/logstash-core.gemspec
```diff
@@ -27,6 +27,8 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency "gems", "~> 0.8.3"  #(MIT license)
   gem.add_runtime_dependency "concurrent-ruby", "1.0.0"
   gem.add_runtime_dependency "jruby-openssl", ">= 0.9.11" # Required to support TLSv1.2
+  gem.add_runtime_dependency "sinatra", '~> 1.4', '>= 1.4.6'
+  gem.add_runtime_dependency "puma", '~> 2.15', '>= 2.15.3'
```
ph: +1 for strict version

logstash-core/lib/logstash/webservers/puma.rb
```diff
@@ -0,0 +1,67 @@
...
+    @events  = Puma::Events.strings
+    @binder  = Puma::Binder.new(@events)
+    @binder.import_from_env
+  end
+
+  def run
+    setup_signals
+    log "=== puma start: #{Time.now} ==="
+
+    @runner = Puma::Single.new(self)
```
ph: if we use `Puma::Single` does it mean that we have a single thread webserver?

logstash-core/lib/logstash/webservers/puma.rb
```diff
@@ -0,0 +1,67 @@
...
+    graceful = options.fetch(:graceful, true)
+
+    if graceful
+      @runner.stop_blocked
+    else
+      @runner.stop
+    end
+    @status = :stop
+    log "=== puma shutdown: #{Time.now} ==="
+  end
```
ph: Same as previous comment on `TRAP`, graceful and the force calls should be triggered from the agent.

logstash-core/api/spec/spec_helper.rb
```diff
@@ -0,0 +1,11 @@
...
+ROOT = File.expand_path(File.join(File.dirname(__FILE__), ".."))
+$LOAD_PATH.unshift File.join(ROOT, 'lib')
+Dir.glob('lib/**').each{ |d| $LOAD_PATH.unshift(File.join(ROOT, d)) }
+
+require "logstash/devutils/rspec/spec_helper"
+
+require 'rack/test'
+require 'rspec'
+
+ENV['RACK_ENV'] = 'test'
```
ph: Silly rack :dancers:

logstash-core/lib/logstash/webservers/puma.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  private
+
+  def parse_options
+    @config = Puma::Configuration.new(cli_options)
```
ph: Can we pass the configuration to puma without using a `config.ru`, the config.ru feel strange to me?

logstash-core/lib/logstash/webserver.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def env
+    ENV['RACK_ENV'] || "development"
+  end
+
+  def set_environment
+    @options[:environment] = env
+    ENV['RACK_ENV']        = env
+  end
+end
```
ph: Do we really need a base class for puma, I don't think we plan to support different web server in the mid term?
I would be in favor to only have a Webserver class which hide the implementation details of Puma, wdyt?

logstash-core/lib/logstash/agent.rb
ph: I saw the previous note on your reviewer, so ignore previous comment I will try to reproduce your issue first.

ph:
@purbon I like how the command/code architecture for the Sinatra/Command interaction I think this would give us a lot of flexibility. +1

I know we had this discussion in video but I think the "api" should be defined in `logstash-core/lib/logstash/api` I don't think it should be in a top level directory, this feel strange to me and I think we will get strong opinions on this later on when we want to merge that into master.

The other point, I think we should use the current logstash logger instead of creating a new one, unless the need arise. Using a new logger mean more administration burden on the admin, also I don't see a lot of values with the access log since this port should be protected in some way and would be minimally hit. 

The other problem I see, I don't think Puma should register itself to the `TRAP` signals, this should be handled higher in the agent.

purbon:
@ph 

> I know we had this discussion in video but I think the "api" should be defined in logstash-core/lib/logstash/api I don't think it should be in a top level directory, this feel strange to me and I think we will get strong opinions on this later on when we want to merge that into master.

Well for this reason I also defended the arguments why it should not go into core, for me this feels completely strange, where to locate the API, I'm not saying we don't bundle it together, but I would rather make it a top level location than within logstash-core as it should be a different living thing only interfacing with the core. 

> The other point, I think we should use the current logstash logger instead of creating a new one, unless the need arise. Using a new logger mean more administration burden on the admin, also I don't see a lot of values with the access log since this port should be protected in some way and would be minimally hit.

I agree we should have only one logger here.

> The other problem I see, I don't think Puma should register itself to the TRAP signals, this should be handled higher in the agent.

makes sense.

logstash-core/api/config.ru
purbon: make sense.

logstash-core/api/lib/app.rb
purbon: makes sense, I used this for simplicity for now. will update.

logstash-core/api/lib/app.rb
purbon: nope,

> A symbol specifying the deployment environment; typically set to one of :development, :test, or :production. The :environment defaults to the value of the RACK_ENV environment variable (ENV['RACK_ENV']), or :development when no RACK_ENV environment variable is set.

logstash-core/api/lib/app.rb
purbon: makes sense

logstash-core/api/config.ru
ph: Well, if we start he server in the agent, should the agent be aware of that? @jsvd Should the agent start the logstash webservice?

purbon: agent is not aware, as of now of status, but can be pass through constructor, works for me.

ph: I think it would make sense.

logstash-core/api/lib/app/pipeline/stats_command.rb
purbon: this was the idea, also to isolate from snapshot and get all info from this api, or also other services.

logstash-core/api/lib/app/service.rb
purbon: can you please introduce me to reasons why using it? what are the benefits?

logstash-core/api/lib/app/service.rb
purbon: we could, but this interface is going to change as for example makes no sense to have ```to_event.to_hash``` to get data out of the snapshot :+1: We can narrow down the possible errors on the fly with the different interactions.

logstash-core/lib/logstash/agent.rb
purbon: yeah, for now is broken, so I rather keep it here :cry:, but I do agree makes sense to be moved there.

logstash-core/lib/logstash/webservers/puma.rb
purbon: yeah, will add a comment, is basically because of https://github.com/puma/puma/blob/master/lib/puma/single.rb#L82 providing you the option to write  the server state out somehow.

logstash-core/lib/logstash/webservers/puma.rb
purbon: this is the interface to stop actually being called from the agent.

logstash-core/lib/logstash/webservers/puma.rb
purbon: I can name this file as you with, basically puma waits for a ruby file as all other rack applications, this file is going to be one actually starting up the web application.

logstash-core/lib/logstash/webserver.rb
purbon: we can have only one, makes it easy.

ph:
Concerning the directory, If its top level directory why the api use a `LogStash` namespace?

purbon:
@ph It uses a Logstash namespace because it life for logstash, but is no core component as the core for me is the pipeline/agent and its surrounding components. also the runner is wrongly places in the logstash-core.

logstash-core/lib/logstash/webservers/puma.rb
purbon: At my understanding, nope, we can change this by adding more thread with the ```:min_thread``` and ```:max_thread``` configuration options

purbon:
@ph I went throught most/all of your request ! thanks a lot for your feedback, very valuable.

logstash-core/api/lib/app/service.rb
ph: agree

ph:
I am OK to merge this in the feature branch and do new PRs that implement the real methods and refine the implementations.

purbon:
cool, as soon as you merge #4358 I will rebase and merge this to then wire up the methods in. makes sense?

purbon:
closing this as being already merged.

PR #4336, opened 2015-12-10
Keep the namespace context of the metric up to the collector service

ph:
Small PRs for initialize rework of the collector to be able to a reference of the current namespace context.
This will allow to write structured queries against the snapshot.
fixed a few typos.


ph:
There is still a few core test failling with this branch I will look into it with another PRs.

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -29,7 +29,7 @@ def push(*args)
     end
 
     def self.snapshot_rotation_time=(time)
-      @snapsho_rotation_time = ime
+      @snapshot_rotation_time = time
```
purbon: good catch on this, do we have unit test for the collector?

purbon:
Hi,
 please correct me if I understood something wrong, I understand this changes introduces basically namespacing for keys, so we will end up having in the snapshot a keys in a form of ```root.sec0.sec1``` and so on, right?

I see this as a good first step towards a more, lets say "web api" friendly request pattern. Next steps might be trying to see within the snapshot how to best represent. this information.

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -4,36 +4,48 @@
...
+      validate_key!(key)
+      collector.push(namespace_information, key, :gauge, value)
     end
 
-    def namespace(key)
-      Metric.new(collector, merge_keys(key.to_sym))
+    def namespace(sub_namespace)
+      raise MetricNoNamespaceProvided if sub_namespace.nil? || sub_namespace.empty?
+
+      new_namespace = namespace_information.dup
```
purbon: small comment here, why ```dup``` and not ```clone```? is there any particular reason?

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -4,36 +4,48 @@
...
   class Metric
-    attr_reader :collector, :base_key
+    attr_reader :collector, :namespace_information
 
-    def initialize(collector, base_key = nil)
+    public
+    def initialize(collector, namespace = nil)
       @collector = collector
-      @base_key = base_key
+      @namespace_information = Array(namespace).map(&:to_sym)
```
purbon: do we aim to introduce some kind of separator between the different namespace components? this will be useful when performing operations in the snapshot to organize the data.

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -45,20 +57,14 @@ def time(key, &block)
...
     private
-    def merge_keys(key)
-      valid_key!(key)
-      [@base_key, key.to_sym]
-    end
-    
-    def valid_key!(key)
-      raise MetricNoKeyProvided if key.nil? || key == ""
+    def validate_key!(key)
+      raise MetricNoKeyProvided if key.nil? || key.empty?
```
purbon: should we also perform here a validation for a "valid" key? I say a key that belongs to this metric object, for now is actually possible to update anything with any metrics object.

logstash-core/lib/logstash/instrument/metric_type/counter.rb
```diff
@@ -22,7 +22,7 @@ def decrement(value = 1)
       @counter.decrement(value)
     end
 
-    def execute(type, key, action, time, value)
+    def execute(namespace, key, type, action, value)
```
purbon: when are namespace, key, and type used?

logstash-core/lib/logstash/instrument/probe/os.rb
```diff
@@ -59,7 +59,6 @@ def self.underscore(str)
         str.gsub!(/([a-z\d])([A-Z])/,'\1_\2')
         str.tr!("-", "_")
         str.downcase!
-        puts str
```
purbon: yay, good catch on that :+1:

logstash-core/lib/logstash/instrument/snapshot.rb
```diff
@@ -14,8 +14,8 @@ def initialize
     end
 
     def push(*args)
-      type, key, _ = args
-      metric = @metrics.fetch_or_store(key, concrete_class(type, key))
+      namespace, key, type, _ = args
+      metric = @metrics.fetch_or_store([namespace, key].join('-'), concrete_class(type, key))
```
purbon: good job here with the ```join('-')``` I wonder if we should really do it here, or we can have it pre computed? I imagine this join being something expensive at the end of the day. what do you think?

purbon:
@ph some comments made, in general looks good, good first step to have proper format in the snapshot, but still a bit to do there to get proper access patterns.

logstash-core/lib/logstash/instrument/collector.rb
ph: Not yet, Next on the list

logstash-core/lib/logstash/instrument/metric.rb
ph: +1 should be clone, I am not sure why ruby decide to have a subtle difference between the two.

logstash-core/lib/logstash/instrument/metric.rb
ph: What do yo mean by separator?

logstash-core/lib/logstash/instrument/collector.rb
purbon: I recommend we add even simple ones for now, this would catch stupid stuff like this.

logstash-core/lib/logstash/instrument/metric.rb
ph: If I understand correctly you are talking about type checking for the collected metric?
If that the case yes this is a problem with the actual code! +1

Lets say that we have a key of "root.pipeline01.events_in" of type `Counter` and we try to use another type with the same key it should raise an error? Correct?

One way to fix that issue could be to add the type to the actual key automatically, so the previous key would become.

`root.pipeline01.events_in.counter` ?

logstash-core/lib/logstash/instrument/metric.rb
purbon: to have something in between each term of the namespace, ```a.b.c```

logstash-core/lib/logstash/instrument/metric_type/counter.rb
ph: This is a generic call, I wasnt sure it was needed I just fixed the signature of the method.
I need to revisit that.

logstash-core/lib/logstash/instrument/snapshot.rb
ph: This is an expensive call yes, I wanted to change the collection of the metric without breaking all the thing, so this was needed to reduce the size of this PR.

One way we could optimize this would be to use an hash map with precompiled keys?

logstash-core/lib/logstash/instrument/metric.rb
ph: After rereading the other code, I see what you mean here, the `[namespace, key].join('-')` doesnt return what you expect.

purbon: but only between namespace and key, as I understand from the test namespace is not just one word, right? otherwise why using array here?

logstash-core/lib/logstash/instrument/metric.rb
purbon: yes

logstash-core/lib/logstash/instrument/metric_type/counter.rb
purbon: ok, because the signature feels very strange to me.

logstash-core/lib/logstash/instrument/snapshot.rb
purbon: might be, or also push this call upstream somehow, not sure. I see this as a precomputed value as much as possible.

logstash-core/lib/logstash/instrument/metric_type/counter.rb
ph: True, I will check it when change the snapshot format :(

logstash-core/lib/logstash/instrument/metric.rb
ph: I think it should be done at the data structure level to add this information and not on the library level what do you think?

purbon: You introduced here a validation method to check, for now only the presence of the key. I think if we keep this method, we should add some other checks, or change the name of the method to exist? or something like.

more validation could be done in the user level, but it all depends on how flexible/strict we aim to be, keeping in  mind possible errors.

ph: +1 for exist

ph:
I will close this an reopen a new one, this PR was also made on the logstash repo and not on my fork.. :(

ph:
Moving discussion to https://github.com/elastic/logstash/pull/4358

Issue #3224, opened 2015-05-12
Installing plugins in linux (from source code is broken)


purbon:
As discovered from http://build-eu-00.elastic.co/view/LS%201.5/job/Logstash%20Regression%201.5%20(Release)/3/jdk=JDK7,label=metal-pool/console

```
Using sawyer 0.6.0
Using octokit 3.8.0
Using rubyzip 1.1.7
Using bundler 1.9.7
Bundle complete! 114 Gemfile dependencies, 194 gems now installed.
Gems in the group development were not installed.
Bundled gems are installed into ./vendor/bundle.
[plugin:install-core] Installing core plugins
/home/jenkins/workspace/Logstash Regression 1.5 (Release)/jdk/JDK7/label/metal-pool/bin/plugin: 4: .: Can't open /home/jenkins/workspace/bin/logstash.lib.sh
rake aborted!

/home/jenkins/workspace/Logstash Regression 1.5 (Release)/jdk/JDK7/label/metal-pool/rakelib/plugin.rake:7:in `install_plugins'
/home/jenkins/workspace/Logstash Regression 1.5 (Release)/jdk/JDK7/label/metal-pool/rakelib/plugin.rake:34:in `(root)'
org/jruby/RubyProc.java:271:in `call'
org/jruby/RubyArray.java:1613:in `each'
org/jruby/RubyArray.java:1613:in `each'
org/jruby/RubyArray.java:1613:in `each'
org/jruby/RubyKernel.java:1087:in `load'
Tasks: TOP => test:install-core => plugin:install-core
(See full trace by running task with --trace)
WARN: Unresolved specs during Gem::Specification.reset:
      ffi (>= 0)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
rake aborted!
Could not find rake-10.4.2 in any of the sources
/home/jenkins/workspace/Logstash Regression 1.5 (Release)/jdk/JDK7/label/metal-pool/vendor/bundle/jruby/1.9/gems/bundler-1.9.7/lib/bundler/spec_set.rb:92:in `materialize'
org/jruby/RubyArray.java:2444:in `map!'
/home/jenkins/workspace/Logstash Regression 1.5 (Release)/jdk/JDK7/label/metal-pool/vendor/bundle/jruby/1.9/gems/bundler-1.9.7/lib/bundler/spec_set.rb:85:in `materialize'
[...]

PR #4338, opened 2015-12-11
Fix LOGSTASH_HOME having spaces in between

### LINKED ISSUE #4331: Fix for #4303 breaks with $JAVA_OPTS set
The fix for #4303 breaks when `$JAVA_OPTS` is set.

[When it's set](https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L51), then [HEAP_DUMP_PATH](https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L66) will be empty. Which then leads to [passing just -J](https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L179) to jruby, which is not recognized as a valid option.

purbon:
There are issues when having spaces in the path, when using `$(cd `dirname $SOURCEPATH`/..; pwd)` if spaces are present the lookup gets wrong, this happen also when running tasks with rake like `rake test:install-core`

Proposed solution is break up the original expression so it could be made each operation space friendly. 

Fixes #4331 and #3224

### DIFF
--- a/bin/logstash.lib.sh
+++ b/bin/logstash.lib.sh
@@ -23,7 +23,6 @@ else
   SOURCEPATH=$0
 fi
 
-#LOGSTASH_HOME=$(cd `dirname $SOURCEPATH`/..; pwd)
 DIR=`dirname "$SOURCEPATH"`
 cd "${DIR}/.."
 LOGSTASH_HOME="$(pwd)"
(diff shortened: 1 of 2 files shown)


jsvd:
:(

```
% JAVA_OPTS='-server -XX:HeapDumpPath="/tmp/a b/be.hprof"' bin/logstash -v
WARNING: Default JAVA_OPTS will be overridden by the JAVA_OPTS defined in the environment. Environment JAVA_OPTS are -server -XX:HeapDumpPath="/tmp/a b/be.hprof"
Error: Could not find or load main class b.be.hprof"
```

purbon:
@jsvd good test, thanks.

jsvd:
I think the only solution here will be to extract this particular option from the JAVA_OPTS array, and set HEAP_DUMP_PATH so it can be passed separately into the cli arguments

purbon:
Good catch with this issue, I was not aware as he reported a different issue, 

```Which then leads to passing just -J to jruby, which is not recognized as a valid option.``` basically I understand he said that if using JAVA_OPTS, we had no HEAP_DUMP_PATH so the script was wrong. Will fix also your test, good catch.

jsvd:
I'm ok with merging this and opening another issue for the particular issue I raised. it shouldn't be a common use case.

purbon:
@jsvd I guess some other vars might be broken also when having spaces.

jsvd:
can you confirm that this works with https://github.com/elastic/logstash/pull/4292?

PR #4340, opened 2015-12-11
Next generation pipeline

andrewvc:
This replaces #4254 which got just a bit too big. It's squashed and rebased off master.

### DIFF
--- a/logstash-core/lib/logstash/output_delegator.rb
+++ b/logstash-core/lib/logstash/output_delegator.rb
@@ -1,0 +1,68 @@
+# encoding: utf-8
+require "concurrent/atomic/atomic_fixnum"
+
+# This class goes hand in hand with the pipeline to provide a pool of
+# free workers to be used by pipeline worker threads. The pool is
+# internally represented with a SizedQueue set the the size of the number
+# of 'workers' the output plugin is configured with.
+#
+# This plugin also records some basic statistics
+module LogStash; class OutputDelegator
+  attr_reader :workers, :config, :worker_count
+
+  # The *args this takes are the same format that a Outputs::Base takes. A list of hashes with parameters in them
+  # Internally these just get merged together into a single hash
+  def initialize(logger, klass, *args)
+    @logger = logger
+    @config = args.reduce({}, :merge)
+    @klass = klass
+    @worker_count = @config["workers"] || 1
+
+    @worker_queue = SizedQueue.new(@worker_count)
+
+    @workers = @worker_count.times.map do
+      w = @klass.new(*args)
+      w.register
+      @worker_queue << w
+      w
+    end
+
+    @events_received = Concurrent::AtomicFixnum.new(0)
+  end
+
+  def config_name
+    @klass.config_name
+  end
+
+  def register
+    @workers.each {|w| w.register}
+  end
+
+  def multi_receive(events)
+    @events_received.increment(events.length)
+
+    worker = @worker_queue.pop
+    begin
+      worker.multi_receive(events)
+    ensure
+      @worker_queue.push(worker)
+    end
+  end
+
+  def do_close
+    @logger.debug("closing output delegator", :klass => self)
+
+    @worker_count.times do
+      worker = @worker_queue.pop
+      worker.do_close
+    end
+  end
+
+  def events_received
+    @events_received.value
+  end
+
+  def busy_workers
+    @worker_queue.size
+  end
+end end
--- a/Gemfile
+++ b/Gemfile
@@ -4,17 +4,13 @@
 source "https://rubygems.org"
 gem "logstash-core", "3.0.0.dev", :path => "./logstash-core"
 gem "logstash-core-event", "3.0.0.dev", :path => "./logstash-core-event"
-# gem "logstash-core-event-java", "3.0.0.dev", :path => "./logstash-core-event-java"
 gem "file-dependencies", "0.1.6"
 gem "ci_reporter_rspec", "1.0.0", :group => :development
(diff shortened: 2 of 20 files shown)


andrewvc:
I've done some benchmarking parsing apache logs with the cloud benchmarker and indexing them to ES. The results are pretty interesting. The upshot is, for this benchmark, **the ng_pipeline was 28.67% faster than master**. 

This is a preliminary benchmark of just apache log file parsing (with grok / geoip / useragent) with the cloud benchmarker. I anticipate other benchmarks to act differently. The comments below discuss its performance on this task.

To start with the new pipeline seems quite a bit faster BUT the tuning is all different. There are three parameters to tune the new pipeline:

* `-w` number of workers
* `-b` batch size
* `-u` batch pop latency

The key takeaways were:

* Have more takeaways than cores for either pipeline. `-w 8` seemed to be the sweet spot.
* The `-u` setting doesn't change much if anything. We should leave it in, just in case, but the default of 5ms works fine.
* Tuning `-b` is very important! For the apache logs a larger batch size yielded significant gains. We may strongly want to consider size based dynamic batches to help optimize this. Currently its hard to know the size of an event, but with the persistent pipeline that should be easier.
* Adjusting master's `SizedQueue` size did nothing

pipeline_benchmarks_-_google_sheets

andrewvc:
I should mention that watching these benchmarks in htop it seemed that the classic pipeline didn't saturate the cores as effectively and used more `system` time. I think there are some constant time operations in it that impede thread liveness.

colinsurprenant:
I just completed a round of tests & benchmarks using 2 different configs and sets of inputs. I will put my results in a table but before I wanted to report I problem I encountered using many workers and bigger batch size so that other can see if they can reproduce.

I am seeing the following error output and getting various output delays (output completely stops for seconds)

```
Concurrent::TimeoutError {:class=>"Concurrent::TimeoutError", :backtrace=>nil, :level=>:warn}
```

The config & data used is in https://gist.github.com/colinsurprenant/8143a0dfd30830d57b03 and the command line where I started noticing these errors is with 

```sh
$ while true; do cat syslog.dat; done | bin/logstash -w 8 -b 8000 -f syslog.conf
```

using `syslog.dat` and `syslog.conf` in the gist ^^

jordansissel:
@andrewvc your most recent patch (the while loop timer) LGTM, butI haven't tested it.

ph:
@andrewvc what was the issue with the task?

colinsurprenant:
@andrewvc perfect! I much prefer this simpler flushing implementation, plus it uses `Stud.stoppable_sleep` which is preferable for handling the shutdown sequence. 

I haven't been able to reproduce the `Concurrent::TimeoutError` exception :thumbsup: 

Before I post my benchmarks results, I realize that the high workers + bigger batch runs might have been slowed down by the heap memory pressure, so I will re-run them and post results!

andrewvc:
Thanks @colinsurprenant ! In retrospect, the benchmarks I ran above probably saw declining numbers with greater batch sizes due to memory pressure more than CPU. That said, I think I'll hold off on running further benchmarks till a future date since the main point is that the new pipeline is much more efficient than the old one given similar resources.

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def events_received
+    @events_received.value
+  end
+
+  def busy_workers
+    @worker_queue.size
+  end
+end end
```
ph: woa cleaner +1

colinsurprenant:
Here are my results.

- MBP 13r, 2.8 GHz Intel Core i7, 16GB
- Java 8 1.8.0_45-b14
- logstash config:
```
input {
  stdin {
    codec => json_lines
  }
}

filter {
  mutate {
    add_field => ["[fields][date]",  "%{+YYYY-MM-dd:HH:mm:ss}"]
    add_field => ["[fields][mixed]",  "foo %{+YYYY-MM-dd} %{foo}"]
    add_field => ["[fields][foo]",  "%{foo}"]
    add_field => ["[fields][bar]",  "%{bar}"]
    remove_field => ["foo"]
    remove_field => ["bar"]
  }
}

output {
}
```
- base command line, adjust `-w` and `-b` per bench:
```sh
$ yes '{"foo":"testfoo","bar":"testbar"}' | LS_HEAP_SIZE=4g bin/logstash -w X -b Y -f bench.conf | pv -Wlart > /dev/null
```

- results for a 1 minute run, the average k lines per sec (or kTPS). Note that only a single run has been measured for each scenario, and there is slight variations from run ro run with same parameters.

screen shot 2015-12-15 at 1 30 56 pm

colinsurprenant:
LGTM

ph:
did another run LGTM, nice work @andrewvc

andrewvc:
Thanks for putting in so much time to review and find bugs in this PR @colinsurprenant and @ph !

colinsurprenant:
one observation: as we can see, these isn't much a difference in the previous performance numbers, for the same workers and default batch the ng-pipeline is slightly faster. I will not post all the results but I performed a few similar tests using the config at https://gist.github.com/colinsurprenant/8143a0dfd30830d57b03 and the performance gain is higher, for example from **5.5k/s** in master to **6.6k/s** with ng-pipeline which is a lot more significative.

In the previous benchmarks, the input side had more to do because of the JSON parsing by the codec and the inputs are not parallelized so I think that the input was the actual choke point while in the other tests, the input is straight stdin line input and everything is done in the filters so the hard work is effectively in the filter stage which benefits the ng batching etc.

PR #4341, opened 2015-12-11
using symbols in Event data does not make sense

colinsurprenant:
Changed "illegal" object from symbol to string since in Java symbols get converted to string and symbols are not valid objects to start with as Event data.

Relates to #4191


suyograo:
LGTM

colinsurprenant:
since this is so trivial, will go ahead and merge.

colinsurprenant:
merged in master and 2.x

PR #4342, opened 2015-12-12
[wip] reload config from files

jsvd:
add ability to detect changes in the configuration file and restart the pipeline with the changes.

If -f config_file is given and -r is enabled then the agent will check the file every 5 seconds and:
* if there's a difference, stop the existing pipeline, create and start a new one
  - if there's an error while creating the new pipeline, the previous configuration is used again
* if there's no difference, nothing happens
Sending a SIGHUP to the logstash also triggers a configuration reload

**TODO**

~~Since the shutdown controller doesn't yet handling multiple pipelines, it is disabled for now~~
pass runner pipeline settings to the pipelines, through agent

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -82,19 +82,13 @@ class LogStash::Agent
     return unless @pipelines[id]
     @logger.warn("stopping pipeline", :id => id)
     @pipelines[id].shutdown do
-      # TODO uncomment once shutdown controller can be stopped
-      #LogStash::ShutdownController.start(@pipelines[id])
+      LogStash::ShutdownController.start(@pipelines[id], @pipeline_threads[id])
     end
     @pipeline_threads[id].join
   end
 
   def shutdown_pipelines
-    @pipelines.each do |id, pipeline|
-      stop_pipeline(id)
-      #pipeline.shutdown do
-      #  ::LogStash::ShutdownController.start(pipeline)
-      #end
-    end
+    @pipelines.each { |id, _| stop_pipeline(id) }
   end
 
   def running_pipelines?
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -23,6 +23,7 @@ class LogStash::Agent
     @auto_reload = params[:auto_reload]
     @pipeline_settings = params[:pipeline_settings]
     @reload_interval = params[:reload_interval] || 5 # seconds
+    @upgrade_mutex = Mutex.new
   end
 
   def execute
@@ -52,7 +53,7 @@ class LogStash::Agent
     if valid_state?(new_state)
       if new_state?(@state, new_state)
         @logger.warn("fetched new state. upgrading..", :state => new_state)
-        upgrade_state(new_state)
+        @upgrade_mutex.synchronize { upgrade_state(new_state) }
       else
         @logger.debug("same state, ignoring..")
       end
@@ -93,7 +94,9 @@ class LogStash::Agent
   end
 
   def running_pipelines?
-    @pipeline_threads.select {|_, pipeline| pipeline.alive? }.any?
+    @upgrade_mutex.synchronize do
+      @pipeline_threads.select {|_, pipeline| pipeline.alive? }.any?
+    end
   end
 
   # Override the methods below if you're implementing your own agent
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -12,26 +12,35 @@ LogStash::Environment.load_locale!
 class LogStash::Agent
   attr_reader :logger
 
-  def initialize(logger)
+  def initialize(logger, config_string, config_path)
     @logger = logger
     @pipelines = {}
-  end
-
-  def config_valid?(config_str)
-    begin
-      # There should be a better way to test this ideally
-      LogStash::Pipeline.new(config_str)
-    rescue Exception => e
-      e
-    end
+    @pipeline_threads = {}
+    @state = ""
+    @config_loader = LogStash::Config::Loader.new(@logger, false)
+    @config_str = config_string
+    @config_path = config_path
   end
 
   def execute
-    # Make SIGINT/SIGTERM shutdown the pipeline.
     sigint_id = trap_sigint()
     sigterm_id = trap_sigterm()
 
-    @pipelines.each {|_, p| p.run } # blocking operation. works now because size <= 1
+    @logger.info "starting agent"
+    loop do
+      new_state = fetch_state
+      if valid_state?(new_state)
+        if new_state?(@state, new_state)
+          @logger.info "new state: #{new_state}"
+          upgrade_state(new_state)
+        else
+          @logger.debug("same state, ignoring..")
+        end
+      else
+        @logger.error("invalid state #{new_state}")
+      end
+      sleep 10
+    end
     return 0
   rescue => e
     @logger.fatal I18n.t("oops", :error => e)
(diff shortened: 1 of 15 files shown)


jsvd:
example of changing the config file from:

```
input { generator { count => 10000 } }
filter { sleep { time => 0.5 } }
output { stdout { codec => dots } }
```

```
input { generator { count => 10000 } }
filter { sleep { time => 0.5 } }
output { stdout { codec => plain } }
```

you can see the moment where the changes are detected, the pipeline is gracefully shutdown and the new one is created:

```
% bin/logstash -f config --verbose
Creating new agent {:class=>LogStash::Agent, :level=>:info}
starting agent {:level=>:info}
new state: input { generator { count => 10000 } }
filter { sleep { time => 0.5 } }
output { stdout { codec => dots } }
{:level=>:info}
starting pipeline {:id=>"base", :level=>:info}
Settings: Default filter workers: 2
Worker threads expected: 2, worker threads started: 2 {:level=>:info}
Pipeline started {:level=>:info}
Logstash startup completed
................................................................................new state: input { generator { count => 10000 } }
filter { sleep { time => 0.5 } }
output { stdout { codec => plain } }
{:level=>:info}
stopping pipeline {:id=>"base", :level=>:info}
.......................Pipeline shutdown complete. {:level=>:info}
Logstash shutdown completed
starting pipeline {:id=>"base", :level=>:info}
Settings: Default filter workers: 2
Worker threads expected: 2, worker threads started: 2 {:level=>:info}
Pipeline started {:level=>:info}
Logstash startup completed
2015-12-12T00:44:49.296Z Joaos-MBP.lan Hello world!
2015-12-12T00:44:49.296Z Joaos-MBP.lan Hello world!
2015-12-12T00:44:49.296Z Joaos-MBP.lan Hello world!
2015-12-12T00:44:49.296Z Joaos-MBP.lan Hello world!
2015-12-12T00:44:49.296Z Joaos-MBP.lan Hello world!
2015-12-12T00:44:49.297Z Joaos-MBP.lan Hello world!
....
```

logstash-core/lib/logstash/agent.rb
```diff
@@ -10,80 +10,130 @@
 LogStash::Environment.load_locale!
 
 class LogStash::Agent
-  attr_reader :logger
+  attr_reader :logger, :state
 
-  def initialize(logger)
-    @logger = logger
+  def initialize(params)
```
jsvd: I'm not a fan of having this global `params` argument but different agents can have different sources of events...
any suggestions here @andrewvc ?

jsvd:
@andrewvc review welcome for this

logstash-core/lib/logstash/agent.rb
andrewvc: I'm not a huge fan either, but I think its appropriate for this stage. Once we have more use cases we can determine if another approach makes more sense.

logstash-core/lib/logstash/agent.rb
```diff
@@ -10,80 +10,125 @@
...
+
+  def valid_state?(new_state)
+    new_state.is_a?(String)
+  end
+
+  def new_state?(old_state, new_state)
+    old_state != new_state
+  end
+
+  def clean_state
```
andrewvc: Maybe this should be a constant? `CLEAN_STATE`

logstash-core/lib/logstash/agent.rb
```diff
@@ -10,80 +10,125 @@
...
-  attr_reader :logger
+  attr_reader :logger, :state
 
-  def initialize(logger)
-    @logger = logger
+  def initialize(params)
+    @logger = params[:logger]
     @pipelines = {}
+    @pipeline_threads = {}
+    @state = clean_state
```
andrewvc: This is confusing because its not a `state` like `:running`, but a config string. I'd like a name that better describes that. Maybe `current_config`?

logstash-core/lib/logstash/agent.rb
```diff
@@ -10,80 +10,125 @@
...
-      LogStash::Pipeline.new(config_str)
-    rescue Exception => e
-      e
+  def execute
+    @thread = Thread.current
+    @logger.info("starting agent", :state => @state)
+
+    if @auto_reload
+      Stud.interval(@reload_interval) do
+        break unless clean_state? || running_pipelines?
```
andrewvc: `break unless clean_state? || running_pipelines?` takes a minute to understand.

`reload_state! if empty_config? || running_pipelines?` is much more comprehensible IMHO

Since this is duplicated in later lines it may be even better to refactor this to:

`reload_state! if config_needs_reload?`

logstash-core/lib/logstash/agent.rb
```diff
@@ -10,80 +10,125 @@
...
 
-  # Emit a failure message and abort.
-  def fail(message)
-    raise LogStash::ConfigurationError, message
-  end # def fail
+  def stop_pipeline(id)
+    return unless @pipelines[id]
+    @logger.warn("stopping pipeline", :id => id)
+    @pipelines[id].shutdown do
+      LogStash::ShutdownController.start(@pipelines[id], @pipeline_threads[id])
```
andrewvc: Just realizing that "start the shutdown controller is confusing". Maybe we should have named it `LogStash::ShutdownController#begin_shutdown`

logstash-core/lib/logstash/agent.rb
```diff
@@ -10,80 +10,125 @@
...
+    @pipelines.each { |id, _| stop_pipeline(id) }
   end
 
-  def trap_sigterm
-    Stud::trap("TERM") do
-      @logger.warn(I18n.t("logstash.agent.sigterm"))
-      shutdown_pipelines
-    end
+  def running_pipelines?
+    @pipeline_threads.select {|_, pipeline| pipeline.alive? }.any?
```
andrewvc: Maybe call `.values` instead of using `_`. Totally optional, I'll leave it to you to make the choice here.

logstash-core/lib/logstash/agent.rb
```diff
@@ -10,80 +10,125 @@
...
-        Thread.new(@logger) {|logger| sleep 5; logger.warn(I18n.t("logstash.agent.slow_shutdown")) }
-        @interrupted_once = true
-        shutdown_pipelines
-      end
-    end
+  # Override the methods below if you're implementing your own agent
+  def upgrade_state(new_state)
+    stop_pipeline("base")
+    add_pipeline("base", new_state, @pipeline_settings)
+  rescue => e
```
andrewvc: Should we be rescuing `Exception` after the work that @colinsurprenant and @guyboertje did?

andrewvc:
Looking good overall! Left some comments.

logstash-core/lib/logstash/agent.rb
jsvd: different agents will have different notions of clean_state, for the default agent is "", but for another can be a `{"version" => 0}` hashmap for example

logstash-core/lib/logstash/agent.rb
jsvd: Some documentation can set the right expectations, as this is more of an internal representation of the agent state, and not the current status or "lifetime phase".

Here `state` means the same as a node state or cluster state in elasticsearch, which is a document string or hash that represents the internal state.
This can also be interpreted as the State in erlang's OTP framework: any kind of data structure that represents state.

My problem with calling this current_config is that another agent can have much more metadata about its state than just the config itself (versioning, timestamps, last known error, etc)

what do you think of just adding a comment setting this expectation?

logstash-core/lib/logstash/agent.rb
jsvd: I'm trying to avoid adding the concept of "config" in the core of the agent's api, but I agree this is not very readable. will try to improve :+1:

logstash-core/lib/logstash/agent.rb
jsvd: I see..maybe we can rename ShutdownController to ShutdownWatcher? `begin_shutdown` isn't fair either as it doesn't actually do a shutdown, just monitors a pipeline during shutdown and forces exit in some scenarios.

logstash-core/lib/logstash/agent.rb
jsvd: :+1:

logstash-core/lib/logstash/agent.rb
andrewvc: I think a comment helps. Ideally that comment would include the contents of your reply here and mention that other agents might use something richer.

logstash-core/lib/logstash/agent.rb
andrewvc: The empty config thing was assuming  state was changed to config. Since that won't be the case feel free to adjust the var names as needed.

logstash-core/lib/logstash/agent.rb
andrewvc: Well, this isn't a high priority change for me just FYI. But I with `ShutdownWatcher`, much more clear!

logstash-core/lib/logstash/agent.rb
andrewvc: makes sense, disregard then :)

logstash-core/lib/logstash/agent.rb
jsvd: what do you think about it now? I think the if vs unless makes it simpler to read, but makes it more verbose

logstash-core/lib/logstash/agent.rb
jsvd: +1 will open an issue for this refactor, it can target master directly

andrewvc:
@jsvd while I did notice that you merged in the commits for #4350 this branch is not yet rebased on master so it doesn't include any ng_pipeline stuff. There are significant conflicts there. I'm glad to pair with you at some point to make the merge as pain-free as possible.

logstash-core/lib/logstash/agent.rb
jsvd: @guyboertje what do you advise here?

andrewvc:
LGTM pending @guyboertje 's input on the rescue exception

andrewvc:
@jsvd I rebased `feature/management` on master to bring us closer to current. That just broke the merge-ability of this commit. I hope that wasn't the wrong thing to do, but I felt that for a feature branch rebases back to master should just be pushed directly.

andrewvc:
@jsvd would you be able to replace your code with the rebased version I have here? https://github.com/elastic/logstash/pull/4372

There was a nasty mess of merge conflicts after I rebased feature/management off master.

jsvd:
closed in favor of 4372

Issue #4353, opened 2015-12-15
We need to be able to set a node name from the command line the name of the logstash instance


ph:
This name will be used for the agent and also from the metrics collection.

After discussing with @andrewvc we think that `-n` of `--name` would be a good name. The name will be send to Agent for now and show in the log.

If no name is given we will fallback to the current hostname.

andrewvc:
@ph will this also include a per-run UUID?

ph:
@andrewvc what do you mean by per run UUID?

andrewvc:
Errr, in the design discussions @jsvd and I have had there would be two IDs:

1. A name, that is overridable by the user and is an arbitrary string. Defaults to network hostname.
2. A per-run UUID. This is useful in separating out actual runs. Also useful in that if you migrate the host to a new box you are at least guaranteed a UUID change.

The name is more generally useful, but the UUID has its utility as well. Thoughts @ph?

ph:
@andrewvc OK and this UUID is valid for the current instance of Logstash, when logstash restart we will get a new one?

purbon:
@ph @andrewvc just a proposal, can we make more explicit this is the node name? only with ```--name``` is not clear to me which kind of name we are setting here. makes sense to you?

ph:
@purbon I am ok with `--node-name` @andrewvc WDYT ?

PR #4356, opened 2015-12-15
Allow the Agent to define some identifiers

ph:
This PR allow the user to set the name of the logstash instance with the
`-n` or `--name` option, if no name is provided the agent will fallback
to the current hostname of the machine.

This PR also add the `node_uuid` to the agent which is unique between
each new instance of the agent, it will lazy generate a UUID v4 on
invocation.

Fixes: #4353

### DIFF
--- a/logstash-core/spec/logstash/runner_spec.rb
+++ b/logstash-core/spec/logstash/runner_spec.rb
@@ -24,8 +24,8 @@ describe LogStash::Runner do
       let(:args) { ["-e", ""] }
 
       it "should execute the agent" do
-        expect(subject.agent).to receive(:add_pipeline).once
-        expect(subject.agent).to receive(:execute).once
+        expect_any_instance_of(LogStash::Agent).to receive(:add_pipeline).once
+        expect_any_instance_of(LogStash::Agent).to receive(:execute).once
         subject.run(args)
       end
     end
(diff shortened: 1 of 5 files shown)


andrewvc:
@ph would LGTM except you have a failing spec! Ping me when the build passes :)

ph:
o_O oh oh, let me verify that

logstash-core/spec/logstash/runner_spec.rb
```diff
@@ -24,8 +24,8 @@ def run(args); end
       let(:args) { ["-e", ""] }
 
       it "should execute the agent" do
-        expect(subject.agent).to receive(:add_pipeline).once
-        expect(subject.agent).to receive(:execute).once
+        expect_any_instance_of(LogStash::Agent).to receive(:add_pipeline).once
+        expect_any_instance_of(LogStash::Agent).to receive(:execute).once
```
ph: This expectation was needed to change, if we setup the `Agent` in the `Runner#initialize` method the `node_name` value wasn't not available before the `#run` is executed.

ph:
@andrewvc ping tests pass

andrewvc:
@ph LGTM

jsvd:
will we ever need a composed node_name and node_uuid concept?
like `agent.full_name = "#{node_name}.#{node_uuid}"`?

anyway LGTM

andrewvc:
@jsvd I don't believe we'll ever need that. The UUID should be globally unique.

andrewvc:
@jsvd if we want we could display that to the user and only actually use the UUID, but that's more a UX issue than internals which I think this PR is concerned with.

ph:
@andrewvc @jsvd This should only be merged to master correct?

andrewvc:
@ph yep!

PR #4358, opened 2015-12-15
[WIP] Metrics/structured data

ph:
Still a work in progress.


ph:
This PR miss an average type, this is needed to have a meaning full metric of the time spend inside a specific plugin.

logstash-core/lib/logstash/inputs/metrics.rb
```diff
@@ -27,8 +32,8 @@ def stop
       LogStash::Instrument::Collector.instance.delete_observer(self)
     end
 
-    def update(time, snapshot)
-      @logger.debug("Metrics input: received a new snapshot", :snapshot => snapshot, :event => snapshot.to_event) if @logger.debug?
+    def update(time_of_snapshot, snapshot)
```
purbon: why this renaming? I guess a better name might be ```timestamp```. I mean this because now you made the var dependant on the content of the second parameter, if we call this ```timestamp```, this var would have a general meeting, time of the update. Also if this is the time of the snapshot, can this be included in the snapshot ?

purbon:
@ph I know is WiP, just dropping here a few comments :smiling_imp:

logstash-core/lib/logstash/inputs/metrics.rb
```diff
@@ -39,7 +44,10 @@ def update(time, snapshot)
       #   - We can use a synchronization mechanism between the called thread (update method)
       #   and the plugin thread (run method)
       #   - How we handle back pressure here?
-      @queue << snapshot.to_event
+      #   - one snashot should be only one event
```
purbon: this comment line is wrong according to the code you just wrote here.

logstash-core/lib/logstash/inputs/metrics.rb
```diff
@@ -39,7 +44,10 @@ def update(time, snapshot)
...
+      snapshot.to_events.each do |event|
```
purbon: I wonder if makes sense to make this call somehow async? now this might mean a ton of work from the snapshot to transform to events and then iterate to inject each new event in the queue. what do you think?

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -1,79 +1,109 @@
...
   class Collector
-    # TODO: move to a timer based  API with concurrent ruby timer class
-    # Allow to accept an external flush
-    # When the flush is done we should not record any new metric
     include LogStash::Util::Loggable
     include Observable
     include Singleton
 
-    SNAPSHOT_ROTATION_TIME = 1 #seconds
+    SNAPSHOT_ROTATION_TIME = 1 # seconds
```
purbon: I always have troubles with the units and vars/constants? any idea on how to make this less error prone?

purbon: I would also recommend to freeze this them.

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -1,79 +1,109 @@
...
     def push(*args)
-      snapshot.push(*args)
-    end
+      namespaces_path, key, type, other = args
 
-    def self.snapshot_rotation_time=(time)
-      @snapsho_rotation_time = ime
+      begin
+        metric = @metric_store.fetch_or_store(namespaces_path, key) do
+          concrete_class(type).new(namespaces_path, key)
```
purbon: I wonder about this kind of construct ```concrete_class(type).new(namespaces_path, key)``` what are the odds of this being a problem in terms of performance? what do you think?

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -1,79 +1,109 @@
...
-      end
-
-      @current_snapshot
+    private
+    # Use the string to generate a concrete class for this metrics
+    # 
+    # @param [String] The name of the class
+    # @raise
+    def concrete_class(type)
+      LogStash::Instrument::MetricType.const_get(type.capitalize)
```
purbon: what happen in the type is something like ```MetricType``` capitalization might not be enough, right?

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -1,79 +1,109 @@
...
+    #
+    # `#update(created_at, metric_store)`
     def publish_snapshot
-      changed
-      notify_observers(Concurrent.monotonic_time, @current_snapshot) 
-    end
-
-    def update_last_rotation
-      @last_rotation = Concurrent.monotonic_time
+      created_at = Time.now
```
purbon: any reason for this change? I mean I get boths pros and cons of both, just wondering why you started why monotonic and now change :-P

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -0,0 +1,94 @@
+# encoding: utf-8
+require "concurrent"
+require "logstash/event"
+
+module LogStash module Instrument
+  # The Metric store the data structure that make sure the data is
+  # saved in a retrievable way, this is a wrapper around multiples ConcurrentHashMap
+  # acting as a tree like structure.
+  class MetricStore
+    class ConcurrentMapExpectedError < Exception; end
```
purbon: I see this exception too implementation dependant, can we make it more Metric store related?

logstash-core/lib/logstash/inputs/metrics.rb
ph: Actually we could create a Snapshot which basically wrap the metric and the time WDYT?

ph: This would make the update method receive only one thing.

logstash-core/lib/logstash/inputs/metrics.rb
ph: You are right but until the elasticsearch know how to handle `Logstash::Event` of type snapshot, I have to make them unique so it actually create meaningful document in ES.

logstash-core/lib/logstash/instrument/collector.rb
ph: +1 for freezing, not sure If I have a trick to reduce the errors :(

logstash-core/lib/logstash/inputs/metrics.rb
ph: Agree this could be a costly operation if we have a large quantity of metrics, I think we need to benchmark this to find a best approach ?

logstash-core/lib/logstash/instrument/collector.rb
ph: Thats a really good point, In the first iteration I went with a switch case,
What we could do we could cache the resolution in an hash so this operation would become o(1) after the first run. But on the other hand it probably do a hash lookup behind the scene in the class hierarchy, benchmark needed?

logstash-core/lib/logstash/instrument/collector.rb
ph: Correct, but since we are controlling the primitive types I don`t see it as a problem right now?

logstash-core/lib/logstash/instrument/collector.rb
ph: I have to benchmark it, I though it was premature optimization and made the code harder to read :(

logstash-core/lib/logstash/instrument/metric_store.rb
ph: +1, what do you think about `NamespaceExpectedError` ?

ph:
@purbon really good comments thanks for that run!

logstash-core/lib/logstash/instrument/collector.rb
ph: Actually the `concrete_class` method might belong in `LogStash::Instrument::MetricType` ? WDYT

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -0,0 +1,94 @@
...
+    
+    # This method allow to retrieve values for a specific path,
+    #
+    # Also support `*` as a globbing path
+    #
+    # @param [Array] The path where values should be located
+    # @return nil if the values are not found
+    def get(paths)
+      fetch_recursively(path)
+    end
```
ph: Need implementation, need to sync with your @purbon

logstash-core/lib/logstash/instrument/collector.rb
suyograo: Can we name this `SNAPSHOT_ROTATION_TIME_SECS` or something to include units?

ph: Or we could use the a `time + unit` notation, in this case it would be `1s`, this is similar to what ES and beats is currently doing.

ph:
added a #get, changed also how we instantiate different metric type, I have opted to used a predefined hash, it cannot be faster than that.

ph:
@purbon actually `Time.now` is faster

```
~/e/m/monotonic ❯❯❯ ruby fast.rb                                                                                                                                                                              ⏎
            Time.now   158.399k i/100ms
Concurrent.monotonic    85.444k i/100ms
-------------------------------------------------
            Time.now      4.215M (± 5.8%) i/s -     21.067M
Concurrent.monotonic      1.726M (± 6.2%) i/s -      8.630M

Comparison:
            Time.now:  4215079.2 i/s
Concurrent.monotonic:  1726232.1 i/s - 2.44x slower
```

logstash-core/lib/logstash/inputs/metrics.rb
purbon: what I understand from like 48 on is that from one snapshot, so a container of metric_store you generate multiple events, so the line in 47 is misleading. 

Another thing is that the logic to convert to an event should be in the input, not in the snapshot, at my understanding the snapshot is just a container, so should not be aware of format conversions, what do you think?

logstash-core/lib/logstash/instrument/collector.rb
purbon: +1 on time plus units, as a high school teacher told me, you should be writing units always somehow :-P

logstash-core/lib/logstash/instrument/collector.rb
purbon: agree with it, we can always go back if necessary. but also looks reasonable to me to think getting time should be fast.

logstash-core/lib/logstash/instrument/metric_store.rb
purbon: I don't understand the ```Expected``` part ? why is expected? if expected can you do anything to prevent it?

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -1,79 +1,100 @@
...
+    # its the job of the collector to update the store with new metric
+    # of update the metric
+    #
+    # If there is a problem with the key or the type of metric we will record an error 
+    # but we wont stop processing events, theses errors are not considered fatal.
+    # 
     def push(*args)
-      snapshot.push(*args)
-    end
+      namespaces_path, key, type, other = args
```
purbon: what is other? is not clear what kind of data it can hold.

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -1,79 +1,100 @@
...
 
-    def self.snapshot_rotation_time=(time)
-      @snapsho_rotation_time = ime
+      begin
+        metric = @metric_store.fetch_or_store(namespaces_path, key) do
+          LogStash::Instrument::MetricType.create(type, namespaces_path, key)
+        end
+        metric.execute(*other)
```
purbon: as shared ealier, the parameter kind of is not clear. Can we workout a clear name?

purbon:
I see we have here, also in some other parts of LS some configuration variables, like in our case for example ``` SNAPSHOT_ROTATION_TIME_TIMEOUT ``` I think we should make them configurable, for example with another config flie. I don't say we should do it in this PR, but what do you think?

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -1,79 +1,100 @@
...
 
-    def self.snapshot_rotation_time
-      @snapshot_rotation_time || SNAPSHOT_ROTATION_TIME
+    # Monitor the `Concurrent::TimerTask` this update is triggered on every successful or not
+    # run of the task, this method will record the execution in the log
+    #
+    # @param [Time] Time of execution
+    # @param [result] Result of the execution
+    # @param [Exception] Exception
+    def update(time_of_execution, result, exception)
```
purbon: I think the name of this method is wrong, or at less incomplete. it gets confusing with the behaviour it have.

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -4,36 +4,38 @@
...
 
-    def namespace(key)
-      Metric.new(collector, merge_keys(key.to_sym))
-    end
-
-    # I think this should have his own values.
     def time(key, &block)
+      validate_key!(key)
       if block_given?
         start_time = Concurrent.monotonic_time
```
purbon: should we change this also to use ```Time.now```?

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -0,0 +1,107 @@
...
+      to_events_recursively(@store).flatten
+    end
+
+    private
+    def get_recursively(key_paths, map)
+      key_candidate = key_paths.shift
+
+      if key_paths.empty?
+        return map[key_candidate]
+      elsif (next_map = map[key_candidate]).is_a?(Concurrent::Map)
```
purbon: I know this is personal taste, but always found this kind of lines  doing too much things at once, can we break it somehow? it makes reading the code easier, at less to me.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -0,0 +1,107 @@
...
+    # @param [Array] The path where the values should be located
+    # @param [Object] The default object if the value is not found in the path
+    # @return [Object] Return the new_value of the retrieve object in the tree
+    def fetch_or_store(namespaces, key, default_value = nil)
+      fetch_or_store_namespaces(namespaces).fetch_or_store(key, block_given? ? yield(key) : default_value)
+    end
+    
+    # This method allow to retrieve values for a specific path,
+    #
+    # Also support `*` as a globbing path
```
purbon: I can not see how this method support ```*``` for searching all sub trees? I'm missing something?

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -0,0 +1,107 @@
...
+    #
+    # @param [Array] The path where values should be located
+    # @raise [ConcurrentMapExpected] Raise if the retrieved object isn't a `Concurrent::Map`
+    # @return [Concurrent::Map] Map where the metrics should be saved
+    def fetch_or_store_namespaces(namespaces_path)
+      path_map = fetch_or_store_namespace_recursively(@store, namespaces_path)
+      
+      # This mean one of the namespace and key are colliding
+      # and we have to deal it upstream.
+      unless path_map.is_a?(Concurrent::Map)
```
purbon: Can you explain to me when this can happen? is this for something like Map->value->Map ? should not be possible right?

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    # Recursively fetch or create the namespace paths through the `MetricStove`
+    # This algorithm use an index to known which keys to search in the map.
+    # This doesn't cloning the array if we want to give a better feedback to the user
+    #
+    # @param [Concurrent::Map] Map to search for the key
+    # @param [Array] List of path to create
+    # @param [Fixnum] Which part from the list to create
+    #
+    def fetch_or_store_namespace_recursively(map, namespaces_path, idx = 0)
```
purbon: not sure I understand the purpose of the ``` idx``` param? I think is also not used, right?

logstash-core/lib/logstash/instrument/metric_type/base.rb
```diff
@@ -0,0 +1,16 @@
...
+module LogStash module Instrument module MetricType
+  class Base
+    def initialize(namespaces, key)
+      @namespaces = namespaces
+      @key = key
+    end
+
+    private
+    def type
+      @type ||= self.class.name.split("::").last.downcase
```
purbon: I'm seeing a some class modification methods, like the one we have somewhere else about creating a class from a name. I'm starting to see we should group them in some kind of helper class, what do you think?

logstash-core/spec/logstash/inputs/metrics_spec.rb
```diff
@@ -0,0 +1,32 @@
+# encoding: utf-8
+require "logstash/inputs/metrics"
+require "spec_helper"
+
+describe LogStash::Inputs::Metrics do
+  let(:queue) { [] }
+
+  describe "#run" do
+    it "should register itself to the collector observer" do
+      expect_any_instance_of(LogStash::Instrument::Collector).to receive(:add_observer).with(subject)
```
purbon: I know this is the easy way to do it, and works, but could we figure it out a way to not use ```expect_any_instance_of```? the usage of this method is highly discouraged for the same rspec developers.

logstash-core/spec/logstash/inputs/metrics_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  describe "#stop" do
+    it "should remove itself from the the collector observer" do
+      expect_any_instance_of(LogStash::Instrument::Collector).to receive(:delete_observer).with(subject)
+      t = Thread.new { subject.run(queue) }
+      sleep(0.1) # give a bit of time to the thread to start
+      subject.stop
+    end
+
+    it "should unblock the input" do
+      t = Thread.new { subject.run(queue) }
```
purbon: I see this part being a pattern in this test, can we might be wrap this behaviour in a helper method?

logstash-core/spec/logstash/instrument/collector_spec.rb
```diff
@@ -0,0 +1,47 @@
+# encoding: utf-8
+require "logstash/instrument/collector"
+require "spec_helper"
+
+describe LogStash::Instrument::Collector do
+  subject { LogStash::Instrument::Collector.instance }
+  describe "#push" do
+    let(:namespaces_path) { [:root, :pipelines, :pipelines01] }
```
purbon: s/pipelines01/pipeline01 is better ot use singular for the last value.

logstash-core/spec/logstash/instrument/collector_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    context "when there is a conflict with the metric key" do
+      let(:conflicting_namespaces) { [namespaces_path, key].flatten }
+
+      it "logs an error but dont crash" do
+        subject.push(namespaces_path, key, :counter, :increment)
+
+        expect(subject.logger).to receive(:error)
+          .with("Collector: Cannot record metric", hash_including({ :exception => instance_of(LogStash::Instrument::MetricStore::ConcurrentMapExpectedError) }))
```
purbon: might we break this line a bit? looks long to me.

logstash-core/spec/logstash/instrument/metric_store_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    it "support a block as argument" do
+      expect(subject.fetch_or_store(namespaces, key) { counter }).to eq(counter)
+    end
+  end
+
+  context "when the metric object exist in the namespace"  do
+    let(:new_counter) { LogStash::Instrument::MetricType::Counter.new(namespaces, key) }
+
+    before do
```
purbon: I would recommend using specific (:each or :all) labels, always found confusing relaying on defaults.

logstash-core/spec/logstash/instrument/metric_store_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    it "raise an exception" do
+      expect { subject.fetch_or_store(conflicting_namespaces, :new_key, counter) }.to raise_error(LogStash::Instrument::MetricStore::ConcurrentMapExpectedError)
+    end
+  end
+
+  describe "#get" do
+    let(:metric_events) {
+      [
+        [[:node, :sashimi, :pipelines, :pipeline01, :plugins, :"logstash-output-elasticsearch"], :event_in, :increment],
```
purbon: sorry for the question is irrelevant, but why the ```"``` for the ```logstash-output-eleasticsearch``` field?

logstash-core/spec/logstash/instrument/metric_store_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+        [[:node, :sashimi, :pipelines, :pipeline01, :plugins, :"logstash-output-elasticsearch"], :event_in, :increment],
+        [[:node, :sashimi, :pipelines, :pipeline01], :processed_events_in, :increment],
+        [[:node, :sashimi, :pipelines, :pipeline01], :processed_events_out, :increment],
```
purbon: can we add one with ```*```?

logstash-core/spec/logstash/instrument/metric_store_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+
+    it "retrieves end of branch values" do
+      metrics = subject.get(:node, :sashimi, :pipelines, :pipeline01, :plugins, :"logstash-output-elasticsearch")
+      expect(metrics.size).to eq(1)
+    end
+
+    it "retrieves branch" do
+      metrics = subject.get(:node, :sashimi, :pipelines, :pipeline01)
+      expect(metrics.size).to eq(3)
```
purbon: not sure I understand this should fetch 3 or 4? I'm confused between line 47 and 48.

logstash-core/spec/logstash/instrument/metric_store_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      # Lets add a few metrics in the store before trying to convert them
+      metric_events.each do |namespaces, metric_key, action|
+        metric = subject.fetch_or_store(namespaces, metric_key, LogStash::Instrument::MetricType::Counter.new(namespaces, key))
+        metric.execute(action)
+      end
+    end
+
+    it "converts all metric to `Logstash::Event`" do
+      events = subject.to_events
+      expect(events.size).to eq(2)
```
purbon: you should add an specific test for this. I see this being a pattern in your test here, you include a duplicated metrics event, and test some deduplication like his. I think is better to include a separated case as this is very important behaviour and here for example is mix with the event conversion.

logstash-core/lib/logstash/instrument/collector.rb
ph: it depends on the type of metric created, would `metric_type_params` be a better name?

logstash-core/lib/logstash/instrument/collector.rb
ph: I cannot change the method name is part of the `Concurrent::Task` Api, I can make the comment clearer? Or I can make a specific class with only that method?

logstash-core/lib/logstash/instrument/metric.rb
ph: yes + 1, I did the test yesterday after pushing the commit and didnt revisit it.

logstash-core/lib/logstash/instrument/metric_store.rb
ph: If you are talking about the inline assignment in the condition I am okay in breaking it up.

purbon: yes, the inline :smile:

logstash-core/lib/logstash/instrument/metric_store.rb
ph: Its used in the next line of this comment.

logstash-core/lib/logstash/instrument/metric_type/base.rb
ph: I don't create a class for the name anymore, see https://github.com/elastic/logstash/pull/4358/files#diff-e456278bc07b599d01b2c59ec369bbcdR6


I could move the `self.class.name.split("::")` in `logstash/util` ? WDTY

logstash-core/spec/logstash/inputs/metrics_spec.rb
ph: +1 I can use `LogStash::Instrument::Collector.instance`

logstash-core/spec/logstash/instrument/collector_spec.rb
ph: +1

logstash-core/spec/logstash/inputs/metrics_spec.rb
ph: +1

logstash-core/spec/logstash/instrument/collector_spec.rb
ph: +1

logstash-core/spec/logstash/instrument/metric_store_spec.rb
ph: I wanted to use a symbol and not a string, since all the other are symbols.

Symbols doesn't allow you to use `-` init unless you use quotes, without quote it think you are actually doing  subtraction.

```
jruby-1.7.23 :002 > :logstash-output-eleasticsearch
NameError: undefined local variable or method `output' for main:Object
```

with quotes:
```
jruby-1.7.23 :003 > :"logstash-output-eleasticsearch"
 => :"logstash-output-eleasticsearch"
```

logstash-core/spec/logstash/inputs/metrics_spec.rb
purbon: this works.

logstash-core/spec/logstash/instrument/metric_store_spec.rb
ph: I think you are this is maybe not obvious, but let me explain:
line 47 is creating the metric type counter. (value = 1)
 [[:node, :sashimi, :pipelines, :pipeline01, :plugins, :"logstash-output-elasticsearch"], :event_in, :increment]

line 48 is incrementing the previous counter. (Value = 2)

They are not different different unique metric.

logstash-core/spec/logstash/instrument/metric_store_spec.rb
ph: OK, make sense to split.

logstash-core/lib/logstash/instrument/collector.rb
ph: or `metric_type_options` ?

logstash-core/spec/logstash/instrument/metric_store_spec.rb
purbon: I understand, so then is clear this line is confusing, I would totally recommend simplifying this test to help better understanding.

logstash-core/lib/logstash/instrument/metric_store.rb
ph: I will fix that.

logstash-core/lib/logstash/instrument/metric_type/base.rb
purbon: works for me.

logstash-core/lib/logstash/instrument/metric_store.rb
ph: But when I did the code yesterday I found adding the `*` was yielding confusing results and wasnt making sense in my brain. 

With the current features, I don't when this could be useful to support a `*` ? Whats your opinion on that?

logstash-core/lib/logstash/instrument/collector.rb
ph: Time.now is faster, I assumed Concurrurent.monotonic_time was faster and this seems to be wrong.

https://github.com/elastic/logstash/pull/4358#issuecomment-165324127

logstash-core/lib/logstash/inputs/metrics.rb
ph: Yes, like discussed yesterday, I will do that in another PR, if its okay with you?

logstash-core/lib/logstash/instrument/collector.rb
ph: See this line: https://github.com/elastic/logstash/pull/4358/files#diff-b949194a3f253e60f88091a1ff603dd1R86

logstash-core/lib/logstash/inputs/metrics.rb
purbon: yes, works for me

logstash-core/lib/logstash/instrument/metric_store.rb
ph: It's unlikely to happen but still can happen.

I should never happen to plugins developers since their metrics instance will be correctly configured and scope for their plugin at runtime. its mostly act as a guards.

logstash-core/spec/logstash/instrument/metric_store_spec.rb
ph: I can remove it and make it a separate test.

logstash-core/lib/logstash/instrument/metric_store.rb
ph: I have clarified this over zoom.

logstash-core/spec/logstash/inputs/metrics_spec.rb
ph: how would you call this?

logstash-core/lib/logstash/instrument/metric_store.rb
purbon: saw it, sorry :+1:

purbon:
squash and merge :shipit:

purbon:
we can improve the pending parts later on.

purbon:
closing this as it was merged already

ph:
good catch :D

Issue #4371, opened 2015-12-18
[meta] failing plugins specs under master


colinsurprenant:
Here's the survey of the currently failing specs under master at 066e778f63109d9fde2b4f00a70a68106ebe0204

#### devutils pipeline mocking
- [x] https://github.com/elastic/logstash-devutils/issues/42

#### bad in-specs pipeline mocking
- [x] https://github.com/logstash-plugins/logstash-input-elasticsearch/issues/37
- [x] https://github.com/logstash-plugins/logstash-input-file/issues/86
- [x] https://github.com/logstash-plugins/logstash-input-pipe/issues/12
- [x] https://github.com/logstash-plugins/logstash-input-gelf/issues/31
- [x] https://github.com/logstash-plugins/logstash-input-ganglia/issues/15
- [x] https://github.com/logstash-plugins/logstash-input-generator/issues/12
- [x] https://github.com/logstash-plugins/logstash-input-tcp/issues/36

#### out-of-order batches?
- [x] https://github.com/logstash-plugins/logstash-output-file/issues/25

## unknown origin

I suspect this are actually related to https://github.com/elastic/logstash-devutils/issues/42

#### hanging specs
- [x] https://github.com/logstash-plugins/logstash-input-tcp/issues/37
- [x] https://github.com/logstash-plugins/logstash-input-syslog/issues/24

## other
- [x] https://github.com/logstash-plugins/logstash-codec-rubydebug/issues/8
- [x] https://github.com/logstash-plugins/logstash-input-log4j/issues/27

## solutions
- [x] short term fix proposal in #4373
- [ ] real refactor required in #4382

colinsurprenant:
I just added the devutils pipeline mocking problems issue.

colinsurprenant:
proposed short-term fix by @andrewvc https://github.com/elastic/logstash/pull/4373

colinsurprenant:
real "long-term" refactor required in #4382

colinsurprenant:
#4384 solves most issues. updated the description tasks.

colinsurprenant:
added https://github.com/logstash-plugins/logstash-input-log4j/issues/27 can either @andrewvc or @jsvd  take a look? not sure if related to changes introduced by your last pipeline modifications

colinsurprenant:
log4j fixed by https://github.com/logstash-plugins/logstash-input-log4j/pull/28  thanks @andrewvc

ph:
[...]

PR #4373, opened 2015-12-18
Better compatibility with specs that depend on output_func monkeypatc…

andrewvc:
This is a temporary fix until old logstash plugin specs can be refactored to not
mess with the pipeline code.

Should fix a lot of the issues in https://github.com/elastic/logstash/issues/4371


colinsurprenant:
@andrewvc I just worked on this and the correct fix is in #4384. Please feel free to review.

Issue #3701, opened 2015-08-05
USE_RUBY=1 rake test does not work


purbon:
Hi, 
  on master doing:

```
purbon-elastic% USE_RUBY=1 rake test:core        
rake aborted!
Bundler::GemNotFound: Could not find benchmark-ips-2.2.0 in any of the sources
/Users/purbon/work/logstash/lib/bootstrap/bundler.rb:53:in `setup!'
/Users/purbon/work/logstash/rakelib/test.rake:15:in `(root)'
Tasks: TOP => test:core => test:setup
(See full trace by running task with --trace)
```

does not work, having bundler complaining about missing gems that in reality are properly installed and reachable.

PR #4374, opened 2015-12-18
[WIP] DEVELOPING.md guidelines on the development workflow

colinsurprenant:
Guidelines for the logstash development workflow.

This is a WIP, please feel free to suggest, comment, amend, ...

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -1,0 +1,126 @@
+# Logstash Development Workflow
+
+This explains how to develop in the logstash core code-base. We will use to following fictional directories:
+
+- `/logstash/` is the dir containing your local fork clone of logstash
+- `/plugins/...` is the dir containing your local fork clones of logstash plugins or any other local private plugins
+
+## Setup
+
+Logstash uses [JRuby](http://jruby.org/) which gets embedded in the `vendor/jruby/` directory. It is **highly recommended** but not mandatory that you also use JRuby as your local Ruby interpreter and for this you should consider using a Ruby version manager such as [RVM](https://rvm.io/) or [rbenv](https://github.com/sstephenson/rbenv). It is possible to run the rake tasks and the `bin/` commands without having JRuby locally installed in which case the embedded JRuby will be used automatically.
+
+You will also need to have a JVM installed, we recommend using Java 1.8.
+
+`Rake` and `Bundler` should also be installed in your environment.
+
+Note that if you have a local JRuby installed you can force logstash to use your local JRuby instead of the embedded JRuby with the `USE_RUBY=1` environment variable.
+
+
+## Bootstrapping
+
+The logstash environment need to be bootstrapped to install base dependencies in the `vendor/` directory. Use the following command to bootstrap:
+
+```sh
+$ rake bootstrap
+```
+
+After bootstrap, logstash can be launched using `bin/logstash` but note that at this point, no plugin is installed.
+
+### Installing Core Plugins
+
+The logstash core plugins are required to run the core tests/specs. These core plugins and their development dependencies can be installed using:
+
+```sh
+$ rake test:install-core
+```
+
+### Installing Default Plugins
+
+Logstash is packaged with a default set of plugins. To install the set of default plugins and their development dependencies use:
+
+```sh
+$ rake test:install-default
+```
+
+### Installing All Plugins
+
+```sh
+$ rake test:install-all
+```
+
+## Testing
+
+Testing in logstash basically consist on 2 sets of tests, the core tests and the plugin tests
+
+### Core Tests
+
+The Logstash core tests are against core logstash code. Running all core tests can be done using:
+
+```sh
+rake test:core
+```
+
+Or individual test suites can be run using:
+
+```sh
+bin/rspec spec
+bin/rspec logstash-core/spec
+bin/rspec logstash-core-event/spec
+```
+
+Note that it is important to use `bin/rspec` and not just `rspec` because `bin/rspec` is aware of our specific environment. Both commands uses the same arguments.
+
+### Plugin tests
+
+It is possible to run all the **installed** plugins tests/specs from within the logstash development env which has the benefit of making sure all installed plugins tests/specs are passing againt the current development code-base. To run the tests/specs of all installed plugins use:
+
+```sh
+rake test:plugins
+```
+
+Normally you will want to make sure that the default set of plugins are passing the tests/specs. You read above on how install the plugins in your development envirinment.
+
+
+### Installing Development Dependencies
+
+If for any reason, for example a plugins was installed directly with `bin/plugin install ...`, it is possible to install the missing development dependencies which are required to run the tests/specs using:
+
+```sh
+rake plugin:install-development-dependencies
+```
+
+or
+
+```sh
+bin/plugin install --development
+```
+
+both are equivalent.
+
+## Using Alternate Core Modules
+
+Logstash depends on two core modules:
+
+- logstash-core
+
+  contains the pipeline related code.
+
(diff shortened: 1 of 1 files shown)


DEVELOPING.md
```diff
@@ -0,0 +1,126 @@
...
+bin/rspec spec
+bin/rspec logstash-core/spec
+bin/rspec logstash-core-event/spec
+```
+
+Note that it is important to use `bin/rspec` and not just `rspec` because `bin/rspec` is aware of our specific environment. Both commands uses the same arguments.
+
+### Plugin tests
+
+It is possible to run all the **installed** plugins tests/specs from within the logstash development env which has the benefit of making sure all installed plugins tests/specs are passing againt the current development code-base. To run the tests/specs of all installed plugins use:
```
ph: Following the discussion we had last week, I think we should change the wording to enforce it more and make it mandatory? I though it was the goal of this PR,

colinsurprenant: maybe do it in 2 sections: the first which explains how to do each thing and the second wich is more like a checklist of what you should do before submitting a PR?

ph: +1 for that suggestion.

DEVELOPING.md
```diff
@@ -0,0 +1,126 @@
+# Logstash Development Workflow
+
+This explains how to develop in the logstash core code-base. We will use to following fictional directories:
```
jsvd: s/we will use to/we will use the

DEVELOPING.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+After bootstrap, logstash can be launched using `bin/logstash` but note that at this point, no plugin is installed.
+
+### Installing Core Plugins
+
+The logstash core plugins are required to run the core tests/specs. These core plugins and their development dependencies can be installed using:
+
+```sh
+$ rake test:install-core
```
jsvd: looking at this, we should really move the install-{core,default-all} plugin installation take tasks to the plugin namespace

DEVELOPING.md
```diff
@@ -0,0 +1,126 @@
...
+- `/logstash/` is the dir containing your local fork clone of logstash
+- `/plugins/...` is the dir containing your local fork clones of logstash plugins or any other local private plugins
+
+## Setup
+
+Logstash uses [JRuby](http://jruby.org/) which gets embedded in the `vendor/jruby/` directory. It is **highly recommended** but not mandatory that you also use JRuby as your local Ruby interpreter and for this you should consider using a Ruby version manager such as [RVM](https://rvm.io/) or [rbenv](https://github.com/sstephenson/rbenv). It is possible to run the rake tasks and the `bin/` commands without having JRuby locally installed in which case the embedded JRuby will be used automatically.
+
+You will also need to have a JVM installed, we recommend using Java 1.8.
+
+`Rake` and `Bundler` should also be installed in your environment.
```
jsvd: these MUST be installed

DEVELOPING.md
jsvd: +1 for "what things are/do" and "checklists for workflow actions" sections

DEVELOPING.md
```diff
@@ -0,0 +1,126 @@
+# Logstash Development Workflow
```
jsvd: I wouldn't call this top level document the Workflow, as it mainly describes the available tooling and tasks available to the developer
Maybe just "Logstash Development" ?

Then we can add the second main section which describes the processes of:
* creating and submitting a PR
* reviewing and merging a PR
* developing in a feature branch

DEVELOPING.md
purbon: I agree with this, one thing is to actually describe the actions and another the workflow.

DEVELOPING.md
purbon: makes sense to me.

DEVELOPING.md
purbon: I think we should highlight this a bit more! we got issues from time to time that end up being because people don't have them installed beforehand. Also is important to notice the need of rubygems version > 2.5.x

see 

```
If you are seeing errors that look like

$ rake bootstrap
Installing minitar >= 0 because the build process needs it.
[bootstrap] Fetching and installing gem: minitar (>= 0)
rake aborted!
LoadError: no such file to load -- archive/tar/minitar
/Users/<user>/projects/logstash/rakelib/vendor.rake:17:in `untar'
/Users/<user>/projects/logstash/rakelib/vendor.rake:86:in `(root)'
Tasks: TOP => bootstrap => vendor:all => vendor:jruby
then you may need to update your version of rubygems. Run gem -v to see the version of rubygems installed. Version 2.5.2 or higher should work. To update rubygems run gem update --system (you may need to run with sudo if you're using your system Ruby environment).
```

from current readme file.

DEVELOPING.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Setup
+
+Logstash uses [JRuby](http://jruby.org/) which gets embedded in the `vendor/jruby/` directory. It is **highly recommended** but not mandatory that you also use JRuby as your local Ruby interpreter and for this you should consider using a Ruby version manager such as [RVM](https://rvm.io/) or [rbenv](https://github.com/sstephenson/rbenv). It is possible to run the rake tasks and the `bin/` commands without having JRuby locally installed in which case the embedded JRuby will be used automatically.
+
+You will also need to have a JVM installed, we recommend using Java 1.8.
+
+`Rake` and `Bundler` should also be installed in your environment.
+
+Note that if you have a local JRuby installed you can force logstash to use your local JRuby instead of the embedded JRuby with the `USE_RUBY=1` environment variable.
```
purbon: I would add a note for https://github.com/elastic/logstash/issues/3701, USE_RUBY is only working when running logstash, not when executing test.

DEVELOPING.md
```diff
@@ -0,0 +1,126 @@
...
+
+The logstash environment need to be bootstrapped to install base dependencies in the `vendor/` directory. Use the following command to bootstrap:
+
+```sh
+$ rake bootstrap
+```
+
+After bootstrap, logstash can be launched using `bin/logstash` but note that at this point, no plugin is installed.
+
+### Installing Core Plugins
```
purbon: I find this name a bit misleading, we got nothing as core plugins, but plugins that are used when running core test, I would recommend making this differentiation a bit more clear as It could be misleading for some people. I know we explain the test bit inside the paragraph, but you know, some people only read titles :smile:

DEVELOPING.md
purbon: yes and no, there is actually a task for this in the plugin namespace, but this one is also doing bootstrap and other things, see https://github.com/elastic/logstash/blob/master/rakelib/test.rake#L75. I think this an important separation of concerns, what we might improve probably is naming, but not sure how...

DEVELOPING.md
```diff
@@ -0,0 +1,126 @@
...
+### Installing Default Plugins
+
+Logstash is packaged with a default set of plugins. To install the set of default plugins and their development dependencies use:
+
+```sh
+$ rake test:install-default
+```
+
+### Installing All Plugins
+
```
purbon: probably add a bit of description here as it is with the others?

DEVELOPING.md
```diff
@@ -0,0 +1,126 @@
...
+## Using Alternate Core Modules
+
+Logstash depends on two core modules:
+
+- logstash-core
+
+  contains the pipeline related code.
+
+- logstash-core-event
```
purbon: we should update this to add the new reference to the core-event-java :+1:

suyograo:
I added some dev notes recently since this PR was stalled. Closing.

Issue #4382, opened 2015-12-21
refactor specs that depends on implementation dependant pipeline mock


colinsurprenant:
The ng-pipeline merge in master has broken lots of plugin tests/specs which were dependant on mocking  the pipeline based on its previous implementation. We need to review and refactor these tests to avoid implementation dependant tests/specs.

This relates to the currently failing specs meta issue #4371

A short term fix is proposed in #4373

PR #4384, opened 2015-12-21
fix to support legacy plugins specs monkeypatching pipelines

colinsurprenant:
because lots of plugins monkeypatch the pipeline and `output_func` returns nil we need to check for this until we properly refactor this specs as discussed in #4382 

Replaces #4373


andrewvc:
LGTM @colinsurprenant

colinsurprenant:
merged into master, 2.x and 2.2

PR #4395, opened 2015-12-23
rename ShutdownController to ShutdownWatcher

jsvd:
slightly adjust the metaphor for the class that is responsible for
monitoring a pipeline's shutdown process, report on its status and
intervene if it's stalled


logstash-core/spec/logstash/shutdown_watcher_spec.rb
```diff
@@ -1,12 +1,12 @@
...
-require "logstash/shutdown_controller"
+require "logstash/shutdown_watcher"
 
-describe LogStash::ShutdownController do
+describe LogStash::ShutdownWatcher do
 
   let(:check_every) { 0.01 }
   let(:check_threshold) { 100 }
-  subject { LogStash::ShutdownController.new(pipeline, check_every) }
+  subject { LogStash::ShutdownWatcher.new(pipeline, check_every) }
```
ph: you can use `described_class`  here so the only change would have been the describe block

suyograo:
@jsvd ping. would like to close this out for 2.2

logstash-core/spec/logstash/shutdown_watcher_spec.rb
jsvd: I know, but I really don't like "described_class", it's a terrible name, while having the class explicitly there makes it more readable IMO.
That said, I can change you feel strongly on this.

ph: I dont feel strong about it, just make it easier to refactor names :D

ph:
LGTM, I prefer `Watcher` instead of the controller, I think it describe more that the class is actually doing.

jsvd:
Closed by
master (be91439)
2.x (836482e)
2.2 (b887705)

PR #4408, opened 2015-12-30
Add a `WrappedSynchronousQueue#offer` method

ph:
Expose the Java's `#offer` method of the `SynchronousQueue` class,
This method allow the input developper to correctly apply back pressure
to the network clients.

If you use `#push` it will block forever until the pipeline free some
space, this behavior is sufficient if the pipeline outputs are healthy.
But if the output stale the backpressure will be applied up to the input
(producers) when reading a file this scenario is fine since we will just
stop reading the file until the thread unblock again.

In the context of network clients, the story is a bit different the
clients will timeout and try to reconnect to restransmit their payload
creating multiple new connection thread block on the queue. In some case
this will lead into a OOM issues.

This PR is the first step to communicate that the queue is under
pressure.

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -9,10 +9,24 @@ module LogStash; module Util
       @queue = java.util.concurrent.SynchronousQueue.new()
     end
 
+    # Push an object to the queue if the queue is full
+    # it will block until the object can be added to the queue.
+    #
+    # @param [Object] Object to add to the queue
     def push(obj)
       @queue.put(obj)
     end
     alias_method(:<<, :push)
+
+    # Offer an object to the queue, wait for the specified amout of time.
+    # If adding to the queue was successfull it wil return true, false otherwise.
+    #
+    # @param [Object] Object to add to the queue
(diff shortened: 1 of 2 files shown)


andrewvc:
LGTM, very good idea adding this in, and thanks for the detailed description :)

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -9,11 +9,25 @@ def initialize()
...
       @queue.put(obj)
     end
     alias_method(:<<, :push)
 
+    # Offer an object to the queue, wait for the specified amout of time.
+    # If adding to the queue was successfull it wil return true, false otherwise.
+    #
+    # @param [Object] Object to add to the queue
+    # @param [Integer] Time in milliseconds to wait before giving up
+    # @param [Boolean] True if adding was successfull if not it return false
```
jsvd: is this also @param since it's the return value? I'm not too familiar with the syntax

jsvd:
minor question, but code LGTM and tests pass :+1:

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
ph: good catch should be `@return`

PR #4409, opened 2015-12-31
Add http poller to default plugins list

andrewvc:
This plugin is fairly popular, won't increase the bundle size at all, and should probably be a default plugin.

I'd like to target it for the 2.2 release if possible!

### DIFF
--- a/rakelib/default_plugins.rb
+++ b/rakelib/default_plugins.rb
@@ -56,6 +56,7 @@ module LogStash
       logstash-input-generator
       logstash-input-graphite
       logstash-input-http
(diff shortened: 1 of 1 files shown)


jordansissel:
Build failed due to a rubygems problem, not due to this change:

```
ERROR:  Could not find a valid gem 'minitar' (>= 0), here is why:
          Unable to download data from https://rubygems.org/ - bad response Not Found 404 (https://rubygems.global.ssl.fastly.net/latest_specs.4.8.gz)
```

+1 on including http_poller in the default release. Change LGTM.

PR #4391, opened 2015-12-22
Improved semantics for OutputDelegator

andrewvc:
This commit greatly improves the semantics for the OutputDelegator. It
now:
* Has an API for declaring if an output is threadsafe which lets us economize on the number of instances (we only need one shared across N threads)
* By default sets the number of workers for an output to be == the
  number of pipeline workers (for non-threadsafe outputs)
* Moves the worker_safe API (which declares whether the plugin supports > 1 workers) to the class, not instance level
* Strongly enforces the worker safety API. Previously a warning would be logged but nothing else.
* Should Provide better performance for threadsafe plugins. A plugin like the file output should only need a mutex around the file write and could be written to be threadsafe. This opens the door to multiple threads running, say, their codecs simultaneously.

This commit is built on top of some of the ideas in https://github.com/elastic/logstash/issues/4367

PR #4410, opened 2015-12-31
Warn on potential memory issues when a large number of inflight messages are present

### LINKED ISSUE #4349: Add warning when batch size is configured too high
#4155 introduces a configurable batch size which can be tweaked to improve throughput. However, bigger batch size means more events inflight and this needs a bigger java heap configuration. We should add a warning when the limit is too high (maybe 10K events in total)

andrewvc:
Fixes #4349 . This warns on inflight max rather than batch size since even a low batch size could be dangerous if too many workers are present. `inflight_max == batch_size * pipeline_workers`.

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -17,7 +17,7 @@ require "logstash/pipeline_reporter"
 require "logstash/output_delegator"
 
 module LogStash; class Pipeline
-  attr_reader :inputs, :filters, :outputs, :worker_threads, :events_consumed, :events_filtered, :reporter, :pipeline_id
+  attr_reader :inputs, :filters, :outputs, :worker_threads, :events_consumed, :events_filtered, :reporter, :pipeline_id, :logger
 
   DEFAULT_SETTINGS = {
     :default_pipeline_workers => LogStash::Config::CpuCoreStrategy.fifty_percent,
@@ -26,6 +26,7 @@ module LogStash; class Pipeline
     :flush_interval => 5, # in seconds
     :flush_timeout_interval => 60 # in seconds
   }
+  MAX_INFLIGHT_WARN_THRESHOLD = 10_000
 
   def initialize(config_str, settings = {})
     @pipeline_id = settings[:pipeline_id] || self.object_id
@@ -157,11 +158,16 @@ module LogStash; class Pipeline
       pipeline_workers = safe_pipeline_worker_count
       batch_size = @settings[:pipeline_batch_size]
       batch_delay = @settings[:pipeline_batch_delay]
+      max_inflight = batch_size * pipeline_workers
       @logger.info("Starting pipeline",
                    :id => self.pipeline_id,
                    :pipeline_workers => pipeline_workers,
                    :batch_size => batch_size,
-                   :batch_delay => batch_delay)
+                   :batch_delay => batch_delay,
+                   :max_inflight => max_inflight)
+      if max_inflight > MAX_INFLIGHT_WARN_THRESHOLD
+        @logger.warn "CAUTION: Recommended inflight events max exceeded! Logstash will run with up to #{max_inflight} events in memory in your current configuration. If your message sizes are large this may cause instability with the default heap size. Please consider setting a non-standard heap size, changing the batch size (currently #{batch_size}), or changing the number of pipeline workers (currently #{pipeline_workers})"
+      end
 
       pipeline_workers.times do |t|
         @worker_threads << Thread.new do
--- a/logstash-core/spec/logstash/pipeline_spec.rb
+++ b/logstash-core/spec/logstash/pipeline_spec.rb
@@ -276,6 +276,36 @@ describe LogStash::Pipeline do
     end
   end
 
+  describe "max inflight warning" do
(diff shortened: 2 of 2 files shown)


suyograo:
LGTM

andrewvc:
For testers: If you run with `-b 10001` you should see the warning.

ph:
+1 for the warning.

Shouldn't we have a test for that?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -157,11 +157,16 @@ def start_workers
...
       batch_delay = @settings[:pipeline_batch_delay]
+      max_inflight = batch_size * pipeline_workers
       @logger.info("Starting pipeline",
                    :id => self.pipeline_id,
                    :pipeline_workers => pipeline_workers,
                    :batch_size => batch_size,
-                   :batch_delay => batch_delay)
+                   :batch_delay => batch_delay,
+                   :max_inflight => max_inflight)
+      if max_inflight > 10_000
```
ph: Maybe move 10_000 to a constant?

andrewvc:
@ph you're right, we should test for the warning msg.

ph:
@andrewvc I think moving to a constant + a test and it will be great. I think this is a good addition to help users.

andrewvc:
@ph will add both :)

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -17,7 +17,7 @@
 require "logstash/output_delegator"
 
 module LogStash; class Pipeline
-  attr_reader :inputs, :filters, :outputs, :worker_threads, :events_consumed, :events_filtered, :reporter, :pipeline_id
+  attr_reader :inputs, :filters, :outputs, :worker_threads, :events_consumed, :events_filtered, :reporter, :pipeline_id, :logger
```
andrewvc: Adding :logger here makes testing easier

andrewvc:
@ph how's this look now?

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -276,6 +276,36 @@ class TestPipeline < LogStash::Pipeline
...
+    let(:pipeline) { LogStash::Pipeline.new(config, :pipeline_batch_size => batch_size, :pipeline_workers => 1) }
+    let(:logger) { pipeline.logger }
+    let(:warning_prefix) { /CAUTION: Recommended inflight events max exceeded!/ }
+
+    before(:each) do
+      allow(LogStash::Plugin).to receive(:lookup).with("input", "dummyinput").and_return(DummyInput)
+      allow(LogStash::Plugin).to receive(:lookup).with("codec", "plain").and_return(DummyCodec)
+      allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(DummyOutput)
+      allow(logger).to receive(:warn)
+      thread = Thread.new { pipeline.run }
```
ph: Do we really need to start a pipeline for testing that error? 
Since all the validation is done in the initialize?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -157,11 +158,16 @@ def start_workers
...
+      max_inflight = batch_size * pipeline_workers
       @logger.info("Starting pipeline",
                    :id => self.pipeline_id,
                    :pipeline_workers => pipeline_workers,
                    :batch_size => batch_size,
-                   :batch_delay => batch_delay)
+                   :batch_delay => batch_delay,
+                   :max_inflight => max_inflight)
+      if max_inflight > MAX_INFLIGHT_WARN_THRESHOLD
+        @logger.warn "CAUTION: Recommended inflight events max exceeded! Logstash will run with up to #{max_inflight} events in memory in your current configuration. If your message sizes are large this may cause instability with the default heap size. Please consider setting a non-standard heap size, changing the batch size (currently #{batch_size}), or changing the number of pipeline workers (currently #{pipeline_workers})"
```
ph: We usually use the `hash` to pass variable to the logger, BUT In this context I believe string interpolation will be clearer and useful to the user.

 So I am OK to use that here.

ph:
Small comment concerning the test but LGTM

logstash-core/lib/logstash/pipeline.rb
andrewvc: Great, yeah, my feeling is this would never be machine parsed, and should be as user-friendly as possible.

logstash-core/spec/logstash/pipeline_spec.rb
andrewvc: This validation is actually done in start_workers. It needs to be there to avoid calling `safe_pipeline_worker_count` twice.

That should be refactored at some point, but I don't think that needs to happen here.

https://github.com/elastic/logstash/pull/4391/files should fix that issue a little bit, and ultimately removing the multiline filter should fully fix it (we won't have to do the check).

ph: So many workaround for the multiline :(

andrewvc: @ph I know. /me shakes fist at the heavens while cursing multiline

PR #4414, opened 2016-01-04
Up default # of pipeline workers to 100% of cores

andrewvc:
This is predicated on the fact that with the ng_pipeline it is expected that workers
will spend a significant amount of time in iowait due to outputs like the Elasticsearch
output. In benchmarks based on real-world Apache log files the best performance came out
of scenarios where pipeline_workers > num_cpu_cores. Setting this to default to the # of
cores is a defensive decision that should handle cases where users have particularly IO
heavy inputs.

For most users we should recommend setting the number of workers to be as high as possible
until performance decreases.

Previous benchmark information:
https://github.com/elastic/logstash/pull/4340

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -20,7 +20,7 @@ module LogStash; class Pipeline
   attr_reader :inputs, :filters, :outputs, :worker_threads, :events_consumed, :events_filtered, :reporter, :pipeline_id, :logger
 
   DEFAULT_SETTINGS = {
-    :default_pipeline_workers => LogStash::Config::CpuCoreStrategy.fifty_percent,
+    :default_pipeline_workers => LogStash::Config::CpuCoreStrategy.maximum,
     :pipeline_batch_size => 125,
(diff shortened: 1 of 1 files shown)


suyograo:
I am +1 on this. I'm a little concerned about inputs having codecs that would need cores, but I see the argument about IO wait cycles to yield processing power to other parts of the pipeline.

suyograo:
Others?

andrewvc:
@suyograo I share your concern, but in those cases I think the OS scheduler should do a good enough job anyway. If they are truly saturating their system then they need to add another instance or enlarge their current instance.

ph:
+1, I believe changing the default would be OK, since it will give a boost for most people and shouldn't negatively impact people. As I recall the number we initially choose was an arbitrary safe bet.

guyboertje:
+1

andrewvc:
@ph @guyboertje @suyograo are these +1s  LGTMs? Am I good to merge?

ph:
s/+1/lgtm

guyboertje:
LGTM

suyograo:
LGTM

PR #4421, opened 2016-01-05
Add a simplier way to log error messages inside logstash

ph:
This PR introduce a **Loggable** module that we can include into any classes
and it will add the instances/class methods necessary to do the logging.

The logging configuration is centralized at once place inside the
Runner, and this remove the need to pass a logger at initialize time or
by defining a custom accessor for changing it, drying up the code a bit.

by default the class will use the default logger defined by the module
but it can be redefined on specific classe.

This pull request also change the configuration order of the logger, the logger need to be configured with the right log level before its use, which was not the case before.


ph:
@colinsurprenant is thing workflow/code conflicting with the java event? If so I might need help for fixing it.

andrewvc:
Great idea IMHO. Weirdly, the tests pass for me, but not on CI!

I'd say LGTM, but these test failures need to be sussed out first!

ph:
@andrewvc Looking at the CI, oh computers :D

ph:
@andrewvc The instance isnt the same when run on the CI, might be an ordering issue trying to fix it.

ph:
Suggested by @jordansissel to look at https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/logging.rb and see what we can use/reuse.

I am not mistaken this "logging" class is not used anywhere, only required?

ph:
I have removed the requires on `logstash/logging` everywhere in the code but I cannot remove the file `logging.rb` because its required by `logstash-devutils` (https://github.com/elastic/logstash-devutils/blob/master/lib/logstash/devutils/rspec/spec_helper.rb#L15) for some obscure reason.

So before I create a new PR on devutils, are we OK with using `Loggable` as the name of this module or `logging` would be better? WDYT? (I wasn't aware of that file before creating this pr.)

ph:
Using logging.rb is the path to least resistance ;)

logstash-core/lib/logstash/util/loggable.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+    end
+
+    def self.included(base)
+      class << base
+        def logger
+          @logger || Loggable.logger
+        end
+
+        def logger=(logger)
```
colinsurprenant: cosmetic: you use two different method param names for almost the same code, maybe use the same names for the sake of consistency? 
```ruby
def logger=(new_logger)

def logger=(logger)
```

ph: True

colinsurprenant:
Yup! I like this! 

- I believe it should be fairly easy to adjust the Java Event code to use this new *loggable*
- I personally prefer loggable.rb to logging.rb
- Are we sure https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/logging.rb is not used anymore?

ph:
@colinsurprenant the logging is only used when running test.

https://github.com/ph/logstash-devutils/blob/15c59bfaab592f22531d333a25e4dcac39256f18/lib/logstash/devutils/rspec/spec_helper.rb#L28-L32

This is really confusing, we never use the `$logger` anywhere in the code? This global variable is not even used in `cabin`, So... This look like dead code that's never called or used by anything?

Lets make just a PR on devutils to remove it?

ph:
Created a PR for devutils https://github.com/elastic/logstash-devutils/pull/43

ph:
jenkins test this please

colinsurprenant:
should we put that on-hold until we decide on #4548 ?

ph:
@colinsurprenant I think they address different needs, this way or initializing the logger should be compatible with any kind of logger. Actually It could make the context aware logger easier to make its its a module we can scope it at include time.

talevy:
what is the state of this?

suyograo:
@ph bump. Should we close this PR?

talevy:
to be honest, I think some of these features are already in `master`. The Loggable feature, that is.

ph:
And its also in the log4j branch, so lets close this.

PR #4422, opened 2016-01-05
Remove sized_queue.rb

ph:
Since Logstash 2.2, we now use Java SynchronousQueue instead of ruby
sizequeue


ph:
I think we forgot to do that :D

andrewvc:
LGTM , but we still do use http://ruby-doc.org/stdlib-2.0.0/libdoc/thread/rdoc/SizedQueue.html in OutputDelegator FYI.

ph:
@andrewvc Correct!

PR #4425, opened 2016-01-06
Prevent field @timestamp from being removed

### LINKED ISSUE #4370: NoMethodError if field @timestamp is missing
If the field `@timestamp` is missing in an event, the following error occurs:

~~~
NoMethodError: undefined method `to_iso8601' for nil:NilClass
           to_s at /opt/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/event.rb:109
         encode at /opt/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-codec-line-2.0.2/lib/logstash/codecs/line.rb:53
        receive at /opt/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-output-stdout-2.0.3/lib/logstash/outputs/stdout.rb:49
         handle at /opt/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/outputs/base.rb:81
    output_func at (eval):47
   outputworker at /opt/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/pipeline.rb:276
  start_outputs at /opt/logstash-2.1.0/vendor/bundle/jruby/1.9/gems/logstash-core-2.1.0-java/lib/logstash/pipeline.rb:193
~~~

**Steps to reproduce**

Start Logstash with the following config:

~~~
input {
  generator {}
}
filter {
  mutate {
    rename => {
      "@timestamp" => "timestamp"
    }
  }
}
output {
  stdout {}
}
~~~

breml:
Fixes #4370

### DIFF
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -265,7 +265,7 @@ describe LogStash::Event do
 
       context "remove" do
         it "should raise an exception if you attempt to remove @timestamp" do
-          expect{subject.remove("@timestamp")}.to raise_error(TypeError)
+          expect{subject.remove("@timestamp")}.to raise_error(InvalidOperationException)
         end
       end
     end
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -267,6 +267,11 @@ describe LogStash::Event do
         it "should raise an exception if you attempt to remove @timestamp" do
           expect{subject.remove("@timestamp")}.to raise_error(InvalidOperationException)
         end
+        it "should not raise an exception if you attempt to remove timestamp" do
+          subject["timestamp"] = 1456933880
+          expect{subject.remove("timestamp")}.not_to raise_error
+          expect(subject["timestamp"]).to be_nil
+        end
       end
     end
 
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -63,7 +63,7 @@ describe LogStash::Event do
       it "should raise error when formatting %{+%s} when @timestamp field is missing" do
         str = "hello-%{+%s}"
         subj = subject.clone
-        subj.remove("[@timestamp]")
+        subj.instance_variable_get(:@accessors).del("[@timestamp]")
         expect{ subj.sprintf(str) }.to raise_error(LogStash::Error)
       end
 
@@ -80,7 +80,7 @@ describe LogStash::Event do
       it "should raise error with %{+format} syntax when @timestamp field is missing", :if => RUBY_ENGINE == "jruby" do
         str = "logstash-%{+YYYY}"
         subj = subject.clone
-        subj.remove("[@timestamp]")
+        subj.instance_variable_get(:@accessors).del("[@timestamp]")
         expect{ subj.sprintf(str) }.to raise_error(LogStash::Error)
       end
 
@@ -260,6 +260,12 @@ describe LogStash::Event do
         it "should join array, removing duplicates" do
           subject.append(LogStash::Event.new({"field1" => [ "append1","original1" ]}))
           expect(subject[ "field1" ]).to eq([ "original1", "original2", "append1" ])
+        end
+      end
+
+      context "remove" do
+        it "should raise an exception if you attempt to remove @timestamp" do
+          expect{subject.remove("@timestamp")}.to raise_error(TypeError)
         end
       end
     end
--- a/logstash-core-event/lib/logstash/event.rb
+++ b/logstash-core-event/lib/logstash/event.rb
@@ -179,6 +179,9 @@ class LogStash::Event
 
   # Remove a field or field reference. Returns the value of that field when deleted
   def remove(fieldref)
+    if fieldref == TIMESTAMP || fieldref == "[#{TIMESTAMP}]"
+      raise TypeError, "The field '@timestamp' must not be removed"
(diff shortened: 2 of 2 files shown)


jordansissel:
@breml overall looks ok to me; I'm not sure if TypeError is the best exception to use for this, though. Thoughts? We can create a new exception type if it makes sense.

jordansissel:
jenkins, test it

breml:
@jordansissel You are right, I just followed the implementation on lIne 127 (https://github.com/u-s-p/logstash/blob/5b24e2f276de755eb1c3b330aa3c50ef848718ce/logstash-core-event/lib/logstash/event.rb#L127), but this is not the same case. I don't know the exception system of ruby well enough to make a sound 	suggestion. I appreciate your help on this.

suyograo:
@breml @jordansissel How about introducing `InvalidOperationException` ? `TypeError` does seem at odds here.

breml:
@suyograo @jordansissel Just updated this PR with InvalidOperationException.

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -63,7 +63,7 @@
       it "should raise error when formatting %{+%s} when @timestamp field is missing" do
         str = "hello-%{+%s}"
         subj = subject.clone
-        subj.remove("[@timestamp]")
+        subj.instance_variable_get(:@accessors).del("[@timestamp]")
         expect{ subj.sprintf(str) }.to raise_error(LogStash::Error)
```
suyograo: Can you check for `InvalidOperationException` here?

suyograo:
@breml can you also add a test where you add a dummy field called `timestamp` not `@timestamp` and assert if you can delete it correctly? Thats a scenrio i've seen in the field, wanted to make sure we protect against that.

suyograo:
@breml ping! I'd love to get this in soon :)

breml:
@suyograo Here you go. (I also have some plugin related issues, where I would love to have some progress soon :)

suyograo:
@breml unfortunately I had to revert this in https://github.com/elastic/logstash/pull/4756 because `rake tests:core` (unit tests) were failing. There are lots of tests checking `@timestamp`. Can you reopen a new PR based on this which has all the tests passing? Thanks much

breml:
@suyograo After several hours of unsuccessful debugging, I think I have found the issue. With https://github.com/elastic/logstash/commit/23dcf467e490e417bf377a0b2a32a73653434879#diff-efbf03f6386a8f9610003e5858706663R15 the change from logstash-core-event to logstash-core-event-java (nicely hidden in a commit about "metrics api") was introduced (intentional?). 

For the first set of errors, the explanation is:
I extended the logstash-event with a new  exception (InvalidOperationException, https://github.com/u-s-p/logstash/commit/c077c9cb66afed26f7855173d24a72abca1d0e0d), this exception is not present in the java event implementation.

For the second set of errors, there seems to be a difference in the behavior between the ruby and the java implementation regarding `subj.instance_variable_get(:@accessors).del("[@timestamp]")`. I assume that `@accessors` is not available in the java implementation.
The reason why I use `instance_variable_get` is, for the respective test case I have to remove the `@timestamp` field, but I can not use the `remove` method, because in the `remove` method it is not allowed to remove the `@timestamp` field.

**I don't think I'm able to resolve this one, as I don't have any knowledge about the new java based event implementation.**

purbon:
@breml turning logstash-core-event-java on in master was done a while ago before we merged the metrics PR, I'm curios why the diff looks like this, there might be something fishy going there for sure.

re: for the new java event, @colinsurprenant or @guyboertje might be able to help here more, any ideas?

purbon:
@breml logstash-core-event-java was enabled back on `Jan 26, 2016`, see https://github.com/elastic/logstash/commit/5a522a00370fb7bd7275d4ae7d7ab996a5fd329e#diff-8b7db4d5cc4b8f6dc8feb7030baa2478, while the commit you point if from `Feb 4`. But I guess the error might be coming out now because in your commit, we actually also updated the Gemfile.lock, that was not updated in the first one. .... :cry: what do you think?

breml:
@purbon At the moment I think, that this is quite frustrating, because the PR was put together in much less time than I needed today to debug this issue. And my PR kind of updates legacy code, which will never be used. This doesn't feel good.

purbon:
@breml I apologies for the inconvenience you're facing, the move to the new java event bring us a ton of new performance, you can see it http://logstash-meter.elastic.co/ for example. This will also enable us to move forward with more persistency and making easier for people to write plugins in any JVM language. Your PR might found a bug in the new java implementation, this is much appreciate it, but I do understand your feelings.

breml:
@purbon Thanks for the link to the metrics. More logstash throughput is always welcome. Hopefully you are right and my PR has found a bug in the new java implementation. In this case it would be a least valuable.

suyograo:
> I don't think I'm able to resolve this one, as I don't have any knowledge about the new java based event implementation.

@breml we can definitely port this to the java implementation. 

> And my PR kind of updates legacy code, which will never be used. This doesn't feel good.
 I wouldn't think about it this way. 2.x release will live for a long time with Ruby implementation of Event, so your change will definitely be used. Also, the same logic will be in Java Event (not able to remove `@timestamp`), so overall your change really benefits us. I'll take another look at the failing tests and Java event related issues.

breml:
@suyograo Thanks for your feedback. Ping me if the Java implementation is ready. I will push the PR again. OK?

PR #4434, opened 2016-01-08
Bump version to 2.2.0.snapshot2

suyograo:
Bump version, update gem and lock file to point to RubyGems.org
in preperation for 2.2.0 release

### DIFF
--- a/Gemfile.jruby-1.9.lock
+++ b/Gemfile.jruby-1.9.lock
@@ -1,28 +1,3 @@
-PATH
-  remote: ./logstash-core-event
-  specs:
-    logstash-core-event (2.2.0-java)
-
-PATH
-  remote: ./logstash-core
-  specs:
-    logstash-core (2.2.0-java)
-      cabin (~> 0.7.0)
-      clamp (~> 0.6.5)
-      concurrent-ruby (= 0.9.2)
-      filesize (= 0.0.4)
-      gems (~> 0.8.3)
-      i18n (= 0.6.9)
-      jrjackson (~> 0.3.7)
-      jruby-openssl (>= 0.9.11)
-      logstash-core-event (~> 2.2.0)
-      minitar (~> 0.5.4)
-      pry (~> 0.10.1)
-      rubyzip (~> 1.1.7)
-      stud (~> 0.0.19)
-      thread_safe (~> 0.3.5)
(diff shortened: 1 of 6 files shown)


colinsurprenant:
yup, LGTM!

colinsurprenant:
@suyograo checking error, I think it may just be a missing `require` - this is the first time we actually test a "real" package with core-event gem in rubygems.

colinsurprenant:
@suyograo oh sweetjesus :sweat_smile: I finally just found the problem. the core-even gem is malformed, it does not include the correct `/lib` directory... fixing!

colinsurprenant:
actually - both core gems have the root logstash content because they were build from the top level logstash dir. we need to `cd` into each and launch `gem build` in the root of each gem.

the proper fix here is simply to rebuild and push both core gems and update this PR to new gem versions.

suyograo:
Fixed the previous error. Built from within gem dirs, yanked old snapshot1 gems and republished snapshot2 gems

colinsurprenant:
LGTM + Jenkins all green :shipit:

Gemfile.jruby-1.9.lock
```diff
@@ -259,11 +251,12 @@ GEM
       logstash-core (>= 2.0.0.beta2, < 3.0.0)
       nokogiri
       xml-simple
-    logstash-input-beats (2.0.3)
+    logstash-input-beats (2.1.0)
```
ph: Oh yes!

Gemfile.jruby-1.9.lock
```diff
@@ -430,19 +428,19 @@ GEM
...
       logstash-filter-json
       logstash-output-file
-    logstash-output-elasticsearch (2.3.0-java)
+    logstash-output-elasticsearch (2.4.0-java)
       cabin (~> 0.6)
       concurrent-ruby
       elasticsearch (~> 1.0, >= 1.0.13)
       logstash-core (>= 2.0.0, < 3.0.0)
-      manticore (~> 0.4.2)
+      manticore (>= 0.5.2, < 1.0.0)
```
ph: Small comment here, this is a pretty open dependency, from what it was before?

ph:
Look good to me!

PR #4435, opened 2016-01-08
Add rake tasks to create uber packages

suyograo:
Add rake tasks to create zip and tar.gz for LS package with all plugins
installed from losgtash-repo

### DIFF
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -94,13 +94,25 @@ namespace "artifact" do
   end
 
   task "prepare" => ["bootstrap", "plugin:install-default", "install-logstash-core", "install-logstash-core-event", "clean-bundle-config"]
+  task "prepare-all" => ["bootstrap", "plugin:install-all", "install-logstash-core", "install-logstash-core-event", "clean-bundle-config"]
 
-  desc "Build a tar.gz of logstash with all dependencies"
+  desc "Build a tar.gz of default logstash plugins with all dependencies"
   task "tar" => ["prepare"] do
+    puts("[artifact:tar] Building tar.gz of default plugins")
+    build_tar
+  end
+
+  desc "Build a tar.gz of all logstash plugins from logstash-plugins github repo"
+  task "tar-all-plugins" => ["prepare-all"] do
+    puts("[artifact:tar] Building tar.gz of all plugins")
+    build_tar "all-plugins"
+  end
+
+  def build_tar(tar_suffix = "")
     require "zlib"
     require "archive/tar/minitar"
     require "logstash/version"
-    tarpath = "build/logstash-#{LOGSTASH_VERSION}.tar.gz"
+    tarpath = "build/logstash-#{tar_suffix}-#{LOGSTASH_VERSION}.tar.gz"
     puts("[artifact:tar] building #{tarpath}")
     gz = Zlib::GzipWriter.new(File.new(tarpath, "wb"), Zlib::BEST_COMPRESSION)
     tar = Archive::Tar::Minitar::Output.new(gz)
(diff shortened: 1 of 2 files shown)


rakelib/artifacts.rake
```diff
@@ -94,13 +94,25 @@ namespace "artifact" do
...
+    puts("[artifact:tar] Building tar.gz of all plugins")
+    build_tar "all-plugins"
+  end
+
+  def build_tar(tar_suffix = "")
     require "zlib"
     require "archive/tar/minitar"
     require "logstash/version"
-    tarpath = "build/logstash-#{LOGSTASH_VERSION}.tar.gz"
+    tarpath = "build/logstash-#{tar_suffix}-#{LOGSTASH_VERSION}.tar.gz"
```
ph: This will not be backward compatible with the previous naming convention since if no suffix is provided we will have two `-` in the name. I think the easy fix is to set `NIL` as the default value of `tar_suffix` and then use this method.

 tarpath = ["build/logstash", tar_suffix, "{LOGSTASH_VERSION}.tar.gz"].compact.join("-")

rakelib/artifacts.rake
```diff
@@ -132,9 +144,21 @@ namespace "artifact" do
...
+  desc "Build a zip of all logstash plugins from logstash-plugins github repo"
+  task "zip-all-plugins" => ["prepare-all"] do
+    puts("[artifact:zip] Building zip of all plugins")
+    build_zip "all-plugins"
+  end
+
+  def build_zip(zip_suffix = "")
     require 'zip'
-    zippath = "build/logstash-#{LOGSTASH_VERSION}.zip"
+    zippath = "build/logstash-#{zip_suffix}-#{LOGSTASH_VERSION}.zip"
```
ph: See my comment on `build_tar`

rakelib/plugin.rake
```diff
@@ -52,7 +52,18 @@ namespace "plugin" do
...
     puts("[plugin:install-all] Installing all plugins from https://github.com/logstash-plugins")
-    install_plugins("--no-verify", *LogStash::RakeLib.fetch_all_plugins)
+    p = *LogStash::RakeLib.fetch_all_plugins
+    # Install plugin one by one, ignoring plugins that have issues. Otherwise, one bad plugin will
+    # blow up the entire install process.
+    # TODO Push this downstream to #install_plugins
+    p.each do |plugin|
+      begin
+        install_plugins("--no-verify", plugin)
+      rescue
```
ph: Not sure if the Gemfile will be in a consistent state if the one of the plugin fails, but its something we can/should monitor.

I think we will want to use Jenkins to run this command? We will need to make sure we bubble theses error up if we have too many fails this probably mean something is wrong with the build or the dependency tree?

ph:
@suyograo In your test did you encounter this message?
```
Exception: java.lang.ThreadDeath thrown from the UncaughtExceptionHandler in thread "Thread-67"
Installing logstash-codec-netflow
Installation successful
Installing logstash-codec-oldlogstashjson
Installation successful
Installing logstash-codec-plain
[... remove for clarity]
Installing logstash-filter-date
Installation successful
Installing logstash-filter-dns
Installation successful

Exception: java.lang.ThreadDeath thrown from the UncaughtExceptionHandler in thread "Thread-103"

**Comment by ph** [comment]
I also encountered theses issues, which might not be related to your PR, `cross-dependency` problems since we call install on individual plugins?

```
ERROR: Installation Aborted, message: Bundler could not find compatible versions for gem "logstash-codec-plain":
  In snapshot (Gemfile.lock):
    logstash-codec-plain (= 2.0.2)

  In Gemfile:
    logstash-input-generator (>= 0) java depends on
      logstash-codec-plain (>= 0) java

    logstash-input-generator (>= 0) java depends on
      logstash-codec-plain (>= 0) java

    logstash-input-generator (>= 0) java depends on
      logstash-codec-plain (>= 0) java

    logstash-input-generator (>= 0) java depends on
      logstash-codec-plain (>= 0) java

    logstash-input-generator (>= 0) java depends on
      logstash-codec-plain (>= 0) java

    logstash-input-generator (>= 0) java depends on
      logstash-codec-plain (>= 0) java

    logstash-input-github (>= 0) java depends on
      logstash-codec-plain (~> 0.1.6) java

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
Bundler could not find compatible versions for gem "logstash-core":
  In snapshot (Gemfile.lock):
    logstash-core (= 3.0.0.dev)

jordansissel:
Failure to install logstas-input-heroku also: https://gist.github.com/jordansissel/1dc7678a6827f8bfcfc2

jordansissel:
I'm OK with merging and we can resolve the plugins that don't install correctly later since my feeling is that the plugins that fail to install are problems with those plugins (need updated version constraints) and not a problem with this PR.

LGTM

ph:
Theses comment https://github.com/elastic/logstash/pull/4435/files#r49224181 were not applied before merging.

The nightly build have changed names 

screenshot 2016-01-08 20 59 19

suyograo:
@ph apologies. I will make a PR to fix them

ph:
@suyograo dont worry not a big issue :)

suyograo:
@ph comment fixed in https://github.com/elastic/logstash/pull/4440

PR #4448, opened 2016-01-11
Fix declare_workers_not_supported docs to do correct respond_to?

andrewvc:
Previously respond_to? checked for the wrong method!


jordansissel:
Haven't tested, but the fix looks correct to me. LGTM

jsvd:
LGTM

suyograo:
@andrewvc just curious how the specs did not catch this? Can we add a spec covering this?

PR #4453, opened 2016-01-11
Update offline plugin docs

### LINKED ISSUE #4426: Add example for updating all plugins from local cache
https://www.elastic.co/guide/en/logstash/2.1/offline-plugins.html#_install_or_update_a_local_plugin

Let's add a 3rd example for a common use case - on how to update all local plugins with a single command from the cache after unpacking.

```
Example 3. Updating all local plugins
bin/plugin update --local
```

suyograo:
Add Updating all local plugins in one command

Fixes #4426


ppf2:
LGTM thx!

PR #4455, opened 2016-01-11
Validate pipeline numeric options to ensure positive integers.

### LINKED ISSUE #4449: Need to reject --pipeline-batch-size 0

andrewvc:
There were no specs here before, and IMHO they're a bit overkill for
this section of code, so I'm not changing that.

Fixes #4449

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -29,17 +29,17 @@ class LogStash::Runner < Clamp::Command
   option ["-w", "--pipeline-workers"], "COUNT",
     I18n.t("logstash.runner.flag.pipeline-workers"),
     :attribute_name => :pipeline_workers,
-    :default => LogStash::Pipeline::DEFAULT_SETTINGS[:default_pipeline_workers], &:to_i
+    :default => LogStash::Pipeline::DEFAULT_SETTINGS[:default_pipeline_workers] {|s| validate_positive_integer(s) }
 
   option ["-b", "--pipeline-batch-size"], "SIZE",
          I18n.t("logstash.runner.flag.pipeline-batch-size"),
          :attribute_name => :pipeline_batch_size,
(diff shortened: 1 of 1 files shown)


logstash-core/lib/logstash/runner.rb
```diff
@@ -79,6 +79,15 @@ class LogStash::Runner < Clamp::Command
     I18n.t("logstash.runner.flag.node_name"),
     :attribute_name => :node_name
 
+  def validate_positive_integer(str_arg)
+    int_arg = str_arg.to_i
+    if str_arg !~ /^\d+$/ || int_arg < 1
```
suyograo: shouldn't this be `int_arg <= 0`?

logstash-core/lib/logstash/runner.rb
```diff
@@ -79,6 +79,15 @@ class LogStash::Runner < Clamp::Command
     I18n.t("logstash.runner.flag.node_name"),
     :attribute_name => :node_name
 
+  def validate_positive_integer(str_arg)
```
suyograo: IMO we should move this to `utils` dir, but we can do that in a separate PR.

logstash-core/lib/logstash/runner.rb
andrewvc: Since its an integer `int_arg < 1 == int_arg <= 0`.

logstash-core/lib/logstash/runner.rb
andrewvc: I think you're mostly right, but its a little premature. I'd agree if we had two validations, but since its only one, I'd rather not create a one method utility class.

logstash-core/lib/logstash/runner.rb
suyograo: oh man, i don't know what i was thinking there (facepalm) I swear i read it as `int_arg < 0`

suyograo:
LGTM

Issue #4446, opened 2016-01-11
Design: Metrics API endpoints for Logstash


purbon:
As part of the ongoing metrics implementation (#3908), this issue specs all the user facing endpoints for Phase 1.  
Aim is to be consistent as much as possible with ES API, so user experience is similar across products.

### 1. Root resource

Request

``` GET /  ```

Response

```
{
   "hostname": "skywalker",
    "version" : {
        "number" : "2.1.0",
        "build_hash" : "72cd1f1a3eee09505e036106146dc1949dc5dc87"
    }
  }
```

**Note:** For phase 1, we will only have a single pipeline (metrics is an internal pipeline and will not be exposed) so this API will only show `main`

### 2. Stats Info

We would like to slice viewing of runtime stats under 2 dimensions: 
* Node level: These are instance level stats. Logstash is not clustered today, and nodes do not communicate with other nodes. Hence the name `_node` instead of `_nodes`. The action `_nodes` imply communication with other nodes in a cluster.
* Pipeline level: When multiple pipelines are supported by a single Logstash instance, you can view stats per pipeline. 

```
/_node/stats/<types>
/_pipelines/stats/<types>
```
`types` are like events, jvm, os

**Filtering by pipeline: (not for Phase 1)**

`/_pipelines/apache, metric_beat/stats/<types>`

This will show stats for pipelines named `apache` *and* `metric_beat`.

#### 2.1 Events data

Get events information since startup. 

Request

``` GET /_node/stats/events  ```

Response

```
{
    "events": {
        "in": 91,
        "filtered": 91,
        "out": 91
    }
}
```

#### 2.2 JVM data

JVM stats, garbage collection

Request

``` GET /_stats/jvm ```

Response

```
"jvm":{  
   "timestamp":1453233447702,
   "uptime_in_millis":211125811,
   "mem":{  
      "heap_used_in_bytes":58442576,
      "heap_used_percent":5,
      "heap_committed_in_bytes":259522560,
      "heap_max_in_bytes":1037959168,
      "non_heap_used_in_bytes":56332256,
      "non_heap_committed_in_bytes":57475072,
      "pools":{  
         "young":{  
            "used_in_bytes":41672000,
            "max_in_bytes":286326784,
            "peak_used_in_bytes":71630848,
            "peak_max_in_bytes":286326784
         },
         "survivor":{  
            "used_in_bytes":260552,
            "max_in_bytes":35782656,
            "peak_used_in_bytes":8912896,
[...]

PR #4458, opened 2016-01-12
Data collection wire up

purbon:
In this PR there is the intention to connect the data generation layer (logstash-core metrics collection) with the web api and show some real data. This changes include a first organization of the resources to match the initial design at #4446.

Pending tasks:

* [x] Wire up static data like list of plugins, etc.
* [x] Add basic API tests.
* [x] Adapt the API to match the initial designs at #4446, this will include an initial root resource sketch plus the ```events``` information.


logstash-core/api/lib/app/pipeline/stats_command.rb
```diff
@@ -7,14 +7,9 @@ def run
...
     stats = service.get(:pipeline_stats)
-    snapshot = { "events" => {} }
-    stats.each_pair do |key, value|
-      if key.start_with?("events_")
-        field = key.gsub("events_", "")
-        snapshot["events"][field] = value
-      end
-    end
-    snapshot
+    { "events" => { "in" => stats[:base][:events_in].value, "filtered" => stats[:base][:events_filtered].value } }
```
ph: Is there any reason we don't use symbols in the hash definition? I think we should since everything from the metric store uses a symbol for index access.

logstash-core/api/lib/app/root.rb
```diff
@@ -6,7 +6,7 @@ class Root < BaseApp
 
     get "/" do
       content = { "name" => "Logstash API",
-                  "version" => { "number" => "0.1.0" },
+                  "version" => { "number" => LOGSTASH_CORE_VERSION },
```
ph: Might be good to require the file that declare this constant? no?

logstash-core/api/lib/app/service.rb
```diff
@@ -1,22 +1,26 @@
 # encoding: utf-8
 require "logstash/instrument/collector"
+require "logstash/util/loggable"
 
 class LogStash::Api::Service
 
   include Singleton
+  include LogStash::Util::Loggable
```
ph: +1

logstash-core/api/lib/app/system/info_command.rb
```diff
@@ -4,20 +4,30 @@
...
-    report
   end
 
   private
 
-  def type
-    ["input", "output", "filter"][rand(3)]
+  def hostname
+    `hostname`.strip
   end
```
ph: You can use the `Socket.gethostname` so we don't need to shell out for that.

logstash-core/api/lib/app/root.rb
ph: I think it should be `LOGSTASH_VERSION` instead of `LOGSTASH_CORE_VERSION` the latter is the version of the gem that we use and its _might_ be out of sync with the release version.

logstash-core/api/lib/app/system/info_command.rb
purbon: hostname is standard and cross platform? I would not like to have a machine where the hostname is an IP because of a broken DNS lookup, what do you think?

logstash-core/api/lib/app/pipeline/stats_command.rb
purbon: no specific reason, in LS I always saw strings, not symbols so this is why I stick with strings, but can use symbols :+1:

logstash-core/api/lib/app/system/info_command.rb
ph: *I believe* `Socket.gethostbyname` doesn't do dns check, if you want to do that you have to use 
``
Socket.getbyhostname(Socket.gethostname)

ph: When I disconnect my laptop and I try it, I don't have have any errors and my hostname is also invalid.
```ruby
~/e/l/logstash-input-beats git:fix/timing-issue-with-test ❯❯❯ irb                                                                                                                                                                   ◼
jruby-1.7.23 :001 > require 'Socket'
 => true
jruby-1.7.23 :002 > Socket.gethostname
 => "sashimi.local"
jruby-1.7.23 :003 >

File: logstash-core/api/lib/app/system/info_command.rb [hunk 5]
```diff
@@ -4,20 +4,30 @@
...
+      spec.name
+    end
   end
+
+  def all_available_plugins
+    Gem::Specification.find_all.select do |spec|
+      spec.metadata && spec.metadata["logstash_plugin"] == "true"
+    end
+  end
+
```
ph: Don't we have already something in core that can return that?
If its not the case, I think it should be there where we do all the bundle/gem/manipulations? WDYT?

logstash-core/api/spec/lib/api/root_spec.rb
```diff
@@ -1,5 +1,6 @@
 # encoding: utf-8
 require_relative "../../spec_helper"
+require "json"
```
ph: Lets use our own json serializer `LogStash::Json`

logstash-core/api/lib/app/system/info_command.rb
```diff
@@ -4,20 +4,30 @@
...
-               "host_name" => "foobar",
-               "plugins"   => { "count" => 10, "list" => [] }
+    plugins_list = available_plugins
+    {
+      "version"   => LOGSTASH_CORE_VERSION,
+      "hostname" => hostname,
+      "plugins"   => { "count" => plugins_list.count, "list" => plugins_list.sort }
     }
-    report
   end
```
ph: Do we want to return the logstash version on every call? I suspect the call on `/` should be enough no?

logstash-core/api/spec/lib/api/root_spec.rb
purbon: sure, this should have been changed already.

logstash-core/api/lib/app/system/info_command.rb
purbon: @ph to reuse it here should be extracted and refactored properly, but yes there is something in the plugin manager.

ph: I think we should do it, but since we want to iterate quickly I am +1 for a specific PR only with that feature? Lets create an issue and do it in a specific PR.

purbon: as this is not part of our initial effort, I would hold the guns for now. But yes, extracting logic is something we should do. For now I will just remove this command as is not in use anywhere and will confuse you are reviewer.

logstash-core/api/spec/lib/commands/events_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+require 'ostruct'
+
+describe LogStash::Api::StatsEventsCommand do
+
+  let(:service) { double("snapshot-service") }
+
+  subject { described_class.new(service) }
+
+  let(:stats) do
+    { :base => { :events_in => OpenStruct.new(:value => 100), :events_filtered => OpenStruct.new(:value => 200) } }
```
ph: you can use double(bleh, :value => 100)?

ph:
LGTM

PR #4459, opened 2016-01-12
remove link to maintainer guide in contributing since already in index

### LINKED ISSUE #4460: Docs include maintainer-guide twice thus generating an error
http://build-us-00.elastic.co/job/Docs/3611/console happens because the [[community-maintainer]] tag is defined twice because the document itself is included both in the index.asciidoc and the contributing-to-logstash.asciidoc

jsvd:
the maintainer guide is already included in https://github.com/elastic/logstash-docs/blob/master/docs/index.asciidoc so there's no need to also include in the contributing-to-logstash.
@palecur @suyograo is this the right approach to fix the double include?

Also..is there a reason to not have the index.asciidoc file of logstash-docs in this repo too?

fixes #4460


ph:
LGTM

Issue #4443, opened 2016-01-11
Logstash 2.2 falsely claims worker count was changed by user


jsvd:
```
/tmp/logstash-2.2.0.snapshot2 % bin/logstash -e "input { stdin {} } filter { multiline { pattern => 'hey' what => previous negate => true } }"
{:pipeline_workers=>4, :pipeline_batch_size=>125, :pipeline_batch_delay=>5, :pipeline_id=>"base"}
Settings: User set pipeline workers: 4
Warning: Manual override - there are filters that might not work with multiple worker threads {:worker_threads=>4, :filters=>["multiline"], :level=>:warn}
```

This is because [agent.rb has a default value for filter workers](https://github.com/elastic/logstash/blob/2.2/logstash-core/lib/logstash/agent.rb#L27) pulled from `LogStash::Pipeline::LOGSTASH_DEFAULTS`.
Then the [pipeline.rb code checks if the setting was passed](https://github.com/elastic/logstash/blob/2.2/logstash-core/lib/logstash/pipeline.rb#L89-L110) (overriden by user), which will always be true.

purbon:
Good catch @jsvd. As I understand the only way to do a manual override is through the ```-w``` param, right? So I guess here the ones who should show this message is the agent, and just check if we passed this parameter? makes sense?

jsvd:
Doing this check before Clamp is very hacky, it would involve grepping ARGV to find either -w N, --filter-workers N, --filter-workers=N.

After Clamp, the two options here are:
a) moving the override check closer to the agent.rb Clamp::Command
b) pass down the distinction between default settings and user defined settings to the pipeline

Both options above assume clamp can queried about the declared options, this needs to be checked.

purbon:
I don't think B makes sense, should be the agent the one who deals with Clamp options and actually report with this kind of changes. When I said checking for ```-w``` i was actually thinking on your option a. What do you think?

jsvd:
closed by #4466

PR #4466, opened 2016-01-12
distinguish between user set and default pipeline settings

jsvd:
fixes https://github.com/elastic/logstash/issues/4443

the fix takes advantage of clamp's ability to define the setters for the options. These setters are only invoked when the user has explicitly set the option.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -24,17 +24,18 @@ class LogStash::Agent < Clamp::Command
   option ["-w", "--pipeline-workers"], "COUNT",
          I18n.t("logstash.runner.flag.pipeline-workers"),
          :attribute_name => :pipeline_workers,
-         :default => LogStash::Pipeline::DEFAULT_SETTINGS[:default_pipeline_workers] {|s| validate_positive_integer(s) }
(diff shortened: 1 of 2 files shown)


andrewvc:
Tested locally, LGTM

logstash-core/lib/logstash/agent.rb
```diff
@@ -66,6 +67,23 @@ class LogStash::Agent < Clamp::Command
...
+  end
+
+  def pipeline_batch_size=(pipeline_batch_size_value)
+    @pipeline_settings[:pipeline_batch_size] = validate_positive_integer(pipeline_batch_size_value)
+  end
+
+  def pipeline_batch_delay=(pipeline_batch_delay_value)
+    @pipeline_settings[:pipeline_batch_delay] = validate_positive_integer(pipeline_batch_delay_value)
+  end
+
```
ph: Neat! didnt know that from clamp.

ph:
Tests passes, codec LGTM, tested locally !

suyograo:
@jsvd can you commit a diff patch for 2.x and 2.2? Master has deviated

jsvd:
2.2 0b35848
2.x 2952abe

I'll create a new PR against master due to it's heavy deviation

jsvd:
patch against master ended up being very straight forward, merged in 69ca10a

PR #4467, opened 2016-01-12
Fix for 4444: Too verbose deprecations warnings

guyboertje:
Pass a deeply cloned copy of args to each output constructor invocation

### DIFF
--- a/logstash-core/lib/logstash/output_delegator.rb
+++ b/logstash-core/lib/logstash/output_delegator.rb
@@ -25,7 +25,9 @@ module LogStash class OutputDelegator
     # Older plugins invoke the instance method Outputs::Base#workers_not_supported
     # To detect these we need an instance to be created first :()
     # TODO: In the next major version after 2.x remove support for this
-    @workers = [@klass.new(*args)]
+
+    # deep_clone because the output constructor mutates args in-place
+    @workers = [@klass.new(*deep_clone(args))]
     @workers.first.register # Needed in case register calls `workers_not_supported`
 
     # DO NOT move this statement before the instantiation of the first single instance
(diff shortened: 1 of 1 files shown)


guyboertje:
I could not create a failing test. Spec uses doubles. I tested by changing the code in the downloaded snapshot until the fault went away then copied the change into repo branch.

ph:
@guyboertje wouldnt be possible to use a dummy plugin/spy to write a test for this?

ph:
pseudo code, did not test.
```ruby
class MutateSpy < LogStash::Ouputs::Base
  attr_reader :counter
  def initialize(options)
    super(options)
     options["counter"] += 1
     # assuming the initial value is 0
     @counter = options["counter"]  
   end
end

#And the expectation will look like this
@outputdeletagors.workers.each do |w| expect(w.counter).to eq(1)
```

guyboertje:
On reflection a DummyInput will work. Problem lies with the spec design - its not simple to sandwich another context block in at the appropriate place. The spec only needs to verify that the args passed to the Output delegator are not mutated by successive calls to the DummyInput constructor when pipeline_workers are > 1.  Without my change, the spec will fail because the the args are mutated.

ph:
@guyboertje Correct, but I would be ok to have a test to only check for the mutated.

But, something make me think this bug, we possibly have it in all the type of plugins we have right?
But this is only exposed by the way we define workers, since the initialize isn't side effect free.

guyboertje:
The args are mutated intentionally - but only due to default value merge and validation.  [See /config/mixin.rb#L47]( https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/config/mixin.rb#L47).

Your counter will not be altered.  It starts with the keys given by the user in the config then it adds the configs defined by the output. So if the resultant hash is given to subsequent output constructors - you get deprecation warnings. If workers is 6 then we got 5 deprecation warnings.

ph:
@guyboertje thats a bummer :(

guyboertje:
Now after @ph and I have discussed this, the solution offered with this PR is wrong. Closing.

guyboertje:
Better proposal - in plugin.rb
```ruby
  def initialize(params=nil)
    @params = Marshal.load(Marshal.dump(params))
  end
```

ph:
The only problem, we could have is if someone override the initialize in a subclass, oh well can't save all the cases? I am not aware of a lot of plugin that implement a custom initialize.

PR #4472, opened 2016-01-13
stats/jvm integration

### LINKED ISSUE #3909: API: Hot Threads information
Currently there is no way to get information about the running threads inside logstash other than collecting a thread dump with `jstack` It would be great to implement an API endpoint similar to the elasticsearch's `hot threads`.  Ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-hot-threads.html

I think this can be done easily by using the java management api, we could probably use elasticsearch's class directly and return the same results.

In my initial experimentations you have to use the java api because the ruby api will hide the pure java threads, like the one used in Manticore or the rabbit mq client.

purbon:
This PR integrated information into the webapi coming out of the java virtual machine. This include:

* hot threads information.
* memory usage.

Pending tasks:

* [x] Add a global stats resource where most of the relevant stats information are show.
* [x] Format the hot threads not as a hash, but as plain text reply (1) (2)
* [x] Investigate the best way to represent memory consumption.
* [x] Add options to filter the number of threads returned by the hot threads api, also an option to ignore idle threads or not (same behaviour as elasticsearch API does).

(1) This is because the expected consumer of this API is going to be a human and not a software client, having a plain text message makes everything easy to comprehend.
(2) This PR have an initial version of the hot threads report, this should be improved in next versions with more detailed information.

Fixes #3909


purbon:
@ph 

``` rake test:core``` fail due to:

```
Gem::LoadError: You have already activated jruby-openssl 0.9.14, but your Gemfile requires jruby-openssl 0.9.13. Prepending `bundle exec` to your command may solve this.
/Users/purbon/work/logstash/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/runtime.rb:34:in `setup'
/Users/purbon/work/logstash/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/runtime.rb:19:in `setup'
/Users/purbon/work/logstash/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler.rb:122:in `setup'
/Users/purbon/work/logstash/lib/bootstrap/bundler.rb:64:in `setup!'
/Users/purbon/work/logstash/rakelib/test.rake:15:in `(root)'
```

@ph should we rebase with master?

**Comment by purbon** [comment]
@ph I confirm, in master this does not happen, so I foresee a rebase.

**Comment by purbon** [comment]
@ph already did the change to use gauge. Your thoughts here more than welcome.

**Comment by ph** [comment]
I think the rebase was already done, but lock file isnt correct.

**Comment by purbon** [comment]
@ph This error is with an updated lock file.

**Comment by ph** [comment]
I am currently rebasing master to the `feature/metrics` strangely I have a few conflict to merge.
I will squash the commits and and keep ownership.

File: logstash-core/api/lib/app/stats.rb [hunk 1]
```diff
@@ -17,7 +31,25 @@ class Stats < BaseApp
...
+    end
+
+    # return hot threads information
+    get "/jvm/hot_threads" do
+      top_threads_count = params["threads"].to_i || 3
+      ignore_idle_threads = params["ignore_idle_threads"] || true
+      options = {
+        :threads => top_threads_count,
+        :ignore_idle_threads => as_boolean(ignore_idle_threads)
+      }
```
ph: Did you thought about using an class for that and we could just pass if to the `command.run`.
This class could take care of the default values and values casting? 

We can make a class for each endpoint that requires parsing user data.

IIRC in django/rails world we call them form object.

logstash-core/api/lib/app/stats/hotthreads_command.rb
```diff
@@ -0,0 +1,43 @@
+# encoding: utf-8
+require "app/command"
+require 'monitoring'
+require "socket"
+
+class LogStash::Api::HotThreadsCommand < LogStash::Api::Command
+
+  SKIPPED_THREADS = [ "Finalizer", "Reference Handler", "Signal Dispatcher" ]
```
ph: We might want to freeze this constant.

logstash-core/api/lib/app/stats.rb
purbon: We can do it, what about doing it in a feature PR?

- purbon

logstash-core/api/lib/app/stats/hotthreads_command.rb
```diff
@@ -0,0 +1,43 @@
...
+      break if i >= top_count
+      if ignore
+        next if SKIPPED_THREADS.include?(thread_name)
+        next if thread_name.match(/Ruby-\d+-JIT-\d+/)
+      end
+      report << "<p> #{build_report(container)} </p>"
+      i += 1
+    end
+    report
+  end
```
ph: Lets use plain text for formatting and return the correct mime type for the request.

logstash-core/api/lib/app/stats/memory_command.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  private
+
+  def dump(hash)
+    hash.inject({}) do |h, (k,v)|
+      h[k] = v.value
+      h
+    end
+  end
+
```
ph: I am wondering should we just implemented `to_s` on the `MetricType`? The json serializer should just call that?

ph: This would make the code simplier?

logstash-core/lib/logstash/instrument/periodic_poller/jvm.rb
```diff
@@ -1,13 +1,92 @@
...
+    private
+
+    ##
+    #  (TODO) usage of increment need to be changed to gauge as soon as this
+    #  type is implemented. This for all the metrics collectors in this class.
+    ##
+
+    def collect_heap_metrics(data)
+      heap = aggregate_information_for(data["heap"].values)
+      heap[:used_percent] = (heap[:used_in_bytes] / heap[:max_in_bytes].to_f)*100.0
```
ph: Not sure we need `to_f` in this case since we will convert it back to `to_i` later on.

logstash-core/lib/logstash/instrument/periodic_poller/jvm.rb
```diff
@@ -1,13 +1,92 @@
...
+      collect_non_heap_metrics(raw)
+      collect_pools_metrics(raw)
+    end
+
+    private
+
+    ##
+    #  (TODO) usage of increment need to be changed to gauge as soon as this
+    #  type is implemented. This for all the metrics collectors in this class.
+    ##
```
ph: We can remove this comment

logstash-core/api/spec/lib/commands/jvm_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      it "return heap information" do
+        expect(report.keys).to include("heap")
+      end
+
+      it "return non heap information" do
+        expect(report.keys).to include("non_heap")
+      end
+
+    end
+  end
```
ph: We should check for the nested key too in the tests.

ph:
Added a few comments, almost ready to merge in.

logstash-core/api/lib/app/stats/memory_command.rb
purbon: More than a ```to_s``` method, to have a marshal method that dump a structure as a hash, meaning without the require of a value method. I see this a very common use case for consumers, so yes, I would agree we should implemented in the ```metric_type``` or ```metric_store``` level.

logstash-core/lib/logstash/instrument/periodic_poller/jvm.rb
purbon: We have to do it, see:

```
irb(main):003:0> (76/100.0)*100
=> 76.0
irb(main):004:0> (76/100)*100
=> 0
irb(main):005:0> (76/100)*100.0
=> 0.0
```

otherwise the division is clearly wrong.

ph: Ah silly ruby and his type system,  yeah you are right.

logstash-core/api/lib/app/stats/memory_command.rb
ph: The MetricType already has a `to_hash` method? But I doubt the format is compatible with our return values.

purbon: this is not helpful, it has the same problems. What need to be done is actually return an adaptor for the ```Concurrent::Map``` used as internal store what actually have a ```to_hash``` method.

ph: Agree +1 this would be a better solution.

logstash-core/api/spec/fixtures/memory.json
```diff
@@ -0,0 +1,42 @@
...
+		},
+		"survivor": {
+			"used_in_bytes": 1,
+			"committed_in_bytes": 2,
+			"max_in_bytes": 3,
+			"peak_used_in_bytes": 4,
+			"peak_max_in_bytes": 5
+		}
+	}
+}
```
ph: Hehe I see what you did here!

purbon: This fixture, will go away as soon as I fix the issue with the api not getting data soon enough. For now is a necessary pain.

ph:
OK, rebase this and merge it, so we can iterate.

Issue #4444, opened 2016-01-11
Too verbose deprecations warnings


purbon:
With this very simple config,

```
input {
  stdin {}
}

output {
  stdout { codec => rubydebug }
  elasticsearch {
    hosts => [ "127.0.0.1:9200" ]
  }
}
```

when I start logstash I see this warning message,

```
You are using a deprecated config setting "retry_max_items" set in elasticsearch. Deprecated settings will continue to work, but are scheduled for removal from logstash in the future.  If you have any questions about this, please visit the #logstash channel on freenode irc. {:name=>"retry_max_items", :plugin=><LogStash::Outputs::ElasticSearch hosts=>["127.0.0.1:9200"], codec=><LogStash::Codecs::Plain charset=>"UTF-8">, workers=>1, index=>"logstash-%{+YYYY.MM.dd}", manage_template=>true, template_name=>"logstash", template_overwrite=>false, flush_size=>500, idle_flush_time=>1, doc_as_upsert=>false, max_retries=>3, script_type=>"inline", script_var_name=>"event", scripted_upsert=>false, retry_max_interval=>2, retry_max_items=>500, action=>"index", path=>"/", ssl=>false, ssl_certificate_verification=>true, sniffing=>false, sniffing_delay=>5>, :level=>:warn}
You are using a deprecated config setting "retry_max_items" set in elasticsearch. Deprecated settings will continue to work, but are scheduled for removal from logstash in the future.  If you have any questions about this, please visit the #logstash channel on freenode irc. {:name=>"retry_max_items", :plugin=><LogStash::Outputs::ElasticSearch hosts=>["127.0.0.1:9200"], codec=><LogStash::Codecs::Plain charset=>"UTF-8">, workers=>1, index=>"logstash-%{+YYYY.MM.dd}", manage_template=>true, template_name=>"logstash", template_overwrite=>false, flush_size=>500, idle_flush_time=>1, doc_as_upsert=>false, max_retries=>3, script_type=>"inline", script_var_name=>"event", scripted_upsert=>false, retry_max_interval=>2, retry_max_items=>500, action=>"index", path=>"/", ssl=>false, ssl_certificate_verification=>true, sniffing=>false, sniffing_delay=>5>, :level=>:warn}
[...]

PR #4473, opened 2016-01-13
Fix no. 2 for 4444: Too verbose deprecations warnings

guyboertje:
Better method of ensuring that args are not mutated by plugin initialize.

NOTE:
- I had to introduce `LogStash::Util.deep_clone` as `Marshal.load(Marshal.dump(o))` does not work directly on args because there is logger instance in args and Marshal can't dump IO.
- I had to alter all the codec, filter, input and output base.rb `initialize` to deep_clone params, as well as the plugin.rb `initialize` to do `config_init(@params)`.

### DIFF
--- a/logstash-core/lib/logstash/util.rb
+++ b/logstash-core/lib/logstash/util.rb
@@ -183,4 +183,21 @@ module LogStash::Util
       o
     end
   end
+
+  def self.deep_clone(o)
+    case o
+    when Hash
+      o.inject({}) {|h, (k,v)| h[k] = deep_clone(v); h }
+    when Array
+      o.map {|v| deep_clone(v) }
+    when Fixnum, Symbol, IO, TrueClass, FalseClass, NilClass
+      o
+    when LogStash::Codecs::Base
+      o.clone.tap {|c| c.register }
+    when String
+      o.clone #need to keep internal state e.g. frozen
+    else
+      Marshal.load(Marshal.dump(o))
+    end
+  end
 end # module LogStash::Util
--- a/logstash-core/spec/logstash/plugin_spec.rb
+++ b/logstash-core/spec/logstash/plugin_spec.rb
@@ -108,7 +108,7 @@ describe LogStash::Plugin do
 
       subject.validate({})
     end
-    
+
 
     it 'logs a warning if the plugin use the milestone option' do
       expect_any_instance_of(Cabin::Channel).to receive(:warn)
@@ -117,6 +117,35 @@ describe LogStash::Plugin do
       class LogStash::Filters::Stromae < LogStash::Filters::Base
         config_name "stromae"
         milestone 2
+      end
+    end
+  end
+
+  describe "subclass initialize" do
+    let(:args) { Hash.new }
+
+    [
+      StromaeCodec = Class.new(LogStash::Codecs::Base) do
+        config_name "stromae"
+        config :foo_tag, :validate => :string, :default => "bar"
+      end,
+      StromaeFilter = Class.new(LogStash::Filters::Base) do
+        config_name "stromae"
+        config :foo_tag, :validate => :string, :default => "bar"
+      end,
+      StromaeInput = Class.new(LogStash::Inputs::Base) do
+        config_name "stromae"
+        config :foo_tag, :validate => :string, :default => "bar"
+      end,
+      StromaeOutput = Class.new(LogStash::Outputs::Base) do
+        config_name "stromae"
+        config :foo_tag, :validate => :string, :default => "bar"
+      end
+    ].each do |klass|
+
+      it "subclass #{klass.name} does not modify params" do
+        instance = klass.new(args)
+        expect(args).to be_empty
       end
     end
   end
--- a/logstash-core/lib/logstash/codecs/base.rb
+++ b/logstash-core/lib/logstash/codecs/base.rb
@@ -11,7 +11,7 @@ module LogStash::Codecs; class Base < LogStash::Plugin
 
   def initialize(params={})
     super
-    config_init(params)
+    config_init(@params)
     register if respond_to?(:register)
   end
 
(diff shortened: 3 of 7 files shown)


guyboertje:
Added specs
fixes #4444

andrewvc:
I believe this will also fix https://github.com/logstash-plugins/logstash-output-tcp/issues/13#issuecomment-171389649

andrewvc:
LGTM and tests pass (incl. plugin tests)

logstash-core/lib/logstash/util.rb
```diff
@@ -183,4 +183,21 @@ def self.stringify_symbols(o)
...
+  def self.deep_clone(o)
+    case o
+    when Hash
+      o.inject({}) {|h, (k,v)| h[k] = deep_clone(v); h }
+    when Array
+      o.map {|v| deep_clone(v) }
+    when Fixnum, Symbol, IO, TrueClass, FalseClass, NilClass
+      o
+    when LogStash::Codecs::Base
+      o.clone
```
colinsurprenant: should we do a clone here or instantiate a new codec?

logstash-core/lib/logstash/util.rb
```diff
@@ -183,4 +183,21 @@ def self.stringify_symbols(o)
...
+    when String
+      o.dup
```
purbon: i wonder why dup and not clone? I understand for String would have the same effect, right?

logstash-core/lib/logstash/util.rb
purbon: some question here? do we want to keep internal state, as I could read from the docs:

> In general, clone and dup may have different semantics in descendant classes. While clone is used to duplicate an object, including its internal state, dup typically uses the class of the descendant object to create the new instance.

logstash-core/spec/logstash/plugin_spec.rb
```diff
@@ -120,4 +120,33 @@ class LogStash::Filters::Stromae < LogStash::Filters::Base
       end
     end
   end
+
+  describe "subclass initialize" do
```
purbon: not sure about having this test here as it is, or having an specific test for codecs, filters, outputs, inputs, etc ... I think it would be an awesome addition to have also checking the super call is working as expected.

purbon:
Small comments about unit test location and clone, but for the rest it looks good to me.

I tested manually with


```
input {
  generator { 
    message => "foobar"
    count => 1
  }
}

output {
  stdout { codec => rubydebug }
  elasticsearch {
    hosts => [ "127.0.0.1:9200" ]
    workers => 3
  }
}
```

and see

```
skywalker% ./bin/logstash -f ../manual_testing/2.2/output-es.config 
Logstash startup completed
{
       "message" => "foobar",
      "@version" => "1",
    "@timestamp" => "2016-01-14T08:46:47.604Z",
          "host" => "0.0.0.0",
      "sequence" => 0
}
Logstash shutdown completed
```

what means works as expected.

Thanks @guyboertje for putting this fix in place.

logstash-core/lib/logstash/util.rb
guyboertje: Codecs::Base does this
```ruby
  def clone
    return self.class.new(params)
  end
```
and this will call deep_clone of the params given.

BTW codec base is the only subclass that has a clone method.

If anything we should call register after clone
e.g. 
```ruby
  def self.deep_clone(o)
    case o
    when Hash
      o.inject({}) {|h, (k,v)| h[k] = deep_clone(v); h }
    when Array
      o.map {|v| deep_clone(v) }
    when Fixnum, Symbol, IO, TrueClass, FalseClass, NilClass
      o
    when LogStash::Codecs::Base
      o.clone.tap {|c| c.register }
    when String
      o.dup
    else
      Marshal.load(Marshal.dump(o))
    end
  end
```

guyboertje: @colinsurprenant @purbon - we definitely do want to call clone here.  IMO this exactly what this clone method is for.

logstash-core/lib/logstash/util.rb
guyboertje: @purbon - no it should be clone.  Will change.
```sh
irb(main):007:0> du = "aasasas".freeze.dup
=> "aasasas"
irb(main):008:0> fr = "aasasas".freeze.clone
=> "aasasas"
irb(main):009:0> du.frozen?
=> false
irb(main):010:0> fr.frozen?
=> true
```
Thanks for querying that.

logstash-core/spec/logstash/plugin_spec.rb
guyboertje: I would expand the individual specs (codec does not have one; referee foul!) in:
`logstash/logstash-core/spec/logstash/(filter|inputs|outputs)/base_spec.rb` to do what you suggest.

These are here because it is verifying the ultimate superclass behaviour.

logstash-core/lib/logstash/util.rb
purbon: seeing what clone is doing for codecs, makes  sense for me. 

@guyboertje calling register makes sense too.

logstash-core/spec/logstash/plugin_spec.rb
purbon: sure, having both is good.

guyboertje: Thinking about it more, we need a shared example similar to the shuts-down one that all plugins run because we can't otherwise guarantee that an author does not modify the params if they override `initialize`.

purbon: makes sense to me. Go for it.

logstash-core/lib/logstash/util.rb
colinsurprenant: :thumbsup:

driverpt:
@guyboertje , does this closes #5211 ?

guyboertje:
@driverpt - No, not until the Event#clone method uses `LogStash::Util.deep_clone` - the PR has stalled.

PR #4474, opened 2016-01-13
Metrics/refactor small improvements

ph:
This PR contains:
- a few speed improvements for collecting metrics by removing unnecessary allocations.
- A way to get the uptime for a pipeline and the agent
- a filterdelegator to collect filter stats.

### DIFF
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -7,7 +7,6 @@ module LogStash
     def_delegators :@filter,
       :register,
       :close,
-      :multi_filter,
       :threadsafe?,
       :do_close,
       :do_stop,
@@ -18,18 +17,30 @@ module LogStash
 
       @logger = logger
       @klass = klass
-      @metric = metric
       @filter = klass.new(options)
 
-      define_flush if @filter.respond_to?(:flush)
+      # Scope the metrics to the plugin
+      @metric = metric.namespace(@filter.identifier_name)
+      @filter.metric = @metric
+
+      define_flush_method if @filter.respond_to?(:flush)
     end
 
     def config_name
       @klass.config_name
     end
 
+    def multi_filter(events)
+      @metric.increment(:events_in, events.size)
+
+      new_events = @filter.multi_filter(events)
+
+      @metric.increment(:events_out, new_events.size)
+      return new_events
+    end
+
     private
-    def define_flush
+    def define_flush_method
       define_singleton_method(:flush) do |options = {}|
         @filter.flush(options)
       end
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -1,0 +1,38 @@
+# encoding: utf-8
+#
+module LogStash
+  class FilterDelegator
+    extend Forwardable
+
+    def_delegators :@filter,
+      :register,
+      :close,
+      :multi_filter,
+      :threadsafe?,
+      :do_close,
+      :do_stop,
+      :periodic_flush
+
+    def initialize(logger, klass, metric, *args)
+      options = args.reduce({}, :merge)
+
+      @logger = logger
+      @klass = klass
+      @metric = metric
+      @filter = klass.new(options)
+
+      define_flush if @filter.respond_to?(:flush)
+    end
+
+    def config_name
+      @klass.config_name
+    end
+
+    private
+    def define_flush
+      define_singleton_method(:flush) do |options = {}|
+        @filter.flush(options)
+      end
+    end
+  end
+end
--- a/logstash-core/lib/logstash/inputs/metrics.rb
+++ b/logstash-core/lib/logstash/inputs/metrics.rb
@@ -44,9 +44,9 @@ module LogStash module Inputs
       #   - We can use a synchronization mechanism between the called thread (update method)
       #   and the plugin thread (run method)
       #   - How we handle back pressure here?
-      #   - one snashot should be only one event
-      snapshot.metric_store.to_events.each do |event|
-        @queue << event
+      #   - one snashot should be only one event ?
+      snapshot.metric_store.all.each do |metric|
+        @queue << LogStash::Event.new({ "@timestamp" => snapshot.created_at }.merge(metric.to_hash))
       end
     end
   end
--- a/logstash-core/lib/logstash/instrument/metric.rb
+++ b/logstash-core/lib/logstash/instrument/metric.rb
@@ -38,8 +38,8 @@ module LogStash module Instrument
       if block_given?
         start_time = Time.now
         content = block.call
-        duration = Time.now - start_time
-        gauge(key, duration)
+        duration = (Time.now - start_time) * 1000 # Records in Milliseconds
+        collector.push(namespace_information, key, :mean, :increment, duration)
         return content
       else
         raise MetricNoBlockProvided
--- a/logstash-core/lib/logstash/instrument/metric.rb
+++ b/logstash-core/lib/logstash/instrument/metric.rb
@@ -9,6 +9,29 @@ module LogStash module Instrument
   class MetricNoNamespaceProvided < MetricException; end
 
   class Metric
+    class TimedExecution
+      MILLISECONDS = 1_000_000.0.freeze
+
+      def initialize(metric, key)
+        @metric = metric
+        @key = key
+        start
+      end
+
+      def start
+        @start_time = current_time
+      end
+
+      def stop
+        @metric.report_time(@key, (MILLISECONDS * (current_time - @start_time)).to_i)
+      end
+
+      private
+      def current_time
+        Time.now
+      end
+    end
+
     attr_reader :collector, :namespace_information
 
     public
@@ -33,17 +56,21 @@ module LogStash module Instrument
       collector.push(namespace_information, key, :gauge, :set, value)
     end
 
-    def time(key, &block)
+    def time(key)
       validate_key!(key)
+
       if block_given?
-        start_time = Time.now
-        content = block.call
-        duration = (Time.now - start_time) * 1000 # Records in Milliseconds
-        collector.push(namespace_information, key, :mean, :increment, duration)
+        timer = TimedExecution.new(self, key)
+        content = yield
+        timer.stop
         return content
       else
-        raise MetricNoBlockProvided
+        TimedExecution.new(self, key)
       end
+    end
+
+    def report_time(key, duration)
+      collector.push(namespace_information, key, :mean, :increment, duration)
     end
 
     def namespace(sub_namespace)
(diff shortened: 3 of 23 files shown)


ph:
This PR also add a method a way to format the duration correctly for a human.

`Pipeline#uptime` => time in milliseconds since started
`Agent#uptime` => Time in milliseconds since started

You can have a human string with `LogStash::FormatDuration.format(milliseconds)` this will return something like this `4d 45m`

logstash-core/spec/logstash/instrument/null_metric_spec.rb
```diff
@@ -7,4 +7,16 @@
   it "defines the same interface as `Metric`" do
     expect(described_class).to implement_interface_of(LogStash::Instrument::Metric) 
   end
+
+  describe "#time" do
+    it "returns the vaule of the block without recording any metrics" do
+      x = 1
```
purbon: why this variable here?

logstash-core/spec/logstash/output_delegator_spec.rb
```diff
@@ -15,10 +15,20 @@
 
     before do
       allow(out_klass).to receive(:new).with(any_args).and_return(out_inst)
+      allow(out_inst).to receive(:register)
+      allow(out_inst).to receive(:identifier_name)
+      allow(out_inst).to receive(:metric=).with(any_args)
+      allow(logger).to receive(:debug).with(any_args)
+    end
+
+    before do
```
purbon: I would advise to use always ```before(:each)```  or ```before(:all)```, as relying on the default it gets me always confused for the expected behaviour.

logstash-core/lib/logstash/webserver.rb
```diff
@@ -5,8 +5,7 @@
 require 'puma/configuration'
 require 'puma/commonlogger'
 
-module LogStash
-  class WebServer
+module LogStash class WebServer
```
purbon: why this change? it looks to me a bit cosmetic, before it was actually more clear to understand, right?

purbon:
@ph can we add also here a method in the agent that lets query the pipelines? I envision the API holding a reference to the agent and then this being responsible to ask the pipelines. What do you think?

purbon:
@ph can you rebase this branch, files like ```info_command.rb``` were actually removed. This change should happen in the new command used for the root resource.

logstash-core/spec/logstash/instrument/null_metric_spec.rb
ph: you are right I could replace that with a string.

ph:
@purbon rebased, cleaned with your comments.

added a command in agent to retrieve the pipelines.

purbon:
@ph I guess this need now another rebase?

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -37,8 +39,8 @@ def get(*key_paths)
     # Take all the individuals `MetricType` and convert them to `Logstash::Event`
```
purbon: This documentation should be updated.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -37,8 +39,8 @@ def get(*key_paths)
...
     #
     # @return [Array] An array of all metric transformed in `Logstash::Event`
-    def to_events
-      to_events_recursively(@store).flatten
+    def all_metrics
```
purbon: not totally comfortable with a name like ```all_metrics``` for a class named ```MetricStore```, what do you think? might be named it simply ```all``` ?

logstash-core/lib/logstash/instrument/null_metric.rb
```diff
@@ -1,6 +1,16 @@
...
+require "logstash/instrument/metric"
+
 module LogStash module Instrument
  class NullMetric
+   class NullTimedExecution
+     def stop
+     end
+   end
+
+   NULL_TIMED_EXECUTION_INSTANCE = NullTimedExecution.new
```
purbon: I don't understand why do you use constans here for this kind of values? what is your way of thinking here? thanks!

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -135,13 +146,13 @@ def filters?
...
 
   def run
+    @started_at = Time.now
+
     LogStash::Util.set_thread_name("[#{pipeline_id}]-pipeline-manager")
     @logger.terminal(LogStash::Util::DefaultsPrinter.print(@settings))
 
     start_workers
 
-    running = true
```
purbon: good catch, :+1:

logstash-core/lib/logstash/inputs/metrics.rb
```diff
@@ -44,9 +44,9 @@ def update(snapshot)
       #   - We can use a synchronization mechanism between the called thread (update method)
       #   and the plugin thread (run method)
       #   - How we handle back pressure here?
-      #   - one snashot should be only one event
-      snapshot.metric_store.to_events.each do |event|
-        @queue << event
+      #   - one snashot should be only one event ?
+      snapshot.metric_store.all_metrics.each do |metric|
```
purbon: As I shared in the metrics store, I'm somehow in the fence for methods like this one. What do you think about a notation like

```
snapshot.metric_store.each do |metric|
   ...
end
```

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -10,6 +10,28 @@ class MetricNoBlockProvided < MetricException; end
   class MetricNoNamespaceProvided < MetricException; end
 
   class Metric
+    class TimedExecution
+      MILLISECONDS = 1_000_000.0
```
purbon: can we freeze this constant?

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -10,6 +10,28 @@ class MetricNoBlockProvided < MetricException; end
...
+
+      def start
+        @start_time = current_time
+      end
+
+      def stop
+        @metric.report_time(@key, (MILLISECONDS * (current_time - @start_time)).to_i)
+      end
+
+      def current_time
```
purbon: what is the need/purpose of `this ``current_time``` method?

purbon: I would make this method private as is only internal.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -58,13 +60,13 @@ def get_recursively(key_paths, map)
       end
     end
 
-    def to_events_recursively(values)
+    def all_metrics_recursively(values)
```
purbon: same thing for this method, naming looks confusing to me.

logstash-core/lib/logstash/instrument/metric_store.rb
ph: +1

logstash-core/lib/logstash/instrument/metric_store.rb
ph: +1

logstash-core/lib/logstash/instrument/null_metric.rb
ph: I don't want the overhead of creating  a new instance on each call since it will return a noop object.

logstash-core/lib/logstash/inputs/metrics.rb
ph: I am bit worried to use `each` in this case, since I don't want people thing its an enumerable when its not. I am OK to use `.all` like you suggested earlier?

ph: For searching I was thinking more about a `#query` or a `fetch` method

logstash-core/lib/logstash/instrument/metric.rb
ph: we can for sure.

logstash-core/lib/logstash/instrument/null_metric.rb
purbon: I understand. why not having a singleton? just curious.... :smile:

ph: both work I guess ?

purbon: Yes, probably a singleton will be more clear. Change if we feel to do it.

ph:
Changed `all_` for `each` and rebased with feature/metrics.

logstash-core/lib/logstash/filter_delegator.rb
```diff
@@ -0,0 +1,49 @@
...
+    def initialize(logger, klass, metric, *args)
+      options = args.reduce({}, :merge)
+
+      @logger = logger
+      @klass = klass
+      @filter = klass.new(options)
+
+      # Scope the metrics to the plugin
+      @metric = metric.namespace(@filter.identifier_name)
+      @filter.metric = @metric.namespace(:internal)
```
purbon: should we update the ```internal``` namespace here? I mean as discussed in #4492.

logstash-core/lib/logstash/inputs/metrics.rb
purbon: works for me!

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -33,19 +56,23 @@ def gauge(key, value)
...
-        content = block.call
-        duration = Time.now - start_time
-        gauge(key, duration)
+        timer = TimedExecution.new(self, key)
+        content = yield
+        timer.stop
         return content
       else
-        raise MetricNoBlockProvided
+        TimedExecution.new(self, key)
```
purbon: I understand the ``ìf``` part execute the block passed, but not sure I see here if this is executing anything, I'm I missing anything?

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -12,11 +12,18 @@ module LogStash class OutputDelegator
...
     @threadsafe = klass.threadsafe?
     @config = args.reduce({}, :merge)
     @klass = klass
+    
+    # Create an instance of the input so we can fetch the identifier
+    output = @klass.new(*args)
+    
+    # Scope the metrics to the plugin
+    @metric = metric.namespace(output.identifier_name)
+    internal = @metric.namespace(:internal)
```
purbon: same things here, I guess we can remove the internal namespace now.

purbon:
@ph very small comments, mostly leftovers and I'm good with us merging this !!! good work here!

logstash-core/lib/logstash/output_delegator.rb
ph: agree!

logstash-core/lib/logstash/instrument/metric.rb
ph: Something you don't want to capture a block, this method support both kind of invocation.

```ruby
timer = metric.time("responsetime")
# do work
timer.stop

or 

metric.time("responsetime")  do
# do work
end
```

logstash-core/lib/logstash/filter_delegator.rb
ph: agree we should removed.

PR #4477, opened 2016-01-13
Removed the Reporter class, this was used for debugging when developping the collector.

ph:
There is not need for theses classes to exist anymore they were used at the beginning to test the logic of the reporter.


purbon:
LGTM

PR #4482, opened 2016-01-13
add plugins dev instructions

### LINKED ISSUE #4175: document plugins dev Gemfile requirements with new core gems structure
related to #4123

when doing plugin development, the plugin `Gemfile` can/should include both logstash-core and logstash-core-event dependencies when needing to point to local source tree for both. 

```ruby
gem "logstash-core", :path => "/some/path/logstash/logstash-core"
gem "logstash-core-event", :path => "/some/path/logstash/logstash-core-event"
```

Note that this is not required but if you need to test your plugin against local sources for either logstash-core and/or logstash-core-event, you need to add them individually.

colinsurprenant:
provide instructions for running plugins specs against logstash-core-event-java

fixes #4175

### DIFF
--- a/logstash-core-event-java/README.md
+++ b/logstash-core-event-java/README.md
@@ -1,6 +1,6 @@
 # logstash-core-event-java
 
-## dev install
+## Dev install
 
 1- build code with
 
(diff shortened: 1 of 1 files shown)


suyograo:
Closing because this is super old and no longer relevant.

PR #4483, opened 2016-01-13
Clone params for output plugins to fix validation behavior

andrewvc:
The validator for plugin will mutate the params in place, causing
the various output instances to share an instance of the codec!
This is obviously dangerous. This patch restores the previous behavior

Fixes https://github.com/logstash-plugins/logstash-output-tcp/issues/13#issuecomment-171389649


andrewvc:
This is a better fix: https://github.com/elastic/logstash/pull/4473

Issue #4480, opened 2016-01-13
default to logstash-core-event-java implementation


colinsurprenant:
switch to use the core Java Event implementation by default. 

For this we need to complete the following:
- [x] jar-dependencies cleanups in logstash-core-event-java #4176 
- [x] add logstash-core-event-java specific build requirements in the build rake tasks #4479
- [x] default to java event and dependencies cleanups #4494

colinsurprenant:
all these are solved by #4494

colinsurprenant:
merging the result of #4563 in master and closing this issue. Starting from this merge master will default to Java Event and jenkins will be able to run tests in preparation for 2.3.

PR #4494, opened 2016-01-14
default to java event and dependencies cleanups

colinsurprenant:
- default to logstash-core-event-java
- removed dependency on Guava
- cleaned unused dependencies
- cleaned unused imports
- cleaned build.gradle
- removed java 8 code so it compiles on java 7
- use jar-dependencies to handle jackson dependency
- use json-unit for json related tests
- added gem jar creation and cleanup in build.gradle

for merging into feature/core-event-java shared branch. @guyboertje please review.

relates to #4480


colinsurprenant:
oh, we have to hook the gradle build in the rake tasks, will do.

colinsurprenant:
AH WOOT! build success!

logstash-core-event-java/build.gradle
```diff
@@ -1,104 +1,108 @@
...
-    classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.2'
-  }
+    repositories {
+        mavenLocal()
+        mavenCentral()
+        jcenter()
+    }
+    dependencies {
+        classpath 'net.saliman:gradle-cobertura-plugin:2.2.8'
+        classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.2'
```
talevy: should we shadow our jar?

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -50,7 +49,9 @@ public Event()
     public Event(Map data)
     {
         this.data = data;
-        this.data.putIfAbsent(VERSION, VERSION_ONE);
+        if (!this.data.containsKey(VERSION)) {
```
talevy: why not `putIfAbsent`?

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -203,8 +204,12 @@ public String toString() {
         try {
             return (getTimestamp().toIso8601() + " " + this.sprintf("%{host} %{message}"));
         } catch (IOException e) {
-            String host = (String)this.data.getOrDefault("host", "%{host}");
```
talevy: same question here as with the `putIfAbsent`.

logstash-core-event-java/src/main/java/com/logstash/Event.java
talevy: oh, nevermind. for Java7

logstash-core-event-java/src/main/java/com/logstash/Event.java
talevy: oh, nevermind. for Java7

logstash-core-event-java/src/main/java/com/logstash/KeyNode.java
```diff
@@ -39,4 +39,20 @@ public String evaluate(Event event) throws IOException {
             return "%{" + this.key + "}";
         }
     }
+
+    // TODO: (colin) this should be moved somewhere else to make it reusable
+    //   this is a quick fix to compile on JDK7 a not use String.join that is
+    //   only available in JDK8
+    public static String join(List<?> list, String delim) {
```
talevy: can we move this to a utility class since it is more generic than just `KeyNode`?

logstash-core-event-java/build.gradle
colinsurprenant: if you mean the jar we produce (org.logstash namespace) no, I don't see any good reason to shadow

logstash-core-event-java/src/main/java/com/logstash/Event.java
colinsurprenant: +1

logstash-core-event-java/build.gradle
talevy: then, can't we remove this? or am I missing somewhere else this is potentially being used.

logstash-core-event-java/build.gradle
```diff
@@ -1,104 +1,108 @@
...
+    testCompile 'junit:junit:4.12'
+    testCompile 'net.javacrumbs.json-unit:json-unit:1.9.0'
+}
 
 // See http://www.gradle.org/docs/current/userguide/gradle_wrapper.html
 task wrapper(type: Wrapper) {
-  description = 'Install Gradle wrapper'
-  gradleVersion = '2.7'
+    description = 'Install Gradle wrapper'
+    gradleVersion = '2.7'
```
talevy: I don't think we should bump this for this PR, but what do you think of keeping gradle versions in sync between us and elasticsearch? es depends on `2.8` now.

logstash-core-event-java/src/main/java/com/logstash/KeyNode.java
colinsurprenant: yeah, comment says that. will you force me to do it now? ;) it's just that I think we should probably do a namespace cleanup at some point and I didn't want to commit to a Utils class (which I really dislike in general because they often end-up as a bag of everything...)

logstash-core/logstash-core.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
   gem.version       = LOGSTASH_CORE_VERSION
 
-  gem.add_runtime_dependency "logstash-core-event", "~> 3.0.0.dev"
+  gem.add_runtime_dependency "logstash-core-event-java", "~> 3.0.0.dev"
```
talevy: w00t!

talevy:
left some comments, looks good!

logstash-core-event-java/build.gradle
colinsurprenant: oh, totally. I will, missed the new release.

colinsurprenant:
@talevy thanks! let me know what you think about the `join` method ...

logstash-core-event-java/src/main/java/com/logstash/KeyNode.java
talevy: sure, since it is only used here, let's leave it here.
can we add tests for it?

colinsurprenant: there is an indirect test on it in the StringInterpolationTest, it is definitely not complete as far as specifically testing the `join` method per-se. But yeah, properly testing join is a good idea, and we can move those tests whenever we refactor it.

logstash-core-event-java/build.gradle
colinsurprenant: no, AFAIK this is a leftover from what you originally put in place. will remove.

logstash-core-event-java/src/main/java/com/logstash/FieldReference.java
```diff
@@ -11,7 +11,7 @@
     private List<String> path;
     private String key;
     private String reference;
-    private static List<String> EMPTY_STRINGS = new ArrayList(Arrays.asList(new String[]{""}));
+    private static List<String> EMPTY_STRINGS = Arrays.asList(new String[]{""});
```
talevy: why create an array first?

does something like this work?
```
Arrays.asList("")
```

logstash-core-event-java/src/main/java/com/logstash/KeyNode.java
talevy: +1 unit tests for `join`

logstash-core-event-java/src/main/java/com/logstash/FieldReference.java
colinsurprenant: err, right :P

logstash-core-event-java/src/test/java/com/logstash/KeyNodeTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import static org.junit.Assert.assertEquals;
+
+public class KeyNodeTest {
```
talevy: can `value` that is extracted from the keynode be `null`? if so, I think `toString` on it would fail. should we add a test for this if it is possible?

colinsurprenant: value per se cannot be null because:

```java
   if (value != null) {
            if (value instanceof List) {
                return join((List)value, ",");
     
```

but that does not account for the List content, if it contains `null` then `join` would crash. I am not sure if it's possible for the List to contain nulls, @ph? otherwise it would be simple to add a `null` test in `join` to return `""` but I am not sure if that makes sense, specifically for the case where having `null`s in the List would indicate a bug in which case we could argue that crashing in `join` would be the good thing.

ph: I need to check went it can be a list.

colinsurprenant: Ok, so looking at the code, a `List` `value` is and `Event` field content which is an array. The question is, can Event array fields contain null values?

ph: Well I think this is effectively a bug in our logic.

lets take the ruby implementation of the following:
```ruby 
["foo", nil, "bar"].join(",")
=> "foor,,bar"
```

ph: So I think its fair to assume that the list in an event key can be null and minic the previous ruby behavior.

ph: I think `String.join` will use an empty string if the element is null, I havent tested it.

ph: From the doc https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.CharSequence...-

``` 
Note that if an element is null, then "null" is added.

```

ph: I propose we add the check in our `join` implementation to use an empty string when the item is null. WDYT?

colinsurprenant: I think it makes sense. I will verify to see if we have test cases for list null values and add one if its missing and run it on both the ruby and java implementations and make sure the java impl is consistent.

ph: +1

colinsurprenant: I've added spec cases in #4561 which clarifies current behaviour. will adjust the join method accordingly.

colinsurprenant:
moving this PR to #4563 - I mistakenly merged into the shared branch, and un-merged to correct mistake but now github won't believe it's not merged...

PR #4496, opened 2016-01-15
[rake] Fix broken doc generation

suyograo:
Doc generation was not updated after the changes to event and logstash-core
lib dir structure. Also, the rake task now ignores plugins whose doc cannot
be generated instead of bailing out in entirely, and having to start from
scratch again.

### DIFF
--- a/docs/asciidocgen.rb
+++ b/docs/asciidocgen.rb
@@ -3,7 +3,7 @@ require "erb"
 require "optparse"
 
 $: << Dir.pwd
-$: << File.join(File.dirname(__FILE__), "..", "lib")
+$: << File.join(File.dirname(__FILE__), "..", "logstash-core/lib")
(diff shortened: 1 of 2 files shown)


ph:
`bin/logstash docgen` doesnt exist for me in this branch?
Actually when I try to run the task I only get 

```
~/e/logstash git:pr/4496 ❯❯❯ rake docs:generate-docs                                                                                                                                                                              ✱ ◼
Generated asciidoc is under dir LS_HOME/asciidoc_generated
ERROR: too many arguments

See: 'bin/logstash --help'
ERROR: too many arguments

See: 'bin/logstash --help'
ERROR: too many arguments

See: 'bin/logstash --help'
```
I think the rake task is not calling the right script.

ph:
Shouldnt it be `docs/asciidocgen.rb` instead of `bin/logstash`?

PR #4498, opened 2016-01-15
MetricType: Gauge

### LINKED ISSUE #4488: MetricType Gauge
Currently the metric collector only support the Counter type, we need to add support for a Gauge Type to support adhoc values, theses values could be:

- The size of queue 
- an eviction cache
- an uptime
- Number of threads
- a status?

ph:
Add a gauge type for the metric collector, this is useful for values
like size of a queue, actual memory

Fixes #4488


purbon:
LGTM

ph:
I love how easy it is to add something :D

ph:
@purbon You can change your JVM branch to use that instead.

purbon:
@ph sure.

Issue #4401, opened 2015-12-28
Use a settings file to complement command line argument parsing


jsvd:
Currently logstash lists the following command line arguments:

```
    -f, --config CONFIG_PATH      Load the logstash config from a specific file
                                  or directory.  If a directory is given, all
                                  files in that directory will be concatenated
                                  in lexicographical order and then parsed as a
                                  single config file. You can also specify
                                  wildcards (globs) and any matched files will
                                  be loaded in the order described above.
    -e CONFIG_STRING              Use the given string as the configuration
                                  data. Same syntax as the config file. If no
                                  input is specified, then the following is
                                  used as the default input:
                                  "input { stdin { type => stdin } }"
                                  and if no output is specified, then the
                                  following is used as the default output:
                                  "output { stdout { codec => rubydebug } }"
                                  If you wish to use both defaults, please use
                                  the empty string for the '-e' flag.
                                   (default: "")
    -w, --filterworkers COUNT     Sets the number of filter workers to run.
                                   (default: 0)
    -l, --log FILE                Write logstash internal logs to the given
                                  file. Without this flag, logstash will emit
                                  logs to standard output.
    -v                            Increase verbosity of logstash internal logs.
                                  Specifying once will show 'informational'
                                  logs. Specifying twice will show 'debug'
                                  logs. This flag is deprecated. You should use
                                  --verbose or --debug instead.
    --quiet                       Quieter logstash logging. This causes only 
                                  errors to be emitted.
    --verbose                     More verbose logging. This causes 'info' 
                                  level logs to be emitted.
    --debug                       Most verbose logging. This causes 'debug'
                                  level logs to be emitted.
    -V, --version                 Emit the version of logstash and its friends,
                                  then exit.
    -p, --pluginpath PATH         A path of where to find plugins. This flag
                                  can be given multiple times to include
                                  multiple paths. Plugins are expected to be
                                  in a specific directory hierarchy:
                                  'PATH/logstash/TYPE/NAME.rb' where TYPE is
                                  'inputs' 'filters', 'outputs' or 'codecs'
                                  and NAME is the name of the plugin.
    -t, --configtest              Check configuration for valid syntax and then exit.
    --[no-]allow-unsafe-shutdown  Force logstash to exit during shutdown even
                                  if there are still inflight events in memory.
                                  By default, logstash will refuse to quit until all
                                  received events have been pushed to the outputs.
                                   (default: false)
    -h, --help                    print help
```

[...]

PR #4499, opened 2016-01-15
[WIP] implementation of the settings.yml file

jsvd:

TODO
- [ ] configurable location for settings.yml
- [ ] documentation

This PR targets the currently empty (== to master) feature branch https://github.com/elastic/logstash/tree/feature/settings_file


lib/bootstrap/environment.rb
```diff
@@ -64,11 +89,23 @@ def pattern_path(path)
...
   # TODO deprecate these arguments in the next major version. use -i only
   if ARGV == ["irb"] || ARGV == ["pry"]
     puts "Warn: option \"#{ARGV.first}\" is deprecated, use \"-i #{ARGV.first}\" or \"--interactive=#{ARGV.first}\" instead"
     LogStash::Runner.run("bin/logstash", ["--interactive", ARGV.first])
   else
-    LogStash::Runner.run("bin/logstash", ARGV)
+    # The Clamp library supports specifying the same argument multiple times
+    # and it keeps the   in an array. So in order for cli args to override
+    # the settings.yml args we can do `settings_array + ARGV`
+    LogStash::Runner.run("bin/logstash", settings_array + ARGV)
```
purbon: I think I would make ```settings_array+ARGV``` something more clear, might be though passing some method here? on the other side, to unify the two arrays, is ```+``` the best option in terms of performance?

logstash-core/lib/logstash/environment.rb
```diff
@@ -1,7 +1,27 @@
 # encoding: utf-8
 require "logstash/errors"
+require "logstash/config/cpu_core_strategy"
 
 module LogStash
+
+  DEFAULT_SETTINGS = {
+    "node.name" => Socket.gethostname,
```
purbon: would it be good to make this key strings also constants somewhere? what do you think?

purbon: does ```node.name``` means in the yaml something like:

```
- node:
   - name: foo
```

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -156,15 +142,15 @@ def start_workers
...
+      batch_size = @settings["pipeline.batch.size"]
+      batch_delay = @settings["pipeline.batch.delay"]
       max_inflight = batch_size * pipeline_workers
       @logger.info("Starting pipeline",
-                   :id => self.pipeline_id,
-                   :pipeline_workers => pipeline_workers,
-                   :batch_size => batch_size,
-                   :batch_delay => batch_delay,
-                   :max_inflight => max_inflight)
+                   "id" => self.pipeline_id,
```
purbon: why strings vs simbols? do we have a uniform policy to use one or another? or we just use both ?

lib/bootstrap/environment.rb
```diff
@@ -56,6 +56,31 @@ def pattern_path(path)
   end
 end
 
+def flatten_hash(h,f="",g={})
```
purbon: are this two more utility methods? what do you think?

lib/bootstrap/environment.rb
```diff
@@ -64,11 +89,23 @@ def pattern_path(path)
 if $0 == __FILE__
   LogStash::Bundler.setup!({:without => [:build, :development]})
   require ARGV.shift
+
+  if settings = YAML.parse(IO.read("settings.yml"))
+    settings = settings.to_ruby
```
purbon: I'm curious about why ```.to_ruby``` might be need here. any reason?

purbon:
are we have a default for all settings? or is there anyone we foresee to be required ?

lib/bootstrap/environment.rb
jsvd: :+1: on giving `settings_array` a better name.
the idea here is to push to the runner both the settings from the yaml file and the settings from the CLI, so I don't see a problem with `+`. Also we can sacrifice performance for readability here since it will only execute once at startup.

logstash-core/lib/logstash/environment.rb
jsvd: making all those strings constants would introduce another block of code with just constant definitions, create a level of indirection and make things harder to read (IMO).

This PR introduces settings that can be either expressed as a single string "pipeline.batch.size" or in hierarchical fashion:

```yaml
pipeline:
  batch:
    size: 6
```

for examples see the included `settings.yml`

logstash-core/lib/logstash/pipeline.rb
jsvd: I don't believe we have a policy.
The reason I chose strings here is the flexibility of them vs symbols, specially since we want to be able to have settings named "pipeline.batch.delay" or "node.name". doing this with symbols would be cumbersome `:"node.name"`

Also, none of this code will be in the critical path performance-wise, so we can afford strings easily.

lib/bootstrap/environment.rb
jsvd: :+1: I was struggling to decide where to put them, I'll experiment moving them to `LogStash::Util`

lib/bootstrap/environment.rb
jsvd: YAML.parse returns an awkward `Psych::Nodes::Document` object that can be navigated to obtain the structure of the yaml file. `to_ruby` just converts the structure into a simpler ruby object. in this case, a hash of hashes.

jsvd:
@purbon I couldn't understand the question :(

 Some settings will never have a "semantic" default. e.g. "config.path" doesn't have a default. you COULD say it is `nil`, which is the same as not having, IMO

lib/bootstrap/environment.rb
andrewvc: interested, good choice!

lib/bootstrap/environment.rb
```diff
@@ -64,11 +63,23 @@ def pattern_path(path)
 if $0 == __FILE__
   LogStash::Bundler.setup!({:without => [:build, :development]})
   require ARGV.shift
+
+  if settings = YAML.parse(IO.read("settings.yml"))
```
andrewvc: Leaving a note here to make sure we read from a more robust path in a subsequent commit (maybe LS_HOME/settings.yml)

lib/bootstrap/environment.rb
```diff
@@ -64,11 +63,23 @@ def pattern_path(path)
...
   # TODO deprecate these arguments in the next major version. use -i only
   if ARGV == ["irb"] || ARGV == ["pry"]
     puts "Warn: option \"#{ARGV.first}\" is deprecated, use \"-i #{ARGV.first}\" or \"--interactive=#{ARGV.first}\" instead"
     LogStash::Runner.run("bin/logstash", ["--interactive", ARGV.first])
   else
-    LogStash::Runner.run("bin/logstash", ARGV)
+    # The Clamp library supports specifying the same argument multiple times
+    # and it keeps the   in an array. So in order for cli args to override
+    # the settings.yml args we can do `settings_from_yml + ARGV`
+    LogStash::Runner.run("bin/logstash", settings_from_yml + ARGV)
```
andrewvc: niiiice

logstash-core/lib/logstash/environment.rb
andrewvc: I'm with @jsvd on just using a string. I see what you're going for with constants @purbon but I think its just a little too finnicky here IMHO

logstash-core/lib/logstash/pipeline.rb
andrewvc: I generally agree with @purbon here. It'd be nice to have a policy, but until we finish the mythical style guide I'm inclined to just let the author pick. Strings are fine with me (Though `:"foo.bar"` is a valid symbol).

logstash-core/lib/logstash/runner.rb
```diff
@@ -379,4 +370,37 @@ def start_shell(shell, start_binding)
       fail(I18n.t("logstash.runner.invalid-shell"))
     end
   end
-end # class LogStash::Runner
+
+  def build_settings_hash
+    hash = {}
+    self.class.declared_options.each do |opt|
+      option_name = opt.long_switch.sub("--", "").sub("[no-]", "")
```
andrewvc: :see_no_evil:

logstash-core/lib/logstash/util.rb
```diff
@@ -200,4 +200,31 @@ def self.deep_clone(o)
       Marshal.load(Marshal.dump(o))
     end
   end
+
+
+  def self.flatten_hash(h,f="",g={})
```
andrewvc: this could use either more clear variables names or a comment

logstash-core/lib/logstash/util/worker_threads_default_printer.rb
```diff
@@ -6,8 +6,8 @@
 module LogStash module Util class WorkerThreadsDefaultPrinter
 
   def initialize(settings)
-    @setting = settings.fetch(:pipeline_workers, 0)
-    @default = settings.fetch(:default_pipeline_workers, 0)
+    @setting = settings.fetch('pipeline.workers', 0)
```
andrewvc: .... from strings, to symbols, and back again! .... ( I actually don't care, but it is funny :) )

andrewvc:
Generally LGTM, though I think we could use a little testing around the parameter merging.

lib/bootstrap/environment.rb
suyograo: See comments https://github.com/elastic/logstash/issues/4401#issuecomment-172652928

ph:
I can see a use case to read multiples `yml` files and merge the contains of the keys.

Let say I have the following files:

```
config/settings.yml
config/metrics.yml
config/metrics.exporters.yml # This file is copied after installing plugin, which is easier than editing the metrics.yml
```

It is something we should support?

purbon:
+1 for this too,
   On the other side, have we thought on adding some validation for
keywords? Something like unnexpected key for metrics, or is something more
of the metrics component?

/purbon

suyograo:
@ph IMO, we should not support merging of settings file. Its too confusing. Lets keep it simple :)

suyograo:
Also, please lets call this `logstash.yml`. `conf/settings.yml` is too generic.

ph:
@suyograo Okay for not merging files.

ph:
@jsvd what is needed to move this forward another review?

jsvd:
@ph currently I'm adding documentation that describes this new feature, but the code can be reviewed as is. the only missing feature is a way to configure the location of logstash.yml

ph:
@jsvd I will do a review since I need some kind of configuration for the metrics!

jsvd:
closing. reopening this against master

PR #4512, opened 2016-01-18
(Work in progress) Nodes API

purbon:
Work in progress ```_nodes``` api example as will be defined and requested during in #4446 . As a work in progress PR this will change to adapt it to the final decision on API's design.


purbon:
closing this in benefit of #4641

PR #4518, opened 2016-01-19
init: Remove overwrite of JAVA_OPTS

### LINKED ISSUE #4517: Init script: JAVA_OPTS are getting overwritten by LS_JAVA_OPTS
https://github.com/elastic/logstash/blob/master/pkg/logstash.sysv#L144

Intended behavior: (See https://github.com/elastic/logstash/blob/master/bin/logstash#L11)

> LS_JAVA_OPTS="xxx" to append extra options to the defaults JAVA_OPTS provided by logstash
   JAVA_OPTS="xxx" to *completely override* the defauls set of JAVA_OPTS provided by logstash

suyograo:
Fixes #4517

### DIFF
--- a/pkg/logstash.sysv
+++ b/pkg/logstash.sysv
@@ -141,7 +141,6 @@ configtest() {
     return 1
   fi
 
-  JAVA_OPTS=${LS_JAVA_OPTS}
   HOME=${LS_HOME}
   export PATH HOME JAVA_OPTS LS_HEAP_SIZE LS_JAVA_OPTS LS_USE_GC_LOGGING
 


pkg/logstash.sysv
```diff
@@ -141,7 +141,6 @@ configtest() {
     return 1
   fi
 
-  JAVA_OPTS=${LS_JAVA_OPTS}
   HOME=${LS_HOME}
   export PATH HOME JAVA_OPTS LS_HEAP_SIZE LS_JAVA_OPTS LS_USE_GC_LOGGING
```
purbon: we should also remove ```JAVA_OPTS``` from here.

purbon:
Add only small comment, in general LGTM.

pkg/logstash.sysv
suyograo: Why? A user can supply `JAVA_OPTS` which replaces existing one or merge java options by providing `LS_JAVA_OPTS`, right?

purbon: but then why do we need to export it here? I understand we need to export it if we actually redefined itself here.

the options you are talking about are exported in https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L90, I don't see why se need to export it here too, is redundant to my understanding.

jordansissel: We need to export here because if JAVA_OPTS isn't set at all, we set it here: https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L56 - and it not being set previously means we know it isn't exported as an environment variable, so we have to tell the shell to export it so that it is passed down to the JRuby executable - https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L181

It's possible that we should just export it on [this line](https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L66) when we're in the conditional knowing that JAVA_OPTS isn't set. But either way we get the correct behavior.

purbon: I see, so for my understanding https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L90 will do what you say, but only when using the vendored ruby, not when using ```USE_RUBY```, right? I might be missing something, but....

keeping works for me, we can always remove it later if possible.

jordansissel: > A user can supply JAVA_OPTS

if they're supplying it, aren't they supplying it from the environment so there's no need to do `export JAVA_OPTS` again? I think this is mostly a style-and-intent concern, not a functional problem, though.

Issue #4513, opened 2016-01-18
Apply changes from configuration file without restarting


suyograo:
Today, any config changes made to LS pipeline requires a process restart, which is not ideal. Plan is to implement a feature which would track config file for changes and restart the pipeline (same process) with updated changes. This feature can be enabled by passing a CLI option `--auto_reload` with `--reload_interval` which controls how often LS should check the config files for changes.

djschny:
For service oriented apps I wrote in the past that had this behavior, I expanded to include the refreshing from any URL, that way one easy way to get centralized config management was via a web server or similar. For example:
* file://
* http://
* https://
* ftp://

If you're in Java land, accomplishing this is pretty easy too using something similar to `new URL("<my_config_url>").openStream()` and regardless of what protocol you're given a stream to read from.

jsvd:
@djschny we support the same in logstash through open-uri https://github.com/elastic/logstash/pull/4520/files#diff-5e1ebf137d1eab8a085fa7c1c4a6fd2cR37
However we're restricting to https and file (for no go reason).

jsvd:
merged in master:

Branch | Commits
-------|--------
master | a90ed32d6d49d6150b23109c69e7ff6be3016d0f, e38d815ecb57153ef409abfe7bae6c2af0a177c6, 34ed3c34758eb0c280882ee3f34b13a9a7048bad, 7ee5038ccf8f93b54f96d12d6247a79e9405a14d, e32f732dba9e73e718fc25b0490e8f547d05ba93, 7d8d6e4c7ad088026bcfb0e6392d6c8e994dd365

I'll open a new PR to add this to 2.x since the branches have greatly diverged

suyograo:
@jsvd great stuff

lawre:
How does the dynamic reload work with systemd service?  I perused /etc/sysconfig/logstash and saw no variables corresponding to 'auto_reload' or 'reload_interval'

jsvd:
in `#LS_OPTS=""` you can set `LS_OPTS="-r"` to have auto reloading

jsvd:
merged in 2.x

Branch | Commits
-------|--------
2.x | 00995378682e90ff50eb3ed244838e9530997294

PR #4520, opened 2016-01-19
Agent config reloading

jsvd:
implements configuration reloading through SIGHUP and the "-r" flag

this pr is based off of the [Pluggable Agents PR](https://github.com/elastic/logstash/pull/4194) sans the AgentRegistry mechanism

fixes https://github.com/elastic/logstash/issues/4513

This needs to be backported to 2.x branch (for 2.3.0)


logstash-core/lib/logstash/agent.rb
```diff
@@ -11,32 +12,78 @@
...
+    # a subclass of this base Agent could use more complex state objects such
+    # as a hash map containing versioning, update_at timestamp, etc.
+    @state = clean_state
   end
 
   def execute
-    # Make SIGINT/SIGTERM shutdown the pipeline.
-    sigint_id = trap_sigint()
-    sigterm_id = trap_sigterm()
+    @thread = Thread.current
```
andrewvc: Is this used anywhere?

jsvd: Yes by Stud.stop?/!

andrewvc: That could be documented with a comment, its kind of confusing if you don't know why it's being set.

jordansissel: +1. Also maybe adding an api to Stud that makes it more obvious, instead of an ivar setting?

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -1,27 +1,135 @@
...
+    end
+
+    after :each do
+      File.unlink(tmp_config_path)
+    end
+
+    it "should join the config string and config path content" do
+      expect(subject.send(:fetch_state).strip).to eq(cli_config + IO.read(tmp_config_path))
+    end
+
```
andrewvc: Good job on all the tests here!

andrewvc:
Looks LGTM! When it moves out of WIP status let me know and I'll hand test it locally.

logstash-core/lib/logstash/agent.rb
```diff
@@ -11,32 +12,75 @@
...
 
   def execute
-    # Make SIGINT/SIGTERM shutdown the pipeline.
-    sigint_id = trap_sigint()
-    sigterm_id = trap_sigterm()
+    @thread = Thread.current # this var is implicilty used by Stud.stop?
+    @logger.info("starting agent", :state => @state)
+
+    reload_state!
+    return 1 if clean_state?
```
andrewvc: This needs a log message explaining the error. Perhaps: "No config could be found! Will exit."

logstash-core/lib/logstash/agent.rb
```diff
@@ -11,32 +12,75 @@
...
+    shutdown_pipelines
+  end
 
-    @pipelines.each {|_, p| p.run } # blocking operation. works now because size <= 1
-    return 0
+  def reload_state!
+    new_state = fetch_state
+
+    if (reason = valid_state?(new_state)) != true
+      @logger.error("invalid state", :state => new_state, :reason => reason)
```
andrewvc: Hmmm, will this always print in the `return 1` case? If so this is a fine error.

The confusing thing about using the 'state' verbiage, is that the user is dealing with config files. `state` is really an internal term. Thoughts?

jordansissel:
I will review this tomorrow also <3

jsvd:
I stumbled upon a severe limitation of this implementation. the agent now explicitly handles a single pipeline called "main", instead of supporting multiple pipeline states.

I'm adding a new api to the agent `register_pipeline(pipeline_id, pipeline_settings)` which will be added to the agent state and then processed in the state machine.

logstash-core/lib/logstash/agent.rb
```diff
@@ -11,78 +12,165 @@
...
 
-  def trap_sigterm
-    Stud::trap("TERM") do
-      @logger.warn(I18n.t("logstash.agent.sigterm"))
-      shutdown_pipelines
+  def fetch_config(settings)
+    @config_loader.format_config(settings[:config_path], settings[:config_string])
+  end
+
+  def reload_pipeline!(id)
```
andrewvc: I'd add a comment mentioning which mutex you need to use this function safely.

logstash-core/lib/logstash/agent.rb
```diff
@@ -11,78 +12,165 @@
...
-        @logger.fatal(I18n.t("logstash.agent.forced_sigint"))
-        exit
-      else
-        @logger.warn(I18n.t("logstash.agent.sigint"))
-        Thread.new(@logger) {|logger| sleep 5; logger.warn(I18n.t("logstash.agent.slow_shutdown")) }
-        @interrupted_once = true
-        shutdown_pipelines
+  def start_pipeline(id)
+    pipeline = @pipelines[id]
+    return unless pipeline.is_a?(LogStash::Pipeline)
```
andrewvc: What else would this be? Should this raise an exception instead?

logstash-core/lib/logstash/agent.rb
```diff
@@ -11,78 +12,165 @@
...
+
+  def running_pipelines?
+    @upgrade_mutex.synchronize do
+      @pipelines.select {|pipeline_id, _| running_pipeline?(pipeline_id) }.any?
+    end
+  end
+
+  def running_pipeline?(pipeline_id)
+    thread = @pipelines[pipeline_id].thread
+    thread.is_a?(Thread) && thread.alive?
```
andrewvc: Why a class check? What else could this be? Maybe this should be:

```ruby
thread = @pipelines[pipeline_id].thread
thread ? thread.alive? : nil # Ternary SQL-like truth!
```

andrewvc:
Looks fantastic! Generally LGTM, though I think the areas where I left some comments could use some work.

Interestingly I think I found a bug in stdin where this won't work with this: https://github.com/logstash-plugins/logstash-input-stdin/pull/9

logstash-core/lib/logstash/agent.rb
jsvd: it's a way of dealing with a miss, it could be just `return unless pipeline`.

logstash-core/lib/logstash/agent.rb
jsvd: if sometimes prefer using `object.is_a?(Class)` instead of `!object.nil?` because it's both a tighter check and also documents that that object's class is supposed to be.
I can make this just `thread && thread.alive` if you like

jsvd:
@andrewvc yeah, stdin is a pain in the butt, @colinsurprenant created https://github.com/logstash-plugins/logstash-input-stdin/pull/3 to make the stdin interruptible and closable.
eventually this needs to be merged in, though I'm not sure it's ready to deal with the scenario where we need to open and close stdin multiple times.

Also, with multiple pipelines...stdin is a big nono...

jordansissel:
Did some quick tests with a simple config (tcp input, stdout output) and I like the behavior :)

logstash-core/lib/logstash/agent.rb
```diff
@@ -11,78 +12,167 @@
...
-      shutdown_pipelines
+  def fetch_config(settings)
+    @config_loader.format_config(settings[:config_path], settings[:config_string])
+  end
+
+  # since this method modifies the @pipelines hash it is
+  # wrapped in @upgrade_mutex in the parent call `reload_state!`
+  def reload_pipeline!(id)
+    old_pipeline = @pipelines[id]
+    new_pipeline = create_pipeline(old_pipeline.original_settings)
```
jordansissel: The way I read this, I kind of read it as "Create a pipeline based on the old pipeline" -- and this feels weird. Is `pipeline.original_settings` a good way to pass settings around?

logstash-core/lib/logstash/config/loader.rb
```diff
@@ -0,0 +1,99 @@
+require "logstash/config/defaults"
+
+module LogStash; module Config; class Loader
+  def initialize(logger, is_config_test)
+    @logger = logger
+    @is_config_test = is_config_test
```
jordansissel: Hmm.. This flag (is_config_test) feels more like a flag controlling the debug level of the loader. If so, we can probably use the @logger.debug? method for that. Thoughts?

jordansissel:
@andrewvc Can you check into why Travis fails to test this? It seems like it's trying to use the wrong version of ruby somehow?

logstash-core/lib/logstash/agent.rb
jsvd: I'm glad you noticed this. I struggled a bit to find a place the pipeline settings needed for regenerating the pipeline. 
There are 2 sets of settings for the pipeline: the fetching instructions and the parametrization.
The pipeline instance must own the parametrization settings for its runtime, but not the fetching settings (like the cli args `config.path` and `config.string`) since the constructor receives the final "pipeline description string".

So I first tried with a hash that contained pipeline_id, fetch_settings and pipeline instance:

```ruby
@pipeline = {
  "pipelineid1" => {
     :pipeline => LogStash::Pipeline,
     :settings => { "config.path" => "/tmp/config" , ... },
  "pipelineid2" => {
      ...
  }
}
```

But creating a new Pipeline still required getting the runtime settings from the old pipeline and merging with the fetch settings so I opted to put them all "inside".
Also this hash made it awkward to interact with the pipeline object (e.g. `@pipelines["pipelineid1"].pipeline.shutdown`)

I therefore chose to have just:

```ruby
@pipeline = {
  "pipelineid1" => LogStash::Pipeline,
  "pipelineid2" => LogStash::Pipeline
}
```

An alternative here is to have a...*shudder*..pipeline factory class, that harbors the settings and produces pipelines, given the factory's settings.

andrewvc:
@jordansissel I made a mistake with travis. It was only supposed to be testing my travis-ci branch/PR but it wound up applying to the whole repo and used its own default ruby test config. I disabled that behavior, now it only tests if a .travis.yml is present, but some PRs like this one have stuck red statuses :( . I'm realizing this is a problem for Jarvis, not sure that to do.

logstash-core/lib/logstash/config/loader.rb
jsvd: +1 on changing this, I see no reason not to show this when debug is enabled, even if it's not a configuration test.

jordansissel:
LGTM.

(I expect some integration problems to be found after this merges, but the tests are passing for me and my manual tests look pretty awesome)

jsvd:
I'll create a new PR to add this to 2.x

logstash-core/lib/logstash/agent.rb
```diff
@@ -4,85 +4,175 @@
...
+  end
+
+  def running_pipeline?(pipeline_id)
+    thread = @pipelines[pipeline_id].thread
+    thread.is_a?(Thread) && thread.alive?
   end
+
+  def upgrade_pipeline(pipeline_id, new_pipeline)
+    stop_pipeline(pipeline_id)
+    @pipelines[pipeline_id] = new_pipeline
```
beidou2018, line 170: I'm just curious, when replacing the old pipeline with a new one, have the events in the old pipeline's queue been fully processed? If not, will those events be lost?

PR #4525, opened 2016-01-19
Add a `Plugin#id` that return an ID for a plugin

### LINKED ISSUE #3892: Add a unique ID on every plugin and allow user to override it.
In the configuration we can define multiples time the same plugin with different configurations, but to associate the metrics to a specific plugins we need to make sure we can correctly identify it. This ID has a side effect to allow us to update the configuration of a specific plugin if needed via an API..

I propose:
- By default every plugins (input, filter, output and codec) generate a unique ID.
- This ID should be still human readable to understand what it refer too
- Users should be allowed to override it.
- Logstash should be smart enough to raise an issue if two plugins has the same ID.
- If would be awesome if a codec ID contains information about the plugin defining it.
- `--debug` should show the ID in the comment.
- The ID need to be reproducable to associated with the same plugin between each restart.

This is a what a config would look like with user defined ids.
```
input {
 sqs {
  id => "fast-queue"
  codec  => json { id => "json-fast" }
 }
}
```

ph:
This method return an ID for the plugins and be configured by the users
in the configuration like this:

```
elasticsearch {
  id => "ABC"
  ...
}
```

This information will be used when collecting metrics for a specific
plugin. Allowing the user to change it allow to stick between restart.

Fixes #3892

### DIFF
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -20,7 +20,7 @@ module LogStash
       @filter = klass.new(options)
 
       # Scope the metrics to the plugin
-      @metric = metric.namespace(@filter.identifier_name)
+      @metric = metric.namespace(@filter.id.to_sym)
       @filter.metric = @metric
 
       define_flush_method if @filter.respond_to?(:flush)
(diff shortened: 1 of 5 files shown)


ph:
@jordansissel So I changed the logic a bit of the `#id`

By default if the user has not configured anything we will return a UUID.
The user can override this to return something more meaning for them.

purbon:
I guess this will need some rebase here.

On the other side, I was not able to see the change for inputs? is there other system for inputs?

for the rest, it LGTM so happy to be merged.

ph:
The inputs are covered higher directly in the pipeline.

Issue #4502, opened 2016-01-15
Enable Metrics collection by default


ph:
Currently the metrics are not enabled by default and users need to tell Logstash to collect metrics by using the `--metric` option.  We should change that to make sure the metrics are always on by default and we cannot turn them off. 

The reasoning behind this is to make sure we always have visibility inside logstash for debugging or monitoring.

PR #4526, opened 2016-01-19
Change in the default of the metric feature.

ph:
Metric will be on by default for the pipelines and the plugins,
Its not possible to turn it off for the pipeline but you can turn
off metrics collection for specific plugin.

fixes: #4502


purbon:
LGTM, only minor thingI still get

```
````

when starting the command, I thought this was already fix for master, what do you think?

ph:
> Settings: Default pipeline workers: 4
> Settings: Default pipeline workers: 4

Oh this is a good catch, create a specific issue for it, since we are starting 2 independant pipelines both of them are actually reporting that information.

PR #4533, opened 2016-01-20
Remove unnecessary register when doing deep clone for codecs

purbon:
@guyboertje review?

### DIFF
--- a/logstash-core/lib/logstash/util.rb
+++ b/logstash-core/lib/logstash/util.rb
@@ -193,7 +193,7 @@ module LogStash::Util
     when Fixnum, Symbol, IO, TrueClass, FalseClass, NilClass
       o
     when LogStash::Codecs::Base
-      o.clone.tap {|c| c.register }
+      o.clone
     when String
       o.clone #need to keep internal state e.g. frozen
     else
--- a/logstash-core/spec/logstash/plugin_spec.rb
+++ b/logstash-core/spec/logstash/plugin_spec.rb
@@ -144,9 +144,26 @@ describe LogStash::Plugin do
     ].each do |klass|
 
       it "subclass #{klass.name} does not modify params" do
-        instance = klass.new(args)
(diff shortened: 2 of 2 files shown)


guyboertje:
LGTM and thanks @purbon

purbon:
welcome! which branches should this be merged?

jsvd:
may I suggest a test that shows this actually fixes a bug?

purbon:
Sure,
   I would just check that register is call once when doing deep clone,
makes sense guy?

Is there any special problem this fix?

/purbon

jsvd:
On a side note, I believe we could change: ` when Fixnum, Symbol, IO, TrueClass, FalseClass, NilClass` to just `when IO`, all others can be Marshal.dump/load

purbon:
@jsvd I see you improvement as a perfect followup, good catch.

guyboertje:
@purbon, @jsvd 
> Is there any special problem this fix?
> may I suggest a test that shows this actually fixes a bug?
There is no bug.  Some codec may not like register called twice. The `deep_clone` method needs to as fast as possible and registering a second time is a waste of cycles.

> On a side note, I believe we could change: when Fixnum, Symbol, IO, TrueClass, FalseClass, NilClass to just when IO, all others can be Marshal.dump/load

- Fixnum, Symbol can't be cloned or duped
- Values of true, false and nil can't be mutated as such
- IO can't be Marshalled/serialised and we probably need the original, e.g. logger
- Marshal load dump is much more expensive than dup or clone.

This is why I return the original for Fixnum, Symbol, IO, TrueClass, FalseClass, NilClass and only use Marshalling as a last resort.

Merge to master, 2.2 and 2.x

purbon:
@guyboertje adding small test to check for this.

jsvd:
+1 that this is not a bug. thanks for the enlightenment @guyboertje

PR #4535, opened 2016-01-20
Common plugin registry to be reused around Logstash

purbon:
Introducing a common plugin registry class, this class will let us:

* Load current plugins, loaded from our configurations.
* Load any other kind of plugins we want to introduce for logstash.

This new module is more or less feature pare with the current plugin loading behaviour. The list of features are:

* Load plugins if installed as gems.
* Load plugins defined in code base (mostly used for test, _could we remove this?_)
* Raise error if not.

@andrewvc @jsvd @ph your thoughts are much appreciate it.

### DIFF
--- a/logstash-core/lib/logstash/codecs/base.rb
+++ b/logstash-core/lib/logstash/codecs/base.rb
@@ -7,7 +7,12 @@ require "logstash/logging"
 # This is the base class for logstash codecs.
 module LogStash::Codecs; class Base < LogStash::Plugin
   include LogStash::Config::Mixin
+
   config_name "codec"
+
+  def self.plugin_type
+    "codec"
+  end
 
   def initialize(params={})
     super
--- a/logstash-core/lib/logstash/plugin.rb
+++ b/logstash-core/lib/logstash/plugin.rb
@@ -6,6 +6,7 @@ require "logstash/instrument/null_metric"
 require "cabin"
 require "concurrent"
 require "securerandom"
+require "logstash/plugins/registry"
 
 class LogStash::Plugin
   attr_accessor :params
@@ -117,21 +118,19 @@ class LogStash::Plugin
   # Look up a plugin by type and name.
   def self.lookup(type, name)
     path = "logstash/#{type}s/#{name}"
-
-    # first check if plugin already exists in namespace and continue to next step if not
-    begin
-      return namespace_lookup(type, name)
-    rescue NameError
-      logger.debug("Plugin not defined in namespace, checking for plugin file", :type => type, :name => name, :path => path)
+    LogStash::Registry.instance.lookup(type ,name) do |plugin_klass, plugin_name|
+      is_a_plugin?(plugin_klass, plugin_name)
     end
-
-    # try to load the plugin file. ex.: lookup("filter", "grok") will require logstash/filters/grok
-    require(path)
-
-    # check again if plugin is now defined in namespace after the require
-    namespace_lookup(type, name)
   rescue LoadError, NameError => e
+    logger.debug("Problems loading the plugin with", :type => type, :name => name, :path => path)
     raise(LogStash::PluginLoadingError, I18n.t("logstash.pipeline.plugin-loading-error", :type => type, :name => name, :path => path, :error => e.to_s))
+  end
+
+  public
+  def self.declare_plugin(type, name)
+    path = "logstash/#{type}s/#{name}"
+    registry = LogStash::Registry.instance
+    registry.register(path, self)
   end
 
   private
@@ -165,4 +164,5 @@ class LogStash::Plugin
   def self.logger
     @logger ||= Cabin::Channel.get(LogStash)
   end
+
 end # class LogStash::Plugin
--- a/logstash-core/lib/logstash/plugins/registry.rb
+++ b/logstash-core/lib/logstash/plugins/registry.rb
@@ -1,0 +1,83 @@
+# encoding: utf-8
+require 'singleton'
+require "rubygems/package"
+
+module LogStash
+  class Registry
+
+    ##
+    # Placeholder class for registered plugins
+    ##
+    class Plugin
+      attr_reader :type, :name
+
+      def initialize(type, name)
+        @type  = type
+        @name  = name
+      end
+
+      def path
+        "logstash/#{type}s/#{name}"
+      end
+
+      def cannonic_gem_name
+        "logstash-#{type}-#{name}"
+      end
+
+      def installed?
+        find_plugin_spec(cannonic_gem_name).any?
+      end
+
+      private
+
+      def find_plugin_spec(name)
+        specs = ::Gem::Specification.find_all_by_name(name)
+        specs.select{|spec| logstash_plugin_spec?(spec)}
+      end
+
+      def logstash_plugin_spec?(spec)
+        spec.metadata && spec.metadata["logstash_plugin"] == "true"
+      end
+
+    end
+
+    include Singleton
+
+    def initialize
+      @registry = {}
+      @logger = Cabin::Channel.get(LogStash)
+    end
+
+    def lookup(type, plugin_name, &block)
+
+      plugin = Plugin.new(type, plugin_name)
+
+      if plugin.installed?
+        return @registry[plugin.path] if registered?(plugin.path)
+        require plugin.path
+        klass = @registry[plugin.path]
+        if block_given? # if provided pass a block to do validation
+          raise LoadError unless block.call(klass, plugin_name)
+        end
+        return klass
+      else
+        # The plugin was defined directly in the code, so there is no need to use the
+        # require way of loading classes
+        return @registry[plugin.path] if registered?(plugin.path)
+        raise LoadError
+      end
+    rescue => e
+      @logger.debug("Problems loading a plugin with", :type => type, :name => plugin, :path => plugin.path, :error => e) if @logger.debug?
+      raise LoadError, "Problems loading the requested plugin named #{plugin_name} of type #{type}."
+    end
+
+    def register(path, klass)
+      @registry[path] = klass
+    end
+
+    def registered?(path)
+      @registry.has_key?(path)
+    end
+
+  end
+end
--- a/logstash-core/lib/logstash/config/mixin.rb
+++ b/logstash-core/lib/logstash/config/mixin.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 require "logstash/namespace"
 require "logstash/config/registry"
+require "logstash/plugins/registry"
 require "logstash/logging"
 require "logstash/util/password"
 require "logstash/version"
@@ -191,8 +192,12 @@ module LogStash::Config::Mixin
     def config_name(name = nil)
       @config_name = name if !name.nil?
       LogStash::Config::Registry.registry[@config_name] = self
+      if self.respond_to?("plugin_type")
+        declare_plugin(self.plugin_type, @config_name)
+      end
       return @config_name
     end
+    alias_method :config_plugin, :config_name
 
     # Deprecated: Declare the version of the plugin
     # inside the gemspec.
--- a/logstash-core/lib/logstash/filters/base.rb
+++ b/logstash-core/lib/logstash/filters/base.rb
@@ -117,6 +117,10 @@ class LogStash::Filters::Base < LogStash::Plugin
   # Optional.
   config :periodic_flush, :validate => :boolean, :default => false
 
+  def self.plugin_type
+    "filter"
+  end
+
   public
   def initialize(params)
     super
--- a/logstash-core/lib/logstash/inputs/base.rb
+++ b/logstash-core/lib/logstash/inputs/base.rb
@@ -10,6 +10,7 @@ require "logstash/util/decorators"
 # This is the base class for Logstash inputs.
 class LogStash::Inputs::Base < LogStash::Plugin
   include LogStash::Config::Mixin
+
   config_name "input"
 
   # Add a `type` field to all events handled by this input.
@@ -47,6 +48,10 @@ class LogStash::Inputs::Base < LogStash::Plugin
 
   attr_accessor :params
   attr_accessor :threadable
+
+  def self.plugin_type
+    "input"
+  end
 
   public
   def initialize(params={})
--- a/logstash-core/lib/logstash/outputs/base.rb
+++ b/logstash-core/lib/logstash/outputs/base.rb
@@ -57,6 +57,10 @@ class LogStash::Outputs::Base < LogStash::Plugin
     self.class.declare_workers_not_supported!(message)
   end
 
+  def self.plugin_type
+    "output"
+  end
+
   public
   def initialize(params={})
     super
--- a/logstash-core/spec/logstash/plugin_spec.rb
+++ b/logstash-core/spec/logstash/plugin_spec.rb
@@ -34,6 +34,19 @@ describe LogStash::Plugin do
     expect(LogStash::Plugin.lookup("filter", "lady_gaga")).to eq(LogStash::Filters::LadyGaga)
   end
 
+  describe "plugin signup in the registry" do
+
+    let(:registry) { LogStash::Registry.instance }
+
+    it "should be present in the registry" do
+      class LogStash::Filters::MyPlugin < LogStash::Filters::Base
+        config_name "my_plugin"
+      end
+      path     = "logstash/filters/my_plugin"
+      expect(registry.registered?(path)).to eq(true)
+    end
+  end
+
   describe "#inspect" do
     class LogStash::Filters::MyTestFilter < LogStash::Filters::Base
       config_name "param1"
--- a/logstash-core/spec/logstash/plugins/registry_spec.rb
+++ b/logstash-core/spec/logstash/plugins/registry_spec.rb
@@ -1,0 +1,57 @@
+# encoding: utf-8
+require "spec_helper"
+require "logstash/plugins/registry"
+require "logstash/inputs/base"
+
+# use a dummy NOOP input to test plugin registry
+class LogStash::Inputs::Dummy < LogStash::Inputs::Base
+  config_name "dummy"
+
+  def register; end
+
+end
+
+describe LogStash::Registry do
+
+  let(:registry) { described_class.instance }
+
+  context "when loading installed plugins" do
+
+    let(:plugin) { double("plugin") }
+
+    it "should return the expected class" do
+      klass = registry.lookup("input", "stdin")
+      expect(klass).to eq(LogStash::Inputs::Stdin)
+    end
+
+    it "should raise an error if can not find the plugin class" do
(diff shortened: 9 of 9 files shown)


logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -0,0 +1,96 @@
...
+          klass = load_plugin(type, plugin, &block)
+          register(plugin_path, klass)
+          return klass
+        else # Plugin requested not currently installed.
+          @logger.debug("Requested plugin is not currently installed or defined", :type => type, :name => plugin, :path => plugin_path) if @logger.debug?
+          raise LoadError, "Requested plugin is not currently installed or defined" # (TODO) Need to be added to I18n when agreed on message.
+        end
+      end
+
+      def register(path, plugin)
```
andrewvc: Have you considered having register use the signature `register(type, plugin)`? 

It looks like all plugins need to have a path of `/logstash/#{type}s/#{plugin}`. Shouldn't we enforce this formatting here? We may also want to enforce plugin names, maybe with something like `/a-z[0-9_-]`

andrewvc:
The code here looks good! However, I'm not a fan of this method of plugin detection/registration.

This patch seems to use the current pattern of guessing class names and requiring logstash to explicitly lookup a plugin by name. In my last PR for the agent plugin registry I proposed having an explicit plugin register step, where the plugin itself registers with Logstash, rather than logstash 'discovering' the plugin. This would look like.

```ruby
class MyLogstashPlugin < LogStash::SomePluginClass
  LogStash::PluginRegistry(:my_plugin_type, "plugin_name")  

  def foo
  end
end
```

The advantages here are:

* Decouples the class names from the plugin names, giving plugin authors greater flexibility
* Allows for plugins that inject their behavior. Currently plugins must be explicitly referenced from a logstash config to be used.
* Allows for us to present the user with a list of valid plugins
* Does not rely on 'magic' detection via class names

I believe that we should support the old plugin loading mechanism for inputs/outputs/filters, but that this new one should use explicit registration.

ph:
> I believe that we should support the old plugin loading mechanism for inputs/outputs/filters, but that this new one should use explicit registration.

I strongly think this is something we should do, the layout structure is the last reminiscent of 1.4.2 structure.

logstash-core/lib/logstash/plugins/registry.rb
purbon: I think there is an small confusion here, the last argument of register is the plugin class, not the plugin name. But yes, I do agree this signature could be improved :+1:

purbon:
Thanks a lot for your feedback guys, much appreciate.

> I believe that we should support the old plugin loading mechanism for inputs/outputs/filters, but that this new one should use explicit registration.

I agree with you two @ph and @andrewvc, my very first idea was to actually support the former behaviour as this will help us not break compatibility with the way plugins work, but as you can see the registry have a public registry method, letting the user of it actually do what you propose @andrewvc . My intention if to add in this very same PR a couple of helpers that let plugin authors actually have something like

```
   class MyPlugin < BaseLogStashPlugin
        announce 
    end
```

or something alike, I see your way of doing @andrewvc require the user to pass actually some information we already know.

what do you think?

purbon:
@andrewvc Added an improvement re explicit registration for plugins, please let me know your thoughts about it. There are a few things I'm thinking about.

* Changes should be backwards compatible with current plugin system.
* Notation should be as easy and simply as possible.
* Avoid as much duplication as possible, this means avoiding user having to provide twice information we're already able to have.

There are a few improvements possible, but also have their challenges, for example changing the plugin path key for the very same. Happy to discuss pros and cons.

logstash-core/lib/logstash/plugins/mixin.rb
```diff
@@ -0,0 +1,18 @@
+# encoding: utf-8
+
+module LogStash
+  module Plugins
+    module Mixin
+
+      def annotate(name="")
+        if !name.empty?
+          super("filter", name)
```
andrewvc: Why does this only work to set "filter" type plugins? I assume that this mixin is only called directly like filters?

andrewvc:
Thanks for adding the explicit registration code @purbon , I think we can sugar it a bit more however.

Currently it looks something like 
```ruby
class LogStash::Filters::MyPlugin < LogStash::Filters::Base
  config_name "my_plugin"
  annotate
end
```

What do you think about having `config_name` do this work? The other nice thing about this is it lets us completely remove implicit plugin loading in a backward compatible way!

BTW, looking at the code `annotate` only works for filters so far, correct?

logstash-core/lib/logstash/plugins/mixin.rb
purbon: Good catch, for now I left this like this because I wanted to see if you liked the procedure. Easy to extend to others following this idea.

purbon:
@andrewvc updated this PR to reflect your last proposal, that I agree with! can you let me know what do you think?.

PD: I also introduced a plugin_type accessor for base classes, this is used to properly fetch the type and can lead to some simplification later on in the config_mixin code base.

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -0,0 +1,85 @@
+# encoding: utf-8
+require 'singleton'
+require "rubygems/package"
+
+module LogStash
+  class Registry
+
+    ##
+    # Placeholder class for registred plugins
```
ph: registered

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -0,0 +1,85 @@
...
+      else
+        # The plugin was defined directly in the code, so there is no need to use the
+        # require way of loading classes
+        return registry[plugin.path] if registered?(plugin.path)
+        raise LoadError
+      end
+    rescue => e
+      @logger.debug("Problems loading a plugin with", :type => type, :name => plugin, :path => plugin.path, :error => e) if @logger.debug?
+      raise LoadError, "Problems loading the requested plugin named #{plugin_name} of type #{type}."
+    end
```
ph: Do we really need to `#lookup` method? There is one already defined in `LogStash::Plugin` can we move all the logic of handling the loading/requiring in the same place?

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      def logstash_plugin_spec?(spec)
+        spec.metadata && spec.metadata["logstash_plugin"] == "true"
+      end
+
+    end
+
+    include Singleton
+
+    attr_reader :registry
```
ph: I think this attribute should be private.

ph:
This PR add the registry, really well coded btw, but I think it will still rely on the current directory structure to work am I correct? In that case the directory structure would be something like

1. We have a file named "logstash/inputs/mqtt.rb"
2. we define a logstash config with the following format " input { mqtt }"
3. The code will require the file
4. the file will register the plugin to the registry?

Correct?

I am OK with that flow for a first version, what I would like is when we define a plugin in the config we load a generic files that define the plugin behavior or file.

**really high level thoughts here***
"lib/plugins.rb"

```ruby
PLUGIN_CLASS = LogStash::Input::MQTT
```

logstash-core/lib/logstash/plugin.rb
```diff
@@ -68,47 +69,23 @@ def debug_info
...
-    namespace = LogStash.const_get(type_const)
-    # the namespace can contain constants which are not for plugins classes (do not respond to :config_name)
-    # namespace.constants is the shallow collection of all constants symbols in namespace
-    # note that below namespace.const_get(c) should never result in a NameError since c is from the constants collection
-    klass_sym = namespace.constants.find { |c| is_a_plugin?(namespace.const_get(c), name) }
-    klass = klass_sym && namespace.const_get(klass_sym)
-    raise(NameError) unless klass
-    klass
+  public
+  def self.annotate(type, name)
```
andrewvc: I think we should rename this. Maybe something like `register_plugin`.

logstash-core/lib/logstash/plugin.rb
```diff
@@ -68,47 +69,23 @@ def debug_info
...
+    path = "logstash/#{type}s/#{name}"
+    registry = LogStash::Registry.instance
```
andrewvc: Why the assignment here? It's only used once on the next line.

andrewvc:
WRT `require`, I think we should take this opportunity to make it more generic. IMHO if you install a plugin we should do the following.

1. Validate the plugin's name is in the format 'logstash-namespace-name'
2. Generate the gemfile line `gem 'logstash-output-foo', :require => 'logstash/outputs/foo'.

If we do this we can move the 'require' bit to bundler, and let our code focus on the registration. Additionally, we'll need this to allow multiple plugins to live under the same gem. 

Thoughts? @purbon @ph? I know this expands the scope a bit, but we only really get one shot to do this refactor correctly.

andrewvc:
BTW, code looks great now! Just want to make sure we iron out all the use cases!

ph:
> WRT require, I think we should take this opportunity to make it more generic. IMHO if you install a plugin we should do the following. 

Agree!

> Validate the plugin's name is in the format 'logstash-namespace-name''
Good idea

> Generate the gemfile line `gem 'logstash-output-foo', :require => 'logstash/outputs/foo'.

I was thinking about using the same behavior as the rubygems plugins. Also some plugins might need hooks in the system, copy a file on `post_install`, etc...

Rubygems' doc extract.

> As of RubyGems 1.3.2, RubyGems will load plugins installed in gems or $LOAD_PATH. Plugins must be named ‘rubygems_plugin’ (.rb, .so, etc) and placed at the root of your gem’s #require_path. Plugins are discovered via Gem::find_files then loaded. Take care when implementing a plugin as your plugin file may be loaded multiple times if multiple versions of your gem are installed.
> http://guides.rubygems.org/plugins/


> Thoughts? @purbon @ph? I know this expands the scope a bit, but we only really get one shot to do this refactor correctly.

Agree.

logstash-core/lib/logstash/plugin.rb
purbon: to reduce line length, I know variable is not useful after the call.

logstash-core/lib/logstash/plugins/registry.rb
purbon: I don't see all our "plugins" inheriting from ```LogStash::Plugin```, are we? this is why I created the registry as a separate item. I see mostly this being used for loading all kind of code within logstash. What do you think?

logstash-core/lib/logstash/plugins/registry.rb
purbon: I don't think is even necessary to expose it anymore, good spot.

purbon:
@ph 

> This PR add the registry, really well coded btw, but I think it will still rely on the current directory structure to work am I correct? In that case the directory structure would be something like
> 
> 1. We have a file named "logstash/inputs/mqtt.rb"
> 2. we define a logstash config with the following format " input { mqtt }"
> 3. The code will require the file
> 4. the file will register the plugin to the registry?
> 
> Correct?

Yes this is 100% correct and accurate. For a first version I wanted to "transport" much of the current behaviour to a generic instance. Then if next iterations of this idea we could innovate more versatile approaches.

> I am OK with that flow for a first version, what I would like is when we define a plugin in the config we load a generic files that define the plugin behavior or file.
> 
> **really high level thoughts here***
> "lib/plugins.rb"
> 
> ```ruby
> PLUGIN_CLASS = LogStash::Input::MQTT
> ```

I know your thoughts are really high level here, but I like the idea of having plugins define a more clever way to enable registration than standardizing a giving path to enable the ```require```. I will open a design issue to discuss next steps on plugin management, we can have all discussion there and follow up with implementation after it.

purbon:
@ph I just moved your thoughts at #4612 to discuss.

purbon:
@andrewvc let's have this discussion at #4612 and agree on next step. As your proposal really change the fact we work with plugins, I would say we should do it with another set of changes. This does not mean I don't like the idea, but would prefer to take baby steps :+1:

logstash-core/lib/logstash/plugin.rb
purbon: not sure I like this name, it gets confusing with the register method. what are we registering then .... -_-!

purbon:
@andrewvc @ph @suyograo looking forward to revamp this, what is missing in your opinion?

logstash-core/lib/logstash/plugin.rb
ph: +1 got `register_plugin`

I would argue that `register` was a bad name to what its actually do and it should have been initialize, the historical reason was to not to impose burden on the plugin developer to call super.

More arguments in favor of `register_plugin`, If you look at the definition of **annotate**
```
add notes to (a text or diagram) giving explanation or comment: documentation should be annotated with explanatory notes | (as adj. annotated) : an annotated bibliography.
```

vs
**Register**
```
enter or record on an official list or directory:
```

ph:
@purbon  I have reread this PR and updated #4612

purbon:
rebased this PR, @andrewvc also looking forward for your thoughts on this.

purbon:
@ph will take a look at #4612 and follow up with discussions there.

logstash-core/lib/logstash/codecs/base.rb
```diff
@@ -7,8 +7,13 @@
 # This is the base class for logstash codecs.
 module LogStash::Codecs; class Base < LogStash::Plugin
   include LogStash::Config::Mixin
+
   config_name "codec"
 
+  def self.plugin_type
+    "codec"
```
andrewvc: This should raise a `NotImplementedError` instead of having a default.

andrewvc: Oh, my mistake! Thought this was a different file!

logstash-core/lib/logstash/plugin.rb
```diff
@@ -117,23 +118,21 @@ def config_name
...
-
-    # check again if plugin is now defined in namespace after the require
-    namespace_lookup(type, name)
   rescue LoadError, NameError => e
+    logger.debug("Problems loading the plugin with", :type => type, :name => name, :path => path)
     raise(LogStash::PluginLoadingError, I18n.t("logstash.pipeline.plugin-loading-error", :type => type, :name => name, :path => path, :error => e.to_s))
   end
 
+  public
+  def self.annotate(type, name)
```
andrewvc: Have you considered naming this something more meaningful? I recall before the rebase that there was a discussion about naming this something else, like maybe `declare_plugin!` or the like.

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -0,0 +1,83 @@
...
+      def initialize(type, name)
+        @type  = type
+        @name  = name
+      end
+
+      def path
+        "logstash/#{type}s/#{name}"
+      end
+
+      def gem_name
```
andrewvc: I think one of the goals of this change is to make gem names orthogonal to what plugin gets registered. I think we could make this more clear with a method name like `guessed_gem_name`. One of my goals is to be able to combine plugins like logstash-input-rabbitmq and logstash-output-rabbitmq into one gem.

logstash-core/lib/logstash/plugins/registry.rb
andrewvc: I agree in that a plugin just has to be a class of any type. `LogStash::Plugin` is legacy IMHO, and specific to input/output/codec/filter plugins. Different plugin types may interface alternate interfaces. (Like the new agent plugin type).

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -0,0 +1,83 @@
...
+    def initialize
+      @registry = {}
+      @logger = Cabin::Channel.get(LogStash)
+    end
+
+    def lookup(type, plugin_name, &block)
+
+      plugin = Plugin.new(type, plugin_name)
+
+      if plugin.installed?
```
andrewvc: Correct me if I'm wrong, but this still requires plugins to have gems named `logstash-#{type}-#{name}` yes? I don't think we should look for the spec, but rather just if its in the registry.

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -0,0 +1,83 @@
...
+require 'singleton'
+require "rubygems/package"
+
+module LogStash
+  class Registry
+
+    ##
+    # Placeholder class for registered plugins
+    ##
+    class Plugin
```
andrewvc: Using a class here seems heavy. Is there a way we can do this just by adding these as class methods to `LogStash::Registry`?

logstash-core/lib/logstash/plugin.rb
ph: I think you were suggesting for `register_plugin` which was more explicit, I was +1 on that change.

logstash-core/lib/logstash/plugins/registry.rb
ph: +1

logstash-core/lib/logstash/plugin.rb
purbon: yeah, we should rename this, not sure why it got back here (rebase mistake I guess).

logstash-core/lib/logstash/plugins/registry.rb
purbon: I don't think so, I like the separation of concerns here as it makes the different code path less polluted. If properly organized it could also benefit other parts of the stack with easier testing.

logstash-core/lib/logstash/plugins/registry.rb
purbon: The only point this is used is to check if the plugin, coming from a non random gem name, is installed or not. 

If installed, it add the plugin to the registry himself, basically to support the old style plugins. When having more than 1 plugin in the same gem you will need somehow to register all plugins, because without any pattern there is no way to get this automatically done.

will try to simplify the logic a bit, not sure all steps are 100% necessary.

logstash-core/lib/logstash/plugins/registry.rb
purbon: The logic here try to fit for backwards compatible plugins, we _don't want_ to require to update all plugins after we add this change. 

If the plugin is not detected to be installed it will fallback to only check if it is in the registry, or raise an error.

logstash-core/lib/logstash/plugins/registry.rb
ph: I agree with you @purbon, for this.

logstash-core/lib/logstash/plugins/registry.rb
andrewvc: You're right! I withdraw this comment!

andrewvc:
LGTM!

ph:
LGTM

purbon:
yay, thanks for your thoughtful reviews.

purbon:
please jenkins, test this.

purbon:
please jenkins, test this

PR #4538, opened 2016-01-21
search through nested directory structure for jars in vendor directory

talevy:
It is not always the case that jars will be located in one level within the `runtime-jars` and `test-jars` directory. For example, when a project's jar dependencies is vendored using [jar-dependencies](https://github.com/mkristian/jar-dependencies), the jars are nested.

### DIFF
--- a/logstash-core/lib/logstash/environment.rb
+++ b/logstash-core/lib/logstash/environment.rb
@@ -33,11 +33,11 @@ module LogStash
     end
 
     def load_runtime_jars!(dir_name="vendor", package="jar-dependencies")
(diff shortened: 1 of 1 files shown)


purbon:
@talevy small question, might be out of having small knowledge of jar-dependencies? Aren't jar-dependencies also pulling the jars in ? just curious about it :smile:

purbon:
btw, test are red here ... do you know why?

suyograo:
@talevy @purbon bump. can we either merge or close this?

talevy:
Hmm. I do not have a necessary use case of this anymore since we refactored ways we do things a bit here. I will close this for now. If it ever comes up again, I'll re-open.

PR #4547, opened 2016-01-21
Fix hanging runner test

### LINKED ISSUE #4546: runner_spec hangs when executing
see 

```
skywalker% bin/rspec --format documentation logstash-core/spec/logstash/runner_spec.rb
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}

LogStash::Runner
  pipeline settings
    when pipeline workers is defined by the user
```

this might be probably due to the way the webserver starts.

purbon:
by actually mocking out the agent execute call, so test execution are only testing the runner and not the following methods.

Fixes #4546

### DIFF
--- a/logstash-core/spec/logstash/runner_spec.rb
+++ b/logstash-core/spec/logstash/runner_spec.rb
@@ -105,6 +105,7 @@ describe LogStash::Runner do
     let(:pipeline) { double("pipeline") }
 
     before(:each) do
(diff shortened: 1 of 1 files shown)


ph:
LGTM

PR #4554, opened 2016-01-22
Fixed small bugs that pervented the metrics collection to work as expected

### LINKED ISSUE #4553: Missing method when the metric store gets flushed
Error when running metrics branch,

```
Collector: Something went wrong went sending data to the observers {:execution_time=>2016-01-22 10:34:51 +0100, :result=>nil, :exception=>#<NoMethodError: undefined method `all' for #<LogStash::Instrument::MetricStore:0x5a711356>>, :level=>:error}
```

this happen when the metric store gets flushed.

purbon:
This is a collection of small fixes that make the metric collection work again, it also introduces a test to make sure the metrics input can fetch data from the metric store when receiving an update call. This bug was introduced when changing the behaviour of  ```MetricStore.each method```, used during the previous procedure.

It also fixes an important issue that made the api service gathering code to load after a resource was loaded, now the service provider is loaded as soon as the webapi is started.

Fixed #4553 and #4555

@ph thoughts

### DIFF
--- a/logstash-core/api/init.ru
+++ b/logstash-core/api/init.ru
@@ -9,6 +9,8 @@ require 'app/stats'
 env = ENV["RACK_ENV"].to_sym
 set :environment, env
 
+set :service, LogStash::Api::Service.instance
+
 run LogStash::Api::Root
 
 namespaces = { "/_stats" => LogStash::Api::Stats }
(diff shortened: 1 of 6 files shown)


ph:
LGTM

PR #4561, opened 2016-01-25
add specs for nil values in array fields

colinsurprenant:
specs for nil values in `Event` array fields.
this relates to discussion in Java Event implementation #4494


ph:
LGTM, I think it cover the use case we discussed.

But, did we ever tested what happen when we have multiple level of arrays? 

```ruby
subject["[foo]"] = [[12, nil], 56]
expect(subject.sprintf("%{[foo]}")).to eq(??)
``

colinsurprenant:
Per the logic it should be `"12,,56"`, will add a test case for this too.

colinsurprenant:
I'll go ahead and merge that in master and 2.x

ph:
lgtm

PR #4584, opened 2016-01-27
Extended installation of packaged plugins

### LINKED ISSUE #4585: Plugin Manager: Extend functionality to installation plugin packs
Plugin packs are a collection of multiple plugins, which contain gems and their dependencies. This is mainly an upper layer on top of current install command that will basically pull all plugins and install them from the package. Also compliments the current offline installation features.

Some characteristics about packages are:

* They could be installed either from a web url or from the local file system.
* Are compressed in a zip file and have the zip extension.
* They can have one or multiple plugins, including their dependencies.

Tasks:

* [x] Enhance the current install command to handle packages.
* [x] Handle plugins from a web url.
* [x] Handle plugins from the file system.
* [x] Handle plugins by name.
* [x] Introduce version checks to be sure plugins works with expected logstash versions.
* [x] Add user agent identification when fetching plugins from a web url.
* [x] Add all necessary unit

Pending/Deferred to the near feature:

* [ ] Add all necessary integration tests.

The addition of integration test is deferred to the creation of a reasonably automated infrastructure to run them automatically.

purbon:
This PR adds the idea of packaged plugins to the plugin manager, basically a collection of plugins that are going to be installed from a given zip file. This file could life either in a local file system or in a remote installation.

This set of changes also need to make available the ```LOGSTASH_VERSION``` constant at the plugin manager level, this requires to finally move the ```logstash/version``` file from logstash core where it was before, updating all dependencies, etc.  Now logstash-core only have it's own version, and the main logstash version is under ```lib```.

Fixes #4585


purbon:
@ph do you want to start giving some thoughts on this PR? thanks a lot. Happy to provide you with as much context as you might need.

lib/pluginmanager/commands/install_command.rb
```diff
@@ -0,0 +1,130 @@
...
+
+  def is_zip_file?(plugin)
+    ::File.extname("#{plugin}") == ".zip"
+  end
+
+  # Build a package URI from a given package param
+  # @param [String] A package name
+  # @return [URI] A URI for the given package
+  def build_uri(plugin)
+    return URI("http://127.0.0.1:8080/#{plugin}.zip") unless is_zip_file?(plugin)
```
electrical: Why is this static to this IP and port?

purbon: This need to be change and make configurable.

lib/pluginmanager/sources/http.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  class HTTP < Base
+
+    attr_reader :uri, :fallback
+
+    def initialize(uri)
+      super
+      @fallback = !uri.start_with?("http")
+      if fallback
+        @uri = URI("http://localhost:8080/#{uri}.zip")
```
purbon: This address should be made configuratble.

ph:
I need a bit more time, since of the complexity of this PR.

purbon:
@ph looking forward to your thoughts on here.

purbon:
@ph your thoughts on here are very welcome.

purbon:
Added basic integration test in 7354974846580c589852e1cd5bafda7c43575862, however they have one small issue and is that they need for now us to create a new "good" artifact to be installed every time we release a new version. I'm sure we can fix that as soon as we've regular releases of packs.

purbon:
@ph rebased

lib/pluginmanager/install.rb
```diff
@@ -32,6 +52,12 @@ def execute
   end
 
   private
+
+  def merge_into_plugin_list(packs)
+    (plugins_arg << packs).flatten!
```
ph: You can use `Array#concat`

spec/spec_helper.rb
```diff
@@ -1,4 +1,5 @@
 # encoding: utf-8
+require "clamp"
```
ph: I dont think this library is needed for all the tests, I think this should only required on the corresponding file.

lib/pluginmanager/sources/http.rb
```diff
@@ -0,0 +1,78 @@
...
+      path   = File.join(dest, uri.path.split("/").last)
+      status = nil
+      Net::HTTP.start(*http_start_args(uri)) do |http|
+        response = http.get(uri.request_uri, header_options)
+        status   = response.code
+        break unless status == "200"
+        write_to_file(path, response.body)
+      end
+      [path, status]
+    end
```
ph: It would be nice if this would use stream download instead of keeping the whole package into memory.

ph: If something went wrong during downloading should we delete the file?
What happen if the file already exist, the content get replaced?

lib/pluginmanager/commands/install_command.rb
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  def find_packs(args=[])
+    packs = []
+    args.clone.each do |arg|
+      next if (arg.start_with?("logstash-") || File.extname(arg) == ".gem")
+      source = build_source(arg)
+      # We should not remove packages called by name that does not exist, they
+      # could still be plugins to be installed from rubygems
+      if !( source.is_a?(LogStash::PluginManager::Sources::HTTP) &&
```
ph: remove space after parenthese

lib/pluginmanager/commands/install_command.rb
```diff
@@ -0,0 +1,108 @@
...
+  # Build a package URI from a given package param
+  # @param [String] A package name
+  # @return [URI] A URI for the given package
+  def build_source(arg)
+    zip_file = ::File.extname("#{arg}") == ".zip"
+    if !zip_file || ( zip_file && arg.start_with?("http") )
+      return LogStash::PluginManager::Sources::HTTP.new(arg)
+    end
+    LogStash::PluginManager::Sources::Local.new(arg)
+  end
```
ph: I think this logic should a factory method `LogStash::PluginManager::Sources`, This would make it more, like give me the correct interface for this path.

lib/pluginmanager/sources/http.rb
```diff
@@ -0,0 +1,78 @@
...
+    def initialize(uri)
+      super
+      @fallback = !uri.start_with?("http")
+      if fallback
+        @uri = URI(fallback_uri(uri))
+      else
+        @uri = URI(uri)
+      end
+      @header_options = {'User-Agent' => USER_AGENT}
+    end
```
ph: After looking at the implementation of https://github.com/elastic/logstash/pull/4584/files#diff-a216a3887000aed590754e380e388739R96

I don't think the `fallback` will be called.

lib/pluginmanager/sources/http.rb
```diff
@@ -0,0 +1,78 @@
...
+    # by name, if you require to install a pack from a custom URL
+    # you will need to pass the full URL.
+    ##
+    ROOT_URI = "https://download.elastic.co".freeze
+
+    ##
+    # User agent string used to identify the requester is a
+    # LogStash agent
+    ##
+    USER_AGENT = "Logstash/#{LOGSTASH_VERSION}".freeze
```
ph: good idea for stats.

lib/pluginmanager/sources/http.rb
```diff
@@ -0,0 +1,78 @@
...
+      response = Net::HTTP.start(*http_start_args(uri)) do |http|
+        http.head(uri.request_uri, header_options)
+      end
+      response.kind_of?(Net::HTTPSuccess)
+    rescue
+      false
+    end
+
+    def fetch(dest="")
+      path   = File.join(dest, uri.path.split("/").last)
```
ph: `File.basename` instead of split?

lib/pluginmanager/sources/local.rb
```diff
@@ -0,0 +1,26 @@
...
+    def exist?
+      ::File.exist?("#{uri}")
+    end
+
+    def fetch(dest="")
+      base_dir = (dest.empty? ? LogStash::Environment::LOGSTASH_HOME : dest )
+      current_path = ::File.join(base_dir, ::File.basename("#{uri}"))
+      ::FileUtils.cp("#{uri}", current_path)
+      [ current_path, "200" ]
+    end
```
ph: "200" is an implementation details which doesn't have any relation with the local file system.
Would it better to return an object with a success method?

lib/pluginmanager/sources/local.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  class Local < Base
+
+    def exist?
+      ::File.exist?("#{uri}")
+    end
+
+    def fetch(dest="")
+      base_dir = (dest.empty? ? LogStash::Environment::LOGSTASH_HOME : dest )
+      current_path = ::File.join(base_dir, ::File.basename("#{uri}"))
```
ph: instead of using string interpolation for `File.basename` I suggest we use either `URI#path` or `URI#to_s`

ph: But in reality this should not be needed since `uri` should be a string in the context of local source, no?

ph: Oh well, the base class force URI() on everything.

```
u = URI("stove.io/download.zip")
=> #<URI::Generic:0x28b576a9 URL:stove.io/download.zip>
```
And  it fallback to a generic scheme, interesting.

lib/pluginmanager/sources/http.rb
```diff
@@ -0,0 +1,78 @@
...
+    USER_AGENT = "Logstash/#{LOGSTASH_VERSION}".freeze
+
+    def initialize(uri)
+      super
+      @fallback = !uri.start_with?("http")
+      if fallback
+        @uri = URI(fallback_uri(uri))
+      else
+        @uri = URI(uri)
+      end
```
ph: OK, this part of the initialize was confusing me, the parent class set the `@uri` properties when we pass the `value` to the method call. So if this is the case we could rewrite this method like this.

```ruby
def initialize(uri)
  @fallback = !uri.start_with?("http")

  if fallback
    @uri = super(fallback_uri(uri))
  else
    @uri = super(@uri)
  end
  @header_options = {'User-Agent' => USER_AGENT}
end
```

lib/pluginmanager/sources/local.rb
```diff
@@ -0,0 +1,26 @@
...
+require "rubygems/package"
+require "pluginmanager/sources/base"
+require "fileutils"
+
+module LogStash::PluginManager::Sources
+
+  class Local < Base
+
+    def exist?
+      ::File.exist?("#{uri}")
```
ph: uri.path

lib/pluginmanager/sources/base.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def valid?
+      return false if version.nil? || version.empty?
+      is_zip_file? && valid_version?
+    end
+
+    def valid_version?
+      LOGSTASH_VERSION == version
+    end
```
ph: I find it a bit fragile to check for compatibility by using a version in the name of the package, is all the other xpack implementation doing the same thing?

lib/pluginmanager/sources/base.rb
```diff
@@ -0,0 +1,47 @@
...
+      raise "NotImplemented"
+    end
+
+    def version
+      File.basename("#{uri}", ".zip").split("-")[1]
+    end
+
+    def to_s
+      "#{uri}"
+    end
```
ph: maybe uri.to_s

lib/pluginmanager/commands/install_command.rb
```diff
@@ -0,0 +1,108 @@
...
+      (fields.count > 1 ? fields[0...-1] : fields).join("-")
+    end
+  end
+
+  # Fetch a package from a URI, either from a web uri or from the local file system.
+  # @param [URI] The package url
+  # @param [String] A temporary directy used to store the fetched package.
+  # @return [String] The location of the fetched package.
+  def fetch_pack(source, temp_dir)
+    file, _ =  source.fetch(temp_dir)
```
ph: Why do we return the status code if we don't do anything with it?

lib/pluginmanager/commands/install_command.rb
```diff
@@ -0,0 +1,108 @@
...
+    filename = File.basename(file, ".zip").split("-")[0]
+    output_dir = File.join(File.dirname(file), "out")
+    zip.extract(file, output_dir)
+    return File.join(output_dir, "logstash", filename)
+  end
+
+  def find_packs(args=[])
+    packs = []
+    args.clone.each do |arg|
+      next if (arg.start_with?("logstash-") || File.extname(arg) == ".gem")
```
ph: Can we move theses conditionals to predicates?

lib/pluginmanager/commands/install_command.rb
```diff
@@ -0,0 +1,108 @@
...
+    packs
+  end
+
+  # Finds and makes sure if there are packages willing to be installed they are
+  # available inside the system.
+  # @return [Array] The list of plugins to be installed from within the package.
+  def fetch_and_copy_packs(args=[], &block)
+    FileUtils.mkdir_p(LogStash::Environment::CACHE_PATH) unless args.empty?
+    plugins = []
+    Dir.mktmpdir("logstash-plugin-manager") do |temp_dir|
```
ph: We should use `Stud` for this https://github.com/jordansissel/ruby-stud/blob/master/lib/stud/temporary.rb#L45

lib/pluginmanager/commands/install_command.rb
```diff
@@ -0,0 +1,108 @@
+# encoding: utf-8
+require "pluginmanager/command"
+require "bootstrap/util/compress"
+require "fileutils"
+require "pluginmanager/sources/http"
+require "pluginmanager/sources/local"
+
+class LogStash::PluginManager::InstallCommand < LogStash::PluginManager::Command
```
ph: Why this code is not in the `commands/install` file?

lib/pluginmanager/sources/http.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    def exist?
+      response = Net::HTTP.start(*http_start_args(uri)) do |http|
+        http.head(uri.request_uri, header_options)
+      end
+      response.kind_of?(Net::HTTPSuccess)
+    rescue
+      false
+    end
```
ph: good usage of head here.

spec/pluginmanager/commands/install_spec.rb
```diff
@@ -0,0 +1,68 @@
+# encoding: utf-8
+require "spec_helper"
+require "pluginmanager/commands/install_command"
+require "uri"
+require 'webmock/rspec'
+
+describe LogStash::PluginManager::InstallCommand do
```
ph: Should the versions referenced in theses files uses the `LOGSTASH_VERSION`? They still mention 3.0.0-dev and master should be 5.0.0?

spec/pluginmanager/commands/install_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+
+    it "select plugins to be installed" do
+      files = [ "/foo/bar/plugin.gem", "/foo/bar/bar-2.0.gem" ]
+      expect(Dir).to receive(:glob).with(File.join("/foo/bar", "*.gem")).and_return(files)
+      expect(subject.select_plugins("/foo/bar")).to eq(["plugin", "bar"])
+    end
+
+    context "#find packs" do
+      let(:syslog_pack_url) { "http://download.elastic.co:443/syslog_pack-3.0.0.dev.zip" }
```
ph: I am confused here, :443 and no ssl?

spec/pluginmanager/sources_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+
+  describe "#fetch" do
+
+    before(:each) do
+      File.delete("pack.zip") if File.exist?("pack.zip")
+    end
+
+    after(:each) do
+      File.delete("pack.zip") if File.exist?("pack.zip")
+    end
```
ph: Do we need both callbacks?

spec/pluginmanager/sources_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    it "should start the download process" do
+      stub_request(:get, uri).and_return(:body => "foo", :status => 404)
+      Dir.mktmpdir do |tmp_dir|
```
ph: Lets use `Stud` for this.

spec/pluginmanager/sources_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+      it "fails the validation" do
+        expect(subject.valid?).to be_falsey
+      end
+    end
+
+  end
+
+  describe "#fetch" do
+
+    let(:destination) { "/destionation/path" }
```
ph: type in destination

spec/pluginmanager/sources_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    it "copy a file from a given path to a destination" do
+      destination_path = File.join(destination, "bar-#{LOGSTASH_VERSION}.zip")
+      expect(::FileUtils).to receive(:cp).with(uri, destination_path).and_return(true)
+      subject.fetch(destination)
+    end
```
ph: I am wondering if we could check if the file was created instead of doing expectation on `FileUtils?`

spec/pluginmanager/sources_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    it "uses LOSTASH_HOME if no destination is given" do
+      destination_path = File.join(LogStash::Environment::LOGSTASH_HOME, "bar-#{LOGSTASH_VERSION}.zip")
+      expect(::FileUtils).to receive(:cp).with(uri, destination_path).and_return(true)
+      subject.fetch
```
ph: Same comment as above.

integration/plugin_manager/plugin_install_spec.rb
```diff
@@ -1,41 +1,90 @@
...
-      result = command("bin/logstash-plugin install #{plugin_name}")
+    let(:pack)        { "https://s3.amazonaws.com/test.elasticsearch.org/logstash/jdbc_pack-#{LOGSTASH_VERSION}.zip" }
+    let(:plugin_name) { "logstash-input-jdbc" }
+
+    it "install sucessfully" do
+      result = command("bin/logstash-plugin install #{pack}")
       expect(result.exit_status).to eq(0)
-      expect(result.stdout).to match(/^Validating\s#{plugin_name}\nInstalling\s#{plugin_name}\nInstallation\ssuccessful$/)
+      expect(result.stdout).to match(/^Validating\s#{pack}\nValidating\s#{plugin_name}\nInstalling\s#{plugin_name}\nInstallation\ssuccessful$/)
     end
```
ph: Should we have a test to check if the version of plugin in the pack is correctly installed `bin\plugin list --verbose PLUGIN`

ph:
I did a pass on the code. I think this solution will work but I haven't tested all the use cases.

spec/pluginmanager/sources_spec.rb
purbon: why? this does the work, isn't? what is the benefit of pushing stud for this?

ph: Because its a library we maintain and we use it everywhere when dealing with temporary files and directory. I think consistency around this need is good.

spec/pluginmanager/sources_spec.rb
purbon: we might not need both, I guess I did it to be extra sure the file was clean up :-P

spec/pluginmanager/sources_spec.rb
purbon: sure, but still no technical benefit for this use case here, I must
disagree of using for using libs :-)

lib/pluginmanager/sources/base.rb
purbon: yes, a pack could only be installed for the version I was created. For now the only way we've to fetch this is thought the URL and filename, very fragile ... but we've not much more available.

lib/pluginmanager/sources/local.rb
purbon: yes, ^^ is the reason why I used to_s to convert it ... you should use ``file://` to be detected as local uri, otherwise anything without schema is generic.

lib/pluginmanager/commands/install_command.rb
purbon: because I tried to extract some logic away from the `pluginmanager/install.rb`  in a way that things like utility, fetch, verification functions are more wrap here ... this works on a similar way as `pack_command.rb` or the `command.rb` interface in the main directory.

spec/pluginmanager/commands/install_spec.rb
purbon: version here does not mather, is just to write the test and validate them, nothing else. I would rather not update it or interpolate any LOGSTASH_VERSION, but this is a matter of taste to me.

spec/pluginmanager/commands/install_spec.rb
purbon: this is the expected url resolution by `webmock/rspec`

spec/pluginmanager/sources_spec.rb
ph: Every time, I do not use Stud in one of my PR for this kind of behavior, I get asked to use it. So I pass this information around. And is not like I asked you to include a new dependency to the gemfile. Its already in the Logstash project and we maintain it.

For the technical point, this library allow user to define the temporary directory in multiple way  that the default library don't. As a user I would be confused if not all the temporary files where saved at the same place.

https://github.com/jordansissel/ruby-stud/blob/master/lib/stud/temporary.rb#L13

integration/plugin_manager/plugin_install_spec.rb
purbon: makes sense to me, however end of the day this is also testing this, without the `Installation\ssuccessful` here we would rather have nothing installed, that is what we aim to check here.

If I understand you properly you want to add another check to be sure if `pack (a)`, include `plugin (p), version 1`, the version installed is version 1, isn't? wouldn't this be something check during pack build time that the version is this one? packs don't to more internal validation than just pushing the plugins packaged with them.

lib/pluginmanager/commands/install_command.rb
purbon: at this point we are just sorting strings here, so having predicates is not easy unless we transform in earlier steps this strings to more complex objects.

integration/plugin_manager/plugin_install_spec.rb
ph: Yes I would prefer to have this kind of check, We have seen case that the plugin manager was returning the `succesful` when it did not actually succeed. Its probably a bit harder to write but should make it more robust IMHO.

lib/pluginmanager/commands/install_command.rb
ph: I did not imply complex object, I would have been fine with the following.

```ruby
def logstash_pack?(file)
   file.start_with?("logstash-") 
end
```

Event is we only do assertion on the string itself using these kind of method hide the complexity away.

lib/pluginmanager/sources/base.rb
ph: Yeah its a sad reality  :(

On a related note, the only way to enforce a plugin from a pack to work only on specific version of Logstash would be to make it a strict dependency on version of the core.

OK lets move on.

spec/pluginmanager/commands/install_spec.rb
ph: OK since the call is wrapped in webmock, but I still find it strange.

suyograo:
@ph @purbon do we need another pass on this?

ph:
@suyograo I still have a few question in the PR who were not answered and we will need a second reviewer for it.

integration/plugin_manager/plugin_install_spec.rb
ph: Since we have merged the qa/acceptance test, can we migrate the test there?

ph:
@purbon I took a quick look at the state of this before we sync.

Lets try to discuss theses issues:


- Move test to the QA/Acceptance framework.
- Can we add metadata to the pack itself? I find it fragile to rely on the name when we do validation to check if we can install it.
   - Either to have a manifest file in the archive to check against.
   - Or in the file header if possible?
-  What happen to the file if the download fails.
- Http Error Code handling: https://github.com/elastic/logstash/pull/4584/files#diff-a216a3887000aed590754e380e388739R40

Issue #4492, opened 2016-01-14
Internal structure of the metric of the metric store


ph:
When merging the JVM and the web api we found out that the current way of defining namespace and keys wasn't flexible enough.

Lets try to define all all the possibilities we can have.

# Top level
| metrics | type | phase | Status |
----------- | ------ | -------- | --------- |
stats.events.in | counter | 1 | Done in the PR, need to change namespace
stats.events.out | counter | 1 | Done in the PR, need to change namespace
stats.uptime_in_milliseconds | gauge value | 1
stats.logtash_version | gauge value | 1

| metrics | type | phase |
----------- | ------ | -------- |
stats.jvm.memory.heap.used_in_bytes | gauge value | 1
stats.jvm.memory.used_in_bytes | gauge | 1 
stats.jvm.memory.used_percent | gauge value | 1
stats.jvm.memory.committed_in_bytes | gauge | 1
stats.jvm.memory.max_in_bytes | gauge | 1

| metrics | type | phase |
----------- | ------ | -------- |
stats.os.cpu.XXX | X | X

# Detailed pipelines level, collected at the pipeline level
In the following case **XXX** is a **UUID** of user defined string in the logstash configuration.

| metrics | type | phase | status |
----------- | ------ | -------- | -------- |
stats.pipelines.pipeline_X.events.in | counter | 1  | Done in the PR, need to change namespace
stats.pipelines.pipeline_X.events.out | counter | 1 | Done in the PR, need to change namespace
stats.pipelines.pipeline_X.plugins.inputs.file_XXXXXX.events.out | counter |  1 | Done in the PR, need to change namespace
stats.pipelines.pipeline_X.plugins.filters.grok_XXXXXX.events.out | counter |  1 | Done in the PR, need to change namespace
stats.pipelines.pipeline_X.plugins.filters.grok_XXXXXX.events.in | counter |  1 | Done in the PR, need to change namespace
stats.pipelines.pipeline_X.plugins.outputs.elasticsearch_XXXX.events.in | counter |  1 | Done in the PR, need to change namespace
stats.pipelines.pipeline_X.plugins.outputs.elasticsearch_XXXX.events.out | counter |  need to be done at the plugin level or with a ack.

# Plugins Level metrics
Theses statistic are collected by the code inside the plugin and are defined by the plugin authors.

| metrics | type | phase |
----------- | ------ | -------- |
[...]

Issue #4506, opened 2016-01-15
Add guards for the `MetricStore`


ph:
The current implementation of the `MetricStore` can grow unbounded,  we should add guards on the size of it to make sure its stays under control and log that information.

ph:
Lets push it to make it work for beta1

ph:
updated with another guard

ph:
> I understand, as I shared earlier I think I would add some kind of defense/validation to prevent people falling on creating unexpected namespaces, what do you think?

I am including your comment in this issues, do you have a clear scenario when it could be a problem and what would be the strategy to stop it from hapenning?

PR #4602, opened 2016-02-01
Metrics/flexibility

### LINKED ISSUE #4461: Internal: Filter metrics by namespaces
As a consumer of the metrics collector I aim to get meaningful namespaces that let me select and fetch data easy. For now we've:

* root.base.events_in
* root.base.events_filtered

I would propose to have something that matches more close, semantically speaking the resources defined for #4446, so we can make something like.

* root.stats.events, to get all events information
* root.stats.jvm, to get all jvm information
* root.stats.networking, to get all events information.

I also propose to rename root to a word with a bit more context, what do you think? might be something like metrics? or logstash? then we could have namespaces like:

* logstash.stats.events
* logstash.stats.jvm 

what do you think?

### LINKED ISSUE #4529: Hide the implementations details of the MetricStore when retrieving metrics
Hi,
  as a user of the metric store I don't expect to interact with implementation details, today when asking the metric store for data it returns an instance of a ```Concurrent::Map```, this makes the consumer of this api aware of an implementation detail that might change in the feature.

We should return a plain ruby hash, or similar.

The way the Metric store is implemented also make the values of this ```Concurrent::Map``` be a MetricType instance. This is problematic as to retrieve a value out of it a user must call it's value method. The returned value object for the metric store should be a plain ruby object or similar.

ph:
Refactoring of the metric:

- Conform to the metric defined in Follow #4492
- Better filtering of the metrics from the metric store Fixes #4461
- Hide the implementation details of the metric store It return an hash, Fixes #4529
- MetricStore#each now support a path to only fetch a subset on the metrics.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -22,7 +22,7 @@ class LogStash::Agent
 
   def initialize(options = {})
     @pipelines = {}
-     
+
     @node_name = options[:node_name] || Socket.gethostname
     @collect_metric = options.fetch(:collect_metric, false)
     @logger = options[:logger]
@@ -59,7 +59,7 @@ class LogStash::Agent
 
   def add_pipeline(pipeline_id, config_str, settings = {})
     settings.merge!(:pipeline_id => pipeline_id,
-                    :metric => metric.namespace(pipeline_id))
+                    :metric => metric)
 
     @pipelines[pipeline_id] = LogStash::Pipeline.new(config_str, settings)
   end
@@ -69,7 +69,7 @@ class LogStash::Agent
   end
 
   # Calculate the Logstash uptime in milliseconds
-  # 
+  #
   # @return [Fixnum] Uptime in milliseconds
   def uptime
     ((Time.now.to_f - started_at.to_f) * 1000.0).to_i
@@ -92,7 +92,7 @@ class LogStash::Agent
   def start_background_services
     if collect_metric?
       @logger.debug("Agent: Starting metric periodic pollers")
-      @periodic_pollers.start 
+      @periodic_pollers.start
     end
   end
 
@@ -106,7 +106,7 @@ class LogStash::Agent
   def configure_metric
     if collect_metric?
       @logger.debug("Agent: Configuring metric collection")
-      @metric = LogStash::Instrument::Metric.create(:root)
+      @metric = LogStash::Instrument::Metric.create
       add_metric_pipeline
     else
       @metric = LogStash::Instrument::NullMetric.new
@@ -131,7 +131,7 @@ class LogStash::Agent
       }
       output {
         elasticsearch {
-          flush_size => 10
+          flush_size => 1
           hosts => "127.0.0.1"
           index => "metrics-%{+YYYY.MM.dd}"
         }
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -52,7 +52,10 @@ module LogStash
         # we also need to trace the number of events
         # coming from a specific filters.
         new_events = @filter.flush(options)
-        @metric_events.increment(:out, new_events.size) unless new_events.nil? # Logstash-filter-aggregates return nil.
+
+        # Filter plugins that does buffering or spooling of events like the
+        # `Logstash-filter-aggregates` can return `NIL` and will flush on the next flush ticks.
+        @metric_events.increment(:out, new_events.size) unless new_events.nil?
         new_events
       end
     end
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -23,6 +23,7 @@ module LogStash
       @metric = metric.namespace(@filter.id.to_sym)
       @filter.metric = @metric
 
+      # Not all the filters will do bufferings
       define_flush_method if @filter.respond_to?(:flush)
     end
 
@@ -31,18 +32,26 @@ module LogStash
     end
 
     def multi_filter(events)
-      @metric.increment(:events_in, events.size)
+      @metric.increment(:in, events.size)
 
       new_events = @filter.multi_filter(events)
 
-      @metric.increment(:events_out, new_events.size)
+      # There is no garantee in the context of filter
+      # that EVENTS_INT == EVENTS_OUT, see the aggregates and
+      # the split filter.
+      @metric.increment(:out, new_events.size)
+
       return new_events
     end
 
     private
     def define_flush_method
       define_singleton_method(:flush) do |options = {}|
-        @filter.flush(options)
+        # we also need to trace the number of events
+        # coming from a specific filters.
+        new_events = @filter.flush(options)
+        @metric.increment(:out, new_events.size) unless new_events.nil? # Logstash-filter-aggregates return nil.
+        new_events
       end
     end
   end
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -20,8 +20,10 @@ module LogStash
       @filter = klass.new(options)
 
       # Scope the metrics to the plugin
-      @metric = metric.namespace(@filter.id.to_sym)
-      @filter.metric = @metric
+      namespaced_metric = metric.namespace(@filter.id.to_sym)
+      @filter.metric = metric
+
+      @metric_events = namespaced_metric.namespace(:events)
 
       # Not all the filters will do bufferings
       define_flush_method if @filter.respond_to?(:flush)
@@ -32,14 +34,14 @@ module LogStash
     end
 
     def multi_filter(events)
-      @metric.increment(:in, events.size)
+      @metric_events.increment(:in, events.size)
 
       new_events = @filter.multi_filter(events)
 
       # There is no garantee in the context of filter
       # that EVENTS_INT == EVENTS_OUT, see the aggregates and
-      # the split filter.
-      @metric.increment(:out, new_events.size)
+      # the split filter
+      @metric_events.increment(:out, new_events.size) unless new_events.nil?
 
       return new_events
     end
@@ -50,7 +52,7 @@ module LogStash
         # we also need to trace the number of events
         # coming from a specific filters.
         new_events = @filter.flush(options)
-        @metric.increment(:out, new_events.size) unless new_events.nil? # Logstash-filter-aggregates return nil.
+        @metric_events.increment(:out, new_events.size) unless new_events.nil? # Logstash-filter-aggregates return nil.
         new_events
       end
     end
--- a/logstash-core/lib/logstash/inputs/metrics.rb
+++ b/logstash-core/lib/logstash/inputs/metrics.rb
@@ -11,6 +11,7 @@ module LogStash module Inputs
   # Elasticsearch
   class Metrics < LogStash::Inputs::Base
     config_name "metrics"
+    milestone 3
 
     def register
     end
@@ -35,19 +36,22 @@ module LogStash module Inputs
     def update(snapshot)
       @logger.debug("Metrics input: received a new snapshot", :created_at => snapshot.created_at, :snapshot => snapshot, :event => snapshot.metric_store.to_event) if @logger.debug?
 
-      # TODO: (ph)
-      # - Obviously the format here is wrong and we need to
-      # transform it from Snapshot to an event
-      # - There is another problem, if the queue is full this could block the snapshot thread.
-      # There is a few possible solution for this problem:
-      #   - We can use a future
-      #   - We can use a synchronization mechanism between the called thread (update method)
-      #   and the plugin thread (run method)
-      #   - How we handle back pressure here?
-      #   - one snashot should be only one event ?
+      # The back pressure is handled in the collector's
+      # scheduled task (running into his own thread) if something append to one of the listener it will
+      # will timeout. In a sane pipeline, with a low traffic of events it shouldn't be a problems.
       snapshot.metric_store.each do |metric|
-        @queue << LogStash::Event.new({ "@timestamp" => snapshot.created_at }.merge(metric.to_hash))
+        @queue << LogStash::Event.new({ "@timestamp" => snapshot.created_at }.merge(convert_to_hash(metric)))
       end
+    end
+
+    # Transform the MetricType into a hash to create a new event
+    def convert_to_hash(metric)
+      {
+        "namespaces" => metric.namespaces,
+        "key" => metric.key,
+        "type" => metric.type,
+        "value" => metric.value
+      }
     end
   end
 end;end
--- a/logstash-core/lib/logstash/inputs/metrics.rb
+++ b/logstash-core/lib/logstash/inputs/metrics.rb
@@ -40,18 +40,8 @@ module LogStash module Inputs
       # scheduled task (running into his own thread) if something append to one of the listener it will
       # will timeout. In a sane pipeline, with a low traffic of events it shouldn't be a problems.
       snapshot.metric_store.each do |metric|
-        @queue << LogStash::Event.new({ "@timestamp" => snapshot.created_at }.merge(convert_to_hash(metric)))
+        @queue << LogStash::Event.new({ "@timestamp" => snapshot.created_at }.merge(metric.to_hash))
       end
-    end
-
-    # Transform the MetricType into a hash to create a new event
-    def convert_to_hash(metric)
-      {
-        "namespaces" => metric.namespaces,
-        "key" => metric.key,
-        "type" => metric.type,
-        "value" => metric.value
-      }
     end
   end
 end;end
--- a/logstash-core/lib/logstash/instrument/collector.rb
+++ b/logstash-core/lib/logstash/instrument/collector.rb
@@ -19,7 +19,7 @@ module LogStash module Instrument
     include Observable
     include Singleton
 
-    SNAPSHOT_ROTATION_TIME_SECS = 10 # seconds
+    SNAPSHOT_ROTATION_TIME_SECS = 1 # seconds
     SNAPSHOT_ROTATION_TIMEOUT_INTERVAL_SECS = 10 * 60 # seconds
 
     def initialize
@@ -34,9 +34,9 @@ module LogStash module Instrument
     # its the job of the collector to update the store with new metric
     # of update the metric
     #
-    # If there is a problem with the key or the type of metric we will record an error 
+    # If there is a problem with the key or the type of metric we will record an error
     # but we wont stop processing events, theses errors are not considered fatal.
-    # 
+    #
     def push(namespaces_path, key, type, *metric_type_params)
       begin
         metric = @metric_store.fetch_or_store(namespaces_path, key) do
@@ -69,7 +69,7 @@ module LogStash module Instrument
     # @param [Exception] Exception
     def update(time_of_execution, result, exception)
       return true if exception.nil?
-      logger.error("Collector: Something went wrong went sending data to the observers", 
+      logger.error("Collector: Something went wrong went sending data to the observers",
                    :execution_time => time_of_execution,
                    :result => result,
                    :exception => exception)
@@ -78,7 +78,7 @@ module LogStash module Instrument
     # Snapshot the current Metric Store and return it immediately,
     # This is useful if you want to get access to the current metric store without
     # waiting for a periodic call.
-    # 
+    #
     # @return [LogStash::Instrument::MetricStore]
     def snapshot_metric
       Snapshot.new(@metric_store)
--- a/logstash-core/lib/logstash/instrument/metric.rb
+++ b/logstash-core/lib/logstash/instrument/metric.rb
@@ -8,80 +8,70 @@ module LogStash module Instrument
   class MetricNoBlockProvided < MetricException; end
   class MetricNoNamespaceProvided < MetricException; end
 
+  # This class provide the interface between the code, the collector and the format
+  # of the recorded metric.
   class Metric
-    class TimedExecution
-      MILLISECONDS = 1_000_000.0.freeze
+    attr_reader :collector
 
-      def initialize(metric, key)
-        @metric = metric
-        @key = key
-        start
-      end
-
-      def start
-        @start_time = current_time
-      end
-
-      def stop
-        @metric.report_time(@key, (MILLISECONDS * (current_time - @start_time)).to_i)
-      end
-
-      private
-      def current_time
-        Time.now
-      end
-    end
-
-    attr_reader :collector, :namespace_information
-
-    public
-    def initialize(collector, namespace = nil)
+    def initialize(collector)
       @collector = collector
-      @namespace_information = Array(namespace).map(&:to_sym)
     end
 
-    def increment(key, value = 1)
+    def increment(namespace, key, value = 1)
       validate_key!(key)
-      collector.push(namespace_information, key, :counter, :increment, value)
+      collector.push(namespace, key, :counter, :increment, value)
     end
 
-    def decrement(key, value = 1)
+    def decrement(namespace, key, value = 1)
       validate_key!(key)
-      collector.push(namespace_information, key, :counter, :decrement, value)
+      collector.push(namespace, key, :counter, :decrement, value)
     end
 
-    # might be worth to create a block interface for time based gauge
-    def gauge(key, value)
+    def gauge(namespace, key, value)
       validate_key!(key)
-      collector.push(namespace_information, key, :gauge, :set, value)
+      collector.push(namespace, key, :gauge, :set, value)
     end
 
-    def time(key)
+    def time(namespace, key)
       validate_key!(key)
 
       if block_given?
-        timer = TimedExecution.new(self, key)
+        timer = TimedExecution.new(self, namespace, key)
         content = yield
         timer.stop
         return content
       else
-        TimedExecution.new(self, key)
+        TimedExecution.new(self, namespace, key)
       end
     end
 
-    def report_time(key, duration)
-      collector.push(namespace_information, key, :mean, :increment, duration)
+    def report_time(namespace, key, duration)
+      collector.push(namespace, key, :mean, :increment, duration)
     end
 
-    def namespace(sub_namespace)
-      raise MetricNoNamespaceProvided if sub_namespace.nil? || sub_namespace.empty?
+    # This method return a metric instance tied to a specific namespace
+    # so instead of specifying the namespace on every call.
+    #
+    # Example:
+    #   metric.increment(:namespace, :mykey, 200)
+    #   metric.increment(:namespace, :mykey_2, 200)
+    #
+    #   namespaced_metric = metric.namespace(:namespace)
+    #   namespaced_metric.increment(:mykey, 200)
+    #   namespaced_metric.increment(:mykey_2, 200)
+    # ```
+    #
+    # @param name [Array<String>] Name of the namespace
+    # @param name [String] Name of the namespace
+    def namespace(name)
+      raise MetricNoNamespaceProvided if name.nil? || name.empty?
 
-      new_namespace = namespace_information.clone
-      new_namespace << sub_namespace
-
-      Metric.new(collector, new_namespace)
+      NamespacedMetric.new(self, name)
     end
 
+    # Create a Metric instrance using the default Collector singleton reference
+    #
+    #
     def self.create(namespace, collector = LogStash::Instrument::Collector.instance)
       Metric.new(collector, namespace)
     end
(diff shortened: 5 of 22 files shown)


ph:
:bug: Fixing the build. weird I did not have this before pushing.

logstash-core/lib/logstash/agent.rb
```diff
@@ -22,7 +22,7 @@ class LogStash::Agent
 
   def initialize(options = {})
     @pipelines = {}
-     
+
```
purbon: always found this changes very relevant, no just kidding sometimes git thinks we deleted and added a new line :panda_face:

logstash-core/lib/logstash/agent.rb
```diff
@@ -106,7 +106,7 @@ def stop_background_services
   def configure_metric
     if collect_metric?
       @logger.debug("Agent: Configuring metric collection")
-      @metric = LogStash::Instrument::Metric.create(:root)
```
purbon: Curious about it, As I understand we spoke about having as namespaces something like ```stats.events.in```, how does it much here the removal of ```root```? might be I lost it, but I found not references to stats anywhere :smiley_cat:

logstash-core/lib/logstash/filter_delegator.rb
```diff
@@ -31,18 +34,26 @@ def config_name
...
     end
 
     private
     def define_flush_method
       define_singleton_method(:flush) do |options = {}|
-        @filter.flush(options)
+        # we also need to trace the number of events
+        # coming from a specific filters.
+        new_events = @filter.flush(options)
+        @metric_events.increment(:out, new_events.size) unless new_events.nil? # Logstash-filter-aggregates return nil.
```
purbon: re:
> # Logstash-filter-aggregates return nil

is this a common option ? could happen with other filters? what do you think?

logstash-core/lib/logstash/inputs/metrics.rb
```diff
@@ -11,6 +11,7 @@ module LogStash module Inputs
   # Elasticsearch
   class Metrics < LogStash::Inputs::Base
     config_name "metrics"
+    milestone 3
```
purbon: I thought we were not using milestones anymore :+1:

logstash-core/lib/logstash/inputs/metrics.rb
```diff
@@ -35,19 +36,22 @@ def stop
...
-      #   - We can use a synchronization mechanism between the called thread (update method)
-      #   and the plugin thread (run method)
-      #   - How we handle back pressure here?
-      #   - one snashot should be only one event ?
+      # The back pressure is handled in the collector's
+      # scheduled task (running into his own thread) if something append to one of the listener it will
+      # will timeout. In a sane pipeline, with a low traffic of events it shouldn't be a problems.
       snapshot.metric_store.each do |metric|
-        @queue << LogStash::Event.new({ "@timestamp" => snapshot.created_at }.merge(metric.to_hash))
+        @queue << LogStash::Event.new({ "@timestamp" => snapshot.created_at }.merge(convert_to_hash(metric)))
```
purbon: I wonder about this conversion happening here? would not be more coherent that the to_hash method gets implemented in the metric level? I guess it collide with the other sanitization, but the format specified in the ```convert_to_hash``` is not really doing rocket changes. What do you think?

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -19,7 +19,7 @@ class Collector
     include Observable
     include Singleton
 
-    SNAPSHOT_ROTATION_TIME_SECS = 10 # seconds
+    SNAPSHOT_ROTATION_TIME_SECS = 1 # seconds
```
purbon: I know, this is still work in progress, but why this change again? I mean we changed from 1 to 10 and not back to 1, what is your thinking here?

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -8,87 +8,102 @@ class MetricNoKeyProvided < MetricException; end
...
+    def initialize(collector)
       @collector = collector
-      @namespace_information = Array(namespace).map(&:to_sym)
     end
 
-    def increment(key, value = 1)
+    def increment(namespace, key, value = 1)
       validate_key!(key)
-      collector.push(namespace_information, key, :counter, :increment, value)
+      collector.push(namespace, key, :counter, :increment, value)
```
purbon: not sure I like this change, does this means now we could use "any" metric object to run operations over other keys? I understand this provide us more flexibility, but also introduce more degrees of complexity, if we go that path, do we want to make sure the namespace is consistent before issuing the push?

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -8,87 +8,102 @@ class MetricNoKeyProvided < MetricException; end
...
+    #   namespaced_metric.increment(:mykey, 200)
+    #   namespaced_metric.increment(:mykey_2, 200)
+    # ```
+    #
+    # @param name [Array<String>] Name of the namespace
+    # @param name [String] Name of the namespace
+    def namespace(name)
+      raise MetricNoNamespaceProvided if name.nil? || name.empty?
+
+      NamespacedMetric.new(self, name)
```
purbon: oh! now I understand changes inside the Metric class, :smile:

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -8,87 +8,102 @@ class MetricNoKeyProvided < MetricException; end
...
     end
 
-    def self.create(namespace, collector = LogStash::Instrument::Collector.instance)
-      Metric.new(collector, namespace)
+    # Create a Metric instrance using the default Collector singleton reference
+    #
+    #
+    def self.create(collector = LogStash::Instrument::Collector.instance)
+      Metric.new(collector)
```
purbon: should all be ```NamespacedMetric```? what is the benefit of having the first one be a base Metric?

logstash-core/lib/logstash/instrument/metric.rb
```diff
@@ -8,87 +8,102 @@ class MetricNoKeyProvided < MetricException; end
...
       raise MetricNoKeyProvided if key.nil? || key.empty?
     end
+
+    # Allow to calculate the execution of a block of code.
+    # This class support 2 differents syntax a block or the return of
+    # the object itself, but in the later case the metric wont be recorded
+    # Until we call `#stop`.
+    #
+    # @see LogStash::Instrument::Metric#time
+    class TimedExecution
```
purbon: I understand this class is under private scope, right? I wonder what happen in the case of the time function returning an instance of ```TimedExecution``` would this lead to some problem?

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -8,7 +8,14 @@ module LogStash module Instrument
   # saved in a retrievable way, this is a wrapper around multiples ConcurrentHashMap
   # acting as a tree like structure.
   class MetricStore
-    class NamespacesExpectedError < Exception; end
+    class NamespacesExpectedError < StandardError; end
```
purbon: +1 on this.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -8,7 +8,14 @@ module LogStash module Instrument
...
+    class MetricNotFound < StandardError; end
+
+    KEY_PATH_SEPARATOR = "/"
```
purbon: can we freeze constants ? :+1:

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -28,49 +35,107 @@ def fetch_or_store(namespaces, key, default_value = nil)
...
       end
+      return new_hash
     end
 
+    def extract_filter_keys(key)
+      key.to_s.strip.split(FILTER_KEYS_SEPARATOR).map(&:to_sym)
+    end
+
+    def transform_to_array(map)
+      map.values.collect do |value|
```
purbon: I wonder about the output here, are we really only interested on map values?

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -28,49 +35,107 @@ def fetch_or_store(namespaces, key, default_value = nil)
...
+        value.is_a?(Hash) ? transform_to_array(value) : value
+      end.flatten
+    end
+
+    def transform_to_hash(map, new_hash = Hash.new({}))
```
purbon: I saw this notation ```Hash.new({})``` also in other places? any reason to use it? I mean you could just just ```Hash.new``` or ```{}```, right?

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -28,49 +35,107 @@ def fetch_or_store(namespaces, key, default_value = nil)
...
+      map.each_pair do |key, value|
+        if value.is_a?(Concurrent::Map)
```
purbon: can we leverage the internal ```to_hash``` method provided by concurrent map?

logstash-core/lib/logstash/instrument/metric_store.rb
purbon: not sure I like this name, also the ```transform_to_array```, are we transforming or cloning? we don't really do any data transformation, just data structure, right? also why not just using the typical ruby ```to_array``` or ```to_hash```?

logstash-core/lib/logstash/instrument/metric_type/base.rb
```diff
@@ -1,24 +1,24 @@
...
-    def to_event(created_at = Time.now)
-      LogStash::Event.new(to_hash.merge({ "@timestamp" => created_at }))
+    def inspect
+      "#{self.class.name} - namespaces: #{namespaces} key: #{key} value: #{value}"
     end
 
-    def inspect
-      "#{self.class.name} - namespaces: #{@namespaces} key: #{@key} value: #{value}"
+    def to_json
+      LogStash::Json.dump(value)
```
purbon: +1

logstash-core/lib/logstash/instrument/metric_type/counter.rb
```diff
@@ -25,14 +25,5 @@ def execute(action, value = 1)
     def value
       @counter.value
     end
-
-    def to_hash
```
purbon: why not having this info in metric_type base? I see this as a common structure, at least for now for the types, right?

logstash-core/lib/logstash/instrument/namespaced_metric.rb
```diff
@@ -0,0 +1,38 @@
...
+    # @param metric [LogStash::Instrument::Metric] The metric instance to proxy
+    # @param namespace [Array] The namespace to use
+    def initialize(metric, namespace_name)
+      @metric = metric
+      @namespace_name = Array(namespace_name)
+    end
+
+    # Get only the instance methods defined in the class
+    # and not the whole hierarchy of methods.
+    LogStash::Instrument::Metric.public_instance_methods(false).each do |method|
```
purbon: is there any specific reason not to use common inheritance patterns? why not using composition, so we could leverage  metric class methods without exposing the things we don't want ? what do you think?

logstash-core/lib/logstash/instrument/null_metric.rb
```diff
@@ -2,41 +2,39 @@
...
-
-   # Allow to reuse the same variable when creating subnamespace
-   NULL_METRIC_INSTANCE = NullMetric.new
+    attr_reader :namespace_name
 
-   attr_reader :collector, :namespace_information
-   def initialize
+   # Get only the instance methods defined in the class
+   # and not the whole hierarchy of methods.
+   LogStash::Instrument::Metric.public_instance_methods(false).each do |method|
```
purbon: same thing here, why not using composition?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -211,17 +212,22 @@ def start_workers
   end
 
   # Main body of what a worker thread does
-  # Repeatedly takes batches off the queu, filters, then outputs them
+  # Repeatedly takes batches off the queue, filters, then outputs them
   def worker_loop(batch_size, batch_delay)
     running = true
 
+    namespace_events = metric.namespace([:stats, :events])
```
purbon: If I understand properly, stats namespace gets initialized here, right? I guess is better to have it at metrics creation time, isn't? what do you think?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -412,12 +423,14 @@ def shutdown_workers
   def plugin(plugin_type, name, *args)
     args << {} if args.empty?
 
+    pipeline_scoped_metric = metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :plugins])
```
purbon: why the ```to_s.to_sym``` chained conversion? any specific fear to chain them?

logstash-core/lib/logstash/plugin.rb
```diff
@@ -92,7 +92,7 @@ def debug_info
   end
 
   def metric=(new_metric)
-    @metric = new_metric.namespace(@id)
+    @metric = new_metric
```
purbon: This is interesting change, you moved the responsibility to namespace the metric with plugin id from the same plugin to the user of this plugin interface, isn't ? any reason for this? I see this as a more natural behaviour of the plugin base class as we aim to have all plugins request namespaced.

purbon:
@ph did a first round of reviews, good work man! happy to chat with you more in deep if necessary to explain context of your changes :+1:

logstash-core/lib/logstash/agent.rb
ph: Gif diff are weird sometime, you know you can pass `?w=1` to the PR url,  this will make the diff ignore all the whitespace changes. The downside you cannot comment on the pr in that context... We can't have all the nice things.

logstash-core/lib/logstash/agent.rb
ph: This is effectively a good catch I will recheck that.

logstash-core/lib/logstash/filter_delegator.rb
ph: It can happen in any filters that does some kind of internal buffering like the multiline filter.
I can make the comment more generic if its the issues? I wanted to add context for the reviewer for why I was doing that check in this case.

logstash-core/lib/logstash/inputs/metrics.rb
ph: Yes we arent, but not having this make logstash display a yellow warning on boot, I have another PR #4603 that take care of that problem.

logstash-core/lib/logstash/inputs/metrics.rb
ph: It could but If I remember correctly LogStash::Json.dump(hash) will call to_hash on the object before calling `to_json`, I will double check it.

logstash-core/lib/logstash/instrument/collector.rb
ph: It should be in fact exposed to the user when the configuration is done, so right now I think using 1 is not a real issue until I have benchmarked the code.

logstash-core/lib/logstash/instrument/metric.rb
ph: Yes, but this is intended for custom internal pipeline. Lets say you want to update 2 different keys in the same code context.

```
stats.pipelines.events_in
stats.pipelines.pipelines01.plugins.XXXXX.events_in
```

This require 2 different namespace, I think is an OK trade off, we protect the plugins author to create metrics anywhere they want, but we are still able to do it.

logstash-core/lib/logstash/instrument/metric.rb
ph: Yes!

logstash-core/lib/logstash/instrument/metric.rb
ph: This is related to the with change come great responsabilities and the fact that we can create different keys in different context in the pipeline.

This is just a simple factory to use the collector singleton.

logstash-core/lib/logstash/instrument/metric.rb
ph: Private here actually mean this: I don't want the external world to be able to create this class.
But this doesn't prevent the class from returning an instance of it which we can interact with.

We have a test covering that specific use case if you look at the rspec.

logstash-core/lib/logstash/instrument/metric_store.rb
ph: We we can, I always forget to do it.

logstash-core/lib/logstash/instrument/metric_store.rb
ph: We don't need it, since the metric itself knows what namespace and keys its associated.
This make debugging a metric easier and more contained, the structure isn't really important.

I think using `#each` or `#get` depends on what you actually need to output to the user.
In the context on of the api we have a context of hierachy and how an metric rely to other metrics. but in a context of sending event to other system we only need to make sure that a metric has enough information to be meaningful to the external system.

logstash-core/lib/logstash/instrument/metric_store.rb
ph: There is a small difference between `Hash.new()` and `Hash.new({})`, the argument is actually the default values for the keys, this make the code simpler because I don't have to worry about a default values for a key.

```ruby
jruby-1.7.23 :001 > m = Hash.new
 => {}
jruby-1.7.23 :002 > m[:j][:k]
NoMethodError: undefined method `[]' for nil:NilClass

jruby-1.7.23 :003 > m = Hash.new({})
 => {}
jruby-1.7.23 :004 > m[:j][:k] = "dd"

```

ph: I am not a big fan of `transform_*` name in that context , but for me `to_array` or `to_hash` actually mean I am converting the current object to a hash or an array which isn't exactly the case since it doesn't act on the object itself but on a generated subset.

logstash-core/lib/logstash/instrument/metric_store.rb
ph: Yes and no, it doesn't work with nested `Concurrent::Map`, It will just transform the first level.

logstash-core/lib/logstash/instrument/metric_type/counter.rb
ph: I can revisit this, but IIRC I had issue when mixing it with a `to_json` method.

I think `LogStash::Json.dump`  was calling to_hash before calling the `to_json` :(

logstash-core/lib/logstash/instrument/namespaced_metric.rb
ph: Well, the signature of the method change, I can't use it :(

logstash-core/lib/logstash/instrument/null_metric.rb
ph: This reduce the number of lines to code without affecting performance. Actually all theses methods are noop and need to exist.

purbon: and in the other side it makes readability much harder, I wonder if this method is actually faster or slower than pure composite here. What do you think?

logstash-core/lib/logstash/pipeline.rb
ph: If its a number, `to_sym` wont work, it only work on string. :(

logstash-core/lib/logstash/plugin.rb
ph: Thanks for noticing it.

I have struggled with this change a bit, when I changed for a `NamespacedMetric` I though it was more natural to provide a context to the plugin. Since the plugins are actually instantiated in the delegators and can be run in multiples thread I though It was better generate the metric context there to make it more granular.

As a plugin developer he doesnt need to know what logic need to be done to have his keys to be correctly routed to the backend. 

Do you see any advantages to make it there? I am happy to discuss that!

logstash-core/lib/logstash/instrument/null_metric.rb
ph: - Agree on readability.
- I do not believe its slower, we actually use this pattern with the config ast and the filter function.
- Theses methods are created once, when the file is required.

logstash-core/lib/logstash/pipeline.rb
ph: Actually it will initialized at metrics creation time, its a lazy call.

logstash-core/lib/logstash/instrument/metric_store.rb
guyboertje: @ph and @purbon - I have been bitten before on this form of Hash default - the same instance will be used for all missing keys. Make sure this does not cause problems...
```sh
irb(main):001:0> d = Hash.new
=> {}
irb(main):002:0> h = Hash.new(d)
=> {}
irb(main):003:0> h[:a][:b] = 42
=> 42
irb(main):004:0> h[:b][:c] = 43
=> 43
irb(main):005:0> d
=> {:b=>42, :c=>43}
irb(main):006:0> h[:b][:c]
=> 43
irb(main):007:0> h[:a][:c]
=> 43
irb(main):008:0> h[:z][:b]
=> 42
```

ph: Good points!  Thanks @guy I will check that to make sure this is not a problem.

logstash-core/lib/logstash/agent.rb
ph: I rechecked the logic the namespace are created just outside the worker loop in the pipeline.
I understand the logic is a bit more harder to follow.

https://github.com/elastic/logstash/pull/4602/files#diff-bc1e9ea81caa54ea0d56abf4415fdf00R220

The problem is this part of the code actually update two different namespace and increment a counter. The pipeline specific and the application level counter for the number of `events_in`.

This is actually a local aggregation, to reduce complexity in the code base I preferred to update two different counter. What we could do in a future version is to allow to register aggregation callback or create aggregated view on metric store. I think this is a bit more complex to implement and it could be done in a future iteration.

logstash-core/lib/logstash/filter_delegator.rb
ph: Added a clearer comment on that part of the code.

logstash-core/lib/logstash/instrument/collector.rb
ph: Lets make it configurable in a new PR.

logstash-core/lib/logstash/inputs/metrics.rb
ph: Oh this is subtile, I found what problem I had before, Jrjackson does the serializing a bit different than the default `json` library, I think this is for performance reason @guyboertje ?

```ruby
require "json"
require "jrjackson"

class MyTest
  attr_reader :message

  def initialize(message)
    @message = message
  end

  def to_hash
    { "message" => "bye world" }
  end

  def to_json(opts = {})
    message
  end

  def to_json_data
    "to_json_data:  #{message}"
  end
end

class MyTest2
  attr_reader :message

  def initialize(message)
    @message = message
  end

  def to_hash
    { "message" => "bye world" }
  end

  def to_json(opts = {})
    message
  end
end

m = MyTest.new("Hello PH")
m2 = MyTest2.new("Hello PH 2")

puts JSON.dump(m)
puts JrJackson::Json.dump(m)
puts JrJackson::Base.generate(m, {})
puts JrJackson::Base.generate(m2, {})
```

```
Hello PH
Hello PH
"to_json_data:  Hello PH"
{"message":"bye world"}
```

If I override `to_json_data` instead of `to_json` it wont call `to_hash`, @guyboertje is there any specific reason why `JrJackson::Json.dump` and `JrJackson::Base.generate` dont call the same methods?

logstash-core/lib/logstash/instrument/metric_store.rb
ph: TIL, there was actually no error in the code, since in some case we were assigning an empty hash anyway, I have removed the default hash value.

logstash-core/lib/logstash/instrument/metric_type/counter.rb
ph: I move it back to the base class and used `to_json_data` in the base class.

logstash-core/lib/logstash/inputs/metrics.rb
guyboertje: I will reply in Slack because this conversation is hidden under an outdated diff.

logstash-core/lib/logstash/agent.rb
purbon: If I understood your comment properly, you say that https://github.com/elastic/logstash/pull/4602/files#diff-bc1e9ea81caa54ea0d56abf4415fdf00R220 is basically creating the namespace?

logstash-core/lib/logstash/inputs/metrics.rb
purbon: good for #4603 !!

logstash-core/lib/logstash/instrument/collector.rb
purbon: +1

logstash-core/lib/logstash/instrument/metric.rb
purbon: Ok! I understand. Good we are not passing this to plugins, can we might be add a validation/defense to check the namespace actually exist? I see a potential error where people actually create metrics to unknown locations.

logstash-core/lib/logstash/instrument/metric.rb
purbon: I understand, as I shared earlier I think I would add some kind of defense/validation to prevent people falling on creating unexpected namespaces, what do you think?

logstash-core/lib/logstash/instrument/metric.rb
purbon: :smile:

logstash-core/lib/logstash/instrument/metric_store.rb
purbon: Interesting,  my question here goes more on the side of if I ask for namespace `foo.bar` i might expect to also get data for `foo.bar.zoo`, what do you think?

logstash-core/lib/logstash/instrument/metric_type/counter.rb
purbon: @ph might be need to fix `LogStash::Json.dump` any reason to do that like this?

logstash-core/lib/logstash/instrument/metric.rb
ph: Can we do it as the part of the #4506 ?

logstash-core/lib/logstash/instrument/metric_store.rb
ph: lets say that you have the following structure
```
{ :foo => { :bar = > { :zoo => 1, :hello => { :ninja => 2 } } }
```
If you do `each("foo/bar")` you will get [1, 2] since they are both under `foo.bar`

logstash-core/lib/logstash/pipeline.rb
purbon: so you mean this will create the namespace first time we actually store something? what happen then for things with value zero?

logstash-core/lib/logstash/pipeline.rb
purbon: makes sense

logstash-core/lib/logstash/pipeline.rb
ph: We dont have values until its gathered?

logstash-core/lib/logstash/agent.rb
ph: The namespace will only be created if we actually collect a metric `increment` in that case.

logstash-core/lib/logstash/instrument/metric.rb
ph: I suggest we make it part of #4506

logstash-core/lib/logstash/plugin.rb
purbon: This is actually good question, my expectations somehow match with yours, plugins should not be aware of namespaces, they should just get the right metrics object. But if you ask me about internal logstash location for this change I think it belongs to the plugin base class. Having this change in the delegators makes you have duplicated code and introduced potential issues with plugins when the metric object is passed without the "right" namespace. What do you think?

logstash-core/lib/logstash/pipeline.rb
purbon: I know that, just wondering what will be the expected output and behaviour for a metric namespace created, but without any data. 

I know now we create the namespace the same time we insert the data, so having no values stored means no namespace created. Do you think it could be the case we aim to show a metric with value zero?

logstash-core/lib/logstash/instrument/metric.rb
purbon: +1

logstash-core/lib/logstash/instrument/metric.rb
purbon: sure, just update the issue with it.

logstash-core/lib/logstash/instrument/metric_store.rb
purbon: this makes this confusing, isn't?

purbon:
I'm good to merge this, if we open two issues to fix in next PR's this features:

* https://github.com/elastic/logstash/pull/4602#discussion_r51553770
* https://github.com/elastic/logstash/pull/4602#discussion_r51736794
* https://github.com/elastic/logstash/pull/4602#discussion_r51736491

what do you think?

logstash-core/lib/logstash/instrument/metric_store.rb
ph: Well, It depends, it you need to retain structure you can use `#get` instead of each?

logstash-core/lib/logstash/instrument/metric.rb
ph: +1 awesome

logstash-core/lib/logstash/instrument/metric_store.rb
purbon: I see both providing equivalent features, if the two of them have different output we should be sure to document this very clear, otherwise make the api confusing.

Issue #4562, opened 2016-01-25
Milestone warning is displayed when pluginpath is used


suyograo:
`--pluginpath ` can be used with LS to load plugins which are not converted into RubyGems yet. LS will output a warning when used with this option:

> 
:message=>"abc plugin is using the 'milestone' method to declare the version of the plugin this method is deprecated in favor of declaring the version inside the gemspec.", :level=>:warn} 

This is a benign message, and can be ignored, but we need to remove this warning so it does not confuse users.

ph:
I came around this issue myself when working on metrics, I suggest we move the log level to `debug`?

ph:
And we remove the support for milestone in the next major version.

PR #4603, opened 2016-02-01
Hide milestone confusing message to the user

ph:
This PR change the log level when a user load a plugin with a milestone
option to debug instead of warn.

Fixes: #4562

### DIFF
--- a/logstash-core/lib/logstash/config/mixin.rb
+++ b/logstash-core/lib/logstash/config/mixin.rb
@@ -147,7 +147,7 @@ module LogStash::Config::Mixin
     # inside the gemspec.
     def milestone(m = nil)
       @logger = Cabin::Channel.get(LogStash)
-      @logger.warn(I18n.t('logstash.plugin.deprecated_milestone', :plugin => config_name))
+      @logger.debug(I18n.t('logstash.plugin.deprecated_milestone', :plugin => config_name))
     end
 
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -147,7 +147,7 @@ def plugin_status(status = nil)
     # inside the gemspec.
     def milestone(m = nil)
       @logger = Cabin::Channel.get(LogStash)
-      @logger.warn(I18n.t('logstash.plugin.deprecated_milestone', :plugin => config_name))
+      @logger.debug(I18n.t('logstash.plugin.deprecated_milestone', :plugin => config_name))
```
suyograo: Hmm, I think we need this warning. Maybe at `info`. Question is, should this messages be logged when `pluginpath` option is used.

ph: Well, From the issue I understood that this was confusing people, since it doesn't really interfere with the plugins. I though it was more of a developer focus message than a user focus message this is why I went ahead with debug?

This was my reasoning behind it.

suyograo:
LGTM

logstash-core/lib/logstash/config/mixin.rb
suyograo: Makes sense.

PR #4605, opened 2016-02-01
Make sure that sprintf return a new clone

### LINKED ISSUE #4592: Undocumented behaviour with add_field
I recognized that after calling the add_field method to get a copy of a field both fields point to the same content. So changes to the new field also cause the same change to the old field.
Let us assume I have created a field Mapping with the uppercase string "FOOBAR" and I do the following steps:
- mutate{add_field => {"mappingout" => "%{Mapping}"} }
- mutate{lowercase => ["mappingout"]}
Now both fields contain the lowercase string foobar
The following workaround is applicable:
- mutate{add_field => {"mappingout" => "tt%{Mapping}"} }
- mutate{lowercase => ["mappingout"]}
Using the workaround Mapping contains the original uppercase string FOOBAR and mappingout the lowercase ttfoobar.

I am not sure wether this behaviour is accidently or not. If not the documentation of add_field should be adapted and highlight this.

ph:
In some cases `Event#sprintf` was returning a reference to the value and
not a new instance. This would lead in some weird behavior when using
`add_fields` with the sprintf syntax.

Fixes #4592

### DIFF
--- a/logstash-core-event/lib/logstash/string_interpolation.rb
+++ b/logstash-core-event/lib/logstash/string_interpolation.rb
@@ -125,7 +125,9 @@ module LogStash
       when Hash
         LogStash::Json.dump(value)
       else
-        value
+        # Make sure we dont work on the refence of the value
+        # The Java Event implementation was always returning a string.
+        "#{value}"
       end
     end
   end
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -1,5 +1,6 @@
 # encoding: utf-8
 require "spec_helper"
+require "logstash/util/decorators"
 require "json"
 
 describe LogStash::Event do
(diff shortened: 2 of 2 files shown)


ph:
@guyboertje Actually the failing test made me realize that the Java version already return a string all the time. So "#{value}" would make it to parity with the java version what do you think?

guyboertje:
@ph - +1 return string.  Let keep parity. If we get complaints we can fix in both places. Better to know now than when we start using Java Event. :smile:

logstash-core-event/lib/logstash/string_interpolation.rb
```diff
@@ -125,7 +125,9 @@ def evaluate(event)
       when Hash
         LogStash::Json.dump(value)
       else
-        value
+        # Make sure we dont work on the refence of the value
+        # The Java Event implementation was always returning a string.
+        "#{value}"
```
jsvd: is there any benefit to doing value.to_s instead? not sure.

jsvd: also maybe

```
# ...
case String
  value
else
  "#{value}"
end
```

guyboertje: it needs to be a new string instance such that if you upcase! the returned value it does not mutate the original.

In jruby to_s on String returns a reference to the original.

guyboertje:
LGTM

logstash-core-event/lib/logstash/string_interpolation.rb
jsvd: :facepalm: nevermind.

jsvd:
LGTM

PR #4608, opened 2016-02-02
Initial .travis.yml

andrewvc:
This patch:

* Moves us to Travis for PR based pull requests
* Uses `rake test:install-core && rake test:core-fail-fast` rather than the `ci/*` scripts for simplicity. Travis starts clean ever time
* Caches the `vendor/bundle` and `.gradle` directories for speed. Infrequently running Jenkins jobs can check for issues the latest versions of gems.
* Updates README.md to point to the new tests, and hides buttons to the defunct all-plugins tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ cache:
     - ~/.gradle/
 rvm:
   - jruby-1.7.25
-  - jruby-9.0.5.0
 jdk:
   - oraclejdk8
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+sudo: false
+language: ruby
+cache:
+  directories:
+    - vendor/bundle
+    - ~/.gradle/
+rvm:
+  - jruby-1.7.25
+  - jruby-9.0.5.0
+jdk:
+  - oraclejdk8
+before_script:
+  - rake test:install-core
+  - echo "--order rand" > .rspec
+  - echo "--format documentation" >> .rspec
+script: rake test:core
--- a/README.md
+++ b/README.md
@@ -2,10 +2,9 @@
 
 ### Build status
 
-| Branch | master | 2.x | 2.1
+| Test | master | 5.0 | 2.3 |
 |---|---|---|---|
-| core | [![Build Status](http://build-eu-00.elastic.co/view/LS%20Master/job/logstash_regression_master/badge/icon)](http://build-eu-00.elastic.co/view/LS%20Master/job/logstash_regression_master/) | [![Build Status](http://build-eu-00.elastic.co/view/LS%202.x/job/logstash_regression_2x/badge/icon)](http://build-eu-00.elastic.co/view/LS%202.x/job/logstash_regression_2x/) | [![Build Status](http://build-eu-00.elastic.co/view/LS%202.x/job/logstash_regression_21/badge/icon)](http://build-eu-00.elastic.co/view/LS%202.x/job/logstash_regression_21/) |
-| integration | [![Build Status](http://build-eu-00.elastic.co/view/LS%20Master/job/Logstash_Master_Default_Plugins/badge/icon)](http://build-eu-00.elastic.co/view/LS%20Master/job/Logstash_Master_Default_Plugins/) | [![Build Status](http://build-eu-00.elastic.co/view/LS%202.x/job/Logstash_Default_Plugins_2x/badge/icon)](http://build-eu-00.elastic.co/view/LS%202.x/job/Logstash_Default_Plugins_2x/) | [![Build Status](http://build-eu-00.elastic.co/view/LS%202.x/job/Logstash_Default_Plugins_21/badge/icon)](http://build-eu-00.elastic.co/view/LS%202.x/job/Logstash_Default_Plugins_21/) |
+| core | [![Build Status](https://travis-ci.org/elastic/logstash.svg?branch=master)](https://travis-ci.org/elastic/logstash) | [![Build Status](https://travis-ci.org/elastic/logstash.svg?branch=2.3)](https://travis-ci.org/elastic/logstash) | [![Build Status](https://travis-ci.org/elastic/logstash.svg?branch=2.3)](https://travis-ci.org/elastic/logstash) |
 
 Logstash is a tool for managing events and logs. You can use it to collect
 logs, parse them, and store them for later use (like, for searching).  If you


.travis.yml
```diff
@@ -0,0 +1,9 @@
+sudo: false
+language: ruby
+cache:
+  directories:
+    - vendor/bundle
+    - .gradle
+rvm:
+  - jruby-1.7.24
+script: rake bootstrap test:install-default && rake test:core test:plugins
```
purbon: Easier if you use the scripts under the ```ci``` directory, so in case we change the tasks names (for example between versions) we don't have to update the set config.

purbon: Inside that scripts you will find the right flow for test described.

andrewvc: @purbon I noticed that the ci setup removes the 'vendor' directory. Do we really want this for every test run? That will really slow down the specs! Maybe we can discuss tomorrow?

purbon: @andrewvc the idea is to start with a _clean_ environment, we could also achieve this by cleaning the hole workspace. Happy to see ways to achieve this with you.

andrewvc: Well, in that case its super easy in travis, since it is by default clean. I just need to disable the caching directives I specified above. I guess what I'm saying is that I'm not sure if it's worth the slowdown, but I suppose that can be  resolved in a future PR. Should we also not cache `.gradle` then? I assume that's fine to cache. Right @colinsurprenant ?

andrewvc:
@purbon how's this looking now?

purbon:
looks much better now! anyway to actually test the tests?

purbon:
can we have other scripts for all plugins / default plugins tests?

andrewvc:
@purbon this is now ready for your review!

README.md
```diff
@@ -2,10 +2,9 @@
...
-| integration | [![Build Status](http://build-eu-00.elastic.co/view/LS%20Master/job/Logstash_Master_Default_Plugins/badge/icon)](http://build-eu-00.elastic.co/view/LS%20Master/job/Logstash_Master_Default_Plugins/) | [![Build Status](http://build-eu-00.elastic.co/view/LS%202.x/job/Logstash_Default_Plugins_2x/badge/icon)](http://build-eu-00.elastic.co/view/LS%202.x/job/Logstash_Default_Plugins_2x/) | [![Build Status](http://build-eu-00.elastic.co/view/LS%202.x/job/Logstash_Default_Plugins_21/badge/icon)](http://build-eu-00.elastic.co/view/LS%202.x/job/Logstash_Default_Plugins_21/) |
+| core | [![Build Status](https://travis-ci.org/elastic/logstash.svg?branch=master)](https://travis-ci.org/elastic/logstash) | [![Build Status](https://travis-ci.org/elastic/logstash.svg?branch=2.3)](https://travis-ci.org/elastic/logstash) | [![Build Status](https://travis-ci.org/elastic/logstash.svg?branch=2.3)](https://travis-ci.org/elastic/logstash) |
```
ph: I clicked view on the readme and the table seems broken?

.travis.yml
```diff
@@ -0,0 +1,11 @@
...
+language: ruby
+cache:
+  directories:
+    - vendor/bundle
+    - .gradle
+rvm:
+  - jruby-1.7.24
+jdk:
+  - oraclejdk8
+script: rake test:install-core && rake test:core-fail-fast
```
ph: We want to fail fast, I would prefer to see a complete run?

andrewvc: @ph I believe that was the old behavior. I'm glad to switch it, slow failure makes more sense to me. Also, I'll have it use verbose rspec formatting.

.travis.yml
```diff
@@ -0,0 +1,11 @@
+sudo: false
+language: ruby
+cache:
+  directories:
+    - vendor/bundle
+    - .gradle
+rvm:
+  - jruby-1.7.24
```
ph: We can use .25, there a PR ready about to be merged at any moment.
Can we use ERB in travis template? That would be cool to synchronize the version with this value https://github.com/ph/logstash/blob/263232a5e4b54c5fb286845bdc630ff494c9eb0a/rakelib/vendor.rake#L3

.travis.yml
ph: Another question, can we specify a JVM on travis?

README.md
ph: screenshot 2016-04-18 16 18 50

.travis.yml
```diff
@@ -0,0 +1,16 @@
+sudo: false
+language: ruby
+cache:
+  directories:
+    - vendor/bundle
+    - .gradle
+rvm:
+  - jruby-1.7.25
+  - jruby-9.0.5.0
```
suyograo: not sure we need to test on 9K on every PR, yet.

andrewvc: Hmm, well, it runs in parallel so it shouldn't slow things down. I just figure that if the tests pass that's a bonus since we'll have to move to it at some point?

WDYT?

suyograo: >  just figure that if the tests pass that's a bonus

Ok, makes sense to run both... Just don't want us to block a PR because it fails on JRuby 9K (yet)

suyograo:
LGTM

README.md
```diff
@@ -2,10 +2,9 @@
...
```
suyograo: Shouldn't this be master, 5.0 and 2.3. 2.3 is repeated twice here.

andrewvc: Ack, good catch! Will fix :)

ph:
Why we also have circle ci?
Other than that LGTM

andrewvc:
@ph I thought I'd add that in to test it out. I got frustrated waiting ~30m for a travis-ci build and threw that in there:

The output looks pretty nice! https://circleci.com/gh/elastic/logstash/10#tests maybe we should consider it? Circle will actually let us pay for dedicated boxes.

andrewvc:
@purbon this actually still is a WIP, but it's getting there!

ph:
@andrewvc I have paid for circle.ci in a previous jobs, I had no issue with it.

ph:
@andrewvc I am not sure we should try circleci for now, we are trying to reduce the noise level :)

andrewvc:
Yeah I agree. I'll move circle to a separate branch so it'll be an option
if we need it.

suyograo:
LGTM

andrewvc:
@suyograo I'm good to (finally) merge this if you are!

suyograo:
@andrewvc lets do it

Issue #4620, opened 2016-02-03
Metrics to stdout not working with logstash 2.2.0


matejzero:
Hello,

today I updated logstash from 2.1.1 to 2.2.0 on my test server and metrics output stopped working.

I have the following setup (it's a test / benchmark server):
- file input plugins reads a file
- metrics plugins counts the number of events and adds tag 'metric'
- if there is 'metric' tag in the event, print it out to stdout, else send event to /dev/null

Somehow, I don't get any text on stdout.
Is it connected with the new pipeline? Is metrics plugin even useful now, with the new pipeline design?

My config:
```
input {
  file {
    path => "/mnt/logstash-test-log.1"
    start_position  => "beginning"
    sincedb_path => "/dev/null"
    ignore_older => 999999999999
    add_field => { "syslog_format" => "true" }
  }
}
filter {
    metrics {
        meter => "events"
        add_tag => "metric"
        add_field => { "metric_type" => "Input" }
        flush_interval => 10
    }
}
output {
  if "metric" in [tags] {
    stdout {
      codec => line {
        format => "%{[metric_type]}:  1m rate: %{[events][rate_1m]} ( %{[events][count]} ) | 5m rate: %{[events][rate_5m]} ( %{[events][count]} )"
      }
    }
  }
}
```

The same config is working on 2.1.1.

Matej

andrewvc:
@matejzero thanks for reporting this. We've traced this to a bug in the flusher code for logstash 2.2.0 , there is a race condition where it sometimes will not start. This should be fixed by this new PR I just made: https://github.com/elastic/logstash/issues/4620

matejzero:
@andrewvc: great, thanks... Will try the fix tomorrow at work

andrewvc:
Thanks for the bug report @matejzero ! This was a bad bug. We hope to release a 2.2.1 version sometime this week to get this out.

matejzero:
I did some testing and it looks like it's working as it should. Thanks for the fix.

suyograo:
Released a snapshot version for anyone looking: 
https://download.elastic.co/logstash/logstash/logstash-2.2.1-SNAPSHOT.zip
https://download.elastic.co/logstash/logstash/logstash-2.2.1-SNAPSHOT.tar.gz

PR #4623, opened 2016-02-03
Fix flusher initialization to prevent race breaking flushing

andrewvc:
If start_flusher is called before @running is set to true it may shut itself
down before it truly starts. This fixes that bug.

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -72,8 +72,6 @@ module LogStash; class Pipeline
     @ready = Concurrent::AtomicBoolean.new(false)
     @running = Concurrent::AtomicBoolean.new(false)
     @flushing = Concurrent::AtomicReference.new(false)
-
-    start_flusher
   end # def initialize
 
   def ready?
@@ -131,6 +129,7 @@ module LogStash; class Pipeline
     # Generally this happens if SIGINT is sent and `shutdown` is called from an external thread
 
     @running.make_true
+    start_flusher # Launches a non-blocking thread for flush events
     wait_inputs
     @running.make_false
 
@@ -415,6 +414,9 @@ module LogStash; class Pipeline
   end
 
   def start_flusher
+    # Invariant to help detect improper initialization
+    raise "Attempted to start flusher on a stopped pipeline!" if @running.false?
+
     @flusher_thread = Thread.new do
       while Stud.stoppable_sleep(5, 0.1) { @running.false? }
         flush
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -128,10 +128,10 @@ module LogStash; class Pipeline
     # Block until all inputs have stopped
     # Generally this happens if SIGINT is sent and `shutdown` is called from an external thread
 
-    @running.make_true
+    transition_to_running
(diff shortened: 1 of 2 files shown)


jsvd:
may I suggest a test covering this?

andrewvc:
Fixes https://github.com/elastic/logstash/issues/4620

ph:
LGTM, Good refactoring of the booleans

andrewvc:
@jsvd can you take a look and be the second LGTM here?

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -232,6 +232,24 @@ class TestPipeline < LogStash::Pipeline
...
+        allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(DummyOutput)
+      end
+
+      let(:config) { "input { dummyinput {} } output { dummyoutput {} }"}
+
+      it "should start the flusher thread only after the pipeline is running" do
+        pipeline = TestPipeline.new(config)
+
+        expect(pipeline).to receive(:transition_to_running).ordered.and_call_original
+        expect(pipeline).to receive(:start_flusher).ordered.and_call_original
```
jsvd: I think this test should check that the call to `flush` was made, instead of `start_flusher`

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -72,8 +72,6 @@ def initialize(config_str, settings = {})
     @ready = Concurrent::AtomicBoolean.new(false)
     @running = Concurrent::AtomicBoolean.new(false)
     @flushing = Concurrent::AtomicReference.new(false)
-
-    start_flusher
   end # def initialize
 
   def ready?
```
jsvd: maybe we can take the opportunity to remove "ready?" and replace with running?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -415,10 +430,13 @@ def flush_filters(options = {}, &block)
   end
 
   def start_flusher
+    # Invariant to help detect improper initialization
+    raise "Attempted to start flusher on a stopped pipeline!" if stopped?
```
jsvd: is there a scenario where we predict this to happen? `start_flusher` is called immediately after making `@running` true.

logstash-core/spec/logstash/pipeline_spec.rb
andrewvc: I think that should be a different unit test. `start_flusher` calls `flush` on a periodic basis. We don't want this test to yield a false positive in the case where somehow `flush` is only called once (or twice, or three times!) but the flusher itself is not started.

I do think we could use a separate unit test for the flusher's behavior.

logstash-core/lib/logstash/pipeline.rb
andrewvc: No, there isn't, but I felt this behavior was important enough that a future-proofing invariant was called for. This may catch a dumb mistake down the road.

jsvd:
LGTM

logstash-core/lib/logstash/pipeline.rb
andrewvc: I think that can wait for a future refactor, I'd like to keep this scope to the bug at hand.

Issue #4599, opened 2016-02-01
New java event interfere the plugin manager uninstall command with cached gems


purbon:
The introduction, and enablement, of the new java event in master broke the way the offline installation works, mostly affecting the uninstall command when there are plugins cached.

Due to the fact that we empty the internal bundler caches, aka app cache during package time, we introduced a hack that forced this caches to be rebuild, see

```
      ::Bundler::Source::Rubygems.module_exec do
        def cached_gem(spec)
          cached_built_in_gem(spec)
        end
      end
    end
```

https://github.com/elastic/logstash/blob/master/lib/bootstrap/bundler.rb#L35-L40

This basically pulls all requested specs from rubygems, in case they are not actually installed inside the internal cache. But as the new `logstash-core-event-java` is not at rubygems, this process fails. Uninstallation still works, but users see a set of nasty messages that can be misleading.

If no cache is present:

```
skywalker% bin/plugin install logstash-input-stdin
Validating logstash-input-stdin
Installing logstash-input-stdin
Installation successful
skywalker% ls vendor                              
_  bundle  jruby
skywalker% bin/plugin uninstall logstash-input-stdin
Uninstalling logstash-input-stdin
```

if there are plugins cached:

```
skywalker% ls vendor 
_  bundle  cache  jruby
skywalker% bin/plugin install logstash-input-stdin
Validating logstash-input-stdin
Installing logstash-input-stdin
Installation successful
skywalker% bin/plugin uninstall logstash-input-stdin
Uninstalling logstash-input-stdin
Error NoMethodError, retrying 1/10
undefined method `source_uri' for nil:NilClass
WARNING: SSLSocket#session= is not supported
Error NoMethodError, retrying 2/10
undefined method `source_uri' for nil:NilClass
WARNING: SSLSocket#session= is not supported
```

after that uninstalled plugin is gone, but the set of Error and warnings looks bad :smiley: :+1: 

@colinsurprenant what do you think about releasing event-java to rubygems? is this an option?

colinsurprenant:
this problem should not be specific to `logstash-core-event-java`, the mechanism/setup should be **exactly** the same with the Ruby `logstash-core-event`

there are 3 logstash execution contexts:
1- local dev dir: where the gems are located in the source tree respective dirs `logstash-core-event/` or `logstash-core-event-java/`.
[...]

PR #4630, opened 2016-02-04
force ruby-maven version

colinsurprenant:
as of Feb 3rd 2016, the ruby-maven gem is resolved to version 3.3.3 and that version has an rdoc problem that causes a bundler exception. 3.3.9 is the current latest version which does not have this problem.

/cc @purbon @ph


purbon:
LGTM, also test this locally and works.

ph:
LGTM, tested locally  Fixed the issue I had this morning. Good catch all!

colinsurprenant:
will fix #4599

colinsurprenant:
thanks. rebasing and merging.

colinsurprenant:
merged in master and 2.x

PR #4631, opened 2016-02-04
add Event#from_json method

### LINKED ISSUE #4595: introduce Event#from_json to avoid type conversions in Java Event
As described in #4481 I suggest we add `Event#from_json` for json deserialization to avoid double type conversions when used in the context of the Java Event. 

The best example is the json_lines codec which uses the `LogStash::Json.load` method [to deserialize json](https://github.com/logstash-plugins/logstash-codec-json_lines/blob/1eb838d12321a05f291c3836dce51c230605f915/lib/logstash/codecs/json_lines.rb#L55) which produces a Ruby object tree that is in turn passed to the Java Event which has to convert back into  native Java objects. 

By introducing  `Event#from_json` we would be able to short-circuit this double type conversion and improve significantly the deserialization speed. To keep backward compatibility, the codec could just first check for `Event.method_defined?(:from_json)` and choose the best strategy.

colinsurprenant:
will fix #4595 

This introduces `Event#from_json` which takes as argument a json string and returns a new `Event` object. The idea here is to update the json_lines codec to use this method to avoid double type conversion by first parsing the json to a Ruby structure then passing that structure to the Java Event and go through unnecessary type conversions. the POC has shown a ~50% performance increase using this. 

Note that this PR adds specs for the input corner cases behaviour of the current deserialization behaviour to make sure `from_json` behaves the same way.

### DIFF
--- a/logstash-core-event-java/src/main/java/com/logstash/Event.java
+++ b/logstash-core-event-java/src/main/java/com/logstash/Event.java
@@ -1,7 +1,7 @@
 package com.logstash;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.logstash.ext.JrubyTimestampExtLibrary;
-import org.codehaus.jackson.map.ObjectMapper;
 import org.joda.time.DateTime;
 import org.jruby.RubySymbol;
 
@@ -153,8 +153,38 @@ public class Event implements Cloneable, Serializable {
         }
     }
 
-    public String toJson() throws IOException {
+    public String toJson()
+            throws IOException
+    {
         return mapper.writeValueAsString((Map<String, Object>)this.data);
+    }
+
+    public static Event[] fromJson(String json)
+            throws IOException
+    {
+        Event[] result;
+
+        if (json == null || json.trim().isEmpty()) {
+            return new Event[]{ new Event() };
+        }
+
+        Object o = mapper.readValue(json, Object.class);
+        // we currently only support Map or Array json objects
+        if (o instanceof Map) {
+            result = new Event[]{ new Event((Map)o) };
+        } else if (o instanceof List) {
+            result = new Event[((List) o).size()];
+            int i = 0;
+            for (Object e : (List)o) {
+                if (!(e instanceof Map)) {
+                    throw new IOException("incompatible inner json array object type=" + e.getClass().getName() + " , only hash map is suppoted");
+                }
+                result[i++] = new Event((Map)e);
+            }
+        } else {
+            throw new IOException("incompatible json object type=" + o.getClass().getName() + " , only hash map or arrays are suppoted");
+        }
+        return result;
     }
 
     public Map toMap() {
--- a/logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
+++ b/logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
@@ -35,6 +35,10 @@ import java.util.List;
 
 public class JrubyEventExtLibrary implements Library {
 
+    private static RubyClass PARSER_ERROR = null;
+    private static RubyClass GENERATOR_ERROR = null;
+    private static RubyClass LOGSTASH_ERROR = null;
+
     public void load(Ruby runtime, boolean wrap) throws IOException {
         RubyModule module = runtime.defineModule("LogStash");
 
@@ -54,6 +58,19 @@ public class JrubyEventExtLibrary implements Library {
         clazz.setConstant("VERSION_ONE", runtime.newString(Event.VERSION_ONE));
         clazz.defineAnnotatedMethods(RubyEvent.class);
         clazz.defineAnnotatedConstants(RubyEvent.class);
+
+        PARSER_ERROR = runtime.getModule("LogStash").defineOrGetModuleUnder("Json").getClass("ParserError");
+        if (PARSER_ERROR == null) {
+            throw new RaiseException(runtime, runtime.getClass("StandardError"), "Could not find LogStash::Json::ParserError class", true);
+        }
+        GENERATOR_ERROR = runtime.getModule("LogStash").defineOrGetModuleUnder("Json").getClass("GeneratorError");
+        if (GENERATOR_ERROR == null) {
+            throw new RaiseException(runtime, runtime.getClass("StandardError"), "Could not find LogStash::Json::GeneratorError class", true);
+        }
+        LOGSTASH_ERROR = runtime.getModule("LogStash").getClass("Error");
+        if (LOGSTASH_ERROR == null) {
+            throw new RaiseException(runtime, runtime.getClass("StandardError"), "Could not find LogStash::Error class", true);
+        }
     }
 
     public static class ProxyLogger implements Logger {
@@ -107,7 +124,7 @@ public class JrubyEventExtLibrary implements Library {
             args = Arity.scanArgs(context.runtime, args, 0, 1);
             IRubyObject data = args[0];
 
-            if (data.isNil()) {
+            if (data == null || data.isNil()) {
                 this.event = new Event();
             } else if (data instanceof RubyHash) {
                 HashMap<String, Object>  newObj = Javafier.deep((RubyHash) data);
@@ -215,7 +232,7 @@ public class JrubyEventExtLibrary implements Library {
             try {
                 return RubyString.newString(context.runtime, event.sprintf(format.toString()));
             } catch (IOException e) {
-                throw new RaiseException(getRuntime(), (RubyClass)getRuntime().getModule("LogStash").getClass("Error"), "timestamp field is missing", true);
+                throw new RaiseException(getRuntime(), LOGSTASH_ERROR, "timestamp field is missing", true);
             }
         }
 
@@ -251,9 +268,38 @@ public class JrubyEventExtLibrary implements Library {
 
         @JRubyMethod(name = "to_json", rest = true)
         public IRubyObject ruby_to_json(ThreadContext context, IRubyObject[] args)
-            throws IOException
         {
-            return RubyString.newString(context.runtime, event.toJson());
+            try {
+                return RubyString.newString(context.runtime, event.toJson());
+            } catch (Exception e) {
+                throw new RaiseException(context.runtime, GENERATOR_ERROR, e.getMessage(), true);
+            }
+        }
+
+        // @param value [String] the json string. A json object/map will convert to an array containing a single Event.
+        // and a json array will convert each element into individual Event
+        // @return Array<Event> array of events
+        @JRubyMethod(name = "from_json", required = 1, meta = true)
+        public static IRubyObject ruby_from_json(ThreadContext context, IRubyObject recv, RubyString value)
+        {
+            Event[] events;
+            try {
+                events = Event.fromJson(value.asJavaString());
+            } catch (Exception e) {
+                throw new RaiseException(context.runtime, PARSER_ERROR, e.getMessage(), true);
+            }
+
+            RubyArray result = RubyArray.newArray(context.runtime, events.length);
+
+            if (events.length == 1) {
+                // micro optimization for the 1 event more common use-case.
+                result.set(0, RubyEvent.newRubyEvent(context.runtime, events[0]));
+            } else {
+                for (int i = 0; i < events.length; i++) {
+                    result.set(i, RubyEvent.newRubyEvent(context.runtime, events[i]));
+                }
+            }
+            return result;
         }
 
         @JRubyMethod(name = "validate_value", required = 1, meta = true)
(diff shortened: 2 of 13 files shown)


colinsurprenant:
I will push a json_lines codec PR for this shortly to show the usage in a backward compatible way.

colinsurprenant:
the build fail is because of the ruby-maven problem I believe, fixed in #4630. I will merge and rebase here to see.

andrewvc:
@colinsurprenant isn't this a bit strange? Shouldn't we be just rewriting JSON Lines to be a java plugin?

andrewvc:
Just to clarify, we've basically moved one codec into the Event class, isn't this not a pattern we want to encourage?

colinsurprenant:
@andrewvc well, ultimately yes, having the codec in Java will be awesome, until then, this is the best option I believe. In any case, I think that if we have a `to_json` method, it does somehow make sense to also have a `from_json`?  Since json ingestion support is such an important use-case to support, I think this will provide good value by bringing the performance improvement of today's Java Event implementation to to json ingestion use-case. WDYT?

colinsurprenant:
rebased against master, let's see if it solves the ruby-maven bobo

colinsurprenant:
@andrewvc also, this does not moves the codec into the event class, it moves the equivalent of [`LogStash::Json.load`](https://github.com/elastic/logstash/blob/1260f63a2acb913442377101a827cd59ea766ad9/logstash-core/lib/logstash/json.rb#L34-L42)  in `Event` 
The codec still does the line tokenization etc

colinsurprenant:
now the error is
```
rspec ./logstash-core/spec/logstash/runner_spec.rb:36 # LogStash::Runner argument parsing with no arguments should show help
```
which seems unrelated to these commits :confused:

colinsurprenant:
Ok, I am able to reproduce the build failure when on Java 7 and running
```
$ SPEC_OPTS="--seed 61716" rake test:core
```
IMO it is not related to this PR. Will create a separate issue.

The complete error is

```
Failures:

  1) LogStash::Runner argument parsing with no arguments should show help
     Failure/Error: subject.run(args)
       (#<Cabin::Channel:0x15462876>).warn("You may be interested in the '--configtest' flag which you can use to validate\nlogstash's configuration before you choose to restart a running system.")
           expected: 1 time with any arguments
           received: 2 times with arguments: ("You may be interested in the '--configtest' flag which you can use to validate\nlogstash's configuration before you choose to restart a running system.")
     # ./logstash-core/lib/logstash/runner.rb:164:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67:in `run'
     # ./logstash-core/spec/logstash/runner_spec.rb:40:in `(root)'

Finished in 13.78 seconds (files took 3.14 seconds to load)
1202 examples, 1 failure

Failed examples:

rspec ./logstash-core/spec/logstash/runner_spec.rb:36 # LogStash::Runner argument parsing with no arguments should show help

Randomized with seed 61716
```

andrewvc:
@colinsurprenant makes sense for now! So long as we don't add any more from/to methods I'm :+1:

colinsurprenant:
ok, cool. who's next?  :grimacing:

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -251,9 +251,24 @@ public IRubyObject ruby_to_java(ThreadContext context)
 
         @JRubyMethod(name = "to_json", rest = true)
         public IRubyObject ruby_to_json(ThreadContext context, IRubyObject[] args)
-            throws IOException
         {
-            return RubyString.newString(context.runtime, event.toJson());
+            try {
+                return RubyString.newString(context.runtime, event.toJson());
+            } catch (Exception e) {
+                throw new RaiseException(context.runtime, context.runtime.getModule("LogStash").defineOrGetModuleUnder("Json").getClass("GeneratorError"), e.getMessage(), true);
```
talevy: this usage of `defineOrGetModuleUnder` is slightly confusing.

If you end up defining it, then wouldn't the proceeding `getClass` return null? since nothing is using it after that fact, I suppose it is fine

talevy:
in general, what is our stance on testing for our java-core + ruby integration? It looks like you chose to not introduce java tests in favor of doing ruby testing, should we stick to doing both? or is that only important once more of core becomes java

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
colinsurprenant: it is and AFAIK this is the only way to get a reference of an inner module. maybe this could be established earlier on in the Library load so that if this namespace does not exist it will crash on startup and not at exception time.  thanks, good observation.

colinsurprenant:
@tal good point. in this case I did not include specific java tests and probably I should. there is a potential redundancy issue but I think it's fair that some tests overlaps but both are required since they both define different interfaces.

talevy:
I don't view it as redundant. Since we going to be using direct java more and more, I view it as future-proofing :)

colinsurprenant:
@talevy fait enough, I agree.

colinsurprenant:
@talevy I think I covered what we discussed, initializing the error classes at load time and added Java Event#fromJson tests.

suyograo:
heh, tests pass now on Java 7 http://build-eu-00.elastic.co/job/Logstash_PR/2170/

tal:
Hi

talevy:
hi @tal, meet @talevy

suyograo:
We need a @tal vs @talevy PR showdown

talevy:
I'm fine with that, we can assign all my tickets to the real @tal.

colinsurprenant:
lol - sorry to have pulled you in @tal and thanks for stopping by and say Hi :grimacing:

colinsurprenant:
the related PR for the json_lines codec is at https://github.com/logstash-plugins/logstash-codec-json_lines/pull/19

guyboertje:
@colinsurprenant - are you doing the json codec too?

guyboertje:
@colinsurprenant - I am presuming that the Java tests are only done in your IDE or is there some rspec magic that runs the Java tests too?

colinsurprenant:
@guyboertje `gradle build` includes running tests, alternatively `gradle test` will only compile and run tests. If you don't have gradle 2.8+ installed locally you can use the provided wrapper `gradlew`.

colinsurprenant:
@guyboertje good point, I actually forgot about the *plain* json codec and there is a subtlety in there... it has 2 behaviours depending on the input json structure *type*:
- Object/Hash : it generates an `Event` just like the json_lines codec, all good.
- Array : it generates an array of `Event`, one for each json array element. 

I am not sure on the best strategy for this latter case. I could easily change the contract of `from_json` to output `Event[]` and deal with the array case in there. I think that would be the only way to deal with that efficiently otherwise we are stuck at doing the deserialization before, in the codec, in Ruby, then checking the resulting object type.

Hm. I believe changing the contract to 
```
public static Event[] fromJson(String json)
```
or
```
public static List<Event> fromJson(String json)
``` 

would be acceptable? the added benefit would be to also support the array case in json_lines?

Thoughts?

guyboertje:
@colinsurprenant 

> gradle test will only compile and run tests.

The reason for my question about the Java tests is: "Can the reviewer rely on the Jenkins test green tick in this PR to know that ALL the tests have been run and pass?".

I guess at the moment the reviewer must assume that if you are committing the jar file - the Java tests must have passed.

Is this the best we can do or can we make Travis do the `gradle test` and `bundle exec rspec`?

guyboertje:
@colinsurprenant - I saw the json codec with its two operations, that is way I asked. :smile:

I vote we change the `fromJson` contract to `List<Event>` - the jruby wrapper would convert this to a RubyArray of RubyEvent.  In the codecs we can expect an array of events even if it has just one element.

colinsurprenant:
@guyboertje 
- yup, agree with `List<Event>`.
- the build process actually calls `gradlew jar` which does not includes running tests. good catch. we should definitely figure what the best strategy is for having the gradle test task invoked *somewhere* in the build process.
- «I saw the json codec with its two operations» :grimacing:  I appreciate your politeness in gently bringing this to my attention, maybe this has to do with this very British trait ;) but I wouldn't have been offended if you just stated your observation right there in the first place. But that's fine, the whole process actually made me look good :smile:  so thank you :wine_glass:

guyboertje:
@colinsurprenant - you give me more credit than I'm due.  I did not leap as far as thinking that what happens in the json codec will affect the Java Event fromJson contract design. I only meant that there is an optimisation opportunity therein.

But I was also thinking, what happens if the JSON is a well formed array of scalars - but not suited for LS Event generation?

colinsurprenant:
@guyboertje in this case I think the best/only thing to do is to honour the *legacy* behaviour which will be to throw an exception but the new implementation will throw an exception sooner by making sure the json array only contains maps, as for the *legacy* implementation would crash later when trying to initialize a new Event with a scalar parameter value. 

I just finished the refactor to handle the json *array of map* case and I am a bit unhappy because it causes non negligible overhead in creating collections and converting these to Ruby and then the codec has to iterate the result. I am trying to see how this could be optimized...

colinsurprenant:
@guyboertje I think I'll play the dynamic game and have `from_json` Ruby interface either return a single `Event` or `Array<Event>` and it's up to the method consumer to either *know* the passed json is a map yielding a single `Event` object otherwise testing on the returned object type `result.is_a?(Array)` or force-normalizing into an array `Array(result)`. I will try to see if I can improve by doing this at the cost of a *weirder* interface. Suggestions welcome :stuck_out_tongue:

colinsurprenant:
I finally decided on a middle ground, `Event#from_json` will always return an Array which makes for a more coherent interface but the implementation is using Java arrays instead of collections and this yields acceptable performance. 

Note that this last commit introduces a bit more changes because I decided to upgrade Jackson which was really old and it is using a new namespace in v2

guyboertje:
:+1: on returning an array.

colinsurprenant:
@talevy  ^^ added test for partial invalid json array

talevy:
sweet. any other todos? things look good to me

colinsurprenant:
no, all good on my side. wanna merge. :pray:

talevy:
LGTM

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -153,10 +153,40 @@ public boolean includes(String reference) {
...
+        Event[] result;
+
+        if (json == null || json.trim().isEmpty()) {
+            return new Event[]{ new Event() };
+        }
+
+        Object o = mapper.readValue(json, Object.class);
+        // we currently only support Map or Array json objects
+        if (o instanceof Map) {
+            result = new Event[]{ new Event((Map)o) };
```
guyboertje: nice :+1:

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -153,10 +153,40 @@ public boolean includes(String reference) {
...
+                    throw new IOException("incompatible inner json array object type=" + e.getClass().getName() + " , only hash map is suppoted");
+                }
+                result[i++] = new Event((Map)e);
+            }
+        } else {
+            throw new IOException("incompatible json object type=" + o.getClass().getName() + " , only hash map or arrays are suppoted");
+        }
+        return result;
+    }
+
```
guyboertje: nice to catch either "[1,2,3,4]" or "\"fubar\"", both valid JSON but useless to us.

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -251,9 +268,38 @@ public IRubyObject ruby_to_java(ThreadContext context)
...
+                events = Event.fromJson(value.asJavaString());
+            } catch (Exception e) {
+                throw new RaiseException(context.runtime, PARSER_ERROR, e.getMessage(), true);
+            }
+
+            RubyArray result = RubyArray.newArray(context.runtime, events.length);
+
+            if (events.length == 1) {
+                // micro optimization for the 1 event more common use-case.
+                result.set(0, RubyEvent.newRubyEvent(context.runtime, events[0]));
```
guyboertje: nice shortcut

logstash-core-event-java/src/main/java/com/logstash/Event.java
colinsurprenant: yup, also added good test harness around these cases.

guyboertje:
LGTM

colinsurprenant:
great! squashing and merging.

colinsurprenant:
merged into master and 2.x

PR #4639, opened 2016-02-05
remove the secondary pipeline

ph:


ph:
This should remove the manticore and timeouts errors that you have experienced.

purbon:
@ph why do we remove it for now? could we just add a flag and output everything for now to ```output/null```? then as soon as we introduce the settings we can enable proper shipping. what do you think?

ph:
I prefer to remove until we have settings in place, things will certainly changes with that in place.

purbon:
LGTM

PR #4641, opened 2016-02-05
Metrics API resources definition

purbon:
This make all designed resources for the metrics api available, this follows the initial sketches created in #4446.

As discussed in #4446, this adds:

* /
* /_node/hot_threads
* /_node/stats/
* /_node/stats/events
* /_node/stats/jvm
* /_stats/jvm

you can find all resource definitions within the app directory of the api sinatra app.

resources


logstash-core/lib/logstash/api/lib/app/commands/stats/events_command.rb
```diff
@@ -8,9 +8,9 @@ def run
     #need to be tailored to the right metrics for this command.
     stats = service.get(:events_stats)
     {
-      :in => stats[:base][:events_in].value,
-      :out => 0,
-      :dropped => stats[:base][:events_filtered].value
+      :in => stats[:stats][:events][:in].value,
+      :out => stats[:stats][:events][:out].value,
+      :dropped => stats[:stats][:events][:filtered].value
```
ph: Lets agree on a definition for dropped here.

For me `dropped` events are events that we have removed from the pipeline, either a plugin decided to remove the events from the pipeline or a event became poisonous, or an event raised an exception.

ph: `Filtered` is a different beast, since an event IN the fitlers queue doesn't necessary mean and event out. I have added the filtered counter.

Also this counter will also be updated by the flush ticks.

logstash-core/lib/logstash/agent.rb
```diff
@@ -106,6 +106,7 @@ def stop_background_services
   def configure_metric
     if collect_metric?
       @logger.debug("Agent: Configuring metric collection")
+      LogStash::Instrument::Collector.instance.agent = self
```
ph: Do the collector really need to know about the agent? Would it be better to start the webserver and give it the reference to the current agent?

logstash-core/lib/logstash/api/lib/app/commands/stats/hotthreads_command.rb
```diff
@@ -0,0 +1,96 @@
+# encoding: utf-8
+require "app/command"
+require 'monitoring'
+require "socket"
+
+class LogStash::Api::HotThreadsCommand < LogStash::Api::Command
+
+  def run(options={})
+    hash      = JRMonitor.threads.generate
```
ph: No sure we need that many spaces.

logstash-core/lib/logstash/api/lib/app/commands/stats/hotthreads_command.rb
```diff
@@ -0,0 +1,96 @@
...
+    def initialize(dump, options={})
+      @dump    = dump
+      @options = options
+      @top_count = options.fetch(:threads, 3)
+      @ignore    = options.fetch(:ignore_idle_threads, true)
+    end
+
+    def to_s
+      hash = to_hash
+      report = "::: {#{hash[:hostname]}} \n Hot threads at #{hash[:time]}, busiestThreads=#{top_count}:\n"
```
ph: Can we move the string from the code to the locales.yml?

logstash-core/lib/logstash/api/lib/app/commands/stats/hotthreads_command.rb
```diff
@@ -0,0 +1,96 @@
...
+    SKIPPED_THREADS = [ "Finalizer", "Reference Handler", "Signal Dispatcher" ].freeze
+
+    attr_reader :top_count, :ignore, :dump
+
+    def initialize(dump, options={})
+      @dump    = dump
+      @options = options
+      @top_count = options.fetch(:threads, 3)
+      @ignore    = options.fetch(:ignore_idle_threads, true)
+    end
```
ph: No sure about align of `=`,  I personally find them to be hard to maintain with changes.

logstash-core/lib/logstash/api/lib/app/commands/stats/hotthreads_command.rb
```diff
@@ -0,0 +1,96 @@
...
+        if ignore
+          next if SKIPPED_THREADS.include?(thread_name)
+          next if thread_name.match(/Ruby-\d+-JIT-\d+/)
+        end
+        block.call(thread_name, _hash)
+        i += 1
+      end
+    end
+
+    def build_report(hash)
```
ph: If this part of the code really called? Its private and I don't see an internal call to it.

logstash-core/lib/logstash/api/lib/app/commands/stats/hotthreads_command.rb
```diff
@@ -0,0 +1,96 @@
...
+      end
+      report
+    end
+
+    def hostname
+      Socket.gethostname
+    end
+
+    def cpu_time(hash)
+      hash["cpu.time"] / 1000
```
ph: Move 100 to a constant?

logstash-core/lib/logstash/api/lib/app/commands/stats/memory_command.rb
```diff
@@ -0,0 +1,21 @@
...
+      :heap => memory["heap"],
+      :non_heap => memory["non_heap"],
+      :pools => memory["pools"].inject({}) do |acc, (type, hash)|
+                  acc[type] = hash
+                  acc
+               end
+    }
+  rescue
+    {} # Something happen, so we just return an empty hash.
+  end
```
ph: This rescue block is rescuing everything, what are the actuals exceptions you are thinking that can happen here?

logstash-core/lib/logstash/agent.rb
purbon: I don't know how can I pass an instance to  the Sinatra App by passing it
to the webserver (aka puma agent here). Only way I know to pass it is
through code, like the collector way, we could add another way. Please let
me know if you know how to pass ref thought webserver.

logstash-core/lib/logstash/api/lib/app/commands/system/basicinfo_command.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  attr_reader :agent
+
+  def run
+    @agent = service.agent
+    {
+      "version"   => LOGSTASH_VERSION,
+      "hostname" => @agent.node_name,
+    }
+  end
```
ph: There is some conflicting codes with the reference to `agent`, first we declare an `attr_reader :agent`, the value is set in the `#run` method but we never use it. We always use the *instance variable*.

Could we make the `agent`  handling part of the `LogStash::Api::Command` base class?
Could this be useful to other command?

logstash-core/lib/logstash/api/lib/app/commands/system/basicinfo_command.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  private
+
+  def pipelines
```
ph: Confusion, I think this code is never called, its marked as private?

logstash-core/lib/logstash/api/lib/app/modules/node.rb
```diff
@@ -0,0 +1,24 @@
...
+require "app"
+
+module LogStash::Api
+  class Node < BaseApp
+
+    helpers AppHelpers
+
+    # return hot threads information
+    get "/hot_threads" do
+      top_threads_count = params["threads"] || 3
```
ph: Move to a constant?

ph: with a default values hash?

logstash-core/lib/logstash/api/lib/app/service.rb
```diff
@@ -30,9 +34,10 @@ def update(snapshot)
   def get(key)
     metric_store = @snapshot.metric_store
     if key == :jvm_memory_stats
-      metric_store.get(:root, :jvm, :memory)
+      data = metric_store.get_with_path("jvm/memory")[:jvm][:memory]
+      LogStash::Json.dump(data)
     else
-      { :base => metric_store.get(:root, :base) }
+      metric_store.get_with_path("stats/events")
     end
```
ph: The return value of this method is mixed depending on the key we will return a json string or a ruby hash?
I think the serialization should be handled in the representation layer in Sinatra?

logstash-core/lib/logstash/api/lib/app/commands/stats/hotthreads_command.rb
ph: Also we can move the default values in constant?

ph:
First run of review in :)

logstash-core/lib/logstash/api/lib/app/commands/stats/hotthreads_command.rb
purbon: sure, makes sense to me.

logstash-core/lib/logstash/agent.rb
purbon: We forward this change for later, we will add issue and comment to properly refactor the runner to pass all agent informations.

logstash-core/lib/logstash/api/lib/app/commands/stats/memory_command.rb
```diff
@@ -0,0 +1,27 @@
...
+      :non_heap_committed_in_bytes => memory["non_heap"]["committed_in_bytes"],
+      :pools => memory["pools"].inject({}) do |acc, (type, hash)|
+          hash.delete("committed_in_bytes")
+          acc[type] = hash
+          acc
+    end
+    }
+  rescue
+    {} # Something happen, so we just return an empty hash.
+  end
```
ph: What could go wrong here?
We just hide the error and we don't do any logging?

logstash-core/lib/logstash/api/lib/app/service.rb
ph: ^ This is still a problem, no?

logstash-core/lib/logstash/api/lib/app/commands/stats/memory_command.rb
purbon: this should be removed, if something goes wrong here should be catched by the sinatra error handler.

purbon:
@ph I think I made all changes required, please take another look and give me your green light to merge :+1: if possible! good work man!

purbon:
@suyograo added last commits here, can you please review also data formats are the ones you expect, just to be sure.

This defines the list of resources we discussed in #4446:

* /
* /_node/hot_threads
* /_node/stats/
* /_node/stats/events
* /_node/stats/jvm
* /_stats/jvm

thank you for your time.

ph:
Code LGTM.

Issue #4306, opened 2015-12-05
Event#clone do not copy @metadata


wiibaa:
Question triggered by report in https://github.com/logstash-plugins/logstash-filter-clone/issues/5

What is the expected behaviour of the _#clone_ method regarding @metadata ?
User expectaction is that an exact copy is done data + metadata, but it is not currently the case.

@colinsurprenant as you are in an heavy review of the event class, could you please comment on this ?

purbon:
There is also another collateral effect related to that, as you can see from https://github.com/logstash-plugins/logstash-filter-multiline/blob/master/lib/logstash/filters/multiline.rb#L233-L234, override does not override metadata. I guess we should also track this as the expectations for this methods are confussing.

wiibaa:
@purbon @colinsurprenant I'm working on a workaround for this in the clone filter but I'm stuck deciding the correct way of doing a shallow copy, otherwise the events would be sharing their @metadata field. 
`clone["@metadata"] = Marshal.load(Marshal.dump(event["@metadata"]))` looks ugly but seems rather common. Thoughts ?

jordansissel:
@wiibaa seems like we're missing a nice `clone` method that an the Event could expose and solve this. Thoughts?

original-brownbear:
fixed in #4640

PR #4642, opened 2016-02-06
Clone metadata when Event#clone is called

### LINKED ISSUE #4632: master build failure on java 7
there is a build failure in #4631 which does not seems to be related to the PR. In fact, I just finished checking on bare master and it also reproduces.

The failure happens only in **Java 7** and can be reproduced with 
```sh
rake bootstrap
rake test:install-core
SPEC_OPTS="--seed 61716" rake test:core
```

the error is

```
Failures:

  1) LogStash::Runner argument parsing with no arguments should show help
     Failure/Error: subject.run(args)
       (#<Cabin::Channel:0x57ec8527>).warn("You may be interested in the '--configtest' flag which you can use to validate\nlogstash's configuration before you choose to restart a running system.")
           expected: 1 time with any arguments
           received: 2 times with arguments: ("You may be interested in the '--configtest' flag which you can use to validate\nlogstash's configuration before you choose to restart a running system.")
     # ./logstash-core/lib/logstash/runner.rb:164:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67:in `run'
     # ./logstash-core/spec/logstash/runner_spec.rb:40:in `(root)'

Finished in 13.06 seconds (files took 3.31 seconds to load)
1005 examples, 1 failure

Failed examples:

rspec ./logstash-core/spec/logstash/runner_spec.rb:36 # LogStash::Runner argument parsing with no arguments should show help

Randomized with seed 61716
```

### LINKED ISSUE #4640: `LogStash::Event#clone` should also clone the `@metadata`

suyograo:
This partially fixes #4640, only for Ruby implementation. We need to
make a similar change to Java implementation

### DIFF
--- a/logstash-core-event/lib/logstash/event.rb
+++ b/logstash-core-event/lib/logstash/event.rb
@@ -93,19 +93,8 @@ class LogStash::Event
 
   # Create a deep-ish copy of this event.
   def clone
-    copy = {}
-    @data.each do |k,v|
-      # TODO(sissel): Recurse if this is a hash/array?
-      copy[k] = begin v.clone rescue v end
-    end
-
-    cloned = self.class.new(copy)
-
-    # Clone metadata too
-    @metadata.each do |k,v|
-      cloned["@metadata"][k] = begin v.clone rescue v end
-    end
-    cloned
+    copy = LogStash::Util.deep_clone(@data)
+    self.class.new(copy.merge({ "@metadata" => LogStash::Util.deep_clone(@metadata.clone) }))
   end
 
   def to_s
--- a/logstash-core-event/lib/logstash/event.rb
+++ b/logstash-core-event/lib/logstash/event.rb
@@ -99,7 +99,13 @@ class LogStash::Event
       copy[k] = begin v.clone rescue v end
     end
 
-    self.class.new(copy)
+    cloned = self.class.new(copy)
+
+    # Clone metadata too
+    @metadata.each do |k,v|
+      cloned["@metadata"][k] = begin v.clone rescue v end
+    end
+    cloned
   end
 
   def to_s
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -669,7 +669,7 @@ describe LogStash::Event do
     it "should clone metadata fields with multiple keys" do
       cloned = event5.clone
       expect(cloned.get(fieldref)).to eq("pants")
-      expect(cloned.get("[@metadata][smarty]").to eq("pants2")
+      expect(cloned.get("[@metadata][smarty]")).to eq("pants2")
       expect(cloned.to_hash_with_metadata).to include("@metadata")
     end
   end
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -630,5 +630,48 @@ describe LogStash::Event do
       expect(event.get("[a][0][b]")).to eq(98)
     end
   end
+
+  describe "#clone" do
+    let(:fieldref) { "[@metadata][fancy]" }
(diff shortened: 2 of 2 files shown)


logstash-core-event/lib/logstash/event.rb
```diff
@@ -98,7 +98,13 @@ def clone
...
     end
 
-    self.class.new(copy)
+    cloned = self.class.new(copy)
+    
+    # Clone metadata too
+    @metadata.each do |k,v|
+      cloned["@metadata"][k] = begin v.clone rescue v end
+    end
+    cloned
```
ph: I took a bit of time to go through the code I don't think we need the loop in that case.
I think this code should be enough in most case.

```
self.class.new(copy.merge({ "@metadata" => LogStash::Util.deep_clone(@metadata.clone) }))
```

It really handle recursive cloning of the values, we might want to use the same behavior for the `@data`

ph: Using clone on an hash is not enough if it contains strings..

```
irb(main):001:0> m = "Hello"
=> "Hello"
irb(main):002:0> k = { "message" => m }
=> {"message"=>"Hello"}
irb(main):003:0> kcloned = k.clone
=> {"message"=>"Hello"}
irb(main):004:0> m << " World"
=> "Hello World"
irb(main):005:0> kcloned
=> {"message"=>"Hello World"}
irb(main):006:0> k
=> {"message"=>"Hello World"}
```

wiibaa: @ph Sorry to barge in, but I'm interested in this for the clone filter, to my understanding, either you need to loop in the hash (and nested hash etc)  or you can use a Marshal "hack" 
`copied_data = Marshal.load(Marshal.dump(complex_data))`
Do you think it is usable performance wise ?
Reference => http://thingsaaronmade.com/blog/ruby-shallow-copy-surprise.html

ph: @wiibaa you are welcome anytime :)

I believe we should provide a `#clone` method to the user and this method should sanely does what is supposed to do.
Clone the object and make sure we drop any references, the Marshal dance in some cases is the only way to make it work. But has the side effect to be really slow in a tight event loop this is not something we want. (LogStash::Util.deep_clone is using marshal)

The problem is usually show if you do in place modification, mutability is a silly world :)

A quick search show 2 plugins using an explicit `Event#clone` the split filter and the dns filter.

ph: Also to note `deep_clone` will use marshal as the last resort.
```ruby
def self.deep_clone(o)
    case o
    when Hash
      o.inject({}) {|h, (k,v)| h[k] = deep_clone(v); h }
    when Array
      o.map {|v| deep_clone(v) }
    when Fixnum, Symbol, IO, TrueClass, FalseClass, NilClass
      o
    when LogStash::Codecs::Base
      o.clone
    when String
      o.clone #need to keep internal state e.g. frozen
    else
      Marshal.load(Marshal.dump(o))
    end
end
```

ph:
This PR fix https://github.com/elastic/logstash/issues/4306

ph:
Lets target this for 2.3.2?

logstash-core-event/lib/logstash/event.rb
```diff
@@ -98,7 +98,13 @@ def clone
       copy[k] = begin v.clone rescue v end
```
guyboertje: We should really use `LogStash::Util.deep_clone(@data)` because otherwise we get `TypeError: can't clone Fixnum` type errors.

This applies to `@metadata` too.

guyboertje:
@suyograo - please adapt the event clone method to use `LogStash::Util.deep_clone` for the data and metadata.

ph:
related #5567

suyograo:
Closed in favor of https://github.com/elastic/logstash/pull/5656. Doing this first for Ruby Event in 2.x. Will port it to Java Event next

PR #4643, opened 2016-02-08
support live reloading of multiple pipelines

jsvd:
NOTE: this is a backport from master (pr #4520)

* reload config from files
* add reload config flag and allow reload through sighup
* move signal handling to runner
* add "-r" flag to detect changes in config and reload
* make SIGHUP reload config file
* make agent execute either loop to detect changes or sleep forever
* sigint/sigterm shut down the agent.execute task
* stop agent if pipeline terminates
* Better shutdown logic for controller and pipeline
* support multiple pipeline reloading and other cleanups
* serialize upgrade_state and running_pipelines?
* remove useless @thread from pipeline
* remove agent registry and cleanup
* reduce reload_interval from 5 to 3
* remove useless MissingAgentError class
* fix exit codes for reload_state failures
* explain why thread var exists in agent.rb
* properly pass backtrace exceptions to oops
* remove is_config_test from loader
* better documentation on agent/runner/pipeline methods

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -3,12 +3,16 @@ require "clamp" # gem 'clamp'
 require "logstash/environment"
 require "logstash/errors"
 require "logstash/config/cpu_core_strategy"
+require "stud/trap"
+require "logstash/config/loader"
 require "uri"
 require "net/http"
 require "logstash/pipeline"
-LogStash::Environment.load_locale!
 
 class LogStash::Agent < Clamp::Command
+
+  attr_reader :pipelines
+
   DEFAULT_INPUT = "input { stdin { type => stdin } }"
   DEFAULT_OUTPUT = "output { stdout { codec => rubydebug } }"
 
@@ -22,18 +26,18 @@ class LogStash::Agent < Clamp::Command
     :default => "", :attribute_name => :config_string
 
   option ["-w", "--pipeline-workers"], "COUNT",
-         I18n.t("logstash.runner.flag.pipeline-workers"),
+         I18n.t("logstash.agent.flag.pipeline-workers"),
          :attribute_name => :pipeline_workers,
          :default => LogStash::Pipeline::DEFAULT_SETTINGS[:default_pipeline_workers]
 
 
   option ["-b", "--pipeline-batch-size"], "SIZE",
-         I18n.t("logstash.runner.flag.pipeline-batch-size"),
+         I18n.t("logstash.agent.flag.pipeline-batch-size"),
          :attribute_name => :pipeline_batch_size,
          :default => LogStash::Pipeline::DEFAULT_SETTINGS[:pipeline_batch_size]
 
   option ["-u", "--pipeline-batch-delay"], "DELAY_IN_MS",
-         I18n.t("logstash.runner.flag.pipeline-batch-delay"),
+         I18n.t("logstash.agent.flag.pipeline-batch-delay"),
          :attribute_name => :pipeline_batch_delay,
          :default => LogStash::Pipeline::DEFAULT_SETTINGS[:pipeline_batch_delay]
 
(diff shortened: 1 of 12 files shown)


jsvd:
@andrewvc review welcome on this port to 2.x branch

suyograo:
@jsvd rebase?

andrewvc:
@jsvd this now has merge conflicts (apologies for the late delay) can you rebase? One of them is with the patched 'oops' handling I added in master. We shouldn't be formatting the backtrace with I18N but rather passing it as a key to Lumberjack

logstash-core/lib/logstash/agent.rb
```diff
@@ -3,12 +3,16 @@
 require "logstash/environment"
 require "logstash/errors"
 require "logstash/config/cpu_core_strategy"
+require "stud/trap"
+require "logstash/config/loader"
 require "uri"
 require "net/http"
 require "logstash/pipeline"
-LogStash::Environment.load_locale!
```
andrewvc: Why was this removed?

jsvd: I didn't see a need for it (didn't break anything). we still do `load_locale!` in 3 places:

logstash-core/lib/logstash/config/mixin.rb:LogStash::Environment.load_locale!
logstash-core/lib/logstash/environment.rb:    def load_locale!
logstash-core/lib/logstash/runner.rb:LogStash::Environment.load_locale!

Which should cover the paths to loading agent.rb

logstash-core/lib/logstash/agent.rb
```diff
@@ -320,63 +296,195 @@ def configure_plugin_paths(paths)
...
+  end
+
+  def shutdown_pipelines
+    @pipelines.each { |id, _| stop_pipeline(id) }
+  end
+
+  def stop_pipeline(id)
+    pipeline = @pipelines[id]
+    return unless pipeline
+    @logger.warn("stopping pipeline", :id => id)
```
andrewvc: Why is this logger.warn? Shouldn't this be at the info level?

logstash-core/lib/logstash/agent.rb
```diff
@@ -320,63 +296,195 @@ def configure_plugin_paths(paths)
...
+  def reload_pipeline!(id)
+    old_pipeline = @pipelines[id]
+    new_pipeline = create_pipeline(old_pipeline.original_settings)
+    return if new_pipeline.nil?
+
+    if old_pipeline.config_str == new_pipeline.config_str
+      @logger.debug("no configuration change for pipeline",
+                    :pipeline => id, :config => old_pipeline.config_str)
+    else
+      @logger.warn("fetched new config for pipeline. upgrading..",
```
andrewvc: Shouldn't this be logged at info, not warn?

logstash-core/lib/logstash/agent.rb
jsvd: True... By default we set the logger to only show >= warn and I thought this was an important enough message to show in normal usage.

andrewvc: I think in that case we should do a logger.info + a logger.terminal. I don't think it should appear on-disk IMHO.

The real problem is that info should be safe to turn on by default IMHO. warn should be used for things that demand operator attention.

andrewvc: What do you think @jordansissel ? How does this integrate with your long term plans for log level revision in LS?

jsvd: an alternative for now is to just do a @logger.log, which will bypass the level checking, iirc.
I don't see a reason why this message shouldn't be in the disk logs

logstash-core/locales/en.yml
```diff
@@ -5,7 +5,7 @@
 en:
   oops: |-
     The error reported is: 
-      %{error}
```
andrewvc: This will have a merge conflict now.

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -1,39 +1,167 @@
...
 
-      context "unix" do
-        let(:path) { './test.conf' }
-        it 'works with relative path' do
-          subject.load_config(path)
+  before :each do
+    [:info, :warn, :error, :fatal, :debug].each do |level|
+      allow(logger).to receive(level)
+    end
+    [:info?, :warn?, :error?, :fatal?, :debug?].each do |level|
```
andrewvc: Maybe just have one array with both ? and non-? things?

logstash-core/lib/logstash/agent.rb
andrewvc: @jsvd I'm fine with `@logger.log`

logstash-core/lib/logstash/agent.rb
andrewvc: :+1:

andrewvc:
Once the logger stuff is changed to @logger.log I'm LGTM here.

jsvd:
closed by 0099537

PR #4652, opened 2016-02-09
Metrics API resources definition

purbon:
follows the initial sketches created in #4446.

As discussed in this adds:

/
/_node/hot_threads
/_node/stats/
/_node/stats/events
/_node/stats/jvm
/_stats/jvm

* introduces also small refactors and cleanup necessary to improve the webapi code quality.

* adds also a way to handle references from the webapi to the internals of logstash by passing the agent references to the collector, this is not perfect, but for now it solve the communication situation until further refactor.

Supersede #4641 as this one is directed to the right branch after we made all available again under the feature/metrics branch.


purbon:
Merging this as spoke https://github.com/elastic/logstash/pull/4641#issuecomment-181542918

PR #4655, opened 2016-02-09
Fix web api HTTP to be 9600 as Default

### LINKED ISSUE #4648: http server default port
Change the default port of the http server, the current default is 9292. This port conflict with ES's range.
@suyograo  suggested we move to 9600, a quick google search doesnt show any popular application using that port.

Also we should be able to change it when starting logstash with a `cli` command, 
`--http-port` this can be moved to the settings file later to `http.port` setting.

purbon:
Fixes #4648

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -34,6 +34,8 @@ class LogStash::Agent
     @pipelines = {}
     @started_at = Time.now
     @node_name = params[:node_name] || Socket.gethostname
+    @web_api_http_port = params[:web_api_http_port]
+
     @config_loader = LogStash::Config::Loader.new(@logger)
     @reload_interval = params[:reload_interval] || 3 # seconds
     @upgrade_mutex = Mutex.new
@@ -108,7 +110,7 @@ class LogStash::Agent
 
   private
   def start_webserver
-    options = { :debug => debug }
+    options = { :debug => debug, :http_port => @web_api_http_port }
     @webserver = LogStash::WebServer.new(@logger, options)
     Thread.new(@webserver) do |webserver|
       LogStash::Util.set_thread_name("Api Webserver")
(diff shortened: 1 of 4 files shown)


purbon:
@ph @jsvd thoughts and reviews more than welcome.

logstash-core/lib/logstash/webserver.rb
```diff
@@ -17,11 +17,14 @@ class WebServer
...
       @logger      = logger
       @options     = {}
-      @cli_options = options.merge({ :rackup => ::File.join(::File.dirname(__FILE__), "api", "init.ru")  })
+      @cli_options = options.merge({ :rackup => ::File.join(::File.dirname(__FILE__), "api", "init.ru"), 
+                                     :binds => ["tcp://0.0.0.0:#{options[:http_port]}"] })
       @status      = nil
 
       parse_options
 
+      puts @options
```
ph: Debugging?

logstash-core/lib/logstash/agent.rb
```diff
@@ -34,6 +34,8 @@ def initialize(params)
     @pipelines = {}
     @started_at = Time.now
     @node_name = params[:node_name] || Socket.gethostname
+    @web_api_http_port = params[:web_api_http_port] || 9600
```
ph: I think the default value should be handled inside the webservice class, It's not a concern of the agent to setup the default values for the port. Also moving it there align with our discussion from yesterday concerning a runner class.

logstash-core/lib/logstash/webserver.rb
purbon: crap yes.

logstash-core/lib/logstash/agent.rb
purbon: ++

purbon:
@ph comments amended, thanks a lot for your time.

purbon:
please jenkins, test this.

ph:
Do you think we should add a test for this? Maybe a unit test against the webserver?

purbon:
we should for sure, but unit test this would be a bit complicated as it
will need to mock the web app somehow.... can we open an issue to add this
in next iteration? will need to think about the right way to test it.

- purbon

ph:
+1 create an issue1 lgtm

PR #4658, opened 2016-02-09
Add a pretty print option to the metrics web api

### LINKED ISSUE #4645: Add a JSON preatty print option to the web api
It would be nice to have an option like:

* pretty: When appending ?pretty to any request made, the JSON returned will be pretty formatted.

purbon:
Fixes #4645

### DIFF
--- a/logstash-core/spec/logstash/json_spec.rb
+++ b/logstash-core/spec/logstash/json_spec.rb
@@ -56,6 +56,20 @@ describe "LogStash::Json" do
         expect(LogStash::Json.dump(array)).to eql(json_array)
       end
 
+      context "pretty print" do
+
+        let(:hash) { { "foo" => "bar", :zoo => 2 } }
+
+        it "should serialize with pretty print" do
+          pprint_json = LogStash::Json.dump(hash, :pretty => true)
+          expect(pprint_json).to include("\n")
+        end
+
+        it "should by default do no pretty print" do
+          pprint_json = LogStash::Json.dump(hash)
+          expect(pprint_json).not_to include("\n")
+        end
+      end
     end
 
   else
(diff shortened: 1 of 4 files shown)


ph:
Can we have a small test for the `Logstash::Json` class?

purbon:
@ph not sure how to differentiate between preatty and not pretty, any criteria you might think of?

ph:
@purbon The json format with the pretty option should have line data formatting in it ("\n" ...)
I think is important to test it since we wrap the library in a common accessor, if we ever change jrjackson for something else we should make sure it support a pretty output.

purbon:
ok, will add a test looking for CR characters. :smile:

purbon:
Test added. :+1:

logstash-core/spec/logstash/json_spec.rb
```diff
@@ -56,6 +56,20 @@
         expect(LogStash::Json.dump(array)).to eql(json_array)
       end
 
+      context "pretty print" do
+
+        let(:hash) { { "foo" => "bar", :zoo => 2 } }
+
+        it "should serialize with pretty print" do
+          pprint_json = LogStash::Json.dump(hash, :pretty => true)
+          expect(pprint_json).to include("\n")
```
ph: I think I would make it stricter with comparing the results with the actual generated string, But I guess this work too?
```
irb(main):001:0> JrJackson::Base.generate({ "foo" => "bar", :zoo => 2 }, :pretty => true)
=> "{\n  \"foo\" : \"bar\",\n  \"zoo\" : 2\n}"
irb(main):002:0> JrJackson::Base.generate({ "foo" => "bar", :zoo => 2 }, :pretty => false)
=> "{\"foo\":\"bar\",\"zoo\":2}"

ph:
Small, comment for the tests but this lgtm.

logstash-core/spec/logstash/json_spec.rb
purbon: I'm not sure if doing it strictly could actually leak to some specific implementation details, should all pretty print be the same? or just include CR? I mean for example in the indentation. For now I think I will leave it like this and we can always improve it later on.

PR #4661, opened 2016-02-09
Use the right keys in locale to describe CLI flags

suyograo:
Fixes `-w, --pipeline-workers COUNT  translation missing: en.logstash.runner.flag.pipeline-workers (default: 8)`


andrewvc:
LGTM

PR #4653, opened 2016-02-09
Add metrics collection and api for logstash-core and plugins

purbon:
This PR add initial support for metrics collection within the core logstash pipeline, including plugins being used with it and a REST api to query them for monitoring purposes.

The webapi defines an initial collection of resources as discussed in #4446  that are,

* /
* /_node/hot_threads
* /_node/stats/
* /_node/stats/events
* /_node/stats/jvm
* /_stats/jvm

For more context on the metrics collection and API efforts you can check https://github.com/elastic/logstash/issues/3908.

After the merge of #4655, the default port is 9600 as discussed in #4648. You can operate this through a CLI flag.

After the merge of #4658, now all resources returning json can do pretty print if they have the key pretty in the url. See #4645 for more details.

/cc @ph

PR #4662, opened 2016-02-09
Set of small fixes proposed with merge to master review

purbon:
From https://github.com/elastic/logstash/pull/4653#issuecomment-181961527:

> about idle threads, we are actually ignoring (at my understanding) the idle threads ES id ignoring, do you propose to ignore all waiting threads?
about versions, yeah we can add both's this should not be complicated to archive.
about 404? I agree, we should return empty body with status 404.
thanks a lot for your time.

Basically adds this PR:

* default content for 404 errors
* bring back the threads param to select the number of threads to be shown in the web api.
* default the hot threads number to 3
* Ignore all threads that fall under one of the next criteria, an internal JVM thread, an internal JRuby thread pool thread, or an internal JRuby JIT thread. This will obviously be shown if the right flag is passed.

### DIFF
--- a/logstash-core/lib/logstash/api/lib/app.rb
+++ b/logstash-core/lib/logstash/api/lib/app.rb
@@ -27,7 +27,9 @@ module LogStash::Api
 
     not_found do
       status 404
-      respond_with({})
+      as   = params.has_key?("human") ? :string : :json
+      text = as == :string ? "" : {}
+      respond_with(text, :as => as)
     end
 
     error do
--- a/logstash-core/lib/logstash/api/lib/app.rb
+++ b/logstash-core/lib/logstash/api/lib/app.rb
@@ -25,6 +25,11 @@ module LogStash::Api
       @factory = CommandFactory.new(settings.service)
     end
 
+    not_found do
+      status 404
+      respond_with({})
+    end
+
     error do
       logger.error(env['sinatra.error'].message, :url => request.url, :ip => request.ip, :params => request.params)
     end
--- a/logstash-core/lib/logstash/api/lib/app/commands/system/basicinfo_command.rb
+++ b/logstash-core/lib/logstash/api/lib/app/commands/system/basicinfo_command.rb
@@ -8,7 +8,7 @@ class LogStash::Api::SystemBasicInfoCommand < LogStash::Api::Command
     {
       "hostname" => hostname,
       "version" => {
-        "number" => LOGSTASH_VERSION,
+        "number" => LOGSTASH_VERSION
       }
     }
   end
--- a/logstash-core/lib/logstash/api/lib/app/commands/system/basicinfo_command.rb
+++ b/logstash-core/lib/logstash/api/lib/app/commands/system/basicinfo_command.rb
@@ -8,7 +8,7 @@ class LogStash::Api::SystemBasicInfoCommand < LogStash::Api::Command
     {
       "hostname" => hostname,
       "version" => {
-        "number" => LOGSTASH_VERSION
+        "number" => LOGSTASH_VERSION,
       }
     }
   end
--- a/logstash-core/lib/logstash/webserver.rb
+++ b/logstash-core/lib/logstash/webserver.rb
@@ -62,7 +62,7 @@ module LogStash
         @runner.stop_blocked
       else
         @runner.stop
-      end rescue ""
+      end rescue nil
 
       @status = :stop
       log "=== puma shutdown: #{Time.now} ==="
--- a/logstash-core/lib/logstash/webserver.rb
+++ b/logstash-core/lib/logstash/webserver.rb
@@ -62,7 +62,8 @@ module LogStash
         @runner.stop_blocked
       else
         @runner.stop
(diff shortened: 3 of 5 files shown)


ph:
LGTM, but we will need a round of adding tests.

purbon:
@ph for sure, this will be my priority next.

logstash-core/lib/logstash/api/lib/app/commands/system/basicinfo_command.rb
```diff
@@ -8,8 +8,20 @@ def run
     {
       "hostname" => hostname,
       "version" => {
-        "number" => LOGSTASH_VERSION
+        "number" => LOGSTASH_VERSION,
+        "logstash-core" => {
+          "version" => {
+            "number" => fetch_logstash_core_version
```
suyograo: Hey, thinking about this more, lets keep this simple for now. LS Core and LS Event Versions are the same. If and when it diverges, we can display it. Sorry I'd like this to be super easy for users to understand. Having multiple versions just confuses users.

@purbon Lets go back to 

```
{
        "hostname" => hostname,
        "version" => {
           "number" => "2.2.0"
           "build_hash => "72cd1f1a3eee09505e036106146dc1949dc5dc87"
    }
}

```

purbon: make sense, no problem. reverting this change.

purbon:
related with: #4667

purbon:
@suyograo @ph @jsvd about ignoring idle theads and what does this mens, I just checked the ES code base and what they do is to ignore this threads.

```

    private static boolean isIdleThread(ThreadInfo threadInfo) {
        String threadName = threadInfo.getThreadName();

        // NOTE: these are likely JVM dependent
        if (threadName.equals("Signal Dispatcher") ||
            threadName.equals("Finalizer") ||
            threadName.equals("Reference Handler")) {
            return true;
        }

        for (StackTraceElement frame : threadInfo.getStackTrace()) {
            String className = frame.getClassName();
            String methodName = frame.getMethodName();
            if (className.equals("java.util.concurrent.ThreadPoolExecutor") &&
                methodName.equals("getTask")) {
                return true;
            }
            if (className.equals("sun.nio.ch.SelectorImpl") &&
                methodName.equals("select")) {
                return true;
            }
            if (className.equals("org.elasticsearch.threadpool.ThreadPool$EstimatedTimeThread") &&
                methodName.equals("run")) {
                return true;
            }
            if (className.equals("org.elasticsearch.indices.ttl.IndicesTTLService$Notifier") &&
                methodName.equals("await")) {
                return true;
            }
            if (className.equals("java.util.concurrent.LinkedTransferQueue") &&
                methodName.equals("poll")) {
                return true;
            }
        }

        return false;
```

see https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/monitor/jvm/HotThreads.java#L94-L129 for details, so I'm not sure we should ignore all waiting or timed waiting threads so easy, I think we should go towards a more detailed ignore function as they do. What do you think?

logstash-core/lib/logstash/api/lib/app/commands/system/basicinfo_command.rb
ph: @suyograo +1 for making it simple :)

ph:
@purbon I understand your issue and I agree, even if we saw a lot of thread dump we don't know the granularity details that we can hide without making the dump useless. I think better filter rules will come with report of users and we will be able to improve our initial filtering. Also this will evolve with our growing usage of Java inside logstash.

purbon:
@ph I updated this PR with a more detailed function, partially inspired by the ES code,  will be adding comments now to clarify possible doubts.

purbon:
@ph last set of changes pushed, final LGTM?

logstash-core/lib/logstash/webserver.rb
```diff
@@ -62,7 +62,8 @@ def stop(options={})
         @runner.stop_blocked
       else
         @runner.stop
-      end
+      end rescue ""
```
ph: can you use `rescue nil` instead of an empty string?

logstash-core/lib/logstash/api/lib/app.rb
```diff
@@ -25,6 +25,11 @@ def initialize(app=nil)
       @factory = CommandFactory.new(settings.service)
     end
 
+    not_found do
+      status 404
+      respond_with({})
+    end
+
```
ph: Should we check for `?human` here to return a text version of the 404?

ph:
Added 2 minor comments, other than that LGTM.

logstash-core/lib/logstash/webserver.rb
purbon: sure

logstash-core/lib/logstash/api/lib/app.rb
purbon: we can yes.

PR #4665, opened 2016-02-09
Threadsafety problem when creating the initial metric instance

ph:
This PR fix an issue where we were incorrectly creating the first
instance of the metric. This was causing missing increments

### DIFF
--- a/logstash-core/lib/logstash/instrument/collector.rb
+++ b/logstash-core/lib/logstash/instrument/collector.rb
@@ -61,6 +61,9 @@ module LogStash module Instrument
       end
     end
 
+    def clear
+      @metric_store = MetricStore.new
+    end
 
     # Monitor the `Concurrent::TimerTask` this update is triggered on every successful or not
     # run of the task, TimerTask implement Observable and the collector acts as
--- a/logstash-core/lib/logstash/instrument/metric_store.rb
+++ b/logstash-core/lib/logstash/instrument/metric_store.rb
@@ -21,6 +21,10 @@ module LogStash module Instrument
       # We keep the structured cache to allow
       # the api to search the content of the differents nodes
       @store = Concurrent::Map.new
+
+      # This hash has only one dimension
+      # and allow fast retrieval of the metrics
+      @fast_lookup = Concurrent::Map.new
     end
 
     # This method use the namespace and key to search the corresponding value of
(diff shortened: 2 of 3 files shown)


logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -21,6 +21,10 @@ def initialize
       # We keep the structured cache to allow
       # the api to search the content of the differents nodes
       @store = Concurrent::Map.new
+
+      # This hash has only one dimension
+      # and allow fast retrieval of the metrics
+      @fast_lookup = Concurrent::Map.new
```
purbon: does this means we're keeping an internal cache for keys? is this a performance oriented optimization? .... I might not be for this change, but what about having actual pointers and not a copy of the metrics stored in two places? what do you think?

logstash-core/lib/logstash/instrument/collector.rb
```diff
@@ -61,6 +61,9 @@ def push(namespaces_path, key, type, *metric_type_params)
       end
     end
 
+    def clear
```
purbon: I only see this method being used in test, is this intentional?

purbon:
I tested the behaviour and works good for me. All event accounting is reasonable (ev in >= env filt >= env out).

logstash-core/lib/logstash/instrument/metric_store.rb
ph: Its for two things, its more performant to `o(1)` than do o(n) where `n` is the path of the namespace and the key.

Its not a copy of the metric, its a pointer.

logstash-core/lib/logstash/instrument/collector.rb
ph: Yes, I wanted to make sure each run were isolated, since the collector is still singleton. I have added this method.

logstash-core/lib/logstash/instrument/metric_store.rb
purbon: I' sorry, for me is a copy not a pointer to the right map place, now you are saving two times the same data.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -28,10 +32,20 @@ def initialize
...
-    # @param [Object] The default object if the value is not found in the path
+    # @param [Symbol] The metric key
     # @return [Object] Return the new_value of the retrieve object in the tree
     def fetch_or_store(namespaces, key, default_value = nil)
-      fetch_or_store_namespaces(namespaces).fetch_or_store(key, block_given? ? yield(key) : default_value)
+      provided_value =  block_given? ? yield(key) : default_value
+
+      # Check if the value is in the `fast_lookup` if it is return it, if not compute it
+      # and insert it into the store for
+      if found_value = @fast_lookup.put_if_absent([namespaces, key], provided_value)
```
ph:
```
We first check in the `@fast_lookup` store to see if we have already see that metrics before, This give us a `o(1)` access, which is faster than searching through the structured data store (Which is a `o(n)` operation where `n` is the number of element in the namespace and the value of the key). If the metric is already present in the `@fast_lookup`, the call  to `#put_if_absent` will return the value. This value is send back directly to the caller.

BUT. If the value is not present in the `#fastlookup` the value will be inserted and `#puf_if_absent` will return nil.
With this returned value  of `nil` we assume that we don't have it in the `@metric_store` for structured search so we add it there too.

The problem with only using the `@metric_store` directly all the time would require us to use the mutex around the structure since its a multi-level hash, without that it wont return a consistent value of the metric and this would slow down the code.
```

@purbon  can you review this comment ^?

purbon: LGTM man, ship it, ship it ! thanks a lot for clarification.

PR #4671, opened 2016-02-10
from_json should not generate new Event on blank/empty json input

colinsurprenant:
`Event#from_json` was previously generating empty `Event` on blank/empty json input but it should just ignore it and not generate empty event.

This complements logstash-plugins/logstash-codec-json_lines/pull/22

### DIFF
--- a/logstash-core-event-java/src/main/java/com/logstash/Event.java
+++ b/logstash-core-event-java/src/main/java/com/logstash/Event.java
@@ -162,12 +162,12 @@ public class Event implements Cloneable, Serializable {
     public static Event[] fromJson(String json)
             throws IOException
     {
-        Event[] result;
-
+        // empty/blank json string does not generate an event
         if (json == null || json.trim().isEmpty()) {
-            return new Event[]{ new Event() };
+            return new Event[]{ };
         }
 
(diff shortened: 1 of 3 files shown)


jordansissel:
Code LGTM. Looks like jenkins build passed. LGTM :)

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -162,12 +162,12 @@ public String toJson()
     public static Event[] fromJson(String json)
             throws IOException
     {
-        Event[] result;
-
+        // empty/blank json string does not generate an event
         if (json == null || json.trim().isEmpty()) {
-            return new Event[]{ new Event() };
+            return new Event[]{ };
```
talevy: why not return null to signify this?

colinsurprenant: because it makes one less return state to check. empty array will bubble up seamlessly to an empty Ruby array. returning null would require checking for null and acting on it.

jsvd:
LGTM

colinsurprenant:
merged in master and 2.x

Issue #4669, opened 2016-02-10
Removal of --filterworkers considered harmful


magnusbaeck:
I just learned the hard way that `--filterworkers` was removed in Logstash 2.2 (specifically, commit https://github.com/elastic/logstash/commit/028576ba6f141f57c180c8f3dbffa14cb24967c0). There are three problems with this:

* This was not mentioned in the [changelog](https://github.com/elastic/logstash/blob/2.2/CHANGELOG.md).
* There was no backwards compatible deprecation period which arguably would've made sense to match people's expectations for a non-major update like 2.2.
* The [command line docs for 2.2](https://github.com/elastic/logstash/blob/2.2/docs/static/command-line-flags.asciidoc) haven't been updated and still list `--filterworkers` as available.

This change broke Logstash Filter Verifier (https://github.com/magnusbaeck/logstash-filter-verifier/issues/4). Suggestions:

* Reinstate `--filterworkers` in 2.2.1 as a (possibly undocumented) alias of `-w`/`--pipeline-workers` to avoid additional breakage, and schedule the flag for permanent removal in 3.0.
* Update the changelog.
* Update the command-line documentation.

I can supply a PR for the latter two if you like. Do you want it to the 2.2 branch?

andrewvc:
Thanks for bringing this up @magnusbaeck . You're right, we definitely should reistate that option as an alias. I'm going to start work on that today. We'll bring it back as a deprecated option.

The tricky thing here is that conceptually filter workers are gone, since the new pipeline doesn't have that concept. On the other hand we shouldn't be breaking configs with a minor release, and it's compatible enough. Apologies for the troubles you've faced here @magnusbaeck .

suyograo:
For anyone reading this, `-w` flag still works. Agree `--filterworkers` should be reinstated as @andrewvc mentioned. We'll get this out asap.

jordansissel:
@magnusbaeck thanks for catching this!

suyograo:
@andrewvc keeping this open until we port this to master and 2.x

andrewvc:
@suyograo I don't believe we should port to master since the next major won't support this functionality. This is already merged into 2.x. I think we can close it.

suyograo:
Yep, not to master.

magnusbaeck:
Thanks for the quick turnaround! But the 2.2.0 changelog should still be updated, right?

suyograo:
@magnusbaeck yes, I will update the 2.2.0 changelog.

PR #4672, opened 2016-02-10
This brings back --filter-workers for the 2.x series. This will be go…

andrewvc:
…ne in the next major version

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -36,6 +36,10 @@ class LogStash::Agent < Clamp::Command
          :attribute_name => :pipeline_batch_delay,
          :default => LogStash::Pipeline::DEFAULT_SETTINGS[:pipeline_batch_delay]
 
+  option ["--filterworkers"], "COUNT",
+         I18n.t("logstash.agent.flag.filterworkers"),
+         :attribute_name => :filter_workers
+
   option ["-l", "--log"], "FILE",
     I18n.t("logstash.agent.flag.log"),
     :attribute_name => :log_file
@@ -134,6 +138,12 @@ class LogStash::Agent < Clamp::Command
     end
     configure
 
+
+    if filter_workers
+      @logger.warn("--filter-workers is deprecated! Please use --pipeline-workers or -w. This setting will be removed in the next major version!")
+      self.pipeline_workers = filter_workers
+    end
+
     # You must specify a config_string or config_path
     if @config_string.nil? && @config_path.nil?
       fail(help + "\n" + I18n.t("logstash.agent.missing-configuration"))
@@ -156,6 +166,7 @@ class LogStash::Agent < Clamp::Command
         @config_string += DEFAULT_OUTPUT
       end
     end
+
 
     begin
       pipeline = LogStash::Pipeline.new(@config_string, @pipeline_settings)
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/agent.rb
```diff
@@ -134,6 +138,12 @@ def execute
     end
     configure
 
+
+    if filter_workers
+      @logger.warn("--filter-workers is deprecated! Please use --pipeline-workers. This setting will be removed in the next major version!")
+      self.pipeline_workers=filter_workers
```
jordansissel: spacing around `=`

logstash-core/lib/logstash/agent.rb
```diff
@@ -36,6 +36,10 @@ class LogStash::Agent < Clamp::Command
          :attribute_name => :pipeline_batch_delay,
          :default => LogStash::Pipeline::DEFAULT_SETTINGS[:pipeline_batch_delay]
 
+  option ["--filter-workers"], "COUNT",
```
suyograo: Its `--filterworkers` not `--filter-workers`

suyograo:
```
suyog@machine:/tmp/logstash-1.5.5$ bin/logstash --help
Usage:
    /bin/logstash agent [OPTIONS]

Options:
    -f, --config CONFIG_PATH      Load the logstash config from a specific file
                                  or directory.  If a directory is given, all
                                  files in that directory will be concatenated
                                  in lexicographical order and then parsed as a
                                  single config file. You can also specify
                                  wildcards (globs) and any matched files will
                                  be loaded in the order described above.
    -e CONFIG_STRING              Use the given string as the configuration
                                  data. Same syntax as the config file. If no
                                  input is specified, then 'stdin
                                  { type => stdin }' is the default input.
                                  If no output is specified, then  'stdout
                                  { codec => rubydebug } }' is default output.
                                   (default: "")
    -w, --filterworkers COUNT     Sets the number of filter workers to run.
```

jordansissel:
Can you update the commit message to say `Fixes #4669` on a line by itself at the end?

logstash-core/lib/logstash/agent.rb
```diff
@@ -134,6 +138,12 @@ def execute
     end
     configure
 
+
+    if filter_workers
+      @logger.warn("--filter-workers is deprecated! Please use --pipeline-workers. This setting will be removed in the next major version!")
```
suyograo: Add 
>
Please use --pipeline-workers or -w. This setting..

logstash-core/lib/logstash/agent.rb
jordansissel: Good catch.

logstash-core/locales/en.yml
```diff
@@ -157,6 +157,8 @@ en:
           Check configuration for valid syntax and then exit.
         pipeline-workers: |+
           Sets the number of pipeline workers to run.
+        filter-workers: |+
```
suyograo: `filterworkers`

logstash-core/lib/logstash/agent.rb
andrewvc: can do.

andrewvc:
@suyograo @jordansissel I believe I have addressed all concerns here with my last force push.

suyograo:
LGTM

PR #4670, opened 2016-02-10
Fix an issue with the events count and the naming of the recorded metric at the plugin level.

ph:
- Add more test for the FilterDelegator, we did not have any before.
- Fix an issue when the plugin buffer events to keep the accurate count
- Fix an issue with the naming of the metrics at the plugin level

PR #4675, opened 2016-02-11
Make proper test integration for the metrics api

purbon:
Make usage of proper test integration with the main core pipeline for the logstash web api, this basically means this PR is cleaning up all unnecessary mocks and going through a real agent and pipeline for testing purposes.

Still room to improve the test, this this is a first pass, next steps are to actually improve the api endpoint test for expected behaviour.

Small set of test improvements, like:

* Cleanup non used fixture.
* Improve node_stats tests by adding more context and format checks.
* Add a new node_specs endpoint test that checks the expected data is
  returned for the hot_threads resource.

@ph what do you think?

//cc @jsvd


logstash-core/lib/logstash/api/lib/app/service.rb
```diff
@@ -40,4 +44,15 @@ def get(key)
     end
     LogStash::Json.dump(data)
   end
+
+  private
+
+  def has_counters?
```
purbon: This was actually a source of concern to me, as for testing I do really need to wait till some events are generated. Obviously open to improve this part of the code.

rakelib/default_plugins.rb
```diff
@@ -119,6 +119,8 @@ module RakeLib
       logstash-input-stdin
       logstash-input-tcp
       logstash-output-stdout
+      logstash-output-null
```
suyograo: why this change? Can't we add this to dev dependencies?

purbon: we could yes, the idea is to use this in the agent pipeline for this test, so nothing is really output, but still everything is flowing for all the stages.

logstash-core/lib/logstash/api/lib/app/service.rb
ph: What I did in my test is to use the generator and custom output with an accessor for the number of events received.
So I wait until the output has received all the events generated by the generator.

logstash-core/lib/logstash/api/lib/app/service.rb
```diff
@@ -40,4 +44,15 @@ def get(key)
...
+    return false if @snapshot.nil?
+    if @snapshot_rotation_mutex.try_lock
+      has = @snapshot.metric_store.all.clone.map { |t| t.class.to_s }.include?("LogStash::Instrument::MetricType::Counter")
+      @snapshot_rotation_mutex.unlock
+      return has
+    end
+  end
```
ph: Since I have the `@fast_lookup` now I can add a `size` or has_metrics? Method would that help you?

ph: Also since all metrics are lazy created if you do `@snapshot.metric_store.all.size > 0` this will tell you if we have any metrics in the store.  

But lets create a predicated for it. `#empty?` `#present?` Naming is hard :)

logstash-core/spec/api/lib/api/node_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    context "when asking for human output" do
+      before(:all) do
+        get "/hot_threads?human"
+      end
+
+      let(:payload) { last_response.body }
+
+      it "shuold return a plain text description" do
```
ph: small typo `shuold` -> `should`, Should we check for `content_type`?

logstash-core/spec/api/spec_helper.rb
```diff
@@ -17,3 +17,40 @@ def read_fixture(name)
...
+    allow(LogStash::Instrument::Collector.instance).to receive(:agent).and_return(agent)
+  end
+end
+
+def wait_until_snapshot_received
+  while !LogStash::Api::Service.instance.started? do
+    sleep 0.5
+  end
+end
+wait_until_snapshot_received
```
ph: I have a small question could this `spec_helper` make some conflict with our other `spec/spec_helper` ?

ph:
@purbon I really like that! thanks for making it happen. I have similar test for the pipeline in #4670

logstash-core/lib/logstash/api/lib/app/service.rb
purbon: that would help, but my issue here is more on the side of having counters there. it takes some time for the pipeline to actually generate some events acouting information, and to test for example the  `/events` resource I do need this to be there. JVM periodic pollers is working like charm and providing data very fast.

logstash-core/lib/logstash/api/lib/app/service.rb
purbon: That could be a good idea, but I'm more on the side of letting the collector notify when he has received events, or any other side of information we aim for. On the other side, I'm starting to believe we should cleanup all this real/dummy plugins and move to a unified solution here.

ph: Well, we can also change the behavior if the collector didn't collect any new metrics we dont send the snapshot?

logstash-core/lib/logstash/api/lib/app/service.rb
ph: true, if you are targetting the events.

rakelib/default_plugins.rb
purbon: Sorry to say, we could not as we need to have this as runtime dependencies because of the plugins system.

logstash-core/lib/logstash/api/lib/app/service.rb
purbon: well, it collects metrics somehow, as it collect the periodic poller metrics anyway. this is only a testing issue for now.

logstash-core/spec/api/spec_helper.rb
purbon: I don't think so as is only used by the API tests and is only locally required by them.

purbon:
closing this in benefit of a new PR to master, as the feature/metrics branch got merged recently.

PR #4676, opened 2016-02-11
Collection of fixes from last round if reviews in the feature metrics PR

purbon:
Hi,
  this actually makes sure to fix a few of the concerns coming out of the last round of reviews in the metrics api, in detail:

* make sure puma version >= 2.16 (same as in the http input)
* add http-host option to the web server.
* fix typos and wrongly removed lines.

@ph what do you think?

/cc @jsvd

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@
 
 source "https://rubygems.org"
 gem "logstash-core", "3.0.0.dev", :path => "./logstash-core"
+# gem "logstash-core-event", "3.0.0.dev", :path => "./logstash-core-event"
 gem "logstash-core-event-java", "3.0.0.dev", :path => "./logstash-core-event-java"
 gem "file-dependencies", "0.1.6"
 gem "ci_reporter_rspec", "1.0.0", :group => :development
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -84,7 +84,7 @@ class LogStash::Runner < Clamp::Command
 
   option ["--http-host"], "WEB_API_HTTP_HOST",
     I18n.t("logstash.web_api.flag.http_host"),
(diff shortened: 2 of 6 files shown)


logstash-core/lib/logstash/runner.rb
```diff
@@ -82,7 +82,11 @@ class LogStash::Runner < Clamp::Command
     I18n.t("logstash.runner.flag.auto_reload"),
     :attribute_name => :auto_reload, :default => false
 
-  option ["-ap", "--http-port"], "WEB_API_HTTP_PORT",
+  option ["--http-host"], "WEB_API_HTTP_HOST",
+    I18n.t("logstash.web_api.flag.http_host"),
+    :attribute_name => :web_api_http_host, :default => "0.0.0.0"
```
suyograo: Can we use `127.0.0.1` as default instead of `0.0.0.0`

Gemfile
```diff
@@ -3,6 +3,7 @@
 
 source "https://rubygems.org"
 gem "logstash-core", "3.0.0.dev", :path => "./logstash-core"
+# gem "logstash-core-event", "3.0.0.dev", :path => "./logstash-core-event"
```
ph: I think this is not the last time we are bitten by this line.

logstash-core/lib/logstash/runner.rb
ph: +1

ph:
Only the minor comments concerning 127.0.0.1 and it lgtm.
We should bind to localhost only by default and not all the interface.

logstash-core/lib/logstash/runner.rb
purbon: sure

Gemfile
purbon: yeah, I do really think we should remove it ....

PR #4677, opened 2016-02-11
Reviews from the metric collection and api pr.

ph:


logstash-core/lib/logstash/agent.rb
```diff
@@ -18,7 +18,9 @@
 LogStash::Environment.load_locale!
 
 class LogStash::Agent
-  attr_reader :metric, :debug, :node_name, :started_at, :pipelines, :logger
+  STARTED_AT = Time.now
```
purbon: you should freeze the constant.

logstash-core/lib/logstash/api/lib/app/command.rb
```diff
@@ -23,7 +23,7 @@ def uptime
     end
 
     def started_at
-      (service.agent.started_at.to_f*1000.0).to_i
+      (LogStash::Agent::STARTED_AT.to_f * 1000.0).to_i
```
purbon: should you always store the .to_f?

purbon:
left only small comment, in general LGTM.

logstash-core/lib/logstash/agent.rb
ph: +1

logstash-core/lib/logstash/api/lib/app/command.rb
ph: I don't think its needed, its not hit really often. I don't believe we need this optimization

PR #4681, opened 2016-02-12
feature/metrics cleanups 

purbon:
* cleanups debug variable used inside the pipeline.
* fixed small typo in the memory command.

### DIFF
--- a/logstash-core/lib/logstash/webserver.rb
+++ b/logstash-core/lib/logstash/webserver.rb
@@ -20,7 +20,8 @@ module LogStash
       http_port    = options[:http_port] || 9600
       @options     = {}
       @cli_options = options.merge({ :rackup => ::File.join(::File.dirname(__FILE__), "api", "init.ru"),
-                                     :binds => ["tcp://#{http_host}:#{http_port}"] })
+                                     :binds => ["tcp://#{http_host}:#{http_port}"],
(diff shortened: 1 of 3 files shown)


purbon:
@jsvd @ph thoughts?

logstash-core/lib/logstash/webserver.rb
```diff
@@ -20,7 +20,8 @@ def initialize(logger, options={})
       http_port    = options[:http_port] || 9600
       @options     = {}
       @cli_options = options.merge({ :rackup => ::File.join(::File.dirname(__FILE__), "api", "init.ru"),
-                                     :binds => ["tcp://#{http_host}:#{http_port}"] })
+                                     :binds => ["tcp://#{http_host}:#{http_port}"],
+                                     :debug => logger.debug? })
```
jsvd: is this still necessary? isn't the logger passed to the webserver?

purbon: Is necessary to tell puma he is suposed to work on debug mode, only
possible throught cli options.

purbon: and to provide more context, no logger is not passed to any webserver here. This works just as interface to puma. Logger is also used inside the sinatra app, but this is another moving component.

jsvd:
LGTM

Issue #4667, opened 2016-02-10
Improve current test suites for the WebApi integration


purbon:
We need to improve a bit the coverage and resiliency of the web api tests, this will include being sure that:

* Mocks are done where it makes sense, integration with the pipeline and collector interface should not be mocked.
* All resources are tested for presence, this should be done with an easy and simple test.
* Test parameters are properly checked.
* Error and misbehave conditions are checked.

purbon:
Partial fix for this implemented in #4782, then is just time to make the test more complete.

PR #4687, opened 2016-02-12
Fix an issue with the expectation on the agent#uptime

ph:
Fixes: https://github.com/elastic/logstash/issues/4686

### DIFF
--- a/logstash-core/spec/api/lib/commands/jvm_spec.rb
+++ b/logstash-core/spec/api/lib/commands/jvm_spec.rb
@@ -12,6 +12,7 @@ describe "JVM stats" do
     before(:each) do
       allow(agent).to receive(:node_name).and_return("foo")
       expect_any_instance_of(LogStash::Api::Service).to receive(:agent).and_return(agent)
+      allow(subject).to receive(:uptime).and_return(10)
     end
 
     context "#schema" do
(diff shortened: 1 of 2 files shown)


ph:
@jsvd Should fix the problem you have found concerning the JVM hot threads.

https://github.com/elastic/logstash/issues/4667 will make this code obsolete.

suyograo:
LGTM

logstash-core/spec/api/lib/commands/jvm_spec.rb
```diff
@@ -12,6 +12,7 @@
     before(:each) do
       allow(agent).to receive(:node_name).and_return("foo")
       expect_any_instance_of(LogStash::Api::Service).to receive(:agent).and_return(agent)
+      allow(subject).to receive(:uptime).and_return(10)
```
jsvd: I suggest 42

suyograo: Is `10` mocked or injected somewhere?

ph: Its a double, not a real instance so in this case the number doesn't matter #4667 is better testing strategy since we actually starting a logstash instance under the hood with a config and test the api end points.

Issue #3944, opened 2015-09-20
Add support for environment variable injection in logstash plugin configuration


fbaligand:
It would be great to support environment variable injection in logstash configuration, like this : 
```
tcp {
port: "${TCP_PORT}"
}
```
It would be very useful to have a logstash configuration independent from its environment. And so, have the same logstash configuration among different environments (dev, test, prod, ...)

Numerous frameworks support such a feature like spring or log4j.

jordansissel:
As a workaround, you can use a template tool (m4, sed, etc) to achieve
this. Run it on your config before starting logstash. This is what I've
done in the past with good success. Hope this helps :)

fbaligand:
Thanks for your reply @jordansissel.
I know this workaround because I have read your answer in some forums or issues ;)
But it would be really great to support it dynamically in logstash !
It avoids to generate an instance configuration from a template configuration, each time we start logstash.

cstockton:
Given the robustness of logstash configuration files I am a bit surprised there is not a cleaner way to do this. All of our logstash instances are containerized within Docker, getting our various containers to talk to eachother is done via Docker's container linking.. they provide you with some environmental variables so your containers can find each other.

That said, even if something like M4 was suitable for configuration preprocessing in a grammar as expressive as logstash's (it isn't) .. it would be difficult (messy at best) to hook the preprocessor into the various container deployment and orchestration systems. Your containers will be expected to be built ahead of time and be environment agnostic. Typically getting all the context they need at runtime from the environment variables.

@fbaligand The solution I have today uses the environment filter, but it is not ideal because it is including all of the environment vars in the message. Maybe you found a better way?

```bash
filter {
  ...
  environment {
    add_field_from_env => {
      "MY_ENV_ADDR" => "MY_ENV_PROD_PORT_12285_TCP_ADDR"
      "MY_ENV_PORT" => "MY_ENV_PROD_PORT_12285_TCP_PORT"
    }
  }
}
...
output {
  http {
    codec => "json"
    http_method => "post"
    url => "http://%{MY_ENV_ADDR}:%{MY_ENV_PORT}/..."
  }
}
```

Is there a way to specify local fields i.e.: not to be included in final output, but used for transit through the pipeline? Or maybe there is a syntax for defining fields to namespace them? There might be a way to do what I am trying to do in the logstash world, just haven't found it.

[...]

PR #4710, opened 2016-02-21
Fix issue #3944 : Add support for environment variable injection in logstash plugin configuration

fbaligand:
This PR implements issue #3944.

The implementation is coherent with beats implementation : 
https://github.com/elastic/beats/pull/715

The implementation follows these specifications :
- replaces `${var}` or `$var` in config files according to the values of the current environment variables - if - if environment variable is not existing, placeholder is replaced by empty string and a warn log is added
- add info log about replacements (like beats)
- process string, number, boolean, array and hash param values
- process default value with ":" separator : `${var:defaultValue}`

### DIFF
--- a/docs/static/configuration.asciidoc
+++ b/docs/static/configuration.asciidoc
@@ -599,6 +599,184 @@ output {
 }
 ----------------------------------
 
+[[environment-variables]]
+=== Using Environment Variables in Configuration
+==== Overview
+
+* You can set environment variable references into Logstash plugins configuration using `${var}` or `$var`.
+* Each reference will be replaced by environment variable value at Logstash startup.
+* The replacement is case-sensitive.
+* References to undefined variables raise a Logstash configuration error.
+* A default value can be given by using the form `${var:default value}`.
+* You can add environment variable references in any plugin option type : string, number, boolean, array or hash.
+* Environment variables are immutable. If you update the environment variable, you'll have to restart Logstash to pick the updated value.
+
+==== Examples
+
+[cols="a,a,a"]
+|==================================
+|Logstash config source	|Environment 	|Logstash config result
+
+|
+[source,ruby]
+----
+input {
+  tcp {
+    port => "$TCP_PORT"
+  }
+}
+----
+
+|
+[source,shell]
+----
+export TCP_PORT=12345
+----
+|
+[source,ruby]
+----
+input {
+  tcp {
+    port => 12345
+  }
+}
+----
+|
+[source,ruby]
+----
+input {
+  tcp {
+    port => "${TCP_PORT}"
+  }
+}
+----
+
+|
+[source,shell]
+----
+export TCP_PORT=12345
+----
+|
+[source,ruby]
+----
+input {
+  tcp {
+    port => 12345
+  }
+}
+----
+|
+[source,ruby]
+----
+input {
+  tcp {
+    port => "${TCP_PORT}"
+  }
+}
+----
+
+|
+No TCP_PORT defined
+|
+Raise a logstash configuration error
+|
+[source,ruby]
+----
+input {
+  tcp {
+    port => "${TCP_PORT:54321}"
+  }
+}
+----
+
+|
+No TCP_PORT defined
+|
+[source,ruby]
+----
+input {
+  tcp {
+    port => 54321
+  }
+}
+----
+|
+[source,ruby]
+----
+input {
+  tcp {
+    port => "${TCP_PORT:54321}"
+  }
+}
+----
+
+|
+[source,shell]
+----
+export TCP_PORT=12345
+----
+|
+[source,ruby]
+----
+input {
+  tcp {
+    port => 12345
+  }
+}
+----
+|
+[source,ruby]
+----
+filter {
+  mutate {
+    add_tag => [ "tag1", "${ENV_TAG}" ]
+  }
+}
+----
+
+|
+[source,shell]
+----
+export ENV_TAG="tag2"
+----
+|
+[source,ruby]
+----
+filter {
+  mutate {
+    add_tag => [ "tag1", "tag2" ]
+  }
+}
+----
+|
+[source,ruby]
+----
+filter {
+  mutate {
+    add_field => {
+      "my_path" => "${HOME}/file.log"
+    }
+  }
+}
+----
+|
+[source,shell]
+----
+export HOME="/path"
+----
+|
+[source,ruby]
+----
+filter {
+  mutate {
+    add_field => {
+      "my_path" => "/path/file.log"
+    }
+  }
+}
+----
+|==================================
+
 [[config-examples]]
 === Logstash Configuration Examples
 The following examples illustrate how you can configure Logstash to filter events, process Apache logs and syslog messages, and use conditionals to control what events are processed by a filter or output.
--- a/logstash-core/lib/logstash/config/mixin.rb
+++ b/logstash-core/lib/logstash/config/mixin.rb
@@ -37,6 +37,9 @@ module LogStash::Config::Mixin
 
   PLUGIN_VERSION_1_0_0 = LogStash::Util::PluginVersion.new(1, 0, 0)
   PLUGIN_VERSION_0_9_0 = LogStash::Util::PluginVersion.new(0, 9, 0)
+  
+  ENV_PLACEHOLDER_REGEX = Regexp.new(/\$\w+|\$\{\w+(\:[^}]*)?\}/)
+
 
   # This method is called when someone does 'include LogStash::Config'
   def self.included(base)
@@ -99,6 +102,23 @@ module LogStash::Config::Mixin
       end
     end
 
+    # Resolve environment variables references
+    params.each do |name, value|
+      if (value.is_a?(Hash))
+        value.each do |valueHashKey, valueHashValue|
+          value[valueHashKey.to_s] = replace_env_placeholders(valueHashValue)
+        end
+      else
+        if (value.is_a?(Array))
+          value.each_index do |valueArrayIndex|
+            value[valueArrayIndex] = replace_env_placeholders(value[valueArrayIndex])
+          end
+        else
+          params[name.to_s] = replace_env_placeholders(value)
+        end
+      end
+    end
+
     if !self.class.validate(params)
       raise LogStash::ConfigurationError,
         I18n.t("logstash.runner.configuration.invalid_plugin_settings")
@@ -125,6 +145,42 @@ module LogStash::Config::Mixin
 
     @config = params
   end # def config_init
+
+  # Replace all environment variable references in 'value' param by environment variable value and return updated value
+  # Process following patterns : $VAR, ${VAR}, ${VAR:defaultValue}
+  def replace_env_placeholders(value)
+    if (value.is_a?(String))
+      while value =~ ENV_PLACEHOLDER_REGEX
+        valueParts = value.partition(ENV_PLACEHOLDER_REGEX)
+        placeHolder = valueParts[1]
+        if placeHolder.start_with?("${")
+          envVarName = placeHolder.slice(2..placeHolder.length - 2)
+        else
+          envVarName = placeHolder.slice(1..placeHolder.length - 1)
+        end
+        if envVarName.include?(':')
+          placeHolderParts = envVarName.split(':')
+          envVarName = placeHolderParts[0]
+          defaultValue = placeHolderParts[1] || ""
+        else
+          defaultValue = ""
+        end
+        
+        envVarValue = ENV[envVarName]
+        if envVarValue.nil? 
+          envVarValue = defaultValue
+          if defaultValue.empty?
+            @logger.warn("In plugin '#{self.class.config_name}', " +
+                         "referenced environment variable '#{envVarName}' does not exist. " +
+                         "This reference has been replaced by empty string.")
+          end
+        end
+        value = valueParts[0] << envVarValue << valueParts[2]
+        @logger.info("Replacing config environment variable '#{placeHolder}' with #{envVarValue}")
+      end
+    end
+    return value
+  end # def replace_env_placeholders
 
   module DSL
     attr_accessor :flags
--- a/logstash-core/lib/logstash/config/mixin.rb
+++ b/logstash-core/lib/logstash/config/mixin.rb
@@ -38,8 +38,7 @@ module LogStash::Config::Mixin
   PLUGIN_VERSION_1_0_0 = LogStash::Util::PluginVersion.new(1, 0, 0)
   PLUGIN_VERSION_0_9_0 = LogStash::Util::PluginVersion.new(0, 9, 0)
   
-  ENV_PLACEHOLDER_REGEX = Regexp.new(/\$\w+|\$\{\w+(\:[^}]*)?\}/)
-
+  ENV_PLACEHOLDER_REGEX = /\$(?<name>\w+)|\$\{(?<name>\w+)(\:(?<default>[^}]*))?\}/
 
   # This method is called when someone does 'include LogStash::Config'
   def self.included(base)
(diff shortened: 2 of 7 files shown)


jordansissel:
Howdy! Thank you for this proposed change. Let's discuss the details of the proposal and possible solutions in #3944 before moving forward on this PR.

suyograo:
@fbaligand Thanks much for your contribution. We'll review this shortly. Can you please add some documentation (in a separate section in asciidoc). We can help you if you need any info regarding asciidoc..

fbaligand:
Hi @suyograo,

Happy to see you appreciate my contribution. 
For asciidoc, where exactly do you want I add asciidoc ? In mixin.rb file ? Else where ?

suyograo:
@fbaligand can you add a new section "Using Environment Variables in Configuration" here https://github.com/elastic/logstash/blob/master/docs/static/configuration.asciidoc

I can help review and augment it. Lets provide examples, syntax info etc.

logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -126,6 +146,42 @@ def config_init(params)
...
+        if envVarValue.nil? 
+          envVarValue = defaultValue
+          if defaultValue.empty?
+            @logger.warn("In plugin '#{self.class.config_name}', " +
+                         "referenced environment variable '#{envVarName}' does not exist. " +
+                         "This reference has been replaced by empty string.")
+          end
+        end
+        value = valueParts[0] << envVarValue << valueParts[2]
+        @logger.info("Replacing config environment variable '#{placeHolder}' with #{envVarValue}")
```
suyograo: Can we make this debug logs?

logstash-core/spec/logstash/config/mixin_spec.rb
```diff
@@ -151,4 +151,50 @@
...
+      plugin_class.new({
+        "oneString" => "${notExistingVar}",
+        "oneBoolean" => "${notExistingVar:true}",
+        "oneNumber" => "${MIXIN_SPEC_ENV_VAR}",
+        "oneArray" => [ "first array value", "$MIXIN_SPEC_ENV_VAR" ],
+        "oneHash" => { "key" => "$MIXIN_SPEC_ENV_VAR" }
+      })
+    }
+     
+    it "should have oneString param as empty string (env var not found)" do
```
suyograo: Can we add spec for multiple env variables in the same config? For example `MIXIN_SPEC_ENV_VAR` `MIXIN_SPEC_ENV_VAR2` .. and make sure they are replaced accordingly.

suyograo:
I'd also like to see the following tests:
* `${TEST_TCP_PORT}:5600`: Defaults not applied, config error is thrown
* `${TEST_TCP_PORT-5600}`: Throws config error, bad delimiter
* `${TEST_HOST_NAME:abc def}`: Spaces in env variable
* `${TEST_HOST_NAME: }`: Default as space
* `${TEST_HOST_NAME:}`: Default is non-space. Replaced by empty string
* `${test_home_name:abc}` and `${TEST_HOST_NAME:def}`: Should be case-sensitive

suyograo:
What do you think of just supporting syntax: `${ENV_VAR}` and rejecting `$ENV_VAR` ? Seems to make it simpler?

logstash-core/lib/logstash/config/mixin.rb
fbaligand: For asciidoc, I can do this, no problem.
Concerning the content, does the beats PR description (https://github.com/elastic/beats/pull/715) fits what you want ?

suyograo: Yes, it would be lovely to show a real world example. The ones users had requested in #3944. Like injecting `TCP_PORT` or adding `tags` using environment variable

fbaligand: Concerning logs, I make a warn log only if environment variable is not found.
And for info log, I make exactly the same log content and same log content than Beats PR : 
https://github.com/elastic/beats/pull/715

That said, if you want that I make these 2 logs using debug level, no problem.

logstash-core/spec/logstash/config/mixin_spec.rb
fbaligand: no problem, I'll do it.

logstash-core/lib/logstash/config/mixin.rb
suyograo: related to this, @jordansissel had concerns about logging if no environment variable is found. He mentioned we should throw a config error. I'd love for him to chime in here..

I think env variable replacement can be `debug` instead of `info`. LS, out of the box is configured to  `info` level at console.

jordansissel: @suyograo LS uses warn by default. My rough guide for log levels has been info == user-level troubleshooting and debug == developer-level troubleshooting. Given this, I think info level makes sense.

I'll comment separately about the 'no env var found' stuff ;P

jordansissel:
I'm OK with most of this proposal, and I would like to make one modification to the proposed behavior: It should be an error to use an environment variable that doesn't exist except in cases where a default value is provided. This should be a configuration error that causes Logstash to fail to use the proposed configuration.

In terms of documentation, I given the ambiguity available (what syntax is valid, what is an error, etc), it'd be ideal if we could launch this with docs that help users understand the limits and boundaries of this new feature.

While driving today, I spent some time trying to convince myself to like this feature (even if I personally may not want it, it is clear that many users _do_ want it). I found at least one case, yay! In cases where Logstash has a default value that may depend on the environment, such as the file input's use of $HOME for a default, right now we have some confusing documentation that says for [file input's `sincedb_path`](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-file.html#plugins-inputs-file-sincedb_path): "There is no default value for this setting." -- Yet, further down the doc says "The default will write sincedb files to some path matching $HOME/.sincedb"  -- a statement which is kind of in conflict with "there is no default". Having env var support can help us resolve these situations as I can envision a future where we simply say `The default value is ${HOME}` and my proposed modification - a missing env var being an error - would cause a configuration error to be reported that should be pretty clear to users.

I am happy to work with you to get the docs written and reviewed. Thanks for your efforts on this.

logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -99,6 +102,23 @@ def config_init(params)
       end
     end
 
+    # Resolve environment variables references
+    params.each do |name, value|
+      if (value.is_a?(Hash))
+        value.each do |valueHashKey, valueHashValue|
+          value[valueHashKey.to_s] = replace_env_placeholders(valueHashValue)
```
jordansissel: Why use the string (`.to_s`) value of `valueHashKey` here?

logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -126,6 +146,42 @@ def config_init(params)
     @config = params
   end # def config_init
 
+  # Replace all environment variable references in 'value' param by environment variable value and return updated value
+  # Process following patterns : $VAR, ${VAR}, ${VAR:defaultValue}
+  def replace_env_placeholders(value)
+    if (value.is_a?(String))
+      while value =~ ENV_PLACEHOLDER_REGEX
```
jordansissel: I think there's a shorter/simpler way to do this. Doing it shorter will require two things:

1) Using named captures in your ENV_PLACEHOLDER_REGEX
2) Using `String#gsub`

Rough proposal (I haven't tested thoroughly). Demonstrated in irb:

```ruby
>> ENV_PLACEHOLDER_REGEX = Regexp.new(/\$(?<env>\w+)|\$\{(?<env>\w+)(\:(?<default>[^}]*))?\}/)
>> text = "hello ${world} ${fancy} whoa ${missing:cool default}"
=> "hello ${world} ${fancy} whoa ${missing:cool default}"

# Demonstrating it using String#gsub and regexp named captures
>> text.gsub(ENV_PLACEHOLDER_REGEX) { |t| m = Regexp.last_match; value = map[m["env"]]; value.nil? ? m["default"] : value }
=> "hello earth pants whoa cool default"
```

Using named captures will allow you to lookup the env var name requested instead of having to compute it (as in lines 154-160) and also let you easily look up the default value (if provided).

---

I am not sure I am making sense here, so let me know if I'm being confusing. My overall feeling is that moving to use String#gsub + named captures allows you to shrink lines 152-180 to be something much more concise (and hopefully easier to debug should things go wrong)

logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -126,6 +146,42 @@ def config_init(params)
     @config = params
   end # def config_init
 
+  # Replace all environment variable references in 'value' param by environment variable value and return updated value
+  # Process following patterns : $VAR, ${VAR}, ${VAR:defaultValue}
+  def replace_env_placeholders(value)
+    if (value.is_a?(String))
```
jordansissel: Seems like we do nothing if the 'value' is not a string. Should it be an error to call this method with anything but a string?

logstash-core/lib/logstash/config/mixin.rb
jordansissel: I've opened https://github.com/fbaligand/logstash/pull/1 against your fork which implements my suggestion.

It also implements my proposal to fail with an error if the user provides a placeholder that refers to a nonexistent env var and also doesn't provide a default.

jordansissel:
Submitted a [PR](https://github.com/fbaligand/logstash/issues/1) for this PR (inception!) with my proposed two changes. @fbaligand - Let me know what you think. If you like it, merge it into your fork and it should show up here automatically (in theory, I don't remember the last time I did a PR against a PR)

fbaligand:
Wow, lots of comments ! ;)

@suyograo 
Concerning new tests, I agree that more tests are welcome.
But before adding them, I must tell you the current implementation behaviour (which is not necessarily what you expect), so that we can discuss about desired behaviour :

- `${TEST_TCP_PORT}:5600`: presently, no error is thrown, the result is just : `<env value>:5600` (only the expression matching expected regex pattern is replaced. And to me, that is very very difficult to detect patterns which don't match expected pattern)
- `${TEST_TCP_PORT-5600}`: presently, no error is thrown, the result is equal to input, because `-` doesn't match `\w`, so the full expression does not match expected regex pattern.
- `${TEST_HOST_NAME:abc def}`: should work as expected
- `${TEST_HOST_NAME: }`: should work as expected
- `${TEST_HOST_NAME:}`: should work as expected, replacing it by empty string
- `${test_home_name:abc}` and `${TEST_HOST_NAME:def}`: should work as expected

fbaligand:
@suyograo  @jordansissel 
Concerning support for `${var}` and `$var` ; and a "not found env variable" replaced by empty string ; I followed exactly the same specifications than Beats (https://github.com/elastic/beats/pull/715).
My aim was to have a coherent behaviour between this two elastic projects.

So what do you think about ?
Do we do differently in logstash ?

logstash-core/lib/logstash/config/mixin.rb
fbaligand: The only reason was because in `mixin.rb`, there are numerous `name.to_s` while processing params (ex: line 87).
But I agree that for a hash key, there's probably no need to do that.

Tell me if you want I remove it.

logstash-core/lib/logstash/config/mixin.rb
fbaligand: I agree that using named captures is cleaner !
And don't worry, you was not confusing at all, I have fully understood what you mean :)

So it's ok to me to do like that.

logstash-core/lib/logstash/config/mixin.rb
fbaligand: If 'value' is not a string, the method will raise an error because it uses 'string' methods on 'value'.
Maybe it would be cleaner to make this test at line 117 ?

fbaligand:
@jordansissel Thanks for the PR to my PR ! :))
As you say, we remake Inception film :)

I just reviewed it and make two comments.
But globally, it is OK.

The only remaining question is what I say in my previous comment : if we raise an error when env variable is not found, logstash behaviour would be different from Beats.

So let me know what you think about that.

fbaligand:
For documentation, I agree that "documentation by example" is really nice.
I will do so.

suyograo:
@fbaligand I am ok with throwing an exception rather than silently replacing with empty string. I think it is ok to deviate from beats here, and if users provide feedback we can always go back to previous behavior. 

Also, lets keep the support for both `${var}` and `$var`. 

Can you merge Jordan's PR into yours?

fbaligand:
I just merged Jordan's PR !
It's fun to inverse the roles :)

fbaligand:
@suyograo 
What do you think about my comment on your suggested tests ?

suyograo:
@fbaligand thanks. Can you add an explicit test for validating these, we are pretty close to merging this:

> `${TEST_HOST_NAME:abc def}`: should work as expected
`${TEST_HOST_NAME:}`: should work as expected, replacing it by empty string
`${test_home_name:abc}` and `${TEST_HOST_NAME:def}`: should work as expected

I am ok with the other scenarios.

fbaligand:
@suyograo @jordansissel 
I just added 'Using Environment Variables in Configuration' section in configuration documentation.
Could you review it ?
Don't hesitate to make a PR on my PR :)

docs/static/configuration.asciidoc
```diff
@@ -599,6 +599,199 @@ output {
 }
 ----------------------------------
 
+[[environment-variables]]
+=== Using Environment Variables in Configuration
+== Overview
+
+You can set environment variable references into logstash plugins configuration using `${var}` or `$var`.
```
suyograo: Can you please capitalize Logstash?. Basically `logstash` -> to `Logstash`

docs/static/configuration.asciidoc
```diff
@@ -599,6 +599,199 @@ output {
...
+=== Using Environment Variables in Configuration
+== Overview
+
+You can set environment variable references into logstash plugins configuration using `${var}` or `$var`. 
+Each reference will be replaced by environment variable value at logstash startup.
+
+* The replacement is case-sensitive.
+* References to undefined variables raise a logstash configuration error.
+* A default value can be given by using the form `${var:default value}`.
+* You can add environment variable references in any plugin option type : string, number, boolean, array or hash.
```
suyograo: Can you also add that environment variables are immutable. If you update the environment variable, you'll have to restart Logstash to pick the updated value.

docs/static/configuration.asciidoc
```diff
@@ -599,6 +599,199 @@ output {
...
+
+== Examples
+
+[cols="asciidoc,asciidoc,asciidoc",options="header",grid="cols"]
+|=======
+|Logstash config source	|Environment 	|Logstash config result
+
+|
+name: $NAME|export NAME=elastic 	|name: elastic
```
suyograo: nice

docs/static/configuration.asciidoc
```diff
@@ -599,6 +599,199 @@ output {
...
+[source,shell]
+----
+export LS_HOME="/path"
+----
+|
+[source,ruby]
+----
+filter {
+  mutate {
+    add_field => { "ls_home" => "/path" }
```
suyograo: Love the examples @fbaligand

docs/static/configuration.asciidoc
fbaligand: Oups : the first table is a mistake : 
I copy the Beats examples, to be inspired in Logstash examples.
But then, I forget to drop Beats initial table.
=> I drop it.

fbaligand:
@suyograo 
I just fix up documentation :
- replace logstash by Logstash
- remove first table that was a bad copy/paste

fbaligand:
@suyograo
I think I will add your suggested tests tomorrow.

fbaligand:
I just pushed slight enhancements in configuration documentation.
Tell me if you're OK with that.

All documentation updates are squashed into only one commit.

fbaligand:
@suyograo 
I just added the following tests (which work fine) :
- `"key2" => "$FunString is ${FunBool}"`
- `"key3" => "${FunBool:false} or ${funbool:false}"`
- `"oneArray" => [ "first array value", "${notExistingVar:foo}", "${notExistingVar:}", "${notExistingVar: }", "${notExistingVar:foo bar}" ]`

logstash-core/lib/logstash/config/mixin.rb
jordansissel: It should be ok, in theory, if it's not, it probably just means we found a bug caused by something else within logstash going funny. :)

jordansissel:
LGTM

suyograo:
LGTM as well :) Thanks for this @fbaligand !

jordansissel:
:heart:

Thanks for working on this :)

fbaligand:
@suyograo @jordansissel 
It was great to work with you !
Happy to see that PR is now merged !
Thanks :)

thasmo:
Awesome! Just learned about logstash, set up my first instance via Docker and gonna deploy it in a second. Can't wait to have 2.3 released with this **great** feature.

fbaligand:
@thasmo 
Happy to see you appreciate this feature ! :)

PR #4713, opened 2016-02-23
Ensure only .conf files are loaded when logstash start

rhoml:
We had a bug in our instances where some automated task was creating a `Megasas.log` file in the current path of the user logged in.
This was causing logstash not to start and we discovered that logstash tries to load all files in `/etc/logstash/conf.d` it would be nice if it only tried to load .conf files so we can avoid this kind of issues. 
Even the puppet-logstash creates logstash.conf so is my believe that we should only load .conf files.


suyograo:
Fixed in #7130

PR #4714, opened 2016-02-23
Add ignore_older option and related tip.

raf64flo:
Add option to handle example file even if it is older than 86400s.


suyograo:
LGTM

PR #4717, opened 2016-02-24
Fix oops backtrace logging

andrewvc:
'Oops' backtrace logging doesn't seem to actually dump anything useful (it REALLY needs a backtrace given that it only happens in bizarre circumstances). It also seems that the backtrace isn't properly formatted (see https://github.com/logstash-plugins/logstash-output-rabbitmq/issues/41 ).

At any rate, this should use structured logging.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -86,7 +86,7 @@ class LogStash::Agent
         begin
           reload_pipeline!(pipeline_id)
         rescue => e
-          @logger.error I18n.t("oops", :error => e, :backtrace => e.backtrace)
+          @logger.error(I18n.t("oops"), :message => e.message, :class => e.class.name, :backtrace => e.backtrace)
         end
       end
     end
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/agent.rb
```diff
@@ -86,7 +86,7 @@ def reload_state!
         begin
           reload_pipeline!(pipeline_id)
         rescue => e
-          @logger.error I18n.t("oops", :error => e, :backtrace => e.backtrace)
+          @logger.error I18n.t("oops"), :message => e.message, :class => e.class.name, :backtrace => e.backtrace
```
suyograo: closed bracket?

suyograo: ah, was confusing. Can we add braces around `@logger.error( I18n.t("oops"), :message => `

suyograo:
minor issue with style, but LGTM

suyograo:
LGTM

PR #4718, opened 2016-02-24
Add ability to configure custom gc log file

siddharthasahu:
This adds ability to add gc log file to the custom $LS_LOG_DIR path


siddharthasahu:
Finished signing the CLA

PR #4725, opened 2016-02-26
[Docs] Add section on upgrading to 2.2

suyograo:


docs/static/upgrading.asciidoc
```diff
@@ -77,3 +77,20 @@ of workers by passing a command line flag such as:
...
+=== Upgrading Logstash to 2.2
+
+Logstash 2.2 re-architected the pipeline stages to provide more performance and help future enhancements in resiliency. 
+The new pipeline introduced micro-batches, processing groups of events at a time. The default batch size is 
+125 per worker. Also, the filter and output stages are executed in the same thread, but still, as different stages.
+The CLI flag `--pipeline-workers` or `-w` control the number of execution threads, which is set by default to number of cores.
+
+**Performance Tuning in 2.2**
+Since both filters and output workers are on the same thread, this could lead to threads being idle in I/O wait state. 
+Thus, in 2.2, you can safely set `-w` to a number which is 2x number of cores. A common way to tune performance is
```
andrewvc: I wouldn't mention a specific number. In some cases it may be OK to have 3x # of cores (though that would be an extreme).

docs/static/upgrading.asciidoc
```diff
@@ -77,3 +77,20 @@ of workers by passing a command line flag such as:
...
+Logstash 2.2 re-architected the pipeline stages to provide more performance and help future enhancements in resiliency. 
+The new pipeline introduced micro-batches, processing groups of events at a time. The default batch size is 
+125 per worker. Also, the filter and output stages are executed in the same thread, but still, as different stages.
+The CLI flag `--pipeline-workers` or `-w` control the number of execution threads, which is set by default to number of cores.
+
+**Performance Tuning in 2.2**
+Since both filters and output workers are on the same thread, this could lead to threads being idle in I/O wait state. 
+Thus, in 2.2, you can safely set `-w` to a number which is 2x number of cores. A common way to tune performance is 
+keep increasing the `-w` beyond the # of cores until performance no longer improves. A note of caution - make sure you also 
+keep heapsize in mind, because the number of in-flight events are `# of workers * batch_size`. More in-flight events
```
andrewvc: Might be worth completing the equation to be `#workers * batch_size * average_event size`

docs/static/upgrading.asciidoc
```diff
@@ -77,3 +77,20 @@ of workers by passing a command line flag such as:
 
 [source,shell]
 bin/logstash `-w 1`
+
+=== Upgrading Logstash to 2.2
+
+Logstash 2.2 re-architected the pipeline stages to provide more performance and help future enhancements in resiliency. 
+The new pipeline introduced micro-batches, processing groups of events at a time. The default batch size is 
+125 per worker. Also, the filter and output stages are executed in the same thread, but still, as different stages.
```
andrewvc: Also add 'this will by default also be the bulk size used for the elasticsearch output. The Elasticsearch output's `flush_size` now acts only as a maximum bulk size (still defaulting to 500).'

suyograo: Moved to a new section

suyograo:
@andrewvc how does it look now?

andrewvc:
We need a section stating:

```
The default number of output workers in Logstash 2.2 is now equal to the number of pipeline workers (-w) unless overridden in the Logstash config file. This can be problematic for some users as the extra workers may consume extra resources like file handles, especially in the case of the Elasticsearch output. Users with more than one Elasticsearch host may want to override the ‘workers’ setting for the Elasticsearch output in their Logstash config to constrain that number to a low value, such as 1-4. A [patch](https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/377) is forthcoming to remove the need for this in the Elasticsearch output worker.
```

before this is LGTM

andrewvc:
LGTM

PR #4727, opened 2016-02-28
WIP Java graph execution model

andrewvc:
This is a preliminary PR to discuss a java based graph execution model. The code is not production ready, but the results are very nice!

Some early results:

```bash
## -w 8 (on a 4 core box with HT)
# Java graph
~/p/logstash (native-graph) $ rm /tmp/graph-out.json; head -n 1000000 ~/projects/ls_apache_materials/apache_access_logs | time  bin/logstash -w 8 -b 250 -f conditional-apache.yml; wc -l /tmp/graph-out.json
Settings: User set pipeline workers: 8, Default pipeline workers: 8
Logstash startup completed
Logstash shutdown completed
stopping pipeline {:id=>"main", :level=>:warn}
       94.25 real       285.85 user        12.02 sys
 1000000 /tmp/graph-out.json

# Master
~/p/logstash-alt (fix_oops_backtrace_logging) $ rm -f /tmp/ls-out.json; head -n 1000000 ~/projects/ls_apache_materials/apache_access_logs | time bin/logstash -w 8 -b 250  -f ../logstash/conditional-apache.cfg; wc -l /tmp/ls-out.json
Settings: User set pipeline workers: 8, Default pipeline workers: 8
Logstash startup completed
Logstash shutdown completed
stopping pipeline {:id=>"main", :level=>:warn}
      120.53 real       527.25 user        22.15 sys
 1000000 /tmp/ls-out.json

## -w 1
# Java graph
~/p/logstash (native-graph) $ rm /tmp/graph-out.json; head -n 1000000 ~/projects/ls_apache_materials/apache_access_logs | time  bin/logstash -w 1 -b 250 -f conditional-apache.yml; wc -l /tmp/graph-out.json
Settings: User set pipeline workers: 1, Default pipeline workers: 8
Logstash startup completed
Logstash shutdown completed
stopping pipeline {:id=>"main", :level=>:warn}
      194.75 real       244.30 user        11.02 sys
 1000000 /tmp/graph-out.json

# Master
rm -f /tmp/ls-out.json; head -n 1000000 ~/projects/ls_apache_materials/apache_access_logs | time bin/logstash -w 1 -b 250  -f ../logstash/conditional-apache.cfg; wc -l /tmp/ls-out.json
Settings: User set pipeline workers: 1, Default pipeline workers: 8
Logstash startup completed
Logstash shutdown completed
stopping pipeline {:id=>"main", :level=>:warn}
      301.91 real       353.34 user        17.05 sys
 1000000 /tmp/ls-out.json
```

This benchmark uses a new graph IR which it sorts topologically then [traverses](https://github.com/andrewvc/logstash/blob/native-graph/logstash-core-event-java/src/main/java/com/logstash/pipeline/PipelineGraph.java#L33). Currently only outputs are batched since we execute the whole filter execution path as a single function, once per event. This new pipeline passes micro-batches to each filter letting filters potentially exploit the new batch nature of the pipeline (none do this yet).

The IR / standard config used here was: https://github.com/andrewvc/logstash/blob/native-graph/conditional-apache.yml / https://github.com/andrewvc/logstash/blob/native-graph/conditional-apache.cfg . The idea here is that we could just compile the current logstash config into this new IR using https://github.com/colinsurprenant/logstash-antlr-config

This new pipeline is written in Java. I saw similar (though not as impressive) results with a much hackier ruby version I wrote on the bus. The graph execution model is more likely than the Javaification to be responsible for most of the results. I decided that there was no point in writing a more fleshed out ruby version since we're moving to Java anyway.

Key points:
* New pipeline ~27% faster than current with -w 8, 50% faster than current with -w 1. 
* New pipeline 84% faster measured on CPU time with -w 8, only 44% faster on CPU time with -w 1. I suspect improvements to the batcher might fix this.
* New pipeline user time being 15% faster on -w 1 than -w 8, with wall time being 200% faster on -w 8 indicates that the pipeline wasn't saturated at -w 8. We need a better test harness that can more fully saturate workers. We should use this ratio to test pipeline saturation in future tests.
So, ~27% wall clock speedup for the apache pipeline + a 84% speedup in userspace.
Note that this speedup comes despite having a reasonably complex grok pattern + other CPU intensive filters (geoip, user agent) and a simple boolean condition.
* There are some differences in stats tracking with the new pipeline. But I doubt they account for a significant % of the performance
* We do an [additional sort](https://github.com/andrewvc/logstash/blob/native-graph/logstash-core-event-java/src/main/java/com/logstash/pipeline/PipelineGraph.java#L64) before each output with the new pipeline to maintain ordering in the -w 1 case. We should skip this with -w 2.

The code here ain't great and would need new tests. This is a WIP to solicit feedback.


andrewvc:
In hindsight I should have used a topological sort to determine the execution order. Will do that when I find the time

ph:
There is some merge artifacts in this PR, can you remove them?

andrewvc:
@ph you mean the Gemfile.lock? I just removed those things. I left the YAML in on purpose, since those are test files.

ph:
I mean 393  logstash-core/lib/logstash/agent_LOCAL_99722.rb


On Wed, Mar 2, 2016 at 3:59 PM, Andrew Cholakian <notifications@github.com>
wrote:

> @ph <https://github.com/ph> you mean the Gemfile.lock? I just removed
> those things. I left the YAML in on purpose, since those are test files.
>
> —
> <https://github.com/elastic/logstash/pull/4727#issuecomment-191427565>.
>



-- 
--
ph
Software Engineer

andrewvc:
Ah, thanks @ph, removed them!

suyograo:
Closing this in lieu of https://github.com/elastic/logstash/tree/lir

PR #4738, opened 2016-02-29
Fix order of `--configtest` arg

untergeek:
It should come before the `-f`, otherwise Logstash might bind a port and
start running, rather than only perform a configtest, then exit.

fixes 4737


suyograo:
LGTM

PR #4756, opened 2016-03-03
Revert #4425

suyograo:
#4425 introduced protection against removing `@timestamp` but there
 are a lot of tests that fail now. Reverting until we fix this properly


ph:
LGTM

breml:
@suyograo see https://github.com/elastic/logstash/pull/4425#issuecomment-192334356

PR #4759, opened 2016-03-04
Fix configtest option in init scripts

### LINKED ISSUE #4319: Specifying JMX parameters in LS_JAVA_OPTS causes Logstash not to restart properly
We added LS_JAVA_OPTS [here](https://www.elastic.co/blog/logstash-1-5-0-rc3-released).  When using the RPM, there are multiple places you can specify LS_JAVA_OPTS, eg. in the /etc/sysconfig/logstash and /etc/init.d/logstash files.  But it looks like the /etc/init.d/logstash is the right place to specify it.

Here is an example when seeing up a JMX connection, if you specify it in /etc/sysconfig/logstash file, this causes issues when using /etc/init.d/logstash restart - it will eventually end up launching multiple RMI servers and fail with a bind exception, eg.

```
sudo /etc/init.d/logstash restart
Configuration error. Not restarting. Re-run with configtest parameter for details

sudo /etc/init.d/logstash configtest 
WARNING: Default JAVA_OPTS will be overridden by the JAVA_OPTS defined in the environment. Environment JAVA_OPTS are -Djava.rmi.server.hostname=localhost -Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.port=9994 -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false 
Error: Exception thrown by the agent : java.rmi.server.ExportException: Port already in use: 9994; nested exception is: 
java.net.BindException: Address already in use
```

So I moved the LS_JAVA_OPTS settings for the JMX connection from /etc/sysconfig/logstash to /etc/init.d/logstash.  After that, I am no longer able to reproduce the restart issue.

Not sure if this is a bug. If not, it will be nice to document this better in our guide (and as a comment within the sysconfig file, etc..).

suyograo:
No need to export all the `JAVA*` related environment variables
for configtest. It needs a minimal set of variables to execute
`bin/logstash --config-test`. This was indirectly causing issues
when used with JMX options in `LS_JAVA_OPTS`. JMX needs a remote port defined
and `configtest` was trying to connect to the port twice which caused
restart (#4319) to fail

Fixes #4319

### DIFF
--- a/pkg/logstash.sysv
+++ b/pkg/logstash.sysv
@@ -142,7 +142,7 @@ configtest() {
   fi
 
   HOME=${LS_HOME}
-  export PATH HOME JAVA_OPTS LS_HEAP_SIZE LS_JAVA_OPTS LS_USE_GC_LOGGING
+  export PATH HOME
 
   test_args="--configtest -f ${LS_CONF_DIR} ${LS_OPTS}"
   $program ${test_args}


untergeek:
Such a splash from such a small oversight...

LGTM™

PR #4763, opened 2016-03-04
Fix threadsafe workers in output delegators to not instantiate extras

andrewvc:
There were a number of issues with the way OutputDelegator worked. The main issue is that it still created extra worker objects! This refactor significantly cleans that up and actually ensures only one worker is created in the threadsafe case

### DIFF
--- a/logstash-core/lib/logstash/output_delegator.rb
+++ b/logstash-core/lib/logstash/output_delegator.rb
@@ -1,5 +1,6 @@
 # encoding: utf-8
 require "concurrent/atomic/atomic_fixnum"
+java_import "java.util.concurrent.CopyOnWriteArrayList"
 
 # This class goes hand in hand with the pipeline to provide a pool of
 # free workers to be used by pipeline worker threads. The pool is
@@ -8,72 +9,27 @@ require "concurrent/atomic/atomic_fixnum"
 #
 # This plugin also records some basic statistics
 module LogStash class OutputDelegator
-  attr_reader :workers, :config, :worker_count, :threadsafe
+  attr_reader :workers, :config, :threadsafe
 
   # The *args this takes are the same format that a Outputs::Base takes. A list of hashes with parameters in them
   # Internally these just get merged together into a single hash
-  def initialize(logger, klass, default_worker_count, metric, *args)
+  def initialize(logger, klass, default_worker_count, metric, *plugin_args)
     @logger = logger
     @threadsafe = klass.threadsafe?
-    @config = args.reduce({}, :merge)
+    @config = plugin_args.reduce({}, :merge)
     @klass = klass
+    @workers = java.util.concurrent.CopyOnWriteArrayList.new
+    @default_worker_count = default_worker_count
+    @registered = false
 
     # Create an instance of the input so we can fetch the identifier
-    output = @klass.new(*args)
+    output = @klass.new(@config)
 
     # Scope the metrics to the plugin
     namespaced_metric = metric.namespace(output.plugin_unique_name.to_sym)
     @metric_events = namespaced_metric.namespace(:events)
 
-
-    # We define this as an array regardless of threadsafety
-    # to make reporting simpler, even though a threadsafe plugin will just have
-    # a single instance
-    #
-    # Older plugins invoke the instance method Outputs::Base#workers_not_supported
-    # To detect these we need an instance to be created first :()
-    # TODO: In the next major version after 2.x remove support for this
-    @workers = [@klass.new(*args)]
-    @workers.first.register # Needed in case register calls `workers_not_supported`
-
-    # DO NOT move this statement before the instantiation of the first single instance
-    # Read the note above to understand why
-    @worker_count = calculate_worker_count(default_worker_count)
-    @logger.debug("Will start workers for output", :worker_count => @worker_count, :class => klass)
-
-    warn_on_worker_override!
-    # This queue is used to manage sharing across threads
-    @worker_queue = SizedQueue.new(@worker_count)
-
-    @workers += (@worker_count - 1).times.map do
-      inst = @klass.new(*args)
-      inst.metric = @metric
-      inst.register
-      inst
-    end
-
-    @workers.each { |w| @worker_queue << w }
-
     @events_received = Concurrent::AtomicFixnum.new(0)
-
-
-    # One might wonder why we don't use something like
-    # define_singleton_method(:multi_receive, method(:threadsafe_multi_receive)
-    # and the answer is this is buggy on Jruby 1.7.x . It works 98% of the time!
-    # The other 2% you get weird errors about rebinding to the same object
-    # Until we switch to Jruby 9.x keep the define_singleton_method parts
-    # the way they are, with a block
-    # See https://github.com/jruby/jruby/issues/3582
-    if threadsafe?
-      @threadsafe_worker = @workers.first
-      define_singleton_method(:multi_receive) do |events|
-        threadsafe_multi_receive(events)
-      end
-    else
-      define_singleton_method(:multi_receive) do |events|
-        worker_multi_receive(events)
-      end
-    end
   end
 
   def threadsafe?
@@ -98,11 +54,14 @@ module LogStash class OutputDelegator
     @config["workers"] && @config["workers"] > 1 && @klass.workers_not_supported?
   end
 
-  def calculate_worker_count(default_worker_count)
+  def target_worker_count
+    # Remove in 5.0 after all plugins upgraded to use class level declarations
+    raise ArgumentError, "Attempted to detect target worker count before instantiating a worker to test for legacy workers_not_supported!" if @workers.size == 0
+
     if @threadsafe || @klass.workers_not_supported?
       1
     else
-      @config["workers"] || default_worker_count
+      @config["workers"] || @default_worker_count
     end
   end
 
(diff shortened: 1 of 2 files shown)


suyograo:
@jsvd @jordansissel  review please?

suyograo:
@andrewvc rebase? 

@jsvd review ping

andrewvc:
@suyograo @jsvd just rebased. This is really important to get in to 2.3 and is required for all the threadsafety work on the ES output / ruby client.

andrewvc:
@suyograo @jsvd tests now pass!

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -8,72 +9,27 @@
...
   # The *args this takes are the same format that a Outputs::Base takes. A list of hashes with parameters in them
   # Internally these just get merged together into a single hash
-  def initialize(logger, klass, default_worker_count, metric, *args)
+  def initialize(logger, klass, default_worker_count, metric, *plugin_args)
     @logger = logger
     @threadsafe = klass.threadsafe?
-    @config = args.reduce({}, :merge)
+    @config = plugin_args.reduce({}, :merge)
     @klass = klass
+    @workers = java.util.concurrent.CopyOnWriteArrayList.new
```
purbon: not sure how do we feel about actually having pure java code inside our ruby side? would it not be better to actually wrap it somehow?

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -111,7 +70,59 @@ def config_name
...
+    @registered = true
+    # We define this as an array regardless of threadsafety
+    # to make reporting simpler, even though a threadsafe plugin will just have
+    # a single instance
+    #
+    # Older plugins invoke the instance method Outputs::Base#workers_not_supported
+    # To detect these we need an instance to be created first :()
+    # TODO: In the next major version after 2.x remove support for this
+    @workers << @klass.new(@config)
+    @workers.first.register # Needed in case register calls `workers_not_supported`
```
purbon: not sure I understand why we call register only to the first element of @workers, but @workers is a global array and this methods can only be called once. isn't this register call not going to happen always regarder of the way the output plugin works?

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -111,7 +70,59 @@ def config_name
...
+
+    @logger.debug("Will start workers for output", :worker_count => target_worker_count, :class => @klass)
+
+    # Threadsafe versions don't need additional workers
+    setup_additional_workers!(target_worker_count) unless @threadsafe
+    # We skip the first worker because that's pre-registered to deal with legacy workers_not_supported
+    @workers.subList(1,@workers.size).each(&:register)
```
purbon: don't the difference between this line and `83`, I guess we could directly check for workers not supported, what do you think?

on the other side, here goes one of my concerts with _non ruby_ code mixup, here you use a java method in between of ruby stuff :tada:

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -111,7 +70,59 @@ def config_name
...
+    setup_multi_receive!
+  end
+
+  def setup_additional_workers!(target_worker_count)
```
purbon: Is this a destructive method? I see you commonly about the `!` methods in a lot of places :smile_cat:

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -111,7 +70,59 @@ def config_name
...
+      inst.metric = @metric
+      @workers << inst
+    end
+
+    # This queue is used to manage sharing across threads
+    @worker_queue = SizedQueue.new(target_worker_count)
+    @workers.each {|w| @worker_queue << w }
+  end
+
+  def setup_multi_receive!
```
purbon: same thing here.

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -153,10 +168,14 @@ def busy_workers
     if @threadsafe
       0
     else
-      @workers.size - @worker_queue.size
+      @workers.size - (@worker_queue ? @worker_queue.size : 0)
```
purbon: curious to know a couple of things here:

* why not using `worker_count` like the method you defined in next lines?
* also, will not `@worker_queue.size` return zero when zero? why do you need the if here?

thanks!

logstash-core/spec/logstash/output_delegator_spec.rb
```diff
@@ -31,6 +31,7 @@
 
     context "after having received a batch of events" do
       before do
```
purbon: Is a matter of style, but I do recommend using explicit params here, relaying in defaults it actually makes me think all time which was the default :+1:

logstash-core/spec/logstash/output_delegator_spec.rb
```diff
@@ -80,6 +88,7 @@
           allow(out_inst).to receive(:metric=).with(any_args)
           allow(out_inst).to receive(:id).and_return("a-simple-plugin")
           allow(out_klass).to receive(:workers_not_supported?).and_return(false)
+          subject.register
```
purbon: not sure I understand the duplication of some of this `allow` and `register` calls here, can you elaborate a bit more?

logstash-core/spec/logstash/output_delegator_spec.rb
```diff
@@ -96,14 +105,18 @@
         end
 
         it "should not utilize the worker queue" do
-          expect(subject.send(:worker_queue)).not_to receive(:pop)
-          subject.multi_receive(events)
+          expect(subject.send(:worker_queue)).to be_nil
```
purbon: I don't see any method call here? jut retrieving the queue? is this only checking for initialization?

logstash-core/lib/logstash/output_delegator.rb
jsvd: you can ways use the snake case version of the methods, e.g. @workers.sub_list(1, ..)

logstash-core/lib/logstash/output_delegator.rb
andrewvc: This is kind of crappy, but its required for backwards compat. The only way we know if a worker supports workers or not is if its register method calls 'workers_not_supported'. So, to determine if we should make multiple workers we need to: 

1. make a single worker and invoke register
1. Check if it invoked `workers_not_supported`
1. Start additional workers if so.

It sucks, and we should deprecate it in favore of [`declare_workers_supported!`](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/outputs/base.rb#L39) which is done at the class level

logstash-core/lib/logstash/output_delegator.rb
andrewvc: See the comment above for why.

@jsvd I prefer not to use snake case for Java objects because it keeps the delineation between ruby and java objects clear.

logstash-core/lib/logstash/output_delegator.rb
andrewvc: I've always seen `!`` as being about side effects not destruction, as in `create!` in rails.

logstash-core/lib/logstash/output_delegator.rb
andrewvc: I don't see the value since using it is so easy. We no longer support MRI, and I think a wrapper just means a larger area we need to test and more places things can go wrong.

logstash-core/lib/logstash/output_delegator.rb
colinsurprenant: agree, `!` signals side-effects in the method.

logstash-core/lib/logstash/output_delegator.rb
andrewvc: I actually don't remember why I made this change. `@worker_queue` should always be defined so I'm wondering if there's a race here? At any rate I'll investigate this.

You're right, this should use `worker_count` as well.

logstash-core/spec/logstash/output_delegator_spec.rb
andrewvc: Not sure I see the duplication.

logstash-core/spec/logstash/output_delegator_spec.rb
andrewvc: This is expecting there to not be a worker queue defined if we're using a threadsafe worker.

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -138,9 +149,13 @@ def worker_multi_receive(events)
   def do_close
     @logger.debug("closing output delegator", :klass => self)
 
-    @worker_count.times do
-      worker = @worker_queue.pop
-      worker.do_close
+    if @threadsafe
+      @workers.each(&:do_close)
+    else
+      worker_count.times do
```
jsvd: is it possible that there is one or more workers popped from the list when do_close is called?
If threadsafe is false, `worker_count` is called to know how many pops/do_close we need to do. but worker_count computes @workers.size on each call

jsvd:
my only concern with this is https://github.com/elastic/logstash/pull/4763/files#r57338275.
that aside I'm running all the tests with the default plugins, will report back with the result

logstash-core/lib/logstash/output_delegator.rb
andrewvc: This is not possible because the pipeline does not start closing outputs or filters until all worker threads are stopped. https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/pipeline.rb#L442

logstash-core/lib/logstash/output_delegator.rb
andrewvc: Ah, I found the reason for this. The pipeline reporter can run before the outputs are registered trying to pull a value here. In that case @worker_queue is empty, we just return 0. I'll add a comment here so it isn't confusing.

andrewvc:
@jsvd @purbon I believe I have addressed all your concerns with my latest push, please let me know if you have further questions.

jsvd:
LGTM

PR #4764, opened 2016-03-04
improve pipeline start and stop logging

### LINKED ISSUE #4728: Change shutdown verbiage for config reload feature
With `--auto-reload` enabled, LS can dynamically apply config changes made to the file. The message emitted when a new config applies is misleading. Today it says:
```
Logstash shutdown completed
```

should be changed to 

```
Config change detected, pipeline <ID> has been shutdown
```

On startup of the pipeline, it should be:
```
New pipleine <ID> started with latest config
```

Also, I don't see any point in logging the new updated config . For users with large configs, this is going to be a lot of noise. Lets move this message to debug:

```
fetched new config for pipeline. upgrading .. {:pipeline => main .... 
```

jsvd:
fixes #4728

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -167,8 +167,7 @@ module LogStash; class Pipeline
 
     start_workers
 
-    @logger.info("Pipeline started")
-    @logger.terminal("Logstash startup completed")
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

andrewvc:
LGTM

PR #4770, opened 2016-03-06
Fixed an issue with accidental multiple instantiation of plugins

Ulrhol:
I found this issue while fixing the websocket output plugin. It seems most plugins are accidentally instantiated more times than expected. This happens regardless of how you define your # of workers, if you define "workers_not_supported" or if you use the 2.2+ API with "declare_workers_not_supported".


suyograo:
@andrewvc can you take a look at this?

suyograo:
oh, looks like https://github.com/elastic/logstash/pull/4763 fixes this.

Ulrhol:
Agree. #4763 seems to be solving the same issue but in a more appropriate way.

suyograo:
Ok, closing this. Thanks @Ulrhol

Issue #4651, opened 2016-02-09
Clarify offline plugin instructions from plugin command


ppf2:
To reproduce, create 2 separate LS 2.1 directories, one for the logstash-2.1.0_source (with internet access to prepare the offline package), the other logstash-2.1.0_target (to unpack the offline package).

First prepare the offline package in logstash-2.1.0_source:

```
./plugin update logstash-input-kafka 
./plugin update logstash-output-elaticsearch 
./plugin pack 
```

In the logstash-2.1.0_target location:

```
./plugin list --verbose |grep kafka
logstash-input-kafka (2.0.2)
logstash-output-kafka (2.0.1)
```

```
./plugin unpack ../../logstash-2.1.0_source/plugins_package.tar.gz 
Unpacking ../../logstash-2.1.0_source/plugins_package.tar.gz
Unpacked at /Users/test_user/Downloads/testing/logstash-2.1.0_target/vendor/cache
The unpacked plugins can now be installed in local-only mode using bin/plugin install --local [plugin name]
```

See how it tells you to use the `install --local plugin_name` option for the plugin command.

If you try this (now with internet access disabled), you will see that the plugin install using the local cache will fail:

```
./plugin install --local logstash-input-kafka
Installing logstash-input-kafka
Error Bundler::HTTPError, retrying 1/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 2/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 3/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 4/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 5/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 6/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 7/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 8/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 9/10
Could not fetch specs from https://rubygems.org/
Error Bundler::HTTPError, retrying 10/10
Could not fetch specs from https://rubygems.org/
Too many retries, aborting, caused by Bundler::HTTPError
ERROR: Installation Aborted, message: Could not fetch specs from https://rubygems.org/
```

Same thing if you use the update command:

```
[...]

PR #4779, opened 2016-03-09
Internal gem cache rebuild and improved experience when doing updates.

purbon:
As we clean up the internal ruby gems cache, but keep the *.gemspec files around, the pack command gets confused expecting to have gem files it really does not have. This causes the usage of offline installation and update to be broken when there is a dependency mismatch.

This situation is raised when the user does, while working with an old version (not necessary too old, but that uses old dependencies), see #4651 for details:
- Run `bin/plugin update [plugin name]`. < could be with more than one plugin, or just one>
- Run `bin/plugin pack`

If for example plugin name depends on A version 1.0, but the new version actually can use a new version, lets say 2.0, the package will contain only 2.0 as is the _current state_ in the LS used to create the bundle. This will cause problems when using this package in a raw LS, there bundler might think it needs 1.0 version.  This might also happen if 2 or more plugins are updated, due to internal cache checkups done in Bundler, see https://github.com/bundler/bundler/blob/master/lib/bundler/cli/update.rb#L53 for more details.

To address this situation I recommend, for now and without disregarding feature improvements, to:
- Create an initial pack before doing the update. Name this file something like `initial_plugin_package.tar.gz`, so there are no override in the feature.
- Update the desired plugins.
- Create a new package, that will include the new plugins.
- Merge the initial and the new package into a single package.

The merged package will include old and new dependencies, making sure _all_ required dependencies are available to the offline environment. Complementary improvements in the document should be made, so this is known. This is useful if you only want to update 1 plugin, but got more updated in your package.

There is also the situation when you know and control which plugins are _new_ in your package for offline installation, then I recommend to use `bin/plugin update --local` as this will make sure bundler does the right internal dependency resolution.

This PR also makes sure errors for missing gems in case of doing an offline install are more clear, like this users could be aware of what is missing inside their `vendor/cache` directories. 

What do you think?

I tried my best to describe the situation here, but please don't hesitate to ask for clarification, I know the process is a bit tricky.

Related #4651

### DIFF
--- a/lib/bootstrap/bundler.rb
+++ b/lib/bootstrap/bundler.rb
@@ -28,18 +28,31 @@ module LogStash
         end
       end
 
-      # This patch makes rubygems fetch directly from the remote servers
-      # the dependencies he need and might not have downloaded in a local
-      # repository. This basically enabled the offline feature to work as
-      # we remove the gems from the vendor directory before packacing.
-      ::Bundler::Source::Rubygems.module_exec do
-        def cached_gem(spec)
-          cached_built_in_gem(spec)
+      ::Bundler::Runtime.module_exec do
+        def rebuild_gem_cache
+          specs.each do |spec|
+            next if spec.name == "bundler"
+            next if spec.source.is_a?(::Bundler::Source::Path)
+            spec.source.send(:cached_built_in_gem, spec) if spec.source.respond_to?(:cached_built_in_gem)
+          end
         end
       end
+
     end
 
-    def setup!(options = {})
+    # This basically enabled the offline feature to work as
+    # we remove the gems from the vendor directory before packacing, so we need
+    # to fetch them back before moving forward.
+    def rebuild_gem_cache(options={})
+      configure!(options)
+      ::Bundler.load.rebuild_gem_cache
+    end
+
+
+    # Configure the minimum required set of options for Bundler like GEM_HOME location,
+    # GEMFILE_PATH, etc. It also apply internal patches and makes sure it setup to work
+    # as expected for internal LS usage.
+    def configure!(options={})
       options = {:without => [:development]}.merge(options)
       options[:without] = Array(options[:without])
 
@@ -61,6 +74,10 @@ module LogStash
       ::Bundler.settings[:gemfile] = Environment::GEMFILE_PATH
 
       ::Bundler.reset!
+    end
+
+    def setup!(options = {})
+      configure!(options)
       ::Bundler.setup
     end
 
--- a/lib/pluginmanager/pack.rb
+++ b/lib/pluginmanager/pack.rb
@@ -11,6 +11,8 @@ class LogStash::PluginManager::Pack < LogStash::PluginManager::PackCommand
     puts("Packaging plugins for offline usage")
 
     validate_target_file
+
+    LogStash::Bundler.rebuild_gem_cache
     LogStash::Bundler.invoke!({:package => true, :all => true})
     archive_manager.compress(LogStash::Environment::CACHE_PATH, target_file)
     FileUtils.rm_rf(LogStash::Environment::CACHE_PATH) if clean?


ph:
@purbon took the time this morning to run me through the situation of bundler and multiples version of gem in an offline installation context. 

let me just rephrase the problem for personal notes.

- We ship Logstash without any gem to obviously saves space in the packages.
- Logstash ship with a *.lock with frozen dependencies.
- When we run `bin/plugin update`, the lock is updated to reflect the new versions of the dependencies.
- We generate the **package.tar.gz** with the **newer** gems versions.
- On the offline logstash everything is fine if we update all the plugins since we can find all the *.gem in the cache.
- It doesn't work when we try to update a specific plugin, bundler will try to find all the dependencies and wont be able to find the `*.gem` for the gems shipped with the original package.

I see two possible solution for this problem.
- Hack bundler to relax the gem file lookup 
- Make sure the gem version are present all the time. (shipped gem and updated gems, this is the solution that @purbon has implemented.)

I think his suggestion is the less evil thing we could possible do right now to improve the less desirable situation for people behind firewalls. But this could lead to another problem, if the logstash version of the target installation vs the version of the packaging logstash are different it will.

I believe we should add a metadata file to the plugins pack to add the version of the logstash that was used to create it. The target could read that file before doing the unpack to know if the package would be compatible.

purbon:
Hi,

  you described it very good, just a minor correction.

On Wed, Mar 9, 2016 at 3:46 PM Pier-Hugues Pellerin <

>
>
> I see two possible solution for this problem.
>
>    - Hack bundler to relax the gem file lookup
>    - Make sure the gem version are present all the time. (shipped gem and
>    updated gems, this is the solution that @purbon
>    <https://github.com/purbon> has implemented.)
>
> I think his suggestion is the less evil thing we could possible do right
> now to improve the less desirable situation for people behind firewalls.
>

I made sure that logstash complains explicitly when something is not in the
internal cache, the procedure to have all gem versions present all the time
end up being responsability of the user (we should add this to
documentation). There is actually no way to keep old versions always, as
they get cleanup from the env as soon as possible (as it makes sense).

- purbon

ph:
I am +1 to explicitly communicate that to the user, but still think we should check the logstash version of the creator of the package agains't the version of installation that use it.

purbon:
What are the benefits of doing this? or just letting bundler check them for
us? I mean by the end of the day it will check it himself, right?

ph:
@purbon True, bundler will only complain about the version it cannot find, not its was created in an incompatible version of logstash. 

I think you are right, might not be completely needed.  I will bring it up in our bundler sync.

colinsurprenant:
1. For reference, [this is where](https://github.com/elastic/logstash/blob/f61e9a8cc5ee58e93a0a16b072c3754949598be6/rakelib/artifacts.rake#L25) the `*.gem` file are discarded from the artifact creation.

2. I think that what we do in (1) made sense from a package size perspective before we decided to support offline plugin installation but now it does not make sense because it creates a *state* problem in Bundler. 

3. I will argue that the **real** problem here is our default plugins pre-installation & packaging strategy. the idea of providing a readily available default plugins set makes sense from a getting-started perspective but I think we should revisit how we could achieve this without disrupting the installed plugins *state*.

4. One idea would be that the `bin/plugin pack` and `bin/plugin unpack` commands freezes and restores the **complete** plugins state, including `Gemfile` and `Gemfile.jruby-1.9.lock` and this would allows just running logstash on the target system exactly as it was on the source system without having to do any `bin/plugin install` or `bin/plugin update` command after the  `bin/plugin unpack`. We would just document that `bin/plugin install` or `bin/plugin update` require connectivity. I don't thinks it actually *make sense* to support `bin/plugin install` or `bin/plugin update` offline!

5. Similar to (4), instead of pre installing plugins and deleting the `*.gem` files at packaging time, we could  ship the core logstash package without any plugins or just a **very limited set** of core plugins **including** their `.gem` files and provide a few additional plugins packages that can be downloaded and  installed the same way as (4). Obviously one of these additional package would be the infamous "default" plugins package that would be pretty big since it would also include all the `.gem` files but since its an optional and non essential download, it would be acceptable **and** we would also provide other smaller plugins combinations packages like a "getting started" that would cover the courses and workshop, etc.

6. by doing this we would:
  - stop confusing bundler state and avoid more dangerous and **non** future-proof monkey patching
  - move forward with the plugin-based vision of **not** having to pre-package all plugins with the base logstash package
  - still provide the facility for someone to download and easily install the whole default plugins package which would be a bigger file but with a very simple installation since it would just involve copying files at the right place and not call any bundler install/update.

Thoughts?

purbon:
Moin moin,
   thanks for your thoughts Colin, I will try to provide some feedback on the topics, for the reference.

> 1.- For reference, this is where the *.gem file are discarded from the artifact creation.

Yes, this is basically cleaning up the internal Bundler.app_cache,  https://github.com/bundler/bundler/blob/2d7d9c8b4404a74fbb4baacc1cee809318f3123e/lib/bundler.rb#L183, used in many parts of the code base like https://github.com/bundler/bundler/blob/master/lib/bundler/source/rubygems.rb#L24 to actually determine the main bundler cache of gems.

> 2.- I think that what we do in (1) made sense from a package size perspective before we decided to support offline plugin installation but now it does not make sense because it creates a state problem in Bundler.

Agree, just want to make sure one thing, as you said many times, we must follow the bundler contract as close as possible, and in this case we're "hacking" bundler. For the record, I found fundler to have tendency to cause pain if you mess with this pre conditions, I would like us to stick closer to it as much as  possible.

> 3.- I will argue that the real problem here is our default plugins pre-installation & packaging strategy. the idea of providing a readily available default plugins set makes sense from a getting-started perspective but I think we should revisit how we could achieve this without disrupting the installed plugins state.

Yes and no, the preinstalled idea is not bad, but we also want small packages, so we can not get boths. To fix the later, we went the route of removing the `*.gem` files that makes bundler confused somehow. The problem for me is that we can not archive both and still make use of the package feature for offline installations. 

To fix that, I think we agree on the idea of proving logstash in some flavors, with different sized and including all `.gem` files in the internal cache, so we keep bundler happy.

> 4.- One idea would be that the bin/plugin pack and bin/plugin unpack commands freezes and restores the complete plugins state, including Gemfile and Gemfile.jruby-1.9.lock and this would allows just running logstash on the target system exactly as it was on the source system without having to do any bin/plugin install or bin/plugin update command after the bin/plugin unpack. We would just document that bin/plugin install or bin/plugin update require connectivity. I don't thinks it actually make sense to support bin/plugin install or bin/plugin update offline!

If i understand you properly, this is equivalent of just going `bin/plugin update --local` where bundler is actually not going to complain and be happy (_and update everything_). Keep in mind this only happen when installing a single plugin. Some people, like the ones that started this ticket, does not want to update everything.

> 5.- Similar to (4), instead of pre installing plugins and deleting the *.gem files at packaging time, we could ship the core logstash package without any plugins or just a very limited set of core plugins including their .gem 

+1 on that, I totally think is the way to go to reduce problems.


> 5.-  files and provide a few additional plugins packages that can be downloaded and installed the same way as (4).

I don't thinks (4) is actually necessary, the only thing that happen here is to make sure all dependencies are always available, and by providing them in the packages we don't have to do this.

> 5.- Obviously one of these additional package would be the infamous "default" plugins package that would be pretty big since it would also include all the .gem files but since its an optional and non essential download, it would be acceptable and we would also provide other smaller plugins combinations packages like a "getting started" that would cover the courses and workshop, etc.

we could also make topic related packages, like for file logs, for network, etc...

And __related to this PR proposal__:

This basically removes the monkey patch, so just add an independent method that lets rebuild the internal cache, before it was doing through patching. This makes the error explicit when something is missing, so the installation is stopped and the user knows it needs to add it.

Sorry for the long reply, just trying to provide context and answers :+1:

colinsurprenant:
> If i understand you properly, this is equivalent of just going bin/plugin update --local where bundler is actually not going to complain and be happy (and update everything).

What I am saying is that there would be no more install or update necessary - you create a plugins installation state on one machine with connectivity and you copy over the other machine the complete state : the whole `vendor/bundle`, the `Gemfile` and the `Gemfile.jruby-1.9.lock`. 

But as you mention, we would loose the possibility to update/install individual plugins, you would just restart logstash with the new bundler state, as a whole.

colinsurprenant:
With such a pack, there would be no installing or updating individual plugins, to update or install a new plugin, the pack plugins state could be (recreated and) moved back to another connected machine, unpacked, perform new plugin install or update, repacked and moved back to the disconnected machine. 

Let me sequence the new workflow under that idea:

- On connected machine with logstash
  - install, update, remove plugins to satisfaction
  - run `bin/plugin pack`
  - copy pack file to disconnected machine

- On disconnected machine with logstash
  - stop logstash
  - run `bin/plugin unpack`
  - start logstash
  - now disconnected logstash run exactly with same set of plugins as connected logstash

Now if you need to update a single plugins. there are 2 possibilities:

1. You still have the disconnected logstash pack file around:
  - move disconnected pack to a connected logstash
  - run `bin/plugin unpack`
  - install, update, remove plugins to satisfaction
  - run `bin/plugin pack`
  - copy pack file to disconnected machine

1. You do **not** have the disconnected logstash pack file around:
  - run `bin/plugin pack` on disconnected logstash
  - move disconnected pack to a connected logstash
  - run `bin/plugin unpack`
  - install, update, remove plugins to satisfaction
  - run `bin/plugin pack`
  - copy pack file to disconnected machine

purbon:
@colinsurprenant I agree this approach will solve some issues, is basically what I said, the same as doing now `bin/plugin update --local`, although not sure people will like the idea of updating everything at once, I see people more willing to have a notation like:

-> `bin/plugin update foo`

I know it might sound strange, as you achieve the same thing, what do you think?

- purbon

ph:
> One idea would be that the bin/plugin pack and bin/plugin unpack commands freezes and restores the complete plugins state, including Gemfile and Gemfile.jruby-1.9.lock and this would allows just running logstash on the target system exactly as it was on the source system without having to do any bin/plugin install or bin/plugin update command after the bin/plugin unpack.

I think this is a really interesting takes, this would simplify a lot of things code and management wise. IMHO. 

> We would just document that bin/plugin install or bin/plugin update require connectivity. I don't thinks it actually make sense to support bin/plugin install or bin/plugin update offline!

I think people will be find with that. We have two use case of installation:

- close infra without internet access, `bin/plugin install don't work` they will need to `pack/unpack` we could mark the installation as remotely managed.
- Behind firewalls with proxy access `bin/plugin install work`

@purbon `bin/plugin update foo` this will work on the system used to package the gems, the target system will only use one command which is the `unpack` command.

ph:
Also for an automation stand point, this would be easier to manage for and admin.

- Update the pack
- push the pack
- unpack (pick up any updated plugins)
- restart logstash

purbon:
On Thu, Mar 10, 2016 at 4:23 PM Pier-Hugues Pellerin <

>
>
> @purbon <https://github.com/purbon> bin/plugin update foo this will work
> on the system used to package the gems, the target system will only use one
> command which is the unpack command.
>

Yes, my fear is with not having update in the offline situation, but if you
guys are ok with it :-P, I would aim for having the same interface for all
situations (as we can archive this if we include all gems internally).

- purbon

purbon:
For this I'm +1, but we can already do that with

bin/plugin update --local

no need to change anything.

- purbon

ph:
I think we should make the `bin/plugin install/update` on the target system yield that the dependencies are remotely managed? WDYT?

purbon:
what do you mean by remotely managed?
- purbon

ph:
Target logstash has no internet access, the logtash with internet access remotely manages the target logstash dependencies. *Remotely managed* might not be the correct wording here.

But maybe this is not needed, I though it might be great to give the feedback to the user.

purbon:
I all +1 to provide more clear and extensive feedback to user, part of this
PR is doing this.

PD: We're having discussion on a PR while we should be having this in the
issue :-P

- purbon

colinsurprenant:
I moved this idea discussion in #4787, sorry for the duplication but I just realized that we were having this discussion in a PR which derails the original PR intent. This discussion deserves a separate issue.

seang-es:
Can this also get a backport to 2.x when this is complete?  Thanks!

ph:
@seang-es I dont see any reason why it would prevent this PR to be backported to 2.X

purbon:
I agree with @ph, end of the day this helps the offline experience by providing clever error message.

@ph can we move forward with this PR changes? it would be nice to have them so the error communication is clear to the end user, what do you think?

ph:
@purbon Did it was rebased with the current master?

I have tried the following:
- create an artifact with `rake artifact:tar
- extract this tarball and use `bin/plugin pack` and it failed when trying to packaige the `logstash-core-event-java`, probably because its only available on rubygems. We might need some special code to handle the case of unreleased gems.

**Error log**
```
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Error Bundler::InstallError, retrying 2/10
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Error Bundler::InstallError, retrying 3/10
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Error Bundler::InstallError, retrying 4/10
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Error Bundler::InstallError, retrying 5/10
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Error Bundler::InstallError, retrying 6/10
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Error Bundler::InstallError, retrying 7/10
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Error Bundler::InstallError, retrying 8/10
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Error Bundler::InstallError, retrying 9/10
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Error Bundler::InstallError, retrying 10/10
An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
Too many retries, aborting, caused by Bundler::InstallError
Bundler::InstallError: An error occurred while installing logstash-core-event-java (3.0.0.dev), and Bundler cannot continue.
Make sure that `gem install logstash-core-event-java -v '3.0.0.dev'` succeeds before bundling.
  install_gem_from_spec at /tmp/logstash-3.0.0.dev/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/installer.rb:145
   install_sequentially at /tmp/logstash-3.0.0.dev/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/installer.rb:269
                   each at /private/tmp/logstash-3.0.0.dev/vendor/jruby/lib/ruby/1.9/forwardable.rb:201
   install_sequentially at /tmp/logstash-3.0.0.dev/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/installer.rb:268
```

ph:
This PR was replaced with #6404

PR #4780, opened 2016-03-09
Make sure jvm info in _stats and _node/stats follow the same format

purbon:
Small PR to improve data consistency change to make sure all jvm info follow the same format in `_stats` or `_node/stats`.

Please don't focus too much on test, this will be improve with #4667, I mean I now they are bad right now :+1:


ph:
LGTM

Issue #4674, opened 2016-02-11
Bad file descriptor (Errno::EBADF) error when shutting down the web api webserver 


purbon:
Sometimes when shutting down logstash puma shows this error:

```
Error in reactor loop escaped: Bad file descriptor - Bad file descriptor (Errno::EBADF)
org/jruby/RubyIO.java:3681:in `select'
/Users/purbon/work/logstash/vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/reactor.rb:29:in `run_internal'
/Users/purbon/work/logstash/vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/reactor.rb:138:in `run_in_thread'
```

is not something causing problem in runtime, but should be investigated and avoided.

purbon:
For now I'm unable to reproduce this, something might have changed internally that made this error go away. @ph @suyograo If you see this error, please let me know, but for now I think is just gone :+1:

purbon:
I just deleted my previous comment, I can see this from running full integrated test:core as of #4667

purbon:
A fix for this goes into #4782

PR #4782, opened 2016-03-09
Fix webserver issue + Improve WebAPI testing

purbon:
This PR provides a solution for:

* #4674 by making sure puma is only handling stuff it can handle, like this in the shutdown face there are no issues raised. This change can be removed when using another kind of interface, but is necessary for using the current one.

* Make our path toward #4667 easier as test are more properly written and structured, this mostly let's us test with direct real life data out of the current pipeline, so we verify data is properly composed and shown.

what do you think?


purbon:
please jenkins, test this.

rakelib/default_plugins.rb
```diff
@@ -119,6 +119,8 @@ module RakeLib
       logstash-input-stdin
       logstash-input-tcp
       logstash-output-stdout
+      logstash-output-null
+      logstash-codec-rubydebug
```
ph: rubydebug? we need that fore core testing?

purbon: not really, good catch.

logstash-core/spec/api/lib/api/node_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    context "when asking for human output" do
+      before(:all) do
+        get "/hot_threads?human"
+      end
+
+      let(:payload) { last_response.body }
+
+      it "should return a plain text description" do
+        expect{ JSON.parse(payload) }.to raise_error
+      end
```
ph: should we check the header to make sure we send plain text?

logstash-core/spec/api/lib/api/node_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    context "when requesting idle threads" do
+
+      before(:all) do
+        get "/hot_threads?ignore_idle_threads=false&threads=10"
+      end
+
+      let(:payload) { JSON.parse(last_response.body) }
```
ph: I am starting to be a big fan of subject's name feature.
I think case this would be `subject(:payload)`, but this is just syntactic sugar since under the hood it uses `let`

logstash-core/spec/api/lib/api/node_spec.rb
purbon: makes sense to check both.

logstash-core/spec/api/lib/api/node_spec.rb
purbon: well, I agree, but not for this case. subject of this test is not the payload, but the small sinatra app.

rakelib/default_plugins.rb
purbon: @ph interesting, I removed the rubydebug and test hangs :-1:, might this be due to using the custom pipeline as in the spec_helper?

logstash-core/spec/api/lib/api/node_spec.rb
ph: We might wand to check for json to, an rspec matcher would make it easy to test WDYT?

logstash-core/spec/api/lib/api/node_spec.rb
ph: Actually we test the content of the payload? 
https://github.com/elastic/logstash/pull/4782/files#diff-1ce42ef623c4bfed70c7f5e063dd47fcR62

logstash-core/lib/logstash/webserver.rb
```diff
@@ -24,7 +24,12 @@ def initialize(logger, options={})
       @options     = {}
       @cli_options = options.merge({ :rackup => ::File.join(::File.dirname(__FILE__), "api", "init.ru"),
                                      :binds => ["tcp://#{http_host}:#{http_port}"],
-                                     :debug => logger.debug? })
+                                     :debug => logger.debug?,
+                                     # Prevent puma from queueing request when not able to properly handling them,
+                                     # fixed https://github.com/elastic/logstash/issues/4674. See
+                                     # https://github.com/puma/puma/pull/640 for mode internal details in PUMA.
+                                     :queue_requests => false
```
ph: Thanks for adding context.

logstash-core/spec/api/lib/api/node_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+require "sinatra"
+require "app/modules/node"
+
+describe LogStash::Api::Node do
+
+  include Rack::Test::Methods
+
+  def app()
+    described_class
+  end
```
ph: defining `#app` is for the `Rack::Test::Methods`? correct?

logstash-core/spec/api/lib/api/node_stats_spec.rb
```diff
@@ -11,24 +11,57 @@ def app()
...
+      expect(last_response).to be_ok
+    end
 
-  let(:events) do
-    { :in => 10, :out => 20 }
+    ["events", "jvm", "start_time_in_millis"].each do |key|
+      it "contains #{key} information" do
+        expect(payload).to include(key)
+      end
+    end
```
ph: +1 for this

rakelib/default_plugins.rb
ph: @purbon I dont believe to be the source of the problem, since the custom pipeline in your test is not using the rubydebug at all? So I am not sure where its used?

ph:
Did a review added minor comments concerning `ruby-debug` adress theses and it will be ready to merge.

logstash-core/spec/api/lib/api/node_spec.rb
purbon: yes, this is how Rack::Test mockup runs the internal app.

logstash-core/lib/logstash/webserver.rb
purbon: Man, context is much need here, otherwise in 2 weeks I promise I forget why we added :+1:

logstash-core/spec/api/lib/api/node_spec.rb
purbon: we check the content of the payload, but test how the module (subject). If done as you said, subject would become a different thing for nearly all test. 

by doing it right, imagine now I add here another test that checks for the content type in the response, then the subject changes again (in your terms).

logstash-core/spec/api/lib/api/node_spec.rb
purbon: yes and no, I will go for having two test as we are really testing two things, one that it return a good content type and another that make the JSON parser happy.

rakelib/default_plugins.rb
purbon: will do a more indeep check, just interesting to see ....

purbon: interesting, if I remove the `rubydebug`, funny thing `rake test:core` hangs, while bin/rspec not :disappointed:

purbon:
@ph finally got the rogue thread problem (and why the rubygems codec was in place, due to wrong pipeline/agent initialization.

what do you think? can this be merged?

ph:
LGTM

purbon:
@jsvd can you take a look here.

logstash-core/lib/logstash/api/lib/app/service.rb
```diff
@@ -23,6 +23,10 @@ def agent
     LogStash::Instrument::Collector.instance.agent
   end
 
+  def started?
+    !@snapshot.nil? && has_counters?
```
jsvd: since we check if @snapshot is nil in has_counters?, isn't this first check redundant?

purbon: basically yes, this end up here I guess as legacy :-P

logstash-core/lib/logstash/api/lib/app/service.rb
```diff
@@ -40,4 +44,20 @@ def get(key)
...
     LogStash::Json.dump(data)
   end
+
+  private
+
+  def has_counters?
+    return false if @snapshot.nil?
+    if @snapshot_rotation_mutex.try_lock
+      has = metric_types.include?("LogStash::Instrument::MetricType::Counter")
+      has = has &&  metric_types.include?("LogStash::Instrument::MetricType::Gauge")
```
jsvd: you can also do 
```ruby
(["LogStash::Instrument::MetricType::Counter", "LogStash::Instrument::MetricType::Gauge"] - metric_types).empty?
```

logstash-core/lib/logstash/api/lib/app/service.rb
```diff
@@ -40,4 +44,20 @@ def get(key)
     end
     LogStash::Json.dump(data)
   end
+
+  private
+
+  def has_counters?
+    return false if @snapshot.nil?
+    if @snapshot_rotation_mutex.try_lock
```
jsvd: the mutex is locked we return false here, is that what we want? Why not block until it's free?

```ruby
@snapshot_rotation_mutex.synchronize do
  metric_types.include?("LogStash::Instrument::MetricType::Counter") &&
  metric_types.include?("LogStash::Instrument::MetricType::Gauge")
end
```

logstash-core/lib/logstash/api/lib/app/service.rb
```diff
@@ -40,4 +44,20 @@ def get(key)
...
+      has = metric_types.include?("LogStash::Instrument::MetricType::Counter")
+      has = has &&  metric_types.include?("LogStash::Instrument::MetricType::Gauge")
+      @snapshot_rotation_mutex.unlock
+      return has
+    end
+  end
+
+  def metric_types
+    @snapshot.metric_store.all.clone.map { |t| t.class.to_s }
```
jsvd: why is clone necessary here?

logstash-core/spec/api/spec_helper.rb
```diff
@@ -11,9 +9,107 @@
...
+    args = [
+      :logger => Cabin::Channel.get(LogStash),
+      :auto_reload => false,
+      :collect_metric => true,
+      :debug => false,
+      :node_name => "test_agent",
+      :web_api_http_port => rand(9600..9700)
+    ]
+
+    @config_str   = "input { generator {count => 0} } output { null { } }"
```
jsvd: I believe you can just do output {} to have an empty output section, this eliminating the need for the logstash-output-null

logstash-core/lib/logstash/api/lib/app/service.rb
purbon: makes sense, good idea.

logstash-core/lib/logstash/api/lib/app/service.rb
purbon: I want to avoid messing up with probable internal references here, but yeah! not much need at the end of the day :+1:

logstash-core/spec/api/spec_helper.rb
purbon: but this will output stuff to stdout, and I don't want that.

purbon:
@jsvd thanks a lot for your feedback, I applied most of your comments, would you be so nice to let me know if everything is now ok to be merged? thanks a lot.

logstash-core/spec/api/spec_helper.rb
jsvd: nop:

```
/tmp/logstash-2.2.2 % bin/logstash -e "input { generator { count => 10000 } } output {} "
Logstash startup completed
Logstash shutdown completed
```

purbon: interesting, thanks Joao!

purbon: fixed with the last commit, do you have anything else for me?

ph:
LGTM, I have rechecked @jsvd's comment, good feedback Joao!

jsvd:
LGTM

purbon:
thanks @ph, @jsvd squashing and merging.

PR #4783, opened 2016-03-10
Bump version to 2.3.0.snapshot1

suyograo:
Published snapshot version for logstash-core-event-java gem


logstash-core/logstash-core.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
   gem.version       = LOGSTASH_CORE_VERSION
 
-  gem.add_runtime_dependency "logstash-core-event-java", "~> 3.0.0.dev"
+  gem.add_runtime_dependency "logstash-core-event-java", "2.3.0.snapshot1"
```
ph: @colinsurprenant is that the correct constraints? `~>` version strict? I believe the strict is the way to go just wanted to make sure.

Gemfile.jruby-1.9.lock
```diff
@@ -40,26 +30,41 @@ GEM
...
       json (~> 1.8)
-      rest-client (>= 1.6.8, < 2)
       simplecov (~> 0.11.0)
       term-ansicolor (~> 1.3)
       thor (~> 0.19.1)
       tins (~> 1.6.0)
     diff-lcs (1.2.5)
     docile (1.1.5)
-    domain_name (0.5.25)
+    domain_name (0.5.20160309)
```
ph: Tthey went over the top in their version. oh semver :P

ph:
minor comments concerning the version constraint for the `logstash-java-event`, I have created a tarballs and tested it with a beats/filebeat configuration and it seens to work for me.

Lets get the opinion of @colinsurprenant for the version and we should be able to merge/release this.

Gemfile.jruby-1.9.lock
colinsurprenant: wtf

logstash-core/logstash-core.gemspec
colinsurprenant: han. I don't know anymore :/  - let's clarify our core gems versions constrains strategy. I think that for the snapshotX it doesn't matter much and we can keep a fixed version constrain since we will always release new complete snapshots

Gemfile.jruby-1.9.lock
ph: I have checked why, its for a internal reference when they update TLD database.. but oh well

logstash-core/logstash-core.gemspec
ph: +1, and lets make a "release document" or/and a tools for the version bump.

colinsurprenant:
changes LGTM, I did not test package generation etc. I think we're good on the version constrains for the snapshotX releases, but we should clarify this before the release.

logstash-core/logstash-core.gemspec
suyograo: Yeah that was my idea too. For snapshots, we need to pin to the corresponding versions since a lot of stuff *may* be changing with Event.

ph:
LGTM

suyograo:
Merged with ef06df9200565bf3dea0301b3467bb3b1de8e27b

Issue #4730, opened 2016-02-28
Document various metrics API


suyograo:
See #4446 

Should be added to asciidoc

suyograo:
Fixed in #4785

PR #4785, opened 2016-03-10
Add documentation for monitoring API

dedemorton:
@suyograo and other reviewers: The monitoring API documentation for Alpha 1 (issue #4730 ) is ready to review. When looking through this PR, please look for and respond to inline comments that contain questions that begin with //QUESTION.

I've uploaded the built files so you can see how they look formatted:   https://logstashreference.firebaseapp.com/monitoring.html 

At some point, we might want to have a top-level container called "Monitoring Logstash" but that doesn't make sense until we have more content about monitoring (not just API docs).

### DIFF
--- a/docs/static/monitoring-apis.asciidoc
+++ b/docs/static/monitoring-apis.asciidoc
@@ -1,0 +1,293 @@
+[[monitoring]]
+== Monitoring APIs
+
+Logstash provides the following monitoring APIs to retrieve runtime metrics
+about Logstash:
+
+* <<root-resource-api>>
+* <<stats-info-api>>
+* <<hot-threads-api>>
+
+//NOTE: Need to add this to the doc after Alpha 1: * <<plugins-api>>
+
+[float]
+[[monitoring-common-options]]
+=== Common Options
+
+The following options can be applied to all of the Logstash monitoring APIs.
+
+[float]
+==== Pretty Results
+
+When appending `?pretty=true` to any request made, the JSON returned
+will be pretty formatted (use it for debugging only!). Another option is
+to set `?format=yaml` which will cause the result to be returned in the
+(sometimes) more readable yaml format.
+
+[float]
+==== Human-Readable Output
+
+NOTE: For Alpha 1, the `human` option is supported for the <<hot-threads-api>>
+only. When you specify `human=true`, the results are returned in plain text instead of
+JSON format. The default is false.
+
+Statistics are returned in a format suitable for humans
+(eg `"exists_time": "1h"` or `"size": "1kb"`) and for computers
+(eg `"exists_time_in_millis": 3600000` or `"size_in_bytes": 1024`).
+The human-readable values can be turned off by adding `?human=false`
+to the query string. This makes sense when the stats results are
+being consumed by a monitoring tool, rather than intended for human
+consumption.  The default for the `human` flag is
+`false`.
+
+
+/////
+COMMENTED OUT because this API was moved to Alpha 2. Feel free to add review
+comments, tho, if you notice inaccuracies.
+
+[[plugins-api]]
+=== Plugins API
+
+experimental[]
+
+The plugins API gets information about all Logstash plugins that are currently installed.
+This API basically returns the output of running the `bin/plugins list --verbose` command.
+
+[source,js]
+--------------------------------------------------
+GET /_plugins
+--------------------------------------------------
+
+The output is a JSON document.
+
+Example response:
+
+[source,js]
+--------------------------------------------------
+"total": 102
+"plugins" : [
+  {
+      "name": "logstash-output-pagerduty"
+      "version": "2.0.2"
+  },
+  {
+      "name": "logstash-output-elasticsearch"
+      "version": "2.1.2"
+  }
+....
+] 
+--------------------------------------------------
+/////
+
+[[root-resource-api]]
+=== Root Resource API
+
+experimental[]
+
+The root resource API retrieves general information about the Logstash instance, including
+the host name and version information.
+
+[source,js]
+--------------------------------------------------
+GET /
+--------------------------------------------------
+
+Example response:
+
+[source,js]
+--------------------------------------------------
+{
+   "hostname": "skywalker",
+    "version" : {
+        "number" : "2.1.0",       
+    }
+  }
+--------------------------------------------------
+
+
+See <<monitoring-common-options, Common Options>> for a list of options that can be applied to all
+Logstash monitoring APIs.
+
+[[stats-info-api]]
+=== Stats Info API
+
+experimental[]
+
+The stats info API retrieves runtime stats about Logstash. 
+
+// COMMENTED OUT until Logstash supports multiple pipelines: To retrieve all stats for the Logstash instance, use the `_node/stats` endpoint:
+
+[source,js]
+--------------------------------------------------
+GET /_node/stats/<types>
+--------------------------------------------------
+
+////
+COMMENTED OUT until Logstash supports multiple pipelines: To retrieve all stats per pipeline, use the `_pipelines/stats` endpoint:
+
+[source,js]
+--------------------------------------------------
+GET /_pipelines/stats/<types>
+--------------------------------------------------
+////
+
+Where `<types>` is optional and specifies the types of stats you want to return.
+
+By default, all stats are returned. You can limit this by combining any of the following types: 
+
+[horizontal]
+`events`::
+	Gets event information since startup. 
+`jvm`::
+	Gets JVM stats, including stats about garbage collection. 
+
+For example, the following request returns a JSON document that shows the number of events
+that were input, filtered, and output by Logstash since startup:
+
+[source,js]
+--------------------------------------------------
+GET /_node/stats/events
+--------------------------------------------------
+
+Example response:
+
+[source,js]
+--------------------------------------------------
+{
+    "events": {
+        "in": 91,
+        "filtered": 91,
+        "out": 91
+    }
+}
+--------------------------------------------------
+
+The following request returns a JSON document containing JVM stats:
+
+[source,js]
+--------------------------------------------------
+GET /_node/stats/jvm
+--------------------------------------------------
+
+Example response:
+
+[source,js]
+--------------------------------------------------
+"jvm":{  
+   "timestamp":1453233447702,
+   "uptime_in_millis":211125811,
+   "mem":{  
+      "heap_used_in_bytes":58442576,
+      "heap_used_percent":5,
+      "heap_committed_in_bytes":259522560,
+      "heap_max_in_bytes":1037959168,
+      "non_heap_used_in_bytes":56332256,
+      "non_heap_committed_in_bytes":57475072,
+      "pools":{  
+         "young":{  
+            "used_in_bytes":41672000,
+            "max_in_bytes":286326784,
+            "peak_used_in_bytes":71630848,
+            "peak_max_in_bytes":286326784
+         },
+         "survivor":{  
+            "used_in_bytes":260552,
+            "max_in_bytes":35782656,
+            "peak_used_in_bytes":8912896,
+            "peak_max_in_bytes":35782656
+         },
+         "old":{  
+            "used_in_bytes":16510024,
+            "max_in_bytes":715849728,
+            "peak_used_in_bytes":16510024,
+            "peak_max_in_bytes":715849728
+         }
+      }
+   }
+--------------------------------------------------
+
+See <<monitoring-common-options, Common Options>> for a list of options that can be applied to all
+Logstash monitoring APIs.
+
+[[hot-threads-api]]
+=== Hot Threads API
+
+experimental[]
+
+The hot threads API gets the current hot threads for Logstash. A hot thread is a
+Java thread that has high CPU usage and executes for a longer than normal period
+of time.
+
+[source,js]
+--------------------------------------------------
+GET /_node/hot_threads
+--------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+
+* <<hot-threads-api>>
+* <<root-resource-api>>
+* <<stats-info-api>>
+
+//NOTE: Need to add this to the doc after Alpha 1: * <<plugins-api>>
+
+=== Testing Logstash APIs in Sense
+
+//QUESTION: What do you think of this approach? I felt that the links for invoking Sense would be useful, but I wanted to avoid confusing users. If they don't notice this topic, they might still be confused. Do you think I should keep the links that launch Sense, or should I remove them?
```
ph: I think sense is a great idea, but I am not sure it will work out of the box since the target is not an elasticsearch instance and I believe sense uses the kibana configuration to get the address of the server?

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+=== Human-Readable Output
+
+Statistics are returned in a format suitable for humans
+(eg `"exists_time": "1h"` or `"size": "1kb"`) and for computers
+(eg `"exists_time_in_millis": 3600000` or `"size_in_bytes": 1024`).
+The human-readable values can be turned off by adding `?human=false`
+to the query string. This makes sense when the stats results are
+being consumed by a monitoring tool, rather than intended for human
+consumption.  The default for the `human` flag is
+`false`.
```
ph: This is not exactly true, the API will always return the two formats the machine format and the human format in the JSON response. So you will get both `exists_time` and `exists_time_in_millis` in your response.

The human params currently only works the `hot_threads` api call and it currently do the following it will return a plain text string instead of a json document which make the response easier to read than a formatted json.

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+
+[[hot-threads-api]]
+=== Hot Threads API
+
+experimental[]
+
+//QUESTION: Which of the params shown for the Elasticsearch hot_threads API also available here? (I'm talking about the threads, interval, type, and ignore_idle_threads parameters shown here: https://www.elastic.co/guide/en/elasticsearch/reference/2.2/cluster-nodes-hot-threads.html) I ask even though these parameters are not mentioned in issue #4446 because I tested ?threads, and it works, but when I tested the other params, I couldn't really tell if they were working.
```
ph: The current query string params support by this call are  the following.

`thread` => integer, default 3, limit the number of hot threads returned by the api, they ar ethe threads that consume the most resources.
`human` => boolean default false, return the result in plain text instead of the json format.
`ignore_idle_threads` => default true, will not return idle threads.

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+    "percent_of_cpu_time" : 0.15,
+    "state" : "timed_waiting",
+    "path" : "/Users/suyog/ws/elastic/logstash/build/logstash-3.0.0.dev/vendor/local_gems/f5685da5/logstash-core-3.0.0.dev-java/lib/logstash/pipeline.rb:496",
+    "traces" : "\t\tjava.lang.Object.wait(Native Method)\n\t\torg.jruby.RubyThread.sleep(RubyThread.java:1002)\n\t\torg.jruby.RubyKernel.sleep(RubyKernel.java:803)\n\t\torg.jruby.RubyKernel$INVOKER$s$0$1$sleep.call(RubyKernel$INVOKER$s$0$1$sleep.gen)\n\t\torg.jruby.internal.runtime.methods.JavaMethod$JavaMethodN.call(JavaMethod.java:667)\n\t\torg.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:206)\n\t\torg.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:168)\n\t\torg.jruby.ast.FCallOneArgNode.interpret(FCallOneArgNode.java:36)\n\t\torg.jruby.ast.NewlineNode.interpret(NewlineNode.java:105)\n\t\torg.jruby.ast.BlockNode.interpret(BlockNode.java:71)\n"
+  } ]
+--------------------------------------------------
+
+//QUESTION: Do you think we should describe the fields that are returned by each API? I haven't done this because I feel that the fields are intuitively named and that the descriptions would be redundant. WDYT?
```
ph: I agree I believe they are self explicit and we might add more fields so it would make the doc harder to keep in sync with the actual code.

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+] 
+--------------------------------------------------
+/////
+
+[[root-resource-api]]
+=== Root Resource API
+
+experimental[]
+
+//QUESTION: Is "root resource API" a good name for this API?
```
ph: I think this is an OK name, I wish naming thing was easier from time to time ;)
@purbon do you have a better name for this one?

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+=== Root Resource API
+
+experimental[]
+
+//QUESTION: Is "root resource API" a good name for this API? 
+
+The root resource API retrieves general information about the Logstash instance, including
+the host name, version, and build hash.
+
+//QUESTION: Actually I tested the REST call on the latest build, the response did not include the build hash. Is this field still part of the output?
```
ph: I think in the current version we don't return the `build_hash`, I have created an issue to discuss if we still want that feature or not.

See https://github.com/elastic/logstash/issues/4786

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+[source,js]
+--------------------------------------------------
+GET /_node/stats/<types>
+--------------------------------------------------
+//AUTOSENSE
+
+To retrieve all stats per pipeline, use the `_pipelines/stats` endpoint:
+
+//QUESTION: Should we mention that for Alpha 1 that the API will only show one pipeline, main? Actually, when I try the REST call, it returns an empty document. Is the _pipelines/stats endpoint even supported in Alpha 1? I'm using a very basic Logstash config. Do I need to configure something to see pipeline stats? 
+
```
ph: The stats are dynamically created at runtime, if the pipeline don't do anything no stats will be collected. If you are using the `generator` input to send to elasticsearch you should get some metrics. Can you share you configuration?

ph: I think the current expectation for people is that logstash only run one pipeline, so I think we don't need to explain that the stats will be from the main pipeline? This could change in the future and when it happen we can update the documentation to reflect the new reality.

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+[source,js]
+--------------------------------------------------
+GET /_pipelines/stats/<types>
+--------------------------------------------------
+//AUTOSENSE
+
+Where `<types>` is optional and specifies the types of stats you want to return.
+
+//QUESTION: I'm assuming that I should be able to specify more than one type by using a comma separator, but when I specify more than one type, the call returns an empty document. Am I mistaken? For example, the following command returns {}: curl localhost:9600/_node/stats/events,jvm. I know that right now it doesn't make sense to specify the two separated by a comma because you get that by default, but I'm assuming that you plan to add other types over time. If you don't, let me know, and I'll reword the intro to the list of types.
```
ph: This might be a bug @purbon can you double check that?

ph:
I think an initial review, I think I have answered all your questions. Good stuff! Thanks @dedemorton

docs/static/monitoring-apis.asciidoc
dedemorton: @ph Is this true for Alpha1 or in general?

I was under the impression from my conversation with @suyograo and Alvin that these parameters would provide the same behavior that they provide in the Elasticsearch APIs (documented [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html)). 

Are you planning to add wider support for `?human` post-Alpha1? If so, I'm wondering if it would make sense to document the behavior as we'll deliver it in 5.0.0 and add an Alpha1 note indicating that the parameter is only supported for the hot threads API.

docs/static/monitoring-apis.asciidoc
acchen97: Not sure how this would work technically, perhaps we can leave out the Sense portions for now.

docs/static/monitoring-apis.asciidoc
suyograo: +1, no need to describe individual fields

docs/static/monitoring-apis.asciidoc
suyograo: Yeah, root resource is fine.

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
+[[monitoring]]
+== Monitoring APIs
+
+Logstash provides the following monitoring APIs to retrieve runtime metrics
+about Logstash:
+
+* <<hot-threads-api>>
```
acchen97: Can we shift hot threads to the last item (under stats info)?

docs/static/monitoring-apis.asciidoc
suyograo: +1 on leaving sense out. IMO most users of monitoring APIs are for operational health and not for developing. So they might just use `curl` or tie this into an existing monitoring solution

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+--------------------------------------------------
+//AUTOSENSE
+
+To retrieve all stats per pipeline, use the `_pipelines/stats` endpoint:
+
+//QUESTION: Should we mention that for Alpha 1 that the API will only show one pipeline, main? Actually, when I try the REST call, it returns an empty document. Is the _pipelines/stats endpoint even supported in Alpha 1? I'm using a very basic Logstash config. Do I need to configure something to see pipeline stats? 
+
+[source,js]
+--------------------------------------------------
+GET /_pipelines/stats/<types>
```
acchen97: Let's remove this _pipelines option as its a future item.

docs/static/monitoring-apis.asciidoc
dedemorton: @ph Yes, that's correct. The topic describes what users need to do to get it to work. I'm curious to hear what other people think. I'll gladly remove the Sense links if people think they are too confusing. They are a nice (easy-to-implement) extra, but not required. @suyograo Do you have an opinion?

suyograo: @dedemorton +1 on leaving sense out. IMO most uses of monitoring APIs are for operational health and not for developing or learning (which is where Sense is invaluable). So operators might just use `curl` or tie this into an existing monitoring solution.

docs/static/monitoring-apis.asciidoc
ph: > Are you planning to add wider support for ?human post-Alpha1? If so, I'm wondering if it would make sense to document the behavior as we'll deliver it in 5.0.0 and add an Alpha1 note indicating that the parameter is only supported for the hot threads API.

I think this would be a great idea, since we will probably add more support to the `human` option.

docs/static/monitoring-apis.asciidoc
dedemorton: @acchen97 Right now the APIs are listed in alphabetical order. One of my pet peeves about our documentation is that we don't always have a clear organizational pattern, which leaves it up to users to figure out where we've buried stuff. For reference info, I prefer to alphabetize the topics to make it easier for people to find stuff by name (which admittedly isn't important with a small set of topics). If we are planning to add a bunch of other monitoring APIs later, then I would prefer to keep the current organization and be consistent in all the APIs that we document. I'm open to your suggestion, though, if you feel strongly.

docs/static/monitoring-apis.asciidoc
```diff
@@ -0,0 +1,312 @@
...
+        "out": 91
+    }
+}
+--------------------------------------------------
+
+The following request returns a JSON document containing JVM stats:
+
+[source,js]
+--------------------------------------------------
+GET /_stats/jvm
```
dedemorton: Good catch. Fixed.

docs/static/monitoring-apis.asciidoc
dedemorton: @acchen97 Actually, ignore my previous comment. I'll move the section to the end for now and we can decide on the final organization when/if we have more APIs to worry about and the order matters more. :-)

dedemorton:
@suyograo, @ph  @acchen97 : I've pushed changes based on your feedback. Let me know if the changes look OK and whether this is ready to merge into the main docs. 

I've refreshed the built files so you can see the doc in the right context: https://logstashreference.firebaseapp.com/monitoring.html (you might need to reload the page to see the changes)

suyograo:
@dedemorton looking good. What about a top level section "Common Options" we can then document `?pretty` and `?human` settings?

dedemorton:
@suyograo I've moved the section about common options to the main container topic (I think that's what you were asking me to do). Also, I've added links to each reference topic to point back to the Common Options section. I've refreshed the draft at https://logstashreference.firebaseapp.com/monitoring.html.

suyograo:
LGTM. Please merge

acchen97:
Good work @dedemorton thanks! :+1:

Issue #4777, opened 2016-03-09
Add plugins list API


suyograo:
Follow up from #4446 

### Plugins

Expose an endpoint to get information on all plugins currently installed. This is basically the output of `bin/plugins list --verbose`

```GET /_plugins```

Response:

```
"total": 102
"plugins" : [
  {
      "name": "logstash-output-pagerduty"
      "version": "2.0.2"
  },
  {
      "name": "logstash-output-elasticsearch"
      "version": "2.1.2"
  }
....
]  
```

purbon:
changed the label to be in sync with the PR.

PR #4794, opened 2016-03-11
/_plugins resource creation

purbon:
Hi,
  this PR is an initial version for the `/_plugins` resource as defined in #4777. Is important to notice this involve some code duplication for the plugin listing, this will be clean up in a feature, _and probably longer refactoring PR_.

### DIFF
--- a/logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
+++ b/logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
@@ -1,0 +1,28 @@
+# encoding: utf-8
+require "app/command"
+
+class LogStash::Api::PluginsCommand < LogStash::Api::Command
+
+  def run
+    { :total => plugins.count, :plugins => plugins }
+  end
+
+  private
+
+  def plugins
+    @plugins ||= find_plugins_gem_specs.map do |spec|
+      { :name => spec.name, :version => spec.version.to_s }
+    end.sort_by do |spec|
+      spec[:name]
+    end
+  end
+
+  def find_plugins_gem_specs
+    @specs ||= Gem::Specification.find_all.select{|spec| logstash_plugin_gem_spec?(spec)}
+  end
+
+  def logstash_plugin_gem_spec?(spec)
+    spec.metadata && spec.metadata["logstash_plugin"] == "true"
+  end
+
+end
--- a/logstash-core/lib/logstash/api/lib/app/modules/plugins.rb
+++ b/logstash-core/lib/logstash/api/lib/app/modules/plugins.rb
@@ -1,0 +1,15 @@
+# encoding: utf-8
+require "app"
+
+module LogStash::Api
+  class Plugins < BaseApp
+
+    helpers AppHelpers
+
+    get "/" do
+      command = factory.build(:plugins_command)
+      respond_with(command.run())
+    end
+
+  end
+end
--- a/logstash-core/spec/api/lib/api/plugins_spec.rb
+++ b/logstash-core/spec/api/lib/api/plugins_spec.rb
@@ -1,0 +1,57 @@
+# encoding: utf-8
+require_relative "../../spec_helper"
+require "sinatra"
+require "app/modules/plugins"
+require "logstash/json"
+
+describe LogStash::Api::Plugins do
+
+  include Rack::Test::Methods
+
+  def app()
+    described_class
+  end
+
+  before(:all) do
+    get "/"
+  end
+
+  let(:payload) { LogStash::Json.load(last_response.body) }
+
+  it "respond to plugins resource" do
+    expect(last_response).to be_ok
+  end
+
+  it "return valid json content type" do
+    expect(last_response.content_type).to eq("application/json")
+  end
+
+  context "#schema" do
+    it "return the expected schema" do
+      expect(payload.keys).to include("plugins", "total")
+      payload["plugins"].each do |plugin|
+        expect(plugin.keys).to include("name", "version")
+      end
+    end
+  end
+
+  context "#values" do
+
+    it "return totals of plugins" do
+      expect(payload["total"]).to eq(payload["plugins"].count)
+    end
+
+    it "return a list of available plugins" do
+      payload["plugins"].each do |plugin|
+        expect(plugin).to be_available?
+      end
+    end
+
+    it "return non empty version values" do
+      payload["plugins"].each do |plugin|
+        expect(plugin["version"]).not_to be_empty
+      end
+    end
+
+  end
+end
--- a/docs/static/monitoring-apis.asciidoc
+++ b/docs/static/monitoring-apis.asciidoc
@@ -7,8 +7,7 @@ about Logstash:
 * <<root-resource-api>>
 * <<stats-info-api>>
 * <<hot-threads-api>>
-
-//NOTE: Need to add this to the doc after Alpha 1: * <<plugins-api>>
+* <<plugins-api>>
 
 [float]
 [[monitoring-common-options]]
@@ -39,11 +38,6 @@ to the query string. This makes sense when the stats results are
 being consumed by a monitoring tool, rather than intended for human
 consumption.  The default for the `human` flag is
 `false`.
-
-
-/////
-COMMENTED OUT because this API was moved to Alpha 2. Feel free to add review
-comments, tho, if you notice inaccuracies.
 
 [[plugins-api]]
 === Plugins API
@@ -77,7 +71,6 @@ Example response:
 ....
 ] 
 --------------------------------------------------
-/////
 
 [[root-resource-api]]
 === Root Resource API
--- a/logstash-core/lib/logstash/api/init.ru
+++ b/logstash-core/lib/logstash/api/init.ru
@@ -7,6 +7,7 @@ require 'app/root'
 require 'app/modules/stats'
 require 'app/modules/node'
 require 'app/modules/node_stats'
+require 'app/modules/plugins'
 
 env = ENV["RACK_ENV"].to_sym
 set :environment, env
@@ -20,7 +21,8 @@ run LogStash::Api::Root
 
 namespaces = { "/_node" => LogStash::Api::Node,
                "/_node/stats" => LogStash::Api::NodeStats,
-               "/_stats" => LogStash::Api::Stats }
+               "/_stats" => LogStash::Api::Stats,
+               "/_plugins" => LogStash::Api::Plugins }
 
 namespaces.each_pair do |namespace, app|
   map(namespace) do
--- a/logstash-core/lib/logstash/api/lib/app/command_factory.rb
+++ b/logstash-core/lib/logstash/api/lib/app/command_factory.rb
@@ -4,6 +4,7 @@ require "app/commands/system/basicinfo_command"
 require "app/commands/stats/events_command"
 require "app/commands/stats/hotthreads_command"
 require "app/commands/stats/memory_command"
+require "app/commands/system/plugins_command"
 
 module LogStash::Api
   class CommandFactory
@@ -16,7 +17,8 @@ module LogStash::Api
         :system_basic_info => SystemBasicInfoCommand,
         :events_command => StatsEventsCommand,
         :hot_threads_command => HotThreadsCommand,
-        :memory_command => JvmMemoryCommand
+        :memory_command => JvmMemoryCommand,
+        :plugins_command => PluginsCommand
       )
     end
 
--- a/logstash-core/spec/api/lib/api/node_spec.rb
+++ b/logstash-core/spec/api/lib/api/node_spec.rb
@@ -2,6 +2,7 @@
 require_relative "../../spec_helper"
 require "sinatra"
 require "app/modules/node"
+require "logstash/json"
 
 describe LogStash::Api::Node do
 
@@ -22,7 +23,7 @@ describe LogStash::Api::Node do
     end
 
     it "should return a JSON object" do
-      expect{ JSON.parse(last_response.body) }.not_to raise_error
+      expect{ LogStash::Json.load(last_response.body) }.not_to raise_error
     end
 
     context "#threads count" do
@@ -31,7 +32,7 @@ describe LogStash::Api::Node do
         do_request { get "/hot_threads?threads=5" }
       end
 
-      let(:payload) { JSON.parse(last_response.body) }
+      let(:payload) { LogStash::Json.load(last_response.body) }
 
       it "should return a json payload content type" do
         expect(last_response.content_type).to eq("application/json")
@@ -65,7 +66,7 @@ describe LogStash::Api::Node do
         do_request { get "/hot_threads?ignore_idle_threads=false&threads=10" }
       end
 
-      let(:payload) { JSON.parse(last_response.body) }
+      let(:payload) { LogStash::Json.load(last_response.body) }
 
       it "should return JIT threads" do
         thread_names = payload["threads"].map { |thread_info| thread_info["name"] }
--- a/logstash-core/spec/api/lib/api/node_stats_spec.rb
+++ b/logstash-core/spec/api/lib/api/node_stats_spec.rb
@@ -2,6 +2,7 @@
 require_relative "../../spec_helper"
 require "sinatra"
 require "app/modules/node_stats"
+require "logstash/json"
 
 describe LogStash::Api::NodeStats do
 
@@ -11,7 +12,7 @@ describe LogStash::Api::NodeStats do
     described_class
   end
 
-  let(:payload) { JSON.parse(last_response.body) }
+  let(:payload) { LogStash::Json.load(last_response.body) }
 
   context "#root" do
 
@@ -32,7 +33,7 @@ describe LogStash::Api::NodeStats do
 
   context "#events" do
 
-    let(:payload) { JSON.parse(last_response.body) }
+    let(:payload) { LogStash::Json.load(last_response.body) }
 
     before(:all) do
       do_request { get "/events" }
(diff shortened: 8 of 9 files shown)


purbon:
@ph your thoughts are for sure very welcome here.

logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
```diff
@@ -0,0 +1,26 @@
...
+  private
+
+  def plugins
+    @plugins ||= find_plugins_gem_specs.map do |spec|
+      { :name => spec.name, :version => spec.version.to_s }
+    end
+  end
+
+  def find_plugins_gem_specs
+    @specs ||= Gem::Specification.find_all.select{|spec| logstash_plugin_gem_spec?(spec)}
```
ph: I am starting to be a fan of `&method(:logstash_plugin_gem_spec)`
```
irb(main):001:0> a = [1, 2, 3, 4, 5]
=> [1, 2, 3, 4, 5]
irb(main):002:0> def higher?(m); m > 3;end
=> nil
irb(main):003:0> a.select(&method(:higher?))
=> [4, 5]
irb(main):004:0>
``

Comment (purbon) [inline logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb]: Is a good idea, but in this case I think the code is clear enough, so not much of need to make it shorter ... I'm a fan of clear and easy to understand code :+1:

Comment (ph) [inline logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb]: I know you have noted about code duplication in the description but this method is the same as
https://github.com/elastic/logstash/blob/ba96b1f0b3d2b14c63c3c8cd34af647088646710/lib/pluginmanager/util.rb#L98

Any reason why we don't use it?

File: logstash-core/spec/api/lib/api/plugins_spec.rb [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+
+  it "return the expected schema" do
+    expect(payload.keys).to include("plugins", "total")
+  end
+
+  it "return a list of available plugins" do
+    payload["plugins"].each do |plugin|
+      expect(plugin).to be_available?
+    end
+  end
```
ph: We might wand to check for the version? WDYT

logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
purbon: We can not call the plugin manager from the logstash-core, as the code is organized right now it makes no sense, I mean without a more deep refactoring of this classes you mention (they do really need to put moved and organized cleaner and better for being reusable).

logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def run
+    { :total => plugins.count, :plugins => plugins }
+  end
+
+  private
+
+  def plugins
+    @plugins ||= find_plugins_gem_specs.map do |spec|
+      { :name => spec.name, :version => spec.version.to_s }
```
ph: I believe `to_s` is unnecessary here since converting the hash to json should call this.

logstash-core/spec/api/lib/api/plugins_spec.rb
purbon: what is the point? which version? logstash-core one?

logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
purbon: is it necessary as otherwise I can not be serialized.

logstash-core/spec/api/lib/api/plugins_spec.rb
ph: at least we return a version key no, `be_available` could do that.
If the api only return the plugin name its a bit useless.

logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
ph: The path issue is killing us for that. I am OK for duplication in that case.

logstash-core/spec/api/lib/api/plugins_spec.rb
purbon: check the command and you will see what is returned, see https://github.com/elastic/logstash/pull/4794/files#diff-8000a2194e946f5453dccc60c4d6424bR14

this test is just checking that the command is actually returning stuff from the system, nothing else.

ph: updated previous comment to make it clearer.

logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
ph: OK

logstash-core/spec/api/lib/api/plugins_spec.rb
ph: @purbon I checked the code before raising the issue, I am just saying I want to make sure we return the version key in the JSON response. I believe we need to have a test for that.

purbon: we can do more in deep format check if you want, yes. But for now in all
test I tried to be a bit flexible as we still can change a lot of the
formats :-P

- purbon

suyograo: +1 on checking if version is returned

purbon:
@ph fixed your concerns with the schema checks, not it checks for versions. Let me know what do you think for this PR to move forward :+1:

logstash-core/spec/api/lib/api/plugins_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  it "return valid json content type" do
+    expect(last_response.content_type).to eq("application/json")
+  end
+
+  it "return the expected schema" do
+    expect(payload.keys).to include("plugins", "total")
+    payload["plugins"].each do |plugin|
+      expect(plugin.keys).to include("name", "version")
```
ph: Sorry to be picky, I think we should make this test a bit more robust and check to make sure the value isn't nil. 

TIL that the `include` matcher only check for the presence of they, maybe it was obvious to everyone that it was calling `include?` on the hash, but for me it did not ring a bell until this morning.

This code change make the test pass.
```diff
diff --git a/logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb b/logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
index 5b1b045..c1ffb77 100644
--- a/logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
+++ b/logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
@@ -11,7 +11,7 @@ class LogStash::Api::PluginsCommand < LogStash::Api::Command

   def plugins
     @plugins ||= find_plugins_gem_specs.map do |spec|
-      { :name => spec.name, :version => spec.version.to_s }
+      { :name => nil, :version => nil }
     end
   end
```

ph:
One last comment it will get my LGTM.

logstash-core/spec/api/lib/api/plugins_spec.rb
purbon: yes it makes the test pas, but this situation is impossible to happen for how, if spec.version.to_s return nil, this means we installed plugins without version ... this will always return.

ph: Agree, it wont happen when you `know` the implementation, but when we refactor the implementation to share code with the plugin manager this statement will certainly changes.

So IMHO, I think we shouldn't use `include?` unless we don't actually mind sending nil values

ph:
Minor details, we should sort on the plugin name?
```json
~/e/logstash git:pr/4794 ❯❯❯ curl http://localhost:9600/_plugins\?pretty\=1                                                                                                                                                                                                                                               ✱
{
  "total" : 13,
  "plugins" : [ {
    "name" : "logstash-codec-json",
    "version" : "2.1.1"
  }, {
    "name" : "logstash-codec-line",
    "version" : "2.1.0"
  }, {
    "name" : "logstash-codec-json_lines",
    "version" : "2.1.1"
  }, {
    "name" : "logstash-codec-plain",
    "version" : "2.0.2"
  }, {
    "name" : "logstash-filter-clone",
    "version" : "2.0.4"
  }, {
    "name" : "logstash-patterns-core",
    "version" : "2.0.2"
  }, {
    "name" : "logstash-filter-grok",
    "version" : "2.0.3"
  }, {
    "name" : "logstash-filter-multiline",
    "version" : "2.0.3"
  }, {
    "name" : "logstash-filter-mutate",
    "version" : "2.0.3"
  }, {
    "name" : "logstash-input-generator",
    "version" : "2.0.2"
  }, {
    "name" : "logstash-input-stdin",
    "version" : "2.0.2"
  }, {
    "name" : "logstash-input-tcp",
    "version" : "3.0.2"
  }, {
    "name" : "logstash-output-stdout",
    "version" : "2.0.4"
  } ]
}%
```
This should be the correct order
```
/e/logstash-plugins ❯❯❯ curl -s 'http://localhost:9600/_plugins?pretty=1' | jq '.plugins[].name' |sort
"logstash-codec-json"
"logstash-codec-json_lines"
"logstash-codec-line"
"logstash-codec-plain"
"logstash-filter-clone"
"logstash-filter-grok"
"logstash-filter-multiline"
"logstash-filter-mutate"
"logstash-input-generator"
"logstash-input-stdin"
"logstash-input-tcp"
"logstash-output-stdout"
"logstash-patterns-core"

**Comment by purbon** [comment]
we can do that yes,

- purbon

On Mon, Mar 14, 2016 at 4:09 PM Pier-Hugues Pellerin <

> Minor details, we should sort on the plugin name?
>
> ~/e/logstash git:pr/4794 ❯❯❯ curl http://localhost:9600/_plugins\?pretty\=1                                                                                                                                                                                                                                               ✱
> {
>   "total" : 13,
>   "plugins" : [ {
>     "name" : "logstash-codec-json",
>     "version" : "2.1.1"
>   }, {
>     "name" : "logstash-codec-line",
>     "version" : "2.1.0"
>   }, {
>     "name" : "logstash-codec-json_lines",
>     "version" : "2.1.1"
>   }, {
>     "name" : "logstash-codec-plain",
>     "version" : "2.0.2"
>   }, {
>     "name" : "logstash-filter-clone",
>     "version" : "2.0.4"
>   }, {
>     "name" : "logstash-patterns-core",
>     "version" : "2.0.2"
>   }, {
>     "name" : "logstash-filter-grok",
>     "version" : "2.0.3"
>   }, {
>     "name" : "logstash-filter-multiline",
>     "version" : "2.0.3"
>   }, {
>     "name" : "logstash-filter-mutate",
>     "version" : "2.0.3"
>   }, {
>     "name" : "logstash-input-generator",
>     "version" : "2.0.2"
>   }, {
>     "name" : "logstash-input-stdin",
>     "version" : "2.0.2"
>   }, {
>     "name" : "logstash-input-tcp",
>     "version" : "3.0.2"
>   }, {
>     "name" : "logstash-output-stdout",
>     "version" : "2.0.4"
>   } ]
> }%
>
> This should be the correct order
>
> /e/logstash-plugins ❯❯❯ curl -s 'http://localhost:9600/_plugins?pretty=1' | jq '.plugins[].name' |sort
> "logstash-codec-json"
> "logstash-codec-json_lines"
> "logstash-codec-line"
> "logstash-codec-plain"
> "logstash-filter-clone"
> "logstash-filter-grok"
> "logstash-filter-multiline"
> "logstash-filter-mutate"
> "logstash-input-generator"
> "logstash-input-stdin"
> "logstash-input-tcp"
> "logstash-output-stdout"
> "logstash-patterns-core"
>
> —
> <https://github.com/elastic/logstash/pull/4794#issuecomment-196359947>.
>

File: logstash-core/spec/api/lib/api/plugins_spec.rb (hunk 4 above)
Comment (purbon) [inline logstash-core/spec/api/lib/api/plugins_spec.rb]: Hi,
   do you want to also make sure we return plugin names with content and
following our format?

 should I extend this way of testing to each other test for the webapi?

  - purbon

File: logstash-core/spec/api/lib/api/plugins_spec.rb [hunk 5]
```diff
@@ -0,0 +1,45 @@
...
+
+  def app()
+    described_class
+  end
+
+  before(:all) do
+    get "/"
+  end
+
+  let(:payload) { JSON.parse(last_response.body) }
```
ph: as reported by @colinsurprenant we might want to use our own serialized for the test?

logstash-core/spec/api/lib/api/plugins_spec.rb
colinsurprenant: (excuse my intrusion) 

`nil` is a bitch, I wish it did not exist :P but I believe that when testing an interface where `nil` values are illegal, we should add a specific test for it. now, as far as testing for proper content format, it's not a bad idea, but pragmatically I thinks that just checking that the return type is `String` and non-empty would be sufficient IMO.

ph: > do you want to also make sure we return plugin names with content and
following our format?

By the format you mean the `logstash-*? This test doesnt involve mocking the `Gem::Specification` layer? 

I think this is a good idea @purbon.

>  should I extend this way of testing to each other test for the webapi?
No, Lets keep this PR targetted on the `_plugins`

ph: > nil is a bitch, I wish it did not exist

Well how many ruby programmer lives were destroyed by a nil value hidden somewhere between monkey patches and alias method chain.

purbon: @colinsurprenant no need to excuse, just wondering here how deep to go in the test, if we test for schema and values or only schema. For this, as we know the implementation is not that hard not to test for nil, but will add it.

purbon:
@ph @colinsurprenant fixed all your concerns, is there anything else left?

logstash-core/spec/api/lib/api/plugins_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    it "return a list of available plugins" do
+      payload["plugins"].each do |plugin|
+        expect(plugin).to be_available?
+      end
+    end
+
+    it "return non empty version values" do
+      payload["plugins"].each do |plugin|
+        expect(plugin["version"]).not_to be_empty
```
ph: I think the correct assertion is `be_nil` since `NIL` wont implement empty?

```
Failure/Error: expect({}[:hello]).not_to be_empty
expected  to respond to `empty?`
```

purbon: Hi, actually this is the way to implement it and not to check for null. We
aim to check for "something" to be returned so we check for empty.

If null is provided, test is going to fail anyway.

/purbon

ph: OK, I don't mind then, LGTM.

purbon: Good, then i will move forward and merge as this is a simple PR.

/purbon

purbon:
After talking to @jsvd, he is think this PR is big enough to be reviewed by two people. Your thoughts are welcome man.

purbon:
@dedemorton I added back your commented out section for plugins api, is this good enough to be merged? let me know your thoughts.

logstash-core/spec/api/lib/api/plugins_spec.rb
colinsurprenant: > just wondering here how deep to go in the test

@purbon I think I explained my POV on this in my previous comment?

dedemorton:
@purbon Yes, I think it's fine to merge this.

logstash-core/lib/logstash/api/lib/app/commands/system/plugins_command.rb
jsvd: no dependencies of the plugin? :(

purbon: we only spoke about adding now the name and the version. we can definitely
add more info in next iterations.

purbon: see the issue for ref specs on the implementation.

logstash-core/lib/logstash/api/lib/app/modules/plugins.rb
```diff
@@ -0,0 +1,15 @@
+# encoding: utf-8
+require "app"
+
+module LogStash::Api
+  class Plugins < BaseApp
+
+    helpers AppHelpers
+
+    get "/" do
+       command = factory.build(:plugins_command)
```
jsvd: please fix indentation

jsvd:
Minor comment + question about also providing list of dependencies for a plugin.
For debugging purpose this is quite useful. However now that I think about it, what we can extract from the gemspec is the requirement list, not the actual version of the gems installed..
This would be awesome: `:dependencies => { "manticore" => { :depends => "~> 0.5.3", :installed => "0.5.6" }}`

ph:
@jsvd This is a good idea, lets make an issue to track it. I am not sure of the complexity to implement that feature and I think it should be hidden behind a query params. 

I don't see it to be useful other than debugging the dependencies issues when something is wrong with the plugin manager.

purbon:
The feature is good, I'm all in for having an issue and implementing it in
a feature iteration.

- purbon

jsvd:
LGTM

logstash-core/lib/logstash/api/lib/app/modules/plugins.rb
purbon: fixing this and merging as you gave a :heart: lgtm :+1:

PR #4801, opened 2016-03-11
Clear the Collector metric store

### LINKED ISSUE #4542: Delete metrics in the metric store when a pipeline reloaded
Currently the `MetricStore` will keep all the collected metrics since Logstash starts, the only way to clean the store is to physically restart Logstash. This is a problem with the introduction of config reloading, which could result in unwanted or inaccurate metric collection.

We collect metrics at two different level: **application** and the **pipeline**.

At the application level, we are gathering information about the jvm, cpu, os and events in/outs.
At the pipeline levels we also gather in/out and in/out of the different plugins.

So this generate a few questions at how we will handle the lifecycle of the metrics:

What should happen when a new config is reloaded inside logstash? Do we think that bringing a new config into Logstash is similar to a restart and we clear all the metrics collected before starting the new pipeline **OR** we clean only the information gathered at the pipeline level? 

In the first version of the metrics we only show that we gather information at the system level it might make more sense to clear the whole collected metrics? Since it will be the only way for people to know on their graph that something restarted logstash? (The spike)

If we consider clearing only the pipeline level we need to make a change to the way the ID of the pipeline are dealt with.. When we start logstash the default pipeline is called `main`, when logstash reload a new config the pipeline created from that config is also named `main`, We need to introduce a UUID to make sure no 2 pipeline can have the same ID in the metric store.

ph:
When a new pipeline is started we need to clear the metric store to make
sure the collected are relevant to the actual configuration.

Fix #4542

### DIFF
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -1,6 +1,8 @@
 # encoding: utf-8
-require 'spec_helper'
-require 'stud/temporary'
+require "spec_helper"
+require "stud/temporary"
+require "logstash/inputs/generator"
+require_relative "../support/mocks_classes"
 
 describe LogStash::Agent do
 
@@ -271,6 +273,65 @@ describe LogStash::Agent do
   context "#uptime" do
     it "return the number of milliseconds since start time" do
       expect(subject.uptime).to be >= 0
+    end
+  end
+
+  context "metrics after config reloading" do
+    let(:dummy_output) { DummyOutput.new }
+    let(:config) { "input { generator { } } output { dummyoutput { } }" }
+    let(:new_config_generator_counter) { 50 }
+    let(:new_config) { "input { generator { count => #{new_config_generator_counter} } } output { dummyoutput {} }" }
+    let(:config_path) do
+      f = Stud::Temporary.file
+      f.write(config)
+      f.close
+      f.path
+    end
+    let(:interval) { 0.2 }
+    let(:pipeline_settings) { { :pipeline_workers => 4,
+                                :config_path => config_path } }
+
+    let(:agent_args) do
+      super.merge({ :auto_reload => true,
+                    :reload_interval => interval,
+                    :collect_metric => true })
+    end 
+
+    before :each do
+      allow(DummyOutput).to receive(:new).at_least(:once).with(anything).and_return(dummy_output)
+      allow(LogStash::Plugin).to receive(:lookup).with("input", "generator").and_return(LogStash::Inputs::Generator)
+      allow(LogStash::Plugin).to receive(:lookup).with("codec", "plain").and_return(LogStash::Codecs::Plain)
+      allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(DummyOutput)
+
+      @t = Thread.new do
+        subject.register_pipeline("main",  pipeline_settings)
+        subject.execute
+      end
+
+      sleep(2)
+    end
+
+    after :each do
+      Stud.stop!(@t)
+      @t.join
+    end
+
+    it "resets the metric collector" do
+      # We know that the store has more events that the next expect
+      sleep(0.01) while dummy_output.events.size < new_config_generator_counter
+      snapshot = LogStash::Instrument::Collector.instance.snapshot_metric
+      expect(snapshot.metric_store.get_with_path("/stats/events")[:stats][:events][:in].value).to be > new_config_generator_counter
+
+      # update the configuration and give some time to logstash to pick it up and do the work
+      IO.write(config_path, new_config)
+
+      sleep(interval * 3) # Give time to reload the config
+      
+      # Since thre is multiple threads involved and with the configuration reload, 
+      # It can take some time to the states be visible in the store
+      sleep(0.01) while dummy_output.events.size < new_config_generator_counter
+      snapshot = LogStash::Instrument::Collector.instance.snapshot_metric
+      expect(snapshot.metric_store.get_with_path("/stats/events")[:stats][:events][:in].value).to eq(new_config_generator_counter)
     end
   end
 end
--- a/logstash-core/spec/support/mocks_classes.rb
+++ b/logstash-core/spec/support/mocks_classes.rb
@@ -1,0 +1,26 @@
+# encoding: utf-8
+require "logstash/outputs/base"
+
+class DummyOutput < LogStash::Outputs::Base
+  config_name "dummyoutput"
+  milestone 2
+
+  attr_reader :num_closes, :events
+
+  def initialize(params={})
+    super
+    @num_closes = 0
+    @events = []
+  end
+
+  def register
+  end
+
+  def receive(event)
+    @events << event
+  end
+
+  def close
+    @num_closes = 1
+  end
+end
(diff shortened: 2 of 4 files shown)


logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -157,11 +157,17 @@
...
         expect(subject.pipelines[pipeline_id].config_str).to eq(new_config)
       end
+
       it "starts the pipeline" do
         expect(subject).to receive(:stop_pipeline)
         expect(subject).to receive(:start_pipeline)
         subject.send(:reload_state!)
       end
+
+      it "clears the metric collector" do
```
suyograo: Can we add a more concrete test where we send some events, make sure metric values are incremented and then send `reload_state`, then check if the values in metrics are zero?

ph: Yes we can do that.

ph: Good catch you have found a race condition :tada:

purbon:
If @suyograo test improvement is added, I'm good with this merge, the important thing to focus here is the clear function is thread safe and Suyog's test proposal will test for that.

ph:
The problem isnt actually a race condition its the way the tests are done in the `agent_spec`, I have been clearing that out and I am getting a bunch of weird rspec issues concerning double used in mutliples places, IOErrors and weird ruby stack traces. Still wondering what is the real cause here. I think we have to much stubbing of methods in the test which hides problems.

ph:
When I say this is a race condition I was wrong, only one thread at a time can change the value of the metric store so I don't think this is the issue, but it __might__ hide a visibility problem.

I have written a more appropriate test that actually does the integration test similar to what I did in the `pipeline_spec` but the other previously mentioned issues are getting in my way. The new test actually check the value of the store with the events that was generated from the generator and asserts the count.

Issues I have:
1. same `double` instance used in multiple context.
2. Already used port, (the other test deactivate the http server)

The number 1, is in my opinion generated because we send the logger to other library, we don't really know what they do with it.

1. This is weird, since we actually shutdown the server.

purbon:
I agree with 1, but not sure this is the issue ... doubles should not be reused between context, we're keeping a reference to this double somewhere :crying_cat_face: 


number 2 is strange, for the webapi test I started like this test, but endup having to use a mock agent to properly mockup the server start and only start the internal rack:test interface, worst investigating more I guess.

purbon:
@ph still WiP? or you want more detail reviews?

ph:
still wip.

ph:
Easy review, easy fix when you think about it :P

purbon:
ok, going for it.

purbon:
please jenkins, test this.

purbon:
please jenkins, test this.

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -197,4 +199,65 @@
...
     end
   end
+
+  context "metrics after config reloading" do
+    let(:dummy_output) { DummyOutput.new }
+    let(:config) { "input { generator { } } output { dummyoutput { } }" }
+    let(:new_config_generator_counter) { 50 }
+    let(:new_config) { "input { generator { count => #{new_config_generator_counter} } } output { dummyoutput {} }" }
+    let(:config_path) do
+      f = Stud::Temporary.file
```
purbon: not related with this PR, just a comment for feature interactions.

We should be able to actually inject this configuration into the pipeline as plain string and not from a temporal path, what do you think? If we do so the pipeline testing and reasoning will be much easier, as then the agent can reinject the new plain string config on it.

what do you think?

//cc @jsvd

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -197,4 +199,65 @@
...
+                    :collect_metric => true })
+    end 
+
+    before :each do
+      allow(DummyOutput).to receive(:new).at_least(:once).with(anything).and_return(dummy_output)
+      allow(LogStash::Plugin).to receive(:lookup).with("input", "generator").and_return(LogStash::Inputs::Generator)
+      allow(LogStash::Plugin).to receive(:lookup).with("codec", "plain").and_return(LogStash::Codecs::Plain)
+      allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(DummyOutput)
+
+      @t = Thread.new do
```
purbon: not necessary to fix for this version, but generally this is a bad costume in rspec as it very error prone. :speak_no_evil:

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -197,4 +199,65 @@
...
+    after :each do
+      Stud.stop!(@t)
+      @t.join
+    end
+
+    it "resets the metric collector" do
+      # We know that the store has more events that the next expect
+      sleep(0.01) while dummy_output.events.size < new_config_generator_counter
+      snapshot = LogStash::Instrument::Collector.instance.snapshot_metric
+      expect(snapshot.metric_store.get_with_path("/stats/events")[:stats][:events][:in].value).to be > new_config_generator_counter
```
purbon: `<style comment>`
it would be nice to break up this line a bit, so it gets more readable.
`</style comment>`

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -197,4 +199,65 @@
...
+
+      # update the configuration and give some time to logstash to pick it up and do the work
+      IO.write(config_path, new_config)
+
+      sleep(interval * 3) # Give time to reload the config
+      sleep(0.01) while dummy_output.events.size < new_config_generator_counter
+
+      # try a few times
+      wait_for do
+        expect(true).to eq(false)
```
purbon: why do we use this here? is a bit obscure, I do recommend looking for a more reasonable alternative.

logstash-core/spec/support/mocks_classes.rb
```diff
@@ -0,0 +1,26 @@
+# encoding: utf-8
+require "logstash/outputs/base"
+
+class DummyOutput < LogStash::Outputs::Base
```
purbon: `<not for this change, but ....>`
I'm now really seeing _too much_ mocks/dummy/whatever kind of plugins, is time to wipe all the real plugin dependencies and organize all this mocks properly.

purbon:
Left a few comments, most of them generally about style, rants about already having a lot of mocks/dummy, that need to be properly organized.

Only important comment is https://github.com/elastic/logstash/pull/4801#discussion_r57318710, this is important to fix because is actually really obscure code. As soon as this is fix, the rest it's LGTM.

logstash-core/spec/logstash/agent_spec.rb
ph: bad commit, good catch.

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -273,4 +275,63 @@
...
+      IO.write(config_path, new_config)
+
+      sleep(interval * 3) # Give time to reload the config
+      
+      # Since thre is multiple threads involved and with the configuration reload, 
+      # It can take some time to the states be visible in the store
+      sleep(0.01) while dummy_output.events.size < new_config_generator_counter
+      snapshot = LogStash::Instrument::Collector.instance.snapshot_metric
+      expect(snapshot.metric_store.get_with_path("/stats/events")[:stats][:events][:in].value).to eq(new_config_generator_counter)
+    end
```
ph: @purbon I have removed the `wair_for` here So It is okay to get a LGTM from you?

I ran the test of this PR 6 times to check if this could raise a timing issue, I believe its should be okay to only check if the events are correctly send at the output end.

purbon: ++

purbon:
please jenkins, test this.

purbon:
LGTM

PR #4811, opened 2016-03-13
to see result of filter in stdout

brizjin:
Needs to use codec rubydebug for example


suyograo:
@brizjin can you please perform step 2 of https://github.com/elasticsearch/logstash/blob/master/CONTRIBUTING.md#contribution-steps

Issue #1569, opened 2014-07-24
Internal logging: A return to JSON


jordansissel:
Back in the day, logstash used json for its log format. However, non-UTF-8 problems and bad logging styles caused crashes *in* the logger, so we moved to ruby's `Object#inspect`.

Now that we have a better handle on UTF-8 problems *and* we have a better handle on how to do logging, it's probably time to make Cabin go back to emitting json.

purbon:
It would be really nice as some people ask how to ingest logstash/elasticsearch logs back into the same platform to then be able to monitor it.

simmel:
I've done this a very hackish way with regex (and we know [where that leads](http://stackoverflow.com/a/1732454))

```
filter {
  # Make sure the message is begins and ends with {} because otherwise we can't
  # parse it anyway.
  if [program] == "logstash" and [message] =~ /^{\s*:.*:level=>:\w+}$/ {

    # XXX Transform Ruby debug output (i.e. a Ruby hash serialized) into JSON
    # the hackish way. This is very ugly but there is no other way except doing
    # eval on it and we know there that path leads...
    # Turns this:
    # {:message=>"meow", :timestamp_field=>"\"2014-12-16T08:58:41.517Z\"", :nothing=>nil, :level=>:error}
    # into this:
    # {"message": "meow", "timestamp_field":"\"2014-12-16T08:58:41.517Z\"", "nothing": null, "level": "error"}
    mutate {
      gsub => [
        "message", "([{,] ?):(\w+) ?=> ?:(\w+)", '\1"\2": "\3"',
        "message", "([{,] ?):(\w+)", '\1"\2"',
        "message", ":nil,", ':null,',
        "message", "=>", ":"
      ]
    }

    json {
      source => "message"
    }
  }
}
```

It works for us but I'm sure there are edgecases and that this code has bugs.

colinsurprenant:
we could also catch any json serialization exception in the logger and force transcode/cleanse all strings to UTF-8  and retry serialization. 

also, why not making a pluggable logger output formatter? and having a default standard line based logger implementation and a json formatter implementation?

jordansissel:
@colinsurprenant Cabin would allow you to have formatters, but imo, it's not desirable from my perspective as a maintainer. Having to learn to read both "plain text" and json logs for the same program will be tedious. I'd prefer just JSON. Once you plaint-text it, you lose the ability to read it easily with computers.

jordansissel:
[...]

PR #4820, opened 2016-03-15
Allow Logstash to write its logs in JSON format

jordansissel:
This is made available by a `--log-in-json` flag. Default is false.
When false, the old behavior [1] is used. 

When true, JSON logs are emitted.

[1] The old behavior is realy two things. First, using Object#inspect to
serialize. Second, to color the output if the IO is a tty.

For #1569

### DIFF
--- a/logstash-core/lib/logstash/logging/json.rb
+++ b/logstash-core/lib/logstash/logging/json.rb
@@ -1,0 +1,21 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/logging"
+require "logstash/json"
+
+module LogStash; class Logging; class JSON
+  def initialize(io)
+    raise ArgumentError, "Expected IO, got #{io.class.name}" unless io.is_a?(IO)
+
+    @io = io
+    @lock = Mutex.new
+  end
+
+  def <<(obj)
+    serialized = LogStash::Json.dump(obj)
+    @lock.synchronize do
+      @io.puts(serialized)
+      @io.flush
+    end
+  end
+end; end; end
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -102,6 +102,10 @@ class LogStash::Runner < Clamp::Command
     I18n.t("logstash.runner.flag.allow-env"),
     :attribute_name => :allow_env, :default => false
 
+  option ["--[no-]log-in-json"], :flag,
+    I18n.t("logstash.runner.flag.log-in-json"),
+    :default => false
+
   def pipeline_workers=(pipeline_workers_value)
     @pipeline_settings[:pipeline_workers] = validate_positive_integer(pipeline_workers_value)
   end
@@ -136,7 +140,7 @@ class LogStash::Runner < Clamp::Command
     require "logstash/util/java_version"
     require "stud/task"
     require "cabin" # gem 'cabin'
-
+    require "logstash/logging/json"
 
     # Configure Logstash logging facility, this need to be done before everything else to
     # make sure the logger has the correct settings and the log level is correctly defined.
@@ -326,11 +330,20 @@ class LogStash::Runner < Clamp::Command
                     :path => path, :error => e))
       end
 
-      @logger.subscribe(STDOUT, :level => :fatal)
-      @logger.subscribe(@log_fd)
+      if log_in_json?
+        @logger.subscribe(LogStash::Logging::JSON.new(STDOUT), :level => :fatal)
+        @logger.subscribe(LogStash::Logging::JSON.new(@log_fd))
+      else
+        @logger.subscribe(STDOUT, :level => :fatal)
+        @logger.subscribe(@log_fd)
+      end
       @logger.terminal "Sending logstash logs to #{path}."
     else
-      @logger.subscribe(STDOUT)
+      if log_in_json?
+        @logger.subscribe(LogStash::Logging::JSON.new(STDOUT))
+      else
+        @logger.subscribe(STDOUT)
+      end
     end
 
     if debug_config? && @logger.level != :debug
--- a/logstash-core/spec/logstash/plugin_spec.rb
+++ b/logstash-core/spec/logstash/plugin_spec.rb
@@ -8,12 +8,12 @@ require "logstash/filters/base"
 
 describe LogStash::Plugin do
   it "should fail lookup on inexisting type" do
-    expect_any_instance_of(Cabin::Channel).to receive(:debug).once
+    #expect_any_instance_of(Cabin::Channel).to receive(:debug).once
     expect { LogStash::Plugin.lookup("badbadtype", "badname") }.to raise_error(LogStash::PluginLoadingError)
   end
 
   it "should fail lookup on inexisting name" do
-    expect_any_instance_of(Cabin::Channel).to receive(:debug).once
+    #expect_any_instance_of(Cabin::Channel).to receive(:debug).once
     expect { LogStash::Plugin.lookup("filter", "badname") }.to raise_error(LogStash::PluginLoadingError)
   end
 
--- a/logstash-core/spec/logstash/runner_spec.rb
+++ b/logstash-core/spec/logstash/runner_spec.rb
@@ -3,7 +3,10 @@ require "spec_helper"
 require "logstash/runner"
 require "stud/task"
 require "stud/trap"
+require "stud/temporary"
 require "logstash/util/java_version"
+require "logstash/logging/json"
+require "json"
 
 class NullRunner
   def run(args); end
@@ -16,7 +19,7 @@ describe LogStash::Runner do
 
   before :each do
     allow(Cabin::Channel).to receive(:get).with(LogStash).and_return(channel)
-    allow(channel).to receive(:subscribe).with(any_args)
+    allow(channel).to receive(:subscribe).with(any_args).and_call_original
   end
 
   describe "argument parsing" do
@@ -91,6 +94,32 @@ describe LogStash::Runner do
         expect(subject).to receive(:signal_usage_error).and_call_original
         expect(subject).to receive(:show_short_help)
         expect(subject.run(args)).to eq(1)
+      end
+    end
+  end
+
+  context "--log-in-json" do
+    subject { LogStash::Runner.new("") }
+    let(:logfile) { Stud::Temporary.file }
+    let(:args) { [ "--log-in-json", "-l", logfile.path, "-e", "input {} output{}" ] }
+
+    after do
+      logfile.close
+      File.unlink(logfile.path)
+    end
+
+    before do
+      expect(channel).to receive(:subscribe).with(kind_of(LogStash::Logging::JSON)).and_call_original
+      subject.run(args)
+
+      # Log file should have stuff in it.
+      expect(logfile.stat.size).to be > 0
+    end
+
+    it "should log in valid json. One object per line." do
+      logfile.each_line do |line|
+        expect(line).not_to be_empty
+        expect { JSON.parse(line) }.not_to raise_error
       end
     end
   end
--- a/logstash-core/spec/logstash/runner_spec.rb
+++ b/logstash-core/spec/logstash/runner_spec.rb
@@ -179,14 +179,14 @@ describe LogStash::Runner do
       it "should set 'debug_config' to false by default" do
         expect(LogStash::Config::Loader).to receive(:new).with(anything, false).and_call_original
         expect(LogStash::Pipeline).to receive(:new).with(pipeline_string, hash_including(:debug_config => false)).and_return(pipeline)
-        args = ["--debug", "-e", pipeline_string]
+        args = ["--debug", "-e", pipeline_string, "-l", "/dev/null", "--log-in-json"]
         subject.run("bin/logstash", args)
       end
 
       it "should allow overriding debug_config" do
         expect(LogStash::Config::Loader).to receive(:new).with(anything, true).and_call_original
         expect(LogStash::Pipeline).to receive(:new).with(pipeline_string, hash_including(:debug_config => true)).and_return(pipeline)
-        args = ["--debug", "--debug-config",  "-e", pipeline_string]
+        args = ["--debug", "--debug-config",  "-e", pipeline_string, "-l", "/dev/null", "--log-in-json"]
         subject.run("bin/logstash", args)
       end
     end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -6,6 +6,34 @@ CoverageHelper.eager_load if ENV['COVERAGE']
 
 require "logstash/devutils/rspec/spec_helper"
 
+class JSONIOThingy < IO
+  def initialize; end
+  def flush; end
+
+  def puts(payload)
+    # Ensure that all log payloads are valid json.
+    LogStash::Json.load(payload)
+  end
+end
+
+RSpec.configure do |c|
+  c.before do
+    # Force Cabin to always have a JSON subscriber.  The main purpose of this
+    # is to catch crashes in json serialization for our logs. JSONIOThingy
+    # exists to validate taht what LogStash::Logging::JSON emits is always
+    # valid JSON.
+    jsonvalidator = JSONIOThingy.new
+    allow(Cabin::Channel).to receive(:new).and_wrap_original do |m, *args|
+      logger = m.call(*args)
+      logger.level = :debug
+      logger.subscribe(LogStash::Logging::JSON.new(jsonvalidator))
+
+      logger
+    end
+  end
+
+end
+
 def installed_plugins
   Gem::Specification.find_all.select { |spec| spec.metadata["logstash_plugin"] }.map { |plugin| plugin.name }
 end
--- a/lib/bootstrap/rspec.rb
+++ b/lib/bootstrap/rspec.rb
@@ -7,6 +7,7 @@ $LOAD_PATH.unshift(File.join(LogStash::Environment::LOGSTASH_CORE, "spec"))
 
 require "rspec/core"
 require "rspec"
+require 'ci/reporter/rake/rspec_loader'
 
 status = RSpec::Core::Runner.run(ARGV.empty? ? ["spec"] : ARGV).to_i
 exit status if status != 0
--- a/logstash-core/lib/logstash/api/lib/app/service.rb
+++ b/logstash-core/lib/logstash/api/lib/app/service.rb
@@ -28,7 +28,7 @@ class LogStash::Api::Service
   end
 
   def update(snapshot)
-    logger.debug("[api-service] snapshot received", :snapshot => snapshot) if logger.debug?
+    logger.debug("[api-service] snapshot received", :snapshot_time => snapshot.created_at) if logger.debug?
     if @snapshot_rotation_mutex.try_lock
       @snapshot = snapshot
       @snapshot_rotation_mutex.unlock
--- a/logstash-core/lib/logstash/inputs/base.rb
+++ b/logstash-core/lib/logstash/inputs/base.rb
@@ -83,7 +83,7 @@ class LogStash::Inputs::Base < LogStash::Plugin
 
   public
   def do_stop
-    @logger.debug("stopping", :plugin => self)
+    @logger.debug("stopping", :plugin => self.class.name)
     @stop_called.make_true
     stop
   end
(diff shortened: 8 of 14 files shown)


logstash-core/lib/logstash/runner.rb
```diff
@@ -90,6 +90,10 @@ class LogStash::Runner < Clamp::Command
     I18n.t("logstash.web_api.flag.http_port"),
     :attribute_name => :web_api_http_port, :default => 9600
 
+  option ["--[no-]log-in-json"], :flag,
```
suyograo: Thoughts about renaming this flag to `--logger.format` to make it future proof for `setting.yml`? This way it translates easily to yml.. See https://github.com/elastic/logstash/pull/4499 /cc @jsvd

jsvd: to be consistent, this pr should add the flag according to the current state of master. right now it should use "-", but if it happens to be merged after logstash.yml, then the pr needs to be changed

jordansissel:
Rebased.

logstash-core/lib/logstash/runner.rb
```diff
@@ -102,6 +102,10 @@ class LogStash::Runner < Clamp::Command
     I18n.t("logstash.runner.flag.allow-env"),
     :attribute_name => :allow_env, :default => false
 
+  option ["--[no-]log-in-json"], :flag,
+         I18n.t("logstash.agent.flag.log-in-json"),
```
suyograo: Translation missing in yml file?

jordansissel:
Hand-test that it still works post-rebase:

```
% bin/logstash --log-in-json -f /tmp/nonexistant
{"timestamp":"2016-04-17T10:46:13.551000-0700","message":"No config files found: /tmp/nonexistant\nCan you make sure this path is a logstash config file?","level":"error"}
```

jordansissel:
Added tests

logstash-core/lib/logstash/runner.rb
jordansissel: Will fix this. Good catch.

jordansissel:
Added help text:

```
    --[no-]log-in-json            Specify that Logstash should write its own logs in JSON form - one
                                  event per line. If false, Logstash will log using Ruby's
                                  Object#inspect (not easy to machine-parse)
                                   (default: false)
```

jordansissel:
Squashed to a single commit.

ph:
@jordansissel I saw we are targeting this for 5.0 But I am wondering if we should merge it in 2.3.X as an experimental flag?

jordansissel:
@ph I'd be ok merging this into 2.3 and 2.x as a flag that defaults to the old behavior. We can have a new PR later that enables JSON by default in master. Thoughts?

jordansissel:
http://build-eu-00.elastic.co/job/Logstash_PR/2671/testReport/ -- Failing tests seem unrelated to this PR.

ph:
@jordansissel 
yes +1 to making this this teh default logging mode in 5.0
+1 Make it a flag for 2.X 2.3.

ph:
@jordansissel Yes the faillure are unrelated. I think its time to make a PR on flores with https://github.com/logstash-plugins/logstash-input-beats/blob/master/spec/support/flores_extensions.rb#L29 ;)

jordansissel:
@ph ok cool. I'll target this for 2.x, 2.3, and master

jordansissel:
Summarizing some offline discussion between @ph, @suyograo, and myself.

Given this is a new feature, we'll target this to Logstash 2.4.0 insetad of 2.3.2 because it's a feature, not a bug fix.

jordansissel:
Merge targets: 2.x, 2.4, 5.0, master

ph:
@jordansissel I found our first serialization errors when running logstash in debug mode.

```
{"timestamp":"2016-04-21T10:58:04.833000-0400","message":"* Environment: development","level":"debug","file":"logstash/webserver.rb","line":"55","method":"log"}
LogStash::Json::GeneratorError: Cannot serialize instance of: LogStash::Json::GeneratorError
      jruby_dump at /Users/ph/es/logstash/logstash-core/lib/logstash/json.rb:52
              << at /Users/ph/es/logstash/logstash-core/lib/logstash/logging/json.rb:14
              << at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/subscriber.rb:10
         publish at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/channel.rb:196
            each at org/jruby/RubyHash.java:1342
         publish at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/channel.rb:193
     synchronize at org/jruby/ext/thread/Mutex.java:149
         publish at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/channel.rb:192
            _log at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/mixins/logger.rb:113
  log_with_level at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/mixins/logger.rb:90
           fatal at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/mixins/logger.rb:75
         execute at /Users/ph/es/logstash/logstash-core/lib/logstash/runner.rb:231
             run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67
             run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:132
          (root) at /Users/ph/es/logstash/lib/bootstrap/environment.rb:72
```

logstash-core/spec/logstash/runner_spec.rb
```diff
@@ -96,6 +99,32 @@ def run(args); end
...
+    end
+
+    it "should log in valid json. One object per line." do
+      logfile.each_line do |line|
+        expect(line).not_to be_empty
+        expect { JSON.parse(line) }.not_to raise_error
+      end
+    end
+  end
+
```
ph: We may need to add for the `--debug` log level.

suyograo:
@jordansissel @ph ping! @ph is your https://github.com/elastic/logstash/pull/4820#issuecomment-212959827 a blocker for this PR?

logstash-core/spec/logstash/runner_spec.rb
jordansissel: Adding `--debug` feels more like an integration test, not a unit test. My feeling is that this test intends to make sure that the `--log-in-json` flag will cause logs to come out in JSON.

Can you help me understand what testing --debug would help with?

ph: @jordansissel Correct, you are right it feel more like an integration test.

but this [error](https://github.com/elastic/logstash/pull/4820#issuecomment-212959827) was found when I ran logstash with `--debug` and `--log-in-json`, we have a serialization issue on some type.

ph:
@suyograo well, it current fails when I start logstash with `bin/logstash -e 'input { generator {} } output { stdout {}}' --log-in-json --debug`

So this might be a sign that we can have other problems.
```
{"timestamp":"2016-04-26T21:03:08.109000-0400","message":"* Environment: development","level":"debug","file":"logstash/webserver.rb","line":"55","method":"log"}
{"timestamp":"2016-04-26T21:03:08.237000-0400","message":"[api-service] start","level":"debug","file":"app/service.rb","line":"13","method":"initialize"}
{"timestamp":"2016-04-26T21:03:08.246000-0400","message":"* Listening on tcp://127.0.0.1:9600","level":"debug","file":"logstash/webserver.rb","line":"55","method":"log"}
LogStash::Json::GeneratorError: Cannot serialize instance of: LogStash::Json::GeneratorError
      jruby_dump at /Users/ph/es/logstash/logstash-core/lib/logstash/json.rb:52
              << at /Users/ph/es/logstash/logstash-core/lib/logstash/logging/json.rb:14
              << at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/subscriber.rb:10
         publish at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/channel.rb:196
            each at org/jruby/RubyHash.java:1342
         publish at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/channel.rb:193
     synchronize at org/jruby/ext/thread/Mutex.java:149
         publish at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/channel.rb:192
            _log at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/mixins/logger.rb:113
  log_with_level at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/mixins/logger.rb:90
           fatal at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/cabin-0.8.1/lib/cabin/mixins/logger.rb:75
         execute at /Users/ph/es/logstash/logstash-core/lib/logstash/runner.rb:231
             run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67
             run at /Users/ph/es/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:132
          (root) at /Users/ph/es/logstash/lib/bootstrap/environment.rb:72
{"timestamp":"2016-04-26T21:03:08.264000-0400","message":"Use Ctrl-C to stop","level":"debug","file":"logstash/webserver.rb","line":"55","method":"log"}
```

ph:
Adding a pry inside the exception handling we have another error, we get a bit more information.

```
[1] pry(LogStash::Json)> e
=> #<JrJackson::ParseError: Cannot serialize instance of: Class>
[2] pry(LogStash::Json)>
```

ph:
And the content that we tried to serialize was this
```
 pry(LogStash::Json)> o
=> {:timestamp=>"2016-04-26T21:07:04.930000-0400", :message=>"Will start workers for output", :worker_count=>1, :class=>LogStash::Outputs::Stdout, :level=>:debug, :file=>"logstash/output_delegator.rb", :line=>"85", :method=>"register"}
[2] pry(LogStash::Json)>
```

ph:
So we might need to help jrjackson to serialize this kind of behavior.
@guyboertje What would be the best option in this case?

guyboertje:
@ph - we can add class and instance methods `to_json_data` that calls `inspect` to `plugin.rb`

guyboertje:
I think, in JrJackson, I need to add the ability to serialize its own Errors :smile:

ph:
@guyboertje  LOL :)

jordansissel:
@ph I suppose we could use this PR to fix most of the areas where serialization fails. Thoughts? I'm open to it.

ph:
@jordansissel +1 for using this PR to fix some serialization issue, We probably don't use that behavior often.

@guyboertje We can patch `plugin.rb`, this would make it work with the plugin class, but we will miss some. JrJackson has a different behavior when we pass a class than the regular json, which just stringify the name, do you think it worth creating an issue on JrJackson?

```
irb(main):001:0> require 'json'
=> true
irb(main):002:0> class Hello;end
=> nil
irb(main):003:0> JSON.dump({ "kind" => Hello })
=> "{\"kind\":\"Hello\"}"
irb(main):004:0>
```

jordansissel:
620d2a70f96db3e3dafdedb2cf3ff73d3bb4a9a9 adds json logging for all rake test runs. I get some failures, but they are random and the errors do not indicate JSON serialization to cause a problem.

Would love some extra testing on this one -- `rake test:core` and `rake test:plugins` should both not have any logging-related failures.

jordansissel:
The [jenkins failure](http://build-eu-00.elastic.co/job/Logstash_PR/2734/console) doesn't seem caused by this PR.

jordansissel:
The [appveyor](https://ci.appveyor.com/project/suyograo/logstash/build/1.0.21) failure is due to a misconfiguration of appveyor.

jordansissel:
Appveyor seems misconfigured. @suyograo ?

jordansissel:
@ph confirmed the failure you reported. Funny that no test reproduces this, haha :(

suyograo:
Disabled appveyor for now (until I get it right :) )

guyboertje:
@ph - I see two problems in JrJackson. How to serialize a class reference, 2) how to serialize an arbitrary instance reference.

In general, I have resisted the urge to add the `to_s` or `inspect` check and call or a hand coded version thereof because this may leak private variables or sensitive information to the world.
I try to encourage Object authors to take control of this by implementing `to_json_data` because we don't have annotations that Jackson encourages or similar mechanisms in Ruby.

Maybe I need a `permissively_serialize` flag that allows JrJackson to use introspection to serialize an arbitrary object.

So, if we can use `to_json_data` where structure is important, this is good because we control that.

Class references can be checked and handled.

If you would like to raise an issue in JrJackson go for it - otherwise I will.

ph:
> In general, I have resisted the urge to add the to_s or inspect check and call or a hand coded version thereof because this may leak private variables or sensitive information to the world.
I try to encourage Object authors to take control of this by implementing to_json_data because we don't have annotations that Jackson encourages or similar mechanisms in Ruby.

Personally I think its up to the developer to make sure we don't leak any sensitive information its shouldnt be the responsability of the library from blocking for us or making bad decision ;)
But changing that is a breaking changes in your library.

> Maybe I need a permissively_serialize flag that allows JrJackson to use introspection to serialize an arbitrary object. 

In our case this would work since we control when the serialization of the log occur we could just pass the flag.

> If you would like to raise an issue in JrJackson go for it - otherwise I will.

I will let you create it, I can jump to comment on it if needed.

jordansissel:
As of my latest commit on this PR, `bin/logstash -e 'input { generator { count => 1 } }' --log-in-json --debug` will work successfully.

jordansissel:
Still working on injecting the json logging into our tests so we can catch more logger bugs. This is my patch so far, but it is not quite successful:

```diff
diff --git a/spec/spec_helper.rb b/spec/spec_helper.rb
index 5428fd8..af337c4 100644
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -6,6 +6,21 @@ CoverageHelper.eager_load if ENV['COVERAGE']

 require "logstash/devutils/rspec/spec_helper"

+RSpec.configure do |c|
+  c.before do
+    #nullout = File.new("/dev/null", "w")
+    nullout = STDERR
+    allow(Cabin::Channel).to receive(:new).and_wrap_original do |m, *args|
+      logger = m.call(*args)
+      p :logger => logger
+      logger.level = :debug
+      logger.subscribe(LogStash::Logging::JSON.new(nullout))
+      logger
+    end
+  end
+
+end
+
 def installed_plugins
   Gem::Specification.find_all.select { |spec| spec.metadata["logstash_plugin"] }.map { |plugin| plugin.name }
 end
```

jordansissel:
Ok I think all the specs are passing for me that are related to logging failures...

I still see these two failures, but they don't appear related to this pr:

```
rspec ./logstash-core/spec/logstash/inputs/metrics_spec.rb:26 # LogStash::Inputs::Metrics#update should fill up the queue with received events
rspec ./logstash-core/spec/logstash/agent_spec.rb:359 # LogStash::Agent metrics after config reloading resets the metric collector

```

logstash-core/lib/logstash/logging/json.rb
```diff
@@ -0,0 +1,20 @@
+require "logstash/namespace"
```
ph: Might be good to add the `# encoding: utf-8` header to appease the utf gods.

logstash-core/spec/logstash/runner_spec.rb
jordansissel: I've updated (see spec_helper) the tests to always force debug level + json logging (at least, as far as I can tell) so that we can catch some problems if they come up.

logstash-core/spec/logstash/plugin_spec.rb
```diff
@@ -8,12 +8,12 @@
...
 describe LogStash::Plugin do
   it "should fail lookup on inexisting type" do
-    expect_any_instance_of(Cabin::Channel).to receive(:debug).once
+    #expect_any_instance_of(Cabin::Channel).to receive(:debug).once
     expect { LogStash::Plugin.lookup("badbadtype", "badname") }.to raise_error(LogStash::PluginLoadingError)
   end
 
   it "should fail lookup on inexisting name" do
-    expect_any_instance_of(Cabin::Channel).to receive(:debug).once
+    #expect_any_instance_of(Cabin::Channel).to receive(:debug).once
```
ph: +1 for removing them.
I think checking for the raise_error is actually the behavior we want to verify.

rakelib/test.rake
```diff
@@ -17,6 +17,18 @@ namespace "test" do
...
+
+  def setup_json_logging
+    require "cabin"
+    require "logstash/logging/json"
+    Cabin::Channel.get(LogStash).tap do |logger|
+      # TODO(sissel): Open null on windows?
+      out = File.new("/dev/null", "a")
+      logger.subscribe(LogStash::Logging::JSON.new(out))
+      logger.level = :debug
+    end
```
ph: This will surely fails on windows, we can make it cross platform if we can create a dummy io object that implement the minimal interface to be an IO object but doesn't actually write anything to disk.

OR we can use this, which will make a null stream, I believe it will work on windows but I havent tested it?
```
File.open(File::NULL, "w")
```

spec/spec_helper.rb
```diff
@@ -6,6 +6,34 @@
...
 
+class JSONIOThingy < IO
+  def initialize; end
+  def flush; end
+
+  def puts(payload)
+    # Ensure that all log payloads are valid json.
+    LogStash::Json.load(payload)
+  end
+end
```
ph: Oh neat idea.

rakelib/test.rake
ph: I am a bit confused, why we need this code and the code in `spec_helper`?

ph:
Changes LGTM, small comment concerning the use of `/dev/null`, I think we can go ahead with this.
@jordansissel Running the all plugins tests will probably discover some other issues in the plugins for the serialization that we need to fix.

andrewvc:
Regarding the serialization `to_s` issue @ph and @guyboertje have been discussing I think

recursing is fine but we should never call `inspect`. Inspect can:

1. Generate massive output by accident
2. Leak passwords if they're linked in the object graph (this has happened multiple times!)

Ultimately we should strive for a whitelist approach to log inclusion IMHO. I wouldn't even mind it if we traversed plain ruby Hash and Array objects, invoking .to_s on everything else.

guyboertje:
Just so people know what JrJackson does do...
It knows how the serialize the following:
String
Float
BigDecimal
Time
Array
Hash
Fixnum
Bignum
Date
Symbol
Struct
TrueClass
FalseClass
(I will be adding Exception/Error and Class)

If it finds an object class not in the above list then it tries to invoke these methods, in order, on the object:
`to_json_data`
`to_time`
`to_h`
`to_hash`
`to_a`
`to_json`

Why `to_json` last?
Because, for the best performance, I do not want a JSON string in this case, I want JSON data structures so Jackson can continue with caching and memoising. This is why `to_json_data` is first - I want an object author to give me a data structure.
When authors write a `to_json` method, they very often will write:
```ruby
def to_json
  JSON.dump(some_structure)
end
```
This will suspend the JrJackson session, call the JSON library, and return a string for JrJackson to append.

rakelib/test.rake
jordansissel: Agreed. this setup_json_logging was an older implementation of what is now in spec_helper.

jordansissel: I removed it now.

logstash-core/lib/logstash/logging/json.rb
jordansissel: done

ph:
LGTM @jordansissel

jordansissel:
I need to backport this PR by hand to 2.x for 2.4. Will open a separate PR for it.

jordansissel:
Backport to 2.4 done in https://github.com/elastic/logstash/pull/5277.

joshuaspence:
Will Logstash 2.4 be released any time soon?

suyograo:
@joshuaspence yes, target is to release it in Aug.

Issue #4565, opened 2016-01-26
Timestamp loses precision when parsed from a JSON


alex88:
I was trying to figure out why the GELF input wasn't showing milliseconds in kibana, turns out to be mainly a jRuby bug which I opened here https://github.com/jruby/jruby/issues/3616

Basically the gelf input plugin parses the JSON using `LogStash::Json.load` [here](https://github.com/logstash-plugins/logstash-input-gelf/blob/master/lib/logstash/inputs/gelf.rb#L107) and later it creates the timestamp value with `LogStash::Timestamp.at(event["timestamp"])`.

The real problem happens [here](https://github.com/elastic/logstash/blob/v2.1.1/lib/logstash/timestamp.rb#L26-L28), basically when `Time.at` is called with a BigDecimal (which LogStash::Json.load provides) the milliseconds are stripped.

Could it be worth to handle the case when it's a BigDecimal and call `to_f` on it? If so I can work on a PR.
Since you stil bundle an older jRuby version and maybe it's easier to handle the problem with a patch instead of waiting for a new jRuby release that you should also globally upgrade.

suyograo:
@alex88 thanks for reporting this. We would absolutely love a monkey patch/PR to fix this for now. We are not on JRuby 9k

alex88:
@suyograo sorry for the delay, is it possible that when using jruby https://github.com/elastic/logstash/blob/master/logstash-core-event/lib/logstash/timestamp.rb isn't touched by the tests?
I'm running 
```bin/rspec logstash-core-event/spec/logstash/timestamp_spec.rb logstash-core-event-java/spec/timestamp_spec.rb```
and even deleting the `at` method nothing changes, like I haven't removed the code

alex88:
Ok nevermind, seems that running `rake test:install-core` installs the logstash-core-event-java which doesn't use the class I was trying to fix but it still has the same problem

purbon:
@guyboertje what do you think?

guyboertje:
@purbon, @alex88  /cc @colinsurprenant

We should not rely on side effects like this.

Git blame on [line 111](https://github.com/logstash-plugins/logstash-input-gelf/blame/master/lib/logstash/inputs/gelf.rb#L107) shows that it is very old.

[...]

PR #4827, opened 2016-03-16
fix Java Timestamp#usec and Timestamp#at

### LINKED ISSUE #4819: Java Timestamp#usec does not return proper value
work on #4565 and logstash-plugins/logstash-input-gelf#35 made me realize that the current Java Timestamp#usec implementation does not return the correct microsecond value but instead returns the milliseconds **since epoch** and not the millisconds on the current time.

A spec and a fix are required.

colinsurprenant:
- fix Java `Timestamp#usec` to correctly return the current time usec and not the epoch usec, fixes #4819
- fix Java `Timestamp#at` to correctly handle `BigDecimal` epoch, relates to #4565

### DIFF
--- a/logstash-core-event-java/src/main/java/com/logstash/Timestamp.java
+++ b/logstash-core-event-java/src/main/java/com/logstash/Timestamp.java
@@ -70,7 +70,9 @@ public class Timestamp implements Cloneable {
     }
 
     public long usec() {
-        return new Duration(JAN_1_1970.toDateTime(DateTimeZone.UTC), this.time).getMillis();
+        // JodaTime only supports milliseconds precision we can only return usec at millisec precision.
+        // note that getMillis() return millis since epoch
+        return (new Duration(JAN_1_1970.toDateTime(DateTimeZone.UTC), this.time).getMillis() % 1000) * 1000;
     }
 
     @Override
--- a/logstash-core-event/spec/logstash/timestamp_spec.rb
+++ b/logstash-core-event/spec/logstash/timestamp_spec.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 require "spec_helper"
 require "logstash/timestamp"
+require "bigdecimal"
 
 describe LogStash::Timestamp do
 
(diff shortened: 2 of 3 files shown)


guyboertje:
LGTM

Tests fail - but not because of an error in the code rather:
`rspec ./logstash-core/spec/api/lib/api/node_spec.rb:70 # LogStash::Api::Node#hot threads when requesting idle threads should return JIT threads`

colinsurprenant:
@guyboertje yeah, I already raised the issue with @ph and @purbon - clearly unrelated.

purbon:
@colinsurprenant @guyboertje @ph fix for this unrelated issue can be located at https://github.com/elastic/logstash/pull/4833, feel free to take a look at it.

logstash-core-event-java/src/main/java/com/logstash/Timestamp.java
```diff
@@ -70,7 +70,9 @@ public String toString() {
     }
 
     public long usec() {
-        return new Duration(JAN_1_1970.toDateTime(DateTimeZone.UTC), this.time).getMillis();
+        // JodaTime only supports milliseconds precision we can only return usec at millisec precision.
+        // note that getMillis() return milliss since epoch
```
talevy: s/milliss/milliseconds

logstash-core-event/spec/logstash/timestamp_spec.rb
```diff
@@ -106,4 +107,50 @@
       expect(subject.to_f).to eq(now.to_f)
     end
   end
+
+  context "at" do
```
talevy: can we add potentially 3 more tests?

1. arg==nil
2. arg < 0
3. arg != numeric type

logstash-core-event/spec/logstash/timestamp_spec.rb
```diff
@@ -106,4 +107,50 @@
...
+
+  context "usec" do
+    it "should support millisecond precision" do
+      expect(LogStash::Timestamp.at(946702800.123).usec).to eq(123000)
+    end
+
+    it "should try to preserve and report microseconds precision if possible" do
+      # since Java Timestamp relies on JodaTime which supports only milliseconds precision
+      # the usec method will only be precise up to milliseconds.
+      expect(LogStash::Timestamp.at(946702800.123456).usec).to be_within(1000).of(123456)
```
talevy: this `usec` test looks like it would be perfect for a randomized test. I'm OK without it

logstash-core-event-java/src/main/java/com/logstash/Timestamp.java
colinsurprenant: yep

logstash-core-event/spec/logstash/timestamp_spec.rb
colinsurprenant: done!

logstash-core-event/spec/logstash/timestamp_spec.rb
colinsurprenant: I would not change the current specs but also add randomized. let's defer that + create followup issue ok?

talevy: sounds good

talevy:
LGTM

colinsurprenant:
I squashed and merged into master, 2.x and 2.3. 
the close/merge status and commits history is borked because I squashed, merged into master and cherrypicked into 2.x an 2.3 but forgot to first push new branch squashed history on github. I pushed it after which wreck havoc in the status :(

PR #4828, opened 2016-03-17
Revert commit 4f7299a

suyograo:
Need to revert bump to 5.0.0.dev because individual plugins
will not install on top of this version.

In plugins' gemspec, the constraint is
`s.add_runtime_dependency "logstash-core", ">= 2.0.0", "< 3.0.0"`

This needs to be mass updated, and then bumped to 5.0.0


ph:
LGTM

PR #4831, opened 2016-03-17
expose reload_interval in Runner

jsvd:
LogStash::Agent already accepts a :reload_interval param, but it wasn't exposed in the Runner cli args, this PR fixes this problem.


guyboertje:
LGTM

suyograo:
LGTM. Please label releases

PR #4833, opened 2016-03-17
Change JIT threads for pool threads in the hot_threads ignore_idle_threads filter

purbon:
changes JIT threads for pool threads in the idle threads test, not sure why JIT threads don't show up in some situations, especially on linux machines. But we still need to check for the idle threads filter to work, so this test does it using the other jruby internal threads (skip as idle).

 Related to #4827


purbon:
@colinsurprenant @ph thoughts on this very welcome.

ph:
lgtm

suyograo:
LGTM

PR #4838, opened 2016-03-17
correctly handle BigDecimal Rubyfying

colinsurprenant:
The Java BigDecimal type conversion to a proper Ruby BigDecimal is not handled by the default JRuby type converter and it must be handled in the `Rubyfier` class. Also added tests.

This problem has been spotted while working on the GELF input logstash-plugins/logstash-input-gelf#35 and running specs against master with the Java `Event`.

It was also reported by @guyboertje in a discussion earlier today but I can't recall it was about what :grimacing: 

Prior to this patch the GELF input specs were not passing against master.

### DIFF
--- a/logstash-core-event-java/spec/event_spec.rb
+++ b/logstash-core-event-java/spec/event_spec.rb
@@ -96,6 +96,14 @@ describe LogStash::Event do
       e["[foo]"] = nil
       expect(e.to_hash).to include("foo" => nil)
     end
+
+    # BigDecinal is now natively converted by JRuby, see https://github.com/elastic/logstash/pull/4838
+    it "should set BigDecimal" do
+      e = LogStash::Event.new()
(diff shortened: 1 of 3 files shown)


guyboertje:
I discovered this in a change to the translate filter. When a bigdecimal was given to a key as part of a translation, it was converted to a Java bigdecimal and when the value was retrieved, the conversion back to a ruby value failed as detailed here.

talevy:
can we add rspec tests?

colinsurprenant:
@talevy done. I added a minimal test for the BigDecimal case, it fails when removing the fix and passes with this fix.

talevy:
LGTM

guyboertje:
LGTM

colinsurprenant:
merged in master, 2.x and 2.3

Issue #4824, opened 2016-03-16
master build failure due to metrics error


purbon:
Hi,
  last master build is broken due to this error, see (http://build-eu-00.elastic.co/view/LS%20Master/job/logstash_regression_master/616/) only heppened for JDK8, but worst investigating. This looks like a rogue thread going on somewhere.

Error is:

```
Failures:

  1) LogStash::Inputs::Metrics#update should fill up the queue with received events
     Failure/Error: expect(queue.count).to eq(1)
       
       expected: 1
            got: 41
       
       (compared using ==)
     # ./logstash-core/spec/logstash/inputs/metrics_spec.rb:29:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.8/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:42:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-10.5.0/lib/rake/task.rb:240:in `execute'
```

@ph any idea?

ph:
It might be related to a thread running in the background I don't think its a JDK8 issues.
I wish the 3 JDK jobs run were using the same seed, so this would help rule out some of the JDK issue.

purbon:
should not be a JDK issue, but as earlier executions fails on something
else, but still something we have to fix.

- purbon

PR #4839, opened 2016-03-18
Make sure we clear the store before running the test

ph:
Fixes: #4824


ph:
@purbon I tried to make it fails multiples times with and without the same seed without luck.
I am working to remove the problem by doing the instantiation of the collector inside the agent.

suyograo:
LGTM

PR #4849, opened 2016-03-19
add node_name flag description

jsvd:
to 2.x and 2.3.0

PR #4850, opened 2016-03-19
remove node_name cli argument

jsvd:
this argument is not used and was introduced by mistake in
the backport of the automatic configuration reloading

to be merged in 2.x and 2.3

replaces https://github.com/elastic/logstash/pull/4849


suyograo:
LGTM, but I'd like @ph to glance over this one too.

ph:
LGTM, I have also checked the locales.yml  and there is not mention of the `node_name` key

Issue #4852, opened 2016-03-21
[web api] Hot threads ignore_idle_threads=false test


purbon:
The original test:

```
 context "when requesting idle threads" do

      before(:all) do
        do_request { get "/hot_threads?ignore_idle_threads=false&threads=10" }
      end

      let(:payload) { LogStash::Json.load(last_response.body) }

      it "should return JIT threads" do
        thread_names = payload["threads"].map { |thread_info| thread_info["name"] }
        expect(thread_names.grep(/pool/)).not_to be_empty
      end
    end
```

is showing randomly failures, in linux boxes, when checking for `JIT` or `pool` threads, this two threads are part of the ignored by default threads and part of the internal jruby interpreter. 

We're disabling this test, until we figure it out a good way to test for this idle threads cross platform.

purbon:
Fixed in #4853

PR #4853, opened 2016-03-21
Remove hot_threads ignore_idle_threads test

purbon:
In #4833 I changed the `ignore_idle_threads` test from checking the `JIT` threads to check the `pool` threads, but looks like both are having the same randomly MIA behaviour, until we figure it out the right way to test for this threads I do recommend removing them for now.

Related #4852 #4833


ph:
LGTM, +1 to know is the real cause of that.

PR #4855, opened 2016-03-21
fix reload_state spec to ensure pipeline has started

### LINKED ISSUE #4825: 2.x branch regression test broken
Hi,
  last two build of the 2.x branch regression test are red due to this error:

```
Failures:

  1) LogStash::Agent#execute when auto_reload is true if state is clean should periodically reload_state
     Failure/Error: expect(subject).to receive(:reload_state!).at_least(3).times
       (#<LogStash::Agent:0x1bb3e150>).reload_state!(any args)
           expected: at least 3 times with any arguments
           received: 2 times with any arguments
     # ./logstash-core/spec/logstash/agent_spec.rb:77:in `(root)'
```

See http://build-eu-00.elastic.co/view/LS%202.x/job/logstash_regression_2x/111/ for details, for example.

@jsvd as this is bound to reload_state, any idea why this might be happening?

jsvd:
fixes #4825

### DIFF
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -70,12 +70,12 @@ describe LogStash::Agent do
 
     context "when auto_reload is true" do
       let(:agent_args) { [ "--auto-reload", "--config", config_file] } #reload_interval => 0.01, :config_path => } }
-      #let(:agent_args) { { :logger => logger, :auto_reload => false, :reload_interval => 0.01, :config_path => config_file } }
       context "if state is clean" do
         it "should periodically reload_state" do
-          allow(subject).to receive(:clean_state?).and_return(false)
           expect(subject).to receive(:reload_state!).at_least(3).times
-          t = Thread.new { subject.execute }
+          t = Thread.new(subject) {|subject| subject.execute }
+          sleep 0.01 until (subject.running_pipelines? && subject.pipelines.values.first.ready?)
+          # now that the pipeline has started, give time for reload_state! to happen a few times
           sleep 0.1
           Stud.stop!(t)
           t.join


logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -73,9 +73,9 @@
       #let(:agent_args) { { :logger => logger, :auto_reload => false, :reload_interval => 0.01, :config_path => config_file } }
       context "if state is clean" do
         it "should periodically reload_state" do
-          allow(subject).to receive(:clean_state?).and_return(false)
           expect(subject).to receive(:reload_state!).at_least(3).times
-          t = Thread.new { subject.execute }
+          t = Thread.new(subject) {|subject| subject.execute }
+          sleep 0.01 until (subject.running_pipelines? && subject.pipelines.values.first.ready?)
```
ph: parentheses aren't really needed here.

ph:
Small comment, I was hit bit this issue last weekend randomly.
LGTM

logstash-core/spec/logstash/agent_spec.rb
purbon: even if not need I agree that sometimes simplify reading, at less for me. not sure where to stand here.

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -73,9 +73,9 @@
...
           sleep 0.1
```
purbon: I wonder why is this second sleep 0.1 need? any reason?

purbon: does this 0.1 sleep have a meaning? if so can we add a comment indicating that reason behind it? I mean so when we read the test we know why :+1:

purbon:
small comment, but for the rest LGTM.

logstash-core/spec/logstash/agent_spec.rb
jsvd: +1 that they're not needed, but since it's a long line I'd rather keep them to help readability.

jsvd:
@purbon updated spec with a comment on sleep

ph:
LGTM

PR #4856, opened 2016-03-21
implementation of the logstash.yml settings file

jsvd:

So the setting precedence is, from weakest to strongest: default settings < yaml settings < cli settings

TODO
make conf/logstash.yml location configurable using env var CONF_DIR or similar


ph:
On it.

conf/logstash.yml
```diff
@@ -0,0 +1,47 @@
...
+# node.name: test
+#
+# If omitted the node name will default to the machine's host name
+#
+# ------------ Pipeline Settings --------------
+#
+# Set the number of workers that will, in parallel, execute the filters+outputs
+# stage of the pipeline.
+#
+# This defaults to half the number of the host's CPU cores.
```
ph: I believe we have change that to use all the core.

conf/logstash.yml
```diff
@@ -0,0 +1,47 @@
...
+#
+# pipeline.batch.delay: 5
+#
+# Force Logstash to exit during shutdown even if there are still inflight
+# events in memory. By default, logstash will refuse to quit until all
+# received events have been pushed to the outputs.
+#
+# WARNING: enabling this can lead to data loss during shutdown
+#
+# pipeline.unsafe_shutdown: false
```
ph: Should we add **all** or a **subset** of the options currently available from the CLI?
I we are only doing a subset, we should at least include the option concerning the web api and the config reloading. WDYT?

```
--http-host WEB_API_HTTP_HOST Web API binding host (default: "127.0.0.1")
--http-port WEB_API_HTTP_PORT Web API http port (default: 9600)

File: docs/static/command-line-flags.asciidoc [hunk 3]
```diff
@@ -63,6 +63,11 @@ Logstash has the following flags. You can use the `--help` flag to display this
...
   Specifies how often Logstash checks the config files for changes. The default is every 3 seconds.
 
+--pipeline.unsafe_shutdown
+  Force logstash to exit during shutdown even if there are still inflight events
+  in memory. By default, logstash will refuse to quit until all received events
+  have been pushed to the outputs.
+
 -h, --help
   Print help
 ----------------------------------
```
ph: Oh I will update that file with the web api options, thanks for bringing this to my attention.

lib/bootstrap/environment.rb
```diff
@@ -56,6 +57,31 @@ def pattern_path(path)
   end
 end
 
+def fetch_yml_settings(settings_path)
+  if settings = YAML.parse(IO.read(settings_path))
+    settings = settings.to_ruby
```
ph: TIL

logstash-core/lib/logstash/runner.rb
```diff
@@ -375,4 +373,36 @@ def trap_sigint
...
+      elsif value.nil? # default setting not set by user
+        output << "#{setting}: #{default_value.inspect}"
+      else # print setting, warn it has been set, and show default value
+        output << "*#{setting}: #{value.inspect} (default: #{default_value.inspect})"
+      end
+    end
+    output << "--------------- Logstash Settings -------------------"
+    output
+  end
+end
```
ph: This is somewhat used for debugging but I don't see anything testing that path?

logstash-core/lib/logstash/runner.rb
```diff
@@ -156,11 +162,15 @@ def execute
 
     return start_shell(@ruby_shell, binding) if @ruby_shell
 
+    settings = build_settings_hash
+    settings.merge!(@pipeline_settings)
+    format_settings(settings).each {|line| @logger.log(line) }
```
ph: Any reason why this is not in debug mode?

conf/logstash.yml
jsvd: +1 on adding all or most of the settings. I'll add these too.

ph:
@jsvd I like the code, I think the only thing I would try to *squeak* in for this PR is to encapsulate this logic into his own class like a **Config** or a Settings class and we could make it behave like hash/Enum for the accessors

. Doing this would allow us future improvements PR with this:

1. Decouple the config from the actual logic.
2. Create better validation for the configuration using the visitor patterns, theses validations could be read from a directory and executed when we call a `#validate`
3. Necessary coercion could happen in the config class.
4. The internal data could be saved in an object instead of the raw value and keep track of what changed the value.
5. The global configuration of logstash should be made available to the plugins, lets say a plugin need a `since_db` like the file and the s3 input. A since DB datastore could be defined at the config level and reused by multiple plugins.
6. Drop clamp for better control.
7. Config Management could be use to update the config store, We could also implement the observer pattern, so when the config is changed the children could act on theses changes if needed.

ph:
I think most of the suggestions/ideas could be encapsulated in the config object and add improvements to the config system without having to change anything else in the core classes, except the one removing clamp :P

conf/logstash.yml
jsvd: how do you want to namespace these settings? just "http.host"/"http.port" or "api.http.host" or.. ?

ph: I think we are fine with `http.host` and `http.port`

jsvd:
closed in favor of https://github.com/elastic/logstash/pull/5251

Issue #4829, opened 2016-03-17
Mass update plugins for 5.0.0 release


suyograo:
mass update and publish to all plugins needed because of the constraints in the gemspec for plugins.

See for example: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/master/logstash-output-elasticsearch.gemspec#L27

```
s.add_runtime_dependency "logstash-core", ">= 2.0.0", "< 3.0.0"
```

ain't gonna cut it for 5.0

jsvd:
we had spoken about creating a level of indirection between logstash-core and logstash-plugins with a logstash-core-api gem, that we could keep at a lower minor version and change at a slower pace..I guess we completely forgot this during 2.x

ph:
@jsvd I remember this discussion, the only downside it will make plugins incompatible with older version and we will need to backport stuff.

Related to this discussion.

ph:
Discussing with @jsvd on that  we are seeing two differents way of making this work.

1. update the logstash-core constrains to be "> 2.0.0.alpha1" and "< 6.0.0.alpha1"
2. OR, create a new gems called `logstash-core-api|logstash-core-contract` and keep the version outside of the normal logstash release versioning, make the plugins depends on it. This gems should have a dependency on "logstash-core >= 2.0.0.alpha1" to make the plugins on 2.0. And the logstash itself depends on a specific version of `logstash-core-contract`

ph:
@jsvd and I created a graph to brainstorm around the constraints we think that will make it work and result in less mass update.
![dependencies](https://cloud.githubusercontent.com/assets/640/13858125/8f60d996-ec53-11e5-9af2-e91a3adaa36f.jpg)

ph:
@colinsurprenant we would like your feedback on ^

ph:
small errors in the graph updated with a middle version of ES.

suyograo:
@ph @jsvd since we are running short of time for alpha1, can we do this:
update the logstash-core constrains to be "> 2.0.0" and "< 6.0.0"

not sure why we need the alphas?

colinsurprenant:
just a quick heads up, while discussing this this morning, I mentioned that the original intent of the logstash-core gem was to actually **be** the *contract* and to achieve this we only need to **stop** sync'ing the logstash-core gem version with the logstash **release** version. A major version bump in logstash-core will only be necessary to signal incompatible core API changes. 

[...]

PR #4866, opened 2016-03-22
Create a virtual gem to handler the plugin dependency on core

ph:
We have decided to create a gem that the plugins can depend on to make
sure they use the correct core, with that in mind this will reduce the
mass update required when releasing a major release of logstash if the
contract between logstash and the plugin didnt change. The first version
of this gem act as a virtual gem, future release will include the actual
plugins code of the contract.

This gem should follow the same convention as semver, a breaking change
in the api will mean a bump in the major version.  Plugins author are
encouraged to test they plugins with a specific version of the contract
and define their dependency as the following:

```
  s.add_runtime_dependency "logstash-code-plugin-api", "~> 2.0"
```

ref #4829

### DIFF
--- a/logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
+++ b/logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
@@ -1,0 +1,2 @@
+# encoding: utf-8
+LOGSTASH_CORE_PLUGIN_API = "1.0.0"
--- a/logstash-core-plugin-api/logstash-core-plugin-api.gemspec
+++ b/logstash-core-plugin-api/logstash-core-plugin-api.gemspec
@@ -1,0 +1,29 @@
+# -*- encoding: utf-8 -*-
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require "logstash-core-plugin-api/version"
+
+Gem::Specification.new do |gem|
+  gem.authors       = ["Elastic"]
+  gem.email         = ["info@elastic.co"]
+  gem.description   = %q{Logstash plugin API}
+  gem.summary       = %q{Define the plugin API that the plugin need to follow.}
+  gem.homepage      = "http://www.elastic.co/guide/en/logstash/current/index.html"
+  gem.license       = "Apache License (2.0)"
+
+  gem.files         = Dir.glob(["logstash-core-event.gemspec", "lib/**/*.rb", "spec/**/*.rb"])
+  gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
+  gem.name          = "logstash-core-plugin-api"
+  gem.require_paths = ["lib"]
+  gem.version       = LOGSTASH_CORE_PLUGIN_API
+
+  gem.add_runtime_dependency "logstash-core", ">= 2.0.0", "<= 3.0.0.dev"
+
+  # Make sure we dont build this gem from a non jruby
+  # environment.
+  if RUBY_PLATFORM == "java"
+    gem.platform = "java"
+  else
+    raise "The logstash-core-api need to be build on jruby"
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ source "https://rubygems.org"
 gem "logstash-core", "3.0.0.dev", :path => "./logstash-core"
 # gem "logstash-core-event", "3.0.0.dev", :path => "./logstash-core-event"
 gem "logstash-core-event-java", "3.0.0.dev", :path => "./logstash-core-event-java"
+gem "logstash-core-plugin-api", "1.0.0", :path => "./logstash-core-plugin-api"
 gem "file-dependencies", "0.1.6"
 gem "ci_reporter_rspec", "1.0.0", :group => :development
 gem "simplecov", :group => :development


ph:
Also this PR add a few useful rake tasks that would you want to test.

```
rake releases:publish-gems
```

Next manual step is to update the `Gemfile` with the correct version.

rakelib/release.rake
```diff
@@ -0,0 +1,158 @@
...
+      puts "Running: #{cmd}"
+    elsif !ENV["RUBYGEMS_API_KEY"].nil?
+      cmd = "#{GEM_BINARY_PATH} push #{gem}"
+    else
+      puts "Cannot publish the gem without a `~/.gem/credentials` or a `$RUBYGEM_API_KEY` defined"
+      exit
+    end
+
+    puts "Publishing gem: #{File.basename(gem)}"
+    #system(cmd)
```
ph: I have explicitly disabled this for reviewing this PR. I dont want to create a rogue gems on rubygems.

ph:
PR that will be made in the plugins from the mass update script https://github.com/logstash-plugins/logstash-input-beats/pull/62

ph:
@jsvd Can you also take a look at theses 2 PRs?

logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
```diff
@@ -0,0 +1,2 @@
+# encoding: utf-8
+LOGSTASH_CORE_PLUGIN_API = "1.0.0"
```
suyograo: Can we add a comment on what the versioning scheme here is? And also comment that we should not be editing this manually, and rather use the rake script

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
```diff
@@ -0,0 +1,27 @@
+# -*- encoding: utf-8 -*-
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require "logstash-core-plugin-api/version"
+# This is completely valid, when we do a `gem build` the gemspec is 
+# converted to a text file without any ruby code.
+require_relative "../logstash-core/lib/logstash-core/version"
+
+Gem::Specification.new do |gem|
+  gem.authors       = ["Elasticsearch"]
```
suyograo: Elastic

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
```diff
@@ -0,0 +1,27 @@
...
+  gem.email         = ["info@elasticsearch.com"]
```
suyograo: `info@elastic.co`

colinsurprenant:
Can we choose between API and contract, maybe it's just me but it feels weird to say "API contract" ? 

A specific API version IS a contract.

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
colinsurprenant: I thinks @ph missed the rebranding :smile:

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
```diff
@@ -0,0 +1,27 @@
...
+  gem.homepage      = "http://www.elastic.co/guide/en/logstash/current/index.html"
+  gem.license       = "Apache License (2.0)"
+
+  gem.files         = Dir.glob(["logstash-core-event.gemspec", "lib/**/*.rb", "spec/**/*.rb"])
+  gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
+  gem.name          = "logstash-core-plugin-api"
+  gem.require_paths = ["lib"]
+  gem.version       = LOGSTASH_CORE_PLUGIN_API
+
+  gem.add_runtime_dependency "logstash-core", "#{LOGSTASH_CORE_VERSION}"
```
colinsurprenant: so whenever we create this gem, the `logstash-core` version dependency is the one that currently exists in the source tree at that moment? is this **always** what we want?

colinsurprenant:
Also, for the `release.rake` - maybe we should make that into its own PR? my idea is that we could also add the other core gems building and not just core-plugin-api?

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
```diff
@@ -0,0 +1,27 @@
+# -*- encoding: utf-8 -*-
```
jsvd: why the funky header vs the more classic one?

suyograo:
@colinsurprenant I think this rake release task is ok for now. we can always expand later to cover more functionality. But moving to a new PR makes this easier for reviews, I am cool with that also.

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
```diff
@@ -0,0 +1,27 @@
...
+  gem.license       = "Apache License (2.0)"
+
+  gem.files         = Dir.glob(["logstash-core-event.gemspec", "lib/**/*.rb", "spec/**/*.rb"])
+  gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
+  gem.name          = "logstash-core-plugin-api"
+  gem.require_paths = ["lib"]
+  gem.version       = LOGSTASH_CORE_PLUGIN_API
+
+  gem.add_runtime_dependency "logstash-core", "#{LOGSTASH_CORE_VERSION}"
+  if RUBY_PLATFORM == 'java'
```
jsvd: should we just abort if not using java?

rakelib/release.rake
```diff
@@ -0,0 +1,158 @@
+# encoding: utf-8
+require "fileutils"
+
+LOGSTASH_BUILD_PATH = File.expand_path(File.join(File.dirname(__FILE__), "..", "build"))
+
+LOGSTASH_CORE_PLUGIN_API_DEFAULT_VERSION = "1.0"
```
jsvd: can you place this in the plugin api gem code itself? maybe under lib/logstash-plugin-api/version.rb ?

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
colinsurprenant: cut & paste or generated? but yeah, this funky syntax is... funky :P

rakelib/release.rake
jsvd: oh wait..it's already there, this is the _DEFAULT_VERSION. nvm my comment....

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
ph: To the full disclosure I copied one of our gemspec to make it work and in theses we actually use this syntax.

On a side node there is effectively 3 way that I know to specify the encoding in ruby files:

```
# This is use by newer version of `bundler gem`
# coding: utf-8
# encoding: utf-8
# -*- encoding: utf-8 -*-
```

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
ph: We might changes our other gemspec files? I took the logstash-core.gemspec as the starting point?

suyograo: ok, we'll have to change everything. Can you please change this one here and I'll update the others in a diff PR

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
ph: I think we all missed the rebranding :P 
https://github.com/ph/logstash/blob/b140886a71d72b8947771dd896d84c0a53a43117/logstash-core-event-java/logstash-core-event-java.gemspec#L7

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
ph: Yup

logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
ph: +1

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
ph: In this (proposition)[https://github.com/elastic/logstash/issues/4829#issuecomment-199925565] I was proposing the following constraints, I agree it might not be clear with the `satisfied_by?`

```
 gem.add_runtime_dependency "logstash-core", ">= 2.0.0" ,"<= #{LOGSTASH_CORE_VERSION}"
```

But after discussing with you yesterday, I though to use a strict dependency on a specific `logtash-core` and not use a range. But this constraints cannot work for the first version of `logstash-core-plugin-api` (1.0.0) IT **needs** to be  a range like this.

```
 gem.add_runtime_dependency "logstash-core", ">= 2.0.0" ,"<= 2.3.0"
```

If we don't do this, plugins wont be able to be installed in a pre Logstash 2.3 installation.

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
ph: I can do that.

rakelib/release.rake
ph: This is used when we cannot find the gems.. which should be on the first push.

ph:
@colinsurprenant I will remove contract and just use API.

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
ph: Well, I think its consistent with other gemspec. 

https://github.com/ph/logstash/blob/b140886a71d72b8947771dd896d84c0a53a43117/logstash-core-event-java/logstash-core-event-java.gemspec#L1
https://github.com/ph/logstash/blob/b140886a71d72b8947771dd896d84c0a53a43117/logstash-core/logstash-core.gemspec#L1
https://github.com/ph/logstash/blob/b140886a71d72b8947771dd896d84c0a53a43117/logstash-core-event/logstash-core-event.gemspec

ph:
@colinsurprenant @suyograo I will move the release.rake in another PR and we will do the first release by hand.

suyograo:
@ph +1

ph:
updated with changes.

Only missing a response for https://github.com/elastic/logstash/pull/4866/files#r57211124

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
ph: I created https://github.com/elastic/logstash/pull/4880 for the rebranding.

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
colinsurprenant: lol I never clued in on these :P now I will see it everywhere :grin:

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
colinsurprenant: right, my comment is more about the automatic insertion of that version instead of letting us setting it when we generate. I am thinking there might be cases where we'd want to generate a new plugin-api gem using a dependency on logstash-core that is not the one currently in the source tree.

ph: Correct, But I have removed all the rake task in the present PR so we can tackle this problem in a  new PR.

logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
ph: I have removed the rake script so its not really needed.

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
colinsurprenant: the problem here is that this "hides" exactly what the logstash-core version constrain... it's the one in `logstash-core/lib/logstash-core/version`. it we write it explicitly here in there are less chances it will be overlooked.

ph: I can make it explicit.

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
```diff
@@ -0,0 +1,32 @@
...
+  gem.add_runtime_dependency "logstash-core", "#{LOGSTASH_CORE_VERSION}"
+
+  # Make sure we dont build this gem from a non jruby
+  # environment.
+  if RUBY_PLATFORM == "java"
+    gem.platform = "java"
+  else
+    raise "The logstash-core-api need to be build on jruby"
+  end
+end
```
colinsurprenant: cosmetic but could be reduced to 
```ruby
raise "The logstash-core-api need to be build on jruby" if RUBY_PLATFORM != "java"
gem.platform = "java"
```
but I both are fine eally.

ph:
updated with the latest changes.

colinsurprenant:
LGTM

jsvd:
LGTM

PR #4867, opened 2016-03-22
Missing Command line flags from the documentation

ph:
With the addition of the API we have added 2 options to control the
behavior of the webserver, the port and IP address to bind. This PR add
them to the documentation.


ph:
When doing a code review I found out that theses options were not in the documentation.

purbon:
@ph test are being dancing funky here, what do you think the reason might be? the mass update?

purbon:
LGTM, is just a doc change.

purbon:
Actually the problem is due to us testing with JDK7 and having some only JDK8 code loaded, changed the PR test to be testing with JDK8 for now.

purbon:
please jenkins, test this.

suyograo:
LGTM

Issue #2852, opened 2015-03-18
Create automated release acceptance test


suyograo:
We need a Jenkins job that verifies the release artifacts and it needs to have more coverage than our existing Package Acceptance tests. This should have tests for plugin CLI, file inputs and test the existence of jars, vendor directories etc.

Test plan:
- [x] `bin/logstash version`: List current version
- [ ] `bin/logstash` with json input and check if parsing is successful
- [ ] `bin/plugin list`: Should list all plugins installed by default. Assert the count (currently 103)
- [x] `bin/plugin list --verbose`: Should list version info (Assert for stdin plugin)
- [x] `bin/plugin list stdin`: Should list logstash-input-stdin
- [x] `bin/plugin uninstall logstash-input-stdin`: Should uninstall, check for exit code
- [ ] `bin/plugin uninstall localgem`: Should uninstall, check for exit code.
- [x] `bin/plugin list stdin`: Should not list plugin
- [ ] `bin/plugin install logstash-input-wmi`
- [ ] `bin/plugin list --verbose wmi` : Should show current version (assert after getting from rubygems API)
- [x] `bin/plugin update` Mass update
- [x] Install locally built gem `bin/plugin install <file_path> --no_verify`
- [ ] `bin/plugin update` Try mass update after installing local gem
- [x] Test ES integration with geo, grok, and useragent filter with Apache logs. Send 50 Apache logs and check exact fields after doing a search on ES. Do this test on node, http and transport protocol
- [ ] Kafka input/output test. Bring up Kafka, send events through Logstash producer, and receive through Logstash consumer (@talevy has the scripts for these)
- [x] File input test: Follow test file (with known number of events) and assert if the events are received. Test `start_position => "beginning"` and also test tailing option
- [ ] File input test: Follow test file (with known number of events) and assert if the events are received. Test test tailing option
- [ ] Check for bundled jars. `find . -name *.jar | wc -l` should be 155.
- [ ] Add a `HTTP_PROXY` test for installing plugins

This should be run on all supported platforms

ph:
Linux is working fine with `test-kitchen` is just a matter of create the appropriate docker file.
There is some work done for windows, I will try to experiment with it.

purbon:
:dancer:

ph:
@talevy can you point me to the kafka scripts?

ph:
- [ ] use the `rakelib/default_plugins.rb` for the number of plugins installed assertion.
[...]

PR #4868, opened 2016-03-22
[POC] [Work in progress] Integration & Acceptance test

purbon:
Add POC sample for provisioning and running acceptance test in different OS for packaged and source code logstash distributions. As a way of experimentation this add a few ways to run this test:
    
    * by simply calling vagrant ssh in a rake task (see vm.rake)
    * by using the serverspec lib (has some restrictions but also some perks
      good to use)
    * by using a custom net:ssh wrapper to connect to the VM boxes.
    
The VM handling here is done using vagrant as a reliable way to create images, for the source of the imatges it uses the elastic public atlas of machines (https://atlas.hashicorp.com/elastic). There are scripts, for now provided for ubuntu and centos, also rake tasks to set up and destroy the machines.

The basic idea is to have a way to solve #2852, but also have bit more elaborated test stories for acceptance. Also giving the collateral benefit of having regular tests run in different OS, feature that is expected to also be provided by the internal CI system.

In this POC we plan to solve this scenario by:

* testing vs local generated VM's (using vagrant)
* testing vs any external system (machine connection is done throught SSH)
* testing locally (without any VM overhead)


purbon:
please jenkins, test this.

purbon:
closing this in benefit of a feature PR that will implement the first step of #5070

PR #4870, opened 2016-03-23
Bump to 2.3.0.snapshot3

suyograo:


colinsurprenant:
LGTM - @guyboertje do we need/want to also bump filewatch? was something merged yesterday?

Gemfile.jruby-1.9.lock
```diff
@@ -132,8 +132,8 @@ GEM
     logstash-codec-graphite (2.0.2)
       logstash-codec-line
       logstash-core (>= 2.0.0.beta2, < 3.0.0)
-    logstash-codec-json (2.1.0)
-      logstash-core (>= 2.0.0.beta2, < 3.0.0)
+    logstash-codec-json (2.1.1)
+      logstash-core (>= 2.0.0, < 3.0.0)
```
jsvd: strange, why was this dependency on logstash-core changed and the other weren't?

colinsurprenant: good catch. not sure wha't happening here. verify plugins gemspec? maybe a mass update consequence? @ph ?

ph: Yes, I think this is the case.  Not sure why the dependency changed.
Also this should be `logstash-codec-json 2.1.1` and no 2.1.0 ?

https://rubygems.org/gems/logstash-codec-json/versions/2.1.1

2.1.1 was release before my mass update, any reason why we arent picking up the latest version?

colinsurprenant:
It looks like we have some inconsistencies in the plugins gemspecs? what do you guys @ph @jsvd thinks we should do?

Gemfile.jruby-1.9.lock
ph: My previous comment wasn't accurate.

ph: It was updated in this commit in https://github.com/logstash-plugins/logstash-codec-json/commit/fd0b6efbb851ece48a37466eb10eff69485dec35 in 2.1.1

ph:
@colinsurprenant I have checked the dependencies and I think we are safe with how they are.
The constraints was changed to a more restrictive one, which is fine in that case since we just remove the beta.

colinsurprenant:
but I'd like to understand why we have an inconsistent constrain here?

Gemfile.jruby-1.9.lock
jsvd: as long as a single `logstash-core (>= 2.0.0, < 3.0.0)` doesn't change how bundler picks prereleased gems (I believe it shouldn't), we're ok

colinsurprenant: ok, so, we need to followup with an issue to make sure the gemspecs are consistent and that they should not be changed during a normal dev workflow, so this should have been caught in the PR review too. /cc @andrewvc since you made that commit

I'm good to move on if that new constrain does not  break things

colinsurprenant:
@ph thanks for the clarifications on the inconsistency origin, let's followup with an issue to make sure plugins gemspecs constrains are consistent and make sure we talk about it so that we don't mess with that unless there is a valid reason to do so, knowing about it will also help in catching that in review.

colinsurprenant:
are we making a new PR or updating for filewatch @suyograo  ?

suyograo:
@colinsurprenant updated this for fileinput: https://github.com/elastic/logstash/pull/4870/files#diff-efbf03f6386a8f9610003e5858706663R280

andrewvc:
LGTM

suyograo:
thanks @andrewvc

PR #4882, opened 2016-03-23
Create a virtual gem to handler the plugin dependency on core

ph:
We have decided to create a gem that the plugins can depend on to make
sure they use the correct core, with that in mind this will reduce the
mass update required when releasing a major release of logstash if the
contract between logstash and the plugin didnt change. The first version
of this gem act as a virtual gem, future release will include the actual
plugins code of the contract.

This gem should follow the same convention as semver, a breaking change
in the api will mean a bump in the major version.  Plugins author are
encouraged to test their plugins with a specific version of the contract
and define their dependency as the following:

```
  s.add_runtime_dependency "logstash-core-plugin-api", "~> 2.0"
```

This PR also introduce rake task to help managing this gem, since every
release of logstash will produce a new version of this gem with the
minor version bump. A **logstash-core-plugin-api** will declare a strict
dependency on a specific **logstash-core** version, like this:

```
  s.add_runtime_dependency "logstash-core", "2.3.0"
```

On the other hand each release of logtash will also declare a strict
dependency on a specific version of the `logstash-core-plugin-api` gem.

ref #4829

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,9 @@
 # If you modify this file manually all comments and formatting will be lost.
 
 source "https://rubygems.org"
-gem "logstash-core", "2.3.0.snapshot1"
-gem "logstash-core-event-java", "2.3.0.snapshot1"
+gem "logstash-core", "2.3.0.snapshot3"
+gem "logstash-core-event-java", "2.3.0.snapshot3"
+gem "logstash-core-plugin-api", "1.0.0"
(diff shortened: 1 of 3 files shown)


suyograo:
LGTM

suyograo:
Merged in https://github.com/elastic/logstash/commit/f42bc1bf95a9c5feeda2be5b984df700223978b8

PR #4885, opened 2016-03-24
Fix broken link

dedemorton:
It looks like this broken link was fixed in the logstash-docs repo, but not in logstash. Fixing it now so the problem doesn't persist. OK to merge?


suyograo:
LGTM

PR #4889, opened 2016-03-24
Add badges to plugin page

wiibaa:
As a user it is difficult to know which version of the plugin is bundled in a specific logstash release and to check if this is the latest one.
This PR is to discuss if this is a good feature to add and the correct way to add it to achieve something like a clear statement: 
_last release of logstash use v2.1.2 of date filter but a new version of the gem is available_
plugin-badges

My understanding is that the documentation package is build by a specific version of logstash, so reading the installed gemspec file for each plugin seems the easiest, thoughts ?

### DIFF
--- a/docs/asciidocgen.rb
+++ b/docs/asciidocgen.rb
@@ -170,6 +170,9 @@ class LogStashConfigAsciiDocGenerator
     # Now parse the real library
     code = File.new(file).read
 
+    # Read the plugin version from its gemspec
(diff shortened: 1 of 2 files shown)


suyograo:
@wiibaa This is an excellent idea! To be clear, this badge will be placed in the plugin documentation, not in the GitHub readme.

In addition to the badge, I would also add a line like you suggested:

```
Latest release of Logstash (v2.2.0) used the v2.1.2 of date filter but a new version of the gem (v2.1.3) is available
```

WDYT?

suyograo:
this has been fixed in the plugin documentation. Closing

PR #4904, opened 2016-03-28
Revert to use only a single worker by default on outputs

andrewvc:
The move to auto-scale output workers was great in theory, but a lot of outputs
just weren't built to support it well, they often used too many resources or had
logical errors.

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -37,6 +37,8 @@ module LogStash; class Pipeline
     :config_str,
     :original_settings
 
+  DEFAULT_OUTPUT_WORKERS = 1
+
   DEFAULT_SETTINGS = {
     :default_pipeline_workers => LogStash::Config::CpuCoreStrategy.maximum,
     :pipeline_batch_size => 125,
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/pipeline.rb
```diff
@@ -469,7 +469,7 @@ def plugin(plugin_type, name, *args)
   end
 
   def default_output_workers
-    @settings[:pipeline_workers] || @settings[:default_pipeline_workers]
+    1
```
suyograo: Can we make this a constant?

suyograo:
LGTM

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -37,6 +37,8 @@ module LogStash; class Pipeline
     :config_str,
     :original_settings
 
+  DEFAULT_OUTPUT_WORKERS = 1
+
   DEFAULT_SETTINGS = {
```
jsvd: any reason not to place DEFAULT_OUTPUT_WORKERS in the DEFAULT_SETTINGS hash?

untergeek:
LGTM™ too.

logstash-core/lib/logstash/pipeline.rb
andrewvc: I'd rather not have defaults in there in general. I think we should probably pull `default_pipeline_workers` out of there IMHO.

andrewvc: Just to clarify, I think its confusing that we have dynamic settings and defaults in that hash. Defaults make more sense as constants IMHO.

PR #4905, opened 2016-03-28
Revert to use only a single worker by default on outputs

andrewvc:
Port of https://github.com/elastic/logstash/pull/4904 to 2.x series Logstash


suyograo:
LGTM

Issue #4909, opened 2016-03-29
Test on master failing with NPE on HotThreadsCommand#schema


jsvd:
https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-os-compatibility/os=ubuntu/lastBuild/console

Error is:

```
2) JVM stats LogStash::Api::HotThreadsCommand#schema return hot threads information
     Failure/Error: Unable to find matching line from backtrace
     Java::JavaLang::NullPointerException:
     # com.purbon.jrmonitor.monitors.HotThreadsMonitor$ThreadReport.<init>(HotThreadsMonitor.java:36)
     # com.purbon.jrmonitor.monitors.HotThreadsMonitor.detect(HotThreadsMonitor.java:146)
     # com.purbon.jrmonitor.reports.RubyThreadsReport.generate(RubyThreadsReport.java:41)
     # com.purbon.jrmonitor.reports.RubyThreadsReport$INVOKER$s$generate.call(RubyThreadsReport$INVOKER$s$generate.gen)
```

purbon:
working on this, //cc @ph

purbon:
should be fixed by the release of https://rubygems.org/gems/jruby-monitoring/versions/0.3.1

jsvd:
nice. thanks! can you link to the issue in the jruby-monitoring project?

jsvd:
also, https://github.com/elastic/logstash/blob/master/logstash-core/logstash-core.gemspec#L33 should probably bumped to be at least 0.3.1, no?

purbon:
is going to fetch it anyway, but yes we can bump the versions.

ph:
lets bump it, quick turn around @purbon and good find @jsvd

purbon:
done https://github.com/elastic/logstash/pull/4911

PR #4911, opened 2016-03-29
bump jruby-monitoring version to >= 0.3.1 to include last fixes

purbon:
Forces to get, _min_ the jruby-monitoring with the fix for #4909

### DIFF
--- a/logstash-core/logstash-core.gemspec
+++ b/logstash-core/logstash-core.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency 'puma', '~> 2.16', '>= 2.16.0'
   gem.add_runtime_dependency "jruby-openssl", "0.9.13" # Required to support TLSv1.2
   gem.add_runtime_dependency "chronic_duration", "0.10.6"
-  gem.add_runtime_dependency "jruby-monitoring", '~> 0.1'
+  gem.add_runtime_dependency "jruby-monitoring", '~> 0.3.1'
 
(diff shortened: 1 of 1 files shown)


ph:
LGTM

suyograo:
LGTM

logstash-core/logstash-core.gemspec
```diff
@@ -30,7 +30,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency 'puma', '~> 2.16', '>= 2.16.0'
   gem.add_runtime_dependency "jruby-openssl", "0.9.13" # Required to support TLSv1.2
   gem.add_runtime_dependency "chronic_duration", "0.10.6"
-  gem.add_runtime_dependency "jruby-monitoring", '~> 0.1'
+  gem.add_runtime_dependency "jruby-monitoring", '~> 0.3.1'
```
jsvd: I'm ok with this change, but I must warn that it changes the current behaviour:

```
jruby-1.7.23 :001 > Gem::Requirement.new("~> 0.3.1").satisfied_by?(Gem::Version.new("0.5"))
 => false 
jruby-1.7.23 :002 > Gem::Requirement.new("~> 0.1").satisfied_by?(Gem::Version.new("0.5"))
 => true 
```

An alternative would be to do:

```
gem.add_runtime_dependency "jruby-monitoring", '~> 0.1', '>= 0.3.1'
```

purbon: I know it changes requirements, but is also good to be more strict to avoid confusions and errors in the feature.

PR #4913, opened 2016-03-29
Retrieve JAVA_HOME from windows symlink

StefanScherer:
I struggled today as the latest Chocolatey package to install Java on Windows does no longer set the JAVA_HOME environment automatically.

So I tried another approach to get rid of the JAVA_HOME environment variable ot start logstash on Windows.

I have seen that the Java MSI installs some symlinks on Windows with the path to the real java.exe which includes the path to the Java version installed.

```
Microsoft Windows [Version 10.0.10586]
(c) 2015 Microsoft Corporation. All rights reserved.

C:\Users\vagrant>where java
C:\ProgramData\Oracle\Java\javapath\java.exe

C:\Users\vagrant>dir %ProgramData%\Oracle\java\javapath
 Volume in drive C is Windows 10
 Volume Serial Number is 9C9F-63C7

 Directory of C:\ProgramData\Oracle\java\javapath

03/29/2016  11:32 AM    <DIR>          .
03/29/2016  11:32 AM    <DIR>          ..
03/29/2016  11:32 AM    <SYMLINK>      java.exe [C:\Program Files\Java\jre1.8.0_77\bin\java.exe]
03/29/2016  11:32 AM    <SYMLINK>      javaw.exe [C:\Program Files\Java\jre1.8.0_77\bin\javaw.exe]
03/29/2016  11:32 AM    <SYMLINK>      javaws.exe [C:\Program Files\Java\jre1.8.0_77\bin\javaws.exe]
               3 File(s)              0 bytes
               2 Dir(s)  49,970,565,120 bytes free

C:\Users\vagrant>dir %ProgramData%\Oracle\java\javapath\java.exe
 Volume in drive C is Windows 10
 Volume Serial Number is 9C9F-63C7

 Directory of C:\ProgramData\Oracle\java\javapath

03/29/2016  11:32 AM    <SYMLINK>      java.exe [C:\Program Files\Java\jre1.8.0_77\bin\java.exe]
               1 File(s)              0 bytes
               0 Dir(s)  48,065,687,552 bytes free
```

With the changes in setup.bat the batch script is able to generate JAVA_HOME from that symlink. Tested on Windows 7 and Windows 10 and Java 1.8.0_74 and 77.

### DIFF
--- a/bin/setup.bat
+++ b/bin/setup.bat
@@ -15,6 +15,10 @@ goto finally
 
 :setup_jruby
 REM setup_java()
+if not defined JAVA_HOME IF EXIST %ProgramData%\Oracle\java\javapath\java.exe (
+    for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_HOME=%%a
+)
+if defined JAVA_HOME set JAVA_HOME=%JAVA_HOME:\bin\java.exe=%
 if not defined JAVA_HOME goto missing_java_home
 REM ***** JAVA options *****
 
--- a/bin/setup.bat
+++ b/bin/setup.bat
@@ -15,10 +15,15 @@ goto finally
 
 :setup_jruby
 REM setup_java()
+set JAVA_FROM_SYMLINK=
 if not defined JAVA_HOME IF EXIST %ProgramData%\Oracle\java\javapath\java.exe (
     for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_HOME=%%a
+    set JAVA_FROM_SYMLINK=1
 )
 if defined JAVA_HOME set JAVA_HOME=%JAVA_HOME:\bin\java.exe=%
+if "%JAVA_FROM_SYMLINK%" == "1" (
+    echo Using JAVA_HOME=%JAVA_HOME% retrieved from %ProgramData%\Oracle\java\javapath\java.exe
+)
 if not defined JAVA_HOME goto missing_java_home
 REM ***** JAVA options *****
 
--- a/bin/setup.bat
+++ b/bin/setup.bat
@@ -15,13 +15,11 @@ goto finally
 
 :setup_jruby
 REM setup_java()
-set JAVA_FROM_SYMLINK=
 if not defined JAVA_HOME IF EXIST %ProgramData%\Oracle\java\javapath\java.exe (
-    for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_HOME=%%a
-    set JAVA_FROM_SYMLINK=1
+    for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_EXE=%%a
 )
-if defined JAVA_HOME set JAVA_HOME=%JAVA_HOME:\bin\java.exe=%
-if "%JAVA_FROM_SYMLINK%" == "1" (
+if defined JAVA_EXE set JAVA_HOME=%JAVA_EXE:\bin\java.exe=%
+if defined JAVA_EXE (
     echo Using JAVA_HOME=%JAVA_HOME% retrieved from %ProgramData%\Oracle\java\javapath\java.exe
 )
 if not defined JAVA_HOME goto missing_java_home


StefanScherer:
Your CLA is hard to sign as it forces me to turn off default security features in Safari.

StefanScherer:
OK, it works in Chrome. I've signed the CLA.

bin/setup.bat
```diff
@@ -15,6 +15,10 @@ goto finally
 
 :setup_jruby
 REM setup_java()
+if not defined JAVA_HOME IF EXIST %ProgramData%\Oracle\java\javapath\java.exe (
+    for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_HOME=%%a
```
suyograo: @StefanScherer Can we log/echo that we are deducing the `JAVA_HOME` from this location?

bin/setup.bat
```diff
@@ -15,10 +15,15 @@ goto finally
 
 :setup_jruby
 REM setup_java()
+set JAVA_FROM_SYMLINK=
 if not defined JAVA_HOME IF EXIST %ProgramData%\Oracle\java\javapath\java.exe (
     for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_HOME=%%a
+    set JAVA_FROM_SYMLINK=1
 )
 if defined JAVA_HOME set JAVA_HOME=%JAVA_HOME:\bin\java.exe=%
+if "%JAVA_FROM_SYMLINK%" == "1" (
```
StefanScherer: @suyograo I've added the echo to show the JAVA_HOME retrieved from the symlink.

I would have added that into the first if block, but unfortunately the cmd shell does not give me a value for `%JAVA_HOME%` inside that if block. So I've added another helper variable to do the echo afterwards.

So the output looks like this if Java is installed without the JAVA_HOME system environment set.

```
PS C:\vagrant\resources\logstash\bin> .\setup.bat
Using JAVA_HOME=C:\Program Files\Java\jre1.8.0_77 retrieved from C:\ProgramData\Oracle\java\javapath\java.exe
```

And if no Java is installed or this symlink does not exist it still shows this:

```
PS C:\vagrant\resources\logstash\bin> .\setup.bat
JAVA_HOME environment variable must be set!
```

bin/setup.bat
```diff
@@ -15,6 +15,15 @@ goto finally
 
 :setup_jruby
 REM setup_java()
+set JAVA_FROM_SYMLINK=
+if not defined JAVA_HOME IF EXIST %ProgramData%\Oracle\java\javapath\java.exe (
+    for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_HOME=%%a
+    set JAVA_FROM_SYMLINK=1
+)
+if defined JAVA_HOME set JAVA_HOME=%JAVA_HOME:\bin\java.exe=%
```
suyograo: Why is this line needed? `JAVA_HOME` should not include the `bin` directory..

StefanScherer: This line cuts off the `\bin\java.exe` from the value retrieved from the symlink.
I've added another commit to make it more clear. There I use a variable `JAVA_EXE` and then build `JAVA_HOME` from it.

PTAL

bin/setup.bat
```diff
@@ -15,6 +15,13 @@ goto finally
 
 :setup_jruby
 REM setup_java()
+if not defined JAVA_HOME IF EXIST %ProgramData%\Oracle\java\javapath\java.exe (
+    for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_EXE=%%a
```
suyograo: Sorry if i wasn't clear. I think this line should be 

```
for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_HOME=%%a
```

See https://docs.oracle.com/cd/E19182-01/820-7851/inst_cli_jdk_javahome_t/

> Set JAVA_HOME:
> 
> Right click My Computer and select Properties.
> 
> On the Advanced tab, select Environment Variables, and then edit JAVA_HOME to point to where the JDK software is located, for example, C:\Program Files\Java\jdk1.6.0_02.

bin/setup.bat
```diff
@@ -15,6 +15,13 @@ goto finally
...
+)
+if defined JAVA_EXE set JAVA_HOME=%JAVA_EXE:\bin\java.exe=%
```
StefanScherer: @suyograo No problem. I thought about my solution and find that a temporary variable JAVA_EXE fits better here. So I made the last commit. This for loop sets eg. JAVA_EXE = C:\Program Files\Java\jdk1.6.0_02\bin\java.exe as it only can read the complete symlink to the java.exe binary file.

Here in line 21 the JAVA_HOME variable is then set with the value of JAVA_EXE, but without the \bin\java.exe, so we set here JAVA_HOME = C:\Program Files\Java\jdk1.6.0_02

The code works as it is needed for a correct JAVA_HOME env in the script, I've tested this script on a machine with Java 8 installed, but no user/system JAVA_HOME environment set.

suyograo:
@StefanScherer thanks for the explanation and contributing to Logstash.

LGTM

ph:
I've tested this behavior under windows 2012 after unsetting my JAVA_HOME logstash was able to find it.

StefanScherer:
I'll check if elasticsearch also needs JAVA_HOME in the bat start scripts. We found that "problem" as we started to implement MSI installers for ELK and wanted to get rid of the manual step of setting JAVA_HOME. The user has to install Java and then the three MSI's should do the rest.

PR #4941, opened 2016-03-31
Add Kafka beta version to 5.0.0-alpha1 package

suyograo:
Beta does not get picked up automatically by bundler as the latest release. You have to explicitly ping the version. We can remove this once Kafka plugins are not beta.


ph:
I was curious with that changes and I have stumbled accross an issue, if you checkout this PR and setup your environment with `rake bootstrap` and you try to run the `bin/logstash-plugin update` you will get this error:

```
~/e/logstash git:pr/4941 ❯❯❯ bin/logstash-plugin update                                                                                                                                                   ⏎ ✱
ERROR: Updated Aborted, message: undefined method `version' for nil:NilClass
~/e/logstash git:pr/4941 ❯❯❯
```

This is strange I have tried the following in the 5.0 branch which should do the same result. Add this line to the Gemfile:
```
gem "logstash-input-beats", "2.2.3"
```

And if you do this:
```
bin/plugin install --no-verify
```
Theses commands successfully update the current packages.

So I think pre-release version of gems breaks the update flow.

ph:
I have marked the other issue for alpha2, I don't think the problem is block for the alpha1.

LGTM

PR #4944, opened 2016-03-31
fix --config-test and add tests for it

### LINKED ISSUE #4933: --configtest not working on 2.3 (undefined method `config_valid?')
I'm getting the following error on 2 instances of logstash that were upgraded from 2.0 to 2.3, both are running ubuntu 12.04 amd64.

The error occurs when running `logstash -f /etc/logstash/conf.d --configtest`.
Since the new init script uses configtest when doing a restart, it means `/etc/init.d/logstash restart` cannot be used. 

```
An unexpected error occurred! {:error=>#<NoMethodError: undefined method `config_valid?' for LogStash::Pipeline:Class>, :class=>"NoMethodError", :backtrace=>["/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.0-java/lib/logstash/agent.rb:169:in `execute'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.0-java/lib/logstash/runner.rb:94:in `run'", "org/jruby/RubyProc.java:281:in `call'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.0-java/lib/logstash/runner.rb:99:in `run'", "org/jruby/RubyProc.java:281:in `call'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/stud-0.0.22/lib/stud/task.rb:24:in `initialize'"], :level=>:warn}
```

jsvd:
fixes #4933

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -34,16 +34,6 @@ module LogStash; class Pipeline
     "LogStash::Inputs::Stdin"
   ]
 
-  def self.validate_config(config_str, settings = {})
-    begin
-      # There should be a better way to test this
-      self.new(config_str, settings)
-      true
-    rescue => e
-      e.message
-    end
-  end
-
   def initialize(config_str, settings = {})
     @config_str = config_str
     @original_settings = settings
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -38,6 +38,7 @@ module LogStash; class Pipeline
     begin
       # There should be a better way to test this
       self.new(config_str, settings)
+      true
     rescue => e
       e.message
     end
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -10,7 +10,7 @@ describe LogStash::Agent do
   subject { LogStash::Agent.new("", "") }
 
   before :each do
-    [:log, :info, :warn, :error, :fatal, :debug].each do |level|
+    [:log, :info, :warn, :error, :fatal, :debug, :terminal].each do |level|
       allow(logger).to receive(level)
     end
     [:info?, :warn?, :error?, :fatal?, :debug?].each do |level|
@@ -293,6 +293,24 @@ describe LogStash::Agent do
       expect(File).to receive(:directory?).and_return(false)
       expect(LogStash::Environment).not_to receive(:add_plugin_path)
       expect{subject.configure_plugin_paths(single_path)}.to raise_error(LogStash::ConfigurationError)
+    end
+  end
+
+  context "--config-test" do
+    let(:cli_args) { ["-t", "-e", pipeline_string] }
+
+    context "with a good configuration" do
+      let(:pipeline_string) { "input { } filter { } output { }" }
+      it "should exit successfuly" do
+        expect(subject.run(cli_args)).to eq(0)
+      end
+    end
+
+    context "with a bad configuration" do
+      let(:pipeline_string) { "rlwekjhrewlqrkjh" }
+      it "should fail by returning a bad exit code" do
+        expect(subject.run(cli_args)).to eq(1)
+      end
     end
   end
 
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -296,7 +296,7 @@ describe LogStash::Agent do
     end
   end
 
-  context "--config-test" do
+  describe "--config-test" do
     let(:cli_args) { ["-t", "-e", pipeline_string] }
 
     context "with a good configuration" do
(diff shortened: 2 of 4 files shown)


purbon:
reviewing this.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -38,6 +38,7 @@ def self.validate_config(config_str, settings = {})
     begin
       # There should be a better way to test this
       self.new(config_str, settings)
+      true
     rescue => e
       e.message
```
purbon: should we make this return false in case something is raised?

purbon: should we make this return false in case something is raised?
 
would it also be better to use a global rescue for this method?  smething like

```
def some_method(args)
  self.new(....)
rescue
 e.message
 false 
end
```

but i'm sure this has broader implications as expectations of methods using this validate function might be expecting this message somehow, 

what do you think?

purbon:
related to #4946

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -296,6 +296,24 @@
     end
   end
 
+  context "--config-test" do
```
purbon: I would change this to `describe` as is more of a general are with two specific context for ```--config-test```

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -296,6 +296,24 @@
...
+    let(:cli_args) { ["-t", "-e", pipeline_string] }
+
+    context "with a good configuration" do
+      let(:pipeline_string) { "input { } filter { } output { }" }
```
purbon: I see this config string being used in many places, I think I count at less 4 in this same spec. Can we dry it out a bit?

logstash-core/lib/logstash/pipeline.rb
purbon: I think I would go for a more uniform interface for this method, also might be good to add some documentation so we don't make mistakes with return expectations.

jsvd: yes the argument for not using false there is the need for that message. I guess this can be refactored in another way, potentially using exceptions, but I guess for now it ok.

logstash-core/spec/logstash/agent_spec.rb
jsvd: :+1:

logstash-core/spec/logstash/agent_spec.rb
jsvd: I think that, at least in this case, drying it makes it much harder to read, as the test is:

```
"testing ok config"
config = "ok config"
assertion with ok config 
```

drying would force the reader to look for ok config string somewhere in the code

purbon: I'm on the fence here, it just feels awkward, and for me is an indication that probably this test should be located within other "broader" sections. 

For this change, it works and does the work, so I'm good with it (in a pragmatic way :+1: )

logstash-core/lib/logstash/pipeline.rb
jsvd: I'll document the method signature, is that enough? 
I believe the alternative implementation would be with just using an exception. Ruby doesn't have a good convention for return either true or an error message, as some other languages do (obviously erlang comes to mind :D )

purbon: well yes an no, at minimum I would introduce some clear documentation ... the return value interface here is just misleading and confusing as it return two kind of values. We need to provide clarity in order to avoid feature issues.

And actually there is only one user of this interface and is the one you just fix.

And yes, your fix fixed the issue for this change, but we could also help out on fixing the interface as is mostly part of why this happened.

guyboertje: I think that returning a `string` or `false` is keeping the contract of returning something `truthy` because a string is `truthy`

purbon:
LGTM

guyboertje:
Consider moving the config loader to the `validate_config` method in pipeline.rb
```ruby
def self.validate_config(path, config, logger, i18n_key = "logstash.error")
  config_str = LogStash::Config::Loader.new(logger).format_config(path, config)
  begin
    # There should be a better way to test this
    self.new(config_str, settings)
    logger.terminal "Configuration OK"
    return 0
  rescue => e
    logger.fatal I18n.t(i18n_key, :error => e.message)
    return 1
  end
end
```
so the call in agent is simply...
```ruby

if config_test?
  return LogStash::Pipeline.validate_config(config_path, config_string,
        @logger, "logstash.agent.invalid-configuration")
end
```

jsvd:
I'm a bit on the fence on moving the code back to pipeline.rb
the main argument is that the validate_config just requires tons of arguments (4 and should be 5 with the settings), which I'm pretty sure violates some envy pattern and angries 2 out of 4 GoF members.

That said I think there might be value in moving the code from your validate_config to a method in agent.rb itself. Honestly I don't think the current proposal is so obviously wrong it justifies further refactoring, but can be convinced otherwise.

guyboertje:
@jsvd - I am ambivalent too.

LGTM

PR #4946, opened 2016-03-31
fix --config-test and add a couple of tests

### LINKED ISSUE #4933: --configtest not working on 2.3 (undefined method `config_valid?')

jsvd:
fixes #4933

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -52,16 +52,6 @@ module LogStash; class Pipeline
     "LogStash::Inputs::Stdin"
   ]
 
-  def self.validate_config(config_str, settings = {})
-    begin
-      # There should be a better way to test this
-      self.new(config_str, settings)
-      true
-    rescue => e
-      e.message
-    end
-  end
-
   def initialize(config_str, settings = {})
     @config_str = config_str
     @original_settings = settings
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -56,6 +56,7 @@ module LogStash; class Pipeline
     begin
       # There should be a better way to test this
       self.new(config_str, settings)
+      true
     rescue => e
       e.message
     end
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -166,9 +166,9 @@ class LogStash::Runner < Clamp::Command
     end
 
     if config_test?
-      config_loader = LogStash::Config::Loader.new(@logger, config_test?)
+      config_loader = LogStash::Config::Loader.new(@logger)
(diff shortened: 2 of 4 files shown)


logstash-core/lib/logstash/pipeline.rb
```diff
@@ -56,6 +56,7 @@ def self.validate_config(config_str, settings = {})
     begin
       # There should be a better way to test this
       self.new(config_str, settings)
+      true
     rescue => e
       e.message
```
purbon: please see: https://github.com/elastic/logstash/pull/4944#discussion_r58017201 on related change comment.

purbon:
This is basically a port of https://github.com/elastic/logstash/pull/4944 to 5.0, in order not to repeat comments here, I would focus on getting first #4944 clear and then we can focus on special things here.

logstash-core/lib/logstash/pipeline.rb
guyboertje: Consider using nil instead of true so that the test on line 172 becomes so much more readable, nil being the absence of an error.  This is similar to the way Ruby returns nil on compact! in the absence of a change.
```ruby
if config_error
  @logger.fatal I18n.t("logstash.error", :error => config_error)
  return 1
else
  @logger.terminal "Configuration OK"
  return 0
end
```

purbon: this is for sure a more clean solution, I try to keep comments on #4944 as basically is the same code ported here.

jsvd:
another round of review welcome. essentially corrected the same things as in #4944

guyboertje:
LGTM

purbon:
LGTM

Issue #4902, opened 2016-03-28
Publish logstash-core-event-java artifact to Maven Central


suyograo:
Publish this jar to maven so developers can directly use this Event object in their pure Java plugins.

See https://github.com/elastic/logstash/issues/4515#issuecomment-200277819

driverpt:
@suyograo , you have any ETA on this ?

driverpt:
Created PR #4947 to help with this issue.

talevy:
Elasticsearch publishes their artifacts to Sonatype. I think this should also be where we publish our java artifacts

talevy:
I also think it is appropriate to hold off on publishing the artifact to a central repository until we shift more components to Java and solidify the API a bit more

driverpt:
@talevy , i recommend at least bintray/jcentral

colinsurprenant:
@driverpt it is too early to officially publish the Java Event to Maven IMO. This is definitely the plan at some point but the work done so far has been mainly to support the JRuby compatibility layer and we haven't given much thoughts on the pure Java API side for the context of writing pure Java plugins **yet**. 

Ultimately this is the goal and we are getting there but it is a bit premature to commit it as a published jar at this point. Does that make sense?

Otherwise, I am all in favour of enhancements that could help **you** build & generate jars for usage in your project until we publish the jars as proposed in #4947

w4:
bump, this is relevant again :)

cc @danhermann

PR #4947, opened 2016-03-31
Add Maven support

driverpt:
This allows to perform `./gradlew install` and include it in Pure-Java Libraries.

### DIFF
--- a/logstash-core-event-java/build.gradle
+++ b/logstash-core-event-java/build.gradle
@@ -6,6 +6,7 @@ buildscript {
     }
     dependencies {
         classpath 'net.saliman:gradle-cobertura-plugin:2.2.8'
+        classpath 'com.bmuschko:gradle-nexus-plugin:2.3.1'
     }
 }
 
@@ -25,6 +26,7 @@ gradle.projectsEvaluated {
 apply plugin: 'java'
 apply plugin: 'idea'
 apply plugin: 'maven'
+apply plugin: 'com.bmuschko.nexus'
 
 group = 'org.logstash'
 version = '5.0.0.dev'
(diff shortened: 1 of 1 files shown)


talevy:
@driverpt Thanks for the contribution!

I think it would be more appropriate to publish our artifacts to Sonatype.

we can take a look at using this plugin: https://github.com/bmuschko/gradle-nexus-plugin just as ES does for publishing

driverpt:
@talevy , are the artifacts automatically published to Maven Central after Sonatype ?

talevy:
yes, jars will be published to Maven Central via Sonatype

driverpt:
@talevy , i just updated it.

colinsurprenant:
as discussed in #4902 we will not publish to Maven/Sonatype just yet. I am ok if @tal and/or myself review this for the eventual publishing but in the meantime is there anything else we could do to help you work with the current version of the Java Event? Can't you just build the jar locally and use it in your project?

driverpt:
@colinsurprenant , since logstash-core-event-java is already used in v2.3.0, it makes perfect sense to publish it to Maven Central, the only thing that the Developers need to know is to call the `event.to_java` method when passing it as a parameter.

colinsurprenant:
@driverpt it was reverted in 2.3.1 because we realized it introduced a non-backward compatible behaviour that could show up in Ruby filter scripts and/or customs plugins that we don't control.

But this has nothing to do with the publishing or not to Maven. As I explained in #4902 «it is too early to officially publish the Java Event to Maven IMO. This is definitely the plan at some point but the work done so far has been mainly to support the JRuby compatibility layer and we haven't given much thoughts on the pure Java API side for the context of writing pure Java plugins yet.Ultimately this is the goal and we are getting there but it is a bit premature to commit it as a published jar at this point.»

Does that make sense?

driverpt:
@colinsurprenant , you're right, it was reverted, but nevertheless you launched a Version with that Artifact, so, to me, because of that, that's more than enough reason to publish an Artifact to Maven Central.

If the feature is unstable, it probably shouldn't have been launched in the first place, but if it's launched, there should be a Maven Central Artifact

colinsurprenant:
@driverpt «unstable feature» is too broad of a statement here. Obviously we are doing our best to make stable code, but at this moment we do not have any requirement to commit on a stable interface before we are ready to do so.

Publishing something as a standalone library is a statement in relation to its interface and not all written classes necessarily qualifies as a being mature enough **in that respect**. As I tried to explain, we work iteratively and the focus has been on making it work in relation to the Ruby interface. Again, for now it is still an **internal component** of the logstash-core-event-java **gem**. There is no standalone published artifact of the Java Event class meant for integration with Java code **just yet**. We are totally ok if you discovered it and started using it but you have to understand that you are a bit on your own and will have to track changes and we will not pay attention to any breaking changes at this point, until we feel its good enough to exist as a publicly usable component. For now consider it as an internal component. 

Again, it will happen soon, but not today.

suyograo:
@driverpt will close this since we are not moving forward with this at this time. We can reopen when we are ready to publish to maven. Until then, please use the builds to generate the jar artifacts for `Event` class.

PR #4952, opened 2016-03-31
ensure Plugin#original_params is hides password fields

jsvd:


logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -142,6 +142,11 @@ def config_init(params)
       instance_variable_set("@#{key}", value)
     end
 
+    # now that we know the parameters are valid, we can obfuscate the original copy
+    # of the parameters before storing them as an instance variable
+    self.class.validate_check_parameter_values(original_params)
```
purbon: Isn't this actually doing the validation twice? I saw https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/config/mixin.rb#L252-L265 that is basically the validation method.

I know this fixes the problem, I'm just wondering here if we should have a diff method or some kind of alias that show we're actually running the change to hide the password.

what do you think?

logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -142,6 +142,11 @@ def config_init(params)
       instance_variable_set("@#{key}", value)
     end
 
+    # now that we know the parameters are valid, we can obfuscate the original copy
```
purbon: what about changing `obfuscate` to `hide` when talking about security, obfuscate has this idea of  security through obscurity, what do you think?

logstash-core/lib/logstash/config/mixin.rb
jsvd: I struggled a bit with that, and I did a sample implementation of a method just to do that, but it's somewhat awkward and leads to some duplication of code.
I selected `validate_check_parameter_values` because it only takes care of validating arguments against their validators, so it doesn't populate default values, etc

logstash-core/lib/logstash/config/mixin.rb
jsvd: I don't have feelings either way. imo, it's correct in the sense of english (it can mean blur, complicate, etc). as for security by obscurity..this is opensource :D

logstash-core/lib/logstash/config/mixin.rb
ph: I am doing some testing for that, not sure if it could have some side effect in specific cases.

ph: I am OK with this fix event if we are actually validating twice the submitted values.
The actual problem is we should split the value assignment from the validation.

ph:
LGTM

logstash-core/lib/logstash/config/mixin.rb
purbon: I would say, we should add an issue to cleanup this deb, are you ok with
that?

jsvd:
refactored this so password hiding is done in a method created for this purpose alone. review welcome

ph:
I prefer that refactor,  thanks @jsvd, LGTM even more.

logstash-core/lib/logstash/config/mixin.rb
andrewvc: I agree with @purbon , an obfuscation implies that something can be reversed (an obfuscated password is something like HTTP basic, where the pass is base64 encoded). This is hidden or masked. I think this is an important distinction to make here if someone is reading this for a security review.

purbon:
LGTM, but I would so as @andrewvc also recommended s/obfuscation/hidden (or similar)

andrewvc:
LGTM pending a change to that comment regarding obfuscation. I tested this manually and saw the password properly masked in the logs.

PR #4954, opened 2016-03-31
Add a systemd unit for logstash

mmuehlenhoff:
This is feature equivalent to the sysvinit script for Debian/Ubuntu.

We're using this systemd unit in the Wikimedia Foundation's Logstash cluster (on Debian jessie).


purbon:
Thanks for your contribution @mmuehlenhoff, you will need to  sign the CLA at https://www.elastic.co/contributor-agreement/ so we can continue with your change, is this ok for you?

mmuehlenhoff:
I've forwarded this to our Legal Department. Signing the CLA should be no problem, but might take up to a few days.

mmuehlenhoff:
Our Legal department has signed the CCLA and I've just filled the CLA form. This probably needs some processing on Elastic's side, so let me know if you need anything else from me.

mmuehlenhoff:
The CLA has now been signed

jordansissel:
Thanks for your efforts! <3

@untergeek is working on having our packages ship with all the lovely service scripts (systemd, upstart, sysv, etc) as a separate task -- his work will show up in Logstash 5.0. I think we can close this in favor of that work.

suyograo:
@mmuehlenhoff Thanks for your contribution. In the last couple of weeks, we've been working on adding systemd support to Logstash and @untergeek is crafting a solution that involves shipping pleaserun with LS. This way we can create startup scripts based on where LS gets installed.

Will be closing this PR

karmi:
Hi @mmuehlenhoff, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/logstash/pull/4954.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @mmuehlenhoff, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/logstash/pull/4954.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #4956, opened 2016-03-31
Make the `logstash-plugin` works when you have an installed pre-release gem.

### LINKED ISSUE #4955: Using pre-release gems in a Gemfile break update flow.
Discovered when testing PR #4941.

When you are using a pre-release gems in the Gemfile it breaks the plugin update all flow.
```ruby
#Gemfile
gem "logstash-input-kafka", "3.0.0.beta4"
```

```
~/e/logstash git:pr/4941 ❯❯❯ bin/logstash-plugin update                                                                                                                                                   ⏎ ✱
ERROR: Updated Aborted, message: undefined method `version' for nil:NilClass
~/e/logstash git:pr/4941 ❯❯❯
```

This is strange I have tried the following in the 5.0 branch which should do the same result. Add this line to the Gemfile:
```
gem "logstash-input-beats", "2.2.3"
```

And if you do this:
```
bin/plugin install --no-verify
```
Theses commands successfully update the current packages, so the problem I think is only related to the pre-release version of the gem.

ph:
Fixes #4955

### DIFF
--- a/lib/pluginmanager/util.rb
+++ b/lib/pluginmanager/util.rb
@@ -4,7 +4,7 @@ require "rubygems/package"
 module LogStash::PluginManager
 
   class ValidationError < StandardError; end
-  DEFAULT_GEM_REQUIREMENTS = Gem::Requirement.new([">= 0.b"])
+  DEFAULT_GEM_REQUIREMENTS = Gem::Requirement.new([">= 0.a"])
 
   # check for valid logstash plugin gem name & version or .gem file, logs errors to $stdout
   # uses Rubygems API and will remotely validated agains the current Gem.sources
--- a/lib/pluginmanager/util.rb
+++ b/lib/pluginmanager/util.rb
@@ -4,6 +4,7 @@ require "rubygems/package"
 module LogStash::PluginManager
 
   class ValidationError < StandardError; end
+  DEFAULT_GEM_REQUIREMENTS = Gem::Requirement.new([">= 0.b"])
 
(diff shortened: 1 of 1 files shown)


ph:
@talevy right it should an .a instead of b

lib/pluginmanager/util.rb
```diff
@@ -4,6 +4,7 @@
 module LogStash::PluginManager
 
   class ValidationError < StandardError; end
+  DEFAULT_GEM_REQUIREMENTS = Gem::Requirement.new([">= 0.b"])
```
ph: should 0.a not 0.b as mention by @talevy

purbon:
I don't see test on this, but are you sure with this change.

* You are not updating to `--pre` versions with the current interface.
* Install `--pre`  versions is explicit and not enabled by default?

What do you think?

Adding specific test for it is required and encouraged :+1:

talevy:
what is the state of this?

talevy:
pinging again

ph:
@talevy I am adding an acceptance test for it.

ph:
@purbon 

```
You are not updating to --pre versions with the current interface.
Install --pre versions is explicit and not enabled by default?
```
Installing `--pre` version will be explicit after this change since `prereleased` will be hidded from the api. 

So if people want to install a pre release they will have to explicitly use a version when calling `bin/logstash-plugin install`

ph:
With the upcoming changes from #6168 this wont be necessary.

PR #4960, opened 2016-03-31
Fix links causing build errors and add missing files

dedemorton:
The link issues were fixed in the logstash-docs repo, but not the logstash repo.

Also adding content changes that were made in the logstash-docs repo, but not the logstash repo.


dedemorton:
Note that this PR is simply to get the changes into the logstash repo where they belong.

ph:
LGTM

Issue #4958, opened 2016-03-31
Disable env variable interpolation by default


andrewvc:
Env variable interpolation breaks current configs (as seen in https://github.com/elastic/logstash/issues/4953 ). We need to make this disabled by default until we have a better implementation with real quoting and escaping.

suyograo:
Fixed in https://github.com/elastic/logstash/pull/4963

PR #4963, opened 2016-03-31
add experimental flag for environment variables in config

talevy:
closes: https://github.com/elastic/logstash/issues/4958
first introduced by: https://github.com/elastic/logstash/pull/4710

the proposed solution is to re-introduce environment variable interpolation within configs to be hidden behind an experimental command line flag.

Along with this, there is one functional change:
1. you can no longer reference environment variables using `$VAR` syntax, and must use the curly braces: `${VAR}`

In the future, we may want to introduce the ability to escape use `\\` so that strings with `${...}` can be interpreted as literals.

things missing:

- [x] end-to-end flag testing

### DIFF
--- a/logstash-core/lib/logstash/config/mixin.rb
+++ b/logstash-core/lib/logstash/config/mixin.rb
@@ -38,7 +38,8 @@ module LogStash::Config::Mixin
   PLUGIN_VERSION_1_0_0 = LogStash::Util::PluginVersion.new(1, 0, 0)
   PLUGIN_VERSION_0_9_0 = LogStash::Util::PluginVersion.new(0, 9, 0)
 
-  ENV_PLACEHOLDER_REGEX = /\$(?<name>\w+)|\$\{(?<name>\w+)(\:(?<default>[^}]*))?\}/
+  ALLOW_ENV_FLAG = "__ALLOW_ENV__"
+  ENV_PLACEHOLDER_REGEX = /\$\{(?<name>\w+)(\:(?<default>[^}]*))?\}/
 
   # This method is called when someone does 'include LogStash::Config'
   def self.included(base)
@@ -47,6 +48,14 @@ module LogStash::Config::Mixin
   end
 
   def config_init(params)
+    # HACK(talevy): https://github.com/elastic/logstash/issues/4958
+    # Currently, the regular plugins params argument is hijacked
+    # to pass along the `allow_env` configuration variable. This was done as to 
+    # not change the method signature of Plugin. This also makes it difficul to 
+    # reason about at the same time. ALLOW_ENV_FLAG is a special param that users 
+    # are now forbidden to set in their configuration definitions.
+    allow_env = params.delete(LogStash::Config::Mixin::ALLOW_ENV_FLAG) { false }
+
     # Validation will modify the values inside params if necessary.
     # For example: converting a string to a number, etc.
     
@@ -102,21 +111,24 @@ module LogStash::Config::Mixin
     end
 
     # Resolve environment variables references
-    params.each do |name, value|
-      if (value.is_a?(Hash))
-        value.each do |valueHashKey, valueHashValue|
-          value[valueHashKey.to_s] = replace_env_placeholders(valueHashValue)
-        end
-      else
-        if (value.is_a?(Array))
-          value.each_index do |valueArrayIndex|
-            value[valueArrayIndex] = replace_env_placeholders(value[valueArrayIndex])
+    if allow_env
+        params.each do |name, value|
+        if (value.is_a?(Hash))
+          value.each do |valueHashKey, valueHashValue|
+            value[valueHashKey.to_s] = replace_env_placeholders(valueHashValue)
           end
         else
-          params[name.to_s] = replace_env_placeholders(value)
+          if (value.is_a?(Array))
+            value.each_index do |valueArrayIndex|
+              value[valueArrayIndex] = replace_env_placeholders(value[valueArrayIndex])
+            end
+          else
+            params[name.to_s] = replace_env_placeholders(value)
+          end
         end
       end
     end
+
 
     if !self.class.validate(params)
       raise LogStash::ConfigurationError,
@@ -154,7 +166,6 @@ module LogStash::Config::Mixin
   # Process following patterns : $VAR, ${VAR}, ${VAR:defaultValue}
   def replace_env_placeholders(value)
     return value unless value.is_a?(String)
-    #raise ArgumentError, "Cannot replace ENV placeholders on non-strings. Got #{value.class}" if !value.is_a?(String)
 
     value.gsub(ENV_PLACEHOLDER_REGEX) do |placeholder|
       # Note: Ruby docs claim[1] Regexp.last_match is thread-local and scoped to
@@ -168,7 +179,6 @@ module LogStash::Config::Mixin
       if replacement.nil?
         raise LogStash::ConfigurationError, "Cannot evaluate `#{placeholder}`. Environment variable `#{name}` is not set and there is no default value given."
       end
-      @logger.info? && @logger.info("Evaluating environment variable placeholder", :placeholder => placeholder, :replacement => replacement)
       replacement
     end
   end # def replace_env_placeholders
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -45,7 +45,8 @@ module LogStash; class Pipeline
     :pipeline_batch_delay => 5, # in milliseconds
     :flush_interval => 5, # in seconds
     :flush_timeout_interval => 60, # in seconds
-    :debug_config => false
+    :debug_config => false,
+    :allow_env => false
   }
   MAX_INFLIGHT_WARN_THRESHOLD = 10_000
 
@@ -61,6 +62,7 @@ module LogStash; class Pipeline
     @settings = DEFAULT_SETTINGS.clone
     settings.each {|setting, value| configure(setting, value) }
     @reporter = LogStash::PipelineReporter.new(@logger, self)
+    @allow_env = settings[:allow_env]
 
     @inputs = nil
     @filters = nil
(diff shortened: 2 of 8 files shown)


logstash-core/spec/logstash/config/mixin_spec.rb
```diff
@@ -213,7 +217,7 @@
           "oneString" => "${FunString:foo}",
           "oneBoolean" => "${FunBool:false}",
           "oneArray" => [ "first array value", "${FunString:foo}" ],
-          "oneHash" => { "key1" => "${FunString:foo}", "key2" => "$FunString is ${FunBool}", "key3" => "${FunBool:false} or ${funbool:false}" }
```
talevy: funny, there was only one place actually using the `$VAR` scheme

talevy:
Example Usage

```bash
$ rake artifact:tar
$ tar -xzf build/logstash-5.0.0.dev.tar.gz
...
```

With `--allow-env` flag
```
$ FOO=foo bin/logstash -e 'input { generator { message => "${FOO}" count => 1 } }' --allow-env
{:pipeline_id=>"main", :config_string=>"input { generator { message => \"${FOO}\" count => 1 } }", :config_path=>nil, :allow_env=>true}
Settings: Default pipeline workers: 8
Pipeline main started
{
      "sequence" => 0,
    "@timestamp" => 2016-04-01T01:05:57.634Z,
      "@version" => "1",
          "host" => "elasticbox-2.local",
>>>>>       "message" => "foo"
}
Pipeline main has been shutdown
stopping pipeline {:id=>"main", :level=>:warn}
```

Without `--allow-env` flag
```
$ FOO=foo bin/logstash -e 'input { generator { message => "${FOO}" count => 1 } }'
{:pipeline_id=>"main", :config_string=>"input { generator { message => \"${FOO}\" count => 1 } }", :config_path=>nil, :allow_env=>false}
Settings: Default pipeline workers: 8
Pipeline main started
{
      "sequence" => 0,
    "@timestamp" => 2016-04-01T01:07:16.409Z,
      "@version" => "1",
          "host" => "elasticbox-2.local",
>>>>>       "message" => "${FOO}"
}
Pipeline main has been shutdown

```

logstash-core/lib/logstash/plugin.rb
```diff
@@ -45,6 +45,7 @@ def eql?(other)
   end
 
   def initialize(params=nil)
+    @allow_env = params.delete(LogStash::Config::Mixin::ALLOW_ENV_FLAG) { false }
```
andrewvc: I understand why we wrote this this way (we both discussed it!), but I think a comment mentioning something like 'this is a temporary hack' would be good. It'd be good to announce our intent to rewrite this weirdness at a later date here :)

logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -163,7 +166,6 @@ def replace_env_placeholders(value)
       if replacement.nil?
         raise LogStash::ConfigurationError, "Cannot evaluate `#{placeholder}`. Environment variable `#{name}` is not set and there is no default value given."
       end
-      @logger.info? && @logger.info("Evaluating environment variable placeholder", :placeholder => placeholder, :replacement => replacement)
```
andrewvc: :+1: for removing this leak of potentially confidential stuff

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -191,6 +191,40 @@
...
   end
 
+  describe "Environment Variables In Configs" do
+    let(:agent_args) { {
+      :logger => logger,
+      :auto_reload => false,
+      :reload_interval => 0.01
+    } }
+    let(:pipeline_id) { "main" }
+    let(:pipeline_config) { "input { generator { message => '${FOO}-bar' } } filter { } output { }" }
```
andrewvc: Creative test!

andrewvc:
Verified that it behaves correctly locally. I agree we need better end to end flag testing.

logstash-core/lib/logstash/plugin.rb
talevy: yeah, good call. I will leave a comment here

talevy:
Even thought we don't have a convention for testing things end-to-end, I am inclined to add a rake test for this. Since this is more of a hotfix, I was hesitant to introduce this element to the testing framework. Definitely a followup

andrewvc:
@talevy I think if we just add a spec a la what's in runner_spec currently ensuring the value is passed in that is sufficient.

logstash-core/lib/logstash/plugin.rb
ph: Can we reference an issue in that comment? that would be awesome.

logstash-core/lib/logstash/codecs/base.rb
```diff
@@ -11,7 +11,7 @@ module LogStash::Codecs; class Base < LogStash::Plugin
 
   def initialize(params={})
     super
-    config_init(@params)
+    config_init(@allow_env, @params)
```
ph: I am confused, where `@allow_env` is coming from, its not coming from the params in this method and is not declared any where?

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -191,6 +191,40 @@
...
+      it "does not interpolate environment variables" do
+        expect(subject).to receive(:fetch_config).and_return(pipeline_config)
+        subject.register_pipeline(pipeline_id, pipeline_settings)
+        expect(subject.pipelines[pipeline_id].inputs.first.message).to eq("${FOO}-bar")
+      end
+    end
+
+    context "when allow_env is true" do
+      before :each do
+        ENV["FOO"] = "foo"
```
ph: We should not make sure we are not changing the env.

We can either use a stub on `ENV` or use this flow to reset the value after the test run.

```ruby
before :each do
@foo_content =  ENV["FOO"]
ENV["FOO"] = "foo"
end

after :each do
ENV["FOO"] = @foo_content
end
```

andrewvc:
@talevy I've had similar such tests here: https://github.com/elastic/logstash/pull/4965

logstash-core/spec/logstash/agent_spec.rb
talevy: ah yes, good call

logstash-core/lib/logstash/codecs/base.rb
talevy: haha, that is the same question I had for `@params`, where is that made?

The answer is `plugin.rb`. when we call `super`, it does all this logic

ph: Oh, got it.. feel a bit of hacky but we still need to clean all of this, thanks for clarifying it for me.

talevy: I'll be the first to agree with you on that point :smile:

ph:
Added minor comments on the PR, a test on the flag itself would be great

Tested on my end, works for me.

```
~/e/logstash git:pr/4963 ❯❯❯ bin/logstash -e 'input { generator  { message => "PH is ${QUALITY}" count => 1 }}' --allow-env                                                                                                                                                                                               ✱
Pipeline main started
{
      "sequence" => 0,
    "@timestamp" => 2016-04-01T15:20:35.082Z,
      "@version" => "1",
          "host" => "sashimi",
       "message" => "PH is awesome"
}
Pipeline main has been shutdown
stopping pipeline {:id=>"main", :level=>:warn}
~/e/logstash git:pr/4963 ❯❯❯ bin/logstash -e 'input { generator  { message => "PH is ${QUALITY}" count => 1 }}'                                                                                                                                                                                                           ✱
Pipeline main started
{
      "sequence" => 0,
    "@timestamp" => 2016-04-01T15:22:02.888Z,
      "@version" => "1",
          "host" => "sashimi",
       "message" => "PH is ${QUALITY}"
}
Pipeline main has been shutdown
stopping pipeline {:id=>"main", :level=>:warn}
```

ph:
I know its temporary, but I believe we should add this flag to this document https://github.com/elastic/logstash/blob/master/docs/static/command-line-flags.asciidoc


@talevy Address the spec and it will LGTM for me.

logstash-core/lib/logstash/codecs/base.rb
colinsurprenant: can't we add the allow_env key into the @param?

talevy: the idea was to split it out as soon as possible. Since it is hijacking the `params`, I wanted to get it out of there as fast as possible

talevy: on second thought, I'll leave it in so that `config_init` is responsible to take it out

ph:
Test LGTM, I agree with @colinsurprenant's comment concerning the **params** and the **allow_env(( all the work can be done in the `config_init` method.

talevy:
alright, updated to reflect comments. let me know what you think!

ph:
LGTM, @colinsurprenant, @purbon, @andrewvc  Can you do a final review.

logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -47,6 +48,14 @@ def self.included(base)
   end
 
   def config_init(params)
+    # HACK(talevy): https://github.com/elastic/logstash/issues/4958
```
andrewvc: +1 on this explanatory comment!

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -69,6 +69,7 @@ def initialize(config_str, settings = {})
     @settings = DEFAULT_SETTINGS.clone
     settings.each {|setting, value| configure(setting, value) }
     @reporter = LogStash::PipelineReporter.new(@logger, self)
+    @allow_env = settings[:allow_env] || false
```
andrewvc: The current pattern we use is to set defaults in the DEFAULT_SETTINGS hash

andrewvc:
LGTM aside from the one nit I raised about settings value. (Tested locally to boot).

Maybe we should print out all merged env values when the `--debug-config` flag is enabled. That would require a rebase on top of #4965 . I'm fine with that being in a future PR however.

logstash-core/lib/logstash/pipeline.rb
talevy: I'll reuse that.

as a side note now that I am looking at it. the whole `configure(setting)` thing is difficult to grok

ph:
LGTM,.

Issue #4964, opened 2016-04-01
Debug mode can leak sensitive variables to log file via generated code


andrewvc:
When run in debug mode logstash will print out its full pipeline ruby source which includes all arguments to filters, including password fields, in plaintext.

PR #4965, opened 2016-04-01
Hide generated ruby code unless --debug-config argument passed

andrewvc:
We now hide this because displaying it is dangerous. Generated code may contain plaintext copies of 'password' type fields. Users can force this to appear with the --debug-config flag.

Fixes https://github.com/elastic/logstash/issues/4964

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -44,7 +44,8 @@ module LogStash; class Pipeline
     :pipeline_batch_size => 125,
     :pipeline_batch_delay => 5, # in milliseconds
     :flush_interval => 5, # in seconds
-    :flush_timeout_interval => 60 # in seconds
+    :flush_timeout_interval => 60, # in seconds
+    :debug_config => false
   }
   MAX_INFLIGHT_WARN_THRESHOLD = 10_000
 
@@ -90,7 +91,10 @@ module LogStash; class Pipeline
 
     # The config code is hard to represent as a log message...
     # So just print it.
-    @logger.debug? && @logger.debug("Compiled pipeline code:\n#{code}")
+
+    if @settings[:debug_config] && logger.debug?
+      logger.debug("Compiled pipeline code", :code => code)
+    end
 
     begin
       eval(code)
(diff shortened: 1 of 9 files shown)


andrewvc:
This could probably use better end to end flag testing. I will attempt to add some tomorrow.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -90,7 +91,10 @@ def initialize(config_str, settings = {})
 
     # The config code is hard to represent as a log message...
     # So just print it.
-    @logger.debug? && @logger.debug("Compiled pipeline code:\n#{code}")
+
+    if @settings[:debug_config] && logger.debug?
+      logger.debug("Compiled pipeline code:\n#{code}")
```
jsvd: I suggest here `logger.debug("Compiled pipeline code", :code => code)`

logstash-core/lib/logstash/runner.rb
```diff
@@ -189,7 +193,8 @@ def execute
 
     @agent.register_pipeline("main", @pipeline_settings.merge({
                           :config_string => config_string,
-                          :config_path => config_path
+                          :config_path => config_path,
+                          :debug_compiled => @debug_compiled
```
jsvd: here it's `debug_compiled` but in pipeline.rb it's `debug_config`?

jsvd:
this PR removes some of the compiled print statements but there are many others.
have you tried running `bin/logstash -f config --debug` with this PR and a config with a pwd field? I'm getting a lot of hits on the password

andrewvc:
I've added some tests to runner_spec that I think close the gap here.

ph:
When we run `--debug-config` without the `--debug` we should tell the user they need to be enable the `--debug` flag too. We also need to update the documentation https://github.com/elastic/logstash/blob/master/docs/static/command-line-flags.asciidoc

ph:
Just missing updating the `--debug-config` in the `command-line-flags.asciidoc` and everything is LGTM.

jsvd:
also, I believe there's a log call that needs to be mocked:

```
un options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}
...........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................Updating dummy
......Validating dummy
.....................................................ERROR: Configuration reloading also requires passing a configuration path with '-f yourlogstash.conf'
........Agent: Configuring metric collection {:level=>:debug, :file=>"logstash/agent.rb", :line=>"154", :method=>"setup_metric_collection"}
Agent: Stopping metric periodic pollers {:level=>:debug, :file=>"logstash/agent.rb", :line=>"147", :method=>"stop_background_services"}
.Agent: Configuring metric collection {:level=>:debug, :file=>"logstash/agent.rb", :line=>"154", :method=>"setup_metric_collection"}
Agent: Stopping metric periodic pollers {:level=>:debug, :file=>"logstash/agent.rb", :line=>"147", :method=>"stop_background_services"}
...............................Error parsing @timestamp string value=foo
Unrecognized @timestamp value type=class java.lang.Long
...............................................................................................................................................................................................Plugin foo version 9.0.0.0 does not exist
.........................................................................................................

Finished in 35.16 seconds (files took 6.33 seconds to load)
1091 examples, 0 failures

Randomized with seed 26190

```

other than that and what ph mentioned, LGTM

PR #4974, opened 2016-04-01
Add a note concerning the ruby filter for the change in the java event

ph:


docs/static/breaking-changes.asciidoc
```diff
@@ -70,3 +70,32 @@ Starting with the 2.0 release of Logstash, the default value of the `filter_work
 plugins is half of the available CPU cores, instead of 1. This change increases parallelism in filter execution for
 resource-intensive filtering operations. You can continue to use the `-w` flag to manually set the value for this option,
 as in previous releases.
+
+
+[float]
+=== Ruby filter Changes
```
dedemorton: Ruby Filter Changes (use title capitalization--by the way, the title of this topic should also be changed to say Breaking Changes)

docs/static/breaking-changes.asciidoc
```diff
@@ -70,3 +70,32 @@ Starting with the 2.0 release of Logstash, the default value of the `filter_work
...
+
+With the migration to the Java Event we have changed how we can access the internal data, the object doesn't return a
```
dedemorton: I suggest changing this section to say: "With the migration to the Java Event (https://github.com/elastic/logstash/issues/4191[Issue 4191]), we have changed how you can access internal data. The Event object no longer returns a reference to the data. Instead, it returns a copy. This might change how you do manipulation of your data, especially when working with nested hashes."

docs/static/breaking-changes.asciidoc
```diff
@@ -70,3 +70,32 @@ Starting with the 2.0 release of Logstash, the default value of the `filter_work
...
+reference to the data but a copy instead. This change how you have to do manipulation of your data especially when working with nested hashes.
+When working with nested hashes its recommended to use the `fieldref` syntax instead of using multiple brakets.
```
dedemorton: Change to say, "When working with nested hashes, it’s recommended that you use...."

docs/static/breaking-changes.asciidoc
```diff
@@ -70,3 +70,32 @@ Starting with the 2.0 release of Logstash, the default value of the `filter_work
...
+----------------------------------
+filter { 
+  ruby {
+    codec => "event['uuid'] = event['uuid'].gsub(/b/, '')" # instead of using event['uuid'].gsub!(/b/, '')
+  }
+}
+----------------------------------
+
+
+**Examples:**
```
dedemorton: You can remove this. You don't need a second Examples heading.

dedemorton:
@ph I'm done with my review.

docs/static/breaking-changes.asciidoc
```diff
@@ -70,3 +70,31 @@ Starting with the 2.0 release of Logstash, the default value of the `filter_work
...
+----------------------------------
+filter { 
+  ruby {
+    codec => "event['uuid'] = event['uuid'].gsub(/b/, '')" # instead of using event['uuid'].gsub!(/b/, '')
+  }
+}
+----------------------------------
+
+[source, js]
+----------------------------------
```
ph: @dedemorton For the source blocks should I only use one? I am not sure how it will render.

ph:
@dedemorton Updated with your recommendations and I have added a small question.

docs/static/breaking-changes.asciidoc
dedemorton: @ph I would try using two separate blocks with an extra line in between. I usually add a bit of text to introduce examples (helps avoid this problem and provides some context). But I'd built it and see how it looks.

dedemorton:
LGTM

PR #4987, opened 2016-04-03
(2.x Backport) Hide sensitive AST/config data unless --debug-config argument passed

andrewvc:
(This is a 2.x backport of https://github.com/elastic/logstash/pull/4965)

We now hide this because displaying it is dangerous. Generated code
and other AST data may contain plaintext copies of 'password' type fields.
Users can force this to appear with the --debug-config flag.

Fixes https://github.com/elastic/logstash/issues/4964

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -11,7 +11,7 @@ require "logstash/pipeline"
 
 class LogStash::Agent < Clamp::Command
 
-  attr_reader :pipelines
+  attr_reader :pipelines, :config_loader
 
   DEFAULT_INPUT = "input { stdin { type => stdin } }"
   DEFAULT_OUTPUT = "output { stdout { codec => rubydebug } }"
@@ -58,6 +58,10 @@ class LogStash::Agent < Clamp::Command
   option "--verbose", :flag, I18n.t("logstash.agent.flag.verbose")
   option "--debug", :flag, I18n.t("logstash.agent.flag.debug")
 
+  option ["--debug-config"], :flag,
+         I18n.t("logstash.runner.flag.debug_config"),
+         :attribute_name => :debug_config, :default => false
+
   option ["-V", "--version"], :flag,
     I18n.t("logstash.agent.flag.version")
 
@@ -89,7 +93,6 @@ class LogStash::Agent < Clamp::Command
     @pipelines = {}
     @pipeline_settings ||= { :pipeline_id => "main" }
     @upgrade_mutex = Mutex.new
-    @config_loader = LogStash::Config::Loader.new(@logger)
   end
 
   def pipeline_workers=(pipeline_workers_value)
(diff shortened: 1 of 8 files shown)


logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -10,6 +10,8 @@
   subject { LogStash::Agent.new("", "") }
 
   before :each do
+    allow(LogStash::Config::Loader).to receive(:new).with(any_args).and_call_original
```
andrewvc: Aha, I think this is the cause of the threadsafety issues.

purbon: If this is reused in different threads, then this is for sure the issue.

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -343,5 +343,25 @@
     end
   end
 
+  describe "debug_config" do
+    let(:pipeline_string) { "input {} output {}" }
+    let(:pipeline) { double("pipeline") }
+
+    it "should set 'debug_config' to false by default" do
+      expect(LogStash::Pipeline).to receive(:new).with(pipeline_string, hash_including(:debug_config => false)).and_return(pipeline)
```
purbon: much simply workflow as before :+1:, good work with the simplification!

purbon:
please jenkins, test it.

purbon:
@andrewvc test still fail, do you have an idea why?

jsvd:
LGTM

ph:
LGTM

andrewvc:
There is a bug here, it doesn't actually print the config anymore.

jsvd:
patch to fix this:

```patch
% git diff logstash-core
diff --git a/logstash-core/lib/logstash/agent.rb b/logstash-core/lib/logstash/agent.rb
index 11bc0f3..fafbb3e 100644
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -110,7 +110,7 @@ class LogStash::Agent < Clamp::Command
 
   def debug_config=(debug_config)
     @config_loader.debug_config = debug_config
-    @debug_config
+    @debug_config = debug_config
   end
 
   def validate_positive_integer(str_arg)
```

logstash-core/lib/logstash/agent.rb
```diff
@@ -110,7 +110,7 @@ def pipeline_batch_delay=(pipeline_batch_delay_value)
 
   def debug_config=(debug_config)
     @config_loader.debug_config = debug_config
-    @debug_config
+    @debug_config = true
```
jsvd: any reason to do `@debug_config = true` instead of `@debug_config = debug_config`?

ph:
I am testing it.

ph:
LGTM, tests passes and the manual test too.

Issue #4991, opened 2016-04-04
service logstash restart does not take KILL_ON_STOP_TIMEOUT into account


purbon:
Tested with: Ubuntu 14.04 package when calling `service logstash restart` the script does not take into account the var `KILL_ON_STOP_TIMEOUT` leaving former LS instance alive.

see:

```
root@vagrant-ubuntu-trusty-64:/etc/init.d# echo $KILL_ON_STOP_TIMEOUT
1
root@vagrant-ubuntu-trusty-64:/etc/init.d# /etc/init.d/logstash restart
Killing logstash (pid 13256) with SIGTERM
Waiting logstash (pid 13256) to die...
Waiting logstash (pid 13256) to die...
Waiting logstash (pid 13256) to die...
Waiting logstash (pid 13256) to die...
Waiting logstash (pid 13256) to die...
Waiting logstash (pid 13256) to die...
Waiting logstash (pid 13256) to die...
Waiting logstash (pid 13256) to die...
Waiting logstash (pid 13256) to die...
logstash stop failed; still running.
logstash started.
```

this might happen when LS resist to stop, for example when there is an stale output.

The expected workaround for this is to use `start` and `stop` when using the KILL_ON_STOP_TIMEOUT.

cgreatorexatonshape:
+1 with additional information, cross-post from the forums...

When logstash doesn't stop in a timely manner, the init script exits with a bug trying to handle the timeout, and when calling it with restart, it ends up starting multiple copies.

sudo service logstash restart
Killing logstash (pid 23698) with SIGTERM
Waiting logstash (pid 23698) to die...
Waiting logstash (pid 23698) to die...
Waiting logstash (pid 23698) to die...
Waiting logstash (pid 23698) to die...
Waiting logstash (pid 23698) to die...
Waiting logstash (pid 23698) to die...
Waiting logstash (pid 23698) to die...
Waiting logstash (pid 23698) to die...
Waiting logstash (pid 23698) to die...
/etc/init.d/logstash: 99: [: -eq: unexpected operator
logstash stop failed; still running.
logstash started.

Seen in version 2.3.1 on ubuntu 14 LTS

The relevant line in the code is...

  if [ $KILL_ON_STOP_TIMEOUT -eq 1 ] ; then
...and that variable doesn't exist anywhere in a default configuration, hence the init script bugs out.

[...]

PR #4992, opened 2016-04-04
Fix for ubuntu package stop function 

### LINKED ISSUE #4940: /etc/init.d/logstash: 99: /etc/init.d/logstash: [[: not found
When I use `service logstash restart` on Ubuntu 14.04 LTS, will raise this error.

Logstash version is `2.2.3`, and installed from official repo.

This because Ubuntu does not use `sh` alias for `bash`, and `sh` does not support built-in `[[` term.

purbon:
by having use proper portable `[` command instead of the bash `[[` based one when stop checks for `KILL_ON_STOP_TIMEOUT` env to be set.

Fixes #4940

As of a solution I when for using the posix `[` command, instead of moving the hole scripts to bash due to the reasons explained in:

* https://wiki.ubuntu.com/DashAsBinSh#A.5B
* and due to the reason that in this script this was actually an exception command, as rest of the similar checks were using the `[` options.

> The major reason to switch the default shell was efficiency. bash is an excellent full-featured shell appropriate for interactive use; indeed, it is still the default login shell. However, it is rather large and slow to start up and operate by comparison with dash.

From the same ubuntu people, notice that in ubuntu now `bin/sh` -> `bin/dash`

related issue: #4991

### DIFF
--- a/pkg/logstash.sysv
+++ b/pkg/logstash.sysv
@@ -96,7 +96,7 @@ stop() {
       sleep 1
     done
     if status ; then
-      if [[ $KILL_ON_STOP_TIMEOUT -eq 1 ]] ; then
+      if [ $KILL_ON_STOP_TIMEOUT -eq 1 ] ; then
         echo "Timeout reached. Killing $name (pid $pid) with SIGKILL. This may result in data loss."
         kill -KILL $pid
         echo "$name killed with SIGKILL."


purbon:
@ph @jsvd your review here is much welcome.

To reproduce you need to test in an ubuntu based machine by calling stop, or restart, when for example using this dummy config:

```
root@vagrant-ubuntu-trusty-64:/etc/init.d# cat ../logstash/conf.d/dummy.conf 
input {
  generator {}
}

output {
 stdout { codec => dots }
 elasticsearch {
   hosts => [ "127.0.0.1:9200" ]
 }
}
```
and the expected elasticsearch output is down (for any reason)

jsvd:
LGTM, I've checked and we don't use [[]] any where else :+1:

purbon:
please jenkins, test it

purbon:
please jenkins, test it

ph:
LGTM

PR #4995, opened 2016-04-04
revert to jruby 1.7.23 for regex regression

### LINKED ISSUE #4977: JRuby 1.7.24 regex thread safety regression
per logstash-plugins/logstash-filter-useragent#25 and logstash-plugins/logstash-filter-useragent#26 it seems that the JRuby regex code is not thread safe. 

We need to
1- confirm that the JRuby regex implementation is not thread safe
2- if it's a bug, report and/or submit fix
3- verify out usage of regexes to make sure this problem does not surface elsewhere.

colinsurprenant:
fixes #4977 until fixed upstream in JRuby jruby/jruby#3670

### DIFF
--- a/rakelib/vendor.rake
+++ b/rakelib/vendor.rake
@@ -1,6 +1,6 @@
 namespace "vendor" do
   VERSIONS = {
-    "jruby" => { "version" => "1.7.24", "sha1" => "0c321d2192768dfec419bee6b44c7190f4db32e1" },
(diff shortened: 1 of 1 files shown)


ph:
LGTM, rake test:plugins went through for me without issues.

jsvd:
LGTM

colinsurprenant:
merged into master, 2.3, 2.x, 5.0

PR #4998, opened 2016-04-04
Cherry pick doc changes from 2.3

dedemorton:
Contains doc changes from 2.3 that are not currently in 2.x.


docs/static/logstash-docs-home.asciidoc
```diff
@@ -1,30 +0,0 @@
-[[logstash-docs-home]]
```
andrewvc: Why was this removed?

andrewvc:
@dedemorton LGTM aside from the one deletion. Why was that page removed?

dedemorton:
@andrewvc The page that I deleted is no longer being used anywhere in the docs. Suyog said it was OK to remove it. Is there some reason why it should not be removed?

dedemorton:
I'm going to merge this now because the doc build really happens in logstash-docs anyhow, where the file has already been removed. :-)

Issue #4834, opened 2016-03-17
Java Event type casting conversion optimization


colinsurprenant:
While discussing over a possible type conversion bug in the Java Event @guyboertje suggested a very interesting new strategy to minimize the Ruby<->Java conversion costs.

Currently when setting an Event field value from the Ruby world, that value is converted into a pure Java object and stored into the Java event. Every time that field is read from the Ruby world later on, it is converted back into a proper Ruby object. There is obviously a cost to that type conversion which occurs at every field set and get.

@guyboertje's proposal is to create a Java value class for the Java Event that would store both Java and Ruby values representations. The actual type conversion would be done once and lazily only when required. For example, upon setting an initial value from the Ruby world into the Java Event, only the original Ruby value object would be stored into that new Java value class. Further read of that value fro the Ruby would would not require any type conversion, the original Ruby object could be directly be returned. If/when the Java Event need to access the Java object representation of that original Ruby value, then the original Ruby object would be converted once and store in that value class. If that Java representation of that value is read again, no conversion would be necessary.

The same idea would apply for the reverse case when a value would be originally set from the Java world and eventually read from the Ruby world.

This could very well significantly improve field access performance!

guyboertje:
Woohoo - I want to do this, pick me, pick me.

colinsurprenant:
@guyboertje you definitely are on the short list for a honorific "JRuby<->Java type conversion" PhD :laughing: so yeah, please do :thumbsup:

guyboertje:
It is a little more complicated that I thought.  Timestamps being one problem where the RubyTimestamp wraps Java Timestamp.

colinsurprenant:
which should not be a problem since we do have a proper Ruby class for it (as a JRuby ext) so the same logic can be applied?

colinsurprenant:
Can we close this since #4999 is now merged?

PR #4999, opened 2016-04-04
Implementation - Java Event type casting conversion optimization

guyboertje:
See #4834

See #5614 for an explanation of BiValues


guyboertje:
@colinsurprenant - Is this merge-worthy?

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
```diff
@@ -0,0 +1,98 @@
...
+import org.jruby.Ruby;
+import org.jruby.RubyNil;
+import org.jruby.runtime.builtin.IRubyObject;
+
+import java.io.Serializable;
+
+public class RubyJavaObject implements Serializable {
+    private IRubyObject rubyValue;
+    private Object javaValue;
+    private boolean is_ruby_nil;
```
andrewvc: Should this be camel case?

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public boolean hasJavaValue() {
+        return null != javaValue;
+    }
+
+    public boolean isRubyNil() {  return is_ruby_nil; }
+
+    @Override
+    public String toString() {
+        final StringBuffer sb = new StringBuffer("RubyJavaObject{");
```
andrewvc: Curious as to the rationale behind using StringBuffer here. Is thread safety a concern here?

guyboertje: No rationale, just copy/pasta from JRuby.

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
guyboertje: I guess so. Will change. Thanks

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
andrewvc: Probably makes sense to use a StringBuilder IMHO. Should be a little faster, but its not a big deal either way.

guyboertje:
@colinsurprenant + @andrewvc : bump!

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -4,6 +4,7 @@
 import com.logstash.ext.JrubyTimestampExtLibrary;
 import org.joda.time.DateTime;
 import org.jruby.RubySymbol;
+import org.jruby.runtime.builtin.*;
```
colinsurprenant: is this needed in the Event Class?

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -119,7 +121,16 @@ public void setTimestamp(Timestamp t) {
         this.data.put(TIMESTAMP, this.timestamp);
     }
 
+    public Object getFieldForRuby(String reference) {
```
colinsurprenant: shouldn't we move Ruby *specific* methods in the JRuby ext?

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -119,7 +121,16 @@ public void setTimestamp(Timestamp t) {
...
+        return _getField(reference);
+    }
+
     public Object getField(String reference) {
+        Object val = _getField(reference);
+        return Valuefier.unconvert(val);
+    }
+
+    private Object _getField(String reference) {
```
colinsurprenant: I am not sure I like the underscore naming here `_getField` - it does not convey the specific purpose of this method, not is it documented

colinsurprenant:
@guyboertje one of the design consideration of the *pure* Java Event was to **not** have to hold any JRuby objects to avoid future serialization problems as `IRubyObject`s are not `Serializable` because they hold reference to the JRuby runtime. I just want to make sure that this is taken into consideration as the Java Event need to be stay `Serializable`.

guyboertje:
@colinsurprenant : I can make the `IRubyObject rubyValue` reference in RubyJavaObject volatile. That will prevent it from being serialized no? And when it is deserialized I imagine that the rubyValue will be null and so will be converted if asked for by the Jruby Ext.

logstash-core-event-java/src/main/java/com/logstash/Event.java
guyboertje: no - will remove

logstash-core-event-java/src/main/java/com/logstash/Event.java
guyboertje: See reply to `_getField`

logstash-core-event-java/src/main/java/com/logstash/Event.java
guyboertje: How about I make it public and call it `getUnconvertedField` and call this method from the JRubyExt - because the JRubyExt needs to extract the Ruby Value, which after all is the whole point behind this enhancement. Then I remove `getFieldForRuby`.

colinsurprenant:
@guyboertje yes, I think that would work. If all the JRuby objects are well encapsulated in the `RubyJavaObject` class and that this class is correctly `Serialisable` by making the `IRubyObject` values `transcient` that would work I think. We still need to verify the case where a Ruby object was set and not converted to Java, in which case if/when we serialize we may loose that original value altogether. 

This raises the question of doing lazy Ruby->Java conversion, knowing that persistence will force that conversion upon serialization?

guyboertje:
@colinsurprenant - this means implementing the Serializable interface properly in JavaRubyObject.

colinsurprenant:
@guyboertje yes. how do you see the fact that persistence will end up in always forcing a Ruby->Java conversion? IMO it's ok. I am just wondering that if we know we are going through persistence, is there value to do the Ruby->Java  lazily vs doing it upfront? thinking-out-loud here

guyboertje:
@colinsurprenant - I'm thinking its OK too.
So when do you see serialization occurring - going into the PQ and deserialization - coming off disk?

.gitignore
```diff
@@ -28,3 +28,7 @@ rspec.xml
 .vendor
```
colinsurprenant: I don't think we should introduce `.gitignore` in a PR ...

logstash-core-event-java/.gitignore
```diff
@@ -7,3 +7,5 @@ build
 # Intellij
```
colinsurprenant: same - I don't think we should introduce .gitignore in a PR ...

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -119,7 +120,12 @@ public void setTimestamp(Timestamp t) {
         this.data.put(TIMESTAMP, this.timestamp);
     }
 
-    public Object getField(String reference) {
+     public Object getField(String reference) {
```
colinsurprenant: indentation?

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -139,8 +145,15 @@ public void setField(String reference, Object value) {
         } else if (reference.startsWith(METADATA_BRACKETS)) {
             this.metadata_accessors.set(reference.substring(METADATA_BRACKETS.length()), value);
         } else {
-            this.accessors.set(reference, value);
+            this.accessors.set(reference, setableValue(value));
+        }
+    }
+
+    private Object setableValue(Object value) {
```
colinsurprenant: I am not sure I like `setableValue`, what about simply `rubyJavaValue` ? I know it can also return Map and List... not sure ... ?

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
```diff
@@ -0,0 +1,124 @@
...
+import java.io.InvalidObjectException;
+import java.io.ObjectStreamException;
+import java.io.Serializable;
+import java.io.ObjectInputStream;
+
+public final class RubyJavaObject implements Serializable {
+    private static final long serialVersionUID = 4358368393114318978L;
+
+    private transient IRubyObject rubyValue;
+    private transient boolean isRubyNil;
```
colinsurprenant: is there a particular reason why we keep a specific boolean for tagging nil values? it seems that `hasRubyValue() && rubyValue.isNil()` would work and avoid managing this boolean state?

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
```diff
@@ -0,0 +1,124 @@
...
+    private Object javaValue;
+
+    public RubyJavaObject(Object javaValue) {
```
colinsurprenant: the fact that we did not create a class for holding Java values and have opted for `Object` result in having method overloading relying on the implicit more specific type rule to pick the right method which is not very clear IMO. It is probably too big a change to consider refactoring this to use a "proper" object model for values, but maybe we could a t least add a few unit tests to show that the `RubyJavaObject` method overloading really work?

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
```diff
@@ -0,0 +1,124 @@
...
+        if (hasRubyValue() && isRubyNil) {
+            return null;
+        }
+        if (javaValue == null) {
+            javaValue = Javafier.deep(rubyValue);
+        }
+        return javaValue;
+    }
+
+    public Object javaGetRubyValue() {
```
colinsurprenant: I am not sure what `javaGetRubyValue` means? is it meant as returning the unconverted rubyValue ?

logstash-core-event-java/src/main/java/com/logstash/Rubyfier.java
```diff
@@ -45,8 +50,7 @@ private static RubyArray deepList(Ruby runtime, final List list) {
         final RubyArray array = runtime.newArray(length);
 
         for (Object item : list) {
-            // use deepOnly because RubyArray.add already calls JavaUtil.convertJavaToUsableRubyObject on item
-            array.add(deepOnly(runtime, item));
+            array.add(deep(runtime, item));
```
colinsurprenant: I think at this point the `deepOnly` method is not needed anymore? should we keep it?

logstash-core-event-java/src/main/java/com/logstash/Valuefier.java
```diff
@@ -0,0 +1,105 @@
+package com.logstash;
```
colinsurprenant: is that class pretty much the same idea or complement the `Javafier` class?

colinsurprenant:
left some inline comments. really great job so far! 
maybe we're a bit thin on tests?

logstash-core-event-java/src/main/java/com/logstash/Event.java
guyboertje: how about `maybeDualize(value)`

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
guyboertje: So my understanding is that the Jruby extension will always call methods that involve IRubyObject and have no use for the methods that involve Object and that the retrieving of the Ruby value will convert if necessary.

However the java side may need to retrieve the Ruby value without conversion - during (de)serialization perhaps.

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
guyboertje: The general intention is that one or other internal value **should** be set.

I want to avoid the implicit assumption that javaValue of null means a rubyValue of RubyNil.

See `javaGetRubyValue` and the `equals` method.

logstash-core-event-java/src/main/java/com/logstash/Rubyfier.java
guyboertje: I guess its stored in git so if we remove it now we can get it back if its needed again.

Will remove.

logstash-core-event-java/src/main/java/com/logstash/Valuefier.java
guyboertje: The `Javafier` only deals with Ruby objects. I chose not to make it aware of `RubyJavaValue` because it is a Java object.

The `Rubyfier` only deals with Java objects and unwraps `RubyJavaValue` values.

The `Valuefier` deals with either Ruby or Java objects.
It converts them to their opposite equivalents and wraps values in a `RubyJavaValue`.
It unconverts on the Java side and unwraps RubyJavaValue values to Java.

`Javafier.deep` is now only called in `Valuefier` or `RubyJavaValue`
`Rubyfier.deep` is only called in `RubyJavaValue` or the JRuby Extension.

BTW, there is a bug in the `unconvertMap` method - calls convert instead of unconvert. (will fix)

logstash-core-event-java/src/main/java/com/logstash/Event.java
guyboertje: Or `wrapRubyJavaObject`

.gitignore
guyboertje: I am only updating it.

logstash-core-event-java/src/main/java/com/logstash/RubyJavaObject.java
guyboertje: You are correct. After fixing the Javafier for JavaProxy instances, It occurs to me to have specific BiValues for each non-collection ruby/java pair. If so, we move the Javafier/Rubyfier specifics into these types.
WDYT?
I will make this change and we can discuss in here.

guyboertje:
Benchmarks:
Master:
```
bin/bundle exec ruby ./benchmark/event_accessor.rb
Same Event instance

 ----------> event.get("[foo][bar][foobar]") => morebar

Warming up --------------------------------------
          Deep fetch   133.669k i/100ms
          Deep fetch      2.607M (± 5.3%) i/s -     26.065M
```

This PR:
```
bin/bundle exec ruby ./benchmark/event_accessor.rb
Same Event instance

 ----------> event.get("[foo][bar][foobar]") => morebar

Warming up --------------------------------------
          Deep fetch   182.242k i/100ms
          Deep fetch      5.746M (± 6.0%) i/s -     57.224M
```

guyboertje:
Working through the Jenkins bugs now.

guyboertje:
@colinsurprenant - updated for JavaProxy handling.

guyboertje:
@talevy - please review this while Colin and I work together on PQ.

.gitignore
```diff
@@ -33,3 +33,4 @@ qa/.vagrant
 qa/.rspec
 qa/acceptance/.vagrant
 qa/Gemfile.lock
+
```
talevy: empty line?

guyboertje: came from the rebase from master

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -253,10 +265,16 @@ private Timestamp initTimestamp(Object o) {
             } else if (o instanceof String) {
                 // second most frequent
                 return new Timestamp((String) o);
+            } else if (o instanceof StringBiValue) {
+                return new Timestamp(((StringBiValue) o).javaValue());
+            }else if (o instanceof TimeBiValue) {
```
colinsurprenant: cosmetic: `}else if` -> `} else if`

logstash-core-event-java/src/main/java/com/logstash/Rubyfier.java
```diff
@@ -1,65 +1,60 @@
...
     }
 
     private static RubyHash deepMap(Ruby runtime, final Map<?, ?> map) {
         RubyHash hash = RubyHash.newHash(runtime);
-
-        for (Map.Entry<?, ?> entry : map.entrySet()) {
-            // use deepOnly on value because RubyHash.put already calls JavaUtil.convertJavaToUsableRubyObject on items
-            hash.put(entry.getKey(), deepOnly(runtime, entry.getValue()));
+        for (Map.Entry entry : map.entrySet()) {
+            IRubyObject k = JavaUtil.convertJavaToUsableRubyObject(runtime, entry.getKey());
```
colinsurprenant: not sure I understand why we need to call JavaUtil.convertJavaToUsableRubyObject here?

logstash-core-event-java/src/main/java/com/logstash/Valuefier.java
colinsurprenant: bump for fix

logstash-core-event-java/src/main/java/com/logstash/bivalues/ObjectBiValue.java
```diff
@@ -0,0 +1,41 @@
...
+import org.jruby.javasupport.JavaUtil;
+import org.jruby.runtime.builtin.IRubyObject;
+
+import java.io.ObjectStreamException;
+
+// Guy: I am not sure we need this class
+// it was meant to be supplied as a fall-back when BiValues.newBiValue(obj) returns nothing.
+// i.e. when JRuby wraps a value in a JavaProxy, however, we don't want Maps, Lists or arrays in a BiValue
+// or when there is a JRuby class that has no java equivalent, for example, a custom class.
+// in the case of a ConcreteProxy, toJava(Object.class) will return the wrapped Java object.
```
colinsurprenant: so do we trash it? doesn't that defeat the purpose of not accepting "unknown" objects?

logstash-core-event-java/src/main/java/com/logstash/ConvertedList.java
```diff
@@ -0,0 +1,224 @@
...
+    private final List<T> delegate;
+
+    public ConvertedList(List<T> delegate) {
+        this.delegate = delegate;
+    }
+    public ConvertedList() {
+        this.delegate = new ArrayList<>();
+    }
+
+    public static List<Object> newFromList(List<Object> list) {
```
colinsurprenant: I think it would make more sense to have the return object be a `ConvertedList` and not the `List` generic since this is a static factory that generates `ConvertedList` specifically, just like the class constructor.

colinsurprenant:
@guy can you high-level describe the intent/strategy of the `ConvertedMap` and `ConvertedList` classes? What I understand is that getting a map/list object will always trigger a deep map/list copy to the right target world (Ruby/Java) and the leaf values will leverage bivalues thus gaining on the bivalues conversion caching. But the map/list object themselves will always be recreated every time they are accessed?

logstash-core-event-java/src/main/java/com/logstash/Valuefier.java
guyboertje: fixed in the BiValue redesign

logstash-core-event-java/src/main/java/com/logstash/bivalues/ObjectBiValue.java
guyboertje: I will remove it for the final squash.

logstash-core-event-java/src/main/java/com/logstash/Rubyfier.java
guyboertje: I guess we can assume that `RubyHash.put` will always convert it - will revert.

guyboertje:
RubyHash and Java HashMap are converted into a ConvertedMap instance. Likewise, RubyArray and ArrayList are converted into a ConvertedList. These two classes are used so we can be certain that the internal data Map is always a collection of ConvertedMap, ConvertedList or BiValue. Any value not of these three types must be converted. If we were to unconvert a normal HashMap we don't know if the leaf nodes are BiValues or not.  In Javafier and Rubifier there was always this uncertainty of whether the Maps and Lists encountered were converted or not.

suyograo:
@colinsurprenant @guyboertje can you work to merge this before beta1? Thanks.

colinsurprenant:
I'd really like to see the performance impact using our logstash-benchmark tool, @purbon  ?

colinsurprenant:
Overall LGTM
- there are still some missing commits followups:
  - https://github.com/elastic/logstash/pull/4999#discussion-diff-69772533
  - https://github.com/elastic/logstash/pull/4999#discussion-diff-69771506
- are all tests: core* and plugins passing?
- I'd like to see a first performance run in *logstash-benchmarks* 

So I am definitely +1 on moving forward with this, my only concern given the extent of the changes is the lack of real-world mileage but on the other hand we are still in a beta cycle so the risk is not as high.

suyograo:
+1 on merging. We can always measure performance and readjust accordingly.

suyograo:
As long as tests pass, and other comments are ok'd, that is..

guyboertje:
> there are still some missing commits followups:
#4999 (diff)
#4999 (diff)
are all tests: core* and plugins passing?

updated for comments above
tests core and plugins pass (although I had to change the initTimestamp method to handle NullBiValue) so al-in-all good call 🏁

suyograo:
Please squash and merge @guyboertje

purbon:
@colinsurprenant @guyboertje keep in mind please that after you merge and don't need this PR to benchmarked anymore, remove it from https://github.com/elastic/logstash-performance-testing/blob/master/web/api/config.yml .  Thanks a lot guys!

PR #5001, opened 2016-04-04
Prevent config leak on output delegator close

andrewvc:
On shutdown logstash may leak config information for outputs including
plaintext passwords


ph:
LGTM

colinsurprenant:
LGTM

PR #5009, opened 2016-04-04
Update terminology used for security product

dedemorton:
Product name update for 5.0. I didn't update the corresponding doc link for alpha 1 (it still points to current) because the location of security content is currently in flux. 

The following variable will be added to the index.asciidco in logstash-docs:

:security:              X-Pack Security


talevy:
LGTM

ph:
Lgtm

PR #5000, opened 2016-04-04
Hide sensitive AST/config data unless --debug-config argument passed

andrewvc:
We now hide this because displaying it is dangerous. Generated code
and other AST data may contain plaintext copies of 'password' type fields.
Users can force this to appear with the --debug-config flag.

Fixes https://github.com/elastic/logstash/issues/4964

PR #5025, opened 2016-04-05
Hide sensitive AST/config data unless --debug-config argument passed

### LINKED ISSUE #4964: Debug mode can leak sensitive variables to log file via generated code

purbon:
From former PR #5000, 

> We now hide this because displaying it is dangerous. Generated code
and other AST data may contain plaintext copies of 'password' type fields.
Users can force this to appear with the --debug-config flag.

test are passing properly for this one, when testing this please make sure to use the local code base and not the released gem as there is no snapshot currently available for this branch that contains the code. 

Fixes #4964

### DIFF
--- a/docs/static/command-line-flags.asciidoc
+++ b/docs/static/command-line-flags.asciidoc
@@ -42,6 +42,11 @@ Logstash has the following flags. You can use the `--help` flag to display this 
 --debug
  Increase verbosity to the last level (trace), more verbose.
 
+--debug-config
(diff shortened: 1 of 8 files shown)


purbon:
please jenkins, test this.

ph:
@purbon I will be closing this one since https://github.com/elastic/logstash/pull/5000 got merged in.

PR #5026, opened 2016-04-05
Add ruby breaking change to the 5.0 docs

dedemorton:
Adds note about the ruby filter change plus fixes a minor formatting issue.


ph:
LGTM

andrewvc:
@dedemorton great work! I think we’ve discovered that there may be a way for us to preserve the old ruby filter behavior after all. It’s still something under active discussion however. We’re not sure if it is 100% the way we want to go. We'll be opening a design discussion issue on it very soon.

dedemorton:
Closing this because the problem is being fixed in the code.

PR #5038, opened 2016-04-06
revert to logstash-core-event in 2.3 branch

colinsurprenant:
revert to logstash-core-event in the 2.3 branch. 
to complete this we will need to move forward with a snapshot release and publish a new logstash-core gem which will drop the dependency on logstash-core-event-java and generate a new .lock file and update it in the 2.3 branch.


colinsurprenant:
this has been merged in #5039

PR #5039, opened 2016-04-06
Gemfile/2.3.1 snapshot1

ph:


logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
```diff
@@ -1,3 +1,3 @@
 # encoding: utf-8
-# 1.3.0 tracks LS 2.3.0
-LOGSTASH_CORE_PLUGIN_API = "1.3.0"
+# 1.7.0 tracks LS 2.3.1
```
jsvd: 1.7.0 tracks ls 2.3.1.snapshot1

colinsurprenant:
LGTM - verified that the next plugin-api version is correctly 1.7.0.

jsvd:
LGTM

PR #5049, opened 2016-04-07
Fix Rakelib package.rake to use logstash-plugin pack

purbon:
to generate the plugins bundle, so the command rake package:plugins-all and rake package:plugins-default works creating the offline bundle.

This should be merged to master and 5.0 branches


ph:
oops LGTM :)

jsvd:
LGTM

purbon:
:shipit:

PR #5050, opened 2016-04-07
Fix Rakelib package.rake to use bin/plugin pack

purbon:
to generate the plugins bundle, so rake package:plugins-all and rake package:plugins-default works as expected generating the bundles for offline installation.

This should be merged with branches < 5.0 as they still use the old `bin/plugin` command.

### DIFF
--- a/rakelib/package.rake
+++ b/rakelib/package.rake
@@ -1,7 +1,7 @@
 namespace "package" do
 
   task "bundle" do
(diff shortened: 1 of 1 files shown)


purbon:
related to #5049

ph:
oops LGTM

jsvd:
LGTM

purbon:
:shipit:

PR #5056, opened 2016-04-07
Lock file for 2.3.1

ph:


Gemfile.jruby-1.9.lock
```diff
@@ -167,16 +167,16 @@ GEM
...
       pry (~> 0.10.1)
       rubyzip (~> 1.1.7)
       stud (~> 0.0.19)
       thread_safe (~> 0.3.5)
       treetop (< 1.5.0)
-    logstash-core-event (2.3.1.snapshot1-java)
-    logstash-core-plugin-api (1.7.0-java)
-      logstash-core (>= 2.0.0, <= 2.3.1.snapshot1)
+    logstash-core-event (2.3.1-java)
+    logstash-core-plugin-api (1.8.0-java)
```
purbon: I guess this minor update makes sense, isn't?

colinsurprenant:
LGTM

Gemfile.jruby-1.9.lock
ph: Yes, we update minor on every release we made of logstash, include `pre-release` version, this stategy allow us to expand the version constraint.  Look at #5055 and https://rubygems.org/search?utf8=%E2%9C%93&query=logstash-core-plugin-api history release and the constraints.

purbon: ok, thanks for the clarification

then go, go :shipit: when jenkins is green :LGTM:

Gemfile.jruby-1.9.lock
```diff
@@ -167,16 +167,16 @@ GEM
       i18n (= 0.6.9)
       jrjackson (~> 0.3.7)
       jruby-openssl (= 0.9.13)
-      logstash-core-event (= 2.3.1.snapshot1)
+      logstash-core-event (= 2.3.1)
```
andrewvc: Out of curiosity, why is core-event not a -java dep?. I'm not really sure how this works.

ph: Actually bundler will use the platform to fetch the correct gems, there is never the `-java` in the name of the gems. See https://rubygems.org/gems/logstash-core-event, there is a `-java` after the version number. You can release multiples gem with the same name but with different platform target.

 The exception if the `logstash-core-event-java` see https://rubygems.org/gems/logstash-core-event-java

ph: look at the nokogiri gems.

https://rubygems.org/gems/nokogiri/versions

ph: Since I only publish the `logstash-core-event` using jruby the gems is only available on the jruby platform.

andrewvc: Sounds good to me!

andrewvc:
LGTM

PR #5063, opened 2016-04-07
Small experiment to have test run with local code or either released code

### LINKED ISSUE #5062: Executing Logstash test vs released logstash-core gem or vs the local code
There is actually an annoying gem in our testing infrastructure, we already bite this a few times (including myself), I will try to explain it here, including a solution for this (my proposal).

If you encounter yourself running test for a branch that has already been released, let's say 5.0 (for this example), that has a logstash-core gem (version 5.0.0.alpha1) online. This means the gemfile in this branch points to this gem, so our test will be run by default by pulling this gem. The problems comes when we had merged some code that might not be already published (missing snapshot available), then strange errors will show up.

 This is a problem because:

* Errors shown are strange 
* This causes us to have a delay until we figure it out there is missing code available online.
* rake test:core is not really testing (for this branches) the code changes, but the codes available online.
* CI gets confused for the same reason.

Test that should happen here:

* Is the current code base working properly.
* Is the latest codebase available online.

Solution for this:

We need to kind of test runners, one that runs test with the Gemfiles pointing to the code in the local repository and another that runs the same kind of test, but pointing the core gems to the last ones available online. This means we should have to kind of tasks to run tests:

* rake test:core (default run the test with the local codebase)
* rake test:released:core (run the test pointing to a released gem)

purbon:
This PR adds and changes some commands behaviours

* Changes the `rake test:core` command to run with local code base, this is done by updating (on the fly) the Gemfile and Gemfile.lock, that late will be reverted.

* Add the `rake test:released:core` command that basically runs with the current Gemfile status.

* Add a warning in the `bin/rspec` command to be sure users know vs witch codebase they are running the test (either released gems or local codebase).

The idea of this is to help fixing the inconsistencies when we run test in a branch that has already been released, so the Gemfile is actually pointing to released gem, so In case there is code merged, but not published test jobs fail.

Fixes #5062


purbon:
@andrewvc your thoughts on here are very appreciate it.

purbon:
please jenkins, test this

purbon:
@suyograo @ph also your thoughts are welcome here, happy to move this forward soon.

lib/bootstrap/rspec.rb
```diff
@@ -8,5 +8,14 @@
 require "rspec/core"
 require "rspec"
 
+require_relative "../rake/rspec"
+gemfiles_helpers = GemfileHelpers.new.load
+puts "WARNING: You're running the test with the current set of logstash core gems"
```
andrewvc: I have no idea how this works. Why do we know this at this point?

andrewvc:
On my box running `rake test:core` does not run any tests.

Output:

```
~/p/logstash (purbon-feature/test_from_code) $ rake test:core
^[[5~Invoking bundler update...
Resolving dependencies...
Using addressable 2.3.8
Using cabin 0.8.1
Using arr-pm 0.0.10
Using backports 3.6.8
Using ffi 1.9.10
Using childprocess 0.5.9
Using numerizer 0.1.1
Using chronic_duration 0.10.6
Using clamp 0.6.5
Using coderay 1.1.1
Using concurrent-ruby 1.0.0
Using multipart-post 2.0.0
Using faraday 0.9.2
Using minitar 0.5.4
Using file-dependencies 0.1.6
Using filesize 0.0.4
Using json 1.8.3
Using fpm 1.3.3
Using gems 0.8.3
Using i18n 0.6.9
Using jar-dependencies 0.3.2
Using jrjackson 0.3.8
Using jruby-monitoring 0.3.1 (was 0.3.0)
Using jruby-openssl 0.9.13
Using ruby-maven-libs 3.3.3
Using ruby-maven 3.3.10
  jar dependencies for logstash-core-event-java-5.0.0.dev-java.gemspec . . .
      com.fasterxml.jackson.core:jackson-core:2.7.1:compile
      com.fasterxml.jackson.core:jackson-databind:2.7.1-1:compile
Using logstash-core-event-java 5.0.0.dev from source at ./logstash-core-event-java
Using method_source 0.8.2
Using slop 3.6.0
Using spoon 0.0.4
Using pry 0.10.3
Using puma 2.16.0
Using rubyzip 1.1.7
Using rack 1.6.4
Using rack-protection 1.5.3
Using tilt 2.0.2
Using sinatra 1.4.7
Using stud 0.0.22
Using thread_safe 0.3.5
Using polyglot 0.3.5
Using treetop 1.4.15
Using logstash-core 5.0.0.dev from source at ./logstash-core
Using logstash-core-plugin-api 1.0.0 from source at ./logstash-core-plugin-api
Using sawyer 0.6.0
Using octokit 3.8.0
Using bundler 1.9.10
Bundle complete! 18 Gemfile dependencies, 46 gems now installed.
Gems in the group development were not installed.
Resolving dependencies...
```

andrewvc:
Hmmm, It does the right think if I `rake test:install-core` first. It just doesn't do the right thing if I do a clean  checkout and run `rake test:core`.

lib/rake/rspec.rb
```diff
@@ -0,0 +1,173 @@
+# encoding: utf-8
+require_relative "../bootstrap/environment"
+require 'pathname'
+require_relative "../pluginmanager/gemfile"
+
+class GemfileHelpers
```
andrewvc: Should this be a module? It doesn't seem very object like. At the very least maybe a singleton?

andrewvc:
Conceptually this makes sense to me and is a great way around the huge pain of being pinned to the wrong version. Great idea @purbon ! It's hacky, of course, but I don't have any better ideas, so it's good :)

LGTM

colinsurprenant:
Before moving forward with this we should wait until we finalise the unified-build integration strategy which has to deal with the problem of embedded vs published core gems. I'd like this to be solved in a consistent "correct" way before moving ahead with just a quick fix on this.

purbon:
@colinsurprenant I agree with you and the idea of doing the right thing, for me know most of the pain comes out having us dealing with imaginary issues because the issue described here.

I don't think we have to wait to the unified release, I do really would like to get this fixed sooner as is annoying and a waste of time when you encounter this. Can we discuss might be the right solution for this?

colinsurprenant:
@purbon I am not saying we have to wait for the unified release, I am saying we should wait until we go through the process of agreeing on the right way to solve this moving forward and this will happen in the short term. It's an annoyance but there is no emergency to move ahead with something that does not address the bigger picture.

purbon:
closing this until a "correct" way to get this implemented.

PR #5069, opened 2016-04-08
Adds a method to list all defined methods in the API

purbon:
This PR adds a _directory_ listing resource, this kind of methods are useful to discover defined methods in an API.

### DIFF
--- a/logstash-core/lib/logstash/api/lib/app.rb
+++ b/logstash-core/lib/logstash/api/lib/app.rb
@@ -25,6 +25,19 @@ module LogStash::Api
       @factory = CommandFactory.new(settings.service)
     end
 
+    get "/_directory" do
+      data = {}
+      ROUTES.each_pair do |namespace, _routes|
+        routes = {}
+        _routes.each_pair do |verb, methods|
+          next if "HEAD" == verb
+          routes[verb] = methods.map { |m| "http://localhost:9600#{namespace}/#{m[0].source.gsub("\\A/","").gsub("\\z", "")}" }
+        end
+        data[namespace] = routes
+      end
+      respond_with data
+    end
+
     not_found do
       status 404
       as   = params.has_key?("human") ? :string : :json
--- a/logstash-core/lib/logstash/api/lib/app.rb
+++ b/logstash-core/lib/logstash/api/lib/app.rb
@@ -25,7 +25,7 @@ module LogStash::Api
       @factory = CommandFactory.new(settings.service)
     end
 
-    get "/_directory" do
+    get "/_meta/api" do
       data = {}
       ROUTES.each_pair do |namespace, _routes|
         routes = {}
--- a/logstash-core/lib/logstash/api/lib/app.rb
+++ b/logstash-core/lib/logstash/api/lib/app.rb
@@ -31,7 +31,7 @@ module LogStash::Api
         routes = {}
         _routes.each_pair do |verb, methods|
           next if "HEAD" == verb
-          routes[verb] = methods.map { |m| "http://localhost:9600#{namespace}/#{m[0].source.gsub("\\A/","").gsub("\\z", "")}" }
+          routes[verb] = methods.map { |m| "http://#{request.host}:#{request.port}#{namespace}/#{m[0].source.gsub("\\A/","").gsub("\\z", "")}" }
         end
         data[namespace] = routes
       end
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/api/lib/app.rb
```diff
@@ -25,6 +25,19 @@ def initialize(app=nil)
       @factory = CommandFactory.new(settings.service)
     end
 
+    get "/_directory" do
+      data = {}
+      ROUTES.each_pair do |namespace, _routes|
+        routes = {}
+        _routes.each_pair do |verb, methods|
+          next if "HEAD" == verb
+          routes[verb] = methods.map { |m| "http://localhost:9600#{namespace}/#{m[0].source.gsub("\\A/","").gsub("\\z", "")}" }
```
ph: This doesn't take into consideration the user configuration of the host and port?

logstash-core/lib/logstash/api/lib/app.rb
```diff
@@ -25,6 +25,19 @@ def initialize(app=nil)
       @factory = CommandFactory.new(settings.service)
     end
 
+    get "/_directory" do
```
ph: Is there any prior art for the name of this route? Maybe if we look at hypermedia type of api?

logstash-core/lib/logstash/api/lib/app.rb
purbon: nope, if we want to add a method like this we certainly should do that change.

logstash-core/lib/logstash/api/lib/app.rb
purbon: The most I saw is having proper documentation in our case we've
https://www.elastic.co/guide/en/logstash/5.0/monitoring.html, or custom
solutions, any pointer would be helpful. Basic idea of resources like this
is enable easy discoverly, as we know no one really looks at documentation.

- purbon

ph: +1 I like the idea and its useful for developing clients.

logstash-core/lib/logstash/api/lib/app.rb
suyograo: Can we use `/_meta/api` instead of `_directory`?

logstash-core/lib/logstash/api/lib/app.rb
suyograo: Can we use `/_meta/api` instead of `_directory`?

suyograo:
Lets resurrect this. Also can you show us what the output will look like?

logstash-core/lib/logstash/api/lib/app.rb
purbon: sure, works for me

purbon:
screen shot 2016-05-18 at 12 12 09

the resource output for now it looks like ^^, basically listing all resources grouped per URI and HTTP verb. As you can see in the code, this will fetch all defined resources, no need to update this method in any way to be up to date.

logstash-core/lib/logstash/api/lib/app.rb
purbon: @ph amended. thanks for taking a review on this.

suyograo:
@purbon what about adding more info to these, like:
```
{ 
   /_node {
      GET: 
          [
            {
                "url": "http://localhost:9600/_node/hot_threads",
                "description": "Gets the current hot threads for Logstash.",
                "docs": "https://www.elastic.co/guide/en/logstash/current/hot-threads-api.html"
            }
          ]
 ....
}

this makes it useful during offline access.

purbon:
while it might be interesting to have something like what you proposed in https://github.com/elastic/logstash/pull/5069#issuecomment-220196251, my initial was to have a _simple_ and _automatic_ endpoint, that basically require no configuration or external documentation to get it.

The only way I might think to get this somehow automated is by having the requested additional information as endpoints documentation (in code). What do you think?. Idea is to have as less mantaintanance for this as possible.

purbon:
This PR is dependant on #5307, when this refactor gets in, this code should be adapted, lucky there is not much we should adapt.

suyograo:
@purbon Without description, I don't see how this is useful to a new user. Also I agree with you on this becoming outdated as we add more APIs. We have to make it easier to maintain. What are your thoughts on closing this until users actually feel its hard to navigate or discover our APIs? I rather we put our efforts at this point in improving docs and making URI, as self-explanatory as possible..

purbon:
sure, lets close it.

PR #5071, opened 2016-04-08
KILL_ON_STOP_TIMEOUT fix

samuraiii:
Possible way to fix unary error ...

### DIFF
--- a/pkg/logstash.sysv
+++ b/pkg/logstash.sysv
@@ -36,7 +36,7 @@ LS_CONF_DIR=/etc/logstash/conf.d
 LS_OPEN_FILES=16384
 LS_NICE=19
 LS_OPTS=""
-
+KILL_ON_STOP_TIMEOUT=0
 
 [ -r /etc/default/$name ] && . /etc/default/$name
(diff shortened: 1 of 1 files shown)


purbon:
Hi, 
   thanks a lot for your contribution, at my understanding the idea of this var is that people can actually set it on the environment, with this it will be always be set, isn't?

On the other side, to continue you will need to sign the CLA process: https://www.elastic.co/contributor-agreement/

/cheers

samuraiii:
Yes it will be set but (default as I uderstand should be this value), it is by default set in /etc/sysconfig/logstash but there should be just differences from default ... (file should be commented whole on clean install and all defaults should be set in /etc/init.d/logstash and possible overridden by in  /etc/sysconfig/logstash as you do it now with all other variables).

samuraiii:
You can commit that fix yourself I am not going to sign that for one-liner fix...

purbon:
Oh sorry to read that, I would be very happy to merge your fix, but can not do without the CLA, I hope you understand. I will close this PR, but keep the issue open to implement and think about this issue. thanks a lot for your time and understanding.

PR #5075, opened 2016-04-08
build core gems in same process as rake command instead of spawning new process

talevy:


rjernst:
+1

colinsurprenant:
+1 on the idea, yes.

rakelib/plugin.rake
```diff
@@ -86,9 +87,15 @@ namespace "plugin" do
 
     Rake::Task["plugin:clean-local-core-gem"].invoke(name, path)
 
-    puts("[plugin:build-local-core-gem] Building #{File.join(path, name)}.gemspec")
+    gemspec = "#{File.join(path, name)}.gemspec"
 
-    system("cd #{path}; gem build #{name}.gemspec")
+    puts("[plugin:build-local-core-gem] Building #{gemspec}")
+
+    spec = Gem::Specification.load gemspec
```
colinsurprenant: any reason for omitting parenthesis on method call? this is not really a dsl context here, same comment below

talevy: reason: I extracted these steps out of the original `gem build` code that was being run via the GemRunner. I can add the parens

purbon:
This is a very good idea @talevy, on the other side, what do you think about encapsulating all this logic into their own classes (that will be easy to test) and then doing only one call method here? would this be a good approach to test this rake calls?

talevy:
yes, @purbon. I will see what I can do about testing. currently the gem builds. but something weird is happening in the actual packaging of the tar artifact

talevy:
@purbon, after realizing a mistake I made with regards to the loading path of all the `spec.files`, the actual code has been reduced to simply executing the package build code. so wrapping this up in a class seems like too much.

jordansissel:
Code change LGTM. Waiting on rake to finish testing.

talevy:
@jordansissel 

here is a more isolated way of testing the task:

```
# logstash-core-plugin-api
rake plugin:build-local-core-gem[logstash-core-plugin-api,./logstash-core-plugin-api]
# logstash-core-event-java
rake plugin:build-local-core-gem[logstash-core-event-java,./logstash-core-event-java]
```

jordansissel:
Built an artifact and it works successfully - behavior seems the same. +1 for not using system().

LGTM

colinsurprenant:
sweet change! LGTM
:shipit:

PR #5082, opened 2016-04-11
Added reload parameter

skokhanovskiy:
Added reload configuration support from #4513 to the init script

### DIFF
--- a/pkg/logstash.sysv
+++ b/pkg/logstash.sysv
@@ -127,6 +127,12 @@ status() {
   fi
 }
 
+reload() {
+  if status ; then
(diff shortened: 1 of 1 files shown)


purbon:
Thanks a lot for your contribution @skokhanovskiy , to move forward with this It would be necessary you sign the http://www.elasticsearch.org/contributor-agreement/, keep in mind to use the same email you used to create this PR as this will highly simplify internal procedures.

Looking forward to it,

skokhanovskiy:
I signed the agreement after the pull request. I must create a new one?

PR #5096, opened 2016-04-12
Make the default list of plugins bundled with logstash a json file

purbon:
So it can be consumed either by LS or external entities that might be interested to know the list of default plugins (for example the new plugins CI status page)

@ph your thoughts here? this basically implements your proposal to have the logstash-ci status page split re plugins.

### DIFF
--- a/rakelib/default_plugins.rb
+++ b/rakelib/default_plugins.rb
@@ -1,142 +1,7 @@
+require "logstash/json"
+
 module LogStash
   module RakeLib
-
-    # plugins included by default in the logstash distribution
-    DEFAULT_PLUGINS = %w(
-      logstash-input-heartbeat
-      logstash-codec-collectd
-      logstash-output-xmpp
-      logstash-codec-dots
-      logstash-codec-edn
-      logstash-codec-edn_lines
-      logstash-codec-fluent
-      logstash-codec-es_bulk
-      logstash-codec-graphite
-      logstash-codec-json
-      logstash-codec-json_lines
-      logstash-codec-line
-      logstash-codec-msgpack
-      logstash-codec-multiline
-      logstash-codec-netflow
-      logstash-codec-plain
-      logstash-codec-rubydebug
-      logstash-filter-clone
-      logstash-filter-csv
-      logstash-filter-date
-      logstash-filter-dns
-      logstash-filter-drop
-      logstash-filter-fingerprint
-      logstash-filter-geoip
-      logstash-filter-grok
-      logstash-filter-json
-      logstash-filter-kv
-      logstash-filter-metrics
-      logstash-filter-mutate
-      logstash-filter-ruby
-      logstash-filter-sleep
-      logstash-filter-split
-      logstash-filter-syslog_pri
-      logstash-filter-throttle
-      logstash-filter-urldecode
-      logstash-filter-useragent
-      logstash-filter-uuid
-      logstash-filter-xml
-      logstash-input-couchdb_changes
-      logstash-input-elasticsearch
-      logstash-input-exec
-      logstash-input-file
-      logstash-input-ganglia
-      logstash-input-gelf
-      logstash-input-generator
-      logstash-input-graphite
-      logstash-input-http
-      logstash-input-http_poller
-      logstash-input-imap
-      logstash-input-irc
-      logstash-input-jdbc
-      logstash-input-log4j
-      logstash-input-lumberjack
-      logstash-input-pipe
-      logstash-input-rabbitmq
-      logstash-input-redis
-      logstash-input-s3
-      logstash-input-snmptrap
-      logstash-input-sqs
-      logstash-input-stdin
-      logstash-input-syslog
-      logstash-input-tcp
-      logstash-input-twitter
-      logstash-input-udp
-      logstash-input-unix
-      logstash-input-xmpp
-      logstash-input-kafka
-      logstash-input-beats
-      logstash-output-cloudwatch
-      logstash-output-csv
-      logstash-output-elasticsearch
-      logstash-output-file
-      logstash-output-graphite
-      logstash-output-http
-      logstash-output-irc
-      logstash-output-kafka
-      logstash-output-nagios
-      logstash-output-null
-      logstash-output-pagerduty
-      logstash-output-pipe
-      logstash-output-rabbitmq
-      logstash-output-redis
-      logstash-output-s3
-      logstash-output-sns
-      logstash-output-sqs
-      logstash-output-statsd
-      logstash-output-stdout
-      logstash-output-tcp
-      logstash-output-udp
-      logstash-output-webhdfs
-    )
-
-    # plugins required to run the logstash core specs
-    CORE_SPECS_PLUGINS = %w(
-      logstash-filter-clone
-      logstash-filter-mutate
-      logstash-filter-multiline
-      logstash-input-generator
-      logstash-input-stdin
-      logstash-input-tcp
-      logstash-output-stdout
-    )
-
-    TEST_JAR_DEPENDENCIES_PLUGINS = %w(
-      logstash-input-kafka
-    )
-
-    TEST_VENDOR_PLUGINS = %w(
-      logstash-codec-collectd
-    )
-
-    ALL_PLUGINS_SKIP_LIST = Regexp.union([
-      /^logstash-filter-yaml$/,
-      /example$/,
-      /drupal/i,
-      /^logstash-output-logentries$/,
-      /^logstash-input-jdbc$/,
-      /^logstash-output-newrelic$/,
-      /^logstash-output-slack$/,
-      /^logstash-input-neo4j$/,
-      /^logstash-output-neo4j$/,
-      /^logstash-input-perfmon$/,
-      /^logstash-output-webhdfs$/,
-      /^logstash-input-rackspace$/,
-      /^logstash-output-rackspace$/,
-      /^logstash-input-dynamodb$/,
-      /^logstash-filter-language$/,
-      /^logstash-input-heroku$/,
-      /^logstash-output-google_cloud_storage$/,
-      /^logstash-input-journald$/,
-      /^logstash-input-log4j2$/,
-      /^logstash-codec-cloudtrail$/
-    ])
-
 
     # @return [Array<String>] list of all plugin names as defined in the logstash-plugins github organization, minus names that matches the ALL_PLUGINS_SKIP_LIST
     def self.fetch_all_plugins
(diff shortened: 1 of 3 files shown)


rakelib/default_plugins.rb
```diff
@@ -1,114 +1,10 @@
...
-      logstash-output-s3
-      logstash-output-sns
-      logstash-output-sqs
-      logstash-output-statsd
-      logstash-output-stdout
-      logstash-output-tcp
-      logstash-output-udp
-      logstash-output-kafka
-    )
+    DEFAULT_PLUGINS = JSON.parse(File.read("rakelib/default_plugins.json"))["default_plugins"]
```
ph: Any reason why don't use our json gems?

purbon: I went to this because I was not sure we load actually logstash-core when running rake.

ph:
@purbon Lets take the time to drop the content of all the constants from this file in the JSON file. (CORE_SPECS_PLUGINS, TEST_JAR_DEPENDENCIES_PLUGINS, TEST_VENDOR_PLUGINS, ALL_PLUGINS_SKIP_LIST)

I think its better that we don't have to edit multiples places when we are dealing with the plugins, WDTY @suyograo  and @purbon  ?

rakelib/default_plugins.rb
ph: Good point, can you test if its the case?

purbon:
I understand the idea of having this in one file, however we do think this separation is good for now. We add the json file in order to enable others to read it (not just logstash), so:

* Other components for now are not interested to read about CORE_SPECS for example.
* Nearly all this are going to be gone as soon as we improve our testing and packages procedures, so for me they are living zombies for now.
* We do need at the end of the day to read this file and make it accessible from ruby.

and must say I understand your idea of having all in a single place, this is are the reasons why I just keep the default ones there.

rakelib/default_plugins.rb
purbon: works, pushing the change.

colinsurprenant:
Any reason not to have just a (more readable) plain json array, sorted by type and plugin name? in the end this is human-edited, for example:

```json
[
"logstash-codec-collectd",
"logstash-codec-dots",
"logstash-codec-edn",
"logstash-codec-edn_lines",

"logstash-input-heartbeat",

"logstash-output-xmpp",
"logstash-output-zeromq",
]
```

colinsurprenant:
also agree with @ph, if we are to extract one list of plugins, lets be consistent and extract all plugin lists.

guyboertje:
@purbon, @ph, @colinsurprenant:
I can't help feeling this should be a JSON object with the key being the plugin and the value another object e.g. `{"default" : true,"all_plugins_skip_list" : true}`
Plus abstracting the reading and querying of the file into a ruby class that rake tasks or others can use.
Or have several JSON array based files that the ruby class reads - maybe glob for `*_plugins_list.json`.
e.g. `default_plugins_list.json`, `core_specs_plugins_list.json` etc.

Basically isolate the format (and its evolution) of the file(s) behind an API.
If the need for CORE_SPECS_PLUGINS goes away then the API returns `[]` if queried for.

purbon:
@guyboertje this will make the format a big more interesting I see, but end of the day I tried to make it really simple and reduce process in consumers of this file.

suyograo:
+1 on making this a JSON object that @guyboertje mentioned. I'd like to use it to add more metadata, for example, to add a warning blob for unsupported plugins in docs.

ph:
> warning blob for unsupported plugins in docs.
Oh this is a nice idea.

purbon:
Ok, will make it like this.

/purbon

purbon:
Changes pushed to include metadata in the file, opinions welcome.

rakelib/default_plugins.rb
```diff
@@ -168,5 +17,24 @@ def self.is_released?(plugin)
       require 'gems'
       Gems.info(plugin) != "This rubygem could not be found."
     end
+
+    def self.fetch_plugins_for(clazz)
+      LogStash::Json.load(::File.read("rakelib/plugins-metadata.json")).select do |_, metadata|
+        metadata[clazz]
+      end.keys
+    end
```
ph: Its not really a clazz, more like a type or field?

purbon: works for me, will update.

- purbon

ph:
@purbon I think @guyboertje meant to encapsulate the logic of dealing with the `plugins-metadata.json` into his own class, to allow us to enrich it with other methods to query information for specific plugins, this would be perfectly handy when we do docs generation.

@guyboertje Would you mind clarifying your idea?

purbon:
sorry, did not get it like this, but the idea is ok for me, even this class
will be mostly a wrapper for what the rakelib file is doing.

purbon:
@ph @guyboertje most of this logic will be just the small function I shared, but will add more if you with in a class, however for now I don't see much of benefit .... unless we move this class to core, what it might make a bit more sense to me. What do you think?

rakelib/plugins-metadata.json
```diff
@@ -0,0 +1,905 @@
+{
+  "logstash-codec-cloudtrail": {
+    "default-plugins": false,
```
colinsurprenant: is there particular reason for choosing to list by plugins instead of by types? for example:
```
{
  "default-plugins" : [
    "logstash-codec-dots",
    "logstash-codec-collectd",
    ...
  ],
  "core-specs": [
    ...
  ],
  ...
}
```

guyboertje:
I think it should be a class in core or devutils maybe.
The class is responsible for finding and loading the json file(s).
It should act like a mini database not quite SQL queries but a simple query mechanism.
"All entries having property X"
"All entries having property X and Y"
"All entries having property X or Y"
"All entries not having property X"

Maybe map/reduce - map "All with X" reduce "emit properties M and N"

WDYT?

colinsurprenant:
@guyboertje are we seeing any other properties than being part of a group? Sorry to say be it feels we are over-engineering this very simple use-case which is to externalize the plugins groupings, in json, so that these can be consumed by potential external processes like CI. IMO internally in LS, we could just have a method which returns a hash with group names as keys and list of plugins as values, no?

rakelib/plugins-metadata.json
guyboertje: @colinsurprenant @ph @purbon - it very much depends on whether the format is designed for humans first or the encapsulating class first.

For example (foolish but bear with me), we could simply have `"logstash-codec-dots" : {"id" : 101, (more properties)}` and some indexes ` "default-plugins" : [101, 104, 110]`
This would be friendly to the class but hard for humans.

guyboertje:
@colinsurprenant - I think it can be what ever we want - from one simple class with a simple file format to ES hosted docs with a front-end.

I had not thought of other systems reading the json files. I only imagined that it would be used in our code or maybe in the terminal via rake.

I guess CI tools could use rake.

guyboertje:
@colinsurprenant - Other properties: "unsupported", "community maintained".

suyograo:
> Other properties: "unsupported", "community maintained".

+1.

colinsurprenant:
> Other properties: "unsupported", "community maintained".

And for what use-case exactly?

I am not sure why we need to go beyond simple grouping here. Plugins already have metadata/properties embedded in their gemspecs. Not sure why we need to duplicate that here for as-of-yet non-existing use cases?

I'm not following.

purbon:
In that case I agree, in part, with @colinsurprenant, the first and only reason this was introduced is to power the feature ci plugin status page, and this was introduced because we were asking ourself to have a distinction between default and non default plugins.

but I also see the benefits to introduce a bit more metadata for things like documentation, however I see a few things here:

* We've metadata in each plugin, aka gem, this can be used in documentation use case.
* I would rather keep this file simple, and only classification based (aka default, test, etc...) as this will give the easiest (DRY solution). 

We can always iterate, and see improvements, but for now I would rather stick to the current solution here, or also a simplification. 

What do you think?

ph:
As the original reviewer of this PR and read the comments from others.

I usually would be in favor of @guyboertje's comment and encapsulating everything in a class and I dont think it would be overengineering stuff. **BUT**, in the other hand its only used currently used in the build process so I am OK in the quicker solution to just externalize the data.

ph:
LGTM, lets move forward, this is a first step to dry it out.

ph:
Can we make this move forward?

guyboertje:
LGTM

ph:
@purbon lets make sure we sync the content of the file, will do another round to make sure its correct.

The `LGTM` was more I am OK with the current way of making it work/

purbon:
sure, will sync and ask you two again, I was more looking to know if there is any pending thing, a part from rebasing 👍  thanks!

purbon:
@guyboertje @ph list updated and PR rebased, let me know what do you think and if this is good to be merged? I understand this would be master and 5.0, isn't?  or do you see it also inside other 2.x branches?

What do you think?

ph:
@purbon I would to have this merged for master, 2.X and 5.0, I don't think it will be too hard to do it. The list of plugins is probably different between branches though.

The build is failling, but this might not be related to the current code?

purbon:
@ph checking the build, I think the rebase got the branch screwed, fixing.

purbon:
jenkins, please test this.

purbon:
please jenkins, test this.

ph:
retriggering build.

purbon:
please jenkins, test this.

ph:
LGTM, travis run successfully.

ph:
@purbon I believe this is ready to merge.

suyograo:
@purbon I can't even build the artifacts with this

```
suyog@machine:~/ws/elastic/logstash (pr/5096)$ rake artifact:tar
rake aborted!
no such file to load -- concurrent
org/jruby/RubyKernel.java:1040:in `require'
/Users/suyog/ws/elastic/logstash/logstash-core/lib/logstash/config/defaults.rb:3:in `(root)'
org/jruby/RubyKernel.java:1040:in `require'
/Users/suyog/ws/elastic/logstash/logstash-core/lib/logstash/config/cpu_core_strategy.rb:1:in `(root)'
org/jruby/RubyKernel.java:1040:in `require'
/Users/suyog/ws/elastic/logstash/logstash-core/lib/logstash/config/cpu_core_strategy.rb:3:in `(root)'
org/jruby/RubyKernel.java:1040:in `require'
/Users/suyog/ws/elastic/logstash/logstash-core/lib/logstash/environment.rb:1:in `(root)'
org/jruby/RubyKernel.java:1040:in `require'
/Users/suyog/ws/elastic/logstash/logstash-core/lib/logstash/environment.rb:3:in `(root)'
org/jruby/RubyKernel.java:1040:in `require'
/Users/suyog/ws/elastic/logstash/logstash-core/lib/logstash/json.rb:1:in `(root)'
/Users/suyog/ws/elastic/logstash/logstash-core/lib/logstash/json.rb:2:in `(root)'
org/jruby/RubyKernel.java:1059:in `load'
/Users/suyog/ws/elastic/logstash/rakelib/default_plugins.rb:1:in `(root)'
/Users/suyog/ws/elastic/logstash/rakelib/default_plugins.rb:1:in `(root)'
file:/Users/suyog/.rvm/rubies/jruby-1.7.25/lib/jruby.jar!/jruby/kernel19/kernel.rb:1:in `(root)'
file:/Users/suyog/.rvm/rubies/jruby-1.7.25/lib/jruby.jar!/jruby/kernel19/kernel.rb:13:in `require_relative'
/Users/suyog/ws/elastic/logstash/rakelib/plugin.rake:1:in `(root)'
org/jruby/RubyKernel.java:1059:in `load'
org/jruby/RubyKernel.java:1079:in `eval'
/Users/suyog/.rvm/gems/jruby-1.7.25/bin/jruby_executable_hooks:15:in `(root)'
```

suyograo:
master build is fine. Its something with this PR.

ph:
@suyograo I just did another test with this PR and I successfully build an artefact, but I do get a warning for redefined constants.

```
/t/logstash git:pr/5096 ❯❯❯ rake artifact:tar
/private/tmp/logstash/lib/bootstrap/environment.rb:14 warning: already initialized constant LOGSTASH_HOME
/private/tmp/logstash/lib/bootstrap/environment.rb:14 warning: already initialized constant LOGSTASH_HOME
Invoking bundler install...
```

*Notes, I did this from a clean clone*

suyograo:
@purbon this needs a rebase with master as well.

jordansissel:
> to add a warning blob for unsupported plugins in docs

Slight tangent: There is no concept of "unsupported plugins". "Supported" plugins is only relevant in commercial customer support relationships and is not relevant to any of the open source customers.

ph:
Rebased in #5793

Issue #5081, opened 2016-04-11
CHANGELOG.md inconsistent format => not displaying correctly on github


wiibaa:
A lot of CHANGELOG.md files do not contains a blank line to separate first level titles, therefore they are not rendered correctly, for example https://github.com/logstash-plugins/logstash-filter-date/blob/master/CHANGELOG.md jumps from 2.0 to 2.1.6

Also from https://raw.githubusercontent.com/logstash-plugins/logstash-input-jdbc/master/CHANGELOG.md second level heading are used 

@suyograo can the **detailed markdown format of changelog** be strengthened and clearly specified in https://www.elastic.co/guide/en/logstash/current/_versioning_plugins.html. It could improve readability a lot when reading it in different plugins, or they could even be injected in ELK to display evolution of the Logstash plugin ecosystem or power a crazy logstash-caniuse (as http://caniuse.com) if someone has some free-time

@ph can a bulk magic be applied here to correct / validate these files with advanced regex/grokking and maybe integrated in the CI ?

I may be reading changelog too much lately... :D

ph:
@wiibaa Don't stop reading them!

It might be possible to create a script to clean them up
- lookup for `#` + the version scheme and add new line?
- Then we can clean them to make sure the headers uses the first level title.

For the CI you mean, correcting the format on push? Or making sure we have one?
I think the format validation should be part of the review process.

Concerning: https://www.elastic.co/guide/en/logstash/current/_versioning_plugins.html Would you mind doing a PR for that? I agree we need to use the same structure everywhere.

wiibaa:
@ph I made a first try in the filter-date plugin, please have a look. 

First I strongly suspect that the missing newline was introduced by the mass update, and followed by human-error copy-pasting, can you add a newline in the mass update when writing the changelog already please.

When possible, I will prepare a PR to the doc, my idea is to propose some general rules first:
* version should be a second-level header
* each change should be started with ` - ` (one space one dash one space)
* optional usage of comma-separated tags ended by a colon
* multi-line description must be followed by 3 spaces => aligned under first line
* newline after version block

Translated in regex to validated the file, and tested against my proposal for date filter it becomes
(not incuding a specific match for tags yet)
`/\A(?:^## \d\.\d\.\d(?: \(yanked\))?$(?:\n^ - .*$(?:\n^   .*$)*)+\n+)+\z/`

[...]

PR #5103, opened 2016-04-13
[Maintainer doc] Detail expected format of CHANGELOG.md

wiibaa:
Starting from #5081 reporting CHANGELOG rendering issue on github and after reviewing several CHANGELOG.md in different plugins, I thought that there is room for improvement in harmonizing how the changelog is actually formatted.

* This PR contains documentation proposal, [preview available here](http://gist.asciidoctor.org/?github-wiibaa/logstash/changelog-format-doc//docs/static/maintainer-guide.asciidoc&stylesheet=foundation.css#_changelog_management)
* https://github.com/logstash-plugins/logstash-filter-date/pull/64 apply the defined rules to the date filter plugin, you can compare [current](https://github.com/logstash-plugins/logstash-filter-date/blob/master/CHANGELOG.md) to [proposed](https://github.com/wiibaa/logstash-filter-date/blob/format-changelog/CHANGELOG.md)

@ph please review and discuss with the team, I'm waiting for your green light before looking to other plugins changelog ;)

### DIFF
--- a/docs/static/maintainer-guide.asciidoc
+++ b/docs/static/maintainer-guide.asciidoc
@@ -114,7 +114,7 @@ Please see following annotated example and see a concrete example in https://raw
 
 ----
 <1> Latest version is the first line of CHANGELOG.md
-<2> Each version identifier should be a level-1 section title using `##`
+<2> Each version identifier should be a level-2 header using `##`
 <3> One change description is described as a list item using a dash `-` aligned under the version identifier
 <4> One change can be tagged by a word and suffixed by `:`. +
     Common tags are `bugfix`, `feature`, `doc`, `test` or `internal`.
--- a/docs/static/maintainer-guide.asciidoc
+++ b/docs/static/maintainer-guide.asciidoc
@@ -88,10 +88,40 @@ should kept to the minimum.
 
 Every plugin should have a changelog (CHANGELOG.md).  If not, please create one.  When changes are made to a plugin, make sure to include a changelog entry under the respective version to document the change.  The changelog should be easily understood from a user point of view.  As we iterate and release plugins rapidly, users use the changelog as a mechanism for deciding whether to update.
 
-Changes that are not user facing should be tagged as “Internal: ”.  For example:
+Changes that are not user facing should be tagged as `internal:`.  For example:
 
-* "Internal: Refactored specs for better testing"
-* "Default timeout configuration changed from 10s to 5s"
+[source,markdown]
+ - internal: Refactored specs for better testing
+ - config: Default timeout configuration changed from 10s to 5s
+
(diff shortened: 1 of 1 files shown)


docs/static/maintainer-guide.asciidoc
```diff
@@ -88,10 +88,40 @@ should kept to the minimum.
...
+ - tag: Multi-line description        // <3> <6>
+   must be indented and can use
+   additional markdown syntax
+                                      // <7>
+## 1.0.0                              // <8>
+[...]
+
+----
+<1> Latest version is the first line of CHANGELOG.md
+<2> Each version identifier should be a level-1 section title using `##`
```
ph: I though `#` was header level 1 and `##` was header level 2?

ph:
@elastic/logstash Hey guys, what do you think of this?

docs/static/maintainer-guide.asciidoc
wiibaa: My last read was asciidoctor where they separate document title from section title http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#section-titles

But in markdown I should indeed say **header level-2**

suyograo:
@wiibaa Good suggestion, but what about having just 2 tags: `internal` and `docs`. I want to make sure creating CHANGELOG is not a lot of overhead. WDYT?

wiibaa:
@suyograo As discussed with @ph in #5081 an additional goal could be to simplify (or automate) the creation of the [main logstash changelog](https://github.com/elastic/logstash/blob/2.3/CHANGELOG.md)

In recent versions, we can see how it is carefully crafted, so the final presentation of info in this changelog is the driver for the usage of tags upstream. So you decide what you need first, but *I see a pattern :)*. Please see table below

Logstash changelog for humans | tag in plugins changelog
--- | ---
<ul><li>**Added**...</li><li>**Improved**...</li><li>**Better**...</li> <li>**Support**...</li></ul> | `improvment`
<ul><li>**Fixed**...</li><li>**Fixed bug**...</li><li>**Resolved**...</li></ul> | `bugfix`
<ul><li>**Breaking change**...</li><li>**Backward incompatible**...</li></ul> | `config,breaking`

IMHO I do not see a big overhead using tags in the changelog, because they should be derived from the labels used for triage on the github issues. The matter here is to define a sane starting point that bring clear added value.

Thinking about the main changelog and giving my 2 cents, I would be tempted to order and group the changes in:
* **config change first** new / deprecated / breaking
* bugfix and improvment (improvment with config change already covered in 1st section)
* internal

Thoughts ?

suyograo:
@wiibaa Thanks for your explanation. LGTM

PR #5104, opened 2016-04-13
Correct syntax in the conditional mutate example

endemics:
The example as an incorrect syntax. As per https://www.elastic.co/guide/en/logstash/current/plugins-filters-mutate.html#plugins-filters-mutate-remove_field it should be `remove_field` and not `remove`

### DIFF
--- a/docs/static/configuration.asciidoc
+++ b/docs/static/configuration.asciidoc
@@ -387,7 +387,7 @@ For example, the following conditional uses the mutate filter to remove the fiel
 ----------------------------------
 filter {
   if [action] == "login" {
-    mutate { remove => "secret" }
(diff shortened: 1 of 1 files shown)


endemics:
I've signed the CLA afterward

suyograo:
@endemics you signed the CLA with `3ntr0p13+github@gmail.com` but this patch is from a different email: `https://github.com/elastic/logstash/pull/5104.patch`. Can you create a PR from the same email address?

endemics:
@suyograo OK I didn't realise the email needed to match, I thought the key was the github user. I've signed the CLA again with the email from this patch, so hopefully this time it should be all good.

PR #5107, opened 2016-04-13
Make sure to use non existence plugins in registry test

purbon:
This helps fixing one of the issues having in CI when a branch already had a release as in that case the elasticsearch input plugin will be there for sure.

For other issues in https://logstash-ci.elastic.co/job/elastic+logstash+5.0+multijob-intake/, we should make sure to release a new snapshot, until this does not happen the build is going to be red ( due to the way test works ). Waiting for something like #5063 to be merged and then we could run test with not released code.

// @suyograo


suyograo:
LGTM

ph:
The change LGTM

PR #5113, opened 2016-04-13
Fix backward compatibility break with fieldref and Float/Fixnum

### LINKED ISSUE #4961: Issue with logstash 2.3 and coordinates returned by geoip filter
I've run into an issue using the geoip filter in logstash 2.3 where the plugin is returning valid coordinates after looking up an IP address, but when logstash tries to send the data to elastic the coordinates have been changed to an invalid number.  I added some additional logging to the geoip filter so I could see what was getting sent back to logstash.

When this happens, the invalid latitude that logstash tries to send to elastic is the same across all events until I restart logstash.  After a restart the latitude changes to a new invalid value.

So this is getting returned to logstash from the geoip filter:

```
{:level=>:error, "ip"=>"REDACTED", "country_code2"=>"US", "country_code3"=>"USA", "country_name"=>"United States", "continent_code"=>"NA", "region_name"=>"MA", "city_name"=>"Cambridge", "postal_code"=>"02142", "latitude"=>42.362599999999986, "longitude"=>-71.0843, "dma_code"=>506, "area_code"=>617, "timezone"=>"America/New_York", "real_region_name"=>"Massachusetts", "location"=>[-71.0843, 42.362599999999986]}`
```

And then this is what happens when logstash tries to send to elastic:

```
Failed action.  {:status=>400, :action=>["index", {:_id=>nil, :_index=>"firewall-2016.03.30", :_type=>"syslog", :_routing=>nil}, 2016-03-30T20:24:01.940Z 10.1.0.9 <182>Mar 30 2016 13:24:01: %ASA-6-302013: Built outbound TCP connection 2218122914 for outside:REDACTED/80 (REDACTED/80) to inside:REDACTED/7180 (REDACTED/35875)
], :response=>{"create"=>{"_index"=>"firewall-2016.03.30", "_type"=>"syslog", "_id"=>"AVPJMpNujqG9ypkgPRCU", "status"=>400, "error"=>{"type"=>"mapper_parsing_exception", "reason"=>"failed to parse", "caused_by"=>{"type"=>"illegal_argument_exception", "reason"=>"illegal latitude value [269.82439041137695] for src_ip_coords"}}}}, :level=>:warn}
```

I can't for the life of me figure out where the invalid latitude is coming from.  I've removed actual IP addresses to protect the innocent, but the src_ip field contains a valid public IP address that is successfully filtered by geoip in logstash 2.2.1 :)

The same data and logstash configuration works fine in 2.2.1, this is something new that presented with 2.3

ph:
The change introduced with https://github.com/elastic/logstash/issues/4592 to deal with reference had
a side effect to break existing configuration when using fieldref accessing a Fixnum or a Float.

The problem was instead of returning the original type , the value
was converted to a string. This string could break some mappings that
were expecting a float.

Fixes #4961

### DIFF
--- a/logstash-core-event/lib/logstash/string_interpolation.rb
+++ b/logstash-core-event/lib/logstash/string_interpolation.rb
@@ -124,9 +124,15 @@ module LogStash
         value.join(",")
       when Hash
         LogStash::Json.dump(value)
+      when Numeric
+        value
+      when TrueClass
+        value
+      when FalseClass
+        value
       else
-        # Make sure we dont work on the refence of the value
-        # The Java Event implementation was always returning a string.
+        # If we dont know how to deal with the type we return a string
+        # to make sure we don't return a reference.
         "#{value}"
       end
     end
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -82,12 +82,40 @@ describe LogStash::Event do
         expect(event["reference_test"]).not_to eq(data)
       end
 
-      it "should not return a Fixnum reference" do
+      # TODO: This was a bug and should only be true in the context of 2.3.X
+      # see https://github.com/elastic/logstash/issues/5114 for more details.
+      it "should return a Fixnum" do
         data = 1
         event = LogStash::Event.new({ "reference" => data })
         LogStash::Util::Decorators.add_fields({"reference_test" => "%{reference}"}, event, "dummy-plugin")
-        data += 41
-        expect(event["reference_test"]).to eq("1")
+        expect(event["reference_test"]).to eq(1)
+      end
+
+      # TODO: This was a bug and should only be true in the context of 2.3.X
+      # see https://github.com/elastic/logstash/issues/5114 for more details.
+      it "should return a Float" do
+        data = 1.999
+        event = LogStash::Event.new({ "reference" => data })
+        LogStash::Util::Decorators.add_fields({"reference_test" => "%{reference}"}, event, "dummy-plugin")
+        expect(event["reference_test"]).to eq(1.999)
+      end
+
+      # TODO: This was a bug and should only be true in the context of 2.3.X
+      # see https://github.com/elastic/logstash/issues/5114 for more details.
+      it "should return true" do
+        data = true
+        event = LogStash::Event.new({ "reference" => data })
+        LogStash::Util::Decorators.add_fields({"reference_test" => "%{reference}"}, event, "dummy-plugin")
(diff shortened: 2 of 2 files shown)


logstash-core-event/lib/logstash/string_interpolation.rb
```diff
@@ -124,9 +124,13 @@ def evaluate(event)
         value.join(",")
       when Hash
         LogStash::Json.dump(value)
+      when Fixnum
```
jsvd: I suggest `when Numeric` instead of the 2 when clauses

ph: Didnt know about that.

logstash-core-event/lib/logstash/string_interpolation.rb
```diff
@@ -124,9 +124,11 @@ def evaluate(event)
         value.join(",")
       when Hash
         LogStash::Json.dump(value)
+      when Numeric
+        value
       else
-        # Make sure we dont work on the refence of the value
-        # The Java Event implementation was always returning a string.
+        # If we dont know how to deal with the type we return a string
+        # to make sure we don't return a reference.
```
jsvd: should we also case TrueClass/FalseClass?

ph: If you want to be full backward compatible yes, we should we didn't have any test before covering the boolean cases.

colinsurprenant:
Just to make sure, did we test the previous versions handling of the booleans TrueClass/FalseClass to make sure we are doing the same thing?

ph:
@colinsurprenant Yes the booleans where returned directly and were not converted to string.

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -82,12 +82,32 @@
         expect(event["reference_test"]).not_to eq(data)
       end
 
-      it "should not return a Fixnum reference" do
+      it "should return a Fixnum" do
```
jordansissel: Let's put these in a context for things that should only work in 2.3 and older as a way of documenting for ourselves that these should be removed (and also fail) on 5.0

jordansissel:
Other than my one comment, LGTM

logstash-core-event/spec/logstash/event_spec.rb
ph: Ill add a comment on this test for that with a reference to the other issues.

colinsurprenant:
Perfect. LGTM.

suyograo:
here's one more LGTM

ph:
Changes done waiting on the jenkins run.

PR #5125, opened 2016-04-15
Normalize the name of the generated artifacts

### LINKED ISSUE #5100: Normalize the naming of the logstash generated artifacts with the other products
The artifacts produced by the build process don't use the same naming when generating a pre-release package (alpha or beta). Some of the file contains a tidle or a hyphen. I think we should standardize on the hyphen like other ES products like beats.  This will make sure that our download statistic are correctly compiled.

```
logstash_5.0.0~alpha1-1_all.deb
logstash-5.0.0-alpha1.tar.gz
logstash-5.0.0-alpha1.zip
logstash-5.0.0~alpha1-1.noarch.rpm
```

We need to take care of the following, the debian and the rpm still need to use the `~` in their metadata to make sure that alpha1 is lower than GA.

cc @drewr @dedemorton

ph:
This PR introduce a unified naming scheme for all our artifacts and will
produce the following files:

logstash-5.0.0-alpha1_all.deb
logstash-5.0.0-alpha1.tar.gz
logstash-5.0.0-alpha1.zip
logstash-5.0.0-alpha1.noarch.rpm

Fix #5100

### DIFF
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -226,6 +226,9 @@ namespace "artifact" do
 
     case platform
       when "redhat", "centos"
+        # produce: logstash-5.0.0-alpha1.noarch.rpm
+        package_filename = "logstash-#{LOGSTASH_VERSION}.ARCH.TYPE"
+
         File.join(basedir, "pkg", "logrotate.conf").tap do |path|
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

jsvd:
LGTM

PR #5132, opened 2016-04-18
fix for accessors lut invalidation

### LINKED ISSUE #5127: Accessors path invalidation problem
This problem was originally reported by @guyboertje in #5212. I am copying the invalidation bug sequence here:

```sh
> event = LogStash::Event.new
=> #<LogStash::Event:0x25ce435 @metadata_accessors=#<LogStash::Util::Accessors:0x7ea71fc2 @store={}, @lut={}>, @cancelled=false, @data={"@version"=>"1", "@timestamp"=>"2016-04-15T14:19:31.561Z"}, @metadata={}, @accessors=#<LogStash::Util::Accessors:0x7cd5fcf4 @store={"@version"=>"1", "@timestamp"=>"2016-04-15T14:19:31.561Z"}, @lut={}>>
> event["[a][0]"]
=> nil
> event["[a][0]"] = 42
=> 42
> event["[a]"]
=> {"0"=>42}
> event["[a]"] = [42,24]
=> [42, 24]
> event["[a]"]
=> [42, 24]
> event["[a][0]"]
=> 42
> event["[a]"] = [24,42]
=> [24, 42]
> event["[a][0]"]
=> 42
> event["[a][0]"] = {"a"=> 99, "b" => 98}
=> {"a"=>99, "b"=>98}
> event["[a][0]"]
=> {"a"=>99, "b"=>98}
> event["[a]"]
=> [24, 42]
> event["[a][0]"]
=> {"a"=>99, "b"=>98}
> event["[a][0][b]"]
TypeError: can't convert String into Integer
  from org/jruby/RubyFixnum.java:1143:in `[]'
```

I wrote a spec with this sequence and confirm the problem on **both** the Ruby and Java Event implementations.

colinsurprenant:
This is a fix for the Accessors fieldref cache invalidation in the Ruby Event only. Before considering merging this we should assess the performance impact and see if using a hierarchical cache would provide better performance. Depending on the direction we choose for the fix, the same idea will have to be applied to the Java Event.

A stale cache spec has been added. 

Fixes #5127 

**[edit 04/20]**
Per comments below, we agreed that this fix is acceptable performance-wise and the same fix has been applied to the Java Event. Doing so uncovered another bug that resulted in a `expecting List or Map ` ClassCastException, also fixed and related test added.

### DIFF
--- a/logstash-core-event-java/src/main/java/com/logstash/Accessors.java
+++ b/logstash-core-event-java/src/main/java/com/logstash/Accessors.java
@@ -39,7 +39,7 @@ public class Accessors {
                 }
                 return ((List<Object>) target).remove(i);
             } else {
-                throw new ClassCastException("expecting List or Map");
+                throw newCollectionException(target);
             }
         }
         return null;
@@ -67,7 +67,7 @@ public class Accessors {
         target = this.data;
         for (String key : field.getPath()) {
             target = fetch(target, key);
-            if (target == null) {
+            if (! isCollection(target)) {
                 return null;
             }
         }
@@ -80,9 +80,13 @@ public class Accessors {
     private Object findCreateTarget(FieldReference field) {
         Object target;
 
-        if ((target = this.lut.get(field.getReference())) != null) {
-            return target;
-        }
+        // flush the @lut to prevent stale cached fieldref which may point to an old target
+        // which was overwritten with a new value. for example, if "[a][b]" is cached and we
+        // set a new value for "[a]" then reading again "[a][b]" would point in a stale target.
+        // flushing the complete @lut is suboptimal, but a hierarchical lut would be required
+        // to be able to invalidate fieldrefs from a common root.
+        // see https://github.com/elastic/logstash/pull/5132
+        this.lut.clear();
 
         target = this.data;
         for (String key : field.getPath()) {
@@ -95,10 +99,8 @@ public class Accessors {
                     int i = Integer.parseInt(key);
                     // TODO: what about index out of bound?
                     ((List<Object>)target).set(i, result);
-                } else if (target == null) {
-                    // do nothing
-                } else {
-                    throw new ClassCastException("expecting List or Map");
+                } else if (target != null) {
+                    throw newCollectionException(target);
                 }
             }
             target = result;
@@ -133,8 +135,8 @@ public class Accessors {
             return result;
         } else if (target == null) {
             return null;
-        } {
-            throw new ClassCastException("expecting List or Map");
+        } else {
+            throw newCollectionException(target);
         }
     }
 
@@ -157,8 +159,19 @@ public class Accessors {
                 ((List<Object>) target).set(i, value);
             }
         } else {
-            throw new ClassCastException("expecting List or Map");
+            throw newCollectionException(target);
         }
         return value;
+    }
+
+    private boolean isCollection(Object target) {
+        if (target == null) {
+            return false;
+        }
+        return (target instanceof Map || target instanceof List);
+    }
+
+    private ClassCastException newCollectionException(Object target) {
+        return new ClassCastException("expecting List or Map, found "  + target.getClass());
     }
 }
--- a/logstash-core-event/spec/logstash/event_spec.rb
+++ b/logstash-core-event/spec/logstash/event_spec.rb
@@ -602,4 +602,33 @@ describe LogStash::Event do
       expect(event1.to_s).to eq("#{timestamp.to_iso8601} #{event1["host"]} #{event1["message"]}")
     end
   end
+
+  describe "Event accessors" do
+    let(:event) { LogStash::Event.new({ "message" => "foo" }) }
+
+    it "should invalidate target caching" do
+      expect(event["[a][0]"]).to be_nil
+
+      expect(event["[a][0]"] = 42).to eq(42)
+      expect(event["[a][0]"]).to eq(42)
+      expect(event["[a]"]).to eq({"0" => 42})
+
+      expect(event["[a]"] = [42, 24]).to eq([42, 24])
+      expect(event["[a]"]).to eq([42, 24])
+
+      expect(event["[a][0]"]).to eq(42)
+
+      expect(event["[a]"] = [24, 42]).to eq([24, 42])
+      expect(event["[a][0]"]).to eq(24)
+
+      expect(event["[a][0]"] = {"a "=> 99, "b" => 98}).to eq({"a "=> 99, "b" => 98})
+      expect(event["[a][0]"]).to eq({"a "=> 99, "b" => 98})
+
+      expect(event["[a]"]).to eq([{"a "=> 99, "b" => 98}, 42])
+      expect(event["[a][0]"]).to eq({"a "=> 99, "b" => 98})
+      expect(event["[a][1]"]).to eq(42)
+      expect(event["[a][0][b]"]).to eq(98)
+    end
+  end
 end
+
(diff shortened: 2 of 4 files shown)


colinsurprenant:
Ah yes, obviously, you have to make sure `logstash-core-event` is selected in the `Gemfile` and `logstash-core.gemspec`

jsvd:
From my benchmarks of operations on LogStash::Event, the performance change with this patch is as follows:
* in a set-heavy benchmark, this patch is ~7% __faster__ than the current implementation
* in a read-heavy benchmark, this patch is ~4.8%  __slower__ than current implementation
* in a mixed benchmark, this patch is ~4.5% __faster__ than the current implementation

All these are measured with an uncertainty around +/- 4%

So IMO, since this actually fixes a bug, I think the PR is on the right track 👍

colinsurprenant:
@jsvd thanks for measuring this!

colinsurprenant:
- We agreed that this fix is acceptable performance-wise and the same fix has been applied to the Java Event.
- Doing so uncovered another bug that resulted in a `expecting List or Map ` ClassCastException, also fixed and related test added.

Note that `rake test:core` reports the error under both Event implementations if/when removing the fix. The Java implementation also includes specific Java unit tests.

@jsvd @guyboertje @tal please review?

colinsurprenant:
The build failure is not related to this PR
```
Failures:

  1) LogStash::Agent metrics after config reloading resets the metric collector
     Failure/Error: expect(snapshot.metric_store.get_with_path("/stats/events")[:stats][:events][:in].value).to eq(new_config_generator_counter)
     NoMethodError:
       undefined method `value' for nil:NilClass
     # ./logstash-core/spec/logstash/agent_spec.rb:374:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:248:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:243:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:180:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:173:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:150:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:115:in `run_with_threads'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:100:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:78:in `run'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:176:in `standard_exception_handling'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:75:in `run'

Finished in 38.1 seconds (files took 2.1 seconds to load)
1179 examples, 1 failure

Failed examples:

rspec ./logstash-core/spec/logstash/agent_spec.rb:359 # LogStash::Agent metrics after config reloading resets the metric collector

Randomized with seed 54320
```
@ph @purbon  ?

ph:
@colinsurprenant I will look into this failling test. first time I see it.

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -602,4 +602,33 @@
       expect(event1.to_s).to eq("#{timestamp.to_iso8601} #{event1["host"]} #{event1["message"]}")
     end
   end
+
+  describe "Accessors invalidation" do
+    let(:event) { LogStash::Event.new({ "message" => "foo" }) }
+
+    it "should perperly invalidate accessors" do
```
guyboertje: properly

guyboertje:
LGTM

logstash-core-event-java/src/main/java/com/logstash/Accessors.java
```diff
@@ -67,7 +67,7 @@ private Object findTarget(FieldReference field) {
         target = this.data;
         for (String key : field.getPath()) {
             target = fetch(target, key);
-            if (target == null) {
+            if (! targetIsCollection(target)) {
```
jsvd: this method only checks that the argument is a collection, I suggest renaming it as `isCollection`

logstash-core-event-java/src/main/java/com/logstash/Accessors.java
```diff
@@ -134,7 +138,7 @@ private Object fetch(Object target, String key) {
         } else if (target == null) {
             return null;
         } {
-            throw new ClassCastException("expecting List or Map");
+            throw new ClassCastException("expecting List or Map, found " + target.getClass().toString());
```
jsvd: for consistency sake I suggest either putting or not having a comma on these log messages

jsvd:
2 minor comments, otherwise tests went well, LGTM

logstash-core-event-java/src/main/java/com/logstash/Accessors.java
colinsurprenant: I was totally expecting this comment :D

logstash-core-event-java/src/main/java/com/logstash/Accessors.java
colinsurprenant: +1

logstash-core-event/spec/logstash/event_spec.rb
colinsurprenant: 👍

logstash-core-event-java/src/main/java/com/logstash/Accessors.java
```diff
@@ -134,7 +138,7 @@ private Object fetch(Object target, String key) {
         } else if (target == null) {
             return null;
         } {
```
colinsurprenant: woa, this is a weird syntax quirk

talevy: missing else

logstash-core-event-java/src/main/java/com/logstash/Accessors.java
```diff
@@ -98,7 +102,7 @@ private Object findCreateTarget(FieldReference field) {
                 } else if (target == null) {
                     // do nothing
                 } else {
```
talevy: make this `else if (target != null)` and remove above empty block?

colinsurprenant: +1

logstash-core-event-java/src/main/java/com/logstash/Accessors.java
colinsurprenant: yup... weird omission, it was still "correct" but ... anyhow. fixing :D

colinsurprenant:
ok, pushed changes. any final review @jsvd @talevy ?

jsvd:
LGTM

colinsurprenant:
merged in master, 5.0, 2.x. will wait after 2.3.2 release (today) to merge fix in 2.3, we prefer to get a bit more mileage with this fix before it makes it into a release.

Issue #5070, opened 2016-04-08
Artifacts acceptance testing stories


purbon:
The motivation of this issue is part of a collection where we're describing the different the set of tests a Logstash deliverable should be put into in order to give green light for it’s release.

Each deliverable package should be by put into this set of test to get the green light to be distributed:

### Artifact management

* It should be installable on each machine.
* It should be able to upgrade without issues (from older to newer version).
* It should be able to remove the package.

### Artifact execution

* Have a working plugin manager but executing the same set of test defined in #5020 

_using a dummy config generator, stdout, no filter_

* It should be able to start
* It should be able to stop
  * if KILL_ON_STOP_TIMEOUT is defined the process should be killed ( only 1 logstash should be alive)
* It should be able to restart
   * if KILL_ON_STOP_TIMEOUT is defined the process should be killed ( only 1 logstash should be alive)
* It should log stdout to their log
* should be able to run configtest

_using a more complex config: generator, stdout, elasticsearch, no filter_

* It should be able to start
* It should be able to stop
  * if KILL_ON_STOP_TIMEOUT is defined the process should be killed ( only 1 logstash should be alive)
* It should be able to restart 
   * if KILL_ON_STOP_TIMEOUT is defined the process should be killed ( only 1 logstash should be alive) **(also when the elasticsearch is down)**

related issues: #5013 #5020

Issue #5013, opened 2016-04-05
[meta] Improvements in current Logstash testing infrastructure 


purbon:
The motivation of this issue is to track and discuss the logstash efforts to improvements our current testing efforts in regards of unit, integration and acceptance testing.

* Test should have proper organization, so it's easy for developers to know they should fulfil all necessary gaps before thinking the feature is testing complete.
  - Coverage can help knowing there are missing parts.

* Test should be able to run locally and remotely. Developers should not have to rely on the CI infra, but they should be able to either run vs a local VM (for example to debug issues in specific distros), or vs their local environment (properly sandboxed). 

* Invoking each kind of test should be a one command call, including acceptance and integration test.

* A set of integrations test, using common configuration should be created. This should be part of the logstash-core project and don't necessary have to use the logstash command interface, just create internal pipelines. This must be our [white box testing](https://en.wikipedia.org/wiki/White-box_testing#Advantages) efforts, and must be mostly focused on testing pipeline interaction with plugins. Some of this is already available inside our testing code base, but should be more clearly DRY'ed.

* Another set of integration test should be created where the integration is tested from the user point of view, this will include external components like elasticsearch, kafka, etc. In correlation with the acceptance test layer this will benefit us checking how LS is behaving. This will necessarily need a one command that deals with this external components, providing users of this test with simplicity. There has been a PoC experiment by @talevy at https://github.com/talevy/elastic-stack-tests, we shuold also leverage existing cross platform efforts.

* We should have complete acceptance test that run regularly (each commit/nightly/...). Acceptance test are basically automatic validations for our deliverables from the user point of view, so they should check for common user actions and behaviour. Related  #4868 #2852 .

The initial plan to get this up and running would be.

- [ ] Organize internal testing code base as clean as possible so developers know which testing parts they should always fill.

- [ ]  Provide internal helpers to facilitate the creation of internal pipelines and later on the retrieval of test expectations out of them.

- [ ]  Create a [first iteration](https://en.wikipedia.org/wiki/Minimum_viable_product) of an acceptance testing layer, this will include testing:
   - [ ]  Plugin manager behaviour, including CLI flags.
   - [ ]  Standard apache config.
   - [ ]  `bin/logstash` CLI flags.
   - [ ]  Package installation and upgrade behaviours.
   
 All this should be tested in all supported OS, and by using the generated packages from our codebase (aka snapshots).

[...]

PR #5138, opened 2016-04-19
[WiP] First version of QA acceptance test framework

purbon:
This is the first version, still work in progress, PR to add a complete acceptance test framework to logstash. The code in here is submitted for early review for interested parties, so is still subject to changes when necessary.

The basic functionality of this test are:

* A command executor responsible of sending commands to either a remote machine or to the local environment. This is used during the test execution to process the necessary test steps.
* A collection of rspec helpers that enable a simple and easy description of each test case.
* A collection of public vagrant machines used to run the acceptance test.

To run this you should:

* Cd into the `qa/acceptance` directory.
* Run `vagrant up`  to bootstrap all described machines, you can also bootstrap only an small set of them.
* Cd to the main `qa` directory.
* Run `rake test:ssh_config`: This will generate the current ssh configuration used for the test to locate the machines, look for a file named `.ssh_config` who will have the json information.
* Run `bundle exec rake test:acceptance:all` to run the acceptance test for all described platforms, you can also run  `bundle exec rake test:acceptance:debian` or  `bundle exec rake test:acceptance:centos` to target only one platform.

Hopefully this will report properly without issues.

I would like to have prefered feedback on:

* The rspec helpers workflow and syntax, is this too boilerplate? do you have something missing?
* Some other questions you might have.

This PR content might change, for example around the syntax or the way the rake task description works, all to make it simpler.

Looking forward to get your feedback!

related issue: #5013 #5070

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,8 @@ rspec.xml
 .vendor
 integration_run
 .mvn/
+qa/.vm_ssh_config
+qa/.vagrant
+qa/.rspec
+qa/acceptance/.vagrant
+qa/Gemfile.lock
--- a/qa/Rakefile
+++ b/qa/Rakefile
@@ -1,0 +1,53 @@
+require "rspec"
+require "rspec/core/runner"
+require "rspec/core/rake_task"
+require_relative "vagrant-helpers"
+
+task :spec    => 'spec:all'
+task :default => :spec
+
+namespace :test do
+
+  desc "Generate a valid ssh-config"
+  task :ssh_config do
+    require "json"
+    cd "acceptance" do
+      raw_ssh_config    = LogStash::VagrantHelpers.fetch_config[:stdout].split("\n");
+      parsed_ssh_config = LogStash::VagrantHelpers.parse(raw_ssh_config)
+      File.write("../.vm_ssh_config", parsed_ssh_config.to_json)
+    end
+  end
+
+  desc "Bootstrap all the VM's used for this tests"
+  task "setup" do
+    puts "bootstraping all VM's defined in acceptance/Vagrantfile"
+    cd "acceptance" do
+      LogStash::VagrantHelpers.bootstrap
+    end
+  end
+
+  namespace :acceptance do
+
+    desc "Run all acceptance"
+    task :all do
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+    end
+
+    desc "Run acceptance test in debian machines"
+    task :debian do
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/debian/**/*_spec.rb"]]))
+    end
+
+    desc "Run acceptance test in centos machines"
+    task :centos do
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/centos/**/*_spec.rb"]]))
+    end
+
+    desc "Run one single machine acceptance test"
+    task :single, :machine do  |t, args|
+      ENV['LS_VAGRANT_HOST'] = args[:machine]
+      platform = LogStash::VagrantHelpers.translate(args[:machine])
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+    end
+  end
+end
--- a/qa/acceptance/Vagrantfile
+++ b/qa/acceptance/Vagrantfile
@@ -1,0 +1,43 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+
+  config.vm.define "ubuntu-1204" do |machine|
+    machine.vm.box = "elastic/ubuntu-12.04-x86_64"
+    common_config(machine, "ubuntu")
+  end
+
+  config.vm.define "ubuntu-1404" do |machine|
+    machine.vm.box = "elastic/ubuntu-14.04-x86_64"
+    common_config(machine, "ubuntu")
+  end
+
+  config.vm.define "centos-6" do |machine|
+    machine.vm.box = "elastic/centos-6-x86_64"
+    common_config(machine, "ubuntu")
+  end
+
+  config.vm.define "centos-7" do |machine|
+    machine.vm.box = "elastic/centos-7-x86_64"
+    common_config(machine, "ubuntu")
+  end
+
+end
+
+def common_config(machine, type)
+  machine.vm.provider "virtualbox" do |v|
+    v.memory = 4096
+    v.cpus = 4
+  end
+  machine.vm.synced_folder "../../build", "/logstash-build", create: true
+  machine.vm.provision :shell do |sh|
+    sh.path = "sys/#{type}/bootstrap.sh"
+    sh.privileged = true
+  end
+
+  machine.vm.provision :shell do |sh|
+    sh.path = "sys/#{type}/user_bootstrap.sh"
+    sh.privileged = false
+  end
+end
--- a/qa/acceptance/spec/centos/lib/install_spec.rb
+++ b/qa/acceptance/spec/centos/lib/install_spec.rb
@@ -1,0 +1,33 @@
+# encoding: utf-8
+require_relative '../spec_helper'
+require          'logstash/version'
+
+describe "artifacts", :platform => :centos do
+
+  shared_examples "installable" do |host, name|
+
+    before(:each) do
+      install("/logstash-build/logstash-#{LOGSTASH_VERSION}.noarch.rpm", host)
+    end
+
+    it "is installed on #{name}" do
+      expect("logstash").to be_installed.on(host)
+    end
+
+    it "is running in #{name}" do
+      start_service("logstash", host)
+      expect("logstash").to be_running.on(host)
+      stop_service("logstash", host)
+    end
+
+    it "is removable on #{name}" do
+      uninstall("logstash", host)
+      expect("logstash").to be_removed.on(host)
+    end
+  end
+
+  config = ServiceTester.configuration
+  config.servers.each do |host|
+    it_behaves_like "installable", host, config.lookup[host]
+  end
+end
--- a/qa/acceptance/spec/centos/spec_helper.rb
+++ b/qa/acceptance/spec/centos/spec_helper.rb
@@ -1,0 +1,15 @@
+# encoding: utf-8
+require_relative '../spec_helper'
+
+def default_vagrant_boxes
+  [
+    "centos-6",
+    "centos-7"
+  ]
+end
+
+selected_boxes = SpecsHelper.find_selected_boxes(default_vagrant_boxes)
+
+SpecsHelper.configure(selected_boxes)
+
+puts "[Acceptance specs] running on #{ServiceTester.configuration.lookup.values}" if !selected_boxes.empty?
--- a/qa/acceptance/spec/config_helper.rb
+++ b/qa/acceptance/spec/config_helper.rb
@@ -1,0 +1,28 @@
+# encoding: utf-8
+require "json"
+
+module SpecsHelper
+
+  def self.find_selected_boxes(default_boxes=[])
+    if ENV.include?('LS_VAGRANT_HOST') then
+      default_boxes.include?(ENV['LS_VAGRANT_HOST']) ? ENV['LS_VAGRANT_HOST'] : []
+    else
+      default_boxes
+    end
+  end
+
+  def self.configure(vagrant_boxes)
+    setup_config = JSON.parse(File.read(".vm_ssh_config"))
+    ServiceTester.configure do |config|
+      config.servers = []
+      config.lookup  = {}
+      setup_config.each do |host_info|
+        next unless vagrant_boxes.include?(host_info["host"])
+        url = "#{host_info["hostname"]}:#{host_info["port"]}"
+        config.servers << url
+        config.lookup[url] = host_info["host"]
+      end
+    end
+  end
+
+end
--- a/qa/acceptance/spec/debian/spec_helper.rb
+++ b/qa/acceptance/spec/debian/spec_helper.rb
@@ -1,0 +1,15 @@
+# encoding: utf-8
+require_relative '../spec_helper'
+
+def default_vagrant_boxes
+  [
+    "ubuntu-1204",
+    "ubuntu-1404"
+  ]
+end
+
+selected_boxes = SpecsHelper.find_selected_boxes(default_vagrant_boxes)
+
+SpecsHelper.configure(selected_boxes)
+
+puts "[Acceptance specs] running on #{ServiceTester.configuration.lookup.values}" if !selected_boxes.empty?
--- a/qa/acceptance/sys/centos/user_bootstrap.sh
+++ b/qa/acceptance/sys/centos/user_bootstrap.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+cd
+wget --no-verbose https://download.elastic.co/logstash/logstash/packages/centos/logstash-2.3.1-1.noarch.rpm
--- a/qa/acceptance/sys/ubuntu/bootstrap.sh
+++ b/qa/acceptance/sys/ubuntu/bootstrap.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+apt-get update
+apt-get install -y openjdk-7-jdk
+
+##
+# Install logstash manually from a URL
+##
+BRANCH=${LOGSTASH_BRANCH:-'master'}
+BUILD_URL='https://s3-eu-west-1.amazonaws.com/build-eu.elasticsearch.org/logstash'
+URL="$BUILD_URL/$BRANCH/nightly/JDK8/logstash-latest-SNAPSHOT.deb"
+wget --no-verbose $URL
--- a/qa/acceptance/sys/ubuntu/user_bootstrap.sh
+++ b/qa/acceptance/sys/ubuntu/user_bootstrap.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+cd
+wget --no-verbose https://download.elastic.co/logstash/logstash/packages/debian/logstash_2.3.1-1_all.deb
--- a/qa/rspec/centos/commands.rb
+++ b/qa/rspec/centos/commands.rb
@@ -1,0 +1,58 @@
+# encoding: utf-8
+module ServiceTester
+  class CentosCommands
+
+    def installed?(hosts, package)
+      stdout = ""
+      at(hosts, {in: :serial}) do |host|
+        cmd = exec!("yum list installed  #{package}")
+        stdout = cmd.stdout
+      end
+      stdout.match(/^Installed Packages$/)
+      stdout.match(/^logstash.noarch/)
+    end
+
+    def install(package, host=nil)
+      hosts  = (host.nil? ? servers : Array(host))
+      errors = {}
+      at(hosts, {in: :serial}) do |_host|
+        cmd = sudo_exec!("yum install -y  #{package}")
+        errors[_host] = cmd.stderr unless cmd.stderr.empty?
+      end
+      errors
+    end
+
+    def uninstall(package, host=nil)
+      hosts = (host.nil? ? servers : Array(host))
+      at(hosts, {in: :serial}) do |_|
+        sudo_exec!("yum remove -y #{package}")
+      end
+    end
+
+    def removed?(hosts, package)
+      stdout = ""
+      at(hosts, {in: :serial}) do |host|
+        cmd = sudo_exec!("yum list installed #{package}")
+        stdout = cmd.stderr
+      end
+      stdout.match(/^Error: No matching Packages to list$/)
+    end
+
+    def running?(hosts, package)
+      stdout = ""
+      at(hosts, {in: :serial}) do |host|
+        cmd = sudo_exec!("service #{package} status")
+        stdout = cmd.stdout
+      end
+      stdout.match(/^#{package} is running$/)
+    end
+
+    def service_manager(service, action, host=nil)
+      hosts = (host.nil? ? servers : Array(host))
+      at(hosts, {in: :serial}) do |host|
+        sudo_exec!("service #{service} #{action}")
+      end
+    end
+
+  end
+end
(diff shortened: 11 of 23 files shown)


purbon:
might be you @ph or @suyograo wanna share some of your thoughts about it?

ph:
General comment, can you add the `encoding: utf-8` to all the files.

qa/acceptance/sys/ubuntu/user_bootstrap.sh
```diff
@@ -0,0 +1,31 @@
...
+##
+#rbenv install jruby-1.7.24
+#rbenv global  jruby-1.7.24
+
+#cd logstash
+#rake bootstrap
+#rake test:install-core
+
+cd
+wget https://download.elastic.co/logstash/logstash/packages/debian/logstash_2.3.1-1_all.deb
```
purbon: this is now getting the latest 2.x branch released, however it should get it from local a shared dir as much as possible. The basic idea should be that:

* Pull last released version from URL.
* Push the generated version from local shared directory.

Will make sure to have this implemented next.

What do you think?

qa/Rakefile
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    desc "Run acceptance test in centos machines"
+    task :centos do
+      require "rspec/core/runner"
+      require "rspec"
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/centos/**/*_spec.rb"]]))
+    end
+  end
+end
```
ph: Can we put all the requires a the top? I don't see the need to lazy require here.

qa/Rakefile
```diff
@@ -0,0 +1,66 @@
...
+    hosts
+  end
+
+  desc "generate a valid ssh-config"
+  task :ssh_config do
+    require "json"
+    cd "acceptance" do
+      File.write("../.ssh_config", parse(fetch_config[:stdout].split("\n")).to_json)
+    end
+  end
```
ph: Can we give it a better name? With that naming I thought is was a file generated by ssh.
`.vagrant_ssh_config` would make it clear its related to our vagrant setup.

qa/acceptance/Vagrantfile
```diff
@@ -0,0 +1,64 @@
...
+    sh.privileged = true
+  end
+
+  ls.vm.provision :shell do |sh|
+    sh.path = "sys/ubuntu/user_bootstrap.sh"
+    sh.privileged = false
+  end
+
+  #ls.vm.network "forwarded_port", guest: 9600, host: 9600
+end
```
ph: Did you consider a hash based configuration for the machine definition like this one?
I find it more concise easier to deal with, but It might be matter of taste.

```ruby
boxes = {
"ubuntu-1204" => { "box" => "elastic/ubuntu-12.04-x86_64" }
"centos-6" => { "box" => "elastic/centos-6-x86_64" }
}

boxes.each |machine_name, config|
machine_type = machine_name.gsub(/[^a-ZA-Z]/,'')
 config.vm.define machine_name do |ls|
     ls.vm.box = config["box"]
   ls.vm.provider "virtualbox" do |v|
     v.memory = 4096
     v.cpus = 4
   end
   ls.vm.synced_folder "../../build", "/logstash-build"
   ls.vm.provision :shell do |sh|
     sh.path = "sys/#{machine_type}/bootstrap.sh"
     sh.privileged = true
  end
 
   ls.vm.provision :shell do |sh|
     sh.path = "sys/{machine_type}/user_bootstrap.sh"
     sh.privileged = false
   end
end
```

I like the idea of letting vagrant decide the port/ips.

qa/acceptance/spec/centos/lib/install_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      uninstall("logstash", host)
+      expect("logstash").to be_removed.on(host)
+    end
+  end
+
+  config = ServiceTester.configuration
+  config.servers.each do |host|
+    it_behaves_like "installable", host, config.lookup[host]
+  end
+end
```
ph: I am usually in favor of explicit vs implicit, but I am wondering with all this `host` repetition.
Should we make the `host` implicit on the matcher? Our test will 99% of the time target one machine, the current one.

This would change the test to this format.
```ruby
it "is running in #{name}" do
 start_service("logstash")
 expect("logstash").to be_running
 stop_service("logstash")
end
```

qa/acceptance/spec/centos/spec_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+require_relative '../../../rspec/helpers'
+require_relative '../../../rspec/matchers'
+require_relative '../config_helper'
+
+def vagrant_boxes
+  [
+    "centos-6",
+    "centos-7"
+  ]
+end
```
ph: If we use a hash for settings up the vagrant boxes we would be able to reuse the data here.
see my comment on https://github.com/elastic/logstash/pull/5138/files#r60282516

qa/acceptance/spec/config_helper.rb
```diff
@@ -0,0 +1,19 @@
+require "json"
```
ph: I believe this is a good place to not require anything from the core.

qa/acceptance/Vagrantfile
ph: I used the name of the distribution for the `machine_type` but it could be a key in the config.

qa/Rakefile
```diff
@@ -0,0 +1,66 @@
...
+
+namespace :test do
+
+  def fetch_config
+    require "open3"
+    Open3.popen3("vagrant ssh-config") do |stdin, stdout, stderr, wait_thr|
+      { :stdout => stdout.read.chomp, :stderr => stderr.read.chomp,
+        :exit_status => wait_thr.value.exitstatus }
+    end
+  end
```
ph: Maybe add a return a message if vagrant is not installed or not available in the current $PATH?

by curiosity, since vagrant is written in ruby, do they have a gem that wrap vagrant? (Not really needed, was a simple curiosity)

qa/Rakefile
```diff
@@ -0,0 +1,66 @@
...
+      elsif line.match(/Port\s(.*)$/)
+        host[:port]     = line.gsub("Port","").strip
+      elsif line.empty?
+        hosts << host
+        host = {}
+      end
+    end
+    hosts << host
+    hosts
+  end
```
ph: sad 🐼   that vagrant doesn't return a machine readable format

qa/acceptance/sys/ubuntu/bootstrap.sh
```diff
@@ -0,0 +1,22 @@
...
+#apt-get update
+
+##
+# Install logstash manually from a URL
+##
+BRANCH=${LOGSTASH_BRANCH:-'master'}
+BUILD_URL='https://s3-eu-west-1.amazonaws.com/build-eu.elasticsearch.org/logstash'
+URL="$BUILD_URL/$BRANCH/nightly/JDK8/logstash-latest-SNAPSHOT.deb"
+wget $URL
+#dpkg -i "$HOME/logstash-latest-SNAPSHOT.deb"
```
ph: Maybe not needed with the current dependencies, but I think we should future proof it and add this line to make sure apt-get doesn't trigger a dialog.

`export DEBIAN_FRONTEND=noninteractive`

qa/acceptance/spec/debian/spec_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+end
+
+RunnerTool.configure
+SpecsHelper.configure(vagrant_boxes)
+
+puts "[Acceptance specs] running on #{ServiceTester.configuration.lookup.values}"
+
+RSpec.configure do |c|
+  c.include ServiceTester
+end
```
ph: This code is almost the same as https://github.com/elastic/logstash/pull/5138/files#diff-560541d08dfcd2f1980e7a56c37e1b3eR1 we should dry it up.

qa/acceptance/sys/centos/user_bootstrap.sh
```diff
@@ -0,0 +1,31 @@
...
+# install jruby-1.7.24 and make it global
+##
+#rbenv install jruby-1.7.24
+#rbenv global  jruby-1.7.24
+
+#cd logstash
+#rake bootstrap
+#rake test:install-core
+
+cd
```
ph: Left over?

qa/acceptance/sys/ubuntu/user_bootstrap.sh
ph: What if we keep the scripts the same, they always install a package from the shared directory.

But we add code that is run (could be a rake task) before that to make sure the package is available in the shared directory. So the script could either download the package or ask logstash to build a new one.

qa/rspec/centos/commands.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def uninstall(package, host=nil)
+      hosts = (host.nil? ? servers : Array(host))
+      at(hosts, {in: :serial}) do |_|
+        sudo_exec!("yum remove -y #{package}")
+      end
+    end
+
+
```
ph: sticky enter key?

ph:
I suggest we add a `README.md` in the qa directory and we will populate it as we go with the feature branch, we can add a link to it from the main readme file. WDYT?

.gitignore
```diff
@@ -28,3 +28,8 @@ rspec.xml
 .vendor
 integration_run
 .mvn/
+qa/.ssh_config
+qa/.vagrant
+qa/.rspec
+qa/acceptance/.vagrant
+qa/Gemfile.lock
```
ph: We might want to add `aq/vendor` to this list, in case anyone is doing `bundler install --path vendor` like myself.

qa/Rakefile
ph: I am **suggesting** that we encapsulate the methods that deal with the `ssh_config` file into a class, so we can reuse it or unit test it more easily. 

IMHO  code in the rakefile should be minimal.

qa/Rakefile
```diff
@@ -0,0 +1,66 @@
+require 'rspec'
```
ph: Can you add these lines to this file? This will make sure the code actually read the `Gemfile.lock` and the the `.bundler/config`, This is needed if you want to run `bundle install --path vendor`

```ruby
require "rubygems" # Might not be required on all ruby, but doesnt hurt.
require "bundler/setup"
```

qa/acceptance/Vagrantfile
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+end
+
+def centos_common_config(ls)
+  ls.vm.provider "virtualbox" do |v|
+    v.memory = 4096
+    v.cpus = 4
+  end
+ls.vm.synced_folder "../../build", "/logstash-build"
```
ph: I found a small issue when I ran the test, when you start vagrant the `build` folder need to exist before it succesfuly start. Since the `build` is not checked in the repository its possible that the directory wont exist when we try to start the tests.

I suggest for that we try to create it before configuring it.

qa/acceptance/sys/centos/user_bootstrap.sh
```diff
@@ -0,0 +1,31 @@
...
+wget https://download.elastic.co/logstash/logstash/packages/centos/logstash-2.3.1-1.noarch.rpm
```
ph: Can we run `wget` in  `–quiet ` mode, because it generates a weird output when provisioning the machine.

ph:
I did a a first pass on the code and tried it locally.
I was not able to make it run correctly,

I did the following:
```sh
vagrant up
rake test:ssh_config
rake test:acceptance:debian
```

@purbon @talevy should we also review [runner-tool](https://github.com/purbon/runner-tool) I would be inclined to say yes since we will all depends on it? 

**A few future improvements**

The initial `vagrant up` takes a long time, I am OK with that :)

This Code is completely independent of Logstash and it wont use the bundled jruby to run, are we OK with that strategy? I believe its fine for the first iteration.

We need some kind of `rake test:setup` make sure the system is ready to run test, I had to go through the task and fine what was needed to be run.

qa/acceptance/sys/ubuntu/user_bootstrap.sh
purbon: I think we have to keep both, one in the bootstrap script to get the "latest" release and a generated version in the shared directory.

the latest will only be used when performing the upgrade tests.

what do you think?

qa/Rakefile
purbon: Yes, I agree we should be starting this direction in all our rake tasks. For now code is kind of minimal already, but I'm sure we could simplify it more.

qa/Rakefile
purbon: it's is something generated by ssh, but it contains data to access the vm's over ssh ...., so you know ... naming is hard 💯

purbon: instead of your `.vagrant_ssh_config`, I will be using `.vm_ssh_config` as is less tool dependant and more descriptive as it relates to the vm's used for testing.

qa/acceptance/spec/centos/lib/install_spec.rb
purbon: @ph I agree with you, I find this host usage unnecessary repetitive. I actually spoke about it with @jsvd and the problem we found here is how to easy access to this host variable (param), is not easy to transparently make this option pass to the helpers level in rspec.

do you have an idea?

on the other side, he was ok with this syntax, so I guess we have many opinions here.

qa/acceptance/spec/config_helper.rb
purbon: we, no problem! I stick too much to old behaviour if using plain json parser, 👍 I should remember myself all time!

qa/acceptance/sys/ubuntu/bootstrap.sh
purbon: I understand, but no really sure we need this, this part is more of a leftover from where I pulled this scripts. however I will put this as part of the initial `.sh` script header.

purbon:
@ph I think I addressed most of your comments, thank you very much for your feedback here, all very valuable!

respect on running the test, there is a small situation I found out, as this is using the `runner-tool` defined in the Gemfile (actually pulled with `:git` option) for rake to find it I somehow need to run `rake test:acceptance:debian` as `bundle exec rake test:acceptance:debian`.

but can not for example run the tasks that interact with vagrant like this because bundler then gets confused with vagrant and the dependencies it needs, if you have any idea how to make this easier I'm very welcome!

purbon:
On Tue, Apr 19, 2016 at 9:20 PM Pier-Hugues Pellerin <

> @purbon <https://github.com/purbon> @talevy <https://github.com/talevy>
> should we also review runner-tool <https://github.com/purbon/runner-tool>
> I would be inclined to say yes since we will all depends on it?
>

sure, the tool is still work in progress, beta, alpha status, but all
comments are for sure very welcome there! collective intelligence is for
sure better than lock down :-P


> *A few future improvements*
>
> The initial vagrant up takes a long time, I am OK with that :)
>
well, we have not much of room to improve this, every time we
bootstrap/provision a new set of VM's this takes time, the idea here is to
have this in mind and use our resources there wisely.


> This Code is completely independent of Logstash and it wont use the
> bundled jruby to run, are we OK with that strategy? I believe its fine for
> the first iteration.
>
The question is why do I want to run this acceptance test with the bundled
ruby, I don't see any benefit for now on it, only problems (as for example
how to redirect it to it, must say our current switch is rather fragile).


> We need some kind of rake test:setup make sure the system is ready to run
> test, I had to go through the task and fine what was needed to be run.
>
Added an initial version of it that basically does vagrant up, but we could
for sure add more detailed tasks.

- purbon

qa/acceptance/sys/ubuntu/user_bootstrap.sh
ph: Yes we need both for the upgrade scenario.

qa/Rakefile
ph: Since its new code, lets do it now, if we don't do it we will never do it ;)

purbon: it's is already done in the current status of the PR 👍

qa/acceptance/spec/centos/lib/install_spec.rb
ph: Oh yeah, helper scoping in rspec is kinda tricky. But I I though their module got mixed in the generated class. Do they have access to `let` statements?

> he was ok with this syntax, so I guess we have many opinions here.

I find it unecessary, I haven't experimented on this PR but I think we get remove it.

ph:
> but can not for example run the tasks that interact with vagrant like this because bundler then gets confused with vagrant and the dependencies it needs, if you have any idea how to make this easier I'm very welcome!


Did you try the following `Bundler.with_clean_env`? See: http://bundler.io/v1.3/man/bundle-exec.1.html
```ruby
 def self.execute(cmd)
   Bundler.with_clean_env do
      Open3.popen3(cmd) do |stdin, stdout, stderr, wait_thr|
       { :stdout => stdout.read.chomp, :stderr => stderr.read.chomp,
        :exit_status => wait_thr.value.exitstatus }
      end
   end
end

**Comment by ph** [comment]
> The question is why do I want to run this acceptance test with the bundled
ruby, I don't see any benefit for now on it, only problems (as for example
how to redirect it to it, must say our current switch is rather fragile)

Agree its fragile.


> sure, the tool is still work in progress, beta, alpha status, but all
comments are for sure very welcome there! collective intelligence is for
sure better than lock down :-P

To be able to do that, we would need some kind of PR, its is possible for you to create that?
When its ready I think we should move it to the `elastic` organization?

File: qa/Rakefile [hunk 18]
```diff
@@ -0,0 +1,46 @@
...
+require_relative "vagrant-helpers"
+
+task :spec    => 'spec:all'
+task :default => :spec
+
+namespace :test do
+
+  desc "Generate a valid ssh-config"
+  task :ssh_config do
+    require "logstash/json"
```
ph: I do think this will work if you do `bundle exec` since the library will not be available anywhere.
You either need to use the core json or add a dependency on jrjackson.

qa/acceptance/Vagrantfile
```diff
@@ -0,0 +1,60 @@
...
+  ls.vm.provision :shell do |sh|
+    sh.path = "sys/ubuntu/bootstrap.sh"
+    sh.privileged = true
+  end
+
+  ls.vm.provision :shell do |sh|
+    sh.path = "sys/ubuntu/user_bootstrap.sh"
+    sh.privileged = false
+  end
+end
```
ph: What is your opinion on this proposition ? 
https://github.com/elastic/logstash/pull/5138#discussion_r60282516

ph:
Added a note concerning the configuration.

qa/acceptance/spec/centos/lib/install_spec.rb
purbon: sure is not necessary, but also not nice to pass this var thought a let statement, I guess the best way here might be pass with metadata, but not sure this is accessible in helpers, will need to investigate.

purbon:
Hi,
 I think I already shared with you, but https://github.com/elastic/logstash/pull/5138#issuecomment-212416758 does not work as it shows this error:

```
Vagrant experienced a version conflict with some installed plugins!\nThis usually happens if you recently upgraded Vagrant. As part of the\nupgrade process, some existing plugins are no longer compatible with\nthis version of Vagrant. The recommended way to fix this is to remove\nyour existing plugins and reinstall them one-by-one. To remove all\nplugins:\n\n    rm -r ~/.vagrant.d/plugins.json ~/.vagrant.d/gems\n\nNote if you have an alternate VAGRANT_HOME environmental variable\nset, the folders above will be in that directory rather than your\nuser's home directory.\n\nThe error message is shown below:\n\nBundler could not find compatible versions for gem \"bundler\":\n  In Gemfile:\n    vagrant (= 1.7.4) ruby depends on\n      bundler (<= 1.10.5, >= 1.5.2) ruby\n\n  Current Bundler version:\n    bundler (1.10.6)\nThis Gemfile requires a different version of Bundler.\nPerhaps you need to update Bundler by running `gem install bundler`?\nCould not find gem 'bundler (<= 1.10.5, >= 1.5.2) ruby', which is required by gem 'vagrant (= 1.7.4) ruby', in any of the sources.
```

I guess this is due to the fact that we would invoke vagrant with the bundler exec context, so it's own dependencies it gets confused. For this time we're good I guess by having this controversy, we can always simplify and fix later on.

qa/acceptance/spec/centos/spec_helper.rb
purbon: This point here is only for filtering which boxes belongs to which execution, on next changes we could select here only 1 box.

not sure what we define in the vagrantfile will become later available here.

.gitignore
purbon: we could, but not sure if makes sense! I know vendor is common directory, but not sure to add it as anyone could use different directories.

qa/acceptance/Vagrantfile
purbon: It for sure makes the configuration shorter for now, my thoughts here are that for now it simplify, as we have nothing platform specific, but if you take a look at https://github.com/elastic/elasticsearch/blob/master/Vagrantfile , the source of this file it gets some thing that are platform specific, so for now I stayed to something like this, but I'm good to make it as DRY as possible.

purbon:
Hi @ph && @talevy , as spoken with @talevy if you have nothing else pending here that you worry about I will move to have a shared branch with this PR as base, we can always iterate from there.

Thanks folks!

ph:
@purbon I don't really agree about how the config of the vagrantfile is made but I think it shouldn't block the creation of the branch.

purbon:
re vagrant file config, we can always iterate and make it simple, thanks!
will move forward to create the shared branch.

thanks

purbon:
@talevy as discussed, I added a few lines of code in https://github.com/elastic/logstash/pull/5138/commits/e61663e143b3d6db654e18139fb7bdeb2caf6c5e as a first pass on passing hostname information and run only 1 acceptance test, how this helps you.

purbon:
squashing and creating the shared branch as agreed here.

purbon:
closing this as this got manually upgraded to a shared branch at https://github.com/elastic/logstash/tree/feature/acceptance_test

@ph @talevy looking forward to more code collaboration here! point your PR's there 👍

PR #5150, opened 2016-04-20
Fix tagging so that logstash version resolves in example

dedemorton:
Added subs="attributes" so that the {logstash_version} variable resolves correctly in the output.


suyograo:
LGTM

PR #5162, opened 2016-04-21
Have ci scripts abort if any command exits nonzero

jordansissel:
This should fix scenarios where `rake bootstrap` fails due to bundler
failures and then some future command like `rake test:core` fails
because there's missing gems.


purbon:
please jenkins, test this!

jordansissel:
Rebased.

suyograo:
LGTM

Issue #5148, opened 2016-04-20
Logstash 5.0 doc fixes


acchen97:
**5.0 beta1**
- [x] Replace bin/plugin with bin/logstash-plugin
- [x] Kafka comp matrix updates https://github.com/elastic/logstash/issues/5645
- [x] Monitoring API updates https://github.com/elastic/logstash/issues/5732
- [x] Bring settings file docs to config options parity - https://www.elastic.co/guide/en/logstash/master/logstash-settings-file.html
- [x] Add messaging in command line flags page to encourage use of settings file instead

**5.0 GA**
- [x] Delineation of Elastic vs Community supported plugins
- [ ] Getting started guide - add jdbc and http input examples, ES + S3 outputs, incremental building of a config file (showcasing auto-reload option)
- [x] 5.0 Upgrades - upgrade guide for 2.0 and 2.2 should be consolidated into an overarching upgrades section
- Plugin specific doc additions:
  - [x] Plugin version
  - [x] Logstash core version compatibility, i.e. Logstash 2.4 - 5.x
  - [ ] Two links for plugin version navigation: master and latest compatible verison
  - [x] Version release date
  - [x] Link to change log
  - [ ] Add plugin code example snippet for most popular plugins - Alvin to define list
- Github page
  - [ ] Update top description of each plugin page
  - [ ] Add first [few] paragraphs plugin summary from documentation at top of README, also include official plugin docs link (master version)

**Completed**
- [x]  Community maintainer guide needs to be back to one page
- [x] Add a glossary of terms
- [x] Additional metrics for monitoring API - https://github.com/elastic/logstash/issues/4923
  - Better document/link http host & port info/settings from monitoring API page - https://github.com/elastic/logstash/issues/5285
- [x] Settings file for config options - https://github.com/elastic/logstash/issues/4401
- [x] Getting started guide updates
- [x] Pipeline changes, messaging against tuning output workers
- [x] Env variable page updates - funky rendering

**Backlog**
- [ ] New section "Adding & Collecting New Plugin Metrics" - adding guidance on how to add metrics to collect in plugins along with new metrics style guide https://github.com/elastic/logstash/issues/5716
- [ ] New windows doc page - just under the install scope
- [ ] Deploying/scaling guide updates

dedemorton:
Added glossary in PR #5237

acchen97:
[...]

PR #5167, opened 2016-04-21
Make structural changes to the plugin maintainer guide

dedemorton:
This PR addresses the first item listed in #5148 

Summary of changes:

* Added float tags so that all the content for the plugin maintainer guide appears in the same topic.
* Moved the plugin maintainer guide to be a sub-topic in the TOC (instead of a top-level topic).
* Moved the Contributing to Logstash section to the end of the TOC. Note that you don't see this change in the diff because the index.asciidoc file lives in the logstash-docs repo.

The TOC now looks like this:
newtoc


suyograo:
LGTM

PR #5168, opened 2016-04-21
Fixed sysv force kill

KnightOfNight:
I have completed the contributor agreement.

I have tested this on my own servers and found no issues.

### DIFF
--- a/pkg/logstash.default
+++ b/pkg/logstash.default
@@ -38,4 +38,4 @@
 # If this is set to 1, then when `stop` is called, if the process has
 # not exited within a reasonable time, SIGKILL will be sent next.
 # The default behavior is to simply log a message "program stop failed; still running"
-#LS_KILL_ON_STOP_TIMEOUT=0
+#KILL_ON_STOP_TIMEOUT=0
--- a/pkg/logstash.default
+++ b/pkg/logstash.default
@@ -38,4 +38,4 @@
 # If this is set to 1, then when `stop` is called, if the process has
 # not exited within a reasonable time, SIGKILL will be sent next.
 # The default behavior is to simply log a message "program stop failed; still running"
-KILL_ON_STOP_TIMEOUT=0
+#LS_KILL_ON_STOP_TIMEOUT=0
--- a/pkg/logstash.sysv
+++ b/pkg/logstash.sysv
@@ -35,7 +35,7 @@ LS_LOG_FILE="${LS_LOG_DIR}/$name.log"
 LS_CONF_DIR=/etc/logstash/conf.d
 LS_OPEN_FILES=16384
 LS_NICE=19
-LS_KILL_ON_STOP_TIMEOUT=0
+KILL_ON_STOP_TIMEOUT=0
 LS_OPTS=""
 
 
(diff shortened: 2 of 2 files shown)


pkg/logstash.sysv
```diff
@@ -96,12 +97,13 @@ stop() {
       sleep 1
     done
     if status ; then
-      if [ $KILL_ON_STOP_TIMEOUT -eq 1 ] ; then
+      if [ $LS_KILL_ON_STOP_TIMEOUT -eq 1 ] ; then
```
jordansissel: While we're here, can you quote the variable "$LS_KILL_ON_STOP_TIMEOUT" so it won't complain if someone has somehow unset this value?

KnightOfNight: Actually no.  If you quote a variable and then try to integer compare it with -eq, it generates an error.  However I can should and will change the operator from -eq to == in keeping with the fact that the variable is a string because it wasn't typeset -i.

KnightOfNight:
Not sure why I'm seeing the CLA error.  I completed it.

suyograo:
@KnightOfNight yeah, the echosign app that signs the CLA is slow to provide updates. Confirmed CLA signed

pkg/logstash.default
```diff
@@ -38,4 +38,4 @@
 # If this is set to 1, then when `stop` is called, if the process has
 # not exited within a reasonable time, SIGKILL will be sent next.
 # The default behavior is to simply log a message "program stop failed; still running"
-KILL_ON_STOP_TIMEOUT=0
+#LS_KILL_ON_STOP_TIMEOUT=0
```
purbon: This is introducing a breaking change here, is there any reason not to keep the name for now? I agree the rename is better, but if possible not to break the var usage, we should keep this as the same. Changing names will be good for 5.0 release.

I mean this option is working for stop function, so we should somehow keep it like this.

purbon:
related to: #4991

pkg/logstash.sysv
jordansissel: Quoting it doesn't make it any more a string than it already is ;)

I misspoke. I was thinking of `=` which has unexpected behavior (error) when you do not quote. It's still true of `-eq` though you always get an error (even if the error is different) if you don't have something coercible into a number:

```
$ [ $FOO = 1 ]
sh: 1: [: =: unexpected operator
$ [ "$FOO" = 1 ]
$ [ $FOO -eq 1 ]
sh: 3: [: -eq: unexpected operator
$ [ "$FOO" -eq 1 ]
sh: 4: [: Illegal number:
```

Still, we should quote this variable so the error is most informative "illegal number" instead of the very confusing "unexpected operator"

KnightOfNight: Sorry for the confusion, I agree with the quoting, I just failed to mention it here in comments, but it was in my latest push.  However, I still believe that using = rather than -eq is preferable because then no error is generated in any case.

pkg/logstash.default
KnightOfNight: This is a fair point.  I'll change the name back and PR the name change for v5.

suyograo:
LGTM

suyograo:
FIxed in https://github.com/elastic/logstash/commit/e7c05f9c0f5061d85807887e10b477167e3fed58

PR #5173, opened 2016-04-21
restructure tasks to have one VM run at a time during acceptance testing

talevy:
My initial attempt at making the acceptance test infra run one virtual-machine at a time, and bootstrap and clean up after itself.

### DIFF
--- a/qa/Rakefile
+++ b/qa/Rakefile
@@ -1,0 +1,53 @@
+require "rspec"
+require "rspec/core/runner"
+require "rspec/core/rake_task"
+require_relative "vagrant-helpers"
+
+task :spec    => 'spec:all'
+task :default => :spec
+
+namespace :test do
+
+  desc "Generate a valid ssh-config"
+  task :ssh_config do
+    require "json"
+    cd "acceptance" do
+      raw_ssh_config    = LogStash::VagrantHelpers.fetch_config[:stdout].split("\n");
+      parsed_ssh_config = LogStash::VagrantHelpers.parse(raw_ssh_config)
+      File.write("../.vm_ssh_config", parsed_ssh_config.to_json)
+    end
+  end
+
+  desc "Bootstrap all the VM's used for this tests"
+  task "setup" do
+    puts "bootstraping all VM's defined in acceptance/Vagrantfile"
+    cd "acceptance" do
+      LogStash::VagrantHelpers.bootstrap
+    end
+  end
+
+  namespace :acceptance do
+
+    desc "Run all acceptance"
+    task :all do
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+    end
+
+    desc "Run acceptance test in debian machines"
+    task :debian do
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/debian/**/*_spec.rb"]]))
+    end
+
+    desc "Run acceptance test in centos machines"
+    task :centos do
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/centos/**/*_spec.rb"]]))
+    end
+
+    desc "Run one single machine acceptance test"
+    task :single, :machine do  |t, args|
+      ENV['LS_VAGRANT_HOST'] = args[:machine]
+      platform = LogStash::VagrantHelpers.translate(args[:machine])
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+    end
+  end
+end
--- a/qa/Rakefile
+++ b/qa/Rakefile
@@ -18,36 +18,47 @@ namespace :test do
     end
   end
 
-  desc "Bootstrap all the VM's used for this tests"
-  task "setup" do
-    puts "bootstraping all VM's defined in acceptance/Vagrantfile"
-    cd "acceptance" do
-      LogStash::VagrantHelpers.bootstrap
-    end
-  end
-
   namespace :acceptance do
 
     desc "Run all acceptance"
     task :all do
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+      all_boxes = LogStash::VagrantHelpers::DEFAULT_CENTOS_BOXES + LogStash::VagrantHelpers::DEFAULT_DEBIAN_BOXES 
+      for box in all_boxes do
+        LogStash::VagrantHelpers.bootstrap(box)
+        exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+        LogStash::VagrantHelpers.destroy(box)
+      end
     end
 
     desc "Run acceptance test in debian machines"
     task :debian do
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/debian/**/*_spec.rb"]]))
+      for box in LogStash::VagrantHelpers::DEFAULT_DEBIAN_BOXES do
+        LogStash::VagrantHelpers.bootstrap(box)
+        exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/debian/**/*_spec.rb"]]))
+        LogStash::VagrantHelpers.destroy(box)
+      end
     end
 
     desc "Run acceptance test in centos machines"
     task :centos do
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/centos/**/*_spec.rb"]]))
+      for box in LogStash::VagrantHelpers::DEFAULT_CENTOS_BOXES do
+        LogStash::VagrantHelpers.bootstrap(box)
+        exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/centos/**/*_spec.rb"]]))
+        LogStash::VagrantHelpers.destroy(box)
+      end
     end
 
     desc "Run one single machine acceptance test"
     task :single, :machine do  |t, args|
-      ENV['LS_VAGRANT_HOST'] = args[:machine]
-      platform = LogStash::VagrantHelpers.translate(args[:machine])
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+      machine = args[:machine]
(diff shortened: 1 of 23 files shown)


qa/Rakefile
```diff
@@ -18,36 +18,47 @@ namespace :test do
...
+        LogStash::VagrantHelpers.bootstrap(box)
+        exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/centos/**/*_spec.rb"]]))
+        LogStash::VagrantHelpers.destroy(box)
+      end
     end
 
     desc "Run one single machine acceptance test"
     task :single, :machine do  |t, args|
-      ENV['LS_VAGRANT_HOST'] = args[:machine]
-      platform = LogStash::VagrantHelpers.translate(args[:machine])
```
talevy: @purbon, did you have this implemented? I couldn't find this anywhere

qa/acceptance/Vagrantfile
```diff
@@ -27,8 +27,8 @@ end
 
 def common_config(machine, type)
   machine.vm.provider "virtualbox" do |v|
-    v.memory = 4096
```
talevy: my computer can't hang

qa/vagrant-helpers.rb
```diff
@@ -1,15 +1,26 @@
...
 module LogStash
   class VagrantHelpers
+    DEFAULT_CENTOS_BOXES = [ "centos-6", "centos-7" ]
+    DEFAULT_DEBIAN_BOXES = [ "ubuntu-1204", "ubuntu-1404"]
 
-    def self.bootstrap
-      execute("vagrant up")
+    def self.bootstrap(box)
+      Dir.chdir("acceptance") do
+        p execute("vagrant up #{box} --provider virtualbox")
```
talevy: major refactoring needs to happen here. but the `--provider virtualbox` should stay. some environments have docker as the provider, or something else as the default.

qa/vagrant-helpers.rb
```diff
@@ -1,15 +1,26 @@
...
+        raw_ssh_config    = execute("vagrant ssh-config")[:stdout].split("\n");
+        parsed_ssh_config = parse(raw_ssh_config)
+        File.write(".vm_ssh_config", parsed_ssh_config.to_json)
```
talevy: TODO: do not rely on sharing state via a file

qa/Rakefile
purbon: this line is an stupid leftover, but this var has been pushed by me just yesterday, and somehow it filters the stuff on later stages.

qa/acceptance/Vagrantfile
purbon: good catch, this is a leftover (_i actually never removed_) from where I pull this config files (aka also running elasticsearch so need more mem).

qa/vagrant-helpers.rb
purbon: 👍 , good catch ... might be there is a way to tell this in the vagrantfile?

qa/vagrant-helpers.rb
purbon: Well, this is a bit of story here, let me explain my thoughts .... When I first thought about it my idea is/was to decouple everything as much as possible, so dumping the necessary ssh connection is something kinda necessary to do here.

On the other side, the only thing I can see is polling this same call for each time we run an acceptance test, and this would somehow slowdown the process. What do you think?

qa/Rakefile
```diff
@@ -18,36 +18,47 @@ namespace :test do
...
-  end
-
   namespace :acceptance do
 
     desc "Run all acceptance"
     task :all do
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+      all_boxes = LogStash::VagrantHelpers::DEFAULT_CENTOS_BOXES + LogStash::VagrantHelpers::DEFAULT_DEBIAN_BOXES 
+      for box in all_boxes do
+        LogStash::VagrantHelpers.bootstrap(box)
```
purbon: I'm on the fence here, my thoughts at the beginning where to design this as decoupled as possible so:

* Running test does nothing to do with bootstrapping the VM's.
* We get the machines bootstrapped/destroyed (__with rake tasks__), but all at once at the beginning of the process.

I guess the main idea here is to make this process as fast as possible, have you tested which approach is:

* Easier to operate.
* Faster to run the test, not just one but might be more than one. The common workflow I see here is to run all test for each platform (aka _debian_, _centos_, ....)
* Let's you investigate what went wrong if a test fails.

what do you think?

qa/Rakefile
```diff
@@ -18,36 +18,47 @@ namespace :test do
...
+        exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+        LogStash::VagrantHelpers.destroy(box)
```
purbon: I'm not confortable with this, I would like we are able to jump into the VM's if something wrong happen. this might help is debug and investigate what might be wrong,

qa/Rakefile
```diff
@@ -18,36 +18,47 @@ namespace :test do
...
-      LogStash::VagrantHelpers.bootstrap
-    end
-  end
-
   namespace :acceptance do
 
     desc "Run all acceptance"
     task :all do
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+      all_boxes = LogStash::VagrantHelpers::DEFAULT_CENTOS_BOXES + LogStash::VagrantHelpers::DEFAULT_DEBIAN_BOXES
```
purbon: I like the introduction of this constants! 👍

qa/Rakefile
```diff
@@ -18,36 +18,47 @@ namespace :test do
...
 
     desc "Run one single machine acceptance test"
     task :single, :machine do  |t, args|
-      ENV['LS_VAGRANT_HOST'] = args[:machine]
-      platform = LogStash::VagrantHelpers.translate(args[:machine])
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/**/*_spec.rb"]]))
+      machine = args[:machine]
+      ENV['LS_VAGRANT_HOST'] = machine
+      LogStash::VagrantHelpers.bootstrap(machine)
+      if LogStash::VagrantHelpers::DEFAULT_DEBIAN_BOXES.include? machine
```
purbon: a matter of taste, but around `include?` I think is better to have parentheses.

qa/acceptance/spec/centos/spec_helper.rb
```diff
@@ -1,14 +1,8 @@
...
-def default_vagrant_boxes
-  [
-    "centos-6",
-    "centos-7"
-  ]
-end
-
-selected_boxes = SpecsHelper.find_selected_boxes(default_vagrant_boxes)
+
+selected_boxes = SpecsHelper.find_selected_boxes(LogStash::VagrantHelpers::DEFAULT_CENTOS_BOXES)
```
purbon: 👍

qa/vagrant-helpers.rb
```diff
@@ -1,15 +1,26 @@
...
+      end
     end
 
-    def self.fetch_config
-      execute("vagrant ssh-config")
+    def self.destroy(box)
+      Dir.chdir("acceptance") do
+        execute("vagrant destroy #{box} -f")
+        File.delete(".vm_ssh_config")
```
purbon: are we sure we aim to remove this file after a destroy is called? not sure as we need this to have the state of ALL the machines for now.

purbon:
@talevy good work here! thanks for your ideas! I left you some comments here! happy to provide some context if some comments are not clear enough 😄

qa/Rakefile
purbon: I guess the best approach here would be to halt each vm after the run, and then make sure that before we start any new run we do with a clean VM, what do you think?

talevy: true. I had a personal debate over destroy vs halt. halt wins!

qa/Rakefile
talevy: agree. will change

qa/vagrant-helpers.rb
ph: I see two different solution:

- The actual problem here is that the IP and port can change when we run vagrant up, but `nothing` (maybe CI?) prevent us from using static IPs and port for the tests in our Vagrantfile. This would make the run idempotent.

- OR Since we know when we start a machine and stop a machine, we know when to refresh/invalidate the vagrant status

qa/Rakefile
ph: I would be inclined to say that the debugging workflow involve the quickest route to run, which usually mean only 1 vm, fix the problem with a red/green TDD and then test all the other vm.

Historically, I used kitchen-ci and the test  went to trash because it was slow to get feedback.

qa/Rakefile
ph: Actually we might want the following.

```
if the box doesnt exist
  bootstrap it
  take a clean snapshot
else the box exist
  restore a clean state snapshot?
end
run the tests
halt the machine
```

qa/vagrant-helpers.rb
ph: Just a note, using static port and ips is a old strategy when I was doing ansible works.

purbon: I would not go to the static IP strategy, as is simpler to just call it by name and fetch the ssh config, this give us more versatility for either more complex setups, but also for local environments, let each box decide. I'm -1 on going static anyway as is going to be unnecessary painful.

purbon:
@talevy now that are on it, could you add here also features to do something like:

```
Destroy all previous VM's.
Bootstrap all necessary VM's
Run the tests.
Halt all the machines
```

this is required to run for each platform as we test one kind of artifact for the platforms it supports, see https://github.com/elastic/logstash/issues/5181#issuecomment-214197071 for a long story around this. I think your improvements here are going to make having something like ^^ easy also.

purbon:
@talevy @ph can we make something to move this forward and merge to advance? we need this soon to be tested and move forward with the rest. what do you think?

ph:
Last time I've @talevy he was still blocked by his vagrant stuff, I will give a go today.

talevy:
Closing in favor of opening a new PR to resolve this issue. It has diverged from `feature/acceptance_test` too much to be comfortable to fix

Issue #5160, opened 2016-04-21
Failling test on auto_reload


ph:
```
And another one, non metric related (seed was 40152)

  1) LogStash::Agent#execute when auto_reload is true when calling reload_state! with a config that does not contain reload incompatible plugins does not reload if new config contains reload incompatible plugins
     Failure/Error: expect(subject).to receive(:upgrade_pipeline).at_least(2).times
       (#<LogStash::Agent:0x278d2891>).upgrade_pipeline(any args)
           expected: at least 2 times with any arguments
           received: 1 time with any arguments
     # ./logstash-core/spec/logstash/agent_spec.rb:156:in `(root)'

```
@jsvd Want to take a look?

jsvd:
thanks for reporting, checking. is this on master, or 2.x?

ph:
@jsvd master, really random :(

jsvd:
closed by #5175

PR #5175, opened 2016-04-22
make the plugin incompatibility tests less deterministic

jsvd:
upgrade_pipeline has the nasty side effect of..upgrading the pipeline. This test checks that upgrade_pipeline is called and, since the configuration changed ONCE, upgrade_pipeline should be called ONCE.

When I did the test I relaxed this concern because the expectation of calling a method mocks it. Which means the original isn't called, which means the pipeline isn't upgraded so, in next cycle of reloading, the upgrade is tried again.

By ensuring the method is actually called and performs its side effect, we can deterministically test that upgrade_pipeline should be called once and only once.

fixes https://github.com/elastic/logstash/issues/5160

### DIFF
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -82,7 +82,7 @@ describe LogStash::Agent do
         context "with a config that contains reload incompatible plugins" do
           let(:second_pipeline_config) { "input { stdin {} } filter { } output { }" }
 
-          it "does not reload if new config contains reload incompatible plugins" do
+          it "does not upgrade the new config" do
             t = Thread.new { subject.execute }
             sleep 0.01 until subject.running_pipelines? && subject.pipelines.values.first.ready?
             expect(subject).to_not receive(:upgrade_pipeline)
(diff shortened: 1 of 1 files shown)


purbon:
@jsvd have you tried running this, let's say for N times?

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -150,10 +150,10 @@
         context "with a config that does not contain reload incompatible plugins" do
           let(:second_pipeline_config) { "input { generator { } } filter { } output { }" }
 
-          it "does not reload if new config contains reload incompatible plugins" do
+          it "does upgrade the new config" do
             t = Thread.new { subject.execute }
             sleep 0.01 until subject.running_pipelines? && subject.pipelines.values.first.ready?
-            expect(subject).to receive(:upgrade_pipeline).at_least(2).times
+            expect(subject).to receive(:upgrade_pipeline).once.and_call_original
```
purbon: I wonder why this change of expectations here?, do you remember why you initially started with `at_least(2)` and now require only once? do you have a reason behind?

jsvd: `upgrade_pipeline` has the nasty side effect of..upgrading the pipeline. This test checks that upgrade_pipeline is called, and, since the configuration changed ONCE, `upgrade_pipeline` should be called ONCE.

When I did the test I relaxed this concern because the expectation of calling a method mocks it. Which means the original isn't called, which means the pipeline isn't upgraded so, in next cycle of reloading, the upgrade is tried again.

By ensuring the method is actually called and performs its side effect, we can deterministically test that upgrade_pipeline should be called once and only once.

jsvd:
@purbon well I can say I tried it more than N times (for some N).

purbon:
❤️ https://github.com/elastic/logstash/pull/5175#issuecomment-213426970 ❤️

logstash-core/spec/logstash/agent_spec.rb
purbon: makes sense to me.

ph:
LGTM, have tested it

purbon:
LGTM, :shipit: :shipit:

Issue #5176, opened 2016-04-22
Using the vagrant CLI from ruby


purbon:
Hi,
  when developing the new acceptance test framework we somehow face with the need to call the vagrant tool, for example to setup the machines, halt them, etc.. This is done using the vagrant CLI tool, however if we call this tool using the `bundle exec` pattern, we get this error:

> Vagrant experienced a version conflict with some installed plugins! This usually happens if you recently upgraded Vagrant. As part of the upgrade process, some existing plugins are no longer compatible with this version of Vagrant. The recommended way to fix this is to remove your existing plugins and reinstall them one-by-one. To remove all plugins:    rm -r ~/.vagrant.d/plugins.json ~/.vagrant.d/gems Note if you have an alternate VAGRANT_HOME environmental variable set, the folders above will be in that directory rather than your user's home directory. The error message is shown below: Bundler could not find compatible versions for gem \"bundler\":  In Gemfile:   vagrant (= 1.8.1) was resolved to 1.8.1, which depends on      bundler (<= 1.10.6, >= 1.5.2)  Current Bundler version:   bundler (1.11.2) This Gemfile requires a different version of Bundler. Perhaps you need to update Bundler by running `gem install bundler`? Could not find gem 'bundler (<= 1.10.6, >= 1.5.2)', which is required by gem 'vagrant (= 1.8.1)', in any of the sources.

For now the _"expected" workaround_ is not to use that pattern when calling vagrant, however it would be nice to get a uniform behaviour if possible.

My current thoughts on this error are that bundler gets somehow confused with the vagrant tool requirements.

I've tried an approach proposed by @ph at https://github.com/elastic/logstash/pull/5138#issuecomment-212416758, but this actually does not help much.

We need to investigate this as this is an actual problem for the workflow to be as easy and simple as possible.

PR #5180, opened 2016-04-22
Fixing the vagrant cli error under bundler and a refactor of the vagrant call

ph:
Vagrant is a ruby application and use bundler to manager his dependency,
when you invoke this kind of application inside a rake task you have to
clear the environment to make sure the application his able to run his
own dependency and not only the runner dependency.

This PR also introduce a small refactor of the Execution of local
command to allow the command to give better feedback when its actually
failling.

Fixes: #5176

### DIFF
--- a/qa/vagrant-helpers.rb
+++ b/qa/vagrant-helpers.rb
@@ -1,11 +1,12 @@
 # encoding: utf-8
 require "open3"
+require "bundler"
 
 module LogStash
-  class VagrantHelpers
+  class CommandExecutor
     class CommandError < StandardError; end
 
-    class ExecuteResponse
+    class CommandResponse
       attr_reader :stdin, :stdout, :stderr, :exitstatus
 
       def initialize(stdin, stdout, stderr, exitstatus)
@@ -20,12 +21,36 @@ module LogStash
       end
     end
 
+    def self.run(cmd)
+      # This block is require to be able to launch a ruby subprocess
+      # that use bundler.
+      Bundler.with_clean_env do
+        Open3.popen3(cmd) do |stdin, stdout, stderr, wait_thr|
+          CommandResponse.new(stdin, stdout.read.chomp, stderr.read.chomp, wait_thr.value.exitstatus)
+        end
+      end
+    end
+
+    # This method will raise an exception if the `CMD`
+    # was not run successfully and will display the content of STDERR
+    def self.run!(cmd)
+      response = run(cmd)
+    
+      unless response.success?
+        raise CommandError, "CMD: #{cmd} STDERR: #{response.stderr}"
+      end
+      response
+    end
+  end
+
+  class VagrantHelpers
+
     def self.bootstrap
-      execute_successfully("vagrant up")
+      CommandExecutor.run!("vagrant up")
     end
 
     def self.fetch_config
-      execute_successfully("vagrant ssh-config")
+      CommandExecutor.run!("vagrant ssh-config")
     end
 
     def self.parse(lines)
@@ -44,23 +69,6 @@ module LogStash
       end
       hosts << host
       hosts
-    end
-
-    private
-
-    def self.execute(cmd)
-      Open3.popen3(cmd) do |stdin, stdout, stderr, wait_thr|
-        ExecuteResponse.new(stdin, stdout.read.chomp, stderr.read.chomp, wait_thr.value.exitstatus)
-      end
-    end
-
-    def self.execute_successfully(cmd)
-      response = execute(cmd)
-    
-      unless response.success?
-        raise CommandError, "CMD: #{cmd} STDERR: #{response.stderr}"
-      end
-      response
     end
   end
 end
--- a/qa/vagrant-helpers.rb
+++ b/qa/vagrant-helpers.rb
@@ -3,13 +3,29 @@ require "open3"
 
 module LogStash
   class VagrantHelpers
+    class CommandError < StandardError; end
+
+    class ExecuteResponse
+      attr_reader :stdin, :stdout, :stderr, :exitstatus
+
+      def initialize(stdin, stdout, stderr, exitstatus)
+        @stdin = stdin, 
+        @stdout = stdout
+        @stderr = stderr
+        @exitstatus = exitstatus
+      end
+
+      def success?
+        exitstatus == 0
+      end
+    end
 
     def self.bootstrap
-      execute("vagrant up")
+      execute_successfully("vagrant up")
     end
 
     def self.fetch_config
-      execute("vagrant ssh-config")
+      execute_successfully("vagrant ssh-config")
     end
 
     def self.parse(lines)
@@ -34,10 +50,17 @@ module LogStash
 
     def self.execute(cmd)
       Open3.popen3(cmd) do |stdin, stdout, stderr, wait_thr|
-        { :stdout => stdout.read.chomp, :stderr => stderr.read.chomp,
-          :exit_status => wait_thr.value.exitstatus }
+        ExecuteResponse.new(stdin, stdout.read.chomp, stderr.read.chomp, wait_thr.value.exitstatus)
       end
     end
 
+    def self.execute_successfully(cmd)
+      response = execute(cmd)
+    
+      unless response.success?
+        raise CommandError, "CMD: #{cmd} STDERR: #{response.stderr}"
+      end
+      response
+    end
   end
 end
--- a/qa/Rakefile
+++ b/qa/Rakefile
@@ -14,7 +14,7 @@ namespace :test do
   task :ssh_config do
     require "json"
     cd "acceptance" do
-      raw_ssh_config    = LogStash::VagrantHelpers.fetch_config[:stdout].split("\n");
+      raw_ssh_config    = LogStash::VagrantHelpers.fetch_config.stdout.split("\n");
       parsed_ssh_config = LogStash::VagrantHelpers.parse(raw_ssh_config)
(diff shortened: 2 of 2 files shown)


qa/vagrant-helpers.rb
```diff
@@ -34,10 +50,17 @@ def self.parse(lines)
 
     def self.execute(cmd)
       Open3.popen3(cmd) do |stdin, stdout, stderr, wait_thr|
-        { :stdout => stdout.read.chomp, :stderr => stderr.read.chomp,
-          :exit_status => wait_thr.value.exitstatus }
+        ExecuteResponse.new(stdin, stdout.read.chomp, stderr.read.chomp, wait_thr.value.exitstatus)
       end
     end
 
+    def self.execute_successfully(cmd)
```
talevy: for simplicity, I think we should just have one `execute` command with this behavior

qa/vagrant-helpers.rb
```diff
@@ -34,10 +50,17 @@ def self.parse(lines)
 
     def self.execute(cmd)
       Open3.popen3(cmd) do |stdin, stdout, stderr, wait_thr|
-        { :stdout => stdout.read.chomp, :stderr => stderr.read.chomp,
-          :exit_status => wait_thr.value.exitstatus }
+        ExecuteResponse.new(stdin, stdout.read.chomp, stderr.read.chomp, wait_thr.value.exitstatus)
```
purbon: nice, do you know why this helps on getting the error out?

ph: Well wrapping this in a class can allow us to iterate on the response object, It could have been a struct. It just easier to refactor than using a hash.

talevy: it is nice to have a bit more structure, and the added sugar of methods like `success?` I also feel like this structure will lend itself to possible future improvements where we want to wrap async executions (background processes).

qa/vagrant-helpers.rb
ph: I don't agree with that, If I want to reuse the executor I want to control the behavior of it to make specific assertions. To be clear I want to reuse that code for the CLI test and make custom helper that will handle the right feedback and not depend on the implementation exception.

Its pretty common to use a `!` in ruby if the method can raise an exception. See `save and save!` in ActiveRecord.

talevy: so should we do that? `execute` and `execute!`?

qa/vagrant-helpers.rb
purbon: sure, but does this helps on running vagrant when doing `bundle exec rake test .... ` ? this is my question, for sure I undersand the OO design principles.

qa/vagrant-helpers.rb
```diff
@@ -1,15 +1,56 @@
...
+
+      def success?
+        exitstatus == 0
+      end
+    end
+
+    def self.run(cmd)
+      # This block is require to be able to launch a ruby subprocess
+      # that use bundler.
+      Bundler.with_clean_env do
```
purbon: I'm wondering why this might work here, and not when I tested this same code inside the Rake task, any idea? I always got this same error: see #5176, we should make sure we can call `vagrant` with `rake foo` and with `bundler exec rake foo`

qa/vagrant-helpers.rb
ph: No it doesn't but this https://github.com/elastic/logstash/pull/5180/files does.

qa/vagrant-helpers.rb
```diff
@@ -1,15 +1,56 @@
 # encoding: utf-8
 require "open3"
+require "bundler"
 
 module LogStash
+  class CommandExecutor
```
purbon: what do you think about breaking this into more separate files? and folders?

qa/vagrant-helpers.rb
talevy: > if you go ahead and merge, please make sure this works 👍 

will verify before LGTM

ph: I've tested with both, bundle and rake I do see any errors on my side.

But I don't think we should aim for making `rake` without prefixing it work, its always better to use `bundle exec` to make sure the gem in the current directory are used.

qa/vagrant-helpers.rb
ph: I am OK to move it, WDYT about `./lib/logstash/command_executor` ?

qa/vagrant-helpers.rb
ph: @talevy CommandResponse.stdout could return a future value.

qa/vagrant-helpers.rb
purbon: are you talking about the main lib directory? or something inside the qa folder? my thoughts are more on keeping the qa framework all united inside the same directory, so later we could make it easy a single gem, what do you think?

purbon:
@ph I tested this PR in my machine and got this report:

```
skywalker% rake test:ssh_config
cd acceptance
cd -
skywalker% cat .vm_ssh_config 
[{"host":"ubuntu-1204","hostname":"127.0.0.1","port":"2200"},{"host":"ubuntu-1404","hostname":"127.0.0.1","port":"2201"},{"host":"centos-6","hostname":"127.0.0.1","port":"2202"},{"host":"centos-7","hostname":"127.0.0.1","port":"2203"}]%                                  skywalker% rm .vm_ssh_config 
skywalker% bundle exec rake test:ssh_config
cd acceptance
rake aborted!
LogStash::CommandExecutor::CommandError: CMD: vagrant ssh-config STDERR: Vagrant experienced a version conflict with some installed plugins!
This usually happens if you recently upgraded Vagrant. As part of the
upgrade process, some existing plugins are no longer compatible with
this version of Vagrant. The recommended way to fix this is to remove
your existing plugins and reinstall them one-by-one. To remove all
plugins:

    rm -r ~/.vagrant.d/plugins.json ~/.vagrant.d/gems

Note if you have an alternate VAGRANT_HOME environmental variable
set, the folders above will be in that directory rather than your
user's home directory.

The error message is shown below:

Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    vagrant (= 1.7.4) ruby depends on
      bundler (<= 1.10.5, >= 1.5.2) ruby

  Current Bundler version:
    bundler (1.10.6)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?
Could not find gem 'bundler (<= 1.10.5, >= 1.5.2) ruby', which is required by gem 'vagrant (= 1.7.4) ruby', in any of the sources.
/Users/purbon/work/logstash/qa/vagrant-helpers.rb:40:in `run_successfully!'
/Users/purbon/work/logstash/qa/vagrant-helpers.rb:53:in `fetch_config'
/Users/purbon/work/logstash/qa/Rakefile:15:in `(root)'
/Users/purbon/work/logstash/qa/Rakefile:14:in `(root)'
Tasks: TOP => test:ssh_config
skywalker% 
```

1.- is run without bundle, 2.- is run with bundle exec.

any thoughts on why this might happen? my thoughts are on the version available of bundler in my installation vs the one required for vagrant.

you said this works for you, how is your setup/installation look like?

purbon:
I did a test by having the same bundler version as vagrant, 

```
skywalker% gem list bundler

*** LOCAL GEMS ***

bundler (1.10.5, 1.9.10)
```

and it worked:

```
skywalker% bundle exec rake test:ssh_config      
cd acceptance
cd -
skywalker% 
```

so my issue here was more related to different bundler versions, not sure how to handle this in a nice way? I mean I'm sure as It happened to me, we'll have N different bundler versions ad we can not force users of this to have a given one.

What do you think?

qa/vagrant-helpers.rb
ph: I am talking about creating that inside the qa folder.

ph:
I haven't seen that specific behavior with a bundler version.

You were running 1.9.X and installed a 1.10.X ? If I look at my current gems I am running an even newer version than you. Which is probably not the version shipped with vagrant.

```
~ ❯❯❯ gem list | grep bundler
bundler (1.11.2)
```
It might be a bug with `Bundler.with_clean_env` with older version of bundler that got fixed in newer version.

> so my issue here was more related to different bundler versions, not sure how to handle this in a nice way? 

The only way I remember to enforce a dependency on a specific version of bundler is to add a `If` in the gemfile that check for the current version of bundler.

ph:
Something like:

```
if Gem::Version.new(Bundler::VERSION) < Gem::Version.new('XXXX')
```

I agree it feel weird.

The only problem I can see is that the qa test requires a newer version than the one shipped with logstash..

qa/vagrant-helpers.rb
```diff
@@ -1,15 +1,56 @@
...
+      Bundler.with_clean_env do
+        Open3.popen3(cmd) do |stdin, stdout, stderr, wait_thr|
+          CommandResponse.new(stdin, stdout.read.chomp, stderr.read.chomp, wait_thr.value.exitstatus)
+        end
+      end
+    end
+
+    # This method will raise an exception if the `CMD`
+    # was not run successfully and will display the content of STDERR
+    def self.run_successfully!(cmd)
```
purbon: I would remove this to be ```ruby def self.run!(cmd)  ```  I somehow find the successfully part of the name misleading.

purbon:
left small comment, for the rest all LGTM, :shipit: :shipit:

ph:
updated with your comment

ph:
rebased

Issue #5186, opened 2016-04-25
Platform information need to be abstracted outside of the spec tests


talevy:
the platform mappings to each vagrant box defined in the acceptance testing framework need to be configured outside of the rspec. This will allow the specs to be generalized across both platforms and make the tests more DRY.

ph:
We have decided to make the mapping of the platform outisde of the vagrant file inside a json file.
This will have the benefit to make them available inside the rake task and inside vagrant.

ph:
@talevy @purbon I am suggesting the following structure for the json file

```json
# platforms.json
{
"ubuntu-1204": { "box": "elastic/ubuntu-12.04-x86_64", "type": "ubuntu"},
"centos-6": { "box": "elastic/centos-6-x86_64", "type": "centos" }
}
```
And this will make this possible in the vagrant file.
```ruby
boxes.each |machine_name, config|
machine_type = config["type"]
 config.vm.define machine_name do |ls|
     ls.vm.box = config["box"]
   ls.vm.provider "virtualbox" do |v|
     v.memory = 4096
     v.cpus = 4
   end
   ls.vm.synced_folder "../../build", "/logstash-build"
   ls.vm.provision :shell do |sh|
     sh.path = "sys/#{machine_type}/bootstrap.sh"
     sh.privileged = true
  end

   ls.vm.provision :shell do |sh|
     sh.path = "sys/{machine_type}/user_bootstrap.sh"
     sh.privileged = false
   end
end
```

ph:
Lets try to define the scenarios that we want to support.
```
rake test:acceptance:all # Run all tests on all platforms
rake test:acceptance:ubuntu # run all the spec on all the platform of type `ubuntu`
rake test:acceptance:centos # run all the spec on all the platform of type `centos`
rake test:acceptance:single[ubuntu-1204] # run all the specs on the ubuntu-1204 machine
```

I think we should also be able to support the following

```
HOST=ubuntu-1204 bundle exec spec spec/my_specific_spec.rb:45 # Run a specific example on a machine.
```
```

ph:
Concerning the structure of the spec itself, we try to make them platform agnostic.
As an example the `install_spec` would look like something like this.

```diff
diff --git a/qa/acceptance/spec/debian/lib/install_spec.rb b/qa/acceptance/spec/debian/lib/install_spec.rb
index 4b10baa..5070a1b 100644
--- a/qa/acceptance/spec/debian/lib/install_spec.rb
[...]

PR #5187, opened 2016-04-25
Platforms are now defined in a json file

ph:
Not ready for major review.

Fix: #5186


ph:
@purbon @talevy first try to move the platform definition out of the vagrantfile. WDYT?

purbon:
@ph major comments here is I would merge the platform.json file and the ssh connection config file, it does not make sense to have two config files, what do you think?

qa/Rakefile
```diff
@@ -33,14 +33,11 @@ namespace :test do
...
 
-    desc "Run acceptance test in debian machines"
-    task :debian do
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/debian/**/*_spec.rb"]]))
-    end
-
-    desc "Run acceptance test in centos machines"
-    task :centos do
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/centos/**/*_spec.rb"]]))
+    platforms.types.each do |type|
```
purbon: would this be better to be an static method?

qa/Vagrantfile
```diff
@@ -0,0 +1,27 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+require_relative "./platform_config.rb"
+
+Vagrant.configure(2) do |config|
+  platforms = PlatformConfig.new
```
purbon: I know you love this way of doing vagrant files, but can we keep the specific config areas for each platform? as soon as we've specific needs for a platform, or even a version, this kind of config file will be a mess.

ph: Can you give me an example of a change that could make this a mess?

ph: The platform is encapsulated into his own class so we need specific platform dependent stuff we can also do it there.

ph: I suggest that if we ever have the need a different settings, we make the changes. But right now since everything is the same we should just use the dry up version instead.

ph:
@purbon Concerning the merging of the `ssh_config` and the platform.

 I don't think its doable `ssh_config` is dependent of the current run on a specific machine, so this file will be different for every developers and we don't want to have it in our git repo. The content of the platforms.json is static and need to be present in our source control to make sure we use the same vms configuration.

qa/Vagrantfile
purbon: I'm talking about different bootstrapping needs, but you got your point, we can always extend it when necessary, what I don't want us to end up is with a very DRY vagrantfile, but _complicated_ to follow and understand.

purbon:
@ph we can merge them at update time, so doing the ssh config is just updating the main config file, this is only to keep sanity and have only one source of data, not many. My main concern here is at debug time, if I'm looking for the ubuntu-1204 characteristics I don't want to go to 2 different files.

ph:
@purbon I know we can do it at update time, but don't you agree it will be a pain to make sure we don't push it to the repo? This will become another `.lock` file and I personally I don't want that.

I am even wondering if we just get the performance hit of running vagrant status, we could cache the result if we plan to run mutliple vms from a single rake task.

ph:
Cache as in memory cache, not disk cache.

ph:
rebased with the branch

ph:
> if I'm looking for the ubuntu-1204 characteristics I don't want to go to 2 different files.

Thats a good point, but the real use case is not to use the port or ip. I don't need that information to log into a machine to debug something. I can just do `vagrant ssh ubuntu-1204` and all the complexity is handled by vagrant itself.

purbon:
@ph I would rebase and merge this, LGTM! :shipit: :shipit:

purbon:
pd, squash commits, 👍 , keep it in mind :-P

Issue #5179, opened 2016-04-22
Bundler install broken in latest master & 5.0 for all redhat based distros


purbon:
Hi,
  running the rake test:core, previous bootstrap, in fedora (also in other linuxes) break with latest master. see https://gist.github.com/purbon/495be8d1dbe735ae9c7e95bcb6702c04 for details. This might be due to some issue with https and certificates as running the `rake bootstrap` with the Gemfile pointed to `http://rubygems.org` works without problems.

Also using jruby .23 (latest used in logstash) fixes the issue, this makes me belive is some incompatiblity with the new introduced jruby-openssl version, see https://github.com/elastic/logstash/commit/5de3ce40d1c2f12395ecee4685ca98fa7b206032 that is where this problem started on https://logstash-ci.elastic.co/job/elastic-logstash-master/85/ 

testing now other implication here as using the new packaged on this platforms and try to install plugins.

__update__

plugin manager is also broken for _fedora_ and might be for other linux, see

```
[root@localhost vagrant]# /opt/logstash/bin/logstash --version
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
logstash 5.0.0.dev
[root@localhost vagrant]# /opt/logstash/bin/logstash-plugin list --verbose | grep jdbc
logstash-input-jdbc (3.0.2)
[root@localhost vagrant]# /opt/logstash/bin/logstash-plugin uninstall logstash-input-jdbc
Uninstalling logstash-input-jdbc
Error NoMethodError, retrying 1/10
undefined method `path' for #<Bundler::Source::Rubygems:0x5344533b>
WARNING: SSLSocket#session= is not supported
Error NoMethodError, retrying 2/10
undefined method `path' for #<Bundler::Source::Rubygems:0x7d377038>
WARNING: SSLSocket#session= is not supported
Error NoMethodError, retrying 3/10
undefined method `path' for #<Bundler::Source::Rubygems:0x7ff5c03e>
WARNING: SSLSocket#session= is not supported
Error NoMethodError, retrying 4/10
undefined method `path' for #<Bundler::Source::Rubygems:0x26e9526>
WARNING: SSLSocket#session= is not supported
^C[root@localhost vagrant]/opt/logstash/bin/logstash-plugin uninstall logstash-output-elasticsearch
[...]

PR #5198, opened 2016-04-26
Set the HEAP Size to 1gig when running rake tasks

ph:
Logstash uses by default in the shell script 1 gig of memory for the
JVM. but when we use rake it uses the default settings of JRUBY and
start a vm with 500megs of memory. The recent changes in JRUBY and
JRuby-OpenSSL have increased the base memory and make the installing of
plugins fail with a OOM issues. This PR make sure the CI scripts and the rake
jruby launch uses the default of 1gig.

Fixes: #5179

### DIFF
--- a/rakelib/z_rubycheck.rake
+++ b/rakelib/z_rubycheck.rake
@@ -32,7 +32,7 @@ if ENV['USE_RUBY'] != '1'
     # if required at this point system gems can be installed using the system_gem task, for example:
     # Rake::Task["vendor:system_gem"].invoke(jruby, "ffi", "1.9.6")
 
-    exec(jruby, "-J-Xmx1024m", "-S", rake, *ARGV)
+    exec(jruby, "-J-Xmx1g", "-S", rake, *ARGV)
   end
 end
 
--- a/rakelib/z_rubycheck.rake
+++ b/rakelib/z_rubycheck.rake
@@ -32,7 +32,7 @@ if ENV['USE_RUBY'] != '1'
     # if required at this point system gems can be installed using the system_gem task, for example:
     # Rake::Task["vendor:system_gem"].invoke(jruby, "ffi", "1.9.6")
 
-    exec(jruby, "-S", rake, *ARGV)
+    exec(jruby, "-J-Xmx1024m", "-S", rake, *ARGV)
   end
 end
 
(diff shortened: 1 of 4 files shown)


rakelib/z_rubycheck.rake
```diff
@@ -32,7 +32,7 @@ if ENV['USE_RUBY'] != '1'
     # if required at this point system gems can be installed using the system_gem task, for example:
     # Rake::Task["vendor:system_gem"].invoke(jruby, "ffi", "1.9.6")
 
-    exec(jruby, "-S", rake, *ARGV)
+    exec(jruby, "-J-Xmx1024m", "-S", rake, *ARGV)
```
jordansissel: why 1024m and not 1g like the rest of this patch?

ph: Because I wanted to type more characters ;)

colinsurprenant:
wouldn't it be possible to DRY this and maintain only 1 place where the default JVM heap size is specified for both the build/rake tasks and the logstash startup?

ph:
@colinsurprenant Yes, but probably not easy, we also use the settings in the startup script for different OS. I suggest the following, go ahead with this patch and create an issue for refactor of the common settings to an external file?  Maybe some we could use some kind of templating?

```
pkg/logstash.default
15:#LS_HEAP_SIZE="1g"
pkg/logstash.sysv
32:LS_HEAP_SIZE="1g"
pkg/logstash.sysv.redhat
38:LS_HEAP_SIZE="1g"
pkg/logstash.sysv.debian
39:LS_HEAP_SIZE="1g"
pkg/logstash.upstart.ubuntu
25:  LS_HEAP_SIZE="1g"
```

colinsurprenant:
@ph it depends if this is urgent to merge? is it really that difficult to externalize that setting and read it in all these contexts?

ph:
@colinsurprenant I will check if I can find a nice way to do, but it certainly requires more changes/testing

colinsurprenant:
@ph ok! that'd be nice, otherwise I'm +1 on what you proposed (quick change and refactor issue) ... it's just that by experience I know that these non-urgent refactor issues are often forgotten :(

ph:
@colinsurprenant I am usually a scout and try to improve on each PR, but I would like to have full package acceptance testing for that kind of changes :(

jordansissel:
LGTM build succeded :P

colinsurprenant:
scout

PR #5199, opened 2016-04-26
Enable DEBUG=1 when using bundler related task

ph:
When running any logstash's command you can use DEBUG=1 which should
give the user a bit more information about what LS is doing.

This PR changes the behavior of bundler to make sure its run with the
`--verbose` flag.

### DIFF
--- a/lib/bootstrap/bundler.rb
+++ b/lib/bootstrap/bundler.rb
@@ -104,13 +104,22 @@ module LogStash
       ::Bundler.settings[:gemfile] = LogStash::Environment::GEMFILE_PATH
       ::Bundler.settings[:without] = options[:without].join(":")
 
+      if !debug?
+        # Will deal with transient network errors
+        execute_bundler_with_retry(options)
+      else
+        options[:verbose] = true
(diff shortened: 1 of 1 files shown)


lib/bootstrap/bundler.rb
```diff
@@ -104,13 +104,22 @@ def invoke!(options = {})
       ::Bundler.settings[:gemfile] = LogStash::Environment::GEMFILE_PATH
       ::Bundler.settings[:without] = options[:without].join(":")
 
+      if !debug?
+        # Deal with transient network errors
+        return execute_bundler_with_retry(options)
```
jsvd: any reason for the explicit return?

jsvd:
LGTM

lib/bootstrap/bundler.rb
ph: Right there is no need for that.

ph:
@jsvd which branch should we target this?

jsvd:
All...master 5.0 2.x 2.3

PR #5202, opened 2016-04-26
bug fix, changed == to =

KnightOfNight:
double equals only works in BASH, not /bin/sh which may only be a subset of BASH.


ph:
LGTM, sad simple mistake :( 

== is a bash-specific alias for =

ph:
merge in 2.3 ccc8d79089c233b4c6931fcdd44c0d4bee84eec9

Issue #5163, opened 2016-04-21
already initialized constant warning in specs


colinsurprenant:
this warning shows up in `rake test:core` 

```
.../logstash/logstash-core/lib/logstash/api/init.ru:1 warning: already initialized constant ROOT
```

would it be possible to get rid of this ROOT constant redefinition?

PR #5208, opened 2016-04-27
Remove constant redefinition

ph:
Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: https://www.elastic.co/contributor-agreement/

The Rack config was redefining a constant in the configuration, since
this value isn't used anywhere else there was no need to use a constant.

The conflicting ROOT contants was defined in `lib/logstash/patches/profile_require_calls.rb`

Fixes: #5163


jordansissel:
Haven't tested, but the change LGTM

PR #5213, opened 2016-04-27
update SNAPSHOT tests to run

talevy:
I was having trouble running these specs locally. these changes cleanup some path and naming issues.


ph:
LGTM

PR #5214, opened 2016-04-27
Support ( in path and actually set JAVA_OPTS

ecki:
The current logic for printing %JAVA_HOME% inside if ( ) block failes if the path contains a `(` which is the case for 32bit java (`C:\Program Files (x86)\java\jdk`)

The current logic for appending to JAVA_OPTS does not work as it will be expanded while interpreting the else block. Use goto instead.

Personally I do not at all like that this glutters all my global environemnt, but hey I guess to each its own taste :)

### DIFF
--- a/bin/setup.bat
+++ b/bin/setup.bat
@@ -19,9 +19,8 @@ if not defined JAVA_HOME IF EXIST %ProgramData%\Oracle\java\javapath\java.exe (
     for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_EXE=%%a
 )
 if defined JAVA_EXE set JAVA_HOME=%JAVA_EXE:\bin\java.exe=%
-if defined JAVA_EXE (
-    echo Using JAVA_HOME=%JAVA_HOME% retrieved from %ProgramData%\Oracle\java\javapath\java.exe
-)
+if defined JAVA_EXE echo Using JAVA_HOME=%JAVA_HOME% retrieved from %ProgramData%\Oracle\java\javapath\java.exe
+
 if not defined JAVA_HOME goto missing_java_home
 REM ***** JAVA options *****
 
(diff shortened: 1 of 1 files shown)


suyograo:
This was introduced by @StefanScherer in https://github.com/elastic/logstash/pull/4913

@StefanScherer can you check this out and let us know if it works in your testing?

suyograo:
@ecki Hi, can you please perform step 2 of https://github.com/elasticsearch/logstash/blob/master/CONTRIBUTING.md#contribution-steps

bin/setup.bat
```diff
@@ -19,9 +19,8 @@ if not defined JAVA_HOME IF EXIST %ProgramData%\Oracle\java\javapath\java.exe (
     for /f "tokens=2 delims=[]" %%a in ('dir %ProgramData%\Oracle\java\javapath\java.exe') do @set JAVA_EXE=%%a
 )
 if defined JAVA_EXE set JAVA_HOME=%JAVA_EXE:\bin\java.exe=%
-if defined JAVA_EXE (
-    echo Using JAVA_HOME=%JAVA_HOME% retrieved from %ProgramData%\Oracle\java\javapath\java.exe
-)
+if defined JAVA_EXE echo Using JAVA_HOME=%JAVA_HOME% retrieved from %ProgramData%\Oracle\java\javapath\java.exe
```
StefanScherer: This is definitely the more elegant way without the braces solving the problem showing and old variable

StefanScherer:
@suyograo I've got some minutes to test the PR:

```
PS C:\Users\vagrant\logstash> git fetch origin pull/5214/head:test
From https://github.com/elastic/logstash
 * [new ref]         refs/pull/5214/head -> test
PS C:\Users\vagrant\logstash> git checkout test
Switched to branch 'test'
PS C:\Users\vagrant\logstash> notepad bin\setup.bat
PS C:\Users\vagrant\logstash> .\bin\setup.bat
Using JAVA_HOME=C:\Program Files\Java\jre1.8.0_77 retrieved from C:\ProgramData\Oracle\java\javapath\java.exe
Unable to find JRuby.
If you are a user, this is a bug.
If you are a developer, please run 'rake bootstrap'. Running 'rake' requires the 'ruby' program be available.
```

It still works finding and setting JAVA_HOME, I only cloned the repo without JRuby, so the other error is normal.

LGTM

Issue #5212, opened 2016-04-27
Make the acceptance test use the the current logstash package.


ph:
The current acceptance test only work with the remote logstash package.

I don't think this is necessary and should be refactored to work with the current logstash version which is a Logstash snapshot version by itself.

So running the acceptance test would do the following.

- Make the artifacts, tar, deb, rpm
- Run the acceptance test on the generated artifact.

ph:
The packages will be generated and make available in a `vagrant shared folder` and will be renamed to 


```
logstash-current-snapshot.deb
logstash-current-snapshot.deb
logstash-current-snapshot.zip
logstash-current-snapshot.tar.gz
```

We will need some logic in place to simulate an upgrade scenario, fetch a previous package?

purbon:
closing this as this is already an option and the way the test are designed.

PR #5218, opened 2016-04-28
Add common ci helper script to run acceptance test

ph:
This change introduce a new command ci/ci_acceptance which is
responsable of building the packages, bootstraping the acceptance test
environment, launching the VMs and running the tests.

You can use the following command to target specific platform.

```
ci/ci_acceptance.sh all
ci/ci_acceptance.sh centos
ci/ci_acceptance.sh debian
```

This PR also add a new rake task to build all the artifacts in a single
run called rake artifact:all it make sure we only install the plugin
only once and make the build process faster.

Fixes: #5212

### DIFF
--- a/ci/ci_acceptance.sh
+++ b/ci/ci_acceptance.sh
@@ -1,0 +1,44 @@
+#!/bin/sh
+set -e
+
+# Since we are using the system jruby, we need to make sure our jvm process
+# uses at least 1g of memory, If we don't do this we can get OOM issues when
+# installing gems. See https://github.com/elastic/logstash/issues/5179
+export JRUBY_OPTS="-J-Xmx1g"
+
+SELECTED_TEST_SUITE=$1
+
+if [[ $SELECTED_TEST_SUITE == $"centos" ]]; then
+  rake artifact:rpm
+  echo "Acceptance: Installing dependencies"
+  cd qa
+  bundle install
+
+  echo "Acceptance: Running the tests"
+  bundle exec rake test:setup
+  bundle exec rake test:ssh_config
+  bundle exec rake test:acceptance:centos
+elif [[ $SELECTED_TEST_SUITE == $"debian" ]]; then
+  rake artifact:deb
+  echo "Acceptance: Installing dependencies"
+  cd qa
+  bundle install
+
+  echo "Acceptance: Running the tests"
+  bundle exec rake test:setup
+  bundle exec rake test:ssh_config
+  bundle exec rake test:acceptance:debian
+elif [[ $SELECTED_TEST_SUITE == $"all" ]]; then
+  echo "Building Logstash artifacts"
+  rake artifact:all
+
+  echo "Acceptance: Installing dependencies"
+  cd qa
+  bundle install
+
+  echo "Acceptance: Running the tests"
+  bundle exec rake test:setup
+  bundle exec rake test:ssh_config
+  bundle exec rake test:acceptance:all
+  cd ..
+fi
--- a/ci/ci_acceptance.sh
+++ b/ci/ci_acceptance.sh
@@ -9,6 +9,7 @@ export JRUBY_OPTS="-J-Xmx1g"
 SELECTED_TEST_SUITE=$1
 
 if [[ $SELECTED_TEST_SUITE == $"centos" ]]; then
+  echo "Generating the RPM, make sure you start with a clean environment before generating other packages."
   rake artifact:rpm
   echo "Acceptance: Installing dependencies"
   cd qa
@@ -19,6 +20,7 @@ if [[ $SELECTED_TEST_SUITE == $"centos" ]]; then
   bundle exec rake test:ssh_config
   bundle exec rake test:acceptance:centos
 elif [[ $SELECTED_TEST_SUITE == $"debian" ]]; then
+  echo "Generating the DEB, make sure you start with a clean environment before generating other packages."
   rake artifact:deb
   echo "Acceptance: Installing dependencies"
   cd qa
--- a/qa/Vagrantfile
+++ b/qa/Vagrantfile
@@ -12,15 +12,10 @@ Vagrant.configure(2) do |config|
         v.memory = 2096
         v.cpus = 4
       end
-      machine.vm.synced_folder "../../build", "/logstash-build", create: true
+      machine.vm.synced_folder "../build", "/logstash-build", create: true
       machine.vm.provision :shell do |sh|
         sh.path = "sys/#{platform.type}/bootstrap.sh"
         sh.privileged = true
-      end
-
-      machine.vm.provision :shell do |sh|
-        sh.path = "sys/#{platform.type}/user_bootstrap.sh"
-        sh.privileged = false
       end
     end
   end
--- a/qa/acceptance/spec/centos/lib/install_spec.rb
+++ b/qa/acceptance/spec/centos/lib/install_spec.rb
@@ -1,13 +1,13 @@
 # encoding: utf-8
-require_relative '../spec_helper'
-require          'logstash/version'
+require_relative "../spec_helper"
+require "logstash/version"
 
 describe "artifacts", :platform => :centos do
 
   shared_examples "installable" do |host, name|
 
     before(:each) do
-      install("/home/vagrant/logstash-latest-SNAPSHOT.rpm", host)
+      install("/logstash-build/logstash-#{LOGSTASH_VERSION}.noarch.rpm", host)
     end
 
     it "is installed on #{name}" do
--- a/qa/sys/centos/bootstrap.sh
+++ b/qa/sys/centos/bootstrap.sh
@@ -2,11 +2,3 @@
 
 yum update
 yum install -y java-1.8.0-openjdk-devel.x86_64
-
-##
-# Install logstash manually from a URL
-##
-BRANCH=${LOGSTASH_BRANCH:-'master'}
-BUILD_URL='https://s3-eu-west-1.amazonaws.com/build-eu.elasticsearch.org/logstash'
-URL="$BUILD_URL/$BRANCH/nightly/JDK8/logstash-latest-SNAPSHOT.rpm"
-wget --no-verbose $URL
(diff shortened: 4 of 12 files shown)


talevy:
can you update the readme file?

ph:
Done @talevy, forgot to squash before pushing and description of this PR was really bad.

talevy:
`bin/ci_acceptance`? I see `ci/ci_acceptance.sh`

talevy:
```bash
$ rake artifact:all
Could not find numerizer-0.1.1 in any of the sources
Run `bundle install` to install missing gems.
```

ph:
Weird let me retest that command directly with a clean checkout

ph:
@talevy will fix the typo in the commit message itself.

Concerning the `Could not find numerizer-0.1.1 in any of the sources` Do you have any `.bundle/config` from a previous debugging session. This can happen if you use `bundle exec rake` without using our shim.

I have started from a clean env and did get any issue.
``
/e/logstash git:fix/use-current-branch-artifact ❯❯❯ git status
On branch fix/use-current-branch-artifact
nothing to commit, working directory clean
~/e/logstash git:fix/use-current-branch-artifact ❯❯❯ ruby -v
jruby 1.7.25 (1.9.3p551) 2016-04-13 867cb81 on Java HotSpot(TM) 64-Bit Server VM 1.8.0_20-b26 +jit [darwin-x86_64]
~/e/logstash git:fix/use-current-branch-artifact ❯❯❯ gem list

*** LOCAL GEMS ***

bundler (1.11.2)
jar-dependencies (0.2.6)
jruby-openssl (0.9.16 java)
json (1.8.0 java)
rake (10.1.0)
rdoc (4.1.2)
~/e/logstash git:fix/use-current-branch-artifact ❯❯❯
```

File: qa/Vagrantfile [hunk 1]
```diff
@@ -12,16 +12,11 @@ Vagrant.configure(2) do |config|
...
         v.cpus = 4
       end
-      machine.vm.synced_folder "../../build", "/logstash-build", create: true
+      machine.vm.synced_folder "../build", "/logstash-build", create: true
       machine.vm.provision :shell do |sh|
         sh.path = "sys/#{platform.type}/bootstrap.sh"
         sh.privileged = true
       end
-
-      machine.vm.provision :shell do |sh|
```
purbon: why are we removing this step in the bootstrap?

qa/acceptance/spec/centos/lib/install_spec.rb
```diff
@@ -1,13 +1,13 @@
...
+require_relative "../spec_helper"
+require "logstash/version"
 
 describe "artifacts", :platform => :centos do
 
   shared_examples "installable" do |host, name|
 
     before(:each) do
-      install("/home/vagrant/logstash-latest-SNAPSHOT.rpm", host)
+      install("/logstash-build/logstash-#{LOGSTASH_VERSION}.noarch.rpm", host)
```
purbon: I wonder why this is necessary? might be we changed the way packages are generated and I did notice?

qa/sys/centos/bootstrap.sh
```diff
@@ -2,11 +2,3 @@
 
 yum update
 yum install -y java-1.8.0-openjdk-devel.x86_64
-
-##
-# Install logstash manually from a URL
-##
-BRANCH=${LOGSTASH_BRANCH:-'master'}
```
purbon: This files and processes are important because one acceptance test you should run is update from latest available release, so you need to bootstrap this with the latest rpm, deb, ... available to download.

ph: shouldn't this be done in the corresponding test suite, so the expectation can be controlled?

ci/ci_acceptance.sh
```diff
@@ -0,0 +1,15 @@
...
+rake artifact:all
+
+echo "Acceptance: Installing dependencies"
+cd qa
+bundle install
+
+echo "Acceptance: Running the tests"
+bundle exec rake test:setup
+bundle exec rake test:ssh_config
+bundle exec rake test:acceptance:all
```
purbon: good addition, but would be awesome we could have this script parametrized by platform type, this is necessary in order to simplify the process in CI environments.

qa/acceptance/spec/centos/lib/install_spec.rb
ph: This revert the previous commit. https://github.com/elastic/logstash/commit/795d5d0619b57431179afc50fdb18172af505f4b#diff-2aa36be24fd3a114b68db38ebed4a070R10

qa/sys/centos/bootstrap.sh
purbon: I don't think so, if we do that at bootstrapping process we can pull from a newly created "logstash-latest" url and keep this done only once during slowest part of process, doing that in test execution it unnecessary slows the process. But this are my preference, it does not means is perfect 👍

qa/acceptance/spec/centos/lib/install_spec.rb
purbon: ok, now I see .... github diff is confusing sometimes to me 👍

ci/ci_acceptance.sh
ph: Why would we need a platform type here?

purbon: we need it to get test that run for a given platform, it will give us
faster and detailed feedback in comparison to run all at once.

ph: I will add it.

ph: @purbon What would be the parameter name and the possible values?

purbon: let's wait on actually decide how do we aim to run this in CI environment, this script is here to empower it, so ....  I'm good with whatever path we take, but this script should help automated testing.

purbon:
@ph this branch needs a bit of rebase 👍

ph:
rebased updated description.

You know have 3 possible command.

```
ci/acceptance.sh debian
ci/acceptance.sh centos
ci/acceptance.sh all
```

When you test a specific platform you have to reset your environment before trying to test the other environment. `ci/acceptance.sh all` doesn't have this problem since it bundle all the build into the same task.

We may address this problem into a future PR.

qa/sys/centos/bootstrap.sh
purbon: My preference would be that test are mostly responsible of testing assertions and make this process as simple as possible, we should avoid any logic in there that is too much tied with the necessary moving parts to run the test (like downloading the artifacts).

for me this point is a blocker to merge this PR, we should get some consensus here 👍 , looking forward to it.

purbon:
> When you test a specific platform you have to reset your environment before trying to test the other environment. ci/acceptance.sh all doesn't have this problem since it bundle all the build into the same task.

this might be a problem for development environments mostly, isn't? can we might be add an small warning on top of this rspec calls? what do you think?

on CI, even if we use any of this, the workspace will be different so this will not be an issue.

ph:
> this might be a problem for development environments mostly, isn't? can we might be add an small warning on top of this rspec calls? what do you think?


I can certainly add a warning, where would be the best place?

purbon:
might some place in the CI scripts? not sure really.

qa/sys/centos/bootstrap.sh
ph: Lets try to find a middle ground  here.

The problem with the downloading the file at provisioning time is this operation is done once unless we destroy the machine **OR** manually run the vagrant provision.
Are we at risk to run the tests not on the latest released version?

This problem doesn't appear on CI because on every run the machine are rebuild from scratch, but locally as human we wont be doing that since Its really time consuming.

I agree having the spec download the file could be slower, but nothing prevent us from persisting it. We can do a head check on it to see if we need to update it? This would fix the problem of the slow.

Another question, am I wrong or this file is only needed in the upgrade scenario? Sot this file is only needed to be downloaded in a `before :all` block? 

In a normal test run on CI, downloading from bootstrapping vs once in a before :all should yield the same execution time?

ph:
added a small message concerning the rpm/debian in the ci/ci_acceptance.sh

qa/sys/centos/bootstrap.sh
purbon: > The problem with the downloading the file at provisioning time is this operation is done once unless we destroy the machine OR manually run the vagrant provision command.

nope, the situation here is the same as downloading the thing inside the rspec tests. Download should happen 1, then is responsibility of the test framework to keep the right state inside the tests.

> Are we at risk to run the tests not on the latest released version in local dev machine?

nope, as I said, risks are the same independently where do you download the artifact. we should be able to create a URL that lets us point to latest released.

>  but locally as human we won't be doing that since Its really time consuming.

this is specially my point here, I see test framework being sure that the state is the right one, and then we can reuse the download artifacts any time we run the test with the same machine.

> I agree having the spec download the file could be slower, but nothing prevent us from persisting it. 

this is what we get by doing this at provision time.

> before :all 

problem with this is that we download any time the rspec examples are executed, I want us to avoid that by doing this in  provision time.

ph: > nope, the situation here is the same as downloading the thing inside the rspec tests. Download should happen 1, then is responsibility of the test framework to keep the right state inside the tests.

How do you plan to resolve that?

purbon: we have to see how to do it, but end of the day is the same as if I download of have the rpm already available after bootstrap time. Let me explain better, only thing that is under discussion here is, 

* For you download + install should happen in rspec
* For me download should happen in bootstrap + install should happen in rspec.

Does this makes sense to you?

ph: >For you download(if needed) + install should happen in rspec

For me I though it was simplier, everything for this current test is in the same file, maybe I don't understand the advantages of doing that a bootstrap time.

But I don't think this should be a blocker for this current PR, since there is no tests for the upgrade scenario.
I will revert theses lines and let you go ahead with your solution.

ph: s/will/can

purbon:
I'm good with merging this as this, we can always bring back this scripts when the hole infra for downloading the latest logstash is ready.

purbon:
LGTM

PR #5219, opened 2016-04-28
Fix missing translation on debug_config and other keys from the agent

### LINKED ISSUE #5216: translation missing: en.logstash.runner.flag.debug_config
When I run logstash 2.3.2 on german Win64 with --help option the resulting usage text displays a missing translation for

    --debug-config                translation missing: en.logstash.runner.flag.debug_config (default: false)

BTW: it is not the first time I had this, maybe add an test calling `--help` and grepping for the error message before shipping?

ph:
Fix the issues found when running the i18n static analysis.

Fix #5216


jsvd:
LGTM

PR #5170, opened 2016-04-21
[WIP] refactor Ruby Event getter and setter

colinsurprenant:
This is a WIP for the Event  getter and setter refactor. 

Per #5141 and #5140 

This code is pushed in `feature/event_interface` **shared branch**.

For this to work the plugins using a local `:path` in the included modified `Gemfile` need to be tweeked to use the new interface. 

- [x] Modify the Ruby Event interface
- [x] Pass core specs using Ruby Event
- [x] Modify Java Event interface
- [x] Pass core specs using Java Event
- [x] bump logstash-core-plugin-api for new Event interface
- [x] publish modified plugins required for core specs using the new logstash-core-plugin-api version

PR #5224, opened 2016-04-29
add --preserve option to plugin install

colinsurprenant:
Allow using the `--preserve` option to the plugin install command to preserve any current gem options in the `Gemfile` for the installation of a plugin gem which is already in the `Gemfile`.

This is required by #5170 to support specifying plugin gems from a github repo/branch in the `Gemfile`. I believe this will also help avoiding the unnecessary publication of core plugins snapshots.

Ultimately this allows running the rake test:install* command by keeping current `Gemfile` gems options, for example a custom in-dev `:path` option, instead of overwriting them and forcing the use of the published gem.

### DIFF
--- a/lib/pluginmanager/install.rb
+++ b/lib/pluginmanager/install.rb
@@ -9,6 +9,7 @@ class LogStash::PluginManager::Install < LogStash::PluginManager::Command
   parameter "[PLUGIN] ...", "plugin name(s) or file", :attribute_name => :plugins_arg
   option "--version", "VERSION", "version of the plugin to install"
   option "--[no-]verify", :flag, "verify plugin validity before installation", :default => true
+  option "--preserve", :flag, "preserve current gem options", :default => false
   option "--development", :flag, "install all development dependencies of currently installed plugins", :default => false
   option "--local", :flag, "force local-only plugin installation. see bin/logstash-plugin package|unpack", :default => false
 
@@ -90,7 +91,15 @@ class LogStash::PluginManager::Install < LogStash::PluginManager::Command
 
     # Add plugins/gems to the current gemfile
     puts("Installing" + (install_list.empty? ? "..." : " " + install_list.collect(&:first).join(", ")))
-    install_list.each { |plugin, version, options| gemfile.update(plugin, version, options) }
+    install_list.each do |plugin, version, options|
+      if preserve?
+        plugin_gem = gemfile.find(plugin)
+        puts("Preserving Gemfile gem options for plugin #{plugin}") if plugin_gem && !plugin_gem.options.empty?
+        gemfile.update(plugin, version, options)
+      else
+        gemfile.overwrite(plugin, version, options)
+      end
+    end
 
     # Sync gemfiles changes to disk to make them available to the `bundler install`'s API
     gemfile.save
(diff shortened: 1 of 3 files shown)


lib/pluginmanager/install.rb
```diff
@@ -90,7 +91,13 @@ def install_gems_list!(install_list)
 
     # Add plugins/gems to the current gemfile
     puts("Installing" + (install_list.empty? ? "..." : " " + install_list.collect(&:first).join(", ")))
-    install_list.each { |plugin, version, options| gemfile.update(plugin, version, options) }
+    install_list.each do |plugin, version, options|
+      if preserve?
```
talevy: maybe add a debug log message here to make the behavior known on a per plugin basis?

purbon:
what about having a few test for this in the great integration test that @ph created? or might be add some checkups as unit test in https://github.com/elastic/logstash/tree/master/spec/pluginmanager ? what do you think?

purbon:
but in general I do like this option!!! good work!

lib/pluginmanager/install.rb
colinsurprenant: yap, added!

colinsurprenant:
@purbon the integration tests in place for the end-user usage should be sufficient since the `--preserve` option is off by default.  We could add integration tests for the behaviour of the option but since this is typically a development option I'm good in doing that in another PR.

lib/pluginmanager/install.rb
```diff
@@ -90,7 +91,15 @@ def install_gems_list!(install_list)
...
+        plugin_gem = gemfile.find(plugin)
+        puts("Preserving Gemfile gem options for plugin #{plugin}") if plugin_gem && !plugin_gem.options.empty?
```
ph: I think its would be cool to report the options in the logger.

colinsurprenant: there is not logger instance in the plugin manager

ph:
OK, I did some initial testing with the command line with this PR, work great so far.
I think we have a lot of collateral tests that will actually check the install command.

I would still prefer to have specific test for it like @purbon mentioned but they can be done in another PR.

Code LGTM.

lib/pluginmanager/install.rb
ph: err I should have said, it would be cool to report what the options this plugins is currently using the the puts output.

colinsurprenant:
ok, will open followup issue for integration tests.

lib/pluginmanager/install.rb
colinsurprenant: ahh yes. I'd prefer not too because the installed gem option are not just discarded in favour of the options in the Gemfile, they are actually merged in the Gemfile options but at this layer, we don't know yet the result of that merge. Very practically, I don't think the plugins manager ever pass options to install just yet so it can be assumed today that the current options in the Gemfile will be the final options but that could eventually not be the case ... so I don't think it's really worth investing too much time on this now, if we think it's really valuable we can take another look at that point?

colinsurprenant:
rebased and merged into master, 5.0 and 2.x

lib/pluginmanager/install.rb
ph: we agreed on zoom to not add this.

PR #5233, opened 2016-05-02
Release 5.0 alpha2

suyograo:


ph:
LGTM, I have double checked the GeoIP and kafka version they have their latest beta.

PR #5236, opened 2016-05-02
Update alpha note to use variable for version

dedemorton:
Changed so that the source uses a variable rather than hard coding version info.


suyograo:
LGTM

Issue #3414, opened 2015-06-10
[docs] Improve getting started docs/tutorial


suyograo:
- [x] Split the getting started to a short version and long (advanced) version
- [x] Update this content with plugins
- [x] Add more examples for sending logs and non-log data like tweets to destinations
- [x] Remove any old, outdated references
- [x] Add glossary of common terms

suyograo:
@dedemorton as we spoke, please close this after we add a glossary of terms

PR #5237, opened 2016-05-03
Add glossary to the Logstash doc

dedemorton:
This PR completes the changes listed in issue #3414.

### DIFF
--- a/docs/static/glossary.asciidoc
+++ b/docs/static/glossary.asciidoc
@@ -1,0 +1,79 @@
+[[glossary]]
+== Glossary of Terms
+
+[[glossary-metadata]]@metadata ::
+  A special field for storing content that you don't want to include in output <<glossary-event,events>>. For example, the `@metadata`
+  field is useful for creating transient fields for use in <<glossary-conditional,conditional>> statements.
+    
+[[glossary-codec-plugin]]codec plugin::
+  A Logstash <<glossary-plugin,plugin>> that changes the data representation of an <<glossary-event,event>>. Codecs are essentially stream filters that can operate as part of an input or output. Codecs enable you to separate the transport of messages from the serialization process. Popular codecs include json, msgpack, and plain (text).
+  
+[[glossary-conditional]]conditional::
+  A control flow that executes certain actions based on whether a statement (also called a condition) is true or false. Logstash supports `if`, `else if`, and `else` statements. You can use conditional statements to apply filters and send events to a specific output based on conditions that you specify. 
+    
+[[glossary-event]]event::
+	A single unit of information, containing a timestamp plus additional data. An event arrives via an input, and is subsequently parsed, timestamped, and passed through the Logstash <<glossary-pipeline,pipeline>>.
+    
+[[glossary-field]]field::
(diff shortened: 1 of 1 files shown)


untergeek:
I like it! LGTM™

suyograo:
@dedemorton Can we add 
`@metadata`: https://www.elastic.co/guide/en/logstash/current/event-dependent-configuration.html#metadata
`conditionals`: https://www.elastic.co/guide/en/logstash/current/event-dependent-configuration.html#conditionals
`field reference`: https://www.elastic.co/guide/en/logstash/current/event-dependent-configuration.html#logstash-config-field-references

dedemorton:
@suyograo I've added the terms plus a few more. I had a "face palm" moment this afternoon when I realized that there was another [glossary file in the repo](https://github.com/elastic/logstash/blob/master/docs/static/logstash-glossary.asciidoc) (wasn't hooked up to the TOC). IMHO, the old glossary defines a lot of terminology that isn't Logstash-specific, so I didn't include those terms, but I did try to merge (and massage) the definitions that I thought we would want to preserve. I didn't include industry terminology or descriptions of other products because that's why we have google, right? I think a shorter glossary that focuses on Logstash terminology is preferable.

BTW, I think that "worker" is better than "pipeline worker" but I can change it if you feel that pipeline worker is necessary.

docs/static/glossary.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+[[glossary-filter-plugin]]filter plugin::
+  A Logstash <<glossary-plugin,plugin>> that performs intermediary processing on an event. Typically, filters act upon
+  event data after it has been ingested via inputs, by mutating, enriching, and/or modifying the data according to
+  configuration rules. Filters are often applied conditionally depending on the characteristics of the event. Popular
+  filter plugins include grok, mutate, drop, clone, and geoip. Filter stages are optional.
+  
+[[glossary-gem]]gem::
+  A self-contained package of code that's hosted on https://rubygems.org[RubyGems.org]. You can use the Logstash plugin
+  manager to manage Lostash gems.
```
suyograo: typo `Lostash`

Also can you add that Logstash plugins are packaged as Ruby Gems? Thats the connection here.

suyograo:
@dedemorton face palm indeed. I completely forgot we had an old one lying around as well. 

Another common term used in Logstash context: 
"Message Buffer/Queue": A message buffer is an external software (Redis, Kafka, RabbitMQ) that stores messages from the Logstash shipper instance as an intermediate store, waiting to be processed by the Logstash indexer instance. This typically helps to manage bursts of data.

dedemorton:
@suyograo We also use the term "broker" in the docs to mean (I think) the same thing. I'm a bit concerned about defining a message buffer (or queue) as external software because it's a general term that we might also want to use to describe internal queuing (for example, when we talk about things like persistent queues). How about this instead:

message broker:  Also referred to as a _message buffer_ or _message queue_, a message broker is external software (such as Redis, Kafka, or RabbitMQ) that stores messages from the Logstash shipper instance as an intermediate store, waiting to be processed by the Logstash indexer instance.

suyograo:
LGTM, please merge

PR #5241, opened 2016-05-03
Update release notes for 5.0 alpha2

dedemorton:
A couple questions:
* We say "the init script" and we talk about "init scripts" (plural). Which should I use? 
* Would you rather see the release notes for all the alpha releases in the same topic? If so, I can add float tags. If not, I can let the doc build chunk them into separate topics. Here's what the default (no float tags) looks like. Having separate topics might  make it a tad bit easier for users to scan for changes that are relevant to them.

releasenotes

### DIFF
--- a/docs/static/releasenotes.asciidoc
+++ b/docs/static/releasenotes.asciidoc
@@ -1,5 +1,48 @@
 [[releasenotes]]
-== Logstash 5.0-alpha1 Release Notes
+== Release Notes
+
+This section summarizes the changes in each release.
+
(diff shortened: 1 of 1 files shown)


suyograo:
> We say "the init script" and we talk about "init scripts" (plural). Which should I use?

It is `init script` not plural

> If not, I can let the doc build chunk them into separate topics

IMO separate is better so users can see what change for the exact release

suyograo:
LGTM

PR #5243, opened 2016-05-04
Make Java 8 minimum version for LS 5.0

### LINKED ISSUE #5242: Make Java 8 strict requirement for Logstash
If Logstash is attempted to run under Java 7, we should bail out with a warning. Only Java 8 should be supported in 5.0

suyograo:
Fixes #5242

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -149,8 +149,12 @@ class LogStash::Runner < Clamp::Command
       return 1
     end
 
-    # Print a warning to STDERR for bad java versions
-    LogStash::Util::JavaVersion.warn_on_bad_java_version
+    # Exit on bad java versions
+    java_version = LogStash::Util::JavaVersion.version
+    if LogStash::Util::JavaVersion.bad_java_version?(java_version)
+      $stderr.puts "Java version 1.8.0 or later is required. (You are running: #{java_version})"
+      return 1
+    end  
 
     LogStash::ShutdownWatcher.unsafe_shutdown = unsafe_shutdown?
     LogStash::ShutdownWatcher.logger = @logger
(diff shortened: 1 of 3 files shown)


logstash-core/lib/logstash/runner.rb
```diff
@@ -149,8 +149,12 @@ def execute
...
     end
 
-    # Print a warning to STDERR for bad java versions
-    LogStash::Util::JavaVersion.warn_on_bad_java_version
+    # Exit on bad java versions
+    java_version = LogStash::Util::JavaVersion.version
+    if LogStash::Util::JavaVersion.bad_java_version?(java_version)
+      $stderr.puts "Java version 1.8.0 or later is required. (You are running: #{java_version})"
+      return 1
+    end
```
ph: Shouldn't we use `signal_usage_error` and `I18n` for the string?

Like this https://github.com/elastic/logstash/pull/5243/files#diff-da023432533015a65b7931c1c17d4c60R177

`signal_usage_error` will exit with the right status code.

ph: We might want to delete the `warn_on_bad_java_version` since we wont use it anymore, IIRC we had also specs covering it.

https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/util/java_version.rb#L10

suyograo: @ph `signal_usage_error` is a clamp method.. Typically that is used for CLI usage issues. Java requirement is more of a fatal error. See similar usages: https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/runner.rb#L147

ph: good reasoning

ph:
LGTM

PR #5245, opened 2016-05-04
Remove oldlogstashjson from the default plugin list

ph:
This plugins was used to migrate from logstash json schema 0 to 1.


jordansissel:
LGTM

Issue #5235, opened 2016-05-02
Logstash Auto-Reload may have resource leak


kenwdelong:
I'm running a small ELK stack in a Docker container.  When I updated to Logstash 2.3.x, I started to experience resource leaks.

![image](https://cloud.githubusercontent.com/assets/6744405/14970266/628d97d6-107c-11e6-9c7d-a8cbcf3c191f.png)

I finally traced it down to the auto-reload feature for Logstash, which I had unknowingly merged from the upstream of my fork.

With auto reload on, within 24 hours Logstash was running at 50% CPU and 16% memory on my EC2 t2.medium instance.  Without it, Logstash stays at about 1-2% CPU and 5.5% memory indefinitely.  My ELK stack takes in about 787,000 messages per day. Nothing else runs on that instance.  I'm using 2016-03 Amazon Linux and Docker 1.9.1.

You can see what I'm running successfully here: https://github.com/kenwdelong/elk-docker/tree/ES232LS232K450, as compared to what I was running before unsuccessfully: https://github.com/kenwdelong/elk-docker/tree/ES231L231K450.  If you diff the commits you'll see the difference in logstash-init. (there are also minor version upgrades to ES and LS, but I confirmed that was not the problem).

jsvd:
Thanks for the detailed report @kenwdelong.

Could you post logs from logstash and a heapdump when logstash is consuming those resource? That would really help with debugging.

kenwdelong:
I don't think I can get a heap dump - I'd have to put the buggy Docker image back into production for 24h, and that's not really an option.  I did grab a thread dump when it was in a bad state.  I'll attach that here.  I didn't find anything interesting (at least to me).

[logstashStackTrace.txt](https://github.com/elastic/logstash/files/246129/logstashStackTrace.txt)

The Docker build that was bad is tagged here: https://hub.docker.com/r/kenwdelong/elk-docker/tags/ as ES231L231K450.  I could try spinning up an EC2 instance, and then just running netcat to pump in dummy data and see if I can reproduce it.

I created a new t2.small instance and ran these commands:
sudo su
yum update
yum install docker
service docker start
docker pull kenwdelong/elk-docker:ES231L231K450
docker run -d -p 5601:5601 -p 5000:5000 --name elk kenwdelong/elk-docker:ES231L231K450
echo '{  "@timestamp": 1461677555757, "@version": 1, message: "this is a test" }' > test.json
[...]

PR #5250, opened 2016-05-05
optimize resource usage when recompiling pipeline

jsvd:
* avoid usage of class variable
* avoid creating pipeline object if fetched config is same as before

fixes https://github.com/elastic/logstash/issues/5235

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -189,16 +189,20 @@ class LogStash::Agent
   # wrapped in @upgrade_mutex in the parent call `reload_state!`
   def reload_pipeline!(id)
     old_pipeline = @pipelines[id]
+    if old_pipeline.config_str == fetch_config(old_pipeline.original_settings)
+      @logger.debug("no configuration change for pipeline",
+                    :pipeline => id, :config => old_pipeline.config_str)
+      return
+    end
+
     new_pipeline = create_pipeline(old_pipeline.original_settings)
     return if new_pipeline.nil?
 
-    if old_pipeline.config_str == new_pipeline.config_str
-      @logger.debug("no configuration change for pipeline",
-                    :pipeline => id, :config => old_pipeline.config_str)
-    elsif new_pipeline.non_reloadable_plugins.any?
+    if new_pipeline.non_reloadable_plugins.any?
       @logger.error(I18n.t("logstash.agent.non_reloadable_config_reload"),
                     :pipeline_id => id,
                     :plugins => new_pipeline.non_reloadable_plugins.map(&:class))
+      return
     else
       @logger.warn("fetched new config for pipeline. upgrading..",
                    :pipeline => id, :config => new_pipeline.config_str)
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -165,12 +165,15 @@ class LogStash::Agent
     @collect_metric
   end
 
-  def create_pipeline(settings)
-    begin
-      config = fetch_config(settings)
-    rescue => e
-      @logger.error("failed to fetch pipeline configuration", :message => e.message)
-      return
+  def create_pipeline(settings, config=nil)
+
+    if config.nil?
+      begin
+        config = fetch_config(settings)
+      rescue => e
+        @logger.error("failed to fetch pipeline configuration", :message => e.message)
+        return
+      end
     end
 
     begin
@@ -189,13 +192,14 @@ class LogStash::Agent
   # wrapped in @upgrade_mutex in the parent call `reload_state!`
   def reload_pipeline!(id)
     old_pipeline = @pipelines[id]
-    if old_pipeline.config_str == fetch_config(old_pipeline.original_settings)
+    new_config = fetch_config(old_pipeline.original_settings)
+    if old_pipeline.config_str == new_config
       @logger.debug("no configuration change for pipeline",
-                    :pipeline => id, :config => old_pipeline.config_str)
+                    :pipeline => id, :config => new_config)
       return
     end
 
-    new_pipeline = create_pipeline(old_pipeline.original_settings)
+    new_pipeline = create_pipeline(old_pipeline.original_settings, new_config)
     return if new_pipeline.nil?
 
     if new_pipeline.non_reloadable_plugins.any?
--- a/logstash-core/lib/logstash/config/config_ast.rb
+++ b/logstash-core/lib/logstash/config/config_ast.rb
@@ -137,7 +137,10 @@ module LogStash; module Config; module AST
   class PluginSection < Node
     # Global plugin numbering for the janky instance variable naming we use
     # like @filter_<name>_1
-    @@i = 0
+    def initialize(*args)
+      super(*args)
+      @i = 0
+    end
 
     # Generate ruby code to initialize all the plugins.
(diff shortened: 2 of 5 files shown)


jordansissel:
Read the patch on my phone. Change looks good, but I haven't tested.

-Jordan

logstash-core/lib/logstash/agent.rb
```diff
@@ -165,12 +165,15 @@ def collect_metrics?
...
-      @logger.error("failed to fetch pipeline configuration", :message => e.message)
-      return
+  def create_pipeline(settings, config=nil)
+
+    if config.nil?
+      begin
+        config = fetch_config(settings)
+      rescue => e
+        @logger.error("failed to fetch pipeline configuration", :message => e.message)
+        return
```
guyboertje: `return` here is needed, but is it on line 183?

guyboertje:
LGTM (not tested, not sure there is a definitive test)

Question, is there a strategy to compare configs that are structurally the same but not byte for byte?

logstash-core/lib/logstash/agent.rb
jsvd: it doesn't need, I think in this case the visual similarity between the two blocks helps reason about the behaviour of the code. but I can remove if ya want :)

jsvd:
@guyboertje as soon as we have a better grasp of what our long term ast representation infrastructure will be, we can build a finer grained comparison with it

jsvd:
@guyboertje currently the test is monitor the garbage collection with yk or visual vm. A tip is to look for the count of org.jruby.runtime.ivars.StampedVariableAccessor

Also increasing the reload interval and testing throughput will show an increase scenarios of cpu starvation.

colinsurprenant:
@jsvd see [my last comment](https://github.com/elastic/logstash/issues/5235#issuecomment-217540812) in #5235 - maybe we could refactor the dynamic ivars and methods that way?

jsvd:
from my tests when reusing `@@i` + avoiding pipeline creation,the number of ivars is reduced dramatically.
The slow down will only be noticeable if the pipeline is completely changed about once an hour for a month.we don't even have enough plugins to trigger that many total changes.
thought experiment: a pipeline changing once an hour for a month, never reusing any of its 10 plugins, would create 10*24*31 = 10800 ivars

jsvd:
@colinsurprenant with this in perspective i'd rather not introduce more code with unknown behaviour. wdyt?

guyboertje:
@jsvd - re: your thought experiment, is there a way for LS main thread to test for this hidden ivar count? It occurs to me that, after this PR, reloading configs in a way that generates that many ivars - must be A) some automation gone wrong or B) a bug. Surely, it can't be as a result of natural DevOps behaviour.

jsvd:
@guyboertje unfortunatley no, there's nothing in ruby that exposes this growing table, afaik.
The only way we can measure is with a heapdump and count of instances

guyboertje:
@jsvd - Not even as a JRuby extension?

jsvd:
That might do it, I don't know. Honestly I don't think it is worth the effort/time/overhead.
The current behaviour happens for a known reason, the architecture that allows it was not designed with the concern of multiple pipelines in mind, this PR changes that

guyboertje:
I am not trying to make more work, only whether detecting this can be achieved in the future if we need to. 😄

jsvd:
Here it is, in JRuby it's possible: https://github.com/jruby/jruby/issues/3859#issuecomment-217868853

colinsurprenant:
@jsvd but can't we avoid dynamically creating ivars altogether by having one ivar (hash or struct or whichever container) to hold whatever we assign to theses ivars in the first place and provide Pipeline accessors to these values? It does not feel like such a big change and would get rid of this and actually make to code more explicit and less magic?

jsvd:
we definitely can, as per your suggestion.
As for the magic, it's subjective, in both scenarios we're generating code, I'd have to actually write the code and compare if the generated code is that much cleaner. Also, performance wise I'm not sure of the impact of using the hash instead of instance variables.
There's no question this whole compilation process needs to get simpler but I don't think it's the scope of this PR: it's preventing the ivar explosion in a singleton class.

colinsurprenant:
@jsvd +1 this is definitely a good improvement and we will revisit the whole pipeline creation in the future. 
Is there a pragmatic way to have specs to assess the `@@i` vs `@i` strategy?

jsvd:
you mean the compilation correctness or absense of lookup explosion? the first should be taken care of by existing tests, the latter...we can perhaps create a test with https://github.com/jruby/jruby/issues/3859#issuecomment-217868853 ?

colinsurprenant:
I am not sure but I would try to avoid testing very specific implementations details and try to test against the required behaviour. I think it will be hard here though.

jsvd:
the problem is not closures being set to ivars, is just the plain creation of ivars, if we reuse ivars, the problem goes away.

colinsurprenant:
@jsvd yes, that's what I meant, the ivars assignments is the problem, not the closures themselves. What I am suggesting is that it *seems* fairly simple to completely move away from that strategy and not create ivars for the closures in the first place, no?

jsvd:
I don't mean to fight you on this, it's was just a matter of pragmatism as I've been mostly away this week and we need to fix this.
I agree that using a hash ivar will reduce issues, I am just, at this point, avoiding changing more than necessary to avoid other unpredictable behaviors.
Anyway, next week I'll format a patch with your idea and do some testing on it.

colinsurprenant:
@jsvd absolutely no fight 💪  💪  😎  
I am sure we can resolve this like perfect gentleman by doing a proper standoff 😆

colinsurprenant:
Finally, instead of trying to explain my idea I created a quick & dirty example in https://github.com/elastic/logstash/pull/5299 - if this makes sense feel free to reimplement here as   you wish!

jsvd:
Added ivar reduction strategy by using a hash to store generated objects, code from @colinsurprenant 
I've tested manually and did some performance check, and all seems good, no drop or increase in eps

colinsurprenant:
LGTM

jsvd:
closed by:


[logstash:master] 6 new commits by Joao Duarte and 1 other:
d163917  avoid class variable during pipeline compilation - Joao Duarte
60c04c2  avoid creating pipeline if fetched config is the same as previous - Joao Duarte
c8e8bc4  reduce overhead of create_pipeline if config is known - Joao Duarte
58bc205  use closure table instead of ivars - Colin Surprenant
afe0e16  avoid dynamic use of singleton methods in pipeline compilation - Joao Duarte Show more...

[08:49] 
[logstash:5.0] 6 new commits by Joao Duarte and 1 other:
fb40dee  avoid class variable during pipeline compilation - Joao Duarte
c79ca77  avoid creating pipeline if fetched config is the same as previous - Joao Duarte
723046f  reduce overhead of create_pipeline if config is known - Joao Duarte
98abb22  use closure table instead of ivars - Colin Surprenant
ce473ab  avoid dynamic use of singleton methods in pipeline compilation - Joao Duarte Show more...

[08:49] 
[logstash:2.3] 1 new commit by Joao Duarte:
7cc3171  reduce overhead of pipepile verification/compilation - Joao Duarte

[08:49] 
[logstash:2.x] 1 new commit by Joao Duarte:
3106f3c  reduce overhead of pipepile verification/compilation - Joao Duarte

PR #5251, opened 2016-05-06
implementation of the logstash.yml settings file

jsvd:

So the setting precedence is, from weakest to strongest: default settings < yaml settings < cli settings

TODO
- [x] make conf/logstash.yml location configurable using env var CONF_DIR or similar
- [x] tests for Settings class
- [x] tests for Setting class
- [ ] correct documentation


jsvd:
@suyograo the logstash.yml currently it's in `::File.join(LOGSTASH_HOME, "conf", "logstash.yml")`
as noted in the TODO in the issue, I'd still like to have a proper environment variable just for the conf directory

conf/logstash.yml
```diff
@@ -0,0 +1,105 @@
...
+# How often to check if the pipeline configuration has changed (in seconds)
+#
+# config.reload_interval", Numeric, 3),
+#
+# ------------ Metrics Settings --------------
+#
+# Collect metrics?
+# WARNING: Enabling this has a performance impact
+#
+# metric.collect: false
```
ph: Metric are on by default and cannot be turned it off.
better to remove that comment

lib/bootstrap/environment.rb
```diff
@@ -56,6 +57,31 @@ def pattern_path(path)
...
+    puts "Warn: option \"#{argv.first}\" is deprecated, use \"-i #{argv.first}\" or \"--interactive=#{argv.first}\" instead"
+    ["--interactive", argv.first]
+  else
+    # The Clamp library supports specifying the same argument multiple times
+    # and it keeps the last occurrence in an array. So in order for cli args
+    # to override the logstash.yml args, we can do `settings_from_yml + argv`
+    settings_from_yml = fetch_yml_settings(LogStash::Environment::SETTINGS_PATH)
+    settings_from_yml + argv
+  end
+end
```
ph: Theses methods are globally defined methods, shouldn't they be in their own module?

logstash-core/lib/logstash/environment.rb
```diff
@@ -1,7 +1,38 @@
...
+           BooleanSetting.new("debug", false),
+           BooleanSetting.new("debug.config", false),
+           BooleanSetting.new("verbose", false),
+           BooleanSetting.new("quiet", false),
+           BooleanSetting.new("version", false),
+           BooleanSetting.new("help", false),
+                  Setting.new("log.path", String, nil, false),
+                  Setting.new("web_api.http.host", String, "127.0.0.1"),
+                  Setting.new("web_api.http.port", Numeric, 9600),
+  ].each {|setting| SETTINGS.register(setting) }
```
ph: I am wondering if the `BooleanSetttings` should be defined in the Setting class, this would make the Settings acts as a module.

```ruby
class Setting
  attr_reader :name

  def initialize(name)
    @name = name
  end

  class Boolean < Setting
    def initialize(name)
      @name = "boolean-#{name}"
    end
  end
end

puts Setting.new("Bojourno").name
puts Setting::Boolean.new("adios").name
```

ph: I really like How clean it make the code.

logstash-core/lib/logstash/environment.rb
```diff
@@ -1,7 +1,38 @@
...
+
+  [
+                  Setting.new("node.name", String, Socket.gethostname),
+           BooleanSetting.new("config.allow_env", false),
+  ExistingFilePathSetting.new("config.path", nil, false),
+                  Setting.new("config.string", String, nil, false),
+           BooleanSetting.new("config.test", false),
+           BooleanSetting.new("config.auto_reload", false),
+                  Setting.new("config.reload_interval", Numeric, 3),
+           BooleanSetting.new("metric.collect", true) {|v| v == true }, # metric collection cannot be disabled
```
ph: This might be a comment of style, but I find it a bit weird to do a `.new() { }`, I havent seen that before in the context of  initialization. But I do agree it makes the code shorter and its valid ruby code. :)

logstash-core/lib/logstash/patches/clamp.rb
```diff
@@ -0,0 +1,47 @@
+require 'clamp'
+require 'logstash/environment'
+
+module Clamp
+  module Attribute
+    class Instance
+      def default_from_environment
+        # we don want uncontrolled var injection from the environment
```
ph: s/don/dont

logstash-core/lib/logstash/patches/clamp.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  # Create a subclass of Clamp::Command that enforces the use of
+  # LogStash::SETTINGS for setting validation
+  class StrictCommand < Command
+    class << self
+      include ::Clamp::Option::StrictDeclaration
+    end
+  end
+end
```
ph: neat hack

logstash-core/lib/logstash/runner.rb
```diff
@@ -394,4 +353,7 @@ def trap_sigint
     end
   end
 
-end # class LogStash::Runner
+  def setting(key)
+    @settings.get_value(key)
+  end
+end
```
ph: I expect we see this pattern used in a lot of places.
I which we had some kind of module or base classes to define theses useful methods.

logstash-core/lib/logstash/settings.rb
```diff
@@ -0,0 +1,194 @@
...
+module LogStash
+  class Settings
+
+    def initialize
+      @settings = {}
+    end
+
+    def register(setting)
+      if @settings.key?(setting.name)
+        raise ArgumentError.new("Setting \"#{setting.name}\" has already been registered as #{setting.inspect}")
```
ph: I would define a custom error that inherits from `ArgumentError` It believe it make things easier to debug.

logstash-core/lib/logstash/settings.rb
```diff
@@ -0,0 +1,194 @@
...
+
+    def get_subset(setting_regexp)
+      regexp = setting_regexp.is_a?(Regexp) ? setting_regexp : Regexp.new(setting_regexp)
+      settings = self.class.new
+      @settings.each do |setting_name, setting|
+        next unless setting_name.match(regexp)
+        settings.register(setting.clone)
+      end
+      settings
+    end
```
ph: handy method

logstash-core/lib/logstash/environment.rb
```diff
@@ -1,7 +1,38 @@
...
+                  Setting.new("ruby_shell", String, nil, false),
+           BooleanSetting.new("debug", false),
+           BooleanSetting.new("debug.config", false),
+           BooleanSetting.new("verbose", false),
+           BooleanSetting.new("quiet", false),
+           BooleanSetting.new("version", false),
+           BooleanSetting.new("help", false),
+                  Setting.new("log.path", String, nil, false),
+                  Setting.new("web_api.http.host", String, "127.0.0.1"),
+                  Setting.new("web_api.http.port", Numeric, 9600),
```
ph: I like how flexible your code is, maybe we should create up front concrete class fro `Numeric` and `String` since we will use them a lot?

ph:
@jsvd I really like the refactor that you did, I think this will open new features in our code!
Added small comments and would you like to rebase the code I would like to do some manual testing on it.

logstash-core/lib/logstash/environment.rb
jordansissel: +1 @ph's suggestion. We might want something like that for validation anyway since things like Port isn't really just a numeric, but a number between 1 and 65535 right?

jordansissel: If I squint really hard, the config validator has similar constraints (type and value validation, etc). Maybe we can have them use the same thing (eventually, not needed for v1 of this) ;P

jsvd: Agree, that was actually one of the purposes of structuring the code this way, so we can have higher level setting classes :+1:

lib/bootstrap/environment.rb
jsvd: :+1: doing it now

logstash-core/lib/logstash/environment.rb
ph: @jordansissel This was the the secret goal ;)

logstash-core/lib/logstash/runner.rb
```diff
@@ -12,129 +13,141 @@
...
+      match = settings_arg.match(/--settings.path=(.*)/)
+      settings_path = match[1]
+    elsif ENV['LS_SETTINGS_DIR']
+      settings_path = ENV['LS_SETTINGS_DIR']
+    else
+      settings_path = nil
+    end
+
+    LogStash::SETTINGS.set("settings.path", settings_path) if settings_path
+    LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("settings.path"))
```
ph: I see the usefulness of having `from_yaml` but could add a `load` method to `LogStash::Settings` that will do the same thing.

spec/bootstrap/environment_spec.rb
```diff
@@ -0,0 +1,6 @@
+# encoding: utf-8
+require "spec_helper"
+require "bootstrap/environment"
+
+describe LogStash::Environment do
+end
```
ph: Do we really need this file?

ph:
Jenkins test this please

ph:
Jenkins please test this

ph:
Changes look good but I have a lot of test faillings locally.

```
 64) JVM stats LogStash::Api::HotThreadsCommand#schema return hot threads information
     Failure/Error: report = do_request { subject.run }
     ArgumentError:
       Setting "config.reload.auto" hasn't been registered
     # ./logstash-core/lib/logstash/settings.rb:20:in `get_setting'
     # ./logstash-core/lib/logstash/settings.rb:52:in `set_value'
     # ./logstash-core/lib/logstash/settings.rb:65:in `merge'

  65) LogStash::Runner pipeline settings debug_config should allow overriding debug.config
     Failure/Error: subject.run("bin/logstash", args)
     SystemExit:
       exit
     # ./vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:137:in `run'
     # ./logstash-core/spec/logstash/runner_spec.rb:220:in `(root)'

[...]
Finished in 28.39 seconds (files took 4.22 seconds to load)
1844 examples, 67 failures

Failed examples:
Randomized with seed 46916
```

jsvd:
wow, I know a few tests are failing, but not 67..even jenkins showed only 7 http://build-eu-00.elastic.co/job/Logstash_PR/2833/

jsvd:
currently only 3 errors in http://build-eu-00.elastic.co/job/Logstash_PR/2838/console:

```
Failures:

  1) LogStash::Agent metrics after config reloading resets the metric collector
     Failure/Error: expect(value).to eq(new_config_generator_counter)
       
       expected: 50
            got: 7636
       
       (compared using ==)
     # ./logstash-core/spec/logstash/agent_spec.rb:394:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:248:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:243:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:180:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:173:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:150:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:115:in `run_with_threads'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:100:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:78:in `run'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:176:in `standard_exception_handling'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:75:in `run'

  2) LogStash::Pipeline when collecting metrics in the pipeline pipelines populates the filter metrics
     Failure/Error: expect(collected_metric[:stats][:pipelines][:main][:plugins][:filters][plugin_name][:events][metric_key].value).to eq(number_of_events)
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./logstash-core/spec/logstash/pipeline_spec.rb:626:in `(root)'
     # ./logstash-core/spec/logstash/pipeline_spec.rb:624:in `(root)'
     # ./logstash-core/spec/logstash/pipeline_spec.rb:623:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:248:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:243:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:180:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:173:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:150:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:115:in `run_with_threads'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:100:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:78:in `run'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:176:in `standard_exception_handling'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:75:in `run'

  3) LogStash::Pipeline when collecting metrics in the pipeline pipelines populates the output metrics
     Failure/Error: expect(collected_metric[:stats][:pipelines][:main][:plugins][:outputs][plugin_name][:events][:out].value).to eq(number_of_events)
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./logstash-core/spec/logstash/pipeline_spec.rb:633:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:248:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:243:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:180:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:173:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:150:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:115:in `run_with_threads'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:100:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:78:in `run'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:176:in `standard_exception_handling'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:75:in `run'

Finished in 42.73 seconds (files took 2 seconds to load)
1893 examples, 3 failures
```

spec/bootstrap/environment_spec.rb
jsvd: no, I can remove it..I just thought of keeping it to encourage future tests in this module :)

logstash-core/lib/logstash/runner.rb
jsvd: not sure what you mean with this, like:

```ruby
  class Settings
    def load
       LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("settings.path"))
    end
```
?

jsvd:
also, I've been frequently rebasing the code against master to avoid rot, so it's at most 1 or 2 commits off (already rebased against log-with-json path)

spec/bootstrap/environment_spec.rb
ph: leave it :P

logstash-core/lib/logstash/runner.rb
ph: Yes this is what I mean, but its only sugar. I guess.

On Wed, May 11, 2016 at 6:55 AM, João Duarte <notifications@github.com>
wrote:

> In logstash-core/lib/logstash/runner.rb
> <https://github.com/elastic/logstash/pull/5251#discussion_r62827835>:
>
> >      super(*args)
> >    end
> >
> > +  def run(args)
> > +    if i=args.find_index("--settings.path")
> > +      settings_path = args[i+1]
> > +    elsif settings_arg = args.find {|v| v.match(/--settings.path=/) }
> > +      match = settings_arg.match(/--settings.path=(.*)/)
> > +      settings_path = match[1]
> > +    elsif ENV['LS_SETTINGS_DIR']
> > +      settings_path = ENV['LS_SETTINGS_DIR']
> > +    else
> > +      settings_path = nil
> > +    end
> > +
> > +    LogStash::SETTINGS.set("settings.path", settings_path) if settings_path
> > +    LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("settings.path"))
>
> not sure what you mean with this, like:
>
>   class Settings
>     def load
>        LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("settings.path"))
>     end
>
> ?
>
> —
> You are receiving this because you modified the open/close state.
> <https://github.com/elastic/logstash/pull/5251/files/c2b8f0e86d84bbbffcaca61e30de25288ffa0f9a#r62827835>
>



-- 
--
ph
Software Engineer

ph:
Weird concerning the failures, I will run it again :(

ph:
OK, tests are resolved for me. this what happen when you work with artefact for too much times, the `logstash-core` gems are cached in the vendor directory.

ph:
OK, the only thing I would still apply is to use custom exceptions instead of the core one.

Other than that code LGTM, lets get another reviewer in.

jsvd:
will do. thank you very much @ph for the valuable feedback

lib/bootstrap/environment.rb
```diff
@@ -56,20 +56,13 @@ def pattern_path(path)
...
-
 # when launched as a script, not require'd, (currently from bin/logstash and bin/logstash-plugin) the first
 # argument is the path of a Ruby file to require and a LogStash::Runner class is expected to be
 # defined and exposing the LogStash::Runner#main instance method which will be called with the current ARGV
 # currently lib/logstash/runner.rb and lib/pluginmanager/main.rb are called using this.
 if $0 == __FILE__
   LogStash::Bundler.setup!({:without => [:build, :development]})
   require ARGV.shift
-  # TODO deprecate these arguments in the next major version. use -i only
-  if ARGV == ["irb"] || ARGV == ["pry"]
```
purbon: good catch to clean up here, but might be would have been great to keep the cleanups in a different PR. other than that, great you remove it and nothing againts :+1:

logstash-core/lib/logstash/agent.rb
```diff
@@ -20,29 +20,32 @@
...
-    @reload_interval = params[:reload_interval] || 3 # seconds
+    @config_loader = LogStash::Config::Loader.new(@logger)
+    @reload_interval = setting("config.reload.interval")
     @upgrade_mutex = Mutex.new
 
-    @collect_metric = params.fetch(:collect_metric, false)
-    setup_metric_collection
+    @collect_metric = setting("metric.collect")
+    @metric = create_metric_collector
+    @periodic_pollers = LogStash::Instrument::PeriodicPollers.new(metric)
```
purbon: can we might be wrap this also into a method? in my opinion periodic pollers are also part of the collector aka setup the hole metrics system.  might be you can share a bit more about your thoughts to make this change?

logstash-core/lib/logstash/agent.rb
```diff
@@ -262,4 +268,8 @@ def clean_state?
   def reset_collector
     LogStash::Instrument::Collector.instance.clear
   end
+
+  def setting(key)
```
purbon: I see who ways to access this settings, one when is from the ivar (using this function) and another when is not  using the ivar. This two methods looks confusing end of the day. what do you think about having them more "uniform"?

logstash-core/lib/logstash/environment.rb
```diff
@@ -1,7 +1,40 @@
 # encoding: utf-8
 require "logstash/errors"
+require "logstash/config/cpu_core_strategy"
+require "logstash/settings"
 
 module LogStash
+
+  [
+            Setting::String.new("node.name", Socket.gethostname),
```
purbon: should we fix the indentation on this file?

logstash-core/lib/logstash/environment.rb
```diff
@@ -1,7 +1,40 @@
...
+           Setting::Boolean.new("debug.config", false),
+           Setting::Boolean.new("verbose", false),
+           Setting::Boolean.new("quiet", false),
+           Setting::Boolean.new("version", false),
+           Setting::Boolean.new("help", false),
+            Setting::String.new("log.path", nil, false),
+           Setting::Boolean.new("log.json", false),
+            Setting::String.new("web_api.http.host", "127.0.0.1"),
+              Setting::Port.new("web_api.http.port", 9600),
+  ].each {|setting| SETTINGS.register(setting) }
```
purbon: Might be not for now, but what do you think about having a less mixup default values registration? a few proposals:

* Why not having this in their own location, might be some place inside the _settings_ module?
* I guess in most of the cases we could infer the right type from the data type, most clear case is the boolean case. Might be we can avoid having to explicitly use this `Setting::Boolean` classes.

I ask specially the first one, because it might get simpler for maintenance and also when we might like to introduce somethings like validation for values, or what I do really find awesome, the `did you mean ....? ` followup.

What do you think?

purbon: But I do understand why it required to be in the environment, however code wise I would probably have this default in their own house.

logstash-core/lib/logstash/patches/clamp.rb
```diff
@@ -0,0 +1,47 @@
+require 'clamp'
+require 'logstash/environment'
+
+module Clamp
+  module Attribute
+    class Instance
+      def default_from_environment
+        # we don't want uncontrolled var injection from the environment
```
purbon: good you added a few context comment 👍 , on the other side to make this even better,  the reason might obvious, but why don't we want it?

logstash-core/lib/logstash/patches/clamp.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  module Option
+
+    module StrictDeclaration
+
+      include Clamp::Attribute::Declaration
+
+      # Instead of letting Clamp set up accessers for the options
```
purbon: s/accessers/accessors

logstash-core/lib/logstash/patches/clamp.rb
```diff
@@ -0,0 +1,47 @@
...
+        define_method(option.read_method) do
+          LogStash::SETTINGS.get_value(option.attribute_name)
+        end
+      end
+    end
+  end
+
+  # Create a subclass of Clamp::Command that enforces the use of
+  # LogStash::SETTINGS for setting validation
+  class StrictCommand < Command
```
purbon: 😍

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -30,39 +28,26 @@ module LogStash; class Pipeline
...
-    :metric,
     :logger,
     :started_at,
     :thread,
     :config_str,
-    :original_settings
-
-  DEFAULT_OUTPUT_WORKERS = 1
-
-  DEFAULT_SETTINGS = {
```
purbon: I love the smell of red in a PR 👍

logstash-core/lib/logstash/agent.rb
purbon: might be this should be coming out of the same settings class.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -78,7 +63,7 @@ def initialize(config_str, settings = {})
     #
     # This need to be configured before we evaluate the code to make
     # sure the metric instance is correctly send to the plugin.
-    @metric = settings.fetch(:metric, Instrument::NullMetric.new)
+    @metric = Instrument::NullMetric.new
```
purbon: 👍 for this improvement.  might be would be good to add an small comment to provide context that metric it has to be initialized (if not willing to be empty) through the accessor.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -217,15 +192,15 @@ def start_workers
...
+      batch_size = @settings.get("pipeline.batch.size")
+      batch_delay = @settings.get("pipeline.batch.delay")
       max_inflight = batch_size * pipeline_workers
       @logger.info("Starting pipeline",
-                   :id => self.pipeline_id,
-                   :pipeline_workers => pipeline_workers,
-                   :batch_size => batch_size,
-                   :batch_delay => batch_delay,
-                   :max_inflight => max_inflight)
+                   "id" => self.pipeline_id,
```
purbon: _curious about_, any reason for this change from symbols to strings?

jsvd: to map closely to the settings, that are also strings.
The reason for settings to be strings, is to avoid the ugly parameter of `:"pipeline.batch.delay"`, since we need to use dots

logstash-core/lib/logstash/runner.rb
```diff
@@ -12,129 +13,146 @@
...
+    I18n.t("logstash.runner.flag.settings-dir"),
+    :attribute_name => "settings.dir",
+    :default => LogStash::SETTINGS.get_default("settings.dir")
 
   attr_reader :agent
 
   def initialize(*args)
     @logger = Cabin::Channel.get(LogStash)
-    @pipeline_settings ||= { :pipeline_id => "main" }
+    @settings = LogStash::SETTINGS
```
purbon: why having an ivar here is necessary? if we aim to simplify the length of the constant could be a reason 👍, might be having some accessor ?

logstash-core/lib/logstash/runner.rb
```diff
@@ -12,129 +13,146 @@
...
     super(*args)
   end
 
+  def run(args)
+    if i=args.find_index("--settings.dir")
```
purbon: what is `i`? can we might be use a more representative name?

purbon: should this be part of his own function? something like grasp/fetch settings location ?

logstash-core/lib/logstash/runner.rb
```diff
@@ -158,32 +176,34 @@ def execute
...
     if version?
       show_version
       return 0
     end
 
-    return start_shell(@ruby_shell, binding) if @ruby_shell
+    return start_shell(setting("interactive"), binding) if setting("interactive")
 
-    if config_string.nil? && config_path.nil?
+    @settings.format_settings.each {|line| @logger.log(line) }
```
purbon: If i understand this properly, this means we log all settings every time we run logstash? is this not doing too much?

logstash-core/lib/logstash/runner.rb
```diff
@@ -194,22 +214,9 @@ def execute
...
-                          :web_api_http_host => @web_api_http_host,
-                          :web_api_http_port => @web_api_http_port)
-
-    @agent.register_pipeline("main", @pipeline_settings.merge({
-                          :config_string => config_string,
-                          :config_path => config_path,
-                          :debug_config => debug_config?,
-                          :allow_env => allow_env?
-                          }))
+    @agent = create_agent(@settings)
```
purbon: 👍 for this simplification, lovely to see

logstash-core/lib/logstash/runner.rb
purbon: another place https://github.com/elastic/logstash/pull/5251/files#diff-58bacb92415b22a57bdf05dc7459f937R272 with very similar code, I guess having more uniform access to settings would be nice to have.

logstash-core/lib/logstash/settings.rb
purbon: +1

purbon: but also this might work, as the context is already set to settings, it all depends who might be the consumer of this exception.

logstash-core/lib/logstash/agent.rb
jsvd: imo it's just a matter of need depending on the class.
Usually the pattern would be `LogStash::SETTINGS.get(key)`, but to ease testing and avoid this direct access we pass the Settings instance as an argument, so that turns into `@settings.get(key)`, but in classes that make a very high use of settings like the runner I wanted to be less verbose, hence the accessor method, to reduce to `setting(key)`.

From this perspective, what do you suggest?

logstash-core/lib/logstash/environment.rb
jsvd: the indentation was on purpose, to align the settings names, to make it easier to read.

logstash-core/lib/logstash/environment.rb
jsvd: I don't think this list should live in settings.rb since that file just creates the concepts for settings handling. This block of text actually defines settings important for LogStash, before they're needed in the runner, which is the environment.

As for inferring, I'm not sure I understand. in a setting like `Setting::Boolean.new("debug", false)`, you want to just do `Setting.new("debug", false)` and infer it's a Setting::Boolean from the default value's class? I think we could do it, but it would be inconsistent with the settings that have no default value and we cannot infer those.

logstash-core/lib/logstash/patches/clamp.rb
jsvd: The reason is simply a matter of design, I built this PR so settings can be retrieved from the cli, then yaml file, then defaults. Also allowing env would make it a bit too complex, IMO.

Adding an extra comment describing this.

logstash-core/lib/logstash/runner.rb
jsvd: we use an accessor in runner, to do `setting(setting_name)`

logstash-core/lib/logstash/runner.rb
jsvd: I was unsure about this, I like it being explicit, but I'm ok with changing this to another log level.
What do you suggest, info?

docs/static/command-line-flags.asciidoc
```diff
@@ -41,42 +41,47 @@ Logstash has the following flags. You can use the `--help` flag to display this
...
--r, --[no-]auto-reload
+-r, --auto.reload
   Monitor configuration changes and reload the configuration whenever it is changed.
 
---allow-env
+--allow.env
   EXPERIMENTAL: Enable environment variable templating within configuration parameters.
   
---reload-interval RELOAD_INTERVAL
+--reload.interval RELOAD_INTERVAL
```
jordansissel: Should this (and others) match the same setting name? By this, I mean, should this flag be `--config.reload.interval` ?

logstash-core/lib/logstash/agent.rb
jordansissel: +1 on using an instance of Settings passed in (@settings) instead of a global LogStash::SETTINGS for reasons you stated (easier testing)

logstash-core/lib/logstash/environment.rb
jordansissel: If this style is undesirable, we could do this the same way Plugin config settings are given:

```ruby
setting "node.name", Setting::String, :default => Socket.gethostname
```

```ruby
def self.setting(name, type, **stuff)
  SETTINGS.register(klass.new(name, ...whtaever..))
end
```

logstash-core/lib/logstash/patches/clamp.rb
jordansissel: +1.

I think it might be worth allowing ${FOO} environment values in the config yaml itself, but I'm content to wait until after the first release of this (and some user feedback) to add that.

logstash-core/lib/logstash/runner.rb
```diff
@@ -12,129 +13,136 @@
...
   def initialize(*args)
     @logger = Cabin::Channel.get(LogStash)
-    @pipeline_settings ||= { :pipeline_id => "main" }
+    @settings = LogStash::SETTINGS
     super(*args)
   end
 
+  def run(args)
+    settings_path = fetch_settings_path(args)
+    LogStash::SETTINGS.set("settings.dir", settings_path) if settings_path
```
jordansissel: If we have @settings, why use LogStash::SETTINGS here?

logstash-core/lib/logstash/runner.rb
```diff
@@ -12,129 +13,136 @@
...
+    LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("settings.dir"))
+    super(*[args])
```
jordansissel: why `super(*[args])` and not `super` or `super(*args)` ? Feels weird. I might be missing something obvious though ;P

logstash-core/lib/logstash/runner.rb
```diff
@@ -244,10 +241,10 @@ def execute
   def show_version
     show_version_logstash
 
-    if [:info, :debug].include?(verbosity?) || debug? || verbose?
+    if debug? || verbose?
       show_version_ruby
       show_version_java if LogStash::Environment.jruby?
-      show_gems if [:debug].include?(verbosity?) || debug?
+      show_gems if debug? && verbose?
```
jordansissel: Is this right? showing gems only if both `--debug` and `--verbose` are passed

logstash-core/lib/logstash/runner.rb
```diff
@@ -407,4 +385,26 @@ def trap_sigint
     end
   end
 
-end # class LogStash::Runner
+  def setting(key)
+    @settings.get_value(key)
+  end
+
+  # where can I find the logstash.yml file?
+  # 1. look for a "--setings.dir path"
```
jordansissel: typo "settings.dir"

logstash-core/lib/logstash/runner.rb
```diff
@@ -12,129 +13,136 @@
...
 
   def initialize(*args)
     @logger = Cabin::Channel.get(LogStash)
-    @pipeline_settings ||= { :pipeline_id => "main" }
+    @settings = LogStash::SETTINGS
     super(*args)
   end
 
+  def run(args)
+    settings_path = fetch_settings_path(args)
```
jordansissel: what happens if this path doesn't exist?

logstash-core/lib/logstash/settings.rb
```diff
@@ -0,0 +1,241 @@
...
+        # so let's use the proc validator instead
+        super(name, ::Numeric, default, strict)
+      end
+    end
+
+    class Port < Setting
+      def initialize(name, default=nil, strict=true)
+        # Ruby doesn't have a single class to represent booleans
+        # so let's use the proc validator instead
+        super(name, ::Numeric, default, strict) {|value| value > 1 && value < 65535 }
```
jordansissel: >=1 and <= 65535

or, > 0 && < 65536

logstash-core/lib/logstash/settings.rb
```diff
@@ -0,0 +1,241 @@
...
+      def initialize(name, default=nil, strict=true)
+        # Ruby doesn't have a single class to represent booleans
+        # so let's use the proc validator instead
+        super(name, ::String, default, strict)
+      end
+    end
+
+    class Numeric < Setting
+      def initialize(name, default=nil, strict=true)
+        # Ruby doesn't have a single class to represent booleans
```
jordansissel: copy paste? Comment mentions booleans. SImilar for elsewhere in this file.

logstash-core/lib/logstash/settings.rb
```diff
@@ -0,0 +1,241 @@
...
+    end
+
+    def set?
+      @value_is_set
+    end
+
+    def set(value)
+      validate(value)
+      @value = value
+      @value_is_set = true unless @value_is_set
```
jordansissel: Why have the `unless ...` here?

jordansissel:
Reviewed the code, looking good so far! :)

Will test soon.

suyograo:
I created tar.gz using this PR, and urwrapped it in `/tmp`

```
suyog@machine:/tmp/logstash-5.0.0.dev$ bin/logstash version
Errno::ENOENT: No such file or directory - /tmp/logstash-5.0.0.dev/conf/logstash.yml
       read at org/jruby/RubyIO.java:3804
       read at org/jruby/RubyIO.java:3987
  read_yaml at /tmp/logstash-5.0.0.dev/vendor/local_gems/1889b1c9/logstash-core-5.0.0.dev-java/lib/logstash/settings.rb:100
  from_yaml at /tmp/logstash-5.0.0.dev/vendor/local_gems/1889b1c9/logstash-core-5.0.0.dev-java/lib/logstash/settings.rb:94
        run at /tmp/logstash-5.0.0.dev/vendor/local_gems/1889b1c9/logstash-core-5.0.0.dev-java/lib/logstash/runner.rb:142
        run at /tmp/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:132
     (root) at /tmp/logstash-5.0.0.dev/lib/bootstrap/environment.rb:66
```

We need to ship a default `logstash.yml` so packages can use them. FWIW, see https://github.com/elastic/elasticsearch/blob/master/distribution/src/main/resources/config/elasticsearch.yml

docs/static/command-line-flags.asciidoc
suyograo: +1

logstash-core/lib/logstash/runner.rb
```diff
@@ -12,129 +13,136 @@
...
-  option ["-b", "--pipeline-batch-size"], "SIZE",
-         I18n.t("logstash.runner.flag.pipeline-batch-size"),
-         :attribute_name => :pipeline_batch_size,
-         :default => LogStash::Pipeline::DEFAULT_SETTINGS[:pipeline_batch_size]
+  option ["-u", "--pipeline.batch.delay"], "DELAY_IN_MS",
+    I18n.t("logstash.runner.flag.pipeline-batch-delay"),
+    :attribute_name => "pipeline.batch.delay",
+    :default => LogStash::SETTINGS.get_default("pipeline.batch.delay"), &:to_i
+
+  option ["--[no-]pipeline.unsafe_shutdown"], :flag,
```
suyograo: Can we remove the `--no` prefix: `--[no-]pipeline.unsafe_shutdown`. IMO its super confusing. When you are in `logstash.yml` would you be setting:

`pipeline.unsafe_shutdown`: [true|false]

Why not flip the CLI args to take `true` and `false`. 

Ideally it would be nice to be consistent with CLI long form option and the setting in `logstash.yml`. WDYT? This way a user who is testing stuff (wants transient stuff, so uses the CLI) does not get confused

`bin/logstash --pipeline.unsafe.shutdown true`

logstash-core/lib/logstash/runner.rb
```diff
@@ -12,129 +13,136 @@
...
-         :attribute_name => :pipeline_batch_delay,
-         :default => LogStash::Pipeline::DEFAULT_SETTINGS[:pipeline_batch_delay]
+  # Plugins Settings
+  option ["-p", "--plugin.paths"] , "PATH",
+    I18n.t("logstash.runner.flag.pluginpath"),
+    :multivalued => true, :attribute_name => "plugin.paths",
+    :default => LogStash::SETTINGS.get_default("plugin.paths")
 
+  # Logging Settings
   option ["-l", "--log"], "FILE",
```
suyograo: Should be "--log.path"

suyograo:
Output of `bin/logstash --help` (long form options) does not match what is in `logstash.yml`

suyograo:
Also, why do we need to log all the settings when LS starts up? IMO, its unnecessary. We are building APIs to interrogate this info.

```
suyog@machine:/tmp/logstash-5.0.0.dev$ bin/logstash -e 'input { stdin {} } output { stdout {} }'
-------- Logstash Settings (* means modified) ---------
node.name: "Suyogs-MBP-2"
config.allow_env: false
*config.string: "input { stdin {} } output { stdout {} }"
config.test: false
config.reload.auto: false
config.reload.interval: 3
metric.collect: true
settings.dir: "/tmp/logstash-5.0.0.dev/conf"
pipeline.id: "main"
*pipeline.workers: 4 (default: 8)
pipeline.output.workers: 1
pipeline.batch.size: 125
pipeline.batch.delay: 5
pipeline.unsafe_shutdown: false
plugin.paths: []
debug: false
debug.config: false
verbose: false
quiet: false
version: false
help: false
log.json: false
web_api.http.host: "127.0.0.1"
web_api.http.port: 9600
```

logstash-core/lib/logstash/environment.rb
```diff
@@ -1,7 +1,40 @@
...
+           Setting::Boolean.new("debug", false),
+           Setting::Boolean.new("debug.config", false),
+           Setting::Boolean.new("verbose", false),
+           Setting::Boolean.new("quiet", false),
+           Setting::Boolean.new("version", false),
+           Setting::Boolean.new("help", false),
+            Setting::String.new("log.path", nil, false),
+           Setting::Boolean.new("log.json", false),
+            Setting::String.new("web_api.http.host", "127.0.0.1"),
+              Setting::Port.new("web_api.http.port", 9600),
```
arnonhongklay: Snake indent??

suyograo:
@jsvd nice work! I tested few scenarios (override rules, validation in yml, etc) and they all work fine.

logstash-core/lib/logstash/runner.rb
jsvd: just because doing

def m(*args)

wil wrap args in an array, so that you can then do super(*args)

but in this case we want to work with the actual args, so I removed the * from the argument, so args isn't wrapped.
This means that doing super(*args) will fail since it's not wrapped, hence super(*[args])

logstash-core/lib/logstash/runner.rb
jsvd: that's a good point, I need to check how to do it, maybe just forget about clamp flags and make them a normal flag, and have the Settings::Boolean class coerce either "true" or "false" to the ruby classes, and fail otherwise

jsvd:
Addressed most questions, except the change of boolean flags to a --flag=true formar

ph:
Woa, @jsvd I've tested a few scenario and they look to work fine, awesome job. its worth it.

jsvd:
I've rebased this PR against master to catch up with the removal of the allow_env flag

logstash-core/lib/logstash/runner.rb
suyograo: +1 to that approach

suyograo:
@jsvd did you see my comment about including the `logstash.yml` in https://github.com/elastic/logstash/blob/master/rakelib/artifacts.rake#L5. Otherwise packages won't work

jsvd:
@suyograo about your comment on boolean flags, what do we do with simple flags like -V (version) and --debug? I'd hate to see a user have to type `bin/logstash --version=true` or even --debug=true

One thing we can do is ensure all boolean flags are pharsed so that their default is false, this way to enable a boolean flag you only need to write the flag's name, e.g. --debug, and you rarely have to disable them with the "ugly" --no-debug

suyograo:
> what do we do with simple flags like -V (version) and --debug? I'd hate to see a user have to type bin/logstash --version=true or even --debug=true

IMO, we should leave `version` and `--debug` as is in the CLI and not port them to `settings.yml` file. 

> One thing we can do is ensure all boolean flags are pharsed so that their default is false, this way to enable a boolean flag you only need to write the flag's name, e.g. --debug, and you rarely have to disable them with the "ugly" --no-debug

+1. I like this.

rakelib/artifacts.rake
```diff
@@ -13,6 +13,7 @@ namespace "artifact" do
       "lib/pluginmanager/**/*",
       "patterns/**/*",
       "vendor/??*/**/*",
+      "conf/logstash.yml",
```
suyograo: Lets change this to `config/logstash.yml` to match ES.

jsvd:
this PR page is getting way too slow from the change + ~100 comments. I can't do much about the change size but I can reopen a PR to declutter the code from comments, as most have been addressed

PR #5254, opened 2016-05-06
Update default plugins for 5.0

suyograo:
5.0 is a chance for us to review and remove any plugins that we don't have to bundle by default. These plugins are still supported and maintained, but the default package should only contain most popular, most used plugins. 

**Removed plugins:**

logstash-filter-anonymize: Replaced with newer logstash-filter-fingerprint
logstash-input-zeromq/output: Not a lot of users. Most users seem to prefer rabbitmq, redis, kafka as broker
logstash-codec-netflow: not sure about this.
logstash-filter-multiline: deprecated
logstash-output-hipchat: not maintained. Still at v1 of API.
logstash-output-juggernaut: less usage

Please comment if you'd like to remove more/less

/cc @acchen97 @jordansissel

### DIFF
--- a/rakelib/default_plugins.rb
+++ b/rakelib/default_plugins.rb
@@ -4,7 +4,6 @@ module LogStash
     # plugins included by default in the logstash distribution
     DEFAULT_PLUGINS = %w(
       logstash-input-heartbeat
-      logstash-output-zeromq
       logstash-codec-collectd
       logstash-output-xmpp
       logstash-codec-dots
@@ -18,10 +17,8 @@ module LogStash
       logstash-codec-line
       logstash-codec-msgpack
       logstash-codec-multiline
-      logstash-codec-netflow
       logstash-codec-plain
       logstash-codec-rubydebug
-      logstash-filter-anonymize
       logstash-filter-checksum
       logstash-filter-clone
       logstash-filter-csv
@@ -34,7 +31,6 @@ module LogStash
       logstash-filter-json
       logstash-filter-kv
       logstash-filter-metrics
-      logstash-filter-multiline
       logstash-filter-mutate
       logstash-filter-ruby
       logstash-filter-sleep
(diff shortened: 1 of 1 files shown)


ph:
Concerning `logstash-codec-netflow`, we have a new community maintainer for it, and I can say he is really active.

`logstash-output-hipchat` Was completely rewritten by me a few month ago and It does support the v2 api and we use the official hipchat gem for it.

suyograo:
@ph we also need to measure if folks are actively using it. Remember these plugins can still be installed via `bin/logstash-plugin install`

ph:
@suyograo Agree, we need to find a way to make the rubygems stats meaningful, right now theses stats are obfuscated by our build process.

acchen97:
@suyograo I'm +1 on keeping netflow codec in for now.  You've removed most of the ones I had in mind, here's a few others proposals.

Remove:
- ganglia output
- gelf output
- nagios output
- nagios_nsca output
- null output

Add:
- translate filter
- webhdfs output

ph:
@acchen97 I am +1 with your proposed changes

jsvd:
I know this has been discussed already but I'm a bit on the fence with removing the multiline filter, as we don't have an alternative solution to dealing with multiline after codec, even if it means using a single input and -w 1

Do we have any statistics on how much this plugin is still used today?

Also, I suggest removing the checksum filter, should have been gone a long time ago https://github.com/logstash-plugins/logstash-filter-checksum/issues/2

acchen97:
@jsvd this is just for removing multiline filter from default package.

> Also, I suggest removing the checksum filter, should have been gone a long time ago logstash-plugins/logstash-filter-checksum#2
+1 on removing this from logstash-plugins org, removing [output-elasticsearch_http](https://github.com/logstash-plugins/logstash-output-elasticsearch_http) makes sense too.

suyograo:
@jsvd we are removing multiline filter from the defaults. Since it is deprecated, if someone has the edge case of using it, then can always install it

suyograo:
@ph @acchen97 can you guys take another look? Updated

acchen97:
LGTM

ph:
LGTM

PR #5258, opened 2016-05-09
Add artifact update test scenario to acceptance test

purbon:
This PR adds:

* The update artifact scenario to the acceptance test.
* Does small code refactoring, basically aggregating some concepts together.

Feature improvements for now:

* Simplify the reference to the latest logstash released. Now it needs more than 1 place to be updated.


purbon:
@tal your thoughts are very welcome here.

qa/acceptance/spec/shared_examples/updated.rb
```diff
@@ -0,0 +1,21 @@
+require_relative '../spec_helper'
+require          'logstash/version'
+
+RSpec.shared_examples "updated" do |logstash|
+
+  before (:all) { logstash.snapshot }
+  after  (:all) { logstash.restore }
+
+  it "can update on #{logstash.host}" do
+    logstash.install("5.0.0-alpha2", "./")
```
talevy: should this be versioned from a constant, file, or environment variable?

talevy:
@purbon. I am @talevy, not `@tal` :)

talevy:
small comments, otherwise LGTM

qa/acceptance/spec/shared_examples/updated.rb
purbon: I agree, I guess we should somehow make it available for most of the parts here, this is also need in the provision scripts, will see if I think of a better way to do it.

purbon: updated this to be read from the same config file as the vagrant platforms, I guess this is the best that works to have everything DRY.... next thing is obviously to name this file properly, but this can be done later.

purbon:
@talevy updated this PR with your comments, let me know if you are ok with it 👍

qa/sys/redhat/user_bootstrap.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+VERSION=`cat /vagrant/config/platforms.json | grep  latest | cut -d":" -f2 | sed 's/["\|,| ]//g'`
```
talevy: can we make this less error prone and more explicit?

purbon: Question is how? In pure bash this is the more generic way i could think
of, improvements welcome for sure!

talevy: I guess I am OK with it for now

talevy:
LGTM

Issue #5228, opened 2016-04-30
Acceptance test should be able to run one specific test easily


purbon:
A request coming from https://github.com/elastic/logstash/pull/5226#discussion_r61630135 ask for a very interesting point, having a simple option to run one single test. Acceptance test are very heavy test, so in case of debugging the whole process is faster for end user.

My current thoughts on implementing are going to the direction of using  the filtering options from rspec [[examples matching](https://www.relishapp.com/rspec/rspec-core/v/3-4/docs/command-line/example-option)]  [[group filtering](https://www.relishapp.com/rspec/rspec-core/v/3-4/docs/filtering)].

The interface should be in my opinion something like:

```
$ rake test:acceptance:example[example_string_name]
```

we could also enable in the feature to use the local `bin/rspec` shim with this test so the full rspec interface will be available.

Another interesting interface many people enjoy using is the one that lets examples to be run by line number. Do you find this interface also useful?

ph:
> Another interesting interface many people enjoy using is the one that lets examples to be run by line number. Do you find this interface also useful?

I  have never used filtering by example name, I find it harder to remember the name.

Run a test by line number is something I do everyday with all of our projects.
When a test fails they give you the line number and I use it until I have fixed the problem.

PR #5259, opened 2016-05-09
Migration of the CLI script to run under Vagrant VMs.

ph:
This is the First pass to part the `integration` test for the CLI, the
Logstash and the logstash-plugin command.

@purbon I have an issue one of the test is taking a really long time to complete the uninstall one and it seems to kill to the connection to the underlying VM can you investigate that?

### DIFF
--- a/qa/Gemfile
+++ b/qa/Gemfile
@@ -3,3 +3,5 @@ gem "runner-tool", :git => "https://github.com/purbon/runner-tool.git"
 gem "rspec", "~> 3.1.0"
 gem "rake"
 gem "pry", :group => :test
+gem "stud"
+gem "logstash-devutils"
--- a/qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
@@ -1,0 +1,58 @@
+# encoding: utf-8
+require_relative "../../../spec_helper"
+require "logstash/version"
+require "fileutils"
+require "logstash/devutils/rake"
+
+shared_examples "logstash install" do |logstash|
+  before(:each) do
+    logstash.install(LOGSTASH_VERSION)
+  end
+
+  after(:each) do
+    logstash.uninstall
+  end
+
+  describe "on #{logstash.hostname}" do
+    context "with a direct internet connection" do
+      context "when the plugin exist" do
+        context "from a local `.GEM` file" do
+          let(:gem_name) { "logstash-filter-qatest-0.1.1.gem" }
+          let(:gem_path_on_vagrant) { "/tmp/#{gem_name}" }
+          before(:each) do
+            logstash.download("https://rubygems.org/gems/#{gem_name}", gem_path_on_vagrant)
+          end
+
+          after(:each) { logstash.delete_file(gem_path_on_vagrant) }
+
+          it "successfully install the plugin" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install #{gem_path_on_vagrant}")
+            expect(command).to install_successfully
+            expect(logstash).to have_installed?("logstash-filter-dns")
+          end
+        end
+
+        context "when fetching a gem from rubygems" do
+          it "successfully install the plugin" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install logstash-filter-qatest")
+            expect(command).to install_successfully
+            expect(logstash).to have_installed?("logstash-filter-qatest")
+          end
+
+          it "allow to install a specific version" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install --version 0.1.0 logstash-filter-qatest")
+            expect(command).to install_successfully
+            expect(logstash).to have_installed?("logstash-filter-qatest", "0.1.0")
+          end
+        end
+      end
+
+      context "when the plugin doesnt exist" do
+        it "fails to install and report an error" do
+          command = logstash.run_command_in_path("bin/logstash-plugin install --no-verify logstash-output-impossible-plugin")
+          expect(command.stderr).to match(/Plugin not found, aborting/)
+        end
+      end
+    end
+  end
+end
--- a/qa/acceptance/spec/shared_examples/cli/logstash-plugin/list.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash-plugin/list.rb
@@ -1,0 +1,48 @@
+# encoding: utf-8
+require_relative "../../../spec_helper"
+require "logstash/version"
+require "fileutils"
+
+shared_examples "logstash list" do |logstash|
+  describe "logstash-plugin list on #{logstash.hostname}" do
+    before(:all) do
+      logstash.install(LOGSTASH_VERSION)
+    end
+
+    after(:all) do
+      logstash.uninstall
+    end
+
+    context "without a specific plugin" do
+      it "display a list of plugins" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list")
+        expect(result.stdout.split("\n").size).to be > 1
+      end
+
+      it "display a list of installed plugins" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list --installed")
+        expect(result.stdout.split("\n").size).to be > 1
+      end
+
+      it "list the plugins with their versions" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list --verbose")
+        result.stdout.split("\n").each do |plugin|
+          expect(plugin).to match(/^logstash-\w+-\w+\s\(\d+\.\d+.\d+\)/)
+        end
+      end
+    end
+
+    context "with a specific plugin" do
+      let(:plugin_name) { "logstash-input-stdin" }
+      it "list the plugin and display the plugin name" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list #{plugin_name}")
+        expect(result).to run_successfully_and_output(/^#{plugin_name}$/)
+      end
+
+      it "list the plugin with his version" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list --verbose #{plugin_name}")
+        expect(result).to run_successfully_and_output(/^#{plugin_name} \(\d+\.\d+.\d+\)/)
+      end
+    end
+  end
+end
--- a/qa/acceptance/spec/shared_examples/cli/logstash-plugin/update.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash-plugin/update.rb
@@ -1,0 +1,40 @@
+# encoding: utf-8
+require_relative "../../../spec_helper"
+require "logstash/version"
+
+shared_examples "logstash update" do |logstash|
+  describe "logstash update on #{logstash.hostname}" do
+    before :each do
+      logstash.install(LOGSTASH_VERSION)
+    end
+
+    after :each do
+      logstash.uninstall
+    end
+
+    let(:plugin_name) { "logstash-filter-qatest" }
+    let(:previous_version) { "0.1.0" }
+
+    before do
+      logstash.run_command_in_path("bin/logstash-plugin install --version #{previous_version} #{plugin_name}")
+      # Logstash wont update when we have a pinned versionin the gemfile so we remove them
+      logstash.replace_in_gemfile(',\s"0.1.0"', "")
+      expect(logstash).to have_installed?(plugin_name, previous_version)
+    end
+
+    context "update a specific plugin" do
+      it "has executed succesfully" do
+        cmd = logstash.run_command_in_path("bin/logstash-plugin update #{plugin_name}")
+        expect(cmd.stdout).to match(/Updating #{plugin_name}/)
+        expect(logstash).not_to have_installed?(plugin_name, previous_version)
+      end
+    end
+
+    context "update all the plugins" do
+      it "has executed succesfully" do
+        logstash.run_command_in_path("bin/logstash-plugin update")
+        expect(logstash).to have_installed?(plugin_name, "0.1.1")
+      end
+    end
+  end
+end
--- a/qa/rspec/commands/base.rb
+++ b/qa/rspec/commands/base.rb
@@ -4,8 +4,8 @@ require_relative "../../vagrant/helpers"
 module ServiceTester
 
   class Base
-
     LOCATION="/logstash-build".freeze
+    LOGSTASH_PATH="/opt/logstash/".freeze
 
     def snapshot(host)
       LogStash::VagrantHelpers.save_snapshot(host)
@@ -23,5 +23,44 @@ module ServiceTester
       service_manager(service, "stop", host)
     end
 
+    def run_command(cmd, host)
+      hosts = (host.nil? ? servers : Array(host))
+
+      response = nil
+      at(hosts, {in: :serial}) do |_host|
+        response = sudo_exec!(cmd)
+      end
+      response
+    end
+
+    def replace_in_gemfile(pattern, replace, host)
+      cmd = "/bin/env sed -i.sedbak 's/#{pattern}/#{replace}/' /opt/logstash/Gemfile"
+      run_command(cmd, host)
+    end
+
+    def run_command_in_path(cmd, host)
+      run_command("#{File.join(LOGSTASH_PATH, cmd)}", host)
+    end
+
+    def plugin_installed?(host, plugin_name, version = nil)
+      if version.nil?
+        cmd = run_command_in_path("bin/logstash-plugin list", host)
+        search_token = plugin_name
+      else
+        cmd = run_command_in_path("bin/logstash-plugin list --verbose", host)
+        search_token ="#{plugin_name} (#{version})"
+      end
+
+      plugins_list = cmd.stdout.split("\n")
+      plugins_list.include?(search_token)
+    end
+
+    def download(from, to, host)
+      run_command("wget #{from} -O #{to}", host)
+    end
+
+    def delete_file(path, host)
+      run_command("rm -rf #{path}", host)
+    end
   end
 end
--- a/qa/rspec/commands/debian.rb
+++ b/qa/rspec/commands/debian.rb
@@ -15,7 +15,7 @@ module ServiceTester
     end
 
     def package_for(version, base=ServiceTester::Base::LOCATION)
-      File.join(base, "logstash-#{version}_all.deb")
+      File.join(base, "logstash-#{version}.deb")
     end
 
     def install(package, host=nil)
@@ -60,6 +60,5 @@ module ServiceTester
         sudo_exec!("service #{service} #{action}")
       end
     end
-
   end
 end
--- a/qa/rspec/matchers/be_installed.rb
+++ b/qa/rspec/matchers/be_installed.rb
@@ -3,7 +3,6 @@ require 'rspec/expectations'
 require_relative '../helpers'
 
 RSpec::Matchers.define :be_installed do
-
   match do |subject|
     subject.installed?(subject.hosts, subject.name)
   end
--- a/qa/rspec/matchers/cli_matchers.rb
+++ b/qa/rspec/matchers/cli_matchers.rb
@@ -1,0 +1,31 @@
+# encoding: utf-8
+RSpec::Matchers.define :be_successful do
+  match do |actual|
+    actual.exit_status == 0
+  end
+end
+
+RSpec::Matchers.define :fail_and_output do |expected_output|
+  match do |actual|
+    actual.exit_status == 1 && actual.stderr =~ expected_output
+  end
+end
+
+RSpec::Matchers.define :run_successfully_and_output do |expected_output|
+  match do |actual|
+    (actual.exit_status == 0 || actual.exit_status.nil?) && actual.stdout =~ expected_output
+  end
+end
+
+RSpec::Matchers.define :have_installed? do |name,*args|
+  match do |actual|
+    version = args.first
+    actual.plugin_installed?(name, version)
+  end
+end
+
+RSpec::Matchers.define :install_successfully do
+  match do |cmd|
+    expect(cmd).to run_successfully_and_output(/Installation successful/)
+  end
+end
--- a/qa/Rakefile
+++ b/qa/Rakefile
@@ -28,7 +28,6 @@ namespace :qa do
 
       message  = "bootstraping all VM's defined in acceptance/Vagrantfile"
       message  = "#{message} for #{args[:platform]}: #{machines}" if !args[:platform].nil?
-      puts message
 
       LogStash::VagrantHelpers.destroy(machines)
       LogStash::VagrantHelpers.bootstrap(machines)
@@ -49,7 +48,7 @@ namespace :qa do
   namespace :acceptance do
     desc "Run all acceptance"
     task :all do
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/lib/*_spec.rb"]]))
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/lib/**/*_spec.rb"]]))
     end
 
     platforms.types.each do |type|
@@ -63,7 +62,7 @@ namespace :qa do
     desc "Run one single machine acceptance test"
     task :single, :machine do |t, args|
       ENV['LS_VAGRANT_HOST']  = args[:machine]
-      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/lib/*_spec.rb"]]))
+      exit(RSpec::Core::Runner.run([Rake::FileList["acceptance/spec/lib/**/**/*_spec.rb"]]))
     end
   end
 end
--- a/qa/acceptance/spec/lib/cli_operation_spec.rb
+++ b/qa/acceptance/spec/lib/cli_operation_spec.rb
@@ -1,0 +1,19 @@
+# encoding: utf-8
+require_relative "../spec_helper"
+require_relative "../shared_examples/cli/logstash/version"
+require_relative "../shared_examples/cli/logstash-plugin/install"
+require_relative "../shared_examples/cli/logstash-plugin/list"
+require_relative "../shared_examples/cli/logstash-plugin/uninstall"
+require_relative "../shared_examples/cli/logstash-plugin/update"
+
+describe "CLI operation" do
+  config = ServiceTester.configuration
+  config.servers.each do |address|
+    logstash = ServiceTester::Artifact.new(address, config.lookup[address])
+    it_behaves_like "logstash version", logstash
+    it_behaves_like "logstash install", logstash
+    it_behaves_like "logstash list", logstash
+    it_behaves_like "logstash uninstall", logstash
+    it_behaves_like "logstash update", logstash
+  end
+end
--- a/qa/acceptance/spec/shared_examples/cli/logstash-plugin/uninstall.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash-plugin/uninstall.rb
@@ -1,0 +1,35 @@
+# encoding: utf-8
+require_relative "../../../spec_helper"
+require "logstash/version"
+require "fileutils"
+
+shared_examples "logstash uninstall" do |logstash|
+  describe "logstash uninstall on #{logstash.hostname}" do
+    before :each do
+      logstash.install(LOGSTASH_VERSION)
+    end
+
+    after :each do
+      logstash.uninstall
+    end
+
+    context "when the plugin isn't installed" do
+      it "fails to uninstall it" do
+        result = logstash.run_command_in_path("bin/logstash-plugin uninstall logstash-filter-qatest")
+        expect(result.stderr).to match(/ERROR: Uninstall Aborted, message: This plugin has not been previously installed, aborting/)
+      end
+    end
+
+    # Disabled because of this bug https://github.com/elastic/logstash/issues/5286
+    xcontext "when the plugin is installed" do
+      it "succesfully uninstall it" do
+        result = logstash.run_command_in_path("bin/logstash-plugin install logstash-filter-qatest")
+        expect(logstash).to have_installed?("logstash-filter-qatest")
+
+        result = logstash.run_command_in_path("bin/logstash-plugin uninstall logstash-filter-qatest")
+        expect(result.stdout).to match(/^Uninstalling logstash-filter-qatest/)
+        expect(logstash).not_to have_installed?("logstash-filter-qatest")
+      end
+    end
+  end
+end
--- a/qa/acceptance/spec/shared_examples/cli/logstash/version.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash/version.rb
@@ -1,0 +1,22 @@
+# encoding: utf-8
+require_relative "../../../spec_helper"
+require "logstash/version"
+
+shared_examples "logstash version" do |logstash|
+  describe "logstash --version" do
+    before :all do
+      logstash.install(LOGSTASH_VERSION)
+    end
+
+    after :all do
+      logstash.uninstall
+    end
+
+    context "on #{logstash.hostname}" do
+      it "returns the right logstash version" do
+        result = logstash.run_command_in_path("bin/logstash --version")
+        expect(result).to run_successfully_and_output(/#{LOGSTASH_VERSION}/)
+      end
+    end
+  end
+end
--- a/qa/rspec/commands.rb
+++ b/qa/rspec/commands.rb
@@ -53,6 +53,30 @@ module ServiceTester
       client.uninstall(name, host)
     end
 
+    def run_command_in_path(cmd)
+      client.run_command_in_path(cmd, host)
+    end
+
+    def run_command(cmd)
+      client.run_command(cmd, host)
+    end
+
+    def plugin_installed?(name, version = nil)
+      client.plugin_installed?(host, name, version)
+    end
+
+    def download(from, to)
+      client.download(from, to , host)
+    end
+    
+    def replace_in_gemfile(pattern, replace)
+      client.replace_in_gemfile(pattern, replace, host)
+    end
+
+    def delete_file(path)
+      client.delete_file(path, host)
+    end
+
     def to_s
       "Artifact #{name}@#{host}"
     end
--- a/qa/rspec/commands/redhat.rb
+++ b/qa/rspec/commands/redhat.rb
@@ -15,7 +15,7 @@ module ServiceTester
     end
 
     def package_for(version, base=ServiceTester::Base::LOCATION)
-      File.join(base, "logstash-#{version}.noarch.rpm")
+      File.join(base, "logstash-#{version}.rpm")
     end
 
     def install(package, host=nil)
--- a/qa/rspec/matchers.rb
+++ b/qa/rspec/matchers.rb
@@ -1,3 +1,4 @@
 # encoding: utf-8
-require_relative './matchers/be_installed'
-require_relative './matchers/be_running'
+require_relative "./matchers/be_installed"
+require_relative "./matchers/be_running"
+require_relative "./matchers/cli_matchers"
--- a/qa/rspec/matchers/be_running.rb
+++ b/qa/rspec/matchers/be_running.rb
@@ -3,7 +3,6 @@ require 'rspec/expectations'
 require_relative '../helpers'
 
 RSpec::Matchers.define :be_running do
-
   match do |subject|
     subject.running?(subject.hosts, subject.name)
   end
--- a/qa/support/logstash-filter-qatest/logstash-filter-qatest-newer.gemspec
+++ b/qa/support/logstash-filter-qatest/logstash-filter-qatest-newer.gemspec
@@ -1,0 +1,25 @@
+Gem::Specification.new do |s|
+  s.name            = 'logstash-filter-qatest'
+  s.version         = '0.1.1'
+  s.licenses        = ['Apache License (2.0)']
+  s.summary         = "This plugin is only used in the acceptance test"
+  s.description     = "This plugin is only used in the acceptance test"
+  s.authors         = ["Elasticsearch"]
+  s.email           = 'info@elasticsearch.com'
+  s.homepage        = "http://www.elasticsearch.org/guide/en/logstash/current/index.html"
+  s.require_paths = ["lib"]
+
+  # Files
+  s.files = `git ls-files`.split($\)+::Dir.glob('vendor/*')
+
+  # Tests
+  s.test_files = s.files.grep(%r{^(test|spec|features)/})
+
+  # Special flag to let us know this is actually a logstash plugin
+  s.metadata = { "logstash_plugin" => "true", "logstash_group" => "filter" }
+
+  # Gem dependencies
+  s.add_development_dependency 'logstash-devutils'
+  s.add_runtime_dependency "logstash-core-plugin-api", "~> 2.0"
+end
+
--- a/qa/support/logstash-filter-qatest/logstash-filter-qatest-old.gemspec
+++ b/qa/support/logstash-filter-qatest/logstash-filter-qatest-old.gemspec
@@ -1,0 +1,24 @@
+Gem::Specification.new do |s|
+  s.name            = 'logstash-filter-qatest'
+  s.version         = '0.1.0'
+  s.licenses        = ['Apache License (2.0)']
+  s.summary         = "This plugin is only used in the acceptance test"
+  s.description     = "This plugin is only used in the acceptance test"
+  s.authors         = ["Elasticsearch"]
+  s.email           = 'info@elasticsearch.com'
+  s.homepage        = "http://www.elasticsearch.org/guide/en/logstash/current/index.html"
+  s.require_paths = ["lib"]
+
+  # Files
+  s.files = `git ls-files`.split($\)+::Dir.glob('vendor/*')
+
+  # Tests
+  s.test_files = s.files.grep(%r{^(test|spec|features)/})
+
+  # Special flag to let us know this is actually a logstash plugin
+  s.metadata = { "logstash_plugin" => "true", "logstash_group" => "filter" }
+
+  # Gem dependencies
+  s.add_development_dependency 'logstash-devutils'
+  s.add_runtime_dependency "logstash-core-plugin-api", "~> 2.0"
+end


qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+            command = run_command_in_path("bin/logstash-plugin install --version 2.1.2 logstash-filter-dns", logstash.host)
+            expect(command).to run_successfully_and_output(/Installation successful/)
+            expect(plugin_installed?(logstash.host, "logstash-filter-dns", "2.1.2")).to be_truthy
+          end
+        end
+      end
+
+      context "when the plugin doesnt exist" do
+        it "fails to install and report an error" do
+          # LOOK WITH PERE WHY EXITSTATUS IS NIL
```
ph: @purbon If you uncomment the check `fail_and_output` that does an explicit check on the status code, the remote command will return `NIL` instead of 1.

Gemfile.jruby-1.9.lock
```diff
@@ -1,5 +1,12 @@
 PATH
-  remote: ./logstash-core
+  remote: vendor/local_gems/9975e01f/logstash-core-event-java-5.0.0.dev-java
```
purbon: I guess this might be some kind of left over right?

Gemfile
```diff
@@ -2,16 +2,13 @@
 # If you modify this file manually all comments and formatting will be lost.
 
 source "https://rubygems.org"
-gem "logstash-core", "5.0.0.dev", :path => "./logstash-core"
-gem "logstash-core-event-java", "5.0.0.dev", :path => "./logstash-core-event-java"
-gem "logstash-core-plugin-api", "1.0.0", :path => "./logstash-core-plugin-api"
+gem "logstash-core", "5.0.0.dev", :path => "vendor/local_gems/ee2f9cf4/logstash-core-5.0.0.dev-java"
```
purbon: Same thing here, this looks like some kind of leftover from running all plugins test?

ph: Will revert that.. I guess I needed to generate an artifact to test the artifact. ;)

purbon: sure, I understand! thanks for reverting this 👍

qa/.ssh_config
```diff
@@ -0,0 +1 @@
+[{}]
```
purbon: this file should not be committed, we should add it to the gitignore. I propose we add one for the qa level, easy to manage.

qa/rspec/centos/commands.rb
```diff
@@ -0,0 +1,58 @@
+# encoding: utf-8
+module ServiceTester
+  class CentosCommands
```
purbon: This file was renamed to be redhat, as is really the _source_ of this kind of distros. I think you should rebase this PR, so you get all this things merged for you.

qa/rspec/cli/cli_tester.rb
```diff
@@ -0,0 +1,46 @@
+# encoding: utf-8
+require_relative "../../vagrant-helpers"
+require "logstash/devutils/rake" # provide download helper
+require "stud/temporary"
+
+module CliTester
+  DOWNLOAD_DIRECTORY = File.join(File.dirname(__FILE__), "..", "..", "vendor", "download")
+
+  def run_command(cmd, host)
+    hosts = (host.nil? ? servers : Array(host))
```
purbon: The code in the feature branch include a base class for commands, see https://github.com/elastic/logstash/blob/feature/acceptance_test/qa/rspec/commands/base.rb I see part of this code belonging there as is more like a generic.

purbon: I thinks this apply to most parts of this module.

qa/rspec/matchers/be_installed.rb
```diff
@@ -14,3 +13,9 @@
     subject.removed?(subject.hosts, subject.name)
   end
 end
+
+RSpec::Matchers.define :be_removed do
```
purbon: Why having this matcher twice? not sure I understand why this is being highlight here.

purbon:
* qa/rspec/matchers/be_succesful.rb
* qa/rspec/matchers/run_successfully_and_output.rb
*  qa/rspec/matchers/fail_and_output.rb

looks to me like similar/related matchers, can we somehow group them? I ask not to end up with M files all of them including very small and simple matchers. what do you think?

qa/sys/centos/bootstrap.sh
```diff
@@ -0,0 +1,17 @@
...
+yum update
+yum install -y java-1.8.0-openjdk-devel.x86_64
+
+##
+# Install logstash manually from a URL
+##
+BRANCH=${LOGSTASH_BRANCH:-'master'}
+BUILD_URL='https://s3-eu-west-1.amazonaws.com/build-eu.elasticsearch.org/logstash'
+URL="$BUILD_URL/$BRANCH/nightly/JDK8/logstash-latest-SNAPSHOT.rpm"
+wget --no-verbose $URL
```
purbon: I wonder why this have to happen in the privileged provision script (run under root), any idea? I would move them to be in the user_bootstrap.

on the other side, what is the benefit from using the snapshot or simply using the local shared directory? with the latest we're testing with 1 less download. What do you think?

qa/sys/ubuntu/bootstrap.sh
```diff
@@ -0,0 +1,17 @@
...
+apt-get update
+apt-get install -y openjdk-7-jdk
+
+##
+# Install logstash manually from a URL
+##
+BRANCH=${LOGSTASH_BRANCH:-'master'}
+BUILD_URL='https://s3-eu-west-1.amazonaws.com/build-eu.elasticsearch.org/logstash'
+URL="$BUILD_URL/$BRANCH/nightly/JDK8/logstash-latest-SNAPSHOT.deb"
+wget --no-verbose $URL
```
purbon: same comments as https://github.com/elastic/logstash/pull/5259/files#r62646255

qa/vagrant-helpers.rb
```diff
@@ -46,15 +46,15 @@ def self.run!(cmd)
   class VagrantHelpers
 
     def self.halt(machines="")
-      CommandExecutor.run!("vagrant halt #{machines.join(' ')}")
+      CommandExecutor.run!("vagrant halt #{Array(machines).join(' ')}")
```
purbon: good catch 👍 , https://en.wikipedia.org/wiki/Defensive_programming on top 💯

qa/vagrant-helpers.rb
```diff
@@ -66,6 +66,13 @@ def self.parse(lines)
       lines.each do |line|
         if line.match(/Host\s(.*)$/)
           host = { :host => line.gsub("Host","").strip }
+    end
+
+    def self.parse(lines)
```
purbon: not sure I understand this changes, got the file screw up somehow?

ph:
@purbon Thanks for your comments I will go through the changes.

qa/sys/centos/bootstrap.sh
ph: I think I messed up the rebase, I didnt not change anything about the sys/*.sh files in this PR>

qa/rspec/cli/cli_tester.rb
ph: I can move them to the `ServiceTester` class.

purbon: why not using the command base?

qa/sys/centos/bootstrap.sh
purbon: no problem!

qa/Gemfile
```diff
@@ -3,3 +3,5 @@ gem "runner-tool", :git => "https://github.com/purbon/runner-tool.git"
 gem "rspec", "~> 3.1.0"
 gem "rake"
 gem "pry", :group => :test
+gem "stud"
+gem "logstash-devutils"
```
ph: I wanted to reuse the download method available our library.

qa/rspec/cli/cli_tester.rb
ph: I believe we are talking about the same thing, but our naming doesn't match.

If you look at this file https://github.com/elastic/logstash/blob/feature/acceptance_test/qa/rspec/commands/base.rb#L3,
This file define the `ServiceTester:Base` class but the path reference a `Command` module.

When I have answered your comment I looked at the content of the file not the path.

ph:
I screw up the rebase on this PR, I am working on fixing it.

qa/vagrant-helpers.rb
ph: well the ci/acceptance.sh all was failling because of this.

ph:
I've pushed the rebase of this PR to the branch, I believe its much cleaner.

But there is test failing but I would like your opinion at how to properly fix them.
(Actually theses tests were running before we pushed the gems for 5.0.)

I am wondering if we should make a dummy gems to test the following cases, this would allow us to automate or better control the version and the gemspec.

- Updating from version A -> B (this is bit hard to test if the plugins has only one version available)
- unstalling (a gem with no other dependency, we might be able to use the stdin or the null object?)
- installing a specific version of a gem

I am asking this because using available gems make the task a lot harder to control the testing environment. What do you think would be the best way to proceed?

purbon:
on it, thanks for the rebase 👍

qa/vagrant-helpers.rb
purbon: not I wonder if this is something we want to do here or also improve the default param value. I guess both are going to be need.

qa/vagrant-helpers.rb
```diff
@@ -46,15 +46,15 @@ def self.run!(cmd)
   class VagrantHelpers
 
     def self.halt(machines="")
```
purbon: I would also change this to be 

```ruby
def self.halt(machines=[])
```

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      context "when the plugin exist" do
+        context "from a local `.GEM` file" do
+          let(:gem_path) { download_gem("logstash-filter-dns", "2.1.3") }
+          let(:gem_path_on_vagrant) { "/vagrant/vendor/download/logstash-filter-dns-2.1.3.gem" }
+          # after(:each) { FileUtils.rm_rf(gem_path) }
+
+          it "successfully install the plugin" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install #{gem_path_on_vagrant}")
+            expect(command).to run_successfully_and_output(/Installation successful/)
+            expect(logstash.plugin_installed?(logstash.host, "logstash-filter-dns")).to be_truthy
```
purbon: why is `logstash.host` required as parameter here? the artifact should already know about the host is running on right? also the notation it gets simplier. What do you think?

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+          end
+        end
+
+        context "when fetching a gem from rubygems" do
+          it "successfully install the plugin" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install logstash-filter-dns")
+            expect(command).to run_successfully_and_output(/Installation successful/)
+            expect(logstash.plugin_installed?(logstash.host, "logstash-filter-dns")).to be_truthy
```
purbon: same things as https://github.com/elastic/logstash/pull/5259/files#r62805705

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+          end
+
+          it "allow to install a specific version" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install --version 2.1.2 logstash-filter-dns")
+            expect(command).to run_successfully_and_output(/Installation successful/)
+            expect(logstash.plugin_installed?(logstash.host, "logstash-filter-dns", "2.1.2")).to be_truthy
```
purbon: also same as https://github.com/elastic/logstash/pull/5259/files#r62805705

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    logstash.uninstall
+  end
+
+  describe "on #{name}" do
+    context "with a direct internet connection" do
+      context "when the plugin exist" do
+        context "from a local `.GEM` file" do
+          let(:gem_path) { download_gem("logstash-filter-dns", "2.1.3") }
+          let(:gem_path_on_vagrant) { "/vagrant/vendor/download/logstash-filter-dns-2.1.3.gem" }
+          # after(:each) { FileUtils.rm_rf(gem_path) }
```
purbon: leftover?

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+          it "successfully install the plugin" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install #{gem_path_on_vagrant}")
+            expect(command).to run_successfully_and_output(/Installation successful/)
```
purbon: I wonder if this is something we should check inside the matcher? so in all places we check for successful install we have them grouped in a single matcher. This would be a new specific matcher for me, this one works also good. What do you think?

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+describe "CLI/logstash-plugin install" do
+  config = ServiceTester.configuration
+  config.servers.each do |address|
+    logstash = ServiceTester::Artifact.new(address, config.lookup[address])
+    it_behaves_like "logstash install", logstash
+  end
+end
+
+DOWNLOAD_DIRECTORY = File.join(File.dirname(__FILE__), "..", "..", "vendor", "download").freeze
+def download_gem(name, version)
```
purbon: might be something to wrap inside of a helpers file? so test are more just the specs definition.

qa/acceptance/spec/lib/cli/logstash-plugin/list_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    after :all do
+      logstash.uninstall
+    end
+
+    context "without a specific plugin" do
+      it "display a list of plugins" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list")
+        # expect(result.exit_status).to eq(0)
```
purbon: leftover

qa/acceptance/spec/lib/cli/logstash-plugin/list_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        expect(result.stdout.split("\n").size).to be > 1
+      end
+
+      it "display a list of installed plugins" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list --installed")
+        # expect(result.exit_status).to eq(0)
```
purbon: leftover

purbon:
I see this pattern repeated over the full set of test, for each you build this kind of setup, 

```ruby
describe "CLI/logstash-plugin list " do
   config = ServiceTester.configuration
   config.servers.each do |address|
     logstash = ServiceTester::Artifact.new(address, config.lookup[address])
     it_behaves_like "logstash list", logstash
   end
 end
```

what about having something like (see [spec/lib/artifact_operation_spec.rb](https://github.com/elastic/logstash/blob/feature/acceptance_test/qa/acceptance/spec/lib/artifact_operation_spec.rb)):

```ruby
describe "artifacts operation" do
  config = ServiceTester.configuration
  config.servers.each do |address|
    logstash = ServiceTester::Artifact.new(address, config.lookup[address])
    it_behaves_like "installable", logstash
    it_behaves_like "runnable",    logstash
  end
end
```

where we setup shared examples one way and the test composition another, what do you think? one benefit I see is having things not mixed up.

qa/acceptance/spec/lib/cli/logstash-plugin/list_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        expect(result.stdout.split("\n").size).to be > 1
+      end
+
+      it "list the plugins with their versions" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list --verbose")
+        result.stdout.split("\n").each do |plugin|
+          expect(plugin).to match(/^logstash-\w+-\w+\s\(\d+\.\d+.\d+\)/)
+        end
+        # expect(result.exit_status).to eq(0)
```
purbon: leftover?

qa/acceptance/spec/lib/cli/logstash-plugin/uninstall_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  after :each do
+    logstash.uninstall
+  end
+
+  context "when the plugin isn't installed" do
+    it "fails to uninstall it" do
+      result = logstash.run_command_in_path("bin/logstash-plugin uninstall logstash-filter-cidr")
+      expect(result.stderr).to match(/ERROR: Uninstall Aborted, message: This plugin has not been previously installed, aborting/)
+      # expect(result.exit_status).to eq(1)
```
purbon: leftover?

qa/acceptance/spec/lib/cli/logstash-plugin/uninstall_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  context "when the plugin is installed" do
+    it "succesfully uninstall it" do
+      result = logstash.run_command_in_path("bin/logstash-plugin install logstash-filter-dns")
+      expect(logstash.plugin_installed?(logstash.host, "logstash-filter-ruby")).to be_truthy
+
+      result = logstash.run_command_in_path("bin/logstash-plugin uninstall logstash-filter-dns")
+      expect(result.stdout).to match(/^Uninstalling logstash-filter-dns/)
+      # expect(result.exit_status).to eq(0)
```
purbon: leftover?

qa/acceptance/spec/lib/cli/logstash-plugin/update_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    logstash.run_command_in_path("bin/logstash-plugin install --version #{previous_version} #{plugin_name}")
+    cmd = logstash.run_command_in_path("bin/logstash-plugin list --verbose #{plugin_name}")
+    expect(cmd.stdout).to match(/#{plugin_name} \(#{previous_version}\)/)
+  end
+
+  context "update a specific plugin" do
+    subject { logstash.run_command_in_path("bin/logstash-plugin update #{plugin_name}") }
+
+    it "has executed succesfully" do
+      # expect(subject.exit_status).to eq(0)
```
purbon: leftover?

qa/acceptance/spec/lib/cli/logstash-plugin/update_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(subject.stdout).to match(/Updating #{plugin_name}/)
+    end
+  end
+
+  context "update all the plugins" do
+    subject { logstash.run_command_in_path("bin/logstash-plugin update") }
+
+    it "has executed succesfully" do
+      # expect(subject.exit_status).to eq(0)
```
purbon: leftover?

purbon:
> I've pushed the rebase of this PR to the branch, I believe its much cleaner.

awesome, much cleaner, thanks a lot!

> I am wondering if we should make a dummy gems to test the following cases, this would allow us to automate or better control the version and the gemspec

I'm +1 for this, test should be always using controlled scenarios, so I agree completely for this.

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
```diff
@@ -0,0 +1,74 @@
+# encoding: utf-8
+require_relative "../../../spec_helper"
+require "logstash/version"
+require "fileutils"
+require "logstash/devutils/rake"
```
purbon: I wonder why this is necessary, in my test if required it complains with:

```
NoMethodError: undefined method `desc' for main:Object
           (root) at /Users/purbon/.rbenv/versions/jruby-1.7.20/lib/ruby/gems/shared/gems/logstash-devutils-0.0.19-java/lib/logstash/devutils/rake/publish.rake:21
             load at org/jruby/RubyKernel.java:1091
           (root) at /Users/purbon/.rbenv/versions/jruby-1.7.20/lib/ruby/gems/shared/gems/logstash-devutils-0.0.19-java/lib/logstash/devutils/rake.rb:1
          require at org/jruby/RubyKernel.java:1072
           (root) at /Users/purbon/.rbenv/versions/jruby-1.7.20/lib/ruby/gems/shared/gems/logstash-devutils-0.0.19-java/lib/logstash/devutils/rake.rb:2
             load at org/jruby/RubyKernel.java:1091
           (root) at /Users/purbon/work/logstash/qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:1
```

purbon: I see you required it for the download version, might be we can do a finer grained requirement?

purbon: I know where the error is coming from, I'm not running this exemples with rake, so the whole rake context is not in. As a syntax like:  ```bundle exec rspec acceptance/spec/lib/cli/logstash-plugin/install_spec.rb``` should be possible I do recommend we use code here that is not rake dependant.

purbon: we should move most of the code within https://github.com/elastic/logstash-devutils/blob/master/lib/logstash/devutils/rake/vendor.rake  to a real class that can be propery reused, there is also some duplicated code in our logstash-core rake tasks.

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
purbon: what about having something like:

```ruby
  expect(logstash).to include_plugin?("logstash-filter-dns")
```

or 

```ruby
  expect(logstash).to have_installed?("logstash-filter-dns")
```

what do you think?

purbon:
I executed only the install_spec tests and got this errors:

```bash
skywalker% bundle exec rake qa:acceptance:single["ubuntu-1404"]
[Acceptance specs] running on ["ubuntu-1404"]
FFF.

Failures:

  1) CLI/logstash-plugin install behaves like logstash install on RSpec::ExampleGroups::CLILogstashPluginInstall::BehavesLikeLogstashInstall with a direct internet connection when the plugin exist from a local `.GEM` file successfully install the plugin
     Failure/Error: expect(command).to run_successfully_and_output(/Installation successful/)
       expected #<RunnerTool::Command:0x6dc2ac53 @host="127.0.0.1:2222", @cmd="/opt/logstash/bin/logstash-plugin install /vagrant/vendor/download/logstash-filter-dns-2.1.3.gem", @stderr="Plugin /vagrant/vendor/download/logstash-filter-dns-2.1.3.gem does not exist\nERROR: Installation aborted, verification failed for /vagrant/vendor/download/logstash-filter-dns-2.1.3.gem\n", @exit_code=1, @stdout="Validating /vagrant/vendor/download/logstash-filter-dns-2.1.3.gem\n"> to run successfully and output /Installation successful/
     Shared Example Group: "logstash install" called from ./acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:62
     # ./acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:26:in `(root)'
     # ./Rakefile:65:in `(root)'

  2) CLI/logstash-plugin install behaves like logstash install on RSpec::ExampleGroups::CLILogstashPluginInstall::BehavesLikeLogstashInstall with a direct internet connection when the plugin exist when fetching a gem from rubygems successfully install the plugin
     Failure/Error: expect(logstash.plugin_installed?(logstash.host, "logstash-filter-dns")).to be_truthy
       expected: truthy value
            got: false
     Shared Example Group: "logstash install" called from ./acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:62
     # ./acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:35:in `(root)'
     # ./Rakefile:65:in `(root)'

  3) CLI/logstash-plugin install behaves like logstash install on RSpec::ExampleGroups::CLILogstashPluginInstall::BehavesLikeLogstashInstall with a direct internet connection when the plugin exist when fetching a gem from rubygems allow to install a specific version
     Failure/Error: expect(logstash.plugin_installed?(logstash.host, "logstash-filter-dns", "2.1.2")).to be_truthy
     ArgumentError:
       wrong number of arguments calling `plugin_installed?` (3 for 2)
     Shared Example Group: "logstash install" called from ./acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:62
     # ./acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:41:in `(root)'
     # ./Rakefile:65:in `(root)'

Finished in 5 minutes 38 seconds (files took 2.14 seconds to load)
4 examples, 3 failures

Failed examples:

rspec ./acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:24 # CLI/logstash-plugin install behaves like logstash install on RSpec::ExampleGroups::CLILogstashPluginInstall::BehavesLikeLogstashInstall with a direct internet connection when the plugin exist from a local `.GEM` file successfully install the plugin
rspec ./acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:32 # CLI/logstash-plugin install behaves like logstash install on RSpec::ExampleGroups::CLILogstashPluginInstall::BehavesLikeLogstashInstall with a direct internet connection when the plugin exist when fetching a gem from rubygems successfully install the plugin
rspec ./acceptance/spec/lib/cli/logstash-plugin/install_spec.rb:38 # CLI/logstash-plugin install behaves like logstash install on RSpec::ExampleGroups::CLILogstashPluginInstall::BehavesLikeLogstashInstall with a direct internet connection when the plugin exist when fetching a gem from rubygems allow to install a specific version
skywalker% 
```

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
ph: I have remove the host, will make a matcher for that.

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
purbon: Interesting, running this example alone and only for one machine give me this output

```ruby
<RunnerTool::Command:0x11f817fc @host="127.0.0.1:2222", @cmd="/opt/logstash/bin/logstash-plugin install --no-verify logstash-output-impossible-plugin", @stderr="Plugin not found, aborting\nERROR: Installation Aborted, message: Could not find gem 'logstash-output-impossible-plugin (>= 0) java' in any of the gem sources listed in your Gemfile or installed on this machine.\n", @exit_code=1, @stdout="Installing logstash-output-impossible-plugin\n">
```

so exit_code == 1

qa/acceptance/spec/lib/cli/logstash-plugin/install_spec.rb
ph: > This would be a new specific matcher for me, this one works also good. 

Not sure I am following your point, can you give a bit more explanation or code example?

purbon: sure, 
   my basic idea is that you check here for an output that is going to be the same for all plugin installations, so having a matcher is probably the most clean way. I see ourself updating this test when we change this text 👍

ph: thanks its clearer now.

ph:
I've pushed a new version I think I have most if not all of your comments in it.

Few things to node, I've discovered that the `bin/logtash-plugin uninstall` command doesn't current work, I've the #5286 to track it and disabled the test.

I have created the `logstash-filter-qatest` which is really a dummy plugins that we can control and whatever we want with it. This plugin has no cross dependency and for the simplicity of it I have created 2 gemspec in the support folder.

I am still running into issues with return code, I've keep the code commented. But I can show it to you if you want? 

Running the complete suite on one vm, takes between 12 and 15 minutes on my machines.

purbon:
on it for more review and test the ssh issue. might you need to rebase this again as it's basically out of sync due to my last merged in the fearture branch I guess.

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  describe "on #{name}" do
+    context "with a direct internet connection" do
+      context "when the plugin exist" do
+        context "from a local `.GEM` file" do
+          let(:gem_name) { "logstash-filter-qatest-0.1.1.gem" }
+          let(:gem_path_on_vagrant) { "/tmp/#{gem_name}" }
+          before(:each) do
+            logstash.download("https://rubygems.org/gems/#{gem_name}", gem_path_on_vagrant)
```
purbon: [not related with this PR, just a dream for the feature] might be a good improvement for next iteration would to have this gem somehow local, isn't? what do you think?

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
```diff
@@ -0,0 +1,59 @@
...
+          end
+
+          after(:each) { logstash.delete_file(gem_path_on_vagrant) }
+
+          it "successfully install the plugin" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install #{gem_path_on_vagrant}")
+            expect(command).to install_successfully
+            expect(logstash).to have_installed?("logstash-filter-dns")
```
purbon: 😍

purbon:
@ph 

```
skywalker% LS_VAGRANT_HOST=ubuntu-1404 bundle exec rspec acceptance/spec/lib/cli_operation_spec.rb 
[Acceptance specs] running on ["ubuntu-1404"]
NoMethodError: undefined method `desc' for main:Object
```

still does not work as this PR requires rake context to be loaded, how hard do we feel about it?

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
ph: OK for next iteration

ph:
@purbon Concerning the`LS_VAGRANT_HOST`, I would prefer it to work, but I didnt know it was supposed to work in the current code so I used the `bundle exec rake "qa:acceptance:single[ubuntu-1404]"` task instead for testing and uncommenting the test I did not want.

purbon:
no problem at all, well the idea is that is one of the methods we get to actually run 1 example and not the full suite, so this will actually provide a solution for #5228. What do you think? and on the other side, is an unnecessary dependency on rake for the test suite.

purbon:
trying to build a deb artifact to be used in the integration test from this branch actually give this error https://gist.github.com/purbon/45c9e06f48da564ab6e483b898cb670f this makes testing this PR impossible for now 😭

qa/rspec/matchers/cli_matchers.rb
```diff
@@ -0,0 +1,31 @@
...
+
+RSpec::Matchers.define :run_successfully_and_output do |expected_output|
+  match do |actual|
+    (actual.exit_status == 0 || actual.exit_status.nil?) && actual.stdout =~ expected_output
+  end
+end
+
+RSpec::Matchers.define :have_installed? do |name,*args|
+  match do |actual|
+    version = args.first
```
purbon: you can actually have a cleaner interface to this if you use something like http://www.relishapp.com/rspec/rspec-expectations/v/3-4/docs/custom-matchers/define-matcher-with-fluent-interface

this will allow you have something like

```ruby
  expect(logstash).to have_installed?("logstash-input-foo").with_version("0.1.1")
```

what do you think?

purbon:
interesting, rake artifact:deb works on master, but not in this branch 😿

purbon:
@ph interesting, re https://github.com/elastic/logstash/pull/5259#issuecomment-220249566 basically I took this branch and did a rebase, then the artifact build properly, _computers_.....

qa/rspec/commands/base.rb
```diff
@@ -0,0 +1,60 @@
+# encoding: utf-8
+
+module ServiceTester
+
+  class Base
+    LOCATION="/logstash-build".freeze
+
+    def logstash_path
+      "/opt/logstash/"
```
purbon: why not having this as a constant?

qa/acceptance/spec/shared_examples/cli/logstash-plugin/update.rb
```diff
@@ -0,0 +1,38 @@
...
+  after :each do
+    logstash.uninstall
+  end
+
+  let(:plugin_name) { "logstash-filter-qatest" }
+  let(:previous_version) { "0.1.0" }
+
+  before do
+    logstash.run_command_in_path("bin/logstash-plugin install --version #{previous_version} #{plugin_name}")
+    # Logstash wont update when we have a pinned versionin the gemfile so we remove them
```
purbon: does this behaviour actually make sense? I thought we alway wanted to update all plugins, even if we got them pinned with a given version, isn't?

purbon: this looks to me like a bug.

ph: Lets create an issue to discuss that.

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
```diff
@@ -0,0 +1,59 @@
...
+          let(:gem_path_on_vagrant) { "/tmp/#{gem_name}" }
+          before(:each) do
+            logstash.download("https://rubygems.org/gems/#{gem_name}", gem_path_on_vagrant)
+          end
+
+          after(:each) { logstash.delete_file(gem_path_on_vagrant) }
+
+          it "successfully install the plugin" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install #{gem_path_on_vagrant}")
+            expect(command).to install_successfully
```
purbon: s/intall_sucessfully/run_sucessfully

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
```diff
@@ -0,0 +1,59 @@
...
+          let(:gem_name) { "logstash-filter-qatest-0.1.1.gem" }
+          let(:gem_path_on_vagrant) { "/tmp/#{gem_name}" }
+          before(:each) do
+            logstash.download("https://rubygems.org/gems/#{gem_name}", gem_path_on_vagrant)
+          end
+
+          after(:each) { logstash.delete_file(gem_path_on_vagrant) }
+
+          it "successfully install the plugin" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install #{gem_path_on_vagrant}")
```
purbon: I'm wondering if we can DRY the multiple `bin/logstash-plugin references into a single place? not really bug, just something to simplify mantanence in the feature i think.

qa/acceptance/spec/shared_examples/cli/logstash-plugin/list.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    after :all do
+      logstash.uninstall
+    end
+
+    context "without a specific plugin" do
+      it "display a list of plugins" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list")
+        # expect(result.exit_status).to eq(0)
```
purbon: can we clean up all this commented lines, we can always put them back when necessary.

qa/acceptance/spec/shared_examples/cli/logstash-plugin/list.rb
```diff
@@ -0,0 +1,51 @@
+# encoding: utf-8
+require_relative "../../../spec_helper"
+require "logstash/version"
+require "fileutils"
+
+shared_examples "logstash list" do |logstash|
+  describe "logstash-plugin list on #{logstash.host}" do
+    before :all do
```
purbon: add () ?

qa/acceptance/spec/shared_examples/cli/logstash-plugin/list.rb
```diff
@@ -0,0 +1,51 @@
...
+      logstash.install(LOGSTASH_VERSION)
+    end
+
+    after :all do
```
purbon: add () ?

qa/acceptance/spec/shared_examples/cli/logstash-plugin/update.rb
ph: At first I though it was a bug than I did a quick chat with @colinsurprenant about and I think the conclusion was that we need some kind of flag to allow override of pinned version. I will create an issue to discuss it.

qa/acceptance/spec/shared_examples/cli/logstash-plugin/list.rb
```diff
@@ -0,0 +1,51 @@
...
+      it "display a list of plugins" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list")
+        # expect(result.exit_status).to eq(0)
+        expect(result.stdout.split("\n").size).to be > 1
+      end
+
+      it "display a list of installed plugins" do
+        result = logstash.run_command_in_path("bin/logstash-plugin list --installed")
+        # expect(result.exit_status).to eq(0)
+        expect(result.stdout.split("\n").size).to be > 1
```
purbon: can we wrap this in a helper/matcher, something like `expect(logstash). to have_installed_at_least_one_plugin` ? easier to maintain and cleaner, what do you think?

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
```diff
@@ -0,0 +1,59 @@
+# encoding: utf-8
+require_relative "../../../spec_helper"
+require "logstash/version"
+require "fileutils"
+require "logstash/devutils/rake"
```
purbon: we should remove this from here, test can we run with rake and with directly the rspec tool, for example to run only 1 example as requested #5228

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
```diff
@@ -0,0 +1,59 @@
...
+
+shared_examples "logstash install" do |logstash|
+  before(:each) do
+    logstash.install(LOGSTASH_VERSION)
+  end
+
+  after :each do
```
purbon: add ()?

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
```diff
@@ -0,0 +1,59 @@
...
+    logstash.uninstall
+  end
+
+  describe "on #{name}" do
```
purbon: can we use here the host name information, now it gets here the IP:PORT what can really change end of the day.

qa/acceptance/spec/shared_examples/cli/logstash-plugin/update.rb
purbon: create an issue please 👍

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
ph: After a small discussion, `name` should return the distribution like `ubuntu-1404`, `centos-6`

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
ph: The matchers actually do an expect on the stdout that is really tied to /install/, so I think install_successfully is a better name in that context.

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
ph: Maybe in the future, but I really like how they match the actual usage of the logstash cli tool, not sure if it would be a good thing to hide it behind an helper.

qa/acceptance/spec/shared_examples/cli/logstash-plugin/list.rb
ph: done

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
purbon: is not that much about hiding, but about creating an abstraction that make
this more DRY. but I agree we can do it later.

- purbon

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
purbon: I created https://github.com/elastic/logstash/pull/5323 to address this, howerver I find this line is wrong, name is here an rspec variable, we should either use something like `logstash.hostname` as in the other PR.

qa/rspec/commands/debian.rb
```diff
@@ -15,7 +15,7 @@ def installed?(hosts, package)
     end
 
     def package_for(version, base=ServiceTester::Base::LOCATION)
-      File.join(base, "logstash-#{version}_all.deb")
+      File.join(base, "logstash-#{version}.deb")
```
purbon: ❤️

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
purbon: @ph lets make sure we open an issue to improve this, or add a comment to #5228, whatever if better.

purbon:
@ph sorry to report another error, I just tried this in a fresh ubuntu-1404 vm with a newly created deb artifact from this PR, this is the output I get https://gist.github.com/purbon/088cb194b9b55fe7e31005f046e66bcb

any idea what might be wrong 😢

ph:
You haven't done anything wrong I am able to reproduce this issue, if you add `DEBUG=1` before running the command you will se our friend again. 

```
Errno::ENOENT: No such file or directory - /opt/logstash/vendor/local_gems/3ddbb6f5/logstash-core-event-java-5.0.0.dev-java/deps.lst
```

Not sure why the previous package I have generated did not encounter this issues.

ph:
@purbon Yeah, I dont think the problem is related to this PR. I think package generated from non released gems are busted and don't work. I don't recall a change that could have made this regression.

purbon:
PR starts to look quite good, can you squash it? if everything goes good
with my test will try to merge tomorrow morning.

- purbon

purbon:
@ph we should definetly open an issue for this so we can fix it for release soon.

ph:
@purbon I have createdd https://github.com/elastic/logstash/issues/5325, I think we might have found our first bug with the acceptance test.

purbon:
:+1: //cc @suyog

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
purbon: naming is hard 👍

qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
```diff
@@ -0,0 +1,59 @@
...
+shared_examples "logstash install" do |logstash|
+  before(:each) do
+    logstash.install(LOGSTASH_VERSION)
+  end
+
+  after(:each) do
+    logstash.uninstall
+  end
+
+  describe "on #{name}" do
```
purbon: we should use here logstash.hostname instead of name, name is a custom variable created by rspec, so what it prints is actually this.

```
  behaves like logstash install
    on RSpec::ExampleGroups::CLIOperation::BehavesLikeLogstashInstall
      with a direct internet connection
```

see https://github.com/elastic/logstash/pull/5323 as example.

purbon:
@ph can we rename `qa/supports/` to `qa/support/`

purbon:
Hi, 
  I did my last two comments, I think if we address https://github.com/elastic/logstash/pull/5259#discussion_r64011561 and https://github.com/elastic/logstash/pull/5259#issuecomment-220557916, squash it then I'm happy to get this merged.

You could also see a set of follow up issues I created with stuff that need to be improved.

Looking forward to get this merged.

ph:
@purbon Making the changes and squashing it.

purbon:
As spoken, it would be better if we rollback e49f0efaf8b9aab9790cabaa5fbd8bbbb3e04c18 and actually push another PR where we actually see where do we need this fix exactly, it will help us understand this problem and keep PR not mix up 👍, thanks a lot for fixing this @ph!

ph:
This PR requires depends on code from theses PR:

https://github.com/elastic/logstash/pull/5337 
https://github.com/elastic/logstash/pull/5323

And is blocked by https://github.com/elastic/logstash/issues/5324, I cannot run the QA test now.

purbon:
I can merge https://github.com/elastic/logstash/pull/5323 now so we can rebase this before merge, makes sense?

purbon:
https://github.com/elastic/logstash/pull/5337 is only for debian, discovered with ubuntu-1404, right?

purbon:
I'm working on a solution for https://github.com/elastic/logstash/issues/5324 as we speak.

purbon:
I run a few test here and i'm good to get this merged, LGTM!!!!! :shipit: :shipit: :shipit: :shipit: :shipit:, check https://github.com/elastic/logstash/issues/5324#issuecomment-220824469 for detailed comments on it.

so LGTM! merge it!

PR #5260, opened 2016-05-09
Add more platforms to our Acceptance testing QA layer

purbon:
This PR adds:

* Makes sure the vagrant ssh-config usage is only done with running machines, so it does not complain when executed (Fix an issue)
* Add a more complete list of linux flavors to the QA system, this include latest ubuntu, oel, fedora, debian and suse.

### DIFF
--- a/qa/rspec/commands.rb
+++ b/qa/rspec/commands.rb
@@ -1,6 +1,8 @@
 # encoding: utf-8
 require_relative "./commands/debian"
+require_relative "./commands/ubuntu"
 require_relative "./commands/redhat"
+require_relative "./commands/suse"
 require "forwardable"
 
 module ServiceTester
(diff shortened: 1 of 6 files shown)


purbon:
@tal your thoughts are very welcome

qa/rspec/commands.rb
```diff
@@ -50,10 +52,16 @@ def to_s
 
   class CommandsFactory
 
-    def self.fetch(type)
+    def self.fetch(type, host)
       case type
       when "debian"
-        return DebianCommands.new
+        if host.start_with?("ubuntu")
```
talevy: why not have `ubuntu` be its own type for simplicity?

talevy:
small comments, otherwise LGTM

qa/rspec/commands.rb
purbon: actually ubuntu is debian based, so most of the interaction will be similar, but not the same for some outputs. This is the main reason. 

On the other side, you can see the opensuse example, is redhat (rpm) based, but the hole interaction is completely different, so I decided to have a different type. 

Makes sense to you?

talevy: I see, that is fine. yeah

talevy:
LGTM

purbon:
thanks @talevy 🚢  it!!!

PR #5265, opened 2016-05-10
undo allow_env flag

### LINKED ISSUE #5263: Remove feature flag for environment variable support
In 2.3 we [added a feature flag](https://github.com/elastic/logstash/pull/4963/files) for environment variable support in config since it broke stuff when passwords had `$`. Since this is a very popular feature, we should make it work out of the box for 5.0, where we have a chance to document breaking changes. 

- [ ] Only allow `${my_env_var}` syntax for expanding environment variables in config. Do not expand `$my_var`
- [ ] Remove `--allow-env` 
- [ ] Add tests for values with `$`
- [ ] Update documentation

talevy:
allow-env flag was introduced because of a backwards incompatible
change, this flag is no longer needed for 5.0 and onwards.

Closes #5263.

### DIFF
--- a/docs/static/command-line-flags.asciidoc
+++ b/docs/static/command-line-flags.asciidoc
@@ -65,9 +65,6 @@ Logstash has the following flags. You can use the `--help` flag to display this 
 -r, --[no-]auto-reload
   Monitor configuration changes and reload the configuration whenever it is changed.
 
---allow-env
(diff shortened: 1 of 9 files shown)


suyograo:
Can you update docs here: https://www.elastic.co/guide/en/logstash/current/command-line-flags.html

talevy:
updated

suyograo:
LGTM

PR #5274, opened 2016-05-10
Publishing a new set of gem for the 5.0.0-alphas

### LINKED ISSUE #5273: Plugin test fails to run in the `plugin-api-v1` branch of the plugin
This was originally reported by @andrewvc,

When you try to run the tests inside the v1 branch you could get the following

```
  6) outputs/elasticsearch with timeout set should fail after the timeout
     Failure/Error: Thread.new { eso.receive(LogStash::Event.new) }
     NoMethodError:
       undefined method `[]' for 2016-05-09T23:47:29.890Z %{host} %{message}:LogStash::Event
```
By doing a bit more investigation and by reading the log of the `bundle install` command we can see that `bundler` is picking up the last version of the `logstash-core-event-java` which implement the interface changes.

```
Using logstash-core-event-java 5.0.0.beta1.snapshot1
Using logstash-core 5.0.0.alpha2
Using logstash-core-plugin-api 1.13.0
```

This mean one of our dependency is incorrectly defined, in this case the `logstash-core` define the dependency on the `logstash-core-event-java` as this:

```
logstash-core-event-java ~> 5.0.0.alpha2 # incorrect version.
logstash-core-event-java = 5.0.0.alpha2 # correctly strict defined version.

# Verification
 Gem::Requirement.new("~> 5.0.0.alpha2").satisfied_by?(Gem::Version.new("5.0.0.beta1.snapshot1"))
=> true
```

The problem with declaring a dependency like this is that it will accept minor version and **beta1.snapshot** is considered a minor version update. 

**Actions to take to fix the problem**

- [x] We will publish a new `logstash-core-5.0.0.alpha3.snapshot1`
- [x] We will publish a new `logstash-core-plugin-api-0.14` depending on `logstash-core-5.0.0.alpha3.snapshot1`
- [x] We will yank the `logstash 5.0.0.beta1.snapshot1`,`logstash-core-event-java.beta1.snapshot1` and `logstash-core-plugin-api-2.0`
- [x] We will publish `logstash 5.0.0.beta1.snapshot2`,`logstash-core-event-java.beta1.snapshot2` and `logstash-core-plugin-api-2.1`

By releasing a new  `logstash-core-5.0.0.alpha3.snapshot1`, the bundle install will pick it up and use the right dependency on the event.

ph:
This Change is necessary to allow plugins 2.X to be build again.
Fixes #5273

### DIFF
--- a/logstash-core/logstash-core.gemspec
+++ b/logstash-core/logstash-core.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
   gem.version       = LOGSTASH_CORE_VERSION.gsub(/-/, '.')
 
-  gem.add_runtime_dependency "logstash-core-event-java", "~> 5.0.0.alpha2"
+  gem.add_runtime_dependency "logstash-core-event-java", "5.0.0.alpha2"
 
   gem.add_runtime_dependency "cabin", "~> 0.8.0" #(Apache 2.0 license)
   gem.add_runtime_dependency "pry", "~> 0.10.1"  #(Ruby license)
--- a/logstash-core/logstash-core.gemspec
+++ b/logstash-core/logstash-core.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
   gem.version       = LOGSTASH_CORE_VERSION.gsub(/-/, '.')
 
-  gem.add_runtime_dependency "logstash-core-event-java", "5.0.0.alpha2"
+  gem.add_runtime_dependency "logstash-core-event-java", "5.0.0.alpha3.snapshot1"
 
   gem.add_runtime_dependency "cabin", "~> 0.8.0" #(Apache 2.0 license)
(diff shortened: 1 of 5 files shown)


logstash-core/logstash-core.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
   gem.version       = LOGSTASH_CORE_VERSION.gsub(/-/, '.')
 
-  gem.add_runtime_dependency "logstash-core-event-java", "~> 5.0.0.alpha2"
+  gem.add_runtime_dependency "logstash-core-event-java", "5.0.0.alpha2"
```
suyograo: alpha3

suyograo:
Might as well release alpha3 for java event.

ph:
@suyograo Concerning the java event I though It was unecessary, if we want to do it we need to release the code from the tagged branch to make sure we don't publish the new api.

ph:
updated logstash-core-java-event

suyograo:
LGTM

ph:
merged in https://github.com/elastic/logstash/tree/v5.0.0.alpha3.snapshot1

PR #5276, opened 2016-05-10
Backport/json logging pr 4820

jordansissel:
Allow Logstash to write its logs in JSON format

This is made available by a --log-in-json flag. Default is false.
When false, the old behavior [1] is used.

When true, JSON logs are emitted.

[1] The old behavior is really two things. First, using Object#inspect to
serialize. Second, to color the output if the IO is a tty.

For #1569

This is a manual backport of PR #4820 into the 2.x branch.


jordansissel:
Derp. wrong branch.

PR #5284, opened 2016-05-11
add missing to_json_with_metadata method found in ls-event-ruby

talevy:
It would be nice to have a `to_json_with_metadata` in Java Event. It existed in the ruby version.

aside from that, some other small changes like fixing the `toMap` to return a copy, rather than a mutable reference to the original event's data

### DIFF
--- a/logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
+++ b/logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
@@ -245,19 +245,13 @@ public class JrubyEventExtLibrary implements Library {
         @JRubyMethod(name = "to_hash")
         public IRubyObject ruby_to_hash(ThreadContext context) throws IOException
         {
-            return Rubyfier.deep(context.runtime, this.event.toMap());
+            return Rubyfier.deep(context.runtime, this.event.getData());
         }
 
         @JRubyMethod(name = "to_hash_with_metadata")
         public IRubyObject ruby_to_hash_with_metadata(ThreadContext context) throws IOException
         {
-            Map data = this.event.toMap();
-            Map metadata = this.event.getMetadata();
-
-            if (!metadata.isEmpty()) {
-                data.put(Event.METADATA, metadata);
-            }
-            return Rubyfier.deep(context.runtime, data);
+            return Rubyfier.deep(context.runtime, this.event.toMapWithMetadata());
         }
 
         @JRubyMethod(name = "to_java")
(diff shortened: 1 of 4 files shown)


colinsurprenant:
I am not sure `toMap` should necessarily enforce a deep data map cloning? My reasoning is that under some circumstances the price to pay for deeply cloning my be high and might not always be required? The returned hash should not be mutated and that should be documented but with the lack of proper & efficient immutable data structures maybe we should avoid always deep cloning?

talevy:
If someone didn't want to incur the price of cloning, wouldn't they just use `Event#getData()`?

colinsurprenant:
@talevy good point :P 
Then I am not sure why we introduced `toMap` to start with? ... the JRuby ext `to_hash` currently clones the `toMap` result per https://github.com/elastic/logstash/blob/b2f04ba9b92d58330ad78942bf7c85890c2c956c/logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java#L248 (which is ok).

So, ok, I agree with this:
- let's keep `getData` to retrieve original internal data map
- let's have `toMap` return a *safe* deeply cloned data map

you'll have to update ^^ `to_hash` to not clone the clone.

Can you add proper tests to assert this and also document the methods, especially `toMap` ?

talevy:
@colinsurprenant, yup. on it. thanks!

talevy:
unfortunately, `ruby_to_hash_with_metadata` will have to copy twice

talevy:
@colinsurprenant updated

logstash-core-event-java/lib/logstash-core-event-java_jars.rb
```diff
@@ -1,6 +1,14 @@
 # this is a generated file, to avoid over-writing it just delete this comment
```
colinsurprenant: why change this file? this is a generated file. If jar_dependencies is not loaded the problem is probably elsewhere

logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -245,19 +245,13 @@ public IRubyObject ruby_to_s(ThreadContext context)
         @JRubyMethod(name = "to_hash")
         public IRubyObject ruby_to_hash(ThreadContext context) throws IOException
         {
-            return Rubyfier.deep(context.runtime, this.event.toMap());
+            return Rubyfier.deep(context.runtime, this.event.getData());
```
colinsurprenant: why change this to getData()?

talevy: the multiple copy problem.
Rubyfier.deep already copies the data structure, no? Figured it would be a read-only operation, so why copy twice

logstash-core-event-java/lib/logstash-core-event-java_jars.rb
talevy: woah. didn't mean to push this file. unintended

jordansissel:
Do we need this method? I think only the rubydebug codec uses it. Maybe we can achieve what rubydebug needs without this method?

talevy:
@jordansissel, So long as we keep ruby-event in core, I think it is important to at least keep it somewhat in sync with functionality

talevy:
Closing this because it is no longer needed

PR #5296, opened 2016-05-13
have Event#toMap return copy

talevy:


colinsurprenant:
I think we need to verify current usages of `toMap()` to make sure there's no double-cloning involved like https://github.com/talevy/logstash/blob/1339b096b18de958eafb6cef08627f6a5a5c64c7/logstash-core-event-java/src/main/java/com/logstash/ext/JrubyEventExtLibrary.java#L248

talevy:
ah, I forgot about this one that I changed in the previous PR. will update

talevy:
@colinsurprenant thanks, updated. I see no other usages

logstash-core-event-java/src/test/java/com/logstash/EventTest.java
```diff
@@ -4,6 +4,9 @@
 
 import java.io.IOException;
 import java.util.*;
+
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.core.IsSame.sameInstance;
```
colinsurprenant: are we introducing a new `org.hamcrest` dependency?

talevy: looks like it has already been included. I just autocompleted. unless I am mistaken?

talevy: I liked the `sameInstance` assertion from there, so I started typing it. and it worked. so I stuck with it

talevy:
Looks like it is somehow shadowed inside the `json-unit` testCompile dependency?

colinsurprenant:
I don't really mind if we introduce a new testCompile dep but I am not sure how this can work if it is not included in `build.gradle`?

talevy:
It is being brought in by `json-unit`

talevy:
well. it doesn't matter anymore. I removed the code dependency

colinsurprenant:
LGTM

PR #5299, opened 2016-05-13
[WIP] use closure table instead of ivars

colinsurprenant:
This is a quick & dirty patch to demonstrate the idea of using a `@closures` table instead of dynamically creating ivars which potentially leads to a slowdown when/if multiple pipelines are created.

Note that all core specs are passing with this patch. I did not run plugins tests and neither assessed performance impact.

If this idea has traction, feel free to dump this and integrate in another PR.

### DIFF
--- a/logstash-core/spec/logstash/pipeline_spec.rb
+++ b/logstash-core/spec/logstash/pipeline_spec.rb
@@ -616,4 +616,20 @@ describe LogStash::Pipeline do
       end
     end
   end
+
+  context "Pipeline object" do
+    before do
+      allow(LogStash::Plugin).to receive(:lookup).with("input", "generator").and_return(LogStash::Inputs::Generator)
+      allow(LogStash::Plugin).to receive(:lookup).with("codec", "plain").and_return(DummyCodec)
+      allow(LogStash::Plugin).to receive(:lookup).with("filter", "dummyfilter").and_return(DummyFilter)
+      allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(DummyOutput)
+    end
+
+    let(:pipeline1) { LogStash::Pipeline.new("input { generator {} } filter { dummyfilter {} } output { dummyoutput {}}") }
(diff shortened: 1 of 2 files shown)


colinsurprenant:
oh, and I did not pay attention to the `define_singleton_method` functions  but we could probably use the same idea?

guyboertje:
For Prior Art, JRuby does similar for Thread internal variables.

```java

@JRubyClass(name="Thread")
public class RubyThread extends RubyObject implements ExecutionContext {

    private static final Logger LOG = LoggerFactory.getLogger("RubyThread");

    /** The thread-like think that is actually executing */
    private ThreadLike threadImpl;

    /** Normal thread-local variables */
    private transient Map<IRubyObject, IRubyObject> threadLocalVariables;

    /** Context-local variables, internal-ish thread locals */
    private final Map<Object, IRubyObject> contextVariables = new WeakHashMap<Object, IRubyObject>();

```

guyboertje:
Although, consider the name `@pipeline_vars` or `@pipeline_variables`.

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -616,4 +616,20 @@ class TestPipeline < LogStash::Pipeline
...
+      allow(LogStash::Plugin).to receive(:lookup).with("codec", "plain").and_return(DummyCodec)
+      allow(LogStash::Plugin).to receive(:lookup).with("filter", "dummyfilter").and_return(DummyFilter)
+      allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(DummyOutput)
+    end
+
+    let(:pipeline1) { LogStash::Pipeline.new("input { generator {} } filter { dummyfilter {} } output { dummyoutput {}}") }
+    let(:pipeline2) { LogStash::Pipeline.new("input { generator {} } filter { dummyfilter {} } output { dummyoutput {}}") }
+
+    it "should not add ivars" do
+       expect(pipeline1.instance_variables).to eq(pipeline2.instance_variables)
```
jsvd: curiously, my current pr passes this test :)

colinsurprenant:
closing, commit has been added into #5250

PR #5301, opened 2016-05-14
Move root resource doc to main page for monitoring APIs

dedemorton:
Moving this content because Suyog and Alvin rightfully suggested that it didn't warrant its own topic.


suyograo:
LGTM

PR #5306, opened 2016-05-16
Remove the pull request template.

jordansissel:
This was a good idea, but now every PR seems to be submitted with the
description being "Thanks for contributing to logstash!" and that's
pretty annoying.


suyograo:
LGTM

Issue #4923, opened 2016-03-30
More metrics for monitoring API


acchen97:
Outlining plan to augment the following monitoring API endpoints with additional metrics ([phase 1](https://github.com/elastic/logstash/issues/4446)):
* Root Resource (/) - retrieve general node information
* Node Info (/_node) - retrieve node information
* Node Stats (/_node/stats) - retrieve current node statistics
* Hot Threads (/_node/hot_threads) - retrieve current node hot threads
* Plugins (/_plugins) - retrieve plugin information and statistics

# Additional Metrics
### Node Stats
| Metrics | Namespace | API Resource |
----------- | ------ | -------- | --------- |
Events In [bytes] | stats.events.in | /_node/stats/events |
Events Filtered [bytes] | stats.events.filtered | /_node/stats/events |
Events Out [bytes] | stats.events.out | /_node/stats/events |
Thread Count | stats.jvm.threads.count | /_node/stats/jvm |
Peak Thread Count | stats.jvm.threads.peak_count | /_node/stats/jvm |
Number of Open File Descriptors | stats.process.open_file_descriptors | /_node/stats/process |
Peak Open File Descriptors | stats.process.peak_open_file_descriptors | /_node/stats/process |
Max File Descriptors | stats.process.max_file_descriptors | /_node/stats/process |
CPU Usage (%) | stats.process.cpu.percent | /_node/stats/process/cpu |
CPU Time | stats.process.cpu.total_in_millis | /_node/stats/process/cpu |
Virtual Memory - Total | stats.process.mem.total_virtual_in_bytes | /_node/stats/process/mem

```
{
    "events": {
        "in": 91,
        "bytes_in": 900000
        "filtered": 91,
        "bytes_filtered": 900000
        "out": 91
        "bytes_out": 900000
    },
    "threads": {
        "count": 12,
        "peak_count": 12
    },
    "process": {
        "open_file_descriptors": 200,
        "max_file_descriptors": 400,
        "cpu": {
            "percent": 20,
            "total_in_millis": 30081091
        },
        "mem": {
            "total_virtual_in_bytes": 5621555200
        }
    }
}
```

### Node Info
| Metrics | Namespace | API Resource |
----------- | ------ | -------- | --------- |
Pipeline Workers | pipeline.workers | /_node/pipeline |
Pipeline Batch Size | pipeline.batch_size | /_node/pipeline |
Pipeline Batch Delay | pipeline.batch_delay | /_node/pipeline |
[...]

PR #5307, opened 2016-05-16
WebAPI refactor && JVM thread stats

### LINKED ISSUE #5328: API Webserver Shutdown does not free port
Running tests with `bin/rspec  ./logstash-core/spec --seed 49545` exposes a flaw. The `stop_webserver` call does not actually free the port. This makes tests with this seed fail.

andrewvc:
WIP for adding the stats mentioned in #4923 

Fixes #5328 

Refactors: 

1. All classes renamed to be consistent with their location on the FS path.
1. Rack app changed to no longer use init.ru, but use direct class instantiation (as a result we don't need a nested 'lib' folder under logstash/api.
1. Commands refactored to handle multiple responsibilities with multiple method names. (The #run method meant we needed tons of little classes that were 90% boilerplate. Check the new node_stats command).
1. Intent is to refactor the commands directory to be a direct mirror of the web API. This will simplify automated stats shipping.

### Node Stats

* [x] stats.events.in | /_node/stats/events
* [x] stats.events.filtered | /_node/stats/events |
* [x] stats.events.out | /_node/stats/events |
* [x] stats.jvm.threads.count | /_node/stats/jvm |
* [x] stats.jvm.threads.peak_count | /_node/stats/jvm |
* [x] stats.process.open_file_descriptors | /_node/stats/process |
* [x] stats.process.peak_open_file_descriptors | /_node/stats/process |
* [x] stats.process.max_file_descriptors | /_node/stats/process |
* [x] stats.process.cpu.percent | /_node/stats/process/cpu |
* [x] stats.process.cpu.total_in_millis | /_node/stats/process/cpu |
* [x] stats.process.mem.total_virtual_in_bytes | /_node/stats/process/mem


logstash-core/spec/support/resource_dsl_methods.rb
```diff
@@ -0,0 +1,47 @@
+module ResourceDSLMethods
```
purbon: should we keep this support class within the api directory?

andrewvc: I think so, only API stuff would use it.

logstash-core/lib/logstash/api/app.rb
```diff
@@ -0,0 +1,15 @@
...
+require "cabin"
+require "logstash/json"
+require "helpers/app_helpers"
+require "app/service"
+require "app/command_factory"
+require "logstash/util/loggable"
+
+module LogStash::Api
+  class App
+
```
suyograo: blank lines?

logstash-core/lib/logstash/api/app_helpers.rb
```diff
@@ -20,4 +20,8 @@ def as_boolean(string)
     return false  if string == false  || string.blank? || string =~ (/(false|f|no|n|0)$/i)
     raise ArgumentError.new("invalid value for Boolean: \"#{string}\"")
   end
+
+  def define_resource(commands)
+    # TODO: stuff
```
suyograo: remove this method?

logstash-core/lib/logstash/api/app.rb
andrewvc: I'll just pull this class. I'm realizing now that it no longer has a purpose.

andrewvc:
Thanks for the review @suyograo I've fixed those issues.

Tests won't pass without the master version of https://github.com/elastic/jrmonitor

I'm of the opinion that we should just pull it into the main Logstash repo. I think it's too much effort to maintain it as a separate gem.

logstash-core/lib/logstash/webserver.rb
```diff
@@ -59,40 +56,34 @@ def error(str)
...
+        run LogStash::Api::Modules::Root
+        namespaces.each_pair do |namespace, app|
+          map(namespace) do
+            run app
+          end
+        end
+      end
+    end
+
+    def rack_namespaces
```
purbon: why should the webserver class care about the resources (web endpoints) defined in the webapp? for me having them here complicate the separation of concerns.

logstash-core/lib/logstash/webserver.rb
```diff
@@ -31,24 +35,17 @@ def initialize(logger, options={})
...
     def run
       log "=== puma start: #{Time.now} ==="
 
-      @runner = Puma::Single.new(self)
-      @status = :run
-      @runner.run
-      stop(:graceful => true)
+      stop # Just in case
+
+      @server = ::Puma::Server.new rack_app
```
andrewvc: Check that in the future we could adjust thread settings

logstash-core/lib/logstash/api/modules/base.rb
```diff
@@ -0,0 +1,41 @@
+require "logstash/api/app_helpers"
+require "logstash/api/command_factory"
+
+module LogStash
+  module Api
+    module Modules
+      class Base < ::Sinatra::Base
```
purbon: I saw initially the modules as the resources (web api endpoints ) definition, base class should not be part of the modules as it does not really define web api resources. 

On the other side, just my experience and way of thinkings (does not have to be yours), but I like the idea of having a main app class, or file, in that case I guess we could for example move this base class to the the `logstash/api` directory, what do you think?

logstash-core/lib/logstash/util/thread_dump.rb
```diff
@@ -0,0 +1,55 @@
...
+          i += 1
+        end
+      end
+
+      def idle_thread?(thread_name, data)
+        idle = false
+        if SKIPPED_THREADS.include?(thread_name)
+          # these are likely JVM dependent
+          idle = true
+        elsif thread_name.match(/Ruby-\d+-JIT-\d+/)
```
purbon: now that you're cleaning up, this is also welcome to clean up and aggregate somehow, what do you think?

logstash-core/lib/logstash/webserver.rb
purbon: as spoken, this might also be part of the main app class, what do you think?

logstash-core/lib/logstash/api/commands/stats.rb
```diff
@@ -0,0 +1,105 @@
+require "logstash/api/commands/base"
+require 'logstash/util/thread_dump'
+
+module LogStash
+  module Api
+    module Commands
+      class Stats < ::LogStash::Api::Commands::Base
```
purbon: why using full namespace here? if location is already scope, right?

purbon:
I guess this PR might need a rebase.

logstash-core/lib/logstash/agent.rb
```diff
@@ -17,6 +17,8 @@
 
 LogStash::Environment.load_locale!
 
+AGENT_LIST = {}
```
purbon: Where is used?

purbon:
@andrewvc as spoken, I gave a round of comments, assigning this back to you to follow up on them. Let  me know when all is up and I will review again 👍

logstash-core/lib/logstash/agent.rb
andrewvc: Good call! This was left in from debugging some resource issues. Will remove :)

logstash-core/lib/logstash/api/commands/stats.rb
andrewvc: Yep, this can be shortened. I'll just make it 'Commands::Base' because 'Base' is so generic it can be a little confusing.

logstash-core/lib/logstash/api/modules/base.rb
andrewvc: Well, I think its fairly common to have a 'base' class that's a sibling to the classes that inherit from it.

If we move it up a level it'd have to be renamed to `LogStash::Api::BaseModule` and moved to `api/base_module.rb`.

I'm fine with either approach. I take it you prefer the latter?

logstash-core/lib/logstash/util/thread_dump.rb
andrewvc: Makes sense, however, I think the scope has already blown up quite a bit, I'd rather do that in a separate PR.

logstash-core/lib/logstash/webserver.rb
andrewvc: I'll extract this out to a separate class for sure :)

andrewvc:
@purbon I believe I have addressed all your PR feedback. How's it look now?

logstash-core/lib/logstash/agent.rb
purbon: still not removed, 👍  guess you miss the poor kid here :-P

logstash-core/lib/logstash/api/modules/base.rb
purbon: well, in that case for me (in my head) this is more like a utility base class, right? so ... let me explain. by having this and the actual resources in separate location we could achieve a more clear separation of concerns, right? I say that because end of the day, there is no actual resource setup with this class. What do you think?

logstash-core/lib/logstash/api/modules/node.rb
```diff
@@ -0,0 +1,30 @@
+# encoding: utf-8
+module LogStash
+  module Api
+    module Modules
+      class Node < ::LogStash::Api::Modules::Base
+        get "/" do
+          payload = {
```
purbon: can we might, use `payload = {}`

purbon: also, now that I think about it, this resource looks empty to me, why is like this?

logstash-core/lib/logstash/api/rack_app.rb
```diff
@@ -0,0 +1,32 @@
...
+            end
+          end
+        end
+      end
+
+      def self.rack_namespaces
+        {
+          "/_node" => LogStash::Api::Modules::Node,
+          "/_node/stats" => LogStash::Api::Modules::NodeStats,
+          "/_plugins" => LogStash::Api::Modules::Plugins
```
purbon: why is module stats not used here? see https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/api/init.ru#L22-L25 as the current namespaces definitions.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -100,6 +100,10 @@ def get(*key_paths)
       new_hash
     end
 
+    def get_shallow(*key_paths)
```
purbon: can we add a comment here indicating what this function actually does?

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -180,6 +180,6 @@ def worker_count
   end
 
   private
-  # Needed for testing, so private
+0  # Needed for testing, so private
```
purbon: why is a `0` added here before the comment, is this a typo probably?

logstash-core/lib/logstash/util/thread_dump.rb
purbon: no problem, 👍

logstash-core/lib/logstash/webserver.rb
```diff
@@ -31,24 +30,17 @@ def initialize(logger, options={})
...
     def run
       log "=== puma start: #{Time.now} ==="
 
-      @runner = Puma::Single.new(self)
-      @status = :run
-      @runner.run
-      stop(:graceful => true)
+      stop # Just in case
+
+      @server = ::Puma::Server.new(LogStash::Api::RackApp.app)
```
purbon: now looks awesome!!! 👍

logstash-core/spec/api/lib/api/node_stats_spec.rb
```diff
@@ -1,68 +1,43 @@
...
 require "logstash/json"
 
-describe LogStash::Api::NodeStats do
-
+describe LogStash::Api::Modules::NodeStats do
   include Rack::Test::Methods
+  extend ResourceDSLMethods
 
-  def app()
+  def app() # Used by Rack::Test::Methods
```
purbon: might be have this comment in line 10?

purbon:
@andrewvc good work, It looks much better, I just left you a set of new comments, some minor, but some important ones to arrange somehow. see:

* https://github.com/elastic/logstash/pull/5307#discussion_r64537711
* https://github.com/elastic/logstash/pull/5307#discussion_r64537096
* https://github.com/elastic/logstash/pull/5307#discussion_r64536031

A typo:

* https://github.com/elastic/logstash/pull/5307#discussion_r64537984


re https://github.com/elastic/logstash/pull/5307#discussion_r64245604 I explained my way of thinkig about it, might I'm good to discuss with you, my raw motivation here is to have separated resource (api end points) definition from actually sinatra code setup.

happy to help,

- purbon

logstash-core/lib/logstash/output_delegator.rb
andrewvc: Ahhhh, good catch. Total typo.

logstash-core/lib/logstash/api/modules/node.rb
andrewvc: My mistake, I was partway through working on this and forgot about it. Will fix it up.

purbon:
As spoken @andrewvc, If  all comments from https://github.com/elastic/logstash/pull/5307#issuecomment-221515453 are addresses, except the https://github.com/elastic/logstash/pull/5307#discussion_r64245604 that we know is a complicated story 👍 , I'm good to get it merged, so pre LGTM!! :shipit: :shipit:

suyograo:
Jenkins, please test this.

PR #5322, opened 2016-05-19
Fix ssh dump config + suse boostrap user level script

purbon:
two simple fixes, that will make the bootstrap process work.

* ssh config should only fetch active machines.
* suse needs a non privileged bootstrap script (for now empty)


ph:
LGTM

PR #5323, opened 2016-05-19
Use hostname in the acceptance test descriptions

purbon:
@ph think if this way is something you might also like to introduce for #5259, for all acceptance test definetly something better to have names than IP's, isn't?


ph:
yup I have replace the host with that too, good call.

LGTM

PR #5332, opened 2016-05-20
Add changes that support settings files

dedemorton:
Adds doc about settings file to resolve an item listed in  #5148 

Summary of changes:
- Formatting improvement requested by @acchen97 
- Add mention of the settings file as an option instead of specifying flags
- Fix a few minor mistakes and typos
- Add "coming" annotation to clarify that the changes are for alpha3
- Added node.name and pipeline.output.workers because it was missing from the doc (I'll add a note below about pipeline.output.workers)

### DIFF
--- a/docs/static/command-line-flags.asciidoc
+++ b/docs/static/command-line-flags.asciidoc
@@ -1,11 +1,9 @@
 [[command-line-flags]]
-=== Command-line flags
+=== Command-Line Flags
 
 Logstash has the following flags. You can use the `--help` flag to display this information.
 
-As an alternative to setting command-line flags, you can specify options in the Logstash settings file,
-`logstash.yml`, located in `LOGSTASH_HOME/config`. Any flags that you set at the command line override the corresponding
-options in the `logstash.yml` file. 
+You can also control Logstash execution by specifying options in the Logstash settings file. For more info, see <<logstash-settings-file>>.  
 
 coming[5.0.0-alpha3, Command-line flags have dots instead of dashes in their names]
 
@@ -20,13 +18,10 @@ coming[5.0.0-alpha3, Command-line flags have dots instead of dashes in their nam
  is specified, `stdout { codec => rubydebug }}` is default.
 
 *`-w, --pipeline.workers COUNT`*::
- Sets the number of pipeline workers (threads) to run for filter processing (default: half the number of cores).
+ Sets the number of pipeline workers (threads) to run for filter processing (default: number of cores).
  If you find that events are backing up, or that the CPU is not saturated, consider increasing
  this number to better utilize machine processing power.
  
-*`--pipeline.output.workers COUNT`*::
- Sets the number of pipeline workers (threads) to run for each output plugin instance.
-
 *`-b, --pipeline.batch.size SIZE`*::
  This parameter defines the maximum number of events an individual worker thread will collect
  before attempting to execute its filters and outputs. Default is 125 events.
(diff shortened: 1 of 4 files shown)


docs/static/command-line-flags.asciidoc
```diff
@@ -3,81 +3,91 @@
...
+ is specified, `stdout { codec => rubydebug }}` is default.
 
--w, --pipeline.workers COUNT
- Sets the number of pipeline workers (threads) to run for filter processing (default: number of cores).
+*`-w, --pipeline.workers COUNT`*::
+ Sets the number of pipeline workers (threads) to run for filter processing (default: half the number of cores).
  If you find that events are backing up, or that the CPU is not saturated, consider increasing
  this number to better utilize machine processing power.
+ 
+*`--pipeline.output.workers COUNT`*::
```
dedemorton: @suyograo or @jsvd I'm not convinced this is a valid CLI flag.

pipeline.output.workers DOES NOT exist here: 
https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/runner.rb

But it does exist here:
https://github.com/elastic/logstash/blob/master/config/logstash.yml#L34

suyograo:
@dedemorton Can we also add a separate section "Using the Settings File" or something similar to call out the existence of `config/logstash.yml` in detail? We can then reference this from the Command Line  section: https://github.com/elastic/logstash/pull/5332/files#diff-ab5ecd26344a9aea61a0a0581e8f283cR6

Since this is new its good to push its visibility.

docs/static/command-line-flags.asciidoc
suyograo: @jsvd did you mean `workers` in `output` plugins?

suyograo: good catch @dedemorton

jsvd: @andrewvc should we expose `pipeline.output.workers` to the command line and make it known?

docs/static/command-line-flags.asciidoc
```diff
@@ -3,81 +3,91 @@
...
  Use the given string as the configuration data. Same syntax as the config file.
- If not input is specified, 'stdin { type => stdin }' is default. If no output
- is specified, 'stdout { codec => rubydebug }}' is default.
+ If no input is specified, `stdin { type => stdin }` is default. If no output
+ is specified, `stdout { codec => rubydebug }}` is default.
 
--w, --pipeline.workers COUNT
- Sets the number of pipeline workers (threads) to run for filter processing (default: number of cores).
+*`-w, --pipeline.workers COUNT`*::
+ Sets the number of pipeline workers (threads) to run for filter processing (default: half the number of cores).
```
jsvd: I mislead you here, @dedemorton, it's actually the number of cores, not half (set in [here](https://github.com/elastic/logstash/blob/7783f76a4c29e2d944c807ceea1b4df85383c423/logstash-core/lib/logstash/environment.rb#L18))

docs/static/command-line-flags.asciidoc
suyograo: IMHO, we should not. Its just inconsistent to have 1 config option being exposed this way. We should not set this precedent.

dedemorton:
@suyograo Sure, I can add the info to a separate topic. I decided to err on the side of minimalism here, but I can see how having a separate topic might raise visibility that the file exists.

jsvd:
LGTM

suyograo:
LGTM

Issue #5325, opened 2016-05-19
Cannot install plugin with a package generated with `logstash-core` gems install installed in /vendor


ph:
@purbon found this issue when testing the PR I've made for the cli migration.

---

For the purpose of the QA test I have created a dummy filter named **logstash-filter-qatest**, this plugin is used for testing the install/update scenario and doesn't depends on any dependencies.

But the problem is easy to reproduce:

- Generate a a package (tar, debian or rpm)
- Move it to a  VM
- install / uncompress the package
- try to install the plugin with `DEBUG= 1bin/logstash-plugin install logstash-filter-qatest`

The install process will fails and you will get the following message.

```
0:  ruby-maven (3.3.10) from /home/vagrant/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/specifications/ruby-maven-3.3.10.gemspec
  jar dependencies for logstash-core-event-java-5.0.0.dev-java.gemspec . . .
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-parseable POM /home/vagrant/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/gems/jar-dependencies-0.3.2/lib/jars/gemspec_pom.rb: only whitespace content allowed before start tag and not # (position: START_DOCUMENT seen #... @1:1)  @ line 1, column 1
 @
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]
[ERROR]   The project  (/home/vagrant/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/gems/jar-dependencies-0.3.2/lib/jars/gemspec_pom.rb) has 1 error
[ERROR]     Non-parseable POM /home/vagrant/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/gems/jar-dependencies-0.3.2/lib/jars/gemspec_pom.rb: only whitespace content allowed before start tag and not # (position: START_DOCUMENT seen #... @1:1)  @ line 1, column 1 -> [Help 2]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
[...]

PR #5337, opened 2016-05-20
Force updating the certificate on the target vm

ph:
Ruby-maven tries to connect to an http host and the certificates shipped
with Ubuntu 1404 cannot verify it. This PR make this command run after
bootstrapping the debian machine.

```
sudo update-ca-certificates -f
```

This should work under all debian based distribution.

Fixes: #5325

### DIFF
--- a/qa/Vagrantfile
+++ b/qa/Vagrantfile
@@ -17,6 +17,14 @@ Vagrant.configure(2) do |config|
         sh.path = "sys/#{platform.type}/bootstrap.sh"
         sh.privileged = true
       end
+
+      if platform.bootstrap
+        machine.vm.provision :shell do |sh|
+          sh.path = platform.bootstrap
+          sh.privileged = true
+        end
+      end
+
       machine.vm.provision :shell do |sh|
         sh.path = "sys/#{platform.type}/user_bootstrap.sh"
         sh.privileged = false
--- a/qa/config/platforms.json
+++ b/qa/config/platforms.json
@@ -2,7 +2,7 @@
   "latest": "5.0.0-alpha2",
   "platforms" : {
     "ubuntu-1204": { "box": "elastic/ubuntu-12.04-x86_64", "type": "debian" },
-    "ubuntu-1404": { "box": "elastic/ubuntu-14.04-x86_64", "type": "debian" },
+    "ubuntu-1404": { "box": "elastic/ubuntu-14.04-x86_64", "type": "debian", "bootstrap": "sys/ubuntu-1404/bootstrap.sh" },
     "ubuntu-1504": { "box": "elastic/ubuntu-15.04-x86_64", "type": "debian" },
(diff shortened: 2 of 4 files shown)


qa/sys/debian/bootstrap.sh
```diff
@@ -5,3 +5,4 @@ apt-get update
 apt-get install -y openjdk-8-jdk
 update-alternatives --config java
 update-alternatives --config javac
+sudo update-ca-certificates -f
```
purbon: you don't need to use sudo here, by default this script runs with privileges

purbon: for the way we run this scripts it will run for all debians, we might also not need this for all debian's, would be nice to check if this happen for others, and if not, basically only do this for ubuntu-1404.

purbon:
Thanks man!

mkristian:
I have to add my remark to it: I did see the very same error on the openjdk8 debian package a few weeks ago. usually the jdk packages in debian do sync the jvm keystore with installed cacerts. this is somehow missing on the openjdk8 package.

purbon:
good to know @mkristian, easy to cross check for us, will do and update the PR accordingly 👍

ph:
@purbon Only 1404 is affected by this, ubuntu latest LTS ship with updated certificates from the start.
How do we want to process, run a specific script for this specific releases?

ph:
I suggest we update the `platform.json` and add a new attribute named **provisioning** or **bootstrap** that point to a shell script. Then we changes the `Vagrantfile` to execute this script when specified? WDYT?

purbon:
Yes, we could add might be specific platform script to do this, like the
thing that is different for this platform. What do you think?

/purbon

ph:
@purbon updated with a specific script for ubuntu-1404

qa/Vagrantfile
```diff
@@ -17,6 +17,14 @@ Vagrant.configure(2) do |config|
         sh.path = "sys/#{platform.type}/bootstrap.sh"
         sh.privileged = true
       end
+
+      if platform.bootstrap
```
purbon: not sure I get this way of adding the new specific platform specific scripts here, I see a few open questions for me:

* bootstrap seems to be a confusing name, specially If I only take a look at the Vagrantfile. Is not clear why this file and code is there for.
* On the other side, we now mix, have some scripts defined in the vagrantfile and some in the json file. My take here would be to have either all in the json file or have them defined only in the vagrantfile, otherwise there is an unnecessary mix in the configuration (ruby (vagrantfile) vs json).

What do you think?

purbon: see https://github.com/elastic/logstash/pull/5346 for another proposal how to solve this, the idea there is to wrap the logic to select this files in an "extended" version of the platform logic, as we try to have a minimalist Vagrantfile, I think that should be the place.

ph:
closing in favor of #5346

PR #5341, opened 2016-05-20
Feature/release packaging

untergeek:
Updates to work with @jsvd's work on `logstash.yml` settings.

This uses a wrapper around `pleaserun` (bin/system-install) to generate the proper startup files for whatever environment you're using.


Gemfile
```diff
@@ -23,3 +23,6 @@ gem "rubyzip", "~> 1.1.7", :group => :build
 gem "gems", "~> 0.8.3", :group => :build
 gem "rack-test", :require => "rack/test", :group => :development
 gem "flores", "~> 0.0.6", :group => :development
+gem "pleaserun"
+gem "mustache"
```
jordansissel: I don't see Mustache invoked in this PR. Pleaserun uses it, but lists it as a dependency. Do we need it listed here?

Same question for Clamp.

bin/logstash.lib.sh
```diff
@@ -140,7 +132,9 @@ setup_ruby() {
 jruby_opts() {
   printf "%s" "--1.9"
   for i in $JAVA_OPTS ; do
-    printf "%s" " -J$i"
+    if [ "x$i" != "x" ] && [ "x$i" != "x " ]; then
```
jordansissel: you can check if '$i' is empty by doing `if [ -z "$i" ]`. 

Also, when would $i be a space?

bin/system-install
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+unset CDPATH
+. "$(cd `dirname $0`/..; pwd)/bin/logstash.lib.sh"
+setup
+
+if [ -z "$1" ]; then
+  [ -r ${LOGSTASH_HOME}/config/startup.options ] && . ${LOGSTASH_HOME}/config/startup.options
+  [ -r /etc/logstash/startup.options ] && . /etc/logstash/startup.options
```
jordansissel: This means startup options in /etc/logstash take priority over ones in LOGSTASH_HOME. I think this is ok, but is this what you meant?

untergeek: Yes, this is what is intended.  This is necessary because the package includes LOGSTASH_HOME as /usr/share/logstash, but we want to use what's in /etc/logstash first.

bin/system-install
```diff
@@ -0,0 +1,41 @@
...
+tempfile=$(mktemp)
+if [ "x${PRESTART}" == "x" ]; then
+  opts=("--log" "$tempfile" "--overwrite" "--install" "--name" "${INIT_NAME}" "--user" "${LS_USER}" "--group" "${LS_GROUP}" "--description" "${INIT_DESCRIPTION}" "--nice" "${LS_NICE}" "--limit-open-files" "${LS_OPEN_FILES}")
+else
+  opts=("--log" "$tempfile" "--overwrite" "--install" "--name" "${INIT_NAME}" "--user" "${LS_USER}" "--group" "${LS_GROUP}" "--description" "${INIT_DESCRIPTION}" "--nice" "${LS_NICE}" "--limit-open-files" "${LS_OPEN_FILES}" "--prestart" "${PRESTART}")
+fi
+
+program="$(cd `dirname $0`/..; pwd)/bin/logstash"
+
+capture="$(ruby_exec "${LOGSTASH_HOME}/lib/systeminstall/pleasewrap.rb" "${opts[@]}" ${program} ${LS_OPTS})"
```
jordansissel: Should you do something with `$capture` ?

untergeek: Strangeness with shell actions and functions.  If I do not capture the output, it doesn't work.  I'm not sure why.  I'd be happy to make it look less weird if I could figure out why and replace it with something sensible.

config/startup.options
```diff
@@ -0,0 +1,57 @@
...
+# Open file limit
+LS_OPEN_FILES=16384
+
+# Nice level
+LS_NICE=19
+
+# If this is set to 1, then when `stop` is called, if the process has
+# not exited within a reasonable time, SIGKILL will be sent next.
+# The default behavior is to simply log a message "program stop failed; still running"
+#KILL_ON_STOP_TIMEOUT=0
```
jordansissel: If we start shipping pleaserun-generated scripts, then I don't think this setting is needed anymore... sorta. The systemd and upstart implementations of the logstash service won't support this, and the pleaserun-generated sysv script doesn't, either.

I'm not sure what the right solution here.

bin/logstash.lib.sh
untergeek: I got some weird errors when testing shell options, but that may have been from something else.  I'll retest with this as just `-z`

config/startup.options
```diff
@@ -0,0 +1,57 @@
...
+
+# Change these to have the init script named and described differently
+# This is useful when running multiple instances of Logstash on the same
+# physical box or vm
+INIT_NAME="logstash"
```
jordansissel: Let's call this SERVICE_NAME and SERVICE_DESCRIPTION

config/startup.options
untergeek: I was not aware of that.  Not a big deal to remove it.

config/startup.options
untergeek: Will make the change.

config/startup.options
```diff
@@ -0,0 +1,57 @@
+################################################################################
+# These settings are ONLY used by $LS_HOME/bin/system-install to create a custom
```
jordansissel: If this file is only used by system-install, maybe we can combine the logstash jvm options w/ a new file maybe called 'service.options' or something, and that file probably only has two settings (the name and description of the service?)

config/startup.options
```diff
@@ -0,0 +1,57 @@
...
+# If this is set to 1, then when `stop` is called, if the process has
+# not exited within a reasonable time, SIGKILL will be sent next.
+# The default behavior is to simply log a message "program stop failed; still running"
+#KILL_ON_STOP_TIMEOUT=0
+
+# Change these to have the init script named and described differently
+# This is useful when running multiple instances of Logstash on the same
+# physical box or vm
+INIT_NAME="logstash"
+INIT_DESCRIPTION="logstash"
```
jordansissel: Should there also be a setting to configure what user Logstash runs as?

pkg/centos/before-remove.sh
```diff
@@ -1,6 +1,18 @@
 if [ $1 -eq 0 ]; then
-  /sbin/service logstash stop >/dev/null 2>&1 || true
-  /sbin/chkconfig --del logstash
+  if [ ! -f "/etc/systemd/system/logstash.service" ]; then
```
jordansissel: Will this remove the upstart config on CentOS 6 ?

config/startup.options
untergeek: Isn't that LS_USER and LS_GROUP?  Aren't they already there?

pkg/debian/before-remove.sh
```diff
@@ -1,13 +1,23 @@
 #!/bin/sh
 
 if [ $1 = "remove" ]; then
-  service logstash stop >/dev/null 2>&1 || true
+  if [ ! -f "/etc/systemd/system/logstash.service" ]; then
+    service logstash stop >/dev/null 2>&1 || true
```
jordansissel: Will this remove the upstart and/or sysv scripts?

pkg/jvm.options
```diff
@@ -0,0 +1,74 @@
...
+# Set the locale language
+#-Duser.language=en
+
+# Set the locale country
+#-Duser.country=US
+
+# Set the locale variant, if any
+#-Duser.variant=
+
+## Basic settings
```
jordansissel: Feels weird to have a section above saying "Expert settings" and then below saying "Basic settings"

Are these expert basic settings? ;P

pkg/centos/before-remove.sh
untergeek: Good catch.

pkg/jvm.options
untergeek: Hahaha.  That's how it's formatted in the ES files, I think

config/startup.options
untergeek: The JVM options should be in `jvm.options` because Elasticsearch is doing that, and we're doing our best to mirror the configuration profile.  The `jvm.options` _are_ used at each invocation of Logstash, though, as `logstash.lib.sh` is configured to do.  Those can be changed, then restart Logstash to see the changes, where `startup.options` are only for generating init scripts.

pkg/jvm.options
```diff
@@ -0,0 +1,74 @@
...
+
+# set the I/O temp directory
+-Djava.io.tmpdir=$HOME
+
+# set to headless, just in case
```
jordansissel: less just in case, I suppose, more that we'll never want to use AWT or connect to an X server ;P

pkg/ubuntu/before-remove.sh
```diff
@@ -1,13 +1,23 @@
 #!/bin/sh
 
 if [ $1 = "remove" ]; then
-  service logstash stop >/dev/null 2>&1 || true
+  if [ ! -f "/etc/systemd/system/logstash.service" ]; then
+    service logstash stop >/dev/null 2>&1 || true
```
jordansissel: Will this remove the upstart job file? Should it?

config/startup.options
suyograo: +1 on being consistent with ES

bin/system-install
jordansissel: you should be able to do this:

```shell
ruby_exec "${LOGSTASH_HOME}/lib/systeminstall/pleasewrap.rb" "${opts[@]}" "${program}" ${LS_OPTS}
exit_code=$?
```

Does that not work? Hmm..

config/startup.options
jordansissel: I think we should remove it, but for users who rely on it I'm not sure what to tell them. 

`systemctl` manpage says it has a `kill` subcommand which will let you SIGKILL (or whatever) a service:

>        kill PATTERN...
>           Send a signal to one or more processes of the unit. Use --kill-who=
>           to select which process to kill. Use --signal= to select the signal
>           to send.

Given I think most of our supported platforms use systemd, maybe removing this feature is not a negative impact on users since folks who need to force logstash to terminate can just do `systemctl kill --signal=KILL logstash` or something.

jordansissel: So, +1 to remove it.

config/startup.options
jordansissel: i totally missed that. Yes. <3

pkg/jvm.options
jordansissel: Ahh well. Something we can improve on the next PR! Nothing to change for now. <3

pkg/jvm.options
untergeek: Indeed.  Just mirroring the options that came with Elastic.  We can review these any time we want.

suyograo:
Tested new dir structure and its consistent with ES layout.

```
suyog@ubuntu:~$ ls /usr/share/logstash/
bin  CHANGELOG.md  CONTRIBUTORS  Gemfile  Gemfile.jruby-1.9.lock  lib  LICENSE  NOTICE.TXT  vendor
suyog@ubuntu:~$ ls /etc/logstash/
conf.d  jvm.options  logstash.yml  startup.options
```

LGTM.

jordansissel:
Eeee much excite.

Issue #5338, opened 2016-05-20
QA Acceptance test: debian 8 doesn't bootstrap


ph:
`bootstrap.sh`

```
==> debian-8: Running provisioner: shell...
    debian-8: Running: /var/folders/yl/9hpq_qs528g60cv9trnt5z180000gn/T/vagrant-shell20160520-25625-a6gt3v.sh
==> debian-8: stdin: is not a tty
==> debian-8: /tmp/vagrant-shell: line 3: add-apt-repository: command not found
==> debian-8: Ign http://apt.puppetlabs.com jessie InRelease
==> debian-8: Get:1 http://apt.puppetlabs.com jessie Release.gpg [836 B]
==> debian-8: Get:2 http://security.debian.org jessie/updates InRelease [63.1 kB]
==> debian-8: Get:3 http://apt.puppetlabs.com jessie Release [40.6 kB]
==> debian-8: Get:4 http://apt.puppetlabs.com jessie/PC1 amd64 Packages [7,150 B]
==> debian-8: Get:5 http://security.debian.org jessie/updates/main Sources [136 kB]
==> debian-8: Get:6 http://security.debian.org jessie/updates/contrib Sources [1,439 B]
==> debian-8: Get:7 http://security.debian.org jessie/updates/non-free Sources [14 B]
==> debian-8: Get:8 http://security.debian.org jessie/updates/main amd64 Packages [247 kB]
==> debian-8: Ign http://apt.puppetlabs.com jessie/PC1 Translation-en_US
==> debian-8: Ign http://apt.puppetlabs.com jessie/PC1 Translation-en
==> debian-8: Ign http://cloudfront.debian.net jessie InRelease
==> debian-8: Get:9 http://security.debian.org jessie/updates/contrib amd64 Packages [2,506 B]
==> debian-8: Get:10 http://cloudfront.debian.net jessie Release.gpg [2,373 B]
==> debian-8: Get:11 http://security.debian.org jessie/updates/non-free amd64 Packages [14 B]
==> debian-8: Get:12 http://security.debian.org jessie/updates/contrib Translation-en [1,211 B]
==> debian-8: Get:13 http://security.debian.org jessie/updates/main Translation-en [135 kB]
==> debian-8: Get:14 http://cloudfront.debian.net jessie Release [148 kB]
==> debian-8: Get:15 http://security.debian.org jessie/updates/non-free Translation-en [14 B]
[...]

PR #5346, opened 2016-05-23
Fixes debian-8 bootstrap plus add specific platform dependent loading scripts

purbon:
Hi,
 this PR provides a fix for #5338 by providing specific bootstrap script for debian-8, it also include small change necessary to make sure all packages use the new schema name (see #5100 for details).

This PR also adds:

* custom script for `sles-12` to make sure it gets the latest java8, including also a general suse bootstrap script to download latest rpm for it.
* custom script for `ubuntu-1404` to make sure it updates the CA certificates, basically providing the solution need for #5337

_Important:_

This way this PR solve the specific bootstrap scripts is different than the one proposed in #5337, before moving forward with both we should make sure to get a common solution for it.

### DIFF
--- a/qa/platform_config.rb
+++ b/qa/platform_config.rb
@@ -3,7 +3,7 @@ require "json"
 
 class PlatformConfig
 
-  Platform = Struct.new(:name, :box, :type)
+  Platform = Struct.new(:name, :box, :type, :specific)
 
   DEFAULT_CONFIG_LOCATION = File.join(File.dirname(__FILE__), "config", "platforms.json").freeze
 
@@ -15,7 +15,7 @@ class PlatformConfig
 
     data = JSON.parse(File.read(@config_path))
     data["platforms"].each do |k, v|
-      @platforms << Platform.new(k, v["box"], v["type"])
+      @platforms << Platform.new(k, v["box"], v["type"], v["specific"])
     end
     @platforms.sort! { |a, b| a.name <=> b.name }
     @latest = data["latest"]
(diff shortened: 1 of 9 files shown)


qa/platform_config.rb
```diff
@@ -1,9 +1,34 @@
...
+
+    def initialize_bootstrap_scripts(data)
+      @bootstrap = OpenStruct.new(:privileged     => "sys/#{type}/bootstrap.sh",
+                                  :non_privileged => "sys/#{type}/user_bootstrap.sh")
+      ##
+      # for now the only specific boostrap scripts are ones need
+      # with privileged access level, whenever others are also
+      # required we can update this section as well with the same pattern.
+      ##
+      @bootstrap.privileged = "sys/#{type}/#{name}/bootstrap.sh" if data["specific"]
```
ph: OK with this strategy to override the default scripts

qa/platform_config.rb
```diff
@@ -1,9 +1,34 @@
...
+    def initialize(name, data)
+      @name = name
+      @box  = data["box"]
+      @type = data["type"]
+      initialize_bootstrap_scripts(data)
+    end
+
+    private
+
+    def initialize_bootstrap_scripts(data)
```
ph: Can we come up with a better name? Its more like a configure to me.

purbon: what do you prefer? something like configure_bootstrap_scripts?

ph:
Tested with
```
vagrant up debian-8
vagrant up ubuntu-1404
vagrant up sles-12
```
minor comment, LGTM

PR #5351, opened 2016-05-24
Backport the plugin manager generate command to 2.x branch.

### LINKED ISSUE #4266: It should be the option to generate a plugin skeleton using the plugin manager
It will make the plugin creator life a bit more easy if there is an option to generate a simple plugin skeleton from the plugin manager itself. This should be easy implemented based on the current example repositories and some templates.

purbon:
Add a new command that generate a plugin skeleton that can be used to create a new plugin

Fixes #4266

backport of #4190


suyograo:
LGTM

PR #5355, opened 2016-05-24
Bugfix from manual test of new plugin generator on master

wiibaa:
Happy to be one of the first to test very sweet work by @purbon 
Here are some minor corrections

* fix to allow direct execution of 'bundle install' => direct run of bundler was throwing an error due to unescaped single quote in gemspec
* make plugin name parameter required => forgetting the --name param resulted in a cryptic error "cannot call downcase() on Nil", now report that --name is mandatory
* display correct path of created files, the listing of created file did not contains subfolders (i.e lib/logstash/...)			
* use new event api in generator filter template => bundle exec rspec was failing on master for filter plugin due to usage of old event api
* fix config_name for generated input plugin => typo in the template of the input for the config_name setting

### DIFF
--- a/lib/pluginmanager/generate.rb
+++ b/lib/pluginmanager/generate.rb
@@ -4,6 +4,7 @@ require "pluginmanager/templates/render_context"
 require "erb"
 require "ostruct"
 require "fileutils"
+require "pathname"
 
 class LogStash::PluginManager::Generate < LogStash::PluginManager::Command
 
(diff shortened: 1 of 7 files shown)


suyograo:
nice catch @wiibaa. LGTM

wiibaa:
@suyograo @purbon can you please check and take care of backport to 2.x if needed ?

purbon:
First, sorry for the bugs introduced and thanks a lot to spot them and fix
them. I appreciate.

Re backport, sure. Im on it.

/purbon

PR #5359, opened 2016-05-25
Make the api port bind to a range between 9600 and 9700 (default)

### LINKED ISSUE #5326: Bind http.port to a range instead of just 9600
Can we bind the `http.port` to a range `[9600-9700]` instead of just `9600`. This makes running multiple LS on the same  box easier. Also follows ES user experience. 

Should indirectly fix https://github.com/elastic/logstash/issues/5314

purbon:
It provides a fix for the proposal in #5326 making the webserver try to bind to the first free port between the range 9600 and 9700 (not included). If the port is specified in the settings file it used it.

Fixes #5326

### DIFF
--- a/logstash-core/lib/logstash/webserver.rb
+++ b/logstash-core/lib/logstash/webserver.rb
@@ -8,19 +8,20 @@ require "puma/commonlogger"
 module LogStash 
   class WebServer
 
+    class HostBindingError < StandardError; end
+
     extend Forwardable
 
     attr_reader :logger, :status, :config, :options, :cli_options, :runner, :binder, :events
 
     def_delegator :@runner, :stats
 
-    DEFAULT_HOST = "127.0.0.1".freeze
-    DEFAULT_PORT = 9600.freeze
+    DEFAULT_PORT_RANGE=(9600...9700).freeze
 
     def initialize(logger, options={})
       @logger      = logger
-      http_host    = options[:http_host] || DEFAULT_HOST
-      http_port    = options[:http_port] || DEFAULT_PORT
+      http_host    = options[:http_host]
+      http_port    = options[:http_port] || pick_default_port(http_host, DEFAULT_PORT_RANGE)
       @options     = {}
       @cli_options = options.merge({ :rackup => ::File.join(::File.dirname(__FILE__), "api", "init.ru"),
                                      :binds => ["tcp://#{http_host}:#{http_port}"],
@@ -30,8 +31,10 @@ module LogStash
                                      # https://github.com/puma/puma/pull/640 for mode internal details in PUMA.
                                      :queue_requests => false
       })
-      @status      = nil
 
+      logger.terminal("Binding Logstash WebAPI to tcp://#{http_host}:#{http_port}")
+
+      @status      = nil
       parse_options
 
       @runner  = nil
@@ -79,6 +82,17 @@ module LogStash
     end
 
     private
+
+    def pick_default_port(http_host, range=(9600...9700))
+      range.step(1) do |current_port|
+        begin
+          TCPServer.new(http_host, current_port)
+          return current_port
+        rescue Errno::EADDRINUSE
+        end
+      end
+      raise HostBindingError.new("Range #{range} is full")
+    end
 
     def env
       @options[:debug] ? "development" : "production"
--- a/logstash-core/lib/logstash/webserver.rb
+++ b/logstash-core/lib/logstash/webserver.rb
@@ -32,7 +32,7 @@ module LogStash
                                      :queue_requests => false
       })
 
-      logger.terminal("Binding Logstash WebAPI to tcp://#{http_host}:#{http_port}")
+      logger.terminal("Binding Logstash HTTP server to #{http_host}:#{http_port}")
 
       @status      = nil
       parse_options
@@ -89,9 +89,10 @@ module LogStash
           TCPServer.new(http_host, current_port).close
           return current_port
         rescue Errno::EADDRINUSE
+          log("Address #{http_host}:#{current_port} in use")
         end
       end
-      raise HostBindingError.new("Range #{range} is full")
+      raise HostBindingError.new("Unable to bind to the specified port range (#{range}) for HTTP server")
     end
 
     def env
--- a/logstash-core/lib/logstash/webserver.rb
+++ b/logstash-core/lib/logstash/webserver.rb
@@ -16,7 +16,7 @@ module LogStash
 
     def_delegator :@runner, :stats
 
-    DEFAULT_PORT_RANGE=(9600...9700).freeze
+    DEFAULT_PORT_RANGE=(9600..9700).freeze
 
     def initialize(logger, options={})
       @logger      = logger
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/webserver.rb
```diff
@@ -8,19 +8,20 @@
...
 
+    class HostBindingError < StandardError; end
+
     extend Forwardable
 
     attr_reader :logger, :status, :config, :options, :cli_options, :runner, :binder, :events
 
     def_delegator :@runner, :stats
 
-    DEFAULT_HOST = "127.0.0.1".freeze
```
suyograo: Why can't we leave this as is?

logstash-core/lib/logstash/webserver.rb
```diff
@@ -30,8 +31,10 @@ def initialize(logger, options={})
                                      # https://github.com/puma/puma/pull/640 for mode internal details in PUMA.
                                      :queue_requests => false
       })
-      @status      = nil
 
+      logger.terminal("Binding Logstash WebAPI to tcp://#{http_host}:#{http_port}")
```
suyograo: Can we change to  `Binding Logstash HTTP server to #{http_host}:#{http_port}"` We could be adding more than web API in the future..

logstash-core/lib/logstash/webserver.rb
```diff
@@ -80,6 +83,17 @@ def stop(options={})
 
     private
 
+    def pick_default_port(http_host, range=(9600...9700))
+      range.step(1) do |current_port|
+        begin
+          TCPServer.new(http_host, current_port)
+          return current_port
+        rescue Errno::EADDRINUSE
+        end
```
suyograo: Add a debug statement with the tried port?

logstash-core/lib/logstash/webserver.rb
```diff
@@ -80,6 +83,17 @@ def stop(options={})
...
+      end
+      raise HostBindingError.new("Range #{range} is full")
```
suyograo: `Unable to bind to the specified port range (#{range}) for HTTP server`

logstash-core/lib/logstash/webserver.rb
purbon: we can, but is confusing, most of this defaults are also defined in the settings side, so if we change one, but not the other is confusing, don't you think?

suyograo:
Does not work with the default port. I created a tar package with no changes to default and it didn't work

`http.port: 9600`

```
suyog@machine:/tmp/logstash-5.0.0.dev$ bin/logstash -e 'input { stdin {} } output { stdout {} }'
Pipeline main started
Binding Logstash WebAPI to tcp://127.0.0.1:9600
An unexpected error occurred! {:error=>#<Errno::EADDRINUSE: Address already in use - bind - Address already in use>, :backtrace=>["org/jruby/ext/socket/RubyTCPServer.java:118:in `initialize'", "org/jruby/RubyIO.java:871:in `new'", "/tmp/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/binder.rb:234:in `add_tcp_listener'", "/tmp/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/binder.rb:98:in `parse'", "org/jruby/RubyArray.java:1613:in `each'", "/tmp/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/binder.rb:84:in `parse'", "/tmp/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/runner.rb:120:in `load_and_bind'", "/tmp/logstash-5.0.0.dev/vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/single.rb:79:in `run'", "/tmp/logstash-5.0.0.dev/vendor/local_gems/776c9074/logstash-core-5.0.0.dev-java/lib/logstash/webserver.rb:53:in `run'", "/tmp/logstash-5.0.0.dev/vendor/local_gems/776c9074/logstash-core-5.0.0.dev-java/lib/logstash/agent.rb:137:in `start_webserver'"], :level=>:fatal}
```

If I specify the port explicitly, it works:

```
suyog@machine:/tmp/logstash-5.0.0.dev$ bin/logstash -e 'input { stdin {} } output { stdout {} }' --http.port=9605
Pipeline main started
Binding Logstash WebAPI to tcp://127.0.0.1:9605
```

Also, the implicit range should be [9600-9700] which should be in the code and not configurable.. ES chooses `[9200-9300]` https://www.elastic.co/guide/en/elasticsearch/reference/master/modules-http.html#_settings_2

logstash-core/lib/logstash/webserver.rb
suyograo: ok, its good as you have it in your PR

purbon:
addressed all your comments. thanks a lot for your review.

logstash-core/lib/logstash/webserver.rb
```diff
@@ -80,6 +83,18 @@ def stop(options={})
 
     private
 
+    def pick_default_port(http_host, range=(9600..9700))
+      range.step(1) do |current_port|
+        begin
+          TCPServer.new(http_host, current_port).close
```
suyograo: hmm, this will not work.. isn't there a race condition across 2 instance? What if LS instance A checked the port for free-ness and returned ok, but before it actually acquires the port the other LS instance does the same? I think the check and acquire port should happen at the same time..

purbon: I understand your concerns, here, however I don't see a resonable way to add this behaviour inside puma. The only feasible way to make this happen, without any risk of race condition would be to use a system call inside puma, other than that there is no real way to prevent actually to instances to check for the same resource isn't? (actually talking about two codebase copies, isn't?

what do you think?

logstash-core/lib/logstash/webserver.rb
```diff
@@ -80,6 +83,18 @@ def stop(options={})
 
     private
 
+    def pick_default_port(http_host, range=(9600..9700))
```
jsvd: this should be `pick_default_port(http_host, range=DEFAULT_PORT_RANGE)`

purbon: agreed

logstash-core/lib/logstash/webserver.rb
jsvd: I'd also prefer that the bind-fail-next-bind loop was done with puma itself instead of looping before and being vulnerable to a race condition.

it should be simple to wrap the puma init code in an rescue/retry block, no?

purbon: ah! now I understand, you prefer to actually make puma up and down, might try that, however seems a bit too much (bring puma up and down) for the whole thing, will see If I see a simpler way to implement this.

jsvd: well, usually it should happen at most once or twice, unless you're starting dozens of LS instances in the same box

purbon: I can implement it like this, no problem. however it looks like too much brute force as this will also check not just when logstash runs, but also if they have other devices using one port in our range (and this we never know). I'm sure we will never use the full set, but a few might happen. 

with acking puma to boot to check this we might prevent a possible race condition, but will have a slowdown. 

I'm fine with it.

purbon:
Closing this in benefit of being implemented after #5307 it gets merged as the logic requested is going to be easier to be implemented there.

@andrewvc might be you can borrow some code from here to make this change algo happen in your PR, isn't?

Issue #5361, opened 2016-05-25
stdin input plugin specs failure


colinsurprenant:
stdin input plugin  specs are failing with the error below. I suspect this will also affect other plugins.

The problem is that the settings setup at https://github.com/elastic/logstash/blob/5e5c0c0844e1c4e833bf6bcc23ac588c2f4a1f63/logstash-core/lib/logstash/environment.rb#L16  uses `Environment::LOGSTASH_HOME` which comes from bootstrap but has not been loaded when required from the plugins rspec context. 

There is a missing `bootstrap/environment.rb` require in the code path. At this point not sure where it should be added. 

To reproduce simply run rspec on stdin input plugin in master.

```
NameError: uninitialized constant LogStash::Environment
    const_missing at org/jruby/RubyModule.java:2743
         LogStash at /Users/colin/dev/src/elasticsearch/logstash-plugins/logstash-input-stdin/vendor/gems/logstash-core-5.0.0.alpha3.snapshot4-java/lib/logstash/environment.rb:8
           (root) at /Users/colin/dev/src/elasticsearch/logstash-plugins/logstash-input-stdin/vendor/gems/logstash-core-5.0.0.alpha3.snapshot4-java/lib/logstash/environment.rb:6
          require at org/jruby/RubyKernel.java:1040
           (root) at /Users/colin/dev/src/elasticsearch/logstash-plugins/logstash-input-stdin/vendor/gems/logstash-devutils-1.0.1-java/lib/logstash/devutils/rspec/spec_helper.rb:1
          require at org/jruby/RubyKernel.java:1040
           (root) at /Users/colin/dev/src/elasticsearch/logstash-plugins/logstash-input-stdin/vendor/gems/logstash-devutils-1.0.1-java/lib/logstash/devutils/rspec/spec_helper.rb:16
             load at org/jruby/RubyKernel.java:1059
           (root) at /Users/colin/dev/src/elasticsearch/logstash-plugins/logstash-input-stdin/spec/inputs/stdin_spec.rb:1
             each at org/jruby/RubyArray.java:1613
           (root) at /Users/colin/dev/src/elasticsearch/logstash-plugins/logstash-input-stdin/spec/inputs/stdin_spec.rb:2
           (root) at /Users/colin/dev/src/elasticsearch/logstash-plugins/logstash-input-stdin/vendor/gems/rspec-core-3.4.4/lib/rspec/core/configuration.rb:1
[...]

PR #5363, opened 2016-05-26
Initialize the `path.settings` inside the runner class

ph:
The `path.settings` requires the LOGSTASH_HOME constant to be defined,
the problem is that constant is only defined when you are actually
inside the logstash application, This was causing a bug when you were
testing plugin individually because that constant wasn't defined.

Fixes: #5361


colinsurprenant:
LGTM

PR #5367, opened 2016-05-26
Add Alpha3 labels to new content

dedemorton:
Need to flag more sections as "coming" to prevent confusion for current alpha2 users.


talevy:
LGTM

PR #5369, opened 2016-05-26
Fix failing tests on master and 5.0

suyograo:
License tests was failing because of GPL2 in mime-types.Added an
exception for this gem because v2.6.2 is triple-licensed and we pick
MIT.

Metrics test is failing because of address in use which is fixed in
PR #5307. Skipping this test for now. @ph is ok with this.

### DIFF
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -373,7 +373,7 @@ describe LogStash::Agent do
       @t.join
     end
 
-    it "resets the metric collector" do
+    xit "resets the metric collector" do
       # We know that the store has more events coming in.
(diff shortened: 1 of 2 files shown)


ph:
Lgtm

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -373,7 +373,7 @@
       @t.join
     end
 
-    it "resets the metric collector" do
+    xit "resets the metric collector" do
```
arnonhongklay: what is xit?

talevy: https://relishapp.com/rspec/rspec-core/v/2-0/docs/pending/pending-examples#temporarily-pending-by-changing-"it"-to-"xit"

PR #5376, opened 2016-05-27
Remove ps command for checking init system type

dedemorton:
Removed because the command only tells you whether systemd is used.


untergeek:
LGTM™

PR #5377, opened 2016-05-27
make the necessary changes to compile with the new package system

purbon:
Introduce necessary changes to compile with the new package system introduced in #5341

/cc @ph @untergeek thoughts of you very welcome here.


untergeek:
Excellent!  

LGTM™

Issue #5366, opened 2016-05-26
Refactor interaction between the Collector/Metrics with the WebApi/Pipeline/Plugin


ph:
Currently the Metrics use a singleton instance of the `Collector` to save or access the metric store. This make accessing it really simple but cause other issues when testing it ( see https://github.com/elastic/logstash/issues/5291)  

I suggest we do the following refactor:

- It will be the Agent responsability to create the Collector instance
- The pipeline will receive a scoped metric instance associated with right collector
- The plugin will receive a scoped metric instance from the pipeline (already done)
- The Webserver will receive an instance of the collector to access the metric store directly and **wont** register as an observer. This will give fresher result on the API.
- When a pipeline reload occur the collector will be cleared and the `PeriodPoller` restarted.
- The input metric will register to the Observer to receive notification of new snapshot.
- The input metric will use the `metric` accessor to gain access to the collector.

ph:
Because of #5400 I've switched the priority to P1 and will go ahead with the refactor.

ph:
cc @andrew

ph:
cc @andrewvc

andrewvc:
One other improvement would be to use a lazy cache to synchronize snapshots rather than a thread that does this periodically. It would fix initialization order issues and be a simpler design. We would need code like:

```ruby
def snapshot
  @snapshot_mutex.synchronize do
     return @snapshot if @snapshot.last_updated > System.nano_time 
     
   end
end
```

ph:
Even with removing the singleton, the `LogStash::Agent metrics after config reloading resets the metric collector` is still failling. Doing a bit more digging.

ph:
I think it might be related to the test case and the agent doesnt reload the new configuration.

ph:
I've moved stuff around, to know this.
- when the agent start the collector is correctly save the metric
- the metric store is correctly cleaned when a pipeline is restarted
- Some metric are correctly saved to the new metric store like the jvm stats but newly created pipeline.

ph:
faillings test fixed, thread safety when we were creating pipelining and settings the metric.

ph:
I finished doing this refactor, I will run the test on a dedicated box to make sure we dont have any issue like #5400.

PR #5381, opened 2016-05-27
More node stats

andrewvc:
This resolves the non-plugin portion of #4923 

Fixes https://github.com/elastic/logstash/issues/5475

It also DRYs up some of the metrics/API code.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -46,7 +46,6 @@ class LogStash::Agent
 
     @collect_metric = setting("metric.collect")
 
-
     # Create the collectors and configured it with the library
     configure_metrics_collectors
   end
--- a/logstash-core/lib/logstash/api/commands/base.rb
+++ b/logstash-core/lib/logstash/api/commands/base.rb
@@ -19,6 +19,10 @@ module LogStash
         def started_at
           (LogStash::Agent::STARTED_AT.to_f * 1000.0).to_i
         end
+
+        def extract_metrics(path, *keys)
+          service.extract_metrics(path, *keys)
+        end
       end
     end
   end
--- a/logstash-core/lib/logstash/api/commands/node.rb
+++ b/logstash-core/lib/logstash/api/commands/node.rb
@@ -1,0 +1,117 @@
+require "logstash/api/commands/base"
+
+module LogStash
+  module Api
+    module Commands
+      class Node < Commands::Base
+        def all
+          {
+            :pipeline => pipeline,
+            :os => os,
+            :jvm => jvm,
+            :hot_threads => hot_threads
+          }           
+        end
+        
+        def pipeline
+          extract_metrics(
+            [:stats, :pipelines, :main, :config],
+            :workers, :batch_size, :batch_delay
+          )
+        end
+
+        def os
+          {
+            :name => java.lang.System.getProperty("os.name"),
+            :arch => java.lang.System.getProperty("os.arch"),
+            :version => java.lang.System.getProperty("os.version"),
+            :available_processors => java.lang.Runtime.getRuntime().availableProcessors()
+          }
+        end
+
+        def jvm
+          memory_bean = ManagementFactory.getMemoryMXBean()
+          {
+            :pid =>  ManagementFactory.getRuntimeMXBean().getName().split("@").first.to_i,
+            :version => java.lang.System.getProperty("java.version"),
+            :vm_name => java.lang.System.getProperty("java.vm.name"),
+            :vm_version => java.lang.System.getProperty("java.version"),
+            :vm_vendor => java.lang.System.getProperty("java.vendor"),
+            :vm_name => java.lang.System.getProperty("java.vm.name"),            
+            :start_time_in_millis => started_at,
+            :mem => {
+              :heap_init_in_bytes => (memory_bean.getHeapMemoryUsage().getInit() < 0 ? 0 : memory_bean.getHeapMemoryUsage().getInit()),
+              :heap_max_in_bytes => (memory_bean.getHeapMemoryUsage().getMax() < 0 ? 0 : memory_bean.getHeapMemoryUsage().getMax()),
+              :non_heap_init_in_bytes => (memory_bean.getNonHeapMemoryUsage().getInit() < 0 ? 0 : memory_bean.getNonHeapMemoryUsage().getInit()),
+              :non_heap_max_in_bytes => (memory_bean.getNonHeapMemoryUsage().getMax() < 0 ? 0 : memory_bean.getNonHeapMemoryUsage().getMax())
+            }
+          }
+        end
+
+        def hot_threads(options={})
+          HotThreadsReport.new(self, options)
+        end
+
+        class HotThreadsReport
+          HOT_THREADS_STACK_TRACES_SIZE_DEFAULT = 10.freeze
+          
+          def initialize(cmd, options)
+            @cmd = cmd
+            filter = { :stacktrace_size => options.fetch(:stacktrace_size, HOT_THREADS_STACK_TRACES_SIZE_DEFAULT) }
+            jr_dump = JRMonitor.threads.generate(filter)
+            @thread_dump = ::LogStash::Util::ThreadDump.new(options.merge(:dump => jr_dump))
+          end
+          
+          def to_s
+            hash = to_hash
+            report =  "#{I18n.t("logstash.web_api.hot_threads.title", :hostname => hash[:hostname], :time => hash[:time], :top_count => @thread_dump.top_count )} \n"
+            report << '=' * 80
+            report << "\n"
+            hash[:threads].each do |thread|
+              thread_report = ""
+              thread_report = "#{I18n.t("logstash.web_api.
+                                hot_threads.thread_title", :percent_of_cpu_time => thread[:percent_of_cpu_time], :thread_state => thread[:state], :thread_name => thread[:name])} \n"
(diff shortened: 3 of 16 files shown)


suyograo:
LGTM, looks straightforward

logstash-core/lib/logstash/agent.rb
```diff
@@ -47,13 +47,14 @@ def initialize(settings = LogStash::SETTINGS)
     @collect_metric = setting("metric.collect")
     @metric = create_metric_collector
     @periodic_pollers = LogStash::Instrument::PeriodicPollers.new(create_metric_collector)
+    
+    start_background_services
```
andrewvc: This was put here to address #5366 but doesn't fix it completely.

logstash-core/lib/logstash/api/commands/node.rb
```diff
@@ -0,0 +1,117 @@
...
+            :mem => {
+              :heap_init_in_bytes => (memory_bean.getHeapMemoryUsage().getInit() < 0 ? 0 : memory_bean.getHeapMemoryUsage().getInit()),
+              :heap_max_in_bytes => (memory_bean.getHeapMemoryUsage().getMax() < 0 ? 0 : memory_bean.getHeapMemoryUsage().getMax()),
+              :non_heap_init_in_bytes => (memory_bean.getNonHeapMemoryUsage().getInit() < 0 ? 0 : memory_bean.getNonHeapMemoryUsage().getInit()),
+              :non_heap_max_in_bytes => (memory_bean.getNonHeapMemoryUsage().getMax() < 0 ? 0 : memory_bean.getNonHeapMemoryUsage().getMax())
+            }
+          }
+        end
+
+        def hot_threads(options={})
```
andrewvc: This was moved from Stats because it was never a Stats API call. This is its proper home.

logstash-core/lib/logstash/api/commands/stats.rb
```diff
@@ -5,17 +5,32 @@ module LogStash
   module Api
     module Commands
       class Stats < Commands::Base
-
         def jvm
-          {:threads => service.get_shallow(:jvm, :threads)}
+          {
```
andrewvc: This was refactored to use `extract_metrics` to make the mapping of metrics to field names more explicit. This way we can add metrics internally to a namespace without necessarily exposing them to the API.

logstash-core/lib/logstash/api/commands/stats.rb
```diff
@@ -34,70 +49,6 @@ def memory
               acc
             end
           }
-        end
-
-        def hot_threads(options={})
```
andrewvc: Moved to the nodes API where it should have been all along.

andrewvc:
@purbon @ph some comment would be welcome here! We won't get passing tests till we address #5366

logstash-core/lib/logstash/api/commands/base.rb
```diff
@@ -19,6 +19,43 @@ def uptime
...
+        #   [:cpu, [:total_in_millis, :percent]]
+        # )
+        # 
+        # Returns:
+        # # From the jvm.process metrics namespace
+        # {
+        #   :open_file_descriptors => 123
+        #   :cpu => { :total_in_millis => 456, :percent => 789 }
+        # }
+        def extract_metrics(path, *keys)
```
purbon: might this function something to be useful for other consumers of the api?

logstash-core/lib/logstash/api/commands/node.rb
```diff
@@ -0,0 +1,117 @@
...
+          {
+            :pid =>  ManagementFactory.getRuntimeMXBean().getName().split("@").first.to_i,
+            :version => java.lang.System.getProperty("java.version"),
+            :vm_name => java.lang.System.getProperty("java.vm.name"),
+            :vm_version => java.lang.System.getProperty("java.version"),
+            :vm_vendor => java.lang.System.getProperty("java.vendor"),
+            :vm_name => java.lang.System.getProperty("java.vm.name"),            
+            :start_time_in_millis => started_at,
+            :mem => {
+              :heap_init_in_bytes => (memory_bean.getHeapMemoryUsage().getInit() < 0 ? 0 : memory_bean.getHeapMemoryUsage().getInit()),
```
purbon: can we make all this super long lines shorted and easier to digest? I think we should be avoiding ternary like this as longer as the line it gets more hard to follow the logic.

logstash-core/spec/api/lib/api/support/resource_dsl_methods.rb
```diff
@@ -14,7 +14,13 @@ def hash_to_mapping(h, path=[], mapping={})
 
   def test_api(expected, path)
     context "GET #{path}" do
-      let(:payload) { LogStash::Json.load(last_response.body) }
+      let(:payload) do
+        begin
+          LogStash::Json.load(last_response.body)
+        rescue LogStash::Json::ParserError => e
+          require 'pry'; binding.pry
```
purbon: pry when something wrong happens? I think this might be good for debugging but can cause unnecessary problems if raised unintentional

logstash-core/lib/logstash/api/rack_app.rb
```diff
@@ -90,6 +90,8 @@ def self.app(logger, environment)
           run LogStash::Api::Modules::Root
           namespaces.each_pair do |namespace, app|
             map(namespace) do
+
```
purbon: empty?

logstash-core/lib/logstash/api/commands/base.rb
andrewvc: Yes. @ph proposed moving this into the collector. I agree with him on this. I'll wait till #5366 is addressed before I attempt that move.

logstash-core/lib/logstash/api/commands/node.rb
andrewvc: I agree that it isn't so great to read, but I think its better than splitting into a separate line where it takes up less horizontal but more vertical space.

I'm basically stealing this from https://github.com/elastic/elasticsearch/blob/34da01f794ba892176376aafc18e901cacdd0fce/core/src/main/java/org/elasticsearch/monitor/jvm/JvmInfo.java#L72

I feel like the ES team probably made the same tradeoff

logstash-core/lib/logstash/api/rack_app.rb
andrewvc: good catch!

logstash-core/spec/api/lib/api/support/resource_dsl_methods.rb
andrewvc: That is a bug! I def need to remove this :). Thanks for finding this.

andrewvc:
@ph @purbon ready for another round of review!

suyograo:
@andrewvc agreed that `hot_threads` should be under `_node`, but I wonder if we should omit it when we call `GET curl localhost:9600/_node?pretty` and only show it when someone explicitly calls `GET curl localhost:9600/_node/hot_threads?pretty`

suyograo:
Also, would you mind moving `_plugins` from `localhost:9600/_plugins` to `localhost:9600/_node/plugins` as noted in #5475 ?

suyograo:
Tested this PR, info looks good.

```
suyog@machine:~$ curl localhost:9600/_node?pretty
{
  "pipeline" : {
    "workers" : 8,
    "batch_size" : 125,
    "batch_delay" : 5
  },
  "os" : {
    "name" : "Mac OS X",
    "arch" : "x86_64",
    "version" : "10.11.1",
    "available_processors" : 8
  },
  "jvm" : {
    "pid" : 92317,
    "version" : "1.8.0_65",
    "vm_name" : "Java HotSpot(TM) 64-Bit Server VM",
    "vm_version" : "1.8.0_65",
    "vm_vendor" : "Oracle Corporation",
    "start_time_in_millis" : 1465867138764,
    "mem" : {
      "heap_init_in_bytes" : 268435456,
      "heap_max_in_bytes" : 1037959168,
      "non_heap_init_in_bytes" : 2555904,
      "non_heap_max_in_bytes" : 0
    }
  },
```

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -110,6 +110,54 @@ def get_shallow(*key_paths)
...
+          # This is a Metric object, we need to extract its value
+          # If the metric didn't exist it might be nil, but we still want its key
+          # to exist with a nil value
+          acc[k] = v ? v.value : nil; acc
+        end
+        
+        acc
+      end
+    end    
+
```
ph: Thanks for moving this inside the metric store

ph: Can we add a unit test for it?

ph:
Tested this PR, info looks goods.
Only a minor comments concerning `extract_metrics`, I think we should have test covering it, its a recursive method and I would be more confident if we have test targetting it.

andrewvc:
@ph I've added extensive unit tests to `extract_metrics` and fully refactored it to make it act correctly with deep metrics and weird stuff like `[:foo, [:pipeline1, :pipeline2], :common]`

logstash-core/lib/logstash/instrument/metric_store.rb
andrewvc: Done!

andrewvc:
@suyograo @ph I believe I've made all required changes. How's it looking?

suyograo:
Tested, LGTM from my end.

andrewvc:
@ph can I get an LGTM from you to proceed

ph:
Tested L G T M

PR #5383, opened 2016-05-27
Fix starting logstash from command-line

### LINKED ISSUE #5379: Set path.config to point to /etc/logstash/conf.d for release packages
When LS is installed as a service, the post install script needs to overwrite `path.config` to point to `/etc/logstash/conf.d`

Otherwise you get this error:

```
`ERROR: No configuration file was specified. Perhaps you forgot to provide the '-f yourlogstash.conf' flag?

```

untergeek:
This fix is for packages only (RPM/DEB)

This puts the requisite changes into /etc/logstash/logstash.yml so that starting logstash from the command-line, e.g. `/usr/share/logstash/bin/logstash --path.settings /etc/logstash` will read the pipeline from `/etc/logstash/conf.d` and log to `/var/log/logstash/logstash.log`.

fixes #5379

### DIFF
--- a/config/jvm.options
+++ b/config/jvm.options
@@ -40,7 +40,7 @@
 ## basic
 
 # set the I/O temp directory
--Djava.io.tmpdir=$HOME
+#-Djava.io.tmpdir=$HOME
 
 # set to headless, just in case
 -Djava.awt.headless=true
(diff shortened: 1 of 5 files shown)


suyograo:
LGTM

suyograo:
LGTM for new commit.

PR #5387, opened 2016-05-28
Restructure getting started content

dedemorton:
@acchen97 @suyograo Here is the first round of changes to the getting started guide. I created a temporary build so you can see the built docs: https://logstashgs.firebaseapp.com/ 

What's remaining: I need to create a more interesting set of examples in the section about multiple inputs and outputs (to cover other stages that Alvin suggested). The use case should show how Logstash unifies the data; right now it's just an odd assortment of stages.

Major changes:
* Streamlined content to remove configuration details (like TLS) that were too advanced for getting started content. 
* Moved content not related to getting started to other sections. 
* Refined examples because they didn't work exactly as written (and the returned docs did not match the examples).
* Included a few comments to explain why I did some things - take a look if you have questions.

Note that I could not test the file input or output because it was broken in the snapshot. But at least all the other problems are fixed.


docs/static/advanced-pipeline.asciidoc
```diff
@@ -107,7 +101,9 @@ Bytes served:: `bytes`
 Referrer URL:: `referrer`
 User agent:: `agent`
 
-Edit the `first-pipeline.conf` file to add the following text:
+//QUESTION: The name/value pairs above are backwards. Is this really how grok parses the data, or is this a mistake?
```
jsvd: humm..I don't follow what you mean here by backwards, can you explain? thanks

docs/static/advanced-pipeline.asciidoc
```diff
@@ -145,60 +134,75 @@ After processing, the sample line has the following JSON representation:
...
 
-With this configuration, Logstash uses http protocol to connect to Elasticsearch. The above example assumes Logstash
-and Elasticsearch to be running on the same instance. You can specify a remote Elasticsearch instance using `hosts`
-configuration like `hosts => "es-machine:9092"`.
+The `geoip` plugin configuration requires you to specify the name of the source field that contains the IP address to look up. In this example, the `clientip` field contains the IP address.
 
-[float]
-[[configuring-geoip-plugin]]
-===== Enhancing Your Data with the Geoip Filter Plugin
+Make sure that the `geoip` section is after the `grok` section of the configuration file and that both the `grok` and
```
jsvd: maybe.. "Since filters are evaluated in sequence, make sure.."?

docs/static/advanced-pipeline.asciidoc
```diff
@@ -145,60 +134,75 @@ After processing, the sample line has the following JSON representation:
...
+Elasticsearch cluster. Edit the `first-pipeline.conf` file and replace the entire `output` section with the following
+text:
 
 [source,json]
 --------------------------------------------------------------------------------
-geoip {
-    source => "clientip"
+output {
+    elasticsearch {
+        hosts => "localhost:9200"
```
jsvd: although it is not necessary, I wonder if we should use `hosts => [ "localhost:9200" ]` to encourage the idea of more nodes..

docs/static/advanced-pipeline.asciidoc
```diff
@@ -234,212 +239,232 @@ Try a test query to Elasticsearch based on the fields created by the `grok` filt
...
+<1> Absolute path to the file or files that Filebeat processes.
+<2> Adds a field called `type` with the value `syslog` to the event.
 
-Save this configuration file as `filebeat.yml`.
+Save your changes. 
+
+To keep the example configuration simple, you won't specify TLS/SSL settings as you would in a real world
+scenario.
+
+//NOTE: I'VE REMOVED THE CA SETTINGES BECAUSE SETTING UP A CA AND GENERATING CERTIFICATES IS TOO COMPLEX FOR A GETTING STARTED GUIDE. A SIMPLE CONFIG IS BETTER FOR TEACHING CONCEPTS.
```
jsvd: 👍 it made sense before with logstash-forwarder because ssl was mandatory, but not anymore

docs/static/advanced-pipeline.asciidoc
dedemorton: @jsvd I only meant that shouldn't the Field Name column should be the first column in the table. Seems backwards.

suyograo:
LGTM

dedemorton:
Adds fix for the getting started doc items tracked in #5148

PR #5388, opened 2016-05-30
Remove unnecessary dependancy to rake

purbon:
Initially the dependency to rake was introduced, according to @ph comments in #5259 to reuse the download function from devutils, however the code to download is actually done using wget, see

```
    def download(from, to, host)
      run_command("wget #{from} -O #{to}", host)
    end
```

as the dependency to rake actually make using `bundle exec rspec` syntax impossible is better to get it removed and can always come back when we approach proper reuse of devutils code.

//cc @ph @untergeek your thoughts are very welcome

### DIFF
--- a/qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
@@ -2,7 +2,6 @@
 require_relative "../../../spec_helper"
 require "logstash/version"
 require "fileutils"
-require "logstash/devutils/rake"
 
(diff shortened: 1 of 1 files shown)


ph:
Why we have so many commits for that change?

purbon:
Needs rebase :-/

ph:
@purbon Agree, but let rebase that I will take a look after ;)

purbon:
Sure, no problem! I understand.

purbon:
@ph rebased 👍

ph:
lgtm

Issue #5371, opened 2016-05-26
Unskip metrics test after address in use issues are resolved.


suyograo:
Disabled test: https://github.com/elastic/logstash/blob/master/logstash-core/spec/logstash/agent_spec.rb#L376 needs to be enabled.

```
LogStash::Agent metrics after config reloading resets the metric collector
     Failure/Error: Unable to find matching line from backtrace
     Errno::EADDRINUSE:
       Address already in use - bind - Address already in use
     # ./vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/binder.rb:234:in `add_tcp_listener'
     # ./vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/binder.rb:98:in `parse'
     # ./vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/binder.rb:84:in `parse'
     # ./vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/runner.rb:120:in `load_and_bind'
     # ./vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/single.rb:79:in `run'
     # ./vendor/bundle/jruby/1.9/gems/logstash-core-5.0.0.alpha3.snapshot5-java/lib/logstash/webserver.rb:50:in `run'
     # ./vendor/bundle/jruby/1.9/gems/logstash-core-5.0.0.alpha3.snapshot5-java/lib/logstash/agent.rb:136:in `start_webserver'
```

http://build-eu-00.elastic.co/job/logstash_regression_50/jdk=JDK8,label=metal-pool/133/console

PR which will fix: https://github.com/elastic/logstash/pull/5307

ph:
cc @andrewvc

PR #5392, opened 2016-05-30
Remove XIT for metric test in the agent_spec.

ph:
The XIT was added to make sure we were able to build LS packages.
This PR revert the xit since the bug was fixed in #5307

Fixes: #5371


jsvd:
LGTM

PR #5393, opened 2016-05-30
abort on incorrect path.settings value

### LINKED ISSUE #5389: logstash.yml file not found if run the logstash command directly
If installing logstash as a package, but then using the logstash command directly, for example to check a config or to check for a version, it can not find the config file.

see 

```ruby
vagrant@packer-virtualbox-iso-1452039329:~$ sudo /usr/share/logstash/bin/logstash --version
Errno::ENOENT: No such file or directory - /usr/share/logstash/config/logstash.yml
       read at org/jruby/RubyIO.java:3804
       read at org/jruby/RubyIO.java:3987
  read_yaml at /usr/share/logstash/vendor/local_gems/47abefa3/logstash-core-5.0.0.dev-java/lib/logstash/settings.rb:100
  from_yaml at /usr/share/logstash/vendor/local_gems/47abefa3/logstash-core-5.0.0.dev-java/lib/logstash/settings.rb:94
        run at /usr/share/logstash/vendor/local_gems/47abefa3/logstash-core-5.0.0.dev-java/lib/logstash/runner.rb:141
        run at /usr/share/logstash/vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:132
     (root) at /usr/share/logstash/lib/bootstrap/environment.rb:66
vagrant@packer-virtualbox-iso-1452039329:~$ 

```

logstash expect to have the file in the LS_HOME location, `/usr/share/logstash` however it also has one in `/etc/logstash` and not sure if users will also think that when running as a command it will pick up a config file in the $HOME directory.

I guess is resonable to think about having only 1 reference to the config file and this could be the one inside `/etc/logstash`. 

This has been tested with the `deb` package, but will also have the same behaviour with the rpm I guess.

//cc @jsvd this is the bug I told you. @untergeek might be also of your interest.

jsvd:
fixes #5389

example output:

```
~/projects/logstash (git)-[master] % bin/logstash --version
Warning: Could not find logstash.yml in 'path.settings' (value: "/Users/joaoduarte/projects/logstash/config"), using internal defaults.
logstash 5.0.0.dev
```

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -139,9 +139,12 @@ class LogStash::Runner < Clamp::StrictCommand
     @settings.set("path.settings", settings_path) if settings_path
 
     begin
-    LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("path.settings"))
+      LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("path.settings"))
     rescue => e
-      puts "Warning: Could not find logstash.yml in 'path.settings' (value: \"#{LogStash::SETTINGS.get("path.settings")}\"), using internal defaults."
+      @logger.subscribe(STDOUT)
+      @logger.warn("Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in \"--path.settings=/path/..\" in the command line options")
+      @logger.fatal("Failed to load settings file from \"path.settings\". Aborting...", "path.settings" => LogStash::SETTINGS.get("path.settings"), "exception" => e.class, "message" => e.message)
+      exit(-1)
     end
 
     super(*[args])
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -138,7 +138,12 @@ class LogStash::Runner < Clamp::StrictCommand
 
     @settings.set("path.settings", settings_path) if settings_path
 
+    begin
     LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("path.settings"))
+    rescue => e
+      puts "Warning: Could not find logstash.yml in 'path.settings' (value: \"#{LogStash::SETTINGS.get("path.settings")}\"), using internal defaults."
+    end
+
     super(*[args])
   end
 


logstash-core/lib/logstash/runner.rb
```diff
@@ -138,7 +138,12 @@ def run(args)
 
     @settings.set("path.settings", settings_path) if settings_path
 
+    begin
     LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("path.settings"))
+    rescue => e
```
ph: Should we be more strict on the catched exception?

suyograo:
I don't think we should use defaults here when the settings are missing. This is just confusing. I agree completely about having a better error message, but failing fast is ideal here, IMO. We can augment the message to say if you've used packages, we look for the settings file in '/etc/logstash'. Thoughts?

purbon:
@suyograo then this is more solution #2 from https://github.com/elastic/logstash/issues/5389, I guess we should discuss there, isn't?

on the other hand, in #5389 is explained, problem is when using a LS installed with packaged, but calling the logstash command directly.

jsvd:
pivoted this pr in conformity with discussion in #5389

logstash-core/lib/logstash/runner.rb
```diff
@@ -138,7 +138,15 @@ def run(args)
 
     @settings.set("path.settings", settings_path) if settings_path
 
-    LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("path.settings"))
+    begin
+      LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("path.settings"))
+    rescue => e
```
suyograo: Do we catch path not found? is it possible?

jsvd:
```
% bin/logstash --version --path.settings=/ewrwrwe/rw/rew/rwe/
Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in "--path.settings=/path/.." in the command line options {:level=>:warn}
Failed to load settings file from "path.settings". Aborting... {"path.settings"=>"/ewrwrwe/rw/rew/rwe/", "exception"=>Errno::ENOENT, "message"=>"No such file or directory - /ewrwrwe/rw/rew/rwe/logstash.yml", :level=>:fatal}
```

jsvd: this should also catch stuff like permission denied,  invalid yaml file, etc

suyograo: sounds good.

suyograo:
LGTM

PR #5399, opened 2016-05-31
pass the settings file location as required by the new package structure

purbon:
Fix the version test to actually give the setting file location, this is due to the requirements introduced by the new package schema.


untergeek:
I think it would be wiser to not require `--path.settings` when asking for `--version`

purbon:
agreed, but until we do that change we've to live with it also here 👍 , followup issue coming

untergeek:
LGTM™ as an interim solution.

PR #5401, opened 2016-05-31
Kryo byte serialization

guyboertje:
Notes to reviewers:

- This is an initial naive implementation.
- I believe it is enough to give Colin something to use for PQ in tests and benchmarks.


logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -253,16 +297,16 @@ private Timestamp initTimestamp(Object o) {
             } else if (o instanceof String) {
                 // second most frequent
                 return new Timestamp((String) o);
-            } else if (o instanceof JrubyTimestampExtLibrary.RubyTimestamp) {
-                return new Timestamp(((JrubyTimestampExtLibrary.RubyTimestamp) o).getTimestamp());
+//            } else if (o instanceof JrubyTimestampExtLibrary.RubyTimestamp) {
+//                return new Timestamp(((JrubyTimestampExtLibrary.RubyTimestamp) o).getTimestamp());
```
talevy: why?

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -253,16 +297,16 @@ private Timestamp initTimestamp(Object o) {
...
             } else if (o instanceof Timestamp) {
                 return new Timestamp((Timestamp) o);
             } else if (o instanceof DateTime) {
                 return new Timestamp((DateTime) o);
             } else if (o instanceof Date) {
                 return new Timestamp((Date) o);
-            } else if (o instanceof RubySymbol) {
-                return new Timestamp(((RubySymbol) o).asJavaString());
+//            } else if (o instanceof RubySymbol) {
```
talevy: why?

logstash-core-event-java/src/jmh/java/com/logstash/ThreadStateBase.java
```diff
@@ -0,0 +1,106 @@
+package com.logstash;
+
+import java.math.BigDecimal;
+import java.util.HashMap;
+import java.util.Map;
+
+public abstract class ThreadStateBase {
```
talevy: I am not sure about all of the necessity of this class, maybe missing something about how Kyro serializes things in a multithreaded manner?

logstash-core-event-java/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Fri Jan 22 14:29:02 EST 2016
+#Mon May 16 10:40:54 BST 2016
```
talevy: nice 🇬🇧

logstash-core-event-java/build.gradle
```diff
@@ -92,11 +99,17 @@ idea {
...
+    compile 'com.fasterxml.jackson.core:jackson-core:2.7.3'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.7.3'
+    compile 'com.esotericsoftware:kryo:3.0.3'
+
     provided 'org.jruby:jruby-core:1.7.22'
     testCompile 'junit:junit:4.12'
     testCompile 'net.javacrumbs.json-unit:json-unit:1.9.0'
+
+    runtime 'com.fasterxml.jackson.core:jackson-databind:2.7.4'
+    runtime 'org.jruby:jruby-complete:1.7.22'
```
talevy: we need to figure this whole jruby provided vs. runtime dependency thing.

I think it makes sense to have jmh profiling be its own submodule with its own dependencies. that way it can require what it needs and not interrupt java event's dependencies.

what do you think?

talevy:
This is cool, I am not too familiar with Kyro. I know other serialization-heavy class-passing jvm projects use it, so it is cool to see support for this in Logstash!

talevy:
@guyboertje, do you mind sharing some benchmarks from your jmh run? or leave some notes on how to execute the benchmarks?

logstash-core-event-java/build.gradle
guyboertje: Sure. My gradle foo is weak. Hence the commenting out of the JRuby bits that you comments "Why?" on below.

logstash-core-event-java/gradle/wrapper/gradle-wrapper.properties
guyboertje: Done by IntelliJ AFAIK

logstash-core-event-java/src/jmh/java/com/logstash/ThreadStateBase.java
guyboertje: This class is for JMH only and that is why it is in `src/jmh/java`.

logstash-core-event-java/src/main/java/com/logstash/Event.java
guyboertje: Could not get JMH to compile JRuby without making it a compile dependency for the normal Jar.

I will try again.

logstash-core-event-java/src/main/java/com/logstash/Event.java
guyboertje: See above reply to previous `why?`

guyboertje:
@tal, @colinsurprenant:
`./gradlew clean jmh`
executes the jmh benchmarks

suyograo:
@guyboertje sorry I accidentally closed this. Can you revive this again after a rebase?

guyboertje:
@suyograo - we are not using Kryo - this can remain closed.

PR #5410, opened 2016-05-31
remove .lock file

talevy:
The lock file in master does not seem to be used and is only lagging behind release branches


ph:
I think we can remove the lock from the master branch only, lock are still **needed** in releases branches to make sure we can reproduce a package from a release tag.

Having the lock file in the master branch, I think It give us more pain than actual benefits. Everytime that we bootstrap an environment or want to rebase the current branch we have to stash this file.
OR you can do `git update-index --no-assume-unchanged <LOCKFILE>` that make git ignore local changes to the file.

But I think **there is a dangerous thing**, this commit need to never make it to other branches.

---

A bit of history, In the real world, rails or ruby developers checkout their **lock** file if they want to make sure every developers run the same version of the dependency. Checking this file also speedup bundle install since bundler doesn't need to calculate the dependency graph.

ph:
@colinsurprenant Do you remember why we keep the lock file in the master branch? I understand the need in the releases branch but I can't remember why we decided to keep it in the master branch.

jordansissel:
Agreed, master (and perhaps all branch heads) should not have .lock files. Tagged releases should, but not branches, imo. +1 on removing gemfile lock here :)

talevy:
Hey all. thanks for chiming in here, but I am unclear as to whether you are "LGTM" with this PR

suyograo:
Lets wait for @colinsurprenant to chime here.

talevy:
@colinsurprenant. what say you?

jsvd:
btw I'm +1 on this change but curious if there's any reason behind it as well.
I actually made the same change in https://github.com/elastic/logstash/pull/5460

talevy:
@jsvd. same change, eh? OK, I am in no rush. I am happy to have your PR carry this calling. Closing in favor of refactoring present in #5460

PR #5411, opened 2016-05-31
Improve error handling in API / Rack App.

andrewvc:
Setting error handling conditions in sinatra dynamically is not possible.
The values such as `show_exceptions` get set, but have no effect.
This is esp. important in testing where we want failures to raise exceptions.
Otherwise, debugging is a matter of parsing instructable HTML responses containing
the debug info.

We do not want this behavior by default however, as doing so would mean a single API
bug would take down a production app.

This change adds a new CLI flag 'http.environment' that is defaulted to 'production'.
If set to 'test' exceptions get raised.

This change also removes sinatra error handlers that previously did nothing, and injects
custom rack middleware to properly log and handle errors.

### DIFF
--- a/logstash-core/lib/logstash/api/modules/base.rb
+++ b/logstash-core/lib/logstash/api/modules/base.rb
@@ -7,14 +7,15 @@ module LogStash
       class Base < ::Sinatra::Base
         helpers AppHelpers
 
+        # These options never change
+        # Sinatra isn't good at letting you change internal settings at runtime
+        # which is a requirement. We always propagate errors up and catch them
+        # in a custom rack handler in the RackApp class
         set :environment, :production
+        set :raise_errors, true
+        set :show_exceptions, false
 
         attr_reader :factory
-
-        if settings.environment != :production
-          set :raise_errors, true
-          set :show_exceptions, :after_handler
-        end
 
         include LogStash::Util::Loggable
 
@@ -31,12 +32,6 @@ module LogStash
           text = as == :string ? "" : {}
           respond_with(text, :as => as)
         end
-
-        error do
-          e = env['sinatra.error']
-          logger.error(e.message, :url => request.url, :ip => request.ip, :params => request.params, :class => e.class.name, :backtrace => e.backtrace)
-        end
-
       end
     end
   end
--- a/logstash-core/lib/logstash/api/rack_app.rb
+++ b/logstash-core/lib/logstash/api/rack_app.rb
@@ -1,3 +1,5 @@
+require "sinatra"
+require "rack"
 require "logstash/api/modules/base"
 require "logstash/api/modules/node"
 require "logstash/api/modules/node_stats"
@@ -8,9 +10,83 @@ require "logstash/api/modules/stats"
 module LogStash
   module Api
     module RackApp
-      def self.app
+      # Cabin is not compatible with CommonLogger, and this gives us more control anyway
+      METADATA_FIELDS = [:request_method, :path_info, :query_string, :http_version, :http_accept].freeze
+      def self.log_metadata(status, env)
+        METADATA_FIELDS.reduce({:status => status}) do |acc, field|
+          acc[field] = env[field.to_s.upcase]
+          acc
+        end        
+      end
+      
+      class ApiLogger
+        LOG_MESSAGE = "API HTTP Request".freeze
+        
+        def initialize(app, logger)
+          @app = app
+          @logger = logger
+        end
+
+        def call(env)
+          res = @app.call(env)
+          status, headers, body = res
+          
+          if fatal_error?(status)
+            @logger.warn? && @logger.warn(LOG_MESSAGE, RackApp.log_metadata(status, env))                      
+          else          
+            @logger.info? && @logger.info(LOG_MESSAGE, RackApp.log_metadata(status, env))                      
+          end
+
+          res          
+        end
+
+        def fatal_error?(status)
+          status >= 500 && status < 600
+        end
+      end
+      
+      class ApiErrorHandler
+        LOG_MESSAGE = "Internal API server error".freeze
+        
+        def initialize(app, logger)
+          @app = app
+          @logger = logger
+        end
+
+        def call(env)
+          @app.call(env)
+        rescue => e
+          body = RackApp.log_metadata(500, env).
+                   merge({
+                           :error => "Unexpected Internal Error",
+                           :class => e.class.name,
+                           :message => e.message,
+                           :backtrace => e.backtrace
+                         })
+
+          @logger.error(LOG_MESSAGE, body)
+          
+          [500,
+           {'Content-Type' => 'application/json'},
+           [LogStash::Json.dump(body)]
+          ]
+        end
+      end
+      
+      def self.app(logger, environment)
         namespaces = rack_namespaces 
         Rack::Builder.new do
+          # Custom logger object. Rack CommonLogger does not work with cabin
+          use ApiLogger, logger
+          
+          # In test env we want errors to propogate up the chain
+          # so we get easy to understand test failures.
+          # In production / dev we don't want a bad API endpoint
+          # to crash the process
+          if environment != "test"
+            use ApiErrorHandler, logger
+          end
+          
           run LogStash::Api::Modules::Root
           namespaces.each_pair do |namespace, app|
             map(namespace) do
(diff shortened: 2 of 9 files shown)


logstash-core/lib/logstash/runner.rb
```diff
@@ -115,6 +115,11 @@ class LogStash::Runner < Clamp::StrictCommand
     :attribute_name => "http.port",
     :default => LogStash::SETTINGS.get_default("http.port"), &:to_i
 
+  option ["--http.environment"], "HTTP_ENVIRONMENT",
```
suyograo: Should we even add these CLI flags now that we have `logstash.yml` file?

andrewvc: Is it possible to add a file-only setting? I'm fine with that but it has to be able to be set in the runner instantiation.

suyograo: I think so.. maybe @jsvd can comment..

andrewvc: I think that makes sense. The only reason this is a configuration option is so that it can be set [here](https://github.com/elastic/logstash/pull/5411/files#diff-e0c0c10829ae5c5711afbe0289778cebR49) by the test suite.

It does need to be set at runner instantiation time however!

logstash-core/lib/logstash/api/modules/base.rb
```diff
@@ -7,15 +7,16 @@ module Modules
       class Base < ::Sinatra::Base
         helpers AppHelpers
 
+        # These options never change
+        # Sinatra isn't good at letting you change internal settings at runtime
+        # which is a requirement. We always propagate errors up and catch them
+        # in a custom rack handler in the RackApp class
         set :environment, :production
+        set :raise_errors, true
+        set :show_exceptions, false
```
purbon: why not showing exceptions for test? might this help for test I understand, but could we hide some errors also?

logstash-core/lib/logstash/environment.rb
```diff
@@ -29,6 +29,7 @@ module LogStash
             Setting::String.new("log.format", "plain", true, ["json", "plain"]),
             Setting::String.new("http.host", "127.0.0.1"),
               Setting::Port.new("http.port", 9600),
+            Setting::String.new("http.environment", "production"),
```
purbon: why should we have this env here? talking about http env works for us (developers), but not really for end users, they should not care about it, and is also something that a release is always production, right?

logstash-core/lib/logstash/runner.rb
purbon: I don't think this flag is something the user should care about, end of the day we should be setting the hole environment (for all logstash) or nothing, this makes other moving part of logstash in strange status.

I think is better to have this variable injected somehow when we run test, or either be setup in regular mode when run from command line. Does it makes sense?

happy to provide more clarity.

logstash-core/lib/logstash/api/rack_app.rb
```diff
@@ -8,9 +8,84 @@
 module LogStash
   module Api
     module RackApp
-      def self.app
+      # Cabin is not compatible with CommonLogger, and this gives us more control anyway
+      class ApiLogger
```
purbon: I understand the motivations your are going with this solution, and not saying is bad, just a few open questions I have that would be ncie to provide closure somehow.

* What do you think about performance? as more rack middleware we use, more roundtrips the request is doing, have you thought about it? 

* This Rack applications should be properly tested as they include another layer of interaction for every request that hits the api.

logstash-core/lib/logstash/api/rack_app.rb
```diff
@@ -8,9 +8,84 @@
...
+            @logger.info? && @logger.info(LOG_MESSAGE, log_metadata(status, env))                      
+          end
+
+          res          
+        end
+
+        def log_metadata(status, env)
+          {
+            :status => status,
+            :method => env["REQUEST_METHOD"],
```
purbon: can we automate this, might be using an inject loop? I ask because it makes sense to actually log all the event.

logstash-core/lib/logstash/api/modules/base.rb
andrewvc: Show exceptions only affects the HTML returned (which is inscrutable when dumped to a terminal). We want to raise errors for tests instead so they are reported as a test failure.

logstash-core/lib/logstash/api/rack_app.rb
```diff
@@ -8,9 +8,84 @@
...
+            :path => env["REQUEST_PATH"],
+            :query_string => env["QUERY_STRING"],
+            :http_version => env["HTTP_VERSION"],
+            :http_accept => env["HTTP_ACCEPT"]
+          }
+        end
+          
+      end
+      
+      class ApiErrorHandler
```
purbon: why having two different handlers? one for http errors and another for logging exceptions?

logstash-core/lib/logstash/api/rack_app.rb
```diff
@@ -8,9 +8,84 @@
...
+        def initialize(app, logger)
+          @app = app
+          @logger = logger
+        end
+
+        def call(env)
+          @app.call(env)
+        rescue => e                   
+          body = {
+            :error => "Unexpected Internal Error",
```
purbon: do we want to wrap this in method?

logstash-core/lib/logstash/api/rack_app.rb
andrewvc: I think the performance impact here is negligible. You would need to get to thousands of TPS before this rack middleware had a measurable impact in my estimation.

I agree this still needs tests, that's why its marked as a WIP.

purbon:
good work Andrew, this is really how it should be done the logging 👍 , glad to see so many improvements here 👍

andrewvc:
@purbon I did wind up refactoring the rack middleware, but after trying to combine it into one decided to leave it as two. It would have complicated the tests more than necessary, and an extra piece of rack middleware basically costs one function call so I'm not worried about perf.

logstash-core/lib/logstash/api/rack_app.rb
```diff
@@ -8,9 +10,79 @@
...
+        def initialize(app, logger)
+          @app = app
+          @logger = logger
+        end
+
+        def call(env)
+          res = @app.call(env)
+          status, headers, body = res
+          
+          if status >= 500 && status < 600
```
ph: I would see this move into his own predicate, I did a quick look a IANA, but they dont seem to have a name for this range.
```
fatal_error?(status)
```

logstash-core/lib/logstash/environment.rb
ph: @andrewvc Can we add a comment here why it was necessary and maybe pointing to this PR? I think in a few month we will wonders why we did such a thing.

logstash-core/lib/logstash/webserver.rb
```diff
@@ -1,25 +1,25 @@
...
     extend Forwardable
 
-    attr_reader :logger, :status, :config, :options, :cli_options, :runner, :binder, :events, :http_host, :http_port
+    attr_reader :logger, :status, :config, :options, :cli_options, :runner, :binder, :events, :http_host, :http_port, :http_environment
 
     def_delegator :@runner, :stats
 
     DEFAULT_HOST = "127.0.0.1".freeze
     DEFAULT_PORT = 9600.freeze
+    DEFAULT_ENVIRONMENT = 'production'
```
ph: #freeze?

logstash-core/spec/api/lib/rack_app_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        [404, {}, ["404 Page not found"]]
+      end
+    end
+  end
+
+  let(:logger) { Cabin::Channel.get }
+
+  describe LogStash::Api::RackApp::ApiErrorHandler do
+    let(:app) do
+      # Scoping in rack builder is weird, these need to be locals
```
ph: Oh yeah ruby.

andrewvc: Yeah, that irritated me quite a bit when it didn't work at first!

andrewvc:
Jenkins, please test this

ph:
Codes LGTM, I like that we have 2 separate rack handler since they target 2 different need.

When build pass feel free to merge.

PR #5418, opened 2016-06-01
Add docs about plugin generator

suyograo:


docs/static/plugin-generator.asciidoc
```diff
@@ -0,0 +1,19 @@
+[[plugin-generator]]
+== Generating Plugins
+
+You can now create your own Logstash plugin in seconds! The generate subcommand of `bin/logstash-plugin` creates the foundation
```
purbon: might be write, `You can now create your own Logstash plugins easier!`, the in a second is don't know, strange ... because end of the day they still have to do so many other things to finish the plugin, right?

docs/static/plugin-generator.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+can start adding custom code process data with Logstash.
+
+**Example Usage**
+
+[source,sh]
+--------------------------------------------
+bin/logstash-plugin generate --type input --name xkcd --path ~/ws/elastic/plugins
+-------------------------------------------
+
+* `--type`: Type of plugin - input, filter, output and codec
```
purbon: codecs are on their way, but yes 👍

purbon:
only one small comment, in general looks awesome! good work

docs/static/plugin-generator.asciidoc
suyograo: Should be fine to document it :)

PR #5419, opened 2016-06-01
Fix argument check for the generator command.

### LINKED ISSUE #5412: Logstash Plugin Generator Allows creationg of plugin without setting --type
new logstash plugin generator errors out when creating new plugin

```
~/Code/es/talevy-logstash (master)$ bin/logstash-plugin generate --name codec-confluent-avro
 Creating /Users/tal/Code/es/talevy-logstash/logstash--codec-confluent-avro
ERROR: Plugin creation Aborted, message: No such file or directory - No such directory: /Users/tal/Code/es/talevy-logstash/lib/pluginmanager/templates/-plugin
```

Expected:

- no error
- and creating `logstash-codec-confluent-avro` instead of `logstash--codec-confluent-avro`

purbon:
* add more control on valid arguments for codecs in the generator command.

Partially fixed #5412 

follow up PR adding codecs is on its way.

### DIFF
--- a/lib/pluginmanager/generate.rb
+++ b/lib/pluginmanager/generate.rb
@@ -10,15 +10,12 @@ class LogStash::PluginManager::Generate < LogStash::PluginManager::Command
 
   TYPES = [ "input", "filter", "output" ]
 
-  option "--type", "TYPE", "Type of the plugin {input, filter, output}s" do |arg|
-    raise(ArgumentError, "should be one of: input, output or filter") unless TYPES.include?(arg)
-    arg
-  end
-
+  option "--type", "TYPE", "Type of the plugin {input, filter, output}s", :required => true
   option "--name", "PLUGIN", "Name of the new plugin", :required => true
   option "--path", "PATH", "Location where the plugin skeleton will be created", :default => Dir.pwd
 
   def execute
+    validate_params
     source = File.join(File.dirname(__FILE__), "templates", "#{type}-plugin")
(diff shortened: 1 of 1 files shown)


suyograo:
Lets wait for codec support so we can fix this right once for all.

purbon:
both fix are independent, you still need to validate the right values for this field.

suyograo:
Yeah, understood, but why not add codec support first and then fix https://github.com/elastic/logstash/pull/5419/files#diff-1d7289384dcead3ba670b14383780487R13
to include codec?

purbon:
check #5421, it might take long to actually decide if the codec example is good enough, so this is why I break up the fix, one for fixing each problem.

lib/pluginmanager/generate.rb
```diff
@@ -35,6 +32,10 @@ def execute
 
   private
 
+  def validate_params
+    raise(ArgumentError, "should be one of: input, output or filter") unless TYPES.include?(type)
```
talevy: will be `input, output, filter, or codec` once the other PR is in. yeah?

untergeek: Yep.  See #5421

untergeek:
LGTM™

lib/pluginmanager/generate.rb
purbon: agreed, just tried to split the PR as I was not expecting the codec to go in so fast 👍

PR #5421, opened 2016-06-01
Add codecs to the plugin generator

purbon:
Initially as the codec example repo was empty, codecs were not added to the plugin generator, however It makes a lot of sense to have them here, so here they are.

This PR adds what is into https://github.com/logstash-plugins/logstash-codec-example/pull/1 as initial PR for the codecs example, I'm sure we can iterate over the example meaning and composition, but i focused to provide the same content.

@untergeek as you made the example codec, your thoughts to what do here are very welcome.

This helps fixing #5412 together with #5419.

### DIFF
--- a/lib/pluginmanager/generate.rb
+++ b/lib/pluginmanager/generate.rb
@@ -8,9 +8,9 @@ require "pathname"
 
 class LogStash::PluginManager::Generate < LogStash::PluginManager::Command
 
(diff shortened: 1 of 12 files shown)


untergeek:
Brilliant!  LGTM™

purbon:
awesome, thanks @untergeek, can you also take a look at #5419, going to merge boths as they are related.

PR #5439, opened 2016-06-02
Add URI config validator/type

### LINKED ISSUE #2546: URL / URI parser would be very handy
Would be awesome if URL could get parsed, same style as user agent filter.
Fields to extract:
- protocol
- host
- port
- uri
- query string
- list of key / values for each parameters

Same applies if just URI.
Since many logs include URLs, it would make visualizations on those very easy. Getting aggregations on host, uri, params would be very handy.

andrewvc:
Often times plugins (like the Elasticsearch output) can naturally use URIs for their configuration.
Unfortunately using the :string type here means that the password portion of the URI can easily be leaked.

This wraps the URI class in a new LogStash::Util::SafeURI class that proxies all regular URI methods but masks
the password when `#to_s` and `#inspect` are invoked.

### DIFF
--- a/logstash-core/lib/logstash/config/mixin.rb
+++ b/logstash-core/lib/logstash/config/mixin.rb
@@ -4,6 +4,7 @@ require "logstash/config/registry"
 require "logstash/plugins/registry"
 require "logstash/logging"
 require "logstash/util/password"
+require "logstash/util/safe_uri"
 require "logstash/version"
 require "logstash/environment"
 require "logstash/util/plugin_version"
@@ -513,6 +514,12 @@ module LogStash::Config::Mixin
             end
 
             result = value.first.is_a?(::LogStash::Util::Password) ? value.first : ::LogStash::Util::Password.new(value.first)
+          when :uri
+            if value.size > 1
+              return false, "Expected uri (one value), got #{value.size} values?"
+            end
+            
+            result = value.first.is_a?(::LogStash::Util::SafeURI) ? value.first : ::LogStash::Util::SafeURI.new(value.first)
           when :path
             if value.size > 1 # Only 1 value wanted
               return false, "Expected path (one value), got #{value.size} values?"
@@ -550,6 +557,10 @@ module LogStash::Config::Mixin
       params.each do |key, value|
         if @config[key][:validate] == :password && !value.is_a?(::LogStash::Util::Password)
           params[key] = ::LogStash::Util::Password.new(value)
+        end
+
+        if @config[key][:validate] == :uri && !value.is_a?(::LogStash::Util::SafeURI)
+          params[key] = ::LogStash::Util::SafeURI.new(value)
         end
       end
     end
--- a/logstash-core/spec/logstash/config/mixin_spec.rb
+++ b/logstash-core/spec/logstash/config/mixin_spec.rb
@@ -102,6 +102,76 @@ describe LogStash::Config::Mixin do
     end
   end
 
+  context "when validating :uri" do
+    let(:klass) do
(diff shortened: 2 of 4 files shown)


logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -513,6 +514,12 @@ def validate_value(value, validator)
             end
 
             result = value.first.is_a?(::LogStash::Util::Password) ? value.first : ::LogStash::Util::Password.new(value.first)
+          when :uri
+            if value.size > 1
+              return false, "Expected uri (one value), got #{value.size} values?"
+            end
+            
+            result = value.first.is_a?(::LogStash::Util::SafeURI) ? value.first : ::LogStash::Util::SafeURI.new(value.first)
```
talevy: > value.first.is_a?(::LogStash::Util::SafeURI)

When would this be true if `validate_value` is called on the configuration?

Currently, the only time I see this would be true if someone was to set a value as the default value, right?

talevy:
Left small comment, OTT LGTM

logstash-core/spec/logstash/config/mixin_spec.rb
```diff
@@ -102,6 +102,58 @@
...
+        config :uri, :validate => :uri
+      end
+    end
+    
+    let(:scheme) { "myscheme" }
+    let(:user) { "myuser" }
+    let(:password) { "fancypants" }
+    let(:hostname) { "myhostname" }
+    let(:path) { "/my/path" }
+    let(:uri_str) { "#{scheme}://#{user}:#{password}@#{hostname}#{path}" }
```
suyograo: Can we add another test -- URI with no username/password to validate if the behavior is the same?

suyograo:
+1 for this, but this does not fix https://github.com/elastic/logstash/issues/2546 which is asking for a separate filter to extract fields from URI.. Just sayin' :)

logstash-core/lib/logstash/config/mixin.rb
andrewvc: Weird stuff happens during params parsing and values can be reparsed. I lifted this code from the Password validation which does the same thing. See https://github.com/andrewvc/logstash/blob/5956d8ecd089ef348c9b937ee8a06bcb48eb3221/logstash-core/lib/logstash/config/mixin.rb#L511

logstash-core/spec/logstash/config/mixin_spec.rb
andrewvc: Great call, in fact I just found a bug doing that!

andrewvc:
@suyograo @talevy all done! I would like to merge this into v2.4.0 as well as v5.0.0 if possible so that I can have the ES output use it in 2.4.x+

wiibaa:
@andrewvc is it necessary with this new type to add a section under **value types** in https://github.com/elastic/logstash/blob/master/docs/static/configuration.asciidoc ?

suyograo:
LGTM +1 on merging to 2.x branch

andrewvc:
@wiibaa you are absolutely right. I'll update the docs

suyograo:
@andrewvc docs look good.

logstash-core/lib/logstash/config/mixin.rb
talevy: ah yes. when plugins are cloned if I remember correctly. cool

andrewvc:
@suyograo and @talevy I realize now that the PR as I submitted it before only handled scalar values, it must handle arrays as well for it to be useful for the ES output. Mind taking another pass at this latest commit?

talevy:
👀  LGTM

PR #5442, opened 2016-06-03
Simplify the tests locations

purbon:
Hi,
  this PR introduced to small, but powerful improvements:
- It give back the `spec` dir as top main level test directory, this makes the  whole test location more compliant with what rspec expects and simplify unnecessary directories.
- It makes usages of shared examples only for test that will be shared, in this case the CLI as this is intended to be used in several stories.

should be merged in master and 5.0


ph:
I will close this PR it will be replaced by https://github.com/elastic/logstash/pull/6432

PR #5453, opened 2016-06-06
List plugin options

andrewvc:
Add new `:list `property to configuration parameters.

If set to try this will allow the user to specify one or more values.
This generally replaces the :array type, which had fewer type checks.

The array type is still needed for lists of hashes and mixed types.

### DIFF
--- a/docs/static/configuration.asciidoc
+++ b/docs/static/configuration.asciidoc
@@ -83,25 +83,38 @@ The settings you can configure vary according to the plugin type. For informatio
 === Value Types
 
 A plugin can require that the value for a setting be a
-certain type, such as boolean or hash. The following value
+certain type, such as boolean, list, or hash. The following value
 types are supported.
 
 [[array]]
-[float]
 ==== Array
 
-An array can be a single string value or multiple values. If you specify the same
-setting multiple times, it appends to the array.
+This type is now mostly deprecated in favor of using a standard type like `string` with the plugin defining the `:list => true` property for better type checking. It is still needed to handle lists of hashes or mixed types where type checking is not desired.
+
+Example:
+
+[source,js]
+----------------------------------
+  users => [ {id => 1, name => bob}, {id => 2, name => jane} ]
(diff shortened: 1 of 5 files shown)


logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -328,58 +328,84 @@ def validate_check_invalid_parameter_names(params)
...
+      
+      return [is_valid, processed_value]
+    end
+
     def validate_check_parameter_values(params)
       # Filter out parametrs that match regexp keys.
       # These are defined in plugins like this:
       #   config /foo.*/ => ... 
-      is_valid = true
+      all_params_valid = true
```
andrewvc: Renamed this and `success` because it was confusing to have two notions of validity with similar names in the same function.

docs/static/configuration.asciidoc
```diff
@@ -83,25 +83,38 @@ The settings you can configure vary according to the plugin type. For informatio
...
 
 [source,js]
 ----------------------------------
   path => [ "/var/log/messages", "/var/log/*.log" ]
-  path => "/data/mysql/mysql.log"
+  uris => [ "http://elastic.co", "http://example.net" ]
 ----------------------------------
 
-This example configures `path` to be an array that contains an element for each of the three strings.
+This example configures `path`, which is a `string` to be a list that contains an element for each of the three strings. It also will configure the `uris` parameter to be a list of URIs, failing if any of the URIs provided are not valid.
```
jordansissel: Not part of the change, but "three strings" isn't right, there are only two strings in both examples.

jordansissel:
Ignoring [jenkins failure](http://build-eu-00.elastic.co/job/Logstash_PR/3055/console) because it's metrics-related, not related to this PR.

jordansissel:
LGTM

suyograo:
I really don't know why Jenkins is building PRs now. I removed it a while back.

PR #5454, opened 2016-06-07
logstash-kafka plugin  adapt to kafka  0.10.0.0

alwayscode:
kafka  version  0.10.0.0  has changed the producer,  add timestamp  to message, configs, and Producer/ConsumerRecord,  so  the plugin should be written with the new producer api.


suyograo:
@alwayscode thanks for your contributions, but this is being taken care of here: https://github.com/logstash-plugins/logstash-input-kafka/pull/94

LS plugins live in a different organization and have their own repos: https://github.com/logstash-plugins

PR #5457, opened 2016-06-07
Add get and set to 2.x branch but leaving [] and []=

### LINKED ISSUE #5449: Add set and get API to Ruby Event implementation
Java Event APIs (setters/getters) were introduced in 5.0 (as core-plugin-api 2.0). All plugins were updated to use these calls instead of directly using Ruby's hash syntax. While this works well for 5.0, it would be great if we can introduce the setter and getter API in 2.4 as well, for the Ruby Event implementation. Basically `set` and `get` in LS 2.4 would alias to existing `[]` and `[]=` respectively. 

This change would allow plugins developed with 2.0 API to be backward compatible. On top of it, development process becomes easier as well; we don't have to backport every change to v1 _and_ v2 branches, _and_ publish from both!

guyboertje:
Fix #5449

### DIFF
--- a/logstash-core-event/spec/logstash/event_old_api_spec.rb
+++ b/logstash-core-event/spec/logstash/event_old_api_spec.rb
@@ -1,0 +1,662 @@
+# encoding: utf-8
+require "spec_helper"
+require "logstash/util/decorators"
+require "json"
+
+describe LogStash::Event do
+
+  shared_examples "all event tests" do
+    context "[]=" do
+      it "should raise an exception if you attempt to set @timestamp to a value type other than a Time object" do
+        expect{subject["@timestamp"] = "crash!"}.to raise_error(TypeError)
+      end
+
+      it "should assign simple fields" do
+        expect(subject["foo"]).to be_nil
+        expect(subject["foo"] = "bar").to eq("bar")
+        expect(subject["foo"]).to eq("bar")
+      end
+
+      it "should overwrite simple fields" do
+        expect(subject["foo"]).to be_nil
+        expect(subject["foo"] = "bar").to eq("bar")
+        expect(subject["foo"]).to eq("bar")
+
+        expect(subject["foo"] = "baz").to eq("baz")
+        expect(subject["foo"]).to eq("baz")
+      end
+
+      it "should assign deep fields" do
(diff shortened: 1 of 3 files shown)


ph:
LGTM, thanks for making this happen.

guyboertje:
@suyograo - please check that this is OK too.

logstash-core-event/spec/logstash/event_old_api_spec.rb
```diff
@@ -0,0 +1,662 @@
...
+describe LogStash::Event do
+
+  shared_examples "all event tests" do
+    context "[]=" do
+      it "should raise an exception if you attempt to set @timestamp to a value type other than a Time object" do
+        expect{subject["@timestamp"] = "crash!"}.to raise_error(TypeError)
+      end
+
+      it "should assign simple fields" do
+        expect(subject["foo"]).to be_nil
```
suyograo: why can't we test for `subject.get()` and `subject.set(x, y)`. There is already tons of tests for `subject[a]` and `[]=` correct?

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -8,68 +8,68 @@
...
+        expect(subject.get("[foo]")).to eq([nil])
       end
 
       it "should set nil in second element within empty array" do
-        subject["[foo]"] = []
+        subject.set("[foo]", [])
 
-        expect(subject["[foo][1]"] = nil).to eq(nil)
-        expect(subject["[foo]"]).to eq([nil, nil])
+        expect(subject.set("[foo][1]", nil)).to eq(nil)
```
suyograo: These things should not be changed right? both `[]`, `[]=` and `get` should work!

guyboertje:
@suyograo - I copied the `[] []=` based tests to a new file `event_old_api_spec` so if we remove support for `[] []=` we just need to remove the `event_old_api_spec` file.

guyboertje:
@suyograo - I am more concerned that I may have missed some other back ported changes in this PR, i.e. stuff in 2.3 that is not in 2.x. I did check but a second pair of eyes is always good.

suyograo:
LGTM

Issue #5440, opened 2016-06-03
Logstash 5.0.0~alpha3 - Multiple filter sections causing unexpected conditional behaviour


ceeeekay:
With multiple filter sections in a config (e.g., when using separate filter files) if a message matches multiple conditions, the action from the last condition will be replayed the number of times the message matches.

This can also be reproduced in a single file with multiple filter sections:
```
# cat test1.conf 
input {
  stdin {
    codec => json {}
  }
}

filter {
  if [foo] == "111" {
    mutate {
      add_tag => "1"
    }
  }
}

filter {
  if [bar] == "222" {
    mutate {
      add_tag => "2"
    }
  }
}

output {
  stdout {
    codec => rubydebug {}
  }
}
```
```
# /usr/share/logstash/bin/logstash --path.settings /etc/logstash/ --path.config test1.conf 
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
Sending logstash logs to /var/log/logstash/logstash.log.
Pipeline main started
{"foo": "111","bar": "222"}
{
           "foo" => "111",
           "bar" => "222",
      "@version" => "1",
    "@timestamp" => 2016-06-03T03:33:00.719Z,
          "host" => "civet",
          "tags" => [
        [0] "2",
        [1] "2"
    ]
}

```
Note the tag array "2","2" rather than the expected "1","2"

However if the same filters are contained within a single block, the behaviour is as expected:
```
# cat test2.conf 
input {
  stdin {
    codec => json {}
  }
}

filter {
  if [foo] == "111" {
    mutate {
      add_tag => "1"
    }
  }

  if [bar] == "222" {
    mutate {
      add_tag => "2"
    }
  }
}

output {
  stdout {
    codec => rubydebug {}
  }
}
```
```
# /usr/share/logstash/bin/logstash --path.settings /etc/logstash/ --path.config test2.conf 
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
Sending logstash logs to /var/log/logstash/logstash.log.
Pipeline main started
{"foo": "111","bar": "222"}
{
[...]

PR #5459, opened 2016-06-07
ensure plugins are correctly compiled outside of same plugin section

jsvd:
Because the config_ast uses an ivar local to the PluginSection class instances, a configuration such as:

```
input { generator {} }
filter { sleep { time => 1 } }
filter { sleep { time => 2 } }
output { }
```

the second sleep filter will override the first instead of being evaluated after it.
This PR adds a global counter that is reset on each pipeline compilation, avoiding this issue.

fixes https://github.com/elastic/logstash/issues/5440

### DIFF
--- a/logstash-core/lib/logstash/config/config_ast.rb
+++ b/logstash-core/lib/logstash/config/config_ast.rb
@@ -76,6 +76,14 @@ module LogStash; module Config; module AST
     @defered_conditionals_index = val
   end
 
+  def self.plugin_instance_index
+    @plugin_instance_index
+  end
+
+  def self.plugin_instance_index=(val)
+    @plugin_instance_index = val
+  end
+
   class Node < Treetop::Runtime::SyntaxNode
     def text_value_for_comments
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/config/config_ast.rb
```diff
@@ -76,6 +76,14 @@ def self.defered_conditionals_index=(val)
     @defered_conditionals_index = val
   end
 
+  def self.plugin_instance_count
+    @plugin_instance_count
```
colinsurprenant: since this is pretty much the same logic as with `@defered_conditionals_index` I suggest we use the same naming? `@plugin_instance_index` ?

colinsurprenant:
Great, with specs etc! 👍   
As commented in code, I only suggest an identifier rename.
LGTM

jsvd:
Updated, thanks

colinsurprenant:
LGTM² :)

Issue #5416, opened 2016-06-01
Building release packages by not changing files


suyograo:
We need to change the way we build snapshot packages and release packages. Basically, we should be able to promote a release candidate (RC) package to a final release package without changing any files. We can achieve this by:

* Having one source of truth + script to set version #5414 
* Build snapshot packages by default when we run a build. There is no need to specify snapshots in the version info. For example, we would never have `5.0.0-alpha4.snapshot1` in `version.yml`. It would only be `5.0.0-alpha4`.  This is kinda already done in master..
* `rake artifact:tar` should always build a snapshot. by passing `--release` flag, it should promote the sha to be a release instead of the usual snapshot. No changes should be made to any files.

The one gotcha is publishing logstash-core gems to RubyGems.org so plugins can use it for testing.. RubyGems.org has a constraint on unique versions for publish (i.e. we can't replace existing Gem with same versions). This means we can't actually replace existing snapshots when code changes are made and _still_ reference the same version in the gemspec.

suyograo:
@jsvd @jordansissel and I had a zoom conversation yesterday. Rough notes:

### Enabling Multiple Snapshots
Gemfiles in release branches will not use published `logstash-core*` gems from RubyGem. It will be built locally and referenced, like we do in master branch now:

```
gem "logstash-core", "5.0.0-alpha4", :path => "./logstash-core"
gem "logstash-core-event-java", "5.0.0-alpha4", :path => "./logstash-core-event-java"
gem "logstash-core-plugin-api", "2.20.0", :path => "./logstash-core-plugin-api"
```

The version here will come from `pkg/version.yml` which is described in #5414. This way, we can publish as many snapshots as possible (without having to increment snapshot versions) for a particular version. Once we are ready to promote a snapshot to an actual release, we only have to use the `--release` switch and no files (version.yml, gemspecs) will be touched when creating the final package.

### Plugin Tests
Plugins tests will be affected by this. Today, plugins use the latest released `logstash-core*` in their tests. Our suggestion here was to test plugins against Logstash repo directly, eliminating the need for published snapshots. For example, travis.yml can be changed to be test using a matrix: 
     a) Test against head of master
     b) Test against head of latest release branch (for example 5.0). 

For b) travis.yml has to be changed to use:
```
before_install:
- git clone -b v5.0 https://github.com/elastic/logstash
```

and in gemspec:

```
source 'https://rubygems.org'
gemspec
gem "logstash-core", :path => "./logstash/logstash-core"
[...]

Issue #5414, opened 2016-06-01
Extract all version info into one single file


suyograo:
Today there are several places to update versions for different gems. We should extract all of them out and move them to a separate file under `pkg/version.yml`

The structure of version.yml can be:

```
logstash-core: 5.0.0-alpha3
# random comment here
logstash-core-plugin-api: 2.1.6
logstash-core-event-java: 5.0.0-alpha3
# any other dependency here:
```

Once this is done, its easy to set version by the unified build script.

jsvd:
my first approach was to make all version constants to pull from the YAML file.
The problem is that our gems, when packaged, won't contain the yaml file unless we replicate it to all core gems.
The alternative is having indeed a yaml file that is used as SSOT, but then a rake task is used to propagate changes in the YAML to the gemspecs.
Thoughts?

suyograo:
> The alternative is having indeed a yaml file that is used as SSOT, but then a rake task is used to propagate changes in the YAML to the gemspecs. 

Works. As long as there is just one place as source to truth

suyograo:
> but then a rake task is used to propagate changes in the YAML to the gemspecs.

This task should be called only by `rake set-version` though, right?

Is this the workflow you're thinking:

* `rake version:set --version 5.0.0-alpha4` will set this version in `pkg/version.yml` and then call another internal rake task to propagate this to all the Gems?
* `rake version:get` outputs the current version.
* `rake artifact:tar` will build a snapshot artifact from the current sha and version: 5.0.0-alpha4-snapshot.
* `rake artifact:tar --release` will build a release artifact from the current sha and version: 5.0.0-alpha4.

jsvd:
I'm thinking of:

* `rake version:sync`: check if it's necessary to propagate the changes and does it if so. This can be called at the start of ANY rake task automatically, like we do now for vendoring jruby.
* `rake version:set[5.0.0-alpha4]` which changes the yaml file and then calls `rake version:sync`
* `rake version:show` that prints the version of the logstash package
   - `rake version:show[all|logstash|logstash-core|logstash-core-event|..]` show version of subcomponent or all of them
* `rake artifact:tar` create a snapshot artifact from current sha with CURRENT.VERSION-snapshot and writes to a file latest.build `commit_sha1:file_sha1`
[...]

PR #5460, opened 2016-06-07
unified release PR

jsvd:
Purpose:

* manage releases through a minimum number of rake tasks
* simplify building of snapshot builds
* create staged artifacts, candidates for releases, that required no changes to become releases
* this means the snapshot release process will not involve publishing gems, therefore:
* the gem artifacts should only be published to rubygems as a final artifact, at the time of GA

Changes:

* release artifacts no longer depend on gems of core components
* all core components are used locally AS-IS so that no code changes from snapshot to ga
* `versions.yml` describes the versions of all logstash parts and package
* `rake version:set/set_plugin_api/set_package` manage the yaml file and push the changes to the gemspecs/version.rb files
* `rake artifact:prepare` updates the lock file. useful for a release branch like 5.0
* `rake artifact:all` generates SNAPSHOT artifacts: tar.gz, zip, rpm, deb
* `RELEASE=1 rake artifact:all` creates release candidate artifacts + 4 gems: logstash-core, logstash-core-event, logstash-core-event-java and logstash-core-plugin-api

implements #5416 and #5414

for a rough simulation of the release process with a new, 6.0 branch, see https://gist.githubusercontent.com/jsvd/a83a09f1d5029405526f6db092b6f425/raw/32585d0bc9bfb879e6f522e1f3a5ff93bf7e1b04/gistfile1.txt


colinsurprenant:
about
> the gem artifacts should only be published to rubygems as a final artifact, at the time of GA

Does that conflict with, AFAIU, the goal of the Unified Build of publishing release candidates and if they are chosen for release, then to be able to publish as GA as-is, without going through any (re)packaging process?

jsvd:
the key point of unified release is to generate a release from the same sha1 commit used for the snapshot. not repackaging is desirable but, afaik, not mandatory.
The simple fact that a SNAPSHOT package, when doing `bin/logstash -V` will return 5.0.0-SNAPSHOT while the released one shows just 5.0.0 means that repackage will need to happen, but this can be made so no changes happen in git

jordansissel:
> the gem artifacts should only be published to rubygems as a final artifact, at the time of GA

I think this idea (at release-time, to publish anything other than tar/rpm/etc artifacts) was frowned upon by the others working in unified release (right, @rjernst?)

jsvd:
The reason for publshing gems at all is to run tests on the plugins against the latest version of logstash core/event.
For logstash itself we dont need these gems published at all.

jordansissel:
@jsvd right, which means we can probably publish core gems outside the normal scheduling of the main artifact releases, right?

jsvd:
Yes we can..
But I'd like to avoid that extra manual step if possible. At the end of the build, the folder will contain all the gem, tar.gz, zip, deb and rpm. I'd like to publish them all. Thoughts, @rjernst?

jordansissel:
> The reason for publshing gems at all is to run tests on the plugins against the latest version of logstash core/event.

We can do this by directing plugin tests to use a git target for the source instead of rubygems.org (we do this today).We'll have to do that anyway prior to declaring a GA-ready snapshot (otherwise we'll publish a GA without knowing that all plugins are happy with it, right?)

jsvd:
Not necessarily, running rake test:plugins in Logstash will test it against latest release of each plugin.

colinsurprenant:
> the key point of unified release is to generate a release from the same sha1 commit used for the snapshot. 

Is it? Last time we discussed that I understood that the RC packages would be turned into GA releases as-is form a whole package perspective, no repackaging, no version update, nada. If this is in fact still a requirement then RC cycles will either have to use published core gems OR we switch to use embedded core gems in GA releases.

colinsurprenant:
> Not necessarily, running rake test:plugins in Logstash will test it against latest release of each plugin. 

I think @jordansissel was referring to Travis testing the plugins.

jsvd:
Ok, I was addressing the actual concern of "we'll publish a GA without knowing that all plugins are happy with it, right"

As for not repackaging, do you suggest that there is no mention of SNAPSHOT other than tgz filename?
Because this way when you unpack it will unpack to logstash-5.0.0/
I can implement this easily, just seems wrong to do no repackaging whatsoever

jsvd:
Also, we have agreed before to add a new Travis test config setting to each plugin that will checkout a Logstash git clone and test against it.

colinsurprenant:
> As for not repackaging, do you suggest that there is no mention of SNAPSHOT other than tgz filename?

I believe this was the requirement yes, that between an accepted RC package and GA, no bits are changed other than the file name(s). This mean that all RCs should use GA install paths and version numbers. In other words, no repackaging script is run after an accepted RC to produce a GA.

suyograo:
Folks, I think there is some confusion here. There are 2 separate concepts in unified build:
* **Snapshots:** These are daily build snapshots of a release branch. 
   * These will not be turned into, or promoted to a final release. 
   * They are for internal/external testing. 
   * By default, any artifact creation task will always create a snapshot release package. So `rake artifact:tar` on 5.0 branch with version `5.0.0-beta1` will generate `logstash-5.0.0-beta1-SNAPSHOT`. When you unzip it, it should go into `logstash-5.0.0-beta1-SNAPSHOT/*` dir and show version as `5.0.0-beta1-SNAPSHOT`.
   * The fact that this is a snapshot is just a flag, and stored as metadata. The version in `version.yml` will never contain a snapshot postfix. It will always be 5.0.0, 2.4.0 etc. There is a separate field called `metadata: true|false` in version.yml.
  * No Gems will be published, everything will be built locally and used in gemspec.
  * Plugin travis tests will run off the tip of release branches, so 5.0, 2.3, 2.x etc.
* **Release Candidate:** These are potential candidates which are staged for a release.
  * Release candidates are built by passing `-release` to the rake task.
  * Version for release candidate will display: `5.0.0-alpha4`, `2.4.0` etc.
  * Release candidates go through testing phase. If there are bugs found, you fix the issue, get a new sha1 and re-spin another candidate. 
  * If there are no bugs, a release candidate will be renamed/promoted to GA release. It will be moved from staging repo to publicly available repo for all our users to download.
  * The promotion from candidate to release will not touch any files, and there is no metadata to be changed.
  * Once release candidate is promoted, the release is baked. We cannot change anything. At this point, we also publish locally built gems to RubyGems.org so everyone (including plugins) can use this.

rjernst:
@suyograo Has it correct, except for one caveat:

> If there are no bugs, a release candidate will be renamed/promoted to GA release

There is no renaming. Filenames are copied as-is to the release location in s3.

I do have one question about the gems:

> Once release candidate is promoted, the release is baked. We cannot change anything. At this point, we also publish locally built gems to RubyGems.org so everyone (including plugins) can use this.

If you are going to publish to ruby gems every release, then I don't understand why you would want to do this as a separate process.

colinsurprenant:
@rjernst right, @suyograo could you clarify your thoughts on

> Once release candidate is promoted, the release is baked. We cannot change anything. At this point, we also publish locally built gems to RubyGems.org so everyone (including plugins) can use this.

If a release candidate is fully backed and simply copied as-is to a release directory as part of the promotion, this means that all gems will need to be published as part of the build process, prior to this promotion so that the same exact bits will be used. 

Am I missing something?

suyograo:
> gems will need to be published as part of the build process, prior to this promotion so that the same exact bits will be used.

For Logstash, we'll need an extra step as part of the promotion from RC to GA to publish gems to Ruby Gems. For this, the unified build process could get the sha1 which becomes the official GA, build and publish Gems off of this. 

@rjernst does this answer your question as well? That gem publishing should be built-in to the unified release, just as ES would publish final jars to maven..
 
> If you are going to publish to ruby gems every release, then I don't understand why you would want to do this as a separate process?

suyograo:
We can provide a rake task for publishing gems given a sha1. The unified build could call this before promoting the RC to GA.

`rake release:publish-gems sha1`

colinsurprenant:
@suyograo but if the RC uses unpublished gems, publishing gems for the promotion means we have to also update the RC Gemfile to point to the newly published gems which breaks the no-bits-changed as part of the promotion, no?

suyograo:
The point of publishing gems is for external consumption -- plugins. LS package should not change anything since its using the same bits.. This is different than how it works today. The publishing of gems is merely another step to close out the release, unrelated to the release package itself.

colinsurprenant:
@suyograo ok, so are we saying that we will change our current strategy to reference core gems on Rubygems and just embed and reference the core gems locally in all packages?

suyograo:
> ok, so are we saying that we will change our current strategy to reference core gems on Rubygems and just embed and reference the core gems locally in all packages?

Yes

suyograo:
@jsvd can you modify the rake task for `--release` to create an additional step: 
Build gems for `logstash-core`, `logstash-core-plugins-api` and `logstash-core-event-java` and move them to `LS_HOME/build/gems` directory?
Unified build process will then copy these and only publish them when we move from RC to GA.

jsvd:
@suyograo yeah, right now the build task already creates the gems, I'll modify it to only create them if the release flag is passed and move them to the gems directory 👍

jsvd:
@suyograo I have one correction to make about:

> By default, any artifact creation task will always create a snapshot package. So rake artifact:tar on 5.0 branch with version 5.0.0-beta1 will generate logstash-5.0.0-beta1-SNAPSHOT. When you unzip it, it should go into logstash-5.0.0-beta1-SNAPSHOT/* dir and show version as 5.0.0-beta1-SNAPSHOT.

No information should be displayed on the fact that the build is a snapshot other than directory name, right @rjernst ? 

This happens already for beats (no mention of snapshot):

```
...beat-5.0.0-alpha3-SNAPSHOT-darwin-x64 % ./filebeat -version
filebeat version 5.0.0-alpha3 (amd64), libbeat 5.0.0-alpha3
```

and on ES side:

```
% bin/elasticsearch --version
Version: 5.0.0-alpha3, Build: 58a103a/2016-06-01T22:57:12.730Z, JVM: 1.8.0_20
```

jsvd:
Updated code and PR description to reflect requirements

suyograo:
@jsvd can we persist the build hash so we can display it in the API like ES does? Will be nice for debugging. Also ES has a `build_snapshot` field to indicate a snapshot release. But, yeah, its not in the version itself.

```
{
  "name" : "Ironclad",
  "cluster_name" : "elasticsearch",
  "version" : {
    "number" : "5.0.0-alpha2",
    "build_hash" : "e3126df",
    "build_date" : "2016-04-26T12:08:58.960Z",
    "build_snapshot" : false,
    "lucene_version" : "6.0.0"
  },
  "tagline" : "You Know, for Search"
}
```

rjernst:
Do gems have a way to store "metadata"? We store the snapshot flag in jar META-INF

jsvd:
all our equivalents to META-INF: metadata about the code are also part of the code. 
The only file which is produced by the build and not commited is the Gemfile.lock which contains the result of computing the dependency resolution of gem requirements.

So we have two options here, either:
1) have our logstash package code assume a file containing metadata will exist at a certain relative path. This phantom file is generated and injected at build time and does not exist in the git code
2) during build process we change a commited metadata file with the information about `sha1`, `snapshot_build`, etc. These changes are never commited.

The cleaner way seems to be 1), we only need to ensure that logstash, when used from outside of the build (e.g. from a git clone) it works around the absense of this file.
Thoughts about this?

rjernst:
> when used from outside of the build (e.g. from a git clone) it works around the absense of this file

Not sure exactly what you mean, but in ES we have special logic for when ES is not run from a jar (eg in unit tests).

suyograo:
```
suyog@machine:~/ws/elastic/logstash (pr/5460)$ rake version:set 5.0.0-alpha4
Updating "logstash" from "5.0.0" to ""
rake aborted!
can't convert nil into String
org/jruby/RubyString.java:3093:in `gsub'
org/jruby/RubyString.java:3069:in `gsub'
/Users/suyog/ws/elastic/logstash/rakelib/version.rake:63:in `(root)'
org/jruby/RubyHash.java:1610:in `select'
/Users/suyog/ws/elastic/logstash/rakelib/version.rake:59:in `(root)'
org/jruby/RubyProc.java:281:in `call'
org/jruby/RubyArray.java:1613:in `each'
/Users/suyog/ws/elastic/logstash/rakelib/version.rake:109:in `(root)'
org/jruby/RubyProc.java:281:in `call'
org/jruby/RubyArray.java:1613:in `each'
org/jruby/RubyArray.java:1613:in `each'
org/jruby/RubyKernel.java:1059:in `load'
Tasks: TOP => version:sync
```

@jsvd it barfs on set version..

jsvd:
Correct rake arg syntax is `rake version:set[5.0.0-alpha4]`

That said, I'll add arg validation..

jsvd:
```
% rake version:set_package 5.0.0
Invalid version argument: "". Aborting...
% rake version:set_package[5.0.0batdsa]
Invalid version argument: "5.0.0batdsa". Aborting...
% rake version:set_package[5.0.0-alpha4]
Updating "logstash" from "5.0.0" to "5.0.0-alpha4
%
```

suyograo:
Tested both snapshots and RC workflow. The deb and rpm build takes a while, but it is what it is for now.

Snapshot
----------------------------------------
```
rake bootstrap
rake version:set[5.0.0-alpha4]
rake artifact:all
```
Release
-----------------------------------------
```
RELEASE=1 rake artifact:all
```
This puts the final gems in `LS_HOME/build/gems` and artifacts in `LS_HOME/build`


LGTM.

suyograo:
Unrelated to this PR, but we have to check why the artifacts have bloated back to ~100MB. We added GeoLite2 db, but that shouldn't add 20MB

jsvd:
it might, 
```
% pwd
logstash-5.0.0/vendor/bundle/jruby/1.9/gems/logstash-filter-geoip-4.0.0-java
% du -sh vendor/GeoLite2-City.mmdb
 62M	vendor/GeoLite2-City.mmdb
%
```
this is a binary file, so won't compress well. Previously was, in 2.3.2:

```
% du -sh vendor/bundle/jruby/1.9/gems/logstash-filter-geoip-2.0.7/vendor/
 21M	vendor/bundle/jruby/1.9/gems/logstash-filter-geoip-2.0.7/vendor/
```

so a bump of 20 mb in file size actually makes a lot of sense.

logstash-core-plugin-api/logstash-core-plugin-api.gemspec
```diff
@@ -11,13 +11,13 @@ Gem::Specification.new do |gem|
   gem.homepage      = "http://www.elastic.co/guide/en/logstash/current/index.html"
   gem.license       = "Apache License (2.0)"
 
-  gem.files         = Dir.glob(["logstash-core-event.gemspec", "lib/**/*.rb", "spec/**/*.rb"])
+  gem.files         = Dir.glob(["logstash-core-plugin/api.gemspec", "lib/**/*.rb", "spec/**/*.rb"])
```
ph: I think this should be `logstash-core-plugin-api.gemspec` ?

rakelib/artifacts.rake
```diff
@@ -55,20 +65,76 @@ namespace "artifact" do
...
   end
 
+  task "generate_build_metadata" do
+    return if defined?(BUILD_METADATA_FILE)
+    BUILD_METADATA_FILE = Tempfile.new('build.rb')
+    build_sha = `git rev-parse HEAD`.chomp
+    metadata = [
+      "# encoding: utf-8",
+      "BUILD_INFO = { \"build_sha\" => \"#{build_sha}\", \"build_snapshot\" => #{SNAPSHOT_BUILD} }",
+    ]
```
ph: Small suggestion instead of dealing with manually generating the hash you can use `to_s` to get the ruby code.

```
IO.write("/tmp/test", { "build_sha" => "1234"}.to_s)

~/e/logstash git:6.0 ❯❯❯ cat /tmp/test                                                                                                                  ✱ ◼
{"build_sha"=>"1234"}%
```

ph: Do we want to add the `build_date` like ES does?

```
    "build_date" : "2016-04-26T12:08:58.960Z",
```

rakelib/artifacts.rake
```diff
@@ -90,15 +156,15 @@ namespace "artifact" do
     path = matches.first[gem_line_regex, 1]
 
     if path
-      Rake::Task["plugin:install-local-core-gem"].invoke("logstash-core", path)
+      Rake::Task["plugin:build-local-core-gem"].invoke("logstash-core", path)
     else
-      puts("[artifact:install-logstash-core] using logstash-core from Rubygems")
+      exit(1)
```
ph: Why are we explicitly returning a value greather than 0 ?
I don't see any error in the code path that would require us to return 1 ?

rakelib/artifacts.rake
```diff
@@ -110,15 +176,15 @@ namespace "artifact" do
     path = matches.first[gem_line_regex, 2]
 
     if path
-      Rake::Task["plugin:install-local-core-gem"].invoke(name, path)
+      Rake::Task["plugin:build-local-core-gem"].invoke(name, path)
     else
-      puts("[artifact:install-logstash-core] using #{name} from Rubygems")
+      exit(1)
```
ph: should it be 0?

rakelib/version.rake
```diff
@@ -0,0 +1,132 @@
...
+    Rake::Task["version:sync"].invoke; Rake::Task["version:sync"].reenable
+  end
+
+  task :validate, :version do |t, args|
+    unless Regexp.new('^\d+\.\d+\.\d+(?:-\w+\d+)?$').match(args[:version])
+      abort("Invalid version argument: \"#{args[:version]}\". Aborting...")
+    end
+  end
+
+end
```
ph: +1 Well written rake tasks!

rakelib/artifacts.rake
jsvd: it's just that at this point it should never happen, we should always have gem "logstash-core", :path => ...
if not there's something wrong

ph:
I love the new workflow will make things much easier, did a review of the code itself, really minor things: - The status code returned from some rake task
- The build date would be a nice thing?

rakelib/artifacts.rake
ph: make sense with your comment, but maybe add a debug log?

ph:
@jsvd Small question concerning the build, I should be able to run the following method in this exact order right?

```
rake bootstrap
rake version:set[5.0.0-alpha4]
rake artifact:all
RELEASE=1 rake artifact:all
```

jsvd:
correct. bootstrap isn't necessary, I think

jsvd:
added build_date info and addressed PR concerns

ph:
There is a small issue with the generate gems from the process.

```
│   ├── logstash-core-5.0.0.pre.alpha4-java.gem
│   ├── logstash-core-event-java-5.0.0.pre.alpha4-java.gem
``

We should not have a `.pre.` in them, its because we have a `-` in the version

ph:
We sync on zoom, and we believe that the .pre wont be an issue in the future releases.

ph:
The failling test is not related to this PR, I've created a specific issue (#5518)  for it.

I've tested the generated artefact for release and the snapshot and they both work, this code will make things much smoother.


LGTM :shipit: ❤️

PR #5463, opened 2016-06-08
Add an option to install x-packs in logstash

### LINKED ISSUE #4585: Plugin Manager: Extend functionality to installation plugin packs

purbon:
This PR basically supersede the old #4584 and has the motivation to add support for the new packs inside logstash.

> This PR adds the idea of packaged plugins to the plugin manager, basically a collection of plugins that are going to be installed from a given zip file. This file could life either in a local file system or in a remote installation.

pending work:

- [x] Move test to the QA/Acceptance framework.
- [x] What happen to the file if the download fails.
- [x] Http Error Code handling

Fixes #4585


purbon:
@ph re ` What happen to the file if the download fails.`

I've been thinking and I guess the right behaviour is:

1.- If the plugin requested is a valid xpack, and for example the download return a code >= 500 to raise an exception is a reasonable outcome. This is also in case of an exception happen during the download process.
2.- If the download file fails because of a 404 for example, even if the name provided is for a valid and expected xpack, we don't have a resonable way to provide clever error reporting. (1)

(1) this is because we must always try to follow up if the requested plugin is not xpack and an actuall gem.

Makes sense to you?

purbon:
I will remove for now the work in progress, but still work on getting more test for error cases here. I would appreciate your thoughts for now. thanks for your time @ph

ph:
I've rebased this PR there https://github.com/ph/logstash/tree/feature/pack if you want to test it out with the latest master.

ph:
I am OK with that code @purbon, I've created the feature/pack branch and will implements changes on top of your PR.

Good work!

cc @jsvd this PR does the majority of what we want, I think it just need a few changes and we need to test the offline scenario more closely.

I will close this issue and make a future PR when all the work is done.

PR #5472, opened 2016-06-09
Track wallclock time spent in filters and outputs. 

jordansissel:
This PR adds time tracking (wallclock) to all individual filter instances. The goal is to help identify what plugin configurations are consuming the most time.

This data is stored in the internal metrics system and made available via the `/_node/stats/pipelines` HTTP endpoint.

### DIFF
--- a/logstash-core/lib/logstash/api/commands/stats.rb
+++ b/logstash-core/lib/logstash/api/commands/stats.rb
@@ -34,7 +34,8 @@ module LogStash
         end
 
         def pipeline
-          service.get_shallow(:stats, :pipelines)
+          stats = service.get_shallow(:stats, :pipelines)
+          PluginsStats.report(stats)
         end
 
         def memory
@@ -53,7 +54,98 @@ module LogStash
               acc
             end
           }
-        end       
+        end
+
+        def hot_threads(options={})
+          HotThreadsReport.new(self, options)
+        end
+
+        class HotThreadsReport
+          HOT_THREADS_STACK_TRACES_SIZE_DEFAULT = 10.freeze
+          
+          def initialize(cmd, options)
+            @cmd = cmd
+            filter = { :stacktrace_size => options.fetch(:stacktrace_size, HOT_THREADS_STACK_TRACES_SIZE_DEFAULT) }
+            jr_dump = JRMonitor.threads.generate(filter)
+            @thread_dump = ::LogStash::Util::ThreadDump.new(options.merge(:dump => jr_dump))
+          end
+          
+          def to_s
+            hash = to_hash
+            report =  "#{I18n.t("logstash.web_api.hot_threads.title", :hostname => hash[:hostname], :time => hash[:time], :top_count => @thread_dump.top_count )} \n"
+            report << '=' * 80
+            report << "\n"
+            hash[:threads].each do |thread|
+              thread_report = ""
+              thread_report = "#{I18n.t("logstash.web_api.
+                                hot_threads.thread_title", :percent_of_cpu_time => thread[:percent_of_cpu_time], :thread_state => thread[:state], :thread_name => thread[:name])} \n"
+              thread_report = "#{thread[:percent_of_cpu_time]} % of of cpu usage by #{thread[:state]} thread named '#{thread[:name]}'\n"
+              thread_report << "#{thread[:path]}\n" if thread[:path]
+              thread[:traces].each do |trace|
+                thread_report << "\t#{trace}\n"
+              end
+              report << thread_report
+              report << '-' * 80
+              report << "\n"
+            end
+            report
+          end
+
+          def to_hash
+            hash = { :hostname => @cmd.hostname, :time => Time.now.iso8601, :busiest_threads => @thread_dump.top_count, :threads => [] }
+            @thread_dump.each do |thread_name, _hash|
+              thread_name, thread_path = _hash["thread.name"].split(": ")
+              thread = { :name => thread_name,
+                         :percent_of_cpu_time => cpu_time_as_percent(_hash),
+                         :state => _hash["thread.state"]
+                       }
+              thread[:path] = thread_path if thread_path
+              traces = []
+              _hash["thread.stacktrace"].each do |trace|
+                traces << trace
+              end
+              thread[:traces] = traces unless traces.empty?
+              hash[:threads] << thread
+            end
+            hash
+          end
+
+          def cpu_time_as_percent(hash)
+            (((cpu_time(hash) / @cmd.uptime * 1.0)*10000).to_i)/100.0
+          end
+
+          def cpu_time(hash)
+            hash["cpu.time"] / 1000000.0
+          end
+        end # class HotThreadsReport
+
+        module PluginsStats
+          module_function
+
+          def plugin_stats(stats, plugin_type)
+            # Turn the `plugins` stats hash into an array of [ {}, {}, ... ]
+            # This is to produce an array of data points, one point for each
+            # plugin instance.
+            return [] unless stats[:plugins].include?(plugin_type)
+            stats[:plugins][plugin_type].collect do |id, data|
+              { :id => id }.merge(data)
+            end
+          end
+
+          def report(stats)
+            # Only one pipeline right now.
+            stats = stats[:main]
+
+            {
+              :events => stats[:events],
+              :pipeline => {
+                :inputs => plugin_stats(stats, :inputs),
+                :filters => plugin_stats(stats, :filters),
+                :outputs => plugin_stats(stats, :outputs)
+              }
+            }
+          end
+        end # module PluginsStats
       end
     end
   end
--- a/logstash-core/lib/logstash/api/commands/stats.rb
+++ b/logstash-core/lib/logstash/api/commands/stats.rb
@@ -33,6 +33,10 @@ module LogStash
           )
         end
 
+        def pipeline
+          service.get_shallow(:stats, :pipelines)
+        end
+
         def memory
           memory = service.get_shallow(:jvm, :memory)
           {
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -39,7 +39,7 @@ module LogStash
 
       start = Time.now
       new_events = @filter.multi_filter(events)
-      @filter.metric.increment(:time_ms, ((Time.now - start)*1000).to_i)
+      @filter.metric.increment(:duration_in_millis, ((Time.now - start)*1000).to_i)
 
       # There is no garantee in the context of filter
       # that EVENTS_INT == EVENTS_OUT, see the aggregates and
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -37,7 +37,9 @@ module LogStash
     def multi_filter(events)
       @metric_events.increment(:in, events.size)
 
+      start = Time.now
       new_events = @filter.multi_filter(events)
+      @filter.metric.increment(:time_ms, ((Time.now - start)*1000).to_i)
 
       # There is no garantee in the context of filter
       # that EVENTS_INT == EVENTS_OUT, see the aggregates and
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -24,8 +24,8 @@ module LogStash
       namespaced_metric = metric.namespace(@filter.plugin_unique_name.to_sym)
       @filter.metric = metric
 
-      namespaced_metric.gauge(:name, config_name)
       @metric_events = namespaced_metric.namespace(:events)
+      namespaced_metric.gauge(:name, config_name)
 
       # Not all the filters will do bufferings
       define_flush_method if @filter.respond_to?(:flush)
@@ -38,7 +38,7 @@ module LogStash
     def multi_filter(events)
       @metric_events.increment(:in, events.size)
 
-      clock = @filter.metric.time(:duration_in_millis)
+      clock = @metric_events.time(:duration_in_millis)
       new_events = @filter.multi_filter(events)
       clock.stop
 
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -24,6 +24,7 @@ module LogStash
       namespaced_metric = metric.namespace(@filter.plugin_unique_name.to_sym)
       @filter.metric = metric
 
+      namespaced_metric.gauge(:name, config_name)
       @metric_events = namespaced_metric.namespace(:events)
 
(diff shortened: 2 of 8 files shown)


jordansissel:
Sample run:

```
% bin/logstash --http.host '0.0.0.0' -e 'input { generator { } } filter { sleep { time => "0.1" } grok { match => { "message" => "foo" } } } output { null { } }'
```

Ask for pipeline stats:

```
% curl localhost:9600/_node/stats/pipelines?pretty
{
  "pipelines" : {
    "main" : {
      "events" : {
        "in" : 249,
        "filtered" : 124,
        "out" : 124
      },
      "plugins" : {
        "outputs" : {
          "null_8b1226e2-e714-444c-9774-0135833457d3" : {
            "events" : {
              "in" : 124,
              "out" : 124
            }
          }
        },
        "filters" : {
          "sleep_713f61e2-6d7a-436c-9159-b3a953b18451" : {
            "events" : {
              "in" : 145,
              "time_ms" : 14747,
              "out" : 145
            }
          },
          "grok_0236a213-3b5b-4a76-ab75-0a3c75c61526" : {
            "events" : {
              "in" : 145,
              "time_ms" : 75,
              "out" : 144
            }
          }
        }
      }
    }
  }
}
```

jordansissel:
think the jenkins build failure is unrelated

suyograo:
@jordansissel this is awesome! Thoughts about moving plugin specific stats to `_plugin` as discussed here #? IMHO, `_pipeline/stats` should be overall stats for pipelines running in LS. If and when we add multi-pipelines in the future it will show all stats from running pipelines.


So these stats would be:

```
GET curl localhost:9600/_plugins`
```

Response

```
"plugins" : {
        "outputs" : {
          "null_8b1226e2-e714-444c-9774-0135833457d3" : {
            "events" : {
              "in" : 124,
              "out" : 124
            }
          }
        },
        "filters" : {
          "sleep_713f61e2-6d7a-436c-9159-b3a953b18451" : {
            "events" : {
              "in" : 145,
              "time_ms" : 14747,
              "out" : 145
            }
          },
          "grok_0236a213-3b5b-4a76-ab75-0a3c75c61526" : {
            "events" : {
              "in" : 145,
              "time_ms" : 75,
              "out" : 144
            }
          }
        }
      }
```

suyograo:
@jordansissel see https://github.com/elastic/logstash/issues/5475#issuecomment-225127603

suyograo:
Also, what if I want to see time spent in _all_ my grok filters..

Can we add

```
"plugins" : {
        "outputs" : {
            "grok_713f61e2-6d7a-436c-9159-b3a953b18451": {
                 "name": "grok"
                 "events" : {
                       "in" : 145,
                       "time_ms" : 75,
                       "out" : 144
                 }
            }
   } 

```

or using arrays...

```
"plugins" : {
        "outputs" : {
              "grok" : [ 
                   {
                      "id": "grok_713f61e2-6d7a-436c-9159-b3a953b18451"      
                      "events" : {
                           "in" : 145,
                           "time_ms" : 75,
                           "out" : 144
                      }
                  },
                  {
                 "id": "grok_33345e2-6d7a-436c-9159-b3a953b18451"   
                 "events" : {
                           "in" : 145,
                           "time_ms" : 75,
                           "out" : 144
                      }
                  }
             ]
         }
   } 
```

suyograo:
Also, when a config has multiple filters of the same kind, how will a user distinguish which particular filter took most time? IOW, how can a user know what `grok_713f61e2-6d7a-436c-9159-b3a953b18451` is in the configuration?

logstash-core/lib/logstash/filter_delegator.rb
```diff
@@ -37,7 +37,9 @@ def config_name
     def multi_filter(events)
       @metric_events.increment(:in, events.size)
 
+      start = Time.now
       new_events = @filter.multi_filter(events)
+      @filter.metric.increment(:time_ms, ((Time.now - start)*1000).to_i)
```
ph: I didn't think about how the pipeline args could return accurate results with every snapshot, thanks for showing me the obvious here :)

I suggest a simple change, the metric lib has already a method called `#time` that create a `MeanType` instead of a counter. I suggest we change the `Metric#Time` to use a counter and remove the `MeanType`.

`#Time` accepts a block or will return a `TimedExecution` when no block is provided, we can call stop on it to actually record the time to the lib.

So this code could be rewritten like this.
```
new_events = @metric_events.time { @filter.multi_filter(events) }
```

We can use the `@metric_events` directly since it will be correctly scoped to the `plugins/events` node

Related code that need small changes.
https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/instrument/metric.rb#L35
https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/instrument/metric.rb#L84 We will need to change the value of this constant.

logstash-core/spec/logstash/filter_delegator_spec.rb
```diff
@@ -92,6 +93,7 @@ def filter(event)
       end
 
       it "increments the in/out of the metric" do
+        allow(metric).to receive(:increment).with(:time_ms, anything)
```
ph: I think this should be an expect?

logstash-core/spec/logstash/filter_delegator_spec.rb
```diff
@@ -117,6 +119,7 @@ def filter(event)
     end
 
     it "increments the in/out of the metric" do
+      allow(metric).to receive(:increment).with(:time_ms, anything)
```
ph: Same as previous

ph:
@suyograo Each plugins has his own plugins that is generated using `SecureRandom.uuid` at startup, so the `id` settings is not required in the config file.

Settings an id is the only current way to correlate stats from different run or config changes.

ph:
@jordansissel did an initial pass, good work! :)

suyograo:
> Settings an id is the only current way to correlate stats from different run or config changes.

Sure, but how would a user know from this random UUID? The point is to guide the user to find the worst performing grok filter, no? `grok_713f61e2-6d7a-436c-9159-b3a953b18451` will not mean anything to the user.

jordansissel:
@suyograo I feel that the `how would a user know from this random UUID` problem is more a problem with the way plugin identifiers are generated in the first place. We don't have a way to "look up" a plugin by ID right now, so if there are multiples of a given plugin, it becomes nearly impossible to figure out what a given plugin metric belongs to.

In terms of incremental improvement, this patch gets us to the point where you can at least identify what the plugin name is that is behaving poorly, which is better than we have today.

i agree that the metrics themselves are very hard to consume due to the random UUID, but the "random UUID" is chosen far away from the metrics system inside Logstash, so I'm not sure what to propose as a workaround unless we solve the pipeline plugin id problem.

I'm OK delaying this PR (in idea, representing performance of individual plugin instances) until we solve the discoverability problem, "What plugin ids correspond to which part of my config?"

logstash-core/spec/logstash/filter_delegator_spec.rb
jordansissel: This test, to me, intends to test the "in" and "out" metrics. It doesn't mention any assertion about "time_ms", so it feels correct to use `allow` here because we don't care if it's called, but we must allow it otherwise the test will fail.

Thoughts?

ph: Agree, with your explanation, but its either we have a dedicated test to make sure we record the `time_ms` in a filter execution or we rename the example to be more generic and mention `time_ms`, maybe renaming it like this `records filter executions details: in/out/time_ms` would do the trick?

ph:
> I'm OK delaying this PR (in idea, representing performance of individual plugin instances) until we solve the discoverability problem, "What plugin ids correspond to which part of my config?"

I think even if the link between the plugin config and the id are hard to make we should still merge this PR, I think its valuable information.

And like you said the ID and the recording are two different thing and we can make another PR later to improve it.

jordansissel:
@ph indeed. By way of example, I've used this PR to debug a few customer filter performance issues already. For large complex configs, having a very big "Hey, date filter is consuming all your time!" signal is very valuable and knowing this helps isolate a small part of a given configuration for further troubleshooting.

I think we can document this as not-fully-baked-but-still-useful in its current state.

logstash-core/spec/logstash/filter_delegator_spec.rb
jordansissel: I think maybe I miswrote the fix here. THis `allow` belongs in a `before` statement. It's not really part of the test, just saying that such a call is OK.

Maybe I could expand it to allow *any* `:increment` calls and we just assert that `:in` and `:out` are at least invoked, because that's what the test really expects, imo.

suyograo:
> I'm OK delaying this PR (in idea, representing performance of individual plugin instances) until we solve the discoverability problem, "What plugin ids correspond to which part of my config?"

No, this is absolutely gold :) I was just trying to see if we can improve the UX

jordansissel:
> I was just trying to see if we can improve the UX

Same! I haven't figured out a good way to map from the plugin ids to the user's configuration. I think this is the first real case where we even expose the id outside Logstash, so it's good to find these frictions early ;P

I think we can follow up soon with some improvements to how to go from plugin id to plugin configuration (or line in a given config file?).

ph:
> I think we can follow up soon with some improvements to how to go from plugin id to plugin configuration (or line in a given config file?).

I really like this idea

jordansissel:
Ok, next target (after this PR) for me is some /_cat/-like API for humans to help make sense of the active config + metrics to aid in debugging for human readability

jordansissel:
@ph I tried to imrpvoe the changes I made to the specs in https://github.com/elastic/logstash/pull/5472/commits/8e25962757e9b1ed81c0019d5cc29948a019e063 -- I think it's more clear what each spec is checking, now.

acchen97:
Good stuff here @jordansissel 

Further validating this, initial API feedback from users has been the need to understand and debug exactly what pipeline components are slow or having issues i.e. is one of my Kafka inputs ingesting really slow?  +1 on tying stats of a plugin instance to the specific plugin in the config to uniquely identify when multiple instances of the same plugin are instantiated.

ph:
@jordansissel +1 for the spec changes.

suyograo:
+1 on moving forward on this as v1

jordansissel:
Some offline discussion with @suyograo -- we're going to have the per-plugin data wrapped up in an array since each entry is more a data point than a field, The initial reason for considering this was to avoid having field names in Elasticsearch like "grok_1234-14134235235-235234-234"

logstash-core/lib/logstash/filter_delegator.rb
suyograo: @jordansissel can we please change `time_ms` to `time_in_millis` to be consistent with such usages in other APIs?

jordansissel: @ph regarding @metric_events.time { ... }, method call overhead is high in Ruby/JRuby, and my experience has me try to to avoid proc/lambda blocks in performance-needing code paths. For future task, maybe we can have a Stopwatch class, something like:

```ruby
clock = @metric_events.time
@filter.multi_filter(events)
clock.stop # Compute duration and increment metric_events metric.
```

jordansissel: @suyograo yeah that sounds good.

jordansissel:
Two items remaining, I think:
* Change `time_ms` metric name to `time_in_millis` (or perhaps a better name, just `millis` or `duration_in_millis` ?
* Change the JSON structure for the per-plugin-instance stats to be an array [ { ... }, { ... } ] with one object per plugin instance.

suyograo:
+1 on duration_in_millis. Also the array structure

acchen97:
Another +1 for `duration_in_millis`.  To clarify here, for the array structure, you mean an array per plugin type, right?  i.e. the grok filter array (enumerating stats for all grok instances) example in your comment above https://github.com/elastic/logstash/pull/5472#issuecomment-225129095

logstash-core/lib/logstash/filter_delegator.rb
ph: @jordansissel If you don't provide a proc/lambda you already get a stop watch class.
See https://github.com/elastic/logstash/blob/master/logstash-core/spec/logstash/instrument/metric_spec.rb#L85

jordansissel:
* duration_in_millis change is done
* restructuring such that `{ "filters": [ { ... }, { ... }, ... ] }` each plugin instance is an array in its type (filters, inputs, outputs) -- done.

Still to do:

* use TimedExecution instead of doing time math myself in filter_delegator
* figure out how to add `"name": "grok"` (or similar) for each plugin instance. Currently we only have the ID.

ph:
> figure out how to add "name": "grok" (or similar) for each plugin instance. Currently we only have the ID.

I've thought about it yesterday and I've come up with a solution, for the `name` part you can use a `gauge` value, I see it as a snapshot of a value that will never change so we can do it in the initialize of the filter delegator, So this value could be used for filtering the data.

Later we could add a new method on the metric to save information on a namespace that will hide that implementation details.

ph:
@jordansissel I've created the fix to uses copies instead of reference in the **NamedspacedMetric** class. I suggest you cherry-pick the commit in my PR and make the following change in your code.


```diff 
diff --git a/logstash-core/lib/logstash/filter_delegator.rb b/logstash-core/lib/logstash/filter_delegator.rb
index dc5219f..d07e3a8 100644
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -38,7 +38,7 @@ module LogStash
     def multi_filter(events)
       @metric_events.increment(:in, events.size)
 
-      clock = @filter.metric.time(:duration_in_millis)
+      clock = @metric_events.time(:duration_in_millis)
       new_events = @filter.multi_filter(events)
       clock.stop
 ```

**Comment by jordansissel** [comment]
@ph done.

I have also added plugin name and wallclock (duration_in_millis) to output_delegator, also.

**Comment by suyograo** [comment]
what's up with the tests? https://travis-ci.org/elastic/logstash/builds/138207795

**Comment by jordansissel** [comment]
Oops. I'll fix the tests. I changed some internals of metrics and the specs weren't updated.

**Comment by jordansissel** [comment]
Fixed failing tests.

File: logstash-core/lib/logstash/api/commands/stats.rb [hunk 4]
```diff
@@ -49,7 +54,98 @@ def memory
...
+            (((cpu_time(hash) / @cmd.uptime * 1.0)*10000).to_i)/100.0
+          end
+
+          def cpu_time(hash)
+            hash["cpu.time"] / 1000000.0
+          end
+        end # class HotThreadsReport
+
+        module PluginsStats
+          module_function
```
ph: I dont see that often :)

ph:
This PR need a rebase to make sure the `name` node appear in the right place, other this minor thing.

LGTM for merge.

jordansissel:
Rebase done.

PR #5474, opened 2016-06-09
Fix Javafier for MapJavaProxy, ArrayJavaProxy, ConcreteJavaProxy.

### LINKED ISSUE #5465: JavaEvent#set should allow to pass a JavaMapProxy as a value
When working on the netty rewrite of the beats input I came across this issue when implementing the codecs and doing merges of values from a hash.

When you pass a `JavaMapProxy` to the set method it currently doesn't work because the Javafier doesn't know how to handle that kind of value. As noted by @guyboertje it should also handle the case of `ArrayMapProxy`.

```
java.lang.IllegalArgumentException: Missing Ruby class handling for full class name=org.jruby.java.proxies.MapJavaProxy, simple name=MapJavaProxy
	at com.logstash.Javafier.deep(Javafier.java:111)
	at com.logstash.ext.JrubyEventExtLibrary$RubyEvent.ruby_set_field(JrubyEventExtLibrary.java:159)
	at com.logstash.ext.JrubyEventExtLibrary$RubyEvent$INVOKER$i$2$0$ruby_set_field.call(JrubyEventExtLibrary$RubyEvent$INVOKER$i$2$0$ruby_set_field.gen)
	at org.jruby.internal.runtime.methods.AliasMethod.call(AliasMethod.java:66)
	at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:202)
	at org.jruby.ast.FCallTwoArgNode.interpret(FCallTwoArgNode.java:38)
```
cc @guyboertje @colinsurprenant

guyboertje:
+ JUnit tests + Rspec test

Fixes #5465


logstash-core-event-java/src/main/java/com/logstash/Javafier.java
```diff
@@ -1,42 +1,49 @@
...
-import org.jruby.RubyFixnum;
-import org.jruby.RubyTime;
-import org.jruby.RubySymbol;
-import org.jruby.RubyBignum;
-import org.jruby.ext.bigdecimal.RubyBigDecimal;
 import com.logstash.ext.JrubyTimestampExtLibrary;
-import org.jruby.runtime.builtin.IRubyObject;
-import java.math.BigDecimal;
 import org.joda.time.DateTime;
+import org.jruby.*;
```
colinsurprenant: Not sure where this was discussed but I would prefer we keep explicit classes imports?

logstash-core-event-java/src/main/java/com/logstash/Util.java
```diff
@@ -1,13 +1,26 @@
...
+    public static Object getMapFixture() throws IOException {
+        StringBuilder json = new StringBuilder();
+        json.append("{");
+        json.append("\"string\": \"foo\", ");
+        json.append("\"int\": 42, ");
+        json.append("\"float\": 42.42, ");
+        json.append("\"array\": [\"bar\",\"baz\"], ");
+        json.append("\"hash\": {\"string\":\"quux\"} }");
+
+        ObjectMapper mapper = new ObjectMapper();
```
colinsurprenant: why are we going through Jackson to produce the object?

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
guyboertje: I did not pay attention to what intellij did. I let it manage the adding of the import lines unless its confused and then I add one in manually.
I will change it.

logstash-core-event-java/src/main/java/com/logstash/Util.java
guyboertje: Only to mimic the way that @ph induced the error.

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
```diff
@@ -127,6 +134,16 @@ public static Object deep(IRubyObject o) {
             case RubyNil: return deep((RubyNil)o);
             case True: return deep((RubyBoolean.True)o);
             case False: return deep((RubyBoolean.False)o);
+            case MapJavaProxy: return deep(((MapJavaProxy) o).to_hash());
+            case ArrayJavaProxy:
+            case ConcreteJavaProxy:
+                Object obj = JavaUtil.unwrapJavaObject(o);
```
guyboertje: I thought about what you said regarding using a method here and I think we should.
I suggest that we raise an error rather than fall through when we get a ConcreteJavaProxy value that we have not coded for. WDYT?

ph:
@guyboertje I can confirm that this change fix the problem I had with my beats rewrite, thank you very much.

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
colinsurprenant: ah yes, Intellij at some point does that :(

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
```diff
@@ -1,42 +1,60 @@
...
-        // TODO: (colin) investagate why .toJavaArrayUnsafe() which should be faster by avoiding copying produces nil values spec errors in arrays
-        for (IRubyObject o : a.toJavaArray()) {
+        for (IRubyObject o : a) {
             result.add(deep(o));
         }
         return result;
     }
 
+    public static List<Object> deep(RubyArray a) {
+        return deep( a.toJavaArray());
```
colinsurprenant: cosmetic: `deep( a.toJavaArray())` -> `deep(a.toJavaArray())`

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
```diff
@@ -101,6 +119,25 @@ public static Object deep(RubyBoolean.False b) {
...
+        if (obj instanceof List) {
+            return deepList((List<Object>) obj);
+        }
+        StringBuilder sb = new StringBuilder();
+        sb.append("Missing Ruby class handling for full class name=");
+        sb.append(jp.getClass().getName());
+        sb.append(", simple name=");
+        sb.append(jp.getClass().getSimpleName());
+        sb.append(", wrapped object=");
+        sb.append(obj.getClass());
```
colinsurprenant: suggestion: maybe use `String.format` instead?
use a constant string pattern like 
```
private static final String TMPL = "Missing Ruby class handling for full class name=%s, simple name=%s, wrapped object=%s";

...

throw new IllegalArgumentException(String.format(TMPL, jp.getClass().getName(), jp.getClass().getSimpleName(), obj.getClass());
```

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
```diff
@@ -127,6 +164,9 @@ public static Object deep(IRubyObject o) {
             case RubyNil: return deep((RubyNil)o);
             case True: return deep((RubyBoolean.True)o);
             case False: return deep((RubyBoolean.False)o);
+            case MapJavaProxy: return deep(((MapJavaProxy) o).to_hash());
```
colinsurprenant: does the `.to_hash()` call involve any copying? did not dig into this but we should make sure we are not ending up doing a double-conversion here

guyboertje: Will check

logstash-core-event-java/spec/event_spec.rb
```diff
@@ -5,6 +5,7 @@
 require "logstash/event"
 require "json"
 require "java"
+# java_import 'com.logstash.Util'
```
talevy: can we remove this?

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
```diff
@@ -49,6 +81,20 @@ public void visit(IRubyObject key, IRubyObject value) {
         return result;
     }
 
+    private static Object deepAnything(Object o) {
```
talevy: why don't we have this be `deep(Object o)`? and separate the concerns accordingly as we get more specific

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
```diff
@@ -200,5 +226,9 @@ public static Object deep(IRubyObject o) {
         ArrayJavaProxy,
         ConcreteJavaProxy
     }
+
+    private static String missingHandlerString(String fmt, String... subs) {
```
colinsurprenant: not sure why this method is needed...? you can keep it if you want but if I ever modify this area I might inadvertently replace it with a direct `String.format` call 👼

logstash-core-event-java/src/test/java/com/logstash/JavafierTest.java
```diff
@@ -2,17 +2,99 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
 
 public class JavafierTest {
+    public static final Ruby ruby;
```
talevy:
```
public static final Ruby ruby = Ruby.getGlobalRuntime();
```
is better, no?

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
```diff
@@ -1,44 +1,76 @@
...
 
-        // TODO: (colin) investagate why .toJavaArrayUnsafe() which should be faster by avoiding copying produces nil values spec errors in arrays
-        for (IRubyObject o : a.toJavaArray()) {
+        for (IRubyObject o : a) {
             result.add(deep(o));
         }
         return result;
     }
 
+    public static List<Object> deep(RubyArray a) {
```
talevy: this doesn't seem to be used

talevy: err, at least it is not being tested. From my understanding `deep(IRubyObject)` will be called for these instances, since `RubyArray` might stay as generic as an IRubyObject

logstash-core-event-java/spec/event_spec.rb
guyboertje: Sure, will do.

logstash-core-event-java/src/main/java/com/logstash/Javafier.java
guyboertje: This method is called by line 173 - `case RubyArray: return deep((RubyArray)o);`. Otherwise, it would recursively call itself.

guyboertje:
@talevy, @colinsurprenant - I appreciate your concerns about some of `Javafier`'s method names and arguments but, frankly, as soon as this merges, I will have to refactor these JavaProxy changes into  #4999 anyway as it brings sweeping changes to `Javafier`.
#4999 also introduces a TestBase class that does the setup of the ruby runtime, equivalent of 'require "bigdecimal"' and 'require "logstash/timestamp"', once per JUnit Suite.

colinsurprenant:
@guyboertje as noted, my comment was essentially cosmetic. 
LGTM

PR #5486, opened 2016-06-14
Fixes QA acceptance test update issue

### LINKED ISSUE #5468: Acceptance tests plugins update scenario broken.
The update test in the acceptance layer does not work as expected causing many issues as you can see in https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-acceptance/label=metal,suite=debian/15/console 

The source of this error is the code used to remove the version constran in the gemfile for the test gem used in tests, this is required because in logstash if you previously installed an specific version of a plugin you can not use the update command, you should remove it and do the install.

```
  before do
      logstash.run_command_in_path("bin/logstash-plugin install --version #{previous_version} #{plugin_name}")
      # Logstash wont update when we have a pinned versionin the gemfile so we remove them
      logstash.replace_in_gemfile(',\s"0.1.0"', "")
      expect(logstash).to have_installed?(plugin_name, previous_version)
    end
```

we need to improve the interaction specified in line `logstash.replace_in_gemfile(',\s"0.1.0"', "")`, as the cause of having this failing test.

purbon:
For some platforms, the update command has been broken due to the fact that for one of the plugins involved (webhdfs output) the requested versions list got understood by `Net::HTTP` as being ASCII and not `UTF-8`. This PR proposes a fix for that that does not require lots of investment due to the small usage of this feature internally.

This PR also updates the latest logstash released to alpha3, and removed the patch introduced to uniformize the names, now all have no arch so it makes no more sense.

Fixes #5468

### DIFF
--- a/lib/bootstrap/rubygems.rb
+++ b/lib/bootstrap/rubygems.rb
@@ -43,6 +43,16 @@ module LogStash
       end
     end
 
+    ##
+    # Take a plugin name and get the latest versions available in the gem repository.
+    # @param [String] The plugin name
+    # @return [Hash] The collection of registered versions
+    ##
+    def versions(plugin)
+      require "gems"
+      require_relative "patches/gems"
+      Gems.versions(plugin)
+    end
     # Take a gem package and extract it to a specific target
     # @param [String] Gem file, this must be a path
     # @param [String, String] Return a Gem::Package and the installed path
--- a/qa/sys/debian/user_bootstrap.sh
+++ b/qa/sys/debian/user_bootstrap.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 
 VERSION=`cat /vagrant/config/platforms.json | grep  latest | cut -d":" -f2 | sed 's/["\|,| ]//g'`
-LOGSTASH_FILENAME="logstash-${VERSION}_all.deb"
+LOGSTASH_FILENAME="logstash-${VERSION}.deb"
 wget -q https://download.elastic.co/logstash/logstash/packages/debian/$LOGSTASH_FILENAME
-mv $LOGSTASH_FILENAME "logstash-${VERSION}.deb" # necessary patch until new version with the standard name format are released
--- a/lib/bootstrap/patches/gems.rb
+++ b/lib/bootstrap/patches/gems.rb
@@ -1,0 +1,16 @@
+# encoding: utf-8
+require "gems"
+
+# This patch is necessary to avoid encoding problems when Net:HTTP return stuff in ASCII format, but
+# consumer libraries, like the YAML parsers expect them to be in UTF-8. As we're using UTF-8 everywhere
+# and the usage of versions is minimal in our codebase, the patch is done here. If extended usage of this
+# is done in the feature, more proper fix should be implemented, including the creation of our own lib for
+# this tasks.
+module Gems
+  module Request
+    def get(path, data = {}, content_type = 'application/x-www-form-urlencoded', request_host = host)
+      request(:get, path, data, content_type, request_host).force_encoding("UTF-8")
+    end
+  end
+end
+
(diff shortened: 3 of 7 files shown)


qa/sys/debian/user_bootstrap.sh
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 
 VERSION=`cat /vagrant/config/platforms.json | grep  latest | cut -d":" -f2 | sed 's/["\|,| ]//g'`
-LOGSTASH_FILENAME="logstash-${VERSION}_all.deb"
+LOGSTASH_FILENAME="logstash-${VERSION}.deb"
```
jsvd: why is this change in this PR? I'm not familiar with the naming requirements, are we removing architectures from package filenames?

purbon: yeah, before alpha3 we had arch in the package name, not anymore.

lib/bootstrap/rubygems.rb
```diff
@@ -43,6 +43,25 @@ def self.reset
...
+    # @return [Hash] The collection of registered versions
+    ##
+    def versions(plugin)
+      require "gems"
+      # This patch is necessary to avoid encoding problems when Net:HTTP return stuff in ASCII format, but
+      # consumer libraries, like the YAML parsers expect them to be in UTF-8. As we're using UTF-8 everywhere
+      # and the usage of versions is minimal in our codebase, the patch is done here. If extended usage of this
+      # is done in the feature, more proper fix should be implemented, including the creation of our own lib for
+      # this tasks.
+      ::Gems::Request.module_eval do
```
jsvd: can't we move this into `logstash-core/lib/logstash/patches` like the rest of the monkeypatches/fixes we did?

qa/sys/debian/user_bootstrap.sh
jsvd: makes sense, please keep these 2 concerns in separate commits during merge

purbon: it is in separate commits, upgrade to alpha3 is what make this to be changed 👍

lib/bootstrap/rubygems.rb
purbon: this is for plugin manager, I guess this is again an chicken egg problem, so I would say, we can not really do it.

ph: I understand we cannot move this patch to the `logstash-core` because of the PATH handling, But I think the idea behind @jsvd comment is to externalize this patch outside of where its called. If we ever need to remove it, this patch will be really hidden somewhere and not as obvious as a file in the `logstash/patches` directory, the plugin manager has a few required patch to allow it to work correctly. 

Also having this patch inline within the `#versions` method mean jruby will evaluate the `module_eval` for every method call, in the context of jruby this mean the methods cache will be dropped on every method call.

This kind of behavior from gems/rubygems makes me really sad, I still don't understand were in the chain the utf-8 information of the request isn't made.

purbon: I understand, I will put this in a similar location as the ones in core, it makes sense. On the other side, keep in mind this is a command line tool that run once, so the way this works is actually not that important because the module_eval is going to be done one time and only when the versions function is used. Makes sense to you?

purbon:
@ph @jsvd updated this with your request, let me know if you have any more questions 😄

lib/bootstrap/patchs/gems.rb
```diff
@@ -0,0 +1,16 @@
+# encoding: utf-8
```
ph: can we name the directory patches like we do in the core?

ph:
I've tested it with a debian-8 machine defined in the `qa/vagrantfile` and this fixed worked when doing a plugin update. I've removed the patch and the update fails.

Added a minor comment concerning the name of the directory, after that change it will be good to merge.

lib/bootstrap/patchs/gems.rb
purbon: sure

purbon:
renaming the patches directory to be like in core and merging, thanks @ph for your review.

Issue #1511, opened 2014-07-08
Slow even when running version


smithamax:
I'm running logstash on my mac, and on a linux box. It takes ages to start up on both.

even when running a simple `logstash version` it takes more than 10 seconds.

on os x
```
$ time logstash version
logstash 1.4.1

real  0m13.211s
user 0m15.355s
sys   0m0.724s
```

on linux
```
$ time /opt/logstash/bin/logstash version
logstash 1.4.2-modified

real  0m36.963s
user 0m34.432s
sys   0m1.997s
```

Does anyone have any insight into why this takes so long and if there is a solution?

untergeek:
Closing.  Feel free to re-open, if desired.

timmolter:
I'd like to report this issue too:

    $ time /opt/logstash/bin/logstash version
    logstash 1.4.2-modified

    real	1m31.590s
    user	1m47.771s
    sys	0m4.444s

Running any thing takes over a minute to start. How can I debug the problem? Are there some flags I can set to ignore anything? What's the reason for the slow startup?

untergeek:
```
$ time bin/logstash --version
logstash 2.0.0.dev

real	0m4.149s
user	0m10.876s
sys	0m0.494s
```

This is beyond strange.  What hardware are you running on?

timmolter:
I have a Linux Box running 32-bit Lubuntu 12.04 with a quad core Intel Atom, 4GB RAM. The thing is that I run other Java apps on it, and they start up instantly. Any tips to find the bottleneck?

untergeek:
How much RAM is free?  Logstash is pretty hungry.  It will want 500M for the heap, minimum.  It could be that you're hitting swap, which would definitely explain the slow-down.

Which Java are you using?

Also, and Atom processor, 4 cores notwithstanding, is not a high-powered CPU.

timmolter:
344 MB free. That's probably the issue. I'm using Java 8, the latest from Oracle. 

Is Logstash claiming 500MB at startup even if it's not needed? For just `logstash --version` for example. That seems unecessary. Where are the JVM memory flags set?

untergeek:
```
$ cat bin/logstash
#!/bin/sh
# Run logstash from source
#
# This is most useful when done from a git checkout.
#
# Usage:
#     bin/logstash <command> [arguments]
#
# See 'bin/logstash help' for a list of commands.
#
# Defaults you can override with environment variables
[...]

PR #5490, opened 2016-06-14
make -V/--version fast on *nix

jsvd:
```
/tmp/logstash-5.0.0 % time bin/logstash -V
logstash 5.0.0
bin/logstash -V  27.79s user 1.34s system 270% cpu 10.773 total

/tmp/logstash-5.0.0 % patch -p1 < /tmp/patch
patching file bin/logstash

/tmp/logstash-5.0.0 % time bin/logstash -V
logstash 5.0.0
bin/logstash -V  0.01s user 0.02s system 101% cpu 0.029 total
```

Still figuring out how to do the same in Windows BAT scripts..

### DIFF
--- a/bin/logstash
+++ b/bin/logstash
@@ -46,4 +46,11 @@ fi
 . "$(cd `dirname $SOURCEPATH`/..; pwd)/bin/logstash.lib.sh"
 setup
 
-ruby_exec "${LOGSTASH_HOME}/lib/bootstrap/environment.rb" "logstash/runner.rb" "$@"
+if [ "$1" = "-V" ] || [ "$1" = "--version" ];
+then
+  LOGSTASH_VERSION_FILE="${LOGSTASH_HOME}/logstash-core/lib/logstash/version.rb"
+  LOGSTASH_VERSION="$(sed -ne 's/^LOGSTASH_VERSION = "\([^*]*\)"$/\1/p' $LOGSTASH_VERSION_FILE)"
+  echo "logstash $LOGSTASH_VERSION"
+else
+  ruby_exec "${LOGSTASH_HOME}/lib/bootstrap/environment.rb" "logstash/runner.rb" "$@"
+fi
--- a/qa/acceptance/spec/shared_examples/cli/logstash/version.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash/version.rb
@@ -14,8 +14,14 @@ shared_examples "logstash version" do |logstash|
 
     context "on #{logstash.hostname}" do
       it "returns the right logstash version" do
-        result = logstash.run_command_in_path("bin/logstash --path.settings=/etc/logstash --version")
+        result = logstash.run_command_in_path("bin/logstash --version")
         expect(result).to run_successfully_and_output(/#{LOGSTASH_VERSION}/)
+      end
+      context "when also using the --path.settings argument" do
+        it "returns the right logstash version" do
+          result = logstash.run_command_in_path("bin/logstash --path.settings=/etc/logstash --version")
+          expect(result).to run_successfully_and_output(/#{LOGSTASH_VERSION}/)
+        end
       end
     end
   end


jsvd:
this partially addresses #1511

jsvd:
We can merge this as is now and open another PR to accomplish the same on windows. removing WIP tag.

bin/logstash
```diff
@@ -46,4 +46,11 @@ fi
 . "$(cd `dirname $SOURCEPATH`/..; pwd)/bin/logstash.lib.sh"
 setup
 
-ruby_exec "${LOGSTASH_HOME}/lib/bootstrap/environment.rb" "logstash/runner.rb" "$@"
+if [ "$1" = "-V" ] || [ "$1" = "--version" ];
+then
+  LOGSTASH_VERSION_FILE="${LOGSTASH_HOME}/logstash-core/lib/logstash/version.rb"
+  LOGSTASH_VERSION=$(awk '/./{line=$0} END{print line}' $LOGSTASH_VERSION_FILE | cut -d '"' -f 2)
```
jordansissel: If you just want the last line, `sed -ne '$p'` is shorter than the awk you have.

A possibly even shorter example with just sed:

```
LOGSTASH_VERSION="$(sed -ne 's/^LOGSTASH_VERSION = "\([^*]*\)"$/\1/p' logstash-core/lib/logstash/version.rb)"
```

my main concern is whether or not this works on OSX and our supported Linuxes. `awk` and `cut` may not always be installed by default (nor sed, for that matter). Thoughts?

talevy:
```bash
(pr/5490)$ bin/logstash -V
logstash 5.0.0.dev
(pr/5490)$ bin/logstash --version
logstash 5.0.0.dev
```

LGTM

jordansissel:
I like this change. It adds a constraint on the format of the version.rb, which is ok, but maybe we should add an acceptance test to verify the output?

I am +1 on doing Windows in a future PR.

bin/logstash
jsvd: we already have other code that uses sed and awk:
```
bin/logstash:    SOURCEPATH=$(stat -c %N $0 | awk '{print $3}' | sed -e 's/\‘//' -e 's/\’//')
bin/logstash.lib.sh:    SOURCEPATH=$(stat -c %N $0 | awk '{print $3}' | sed -e 's/\‘//' -e 's/\’//')
```

That said I also was a bit reluctant on using "advanced sed" features due to compatibility

jordansissel:
I'm OK if we don't have tests for now, but would love tests to verify this stays working in the future. LGTM

jsvd:
I'll add a second acceptance test in https://github.com/elastic/logstash/blob/master/qa/acceptance/spec/shared_examples/cli/logstash/version.rb to ensure we test this PR's code path

qa/acceptance/spec/shared_examples/cli/logstash/version.rb
```diff
@@ -14,9 +14,15 @@
 
     context "on #{logstash.hostname}" do
       it "returns the right logstash version" do
-        result = logstash.run_command_in_path("bin/logstash --path.settings=/etc/logstash --version")
+        result = logstash.run_command_in_path("bin/logstash --version")
         expect(result).to run_successfully_and_output(/#{LOGSTASH_VERSION}/)
       end
+      context "when also using the --path.settings argument" do
```
purbon: if we want to be strict about semantic, this is not another context, is just another test with different input, so I would remove the context here.

bin/logstash
purbon: agree here with @jordansissel concerns, we should be use most cross platform methods possible, my question is why seed and awk, we could also use `grep` and `cut`, right? mostly as you have in line 52, what do you think?

qa/acceptance/spec/shared_examples/cli/logstash/version.rb
jsvd: my idea was that, although we don't have it here, we could create a test that says that "if this path.settings" is invalid, we expect it to not run.
so this context would have 2 tests

bin/logstash
jsvd: as I said, we are already using awk and sed on this script, so I used them too.

qa/acceptance/spec/shared_examples/cli/logstash/version.rb
purbon: +1 for this, you should definitely add this other test too, should not be hard, right?

bin/logstash
purbon: ok, then no problem :-P

jsvd: I'll change to Jordan awesome one liner suggestion

purbon:
LGTM, change is very powerful. we should think about generalizing this kind of approach for the feature.

jsvd:
also merged into 5.0 and 2.x (for future 2.4.0):

```
[logstash:5.0] 2 new commits by Joao Duarte:
`7bf89aa`  make -V/--version fast on *nix - Joao Duarte
`ae09503`  add acceptance test - Joao Duarte 

[logstash:2.x] 1 new commit by Joao Duarte:
`3b19782`  make -V/--version fast on *nix - Joao Duarte 
```

PR #5499, opened 2016-06-15
[Backport from master] Add new :list property to configuration parameters.

andrewvc:
If set to try this will allow the user to specify one or more values.
This generally replaces the :array type, which had fewer type checks.

The array type is still needed for lists of complex objects, e.g. hashes.


suyograo:
LGTM

jordansissel:
LGTM

PR #5502, opened 2016-06-15
Revert geoip filter to 2.x

suyograo:
Previous update had included 3.x of geoip filter which was a rewrite for GeoIP2. This should not be included in a bug fix release, so reverting to 2.x


jsvd:
LGTM

Issue #5504, opened 2016-06-15
ubuntu-1604 issues in acceptance testing with installation and update


purbon:
seen at https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-acceptance/label=metal,suite=debian/22/console

 ```
Failures:

  1) artifacts operation behaves like installable is running on ubuntu-1604
     Failure/Error: expect(logstash).to be_running
       expected #<ServiceTester::Artifact:0x5149f008 @host="127.0.0.1:2202", @client=#<ServiceTester::Ubuntu1604Commands:0x7072bc39>, @options={"host"=>"ubuntu-1604", "type"=>"debian"}> to be running
     Shared Example Group: "installable" called from /var/lib/jenkins/workspace/elastic+logstash+master+multijob-acceptance/label/metal/suite/debian/qa/acceptance/spec/lib/artifact_operation_spec.rb:12
     # /var/lib/jenkins/workspace/elastic+logstash+master+multijob-acceptance/label/metal/suite/debian/qa/acceptance/spec/shared_examples/installed.rb:17:in `(root)'
     # /var/lib/jenkins/workspace/elastic+logstash+master+multijob-acceptance/label/metal/suite/debian/qa/Rakefile:58:in `(root)'

  2) artifacts operation behaves like updated can be updated an run on ubuntu-1604
     Failure/Error: expect(logstash).to be_running
       expected #<ServiceTester::Artifact:0x5149f008 @host="127.0.0.1:2202", @client=#<ServiceTester::Ubuntu1604Commands:0x7072bc39>, @options={"host"=>"ubuntu-1604", "type"=>"debian"}> to be running
     Shared Example Group: "updated" called from /var/lib/jenkins/workspace/elastic+logstash+master+multijob-acceptance/label/metal/suite/debian/qa/acceptance/spec/lib/artifact_operation_spec.rb:13
     # /var/lib/jenkins/workspace/elastic+logstash+master+multijob-acceptance/label/metal/suite/debian/qa/acceptance/spec/shared_examples/updated.rb:23:in `(root)'
     # /var/lib/jenkins/workspace/elastic+logstash+master+multijob-acceptance/label/metal/suite/debian/qa/Rakefile:58:in `(root)'

Finished in 52 minutes 21 seconds (files took 1.04 seconds to load)
72 examples, 2 failures, 4 pending
```

purbon:
[...]

PR #5506, opened 2016-06-16
Add the idea of experimental platforms for the acceptance testing framework

### LINKED ISSUE #5509: Add the idea of experimental platforms to the Acceptance test framework
This will let us keep focus on platforms inside the support matrix, but also do some testing on upcoming platforms without mixing concepts. For example like happens with current failures inside ubuntu-1604 (#5504).

The experimental layer will need to be included as a new step inside https://logstash-ci.elastic.co .

This will keep the acceptance testing strictly to match the support matrix while letting us test on upcoming supported platforms and be sure they are in good health for our product. What do you think?

purbon:
This will let us keep focus on platforms inside the support matrix, but also do some testing on upcoming platforms without mixing concepts. For example like happens with current failures inside ubuntu-1604 (https://github.com/elastic/logstash/issues/5504). 

The experimental layer will need to be included as a new step inside https://logstash-ci.elastic.co .

This will keep the acceptance testing strictly to match the support matrix while letting us test on upcoming supported platforms and be sure they are in good health for our product. What do you think?

Fixes #5509


ph:
I understand the intend, I find it similar to the possibility of making a test as pending. Marking a test as pending or disabling a test is the best way to make it rot. 

I think Its wiser and easier to communicate to have the the following strategy: when we add a new machine in the platforms matrix, we or the community make a commitment to support it until it make no sense to support it. Since we can run the tests locally we have the possibility to test how logstash behave before deciding to support a platforms. 

So in my opinion I am not in favor of having this feature, I am more into lets reduce the scope. But, if we decide to go ahead with a concept of experimental maybe its the responsibility of the CI to mark a job as experimental and not take it into consideration when this job fails for the dependent job.

This PR target explicitly ubuntu-1604, which is the latest Ubuntu LTS, I think we should concentrate into fixing the root problem before having the possibility to mark  as experimental.

purbon:
Thanks for your feedback @ph for me experimenting is the idea of letting CI evaluate platforms that we still don't have in the support matrix, as this test take long time I don't think anyone will run them locally. 

I think I will create also an issue to have the discussion there.

qa/Rakefile
```diff
@@ -36,10 +42,10 @@ namespace :qa do
     desc "Halt all VM's involved in the acceptance test round"
     task :halt, :platform do |t, args|
       config   = PlatformConfig.new
-      machines = config.select_names_for(args[:platform])
-      message = "halting all VM's defined inside Vagrantfile"
-      message  = "#{message} for #{args[:platform]}: #{machines}" if !args[:platform].nil?
-      puts message
+      experimental = (ENV['LS_QA_EXPERIMENTAL'].to_s.downcase || "false") == "true"
```
suyograo: we should make it `LS_QA_EXPERIMENTAL_OS`

suyograo:
+1 on having support for this in our infra.

qa/Rakefile
purbon: works for me.

purbon:
updated.

suyograo:
@ph I like the idea of us having some infra to test new OS's, so lets move forward with it. How this will be enabled is a different discussion, which needs to be followed up with the infra team.

purbon:
agreed, if we like the idea of having this extra OS's on infra, I will
handle this internally, pleasure! but will for sure be separate from our
acceptance test layer that is what must validate a release to be green.

ph:
@suyograo @purbon OK, But I believe we should mark Ubuntu1604 as experiemental.

purbon:
@ph ubuntu-1604 is not in our support matrix, so until it reach there, is experimental 👍

purbon:
Ubuntu1604 has been always experimental in this pr,  check the config file.

qa/Rakefile
```diff
@@ -24,10 +30,10 @@ namespace :qa do
     desc "Bootstrap all the VM's used for this tests"
     task :setup, :platform do |t, args|
       config   = PlatformConfig.new
-      machines = config.select_names_for(args[:platform])
+      experimental = (ENV['LS_QA_EXPERIMENTAL_OS'].to_s.downcase || "false") == "true"
+      machines = config.select_names_for(args[:platform], {"experimental" => experimental})
```
ph: We are using rake task arguments and an environment variable in the same task can we consolidate in only one way?
Since we are using rake arguments I think we should use that for the experimental option? Opinions?

purbon: My thought on using an env var was because it changes behaviour of same
task, run default or experimental ones. It makes the interface a bit more
clear to me. But i'm good with whatever we agree with.

ph: Good reasoning. I wish rake would provide a clearer interface to arguments.

untergeek:
We should have been testing against Ubuntu 15.10 and the pre-releases of 16.04.  It's kind of embarrassing to not be able to run properly on a brand new LTS release.  

"Not listed on our support matrix" sounds like a bad excuse to me.

purbon:
Don't get me  wrong @untergeek, i all in favor of your approach, this pr just try to have a middle ground in between the we test all, and we do acceptance with thw current suported ones. In ci we should have both. Makes sense to you?

purbon:
This started to have all in ci, as we removed fedoras a few days ago after @suyograo  proposal.

ph:
Lets get a complete run and fix ubuntu.

purbon:
So, should we merge this pr, or just remove ubuntu16 and add it back when we decide to? I'm +1 for as many test as possible, but i'm with the quorum here.

ph:
I haven't run the test locally but the code LGTM, I've paid special attention to the filtering logic.

I wish theses test were instant so we could be able to run them with the ci before merging in master, I am wondering if we could make a feature branch that run theses tests? @purbon could that be possible with our current infra?

purbon:
why do we want to check until acceptance test for pr's? Seems a bit too
much, right?, we should check until integration, might be some layers of
test here should be there. And we should run them in ci, we don't do that
now.... This would be a nice addition for the os compat layer.

PR #5511, opened 2016-06-16
Fix 'Socket' require in environment.rb

andrewvc:
Logstash devutils would raise an error when including this file
due to this. Not really possible to test

### DIFF
--- a/logstash-core/lib/logstash/environment.rb
+++ b/logstash-core/lib/logstash/environment.rb
@@ -2,6 +2,7 @@
 require "logstash/errors"
 require "logstash/config/cpu_core_strategy"
 require "logstash/settings"
+require "socket"
 
 module LogStash
 


ph:
This PR is the same as https://github.com/elastic/logstash/pull/5512 ?

logstash-core/lib/logstash/environment.rb
```diff
@@ -2,6 +2,7 @@
 require "logstash/errors"
 require "logstash/config/cpu_core_strategy"
 require "logstash/settings"
+require "socket"
```
ph: OK got it why we need it. make sense!

https://github.com/elastic/logstash/pull/5511/files#diff-b803768250328614e95dee4f5b27af18R10

ph:
LGTM

PR #5512, opened 2016-06-16
Fix a regression where untyped fields would return null values

andrewvc:
This patch fixes a bug where config fields declared without a type
would just return null.


ph:
LGTM

PR #5522, opened 2016-06-20
make acceptance test framework pick snapshot rpm/deb builds

jsvd:
by default, logstash will generate rpm/deb files called logstash-#{LOGSTASH_VERSION}-SNAPSHOT.rpm/deb

this PR reflects that change so that acceptance tests can be executed correctly

### DIFF
--- a/qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
@@ -5,7 +5,7 @@ require "fileutils"
 
 shared_examples "logstash install" do |logstash|
   before(:each) do
-    logstash.install(LOGSTASH_VERSION)
+    logstash.install({:version => LOGSTASH_VERSION})
   end
 
   after(:each) do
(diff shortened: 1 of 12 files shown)


purbon:
Thanks for doing this change @jsvd, I have however a couple of comments.

As is implemented now all packages use the same naming, so all are `logstash-#{version}.[rpm|deb]`, including the ones downloaded for the update logstash scenario, you can see this in https://github.com/elastic/logstash/blob/master/qa/sys/debian/user_bootstrap.sh for example.

As https://github.com/elastic/logstash/blob/master/qa/acceptance/spec/shared_examples/updated.rb#L14 install only based on the version, we need to find a way to tell is not going to have the snapshot appendinx. I think I'm good for not using anymore only the version and using the full name, what do you think?

Thanks a lot!

- purbon

jsvd:
I'm +1 on not breaking existing code, we need a more generic api that accepts a file path instead of a logstash version.
My knowledge of this code base is limited, so a proper fix will take a bit more time.

purbon:
@jsvd no problem, I will propose a change to you now.

purbon:
finally the proposed fix is https://github.com/jsvd/logstash/pull/2

purbon:
LGTM

jsvd:
LGTM :allthethings:

PR #5530, opened 2016-06-21
Set master's version to 5.0.0-alpha4

suyograo:
Set master branch's version to 5.0.0-alpha4 to match other projects like ES, Kibana and Beats. Alphas will be released from the master branch. This is to make the unified release consistently work across products


talevy:
LGTM

jsvd:
this is no longer valid as alpha4 has been released

Issue #5534, opened 2016-06-22
Harden Gemfile used in release packages


suyograo:
We should harden the [Gemfile](https://github.com/elastic/logstash/blob/5.0/Gemfile) used in release branches to be more restrictive w.r.t. versions. Today, its really easy to slip and include a major version of a plugin in a LS minor release, as long as its compatible with the core APIs. This is definitely not desirable from a user perspective because plugin major versions _can_ break compatibility. And we don't intend to break compatibility in LS minor releases. One protection would be to add tighter constraints to every plugin specified in `Gemfile` such that minor LS versions will only include patch releases for plugins. By using the [pessimistic version constraint](http://guides.rubygems.org/patterns/#pessimistic_version_constraint).

For example:

```
gem "logstash-filter-grok", "~> 3.0"
```

This means grok plugin versions `>= 3.x` is allowed but not ones `>= 4.0`. There would be no danger of accidentally including a major version release of a plugin for LS minor releases. 

For LS major releases, a new constraint file needs to be generated which relaxes the constraint to include the next majors.

Thoughts? /cc @jsvd @ph @jordansissel

jsvd:
> One protection would be to add tighter constraints to every plugin specified in Gemfile such that minor LS versions will only include patch releases for plugins. By using the pessimistic version constraint.

I'm not sure if that is your intent, but I'd guess we want to set the pessimistic constraint to the minor version, not patch. So:

`gem "logstash-filter-grok", "~> 3"`

This way it won't upgrade to 4 but can upgrade to 3.1, 3.2, 3.2.3, etc

jsvd:
done in #5535

Issue #5533, opened 2016-06-22
Add default plugins to master Gemfile


suyograo:
Previously, master branch in LS (since 5.0) had a different `Gemfile` than release branches. With the merge of #5460, both release and master branches use the same strategy of using local, unpublished `logstash-*` plugins for packages. With this, there is really no need to deviate `Gemfile` for release branches. Lets make them same. This will settle a lot of merge conflicts as well. Also, we should really be able to release 5.0.0 from master itself. See #5530

suyograo:
@jsvd @ph @jordansissel thoughts?

jsvd:
I'm -1 on adding the default plugins to the gemfile. you often might be working only in core so you don't need the default plugins slowing down your workflow.
Release branches are, however, dedicated to creating releases so it makes sense there to encode the default plugins list.
We can create a rake task that does this updating/injection for us, though.

ph:
I've gave some thought on this since our original discussion from the unified releases.

@jsvd I agree that this could slow down our workflow, switching branch would often mean in some way to  kickstart an install the plugin, some part of this process will be faster since some gems will be present locally in the gem cache.

@suyograo From your proposition the **Gemfile** will be the same in master and in the branches? The only thing that will differ is the **lockfile** wont be present in the master branch but will be available in the releases branch? Correct?

suyograo:
> The only thing that will differ is the lockfile wont be present in the master branch but will be available in the releases branch? Correct?

Correct. 

I do like the simplicity in this approach -- having consistent `Gemfile` across branches going forward. I understand it may slow down development, but I think thats a technical problem we can solve. This is a problem we have to solve regardless. I feel the merges will get easier too.

Others? @colinsurprenant @jordansissel ?

jordansissel:
I think the right place to add this, and would solve @jsvd's concerns, is in a `group` in the Gemfile. @colin has noted previously (on zoom, I think?) that the Gemfile parser that the plugin manager uses doesn't understand groups, but we could fix that and use a group, maybe call it `group :default_plugins` and go from there.

I am +1 on using the Gemfile to include the default plugins, and in favor of using a group to fence it so that installing plugins during development is optional.

suyograo:
I like the `group` idea!

ph:
Adding a `group` make sense and with that I +1 that issue.

ph:
Just to be clear, to make it right lets make sure we support the 2 group definition syntax:

```ruby
group :default_plugins do
end

[...]

PR #5535, opened 2016-06-22
pin major versions of default plugins in Gemfile

jsvd:
implements https://github.com/elastic/logstash/issues/5534


suyograo:
LGTM, did not test

suyograo:
I'd like @ph to check this out as well.

ph:
The list look sane to me, I've checked the most weird one like the SNS (v4) and the input rabbitmq (v5), The choice of dependency `~> MAJOR` is correct and will pick up the all the minor and release version of the plugins.

I will do a bit of testing with the update scenario.

ph:
I've tested the update scenario using the `logstash-filter-qatest` to make sure that the update still works with defining the plugins like that and everything is fine.

I've created #5543 to make sure we change the code in the update to skip the core gems from the message, I think it will confuse people.

rake artifact:all, rake bootstrap, rake test:install-core, rake test:core command **works**.

LGTM

ph:
As a side note, this should make the unified build process faster since it skip the `logstash-plugin install`

jsvd:
until we decide on https://github.com/elastic/logstash/issues/5533 we should only merge this to 5.0, correct?
Thanks for the review @ph

ph:
Concerning #5533, I think we could still merge this in master or we can merge it later.

Merging it right will make things a tad slower until we implement a solution for it. So its a matter of how much pain we want to have :)

suyograo:
+1 on merging it to master as well as 5.0. See https://github.com/elastic/logstash/issues/5533#issuecomment-228106684

jsvd:
as for merging this to master, should we merge as is? I'm don't see the benefit of having master restrain the major version of the plugins.
Since it's "bleeding edge" it should pick whatever is latest, no?

suyograo:
@jsvd makes sense to leave master without constraints..

PR #5537, opened 2016-06-22
avoid using duplicate logstash.yml for rpm/deb

jsvd:
Currently https://github.com/elastic/logstash/blob/master/config/logstash.yml and https://github.com/elastic/logstash/blob/master/pkg/logstash.yml are the same, so until they start to diverge I'd suggest using the same file.


untergeek:
Hmmm.  Make sure this gets tested.  I had issues when I used the approach you went with, which is why I ended up "duplicating" it. Perhaps I just did something wrong.

untergeek:
And by tested, I mean by the package testing infrastructure.

jsvd:
Tested both RPM and DEB packages..all is fine. Reviews welcome.

untergeek:
LGTM™

Thanks for chasing down the weirdness.

Issue #5523, opened 2016-06-20
Understanding the performance of bin/logstash-plugin install command


purbon:
One of the common comments and concerns about logstash is it's performance, in this issue we try to understand the performance concerns involved in installing a plugin.

Here is a sample execution trying to install the generator input.

```
skywalker% JRUBY_OPTS="--profile" ./bin/logstash-plugin install logstash-input-generator
Profiling enabled; ^C shutdown will now dump profile info
Validating logstash-input-generator
Installing logstash-input-generator
Installation successful

main profile results:
Total time: 17.12

     total        self    children       calls  method
----------------------------------------------------------------
     15.18        0.00       15.18           1  Clamp::Command.run
     15.18        0.00       15.18           2  Clamp::Command#run
     15.18        0.00       15.18           1  Clamp::Subcommand::Execution.execute
     15.18        0.00       15.18           1  LogStash::PluginManager::Install#execute
     14.51        0.06       14.45       26035  Array#each
      9.39        0.00        9.39           1  LogStash::PluginManager::Install#install_gems_list!
      9.39        0.00        9.39           1  LogStash::Bundler.invoke!
      9.27        0.00        9.27        5093  Proc#call
      9.27        0.00        9.27           1  LogStash::Bundler.execute_bundler_with_retry
      9.27        0.00        9.27           1  LogStash::Bundler.capture_stdout
      9.27        0.00        9.27           1  Kernel.loop
      9.27        0.00        9.27           1  LogStash::Bundler.execute_bundler
      9.27        0.00        9.27           1  Bundler::CLI.start
      9.27        0.00        9.27           1  Bundler::Thor::Base::ClassMethods.start
      9.27        0.00        9.27           1  Bundler::Thor.dispatch
      9.26        0.00        9.26           1  Bundler::Thor::Invocation.invoke_command
      9.26        0.00        9.26           1  Bundler::Thor::Command#run
      9.26        0.00        9.26          73  BasicObject#__send__
      9.26        0.00        9.26           1  Bundler::CLI#install
      9.26        0.00        9.25           1  Bundler::CLI::Install#run
      9.15        0.00        9.15           1  Bundler::Installer.install
      9.15        0.00        9.15           1  Bundler::Installer#run
      8.29        0.00        8.29           1  Bundler::Installer#install_sequentially
      7.64        0.00        7.64          72  Bundler::SpecSet#each
      7.54        0.00        7.54          64  Bundler::Installer#install_gem_from_spec
[...]

PR #5538, opened 2016-06-22
Improve acceptance tests execution time

purbon:
Hi,
  after performing some benchmarking for the install command in #5523, in this PR it basically disable jar dependencies and the plugin validation, this two are the biggest offenders in terms of execution time so CLI plugin manager test basically get an speedup of nearly 50% per platform.

See  `13 minutes 48 seconds` without this change, to  `6 minutes 48 seconds`  with this change.

A proper solution to make this two flags faster should be researched, but for now the effort for it is quite important, so decided to push this two small changes here to make our test suites faster sooner.

### DIFF
--- a/qa/rspec/commands/base.rb
+++ b/qa/rspec/commands/base.rb
@@ -29,7 +29,7 @@ module ServiceTester
 
       response = nil
       at(hosts, {in: :serial}) do |_host|
-        response = sudo_exec!(cmd)
+        response = sudo_exec!("JARS_SKIP='true' #{cmd}")
       end
       response
     end
--- a/qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash-plugin/install.rb
@@ -32,10 +32,22 @@ shared_examples "logstash install" do |logstash|
         end
 
         context "when fetching a gem from rubygems" do
+
           it "successfully install the plugin" do
+            command = logstash.run_command_in_path("bin/logstash-plugin install logstash-filter-qatest")
+            expect(command).to install_successfully
+            expect(logstash).to have_installed?("logstash-filter-qatest")
+          end
+
+          it "successfully install the plugin when verification is disabled" do
             command = logstash.run_command_in_path("bin/logstash-plugin install --no-verify logstash-filter-qatest")
             expect(command).to install_successfully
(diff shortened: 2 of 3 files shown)


purbon:
@ph what do you think?

purbon:
@untergeek might be you can also give your feedback here.

jsvd:
is this not needed anywhere? what are the downsides of using --no-verify?

purbon:
@jsvd verify check the plugin is a logstash plugin by checking the rubygems api, but this is extremely slow to do it all the time.

jsvd:
Honestly this makes me a bit itchy, to disable safety checks and default behaviour for performance reasons, specially in an acceptance test scenario.
Certainly we should be testing --no-verify in a separate acceptance test, but since the default is enabled, all other tests should use it.

purbon:
@jsvd I understand your point, my proposal is to have only 1 test doing this verify and disable it for others as is extremely slow process. We can always re enable it when we land a better version that does not take this huge penalty.

what do you think?

purbon:
FYI:on the other side, here we are testing with plugins we all know are logstash plugins, so end of the days we are just checking if the call to rubygems works.

jsvd:
ok, I see, and agree with you. Can we then add:
  * for install and update commands
     * 1 test that shows, with default args, the command fails on a non logstash plugin gem
     * 1 test that shows, with default args, the command succeeds on a logstash plugin gem
     * 1 test that shows, with --no-verify, the command succeeds on a logstash plugin gem

With that in place I'm +1 on doing all other actions with --no-verify

purbon:
this makes sense and is a resonable addition to this PR. will push this
now. thanks Joao for taking time to review this.

purbon:
@jsvd what do you think now?

qa/rspec/commands/base.rb
```diff
@@ -29,7 +29,7 @@ def run_command(cmd, host)
 
       response = nil
       at(hosts, {in: :serial}) do |_host|
-        response = sudo_exec!(cmd)
+        response = sudo_exec!("JARS_SKIP='true' #{cmd}")
```
jsvd: does this contribute a lot to performance of the specs?

purbon: yes, a lot, see https://github.com/elastic/logstash/issues/5523 for more details.

jsvd:
LGTM

PR #5542, opened 2016-06-23
Fix an issue where metric instance was nil when running test from a plugin

ph:
This PR introduces a few changes:

- Make sure that a metric object is always available even if the plugin was not created from a pipeline see https://github.com/logstash-plugins/logstash-filter-geoip/pull/83#issuecomment-227852968 for more details
- When you run logstash in debug mode and an accepting occur we will display the stacktrace, this make debugging issues with the pipeline a bit easier to track.
- The metric object was not correctly setup for input plugin and would have make the metric collection fails for theses one.
- Flores is now setup in the Rspec.configure instead of a file basis.

When this PR is merged we need to release a new snapshot to make sure the plugin test correctly pass.

### DIFF
--- a/logstash-core/lib/logstash/plugin.rb
+++ b/logstash-core/lib/logstash/plugin.rb
@@ -105,9 +105,17 @@ class LogStash::Plugin
   end
 
   def metric
-    @metric_plugin ||= enable_metric ? @metric : LogStash::Instrument::NullMetric.new
+    @metric_plugin ||= begin
+                         # We can disable metric per plugin if we want in the configuration
+                         # we will use the NullMetric in this case.
+                         if @enable_metric
+                           # Fallback when testing plugin and no metric collector are correctly configured.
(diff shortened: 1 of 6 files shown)


ph:
I've marked this PR as blocked because it block people from implementing metric collecting in the plugin.

logstash-core/lib/logstash/plugin.rb
```diff
@@ -105,9 +105,17 @@ def metric=(new_metric)
   end
 
   def metric
-    @metric_plugin ||= enable_metric ? @metric : LogStash::Instrument::NullMetric.new
+    @metric_plugin ||= begin
```
jordansissel: why use `begin` here?

ph: Not need I will move it to a ternary operator

ph:
@jordansissel updated with your changes

Issue #5539, opened 2016-06-22
Backport null metrics to 2.4


suyograo:
We want to make sure 5.x plugins work on 2.4, see https://github.com/elastic/logstash/issues/5449. As we introduce metrics usages in plugins, we should backport a shim for metrics as well. This way plugins don't have to have a check -- `respond_to?(:metric)` or along those lines. The metric call should be a noop

ph:
done in #5544

PR #5544, opened 2016-06-23
Implement a NullMetric Object in the 2.X series of Logstash

ph:
This PR, backport the `NullMetric` class from the master branch into the
2.X series and allow to write plugin that can be run on both Logstash 5
and Logstash 2.X without changing the code and without having to check
if the metric object is set before capturing a metric.

Fixes: #5539


jordansissel:
Code LGTM. Tests are passing.

PR #5545, opened 2016-06-23
Add more tests around the metric classes

ph:
This PR improves the quality of the metric library:
- Add more tests coverage for the `Namespaced` class
- Add more tests coverage for the `NullMetric` class
- Fix an issue when `#increment` was called with more arguments than required on both the `NullMetric` and the `namespaced` class.

I've send this your way @jsvd since you already know a bit of this code.

### DIFF
--- a/logstash-core/spec/logstash/instrument/namespaced_metric_spec.rb
+++ b/logstash-core/spec/logstash/instrument/namespaced_metric_spec.rb
@@ -2,10 +2,11 @@
 require "logstash/instrument/namespaced_metric"
 require "logstash/instrument/metric"
 require_relative "../../support/matchers"
+require_relative "../../support/shared_examples"
 require "spec_helper"
 
 describe LogStash::Instrument::NamespacedMetric do
-  let(:namespace) { :stats }
+  let(:namespace) { :root }
   let(:collector) { [] }
   let(:metric) { LogStash::Instrument::Metric.new(collector) }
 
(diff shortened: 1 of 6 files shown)


logstash-core/spec/logstash/instrument/namespaced_metric_spec.rb
```diff
@@ -27,6 +27,90 @@
...
+      metric = subject.increment(:error_rate)
+      expect(collector).to be_a_metric_event([:root, :error_rate], :counter, :increment, 1)
+    end
+
+    it "a counter by a provided value" do
+      metric = subject.increment(:error_rate, 20)
+      expect(collector).to be_a_metric_event([:root, :error_rate], :counter, :increment, 20)
+    end
+
+    it "raises an exception if the key is an empty string" do
```
jsvd: Not strictly required, but consider adding the `raises an exception if the key is an empty string` and `raise an exception if the key is nil` specs to a shared example like "crud operations". these are repeated in increment/decrement/gauge and, though it is not tested, probably also affect `#time`.

logstash-core/spec/logstash/instrument/null_metric_spec.rb
```diff
@@ -8,7 +8,53 @@
     expect(described_class).to implement_interface_of(LogStash::Instrument::NamespacedMetric)
   end
 
+  let(:key) { "galaxy" }
+
+  describe "#increment" do
```
jsvd: same with these specs, most of the specs for the nullmetric are also valid for the Namespaced metric (the `.to_not raise_error`, namespacing resulting in an instance of self.class, etc)

ph: Correct I will make this a shared example and reuse them.

ph:
@jsvd I've followed your suggestion with the following work, I've created a shared example which is defined in the `namespaced_metric_spec` and reused in the `null_metric_spec` file. I've moved the nil and empty check for the keys in this shared example too.

jsvd:
I'm getting a warning about redefinition when running `rake bootstrap && rake test:install-core && rake test:core`:

```
rake test:install-core  123.52s user 4.83s system 259% cpu 49.450 total
WARNING: Shared example group 'metrics commons operations' has been previously defined at:
  /Users/joaoduarte/projects/logstash/logstash-core/spec/logstash/instrument/namespaced_metric_spec.rb:8
...and you are now defining it at:
  /Users/joaoduarte/projects/logstash/logstash-core/spec/logstash/instrument/namespaced_metric_spec.rb:8
The new definition will overwrite the original one.

Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}
...............................................Updating dummy

```

That said, the change looks good and the tests pass consistently.

ph:
Weird, I have paid attention to the warning but only to the result of the run.
Let me track it down, I might have to actually make the example outside of the rspec file since it might be related at how the spec files are actually loaded.

ph:
@jsvd Had no easy choice, I've moved them into a common file and both spec are using them, I didnt want to deal with class path since things get can messy really quick.

Gemfile
```diff
@@ -21,3 +21,15 @@ gem "gems", "~> 0.8.3", :group => :build
 gem "rack-test", :require => "rack/test", :group => :development
 gem "flores", "~> 0.0.6", :group => :development
 gem "pleaserun"
+gem "logstash-filter-clone"
```
jsvd: please remove these changes

jsvd:
Just remove the changes to the Gemfile and LGTM!

jsvd:
LGTMx2

PR #5554, opened 2016-06-25
do not remove (a possibly) created user/group on rpm uninstallation

liro82:
it would be better to not remove a (possibly) created user/group on uninstallation of logstash rpm.
let me explain why the removal is not a good option:
- user/group may not be created by rpm (see [here](https://fedoraproject.org/wiki/Packaging:UsersAndGroups) to see why)
- existing files on system (prior owned by logstash) will be possibly owned a later / newly created user (see [possibly security threa](https://ma.ttias.be/on-removing-users-with-postun-in-rpm-spec-files/))

within an organization managing multiple logstash server, sysadmins may want to have consistent uid/gid by either
- pre creating user/groups (rpm installation will correctly reuse these users)
- or using directory based users (freeipa, ...)

and thus uninstalling a package (at least for RedHat based systems) it is not the best way to remove a user.

thanks for having a few thoughts (and hopefully merge) about this situation.

### DIFF
--- a/pkg/centos/before-remove.sh
+++ b/pkg/centos/before-remove.sh
@@ -27,11 +27,4 @@ if [ $1 -eq 0 ]; then
       rm /etc/systemd/system/logstash.service
     fi
   fi
-  if getent passwd logstash >/dev/null ; then
-    userdel logstash
-  fi
-
-  if getent group logstash > /dev/null ; then
-    groupdel logstash
-  fi
 fi
--- a/pkg/debian/before-remove.sh
+++ b/pkg/debian/before-remove.sh
@@ -28,11 +28,4 @@ if [ $1 = "remove" ]; then
       rm /etc/systemd/system/logstash.service
     fi
   fi
-  if getent passwd logstash >/dev/null ; then
-    userdel logstash
-  fi
-
-  if getent group logstash > /dev/null ; then
-    groupdel logstash
-  fi
 fi
--- a/pkg/ubuntu/before-remove.sh
+++ b/pkg/ubuntu/before-remove.sh
@@ -28,11 +28,4 @@ if [ $1 = "remove" ]; then
       rm /etc/systemd/system/logstash.service
     fi
   fi
-  if getent passwd logstash >/dev/null ; then
-    userdel logstash
-  fi
-
-  if getent group logstash > /dev/null ; then
-    groupdel logstash
-  fi
 fi


untergeek:
Makes sense.  

For the sake of consistency, would you mind also doing the same for the [debian](https://github.com/elastic/logstash/blob/master/pkg/debian/before-remove.sh#L31-L37) and [ubuntu](https://github.com/elastic/logstash/blob/master/pkg/ubuntu/before-remove.sh#L31-L37) `before-remove.sh` scripts?  I'd be happy to merge the changes after that.

Thanks for contributing!

liro82:
@untergeek sure, just changed and pushed.

thank you!

untergeek:
LGTM™ Thanks!

untergeek:
Stand by.  On further review, we may want to add more logic to this.  If we create a user, we should also be able to delete it.  If we don't create a user, we shouldn't try to delete it.

jordansissel:
I want the package to clean up after itself, and if we cannot remove the user when the package is removed, then I wonder if we should even create the user in the first place?

As for your concerns, which are valid, I will try to suggest solutions:

> user/group may not be created by rpm (see here to see why)

In the after-install script, we can record whether or not the package needed to create a user, and if so, make sure we remove the user on package removal only if the package created the user - otherwise do not remove the user on package removal.

> existing files on system (prior owned by logstash) will be possibly owned a later / newly created user 

What files? Can you provide details? Why would files need to be owned by the logstash user?

One thing I can think of, for file ownership, is this: Logstash should only maybe have ownership (or write permission, anyway) on the logfiles it creates with the sysv init script (with upstart and systemd, we should let those systems receive the logging). Given this example, we could have the package removal script remove the log files upon package removal.

jordansissel:
Just for clarity, I agree with your sentiments that the current behavior of the package scripts is not perfect in all situations, and we can make steps to improve it.

As an out-of-the-box kind of proposal, what about this: What if we provide a package with no scripts at all, so you can just treat the package like a tarball (an rpm without scripts is then just an inert file archive like tar/zip, anyway).

liro82:
ok, not an easy decision...

I totally agree that a package should cleanup if being removed. what a package installs should be removed by a package (considering files). a package includes:

* (template / default) configuration files
* functional code (binary, script, ...)

a package may need a user/group to correctly function (if not existing yet, the package should create one). if one already exists the package should use it (so far the current mode of operation is perfect).

while using the software, logfiles and temporary files will be created (for logstash `/var/log/logstash/*`  and `/var/lib/logstash/*`) which are not removed on uninstall (which for logfile is also the correct handling for my opinion).

consider this use-case:
when installing logstash, it may create a user. afterwards you use logstash (which produces `/var/log/lostash/*` and `/var/lib/logsatsh/*` files). once you uninstall logstash, these created files will remain (and again at least for logfiles they should if you ask me). since the user/group is deleted, once you install logstash again a new user/group is created (but often with different uid/gid). this new installation will now need further manual changes in order to correctly work.

if the uninstallation would not remove the user a next installation will reuse the already created user. in addition logfiles will not be left without a local user on the system.

thats was the background, why starting this PR.

jordansissel:
How about this for a propose solution:

## On package installation:

* create the `logstash` user if it doesn't exist
* make a note if we created the user (somewhere) to be used at package removal-time

## On package removal

* Remove anything created by the after-installation script (/var/lib/logstash, etc)
* If installation created the `logstash` user, then we should remove it. Otherwise, we should not remove it.

andsel:
Closed because the spirit of this PR has been moved to a enhancement issue #11672

PR #5555, opened 2016-06-27
Remove JMS from the skip list of plugins

purbon:
The jms input and output were added to the skip list when they were in initial development, but got never removed from there. Now both gets full content repo and people using them, see https://discuss.elastic.co/t/why-logstash-output-jms-isnt-part-of-the-pluggins-in-the-official-documentation/53874/2 for more details.

This will bring back this plugins to the documentation and the all plugins bundle.

@jordansissel @markwalkom @suyograo

### DIFF
--- a/rakelib/default_plugins.rb
+++ b/rakelib/default_plugins.rb
@@ -116,7 +116,6 @@ module LogStash
 
     ALL_PLUGINS_SKIP_LIST = Regexp.union([
       /^logstash-filter-yaml$/,
-      /jms$/,
       /example$/,
       /drupal/i,
       /^logstash-output-logentries$/,


suyograo:
LGTM

andrewvc:
I dunno. The test coverage is [very low](https://github.com/logstash-plugins/logstash-output-jms/blob/master/spec/outputs/jms_spec.rb) and there isn't any active development on it.

Putting it on the list sends the message that we'd support it.

suyograo:
This list is not for default plugins we package. Not having it here will also skip the document generation which we don't want.

andrewvc:
Ohhh, nevermind then! I got confused.

suyograo:
IOW, having JMS plugin included here does not mean we support it. It just means we can install it while generating docs for all plugins in logstash-plugin github repo

PR #5560, opened 2016-06-28
Add queue client support (and batch object) to pipeline

guyboertje:
Following discussion, this is WrappedSynchronousQueue Client support classes proposal.

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -216,7 +216,7 @@ module LogStash; class Pipeline
         end
       end
     ensure
-      # it is important to garantee @ready to be true after the startup sequence has been completed
+      # it is important to guarantee @ready to be true after the startup sequence has been completed
       # to potentially unblock the shutdown method which may be waiting on @ready to proceed
       @ready.make_true
     end
@@ -227,17 +227,16 @@ module LogStash; class Pipeline
   def worker_loop(batch_size, batch_delay)
     running = true
 
-    @filter_queue_client.set_batch_details(batch_size, batch_delay)
+    @filter_queue_client.set_batch_dimensions(batch_size, batch_delay)
 
     while running
-      batch, signal = @filter_queue_client.take_batch
+      batch = @filter_queue_client.take_batch
       @events_consumed.increment(batch.size)
-      running = false if signal == LogStash::SHUTDOWN
+      running = false if batch.shutdown_signal_received?
       filter_batch(batch)
 
-      if signal # Flush on SHUTDOWN or FLUSH
-        flush_options = (signal == LogStash::SHUTDOWN) ? {:final => true} : {}
-        flush_filters_to_batch(batch, flush_options)
+      if batch.shutdown_signal_received? || batch.flush_signal_received?
+        flush_filters_to_batch(batch)
       end
 
       output_batch(batch)
@@ -253,7 +252,7 @@ module LogStash; class Pipeline
           if e.cancelled?
             batch.cancel(e)
           else
-            batch.add(e)
+            batch.merge(e)
           end
         end
       end
@@ -414,7 +413,7 @@ module LogStash; class Pipeline
   end
 
 
-  # perform filters flush and yeild flushed event to the passed block
+  # perform filters flush and yield flushed event to the passed block
   # @param options [Hash]
   # @option options [Boolean] :final => true to signal a final shutdown flush
   def flush_filters(options = {}, &block)
@@ -458,19 +457,22 @@ module LogStash; class Pipeline
   end
 
   # perform filters flush into the output queue
+  #
   # @param batch [ReadClient::ReadBatch]
   # @param options [Hash]
-  # @option options [Boolean] :final => true to signal a final shutdown flush
   def flush_filters_to_batch(batch, options = {})
+    options[:final] = batch.shutdown_signal_received?
     flush_filters(options) do |event|
-      unless event.cancelled?
+      if event.cancelled?
+        batch.cancel(event)
+      else
         @logger.debug? and @logger.debug("Pushing flushed events", :event => event)
-        batch.add(event)
+        batch.merge(event)
       end
     end
 
     @flushing.set(false)
-  end # flush_filters_to_output!
+  end # flush_filters_to_batch
 
   def plugin_threads_info
     input_threads = @input_threads.select {|t| t.alive? }
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -21,21 +21,21 @@ require "logstash/filter_delegator"
 
 module LogStash; class Pipeline
   attr_reader :inputs,
-      :filters,
-      :outputs,
-      :worker_threads,
-      :events_consumed,
-      :events_filtered,
-      :reporter,
-      :pipeline_id,
-      :logger,
-      :started_at,
-      :thread,
-      :config_str,
-      :settings,
-      :metric,
-      :filter_queue_client,
-      :input_queue_client
+    :filters,
+    :outputs,
+    :worker_threads,
+    :events_consumed,
+    :events_filtered,
+    :reporter,
+    :pipeline_id,
+    :logger,
+    :started_at,
+    :thread,
+    :config_str,
+    :settings,
+    :metric,
+    :filter_queue_client,
+    :input_queue_client
 
   MAX_INFLIGHT_WARN_THRESHOLD = 10_000
 
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -21,24 +21,24 @@ require "logstash/filter_delegator"
 
 module LogStash; class Pipeline
   attr_reader :inputs,
-    :filters,
-    :outputs,
-    :worker_threads,
-    :events_consumed,
-    :events_filtered,
-    :reporter,
-    :pipeline_id,
-    :logger,
-    :started_at,
-    :thread,
-    :config_str,
-    :settings,
-    :metric
+      :filters,
+      :outputs,
+      :worker_threads,
+      :events_consumed,
+      :events_filtered,
+      :reporter,
+      :pipeline_id,
+      :logger,
+      :started_at,
+      :thread,
+      :config_str,
+      :settings,
+      :metric
 
   MAX_INFLIGHT_WARN_THRESHOLD = 10_000
 
   RELOAD_INCOMPATIBLE_PLUGINS = [
-    "LogStash::Inputs::Stdin"
+      "LogStash::Inputs::Stdin"
   ]
 
   def initialize(config_str, settings = LogStash::SETTINGS, namespaced_metric = nil)
@@ -82,14 +82,13 @@ module LogStash; class Pipeline
       raise
     end
 
-    @input_queue = LogStash::Util::WrappedSynchronousQueue.new
+    queue = LogStash::Util::WrappedSynchronousQueue.new
+    @input_queue_client = queue.write_client
+    @filter_queue_client = queue.read_client
+
     @events_filtered = Concurrent::AtomicFixnum.new(0)
     @events_consumed = Concurrent::AtomicFixnum.new(0)
 
-    # We generally only want one thread at a time able to access pop/take/poll operations
-    # from this queue. We also depend on this to be able to block consumers while we snapshot
-    # in-flight buffers
-    @input_queue_pop_mutex = Mutex.new
     @input_threads = []
     # @ready requires thread safety since it is typically polled from outside the pipeline thread
     @ready = Concurrent::AtomicBoolean.new(false)
@@ -99,6 +98,10 @@ module LogStash; class Pipeline
 
   def ready?
     @ready.value
+  end
+
+  def filter_queue_client
+    @filter_queue_client
   end
 
   def safe_pipeline_worker_count
@@ -112,14 +115,14 @@ module LogStash; class Pipeline
     if @settings.set?("pipeline.workers")
       if pipeline_workers > 1
         @logger.warn("Warning: Manual override - there are filters that might not work with multiple worker threads",
-                     :worker_threads => pipeline_workers, :filters => plugins)
+            :worker_threads => pipeline_workers, :filters => plugins)
       end
     else
       # user did not specify a worker thread count
       # warn if the default is multiple
       if default > 1
         @logger.warn("Defaulting pipeline worker threads to 1 because there are some filters that might not work with multiple worker threads",
-                     :count_was => default, :filters => plugins)
+            :count_was => default, :filters => plugins)
         return 1 # can't allow the default value to propagate if there are unsafe filters
       end
     end
@@ -176,8 +179,10 @@ module LogStash; class Pipeline
   end
 
   def start_workers
-    @inflight_batches = {}
-
+    @filter_queue_client.set_events_metric(metric.namespace([:stats, :events]))
+    @filter_queue_client.set_pipeline_metric(
+        metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :events])
+    )
     @worker_threads.clear # In case we're restarting the pipeline
     begin
       start_inputs
@@ -187,19 +192,20 @@ module LogStash; class Pipeline
       pipeline_workers = safe_pipeline_worker_count
       batch_size = @settings.get("pipeline.batch.size")
       batch_delay = @settings.get("pipeline.batch.delay")
+
       max_inflight = batch_size * pipeline_workers
 
-      config_metric = metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :config])   
+      config_metric = metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :config])
       config_metric.gauge(:workers, pipeline_workers)
       config_metric.gauge(:batch_size, batch_size)
       config_metric.gauge(:batch_delay, batch_delay)
-      
+
       @logger.info("Starting pipeline",
-                   "id" => self.pipeline_id,
-                   "pipeline.workers" => pipeline_workers,
-                   "pipeline.batch.size" => batch_size,
-                   "pipeline.batch.delay" => batch_delay,
-                   "pipeline.max_inflight" => max_inflight)
+          "id" => self.pipeline_id,
+          "pipeline.workers" => pipeline_workers,
+          "pipeline.batch.size" => batch_size,
+          "pipeline.batch.delay" => batch_delay,
+          "pipeline.max_inflight" => max_inflight)
       if max_inflight > MAX_INFLIGHT_WARN_THRESHOLD
         @logger.warn "CAUTION: Recommended inflight events max exceeded! Logstash will run with up to #{max_inflight} events in memory in your current configuration. If your message sizes are large this may cause instability with the default heap size. Please consider setting a non-standard heap size, changing the batch size (currently #{batch_size}), or changing the number of pipeline workers (currently #{pipeline_workers})"
       end
@@ -222,73 +228,43 @@ module LogStash; class Pipeline
   def worker_loop(batch_size, batch_delay)
     running = true
 
-    namespace_events = metric.namespace([:stats, :events])
-    namespace_pipeline = metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :events])
+    @filter_queue_client.set_batch_details(batch_size, batch_delay)
 
     while running
-      # To understand the purpose behind this synchronize please read the body of take_batch
-      input_batch, signal = @input_queue_pop_mutex.synchronize { take_batch(batch_size, batch_delay) }
+      batch, signal = @filter_queue_client.take_batch
+      @events_consumed.increment(batch.size)
       running = false if signal == LogStash::SHUTDOWN
-
-      @events_consumed.increment(input_batch.size)
-      namespace_events.increment(:in, input_batch.size)
-      namespace_pipeline.increment(:in, input_batch.size)
-
-      filtered_batch = filter_batch(input_batch)
+      filter_batch(batch)
 
       if signal # Flush on SHUTDOWN or FLUSH
         flush_options = (signal == LogStash::SHUTDOWN) ? {:final => true} : {}
-        flush_filters_to_batch(filtered_batch, flush_options)
+        flush_filters_to_batch(batch, flush_options)
       end
 
-      @events_filtered.increment(filtered_batch.size)
+      output_batch(batch)
 
-      namespace_events.increment(:filtered, filtered_batch.size)
-      namespace_pipeline.increment(:filtered, filtered_batch.size)
+      # @filter_queue_client.ack(batch)
+      # @dlq_client.add(batch)
 
-      output_batch(filtered_batch)
-
-      namespace_events.increment(:out, filtered_batch.size)
-      namespace_pipeline.increment(:out, filtered_batch.size)
-
-      inflight_batches_synchronize { set_current_thread_inflight_batch(nil) }
+      @filter_queue_client.close_batch(batch)
     end
-  end
-
-  def take_batch(batch_size, batch_delay)
-    batch = []
-    # Since this is externally synchronized in `worker_look` wec can guarantee that the visibility of an insight batch
-    # guaranteed to be a full batch not a partial batch
-    set_current_thread_inflight_batch(batch)
-
-    signal = false
-    batch_size.times do |t|
-      event = (t == 0) ? @input_queue.take : @input_queue.poll(batch_delay)
-
-      if event.nil?
-        next
-      elsif event == LogStash::SHUTDOWN || event == LogStash::FLUSH
-        # We MUST break here. If a batch consumes two SHUTDOWN events
-        # then another worker may have its SHUTDOWN 'stolen', thus blocking
-        # the pipeline. We should stop doing work after flush as well.
-        signal = event
-        break
-      else
-        batch << event
-      end
-    end
-
-    [batch, signal]
   end
 
   def filter_batch(batch)
-    batch.reduce([]) do |acc,e|
-      if e.is_a?(LogStash::Event)
-        filtered = filter_func(e)
-        filtered.each {|fe| acc << fe unless fe.cancelled?}
+    batch.each do |event|
+      if event.is_a?(LogStash::Event)
+        filter_func(event).each do |e|
+          #these are both dequeued and generated events
+          if e.cancelled?
+            batch.cancel(e)
+          else
+            batch.add(e)
+          end
+        end
       end
-      acc
     end
+    @filter_queue_client.add_filtered_metrics(batch)
+    @events_filtered.increment(batch.size)
   rescue Exception => e
     # Plugins authors should manage their own exceptions in the plugin code
     # but if an exception is raised up to the worker thread they are considered
@@ -297,38 +273,28 @@ module LogStash; class Pipeline
     # Users need to check their configuration or see if there is a bug in the
     # plugin.
     @logger.error("Exception in pipelineworker, the pipeline stopped processing new events, please check your filter configuration and restart Logstash.",
-                  "exception" => e, "backtrace" => e.backtrace)
+        "exception" => e, "backtrace" => e.backtrace)
     raise
   end
 
   # Take an array of events and send them to the correct output
   def output_batch(batch)
     # Build a mapping of { output_plugin => [events...]}
-    outputs_events = batch.reduce(Hash.new { |h, k| h[k] = [] }) do |acc, event|
+    output_events_map = Hash.new { |h, k| h[k] = [] }
+    batch.each do |event|
       # We ask the AST to tell us which outputs to send each event to
       # Then, we stick it in the correct bin
 
       # output_func should never return anything other than an Array but we have lots of legacy specs
       # that monkeypatch it and return nil. We can deprecate  "|| []" after fixing these specs
-      outputs_for_event = output_func(event) || []
-
-      outputs_for_event.each { |output| acc[output] << event }
-      acc
+      (output_func(event) || []).each do |output|
+        output_events_map[output].push(event)
+      end
     end
-
     # Now that we have our output to event mapping we can just invoke each output
     # once with its list of events
-    outputs_events.each { |output, events| output.multi_receive(events) }
-  end
-
-  def set_current_thread_inflight_batch(batch)
-    @inflight_batches[Thread.current] = batch
-  end
-
-  def inflight_batches_synchronize
-    @input_queue_pop_mutex.synchronize do
-      yield(@inflight_batches)
-    end
+    output_events_map.each { |output, events| output.multi_receive(events) }
+    @filter_queue_client.add_output_metrics(batch)
   end
 
   def wait_inputs
@@ -348,6 +314,7 @@ module LogStash; class Pipeline
 
     @inputs.each do |input|
       input.register
+
       start_input(input)
     end
   end
@@ -359,24 +326,24 @@ module LogStash; class Pipeline
   def inputworker(plugin)
     LogStash::Util::set_thread_name("[#{pipeline_id}]<#{plugin.class.config_name}")
     begin
-      plugin.run(@input_queue)
+      plugin.run(@input_queue_client)
     rescue => e
       if plugin.stop?
         @logger.debug("Input plugin raised exception during shutdown, ignoring it.",
-                      :plugin => plugin.class.config_name, :exception => e,
-                      :backtrace => e.backtrace)
+            :plugin => plugin.class.config_name, :exception => e,
+            :backtrace => e.backtrace)
         return
       end
 
       # otherwise, report error and restart
       if @logger.debug?
         @logger.error(I18n.t("logstash.pipeline.worker-error-debug",
-                             :plugin => plugin.inspect, :error => e.to_s,
-                             :exception => e.class,
-                             :stacktrace => e.backtrace.join("\n")))
+            :plugin => plugin.inspect, :error => e.to_s,
+            :exception => e.class,
+            :stacktrace => e.backtrace.join("\n")))
       else
         @logger.error(I18n.t("logstash.pipeline.worker-error",
-                             :plugin => plugin.inspect, :error => e))
+            :plugin => plugin.inspect, :error => e))
       end
 
       # Assuming the failure that caused this exception is transient,
@@ -413,7 +380,7 @@ module LogStash; class Pipeline
     # Each worker thread will receive this exactly once!
     @worker_threads.each do |t|
       @logger.debug("Pushing shutdown", :thread => t.inspect)
-      @input_queue.push(LogStash::SHUTDOWN)
+      @input_queue_client.push(LogStash::SHUTDOWN)
     end
 
     @worker_threads.each do |t|
@@ -483,7 +450,7 @@ module LogStash; class Pipeline
   def flush
     if @flushing.compare_and_set(false, true)
       @logger.debug? && @logger.debug("Pushing flush onto pipeline")
-      @input_queue.push(LogStash::FLUSH)
+      @input_queue_client.push(LogStash::FLUSH)
     end
   end
 
@@ -501,9 +468,10 @@ module LogStash; class Pipeline
   # @option options [Boolean] :final => true to signal a final shutdown flush
   def flush_filters_to_batch(batch, options = {})
     flush_filters(options) do |event|
+      event.set("flushed", true)
       unless event.cancelled?
         @logger.debug? and @logger.debug("Pushing flushed events", :event => event)
-        batch << event
+        batch.add(event)
       end
     end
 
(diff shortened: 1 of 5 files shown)


talevy:
why the sudden re-indenting?

guyboertje:
@talevy - only because I edited it in IntelliJ - this is the first time I edited ruby files with IntelliJ. I guess it applied some defaults that I was not aware of.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -21,24 +21,24 @@
...
+      :thread,
+      :config_str,
+      :settings,
+      :metric
 
   MAX_INFLIGHT_WARN_THRESHOLD = 10_000
 
   RELOAD_INCOMPATIBLE_PLUGINS = [
-    "LogStash::Inputs::Stdin"
+      "LogStash::Inputs::Stdin"
```
andrewvc: Why did all this whitespace change from 2 -> 4 spaces?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -82,14 +82,13 @@ def initialize(config_str, settings = LogStash::SETTINGS, namespaced_metric = ni
       raise
     end
 
-    @input_queue = LogStash::Util::WrappedSynchronousQueue.new
+    queue = LogStash::Util::WrappedSynchronousQueue.new
+    @input_queue_client = queue.write_client
+    @filter_queue_client = queue.read_client
```
andrewvc: +1 on these as separate objects. Eventually it'd be nice to have a separate write_client per-input. Is this where you're going with this?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -101,6 +100,10 @@ def ready?
     @ready.value
   end
 
+  def filter_queue_client
+    @filter_queue_client
```
andrewvc: This probably should be an `attr_reader`

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -112,14 +115,14 @@ def safe_pipeline_worker_count
     if @settings.set?("pipeline.workers")
       if pipeline_workers > 1
         @logger.warn("Warning: Manual override - there are filters that might not work with multiple worker threads",
-                     :worker_threads => pipeline_workers, :filters => plugins)
+            :worker_threads => pipeline_workers, :filters => plugins)
```
andrewvc: unnecessary whitespace change

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -176,8 +179,10 @@ def stopped?
   end
 
   def start_workers
-    @inflight_batches = {}
-
+    @filter_queue_client.set_events_metric(metric.namespace([:stats, :events]))
```
andrewvc: These methods should probably be `@filter_queue_client.events_metric = ...` to be more ruby-like.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -187,19 +192,20 @@ def start_workers
...
-                   "id" => self.pipeline_id,
-                   "pipeline.workers" => pipeline_workers,
-                   "pipeline.batch.size" => batch_size,
-                   "pipeline.batch.delay" => batch_delay,
-                   "pipeline.max_inflight" => max_inflight)
+          "id" => self.pipeline_id,
+          "pipeline.workers" => pipeline_workers,
+          "pipeline.batch.size" => batch_size,
+          "pipeline.batch.delay" => batch_delay,
+          "pipeline.max_inflight" => max_inflight)
```
andrewvc: whitespace.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -222,73 +228,43 @@ def start_workers
...
+      batch, signal = @filter_queue_client.take_batch
+      @events_consumed.increment(batch.size)
       running = false if signal == LogStash::SHUTDOWN
-
-      @events_consumed.increment(input_batch.size)
-      namespace_events.increment(:in, input_batch.size)
-      namespace_pipeline.increment(:in, input_batch.size)
-
-      filtered_batch = filter_batch(input_batch)
+      filter_batch(batch)
```
andrewvc: I like this cleanup!

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -222,73 +228,43 @@ def start_workers
...
+      output_batch(batch)
 
-      namespace_events.increment(:out, filtered_batch.size)
-      namespace_pipeline.increment(:out, filtered_batch.size)
-
-      inflight_batches_synchronize { set_current_thread_inflight_batch(nil) }
-    end
-  end
+      # @filter_queue_client.ack(batch)
+      # @dlq_client.add(batch)
```
andrewvc: Maybe a bit premature to put this stuff in here?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -222,73 +228,43 @@ def start_workers
...
-        filtered = filter_func(e)
-        filtered.each {|fe| acc << fe unless fe.cancelled?}
+    batch.each do |event|
+      if event.is_a?(LogStash::Event)
+        filter_func(event).each do |e|
+          #these are both dequeued and generated events
+          if e.cancelled?
+            batch.cancel(e)
+          else
+            batch.add(e)
```
andrewvc: Will this duplicate events? It will get back references to the events that are already in it in addition to the new events no?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -297,38 +273,28 @@ def filter_batch(batch)
     # Users need to check their configuration or see if there is a bug in the
     # plugin.
     @logger.error("Exception in pipelineworker, the pipeline stopped processing new events, please check your filter configuration and restart Logstash.",
-                  "exception" => e, "backtrace" => e.backtrace)
+        "exception" => e, "backtrace" => e.backtrace)
```
andrewvc: Whitespace

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -37,5 +37,231 @@ def take
...
+        @inflight_batches[Thread.current] = batch
+      end
+
+      def close_batch(batch)
+        @mutex.synchronize do
+          @inflight_batches.delete(Thread.current)
+        end
+      end
+
+      def add_dequeued_metrics(batch)
```
andrewvc: I'm not sure these metrics functions belong in the queue object here. They're really a separate concern. 

Mixing them in with the pipeline as before was messy, but at least was one less level of indirection. Have you considered breaking these functions out into a separate class? 

I think it makes sense to use something like the observer pattern here (or the ruby FP equivalent) and have the pipeline invoke callbacks. A `PipelineMetrics` class could hook into that.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -37,5 +37,231 @@ def take
...
+
+    def write_client
+      WriteClient.new(self)
+    end
+
+    def read_client
+      ReadClient.new(self)
+    end
+
+    class ReadClient
```
andrewvc: I like this abstraction

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -560,7 +561,7 @@ class TestPipeline < LogStash::Pipeline
 
     let(:pipeline_settings) { { "pipeline.id" => pipeline_id } }
     let(:pipeline_id) { "main" }
-    let(:number_of_events) { 1000 }
+    let(:number_of_events) { 20 } #1000
```
andrewvc: Why this change + comment?

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -604,7 +605,7 @@ class TestPipeline < LogStash::Pipeline
 
       Thread.new { subject.run }
       # make sure we have received all the generated events
-      sleep 1 while dummyoutput.events.size < number_of_events
+      sleep 0.25 until dummyoutput.events.size == number_of_events
```
andrewvc: Using exact equality is a less defensive test than `<`. If we somehow output more events than expected due to a bug we want this to terminate not go on infinitely, so that it can fail.

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
```diff
@@ -3,26 +3,65 @@
...
+    end
+
+    class Que < Array
+      def take() shift(); end
+      def poll(*) shift(); end
+    end
+
+    describe "WriteClient | ReadClient" do
+      context "when writing to the queue" do
+        let(:queue) { Que.new }
```
andrewvc: Typo?

logstash-core/lib/logstash/pipeline.rb
andrewvc: In generally I think we should prefer to avoid mutation. The batch seems to have multiple concerns

1. Representing a transaction around events removed from the queue
2. Representing the original events removed from the queue
3. Storing new events and tracking them as they mutate through the pipeline.

What's the argument behind putting them all in a single batch class? 

An alternate approach might be:

1. Creating a queue transaction object to track items removed from the queue. 
2. Creating a `#read` method off that transaction object that will pull a list of events.
3. Creating a `#close` method on that transaction object that will clear those items off the disk
4. Handling new events as simple arrays throughout the pipeline

The separate transaction object would be really powerful when applied to a future DLQ as well.

WDYT

logstash-core/lib/logstash/pipeline.rb
guyboertje: https://github.com/elastic/logstash/pull/5560#issuecomment-229272836

logstash-core/lib/logstash/pipeline.rb
guyboertje: IntelliJ Ruby support has continuation indent of 4 spaces. I think this is acceptable.

logstash-core/lib/logstash/pipeline.rb
guyboertje: The client is an adapter that presents one API to the each side of any given queue implementation. There also may be different concerns regarding batching and thread-safety guarantees between read and write that may complicate a dual purpose client. Further, separate clients guarantee that an input can't read an event from the queue and a worker can't write an event to the queue - which suits our current requirements.

The DLQ client, for example, should only be able to write events out.

It is possible to have per input client instances and each instance may post metrics on that input's interaction with the queue without the input needing to know or do this.

But given that a batch is isolated to a worker thread then that thread's interaction with the queue to retrieve a batch can post metrics on a per thread per batch basis.

In this PR the batch metrics are thread anonymous - that can be a future enhancement.

logstash-core/lib/logstash/pipeline.rb
guyboertje: It could be but I prefer to leave it as-is. I did originally call this method `filter_queue` but the ivar `@filter_queue_client` but I though it would be more confusing.

logstash-core/lib/logstash/pipeline.rb
guyboertje: I prefer to use the set_* idiom, it allows us to change the arguments e.g. `set_events_metric(metric, :stats)` that infers namespace([:stats, :events]).

logstash-core/lib/logstash/pipeline.rb
guyboertje: Its an illustration. I will remove it before merge.

logstash-core/lib/logstash/pipeline.rb
andrewvc: @guyboertje this isn't anywhere else in our codebase. We don't have an official style guide but I think this is a blocker IMHO.

andrewvc: @guyboertje you can easily reconfigure intellij not to do this FYI.

logstash-core/lib/logstash/pipeline.rb
andrewvc: Well, why not change it to `set_*` if we change the signature. I don't see why we should write non-idiomatic ruby because of a (very unlikely IMHO) future possibility.

andrewvc:
@guyboertje this is mostly good but I can't sign off on this with auto-ident changes from your editor. Not only do the changes not match up to Logstash conventions, but you've reindented lines that have nothing to do with this PR.

suyograo:
+1 on fixing the whitespace here.

logstash-core/lib/logstash/pipeline.rb
guyboertje: This does not duplicate. See ReadBatch # add method - the event is deleted from the `@dequeued` hash and added to the `@filtered` hash. If it is a new event the dequeued.delete will not do anything.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
guyboertje: As we discussed in the meeting and recorded here for non-attendees.
Metrics like logging is a cross-cutting concern. Taking metrics measurements can and should occur anywhere.

It could be argued that the pipeline should take the metrics at certain points in the batch lifecycle - but I found that it made the pipeline logic much harder to see.
Also, as the client and batch implementation go together, this metric `@event_metric.increment(:in, batch.dequeued_size)` if done in the pipeline, the pipeline would need to know batch specific internals. The PQ Batch object might not have a dequeued size.

logstash-core/lib/logstash/pipeline.rb
guyboertje: Further to using the `set_*` idiom, using a 'setter' allows one to control the return value. For these kind of post initialize setup, I might decide to return `self` to enable setter chaining if I see a lot of post initialize setup in successive lines.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -37,5 +37,231 @@ def take
...
+      # We generally only want one thread at a time able to access pop/take/poll operations
+      # from this queue. We also depend on this to be able to block consumers while we snapshot
+      # in-flight buffers
+
+      def initialize(queue)
+        @queue = queue
+        @mutex = Mutex.new
+        @inflight_batches = {}
```
guyboertje: Note that @infilght_batches as a central mechanism for tracking inflight abtches will fail if we have multiple read clients in the pipeline.

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
guyboertje: No its intentional. I did not want people confuse it with the Stdlib Queue class or allow the test to still run but with different/wrong behaviour if the `class Que < Array` block is accidentally removed.
Que to me seemed to imply that its a queue but not a Queue.

andrewvc: It might be easier on readers to make it more verbose. `DummyQueue` would be more descriptive be more aligned with other tests the logstash codebase.

colinsurprenant:
I must say that I am super happy that the biggest issue seems to be related to whitespace/indentation 😆  - that being said, I fully support the strict & coherent whitespace/indentation mafia here 😈

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
suyograo: Was reading this PR, and saw this. Sorry had to comment.

+1 on naming it something more meaningful. @guyboertje only you know why this is `Que` and not `Queue`. I honestly thought it was a typo too.

logstash-core/spec/logstash/pipeline_spec.rb
guyboertje: my bad - need to revert

logstash-core/spec/logstash/pipeline_spec.rb
guyboertje: you are correct the sampling my well miss the point at which the two equate - I should not assume that events.size will never be greater than number_of_events.

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: so  I read that to @andrewvc 's question the answer is "yes", we could have per-input write clients. right?

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: I agree that this should be an `attr_reader` ... this is exactly the purpose of attr_reader.

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: yeah, I often struggle on choosing between a `set_something(...)` method versus `something = ...` - in addition to the return value, there is also the multiple parameters issue.  In general I must say that I tend to prefer the `set_*` naming as I find it is more immediately obvious we are calling a setter and not assigning a local variable, for instance.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -222,73 +228,43 @@ def start_workers
...
   def filter_batch(batch)
-    batch.reduce([]) do |acc,e|
-      if e.is_a?(LogStash::Event)
-        filtered = filter_func(e)
-        filtered.each {|fe| acc << fe unless fe.cancelled?}
+    batch.each do |event|
+      if event.is_a?(LogStash::Event)
+        filter_func(event).each do |e|
+          #these are both dequeued and generated events
+          if e.cancelled?
```
colinsurprenant: isn't it silly that `filter_func` returns cancelled events?

andrewvc: +1. 

IMHO it's bizarre that the filter plugin contract is what it is. It should just transform an array into another array, and assume anything not in the output array is cancelled.

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: I suggest we rename `add` as it is in fact confusing when looking at the code. maybe `filter` or `set_filtered` ?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -297,38 +273,28 @@ def filter_batch(batch)
...
 
       # output_func should never return anything other than an Array but we have lots of legacy specs
       # that monkeypatch it and return nil. We can deprecate  "|| []" after fixing these specs
-      outputs_for_event = output_func(event) || []
-
-      outputs_for_event.each { |output| acc[output] << event }
-      acc
+      (output_func(event) || []).each do |output|
+        output_events_map[output].push(event)
+      end
```
colinsurprenant: 👍

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -499,9 +466,10 @@ def uptime
   # @option options [Boolean] :final => true to signal a final shutdown flush
   def flush_filters_to_batch(batch, options = {})
     flush_filters(options) do |event|
+      event.set("flushed", true)
```
colinsurprenant: was this added for debugging?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -499,9 +466,10 @@ def uptime
   # @option options [Boolean] :final => true to signal a final shutdown flush
   def flush_filters_to_batch(batch, options = {})
```
colinsurprenant: can we also add the api doc for the batch @param ?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
colinsurprenant: ok, good point. I would add a comment to that effect here somewhere and also in the Pipeline where the `read_client` is called so that if someone figures we should have multiple read clients this current limitation will be obvious

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -37,5 +37,231 @@ def take
...
+        @batch_size = 125
+        @wait_for = 5
+      end
+
+      def set_batch_details(batch_size, wait_for)
```
colinsurprenant: I find it a bit weird that this is not set in the constructor. this is not something that will or should change in the life of that object? I think it would make more sense to bubble up these settings from the `start_worker` to the Pipeline constructor where the `ReadClient` is instantiated.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -37,5 +37,231 @@ def take
...
+
+      def inflight_batches
+        @mutex.synchronize do
+          yield(@inflight_batches)
+        end
+      end
+
+      def current_inflight_batch
+        @inflight_batches.fetch(Thread.current, [])
+      end
```
colinsurprenant: I don't see `current_inflight_batch` used anywhere? what is the purpose?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
colinsurprenant: I agree an observer pattern would be perfect but this should be part of another issue/pr. 

Since both `add_filtered_metrics` and `add_output_metrics` are called from the pipeline, I agree with @andrewvc and for the sake of consistency I think `add_dequeued_metrics` should also be called at the pipeline, right after the `take_batch`.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -37,5 +37,231 @@ def take
...
+
+      def dequeue(event)
+        return if event.nil?
+        @dequeued[event] = 0
+      end
+
+      def add(event)
+        @dequeued.delete(event)
+        @filtered[event] = 0
+      end
```
colinsurprenant: as discussed above, I'd suggest we rename to `filter` or `set_filtered`

colinsurprenant:
Is the only purpose of the per-thread inflight batches accounting only for the purpose of the pipeline reporter? I find it introduces lots of weird complexity in the code. It's not the scope of this PR to sort this out but I'd vote we revisit this at some point.

logstash-core/lib/logstash/pipeline.rb
guyboertje: yes

logstash-core/lib/logstash/pipeline.rb
guyboertje: Referee "foul" two against one, not fair :-)

OK I will do it.

logstash-core/lib/logstash/pipeline.rb
guyboertje: The Batch Object API should not be dictated by the current usage of event arrays. The add method accommodates the current 'misuse' of it.

If possible, the `filter_func` should be made to accept the batch and the individual plugin wrapper (filter_delegator and output_delegator) should interact with the batch in accordance with its API.

```
General Batch Public API
.new()
 
#dequeue(event)
Used by the Read Client to put events into the batch, ideally I would
like to protect this method so that Queue Clients can only use it.

#add(event)
Generated events should be appended to the Batch using this method.

#cancel(event)
Use this method to cancel an event and exclude it from further iteration

#fail(event)
Use this method to indicate that a plugin failed while trying to update this event

#each(&block)
Use this method with a block to iterate over the current active set of events

#size()
This method returns the total number of events that this batch is carrying.

[Possible enhancements for different queue implementations]

#select(predicate, &block)
#reject(predicate, &block)

#active_size()
#cancelled_size()
#failed_size()

#filter(predicate)
```

colinsurprenant:
One other thing I'd like to see is the performance impact, if any. Maybe @purbon can help with this?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
guyboertje: I anticipated that, in the near future we may change either of these parameters during a run.
Doing this in a setter would not interrupt the thread locking - it would just allow the next batch to be sourced differently.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
guyboertje: Given that the pipeline used to `set_current_inflight_batch` so setting K,V for `@inflight_batches` hash and this hash is visible to the whole pipeline - if I move the `@inflight_batches` hash to the client, I need a way for the pipeline code to query the client for it. I could not find any code that accesses this hash ivar, though.

suyograo:
For performance testing, you can just use the perf tool which runs various configs against test data. See instructions here: https://github.com/elastic/logstash-performance-testing

logstash-core/lib/logstash/pipeline.rb
guyboertje: yes

guyboertje:
All the whitespace changes have been reverted now.
I believe this is in a merge-able state now.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -21,19 +21,21 @@
...
-    :thread,
-    :config_str,
-    :settings,
-    :metric
+      :filters,
+      :outputs,
+      :worker_threads,
+      :events_consumed,
+      :events_filtered,
+      :reporter,
```
andrewvc: The whitespace in this section is still off

logstash-core/lib/logstash/pipeline_reporter.rb
```diff
@@ -52,19 +52,20 @@ def snapshot
...
-        :inflight_count => inflight_count,
-        :worker_states => worker_states_snap,
-        :output_info => output_info,
-        :thread_info => pipeline.plugin_threads_info,
-        :stalling_threads_info => pipeline.stalling_threads_info
+          :events_filtered => events_filtered,
+          :events_consumed => events_consumed,
+          :worker_count => pipeline.worker_threads.size,
+          :inflight_count => inflight_count,
+          :worker_states => worker_states_snap,
```
andrewvc: Whitespace here still needs to be fixed.

logstash-core/lib/logstash/pipeline_reporter.rb
```diff
@@ -102,13 +103,13 @@ def output_info
...
       {
-        :type => output_delegator.config_name,
-        :config => output_delegator.config,
-        :is_multi_worker => is_multi_worker,
-        :events_received => output_delegator.events_received,
-        :workers => output_delegator.workers,
-        :busy_workers => output_delegator.busy_workers
+          :type => output_delegator.config_name,
+          :config => output_delegator.config,
+          :is_multi_worker => is_multi_worker,
```
andrewvc: whitespace issues

andrewvc:
@guyboertje there are still whitespace issues, I've annotated all the ones I've found.

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: also `#dequeue(event)` is really confusing  because it does not do what the method name generally means: to remove an element from a queue, but adds the event into a `dequeued` hash. 

I maintain that `add` and `dequeue` are very confusing method names - it may seems clear for you now but forward looking for someone reading the pipeline code in the future willl have a very hard time understanding this because we reuse standard method names with very typical meaning and do something completely counter intuitive.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
colinsurprenant: I really don't like all this in-flight batch stuff in the pipeline.

colinsurprenant:
My last *blocking* concern is about the `dequeue` and `add` method names, see comments.

My other issue that should be addressed in another issue is about the in-flight batches tracking - I think this should be refactored to not clutter as much the pipeline code.

suyograo:
Ok, lets create an issue to address metrics for in-flight batches.

guyboertje:
I can refactor away the dequeue method by allowing the constructor to take the events.

We could use `upsert(event)`, `update_cancelled(event)`, `update_failed(event)` - this retains the DB table metaphor from the discussion, as will the use of `select` to work with a subset.

Maybe one universal method `update(:generated, event)`, `update(:cancelled, event)` , `update(:failed, event)` this has the benefit of being self-documenting but also allows the usage to evolve without changing the API.

colinsurprenant:
I definitely would prefer having discrete methods instead of a single `update(:xxx, event)`. AFAICT this API can change since the plugins will still use the queue-like push method?

guyboertje:
@colinsurprenant - We want the callers to change to use the Batch API as we design it, not have the API reflect current calling intent.

colinsurprenant:
@guyboertje ok, that's fine. I still would prefer discrete method names. Otherwise we should just rename all out API methods as `api_method(:action, arguments)` where `:action` defines the methods behaviour and `arguments` is a list of arguments which applies to this action.

guyboertje:
@colinsurprenant:
- `dequeue`, I have removed it.
- `add`, I have listed my reasons for naming it so. You are uncomfortable with that name. I have explained the criteria for the naming of the the API. The names that you suggest to not meet these criteria. I have suggested alternatives that fit with a DB table metaphor - so far you have said that a single method with a 'command' modifier is unacceptable (and I agree with you on that one, it was a counterpoint folly) but you have not OK'ed the DB metaphor name or suggested a name within the criteria.

To recap in the form of an email - the narative around `add` and the Batch API is:

> Dear Queue,
> 
> Thanks ever so much for the batch - it is worth the wait.
> 
> I am going to give each event to a function (I don't know much about what this function does - it looks a bit funky to me) and its going to call me back with an event. It may call me back zero to plenty of times.
> I will check to see if this called back event is cancelled, so I can get you to cancel it.
> I could also check to see whether this called back event is different to the one I gave to the function and then I could get you to add it to the batch.
> However, I would prefer it if you did the 'different' check because I only know about the event I am currently iterating over and you know about all of them, so when I get you to add this event to the batch will you please do the add/ignore on my behalf.
> 
> By the way, I am planning to speak to the parents of this function to get them to teach it to only call me back when it is cancelling an existing event or it is adding some generated events to  the batch - having said that, you should probably keep checking whether it really is an additional event because, you may not know this, this function is dynamically generated and they just make it up as they go along.
> 
> Yours sincerely,
> Pipeline

Also, as a group we discussed that - when an impasse is reached on matters of style/preference/opinion the PR submission is given the benefit of the doubt.

jsvd:
I see some confusion in explaining in one sentence what the `add(event)` api does from the outside perspective (the implementation is irrelevant):

```
#add(event)
Generated events should be appended to the Batch using this method.
```

The sentence above isn't clear IMO...if it was "add the event to the batch", then `add` would be a good name. Is it not this? something else? (no I don't want to read the code)

guyboertje:
@jsvd for me the word`append` is an add with a positional directive baked in - add to the end.

colinsurprenant:
@guy - the fact that a dissertation is required to explain the `add` behaviour speaks for itself IMO.  If we go back to the original comment by @andrewvc when he asked «Will this duplicate events?» this comment was absolutely valid and is still is. 

Currently, someone coming in the pipeline code and looking at this logic **will not understand** what is going on. 

I agree that the original sin is with the `filter_func` returning mutated and new events and **ideally** this is where we should solve that problem but if we don't want to tackle that problem in this PR, I still maintain that the `add` method name **is confusing**. 

At this point I really don't care if we name it `dear_batch_please_verify_this_event_and_if_it_is_a_new_event_add_it_to_the_filtered_list` ... (and not, I don't like the DB metaphor but this is a personal preference and will totally live with it if the name is not confusing (but please not *upsert*))

As a side note, I believe that the current `add` implementation is not right since we are loosing sight of the original dequeued list and putting both dequeued and generated events in the filtered list. At the moment this does not have any impact but I believe a better implementation would be to keep separate track of dequeued & mutated events and newly generated events.

jsvd:
@guy I merely copied a portion of your Batch API, I'm not suggesting append instead of add.

Going along Colin's argument, I think there's value in thinking if non-converging discussion is revealing underlying complexity in the implementation (this is coming from someone who's not familiar with the code).

guyboertje:
@colinsurprenant 
The dissertation explains the add/append/i_think_this_belongs_to_you behaviour - this is true - but this is because this is a PR and the behaviour is established in this PR. It is the correct place for full explanations.

My answer to Andrew's question (about duplicates) should have been - outwardly from the callers point of view "its not your problem". But Andrew's question was completely subjective and based on his prior work unlike a third party looking at the pipeline code afresh, I reckon they would say 'ahh I use `add` to add an event to the batch, that's simple' and trust that the batch internals did the correct thing.

Imagine we have a new discussion about the clean brand new design of say, a Train object and its API. If I presented this:
```
Train(carriages)
  move(from, to)
  add(carriage)
  remove(carriage)
```
would you really complain about the method name `add` without previous implementation details to add subjectivity?

guyboertje:
@jsvd - I know you were not suggesting append as a method name - I was merely saying that in my description the word append is a special case of add.

colinsurprenant:
if "every one" expresses concerns about the clarity of a chosen method name, either you stick with it because you think everyone got it wrong or you take that opportunity to introspect that name choice to try an understand why everyone think it's confusing. your choice.  

On my side everything has been said and at this point, if you still believe `add` is correct, I'll accept to move on **if and only if** there is proper descriptive comments in both the method definition and at method call to explain the behaviour.

guyboertje:
The simple word `gravity` is used to name an incredibly troublesome concept but until someone comes up with a name that is more descriptive and explanatory - we are stuck with it. Of course it needs supplementary information before someone can say they understand the usage of the word in communication.

I would be happy to change `add` method name if someone can suggest one that:
- does not reflect its current usage e.g. set_filtered
- is more intention revealing and requires less documentation.

andrewvc:
How bout `#remove_from_queue` and `#add_new`?

ph:
I've read most of the threads here.

It depends how we want the behavior of the batch to be explicit, when I first read `#add` I though this call add a new event to the batch. I wasn't aware of what would happen if the event was already present, I though It would just create a new one and I would have duplicates.

I completely agree its the responsibility of the batch to make sure it dealt with duplicates and I also think its better to only have method to call either we have a new or `filtered` events this make code easier to reason to and delegate the actual work downstream.

If we take inspiration of other software like RDBMS and assume that every event in the batch is a row, that operation would be called `upsert`, even elasticsearch uses that term.

Why we dont use the same terminalogy here?

`#upsert` or `#insert_or_update(event)`

But my proposition might go agains't this :(

```
does not reflect its current usage e.g. set_filtered
```

guyboertje:
I want to be understand better.
Does the problem with the name `add` stem from it being confused with `dequeue` OR just the name `add` on its own?

guyboertje:
Also I would advise people to think of the Event object and what problems that caused when the API encouraged people to think of it as a Hash.
In this sense the Batch may be though of as an Array so I don't want to use words from the Array class that do not have exactly the same behaviour (forever).

guyboertje:
So the problem with understanding how duplicates would be handled would disappear if the Batch Object class name was BatchSet?

andrewvc:
@guyboertje I mean we could change the batch name, but I think we could just change the method name.

What if we renamed `#add` to `#merge`? That seems to best encapsulate what its doing to me.

colinsurprenant:
- I think I like `merge` better. 
- Personally I don't like the DB metaphor and `upsert`. this is not a DB, it's a collection of events.
- `dequeue

colinsurprenant:
LGTM.

2 caveats:
- For the purpose of this PR I think the shutdown/flush signal handling is ok because it uses the same logic as before, just within the batch. But, with the PQ coming up, signal handling should be moved out of the batch logic, this is not a batch concern and this is what I did in the feature/java_persistence branch. It makes a lot more sense to deal with signals at the pipeline level and outside the events stream.
- I am concerned about the performance penalty we are paying for the added complexity of juggling collections/hashes on most batch operations. I'd like to see how this could be optimized to minimize that performance cost.

guyboertje:
@colinsurprenant:

I have reorganized the way merged events are handled (can be improved more).
We will see in the benchmarks.

guyboertje:
@colinsurprenant - although I consider this PR to now be ready to merge, I want to wait for the midnight perf run in the kibana link before squash/merge.

guyboertje:
Sadly, it does not appear that the benchmarks have improved with the refactor. Time to profile.

suyograo:
@guyboertje what's the scoop with this? Should we hold up the merge for performance improvements? Or should we merge and then improve it? I would prefer merging it in to get some mileage..

guyboertje:
@suyograo 
I think it should be merged. Is Colin's LGTM enough?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
andrewvc: I added that in for future debugging purposes. I think its safe to remove it now if we aren't using it.

andrewvc:
@guyboertje LGTM, feel free to remove the inflight batches stuff if it isn't being used if you'd like.

purbon:
moin, as this is merged, I understand It can be removed from https://github.com/elastic/logstash-performance-testing/blob/master/web/api/config.yml#L10, isn't? I will remove from now, feel free to add again in case of wishing it back.

PR #5561, opened 2016-06-28
Make documentation the default formatter for acceptance tests

purbon:
Same way as regular tests in logstash, this PR make them run under the documentation level. This will help us debug errors in CI when they happen.


untergeek:
LGTM™

Issue #1637, opened 2014-08-15
Pipeline stops processing events because of exception in filterworker "#<NoMethodError: undefined method `>' for nil:NilClass>"


robinsmidsrod:
I'm getting this error in the logstash error log, and when it happens the pipeline stops processing events. The inputs are still able to receive events, but no events get through to the output stage.

```
{:timestamp=>"2014-08-14T08:09:19.415000+0000", :message=>"Exception in filterworker", "exception"=>#<NoMethodError: undefined method `>' for nil:NilClass>, "backtrace"=>["(eval):920:in `initialize'", "org/jruby/RubyProc.java:271:in `call'", "/opt/logstash/lib/logstash/pipeline.rb:262:in `filter'", "/opt/logstash/lib/logstash/pipeline.rb:203:in `filterworker'", "/opt/logstash/lib/logstash/pipeline.rb:143:in `start_filters'"], :level=>:error}
```

The error happens infrequently (about every other day or so), so I'm unsure what exactly is causing it.

I'm using the website DEB of logstash 1.4.2 on Ubuntu 14.04.

```
# java -version
java version "1.7.0_55"
OpenJDK Runtime Environment (IcedTea 2.4.7) (7u55-2.4.7-1ubuntu1)
OpenJDK 64-Bit Server VM (build 24.51-b03, mixed mode)
```

What other information do you need to debug this further?

tmortensen:
I ran into the same issue,  I am glad to report the workaround solved my issue. 
My filter sample is pretty short,  It for sure is a result of testing an empty field. 

http://pastebin.com/xy4r41rQ

chenryn:
I got similar exceptions when I wrote some codes in filters/ruby. I know I should use conditions before my `evnet['fieldname'].split()` operation, but, only 0.01% lines don't have such fieldname exists, why we must jadge this every line?

I image whether logstash can has a mechanism to restart the died filterworker to force the number of filterworkers equal `-w` setting.

derjohn:
Workaround?
  
ruby {
    # because if [somevar] doesnt work in all cases
    code => "if not defined? event['somevar']; event['somevar'] = 'dummyvalue' end"
  }

hmorgado:
Same thing happened to me and a simple verification if the field is not null solved the problem.
```
if [responsetime] {
        ruby {
	    code => "event['responsetime'] = event['responsetime'] / 1000"
        }
}
```

suyograo:
[...]

PR #5562, opened 2016-06-28
Add pipeline.continue_on_error setting to allow logstash to rescue arbitrary…

### LINKED ISSUE #5547: Improved pipeline error handling
Sometime logstash crashes and dies due to faultily written boolean statements. `[foo] > 5` works well until we hit an event where `[foo]` doesn't exist. Usually, the user means to write something like `[foo] and [foo] > 5`, but not always. Perhaps they need special logic for the `nil` case. These bugs can be sneaky however, as seen in #1637 . 

The pipeline can also die if a filter or output has a bug in it. This is especially common in uses of the ruby filter, which often make the same field being present assumptions.

For some users the right thing to do here is to crash logstash so an operator is forced to diagnose and fix the issue. For others, they generally just want it to log an error and keep on running. Logstash crashing is more an issue for these people.

After going back and forth on who was right WRT these issue @jsvd and converged on thinking that the current (crash by default) behavior is the right for a default, it being the more conservative option. 

We should however add a flag enabling the latter behavior for people who just want it to run, and if they miss a few events, not to worry about it. One small concern with the log and drop method is that it could easily fill up Logstash logs on disk, but that problem is to some extent outside of the scope of LS, and more a logrotate thing.

andrewvc:
This will rescue any plugin or pipeline config lang errors. This setting is only settable in the YAML.

Fixes #5547 and #1637

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -50,6 +50,18 @@
 #
 # pipeline.unsafe_shutdown: false
 #
+# Defaults to false. Setting this to true will catch all unexpected errors in the
+# pipeline without terminating logstash.
+# Logstash will do all of the following: drop the erroring event,
+# log an error-level message, and continue processing. This will also increment
+# the 'errored' count in the pipeline stats API at /_node/pipeline/events.
+#
+# If set to false logstash will immediately exit with an error in these
+# conditions, which may be more safe since potentially fewer events would be dropped.
+#
+# pipeline.continue_on_error: true
+#
+#
 # ------------ Pipeline Configuration Settings --------------
 #
 # Where to fetch the pipeline configuration for the main pipeline
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -33,7 +33,9 @@ module LogStash; class Pipeline
     :thread,
     :config_str,
     :settings,
-    :metric
+    :metric,
+    :namespace_events,
+    :namespace_pipeline
 
   MAX_INFLIGHT_WARN_THRESHOLD = 10_000
 
@@ -47,6 +49,7 @@ module LogStash; class Pipeline
     @settings = settings
     @pipeline_id = @settings.get_value("pipeline.id") || self.object_id
     @reporter = LogStash::PipelineReporter.new(@logger, self)
+    @continue_on_error = LogStash::SETTINGS.get("pipeline.continue_on_error")
 
     @inputs = nil
     @filters = nil
@@ -58,6 +61,15 @@ module LogStash; class Pipeline
     # sure the metric instance is correctly send to the plugins to make the namespace scoping work
     @metric = namespaced_metric.nil? ? LogStash::Instrument::NullMetric.new : namespaced_metric
 
+    @namespace_events = metric.namespace([:stats, :events])
+    @namespace_pipeline = metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :events])
+
+    # Zero out these counters
+    [:in, :filtered, :out, :errored].each do |name|
+      @namespace_events.increment(name, 0)
+      @namespace_pipeline.increment(name, 0)
+    end
+    
     grammar = LogStashConfigParser.new
     @config = grammar.parse(config_str)
     if @config.nil?
@@ -221,10 +233,7 @@ module LogStash; class Pipeline
   # Repeatedly takes batches off the queue, filters, then outputs them
   def worker_loop(batch_size, batch_delay)
     running = true
-
-    namespace_events = metric.namespace([:stats, :events])
-    namespace_pipeline = metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :events])
-
+    
     while running
       # To understand the purpose behind this synchronize please read the body of take_batch
       input_batch, signal = @input_queue_pop_mutex.synchronize { take_batch(batch_size, batch_delay) }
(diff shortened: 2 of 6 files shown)


config/logstash.yml
```diff
@@ -50,6 +50,16 @@
 #
 # pipeline.unsafe_shutdown: false
 #
+# Defaults to false. Setting this to true will rescue all exceptions in the
```
jordansissel: "rescue" is a ruby-specific term and 'exceptions' feels programmer-centric

Being that this should target users/operators, are there any better ways to describe this? "Unexpected errors" or something like that?

config/logstash.yml
```diff
@@ -50,6 +50,16 @@
...
+# pipeline.This means that if there are any errors in filters, or if a
+# conditional raises an exception logstash will drop the event, log an
```
jordansissel: suggest: "logstash will do all of the following: drop the erroring event, log an error-level message, and continue processing"

config/logstash.yml
```diff
@@ -50,6 +50,16 @@
...
+# error-level message, and continue processing.
+# If set to false logstash will immediately exit with an error in these
```
jordansissel: Immediately? (Is there data loss of unrelated in-flight events? If so, we should note that)

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -310,15 +302,37 @@ def output_batch(batch)
...
+    end
+  end
+
+  def with_pipeline_rescue(event_or_events)
+    yield
+  rescue Exception => e
+    logger_opts = {:message => e.message,
+                   :class => e.class.name,
+                   :backtrace => e.backtrace}
+    logger_message = "Exception in pipelineworker!"
```
ph: by default logstash only log stacktraces when logstash is run in debug mode, should we do the same here since stacktraces can get quite noisy ?

ph:
Code LGTM, I think @jordansissel's comment concerning the doc is valid and should be addressed before we merge this.

config/logstash.yml
andrewvc: Well, either way you lose data. You actually lose more if you set it to true, because then it just keeps going and dropping data.

andrewvc: I'm going to add in "If set to false logstash will immediately exit with an error in these
# conditions, which may be more safe since potentially fewer events would be dropped."

logstash-core/lib/logstash/pipeline.rb
andrewvc: Good call!

andrewvc:
@ph @jordansissel I have pushed a new version that should address your critiques.

ph:
build failed? 

```
rake aborted!
SyntaxError: /home/travis/build/elastic/logstash/logstash-core/lib/logstash/pipeline.rb:326: syntax error, unexpected '\n'
/home/travis/build/elastic/logstash/logstash-core/lib/logstash/agent.rb:10:in `(root)'
/home/travis/build/elastic/logstash/vendor/bundle/jruby/1.9/gems/logstash-devutils-1.0.2-java/lib/logstash/devutils/rspec/logstash_helpers.rb:1:in `(root)'
/home/travis/build/elastic/logstash/vendor/bundle/jruby/1.9/gems/logstash-devutils-1.0.2-java/lib/logstash/devutils/rspec/logstash_helpers.rb:1:in `(root)'
/home/travis/build/elastic/logstash/vendor/bundle/jruby/1.9/gems/logstash-devutils-1.0.2-java/lib/logstash/devutils/rspec/spec_helper.rb:1:in `(root)'
/home/travis/build/elastic/logstash/vendor/bundle/jruby/1.9/gems/logstash-devutils-1.0.2-java/lib/logstas

File: logstash-core/lib/logstash/pipeline.rb [hunk 5]
```diff
@@ -310,15 +302,35 @@ def output_batch(batch)
...
+
+  def with_pipeline_rescue(event_or_events)
+    yield
+  rescue Exception => e
+    logger_opts = {:message => e.message, :class => e.class.name}
+    logger_message = "Exception in pipelineworker!"
+
+    if @logger.info?
+      logger_opts[:events] = Array(event_or_events).map(&:to_hash)
+    end
```
ph: Maybe if debug? == true add stacktraces?

ph: Do we have any concerns here about sensitives data that could be leak here?

config/logstash.yml
```diff
@@ -50,6 +50,15 @@
 #
 # pipeline.unsafe_shutdown: false
 #
+# Defaults to false. Setting this to true will rescue all unexpected errors in the
```
jordansissel: "rescue" still seems the wrong term. If I put my user hat on, "rescue" feels like a "save the event!" kind of thing, when really we aren't saving or rescuign it, we're throwing it away... it's more like cutting out a bad thing...

Maybe we can combine the 2nd and 3rd sentences "If set to true, Logstash will take the following action on an event if it causes unexpected errors during processing: ... [explanation here]. If set to false, ..."

config/logstash.yml
```diff
@@ -50,6 +50,15 @@
...
+# pipeline. Logstash will do all of the following: drop the erroring event,
+# log an error-level message, and continue processing.
+# If set to false logstash will immediately exit with an error in these
+# conditions, which may be more safe since potentially fewer events would be dropped.
+#
+# pipeline.rescue_all
```
jordansissel: I think `rescue_all` is the wrong term. It's much less about rescuing (events) than it is about resovling error-causing event from the pipeline by destroying the event.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -310,15 +302,35 @@ def output_batch(batch)
...
+    
+    logger_message << "Enable info logging to see failed events" if !@logger.info?
```
suyograo: Lets also bump some metrics that we actually continued on for this event. We can expose this to the user.

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -100,6 +100,90 @@ class TestPipeline < LogStash::Pipeline
...
+        it "should increment the global errored count for metrics" do
+          expect(pipeline.namespace_events).to have_received(:increment).with(:errored, 1)
+        end
+
+        it "should increment the pipeline specific errored count for metrics" do
+          expect(pipeline.namespace_pipeline).to have_received(:increment).with(:errored, 1)
+        end
+      end
+    end
+  end
```
ph: that was easy (tm)

ph:
LGTM

suyograo:
Can we mark this as being experimental in the config docs?

andrewvc:
Will rename this to `continue_on_conditional_error` after closing https://github.com/elastic/logstash/issues/5593 , then this setting will only rescue nil check errors in conditionals

suyograo:
We discussed this internally, and it will be handled in the DLQ feature. Closing this for now.

PR #5566, opened 2016-06-29
Add more reporting to the QA acceptance test layer

purbon:
Till this PR the qa test layer was reporting vagrant commands only when they finished, so we could not spot issues with them when running the automated tests. This PR add's a flag to run them in debug mode, so all output from normal vagrant commands will also be shown when running like this. 

Not sure to make this default true, but this might be a good thing for CI infra. what do you think? 

Also fix the .gitignore file, so the .rspec, as this file was added in #5561should not be excluded from git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,5 @@ integration_run
 .mvn/
 qa/.vm_ssh_config
 qa/.vagrant
-qa/.rspec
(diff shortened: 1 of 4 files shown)


suyograo:
LGTM

PR #5569, opened 2016-06-30
Fix TOC and apply light edits to plugin generator topic

dedemorton:
@suyograo I don't think you meant for the "Generating Plugins" topic to become a parent topic of "Offline Plugin Management" and "Private Gem Repositories" as it is here:

generatingpluginstoc

So I've changed the heading level so that it appears nested under "Working with Plugins". I've also added a few edits since I had the file open.


jsvd:
LGTM

PR #5587, opened 2016-07-05
Make sure errors are reported also to terminal if show_startup_errors is true

### LINKED ISSUE #5549: Logstash exits silently when there are permission issues to log file
Reproducible on 2.3.3.

If the -l log file has permission issues, logstash just silently exists.  Will be nice if we log something to stdout, etc..

```
touch /tmp/z
chmod 0 /tmp/z
bin/logstash -l /tmp/z --debug
# exits silently.
echo $?
1
```

purbon:
Fixes #5549 by not masquerading  reported errors that happen during bootstrap time inside the agent class, this is not the right place to do the reporting as it should happen in runner, but all this is already properly refactored for 5.0 branch. In this PR I tried to keep the same style and logic when proposing a fix.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -133,7 +133,6 @@ class LogStash::Agent < Clamp::Command
   # Emit a warning message.
   def warn(message)
     # For now, all warnings are fatal.
-#    raise LogStash::ConfigurationError.new(message)
     signal_usage_error(message)
   end # def warn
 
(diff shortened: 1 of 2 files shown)


suyograo:
@purbon Can you take a look at the build failures?

jsvd:
LGTM

purbon:
@suyog @jsvd have you thought on backporting this to the newly created 2.4 branch?

suyograo:
Done, thanks for the heads up @purbon

PR #5589, opened 2016-07-05
Update coming annotations to reflect current release state

dedemorton:
I've already changed these in 5.0 (I think when I updated the release notes). Updating in master to keep things in sync.


suyograo:
LGTM

PR #5600, opened 2016-07-07
bump core to 2.4.0.dev and make plugin_api to 1.20

jsvd:
* bump logstash-core/logstash-core-event/logstash-core-event-java to 2.4.0.dev
* bump logstash-core-plugin-api to 1.20.0
  * provides both get/set and []/[]= events apis by allowing to be installed in logstash >= 2.4 and < 3
* minor bumps on gem dependencies

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,10 +2,10 @@
 # If you modify this file manually all comments and formatting will be lost.
 
 source "https://rubygems.org"
-gem "logstash-core", "3.0.0.dev", :path => "./logstash-core"
-gem "logstash-core-event", "3.0.0.dev", :path => "./logstash-core-event"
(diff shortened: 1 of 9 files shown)


suyograo:
@jsvd 1.20.0 is already taken by 2.3.4. Also, we should cut 2.4.0 soon

suyograo:
Just created 2.4 branch. Should we instead repurpose this PR to make it 2.4.0.snapshot1 in branch 2.4?

jsvd:
yep, I agree with that.
as for plugin-api version number, I'll move that to 1.21.
for the remainder of the 2.3.4 snapshots and ga, can we just bump 1.20.1, .2, .3 and so on?

jsvd:
closing PR to target 2.4

PR #5605, opened 2016-07-07
Add relnotes for all 2.3 releases to the doc

dedemorton:
Updated the published relnotes to include items from the changelog for all versions 2.3 > 2.3.4. That way we don't have to keep updating the link on the products page for patch releases.


suyograo:
Looks great! Please merge

Issue #4918, opened 2016-03-29
Clean up table that shows examples of using environment variables


dedemorton:
The formatting of the table under Examples is too complex and makes the content harder to read:
https://www.elastic.co/guide/en/logstash/2.3/environment-variables.html

This table needs to be broken down and simplified. Either convert the table content into text with a few examples, or break the table up, using individual example titles, with the 'Environment' column as the example title, so you have 'Example 1: export TCP_PORT=12345' and then a two column thing for config source/config result, which will be less ugly.

PR #5612, opened 2016-07-09
Convert examples from table to text to fix rendering issue

dedemorton:
The formatting of the original table was hard to read and rendered poorly. Instead, I've broken the content down into paragraphs with examples. To make the info easier to scan, I broke the examples up into sections by use case. I did this mainly to group the examples so that they were easier to scan (rather than to highlight any specific use cases for using env vars).

### DIFF
--- a/docs/static/configuration.asciidoc
+++ b/docs/static/configuration.asciidoc
@@ -636,62 +636,59 @@ output {
 ----------------------------------
 
 [[environment-variables]]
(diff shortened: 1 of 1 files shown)


dedemorton:
This fixes issue https://github.com/elastic/logstash/issues/4918

suyograo:
LGTM

PR #5618, opened 2016-07-11
Add support for hostname:port syntax to SafeURI class

andrewvc:
This is a deviation from what the ruby URI class normally allows.
With this patch users can use "myhost:123" as an option and have it do the
right thing, as opposed to before where they'd get an error and have to
use "//myhost:123".

### DIFF
--- a/logstash-core/spec/logstash/config/mixin_spec.rb
+++ b/logstash-core/spec/logstash/config/mixin_spec.rb
@@ -178,19 +178,13 @@ describe LogStash::Config::Mixin do
       end
     end
 
-    shared_examples("safe URI") do            
+    shared_examples("safe URI") do |options|
+      options ||= {}
+      
       subject { klass.new("uri" => uri_str) }
 
       it "should be a SafeURI object" do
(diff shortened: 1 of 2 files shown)


andrewvc:
I have verified that this fixes the failing specs in https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/450

talevy:
what's the deal with the unchanged toubleshooting asciidoc?

logstash-core/spec/logstash/config/mixin_spec.rb
```diff
@@ -178,22 +178,17 @@
...
-      it "should make password values hidden with #to_s" do
-        expect(subject.uri.to_s).to eql(uri_hidden)
-      end
-
-      it "should make password values hidden with #inspect" do
-        expect(subject.uri.inspect).to eql(uri_hidden)
-      end
-
       it "should correctly copy URI types" do
+
```
talevy: nitpick

talevy:
LGTM

PR #5619, opened 2016-07-11
Add quicktune guide

andrewvc:
This is a guide to help users quickly ascertain performance problems without getting their hands too dirty with internals knowledge.

Still need to figure out where to link to this from.


suyograo:
LGTM

Issue #5607, opened 2016-07-08
Final touches on monitoring API


acchen97:
Final monitoring API bug fixes for beta1 readiness:

- [x] Pipeline stats from `/_node/stats/pipeline` should also be included in the parent `/_node/stats` resource for completeness; verbosity shouldn’t be an issue as this API should be for computers not humans
- [x] `/_node/stats/mem` resource should be removed and the object should instead be nested under `/_node/stats/jvm` along with the ‘thread’ stats
- [x] `/_node/stats/events` object+resource should be removed as these metrics now instead reside under the pipeline stats `/_node/stats/pipeline`
- [x] There is a bug in `/_node/stats/pipeline` where the plugin specific stats should be keyed by “plugins” and not “pipeline"
- [x] Adding input plugin specific stats, like what we have for filters/outputs already
- [x] `/_node/hot_threads?human=true` human option needs to be fixed, currently returns in JSON
- [x] Enabling specification of a comma-separated list of resources https://github.com/elastic/logstash/issues/5609

/cc @jordansissel @suyograo @andrewvc

acchen97:
@purbon thanks for hopping on this. Added one more item at the end here https://github.com/elastic/logstash/issues/5609. Please let me know if it won't fit into beta1. /cc @suyograo

purbon:
Added metrics to tcp input, pending review. see https://github.com/logstash-plugins/logstash-input-tcp/pull/56

acchen97:
> as of now only Adding input plugin specific stats, like what we have for filters/outputs already is missing. investigating working on this.

Cool, this is important to gain monitoring parity across the stages of the pipeline. 👍

acchen97:
This is complete for alpha5, will open a new issue for tracking beta1 readiness.

PR #5621, opened 2016-07-12
Fix human parameter processing

### LINKED ISSUE #5563: human parameter on the hot threads API returns a response that's not really human readable
When I run the hot threads API with `?human=true` specified, the results are not really human readable, and specifying `?human=false` does not seem to change the response.

In previous alpha versions, I believe that we returned a more readable response, as shown in the [documentation](https://www.elastic.co/guide/en/logstash/5.0/hot-threads-api.html) for the hot threads API. I'm opening this issue in case this change was not intentional.

- Version: 5.0.0-alpha4
- Operating System: OS X
- Config File: NA
- Sample Data: NA
- Steps to Reproduce:
Call the hot threads API and specify `?human=true`. The call returns a response that seems to be in XML format, but (IMHO) isn't really human readable. Setting `human=false` returns the same document. This makes me wonder if the `human` parameter is working as expected.

purbon:
Add fixes for the human parameter processing, for now only related to hot threads api, but nothing prevents to have it for others resources in the feature.

Changes in this PR basically make sure that the `.to_s` method requested when the output should be in text is the right now and no extra information is in between.

Fixes #5563 
Related https://github.com/elastic/logstash/issues/5607

### DIFF
--- a/logstash-core/lib/logstash/api/commands/node.rb
+++ b/logstash-core/lib/logstash/api/commands/node.rb
@@ -62,7 +62,7 @@ module LogStash
           end
           
           def to_s
-            hash = to_hash
+            hash = to_hash[:hot_threads]
             report =  "#{I18n.t("logstash.web_api.hot_threads.title", :hostname => hash[:hostname], :time => hash[:time], :top_count => @thread_dump.top_count )} \n"
             report << '=' * 80
             report << "\n"
@@ -98,7 +98,7 @@ module LogStash
               thread[:traces] = traces unless traces.empty?
               hash[:threads] << thread
             end
-            hash
+            { :hot_threads => hash }
           end
 
           def cpu_time_as_percent(hash)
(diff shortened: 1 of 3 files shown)


logstash-core/lib/logstash/api/commands/node.rb
```diff
@@ -62,7 +62,7 @@ def initialize(cmd, options)
           end
           
           def to_s
-            hash = to_hash
+            hash = to_hash[:hot_threads]
```
andrewvc: Isn't this still JSON output? I'm fine with this change, but I wonder if we need another issue to write true human readable hot_threads output.

If I hit `curl 'http://localhost:9600/_node/hot_threads?h'` on this branch I still get JSON.

andrewvc: My mistake, I should have added `?human` instead of `?h`.

andrewvc:
LGTM

logstash-core/lib/logstash/api/commands/node.rb
purbon: @andrewvc what you saw here is that the hot threads should return a class that actually have a `.to_s` method, the one used to serialize the data as "human", if you return a hash for sure the `to_s` method is not there to convert it to human. same thing or any other changes, at less the way the `respond_with` method is implemented right now.

PR #5623, opened 2016-07-12
Collection of small fixes for the web api

### LINKED ISSUE #5607: Final touches on monitoring API

purbon:
List of errors fixed:

- [x] Pipeline stats from `/_node/stats/pipeline` should also be included in the parent `/_node/stats` resource for completeness; verbosity shouldn’t be an issue as this API should be for computers not humans
- [x] `/_node/stats/mem` resource should be removed and the object should instead be nested under `/_node/stats/jvm` along with the ‘thread’ stats
- [x] `/_node/stats/events` object+resource should be removed as these metrics now instead reside under the pipeline stats `/_node/stats/pipeline`
- [x] There is a bug in `/_node/stats/pipeline` where the plugin specific stats should be keyed by “plugins” and not “pipeline"

Fixes #5607

### DIFF
--- a/logstash-core/spec/api/spec_helper.rb
+++ b/logstash-core/spec/api/spec_helper.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 API_ROOT = File.expand_path(File.join(File.dirname(__FILE__), "..", "..", "lib", "logstash", "api"))
 
+require "stud/task"
 require "logstash/devutils/rspec/spec_helper"
 $LOAD_PATH.unshift(File.expand_path(File.dirname(__FILE__)))
 require "lib/api/support/resource_dsl_methods"
--- a/logstash-core/lib/logstash/api/commands/stats.rb
+++ b/logstash-core/lib/logstash/api/commands/stats.rb
@@ -11,7 +11,8 @@ module LogStash
               [:jvm, :threads],
               :count,
               :peak_count
-            )
+            ),
+            :mem => memory
           }
         end
 
--- a/logstash-core/lib/logstash/api/commands/stats.rb
+++ b/logstash-core/lib/logstash/api/commands/stats.rb
@@ -139,7 +139,7 @@ module LogStash
 
             {
               :events => stats[:events],
-              :pipeline => {
+              :plugins => {
                 :inputs => plugin_stats(stats, :inputs),
                 :filters => plugin_stats(stats, :filters),
                 :outputs => plugin_stats(stats, :outputs)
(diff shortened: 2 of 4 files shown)


logstash-core/spec/api/spec_helper.rb
```diff
@@ -1,6 +1,7 @@
 # encoding: utf-8
 API_ROOT = File.expand_path(File.join(File.dirname(__FILE__), "..", "..", "lib", "logstash", "api"))
 
+require "stud/task"
```
andrewvc: Why was this added?

andrewvc:
Aside from the stud/task comment LGTM. Once we close that one out this is 👍

logstash-core/spec/api/spec_helper.rb
purbon: because if you try to run only the api test with `bin/rspec` without this in the helper it does not work.

purbon:
@andrewvc what do you think?

logstash-core/spec/api/spec_helper.rb
andrewvc: What code is using this? I don't see any code in the PR using stud. 

Is there a missing require at the top of a .rb file somewhere? We probably shouldn't put requires in the spec_helper for libraries used by app code.

purbon: code using is https://github.com/elastic/logstash/blob/master/logstash-core/spec/api/spec_helper.rb#L59, i only added this because I run the test with `bin/rspec` only for the api, not for everything. if I don't add the requirement here it does not work.

andrewvc: Makes sense

andrewvc:
LGTM

PR #5628, opened 2016-07-13
fix string setting validation of possible values

jsvd:
currently when a String setting is created with a range of possible values, this range is not properly checked, allowing a setting to have a value outside of the array.
This PR properly introduces validation and tests to confirm it works as it should


jordansissel:
Code change LGTM. Travis seems happy.

LGTM

Issue #5629, opened 2016-07-13
[webapi] Hot Threads code is duplicated, need refactoring.


purbon:
Hot threads code is duplicated in https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/api/commands/node.rb#L54 and https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/api/commands/node.rb#L54 as the function is the same, this code should be refactored.

PR #5630, opened 2016-07-13
Add support for first level response filtering in the webapi

### LINKED ISSUE #5609: Add support to monitoring API for specifying a comma-separated list
When users call the Elasticsearch APIs, they can specify a comma-separated list of resources to limit the data that's returned by the call. For example, see https://www.elastic.co/guide/en/elasticsearch/reference/master/cluster-nodes-stats.html.

The Logstash monitoring APIs in 5.0 should support the same capability.

purbon:
Hi,
  this PR provide a solution to filter the outcome of first level API responses in the webapi, in the same fashion of what ES is doing, with this changes you can now have urls like:

* `http://localhost:9600/_node/stats` that will respond with the complete hash answer

or 

* `http://localhost:9600/_node/stats/jvm,process` that will only show jvm and process information.

For sure the former  `http://localhost:9600/_node/stats/jvm` or `http://localhost:9600/_node/stats/process` are still an option.

Fixes #5609 #5629
Related to #5607

### DIFF
--- a/logstash-core/lib/logstash/api/commands/default_metadata.rb
+++ b/logstash-core/lib/logstash/api/commands/default_metadata.rb
@@ -1,3 +1,5 @@
+# encoding: utf-8
+
 require "logstash/api/commands/base"
 
 module LogStash
--- a/logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
+++ b/logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
@@ -11,7 +11,7 @@ class HotThreadsReport
   end
 
   def to_s
-    hash = to_hash
+    hash = to_hash[:hot_threads]
     report =  "#{I18n.t("logstash.web_api.hot_threads.title", :hostname => hash[:hostname], :time => hash[:time], :top_count => @thread_dump.top_count )} \n"
     report << '=' * 80
     report << "\n"
@@ -47,7 +47,7 @@ class HotThreadsReport
       thread[:traces] = traces unless traces.empty?
       hash[:threads] << thread
     end
-    hash
+    { :hot_threads => hash }
   end
 
   def cpu_time_as_percent(hash)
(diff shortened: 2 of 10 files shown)


purbon:
@andrewvc can you also take a look here? This has been a requested by @acchen97  inside #5609  and #5607 .

purbon:
need to check why test are broken 👎

logstash-core/lib/logstash/api/commands/default_metadata.rb
```diff
@@ -1,3 +1,5 @@
+# encoding: utf-8
```
andrewvc: Thanks for catching these!

logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
```diff
@@ -0,0 +1,61 @@
+# encoding: utf-8
```
andrewvc: Good call extracting this class into its own file.

logstash-core/lib/logstash/api/modules/node_stats.rb
```diff
@@ -13,47 +13,19 @@ class NodeStats < ::LogStash::Api::Modules::Base
...
           payload = {
             :events => events_payload,
             :jvm => jvm_payload,
             :process => process_payload,
-            :mem => mem_payload
+            :mem => mem_payload,
+            :pipeline => pipeline_payload
           }
-
+          payload.select! { |k,v| selected_fields.include?(k) } unless selected_fields.empty?
```
andrewvc: We use this code in two places. Maybe we could make this into a custom helper or add it into `respond_with` by making it take an additional argument?

suyograo:
@purbon most failing tests are based on hot_threads. can you check if they are flaky or genuine?

logstash-core/lib/logstash/api/modules/node_stats.rb
purbon: makes sense.

purbon:
@suyograo investigating.

purbon:
@andrewvc ammended, what do you think?

andrewvc:
LGTM.

PR #5639, opened 2016-07-15
Add rake task to build rpm and deb all-plugins

### LINKED ISSUE #4361: Add DEB and RPM download options for uber package
DEB and RPM download options for the uber package should be created as well for consistency.

https://www.elastic.co/downloads/logstash

suyograo:
Previously we only built zip and tar for all-plugins. Fixes #4361


talevy:
LGTM

Issue #5593, opened 2016-07-06
Better conditional error messages


andrewvc:
Currently, if a logstash config contains a boolean expression like `if [foo] > 2` and `[foo]` is of the wrong type or nil it will raise exceptions that can be confusing. We should do the following:
1. Wrap all conditionals in a `begin; end` block, and re-raise the exception as a wrapped `LogStash::Pipeline::ConditionalError`. This can be caught be the code in https://github.com/elastic/logstash/pull/5562
2. Print out a friendlier error message like: "Your expression [foo] > 2 failed, probably because a field was null. The exact error wash ‘undefined method `>' for true:TrueClass".

andrewvc:
Should be renamed to `continue_on_conditional_error`

jsvd:
Sorry to backtrack here, but I'd like to steer away from focusing this too much on conditional errors.
The spectrum of exceptions that we can tolerate here should be more than conditional errors, but also uncaught filter exceptions and output exceptions.
This is also a commonly requested feature, so this flag would simply make logstash, as whole, try to plow through exceptions (disabled by default).

andrewvc:
@jsvd we really need this feature if we're going to do #5641 , in fact we need it even more than we need #5641 because users are commonly confused with the error messages we currently give out and don't realize its their config that are going to be broken.

We should definitely wrap and re-raise all conditional errors and provide as much context as possible.

C0rn3j:
`(TypeError) no implicit conversion of nil into Integer` is still a thing 8 years later (8.13.x).

It looks like 8.16.0+ will have something for this though? https://github.com/elastic/logstash/pull/16322

PR #5641, opened 2016-07-15
pipeline.abort_on_error setting to abort on unexpected errors (enabled by default)

### LINKED ISSUE #5547: Improved pipeline error handling

jsvd:
This will rescue any plugin or pipeline config lang errors. This setting is only settable in the YAML.

Fixes #5547 and #1637 

a continuation/replacement of  https://github.com/elastic/logstash/pull/5562

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -57,15 +57,18 @@
 #
 # pipeline.unsafe_shutdown: false
 #
-# Catch all unexpected errors in the pipeline without terminating logstash.
+# By default, logstash will immediately exit when presented with an unexcepted
+# exception. This is due to the fact that an unknown exception can have an
+# unpredictable effect on filter/output state and data flow.
+#
+# Disabling abort_on_error makes Logstash catch all unexpected errors in the
+# pipeline without terminating logstash.
 # Logstash will do all of the following: drop the erroring event,
 # log an error-level message, and continue processing. This will also increment
 # the 'errored' count in the pipeline stats API at /_node/pipeline/events.
 #
-# If set to false logstash will immediately exit with an error in these
-# conditions, which may be more safe since potentially fewer events would be dropped.
 #
-# pipeline.continue_on_error: false
+# pipeline.abort_on_error: true
 #
 #
 # ------------ Pipeline Configuration Settings --------------
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -57,6 +57,18 @@
 #
 # pipeline.unsafe_shutdown: false
 #
+# Defaults to false. Setting this to true will catch all unexpected errors in the
+# pipeline without terminating logstash.
+# Logstash will do all of the following: drop the erroring event,
+# log an error-level message, and continue processing. This will also increment
+# the 'errored' count in the pipeline stats API at /_node/pipeline/events.
+#
+# If set to false logstash will immediately exit with an error in these
+# conditions, which may be more safe since potentially fewer events would be dropped.
+#
+# pipeline.continue_on_error: true
+#
+#
 # ------------ Pipeline Configuration Settings --------------
 #
 # Where to fetch the pipeline configuration for the main pipeline
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -57,8 +57,7 @@
 #
 # pipeline.unsafe_shutdown: false
 #
-# Defaults to false. Setting this to true will catch all unexpected errors in the
-# pipeline without terminating logstash.
+# Catch all unexpected errors in the pipeline without terminating logstash.
 # Logstash will do all of the following: drop the erroring event,
 # log an error-level message, and continue processing. This will also increment
 # the 'errored' count in the pipeline stats API at /_node/pipeline/events.
@@ -66,7 +65,7 @@
 # If set to false logstash will immediately exit with an error in these
 # conditions, which may be more safe since potentially fewer events would be dropped.
 #
-# pipeline.continue_on_error: true
+# pipeline.continue_on_error: false
 #
 #
 # ------------ Pipeline Configuration Settings --------------
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -49,7 +49,7 @@ module LogStash; class Pipeline
     @settings = settings
     @pipeline_id = @settings.get_value("pipeline.id") || self.object_id
     @reporter = LogStash::PipelineReporter.new(@logger, self)
-    @continue_on_error = LogStash::SETTINGS.get("pipeline.continue_on_error")
+    @abort_on_error = LogStash::SETTINGS.get("pipeline.abort_on_error")
 
     @inputs = nil
     @filters = nil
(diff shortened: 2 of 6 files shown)


jsvd:
@andrewvc I posted the following in https://github.com/elastic/logstash/issues/5593 but I'll replicate here:

> Sorry to backtrack here, but I'd like to steer away from focusing this PR too much on conditional errors.
> The spectrum of exceptions that we can tolerate here should be more than conditional errors, but also uncaught filter exceptions and output exceptions.
> This is also a commonly requested feature, so this flag would simply make logstash, as whole, try to plow through exceptions (disabled by default).

What do you think?

suyograo:
I too think just catching conditional errors are limiting here.

I like the original scope as well -- to catch _all_ filter and output exceptions and _continue_ processing. Remember this will not be enabled by default, and default LS behavior would still be its original form -- not to drop any event no matter what. This provides the operator the choice if their usecase allows them to continue. I would also argue that since we have metrics in 5.0 its fairly easy to track the rate of dropped events (`error` count). This is an important metric.

As an extra suggestion, when this flag is turned on, maybe we pipe any events dropped (during bad processing) to a file (in `path.data`). We don't even have to call this DLQ feature. The same rules that apply for internal logs apply to this new file. Use `logrotate` or other solutions to manage this.

@jordansissel @andrewvc thoughts?

suyograo:
Outputs are tricky though. -1 on doing this for outputs. IMO it should be fail-fast for outputs. You want to know immediately if there is a connection issue and if your events are not making to the defined outputs.

jsvd:
I think we need to be clearer on the purpose of this option. The documentation in the yaml says "# Catch all unexpected errors in the pipeline without terminating logstash."
For me this is a correct assessment: during the pipeline execution, catch any exception the plugin code or the normal pipeline code doesn't handle, and "deal with it".
This happens for a number of different reasons:
1) unexpected exceptions in filters
2) unexpected exceptions in outputs
3) errors in the AST evaluation (e.g. `[field] > 1` conditionals where `field` is nil) 
4) something else we have no idea what

From this perspective we should catch any/everything.
That said, I'm ok with refocusing this PR with the concern 3), which means renaming the config option and specializing the exception catching for stuff from the AST.

purbon:
Moin,
  after giving a bit of thought here I think I like Joao proposal while at the same time I think Suyog point is fair. We would like to know if output messages are not making it, and with Joao proposal we know it as we increase the monitoring API and also do logging, but we've the option of not making LS raise an exception.

I think this is  fair compromise for errors situations where LS might be able to recover from.

ph:
After looking at the code and the comments from @suyograo and @jsvd.

AST or runtime conditionals applies to both  the filter stage and the output stage making theses stage really similar in that domain. The current code go a little bit higher and catch all exception, which I think is actually a better representation of what a `continue_on_error` is, and I believe its conceptually a better solution.

If a plugin raise an exception up to the pipeline for me its a bug and it should be handled inside the plugin, but for a user theses errors could be **transient** and don't necessary mean logstash should stop the world.

So I am OK with the following flow:

- mark the event as poison, and remove it from future plugin execution or stage.
- log an exception
- increment errors metric that users can monitor.

Also I think we should make it clearer in the log statement that an event was dropped. see https://github.com/elastic/logstash/pull/5641/files#diff-bc1e9ea81caa54ea0d56abf4415fdf00R338

suyograo:
After a meeting with the team, it was evident this feature is really desired. @jordansissel suggested we change the name of the setting to `pipeline.abort_on_error` and make the default false. Users interested in ploughing through exceptions can flip it to true. Also tag this feature `experimental`.

suyograo:
This needs to be backported to 2.4 as well.

jsvd:
@suyograo I believe you meant to say default abort_on_error to true, and users can turn it off if they want.

suyograo:
@jsvd nope, we want this option to be behind a flag, and not enabled out of the box. OOTB, LS is always fail-fast and never-lose-events. So `abort_on_error` is false, and can be turned on for this "keep moving, nothing to see here" option.

jsvd:
One question, if we backport this to 2.4, we'll have to add this as command line flag, since the yaml configuration for settings has only been introduced in the master/5.x branch
It will likely be called --abort-on-error=true/false

jordansissel:
Yeah abort on error should default to true -- which is today's behavior of
crashing on unexpected bugs in plugins.

suyograo:
@jsvd is correct. I was getting confused :) +1 on `--abort-on-error` for 2.4

config/logstash.yml
```diff
@@ -57,6 +57,20 @@
 #
 # pipeline.unsafe_shutdown: false
 #
+# By default, logstash will immediately exit when presented with an unexcepted
```
andrewvc: s/unexcepted/unexpected/ (nice freudian slip though!)

config/logstash.yml
```diff
@@ -57,6 +57,20 @@
...
+# exception. This is due to the fact that an unknown exception can have an
+# unpredictable effect on filter/output state and data flow.
+#
+# Disabling abort_on_error makes Logstash catch all unexpected errors in the
```
andrewvc: ooo, I like making this something you disable vs. enable 👍

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -307,18 +306,37 @@ def output_batch(batch)
...
+    outputs_events.each do |output, events|
+      with_pipeline_rescue(events) do
+        output.multi_receive(events)
+      end
+    end
+  end
+
+  def with_pipeline_rescue(event_or_events)
+    yield
+  rescue Exception => e
```
andrewvc: Do we really want to rescue `Exception` and not `StandardError`?

See https://robots.thoughtbot.com/rescue-standarderror-not-exception

andrewvc:
I'm fine with catching a wider number of things, but we really need #5593 in if we include this IMHO.

The error messages are a much bigger problem for the average user than the crashing.

andrewvc:
Looking good! Will come back for a LGTM after my comments are vetted.

andrewvc:
Thinking through it a bit more the output case really concerns me. Currently if the output borks we lose what we have in memory. In this case a borked output will just drop stuff constantly.

That same potential exists for conditionals but it probably isn't as bad. 

I don't think I'm willing to go to the mat on this, esp. since its an optional feature, but I am concerned here. Since this is experimental I'm ok with releasing it as is and seeing if complaints occur.

logstash-core/lib/logstash/pipeline.rb
andrewvc: Actually, I'm fine with this since those concerns don't apply to child threads.

suyograo:
Closing this since this will be handled in DLQ project

Issue #5650, opened 2016-07-19
[META] Log4j2 logging in core


talevy:
**Goal:**
* Unify logging across Logstash's Java and Ruby code.
* Support changing logging level dynamically at runtime through REST endpoints
* Support complete JSON logging
* Provide a strategy for reading in its own JSON logs as input for self monitoring

**Proposed Implementation:**
Migrate from Cabin to Log4j2 as the logging library for Logstash Core and its plugins.

**Reason for library change:**
Logstash and its plugin's usage of Java libraries that use Log4j/SLF4j is growing. This is leading to un-captured logging by core as it is running the pipeline. [Cabin](https://github.com/jordansissel/ruby-cabin) is the current logging library used by Logstash, but it's support for these java libraries is limited. Configuration of these java loggers is also not managed by Logstash natively, this yields confusing warning messages when using specific plugins, like Kafka. Even Logstash's core (Logstash Event in Java) is straying away from Cabin-friendly logging land.

**Tasks For 5.0-beta1:**
- [X] Implement JSON logger general Objects passed-in as secondary arguments to the log records
- [X] Support SLF4j usage in plugins
- [X] Update `@logger` usage in all plugins in `logstash-plugins` repo to be compatible with the new interface
- [X] Introduce an `ArrayListAppender` to be used by logging-related tests
- [X] `log4j2.properties` file is picked up in `logstash.lib.sh` and found in the `conf` dir.
- [x] Re-use Java Event's Javafier to safely/recursively convert between ruby and java types to be logged by Log4j2 (under debate)
- [X] Update log instances used in Core to use Log4j2 loggers, with Java-compatible arguments
- [x] re-enable `path.logs` to specify logs directory, and override log4j config at startup with it https://github.com/elastic/logstash/pull/5876
- [x] introduce log4j configuration for tests via devutils (https://github.com/elastic/logstash-devutils/pull/60)
- [ ] Fix hardcoded version of java project to rely on `version.rake` to correctly set https://github.com/elastic/logstash/pull/5882
- [x] expose log-level changing API https://github.com/elastic/logstash/pull/5871
- [ ] Logstash Event Core Java needs updating to the new logging infrastructure
- [x] --log-in-json property. until this is made.. one has to manually update the `log4j2.properties` file to log in json

Plugin PRs:

- [x] jdbc input plugin
- [x] https://github.com/logstash-plugins/logstash-input-exec/pull/10
[...]

PR #5651, opened 2016-07-19
Migrate Logstash to Log4j2 Logging

talevy:
Meta issue describing this issue: https://github.com/elastic/logstash/issues/5650

_****Quoted from the meta issue****_

**Goal:**
* Unify logging across Logstash's Java and Ruby code.
* Support changing logging level dynamically at runtime through REST endpoints
* Support complete JSON logging
* Provide a strategy for reading in its own JSON logs as input for self monitoring

**Proposed Implementation:**
Migrate from Cabin to Log4j2 as the logging library for Logstash Core and its plugins.

**Reason for library change:**
Logstash and its plugin's usage of Java libraries that use Log4j/SLF4j is growing. This is leading to un-captured logging by core as it is running the pipeline. [Cabin](https://github.com/jordansissel/ruby-cabin) is the current logging library used by Logstash, but it's support for these java libraries is limited. Configuration of these java loggers is also not managed by Logstash natively, this yields confusing warning messages when using specific plugins, like Kafka. Even Logstash's core (Logstash Event in Java) is straying away from Cabin-friendly logging land.


fbaligand:
Hi, 
Impressive work !

Just a remark : I think that ExampleEvent.java and Main.java should be in `src/test/java` folder.
It's not intended for logstash runtime use.

logstash-core/src/main/java/org/logstash/log/StructuredMessage.java
```diff
@@ -0,0 +1,93 @@
+package org.logstash.log;
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.apache.logging.log4j.message.Message;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+@JsonSerialize(using = CustomLogEventSerializer.class)
```
fbaligand: It seems that annotation `@JsonSerialize(using = CustomLogEventSerializer.class) ` is not useful.

logstash-core/lib/logstash/agent.rb
```diff
@@ -176,18 +176,18 @@ def create_pipeline(settings, config=nil)
...
         return
       end
     end
 
     begin
       LogStash::Pipeline.new(config, settings, metric)
     rescue => e
-      if @logger.debug?
-        @logger.error("fetched an invalid config", :config => config, :reason => e.message, :backtrace => e.backtrace)
+      if @logger.is_debug_enabled
```
suyograo: can we keep the original `debug?` as well? we need to deprecate it and then remove it for plugin developers.

suyograo: Not sure why we need to change it at all...

logstash-core/src/main/java/org/logstash/log/StructuredMessage.java
talevy: true, the current state of the code is part exploratory. so there are lingering paths of code that may or may not be necessary in the next revision

logstash-core/lib/logstash/agent.rb
talevy: this was here with the original intent of not having a light wrapper around Log4j's logger object. In the upcoming refactor, these changes will be reverted as I will introduce a `#debug?` method to the future wrapper

logstash-core/src/main/java/org/logstash/log/StructuredMessageContextSelector.java
```diff
@@ -0,0 +1,33 @@
...
+        protected Logger newInstance(final LoggerContext ctx, final String name, MessageFactory messageFactory)
+        {
+            if (null == messageFactory)
+                messageFactory = StructuredMessageFactory.INSTANCE;
+
+            return super.newInstance(ctx, name, messageFactory);
+        }
+
+    }
+}
```
fbaligand: If I understand well, the only reason why you create a custom ContextSelector, is to define a custom default MessageFactory, right ?

In this case, it exists a specific system property for this need : `log4j2.messageFactory`
Using this property, you don't need to create a custom ContextSelector (which is done to select current log4j context, so it is useful only if you manage multiple log4j contexts).

suyograo:
@talevy can you fix the tests?

Gemfile
```diff
@@ -111,3 +111,24 @@ gem "logstash-output-stdout"
...
+gem "timecop", ">= 0"
+gem "jdbc-derby", ">= 0"
+gem "docker-api", ">= 0"
+gem "longshoreman", ">= 0"
+gem "elasticsearch", ">= 0"
+gem "sinatra", ">= 0"
+gem "webrick", ">= 0"
+gem "poseidon", ">= 0"
+gem "snappy", ">= 0"
+gem "webmock", "~> 1.21.0"
```
ph: this is probably left from installing something, let make sure we dont merge it in master.

logstash-core/lib/logstash/util/loggable.rb
```diff
@@ -1,29 +1,19 @@
...
-          Loggable.logger
-        end
-      end
-    end
-
-    def logger
-      Loggable.logger
-    end
   end
 end; end
```
ph: @talevy I am OK with this code, I cannot think of a better way of doing it if we want control of the emitters.
Also, I prefer this by a lot than passing the logger around.

+1 <3

logstash-core/logstash-core.gemspec
```diff
@@ -14,12 +14,11 @@ Gem::Specification.new do |gem|
   gem.files         = Dir.glob(["logstash-core.gemspec", "lib/**/*.rb", "spec/**/*.rb", "locales/*", "lib/logstash/api/init.ru"])
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.name          = "logstash-core"
-  gem.require_paths = ["lib"]
+  gem.require_paths = ["lib", "vendor/jars"]
```
ph: Can use the same directory as our other projects?

logstash-core-event-java put them in `lib/logstash-core-event-java/
Kafka (in|out) and beats put them in `vendor/jar-dependencies`

logstash-core/logstash-core.iml
```diff
@@ -0,0 +1,346 @@
...
+        </CLASSES>
+        <JAVADOC/>
+        <SOURCES>
+          <root url="jar:///Users/tal/.gradle/caches/modules-2/files-2.1/com.fasterxml.jackson.core/jackson-annotations/2.7.0/d0d412f71000701f765e1e280764e92337c659f5/jackson-annotations-2.7.0-sources.jar!/"/>
+        </SOURCES>
+      </library>
+    </orderEntry>
+  </component>
+  <component name="ModuleRootManager"/>
+</module>
```
ph: should this be in the `.gitignore`?

logstash-core/lib/logstash/api/rack_app.rb
```diff
@@ -6,22 +6,23 @@
...
       METADATA_FIELDS = [:request_method, :path_info, :query_string, :http_version, :http_accept].freeze
       def self.log_metadata(status, env)
-        METADATA_FIELDS.reduce({:status => status}) do |acc, field|
-          acc[field] = env[field.to_s.upcase]
+        METADATA_FIELDS.reduce({"status" => status}) do |acc, field|
+          acc[field.to_s] = env[field.to_s.upcase]
           acc
-        end        
+        end
       end
```
ph: I am not sure why this change is required if we serialize symbol?

logstash-core/lib/logstash/config/mixin.rb
```diff
@@ -396,7 +394,7 @@ def validate_check_parameter_values(params)
             # Used for converting values in the config to proper objects.
             params[key] = processed_value
           else
-            @logger.error(I18n.t("logstash.runner.configuration.setting_invalid",
+            self.logger.error(I18n.t("logstash.runner.configuration.setting_invalid",
                                  :plugin => @plugin_name, :type => @plugin_type,
                                  :setting => key, :value => value.inspect,
                                  :value_type => config_settings[:validate],
```
ph: there is no need to call `self`, The `logger` should be found in the current lookup of the class.

logstash-core/lib/logstash/environment.rb
```diff
@@ -4,6 +4,7 @@
 require "logstash/settings"
 require "socket"
 require "stud/temporary"
+require "logstash/java_integration"
```
ph: Can we move that require with the other `logstash` one? So they are all together.

logstash-core/lib/logstash/logging.rb
```diff
@@ -1,91 +1,3 @@
...
-    properties.setProperty("log4j.appender.logstash.layout",
-                  "org.apache.log4j.PatternLayout")
-    properties.setProperty("log4j.appender.logstash.layout.conversionPattern",
-                  "log4j, [%d{yyyy-MM-dd}T%d{HH:mm:ss.SSS}] %5p: %c: %m%n")
-
-    org.apache.log4j.LogManager.resetConfiguration
-    org.apache.log4j.PropertyConfigurator.configure(properties)
-    logger.debug("log4j java properties setup", :log4j_level => log4j_level)
-  end
-end # class LogStash::Logger
```
ph: woohoo shinny red!

logstash-core/lib/logstash/logging/logger.rb
```diff
@@ -0,0 +1,72 @@
...
+        @logger.info(message, data)
+      end
+
+      def error(message, data = {})
+        @logger.error(message, data)
+      end
+
+      def fatal(message, data = {})
+        @logger.fatal(message, data)
+      end
```
ph: I remember that you mentioned that we could have `trace` but I don't see anything related to log4j2 traces method. If we don't support all the flow tracing we shouldn't mention that option in the settings?

logstash-core/lib/logstash/logging/logger.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      # Point logging at a specific path.
+      def self.configure_logging(path, level)
+        logging_ctx = LogManager.getContext(false)
+        config = logging_ctx.getConfiguration()
+        path ||= LogManager::ROOT_LOGGER_NAME
```
ph: this shorthand assignment will be only executed if the path send to the method is nil?
I suggest to change the signature of the method for this one:

```ruby
def self.configure_logging(level, path = nil)
#...
end
```

I think it's clearer that the path is not mandatory, and if no path is provided the level will change for the whole application. Also, the API will send an empty string if no module is provided it is that, valid/correct?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -144,7 +145,8 @@ def run
 
     start_workers
 
-    @logger.log("Pipeline #{@pipeline_id} started")
+    # TODO(talevy): figure out other way to do this
+    # @logger.log("Pipeline #{@pipeline_id} started")
```
ph: Why are theses lines commented with a todo?

logstash-core/lib/logstash/runner.rb
```diff
@@ -135,7 +135,7 @@ class LogStash::Runner < Clamp::StrictCommand
   attr_reader :agent
 
   def initialize(*args)
-    @logger = Cabin::Channel.get(LogStash)
+    @logger = self.class.logger
```
ph: This shouldn't be needed with the mixin?

logstash-core/lib/logstash/runner.rb
```diff
@@ -183,7 +186,7 @@ def execute
     end
 
     LogStash::ShutdownWatcher.unsafe_shutdown = setting("pipeline.unsafe_shutdown")
-    LogStash::ShutdownWatcher.logger = @logger
+    # LogStash::ShutdownWatcher.logger = @logger
```
ph: This can be removed

config/log4j2.properties
```diff
@@ -0,0 +1,36 @@
+status = error
+name = LogstashPropertiesConfig
+
+property.filename = /tmp/logstash/logstash.log
```
ph: I am not completely familiar with the **log4j** properties files,  but we already have a similar option in the settings.yml and we use them for the logger. Is that option necessary? It makes the assumption that this path exist (windows don't have this.). Same issue with the `rolling.filePattern` probably.

https://github.com/elastic/logstash/pull/5651/files#diff-da023432533015a65b7931c1c17d4c60R168

logstash-core/lib/logstash/plugin.rb
```diff
@@ -46,8 +45,8 @@ def eql?(other)
   end
 
   def initialize(params=nil)
+    @logger = self.class.logger
```
ph: The mixin should provide that instance variable?

logstash-core/lib/logstash/runner.rb
```diff
@@ -296,55 +300,6 @@ def create_agent(*args)
...
-    if setting("config.debug") && @logger.level != :debug
-      @logger.warn("--config.debug was specified, but log.level was not set to \'debug\'! No config info will be logged.")
-    end
-
-    # TODO(sissel): redirect stdout/stderr to the log as well
-    # http://jira.codehaus.org/browse/JRUBY-7003
-  end # def configure_logging
 
   # Emit a failure message and abort.
   def fail(message)
```
ph: ❤️

logstash-core/src/main/java/org/logstash/log/CustomLogEvent.java
```diff
@@ -0,0 +1,19 @@
...
+import org.apache.logging.log4j.Marker;
+import org.apache.logging.log4j.core.config.Property;
+import org.apache.logging.log4j.core.impl.Log4jLogEvent;
+import org.apache.logging.log4j.message.Message;
+
+import java.util.List;
+
+@JsonSerialize(using = CustomLogEventSerializer.class)
+public class CustomLogEvent extends Log4jLogEvent {
+    public CustomLogEvent(final String loggerName, final Marker marker, final String loggerFQCN, final Level level,
```
ph: Can we use the same variable name for the `loggerFQCN` and the the `fqcn` bellow? see https://github.com/elastic/logstash/pull/5651/files#diff-75539cd440190ae1a0fdc3bdc961cdc9R14

ph: Also this https://github.com/elastic/logstash/pull/5651/files#diff-53a87013a5635780f9539e309e57d107R20

ph:
@talevy I did a first pass looking great so far, before I test it manually I think the next would be:

- Revert the `symbols` changes
- Revert the `self` changes

Theses two things will greatly reduce the size of the PR.

logstash-core/lib/logstash/util/loggable.rb
ph: Maybe a unit test for it?

ph:
@talevy Another comment, can we add unit tests for the new java classes, like the json serializer etc ?

logstash-core/build.gradle
```diff
@@ -0,0 +1,73 @@
...
+idea {
+    module {
+        scopes.PROVIDED.plus += [project.configurations.provided]
+    }
+}
+
+
+dependencies {
+  compile 'org.apache.logging.log4j:log4j-api:2.6.1'
+  compile 'org.apache.logging.log4j:log4j-core:2.6.1'
```
fbaligand: It would be nice to switch to last log4j version : 2.6.2

logstash-core/src/test/java/org/logstash/log/ListAppender.java
```diff
@@ -0,0 +1,247 @@
...
+/**
+ * This appender is primarily used for testing. Use in a real environment is discouraged as the
+ * List could eventually grow to cause an OutOfMemoryError.
+ *
+ * This appender will use {@link Layout#toByteArray(LogEvent)}.
+ *
+ * see org.apache.logging.log4j.junit.LoggerContextRule#getListAppender(String) ILC.getListAppender
+ */
+@Plugin(name = "List", category = "Core", elementType = "appender", printObject = true)
+public class ListAppender extends AbstractAppender {
```
fbaligand: This class is a copy/paste from log4j-core test jar.
It would be nicer to remove this class `ListAppender`, and add this dependency in `build.gradle` :
`testCompile(group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.6.1', classifier: 'tests')`

Thus, it would let to make interesting unit tests on log4j2 integration, using this kind of code :
https://logging.apache.org/log4j/2.x/manual/configuration.html#UnitTestingInMaven

config/logstash.yml
```diff
@@ -98,13 +98,12 @@
 # ------------ Debugging Settings --------------
 #
 # Options for log.level:
-#   * warn => warn (default)
-#   * quiet => error
-#   * verbose => info
-#   * debug => debug
+#   * warn (default)
+#   * error
```
fbaligand: Maybe more relevant to inverse lines 'warn' and 'error' ?

logstash-core/lib/logstash/logging/logger.rb
```diff
@@ -0,0 +1,72 @@
...
+      def error(message, data = {})
+        @logger.error(message, data)
+      end
+
+      def fatal(message, data = {})
+        @logger.fatal(message, data)
+      end
+
+      # Point logging at a specific path.
+      def self.configure_logging(path, level)
```
fbaligand: Given log4j2 documentation (https://logging.apache.org/log4j/2.0/faq.html#reconfig_level_from_code), there is a simpler way to do that :
`Configurator.setLevel(path, Level.getLevel(level));`

Thus, `LogManager.getContext(false).getConfiguration()` doesn't exist in last `log4j-core` version.

logstash-core/src/test/java/org/logstash/log/ListAppender.java
talevy: good call, I wasn't sure whether it was published at the time

logstash-core/logstash-core.gemspec
talevy: the reason it is in `vendor/jars` is specifically since `require_jar` from the `jar-dependencies` library looks in that directory. Once we write our own, I will change this. I think I'll leave it as is for now

logstash-core/lib/logstash/config/mixin.rb
talevy: I'd prefer to be explicit here

logstash-core/lib/logstash/logging/logger.rb
talevy: you're right. I'll update

logstash-core/lib/logstash/runner.rb
talevy: true, if I convert all the other usages in `runner` to use the `self.logger`

logstash-core/lib/logstash/runner.rb
talevy: you're right

config/log4j2.properties
talevy: I will take a look at this for sure. thanks for bringing it up. I think the contents of `log4j2.properties` are up for debate and iteration anyways, so I'd like to leave concerns inside this file to a followup PR.

logstash-core/lib/logstash/plugin.rb
talevy: how can a mixin provide an instance variable? the mixin is included in the class once... not for every instance. unless I am missing something?

config/logstash.yml
talevy: I agree

logstash-core/lib/logstash/logging/logger.rb
talevy: oh cool! I wasn't aware of this

logstash-core/src/main/java/org/logstash/log/CustomLogEvent.java
talevy: I think I shall just remove it from being logged. it is not very useful in the context of Logstash

logstash-core/lib/logstash/api/rack_app.rb
talevy: will revert

logstash-core/lib/logstash/environment.rb
talevy: I am not sure what you mean

logstash-core/lib/logstash/logging/logger.rb
```diff
@@ -0,0 +1,68 @@
...
+      def fatal(message, data = {})
+        @logger.fatal(message, data)
+      end
+
+      def trace(message, data = {})
+        @logger.trace(message, data)
+      end
+
+      # Point logging at a specific path.
+      def self.configure_logging(level, path = "")
```
fbaligand: Maybe cleaner to put `path = LogManager::ROOT_LOGGER_NAME` ?

logstash-core/lib/logstash/environment.rb
ph: I mean all the logstash require should be together.
```
  require "logstash/settings"
  require "logstash/java_integration"
  require "stud/temporary"
  require "socket"
 ```

File: config/log4j2.properties (hunk 17 above)
Comment (ph) [inline config/log4j2.properties]: OK for a follow PR, would you mind creating the issue for tracking it?

File: logstash-core/logstash-core.gemspec (hunk 6 above)
Comment (ph) [inline logstash-core/logstash-core.gemspec]: OK, agree with that.

File: logstash-core/src/main/java/org/logstash/log/CustomLogEvent.java (hunk 20 above)
Comment (ph) [inline logstash-core/src/main/java/org/logstash/log/CustomLogEvent.java]: works for me too.

File: logstash-core/lib/logstash/plugin.rb (hunk 18 above)
Comment (ph) [inline logstash-core/lib/logstash/plugin.rb]: I took some time to look it into more details; Its is possible to a mixin to provide the instance variable but its really an hacky solution which requires doing an `alias_method` on the initialize. Lets no go that route @talevy and lets just keep the instance variable for now.

File: logstash-core/lib/logstash/runner.rb (hunk 16 above)
Comment (fbaligand) [inline logstash-core/lib/logstash/runner.rb]: Concerning logstash shutdown, it may be nice to call `LogManager.shutdown()` : this would let to flush all pending messages to logs and close nicely log files.

What you think ?

File: config/logstash.yml (hunk 23 above)
Comment (fbaligand) [inline config/logstash.yml]: Nice :)
In every conf/code listing all log levels, it would be nice to add `fatal` and `trace` levels, as you implemented these levels in `logger.rb`

File: logstash-core/src/test/java/org/logstash/log/ListAppender.java (hunk 22 above)
Comment (fbaligand) [inline logstash-core/src/test/java/org/logstash/log/ListAppender.java]: @talevy
Now that you updated `build.gradle`, could you remove `ListAppender` java class ?

File: config/logstash.yml [hunk 26]
```diff
@@ -98,13 +98,13 @@
...
 # Options for log.level:
-#   * warn => warn (default)
-#   * quiet => error
-#   * verbose => info
-#   * debug => debug
+#   * error
+#   * warn (default)
+#   * info
+#   * debug
+#   * trace
```
fbaligand: can you add also `fatal` level here ?

docs/static/command-line-flags.asciidoc
```diff
@@ -46,16 +46,13 @@ added[5.0.0-alpha3, Command-line flags have dots instead of dashes in their name
 *`--log.level LEVEL`*::
  Valid levels are:
 * `warn`: log warnings
-* `quiet`: log errors
-* `verbose`: log verbose info (for users)
+* `error`: log errors
+* `info`: log verbose info (for users)
 * `debug`: log debugging info (for developers)
```
fbaligand: can you add `fatal` and `trace` level here ?
can you inverse `warn` and `error` lines ?

docs/static/command-line-flags.asciidoc
```diff
@@ -46,16 +46,13 @@ added[5.0.0-alpha3, Command-line flags have dots instead of dashes in their name
...
 
-*`--log.format FORMAT`*::
- Set to "json" to log in JSON format, or "plain" (default) to use `Object#.inspect`.
- 
 *`--path.settings SETTINGS_DIR`*::
  Set the directory containing the `logstash.yml` <<logstash-settings-file,settings file>>.
```
fbaligand: It could be nice to update description here to say this folder contains also logging configuration file ?

logstash-core/build.gradle
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+
+dependencies {
+  compile 'org.apache.logging.log4j:log4j-api:2.6.2'
+  compile 'org.apache.logging.log4j:log4j-core:2.6.2'
+  compile 'com.fasterxml.jackson.core:jackson-core:2.7.4'
+  compile 'com.fasterxml.jackson.core:jackson-databind:2.7.4'
+  testCompile 'org.apache.logging.log4j:log4j-core:2.6.2:tests'
```
fbaligand: It would be nice to add also :
`testCompile 'org.apache.logging.log4j:log4j-api:2.6.2:tests'`
`testCompile 'junit:junit:4.12'`

It would let to make java unit tests.

logstash-core/lib/logstash/environment.rb
```diff
@@ -32,7 +33,7 @@ module Environment
                     Setting.new("path.plugins", Array, []),
             Setting::String.new("interactive", nil, false),
            Setting::Boolean.new("config.debug", false),
-            Setting::String.new("log.level", "warn", true, ["quiet", "verbose", "warn", "debug"]),
+            Setting::String.new("log.level", "warn", true, ["error", "warn", "debug", "info", "trace"]),
```
fbaligand: can you add here also `fatal` level ?

logstash-core/lib/logstash/logging/logger.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      def warn?
+        @logger.is_warn_enabled
+      end
+
+      def fatal?
+        @logger.is_fatal_enabled
+      end
+
+      def fatal?
```
fbaligand: copy/paste fault here :)
Method name is `fatal?`

fbaligand: copy/paste fault here :)
Method name is `trace?`

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -144,7 +145,7 @@ def run
 
     start_workers
 
-    @logger.log("Pipeline #{@pipeline_id} started")
+    puts "Pipeline #{@pipeline_id} started"
```
fbaligand: As a logstash user, I would appreciate to have this log at `info` level for example.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -159,7 +160,8 @@ def run
     shutdown_flusher
     shutdown_workers
 
-    @logger.log("Pipeline #{@pipeline_id} has been shutdown")
```
fbaligand: As a logstash user, I would appreciate to have this log at `info` level for example.

logstash-core/lib/logstash/runner.rb
```diff
@@ -148,25 +148,30 @@ def run(args)
...
     require "logstash/util"
     require "logstash/util/java_version"
     require "stud/task"
-    require "cabin" # gem 'cabin'
-    require "logstash/logging/json"
 
     # Configure Logstash logging facility, this need to be done before everything else to
     # make sure the logger has the correct settings and the log level is correctly defined.
-    configure_logging(setting("path.log"), setting("log.level"))
+    LogStash::Logging::Logger::configure_logging(setting("log.level"), setting("path.log"))
```
fbaligand: Hhmm, several things here :
- Presently, `configure_logging` method implementation updates log level for one logger (path).
- As confirmed by @ph, to set `path.log` and `log.level`, the right way is to set these settings as system properties using `System.setProperty()` and then reference them in config using `${sys:option}`. And this code has to be set at line 156, before initializing log4j.
- variable `@logger` is set before log4j is initialized by your code (at line 138). This implies log4j is initialized a first time using a default configuration (because no log4j file found). This default configuration will log on console with error root level.
- This implies warn logs will not be displayed (like line 151)

logstash-core/locales/en.yml
```diff
@@ -82,9 +82,9 @@ en:
     runner:
       short-help: |-
         usage:
-          bin/logstash -f CONFIG_PATH [-t] [-r] [--quiet|verbose|debug] [-w COUNT] [-l LOG]
-          bin/logstash -e CONFIG_STR [-t] [--quiet|verbose|debug] [-w COUNT] [-l LOG]
-          bin/logstash -i SHELL [--quiet|verbose|debug]
+          bin/logstash -f CONFIG_PATH [-t] [-r] [--error|info|debug|warn] [-w COUNT] [-l LOG]
+          bin/logstash -e CONFIG_STR [-t] [--error|info|debug|warn] [-w COUNT] [-l LOG]
+          bin/logstash -i SHELL [--error|info|debug|warn]
```
fbaligand: can you add `fatal` and `trace` levels ?
can you put `warn` level before `info` level ?

logstash-core/src/main/java/org/logstash/log/StructuredMessageContextSelector.java
fbaligand: @talevy 
What do you think about my previous comment ?

config/log4j2.properties
fbaligand: @talevy 
Why do you have replaced `log4j2.xml` by `log4j2.properties` ?
- XML is the preferred and most documented log4j2 format : look at the configuration examples here https://logging.apache.org/log4j/2.0/manual/appenders.html#RollingFileAppender
- XML format is much more concise and readable : log4j2 properties format is much more verbose with lots of redundant configuration, trying to make hierarchical configuration using flat properties.

Just an example with you rolling file appender in XML :
``` xml
<RollingFile name="RollingFile" fileName="${filename}" filePattern="/tmp/logstash/logstash-%d{MM-dd-yy-HH-mm-ss}-%i.log.gz"  >
    <Policies>
        <TimeBasedTriggeringPolicy interval="2" modulate="true" />
        <SizeBasedTriggeringPolicy size="100MB" />
    </Policies>
    <DefaultRolloverStrategy max ="5" />
    <PatternLayout pattern="[%d] %p %m (%c:%L)%n"/>
</RollingFile>
```

Isn't it more concise and readable ?

talevy: I would gladly discuss this after this PR makes it in.

It may make more sense for us to provide a very "dumb" config as a starter, and this may be done best in a `.properties` format. Likewise, it will keep consistency with the `.properties` that Elasticsearch will have in 5.0.

Since this is a generic config, I will make sure you are free to use your own custom `.xml` config

logstash-core/src/test/java/org/logstash/log/ListAppender.java
talevy: this was left since I have not finished writing any *actual* tests. will be gone in next commit

config/logstash.yml
```diff
@@ -98,13 +98,14 @@
...
+#   * fatal
```
fbaligand: Could you put `fatal` level just before `error` level ?

docs/static/command-line-flags.asciidoc
```diff
@@ -45,17 +45,17 @@ added[5.0.0-alpha3, Command-line flags have dots instead of dashes in their name
 
 *`--log.level LEVEL`*::
  Valid levels are:
+* `error`: log errors
+* `fatal`: log very severe error messages that will usually be followed by the application aborting
```
fbaligand: Can you put `fatal` level just before `error` level ?

logstash-core/build.gradle
```diff
@@ -0,0 +1,76 @@
...
+
+idea {
+    module {
+        scopes.PROVIDED.plus += [project.configurations.provided]
+    }
+}
+
+
+dependencies {
+  runtime 'org.apache.logging.log4j:log4j-1.2-api:2.6.2'
```
fbaligand: In the same idea, it would be nice to add also `slf4j-api` and `log4j-slf4j-impl` at `runtime` scope.
Because numerous java frameworks/libraries use slf4j to log.

logstash-core/lib/logstash/runner.rb
```diff
@@ -148,25 +147,29 @@ def run(args)
...
     rescue => e
-      @logger.subscribe(STDOUT)
-      @logger.warn("Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in \"--path.settings=/path/..\" in the command line options")
-      @logger.fatal("Failed to load settings file from \"path.settings\". Aborting...", "path.settings" => LogStash::SETTINGS.get("path.settings"), "exception" => e.class, "message" => e.message)
+      logger.warn("Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in \"--path.settings=/path/..\" in the command line options")
+      logger.fatal("Failed to load settings file from \"path.settings\". Aborting...", "path.settings" => LogStash::SETTINGS.get("path.settings"), "exception" => e.class, "message" => e.message)
```
fbaligand: `logger` variable is not defined here => should raise an error at runtime.
As it is a blocking error that implies logstash exit, maybe a simple `puts` is the solution (because log4j is not yet initialized).
What you think ?

logstash-core/locales/en.yml
```diff
@@ -82,10 +82,10 @@ en:
...
       short-help: |-
         usage:
-          bin/logstash -f CONFIG_PATH [-t] [-r] [--quiet|verbose|debug] [-w COUNT] [-l LOG]
-          bin/logstash -e CONFIG_STR [-t] [--quiet|verbose|debug] [-w COUNT] [-l LOG]
-          bin/logstash -i SHELL [--quiet|verbose|debug]
-          bin/logstash -V [--verbose|debug]
+          bin/logstash -f CONFIG_PATH [-t] [-r] [] [-w COUNT] [-l LOG]
+          bin/logstash -e CONFIG_STR [-t] [--log.level fatal|error|warn|info|debug] [-w COUNT] [-l LOG]
+          bin/logstash -i SHELL [--log.level fatal|error|warn|info|debug]
+          bin/logstash -V [--log.level fatal|error|warn|info|debug]
```
fbaligand: can you add `trace` level on these 4 lines ?

talevy: I'll never get this right 😄

logstash-core/lib/logstash/logging/logger.rb
fbaligand: @talevy
Could you fix it ?

config/log4j2.properties
fbaligand: I just downloaded elasticsearch 5.0 alpha5, and format is still custom `yaml` format and log4j 1.x is still used (it's like that since elasticsearch 1.x).  
I checked : no branch or PR concerning migration to log4j 2.x is existing.  

Concerning "dumb config as a starter", in log4j 1.x, properties format was really easy for a start.  
But in log4j 2.x, things have changed : the properties format is now more complex and verbose to let the same hierarchical capacities than xml format.
I especially noticed that you have copied/pasted the only log4j2 properties example as it is :)

suyograo: @fbaligand ES will be moving to `.properties`. I don't think it's happening in the 5.0 timeframe, but it is definitely on the cards to move to `.properties` _and_ to log4j2. It would make sense for LS to be future proof :). 

I am +1 on allowing logf4j xml config for advanced users, but for most `.properties` should suffice.

bin/logstash.lib.sh
```diff
@@ -131,13 +131,19 @@ setup_ruby() {
   VENDORED_JRUBY=
 }
 
+logging_opts() {
+    printf "%s" " -J-DLog4jLogEventFactory=org.logstash.log.CustomLogEventFactory"
+    printf "%s" " -J-DLog4jContextSelector=org.logstash.log.StructuredMessageContextSelector"
```
fbaligand: I recommend you to move these 2 properties to `src/main/resources/log4j2.component.properties` because :
- it will avoid you to copy this configuration in `setup.bat` (for windows startup)
- it will let you remove class LogEventFactoryRule where you define a third time these properties

So, doing that, you avoid to repeat yourself 3 times ! => DRY POWER ! :)

config/log4j2.properties
suyograo: IMO, that should be a concern of a future PR (to support `.xml`)

fbaligand: OK, if elasticsearch will do so, I surrender :)
I'm however surprised by this choice, since properties format was firstly abandoned by log4j2 and has been finally added in log4j 2.4, using a more complex format than log4j 1.x

Anyway, concerning support for another log4j2 config file, it requires just 2 lines of code :)

logstash-core/lib/logstash/runner.rb
```diff
@@ -19,9 +18,10 @@
 require "logstash/version"
 
 class LogStash::Runner < Clamp::StrictCommand
+  include LogStash::Util::Loggable
```
fbaligand: With this include, given that `Loggable` module creates a `Logger` instance, it implies a log4j initialization before log4j config file is defined.
I suggest to not include Loggable here but after log4j config file has been precised (still in runner.rb).

logstash-core/lib/logstash/runner.rb
```diff
@@ -148,25 +147,29 @@ def run(args)
...
+      logger.warn("Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in \"--path.settings=/path/..\" in the command line options")
+      logger.fatal("Failed to load settings file from \"path.settings\". Aborting...", "path.settings" => LogStash::SETTINGS.get("path.settings"), "exception" => e.class, "message" => e.message)
       exit(-1)
     end
 
+    # Configure Logstash logging facility, this need to be done before everything else to
+    # make sure the logger has the correct settings and the log level is correctly defined.
+    # TODO(talevy): cleanly support `path.log` setting in log4j
+    log4j_config_location = setting("path.settings") + "/log4j2.properties"
+    LogStash::Logging::Logger::initialize(log4j_config_location)
```
fbaligand: I think it would be great to surround these 2 lines with a `if` test that checks if `log4j.configurationFile` system property is not present (using `System.getProperty` method).
With this simple code, it enables 2 interesting things :
- logstash user can define a custom log4j config location
- logstash user can choose another log4j config format : xml, yaml or json.

logstash-core/lib/logstash/runner.rb
```diff
@@ -148,25 +147,29 @@ def run(args)
...
+    LogStash::Logging::Logger::configure_logging(setting("log.level"))
```
fbaligand: I think we can do better on this line, solving by this way also `path.log` problem :
- either you add `log.level` and `path.log` as system properties (using `System.setProperty()`) just before log4j initialization.  
And then, you reference them in `log4j2.properties` using `${sys:log.level}` and `${sys:path.log}` 
- either you define a custom log4j2 `Lookup` implementation where you inject logstash settings using `LogstashLookup.setSettings(Logstash::SETTINGS.to_hash)`) just before log4j initialization.  
And then, you reference them in `log4j2.properties` using `${logstash:log.level}` and `${logstash:path.log}`

logstash-core/lib/logstash/runner.rb
talevy: I understand that it is not yet initialized... but the logging functionality inside is lazy, and only gets executed when calling from an instance. So it is not an issue

logstash-core/lib/logstash/runner.rb
talevy: Currently, this is how this branch gets printed to the user. I am not necessarily annoyed by the first message about log4j, unless you think it should not be happening. I don't mind either way

```
15:38:06.083 [main] FATAL logstash.runner - Failed to load settings file from "path.settings". Aborting... {"path.settings"=>"joids", "exception"=>Errno::ENOENT, "message"=>"No such file or directory - /Users/tal/Code/es/talevy-logstash/joids/logstash.yml"}
```

logstash-core/lib/logstash/runner.rb
talevy: @fbaligand have you actually tried running Logstash from this branch? I was experiencing some strange behavior where I was not able to update the configuration using the `Configurator` if I used the system properties option. That is why I introduced the `initialize`.

you are absolutely right though, I will update to check if the system property already exists. We should respect a user's system property

logstash-core/lib/logstash/runner.rb
talevy: > Either you add log.level and path.log as system properties (using System.setProperty()) just before log4j initialization.

I would like to avoid system properties

> Either you define a custom log4j2 Lookup implementation where you inject logstash settings

This would be a nice thing to do for initial loading of the configuration, but custom log4j configs that may not use / understand these lookups will not see `log.level` respected upon loading. I understand we can inform users in the docs, but it feels like a simple thing to ignore.

logstash-core/build.gradle
talevy: I will open an issue around this. I am weary of adding these jars so that they get loaded for everyone, even though they are not using plugins that have `slf4j`. This worry is also applied to the `log4j-1.2-api`, maybe this responsibility should be shifted to that of the plugins

logstash-core/lib/logstash/runner.rb
talevy: actually. I changed my mind 😄. I'm moving this to `$stderr.puts`

bin/logstash.lib.sh
talevy: I used to have this, but then was worried this would conflict with other things... but I guess I override it anyways and others can do the same by added extra arguments when running LS

logstash-core/lib/logstash/pipeline.rb
jordansissel: The reason for `log` (no level) here instead of `info` is because there are certain kinds of events in Logstash's life that has caused confusion or frustration for users. I understand log4j2 has no concept of a log with no "level", though. I guess the only equivalent is `info` (since it is not conceptually a warning/error).

logstash-core/lib/logstash/runner.rb
```diff
@@ -148,25 +148,32 @@ def run(args)
...
     rescue => e
-      @logger.subscribe(STDOUT)
-      @logger.warn("Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in \"--path.settings=/path/..\" in the command line options")
-      @logger.fatal("Failed to load settings file from \"path.settings\". Aborting...", "path.settings" => LogStash::SETTINGS.get("path.settings"), "exception" => e.class, "message" => e.message)
+      $stderr.puts "ERROR: Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in \"--path.settings=/path/..\" in the command line options"
+      $stderr.puts "ERROR: Failed to load settings file from \"path.settings\". Aborting... path.setting=#{LogStash::SETTINGS.get("path.settings")}, exception=#{e.class}, message=>#{e.message}"
```
fbaligand: A little word error here : it is `path.settings` and not `path.setting`.  
Thus, I think you should use simple quote (`'`) here : `LogStash::SETTINGS.get("path.settings")` because your whole string is surrounded by double quotes.

logstash-core/lib/logstash/runner.rb
```diff
@@ -148,25 +148,32 @@ def run(args)
...
       exit(-1)
     end
 
+    # Configure Logstash logging facility, this need to be done before everything else to
+    # make sure the logger has the correct settings and the log level is correctly defined.
+    # TODO(talevy): cleanly support `path.logs` setting in log4j
```
fbaligand: Everywhere else, it is indicated `path.log`.
So either you update everywhere else to `path.logs`, either you get back here to `path.log` :)

logstash-core/lib/logstash/runner.rb
jordansissel: @fbaligand Quotes in `#{...}` are safe.

Example:

```ruby
>> puts "#{"hello world"}"
hello world
>> a = { "hello" => "world" }
=> {"hello"=>"world"}
>> puts "hello #{a["hello"]}"
hello world
```

No quoting change needed.

logstash-core/lib/logstash/runner.rb
```diff
@@ -148,25 +148,32 @@ def run(args)
     begin
       LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("path.settings"))
     rescue => e
-      @logger.subscribe(STDOUT)
-      @logger.warn("Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in \"--path.settings=/path/..\" in the command line options")
-      @logger.fatal("Failed to load settings file from \"path.settings\". Aborting...", "path.settings" => LogStash::SETTINGS.get("path.settings"), "exception" => e.class, "message" => e.message)
+      $stderr.puts "ERROR: Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in \"--path.settings=/path/..\" in the command line options"
```
jordansissel: This message is informational, it is not an error.

logstash-core/lib/logstash/runner.rb
```diff
@@ -148,25 +148,32 @@ def run(args)
...
       exit(-1)
     end
 
+    # Configure Logstash logging facility, this need to be done before everything else to
+    # make sure the logger has the correct settings and the log level is correctly defined.
+    # TODO(talevy): cleanly support `path.logs` setting in log4j
+    log4j_config_location = setting("path.settings") + "/log4j2.properties"
+    unless java.lang.System.getProperty("log4j.configurationFile")
+      LogStash::Logging::Logger::initialize(log4j_config_location)
+    end
```
fbaligand: Nice ! :)
I think that line `log4j_config_location = setting("path.settings") + "/log4j2.properties"` should also move in `unless` block.

logstash-core/lib/logstash/runner.rb
jordansissel: Let's use the locale stuff to format this and put the format of the text in `locales/en.yml`.

logstash-core/lib/logstash/runner.rb
jordansissel: Let's put this message in locales/en.yml also

logstash-core/lib/logstash/runner.rb
talevy: I know. there was some discussion on moving this to be `path.logs` to specify the directory for where the logs will go. instead of one log file

jordansissel:
`rake bootstrap` from this PR fails for me right now:

```
% rake bootstrap
... 
Building logstash-core-event-java using gradle
Error: Could not find or load main class org.gradle.wrapper.GradleWrapperMain
```

talevy:
@jordansissel interesting... I didn't realize I didn't push things up correctly. probably missing a moved jar. I'll take a look. 

I moved the gradle wrapper from core-event-java to the parent directory

logstash-core/lib/logstash/runner.rb
talevy: right, I guess I instinctually prefixed with ERROR since it was a message right before a fatal exit.

talevy:
latest commit pushes the jar up. jars are in the `.gitignore`. :(

jordansissel:
Curious - why are we committing jars to source control? Jars can be compiled from source.

jordansissel:
Regarding my previous comment about jars, I filed https://github.com/elastic/logstash/issues/5791 about removing them. No action needed in this PR for the jars.

jordansissel:
BTW, I've been sitting back on reviewing this mostly because of the incredible work by @fbaligand in reviewing this change. I am very grateful for this review work :)

Y'all are awesome. <3

logstash-core/lib/logstash/runner.rb
fbaligand: Oops, my mistake :)

fbaligand:
Thanks @jordansissel !
Happy to see you appreciate my review work ! :)

logstash-core/lib/logstash/runner.rb
fbaligand: Ok, this would let to be coherent with elasticsearch.

logstash-core/src/main/resources/log4j2.component.properties
```diff
@@ -0,0 +1,2 @@
+Log4jLogEventFactory=org.logstash.log.CustomLogEventFactory
+Log4jContextSelector=org.logstash.log.StructuredMessageContextSelector
```
fbaligand: :+1:

logstash-core/lib/logstash/runner.rb
fbaligand: > @fbaligand have you actually tried running Logstash from this branch? I was experiencing some strange behavior where I was not able to update the configuration using the Configurator if I used the system properties option. That is why I introduced the initialize.

I have not tries running Logstash from this branch.
At the end, normally, we won't need to use Configurator to update configuration, given that we inject `log.level` and `path.log` as `Lookup` entries.
Do you agree with that ?

logstash-core/lib/logstash/runner.rb
```diff
@@ -148,25 +148,32 @@ def run(args)
...
     require "logstash/util"
     require "logstash/util/java_version"
     require "stud/task"
-    require "cabin" # gem 'cabin'
-    require "logstash/logging/json"
 
-    # Configure Logstash logging facility, this need to be done before everything else to
-    # make sure the logger has the correct settings and the log level is correctly defined.
-    configure_logging(setting("path.log"), setting("log.level"))
+    LogStash::Logging::Logger::configure_logging(setting("log.level"))
```
fbaligand: > @fbaligand : Either you define a custom log4j2 Lookup implementation where you inject logstash settings  
> @talevy : This would be a nice thing to do for initial loading of the configuration, but custom log4j configs that may not use / understand these lookups will not see log.level respected upon loading. I understand we can inform users in the docs, but it feels like a simple thing to ignore.

- firstly, I think that very few people will make a custom log4j configuration (maybe only me :p). I mean : define a totally different file, not just customizing some logger levels in provided log4j configuration.
- secondly, people that make custom log4j config would certainly take inspiration from provided `log4j2.properties`, see the `Lookup` references, and so, understand the idea. => I don't think it "requires" a documentation.
- thirdly, for `path.log` (or `path.logs`), at my knowledge, you have no other way than `Lookup` to do that
- at last, this is what elasticsearch uses for its log4j configuration

So up to me, go for Lookup ! :)

logstash-core/lib/logstash/runner.rb
fbaligand: To inject `path.log` and `log.level` as log4j2 `Lookup`, I think we could do it easily like this :
- create a java class like this : [LogstashLookup.java](https://gist.github.com/fbaligand/bedeb87a18b876f1d057d43ee171e05c)
- just before log4j initialization, call `LogstashLookup.setLogstashSettings(Logstash::SETTINGS.to_hash)`
- in `log4j2.properties`, add references `${logstash:path.log}` and `${logstash:log.level}`

Great, no ? :)

logstash-core/build.gradle
fbaligand: Here's my opinion about it :
- in logstash, logging system and configuration is common and shared with everybody : core code and plugins (contrary to an applications server for example)
- manage this shared logging system and its integration is logstash's responsibility, and not each plugin's responsibility.
- if it's not done, some plugin logs will be lost, some others will arrive anywhere.
- in particular for slf4j, it's important that slf4j "api" and "impl" versions are aligned. And logstash-core is the best place to ensure that.

=> To me, it is relevant to put these jars here.

@ph @suyograo @jordansissel 
Any opinion on that subject ?

logstash-core/lib/logstash/runner.rb
talevy: I will leave the config as a dumb config for now. will follow-up with the lookup work in another PR

logstash-core/lib/logstash/runner.rb
talevy: I agree with your points. I prefer to followup with the Lookups in a follow-up PR

talevy:
@fbaligand just wanted to say thank you so much for your comments! been a tremendous help. I was not familiar with much about Log4j before this.

talevy:
tests pass now. didn't even notice. nice!

fbaligand:
@talevy 
You're welcome :)
It has been a pleasure to work on this PR with you !

logstash-core/lib/logstash/runner.rb
fbaligand: OK.
OK.
As I worked a little on this, I just give you as I see the thing.

To inject `path.log` and `log.level` as log4j2 `Lookup`, I think we could do it easily like this :
- create a java class like this : [LogstashLookup.java](https://gist.github.com/fbaligand/bedeb87a18b876f1d057d43ee171e05c)
- just before log4j initialization, call `LogstashLookup.setLogstashSettings(Logstash::SETTINGS.to_hash)`
- in `log4j2.properties`, add references `${logstash:path.log}` and `${logstash:log.level}`

Great, no ? :)

logstash-core/src/test/java/org/logstash/log/CustomLogEventTests.java
```diff
@@ -0,0 +1,86 @@
...
+        assertEquals(5, secondMessage.size());
+        assertEquals("WARN", secondMessage.get("level"));
+        assertEquals("JSONEventLogger", secondMessage.get("loggerName"));
+        assertEquals("main", secondMessage.get("thread"));
+        Map<String, Object> logEvent = new HashMap<>();
+        logEvent.put("message", "complex message");
+        logEvent.put("foo", "bar");
+        assertEquals(logEvent, secondMessage.get("logEvent"));
+    }
+}
```
ph: +1 for theses

logstash-core/lib/logstash/api/modules/logging.rb
```diff
@@ -0,0 +1,22 @@
...
+          else
+            LogStash::Logging::Logger::configure_logging(level, path)
+            respond_with({"acknowledged" => true})
+          end
+        end
+
+      end
+    end
+  end
+end
```
ph: Can we add a test for this rest endpoint I don't see any?

logstash-core/lib/logstash/api/modules/logging.rb
```diff
@@ -0,0 +1,22 @@
+# encoding: utf-8
+module LogStash
+  module Api
+    module Modules
+      class Logging < ::LogStash::Api::Modules::Base
+
+        put "/" do
+          level = params["log.level"]
+          path = params["module"]
+          if level.nil?
```
ph:
```ruby
level.nil? || level.empty?
```

level can be present in the hash but with an empty string as the value.

logstash-core/lib/logstash/api/modules/logging.rb
ph: we will need to document that feature.

ph:
I have tested this PR with the `logstash-input-beats`, this plugin is mostly written in Java and it uses log4j2 as the logging mechanism. When I tried to use the plugin with this PR I get the following exception:

```
[2016-08-24T15:33:30,569][ERROR][logstash.pipeline        ] A plugin had an unrecoverable error. Will restart this plugin.
  Plugin: <LogStash::Inputs::Beats port=>5044, id=>"c68d509166089e1ac514c2a813ab45d5e0db306a-1", enable_metric=>true, codec=><LogStash::Codecs::Plain id=>"plain_b7cc8afd-762e-4cfa-9c7b-c39c35a88d96", enable_metric=>true, charset=>"UTF-8">, host=>"0.0.0.0", ssl=>false, ssl_verify_mode=>"none", include_codec_tag=>true, ssl_handshake_timeout=>10000, congestion_threshold=>5, target_field_for_codec=>"message", tls_min_version=>1, tls_max_version=>1.2, cipher_suites=>["TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA38", "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384", "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384", "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384", "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256"], client_inactivity_timeout=>15>
  Error: must log key-value pairs
  Exception: Java::JavaLang::IllegalArgumentException
  Stack: org.logstash.log.StructuredMessage.<init>(org/logstash/log/StructuredMessage.java:31)
org.logstash.log.StructuredMessageFactory.newMessage(org/logstash/log/StructuredMessageFactory.java:23)
org.apache.logging.log4j.spi.MessageFactory2Adapter.newMessage(org/apache/logging/log4j/spi/MessageFactory2Adapter.java:49)
org.apache.logging.log4j.spi.AbstractLogger.logMessage(org/apache/logging/log4j/spi/AbstractLogger.java:2004)
org.apache.logging.log4j.spi.AbstractLogger.logIfEnabled(org/apache/logging/log4j/spi/AbstractLogger.java:1876)
org.apache.logging.log4j.spi.AbstractLogger.info(org/apache/logging/log4j/spi/AbstractLogger.java:1421)
org.logstash.beats.Server.listen(org/logstash/beats/Server.java:59)
java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:483)
RUBY.run(/Users/ph/es/logstash-plugins/logstash-input-beats/lib/logstash/inputs/beats.rb:177)
RUBY.inputworker(/tmp/logstash/logstash-core/lib/logstash/pipeline.rb:336)
```

if you look at the beats code this is due of the usage of string interpolation. see https://github.com/logstash-plugins/logstash-input-beats/blob/master/src/main/java/org/logstash/beats/Server.java#L59
```ruby
  logger.info("Starting server on port: {}", this.port);
```

We love structured log, since I control this library I can rewrite all the statements using the `StructuredMessage`, but this won't be the case for all the other library we will be using, BTW also I am passing our logger to the netty frameworks to get more detailled information about SSL error/Problems with frames.

@talevy Can we find an elegant solution for that?

ph:
https://github.com/elastic/logstash/pull/4820 introduced `--log-in-json` which allow us to write log in the json format.

The refactor of the runner to use the new settings class changes that for ``--log.format json``

Using this command with the current PR doesnt seems to do anything? Can we add back the support?

```
bin/logstash -f ~/es/logstash/ls_tools/beats.conf --log.format json
```

ph:
Other than the 2 previous comment, we are really close.

suyograo:
@ph @talevy how about restoring `--log-in-json` in a subsequent PR?

ph:
@suyograo I am OK to have another PR to add the --log-in-json.

logstash-core/src/main/java/org/logstash/log/StructuredMessageFactory.java
```diff
@@ -0,0 +1,33 @@
...
+        return new StructuredMessage(message);
+    }
+
+    @Override
+    public Message newMessage(String message, Object... params) {
+        if (params.length == 1 && params[0] instanceof Map) {
+            return new StructuredMessage(message, params);
+        } else {
+            return new ParameterizedMessage(message, params);
+        }
```
fbaligand: Now that the returned `Message` instance is not always a `StructuredMessage`, it's maybe relevant to rename this class to something like `LogstashMessageFactory` ?

logstash-core/src/main/java/org/logstash/log/StructuredMessageContextSelector.java
fbaligand: If I understand well, the only reason why you create a custom ContextSelector, is to define a custom default `MessageFactory`, right ?

In this case, it exists a specific log4j2 property to customize default `MessageFactory` : `log4j2.messageFactory`  
Using this property, you don't need to create a custom ContextSelector (which aim is to select current log4j context).  

I think it's better to customize only what is necessary, especially concerning log4j2 internals.

talevy: I will try and remove that and do this. To be honest, I was still understanding what I needed to change and tried to change as little as possible. thanks.

The only reason I have anything customized is because the JSONLayout doesn't give control over how to serialize the `Message`.

logstash-core/src/main/java/org/logstash/log/StructuredMessageFactory.java
talevy: yes. I'll update

logstash-core/src/main/resources/log4j2.component.properties
```diff
@@ -0,0 +1,2 @@
+Log4jLogEventFactory=org.logstash.log.LogstashLogEventFactory
+log4j2.messageFactory=org.logstash.log.LogstashMessageFactory
```
fbaligand: Nice ! :+1:

logstash-core/src/test/java/org/logstash/log/LogstashMessageContextSelector.java
```diff
@@ -0,0 +1,33 @@
...
+        protected Logger newInstance(final LoggerContext ctx, final String name, MessageFactory messageFactory)
+        {
+            if (null == messageFactory)
+                messageFactory = LogstashMessageFactory.INSTANCE;
+
+            return super.newInstance(ctx, name, messageFactory);
+        }
+
+    }
+}
```
fbaligand: Why did you move this class to `src/test/java` instead of remove it ?

talevy: it is being used by the `LoggerContextRule` for capturing the logs into the ListAppender.

```java
@ClassRule
public static LoggerContextRule CTX = new LoggerContextRule(CONFIG, LogstashMessageContextSelector.class);
```

talevy: I will see how this works if I remove it

talevy: guess it does work.. removed

fbaligand: Indeed. As it works for "main" use case, it works also for "test" use case.
=> Nice :+1:

fbaligand:
@talevy Nice remove :+1:

ph:
@talevy @fbaligand I've gone through the last iteration of this PR and it fixed all the issues we had with string interpolation.

The only thing I've noticed is the unchecked warning, we might want to suppress that warning.

``
:compileJavaNote: /private/tmp/logstash/logstash-core/src/main/java/org/logstash/log/StructuredMessage.java uses unchecked or unsafe operations.
```

ph:
Fix the unchecked warning and it LGTM. Lets get this into master and continue on the other tasks!
https://github.com/elastic/logstash/issues/5650


Amazing work @talevy!

talevy:
thanks. I added a `SuppressWarning` for the cast that was going on. should disappear now

ph:
LGTM

talevy:
finally! thanks @fbaligand, @ph, and @jordansissel for the reviews! I will follow-up with some items left over now. the meta issue should have captured all the points in checkpoints and/or PRs: https://github.com/elastic/logstash/issues/5650.

PR #5656, opened 2016-07-20
Implement deep-clone properly with metadata

### LINKED ISSUE #4640: `LogStash::Event#clone` should also clone the `@metadata`

suyograo:
Fixes #4640

### DIFF
--- a/logstash-core-event/lib/logstash/event.rb
+++ b/logstash-core-event/lib/logstash/event.rb
@@ -92,13 +92,8 @@ class LogStash::Event
 
   # Create a deep-ish copy of this event.
   def clone
-    copy = {}
-    @data.each do |k,v|
-      # TODO(sissel): Recurse if this is a hash/array?
-      copy[k] = begin v.clone rescue v end
-    end
-
-    self.class.new(copy)
+    copy = LogStash::Util.deep_clone(@data)
+    self.class.new(copy.merge({ "@metadata" => LogStash::Util.deep_clone(@metadata.clone) }))
   end
 
   def to_s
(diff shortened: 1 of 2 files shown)


logstash-core-event/lib/logstash/event.rb
```diff
@@ -92,13 +92,8 @@ def cancelled?
...
   def clone
-    copy = {}
-    @data.each do |k,v|
-      # TODO(sissel): Recurse if this is a hash/array?
-      copy[k] = begin v.clone rescue v end
-    end
-
-    self.class.new(copy)
+    copy = LogStash::Util.deep_clone(@data)
+    self.class.new(copy.merge({ "@metadata" => LogStash::Util.deep_clone(@metadata.clone) }))
```
ph: I believe the `@metadata.clone` is unnecessary since deep_clone will take care of all the values.

logstash-core-event/spec/logstash/event_spec.rb
```diff
@@ -658,4 +658,47 @@
...
+      expect(cloned.to_hash_with_metadata).to include("@metadata")
+    end
+
+    it "should clone metadata fields with multiple keys" do
+      cloned = event5.clone
+      expect(cloned.get(fieldref)).to eq("pants")
+      expect(cloned.get("[@metadata][smarty]")).to eq("pants2")
+      expect(cloned.to_hash_with_metadata).to include("@metadata")
+    end
+  end
```
ph: I see we have tests that check if the `cloned` contains the right values but we don't have a test that check for mutation problem, I think we should add that notion to the suite.

Maybe a scenario like this:

1. clone the event
2. mutate value in the cloned event
3. check that the original value were not modified.

WDYT?

ph:
LGTM

jordansissel:
I have removed the 2.4.0 label from this PR because it was only merged into 2.x, not 2.4

PR #5667, opened 2016-07-20
Add bin/rspec and bin/rspec.bat to the packaging exclude list

### LINKED ISSUE #5666: bin/rspec is still shipped in artifacts
Since a few versions ago tests are not shipped anymore within LS, but we still ship the `bin/rspec` and collateral for windows (`bin/rspec.bat`). This opens the door to users trying to test their configurations with this command. 

This situation is not ideal as: 
* the distributed packages for example does not include  development dependencies.
* there is no other test included in the artifacts.
* users might be tempted to test a full configuration, not just filters sections.

Testing a configuration can be, probably easier done, by simply calling the `bin/logstash` command and thinking about LS as a black box.

purbon:
To test this PR is necessary to build a package and see this files are not there anymore. Reasoning why this should be removed can be found at #5666.

Fixes #5666

### DIFF
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -46,6 +46,8 @@ namespace "artifact" do
     @exclude_paths << "**/test/files/slow-xpath.xml"
     @exclude_paths << "**/logstash-*/spec"
     @exclude_paths << "bin/bundle"
+    @exclude_paths << "bin/rspec"
+    @exclude_paths << "bin/rspec.bat"
 
     @exclude_paths
   end


purbon:
@ph or @jordansissel your thoughts here are very welcome.

suyograo:
This is a breaking change, but I agree with the reasoning. LGTM

suyograo:
+1 on creating an issue. Will do so.

jsvd:
cherry-picked this into 5.0 as 2c25592

PR #5677, opened 2016-07-22
[WIP] add simulate flag to replace all inputs and outputs with stdin/stdout

talevy:
Sometimes it would be nice to leave an existing config the same, but just to test the pipeline filter behavior. Right now we comment out certain inputs and outputs and replace them with stdin/stdout sometimes to gain insight into the effects of the pipeline on the input. It would be nice to just replace all inputs and outputs with stdin/stdout so we can test pipeline control flow and such without touching the existing pipeline configuration.

### DIFF
--- a/logstash-core/lib/logstash/environment.rb
+++ b/logstash-core/lib/logstash/environment.rb
@@ -24,6 +24,7 @@ module LogStash
                     Setting.new("path.plugins", Array, []),
             Setting::String.new("interactive", nil, false),
            Setting::Boolean.new("config.debug", false),
+           Setting::Boolean.new("config.simulate", false),
             Setting::String.new("log.level", "warn", true, ["quiet", "verbose", "warn", "debug"]),
            Setting::Boolean.new("version", false),
            Setting::Boolean.new("help", false),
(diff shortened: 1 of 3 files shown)


logstash-core/lib/logstash/environment.rb
```diff
@@ -24,6 +24,7 @@ module LogStash
                     Setting.new("path.plugins", Array, []),
             Setting::String.new("interactive", nil, false),
            Setting::Boolean.new("config.debug", false),
+           Setting::Boolean.new("simulate", false),
```
suyograo: can we call it `config.simulate`?

suyograo:
Big +1 on this.

logstash-core/lib/logstash/environment.rb
talevy: yup

jordansissel:
I like the idea :)

Code looks good. I haven't tested. One suggestion so far: add a notification to the user to say explicitly that simulation is enabled, when enabled.

talevy:
Closing in favor of waiting for the work with the LIR to get merged in

PR #5685, opened 2016-07-25
Multiple Plugins in a Configuration File section

tgroshon:
From the flow of the wording, you mean "plugins" not "filters". And that is an important distinction.

### DIFF
--- a/docs/static/configuration.asciidoc
+++ b/docs/static/configuration.asciidoc
@@ -49,7 +49,7 @@ output {
 ----------------------------------
 
 Each section contains the configuration options for one or more plugins. If you specify
-multiple filters, they are applied in the order of their appearance in the configuration file.
+multiple plugins, they are applied in the order of their appearance in the configuration file.
 
 
 [float]


suyograo:
We actually did mean `multiple filters`. Filters are executed in sequence defined by their order in config files.

tgroshon:
Ah, that's great.  So is order important for in `input` or `output` sections?  Like specifying output plugins for S3 and Elasticsearch.  Will S3 being placed before elasticsearch ensure that data goes to S3 if Elasticsearch crashes?  Does it depend on the plugin implementation?

untergeek:
> So is order important in input or output sections?

No.  Without conditionals, all inputs would go to all outputs in parallel.

untergeek:
> Will S3 being placed before elasticsearch ensure that data goes to S3 if Elasticsearch plugin crashes? Or does it depend on the plugin implementation?

No.  Because all data flows to all outputs, if one stream gets blocked (due to filter slowness or input), then _all_ outputs are blocked until the pipeline is clear.

tgroshon:
Thanks for the clarity!

PR #5688, opened 2016-07-26
edit 2.3.1 relnotes entry

dedemorton:
You can't tell very easily from the diff, but there were a few places where words were missing: 

- "The new pure Java implementation of the Event class Logstash 2.3.1 lightning fast" - (needs a proposition and verb to be grammatically correct. I also thought it was confusing to mention that the Java Event implementation is fast in Logstash 2.3.1--maybe you mean to say 2.3.0. I think the version is implicit, so I just removed it).

- "the Java version remains the correct technical direction and we will most likely be reintroducing in Logstash 5.0." (was missing "it")


suyograo:
LGTM

PR #5690, opened 2016-07-26
Update plugin manager --path.plugins

paurullan:
The plugin manager `--path.plugins` (old --pluginpath) is completely wrong in the documentation. It cost me FOUR hours to find the bug and I hope this fix will remove pain for other developers.

This change should be backported to the current documentation.


paurullan:
I signed the CLA under APSL corporation and paurullan user.

suyograo:
confirmed CLA. Thanks for fixing this!

PR #5692, opened 2016-07-26
Doc: quote shell glob pattern

7lima:
... otherwise it may pick up some real file name in the current working directory - which is not what's intended!


suyograo:
LGTM

7lima:
Thank you, suyograo! ;-)

PR #5696, opened 2016-07-27
Doc: Rename "Event Dependent Configuration" Chapter.

7lima:
The current chapter title does not reflect the contents so it's hard to find it again later on when you're looking for info about accessing fields.

### DIFF
--- a/docs/static/configuration.asciidoc
+++ b/docs/static/configuration.asciidoc
@@ -278,7 +278,7 @@ input { # comments can appear at the end of a line, too
 ----------------------------------
 
 [[event-dependent-configuration]]
-=== Event Dependent Configuration
+=== Accessing Event Data and Fields
 
 The logstash agent is a processing pipeline with 3 stages: inputs -> filters ->
 outputs. Inputs generate events, filters modify them, outputs ship them


docs/static/configuration.asciidoc
```diff
@@ -278,7 +278,7 @@ input { # comments can appear at the end of a line, too
 ----------------------------------
 
 [[event-dependent-configuration]]
-=== Event Dependent Configuration
+=== Accessing Event Data and Fields
```
suyograo: how about `Accessing Event Data and Fields In Configuration`? IMO Configuration is key here..

suyograo:
@7lima can you update this?

7lima:
@suyograo If you think the Configuration keyword from the superordinate "Configuring Logstash" chapter is necessary in the section's title then I would choose _"Accessing Event Data and Fields in the Configuration"_ like in the following "Using Environment Variables in the Configuration" section.

suyograo:
@7lima +1 on "Accessing Event Data and Fields in the Configuration"

dedemorton:
This PR was never completed, so I'm replacing it with #6767

PR #5698, opened 2016-07-27
Bump version to 5.0.0-alpha5

suyograo:


Gemfile
```diff
@@ -21,94 +21,93 @@ gem "gems", "~> 0.8.3", :group => :build
...
+gem "logstash-output-rabbitmq"
+gem "logstash-output-redis"
+gem "logstash-output-s3"
+gem "logstash-output-sns"
+gem "logstash-output-sqs"
+gem "logstash-output-statsd"
+gem "logstash-output-stdout"
+gem "logstash-output-tcp"
+gem "logstash-output-udp"
+gem "logstash-output-webhdfs"
```
ph: Are we removing all the version constrains when we release an official version, I though we were keeping them?

versions.yml
```diff
@@ -1,6 +1,6 @@
...
-logstash: 5.0.0-alpha4
-logstash-core: 5.0.0-alpha4
-logstash-core-event: 5.0.0-alpha4
-logstash-core-event-java: 5.0.0-alpha4
-logstash-core-plugin-api: 2.1.9
+logstash: 5.0.0-alpha5
+logstash-core: 5.0.0-alpha5
+logstash-core-event: 5.0.0-alpha5
+logstash-core-event-java: 5.0.0-alpha5
+logstash-core-plugin-api: 2.1.10
```
ph: <3

ph:
minor comments concerning the version constrains in the Gemfile.

Gemfile
suyograo: I think we should do that for a GA, not a pre-release.. its too early now, and plugins are getting upgraded

ph: in slack, we agreed to only do it on GA

ph:
LGTM, comment was addressed on slack.

PR #5701, opened 2016-07-27
add changes made to monitoring APIs for alpha5

dedemorton:
@andrewvc @acchen97 Please review these changes related to issue #5623 

A few outstanding questions/issues (please read and respond):
- `?human=false` should turn off human-readable input, but doesn't. Is this a bug or intentional (if intentional, the behavior is different from elasticsearch).
- Is there any plan to support `?yaml`? I need to either say that we don't support it for alpha, or I need to remove mention of it from the doc. elasticsearch does support this as a common parameter.
- Right now, I have pipeline stats API as a separate topic. However, I wanted to show it as an endpoint under the Node Stats API because users might want to do something like `/_node/stats/jvm,pipeline`. So I've added it, but I point to the "Pipeline Stats API" topic for examples. Another approach would be to remove the "Pipeline Stats API" completely and move the content to a section called "Pipeline Stats" under the Node Stats API. I think that I originally created a separate topic because pipeline stats were not returned with `/_node/stats` (not sure), but now it works. The only reason to keep this as a separate topic is if you think users are likely to be scanning the TOC to find a Pipeline Stats API.
- `/_node/stats/mem` still returns info. Is this a bug or maybe it's intentional to support earlier alpha versions? I'm assuming after reading #5623  that we want to remove mention of the endpoint from our docs because the info is now under jvm. Let me know if this is not correct.

### DIFF
--- a/docs/static/monitoring-apis.asciidoc
+++ b/docs/static/monitoring-apis.asciidoc
@@ -9,7 +9,6 @@ about Logstash:
 * <<node-info-api>>
 * <<plugins-api>>
 * <<node-stats-api>>
-* <<pipeline-stats-api>>
 * <<hot-threads-api>>
 
 
@@ -46,9 +45,7 @@ The following options can be applied to all of the Logstash monitoring APIs.
 ==== Pretty Results
 
 When appending `?pretty=true` to any request made, the JSON returned
-will be pretty formatted (use it for debugging only!). Another option is
-to set `?format=yaml` which will cause the result to be returned in the
-(sometimes) more readable yaml format.
+will be pretty formatted (use it for debugging only!).
 
 [float]
 ==== Human-Readable Output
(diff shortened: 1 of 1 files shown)


suyograo:
> Is there any plan to support ?yaml? I need to either say that we don't support it for alpha, or I need to remove mention of it from the doc. elasticsearch does support this as a common parameter.

There are no plans to support `?yaml` in any release at this point.

acchen97:
+1 to removing the `?yaml` option

> `?human=false` should turn off human-readable input, but doesn't. Is this a bug or intentional (if intentional, the behavior is different from elasticsearch).

The human option (hot threads only) default to false. `?human` and `?human=true` should return human readable. `?human=false` should return the default format. If not, we should fix for consistency. Also, I'm curious whether `?human=true` is outputting properly, its a lot more verbose than the default JSON output.

> Another approach would be to remove the "Pipeline Stats API" completely and move the content to a section called "Pipeline Stats" under the Node Stats API. 

+1 on rolling it under node stats page. It feels like a better flow and is more consistent with the API endpoints documented at the top level.

> /_node/stats/mem still returns info. Is this a bug or maybe it's intentional to support earlier alpha versions? I'm assuming after reading #5623 that we want to remove mention of the endpoint from our docs because the info is now under jvm. Let me know if this is not correct.

Looks like a bug, we should remove that mem resource if its now under `/_node/stats/jvm`. @purbon thoughts?

dedemorton:
@acchen97 @andrewvc @suyograo I've made changes based on the review comments. Let me know if this is ready to merge.

suyograo:
LGTM

dedemorton:
Resolves alpha5 monitoring API changes listed in #5418

PR #5703, opened 2016-07-28
Reorder keys for better integrations.

gen0cide-:
In the current method, it'd be helpful to have the @vars (@timestamp, @version) be ordered ahead of the other keys. This fixes that.

I know it's a slight performance hit, but it allows for tools that heavily use the logstash text formats to pickup these key vars first.


jsvd:
thank you for the PR. can you elaborate a bit more on why this is useful to you? namely how this "allows for tools that heavily use the logstash text formats to pickup these key vars first"

suyograo:
Closing based on lack of response.

PR #5706, opened 2016-07-28
Fix broken plugin tests that didn't have a defined data.home

### LINKED ISSUE #5702: Latest snapshot/master breaks plugin tests
The latest logstash snapshot/master breaks all plugin tests using Logstash devutils because https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/environment.rb#L12 depends on `bootstrap/environment.rb` which is not required during plugin tests.

The real fix is not to fix this require statement but to fix devutils to properly define the data path to `/tmp/#{something}` for test runs.

andrewvc:
This is done by providing a default data.home in tmp for environments other than a packaged logstash distro.

This is a hack and should be replaced by something better in the future.

Fixes #5702

### DIFF
--- a/logstash-core/lib/logstash/environment.rb
+++ b/logstash-core/lib/logstash/environment.rb
@@ -3,8 +3,16 @@ require "logstash/errors"
 require "logstash/config/cpu_core_strategy"
 require "logstash/settings"
 require "socket"
+require "stud/temporary"
 
 module LogStash
+  # In the event that we're requiring this file without bootstrap/environment.rb
+  if !defined?(LogStash::Environment::LOGSTASH_HOME)
+    module Environment
+      LOGSTASH_HOME = Stud::Temporary.directory("logstash-home")
+      Dir.mkdir(::File.join(LOGSTASH_HOME, "data"))
+    end
+  end
 
   [
             Setting::String.new("node.name", Socket.gethostname),


andrewvc:
I have verified this myself by pointing the ES output plugin build at this branch with my Gemfile set to:

```ruby
source 'https://rubygems.org'


gemspec

gem 'logstash-core', :path => '../logstash/logstash-core'
gem 'logstash-core-event-java', :path => '../logstash/logstash-core-event-java'
gem 'logstash-core-plugin-api', :path=> '../logstash/logstash-core-plugin-api'
```

logstash-core/lib/logstash/environment.rb
```diff
@@ -3,8 +3,16 @@
...
 require "logstash/settings"
 require "socket"
+require "stud/temporary"
 
 module LogStash
+  # In the event that we're requiring this file without bootstrap/environment.rb
+  if !defined?(LogStash::Environment::LOGSTASH_HOME)
+    module Environment
+      LOGSTASH_HOME = Stud::Temporary.directory("logstash-home")
+      Dir.mkdir(::File.join(LOGSTASH_HOME, "data"))
```
ph: Stud will not create the temporary directory, so its better to call `FileUtils.mkdir_p` so we also create the parent directory.

andrewvc: @ph really? This code runs on my system!

andrewvc: @ph it looks like it does here https://github.com/jordansissel/ruby-stud/blob/master/lib/stud/temporary.rb#L47

untergeek: +1 to using `FileUtils.mkdir_p`.  If the path is deeply nested (and `/usr/share/logstash/data` certainly fits that bill), then this is a safer approach.

ph:
@andrewvc Small comment but LGTM, tested with beats input and this PR.

logstash-core/lib/logstash/environment.rb
andrewvc: @untergeek did you click through to my link WRT stud? It looks like the Dir is correctly created.

ph: @andrewvc nevermind, I got confused by the `.pathname` method directly.

ph: +1 for using only `Dir.mkdir` in that case.

untergeek: Agreed.

untergeek:
LGTM™

PR #5713, opened 2016-07-30
improve doc on settings file by describing options

dedemorton:
This PR addresses the request in issue #5148 to improve the settings file.

I Frankensteined the descriptions of the CLI options and the comments from the logstash.yml file to come up with descriptions for the doc. I also edited the CLI doc (lightly) to make it more consistent with the descriptions I added for the settings file. I don't like (essentially) duplicating content, but reuse isn't really feasible due to the differences in how the values are specified in the YAML vs from the command line as flags.

### DIFF
--- a/docs/static/settings-file.asciidoc
+++ b/docs/static/settings-file.asciidoc
@@ -31,80 +31,112 @@ pipeline.batch.delay: 5
 
 The `logstash.yml` file includes the following settings:
 
- *`node.name`*::
-  A descriptive name for the node. If no value is specified, the default is the machine's hostname.
- 
- *`path.data`*::
-  The directory that Logstash and its plugins use for any persistent needs. The default is `LOGSTASH_HOME/data`.
- 
- *`pipeline.workers`*::
-  The number of workers that will, in parallel, execute the filter and output stages of the pipeline. 
-  This defaults to the number of the host's CPU cores. If you find that events are backing up, or that the
+[options="header"]
+|=======================================================================
+| Setting | Description | Default value
+
+| `node.name`
+| A descriptive name for the node.
+| Machine's hostname
+
+| `path.data`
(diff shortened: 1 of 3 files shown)


dedemorton:
PS I need to do more work on the CLI topic because I think the order of options is based on the old -h output (which I'm not sure makes sense, but I will investigate).

docs/static/settings-file.asciidoc
```diff
@@ -28,4 +29,82 @@ pipeline.batch.size: 125
 pipeline.batch.delay: 5
 -------------------------------------------------------------------------------------
 
-See <<command-line-flags>> for a description of the available options.
+The `logstash.yml` file includes the following settings:
+
+ *`node.name`*::
```
suyograo: Can we convert this into a table with 3 columns: `setting type`, `setting definition` and `default value`

suyograo: This will make it really easy to read

dedemorton: @suyograo I think you're saying that you would like to see these options presented in a 3 column table instead of a definition list, right? In that case, I would suggest that we have the column headings Setting, Description, and Default value. For this particular content, I OK with formatting the info as a table (though I think the descriptions are a tad long). My only caveat is that we shouldn't use a table format if we expect that we'll ever need to add multi-line code examples to a description. Multi-line code examples in tables of 3 or more columns don't display very well because of how line wrapping works. Definition lists are much more flexible, which is why we tend to use them for reference content. So...I'm open to reformatting. Let me know if you still want me to do this.

suyograo: Yeah, I think a table will be easy to read. Can we try that? +1 on Setting, Description, and Default value as column heading.s

suyograo:
LGTM

PR #5729, opened 2016-08-04
Plugin Manager: Remove validate warning because its useless

### LINKED ISSUE #5686: Plugin manager provides confusing information for upgrades
Situation: When a user tries to update a plugin, the plugin manager tells the user about the latest global version of the plugin and warns the user about possible compatibility problems. If the user says, "yes", then we install the best possible version.

Example:

```
$ ./logstash/logstash-2.2.0/bin/plugin update logstash-filter-metrics
You are updating logstash-filter-metrics to a new version 4.0.1, which may not be compatible with 3.0.0. are you sure you want to proceed (Y/N)?
Y
Updating logstash-filter-metrics
Updated logstash-filter-metrics 3.0.0 to 3.0.2
```

There are several problems here:

* The plugin manager is offering a version that may be incompatible.
* Scenario is using logstash 2.2.0, but says "may not be compatible with 3.0.0" -- it is unclear what version this is for.
* When the user agrees, we don't install the version the user agreed to.

Proposed changes:

1. Don't even offer incompatible versions on updates
2. Make it clear what project is being versioned; example, instead of "3.0.0" say "logstash-filter-metrics 3.0.0"
3. Do not offer a version to a user, who then agrees, only to install a different version. Oops ;)

suyograo:
Fixes #5686

### DIFF
--- a/lib/pluginmanager/update.rb
+++ b/lib/pluginmanager/update.rb
@@ -47,7 +47,7 @@ class LogStash::PluginManager::Update < LogStash::PluginManager::Command
     filtered_plugins = plugins.map { |plugin| gemfile.find(plugin) }
       .compact
       .reject { |plugin| REJECTED_OPTIONS.any? { |key| plugin.options.has_key?(key) } }
-      .select { |plugin| local? || (verify? ? validates_version(plugin.name) : true) }
+      .select { |plugin| local? }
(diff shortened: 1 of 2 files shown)


ph:
Changes LGTM, the QA acceptance test will run this code too.

ph:
```
```

PR #5737, opened 2016-08-08
Update after-install.sh - fix log file permissions

driegel:
When installing Logstash (after previous removal of the Logstash RPM), there is a chance that there are old log files without proper owner.
Running Logstash as service, it may want to append data to these log files. Consequently, the logstash service does not start because it doesn't have write permissions to that files. This patch sets the (new) user logstash to own these historic files.

### DIFF
--- a/pkg/centos/after-install.sh
+++ b/pkg/centos/after-install.sh
@@ -1,5 +1,5 @@
 chown -R logstash:logstash /usr/share/logstash
-chown logstash /var/log/logstash
+chown -R logstash /var/log/logstash
 chown logstash:logstash /var/lib/logstash
 chmod 0644 /etc/logrotate.d/logstash
 sed -i \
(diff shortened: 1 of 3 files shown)


untergeek:
This looks good to me.  Can you also make the same change for:

https://github.com/elastic/logstash/blob/master/pkg/debian/after-install.sh#L4

and

https://github.com/elastic/logstash/blob/master/pkg/ubuntu/after-install.sh#L4 

?

It makes sense for these changes to also exist for the other supported distributions, and to make the changes to each at the same time, rather than with separate PRs.

suyograo:
@driegel Once you make the changes suggested by @untergeek we can merge this in.

driegel:
@suyograo ready to go ...

suyograo:
LGTM

PR #5738, opened 2016-08-08
correct description of pipeline.batch.delay from seconds to ms

dedemorton:
Confirmed with @ph that the description here is incorrect.


ph:
LGTM

PR #5740, opened 2016-08-08
Fix warning when no local plugins are found

suyograo:
Will fix tests failing because of https://github.com/elastic/logstash/pull/5729#issuecomment-237852619


ph:
LGTM

Issue #5704, opened 2016-07-28
Use Jackson CBOR for event serialization


guyboertje:
See #5638

> refactor Event serialization to use JSON/CBOR

See [the conclusion from the Kryo binary experiments](https://github.com/elastic/logstash/issues/5253#issuecomment-222692691)

In all but the the most random of benchmarks Jackson JSON outperformed Kryo. 
Jackson CBOR is mostly Jackson JSON code with the output in binary that conforms to the CBOR standard. In theory the event maps serialized to binary will be deserializable by other Java CBOR implementations because the map contents are standard Java classes.

In Amsterdam Boaz suggested Jackson CBOR which aligned with my thinking in the link above.

suyograo:
Can you please put in some more description?

suyograo:
Fixed in #5746

PR #5746, opened 2016-08-09
Add Jackson based CBOR binary ser/desr

guyboertje:
Move the JSON and CBOR ObjectMapper static instances to a new class.
Use the Afterburner module with both mappers
Use a HashMap to hold the data and metadata hashmaps for serialization and deserialization.
Add a round trip JUnit test for toBinary and fromBinary.

Notes:
In the toBinary method I have not synchronised or cloned the data or the metadata Maps - I figured that there would only be one thread operating on an event at the time of serialization for the PQ.

See #5704


logstash-core-event-java/src/main/java/com/logstash/cbor/BinaryStruct.java
```diff
@@ -0,0 +1,28 @@
+package com.logstash.cbor;
```
talevy: this class does not seem to be used

talevy:
extra context around why the decision to use CBOR arose would be great. maybe we can link it with this issue: https://github.com/elastic/logstash/issues/5704

talevy:
`mapper` can be removed from the class, no? https://github.com/elastic/logstash/pull/5746/files#diff-69f9c9b29223bfb968ccf0525f9fecd7L39

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -167,10 +170,35 @@ public boolean includes(String reference) {
...
+        if (o == null) {
+            throw new IOException("The binary source deserialized to null");
+        }
+        try {
+            HashMap<String, Map<String, Object>> hashMap = (HashMap<String, Map<String, Object>>) o;
+            HashMap<String, Object> dataMap = (HashMap<String, Object>) hashMap.get("DATA");
+            dataMap.put(METADATA, hashMap.get("META"));
+            return new Event(dataMap);
+        } catch (ClassCastException e) {
+            throw new IOException("incompatible from binary object type=" + o.getClass().getName() + " , only HashMap is supported");
```
talevy: seems to be more than just that `HashMap` is the only acceptable object. It must be a very specific HashMap with `DATA` and `META`, maybe catching all these exceptions together makes sense?

logstash-core-event-java/src/main/java/com/logstash/Event.java
```diff
@@ -167,10 +170,35 @@ public boolean includes(String reference) {
...
+    public static Event fromBinary(byte[] source) throws IOException {
+        if (source == null || source.length == 0) {
+            return new Event();
+        }
+        Object o = CBOR_MAPPER.readValue(source, HashMap.class);
+        if (o == null) {
+            throw new IOException("The binary source deserialized to null");
+        }
+        try {
+            HashMap<String, Map<String, Object>> hashMap = (HashMap<String, Map<String, Object>>) o;
```
talevy: do we have to be implementation specific here? can we just cast it to any `Map`?

suyograo:
Why is this getting into master branch? Master is still tracking 5.0. Isn't it better to merge this into the persistence feature branch?

talevy:
how do we intend to support ruby-land types? I notice that this is only java-specific serialization.

Here is an example of what I am referring to [1]:

```java
    @Test
    public void toBinaryRoundtripWithRubySymbols() throws Exception {

        JrubyEventExtLibrary.RubyEvent originalEvent = (JrubyEventExtLibrary.RubyEvent)
                ruby.evalScriptlet("LogStash::Event.new({\"message\" => :symbol})");

        byte[] binary = originalEvent.getEvent().toBinary();

        JrubyEventExtLibrary.RubyEvent convertedEvent = JrubyEventExtLibrary.RubyEvent.newRubyEvent(ruby, Event.fromBinary(binary));


        ThreadContext context = ruby.getCurrentContext();
        RubyString messageString = RubyString.newString(ruby, "message");

        assertEquals(convertedEvent.ruby_get_field(context, messageString), originalEvent.ruby_get_field(context, messageString));
    }
```

The above test fails due to the javafication of the ruby symbol into a Java String, and this type information is lost during serialization. If I am not mistaken, users expecting a specific value to be a symbol from an input -> filter (or between filters) will run into this scenario I believe.

let me know if I understand things correctly


[1] https://github.com/talevy/logstash/commit/ee7d2b18884754d785a4af170906b8361a41ca3e

logstash-core-event-java/src/main/java/com/logstash/cbor/BinaryStruct.java
guyboertje: Sorry, my bad. I thought I had deleted that class. I was experimenting with a struct like object but then opted for a HashMap to carry the data and metadata.

Will remove.

guyboertje:
@suyog - AFAIK, master, because the PQ references the Java Event as a provided dependency.

This is, in part, why we need to discuss the need for release cycle independency of java event, PQ and LS core.

guyboertje:
@talevy - we discussed the handling of ruby types via zoom. Would you mind summarising your understanding of what was concluded in a comment here, for the sake of future readers.

logstash-core-event-java/src/main/java/com/logstash/Event.java
guyboertje: Sure. This ensures that it is a HashMap anyway `Object o = CBOR_MAPPER.readValue(source, HashMap.class);`.

guyboertje:
I will be closing this PR because the changes have been put directly in `feature/java_persistence` branch

PR #5750, opened 2016-08-10
Fix the require_jar warning about jackson jars

### LINKED ISSUE #5159: sync jrjackson & logstash-core-event-java Jackson dependencies versions

suyograo:
jrjackson was using 2.7.3 whereas core was using 2.7.0. Upgraded jrjackson
to 0.4.0 and core libs to match jars in jrjackson

Fixes #5159

```
suyog@machine:~/ws/elastic/logstash (fix/5159)$ bin/logstash -e 'input { generator { "message" => "this is a bats test for -e" count => 1 }} output { stdout {} }'
Pipeline main started
2016-08-10T00:48:46.085Z Suyogs-MBP-2 this is a bats test for -e
Pipeline main has been shutdown
```

### DIFF
--- a/logstash-core-event-java/build.gradle
+++ b/logstash-core-event-java/build.gradle
@@ -92,8 +92,8 @@ idea {
 }
 
 dependencies {
-    compile 'com.fasterxml.jackson.core:jackson-core:2.7.1'
-    compile 'com.fasterxml.jackson.core:jackson-databind:2.7.1-1'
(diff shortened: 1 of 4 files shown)


talevy:
LGTM

Issue #5662, opened 2016-07-20
[Meta] Tighten-up Output Workers for 5.0 Release


andrewvc:
A lot has changed in preparation for 5.0 we need to tighten up a lot of our output worker story before 5.0 . This meta issue will track those changes. Ultimately, we'd love to get rid of output workers(https://github.com/elastic/logstash/issues/5653) but that will probably have to wait till 6.0.

- [x] Educate: Performance Tuning Guide
- [x] Educate: Blog detailing changes and internals of pipeline architecture.
- [x] [Improve nomenclature and methods for 'threadsafe' outputs](https://github.com/elastic/logstash/issues/5661)
- [x] [Remove instance method version of `workers_not_supported`](https://github.com/elastic/logstash/issues/5658)
- [x] [Raise Error when output_workers set for threadsafe plugin](https://github.com/elastic/logstash/issues/5659)
- [x] [Enforce unique ID for plugins](https://github.com/elastic/logstash/issues/5754)
- [x] [Modify docs generator to correctly tag threadsafe outputs](https://github.com/elastic/logstash/issues/5660)
- [x] [Implement `Codec#multi_encode` / `Output#multi_receive_encoded`](https://github.com/elastic/logstash/issues/5768)
- [x]  [Mass update plugins to conform to new output delegator](https://github.com/elastic/logstash/issues/5663)
- [x] [Output plugins not receiving metrics](https://github.com/elastic/logstash/issues/5810)
- [x] [Add workers_not_supported obsolescence message](https://github.com/elastic/logstash/issues/5823)
- [x] [Improve OD public API](https://github.com/elastic/logstash/issues/5813)

suyograo:
Added blog and docs item as well.

Issue #5659, opened 2016-07-20
Raise error when `output_workers` set for threadsafe plugin


andrewvc:
Currently users can still manually set output workers for outputs marked threadsafe. This currently does nothing but should raise a `LogStash::ConfigurationError`

andrewvc:
Fixed in #5752

Issue #5658, opened 2016-07-20
Remove instance method version of `workers_not_supported`


andrewvc:
`Output#workers_not_supported` (see https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/outputs/base.rb#L56)  has been replaced by its class level variant.

We should that method and only allow the class method. We will need to mass update any plugins that use this to ONLY use the class method. This should be backward compatible with 2.4x keeping our plugin API contract sane.

andrewvc:
Fixed in #5752

PR #5752, opened 2016-08-10
Delegator refactor

### LINKED ISSUE #5661: Improve output nomenclature for 'threadsafe' outputs and other types
We now have 'threadsafe' outputs, but this nomenclature is confusing to some. The other types can have threads and work correctly! Perhaps we should change and improve this on the public facing side.

Additionally, we should move to deprecate the worker model (esp. as we move toward https://github.com/elastic/logstash/issues/5653). To that end we should by default mark plugins as only allowing a single worker unless they opt into another strategy.

I propose the following changes.

Replace the class method `Outputs::Base.declare_threadsafe!` with `Outputs::Base.worker_strategy=(strategy)`.

The strategies would be renamed `from->to`:

* `threadsafe->shared`. This indicates that the output can be shared amongst any number of pipeline workers.
* `(default)->single`. Only a single instance of this output can exist. Note that this would mean that most plugins today would no longer support multiple workers. You would need to opt-in rather than out. The effects here need to be discussed.
* `legacy_workers`. Multiple workers are supported, but its clear that this is deprecated. 

So, a plugin might look like the following:

```ruby
class MyOutput < LogStash::Outputs::Base
  worker_strategy :shared #default is :single
end
```

If no strategy is declared then `single` would be assumed as the default value.

andrewvc:
This WIP cleans up everything related to output delegators and specifically cleans up the nomenclature WRT `concurrency`.

This is ready for review, though tests around `Pipeline#plugin` need some work. I need to write some spec helpers to let us properly test alternate registries to satisfy this. Additionally we need to make the plugin registry injectable for proper testability. I'm deliberating doing this in a separate PR.

Fixes #5661  #5658 #5659 #5754 all related to meta issue https://github.com/elastic/logstash/issues/5662

### DIFF
--- a/docs/static/include/pluginbody.asciidoc
+++ b/docs/static/include/pluginbody.asciidoc
@@ -275,40 +275,31 @@ require "logstash/namespace"
 class LogStash::{pluginclass}::{pluginnamecap} < LogStash::{pluginclass}::Base
   config_name "example"
 
-  # If declared logstash will only allow a single instance of this plugin
-  # to exist, regardless of how many CPU cores logstash detects. This is best
-  # used in cases like the File output, where separate threads writing to a single
-  # File would only cause problems.
+  # This sets the concurrency behavior of this plugin. By default it is :legacy, which was the standard
+  # way concurrency worked before Logstash 2.4
+  # 
+  # You should explicitly set it to either :single or :shared as :legacy will be removed in Logstash 6.0
+  # 
+  # When configured as :single a single instance of the Output will be shared among the
+  # pipeline worker threads. Access to the `#multi_receive/#multi_receive_encoded/#receive` method will be synchronized
+  # i.e. only one thread will be active at a time making threadsafety much simpler.
+  # 
+  # You can set this to :shared if your output is threadsafe. This will maximize
+  # concurrency but you will need to make appropriate uses of mutexes in `#multi_receive/#receive`.
   #
-  # respond_to? check needed for backwards compatibility with < 2.2 Logstashes
-  declare_workers_not_supported! if self.respond_to?(:declare_workers_not_supported!)
-
-  # If declared threadsafe logstash will only ever create one
-  # instance of this plugin per pipeline.
-  # That instance will be shared across all workers
-  # It is up to the plugin author to correctly write concurrent code!
-  #
-  # respond_to? check needed for backwards compatibility with < 2.2 Logstashes
-  declare_threadsafe! if self.respond_to?(:declare_threadsafe!)
-
+  # Only the `#multi_receive/#multi_receive_encoded` methods need to actually be threadsafe, the other methods
+  # will only be executed in a single thread
+  concurrency :single
+  
   public
-  def register
-    # Does the same thing as declare_workers_not_supported!
-    # But works in < 2.2 logstashes
-    # workers_not_supported
+  def register    
   end # def register
 
   public
   # Takes an array of events
+  # Must be threadsafe if `concurrency :shared` is set
   def multi_receive(events)
   end # def multi_receive
-
-  public
-  # Needed for logstash < 2.2 compatibility
-  # Takes events one at a time
-  def receive(event)
-  end # def receive
-
 end # class LogStash::{pluginclass}::{pluginnamecap}
 ----------------------------------
 endif::multi_receive_method[]
--- a/logstash-core/lib/logstash/api/commands/stats.rb
+++ b/logstash-core/lib/logstash/api/commands/stats.rb
@@ -70,7 +70,7 @@ module LogStash
             # Turn the `plugins` stats hash into an array of [ {}, {}, ... ]
             # This is to produce an array of data points, one point for each
             # plugin instance.
-            return [] unless stats[:plugins].include?(plugin_type)
+            return [] unless stats[:plugins] && stats[:plugins].include?(plugin_type)
             stats[:plugins][plugin_type].collect do |id, data|
               { :id => id }.merge(data)
             end
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -13,15 +13,14 @@ module LogStash
     ]
     def_delegators :@filter, *DELEGATED_METHODS
 
-    def initialize(logger, klass, metric, *args)
-      options = args.reduce({}, :merge)
-
+    def initialize(logger, klass, metric, plugin_args)
       @logger = logger
       @klass = klass
-      @filter = klass.new(options)
+      @id = plugin_args["id"]
+      @filter = klass.new(plugin_args)
 
       # Scope the metrics to the plugin
-      namespaced_metric = metric.namespace(@filter.plugin_unique_name.to_sym)
+      namespaced_metric = metric.namespace("#{@klass.config_name}_#{@id}".to_sym)
       @filter.metric = namespaced_metric
 
       @metric_events = namespaced_metric.namespace(:events)
--- a/logstash-core/lib/logstash/output_delegator.rb
+++ b/logstash-core/lib/logstash/output_delegator.rb
@@ -1,192 +1,56 @@
-# encoding: utf-8
-require "concurrent/atomic/atomic_fixnum"
-java_import "java.util.concurrent.CopyOnWriteArrayList"
+require "logstash/output_delegator_strategy_registry"
 
-# This class goes hand in hand with the pipeline to provide a pool of
-# free workers to be used by pipeline worker threads. The pool is
-# internally represented with a SizedQueue set the the size of the number
-# of 'workers' the output plugin is configured with.
-#
-# This plugin also records some basic statistics
+require "logstash/output_delegator_strategies/shared"
+require "logstash/output_delegator_strategies/single"
+require "logstash/output_delegator_strategies/legacy"
+
 module LogStash class OutputDelegator
-  attr_reader :workers, :config, :threadsafe
+  attr_reader :metric, :metric_events, :strategy, :namespaced_metric, :metric_events , :plugin_args, :strategy_registry
 
-  # The *args this takes are the same format that a Outputs::Base takes. A list of hashes with parameters in them
-  # Internally these just get merged together into a single hash
-  def initialize(logger, klass, default_worker_count, metric, *plugin_args)
+  def initialize(logger, output_class, metric, strategy_registry, plugin_args)
     @logger = logger
-    @threadsafe = klass.threadsafe?
-    @config = plugin_args.reduce({}, :merge)
-    @klass = klass
-    @workers = java.util.concurrent.CopyOnWriteArrayList.new
-    @default_worker_count = default_worker_count
-    @registered = false
+    @output_class = output_class
+    @metric = metric
+    @plugin_args = plugin_args
+    @strategy_registry ||= ::LogStash::OutputDelegatorStrategyRegistry.instance
+    raise ArgumentError, "No ID specified! Got args #{plugin_args}" unless id
+    
+    build_strategy!
 
-    # Create an instance of the input so we can fetch the identifier
-    output = @klass.new(@config)
-
-    # Scope the metrics to the plugin
-    namespaced_metric = metric.namespace(output.plugin_unique_name.to_sym)
-    output.metric = namespaced_metric
-
-    @metric_events = namespaced_metric.namespace(:events)
-    namespaced_metric.gauge(:name, config_name)
-
-    @events_received = Concurrent::AtomicFixnum.new(0)
-  end
-
-  def threadsafe?
-    !!@threadsafe
-  end
-
-  def warn_on_worker_override!
-    # The user has configured extra workers, but this plugin doesn't support it :(
-    if worker_limits_overriden?
-      message = @klass.workers_not_supported_message
-      warning_meta = {:plugin => @klass.config_name, :worker_count => @config["workers"]}
-      if message
-        warning_meta[:message] = message
-        @logger.warn(I18n.t("logstash.pipeline.output-worker-unsupported-with-message", warning_meta))
-      else
-        @logger.warn(I18n.t("logstash.pipeline.output-worker-unsupported", warning_meta))
-      end
-    end
-  end
-
-  def worker_limits_overriden?
-    @config["workers"] && @config["workers"] > 1 && @klass.workers_not_supported?
-  end
-
-  def target_worker_count
-    # Remove in 5.0 after all plugins upgraded to use class level declarations
-    raise ArgumentError, "Attempted to detect target worker count before instantiating a worker to test for legacy workers_not_supported!" if @workers.size == 0
-
-    if @threadsafe || @klass.workers_not_supported?
-      1
-    else
-      @config["workers"] || @default_worker_count
-    end
+    @namespaced_metric = metric.namespace(id.to_sym)
+    @metric_events = @namespaced_metric.namespace(:events)
+    @namespaced_metric.gauge(:name, id)
   end
 
   def config_name
-    @klass.config_name
+    @output_class.config_name
+  end
+
+  def concurrency
+    @output_class.concurrency
+  end
+
+  def build_strategy!
+    @strategy = strategy_registry.
+                  class_for(self.concurrency).
+                  new(@logger, @output_class, @metric, @plugin_args)
+  end
+
+  def id
+    @plugin_args["id"]
   end
 
   def register
-    raise ArgumentError, "Attempted to register #{self} twice!" if @registered
-    @registered = true
-    # We define this as an array regardless of threadsafety
-    # to make reporting simpler, even though a threadsafe plugin will just have
-    # a single instance
-    #
-    # Older plugins invoke the instance method Outputs::Base#workers_not_supported
-    # To detect these we need an instance to be created first :()
-    # TODO: In the next major version after 2.x remove support for this
-    @workers << @klass.new(@config)
-    @workers.first.register # Needed in case register calls `workers_not_supported`
-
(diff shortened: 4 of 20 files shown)


logstash-core/lib/logstash/api/commands/stats.rb
```diff
@@ -70,7 +70,7 @@ def plugin_stats(stats, plugin_type)
             # Turn the `plugins` stats hash into an array of [ {}, {}, ... ]
             # This is to produce an array of data points, one point for each
             # plugin instance.
-            return [] unless stats[:plugins].include?(plugin_type)
+            return [] unless stats[:plugins] && stats[:plugins].include?(plugin_type)
```
andrewvc: Fixes an issue where this might run before `:plugins` is defined.

logstash-core/lib/logstash/filter_delegator.rb
```diff
@@ -13,15 +13,14 @@ class FilterDelegator
     ]
     def_delegators :@filter, *DELEGATED_METHODS
 
-    def initialize(logger, klass, metric, *args)
-      options = args.reduce({}, :merge)
-
+    def initialize(logger, klass, metric, plugin_args)
```
andrewvc: Renamed options to `plugin_args` to be more clear and consistent across various classes.

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -1,192 +1,60 @@
...
-
-  def worker_count
-    @workers.size
-  end
-
-  private
-  # Needed for testing, so private
-  attr_reader :threadsafe_worker, :worker_queue
-end end
+end; end
```
guyboertje: Looking at this class holistically...
- Method `id` and ivar `@id`. Consider only using the `id` method and provide nil fallback `@plugin_args.fetch('id', SecureRandom.hex(2))

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -1,192 +1,60 @@
...
+    @xopts = xopts # Options specific to the given class
+    @plugin_args = plugin_args
+    build_strategy!
 
-    @metric_events = namespaced_metric.namespace(:events)
-    namespaced_metric.gauge(:name, config_name)
-
-    @events_received = Concurrent::AtomicFixnum.new(0)
+    @id = plugin_args["id"]
+    @namespaced_metric = metric.namespace("#{@output_class.config_name}_#{@id}".to_sym)
```
guyboertje: Consider using the method `config_name` for the metric namespace, further more that would remove the need to create an ivar on line 17.

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -1,192 +1,60 @@
...
-        @logger.warn(I18n.t("logstash.pipeline.output-worker-unsupported-with-message", warning_meta))
-      else
-        @logger.warn(I18n.t("logstash.pipeline.output-worker-unsupported", warning_meta))
-      end
-    end
-  end
+  def build_strategy!
+    # TODO: Make this injected
+    @strategy_registry = LogStash::OutputDelegatorStrategyRegistry.instance
+    @strategy_class = @strategy_registry.get(self.concurrency)
```
guyboertje: Consider raising the "Could not find output delegator strategy of type..." error in the `instance.get()` method (it is probably where it belongs, as any other method using it will have to also test and raise).
However, maybe `class_for` makes more sense .e.g. `.instance.class_for(self.concurrency).new(...)`
Gives...
```ruby
  def build_strategy!
    @output_delegate = LogStash::OutputDelegatorStrategyRegistry.instance
        .class_for(self.concurrency)
        .new(@logger, @output_class, @metric, @xopts, @plugin_args)
  end
```

logstash-core/lib/logstash/output_delegator_strategies/legacy_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+    @workers.each {|w| @worker_queue << w}
+  end
+  
+  def register
+    @workers.each(&:register)
+  end
+  
+  def multi_receive(events)
+    worker = @worker_queue.pop
+  ensure
```
guyboertje: Did you mean to not do anything with events? To be consistent with A) th previous code and B) the other strategies, consider this:
```ruby
  def multi_receive(events)
    output = @worker_queue.pop
    begin
      output.multi_receive(events)
    ensure
      @worker_queue.push(output)
    end
  end
```

logstash-core/lib/logstash/output_delegator_strategy_registry.rb
```diff
@@ -0,0 +1,28 @@
...
+  def classes
+    @map.values
+  end
+
+  def types
+    @map.keys
+  end
+  
+  def get(type)
+    @map[type]
```
guyboertje: As mentioned above, consider changing this name to `class_for(type)` (its not immediately obvious from the class name `OutputDelegatorStrategyRegistry` that classes are stored within) and raising an error when not found here.

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -1,192 +1,60 @@
...
-    else
-      @config["workers"] || @default_worker_count
-    end
+  def id
+    @plugin_args["id"]
   end
 
   def config_name
-    @klass.config_name
+    "#{@output_class.config_name}_#{id}"
```
jordansissel: Apologies if this is inarticulate, but this feels wrong. The `config_name` is like the name of the plugin, right? For this delegator, it becomes `name-id` where `id` is a computed value.

Can we use a better name than `config_name` for this method -- something that better fits the purpose? Maybe `plugin_id` or something?

logstash-core/lib/logstash/output_delegator_strategies/legacy_worker.rb
jordansissel: Agreed. Is legacy missing a complete implementation?

logstash-core/lib/logstash/output_delegator_strategy_registry.rb
```diff
@@ -0,0 +1,28 @@
+module LogStash; class OutputDelegatorStrategyRegistry
+  # This is generally used as a singleton
+  # Except perhaps during testing
+  def self.instance
+    @instance ||= self.new
```
jordansissel: No change required, but ruby has a `Singleton` library in stdlib that does this.

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -20,42 +20,40 @@ class LogStash::Outputs::Base < LogStash::Plugin
...
+  
+  # Set or return concurrency type
+  def self.concurrency(type=nil)
+    if type
+      @concurrency = type
+
+      # TODO remove this in Logstash 6.0
+      # when we no longer support the :legacy_worker type
+      # This is hacky, but it can only be here
+      if type == :legacy_worker
```
jordansissel: Feels inconsistent to have them called `shared`, `single`, and then a two-word name `legacy worker`. Can we call it `legacy` instead of `legacy_worker`? What's the story on the `_worker` suffix?

docs/static/include/pluginbody.asciidoc
```diff
@@ -275,40 +275,29 @@ require "logstash/namespace"
...
   config_name "example"
 
-  # If declared logstash will only allow a single instance of this plugin
-  # to exist, regardless of how many CPU cores logstash detects. This is best
-  # used in cases like the File output, where separate threads writing to a single
-  # File would only cause problems.
+  # This is the default concurrency level of this output plugin.
+  # You do not need to explicitly set it.
+  # When configured as :single a single instance of the Output will be shared among the
+  # pipeline threads. Access to the `#multi_receive/#receive` method will be synchronized
```
jordansissel: should we call this `pipeline workers` instead of `pipeline threads` ?

docs/static/include/pluginbody.asciidoc
```diff
@@ -275,40 +275,29 @@ require "logstash/namespace"
...
+  # i.e. only one thread will be active at a time making threadsafety much simpler.
```
jordansissel: Similar here, thread vs worker terms.

jordansissel:
Overall, I am big +1 on the idea here. My comments are highly localized and mostly around "is this the best term for this thing?" or similar descriptive concerns.

I haven't reviewed the test changes, yet, but I do see they are passing in travis.

logstash-core/lib/logstash/output_delegator.rb
guyboertje: I feel the same way, however I did not comment on this because it is part of the public API. But, if id is now an intrinsic part of the config_name then it is correct.

logstash-core/lib/logstash/output_delegator_strategies/legacy_worker.rb
```diff
@@ -0,0 +1,27 @@
+# Remove this in Logstash 6.0
+module LogStash module OutputDelegatorStrategies class LegacyWorker
+  ::LogStash::OutputDelegatorStrategyRegistry.instance.register(:legacy_worker, self)
```
guyboertje: I am concerned about this register statement.

In practice this is not a problem because we are in full control of how and when these classes are retrieved. That said, the current code would give the casual reader the idea that this is OK but there is a bug lurking in it. The class is registered without methods - they are added shortly after registration. However, if the class is retrieved and instantiated before the methods are added the instance will have no public API.

Consider moving the registration to after the methods are added. Alternatives:
```ruby
module LogStash
  module OutputDelegatorStrategies class LegacyWorker
    # rest of class definition omitted for illustration
  end; end

  OutputDelegatorStrategyRegistry.instance.register(
    :legacy_worker, OutputDelegatorStrategies::LegacyWorker)
end
```
or
```ruby
module LogStash module OutputDelegatorStrategies class LegacyWorker
  # rest of class definition omitted for illustration
  
  ::LogStash::OutputDelegatorStrategyRegistry.instance.register(:legacy_worker, self)
end; end; end
```

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -45,11 +46,21 @@ module LogStash; class Pipeline
...
+    # a unique id when auto-generating plugin ids
+    @plugin_counter ||= 0 
+    
     @logger = Cabin::Channel.get(LogStash)
     @settings = settings
     @pipeline_id = @settings.get_value("pipeline.id") || self.object_id
     @reporter = LogStash::PipelineReporter.new(@logger, self)
 
+    # Maybe make this configurable via an option at some point
+    @output_delegator_strategy_registry = ::LogStash::OutputDelegatorStrategyRegistry.instance
```
guyboertje: This reference does not seem to be used anywhere.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -387,23 +398,38 @@ def shutdown_workers
...
+    plugin = if plugin_type == "output"
+               OutputDelegator.new(@logger, klass, type_scoped_metric, {}, args)
+             #LogStash::OutputDelegator.new(@logger, klass, @settings.get("pipeline.output.workers"), , *args)
+             elsif plugin_type == "filter"
+               LogStash::FilterDelegator.new(@logger, klass, type_scoped_metric, args)
+             else # input
+               new_plugin = klass.new(args)
+               namespaced_metric = type_scoped_metric.namespace(id)
+               new_plugin.metric = namespaced_metric
+               new_plugin
```
guyboertje: Consider renaming the `new_plugin` variable to the self documenting`input`.

logstash-core/lib/logstash/plugin.rb
```diff
@@ -45,8 +45,11 @@ def eql?(other)
     self.class.name == other.class.name && @params == other.params
   end
 
-  def initialize(params=nil)
+  def initialize(params=nil)    
     @params = LogStash::Util.deep_clone(params)
+    # The id should always be defined normally, but in tests that might not be the case
+    # In the future we may make this more strict in the Plugin API
+    @params["id"] ||= params["id"] || "#{self.class.config_name}_#{SecureRandom.uuid}"
```
guyboertje: Do you need the `params["id"] || `?

guyboertje:
This is looking good. Just a few considerations to address and its not WIP anymore.

logstash-core/lib/logstash/output_delegator_strategies/legacy_worker.rb
andrewvc: That is totally a bug, and I'm clearly missing a test to boot. Good catch :)

logstash-core/lib/logstash/output_delegator_strategy_registry.rb
andrewvc: I'm not using singleton here because during testing I do want the ability to create a new instance without the default plugin list. I think we need a better name for this pattern (do you know of one that is already in use?) as `instance` too strongly implies a singleton. 

This is more like 'it works as a singleton by default but it is totally legit to make another instance, which is very useful for testing.'

logstash-core/lib/logstash/outputs/base.rb
andrewvc: Fair point, I'm cool calling it `legacy`. The `worker` suffix was meant to inform the user that this is the one that goes with the `workers` setting. The other ones no longer support that setting.

docs/static/include/pluginbody.asciidoc
andrewvc: How bout I rename it pipeline worker threads. Since we're talking about synchronization I like that we're informing the reader that these are os threads.

logstash-core/lib/logstash/output_delegator.rb
andrewvc: Thanks for the good feedback. Looking at this closer I suspect we no longer need this method, I'm going to try to remove it and see if there are any issues there.

logstash-core/lib/logstash/output_delegator_strategy_registry.rb
andrewvc: OK, so it looks like this should just proxy through to `@output_class.config_name`, and in other places where we use uniqueness `#id` should be invoked directly. Will push this up in my next commit.

logstash-core/lib/logstash/output_delegator.rb
andrewvc: Good point on only using the ID method. In my next commit I'll actually require the ID be set on plugin create. It's a sane requirement and reduces mystery.

Hmmm, @jordansissel and I arrived at the term `@strategy` after a discussion. The thing is that at some point you an `OutputDelegator` and an `OutputProxy/OutputDelegate` AND an `Output` and that all gets very confusing in terms of hierarchy.

As is you say "We have an output delegator, it has a strategy, the strategy controls the output itself", which, to my ears is a bit easier to understand.

logstash-core/lib/logstash/output_delegator.rb
andrewvc: I've actually refactored this to use `#id` since it should be unique :)

logstash-core/lib/logstash/output_delegator.rb
andrewvc: OK, so it looks like this should just proxy through to @output_class.config_name, and in other places where we use uniqueness #id should be invoked directly. Will push this up in my next commit.

logstash-core/lib/logstash/pipeline.rb
andrewvc: Good point. My original intent was to pass this as an argument to the output delegator. I think I'm going to go ahead and make that refactor now as its worth it. 

Ultimately this should simply be the default value, with the calling code able to override it so we can do richer testing.

logstash-core/lib/logstash/plugin.rb
andrewvc: I don't believe so! Good catch

logstash-core/lib/logstash/output_delegator_strategy_registry.rb
```diff
@@ -17,8 +19,14 @@ def types
     @map.keys
   end
   
-  def get(type)
-    @map[type]
+  def class_for(type)
+    klass = @map[type]
+
+    if !klass
+      raise InvalidStrategyError, "Could not find output delegator strategy of type '#{type}'. Valid strategies: #{@strategy_registry.types}"
```
guyboertje: Nice!

andrewvc:
@guyboertje @jordansissel I believe I've addressed all your comments. I've tried to separate things into separate commits to make viewing the changes I made today more visible. I'll squash them all before final merge.

guyboertje:
LGTM

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -32,24 +32,34 @@ module LogStash; class Pipeline
     :started_at,
     :thread,
     :config_str,
+    :config_hash,
     :settings,
     :metric,
     :filter_queue_client,
-    :input_queue_client
+    :input_queue_client,
+    :xopts
```
jordansissel: the `xopts` name is confusing to me. What does it mean? Can we use a word or two to describe `word1_word2` or something instead of this `xopts` name?

jordansissel:
General LGTM. I'd like to see `xopts` renamed to something more descriptive for me, but we can change that later if desired.

logstash-core/lib/logstash/pipeline.rb
andrewvc: The basic idea was to provide extended options. Do you prefer `extended_options`? I dislike `options` because we already have settings, but `settings` unfortunately is for user configurable things and not rich objects ATM.

andrewvc: How about `xoptions`?

PR #5753, opened 2016-08-10
Fix typo in help text

devkmsg:


suyograo:
@devkmsg The text seems to correctly point to `bin/logstash-plugin pack` there is no `bin/logstash-plugin package`

Issue #5744, opened 2016-08-09
`rake test:integration` is failing


talevy:
Looks like the tests under `spec/integration` were moved. this move is causing `rake test:integration` to fail.

error:

```
21) bin/logstash-plugin uninstall when the plugin isn't installed fails to uninstall it
     Failure/Error: result = command("bin/logstash-plugin uninstall logstash-filter-cidr")
     IOError:
       Cannot run program "bin/logstash-plugin" (in directory "/Users/tal/Code/es/talevy-logstash/spec"): error=2, No such fi
le or directory
     # ./spec/integration/support/integration_test_helpers.rb:40:in `command'
     # ./spec/integration/support/integration_test_helpers.rb:39:in `command'
     # ./vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler.rb:245:in `with_clean_env'
     # ./vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler.rb:232:in `with_original_env'
     # ./vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler.rb:238:in `with_clean_env'
     # ./spec/integration/support/integration_test_helpers.rb:38:in `command'
     # ./spec/integration/plugin_manager/plugin_uninstall_spec.rb:7:in `(root)'
     # ./rakelib/test.rake:115:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/task.rb:248:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/task.rb:243:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/task.rb:180:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/task.rb:173:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/application.rb:152:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/application.rb:108:in `top_level'
[...]

PR #5760, opened 2016-08-12
Remove spec/integration

ph:
With the addition of the qa acceptance test suite we have decide that we
don't need to maintain this suite also since the acceptance test are
running really similar test.

Fixes: #5744


suyograo:
LGTM

Issue #5764, opened 2016-08-16
Flag --debug-config doesn't work with -t in Logstash 2.x


jsvd:
Expected behaviour with `--debug --debug-config -t` should be that the debug statements print out the configuration:

```
% bin/logstash -f 'configs/*' --debug --debug-config -t
Reading config file {:config_file=>"/private/tmp/logstash-2.3.4/configs/a", :level=>:debug, :file=>"logstash/config/loader.rb", :line=>"69", :method=>"local_config"}

The following is the content of a file {:config_file=>"/private/tmp/logstash-2.3.4/configs/a", :level=>:debug, :file=>"logstash/config/loader.rb", :line=>"76", :method=>"local_config"}

input { stdin { } }


 {:level=>:debug, :file=>"logstash/config/loader.rb", :line=>"77", :method=>"local_config"}
Reading config file {:config_file=>"/private/tmp/logstash-2.3.4/configs/b", :level=>:debug, :file=>"logstash/config/loader.rb", :line=>"69", :method=>"local_config"}

The following is the content of a file {:config_file=>"/private/tmp/logstash-2.3.4/configs/b", :level=>:debug, :file=>"logstash/config/loader.rb", :line=>"76", :method=>"local_config"}

output { stdout { } }


 {:level=>:debug, :file=>"logstash/config/loader.rb", :line=>"77", :method=>"local_config"}

The following is the merged configuration {:level=>:debug, :file=>"logstash/config/loader.rb", :line=>"84", :method=>"local_config"}

input { stdin { } }

output { stdout { } }



 {:level=>:debug, :file=>"logstash/config/loader.rb", :line=>"85", :method=>"local_config"}
Plugin not defined in namespace, checking for plugin file {:type=>"input", :name=>"stdin", :path=>"logstash/inputs/stdin", :level=>:debug, :file=>"logstash/plugin.rb", :line=>"76", :method=>"lookup"}
Plugin not defined in namespace, checking for plugin file {:type=>"codec", :name=>"line", :path=>"logstash/codecs/line", :level=>:debug, :file=>"logstash/plugin.rb", :line=>"76", :method=>"lookup"}
config LogStash::Codecs::Line/@charset = "UTF-8" {:level=>:debug, :file=>"logstash/config/mixin.rb", :line=>"153", :method=>"config_init"}
[...]

PR #5765, opened 2016-08-16
remove duplicate config_loader creation when -t is enabled

jsvd:
fixes https://github.com/elastic/logstash/issues/5764

### DIFF
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -299,9 +299,9 @@ describe LogStash::Agent do
 
   describe "--config-test" do
     let(:cli_args) { ["-t", "-e", pipeline_string] }
+    let(:pipeline_string) { "input { } filter { } output { }" }
 
     context "with a good configuration" do
-      let(:pipeline_string) { "input { } filter { } output { }" }
       it "should exit successfuly" do
         expect(subject.run(cli_args)).to eq(0)
       end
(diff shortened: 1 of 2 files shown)


logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -313,6 +313,11 @@
         expect(subject.run(cli_args)).to eq(1)
       end
     end
+
+    it "requests the config loader to format_config" do
+      expect(subject.config_loader).to receive(:format_config)
+      subject.run(cli_args)
+    end
```
ph: Should we actually test that the config was correctly sent to the logger?

ph:
I've tested this PR and I got what I believe a weird result.

if you do `bin/logstash --debug --debug-config -t  -e "input { generator {}}"` you wont get any output of the configuration.

If you do `bin/logstash --debug --debug-config -t -f hello.conf -e "input { tcp { port => 9000}}"` You will see only the content of the *hello.conf*, but the pipeline will also setup the TCP input..

jsvd:
Well spotted. Interestingly, the first scenario has always been true since --debug-config was introduced, the piece of config coming from -e isn't included in debug. I'll check what's needed for it, but IMO can be done in another PR.

As for the second thing..it doesn't actually open the tcp port, just does the plugin init (not register). This has happened in older logstash versions.

ph:
> Well spotted. Interestingly, the first scenario has always been true since --debug-config was introduced, the piece of config coming from -e isn't included in debug. I'll check what's needed for it, but IMO can be done in another PR.

Agree for another PR.

> As for the second thing..it doesn't actually open the tcp port, just does the plugin init (not register). This has happened in older logstash versions.

oh that is effectively confusing then.

ph:
LGTM, this PR fix the mentioned issue.

Issue #5732, opened 2016-08-04
Monitoring API enhancements/cleanups for 5.0


acchen97:
This issue will track the final iteration of monitoring API fixes/additions in preparation for beta1.

- **Node Stats**
 - [x] Add `duration_in_millis` metric under `/_node/stats/pipeline/events` https://github.com/elastic/logstash/issues/5731
 - [x] Add JVM GC stats under `/_node/stats/jvm` https://github.com/elastic/logstash/issues/5730
 - [x] Remove the `/_node/mem` resource as it's been properly moved under `/_node/jvm/mem` https://github.com/elastic/logstash/pull/5701#issuecomment-235728761
 - [x] Add config reload stats under new resource type /_node/stats/x #5680 (@jsvd)

- **Node Info**
 - [x] Add config reload boolean to `/_node/pipeline` https://github.com/elastic/logstash/issues/5681
 - [x] Add JVM GC strategy info under `/_node/jvm` https://github.com/elastic/logstash/issues/5755

- **Hot Threads**
 - [x] Ensure `?human` option works correctly for hot_threads API. Additionally, the output for `?human=true` appeared to be more verbose than the default format. Need to confirm scope of changes following alpha5 release. https://github.com/elastic/logstash/pull/5701#issuecomment-235728761

ph:
@Alvin I've looked at the hot thread human format and the JSON one, and I can say that they are currently the same. I've created a PR to fixes some small issue with it and adding a test to make sure the thread count is taken into consideration in the case.

acchen97:
@ph thanks, I'll spin up alpha5 today for a quick fact check here as well.

acchen97:
@ph doing some testing against alpha5. A few notes below, but will update the initial issue accordingly.

- Hot threads appears mostly fixed, the ?human=false appears to still show the human format though.
- The "mem" stats are still showing up in the `/_node` top level, this (along with the `/_node/mem` resource) should be removed as its now properly under `/_node/stats/jvm`
- For adding basic stats [in/out/duration_in_millis] to specific input plugins, created a new issue (https://github.com/elastic/logstash/issues/5769) to track the 12 plugin scope. @suyograo maybe divide/conquer makes sense?
- For additional plugin level stats, let's add a few more things for Beats input and ES output. This issue (https://github.com/elastic/logstash/issues/5428) can be used as reference, but I'll let you run with the scoping based on your discretion of importance vs size of work.

ph:
[...]

PR #5766, opened 2016-08-16
API: Hot Thread Human format fixes

ph:
This PR fixes the following:

- Make sure the `human=false` works
- Make sure that any other way of defining booleans in the query string works (true|false|t|f|0|1)
- A **human** key without a specific value will be considered as true.
- Add test around the number of threads and the human format.
- Fixes a typo in the Human text and make sure we use the localization files.

Related to #5732

### DIFF
--- a/logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
+++ b/logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 
 class HotThreadsReport
+  STRING_SEPARATOR_LENGHT = 80.freeze
   HOT_THREADS_STACK_TRACES_SIZE_DEFAULT = 10.freeze
 
   def initialize(cmd, options)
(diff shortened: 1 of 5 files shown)


logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
```diff
@@ -13,19 +14,16 @@ def initialize(cmd, options)
   def to_s
     hash = to_hash[:hot_threads]
     report =  "#{I18n.t("logstash.web_api.hot_threads.title", :hostname => hash[:hostname], :time => hash[:time], :top_count => @thread_dump.top_count )} \n"
-    report << '=' * 80
+    report << '=' * STRING_SEPARATOR_LENGHT
```
suyograo: Typo `STRING_SEPARATOR_LENGHT` => `STRING_SEPARATOR_LENGTH`

suyograo:
LGTM. Works as expected.

suyograo:
@ph Just fix the typo in the constant and merge!

logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
ph: oh that word is my own personnal curse.

PR #5770, opened 2016-08-16
Sync codec

### LINKED ISSUE #5768: Implement `Codec#multi_encode` / `Output#multi_receive_encoded`
By creating two new synchronous methods `Codec#multi_encode` / `Output#multi_receive_encoded` we can better support the work in the "Delegator Refactor" PR https://github.com/elastic/logstash/issues/5752 . 

While that PR does improve the current plugin situation it also leaves a giant hole, where `:shared` outputs cannot easily use codecs in a threadsafe way. By adding (and encouraging) these new methods, writing shared outputs will be much easier, a key component for their adoption and https://github.com/elastic/logstash/issues/5663

andrewvc:
This is a single new commit layered on top of #5752 . It should only be merged after that.

These methods allow outputs to receive their events pre-encoded for them
by the pipeline. This is mostly useful in the context of `#shared` outputs, for whom
encoding a discrete batch in a threadsafe way is not necessarily straightforward.

It would be advised for codecs to prefer `#multi_encode` as the main way of operating
as the standard `#encode` method is not threadsafe.

For an example of a plugin using this new API see this file on this branch https://github.com/andrewvc/logstash-output-stdout/blob/sync_codec/lib/logstash/outputs/stdout.rb

Fixes #5768

### DIFF
--- a/logstash-core/lib/logstash/codecs/base.rb
+++ b/logstash-core/lib/logstash/codecs/base.rb
@@ -18,6 +18,7 @@ module LogStash::Codecs; class Base < LogStash::Plugin
     super
     config_init(@params)
     register if respond_to?(:register)
+    setup_multi_encode!
   end
 
   public
@@ -28,9 +29,36 @@ module LogStash::Codecs; class Base < LogStash::Plugin
   alias_method :<<, :decode
 
   public
+  # DEPRECATED: Prefer defining encode_sync or multi_encode
   def encode(event)
-    raise "#{self.class}#encode must be overidden"
+    encoded = multi_encode([event])
+    encoded.each {|event,data| @on_event.call(event,data) }
   end # def encode
+
+  public
+  # Relies on the codec being synchronous (which they all are!)
+  # We need a better long term design here, but this is an improvement
+  # over the current API for shared plugins
+  # It is best if the codec implements this directly
+  def multi_encode(events)
+    if @has_encode_sync              
+      events.map {|event| [event, self.encode_sync(event)]}
+    else
+      batch = Thread.current[:logstash_output_codec_batch] ||= []
+      batch.clear
+      
+      events.each {|event| self.encode(event) }
+      batch
+    end
+  end
+
+  def setup_multi_encode!
+    @has_encode_sync = self.methods.include?(:encode_sync)
+
+    on_event do |event, data|
+      Thread.current[:logstash_output_codec_batch] << [event, data]
+    end
+  end
 
   public
   def close; end;
--- a/logstash-core/lib/logstash/outputs/base.rb
+++ b/logstash-core/lib/logstash/outputs/base.rb
@@ -68,6 +68,8 @@ class LogStash::Outputs::Base < LogStash::Plugin
     # If we're running with a single thread we must enforce single-threaded concurrency by default
     # Maybe in a future version we'll assume output plugins are threadsafe
     @single_worker_mutex = Mutex.new
+    
+    @receives_encoded = self.methods.include?(:multi_receive_encoded)
   end
 
   public
@@ -83,7 +85,15 @@ class LogStash::Outputs::Base < LogStash::Plugin
   public
   # To be overriden in implementations
   def multi_receive(events)
-    events.each {|event| receive(event) }
+    if @receives_encoded
+      self.multi_receive_encoded(codec.multi_encode(events))
+    else
+      events.each {|event| receive(event) }
+    end
+  end
+
+  def codec
+    params["codec"]
   end
 
   def concurrency
--- a/logstash-core/spec/logstash/codecs/base_spec.rb
+++ b/logstash-core/spec/logstash/codecs/base_spec.rb
@@ -1,0 +1,74 @@
+# encoding: utf-8
+require "spec_helper"
+
+DATA_DOUBLE = "data".freeze
+
+# use a dummy NOOP output to test Outputs::Base
+class LogStash::Codecs::NOOPAsync < LogStash::Codecs::Base
+  attr_reader :last_result
+  config_name "noop_async"
+
+  def encode(event)
+    @last_result = @on_event.call(event, DATA_DOUBLE)
+  end
+end
+
+class LogStash::Codecs::NOOPSync < LogStash::Codecs::Base
+  attr_reader :last_result
+  config_name "noop_sync"
+
+  def encode_sync(event)
+    DATA_DOUBLE
+  end
+end
+
+class LogStash::Codecs::NOOPMulti < LogStash::Codecs::Base
+  attr_reader :last_result
+  config_name "noop_multi"
+
+  def encode_sync(event)
+    DATA_DOUBLE
+  end
+end
+
+describe LogStash::Codecs::Base do
+  let(:params) { {} }
+  subject(:instance) { klass.new(params.dup) }
+  let(:event) { double("event") }
+  let(:encoded_data) { DATA_DOUBLE }
+  let(:encoded_tuple) { [event, encoded_data] }
+
+  describe "encoding" do
+    shared_examples "encoder types" do |codec_class|
+      let(:klass) { codec_class }
+      
+      describe "#{codec_class}" do
+        describe "multi_encode" do
+          it "should return an array of [event,data] tuples" do
+            expect(instance.multi_encode([event,event])).to eq([encoded_tuple, encoded_tuple])
+          end
+        end
+        
+        describe "#encode" do
+          before do
+            @result = nil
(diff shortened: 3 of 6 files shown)


jordansissel:
For helping with reviewing, at time of writing, the commit for this PR is here: https://github.com/elastic/logstash/pull/5770/commits/c344cb1552815c000807054cefb1ca64c1e90bb5

ph:
@jordansissel Thanks for the commit reference.

andrewvc:
@jordansissel @guyboertje I have updated this PR to:

1. Better exploit parallelism by assuming that codecs are threadsafe
2. Per @guyboertje 's suggestion create a new `#encode_sync` method that codec authors can use. The codec author can implement either `encode` `encode_sync` or `multi_encode` and Logstash will now do the right thing.

I've also added test cases covering all these options. I have amended the last commit and forced pushed, the commit to check is now https://github.com/elastic/logstash/pull/5770/commits/6fd92412c1e58a992a937c7a6c6b8f29fc3517b7

logstash-core/lib/logstash/codecs/base.rb
```diff
@@ -28,10 +29,56 @@ def decode(data)
...
+              batch = @thread_batches[Thread.current]
+              @multi_encode_mutex.synchronize { batch = @thread_batches[Thread.current] = [] } unless batch
+              batch.clear
+              
+              events.each {|event| self.encode(event) }
+              batch
+            end
+
+    if @on_event != @on_event_proc
+      batch.each {|e| @callback.call(e)}
```
jordansissel: Typo? I don't see `@callback` defined in this PR.

logstash-core/lib/logstash/codecs/base.rb
```diff
@@ -28,10 +29,56 @@ def decode(data)
...
+    data
+  end
+
+  public
+  # Relies on the codec being synchronous (which they all are!)
+  # We need a better long term design here, but this is an improvement
+  # over the current API for shared plugins
+  # It is best if the codec implements this directly
+  def multi_encode(events)
+    batch = if @has_encode_sync
```
jordansissel: This long `if..else..end` should be moved to a method by itself to improve readability.

logstash-core/lib/logstash/codecs/base.rb
```diff
@@ -28,10 +29,56 @@ def decode(data)
...
+    batch
+  end
+
+  def setup_multi_encode!
+    @has_encode_sync = self.methods.include?(:encode_sync) && self.class.instance_method(:encode_sync) == LogStash::Codecs::Base.instance_method(:encode_sync)
+    @thread_batches = {}
+    @multi_encode_mutex = Mutex.new
+
+    @on_event_proc = proc do |event, data|
+      @thread_batches[Thread.current] << [event, data]
```
jordansissel: Thread#[] should be used, imo, instead of a shared-locked hash object who's key is a Thread.

Example:

```
Thread.current[:codec_batches] << [event, data]
```

logstash-core/lib/logstash/codecs/base.rb
```diff
@@ -28,10 +29,56 @@ def decode(data)
...
+            else
+              batch = @thread_batches[Thread.current]
+              @multi_encode_mutex.synchronize { batch = @thread_batches[Thread.current] = [] } unless batch
+              batch.clear
+              
+              events.each {|event| self.encode(event) }
+              batch
+            end
+
+    if @on_event != @on_event_proc
```
jordansissel: I squint at this and wonder what the meaning of this check is. Can you move this condition `@on_event != @on_event_proc` to a method call with a descriptive name? Something like

```ruby
def multi_encode?
  @on_event != @on_event_proc
end
```

logstash-core/lib/logstash/codecs/base.rb
```diff
@@ -28,10 +29,56 @@ def decode(data)
...
+      batch.each {|e| @callback.call(e)}
+    end
+    
+    batch
+  end
+
+  def setup_multi_encode!
+    @has_encode_sync = self.methods.include?(:encode_sync) && self.class.instance_method(:encode_sync) == LogStash::Codecs::Base.instance_method(:encode_sync)
```
jordansissel: I propose a class method instead for this:

```ruby
def self.has_encode_sync?
    instance_methods.include?(:encode_sync) && instance_method(:encode_sync) == LogStash::Codecs::Base.instance_method(:encode_sync)
```

logstash-core/lib/logstash/codecs/base.rb
```diff
@@ -28,10 +29,56 @@ def decode(data)
...
+  # Relies on the codec being synchronous (which they all are!)
+  # We need a better long term design here, but this is an improvement
+  # over the current API for shared plugins
+  # It is best if the codec implements this directly
+  def multi_encode(events)
+    batch = if @has_encode_sync              
+              events.map {|event| [event, self.encode_sync(event)]}
+            else
+              batch = @thread_batches[Thread.current]
+              @multi_encode_mutex.synchronize { batch = @thread_batches[Thread.current] = [] } unless batch
```
jordansissel: If we store the batch array in `Thread.current#[...]=` we won't need a mutex.

logstash-core/lib/logstash/outputs/base.rb
```diff
@@ -63,10 +63,7 @@ def self.plugin_type
   def initialize(params={})
     super
     config_init(@params)
-
-    # If we're running with a single thread we must enforce single-threaded concurrency by default
-    # Maybe in a future version we'll assume output plugins are threadsafe
-    @single_worker_mutex = Mutex.new
+    @receives_encoded = self.methods.include?(:multi_receive_encoded)
```
jordansissel: I think this should be a predicate method instead:

```
def receives_encode?
  methods.include?(:multi_receive_encoded)
end
```

logstash-core/spec/logstash/codecs/base_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+          end
+          
+          it "should yield the correct result" do
+            expect(@result).to eq(encoded_tuple)
+          end
+        end
+      end
+    end
+
+    include_examples("encoder types", LogStash::Codecs::NOOPAsync)
```
jordansissel: Love the use of shared examples.

logstash-core/lib/logstash/codecs/base.rb
andrewvc: Yes, I'll be removing this in my next push

logstash-core/lib/logstash/codecs/base.rb
andrewvc: This will be going away in my next refactor! :)

logstash-core/lib/logstash/codecs/base.rb
andrewvc: Will be removed in my next push!

logstash-core/lib/logstash/codecs/base.rb
andrewvc: Hmm, I tend to dislike using `Thread.current` because its like using global variables, easy to step on people's toes. 

The perf loss here is very small since the shared lock is only used once per batch should be relatively low contention since its only used to check that there is an Array object available. The writes to the array objects are not synchronized since they are only ever used by a single thread and so can omit them.

I'm OK to make this change if you feel strongly about it however.

logstash-core/lib/logstash/outputs/base.rb
andrewvc: Since this is called once per batch I wanted to only calculate this once.

I could make a method with `||=` but that seems like adding complexity.

logstash-core/spec/logstash/codecs/base_spec.rb
andrewvc: Thanks! These were fun to write :)

andrewvc:
@jordansissel I believe I have addressed most of your concerns with this latest push. You left a comment about the use of `Thread.current`hat was unfortunately obliterated because the line changed for other reasons. 

My feeling was that `Thread.current[:key]` is messy and is a sort of global variable, which I usually regard as a smell. I think the performance will be good as is. That said, if you feel strongly about it I'm glad to change the code, but I just wanted to make sure you heard my concern first.

andrewvc:
@jordansissel after some consideration I think it makes sense to use the internal hash for `Thread.current` I will do this in a subsequent patch.

andrewvc:
@guyboertje  @jordansissel The lib has now been patched.

guyboertje:
LGTM

jordansissel:
LGTM

Issue #5531, opened 2016-06-21
You can start Logstash with no worker and it will not complain.


ph:
When testing logstash, I was able to make logstash start with no worker.
Logstash doesn't complain it just sit waiting.

```
 bin/logstash -f ls_tools/generator.conf -w
```

```
~/e/logstash git:fix/pipeline-response-for-custom-metric ❯❯❯ bin/logstash -f ls_tools/generator.conf -w                                                                                                                                                                                                               ⏎ ✱ ◼
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
Pipeline main started
```

I haven't check if we have validation of the setting for the worker.

ph:
@jsvd could you take a look?

jsvd:
closed by #5773

PR #5773, opened 2016-08-17
add more numeric setting classes and control cli numeric coersion

jsvd:
fixes https://github.com/elastic/logstash/issues/5531

* CoercibleSetting: a setting that receives something and tries to coerce it to fit its type, like:
  * Boolean: converts from strings "true" and "false"
  * Numeric: converts from strings that represent numbers
* Created Integer and PositiveInteger setting classes
* Removed `&:to_i` coercion in Clamp option declaration which was causing #5531 by converting nil into 0

### DIFF
--- a/logstash-core/lib/logstash/environment.rb
+++ b/logstash-core/lib/logstash/environment.rb
@@ -21,12 +21,12 @@ module LogStash
             Setting::String.new("config.string", nil, false),
            Setting::Boolean.new("config.test_and_exit", false),
            Setting::Boolean.new("config.reload.automatic", false),
-           Setting::Numeric.new("config.reload.interval", 3),
+           Setting::Numeric.new("config.reload.interval", 3), # in seconds
            Setting::Boolean.new("metric.collect", true) {|v| v == true }, # metric collection cannot be disabled
             Setting::String.new("pipeline.id", "main"),
-           Setting::Numeric.new("pipeline.workers", LogStash::Config::CpuCoreStrategy.maximum),
-           Setting::Numeric.new("pipeline.output.workers", 1),
-           Setting::Numeric.new("pipeline.batch.size", 125),
+   Setting::PositiveInteger.new("pipeline.workers", LogStash::Config::CpuCoreStrategy.maximum),
+   Setting::PositiveInteger.new("pipeline.output.workers", 1),
+   Setting::PositiveInteger.new("pipeline.batch.size", 125),
            Setting::Numeric.new("pipeline.batch.delay", 5), # in milliseconds
            Setting::Boolean.new("pipeline.unsafe_shutdown", false),
                     Setting.new("path.plugins", Array, []),
--- a/logstash-core/lib/logstash/settings.rb
+++ b/logstash-core/lib/logstash/settings.rb
@@ -176,18 +176,43 @@ module LogStash
       end
     end
 
-    ### Specific settings #####
-
-    class Boolean < Setting
-      def initialize(name, default, strict=true, &validator_proc)
+    class Coercible < Setting
+      def initialize(name, klass, default=nil, strict=true, &validator_proc)
         @name = name
-        @klass = Object
+        unless klass.is_a?(Class)
+          raise ArgumentError.new("Setting \"#{@name}\" must be initialized with a class (received #{klass})")
+        end
+        @klass = klass
+        @validator_proc = validator_proc
         @value = nil
         @value_is_set = false
-        @validator_proc = validator_proc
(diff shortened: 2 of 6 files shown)


logstash-core/lib/logstash/environment.rb
```diff
@@ -21,12 +21,12 @@ module Environment
...
            Setting::Boolean.new("config.test_and_exit", false),
            Setting::Boolean.new("config.reload.automatic", false),
-           Setting::Numeric.new("config.reload.interval", 3),
+           Setting::Numeric.new("config.reload.interval", 3), # in seconds
            Setting::Boolean.new("metric.collect", true) {|v| v == true }, # metric collection cannot be disabled
             Setting::String.new("pipeline.id", "main"),
-           Setting::Numeric.new("pipeline.workers", LogStash::Config::CpuCoreStrategy.maximum),
-           Setting::Numeric.new("pipeline.output.workers", 1),
-           Setting::Numeric.new("pipeline.batch.size", 125),
+   Setting::PositiveInteger.new("pipeline.workers", LogStash::Config::CpuCoreStrategy.maximum),
```
purbon: While I do understand this is a good option, I wonder if you considered to keep this as before ```Setting::[Type].new``` and then add some kind of method to do the validation later on. What do you think?

logstash-core/lib/logstash/settings.rb
```diff
@@ -176,18 +176,43 @@ def validate(value)
       end
     end
 
-    ### Specific settings #####
-
-    class Boolean < Setting
-      def initialize(name, default, strict=true, &validator_proc)
+    class Coercible < Setting
```
purbon: Does this means we can have settings that could be `Coercible` and some that not?

logstash-core/lib/logstash/settings.rb
```diff
@@ -200,25 +225,61 @@ def coerce(value)
...
-        @value = coerce(coerced_value)
-        @value_is_set = true
-        @value
+    class Numeric < Coercible
+      def initialize(name, default=nil, strict=true)
+        super(name, ::Numeric, default, strict)
+      end
+
+      def coerce(v)
+        return v if v.is_a?(::Numeric)
```
purbon: Totally necessary the ```::Numeric```, however here I am on the fense of

* Using the same number and force people to use the `::` before the name
* Using an specific name that does not collide.

Can you share your pros and cons for going that way? this would be awesome to we can make sure to reduce the error rate when in a few weeks we need to change stuff here and use the `is_a?`, and alike, comparisons.

purbon:
Good work Joao. Left a few comments, mainly around where and how coercion is done, but in general the code looks good and test pass, I am in general good with this being merged!

logstash-core/lib/logstash/environment.rb
jsvd: it is indeed possible to do `Setting::Integer.new("pipeline.batch.size", 125) {|v| v > 0}`, I only created the PositiveInteger class to provide a better error message, otherwise it will return a generic "failed to validate" error.

logstash-core/lib/logstash/settings.rb
jsvd: Yes, some actually try to coerce the input into its type, while most of the others don't.
For example the String setting will fail for something which isn't a String, a Numeric will try to convert a String into a Numeric

logstash-core/lib/logstash/settings.rb
jsvd: Instead of calling it Numeric we can call Number, but we have the same problem in the String setting.
So the only work "fix" here is calling it NumericSetting, StringSetting, etc.
Other suggestions?

logstash-core/lib/logstash/environment.rb
purbon: I understand, I guess one future direction of this, together with upcoming config refactoring would be properly normalize coercion code with the option to be able to return true|false and the error description.

logstash-core/lib/logstash/settings.rb
purbon: makes sense.

logstash-core/lib/logstash/settings.rb
purbon: not much more than that really, end of the day for this settings code is more scoped, so the "risk" or surprises is lower. If we encounter this as a problem we can alware rethink.

purbon:
LGTM ™️

ph:
LGTM, @jsvd Maybe in a future PR we can split the setting.rb into smaller one for clarity.

PR #5774, opened 2016-08-17
--http.port Now accept a port range

ph:
This PR does a few modifications for our webserver:

- Add a **PortRange** setting type.
- allow the `--http.port` option to accept a port range or a single port, when a range is provided Logstash will incrementally try this list to find a free port for the webserver.
- Logstash will report in the log which port it is using. (INFO LEVEL)
- It refactors a few methods of the webserver.
- It adds test for the binding of the ports.

## Notes:

Port range can be defined in the `logstash.yml` or pass via the command line like this.

```
 bin/logstash -e 'input { generator {}} output { null {}}' --log.level verbose --http.port 2000-2005
```

Fixes: #5326

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -91,9 +91,10 @@
 #
 # http.host: "127.0.0.1"
 #
-# Bind port for the metrics REST endpoint
+# Bind port for the metrics REST endpoint, this option also accept a range
+# (9600-9700) and logstash will pick up the first available ports.
 #
-# http.port: 9600
+# http.port: 9600-9700
 #
 # ------------ Debugging Settings --------------
 #
--- a/docs/static/command-line-flags.asciidoc
+++ b/docs/static/command-line-flags.asciidoc
@@ -90,7 +90,8 @@ added[5.0.0-alpha3, Command-line flags have dots instead of dashes in their name
   The bind address for the metrics REST endpoint. The default is "127.0.0.1".
 
 *`--http.port HTTP_PORT`*::
-  The bind port for the metrics REST endpoint. The default is 9600.
+  The bind port for the metrics REST endpoint. The default is 9600-9700.
+  This settings accept a range of the format 9600-9700 and Logstash will pick up the first available port.
 
 *`--pipeline.unsafe_shutdown`*::
   Force Logstash to exit during shutdown even if there are still inflight events
--- a/logstash-core/lib/logstash/settings.rb
+++ b/logstash-core/lib/logstash/settings.rb
@@ -278,8 +278,51 @@ module LogStash
     end
 
     class Port < Integer
+      VALID_PORT_RANGE = 1..65535
+
       def initialize(name, default=nil, strict=true)
-        super(name, default, strict) {|value| value >= 1 && value <= 65535 }
+        super(name, default, strict) { |value| valid?(value) }
+      end
+
+      def valid?(port)
+        VALID_PORT_RANGE.cover?(port)
+      end
+    end
+
+    class PortRange < Coercible
+      PORT_SEPARATOR = "-"
+
+      def initialize(name, default=nil, strict=true)
+        super(name, ::Range, default, strict=true) { |value| valid?(value) }
+      end
+
+      def valid?(range)
+        Port::VALID_PORT_RANGE.first <= range.first && Port::VALID_PORT_RANGE.last >= range.last
+      end
+
+      def coerce(value)
+        case value
+        when ::Range
+          value
+        when ::Fixnum
+          value..value
+        when ::String
+          first, last = value.split(PORT_SEPARATOR)
+          last = first if last.nil?
+          begin
+            (Integer(first))..(Integer(last))
+          rescue ArgumentError # Trap and reraise a more human error
+            raise ArgumentError.new("Could not coerce #{value} into a port range")
+          end
+        else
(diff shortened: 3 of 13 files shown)


config/logstash.yml
```diff
@@ -91,7 +91,8 @@
 #
 # http.host: "127.0.0.1"
 #
-# Bind port for the metrics REST endpoint
+# Bind port for the metrics REST endpoint, this option also accept a range
+# (9000-10000) and logstash will pick up the first available ports.
```
suyograo: We should not allow `9000-10000`. That's a big range and can collide with ES and lots of other applications. How about `9600-9700` inclusive?

docs/static/command-line-flags.asciidoc
```diff
@@ -90,7 +90,8 @@ added[5.0.0-alpha3, Command-line flags have dots instead of dashes in their name
   The bind address for the metrics REST endpoint. The default is "127.0.0.1".
 
 *`--http.port HTTP_PORT`*::
-  The bind port for the metrics REST endpoint. The default is 9600.
+  The bind port for the metrics REST endpoint. The default is 9600-9700.
+  This settings accept a range of the format 9600-9700 and Logstash will pick up the first available port.
```
suyograo: oh sorry didn't see this in context of the above comment.

logstash-core/spec/logstash/settings/port_range_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    it "can update the range" do
+      subject.set(10000)
+      expect(subject.value).to eq(10000..10000)
+    end
+  end
+
+  context "When the value is a string" do
+    subject { LogStash::Setting::PortRange.new("mynewtest", "9000-10000") }
```
suyograo: Can we add a test covering the default range?

logstash-core/lib/logstash/settings.rb
```diff
@@ -217,8 +217,58 @@ def initialize(name, default=nil, strict=true)
...
       def initialize(name, default=nil, strict=true)
-        super(name, ::Numeric, default, strict) {|value| value >= 1 && value <= 65535 }
+        super(name, ::Range, coerce(default), strict=true)
+      end
+
+      def valid?(range)
+        Port::VALID_PORT_RANGE.first <= range.first && Port::VALID_PORT_RANGE.last >= range.last
+      end
+
+      def coerce(value)
```
jsvd: In https://github.com/elastic/logstash/pull/5773 I added a Coercible subclass to Setting for those Settings that need to use a coerce method.
Can you consider reviewing that PR and perhaps adapting the PortRange setting to be Coercible?

config/logstash.yml
ph: just for clarity and reduce confusion, I will use the 9600-9700 as the example.

logstash-core/lib/logstash/settings.rb
ph: @jsvd I reviewed the #5773 and I will gladly rebase this PR on top of it with the coercible subclass.

logstash-core/spec/logstash/settings/port_range_spec.rb
ph: The default range is not set here, but in https://github.com/elastic/logstash/pull/5774/files/b7509e80ea0a4a4b8eff35d15f46a22d62e93c97#diff-b803768250328614e95dee4f5b27af18R4, this spec file is only for testing the `PortRange` class.

To correctly test the default port it require a new assertion in`qa/acceptance`, will check on hard it is to add it.

ph:
@jsvd I've rebased this branch with master and used the new coercible class.

@suyograo Concerning the tests, I have added examples in the runner_spec to cover the usage of `---http.port, ---http.host` to make sure the values are the expected one in the agent/webserver.

We also have tests covering the logic of the webserver class when a port is taken or when all port are taken. So I think the tests are in a good position now.

logstash-core/lib/logstash/settings.rb
```diff
@@ -278,8 +278,48 @@ def initialize(name, default=nil, strict=true)
...
-        super(name, default, strict) {|value| value >= 1 && value <= 65535 }
+        super(name, ::Range, default, strict=true) { |value| valid?(value) }
+      end
+
+      def valid?(range)
+        Port::VALID_PORT_RANGE.first <= range.first && Port::VALID_PORT_RANGE.last >= range.last
+      end
+
+      def coerce(value)
+        if value.is_a?(::Range)
```
jsvd: maybe instead do:

```ruby
case value
when ::Range
  value
when ::Fixnum
  value..value
when ::String
   # ...
else
end
```

jsvd:
LGTM

logstash-core/lib/logstash/settings.rb
ph: Also I need to have a catch all `else`, adding a test for it.

logstash-core/spec/logstash/settings/port_range_spec.rb
ph: There is now a test for the default range https://github.com/elastic/logstash/pull/5774/files#diff-325a7dbc35b78efb1a9ff395d1cd2ffcR238

suyograo:
LGTM

Issue #5778, opened 2016-08-18
Enhanced config test semantics 


purbon:
In the current logstash codebase we use a [config mixin](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/config/mixin.rb) as the way to handle plugin configuration, this mixin is the responsible of (_short version_):

* Parameter validation, making sure that for example required parameters are there.
* Value validation where mostly checks are performed to be sure that the right data type is used, some might include a degree of semantics verification.
* Value coercion, this phase converts the initial value to the right expected one, for example when we deal with passwords.

This class has lived a long life since the beginning of logstash, well @jordansissel might provide here his historician background for sure, but certainly it has it's difficulties for:

* Maintenance, adding new features requires to understand a complex structure that has intrinsic relations between the different methods and responsibilities.
* Testing, because of the way code is structured makes tests require an initial setup making them hard to compose.

In this issue I propose a strategy to enhance this module that will provide:

* A cleaner interface and separation of concerns into different objects responsible of parameters and validation, coercion, etc.
* An easier interface for testing by decoupling the mixin from the actual code responsible of the different tasks.
* An small cleanup of old features like, `Regexp` , `milestones`, etc.
* An extension to the current validators:
 * to let users have custom classes and Procs/Blocks, basically enabling a handy extension point for more semantic validation.
 * Perform cross validations, currently being done in the [register method](https://github.com/logstash-plugins/logstash-input-jdbc/blob/master/lib/logstash/inputs/jdbc.rb#L208) when necessary.

### Validators and Coercers

In the new validation mechanism I propose to break out the config initialization into:

* Validators, simple set of classes responsible of making sure that a required parameter exist, a value has the right type, etc.
* Coercers, another simple set of classes basically responsible of converting config values to the right type.

A validator might be as simple as:

```ruby

module LogStash::Config
  module TypeValidators
    class Ipaddr < Abstract

      def valid?(value)
        octets = value.split(".")
        if octets.length != 4
          add_errors "Expected IPaddr, got #{value.inspect}"
          return false
        end
        octets.each do |o|
          if (o.to_i < 0 || o.to_i > 255)
            add_errors "Expected IPaddr, got #{value.inspect}"
            return false
          end
        end
      end

    end
  end
end
```

Validators could be:

* A Symbol, the former method that will basically pull a logstash-core validator for it.
* A Proc, or anything that respond to call like lambda.
* A custom class provided by the plugin.

Coercer are like:

```ruby

[...]

PR #5779, opened 2016-08-18
[WiP] Enhanced config validation mechanism for logstash plugins

purbon:
This PR has been created as work in progress example to have initial revision.
- introduction of validators and coercers to simplify to simplify the config mixin.
- add default code to handle class and proc based validators
- add the option to pass the params object to the validators to they can actually do cross validation

There are a lot of things that can, could and for sure will change in this PR, opened this as discussed with @jordansissel and @jsvd to have a tool to discuss.
### TODO (pending):
- [ ] Reuse and simplify coercers and validators organization, for example with settings.
- [ ] Improve test coverage and organization.
### Test:

Current test pas at core level and also locally I updated the file and jdbc inputs and the date filter with passing tests, as this touches and important component of LS, more detailed test will be pulled to be sure there no missing bit pending.
## Notes to reviewers

Please focus on:
- If you like the idea and think is powerful enough to help improve the config tests.
- Do you miss an abstraction? could a class be written differently? 

Think will be a living creature to don't think of this as something final, we might not even merge it 😸.

For more details check #5778


purbon:
@andrewvc @jordansissel  @guyboertje  you might like to take a look here, this is actually the first sketch with happy tests all over from the idea I just shared with you.

@jsvd this is the PR we spoke about this morning, looking forward to see how the setting and this coercions could be DRY'ed.

kares:
seems this WiP is no longer relevant. should be restarted with a clean slate if anyone feels otherwise.

Issue #5681, opened 2016-07-22
Have /_node/ show if Logstash is configured to automatically reload configs


jordansissel:
Suggested by @widhalmt in https://github.com/widhalmt/check_logstash/issues/6.

acchen97:
+1 this can be an addition to the settings in /_node/pipeline

PR #5780, opened 2016-08-18
Adding the values for the automatic reloading to the API

ph:
```
config_reload_automatic: true,
config_reload_interval: 3
```

Fixes: #5681

### DIFF
--- a/docs/static/monitoring-apis.asciidoc
+++ b/docs/static/monitoring-apis.asciidoc
@@ -110,7 +110,10 @@ Example response:
   "pipeline": {
     "workers": 8,
     "batch_size": 125,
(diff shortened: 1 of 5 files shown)


ph:
@jordansissel Concerning this PR, It made me think should we filter out some options depending on other option?

IE: if `config_reload_automatic` is set to false we don't output `config_reload_interval`.

I am confused if we should do that or not. I am afraid that if we do it we might lose some important configuration in the API and it introduce a concept of dependency.

ph:
Part of https://github.com/elastic/logstash/issues/5732 effort

jsvd:
LGTM

Issue #5594, opened 2016-07-06
Deprecate --verbose and --debug


jordansissel:
In master/5.0, we removed --verbose and --debug. This was the right decision, but for reducing user pain, I think we should restore them with warnings about them being deprecated.

* `--debug` should warn users to use `--log.level debug` instead, and also imply that setting (debug)
* `--verbose` should warn users to use `--log.level info` instead, and imply that setting (verbose)

jsvd:
closed by #5785

PR #5785, opened 2016-08-19
allow deprecated quiet/debug/verbose flags to be set

jsvd:
fixes https://github.com/elastic/logstash/issues/5594

adds a few classes to clamp to allow creating deprecated options that warn about deprecations but set the proper new flags accordingly.


guyboertje:
LGTM

PR #5782, opened 2016-08-19
Release 2.4.0.snapshot1

suyograo:
This PR moves from 2.4.0.dev1 to snapshot1 in prep for a release. I also published
required gems to RubyGems so it no longer uses locally built gems.

Remember that 2.4.0 core can install 5.x plugins. So, I had to curate the Gemfile to not drag in 5.x plugins. It was pretty easy and I added the version constraints.

<details>
<summary>Gemfile diff between 2.4.0 and 2.3.4</summary>
```
suyog@machine:/tmp$ diff 2.3.4.plugins 2.4.0.plugins
diff --git a/2.3.4.plugins b/2.4.0.plugins
index 9420bec..bae14b9 100644
--- a/2.3.4.plugins
+++ b/2.4.0.plugins
@@ -10,7 +10,7 @@ logstash-codec-json_lines (2.1.3)
logstash-codec-line (2.1.2)
logstash-codec-msgpack (2.0.4)
logstash-codec-multiline (2.0.11)
logstash-codec-netflow (2.1.0)(2.1.1)
logstash-codec-oldlogstashjson (2.0.4)
logstash-codec-plain (2.0.4)
logstash-codec-rubydebug (2.0.7)
@@ -53,8 +53,8 @@ logstash-input-http (2.2.3)
logstash-input-http_poller (2.0.6)
logstash-input-imap (2.0.5)
logstash-input-irc (2.0.5)
logstash-input-jdbc (3.0.2)(3.1.0)
logstash-input-kafka (2.0.8)(2.0.9)
logstash-input-log4j (2.0.7)
logstash-input-lumberjack (2.0.7)
logstash-input-pipe (2.0.4)
@@ -92,7 +92,7 @@ logstash-output-null (2.0.4)
logstash-output-opentsdb (2.0.4)
logstash-output-pagerduty (2.0.4)
logstash-output-pipe (2.0.4)
logstash-output-rabbitmq (3.1.0)(3.1.2)
logstash-output-redis (2.0.5)
logstash-output-s3 (2.0.7)
logstash-output-sns (3.0.4)
suyog@machine:/tmp$
```

</details>

<details>
<summary>Tested update plugin on 2.4.0</summary>

I was surprised that `bin/plugin install --version` will bump to 5.x plugins, but `bin/plugin update` will not.

```
suyog@machine:/tmp/logstash-2.4.0.snapshot1$ bin/logstash-plugin install --version 5.0.3 logstash-input-kafka
Validating logstash-input-kafka-5.0.3
[...]

PR #5790, opened 2016-08-19
Override default behavior of travis to use `bundle install` and uses rake test:core instead

ph:
This PR fixes the issue discovered in https://github.com/elastic/logstash/pull/5782#issuecomment-241070120

When we set the language to ruby and if Travis detects a Gemfile, it will automagically run bundle install, in the PR this is where it failed.


suyograo:
LGTM

PR #5720, opened 2016-08-02
Rewrite of the doc generation

ph:
**Motivation**: We have decided to rewrite the documentation generator to add more features
to our documentation and to fix some outstanding bugs.

This PR addresses the following problem:
- Hard to add new features to the code base.
- The code was tied with Logstash.
- No easy way to publish the doc for a single plugin
- The plugin author was not able to test their documentation changes
- The reported errors were hard to understand.
- No easy way to automate it.
- the minimal requirement on base classes.
# Features
## Generating the documentation for a Logstash instance

You can now generate the documentation for a Logstash distribution.

_prerequisite_:
- You need to have installed the plugins inside Logstash; this is necessary to pickup the right file when generating the documentation.
- You need to have the development gems installed. `bin/logstash-plugin --no-verify --development`

If you have all of the above you can run this command:

``` sh
rake docs:generate-plugins
rake doc:generate-plugins[/tmp/the-new-doc] # Depending on the shell you are using you might need to quote the task name, like this rake "doc:generate-plugins[/tmp/new-doc]"

```

This will give you an output similar to this

``` sh
logstash-input-file > SUCCESS
logstash-input-s3 > SUCCESS
logstash-input-kafka > FAIL
[...]
Exceptions: XXXX
```

The generator will try to generate the doc for all the plugins defined in the _Gemfile_ and installed in Logstash, if anything goes wrong it won't stop the generation of the other plugin. The Task will also report any errors with stack traces at the end, if one plugin fails the build,
you can interrupt the process, and it will output the current errors before exiting.

The docs will be generated in `build/docs/`.
## Generating the documentation for all the plugins from the organization

You can now generate the documentation from the master for all the plugin in the _logstash-plugins_ organization.

_prerequisite_
- You need to go in the `tools/logstash-docgen` directory
- You need to have the dependency installed.
  
  To get started you can run the following commands:
  
  ``` sh
  cd tools/logstash-docgen
  bundle install
  ```

You can use the the `bin/logstash-docgen` command to generate any plugin that you want; this executable can generate all the plugins or specific one from their master branch.

Usages:

``` sh
bin/logstash-docgen --all # will generate the doc for all the plugins
bin/logstash-docgen logtash-input-file logstash-input-s3 # generate doc for 2 plugins
```

**See:** `bin/logstash-docgen --help` for complete usage.

**Notes:**
- The nature of theses tasks require a lot of external execution to make sure all the doc are done in isolation, this process can take a long time.
[...]

PR #5793, opened 2016-08-19
Make the default list of plugins bundled with logstash a json file

ph:
Rebase of #5096


ph:
I've moved this commit to #5720

PR #5794, opened 2016-08-19
Cleanup OutputDelegatorStrategy refactor

andrewvc:
This fixes two issues:

1. This fully removes the xopts parameter which for Shared and Single concurrency would prevent outputs from receiving their parameters
2. This cleans up the injection of the OutputDelegatorStrategyRegistry.

This also bumps the plugin api version to 2.1.12

### DIFF
--- a/logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
+++ b/logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
@@ -1,2 +1,2 @@
 # encoding: utf-8
-LOGSTASH_CORE_PLUGIN_API = "2.1.7"
+LOGSTASH_CORE_PLUGIN_API = "2.1.12"
--- a/logstash-core/lib/logstash/output_delegator.rb
+++ b/logstash-core/lib/logstash/output_delegator.rb
@@ -12,7 +12,8 @@ module LogStash class OutputDelegator
     @output_class = output_class
     @metric = metric
     @plugin_args = plugin_args
-    @strategy_registry ||= ::LogStash::OutputDelegatorStrategyRegistry.instance
+    @strategy_registry = strategy_registry
+    raise ArgumentError, "No strategy registry specified" unless strategy_registry
     raise ArgumentError, "No ID specified! Got args #{plugin_args}" unless id
     
     build_strategy!
--- a/logstash-core/lib/logstash/output_delegator_strategies/legacy.rb
+++ b/logstash-core/lib/logstash/output_delegator_strategies/legacy.rb
@@ -2,7 +2,7 @@
 module LogStash module OutputDelegatorStrategies class Legacy
   attr_reader :worker_count, :workers
   
-  def initialize(logger, klass, metric, plugin_args={})
+  def initialize(logger, klass, metric, plugin_args)
     @worker_count = (plugin_args["workers"] || 1).to_i
     @workers = @worker_count.times.map {|t| klass.new(plugin_args)}
     @worker_queue = SizedQueue.new(@worker_count)
--- a/logstash-core/lib/logstash/output_delegator_strategies/shared.rb
+++ b/logstash-core/lib/logstash/output_delegator_strategies/shared.rb
@@ -1,5 +1,5 @@
 module LogStash module OutputDelegatorStrategies class Shared
-  def initialize(logger, klass, metric, xopts={}, plugin_args={})
+  def initialize(logger, klass, metric, plugin_args)
     @output = klass.new(plugin_args)
   end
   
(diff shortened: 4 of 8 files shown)


logstash-core/lib/logstash/output_delegator_strategies/legacy.rb
```diff
@@ -2,7 +2,7 @@
 module LogStash module OutputDelegatorStrategies class Legacy
   attr_reader :worker_count, :workers
   
-  def initialize(logger, klass, metric, plugin_args={})
+  def initialize(logger, klass, metric, plugin_args)
```
andrewvc: There was no need for defaults, and having them led to this bug in the other strategies

logstash-core/lib/logstash/output_delegator_strategies/shared.rb
```diff
@@ -1,5 +1,5 @@
 module LogStash module OutputDelegatorStrategies class Shared
-  def initialize(logger, klass, metric, xopts={}, plugin_args={})
```
andrewvc: The issue was `plugin_args` was being passed as `xopts`

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -12,7 +12,7 @@ def initialize(logger, output_class, metric, strategy_registry, plugin_args)
     @output_class = output_class
     @metric = metric
     @plugin_args = plugin_args
-    @strategy_registry ||= ::LogStash::OutputDelegatorStrategyRegistry.instance
+    @strategy_registry = strategy_registry || ::LogStash::OutputDelegatorStrategyRegistry.instance
```
talevy: when is `strategy_registry` passed in as a `nil`?

logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
```diff
@@ -1,2 +1,2 @@
 # encoding: utf-8
-LOGSTASH_CORE_PLUGIN_API = "2.1.7"
+LOGSTASH_CORE_PLUGIN_API = "2.1.12"
```
talevy: I'm probably missing something, why the multipoint jump?

logstash-core/lib/logstash/output_delegator.rb
andrewvc: Never, but I like keeping it this way to keep it clean for future testing.

logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
andrewvc: Our git history sadly is lagging behind actual releases: https://rubygems.org/gems/logstash-core-plugin-api/versions/1.2.0-java

logstash-core/lib/logstash/output_delegator.rb
talevy: would it be better to use default values in arguments?

logstash-core-plugin-api/lib/logstash-core-plugin-api/version.rb
talevy: oh, probably because of the release branches. ++

logstash-core/lib/logstash/output_delegator.rb
andrewvc: That is a good call, but I really don't want to have a default value for `plugin_args`. Thinking it through, I'll just remove the `||`. Tests can just be forced to pass a registry.

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -12,7 +12,8 @@ def initialize(logger, output_class, metric, strategy_registry, plugin_args)
     @output_class = output_class
     @metric = metric
     @plugin_args = plugin_args
-    @strategy_registry ||= ::LogStash::OutputDelegatorStrategyRegistry.instance
+    @strategy_registry = strategy_registry
```
jordansissel: I am cautious of this variable name, btw. It shadows the accessor method `strategy_registry`. I don't have any particular recommendations, though. :\

andrewvc: Oh ruby :). I know its weird, but IMHO it is a ruby idiom.

andrewvc:
@talevy @jordansissel Does this need anything to get to LGTM?

talevy:
@andrewvc it needs a trademark filing

LG™

logstash-core/lib/logstash/output_delegator.rb
guyboertje: This is a problem because it is not clear in the next line whether your intention is to test for missing arg `strategy_registry` or for missing setting via the attr_reader.

Consider changing the arg name to perhaps `strat_registry` or using the ivar in line 16 (but then line 17 becomes weird e.g. ivar in L16 and attr_reader in L17)

talevy: is this still an issue even though there should be consistent scoping for such a variable?

```ruby
class Test
  attr_reader :foo

  def initialize(foo)
    @foo = foo
    if foo
      puts "OK"
    else
      puts "NOT OK"
    end
  end
end

test = Test.new("foo")
```

prints:

```
"OK"
```

PR #5801, opened 2016-08-22
Cleanup OutputDelegatorStrategy refactor (5.x port)

andrewvc:
This fixes two issues:

1. This fully removes the xopts parameter which for Shared and Single concurrency would prevent outputs from receiving their parameters
2. This cleans up the injection of the OutputDelegatorStrategyRegistry.

This also bumps the plugin api version to 2.1.12

Fixes #5794


suyograo:
LGTM

PR #5809, opened 2016-08-22
Downgrade FPM and clamp versions to 2.3 levels

suyograo:
This was causing Jenkins to not build deb packages as seen here: http://build-eu-00.elastic.co/job/logstash_create_deb_artifact_24/jdk=JDK8,label=metal-pool/2/console

Also, this previous version bump was clearly an oversight since [2.3](https://github.com/elastic/logstash/blob/2.3/Gemfile#L18) and [5.0](https://github.com/elastic/logstash/blob/5.0/Gemfile#L18) both use `~> 1.3.3` constraints. I had to bump and release snapshots because of a chicken and egg problem. `logstash-core/logstash-core.gemspec` had a dependency on clamp which conflicted with what fpm `1.3.3` brought in. Had to adjust that and republish gems. Can't wait for unified release on 5.0 :/


untergeek:
LGTM™

PR #5812, opened 2016-08-23
make sure sed replacement is posix compliant

jsvd:
the integration specs that target the update command sometimes use `sed` for setup actions before the actual specs are run.

One of these sed substitutions uses \s to match a whitespace, which in some systems might not work since it's not posix compliant,  [[:space]] should be used instead

aims to fix https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-acceptance/98/


suyograo:
LGTM

PR #5815, opened 2016-08-23
base SignalEvent class and cleanup LogStash namespace

colinsurprenant:
- introduces a `SignalEvent` base class and both `ShutdownEvent` and `FlushEvent` classes define the `flush?` and `shutdown?` method. This makes the checking for a signal event a bit faster and will help cleanup signal handling in the PQ branch.

- this previous modification revealed a namespacing problem: 
  - the `LogStash::LogStash::SHUTDOWN` and `LogStash::LogStash::FLUSH` constants were defined.
  - this problem was partially hidden by the fact that within the `Pipeline` class:
  ```
  ...
  end end
  ```
  All constants were prefixed by `Logstash::` which is both unnecessary and inefficient since the lookups would start by trying to lookup `LogStash::Foo` within the `LogStash` module (`LogStash::LogStash::Foo`) and not find it.

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -218,16 +218,16 @@ module LogStash; module Util
           if event.nil?
             # queue poll timed out
             next
-          elsif event == LogStash::SHUTDOWN
+          elsif event.is_a?(LogStash::SignalEvent)
             # We MUST break here. If a batch consumes two SHUTDOWN events
             # then another worker may have its SHUTDOWN 'stolen', thus blocking
             # the pipeline.
-            @shutdown_signal_received = true
-            break
-          elsif event == LogStash::FLUSH
+            @shutdown_signal_received = event.shutdown?
+
             # See comment above
             # We should stop doing work after flush as well.
-            @flush_signal_received = true
+            @flush_signal_received = event.flush?
+
             break
           else
             @originals[event] = true
--- a/logstash-core-event-java/lib/logstash/event.rb
+++ b/logstash-core-event-java/lib/logstash/event.rb
@@ -10,14 +10,26 @@ require "cabin"
 # in the future it might be necessary to refactor using like a BaseEvent
 # class to have a common interface for all pileline events to support
 # eventual queueing persistence for example, TBD.
-class LogStash::ShutdownEvent; end
-class LogStash::FlushEvent; end
-
 module LogStash
-  FLUSH = LogStash::FlushEvent.new
+  class SignalEvent
+    def flush?; raise "abstract method"; end;
+    def shutdown?; raise "abstract method"; end;
+  end
+
+  class ShutdownEvent < SignalEvent
+    def flush?; false; end;
+    def shutdown?; true; end;
+  end
+
+  class FlushEvent < SignalEvent
+    def flush?; true; end;
+    def shutdown?; false; end;
+  end
+
+  FLUSH = FlushEvent.new
 
   # LogStash::SHUTDOWN is used by plugins
-  SHUTDOWN = LogStash::ShutdownEvent.new
+  SHUTDOWN = ShutdownEvent.new
 end
 
 # for backward compatibility, require "logstash/event" is used a lots of places so let's bootstrap the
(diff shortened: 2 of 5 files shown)


logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -218,16 +218,16 @@ def take_originals_from_queue(queue, size, wait)
...
             next
-          elsif event == LogStash::SHUTDOWN
+          elsif event.is_a?(LogStash::SignalEvent)
             # We MUST break here. If a batch consumes two SHUTDOWN events
             # then another worker may have its SHUTDOWN 'stolen', thus blocking
             # the pipeline.
-            @shutdown_signal_received = true
-            break
-          elsif event == LogStash::FLUSH
+            @shutdown_signal_received = event.shutdown?
```
jsvd: I don't see the purpose of this refactor. the pr description talks about being faster + helps the code in the persistent queue branch.

IMO this could be cleaned as:

```
class SignalEvent;end
class ShutdownEvent < SignalEvent; end
class FlushEvent < SignalEvent; end
```

and then:

```ruby
          case event
          when nil
            # queue poll timed out
            next
          when LogStash::ShutdownEvent
            # We MUST break here. If a batch consumes two SHUTDOWN events
            # then another worker may have its SHUTDOWN 'stolen', thus blocking
            # the pipeline.
            @shutdown_signal_received = true
            break
          when LogStash::FlushEvent
            # See comment above
            # We should stop doing work after flush as well.
            @flush_signal_received = true
            break
          else
            @originals[event] = true
          end
```

Performance wise, in the case of a LogStash::Event the above implementation does .nil? and two === calls, while this PR does .nil?, is_a? and two method calls. The performance should be negligible, and eliminates some lines of code that deal with the flush? and shutdown?.

As for helping the PQ branch..no idea there.

jsvd:
+1 on the namespacing cleanup, but I have some doubts on the class refactoring

colinsurprenant:
The addition of the `flush?` and `shutdown?` methods are to avoid having stuff like 
```ruby
if shutdown_signaled?(signal)
   ...
end
if flush_signaled?(signal)
   ...
end
...

def shutdown_signaled?(signal)
  signal.is_a?(::LogStash::SHUTDOWN)
end

def flush_signaled?(signal)
  signal.is_a?(::LogStash::FLUSH) || shutdown_signaled?(signal)
end
```

This is somewhat an antipattern, the signal object should know about itself.  This is a lot of useless  and cumbersome code. This will help clean this in the `feature/java_persistence` branch.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
colinsurprenant: do not forget that that these ifs are called for all events. since signal are very **rare**, replacing the 2 ifs for both `LogStash::ShutdownEvent` and `LogStash::FlushEvent` for a single `if event.is_a?(LogStash::SignalEvent)` is more efficient and cleaner. Granted this is probably negligible and nonetheless, it is a useless if cascade.

jsvd:
signalling is checked by looking at `@shutdown_signal_received` and `@flush_signal_received`, no?
Also, in your example, `flush_signaled` has added logic to it, which could be implemented with Shutdown signal being a subclass of Flush.

Anyway, I'm not going to block this, but I do think the namespacing refactor is a nobrainer merge while the signalling refactor isn't ideal.

Up to you to change, merge, merge to another branch.

"trying not to block stuff on small accounts" LGTM

colinsurprenant:
The benefit for the signal classes refactor is not much apparent here in master but will help cleanup signal handling in the PQ branch. I don't think there is any risk in merging this and IMO this is a bit more clean too. Unless there are any serious concerns I'll go ahead and merge...

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
jsvd: btw, if this is a concern then we can just place `if event.is_a?(LogStash::Event)` as the first condition

colinsurprenant:
merged in master and 5.0

Issue #5818, opened 2016-08-24
Stats: change stats reset operation to only clear destroyed pipelined counts


jsvd:
Currently, when recreating a pipeline (during configuration reload) all stats are reset and recreated.

For stats like jvm, gc, global event count and others there's no need to reset them.

Therefore, when reloading a pipeline, only the stats from that pipeline should be delete from the metric collector.

This helps support https://github.com/elastic/logstash/issues/5680

acchen97:
Just to clarify, this **only** resets `/_node/stats/pipeline` stats, right? This is just the events/plugin pipeline stats, the remaining API resources will persist across pipeline reloads.

jsvd:
Correct!

jsvd:
resolved by #5819

PR #5819, opened 2016-08-24
partially reset counters during pipeline reload

jsvd:
fixes https://github.com/elastic/logstash/issues/5818

this PR makes it so that a pipeline restart only prunes metrics counts from the pipeline being deleted. This means that the global received/filtered/outputted event counts (and all others) will not be reset.


logstash-core/lib/logstash/agent.rb
```diff
@@ -162,9 +162,8 @@ def configure_metrics_collectors
     @periodic_pollers.start
   end
 
-  def reset_metrics_collectors
-    stop_collecting_metrics
-    configure_metrics_collectors
+  def reset_pipeline_metrics(id)
+    @collector.clear("stats/pipelines/#{id}")
   end
```
ph: love the api.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -180,6 +180,16 @@ def each(path = nil, &block)
     end
     alias_method :all, :each
 
+    def prune(path)
+      key_paths = path.gsub(/^#{KEY_PATH_SEPARATOR}+/, "").split(KEY_PATH_SEPARATOR)
```
ph: Maybe a good time to create a `#key_paths(path)` method

We have the same code here https://github.com/elastic/logstash/pull/5819/files#diff-ae63e631134ccf36b5e761ae91487f61R83

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -180,6 +180,16 @@ def each(path = nil, &block)
...
+      key_paths = key_paths.map {|k| k.to_sym }
```
ph: You can use the symbol to proc here.map(&:to_sym) this could be also part of the `#keys_path` method.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -180,6 +180,16 @@ def each(path = nil, &block)
...
+      @structured_lookup_mutex.synchronize do
+        keys_to_delete = @fast_lookup.keys.select {|namespace, _| (key_paths - namespace).empty? }
+        keys_to_delete.each {|k| @fast_lookup.delete(k) }
```
ph: Could probably do that in one iteration with a `delete_if?`

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -294,5 +304,14 @@ def fetch_or_store_namespace_recursively(map, namespaces_path, idx = 0)
...
     end
+
+    def delete_from_map(map, keys)
+      key = keys.first
+      if keys.size == 1
+        map.delete(key)
+      else
+        delete_from_map(map[key], keys[1..-1]) unless map[key].nil?
+      end
+    end
```
ph: <3

ph:
@jsvd Minor comments to address and can we also add unit tests on the **MetricStore** class concerning the new `#prune` method?

logstash-core/lib/logstash/instrument/metric_store.rb
jsvd: the problem is that `@fast_lookup` isn't a ruby Hash but a java ConcurrencyMap, which doesn't have a delete_if method

ph: true nevermind, this is was happen when we mixed java ruby and concurrency, we can't always have nice things.

jsvd:
I've added 2 unit tests to MetricStore#prune

logstash-core/spec/logstash/instrument/metric_store_spec.rb
```diff
@@ -221,4 +221,33 @@
...
+      # Lets add a few metrics in the store before trying to find them
+      metric_events.each do |namespaces, metric_key, action|
+        metric = subject.fetch_or_store(namespaces, metric_key, LogStash::Instrument::MetricType::Counter.new(namespaces, metric_key))
+        metric.execute(action)
+      end
+    end
+
+    it "should remove all keys with the same starting path as the argument" do
+      subject.prune("/node/sashimi/pipelines/pipeline01")
+      expect { subject.get(:node, :sashimi, :pipelines, :pipeline01) }.to raise_error LogStash::Instrument::MetricStore::MetricNotFound
```
ph: Can we do an expect before to make sure the metric exist? This might be defensive but I would still sleeps better if do the check ?

ph:
really small comment and it LGTM.

Issue #5810, opened 2016-08-22
Output Plugins Not Receiving Metrics


andrewvc:
In the aftermath of the `OutputDelegator` refactor output plugins are no longer having their `metric` instance set correctly, and are defaulting to the `NullMetric`. Additionally there are no tests around this.

We should take the opportunity, now that we are in 5.0, to change the signature of `Plugin#initialize(user_options={})` to be `Plugin#initialize(user_options={},meta_options={})` so that we can extend the options with things like metrics. Currently the code initializes the object then invokes the `metrics=` setter, which is a smell and anti-pattern. This is not a public API so it should not break compatibility.

PR #5825, opened 2016-08-24
Properly set Output Metrics in Output Delegator Strategy.

andrewvc:
Fixes https://github.com/elastic/logstash/issues/5810

Without this Outputs never have their `metric` property set. This doesn't perform the full refactor mentioned in #5810 , that will require more work than is merited at the moment.


untergeek:
LGTM™.  Waiting to see what the CI says.

suyograo:
LGTM

PR #5828, opened 2016-08-25
Update API version constraints in template

suyograo:


jsvd:
if we do this, then https://github.com/elastic/logstash/blob/2.4/lib/pluginmanager/generate.rb#L78 is no longer used. so I suggest either:

a) renaming this to plugin_api_dependency => '">= 1.60", "<= 2.99"' in generator.rb and in the templates
b) removing this line from generate.rb

I'd vote for a) to avoid having the same '">= 1.60", "<= 2.99"' string in multiple places

lib/pluginmanager/templates/input-plugin/logstash-input-example.gemspec.erb
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
   s.metadata = { "logstash_plugin" => "true", "logstash_group" => "input" }
 
   # Gem dependencies
-  s.add_runtime_dependency "logstash-core-plugin-api", "~> <%= min_version %>"
+  s.add_runtime_dependency "logstash-core-plugin-api", ">= 1.60", "<= 2.99"
   s.add_runtime_dependency 'logstash-codec-plain'
   s.add_runtime_dependency 'stud', '>= 0.0.22'
   s.add_development_dependency 'logstash-devutils', '>= 0.0.16'
```
wiibaa: @suyograo this line seems out-of-date compared to the others

lib/pluginmanager/templates/codec-plugin/logstash-codec-example.gemspec.erb
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
   s.metadata = { "logstash_plugin" => "true", "logstash_group" => "codec" }
 
   # Gem dependencies
-  s.add_runtime_dependency 'logstash-core-plugin-api', "~> <%= min_version %>"
+  s.add_runtime_dependency "logstash-core-plugin-api", ">= 1.60", "<= 2.99"
```
jsvd: missed this one :) s/">= 1.60", "<= 2.99"/plugin_api_dependency

suyograo: how'd I miss. ugh :/

jsvd:
LGTM

Issue #5604, opened 2016-07-07
[2.x] Exception thrown if shutdown is requested during part of startup.


jordansissel:
The error is this:

```
NoMethodError: undefined method `[]=' for nil:NilClass
```

This is caused when SIGTERM is sent between these lines of code: https://github.com/elastic/logstash/blob/v2.3.2/logstash-core/lib/logstash/agent.rb#L202-L213

This appears to be fixed in master (https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/agent.rb#L54)

jsvd:
closed by #5829

PR #5829, opened 2016-08-25
prevent exception if logstash is interrupted during startup

jsvd:
during startup signal trapping is set up to gracefully shutdown logstash
this signal handling relies on Stud.stop? which depends on @thread being set

Currently we set that variable after setting up the trapping, which
creates a window where an interruption will trigger Stud.stop and find
@thread to be nil.

This patch moves the setting of @thread to be done before trapping, this
fixing the issue.

fixes https://github.com/elastic/logstash/issues/5604


suyograo:
nice. LGTM.

PR #5838, opened 2016-08-26
fix. add vendor files to logstash-core gem

talevy:


ph:
LGTM, it should pickup all the jars from vendor/jars.

suyograo:
LGTM

PR #5843, opened 2016-08-30
fix name of the all plugin rpm and debian packages

jsvd:
in https://github.com/elastic/logstash/commit/4347cf8aecd54bf349da56c371de13fce34898de two rake tasks were added to generate the "all-plugins" rpm and debian packages.

However, currently the packages will be created with the names `logstash-2.4.0_all.deb` and `logstash-2.4.0.noarch.rpm`, instead of having "all-plugins" in the name.

This PR ensure the names of generated packages will be `logstash-all-plugins-2.4.0.noarch.rpm` and `logstash-all-plugins-2.4.0_all.deb`


talevy:
LGTM

PR #5844, opened 2016-08-31
Fix steps in the getting started

dedemorton:
In this PR, I've implemented a subset of the changes that I made for the docs in 5.0 (see https://github.com/elastic/logstash/commit/0f1ff0f428886a0a25bf90777e966b099470270d). There were some structural changes that I didn't want to carry over, so I didn't cherry-pick the changes.

Mainly, I wanted to fix some of the words/steps that were misleading/broken. I'm planning to make more improvements to the GS in 5.0.

### DIFF
--- a/docs/static/advanced-pipeline.asciidoc
+++ b/docs/static/advanced-pipeline.asciidoc
@@ -67,7 +67,8 @@ input {
 <1> The default behavior of the file input plugin is to monitor a file for new information, in a manner similar to the
 UNIX `tail -f` command. To change this default behavior and process the entire file, we need to specify the position
 where Logstash starts processing the file.
-<2> The default behavior of the file input plugin is to ignore files whose last modification is greater than 86400s. To change this default behavior and process the tutorial file (which is probably much older than a day), we need to configure Logstash so that it does not ignore old files.
+<2> Setting `ignore_older` to 0 disables file age checking so that the tutorial file is processed even though
+it's older than a day.
 
 Replace `/path/to/file` with the absolute path to the location of `logstash-tutorial.log` in your file system.
 
--- a/docs/static/advanced-pipeline.asciidoc
+++ b/docs/static/advanced-pipeline.asciidoc
@@ -1,5 +1,5 @@
 [[advanced-pipeline]]
-=== Setting Up an Advanced Logstash Pipeline
+=== Parsing Logs with Logstash
 
 A Logstash pipeline in most use cases has one or more input, filter, and output plugins. The scenarios in this section
 build Logstash configuration files to specify these plugins and discuss what each plugin is doing.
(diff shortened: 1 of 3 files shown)


docs/static/advanced-pipeline.asciidoc
```diff
@@ -67,9 +67,9 @@ input {
 <1> The default behavior of the file input plugin is to monitor a file for new information, in a manner similar to the
 UNIX `tail -f` command. To change this default behavior and process the entire file, we need to specify the position
 where Logstash starts processing the file.
-<2> The default behavior of the file input plugin is to ignore files whose last modification is greater than 86400s. To change this default behavior and process the tutorial file (which date can be much older than a day), we need to specify to not ignore old files.
+<2> The default behavior of the file input plugin is to ignore files whose last modification is greater than 86400s. To change this default behavior and process the tutorial file (which is probably much older than a day), we need to configure Logstash so that it does not ignore old files.
```
guyboertje: Actually, setting ignore_older to zero disables it entirely.

docs/static/advanced-pipeline.asciidoc
```diff
@@ -117,15 +118,8 @@ filter {
     }
 }
 --------------------------------------------------------------------------------
-And change output to see a result :
-[source,json]
---------------------------------------------------------------------------------
-output {
-  stdout { codec => rubydebug }
-}
---------------------------------------------------------------------------------
```
guyboertje: Did you mean to delete this section? Above, your addition targets the filter section but without this below the user will see nothing happening.
```
Edit the `first-pipeline.conf` file and replace the entire `output` section with the following text:

[source,json]
---------------------------------------------------------------------------------
output {
  stdout { codec => rubydebug }
}

The `codec => rubydebug` will output a formatted representation of each event to the terminal.
This allows you to view the results of any changes to you make to the config as you work though
this tutorial.
You may wish to switch back to this output config from time to time to visualize your changes.
```

guyboertje:
Two comments for optional changes - otherwise LGTM.

docs/static/advanced-pipeline.asciidoc
andrewvc: Looking at this again I do think the language is clear that the `0` option "does not ignore old files."

andrewvc: Perhaps for added clarity it could more directly say. "Setting ignore_older to 0 disables file age checking". Not at all a blocker for me though.

docs/static/advanced-pipeline.asciidoc
andrewvc: I think this is correct because it asks the user only to replace the filter section now as opposed to the entire file.

andrewvc:
LGTM

docs/static/advanced-pipeline.asciidoc
dedemorton: @guyboertje I do think that the rubydebug codec is useful, but the tutorial works without it. I removed it from the tutorial because it's not part of the final config file that we tell users to build. My guess is that someone meant to remove it entirely but missed a spot. Users don't know what a codec is yet, so introducing it without more explanation is potentially confusing.  I'm planning more improvements to the tutorial in 5.0, including a more iterative approach to building the parts of the config file so that users can see the results of their work without having to build everything first. When I make those changes, it might make sense to add rubydebug back into the mix. The way it was written before, though, it was confusing.

PR #5855, opened 2016-09-01
Update CLI topic to fix formatting and add missing options

dedemorton:
Overview of changes:
* Changed from example block formatting to defn list (formatting the content as an example block meant the text didn't wrap and users had to scroll to the right to see content).
* Reordered options to match `-h` output for 2.4.
* Fixed wording to better match output of `-h` (though the content in the doc does have slightly more info in places, which makes sense).
* Added one or two options that were missing previously.

The list of options here now matches what I see when I run LS with the `-h` flag. If anything is missing or needs to be added, we'll need to add it to the `-h` output, too.

### DIFF
--- a/docs/static/command-line-flags.asciidoc
+++ b/docs/static/command-line-flags.asciidoc
@@ -3,79 +3,138 @@
 
 Logstash has the following flags. You can use the `--help` flag to display this information.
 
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

PR #5856, opened 2016-09-01
Update CLI topic with missing options

dedemorton:
The list of options here now matches what I see when I run LS with the `-h` flag. If anything is missing or needs to be added, we'll need to add it to the `-h` output, too.  Please verify that the defaults are correct because I used what was listed in in the `-h` output. Not sure if I inadvertently used any calculated values.

Overview of changes:
* Reordered options to match `-h` output in the latest snapshot build in master.
* Fixed wording to better match output of `-h` (though the content in the doc does have slightly more info in places, which makes sense). I made the assumption that the `-h` output is more up-to-date than the docs, so please make sure that I haven't changed something in the doc that was actually correct.
* Added some options that were missing previously.


suyograo:
LGTM

PR #5862, opened 2016-09-05
Make the package install command mandatory to succeed

purbon:
Currently in our QA acceptance framework if the install command failed, it failed silently what is bad for feedback is something wrong happend.

With this changes the install command is mandatory to succeed, otherwise it will raise an exception and the test will brake.

One example failure, would be if we forget to build the current branch artifacts, see

```bash

Failures:

  1) artifacts operation behaves like installable is installed on ubuntu-1204
     Failure/Error: logstash.install({:version => LOGSTASH_VERSION})
     ServiceTester::InstallException:
       dpkg: error processing /logstash-build/logstash-5.0.0-alpha6-SNAPSHOT.deb (--install):
        cannot access archive: No such file or directory
       Errors were encountered while processing:
        /logstash-build/logstash-5.0.0-alpha6-SNAPSHOT.deb
     Shared Example Group: "installable" called from ./acceptance/spec/lib/artifact_operation_spec.rb:12
     # ./rspec/commands/debian.rb:32:in `install'
     # ./rspec/commands.rb:62:in `install'
     # ./acceptance/spec/shared_examples/installed.rb:8:in `(root)'

Finished in 0.638 seconds (files took 1.75 seconds to load)
3 examples, 1 failure, 2 pending
```
this error is explicit to report something wrong has happened when accessing the local snapshot file.

@untergeek this change will help the overall test experience with more clear error reporting, your thougts are very welcome.

### DIFF
--- a/qa/rspec/commands/base.rb
+++ b/qa/rspec/commands/base.rb
@@ -4,6 +4,8 @@ require_relative "system_helpers"
 
 module ServiceTester
 
+  class InstallException < Exception; end
+
   class Base
     LOCATION="/logstash-build".freeze
     LOGSTASH_PATH="/usr/share/logstash/".freeze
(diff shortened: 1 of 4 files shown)


suyograo:
LGTM

PR #5866, opened 2016-09-06
Add a breaking changes note concerning the private key format for the input beats.

ph:
The change of the private key format was a breaking change and should have been documented in the breaking change page.


suyograo:
LGTM

Issue #5822, opened 2016-08-24
Logstash will sometimes terminate with a lot of errors from Puma


jsvd:
During termination, logstash can produce a lot of errors related to Puma.
For the purpose of demonstration, I could only reproduce this in connection with a logstash error, but I'm not sure if it is a necessary requirement to trigger the puma messages:

```
~/projects/logstash (git)-[master] % bin/logstash -e "input { generator { count => 10000 } } output { if [value] > 10 { stdout {} } }"
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.3 - omit version 2.7.0
Pipeline main started
An unexpected error occurred! {:error=>#<NoMethodError: undefined method `>' for nil:NilClass>, :backtrace=>["(eval):24:in `output_func'", "/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/pipeline.rb:295:in `output_batch'", "org/jruby/RubyProc.java:281:in `call'", "/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb:171:in `each'", "org/jruby/RubyHash.java:1342:in `each'", "/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb:170:in `each'", "/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/pipeline.rb:289:in `output_batch'", "/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/pipeline.rb:252:in `worker_loop'", "/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/pipeline.rb:225:in `start_workers'"], :level=>:fatal}
Error in reactor loop escaped: Bad file descriptor - Bad file descriptor (Errno::EBADF)
org/jruby/RubyIO.java:3705:in `select'
/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/reactor.rb:29:in `run_internal'
/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/puma-2.16.0-java/lib/puma/reactor.rb:138:in `run_in_thread'
2016-08-24 15:35:34 +0100: Listen loop error: #<Errno::EBADF: Bad file descriptor - Bad file descriptor>
org/jruby/RubyIO.java:3705:in `select'
[...]

PR #5869, opened 2016-09-07
Silence PUMA errors when an exception occur in logstash core

ph:
When Logstash encounter an exception Puma was getting really noisy about
the file descriptor it used. Its because Puma has an infinite loop into
his reactor and when logstash is dying from the exception the FD get
removed under the puma threads.

Puma is using the constants `STDERR` and `STDOUT` to give information
about the errors, there is no easy way to provide Puma with a custom IO.
This PR hacks the Puma name to redefine both constants and send the
information to the logger using the `debug` level.

Also when we start the server we can pass a custom Puma::Events instance
to record some of the errors happening in the internal threads, this PR
also make sure we send all theses message to the configured logger.

Fixes: #5822

### DIFF
--- a/logstash-core/spec/logstash/webserver_spec.rb
+++ b/logstash-core/spec/logstash/webserver_spec.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 # require "logstash/json"
 require "logstash/webserver"
+require_relative "../support/helpers"
 require "socket"
 require "spec_helper"
 require "open-uri"
@@ -9,11 +10,15 @@ def block_ports(range)
   servers = []
 
   range.each do |port|
-    server = TCPServer.new("localhost", port)
-    Thread.new do
-      client = server.accept rescue nil
+    begin
+      server = TCPServer.new("localhost", port)
+      Thread.new do
+        client = server.accept rescue nil
+      end
+      servers << server
+    rescue => e
+      # The port can already be taken
     end
-    servers << server
   end
 
   sleep(1)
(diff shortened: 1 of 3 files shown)


ph:
@jsvd OK, this was tricky to test, because `IO.select` will make the server does into an infinite loop and we can get concurrency problems if we try to doubles. :(

I have also added a `silence_warnings` method that hack the **$VERBOSE** seting of ruby to hide the redefine constants warning in the tests..

logstash-core/spec/logstash/webserver_spec.rb
```diff
@@ -36,29 +41,56 @@ def free_ports(servers)
...
+                                    agent,
+                                    { :http_host => "localhost", :http_ports => port_range })}
+
+  let(:port_range) { 10000..10010 }
+
+  context "when an exception occur in the server thread" do
+    let(:spy_output) { spy("stderr").as_null_object }
+
+    it "should not log to STDERR" do
+      backup_stderr = STDERR
```
jsvd: maybe move this code to a before?

```ruby

let!(backup_stdout) { STDOUT }
let!(backup_stderr) { STDERR }

before(:each) do
   silence_warnings do
     STDOUT = STDERR = spy_output
   end
end


# ...

after(:each) do
 silence_warnings do
    STDERR = backup_stderr
    STDOUT = backup_stdout
  end
end
```

not sure if rspec allows this, actually

jsvd:
tests pass, tested. amazing job <3
left minor comment, otherwise LGTM

purbon:
@ph @jsvd, every time I see making error silent I worry a little bit, are we sure (100%) the right thing is to silent the errors? _playing devil's advocate here, to be sure we take the right direction to fix this issue_.

jsvd:
The error is still logged at the debug level. All known situations where this occurs is during logstash termination, so it should be ok.

This is indeed an "ugly" hack for an ugly design decision in Puma to directly call on STDERR from deep inside the code. We brainstormed for other options but couldn't find any, so open to suggestions :)

purbon:
I understand, I was just making sure we gave it enough thought, so I am good with it 👍, good work!

logstash-core/spec/logstash/webserver_spec.rb
ph: I did that previously and was hit by, the `variable was use in multiple example` bug :(

ph:
@purbon @jsvd I am not proud of that fix, but sadly I haven't found a better way to do it :(
Actually the fix was pretty simple, writing the unit test was a *pain*, because of the following.


- Mocking `IO.select` actually make the webserver thread into an infinite loop.
- using a double for logger in multiple threads is asking for the dreadly `this variable is shared between example` errors.
- Redefining the `SDTERR` or the `STDOUT` lead to the constant redefinition warning. :(

logstash-core/spec/logstash/webserver_spec.rb
ph: on a side not, I also think its prefereable to keep the code code, the test is also a hack.

PR #5871, opened 2016-09-07
New Settings API endpoint that only supports Logging

talevy:
All settings changes are implicitly transient.

# PUT API

```
PUT _node/settings
{"logger.logstash" : "error"}
```

with response:

```
{"host":"elasticbox","version":"5.0.0alpha6","http_address":"127.0.0.1:9600","acknowledged":true}
```

Or if there is something that goes wrong... you get an http status code of `400`

```
PUT _node/settings
{"logger.logstash" : "notalevel"}
```

with response:

```
{"host":"elasticbox","version":"5.0.0-alpha6","http_address":"127.0.0.1:9600","error":"invalid level[notalevel] for logger[logstash]"}
```

# Get API

```
curl -XGET "http://localhost:9600/_node/logging?pretty"
{
  "host" : "elasticbox-2.local",
  "version" : "6.0.0-alpha1",
  "http_address" : "127.0.0.1:9600",
  "loggers" : {
    "logstash.registry" : "WARN",
    "logstash.instrument.periodicpoller.os" : "WARN",
    "logstash.instrument.collector" : "WARN",
    "logstash.runner" : "WARN",
    "logstash.inputs.stdin" : "WARN",
    "logstash.outputs.stdout" : "WARN",
    "logstash.agent" : "WARN",
    "logstash.api.service" : "WARN",
    "logstash.instrument.periodicpoller.jvm" : "WARN",
    "logstash.pipeline" : "WARN",
    "logstash.codecs.line" : "WARN"
  }
```

### DIFF
--- a/logstash-core/lib/logstash/api/modules/logging.rb
+++ b/logstash-core/lib/logstash/api/modules/logging.rb
@@ -1,0 +1,52 @@
+# encoding: utf-8
+#
+java_import org.apache.logging.log4j.core.LoggerContext
+
+module LogStash
+  module Api
+    module Modules
+      class Logging < ::LogStash::Api::Modules::Base
+        # retrieve logging specific parameters from the provided settings
+        #
+        # return any unused configurations
+        def handle_logging(settings)
+          Hash[settings.map do |key, level|
+            if key.start_with?("logger.")
+              _, path = key.split("logger.")
+              LogStash::Logging::Logger::configure_logging(level, path)
+              nil
+            else
+              [key, level]
+            end
+          end]
+        end
+
+        put "/" do
+          begin
+            request.body.rewind
+            req_body = LogStash::Json.load(request.body.read)
+            remaining = handle_logging(req_body)
+            unless remaining.empty?
+              raise ArgumentError, I18n.t("logstash.web_api.logging.unrecognized_option", :option => remaining.keys.first)
+            end
+            respond_with({"acknowledged" => true})
+          rescue ArgumentError => e
+            status 400
+            respond_with({"error" => e.message})
+          end
+        end
+
+        get "/" do
+          context = LogStash::Logging::Logger::get_logging_context
+          if context.nil?
+            status 500
+            respond_with({"error" => "Logstash loggers were not initialized properly"})
+          else
+            loggers = context.getLoggers.map { |lgr| [lgr.getName, lgr.getLevel.name] }.sort
+            respond_with({"loggers" => Hash[loggers]})
+          end
+        end
+      end
+    end
+  end
+end
--- a/logstash-core/lib/logstash/logging/logger.rb
+++ b/logstash-core/lib/logstash/logging/logger.rb
@@ -6,6 +6,8 @@ module LogStash
       java_import org.apache.logging.log4j.Level
       java_import org.apache.logging.log4j.LogManager
       java_import org.apache.logging.log4j.core.config.Configurator
+      @@config_mutex = Mutex.new
+      @@logging_context = nil
 
       def initialize(name)
         @logger = LogManager.getLogger(name)
@@ -59,13 +61,22 @@ module LogStash
         @logger.trace(message, data)
       end
 
-      # Point logging at a specific path.
       def self.configure_logging(level, path = LogManager::ROOT_LOGGER_NAME)
-        Configurator.setLevel(path, Level.toLevel(level))
-      end # def configure_logging
+        @@config_mutex.synchronize { Configurator.setLevel(path, Level.valueOf(level)) }
+      rescue Exception => e
+        raise ArgumentError, "invalid level[#{level}] for logger[#{path}]"
+      end
 
       def self.initialize(config_location)
-        Configurator.initialize(nil, config_location)
+        @@config_mutex.synchronize do
+          if @@logging_context.nil?
+            @@logging_context = Configurator.initialize(nil, config_location)
+          end
+        end
+      end
+
+      def self.get_logging_context
+        return @@logging_context
       end
     end
   end
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,8 @@ qa/.vm_ssh_config
 qa/.vagrant
 qa/acceptance/.vagrant
 qa/Gemfile.lock
-
+*.ipr
+*.iws
+*.iml
+.gradle
+.idea
--- a/logstash-core/lib/logstash/api/rack_app.rb
+++ b/logstash-core/lib/logstash/api/rack_app.rb
@@ -5,6 +5,7 @@ require "logstash/api/modules/node"
 require "logstash/api/modules/node_stats"
 require "logstash/api/modules/plugins"
 require "logstash/api/modules/root"
+require "logstash/api/modules/logging"
 require "logstash/api/modules/stats"
 
 module LogStash
@@ -103,7 +104,8 @@ module LogStash
           "/_node" => LogStash::Api::Modules::Node,
           "/_stats" => LogStash::Api::Modules::Stats,
           "/_node/stats" => LogStash::Api::Modules::NodeStats,
-          "/_node/plugins" => LogStash::Api::Modules::Plugins
+          "/_node/plugins" => LogStash::Api::Modules::Plugins,
+          "/_node/logging" => LogStash::Api::Modules::Logging
         }
       end
     end
--- a/logstash-core/locales/en.yml
+++ b/logstash-core/locales/en.yml
@@ -83,6 +83,9 @@ en:
           Hot threads at %{time}, busiestThreads=%{top_count}:
         thread_title: |-
           %{percent_of_cpu_time} % of cpu usage, state: %{thread_state}, thread name: '%{thread_name}'
+      logging:
+        unrecognized_option: |-
+          unrecognized option [%{option}]
     runner:
       short-help: |-
         usage:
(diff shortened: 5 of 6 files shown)


logstash-core/lib/logstash/api/modules/settings.rb
```diff
@@ -0,0 +1,35 @@
...
+            else
+              LogStash::Logging::Logger::configure_logging(level, path)
+            end
+          end
+        end
+
+        put "/" do
+          # only logging setting for now
+          begin
+            req_body = JSON.parse(request.body.read)
```
talevy: need to add check for unprocessed settings options

logstash-core/lib/logstash/api/modules/settings.rb
```diff
@@ -0,0 +1,38 @@
+require 'json'
```
ph: logstash/json?

talevy: right, my mad

logstash-core/lib/logstash/api/modules/settings.rb
```diff
@@ -0,0 +1,38 @@
...
+# encoding: utf-8
+module LogStash
+  module Api
+    module Modules
+      class Settings < ::LogStash::Api::Modules::Base
+        def handle_logging(logging_keys, settings)
+          logging_keys.each do |key|
+            level = settings.delete(key)
+            path = key.split("logger.")[1]
+            if level.nil?
```
ph: could it be an empty value?

logstash-core/lib/logstash/api/modules/settings.rb
```diff
@@ -0,0 +1,38 @@
...
+            end
+          end
+        end
+
+        put "/" do
+          # only logging setting for now
+          begin
+            req_body = JSON.parse(request.body.read)
+            handle_logging(req_body.select { |k,_| k.start_with?("logger.") }.keys, req_body)
+            unless req_body.empty?
```
ph: One way to validate the provided is to use `hash#slice(...)`

ph: I think we should really enforce what we allow here.

ph:
@talevy +1 for going forward :)

logstash-core/lib/logstash/api/modules/settings.rb
talevy: right. there is a slight mixture of concerns here. `configure_logging` will throw a NPE anyways. I will remove this check and make the overall catch be more descriptive

talevy:
Where should I put the docs? cc @ph @suyograo @andrewvc 

I feel like the logging settings API should fall under a broader discussion of Logstash’s API. Currently, this is owned by Monitoring APIs in our docs. Since this endpoint doesn’t feel monitoring related… I feel weird placing it there. Although, it technically does inherit all the common options and server of the Monitoring API

logstash-core/lib/logstash/api/modules/settings.rb
```diff
@@ -0,0 +1,36 @@
+# encoding: utf-8
+module LogStash
+  module Api
+    module Modules
+      class Settings < ::LogStash::Api::Modules::Base
+        def handle_logging(logging_keys, settings)
+          logging_keys.each do |key|
+            level = settings.delete(key)
```
jordansissel: I am uneasy about modifying the `settings` object here (with delete). Can you tell me why you use delete, here?

logstash-core/lib/logstash/logging/logger.rb
```diff
@@ -6,6 +6,7 @@ class Logger
       java_import org.apache.logging.log4j.Level
       java_import org.apache.logging.log4j.LogManager
       java_import org.apache.logging.log4j.core.config.Configurator
+      @@config_mutex = Mutex.new
```
jordansissel: Are you using a class var to make this kind of global? Could we do without it? (I am OK if the answer is that we need it)

logstash-core/lib/logstash/api/modules/settings.rb
```diff
@@ -0,0 +1,36 @@
...
+        put "/" do
+          # only logging setting for now
+          begin
+            req_body = LogStash::Json.load(request.body.read)
+            handle_logging(req_body.select { |k,_| k.start_with?("logger.") }.keys, req_body)
+            unless req_body.empty?
+              raise ArgumentError, "unrecognized option [#{req_body.keys.first}]"
+            end
+            respond_with({"acknowledged" => true})
+          rescue Exception => e
```
jordansissel: I prefer StandardError, `Exception` is too high and will catch things like NoMemoryError and SystemExit.

logstash-core/lib/logstash/api/modules/settings.rb
```diff
@@ -0,0 +1,36 @@
...
+            status 400
```
jordansissel: this should be `status 500`

logstash-core/lib/logstash/api/modules/settings.rb
```diff
@@ -0,0 +1,36 @@
...
+          end
+        end
+
+        put "/" do
+          # only logging setting for now
+          begin
+            req_body = LogStash::Json.load(request.body.read)
+            handle_logging(req_body.select { |k,_| k.start_with?("logger.") }.keys, req_body)
+            unless req_body.empty?
+              raise ArgumentError, "unrecognized option [#{req_body.keys.first}]"
```
jordansissel: I wonder, can we do this inside `handle_logging` instead? It feels like an odd side effect to check for success by having `handle_logging` modify the given `req_body` and checking if the modifications leave the value empty afterwards.

logstash-core/lib/logstash/api/modules/logging.rb
```diff
@@ -0,0 +1,51 @@
...
+              nil
+            else
+              [key, level]
+            end
+          end]
+        end
+
+        put "/" do
+          begin
+            req_body = LogStash::Json.load(request.body.read)
```
ph: IIRC, since request.body.read it's an IO object and might have been read already by a before hook or rack handler it is recommended to rewind it before consuming it.

see http://www.sinatrarb.com/intro.html
```ruby
post "/api" do
  request.body.rewind  # in case someone already read it
  data = JSON.parse request.body.read
  "Hello #{data['name']}!"
end
```

logstash-core/lib/logstash/api/modules/logging.rb
```diff
@@ -0,0 +1,51 @@
...
+            remaining = handle_logging(req_body)
+            unless remaining.empty?
+              raise ArgumentError, "unrecognized option [#{remaining.keys.first}]"
```
ph: Can we move this to our **locales** file?

logstash-core/lib/logstash/api/modules/logging.rb
```diff
@@ -0,0 +1,51 @@
...
+            end
+            respond_with({"acknowledged" => true})
+          rescue Exception => e
+            status 400
+            respond_with({"error" => e.message})
+          end
```
ph: I do not  think we should use the exception for the control flow here, because this make all errors inherited from **Exception** return `400/bad_request` and this could consume other errors which should be mark as 500 (Application error).

ph: This makes me think that we should have a special type of exception that could be catcher higher in the code to return a more generic error.

logstash-core/lib/logstash/api/modules/logging.rb
```diff
@@ -0,0 +1,51 @@
...
+        end
+
+        get "/" do
+          context = LogStash::Logging::Logger::get_logging_context
```
ph: Is this case really possible to retrieve a nil context?

logstash-core/lib/logstash/logging/logger.rb
ph: The problem is log4j changed to the running configuration are not thread safe.

@talevy pointed me to the official log4j doc for this, see https://logging.apache.org/log4j/2.x/manual/customconfig.html


```
Applications sometimes have the need to customize logging separate from the actual configuration. Log4j allows this although it suffers from a few limitations:

If the configuration file is changed the configuration will be reloaded and the manual changes will be lost.
Modification to the running configuration requires that all the methods being called (addAppender and addLogger) be synchronized.
``

logstash-core/lib/logstash/api/modules/logging.rb
ph: It would be fine if we only catch `ArgumentError`? because the `.configure_logging` can also return this.

ph:
Did a review of this PR.

- The current branch is failling and the fails are related to this PR.
- I am OK with the current Unit test for this PR.
- Can we do another issue to create an integration test of changing the logger options so we can have it for 5.0 RC?

logstash-core/lib/logstash/api/modules/logging.rb
talevy: I'll catch Exception, and then introduce conditionals inside. problem is that a Java Exception is thrown, and `Error` does not capture it. I'll make this less catchy

logstash-core/lib/logstash/api/modules/logging.rb
talevy: Only if someone hits the api before `initialize` is called in the start of the runner. theoretically it wouldn't happen, but might as well be safe I think

logstash-core/lib/logstash/logging/logger.rb
talevy: thanks @ph, yeah, the above is the reasoning for this ^^^

logstash-core/lib/logstash/api/modules/logging.rb
ph: better safe than sorry

PR #5873, opened 2016-09-07
Fix formatting and edit breaking changes

dedemorton:


docs/static/breaking-changes.asciidoc
```diff
@@ -2,12 +2,13 @@
 == Breaking Changes
 
 **Changes in 2.4**
-=== Beats Output config change
+
+[float]
+== Beats Output Config Change
```
talevy: you're right, this should be "Input"

talevy:
LGTM

PR #5875, opened 2016-09-08
Fix formatting of breaking changes (for real this time)

dedemorton:
I wasn't happy when I looked more closely at the formatting of the 2.4 breaking changes doc, so I made a few more tweaks. 

I've added/clarified a couple of headers because it seemed like things were under headings where they didn't belong. Note that I removed "breaking changes" from the sub-headings because we establish the severity in the title for the page (removing the full title from sub-headings makes the doc slightly easier to scan).


suyograo:
LGTM

PR #5881, opened 2016-09-08
[5.x] Bump version to 5.1.0

suyograo:


Gemfile.jruby-1.9.lock
```diff
@@ -572,7 +572,7 @@ GEM
       snappy-jars (~> 1.1.0)
     snappy-jars (1.1.0.1.2-java)
     snmp (1.2.0)
-    spoon (0.0.4)
+    spoon (0.0.6)
```
jsvd: I suggest avoiding to update other stuff in this PR

suyograo: I can't avoid this because I don't hand edit the Gemfile.lock. It gets updated when I run `bundle install` after the version bumps.

jsvd:
LGTM

Issue #5733, opened 2016-08-05
 The Logstash Slow log


purbon:
###  Description of the problem

Systems like elasticsearch, mysql and other data processing engines have the concept of an slow log, a log file where the system is reporting slow operations. To have this in Logstash makes sense in order to increase the feedback users get.

### The proposed system should be able to:

* Be configurable. A user should be able to change a config variable to decide what means slow for a given component.
* Should be integrated with all plugins flavors in Logstash, this means it should be possible to report slow operations from inputs, codecs, filters and outputs.
* Should report also from core components, ex: pipeline being stalled. 
* Should be logged in a different file than the regular logs.
* As much context as possible should be provided, this means:
 * The component (plugin, pipeline, etc..), the action, duration, event (if applicable), context (configuration?).
* In the slow log there could be also context information useful to debug the problems.

### Solution proposed

> _Concrete details will be updated on demand as this idea gets developed, so take this proposal as an initial dump of ideas that will for sure evolve._

To implement this feature, specially if we just wanted to get slow_log mix with the current log, we could just reuse the logger instance, *however we aim for a more general structure*, this solution is discarded.

Who should be initially reporting the slow operation?, definitely should be the component detecting the situation, we’ll use the grok filter as example to describe the solution. For the grok filter it should be transparent to report an slow operation, to report an slow operation should be as simple as:

```report_slow_operation(action, duration, event, context, ….) ```

But internally in logstash this another situation, the log might need to get populated with collateral information such as current pipeline situation, workers and plugin status, etc This information should be gathered by other components inside the LS core logic, so this kind of mandates a system that is living in core and able to compose all necessary messages into a proper logger output.

The proposed solution will be:

* A component that it stays inside logstash-core.
* Is available to plugins, but also to core components.
* Knows how to receive a message from one of this components and act accordingly.
* Is able to gather contextual information when necessary.

More details could be grasp from an initial proof of concept:  
_[Include PoC link when available]_.
 
### Pros and Cons

* Easier to grab extended context from the logstash agent and pipelines. Providing easier to debug logs for end users.
* Easy interface with for the component consumers.
* Properly single responsibility, providing us with easier to test and maintain components.
* More complex to implementation than the straightforward solution.
* Need to scale properly to multiple producers of content (slow operations).

### Work packages

For the initial version of this component we aim to:

* Make the current file settings available for plugins.
* Build an slow log service able to receive slow log messages, gather context (when choosen to) and output the final log.
[...]

PR #5886, opened 2016-09-12
An slowlog for Logstash

purbon:
This PR is the materialization of the discussions around #5733 adding an slow log component for logstash, this include:
- The interface to time and communicate slow operation for the plugins, see [plugin.rb](https://github.com/elastic/logstash/compare/master...purbon:feature/slowlog?expand=1#diff-689e179dc5d0e4c2edb42dc602c1eebd) for details, also related tests.
- The threshold definition logic, to fetch the metadata from the settings file.
- Use an specific logger per component.
- Wire and communication to the web api component.
- An endpoint for the web api under `_node/slowlog` were a bunch if statistics are displayed about the slow log.

Please refer to the [issue](https://github.com/elastic/logstash/issues/5733#issuecomment-240093470) to gather more in detail context. 

There are however a few open questions I see around, happy to take your feedback:
- How to keep slow log messages under the specific slowlog appenders? not to being attached also to the root logger config, so data is not duplicated. @talevy ?
- An initial discussion about the web endpoint was [kicked off](https://github.com/elastic/logstash/issues/5733#issuecomment-245667778), however not follow up happened, happy to take your input here. At https://github.com/elastic/logstash/issues/5733#issuecomment-245924741 you can see how data looks like right now.
- Integration with https://github.com/elastic/logstash/pull/5871 when it gets merged, however it can be done later, so no blocker for me.


logstash-core/lib/logstash/plugin.rb
```diff
@@ -8,8 +8,26 @@
 require "logstash/plugins/registry"
 
 class LogStash::Plugin
+
+
+  class Timer
```
jordansissel: I wonder if we can make [TimedExecution](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/instrument/metric.rb#L84) and this `Timer` class use the same code? Low priority, since the code is very simple.

logstash-core/lib/logstash/plugin.rb
```diff
@@ -31,9 +49,21 @@ class LogStash::Plugin
...
+  # of reporting slow operations.
+  #
+  # ```
+  #  thresholds => {
+  #    filter.grok.runtime => 12
+  #  }
+  # ```
+  #
+  # The values defined here takes precedence over the ones defined in the logstash config file. (default => {})
+  config :thresholds, :validate => :hash, :default => {}
```
jordansissel: Where would a user learn what threshold names are available?

I also generally want us to avoid `:validate => :hash` if we can, for two reasons: First, that automated documentation has no way of helping users understand what is valid inside the hash. Second, that validation of the hash value is pretty complex and can be error prone -- the result is confused users. We see this user confusion with the `http_poller` input plugin, for example.

jordansissel: I don't know if all plugins will have "thresholds" -- and I want to learn from my mistake of making `codec` part of the base input and output classes (not all of those plugins can even use codecs).  Do all plugins need a thresholds config setting? What if a plugin doesn't implement any thresholds at all, I feel we shouldn't show this setting to the user if a plugin doesn't support it.

logstash-core/lib/logstash/plugin.rb
```diff
@@ -8,8 +8,26 @@
...
 class LogStash::Plugin
+
+
+  class Timer
+
+    def initialize
+      @timer = 0
+    end
+
+    def start
```
jordansissel: Time is pretty hard to test, so I wonder if we can pass a time as an argument:

```ruby
def start(time=Time.now)
  @timer = time
end
```

This will have the same behavior as the current method, and adds the benefit that we can fake time more easily in testing.

logstash-core/lib/logstash/plugin.rb
```diff
@@ -8,8 +8,26 @@
...
+      @timer = Time.now
+    end
+
+    def stop
```
jordansissel: Similar comment as above for using an argument with a default value:

```ruby
def stop(now=Time.now)
  @timer - now
end
```

logstash-core/lib/logstash/logging/slowlog_manager.rb
```diff
@@ -0,0 +1,50 @@
...
+          _report[:number_of_events] = item[1]
+          _report.merge!(stats.to_hash)
+          report[_key] = _report
+        end
+        items[key] = report
+      end
+      items
+    end
+
+    def top_items
```
jordansissel: I'm a little confused. There's a `top_items` method here, and a `TopItems` class that isn't used by this `top_items` method. Is the naming just confusing me? Or is `TopItems` class not used?

jordansissel:
@purbon can you provide a sample plugin (or a PR to an existing plugin) that makes use of these new thresholds/timings so that I can see how you envision plugin authors to record slowlog entries?

logstash-core/lib/logstash/logging/slow_logger/top_items.rb
```diff
@@ -0,0 +1,69 @@
...
+      @time = time
+    end
+
+    def <=>(other)
+      if threshold != other.threshold
+        return threshold <=> other.threshold
+      end
+      return -1 if time > other.time
+      return  0 if time == other.time
+      return  1 if time < other.time
```
guyboertje: Could you achieve the same result with `other.time <=> time`?

logstash-core/lib/logstash/logging/slow_logger/top_items.rb
```diff
@@ -0,0 +1,69 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "set"
+require 'thread'
+
+module LogStash; module Logging; module Util
+
+  class Operation
```
guyboertje: Did you intend for `Operation` to be an "inner" class of `TopItems`? If so it needs to be indented inside `TopItems`

logstash-core/lib/logstash/logging/slow_logger/top_items.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    attr_reader :threshold, :time
+
```
guyboertje: The tests seem to show `Operation.new(string, integer)`. Is the first argument meant to be a string or an integer?
If string, consider the name `:operation` instead.
If an integer, the name `:threshold` is OK.

logstash-core/spec/logstash/plugins/logging_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    subject.add("a.b.c", 15)
+    expect(subject.top_k(2).map { |o| o.threshold }).to eq(["a.b.c", "foo.bar.zet"])
+  end
+
+  it "should keep only N elements in the set" do
+    11.times do |i|
+      subject.add("#{i}.foo.bar.#{i}", rand(20))
+    end
+    expect(subject.size).to eq(10)
+  end
```
guyboertje: Consider expanding this test to verify the expected behaviour of the algorithm at the `SIZE` boundary.

When filled via `add("b", 100)` through `add("b", 109)` then `add("c", 200)` will have `"b", 109 -> 101` and `"c", 200` but `add("a", 50)` will displace `"c", 200`.

Its hard to understand why `"a", 50` is 'worth' more than `"c", 200`

logstash-core/lib/logstash/plugin.rb
purbon: Wise comments @jordansissel, the situation why I chose to implement this that way is because in #5733 we agreed to have thresholds like:

* In the settings file.
* Per plugin, taking precedence over the settings.

So I made the assumption all plugins should have this config option available, If I understand you properly, are you suggesting to have this option only defined in plugins that have and use the slow log?

re the usage of hash, my problem here is I see the thresholds as actually a hash, as you can see in the documentation. What are you thinking as alternate structure?

logstash-core/lib/logstash/logging/slowlog_manager.rb
purbon: The TopItems class is a leftover from early experimentation, I keep it just in case it was interesting during review, but I will remove it 😄  as it might get you confused.

logstash-core/lib/logstash/logging/slow_logger/top_items.rb
purbon: As discussed previously at https://github.com/elastic/logstash/pull/5886#discussion_r78588439 the top item got removed with 52ed6bdd785448c8d829d054caf84eb7d34f64c1. I removed it basically because it's not used anymore in benefit of having a more general structure.

purbon:
@jordansissel I will be pushing a more complete branch, but the idea to report slow operations with this change is to do something like:

```ruby
        timer.start
        if @hit_cache
          address = @hit_cache.getset(raw) { retriable_getaddress(raw) }
        else
          address = retriable_getaddress(raw)
        end
        slow_logger("filters.dns.getaddress", timer.stop)
```

this is an example from the logstash dns plugin.

andrewvc:
I generally like this PR, but I don't think we should be putting slowlog thresholds in the main plugin config. That sort of stuff should be in the `settings.yml`. In my mind the plugin config is for defining what the pipeline does, and should not include settings related to performance. While it is true that some plugins today do have some performance related settings included we should try to minimize these, not add to them.

purbon:
@andrewvc fairpoint, @talevy also pointed out a similar concern directly in the issue, I will c&p you comment there so we can centralize the talking there.

purbon:
Last commit 3e5bd60c1c37ed1801173e8dbd3e039749319a49 reverted the option to have thresholds inside plugins configurations, so for now all available thresholds will have to be as settings. This was discussed and agreed in the issue.

On the other side, it also remove the concerns that @jordansissel raised about using hash and the global config field.

So, what do you think? good to keep moving? @jordansissel @andrewvc

andrewvc:
@purbon glad to see that stuff removed. I haven't done a full review however. Do you need one or are other people doing it? I see @guyboertje and @jordansissel and @suyograo are already assigned.

purbon:
@andrewvc no problem at all, I am very good with them continuing the reviews here.

logstash-core/lib/logstash/logging/slow_logger.rb
```diff
@@ -0,0 +1,42 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/logging/slow_logger/freq_items"
+
+module LogStash
```
suyograo: Can we add comments please?

logstash-core/lib/logstash/logging/slow_logger.rb
```diff
@@ -0,0 +1,42 @@
...
+        @logger     = LogStash::Logging::Logger.new(name)
+        @freq_items = LogStash::Logging::Util::FreqItems.new
+      end
+
+      def log(threshold, time, data={})
+        message          = data.delete(:message)
+        data[:threshold] = threshold
+        data[:time]      = time
+
+        cache_elements(threshold, time)
```
suyograo: What does the `cache_elements` method do? Can you add comments?

logstash-core/lib/logstash/logging/slow_logger.rb
```diff
@@ -0,0 +1,42 @@
...
+    class SlowLogger
+
+      attr_reader :logger, :freq_items
+
+      def initialize(name="loggers.slow", params={})
+        @logger     = LogStash::Logging::Logger.new(name)
+        @freq_items = LogStash::Logging::Util::FreqItems.new
+      end
+
+      def log(threshold, time, data={})
```
suyograo: can we also take `Event` by default in this API? We can then log the full event if someone requests it. I would also add a new setting `slowlog.enable.event_logging`. `true` would log the entire original event. By default, it can be `false`.

logstash-core/lib/logstash/logging/slow_logger.rb
```diff
@@ -0,0 +1,42 @@
...
+        message          = data.delete(:message)
+        data[:threshold] = threshold
+        data[:time]      = time
+
+        cache_elements(threshold, time)
+        logger.warn(message, data)
```
suyograo: Why are we hardcoding to `#warn` here? Shouldn't this be dependent on the threshold setting set by the user? See behavior in ES for an example.

suyograo:
I would also appreciate a bunch of real log lines that would be output as slowlog. For example, in conjunction with the DNS PR. This will give the reviewer a taste of the output without actually running this PR.

suyograo:
Another point -

In your example PR, you use `slow_logger("filters.dns.getaddress", timer.stop)`. I think we should extract the `filters.dns.` prefix and have it available in general for all plugins (for the plugin developers). For example, outputs, depending on which output you are in, you'll have `outputs.elasticsearch` available to you as a prefix. You can then use this prefix + the method you are tracking in this api:

`slow_logger(create_prefix(".getaddress"), timer.stop)`

`create_prefix` will do something like `"#{slow_log_prefix}.getaddress"` only when threshold is activated so we don't do unecessary string joins.

logstash-core/lib/logstash/logging/slow_logger.rb
suyograo: `log(threshold, time, data={})`

is `time` millis, nanos? can you add that in the parameter. Maybe change it to `took_millis` or `took_nanos`?

logstash-core/lib/logstash/logging/slow_logger.rb
suyograo: I think this `logger.warn(message, data)` is not a great logging format. Can we have more structured logging here? Have explicit: 
`took_millis`
`plugin` (which plugin?)
`method`
`event_payload` (if enabled)
`optional_message`

When you use the `json` format to log, these should be nice fields for the user to use in ES (in the future)

suyograo: Maybe even have a separate SlowLogFormatter class to be able to edit in the future..

logstash-core/spec/logstash/plugins/slow_logger_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+  end
+
+  describe LogStash::Logging::SlowLogger do
+
+    subject { described_class.new }
+
+    it "should respond to log" do
+      expect(subject.respond_to?(:log)).to eq(true)
```
suyograo: I would like to see more tests to assert the actual slowlog message and for the different thresholds.

logstash-core/lib/logstash/logging/slow_logger.rb
purbon: Adding the event is totally possible now, but looks good to me to make it a requirement. At the beginning I thought not to force it, so left the data hash there to let users really add whatever they prefer. 

Not completely sure the best way would it be to output base on this setting, might be only output the _context_ when running in debug more? what do you think?

purbon:
@suyograo can you elaborate https://github.com/elastic/logstash/pull/5886#issuecomment-247865400 a little bit more? I want to understand your way of thinking before making conclusions about it. Thanks a lot!

logstash-core/lib/logstash/logging/slow_logger.rb
purbon: done, but not sure why PR did not refresh this comment.

logstash-core/lib/logstash/logging/slow_logger.rb
purbon: removed as not meaninfull anymore.

config/logstash.yml
```diff
@@ -113,3 +113,12 @@
...
 # path.plugins: []
+#
+# ------------ Slow log thresholds ---------
+#
+# An slow log threshold is a time limit for an operation that when is overcome it will
+# be reported at the related log file. Time is seconds.
+#
+# The last building block of the component is the log level (default: warn)
+#
+# filters.dns.getaddress.warn: 5
```
suyograo: I think this should be `filters.dns.threshold.warn: 5`. As a user how will I know there is `getaddress` method in the code? I want to log all the slow operations in DNS filter. the user shoudn't be required to know the sub-methods.

logstash-core/lib/logstash/logging/slow_logger/freq_items.rb
```diff
@@ -0,0 +1,77 @@
...
+        def initialize
+          @n = 0
+          @mean = 0.0
+          @m2 = 0.0
+          @min = -1
+          @max = 0
+        end
+
+        def update(x)
+          @n += 1
```
suyograo: Can you add comments here. Looking at the code, I don't know what you are updating and what the calculations are?

logstash-core/lib/logstash/logging/slow_logger.rb
```diff
@@ -0,0 +1,68 @@
...
+      @settings   = settings
+    end
+
+    def log(event, threshold, took_in_seconds, data={})
+      max_time = setting(threshold).to_i
+      return if max_time == 0 || took_in_seconds <= max_time
+
+      level, operation_namespace = split_fields(threshold)
+
+      data[:event] = event
```
suyograo: Can you add a setting to not log the event if the user chooses to?

logstash-core/lib/logstash/logging/slow_logger.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    attr_reader :logger, :freq_items, :settings
+
+    def initialize(name="loggers.slow", settings=LogStash::SETTINGS, params={})
+      @logger     = LogStash::Logging::Logger.new(name)
+      @freq_items = LogStash::Logging::Util::FreqItems.new
+      @settings   = settings
+    end
+
+    def log(event, threshold, took_in_seconds, data={})
```
suyograo: it would be nice to have another method `log(event, threshold, start_millis, end_millis, data={})` that plugins can use. This way, plugins don't have to do the `elapsed = end - start` calculation and we can keep it DRY.

suyograo:
@purbon How about some integration tests?

config/logstash.yml
purbon: I've written this as a kind of example, but it fits the format I think is appropriate, I would like to get in a namespace that makes the user able to locate which operation, direct or logical groups is invoked.

With `filters.dns.[operation_label].[log_level]` all the necessary info to get the context of the threshold is there, with your naming proposal the user can not know the context where this threshold is defined. Also what happen when there are two thresholds for a plugin?

logstash-core/lib/logstash/logging/slow_logger.rb
purbon: makes sense

purbon:
@suyograo you can see a set of plugin level unit test at https://github.com/elastic/logstash/pull/5886/files#diff-ffbd04ab14eb05984c997b0c2ea1cb1aR14 . So most of the bases are covered either from the slowlog or from the plugin level.

But I want to understand you mind, what kind of integration test are you thinking about? Are you envisioning a full pipeline run that reports slow logs? might be something with your new framework rats?

config/logstash.yml
suyograo: So how would a user know there is a `getaddress` call in DNS filter?

logstash-core/lib/logstash/logging/slow_logger.rb
```diff
@@ -0,0 +1,67 @@
...
+      max_time = setting(threshold).to_i
+      return if max_time == 0 || took_in_seconds <= max_time
+
+      level, operation_namespace = split_fields(threshold)
+      if setting("slowlog.plugins.context")
```
suyograo: Can we call this `slowlog.enable.source` or `slowlog.log.source` or `slowlog.log.event`? There are other context we are logging, which we don't control using this setting.

logstash-core/lib/logstash/environment.rb
```diff
@@ -40,6 +40,7 @@ module Environment
             Setting::String.new("http.host", "127.0.0.1"),
             Setting::PortRange.new("http.port", 9600..9700),
             Setting::String.new("http.environment", "production"),
+            Setting::Boolean.new("slowlog.plugins.context", true),
```
suyograo: Please add this in the default `logstash.yml` as well

suyograo:
> But I want to understand you mind, what kind of integration test are you thinking about? Are you envisioning a full pipeline run that reports slow logs? might be something with your new framework rats?

Yeah, this could wait for RATS

suyograo:
The only big concern I have is about the settings as described here -- https://github.com/elastic/logstash/pull/5886/files#r79742986

config/logstash.yml
purbon: fair point, I initially thought this problem could be addressed with documentation, but having a global switch makes also sense. We can always add this information as operational context.

logstash-core/lib/logstash/logging/slow_logger.rb
purbon: hope now is more clear.

talevy:
closed by https://github.com/elastic/logstash/pull/6128

PR #5887, opened 2016-09-12
Filebeat integration test with new rats

ph:
This PR uses the new framework developed by @suyograo to add integration test.
Few things in done in this pull request:

- Add support for multiple named config in the yaml files.
- Add support for ERB in the YAML files
- Make the ServiceLocator more flexible and use the file name instead of the predefined hash.
- Add a gitignore
- Add keys for the tests and a generator.
- Add test for theses scenario:
  - No TLS
  - TLS Server Auth
  - TLS Mutual Auth


logstash-core-event-java/lib/logstash-core-event-java_jars.rb
```diff
@@ -3,12 +3,12 @@
   require 'jar_dependencies'
 rescue LoadError
   require 'com/fasterxml/jackson/core/jackson-databind/2.7.3/jackson-databind-2.7.3.jar'
-  require 'com/fasterxml/jackson/core/jackson-annotations/2.7.3/jackson-annotations-2.7.3.jar'
+  require 'com/fasterxml/jackson/core/jackson-annotations/2.7.0/jackson-annotations-2.7.0.jar'
```
suyograo: Why this change?

logstash-core-event-java/lib/logstash-core-event-java_jars.rb
```diff
@@ -3,12 +3,12 @@
...
   require 'com/fasterxml/jackson/core/jackson-core/2.7.3/jackson-core-2.7.3.jar'
 end
 
 if defined? Jars
   require_jar( 'com.fasterxml.jackson.core', 'jackson-databind', '2.7.3' )
-  require_jar( 'com.fasterxml.jackson.core', 'jackson-annotations', '2.7.3' )
+  require_jar( 'com.fasterxml.jackson.core', 'jackson-annotations', '2.7.0' )
```
suyograo: Can we go back to 2.7.3?

suyograo:
We need to remove `qa/integration/fixtures/how.input` from this PR. You can add a script to download it or can use a subset of it. See https://github.com/elastic/logstash/blob/feature/ruby_integration_tests/qa/integration/fixtures/kafka_input.input. 

Maybe we can rename this file to `how_sample.input` and reuse them in multiple tests?

qa/integration/fixtures/beats_input_spec.yml
```diff
@@ -0,0 +1,33 @@
...
+  - filebeat
+  - logstash
+config:
+  without_tls: |-
+    input {
+      beats {
+        port => 5044
+      }
+    }
+  tls_server_auth: |-
```
suyograo: nice. Keyed config!

qa/integration/fixtures/beats_input_spec.yml
```diff
@@ -0,0 +1,33 @@
...
+    input {
+      beats {
+        ssl => true
+        port => 5044
+        ssl_certificate => '<%=options[:ssl_certificate]%>'
```
suyograo: Awesome `erb` addition

qa/integration/services/service_locator.rb
```diff
@@ -1,19 +1,30 @@
...
+      @services[name] = klass.new(settings)
+    end
   end
 
   def get_service(name)
-    @services[name]
+    @services.fetch(name)
+  end
+
+  def available_services
```
suyograo: Nice!

suyograo:
@ph just a couple of concerns and this is good to go. Thanks for adding the enhancements!

logstash-core-event-java/lib/logstash-core-event-java_jars.rb
ph: I think it got fetch when I ran the rake test:install-core I will revert it.

qa/integration/services/service_locator.rb
ph: Hash#fetch is awesome we should use it more.. with it its over looking all over the place when you have a nil value.

jsvd:
Can we somehow avoid including that how.input file in the repo? an uncompressed 64 mb file shouldn't be commited to git

qa/integration/services/filebeat_setup.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -ex
+current_dir="$(dirname "$0")"
+
+if [ -n "${FILEBEAT_VERSION}" ]; then
+  echo "Filebeat version is $FILEBEAT_VERSION"
+  version=$FILEBEAT_VERSION
+else
+   version=5.0.0-alpha5
```
jsvd: if possible I suggest picking this version from the versions.yml file, so we don't have to maintain this :)

ph: true, they should match now.

ph:
@suyograo I update this PR with the following:
- removed `how` data,
- renamed `kafka_input.input` to `how_sample.input`
- reverted the jackson changes.

I haven synchronized the versions.yml with the bash script, I think we can do it in the feature branch. WDYT?

suyograo:
LGTM

PR #5889, opened 2016-09-12
add new lock file for 5.0-beta1

suyograo:


ph:
We will miss the changes for the TCP output, since https://github.com/logstash-plugins/logstash-output-tcp/pull/20 isn't merged yet and it will be broken in this release?

ph:
LGTM, We might have still something depending on cabin somewhere. cc @talevy

suyograo:
@ph correct. I can create a lock file again tomorrow after TCP changes get in.

PR #5892, opened 2016-09-13
update lock file for devutils and tcp output

jsvd:
devutils has been released with changes for the log4j logging
tcp output has been changed to conform with the new concurrency strategy framework


suyograo:
LGTM

Issue #5874, opened 2016-09-07
API: Nearly any URL gives a successful response


jordansissel:
```
% curl http://192.168.1.124:9600/asfasdf/asdf/asdf/a/sdf
{"host":"localhost.localdomain","version":"5.0.0-alpha6","http_address":"0.0.0.0:9600"}
```

I expected this to fail somehow. It doesn't.

ph:
Yup this a bug, we might have a catch all url somewhere in the sinatra routes.

purbon:
Related to https://github.com/elastic/logstash/issues/5622 ?

ph:
I will close this in favor of #5622, since its the strategy I took to solve that problem.

Issue #5622, opened 2016-07-12
bug in the metrics api handing of missing resources (content return)


purbon:
![screen shot 2016-07-12 at 12 32 29](https://cloud.githubusercontent.com/assets/68540/16764175/fb8f393e-482c-11e6-8de2-3e1485778f62.png)

when hitting a missing resources in the api, the user gets a 404 (what is the right behaviour), but it also gets back a hash full of content, what is call metadata. The expected behaviour of a 404 code would be to give back 404 plus empty response back as nothing was found to properly answer the request.

is there any missing benefit of having always included this metadata in the requests that I might be missing? specially for error situations? 

related https://github.com/elastic/logstash/pull/5621

ph:
I think we should do the following:

- Return a 404 http code
- Return an tailored json when error happen.
- Explicitly show `status; 404` in the json

This give us more flexibility in the future and also align with other elastic product like ES.

By flexibility I meant human and computer can understand what happen.


As a reference this is the ES output.
```json
{  
   "error":{  
      "root_cause":[  
         {  
            "type":"index_not_found_exception",
            "reason":"no such index",
            "resource.type":"index_or_alias",
            "resource.id":"dddd",
            "index_uuid":"_na_",
            "index":"dddd"
         }
      ],
      "type":"index_not_found_exception",
      "reason":"no such index",
      "resource.type":"index_or_alias",
      "resource.id":"dddd",
      "index_uuid":"_na_",
      "index":"dddd"
   },
   "status":404
}
```

purbon:
@ph like your idea is a good middle approach.

ph:
cc @jordansissel

ph:
fixed in #5622

PR #5897, opened 2016-09-13
Return 404 on non existant resources

ph:
This PR introduce a changes to make sure the API endpoint correctly
return 404 in every case. To make it work it uses an exception that get
pickup up in the base class by the `error` handler.

Using the exception template allow the code to be dry and make sure all
other errors can be returned with the same format.

Example of a 404.
```json
{
"error": { "message": "Not Found"},
"path": "/this-should-not-exist",
"status": 404
}
```

The code will also set the content-type and the right error code in the
header.

Fixes: #5874, #5622

### DIFF
--- a/logstash-core/lib/logstash/api/app_helpers.rb
+++ b/logstash-core/lib/logstash/api/app_helpers.rb
@@ -1,27 +1,43 @@
 # encoding: utf-8
 require "logstash/json"
+require "logstash/api/errors"
 
 module LogStash::Api::AppHelpers
-
+  # This method handle both of the normal flow *happy path*
+  # and the display or errors, if more custom logic is added here
+  # it will make sense to separate them.
+  #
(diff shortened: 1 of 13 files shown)


suyograo:
LGTM

```
suyog@machine:~$ curl localhost:9600/_node/gibberish
{"path":"/_node/gibberish","status":404,"error":{"message":"Not Found"}}suyog@machine:~$
suyog@machine:~$
suyog@machine:~$ curl -I localhost:9600/_node/gibberish
HTTP/1.1 404 Not Found
Content-Type: application/json
X-Content-Type-Options: nosniff
Content-Length: 72
```

Tested on an existing endpoint:

```
suyog@machine:~$ curl localhost:9600/_node
{"host":"Suyogs-MBP-2","version":"6.0.0-alpha1"
```

and get back 200

LGTM 🚢

PR #5898, opened 2016-09-13
Set default log.level to INFO

### LINKED ISSUE #5735: Set default log level to info
Elasticsearch does this, so we must as well.

### LINKED ISSUE #5893: Print "Pipeline main started" message at startup
during the logging changes, this was switched to be logged with log-level `WARN`. We should bring it back to the user in the terminal

talevy:
sets default to INFO and updates some verbose logging to
more appropriate, less verbose log levels where it makes sense.

Closes #5735.
Closes #5893.

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -101,12 +101,12 @@
 # Options for log.level:
 #   * fatal
 #   * error
-#   * warn (default)
-#   * info
+#   * warn
+#   * info (default)
 #   * debug
 #   * trace
 #
-# log.level: warn
+# log.level: info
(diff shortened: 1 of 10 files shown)


jsvd:
there's a reference in the docs about the default log level: https://github.com/elastic/logstash/blob/master/docs/static/command-line-flags.asciidoc line 75

jsvd:
also https://github.com/talevy/logstash/blob/bdb90a13755a1ed7f12c6eda43e4f5a696c62566/config/logstash.yml#L105

suyograo:
LGTM

PR #5900, opened 2016-09-13
[fix-bug] respect log.format and path.logs from cli

### LINKED ISSUE #5895: --log-in-json property needs to be enabled again
LS 2.4 introduced the log-in-json cli param, the new logging framework made this difficult to update. this needs to be fixed

talevy:
also change how `log.level` is configured at startup time to be set
as a property directly in the log4j2.properties file

Closes #5895

### DIFF
--- a/config/log4j2.properties
+++ b/config/log4j2.properties
@@ -2,26 +2,39 @@ status = error
 name = LogstashPropertiesConfig
 
 appender.console.type = Console
-appender.console.name = console
+appender.console.name = plain_console
 appender.console.layout.type = PatternLayout
 appender.console.layout.pattern = [%d{ISO8601}][%-5p][%-25c] %m%n
-# JSON Logging
-# appender.console.layout.type = JSONLayout
-# appender.console.layout.compact = true
-# appender.console.layout.eventEol = true
+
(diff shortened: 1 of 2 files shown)


config/log4j2.properties
```diff
@@ -2,26 +2,39 @@ status = error
...
-rootLogger.level = error
-rootLogger.appenderRef.console.ref = console
-rootLogger.appenderRef.rolling.ref = rolling
+appender.json_rolling.type = RollingFile
+appender.json_rolling.name = json_rolling
+appender.json_rolling.fileName = ${sys:ls.logs}/logstash.log
+appender.json_rolling.filePattern = ${sys:ls.logs}/logstash-%d{yyyy-MM-dd}.log
+appender.json_rolling.policies.type = Policies
+appender.json_rolling.policies.time.type = TimeBasedTriggeringPolicy
+appender.json_rolling.policies.time.interval = 1
```
ph: lets just confirm what is the rotation frequency, should be day?

ph:
LGTM

config/log4j2.properties
suyograo: +1 on 1 day.

PR #5901, opened 2016-09-13
fix tests and bring back configure logging in runner

talevy:
meant to override any custom log4j2.properties file that is introduced


jordansissel:
LGTM.

One curiosity -- what if the user doesn't specify the log.level in settings or flag? Do we still override the log level?

talevy:
@jordansissel yeah, always adhere to the logstash's default.

is there a way to know whether a user explicitly set a setting in the file or in clamp?

PR #5902, opened 2016-09-13
Fixed links and formatting

dedemorton:
Clint made some changes to the static files in logstash-docs to correct some links. This PR gets those changes into the logstash repo.

### DIFF
--- a/docs/static/breaking-changes.asciidoc
+++ b/docs/static/breaking-changes.asciidoc
@@ -1,44 +1,56 @@
 [[breaking-changes]]
 == Breaking changes
 
-**Breaking changes in 5.0**
+This section discusses the changes that you need to be aware of when migrating your application to Logstash {version}.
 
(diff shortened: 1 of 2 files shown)


suyograo:
LGTM

dedemorton:
@suyograo I've pushed another commit because I noticed that there were a bunch of changes missing from master. It maybe doesn't matter since the files in 5.0 are correct, but on the off chance that someone looks at this content in master, I think it's good to have the latest/greatest. Cherry-picking was a little messy, so I just copied the contents from 5.0. Let me know if that's not Ok, and I can track down all the commits and cherry-pick.

dedemorton:
This should be merged into master only. To avoid merge weirdness, I've created a separate PR that we can merge into 5.0/5.x.

suyograo:
@dedemorton sounds good!

PR #5903, opened 2016-09-13
add package specific log4j2 logging props

### LINKED ISSUE #5894: Logstash should not log to Console when running as a service
Logstash is currently logging to both Console and File at all times. It should not log to console when run as as service

suyograo:
Fixes #5894

### DIFF
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -309,10 +309,6 @@ namespace "artifact" do
 
     basedir = File.join(File.dirname(__FILE__), "..")
 
-    File.join(basedir, "pkg", "log4j2.properties").tap do |path|
-      dir.input("#{path}=/etc/logstash")
-    end
-
     # Create an empty /var/log/logstash/ directory in the package
     # This is a bit obtuse, I suppose, but it is necessary until
     # we find a better way to do this with fpm.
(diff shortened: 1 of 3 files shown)


rakelib/artifacts.rake
```diff
@@ -309,9 +309,9 @@ namespace "artifact" do
 
     basedir = File.join(File.dirname(__FILE__), "..")
 
-    File.join(basedir, "pkg", "logrotate.conf").tap do |path|
-      dir.input("#{path}=/etc/logrotate.d/logstash")
-    end
+    File.join(basedir, "pkg", "log4j2.conf").tap do |path|
```
jordansissel: this should be log4j2.properties, right?

suyograo:
```
ubuntu@ip-10-0-2-201:~$ ls -lrt /etc/logstash/
total 24
-rw-rw-r-- 1 root root 1659 Aug 22 11:41 startup.options
-rw-rw-r-- 1 root root 1738 Aug 22 11:41 jvm.options
-rw-rw-r-- 1 root root 2886 Aug 22 14:53 logstash.yml.dpkg-old
-rw-r--r-- 1 root root 1249 Sep 13 21:09 log4j2.properties
-rw-rw-r-- 1 root root 2868 Sep 13 22:05 logstash.yml
drwxrwxr-x 2 root root 4096 Sep 13 22:05 conf.d
ubuntu@ip-10-0-2-201:~$ vi /etc/logstash/log4j2.properties
```

talevy:
LGTM

PR #5904, opened 2016-09-13
remove commented value for path.logs in default settings file

talevy:
the build scripts rely on the value not being present to update with
appropriate value for the package


talevy:
as done here: https://github.com/elastic/logstash/blob/master/pkg/debian/after-install.sh#L10

suyograo:
LGTM

PR #5910, opened 2016-09-14
Fix heading levels in logging topic

dedemorton:
Minor change to fix heading levels (already pushed this to logstash-docs).


suyograo:
LGTM

Issue #5912, opened 2016-09-14
Concurrency Error in the Webserver patches for the Logger STDERR.


ph:
https://logstash-ci.elastic.co/job/elastic+logstash+5.0+multijob-os-compatibility/143/os=amazon/console

````
ERROR: Configuration reloading also requires passing a configuration path with '-f yourlogstash.conf'
INFO: Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in "--path.settings=/path/.." in the command line options
ERROR: Failed to load settings file from "path.settings". Aborting... path.setting=/tmp/a/a/a/a, exception=Errno::ENOENT, message=>No such file or directory - /tmp/a/a/a/a/logstash.yml
.................Validating dummy
............

Failures:

  1) LogStash::WebServer when an exception occur in the server thread should not log to STDERR
     Failure/Error: Unable to find matching line from backtrace
     ConcurrencyError:
       Detected invalid array contents due to unsynchronized modifications with concurrent users
     # /var/lib/jenkins/workspace/elastic+logstash+5.0+multijob-os-compatibility/os/amazon/vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # /var/lib/jenkins/workspace/elastic+logstash+5.0+multijob-os-compatibility/os/amazon/rakelib/test.rake:42:in `(root)'
     # /var/lib/jenkins/workspace/elastic+logstash+5.0+multijob-os-compatibility/os/amazon/vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/task.rb:248:in `execute'
     # /var/lib/jenkins/workspace/elastic+logstash+5.0+multijob-os-compatibility/os/amazon/vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/task.rb:243:in `execute'
     # /var/lib/jenkins/workspace/elastic+logstash+5.0+multijob-os-compatibility/os/amazon/vendor/bundle/jruby/1.9/gems/rake-11.2.2/lib/rake/task.rb:187:in `invoke_with_call_chain'

ph:
dang, its related to the hack we have made to make sure we can silence puma correctly.

ph:
[...]

PR #5918, opened 2016-09-14
Fix a concurrency issue with the Webserver when overrideing Puma STDERR and STDOUT.

ph:
This PR changes how we modify the STDOUT/STDERR in puma, instande of
using `const_set`, we override the constants using a module.
This operation should be thread safe and will make sure the STDERR are
correctly visible.

After when we receive the logger we can swap the null logger with the
real thing in a thread safe way.

Fixes: #5912

### DIFF
--- a/logstash-core/lib/logstash/patches/puma.rb
+++ b/logstash-core/lib/logstash/patches/puma.rb
@@ -1,0 +1,44 @@
+# encoding: utf-8
+#
+# Patch to replace the usage of STDERR and STDOUT
+# see: https://github.com/elastic/logstash/issues/5912
+module LogStash
+  class NullLogger
+    def self.debug(message)
+    end
+  end
+
+  # Puma uses by default the STDERR an the STDOUT for all his error
+  # handling, the server class accept custom a events object that can accept custom io object,
+  # so I just wrap the logger into an IO like object.
+  class IOWrappedLogger
+    def initialize(new_logger)
+      @logger_lock = Mutex.new
+      @logger = new_logger
+    end
+
+    def sync=(v)
+      # noop
+    end
+
+    def logger=(logger)
+      @logger_lock.synchronize { @logger = logger }
+    end
+
+    def puts(str)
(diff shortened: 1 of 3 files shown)


logstash-core/lib/logstash/webserver.rb
```diff
@@ -3,9 +3,40 @@
 require "puma"
 require "puma/server"
 require "concurrent"
+require "thread"
 
 module LogStash
   class WebServer
+    class NullLogger
```
jsvd: since both of these classes are meant as work arounds for behaviour of Puma, have you thought about namespacing them inside the Puma module, instead of Logstash?

logstash-core/lib/logstash/webserver.rb
```diff
@@ -110,3 +122,11 @@ def start_webserver(port)
     end
   end
 end
+
+# Reopen the puma class to create a scoped STDERR and STDOUT
+# This operation is thread safe since its done at the class level
+# and force JRUBY to flush his classes cache.
+module Puma
```
jsvd: can we perhaps do this in https://github.com/elastic/logstash/tree/master/logstash-core/lib/logstash/patches ?

ph: I am +1 on moving them to patches, I would prefer to keep the `NullLogger` in our namespace thought. Its currently only used by puma but its not puma specific.

ph:
@jsvd update with your reviews comments.

logstash-core/lib/logstash/patches/puma.rb
```diff
@@ -0,0 +1,42 @@
+# encoding: utf-8
+
```
jsvd: please add the issue number as a comment here in the code

jsvd:
small comment and LGTM

logstash-core/lib/logstash/patches/puma.rb
ph: added will merge in 5.0 5.X.

PR #5925, opened 2016-09-15
Update Logstash 5 to the latest input beats 3.1.4

ph:
The B1 internal releases contained an old version of the beats input which didn't
work well with the private keys, this PR make sure we use the latest version 3.1.4 in the 5.0 and 5.X branches

Fixes: logstash-plugins/logstash-input-beats#129


suyograo:
LGTM

PR #5928, opened 2016-09-15
QA Integration dont ship with certificates

ph:
We are removing the pre generated certificate from the repository, the
certificates needed for the test will be generated before downloading
filebeat.


suyograo:
LGTM

PR #5929, opened 2016-09-15
revert default logging to `warn`

talevy:
this also introduces a DIAG level that is between ERROR and WARN so that
the "Pipeline started" message is still exposed

Note: The only reason this PR exists is due to the verbosity of many plugins mentioned here: https://github.com/elastic/logstash/issues/5930.

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -101,12 +101,12 @@
 # Options for log.level:
 #   * fatal
 #   * error
-#   * warn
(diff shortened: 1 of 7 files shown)


talevy:
Closing due to the alternate strategy of modifying the liberal usage of `logger.info` in the plugins here: https://github.com/elastic/logstash/issues/5930.

PR #5936, opened 2016-09-16
Amend redhat based package install acceptance tests

### LINKED ISSUE #5935: Broken redhat based acceptance tests
The error checks added a few days ago made the redhat based test raise an error were the multiple installation of an ```rpm``` package will break with an error like:

```bash
Failures:

  1) artifacts operation behaves like installable is installed on centos-6
     Failure/Error: logstash.install({:version => LOGSTASH_VERSION})
     ServiceTester::InstallException:
       Failed to set locale, defaulting to C
       Error: Nothing to do
     Shared Example Group: "installable" called from ./acceptance/spec/lib/artifact_operation_spec.rb:12
     # ./rspec/commands/redhat.rb:33:in `install'
     # ./rspec/commands.rb:62:in `install'
     # ./acceptance/spec/shared_examples/installed.rb:9:in `(root)
```

purbon:
The artifact install acceptance test was installing every time a new logstash artifact, however and since the introduction of error checking for the installation, this has been broken because installing multiple times a package will return always exit code !=0.

This PR makes sure to fix this by:

* make redhat test uninstall a package after test are done.
* make sure install errors are raised only when an error code is != 0, not only when something is written at stderr.

This is a reason why acceptance test are read at [logstash-ci](https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-acceptance/label=metal,suite=redhat/125/console).

Fixes #5935

### DIFF
--- a/qa/rspec/commands/redhat.rb
+++ b/qa/rspec/commands/redhat.rb
@@ -38,7 +38,6 @@ module ServiceTester
       hosts = (host.nil? ? servers : Array(host))
       at(hosts, {in: :serial}) do |_|
         sudo_exec!("yum remove -y #{package}")
-        sudo_exec!("rm -rf /etc/logstash")
       end
     end
 
--- a/qa/rspec/commands/redhat.rb
+++ b/qa/rspec/commands/redhat.rb
@@ -23,17 +23,22 @@ module ServiceTester
     def install(package, host=nil)
       hosts  = (host.nil? ? servers : Array(host))
       errors = []
+      exit_status = 0
       at(hosts, {in: :serial}) do |_host|
         cmd = sudo_exec!("yum install -y  #{package}")
+        exit_status += cmd.exit_status
         errors << cmd.stderr unless cmd.stderr.empty?
       end
-      raise InstallException.new(errors.join("\n")) unless errors.empty?
+      if exit_status > 0 
+        raise InstallException.new(errors.join("\n"))
+      end
     end
 
     def uninstall(package, host=nil)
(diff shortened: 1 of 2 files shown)


qa/rspec/commands/redhat.rb
```diff
@@ -23,11 +23,15 @@ def package_for(filename, base=ServiceTester::Base::LOCATION)
...
       hosts  = (host.nil? ? servers : Array(host))
       errors = []
+      exit_status = 0
       at(hosts, {in: :serial}) do |_host|
         cmd = sudo_exec!("yum install -y  #{package}")
+        exit_status += cmd.exit_status
         errors << cmd.stderr unless cmd.stderr.empty?
       end
-      raise InstallException.new(errors.join("\n")) unless errors.empty?
+      if exit_status > 0
```
jsvd: I see your intent in doing this. I guess the only downside is that anything that is written to stderr by commands that terminate with status == 0 won't be logged/registered/shown anywhere.

I'd perhaps make the errors a list of `{"host" => host, "command" => cmd, "exit_status" => 0, "stderr" => cmd.stderr }` and raise an InstallException if either an exit_status isn't 0 or stderr isn't empty? not sure...thoughts?

purbon: What happened in this exact case is that exit status was 0, but something got written to stderr, so the exception was raised, basically it was a ```Do Nothing``` kind of situation. So for your proposal this would have to raise an exception, what i don't thinks is right. 

However I agree we should add more logging and we can do that certainly in feature changes, works for you?

Issue #5781, opened 2016-08-18
Log4j1.x and SLF4j compatibility jars in classpath need to be included


talevy:
work in https://github.com/elastic/logstash/issues/5650 leads to a mismatch between the popular slf4j and log4j1.x logging libraries. It should be the responsibility of the plugin to include adapters if it knows it needs them. Logstash could potentially include them at runtime for all, but this feels wasteful since they will only be used against a select set of plugins.

re discussion: https://github.com/elastic/logstash/pull/5651#discussion_r75302743

fbaligand:
Here's my opinion about it :
- in logstash, logging system and its configuration is common and shared with everybody : core code and plugins (contrary to an applications server for example)
- manage this shared logging system and its integration is logstash's responsibility, and not each plugin's responsibility.
- if it's not done, some plugin logs will be lost, some others will arrive anywhere.
- in particular for slf4j, it's important that slf4j "api" and "impl" versions are aligned. And logstash-core is the best place to ensure that.

=> That's why I think it is relevant to add these JARs to logtash-core :
- slf4j-api
- log4j-slf4j-impl
- log4j-1.2-api
- and optionally (if some plugins/libraries use Java Util Logging framework) : log4j-jul

@ph @suyograo @jordansissel 
Any opinion on that subject ?

jordansissel:
My feeling is that we should onlyship log4j2 because we need it. It feels
like we can't solve the problem for other libraries. For example, what if a
plugin or dependency needs a conflicting version of a logging library? Or
what if they shade an incompatible logging jar within the dependency?

I don't have enough recent experience with Java build/releng practices to
know what dangers lie in wait for us, but I do have too much uncertainty
around this proposal to agree with it yet -- can you help me understand?

ph:
I would believe core should take care of any logging library needed, but we have another problem. There is no mechanism in place to day in logstash to correctly deal with dependency of the jars. (its a bit of a far west) Plugins or core can ship with duplicates version of the jars, like the jackson library.

> For example, what if a plugin or dependency needs a conflicting version of a logging library? Or
what if they shade an incompatible logging jar within the dependency?


I am not too worried about this currently for the following reason:

- We control most of the plugin
- Good library are good citizens when using logging, like netty or kafka.

fbaligand:
Concerning logging from third party libraries/frameworks :
[...]

PR #5941, opened 2016-09-17
Log4j1.x and SLF4j compatibility jars in classpath need to be included (#5781)

fbaligand:
add log4j2 implementations for slf4j and jcl in dependencies (`build.gradle`), to fix issue #5781

### DIFF
--- a/logstash-core/build.gradle
+++ b/logstash-core/build.gradle
@@ -29,14 +29,19 @@ repositories {
 }
 
 dependencies {
-  runtime 'org.apache.logging.log4j:log4j-1.2-api:2.6.2'
-  compile 'org.apache.logging.log4j:log4j-api:2.6.2'
-  compile 'org.apache.logging.log4j:log4j-core:2.6.2'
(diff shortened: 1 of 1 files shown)


talevy:
hi @fbaligand thanks for these changes!  Since we are a little late for 5.0, we may wait it out to see if slf4j and jcl should be added into 5.1.

For now, the log4j-1.2 compatibility API was included, so plugins like Kafka log under Logstash's loggers.

fbaligand:
I'm surprised you find it's too late given that Logstash 5 is at beta1 stage.

Considering kafka and beats, the present state is not very sane. I mean :
- they both depend on log4j 1.2 and slf4j bridge to log4j 1.2
- that also mean it adds it to classpath

That would be cleaner if they depend on slf4j bridge to log4j 2, and also safer.
slf4j bridge to log4j 1.2 depend on 2 log4j1.2 classes that are not present in log4j-1.2-api.
It can lead to surprising behaviours.

suyograo:
I am not sure this is applicable anymore. Let me know if you think otherwise.

Issue #5952, opened 2016-09-20
Pipeline duration_in_millis metric seems inaccurate


acchen97:
In the metrics, the pipeline level`duration_in_millis` seems to be inaccurately high.

Config used: https://github.com/elastic/logstash-performance-testing/blob/master/lib/lsperfm/defaults/config/complex_syslog.conf

```
{
  "host" : "xxx",
  "version" : "5.0.0-beta1",
  "http_address" : "127.0.0.1:9600",
  "jvm" : {
    "threads" : {
      "count" : 23,
      "peak_count" : 25
    },
    "mem" : {
      "heap_used_in_bytes" : 257892288,
      "heap_used_percent" : 12,
      "heap_committed_in_bytes" : 519045120,
      "heap_max_in_bytes" : 2077753344,
      "non_heap_used_in_bytes" : 177067680,
      "non_heap_committed_in_bytes" : 187940864,
      "pools" : {
        "survivor" : {
          "peak_used_in_bytes" : 8912896,
          "used_in_bytes" : 9154112,
          "peak_max_in_bytes" : 34865152,
          "max_in_bytes" : 69730304,
          "committed_in_bytes" : 17825792
        },
        "old" : {
          "peak_used_in_bytes" : 111477288,
          "used_in_bytes" : 161886952,
          "peak_max_in_bytes" : 724828160,
          "max_in_bytes" : 1449656320,
          "committed_in_bytes" : 357957632
        },
        "young" : {
          "peak_used_in_bytes" : 71630848,
          "used_in_bytes" : 86851224,
          "peak_max_in_bytes" : 279183360,
          "max_in_bytes" : 558366720,
          "committed_in_bytes" : 143261696
        }
      }
    },
    "gc" : {
      "collectors" : {
        "old" : {
          "collection_time_in_millis" : 163,
          "collection_count" : 2
        },
        "young" : {
          "collection_time_in_millis" : 1092,
          "collection_count" : 108
        }
      }
    }
  },
  "process" : {
    "open_file_descriptors" : 57,
    "peak_open_file_descriptors" : 61,
    "max_file_descriptors" : 10240,
    "mem" : {
      "total_virtual_in_bytes" : 5195812864
    },
    "cpu" : {
      "total_in_millis" : 229566984000,
      "percent" : 0
    }
  },
  "pipeline" : {
    "events" : {
      "duration_in_millis" : 83225682,
      "in" : 130,
      "filtered" : 130,
      "out" : 130
    },
[...]

PR #5953, opened 2016-09-20
Correctly calculate the time spend in the filter + output block

ph:
This PR fix an issue where the time was calculated but no work was done
on the event. This code make sure we have at least one event to start
recording the time spend.

This was causing the `events/duratin_in_millis` to not be in sync with
the time spend on the filtera, since `take_batch` was called in a tight
loop and could return an empty array this made the duration was way off.

Fixes: #5952

### DIFF
--- a/logstash-core/lib/logstash/instrument/metric_store.rb
+++ b/logstash-core/lib/logstash/instrument/metric_store.rb
@@ -188,6 +188,10 @@ module LogStash module Instrument
       end
     end
 
+    def size
+      @fast_lookup.size
+    end
+
     private
     def get_all
       @fast_lookup.values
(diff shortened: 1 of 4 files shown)


ph:
let me do a quick sanity check before reviewing this.

ph:
Ready for review!

If you use this configuration

```
input {
generator {}
}
filter {
ruby {
code => "sleep(1)"
}
}
output {
stdout {
codec => json_lines
}
}
```

This gives a consistent result:

```
events: {
duration_in_millis: 1004192,
in: 1500,
filtered: 1000,
out: 1000
},
plugins: {
inputs: [ ],
filters: [
{
id: "ruby_8d8b1db4f73080c8274bc529e41af19fa92dfd7a-2",
events: {
duration_in_millis: 1010838,
in: 1012,
out: 1008
},
name: "ruby"
}
],
```

suyograo:
LGTM

Issue #5638, opened 2016-07-15
[meta] first persistence implementation release tracking


colinsurprenant:
This issue list and track progress toward the first release of the persistence integration into logstash.
## Related issues
- Persistence design issue #5636 
- DLQ issue #5283
### Pre-GA
- [ ] sweep code TODOs and convert to tasks/issues and prioritize
- [ ] Add bootstrap checks to persistent queue https://github.com/elastic/logstash/issues/6552 (@original-brownbear)
- [ ] performance & benchmarks (@colinsurprenant)

### Post-GA
- [ ] batched queue and multi write integration (@colinsurprenant)
- [ ] blog post (@colinsurprenant)
- [ ] code sweep for proper exception handling (@original-brownbear)
- [ ] more queue recovery tests
- [ ] more *pages tests
- [ ] refactor acking bitset to use ES LocalCheckpointService (@original-brownbear)
- [x] refactor *Settings classes into a builder pattern #7053, also see https://github.com/elastic/logstash/pull/5958#discussion_r83122350 (@original-brownbear)
- [ ] finalize streaming FilePageIO
- [ ] compare and benchmark MmapPageIO vs streaming FilePageIO
- [ ] Check Available Disk Space for Persisted Queues  https://github.com/elastic/logstash/pull/6998 (@original-brownbear)

### Completed
- [x] optimistic recovery PR #6519 (@colinsurprenant) 
- [x] prevent PQ corruption if two logstash instances are started from the same config or location #6505 (@colinsurprenant)
- [x] Pipeline initialization vs Persistent Queue initialization #5916 (@colinsurprenant)
- [x] enable file-based tests/specs (@guyboertje)
- [x] pass existing integration tests (using the current benchmarks setup)
- [x] DLQ/eviction implementation (see #5283) - we agreed to **not** change the current event dropping behaviour for the first iteration
- [x] fix queue.open bug with empty pages
- [x] refactor Event serialization to use JSON/CBOR (@guyboertje)
- [x] add checksum support to MmapPageIO (@guyboertje)
- [x] optimize Queue.ack() page inefficient page traversal (@colinsurprenant)
- [x] optimize Queue.firstUnreadPage()  inefficient page traversal (@colinsurprenant)
- [x] clean hardcoded "checkpoint" filenames (@colinsurprenant)
- [x] handle case of single element bigger than page capacity in Queue.write()
- [x] add proper Queue.close() with Page.close() and PageIO.close() and review queue closing sequence and exception handling.
[...]

Issue #5283, opened 2016-05-11
Events eviction/DLQ


colinsurprenant:
In the context of upcoming internal queuing persistence, we will introduce pipeline-wide reliability to make sure that queued events will not be removed until acknowledged as fully processed by the final pipeline stage. This will ensure at-least-once delivery semantic in particular in the case of an unsupervised stop or a crash. 

In that context, events flowing in the pipeline will require to **always** be acknowledged in the normal runtime. The only situation where events are not acknowledged is upon an unsupervised stop or a crash. This requirement to always acknowledge events introduces the need to evict events upon processing errors or catched/uncatched exceptions.

These should be the possible outcomes of event processing:

```
  1. the process is successful, the event/batch is ACK'ed
  2. the process is blocked, ACK cannot happens until the process is unblocked
  3. the process crashes with an uncatched exception, the event/batch is ACK'ed and evicted (DLQ)
  4. the process catches an exception:
    4.1. if there's a retry mechanism, the event/batch is retried
      4.1.1. the retry is successful, the event/batch is ACK'ed 
      4.1.2. the retry fails (max retry?), the event/batch is ACK'ed and evicted (DLQ)
    4.2. there's no retry mechanism, the event/batch is ACK'ed and evicted (DLQ)
```

We need to introduce a proper/holistic logstash-wide events eviction mechanism.

jordansissel:
+1 on the proposal. In terms of implementation, if ack-and-evict isn't atomic, we'll want to evict to the DLQ and ACK *after* the DLQ write succeeds to avoid data loss on machine failures.

acchen97:
> 3. the process crashes with an uncatched exception, the event/batch is ACK'ed and evicted (DLQ)

The ack/evict option makes sense for pipeline crashes, but should we also entertain the scenario of continued processing (from non-empty queue) post restart?  It would need to restart with the same pre-crash config, but seems like a notable use case to consider.

I agree this queue eviction stuff can and should seemingly converge with other error routing cases you noted here, sharing similar "DLQ" type storage infra.  This may be a completely separate discussion, but re: destinations for eviction/error routing, there's a few options and we could potentially architect with an easy workflow for reprocessing this data back through LS.  I feel the initial focus could be just writing to local file or ES (or simply discard if desired), thoughts?  Btw, ingest node did implement some [failure processing](https://www.elastic.co/guide/en/elasticsearch/reference/master/handling-failure-in-pipelines.html) to route error events into another ES index.

[...]

PR #5882, opened 2016-09-08
check versions.yml for correct project version

talevy:
- and fix a test

PR #5958, opened 2016-09-21
Persistent Queue v1

colinsurprenant:
This is a WIP level PR to start getting eyeballs and bootstrapping the review process. Please note the following:
- tasks/issues tracking for the first iteration of PQ is in #5638. There are still important details to look into so try and avoid reporting already tracked issues/tasks in this PR
- by looking at the files changed by filing the PR I can already see a few files that should not be part of this PR, I will look into this shortly

First Iteration Design Goals:
- **prevent inflight data loss** upon app/os/machine crash by aiming at _at-least-once_ delivery semantic. note that at this point there is no change in the per-event processing/delivery failure where best-effort is done at the plugin level and can result in dropped event. This will be addressed in the future Dead Letter Queue feature, see #5283
- **provide backpressure handling** within logstash using this variable length persistent queue. This will help avoid installing external intermediate message queues for the sole purpose of handling logstash backpressure.

Usage:
- by default PQ is enabled and in the `data/queue` path. make sure it exists: `$ mkdir -p data/queue`
- `$ rm -f data/queue/*` will clean the queue if needed when logstash is not running.
- The following new config options are supported:
  - `path.queue: which defaults to`data/queue`
  - `queue.type:` which default to the disk based PQ but can also take the `memory` option to use the in-memory implementation of the PQ (that we use in tests), or `synchronous` which is the legacy in-memory queue. 
  - `queue.page_capacity` which defaults to 250M
  - `queue.queue_capacity` which **for now** is a maximum number of events limit in the queue, by default there is no limit


jordansissel:
I am working through reviewing this, but not currently commenting in github because it takes ... a while ... for github to render this PR in my browser. Details soon :)

logstash-core/lib/logstash/environment.rb
```diff
@@ -46,7 +46,11 @@ module Environment
 
   ].each {|setting| SETTINGS.register(setting) }
 
-  SETTINGS.register Setting::WritableDirectory.new("path.queue", ::File.join(SETTINGS.get("path.data"), "queue"))
+  queue_file_path = ::File.join(SETTINGS.get("path.data"), "queue")
+  # Here we have to create the directory even though it might be overridden by settings.yml
+  # We should consider a Setting::LazyDirectory class that creates the directory on first 'get'
+  FileUtils.mkdir_p([queue_file_path])
```
suyograo: Can we add debug logs that we are creating this file?

colinsurprenant: I am wondering if this is the right place to create the dir if it does not exists? see https://github.com/elastic/logstash/blob/263e973d090a28446b64a81fa812e1109620f790/rakelib/test.rake#L10-L13  where I did this tmp patch to make sure the dir also exists for the tests ... /cc @guyboertje

logstash-core-event-java/build.gradle
```diff
@@ -1,18 +1,13 @@
...
-        mavenLocal()
-        mavenCentral()
-        jcenter()
-    }
-    dependencies {
-        classpath 'net.saliman:gradle-cobertura-plugin:2.2.8'
-    }
-}
+group = 'org.logstash'
+version = '6.0.0-alpha1'
```
suyograo: Can we get this version info from `versions.yml`? See https://github.com/elastic/logstash/pull/5882

logstash-core/lib/logstash/environment.rb
```diff
@@ -40,8 +40,18 @@ module Environment
             Setting::String.new("http.host", "127.0.0.1"),
             Setting::PortRange.new("http.port", 9600..9700),
             Setting::String.new("http.environment", "production"),
+            Setting::String.new("queue.type", "persisted", true, ["persisted", "memory", "synchronous"]),
```
suyograo: Should we expose this to users?

logstash-core/lib/logstash/environment.rb
```diff
@@ -40,8 +40,18 @@ module Environment
...
+            Setting::Numeric.new("queue.page_capacity", 250 * 1024 * 1024),
+            Setting::Numeric.new("queue.queue_capacity", 250 * 1024 * 1024)
```
suyograo: we should take string sizes like ES does (10g/10mb) to make it easier for the users

logstash-core/lib/logstash/environment.rb
suyograo: Not sure what synchronous means here. Is it a blocking queue?

logstash-core/lib/logstash/environment.rb
```diff
@@ -40,8 +40,18 @@ module Environment
...
+
   ].each {|setting| SETTINGS.register(setting) }
 
+  queue_file_path = ::File.join(SETTINGS.get("path.data"), "queue")
+  # Here we have to create the directory even though it might be overridden by settings.yml
+  # We should consider a Setting::LazyDirectory class that creates the directory on first 'get'
+  FileUtils.mkdir_p([queue_file_path])
+  SETTINGS.register Setting::WritableDirectory.new("path.queue", queue_file_path)
```
jordansissel: I'm going to patch WritableDirectory to try to create the directory on validation. Will push to this branch.

logstash-core/lib/logstash/environment.rb
jordansissel: Let's add a new setting "Bytes" for this.

logstash-core/lib/logstash/environment.rb
jordansissel: ac4bb61 implements directory creation on validation (which isn't the right place, it should happen on `get` (my next change)

logstash-core-event-java/build.gradle
colinsurprenant: oh yes +1

logstash-core/lib/logstash/environment.rb
colinsurprenant: synchronous is the current queue. I think we should be able to configure it if is faster and users don't care about reliability  -  and that's an good point, we should think about the one-off use-case, where users just run LS on a single or a set of static files... maybe introduce a cmdline switch for this use-case?

logstash-core/lib/logstash/environment.rb
colinsurprenant: and here we should remove the queue.queue_capacity setting because currently we only support a maximum number of events capacity and capping the queue at `250 * 1024 * 1024` events does not make sense.

colinsurprenant: oh and +1 on using human readable byte-size values

jordansissel: eadf16b adds a Bytes setting and makes the queue capacity settings use it.

logstash-core/lib/logstash/environment.rb
jordansissel: @colinsurprenant I've added directory creation to the `value` method of WritableDirectory. I think this was the right place (also the comment in the code here noted that the getter was the right place to do mkdir also)

logstash-core/lib/logstash/environment.rb
colinsurprenant: I pushed 70e102e to set the  `queue.queue_capacity` to Numeric 0 since for now this "capacity" is the maximum number of events and not the queue size.

logstash-core-event-java/build.gradle
colinsurprenant: waiting on #5882 to be merged to leverage it

logstash-core/lib/logstash/environment.rb
colinsurprenant: let's push this issue up, technically anything is possible but in terms of usability we have to decide, then we'll adjust here.

colinsurprenant: added as a P1 in #5638

logstash-core/lib/logstash/environment.rb
colinsurprenant: I did this temporary hack [in the rake test task](https://github.com/elastic/logstash/blob/7f66d8bd176d01991135d330935ce59bd1ae0469/rakelib/test.rake#L10-L13) - I guess we should use the same mecanism there ... but is SETTINGS exposed at this point? seems tricky

Gemfile
```diff
@@ -113,3 +114,4 @@ gem "logstash-output-stdout"
 gem "logstash-output-tcp"
 gem "logstash-output-udp"
 gem "logstash-output-webhdfs"
+gem "logstash-filter-multiline"
```
talevy: why was this added to the gemfile?

logstash-core-event-java/lib/logstash/event.rb
```diff
@@ -25,10 +25,14 @@ def flush?; true; end;
...
-  FLUSH = FlushEvent.new
+  class NoSignal < SignalEvent
+    def flush?; false; end;
+    def shutdown?; false; end;
+  end
 
-  # LogStash::SHUTDOWN is used by plugins
+  FLUSH = FlushEvent.new
   SHUTDOWN = ShutdownEvent.new
+  NO_SIGNAL = NoSignal.new
```
talevy: how is this being used? I only see it being checked for initialization

logstash-core-event-java/src/main/java/org/logstash/Timestamp.java
```diff
@@ -3,14 +3,14 @@
...
-import org.joda.time.LocalDateTime;
 import org.joda.time.Duration;
+import org.joda.time.LocalDateTime;
 import org.joda.time.format.DateTimeFormatter;
 import org.joda.time.format.ISODateTimeFormat;
 
 import java.util.Date;
 
-@JsonSerialize(using = TimestampSerializer.class)
+@JsonSerialize(using = org.logstash.json.TimestampSerializer.class)
```
talevy: why not import this class?

logstash-core-queue-jruby/build.gradle
```diff
@@ -0,0 +1,111 @@
+group 'org.logstash'
+version = '6.0.0-alpha1'
```
talevy: version should be handled programmatically via `version.yml` as previously mentioned

logstash-core-queue-jruby/build.gradle
```diff
@@ -0,0 +1,111 @@
...
+apply plugin: 'idea'
+
+repositories {
+    mavenCentral()
+}
+
+gradle.projectsEvaluated {
+    tasks.withType(JavaCompile) {
+        options.compilerArgs << "-Xlint:deprecation"
+//        options.compilerArgs << "-Xlint:unchecked" << "-Xlint:deprecation"
```
talevy: can be removed?

logstash-core-queue-jruby/build.gradle
```diff
@@ -0,0 +1,111 @@
...
+idea {
+    module {
+        scopes.PROVIDED.plus += [project.configurations.provided]
+    }
+}
+
+dependencies {
+    testCompile group: 'junit', name: 'junit', version: '4.12'
+    provided group: 'org.jruby', name: 'jruby-core', version: '1.7.25'
+    provided files('../logstash-core-event-java/lib/logstash-core-event-java/logstash-core-event-java.jar')
```
talevy: we should be depending on this in gradle. let me know if you feel like this is out of the scope of the initial merge-ability review.

logstash-core-queue-jruby/gemspec_jars.rb
```diff
@@ -0,0 +1,4 @@
+# This file is generated by Gradle as part of the build process. It extracts the build.gradle
```
talevy: can we just not check this file into git?

logstash-core-queue-jruby/lib/logstash-core-queue-jruby/logstash-core-queue-jruby.rb
```diff
@@ -0,0 +1,24 @@
+# encoding: utf-8
+
+require "java"
+
+module LogStash
+end
+
+# local dev setup
+classes_dir = File.expand_path("../../../build/classes/main", __FILE__)
```
talevy: note: after merging, we should probably clean this use-case up in our code

logstash-core-queue-jruby/lib/logstash-core-queue-jruby/version.rb
```diff
@@ -0,0 +1,3 @@
+# encoding: utf-8
+
+LOGSTASH_CORE_QUEUE_JRUBY_VERSION = "0.0.1"
```
talevy: version.yml, no?

logstash-core-queue-jruby/settings.gradle
```diff
@@ -0,0 +1,3 @@
+rootProject.name = 'logstash-core-queue-jruby'
+
```
talevy: empty newlines

logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedBatchExtLibrary.java
```diff
@@ -0,0 +1,88 @@
...
+                return new RubyAckedBatch(runtime, rubyClass);
+            }
+        }, module);
+
+        clazz.defineAnnotatedMethods(RubyAckedBatch.class);
+    }
+
+    @JRubyClass(name = "AckedBatch", parent = "Object")
+    public static class RubyAckedBatch extends RubyObject {
+        Batch batch;
```
talevy: should be private and final, no?

logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueExtLibrary.java
```diff
@@ -0,0 +1,140 @@
...
+                return new RubyAckedQueue(runtime, rubyClass);
+            }
+        }, module);
+
+        clazz.defineAnnotatedMethods(RubyAckedQueue.class);
+    }
+
+    // TODO:
+    // as a simplified first prototyping implementation, the Settings class is not exposed and the queue elements
+    // are assumed to be logstash Event.
```
talevy: is the queue expected to have elements other than Logstash Event in the future?

logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueMemoryExtLibrary.java
```diff
@@ -0,0 +1,141 @@
...
+        {
+            Batch b;
+
+            try {
+                b = this.queue.readBatch(RubyFixnum.num2int(limit), RubyFixnum.num2int(timeout));
+            } catch (IOException e) {
+                throw context.runtime.newIOErrorFromException(e);
+            }
+
+            // TODO: return proper Batch object
```
talevy: can you elaborate? note gives the impression that this is a big deal

logstash-core/gemspec_jars.rb
```diff
@@ -0,0 +1,9 @@
+# This file is generated by Gradle as part of the build process. It extracts the build.gradle
```
talevy: do we want to have this checked-in?

logstash-core/lib/logstash/environment.rb
```diff
@@ -23,7 +23,7 @@ module Environment
            Setting::Boolean.new("config.test_and_exit", false),
            Setting::Boolean.new("config.reload.automatic", false),
            Setting::Numeric.new("config.reload.interval", 3), # in seconds
-           Setting::Boolean.new("metric.collect", true) {|v| v == true }, # metric collection cannot be disabled
+           Setting::Boolean.new("metric.collect", true), #{|v| v == true }, # metric collection cannot be disabled
```
talevy: why the change?

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -1,6 +1,6 @@
 # encoding: utf-8
 require "concurrent"
-require "logstash/event"
+# require "logstash/event"
```
talevy: erase?

logstash-core/lib/logstash/instrument/snapshot.rb
```diff
@@ -1,6 +1,6 @@
 # encoding: utf-8
 require "logstash/util/loggable"
-require "logstash/event"
+# require "logstash/event"
```
talevy: erase?

logstash-core-event-java/lib/logstash/event.rb
talevy: nevermind, I see how it is being used now

logstash-core/settings.gradle
```diff
@@ -0,0 +1,2 @@
+rootProject.name = 'logstash-core'
+
```
talevy: empty newline

logstash-core-event-java/build.gradle
colinsurprenant: done

Gemfile
colinsurprenant: I believe @guy added for some tests?

logstash-core-queue-jruby/build.gradle
colinsurprenant: +1

logstash-core-queue-jruby/build.gradle
colinsurprenant: I think we should keep that for the upcoming gradle refactor?

logstash-core-queue-jruby/lib/logstash-core-queue-jruby/version.rb
colinsurprenant: yes

logstash-core/lib/logstash/environment.rb
colinsurprenant: no idea actually.

logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
```diff
@@ -0,0 +1,44 @@
...
+        this.closed = new AtomicBoolean(false);
+    }
+
+    // close acks the batch ackable events
+    public void close() throws IOException {
+        if (closed.getAndSet(true) == false) {
+              this.queue.ack(this.seqNums);
+        } else {
+            // TODO: how should we handle double-closing?
+            throw new IOException("double closing batch");
```
colinsurprenant: yes, agree. I am not even sure we should raise on double closing. thoughts?

logstash-core-event-java/src/main/java/org/logstash/Timestamp.java
colinsurprenant: not sure, @guyboertje ?

logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
```diff
@@ -0,0 +1,79 @@
+package org.logstash.ackedqueue;
+
+import org.logstash.common.io.CheckpointIOFactory;
+import org.logstash.common.io.PageIOFactory;
+
+public class FileSettings implements Settings {
```
colinsurprenant: yes, I think I agree here - having a Settings interface and specific Settings implementations seems not required here since both FileSettings and MemorySettings are actually identical objects that differ only in the constructor. 
+1
@guyboertje authored these bits, WDYT?

logstash-core/lib/logstash/environment.rb
colinsurprenant: `queue_capacity` has been renamed to `max_events` to avoid confusion for now.

logstash-core-queue-jruby/build.gradle
colinsurprenant: done

logstash-core-queue-jruby/build.gradle
colinsurprenant: fixed. actually enabled unchecked and added a specific suppress directive.

logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedBatchExtLibrary.java
```diff
@@ -0,0 +1,88 @@
+package org.logstash.ackedqueue.ext;
+
+import org.jruby.*;
```
colinsurprenant: fixed

logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedBatchExtLibrary.java
colinsurprenant: ah no. it cannot be `final`, because it must be assigned in the `initialize` method below. but it can be `private`.

colinsurprenant: fixed

logstash-core/lib/logstash/instrument/metric_store.rb
colinsurprenant: @guyboertje ?

logstash-core/lib/logstash/instrument/snapshot.rb
colinsurprenant: @guyboertje ?

logstash-core/src/main/java/org/logstash/common/io/ByteArrayStreamOutput.java
```diff
@@ -0,0 +1,65 @@
+package org.logstash.common.io;
+
+import java.nio.ByteBuffer;
+
+public class ByteArrayStreamOutput extends StreamOutput {
+    private byte[] bytes;
+
+    private int pos;
```
talevy: I think it would be clearer for this variable to be named `position` since its getter is `getPosition`

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -0,0 +1,248 @@
...
+    public static final int LENGTH_SIZE = Integer.BYTES;
+    public static final int SEQNUM_SIZE = Long.BYTES;
+    public static final int MIN_RECORD_SIZE = SEQNUM_SIZE + LENGTH_SIZE + CHECKSUM_SIZE;
+    public static final int HEADER_SIZE = 1;     // version byte
+    static final List<byte[]> EMPTY_READ = new ArrayList<>(0);
+
+    private final int capacity;
+    private final List<Integer> offsetMap; // has to be extendable
+    private final ByteBuffer buffer;
+    private long minSeqNum; // TODO: to make minSeqNum final we have to pass in the minSeqNum in the constructor and not set it on first write
```
talevy: is anything specific blocking this TODO?

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -0,0 +1,248 @@
...
+
+        this.buffer = ByteBuffer.allocate(capacity);
+        this.buffer.put(initialBytes);
+
+        this.offsetMap = new ArrayList<>();
+    }
+
+    @Override
+    public void open(long minSeqNum, int elementCount) throws IOException {
+        // TODO: do we need to do something there?
```
talevy: what is this question alluding to?

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -0,0 +1,248 @@
...
+        this.buffer.position(0);
+        this.buffer.put(VERSION);
+        this.head = 1;
+        this.minSeqNum = 0L;
+        this.elementCount = 0;
+    }
+
+    @Override
+    public int getCapacity() {
+        return this.capacity;
```
talevy: `this` is not necessary

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    public long getMinSeqNum() {
+        return this.minSeqNum;
```
talevy: `this` is not necessary

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    @Override
+    public void purge() throws IOException {
+        // do nothing
+    }
+
+    @Override
+    public void close() throws IOException {
+        // TBD
```
talevy: what does this mean?

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    private int checksum(byte[] bytes) {
+        checkSummer.reset();
+        checkSummer.update(bytes, 0, bytes.length);
+        return (int) checkSummer.getValue();
+    }
+
+    // TODO: static method for tests - should refactor
```
talevy: seems pretty simple to migrate this over to the tests since these variables are `public`, so this method can be removed

logstash-core/src/main/java/org/logstash/common/io/FileCheckpointIO.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public void purge(String fileName) throws IOException {
+        Path path = Paths.get(dirPath, fileName);
+        Files.delete(path);
+    }
+
+    @Override
+    public void purge() throws IOException {
+        // TODO: dir traversal and delete all checkpoints?
```
talevy: what are the ramifications of leaving this unimplemented? should we throw an UnsupportedOperationException if called?

logstash-core/src/main/java/org/logstash/common/io/BufferedChecksumStreamInput.java
```diff
@@ -0,0 +1,104 @@
...
+        if (reuse == null ) {
+            this.digest = new BufferedChecksum(new CRC32());
+        } else {
+            this.digest = reuse.digest;
+            digest.reset();
+            this.skipBuffer = reuse.skipBuffer;
+        }
+    }
+
+    public long getChecksum() {
```
talevy: is it safe for this to be a `long` while it's stream-byte-writable form is passed as an `int`?

talevy:
in general, can we move many of the `//` comments to proper `/**` javadoc style comments?

logstash-core/src/main/java/org/logstash/common/io/MemoryPageIOStream.java
```diff
@@ -0,0 +1,274 @@
...
+    public int persistedByteCount(byte[] data) {
+        return persistedByteCount(data.length);
+    }
+
+    @Override
+    public int persistedByteCount(int length) {
+        return MIN_RECORD_SIZE + length;
+    }
+
+    public MemoryPageIOStream(int pageNum, int capacity, String dirPath) throws IOException {
```
talevy: `pageNum` unused

logstash-core/src/main/java/org/logstash/common/io/MemoryPageIOStream.java
```diff
@@ -0,0 +1,274 @@
...
+                    throw new IOException(msg);
+                }
+
+                in.resetDigest();
+                byte[] bytes = in.readByteArray();
+                int actualChecksum = (int) in.getChecksum();
+                int expectedChecksum = in.readInt();
+
+                if (actualChecksum != expectedChecksum) {
+                    // explode with tragic error
```
talevy: this check is a noop?

logstash-core/src/main/java/org/logstash/common/io/MemoryPageIOStream.java
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    @Override
+    public void purge() throws IOException {
+        // do nothing
+    }
+
+    @Override
+    public void close() throws IOException {
+        // TBD
```
talevy: similar to above, what is to be determined?

logstash-core/src/main/java/org/logstash/common/io/MemoryPageIOStream.java
```diff
@@ -0,0 +1,274 @@
...
+        return minSeqNum;
+    }
+
+    // used in tests
+    public byte[] getBuffer() {
+        return buffer;
+    }
+
+    // used in tests
+    public String readHeaderDetails() throws IOException {
```
talevy: if this is used in tests, should be package scoped

logstash-core/src/main/java/org/logstash/common/io/MmapPageIO.java
```diff
@@ -0,0 +1,254 @@
...
+import java.nio.channels.FileChannel;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.zip.CRC32;
+import java.util.zip.Checksum;
+
+// TODO: this essentially a copy of ByteBufferPageIO and should be DRY'ed - temp impl to test file based stress test
```
talevy: temp impl for stress tests?

if this is only used for tests, maybe it belongs in the `src/test/java/org/logstash/..` directory?

logstash-core/src/main/java/org/logstash/common/io/MmapPageIO.java
```diff
@@ -0,0 +1,254 @@
...
+
+public class MmapPageIO implements PageIO {
+    public static final byte VERSION = 1;
+    public static final int CHECKSUM_SIZE = Integer.BYTES;
+    public static final int LENGTH_SIZE = Integer.BYTES;
+    public static final int SEQNUM_SIZE = Long.BYTES;
+    public static final int MIN_RECORD_SIZE = SEQNUM_SIZE + CHECKSUM_SIZE;
+    public static final int HEADER_SIZE = 1;     // version byte
+    static final List<byte[]> EMPTY_READ = new ArrayList<>(0);
```
talevy: is this meant to be package-private scoped?

logstash-core/src/main/java/org/logstash/common/io/MmapPageIO.java
```diff
@@ -0,0 +1,254 @@
...
+    @Override
+    public void activate() throws IOException {
+        if (this.channel == null) {
+            RandomAccessFile raf = new RandomAccessFile(this.file, "rw");
+            this.channel = raf.getChannel();
+            this.buffer = this.channel.map(FileChannel.MapMode.READ_WRITE, 0, this.capacity);
+            raf.close();
+            this.buffer.load();
+        } else {
+//            assert this.channel.isOpen() : String.format("closed channel");
```
talevy: can be removed?

logstash-core/src/main/java/org/logstash/common/io/PageIO.java
```diff
@@ -0,0 +1,50 @@
...
+    void activate() throws IOException;
+
+    // issue the proper data container "fsync" sematic
+    void ensurePersisted();
+
+    // delete/unlink/remove data file
+    void purge() throws IOException;
+
+    // add debugging details, to be added when creating a new page
+    // void setPageHeaderDetails(String details);
```
talevy: is this a todo? or can be this be removed

logstash-core/src/main/java/org/logstash/common/io/Streams.java
```diff
@@ -0,0 +1,61 @@
+package org.logstash.common.io;
+
+/*
```
talevy: license should be on top

logstash-core/src/main/java/org/logstash/common/io/Streams.java
```diff
@@ -0,0 +1,61 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Reader;
+/**
```
talevy: empty comment block

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -0,0 +1,388 @@
+package org.logstash.ackedqueue;
+
+import org.junit.Test;
+import org.logstash.common.io.ByteBufferPageIO;
+
+import java.io.IOException;
+import java.util.*;
```
talevy: glob import

logstash-core-event-java/src/main/java/org/logstash/Timestamp.java
guyboertje: Should be imported. This happened in the early days before I had IntelliJ set up for all LS in one UI.

guyboertje: Wait there is two TimestampSerializer classes, 1) in the `org.logstash` and 2) in `org.logstash.json`!!

@talevy @colinsurprenant - Which one do we keep?

logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueMemoryExtLibrary.java
guyboertje: When the Ruby based Read/Write Client and Batch code is stable I want to merge it into JRuby Extn. IIRC this is a TODO to return a ReadBatch or similar because ATM we have a `ReadBatch(RubyAckedBatch(Batch))`

logstash-core/src/main/java/org/logstash/common/io/ByteArrayStreamOutput.java
colinsurprenant: we will move the *Stream code away

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -0,0 +1,248 @@
...
+    static final List<byte[]> EMPTY_READ = new ArrayList<>(0);
+
+    private final int capacity;
+    private final List<Integer> offsetMap; // has to be extendable
+    private final ByteBuffer buffer;
+    private long minSeqNum; // TODO: to make minSeqNum final we have to pass in the minSeqNum in the constructor and not set it on first write
+    private int elementCount;
+    private int head;
+    private byte version;
+    private Checksum checkSummer = new CRC32();
```
colinsurprenant: yes

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: yeah, I would defer that refactor - I don't think it is urgent we do that now...

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: idem

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: this means it should be implemented at some point if needed. not sure we need to do anything since this is the in-memory page implementation which is only used for tests for now.

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: actually how would you suggest we migrate it to the tests? create a test subclass which exposes this method?

logstash-core/src/main/java/org/logstash/common/io/FileCheckpointIO.java
colinsurprenant: need to decide if this is a merge blocked, I don't think it is ... the purging fucntionality is not all fleshed out

logstash-core/src/main/java/org/logstash/common/io/MemoryPageIOStream.java
colinsurprenant: *Stream classes will be moved aside for now ...

logstash-core/src/main/java/org/logstash/common/io/MemoryPageIOStream.java
colinsurprenant: *Stream classes will be moved aside for now ...

logstash-core/src/main/java/org/logstash/common/io/MemoryPageIOStream.java
colinsurprenant: *Stream classes will be moved aside for now ...

logstash-core/src/main/java/org/logstash/common/io/MmapPageIO.java
```diff
@@ -0,0 +1,254 @@
...
+
+    private MappedByteBuffer buffer;
+    private File file;
+    private FileChannel channel;
+
+    private long minSeqNum; // TODO: to make minSeqNum final we have to pass in the minSeqNum in the constructor and not set it on first write
+    private int elementCount;
+    private int head;
+    private byte version;
+    private Checksum checkSummer = new CRC32();
```
colinsurprenant: agree

logstash-core/src/main/java/org/logstash/common/io/MmapPageIO.java
colinsurprenant: han - need to review this.

logstash-core/src/main/java/org/logstash/common/io/Streams.java
colinsurprenant: yup.  *Stream classes will be moved aside for now ...

logstash-core/src/main/java/org/logstash/common/io/Streams.java
colinsurprenant: *Stream classes will be moved aside for now ...

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
talevy: this method seems to be static param arithmetic, cannot it not be done inline?

logstash-core/lib/logstash/instrument/metric_type/base.rb
```diff
@@ -1,5 +1,5 @@
 # encoding: utf-8
-require "logstash/event"
+# require "logstash/event"
```
guyboertje: Lets

logstash-core/lib/logstash/instrument/snapshot.rb
guyboertje: lets

logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueMemoryExtLibrary.java
colinsurprenant: @guyboertje  can you expand a bit the comment so that future readers understand it? thanks.

logstash-core/gemspec_jars.rb
colinsurprenant: they are

logstash-core/lib/logstash/instrument/metric_store.rb
colinsurprenant: removed

logstash-core/lib/logstash/instrument/metric_type/base.rb
colinsurprenant: removed

logstash-core/lib/logstash/instrument/snapshot.rb
colinsurprenant: removed

logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
```diff
@@ -0,0 +1,44 @@
...
+        this.queue = q;
+        this.closed = new AtomicBoolean(false);
+    }
+
+    // close acks the batch ackable events
+    public void close() throws IOException {
+        if (closed.getAndSet(true) == false) {
+              this.queue.ack(this.seqNums);
+        } else {
+            // TODO: how should we handle double-closing?
```
colinsurprenant: it's a TODO. Undecided if we could just ignore double-closing or we should throw an exception. there's a P1 task to that effect. I think we can ignore this for now

logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
colinsurprenant: I think we can ignore for now as per previous comment.

logstash-core/src/main/java/org/logstash/common/io/ByteArrayStreamOutput.java
colinsurprenant: actually, only MemoryPageIOStream is unused. all the rest of the *Stream classes are used for FileCheckpointIO
@guyboertje please followup

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: I can expand on the TBD.

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: but I want to keep it DRY and not do this size evaluation at two different places. let's defer this and not block the merge, this is a minor issue, the more important is that this is done in only one place.

logstash-core/src/main/java/org/logstash/common/io/BufferedChecksumStreamInput.java
colinsurprenant: actually, only MemoryPageIOStream is unused. all the rest of the *Stream classes are used for FileCheckpointIO
@guyboertje please followup

logstash-core/src/main/java/org/logstash/common/io/MemoryPageIOStream.java
colinsurprenant: move in wip package

logstash-core/src/main/java/org/logstash/common/io/MemoryPageIOStream.java
colinsurprenant: moved in wip package

logstash-core/src/main/java/org/logstash/common/io/MmapPageIO.java
colinsurprenant: fixed. moved into a TODO

logstash-core/src/main/java/org/logstash/common/io/Streams.java
colinsurprenant: fixed

logstash-core/src/main/java/org/logstash/common/io/wip/MemoryPageIOStream.java
```diff
@@ -0,0 +1,279 @@
+package org.logstash.common.io.wip;
```
talevy: what is this package?

logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
guyboertje: @colinsurprenant - agree.

logstash-core/src/main/java/org/logstash/common/io/wip/MemoryPageIOStream.java
colinsurprenant: as discussed this is a temporary package to hold the unfinished `MemoryPageIOStream.java` so it can benefit any code refactor etc.

suyograo:
Closing this as it is already merged.

PR #5967, opened 2016-09-23
Fix generated jar dependencies for ls-event-java

talevy:
This file was incorrectly checked-in with a jackson-annotations
dependency of `2.7.3`. jackson-databind depends on `2.7.0` of the
annotations library.


suyograo:
LGTM. I incorrectly did this in https://github.com/elastic/logstash/commit/8bb01d6358c071f4d07803018c5e4f939dbba2f0

PR #5968, opened 2016-09-23
Rats improvements for travis runs

suyograo:
This PR reduces usage of random `sleep` calls in bash scripts. It replaces them with a poll-wait loop in many cases which works consistently. It also removes excess debug information from the output which can make it harder to track test runs.

Also `/services/kafka` etc is moved to `/services/installed/xyz` which means we can add `/services/installed` to `.gitignore`.

### DIFF
--- a/qa/integration/.gitignore
+++ b/qa/integration/.gitignore
@@ -1,5 +1,2 @@
-/services/filebeat
-/fixtures/how.input
-/services/elasticsearch
-/services/kafka
+/services/installed
 /fixtures/certificates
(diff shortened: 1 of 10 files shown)


qa/integration/.gitignore
```diff
@@ -1,5 +1,2 @@
-/services/filebeat
-/fixtures/how.input
-/services/elasticsearch
-/services/kafka
+/services/installed
```
untergeek: I definitely like this.

untergeek:
This is very nice!

LGTM™

PR #5978, opened 2016-09-29
Require 'forwardable' in SafeURI

andrewvc:
SafeURI requires forwardable, which is included elsewhere in logstash. Plugin tests using this validation don't load logstash in the right order for this to work, so we need an explicit require here.

### DIFF
--- a/logstash-core/lib/logstash/util/safe_uri.rb
+++ b/logstash-core/lib/logstash/util/safe_uri.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 require "logstash/namespace"
 require "logstash/util"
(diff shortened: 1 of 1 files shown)


talevy:
LGTM

maxious:
Adding for future searchers, this PR fixes 
```
NameError: uninitialized constant LogStash::Util::SafeURI::Forwardable
  const_missing at org/jruby/RubyModule.java:2726
        SafeURI at /home/lubr/.rvm/gems/jruby-1.7.19/gems/logstash-core-5.0.0.pre.beta1-java/lib/logstash/util/safe_uri.rb:11
```

PR #5991, opened 2016-10-03
Update to new ES output

suyograo:
Looks like we don't need the docker API gems after this PR: https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/486

### DIFF
--- a/Gemfile.jruby-1.9.lock
+++ b/Gemfile.jruby-1.9.lock
@@ -407,7 +407,7 @@ GEM
       logstash-filter-json
       logstash-input-generator
       logstash-output-file
-    logstash-output-elasticsearch (5.1.1-java)
+    logstash-output-elasticsearch (5.1.2-java)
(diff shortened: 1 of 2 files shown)


Gemfile.jruby-1.9.lock
```diff
@@ -503,7 +503,7 @@ GEM
     murmurhash3 (0.1.6-java)
     mustache (0.99.8)
     naught (1.1.0)
-    nokogiri (1.6.8-java)
+    nokogiri (1.6.8.1-java)
```
andrewvc: Why the nokogiri changes?

andrewvc:
LGTM, though curious why nokogiri changed

Gemfile.jruby-1.9.lock
suyograo: not sure. I ran a bundle update, so just a new version. I supposed its harmless since it's a point release..

PR #6006, opened 2016-10-06
Support for multiple pipelines in the same logstash instance [wip]

jsvd:
Solves #6521 

This PR adds a new setting flag called `config.multi_pipeline` which will read the `pipelines` array property of the logstash.yml and load each as a new pipeline.

Example:

```yml
config.multi_pipeline: true
pipelines:
  - pipeline.id: line
    config.string: "input { generator {} } filter { sleep { time => 1 } } output { stdout { codec => line } }"
    pipeline.workers: 1
  - pipeline.id: dots
    config.string: "input { generator {} } filter { sleep { time => 0.25 } } output { stdout { codec => dots } }"
    pipeline.workers: 2
```

This is a work in progress PR as there are still some challenges to overcome:
- [ ] make metrics api work with multiple pipelines
- [ ] define which metrics should be configurable per pipeline. see https://github.com/elastic/logstash/issues/6492

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -159,3 +159,4 @@
 #
 # Where to find custom plugins
 # path.plugins: []
+config.multi_pipeline: true
--- a/logstash-core/lib/logstash/environment.rb
+++ b/logstash-core/lib/logstash/environment.rb
@@ -21,6 +21,7 @@ module LogStash
  Setting::WritableDirectory.new("path.data", ::File.join(LogStash::Environment::LOGSTASH_HOME, "data")),
     Setting::NullableString.new("config.string", nil, false),
            Setting::Boolean.new("config.test_and_exit", false),
+           Setting::Boolean.new("config.multi_pipeline", false),
            Setting::Boolean.new("config.reload.automatic", false),
            Setting::Numeric.new("config.reload.interval", 3), # in seconds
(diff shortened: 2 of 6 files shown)


acchen97:
That was quick, thanks for getting started on this @jsvd. Many users today use the "merge separate config files into one pipeline" feature today. Turning this multiple pipelines feature on would break that, but I think that's fine so long as its off by default. Few items for discussion:

- This current implementation has each config with its own input/filter/output definitions. I can see use cases where a company would want to use a centralized Logstash with a common input, but have different organizations split/isolate the processing of different streams (filters/outputs) through this same input. Is this something that we could further enable beyond just using conditionals today? There's architectural pros/cons for this segregation at the input stage, but we know exposing additional ports can be tough in some scenarios.

- We should also take into account how we would enable RBAC for different pipelines. My initial thoughts tend to veer towards just enabling this with a security layer atop an API that allows for updating the pipelines dynamically.

- Can different pipelines be managed granularly? Like shutting down one pipeline while leaving the others active? Or is it an all-up or all-down kinda thing?

/cc @suyograo @jordansissel

jsvd:
closing this PR as this feature will be merged from the feature branch https://github.com/elastic/logstash/tree/feature/multi_pipeline when finished.

PR #6008, opened 2016-10-07
Add Settings#validate_all.

### LINKED ISSUE #6004: Logstash checks default directory for write permissions always, even when overriden
Working with @jasontedor on the homebrew integration we got the following error:

```
==> /usr/local/Cellar/logstash/5.0.0-beta1/bin/logstash -e --path.data /tmp/logstash-test-20161006-58368-1nv7nm9/data
ArgumentError: Path "/usr/local/Cellar/logstash/5.0.0-beta1/libexec/data" is not a directory or not writable.
  initialize at /usr/local/Cellar/logstash/5.0.0-beta1/libexec/logstash-core/lib/logstash/settings.rb:372
        call at org/jruby/RubyProc.java:281
    validate at /usr/local/Cellar/logstash/5.0.0-beta1/libexec/logstash-core/lib/logstash/settings.rb:174
  initialize at /usr/local/Cellar/logstash/5.0.0-beta1/libexec/logstash-core/lib/logstash/settings.rb:127
  initialize at /usr/local/Cellar/logstash/5.0.0-beta1/libexec/logstash-core/lib/logstash/settings.rb:368
    LogStash at /usr/local/Cellar/logstash/5.0.0-beta1/libexec/logstash-core/lib/logstash/environment.rb:18
      (root) at /usr/local/Cellar/logstash/5.0.0-beta1/libexec/logstash-core/lib/logstash/environment.rb:9
     require at org/jruby/RubyKernel.java:1040
      (root) at /usr/local/Cellar/logstash/5.0.0-beta1/libexec/logstash-core/lib/logstash/runner.rb:1
     require at org/jruby/RubyKernel.java:1040
      (root) at /usr/local/Cellar/logstash/5.0.0-beta1/libexec/lib/bootstrap/environment.rb:67
Error: logstash: failed
```

It appears that when we instantiate the default data directory ( https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/environment.rb#L21 ) we check that directory is writable, even if the user wants a different. This is due to the fact that Setting::WritableDirectory checks for writability in its constructor. We should only check this after all settings are merged.

jordansissel:
All setting validations are now deferred until after setting sources have been processed (flags, logstash.yml, etc)

Deferred validation fixes #6004.

### DIFF
--- a/logstash-core/lib/logstash/settings.rb
+++ b/logstash-core/lib/logstash/settings.rb
@@ -95,6 +95,12 @@ module LogStash
       self.merge(flatten_hash(settings))
     end
 
+    def validate_all
+      @settings.each do |name, setting|
+        setting.validate_value
+      end
+    end
+
     private
     def read_yaml(path)
       YAML.safe_load(IO.read(path)) || {}
@@ -123,8 +129,9 @@ module LogStash
       @validator_proc = validator_proc
       @value = nil
       @value_is_set = false
+      @strict = strict
 
-      validate(default) if strict
+      validate(default) if @strict
       @default = default
     end
 
@@ -136,8 +143,12 @@ module LogStash
       @value_is_set
     end
 
+    def strict?
+      @strict
+    end
+
     def set(value)
-      validate(value)
+      validate(value) if @strict
       @value = value
       @value_is_set = true
       @value
@@ -167,12 +178,18 @@ module LogStash
       self.to_hash == other.to_hash
     end
 
-    private
-    def validate(value)
-      if !value.is_a?(@klass)
-        raise ArgumentError.new("Setting \"#{@name}\" must be a #{@klass}. Received: #{value} (#{value.class})")
-      elsif @validator_proc && !@validator_proc.call(value)
-        raise ArgumentError.new("Failed to validate setting \"#{@name}\" with value: #{value}")
+    def validate_value
+      validate(value)
+    end
+
+    protected
+    def validate(input)
+      if !input.is_a?(@klass)
+        raise ArgumentError.new("Setting \"#{@name}\" must be a #{@klass}. Received: #{input} (#{input.class})")
+      end
+
+      if @validator_proc && !@validator_proc.call(input)
+        raise ArgumentError.new("Failed to validate setting \"#{@name}\" with value: #{input}")
       end
     end
 
@@ -351,6 +368,13 @@ module LogStash
       end
     end
 
+    class NullableString < String
+      def validate(value)
+        return if value.nil?
+        super(value)
+      end
+    end
+
     class ExistingFilePath < Setting
       def initialize(name, default=nil, strict=true)
         super(name, ::String, default, strict) do |file_path|
@@ -364,7 +388,7 @@ module LogStash
     end
 
     class WritableDirectory < Setting
-      def initialize(name, default=nil, strict=true)
+      def initialize(name, default=nil, strict=false)
         super(name, ::String, default, strict) do |path|
           if ::File.directory?(path) && ::File.writable?(path)
             true
@@ -378,3 +402,4 @@ module LogStash
 
   SETTINGS = Settings.new
 end
+
--- a/logstash-core/lib/logstash/environment.rb
+++ b/logstash-core/lib/logstash/environment.rb
@@ -17,9 +17,9 @@ module LogStash
 
   [
             Setting::String.new("node.name", Socket.gethostname),
-            Setting::String.new("path.config", nil, false),
+    Setting::NullableString.new("path.config", nil, false),
  Setting::WritableDirectory.new("path.data", ::File.join(LogStash::Environment::LOGSTASH_HOME, "data")),
-            Setting::String.new("config.string", nil, false),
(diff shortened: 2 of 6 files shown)


jordansissel:
Testing that #6004 is fixed:

master, fails:

```
% chmod 0 ./data
% bin/logstash --debug -e '' --path.data /tmp/example
ArgumentError: Path "/home/jls/projects/logstash/data" is not a directory or not writable.
```

this PR, succeeds:

```
% chmod 0 ./data
% bin/logstash --debug -e '' --path.data /tmp/example
...
[2016-10-06T15:57:13,378][DEBUG][logstash.runner          ] *path.data: "/tmp/example" (default: "/home/jls/projects/logstash/data")
...
```

jordansissel:
I wanted to do a larger refactor, but my goal was to do a minimal change in time for 5.0.0.

This PR brings in some changes already in the feature/java_persistence branch (adds Setting::Bytes, improves WritableDirectory). Bytes wasn't really needed, but was already in the patch I imported, and isn't used yet, so I felt was safe to include in this PR.

logstash-core/lib/logstash/settings.rb
```diff
@@ -364,17 +372,78 @@ def initialize(name, default=nil, strict=true)
...
         end
+
+        # If we get here, the directory exists and is writable.
+        true
+      end
+
+      def value
+        super.tap do |path|
+          if !::File.directory?(path)
+            # Create the directory if it doesn't exist.
```
jsvd: what is the use case for this implicit directory creation?
Not sure how strict we should be about this, IMO we should just check if it's there and it's writable, if not, throw an error, and leave the dir creation to the provisioning tools, thoughts?

also, this could prevent "silent" directory creating (I know dir creating action is logged but stillll.....).

logstash-core/lib/logstash/settings.rb
```diff
@@ -364,17 +372,78 @@ def initialize(name, default=nil, strict=true)
...
+      end
+
+      def coerce(value)
+        case value
+        when ::Numeric
+          value
+        when ::String
+          LogStash::Util::ByteValue.parse(value)
+        else
+          raise ArgumentError.new("Could not coerce '#{value}' into a bytes value")
```
jsvd: I'd suggest also logging the value's class

jordansissel:
Hmm yeah I saw lots of Travis failures last night. I will attend to that
when I am back at work. ❤️

logstash-core/lib/logstash/settings.rb
jordansissel: Two use cases. First, for the PQ that's coming -- the default path si probably going to be `{path.data}/queue/...` and it'd be nice not to have to have all these empty directories. Second, for users who just want to specify `{path.data}` and that would imply the queue path (or similar) would be relative to `{path.data}` and would be nice if they were automatically created if possible. My hope is to avoid having our docs say `Run these 7 mkdir commands before running logstash.`

jordansissel:
Switched the empty-default string settings with a newer NullableString. I'm not happy with this as a solution, but I think as a short term minimal fix it is ok.

There's still two more failing tests. I will get to these shortly.

jsvd:
what if we added a `set?` predicate to know if a setting was ever `.set(..)` ? this way we wouldn't rely on the value to distinguish between an `nil`/`""`/etc?

logstash-core/lib/logstash/settings.rb
```diff
@@ -364,12 +388,45 @@ def initialize(name, default=nil, strict=true)
...
+        true
+      end
+
+      def value
+        super.tap do |path|
+          if !::File.directory?(path)
+            # Create the directory if it doesn't exist.
+            begin
+              # TODO(sissel): Log the fact that we are creating this directory.
+              ::FileUtils.mkdir_p(path)
```
suyograo: @jordansissel can we log (debug?) that we created this dir?

suyograo: ooh just saw your todo. my bad :)

jordansissel: but yes, I agree.

jordansissel: Done.

jordansissel:
I did some surgery to this PR. Changes:

* This change only introduces validate_all and NullableString for #6004 
* Moved Setting::Bytes addition to https://github.com/elastic/logstash/pull/6022
* Moved WritableDirectory improvements to https://github.com/elastic/logstash/pull/6023

jordansissel:
Manual test `% bin/logstash --debug -e '' --path.data /tmp/example` is now passing on this PR once again. Woo.

jordansissel:
@jsvd can you re-review?

jsvd:
any thoughts on my suggestion in https://github.com/elastic/logstash/pull/6008#issuecomment-252567713?

Otherwise LGTM (test failure is unrelated)

jordansissel:
@jsvd Yeah I think adding `#set?` could be useful, though I wonder about the blurred line between a 'set' and 'default' value when it comes to validation. We can work on that later. <3

Issue #5969, opened 2016-09-26
Exception messages are lost in Logstash 2.4


magnusbaeck:
Starting in Logstash 2.4 the exception messages are lost, so you're only getting a stacktrace and have to resort to reading the source code to figure out what Logstash is complaining about. Given this invalid configuration:

```
input { file { path => "foo/bar" } }
output { stdout { } }
```

Logstash 2.3.2 says this:

```
$ /opt/logstash/bin/logstash -f test.config
Settings: Default pipeline workers: 8
Pipeline aborted due to error {:exception=>#<ArgumentError: File paths must be absolute, relative path specified: foo/bar>, :backtrace=>["/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-file-2.2.5/lib/logstash/inputs/file.rb:187:in `register'", "org/jruby/RubyArray.java:1613:in `each'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-input-file-2.2.5/lib/logstash/inputs/file.rb:185:in `register'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.2-java/lib/logstash/pipeline.rb:330:in `start_inputs'", "org/jruby/RubyArray.java:1613:in `each'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.2-java/lib/logstash/pipeline.rb:329:in `start_inputs'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.2-java/lib/logstash/pipeline.rb:180:in `start_workers'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.2-java/lib/logstash/pipeline.rb:136:in `run'", "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-2.3.2-java/lib/logstash/agent.rb:465:in `start_pipeline'"], :level=>:error}
stopping pipeline {:id=>"main"}
```

This is a really bad error message to show to an end user who made a trivial configuration mistake, but it gets even worse in 2.4.0:

```
$ /opt/logstash/bin/logstash -f test.config
Settings: Default pipeline workers: 8
[...]

PR #6011, opened 2016-10-07
avoid logging exception objects in agent.rb

jsvd:
also provide proper information about the error message instead of just the class name + backtrace.

This needs to be merged into 2.x and 2.4

fixes https://github.com/elastic/logstash/issues/5969


suyograo:
LGTM

PR #6023, opened 2016-10-11
WritableDirectory#value now creates the directory if needed.

jordansissel:
This also makes the failure reports from WritableDirectory's validation
more specific (parent directory doesn't exist, or does exist and is
isn't writable, or the path itself isn't writable, etc).


colinsurprenant:
LGTM - did not test.

jordansissel:
Travis finally ran the tests. woot. Green.

PR #6030, opened 2016-10-11
Fix build

dedemorton:
Link needs to be hard coded here because {ref} resolves to a URL in "current" but the breaking changes topic for 5.0 does not exist there yet. 

Note that I've already pushed this change to logstash-docs repo.


suyograo:
LGTM

PR #6034, opened 2016-10-12
fix warning message when set USE_DRIP=1 and no JAVA_OPTS set in system

truebit:
When launching logstash with option `USE_DRIP=1` like 
    
    USE_DRIP=1 ./bin/logstash

in function `setup_drip`, it would set `$JAVA_OPTS`; then in follwing function `setup_java`, it would output warning message that default $JAVA_OPTS would be overwritten. But actually user did not set "default $JAVA_OPTS".

This fix would not output such warning message if user did not set $JAVA_OPTS


ph:
Would you mind doing step 2 of the contributing guide? https://github.com/elasticsearch/logstash/blob/master/CONTRIBUTING.md#contribution-steps

truebit:
CLA signed

suyograo:
LGTM

Issue #6003, opened 2016-10-06
configuration reloading flaky test


jsvd:
```
 1) LogStash::Agent#execute when auto_reload is true if state is clean should periodically reload_state
     Failure/Error: expect(subject).to receive(:reload_state!).at_least(3).times
       (#<LogStash::Agent:0x4039dba2>).reload_state!(any args)
           expected: at least 3 times with any arguments
           received: 2 times with any arguments
     # ./logstash-core/spec/logstash/agent_spec.rb:158:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:42:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:248:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:243:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:180:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:173:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:152:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:117:in `run_with_threads'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:102:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:80:in `run'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:178:in `standard_exception_handling'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:77:in `run'
[...]

Issue #5445, opened 2016-06-03
Intermittent failure of agent spec


andrewvc:
I got  an error on https://travis-ci.org/elastic/logstash/builds/135021910

```

  1) LogStash::Agent#execute when auto_reload is true if state is clean should periodically reload_state
     Failure/Error: expect(subject).to receive(:reload_state!).at_least(3).times
       (#<LogStash::Agent:0x250268ef>).reload_state!(any args)
           expected: at least 3 times with any arguments
           received: 1 time with any arguments
     # ./logstash-core/spec/logstash/agent_spec.rb:158:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.8/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:42:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:248:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:243:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:180:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/task.rb:173:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:150:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:106:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:115:in `run_with_threads'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:100:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:78:in `run'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.1.2/lib/rake/application.rb:176:in `standard_exception_handling'
[...]

PR #6039, opened 2016-10-12
refactor agent specs to make them more predictable

jsvd:
* avoid the use of sleeps
* remove unused mocking of logger
* ensure pipelines are shut down after each test

fixes https://github.com/elastic/logstash/issues/6003 and https://github.com/elastic/logstash/issues/5445


suyograo:
woot! LGTM

PR #6045, opened 2016-10-13
Add Event method_missing to give better error messages for #[] and #[]=

guyboertje:
Reviewer:
- Please read this - https://github.com/logstash-plugins/logstash-filter-ruby/issues/29
- Then assess whether this is a suitable solution.

I elected to add a method_missing handler for the Event class.

I chose not to do this in the JRuby Extension because IMO it will be easier to remove it later.

If it seems like we need to keep it for all time, we can move it to the JRuby Extension.


jsvd:
I like this idea. Since it's a debatable thing, I vote for getting a second or third LGTM.

That said, LGTM

logstash-core-event-java/lib/logstash-core-event-java/logstash-core-event-java.rb
```diff
@@ -25,4 +25,16 @@ module LogStash
...
\ No newline at end of file
+require "logstash/timestamp"
+
+module LogStash class Event
+  def method_missing(method_name, *arguments, &block)
+    if /^\[\]$/.match(method_name.to_s)
+      raise NoMethodError.new("Direct event field references (i.e. event['field']) have been disabled in favor of using event get and set methods (e.g. event.get('field')). Please consult the Logstash 5.0 breaking changes documentation for more details.")
+    end
+    if /^\[\]=$/.match(method_name.to_s)
+      raise NoMethodError.new("Direct event field references (i.e. event['field'] = 'value') have been disabled in favor of using event get and set methods (e.g. event.set('field', 'value')). Please consult the Logstash 5.0 breaking changes documentation for more details.")
```
guyboertje: OK - a frozen one, perhaps.

colinsurprenant:
I am good with this approach, solves the problem and will not interfere in normal operation. 
But, I think this should be added in `logstash/logstash-core-event-java/lib/logstash/event.rb` no? any reason for adding it in `logstash-core-event-java/lib/logstash-core-event-java/logstash-core-event-java.rb` ?

guyboertje:
@colinsurprenant 
My logic **was** to add the method missing after the jar was loaded (or failed trying) - so I added it in the current place.
It occurs to me now that I can still achieve this by adding it after the `require "logstash-core-event"` in the event.rb.
Will move it.

colinsurprenant:
Is is actually not important to have the jar loaded or not ... this is in the Ruby Event ...  this is the purpose of `event.rb`, to add Ruby stuff to the Event class.

logstash-core-event-java/lib/logstash/event.rb
```diff
@@ -31,7 +31,24 @@ def shutdown?; false; end;
   SHUTDOWN = ShutdownEvent.new
 end
 
+module LogStash class Event
```
colinsurprenant: this is nitpicking but why not define the `class Event` in the previous `module LogStash` definition ?

guyboertje: true

colinsurprenant:
Great! :D  if all tests are passing I'm LGTM

guyboertje:
got 2 LGTM - squashing

PR #6042, opened 2016-10-13
Deprecate uninstall for Plugin Manager

suyograo:
Adds remove and deprecates uninstall to be consistent with rest of Elastic stack

Fixes #6041

PR #6051, opened 2016-10-13
Fix acceptance tests

suyograo:
#6042 broke acceptance tests, so fixing it

### DIFF
--- a/qa/acceptance/spec/shared_examples/cli/logstash-plugin/remove.rb
+++ b/qa/acceptance/spec/shared_examples/cli/logstash-plugin/remove.rb
@@ -3,8 +3,8 @@ require_relative "../../../spec_helper"
 require "logstash/version"
 require "fileutils"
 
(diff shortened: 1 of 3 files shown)


andrewvc:
Does this just fix the test name, or does the described name actually affect the test?

jsvd:
I guess I'm not familiar with the framework, but this pr only changes test descriptions, is that actually a fix?

suyograo:
@andrewvc the test fix was renaming `shared_examples "logstash uninstall" do |logstash|` to `shared_examples "logstash remove" do |logstash|`. I edited the describes to make it more descriptive, since it was confusing (whether it's dealing with LS or LS plugins bin)

suyograo:
This fixes the test: https://github.com/elastic/logstash/pull/6051/files#diff-6f164f39ab48c19f918b531c9192b01cR6

andrewvc:
OIC. LGTM (assuming tests pass)

PR #6058, opened 2016-10-14
Add autostart instructions to apt & yum setups

cwurm:
I guess this should also be merged into the 5.x and 2.x branches.

### DIFF
--- a/docs/static/getting-started-with-logstash.asciidoc
+++ b/docs/static/getting-started-with-logstash.asciidoc
@@ -105,6 +105,22 @@ it with:
 sudo apt-get update && sudo apt-get install logstash
 --------------------------------------------------
 
+Configure Logstash to automatically start during bootup. If your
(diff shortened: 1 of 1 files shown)


docs/static/getting-started-with-logstash.asciidoc
```diff
@@ -105,6 +105,22 @@ it with:
 sudo apt-get update && sudo apt-get install logstash
 --------------------------------------------------
 
+Configure Logstash to automatically start during bootup. If your
+distribution is using SysV init, then you will need to run:
+
+[source,sh]
+--------------------------------------------------
+sudo update-rc.d logstash defaults 95 10
```
inqueue: `defaults` indicates the runlevel where `95 10` indicates where the script goes in startup and shutdown sequence

jakommo: ok, then LGTM

Issue #6055, opened 2016-10-13
cancelled events are not evicted from batch


talevy:
Dropping events using `drop {}` does not actually drop events

recreate using:

```
input {
    generator {
        count => 1
        message => "hello"
    }
}

filter {
    drop {}
    mutate { add_tag => ["hello"] }
}

output {
    stdout { codec => rubydebug }
}

```

following change fixes this, but makes things a little less efficient

```ruby
diff --git a/logstash-core/lib/logstash/filters/base.rb b/logstash-core/lib/logstash/filters/base.rb
index 35bf49e..310d87b 100644
--- a/logstash-core/lib/logstash/filters/base.rb
+++ b/logstash-core/lib/logstash/filters/base.rb
@@ -151,10 +151,10 @@ class LogStash::Filters::Base < LogStash::Plugin
     LogStash::Util.set_thread_plugin(self)
     result = []
     events.each do |event|
-      unless event.cancelled?
         result << event
-        filter(event){|new_event| result << new_event}
-      end
+        unless event.cancelled?
+          filter(event){|new_event| result << new_event}
+        end
     end
     result
   end
```

colinsurprenant:
This is not the proper fix. It solves as a side effect. Working on a PR.

colinsurprenant:
proposed fix in #6059

colinsurprenant:
fixed by #6059

PR #6059, opened 2016-10-14
fix for cancelled events issue #6055

colinsurprenant:
Solves the bug described in issue #6055 where dropped events are still being output.

The problem is that Event cancellation is done by mutating the Event itself by calling Event#cancel and thus the Batch object is not informed of this change and cannot keep an internal state of the cancelled events as the events are processed through the `filter_func`. 

Ultimately the proper fix will be to surface the Batch object as the main container for keeping tabs on the Events lifecycle (originals, generated, cancelled, etc).

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -148,7 +148,10 @@ module LogStash; module Util
         @shutdown_signal_received = false
         @flush_signal_received = false
         @originals = Hash.new
-        @cancelled = Hash.new
+
+        # TODO: disabled for https://github.com/elastic/logstash/issues/6055 - will have to properly refactor
+        # @cancelled = Hash.new
+
         @generated = Hash.new
         @iterating_temp = Hash.new
         @iterating = false # Atomic Boolean maybe? Although batches are not shared across threads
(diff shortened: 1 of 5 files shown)


logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -100,6 +100,56 @@ class TestPipeline < LogStash::Pipeline
...
+            drop {}
+          }
+          mutate { add_tag => ["notdropped"] }
+        }
+        output { dummyoutputwitheventsarray {} }
+      CONFIG
+    end
+
+    it "should not propage cancelled events from filter to output" do
+      Thread.abort_on_exception = true
```
jsvd: in the before block we should store the current value of abort_on_exception, set it to true, and then restore it to the original value in the after block

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -148,7 +148,10 @@ def initialize(queue, size, wait)
         @shutdown_signal_received = false
         @flush_signal_received = false
         @originals = Hash.new
-        @cancelled = Hash.new
+
+        # disabled for https://github.com/elastic/logstash/issues/6055
+        # @cancelled = Hash.new
```
suyograo: Can we remove this at all?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -197,7 +205,9 @@ def filtered_size
       end
 
       def cancelled_size
-        @cancelled.size
+      # disabled for https://github.com/elastic/logstash/issues/6055
+      raise("cancelled_size is unsupported ")
+        # @cancelled.size
```
suyograo: same here. remove commented code?

logstash-core/spec/logstash/pipeline_spec.rb
colinsurprenant: ok, good point - that was a copy-pasta. we should review all specs for this then...

suyograo:
@colinsurprenant why are some stuff commented out?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
colinsurprenant: I would leave it for now since this is a quick fix and we may come back to it and "redo" this for some reason

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
colinsurprenant: same answer

colinsurprenant:
because this is really a quick fix and that code will have to be correctly refactored, just not to loose  context

logstash-core/spec/logstash/pipeline_spec.rb
jsvd: I know we have examples of this store, use, restore somewhere in the spec suite of agent_spec.rb

jsvd:
terrible, terrible code. tests pass. LGTM

suyograo:
LGTM

colinsurprenant:
merged in master, 5.0 and 5.x

PR #6061, opened 2016-10-15
Fix user feedback message for package install

suyograo:
We were showing both INFO and ERROR for all cases which is confusing

```
sudo -u logstash bin/logstash -e 'input { stdin { } } output { stdout {} }'
INFO: Logstash has a new settings file which defines start up time settings. This file is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a package and are starting it manually please specify the location to this settings file by passing in "--path.settings=/path/.." in the command line options
ERROR: Failed to load settings file from "path.settings". Aborting... path.setting=/usr/share/logstash/config, exception=Errno::ENOENT, message=>No such file or directory - /usr/share/logstash/config/logstash.yml
```

now:

```
ubuntu@ip-10-0-2-201:/usr/share/logstash$ sudo -u logstash bin/logstash -e 'input { stdin { } } output { stdout {} }'
ERROR: Logstash requires a setting file which is typically located in $LS_HOME/config or /etc/logstash. If you installed Logstash through a packa....
```

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -162,10 +162,14 @@ class LogStash::Runner < Clamp::StrictCommand
 
     begin
       LogStash::SETTINGS.from_yaml(LogStash::SETTINGS.get("path.settings"))
+    rescue Errno::ENOENT 
+      unless cli_help?(args)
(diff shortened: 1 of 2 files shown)


jsvd:
LGTM

PR #6068, opened 2016-10-17
Add IMPORTANT block for path.settings

untergeek:
As this breaking change will catch the unsuspecting off guard, make it very visible in the breaking changes documentation

Apologies for my editor pruning so many trailing spaces.


dedemorton:
LGTM, assuming it builds OK in asciidoc. I've not used that syntax for notes before.

untergeek:
It should.  It's one of the 2 acceptable syntaxes for note blocks.

PR #6075, opened 2016-10-18
Fix bug where old-style plugin libraries were not being loaded

jordansissel:
This fixes a bug found in 5.0.0rc1 where plugins in `path.plugins` (old-style, deprecated, should not be used anymore, please stop writing plugins like this!) sourced plugins would not be loaded.

I am on the fence about merging this, and would rather just this feature and require folks to write plugins as gems. The ruby filter can continue to exist to support users who want single-file plugins..


jsvd:
We can discuss the deprecation of this feature for sure, but I think that if it's there it should work properly..so... LGTM

jordansissel:
#6109 replace this, I believe, and is merged.

PR #6084, opened 2016-10-18
don't rely on complete queue drain in agent specs

jsvd:
with persistence we no longer wait for total queue drain before
terminating the pipeline. Some of the specs for the Agent class use the
count of events created by the generator plugin to ensure the metric
counters are correct.
This commit changes the assertion strategy to not depend on exact
values, but instead on their properties.


colinsurprenant:
LGTM

PR #6092, opened 2016-10-20
Update Gemfile

matarrese:
Hi, I would like to include the last version 4.1.2 of the plugin logstash-input-jdbc


matarrese:
Hi, I have signed the agreement, and I have tested this change.
Hope to get an update on this. 
Thanks,

suyograo:
@matarrese thanks, we'll release a 2.4.1 soon, but not sure we can include 4.1.2 in 2.4.1 since it breaks config. 

What specific fix in JDBC are you looking at?

matarrese:
Hi @suyograo ,sorry for this late answer, this is the fix that I would like to include:
https://github.com/logstash-plugins/logstash-input-jdbc/pull/156

suyograo:
Cool, we'll release this in 2.4.1. I will update the lock file before the release.

PR #6093, opened 2016-10-20
skip reloading a pipeline if it has reloading turned off

jsvd:
this is ground work in preparation for supporting multiple pipelines

Even if the reload setting is globally true, a pipeline can have this disabled on its own settings.

implements #6094


suyograo:
LGTM

Issue #6079, opened 2016-10-18
Allow the disabling of metrics collection


jsvd:
right now it's not possible to turn off metrics collection, as the value of `metric.collect` configuration property must be always `true`.

as a preparation for future logstash features like multi pipelining, there has to be a way to turn off metrics at different levels:

- don't collect metrics at all
- collect only logstash-core metrics (like jvm, hot_threads, list of installed plugins)
- collect core and pipeline metrics (events in/out, plugin stats, hot_threads, etc)

jsvd:
closed by #6097

PR #6097, opened 2016-10-20
allow the disabling of metric collection

jsvd:
previously `metric.collect` was always necessarily true, it wasn't possible to disable metric collection.
Now, it's possible to set  `metric.collect: false` in the settings.yml file and no metrics are computed/exposed in the web api

implements https://github.com/elastic/logstash/issues/6079


suyograo:
Any reason why this setting needs to be exposed to the user? When would a user set `metric.collect: false`? I completely get why we would do this internally for the metrics pipeline, but doesn't feel necessary to be exposed to end-users.

jsvd:
With this PR the setting can indeed be changed by the user but it's something we are encouraging, as we didn't add it to the logstash.yml file not added documentation about it.

However since it's a new feature, with some (albeit marginal) impact to performance and could have bugs (like anything else), it would be nice to be able to turn it off as a last resort.

ph:
LGTM, I am okay to have this flag available but not documented.
With the change with LIR and the pipeline, we might want to rethink how we pass specific pipeline options.

suyograo:
LGTM. Let's not document it for now and we can use it as a last resort if there are severe perf degradation with metrics.

PR #6100, opened 2016-10-20
Update Breaking Changes to provide extra details

pickypg:
This adds extra details to the Breaking Changes document so that users can hopefully better understand what is necessary to change in order to get around breaking changes.

/cc @acchen97 @suyograo @dedemorton

### DIFF
--- a/docs/static/breaking-changes.asciidoc
+++ b/docs/static/breaking-changes.asciidoc
@@ -1,48 +1,17 @@
 [[breaking-changes]]
 == Breaking changes
 
-This section discusses the changes that you need to be aware of when migrating your application to Logstash {version} from the previous major release of Logstash (2.x).
-
-[float]
-=== When to Upgrade
-
-Fresh installations can and should start with the same version across the Elastic Stack.
-
-Elasticsearch 5.0 does not require Logstash {version}. An Elasticsearch 5.0 cluster will happily receive data from a
-Logstash 2.x instance via the default HTTP communication layer. This provides some flexibility to decide when to upgrade
-Logstash relative to an Elasticsearch upgrade. It may or may not be convenient for you to upgrade them together, and it is
-not required to be done at the same time as long as Elasticsearch is upgraded first.
-
-You should upgrade in a timely manner to get the performance improvements that come with Logstash {version}, but do so in
-the way that makes the most sense for your environment.
-
-[float]
-=== When Not to Upgrade
-
-If any Logstash plugin that you require is not compatible with Logstash {version}, then you should wait until it is ready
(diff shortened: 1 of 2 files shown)


docs/static/breaking-changes.asciidoc
```diff
@@ -1,93 +1,256 @@
 [[breaking-changes]]
 == Breaking changes
 
-This section discusses the changes that you need to be aware of when migrating your application to Logstash {version}.
+This section discusses the changes that you need to be aware of when migrating your application to Logstash {version} from the previous major release of Logstash (2.x).
+
+[float]
+=== When to Upgrade
```
dedemorton: Some of this information looks like it belongs in the upgrade doc, not the breaking changes topic. @acchen97 is in the process of scoping out what needs to change in the upgrade doc. The general upgrade stuff should be added to the new content that we add under this topic: https://www.elastic.co/guide/en/logstash/5.0/upgrading-logstash.html

docs/static/breaking-changes.asciidoc
```diff
@@ -1,93 +1,256 @@
...
+
+Fresh installations can and should start with the same version across the Elastic Stack.
+
+Elasticsearch 5.0 does not require Logstash {version}. An Elasticsearch 5.0 cluster will happily receive data from a
```
dedemorton: Be careful when using the version variable because this value will change with every release. Usually when you describe breaking changes, they are specific to a certain release. You don't want the version number to change when the release version changes. This comment applies to other places where you've used `{version}` in this document.

docs/static/breaking-changes.asciidoc
acchen97: @pickypg has been driving this and has done a great job with drafting up the upgrade docs. It almost seems like a drop in replacement for our upgrade logstash section. I do agree with @dedemorton that this stuff fits more into the upgrade section than here (though we should certainly link to the breaking changes).

docs/static/breaking-changes.asciidoc
pickypg: Yeah, I kind of wonder if I should replace `{version}` with literally `5.0` (or even more generally `5.x`?) in the breaking changes doc, since that is really a constant.

suyograo:
@pickypg heads up -- I merged some stuff to breaking changes doc, so this will need a rebase

pickypg:
I'm going to let the build pass before merging, but otherwise I've rebased and fixed based on comments.

suyograo:
@pickypg all good to merge. `continuous-integration/travis-ci/pr` passed.

PR #6104, opened 2016-10-22
Update old references to verbose or debug modes

### LINKED ISSUE #6095: Log level options when starting Logstash
- Version: 5.0.0-rc1
- Operating System: Centos 7.2
- Config File (if you have sensitive info, please remove it): N/A
- Sample Data: N/A
- Steps to Reproduce: `./bin/logstash --help`

If you executes `./bin/logstash --help` it shows that `--verbose`, `--debug` and `--quiet` are deprecated and you should use `--log.level=verbose` and so on.

The problem is that if you start Logstash with one of them as a parameter, it exits and shows this error:

```ERROR: option '--log.level': Invalid value "verbose". Options are: ["fatal", "error", "warn", "debug", "info", "trace"]```

How can I start the process with verbose enabled?

Thanks,
Best regards

jordansissel:
Replace with --log-level=verbose and --log-level=debug

Fixes #6095

### DIFF
--- a/logstash-core/locales/en.yml
+++ b/logstash-core/locales/en.yml
@@ -215,7 +215,7 @@ en:
           Specifying once will show 'informational'
           logs. Specifying twice will show 'debug'
           logs. This flag is deprecated. You should use
-          --verbose or --debug instead.
+          --log-level=info or --log-level=debug instead.
         version: |+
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

PR #6109, opened 2016-10-24
Universal Plugins and Registry of plugins

ph:
This PR changes two things:
- Allow plugins author to decide how their plugins are structured
- Allow a new kind of plugin that allows plugins author to add hooks into logstash core.

---

The way we load plugins in logstash core is tied to the file directory of the plugins and
all the plugins files are lazy loaded when they are required. Nothing prevent you today to
have multiple plugins in the same repo, if they respect the following structure.

```
logstash/(codec|input|output|filter)s/plugin_name
```

This PR allow the plugin to have an arbitrary file structure and be able to register manually to logstash.
It uses a conventions file for the registration; the author needs to create a `logstash_registry.rb` in any of the `LOAD_PATH` of the plugin.
Logstash on boot will read and evaluate that file. The author can now use the following syntax:

``` ruby
require "logstash/out_of_this_world"

LogStash::PluginRegistry.add(:input, "past", LogStash::OutOfThisWorld::Past)
LogStash::PluginRegistry.add(:input, "future", LogStash::OutOfThisWorld::Future)
LogStash::PluginRegistry.add(:input, "present", LogStash::OutOfThisWorld::Present)
```

Add a new kind of plugin called `Universal` that let you register hooks to the core logstash
I've create new plugin that only add logging to the core:  https://github.com/ph/logstash-pack-megateuf

I've pushed the gem to rubygems so you can already install it using `bin/logstash-plugin`

**The documentation generator will need to be adapted to support that new feature.**


ph:
I've marked this PR as WIP, because I am not sure about the naming should we call theses new plugins, :core, :system or like in this PR :universal ?

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -1,86 +1,243 @@
...
+        @instance.additionals_settings(settings)
+      end
+    end
+
+    attr_reader :hooks
 
     def initialize
-      @registry = {}
-      @logger = self.logger
+     @registry = {}
```
jsvd: indentation

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -1,86 +1,243 @@
...
+    # specifications database, most of out code interact directly with really low level
+    # components of bundler/rubygems we need to encapsulate that and this is a start.
+    class GemRegistry
+      LOGSTASH_METADATA_KEY = "logstash_plugin"
 
-      def initialize(type, name)
-        @type  = type
-        @name  = name
+      class << self
+        def available_libraries
```
jsvd: maybe `installed_gems` or `available_gems`?

logstash-core/lib/logstash/event_dispatcher.rb
```diff
@@ -0,0 +1,44 @@
...
+    #
+    # See note on add_listener, this method shouldn't be called really often.
+    def remove_listener(listener)
+      @listeners.remove(listener)
+    end
+
+    def fire(method_name, *arguments)
+      @listeners.each do |listener|
+        if listener.respond_to?(method_name)
+          if arguments.size > 0
```
ph: oh right!

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -1,86 +1,243 @@
...
+            require path
+          rescue LoadError
+            # Plugin might be already defined in the current scope
+            # This scenario often happen in test when we write an adhoc class
+          end
+
+          klass = namespace_lookup(type, plugin_name)
+          plugin = lazy_add(type, plugin_name, klass)
+        rescue => e
+          logger.warn("Problems loading a plugin with",
```
ph: Not all the type, if the namespace_lookup fails to find the name of the plugin class it will catched by this rescue.

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -1,86 +1,243 @@
...
+    end
+
+    def lazy_add(type, name, klass)
+      logger.error("On demand adding plugin to the registry", :name => name, :type => type, :class => klass)
+      add_plugin(type, name, klass)
+    end
+
+    def add(type, name, klass)
+      logger.error("Adding plugin to the registry", :name => name, :type => type, :class => klass)
+      add_plugin(type, name, klass)
```
ph: Yes, I think for debugging purpuse and since we support both legacy declaration of plugins and this new hook style registration It make sense to know when this plugin get added to the registry.

logstash-core/lib/logstash/runner.rb
```diff
@@ -201,6 +202,9 @@ def execute
       logger.warn("--config.debug was specified, but log.level was not set to \'debug\'! No config info will be logged.")
     end
 
+    LogStash::PluginRegistry.setup!
+    @settings.validate_all
```
ph: Agree!

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -392,7 +392,7 @@
     # We need to create theses dummy classes to know how many
     # events where actually generated by the pipeline and successfully send to the output.
     # Theses values are compared with what we store in the metric store.
-    class DummyOutput2 < DummyOutput; end
+    class DummyOutput2 < LogStash::Outputs::DummyOutput; end
```
ph: I would answer it depends.

When the first version of the registry was implemented, it stopped calling the `Plugin.namespace_lookup`, I think it was a regression in the way the legacy plugins where working. The code was kept in the project, if you look in the master branch you will that this method exist.

This PR reintroduce the call to that so it inforce the need to have the correctly defined plugin in the right namespace, which is how legacy style has been working for a lot of version. If the user want to define theirs own structure they will use the add method.

Most of the time in the code when we define a custom or one shot plugin we also make sure that the `Plugin.lookup` return the right class, except in two case in the current core specs.

This is why I've changed the name of the default mocks to make sure it does work and reflect how the namespace lookup actually works.

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -1,86 +1,243 @@
...
+      LogStash::PluginRegistry.lookup(type, name) do |plugin_klass, plugin_name|
+        is_a_plugin?(plugin_klass, plugin_name)
+      end
+    rescue LoadError, NameError => e
+      logger.debug("Problems loading the plugin with", :type => type, :name => name)
+      raise(LogStash::PluginLoadingError, I18n.t("logstash.pipeline.plugin-loading-error", :type => type, :name => name, :error => e.to_s))
+    end
+
+    def lazy_add(type, name, klass)
+      logger.error("On demand adding plugin to the registry", :name => name, :type => type, :class => klass)
```
ph: They should be debug.

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -1,86 +1,243 @@
...
 
-    def registered?(path)
-      @registry.has_key?(path)
+    def key_for(type, plugin_name)
+      "#{type}-#{plugin_name}"
     end
+  end end
 
-  end
+  PluginRegistry = Plugins::Registry.new
```
ph: Well `PluginRegistry` is a constants, it just not Capitalized.


*I double checked that*

ph:
I've pushed the requested changes @jsvd

ph:
@jsvd Yeah, naming is hard :)

logstash-core/lib/logstash/plugins/registry.rb
andrewvc: I would capitalize this for clarity.

andrewvc:
I proposed `universal`. I'm not married to it, but I like it because it implies it can contain any number of plugins performing any number of functions.

Core and system seem to imply that input/output would not be included. Ideally all plugins would eventually move to universal IMHO.

logstash-core/lib/logstash/plugins/registry.rb
ph: Since 2 people want it I will change it, I do not have a strong opinion.

ph:
@andrewvc I am ok with universal, this is was I used in this code after our conversation.

ph:
I've made a push to this PR to change `PluginRegistry` to `PLUGIN_REGISTRY` So its more obvious its a constant.

jsvd:
@ph I'm attempting to use the additional_settings feature, and while it successfully registers the settings, I can't define them in the logstash.yml file, since the `from_yaml` call early on will execute before the register happens.

my current strategy to work around this is using a different yaml setting file which is loaded during `after_initialize`.

ph:
My solution for this is to introduce the concept of `transient_settings`, which will mean the following. When we read the YAML files and we dont find the key in the settings hash we push the key and the value to the `transient_settings` when we do `validate_all` which happen after all the plugins are loaded we try to merge the transient settings. At that moment if the settings doesn't exist we fails loading the settings.yml file.

The problem is we have some sort of a chicken and egg problem.

ph:
@jsvd I've pushed a fix for the settings validation, you should be able to declare plugin settings in the logstash.yml file.

jsvd:
confirmed it works ok, yay!

jsvd:
LGTM

jordansissel:
:heart:

Issue #6108, opened 2016-10-24
Failling test for `LogStash::Setting::WritableDirectory#set`


ph:
```
 1) LogStash::Setting::WritableDirectory#set and #validate_value when the path exists but is a socket behaves like failure should fail
     Failure/Error: let(:socket) { UNIXServer.new(path) }
     ArgumentError:
       too long unix socket path (max: 103bytes)
     Shared Example Group: "failure" called from ./logstash-core/spec/logstash/settings/writable_directory_spec.rb:84

ph:
Some reference http://serverfault.com/questions/641347/check-if-a-path-exceeds-maximum-for-unix-domain-socket

ph:
Discovered in https://github.com/elastic/logstash/pull/6109

PR #6110, opened 2016-10-24
Fix test for LogStash::Setting::WritableDirectory#set

ph:
path for unix domain socket have a hard limit of 104 chars on macos and 108 on
linux, using `Stud::Temporary.pathname` to generate the path will exceed
that limit. So we will use a custom generation of using the tmpdir +
Time.now.to_f which should be good enough for this test.

Fixes: #6108

### DIFF
--- a/logstash-core/spec/logstash/settings/writable_directory_spec.rb
+++ b/logstash-core/spec/logstash/settings/writable_directory_spec.rb
@@ -1,12 +1,17 @@
 # encoding: utf-8
 require "spec_helper"
 require "logstash/settings"
-require "stud/temporary"
+require "tmpdir"
 require "socket" # for UNIXSocket
+
+
+
 
 describe LogStash::Setting::WritableDirectory do
   let(:mode_rx) { 0555 }
-  let(:parent) { Stud::Temporary.pathname }
+  # linux is 108, Macos is 104, so use a safe value
+  # Stud::Temporary.pathname, will exceed that size without adding anything
+  let(:parent) { File.join(Dir.tmpdir, Time.now.to_f.to_s) }
   let(:path) { File.join(parent, "fancy") }
 
   before { Dir.mkdir(parent) }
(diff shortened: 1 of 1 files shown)


logstash-core/spec/logstash/settings/writable_directory_spec.rb
```diff
@@ -1,12 +1,17 @@
...
 
+
+
+
 describe LogStash::Setting::WritableDirectory do
   let(:mode_rx) { 0555 }
-  let(:parent) { Stud::Temporary.pathname }
+  # linux is 108, Macos is 104, so use a safe value
+  # Stud::Temporary.pathname, will exceed that size without adding anything
+  let(:parent) { File.join(Dir.tmpdir, Time.now.to_f.to_s) }
```
jsvd: should we perhaps fix this in stud? we could detect in the `.pathname` method that the filename is too big a retry with a shorter hex? Or even make the length of the hex an argument and lower it to 15 instead of 30?

logstash-core/spec/logstash/settings/writable_directory_spec.rb
```diff
@@ -1,12 +1,17 @@
 # encoding: utf-8
 require "spec_helper"
 require "logstash/settings"
-require "stud/temporary"
+require "tmpdir"
 require "socket" # for UNIXSocket
 
+
```
jsvd: unnecessary white space

logstash-core/spec/logstash/settings/writable_directory_spec.rb
ph: Well, stud is used all over the place and this problem is tied to only this specific use case.

So I think we should just make it explicit in that context and not add any magic or changes in stud's pathname method.

ph: Also, changing the generation of the temporary file in stud is a breaking changes.

ph:
@jordansissel I saw that you have desactivated the test in the PQ branch.

https://github.com/elastic/logstash/pull/5958/commits/2cd728646c0a3a576960598e2da0afa977b40e3e

Is this fix appropriate intead of disabling it?

jsvd:
LGTM

jordansissel:
@jsvd the path length problem is specific to unix sockets, not to any other kind of path (as far as we know). I think hacking the test for now to use what @ph proposes is fine with me. LGTM

PR #6119, opened 2016-10-26
[docs] ES input was not removed

suyograo:
I accidentally added logstash-input-elasticsearch in this list because
in my RC testing, I had manually removed it using bin/logstash-plugin remove


talevy:
LGTM

PR #6127, opened 2016-10-26
Path correction

n3gster:


suyograo:
Fixed in https://github.com/elastic/logstash/pull/6140

PR #6128, opened 2016-10-26
Slowlog For Filters in Logstash

talevy:
Ever wish you can find out what is taking a long time in your Logstash Pipeline?

With the new addition of the [Monitoring API](https://www.elastic.co/guide/en/logstash/5.0/monitoring.html), you can introspect Logstash at the thread level, but there is no easy way to see what actions are taking long. Slowlog in Logstash is the first step in catching some important plugin actions that are slowing the pipeline down.

For demonstration, here is a sample pipeline with a test filter that will `sleep` for varying durations that, at times, will trigger a slowlog event to be logged.

```
input {
  generator { count => 1 }
}

filter {
  sleep {
    time => "2"
  }
}
output {
  null {}
}
```

Now you can configure a slowlog to log to `/path/to/logs/logstash-slowlog-plain-YYYY-MM-dd.log`

Slowlog is currently only configured on filters. If any filter action on an event exceeds a pre-defined threshold, the instance will be logged.
The thresholds are configured in `logstash.yml` as the following configuration keys:

```
slowlog.threshold.warn (default: -1)
slowlog.threshold.info (default: -1)
slowlog.threshold.debug (default: -1)
slowlog.threshold.trace (default: -1)
```

These options take in a time value similar to [Elasticsearch's Slowlog](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-slowlog.html).  By default, these values are set to `-1` to represent an infinite threshold where no slowlog will be invoked.

Here is a sample:

```
slowlog.threshold.warn: 2s
slowlog.threshold.info 1s
slowlog.threshold.debug 500ms
slowlog.threshold.trace 100ms
```
### Contents of Slowlog

Assuming the above pipeline and the example configuration were run in tandem, we may find our slowlog filled with the following content in the case of a filter taking a little over 2 seconds. The log line will include the offending event and the time it took to process within the filter

```
...
[2016-10-27T13:48:19,308][WARN ][slowlog.logstash.filters.sleep] event processing time {:took_in_nanos=>1000614130, :event=>"{\"sequence\":1,\"@timestamp\":\"2016-10-27T20:48:18.150Z\",\"@version\":\"1\",\"host\":\"elasticbox-2.local\",\"mes
sage\":\"Hello world!\"}"}
```
### Old efforts forgotten?

There were some previous efforts to introduce this feature to Logstash. This is picking up that torch and moving forward with the task.

for reference to older discussions, please reference:

| Feature Issue | https://github.com/elastic/logstash/issues/5733 |
| --- | --- |
| The PR | https://github.com/elastic/logstash/pull/5886 |

I hope that many of the discussion from there will continue here. One major difference between
that PR and this is that this only sets up the slowlog functionality itself with no consideration of the API for tracking "top-K" style metrics of slow operations. That can continue in follow-up commits.

### DIFF
--- a/logstash-core/lib/logstash/logging/logger.rb
+++ b/logstash-core/lib/logstash/logging/logger.rb
@@ -3,11 +3,12 @@ require "uri"
 
 module LogStash
   module Logging
+    java_import org.apache.logging.log4j.Level
+    java_import org.apache.logging.log4j.LogManager
+    java_import org.apache.logging.log4j.core.config.Configurator
+    java_import org.apache.logging.log4j.core.config.DefaultConfiguration
+
     class Logger
-      java_import org.apache.logging.log4j.Level
-      java_import org.apache.logging.log4j.LogManager
-      java_import org.apache.logging.log4j.core.config.Configurator
-      java_import org.apache.logging.log4j.core.config.DefaultConfiguration
       @@config_mutex = Mutex.new
       @@logging_context = nil
 
@@ -88,6 +89,38 @@ module LogStash
 
       def self.get_logging_context
         return @@logging_context
+      end
+    end
+
+    class SlowLogger
+      def initialize(name, warn_threshold, info_threshold, debug_threshold, trace_threshold)
+        slowlog_name = ["slowlog", name].join('.')
+        @slowlogger = LogManager.getLogger(slowlog_name)
+        @warn_threshold = warn_threshold
+        @info_threshold = info_threshold
+        @debug_threshold = debug_threshold
+        @trace_threshold = trace_threshold
+      end
+
+      def as_data(plugin_params, event, took_in_nanos)
+        {
+          :plugin_params => plugin_params,
+          :took_in_nanos => took_in_nanos,
+          :took_in_millis => took_in_nanos / 1000,
+          :event => event.to_json
+        }
+      end
+
+      def on_event(message, plugin_params, event, took_in_nanos)
+        if @warn_threshold >= 0 and took_in_nanos > @warn_threshold
+          @slowlogger.warn(message, as_data(plugin_params, event, took_in_nanos))
+        elsif @info_threshold >= 0 and took_in_nanos > @info_threshold
+          @slowlogger.info(message, as_data(plugin_params, event, took_in_nanos))
+        elsif @debug_threshold >= 0 and took_in_nanos > @debug_threshold
+          @slowlogger.debug(message, as_data(plugin_params, event, took_in_nanos))
+        elsif @trace_threshold >= 0 and took_in_nanos > @trace_threshold
+          @slowlogger.trace(message, as_data(plugin_params, event, took_in_nanos))
+        end
       end
     end
   end
--- a/logstash-core/lib/logstash/util/time_value.rb
+++ b/logstash-core/lib/logstash/util/time_value.rb
@@ -1,0 +1,70 @@
+module LogStash
+  module Util
+    class TimeValue
+      def initialize(duration, time_unit)
+        @duration = duration
+        @time_unit = time_unit
+      end
+
+      def self.from_value(value)
+        if value.is_a?(TimeValue)
+          TimeValue.new(value.duration, value.time_unit)
+        elsif value.is_a?(::String)
+          normalized = value.downcase.strip
+          if normalized.end_with?("nanos")
+            TimeValue.new(parse(normalized, 5), :nanosecond)
+          elsif normalized.end_with?("micros")
+            TimeValue.new(parse(normalized, 6), :microsecond)
+          elsif normalized.end_with?("ms")
+            TimeValue.new(parse(normalized, 2), :millisecond)
+          elsif normalized.end_with?("s")
+            TimeValue.new(parse(normalized, 1), :second)
+          elsif normalized.end_with?("m")
+            TimeValue.new(parse(normalized, 1), :minute)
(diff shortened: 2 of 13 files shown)


suyograo:
Would it make sense to add another level configuration in the slowlog? As in, what if I want to see slowlogs for a particular stage?

Example:
```
filters.slowlog.threshold.warn
inputs.slowlog.threshold.warn
outputs.
```

talevy:
@suyograo Even after reading the previous discussions, I still feel that the meaning of "slow" for things like inputs and outputs are still not well defined.

I am inclined to set the scope of this initial PR to just be slowlog for filters so that we can discuss the expansion separately. There is a lot of setup going on here that is re-usable for other contexts, and inputs/outputs enabling doesn't block this configuration.

Currently, All plugin types have a slowlog available to them, but it is only the filter that has a well defined unit of work that is discrete enough to capture a notion of performance. So, adding them later isn't major.

suyograo:
@talevy fair enough, we can iterate on this based on user feedback

talevy:
One other context that may be worth tracking is queue behavior.

`queue.push.slowlog` and `queue.pop.slowlog`. These logs would give us insight into how long the pipeline is stalled maybe? not sure if the metrics api captures this.

For example, I may have a stalled pipeline and it is unclear if we are stalled on pushing items in from the inputs, or stalled on popping them out in outputs. If these operations had a notion of timeout callbacks, we could get logs to inform us of their status

talevy:
I added the discuss label right now because I feel that it is maybe not right to track slowness at the filter-plugin level, but at the pipeline's filter func. what do you think?

logstash-core/lib/logstash/logging/logger.rb
```diff
@@ -79,5 +80,35 @@ def self.get_logging_context
...
+      def initialize(name, warn_threshold, info_threshold, debug_threshold, trace_threshold)
+        slowlog_name = ["slowlog", name].join('.')
+        @slowlogger = LogManager.getLogger(slowlog_name)
+        @warn_threshold = warn_threshold
+        @info_threshold = info_threshold
+        @debug_threshold = debug_threshold
+        @trace_threshold = trace_threshold
+      end
+
+      def as_data(event, time_in_nanos)
```
suyograo: I think we need more context here. At least the filter ID and filter config. Imagine if there were 8 grok filters and you'd have to find the slowest. With the existing code, you'll know grok is slow, but which one?

logstash-core/lib/logstash/util/time_value.rb
```diff
@@ -0,0 +1,70 @@
+module LogStash
+  module Util
+    class TimeValue
```
suyograo: Nice!

qa/integration/specs/slowlog_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    @ls = @fixture.get_service("logstash")
+  }
+
+  after(:all) {
+    @fixture.teardown
+  }
+
+  before(:each) {
+    # backup the application settings file -- logstash.yml
+    FileUtils.cp(@ls.application_settings_file, "#{@ls.application_settings_file}.original")
```
suyograo: Don't need to do this in this particular test

qa/integration/specs/slowlog_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  after(:each) {
+    @ls.teardown
+    # restore the application settings file -- logstash.yml
+    FileUtils.mv("#{@ls.application_settings_file}.original", @ls.application_settings_file)
```
suyograo: Don't need to do this in this particular test

qa/integration/specs/slowlog_spec.rb
suyograo: Ah, my bad missed the edits to the settings file

suyograo:
@talevy overall, I like the simplicity of this approach. My only concern is about lack of context (esp for filters). That would make it pretty complete.

logstash-core/lib/logstash/logging/logger.rb
talevy: right, this is a general problem of context that we have in the pipeline, many of the changes coming in the LIR fix this.

But, this brings up my question above: https://github.com/elastic/logstash/pull/6128#issuecomment-256767708.
would it make more sense to have a slowlog capture the whole `filter_func` instead of watching time on a per plugin basis?

For now, I will add the ID, and took_in_millis.

qa/integration/specs/slowlog_spec.rb
talevy: that being said, it wouldn't bad at all to write these tests. I think our mission has been mostly accomplished. After one or more of these types of tests, I think a lot of code will be more generalized and re-usable. 💯

suyograo:
@talevy filter time is also being recorded here for monitoring API -- https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/filter_delegator.rb#L40 . Can we use this for slowlog instead of doing in two places? As you said, you'll get access to more context here..

talevy:
@suyograo thanks, was not familiar with that code. that would be a way more appropriate place for this

suyograo:
LGTM.

suyograo:
@talevy before merging please add docs for this.

suyograo:
Docs LGTM. this will need a rebase though

talevy:
thanks! rebased, waiting on build

PR #6138, opened 2016-10-27
Fixed documentation inconsistencies

vjestin:
fixed wrong binaries directory for RPM package in doc

When installed with the RPM package, logstash binaries are actually installed in /usr/share/logstash and it seems to be consistent with other Elastic products (tested with elasticsearch 5.0). Futhermore, 'Logstash Directory Layout' page is only talking about /usr/share/logstash.

fixed settings file name

Documentation refers to both 'settings.yml' and 'logstash.yml' files for new settings file. I set 'logstash.yml' everywhere because it's the file name which is used in RPM, APT and zip/tar.gz archives.


suyograo:
@vjestin thanks, can you please perform step 2 of https://github.com/elasticsearch/logstash/blob/master/CONTRIBUTING.md#contribution-steps

vjestin:
CLA signed.

PR #6145, opened 2016-10-28
Update logs documentation with recent changes

### LINKED ISSUE #6144: Settings doc seems to have a typing mistake for path.logs
- Version: 5.0
- Operating System: Any

This [document](https://www.elastic.co/guide/en/logstash/5.0/logstash-settings-file.html) mentions "path.log" however logstash.yml by default ships with "path.logs" which is the correct setting.

- Steps to Reproduce: use "path.log" instead "path.logs" in logstash.yml then start logstash and see the error.

suyograo:
Fixes #6144


ph:
LGTM

PR #6150, opened 2016-10-31
Fix 'Event#remove' for @metadata fields

### LINKED ISSUE #3538: `Event#remove` doesn't work with @metadata
@andrewvc found this issue when working on the http poller.

If you save any information in the `@metadata` field and try to rename the key with the mutate filter.
The filter will create the target field with a nil value and the data will remain with the `@metadata`
The mutate filter actually use the `#remove` method to remove and return the value of the key.
But the current implementation only give access to the `accessors` hash, which include all the `non-metadata` field

See:
https://github.com/elastic/logstash/blob/master/lib/logstash/event.rb#L203

prehor:
`Event#remove` method give access only to the event data hash, which does not include `@metadata` subtree. When you try to remove any field in `@metadata` subtree you will always get a `nil` value and the field remains in `@metadata` subtree.

Fixes #3538


untergeek:
You probably should have rebased your commit.  We simply cannot accept this pull request as-is.

Also, why do you need to remove a field from `@metadata`? When this functionality was added, the reason we omitted remove functionality was because it is expensive in terms of computation—it slows things down.  We figured that `@metadata` was never going to persist to _any_ output, except stdout, and then only if you explicitly told it to include `@metadata`, so why bother with removes?  The whole premise was so that it could be a workspace that just "went away" when you were done, regardless of what was in it.

Please explain your use case, and why you feel that delete needs to exist in `@metadata`.

untergeek:
Just taking a quick look through the PR, it's clear you based this off of a _very_ old pull from the master branch.  A lot has changed since then, as you can probably tell. 

You should probably start over and make your edits against the most recent version of master.

prehor:
I did not know what was the original intent for this behavior of #remove. 

I use @metadata extensively in the normalization of field names and values. Sometimes I need to delete a field in @metadata because its existence is status information that later on I test or process in other way (for example in processing fields created by grok with multiple patterns). 

And also bothers me that 'mutate {rename => {"[@metadata] [somefield]" => "[someotherfield]'}}' silently does nothing so my config files are full of comments like "WARNING: we cannot use rename for @metadata" :-(

untergeek:
Just do a mutate add_field to copy `[@metadata][somefield]` to `someotherfield`. 

I understand that you want `@metadata` to be more flexible. We can have that discussion, but the whole point was to prevent the slowdown inherent in deleting an object from the Event hash (it's really quite expensive). If you can live with the slowdown, you can do it in the regular event space and clean up what you need to clean up afterward as the performance would be the same.

prehor:
When I use add_field or replace instead of rename I have to use string interpolation, which is slow too. Processing temporary fields outside @metadata is dangerous because it can result to output them by accident and it does not satisfy me. Therefore, I think that removing field and the function rename in the mutate filter should also work for @metada. As a system solution I can see adding this functionality to the method #remove (same tests of field name are doing in the methods and #getField #setField too). As a result, the behavior of functions using #remove will be consistent and it will be less confusing for users (findings that mutate function rename does not rename fields in @metadata and that it is not my mistake but unexpected and undocumented feature of logstash cost me four hours of my life).

untergeek:
Again, it is by design that this is the case.  We will discuss internally whether we want to change this behavior.

Either way, this PR is unacceptable in its current form.  You will have to close this, and make your changes against the current master branch.

untergeek:
At the very least, you will need to squash your commits.

untergeek:
We'll have to do some investigation.  The new java event pipeline may (or may not) change the performance paradigm.

prehor:
Thanks very much. I am glad that you will be thinking about my proposal.  

Tomorrow I will create new PR based on recent master.

jsvd:
we can close this as removing fields from `@metadata` works:

```
/tmp/logstash-7.3.0 % bin/logstash -i irb
irb(main):001:0> e = LogStash::Event.new("[@metadata][name]" => "joao")
=> #<LogStash::Event:0x1332e410>
irb(main):002:0> e.remove("[@metadata][name]")
=> "joao"
irb(main):003:0> e.to_hash
=> {"@timestamp"=>2019-09-02T08:33:09.797Z, "@version"=>"1"}
irb(main):004:0> e.to_hash_with_metadata
=> {"@timestamp"=>2019-09-02T08:33:09.797Z, "@version"=>"1"}
irb(main):005:0> 

```

PR #6152, opened 2016-10-31
merge feature/java_persistence into 5.x

colinsurprenant:


ph:
I am taking a look at the fails build on travis since a lot of metric related test fails, with the same reason.

```
2) LogStash::Api::Modules::Node#hot threads Generic JSON testing GET /os should respond OK
     Failure/Error: do_request { get path }
     LogStash::Instrument::MetricStore::MetricNotFound:
       For path: config
```

This usually mean we try to fetch something from the metric store and the key doesn't exist yet, so my hyphothesis for now is something wrong when we setup the environment.

ph:
fun fact, no error on this branch locally.

ph:
Nothing related to a config in the spec.

```
/t/l/l/s/api git:pr/6152 ❯❯❯ ag 'config'                                                                                                                                                                                                                                                                                  ✱
lib/api/node_spec.rb
117:          "config_reload_automatic" => Boolean,
118:          "config_reload_interval" => Numeric

spec_helper.rb
35:  attr_reader :config_str, :agent, :pipeline_settings
38:    @config_str   = "input { generator {count => 100 } } output { dummyoutput {} }"
41:      "config.reload.automatic" => false,
47:      "config.string" => @config_str,
/t/l/l/s/api git:pr/6152 ❯❯❯
```

ph:
This is where the exception originate https://github.com/elastic/logstash/blob/b89d4d049da3f47bba650430357751f804a1c107/logstash-core/lib/logstash/instrument/metric_store.rb#L221

ph:
I think it might be a port issue, I've restarted the job on travis with this PR and it passes.

Too bad I havent saved the seed.

ph:
Actually It cannot be a port issue we use rack test and the webserver is a mock when running the test, we directly use the sinatra classes for tests.

I've started the job again and created this issue: https://github.com/elastic/logstash/issues/6153

colinsurprenant:
Thanks @ph for looking at this. It feels spec order related. I'll restart the build a few times see if we can reproduce.

Gemfile.jruby-1.9.lock
```diff
@@ -8,14 +8,16 @@ PATH
       filesize (= 0.0.4)
```
talevy: we should not be changing the lock file, no?

talevy: except maybe for `logstash-core-queue` relevant parts

colinsurprenant: why not? I mean, if plugins need to be updated across updates, then necessarily their dependencies will also be updated...

colinsurprenant: plus, our version constrains ***should*** be good to allow acceptable dependencies upgrades

talevy: hmm. OK

talevy:
LGTM, barring the follow-up to make unrelated tests green.

colinsurprenant:
We agreed to merge and followup on the intermittent failing tests. In the last build the only failure is the Kafka integration which is know to be flaky.

PR #6165, opened 2016-11-01
-e and -f should merge even without valid conf file

### LINKED ISSUE #6164: Logstash installed via rpm and apt packages cannot be manually started with `-e` flag
Logstash installed via rpm and apt packages cannot be manually started with `-e` flag. `-e` flag's behavior is to merge config string supplied via command line to config file supplied via `-f`. Now, in packages, `logstash.yml` is implicitly setting `path.config` to `/etc/logstash/conf.d/*`. This means whenever a user does 

```
bin/logstash -e 'foo'
```

it is actually merging contents of directory `/etc/logstash.conf.d` 

For a fresh installation, no config files are found and LS bails with an error.

A combination of all these issues causes a lot of confusion for first users and users experienced with pre-5.0 behavior where there was no `logstash.yml`

Related: https://github.com/elastic/logstash/issues/6062
https://discuss.elastic.co/t/logstash-5-not-running/64449

Also, our documentation for getting started instructs to use `e`, so this is needs to be fixed.

suyograo:
Previously if both -e and -f was specified, LS required that
-f still have valid config file(s) before merging. This fixes it
to either have one of -f or -e provided

Fixes #6164

### DIFF
--- a/logstash-core/lib/logstash/config/loader.rb
+++ b/logstash-core/lib/logstash/config/loader.rb
@@ -12,7 +12,18 @@ module LogStash; module Config; class Loader
       # Append the config string.
       # This allows users to provide both -f and -e flags. The combination
       # is rare, but useful for debugging.
-      config_string = config_string + load_config(config_path)
+      loaded_config = load_config(config_path)
+      if loaded_config.empty? && config_string.empty?
+        # If loaded config from `-f` is empty *and* if config string is empty we raise an error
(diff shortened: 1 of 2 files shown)


jordansissel:
idea and code lgtm; I haven't tested it though.

suyograo:
Merged in https://github.com/elastic/logstash/commit/ae60c7dfd44d32892f526c853dab1bde5705e370

PR #6168, opened 2016-11-02
Feature: A way to install/remove a plugin pack

ph:
This PR add a new installation options for plugins, first, let's get some definition so are all on the same page.

#### What is a pack?

A pack in this context is a *bundle* of plugins that can be distributed outside of rubygems; it is similar to what ES and kibana are doing, and
the user interface is modeled after them. See https://www.elastic.co/downloads/x-pack

**Do not mix it with the `bin/logstash-plugin pack/unpack` command.**

##### A pack need to follow a few guidelines:

- it contains one or more plugins that need to be installed
- it is self-contains with the gems and the needed jars
- it is distributed as a zip file
- the file structure needs to follow some rules.

##### A pack can be installed from multiple ways

- As a reserved name name on elastic.co download http server
    - `bin/plugin install logstash-mypack` will check on the download server if a pack for the current specific logstash version exist and it will be downloaded, if it doesn't exist we fallback on rubygems.
    - The file on the server will follow this convention `logstash-mypack-{LOGSTASH_VERSION}.zip`

- As a fully qualified url
    - `bin/plugin install http://test.abc/logstash-mypack.zip`, if it exists it will be downloaded and installed if it does not we raise an error.

- As a local file
    - `bin/plugin install file:///tmp/logstash-mypack.zip`, if it exists it will be installed

##### The structure of the pack

For the installer to work the structure of the pack need to followed, the pack itself doesn't have any metadata it just relies on the structure to do the right thing.
The name of the pack or the archive name is only used to describe the purpose of the pack, i.e., `logstash-monitoring` or `logstash-alerting`, the plugins contains in the pack can have their own name.

Lets assume that we have a pack name `logstash-specialpack` that contains a plugin named `logstash-output-secret`, the structure would look to something like this:

```
.
├── dependencies
│   ├── addressable-2.4.0.gem
│   ├── cabin-0.9.0.gem
│   ├── ffi-1.9.14-java.gem
│   ├── gemoji-1.5.0.gem
│   ├── launchy-2.4.3-java.gem
│   ├── logstash-output-elasticsearch-5.2.0-java.gem
│   ├── manticore-0.6.0-java.gem
│   ├── spoon-0.0.6.gem
│   └── stud-0.0.22.gem
└── logstash-output-secret-0.1.0.gem
```

The gems in the root of the folder will be the gems that need to be added to the logstash *Gemfile*; the dependencies directory contains the dependencies that the gems requires.
When we develop plugins that will be shipped as a pack, we need to pay special attention of how we define the constraints in the gem.

- External dependencies need to be flexible
- Internal dependencies on other plugins need to be flexible
- Core dependencies can be strict if you want a pack to work on a specific version of the core.
- Java dependencies need to be included in the gem. (Jar-dependencies hooks aren't triggered since they require internet access)

The `pack` are build using a tool I've developed called `paquet`, French word for a package, it takes care of reading the gemspec and whitelisting or blacklisting gems needed for the pack.

You can see paquet at https://github.com/ph/paquet, and an example of a pack using it at https://github.com/ph/logstash-output-secret, the most interesting part is the [Rakefile](https://github.com/ph/logstash-output-secret/blob/master/Rakefile) and the [bash script](https://github.com/ph/logstash-output-secret/blob/master/bundle.sh)

#### How a pack is installed in logstash?

Installing a pack diverges from installing a gems or normal plugins, if you look at the plugin manager's install command you see that we take a completely different path.
We do this for two reasons, its faster and we have better control on the process for the offline scenario. Installing a reserved pack, an URL pack or a local pack will go through the same **local** install path.

Lets look at the flow for installing an official pack.

- Does the pack exist on elastic's server? (HEAD request on a specific URL)
- The pack is downloaded into a temporary folder
- The pack gets extracted
- A special folder gets created as a Rubygems' Indexed format
- This folder is added to the available sources
- The definition (lock file graph) is loaded in memory
- The plugin(s) get added directly to the gemfile
- Their dependencies get resolved and added to the graph
- The graph is valid if it can accept theses gems with the local graph (no remote calls are needed)
- If added gems respect the currently installed gems
- The Gemfile and the lockfile get saved to disk
- The gems get extracted to *GEM_HOME* in the bundle directory
- The gems specification get saved in bundle specification folder
- Temporary files get deleted
- The gems now show up in logstash `bin/logstash-plugin list --verbose.'

#### Offline uninstall

Since a pack can now be installed successfully offline the remove command also need to be changed to support an offline workflow.

- Remove use the same local graph stategy.
- We now display information when trying to remove a gem that other gem depends on it.
- We have rats test for theses cases. 

#### TODO:

- Add tests for conflict; I've been having a hard time to add unit test on conflicting gem.
- We use a side effect of `Gem::Indexer` to make the gem available, I've tried really hard to manually supply the source to the graph without luck for now.

####
- ~~Add a post install message~~ Added in the current PR, the feature was short.
- ~~Allow uninstall/remove to be run offline~~


ph:
I have added the concept of `PluginManager.ui` which stand for user interaction, its like a logger but in the same time its not. its a way for us to give more feedback to the user.

ph:
@suyograo @jsvd ready for a first run of review.

ph:
I've added the offline feature to this PR

ph:
I will look at the 2 failures.

Gemfile
```diff
@@ -6,10 +6,11 @@ gem "logstash-core", :path => "./logstash-core"
 gem "logstash-core-queue-jruby", :path => "./logstash-core-queue-jruby"
 gem "logstash-core-event-java", :path => "./logstash-core-event-java"
 gem "logstash-core-plugin-api", :path => "./logstash-core-plugin-api"
+gem "ruby-progressbar"
```
jsvd: let's add a pessimistic version constraint to these gems, specially now with more gems coming out with ruby > 2 compat

lib/pluginmanager/gem_installer.rb
```diff
@@ -0,0 +1,78 @@
...
+    def display_post_install_message
+      PluginManager.ui.info(spec.post_install_message) if display_post_install_message?
+    end
+
+    def display_post_install_message?
+      @display_post_install_message && !spec.post_install_message.nil?
+    end
+
+    def create_destination_folders
+      FileUtils.mkdir_p(gem_home) unless Dir.exist?(gem_home)
```
jsvd: I don't think we need to check for the existence here:

```ruby
jruby-1.7.25 :002 > FileUtils.mkdir_p("/bin")
 => ["/bin"]
```

lib/pluginmanager/gemfile.rb
```diff
@@ -15,21 +15,24 @@ def initialize(io)
       @gemset = nil
     end
 
-    def load
+    def load(with_backup = true)
```
jsvd: I don't see this being used anywhere, but I could be wrong

lib/pluginmanager/gem_installer.rb
```diff
@@ -0,0 +1,78 @@
...
+  # - Generate the specifications
+  # - Copy the data in the right folders
+  class GemInstaller
+    GEM_HOME = Pathname.new(::File.join(LogStash::Environment::BUNDLE_DIR, "jruby", "1.9"))
+    SPECIFICATIONS_DIR = "specifications"
+    GEMS_DIR = "gems"
+
+    attr_reader :gem_home
+
+    def initialize(gem_file, display_post_install_message = false, gem_home = GEM_HOME)
```
jsvd: also, I don't see a situation where gem_home is not `GEM_HOME`
making this constant would simplify some of the path handling below

lib/pluginmanager/xpack/pack.rb
```diff
@@ -0,0 +1,82 @@
...
+require "pluginmanager/errors"
+
+module LogStash module PluginManager module Xpack
+  # A object that represent the directory structure
+  # related to the actual gems in the extracted package.
+  #
+  # Example of a valid structure, where `logstash-output-secret` is the actual
+  # plugin to be installed.
+  #.
+  # ├── dependencies
```
jsvd: from what I understand, xpacks will contain a directory called logstash with this structure inside of them, no?

lib/pluginmanager/install_strategy/elastic_xpack.rb
```diff
@@ -0,0 +1,41 @@
...
+# In the context of the plugin manager no dependencies are currently loaded.
+require_relative "../../../logstash-core/lib/logstash/version"
+require "pluginmanager/xpack/remote_installer"
+require "pluginmanager/utils/http_client"
+require "pluginmanager/ui"
+require "net/http"
+require "uri"
+
+module LogStash module PluginManager module InstallStrategy
+  class ElasticXpack
```
jsvd: looking at this functionality, I think we could remove this notion of xpack and elasticxpack. all this is is another strategy that takes a name and tries to download it from a "pack repository".

So `InstallStrategy` is essentially a `PackFetchStrategy`:
* `RepositoryStrategy` look for the pack in ELASTIC_XPACK_BASE_URI to fetch the pack
* `UriStrategy` directly use the URI to fetch the pack

And these can yield either a `LocalInstaller` or `RemoteInstaller` strategy. Even here we could remove the notion of XPack and just called it Pack. thoughts?

jsvd:
@ph added first round of review.
TL;DR the most important comment is likely https://github.com/elastic/logstash/pull/6168#discussion_r87861777

lib/pluginmanager/gem_installer.rb
ph: > do we ever want to not show the post install message?
  
Yes we do, we only really want the `post_install_message` for plugins that will be added to the gemfile (our gems). But all gems are free to have their own message but it doesn't mean we want to display the output to the users.

lib/pluginmanager/xpack/pack.rb
suyograo: correct. 

x-pack
   |
   -----elasticsearch
   -----kibana
   -----logstash
           |
            ----monitoring

ph: It will be called the name of the pack, so for a monitoring pack you will get something like this.

logstash-monitoring.zip
 - logstash-monitoring/
      - dependencies/
         - mydependecyXXXX.gem
      - logstash-monitoringXXX.gem

lib/pluginmanager/install_strategy/elastic_xpack.rb
```diff
@@ -0,0 +1,41 @@
...
+        if Utils::HttpClient.remote_file_exist?(uri)
+          PluginManager.ui.debug("Found package at: #{uri}")
+          LogStash::PluginManager::Xpack::RemoteInstaller.new(uri)
+        else
+          PluginManager.ui.debug("Package not found at: #{uri}")
+        end
+      rescue SocketError => e
+        # This probably means there is a firewall in place of the proxy is not correctly configured.
+        # So lets skip this strategy but log a meaningful errors.
+        PluginManager.ui.debug("SocketError, skipping Elastic Xpack, exception: #{e}")
```
ph: We don't log whole exception object, since we have string interpolation at hand it will call e.message in this context

lib/pluginmanager/install_strategy_factory.rb
```diff
@@ -0,0 +1,23 @@
...
+  class InstallStrategyFactory
+    AVAILABLES_STRATEGIES = [
+      LogStash::PluginManager::InstallStrategy::UriXpack,
+      LogStash::PluginManager::InstallStrategy::ElasticXpack,
+    ]
+
+    def self.create(plugins_arg)
+      AVAILABLES_STRATEGIES.each do |strategy|
+        if klass = strategy.match?(plugins_arg.first)
+          return klass
```
ph: +1 to change the name here.

lib/pluginmanager/xpack/local_installer.rb
```diff
@@ -0,0 +1,110 @@
...
+      raise PluginManager::FileNotFoundError, "Cant file local file #{local_file}" unless ::File.exist?(local_file)
+      raise PluginManager::InvalidPackError, "Invalid format, the pack must be in zip format" unless valid_format?(local_file)
+
+      PluginManager.ui.info("Installing file: #{local_file}")
+      uncompressed_path = uncompress(local_file)
+      PluginManager.ui.debug("Pack uncompressed to #{uncompressed_path}")
+      pack = LogStash::PluginManager::Xpack::Pack.new(uncompressed_path)
+      raise PluginManager::InvalidPackError, "The pack must contains at least one plugin" unless pack.valid?
+
+      local_source = move_to_local_source(uncompressed_path)
```
ph: Since the zip doesnt ship with a manifest file we use the directory structure to know what need to be added to the gemfile (root gem vs dependencies gems)

The `move_to_local_source` is needed because the Gem::Indexer requires that special directory structure to genere the index files.

lib/pluginmanager/xpack/pack.rb
```diff
@@ -0,0 +1,82 @@
...
+      @gems = Dir.glob(::File.join(source, "**", "*.gem")).collect { |gem| GemInformation.new(gem) }
+    end
+
+    def plugins
+      require "pry"
+      gems.select { |gem| !gem.dependency? }
+    end
+
+    def dependencies
+      gems.select { |gem| gem.dependency? }
```
ph: No it should not be an issue because this gem should not be presence in the dependencies folder.

lib/pluginmanager/xpack/remote_installer.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def execute
+      PluginManager.ui.info("Downloading file: #{remote_url}")
+      downloaded_file = Utils::Downloader.fetch(remote_url, feedback)
+      PluginManager.ui.debug("Downloaded package to: #{downloaded_file}")
+
+      Xpack::LocalInstaller.new(downloaded_file).execute
+    ensure
+      FileUtils.rm_rf(downloaded_file) if downloaded_file
```
ph: We should clear it in a rescue block there https://github.com/elastic/logstash/pull/6168/files#diff-d29ec5d7cc59e605fe75a1aace924041R66

lib/pluginmanager/install_strategy/elastic_xpack.rb
ph: If I understand correctly you want me to rename theses classes more into PackFetchStrategy, RepositoryStrategy, UriStrategy so they reflect more what they actually do?

lib/pluginmanager/bundler/logstash_injector.rb
```diff
@@ -0,0 +1,46 @@
...
+      gemfile = options.delete(:gemfile)
+      lockfile = options.delete(:lockfile)
+
+      bundler_format = Array(new_deps).collect { |plugin|  ::Bundler::Dependency.new(plugin.name, "=#{plugin.version}")}
+
+      injector = new(bundler_format)
+      injector.inject(gemfile, lockfile)
+    end
+
+    def inject(gemfile_path, lockfile_path)
```
ph: ✅

lib/pluginmanager/remove.rb
```diff
@@ -21,21 +21,12 @@ def execute
...
-      gemfile.save
-
-      puts("Removing #{plugin}")
-
-      # any errors will be logged to $stderr by invoke!
-      # output, exception = LogStash::Bundler.invoke!(:install => true, :clean => true)
-      output = LogStash::Bundler.invoke!(:install => true, :clean => true)
-
-      remove_unused_locally_installed_gems!
+    if !::Bundler::LogstashUninstall.uninstall!(plugin)
```
ph: ✅

ph:
I need to change the url of the xpack  Suyog Rao	downloads/logstash-plugins/x-pack/x-pack-5.0.1.zip

lib/pluginmanager/gemfile.rb
ph: This is used when we add a plugin, remove a plugin or update a plugin.
By default it will create a a backup of the gemfile. In the context of the pack we don't need that since that class won't write anything to disk.

Gemfile
ph: ✅

lib/pluginmanager/ui.rb
```diff
@@ -0,0 +1,24 @@
...
+  # require an explicite logger like log4j.
+  class Shell
+    def info(message)
+      puts message
+    end
+    alias_method :error, :info
+    alias_method :warn, :info
+
+    def debug(message)
+      puts message if ENV["debug"]
```
ph: ✅  I have added test for that,  :(

lib/pluginmanager/xpack/remote_installer.rb
ph: ✅

Gemfile
```diff
@@ -90,7 +91,7 @@ gem "logstash-input-twitter"
 gem "logstash-input-udp"
 gem "logstash-input-unix"
 gem "logstash-input-xmpp"
-gem "logstash-input-kafka"
+gem "logstash-input-kafka", "< 6.0.0"
```
jordansissel: ?

ph: Probably coming from a rebase? I will fix that.

lib/pluginmanager/install_strategy_factory.rb
ph: We cannot use `#find` in this context because find will return the first occurrence that matched the predicate, in this case `LogStash::PluginManager::InstallStrategy::UriXpack`, using each and breaking early seems OK to retrieve the installer.

ph:
@jsvd I've pushed the requested changes to this PR mainly:

- Allow to have a logstash directory in the pack
- refactor the name of the classes to remove mention of xpack
- get_installer_for_path instead of match?
- fix versions of dependencies
- and other small issues that you have mentioned about.

It's ready for another round.

lib/pluginmanager/remove.rb
```diff
@@ -21,21 +21,10 @@ def execute
 
     # since we previously did a gemfile.find(plugin) there is no reason why
     # remove would not work (return nil) here
```
jsvd: this comment no longer makes sense since we're not calling remove anymore

lib/pluginmanager/remove.rb
```diff
@@ -21,21 +21,10 @@ def execute
...
-    if gemfile.remove(plugin)
-      gemfile.save
-
-      puts("Removing #{plugin}")
-
-      # any errors will be logged to $stderr by invoke!
-      # output, exception = LogStash::Bundler.invoke!(:install => true, :clean => true)
-      output = LogStash::Bundler.invoke!(:install => true, :clean => true)
+    exit(1) unless !::Bundler::LogstashUninstall.uninstall!(plugin)
```
jsvd: should this be `exit(1) unless ::Bundler::LogstashUninstall.uninstall!(plugin)`?
right now it's the same as `exit(1) if ::Bundler::LogstashUninstall.uninstall!(plugin)`

lib/pluginmanager/remove.rb
```diff
@@ -21,21 +21,10 @@ def execute
...
 
-      remove_unused_locally_installed_gems!
-    end
+    remove_unused_locally_installed_gems!
   rescue => exception
-    gemfile.restore!
     report_exception("Remove Aborted", exception)
```
jsvd: maybe "Uninstall Aborted"?

lib/pluginmanager/ui.rb
```diff
@@ -0,0 +1,24 @@
+# encoding: utf-8
+module LogStash module PluginManager
+  # The command line commands should be able to report but they shouldn't
+  # require an explicite logger like log4j.
```
jsvd: explicit

lib/pluginmanager/utils/downloader.rb
```diff
@@ -0,0 +1,92 @@
...
+        downloaded_file = ::File.open(::File.join(download_to, ::File.basename(remote_file_uri.path)), "w")
+
+        HttpClient.start(remote_file_uri) do |http|
+          request = Net::HTTP::Get.new(remote_file_uri.path)
+
+          http.request(request) do |response|
+            if response.code == "200"
+              download_chunks(response, downloaded_file)
+            elsif response.code == "302"
+              # This is a bit defensive, In the real world we often creates redirects.
```
jsvd: this is already stated on line 44

lib/pluginmanager/bundler/logstash_uninstall.rb
```diff
@@ -0,0 +1,92 @@
...
+module Bundler
+  class LogstashUninstall
+    attr_reader :gemfile_path, :lockfile_path
+
+    def initialize(gemfile_path, lockfile_path)
+      @gemfile_path = gemfile_path
+      @lockfile_path = lockfile_path
+    end
+
+    # To be uninstall the candidate gems need
```
jsvd: uninstalled

lib/pluginmanager/bundler/logstash_uninstall.rb
```diff
@@ -0,0 +1,92 @@
...
+      gemfile.save
+
+      LogStash::PluginManager.ui.info("Successfully removed #{gem_name}")
+    ensure
+      file.close if file
+    end
+
+    def display_cant_remove_message(gem_name, dependencies_from)
+        message =<<-eos
+Warning, we cannot remove #{gem_name}, because other plugins depend on it.
```
jsvd: I suggest:

```
message =<<-eos
Failed to remove \"#{gem_name}\"` because the following plugins or libraries depend on it:

* #{dependencies_from.join("\n* ")}
eos
```

spec/unit/plugin_manager/install_strategy_factory_spec.rb
```diff
@@ -0,0 +1,27 @@
+# encoding: utf-8
+require "pluginmanager/install_strategy_factory"
+
+describe LogStash::PluginManager::InstallStrategyFactory do
+  subject { described_class }
+  let(:plugins_args) { [ "logstash-pack-mega" ] }
+
+  it "returns the first matched strategy" do
+    success = double("urixpack success")
```
jsvd: maybe update this name to urifetch or similar?

lib/pluginmanager/install.rb
```diff
@@ -17,6 +20,29 @@ class LogStash::PluginManager::Install < LogStash::PluginManager::Command
   # but the argument parsing does not support it for now so currently if specifying --version only
   # one plugin name can be also specified.
   def execute
+    # This is a special flow for XPACK related plugins,
```
jsvd: s/XPACK/PACK ?

lib/pluginmanager/install.rb
```diff
@@ -17,6 +20,29 @@ class LogStash::PluginManager::Install < LogStash::PluginManager::Command
...
+    # if we dont detect an xpack we will just use the normal `Bundle install` Strategy`
```
jsvd: same

spec/unit/plugin_manager/install_spec.rb
```diff
@@ -25,4 +30,29 @@
       cmd.execute
     end
   end
+
+  context "xpack" do
```
jsvd: s/xpack/pack ?

spec/unit/plugin_manager/utils/http_client_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        described_class.start(uri)
+      end
+    end
+  end
+
+  describe ".remove_file_exist?" do
+    let(:mock_http) { double("Net::HTTP") }
+
+    before do
+      expect(subject).to receive(:start).with(anything).and_yield(mock_http).at_least(:once)
```
jsvd: usually I prefer to use allow in `before` blocks and `expects` in assertions. any thoughts on that?

ph:
@jsvd thanks for the reviews, I've pushed all the requested changes!

ph:
I've fixed the failling test from the integration suite, forgot to adjust the expectation when I've changed the message.

jsvd:
LGTM. test failure is unrelated

Issue #6160, opened 2016-11-01
PeriodicPoller timeout error in the log


ph:
When the Periodic Poller that run the stats collection for the JVM/OS timeout it will log the message using the `error` level. 

Under a Logstash running really high throughput this can be a problem when resource starvation occurs, and this error can appear in the log. 

I see two solutions:

A. We make the call never timeout and just block until it can receive the details.
B. We change the log level for debug.

I am more in favor of B.

ph:
related to https://discuss.elastic.co/t/logstash-poller-error-concurrent-timeouterror-logstash-5-0-0-beta1/62448

michaelburn:
Hi, I've the same issue with Logstash 5.0.0.
I tried to get rid of these events by modifying the log4j.properties file like this:
...
logger.logstash.instrument.periodicpoller.jvm = fatal
...
But I can't get this to work. Does anybody have the right configuration-entry to exclude the events from the Logstash-Logfile?

Thanks

Issue #6158, opened 2016-11-01
increase the default value of PeriodicPoller JVM/OS


ph:
The current implement of the periodic poller to retrieve stats on the JVM and OS layer are executed every 1s, we have seen report of theses task timeout. 

I am not sure if its a sane default anymore, Maybe 5s or 10s (the default of the collector) would be good enough.

https://discuss.elastic.co/t/logstash-poller-error-concurrent-timeouterror-logstash-5-0-0-beta1/62448/12

PR #6169, opened 2016-11-02
Change how PeriodicPollers log their exception

ph:
In a busy logstash install and the way we interact with core stats from
the jvm or the os, the poller can timeout. This exception is logged as
an error. but this shouldn't impact normal operation.

This PR changes the following:

- Change the interval for 5s instead of 1s
- Make the timeout bigger 120s instead of 60
- Concurrent::TimeoutError will be logged using debug instead of error
- Any other exception will use the error level.
- add tests

Fixes: #6160
Fixes: #6158


ph:
kafka integration failed.

ph:
@suyograo can you take a look?

suyograo:
LGTM

PR #6172, opened 2016-11-02
Make logstash.yml optional

### LINKED ISSUE #6170: Do not require logstash.yml when bootstrapping Logstash
5.0.0 required Logstash to have a valid `logstash.yml` before it could start successfully. This was mostly fine for users who installed Logstash via `tar.gz`, but many many folks who install it via packages still start Logstash manually. Also, our documentation uses `-e` flag for getting started on Logstash and sending their first event. 

Initially, I thought we could disable settings for `-e` -- like we do now for `--help`, but this list can grow and it makes it hard to do this reliably across the bootstrap levels. (agent, runner, loader). 

`logstash.yml` has only defaults defined, and there is no required parameter to start Logstash. We should be able to use the defaults if no `logstash.yml`. Obviously, this is not ideal from a user point of view, so we should log a warning but continue to bootstrap.

suyograo:
5.0.0 required Logstash to have a valid logstash.yml before it could start successfully. This
was mostly fine for users who installed Logstash via tar.gz, but many many folks who install
it via packages still start Logstash manually. Also, our documentation uses -e flag for
getting started on Logstash and sending their first event.logstash.yml has only defaults defined,
and there is no required parameter to start Logstash. We should be able to use the defaults if no
logstash.yml. Obviously, this is not ideal from a user point of view, so we should log a warning but
continue to bootstrap.

Fixes #6170

### DIFF
--- a/logstash-core/lib/logstash/logging/logger.rb
+++ b/logstash-core/lib/logstash/logging/logger.rb
@@ -1,4 +1,5 @@
 require "logstash/java_integration"
+require "uri"
 
 module LogStash
   module Logging
@@ -6,6 +7,7 @@ module LogStash
       java_import org.apache.logging.log4j.Level
       java_import org.apache.logging.log4j.LogManager
       java_import org.apache.logging.log4j.core.config.Configurator
+      java_import org.apache.logging.log4j.core.config.DefaultConfiguration
       @@config_mutex = Mutex.new
       @@logging_context = nil
 
@@ -70,7 +72,16 @@ module LogStash
       def self.initialize(config_location)
         @@config_mutex.synchronize do
           if @@logging_context.nil?
-            @@logging_context = Configurator.initialize(nil, config_location)
+            file_path = URI(config_location).path
+            if ::File.exists?(file_path)
+              logs_location = java.lang.System.getProperty("ls.logs")
+              puts "Sending Logstash's logs to #{logs_location} which is now configured via log4j2.properties"
+              @@logging_context = Configurator.initialize(nil, config_location)
+            else
+              # fall back to default config
+              puts "Could not find log4j2 configuration at path #{file_path}. Using default config which logs to console"
+              @@logging_context = Configurator.initialize(DefaultConfiguration.new)
+            end
           end
         end
       end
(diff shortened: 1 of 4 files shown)


logstash-core/lib/logstash/runner.rb
```diff
@@ -288,7 +279,12 @@ def execute
     show_short_help
     return 1
   rescue => e
-    logger.fatal(I18n.t("oops"), :error => e, :backtrace => e.backtrace)
+    # if logger itself is not initialized
+    if LogStash::Logging::Logger.get_logging_context.nil?
+      puts e
```
jsvd: I suggest here something like:

```ruby
puts "#{I18n.t("oops")} :error => #{e}, :backtrace => #{e.backtrace}"
```

jsvd:
minor suggestion for better user feedback, otherwise LGTM

logstash-core/lib/logstash/logging/logger.rb
```diff
@@ -70,7 +72,16 @@ def self.configure_logging(level, path = LogManager::ROOT_LOGGER_NAME)
       def self.initialize(config_location)
         @@config_mutex.synchronize do
           if @@logging_context.nil?
-            @@logging_context = Configurator.initialize(nil, config_location)
+            file_path = URI(config_location).path
+            if ::File.exists?(file_path)
+              logs_location = java.lang.System.getProperty("ls.logs")
```
talevy: I'm a little confused by this, is this meant to replace

https://github.com/elastic/logstash/blob/5.0/logstash-core/lib/logstash/runner.rb#L191-L194

talevy: or is this meant to hijack the messaging of that annoying "can't find log4j.properties" message that log4j provides

logstash-core/lib/logstash/runner.rb
talevy: also, why isn't this going to `$stderr`?

suyograo:
@talevy updated to use `stderr`.

talevy:
LGTM

Issue #6173, opened 2016-11-02
inefficient tags removal in Filters::Base


colinsurprenant:
While looking at #6142 I noticed that the [tags removal code](https://github.com/elastic/logstash/blob/ae60c7dfd44d32892f526c853dab1bde5705e370/logstash-core/lib/logstash/filters/base.rb#L190-L197) is inefficient. 

For every tag in the `remove_tag` ivar it gets and sets the Event `tags` field. This operation could be done once before and after the loop.

I am also curious to why this code was not moved part of the `Decorators` module?

colinsurprenant:
closed by #6177

PR #6177, opened 2016-11-02
fix tagging edge cases

### LINKED ISSUE #6142: Logstash 5.0 refuse to start
Hi,

My logstash v5 crashed yesterday and it refused to start again

Here's the log file 

https://gist.github.com/JorisAndrade/b5174ff18feee5b3a7e36772d7ac8a10

colinsurprenant:
This fixes #6142 and #6173 

it includes fixes for the following tagging bugs:
- adding tags using `Event#tag` over pre-existing bare `String` value in the `tags` field
- adding tags using config `add_tag` over pre-existing bare `String` value in the `tags` field
- removing tags over pre-existing bare `String` value in the `tags` field
- inefficient `add_tag` and `remove_tag` implementations

### DIFF
--- a/logstash-core-event-java/src/main/java/org/logstash/Event.java
+++ b/logstash-core-event-java/src/main/java/org/logstash/Event.java
@@ -15,6 +15,7 @@ import org.logstash.ackedqueue.Queueable;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
@@ -76,9 +77,20 @@ public class Event implements Cloneable, Serializable, Queueable {
         this.metadata_accessors = new Accessors(this.metadata);
 
         this.cancelled = false;
-        this.timestamp = initTimestamp(data.get(TIMESTAMP));
+
+        Object providedTimestamp = data.get(TIMESTAMP);
+        // keep reference to the parsedTimestamp for tagging below
+        Timestamp parsedTimestamp = initTimestamp(providedTimestamp);
+        this.timestamp = (parsedTimestamp == null) ? Timestamp.now() : parsedTimestamp;
+
         this.data.put(TIMESTAMP, this.timestamp);
         this.accessors = new Accessors(this.data);
+
+        // the tag() method has to be called after the Accessors initialization
+        if (parsedTimestamp == null) {
+            tag(TIMESTAMP_FAILURE_TAG);
+            this.setField(TIMESTAMP_FAILURE_FIELD, providedTimestamp);
+        }
     }
 
     public Map<String, Object> getData() {
@@ -338,22 +350,36 @@ public class Event implements Cloneable, Serializable, Queueable {
             logger.warn("Error parsing " + TIMESTAMP + " string value=" + o.toString());
         }
 
-        tag(TIMESTAMP_FAILURE_TAG);
-        this.data.put(TIMESTAMP_FAILURE_FIELD, o);
-
-        return Timestamp.now();
+        return null;
     }
 
     public void tag(String tag) {
-        List<Object> tags = (List<Object>) this.data.get("tags");
-        if (tags == null) {
-            tags = new ArrayList<>();
-            this.data.put("tags", tags);
+        List<Object> tags;
+        Object _tags = this.getField("tags");
+
+        // short circuit the null case where we know we won't need deduplication step below at the end
+        if (_tags == null) {
+            setField("tags", Arrays.asList(tag));
+            return;
         }
 
+        // assign to tags var the proper List of either the existing _tags List or a new List containing whatever non-List item was in the tags field
+        if (_tags instanceof List) {
+            tags = (List<Object>) _tags;
+        } else {
+            // tags field has a value but not in a List, convert in into a List
+            tags = new ArrayList<>();
+            tags.add(_tags);
+        }
+
+        // now make sure the tags list does not already contain the tag
+        // TODO: we should eventually look into using alternate data structures to do more efficient dedup but that will require properly defining the tagging API too
         if (!tags.contains(tag)) {
             tags.add(tag);
         }
+
+        // set that back as a proper BiValue
+        this.setField("tags", tags);
     }
 
     public byte[] serialize() throws IOException {
--- a/logstash-core-event-java/src/test/java/org/logstash/EventTest.java
+++ b/logstash-core-event-java/src/test/java/org/logstash/EventTest.java
@@ -1,5 +1,6 @@
 package org.logstash;
 
+import org.junit.Assert;
 import org.junit.Test;
 
 import java.io.IOException;
@@ -266,4 +267,28 @@ public class EventTest {
     public void testFromJsonWithPartialInvalidJsonArray() throws Exception {
         Event.fromJson("[{\"foo\":\"bar\"}, 1]");
     }
+
+    @Test
+    public void testTagOnEmptyTagsField() throws Exception {
+        Event e = new Event();
+        e.tag("foo");
+
+        List<String> tags = (List<String>)e.getField("tags");
+        assertEquals(tags.size(), 1);
+        assertEquals(tags.get(0), "foo");
+    }
+
+    @Test
+    public void testTagOnExistingTagsField() throws Exception {
+        Map data = new HashMap();
+        data.put("tags", "foo");
+        Event e = new Event(data);
+        e.tag("bar");
+
+        List<String> tags = (List<String>)e.getField("tags");
+        assertEquals(tags.size(), 2);
+        assertEquals(tags.get(0), "foo");
+        assertEquals(tags.get(1), "bar");
+      }
+
 }
--- a/logstash-core/lib/logstash/filters/base.rb
+++ b/logstash-core/lib/logstash/filters/base.rb
@@ -195,14 +195,20 @@ class LogStash::Filters::Base < LogStash::Plugin
     # note below that the tags array field needs to be updated then reassigned to the event.
     # this is important because a construct like event["tags"].delete(tag) will not work
     # in the current Java event implementation. see https://github.com/elastic/logstash/issues/4140
+
+    tags = event.get("tags")
+    return unless tags
+
+    tags = Array(tags)
     @remove_tag.each do |tag|
-      tags = event.get("tags")
-      break if tags.nil? || tags.empty?
+      break if tags.empty?
+
       tag = event.sprintf(tag)
       @logger.debug? and @logger.debug("filters/#{self.class.name}: removing tag", :tag => tag)
       tags.delete(tag)
-      event.set("tags", tags)
     end
+
+    event.set("tags", tags)
   end # def filter_matched
 
   protected
--- a/logstash-core/lib/logstash/util/decorators.rb
+++ b/logstash-core/lib/logstash/util/decorators.rb
@@ -33,17 +33,20 @@ module LogStash::Util
     end
 
     # tags is an array of string. sprintf syntax can be used.
-    def add_tags(tags, event, pluginname)
-      tags.each do |tag|
-        tag = event.sprintf(tag)
-        self.logger.debug? and self.logger.debug("#{pluginname}: adding tag", "tag" => tag)
+    def add_tags(new_tags, event, pluginname)
+      tags = event.get("tags")
+      tags = tags.nil? ? [] : Array(tags)
+
+      new_tags.each do |new_tag|
+        new_tag = event.sprintf(new_tag)
+        self.logger.debug? and self.logger.debug("#{pluginname}: adding tag", "tag" => new_tag)
         # note below that the tags array field needs to be updated then reassigned to the event.
         # this is important because a construct like event["tags"] << tag will not work
         # in the current Java event implementation. see https://github.com/elastic/logstash/issues/4140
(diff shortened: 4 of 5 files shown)


colinsurprenant:
As reported in #6142 these 2 scenarios now result in the correct behaviour 

```sh
$ echo "{\"tags\":\"foo\"}" | bin/logstash -e 'input{stdin{codec => json_lines{}}} filter{geoip{source => "foo"}} output{stdout{codec => rubydebug{}}}'

[2016-11-02T18:59:52,719][INFO ][logstash.filters.geoip   ] Using geoip database
[2016-11-02T18:59:52,749][INFO ][logstash.pipeline        ] Starting pipeline
[2016-11-02T18:59:52,753][INFO ][logstash.pipeline        ] Pipeline main started
[2016-11-02T18:59:52,830][INFO ][logstash.agent           ] Successfully started Logstash API endpoint
{
    "@timestamp" => 2016-11-02T22:59:52.810Z,
         "geoip" => {},
      "@version" => "1",
          "host" => "colin-mbp13r.local",
          "tags" => [
        [0] "foo",
        [1] "_geoip_lookup_failure"
    ]
}
[2016-11-02T18:59:55,770][WARN ][logstash.agent           ] stopping pipeline


$ echo "{\"tags\":\"foo\"}" | bin/logstash -e 'input{stdin{codec => json_lines{}}} filter{mutate{add_tag => "bar"}} output{stdout{codec => rubydebug{}}}'

[2016-11-02T19:00:19,719][INFO ][logstash.pipeline        ] Starting pipeline
[2016-11-02T19:00:19,727][INFO ][logstash.pipeline        ] Pipeline main started
[2016-11-02T19:00:19,823][INFO ][logstash.agent           ] Successfully started Logstash API endpoint
{
    "@timestamp" => 2016-11-02T23:00:19.835Z,
      "@version" => "1",
          "host" => "colin-mbp13r.local",
          "tags" => [
        [0] "foo",
        [1] "bar"
    ]
}
[2016-11-02T19:00:22,746][WARN ][logstash.agent           ] stopping pipeline
```

logstash-core-event-java/src/main/java/org/logstash/Event.java
```diff
@@ -76,9 +76,19 @@ public Event(Map data)
         this.metadata_accessors = new Accessors(this.metadata);
 
         this.cancelled = false;
-        this.timestamp = initTimestamp(data.get(TIMESTAMP));
+
+        Object provided_timestamp = data.get(TIMESTAMP);
```
talevy: we should try to stop using snake_case variable names.

suyograo:
@colinsurprenant please label with target versions :)

logstash-core-event-java/src/main/java/org/logstash/Event.java
```diff
@@ -76,9 +76,19 @@ public Event(Map data)
...
+        Timestamp parsed_timestamp = initTimestamp(provided_timestamp);
+        this.timestamp = (parsed_timestamp == null) ? Timestamp.now() : parsed_timestamp;
+
         this.data.put(TIMESTAMP, this.timestamp);
         this.accessors = new Accessors(this.data);
+
+        // the tag() method has to be called after the Accessors initialization
+        if (parsed_timestamp == null) {
```
talevy: how is this ever `null` if line 82 checks for this and then sets it to a non-null value

logstash-core-event-java/src/main/java/org/logstash/Event.java
```diff
@@ -338,22 +348,33 @@ private Timestamp initTimestamp(Object o) {
...
         }
 
-        tag(TIMESTAMP_FAILURE_TAG);
-        this.data.put(TIMESTAMP_FAILURE_FIELD, o);
-
-        return Timestamp.now();
+        return null;
     }
 
     public void tag(String tag) {
```
talevy: I think this can be cleaned up a little. what do you think of something like this?

```java
public void tag(String tag) {
    Object existingTags = getUnconvertedField("tags");
    if (existingTags == null) {
        setField("tags", Arrays.asList(tag));
    } else if (existingTags instanceof List) {
        existingTags.add(tag);
    } else if (existingTags instanceof String) {
        List existingTagsAsList = Arrays.asList(existingTags);
        existingTagsAsList.add(tag);
        setField("tags", existingTagsAsList);
    }else {
        // tags is not a List... why did this happen?
        // why do we allow this?
        // how can we prevent this?
        // let's assume that we can't... we should error out?!
    }
}
```

I may be mis-understanding the edge-cases here, though.

Given that both JRuby's and Java's objects implement the `List` and `String` interfaces... I think it is 
safe to use `getUnconvertedField`, and that will be faster (I may be wrong here).

talevy:
what does the timestamp initialization changes exist in this PR? they seem unrelated to the revamped `tag` method, no?

logstash-core-event-java/src/main/java/org/logstash/Event.java
jordansissel: Yeah we should be using unicode more anyway, like `Object providedⓉimestamp ...` ;P

+1 on more java-ish style camelCase.

jordansissel:
I like the proposed solution. Tags were something I never really defined with specifics. I'm happy to skip the code review since it looks like @talevy already did the duty there -- if you want me to review the code, I am happy to do so.

I reviewed the tests, and the two cases that it adds for coverage LGTM :)

logstash-core-event-java/src/main/java/org/logstash/Event.java
colinsurprenant: lol good catch, will change. (switching between Ruby and Java code :P )

logstash-core-event-java/src/main/java/org/logstash/Event.java
colinsurprenant: no it does not. `parsed_timestamp` is only assigned once by `initTimestamp()`. re-read the code. it is like this because we need to defer setting the tag **after** the Accessors have been initialized

logstash-core-event-java/src/main/java/org/logstash/Event.java
colinsurprenant: so yes, obviously, to clarify, all (my) Java code uses the Java camelCase convention, this is an oversight dur to switching between both languages.

logstash-core-event-java/src/main/java/org/logstash/Event.java
```diff
@@ -338,22 +348,33 @@ private Timestamp initTimestamp(Object o) {
...
-        List<Object> tags = (List<Object>) this.data.get("tags");
-        if (tags == null) {
+        List<Object> tags;
+        Object _tags = this.getField("tags");
```
colinsurprenant: hmm. no.

colinsurprenant:
@talevy because the timestamp initialization tags upon invalid timestamp format and the tag method, per the refactor, requires the accessors to be initialized. I will look at your code proposal see if it makes sense.

logstash-core-event-java/src/main/java/org/logstash/Event.java
talevy: woops, I'll re-read. tried to expand to the point that I thought I saw the state of the world, should have checked the raw and not the diff

logstash-core-event-java/src/main/java/org/logstash/Event.java
```diff
@@ -338,22 +348,33 @@ private Timestamp initTimestamp(Object o) {
             logger.warn("Error parsing " + TIMESTAMP + " string value=" + o.toString());
         }
 
-        tag(TIMESTAMP_FAILURE_TAG);
-        this.data.put(TIMESTAMP_FAILURE_FIELD, o);
-
-        return Timestamp.now();
+        return null;
```
talevy: woops, I'll re-read. tried to expand to the point that I thought I saw the state of the world, should have checked the raw and not the diff

colinsurprenant:
@jordansissel my specific concern and I'd love to get your thoughts on it is: the `Event#tag` method deduplicates tags and this has been like that since a while, before the Java Event if I am not mistaken, but the config `add_tag` **does not** deduplicates and this behaviour is actually spec'ed/tested for - but I do not understand why we would want that. When doing the fix I actually added the dedup in `add_tag` but broke tests, so I commented this "fix" out. So even if we decide it makes sense to also dedup in `add_tag`, this will be a change in behaviour and I am on the fence about doing that in a minor/bugfix release. WDYT?

logstash-core-event-java/src/main/java/org/logstash/Event.java
suyograo: why not?

colinsurprenant: 2 things:
- There is not really any `_` prefix convention in Java but I sometimes use var names starting with underscore to note a temporary holder for the same var name without the underscore or to note a "throwaway" tmp var, all in very local contexts.
- I personally like it and find it concise. This is really a very **minor** style issue. Given the insignificant nature of this it is mostly a question of preference and since I wrote the code I think I can opt on keeping it.

logstash-core-event-java/src/main/java/org/logstash/Event.java
colinsurprenant: as per our discussion we keep current logic. I added comments and switched to use Arrays.asList() for more concision.

logstash-core-event-java/src/main/java/org/logstash/Event.java
suyograo: @colinsurprenant I really don't want to block this PR based on a variable naming suggestion, and I agree it is minor, but this notion of "I wrote code so I'll do what I want" is just antithetical to our review process. 

If it is _that_ minor, why don't you accept it? On the contrary, if we move forward with this, we'll introduce a new convention here that I have not seen used in Java and may permeate our code base. We have a chance to correct it. I can also argue this makes reading code hard. How do I know `_tags` is different than input parameter `tags`. it is so easy to miss when reading this code snippet. 

I do understand we don't have a style guide but I am not sure how this is different than accepting camel case as specified above. And, I look to Elasticsearch codebase as a style reference.

suyograo: And if you want a style guide, here is one from [Oracle](http://www.oracle.com/technetwork/java/codeconventions-135099.html):

> Except for variables, all instance, class, and class constants are in mixed case with a lowercase first letter. Internal words start with capital letters. Variable names should not start with underscore _ or dollar sign $ characters, even though both are allowed.

colinsurprenant: We are sweating very small stuff here. I thought we did have this discussion where non-fundamental or ad-hoc style choices, mostly based on personal preferences should be left at the author discretion. I don't see how this sets any styling precedent in any way. this is a local variable in a very local context. Regardless of Oracle style guide there are lots of precedents for using `_var` style of naming, typically for private vars. There is definitely no consensus around that, just like many other styling issues. At some point we have to respect some of the author decision. Obviously if the code is a mess, then it must be fixed but is this what we are talking about here?

Also, if we are going to use ES as an example, I urge you to execute this agaist the ES codebase and see for yourself.

```sh
ack "\s_[a-z][^\s]"  --type=java
```

suyograo: I give up. Let's move on

talevy:
tests seem to be failing on a beats bug in a test


```
[2016-11-03T17:53:18,473][ERROR][logstash.agent           ] Pipeline aborted due to error {:exception=>#<LogStash::ConfigurationError: Using `verify_mode` set to PEER or FORCE_PEER, requires the configuration of `certificate_authorities`>, :backtrace=>["/home/travis/build/elastic/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/1.9/gems/logstash-input-beats-3.1.8-java/lib/logstash/inputs/beats.rb:157:in `register'", "/home/travis/build/elastic/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/logstash-core/lib/logstash/pipeline.rb:348:in `start_inputs'", "org/jruby/RubyArray.java:1613:in `each'", "/home/travis/build/elastic/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/logstash-core/lib/logstash/pipeline.rb:347:in `start_inputs'", "/home/travis/build/elastic/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/logstash-core/lib/logstash/pipeline.rb:223:in `start_workers'", "/home/travis/build/elastic/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/logstash-core/lib/logstash/pipeline.rb:179:in `run'", "/home/travis/build/elastic/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/logstash-core/lib/logstash/agent.rb:251:in `start_pipeline'"]}```


since this is unrelated, I'm good.

LGTM

colinsurprenant:
@tal how do we make the link between the beats test failure and this tagging PR?!

colinsurprenant:
Ok, I get it, I misread the integration error as being Kafka, it is with beats.

colinsurprenant:
Ok, so I will rebase and merge tomorrow morning.

JorisAndrade:
Would be great to see this release for the v5.0.1 :) 

Thanks for your work !

colinsurprenant:
merged in 5.x and master

JorisAndrade:
Perfect ! Thanks

PR #6178, opened 2016-11-03
[docs] Update breaking changes and java requirements

suyograo:


jordansissel:
LGTM.

Side task: We should make it work under Java 9... in the next few months maybe.

Issue #6186, opened 2016-11-03
Tutorial documentation typo


vreuter:
It looks like there's a typo in the tutorial documentation's filebeat example. Instructions are to set the port for the input section of the configuration to 5043, but the documentation states in a couple spots that filebeat will attempt to connect on port 5403.

dedemorton:
Fixed in #6187

PR #6187, opened 2016-11-03
Tutorial documentation: updated expected port number to reflect instructions

vreuter:
Fixes https://github.com/elastic/logstash/issues/6186


karmi:
Hi @vreuter, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/logstash/pull/6187.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @vreuter, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/logstash/pull/6187.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

vreuter:
Yeah sorry about that @karmi ; it should be done now.

PR #6190, opened 2016-11-03
Update branding in LS overview image

dedemorton:
So sad, but the time has come to use the new branding in our docs. Note that I plan to redo this image after I've gotten input from Alvin. But to make sure the branding is up-to-date for now, I'm tweaking the existing image.


suyograo:
LGTM

PR #6196, opened 2016-11-04
mark reload as failure if pipeline.run aborts

### LINKED ISSUE #6195: pipeline reload failures at pipeline.run stage still count as success in metrics
if a pipeline passes the config test check but terminates when executed (due to plugins raising exceptions during their register phase, for example), then logstash will still register that reload as successful, when it should be a failure.

jsvd:
currently a reload is only marked as a failured if it fails the classic
config test check, where we check for parameter names, existing plugins, etc.

This change waits for the pipeline to transition to running before
marking the reload as success, otherwise it is a failure.

fixes #6195

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -245,12 +245,12 @@ class LogStash::Agent
     return unless pipeline.is_a?(LogStash::Pipeline)
     return if pipeline.ready?
     @logger.debug("starting pipeline", :id => id)
-    Thread.new do
+    t = Thread.new do
       LogStash::Util.set_thread_name("pipeline.#{id}")
       begin
         pipeline.run
       rescue => e
-        @reload_metric.namespace([id.to_sym, :reloads]) do |n|
+        @reload_metric.namespace([id.to_sym, :reloads]).tap do |n|
           n.increment(:failures)
           n.gauge(:last_error, { :message => e.message, :backtrace => e.backtrace})
           n.gauge(:last_failure_timestamp, LogStash::Timestamp.now)
@@ -258,7 +258,15 @@ class LogStash::Agent
         @logger.error("Pipeline aborted due to error", :exception => e, :backtrace => e.backtrace)
       end
     end
-    sleep 0.01 until pipeline.ready?
+    while true do
+      if !t.alive?
+        return false
+      elsif pipeline.ready?
+        return true
+      else
+        sleep 0.01
+      end
+    end
   end
 
   def stop_pipeline(id)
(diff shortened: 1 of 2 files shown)


logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -527,5 +527,41 @@ class DummyOutput2 < DummyOutput; end
...
+        end
+
+        subject.send(:"reload_pipeline!", "main")
+      end
+
+      it "does not increase the successful reload count" do
+        snapshot = subject.metric.collector.snapshot_metric
+        value = snapshot.metric_store.get_with_path("/stats/pipelines")[:stats][:pipelines][:main][:reloads][:successes].value
+        expect(value).to eq(0)
+      end
```
ph: Can we use rspec's `change` matcher for this, I think this will make sure the value actually changes (or not)

suyograo:
@jsvd this needs to be rebased from master to get PH's fixes for tests.

logstash-core/spec/logstash/agent_spec.rb
jsvd: interesting, I don't think I ever used this matcher..from what I read it seems it's very useful when the action that changes the value happens during the test itself. but here the action (changing the configuration file) happens in the before by writing to the file itself.

How would you apply the change matcher here?

logstash-core/lib/logstash/agent.rb
```diff
@@ -245,20 +245,28 @@ def start_pipeline(id)
...
     return if pipeline.ready?
     @logger.debug("starting pipeline", :id => id)
-    Thread.new do
+    t = Thread.new do
       LogStash::Util.set_thread_name("pipeline.#{id}")
       begin
         pipeline.run
       rescue => e
-        @reload_metric.namespace([id.to_sym, :reloads]) do |n|
+        @reload_metric.namespace([id.to_sym, :reloads]).tap do |n|
```
jsvd: more like :facepalm: :)

logstash-core/spec/logstash/agent_spec.rb
ph: Well instead of making the trigger in the before :each you do it in the proc?

Something like this might work?

```ruby
expect { update_config }.to change { snapshot.metric_store.get_with_path("/stats/pipelines")[:stats][:pipelines][:main][:reloads][:successes].value }.by(1)
```


So rspec will do the following.

1. change proc
2. expect proc
3. change proc
4. apply the comparator matcher..

jsvd:
@ph updated specs to the change matcher

suyograo:
Still failing at 
```
ERROR: Installation Aborted, message: You have requested:
  addressable ~> 2.4.0
```

suyograo:
Sorry, this will not make 5.0.1, and i don't consider this a blocker, so moving the label to 5.1.0

ph:
I've created #6204 for the addressable problem

ph:
@jsvd Can you rebase this?

suyograo:
Ignore Kafka failure. LGTM

PR #6197, opened 2016-11-04
Update lock file for 5.0.1

suyograo:
Kafka broke compatibility again, so we need to put in the constraint for "< 6.0.0"


ph:
We are missing beats 3.1.8 in the lock file.

talevy:
@suyograo should this also be going into `5.x`?

suyograo:
@talevy 5.x does not have a lock file yet checked-in yet. It gets generated

PR #6205, opened 2016-11-07
remove duplicate list subcommand

glefloch:
When running `./logstash-plugin --help`, there was twice time the list sub command


suyograo:
LGTM

PR #6211, opened 2016-11-07
Add release notes for 5.0.1

suyograo:


docs/static/releasenotes.asciidoc
```diff
@@ -1,12 +1,55 @@
...
+[[logstash-5-0-1]]
+=== Logstash 5.0.1 Release Notes
+
+* Logstash can now start without logstash.yml, even though this is not ideal and not recommended for production. 
+  Requiring `logstash.yml` in 5.0.0 caused confusion among existing users who install LS from a package and tried to start it 
+  manually via `bin/logstash` ({lsissue}6170[Issue 6170]).
+* Fixed a bug when incoming data already had a field called `tags` and Logstash errored trying to append more 
+  tags to it ({lsissue}6142[Issue 6142]).
+* Increased the current polling interval of a background thread that retrieves stats from JVM and OS layer from 1s to 5s ({lsissue}6160[Issue 6160]).
+* Merge config strings when both `-e` and `-f` was supplied even if `-f` does not yield in any config files ({lsissue}6164[Issue 6164]).
```
dedemorton: I don't think this will make very much sense out of context (unless people follow the link to the issue). We need to add a few more words around this. Maybe say something like (not sure if I have the details completely right, but here goes):

"Changed the behavior of how Logstash merges config strings when you start Logstash with the `-e` and `-f` options specified.  Previously, Logstash required the config files specified by `-f` to be valid before merging. Logstash now reads the config specified by `-e` even if `-f` yields no config files ({lsissue}6164[Issue 6164])."

docs/static/releasenotes.asciidoc
```diff
@@ -1,12 +1,55 @@
...
+* Improve user feedback message when starting Logstash manually after a package install ({lsissue}6061[Issue 6061]).
```
dedemorton: Change to: "Improved the user feedback message that displays when....

docs/static/releasenotes.asciidoc
```diff
@@ -1,12 +1,55 @@
...
+
 [float]
-==== General
+==== Input Plugins
+
+*`Beats`*:
+
+* The plugin can now bind to a specific host using the `host` config (https://github.com/logstash-plugins/logstash-input-beats/issues/146[Issue 146]).
+* Fixed an issue where only the first certificate found in the CA chain were taken into consideration to verify clients (https://github.com/logstash-plugins/logstash-input-beats/issues/153[Issue 153]).
```
dedemorton: Change to: "...only the first certificate found in the CA chain was taken into consideration..."

docs/static/releasenotes.asciidoc
```diff
@@ -1,12 +1,55 @@
...
+* Validate the presence of `ssl_certificate_authorities` when `verify_mode` is set to `FORCE_PEER` or `PEER` (https://github.com/logstash-plugins/logstash-input-beats/issues/155[Issue 155]).
+* Fixed a typo in the default ciphers suite and added validations to catch such issues in the future (https://github.com/logstash-plugins/logstash-input-beats/issues/156[Issue 156]).
+
+*`RabbitMQ`*:
+
+* Fix reconnection issues when trying to connect to RabbitMQ when starting Logstash (https://github.com/logstash-plugins/logstash-input-rabbitmq/issues/91[Issue 91]).
```
dedemorton: This seems a bit vague. Not sure if I have the details right, but I think a bit more words around this problem would be useful. Maybe: 

"Fixed issue where Logstash was unable to connect to RabbitMQ because the input plugin was blocking the Logstash pipeline (https://github.com/logstash-plugins/logstash-input-rabbitmq/issues/91[Issue 91]). "

docs/static/releasenotes.asciidoc
```diff
@@ -1,12 +1,55 @@
...
+* To improve debugging, include URL of server when logging connection errors.
```
dedemorton: It's a bit disorienting to use the imperative here. Suggested edit:

"To improve debugging, we now include the URL of the server when logging connection errors."

docs/static/releasenotes.asciidoc
```diff
@@ -1,12 +1,55 @@
...
+
+==== Filter Plugins
+
+*`Grok`*:
+
+* When using the execution cancelation feature, Grok could crash Logstash under high concurrent loads. This has
```
dedemorton: Fix typo. Should be "execution cancellation feature"

docs/static/releasenotes.asciidoc
```diff
@@ -1,12 +1,55 @@
...
+==== Filter Plugins
+
+*`Grok`*:
+
+* When using the execution cancelation feature, Grok could crash Logstash under high concurrent loads. This has 
+  been fixed (https://github.com/logstash-plugins/logstash-filter-grok/issues/97[Issue 97]).
+
+*`XML`*:
+
+* Fix a bug that, when the plugin tried to report an invalid configuration, would report the following
```
dedemorton: A bit choppy. Maybe this is better:

"Fixed a bug where the plugin would report the following error when really it should have reported an invalid configuration:  `translation missing: en.logstash.agent.configuration.invalid_plugin_register`.

suyograo:
@dedemorton thanks for the edits. I updated this PR.

dedemorton:
LGTM

PR #6221, opened 2016-11-08
Include new kafka bug fix release

suyograo:
Also added constraint `gem "logstash-output-elasticsearch", "<= 5.2.0"` because the later versions have a feature that we don't want to include in 5.0.1


talevy:
if travis test is 🍏 📗 💚 🍵 , LGTM

PR #6224, opened 2016-11-08
Persistent UUID for Agent

andrewvc:
Add Persistent UUID for Agent

This adds two new fields 'id', and 'name' to the base metadata for API requests. These fields are now returned at all API endpoints by default.

The `id` field is the persisted UUID, the name field is the custom name the user has passed in (defaulted to the hostname).

I renamed `node_uuid` and `node_name` to just `id` and `name` to be inline with Elasticsearch.

This also fixed a broken test double in `webserver_spec.rb` that was using doubles across threads which created hidden errors.


logstash-core/lib/logstash/agent.rb
```diff
@@ -131,8 +133,42 @@ def shutdown
     shutdown_pipelines
   end
 
-  def node_uuid
-    @node_uuid ||= SecureRandom.uuid
+  def id
+    return @id if @id
+
+    uuid = nil
+    if File.exists?(id_path)
```
ph: `File` doesn't work all the time. because of the File input, would you mind adding `::File` to make sure we use the root namespace.  :(

logstash-core/lib/logstash/agent.rb
```diff
@@ -131,8 +133,42 @@ def shutdown
...
+      begin
+        uuid = File.open(id_path) {|f| f.each_line.first.chomp }
```
ph: I think the file will only contains the UUID? So we can use `File.read`

```ruby
File.read(id_path).chomp
```

logstash-core/lib/logstash/agent.rb
```diff
@@ -131,8 +133,42 @@ def shutdown
...
+                    :class => e.class.name)
+      end
+    end
+
+    @id = uuid
+  end
+
+  def id_path
+    @id_path ||= File.join(settings.get("path.data"), "uuid")
   end
```
ph: the `uuid` part should probably be a constant?

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -13,9 +14,20 @@
...
   let(:config_file_txt) { "input { generator { count => 100000 } } output { }" }
 
+  def clear_data_dir
+    data_path = agent_settings.get("path.data")
+    Dir.foreach(data_path) do |f|
+      next if f == "." || f == ".." || f == ".gitkeep"
+      FileUtils.rm_rf(File.join(data_path, f))
+    end
+  end
+
```
ph: This method should really useful, maybe we should move it to spec/support?

logstash-core/lib/logstash/agent.rb
andrewvc: @ph yes! I wanted to be defensive here. Sometimes people open files with vim and mess around with them.  We def need chomp if someone opens in vim then saves since that'll add a newline.

logstash-core/spec/logstash/agent_spec.rb
andrewvc: Done!

logstash-core/lib/logstash/agent.rb
andrewvc: Done!

suyograo:
LGTM if travis passes (ignore Kafka)

ph:
LGTM

PR #6225, opened 2016-11-08
[WIP] add queue.max_acked_checkpoint and queue.checkpoint_rate settings

colinsurprenant:
this PR adds support for these new queue settings
- `queue.checkpoint.acks`
- `queue.checkpoint.writes`

We decided to postpone `queue.checkpoint.interval` to validate the need for it since queue is now checkpointed on both acks and writes.

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -107,6 +107,21 @@
 #
 # queue.max_events: 0
 #
+# If using queue.type: persisted, the maximum number of acked events before forcing a checkpoint
+# Default is 1024, 0 for unlimited
+#
+# queue.checkpoint.acks: 1024
+#
+# If using queue.type: persisted, the maximum number of written events before forcing a checkpoint
+# Default is 1024, 0 for unlimited
+#
+# queue.checkpoint.writes: 1024
+#
+# If using queue.type: persisted, the interval in milliseconds when a checkpoint is forced on the head page
+# Default is 1000, 0 for no periodic checkpoint.
+#
+# queue.checkpoint.interval: 1000
+#
 # ------------ Metrics Settings --------------
 #
 # Bind address for the metrics REST endpoint
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -94,9 +94,9 @@ module LogStash; class Pipeline
     rescue => e
       raise
     end
-    queue = build_queue_from_settings
-    @input_queue_client = queue.write_client
-    @filter_queue_client = queue.read_client
+    @queue = build_queue_from_settings
+    @input_queue_client = @queue.write_client
+    @filter_queue_client = @queue.read_client
     @signal_queue = Queue.new
     # Note that @infilght_batches as a central mechanism for tracking inflight
     # batches will fail if we have multiple read clients here.
@@ -119,6 +119,9 @@ module LogStash; class Pipeline
     queue_type = settings.get("queue.type")
     queue_page_capacity = settings.get("queue.page_capacity")
     max_events = settings.get("queue.max_events")
+    checkpoint_max_acks = settings.get("queue.checkpoint.acks")
+    checkpoint_max_writes = settings.get("queue.checkpoint.writes")
+    checkpoint_max_interval = settings.get("queue.checkpoint.interval")
 
     if queue_type == "memory_acked"
       # memory_acked is used in tests/specs
@@ -129,7 +132,7 @@ module LogStash; class Pipeline
     elsif queue_type == "persisted"
       # persisted is the disk based acked queue
       queue_path = settings.get("path.queue")
-      LogStash::Util::WrappedAckedQueue.create_file_based(queue_path, queue_page_capacity, max_events)
+      LogStash::Util::WrappedAckedQueue.create_file_based(queue_path, queue_page_capacity, max_events, checkpoint_max_writes, checkpoint_max_acks, checkpoint_max_interval)
     else
       raise(ConfigurationError, "invalid queue.type setting")
     end
@@ -194,6 +197,7 @@ module LogStash; class Pipeline
     shutdown_workers
 
     @filter_queue_client.close
+    @queue.close
 
     @logger.debug("Pipeline #{@pipeline_id} has been shutdown")
 
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
@@ -82,7 +82,15 @@ public abstract class Page implements Closeable {
         return this.elementCount <= 0 ? 0 : Math.max(0, (maxSeqNum() - this.firstUnreadSeqNum) + 1);
     }
 
-    public void ack(List<Long> seqNums) throws IOException {
+    // update the page acking bitset. trigger checkpoint on the page if it is fully acked or if we acked more than the
+    // configured threshold checkpointMaxAcks.
+    // note that if the fully acked tail page is the first unacked page, it is not really necessary to also checkpoint
+    // the head page to update firstUnackedPageNum because it will be updated in the next upcoming head page checkpoint
+    // and in a crash condition, the Queue open recovery will detect and purge fully acked pages
+    //
+    // @param seqNums the list of same-page seqNums to ack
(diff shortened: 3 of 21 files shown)


colinsurprenant:
Options are now
- `queue.checkpoint.max_acks` max number of acks before forcing a checkpoint
- `queue.checkpoint.max_writes` max number of writes before forcing a checkpoint 
- `queue.checkpoint.max_interval` max interval in millisec before forcing a checkpoint

TODO:
- Verify for possible multi-pages checkpointing? many pages can be be ack'ed after a threshold on the number of acks or time interval so when we reach that threshold many tail pages checkpoint could be required

Observation - @guyboertje we have to rethink our options+settings handling, the amount of changes/refactoring to support or modify an option is a bit overwhelming.

config/logstash.yml
```diff
@@ -107,6 +107,21 @@
 #
 # queue.max_events: 0
 #
+# If using queue.type: persisted, the maximum number of acked events before forcing a checkpoint
+# Default is 1024, 0 for unlimited
+#
+# queue.checkpoint.max_acks: 1024
```
suyograo: as discussed on Slack, we should just call them `queue.checkpoint.{acks, writes, interval}`

colinsurprenant: +1

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -94,9 +94,9 @@ def initialize(config_str, settings = SETTINGS, namespaced_metric = nil)
     rescue => e
       raise
     end
-    queue = build_queue_from_settings
-    @input_queue_client = queue.write_client
-    @filter_queue_client = queue.read_client
+    @queue = build_queue_from_settings
```
guyboertje: Is there any particular reason why an IVAR reference to the queue itself is necessary? I had it in mind that the only access to the queue was via the clients.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -194,6 +197,7 @@ def run
     shutdown_workers
 
     @filter_queue_client.close
+    @queue.close
```
guyboertje: the client should close the queue

logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
```diff
@@ -82,7 +82,15 @@ public long unreadCount() {
         return this.elementCount <= 0 ? 0 : Math.max(0, (maxSeqNum() - this.firstUnreadSeqNum) + 1);
     }
 
-    public void ack(List<Long> seqNums) throws IOException {
+    // update the page acking bitset. trigger checkpoint on the page if it is fully acked or if we acked more than the
+    // configured threshold checkpointMaxAcks.
+    // note that if the fully acked tail page is the first unacked page, it is not really necessary to also checkpoint
+    // the head page to update firstUnackedPageNum because it will be updated in the next upcoming head page checkpoint
+    // and in a crash condition, the Queue open recovery will detect and purge fully acked pages
```
guyboertje: 👍

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -114,67 +130,91 @@ public void open() throws IOException {
 
         // reconstruct all tail pages state upto but excluding the head page
         for (int pageNum = headCheckpoint.getFirstUnackedPageNum(); pageNum < headCheckpoint.getPageNum(); pageNum++) {
-            Checkpoint tailCheckpoint = checkpointIO.read(checkpointIO.tailFileName(pageNum));
 
-            if (tailCheckpoint == null) { throw new IOException(checkpointIO.tailFileName(pageNum) + " not found"); }
+            // all tail checkpoints in the sequence should exist, if not abort mission with a NoSuchFileException
```
guyboertje: 👍

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: yes, because we need to properly close the queue on shutdown!

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: hmmm 🤔  lemme see.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -114,67 +130,91 @@ public void open() throws IOException {
...
+                // create a tail page with a null PageIO and add it to tail pages but not unreadTailPages
+                // since it is fully read since fully acked
+                this.tailPages.add(new TailPage(checkpoint, this, null));
             }
         } else {
-            // for some reason we found a fully acked page, let's purge it.
-            p.purge();
+            TailPage p = new TailPage(checkpoint, this, pageIO);
+            this.tailPages.add(p);
+            this.unreadTailPages.add(p);
```
guyboertje: For @jsvd and @colinsurprenant - is this where Joao would increment the on-disk bytesize of the queue?

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -419,23 +451,52 @@ public void ack(List<Long> seqNums) throws IOException {
...
+            TailPageResult result = null;
+
+            if (this.tailPages.size() > 0) {
+                // short-circuit: first check in the first tail page as it is the most likely page where acking will happen
+                TailPage p = this.tailPages.get(0);
+                if (p.getMinSeqNum() > 0 && firstAckSeqNum >= p.getMinSeqNum() && firstAckSeqNum < p.getMinSeqNum() + p.getElementCount()) {
+                    result = new TailPageResult(p, 0);
+                } else {
+                    // dual search strategy: if few tail pages search linearily otherwise perform binary search
+                    result = (this.tailPages.size() > 3) ? binaryFindPageForSeqnum(firstAckSeqNum) : linearFindPageForSeqnum(firstAckSeqNum);
```
guyboertje: 👍

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: re-reading the code I do not think the client should actually close the queue, here's why:
- the queue object returned by `build_queue_from_settings` is the "main" object from which the read and write clients are created. 
- there are 2 clients from the same queue object, which one do you close and when? to me the "client" should "close" itself but not necessarily close its parent object so that, for example, the write side could close itself but the read side could still be opened. 
- it would be easier to understand to uniformly support close on both the read and write clients which does not necessarily affect the queue object itself and leave the queue closing explicit. It make the closing logic easier to understand and follow IMO.

So I'd actually vote to add close on both clients and keep queue closing as-is.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -114,67 +130,91 @@ public void open() throws IOException {
...
+                // ignore
+            }
+
+            if (this.tailPages.size() == 0) {
+                // this is the first tail page and it is fully acked so just purge it
+                this.checkpointIO.purge(this.checkpointIO.tailFileName(checkpoint.getPageNum()));
+            } else {
+                // create a tail page with a null PageIO and add it to tail pages but not unreadTailPages
+                // since it is fully read since fully acked
+                this.tailPages.add(new TailPage(checkpoint, this, null));
```
colinsurprenant: +1

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
guyboertje: merely an aide memoire

guyboertje:
Other than the comments on the queue close - LGTM. 🍰

guyboertje:
queue close comments discussed and waived, all good now.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
colinsurprenant: done

colinsurprenant:
looking at/fixing build failures, rebasing+squashing and merging.

colinsurprenant:
squashed and merged in 5.x and master

PR #6226, opened 2016-11-08
Allow negative array offsets in field references

jordansissel:
This allows references of `[-1]` and similar negative numbers to mean an offset from the end of the list.

### DIFF
--- a/logstash-core-event-java/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core-event-java/src/main/java/org/logstash/Accessors.java
@@ -188,7 +188,7 @@ public class Accessors {
      * @param size the size of the list.
      * @return the positive integer offset for the list given by index i.
      */
-    private static int listIndex(int i, int size) {
+    public static int listIndex(int i, int size) {
         if (i >= size || i < -size) {
             throw new IndexOutOfBoundsException("Index " + i + " is out of bounds for a list with size " + size);
         }
--- a/logstash-core-event-java/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core-event-java/src/main/java/org/logstash/Accessors.java
@@ -34,14 +34,11 @@ public class Accessors {
                 return ((Map<String, Object>) target).remove(field.getKey());
             } else if (target instanceof List) {
                 int i = Integer.parseInt(field.getKey());
-                int size = ((List) target).size();
-                if (i >= size || i < -size) {
-                  return null;
-                } else if (i < 0) {
-                  // Offset from the end of the array.
-                  return ((List<Object>) target).remove(size + i);
-                } else {
-                  return ((List<Object>) target).remove(i);
+                try {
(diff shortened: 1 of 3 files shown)


jordansissel:
Travis appeared to fail for unrelated reasons (kafka)

talevy:
@jordansissel, we should have fixed this invalid integration test in master, mind rebasing?

jordansissel:
Rebased.

logstash-core-event-java/src/main/java/org/logstash/Accessors.java
```diff
@@ -34,10 +34,15 @@ public Object del(String reference) {
                 return ((Map<String, Object>) target).remove(field.getKey());
             } else if (target instanceof List) {
                 int i = Integer.parseInt(field.getKey());
-                if (i < 0 || i >= ((List) target).size()) {
-                    return null;
+                int size = ((List) target).size();
+                if (i >= size || i < -size) {
```
talevy: seems like there is a bunch of logic on what to do when `i > size`, I think we should add tests for this

jordansissel: Yeah, I was thinking of refactoring this to add a method that turns `i` into a list offset anyway, that way there's less branches for these bounds checks all over the place.

I can do that.

jordansissel:
I am planning to add a bunch of new java tests for this PR as well, btw.

logstash-core-event-java/src/main/java/org/logstash/Accessors.java
```diff
@@ -174,4 +180,18 @@ private boolean isCollection(Object target) {
     private ClassCastException newCollectionException(Object target) {
         return new ClassCastException("expecting List or Map, found "  + target.getClass());
     }
+
+    private static int listIndex(int i, List list) {
```
jordansissel: +1

logstash-core-event-java/src/test/java/org/logstash/AccessorsTest.java
```diff
@@ -207,4 +213,38 @@ public void testStaleTargetCache() throws Exception {
         assertEquals(accessors.get("[foo][bar]"), null);
         assertEquals(accessors.get("[foo]"), "boom");
     }
+
+    @RunWith(Theories.class)
+    public static class TestListIndexFailureCases {
+      private static final int size = 10;
+
+      @DataPoint
+      public static final int tooLarge = size;
```
jordansissel: I am experimenting with junit theories for these exception tests.

jordansissel:
@talevy I think you asked what happens on `-0`. `Integer.parseInt("-0") == 0` if that helps you. We don't (and probably shouldn't) have negative-zero mean something other than zero.

jordansissel:
Latest push adds test coverage for Accessors.listIndex

talevy:
LGTM

PR #6232, opened 2016-11-10
Add space in Log4j2

dedemorton:
Uses the correct name used on the Apache website plus fixes the search problem where searching on Log4j does not find the logging topic.


jordansissel:
LGTM

PR #6236, opened 2016-11-10
Add uptime millis stats

### LINKED ISSUE #6214: Add JVM uptime metrics
`_node/stats` should also report uptime in millis.

Example response:

```
"jvm" : {
        "timestamp" : 1478591320220,
        "uptime_in_millis" : 16472892,
        "mem" : {
          "heap_used_in_bytes" : 458343784,
          "heap_used_percent" : 22,
         ...
```

suyograo:
Fixes #6214

### DIFF
--- a/logstash-core/lib/logstash/api/commands/stats.rb
+++ b/logstash-core/lib/logstash/api/commands/stats.rb
@@ -8,8 +8,6 @@ module LogStash
     module Commands
       class Stats < Commands::Base
         def jvm
-          require "pry"
-          binding.pry
           {
             :threads => extract_metrics(
               [:jvm, :threads],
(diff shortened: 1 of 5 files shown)


ph:
Can you add a test for it in https://github.com/elastic/logstash/blob/master/logstash-core/spec/api/lib/api/node_stats_spec.rb 

After that it will LGTM.

suyograo:
@ph there's a test already: https://github.com/elastic/logstash/pull/6236/files#diff-bbdc5fb1e8fb55a0af749c56605c851eR17

ph:
@suyograo I would like to have a unit test for that new feature (in the file I have specified), you have added a rats test which is awesome, but its much slower to run and not everyone is running this test suite locally because of the requirement of building a package before been able to do the test.

ph:
errr nevermind
 I dont have enought coffee.

ph:
LGTM

Issue #6215, opened 2016-11-08
Add CPU load stats metrics


suyograo:
`_node/stats/os`

```
"os" : {
        "timestamp" : 1478591443335,
        "cpu" : {
          "percent" : 26,
          "load_average" : {
            "1m" : 2.826171875,
            "5m": 1.8261718,
            "15m": 1.56566
          }
        },

```

PR #6240, opened 2016-11-11
Monitor the load average of the machine and display it in the api

ph:
This PR add new information in the /_node/stats api and will return the
load average of the machine in the following formats depending of the
platforms that logstash is running on:

**Linux**
```json
{
    "cpu" : {
      "percent" : 26,
      "load_average" : {
        "1m" : 2.826171875,
        "5m": 1.8261718,
        "15m": 1.56566
      }
    }
}
```

**MacOS and other platform that the OperatingMXBean understand**
```json
{
    "cpu" : {
      "percent" : 26,
      "load_average" : {
        "1m" : 2.826171875,
      }
    }
}
```

Load average is not available on Windows

Fixes: https://github.com/elastic/logstash/issues/6215


logstash-core/spec/api/lib/api/node_stats_spec.rb
```diff
@@ -69,7 +69,8 @@
       },
       "cpu"=>{
         "total_in_millis"=>Numeric,
-        "percent"=>Numeric
+        "percent"=>Numeric,
+        "load_average" => { "1m" => Numeric }
```
suyograo: why only `1m`?

logstash-core/lib/logstash/instrument/periodic_poller/jvm.rb
```diff
@@ -33,11 +33,57 @@ def self.get(gc_name)
       end
     end
 
+    class LoadAverage
```
suyograo: Can we move this to a separate file? it's cleaner, IMO

ph: ++

logstash-core/spec/api/lib/api/node_stats_spec.rb
suyograo: @ph not worth it. this is good

ph:
@suyograo I've move the LoadAverage class into his own file, same with the specs.

suyograo:
LGTM

PR #6241, opened 2016-11-11
LIR UI (Read Only)

andrewvc:
A read-only java graph IR. This introduces a new `Compiler` class to compile logstash config languages. This is incremental work, but a good point for review + a merge into a feature branch.

**Notes for Reviewers:**

* The main entry point is `config.rb`
* You'll want to compile the java with `rake compile:all`

**Changes:**

* It also duplicates the treetop grammar into `lscl.treetop` and associated files because we need differently scoped class names.
* Extracted treetop monkey patches into a separate file shared by the old / new parser
* Added tests for many common config cases.

Extracted and cleaned up from #5938 .


logstash-core/spec/logstash/compiler/compiler_spec.rb
```diff
@@ -0,0 +1,554 @@
...
+
+        it "should contain the single input" do
+          expect(output).to ir_eql(j.iPlugin(OUTPUT, "stdout"))
+        end
+      end
+
+      it_should_behave_like "complex grammar", :output
+    end
+  end
+end
```
ph: I think theses test only check for the happy path, can we add some tests around parsing error to make sure we report the lines, columns correctly?

When we have multiples files, the current behavior of logstash is  to merge them and parse them, I am not sure if we test this scenario here.

logstash-core/src/main/java/org/logstash/config/ir/PluginDefinition.java
```diff
@@ -0,0 +1,88 @@
...
+            PluginDefinition oPluginDefinition = (PluginDefinition) o;
+
+            Set<String> allArgs = new HashSet<>();
+            allArgs.addAll(getArguments().keySet());
+            allArgs.addAll(oPluginDefinition.getArguments().keySet());
+
+            // Compare all arguments except the unique id
+            boolean argsMatch = allArgs.stream().
+                    filter(k -> !k.equals("id")).
+                    allMatch(k -> Objects.equals(getArguments().get(k), oPluginDefinition.getArguments().get(k)));
```
ph: refreshing to see this in java.

logstash-core/src/main/java/org/logstash/config/ir/expression/EventValueExpression.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "event.get('" + fieldName + "')";
+    }
+
+    @Override
+    public String toRubyString() {
+        return "event.getField('" + fieldName + "')";
```
ph: should it be `event.get("")`?

logstash-core/src/main/java/org/logstash/config/ir/graph/Edge.java
```diff
@@ -0,0 +1,76 @@
...
+
+        return edges;
+    }
+
+    public Edge(Vertex from, Vertex to) throws InvalidIRException {
+        this.from = from;
+        this.to = to;
+
+        if (this.from == this.to) {
+            throw new InvalidIRException("Cannot create a cyclic vertex!" + to);
```
ph: missing space between  "!" + to

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -0,0 +1,345 @@
+package org.logstash.config.ir.graph;
```
ph: Can we add junit test for this class? I dont see any in the test folder.

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -0,0 +1,345 @@
...
+    public Graph addVertex(Vertex v) {
+        this.vertices.add(v);
+        return this;
+    }
+
+    public void merge(Graph otherGraph) throws InvalidIRException {
+        this.vertices.addAll(otherGraph.getVertices());
+        this.edges.addAll(otherGraph.edges);
+        refresh();
+    }
```
ph: Should this method be fluent and return the current graph?

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -0,0 +1,345 @@
...
+
+        return this;
+    }
+
+    // Return plugin vertices by type
+    public List<PluginVertex> getPluginVertices(PluginDefinition.Type type) {
+       return pluginVertices()
+               .filter(v -> v.getPluginDefinition().getType().equals(type))
+               .collect(Collectors.toList());
+    }
```
ph: Would it be handy to also have an implementation that return a stream? like pluginVertices

logstash-core/src/main/java/org/logstash/config/ir/graph/Vertex.java
```diff
@@ -0,0 +1,104 @@
+package org.logstash.config.ir.graph;
```
ph: maybe we could have some junit test?

andrewvc:
Thanks for the feedback @ph! I've pushed some partial changes, but have plenty of work to do before this is ready for review again. Thanks for the excellent feedback :)

logstash-core/src/main/java/org/logstash/config/ir/expression/EventValueExpression.java
andrewvc: Hmmm, this was running correctly last I checked! The java event internally calls it `getField` in java, as opposed to `get` in ruby. Since the expression execution engine does not use the LogStash jruby environment it doesn't use the Ext wrapper. Maybe it should?

logstash-core/lib/logstash/compiler/lscl/lscl_grammar.treetop
```diff
@@ -0,0 +1,241 @@
...
+    selector_element+
+    <LogStash::Compiler::LSCL::AST::Selector>
+  end
+
+  rule selector_element
+    "[" [^\],]+ "]"
+    <LogStash::Compiler::LSCL::AST::SelectorElement>
+  end
+
+end
```
andrewvc: Correct! I only changed class names.

logstash-core/lib/logstash/compiler/lscl.rb
```diff
@@ -0,0 +1,555 @@
...
+      jdsl.eEventValue(source_meta, text_value)
+    end
+  end
+  
+  class SelectorElement < Node;
+    def expr
+      jdsl.eEventValue(source_meta, text_value)
+    end
+  end
+end; end; end; end; end;
```
andrewvc: Perhaps, but maybe in another PR? This greatly cleans up `config_ast.rb`. I'd prefer to not take on too much work at once. I'm still not sure how I'd organize the file if I broke it up.

ph: OK for another PR!

Well the previous treetop classes was also complex and hard to read I've hop that splitting them up would help.

logstash-core/src/main/java/org/logstash/config/ir/expression/EventValueExpression.java
ph: This is a really good question, since the classes are actually output ruby code I believe it should use the ruby extension to be consistent. If we were outputting java or other jvm languages It would make sense to use the java api. WDYT?

andrewvc: I mean, this is an internal usage hidden from users. The only reason I'd care about using the extension is because it *should* give a speed boost. Though we'd have to measure that of course.

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
andrewvc: Yes! I'm (trying) to follow the JDK8 convention of `getFoos -> Collection` and `foos -> Stream`. This is in violation of that!

logstash-core/lib/logstash/compiler/lscl.rb
```diff
@@ -0,0 +1,555 @@
...
+require "logstash/compiler/treetop_monkeypatches"
+java_import Java::OrgLogstashConfigIr::DSL
+java_import Java::OrgLogstashConfigIr::SourceMetadata
+
+module LogStashCompilerLSCLGrammar; module LogStash; module Compiler; module LSCL; module AST
+  # Helpers for parsing LSCL files
+  module Helpers
+    def source_meta
+      line, column = line_and_column
+      Java::OrgLogstashConfigIr::SourceMetadata.new(source_file, line, column, self.text_value)
```
andrewvc: Not just debug! This will be used to display the original files to users in all sorts of cases.

logstash-core/spec/logstash/compiler/compiler_spec.rb
andrewvc: @ph there is a test ` it "should attach source_metadata with correct info to the statements"` for source metadata. You're right about merged files. As-is, this code relies on the `Agent` to concatenate the files first. I'll add a test for multiple filter sections.

ph:
@andrewvc LGTM lets get that feature branch started.

Issue #3620, opened 2015-07-17
Docs: field reference should be documented in the "How to write a Logstash filter plugin"


EricDeveaud:
documentation on how to write a logstahs filter is clear and really helpful for  beginners 

there's one point that it is not covered is how to use the "%{fieldname}" syntax.

see: https://discuss.elastic.co/t/field-reference-syntax/25589/2

untergeek:
We love documentation improvements.  Where would you put it?  You can click the edit pencil icon and submit documentation changes from within the document itself.

We do mention the need for using sprintf formatting in our [Configuring Logstash](https://www.elastic.co/guide/en/logstash/current/configuration.html#sprintf) page.

EricDeveaud:
main concern in the plugin filter howto is which one from logstahs or from the plugin should take care of the sprintf formatting.

from my reading I thougth that "%{foo}" was interpreted//expanded//evaluated by logstash itself and then provided to the plugin as value.
Magnus Bäck pointed me that it should be the pluggin to take care of expansion. 

I'll try to edit the doc but with my english it may be more confusing than no doc ;-)

Issue #6220, opened 2016-11-08
docs: New plugin docs still use old (2.x) Event API calls


jordansissel:
![image](https://cloud.githubusercontent.com/assets/131818/20111048/a78ee2d6-a59b-11e6-8697-45403028ff30.png)

suyograo:
Fixed in https://github.com/elastic/logstash/pull/6242

PR #6242, opened 2016-11-12
Fix/how to plugins

suyograo:
Fixes https://github.com/elastic/logstash/issues/3620 and https://github.com/elastic/logstash/issues/6220


GlenRSmith:
LGTM

PR #6244, opened 2016-11-12
Remove old version of release notes

dedemorton:
These release notes are an old version, so I'm deleting them. There's no need for relnotes in 5.x because it's not a release branch.


suyograo:
LGTM

PR #6250, opened 2016-11-14
Add docs for Docker image

jarpy:
Provides documentation for running Logstash with the Elastic official Docker image.

### DIFF
--- a/docs/static/docker.asciidoc
+++ b/docs/static/docker.asciidoc
@@ -1,21 +1,22 @@
 [[docker]]
 == Running Logstash on Docker
-Docker images for Logstash are available from the Elastic Docker registry.
+Docker images for Logstash are available from the Elastic Docker
+registry.
 
-Obtaining Logstash for Docker is as simple as issuing a +docker pull+ command
-against the Elastic Docker registry.
+Obtaining Logstash for Docker is as simple as issuing a +docker pull+
+command against the Elastic Docker registry.
 
 ifeval::["{release-state}"=="unreleased"]
 
-However, version {logstash_version} of Logstash has not yet been released, so no Docker
-image is currently available for this version.
+However, version {logstash_version} of Logstash has not yet been
+released, so no Docker image is currently available for this version.
 
 endif::[]
 
 ifeval::["{release-state}"!="unreleased"]
 
-The Docker image for Logstash v{logstash_version} can be retrieved with the following
-command:
+The Docker image for Logstash {logstash_version} can be retrieved with
+the following command:
 
 ["source","sh",subs="attributes"]
 --------------------------------------------
@@ -31,12 +32,12 @@ Logstash differentiates between two types of configuration:
 
 ==== Pipeline Configuration
 
-It is essential to place your pipeline configuration where it can be found by
-Logstash. By default, the container will look in
+It is essential to place your pipeline configuration where it can be
+found by Logstash. By default, the container will look in
 +/usr/share/logstash/pipeline/+ for pipeline configuration files.
 
-In this example we use a bind-mounted volume to provide the configuration via
-the +docker run+ command:
+In this example we use a bind-mounted volume to provide the
+configuration via the +docker run+ command:
 
 ["source","sh",subs="attributes"]
 --------------------------------------------
@@ -46,10 +47,11 @@ docker run --rm -it -v ~/pipeline/:/usr/share/logstash/pipeline/ {docker-image}
 Every file in the host directory +~/pipeline/+ will then be parsed
 by Logstash as pipeline configuration.
 
-If you don't provide configuration to Logstash, it will run with a minimal
-config that listens for messages from the
-<<plugins-inputs-beats,Beats input plugin>> and echoes any that are received
-to `stdout`. In this case, the startup logs will be similar to the following:
+If you don't provide configuration to Logstash, it will run with a
+minimal config that listens for messages from the
+<<plugins-inputs-beats,Beats input plugin>> and echoes any that are
+received to `stdout`. In this case, the startup logs will be similar
+to the following:
 
 ["source","text"]
 --------------------------------------------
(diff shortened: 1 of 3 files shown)


suyograo:
@jarpy should we also add location of logstash.yml and pipeline config files as a new section here? https://github.com/elastic/logstash/blob/master/docs/static/setting-up-logstash.asciidoc#dir-layout

docs/static/docker.asciidoc
```diff
@@ -0,0 +1,108 @@
...
+ifeval::["{release-state}"=="unreleased"]
+
+However, version {logstash_version} of Logstash has not yet been released, so no Docker
+image is currently available for this version.
+
+endif::[]
+
+ifeval::["{release-state}"!="unreleased"]
+
+The Docker image for Logstash v{logstash_version} can be retrieved with the following
```
dedemorton: Remove the "v" before v{logstash_version} (we don't use the convention of prefacing version numbers with a "v".

docs/static/docker.asciidoc
```diff
@@ -0,0 +1,108 @@
...
+--------------------------------------------
+Sending Logstash logs to /usr/share/logstash/logs which is now configured via log4j2.properties.
+[2016-10-26T05:11:34,992][INFO ][logstash.inputs.beats    ] Beats inputs: Starting input listener {:address=>"0.0.0.0:5044"}
+[2016-10-26T05:11:35,068][INFO ][logstash.pipeline        ] Starting pipeline {"id"=>"main", "pipeline.workers"=>4, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>500}
+[2016-10-26T05:11:35,078][INFO ][org.logstash.beats.Server] Starting server on port: 5044
+[2016-10-26T05:11:35,078][INFO ][logstash.pipeline        ] Pipeline main started
+[2016-10-26T05:11:35,105][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
+--------------------------------------------
+
+This configuration is baked into the image at +/usr/share/logstash/pipeline/logstash.conf+.
```
dedemorton: By "baked into the image" do you mean that it's the default config? If so, I would say this. Not sure "baked into the image" will translate very well when we start getting the docs translated....

docs/static/getting-started-with-logstash.asciidoc
```diff
@@ -144,6 +144,12 @@ WARNING: The repositories do not work with older rpm based distributions
 
 See the <<running-logstash,Running Logstash>> document for managing Logstash as a system service.
```
dedemorton: To be consistent with how we introduce xrefs, I would say:  `"See <<running-logstash,Running Logstash>> for details about managing Logstash as a system service."` The link formatting makes it pretty clear that you're jumping to another page.

Now I have to admit that we're not always consistent. :-)

docs/static/getting-started-with-logstash.asciidoc
```diff
@@ -144,6 +144,12 @@ WARNING: The repositories do not work with older rpm based distributions
...
 
+=== Docker
+
+An image is available for running Logstash as a Docker container. It is
+available from the Elastic Docker registry. See the <<docker>> section for
+details on how to configure and run Logstash Docker containers.
```
dedemorton: Same comment as before: Would say, `See <<docker>> for
details about how to configure and run Logstash Docker containers.`

dedemorton:
@jarpy The writing looks pretty clean...just made a few minor suggestions.

jarpy:
Thanks @dedemorton and @suyograo for the review. 

- Added directory layout section. Great idea: not having this already created confusion for a user: elastic/logstash-docker#5
- Made all your suggested fixes, @dedemorton. Thanks again.
- Wrapped a bunch of copy at 78 columns.
- Added `5.0.1` tag since this documentation is for functionality already available in 5.0.0. There's no functional change since then.

docs/static/docker.asciidoc
```diff
@@ -0,0 +1,114 @@
...
+
+Settings files can also be provided through bind-mounts. Logstash
+expects to find them at +/usr/share/logstash/config/+.
+
+It's possible to provide an entire directory containing all needed
+files:
+
+["source","sh",subs="attributes"]
+--------------------------------------------
+docker run --rm -it -v ~/settings/:/usr/share/logstash/config/ {docker-image}
```
dliappis: @jarpy It is worth noting that with bind-mounts the mounted dir will have the ownership and permissions of the host dir. See for example the discussion here: https://discuss.elastic.co/t/elastic-elasticsearch-docker-not-assigning-permissions-to-data-directory-on-run/65812/2

Since logstash doesn't require write access to those files it's not a major problem, but I still suggest having a warning that the files should be readable by the `logstash` user in the Docker image.

docs/static/docker.asciidoc
```diff
@@ -0,0 +1,114 @@
...
+Bind-mounted configuration is not the only option, naturally. If you
+prefer the _Immutable Infrastructure_ approach, you can prepare a
+custom image containing your configuration by using a +Dockerfile+
+like this one:
+
+["source","dockerfile",subs="attributes"]
+--------------------------------------------
+FROM {docker-image}
+RUN rm -f /usr/share/logstash/pipeline/logstash.conf
+ADD pipeline/ /usr/share/logstash/pipeline/
```
dliappis: Same comment for here as before; in ES for example we recommend chown'ing after the ADD operation: https://github.com/dliappis/elasticsearch/blame/d7646bdbb166a0781130f558637da93214c3ec23/docs/reference/setup/install/docker.asciidoc#L246-L248

jarpy: I don't know. It doubles the size of the example by filling it with things that aren't directly relevant to the point.

```
FROM docker.elastic.co/logstash/logstash:5.0.1
RUN rm -f /usr/share/logstash/pipeline/logstash.conf
ADD pipeline/ /usr/share/logstash/pipeline/
ADD config/ /usr/share/logstash/config/
USER root
RUN chown -R logstash:logstash /usr/share/logstash/pipeline/ /usr/share/logstash/config/
RUN chmod 0755 /usr/share/logstash/pipeline/ /usr/share/logstash/config/
RUN chmod 0444 /usr/share/logstash/pipeline/* /usr/share/logstash/config/*
USER logstash
```

What a mess! It obscures the topic of "how to configure Logstash on Docker" with a discussion of "how to manage file ownership and modes in Docker". 

It forces you to switch user (twice) and become very explicit about file permissions, because you now have to be very sure that the `logstash` user _doesn't_ have write access to the files. Logstash is, among other things, capable of [executing arbitrary code](https://www.elastic.co/guide/en/logstash/current/plugins-filters-ruby.html#plugins-filters-ruby-code), so giving the process write access to its own files seems like something you want to avoid. A simple `chown` therefore, is insufficient. If we teach `chown`, we have to teach `chmod` and `USER` too. Maybe all this belongs in a dedicated "Securing Logstash" guide, but I don't think it belongs in the Docker documentation.

docs/static/docker.asciidoc
jarpy: Sure. :) Added a note.

https://github.com/elastic/logstash/pull/6250/commits/2bd4d8e62c0455f126e911c1862916a5fc6f6762

I think we should be careful about teaching people Docker (and Unix) though. As a reader, I would feel like the author was wasting my time if they told me that the configuration files for a process should be readable by the process. _You don't say, author!_

docs/static/docker.asciidoc
```diff
@@ -89,6 +89,11 @@ Alternatively, a single file can be mounted:
 docker run --rm -it -v ~/settings/logstash.yml:/usr/share/logstash/config/logstash.yml {docker-image}
 --------------------------------------------
 
+NOTE: Bind-mounted configuration files will retain the same permissions and
+ownership within the container that they have on the host system. Be sure
+to set persmissions such that the files will be readable and, ideally, not
```
jarpy: Typo fixed in 80281cfe3ca8674e6619d57b7b7bc1d359f4dded.

suyograo:
@dliappis are you good with this? Looking good to me.

dliappis:
@suyograo Yes this is fine. :shipit: !

PR #6257, opened 2016-11-15
Truncate from the end of the message instead of the front.

### LINKED ISSUE #6255: Logger calls truncate the message to 10000 bytes
Discovered in https://github.com/logstash-plugins/logstash-filter-mutate/issues/85

Reproducing (from master):

```
% bundle exec bin/logstash -e 'input { generator { count => 1 } } filter { ruby { init => "logger.error(:HELLO.to_s + (:X.to_s * 9998))"  code => "" } }'
...
```

```
% grep 'logstash.filters.ruby' logs/logstash-plain.log | cut -b 1-100
[2016-11-13T23:16:27,501][ERROR][logstash.filters.ruby    ] LOXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX

```

As you can see, the "HELLO" is cut off (only the last 2 letters, "LO" are shown).

jordansissel:
PatternLayout calls this 'left justify' I think.

Fixes #6255

### DIFF
--- a/config/log4j2.properties
+++ b/config/log4j2.properties
@@ -21,7 +21,7 @@ appender.rolling.policies.time.type = TimeBasedTriggeringPolicy
 appender.rolling.policies.time.interval = 1
 appender.rolling.policies.time.modulate = true
 appender.rolling.layout.type = PatternLayout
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

PR #6259, opened 2016-11-16
Fix filter 'id' naming to be consistent with inputs / outputs

andrewvc:
The filters were the only category prefixing IDs with their plugin names


suyograo:
LGTM

suyograo:
this integration test is failing on master as well after @talevy https://github.com/elastic/logstash/commit/6e01395dde7cb59551829de85f3d90daa35626a3 commit. I'll check this locally to see if it's legit.

PR #6268, opened 2016-11-17
Theses core input are not currently used inside logstash so it make sense to remove them

ph:
The core metric input is not currently used, it will come back as an independent plugin.


jsvd:
LGTM

PR #6271, opened 2016-11-17
Feature: Paquet, a tool to pack dependencies

ph:
This tool allows to generate a logstash pack of plugin, and it is required
by #6168.

This allows the developer to specify which gems it needs to be included in
the pack, it will also take care to get the parent dependencies.

```ruby
require "paquet"

TARGET_DIRECTORY = File.join(File.dirname(__FILE__), "dependencies")

Paquet::Task.new(TARGET_DIRECTORY) do
  pack "manticore"
  pack "gemoji"
  pack "logstash-output-elasticsearch"

  # Ignore some dependencies that the core will provide
  ignore "logstash-core-plugin-api"
  ignore "logstash-core"
end
```


suyograo:
@ph can you rebase, this tests are fixed in master.

tools/paquet/lib/paquet/rspec/tasks.rb
```diff
@@ -0,0 +1,147 @@
+# encoding: utf-8
+require "bundler"
+require "rake"
+require "rake/tasklib"
+require "fileutils"
+require "net/http"
+
+#
```
suyograo: Can we add more comments here about the purpose of this class/tool?

tools/paquet/lib/paquet/rspec/tasks.rb
```diff
@@ -0,0 +1,147 @@
...
+    def add(name)
+      @gems << name
+    end
+
+    def ignore(name)
+      @ignores << name
+    end
+
+    def pack
+      FileUtils.rm_rf(@target_path)
```
suyograo: Can we integrate with the logger and log that we are doing this?

tools/paquet/lib/paquet/rspec/tasks.rb
```diff
@@ -0,0 +1,147 @@
...
+      FileUtils.mkdir_p(@target_path)
+
+      # need to get the current version and dependencies
+      required_gems = collect_required_gems
```
suyograo: maybe remove this line and directly do `download_gems(collect_required_gems)`

tools/paquet/lib/paquet/rspec/tasks.rb
```diff
@@ -0,0 +1,147 @@
...
+        .each do |name|
+          source = "#{RUBYGEMS_URI}/#{name}"
+          destination = File.join(@target_path, name)
+
+          puts "Vendoring: #{name}, downloading: #{source}"
+          download_file(source, destination)
+      end
+    end
+
+    def download_file(source, destination, counter = 10)
```
suyograo: Can we move this to a utils in core or devutils?

jsvd:
@ph have you thought of creating this as separate repo? we don't seem to depend on anything that is logstash related

tools/paquet/lib/paquet/rspec/tasks.rb
```diff
@@ -0,0 +1,147 @@
...
+        http = Net::HTTP.new(uri.host, uri.port)
+        http.use_ssl = true
+        response = http.get(uri.path)
+
+        case response
+        when Net::HTTPSuccess
+          f.write(response.body)
+        when Net::HTTPRedirection
+          download_file(response['location'], destination, counter)
+        end
```
jsvd: let's raise some issue on on the `else` here

tools/paquet/paquet.gemspec
```diff
@@ -0,0 +1,28 @@
...
+  spec.files         = `git ls-files -z`.split("\x0").reject do |f|
+    f.match(%r{^(test|spec|features)/})
+  end
+  spec.bindir        = "exe"
+  spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
+  spec.require_paths = ["lib"]
+
+  spec.add_development_dependency "bundler", "~> 1.13"
+  spec.add_development_dependency "rake", "~> 10.0"
+  spec.add_runtime_dependency "pry"
```
jsvd: debug leftovers?

tools/paquet/paquet.gemspec
```diff
@@ -0,0 +1,28 @@
...
+  spec.authors       = ["Elastic"]
+  spec.email         = ["info@elastic.co"]
+  spec.license       = "Apache License (2.0)"
+
+  spec.summary       = %q{Rake helpers to create a uber gem}
+  spec.description   = %q{This gem add a few rake tasks to create a uber gems that will be shipped as a zip}
+  spec.homepage      = "https://github.com/elastic/paquet"
+
+
+  spec.files         = `git ls-files -z`.split("\x0").reject do |f|
```
jsvd: since the number of files and directories here is quite small and using exec'ing out is also usually bad, I suggest manually listing the files here.

tools/paquet/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+    $ gem install paquet
+
+## Usage
+
+TODO: Write usage instructions here
+
+## Development
+
+After checking out the repo, run `bin/setup` to install dependencies. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
```
jsvd: where are these binaries coming from?

ph:
@jsvd 
> have you thought of creating this as separate repo? we don't seem to depend on anything that is logstash related

I've create a separate repo at the beginning, but after discussing with @suyograo we both agree to make it a part of the main repo, so the code is easier to track/release.

jsvd:
Fair enough, I don't mean to revisit past discussions, I know having devutils separated has created unnecessary troubles, I only suggested this since, in theory, anyone could use this outside of Logstash to package gems.

tools/paquet/lib/paquet/rspec/tasks.rb
ph: This is not the same as the download method in devutils, this one support HTTP redirection and I would prefer to keep the external dependency as little as possible.

tools/paquet/lib/paquet/rspec/tasks.rb
ph: +1

tools/paquet/lib/paquet/rspec/tasks.rb
ph: I don't currently have a logger in that tool for now and I would prefer not to use cabin to make sure we don't run into weird dependency problems with the package.

But I will provide feedback on that operation.

tools/paquet/README.md
ph: oups default doc from bundle gem.

tools/paquet/lib/paquet/rspec/tasks.rb
suyograo: sounds good

ph:
I've commit the changes from the review, added units test and an integration test to test the packaging. Also, I have reorganized the code a bit to separate every classes.

tools/paquet/.gitignore
```diff
@@ -0,0 +1,6 @@
+/.bundle/
```
jsvd: what happens when the first character is `/`? I assume we're not hoping to see a `.bundle` at the root of the filesystem?

tools/paquet/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+And run
+
+```
+bundle exec rake paquet:vendor
+```
+
+The dependencies will be downloaded in your target directory.
+
+## Project Principles
```
jsvd: <3

jsvd:
I only have a question about the .gitignore file, otherwise LGTM

tools/paquet/.gitignore
ph: usually path in the gitignore are relatives, so it should be a problem?

Issue #3786, opened 2015-08-25
Replace message queue with message broker


markwalkom:
It's minor but we seem to be recommending MQ rather than a generic broker [here](https://www.elastic.co/guide/en/logstash/current/deploying-and-scaling.html).

We should update this :)

dedemorton:
Fixed in #6277

PR #6277, opened 2016-11-17
Change wording so it doesn't sound like we are recommending message queues

dedemorton:
Fixes https://github.com/elastic/logstash/issues/3786


suyograo:
LGTM

PR #6282, opened 2016-11-19
Update links and remove old blog links from LS intro

dedemorton:
We had some old links (my bad) and some links that pointed to very old blogs. 

I'm going to open a separate issue to investigate whether we want to update the blogs or pull the content into the docs.


suyograo:
Thanks for catching this. LGTM

Issue #6285, opened 2016-11-21
Beats integrations test are failing


ph:
The current FB integration test are failling,  see : https://travis-ci.org/elastic/logstash
I am able to reproduce the problem locally, when I start FB with the generated configuration file it exit with EOF.

````
stash/qa/integration/services/installed/filebeat/logs]
2016/11/21 19:09:06.013868 beat.go:174: INFO Setup Beat: filebeat; Version: 5.0.1
2016/11/21 19:09:06.013941 logstash.go:90: INFO Max Retries set to: 3
2016/11/21 19:09:06.014057 outputs.go:106: INFO Activated logstash as output plugin.
2016/11/21 19:09:06.014183 publish.go:291: INFO Publisher name: sashimi
2016/11/21 19:09:06.014803 async.go:63: INFO Flush Interval set to: 1s
2016/11/21 19:09:06.014829 async.go:64: INFO Max Bulk Size set to: 2048
2016/11/21 19:09:06.015483 logp.go:219: INFO Metrics logging every 30s
2016/11/21 19:09:06.015687 beat.go:204: INFO filebeat start running.
2016/11/21 19:09:06.015730 registrar.go:66: INFO Registry file set to: /var/folders/yl/9hpq_qs528g60cv9trnt5z180000gn/T/studtmp-f2daa635d65b43699197b703e78da0634dc776b64e13e102dc885804196d
2016/11/21 19:09:06.015781 registrar.go:99: INFO Loading registrar data from /var/folders/yl/9hpq_qs528g60cv9trnt5z180000gn/T/studtmp-f2daa635d65b43699197b703e78da0634dc776b64e13e102dc885804196d
2016/11/21 19:09:06.015813 registrar.go:111: ERR Error decoding states: EOF
2016/11/21 19:09:06.015826 registrar.go:200: ERR Error loading state: EOF
2016/11/21 19:09:06.015833 filebeat.go:93: ERR Could not start registrar: EOF
2016/11/21 19:09:06.015875 logp.go:245: INFO Total non-zero values:
2016/11/21 19:09:06.015886 logp.go:246: INFO Uptime: 5.642798ms
2016/11/21 19:09:06.015891 beat.go:208: INFO filebeat stopped.
2016/11/21 19:09:06.015898 beat.go:285: CRIT Exiting: EOF
Exiting: EOF
```

ph:
This seems to be related when we configure a custom `registry_file`, I've tried using `/dev/null` and a temporary file and it fails.

PR #6286, opened 2016-11-21
use a temporary path instead of a temporary file

ph:
Using an empty file for the registry make FB quits with a EOF error.

Fixes: #6285


suyograo:
LGTM

PR #6287, opened 2016-11-22
Fix heading levels for docker topics

dedemorton:
Currently the docker heading levels mess up the structure of the TOC. Running Lostash on Docker appears as a top level, and topics that aren't related to docker end up nested underneath:

dockertoc

This PR fixes the heading levels so that the TOC looks like this:
dockertocfixed


suyograo:
nice. LGTM

PR #6288, opened 2016-11-22
Allow for exception instances to get serialized in JSON logging

### LINKED ISSUE #6082: Monkey patch Exception to provide a json_data method
At the moment we add the exception object to the hash when logging an exception. This breaks JSON logging.

If we add a json_data method to the Exception class that returns a compact string of the class and message, we will prevent these cases from throwing a JSON generation error during exception handling logging.

guyboertje:
Fixes #6082


ph:
LGTM, lets backport that to 2.4.x

suyograo:
Merged in https://github.com/elastic/logstash/commit/eba128c96861dd672574795526c68ae7c280f9e6

suyograo:
@guyboertje can you backport this to 2.4 branch?

Issue #3920, opened 2015-09-14
Configuration File Ordering Does Matter for some reason


ayashjorden:
Hi all,
This issue is the result of [this](https://discuss.elastic.co/t/logstash-configuration-file-ordering-does-matter/28840) discussion. Per @magnusbaeck [suggestion](https://discuss.elastic.co/t/logstash-configuration-file-ordering-does-matter/28840/6), I'm opening this issue so we all can benefit and understand Logstash better.
I'm running logstash with a couple of `-f` switches and for some unknown reason, the order of the files does matter:

```
input-filter-output:
~/util/logstash-1.5.4/bin/logstash -f ./conf.d/input-50-email-sending.conf -f ./conf.d/filter-50-email-sending.conf -f ./conf.d/output-50-receiver.conf
Logstash startup completed
Logstash shutdown completed

input-output-filter:
~/util/logstash-1.5.4/bin/logstash -f ./conf.d/input-50-email-sending.conf -f ./conf.d/output-50-receiver.conf -f ./conf.d/filter-50-email-sending.conf
Logstash startup completed
Logstash shutdown completed

filter-output-input:
~/util/logstash-1.5.4/bin/logstash -f ./conf.d/filter-50-email-sending.conf -f ./conf.d/output-50-receiver.conf -f ./conf.d/input-50-email-sending.conf
Logstash startup completed
#gave is some time to run...
^CSIGINT received. Shutting down the pipeline. {:level=>:warn}
Logstash shutdown completed
```

With --debug, I found the following:

```
input-filter-output:
Plugin not defined in namespace, checking for plugin file {:type=>"output", :name=>"http", :path=>"logstash/outputs/http", :level=>:debug, :file=>"logstash/plugin.rb", :line=>"133", :method=>"lookup"}
Plugin not defined in namespace, checking for plugin file {:type=>"codec", :name=>"plain", :path=>"logstash/codecs/plain", :level=>:debug, :file=>"logstash/plugin.rb", :line=>"133", :method=>"lookup"}

input-output-filter:
Plugin not defined in namespace, checking for plugin file {:type=>"filter", :name=>"csv", :path=>"logstash/filters/csv", :level=>:debug, :file=>"logstash/plugin.rb", :line=>"133", :method=>"lookup"}

filter-output-input: Logstash starts and stays up:
[...]

PR #6289, opened 2016-11-22
Clarify -f description and add info about supported glob patterns

dedemorton:
Fixes issue https://github.com/elastic/logstash/issues/3920.

Also added a topic about supported glob patterns (might help users understand that they can use pattern matching for more control over specifying config files).

### DIFF
--- a/docs/static/glob-support.asciidoc
+++ b/docs/static/glob-support.asciidoc
@@ -35,16 +35,16 @@ as part of a glob. The pattern `c:\foo*` will not work, so use `foo*` instead.
 
 Here are some common examples of glob patterns: 
 
-*`"/path/to/*.conf"`*::
+`"/path/to/*.conf"`::
 Matches config files ending in `.conf` in the specified path.
 
-*`"/var/log/*.log"`*::
+`"/var/log/*.log"`::
 Matches log files ending in `.log` in the specified path.
 
-*`"/var/log/**/*.log`*::
+`"/var/log/**/*.log`::
 Matches log files ending in `.log` in subdirectories under the specified path.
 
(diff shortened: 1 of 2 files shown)


jordansissel:
Some example ideas:

* Config files: "/path/to/*.conf"
* Logs in a single directory (file input?): "/var/log/*.log"
* ANything ending in `.log` under a given directory: "/var/log/**/*.log"
* A few specific file patterns (maybe individual app logs?): "/path/to/logs/{app1,app2,app3}/data.log"

dedemorton:
@jordansissel OK, I've added your examples. Decided to add a separate section of examples because it made the reference stuff easier to read. Let me know if you think of any other examples to add.

docs/static/glob-support.asciidoc
```diff
@@ -0,0 +1,50 @@
...
+
+Here are some common examples of glob patterns: 
+
+*`"/path/to/*.conf"`*::
+Matches config files ending in `.conf` in the specified path.
+
+*`"/var/log/*.log"`*::
+Matches log files ending in `.log` in the specified path.
+
+*`"/var/log/**/*.log`*::
```
jordansissel: Github's rendering of this looks... odd. Is this expected? 
screenshot from 2016-11-28 13-57-33

dedemorton: @jordansissel Yah, it's sort of a display problem with GitHub because the tagging renders correctly when I build the doc. However, I don't think the examples actually need to be bolded, so I'll remove the bold formatting to make the display in GitHub more reasonable.

jordansissel:
LGTM!

PR #6290, opened 2016-11-22
feature/ruby env

colinsurprenant:
- adds `bin/ruby` command to launch ruby scripts using the logstash ruby launcher
- modifies `bin/cpdump` to use `bin/ruby` instead of hardcoded vendorized jruby

### DIFF
--- a/bin/ruby
+++ b/bin/ruby
@@ -1,0 +1,24 @@
+#!/bin/sh
+# Run a ruby script using the logstash jruby launcher
+#
+# Usage:
+#   bin/ruby [arguments]
+#
+# Supported environment variables:
+#   LS_JVM_OPTS="xxx" path to file with JVM options
+#   LS_JAVA_OPTS="xxx" to append extra options to the defaults JAVA_OPTS provided by logstash
+#   JAVA_OPTS="xxx" to *completely override* the defauls set of JAVA_OPTS provided by logstash
+#
+# Development environment variables:
+#   USE_RUBY=1 to force use the local "ruby" command to launch logstash instead of using the vendored JRuby
+#   DEBUG=1 to output debugging information
+
+# use faster starting JRuby options see https://github.com/jruby/jruby/wiki/Improving-startup-time
+export JRUBY_OPTS="$JRUBY_OPTS -J-XX:+TieredCompilation -J-XX:TieredStopAtLevel=1"
+
+unset CDPATH
+
+. "$(cd `dirname $0`/..; pwd)/bin/logstash.lib.sh"
+setup
+
+ruby_exec "$@"
--- a/bin/cpdump
+++ b/bin/cpdump
@@ -1,7 +1,7 @@
-#!/usr/bin/env vendor/jruby/bin/jruby
+#!/usr/bin/env bin/ruby
 
 require_relative "../lib/bootstrap/environment"
-LogStash::Bundler.setup!({:without => [:build]})
+LogStash::Bundler.setup!({:without => [:build, :development]})
 require "logstash-core"
 require "logstash/environment"
 require "logstash/settings"


bin/ruby
```diff
@@ -0,0 +1,49 @@
...
+
+# sourcepath discovery below is the same as in bin/logstash
+
+unset CDPATH
+# This unwieldy bit of scripting is to try to catch instances where Logstash
+# was launched from a symlink, rather than a full path to the Logstash binary
+if [ -L $0 ]; then
+  # Launched from a symlink
+  # --Test for the readlink binary
+  RL=$(which readlink)
```
jordansissel: These variables and assignments need quotes:

`RL="$(...)"`

bin/ruby
```diff
@@ -0,0 +1,49 @@
...
+
+# use faster starting JRuby options see https://github.com/jruby/jruby/wiki/Improving-startup-time
+export JRUBY_OPTS="$JRUBY_OPTS -J-XX:+TieredCompilation -J-XX:TieredStopAtLevel=1"
+
+# sourcepath discovery below is the same as in bin/logstash
+
+unset CDPATH
+# This unwieldy bit of scripting is to try to catch instances where Logstash
+# was launched from a symlink, rather than a full path to the Logstash binary
+if [ -L $0 ]; then
```
jordansissel: Quote variable `if [ -L "$0" ]; then`

bin/ruby
```diff
@@ -0,0 +1,49 @@
...
+  # Launched from a symlink
+  # --Test for the readlink binary
+  RL=$(which readlink)
+  if [ $? -eq 0 ]; then
+    # readlink exists
+    SOURCEPATH=$($RL $0)
```
jordansissel: quotes `"$(...)"`

bin/ruby
```diff
@@ -0,0 +1,49 @@
...
+  else
+    # readlink not found, attempt to parse the output of stat
+    SOURCEPATH=$(stat -c %N $0 | awk '{print $3}' | sed -e 's/\‘//' -e 's/\’//')
```
jordansissel: quotes `SOURCEPATH="$(...)"`

bin/ruby
```diff
@@ -0,0 +1,49 @@
...
+    if [ $? -ne 0 ]; then
+      # Failed to execute or parse stat
+      echo "Failed to find source library at path $(cd `dirname $0`/..; pwd)/bin/logstash.lib.sh"
+      echo "You may need to launch Logstash with a full path instead of a symlink."
+      exit 1
+    fi
+  fi
+else
+  # Not a symlink
+  SOURCEPATH=$0
```
jordansissel: Quotes

colinsurprenant:
@jordansissel this bit of shell code was copied from `bin/logstash.lib.sh`  - so I suggest we apply your suggestions to both or leave this one as-is for now since we know it is working and open a new issue to address these on both sides. We could/should also try to DRY it?

colinsurprenant:
Bump?

jordansissel:
the quotation comments I made are to fix bugs, not for style, just in case
I didn't communicate that. If the code you copied doesn't quote also we
should fix that but in a separate PR.

I don't want us to accept bugs just because we copied buggy code.

colinsurprenant:
@jordansissel thanks, I understand that - but what I am saying is that this code is exactly the same code that is currently used in `bin/logstash.lib.sh` which is used to launch LS and I think we can say that  it seems to be reliable. For me it does not make sense to "fix" this here without being consistent in `bin/logstash.lib.sh`.

Also, I do not exactly understand the problematic use-case the fixes you are proposing will solve? Unless these problems are actual blockers for this script then I'll go ahead but otherwise, since this is pretty much untested code, it is hard to say what the effect will be, without doing some manual testing - and this is where I rely on the fact that it "seems" to have been working well for `bin/logstash` this far.

So what I suggest is this: I will open a new PR which will propose the changes you are requesting here, but for both this file and also `bin/logstash.lib.sh`. Better yet, I will try to also DRY that code so that there is only one instance of that code reused in both scripts.

Would that work for you?

bin/ruby
jordansissel: we check if $0 here is a symlink, but if the path contains a space, it fails:

```
% "/tmp/hello world/bin/ruby" --version
/tmp/hello world/bin/ruby: line 24: [: /tmp/hello: binary operator expected
```

This is what I refer to with respect to quoting. If you don't quote your variables, any value with a space will be buggy.

jordansissel:
@colinsurprenant I hear you. Let's do this. Let's remove lines 24-44 (the symlink check). Then, in a future PR, we can fix the quoting problems you've identified in the other shell scripts and introduce correct symlink checks into this code after that.

This way, we remove the bugs (quoting problems) and what remains is a feature we know we want to add but is out of scope for this PR.

colinsurprenant:
@jordansissel removed symlink check per discussion!

jordansissel:
LGTM!

colinsurprenant:
rebased and merged into master and 5.x

PR #6294, opened 2016-11-23
update 5.x branch version to 5.2.0

jsvd:
this is the outcome of:

1. running `rake version:set[5.2.0]`
2. manually editing the version of the `logstash-core-queue-jruby/lib/logstash-core-queue-jruby/version.rb`
3. manually pinning logstash kafka plugins to major version 5 in the `Gemfile`
4. running `rake bootstrap`


ph:
LGTM

colinsurprenant:
For my education, why are we pinning the kafka plugin to `~> 5` ?

suyograo:
@colinsurprenant Kafka plugin version 6.0 is not backward compatible with the previous version. This is not a Logstash issue per se, but Kafka plugin version 5.0 tracks 0.10.0.1 of Kafka, while Kafka plugin 6.0 tracks 0.10.1. The clients are not compatible across those versions which means we can't package 6.0 by default.

PR #6295, opened 2016-11-23
Fix bad callout formatting in doc

dedemorton:
Fixes https://github.com/elastic/website-www.elastic.co/issues/1370 (the incorrectly formatted callouts were resulting in additional numbers added to the next topic that contained callouts).


suyograo:
LGTM

PR #6297, opened 2016-11-24
introduce settting to limit byte size of queue 

jsvd:
introduce a setting to define the maximum size of the Queue in bytes.

The current strategy involves keeping a counter of the queue size by:

* bootstrapping the counter during queue initialization by going through each tail page and add its capacity to the counter
* every new write adds the byte count to the counter
* `isFull()` has an extra boolean check that compares current and max size for the queue

TODO:

- [x] account for the purge of pages (decrementing currentSize)
- [x] add ruby code to expose the config setting
- [x] test the unblocking of writes once pages are purged
- [ ] add currentSize of queue as a logstash metric (? to be determined if done in this PR)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
@@ -9,7 +9,7 @@ public class FileSettings implements Settings {
     private PageIOFactory pageIOFactory;
     private Class elementClass;
     private int capacity;
-    private long queueMaxSizeInBytes;
+    private long queueMaxBytes;
     private int maxUnread;
     private int checkpointMaxAcks;
     private int checkpointMaxWrites;
@@ -44,8 +44,8 @@ public class FileSettings implements Settings {
     }
 
     @Override
-    public Settings setQueueMaxSizeInBytes(long size) {
-        this.queueMaxSizeInBytes = size;
+    public Settings setQueueMaxBytes(long size) {
+        this.queueMaxBytes = size;
         return this;
     }
 
@@ -114,7 +114,7 @@ public class FileSettings implements Settings {
     }
 
     @Override
-    public long getQueueMaxSizeInBytes() { return queueMaxSizeInBytes; }
+    public long getQueueMaxBytes() { return queueMaxBytes; }
 
     @Override
     public int getCapacity() {
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
@@ -9,7 +9,7 @@ public class FileSettings implements Settings {
     private PageIOFactory pageIOFactory;
     private Class elementClass;
     private int capacity;
-    private int queueMaxSizeInBytes;
+    private long queueMaxSizeInBytes;
     private int maxUnread;
     private int checkpointMaxAcks;
     private int checkpointMaxWrites;
@@ -44,7 +44,7 @@ public class FileSettings implements Settings {
     }
 
     @Override
-    public Settings setQueueMaxSizeInBytes(int size) {
+    public Settings setQueueMaxSizeInBytes(long size) {
         this.queueMaxSizeInBytes = size;
         return this;
     }
@@ -114,7 +114,7 @@ public class FileSettings implements Settings {
     }
 
     @Override
-    public int getQueueMaxSizeInBytes() { return queueMaxSizeInBytes; }
+    public long getQueueMaxSizeInBytes() { return queueMaxSizeInBytes; }
 
     @Override
     public int getCapacity() {
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
@@ -114,7 +114,9 @@ public class FileSettings implements Settings {
     }
 
     @Override
-    public long getQueueMaxBytes() { return queueMaxBytes; }
+    public long getQueueMaxBytes() {
+        return queueMaxBytes;
+    }
 
     @Override
     public int getCapacity() {
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
@@ -9,6 +9,7 @@ public class FileSettings implements Settings {
     private PageIOFactory pageIOFactory;
     private Class elementClass;
     private int capacity;
+    private int queueMaxSizeInBytes;
     private int maxUnread;
     private int checkpointMaxAcks;
     private int checkpointMaxWrites;
@@ -39,6 +40,12 @@ public class FileSettings implements Settings {
     @Override
     public Settings setElementClass(Class elementClass) {
         this.elementClass = elementClass;
+        return this;
+    }
+
+    @Override
+    public Settings setQueueMaxSizeInBytes(int size) {
+        this.queueMaxSizeInBytes = size;
         return this;
     }
 
@@ -105,6 +112,9 @@ public class FileSettings implements Settings {
     public String getDirPath() {
         return dirForFiles;
     }
+
+    @Override
+    public int getQueueMaxSizeInBytes() { return queueMaxSizeInBytes; }
 
     @Override
     public int getCapacity() {
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/MemorySettings.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/MemorySettings.java
@@ -8,7 +8,7 @@ public class MemorySettings implements Settings {
     private PageIOFactory pageIOFactory;
     private Class elementClass;
     private int capacity;
-    private long queueMaxSizeInBytes;
+    private long queueMaxBytes;
     private final String dirPath;
     private int maxUnread;
     private int checkpointMaxAcks;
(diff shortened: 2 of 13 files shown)


logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -41,9 +41,12 @@
 
     protected volatile long unreadCount;
 
+    protected volatile long currentSize;
+
     private final CheckpointIO checkpointIO;
     private final PageIOFactory pageIOFactory;
     private final int pageCapacity;
```
colinsurprenant: I see that `currentSize` is a `long` while `maxSize` is an `int` ... any reason?

colinsurprenant: so if size is in bytes, it should probably be a `long` since the Java max integer is (2^31) - 1.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -290,7 +300,7 @@ public long write(Queueable element) throws IOException {
     public boolean isFull() {
         // TODO: I am not sure if having unreadCount as volatile is sufficient here. all unreadCount updates are done inside syncronized
         // TODO: sections, I believe that to only read the value here, having it as volatile is sufficient?
-        return (this.maxUnread > 0) ? this.unreadCount >= this.maxUnread : false;
+        return (((this.maxUnread > 0) ? this.unreadCount >= this.maxUnread : false) || (this.currentSize >= this.maxSize));
```
colinsurprenant: mixing the ternary operator with a boolean `||` expression is hard to reason, I would refactor this using `if`/`else` ...

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -255,6 +264,7 @@ public long write(Queueable element) throws IOException {
 
             this.headPage.write(data, seqNum, this.checkpointMaxWrites);
             this.unreadCount++;
+            this.currentSize += data.length;
```
jsvd: That was my initial approach, but if you allocate a page of 10mb and already have 90mb used and the limit is 100mb, does that mean that no writes on the last page will be accepted (since 90+10 is 100 so `isFull()` is `true`?).

The alternative is to only add the page size to counter when the page is full, but that means we'll go over the limit at most 1 page size.

colinsurprenant: - when a page is created, is already uses all the disk space, writing to it does not increase its disk size. 
- because we block after write, we have to assume there is always space for 1 more write which  can mean creating a new page. 
- so either we accept that the max + 1 page can happen or we block after the first write to a new page if that page has reached the threshold even if it means that the page will be mostly "empty" in term of content (but not disk used, as I said an empty page is already pre-allocated on disk to its full capacity) because we cannot know if a 2nd element will fit in or require another page creation.

jsvd:
@colinsurprenant updated the currentSize tracking to only track whole pages, and made the check conservative so we don't go over the limit at the cost of an entire empty page (worst case)

jsvd:
I added the missing ruby code to wire the setting from the yaml to the java classes.

@colinsurprenant another round of review is welcome

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -290,7 +300,11 @@ public long write(Queueable element) throws IOException {
     public boolean isFull() {
         // TODO: I am not sure if having unreadCount as volatile is sufficient here. all unreadCount updates are done inside syncronized
         // TODO: sections, I believe that to only read the value here, having it as volatile is sufficient?
-        return (this.maxUnread > 0) ? this.unreadCount >= this.maxUnread : false;
+        if ((this.maxSizeInBytes > 0) && this.currentSize >= this.maxSizeInBytes) {
+            return true;
+        } else {
+            return ((this.maxUnread > 0) && this.unreadCount >= this.maxUnread);
+        }
```
colinsurprenant: 👍

colinsurprenant:
Looking good so far - suggestion:
you use both `queue_max_size` and `queueMaxSizeInBytes` - what do you think of just `queue_max_bytes` & `queueMaxBytes` ? `queue_max_size` could be confusing (size of what?) and more in line with `queue_max_events` we already have and `queueMaxSizeInBytes` is, well, very explicit :D

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -390,4 +390,121 @@ public void reachMaxUnreadWithAcking() throws IOException, InterruptedException,
...
+
+        // we expect this next write call to block so let's wrap it in a Future
+        Callable<Long> write = () -> {
+            return q.write(element);
+        };
+
+        ExecutorService executor = Executors.newFixedThreadPool(1);
+        Future<Long> future = executor.submit(write);
+
+        while (!q.isFull()) { Thread.sleep(1); }
```
colinsurprenant: Java `Thread.sleep()` accepts a milliseconds parameter, maybe `1` is a bit too small? I suggest `10`?  I know you used the same value as the similar test I wrote but I think I might have been too aggressive :P

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -390,4 +390,121 @@ public void reachMaxUnreadWithAcking() throws IOException, InterruptedException,
...
+        Batch b = q.readBatch(10); // read 1 page (10 events)
+        b.close();  // purge 1 page
+
+        // spin wait until data is written and write blocks
+        while (q.isFull()) { Thread.sleep(1); }
+
+        assertThat(q.isFull(), is(false));
+
+        executor.shutdown();
+    }
```
colinsurprenant: 👍

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -390,4 +390,121 @@ public void reachMaxUnreadWithAcking() throws IOException, InterruptedException,
...
+        assertThat(q.isFull(), is(true));
+
+        Batch b = q.readBatch(9); // read 90% of page (9 events)
+        b.close();  // this should not purge a page
+
+        assertThat(q.isFull(), is(true)); // queue should still be full
+
+        executor.shutdown();
+    }
+
```
colinsurprenant: 👍

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -290,7 +300,11 @@ public long write(Queueable element) throws IOException {
     public boolean isFull() {
         // TODO: I am not sure if having unreadCount as volatile is sufficient here. all unreadCount updates are done inside syncronized
         // TODO: sections, I believe that to only read the value here, having it as volatile is sufficient?
-        return (this.maxUnread > 0) ? this.unreadCount >= this.maxUnread : false;
+        if ((this.maxSizeInBytes > 0) && this.currentSize >= this.maxSizeInBytes) {
+            return true;
```
suyograo: Can we log debug here which `full` criteria was satisfied? We also need metrics, but I think @talevy is looking at it.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -197,6 +206,7 @@ private void add(Checkpoint checkpoint, PageIO pageIO) throws IOException {
             this.tailPages.add(p);
             this.unreadTailPages.add(p);
             this.unreadCount += p.unreadCount();
+            this.currentSize += pageIO.getCapacity();
```
suyograo: this would be a great metric to have. `currentSize`

jsvd:
+1 on renaming the queueMaxSizeInBytes to queueMaxBytes

jsvd:
@colinsurprenant @suyograo do we want to add the current bytes metric in this PR? I vote for separating the 2 features, so we can merge this faster and work on top of master

logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
```diff
@@ -107,6 +114,9 @@ public String getDirPath() {
     }
 
     @Override
+    public long getQueueMaxBytes() { return queueMaxBytes; }
```
jsvd: I thought I was following the current style, but the IDE tricked me :D

screen shot 2016-11-28 at 10 11 39

colinsurprenant: I personally like one liners like that :D

logstash-core/src/main/java/org/logstash/ackedqueue/MemorySettings.java
```diff
@@ -108,6 +115,8 @@ public String getDirPath() {
         return this.dirPath;
     }
 
+    public long getQueueMaxBytes() { return this.queueMaxBytes; }
```
colinsurprenant: same as above ;)

logstash-core/src/main/java/org/logstash/ackedqueue/FileSettings.java
jsvd: since the current style uses multi line, I'll stick with that ;)

colinsurprenant:
@jsvd +1 I think metrics should be in a separate PR.

suyograo:
@jsvd +1 on moving logging and metrics to a separate PR.

colinsurprenant:
LGTM

colinsurprenant:
hm. did you forget to squash?

jsvd:
well, we have agreed before that it was up to us to decide when to squash or not, and I'm actually happy how these commits show the evolution of the feature given the feedback, so I chose not to squash.

colinsurprenant:
ok :'(

acchen97:
@jsvd i didn't see this exposed in the logstash.yml queue settings, can we add it?

suyograo:
PR here for logstash.yml https://github.com/elastic/logstash/pull/6343

PR #6309, opened 2016-11-29
Remove unnecessary log4j-1.2-api depedency.

jordansissel:
This log4j-1.2-api library provides a "log4j 1.2"-like API from the log4j2 library.

We don't seem to use this, and including it seems to be the cause of the
Logstash log4j input rejecting log4j 1.x's SocketAppender with this
message:

    org.apache.log4j.spi.LoggingEvent; class invalid for deserialization

The origin of this error is that log4j2's log4j-1.2-api defines
LoggingEvent without `implements Serializable`.

This commit also includes regenerated gemspec_jars.rb and
logstash-core_jars.rb.

Reference: https://github.com/logstash-plugins/logstash-input-log4j/issues/36

### DIFF
--- a/logstash-core/build.gradle
+++ b/logstash-core/build.gradle
@@ -110,7 +110,6 @@ idea {
 }
 
 dependencies {
-    runtime 'org.apache.logging.log4j:log4j-1.2-api:2.6.2'
     compile 'org.apache.logging.log4j:log4j-api:2.6.2'
     compile 'org.apache.logging.log4j:log4j-core:2.6.2'
     compile 'com.fasterxml.jackson.core:jackson-core:2.7.4'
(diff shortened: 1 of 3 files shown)


suyograo:
LGTM

jordansissel:
Caveat: I am not sure how this will (or if at all) this may affect plugins which use Java libraries that internally that use Log4j 1.x (kafka?). Best to try and find out? :)

suyograo:
@jordansissel the integration test suite already tests the kafka plugins, so we should be good to go.

jordansissel:
I don't mean reads and writes from Kafka. I mean the logging from within
Kafka library may not be configured (I haven't tested at all nor looked at
the code)

porshkevich:
Kafka and kafka-clients in particular use slf4j for logging.

logstash-input-kafka and logstash-output-kafka use bridge org.slf4j:slf4j-log4j12 but may use log4j-slf4j-impl (log4j2 bridge). This can solve the problem.

jordansissel:
@suyograo I think we want this for 5.x also because without this patch the log4j input cannot work.

Issue #6310, opened 2016-11-29
Acceptance test failure in master and 5.x


suyograo:
This could be from the x-pack install commit. Both redhat and debian are failing with the same errors.

https://logstash-ci.elastic.co/job/elastic+logstash+5.x+multijob-acceptance/label=metal,suite=redhat/87/console

https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-acceptance/label=metal,suite=redhat/203/console

ph:
The messages where changes but the expectation where not updated.

PR #6325, opened 2016-11-29
Relax the acceptance test expectation to match only the exception

ph:
Relax the acceptance test expectation to match only the exception

Fixes: #6310


suyograo:
LGTM

PR #6328, opened 2016-11-29
Add feature flag support for RATS

suyograo:
Adds feature flag support for RATS integration tests. First feature to use it
is persistent_queues. You can now enable a FF in travis by using the travis matrix
by adding a env variable called FEATURE_FLAG=<feature_name>.

To use feature flag, you need a logstash.yml file which has this feature turned on.
Unfortunately --path.settings needs an entire dir where logstash.yml is located, not
just the logstash.yml file, so we need to checkin the fixture which has the feature yml
and log4j.

For example, you can now set FEATURE_FLAG=persistent_queues which means travis will
run tests twice, one with this feature enabled and once without it. When the feature is enabled,
the logstash.yml located in qa/integration/fixtures/persistent_queues is picked up whenever LS is
started in tests


qa/integration/services/logstash_service.rb
```diff
@@ -115,6 +115,20 @@ def spawn_logstash(*args)
     end
   end
 
+  def build_child_process(*args)
+    feature_config_dir = @settings.feature_config_dir
+    # if we are using a feature flag and special settings dir to enable it, use it
+    # If some tests is explicitly using --path.settings, ignore doing this, because the tests
+    # chose to overwrite it.
+    if !feature_config_dir.nil? && !args.include?(SETTINGS_CLI_FLAG)
```
colinsurprenant: FYI `if !feature_config_dir.nil?` is equivalent to `if feature_config_dir`

qa/integration/services/logstash_service.rb
```diff
@@ -172,6 +186,16 @@ def process_id
     @process.pid
   end
 
+  def application_settings_file
+    feature_config_dir = @settings.feature_config_dir
+    if feature_config_dir.nil?
```
colinsurprenant: FYI `if feature_config_dir.nil?` is equivalent to `unless feature_config_dir`

colinsurprenant:
Nice! Just a few non essential observations.

colinsurprenant:
LGTM

PR #6331, opened 2016-11-30
add queue stats to node/stats api

### LINKED ISSUE #6182: Queue Monitoring API
With the addition of the new durable queueing infrastructure, it is time to add some monitoring for these structures!

> [P1](https://github.com/elastic/logstash/issues/5638)
> add monitoring info and API to expose PQ stats

thanks to @suyograo, we have an initial proposal on some metrics we wish to include:

| Metric Type | Metric Name | Description |
|- |- |- |
| metadata | type | the type of queue [persisted, memory]
| metadata | page_capacity_bytes | size (in bytes) of page size for the durable queue
| disk | path | directory where the queue data can be found on disk |
| disk | fs_type | the file-system where the queue data is being written to (e.g. ext4) |
| disk | total_bytes | total size of partition queue is located in |
| disk | available_bytes | extra unused space queue can grow into |
| disk | used_bytes | size, in bytes of the queue data |
| disk | pages | number of live pages used to store events |
| io_stats | reads_per_sec | number of read operations per second |
| io_stats | writes_per_sec | number of write operations per second |
| io_stats | read_kb | total kilobytes read from disk |
| io_stats | write_kb | total kilobytes written to disk |
| events | queued_flushed | number of events unread from queue that are committed to disk |
| events | queued_unflushed | uncommitted events queued up in memory to be flushed to disk |
| events | processed | number of events that have been popped/relinquished from queue |
| events | in_flight | number of events that are currently being operated on within the pipeline (being processed) |
| events | max | max events count that can be stored in the bounded queue |
| events | acked | **not sure, help!** |

These stats should be available via the API

talevy:
example queue section:

```javascript
"queue" : {
    "type" : "persisted",
    "capacity" : {
      "max_queue_size_in_bytes" : 1073741824,
      "max_unread_events" : 0
    },
    "data" : {
      "free_space_in_bytes" : 33851523072,
      "current_size_in_bytes" : 262144000,
      "storage_type" : "hfs",
      "path" : "/logstash/data/queue"
    },
    "events" : {
      "acked_count" : 0,
      "unread_count" : 0,
      "unacked_count" : 0
    }
}
```

When retrieving node stats for a logstash instance that is not running with `queue.type: persisted`, the output of this section will simply be:

```javascript
{ "type" : "memory_acked"}
or
{ "type" : "memory"}
```

as discussed in the comments below, this information will be per pipeline and under the `pipeline` section of the stats. All of these stats will be stored in the metrics store.

Closes #6182.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -106,6 +106,30 @@ public class Queue implements Closeable {
         }
     }
 
+    public String getDirPath() {
+        return this.dirPath;
+    }
+
+    public long getMaxBytes() {
+        return this.maxBytes;
+    }
+
+    public long getMaxUnread() {
+        return this.maxUnread;
+    }
+
+    public long getCurrentByteSize() {
+        return this.currentByteSize;
+    }
+
+    public int getPageCapacity() {
+        return this.pageCapacity;
+    }
+
+    public long getUnreadCount() {
+        return this.unreadCount;
+    }
+
     // moved queue opening logic in open() method until we have something in place to used in-memory checkpoints for testing
     // because for now we need to pass a Queue instance to the Page and we don't want to trigger a Queue recovery when
     // testing Page
@@ -583,6 +607,19 @@ public class Queue implements Closeable {
             return this.headPage.getPageNum();
         }
         return this.tailPages.get(0).getPageNum();
+    }
+
+    public long getAckedCount() {
+        return headPage.ackedSeqNums.cardinality() + tailPages.stream()
+                .mapToLong(page -> page.ackedSeqNums.cardinality())
+                .sum();
+    }
+
+    public long getUnackedCount() {
+        long headPageCount = (headPage.getElementCount() - headPage.ackedSeqNums.cardinality());
+        long tailPagesCount = tailPages.stream()
+                .mapToLong(page -> (page.getElementCount() - page.ackedSeqNums.cardinality())).sum();
+        return headPageCount + tailPagesCount;
     }
 
     protected long nextSeqNum() {
--- a/logstash-core/src/test/java/org/logstash/ackedqueue/TestSettings.java
+++ b/logstash-core/src/test/java/org/logstash/ackedqueue/TestSettings.java
@@ -4,6 +4,7 @@ import org.logstash.common.io.ByteBufferPageIO;
 import org.logstash.common.io.CheckpointIOFactory;
 import org.logstash.common.io.FileCheckpointIO;
 import org.logstash.common.io.MemoryCheckpointIO;
+import org.logstash.common.io.MmapPageIO;
 import org.logstash.common.io.PageIOFactory;
 
 public class TestSettings {
@@ -35,10 +36,11 @@ public class TestSettings {
 
     public static Settings getSettingsCheckpointFilePageMemory(int capacity, String folder) {
         Settings s = new FileSettings(folder);
-        PageIOFactory pageIOFactory = (pageNum, size, path) -> new ByteBufferPageIO(pageNum, size, path);
+        PageIOFactory pageIOFactory = (pageNum, size, path) -> new MmapPageIO(pageNum, size, path);
         CheckpointIOFactory checkpointIOFactory = (source) -> new FileCheckpointIO(source);
         s.setCapacity(capacity);
         s.setElementIOFactory(pageIOFactory);
+        s.setCheckpointMaxWrites(1);
         s.setCheckpointIOFactory(checkpointIOFactory);
         s.setElementClass(StringElement.class);
         return s;
--- a/qa/integration/specs/monitoring_api_spec.rb
+++ b/qa/integration/specs/monitoring_api_spec.rb
@@ -47,4 +47,18 @@ describe "Test Monitoring API" do
     end
   end
 
+  it "can retrieve queue stats" do
+    logstash_service = @fixture.get_service("logstash")
+    logstash_service.start_with_stdin
+    logstash_service.wait_for_logstash
+
+    Stud.try(max_retry.times, RSpec::Expectations::ExpectationNotMetError) do
+      result = logstash_service.monitoring_api.node_stats
+      if logstash_service.settings.feature_flag == "persistent_queues"
+        expect(result["queue"]["type"]).to eq "persisted"
+      else
+        expect(result["queue"]["type"]).to eq "memory"
+      end
+    end
+  end
 end
--- a/qa/integration/specs/monitoring_api_spec.rb
+++ b/qa/integration/specs/monitoring_api_spec.rb
@@ -54,10 +54,21 @@ describe "Test Monitoring API" do
 
     Stud.try(max_retry.times, RSpec::Expectations::ExpectationNotMetError) do
       result = logstash_service.monitoring_api.node_stats
+      expect(result["pipeline"]["queue"]).not_to be_nil
       if logstash_service.settings.feature_flag == "persistent_queues"
-        expect(result["queue"]["type"]).to eq "persisted"
+        expect(result["pipeline"]["queue"]["type"]).to eq "persisted"
+        expect(result["pipeline"]["queue"]["data"]["free_space_in_bytes"]).not_to be_nil
+        expect(result["pipeline"]["queue"]["data"]["current_size_in_bytes"]).not_to be_nil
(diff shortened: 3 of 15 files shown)


logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -106,6 +106,30 @@ public Queue(String dirPath, int pageCapacity, long maxBytes, CheckpointIO check
         }
     }
 
+    public String getDirPath() {
+        return dirPath;
```
colinsurprenant: I personally use `this.`  on instance variables to emphasize that fact, I understand this might not be a univeral best-practice but I'd prefer if we could keep that uniform per class at least. Either we agree on removing all usage of `this.` which is not strictly necessary or we stay uniform within classes.

colinsurprenant:
Overall looks good, added a few comments.
Did we want to include specific tests/specs on the newly exposed Ruby methods?

ph:
There is a small issue with this PR, queue stats will only be fetched on request and not saved periodically in the collector.  This mean theses stats won't be available for the monitoring xpack.

There is two way to fix:

1. this either we create a new `Instrument::Periodic` Poller that periodically fetches that information  2. we create an observer to push new metrics as they comme available.

Also, shouldn't the queue stats be under a specific pipeline instead of the `global _node/stats`?, because we are still thinking about multiple pipelines.

ph:
My comments can be done in another PR for 5.2 if we want, I wont block it for that.

colinsurprenant:
@ph 
- what method do you recommend between `Instrument::Periodic` and observer? 
- what do you mean by «push new metrics as they comme available.», specifically «as they comme available»? AFAIU they become available when requested?
- yes, definitely per pipeline because the queue is per pipeline

suyograo:
> yes, definitely per pipeline because the queue is per pipeline

We don't have multi-pipelines as of now, so let's keep them as is until the feature lands. These APIs can always change

ph:
@colinsurprenant I would go with the PeriodicPoller strategy, creating a new one that will monitor the queue and periodically save the stats to the metric store, I don't think we need live stats, so I don't think we need option 2.

@suyograo If multiple pipeline is for 6.0 I am OK to keep it at the node/stats level, we can break the API at that major release.

suyograo:
The monitoring API is marked experimental, which means there is no requirement to keep it backward-compatible. We are still fuzzy on how and when multi-pipelines will work, so I'd like us not to design these APIs with a feature that is not fully baked yet.

logstash-core/src/test/java/org/logstash/ackedqueue/TestSettings.java
```diff
@@ -35,10 +36,11 @@ public static Settings getSettings(int capacity, long size) {
 
     public static Settings getSettingsCheckpointFilePageMemory(int capacity, String folder) {
         Settings s = new FileSettings(folder);
-        PageIOFactory pageIOFactory = (pageNum, size, path) -> new ByteBufferPageIO(pageNum, size, path);
+        PageIOFactory pageIOFactory = (pageNum, size, path) -> new MmapPageIO(pageNum, size, path);
```
talevy: nothing else is using this function

talevy:
> Did we want to include specific tests/specs on the newly exposed Ruby methods?

I was relying on integration tests to properly test the pages of the queue in the logstash context. thought the persistent_queues feature-flag was merged, but I was wrong. I added it there, but I don't have a clean way to enable persisted, so I will wait for that PR to get merged.

talevy:
> There is a small issue with this PR, queue stats will only be fetched on request and not saved periodically in the collector. This mean theses stats won't be available for the monitoring xpack.

@ph, I purposefully left that out of scope. Although, I now realize that the tag `monitoring` may be misleading

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
colinsurprenant: ok thanks - please do the others as well, I did not comment on each one specifically. Thanks!

colinsurprenant:
@suyograo @ph do we have any other metrics which are per-pipeline?

ph:
@colinsurprenant every metric we collect for plugins are related to a specific pipeline, we hide that from the user and display only one pipeline.

colinsurprenant:
@ph @suyograo @talevy ok so I since this is also per-pipeline I suggest we use the same strategy as with the plugins.

suyograo:
Just tested with PQ flag and stats works fine

suyograo:
@talevy: I chatted with @colinsurprenant and @ph about this being ready for 5.2. Since we are not merging this for 5.1, and are in no hurry, we should make the following changes in this PR to be consistent with other metrics API:
* Use the metrics store to record the PQ stats. This would mean using `metric.increment` and other APIs. Currently, this PR returns real-time data to the API, which was ok for 5.1, but we shouldn't do that for 5.2
* Fold these metrics into the `pipeline` semantics. PQs are pipeline-level contructs so these stats are best reported per-pipeline. At this time, there is just a single pipeline called main, so we would just report it under there. For example:

```
GET _node/stats
```

```json
{
 "pipeline":{
  "queue":{
   "type":"persisted",
   "capacity":{
    "page_capacity_in_bytes":262144000,
    "max_queue_size_in_bytes":1073741824,
    "max_unread_events":0
   },
   "data":{
    "free_space_in_bytes":33851523072,
    "current_size_in_bytes":262144000,
    "storage_type":"hfs",
    "path":"/logstash/data/queue"
   },
   "events":{
    "acked_count":0,
    "unread_count":0,
    "unacked_count":0
   }
  },
  "events":{
   "duration_in_millis":7863504,
   "in":100,
   "filtered":100,
   "out":100
  }
  .....
}
```

There are some fixes we can do later to extract the "common" parts of the queue field like "data" and "capacity" to the instance level, but for now this is fine.

talevy:
@ph @colinsurprenant updated to leverage metric store.

The one issue I have with this setup is the fact that `pipeline.queue` is `null` before the pipeline starts, and so the first few seconds of a running logstash instance will have this in the response until the periodic poller runs (after 5s?)

qa/integration/specs/monitoring_api_spec.rb
```diff
@@ -47,4 +47,19 @@
...
+  it "can retrieve queue stats" do
+    logstash_service = @fixture.get_service("logstash")
+    logstash_service.start_with_stdin
+    logstash_service.wait_for_logstash
+
+    Stud.try(max_retry.times, RSpec::Expectations::ExpectationNotMetError) do
+      result = logstash_service.monitoring_api.node_stats
+      expect(result["pipeline"]["queue"]).not_to be_nil
+      if logstash_service.settings.feature_flag == "persistent_queues"
+        expect(result["pipeline"]["queue"]["type"]).to eq "persisted"
```
ph: Maybe we should check for the other fields?

ph:
Concerning the `NULL` maybe we could turn problem around? The problem with the NULL is not that the queue is not correctly setup its because the `tick` didn't happen yet.

If the pipeline had a `#collect_metrics` that would update the current metric, we could call that in the initialize/run method of the pipeline and it could also be called externally by the poller?

@talevy I am just brainstorming ideas here.

ph:
@talevy Woohoo, I've tested it with both `memory` and `persisted` and it work, the only remaining problem is the `NULL` see ma previous comment.

From what I know about the PQ, the stats looks OK.

talevy:
> Concerning the NULL maybe we could turn problem around? The problem with the NULL is not that the queue is not correctly setup its because the tick didn't happen yet.

@ph, when I was testing, I noticed that the periodic pollers run once before the pipeline starts. So the tick did happen, but it happened before the pipeline started.

> If the pipeline had a #collect_metrics that would update the current metric, we could call that in the initialize/run method of the pipeline and it could also be called externally by the poller?

hmm, but the pipeline has no knowledge of the periodic_pollers, doesn't that pose a problem? The weird thing is that this specific periodic poller only makes sense for an initialized pipeline.

ph:
@talevy Correct the periodic flow is, collect_metric => loop(5) { collect_metric }

But the periodic poller are started before a pipeline actually exist.

ph:
@talevy and I discussed it over zoom and we will do something like https://github.com/elastic/logstash/pull/6331#issuecomment-267401415

talevy:
thanks for the review @ph, stats are collected when pipelines are started (same place that reload stats are collected as well)

ph:
LGTM

talevy:
merged into 5.x under: https://github.com/elastic/logstash/commit/f8b4c72a4985539e6b97e7901cc747156ff39268

PR #6338, opened 2016-12-01
Change system-install to allow manual selection

### LINKED ISSUE #6209: Add ability to choose startup style to system-install script
Have it pull override configuration from `startup.options` in `/etc/logstash` or `$LS_HOME/config` and put flags to catch this in system-install.

Not sure if this should be 5.1 or 5.0.1.  Whichever Label can be removed.

untergeek:
Now you can specify 3 arguments, each dependent on the previous:

1. full path to the `startup.options` file
2. the startup variant (sysv, upstart, systemd)
3. the version of the startup variant (usually unnecessary, can safely be left blank)

Present behavior is preserved.  If `system-install` is called with no arguments, it autodetects and installs the detected variant.

fixes #6209

### DIFF
--- a/bin/system-install
+++ b/bin/system-install
@@ -7,12 +7,34 @@ setup
 if [ -z "$1" ]; then
   [ -r ${LOGSTASH_HOME}/config/startup.options ] && . ${LOGSTASH_HOME}/config/startup.options
   [ -r /etc/logstash/startup.options ] && . /etc/logstash/startup.options
+elif [ "$1" == "-h" ] || [ "$1" == "--help" ]; then
+  echo "Usage: system-install [OPTIONSFILE] [STARTUPTYPE] [VERSION]"
+  echo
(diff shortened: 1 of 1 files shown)


untergeek:
Thanks.  Where would we put acceptance tests for this?  It's something that tests the packaging, right?

jordansissel:
@untergeek yep! In truth, I haven't written any recent acceptance tests. I think @ph and @suyograo  are more familiar and can guide on that.

ph:
@untergeek Theses would go into a qa/acceptance test, they are a bit more cumbersome to write than @suyog's amazing rats test. But since we need to test different os/packages I think this is where we should write them. 

I can give you a hand

untergeek:
> I can give you a hand

That will be much appreciated

jordansissel:
LGTM.

We can add test coverage later.

Issue #6339, opened 2016-12-01
bin/logstash-plugin update doesnt work


ph:
Steps to reproduce with an artefact.

1. install a previous version of a plugin
``
bin/logstash-plugin install --version 0.1.0 logstash-filter-qatest
``

2. Edit Gemfile and remove the version constrains.

3. try one of this command

```
bin/logstash-plugin update logstash-filter-qatest
bin/logstash-plugin update
```

You will get the following output.

```
ph@sashimi~/es/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT  master ± bin/logstash-plugin update --no-verify logstash-filter-qatest
Bundler::GemNotFound: Could not find logstash-filter-qatest-0.1.1 in any of the sources
      materialize at /Users/ph/es/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/spec_set.rb:92
             map! at org/jruby/RubyArray.java:2446
      materialize at /Users/ph/es/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/spec_set.rb:85
            specs at /Users/ph/es/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/definition.rb:132
        specs_for at /Users/ph/es/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/definition.rb:177
  requested_specs at /Users/ph/es/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/definition.rb:166
  requested_specs at /Users/ph/es/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/environment.rb:18
            setup at /Users/ph/es/logstash/build/logstash-6.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/1.9/gems/bundler-1.9.10/lib/bundler/runtime.rb:13
[...]

PR #6340, opened 2016-12-01
Do not clean unused gems after updating gems

ph:
The are a difference between the gems installed and the available spec
loaded in memory making bundler remove the gem that we have just
updated.

Fixes: #6339


suyograo:
LGTM

PR #6343, opened 2016-12-02
Add default setting for queue.max_bytes

suyograo:
Follow up from https://github.com/elastic/logstash/pull/6297


jordansissel:
LGTM. Love more docs <3

Issue #5636, opened 2016-07-14
Persistence Design


colinsurprenant:
This issue explains the Logstash persistence design that will be introduced in 5.x. Thanks to design sessions with @bleskes and @guyboertje, we decided to reuse persistence/resiliency concepts from Elasticsearch such as the sequencing, translog and checkpointing strategies. Currently the implementation is all in Java and can be found in the `feature/java_persistence` branch.

Note that from a Logstash perspective, `Event` objects will ultimately be persisted but from this Queue implementation perspective, we use the _Element_ naming which is more generic since any object which implement our `Queuable` interface can be queued/persisted.
### Sequence Number

First for this new design to work, a queue-wide unique and incrementing **sequence number** will be assigned to each Element at queue write/push operation and this sequence number will be used internally by the queue for its internal state tracking.
### Batch Object

A Batch object is also introduced that keeps track of all read/pop’ed elements and is responsible to handle acking and rejection when all processing over the elements batch is completed. all queue read/pop operation returns a batch object. 

Note that for the first iteration we decided to always return batches of elements within the same page which simplifies the design. The tradeoff is that _small_ batches may be produced when reaching the end of each fixed-sized data file.
### Pages, Data Files and Checkpoints

The basic architecture consist of in-memory Page objects with their associated checkpoint files and data files.

![persistence design 2](https://cloud.githubusercontent.com/assets/2010/16636182/81807e5a-43a4-11e6-8c30-6f109194f498.png)

The in-memory Page object keep track of the runtime unread elements with a simple _first unread sequence number_  and acked elements using a bitset for out-of-order acking support. 

Each data file is bounded by size by configuring a global per-page maximum capacity. Data files are append-only and writes can only happen in the head page data file. Once the head page data file fills up, the head page is beheaded and becomes the latest tail page which only consist on creating a new associated checkpoint file using the `checkpoint.{pageNum}` naming and a new empty head page is created with its special checkpoint naming `checkpoint.head`.

A tail page checkpoint file consists of only a few numeric pointers which tracks each page internal state: own page number, the minimum sequence number, the number of elements and the first unacked sequence number. These pointers are sufficient to reconstruct a complete page state when restarting/recovering.

A head page checkpoint file is identical to a tail page checkpoint but also track the first unacked page number, whish is the first page containing unacked elements.

![queue structure](https://cloud.githubusercontent.com/assets/2010/16637013/da4dcf2e-43a9-11e6-8535-105d9c45c86b.png)
[...]

PR #6344, opened 2016-12-02
Add docs for persistent queues

dedemorton:
Adds end user documentation to support PR #5636


docs/static/persistent-queues.asciidoc
```diff
@@ -0,0 +1,127 @@
...
+stage, you can enable persistent queues to buffer events on disk. The queue size
+is variable and configurable, which means that you have more control over
+managing situations that could result in back pressure to the source. 
+
+[[persistent-queues-advantages]]
+==== Advantages of Persistent Queues
+
+Using persistent queues:
+
+* Provides protection from losing in-flight messages when the Logstash process is shut down or the machine is restarted.
```
acchen97: @suyograo I understand instances failures, but is the machine restart portion true? If there was a full machine reboot, what happens to the events in memory prior to fsync?

docs/static/persistent-queues.asciidoc
```diff
@@ -0,0 +1,127 @@
...
+while events are in-flight, Logstash will attempt to deliver messages stored
+in the persistent queue until delivery succeeds at least once. In other words,
+messages stored in the persistent queue may be duplicated, but not lost.
+
+[[persistent-queues-limitations]]
+==== Limitations of Persistent Queues
+
+The current implementation of persistent queues has the following limitations:
+
+* This version does not enable full end-to-end resiliency. Logstash only
```
acchen97: I think its important to note (maybe even leading with) that there **will** be full end-to-end resiliency for some inputs, like Filebeat and winlogbeat who themselves guarantee at least once delivery. @jordansissel what other inputs would fit into this category (besides other MQs)?

docs/static/persistent-queues.asciidoc
```diff
@@ -0,0 +1,127 @@
...
+Example configuration:
+
+[source, yaml]
+queue.type: persisted
+queue.max_bytes: 4g 
+
+[[backpressure-persistent-queues]]
+==== Handling Back Pressure
+
+Logstash has a built-in mechanism that adds back pressure to the data flow when
```
acchen97: "exerts" back pressure may be a better term

docs/static/persistent-queues.asciidoc
```diff
@@ -0,0 +1,127 @@
...
+sets the total capacity of the queue to 8gb:
+
+[source, yaml]
+queue.type: persisted
+queue.max_bytes: 8gb
+
+With these settings specified, Logstash will buffer unACKed events on disk until 
+the size of the queue reaches 8gb. When the queue is full of unACKed events, and
+the size limit has been reached, Logstash will no longer accept new events. This
+process is called back pressure.
```
acchen97: it seems a bit odd that back pressure is "defined" after it's already been used twice here. might make sense to somehow define it at the start of the section?

acchen97:
Added a few comments, but looks good overall. I like the restructuring that you did.

docs/static/persistent-queues.asciidoc
```diff
@@ -0,0 +1,127 @@
...
+
+Each input handles back pressure independently. For example, when the
+<<plugins-inputs-beats,beats>> input encounters back pressure, it no longer
+accepts new connections and waits until the persistent queue has space to accept
+more events. After the filter and output stages finish processing existing
+events in the queue and ACKs them, Logstash automatically starts accepting new
+events.
```
acchen97: I think we should also have a short section on monitoring PQs. This can mostly just link to the monitoring API queue stats docs.

suyograo:
Can we add 2 more sections:

-----------------------------------------
**Controlling Durability**

When the persistent queue feature is enabled, Logstash will store events on disk. The persistent queue exposes this trade-off between performance and durability by providing the following configuration options. 

queue.checkpoint.writes: The number of writes to the queue to trigger a fsync to disk. This configuration controls the durability from the producer side. Keep in mind that a disk flush is a relatively heavy operation which would affect throughput if performed after every write. For instance, if you would like to ensure all messages in Logstash's queue is durable, you can set queue.checkpoint.writes to 1. This could severely impact performance.

queue.checkpoint.acks: The number of acks to the queue to trigger a fsync to disk. This configuration controls the durability from the consumer side.

The process of checkpointing is atomic, which means any update to the file is saved if successful.

-------------------------------------------------------------------

**Shutdown Scenarios**

logstash/machine fault, we lose anything after the last checkpoint. If you need durability around that you have to set max_writes=1

docs/static/persistent-queues.asciidoc
suyograo: monitoring queue is not in 5.1 so we'll need to punt on this.

docs/static/persistent-queues.asciidoc
suyograo: This version does not enable full end-to-end resiliency except for anything sent to beats input.

docs/static/persistent-queues.asciidoc
suyograo: those will be lost. See shutdown section below.

suyograo:
@dedemorton looks great overall.

suyograo:
<3 it. LGTM

Issue #6335, opened 2016-12-01
`Java::JavaLang::OutOfMemoryError:` when running the test


ph:
First happening on https://logstash-ci.elastic.co/job/elastic+logstash+5.1+multijob-os-compatibility/1/os=debian/console

Link: https://logstash-ci.elastic.co/job/elastic+logstash+5.1+multijob-os-compatibility/11/os=debian/console
```
Failures:

  1) LogStash::Agent metrics after config reloading when reloading a good config resets the pipeline metric collector
     Failure/Error: Unable to find matching line from backtrace
     Java::JavaLang::OutOfMemoryError:
       GC overhead limit exceeded
     # com.concurrent_ruby.ext.JavaAtomicFixnumLibrary$1.allocate(JavaAtomicFixnumLibrary.java:29)
     # org.jruby.RubyClass.allocate(RubyClass.java:225)
     # org.jruby.RubyClass.newInstance(RubyClass.java:856)
     # org.jruby.RubyClass$INVOKER$i$newInstance.call(RubyClass$INVOKER$i$newInstance.gen)
     # org.jruby.internal.runtime.methods.JavaMethod$JavaMethodZeroOrOneOrNBlock.call(JavaMethod.java:297)
     # org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:168)
     # rubyjit.LogStash::Instrument::MetricType::Counter$$initialize_5a4ead66925fbf9edda8efe988f74d314972c9e81442407170.__file__(/var/lib/jenkins/workspace/elastic+logstash+5.1+multijob-os-compatibility/os/debian/logstash-core/lib/logstash/instrument/metric_type/counter.rb:10)
     # rubyjit.LogStash::Instrument::MetricType::Counter$$initialize_5a4ead66925fbf9edda8efe988f74d314972c9e81442407170.__file__(/var/lib/jenkins/workspace/elastic+logstash+5.1+multijob-os-compatibility/os/debian/logstash-core/lib/logstash/instrument/metric_type/counter.rb)
     # org.jruby.ast.executable.AbstractScript.__file__(AbstractScript.java:46)
     # org.jruby.internal.runtime.methods.JittedMethod.call(JittedMethod.java:241)
     # org.jruby.runtime.callsite.CachingCallSite.callBlock(CachingCallSite.java:211)
     # org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:217)
[...]

PR #6346, opened 2016-12-02
Change the assertions in the config reloading spec

ph:
The assertions was using dummy outputs and kept received events into an
array in memory, but the test actually only needed to match the number
of events it received, this PR add a DroppingDummyOutput that wont
retain the events in memory.

The previous implementation was causing a OOM issue when running the
test on a very fast machine.

Fixes: #6335


ph:
I have decided not to remove the existing class because t might be used in other tests and could be useful for doing assertions on the actual events that goes through the pipeline.

jsvd:
great find @ph! thanks for looking into this. LGTM

Issue #6252, opened 2016-11-15
Logstash Monitoring API to Capture cgroup Information


skearns64:
In https://github.com/elastic/elasticsearch/pull/21029 , Elasticsearch added the ability to collect cgroup information. This is really helpful when you're running in a container, and the "normal" metrics like CPU and system load are still tracked at the machine level, not the container level. By capturing the cgroup info, we can get a more useful read of CPU load, including whether our container is being throttled, etc. 

I believe Logstash would benefit from having the same cgroup info available in the Logstash monitoring nodes stats API.

PR #6357, opened 2016-12-05
Add cgroup information to the api

ph:
When logstash is run under a linux container we will gather statistic about the cgroup and the
cpu usage. This information will should in the /_node/stats api and the result will look like this:

```
  "os" : {
    "cgroup" : {
      "cpuacct" : {
        "usage" : 789470280230,
        "control_group" : "/user.slice/user-1000.slice"
      },
      "cpu" : {
        "cpu_quota_micros" : -1,
        "control_group" : "/user.slice/user-1000.slice",
        "stat" : {
          "number_of_times_throttled" : 0,
          "time_throttled_nanos" : 0,
          "number_of_periods" : 0
        },
        "cfs_period_micros" : 100000
      }
    }
  }
```

Fixes: #6252

### DIFF
--- a/logstash-core/lib/logstash/api/commands/stats.rb
+++ b/logstash-core/lib/logstash/api/commands/stats.rb
@@ -65,6 +65,14 @@ module LogStash
           }
         end
 
+        def os
+          service.get_shallow(:os)
+        rescue
+          # The only currently fetch OS information is about the linux
+          # containers.
+          {}
+        end
+
         def gc
           service.get_shallow(:jvm, :gc)
         end
--- a/logstash-core/lib/logstash/api/modules/node_stats.rb
+++ b/logstash-core/lib/logstash/api/modules/node_stats.rb
@@ -13,12 +13,16 @@ module LogStash
             :jvm => jvm_payload,
             :process => process_payload,
             :pipeline => pipeline_payload,
-            :reloads => reloads
+            :reloads => reloads,
+            :os => os_payload
           }
           respond_with(payload, {:filter => params["filter"]})
         end
 
         private
+        def os_payload
+          @stats.os
+        end
 
         def events_payload
           @stats.events
--- a/logstash-core/lib/logstash/api/modules/stats.rb
+++ b/logstash-core/lib/logstash/api/modules/stats.rb
@@ -30,8 +30,9 @@ module LogStash
             :jvm => {
               :timestamp => stats_command.started_at,
               :uptime_in_millis => stats_command.uptime,
-              :memory => stats_command.memory
-            }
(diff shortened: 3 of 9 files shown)


ph:
I will take a look at the failling test

ph:
should be fixed now.

ph:
@jsvd this will also be need for the monitoring.

jordansissel:
I tested this manually. Creating a new cgroup for myself:

```
% sudo cgcreate -t jls:jls -g cpu,cpuacct:/jls
% cgexec -g cpu:/jls bin/logstash -e '' --http.host 0.0.0.0
```

Looking good.

My next question is about the following example stat output:

```
  "os" : {
    "cgroup" : {
      "cpuacct" : {
        "usage" : 26969185690,
```

* There's no obvious unit for `usage`. Could we add one?
* Should `cpuacct` be `cpu,cpuacct` instead, since this is what Linux calls it?

ph:
@jordansissel 

I've used the ES reference to return the same information, ES don't explicitly return a unit with the `usage`, I believe this is typo and it should be `usage_nanos` IMHO.

```
os.cgroup.cpuacct.usage (Linux only)
The total CPU time (in nanoseconds) consumed by all tasks in the same cgroup as the Elasticsearch process
```

https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html

jordansissel:
+1 on calling it `usage_nanoseconds` or slightly worse `usage_nanos` -- I leave it up to you :)

jordansissel:
LGTM

usage, usage_nanos, usage_nanoseconds -- I leave it up to you.

ph:
I'll go with usage for now to keep in sync with ES, in the future we might want to have more description on the unit.

jasontedor:
> I believe this is typo and it should be `usage_nanos` IMHO.

It is, I agree with you that it should have been called `usage_nanos` and I'm leaning towards changing it to this to be consistent with the other metrics that carry units.

jasontedor:
It already is `usage_nanos` in the [code](https://github.com/elastic/elasticsearch/blame/43f71015a8d117265a54e01658302f9ad7903821/core/src/main/java/org/elasticsearch/monitor/os/OsStats.java#L399), this was just a doc bug; I pushed elastic/elasticsearch@41ffb008ad285735f8bc3c9b97b23c6589a17fef.

ph:
@jason I will make a new PR so both are in sync.

ph:
@jasontedor done in https://github.com/elastic/logstash/pull/6428 and merged! thanks for ping.

jasontedor:
Thanks @ph.

PR #6364, opened 2016-12-06
correct batch delay help

### LINKED ISSUE #6273: update new short-batch strategy
in 5.x `WrappedSyncronousQueue#take_originals_from_queue` is now using a short-batch queue read strategy which is consistent with the Persistent Queue implementation and which should also reduce perceived latency in some cases.

We need to update docs where the queue read strategy is mentioned and also update the phrasing of the  command line flag `-u, --pipeline.batch.delay DELAY_IN_MS When creating pipeline batches, how long to wait while polling for the next event.`

colinsurprenant:
fixes #6273


dedemorton:
LGTM

suyograo:
LGTM

PR #6373, opened 2016-12-07
Include jars in the lib directory

suyograo:
Include the jars that are now in lib/* instead of vendor/ since we now depend on
jar dependencies to load them. Without this jars are not bundled in published gems


ph:
This is not necessary  since we merged https://github.com/elastic/logstash/pull/6377

PR #6380, opened 2016-12-08
Allow rake version:set to update the version of the logstash-core-queue

ph:
rake version:set now update the queue gem
the queue version is bumped to 6.0.0-alpha1 so all the version of the gems are the same.


suyograo:
LGTM

PR #6382, opened 2016-12-08
Have logstash.lib.sh find jvm.options

### LINKED ISSUE #6379: jvm.options is left out of custom path.settings
This was observed under RHEL 7.2 running Logstash 5.0.2

The [docs](https://www.elastic.co/guide/en/logstash/current/command-line-flags.html) do say that `path.settings` specifies the location of `logstash.yml` and log4j config.  From the `logstash.yml` file, one can then set `path.config` for LS configuration.   So far so good, but it is not clear how `jvm.options` can be moved from /etc/logstash to a custom location.   The possible locations [appear to be statically defined](https://github.com/elastic/logstash/blob/d5d556989dcf8232c83ef9b6558c0bbe299e3455/bin/logstash.lib.sh#L67-L76)

untergeek:
This will work if `--path.settings` is properly used on the command line

fixes #6379

### DIFF
--- a/bin/logstash.lib.sh
+++ b/bin/logstash.lib.sh
@@ -28,6 +28,26 @@ export LOGSTASH_HOME
 SINCEDB_DIR=${LOGSTASH_HOME}
 export SINCEDB_DIR
 
+# This block will iterate over the command-line args Logstash was started with
+# It will find the argument _after_ --path.settings and use that to attempt
+# to derive the location of an acceptable jvm.options file
+# It will do nothing if this is not found.
+# This fix is for #6379
+if [ -z "$LS_JVM_OPTS" ]; then
+  found=0
+  for i in "$@"; do
+     if [ $found -eq 1 ]; then
+       if [ -r "${i}/jvm.options" ]; then
+         export LS_JVM_OPTS="${i}/jvm.options"
+         break
+       fi
+     fi
+     if [ "$i" = "--path.settings" ]; then
+       found=1
+     fi
+  done
+fi
+
 parse_jvm_options() {
   if [ -f "$1" ]; then
     echo "$(grep "^-" "$1" | tr '\n' ' ')"


bin/logstash.lib.sh
```diff
@@ -28,6 +28,28 @@ export LOGSTASH_HOME
...
+     if [ $found -eq 1 ]; then
+       if [ -r "${i}/jvm.options" ]; then
+         export LS_JVM_OPTS="${i}/jvm.options"
+         break
+       fi
+     fi
+     if [ "$i" = "--path.settings" ]; then
+       found=1
+     fi
+     ((count=count+1))
```
jordansissel: Looks like count is not used in this code?

untergeek: Hee hee.  I was using that originally to show which index of the array was being used.  

Pruning!

bin/logstash.lib.sh
```diff
@@ -28,6 +28,26 @@ export LOGSTASH_HOME
...
 export SINCEDB_DIR
 
+# This block will iterate over the command-line args Logstash was started with
+# It will find the argument _after_ --path.settings and use that to attempt
+# to derive the location of an acceptable jvm.options file
+# It will do nothing if this is not found.
+# This fix is for #6379
+if [ -z "$LS_JVM_OPTS" ]; then
+  found=0
+  for i in $@; do
```
jordansissel: This should be `"$@"` otherwise it ignores spacing in fun ways that are hard to debug.

Example:

```
# Set 3 arguments. First has a space.
$ set -- "hello world" "two" "three"

# How many args if we don't quote $@?
$ sh -c 'echo $#' - $@
4

# How many if we quote?
$ sh -c 'echo $#' - "$@"
3
```

untergeek: I guess that's true.  It would not be very likely with our scenario, but we should plan for the worst.

jordansissel:
LGTM

PR #6385, opened 2016-12-09
add webserver address metric to metric store

jsvd:
since the port used by the webserver is decided at runtime during agent.run, we need to make the webserver write the finally chosen port as metric to the metric store.

This PR adds a new metric called http_address at the root of the metric store which represents the final "host:port" address of the webserver api.

### DIFF
--- a/logstash-core/lib/logstash/api/commands/default_metadata.rb
+++ b/logstash-core/lib/logstash/api/commands/default_metadata.rb
@@ -20,7 +20,9 @@ module LogStash
         end
 
         def http_address
-          service.agent.webserver.address
+          @http_address ||= service.get_shallow(:http_address).value
+        rescue ::LogStash::Instrument::MetricStore::MetricNotFound, NoMethodError => e
+          nil
         end
       end
     end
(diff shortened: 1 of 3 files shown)


jsvd:
@ph do you think we should test this at the metric collector side, or on the webserver_spec side?

suyograo:
This can easily be verified in RATS as well.

ph:
@jsvd I think we can verify it on webserver_spec, I think the api was already returning that information but I thin it was not using the metric store and was accessing the information directly directly, we should make sure that this information is coming from a unique place. WDYT.

logstash-core/lib/logstash/api/commands/default_metadata.rb
```diff
@@ -20,7 +20,9 @@ def version
         end
 
         def http_address
-          service.agent.webserver.address
+          @http_address ||= service.get_shallow(:http_address).value
+        rescue ::LogStash::Instrument::MetricStore::MetricNotFound, NoMethodError => e
+          nil
```
ph: NoMethod error in which case we can get that, if a metric is returned the #value should exist.

jsvd: @ph in testing, unfortunately, the metric subsystem isn't initialized, so service is nil

ph: OK, can you create a specific issue with the symptoms to address it in a different PR.

jsvd: https://github.com/elastic/logstash/issues/6436

PR #6389, opened 2016-12-09
Use Mozilla's Modern cipher list -- taken on December 9, 2016

jordansissel:
This upgrades the cipher suite from Intermediate to Modern.

    Additionally, I have updated the Intermediate list constant from today's
    Mozilla security page and added !3DES and !DES to explicitly avoid
    DES-based ciphers given recent news about sweet32.


jordansissel:
I don't know exactly the impact of this change on users. We need some testing (tcp input, http poller input, etc) on plugins that use SSL to see what negative impact this may have.

jordansissel:
Going to revert this in favor of an updated intermediates PR.

jordansissel:
The Modern ciphers list results in zero ciphers available on JRuby. I haven't looked into why.

PR #6400, opened 2016-12-13
Correct doco

deanmac5:
the examples how GET _node but I believe it should be GET _nodes
At least that's what I needed to do to make it work in Kibana dev tools


deanmac5:
I've signed the CLA now

suyograo:
@deanmac5 There is no `_nodes` end-point in LS. It is the singluar `_node`. You must have been confused with ES which has `_nodes`

PR #6408, opened 2016-12-13
Improve the accuracy of the PQ docs.

jordansissel:
* Removed reference to unused `queue.checkpoint.interval`.
* Changed 'unread' to be 'unACKed'
* Remove confusing/contradictory statement about queue acknowledgement
  and outputs

### DIFF
--- a/docs/static/persistent-queues.asciidoc
+++ b/docs/static/persistent-queues.asciidoc
@@ -1,77 +1,74 @@
 [[persistent-queues]]
 === Persistent Queues
 
-WARNING: This functionality is in beta and is subject to change. It should be deployed in production at your own risk.
+WARNING: This functionality is in beta and is subject to change. Deployment in production is at your own risk.
 
 By default, Logstash uses in-memory bounded queues between pipeline stages
-(input → filter and filter → output) to buffer events. The size of these 
-in-memory queues is fixed and not configurable. If Logstash terminates unsafely,
-either as the result of a software failure or the user forcing an unsafe
-shutdown, it's possible to lose queued events. 
+(inputs → pipeline workers) to buffer events. The size of these in-memory
+queues is fixed and not configurable. If Logstash experiences a temporary
+machine failure, the contents of the in-memory queue will be lost. Temporary machine
+failures are scenarios where Logstash or its host machine are terminated
+abnormally but are capable of being restarted. 
 
-To prevent event loss in these scenarios, you can configure Logstash to use
-persistent queues. With persistent queues enabled, Logstash persists buffered
-events to disk instead of storing them in memory. 
+In order to protect against data loss during abnormal termination, Logstash has
+a persistent queue feature which will store the message queue on disk.
+Persistent queues provide durability of data within Logstash.
 
-Persistent queues are also useful for Logstash deployments that require high
-throughput and resiliency. Instead of deploying and managing a message
-broker, such as Redis, RabbitMQ, or Apache Kafka, to handle a mismatch in
-cadence between the shipping stage and the relatively expensive processing
-stage, you can enable persistent queues to buffer events on disk. The queue size
-is variable and configurable, which means that you have more control over
-managing situations that can result in back pressure to the source. See <<backpressure-persistent-queues>>. 
+Persistent queues are also useful for Logstash deployments that need large buffers.
+Instead of deploying and managing a message broker, such as Redis, RabbitMQ, or
+Apache Kafka, to facilitate a buffered publish-subscriber model, you can enable
+persistent queues to buffer events on disk and remove the message broker.
 
-[[persistent-queues-advantages]]
-==== Advantages of Persistent Queues
+In summary, the two benefits of enabling persistent queues are as follows:
 
-Using persistent queues:
-
-* Provides protection from losing in-flight messages when the Logstash process is shut down or the machine is restarted.
-* Handles the surge of events without having to use an external queueing mechanism like Redis or Kafka.
-* Provides an at-least-once message delivery guarantee. If Logstash is restarted
-while events are in-flight, Logstash will attempt to deliver messages stored
-in the persistent queue until delivery succeeds at least once. In other words,
-messages stored in the persistent queue may be duplicated, but not lost.
+* Provides protection from in-flight message loss when the Logstash process is abnormally terminated.
+* Absorbs bursts of events without needing an external buffering mechanism like Redis or Apache Kafka.
 
 [[persistent-queues-limitations]]
 ==== Limitations of Persistent Queues
 
-The current implementation of persistent queues has the following limitations:
+The following are problems not solved by the persistent queue feature:
 
-* This version does not enable full end-to-end resiliency, except for messages
-sent to the <<plugins-inputs-beats,beats>> input. For other inputs, Logstash
-only acknowledges delivery of messages in the filter and output stages, and not
-all the way back to the input or source.
-* It does not handle permanent disk or machine failures. The data persisted to disk is not replicated, so it is still a single point of failure.
+* Input plugins that do not use a request-response protocol cannot be protected from data loss. For example: tcp, udp, zeromq push+pull, and many other inputs do not have a mechanism to acknowledge receipt to the sender. Plugins such as beats and http, which *do* have a acknowledgement capability, are well protected by this queue.
+* It does not handle permanent machine failures such as disk corruption, disk failure, and machine loss. The data persisted to disk is not replicated.
 
 [[persistent-queues-architecture]]
 ==== How Persistent Queues Work
 
-The persistent queue sits between the input and filter stages in the same
+The queue sits between the input and filter stages in the same
 process:
 
-input → persistent queue → filter + output 
+input → queue → filter + output 
 
-The input stage reads data from the configured data source and writes events to
-the persistent queue for processing. As events pass through the pipeline,
-Logstash pulls a batch of events from the persistent queue for processing them
-in the filter and output stages. As events are processed, Logstash uses a
-checkpoint file to track which events are successfully acknowledged (ACKed) as
-processed by Logstash. An event is recorded as ACKed in the checkpoint file if
-the event is successfully sent to the last output stage in the pipeline;
-Logstash does not wait for the output to acknowledge delivery. 
+When an input has events ready to process, it writes them to the queue. When
+the write to the queue is successful, the input can send an acknowledgement to
+its data source.
 
-During a normal, controlled shutdown (*CTRL+C*), Logstash finishes
-processing the current in-flight events (that is, the events being processed by
-the filter and output stages, not the queued events), finalizes the ACKing
-of these events, and then terminates the Logstash process. Upon restart,
-Logstash uses the checkpoint file to pick up where it left off in the persistent
-queue and processes the events in the backlog. 
+When processing events from the queue, Logstash acknowledges events as
+completed, within the queue, only after filters and outputs have completed.
+The queue keeps a record of events that have been processed by the pipeline.
+An event is recorded as processed (in this document, called "acknowledged" or
+"ACKed") if, and only if, the event has been processed completely by the
+Logstash pipeline. 
 
-If Logstash crashes or experiences an uncontrolled shutdown, any in-flight
-events are left as unACKed in the persistent queue. Upon restart, Logstash will
-replay the events from its history, potentially leading to duplicate data being
-written to the output.
+What does acknowledged mean? This means the event has been handled by all
+configured filters and outputs. For example, if you have only one output,
+Elasticsearch, an event is ACKed when the Elasticsearch output has successfully
+sent this event to Elasticsearch. 
+
+During a normal shutdown (*CTRL+C* or SIGTERM), Logstash will stop reading
+from the queue and will finish processing the in-flight events being processed
+by the filters and outputs. Upon restart, Logstash will resume processing the
+events in the persistent queue as well as accepting new events from inputs.
+
+If Logstash is abnormally terminated, any in-flight events will not have been
+ACKed and will be reprocessed by filters and outputs when Logstash is
+restarted. Logstash processes events in batches, so it is possible
+that for any given batch, some of that batch may have been successfully
+completed, but not recorded as ACKed, when an abnormal termination occurs.
+
+For more details specific behaviors of queue writes and acknowledgement, see 
+<<durability-persistent-queues>>.
 
 [[configuring-persistent-queues]]
 ==== Configuring Persistent Queues
@@ -79,18 +76,20 @@ written to the output.
 To configure persistent queues, you can specify the following options in the
 Logstash <<logstash-settings-file,settings file>>:
 
-* `queue.type`: Specify `persisted` to enable persistent queues. By default, persistent queues are disabled (`queue.type: memory`).
+* `queue.type`: Specify `persisted` to enable persistent queues. By default, persistent queues are disabled (default: `queue.type: memory`).
 * `path.queue`: The directory path where the data files will be stored. By default, the files are stored in `path.data/queue`. 
-* `queue.page_capacity`: The size of the page data file. The queue data consists of append-only data files separated into pages. The default size is 250mb. 
-* `queue.max_events`:  The maximum number of unread events that are allowed in the queue. The default is 0 (unlimited).
+* `queue.page_capacity`: The maximum size of a queue page in bytes. The queue data consists of append-only files called "pages". The default size is 250mb. Changing this value is unlikely to have performance benefits.
+// Technically, I know, this isn't "maximum number of events" it's really maximum number of events not yet read by the pipeline worker. We only use this for testing and users generally shouldn't be setting this.
+* `queue.max_events`:  The maximum number of events that are allowed in the queue. The default is 0 (unlimited). This value is used internally for the Logstash test suite.
 * `queue.max_bytes`: The total capacity of the queue in number of bytes. The
 default is 1024mb (1gb). Make sure the capacity of your disk drive is greater
-than the value you specify here. If both `queue.max_events` and 
-`queue.max_bytes` are specified, Logstash uses whichever criteria is reached
-first. 
+than the value you specify here.
 
-You can also specify options that control when the checkpoint file gets updated (`queue.checkpoint.acks`, `queue.checkpoint.writes`, and
-`queue.checkpoint.interval`). See <<durability-persistent-queues>>.
+If both `queue.max_events` and 
+`queue.max_bytes` are specified, Logstash uses whichever criteria is reached
+first. See <<backpressure-persistent-queue>> for behavior when these queue limits are reached.
+
+You can also specify options that control when the checkpoint file gets updated (`queue.checkpoint.acks`, `queue.checkpoint.writes`). See <<durability-persistent-queues>>.
 
 Example configuration:
 
@@ -101,21 +100,19 @@ queue.max_bytes: 4gb
 [[backpressure-persistent-queues]]
 ==== Handling Back Pressure
 
-Logstash has a built-in mechanism that exerts back pressure on the data flow 
-when the queue is full. This mechanism helps Logstash control the rate of data
-flow at the input stage without overwhelming downstream stages and outputs like
-Elasticsearch.
+When the queue is full, Logstash puts back pressure on the inputs to stall data
+flowing into Logstash. This mechanism helps Logstash control the rate of data
+flow at the input stage without overwhelming outputs like Elasticsearch.
 
-You can control when back pressure happens by using the `queue.max_bytes` 
-setting to configure the capacity of the queue on disk. The following example
-sets the total capacity of the queue to 8gb:
+Use `queue.max_bytes` setting to configure the total capacity of the queue on
+disk. The following example sets the total capacity of the queue to 8gb:
 
 [source, yaml]
 queue.type: persisted
 queue.max_bytes: 8gb
 
-With these settings specified, Logstash will buffer unACKed events on disk until 
-the size of the queue reaches 8gb. When the queue is full of unACKed events, and
+With these settings specified, Logstash will buffer events on disk until the
+size of the queue reaches 8gb. When the queue is full of unACKed events, and
 the size limit has been reached, Logstash will no longer accept new events. 
 
 Each input handles back pressure independently. For example, when the
@@ -128,24 +125,38 @@ events.
 [[durability-persistent-queues]]
 ==== Controlling Durability
 
+Durability is a property of storage writes that ensures data will be available after it's written.
+
 When the persistent queue feature is enabled, Logstash will store events on
-disk. The persistent queue exposes the trade-off between performance and
-durability by providing the following configuration options:
+disk. Logstash commits to disk in a mechanism called checkpointing.
 
-* `queue.checkpoint.writes`: The number of writes to the queue to trigger an
-fsync to disk. This configuration controls the durability from the producer
-side. Keep in mind that a disk flush is a relatively heavy operation that will
-affect throughput if performed after every write. For instance, if you want to
-ensure that all messages in Logstash's queue are durable, you can set
-`queue.checkpoint.writes: 1`. However, this setting can severely impact
-performance.
+To discuss durability, we need to introduce a few details about how the persistent queue is implemented.
 
-* `queue.checkpoint.acks`: The number of ACKs to the queue to trigger an fsync to disk. This configuration controls the durability from the consumer side.
+First, the queue itself is a set of pages. There are two kinds of pages: head pages and tail pages. The head page is where new events are written. There is only one head page. When the head page is of a certain size (see `queue.page_capacity`), it becomes a tail page, and a new head page is created. Tail pages are immutable, and the head page is append-only. 
+Second, the queue records details about itself (pages, acknowledgements, etc) in a separate file called a checkpoint file.
 
-The process of checkpointing is atomic, which means any update to the file is
-saved if successful.
+When recording a checkpoint, Logstash will:
+
+* Call fsync on the head page.
+* Atomically write to disk the current state of the queue.
+
+The following settings are available to let you tune durability:
+
+* `queue.checkpoint.writes`: Logstash will checkpoint after this many writes into the queue. Currently, one event counts as one write, but this may change in future releases.
+* `queue.checkpoint.acks`: Logstah will checkpoint after this many events are acknowledged. This configuration controls the durability at the processing (filter + output)
+part of Logstash.
+
+Disk writes have a resource cost. Tuning the above values higher or lower will trade durability for performance. For instance, if you want to the strongest durability for all input events, you can set `queue.checkpoint.writes: 1`.
+
+The process of checkpointing is atomic, which means any update to the file is saved if successful.
 
 If Logstash is terminated, or if there is a hardware level failure, any data
 that is buffered in the persistent queue, but not yet checkpointed, is lost.
 To avoid this possibility, you can set `queue.checkpoint.writes: 1`, but keep in
 mind that this setting can severely impact performance.
+
+[[garbage-collection]]
+==== Disk Garbage Collection
+
+On disk, the queue is stored as a set of pages where each page is one file. Each page can be at most `queue.page_capacity` in size. Pages are deleted (garbage collected) after all events in that page have been ACKed. If an older page has at least one event that is not yet ACKed, that entire page will remain on disk until all events in that page are successfully processed. Each page containing unprocessed events will count against the `queue.max_bytes` byte size.
+


docs/static/persistent-queues.asciidoc
```diff
@@ -1,47 +1,45 @@
...
+WARNING: This functionality is in beta and is subject to change. Deployment in production is at your own risk.
 
 By default, Logstash uses in-memory bounded queues between pipeline stages
-(input → filter and filter → output) to buffer events. The size of these 
-in-memory queues is fixed and not configurable. If Logstash terminates unsafely,
-either as the result of a software failure or the user forcing an unsafe
-shutdown, it's possible to lose queued events. 
+(inputs → pipeline workers) to buffer events. The size of these in-memory
+queues is fixed and not configurable. If Logstash terminates abnormally,
+either as the result of a software crash of  `kill -9`, it's
```
colinsurprenant: of -> or ?

docs/static/persistent-queues.asciidoc
```diff
@@ -1,47 +1,45 @@
...
 
-The current implementation of persistent queues has the following limitations:
+Persistent queues solves many problems, but the following are problems not
+solved:
 
-* This version does not enable full end-to-end resiliency, except for messages
-sent to the <<plugins-inputs-beats,beats>> input. For other inputs, Logstash
-only acknowledges delivery of messages in the filter and output stages, and not
-all the way back to the input or source.
+* Inputs which do not use a request-response protocol cannot be protected from data loss. For example: tcp, udp, zeromq push+pull, and many other inputs do not have a mechanism to acknowledge receipt to the sender. Plugins such as beats and http, which *do* have a acknowledgement capability, are well protected by this queue.
```
jordansissel: Not sure about my wording choice here. Intent was to highlight that protection is only viable for protocols that have a reply mechanism for data. Protocols with no reply mechanism (udp, tcp, etc) would still be queued but we have no way to report back to the sender.

docs/static/persistent-queues.asciidoc
```diff
@@ -53,44 +51,46 @@ process:
...
-the event is successfully sent to the last output stage in the pipeline;
-Logstash does not wait for the output to acknowledge delivery. 
-
-During a normal, controlled shutdown (*CTRL+C*), Logstash finishes
+the persistent queue for processing. Logstash reads a batch of events from the
+persistent queue and sends them to the filter and output stages. As events
+are processed, Logstash records a checkpoint to track which events are
+successfully acknowledged (ACKed) as processed by Logstash. An event is
+recorded as ACKed in the checkpoint if the event is successfully sent to the
+last output stage in the pipeline;
```
dedemorton: This should be a period

docs/static/persistent-queues.asciidoc
```diff
@@ -53,44 +51,46 @@ process:
...
+
+During a normal, controlled shutdown (*CTRL+C* or SIGTERM), Logstash finishes
 processing the current in-flight events (that is, the events being processed by
 the filter and output stages, not the queued events), finalizes the ACKing
 of these events, and then terminates the Logstash process. Upon restart,
-Logstash uses the checkpoint file to pick up where it left off in the persistent
-queue and processes the events in the backlog. 
+Logstash uses last checkpoint to resume processing the events in the backlog.
```
dedemorton: Uses the last checkpoint

docs/static/persistent-queues.asciidoc
```diff
@@ -53,44 +51,46 @@ process:
...
-processed by Logstash. An event is recorded as ACKed in the checkpoint file if
-the event is successfully sent to the last output stage in the pipeline;
-Logstash does not wait for the output to acknowledge delivery. 
-
-During a normal, controlled shutdown (*CTRL+C*), Logstash finishes
+the persistent queue for processing. Logstash reads a batch of events from the
+persistent queue and sends them to the filter and output stages. As events
+are processed, Logstash records a checkpoint to track which events are
+successfully acknowledged (ACKed) as processed by Logstash. An event is
+recorded as ACKed in the checkpoint if the event is successfully sent to the
```
colinsurprenant: «recorded as ACKed in the checkpoint» I would not phrase like this because we actually don't record acking per event in the checkpont - we keep track of the first unacked event in a page . I would just remove "in the checkpoint" - specific ACK tracking IS done in-memory during runtime but not in the checkpoint.

docs/static/persistent-queues.asciidoc
dedemorton: Wording seems clear to me. I would say "Inputs that do not use..." (use "that" for restrictive clauses).

docs/static/persistent-queues.asciidoc
jordansissel: But semicolons are my favorite!

Thanks :)

docs/static/persistent-queues.asciidoc
colinsurprenant: there is not really a "last" checkpoint - there are always 1 checkpoint per page. Do we need to talk about checkpoints here? maybe just saying that logstash will resume processing where it left?

jordansissel: Yeah, the doc had 'checkpoint file' in various places, and I mostly patched this to remove the word 'file' (because that's an implementation detail unimportant to most users)

I think in most case we are trying to convey the concept that a checkpoint can be behind real-time and that the consequence of this is that an abnormal termination will cause data loss for all data written after the most recent checkpoint ("checkpoint" as an action, not a file).

jordansissel: In places where we are describing order of operations, I am wondering if it would be more clear to have a diagram instead of a paragraph of text.

dedemorton: 👍

docs/static/persistent-queues.asciidoc
jordansissel: :heart:

docs/static/persistent-queues.asciidoc
jordansissel: @colinsurprenant I am also with you on maybe not needing to discuss checkpoints in the way this document does today -- I think we should mention it with respect to durability and maybe nowhere else. At present, it's spread around the document and maybe needs to be just in one section.

docs/static/persistent-queues.asciidoc
```diff
@@ -53,44 +51,46 @@ process:
...
 
 To configure persistent queues, you can specify the following options in the
 Logstash <<logstash-settings-file,settings file>>:
 
-* `queue.type`: Specify `persisted` to enable persistent queues. By default, persistent queues are disabled (`queue.type: memory`).
+* `queue.type`: Specify `persisted` to enable persistent queues. By default, persistent queues are disabled (default: `queue.type: memory`).
 * `path.queue`: The directory path where the data files will be stored. By default, the files are stored in `path.data/queue`. 
 * `queue.page_capacity`: The size of the page data file. The queue data consists of append-only data files separated into pages. The default size is 250mb. 
-* `queue.max_events`:  The maximum number of unread events that are allowed in the queue. The default is 0 (unlimited).
+* `queue.max_events`:  The maximum number of unACKed events that are allowed in the queue. The default is 0 (unlimited).
```
colinsurprenant: no, it really *is* the max number of unread events.

jordansissel: In this doc there is no other mention of the word "unread". If we're going to mention it now, it probably needs explanation. It sounds like there's multiple states (read, unread, acked, unacked?) for an event, and that is news to me, can you help me understand more?

docs/static/persistent-queues.asciidoc
colinsurprenant: durability++

docs/static/persistent-queues.asciidoc
colinsurprenant: sure, a queue event can be in the following state:
- unread (which also implies unacked)
- read & unacked (thus "in flight" in the worker) 
- read & acked
Now, read & acked events only exists because purging is done at page level. only when all events within a page are acked is it purged.

docs/static/persistent-queues.asciidoc
jordansissel: @acchen97 I can't tell if the Kafka input actually does this. https://github.com/logstash-plugins/logstash-input-kafka/blob/master/lib/logstash/inputs/kafka.rb#L245-L260 shows no indication that I can see that it does do this acking, but it may do it somewhere deeper inside the Kafka client.

acchen97: @jordansissel I spoke to @talevy and Kafka input does ack.

In this context, I think we can just note Beats, Kafka, RabbitMQ, and HTTP inputs.

jordansissel:
I'm still working on this. Based on feedback and other discussion, I'm going to do some more possibly-larger changes to this doc in this PR. Will update when it's ready for the next round of review.

Thank you all for the feedback so far!

jordansissel:
Ok I did another pass at the content in this document.

I've tried to do the following:

* Focus durability discussion only on the durability section
* Focus on 'ack' terminology
* consistent use of phrase "abnormal termination" when discussing that topic
* Remove "Advantages of Persistent Queue" section since it's just a bullet-point summary of the first section. Moved it into the first section.
* Defined "acknowledge" for pipeline workers, sorta.
* Added caveat on `queue.max_events` as being not intended for user benefit.
* Added necessary background for durability discussion (pages, checkpoints)
* Described garbage collection scenario of fully acked pages.

jordansissel:
@colin @dedemorton @acchen97 Would love your feedback on the current state.

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
-events are left as unACKed in the persistent queue. Upon restart, Logstash will
-replay the events from its history, potentially leading to duplicate data being
-written to the output.
+input → queue → filter + output 
+
+When an input has events ready to process, it writes them to the queue. When
+the write to the queue is successful, the input can send an acknowledgement to
+its data source.
+
+When processing events from the queue, events are only acknowledged as
```
dedemorton: Sentence starts with a dangling modifier (a phrase that modifies a word that isn't clearly stated in the sentence). How about:

When processing events from the queue, Logstash acknowledges events as completed, within the queue, only after the filters and outputs have completed.

[changed to "after" because "once" is easily misread/mistranslated as "one time"]

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+completed, within the queue, once the filters and outputs have completed.
+The queue keeps a record of events which have been processed by the pipeline.
```
dedemorton: "keeps a record of events that have been processed..."

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+An event is recorded as processed (in this document, called "acknowledged" or
+"ACKed") if, and only if, the event has been processed completely by the
+Logstash pipeline. 
+
+What does acknowledged mean? For the common case, this means that it has been
```
dedemorton: I am unsure what you  mean by "common case" here. Do you mean "in most cases"? Also the antecedent for "it" is unclear. Does "it" refer back to the event? If so, how about:

"What does "acknowledged" mean? In most cases, this means that the event has been handled by all configured filters and outputs."

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+handled by all configured filters and outputs. For example, if you have only
+one output, Elasticsearch, an event is only ACKed once the Elasticsearch output
```
dedemorton: ...an event is ACKed when the Elasticsearch output has successfully sent this event to Elasticsearch... [again, "once" is easily misread/mistranslated at "one time"]

docs/static/persistent-queues.asciidoc
```diff
@@ -101,23 +103,24 @@ queue.max_bytes: 4gb
 [[backpressure-persistent-queues]]
 ==== Handling Back Pressure
 
-Logstash has a built-in mechanism that exerts back pressure on the data flow 
-when the queue is full. This mechanism helps Logstash control the rate of data
-flow at the input stage without overwhelming downstream stages and outputs like
+When the queue is full, Logstash puts back pressure on the inputs to stall data
+flowing into Logstash. This mechanism helps Logstash control the rate of data
+flow at the input stage without overwhelming outputs and outputs like
```
dedemorton: You say outputs twice. Maybe you mean to say:

 "...without overwhelming outputs like  Elasticsearch."

docs/static/persistent-queues.asciidoc
```diff
@@ -101,23 +103,24 @@ queue.max_bytes: 4gb
...
 queue.type: persisted
 queue.max_bytes: 8gb
 
-With these settings specified, Logstash will buffer unACKed events on disk until 
-the size of the queue reaches 8gb. When the queue is full of unACKed events, and
+With these settings specified, Logstash will buffer events on disk until the
+size of the queue reaches 8gb. When the queue is full of unACKed events, and
 the size limit has been reached, Logstash will no longer accept new events. 
 
+On disk, the queue is stored as a set of pages where each page is one file. Each page can be at most `queue.page_capacity` in size. Pages are deleted (garbage collected) once all events in that page have been ACKed. If an older page has at least one event that is not yet ACKed, that entire page will remain on disk until all events in that page are successfully processed. Each page containing unprocessed events will count against the `queue.max_bytes` byte size.
```
dedemorton: I wish we didn't have to reveal these details because they seem like internals that users shouldn't need to know about. This paragraph seems like a bit of a non-sequitur. Maybe you just need to add a line that explains how the setting affects back pressure (it's a bit murky to me, TBH).

Also change: "Pages are deleted (garbage collected) **once** " to "Pages are deleted (garbage collected) **after** all events..."

docs/static/persistent-queues.asciidoc
```diff
@@ -128,22 +131,35 @@ events.
...
+// I'm not sure about this particular description of durability. Might need some help.
+Durability is a property of storage writes that indicate the availability of
+that data after it was written.
+
 When the persistent queue feature is enabled, Logstash will store events on
-disk. The persistent queue exposes the trade-off between performance and
-durability by providing the following configuration options:
+disk. Logstash commits to disk in a mechanism we call checkpointing.
+
+To discuss durability, we need to introduce a few details about how the persistent queue is implemented.
```
dedemorton: Hmmm....we're sort of started down this road earlier. I wonder if all the info about pages could be described in one section? Only the most dedicated readers are going to want to plow through these details.

docs/static/persistent-queues.asciidoc
```diff
@@ -128,22 +131,35 @@ events.
...
+
+First, the queue itself is a set of pages. There are two kinds of pages: head page and tail page. The head page is where new events are written. When this head page is of a certain size (see `queue.page_capacity), the becomes a tail page, and a new head page is created. Tail pages are immutable. 
+
+Second, the queue tracks details about the queue (pages, acknowledgements, etc) in a separate file called a checkpoint file.
+
+When Logstash does checkpoint, the following will occur:
```
dedemorton: "When Logstash does checkpoint" sounds a bit off. Maybe "When Logstash records a checkpoint" or "When Logstash does checkpointing".

docs/static/persistent-queues.asciidoc
```diff
@@ -128,22 +131,35 @@ events.
...
+* `queue.checkpoint.acks`: The number of ACKs to the queue after which a checkpoint is written. This configuration controls the durability at the processing (filter + output) part of Logstash.
 
-* `queue.checkpoint.writes`: The number of writes to the queue to trigger an
-fsync to disk. This configuration controls the durability from the producer
-side. Keep in mind that a disk flush is a relatively heavy operation that will
-affect throughput if performed after every write. For instance, if you want to
-ensure that all messages in Logstash's queue are durable, you can set
-`queue.checkpoint.writes: 1`. However, this setting can severely impact
-performance.
+Disk writes are not free have a cost. Tuning these values higher or lower will trade durability for performance. For instance, if you want to ensure that all events from inputs queue are durable, you can set
```
dedemorton: Something is missing or wasn't deleted here: "Disk writes are not free have a cost"

docs/static/persistent-queues.asciidoc
```diff
@@ -128,22 +131,35 @@ events.
...
+Second, the queue tracks details about the queue (pages, acknowledgements, etc) in a separate file called a checkpoint file.
+
+When Logstash does checkpoint, the following will occur:
+
+* The head page is sync'd to disk.
+* A new checkpoint is written atomically to persist the current state of the queue.
+
+The following settings are available to let you tune durability:
+
+* `queue.checkpoint.writes`: The number of writes from inputs after which a checkpoint is written.
```
dedemorton: This description is a bit confusing because of the sentence structure. Does this say what you meant to say?

"The number of times an input writes to the queue before a checkpoint is written?"

docs/static/persistent-queues.asciidoc
```diff
@@ -128,22 +131,35 @@ events.
...
+
+When Logstash does checkpoint, the following will occur:
+
+* The head page is sync'd to disk.
+* A new checkpoint is written atomically to persist the current state of the queue.
+
+The following settings are available to let you tune durability:
+
+* `queue.checkpoint.writes`: The number of writes from inputs after which a checkpoint is written. 
+* `queue.checkpoint.acks`: The number of ACKs to the queue after which a checkpoint is written. This configuration controls the durability at the processing (filter + output) part of Logstash.
```
dedemorton: Does this work?

"The number of ACKs sent to the queue before a checkpoint is written."

docs/static/persistent-queues.asciidoc
dedemorton: I think there is a word missing here (or one that should have been deleted): "For instance, if you want to ensure that all events from inputs queue are durable"

docs/static/persistent-queues.asciidoc
```diff
@@ -128,22 +131,35 @@ events.
...
-side. Keep in mind that a disk flush is a relatively heavy operation that will
-affect throughput if performed after every write. For instance, if you want to
-ensure that all messages in Logstash's queue are durable, you can set
-`queue.checkpoint.writes: 1`. However, this setting can severely impact
-performance.
+Disk writes are not free have a cost. Tuning these values higher or lower will trade durability for performance. For instance, if you want to ensure that all events from inputs queue are durable, you can set
+`queue.checkpoint.writes: 1`.
 
-* `queue.checkpoint.acks`: The number of ACKs to the queue to trigger an fsync to disk. This configuration controls the durability from the consumer side.
+In Logstash 5.0 and 5.1, a `write` to the queue is counted for every event produced by an input.
```
dedemorton: Maybe would be better for info maintainability to say, "Starting with Logstash 5.0...." (Unless you know this behavior is changing after 5.1).

docs/static/persistent-queues.asciidoc
acchen97: I agree with @dedemorton a bit here in that I think this level of detail would be better for a blog post.

docs/static/persistent-queues.asciidoc
```diff
@@ -128,22 +131,35 @@ events.
...
+that data after it was written.
+
 When the persistent queue feature is enabled, Logstash will store events on
-disk. The persistent queue exposes the trade-off between performance and
-durability by providing the following configuration options:
+disk. Logstash commits to disk in a mechanism we call checkpointing.
+
+To discuss durability, we need to introduce a few details about how the persistent queue is implemented.
+
+First, the queue itself is a set of pages. There are two kinds of pages: head page and tail page. The head page is where new events are written. When this head page is of a certain size (see `queue.page_capacity), the becomes a tail page, and a new head page is created. Tail pages are immutable.
```
acchen97: This detail I feel should also be part of a blog post where we talk about PQ implementation architecture and introduce this terminology.

acchen97: We can probably leave it out of the docs.

docs/static/persistent-queues.asciidoc
```diff
@@ -5,72 +5,73 @@ WARNING: This functionality is in beta and is subject to change. Deployment in p
...
-
-* Provides protection from losing in-flight messages when the Logstash process is abnormally terminated.
-* Absorb bursts of events without needing an external queueing mechanism like Redis or Kafka.
-* Depending on configuration, can provide an at-least-once message
-delivery guarantee. If Logstash is restarted while events are in-flight,
-Logstash will attempt to deliver messages stored in the persistent queue until
-delivery succeeds at least once. In other words, messages stored in the
-persistent queue may be duplicated, but not lost.
+* Provides protection from in-flight message loss when the Logstash process is abnormally terminated.
+* Absorb bursts of events without needing an external queueing mechanism like Redis or Apache Kafka.
```
acchen97: I think it's valuable for us to still outline the "at-least-once" messaging here. Maybe another point like this:

"Guarantees at-least-once delivery when used with inputs that support acknowledgements."

jordansissel:
At-least-once isn't quite available yet but is the intent. We found a bug
where the checkpoint is not written to underlying hardware and Colin is (or
has already) working on a fix.

I won't document a detail we can't deliver :)

docs/static/persistent-queues.asciidoc
jordansissel: I would love to avoid writing details about the `page` concept. I don't know how to do this, though, given the behavior of the queue.

The problem is this: A page with *any* unacked events will remain completely on disk. This can mean that for a 250mb page, if only 1KB is unacked, then we will not clean it up. The impact of this capacity planning and behavior for really small queue (byte) sizes, I believe, can become very confusing.

For my concern here, it is mostly predicted user confusion and my strong desire to not omit information that causes users to file bug reports or customers to file support tickets.

docs/static/persistent-queues.asciidoc
jordansissel: Yeah, maybe for `pages` I can have a section dedicated to have data is purged from the queue?

docs/static/persistent-queues.asciidoc
jordansissel: How about "When checkpointing, Logstash does the following:"?

docs/static/persistent-queues.asciidoc
jordansissel: It's confusing because the property of the queue is for all writes, not just from a single input. And internally, it's called a "write" not an "event" which makes this even harder to document.

I really want to say this, "The number of events written to the queue before a checkpoint is written"

But ...

* It's not the number of events, it's number of writes. It just happens that, today, one write is one event.
* and, "before a checkpoint" makes it sound (incorrectly) like that we require *at least* that number of writes before Logstash will write a checkpoint, when really it's a minimum of (number of writes, number of acks) -- whichever happens first.

docs/static/persistent-queues.asciidoc
jordansissel: ++ yeah I didn't clean up my sentence after editing <3

docs/static/persistent-queues.asciidoc
jordansissel: I don't know when this behavior will change, but it's not guaranteed to always be 1:1 and I'm trying (perhaps poorly) to solve this future problem:

* We change the "one write is one event" in the code
* but, we don't change the docs
* and, then a user files a bug about the behavior
* and, then I get annoyed we didn't update the docs.

docs/static/persistent-queues.asciidoc
jordansissel: Pages, I believe, are important, because they matter when Logstash considers which data can be deleted from disk, and I'm trying to document this in enough detail that user questions can be just answered by pointing at the docs.

docs/static/persistent-queues.asciidoc
jordansissel: In 5.1.1 it's not guaranteed because we:

1) `checkpoint.writes: 1` is necessary for such a guarantee and it is not the default.
2) and, we don't fsync the checkpoint file (a bug fixed by #6430)

docs/static/persistent-queues.asciidoc
acchen97: @jordansissel I'm curious in what scenarios this would occur. If all events in a page are successful except for one, won't that event continue to retry for a number of times until it's dropped (and therefore ack'ed)? I feel like we'd be able to solve this altogether when we have DLQs, or introducing something like compaction though that may be overkill.

It sounds like you're positioning this as an unexpected caveat/gotcha to when backpressure may be exerted. Maybe end with "Therefore, if there are unprocessed events in the queue, the 'queue.max_bytes' limit may be hit unexpectedly, exerting backpressure back to the source." Without a mechanism for queue eviction, can users work around this today aside from bumping up their queue.max_bytes?

docs/static/persistent-queues.asciidoc
acchen97: Sounds good. Typo here "*it* becomes a tail page"

acchen97:
@jordansissel thanks for the update, I agree. Let's add back at-least-once after the bug fix and optimistic recovery.

docs/static/persistent-queues.asciidoc
colinsurprenant: the situation where a page is not deleted because all events are not yet acked can happen in these 2 scenarios:
1- outputs are slower than inputs, which will typically result in the accumulation of multiple pages and all events in the page(s) will be unacked until the outputs are ready to process a page. Typically there will be a short overlap when some output start processing a new page and other output finish processing the previous page.
2- an output (or multiple outputs) are "stuck" in a retry loop for some time and some events will remain unacked during this retry period so a page might be kept "alive" (not deleted) because of this particular output "stuck" in retry for some time while other output might actually finish processing the rest of the page and move on to the other page. This situation will be temporary up to be output retry thresholds which is plugin dependant.

docs/static/persistent-queues.asciidoc
colinsurprenant: or "the number or events acknowledged (ACK'ed)  before a checkpoint is written" ?

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
-shutdown, it's possible to lose queued events. 
+(inputs → pipeline workers) to buffer events. The size of these in-memory
+queues is fixed and not configurable. If Logstash terminates abnormally, the
+contents of the in-memory queue will be lost. Such abnormal terminations
+include software crash, `kill -9`, or the host machine loses power.
 
-To prevent event loss in these scenarios, you can configure Logstash to use
-persistent queues. With persistent queues enabled, Logstash persists buffered
-events to disk instead of storing them in memory. 
+Temporary machine failures are scenarios where Logstash or its host machine are
```
jordansissel: I agree.

I've reorganized these two paragraphs to try and have a better flow to it.

docs/static/persistent-queues.asciidoc
jordansissel: ++

docs/static/persistent-queues.asciidoc
jordansissel: ++

docs/static/persistent-queues.asciidoc
jordansissel: ++

docs/static/persistent-queues.asciidoc
jordansissel: ++

docs/static/persistent-queues.asciidoc
jordansissel: I've moved this part about pages and garbage collection into a new section called 'disk garbage collection'

docs/static/persistent-queues.asciidoc
jordansissel: Changed to this:

```
 When recording a checkpoint, Logstash will:

 * Call fsync on the head page.
 * Atomically write to disk the current state of the queue.
```

docs/static/persistent-queues.asciidoc
jordansissel: I created a separate section for page/garbage details.

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+
+In order to protect against data loss during abnormal termination, Logstash has
+a persistent queue feature which will store the message queue on disk.
+Persistent queues provides durability of data within Logstash.
+
+Persistent queues are also useful for Logstash deployments that need large buffers  
+Instead of deploying and managing a message broker, such as Redis, RabbitMQ, or
+Apache Kafka, to facilitate a buffered publish-subscriber model, you can enable
+persistent queues to buffer events on disk and remove the message broker. This
+benefits you by removing one more piece of machinery from your deployment.
```
colinsurprenant: I'm not sure I like the wording "removing one more piece of machinery"  - it feels like a message broker necessarily involves extra hardware (machinery?) but it might not ... 
Anyhow it may just be a language thing so please disregard my comment if ya'll are comfortable with this :D

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+An event is recorded as processed (in this document, called "acknowledged" or
+"ACKed") if, and only if, the event has been processed completely by the
+Logstash pipeline. 
+
+What does acknowledged mean? For most cases, this means that it has been
+handled by all configured filters and outputs. For example, if you have only
+one output, Elasticsearch, an event ACKed when the Elasticsearch output has
+successfully sent this event to Elasticsearch. In other cases, such as
+with the `drop` filter, an event is ACKed when the `drop` filter cancels the
+event.
```
colinsurprenant: «an event is ACKed when the `drop` filter cancels the event»  this is not true. all events in a batch are acked once the batch is done being processed by the filters+outputs worker, thus a cancelled event will actually be acked at the same time as all the other events in the batch.

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
 Logstash <<logstash-settings-file,settings file>>:
 
-* `queue.type`: Specify `persisted` to enable persistent queues. By default, persistent queues are disabled (`queue.type: memory`).
+* `queue.type`: Specify `persisted` to enable persistent queues. By default, persistent queues are disabled (default: `queue.type: memory`).
 * `path.queue`: The directory path where the data files will be stored. By default, the files are stored in `path.data/queue`. 
-* `queue.page_capacity`: The size of the page data file. The queue data consists of append-only data files separated into pages. The default size is 250mb. 
-* `queue.max_events`:  The maximum number of unread events that are allowed in the queue. The default is 0 (unlimited).
+* `queue.page_capacity`: The maximum size of a queue page in bytes. The queue data consists of append-only files called a "pages". The default size is 250mb. Changing this value is unlikely to have performance benefits.
+// Technically, I know, this isn't "maximum number of events" it's really maximum number of events not yet read by the pipeline worker. We only use this for testing and users generally shouldn't be setting this.
```
colinsurprenant: «append-only files called a "pages"» -> append-only files called "pages"  ?

docs/static/persistent-queues.asciidoc
```diff
@@ -128,24 +128,41 @@ events.
...
 
 If Logstash is terminated, or if there is a hardware level failure, any data
 that is buffered in the persistent queue, but not yet checkpointed, is lost.
 To avoid this possibility, you can set `queue.checkpoint.writes: 1`, but keep in
 mind that this setting can severely impact performance.
+
+[[garbage-collection]]
+==== Disk Garbage Collection
+
+On disk, the queue is stored as a set of pages where each page is one ile. Each page can be at most `queue.page_capacity` in size. Pages are deleted (garbage collected) once all events in that page have been ACKed. If an older page has at least one event that is not yet ACKed, that entire page will remain on disk until all events in that page are successfully processed. Each page containing unprocessed events will count against the `queue.max_bytes` byte size.
```
colinsurprenant: one ile -> one file

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
-messages stored in the persistent queue may be duplicated, but not lost.
+(inputs → pipeline workers) to buffer events. The size of these in-memory
+queues is fixed and not configurable. If Logstash experiences a temporary
+machine failure, the contents of the in-memory queue will be lost. Temporary machine
+failures are scenarios where Logstash or its host machine are terminated
+abnormally but are capable of being restarted. 
+
+In order to protect against data loss during abnormal termination, Logstash has
+a persistent queue feature which will store the message queue on disk.
+Persistent queues provides durability of data within Logstash.
```
jordansissel: ++

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+
+Persistent queues are also useful for Logstash deployments that need large buffers
```
jordansissel: ++

docs/static/persistent-queues.asciidoc
jordansissel: Noted. When I reread this paragraph, I think the last sentence is unnecessary anyway so I will remove it. Thanks <3

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+a persistent queue feature which will store the message queue on disk.
+Persistent queues provides durability of data within Logstash.
+
+Persistent queues are also useful for Logstash deployments that need large buffers  
+Instead of deploying and managing a message broker, such as Redis, RabbitMQ, or
+Apache Kafka, to facilitate a buffered publish-subscriber model, you can enable
+persistent queues to buffer events on disk and remove the message broker. This
+benefits you by removing one more piece of machinery from your deployment.
+
+In summary, the two benefits of enabling the persistent queue are as follows:
```
jordansissel: ++

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+
+* Provides protection from in-flight message loss when the Logstash process is abnormally terminated.
+* Absorb bursts of events without needing an external buffering mechanism like Redis or Apache Kafka.
```
jordansissel: ++

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+its data source.
+
+When processing events from the queue, Logstash acknowledges events as
+completed, within the queue, only after filters and outputs have completed.
+The queue keeps a record of events that have been processed by the pipeline.
+An event is recorded as processed (in this document, called "acknowledged" or
+"ACKed") if, and only if, the event has been processed completely by the
+Logstash pipeline. 
+
+What does acknowledged mean? For most cases, this means that it has been
```
jordansissel: Ahh, great point. Thank you :)

docs/static/persistent-queues.asciidoc
jordansissel: Agreed.

docs/static/persistent-queues.asciidoc
```diff
@@ -1,96 +1,98 @@
...
+handled by all configured filters and outputs. For example, if you have only
+one output, Elasticsearch, an event ACKed when the Elasticsearch output has
```
jordansissel: ++

docs/static/persistent-queues.asciidoc
jordansissel: Oops. Good catch.

docs/static/persistent-queues.asciidoc
```diff
@@ -128,24 +128,41 @@ events.
 [[durability-persistent-queues]]
 ==== Controlling Durability
 
+// I'm not sure about this particular description of durability. Might need some help.
+Durability is a property of storage writes that indicate the availability of
```
jordansissel: Thank you! I wasn't sure how to word that.

docs/static/persistent-queues.asciidoc
```diff
@@ -128,24 +128,41 @@ events.
...
+that data after it was written.
+
 When the persistent queue feature is enabled, Logstash will store events on
-disk. The persistent queue exposes the trade-off between performance and
-durability by providing the following configuration options:
+disk. Logstash commits to disk in a mechanism we call checkpointing.
```
jordansissel: ++

docs/static/persistent-queues.asciidoc
```diff
@@ -128,24 +128,41 @@ events.
...
+
+To discuss durability, we need to introduce a few details about how the persistent queue is implemented.
+
+First, the queue itself is a set of pages. There are two kinds of pages: head page and tail page. The head page is where new events are written. When this head page is of a certain size (see `queue.page_capacity), the becomes a tail page, and a new head page is created. Tail pages are immutable, and the head page is append-only..
```
jordansissel: ++ to all.

docs/static/persistent-queues.asciidoc
```diff
@@ -128,24 +128,41 @@ events.
...
 When the persistent queue feature is enabled, Logstash will store events on
-disk. The persistent queue exposes the trade-off between performance and
-durability by providing the following configuration options:
+disk. Logstash commits to disk in a mechanism we call checkpointing.
+
+To discuss durability, we need to introduce a few details about how the persistent queue is implemented.
+
+First, the queue itself is a set of pages. There are two kinds of pages: head page and tail page. The head page is where new events are written. When this head page is of a certain size (see `queue.page_capacity), the becomes a tail page, and a new head page is created. Tail pages are immutable, and the head page is append-only.. 
+
+Second, the queue tracks details about the queue (pages, acknowledgements, etc) in a separate file called a checkpoint file.
```
jordansissel: ++

docs/static/persistent-queues.asciidoc
```diff
@@ -128,24 +128,41 @@ events.
...
-performance.
+* Call fsync on the head page.
+* Atomically write to disk the current state of the queue.
 
-* `queue.checkpoint.acks`: The number of ACKs to the queue to trigger an fsync to disk. This configuration controls the durability from the consumer side.
+The following settings are available to let you tune durability:
 
-The process of checkpointing is atomic, which means any update to the file is
-saved if successful.
+* `queue.checkpoint.writes`: Logstash will checkpoint after this many writes into the queue. At this time, one event counts as one write, but this may change in future releases.
```
jordansissel: Your explanations are delightfully informative. Thank you!

docs/static/persistent-queues.asciidoc
```diff
@@ -128,24 +128,41 @@ events.
...
+* `queue.checkpoint.acks`: Logstah will checkpoint after this many events are acknowledged. This configuration controls the durability at the processing (filter + output)
+part of Logstash.
+
+Disk writes are have a resource cost. Tuning the above values higher or lower will trade durability for performance. For instance, if you want to the strongest durability for all input events, you can set `queue.checkpoint.writes: 1`.
```
jordansissel: You're right, I should contract it to `are've` an obvious contraction of "are" and "have"..

haha ;)

docs/static/persistent-queues.asciidoc
jordansissel: ++

jordansissel: ++

jordansissel:
OK! I think I've resolved all the feedback now.

I've rebased this into a single commit and am going to merge.

Thank you for helping copy edit this important document <3 <3

PR #6413, opened 2016-12-14
Various LIR improvements

andrewvc:
Many refactors
Rebase off newer master
Fixed an issue with dangling leaf nodes.
Improved Diff Reports
Various pieces of code cleanup.
More consistent ID generation.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/Util.java
+++ b/logstash-core/src/main/java/org/logstash/common/Util.java
@@ -1,0 +1,33 @@
+package org.logstash.common;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * Created by andrewvc on 12/23/16.
+ */
+public class Util {
+    // Modified from http://stackoverflow.com/a/11009612/11105
+    public static String sha256(String base) {
+        try {
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            byte[] hash = digest.digest(base.getBytes(StandardCharsets.UTF_8));
+            return bytesToHexString(hash);
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException("Your system is (somehow) missing the SHA-256 algorithm!", e);
+        }
+    }
+
+    public static String bytesToHexString(byte[] bytes) {
+        StringBuilder hexString = new StringBuilder();
+
+        for (byte aHash : bytes) {
+            String hex = Integer.toHexString(0xff & aHash);
+            if (hex.length() == 1) hexString.append('0');
+            hexString.append(hex);
+        }
+
+        return hexString.toString();
+    }
+}
--- a/logstash-core/src/main/java/org/logstash/config/ir/Pipeline.java
+++ b/logstash-core/src/main/java/org/logstash/config/ir/Pipeline.java
@@ -12,7 +12,7 @@ import java.util.stream.Stream;
 /**
  * Created by andrewvc on 9/20/16.
  */
-public class Pipeline {
+public class Pipeline implements IHashable {
     public Graph getGraph() {
         return graph;
     }
@@ -30,32 +30,23 @@ public class Pipeline {
     private final SpecialVertex filterOut;
 
     public Pipeline(Graph inputSection, Graph filterSection, Graph outputSection) throws InvalidIRException {
-        graph = DSL.graph();
-
-        // We don't really care about the edges in the input section, we just want the vertices as roots
-        for (Vertex inV : inputSection.getVertices()) {
-            if (inV instanceof PluginVertex) {
-                graph.addVertex(inV);
-            } else {
-                throw new InvalidIRException("Only plugin vertices are allowed in input sections!");
-            }
-        }
+        Graph tempGraph = inputSection.copy(); // The input section are our roots, so we can import that wholesale
 
         // Connect all the input vertices out to the queue
         queue = new SpecialVertex(SpecialVertex.Type.QUEUE);
-        graph.threadLeavesInto(queue);
+        tempGraph = tempGraph.chain(queue);
 
         // Now we connect the queue to the root of the filter section
-        graph.threadLeavesInto(filterSection);
+        tempGraph = tempGraph.chain(filterSection);
 
         // Now we connect the leaves (and partial leaves) of the graph
         // which should all be filters (unless no filters are defined)
         // to the special filterOut node
         filterOut = new SpecialVertex(SpecialVertex.Type.FILTER_OUT);
-        graph.threadLeavesInto(filterOut);
+        tempGraph = tempGraph.chain(filterOut);
 
         // Finally, connect the filter out node to all the outputs
-        graph.threadLeavesInto(outputSection);
+        this.graph = tempGraph.chain(outputSection);
     }
 
     public List<Vertex> getPostQueue() throws InvalidIRException {
@@ -103,5 +94,10 @@ public class Pipeline {
         return graph.vertices()
                .filter(v -> v instanceof PluginVertex)
                .map(v -> (PluginVertex) v);
+    }
+
+    @Override
+    public String hashSource() {
+        return this.graph.uniqueHash();
     }
 }
--- a/logstash-core/src/main/java/org/logstash/config/ir/SourceMetadata.java
+++ b/logstash-core/src/main/java/org/logstash/config/ir/SourceMetadata.java
@@ -1,5 +1,7 @@
 package org.logstash.config.ir;
 
+import java.util.Objects;
+
 /**
  * Created by andrewvc on 9/6/16.
  */
@@ -38,6 +40,10 @@ public class SourceMetadata {
         this.sourceLine = null;
         this.sourceColumn = null;
         this.sourceText = null;
+    }
+
+    public int hashCode() {
+        return Objects.hash(this.sourceFile, this.sourceLine, this.sourceColumn, this.sourceText);
     }
 
     public String toString() {
--- a/logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
+++ b/logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
@@ -1,75 +1,173 @@
 package org.logstash.config.ir.graph;
 
+import org.logstash.config.ir.IHashable;
 import org.logstash.config.ir.ISourceComponent;
 import org.logstash.config.ir.InvalidIRException;
-import org.logstash.config.ir.PluginDefinition;
 import org.logstash.config.ir.SourceMetadata;
+import org.logstash.config.ir.graph.algorithms.BreadthFirst;
+import org.logstash.config.ir.graph.algorithms.GraphDiff;
 
+import java.lang.reflect.Array;
 import java.util.*;
-import java.util.function.Consumer;
+import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
  * Created by andrewvc on 9/15/16.
  */
-public class Graph implements ISourceComponent {
+public class Graph implements ISourceComponent, IHashable {
     private final Set<Vertex> vertices = new HashSet<>();
     private final Set<Edge> edges = new HashSet<>();
+    private Map<Vertex, Integer> vertexRanks = new HashMap<>();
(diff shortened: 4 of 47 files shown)


logstash-core/src/main/java/org/logstash/config/ir/graph/IfVertex.java
```diff
@@ -20,6 +21,9 @@ public BooleanExpression getBooleanExpression() {
 
     public IfVertex(SourceMetadata meta, BooleanExpression booleanExpression) {
         super(meta);
+        this.id = "if-" +
+                (meta != null ? Integer.toString(meta.hashCode()) : UUID.randomUUID().toString())
```
andrewvc: This is intended to provide a consistent ID for if conditions. Maybe this isn't necessary, but it seems to be useful. Ideally we'd have a user-specified ID here.

logstash-core/src/main/java/org/logstash/config/ir/SourceMetadata.java
```diff
@@ -40,6 +48,10 @@ public SourceMetadata() {
         this.sourceText = null;
     }
 
+    public int hashCode() {
+        return Objects.hash(this.sourceFile, this.sourceLine, this.sourceColumn, this.sourceText);
+    }
+
     public String toString() {
         return sourceFile + ":" + sourceLine + ":" + sourceColumn + ":```\n" + sourceText + "\n```";
     }
```
ph: We have decided to extract the `@JsonProperty` Does this PR need to be updated with that logic or we want to make it another PR for that.

logstash-core/src/main/java/org/logstash/config/ir/graph/DiffResult.java
```diff
@@ -0,0 +1,75 @@
+package org.logstash.config.ir.graph;
```
ph: Thanks for extracting this into his own class, I know this is mostly for debugging but can we have 2-3 unit test to make sure we don't regress it? I am pretty sure we improve this classe for other stuff.

logstash-core/src/main/java/org/logstash/config/ir/graph/DiffResult.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public DiffResult(Collection<Edge> removedEdges, Collection<Edge> addedEdges, List<Vertex> removedVertices, List<Vertex> addedVertices) {
+        this.removedEdges = removedEdges;
+        this.addedEdges = addedEdges;
+        this.removedVertices = removedVertices;
+        this.addedVertices = addedVertices;
+    }
+
+    public String summary() {
+        String template = "(-%d,+%d Edges | -%d,+%d Vertices)";
```
ph: Maybe move this to a constant?

logstash-core/src/main/java/org/logstash/config/ir/graph/IfVertex.java
ph: I think this implementation is OK for now.

The only down side I currently see the implementation is tied to the ruby language implementation.
Maybe we should hide that generation logic behind a method that we can call on both BooleanExpression and IfVertex this will give us more future flexibility? WDYT?

logstash-core/src/main/java/org/logstash/config/ir/graph/IfVertex.java
```diff
@@ -59,6 +63,16 @@ public boolean acceptsOutgoingEdge(Edge e) {
...
 
+    @Override
+    public String getId() {
+        return id;
+    }
+
+    @Override
+    public String typeString() {
+        return "conditional";
+    }
```
ph: Since the possibility are limited should we use an Enum here?

logstash-core/src/test/java/org/logstash/config/ir/IRHelpers.java
```diff
@@ -55,4 +71,18 @@ public static SourceMetadata testMetadata() {
...
+                                    iPlugin(FILTER, "grok"),
+                                    iPlugin(FILTER, "kv")).toGraph();
+        Graph outputSection = iIf(eGt(eEventValue("[baz]"), eValue(1000)),
+                                    iComposeParallel(
+                                            iPlugin(OUTPUT, "s3"),
+                                            iPlugin(OUTPUT, "elasticsearch")),
+                                    iPlugin(OUTPUT, "stdout")).toGraph();
+
+        return new Pipeline(inputSection, filterSection, outputSection);
+    }
```
ph: Where is this `samplePipeline` used? I don't see any reference to it in the current code?

logstash-core/src/main/java/org/logstash/config/ir/graph/IfVertex.java
andrewvc: @ph how is it tied to the ruby lang implementation?

One downside here is that the hashCode isn't really unique enough. Java hash codes aren't intended to prevent collisions. I should probably change that.

logstash-core/src/main/java/org/logstash/config/ir/graph/DiffResult.java
andrewvc: There's no need in Java due to the [String Pool](http://stackoverflow.com/questions/2486191/what-is-the-java-string-pool-and-how-is-s-different-from-new-strings)

ph: TIL

logstash-core/src/main/java/org/logstash/config/ir/graph/IfVertex.java
ph: > booleanExpression.toRubyString()

`toRubyString` feel specific to ruby ?:)

andrewvc:
@ph the ID generation here I came up with is really crappy. I'm going to instead generate real SHA256 hashed ids consistently from subparts of the tree. That way you can compare any two subtrees / graphs etc. It's just the right way to do this that isn't all hacky. Will take a little while longer.

logstash-core/src/test/java/org/logstash/config/ir/IRHelpers.java
andrewvc: It wasn't used before but is now in a new test!

andrewvc:
@ph this is ready for review again.

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -1,182 +1,294 @@
...
     }
 
+    private void calculateRanks() {
+        vertexRanks = BreadthFirst.breadthFirst(this.getRoots()).vertexDistances;
+    }
+
+    public Integer rank(Vertex vertex) {
+        Integer rank = vertexRanks.get(vertex);
+        // This should never happen
+        if (rank == null) throw new RuntimeException("Attempted to get rank from vertex where it is not yet calculated!" + this);
```
ph: missing space after quote?

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -1,182 +1,294 @@
...
+            String joinedErrorMessageGroups = errorMessageGroups.collect(Collectors.joining("\n---\n"));
 
-    // Walk the graph, visiting each edge on it with the provided Consumer<Edge>
-    public void walkEdges(Consumer<Edge> consumer) {
-        // Avoid stream interface to avoid concurrency issues if a new root is added
-        // since streams are lazy you can't mutate the graph while walking it with a
-        // stream. Using this iterative approach we can allow such operations.
-        for (Vertex root : this.getRoots()) {
-            walkEdges(consumer, root);
+            throw new InvalidIRException("Some nodes on the graph are fully redundant!\n" + joinedErrorMessageGroups);
```
ph: Thanks for all the efforts in the validation!

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -341,21 +418,16 @@ public boolean isEmpty() {
...
 
     public Stream<Edge> edges() {
         return this.edges.stream();
     }
+
+    @Override
+    public String hashSource() {
+        return this.vertices.stream().map(Vertex::hashSource).sorted().collect(Collectors.joining("\n"));
+    }
 }
```
ph: I like the proposed style `getRoots` vs `roots` for streaming interface.

logstash-core/src/main/java/org/logstash/common/Util.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Created by andrewvc on 12/23/16.
+ */
+public class Util {
+    // Modified from http://stackoverflow.com/a/11009612/11105
+    public static String sha256(String base) {
+        try{
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            byte[] hash = digest.digest(base.getBytes("UTF-8"));
```
andrewvc: Nice! Did not know...

logstash-core/src/main/java/org/logstash/common/Util.java
```diff
@@ -0,0 +1,31 @@
...
+            return bytesToHexString(hash);
+        } catch(Exception ex){
+            throw new RuntimeException(ex);
+        }
+    }
+
+    public static String bytesToHexString(byte[] bytes) {
```
andrewvc: Hmmm, that's awesome, but I don't know if I want to tie this to the java XML processing functions.

logstash-core/src/main/java/org/logstash/config/ir/Pipeline.java
```diff
@@ -30,32 +30,23 @@ public SpecialVertex getFilterOut() {
...
         // Now we connect the leaves (and partial leaves) of the graph
         // which should all be filters (unless no filters are defined)
         // to the special filterOut node
         filterOut = new SpecialVertex(SpecialVertex.Type.FILTER_OUT);
-        graph.threadLeavesInto(filterOut);
+        tempGraph = tempGraph.chain(filterOut);
 
         // Finally, connect the filter out node to all the outputs
-        graph.threadLeavesInto(outputSection);
+        this.graph = tempGraph.chain(outputSection);
```
andrewvc: Yeah, took a while to get the naming right.

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -1,182 +1,294 @@
...
+        // If this belongs to another graph use a copy
+        if (v.getGraph() != null && v.getGraph() != this) {
+            throw new InvalidIRException("Attempted to add vertex already belonging to a graph!");
+        }
+
+        v.setGraph(this);
+
         this.vertices.add(v);
         this.refresh();
+    }
```
andrewvc: YES! Good catch!

andrewvc: Now fixed

logstash-core/src/main/java/org/logstash/config/ir/graph/Vertex.java
```diff
@@ -94,14 +93,87 @@ public boolean hasOutgoingEdges() {
...
+        return this.getGraph().getIncomingEdges(this).stream();
     }
 
     public Stream<Edge> outgoingEdges() {
-        return outgoingEdges.stream();
+        return this.getGraph().getOutgoingEdges(this).stream();
+    }
+
+    public Stream<Vertex> ancestors() {
+        // Without all the distinct calls this can be slow
```
andrewvc: Oh, I previously had a different algorithm. I'll kill this comment.

logstash-core/src/main/java/org/logstash/config/ir/graph/algorithms/ShortestPath.java
```diff
@@ -0,0 +1,88 @@
...
+import org.logstash.config.ir.graph.Vertex;
+
+import java.util.*;
+import java.util.stream.Stream;
+
+/**
+ * Created by andrewvc on 1/5/17.
+ * This class isn't currently used anywhere, I wrote it for some code that is now removed, however, I'm sure it will be
+ * useful shortly, so we should hold onto it for a while.
+ */
```
andrewvc: Well, I have tests for it, so I'm not too worried about that TBH.

ph:
Lets close this PR since its merge in the feature branch.

PR #6419, opened 2016-12-15
[WIP] Introduce Dead Letter Queue Feature

talevy:
# Dead Letter Queue

From the start of time, Logstash either would hang on or drop events that were not 
successfully processed. This is a real bummer. What if a simple mistake was made that can be
remedied by an extra filter phase, or maybe a conditional on sending to a different output? The goal of the DLQ is to provide a solution for what to do with these events. A solution that persists these events for future handling.

## APIs and Plugins

To provide a mechanism for re-processing dropped events, a few APIs are introduced into Logstash. One is the Plugin API that actually pushes failed events processed by specific plugins into the DLQ. Another API is the Close API that allows users to instruct Logstash to snapshot the current DLQ's state into a separate directory for re-processing by a new logstash-input-dlq plugin.

### Plugin API

Plugins will be provided a `dlq_commit(event, reason)` method where plugin authors can pass in the event to persist in the DLQ and a string explanation of why the event was dropped. A new dlq-entry will be inserted into the DLQ containing more information about the event plugin context like `plugin_id`, `plugin_name`. This information can be used by other logstash pipelines in conjunction with the logstash-input-dlq plugin to appropriately filter and handle dropped events.

### Logstash DLQ APIs

```
POST logstash:9600/dlq/_rollover/NAME_OF_DLQ
```

To make a DLQ open for re-processing, it must be closed. With the above API call, you can trim and name the current DLQ. This will make it accessible for the logstash-input-dlq plugin for reading.

```
GET logstash:9600/dlq/NAME_OF_DLQ
```

It is difficult to know how to build a pipeline using the dlq input if you do not know what is in the DLQ. Using the GET api, you can fetch statistics of what type of plugins resulted in what type of events (and how many).

```
DELETE logstash:9600/dlq/MY_DLQ_NAME
```

Once you are done processing a specific DLQ, you can delete it using the DELETE API as shown above.

### Dead Letter Queue Input Plugin (separate plugins repo)

PR link: https://github.com/logstash-plugins/logstash-input-dlq/pull/1

Here is an example pipeline that would read from a closed DLQ. In this case, it will index all the events with mapping exception to a separate index in Elasticsearch called "failed_to_map"
```ruby
input {
  dlq {
    name => "MY_DLQ_NAME"
  }
}

output {
  if [@metadata][dlq][plugin_name] == "elasticsearch" && [@metadata][dlq][reason] == "mapping_exception" {
    elasticsearch {
      index => "failed_to_map"
    }
  }
}
```

### TODOs

- [ ] RATS tests
- [ ] introduce a read client for the Queue to enable reading the `_current` DLQ without side effects.
- [ ] test how long it takes to `Queue#open` a 10GB queue.
- [ ] filter-multiline is used in core-tests, this change (or something like it) needs to be done: https://github.com/logstash-plugins/logstash-filter-multiline/pull/32


config/logstash.yml
```diff
@@ -131,6 +131,16 @@
...
 #
+# ------------ Dead-Letter Queue Settings --------------
+# Flag to turn on dead-letter queue.
+#
+# dlq.enable: false
+#
+# If using dlq.enable: true, the directory path where the data files will be stored.
+# Default is path.data/dlq
+#
+# path.dlq: false
```
suyograo: I know it's WIP, but just an early feedback -- let's expand this as `path.dead_letter_queue` instead of `path.dlq`

suyograo: Just for user facing stuff, code references are not important.

talevy: you're right, updated.

suyograo:
@talevy how about modifying the APIs to:

* add 'snapshot' name to make it more explicit that it is a point in time snapshot of the DLQ
`POST logstash:9600/_dlq/snapshot/<MY_DLQ_NAME>`

* DELETE logstash:9600/_dlq/MY_DLQ_NAME

Add: Get a list of DLQs and when it was snapshot'd details (all info from filesystem metadata)

* GET logstash:9600/_dlq

RESPONSE

```js
dlq: [{
   "name": "dlq-20160707"
   "size_in_bytes": 
   "snapshot_time":
  },
 {
   "name": "dlq-20160709"
   "size_in_bytes": 
   "snapshot_time":
  },
  ...
```

-------------------------------------------
### Dead Letter Queue Input Plugin

For this plugin, an important aspect is that we don't garbage collect existing pages which allow users to run multiple iterations of DLQ-logic on top of it.

Also, shouldn't this input include a DLQ path?

```
input {
  dlq {
    name => "MY_DLQ_NAME"
    path => "/var/logstash/data/dlq
  }
}
```

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -1,4 +1,4 @@
-# encoding: utf-8
+# kncoding: utf-8
```
jsvd: typo here

talevy: vim helped out here

talevy:
@colin added you as a reviewer to the DLQ PR, still missing tests. My main concern is whether I got the API right. There are other annoying things like the test dependencies on `filter-mutate`

logstash-core/lib/logstash/dlq_manager.rb
```diff
@@ -0,0 +1,87 @@
...
+      raise Error, "cannot delete the current queue"
+    end
+
+    queue = @queues.delete(name)
+    queue.close
+    FileUtils.rm_rf(::File.join(@path, name))
+    true
+  end
+
+  def write_client
```
talevy: TODO: this queue handle may be used during a rollover, in which case, the underlying queue will be `nil`. need to wrap in mutex so that a rollover and a write do not happen at the same time

talevy:
Closing PR in favor of a set of PRs that will close: https://github.com/elastic/logstash/issues/6700

PR #6430, opened 2016-12-16
add explit fsync on checkpoint write

colinsurprenant:
Add explit fsync on checkpoint write. We assumed that the `Files.write` open/write/close implied fsync'ing which it does not.


talevy:
LGTM

jordansissel:
LGTM

Issue #6085, opened 2016-10-18
queue drainage options at shutdown


colinsurprenant:
Historically logstash, for obvious reasons, always drained its internal in-memory queue(s) upon shutdown. 

Now with the persistent queue, this is not required anymore, a clean shutdown can happen at any time and the queue does not require to be drained as persisted events will be picked up where they were left upon restarting logstash. This is the current implementation in #5958. 

I think we should also support an option where logstash can wait until the queue is drained before completing its shutdown, similar to the legacy implementation. One of the problem is with the input plugins which can initiate a shutdown like the generator input: For a given configuration, if filter+output is slower, the input triggered shutdown will leave unprocessed events in the queue and this can be counter intuitive in a "getting started" scenario or create weird problems in specs like in #6084
- What should be the default behaviour? In a "getting started" use-case it is probably expected that all events will be processed before completing a shutdown
- In production use-case I believe a quick-clean shutdown is a better default as the queue could/will grow to a large number of events and waiting for it to drain could be a problem. 
- If we ever switch the in-memory queuing to use the in-memory PQ backend instead of the current queuing implementation, this option should definitely be enabled to respect the in-memory queuing behaviour to not loose inflight event upon shutdown.

Thoughts?

suyograo:
There are only a couple of plugins that initiate a shutdown like generator and ES input. Generator shouldn't be used in production scenarios, and in general used only for testing. ES input can be changed to a streaming case. In general, exposing too many options will just confuse the users and make it really hard to support.  

IMO, for the first iteration:
* In-memory queue: On by default, so we stick to the same semantics. Drain queue before shutting down
* Disk-backed queue: First iteration, do not support wait-until-drain during shutdown. This works nicely with the variable queue semantic. If a user is really interested in draining events in the queue, they have to wait. Later on, if enough users ask for it, we can expose a flag to force drain before shutdown. If this workflow messes up with our testing, we can add a setting but not document/expose it to users.

jsvd:
I'm in favor of not suprising the user for 5.1.0, which means using the in-memory queue (not the in-memory "PQ"), and force the queue drain before shutdown/reload.

The only question for me is: should the drain semantic be implicit or explicit depending on the queue?
If it is implicit then using the in-memory means the drain must happen (and using the PQ means it will not happen), if it's explicit then we need another setting like `drain_on_shutdown` which defaults to draining.

suyograo:
[...]

Issue #6431, opened 2016-12-16
Source repo structure change proposal


jordansissel:
This issue is only intended to impact the source repository and will not impact the logstash artifacts.

We (logstash team; discussed with @colinsurprenant, @ph, @talevy, @jordansissel), have an idea to restructure the files in the source repository.

Goals:
* Merge all component-like things into a single source tree (logstash-core, logstash-core-event, etc)
* Make it easier to compile things (one java source tree; one build.gradle)
* Make it easier to find a given piece of code (instead of looking across 5 components directories)
* Keep general expectations of Ruby and Java programmers in mind.

Here is the proposal in english:

* Move all ruby code into `lib/`.
* Move all java code into `src/{test,main}/java`
* Move `rakelib` into `lib/rakelib`
* Bring `devutils` into the logstash repo following the rules above.
* Move all ruby specs into `spec/`

This will remove the following directories:

* `logstash-core/`
* `logstash-core-event/`
* `logstash-core-event-java/`
* `logstash-core-plugin-api/`
* `logstash-core-queue-jruby/`

The following will not change:

* `docs/`
* `bin/`

Optional, undecided moves:

* `ci/`
* `pkg/`
* `qa/`
* `tools/`

jordansissel:
PR tracking this is here: https://github.com/elastic/logstash/pull/6432

I stress that this PR exists to track this issue's proposal and will be updated accordingly as feedback comes through this issue.

suyograo:
> Optional, undecided moves:

I don't think we should change ci, pkg, qa and tools dir. They seem to be in the right place.

jordansissel:
@suyograo I want to move ci/pkg/qa/tools, and will propose it eventually, but I will make that proposal *after* this issue is resolved so we don't end up in a sad place with code ;)

jordansissel:
@talevy the work I've started on this I have put into this branch: https://github.com/elastic/logstash/tree/master-reorg

Current status of this branch: It is incomplete. `rake test:install-core` works, but `rake test:core` fails because log4j2 is not found.

You are welcome to totally ignore the `master-reorg` branch and start over -- what I've done so far is not much hard work.

suyograo:
This has been merged with #5348 and most of the directory renaming has been done. Closing for now.

PR #6433, opened 2016-12-16
[WIP] queue drain option support

colinsurprenant:
This is in [WIP] state because the metrics on empty batch/queue need to be assessed - The creation of an empty final batch to handle the final flush upon shutdown which did not contain metrics yielded some errors so I added metrics to empty batches. There was originally a check for empty batches read to not start metrics but I believe this was bug because if this empty batch got used in a flush signal (slim chance) it would have crashed because of unstarted metrics for that batch. 

Core tests/specs are passing locally.

This PR adds support for a queue drain option. It relates to issue #6085 and it is necessary for the upcoming BatchedQueue implementation. The goal is to be able to signal logstash to drain the queue before exiting on a normal shutdown.  It is necessary for in-memory queueing support where events can be buffered. 

- It refactors the Wrapped xxx Queue  ReadClient classes to not read the next batch directly in the constructor but instead expose a `read_batch` method so it is now possible to create a new empty batch. 

- The pipeline worker loop becomes a bit clearer to understand with regard to signal handling (shutdown & flush) and shutdown handling for final flush in particular. 

- By default this option is currently turned off and can be turned on in the config using `queue.drain: true`. 

TODO
- [ ] add tests/specs for the drain support.
- [ ] clarify the empty batch/queue metrics situation

### DIFF
--- a/logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueExtLibrary.java
+++ b/logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueExtLibrary.java
@@ -7,6 +7,7 @@ import org.jruby.RubyClass;
 import org.jruby.RubyFixnum;
 import org.jruby.RubyModule;
 import org.jruby.RubyObject;
+import org.jruby.RubyBoolean;
 import org.jruby.anno.JRubyClass;
 import org.jruby.anno.JRubyMethod;
 import org.jruby.runtime.Arity;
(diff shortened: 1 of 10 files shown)


guyboertje:
Looks good to me so far. I was planning to port the Q clients to Java/JRuby soon anyway. I would rather do the port when the source is restructured i.e. #6431

ph:
Sorry didn't notice I need to review this, I will take a look asap.

colinsurprenant:
I rebased and would like to reboot this PR.

colinsurprenant:
@ph ok found and fixed the spec problems, some where related to empty batch handling as discussed, others related to the metrics start where my refactor had the call to `add_starting_metrics(batch)` right after the new empty batch creation but it needed to be after the `read_batch`.

colinsurprenant:
Build successful! 🎉 
@ph final review?

colinsurprenant:
rebased & squashed, merged into master and 5.x

Issue #6200, opened 2016-11-05
Deprecate old-style plugins in 5.x and remove in 6.0


jordansissel:
Before Logstash had support for installable and upgradable plugins, Logstash provided a way to add additional paths to the ruby library search path in order to provide additional custom plugins.

We've had packaged plugins for [over a year](https://www.elastic.co/blog/logstash-1-5-0-ga-released), and we haven't tested or maintained the old-style plugins pretty much since then.

It's too late for 5.0 to remove this feature, but we can deprecate it in 5.1 and remove it in 6.0.

duylong:
I found these interesting options to integrate our own plugins but also to test quickly.

p-b--:
I am not very against this, I am now somewhat familiar with the plugin loading code and understand the complexity can affect a lot. I would say though that the ability to load plugins directly has made developing really easy - like dead easy.

jordansissel:
@p-b-- I often do development without "installing" a plugin in Logstash; there is simple[1] thing you can do in the Logstash `Gemfile` to tell Logstash to load a plugin from a local directory (which still must have a .gemspec, etc). We could document this development-mode trick. Would that help?

[1] "simple" I mean usually only one step required

p-b--:
Are you referring to this?
https://github.com/logstash-plugins/logstash-input-example#2-running-your-unpublished-plugin-in-logstash

I had "found" this yesterday (I had read it initially but my ruby-fu was much weaker then). I'll start testing using that method.

I haven't done it yet but I wanted to clear the steps:

- Update the plugin `.gempsec` file.
- Update the logstash Gemfile with `gem "logstash-filter-awesome", :path => "/your/local/logstash-filter-awesome"`
- Run `bin/logstash-plugin install --no-verify`
- Run `bin/logstash -e 'filter {awesome {}}'`

Would I need to run a `bundle install` in the plugin dir or would the `logstash-plugin install` handle that?

jordansissel:
#6439 PR is open, work in progress, to execute on this issue.

PR #6439, opened 2016-12-19
Deprecate path.plugins

jordansissel:
I've intended to deprecate old-style plugin (older flag --pluginpath, --path.plugin) for years. Let's do it now :)

Details and discussion are in #6200 

My intent primarily is to deprecate this feature, not remove it entirely, by doing the following:

* Remove references in documentation
* Warn on usage

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -154,8 +154,3 @@
 #
 # log.level: info
 # path.logs:
-#
-# ------------ Other Settings --------------
-#
-# Where to find custom plugins
-# path.plugins: []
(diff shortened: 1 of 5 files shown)


ph:
I am +1 to remove that, the more features we add to plugins registry, like the universal plugins, makes this a bit harder to maintain.

Also, nothing prevent users to either install .gem or to define `:path` in their gemfile.

ph:
*OOPS* I havent read the WIP state on this PR, what do you think its missing @jordansissel ?

jordansissel:
what I think is mainly missing is the docs I want to provide to users to upgrade -- specifically, a path for users to go from the old style to the new style plugins.

suyograo:
Added an issue for the upgrade docs. I will merge this one to 6.0 and 5.x.

PR #6445, opened 2016-12-20
ensure metric collection is disabled when metric.collect is false

jsvd:
in some scenarios, even when metric.collect is false, a pipeline could still produce metrics.

This pr fixes that, also adds and fixes a few tests in the pipeline and pipeline_reporter specs

### DIFF
--- a/logstash-core/spec/logstash/pipeline_reporter_spec.rb
+++ b/logstash-core/spec/logstash/pipeline_reporter_spec.rb
@@ -2,31 +2,7 @@
 require "spec_helper"
 require "logstash/pipeline"
 require "logstash/pipeline_reporter"
-
-class DummyOutput < LogStash::Outputs::Base
-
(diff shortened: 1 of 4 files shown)


logstash-core/spec/logstash/pipeline_reporter_spec.rb
```diff
@@ -2,31 +2,7 @@
...
-
-  def receive(event)
-    @events << event
-  end
-
-  def close
-    @num_closes += 1
-  end
-end
+require_relative "../support/mocks_classes"
```
ph: ++

ph:
LGTM, thanks for the cleanup!

suyograo:
I just tested and there are no global events stats reported (zeros) when no events have been sent.

suyograo:
LGTM

Issue #6444, opened 2016-12-20
Universal plugins don't have a specific version and `print_version_notice` prints a warning for them


ph:
Universal plugins should not trigger this warning.

Maybe we don't event need that warning since all plugins now have a version unless you use the plugin path option in the logstash.yml.

This code is pre Logstash 1.5, I think we can safely remove it.

```
[2016-12-20T20:02:39,440][WARN ][logstash.inputs.metrics  ] metrics plugin doesn't have a version. This plugin isn't well
 supported by the community and likely has no maintainer.
```

PR #6448, opened 2016-12-20
Do not log a warning if a plugin is not from in `#print_notice_version`

ph:
When a plugin is loaded using the `plugins.path` option or is from a
universal plugin there no gemspec can be found for the specific plugin.

We should not print any warning on that case.

Fixes: #6444


ph:
I think it make sense to only disable the case that the plugin version is not found and keep the other one around for now.

suyograo:
LGTM

Issue #6449, opened 2016-12-20
Null events from metrics with no keys


suyograo:
```
"events" : {
              "filtered" : 0,
              "in" : null,
              "duration_in_millis" : null,
              "out" : 0
            },
```

When we query the metrics snapshot, using `::extract_metrics`, if the key does not exist, it gives a null. This may be fine for certain consumers, but when indexing to ES, this ain't great because of the strict mapping. One option: `extract_metrics(key, default_value)` so we can return that if key is not found.

ph:
As a rule of thumb we should populate the metric everytime with zeroes values. 
The metric store is dynamic in nature and unless something is record he is not aware that the values can exist.

suyograo:
@ph ++. Looks like it's just the fields `in` and `duration_in_millis` that have not been init'd for global event metrics

ph:
@suyograo I will make a quick fix for 5.2,  theses stats are collected at the batch level and we will refactor that soon, it will be a good time to move to an observer pattern, since the metric collecting are implemented in two different queues. :(

Issue #5679, opened 2016-07-22
Separate pipeline lifecycle out from pipeline code


andrewvc:
Currently the pipeline/batch class contains metrics/logging statements that can make refactoring and navigating the code tricky. To clean things up we should add a notion of observers to the pipeline for common lifecycle events. We need to first define these events based on current needs, but I'm picturing defining observers with something like:

```ruby
class PipelineObserver
def on_init
end
def on_run
end
def on_batch_deque
end
def on_batch_filter
end
def on_stop
end
# etc. etc.
end
```

ph:
Lets make the code simple and reusable accross other classes, Batch / Agent class could probably benefit from that too.

andrewvc:
@ph zooming out I wonder if we should provide a generic event system for logstash instead of a specific observer.

That might be really useful for future plugins.

ph:
I think you read my mind, I would really like that plugins could communicate between different part of the pipeline or even other plugins.

A few things;

- Maybe shutdown / internal event could use it.
- Maybe we could use a future interface so we know when something got picked up, might be overkill or not.

I've been going through netty a lot and handlers can communicate with each other using that same mechanism.

```ruby
backChannel.fire(:do_something, vars... )
backChannel.fire(:do_something, vars, callback)

```

ph:
This could mean that plugins could define a `triggeredEvent` inside the plugin.

PR #6450, opened 2016-12-20
Initialize with default values global events and pipeline events related metric

ph:
The metric store has no concept is a metric need to exist so as a rule
of thumb we need to defined them with 0 values and send them to the
store when we initialize something.

This PR make sure the batch object is recording the right default values

Fixes: #6449


ph:
This part of the code base need to be refactored into a better strategy see with see the lifecycle #5679 proposition.

suyograo:
LGTM

suyograo:
Tested with x-pack

Issue #6447, opened 2016-12-20
Interval for metrics snapshot collector should be configurable


suyograo:
`Collector` object can register objects to observe and periodically send snapshot updates. This snapshot time task thread is not configurable at the moment. We need to expose this interval and also move the ownership of the snapshotting thread to the registered observer.

PR #6456, opened 2016-12-21
Code cleanup for the collector observer

ph:
We have more the responsability of watching the collector inside the
input itself, this feature might come back when we have a new execution
model that can be improved in watching metrics. But this would require
more granular watchers.

No tests were affected by this changes since the code that required that
features was already removed.

Fixes: #6447

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -130,7 +130,6 @@ class LogStash::Agent
   end
 
   def stop_collecting_metrics
-    @collector.stop
     @periodic_pollers.stop
   end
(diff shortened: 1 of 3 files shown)


suyograo:
@ph I don't see where we are providing a way for consumers to use a different polling interval? I thought there would be a register_listener method or something that takes polling interval and other params.

ph:
@suyograo I thought we were reversing the responsability and make the input responsable of creating the threads?

suyograo:
LGTM

jsvd:
+5 −47 ❤️ 



LGTM

PR #6478, opened 2017-01-03
use date 3.1.1

suyograo:


jsvd:
I suggest renaming this PR/commit to something like "add paquet to Gemfile and pin filter-date to 3.1.1"
otherwise LGTM

Issue #6493, opened 2017-01-06
[Refactor] Interaction  Pipeline / Agent


ph:
When working on #6281 we have discovered that we need to cleanup the pipeline interaction with the agent. The agent known a bit too much about the pipeline implementation which make replacing the pipeline with a java implementation a bit more complicated than it need.

- [x] Direct access to the thread #6497
- [x] reloadable?
- [x] visibility of methods #6497
- [x] refactor creating of the queue with a factory https://github.com/elastic/logstash/pull/6498
- [x] devutils' sample code need to be adjusted. https://github.com/elastic/logstash-devutils/pull/65

Nice to have
- Better defined states machines
- plugin factory for metric
- read client factory metric
- better separation of the metric logic

This cleanup is necessary to allow the LIR execution pipeline to be an easy drop in replacement.

ph:
I will also take a look at https://github.com/elastic/logstash/issues/5679

ph:
done in #6632

PR #6497, opened 2017-01-06
Reduce the visibility of the Pipeline

ph:
With the upcoming changes of the #6281 I want to reduce the visibility of the pipeline as much as possible to the outside world.

- Make sure we don't use any reference to the thread to check if a pipeline is running or not and only rely on the `#running?` method.
- use the private method to mark as much as possible the method that doesn't need to be exposed, this is a start I would also like to hide some of the `attr_reader` methods but this would require more refactoring.
- Set the name of the pipeline thread only once in the agent instead of the `Pipeline#run`.
- Remove the type check on pipeline and instead rely on duck typing.

The tests require https://github.com/elastic/logstash-devutils/pull/65 to make the `#sample` method work.

Ref: #6493

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -293,11 +293,10 @@ class LogStash::Agent
 
   def start_pipeline(id)
     pipeline = @pipelines[id]
-    return unless pipeline.is_a?(LogStash::Pipeline)
     return if pipeline.ready?
     @logger.debug("starting pipeline", :id => id)
     t = Thread.new do
-      LogStash::Util.set_thread_name("pipeline.#{id}")
+      LogStash::Util.set_thread_name("[#{id}]-pipeline-manager")
       begin
         pipeline.run
       rescue => e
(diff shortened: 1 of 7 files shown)


ph:
Just a note, I've deviced to use explicit calls to `private :<method_name>` instead of moving code around to change the visibility, this way I make the PR less green.

ph:
Test is falling because we need a new version of devutils. .. https://github.com/elastic/logstash-devutils/pull/65

jsvd:
I've been trying to run core specs with this branch and pr 65 of devutils, but the specs keep hanging

jsvd:
@ph can you rebase please? also https://github.com/elastic/logstash-devutils/pull/65 is good to merge. I suggest release the new devutils so we can see green here

jsvd:
I'm seeing green in the specs now, I think with a rebase we can merge this

note: this will conflict with https://github.com/elastic/logstash/pull/6631

ph:
I will rebase this PR asap, with all the changes with the pipeline its not easy.

ph:
I think its just easier to do a new PR with all the changes that went into agent/pipeline :(

PR #6499, opened 2017-01-06
Refactor `non_reloadable_plugin`

ph:
Instead of using a list of non reloadable plugin we add a new class
method on the base plugin class that the plugin will override.

By default we assume that all plugins are reloadable, only the stdin
shouldn't be reloadable.


ph:
I will take a look at the failling test, doesn't fail locally, will have to add debug log for travis.

ph:
make sure you use https://github.com/elastic/logstash/pull/6499 :)
this is why the test is hanging here..

jsvd:
LGTM

PR #6504, opened 2017-01-09
Minor code cleanup and reorg of TopologicalSort

andrewvc:
This moves the TopologicalSort algorithm class to the proper
graph.algorithms package. It also cleans up a few little syntax things


ph:
LGTM, make sense make this easier to follow to :)
We will have pretty nifty set of algorithms at the end of this.

Issue #6508, opened 2017-01-09
Modify queue API response


suyograo:
After a discussion on zoom with @jordansissel @colinsurprenant and @talevy, we decided to modify the queue response to make it most actionable to the user. 

- [ ] Introduce a page level API `_node/stats/queue` which provides detailed queue and page level stats to track acks per page
- [x] Modify existing response to only include `queue.events` as a single field that reports current unread events in the queue. Remove `acked_count`, `unacked_count` and `unread_count`
- [ ] Remove `current_size_in_bytes` and replace it with `queue_size_on_disk` which is a percent value calculated from `current_size_in_bytes` (the accurate value, not logical) and `max_queue_size_in_bytes`

colinsurprenant:
This is being discussed in #8986

colinsurprenant:
closing this as it is being followed up in #8986 and #8936.

PR #6510, opened 2017-01-09
remove current_size_in_bytes and acked info from node stats

talevy:
re #6508.

- removed `acked_count`, `unacked_count`, and migrated `unread_count` to
top-level `events` field.
- removed `current_size_in_bytes` info from queue node stats


talevy:
looks like the build failed due to a worker-thread flaky test. re-running

Issue #6506, opened 2017-01-09
Add `hooks` to define a new config source


ph:
We will add a new core hooks to allow universal plugins to register a new source for loading a config.

Currently we support the following sources:
- local
- remote HTTP

We have to do a few refactoring allow the way:

- Split the local and http into 2 separate concept
- Add a base class to define the config source
- investigate the `config_hash` to know when we need to reload a pipeline.
- Keep track of the actual source, file or remote file, this will be useful for the LIR.
- There is also case for the default/input/output.

ph:
@jsvd just a head up this might have some implication on multipipeline, I will try to limit them as much as possible.

ph:
This was done in #6514 we can close this.

PR #6514, opened 2017-01-10
Refactor of the config source loader

ph:
This PR changes a few things in the config handling of Logstash without affecting backward compatibility but add a few level of indirection
for future improvements and add more tests coverage for the existing code.

- First instead of loading all the file directly we use a `SOURCE_LOADER`, this class will use the configured settings of the agent
class to create the source configuration, by default we use a local configuration, but this `SOURCE_LOADER` class allow plugin author to override the behavior and implement their
own custom class.

- The object returned by the loader will accept calls `fetch` and will return an array, this change is necessary for the support of the future multiple pipelines.

- The returned object will be a `PipelineConfig`, this object will return the `contrig_string` of all the different config parts a `settings` properties, but the most important it will keep record of all the original source of configuration.
This will allow the LIR to keep track of that data to give better debug information to the user.

- The PR also add tests for skipping temporary files, reading badly encoded files

- This PR also add a concept of `bootstrap_check` that can be modified by plugin.

Fixes: https://github.com/elastic/logstash/issues/6506


logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -0,0 +1,71 @@
...
+    # This return a ConfigLoader object that will
+    # abstract the call to the different sources and will return multiples pipeline
+    def create(settings)
+      loaders = []
+
+      source_loaders do |config_source|
+        if config_source.match?(settings)
+          loaders << config_source.new(settings)
+        end
+      end
```
ph: This pattern might not be necessary, but keep in mind its for plugin author so we need to add flexibility.

ph:
**Notes**: There is lineage information in the `PipelineConfig` and the `ConfigPart, I did that way to provide better debugging.

ph:
@jsvd Can you do an initial review? I need to sync with you with the 3 faillings tests and the way we pass the config around for the agent/pipelines.

logstash-core/spec/logstash/config/source/local_spec.rb
```diff
@@ -0,0 +1,410 @@
+# encoding: utf-8
```
ph: This files looks scarier than it really is, but I wanted to make sure we really cover all the case we had before.
By looking at our code and the tests, I've found out that we were missing a few of them.

ph:
Just realized we will need a new plugin hook which is triggered after the validation occurs, after the validation this is when we will register the new config sources.

logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      if loaders.empty?
+        # This shouldn't happen with the settings object
+        #  Or with any external plugins.
+        raise "Can't find an appropriate config loader with current settings"
```
jsvd: we should have more error information, maybe create an InvalidLoaderSetting exception or some other name?

logstash-core/lib/logstash/runner.rb
```diff
@@ -241,9 +241,11 @@ def execute
     end
 
     if setting("config.test_and_exit")
-      config_loader = LogStash::Config::Loader.new(logger)
-      config_str = config_loader.format_config(setting("path.config"), setting("config.string"))
       begin
+        config_loader = LogStash::Config::SOURCE_LOADER.create(@settings)
+
+        # TODO(ph): multiple pipeline will affect this
+        config_str = config_loader.pipeline_config.config_string
```
jsvd: as mentioned above, we could change this loading strategy to something like 
```ruby
source_loader = LogStash::Config::SOURCE_LOADER.create(@settings)
config = source_loader.fetch.first
LogStash::Pipeline.new(config)
```

logstash-core/lib/logstash/agent.rb
```diff
@@ -260,16 +261,17 @@ def create_pipeline(settings, config=nil)
...
+    @config_loader.pipeline_config
   end
 
   # since this method modifies the @pipelines hash it is
   # wrapped in @upgrade_mutex in the parent call `reload_state!`
   def reload_pipeline!(id)
     old_pipeline = @pipelines[id]
-    new_config = fetch_config(old_pipeline.settings)
-    if old_pipeline.config_str == new_config
+    new_config = fetch_config
```
ph: Yes that's  the idea i've discussed about diffing the current running pipeline vs the new available configs, but I wanted to limit the changes in the `reload_pipeline!`  to not impact too much your work on the multiple pipeline so I have opted to use a similar interface.

logstash-core/lib/logstash/config/source_loader.rb
ph: make sense! will make the change

logstash-core/lib/logstash/agent.rb
ph: just a note, the PipelineConfig object contains the following: config_string, config_parts and the pipeline_id. so we can just pass this object around.

logstash-core/spec/logstash/config/source_loader_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    loaders = []
+
+    subject.source_loaders { |loader| loaders << loader }
+
+    expect(loaders.size).to eq(1)
+    expect(loaders).to include(LogStash::Config::Source::Local)
+  end
+
+  it "allows to override the available source loaders" do
+    subject.configure_sources(DummyLoader)
```
jsvd: This should probably be a DummySource

logstash-core/lib/logstash/agent.rb
ph: Just to note, this si why I've added this comment concerning the execution https://github.com/elastic/logstash/issues/6521#issuecomment-272959448

logstash-core/lib/logstash/agent.rb
```diff
@@ -235,15 +236,15 @@ def collect_metrics?
...
+        config = fetch_config
       rescue => e
         @logger.error("failed to fetch pipeline configuration", :message => e.message)
         return
       end
     end
 
     begin
-      LogStash::Pipeline.new(config, settings, metric)
+      LogStash::Pipeline.new(config.config_string, config.settings, metric)
```
ph: The problem is, If I change the signature of the Pipeline this will make plugins test to fails.

logstash-core/lib/logstash/runner.rb
ph: @jsvd and myself discussed on zoom.

We cannot easily change the signature of the Pipeline because some plugins tests use it directly.
But we can create another class maybe `RubyPipeline` for lack of a better name that use the new signature and the old pipeline will change the internal.

ph:
@jsvd Ready for a review, concerning the change in the Pipeline's signature I would prefer to do it in another PR.  I've created an issue for that https://github.com/elastic/logstash/issues/6601 so we can discuss it.

ph:
This PR seems to broke something in the monitoring pipeline, taking a look.

ph:
```
1) Test Logstash instance should still merge when -e is specified and -f has no valid config files
     Failure/Error: expect(is_port_open?(port1)).to be true
       expected true
            got false
     # ./specs/01_logstash_bin_smoke_spec.rb:69:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/logstash-devutils-1.2.1-java/lib/logstash/devutils/rspec/logstash_helpers.rb:13:in `try'
     # ./specs/01_logstash_bin_smoke_spec.rb:68:in `(root)'
```

This is a legit failure, will check how/if we need to fix that.

ph:
@suyograo @jsvd 

The test failling is the following integration test.
```
  it "should still merge when -e is specified and -f has no valid config files" do
    config_string = "input { tcp { port => #{port1} } }"
    @ls1.spawn_logstash("-e", config_string, "-f" "/tmp/foobartest")
    @ls1.wait_for_logstash

    try(20) do
      expect(is_port_open?(port1)).to be true
    end
  end
```
I know we have a smoke test for this specific case, but I am not sure we should support that scenario. The less surprising scenario would be to display an error to say we can load a config in the specific path because the file cannot be found.

jsvd:
@ph that one is tricky...I guess a user can have a `bin/logstash -e " .... " -f "/etc/logstash.d/*.cfg"`
so if it's a wildcard, it should be "understandable" if there are no files.

That said, seems to be a very corner case..definitely should go away in 6.0, not sure how to deal with it until then.

ph:
@jsvd yeah its tricky, thing is before that PR I dont think we had a clear definition of all the case of config loading :(

logstash-core/lib/logstash/agent.rb
```diff
@@ -260,16 +264,17 @@ def create_pipeline(settings, config=nil)
     end
   end
 
-  def fetch_config(settings)
-    @config_loader.format_config(settings.get("path.config"), settings.get("config.string"))
+  def fetch_config
+    # TODO(ph) multiple pipeline, we only support one config for now
+    @source_loader.fetch.first
```
jsvd: when other pipelines are loaded (like the monitoring pipeline), we aren't sure which one is loaded first.

This single pipeline strategy should be avoided and become something like:

```ruby
def reload_state!
   pipeline_configs = @source_loader.fetch
   pipelines_to_create, pipelines_to_delete, pipelines_to_reload = resolve_state_actions(pipeline_configs)
   pipelines_to_create.each {|p| create_pipeline(p) }
   pipelines_to_delete.each {|p| delete_pipepline(p) }
   pipelines_to_reload.each {|p| reload_pipeline(p) }
end
```

logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -0,0 +1,68 @@
+# encoding: utf-8
+require "logstash/config/source/local"
+require "logstash/errors"
+require "thread"
+require "set"
+
+module LogStash module Config
+  class SourceLoader
+    class Loader
+      def initialize(loaders)
```
jsvd: I think naming isn't very consistent in this namespace, I'll point out some of these observations below

logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -0,0 +1,68 @@
...
+      def fetch
+        @loaders.collect(&:pipeline_configs).flatten
+      end
+    end
+
+    include LogStash::Util::Loggable
+
+    def initialize
+      @source_loaders_mutex = Mutex.new
+      @source_loaders = Set.new([LogStash::Config::Source::Local])
```
jsvd: the current class is called SourceLoader, but this variable is named `source_loaders` and assign to it a set of Source classes

logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    # This return a ConfigLoader object that will
+    # abstract the call to the different sources and will return multiples pipeline
+    def create(settings)
+      loaders = []
+
+      source_loaders do |config_source|
```
jsvd: it seems this source_loaders synchronized iterator should yield source_loader, but yields `config_source` instead

logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -0,0 +1,68 @@
...
+        if config_source.match?(settings)
+          loaders << config_source.new(settings)
```
jsvd: here we are adding config_source objects to a list called loaders

logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -0,0 +1,68 @@
...
+        # but lets add a guard so we fail fast.
+        raise LogStash::InvalidSourceLoaderSettingError, "Can't find an appropriate config loader with current settings"
+      else
+        Loader.new(loaders)
+      end
+    end
+
+    def source_loaders
+      @source_loaders_mutex.synchronize do
+        @source_loaders.each do |config|
```
jsvd: this should yeld source_loader instead of config?

logstash-core/lib/logstash/agent.rb
ph: Just realized, we really need that because lets say we delete the remote config.. Logstash need to stop the corresponding pipeline and wait.

ph:
@jsvd This is starting to be a bit more complex, I wonder If I can make it 2 PR for that. :(

logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -0,0 +1,68 @@
...
+        end
+      end
+
+      if loaders.empty?
+
+        # This shouldn't happen with the settings object or with any external plugins.
+        # but lets add a guard so we fail fast.
+        raise LogStash::InvalidSourceLoaderSettingError, "Can't find an appropriate config loader with current settings"
+      else
+        Loader.new(loaders)
```
jsvd: In order to dynamically add and remove sources to the source loader, we could remove this class and use instances of SourceLoader directly

ph:
Another thing @jsvd and myself discussed, lets say that that the source of configuration is dynamic and the following scenario happen.


- first call reach correctly the source of configuration
- second call the source of configuration is down.

We have 2 different strategy we could use:

1. One the source block until it succeed and manage his own retry loop.
2. The source return a `SuccessReponse` or  `FailResponse` and the retry is managed by the auto-reload.


We were biased to **option 2** because this won't block logstash.

PR #6518, opened 2017-01-10
add queue_size_on_disk stat to persisted queue node stats

talevy:
This uses the Queue's offset state management as a metric for measuring
bytes usage of the queue on the physical disk.

re #6508.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -126,6 +126,10 @@ public class Queue implements Closeable {
         return this.currentByteSize;
     }
 
+    public long getCurrentPhysicallyPersistedByteSize() {
+        return headPage.getPageIO().getHead() + tailPages.stream().mapToLong((p) -> p.getPageIO().getHead()).sum();
+    }
+
     public int getPageCapacity() {
         return this.pageCapacity;
     }
--- a/logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueExtLibrary.java
+++ b/logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueExtLibrary.java
@@ -112,6 +112,11 @@ public class JrubyAckedQueueExtLibrary implements Library {
             return context.runtime.newFixnum(queue.getCurrentByteSize());
         }
 
+        @JRubyMethod(name = "current_persisted_bytes")
+        public IRubyObject ruby_current_persisted_bytes(ThreadContext context) {
+            return context.runtime.newFixnum(queue.getCurrentPhysicallyPersistedByteSize());
+        }
+
         @JRubyMethod(name = "acked_count")
         public IRubyObject ruby_acked_count(ThreadContext context) {
             return context.runtime.newFixnum(queue.getAckedCount());
--- a/logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueMemoryExtLibrary.java
+++ b/logstash-core-queue-jruby/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueMemoryExtLibrary.java
@@ -106,6 +106,11 @@ public class JrubyAckedQueueMemoryExtLibrary implements Library {
             return context.runtime.newFixnum(queue.getCurrentByteSize());
         }
 
+        @JRubyMethod(name = "current_persisted_bytes")
+        public IRubyObject ruby_current_persisted_bytes(ThreadContext context) {
+            return context.runtime.newFixnum(queue.getCurrentPhysicallyPersistedByteSize());
+        }
+
         @JRubyMethod(name = "acked_count")
         public IRubyObject ruby_acked_count(ThreadContext context) {
             return context.runtime.newFixnum(queue.getAckedCount());
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -539,6 +539,7 @@ module LogStash; class Pipeline
         n.gauge(:page_capacity_in_bytes, queue.page_capacity)
         n.gauge(:max_queue_size_in_bytes, queue.max_size_in_bytes)
         n.gauge(:max_unread_events, queue.max_unread_events)
+        n.gauge(:queue_size_on_disk, queue.current_persisted_bytes)
       end
       pipeline_metric.namespace([:data]).tap do |n|
         n.gauge(:free_space_in_bytes, file_store.get_unallocated_space)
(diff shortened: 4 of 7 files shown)


talevy:
I'm not in love with any of the naming and inconsistencies around "queue_size_on_disk", "head", "writePosition", etc. Open to suggestions. might switch it all to `writePosition` exception for when it is exposed via metrics

colinsurprenant:
I see some  Gradle changes in this PR  - is that relevant to the feature?

talevy:
I found those changes required for me to have a sane workflow. I can remove those from here if you want

colinsurprenant:
@talevy I'd prefer yes because your "sane" workflow may break my workflow and I don't think it relates to this PR and nothing prevets you to keep your workflow. I suggest you extract these changes into another PR and then we can discuss about this specifically or defer all this to the upcoming src tree reorg?

talevy:
@colinsurprenant 

here is a sample pipeline

```
input { generator { count => 10000 } }
filter { sleep { time => 10000 } }
```

and a modified output snippet from the stats api that includes both the persisted and `Queue.currentByteSize`

```
      "capacity" : {
        "queue_current_byte_size" : 262144000,
        "queue_size_on_disk" : 1430001,
      },
```

you'll see that `currentByteSize` basically equals `page_capacity_in_bytes*number_of_pages`, while `queue_size_on_disk` is actually the bytes written.

colinsurprenant:
@talevy ah yes and this leads to the reported file size vs the actual disk usage duality. Pages are pre-allocated in a sparse way: when a page is created but unwritten to it is reported as having a file size equals to the `queue.page_capacity` by `ls -l` but on Linux `du` will report is as 0 bytes size, whereas on MacOS it is be reported as `queue.page_capacity` by both `ls` and `du`. 

Example:

1- Linux (ext4)
```
$ truncate -s 10M foo
$ cat foo > bar
$ ls -l foo bar
-rw-rw-r-- 1 colin colin 10485760 Jan 31 19:17 bar
-rw-rw-r-- 1 colin colin 10485760 Jan 31 19:17 foo
$ du foo bar
0	foo
10240	bar
```

2- MacOS
```
$ truncate -s 10M foo: cat foo > bar
$ ls -l foo bar
-rw-r--r--+ 1 colin staff 10485760 Jan 31 13:14 bar
-rw-r--r--+ 1 colin staff 10485760 Jan 31 13:14 foo
$ du  foo bar
10240	foo
10240	bar
```

Also note that the actual disk usage of the queue will be larger than the sum of the write positions of each page because this does not account for the sector size. 

I am not sure what we want to actually measure and report here and we definitely need to have a consistent understanding and strategy with regard to `Queue.currentByteSize` and config option `queue.max_bytes`. 

Thoughts on this @jsvd (wrote the `Queue.currentByteSize` + `queue.max_bytes` support) and @jordansissel ?

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -126,6 +126,10 @@ public long getCurrentByteSize() {
         return this.currentByteSize;
     }
 
+    public long getCurrentPhysicallyPersistedByteSize() {
+        return headPage.getPageIO().getHead() + tailPages.stream().mapToLong((p) -> p.getPageIO().getHead()).sum();
```
jordansissel: This strategy seems good. Sum of the head position (byte value offset) of all pages.
👍

jordansissel:
> I am not sure what we want to actually measure and report here and we definitely need to have a consistent understanding and strategy with regard to Queue.currentByteSize and config option queue.max_bytes

I agree it's a bit tricky.  Allow me a small tangent that questions if we even need "byte size used" as a metric.

There are many values we could present, but I'll try to focus on actionable ones. I also want to specifically avoid instantaneous values that alone may not be actionable.

For a persisted queue, I'm not even sure queue used-size (fullness, etc) is an actionable property, so I'd like to consider, for now, the possibility of not exposing it as a metric. 

Are we thinking these metrics will be used for a proxy for other health? 

I wonder if we only need behavioral metrics: The time spent waiting for queue writes to complete?

End tangent :)

---- 

I not sure which of the values is best to show to users, and I don't know if there is any perfectly accurate value that will align with user expectations..

If we trust that the queue will never exceed max_bytes (which includes any acked events living in unpurged pages), then maybe we do not need to share the "bytes used" metric. If we guarantee that the 'stuff in the queue' is always in the range `[0, max_bytes]`, then I propose we only publish a percentage.

For the computation of percentage, I propose we use "logical bytes used" / max_bytes, where "logical bytes used" is a count of bytes that the Queue has knowingly written. This means that any space allocated-but-not-yet-written will not be counted, even though it would show up in `du` as @colinsurprenant [demonstrated](https://github.com/elastic/logstash/pull/6518#issuecomment-276451360). 

The consequence of my proposal is that the queue could be "full" (blocking) but will show 99% because the following could be true: `queue_size + next_Event_size > max_size && queue_size < max_size` (queue has space, but not enough for the next event).

For users, actionable observations might be:
* Queue growing steadily. Action: investigate why producers are outpacing consumers.
* Queue always near full (at 99%). Action: Same as above.
* Queue empty, but output counts are not increasing: Action: Investigate inputs to see why no data is arriving.

If we encourage users to size their queues appropriately (small, but large enough to absorb bursts from lossy transports like tcp, udp, and syslog), then we may be able to combine the two observations "growing steadily" and "always near full" into a single observation "producers are outpacing consumers"

Using a percentage allows us to choose the right metric to indicate queue-fullness without confusing users with incorrect (or perceived incorrect) computation about actual byte size on disk.

Summarizing above, I think I am proposing:

* A percentage-used metric (size of all events / max_size)
* A total disk usage (using `du`)

colinsurprenant:
- Queue usage  percentage works only when a queue limit is set but we do not make it mandatory as of now. We can limit the queue by number of events `queue.max_events` or by byte size `queue.max_bytes` but both settings can be set to zero for *unlimited*.
- I think it is fine for a user to setup an *unlimited* queue and eventually rely on a **percentage disk used** limit to avoid a disk full and start applying back pressure. I think this is the most flexible configuration as it allow the user to say "use whatever disk space there is available now and I'll free up space or grow my partition if I need more space" and logstash will adapt. 
- I think the actionable observations can be measured or shown using either this proposed `queue_size_on_disk` or even the number of *unacked* or *unread* events. You can clearly see any queue growth or decrease from these.

colinsurprenant:
I just realized this is actually a respin of the discussion we already had in #6612 with @whack and @jsvd. 

This I believe were the action items and/or conclusions:
- We currently provide 2 ways of computing the the queue size,  a) originally proposed by @jsvd by essentially using the page size multiplied by the number of pages and b) this proposal  where he uses the head (write position) of each page to know the exact number of bytes written per page. **we should only use one way of computing the the queue size**. 
- We should create an issue to add support for % disk space-left threshold

colinsurprenant:
closing, this has been rebooted in #7010

PR #6519, opened 2017-01-10
feature/optimistic recovery

colinsurprenant:
Optimistic recovery feature.

With this feature, on startup, the head page data file is opened using the new `PageIO.recover()` method which sequentially reads the data file from the start and try to find valid written elements regardless of the head checkpoint. This is possible thanks to the per-element checksums which allow self-integrity checking in the data file. 

### TODO

- [x] see about capacity as an it versus file size as long (see TODO in PageIO)
- [x] DRY code between open/recover methods
- [x] DRY code between MmapPageIO and ByteBufferPageIO to ease testing
- [x] add more unit tests
- [x] remove debug traces

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -1,5 +1,7 @@
 package org.logstash.ackedqueue;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.logstash.common.io.CheckpointIO;
 import org.logstash.common.io.PageIO;
 import org.logstash.common.io.PageIOFactory;
@@ -60,22 +62,24 @@ public class Queue implements Closeable {
     private final Method deserializeMethod;
 
     // thread safety
-    final Lock lock = new ReentrantLock();
-    final Condition notFull  = lock.newCondition();
-    final Condition notEmpty = lock.newCondition();
+    private final Lock lock = new ReentrantLock();
+    private final Condition notFull  = lock.newCondition();
+    private final Condition notEmpty = lock.newCondition();
+
+    private static final Logger logger = LogManager.getLogger(Queue.class);
 
     public Queue(Settings settings) {
         this(
-                settings.getDirPath(),
-                settings.getCapacity(),
-                settings.getQueueMaxBytes(),
-                settings.getCheckpointIOFactory().build(settings.getDirPath()),
-                settings.getPageIOFactory(),
-                settings.getElementClass(),
-                settings.getMaxUnread(),
-                settings.getCheckpointMaxWrites(),
-                settings.getCheckpointMaxAcks(),
-                settings.getCheckpointMaxInterval()
+            settings.getDirPath(),
+            settings.getCapacity(),
+            settings.getQueueMaxBytes(),
+            settings.getCheckpointIOFactory().build(settings.getDirPath()),
+            settings.getPageIOFactory(),
+            settings.getElementClass(),
+            settings.getMaxUnread(),
+            settings.getCheckpointMaxWrites(),
+            settings.getCheckpointMaxAcks(),
+            settings.getCheckpointMaxInterval()
         );
     }
 
@@ -142,11 +146,8 @@ public class Queue implements Closeable {
         try {
             headCheckpoint = this.checkpointIO.read(checkpointIO.headFileName());
         } catch (NoSuchFileException e) {
-            headCheckpoint = null;
-        }
+            // if there is no head checkpoint, create a new headpage and checkpoint it and exit method
 
-        // if there is no head checkpoint, create a new headpage and checkpoint it and exit method
-        if (headCheckpoint == null) {
             this.seqNum = 0;
             headPageNum = 0;
 
@@ -162,33 +163,45 @@ public class Queue implements Closeable {
         for (int pageNum = headCheckpoint.getFirstUnackedPageNum(); pageNum < headCheckpoint.getPageNum(); pageNum++) {
 
             // all tail checkpoints in the sequence should exist, if not abort mission with a NoSuchFileException
-            Checkpoint tailCheckpoint = this.checkpointIO.read(this.checkpointIO.tailFileName(pageNum));
+            Checkpoint cp = this.checkpointIO.read(this.checkpointIO.tailFileName(pageNum));
 
             PageIO pageIO = this.pageIOFactory.build(pageNum, this.pageCapacity, this.dirPath);
+            pageIO.open(cp.getMinSeqNum(), cp.getElementCount());
 
-            add(tailCheckpoint, pageIO);
+            add(cp, new TailPage(cp, this, pageIO));
         }
 
         // transform the head page into a tail page only if the headpage is non-empty
         // in both cases it will be checkpointed to track any changes in the firstUnackedPageNum when reconstructing the tail pages
 
-        if (headCheckpoint.getMinSeqNum() <= 0 && headCheckpoint.getElementCount() <= 0) {
+        PageIO pageIO = this.pageIOFactory.build(headCheckpoint.getPageNum(), this.pageCapacity, this.dirPath);
+        pageIO.recover(); // optimistically recovers the head page data file and set minSeqNum and elementCount to the actual read/recovered data
+
+        if (pageIO.getMinSeqNum() != headCheckpoint.getMinSeqNum() || pageIO.getElementCount() != headCheckpoint.getElementCount()) {
+            // the recovered page IO shows different minSeqNum or elementCount than the checkpoint, use the page IO attributes
+
+            logger.warn("recovered head data page {} is different than checkpoint, using recovered page information", headCheckpoint.getPageNum());
+            logger.debug("head checkpoint minSeqNum={} or elementCount={} is different than head pageIO minSeqNum={} or elementCount={}", headCheckpoint.getMinSeqNum(), headCheckpoint.getElementCount(), pageIO.getMinSeqNum(), pageIO.getElementCount());
+
+            long firstUnackedSeqNum = headCheckpoint.getFirstUnackedSeqNum();
+            if (firstUnackedSeqNum < pageIO.getMinSeqNum()) {
+                logger.debug("head checkpoint firstUnackedSeqNum={} is < head pageIO minSeqNum={}, using pageIO minSeqNum", firstUnackedSeqNum, pageIO.getMinSeqNum());
+                firstUnackedSeqNum = pageIO.getMinSeqNum();
+            }
+            headCheckpoint = new Checkpoint(headCheckpoint.getPageNum(), headCheckpoint.getFirstUnackedPageNum(), firstUnackedSeqNum, pageIO.getMinSeqNum(), pageIO.getElementCount());
+        }
+        this.headPage = new HeadPage(headCheckpoint, this, pageIO);
+
+        if (this.headPage.getMinSeqNum() <= 0 && this.headPage.getElementCount() <= 0) {
             // head page is empty, let's keep it as-is
 
-            PageIO headPageIO = this.pageIOFactory.build(headCheckpoint.getPageNum(), this.pageCapacity, this.dirPath);
-            this.headPage = new HeadPage(headCheckpoint, this, headPageIO);
-            this.currentByteSize += headPageIO.getCapacity();
+            this.currentByteSize += pageIO.getCapacity();
 
             // but checkpoint it to update the firstUnackedPageNum if it changed
             this.headPage.checkpoint();
         } else {
             // head page is non-empty, transform it into a tail page and create a new empty head page
-
-            PageIO pageIO = this.pageIOFactory.build(headCheckpoint.getPageNum(), this.pageCapacity, this.dirPath);
-
-            TailPage p = new TailPage(headCheckpoint, this, pageIO);
-            p.checkpoint();
-            add(headCheckpoint, pageIO);
+            add(headCheckpoint, this.headPage.behead());
 
             headPageNum = headCheckpoint.getPageNum() + 1;
             newCheckpointedHeadpage(headPageNum);
@@ -207,10 +220,12 @@ public class Queue implements Closeable {
         this.closed.set(false);
     }
 
-    private void add(Checkpoint checkpoint, PageIO pageIO) throws IOException {
+    // add a read tail page into this queue structures but also verify that this tail page
+    // is not fully acked in which case it will be purged
+    private void add(Checkpoint checkpoint, TailPage page) throws IOException {
         if (checkpoint.isFullyAcked()) {
             // first make sure any fully acked page per the checkpoint is purged if not already
-            try { pageIO.purge(); } catch (NoSuchFileException e) { /* ignore */ }
+            try { page.getPageIO().purge(); } catch (NoSuchFileException e) { /* ignore */ }
 
             // we want to keep all the "middle" checkpoints between the first unacked tail page and the head page
             // to always have a contiguous sequence of checkpoints which helps figuring queue integrity. for this
@@ -223,30 +238,30 @@ public class Queue implements Closeable {
             } else {
                 // create a tail page with a null PageIO and add it to tail pages but not unreadTailPages
                 // since it is fully read because also fully acked
+                // TODO: I don't like this null pageIO tail page...
                 this.tailPages.add(new TailPage(checkpoint, this, null));
             }
         } else {
-            TailPage p = new TailPage(checkpoint, this, pageIO);
-            this.tailPages.add(p);
-            this.unreadTailPages.add(p);
-            this.unreadCount += p.unreadCount();
-            this.currentByteSize += pageIO.getCapacity();
+            this.tailPages.add(page);
+            this.unreadTailPages.add(page);
+            this.unreadCount += page.unreadCount();
+            this.currentByteSize += page.getPageIO().getCapacity();
 
             // for now deactivate all tail pages, we will only reactivate the first one at the end
-            pageIO.deactivate();
+            page.getPageIO().deactivate();
         }
 
         // track the seqNum as we rebuild tail pages, prevent empty pages with a minSeqNum of 0 to reset seqNum
         if (checkpoint.maxSeqNum() > this.seqNum) {
             this.seqNum = checkpoint.maxSeqNum();
         }
-
     }
 
     // create a new empty headpage for the given pageNum and imidiately checkpoint it
     // @param pageNum the page number of the new head page
     private void newCheckpointedHeadpage(int pageNum) throws IOException {
         PageIO headPageIO = this.pageIOFactory.build(pageNum, this.pageCapacity, this.dirPath);
+        headPageIO.create();
         this.headPage = new HeadPage(pageNum, this, headPageIO);
         this.headPage.forceCheckpoint();
         this.currentByteSize += headPageIO.getCapacity();
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/TailPage.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/TailPage.java
@@ -14,6 +14,7 @@ public class TailPage extends Page {
     }
 
     // create a new TailPage object for an exiting Checkpoint and data file
+    // @param pageIO the PageIO object is expected to be open/recover/create
     public TailPage(Checkpoint checkpoint, Queue queue, PageIO pageIO) throws IOException {
         super(checkpoint.getPageNum(), queue, checkpoint.getMinSeqNum(), checkpoint.getElementCount(), checkpoint.getFirstUnackedSeqNum(), new BitSet(), pageIO);
 
@@ -21,12 +22,6 @@ public class TailPage extends Page {
         if (checkpoint.getFirstUnackedSeqNum() > checkpoint.getMinSeqNum()) {
             this.ackedSeqNums.flip(0, (int) (checkpoint.getFirstUnackedSeqNum() - checkpoint.getMinSeqNum()));
         }
-
-        if (pageIO != null) {
-            // open the data file and reconstruct the IO object internal state
-            pageIO.open(checkpoint.getMinSeqNum(), checkpoint.getElementCount());
-        }
-
     }
 
     public void checkpoint() throws IOException {
--- a/logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
@@ -1,0 +1,277 @@
+package org.logstash.common.io;
+
+import org.logstash.ackedqueue.SequencedList;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.zip.CRC32;
+import java.util.zip.Checksum;
+
+public abstract class AbstractByteBufferPageIO implements PageIO {
+
+    public class PageIOInvalidElementException extends IOException {
+        public PageIOInvalidElementException() { super(); }
+        public PageIOInvalidElementException(String message) { super(message); }
+    }
+
+    public class PageIOInvalidVersionException extends IOException {
+        public PageIOInvalidVersionException(String message) { super(message); }
+    }
+
+    public static final byte VERSION_ONE = 1;
+    public static final int VERSION_SIZE = Byte.BYTES;
+    public static final int CHECKSUM_SIZE = Integer.BYTES;
+    public static final int LENGTH_SIZE = Integer.BYTES;
+    public static final int SEQNUM_SIZE = Long.BYTES;
+    public static final int MIN_RECORD_SIZE = SEQNUM_SIZE + CHECKSUM_SIZE;
+    public static final int HEADER_SIZE = 1;     // version byte
+    public static final int MIN_CAPACITY = VERSION_SIZE + SEQNUM_SIZE + LENGTH_SIZE + 1 + CHECKSUM_SIZE; // header overhead plus elements overhead to hold a single 1 byte element
+    public static final List<byte[]> EMPTY_READ = new ArrayList<>(0);
+
+    public static final boolean VERIFY_CHECKSUM = true;
+    public static final boolean STRICT_CAPACITY = true;
+
+    protected int capacity; // page capacity is an int per the ByteBuffer class.
+    protected final int pageNum;
+    protected final List<Integer> offsetMap; // has to be extendable
+    protected long minSeqNum; // TODO: to make minSeqNum final we have to pass in the minSeqNum in the constructor and not set it on first write
+    protected int elementCount;
+    protected int head; // head is the write position and is an int per ByteBuffer class position
+    protected byte version;
+    protected Checksum checkSummer;
+
+    public AbstractByteBufferPageIO(int pageNum, int capacity) {
+        this.minSeqNum = 0;
+        this.elementCount = 0;
+        this.version = 0;
+        this.head = 0;
+        this.pageNum = pageNum;
+        this.capacity = capacity;
+        this.offsetMap = new ArrayList<>();
+        this.checkSummer = new CRC32();
+    }
+
+    // @return the concrete class buffer
+    protected abstract ByteBuffer getBuffer();
+
+    @Override
+    public void open(long minSeqNum, int elementCount) throws IOException {
+        getBuffer().position(0);
+        this.version = getBuffer().get();
+        validateVersion(this.version);
+        this.head = 1;
+
+        this.minSeqNum = minSeqNum;
+        this.elementCount = elementCount;
+
+        if (this.elementCount > 0) {
+            // verify first seqNum to be same as expected minSeqNum
+            long seqNum = getBuffer().getLong();
+            if (seqNum != this.minSeqNum) { throw new IOException(String.format("first seqNum=%d is different than minSeqNum=%d", seqNum, this.minSeqNum)); }
+
+            // reset back position to first seqNum
+            getBuffer().position(this.head);
+
+            for (int i = 0; i < this.elementCount; i++) {
+                // verify that seqNum must be of strict + 1 increasing order
+                readNextElement(this.minSeqNum + i, !VERIFY_CHECKSUM);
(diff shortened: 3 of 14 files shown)


colinsurprenant:
It is a bit out of scope of this PR but I decided to extract/DRY code into a new abstract class `AbstractByteBufferPageIO` - since most code between `ByteBufferPageIO` and `MmapPageIO` was identical and in order to add tests for the `recover()` method I would have had to duplicate even more code so I decided to fix that now.

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
```diff
@@ -0,0 +1,262 @@
...
+        if (this.elementCount <= 0) {
+            this.minSeqNum = 0;
+        }
+    }
+
+    // read and vadidate next element at page head
+    // @param verifyChecksum if true the actual element data will be read + checksumed and compared to written checksum
+    private void readNextElement(long expectedSeqNum, boolean verifyChecksum) throws PageIOInvalidElementException {
+        // if there is no room for the seqNum and length bytes stop here
+        if (this.head + SEQNUM_SIZE + LENGTH_SIZE > capacity) { throw new PageIOInvalidElementException("cannot read seqNum and length bytes past buffer capacity"); }
```
jordansissel: "past buffer capacity" sounds strange to me. Maybe it's ok as worded, but it feels more like "We want to read the next element, but there are fewer than X bytes available before the end of the page, so this is probably a partial write that occurred when Logstash was last terminated abnormally" ?

jordansissel: Or something like: "While recovering, we reached the end of the page while trying to read the next entry. This means ... something something partial write ..."

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
```diff
@@ -0,0 +1,262 @@
...
+
+        if (seqNum != expectedSeqNum) { throw new PageIOInvalidElementException(String.format("Element seqNum %d is expected to be %d", seqNum, expectedSeqNum)); }
+
+        int length = buffer.getInt();
+        this.head += LENGTH_SIZE;
+
+        // length must be > 0
+        if (length <= 0) { throw new PageIOInvalidElementException("Element invalid length"); }
+
+        // if there is no room for the proposed data lenght and checksum just stop here
```
jordansissel: typo "length"

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
```diff
@@ -0,0 +1,262 @@
...
+        long seqNum = buffer.getLong();
+        this.head += SEQNUM_SIZE;
+
+        if (seqNum != expectedSeqNum) { throw new PageIOInvalidElementException(String.format("Element seqNum %d is expected to be %d", seqNum, expectedSeqNum)); }
+
+        int length = buffer.getInt();
+        this.head += LENGTH_SIZE;
+
+        // length must be > 0
+        if (length <= 0) { throw new PageIOInvalidElementException("Element invalid length"); }
```
jordansissel: Do you want to check for too-large length also? (larger than page size?)

logstash-core/src/test/java/org/logstash/common/io/MemoryCheckpointTest.java
```diff
@@ -39,7 +41,11 @@ public void writeNewReadExisting() throws IOException {
 
     @Test
     public void readInnexisting() throws IOException {
-        Checkpoint checkpoint = io.read("checkpoint.invalid");
-        assertThat(checkpoint, is(equalTo(null)));
+        try {
+            io.read("checkpoint.invalid");
+            fail("expected NoSuchFileException");
+        } catch (NoSuchFileException e) {
```
jordansissel: Testing suggestion:

* Change annotation to: `@Test(exception = NoSuchFileException)` (or whatever, I'm writing by hand here, not tested)
* Change function to be 1 line: `io.read("checkpoint.invalid")`

logstash-core/src/test/java/org/logstash/common/io/ByteBufferPageIOTest.java
```diff
@@ -63,64 +64,217 @@ public void hasSpaceAfterWrite() throws IOException {
...
+            ByteBufferPageIO io = newEmptyPageIO();
+            byte[] bytes = element.serialize();
+            int checksum = io.checksum(bytes);
+            bytes[1] = 0x01;
+            io.write(bytes, 1L, bytes.length, checksum);
+            return io.dump();
+        }
+    );
+
+    private final List<BufferGenerator> doubleGenerators = Arrays.asList(
```
jordansissel: This would probably be better written using JUnit parameterized tests for the generated dumps.

```java
@RunWith(Parameterized.class)
public class MyTestClass {
    @Parameters
    public static Collection<byte[]> data() {
       ... generate all `io.dump()` calls here and return a Collection<> of them.
    }

    @Parameter
    public byte[] bytes;

    @Test
    public void openInvalidElement() {
         ByteBufferPageIO io = newPageIO(bytes.length, bytes);
         io.open(1L, 1);

         assertThat(io.getElementCount(), is(equalTo(1)));
         assertThat(io.getMinSeqNum(), is(equalTo(1L))); 
    }
}
```

This has nice properties of avoiding the iteration you do on singleGenerators and doubleGenerators. You could also use this to simplify the invalid write generation.

logstash-core/src/test/java/org/logstash/ackedqueue/TestSettings.java
```diff
@@ -9,7 +9,7 @@
 
 public class TestSettings {
 
-    public static Settings getSettings(int capacity) {
+    public static Settings volatileQueueSettings(int capacity) {
```
jordansissel: Why the rename?

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
```diff
@@ -0,0 +1,262 @@
...
+            }
+        }
+    }
+
+    // recover will overwrite/update/set this object minSeqNum, capacity and elementCount attributes
+    // to reflect what it recovered from the page
+    @Override
+    public void recover() throws IOException {
+        getBuffer().position(0);
+        this.version = getBuffer().get();
```
jordansissel: The `version` value should be validated here. Is it a known/supported version?

logstash-core/src/test/java/org/logstash/common/io/MemoryCheckpointTest.java
colinsurprenant: uh ... weird ... not sure why I did not use that in the first place really 🤔

logstash-core/src/test/java/org/logstash/common/io/ByteBufferPageIOTest.java
colinsurprenant: TIL @Parameter ... will look.

logstash-core/src/test/java/org/logstash/ackedqueue/TestSettings.java
colinsurprenant: because the `TestSettings` class had `getSettings()` and `getSettingsCheckpointFilePageMemory()` which has 2 problems:
1- since `TestSettings` now exposes both volatile and persisted settings, `getSettings()` means nothing.
2- I have not idea what `getSettingsCheckpointFilePageMemory()` even mean

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
colinsurprenant: well, it is implicitly checked below with `if (this.head + length + CHECKSUM_SIZE > capacity)` ...

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
colinsurprenant: +1

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
colinsurprenant: we ... it's not necessarily while recovering, this check also happens on `open()` and is basically an integrity check and I have not idea how this can actually happen but it's just to make sure we can do the following get___ without triggering a buffer overflow. In this case I wouldn't try to infer what has caused this but prefer to simply report the fact ... but we can reword if it's not clear...

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
colinsurprenant: yeah, no, we did not add validation on the version yet...

logstash-core/src/test/java/org/logstash/ackedqueue/TestSettings.java
jordansissel: > getSettingsCheckpointFilePageMemory

I kind of want to name all the methods this, now. haha.

Thanks for the clarification <3

colinsurprenant: lol - the real story is, I was reading some tests around `Queue.open()` and stumbled on `getSettingsCheckpointFilePageMemory()` and thought WTF is this setting?!?!  I seriously had NO idea what this settings was about :P

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
colinsurprenant: not sure how to reword.

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
colinsurprenant: so a fiew check is made to make sure length is greater than zero, and then if the length is actually valid with respect to the buffer size:
```
        // length must be > 0
        if (length <= 0) { throw new PageIOInvalidElementException("Element invalid length"); }

        // if there is no room for the proposed data lenght and checksum just stop here
        if (this.head + length + CHECKSUM_SIZE > capacity) { throw new PageIOInvalidElementException("cannot read element payload and checksum past buffer capacity"); }
```

is that ok with you?

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
colinsurprenant: fixed

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
jordansissel: ++

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
jordansissel: I'm not sure how to reword it at this time, either.

How about adding a TODO comment above indicating that we know this isn't a great exception message but at the time of writing we couldn't come up with anything better.

logstash-core/src/test/java/org/logstash/ackedqueue/TestSettings.java
jordansissel: haha.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -162,33 +163,45 @@ public void open() throws IOException {
         for (int pageNum = headCheckpoint.getFirstUnackedPageNum(); pageNum < headCheckpoint.getPageNum(); pageNum++) {
```
talevy: shouldn't this `headCheckpoint.getFirstUnackedPageNum()` happen after you optimistically recover?

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -223,30 +238,30 @@ private void add(Checkpoint checkpoint, PageIO pageIO) throws IOException {
             } else {
                 // create a tail page with a null PageIO and add it to tail pages but not unreadTailPages
                 // since it is fully read because also fully acked
+                // TODO: I don't like this null pageIO tail page...
```
talevy: neither do I

logstash-core/src/main/java/org/logstash/ackedqueue/TailPage.java
```diff
@@ -14,19 +14,14 @@ public TailPage(HeadPage page) {
     }
 
     // create a new TailPage object for an exiting Checkpoint and data file
+    // @param @pageIO is expected to be open/recover/create
```
talevy: can we use proper javadoc linking syntax here?

logstash-core/src/main/java/org/logstash/ackedqueue/TailPage.java
```diff
@@ -14,19 +14,14 @@ public TailPage(HeadPage page) {
...
         super(checkpoint.getPageNum(), queue, checkpoint.getMinSeqNum(), checkpoint.getElementCount(), checkpoint.getFirstUnackedSeqNum(), new BitSet(), pageIO);
 
         // this page ackedSeqNums bitset is a new empty bitset, if we have some acked elements, set them in the bitset
         if (checkpoint.getFirstUnackedSeqNum() > checkpoint.getMinSeqNum()) {
             this.ackedSeqNums.flip(0, (int) (checkpoint.getFirstUnackedSeqNum() - checkpoint.getMinSeqNum()));
         }
-
-        if (pageIO != null) {
-            // open the data file and reconstruct the IO object internal state
-            pageIO.open(checkpoint.getMinSeqNum(), checkpoint.getElementCount());
```
talevy: the only reason we are safe in removing this is because this is the only place we construct?

https://github.com/elastic/logstash/pull/6519/files#diff-69b2c41a7ba6596afa233505525212bbR169

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
talevy: nitpick, can we expand this from being a one-liner?

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -2,6 +2,7 @@
 
 import org.logstash.ackedqueue.Queueable;
 import org.logstash.ackedqueue.SequencedList;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
```
talevy: this doesn't seem to be used. why not have this be thrown for all of those `/** zilch **/` spaces?

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -41,208 +24,41 @@ public ByteBufferPageIO(int capacity) throws IOException {
...
     @Override
-    public int persistedByteCount(int byteCount) {
-        return ByteBufferPageIO._persistedByteCount(byteCount);
-    }
-
-    private long maxSeqNum() {
-        return this.minSeqNum + this.elementCount - 1;
-    }
-
+    protected ByteBuffer getBuffer() { return this.buffer; }
```
talevy: expand to multiline

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -41,208 +24,41 @@ public ByteBufferPageIO(int capacity) throws IOException {
...
 
     // below public methods only used by tests
 
-    public int getWritePosition() {
-        return this.head;
-    }
+    // TODO: static method for tests - should refactor
+    public static int _persistedByteCount(int byteCount) { return SEQNUM_SIZE + LENGTH_SIZE + byteCount + CHECKSUM_SIZE; }
```
talevy: expand to multiline.

can we remove the `_` part? and if it is only used in tests, we can make this package-private.

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -41,208 +24,41 @@ public ByteBufferPageIO(int capacity) throws IOException {
...
 
-    public int getElementCount() {
-        return this.elementCount;
-    }
+    public int getWritePosition() { return this.head; }
```
talevy: multiline

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
```diff
@@ -41,208 +24,41 @@ public ByteBufferPageIO(int capacity) throws IOException {
...
 
-    public byte[] dump() {
-        return this.buffer.array();
-    }
+    public byte[] dump() { return this.buffer.array(); }
```
talevy: multiline

logstash-core/src/main/java/org/logstash/common/io/MmapPageIO.java
```diff
@@ -1,199 +1,74 @@
...
+    @Override
+    public void recover() throws IOException {
+        mapFile(!STRICT_CAPACITY);
+        super.recover();
+    }
 
-                this.offsetMap.add(head);
-                this.head += SEQNUM_SIZE;
+    // memory map data file to this.buffer and read initial version byte
+    // @param strictCapacity if true verify that data file size is same as confgured page capcity, if false update page capcity to actual file size
```
talevy: proper javadoc format here?

logstash-core/src/main/java/org/logstash/common/io/PageIO.java
```diff
@@ -10,10 +10,20 @@
 public interface PageIO extends Closeable {
 
     // the concrete class should be constructed with the pageNum, capacity and dirPath attributes
+    // and open/recover/create must first be called to setup with physical data file
+    //
+    // TODO: we should probably refactor this with a factory to force the creation of a fully
```
talevy: 👍 . why not do this for this PR?

logstash-core/src/main/java/org/logstash/common/io/PageIO.java
```diff
@@ -10,10 +10,20 @@
...
+    //       initialized concrete object with either open/recover/create instead of allowing
+    //       a partially initialized object using the concrete class constructor. Not sure.
 
     // open an existing data container and reconstruct internal state if required
     void open(long minSeqNum, int elementCount) throws IOException;
 
+    // optimistically recover an existing data container and reconstruct internal state
+    // with the actual read/recovered data. this is only useful il reading a head page
```
talevy: proper javadoc? `s/il/if`

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
```diff
@@ -0,0 +1,276 @@
...
+                readNextElement(this.minSeqNum + i, VERIFY_CHECKSUM);
+                this.elementCount += 1;
+            } catch (PageIOInvalidElementException e) {
+                // simply stop at first invalid element
+                break;
+            }
+        }
+
+        // if we were not able to read any element just reset minSeqNum to zero
+        if (this.elementCount <= 0) {
```
jordansissel: the `<` part  may not be necessary here. Since we only increment elementCount and it starts at 0 above.

Maybe use an assert to crash if elementCount < 0, then make this check be `if (this.elementCount == 0)` instead of `<= 0` ?

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
```diff
@@ -0,0 +1,276 @@
...
+        this.minSeqNum = 0L;
+        this.elementCount = 0;
+    }
+
+    @Override
+    public void write(byte[] bytes, long seqNum) throws IOException {
+        write(bytes, seqNum, bytes.length, checksum(bytes));
+    }
+
+    protected int write(byte[] bytes, long seqNum, int length, int checksum) {
```
jordansissel: Possibly out of scope for this change, but since `bytes.length` is known, does this method need to take a `length` parameter?

logstash-core/src/main/java/org/logstash/common/io/wip/MemoryPageIOStream.java
```diff
@@ -70,6 +71,11 @@ public MemoryPageIOStream(int capacity, byte[] initialBytes) throws IOException
     }
 
     @Override
+    public void recover() throws IOException {
+        throw new NotImplementedException();
```
jordansissel: Curiosity - Is this the right exception to throw, or should we make our own class for this?

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
colinsurprenant: yes, because it allows forcing each value independently of the actual bytes content. note that this is not a public method, and it is used by tests to inject incorrect values.

logstash-core/src/main/java/org/logstash/common/io/wip/MemoryPageIOStream.java
colinsurprenant: did not pay much attention to this, as it is "unused" code yet ("wip" namespace) and this is a dev-time place holder just to make sure it crashes if someone starts to work with it and run tests for example - just a friendly remainder it needs to be implemented. :P

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
colinsurprenant: the optimistic recovery only applies to the head page elements.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
colinsurprenant: I will refactor/clean this but not in this PR

logstash-core/src/main/java/org/logstash/ackedqueue/TailPage.java
colinsurprenant: oh - yeah - muscle memory :P

logstash-core/src/main/java/org/logstash/ackedqueue/TailPage.java
colinsurprenant: I removed all PageIO opening from the *Page classes and a properly opened PageIO must be fed to the *Page

logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
colinsurprenant: I personally prefer it as a one-liner - I really like the fact that this validation code uses less visual real estate and makes the general flow more readable. Again, personal preference.

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: will remove. 
the `/** zilch **/` should stay as-is since these are no-ops methods, not `NotImplementedException`s

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: I do prefer this concise style in such cases.

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: you mean `protected` ?

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: I prefer it single line

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: I prefer it single line

logstash-core/src/main/java/org/logstash/common/io/MmapPageIO.java
colinsurprenant: absolutely :P

logstash-core/src/main/java/org/logstash/common/io/PageIO.java
colinsurprenant: well, it is not really related to this PR and non-critical - the other refactors were triggered by extremely non-DRY code which I think was justified while working on this. I think we should open a followup issue

logstash-core/src/main/java/org/logstash/common/io/PageIO.java
colinsurprenant: yup

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
talevy: hmm, I see. so if the head checkpoint is old and this value is incorrect, it doesn't matter

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
talevy: ok

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
talevy: got it

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
talevy: ok

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
talevy: no, just no `public`. just `static int`. will allow tests to see it since I assume they are in `org.logstash.common.io`

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
talevy: ok

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
talevy: ok

logstash-core/src/main/java/org/logstash/common/io/PageIO.java
talevy: oh, ok. I just didn't see this TODO copy/pasted from anywhere, so I assumed this was a related note

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
colinsurprenant: well ... the head checkpoint is the only way to retrieve the `firstUnackedPageNum` so we have no choice to rely on it. I guess the worst case would be that `firstUnackedPageNum` points to a page that is now fully acked which should not be a problem (need to verify) but if that page has been purged it might be problematic. I will open a followup issue to verify this edge case.

colinsurprenant: followup in #6592

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
colinsurprenant: followup in #6593

logstash-core/src/main/java/org/logstash/ackedqueue/TailPage.java
colinsurprenant: fixed

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
colinsurprenant: so - at this point it cannot "easily" be other than public because there are many usage outside that package ... we should defer that to another issue/pr as this is really out of scope for this pr. ok?

colinsurprenant: followup in #6594

logstash-core/src/main/java/org/logstash/common/io/MmapPageIO.java
colinsurprenant: oh you mean using `/** */` ?
I see I have been using `//` type of comments everywhere and this is not javadoc compatible, I will need to do a global cleanup.

colinsurprenant: followup in #6595

colinsurprenant:
@jordansissel @talevy thanks for the reviews/comments - just push a few commits and followups.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
talevy: hmm. after trying to force this to become an issue... I can't. Since purging of tail files only happens on ack... a checkpoint would also be created. There would only be an issue if something killed the process.. in which case, anything can be corrupt

logstash-core/src/main/java/org/logstash/common/io/ByteBufferPageIO.java
talevy: still not removed?

jordansissel:
LGTM

colinsurprenant:
rebased + squashed and merged in master and 5.x

jsvd:
for future reference the commit on master for this PR is https://github.com/elastic/logstash/commit/356dd71e1e69b9fb652757cf0a4b72b89eca3248

Issue #6521, opened 2017-01-11
Support multiple separate pipelines in a single logstash instance


jsvd:
Currently logstash supports running only a single instance of a pipeline, but it would be useful to run multiple pipelines.

Supporting multiple pipelines has several benefits:
 * simplifying event flow conditionals in complex pipeline configs
 * isolating impact of failures and backpressure in multi flow pipelines configs
 * tuning workers, queueing and batching for different event flows

~~The simplest way to implement and use this feature is to add a toggle feature that, when enabled, reads from the `config/logstash.yml` a list of pipelines~~
This feature will have logstash read a `pipelines.yml` file where pipelines can be described individually. Example:

```yaml
# pipelines.yml
- pipeline.id: line
  config.string: "input { generator {} } filter { sleep { time => 0.01 } } output { stdout { codec => line { format => '%{sequence}' } } }"
  pipeline.workers: 1
- pipeline.id: line2
  path.config: "/etc/conf.d/logstash/line2.*.cfg"
  pipeline.workers: 5
```

TODO:

- [x] define set of configurable settings per pipeline https://github.com/elastic/logstash/issues/6492
- [x] enforce validation of set of configurable settings per pipeline https://github.com/elastic/logstash/issues/6541
- [x] make queue path depend on pipeline id https://github.com/elastic/logstash/issues/6513
- [x] make metrics api work with multi pipeline https://github.com/elastic/logstash/issues/6531
- [x] make -e and -f mutually exclusive https://github.com/elastic/logstash/pull/6976
- [x] don't concatenate default plugins on -f https://github.com/elastic/logstash/pull/6974
- [x] reload pipeline if pipeline settings changed https://github.com/elastic/logstash/pull/7073

Check the feature branch [feature/multi_pipeline](https://github.com/elastic/logstash/tree/feature/multi_pipeline) or [pr #6525](https://github.com/elastic/logstash/pull/6525) for progress

ph:
I've been working on #6514 and I would like to suggest a few things for the multiple pipelines, theses comme from the observation of the current code base and also to deal with future features.

I've been working on two features that resolve around how we load our configuration and how we parse it.

- The first is to make sure when a configuration is loaded that we keep the lineage of the file, this information is really necessary to give better tooling for the
user to allow them to debug complex multi files logstash environment. This will be used by the LIR and future monitoring work.

- The second is to decouple the loading of the configuration from the agent into a another class, allowing developper to create universal plugins that can override source of the configuration.

[...]

PR #6525, opened 2017-01-12
implement multi pipeline support

jsvd:
TODO

- [x] support reloading of pipeline if parameters such as worker count changed
- [x] support subset of pipeline settings in the pipelines.yml file using PipelineSettings class
- [x] better testing
- [x] proper pipelines.yml examples
- [x] log message about ignoring pipelines.yml during a `MultiLocal#match?`
- [x] more testing on broken scenarios (incorrect yaml, missing yaml, lack of config.string AND path.config)

Documentation changes
- [x] describe multiple pipelines
- [ ] adapt pipeline api documentation (_node/pipeline and _node/stats/pipeline) to the new plural versions (_node/pipelines/<id>? and _node/stats/pipelines/<id>?)

solves #6521

### DIFF
--- a/config/pipelines.yml
+++ b/config/pipelines.yml
@@ -1,0 +1,69 @@
+# List of pipelines to be loaded by Logstash
+#
+# This document must be a list of dictionaries/hashes, where the keys/values are pipeline settings.
+# Default values for ommitted settings are read from the `logstash.yml` file.
+# When declaring multiple pipelines, each MUST have its own `pipeline.id`.
+#
+# Example of two pipelines:
+#
+# - pipeline.id: test
+#   pipeline.workers: 1
+#   pipeline.batch.size: 1
+#   config.string: "input { generator {} } filter { sleep { time => 1 } } output { stdout { codec => dots } }
+# - pipeline.id: another_test
+#   queue.type: persisted
+#   path.config: "/tmp/logstash/*.config"
+#
+# Available options:
+#
+#   # name of the pipeline
+#   pipeline.id: mylogs
+#
+#   # The configuration string to be used by this pipeline
+#   config.string: "input { generator {} } filter { sleep { time => 1 } } output { stdout { codec => dots } }"
+#
+#   # The path from where to read the configuration text
+#   path.config: "/etc/conf.d/logstash/myconfig.cfg"
+#
+#   # How many worker threads execute the Filters+Outputs stage of the pipeline
+#   pipeline.workers: 1 (actually defaults to number of CPUs)
+#
+#   # How many events to retrieve from inputs before sending to filters+workers
+#   pipeline.batch.size: 125
+#
+#   # How long to wait before dispatching an undersized batch to filters+workers
+#   pipeline.batch.delay: 5
+#
+#   # How many workers should be used per output plugin instance
+#   pipeline.output.workers: 1
+#
+#   # Internal queuing model, "memory" for legacy in-memory based queuing and
+#   # "persisted" for disk-based acked queueing. Defaults is memory
+#   queue.type: memory
+#
+#   # If using queue.type: persisted, the page data files size. The queue data consists of
+#   # append-only data files separated into pages. Default is 250mb
+#   queue.page_capacity: 250mb
+#
+#   # If using queue.type: persisted, the maximum number of unread events in the queue.
+#   # Default is 0 (unlimited)
+#   queue.max_events: 0
+#
+#   # If using queue.type: persisted, the total capacity of the queue in number of bytes.
+#   # Default is 1024mb or 1gb
+#   queue.max_bytes: 1024mb
+#
+#   # If using queue.type: persisted, the maximum number of acked events before forcing a checkpoint
+#   # Default is 1024, 0 for unlimited
+#   queue.checkpoint.acks: 1024
+#
+#   # If using queue.type: persisted, the maximum number of written events before forcing a checkpoint
+#   # Default is 1024, 0 for unlimited
+#   queue.checkpoint.writes: 1024
+#
+#   # If using queue.type: persisted, the interval in milliseconds when a checkpoint is forced on the head page
+#   # Default is 1000, 0 for no periodic checkpoint.
+#   queue.checkpoint.interval: 1000
+#
+#   # Enable Dead Letter Queueing for this pipeline.
+#   dead_letter_queue.enable: false
--- a/logstash-core/lib/logstash/api/commands/node.rb
+++ b/logstash-core/lib/logstash/api/commands/node.rb
@@ -9,7 +9,7 @@ module LogStash
 
         def all(selected_fields=[])
           payload = {
-            :pipeline => pipeline,
+            :pipelines => pipelines,
             :os => os,
             :jvm => jvm
           }
@@ -17,12 +17,20 @@ module LogStash
           payload
         end
 
-        def pipeline(pipeline_id = LogStash::SETTINGS.get("pipeline.id").to_sym)
-          stats = extract_metrics(
-            [:stats, :pipelines, pipeline_id, :config],
+        def pipelines
+          pipeline_ids = service.get_shallow(:stats, :pipelines).keys
+          pipeline_ids.each_with_object({}) do |pipeline_id, result|
+            result[pipeline_id] = pipeline(pipeline_id)
+          end
+        end
+
+        def pipeline(pipeline_id)
+          extract_metrics(
+            [:stats, :pipelines, pipeline_id.to_sym, :config],
             :workers, :batch_size, :batch_delay, :config_reload_automatic, :config_reload_interval
           )
-          stats.merge(:id => pipeline_id)
+        rescue
+          {}
         end
 
         def os
--- a/logstash-core/lib/logstash/api/commands/stats.rb
+++ b/logstash-core/lib/logstash/api/commands/stats.rb
@@ -41,14 +41,21 @@ module LogStash
         def events
           extract_metrics(
             [:stats, :events],
-            :in, :filtered, :out, :duration_in_millis
+            :in, :filtered, :out, :duration_in_millis, :queue_push_duration_in_millis
           )
         end
 
-        def pipeline(pipeline_id = LogStash::SETTINGS.get("pipeline.id").to_sym)
-          stats = service.get_shallow(:stats, :pipelines, pipeline_id)
-          stats = PluginsStats.report(stats)
-          stats.merge(:id => pipeline_id)
+        def pipeline(pipeline_id = nil)
+          if pipeline_id.nil?
+            pipeline_ids = service.get_shallow(:stats, :pipelines).keys
+            pipeline_ids.each_with_object({}) do |pipeline_id, result|
+              result[pipeline_id] = plugins_stats_report(pipeline_id)
+            end
+          else
+            { pipeline_id => plugins_stats_report(pipeline_id) }
+          end
+        rescue # failed to find pipeline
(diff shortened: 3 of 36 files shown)


logstash-core/lib/logstash/api/commands/node.rb
```diff
@@ -9,20 +9,28 @@ class Node < Commands::Base
...
 
-        def pipeline(pipeline_id = LogStash::SETTINGS.get("pipeline.id").to_sym)
+        def pipelines
+          pipeline_ids = service.get_shallow(:stats, :pipelines).keys
+          result = {}
+          pipeline_ids.each do |pipeline_id|
+            result[pipeline_id] = pipeline(pipeline_id)
+          end
+          result
+        end
```
ph: You can use `#each_with_object` here and remove the need to create result.

```ruby
["main", "super-extra"].each_with_object({}) { |v, hsh| hsh[v] = ".." }
# => {"main"=>"..", "super-extra"=>".."}
```

logstash-core/lib/logstash/api/commands/stats.rb
```diff
@@ -41,14 +41,25 @@ def process
...
+              stats = service.get_shallow(:stats, :pipelines, pipeline_id.to_sym)
+              result[pipeline_id] = PluginsStats.report(stats)
+            end
+          else
+            stats = service.get_shallow(:stats, :pipelines, pipeline_id.to_sym)
+            result[pipeline_id] = PluginsStats.report(stats)
+          end
+          result
+        rescue # failed to find pipeline
+          {}
```
ph: I see similar logic in the `true/false` branch could we refactor this into 2 methods?

logstash-core/lib/logstash/bootstrap_check/default_config.rb
```diff
@@ -4,17 +4,9 @@
...
-
-      if settings.get("config.reload.automatic") && settings.get("path.config").nil?
-        # there's nothing to reload
-        raise LogStash::BootstrapCheckError, I18n.t("logstash.runner.reload-without-config-path")
+      # currently none of the checks applies if there are multiple pipelines
+      if settings.get("config.reload.automatic") && settings.get_setting("config.string").set?
+        raise LogStash::BootstrapCheckError, I18n.t("logstash.runner.reload-with-config-string")
       end
     end
   end
```
ph: self note look at what we have decided for backward compatibility.

logstash-core/lib/logstash/config/source/multi_local.rb
```diff
@@ -0,0 +1,64 @@
+# encoding: utf-8
+require "logstash/config/source/local"
+require "logstash/util/loggable"
+require "logstash/pipeline_settings"
+
+module LogStash module Config module Source
+
```
ph: newline?

logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -110,6 +111,11 @@ def add_source(new_source)
       @sources_lock.synchronize { @sources << new_source }
     end
 
+    def reset_sources
+      logger.debug("Clearing all sources")
+      @sources_lock.synchronize { @sources = Set.new }
+    end
```
ph: The SourceLoader is not a singleton anymore so I think this we don't need this?

logstash-core/lib/logstash/pipeline_settings.rb
```diff
@@ -0,0 +1,35 @@
...
+    SETTINGS_WHITE_LIST = [
+      "pipeline.id", "pipeline.system", "pipeline.reloadable",
+      "pipeline.workers", "pipeline.output.workers",
+      "pipeline.batch.size", "pipeline.batch.delay",
+      "config.string", "path.config", "metric.collect", "config.debug",
+      "queue.max_events", "queue.checkpoint.acks", "queue.checkpoint.writes", "queue.checkpoint.interval",
+      "queue.type", "queue.page_capacity", "queue.max_bytes", "path.queue", "queue.drain",
+      "dead_letter_queue.enable",
+      "config.reload.automatic", "config.reload.interval"
+    ]
```
ph: We have a lot of whitelisted option, I think we should have 1 option per line. We can easily sort it and keep it sorted that way.

logstash-core/lib/logstash/settings.rb
```diff
@@ -140,6 +137,11 @@ def validate_all
       end
     end
 
+    def ==(other)
+      return false unless other.kind_of?(::LogStash::Settings)
+      self.to_hash == other.to_hash
+    end
```
ph: +1, I've looked at `to_hash` method of individual setting. I believe it work for normal type, but shouldn't we change the behavior of `to_hash` in the case of `ArrayCoercible`?

logstash-core/lib/logstash/state_resolver.rb
```diff
@@ -21,7 +21,7 @@ def resolve(pipelines, pipeline_configs)
         else
           # TODO(ph): The pipeline should keep a reference to the original PipelineConfig
           # and we could use straight comparison.
```
ph: I think we can remove the todo.

logstash-core/spec/support/helpers.rb
```diff
@@ -45,8 +45,13 @@ def mock_pipeline(pipeline_id, reloadable = true, config_hash = nil)
   settings = mock_settings("pipeline.id" => pipeline_id.to_s,
                            "config.string" => config_string,
                            "config.reload.automatic" => reloadable)
-  pipeline = LogStash::Pipeline.new(config_string, settings)
-  pipeline
+  pipeline_config = mock_pipeline_config(pipeline_id, config_string, settings)
+  LogStash::Pipeline.new(pipeline_config)
+end
```
ph: +1 for the change!

logstash-core/lib/logstash/bootstrap_check/default_config.rb
ph: Just to clarify, if we have pipelines defined in `pipeline.yml` -e doesn't work? What is happening to -f?

logstash-core/lib/logstash/api/commands/node.rb
jsvd: I've never been a fan of those accum iterators, but I guess it makes sense here

logstash-core/lib/logstash/bootstrap_check/default_config.rb
jsvd: I'll add a couple of tests for this, if -e or -f is enabled, then pipelines.yml is ignored.

logstash-core/lib/logstash/settings.rb
jsvd: I'm not sure of your concern here, I added a few tests to the ArrayCoercible in https://github.com/elastic/logstash/pull/6525/files#diff-bb0f568c71d6288772b068b4f1830c33
let me know if this helps

logstash-core/lib/logstash/bootstrap_check/default_config.rb
jsvd: I've added tests to MultiLocal#match? https://github.com/elastic/logstash/pull/6525/files#diff-8b6c09e2ff51e8eb5173de51308f8182R21

Only here it's possible to decide to load or not the pipelines.yml.

qa/integration/Gemfile
```diff
@@ -1,2 +1,3 @@
 source "https://rubygems.org"
 gemspec
+gem "logstash-devutils", :path => "/Users/joaoduarte/projects/logstash-devutils"
```
ph: lets removed that when we publish the new devutils

ph:
If you are using `-e` we ignore anything that could be present in `config/pipelines.yml`, I think we should log something so the user can see that behavior.

qa/integration/Gemfile
jsvd: 🤦‍♂️ this is was not supposed to have been pushed

ph:
@jsvd Outstanding work!  I've tested it locally and everything seems to work. I just have few comments about logging message to the user when using -f/-e flags on the cli.

its almost ready to merge! I think we are only missing the documentation and the few issues that you have discovered this morning?

config/pipelines.yml
```diff
@@ -0,0 +1,8 @@
+# - pipeline.id: test
+#   pipeline.workers: 1
+#   pipeline.batch.size: 1
+#   config.string: "input { generator {} } filter { sleep { time => 1 } } output { stdout { codec => plain { format => '1' } } }"
+# - pipeline.id: another_test
+#   pipeline.workers: 2
+#   pipeline.batch.size: 1
+#   config.string: "input { generator {} } filter { sleep { time => 1 } } output { stdout { codec => dots } }"
```
ph: I am sure that we will list all the options in the documentation that the user can change, but I think we should add the list from https://github.com/elastic/logstash/pull/6525/files#diff-37d77f4017277cc3a8ad8d054cffb4c5R8 with default values here and make it commented.

jsvd:
@ph I believe I addressed all your concerns, the only thing missing is a documentation section in `docs` about multiple pipelines

suyograo:
@jsvd great work here. Excited to ship this puppy :)

Let's add some documentation here before merging. Otherwise, users would not know how to use this.

ph:
I had a discussion with @jsvd concerning our package and the multiple pipeline feature, our current package rely on `path.config` set to a [wildcard](https://github.com/elastic/logstash/commit/cfecf81091d9603a59f049f7a7926dc55393825f) and it will load all the files that match. I am wondering if we should just ship Logstash with multiple pipeline **on** but with a default *main* pipeline set to the same wildcard path, Something like this.

```yaml
- pipeline.id: main
  path.config: /etc/logstash/conf.d/*.conf
```

This mean that if people want to leverage multiple pipeline they only need to edit `logstash.yml`? This also rise questions about what happen to the`path.config` values in the main `logstash.yml`.

Note this comment doesn't block the multiple pipeline feature in the current state, it is more about user experience.

jsvd:
> I am wondering if we should just ship Logstash with multiple pipeline on but with a default main pipeline set to the same wildcard path

The main reason I'm in favor of this is that unless we do it, from a user's perspective, switching to pipelines.yml is going to be complex as the scripts use -f which disables the yaml file reading.
Another advantage of doing this is that we expose more clearly what logstash loads: instead of having --path.config in init scripts, the pipelines.yml openly states it uses a pipeline called main and loads files from conf.d

I guess the arguments against doing this are:
- if you're only going to use 1 pipeline then the extra yml adds a new file you have to deal with for no gain
- historically logstash tends to not enable a new feature by default (like reloading, persistent queue, dead letter queue)

logstash-core/lib/logstash/api/commands/stats.rb
```diff
@@ -41,14 +41,21 @@ def process
...
+            pipeline_ids = service.get_shallow(:stats, :pipelines).keys
+            pipeline_ids.each_with_object({}) do |pipeline_id, result|
+              result[pipeline_id] = plugins_stats_report(pipeline_id)
+            end
+          else
+            { pipeline_id => plugins_stats_report(pipeline_id) }
+          end
+        rescue # failed to find pipeline
+          {}
         end
```
jsvd: we do, but [in the NodeStats API Modules](https://github.com/elastic/logstash/pull/6525/files#diff-5b309943da7f980637d1fb663cacd01aR13)

jsvd:
failures are related to https://github.com/elastic/logstash/issues/7178, so I'm going to go ahead and merge this to master.

Issue #6527, opened 2017-01-12
Setting pipeline.id is ignored by logstash


jsvd:
Setting `pipeline.id` has no effect on logstash:

```
/tmp/logstash-5.1.1 % grep "pipeline.id" config -r
config/logstash.yml:pipeline.id: test
/tmp/logstash-5.1.1 % bin/logstash -e ""
The stdin plugin is now waiting for input:
[2017-01-12T15:12:42,007][INFO ][logstash.pipeline        ] Starting pipeline {"id"=>"main", "pipeline.workers"=>4, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>500}
[2017-01-12T15:12:42,030][INFO ][logstash.pipeline        ] Pipeline main started
```

PR #6530, opened 2017-01-12
ensure pipeline.id is correctly propagated

jsvd:
fixes https://github.com/elastic/logstash/issues/6527

this PR ensures pipeline.id is correctly propagated, and also simplifies the api between agent and runner, by making `Agent#register_pipeline` receive only one argument, the settings object.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -88,9 +88,9 @@ class LogStash::Agent
   # @param pipeline_id [String] pipeline string identifier
   # @param settings [Hash] settings that will be passed when creating the pipeline.
   #   keys should be symbols such as :pipeline_workers and :pipeline_batch_delay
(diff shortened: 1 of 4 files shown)


logstash-core/lib/logstash/agent.rb
```diff
@@ -88,9 +88,9 @@ def execute
   # @param pipeline_id [String] pipeline string identifier
   # @param settings [Hash] settings that will be passed when creating the pipeline.
   #   keys should be symbols such as :pipeline_workers and :pipeline_batch_delay
-  def register_pipeline(pipeline_id, settings = @settings)
+  def register_pipeline(settings = @settings)
```
ph: Can we remove the fact that we default to an instance variable here? It feel a bit weird to me and could result in unexpected behavior.

jsvd: +1, done

jsvd:
changes done 👍

Issue #6512, opened 2017-01-10
Events "in" metric calculation should happen before inserting into queue


suyograo:
GET `_node/stats` gives event `in`, `out`, `filtered` metrics. With the introduction of PQ, we've added an unbounded buffer capacity between input and filter stage. The metric calculation we do for reporting on incoming events should happen before writing to the queue, instead of doing at the [beginning](https://github.com/elastic/logstash/blob/c03f30fedb3eb440fc8e307aaade57b3f2314fcb/logstash-core/lib/logstash/filter_delegator.rb#L38) of the filter stage.

suyograo:
This is evident if you use PQ and generator config + sleep filter as such: https://gist.github.com/suyograo/659098e66526bf07abe89c3b2997ab12

The events in stats will show lesser than what was actually generated. Generator is basically inserting events as fast as possible, which does not get reflected in the in metric.

ph:
I will take a look.

colinsurprenant:
I think the issue title should read "should happen **before**" ?

ph:
This is not a hard fix, I will wrap the write client into an instrumented client.

ph:
Actually, the referenced code is not where the bug is the `:in` is calculated in the `ReadClient` section and should be done at the `WriteClient` level.

I think the definition we are looking for is the following:

- Input
  - global **in** rate
  - per plugin **out** rate
  - **push_queue_duration_in_millis** This will be new with the PR I am making, looking for a better name?
 
- Filter
  - per plugin: **in**, **out**, **duration_in_millis**

- Output
  - per plugin: **in**, **out**, **duration_in_millis**

- Worker Loop
  - **duration_in_millis**
  - global **out**

suyograo:
Can we rename `push_queue_duration_in_millis` to `queue_push_duration_in_millis`

PR #6532, opened 2017-01-13
Record the `events.in` related statistic in the right place

ph:
This PR changes where the `events.in` are calculated, previously the
values were calculated in the `ReadClient` which was fine before the
addition of the PQ, but this make the stats not accurate when the PQ was
enabled and the producer are a lot faster than the consumer.

These commits change the collection of the metric inside an
instrumented `WriteClient` so both implementation of the client queues will use
the same code.

This also make possible to record `events.out` for every inputs and the
time waiting to push to the queue.

The API is now exposing theses values for each plugins, the events level
and and the pipeline.

Using a pipeline with a sleep filter and PQ we will see this kind of
response from the API.

```json
{
  "duration_in_millis": 438624,
  "in": 3011436,
  "filtered": 2189,
  "out": 2189,
  "queue_push_duration_in_millis": 49845
}
```

Fixes: #6512

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -17,6 +17,7 @@ require "logstash/instrument/namespaced_metric"
 require "logstash/instrument/null_metric"
 require "logstash/instrument/namespaced_null_metric"
 require "logstash/instrument/collector"
+require "logstash/instrument/wrapped_write_client"
 require "logstash/output_delegator"
 require "logstash/filter_delegator"
 require "logstash/queue_factory"
@@ -429,7 +430,8 @@ module LogStash; class Pipeline < BasePipeline
   def inputworker(plugin)
     Util::set_thread_name("[#{pipeline_id}]<#{plugin.class.config_name}")
     begin
-      plugin.run(@input_queue_client)
+      input_queue_client = wrapped_write_client(plugin)
+      plugin.run(input_queue_client)
     rescue => e
       if plugin.stop?
         @logger.debug("Input plugin raised exception during shutdown, ignoring it.",
@@ -616,5 +618,9 @@ module LogStash; class Pipeline < BasePipeline
 
   def draining_queue?
     @drain_queue ? !@filter_queue_client.empty? : false
+  end
+
+  def wrapped_write_client(plugin)
+    LogStash::Instrument::WrappedWriteClient.new(@input_queue_client, self, metric, plugin)
   end
 end; end
--- a/logstash-core/spec/support/mocks_classes.rb
+++ b/logstash-core/spec/support/mocks_classes.rb
@@ -1,79 +1,91 @@
 # encoding: utf-8
 require "logstash/outputs/base"
+require "logstash/inputs/base"
 require "thread"
 
-module LogStash module Outputs
-  class DummyOutput < LogStash::Outputs::Base
-    config_name "dummyoutput"
-    milestone 2
+module LogStash
+  module Inputs
+    class DummyInput < LogStash::Inputs::Base
+      config_name "dummyinput"
 
-    attr_reader :num_closes, :events
-
-    def initialize(params={})
-      super
-      @num_closes = 0
-      @events = []
-    end
-
-    def register
-    end
-
-    def receive(event)
-      @events << event
-    end
-
-    def close
-      @num_closes += 1
+      def run(queue)
+        # noop
+      end
     end
   end
+  module Outputs
+    class DummyOutput < LogStash::Outputs::Base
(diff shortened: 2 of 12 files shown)


logstash-core/spec/support/mocks_classes.rb
```diff
@@ -1,79 +1,91 @@
 # encoding: utf-8
```
ph: Not a lot changed here, only the addition of a `DummyInput`, I've just added a new module and I had to reformat the other code.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -100,7 +100,9 @@ def initialize(config_str, settings = SETTINGS, namespaced_metric = nil)
       raise
     end
     @queue = build_queue_from_settings
-    @input_queue_client = @queue.write_client
+
+    @input_queue_client_factory = LogStash::Instrument::WriteClientFactory.new(queue.write_client, self, metric)
```
andrewvc: TBH I don't think we need a factory at all. If we had multiple implementations of the factory *or* if the factory had complicated construction logic I'd say it'd be a good idea, but as is, i'd just do the wrapping here.

logstash-core/spec/logstash/instrument/write_client_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+
+        it "records at the `plugin level" do
+          expect(snapshot_metric[:pipelines][:main][:plugins][:inputs][myid.to_sym][:queue_push_duration_in_millis].value).to be_between(1, 1000)
+        end
+      end
+    end
+  end
+
+  context "AckedMemoryQueue" do
+    let(:queue) { LogStash::Util::WrappedAckedQueue.create_memory_based("", 1024, 10, 1024) }
```
andrewvc: These really need to be converted to use shared examples between the two different queue types.

logstash-core/lib/logstash/instrument/write_client.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    private
+    def record_metric(method_name, arguments,  size = 1)
+      @events_metrics.increment(:in, size)
+      @pipeline_metrics.increment(:in, size)
+      @plugin_metrics.increment(:out, size)
+
+      clock = @events_metrics.time(:queue_push_duration_in_millis)
+
+      result = @write_client.send(method_name, arguments)
```
andrewvc: Alternatively you could have it accept a const value set to self.method(:foo).

ph: I thought creating the proc would be slower than send since it have to keep local context? I need to benchmark that.

logstash-core/lib/logstash/pipeline.rb
ph: We use the same pattern in the LIR execution branch to encapsulate all the metric logic this is why I went that route with the current ruby implementation.

logstash-core/lib/logstash/instrument/write_client.rb
```diff
@@ -0,0 +1,59 @@
+# encoding: utf-8
+module LogStash module Instrument
+  class WriteClient
```
ph: I will go with WrappedWriteClient since its already in an instrumented module.

logstash-core/lib/logstash/pipeline.rb
andrewvc: If it's consistent then I'm fine keeping it.

logstash-core/lib/logstash/instrument/write_client.rb
andrewvc: I could be totally wrong here!

suyograo:
Pushed this to 5.3.0

logstash-core/lib/logstash/instrument/write_client.rb
ph: We are both wrong :) 

```
Warming up --------------------------------------
               yield   171.322k i/100ms
                send   229.940k i/100ms
               yield     10.087M (± 9.8%) i/s -     49.683M
                send     10.764M (± 8.1%) i/s -     53.346M

Comparison:
                send: 10763622.7 i/s
               yield: 10086697.3 i/s - same-ish: difference falls within error

ph: Code at: https://github.com/ph/experimentations/blob/master/benchmark/yield_vs_send/yield_vs_send.rb

ph:
@andrewvc Ready for another round of review, I believe I have addresses all yours changes.

ph:
I've squash + rebase, I will wait travis run before merging this.

PR #6538, opened 2017-01-16
Generate notice file during artifact build

suyograo:
Adjust our existing NOTICE.txt file to be built every time. This ensures
we collect LICENSE and NOTICE information of all bundled Logstash and non-Logstash
dependencies.

Changes:

* if NOTICE file is present in gems, include that text in our own NOTICE.TXT file.
* we also need to include the whole copyright block in NOTICE (meaning copyright + license)

### DIFF
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -84,7 +84,7 @@ namespace "artifact" do
   task "all" => ["prepare", "build"]
 
   desc "Build a tar.gz of default logstash plugins with all dependencies"
-  task "tar" => ["prepare", "generate_build_metadata"] do
+  task "tar" => ["prepare", "generate_build_metadata", "license:generate-notice-file"] do
     puts("[artifact:tar] Building tar.gz of default plugins")
     build_tar
   end
@@ -451,5 +451,4 @@ namespace "artifact" do
       out.cleanup
     end
   end # def package
-
 end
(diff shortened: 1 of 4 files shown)


rakelib/artifacts.rake
```diff
@@ -84,7 +84,7 @@ namespace "artifact" do
   task "all" => ["prepare", "build"]
 
   desc "Build a tar.gz of default logstash plugins with all dependencies"
-  task "tar" => ["prepare", "generate_build_metadata"] do
+  task "tar" => ["prepare", "generate_build_metadata", "license:generate-notice-file"] do
```
ph: Should it be included in the zip/deb/rpm as well?

rakelib/license.rake
```diff
@@ -0,0 +1,84 @@
...
+          if !notice.empty? || !license.empty?
+            notice_file << "==========================================================================\n"
+            notice_file << "RubyGem: #{name} Version: #{dep.to_spec.version}\n"
+            notice_file << notice
+            notice_file << license
+          end
+        end
+      end
+    end
+  end
```
ph: Minor thing, it would be nice if the dependency were sorted by name, but its not mandatory.

ph:
As a note, I was able to generate the tarball with the correct file and I've checked the content and it appear to be OK to me.

suyograo:
@ph yes it should be for all artifacts. Good catch.

suyograo:
@ph updated. I'll leave the sorting bit out for now. Not urgent.

Issue #6513, opened 2017-01-10
set queue path based on pipeline id


jsvd:
With the upcoming multi pipeline features, it's necessary that each pipeline has its own queue, so the files needs to be namespaced in a directory that should be named after the pipeline id.

This extra namespace/directory could either be injected and created during the [build_queue_from_settings](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/pipeline.rb#L140) or somewhere during [create_file_based](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/util/wrapped_acked_queue.rb#L28)

required for multi pipeline feature (issue #6521)

ph:
Could be part of https://github.com/elastic/logstash/issues/6513

jsvd:
@ph still want to do this as part of that PR? I'm ok with merging this and then addressing it in a new PR

ph:
lets merge #6498 and you can create a new pr for it?

suyograo:
Fixed by #6540

PR #6540, opened 2017-01-16
include pipeline id in queue path

jsvd:
this is a preparation for multiple pipelines: since each pipeline will have its own queue, the queue path must include the pipeline id.

This PR is a WIP, since it's also necessary to create the folder inside of `#{path.data}/queue/`

solves https://github.com/elastic/logstash/issues/6513

### DIFF
--- a/logstash-core/lib/logstash/queue_factory.rb
+++ b/logstash-core/lib/logstash/queue_factory.rb
@@ -1,4 +1,5 @@
 # encoding: utf-8
+require "fileutils"
 require "logstash/event"
 require "logstash/namespace"
 require "logstash/util/wrapped_acked_queue"
@@ -24,7 +25,8 @@ module LogStash
         LogStash::Util::WrappedSynchronousQueue.new
       when "persisted"
         # persisted is the disk based acked queue
-        queue_path = settings.get("path.queue")
+        queue_path = ::File.join(settings.get("path.queue"), settings.get("pipeline.id"))
+        FileUtils.mkdir_p(queue_path)
         LogStash::Util::WrappedAckedQueue.create_file_based(queue_path, queue_page_capacity, queue_max_events, checkpoint_max_writes, checkpoint_max_acks, checkpoint_max_interval, queue_max_bytes)
       else
         raise ConfigurationError, "Invalid setting `#{queue_type}` for `queue.type`, supported types are: 'memory_acked', 'memory', 'persisted'"
--- a/logstash-core/spec/logstash/queue_factory_spec.rb
+++ b/logstash-core/spec/logstash/queue_factory_spec.rb
@@ -4,6 +4,7 @@ require "logstash/settings"
 require "stud/temporary"
 
 describe LogStash::QueueFactory do
+  let(:pipeline_id) { "my_pipeline" }
   let(:settings_array) do
     [
       LogStash::Setting::WritableDirectory.new("path.queue", Stud::Temporary.pathname),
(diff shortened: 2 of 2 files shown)


logstash-core/lib/logstash/queue_factory.rb
```diff
@@ -24,7 +24,7 @@ def self.create(settings)
         LogStash::Util::WrappedSynchronousQueue.new
       when "persisted"
         # persisted is the disk based acked queue
-        queue_path = settings.get("path.queue")
+        queue_path = ::File.join(settings.get("path.queue"), settings.get("pipeline.id"))
         LogStash::Util::WrappedAckedQueue.create_file_based(queue_path, queue_page_capacity, queue_max_events, checkpoint_max_writes, checkpoint_max_acks, checkpoint_max_interval, queue_max_bytes)
```
ph: Maybe we could add a test for that?

jsvd:
@colinsurprenant any suggestion on where the sub folder creation should happen? I'm undecided between:

1) In the queue factory class [here](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/queue_factory.rb#L27-L28)
2) In the wrapped acked queue before allocating it, [here](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/util/wrapped_acked_queue.rb#L28-L29)
3) On the java side, maybe [in the constructor of the Queue](https://github.com/elastic/logstash/blob/master/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java#L82-L107) ?
4) ???

logstash-core/lib/logstash/queue_factory.rb
jsvd: I agree, I just created this wip PR to serve as discussion "platform" to decide where to put the "create folder" code.
I'll definitely add a test 👍

ph:
@jsvd oh good question, I do not have a preference.

colinsurprenant:
definitely in the queue factory - the queue code shouldn't have to fiddle with a path, we setup the per-pipeline path as close as possible to the pipeline and hand over a path that the queue will take for granted, otherwise it will be mixed concerns.

logstash-core/lib/logstash/queue_factory.rb
```diff
@@ -24,11 +25,17 @@ def self.create(settings)
         LogStash::Util::WrappedSynchronousQueue.new
       when "persisted"
         # persisted is the disk based acked queue
-        queue_path = settings.get("path.queue")
+        queue_path = ::File.join(settings.get("path.queue"), settings.get("pipeline.id"))
+        create_directory(queue_path)
```
ph: I think we could just directly call `mkdir_p` here, since the create directory doesn't do more.
Unless we want the `#create_directory` method do do some error handling, but IIRC permissions to the path will be validated by the settings option when logstash start.

jsvd: makes sense, done

jsvd:
changes done 👍

suyograo:
@jsvd @colinsurprenant this is a breaking change. Folks who have the old directory layout will not be able to use their queue with this code change. I am ok with breaking this since it's a beta feature, but can we think about ways to add BWC? Maybe we can check if a checkpoint file already exists under `path.queue` and assume it's in a "main" pipeline ID context? Other ideas?

jsvd:
@suyograo very good point, in practice we could add a check that if the id is main then the queue.path doesn't get a subfolder.
Another alternative is to assume that, if you have multi pipeline turned off, then you'll have a single pipeline and its path.queue is unchanged.
Even if multi pipeline doesn't land in 5.3, we could still have the option that and ensure it's always false.

Issue #6509, opened 2017-01-09
invalid yaml produces cryptic error


jpcarey:
Should improve the error thrown when an invalid yaml file is read from logstash.yml.

Current error:
`ERROR: Failed to load settings file from "path.settings". Aborting... path.setting=/etc/logstash/, exception=Psych::SyntaxError, message=>(<unknown>): expected <block end>, but found BlockMappingStart while parsing a block mapping at line 155 column 2`

- Version: 5.1.1
- Operating System: Cetos 7
- Steps to Reproduce:
  - Install RPM
  - In /etc/logstash/logstash.yml, uncomment line 155, leaving the space at the beginning of the line.
  - Start logstash, and you will get the error pasted above.

jsvd:
We're limited to what the ruby's yaml parser gives us:

```
jruby-1.7.25 :017 > YAML.parse("teste:\n    teste.tet:\n  hey: - 1\n")
Psych::SyntaxError: (<unknown>): expected <block end>, but found BlockMappingStart while parsing a block mapping at line 3 column 3
```

The only option is to create a user friendly component that parses these messages, but it seems like a big time investment ..

jpcarey:
I think if the error was more descriptive - that `YAML.parse` failed, and which file - that would be more than sufficient (along with the raw exception, which does contain the line number and character).

Maybe something like:
`ERROR: Failed to parse YAML file: {{path.to.file}}. Aborting... path.setting=/etc/logstash/, exception=Psych::SyntaxError, message=>(<unknown>): expected <block end>, but found BlockMappingStart while parsing a block mapping at line 155 column 2`

jsvd:
@jpcarey makes sense, thank you for the suggestion

suyograo:
Fixed in #6546

PR #6546, opened 2017-01-17
improve user facing error when yaml is incorrect

jsvd:
fixes https://github.com/elastic/logstash/issues/6509

before:

```
% bin/logstash -e ""
ERROR: Failed to load settings file from "path.settings". Aborting... path.setting=/tmp/logstash/config, exception=Psych::SyntaxError, message=>(<unknown>): found character      '\t(TAB)' that cannot start any token. (Do not use \t(TAB) for indentation) while scanning for the next token at line 138 column 1
```

after

```
% bin/logstash -e ""
ERROR: Failed to parse YAML file "/tmp/logstash/config/logstash.yml". Please confirm if the YAML structure is valid (e.g. look for incorrect usage of whitespace or indentation). Aborting... parser_error=>(<unknown>): found character      '\t(TAB)' that cannot start any token. (Do not use \t(TAB) for indentation) while scanning for the next token at line 138 column 1
```

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -170,7 +170,12 @@ class LogStash::Runner < Clamp::StrictCommand
     rescue => e
       # abort unless we're just looking for the help
       unless cli_help?(args)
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

jsvd:
@suyograo I'm thinking of merging this to 5.2 5.x and master, since imo there's no reason to no bring this improved messaging to 5.2.x, wdyt?

suyograo:
+1 to merge to 5.2.0

PR #6559, opened 2017-01-19
[WIP] Initialize `tags` field

joshuaspence:
Currently I am trying to do this:

```
expect(subject.get('tags')).not_to include('_grokparsefailure')
```

This doesn't work, however, because the `tags` field is not always initialized (it is only initialized if the `Event#tag` method is called. As such, I need to do this instead:

```
expect(subject.get('tags') || []).not_to include('_grokparsefailure')
```

This change initializes the `tags` field to `[]` and provides an `Event#tags` method to replace `subject.get('tags')`.


joshuaspence:
I have no idea how to run the tests locally so I am relying on Travis to run them for me.

joshuaspence:
I'm willing to continue working on this, but wanted to get some feedback as to whether or not this is actually a good idea before I proceed.

Issue #6565, opened 2017-01-20
LogStash::Agent metrics after config reloading when reloading a config that raises exception on pipeline.run does not increase the successful reload count


jsvd:
```
Failures:

  1) LogStash::Agent metrics after config reloading when reloading a config that raises exception on pipeline.run does not increase the successful reload count
     Failure/Error: expect { subject.send(:"reload_pipeline!", "main") }.to_not change {
       expected result not to have changed, but did change from 0 to 1
     # ./logstash-core/spec/logstash/agent_spec.rb:575:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:49:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:194:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:180:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:152:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:117:in `run_with_threads'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:102:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:80:in `run'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:178:in `standard_exception_handling'
[...]

PR #6566, opened 2017-01-20
fix return value of start_pipeline when start_workers fails

jsvd:
during Agent#start_pipeline a new thread is launched that executes
a pipeline.run and a rescue block which increments the failed reload counter

After launching the thread, the parent thread will wait for the pipeline
to start, or detect that the pipeline aborted, or sleep and check again.

There is a bug that, if the pipeline.run aborts during start_workers,
the pipeline is still be marked as `ready`, and the thread will continue
running for a very short period of time, while incrementing the failed reload
metric.

During this period of `pipeline.ready? == true` and `thread.alive? == true`,
the parent check code will observe all the necessary conditions to
consider the pipeline.run to be successful and thus increment the success
counter too. This failed reload can then result in both the success and
failure reload count being incremented.

This commit changes the parent thread check to use `pipeline.running?`
instead of `pipeline.ready?` which is the next logical state transition,
and ensures it is only true if `start_workers` runs successfully.

fixes https://github.com/elastic/logstash/issues/6565

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -313,7 +313,7 @@ class LogStash::Agent
     while true do
       if !t.alive?
         return false
-      elsif pipeline.ready?
+      elsif pipeline.running?
         return true
       else
         sleep 0.01


jsvd:
this can be verified by applying the following patch which exposes the race condition:

```patch
diff --git a/logstash-core/lib/logstash/agent.rb b/logstash-core/lib/logstash/agent.rb
index b3e4ccd..c6e86bf 100644
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -301,6 +301,7 @@ class LogStash::Agent
       begin
         pipeline.run
       rescue => e
+        sleep 5
         @instance_reload_metric.increment(:failures)
         @pipeline_reload_metric.namespace([id.to_sym, :reloads]).tap do |n|
           n.increment(:failures)
```

And then:

```
/tmp/logstash (git)-[master] % bin/rspec ./logstash-core/spec/logstash/agent_spec.rb:573
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.4 - omit version 2.7.3
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.0 - omit version 2.7.3
--- jar coordinate com.fasterxml.jackson.core:jackson-core already loaded with version 2.7.4 - omit version 2.7.3
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.4 - omit version 2.7.3
--- jar coordinate com.fasterxml.jackson.core:jackson-core already loaded with version 2.7.4 - omit version 2.7.3
Sending Logstash's logs to  which is now configured via log4j2.properties
Run options:
  include {:locations=>{"./logstash-core/spec/logstash/agent_spec.rb"=>[573]}}
  exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}
F

Failures:

  1) LogStash::Agent metrics after config reloading when reloading a config that raises exception on pipeline.run does not increase the successful reload count
     Failure/Error: expect { subject.send(:"reload_pipeline!", "main") }.to_not change {
       expected result not to have changed, but did change from 0 to 1
     # ./logstash-core/spec/logstash/agent_spec.rb:574:in `(root)'
     # /tmp/logstash/lib/bootstrap/rspec.rb:13:in `(root)'

Finished in 5.82 seconds (files took 1.84 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./logstash-core/spec/logstash/agent_spec.rb:573 # LogStash::Agent metrics after config reloading when reloading a config that raises exception on pipeline.run does not increase the successful reload count

Randomized with seed 10271
```

guyboertje:
LGTM

We really need to embrace FSM more. Especially with all the connections in plugins. Is there a case for a generic reusable starting|running|stopping|stopped|retrying FSM?

jsvd:
clearly we need to rewrite it all in erlang http://erlang.org/doc/man/gen_fsm.html

PR #6571, opened 2017-01-22
No longer supported 'remove'

tac0x2a:
remove is notsupported. should be used `uninstall` instead of remove

```
bin/logstash-plugin remove logstash-output-kafka
ERROR: No such sub-command 'remove'

See: 'bin/logstash-plugin --help'
```


ph:
@tac0x2a Actually `uninstall` is deprecated see https://github.com/elastic/logstash/blob/master/lib/pluginmanager/main.rb#L38 and we now push `remove`, so all the elastic product are aligned on the same interface.

Issue #6341, opened 2016-12-01
Test failure in `When auto_reload is false when calling reload_pipeline`


ph:
```
Failures:
  1) LogStash::Agent#execute when auto_reload is false when calling reload_pipeline! with a config that does not contain reload incompatible plugins does upgrade the new config
     Failure/Error: expect(subject).to receive(:upgrade_pipeline).once.and_call_original
       (#<LogStash::Agent:0x215ac1b2>).upgrade_pipeline(any args)
           expected: 1 time with any arguments
           received: 0 times with any arguments
     # ./logstash-core/spec/logstash/agent_spec.rb:139:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:49:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:248:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:243:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:180:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/task.rb:173:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:152:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:117:in `run_with_threads'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:102:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:80:in `run'
     # ./vendor/bundle/jruby/1.9/gems/rake-11.3.0/lib/rake/application.rb:178:in `standard_exception_handling'
[...]

PR #6574, opened 2017-01-23
use running? predicate instead of ready? in some agent specs

jsvd:
the pipeline class has two state predicates: ready? and running?

ready? becomes true after `start_workers` terminates (succesfuly or not)
running? becomes true before calling `start_flusher`, which means that
`start_workers` is guaranteed to have terminated successfuly

Whenever possible, we should use `running?` instead of `ready?` in the
spec setup blocks. The only place where this may be bad is when the
pipeline execution is short lived (e.g. generator w/small count) and the
spec may never observe pipeline.running? == true

### DIFF
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -155,7 +155,7 @@ describe LogStash::Agent do
 
           it "does not try to reload the pipeline" do
             t = Thread.new { subject.execute }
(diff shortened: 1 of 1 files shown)


jordansissel:
LGTM pending Travis passes.

jsvd:
This build [triggered](https://travis-ci.org/elastic/logstash/jobs/194537661#L971) known spec failure https://github.com/elastic/logstash/issues/6341

Issue #6487, opened 2017-01-05
[docs] incorrect note about unreleased docker image in documentation for 5.x/5.1


weltenwort:
Instead of the instructions for retrieving the logstash docker image, the following paragraph is rendered in the ["Running Logstash on Docker" documentation page](https://www.elastic.co/guide/en/logstash/current/docker.html) even for the 5.x and 5.1 versions:

```
ifeval::["{release-state}"=="unreleased"]

However, version {logstash_version} of Logstash has not yet been
released, so no Docker image is currently available for this version.

endif::[]
```

The `ifeval` condition does not seem have the intended effect.

dedemorton:
#6578 sets the release-state to "released" for 5.1. #6584 fixes the version info in the 5.x (prerelease) docs. "unreleased" is the correct release-state for 5.x.

PR #6578, opened 2017-01-23
Fix versions and release state

dedemorton:
At some point the correct values for the release got overwritten. 

@suyograo Can you confirm that these are the correct values to use for 5.1? I'm not completely sure about the value of `{branch}`. It's used in https://github.com/elastic/logstash/blob/5.1/docs/static/include/pluginbody.asciidoc. 

This PR fixes https://github.com/elastic/logstash/issues/6487


suyograo:
@dedemorton LGTM. Those version numbers are correct now.

Issue #6476, opened 2017-01-03
Logstash slowlog field took_in_millis shows microseconds instead of milliseconds


nick4eva:
I configured Logstash slowlog like this:

slowlog.threshold.warn: 500ms
slowlog.threshold.info: 400ms
slowlog.threshold.debug: 300ms
slowlog.threshold.trace: 100ms

But I see in slowlog file trace entry with "took_in_millis":103180
I don't believe that my event processing 103 seconds, so I looked at source code and found that took_in_millis calculated incorrectly, it should be took_in_nanos / 1000000 instead of took_in_nanos / 1000:

[https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/logging/logger.rb#L109](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/logging/logger.rb#L109)

suyograo:
Fixed in #6481

PR #6579, opened 2017-01-24
Add release notes for 5.2.0

suyograo:


docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+  in the `logstash-plugin` CLI to manage offline plugins. The `pack` and `unpack` subcommands have been 
+  deprecated ({lsissue}6393[Issue 6393]).
+* Fixed an issue where plugins couldn't be installed when HTTP proxy was enabled in Logstash 
+  deployments ({lsissue}6044[Issue 6044], {lsissue}5777[Issue 5777], {lsissue}5966[Issue 5966]).
+* Improved user facing error when `logstash.yml` is invalid ({lsissue}6509[Issue 6509]).
+* Fixed an issue where Logstash slowlog field `took_in_millis` reported values in microseconds instead of 
+  milliseconds ({lsissue}6476[Issue 6476]).
+* Added explicit `fsync` when writing to the checkpoint file to ensure date has been flushed to disk.
+* `jvm.options` is now loaded relative to the location of `logstash.yml` or if a custom `path.settings` is 
+  specified ({lsissue}6379[Issue 6379]).
```
dedemorton: This seems incomplete. Maybe say, "The `jvm.options` file is now loaded relative to the location of `logstash.yml` or, if `path.settings` is  specified, relative to the specified path."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+* Added ability to create a startup init script to the `system-install` utility ({lsissue}6148[Issue 6148]).
+
+[float]
+==== Input Plugins
+
+*`Kafka`*:
+
+* Made error reporting clearer when the connection to Kafka broker fails.
```
dedemorton: "...connection to a Kafka broker fails...."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+* Made a change to set `kerberos` option only when using GSSAPI.
```
dedemorton: "...to set the `kerberos` option...."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+
+*`Ganglia`*:
+
+* Store the name and value of ganglia metrics (https://github.com/logstash-plugins/logstash-input-ganglia/issues/2[Issue 2]).
```
dedemorton: Using an imperative here is a bit awkward and inconsistent. We should say something like: "Added ability to store the name and value of ganglia metrics" or "Enhanced plugin to store the name and value of ganglia metrics..."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+
+*`RabbitMQ`*
+
+* Fixed issue where metadata would be broken for strings values larger than 255 chars in message headers (https://github.com/logstash-plugins/logstash-input-rabbitmq/issues/94[Issue 94])
```
dedemorton: typo. Change to: "...for string values larger...."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+
+==== Filter Plugins
+
+*`GeoIP`*:
+
+* `target` field should be merged and not completely overwritten by looked up GeoIP data (https://github.com/logstash-plugins/logstash-filter-geoip/issues/98[Issue 98]).
```
dedemorton: Phrasing is a bit odd for the relnotes. Maybe say something like: "Fixed a bug that caused the `target` field to be overwritten by looked up GeoIP data. The `target` field is now merged instead of being completely overwritten."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+
+*`Sleep`*:
+
+* Fixed bug causing a crash when `every` config option is set (https://github.com/logstash-plugins/logstash-filter-sleep/issues/5[Issue 5]).
```
dedemorton: Would say, "Fixed a bug that caused the plugin to crash when the `every` config option was set..." [I'm assuming that the plugin crashed...if Logstash crashed, say that instead.]

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+
+* Added support for VMware VDS IPFIX.
+* Fixed 0-length scope field length (Netflow 9, Juniper SRX)
+  
+==== Output Plugins
+
+*`Elasticsearch`*:
+
+* Previously users could specify a proxy configuration as a hash. This has been disabled due to security reasons.
+* Proxy feature has been fixed to work when authentication is specified in the URL (https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/516[Issue 516]).
```
dedemorton: Change to: "The proxy feature has been fixed...."

Also, rather than saying "when authentication is specified" wouldn't it be better to say "when credentials are specified" or "when authentication credentials are specified"? I think of authentication as being the process that checks the credentials.

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+* Fixed a bug that forced users to URL encode the `password` option. Passwords with special characters can be used directly 
+  in the URL or in the `password` option.
+
+*`S3`*:
+
+* This plugin has been updated to use the v2.0 of the AWS SDK which brings in many updates and bug fixes.
```
dedemorton: Just to maintain consistency (for easy scanning), should say, "Updated this plugin to use the v2.0 of the AWS SDK...."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+* Uploading large files to S3 is more efficient since we use S3's `upload_file` method. This
```
dedemorton: For consistency, would say: "Improved efficiency of uploading large files to S3 by using S3's `upload_file` method. This method handles large files in multi-part chunks."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+* Proxy feature has been fixed to work when authentication is specified in the URL (https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/516[Issue 516]).
+* Fixed a bug that forced users to URL encode the `password` option. Passwords with special characters can be used directly 
+  in the URL or in the `password` option.
+
+*`S3`*:
+
+* This plugin has been updated to use the v2.0 of the AWS SDK which brings in many updates and bug fixes.
+* Uploading large files to S3 is more efficient since we use S3's `upload_file` method. This 
+  handles large files in multi-part chunks.
+* Added new option called `server_side_encryption` that allows users to specify type of encryption (https://github.com/logstash-plugins/logstash-output-s3/issues/62[Issue 62])
```
dedemorton: ...to specify the type of encryption..."

suyograo:
@dedemorton applied your suggestions.

dedemorton:
LGTM

nick4eva:
Issue #6481 (PR #6476)  was already fixed in version 5.1.2.
Why it's mentioned in the changelog for version 5.2.0 again?

PR #6584, opened 2017-01-25
Fix version info

dedemorton:
Changed versions to 5.x.0 so that we don't end up with misleading statements like "version 5.1.0 of Logstash has not yet been released, so no Docker image is currently available for this version." 

Even tho the doc is clearly flagged a prerelease, we don't want it to contain info that is blatantly wrong.


suyograo:
LGTM

Issue #6604, opened 2017-01-30
add PQ exclusive access to path.queue using file locking


colinsurprenant:
As discussed in #6505 - we need to introduce exclusive access to the `path.queue` dir in the Persistent Queue to prevent concurrent access and potential corruption to the queue.

colinsurprenant:
implemented in #6656

Issue #6505, opened 2017-01-09
prevent multiple logstash instances from using the same path.data and path.queue


jsvd:
currently there's nothing preventing a user from enabling the persistent queues and starting two logstash instances, likely corrupting the PQ or seeing strange event patterns.

This is a likely scenario in production environments where an init script malfunctions during restart and starts a second logstash instance.

/cc @colinsurprenant

colinsurprenant:
Absolutely, will address this.

colinsurprenant:
I am not sure how this will be possible to address in a practical way - file locking across different processes/JVM has portability issues and race condition problems. 

We could think about a flag file which indicates that the queue is in-use by another process but we have to figure how to deal with it upon a unclean shutdown? - what if LS starts and this flag file is set but there is actually not other LS process actually running?

Any thoughts?

ph:
Maybe it's just to have a `logstash.lock` file in the queue path and try to release it when logstash stop or crash when possible, with the right signal. We will do our possible to release the lock at the end of the execution (keep the pid in the lock?)

But if the *.lock still exist, we can raise a warning and the script will fail to start, these scenario will be easily monitored with monit or even with the monitoring API.

I haven't check the code, but IIRC ES deal with that case too.

jsvd:
After discussing with the team, we decided to broaden the scope of this issue to not just the queue but the path.data too.
So the solution will be to have two .lock files - one in path.data and another in path.queue - and use a file locking mechanism to prevent more than one instance from using those directories.

The mechanism to achieve file locking will depend on the operating system family. On linux systems the .lock file can be opened in exclusive mode, so two logstash instances will not be able to open the same file.
On Windows files are open exclusively already by default, so no extra functionality is required to detect this conflict.

colinsurprenant:
@jsvd @jordansissel  I am not sure we need have different strategies between Windows and Linux if we use the `FileChannel` `FileLock` in a similar way as ES/Lucene per https://github.com/apache/lucene-solr/blob/8c2ef3bc7fbebe8105c2646c81489aa9393ad402/lucene/core/src/java/org/apache/lucene/store/NativeFSLockFactory.java#L86-L129

colinsurprenant:
The queue exclusive access is now solved in #6656

colinsurprenant:
I suggest we close this issue and re-open one specific for locking `path.data`.

PR #6607, opened 2017-01-30
exclusive file locking class and tests

colinsurprenant:
This is a Java class based on Lucene FSLockFactory and friends to support same and  cross-jvm/process exclusive file locking to implement #6604 and #6505.

Included are tests for same process/jvm and `bin/lock` to test cross jvm locking.

### DIFF
--- a/bin/lock
+++ b/bin/lock
@@ -1,0 +1,9 @@
+#!/usr/bin/env bin/ruby
+
+require_relative "../lib/bootstrap/environment"
+LogStash::Bundler.setup!({:without => [:build, :development]})
+require "logstash-core"
+
+lock = Java::OrgLogstash::FileLockFactory.getDefault.obtainLock(ARGV[0], ".lock")
+puts("locking " + File.join(ARGV[0], ".lock"))
+sleep
--- a/logstash-core/src/main/java/org/logstash/FileLockFactory.java
+++ b/logstash-core/src/main/java/org/logstash/FileLockFactory.java
@@ -1,0 +1,98 @@
+// this class is largely inspired by Lucene FSLockFactory and friends, below is the Lucene original Apache 2.0 license:
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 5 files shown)


bin/lock
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env vendor/jruby/bin/jruby
```
talevy: is this file used anywhere?

colinsurprenant: I just pushed a commit to fix this to use the new `bin/ruby`

colinsurprenant: so `bin/lock` is there only to test cross-vm locking.

bin/lock
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bin/ruby
```
talevy: if this is only for tests, shouldn't we put this under our test resources?

logstash-core/src/main/java/org/logstash/FileLockFactory.java
```diff
@@ -0,0 +1,86 @@
+// this class is largely inspired by Lucene FSLockFactory and friends
```
talevy: should put Lucene's copyright here, no?

bin/lock
colinsurprenant: I'm open to suggestions but for now this is executed manually. What do you suggest specifically?

talevy: well, so everything under `bin` isn't copied over to our release distributable? that is my main concern I guess

colinsurprenant: I just added it to the exclusion list

talevy: 👍

logstash-core/src/main/java/org/logstash/FileLockFactory.java
colinsurprenant: yep - added

colinsurprenant:
I agree that the added `bin/lock` is not ideal but it allowed me to manually test the cross-process/JVM locking. I will create an issue to move that in some kind of integration test ... can rats do that? (launch other processes) @suyograo ?

talevy:
@colinsurprenant RATS would be a great place to run any of these tests you are doing manually (that can be automated).

colinsurprenant:
@talevy ok perfect. So I created #6610 for this - I don't think we should block this PR for this and I'd like to move forward with #6604 **today** so let me know if there are any other concerns!

talevy:
LGTM

guyboertje:
LGTM

colinsurprenant:
thanks @talevy @guyboertje  🙌

colinsurprenant:
rebased + squashed and merged into master and 5.x

PR #6617, opened 2017-02-01
make wait time on stop configurable

ctramnitz:
When doing heaving lifting the current 10 seconds may not be sufficient to gracefully stop LS. Make the check period configurable, with 10 being the default (no change to current behaviour).

### DIFF
--- a/pkg/logstash.default
+++ b/pkg/logstash.default
@@ -36,6 +36,9 @@
 #LS_NICE=19
 
 # If this is set to 1, then when `stop` is called, if the process has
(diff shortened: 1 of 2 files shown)


karmi:
Hi @ctramnitz, we have found your signature in our records, but cannot match the e-mails -- your [commit](https://patch-diff.githubusercontent.com/raw/elastic/logstash/pull/6617.patch) has an invalid e-mail address. Can you please add the e-mail you've signed the CLA with to your Github profile (it can be hidden), that should normally solve the issue. Please @-mention me here if the troubles wouldn't go away.

PR #6618, opened 2017-02-01
Default URL if $LOGSTASH_PACK_URL is empty string

jarpy:
When the environment variable $LOGSTASH_PACK_URL is either undefined
or a zero-length string, use the default URL to download the X-Pack
plugin.

You may have already done this, @ph. Please disregard if so.

Thanks.


ph:
thanks @jarpy I have already done that in https://github.com/elastic/logstash/pull/6611, I will merge it in 5.2.x

PR #6621, opened 2017-02-01
support releaseLock

colinsurprenant:
add support to release a lock.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/FileLockFactory.java
+++ b/logstash-core/src/main/java/org/logstash/FileLockFactory.java
@@ -27,9 +27,21 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
+/**
+ * FileLockFactory provides a way to obtain an exclusive file lock for a given dir path and lock name.
+ * The obtainLock() method will return a Filelock object which should be released using the releaseLock()
+ * method. Normally the returned FileLock object should not be manipulated directly. Only the obtainLock()
+ * and releaseLock() methods should be use to gain and release exclusive access.
+ * This is threadsafe and will work across threads on the same JVM and will also work across processes/JVM.
+ *
+ * TODO: because of the releaseLock() method, strictly speaking this class is not only a factory anymore,
+ * maybe we should rename it FileLockManager?
+ */
 public class FileLockFactory {
 
     /**
@@ -39,7 +51,8 @@ public class FileLockFactory {
 
     private FileLockFactory() {}
 
-    private static final Set<String> LOCK_HELD = Collections.synchronizedSet(new HashSet<String>());
+    private static final Set<String> LOCK_HELD = Collections.synchronizedSet(new HashSet<>());
+    private static final Map<FileLock, String> LOCK_MAP =  Collections.synchronizedMap(new HashMap<>());
 
     public static final FileLockFactory getDefault() {
         return FileLockFactory.INSTANCE;
@@ -71,6 +84,7 @@ public class FileLockFactory {
                 channel = FileChannel.open(realLockPath, StandardOpenOption.CREATE, StandardOpenOption.WRITE);
                 lock = channel.tryLock();
                 if (lock != null) {
+                    LOCK_MAP.put(lock, realLockPath.toString());
                     return lock;
                 } else {
                     throw new LockException("Lock held by another program: " + realLockPath);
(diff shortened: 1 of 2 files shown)


logstash-core/src/main/java/org/logstash/FileLockFactory.java
```diff
@@ -39,7 +41,8 @@
 
     private FileLockFactory() {}
 
-    private static final Set<String> LOCK_HELD = Collections.synchronizedSet(new HashSet<String>());
+    private static final Set<String> LOCK_HELD = Collections.synchronizedSet(new HashSet<>());
```
talevy: I'm a little confused by how this is going to be used. I don't necessarily see why there cannot be 
one synchronized object to keep track of these locks.

In what use case do we have context to the file path, and in what contexts would we just have the Lock?

unless I am mistaken, so far this code is only used in the tests, so I am not sure.

colinsurprenant: what is going to be used? this line explicitly or you are generally speaking for the whole class?  ... it is currently in use in my other branch. I am doing this to simplify PRs review and try to not mix concerns. this is a PR about adding support the the lock release.

colinsurprenant: for your synchronisation concerns: this code need to be threadsafe, hence the use of threadsafe structures. exclusive access is necessarily a concurrency concern.

talevy: I guess I am asking for the context of that other branch. Without that context, I can't really say whether I like the usage of both `LOCK_HELD` and `LOCK_MAP`. For example, maybe something can change as to not need both?

If that is not the case, and it is just a matter of fact that some places only have the `Lock`, while others have the `filePath`, I understand.

colinsurprenant: that's not it  -
1- you obtain a lock by specifying a lock dir and a lock name, this returns a FileLock object. 
2- you release a lock by passing back the previously returned FileLock object hence the need for both LOCK_HELD and LOCK_MAP.  
- you want to only be able to release a lock which was first obtained.

talevy: right, but I am not fully understanding why it is the factories job to manage the release of the lock.

FileLocks are closeable, so once you're done locking, you just close the lock and its over. Maybe it makes more sense to simply check whether the lock is both held and closed, and if it is, remove it from `LOCK_HELD` and proceed in obtaining the lock.

But yeah, I am just shooting in the dark here since I don't have context into how it is being used. If everything I say doesn't apply, then ignore me.

logstash-core/src/main/java/org/logstash/FileLockFactory.java
```diff
@@ -95,4 +99,13 @@ public FileLock obtainLock(String lockDir, String lockName) throws IOException {
             throw new LockException("Lock held by this virtual machine: " + realLockPath);
         }
     }
+
+    public void releaseLock(FileLock lock) throws IOException {
```
talevy: only one thread has a reference to the lock at a time, right? so there is no chance that this gets called from two threads with the same lock?

logstash-core/src/main/java/org/logstash/FileLockFactory.java
colinsurprenant: pretty simple: you want to allow releasing a lock if and only if it was first obtained, **thought the `FileLockFactory`**  we don't support any other locking mecanism, it has to be obtained using this class.

Sure, we could rename it to FileLockManager, but I would defer that at this point?

logstash-core/src/main/java/org/logstash/FileLockFactory.java
colinsurprenant: no that should not happen but we have no control on how you use this class really.

logstash-core/src/main/java/org/logstash/FileLockFactory.java
talevy: > pretty simple: you want to allow releasing a lock if and only if it was first obtained

by having a `FileLock` reference, aren't you naturally using something that has been obtained?

logstash-core/src/main/java/org/logstash/FileLockFactory.java
talevy: got it

logstash-core/src/main/java/org/logstash/FileLockFactory.java
colinsurprenant: No, you could use FileLock objects independently of the `FileLockFactory`.

I could have made that more obvious by using a Lock class wrapping the FileLock but I thought it was pretty much unnecessary overhead.

colinsurprenant: I understand that this class is missing documentation on usage, will add that.

colinsurprenant:
thanks @talevy - rebased & squashed, merged into master and 5.x

PR #6630, opened 2017-02-02
add setting class that coerces value to array

jsvd:


logstash-core/lib/logstash/settings.rb
```diff
@@ -489,6 +489,33 @@ def coerce(value)
...
+    class ArrayCoercible < Coercible
+      def initialize(name, klass, default, strict=true, &validator_proc)
+        @element_class = klass
+        super(name, ::Array, default, strict, &validator_proc)
+      end
+
+      def coerce(value)
+        return value if value.is_a?(::Array)
+        Array(value)
+      end
```
ph: No need to check for `is_a?`, Array() will do the right thing.

logstash-core/lib/logstash/settings.rb
```diff
@@ -489,6 +489,33 @@ def coerce(value)
...
+
+        unless input.all? {|el| el.kind_of?(@element_class) }
+          raise ArgumentError.new("Values of setting \"#{@name}\" must be #{@element_class}. Received: #{input.map(&:class)}")
+        end
+
+        if @validator_proc && !@validator_proc.call(input)
+          raise ArgumentError.new("Failed to validate setting \"#{@name}\" with value: #{input}")
+        end
+      end
+    end
```
ph: clever!

PR #6631, opened 2017-02-02
refactor agent pipeline reloading to avoid double live pipelines with same settings

### LINKED ISSUE #6605: Pipeline reload shutdown/initialization sequence
We need to rework the way we sequence the pipeline reload - it really doesn't seem a good idea to allow a two pipeline instances to exist at the same time. I believe a pipeline object should alway be closed/shutdown before a new one can exist when using the same config options. We need to reason on the Pipeline object initialization/closing semantic and implement it correctly. 

We should probably also think about this in #6601. 

A problem we see is that the pipeline reload (upon config change) does not close the "current" pipeline before creating a new pipeline which has the effect of creating a new PQ object using the same data path as the yet non-closed PQ object of the "current" pipeline.

Since there is no guard in place for two PQ instance to open the same data files (#6505, #6604), if one updates the data and the other decides to close and do a final checkpoint with its own state which will most probably be different that the other one, well, the queue becomes corrupt.

Note that while playing with config reloading while LS is actively processing and persisting event in the queue - I was able to trigger 3-4 different types of crash/exceptions related to this dual PQ instances on the same data path.

colinsurprenant:
this fixes #6605 

- refactor agent reload sequence to make sure a pipeline is shutdown before a new one is started with same settings.

- the goal here as stated in https://github.com/elastic/logstash/issues/6605#issuecomment-276751216 is to introduce the minimum changes to fix the dual pipeline problem for the next minor release, knowing that @ph is working on a deeper refactor of the whole config change management.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -207,16 +207,13 @@ class LogStash::Agent
     @collector = LogStash::Instrument::Collector.new
 
     @metric = if collect_metrics?
-                @logger.debug("Agent: Configuring metric collection")
-                LogStash::Instrument::Metric.new(@collector)
-              else
-                LogStash::Instrument::NullMetric.new(@collector)
-              end
+      @logger.debug("Agent: Configuring metric collection")
+      LogStash::Instrument::Metric.new(@collector)
+    else
+      LogStash::Instrument::NullMetric.new(@collector)
+    end
 
-
-    @periodic_pollers = LogStash::Instrument::PeriodicPollers.new(@metric,
-                                                                  settings.get("queue.type"),
-                                                                  self)
+    @periodic_pollers = LogStash::Instrument::PeriodicPollers.new(@metric, settings.get("queue.type"), self)
     @periodic_pollers.start
   end
 
@@ -232,31 +229,40 @@ class LogStash::Agent
     @collect_metric
   end
 
-  def create_pipeline(settings, config=nil)
+  def increment_reload_failures_metrics(id, config, exception)
+    @instance_reload_metric.increment(:failures)
+    @pipeline_reload_metric.namespace([id.to_sym, :reloads]).tap do |n|
+      n.increment(:failures)
+      n.gauge(:last_error, { :message => exception.message, :backtrace => exception.backtrace})
+      n.gauge(:last_failure_timestamp, LogStash::Timestamp.now)
+    end
+    if @logger.debug?
+      @logger.error("fetched an invalid config", :config => config, :reason => exception.message, :backtrace => exception.backtrace)
+    else
+      @logger.error("fetched an invalid config", :config => config, :reason => exception.message)
+    end
+  end
+
+  # create a new pipeline with the given settings and config, if the pipeline initialization failed
+  # increment the failures metrics
+  # @param settings [Settings] the setting for the new pipelines
+  # @param config [String] the configuration string or nil to fetch the configuration per settings
+  # @return [Pipeline] the new pipeline or nil if it failed
+  def create_pipeline(settings, config = nil)
     if config.nil?
       begin
         config = fetch_config(settings)
       rescue => e
         @logger.error("failed to fetch pipeline configuration", :message => e.message)
-        return
+        return nil
       end
     end
 
     begin
       LogStash::Pipeline.new(config, settings, metric)
     rescue => e
-      @instance_reload_metric.increment(:failures)
-      @pipeline_reload_metric.namespace([settings.get("pipeline.id").to_sym, :reloads]).tap do |n|
-        n.increment(:failures)
-        n.gauge(:last_error, { :message => e.message, :backtrace => e.backtrace})
-        n.gauge(:last_failure_timestamp, LogStash::Timestamp.now)
-      end
-      if @logger.debug?
-        @logger.error("fetched an invalid config", :config => config, :reason => e.message, :backtrace => e.backtrace)
-      else
-        @logger.error("fetched an invalid config", :config => config, :reason => e.message)
-      end
-      return
+      increment_reload_failures_metrics(settings.get("pipeline.id"), config, e)
+      return nil
     end
   end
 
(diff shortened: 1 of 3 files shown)


colinsurprenant:
We just discussed with @jordansissel to try to find a solution to get the plugins validation back before the shutdown + start sequence. Looking at it.

logstash-core/lib/logstash/agent.rb
```diff
@@ -232,62 +229,126 @@ def collect_metrics?
...
-      @logger.debug("no configuration change for pipeline",
-                    :pipeline => id, :config => new_config)
+      @logger.debug("no configuration change for pipeline", :pipeline => id, :config => new_config)
       return
     end
 
-    new_pipeline = create_pipeline(old_pipeline.settings, new_config)
+    # check if this pipeline is not reloadable. it should not happen as per the check below
+    # but keep it here as a safety net if a reloadable pipeline was releoaded with a non reloadable pipeline
+    if !old_pipeline.reloadable?
```
colinsurprenant: it should not happen with the code as it is now, but if someone else in the future decide to trigger `update_pipeline` outside this sequence it will prevent it. it is really a defensive check.  I  don't mind removing it but at the same time it is harmless and will prevent it if we end up in a situation were a pipeline was not reloadable. more like an assertion in the code.

logstash-core/lib/logstash/agent.rb
```diff
@@ -232,62 +229,126 @@ def collect_metrics?
...
+      pipeline_validator = LogStash::BasePipeline.new(new_config, old_pipeline.settings)
+    rescue => e
+      increment_reload_failures_metrics(id, new_config, e)
+      return
+    end
+
+    # check if the new pipeline will be reloadable in which case we want to log that as an error and abort
+    if !pipeline_validator.reloadable?
+      @logger.error(I18n.t("logstash.agent.non_reloadable_config_reload"), :pipeline_id => id, :plugins => pipeline_validator.non_reloadable_plugins.map(&:class))
+      # TODO: in the original code the failure metrics were not incremented, should we do it here?
```
colinsurprenant: yeah - I wanted to avoid changing the behaviour since all tests are passing but I can add that  and a test for it I guess.

logstash-core/lib/logstash/agent.rb
```diff
@@ -232,62 +229,126 @@ def collect_metrics?
...
-    return if new_pipeline.nil?
+    # first stop the old pipeline.
+    # IMPORTANT: a pipelines with same settings should not be instantiated before the previous one is shutdown
+    stop_pipeline(pipeline_id)
+    reset_pipeline_metrics(pipeline_id)
 
+    # start new pipeline now that the old one is shutdown
+    new_pipeline = create_pipeline(settings, new_config)
+    if new_pipeline.nil?
+      # this is a scenario where the configuration is valid (compilable) but the new pipeline refused to start
```
colinsurprenant: ah yeah, it was more a comment for the following section below. will clarify.

logstash-core/lib/logstash/agent.rb
colinsurprenant: actually - I think I'd prefer to leave it as is since it will change reporting behaviour, it is not tested for and it is now created to handle exception this is not the result of a exception. I suggest we create a separate issue for this.

colinsurprenant:
@jsvd Thanks. I pushed a commit to address your suggestions. Let me know.

logstash-core/lib/logstash/agent.rb
```diff
@@ -232,62 +229,129 @@ def collect_metrics?
...
+  # @param settings [Settings] the setting for the new pipelines
+  # @param config [String] the configuration string or nil to fetch the configuration per settings
+  def create_pipeline(settings, config = nil)
     if config.nil?
       begin
         config = fetch_config(settings)
       rescue => e
         @logger.error("failed to fetch pipeline configuration", :message => e.message)
-        return
+        return nil
```
ph: explicit return!

logstash-core/lib/logstash/agent.rb
```diff
@@ -232,62 +229,129 @@ def collect_metrics?
...
+
+    # second step create and start a new pipeline now that the old one is shutdown
+
+    new_pipeline = create_pipeline(settings, new_config)
+    if new_pipeline.nil?
+      # this is a scenario where the configuration is valid (compilable) but the new pipeline refused to start
+      # and at this point NO pipeline is running
+      @logger.error("failed to create the reloaded pipeline and no pipeline is currently running", :pipeline => pipeline_id)
+      return
+    end
```
ph: This change make sense here, I've did the same change in behavior in my refactor because of the simplicity,

logstash-core/lib/logstash/agent.rb
```diff
@@ -232,62 +229,129 @@ def collect_metrics?
...
       return
     end
 
-    new_pipeline = create_pipeline(old_pipeline.settings, new_config)
+    # check if the new pipeline will be reloadable in which case we want to log that as an error and abort
+    if !pipeline_validator.reloadable?
+      @logger.error(I18n.t("logstash.agent.non_reloadable_config_reload"), :pipeline_id => id, :plugins => pipeline_validator.non_reloadable_plugins.map(&:class))
+      # TODO: in the original code the failure metrics were not incremented, should we do it here?
+      # increment_reload_failures_metrics(id, new_config, e)
+      return
```
ph: I think we should increment reload failure, because the current state is not consistent to what we are expecting.

logstash-core/lib/logstash/agent.rb
jsvd: fair enough, +1 on creating an issue to track it

jsvd:
functionality LGTM, testing several corner cases revealed no bugs, both on the pipeline reload side and the metrics side.

I'd still like to remove some of the code style changes, like https://github.com/elastic/logstash/pull/6631/files#r99300484 and a few explicit `return nil` (not consistent with the rest of the code)

logstash-core/lib/logstash/agent.rb
colinsurprenant: here's what I said to @jsvd same comment: «I think I'd prefer to leave it as is since it will change reporting behaviour, it is not tested for and it is now created to handle exception this is not the result of a exception. I suggest we create a separate issue for this.»

ph: I will raise that issue again with the refactor I did, we can decide when it land.

colinsurprenant:
@jsvd  thanks. the explicit `return nil` is absolutely not inconsistent and I will leave it as is. all the methods where a bare `return` is done is because these methods don't return a usable value. the `create_pipeline` methods returns either a `Pipeline` object or `nil` if it fails. This makes is explicit.  I added a comment to document the return value.

logstash-core/lib/logstash/agent.rb
colinsurprenant: see comment below

colinsurprenant:
Also, it's funny, I was expecting to send the crowd into a frenzy, all cheering for my changes since the previous formatting was so painful to look at. :man_shrugging:

ph:
@colinsurprenant well, I am still stuck in a 120 chars wide, editor.

colinsurprenant:
@ph time to upgrade. 2017.

logstash-core/lib/logstash/agent.rb
```diff
@@ -232,62 +229,130 @@ def collect_metrics?
...
+      n.increment(:failures)
+      n.gauge(:last_error, { :message => exception.message, :backtrace => exception.backtrace})
+      n.gauge(:last_failure_timestamp, LogStash::Timestamp.now)
+    end
+    if @logger.debug?
+      @logger.error("fetched an invalid config", :config => config, :reason => exception.message, :backtrace => exception.backtrace)
+    else
+      @logger.error("fetched an invalid config", :config => config, :reason => exception.message)
+    end
+  end
```
ph: +1 I did something similar.

logstash-core/lib/logstash/agent.rb
```diff
@@ -232,62 +229,130 @@ def collect_metrics?
...
+    # check if the new pipeline will be reloadable in which case we want to log that as an error and abort
+    if !pipeline_validator.reloadable?
+      @logger.error(I18n.t("logstash.agent.non_reloadable_config_reload"), :pipeline_id => id, :plugins => pipeline_validator.non_reloadable_plugins.map(&:class))
+      # TODO: in the original code the failure metrics were not incremented, should we do it here?
+      # increment_reload_failures_metrics(id, new_config, e)
+      return
+    end
 
-    return if new_pipeline.nil?
+    # we know configis valid so we are fairly comfortable to first stop old pipeline and then start new one
```
ph: missing space between *config* and *is*.

colinsurprenant:
thanks @jsvd and @ph - rebased + squashed, merged in master and 5.x

Issue #6620, opened 2017-02-01
Refactor Agent to be more flexible with Source Loading


ph:
I've been working on the refactoring of the **Config Loading**  in https://github.com/elastic/logstash/pull/6514 to give the ability of a universal plugins to add or remove sources of configuration.

So far this is working fine when we only deal with one pipeline and the auto-reload is off, as soon as we deal with multiple pipeline like the xpack monitoring pipeline this get a bit tricky. 

Let's say that the source of configuration is dynamic and the following happen:

- First call return pipeline1
- Second call return **no pipeline**

I would expect Logstash in this case to stop the **pipeline1** and wait until the dynamic source returns a new pipeline to run, this will even be more needed when multiples pipelines land and the dynamic source return 1 or more pipelines.

We need to do the following changes:

SOURCE_LOADER.fetch will return a **array of pipeline_configs**, the agent will take this list as the source of truth and will diff the current running pipeline into what we actually need to run and create a list of actions that the agent needs to execute to sync the running state.

Defined actions:

- Create Pipeline
- Reload pipeline
- Stop Pipeline

# TODO:
- [x] resolve action
- [x] remove register_pipeline
- [x] agent takes the source loader as a parameter
- [x] Split Action
- [x] Unit tests Actions.

ph:
cc @jsvd

ph:
The strategy I am intending to use is to create simple classes for each **action** that will receive the hash of the pipelines, this will allow us to test each scenario in isolation.

Conditions:

- If the pipeline_config.pipeline_id is not in the pipelines => Create
- If the pipeline_config.pipeline_id is in the pipelines id
  - the **config_hash** don't match => Reload
  - the **config_hash** match => NOOP
- If the any of the current running pipeline_id is not in the received list of pipeline_configs => Stop

ph:
This change will remove the need of `register_pipeline` since the PipelineConfig will provide both the actual config and the settings that we need to use.

ph:
Running 1 or multiple pipeline will use the same code and will have some kind of notion of state this will allow us to add a health status on logstash when multiple pipeline land.

ph:
@jsvd I've been digging through the auto reload code, and I've discovered that we can create `pipeline` which are not auto-reloadable with the pipeline specific settings? 

I thought reloading was a top level concern, in the context of multiple pipelines and dynamic sources do we need that, I am just trying to figure out if we need it or not?

I can certainally skip them when I resolve the converge actions.

[...]

PR #6632, opened 2017-02-02
Refactor of the Agent and the source loading

ph:
**Note for reviewers:**
This PR includes both the changes of the source loading to make it a bit more flexible and to allow universal plugin to add new source.

ref: https://github.com/elastic/logstash/pull/6514  and https://github.com/elastic/logstash/issues/6620

There is a lot a new code for the following reasons:

1. It includes the code for the source loder.
2. it includes a refactor of the `agent_spec` test.
---

Our current implementation of the Agent in Logstash does a bit too much:

1. Start the webserver
2. Start the metrics
3. Fetch the configs
4. start pipeline
5. stop pipeline
6. reload pipeline
7. Metrics and action code are really linked

This PR attempt to clean up a bit of the execution model and how things are linked together, (3 -> 7)

In this code, the agent never interact directly with the configuration file, instead it uses a `SourceLoader` class. For the Agent, the **SourceLoader** is the source of truth.
if its not returned by the loader, the agent should not run it.

### Testing Strategy

I tried to do minimum changes to the current `agent_spec` file, but I've decided to create 2 other spec with using less mocking and threading.

### Scenario with auto-reload on

- Logstash start
- The Agent ask the **SourceLoader** to return the pipelines: `[:main, :heavy_work, :foobar]`
- If the fetch is successful it continue, otherwise it wait next iteration
- The Agent check the current running pipelines: `[]`
- The Agent ask the `StateResolver` how to converge to the required pipelines, the following actions are scheduled:
    - Start **:main**
    - Start **:heavy_work**
    - Start **:foobar**
- The agent execute the actions
- The agent records the stats with the result of the actions
- The Agent sleep for the reload interval
- The Agent ask the **SourceLoader** to return the pipelines: `[:heavy_work(config_hash changed), :foobar]`
- If the fetch is successful it continue
- The Agent check the current running pipelines: `[:main, :heavy_work, :foobar]`
- The Agent ask the `StateResolver` how to converge to the required pipelines, the following actions are scheduled:
    - Stop **:main**
    - Reload **:heavy_work** (config_hash don't match)
- The agent execute the actions
- The agent records the stats with the result of the actions
- The Agent sleep for the reload interval

### A normal scenario without auto reload

- Logstash start
- The Agent ask the **SourceLoader** to return the pipelines: `[:main, :heavy_work, :foobar]`
- If the fetch is successful it continue, otherwise it wait next iteration
- The Agent check the current running pipelines: `[]`
- The Agent ask the `StateResolver` how to converge to the required pipelines, the following actions are scheduled:
    - Start **:main**
    - Start **:heavy_work**
    - Start **:foobar**
- The agent execute the actions
- The agent records the stats with the result of the actions

Both of theses scenario use the same path, the only difference is the refresh call when auto reload is on.

Each for the `Action` are encapsulated into their own class for easier testing and swapping of implementation, the first implementation
target the current ruby pipeline, but this will allow us to experiment with a java pipeline and have a different interface.

This change allow us to test without using any threads in the `agent_spec` or in the `action`.


logstash-core/lib/logstash/errors.rb
```diff
@@ -6,7 +6,12 @@ class ConfigurationError < Error; end
...
   class ShutdownSignal < StandardError; end
   class PluginNoVersionError < Error; end
+  class BootstrapCheckError < Error; end
 
   class Bug < Error; end
   class ThisMethodWasRemoved < Bug; end
+  class ConfigLoadingError < Error; end
+  class InvalidSourceLoaderSettingError < Error; end
+  class PipelineActionError < Error; end
+  class NonReloadablePipelineError < StandardError; end
```
ph: I will clean these errors, I do not use them anymore.

logstash-core/spec/logstash/config/source/local_spec.rb
```diff
@@ -0,0 +1,402 @@
+# encoding: utf-8
```
ph: This could be split into smaller file maybe?

ph:
Also for the reviewer pay attention to the `converge_spec.rb` which show how to use the new source loader to manage the tests.

ph:
I see already one failure for the `agent_spec`, I am pretty sure its a timing issue :(

ph:
I've pushed a fix to help with the flaky test.

logstash-core/lib/logstash/agent.rb
```diff
@@ -204,6 +250,72 @@ def close_pipelines
...
+      #  - FailedAction
+      #  - SuccessfulAction
+      #  - Exception
+      #
+      # This give us a bit more extensibility with the current startup/validation model
+      # that we currently have.
+      begin
+        logger.debug("Executing action", :action => action)
+        action_result = action.execute(@pipelines)
+        converge_result.add(action, action_result)
```
ph: I could probably add a logger statement here when the `action fail`. I have mixed feeling if we should go with `debug` or `error`,

jsvd:
I tried running a simple pipeline and got an error:

```
 % bin/logstash -e ""
Sending Logstash's logs to /Users/joaoduarte/projects/logstash/logs which is now configured via log4j2.properties
[2017-03-13T05:21:40,353][ERROR][logstash.agent           ] An exception happened when converging configuration {:exception=>LogStash::InvalidSourceLoaderSettingError, :message=>"Can't find an appropriate config loader with current settings", :backtrace=>["/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/config/source_loader.rb:52:in `fetch'", "/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/agent.rb:133:in `converge_state_and_update'", "/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/agent.rb:84:in `execute'", "/Users/joaoduarte/projects/logstash/logstash-core/lib/logstash/runner.rb:297:in `execute'", "org/jruby/RubyProc.java:281:in `call'", "/Users/joaoduarte/projects/logstash/vendor/bundle/jruby/1.9/gems/stud-0.0.22/lib/stud/task.rb:24:in `initialize'"]}
```

jsvd:
I tested how auto reload reacts to a broken configuration, and the logging isn't very useful:

I started logstash with a good configuration:

```
% cat cfg
input {
  generator {}
}
filter {
  sleep { time => 1 }
}
output { stdout { codec => plain { format => "." } } }
```

and after a few events I changed it to:

```
% cat cfg
input {
  generator {}
}
filter {
  sleep { time => 1 }
}
output { stdout { codec => plain { format => "." } } }
broken thing
```

And the output is:

```
% bin/logstash -f cfg -w 1 -b 1 -r
[2017-03-13T05:31:09,313][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2017-03-13T05:31:09,331][INFO ][logstash.pipeline        ] Starting pipeline {"id"=>"main", "pipeline.workers"=>1, "pipeline.batch.size"=>1, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>1}
[2017-03-13T05:31:09,343][INFO ][logstash.pipeline        ] Pipeline main started
[2017-03-13T05:31:09,350][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
................................[2017-03-13T05:31:42,382][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
...[2017-03-13T05:31:45,365][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
...[2017-03-13T05:31:48,369][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
...[2017-03-13T05:31:51,368][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
...[2017-03-13T05:31:54,368][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
```

jsvd:
Adding a comment to the end of the config breaks the loading of the pipeline:

```
% cat /tmp/cfg
input { generator {} }
filter { sleep { time => 1 } }
output { stdout { codec => plain { format => "." } } }
# hello
```

on logstash 5.2.1:

```
/tmp/logstash-5.2.1 % bin/logstash -f /tmp/cfg -w 1 -b 1
Sending Logstash's logs to /tmp/logstash-5.2.1/logs which is now configured via log4j2.properties
[2017-03-13T05:42:12,877][INFO ][logstash.pipeline        ] Starting pipeline {"id"=>"main", "pipeline.workers"=>1, "pipeline.batch.size"=>1, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>1}
[2017-03-13T05:42:12,884][INFO ][logstash.pipeline        ] Pipeline main started
[2017-03-13T05:42:12,983][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
..........
```

on this PR:

```
~/projects/logstash (git)-[pr/6632] % bin/logstash -f /tmp/cfg -w 1 -b 1
Sending Logstash's logs to /Users/joaoduarte/projects/logstash/logs which is now configured via log4j2.properties
[2017-03-13T05:42:33,286][ERROR][logstash.agent           ] Failed to execute action {:action=>LogStash::PipelineAction::Create/pipeline_id:main, :exception=>"LogStash::ConfigurationError", :message=>"Expected one of \r, \n at line 4, column 8 (byte 117) after # hello"}
[2017-03-13T05:42:33,301][INFO ][logstash.agent           ] Pipelines running {:count=>0, :pipelines=>[]}
```

jsvd:
reload metrics seem to be broken:
```
curl localhost:9600/_node/stats  | jq ".pipeline.reloads"
null
```

while on 5.2.1:

```
/tmp/logstash-5.2.1 % curl -s localhost:9600/_node/stats  | jq ".pipeline.reloads"
{
  "last_error": null,
  "successes": 0,
  "last_success_timestamp": null,
  "last_failure_timestamp": null,
  "failures": 0
}
```

jsvd:
It seems most pipeline actions are assuming the main pipeline id instead of the configured one. I changed the `pipeline.id: "meh"` in the config file, and in debug mode I see:

```
[2017-03-13T06:06:31,596][DEBUG][logstash.agent           ] Executing action {:action=>LogStash::PipelineAction::Create/pipeline_id:main}
# ....
[2017-03-13T06:06:31,669][DEBUG][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"meh"}
[2017-03-13T06:06:31,676][INFO ][logstash.pipeline        ] Starting pipeline {"id"=>"meh", "pipeline.workers"=>1, "pipeline.batch.size"=>1, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>1}
[2017-03-13T06:06:31,685][INFO ][logstash.pipeline        ] Pipeline meh started
[2017-03-13T06:06:31,696][DEBUG][logstash.pipeline        ] Pipeline started succesfully {:pipeline_id=>"meh"}
[2017-03-13T06:06:31,702][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["meh"]}
```

Same during shutdown:

```
[2017-03-13T06:07:16,325][DEBUG][logstash.agent           ] Shutting down all pipelines {:pipelines_count=>1}
[2017-03-13T06:07:16,325][DEBUG][logstash.agent           ] Converging pipelines
[2017-03-13T06:07:16,325][DEBUG][logstash.agent           ] Needed actions to converge {:actions_count=>1}
[2017-03-13T06:07:16,326][DEBUG][logstash.agent           ] Executing action {:action=>LogStash::PipelineAction::Stop/pipeline_id:main}
[2017-03-13T06:07:16,327][DEBUG][logstash.pipeline        ] Closing inputs
[2017-03-13T06:07:16,327][DEBUG][logstash.inputs.generator] stopping {:plugin=>"LogStash::Inputs::Generator"}
[2017-03-13T06:07:16,327][DEBUG][logstash.pipeline        ] Closed inputs
[2017-03-13T06:07:16,327][DEBUG][logstash.pipeline        ] Closing inputs
.[2017-03-13T06:07:17,127][DEBUG][logstash.inputs.generator] closing {:plugin=>"LogStash::Inputs::Generator"}
[2017-03-13T06:07:17,128][DEBUG][logstash.pipeline        ] Input plugins stopped! Will shutdown filter/output workers.
[2017-03-13T06:07:18,149][DEBUG][logstash.filters.sleep   ] closing {:plugin=>"LogStash::Filters::Sleep"}
[2017-03-13T06:07:18,149][DEBUG][logstash.outputs.stdout  ] closing {:plugin=>"LogStash::Outputs::Stdout"}
[2017-03-13T06:07:18,149][DEBUG][logstash.pipeline        ] Pipeline meh has been shutdown
```

jsvd:
it seems to be possible to start a pipeline that is not reloadable with -r

```
% cat myconfs/*
input { generator { } }
input { stdin {} }
filter { sleep { time => 1 } }
output { stdout { codec => plain { format => "." } } }
output { stdout { codec => rubydebug } }
```

in logstash 5.2.1:

```
/tmp/logstash-5.2.1 % bin/logstash -f "~/projects/logstash/myconfs/*" -r
[2017-03-13T06:42:31,378][ERROR][logstash.agent           ] Logstash is not able to start since configuration auto reloading was enabled but the configuration contains plugins that don't support it. Quitting... {:pipeline_id=>"main", :plugins=>[LogStash::Inputs::Stdin]}
```

in this PR:

```
 ~/projects/logstash (git)-[pr/6632] % bin/logstash -f "myconfs/*" -r
[2017-03-13T06:43:00,570][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2017-03-13T06:43:00,726][INFO ][logstash.pipeline        ] Starting pipeline {"id"=>"meh", "pipeline.workers"=>4, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>500}
[2017-03-13T06:43:00,748][INFO ][logstash.pipeline        ] Pipeline meh started
The stdin plugin is now waiting for input:
```

jsvd:
the --config.debug flag is now ignored even when using --log.level=debug, which means there is no way to see the configuration that logstash is loading.
I tried with both `-t --log.level=debug --config.debug` and without `-t`

Example run [in this gist](https://gist.github.com/jsvd/efe8db4c5256c5a46535257113ae9ac5)

ph:
Thanks for catching all theses issues, I will check the suite to see if are covering them.
I will fix them an update the update the PR, I don't see anything major.

ph:
@jsvd 

Concerning the reload metrics they seems to work on my ends

```json
{
  "last_error": null,
  "successes": 0,
  "last_success_timestamp": null,
  "last_failure_timestamp": null,
  "failures": 0
}
```

```json
{
  "last_error": null,
  "successes": 1,
  "last_success_timestamp": "2017-03-14T21:37:27.788Z",
  "last_failure_timestamp": null,
  "failures": 0
}```

**Comment by ph** [comment]
> It seems most pipeline actions are assuming the main pipeline id instead of the configured one. I changed the pipeline.id: "meh" in the config file, and in debug mode I see:

For that, until we have the changes for the `logstash.yml` to configure multiple pipeline I think we should keep that? WDYT?

**Comment by ph** [comment]
There is a bug with the auto reload, I am still looking into it, I've pushed the others changes

**Comment by ph** [comment]
I've fixed the issue with auto reload and also added more debug information when an action did not successfully run.

**Comment by ph** [comment]
@jsvd Concerning the *auto-reload* and *non reloadable* plugin, I believe we might need change in behavior. The current behavior (5.x) is to refuse to start (die) when a configuration has some non-reloadable plugins. The problem is that in the case of a remote config this would mean Logstash would die without fetching an appropriate config.

So we have two solutions:

1. Either we change the behavior, let logstash start and periodically try to load a configuration that contains will contains reloadable plugins.
2. We ask the source implementation to also create a custom **Create pipeline* action, so we don't kill LogStash.

I think for simplicity 1. is probably a better implementation WDYT?

**Comment by jsvd** [comment]
I'm ok with breaking current behaviour of aborting logstash if reload is enabled and the pipeline is not reloadable.

Also the reload stats bug seems to happen only when the pipeline.id is changed in the yml file

**Comment by ph** [comment]
There seems to be a bug when running multiple pipelines when trying to shutdown logstash making the process hang? I will take a look at that, this was reported on slack by @jsvd

File: logstash-core/lib/logstash/config/source/local.rb [hunk 4]
```diff
@@ -0,0 +1,203 @@
...
+          when 500
+            raise LogStash::ConfigLoadingError, I18n.t("logstash.runner.configuration.fetch-failed", :path => uri.to_s, :message => "500 error on remote host")
+          else
+            raise LogStash::ConfigLoadingError, I18n.t("logstash.runner.configuration.fetch-failed", :path => uri.to_s, :message => "code: #{response.code}, message: #{response.class.to_s}")
+          end
+        end
+      end
+    end
+
+    PIPELINE_ID = LogStash::SETTINGS.get("pipeline.id").to_sym
```
jsvd: since this is evaluated at file loading time, it won't capture changes to `pipeline.id` coming from the yml file, so it's forever `:main`

ph:
I've have a fix locally to fix the problem when trying to shutdown when the processing is slow or almost stuck, I don't really like it but I don't see any other way to make it cleaner.

I have added a `@force_shutdown` in the pipeline that will be set by the `Pipeline#force_shutdown` method, this properties will be check in various methods before we ack a batch, this will allow theses method to exit sooner but could lose data.

I have fixed the other issues that you have found @jsvd, I will do a bit more testing locally there is some log messages during the spec that I want to check.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -200,9 +211,57 @@ def filters?
...
+    @finished_execution = Concurrent::AtomicBoolean.new(false)
+
+    @thread = Thread.new do
+      begin
+        LogStash::Util.set_thread_name("pipeline.#{pipeline_id}")
+        run
+        @finished_execution.make_true
+      rescue => e
+        close
+        logger.error("Pipeline aborted due to error", :exception => e, :backtrace => e.backtrace)
```
ph: I've revisited all the logging statement int **pipeline.rb** and I have added a `#default_logging_keys` methods making it easy to add new values to the logging statements.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -473,10 +532,25 @@ def shutdown(&before_stop)
...
+    stop_inputs
+
+    # We make this call blocking, so we know for sure when the method return the shtudown is
+    # stopped
+    wait_for_workers
+    clear_pipeline_metrics
+  end # def shutdown
+
+  def wait_for_workers
+    @logger.debug("Closing inputs")
```
ph: The biggest problem is we are not carrying any pipeline specific information, I have added that see my previous comment.

ph:
@jsvd I've pushed my changes, I will still keep an eye on the logging I've seen while the spec suite is running. I believe the following is happening, when a test fails, we get a bunch of related locking message on the other tests.

But I don't know for sure.

jsvd:
I just noticed that the :name gauge metric for the plugins isn't being populated/obtained properly.

I used to following pipeline: `bin/logstash -e "input { generator { } } filter { sleep { time => 3 } }" -w 1 -b 1`

on this PR:

```
% curl -s localhost:9600/_node/stats | jq ".pipeline"
{
  "events": {
    "duration_in_millis": 9033,
    "in": 5,
    "queue_push_duration_in_millis": 8725,
    "filtered": 3,
    "out": 3
  },
  "plugins": {
    "inputs": [
      {
        "id": "99ff0cb0ad9fcc7182eb177b93de7e50dc4bed82-2",
        "out": 5,
        "queue_push_duration_in_millis": 8725
      }
    ],
    "filters": [
      {
        "id": "99ff0cb0ad9fcc7182eb177b93de7e50dc4bed82-3",
        "events": {
          "duration_in_millis": 9012,
          "in": 4,
          "out": 3
        }
      }
    ],
    "outputs": [
      {
        "id": "99ff0cb0ad9fcc7182eb177b93de7e50dc4bed82-1",
        "events": {
          "duration_in_millis": 14,
          "in": 3,
          "out": 3
        }
      }
    ]
  },
```

on master:

```
 % curl -s localhost:9600/_node/stats | jq ".pipeline"
{
  "events": {
    "duration_in_millis": 0,
    "in": 2,
    "filtered": 0,
    "out": 0,
    "queue_push_duration_in_millis": 0
  },
  "plugins": {
    "inputs": [
      {
        "id": "cfaba06beefc22561f7d722413e1a0de28e42f22-1",
        "events": {
          "out": 2,
          "queue_push_duration_in_millis": 0
        }
      }
    ],
    "filters": [
      {
        "id": "cfaba06beefc22561f7d722413e1a0de28e42f22-2",
        "events": {
          "in": 1
        },
        "name": "sleep"
      }
    ],
    "outputs": [
      {
        "id": "cfaba06beefc22561f7d722413e1a0de28e42f22-3",
        "name": "stdout"
      }
    ]
  }
```

weirdly enough, on logstash 5.2.1 we don't show the input at all:

```
% curl -s localhost:9600/_node/stats | jq ".pipeline"
{
  "events": {
    "duration_in_millis": 3009,
    "in": 2,
    "filtered": 1,
    "out": 1
  },
  "plugins": {
    "inputs": [],
    "filters": [
      {
        "id": "cfaba06beefc22561f7d722413e1a0de28e42f22-2",
        "events": {
          "duration_in_millis": 3002,
          "in": 2,
          "out": 1
        },
        "name": "sleep"
      }
    ],
    "outputs": [
      {
        "id": "cfaba06beefc22561f7d722413e1a0de28e42f22-3",
        "events": {
          "duration_in_millis": 4,
          "in": 1,
          "out": 1
        },
        "name": "stdout"
      }
    ]
  }
```

ph:
I will check for the `name`, the missing `event` was fixed in https://github.com/elastic/logstash/commit/b1d2b8a39b87436a70a58d350691699d64a05f80

ph:
@jsvd This is a good catch, I didn't change anything in the execution path related to the collection of metrics, I believe this could be related to a thread safety issue but I am not sure yet. I was able to get the :name by adding breakpoints. This will take some time to find, will report asap.

ph:
I've pushed the fix for the missing `names`, the problem wasn't a thread safety issue. I was clearing out the stats before starting a new pipeline. So instead of doing it at the start I only do it on shutdown. This change make sure the metrics are consistent and are not modified. (I will make a PR on master for the missing input name)

Also, this illustrate a weird problem that we have, plugins get initialized when the pipeline instance is created :(

ph:
Looking at the integration test, this is a legit failures.

ph:
@guyboertje Would you mind reviewing this PR, myself or @jsvd can sync with you to get you started!

logstash-core/lib/logstash/state_resolver.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      pipeline_configs.each do |pipeline_config|
+        pipeline = pipelines[pipeline_config.pipeline_id]
+
+        if pipeline.nil?
+          actions << LogStash::PipelineAction::Create.new(pipeline_config, @metric)
+        else
+          # TODO(ph): The pipeline should keep a reference to the original PipelineConfig
+          # and we could use straight comparison.
+          if pipeline_config.config_hash != pipeline.config_hash
```
jsvd: may I suggest abstracting this equivalence? my main reason is that we may decide to have more than 1 reason to consider two PipelineConfig objects equivalent (or not equivalent).
For example, we can consider differences in setting values like pipeline.workers enough to trigger a reload.

logstash-core/lib/logstash/bootstrap_check/default_config.rb
```diff
@@ -0,0 +1,17 @@
+# encoding: utf-8
+require "logstash/errors"
+
+module LogStash module BootstrapCheck
+  class DefaultConfig
+    def self.check(settings)
+      if settings.get("config.string").nil? && settings.get("path.config").nil?
+        raise LogStash::BootstrapCheckError, I18n.t("logstash.runner.missing-configuration")
```
jsvd: we still support the `bin/logstash -i irb/pry` mode, and in this scenario we're going to receive:

```
% bin/logstash -i irb
ERROR: No configuration file was specified. Perhaps you forgot to provide the '-f yourlogstash.conf' flag?
usage:
  bin/logstash -f CONFIG_PATH [-t] [-r] [] [-w COUNT] [-l LOG]
  bin/logstash -e CONFIG_STR [-t] [--log.level fatal|error|warn|info|debug|trace] [-w COUNT] [-l LOG]
  bin/logstash -i SHELL [--log.level fatal|error|warn|info|debug|trace]
  bin/logstash -V [--log.level fatal|error|warn|info|debug|trace]
  bin/logstash --help
```

jsvd:
One thing I've notice while playing with multiple pipelines is that there's no protection at this layer if a source returns more than 1 pipeline with the same pipeline.id
And since most actions done to pipelines are based on this id it's very easy to break other pipelines.

What do you think of failing (and logging) a fetch operation if there's there are duplicated ids in the retrieved PipelineConfig objects?

logstash-core/lib/logstash/state_resolver.rb
ph: I will make an issue to do a followup PR.

logstash-core/lib/logstash/bootstrap_check/default_config.rb
ph: Oh thats a bug! Will make the quick changes. :)

ph:
@jsvd Correct, I think we should bit more strict with the uniqueness of the **pipeline_ids**, I've created this https://github.com/elastic/logstash/issues/6858, Can you give your input?

logstash-core/lib/logstash/state_resolver.rb
ph: @jsvd At first, I was thinking about implementing a comparator on the `PIpelineConfig` object and use that here by comparing the two `pipeline_config`, In that case we can check the string and the settings. WDYT?

ph: Created https://github.com/elastic/logstash/issues/6859

logstash-core/lib/logstash/bootstrap_check/default_config.rb
ph: Actually, there is trace in the test suite for that, I've missed it.

logstash-core/lib/logstash/pipeline_action/base.rb
```diff
@@ -0,0 +1,24 @@
...
+      "#{self.class.name}/pipeline_id:#{pipeline_id}"
+    end
+    alias_method :to_s, :inspect
+
+    def execute(pipelines)
+      raise "`#execute` Not implemented!"
+    end
+
+    def <=>(other)
+      (ORDERING.index(self.class) <=> ORDERING.index(other.class)).nonzero? || self.pipeline_id <=> other.pipeline_id
```
guyboertje: I would prefer to return -1, 0, 1. e.g.
```ruby
(ORDERING.index(self.class) <=> ORDERING.index(other.class)).nonzero? ? self.pipeline_id <=> other.pipeline_id : 0
```

logstash-core/lib/logstash/bootstrap_check/default_config.rb
suyograo: I think using `pipeline.yml` is an overkill, IMO. There should be one file to configure Logstash. In production, users would just use the file source(s). I understand the debate is about reloadable configurations from the settings file, but that is out of scope for this PR.

logstash-core/lib/logstash/pipeline_action/base.rb
ph: Actually, I don't need to check the `pipeline_id`, since we cannot execute 2 differents action for the same `pipeline_id`

logstash-core/lib/logstash/bootstrap_check/bad_java.rb
```diff
@@ -0,0 +1,17 @@
...
+require "logstash/errors"
+
+module LogStash module BootstrapCheck
+  class BadJava
+    def self.check(settings)
+      # Exit on bad java versions
+      java_version = LogStash::Util::JavaVersion.version
+
+      if LogStash::Util::JavaVersion.bad_java_version?(java_version)
+        raise LogStash::BootstrapCheckError, "Java version 1.8.0 or later is required. (You are running: #{java_version})"
```
ph: I've moved the actual code that raise the exection inside the JavaVersion module and call instead `validate_java_version!`, this make all the logic encapsulated in the same file and the bootstrap_check is just a thin wrapper around that.

ph:
@guyboertje @jsvd I've pushed the changes and rebased the PR with master.

jsvd:
are the integration ci failures related to the bug you saw with sudo?

ph:
@jsvd Not at all, I think they are in part related to #6754,  I am stress testing them locally on a linux vm. For now, I don't think we have a big problem in this PR. Yesterday, the PR was green then I've pushed some small changes and it when back to red, the changes shouldn't have any effect on the failling tests.

jsvd:
LGTM, rebase and merge 🤗 🚢 🏆

ph:
As a note, I am able to get some of the integration failures locally on a linux container.

ph:
@jsvd I take my last comment back, one of the integration test is failing for a good reason.

```
  2) Test Logstash instance should still merge when -e is specified and -f has no valid config files
     Failure/Error: expect(is_port_open?(port1)).to be true

       expected true
            got false
     # ./specs/01_logstash_bin_smoke_spec.rb:137:in `(root)'
     # ./vendor/jruby/1.9/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # ./vendor/jruby/1.9/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # ./vendor/jruby/1.9/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # ./vendor/jruby/1.9/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # ./vendor/jruby/1.9/gems/logstash-devutils-1.3.1-java/lib/logstash/devutils/rspec/logstash_helpers.rb:18:in `try'
     # ./specs/01_logstash_bin_smoke_spec.rb:136:in `(root)'
     # ./vendor/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
````

This behavior is a bit strange if you take 5.X, (/tmp/foobar doesn't exist)

```
bin/logstash -f /tmp/foobar # will fait to start, complaining it cannot find files.
bin/logstash -e 'input { generator {}} output { null {}}' -f /tmp/hloa # will complain about it cannot find file but will continue.
```

The behavior in this branch is: You provided settings that don't work and we fails.

Since we target 6.0, WDYT? I more into make it fails, but I wonder if there any negative side effect.


What's strange, I was to make this test work in some run. :(

**Comment by jsvd** [comment]
I think it's pretty easy to say that it should fail in the case of `bin/logstash -e "" -f /tmp/absent/file`
However, in the scenario of `bin/logstash -e "" -f "/tmp/files/*.cfg"` it might be ok if there are no files whatsoever.

I'm ok with failing if there are no files even if -e is present 👍

**Comment by suyograo** [comment]
> I'm ok with failing if there are no files even if -e is present

@jsvd @ph that's not gonna work when we install via packages. We've seen in the past that users install LS via deb/rpm and proceed to start it from the binary install location (and not using service scripts). When this happens, `-f` is implicit because we point to `/etc/logstash/conf`. So anyone using `bin/logstash -e "blah"` will implicitly also be doing `bin/logstash -e "blah" -f /etc/logstash/conf`. Hence, we can't fail if `/etc/logstash/conf*.cfg` does not return any results.

```
bin/logstash -e 'input { generator {}} output { null {}}' -f /tmp/hloa # will complain about it cannot find file but will continue.`
```
This test encapsulated the behavior I described ^^

suyograo:
Can you check what's the behavior if `/tmp/hloa` directory is present, but there are no files underneath? If this works, we can modify the test to create `/tmp/hloa`

ph:
I've fixed the -e -f, in https://github.com/elastic/logstash/pull/6632/commits/a8b430702e01e2eb801df4cc52a2178e58336083

ph:
I've rebased this monstrous PR, I have run the rats test on my linux box over the weekend and seems to only get the intermittent failures that you already working on fixing @jsvd.

jordansissel:
@ph and @jsvd walked me through the design and a light stepping through the patch. Design LGTM.

PR #6645, opened 2017-02-06
make SafeURI class clone deeper

jsvd:


andrewvc:
Mostly LGTM, but WDYT about moving the specs here: https://github.com/elastic/logstash/blob/ceb5103e01564cba0fb26314f22cebd3ed393cb0/logstash-core/spec/logstash/config/mixin_spec.rb#L181 ?

andrewvc:
Actually, I take that back, this is LGTM. We should actually move some of those specs to your new spec file.

PR #6648, opened 2017-02-07
Generate lock file for 5.3

suyograo:


ph:
LGTM, I've double checked the elasticsearch output, we will need to make a PR for 5.2.1 with that updated too.

PR #6650, opened 2017-02-07
release notes for 5.2.1

ph:

PR #6651, opened 2017-02-07
Update lockfile for 5.2.1

ph:
I've skipped the Kv 4.0.0, after looking at the changelog we break some options in that major bump.
linked to #6650


jsvd:
LGTM, Pending on ci..

PR #6656, opened 2017-02-07
PQ exclusive access to path.queue using file locking

### LINKED ISSUE #6604: add PQ exclusive access to path.queue using file locking

colinsurprenant:
This fixes #6604 by implementing exclusive access to the queue at Pipeline initialization. 

The `sample` spec helper in **devutils** `logstash_helpers.rb`does not close the pipeline after instantiating it. I decided to explicitly close the pipelines in all tests using `sample` instead of modifying **devutils** because I am afraid of potential dependency conflicts if we update **devutils** at this point also knowing that it will be refactored into the main source tree soon.

I am not sure if this is the best idea, I am open to suggestions.

#### TODO
- [x] fix the pipeline commented out specs for the multi pipelines cases.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -189,7 +189,22 @@ class LogStash::Agent
     end
   end
 
+  def close_pipeline(id)
+    pipeline = @pipelines[id]
+    if pipeline
+      @logger.warn("closing pipeline", :id => id)
+      pipeline.close
+    end
+  end
+
+  def close_pipelines
+    @pipelines.each  do |id, _|
+      close_pipeline(id)
+    end
+  end
+
   private
+
   def start_webserver
     options = {:http_host => @http_host, :http_ports => @http_port, :http_environment => @http_environment }
     @webserver = LogStash::WebServer.new(@logger, self, options)
@@ -229,17 +244,17 @@ class LogStash::Agent
     @collect_metric
   end
 
-  def increment_reload_failures_metrics(id, config, exception)
+  def increment_reload_failures_metrics(id, message, backtrace = nil)
     @instance_reload_metric.increment(:failures)
     @pipeline_reload_metric.namespace([id.to_sym, :reloads]).tap do |n|
       n.increment(:failures)
-      n.gauge(:last_error, { :message => exception.message, :backtrace => exception.backtrace})
+      n.gauge(:last_error, { :message => message, :backtrace =>backtrace})
       n.gauge(:last_failure_timestamp, LogStash::Timestamp.now)
     end
     if @logger.debug?
-      @logger.error("Cannot load an invalid configuration.", :reason => exception.message, :backtrace => exception.backtrace)
+      @logger.error("Cannot load an invalid configuration", :reason => message, :backtrace => backtrace)
     else
-      @logger.error("Cannot load an invalid configuration.", :reason => exception.message)
+      @logger.error("Cannot load an invalid configuration", :reason => message)
     end
   end
 
@@ -261,7 +276,7 @@ class LogStash::Agent
     begin
       LogStash::Pipeline.new(config, settings, metric)
     rescue => e
-      increment_reload_failures_metrics(settings.get("pipeline.id"), config, e)
+      increment_reload_failures_metrics(settings.get("pipeline.id"), e.message, e.backtrace)
       return nil
     end
   end
@@ -294,15 +309,14 @@ class LogStash::Agent
     begin
       pipeline_validator = LogStash::BasePipeline.new(new_config, old_pipeline.settings)
     rescue => e
-      increment_reload_failures_metrics(id, new_config, e)
+      increment_reload_failures_metrics(id, e.message, e.backtrace)
       return
     end
 
     # check if the new pipeline will be reloadable in which case we want to log that as an error and abort
     if !pipeline_validator.reloadable?
       @logger.error(I18n.t("logstash.agent.non_reloadable_config_reload"), :pipeline_id => id, :plugins => pipeline_validator.non_reloadable_plugins.map(&:class))
-      # TODO: in the original code the failure metrics were not incremented, should we do it here?
-      # increment_reload_failures_metrics(id, new_config, e)
+      increment_reload_failures_metrics(id, "non reloadable pipeline")
       return
     end
 
@@ -331,20 +345,28 @@ class LogStash::Agent
       # this is a scenario where the configuration is valid (compilable) but the new pipeline refused to start
       # and at this point NO pipeline is running
       @logger.error("failed to create the reloaded pipeline and no pipeline is currently running", :pipeline => pipeline_id)
+      increment_reload_failures_metrics(pipeline_id, "failed to create the reloaded pipeline")
       return
     end
+
+    ### at this point pipeline#close must be called if upgrade_pipeline does not succeed
 
     # check if the new pipeline will be reloadable in which case we want to log that as an error and abort. this should normally not
     # happen since the check should be done in reload_pipeline! prior to get here.
     if !new_pipeline.reloadable?
       @logger.error(I18n.t("logstash.agent.non_reloadable_config_reload"), :pipeline_id => pipeline_id, :plugins => new_pipeline.non_reloadable_plugins.map(&:class))
+      increment_reload_failures_metrics(pipeline_id, "non reloadable pipeline")
+      new_pipeline.close
       return
     end
 
+    # @pipelines[pipeline_id] must be initialized before #start_pipeline below which uses it
     @pipelines[pipeline_id] = new_pipeline
 
     if !start_pipeline(pipeline_id)
       @logger.error("failed to start the reloaded pipeline and no pipeline is currently running", :pipeline => pipeline_id)
+      # do not call increment_reload_failures_metrics here since #start_pipeline already does it on failure
+      new_pipeline.close
       return
     end
 
@@ -373,6 +395,8 @@ class LogStash::Agent
           n.gauge(:last_failure_timestamp, LogStash::Timestamp.now)
         end
         @logger.error("Pipeline aborted due to error", :exception => e, :backtrace => e.backtrace)
+
+        # TODO: this is weird, why dont we return directly here? any reason we need to enter the while true loop below?!
       end
     end
     while true do
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -26,7 +26,7 @@ module LogStash; class BasePipeline
 
   attr_reader :config_str, :config_hash, :inputs, :filters, :outputs, :pipeline_id
 
-  def initialize(config_str, settings)
+  def initialize(config_str, settings = SETTINGS)
     @logger = self.logger
     @config_str = config_str
     @config_hash = Digest::SHA1.hexdigest(@config_str)
@@ -59,8 +59,7 @@ module LogStash; class BasePipeline
     begin
       eval(config_code)
     rescue => e
-      # TODO: the original code rescue e but does nothing with it, should we re-raise to have original exception details!?
-      raise
+      raise e
     end
   end
 
@@ -139,7 +138,13 @@ module LogStash; class Pipeline < BasePipeline
 
     super(config_str, settings)
 
-    @queue = LogStash::QueueFactory.create(settings)
+    begin
+      @queue = LogStash::QueueFactory.create(settings)
+    rescue => e
+      @logger.error("Logstash failed to create queue", "exception" => e, "backtrace" => e.backtrace)
+      raise e
+    end
+
     @input_queue_client = @queue.write_client
     @filter_queue_client = @queue.read_client
     @signal_queue = Queue.new
@@ -216,14 +221,18 @@ module LogStash; class Pipeline < BasePipeline
     shutdown_flusher
     shutdown_workers
 
-    @filter_queue_client.close
-    @queue.close
+    close
 
     @logger.debug("Pipeline #{@pipeline_id} has been shutdown")
 
     # exit code
     return 0
   end # def run
+
+  def close
+    @filter_queue_client.close
+    @queue.close
+  end
 
   def transition_to_running
     @running.make_true
@@ -327,7 +336,8 @@ module LogStash; class Pipeline < BasePipeline
     # plugin.
     @logger.error("Exception in pipelineworker, the pipeline stopped processing new events, please check your filter configuration and restart Logstash.",
                   "exception" => e, "backtrace" => e.backtrace)
-    raise
+
+    raise e
   end
 
   # Take an array of events and send them to the correct output
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -16,7 +16,7 @@ describe LogStash::Agent do
   let(:config_file) { Stud::Temporary.pathname }
   let(:config_file_txt) { "input { generator { count => 100000 } } output { }" }
 
-    subject { LogStash::Agent.new(agent_settings) }
+  subject { LogStash::Agent.new(agent_settings) }
 
   before :each do
     # This MUST run first, before `subject` is invoked to ensure clean state
@@ -52,6 +52,10 @@ describe LogStash::Agent do
       }
     end
 
+    after(:each) do
+      subject.close_pipelines
+    end
+
     it "should delegate settings to new pipeline" do
       expect(LogStash::Pipeline).to receive(:new) do |arg1, arg2|
         expect(arg1).to eq(config_string)
@@ -75,7 +79,7 @@ describe LogStash::Agent do
     end
   end
 
-    describe "#execute" do
+  describe "#execute" do
     let(:config_file_txt) { "input { generator { count => 100000 } } output { }" }
 
     before :each do
@@ -105,7 +109,10 @@ describe LogStash::Agent do
         it "should not reload_state!" do
           expect(subject).to_not receive(:reload_state!)
           t = Thread.new { subject.execute }
-          sleep 0.1
+
+          # TODO: refactor this. forcing an arbitrary fixed delay for thread concurrency issues is an indication of
+          # a bad test design or missing class functionality.
+          sleep(0.1)
           Stud.stop!(t)
           t.join
           subject.shutdown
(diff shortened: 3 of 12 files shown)


colinsurprenant:
I just realized that my `sample` strategy in the description is not the best idea because the plugins are using sample so **devutils** will need to be updated :(

ph:
> I just realized that my sample strategy in the description is not the best idea because the plugins are using sample so devutils will need to be updated.

Does this mean you will update devutils and revert the explicit close in this PR?

colinsurprenant:
@ph - exactly. working on this now.

colinsurprenant:
pushed elastic/logstash-devutils/pull/67 to solve the `sample` helper issue.

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -138,8 +138,8 @@ class TestPipeline < LogStash::Pipeline
       Thread.abort_on_exception = true
 
       pipeline = LogStash::Pipeline.new(config, pipeline_settings_obj)
-      Thread.new { pipeline.run }
-      sleep 0.1 while !pipeline.ready?
+      t = Thread.new { pipeline.run }
+      sleep(0.1) until pipeline.ready?
```
ph: never though of using until.

logstash-core/lib/logstash/agent.rb
```diff
@@ -373,6 +395,8 @@ def start_pipeline(id)
           n.gauge(:last_failure_timestamp, LogStash::Timestamp.now)
         end
         @logger.error("Pipeline aborted due to error", :exception => e, :backtrace => e.backtrace)
+
+        # TODO: this is weird, why dont we return directly here? any reason we need to enter the while true loop below?!
```
colinsurprenant: 👍

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -327,6 +337,8 @@ def filter_batch(batch)
     # plugin.
     @logger.error("Exception in pipelineworker, the pipeline stopped processing new events, please check your filter configuration and restart Logstash.",
                   "exception" => e, "backtrace" => e.backtrace)
+
+    # TODO: shouldn't that be: "raise e" to re-raise previous exception??
```
jordansissel: `raise` with no arguments will re-raise the previous exception:

```ruby
def oops
  raise "Oh no!"
end

begin
  oops
rescue => e
  raise
end
```

```
% ruby test.rb
test.rb:2:in `oops': Oh no! (RuntimeError)
        from test.rb:6:in `<main>'
```


From the ruby docs on `raise`:
> With no arguments, raises the exception in $! or raises a RuntimeError if $! is nil

logstash-core/spec/logstash/pipeline_spec.rb
colinsurprenant: yeah, it avoids using the negation.

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -314,14 +344,20 @@
     end
 
     after(:each) do
-      subject.shutdown
+      # new pipelines will be created part of the ugrade process so we need
```
jordansissel: typo "upgrade process"

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -339,18 +375,29 @@
...
+
       it "updates the state" do
         subject.send(:"reload_pipeline!", default_pipeline_id)
         expect(subject.pipelines[default_pipeline_id].config_str).to eq(new_config)
       end
+
       it "starts the pipeline" do
-        expect(subject).to receive(:stop_pipeline)
-        expect(subject).to receive(:start_pipeline)
+        expect(subject).to receive(:start_pipeline).and_return(true)
```
jordansissel: This pattern of mocking start_ and stop_pipeline is repeated enough we could consider ways to refactor these tests.

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -425,12 +473,13 @@ class DummyOutput2 < LogStash::Outputs::DroppingDummyOutput; end
       Thread.abort_on_exception = true
 
       @t = Thread.new do
+        Thread.current.abort_on_exception = true
```
jordansissel: Wait, this is a thing?

BTW, if you want a thread you can capture the exception from, you can use Stud::Task which basically wraps Thread and lets you get the exception - http://www.rubydoc.info/gems/stud/Stud/Task

```ruby
@t = Stud::Task.new do ... end
...
result, exception = @t.wait
```

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -425,12 +473,13 @@ class DummyOutput2 < LogStash::Outputs::DroppingDummyOutput; end
       Thread.abort_on_exception = true
 
       @t = Thread.new do
```
jordansissel: I think this ivar should be a `let` instead?

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -440,6 +447,10 @@ class TestPipeline < LogStash::Pipeline
     let(:settings) { LogStash::SETTINGS.clone }
     subject { LogStash::Pipeline.new(config, settings, metric) }
 
+    after :each do
```
jordansissel: This can just be `after do ...` because `after :each` is default behavior of `after` without arguments.

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -528,9 +539,21 @@ class TestPipeline < LogStash::Pipeline
       allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutputmore").and_return(DummyOutputMore)
     end
 
+    # multiple pipelines cannot be instantiated using the same PQ settings, force memory queue
+    before :each do
```
jordansissel: just `before` is ok here, `:each` is implied w/o arguments.

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -528,9 +539,21 @@ class TestPipeline < LogStash::Pipeline
...
+      pipeline_workers_setting = LogStash::SETTINGS.get_setting("queue.type")
```
jordansissel: My intuition is to say that we should mock the `get_settings` call here.

```ruby
allow(LogStash::SETTINGS).to receive(:get_setting).with("queue.type").and_return(...)
```

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -712,7 +758,12 @@ class TestPipeline < LogStash::Pipeline
       allow(LogStash::Plugin).to receive(:lookup).with("filter", "multiline").and_return(LogStash::Filters::Multiline)
       allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(::LogStash::Outputs::DummyOutput)
 
-      Thread.new { subject.run }
+      # subject has to be called for the first time outside the thread because it will create a race condition
+      # with the subject.ready? call since subject is lazily initialized
+      s = subject
+      @t = Thread.new { s.run }
```
jordansissel: Let's avoid ivars in rspec; can we use `let` instead, here?

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -597,6 +613,8 @@ public void close() throws IOException {
                 // unblock blocked writes. a write is blocked *after* the write has been performed so
                 // unblocking is safe and will return from the write call
                 notFull.signalAll();
+
+                FileLockFactory.getDefault().releaseLock(this.dirLock);
```
jordansissel: Should this be in a `finally` clause instead?

logstash-core/src/test/java/org/logstash/ackedqueue/TestQueue.java
```diff
@@ -1,5 +1,6 @@
 package org.logstash.ackedqueue;
 
+import java.io.IOException;
```
jordansissel: unused?

logstash-core/spec/logstash/agent_spec.rb
colinsurprenant: ugrade < upgrade :D

logstash-core/spec/logstash/agent_spec.rb
colinsurprenant: agree. but before refactoring these I would prefer to wait for the Agent refactor @ph is working on and then do a very good refactor of the specs.

logstash-core/spec/logstash/pipeline_spec.rb
colinsurprenant: 👍

logstash-core/spec/logstash/agent_spec.rb
colinsurprenant: 👍

logstash-core/spec/logstash/pipeline_spec.rb
colinsurprenant: 👍

logstash-core/spec/logstash/pipeline_spec.rb
colinsurprenant: would need to test. I basically copied this from another test in this spec

logstash-core/spec/logstash/pipeline_spec.rb
colinsurprenant: will check

logstash-core/lib/logstash/agent.rb
colinsurprenant: will not modify here. I will let that for your refactor.

logstash-core/spec/logstash/agent_spec.rb
colinsurprenant: I am not sure how to properly refactor this, there is also the `@abort_on_exception = Thread.abort_on_exception` and this is all in a `before :each`.  ... 
I'd prefer to defer that into another issue.

logstash-core/spec/logstash/agent_spec.rb
colinsurprenant: actually, I'll remove that - I added it because I has a doubt that an exception was not propagated but it did not change anything.

logstash-core/spec/logstash/pipeline_spec.rb
colinsurprenant: fixed

logstash-core/spec/logstash/agent_spec.rb
colinsurprenant: actually, I fixed the pipeline thread with a `let` but kept the `@abort_on_exception` thing.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
colinsurprenant: agree.

colinsurprenant:
@ph @jordansissel  
- Pushed a fix for the build error - problem was in Agent, in the config test and exit. 
- Pushed review modifications.

colinsurprenant:
The build passed, the only error was an integration test
```
rspec ./specs/01_logstash_bin_smoke_spec.rb:41 # Test Logstash instance multiple of them can be started on the same box with automatically trying different ports for HTTP server
```
I restarted the build, I seems like an intermittent error and not related to these changes.

ph:
@colinsurprenant I've seen a weird behavior, I think it's related to this PR or at least the PQ.

**Reproduction case:**

```yaml
# conf/logstash.yml
config.reload.automatic: true
queue.type: persisted
```

start logstash with the following config.

```
input {
    generator {}
}
output {
    file {
    path => "/tmp/loooo.log"
    }
}
```

Edit the config and add the following ruby filter block that will make it crash.

```
filter {
    ruby {
    code => "raise 'e1'"
    init => "raise 'e2'"
    }
}

```

You should see this in your log:
```
2017-02-09T16:18:04,199][ERROR][logstash.agent           ] failed to start the reloaded pipeline and no pipeline is currently running {:pipeline=>"main"}
[2017-02-09T16:18:04,219][ERROR][logstash.pipeline        ] A plugin had an unrecoverable error. Will restart this plugin.
  Plugin: <LogStash::Inputs::Generator id=>"151a1570ca83cabc17069859fb99b9bf32af4691-1", enable_metric=>true, codec=><LogStash::Codecs::Plain id=>"plain_14b02ee6-39b0-4701-ab4e-ee307426e2c8", enable_metric=>true, charset=>"UTF-8">, threads=>1, message=>"Hello world!", count=>0>
  Error:
[2017-02-09T16:18:05,223][ERROR][logstash.pipeline        ] A plugin had an unrecoverable error. Will restart this plugin.
  Plugin: <LogStash::Inputs::Generator id=>"151a1570ca83cabc17069859fb99b9bf32af4691-1", enable_metric=>true, codec=><LogStash::Codecs::Plain id=>"plain_14b02ee6-39b0-4701-ab4e-ee307426e2c8", enable_metric=>true, charset=>"UTF-8">, threads=>1, message=>"Hello world!", count=>0>
  Error: Attempted to write an event to a closed AckedQueue
[2017-02-09T16:18:06,226][ERROR][logstash.pipeline        ] A plugin had an unrecoverable error. Will restart this plugin.
  Plugin: <LogStash::Inputs::Generator id=>"151a1570ca83cabc17069859fb99b9bf32af4691-1", enable_metric=>true, codec=><LogStash::Codecs::Plain id=>"plain_14b02ee6-39b0-4701-ab4e-ee307426e2c8", enable_metric=>true, charset=>"UTF-8">, threads=>1, message=>"Hello world!", count=>0>
  Error: Attempted to write an event to a closed AckedQueue
[2017-02-09T16:18:07,231][ERROR][logstash.pipeline        ] A plugin had an unrecoverable error. Will restart this plugin.
  Plugin: <LogStash::Inputs::Generator id=>"151a1570ca83cabc17069859fb99b9bf32af4691-1", enable_metric=>true, codec=><LogStash::Codecs::Plain id=>"plain_14b02ee6-39b0-4701-ab4e-ee307426e2c8", enable_metric=>true, charset=>"UTF-8">, threads=>1, message=>"Hello world!", count=>0>
  Error: Attempted to write an event to a closed AckedQueue
```

I the case of the normal memory queue, the generator is probably just blocked on the queue and we are not interrupting it and it just sit there.

Since the PQ will raise an exception when the queue is closed, the pipeline will try to restart the plugin forever, we need check in the pipeline to make sure when are stopped that we don't restart plugins ?

ph:
I was able to run the unit tests of this PR successfully on my machine, the failling test seems to be a flaky one.

ph:
@colinsurprenant @jordansissel I am OK with the current code, the way the `agent_spec` spec is currently structured make it hard to unit theses kind of behavior, so I would limit the change in that file if we can so I can merge what I am doing.

colinsurprenant:
@ph 
> I was able to run the unit tests of this PR successfully on my machine, the failling test seems to be a flaky one.

which tests? the only one failing in the Travis build is an integration test. Are you talking about this integration test? 

> I am OK with the current code, the way the agent_spec spec is currently structured make it hard to unit theses kind of behavior, so I would limit the change in that file if we can so I can merge what I am doing.

- You are OK with current code - what about the «weird behavior» you reported above?
- what changes to `agent_spec` are you referring to?

ph:
> I was able to run the unit tests of this PR successfully on my machine, the failling test seems to be a flaky one.

I say **unit** tests, but now I have also tried the integration and ran green on my linux box.

> I am OK with the current code, the way the agent_spec spec is currently structured make it hard to unit theses kind of behavior, so I would limit the change in that file if we can so I can merge what I am doing.

I am OK with the code changes in the `agent_spec` AKA **without doing too much refactoring**

> You are OK with current code - what about the «weird behavior» you reported above?

That was more of an open question but really malformed. 

But since you ask me to clarify it, since this PR expose this issue when used with config reloading it should also fix the exposed case. 

Which is: **When the queue is closed the input will raise an exception when trying to write to the queue and the pipeline will restart the input**

- Either the pipeline is not aware that it should be stopped?
- Or the restart loop doesn't check if the pipeline is stopped.

ph:
The code at https://github.com/logstash-plugins/logstash-input-generator/blob/master/lib/logstash/inputs/generator.rb#L65 probably meant the pipeline doesn't know that it shoud be closed.

colinsurprenant:
@ph unit test? you mean specs? but then the specs were not flaky, it was systematically failing on the build and it exposed a bug in the Runner which I solved and the specs are now correctly passing. Not sure what you are referring. The only flaky test is the integration test which is in failed state in Travis now and that @suyograo confirmed to be a flaky test failing for other reasons and that will need be fixed.

For the `agent_spec` I never intended to do any unrelated refactoring - was it expected to be otherwise? You said «I would limit the change in that file» but is there too much change or has anyone requested more changes? Sorry I am confused.

I can reproduce your failing scenario. Looking at it.

ph:
> @ph unit test? you mean specs? but then the specs were not flaky, it was systematically failing on the build and it exposed a bug in the Runner which I solved and the specs are now correctly passing. Not sure what you are referring. The only flaky test is the integration test which is in failed state in Travis now and that @suyograo confirmed to be a flaky test failing for other reasons and that will need be fixed.

> For the agent_spec I never intended to do any unrelated refactoring - was it expected to be otherwise? You said «I would limit the change in that file» but is there too much change or has anyone requested more changes? Sorry I am confused.

I think we say and we agree on the same thing here.

ph:
I am approving the changes since it actually restrict access to the PQ to one pipeline.
We have agree to fix the other issue concerning the register in another PR.

colinsurprenant:
With @ph and @suyograo we agreed to solve @ph discovered issue when reloading a pipeline that crashes with an plugins exception in the `register` method. It is somewhat unrelated to PQ but is still critical we solve - A new PR will be better to focus only on this issue which is tricky to solve.

@jordansissel please let me know if you are good with this PR state.

colinsurprenant:
followup issue is #6688

jordansissel:
When I run this, it works as expected, but the error message isn't clear enough. Can we make it more actionable? 

```
[2017-02-10T11:02:38,035][ERROR][logstash.pipeline        ] Pipeline initialization failure at QueueFactory.create {"exception"=>#<IOError: Lock held by another program: /home/jls/projects/logstash/data/queue/main/.lock>
```

Instead of this error, we could say something like "The queue is already in use by another running Logstash instance." ?

colinsurprenant:
@jordansissel we could. but at the same time, "Lock held by another program" is more generic and will be compatible with any other external tools that will also require exclusive access - in which case the lock will not be help by another LS instance but by another program. 

This exception is thrown by `FileLockFactory` and does not know anything about the context of its usage so I would not change the wording in there - we could catch it in `QueueFactory` and re-raise a new exception but then it would have to be something more generic and include the previous exception, like "the queue exclusive access failed: reason: IOError: Lock held by another program: ..." 

WDYT?

jordansissel:
@colinsurprenant I like your idea. My main concern is I want to feel like the error message will direct users at solving their own problem instead of needing to ask on discuss forums or file support tickets, which is part of why suggested a, as you point out, somewhat overly-specific log message ("Another logstash is using the queue").

+1 on catching and reraising a more specific error. We can say "Logstash or another program is using the Queue already. Only one program may use the queue at a time." and give the path to the queue as you have.

colinsurprenant:
@jordansissel  Would this work: (trying to minimize changes just for alternate wording):

```
Logstash failed to create queue {"exception"=>#<IOError: The queue failed to obtain exclusive access, cause: Lock held by another program: /home/jls/projects/logstash/data/queue/main/.lock>
```

colinsurprenant:
rebased+squashed and merged into master and 5.x. thanks @ph and @jordansissel

colinsurprenant:
also merged in 5.3

PR #6657, opened 2017-02-07
pin minitar dep to 0.5.4

talevy:
a new Minitar release was made: https://rubygems.org/gems/minitar/versions/0.6

It is breaking the build:

```
[bootstrap] Fetching and installing gem: minitar (>= 0)
The `minitar` executable is no longer bundled with `minitar`. If you are
expecting this executable, make sure you also install `minitar-cli`.
Successfully installed minitar-0.6
rake aborted!
IOError: closed stream
org/jruby/RubyIO.java:1826:in `seek'
org/jruby/ext/zlib/JZlibRubyGzipReader.java:205:in `rewind'
/home/travis/build/elastic/logstash/vendor/bundle/jruby/1.9/gems/minitar-0.6/lib/archive/tar/minitar/reader.rb:182:in `rewind'
/home/travis/build/elastic/logstash/vendor/bundle/jruby/1.9/gems/minitar-0.6/lib/archive/tar/minitar/input.rb:89:in `each_entry'
/home/travis/build/elastic/logstash/rakelib/vendor.rake:20:in `untar'
/home/travis/build/elastic/logstash/rakelib/vendor.rake:86:in `(root)'
org/jruby/RubyProc.java:281:in `call'
org/jruby/RubyArray.java:1613:in `each'
/home/travis/build/elastic/logstash/rakelib/z_rubycheck.rake:28:in `(root)'
org/jruby/RubyKernel.java:1059:in `load'
org/jruby/RubyKernel.java:1059:in `load'
org/jruby/RubyKernel.java:1079:in `eval'
/home/travis/.rvm/gems/jruby-1.7.25/bin/jruby_executable_hooks:15:in `(root)'
Tasks: TOP => vendor:jruby
(See full trace by running task with --trace)
```

PR #6659, opened 2017-02-07
Pin minitar

jordansissel:
. Works around https://github.com/halostatue/minitar/issues/23

### DIFF
--- a/rakelib/dependency.rake
+++ b/rakelib/dependency.rake
@@ -9,7 +9,7 @@ namespace "dependency" do
   end # task rbx-stdlib
 
   task "archive-tar-minitar" do
-    Rake::Task["gem:require"].invoke("minitar", ">= 0")
+    Rake::Task["gem:require"].invoke("minitar", "0.5.4")
   end # task archive-minitar
 
   task "stud" do


halostatue:
This particular change should *not* be necessary; `minitar` does not require `minitar-cli`; only `archive-tar-minitar` does. However, this should be changed to `~> 0.5` because when I release 1.0 (eventually), there will be breaking changes (I plan on dropping everything below Ruby 2.0).

talevy:
@halostatue 

it is not the dependency on `minitar-cli` that is causing an issue here at the moment. it is this: https://github.com/elastic/logstash/pull/6657#issue-206033272

halostatue:
@talevy Can I get an issue opened on halostatue/minitar for this? If possible, a minimal repro would be useful, because none of my tests (and I do have tests *mostly* around compressed tar streams) have shown an error like this.

talevy:
@halostatue do you test with JRuby? maybe it is specific to that platform?

halostatue:
@talevy My travis matrix is pretty extensive, including JRuby 9k and jruby-19mode, neither of which is allowed to fail in the current mode. I’d like to get this fixed, but I’m at a loss. I’ve also added a comment to #6657. A `--trace` output for this would be more useful than what I’m seeing.

halostatue:
Copying from #6657: Is there a recommended Vagrantfile for testing this so I can try to debug?

I won’t be able to look at it until tomorrow, but this should not be happening (as far as I know, *that* code didn’t change beyond formatting).

jordansissel:
@halostatue i'll try to come up with something and will follow up on  https://github.com/halostatue/minitar/issues/24 with any additional details.

ph:
I've did some testing with this branch and I can build the artifact without any problem.
```
mkdir -p build/bootstrap
[bootstrap] Fetching and installing gem: minitar (0.5.4)
Fetching: minitar-0.5.4.gem (100%)
Successfully installed minitar-0.5.4
Installing minitar 0.5.4 because the build process needs it.
[bootstrap] Fetching and installing gem: minitar (0.5.4)
Fetching: minitar-0.5.4.gem (100%)
Successfully installed minitar-0.5.4
Installing bundler ~> 1.9.4 because the build process needs it.
```

jordansissel:
Merging for now and I'll update the pin to 0.6.1 when time permits maybe later this week.

halostatue:
I recommend considering upgrading to 0.6 now that things are fixed, as I had to fix a CVE in tarball extraction.

PR #6670, opened 2017-02-09
Bump to v5.5

suyograo:


jordansissel:
@suyograo we don't need 5.5 yet. 

We should have 5.3 and 5.x now, not 5.4 and 5.x.

suyograo:
My bad. I blame my jetlag for this :)

jordansissel:
I'll allow it since you came from the future.

PR #6676, opened 2017-02-09
Bugfix/deep en vars

emilesalem:
as demanded by @suyograo


karmi:
Hi @emilesalem, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/logstash/pull/6676.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @emilesalem, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/logstash/pull/6676.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

suyograo:
@emilesalem There's a lot of conflicts with this PR. I suggest you rebase your local fix branch with master and then do a push. This should resolve most differences.

PR #6681, opened 2017-02-09
Pin gems gem to 0.8.3

jordansissel:
Fixes this error:

    % bundle install
    ...
    Installing gems 1.0.0
    Gem::InstallError: gems requires Ruby version >= 2.1.9.


ph:
LGTM

PR #6691, opened 2017-02-10
cleanup partially initialized pipeline

### LINKED ISSUE #6688: reloading problems with exception handling in Pipeline#start_workers
Followup from a discovered issue by @ph in #6656 

`Pipeline#start_workers` can abort with an exception triggers at different initialization state leaving the pipeline in a partially initialized state.

One of these cases is when a plugin crash in the `register` methods.

Ideally, start_workers should correctly handle exceptions in the different initialization phases to avoid  a partially initialized pipeline by cleaning up before exiting the method.

colinsurprenant:
Fixes #6688 

This basically catched plugins initialization exceptions in `start_workers` and cleanup/stop/close any partially initialized plugins before bubbling up exception. This way `start_workers` should not exit and leave any partially initialized pipeline, it either succeed or cleanup state upon error.

- move `start_inputs` at the end of the `start_workers` sequence
- catch exceptions in input, output and filter plugins registration and close those that were registered
- catch exception in `start_inputs` and call `shutdown_workers` before re-raising
- start input threads only once all inputs plugins have been successfully registered

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -141,7 +141,7 @@ module LogStash; class Pipeline < BasePipeline
     begin
       @queue = LogStash::QueueFactory.create(settings)
     rescue => e
-      @logger.error("Logstash failed to create queue", "exception" => e, "backtrace" => e.backtrace)
+      @logger.error("Logstash failed to create queue", "exception" => e.message, "backtrace" => e.backtrace)
       raise e
     end
 
@@ -250,12 +250,32 @@ module LogStash; class Pipeline < BasePipeline
     @running.false?
   end
 
+  # register_plugin simply calls the plugin #register method and catches & logs any error
+  # @param plugin [Plugin] the plugin to register
+  # @return [Plugin] the registered plugin
+  def register_plugin(plugin)
+    plugin.register
+    plugin
+  rescue => e
+    @logger.error("Error registering plugin", :plugin => plugin.inspect, :error => e.message)
+    raise e
+  end
+
+  # register_plugins calls #register_plugin on the plugins list and upon exception will call Plugin#do_close on all registered plugins
+  # @param plugins [Array[Plugin]] the list of plugins to register
+  def register_plugins(plugins)
+    registered = []
+    plugins.each { |plugin| registered << register_plugin(plugin) }
+  rescue => e
+    registered.each(&:do_close)
+    raise e
+  end
+
   def start_workers
     @worker_threads.clear # In case we're restarting the pipeline
     begin
-      start_inputs
-      @outputs.each {|o| o.register }
-      @filters.each {|f| f.register }
+      register_plugins(@outputs)
+      register_plugins(@filters)
 
       pipeline_workers = safe_pipeline_worker_count
       batch_size = @settings.get("pipeline.batch.size")
@@ -285,6 +305,16 @@ module LogStash; class Pipeline < BasePipeline
           Util.set_thread_name("[#{pipeline_id}]>worker#{t}")
           worker_loop(batch_size, batch_delay)
         end
+      end
+
+      # inputs should be started last, after all workers
+      begin
+        start_inputs
+      rescue => e
+        # if there is any exception in starting inputs, make sure we shutdown workers.
+        # exception will already by logged in start_inputs
+        shutdown_workers
+        raise e
       end
     ensure
       # it is important to guarantee @ready to be true after the startup sequence has been completed
@@ -335,7 +365,7 @@ module LogStash; class Pipeline < BasePipeline
     # Users need to check their configuration or see if there is a bug in the
     # plugin.
     @logger.error("Exception in pipelineworker, the pipeline stopped processing new events, please check your filter configuration and restart Logstash.",
-                  "exception" => e, "backtrace" => e.backtrace)
+                  "exception" => e.message, "backtrace" => e.backtrace)
 
     raise e
   end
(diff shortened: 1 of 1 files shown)


logstash-core/lib/logstash/pipeline.rb
```diff
@@ -253,9 +253,16 @@ def stopped?
   def start_workers
     @worker_threads.clear # In case we're restarting the pipeline
     begin
-      start_inputs
-      @outputs.each {|o| o.register }
-      @filters.each {|f| f.register }
+      begin
+        registered = []
+        @outputs.each { |p| p.register; registered << p }
```
colinsurprenant: good suggestion

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -253,9 +253,16 @@ def stopped?
...
+        @filters.each { |p| p.register; registered << p }
+      rescue => e
+        @logger.error("Error registering plugin", :error => e)
+        # close all sucessfully registered plugins
+        registered.each { |p| p.do_close }
```
colinsurprenant: excellent question. I think it would be the responsibility of the failing method to cleanup after itself because it is the only one that knows in which state it is and what needs to be cleaned. so the contact then becomes:

1- the `register` method  never creates any unknown state, it cleans state upon exception
2- only if register has correctly returned can `do_close` be safely called.

colinsurprenant: if we agree on this, we will need to followup up in the plugins at some point.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -286,6 +293,16 @@ def start_workers
           worker_loop(batch_size, batch_delay)
         end
       end
+
+      # inputs should be started last, after all workers
+      begin
+        start_inputs
+      rescue => e
+        # if there is any exception in starting inputs, make sure we shutdown workers.
+        # exception will already by loogged in start_inputs
```
colinsurprenant: loo-de-lee-doo.

colinsurprenant:
Thanks @jsvd - I will test the `recovered head data page 1 is different than checkpoint` problem you described.

logstash-core/lib/logstash/pipeline.rb
ph: For simplicity, should we make `do_close` have an idempotent contract?

ph: of course ^ is part of a bigger discussion :)

ph:
@jsvd How did you get the `recovered head data page 1 is different than checkpoint?

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: refactored!

ph:
@colin @jsvd am I wrong to think that `recovered head data page 1` is another problem not link to this PR intend?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -250,12 +250,32 @@ def stopped?
...
   end
 
+  # register_plugin simply calls the plugin #register method and catches & logs any error
+  # @param plugin [Plugin] the plugin to register
+  # @return [Plugin] the registered plugin
+  def register_plugin(plugin)
+    plugin.register
+    plugin
+  rescue => e
+    @logger.error("Error registering plugin", :plugin => plugin.inspect, :error => e)
```
jsvd: I know this code snippet was my suggestion but let's avoid logging "raw" exception objects as we've had serialization issues in log4j2 in the past, maybe `:error => e.message` or backtrace

colinsurprenant:
@ph not sure - investigating. but at first sight it seems related to shutdown/startup/restart sequencing.

logstash-core/lib/logstash/pipeline.rb
colinsurprenant: ok but there are other places in the Pipeline code which also log `:error => e` ... if I correct this one then I should correct all the other too.

colinsurprenant:
So the `recovered head data page 1 is different than checkpoint` problem is not related to this PR but is an optimistic recovery - I found the bug and fixed it, I will push a PR specifically for this.

jsvd:
I see a failing, which seems related to the smoke tests when running with the persistent queue feature flag:

```
Failures:

  1) Test Logstash instance multiple of them can be started on the same box with automatically trying different ports for HTTP server
     Failure/Error: expect(is_port_open?(9601)).to be true

       expected true
            got false
     # ./specs/01_logstash_bin_smoke_spec.rb:52:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/logstash-devutils-1.3.1-java/lib/logstash/devutils/rspec/logstash_helpers.rb:18:in `try'
     # ./specs/01_logstash_bin_smoke_spec.rb:51:in `(root)'

Finished in 14 minutes 27 seconds (files took 8.68 seconds to load)
32 examples, 1 failure

Failed examples:

rspec ./specs/01_logstash_bin_smoke_spec.rb:41 # Test Logstash instance multiple of them can be started on the same box with automatically trying different ports for HTTP server

Randomized with seed 63256
```

@colinsurprenant it seems it makes sense that this test fails since the PR that added the .lock file (a second logstash instance won't start without changing pipeline.id)?

colinsurprenant:
@jsvd ah makes sense for the http smoke test! let's open an issue for this. /cc @suyograo

ph:
LGTM

colinsurprenant:
rebased, squashed and merged into master and 5.x. Thanks @ph and @jsvd 🙌

ph:
@colin @suyograo I've created this issue https://github.com/elastic/logstash/issues/6707 for following up on the smoke test changes for PQ.

colinsurprenant:
also merged in 5.3

PR #6694, opened 2017-02-11
Bugfix/deepenvars

emilesalem:
@suyograo this should begin to make things right. Now all we need is some tests.


suyograo:
I added tests in e2ffcf9dca1d64b0ec87666ef0f3ded5bebb3bea and merged it to master (efc0a0c5984e18b39c4ff11a0b7f2164cb7c6b94)

PR #6701, opened 2017-02-13
fix bad head positioning on recovery & tests

colinsurprenant:
This fixes a problem found while testing #6691 where recovery would fail with 
```
[2017-02-13T10:37:56,505][WARN ][org.logstash.ackedqueue.Queue] recovered head data page 1 is different than checkpoint, using recovered page information.
```
- The problem is in the head page positioning after recovery. The head needs to be update only after a fully valid recovered element. 
- Added tests for this case. Tests are failing without fix and passing with fix.
- I also left some debug traces used to diagnose the problem. 
- I also added an explicit fsync on close.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
@@ -1,5 +1,7 @@
 package org.logstash.common.io;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.logstash.ackedqueue.SequencedList;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


logstash-core/src/main/java/org/logstash/common/io/AbstractByteBufferPageIO.java
```diff
@@ -158,9 +164,9 @@ private void readNextElement(long expectedSeqNum, boolean verifyChecksum) throws
 
         // at this point we recovered a valid element
         this.offsetMap.add(elementOffset);
-        this.head += length + CHECKSUM_SIZE;
+        this.head = newHead + length + CHECKSUM_SIZE;
 
-        buffer.position(head);
+        buffer.position(this.head);
     }
```
ph: 👍

colinsurprenant:
rebased, squashed and merged into master and 5.x. Thanks @jsvd and @ph !

PR #6704, opened 2017-02-14
Run integration tests separately

andrewvc:
This will clarify our travis results and make test suites run a bit faster by taking advantage of some parallelism.


suyograo:
I like. LGTM

Issue #6603, opened 2017-01-28
Allow to completely disable thread metrics collection (slow thread dump)


micoq:
On Logstash instances with a lot of Java threads (in my case, ~160 RabbitMQ inputs for ~700 threads), the JVM metric collector can decrease a lot the throughput while performing many thread dumps: All the Java threads are stopped during approximately 5sec (workers, inputs, http server...). It's not a memory or a garbage collector problem/tuning.

We can't completely disable the metric collector (it only disable the metric storage): [The poller is always started](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/agent.rb#L220)

If we comment out this line, we can increase the throughput of 20% ! (1000 ev/s -> 1200 ev/s with an heavy grok configuration). I think it would be a good idea to be able to choose the [metric collectors](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/instrument/periodic_pollers.rb#L15).

You can see the JVM pauses (or _safepoints_) on stdout by adding the following options in `jvm.options` before starting Logstash (you can see the thread dumps safepoints with any configuration):
```
-XX:+PrintSafepointStatistics
-XX:PrintSafepointStatisticsCount=1
```

Setup:
- Logstash version: 5.1.1
- Operating System: RHEL 6

andrewvc:
@micoq I'm on the fence with this one. I think the main culprit here is really the excessive thread creation you documented so well in https://github.com/logstash-plugins/logstash-input-rabbitmq/issues/93 . Let's tackle that and see if we still need the patch to core.

Logstash shouldn't be creating 700 threads in any circumstance. That said, we should be sure to benchmark this feature's performance, and if it has a significant impact either decrease its collection frequency or make it possible to disable it.

andrewvc:
@micoq you are definitely right, however, that we shouldn't start the poller if metrics are disabled. WDYT @ph?

ph:
As of today, we didn't expose an option to completly disable the metric collection we might revisit that.

@micoq you only disabled the OS Poller? Not anything else?

jordansissel:
I think I am missing context, so I have to ask -- What indicates that the metrics poller is causing this issue?

ph:
@micoq Do you disable all the periodic pollers?  in 5.1.1, the os poller is a noop. https://github.com/elastic/logstash/blob/v5.1.1/logstash-core/lib/logstash/instrument/periodic_poller/os.rb

ph:
[...]

PR #6705, opened 2017-02-14
Collecting the ThreadCount and the PeakThreadCount should not trigger a threaddump

ph:
When we use the JRMonitor library to get information about the running
threads it will trigger a thread dump to get the stacktrace information this OK when
we do a direct call to the `hot_threads` API but in the context of the
periodic poller this would mean that the threads need to be stopped to
generate their current stacktrace. 

Which could significantly slow down logstash. This PR use the **ThreadMXBean** but only use the `#getThreadCount` and the `#getPeakThreadCount`. Theses two calls won't generate a hreaddump and won't block the currents
threads.

**To test** add the following option to your `config/jvm.options` and let logstash run for a few minutes to trigger a few periodic poller iteration and stop logstash you will see the report.

```
-XX:+PrintSafepointStatistics
-XX:PrintSafepointStatisticsCount=1
```

Fixes: #6603

### DIFF
--- a/logstash-core/lib/logstash/instrument/periodic_poller/jvm.rb
+++ b/logstash-core/lib/logstash/instrument/periodic_poller/jvm.rb
@@ -44,7 +44,7 @@ module LogStash module Instrument module PeriodicPoller
 
     def collect
       raw = JRMonitor.memory.generate
-      collect_jvm_metrics(raw)      
+      collect_jvm_metrics(raw)
       collect_pools_metrics(raw)
       collect_threads_metrics
       collect_process_metrics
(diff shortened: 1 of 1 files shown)


ph:
## Result:

### Before the change

```
Polling page always armed
ThreadDump                       312
ForceSafepoint                     5
Deoptimize                        22
CollectForMetadataAllocation        36
GenCollectForAllocation          457
CMS_Initial_Mark                   1
CMS_Final_Remark                   1
EnableBiasedLocking                1
RevokeBias                       304
BulkRevokeBias                    20
Exit                               1
    4 VM operations coalesced during safepoint
Maximum sync time     14 ms
Maximum vm operation time (except for Exit VM operation)     65 
```
### After the change:
```
Polling page always armed
ForceSafepoint                     7
Deoptimize                        19
CollectForMetadataAllocation        29
GenCollectForAllocation          354
CMS_Initial_Mark                   2
CMS_Final_Remark                   2
EnableBiasedLocking                1
RevokeBias                       302
BulkRevokeBias                    18
Exit                               1
```

ph:
@andrewvc great work from @micoq for pointing us to this bug.

ph:
The failling test in this PR is related to this issue https://github.com/elastic/logstash/issues/6707 and only affect PQ tests.

PR #6708, opened 2017-02-14
LIR Read Only

andrewvc:
## Overview

This is an Initial merge of LIR code into master to help keep all this stuff up to date. This won't affect any existing execution, but is simply meant to be an incremental step on the way to #6281 , which will come in a separate merge. There is potential for us to find some configs that break LIR, which is why we should just merge this into master for now.

## Design Goals

1. **Move the Conceptual Core of Logstash from a Function to a DAG:** Currently, logstash pipeline sections are declared in an imperative configuration language and then transpiled to one Ruby function per section. While this strategy has served logstash well throughout its history it is limited. A DAG representation allows for easier manipulation before runtime and easier introspection during runtime.
1. **Enable a Faster Runtime**: With a more agnostic DAG representation experimenting with alternate execution methodologies becomes much simpler. Our first stab at this is #6281. Previously, in #4727, I benchmarked a DAG based PoC that showed significant performance improvements for real-world workloads. Additionally, regardless of execution strategy, transpiling to Ruby limits Logstash's performance to Ruby performance on the JVM and excludes experimentation with other approaches. 
1. **Open up the Possibility of Alternate Config Languages:** Currently Logstash supports a single configuration language. By having a more generic IR Logstash could easily be extended with multiple config languages. The downside of this is that it could fragment the Logstash community and potentially result in non-interchangeable configurations. This would need to be discussed at length and have its Pros and Cons weighed.
1. **Improve Pipeline Introspection:** Graphs are a richer construct than ASTs in many ways, and flow analysis can be easily done with textbook graph traversal algorithms. We can leverage the DAG to provide new types of introspection.
1. **Improve Code Modularity:** The current design complects execution and representation of pipelines to a large extent. By separating those concerns experimentation becomes easier.


logstash-core/lib/logstash/config/config_ast.rb
```diff
@@ -2,61 +2,7 @@
 require 'logstash/errors'
 require "treetop"
 
-class Treetop::Runtime::SyntaxNode
```
andrewvc: This was all moved to treetop_monkeypatches.

jordansissel:
First things first. Is LIR prounounced in a fun way? "Lyre" ? "Ell Eye Arr" ? "Lure" ?

ph:
@jordansissel Lire ? which is the french word for reading.

andrewvc:
@jordansissel I say "El 👁 Are" :)

jordansissel:
Did a first pass at reviewing the Graph+Edge+Vertex details and provided feedback to @adnrewvc over Zoom.

Next, I will study the test suite.

jordansissel:
I cannot run the tests:

```
% USE_RUBY=1 bin/bundle exec rspec logstash-core/spec/
...
NameError: uninitialized constant SUPPORT_DIR
    const_missing at org/jruby/RubyModule.java:2719
           (root) at /home/jls/projects/logstash/logstash-core/spec/logstash/compiler/compiler_spec.rb:37
      module_exec at org/jruby/RubyModule.java:2356
         subclass at /home/jls/projects/logstash/vendor/bundle/jruby/1.9/gems/rspec-core-3.1.7/lib/rspec/core/example_group.rb:325
         describe at /home/jls/projects/logstash/vendor/bundle/jruby/1.9/gems/rspec-core-3.1.7/lib/rspec/core/example_group.rb:219
           (root) at /home/jls/projects/logstash/logstash-core/spec/logstash/compiler/compiler_spec.rb:20
      module_exec at org/jruby/RubyModule.java:2356
         subclass at /home/jls/projects/logstash/vendor/bundle/jruby/1.9/gems/rspec-core-3.1.7/lib/rspec/core/example_group.rb:325
         describe at /home/jls/projects/logstash/vendor/bundle/jruby/1.9/gems/rspec-core-3.1.7/lib/rspec/core/example_group.rb:219
           (root) at /home/jls/projects/logstash/logstash-core/spec/logstash/compiler/compiler_spec.rb:16
...
```

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -0,0 +1,434 @@
...
+            if (v==null) throw new InvalidIRException("Could not thread vertex, id not found in graph: !" + id + "\n" + this);
+            argVertices[i] = v;
+        }
+        return threadVertices(edgeFactory, argVertices);
+    }
+
+    // Will not validate the graph after running!
+    // You must invoke validate the graph yourself
+    // after invoking
+    public Collection<Edge> threadVerticesUnsafe(Edge.EdgeFactory edgeFactory, Vertex... argVertices) throws InvalidIRException {
```
jordansissel: "thread" is an odd verb here. Is there etymology or a story to tell about the name?

logstash-core/spec/logstash/compiler/compiler_spec.rb
```diff
@@ -0,0 +1,580 @@
+require "spec_helper"
+require "logstash/compiler"
+java_import Java::OrgLogstashConfigIr::DSL
+
+describe LogStash::Compiler do
+  def j
```
andrewvc: This is used all over this file, so its here as shorthand to make the tests more readable, while still providing more namespacing safety than `import`.

logstash-core/spec/logstash/compiler/compiler_spec.rb
```diff
@@ -0,0 +1,580 @@
...
+    end
+
+    context "plugins" do
+      subject(:c_plugin) { compiled[:input] }
+      let(:source) { "input { #{plugin_source} } " }
+
+      describe "a simple plugin" do
+        let(:plugin_source) { "generator {}" }
+
+        it "should contain the plugin" do
```
andrewvc: I agree that this can be improved. 

I think the problem is that sometimes I group tests in a way that doesn't lend itself to english language sentences, but is useful in the actual file. The `context 'plugins'` block that this is nested in (which is responsible for the first word 'plugin' makes sense when reading the file, but makes for a bad test name.

Maybe the solution is to wind up with a sentence like:

`Plugins: A simple plugin declaration should create a plugin IR instance`.

So then I'd have `context 'Plugins: '`. 

WDYT?

spec/spec_helper.rb
```diff
@@ -50,3 +50,17 @@ def puts(payload)
 def installed_plugins
   Gem::Specification.find_all.select { |spec| spec.metadata["logstash_plugin"] }.map { |plugin| plugin.name }
 end
+
+RSpec::Matchers.define :ir_eql do |expected|
+  match do |actual|
+    if expected.java_kind_of?(org.logstash.config.ir.SourceComponent) && actual.java_kind_of?(org.logstash.config.ir.SourceComponent)
+      expected.sourceComponentEquals(actual)
+    else
+      return false
```
andrewvc: Good point! Implicit return would be `nil`, I think for a matcher its better to return `false` than a falsey value. I will restructure this to be nicer however.

andrewvc: Oh, and it should be `next` not `return` in the first place.

logstash-core/src/main/java/org/logstash/config/ir/imperative/ComposedParallelStatement.java
```diff
@@ -0,0 +1,32 @@
+package org.logstash.config.ir.imperative;
```
andrewvc: @jordansissel we need parallel statements to represent outputs which are unordered, as opposed to filters, which are ordered.

In the future we'll need to make this explicit. Users will then have the option of declaring whether filter or output order matters. You might want ordered outputs or unordered filters, or sub-parts of each to act differently, hence the requirement.

spec/support/lscl_configs/nginx_json_logstash.conf
```diff
@@ -0,0 +1,42 @@
+input {
```
andrewvc: These files are a direct import from the `elastic/examples` repo. I'd rather not maintain them, but view them as 'in the wild' configs.

logstash-core/src/test/java/org/logstash/config/ir/IRHelpers.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.UUID;
+
+import static org.logstash.config.ir.DSL.*;
+import static org.logstash.config.ir.PluginDefinition.Type.*;
+
+/**
+ * Created by andrewvc on 9/19/16.
+ */
+public class IRHelpers {
+    public static void assertSyntaxEquals(SourceComponent left, SourceComponent right) {
```
andrewvc: I just changed `assertGraphEquals` to an overloaded version of `assertSyntaxEquals`. Good call :)

logstash-core/src/test/java/org/logstash/config/ir/IRHelpers.java
```diff
@@ -0,0 +1,139 @@
...
+        String message = String.format("Expected '%s' to equal '%s'", left, right);
+        MatcherAssert.assertThat(message, left.sourceComponentEquals(right));
+    }
+
+    public static void assertGraphEquals(Graph left, Graph right) {
+        String message = String.format("Expected \n'%s'\n to equal \n'%s'\n%s", left, right, GraphDiff.diff(left, right));
+        MatcherAssert.assertThat(message, left.sourceComponentEquals(right));
+    }
+
+    public static Vertex testVertex() {
```
andrewvc: Yep, will change this one and similar in the file.

logstash-core/src/main/java/org/logstash/config/ir/expression/unary/Truthy.java
```diff
@@ -0,0 +1,28 @@
...
+ * Created by andrewvc on 9/21/16.
+ */
+public class Truthy extends UnaryBooleanExpression {
+    public Truthy(SourceMetadata meta, Expression expression) throws InvalidIRException {
+        super(meta, expression);
+    }
+
+    @Override
+    public String toRubyString() {
+        return "!!(" + this.getExpression() + ")";
```
andrewvc: You're right about that, I just did that for completeness. 

As far as the idea of `truthy` being confusing for users, I agree, but I think we can hide it. We need it because LIR is supposed to be non-ruby specific and capable of targeting languages with no notion of `truthiness`, like java, where an explicit nil/true/false check is required.

logstash-core/lib/logstash/compiler/lscl.rb
```diff
@@ -0,0 +1,566 @@
...
+      section_map
+    end
+  end
+
+  class Comment < Node; end
+  class Whitespace < Node; end
+  
+  class PluginSection < Node
+    def expr
+      [*recursive_select(Branch, Plugin).map(&:expr)]
```
andrewvc: Yeah, I think this went through a few rewrites and wound up with some redundancy :)

logstash-core/src/test/java/org/logstash/config/ir/graph/GraphTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private void assertVerticesConnected(Graph graph, String fromId, String toId) {
+        Vertex from = graph.getVertexById(fromId);
+        assertNotNull(from);
+        Vertex to = graph.getVertexById(toId);
+        assertNotNull(to);
+        assertVerticesConnected(from, to);
+    }
+
+    public void assertVerticesConnected(Vertex from, Vertex to) {
```
andrewvc: Maybe. They're only used in this file. I'd rather not extract till its needed.

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
andrewvc: I'm renaming these from `thread*` to `chain*`. Good point.

logstash-core/src/test/java/org/logstash/config/ir/PipelineTest.java
```diff
@@ -0,0 +1,32 @@
...
+import static org.logstash.config.ir.DSL.*;
+import static org.logstash.config.ir.PluginDefinition.Type.*;
+
+/**
+ * Created by andrewvc on 9/20/16.
+ */
+public class PipelineTest {
+    @Test
+    public void testPipelineCreation() throws InvalidIRException {
+        Graph inputSection = iComposeParallel(iPlugin(INPUT, "generator"), iPlugin(INPUT, "stdin")).toGraph();
```
andrewvc: I suppose there is the option to have three separate DSL classes. One each for imperative/graph/expression. That would preclude the ability to use static imports as well though.

jordansissel:
LGTM given --

* This is new code, not yet really used, and very likely to change as we see it used in the future
* The `imperative` parts will hopefully go away  once we move more into the lir/dag execution model where we won't need imperative compilation anymore.

I'm ok moving forward as is.

suyograo:
Please only merge this to master. BTW, the tests are failing as well..

andrewvc:
@suyograo tests were passing under `bin/rspec` and just needed a small tweak to `spec_helper` to work with the `rake` task. Rebased and pushed with those changes.

Issue #6707, opened 2017-02-14
Smoke Test PQ for multiple instance on the same box need to be adjusted for PQ.


ph:
Created from this comment: https://github.com/elastic/logstash/pull/6691#issuecomment-279678775
```
Failures:

  1) Test Logstash instance multiple of them can be started on the same box with automatically trying different ports for HTTP server
     Failure/Error: expect(is_port_open?(9601)).to be true

       expected true
            got false
     # ./specs/01_logstash_bin_smoke_spec.rb:52:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/logstash-devutils-1.3.1-java/lib/logstash/devutils/rspec/logstash_helpers.rb:18:in `try'
     # ./specs/01_logstash_bin_smoke_spec.rb:51:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'

Finished in 14 minutes 27 seconds (files took 8.68 seconds to load)
32 examples, 1 failure

Failed examples:

rspec ./specs/01_logstash_bin_smoke_spec.rb:41 # Test Logstash instance multiple of them can be started on the same box with automatically trying different ports for HTTP server

Randomized with seed 
```

The test are failling we try to started 2 logstash instance with the same pipeline_id and with the same path for PQ.

ph:
Not sure why both tests are failling there: https://github.com/elastic/logstash/pull/6704
I remember that this test was a bit flaky.

PR #6709, opened 2017-02-14
Fix the test for running multiples instances of LS on the same machine with PQ on.

ph:
This PR make the smoke test for the multiple instances of logstash works
and also change the assertions of a few tests to make it more robust.

Instead of a `try` I assert the creation of a file generated by a fixed generator.

Fixes: #6707


suyograo:
LGTM. though there's a new test failure unrelated to this I think

ph:
I believe they are not related, I will rekick the job, just to make sure we will have to fix this test too.

PR #6713, opened 2017-02-15
9k update

guyboertje:


rakelib/vendor.rake
```diff
@@ -1,6 +1,7 @@
 namespace "vendor" do
   VERSIONS = {
-    "jruby" => { "version" => "1.7.25", "sha1" => "cd15aef419f97cff274491e53fcfb8b88ec36785" },
+    "jruby" => { "version" => "9.1.6.0", "sha1" => "6dada28876caf2ac14de83dbb45d5243fbbd4940" },
+    # "jruby" => { "version" => "1.7.25", "sha1" => "cd15aef419f97cff274491e53fcfb8b88ec36785" },
```
suyograo: Can we remove this all together?

suyograo:
@guyboertje thanks! Can you take a look at the test failures on travis?

rakelib/vendor.rake
guyboertje: Sure.

guyboertje:
@suyograo 
I have been looking at the Travis fails. Initially I can't decide whether to modify the travis job itself as a fix or to expect that the travis job as is should run and that I must discover that which is preventing travis from running.

guyboertje:
The core specs pass but the integration specs are being hit with this travis "feature" https://github.com/travis-ci/travis-ci/issues/6471

ph:
I will close this and start working on the feature branch instead.

Issue #6692, opened 2017-02-10
Plugin doc gen script pulls wrong text for plugin docs


dedemorton:
The script for generating the plugin docs is currently pulling in a couple of code comments instead of the description that begins at https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/lib/logstash/filters/geoip.rb#L34.

Here's what it looks like in the doc (including a screen capture because Suyog might fix the doc before you see the problem).

![image](https://cloud.githubusercontent.com/assets/14206422/22848114/dc75e968-efa6-11e6-9a9e-d52037d19a8c.png)

The script should be resilient enough to extract the text even if there's a minor formatting snafu (like an extra line).

dedemorton:
@ph I'm seeing a similar problem here where the script takes text that's meant to be a comment: https://www.elastic.co/guide/en/logstash/5.2/plugins-inputs-http.html

![image](https://cloud.githubusercontent.com/assets/14206422/22849098/c941297c-efae-11e6-8bb9-9b530320aff4.png)

dedemorton:
Here's another one. But in this case, there is no description of what the plugin does. "handles disconnects" comes from a code comment, tho. :-( 

![image](https://cloud.githubusercontent.com/assets/14206422/22849606/9053e08c-efb3-11e6-9953-2947fece9cb5.png)

dedemorton:
Here's another one. 

![image](https://cloud.githubusercontent.com/assets/14206422/22849682/47173652-efb4-11e6-814d-da895a3b5a95.png)

dedemorton:
From https://www.elastic.co/guide/en/logstash/5.2/plugins-outputs-boundary.html, not sure if this is a problem with the doc gen or the code itself:

![image](https://cloud.githubusercontent.com/assets/14206422/22849890/9647845a-efb6-11e6-8799-6920a1207282.png)

dedemorton:
This one is pulling the license agreement and squishing it together with the description:

![image](https://cloud.githubusercontent.com/assets/14206422/22850253/37f9f7a2-efbb-11e6-8b73-220881b4e0e6.png)

dedemorton:
![image](https://cloud.githubusercontent.com/assets/14206422/22850374/272082dc-efbd-11e6-9459-45156b65c425.png)

dedemorton:
I'm guessing this is not the droid we're looking for:

[...]

PR #6714, opened 2017-02-15
Fix a few issues when using the logstash-docgen tools

ph:
Fixes the following issues:

- Make multiline plugin definition correctly take the description See
rabbit mq input or output
- Use a regexp to ignore encoding comment
- Use `TODO` The doc gen will now ignore some TODOs.
- rules are now executed in the correct order,
- make sure to ignore comment not associated with a description of an
attribute

Ref: #6692


ph:
@suyograo I am doing full build, will report any failures here.

ph:
its missing a fix for the google* plugins to remove the license file in the header

ph:
I have added some code to skip the license for the google plugins.

PR #6720, opened 2017-02-16
Don't log config content when reload is not triggered

suyograo:


jsvd:
hum...not sure that this is a bug. it only logs when debug flag is enabled, and might be useful to understand exactly what is being compared, in a scenario where the pipeline isn't being reloaded and should be.

what is the problem this is aiming to fix (there is no description or link to issue)?

jordansissel:
LGTM.

suyograo:
Merged to 5.2.2 as well.

PR #6725, opened 2017-02-16
unify gradle projects

talevy:
Currently, there are three separate gradle projects in this repository.

These gradle files unify them so that one can call `gradle build` from the parent `logstash` directory and build all of the projects together. This also allows one to run `gradle idea` to generate an intellij `.ipr` project for more seamless development between the three projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,6 @@ qa/Gemfile.lock
 *.ipr
 *.iws
 *.iml
-.gradle
 .idea
 logs
 qa/integration/services/installed/
--- a/build.gradle
+++ b/build.gradle
@@ -1,0 +1,1 @@
+apply plugin: 'idea'
--- a/logstash-core-event-java/build.gradle
+++ b/logstash-core-event-java/build.gradle
@@ -106,7 +106,7 @@ dependencies {
     testCompile 'junit:junit:4.12'
     testCompile 'net.javacrumbs.json-unit:json-unit:1.9.0'
     provided 'org.jruby:jruby-core:1.7.25'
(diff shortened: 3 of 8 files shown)


colinsurprenant:
+1 on the idea. testing workflow locally.

talevy:
@colinsurprenant thanks, seems it broke some `.jar` logic. probably because of the project naming, I'll verify

colinsurprenant:
overall seems to work well. One thing which does not work anymore is building a subproject within that subproject dir:
```
cd logstash-core-event-java
gradle build
```
it complains with
```
> Project with path ':logstash-core' could not be found in root project 'logstash-core-event-java'.
```

is that fixable? I am not sure how we can act on specific subprojects with gradle with this new structure.

talevy:
@colinsurprenant without refactoring each project's gradle files significantly, not sure. only non-intrusive way I see to fix that is to update the `settings.gradle` for each subproject to copy what is happening in the parent project. I'll take another stab at it

colinsurprenant:
@talevy I added this to `logstash-core-event-java/settings.gradle` and I think it is working:
```
include ':logstash-core'
project(':logstash-core').projectDir = new File('../logstash-core')
```

talevy:
right, that is what I was referring to, but I'd like to avoid that... if it gets complicated, I'll just forfeit to that.

another question, is it really important that we support this?

colinsurprenant:
I guess it's not a blocker, it's just that this is my workflow, to build and tests the specific module I am working on. 

OTH if "back linking" subprojects to the main project works, is it a problem to do so?

talevy:
added the back-linking. good to go!

colinsurprenant:
great!! 👍

PR #6739, opened 2017-02-20
use queue path in memory acked queue to namespace .lock file

### LINKED ISSUE #6732: in memory acked queue should scope the lock to the pipeline_id
Currently, when we run the spec we used an in memory acked queue, this testing queue will use the lock for the PQ. The problem is the created lock is a global lock and is not scoped to the `pipeline_id`, so creating more than one active pipelines will fail.

Also the created file for the lock is created in $LOGSTASH_HOME, if you start logstash and a test fails a **.lock** file could still be present in a future run since the **clear_data_dir** doesn't clear that file.

```ruby
def clear_data_dir
    data_path = agent_settings.get("path.data")
    Dir.foreach(data_path) do |f|
    next if f == "." || f == ".." || f == ".gitkeep"
    FileUtils.rm_rf(File.join(data_path, f))
  end
end
```

colinsurprenant:
fixes #6732 

simply pass the namespaced queue path to the memory acked queue constructor so that exclusive lock places the `.lock` file in that dir.

### DIFF
--- a/logstash-core/lib/logstash/queue_factory.rb
+++ b/logstash-core/lib/logstash/queue_factory.rb
@@ -16,18 +16,20 @@ module LogStash
       checkpoint_max_writes = settings.get("queue.checkpoint.writes")
       checkpoint_max_interval = settings.get("queue.checkpoint.interval")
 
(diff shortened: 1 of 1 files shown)


colinsurprenant:
merged in master and 5.x

PR #6745, opened 2017-02-21
memory checkpoint namespace

colinsurprenant:
namespace the in-memory checkpoint by dir path to support multiple instances of memory acked queues.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/io/MemoryCheckpointIO.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/MemoryCheckpointIO.java
@@ -1,9 +1,12 @@
 package org.logstash.common.io;
 
 import org.logstash.ackedqueue.Checkpoint;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 
 import java.io.IOException;
 import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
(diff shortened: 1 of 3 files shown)


colinsurprenant:
@ph per our discussion.

ph:
@colinsurprenant This fixed the problem, I just reverted my hack and all my test is green now!

logstash-core/src/main/java/org/logstash/common/io/MemoryCheckpointIO.java
```diff
@@ -12,7 +15,7 @@
     private final String HEAD_CHECKPOINT = "checkpoint.head";
     private final String TAIL_CHECKPOINT = "checkpoint.";
 
-    private static final Map<String, Checkpoint> sources = new HashMap<>();
+    private static final Map<String, Map<String, Checkpoint>> sources = new HashMap<>();
```
ph: Since this could be called from multiple threads should we make this thread safe?

colinsurprenant: all classes below the `Queue` class are not threadsafe, but all Queue operations are wrapped in a mutex so the Checkpoint operations are safe in that respect.

colinsurprenant:
rebased, squashed and merged into master and 5.x

PR #6754, opened 2017-02-24
make integration specs more resilient

jsvd:


qa/integration/services/logstash_service.rb
```diff
@@ -165,6 +165,7 @@ def wait_for_logstash
       end
       tries -= 1
     end
+    raise RuntimeError.new("failed to start Logstash after #{RETRY_ATTEMPTS} seconds. aborting..")
```
ph: You need to check if you ran out of retries, because right now all the execution will go through this path event if it succeeded.

jsvd: 🤦‍♂️

ph:
I've been also trying to see if I can improve the stability of the integration test suite and I came across a weird behavior. I have added a breakpoint when one of theses tests failed, when I try to connect to the port outside the pry prompt it work.. when I used the TCP class it fails.

Issue #6533, opened 2017-01-13
Metrics internal API: Add method to check if key exists


suyograo:
Some metrics are available only on certain platforms/OS, so it may be empty when queried on these systems. The approach [now](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/api/commands/stats.rb#L72) is to do this in a `rescue` block and catch `MetricNotFound` exception. I think a better way is to expose a `has_metric()` which takes a path.

`has_metric?([:os])`
`has_metric?([:os, cgroups])`

ph:
+1 since exception are expensive in ruby/jruby.

jsvd:
doesn't this mean we'd have to do the lookup twice? like

```
if has_metric?([:os])
  get_shallow([:os])
end
```

The only alternative I see is a value that represents a failed lookup..

ph:
@jsvd the `has_metric` could target the fast lookup table?

jsvd:
fair point. even from that perspective it is tricky to ensure that this change actually makes the overall performance better due to the following observations:

1. overall most get operations succeed consistently
2. some get operations on certain keys and certain environments will never succeed 
3. as it is, a succesful lookup does `get_shallow` and a failed one does `get_shallow + exception handling`
4. with the change, both a sucessful and a failed lookup will do `has_metric + get_shallow`
5. on the most common production platforms (linux based), we'll be adding the extra has_metric call needlessly (because all lookups are succesful).

Another suggestion would be to do like `Hash.fetch` and have an extra parameter for default value.

suyograo:
> on the most common production platforms (linux based), we'll be adding the extra has_metric call needlessly (because all lookups are succesful).

I wouldn't do this for _all_ metric calls. Just ones we know may be empty. It would still be faster than relying on exceptions that we have now.

> and have an extra parameter for default value.

I like this better, actually. But we should short circuit immediately in the code and not do any extra work if the key does not exist.

PR #6759, opened 2017-02-27
Implements #has_metric?(*path)

ph:
Add a new method that uses the `fast_lookup` has to find if a specific
metric exist instead of relying on exceptions.

Usage:

```ruby
metric_store.has_metric?(:node, :sashimi, :pipelines, :pipeline01, :plugins, :"logstash-output-elasticsearch", :event_in) # true
metric_store.has_metric?(:node, :sashimi, :pipelines, :pipeline01, :plugins, :"logstash-output-elasticsearch", :do_not_exist) # false
```
Fixes: #6533


jsvd:
LGTM

Issue #6724, opened 2017-02-16
Logstash 5.3 doc fixes


dedemorton:
This list is a work-in-progress. Need to meet with dev to prioritize/refine goals

**5.3:**

- [x] Document Filebeat modules (see https://github.com/elastic/logstash/issues/6542)
- [x] Generate plugin docs for 5.3. Make sure the manual tweaks made in https://github.com/elastic/logstash-docs/pull/353 get in. Make sure the docs for new plugins that got merged get generated.
- [x] Document content-type header integration compatibility (See https://github.com/elastic/logstash/issues/6718)
- [x] Add "at least once" delivery guarantees to the PQ docs.
- [x] Make remaining plugin changes (I'll move items from [here](https://github.com/elastic/logstash/issues/6501) to this list after the dust settles).
- [x] Bug fixes and enhancements tagged for 5:3, including (but not limited to):

Moved "backlog" to https://github.com/elastic/logstash/issues/6850.

PR #6761, opened 2017-02-28
Add at least once delivery to persistent queue docs

dedemorton:
@acchen97  I've added the "at least once" delivery guarantee back into the PQ docs, as you requested. I've used the language that we removed before.

@jordansissel Can you confirm that this language is accurate for 5.3? Asking you because you did so much work to improve the accuracy of the PQ docs during the last round of updates. Can you let me know if there's any content that is no longer true given the changes added in 5.3.

This PR addresses one of the items tracked in https://github.com/elastic/logstash/issues/6724

### DIFF
--- a/docs/static/persistent-queues.asciidoc
+++ b/docs/static/persistent-queues.asciidoc
@@ -21,19 +21,19 @@ persistent queues to buffer events on disk and remove the message broker.
 
 In summary, the benefits of enabling persistent queues are as follows:
 
-* Provides protection from in-flight message loss when the Logstash process is abnormally terminated.
-* Absorbs bursts of events without needing an external buffering mechanism like Redis or Apache Kafka.
-* Provides an at-least-once message delivery guarantee. If Logstash is restarted while events are
-in-flight, Logstash will attempt to deliver messages stored in the persistent queue until delivery
-succeeds at least once. In other words, messages stored in the persistent queue may be duplicated, but
-not lost.
+* Absorbs bursts of events without needing an external buffering mechanism like
+Redis or Apache Kafka.
+* Provides an at-least-once delivery guarantee against message loss during
+a normal shutdown as well as when Logstash is terminated abnormally. If Logstash
+is restarted while events are in-flight, Logstash will attempt to deliver
+messages stored in the persistent queue until delivery succeeds at least once.
 
 [[persistent-queues-limitations]]
 ==== Limitations of Persistent Queues
 
 The following are problems not solved by the persistent queue feature:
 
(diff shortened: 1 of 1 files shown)


acchen97:
Thanks @dedemorton. In the end, its really still two main benefits. The "at-least-once" is part of the "message loss protection" story. Perhaps we can combine the two?

docs/static/persistent-queues.asciidoc
```diff
@@ -19,10 +19,14 @@ Instead of deploying and managing a message broker, such as Redis, RabbitMQ, or
 Apache Kafka, to facilitate a buffered publish-subscriber model, you can enable
 persistent queues to buffer events on disk and remove the message broker.
 
-In summary, the two benefits of enabling persistent queues are as follows:
+In summary, the benefits of enabling persistent queues are as follows:
 
 * Provides protection from in-flight message loss when the Logstash process is abnormally terminated.
 * Absorbs bursts of events without needing an external buffering mechanism like Redis or Apache Kafka.
+* Provides an at-least-once message delivery guarantee. If Logstash is restarted while events are
```
jordansissel: This seems to duplicate information from the first bullet point ("Provides protection from ...") but maybe it's ok to add here given PQ has low user awareness right now and this might help clarify confusion.

jordansissel:
LGTM

docs/static/persistent-queues.asciidoc
dedemorton: @jordansissel How about combining the two statements into the following? I want to make sure this applies to both a controlled shutdown and abnormal termination.

"Provides an at-least-once delivery guarantee against message loss when the Logstash process is shut down or abnormally terminated. If Logstash is restarted while events are in-flight, Logstash will attempt to deliver messages stored in the persistent queue until delivery succeeds at least once. In other words, messages stored in the persistent queue may be duplicated, but not lost.

dedemorton: @acchen97 ^^

jordansissel: @dedemorton ++++

One minor comment "shutdown or abnormally terminated" to me feels like we're equivocating "shut down" and "abnormally terminated" . Thinking outloud, maybe minor rephrase to say "against message loss for both normal shutdown as well as abnormal termination" ?

acchen97: > In other words, messages stored in the persistent queue may be duplicated, but not lost.

This last sentence is odd. I think it makes sense as "messages read from the persistent queue..." or simply "messages emitted from Logstash..." Other than that, LGTM.

dedemorton: @acchen97 Yah, I agree. Maybe we should just drop that last sentence completely. It's not exactly a benefit when it sends duplicate events, right?

dedemorton:
OK to merge this? @jordansissel I changed your wording slightly because "abnormal termination" sounds weirdly clinical.

jordansissel:
@dedemorton +1

LGTM

PR #6767, opened 2017-02-28
Rename Event Dependent Configuration chapter

dedemorton:
Renaming to a more intuitive name as discussed originally in PR #5696.


suyograo:
LGTM

PR #6775, opened 2017-03-01
Add conditional coding for install instructions

dedemorton:
Some users inadvertently go to the master version of the docs and try to download versions of our software that don't yet exist. The conditional coding in this PR addresses that issue. 

It also means that we don't need to manually  remove "prerelease" from the path when we get docs ready for a release. We simply need to set the release state to `:release-state:  released` in the index.asciidoc (we're already doing this for the docker instructions).

@suyograo FYI


untergeek:
So needed.  Thanks @dedemorton !

jordansissel:
LGTM, but I haven't tested the docs build.

PR #6777, opened 2017-03-01
Fix building the doc for the codec multiline and the redis input

ph:
Fix an issue with handling the plugin type with new style declaration
Correctly extract the plugin type when the plugin is defined with

```
module LogStash module Inputs class XXX < ...
```

Also adjust the regular expression to skip `TODO` statement to ignore `:`

### DIFF
--- a/tools/logstash-docgen/lib/logstash/docgen/parser.rb
+++ b/tools/logstash-docgen/lib/logstash/docgen/parser.rb
@@ -83,8 +83,8 @@ module LogStash module Docgen
 
     def release_date(format = "%B %-d, %Y")
       @release_date ||= begin
-
-                          response = open("https://rubygems.org/api/v1/versions/#{canonical_name}.json").read
+                          url ="https://rubygems.org/api/v1/versions/#{canonical_name}.json"
+                          response = open(url).read
                           # HACK: One of out default plugins, the webhdfs, has a bad encoding in the gemspec
                           # which make our parser trip with this error:
                           #
--- a/tools/logstash-docgen/lib/logstash/docgen/static_parser.rb
+++ b/tools/logstash-docgen/lib/logstash/docgen/static_parser.rb
@@ -6,7 +6,7 @@ module LogStash::Docgen
   class StaticParser
     COMMENTS_IGNORE = Regexp.union(
       Regexp.new(/encoding: utf-8/i),
-      Regexp.new(/TODO:/)
+      Regexp.new(/TODO:?/)
     )
 
     VALID_CLASS_NAME = /^LogStash::(Codecs|Inputs|Filters|Outputs)::(\w+)/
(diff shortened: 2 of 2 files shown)


dedemorton:
@ph The docgen tool ran without errors (in master), but I encountered a couple of errors when I ran the doc build:

**logstash-filter-dissect** (this is an easy fix in the doc source if we don't want to modify the tool)

The addition of the "Description" section results in an "Element section content does not follow the DTD" error. I think this is because the content extracted from the source starts with a level 4 section. So when the "Description" section is added, we end up with two level 4 headings in a row, and asciidoc does not allow this.

`````
==== Description

==== *Dissect or how to de-structure text*

The Dissect filter is a kind of split operation...
`````

We could add a line with `&nbsp;` (like we do in other places), or we could have a rule that the description cannot begin with a section heading (in that case, we could just update the asciidoc in the ruby source code).

**plugins-filters-dns**

The doc build raises the following error. When I look at the generated asciidoc, I see that it's caused by `(Resolv::` appearing in the following section: 

`````
[[plugins-filters-dns-failed_cache_size]]
===== `failed_cache_size` 

  * Value type is <<number,number>>
  * Default value is `0`

cache size for failed requests (Resolv::
`````
I'm not sure where `Resolv::` is coming from, but it causes an error because `::` are reserved characters in asciidoc.

**plugins-inputs-beats**

The doc still contains the text "This monkey patch add callback based flow to the codec until its shipped with core. This give greater flexibility to the implementation by sending more data to the actual block." 

I think this is from one of Jordan's TODO items (not sure, tho). Maybe we can fix it in the source, but you might want to verify that the logic you've added for ignoring TODO items is working OK.

**Comment by ph** [comment]
@dedemorton 

I've looked at the DNS filter, The `Resolv` error seems to have ben fixed with this commit:
https://github.com/logstash-plugins/logstash-filter-dns/commit/4d539b2db24bb8843385128473c5772213719af0

>We could add a line with &nbsp; (like we do in other places), or we could have a rule that the description cannot begin with a section heading (in that case, we could just update the asciidoc in the ruby source code).

You mean adding a &nbsp; after the `description` title? I can update the template with that.

**Comment by ph** [comment]
@dedemorton 

```asciidoc
==== Description

<%= description %>

&nbsp;
```
Looking at the template, concerning the non breaking space do you mean adding one before the actual description?

dedemorton:
@ph I think you would have to put it before the description in order to get something like the following.

````
==== Description

&nbsp;

==== Another Heading

Blah blah blah

````
But TBH, I don't really like this fix because having an extra line when there is content makes it look like we've made a formatting mistake. 

I say that we deal with this problem when it comes up by fixing it in the source. The first line of the description cannot start with a level 4 heading. It either needs to start with text of a level 5 heading.

ph:
@dedemorton Yeah, we need to add theses check on our side in our PR review, I wasn't aware of the level heading issue.

PR #6783, opened 2017-03-02
remove old legacy logstash-core-event

talevy:
and migrate spec tests to core

Also, a gem build task that would probably break on non-snapshot builds (which is not caught by CI)


colinsurprenant:
nice nice nice  - LGTM!

talevy:
ported to `5.x` branch under https://github.com/elastic/logstash/commit/1643cfefaf9e62b78c990e0766555751b789e279

PR #6784, opened 2017-03-02
Make location of bin/logstash more obvious and fix TOC org

### LINKED ISSUE #6219: docs/packaging: bin/logstash has non-obvious location
Example: bin/logstash could appear in different places depending on how a user installed it. Our docs mention `bin/logstash` as shorthand, but when installing rpm/deb this is really /usr/share/logstash/bin/logstash.

My main goal is to reduce user confusion, since bin/logstash has a discoverability problem ("Where is it?"), this means users are having a bad time.

Proposed solution: Update all mentions of `bin/logstash` to include a note indicating where to find `bin/logstash` depending on the method of installation (rpm/deb/tar/zip)

dedemorton:
This fixes #6219. 

I don't think it makes sense to update _every_ mention of `bin/logstash` with a description of where to find the executable on all platforms (as suggested in 6219). There are 160 instances of the command across 101 files.  Instead I've mentioned the location in the key places where new users are likely to first encounter the `bin/logstash` command. 

Other changes:
- Reorganized the TOC topics to be more logically ordered. The changes will look like this:
image
- Added info about running Logstash from the command line. The topic (originally command-line.asciidoc) previously contained merely a list of command line options. It was weird to have a topic that showed the command line options without showing the actual command. I also thought it fit into the flow better to have a topic that tells users the command for running Logstash  in the foreground for testing. 
- Had to move the content in the topic "Running Logstash as a Service on Debian or RPM" to its own asciidoc file so that I could reorganize things better. This won't affect linking.
- renamed command-line.asciidoc to running-logstash-command-line.asciidoc

### DIFF
--- a/docs/static/running-logstash-command-line.asciidoc
+++ b/docs/static/running-logstash-command-line.asciidoc
@@ -21,15 +21,16 @@ the `mypipeline.conf` file:
 bin/logstash -f mypipeline.conf
 ----
 
+Any flags that you set at the command line override the corresponding settings
+in the Logstash <<logstash-settings-file,settings file>>, but the settings file
+itself is not changed. It remains as-is for subsequent Logstash runs.
+
 Specifying command line options is useful when you are testing Logstash.
 However, in a production environment, we recommend that you use the Logstash
 <<logstash-settings-file,settings file>> to control Logstash execution. Using
 the settings file makes it easier for you to specify multiple options, and it
 provides you with a single, versionable file that you can use to start up
 Logstash consistently for each run.
-
-Any flags that you set at the command line override the corresponding settings
-in the Logstash <<logstash-settings-file,settings file>>.
 
 [[command-line-flags]]
 ==== Command-Line Flags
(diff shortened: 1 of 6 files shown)


docs/static/running-logstash-command-line.asciidoc
```diff
@@ -1,15 +1,40 @@
...
-Instead of specifying options at the command line, we recommend that you control Logstash execution
-by specifying options in the Logstash <<logstash-settings-file,settings file>>. Using a settings file
-makes it easier for you to specify mutliple options, and it provides you with a single, versionable
-file that you can use to start up Logstash consistently for each run.
+[source,shell]
+----
+bin/logstash -f mypipeline.conf
+----
+
+Specifying command line options is useful when you are testing Logstash.
```
suyograo: We should also say that any command line options specified will overwrite the configuration specified in the settings file. Also, if you provide a configuration via command line, the settings file is not changed and will remain as-is.

suyograo:
minor comment

dedemorton:
@suyograo I've added info to clarify how setting options at the command line affects the settings file. Is this OK to merge?

suyograo:
LGTM.

Issue #6256, opened 2016-11-15
5.0.1 - Erroneous 'warn' log message when --config.debug is used


patrickpilch:
When `--log.level=debug` and `--config.debug` are provided, a `WARN` level log message is produced incorrectly stating the log level was not set to 'debug'.

- Version: **5.0.1**
- Operating System: **OS X Darwin Kernel Version 15.6.0**
- Config File:
```ruby
input { stdin { } }
output { stdout { codec => json } }
```
- Sample Data: **Anything**
- Steps to Reproduce:
```bash
$ ./logstash -f ./test.conf --log.level=debug --config.debug
```
```
...

[2016-11-16T11:01:43,507][WARN ][logstash.runner          ] --config.debug was specified, but log.level was not set to 'debug'! No config info will be logged.

...

[2016-11-16T11:01:43,699][DEBUG][logstash.agent           ] 
input { stdin {} }
output { stdout { codec => json } }



[2016-11-16T11:01:43,700][DEBUG][logstash.agent           ] 
The following is the merged configuration
[2016-11-16T11:01:43,700][DEBUG][logstash.agent           ] 
input { stdin {} }
output { stdout { codec => json } }

```

jsvd:
this is still true in the latest snapshot:

```
% ./bin/logstash -f /tmp/cfg --log.level=debug --config.debug
Sending Logstash's logs to /tmp/logstash-5.2.0/logs which is now configured via log4j2.properties
[2017-01-06T18:08:04,709][WARN ][logstash.runner          ] --config.debug was specified, but log.level was not set to 'debug'! No config info will be logged.
```

suyograo:
Fixed in #6789

PR #6789, opened 2017-03-02
fix wrong warning about log level debug + config.debug flags 

jsvd:
~~during a refactor a bug was introduced in the boolean logic that prevents the config from being shown even if the `--config.debug` flag is enabled:~~

during a refactor a bug was introduced in the boolean logic that shows an incorrect warning that debug was not enabled even though it is:

```ruby
if setting("config.debug") && logger.debug?
  logger.warn("--config.debug was specified, but log.level was not set to \'debug\'! No config info will be logged.")
end
```

this logic used to be:

```ruby
if setting("config.debug") && @logger.level != :debug
  @logger.warn("--config.debug was specified, but log.level was not set to \'debug\'! No config info will be logged.")
end
```

this pr re-establishes the correct behaviour

fixes https://github.com/elastic/logstash/issues/6256

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -207,7 +207,7 @@ class LogStash::Runner < Clamp::StrictCommand
     # override log level that may have been introduced from a custom log4j config file
     LogStash::Logging::Logger::configure_logging(setting("log.level"))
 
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM. I'm merging this.

PR #6796, opened 2017-03-02
[WIP] batched queue

colinsurprenant:
This implements a new in-memory batched queue. 

In this PR, the specs default queue is set to batched queue with queue drain enabled in `spec/spec_helper.rb` to make sure all queued events are processed before shutting down. All specs are passing locally.

Preliminary tests are showing improved pipeline performance, will followup with result details.

- [ ] performance comparaison 
- [ ] queue drain should be enable by default when using batched queue
- [ ] figure if/how we should run specs for both `BatchedQueue` and `AckedQueue` implementations (maybe use a flag/ENV var to specify which queue impl to run specs on and simply run specs twice, once for each impl).
- [ ] not a blocker for it is becoming important to clean/dry/rewrite the queue wrappers - as discussed with @ph I am planning on a Java rewrite for better integration with LIR pipeline executor


andsel:
The PR has conflicts and doesn't describe which problem it address and how, so convert to draft

logstashmachine:
Since this is a community submitted pull request, a Jenkins build has not been kicked off automatically. Can an Elastic organization member please verify the contents of this patch and then kick off a build manually?

PR #6800, opened 2017-03-04
signal notfull on page purge

### LINKED ISSUE #6626: pipeline doesn't recover if page is purged after queue.max_bytes is reached
setup:

```yml
queue.type: persisted
queue.page_capacity: 5mb
queue.max_bytes: 10mb
```

start logstash with a stalling pipeline:

```
bin/logstash -e "input { generator {} } filter {sleep { time => 1 } }" -w 1 -b 1
```

wait for a bit, the start logstash without the sleep filter:

```
bin/logstash -e "input { generator {} } filter { }" -w 1 -b 1
{
      "sequence" => 1,
    "@timestamp" => 2017-02-01T20:43:03.873Z,
      "@version" => "1",
          "host" => "Joaos-MBP-5.lan",
       "message" => "Hello world!"
}
{
      "sequence" => 2,
    "@timestamp" => 2017-02-01T20:43:03.972Z,
      "@version" => "1",
          "host" => "Joaos-MBP-5.lan",
       "message" => "Hello world!"
}
....
{
      "sequence" => 37448,
    "@timestamp" => 2017-02-01T20:43:05.297Z,
      "@version" => "1",
          "host" => "Joaos-MBP-5.lan",
       "message" => "Hello world!"
}
{
      "sequence" => 37449,
    "@timestamp" => 2017-02-01T20:43:05.297Z,
      "@version" => "1",
          "host" => "Joaos-MBP-5.lan",
       "message" => "Hello world!"
}
{
      "sequence" => 0,
    "@timestamp" => 2017-02-01T20:43:44.280Z,
      "@version" => "1",
          "host" => "Joaos-MBP-5.lan",
       "message" => "Hello world!"
}
# event flow stops
```

colinsurprenant:
Fixes #6626

Nasty bug where acking a batch and purging a page did not signal a notFull condition. 

I also changed all signal() for signalAll() because otherwise only 1 worker/thread would resume if many workers where waiting on a condition. All workers/thread should be given a chance to resume.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -330,7 +330,9 @@ public class Queue implements Closeable {
             this.unreadCount++;
 
             // if the queue was empty before write, signal non emptiness
-            if (wasEmpty) { notEmpty.signalAll(); }
+            // a simple signal and not signalAll is necessary here since writing a single element
+            // can only really enable a single thread to read a batch
+            if (wasEmpty) { notEmpty.signal(); }
 
             // now check if we reached a queue full state and block here until it is not full
             // for the next write or the queue was closed.
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -45,7 +45,6 @@ public class Queue implements Closeable {
     protected final List<TailPage> unreadTailPages;
 
     protected volatile long unreadCount;
-
     protected volatile long currentByteSize;
 
     private final CheckpointIO checkpointIO;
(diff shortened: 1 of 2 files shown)


jsvd:
the fix works, new tests work and I also tested manually.
LGTM

colinsurprenant:
I will have to make a post-LTGM change :P 
The `write()` method `notEmpty` signalling was changed from a `signal()` to `signalAll()` but this is unnecessary. Since this `write()` method writes just a single element, only 1 suspended thread on the `notEmpty` signal needs to be waked up. When we eventually introduce a write method that can take multiple elements then `signalAll()` will make sense since potentially many threads could be able to grab batches off this write.

suyograo:
@jsvd are you still good with this after @colinsurprenant post-LGTM change?

jsvd:
the rationale makes sense and behaviour is still good, LGTM

PR #6817, opened 2017-03-16
Add support for Dead Letter Queue RecordIO library

talevy:
This PR adds a few base components for the Dead Letter Queue

- RecordIO Library

This library writes and reads events into a file with a specific format. The format splits up the file into contiguous blocks of the same size. These blocks contain records, which are either complete or partial byte arrays making up the full event being written.

- DeadLetterQueue Manager

These read/write managers build on top of RecordIO to manage a sorted list of recordio files that, together, make up the DLQ

### DIFF
--- a/logstash-core/src/main/java/org/logstash/DLQEntry.java
+++ b/logstash-core/src/main/java/org/logstash/DLQEntry.java
@@ -58,17 +58,12 @@ public class DLQEntry implements Cloneable, Serializable, Queueable {
                 + pluginTypeBytes.length
                 + pluginIdBytes.length
                 + reasonBytes.length
-                + (Integer.BYTES * 5));
-        buffer.putInt(entryTimeInBytes.length);
-        buffer.put(entryTimeInBytes);
-        buffer.putInt(eventInBytes.length);
-        buffer.put(eventInBytes);
-        buffer.putInt(pluginTypeBytes.length);
-        buffer.put(pluginTypeBytes);
-        buffer.putInt(pluginIdBytes.length);
-        buffer.put(pluginIdBytes);
-        buffer.putInt(reasonBytes.length);
-        buffer.put(reasonBytes);
+                + (Integer.BYTES * 5)); // magic number represents the five byte[] + lengths
+        putLengthAndBytes(buffer, entryTimeInBytes);
+        putLengthAndBytes(buffer, eventInBytes);
+        putLengthAndBytes(buffer, pluginTypeBytes);
+        putLengthAndBytes(buffer, pluginIdBytes);
+        putLengthAndBytes(buffer, reasonBytes);
         return buffer.array();
     }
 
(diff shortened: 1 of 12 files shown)


talevy:
Hey Jordan, I am still working on adding a few more tests and cleaning up the documentation, but I feel that you are up-to-speed with what is going on enough to maybe begin a code review. let me know what you think, thanks!

logstash-core/src/main/java/org/logstash/DLQEntry.java
```diff
@@ -0,0 +1,127 @@
...
+        return buffer.array();
+    }
+
+    public static DLQEntry deserialize(byte[] bytes) throws IOException {
+        ByteBuffer buffer = ByteBuffer.allocate(bytes.length);
+        buffer.put(bytes);
+        buffer.position(0);
+
+        int entryTimeLength = buffer.getInt();
+        byte[] entryTimeBytes = new byte[entryTimeLength];
```
talevy: yeah, good point. lots of this can be cleaned up.

logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueWriteManagerTest.java
```diff
@@ -0,0 +1,95 @@
...
+        dir = temporaryFolder.newFolder().toPath();
+    }
+
+    @Test
+    public void testLockFileManagement() throws Exception {
+        Path lockFile = dir.resolve(".lock");
+        DeadLetterQueueWriteManager writer = new DeadLetterQueueWriteManager(dir, 1000, 1000000);
+        assertTrue(Files.exists(lockFile));
+        writer.close();
+        assertFalse(Files.exists(lockFile));
```
talevy: this test is here to only verify the `.lock` file itself, locking behavior of this file is done in `testFileLocking`. should I merge the two tests into one?

logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueReadManagerTest.java
```diff
@@ -0,0 +1,105 @@
...
+    public void setUp() throws Exception {
+        dir = temporaryFolder.newFolder().toPath();
+    }
+
+    @Test
+    public void testReadFromTwoSegments() throws Exception {
+        RecordIOWriter writer = null;
+
+        for (int i = 0; i < 5; i++) {
+            Path segmentPath = dir.resolve(String.format(DeadLetterQueueWriteManager.SEGMENT_FILE_PATTERN, i));
```
talevy: done

talevy:
@jordansissel thank you for the review.

I have updated to reflect your comments, and also added some extra logic around adding the ability to seek to a specific Timestamp.

jordansissel:
Overall LGTM.

PR #6818, opened 2017-03-16
Bump to 5.0.3

suyograo:


Gemfile
```diff
@@ -48,7 +48,7 @@ gem "logstash-filter-fingerprint"
 gem "logstash-filter-geoip"
 gem "logstash-filter-grok"
 gem "logstash-filter-json"
-gem "logstash-filter-kv"
+gem "logstash-filter-kv", "< 4.0.0"
```
suyograo: Don't want to bring in major version here.

Gemfile
```diff
@@ -61,7 +61,7 @@ gem "logstash-filter-useragent"
 gem "logstash-filter-uuid"
 gem "logstash-filter-xml"
 gem "logstash-input-couchdb_changes"
-gem "logstash-input-elasticsearch"
+gem "logstash-input-elasticsearch", "< 4.0.1"
```
suyograo: Don't want to bring in major change here.

Issue #6619, opened 2017-02-01
env http_proxy setting ignored on logstash-plugin


pjanzen:
Hi,

While installing logstash-filter-elasticsearch behind a proxy I found that when I set http_proxy= it is being ignored during install and returns an error "connection refused"

After looking around @google I found that when I extend JRUBY_OPTS with -J-Dhttp.proxyHost=proxy.server -J-Dhttp.proxyPort=1234 I can install plugins from behind a proxy server.

Topic on the forum is here:
https://discuss.elastic.co/t/plugin-not-found/73472/3

- logstash 5.2.0
- Ubuntu 16.04.1 LTS
- openjdk version "1.8.0_121"

If you require any more information please let me know.

Regard,
Paul Janzen.

ph:
We need an integration test for that, I am pretty sure it was fixed in a jar-dependencies releases that  5.2.0 includes.

primeroz:
I am having the same problem,  reference to #6044 where this is supposed to be fixed 

* Centos 6 
* logstash 5.2 
* openjdk version "1.8.0_121"

HTTPS_PROXY environement variable is completely ignored
JRUBY_OPTS="-J-Dhttp.proxyHost=proxy -J-Dhttp.proxyPort=3128" manage to get the plugin installation started but as soon as the dependecy part start i can see it tried to connect directly to the internet on https and fail

`tcp        0      0 10.0.1.97:57734             10.0.1.254:3128             ESTABLISHED 19410/java
tcp        0      0 10.0.1.97:57736             10.0.1.254:3128             ESTABLISHED 19410/java
tcp        0      1 10.0.1.97:45970             151.101.16.215:443          SYN_SENT    20207/java
`

disaster37:
It's not working for me with a auth http proxy. I have added <username></username> and <password></password> on settings.xml and I have added -J-Dhttp.proxyUser= and -J-Dhttp.proxyPassword=

Any idea ?

camelx1:
primeroz, thx, it really helped me

shqiperih:
Just to confirm this workaround does not work on RHEL 6.7 -- set .m2 file set env variables  -- set JRUBY_OPTS still same errors as #6727. I'll have to go back to an earlier installation of logstash 2.3 with all plugins included in the tar file.  Any chance we could get a similar all plugins included newer version of logstash. Possibly in RPM form so we can start as a service

msvechla:
[...]

Issue #6528, opened 2017-01-12
Plugin installation behind forward proxy completely broken in 5.1.1


tkuther:
Just updated to Logstash version 5.1.1 - installed from elastic YUM repo on CentOS 6, Oracle JDK 1.8.
I do need the output-email plugin.

Reading the docs, the most obvious would be to export HTTP_PROXY and HTTPS_PROXY
```
[root@phsysint08 logstash]# export HTTP_PROXY=http://localhost:3128
[root@phsysint08 logstash]# export HTTPS_PROXY=http://localhost:3128
[root@phsysint08 logstash]# export http_proxy=http://localhost:3128
[root@phsysint08 logstash]# export https_proxy=http://localhost:3128
# curl -s -I https://artifacts.elastic.co
HTTP/1.1 200 Connection established
Connection: close

HTTP/1.1 302 Moved Temporarily
Content-Length: 169
Content-Type: text/html
Date: Thu, 12 Jan 2017 15:07:01 GMT
Location: https://www.elastic.co/downloads/
Server: nginx/1.4.6 (Ubuntu)
x-ngx-hostname: www01
Connection: keep-alive

[root@phsysint08 logstash]# DEBUG=1 bin/logstash-plugin install logstash-output-email
DEBUG: exec /usr/share/logstash/vendor/jruby/bin/jruby --1.9 /usr/share/logstash/lib/pluginmanager/main.rb install logstash-output-email
Looking if package named: logstash-output-email exists at https://artifacts.elastic.co/downloads/logstash-plugins/logstash-output-email/logstash-output-email-5.1.1.zip
Errno::ECONNREFUSED: Connection refused - Connection refused
```
^^ Proxy setting gets ignored

Next, try offline installation. Downloaded the .gem
```
[root@phsysint08 logstash]# DEBUG=1 bin/logstash-plugin install /tmp/logstash-output-email-4.0.3.gem 
DEBUG: exec /usr/share/logstash/vendor/jruby/bin/jruby --1.9 /usr/share/logstash/lib/pluginmanager/main.rb install /tmp/logstash-output-email-4.0.3.gem
Looking if package named: /tmp/logstash-output-email-4.0.3.gem exists at https://artifacts.elastic.co/downloads/logstash-plugins//tmp/logstash-output-email-4.0.3.gem//tmp/logstash-output-email-4.0.3.gem-5.1.1.zip
Errno::ECONNREFUSED: Connection refused - Connection refused
```

[...]

PR #6825, opened 2017-03-21
Allow Logstash to accept a PROXY configuration

ph:
Logstash's plugin manager will now follow proxy configuration from the environment.
If you configure `http_proxy` and `https_proxy` variable, the manager will now use them for all the ruby http connection and will also pass that information down to maven.

Fixes: #6619, #6528

### DIFF
--- a/docs/static/plugin-manager.asciidoc
+++ b/docs/static/plugin-manager.asciidoc
@@ -6,6 +6,18 @@ packages called gems and hosted on RubyGems.org. The plugin manager accessed via
 lifecycle of plugins in your Logstash deployment. You can install, remove and upgrade plugins using the Command Line
 Interface (CLI) invocations described below.
 
+[[http-proxy]]
+=== Proxy configuration
+
+The majority of the plugin manager commands require access to the internet to reach https://rubygems.org[RubyGems.org].
+If your organization is behind a firewall you can set these environments variables to configure Logstash to use your proxy.
+
+[source, shell]
+----------------------------------
+export http_proxy=http://localhost:3128
+export https_proxy=http://localhost:3128
+----------------------------------
+
 [float]
 [[listing-plugins]]
 === Listing plugins
--- a/qa/integration/specs/cli/http_proxy_install_spec.rb
+++ b/qa/integration/specs/cli/http_proxy_install_spec.rb
@@ -1,0 +1,65 @@
+# encoding: utf-8
+require_relative "../../framework/fixture"
+require_relative "../../framework/settings"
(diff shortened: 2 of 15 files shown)


ph:
I am waiting on the travis run, seem to be OK on my ubuntu VM.

ph:
This PR will also need windows testing, I believe the `Dir.home` will return the right thing depending of the platform and the `settings.xml` for maven will also work there.

ph:
I've restarted the Travis job, something weird happened.

docs/static/plugin-manager.asciidoc
```diff
@@ -6,6 +6,18 @@ packages called gems and hosted on RubyGems.org. The plugin manager accessed via
 lifecycle of plugins in your Logstash deployment. You can install, remove and upgrade plugins using the Command Line
 Interface (CLI) invocations described below.
 
+[[http-proxy]]
+=== Proxy configuration
+
+The majority of the plugin manager commands require to have direct access to the internet, if your organization is behind a firewall you can
```
suyograo: The majority of the plugin manager commands require access to the internet to reach RubyGems.org. If your organization is behind a firewall you can set these environments variables to configure Logstash to use your proxy.

suyograo:
Are these tests related to the fact we are using a `sudo` in travis? Do these test failures happen on master?

ph:
@suyograo Yup, only because we use `sudo`, we will leverage jenkins for that.

qa/integration/specs/cli/http_proxy_install_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+require_relative "../../framework/fixture"
+require_relative "../../framework/settings"
+require_relative "../../services/logstash_service"
+require_relative "../../services/http_proxy_service"
+require_relative "../../framework/helpers"
+require "logstash/devutils/rspec/spec_helper"
+require "stud/temporary"
+require "fileutils"
+
+# Theses tests doesn't currently work on Travis, since we need to run them in a sudo
```
ph: Until we move to jenkins

suyograo:
LGTM

jdlourenco:
Will this be fixed only in 5.4.0?

I'm using 5.3.0 and I'm only able to update logstash plugins behind a proxy using [this workaround](https://gist.github.com/devlatte/fda1c57ed7a33daea73c8b00799d74a1).

If that's the case I guess the documentation should be updated since it doesn't seem to work as described [here](https://www.elastic.co/guide/en/logstash/current/working-with-plugins.html).

PR #6826, opened 2017-03-21
Do not try to reach the artifact server when no plugins is given

ph:
This PR fix an annoyance when running the `bin/logstash-plugin install
--no-verify` without any plugins, the command was making an unnecessary
call to the artifacts web server.


ph:
This was discovered when working on #6825

suyograo:
LGTM

Issue #6608, opened 2017-01-31
JVM max heap size stats report is double?


jordansissel:
Background: https://discuss.elastic.co/t/api-showing-incorrect-data/73017/5

/_node/stats/jvm vs /_node/jvm -- heap_max_in_bytes is double in the /_node/stats/jvm call the same value in /_node/jvm

* I confess I'm not sure why `/_node/jvm` even exists as an API call? Shouldn't it just be "/_node/stats" ?

I patched Logstash 5.1.1 to see what's going on:

```diff
     def aggregate_information_for(collection)
       collection.reduce(default_information_accumulator) do |m,e|
         e = { e[0] => e[1] } if e.is_a?(Array)
+        p "----"
         e.each_pair do |k,v|
+          if k.include?("max")
+            p :aggregate_for => [ k, v], :max? => k.include?("max")
+          end
           m[:used_in_bytes] += v       if k.include?("used")
           m[:committed_in_bytes] += v  if k.include?("committed")
           m[:max_in_bytes] += v        if k.include?("max")

```

```
"----"
{:aggregate_for=>["usage.max", 1073741824], :max?=>true}
{:aggregate_for=>["peak.max", 1073741824], :max?=>true}
```
^^ I am not completely sure, but the above code `           m[:max_in_bytes] += v        if k.include?("max")` will double add `max_in_bytes` because `usage.max` and `peak.max` are both true results for `k.include?("max")`.

acchen97:
My understanding was that for heap_max_in_bytes the `_node/jvm` was max allocated heap size, and the `_node/stats/jvm` was max heap size used. The term is overloaded though, we should try to make it more clear what they mean.

primeroz:
I am having the same issue on logstash 5.2.1 with java-1.8.0-openjdk-1.8.0.121-0.b13.el6_8.x86_64 on centos 6.8 

In my case it **seems** to only happen when _Xms_ and _Xmx_ are set to the same value 

For example on one of my instances where i have different values ( _-Xmx2g -Xms1g_)
```
[...]

PR #6827, opened 2017-03-21
Fix the heap collecting values

ph:
This PR fixes an issue where the max heap size was reported as the double of
the actual value because it was merging the values of the usage.max and
peak.max into a single value.

Fixes: #6608


suyograo:
Tests?

ph:
@suyograo Any suggestion for writing theses tests? I could probably mock the beans and expect a good result.

ph:
nevermind, I will just write a simple unit test for that.

suyograo:
Nice. LGTM

PR #6791, opened 2017-03-02
Add Logstash configs for filebeat modules

suyograo:
Hand transpiled Logstash configs for FB modules to be included in
the docs for 5.3.

See FB modules for more details: https://github.com/elastic/beats/tree/master/filebeat/module

PR #6832, opened 2017-03-22
Add docs about configuring LS for use with Filebeat modules

dedemorton:
@suyograo Here's what I'm planning to say about Filebeat modules in the 5.3 Logstash docs. I have the LS configs commented out, but will add them after you merge your PR. Let me know if the messaging here is in line with what you expect for 5.3. 

IMO the manual configuration experience would be a lot easier if Filebeat allowed Logstash as an output. Instead of generating an error and closing, Filebeat should generate a warning when Logstash is configured as the output.  Of course, if  `-setup` is specified and ES is not configured as the output, Filebeat should generate an error.

Right now, FBM users who want to use Logstash instead of ingest node miss out on the benefit that FBMs provide for setting up the Filebeat config automatically. The only real benefit of using Filebeat modules at all for Logstash users (AFAIK) is that they can run the `-setup` command to get the dashboards. 

Note that I've added some questions to reviewers within the text of the document.

The Logstash configs that will be included in this doc are in this PR: https://github.com/elastic/logstash/pull/6791

### DIFF
--- a/docs/static/filebeat-modules.asciidoc
+++ b/docs/static/filebeat-modules.asciidoc
@@ -1,0 +1,321 @@
+[[filebeat-modules]]
+
+== Working with Filebeat Modules
+
+Starting with version 5.3, Filebeat comes packaged with pre-built 
+{filebeat}filebeat-modules.html[modules] that contain the configuration needed
+to read, parse, and visualize data from various log file formats, such as Nginx,
+Apache2, and MySQL. Each Filebeat module consists of one or more filesets that
+contain ingest node pipelines, Elasticsearch templates, Filebeat prospector
+configurations, and Kibana dashboards.
+
+Filebeat modules do not currently provide Logstash pipeline configurations.
+In the future, Filebeat modules will provide tighter integration with Logstash
+to offer you a more powerful alternative to using ingest node pipelines.
+For now, you can follow the steps in this section to configure Filebeat and
+build Logstash pipeline configurations that are equivalent to the ingest
+node pipelines available with the Filebeat modules.
+
+Then you'll be able to use the sample Kibana dashboards available with Filebeat
+to visualize your data in Kibana.
+
+NOTE: These manual steps will no longer be required when Logstash support
+is added to Filebeat modules in a future release.
+
+To build and run Logstash configurations that provide capabilities similar to
+Filebeat modules:
+
+. Load the Filebeat index pattern and sample Kibana dashboards.
++
+To do this, you need to run Filebeat with the Elasticsearch output enabled and
+specify the `-setup` flag. For example:
++
+[source,shell]
+----------------------------------------------------------------------
+./filebeat -e -setup -E "output.elasticsearch.hosts=["http://localhost:9200"]"
+----------------------------------------------------------------------
++
+A connection to Elasticsearch is required for this one-time setup step because
+Filebeat needs to create the index pattern and load the sample dashboards into the
+Kibana index. 
++
+After the dashboards are loaded, you'll see the message
++INFO Connected to Elasticsearch version {elasticsearch_version}+. You can ignore
+any `ERR Connecting error publishing events` messages and shut down Filebeat. 
+
+. Configure Filebeat to send log lines to Logstash.
++
+In version 5.3, Filebeat modules won't work when Logstash is configured as
+the output. Therefore you need to configure Filebeat to harvest lines from
+your log files and send them as events to Logstash. 
++
+See <<logstash-config-for-filebeat-modules>> for detailed examples.
+
+. Create a Logstash pipeline configuration that reads from the Beats input and
+parses the log events.
++
+See <<logstash-config-for-filebeat-modules>> for detailed examples.
+
+. Start Filebeat. For example, to start Filebeat in the foreground, use:
++
+[source,shell]
+----------------------------------------------------------------------
+sudo ./filebeat -e -c filebeat.yml -d "publish"
+----------------------------------------------------------------------
++
+See {filebeat}/filebeat-starting.html[Starting Filebeat] for more info.
+
+. Start Logstash, passing in the pipeline configuration file that parses the
+log. For example:
++
+[source,shell]
+----------------------------------------------------------------------
+bin/logstash -f mypipeline.conf
+----------------------------------------------------------------------
++
+You'll see the following message when Logstash is running and listening for
+input from Beats: 
++
+[source,shell]
+----------------------------------------------------------------------
+[2017-03-17T16:31:40,319][INFO ][logstash.inputs.beats    ] Beats inputs: Starting input listener {:address=>"127.0.0.1:5044"}
+[2017-03-17T16:31:40,350][INFO ][logstash.pipeline        ] Pipeline main started
+----------------------------------------------------------------------
+
+. To visualize the data in Kibana, launch the Kibana web interface by pointing
+your browser to port 5601. For example,
+http://127.0.0.1:5601[http://127.0.0.1:5601]. 
+
+[[logstash-config-for-filebeat-modules]]
+=== Configuration Examples
+
+The examples in this section show you how to configure Filebeat and build
+Logstash pipelines that parse:
+
+* <<parsing-apache2>>
+* <<parsing-mysql>>
+* <<parsing-nginx>>
+* <<parsing-system>>
+
+//REVIEWERS: Do we want to add an example that shows how to conditionally select the grok pattern? If not, what guidance should we provide to help users understand how to build a config that works with more than one type of log file?
+
+Of course, the paths that you specify in the Filebeat config depend on the location
+of the logs you are harvesting. The examples show common default locations.
+
+[[parsing-apache2]]
+==== Apache 2 Logs
+
+Here are some configuration examples for shipping and parsing Apache 2 access and
+error logs.
+
+===== Access Logs
+
+// Reviewers: I could provide separate Filebeat config examples for each OS, but I think that might be overkill. WDYT? There's already a bit of repetition here, but worth it IMO to enable copy/paste.
+
+Example Filebeat config:
+
+[source,yml]
+----------------------------------------------------------------------
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/apache2/access.log* 
+    - /var/log/apache2/other_vhosts_access.log*
+  exclude_files: [".gz$"]
+output.logstash:
+  hosts: ["localhost:5044"]
+----------------------------------------------------------------------
+
+
+//REVIEWERS: When testing these configs, I've used a path to a local test file, so please confirm that the log files located at these paths can be parsed given the specified LS config.
+
+Example Logstash pipeline config:
+
+[source,json]
+----------------------------------------------------------------------------
+#include::filebeat_modules/apache2/access/pipeline.conf[]
+----------------------------------------------------------------------------
+
+===== Error Logs
+
+Example Filebeat config:
+
+[source,yml]
+----------------------------------------------------------------------
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/apache2/error.log*
+  exclude_files: [".gz$"]
+output.logstash:
+  hosts: ["localhost:5044"]
+----------------------------------------------------------------------
+
+
+Example Logstash pipeline config:
+
+[source,json]
+----------------------------------------------------------------------------
+#include::filebeat_modules/apache2/error/pipeline.conf[]
+----------------------------------------------------------------------------
+
+[[parsing-mysql]]
+==== MySQL Logs
+
+Here are some configuration examples for shipping and parsing MySQL error and
+slowlog logs.
+
+===== Error Logs
+
+Example Filebeat config:
+
+[source,yml]
+----------------------------------------------------------------------
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/mysql/error.log*
+    - /var/log/mysqld.log*
+  exclude_files: [".gz$"]
+output.logstash:
+  hosts: ["localhost:5044"]
+----------------------------------------------------------------------
+
+
+Example Logstash pipeline config:
+
+[source,json]
+----------------------------------------------------------------------------
+#include::filebeat_modules/mysql/error/pipeline.conf[]
+----------------------------------------------------------------------------
+
+===== Slowlog
+
+Example Filebeat config:
+
+[source,yml]
+----------------------------------------------------------------------
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/mysql/mysql-slow.log*
+    - /var/lib/mysql/hostname-slow.log
+  exclude_files: [".gz$"]
+output.logstash:
+  hosts: ["localhost:5044"]
+----------------------------------------------------------------------
+
+
+Example Logstash pipeline config:
+
+[source,json]
+----------------------------------------------------------------------------
+#include::filebeat_modules/mysql/slowlog/pipeline.conf[]
+----------------------------------------------------------------------------
+
+[[parsing-nginx]]
+==== Nginx Logs
+
+Here are some configuration examples for shipping and parsing Nginx access and
+error logs.
+
+===== Access Logs
+
+Example Filebeat config:
+
+[source,yml]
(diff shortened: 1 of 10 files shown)


docs/static/filebeat-modules.asciidoc
```diff
@@ -0,0 +1,321 @@
...
+== Working with Filebeat Modules
+
+Starting with version 5.3, Filebeat comes packaged with pre-built 
+{filebeat}filebeat-modules.html[modules] that contain the configuration needed
+to read, parse, and visualize data from various log file formats, such as Nginx,
+Apache2, and MySQL. Each Filebeat module consists of one or more filesets that
+contain ingest node pipelines, Elasticsearch templates, Filebeat prospector
+configurations, and Kibana dashboards.
+
+Filebeat modules do not currently provide Logstash pipeline configurations.
```
suyograo: Can we add a sub-section called "Migrating from ingest pipelines". In this section, we can add context on why a user would want to graduate to using LS instead of the ingest pipelines. The following reasons are all valid:

* Use multiple outputs. Ingest was designed to only support ES as an output. For example, users may want to archive their incoming data to S3 as well as indexing it in ES.
* Use the persistent queue feature to handle spikes when ingesting data (from beats and other sources).
* Take advantage of the richer transformation capabilities in Logstash.

docs/static/filebeat-modules.asciidoc
```diff
@@ -0,0 +1,321 @@
...
+In the future, Filebeat modules will provide tighter integration with Logstash
+to offer you a more powerful alternative to using ingest node pipelines.
+For now, you can follow the steps in this section to configure Filebeat and
+build Logstash pipeline configurations that are equivalent to the ingest
+node pipelines available with the Filebeat modules.
+
+Then you'll be able to use the sample Kibana dashboards available with Filebeat
```
suyograo: How about changing sample Kibana -> same Kibana? 

By providing the LS configs, we are telling users they can continue to use the dashboards set up by the filebeat modules

docs/static/filebeat-modules.asciidoc
```diff
@@ -0,0 +1,321 @@
...
+[source,shell]
+----------------------------------------------------------------------
+./filebeat -e -setup -E "output.elasticsearch.hosts=["http://localhost:9200"]"
+----------------------------------------------------------------------
++
+A connection to Elasticsearch is required for this one-time setup step because
+Filebeat needs to create the index pattern and load the sample dashboards into the
+Kibana index. 
++
+After the dashboards are loaded, you'll see the message
```
suyograo: Hmm this error is weird.. Does this happen only with LS output enabled?

docs/static/filebeat-modules.asciidoc
```diff
@@ -0,0 +1,321 @@
...
++INFO Connected to Elasticsearch version {elasticsearch_version}+. You can ignore
+any `ERR Connecting error publishing events` messages and shut down Filebeat. 
+
+. Configure Filebeat to send log lines to Logstash.
++
+In version 5.3, Filebeat modules won't work when Logstash is configured as
```
suyograo: I think this line is confusing. We should just remove it. 

I guess from this point on -- once you've run setup from FBM -- you are using FB as a regular shipper to LS. We should say, "remove the ES output from filebeat.yml and add LS as an output". 

```
output.logstash:
  hosts: ["localhost:5044"]
```

That should just work then.

docs/static/filebeat-modules.asciidoc
```diff
@@ -0,0 +1,321 @@
...
+
+The examples in this section show you how to configure Filebeat and build
+Logstash pipelines that parse:
+
+* <<parsing-apache2>>
+* <<parsing-mysql>>
+* <<parsing-nginx>>
+* <<parsing-system>>
+
+//REVIEWERS: Do we want to add an example that shows how to conditionally select the grok pattern? If not, what guidance should we provide to help users understand how to build a config that works with more than one type of log file?
```
suyograo: I am not sure why we need to talk about conditionals here. Can you explain more? Each FBM handles a different log type (mysql, apache2, ..) , so we would do the same in LS as well, no?

docs/static/filebeat-modules.asciidoc
```diff
@@ -0,0 +1,321 @@
...
+Filebeat modules:
+
+. Load the Filebeat index pattern and sample Kibana dashboards.
++
+To do this, you need to run Filebeat with the Elasticsearch output enabled and
+specify the `-setup` flag. For example:
++
+[source,shell]
+----------------------------------------------------------------------
+./filebeat -e -setup -E "output.elasticsearch.hosts=["http://localhost:9200"]"
```
suyograo: Is this missing `-modules=nginx` flag?

docs/static/filebeat-modules.asciidoc
```diff
@@ -0,0 +1,321 @@
...
+
+===== Access Logs
+
+// Reviewers: I could provide separate Filebeat config examples for each OS, but I think that might be overkill. WDYT? There's already a bit of repetition here, but worth it IMO to enable copy/paste.
+
+Example Filebeat config:
+
+[source,yml]
+----------------------------------------------------------------------
+filebeat.prospectors:
```
suyograo: Do we need to specify the `filebeat.yml` here? If a user has already done `./filebeat --module=nginx` they should have the relevant `filebeat.yml` file, right?

suyograo:
@dedemorton @acchen97 I'm a bit confused on how to message this to our users. 

IMO, if someone is moving from `FBM -> ES` to `FBM -> LS -> ES`, I would consider this as an advanced user. I think at this point, we should assume that modules are already setup, dashboards are already available in ES, but the user now wants to use LS as the transform engine instead of ingest. If we think about it this way, we could just ask the user to start `./filebeat` with LS output and provide documentation on how to use LS config equivalents of ingest.

docs/static/filebeat-modules.asciidoc
```diff
@@ -0,0 +1,321 @@
+[[filebeat-modules]]
+
+== Working with Filebeat Modules
+
+Starting with version 5.3, Filebeat comes packaged with pre-built 
+{filebeat}filebeat-modules.html[modules] that contain the configuration needed
```
acchen97: Myabe "configurations needed to collect, parse, enrich, and visualize data from various log file formats out-of-the-box"?

docs/static/filebeat-modules.asciidoc
acchen97: +1, I actually prefer the term "Graduating" in this context.

For the last bullet, we can extend to say "like external lookups."

docs/static/filebeat-modules.asciidoc
acchen97: +1, the point is its a "seamless" transition.

docs/static/filebeat-modules.asciidoc
acchen97: Can we maybe have Logstash just define this index mapping and pattern on LS startup?

docs/static/filebeat-modules.asciidoc
acchen97: +1

docs/static/filebeat-modules.asciidoc
acchen97: I would think so, if they already set the LS output explicitly.

acchen97:
@dedemorton thanks for this, left a few comments, but looks good overall. We should link off to these docs from the FBM documentation as well.

docs/static/filebeat-modules.asciidoc
dedemorton: @suyograo That's how things should work, but that's not how they work right now. In 5.3, if you run a Filebeat module with the logstash output enabled, you see the error `CRIT Exiting: Filebeat modules configured but the Elasticsearch output is not configured/enabled`:

`````
Rhodas-MacBook-Pro:filebeat-5.3.0-darwin-x86_64 dedemorton$ sudo ./filebeat -e -modules=system
2017/03/23 17:03:16.622482 beat.go:285: INFO Home path: [/Users/dedemorton/BuildTesting/5.3.0_e5a8c674/filebeat-5.3.0-darwin-x86_64] Config path: [/Users/dedemorton/BuildTesting/5.3.0_e5a8c674/filebeat-5.3.0-darwin-x86_64] Data path: [/Users/dedemorton/BuildTesting/5.3.0_e5a8c674/filebeat-5.3.0-darwin-x86_64/data] Logs path: [/Users/dedemorton/BuildTesting/5.3.0_e5a8c674/filebeat-5.3.0-darwin-x86_64/logs]
2017/03/23 17:03:16.622515 beat.go:186: INFO Setup Beat: filebeat; Version: 5.3.0
2017/03/23 17:03:16.622554 metrics.go:23: INFO Metrics logging every 30s
2017/03/23 17:03:16.622572 logstash.go:90: INFO Max Retries set to: 3
2017/03/23 17:03:16.622625 outputs.go:108: INFO Activated logstash as output plugin.
2017/03/23 17:03:16.622682 publish.go:295: INFO Publisher name: Rhodas-MacBook-Pro.local
2017/03/23 17:03:16.622771 async.go:63: INFO Flush Interval set to: 1s
2017/03/23 17:03:16.622778 async.go:64: INFO Max Bulk Size set to: 2048
2017/03/23 17:03:16.623475 beat.go:221: INFO filebeat start running.
2017/03/23 17:03:16.623535 metrics.go:51: INFO Total non-zero values: 
2017/03/23 17:03:16.623542 metrics.go:52: INFO Uptime: 4.588856ms
2017/03/23 17:03:16.623546 beat.go:225: INFO filebeat stopped.
2017/03/23 17:03:16.623551 beat.go:339: CRIT Exiting: Filebeat modules configured but the Elasticsearch output is not configured/enabled
Exiting: Filebeat modules configured but the Elasticsearch output is not configured/enabled
`````

That's why I've outlined the steps the way I have. TBH, I think the workflow is kind of jenky overall.

File: docs/static/filebeat-modules.asciidoc (hunk 3 above)
Comment (dedemorton) [inline docs/static/filebeat-modules.asciidoc]: @suyograo You can't use the `-setup` step with logstash configured as output because the dashboards need to be loaded into an Elasticsearch instance.

File: docs/static/filebeat-modules.asciidoc (hunk 4 above)
Comment (dedemorton) [inline docs/static/filebeat-modules.asciidoc]: See earlier comment. The problem here is the configuration options get set dynamically when you run the Filebeat modules (I don't think the config for the module is persisted anywhere). So you can just run a module once to get Filebeat configured. If you want to use Logstash as the output in 5.3, my understanding is that you need to configure Filebeat, too.  @tsg Can you confirm that this is true and that I'm not misunderstanding something fundamental here?

File: docs/static/filebeat-modules.asciidoc (hunk 5 above)
Comment (dedemorton) [inline docs/static/filebeat-modules.asciidoc]: @suyograo For example,  let's say that a user wants to build a Logstash pipeline that parses both apache2 error and access logs. Wouldn't they need conditional logic or something similar to do that? Right now, we show a separate Logstash config for each fileset. 

See Tudor's comment here: https://github.com/elastic/logstash/issues/6542#issuecomment-279364410

File: docs/static/filebeat-modules.asciidoc (hunk 7 above)
Comment (dedemorton) [inline docs/static/filebeat-modules.asciidoc]: @suyograo I don't think it works that way. The `filebeat.yml` isn't updated when you run the module. The settings are determined at runtime based on what's in the manifest file (and what the user passes as variables on the command line). @tsg Can you confirm? It's possible that I've missed something big here. :-)

File: docs/static/filebeat-modules.asciidoc (hunk 8 above)
Comment (dedemorton) [inline docs/static/filebeat-modules.asciidoc]: Sure, but I'll remove "out-of-the-box" because it's grammatically ambiguous (could be misread--and mistranslated--as modifying "log file formats"

File: docs/static/filebeat-modules.asciidoc (hunk 6 above)
Comment (tsg) [inline docs/static/filebeat-modules.asciidoc]: I think @suyograo is right here, the `-modules=nginx` flag is needed and works because this is specifically meant to use the ES output (`-E "output.elasticsearch.hosts=["http://localhost:9200"]"`).

File: docs/static/filebeat-modules.asciidoc (hunk 7 above)
Comment (tsg) [inline docs/static/filebeat-modules.asciidoc]: Yeah, this is needed for 5.3, but hopefully we'll be able to simplify it with 5.4. I'd say a single `filebeat.yml` example per module is enough.

Comment (suyograo) [inline docs/static/filebeat-modules.asciidoc]: I had missed this in my understanding of how FBM works. Agree with what Dede has here, and making it better for 5.4

File: docs/static/filebeat-modules.asciidoc [hunk 9]
```diff
@@ -0,0 +1,321 @@
...
+Example Filebeat config:
+
+[source,yml]
+----------------------------------------------------------------------
+filebeat.prospectors:
+- input_type: log
+  paths:
+    - /var/log/mysql/mysql-slow.log*
+    - /var/lib/mysql/hostname-slow.log
+  exclude_files: [".gz$"]
```
tsg: Here the multiline configuration is missing. You can take it from: https://github.com/elastic/beats/blob/master/filebeat/module/mysql/slowlog/config/slowlog.yml#L7

It would be good if these samples would resemble what FBM do as much as possible. I can help with that if want.

docs/static/filebeat-modules.asciidoc
dedemorton: whoops...you're right. I was looking at this out of context. :-)

docs/static/filebeat-modules.asciidoc
dedemorton: @tsg I think the multiline is defined in the Logstash config. You're at a bit of a disadvantage because you can't see the Logstash configs in this review (Suyog is still working on them). They are in this PR: https://github.com/elastic/logstash/pull/6791.

docs/static/filebeat-modules.asciidoc
dedemorton: My mistake. This error happens only when you have Elasticsearch and Logstash configured as outputs. I'll removed that sentence from the topic.

dedemorton:
@suyograo I've updated the Filbeat configs, but I still don't think this is 100% there. 

Here's a list of the problems that I'm seeing: 

The apache access log darwin-2.4.23.log still returns "_grokparsefailure"  tag (works ok on ubuntu-2.2.22.log, tho)

The nginx error log seems to parse OK, but in the dashboard I under "Nginx Errors over time," I don't see any data, just the message: "Area charts require more than one data point. Try adding an X-Axis Aggregation" (Not sure this has anything to do with LS parsing, tho. I might have seen this with the FBMs.)

The system auth log auth-ubuntu1204.log returns  "_grokparsefailure"  and _geoip_lookup_failure tags. I don't have time right now to figure out if I've done something wrong in my test files....

Didn't have time to test the syslog because I got Kibana into some weird state where it wasn't showing any data.

dedemorton:
@suyograo I did a bit more testing of the LS configs today. I didn't want to merge this while you were away, but I'd like to get it merged soon. For the problems I listed in the previous comment:

- For 1 and 3, I think the LS config is OK because it does work on some of the logs. I assume I'm getting geoIP lookup failures because I'm using localhost, so it's probably OK to ignore those failures.
- For 2, I think the problem is the dashboard, not the LS config. But what do I know? WDYT?
- For 4, not sure what to do. Maybe you or @tsg can take a look and see if the multiline config in Filebeat is wrong. I'm pretty sure I copied the config that's used in the Filebeat module.

One other thing: I think the setup step is a little hacky. This step may actually result in a filebeat index getting created for system events...because the default Filebeat config harvests the system log by default. I'm not sure if that matters...or whether I should mention it. But I think some people might wonder why the system info is being captured. If I comment out that section in the filebeat.yml, I think the setup still happens, but Filebeat stops with a critical error. That's not necessarily bad because we want users to stop Filebeat anyhow, but it still seems kind of hacky. Wondering what you and Tudor think about this.

suyograo:
Apologies for the delay; I'll re-run this today and update here.

suyograo:
I tested the syslog config and it does not work as you mentioned. I don't know what the issue is here since the multiline config looks ok. I am thinking we should merge the PR as is which will give users some "template" to use for migration and we can iterate on it. I have a feeling this is one of the differences between the grok implementation in ES and LS

suyograo:
LGTM

PR #6836, opened 2017-03-23
Introduce Dead Letter Queue to Logstash API

talevy:
This PR introduces the DLQ core library from #6817 into Logstash Ruby Core in three forms

- Settings for managing the dead letter queue
- APIs for monitoring the dead letter queue
- Addition of the `dlq_commit(event, reason)` method to the Plugin API


Gemfile
```diff
@@ -116,3 +116,26 @@ gem "logstash-output-tcp"
 gem "logstash-output-udp"
 gem "logstash-output-webhdfs"
 gem "logstash-filter-dissect"
+gem "logstash-filter-multiline", :path => "~/Code/es/logstash-filter-multiline"
```
talevy: still need to clean these two up

logstash-core/lib/logstash/plugin.rb
```diff
@@ -122,12 +123,22 @@ def metric
...
   # return the configured name of this plugin
   # @return [String] The name of the plugin defined by `config_name`
   def config_name
     self.class.config_name
   end
 
+  # commit event to dlq
+  def dlq_commit(event, reason)
+    if @dlq_manager.nil?
+      raise ArgumentError, "DLQ is not enabled, plugin cannot commit to it"
```
jordansissel: I wonder ... how should plugin authors handle this error? What action should be taken if I try to dlq something and this ArgumentError happens?

logstash-core/lib/logstash/plugin.rb
```diff
@@ -122,12 +123,22 @@ def metric
...
   end
+
   # return the configured name of this plugin
   # @return [String] The name of the plugin defined by `config_name`
   def config_name
     self.class.config_name
   end
 
+  # commit event to dlq
+  def dlq_commit(event, reason)
```
jordansissel: I'm not convinced that dlq_commit needs to be available for all plugin types (input, filter, codec, output). I feel like they should only happen for things after the queue?

For example, can you tell me a story about when an input would want to DLQ (At the time of input, we often don't have an event to actually dlq; same goes for codec when we only have bytes, not an event)

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -291,10 +294,10 @@ def each(&blk)
         # TODO: for https://github.com/elastic/logstash/issues/6055 = will have to properly refactor
         @iterating = true
         @originals.each do |e, _|
-          blk.call(e) unless e.cancelled?
+          blk.call(e) unless (e.respond_to?(:cancelled?) && e.cancelled?)
```
jordansissel: Do we have events that don't respond to `cancelled?` ?

docs/static/settings-file.asciidoc
```diff
@@ -134,6 +134,14 @@ The `logstash.yml` file includes the following settings:
 | The interval in milliseconds when a checkpoint is forced on the head page when persistent queues are enabled (`queue.type: persisted`). Specify `queue.checkpoint.interval: 0` for no periodic checkpoint.
 | 1000
 
+| `dead_letter_queue.enable`
+| Flag to instruct Logstash to enable the DLQ feature supported by plugins.
+| `false`
+
+| `path.dql`
```
jordansissel: typo? `path.dead_letter_queue`, right?

logstash-core/lib/logstash/api/commands/dlq.rb
```diff
@@ -0,0 +1,21 @@
...
+require_relative "hot_threads_reporter"
+
+module LogStash module Api module Commands
+  class Dlq < Commands::Base
+    def initialize(service = LogStash::Api::Service.instance)
+      super(service)
+      @pipeline_id, @pipeline = @service.agent.pipelines.first
+    end
+
+    def list
```
jordansissel: Can you tell me the intent of this API?

talevy:
thank you for the review @jordansissel! this one is a bit more half-baked than the recordio one 😄, it was basically my original PR with a bunch of stuff removed. I'll tend to your comments shortly!

logstash-core/lib/logstash/plugin.rb
talevy: I was working with someone the other day that wanted to dead-letter-queue events that could not be deserialized using the avro-codec upon input from kafka. So that would be an example of the input being able to DLQ.

That being said, we may not want to handle such a scenario, and I am cool with that!

logstash-core/lib/logstash/plugin.rb
suyograo: Shouldn't this be a no-op (with logging) if the DLQ is not enabled?

logstash-core/lib/logstash/plugin.rb
suyograo: I see the point about codecs and see it useful (depending on the codec). If the codec is complicated and throws an error we may want to DLQ this. That raises the question, if the event is not fully formed, what then?

I'd say we leave this in here for now since a codec has to actually implement using the DLQ.

Gemfile
```diff
@@ -116,3 +116,26 @@ gem "logstash-output-tcp"
...
+gem "jdbc-derby", ">= 0"
+gem "coveralls", ">= 0"
+gem "addressable", "~> 2.3.0"
+gem "elasticsearch", ">= 0"
+gem "sinatra", ">= 0"
+gem "webrick", ">= 0"
+gem "poseidon", ">= 0"
+gem "snappy", ">= 0"
+gem "webmock", "~> 1.21.0"
+>>>>>>> initial dlq
```
suyograo: merge remnants

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
talevy: I'll see if I can revert this. may have been due to some rash mocking decisions

talevy: ah, now I know what it may have been. Shutdown signal events. I'll clean this up

config/logstash.yml
```diff
@@ -131,6 +131,15 @@
 #
 # queue.checkpoint.interval: 1000
 #
+# ------------ Dead-Letter Queue Settings --------------
+# Flag to turn on dead-letter queue.
+#
+# dead_letter_queue.enable: false
+#
+# If using dead_letter_queue.enable: true, the directory path where the data files will be stored.
```
suyograo: We should use the same strategy as PQ configuration. Mention in the comment here that DLQ default is `path.data/dead_letter_queue` and leave the setting value empty here.

See https://github.com/elastic/logstash/blob/master/config/logstash.yml#L96

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -41,6 +41,10 @@ def register
     @strategy.register
   end
 
+  def do_register(dlq_manager=nil)
```
suyograo: Can we add a comment here why we are doing this in a separate method vs in the constructor? Will be helpful context.

logstash-core/spec/logstash/codecs/base_spec.rb
```diff
@@ -8,6 +8,9 @@ class LogStash::Codecs::NOOPAsync < LogStash::Codecs::Base
   attr_reader :last_result
   config_name "noop_async"
 
+  def register
```
suyograo: Do we need this?

logstash-core/spec/logstash/codecs/base_spec.rb
```diff
@@ -17,6 +20,9 @@ class LogStash::Codecs::NOOPSync < LogStash::Codecs::Base
   attr_reader :last_result
   config_name "noop_sync"
 
+  def register
```
suyograo: Do we need this?

logstash-core/spec/logstash/codecs/base_spec.rb
```diff
@@ -26,6 +32,9 @@ class LogStash::Codecs::NOOPMulti < LogStash::Codecs::Base
   attr_reader :last_result
   config_name "noop_multi"
 
+  def register
```
suyograo: Do we need this?

logstash-core/spec/logstash/filter_delegator_spec.rb
```diff
@@ -1,4 +1,4 @@
-# encoding: utf-8
+# encodingrspec ./logstash-core/spec/logstash/output_delegator_spec.rb:36: utf-8
```
suyograo: Why this change?

logstash-core/spec/logstash/pipeline_dlq_commit_spec.rb
```diff
@@ -0,0 +1,138 @@
+# encoding: utf-8
+require "spec_helper"
+require "tmpdir"
+
+class CommittingInput < LogStash::Inputs::Base
```
suyograo: Maybe call this `DLQCommittingInput`, `DLQComittingFilter` etc?

logstash-core/spec/logstash/filter_delegator_spec.rb
talevy: vim mistype

config/logstash.yml
talevy: will do

talevy: done

logstash-core/lib/logstash/output_delegator.rb
talevy: maybe the better place would be in the Plugin class, but I agree. I'll add this

suyograo:
@guyboertje can you take a look at this PR?

logstash-core/lib/logstash/plugin.rb
jordansissel: > could not be deserialized using the avro-codec upon input from kafka

In this scenario, I would say we should reject this and leave it in kafka? 

My thought with the DLQ is that we put undeliverable messages into it with the idea that human intervention could make a correction and it could then be delivered. With an avro decoding failure (because someone is sending garbage bytes, maybe?), I don't know what kind of human intervention we could expect to resolve this?

jordansissel: I would still prefer `dlq_commit` not be given to all plugins.

talevy:
Discussed this offline with @suyograo and @jordansissel.

Closing in favor of implementing the DLQ management as more of a self-service fashion, where one would request an instance of a queue from the Logstash runtime rather than being provided one. This move will enable exposing the DLQ to only those that need it, rather than all plugins even if they do not use it.

This PR has a few extra fixes (unrelated to the API), and will be followed-up in another PR

PR #6840, opened 2017-03-24
Create an offline pack on Windows generate corrupted pack

ph:
When generating a pack on windows, Paquet was downloading the gems in plain text instead of binary mode, this was making them corrupted and the rubygems class was not able to unpack them correctly. This PR change the File open mode to be **wb** instead of **w**.

## Todo

- [ ] Release a new version of paquet.
- [ ] Make sure Logstash uses that version in a future release.


suyograo:
LGTM

PR #6843, opened 2017-03-27
Release notes and breaking changes for 5.3

suyograo:
We usually avoid breaking changes in point releases, but unfortunately, there is one in PQ -- which is a beta feature, so breaking change is OK. Second, there is a breaking change in a plugin (again, this is something we avoid at best). Hence, I added breaking changes section per release.

### DIFF
--- a/docs/static/releasenotes.asciidoc
+++ b/docs/static/releasenotes.asciidoc
@@ -10,9 +10,9 @@ This section summarizes the changes in the following releases:
 
 
 * Persistent queues:
-    ** Changed the default queue location on disk to include the pipeline's ID as it's parent directory.
-       By default, the queue now gets created under path.data/queue/main. This breaking change was made in
-       anticipation of an upcoming feature where multiple, isolated pipelines could be run on the same Logstash
+    ** Changed the default queue location on disk to include the pipeline's ID in the path hierarchy.
+       By default, the queue is now created under <path.data>/queue/main. This breaking change was made to
+       accommodate an upcoming feature where multiple, isolated pipelines could be run on the same Logstash
        instance.
     ** Added a _recovery_ process that runs when Logstash starts up to recover data written to the persistent
        queue, but not yet checkpoint'd. This is useful in a situation when the input had written data to the queue,
@@ -44,11 +44,11 @@ This section summarizes the changes in the following releases:
 
 *`KV`*:
 
-* Breaking: `trim` and `trimkey` options are renamed to `trim_value` and `trim_key`.
+* Breaking: `trim` and `trimkey` options are renamed to `trim_value` and `trim_key` respectively (https://github.com/logstash-plugins/logstash-filter-kv/issues/10[Issue 10]).
 * `trim_value` only removes the specified leading and trailing characters from the value. Similarly, `trim_key`
   only removes the specified leading and trailing characters from the key (https://github.com/logstash-plugins/logstash-filter-kv/issues/10[Issue 10]).
 * Added new options `remove_char_value` and `remove_char_key` to remove the specified characters from keys
-  (or values) regardless of where these characters are found.
+  (or values) regardless of where these characters are found (https://github.com/logstash-plugins/logstash-filter-kv/issues/10[Issue 10]).
 
 *`Grok`*:
 
--- a/docs/static/releasenotes.asciidoc
+++ b/docs/static/releasenotes.asciidoc
@@ -11,18 +11,18 @@ This section summarizes the changes in the following releases:
 
 * Persistent queues:
     ** Changed the default queue location on disk to include the pipeline's ID in the path hierarchy.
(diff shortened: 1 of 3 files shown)


docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+
+* Persistent queues:
+    ** Changed the default queue location on disk to include the pipeline's ID in the path hierarchy.
+       By default, the queue is now created under <path.data>/queue/main. This breaking change was made to
+       accommodate an upcoming feature where multiple, isolated pipelines could be run on the same Logstash
+       instance.
+    ** Added a _recovery_ process that runs when Logstash starts up to recover data written to the persistent
+       queue, but not yet checkpoint'd. This is useful in a situation when the input had written data to the queue,
+       but Logstash crashed before writing to the checkpoint file.
+    ** Added exclusive access to the persistent queue on disk, as defined by the path.queue setting. Using a file
```
dedemorton: path.queue should be wrapped back ticks.

dedemorton: Suggest changing this to say, "Using a file lock guards against corruption by ensuring that only a single Logstash instance has access to write to the queue on the same path."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+       lock we ensure that only a single Logstash instance has access to write to the queue on the same path, to
+       guard against corruption ({lsissue}6604[Issue 6604]).
+    ** You can now safely reload the pipeline config when using persistent queues. Previously, the reloading of
```
dedemorton: Change to say, "Previously, reloading the config could result...."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+       config could result in data corruption.  In 5.3, the reload sequence has been changed to reliably shutdown the
```
dedemorton: Change to: "...has been changed to reliably shut down the..."
shut down is the verb form
shutdown is the noun form

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+       first pipeline before a new one is started with same settings.
```
dedemorton: Change to: "...with the same settings."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+    ** Fixed an issue where Logstash would stop accepting new events when queue capacity is reached even though events
+       were successfully acknowledged ({lsissue}6626[Issue 6626]).
+
+* Fixed a warning message when --config.debug is used with --log.level=debug  ({lsissue}6256[Issue 6256]).
```
dedemorton: Would change this to say: "Fixed a warning message that is displayed when...."

I'd also wrap `--config.debug` is used with `--log.level=debug` in back ticks.

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+[float]
+==== Input Plugins
+
+*`S3`*:
+
+* We now include the S3 key information in the metadata (https://github.com/logstash-plugins/logstash-input-s3/issues/105[Issue 105]).
+
+*`Unix`*:
+
+* Do not overwrite `host` and `path` fields if they are already provided by `add_field` config.
```
dedemorton: Maybe this should say, "The `host` and `path` fields are no longer overwritten if they are already provided by `add_field` config."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+
+==== Filter Plugins
+
+*`KV`*:
+
+* Breaking: `trim` and `trimkey` options are renamed to `trim_value` and `trim_key` respectively (https://github.com/logstash-plugins/logstash-filter-kv/issues/10[Issue 10]).
```
dedemorton: The `trim`...

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,66 @@
...
+==== Output Plugins
+
+*`S3`*:
+
+* Fixed to use the correct `signature_version` for the SDK v2 library (https://github.com/logstash-plugins/logstash-output-csv/issues/129[Issue 129]).
+* Fixed an issue which resulted in uploading empty files to S3 when using gzip compression (https://github.com/logstash-plugins/logstash-output-s3/issues/95[Issue 95]).
+
+*`CSV`*:
+
+* Updated to work with 5.0 event API and threading contracts (https://github.com/logstash-plugins/logstash-output-csv/issues/10[Issue 10]).
```
dedemorton: minor: "Updated to work with the 5.0 event..."

PR #6845, opened 2017-03-27
Fix typo in logstash core

dedemorton:
This gets propagated into a bunch of files in the plugin docs. :-(

Let me know if I also need to bump the version in the gemspec. Wasn't sure if that would be required....


ph:
LGTM

PR #6852, opened 2017-03-29
Cleanup some RecordIO cases

talevy:
- rename segment files to `%d.log` instead of having a length of 20 characters
- rename test file
- check for `null` events when seeking by timestamp, meaning there was no data found and we reached the end of the block that was supposed to have the event according to some timestamp comparison.


suyograo:
LGTM

jordansissel:
LGTM

PR #6860, opened 2017-03-30
Search java executable in %PATH%

manni19682:
In their latest JDK 8u121 (maybe earlier, i didn't checked that) Oracle changed the java installation on windows a bit. Searching the java executable via the link `%ProgramData%\Oracle\java\javapath\java.exe` unfortunately doesn't work any more. I replaced that by searching the java executable in the PATH as it is already done in the elasticsearch start scripts.

### DIFF
--- a/bin/setup.bat
+++ b/bin/setup.bat
@@ -15,11 +15,11 @@ goto finally
 
 :setup_jruby
 REM setup_java()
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

PR #6861, opened 2017-03-30
Make sure prepare_offline_pack command only accept a filename

ph:
This PR changes the behavior of the command when you were specify a directory
instead of a filename, if the directory exists it would have deleted the
directory and the files in it.

The new flow and validation is now safer, if you specify a directory
Logstash will tell you to specify a complete filename.

If the ZIP extension is missing, the command line will ask you to make
sure you target a right filename.

If the output file already exist it will *not* delete it but instead
will return a warning message, you can force a delete by using the
`--overwrite` option.

### DIFF
--- a/docs/static/offline-plugins.asciidoc
+++ b/docs/static/offline-plugins.asciidoc
@@ -28,14 +28,15 @@ access the Internet.
 +
 [source, shell]
 -------------------------------------------------------------------------------
-bin/logstash-plugin prepare-offline-pack --output OUTPUT [PLUGINS]
+bin/logstash-plugin prepare-offline-pack --output OUTPUT [PLUGINS] --overwrite
 -------------------------------------------------------------------------------
 +
 where:
 +
-* `OUTPUT` specifies the location where the compressed plugin pack will be written. The default location is
-+/LOGSTASH_HOME/logstash-offline-plugins-{logstash_version}.zip+.
+* `OUTPUT` specifies the zip file where the compressed plugin pack will be written. The default file is
++/LOGSTASH_HOME/logstash-offline-plugins-{logstash_version}.zip+. If you are using 5.2.x and 5.3.0, this location should be a zip file whose contents will be overwritten.
 * `[PLUGINS]` specifies one or more plugins that you want to include in the pack.
+* `--overwrite` specifies if you want to override an existing file at the location
 
 Examples:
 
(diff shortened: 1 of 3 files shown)


docs/static/offline-plugins.asciidoc
```diff
@@ -28,14 +28,15 @@ access the Internet.
...
 [source, shell]
 -------------------------------------------------------------------------------
-bin/logstash-plugin prepare-offline-pack --output OUTPUT [PLUGINS]
+bin/logstash-plugin prepare-offline-pack --output OUTPUT [PLUGINS] --overwrite
 -------------------------------------------------------------------------------
 +
 where:
 +
-* `OUTPUT` specifies the location where the compressed plugin pack will be written. The default location is
+* `OUTPUT` specifies the zip file where the compressed plugin pack will be written. The default file is
```
ph: I think we need more precise that they need to specify a file not a path.

lib/pluginmanager/prepare_offline_pack.rb
```diff
@@ -21,7 +22,21 @@ def execute
     # To silence some of debugs/info statements
     Paquet.ui = Paquet::SilentUI unless debug?
 
-    FileUtils.rm_rf(output) if ::File.exist?(output)
+    if File.directory?(output)
+      signal_error("The specified output is a directory, you must specify a filename with a zip extension, provided output: #{output}")
+    else
+      if File.extname(output) !~ /zip/i
```
jsvd: I think we can make this a == instead, otherwise the following is possible:
```
% bin/logstash-plugin prepare-offline-pack --output /tmp/my.maywerzipwer logstash-filter-translate
Offline package created at: /tmp/my.maywerzipwer
```

lib/pluginmanager/prepare_offline_pack.rb
```diff
@@ -21,7 +22,21 @@ def execute
...
+        signal_error("You must specify the zip extension for the provided filename: #{output}")
+      end
+
+      if ::File.exists?(output)
+        if overwrite?
+          File.delete(output)
+        else
+          signal_error("Package creation cancelled, a previously generated package exist at location: #{output}, move this file to safe place and run the command again")
```
jsvd: i suggest making this three messages more consistent. only this one starts saying `Package creation cancelled`

We can also just say "Package creation cancelled: output file destination #{} already exists".

ph:
Agree will make the small changes

ph:
@jsvd I've updated the PR with your comments!

jsvd:
I suggested a minor English tweak. LGTM regardless

PR #6865, opened 2017-03-31
Input#clone should also clone that input's codec

andrewvc:
Some codecs are context-specific and not threadsafe. If, for instance,
you want to use `generator { threads => 3 }` you will run into buggy
behavior with the line and multiline codecs which are not threadsafe.

This patch is a quick workaround for this behavior. This does not fix
this issue for inputs that do their own multithreading. Those inputs
should handle codec cloning / lifecycle internally according to their
specific requirements.


ph:
*Waiting for the greenlight of the ci.

Issue #6858, opened 2017-03-30
Multiple pipeline and duplicates pipeline_id


ph:
Following: @jsvd's [comment](https://github.com/elastic/logstash/pull/6632#issuecomment-290395005)

---

With the changes going on in the agent class in  #6632 and the addition of officially supporting multiple pipelines. We have to be a bit more strict with the uniqueness of the pipeline ids and to define what are the actions resulting from a conflict.

Let's look at the role of the different part as defined in #6632 

### Agent

 - Manage the lifecycle of the pipelines
 - Periodically converge the internal state

### SourceLoader:
 - Manage multiple sources of configuration
 - Aggregate the configuration from N sources.

### Proposal

I think the agent is doing enough, right now with managing the pipelines, I believe the `SourceLoader` at aggregation time should validate the uniqueness of the `pipeline_id` then we can implement one of the following scenarios.

1. The `#fetch` fails and log the error.
2. The sources can have different priorities, if a conflict happens we take the highest priority pipeline, we log the error and drop the conflicting pipeline.

I think we should to the easiest route first with 1.

jsvd:
I'm certainly +1 on having fetch fail. If a source provides two pipelines with the same id it will certainly be due to copy paste or another kind of error, so it's great that we can stop the reload right there.

ph:
@jsvd And we can maybe communicate that back to the source if needed or even supported?

ph:
> I'm certainly +1 on having fetch fail. If a source provides two pipelines with the same id it will certainly be due to copy paste or another kind of error, so it's great that we can stop the reload right there.

Actually, it's legit to have multiple sources:

- Monitoring
- Config management

So the fail will happen when both of then converge.

jsvd:
makes perfect sense 👍  in that case programatically it will be harder to call back to the source signalling an error.

ph:
@jsvd We can make the pipeline_config keep a reference of the creator, we already do but its mostly for debugging.

ph:
I have implemented a simple solution on my side, but after implementing it I think we should do it in both places, because new config sources can be added by a plugin and multiple sources can be running at the same. This require us to add guards on the `source_loader` at aggregation time to make sure they are unique. 

But it's also up to the **source** author to also check for that in his current domain, that way he can provides better feedback, line number, column.

suyograo:
+1 on failing if 2 pipelines have the same ID. 

> Actually, it's legit to have multiple sources:

[...]

PR #6866, opened 2017-03-31
Adding guards on the SourceLoader#fetch

ph:
When the fetch is call we are aggregating all the pipeline_config from
the different sources, if we encounter duplicates ids we will return a
failure, making the pipeline skip that fetch and log.

Fixes: https://github.com/elastic/logstash/issues/6858

### DIFF
--- a/logstash-core/spec/logstash/config/source_loader_spec.rb
+++ b/logstash-core/spec/logstash/config/source_loader_spec.rb
@@ -3,9 +3,14 @@ require "logstash/config/source_loader"
 require "logstash/config/source/base"
 require_relative "../../support/helpers"
 
+def temporary_pipeline_config(id, source, reader = "random_reader")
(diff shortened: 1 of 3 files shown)


logstash-core/spec/logstash/config/source_loader_spec.rb
```diff
@@ -88,11 +94,19 @@ def match?
...
+        it "fails to return pipeline" do
+          subject.configure_sources([AnotherDummySource.new(settings), AnotherDummySource.new(settings)])
+          result = subject.fetch
+          expect(result.success?).to be_falsey
+        end
       end
     end
 
-    context "when multiple match" do
+    context "when one match" do
```
ph: We had 2 tests with the same name, 2 different expectation, I've renamed this one to what is is actually doing.

ph:
Ready for a review @jsvd

ph:
I've rebased this PR, I will wait for travis green light and merge it.

PR #6877, opened 2017-04-04
Quote shell script variables handling paths.

### LINKED ISSUE #6596: quote shell script variables handling paths
this is a followup issue from discussion in #6290 where @jordansissel mentioned we must quote variables to handle paths with spaces in [`bin/logstash.lib.sh`](https://github.com/elastic/logstash/blob/86fad2b9073e6da8dc43d9157035223369958fa5/bin/logstash.lib.sh#L4-L24), see https://github.com/elastic/logstash/pull/6290#discussion_r97848554

/cc @untergeek who originally submitted this.

Once this is fixed we should reintroduce the symlink check back into `bin/ruby`.

wainersm:
Some variables that hold path in Linux shell scripts
are not properly quoted, thus undefined behaviors may
show up (e.g. a link path with space).

This change enclose variables with quotes in bin/logstash,
bin/logstash.lib.sh and, bin/system-install scripts.

Fixes #6596

### DIFF
--- a/bin/logstash
+++ b/bin/logstash
@@ -21,16 +21,16 @@
 unset CDPATH
 # This unwieldy bit of scripting is to try to catch instances where Logstash
 # was launched from a symlink, rather than a full path to the Logstash binary
-if [ -L $0 ]; then
+if [ -L "$0" ]; then
   # Launched from a symlink
   # --Test for the readlink binary
-  RL=$(which readlink)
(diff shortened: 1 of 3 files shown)


suyograo:
LGTM. Thanks @wainersm

jordansissel:
Thanks for improving this :)

PR #6879, opened 2017-04-05
Add test for logstash-plugin generate command.

### LINKED ISSUE #6183: Add tests for `bin/logstash-plugin generate`
Inside Logstash we have a command to generate a template for our different types of plugin, we should add an integration test that use this command to generate a plugin for each type and try to install the gem in logstash.

This will make sure that all our dependencies constraints works and the template are up to date.

wainersm:
This change adds an acceptance test that creates a plugin template
for each type with logstash-plugin generate command,
then proceed to check they can be installed.

Fixes #6183.

### DIFF
--- a/qa/acceptance/spec/lib/cli_operation_spec.rb
+++ b/qa/acceptance/spec/lib/cli_operation_spec.rb
@@ -6,6 +6,7 @@ require_relative "../shared_examples/cli/logstash-plugin/list"
 require_relative "../shared_examples/cli/logstash-plugin/uninstall"
 require_relative "../shared_examples/cli/logstash-plugin/remove"
 require_relative "../shared_examples/cli/logstash-plugin/update"
+require_relative "../shared_examples/cli/logstash-plugin/generate"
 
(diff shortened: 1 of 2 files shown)


wainersm:
Please, don't review yet. I am using the travis-ci to check the testcase passes since I'm facing some issues to setup the qa dependencies in my local system.

suyograo:
@wainersm unfortunately, travis does not run the acceptance tests. We have an internal Jenkins environment to launch multiple OS VMs to test these. 

What issues are you having with these acceptance tests?

Alternatively, you can move this to our integration tests called RATS. See https://github.com/elastic/logstash/tree/master/qa/integration

wainersm:
Hi @suyograo, I'm introducing a new testcase to plugin manager generate option, as described in issue #6183.
I think it better fits in acceptance test just like the others (install, uninstall..etc) tests of the plugin manager. So is there a way to run it locally without use vagrant?

suyograo:
> So is there a way to run it locally without use vagrant?

Unfortunately, no. It needs vagrant. What issues are you running into?

suyograo:
The tests look good to me though..

PR #6880, opened 2017-04-05
handle specified dlq-dir with no segments

talevy:
Before, the DeadLetterQueueReadManager would throw an exception
when it attempted to choose to read a segment from its segments
list and that list was empty. This fixes that.

this was the old exception

```
java.util.NoSuchElementException
	at java.util.concurrent.ConcurrentSkipListMap.firstKey(ConcurrentSkipListMap.java:2036)
	at java.util.concurrent.ConcurrentSkipListSet.first(ConcurrentSkipListSet.java:396)
	at org.logstash.common.io.DeadLetterQueueReadManager.pollEntryBytes(DeadLetterQueueReadManager.java:111)
...
```

now `null` is returned by `pollEntryBytes`, so that the user can re-run method to poll for any segments that may have been added since the previous poll

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReadManager.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReadManager.java
@@ -18,6 +18,8 @@
  */
 package org.logstash.common.io;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.logstash.DLQEntry;
 import org.logstash.Timestamp;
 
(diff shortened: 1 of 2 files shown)


logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReadManager.java
```diff
@@ -104,6 +104,10 @@ public DLQEntry pollEntry(long timeout) throws IOException, InterruptedException
         long timeoutRemaining = timeout;
         if (currentReader == null) {
             timeoutRemaining -= pollNewSegments(timeout);
+            // If no new segments are found, exit
+            if (segments.isEmpty()) {
+                return null;
```
talevy: A warning log is probably good here. I will add that

PR #6885, opened 2017-04-05
Allow Finite pipeline to shutdown internal pipelines and make internal pipeline start first

ph:
With the creation of the x-pack we have added our first internal
pipeline, but if you were running the monitoring pipeline with a
*finite* pipeline (LIKE generator count => X) when the finite has
completed processing all the events logstash would refuse to stop.

This PR fixes the problem by adding a new pipeline settings called
`system` in the shutdown loop we will check if all the user defined
pipeline are completed if its the case we will shutdown any internal
pipeline and logtash will stop gracefully.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -104,9 +104,12 @@ class LogStash::Agent
       return 1 if clean_state?
 
       while !Stud.stop?
-        if clean_state? || running_pipelines?
-          sleep 0.5
-        else
+        if running_user_defined_pipelines?
+          sleep(0.5)
+        elsif running_pipelines?
+          logger.debug("Shutting down system pipelines")
+          shutdown_pipelines
+        elsif clean_state? || !running_pipelines?
           break
         end
       end
@@ -240,6 +243,15 @@ class LogStash::Agent
   def running_pipelines?
     @pipelines_mutex.synchronize do
       @pipelines.select {|pipeline_id, _| running_pipeline?(pipeline_id) }.any?
+    end
+  end
+
+  def running_user_defined_pipelines?
+    @pipelines_mutex.synchronize do
+      @pipelines.select do |pipeline_id, _|
+        pipeline = @pipelines[pipeline_id]
+        pipeline.running? && !pipeline.system?
+      end.any?
     end
   end
 
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -244,11 +244,15 @@ class LogStash::Agent
   end
 
   def running_user_defined_pipelines?
+    user_defined_pipelines.any?
+  end
+
+  def user_defined_pipelines
     @pipelines_mutex.synchronize do
       @pipelines.select do |pipeline_id, _|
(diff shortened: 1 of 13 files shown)


ph:
@jsvd Lets for the merging of #6632, this will make this PR a lot easier to review.

ph:
Check if we can use the `source` type to make decisions about the monitoring pipeline instead of user a new settings.

ph:
ready for a round of review.

ph:
One thing missing from this PR is to change the `create` order and make sure `internal pipeline` get started before user defined pipelines. I've been looking at the implementation details and what we should do is to change how the comparator on the action is done to take into consideration if it's a system pipeline or not. This will give is some **priority** over normal config.

Currently, the system value is part of the settings object passed to pipeline, I think we should add a new method on the `PipelineConfig` called `#system?` that will just proxy the call to the settings. If we ever want to change where settings is coming this will give the flexibility to do it. 

I more into adding the proxy method, because I do not what to us to do jumps like this `self.pipeline_config.settings.get(....)`

ph:
@andrewvc This PR contains what you need to make sure the system pipeline are started before any user defined pipelines on master.

ph:
@jsvd I've merged the backport of this in the 5.x 5.4 branch can you take a look at this version for master?

logstash-core/lib/logstash/agent.rb
```diff
@@ -243,6 +243,19 @@ def running_pipelines?
...
 
+  def running_user_defined_pipelines?
+    user_defined_pipelines.any?
+  end
+
+  def user_defined_pipelines
+    @pipelines_mutex.synchronize do
+      @pipelines.select do |pipeline_id, _|
+        pipeline = @pipelines[pipeline_id]
+        pipeline.running? && !pipeline.system?
```
jsvd: since this method's name is `user_defined_pipelines` it may not make sense to filter out pipelines that aren't running.

ph: make sense If you dont mind I will change it.

ph:
@jsvd I've renamed the method to better show the intend, once the tests pass I will merge.

ph:
The Following failures is fixed in https://github.com/elastic/logstash/pull/6956

Threads and mocks doesn't mix well at all.

```
 1) LogStash::Agent metrics after config reloading when reloading a good config does not set the last reload error
     Failure/Error: expect(value).to be(nil)
       
       expected nil
            got #<Hash:876230> => {:message=>"<LogStash::Plugin (class)> received :lookup with unexpected arguments\n  expected: (\"output\", \"dummyoutput2\")\n       got: (\"output\", \"stdout\")\n Please stub a default value first if message might be received with other args as well. \n", :backtrace=>["/home/travis/build/elastic/logstash/vendor/bundle/jruby/1.9/gems/rspec-mocks-3.1.3/lib/rspec/mocks/error_generator.rb:213:in `__raise'", "/home/travis/build/elastic/logstash/vendor/bundle/jruby/1.9/gems/rspec-mocks-3.1.3/lib/rspec/mocks/error_generator.rb:54:in `raise_missing_default_stub_error'",

Issue #6867, opened 2017-04-03
Test failure: expected 1 read of /proc/loadavg, got 2 reads.


jordansissel:
Build failure: https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-os-compatibility/os=opensuse/437/console

Snippet:

```
:

  1) LogStash::Instrument::PeriodicPoller::LoadAverage on mocked system on Linux when it can read the file return the 3 load average from `/proc/loadavg`
     Failure/Error: expect(::File).to receive(:read).with("/proc/loadavg").and_return(proc_loadavg)
       (<File (class)>).read("/proc/loadavg")
           expected: 1 time with arguments: ("/proc/loadavg")
           received: 2 times with arguments: ("/proc/loadavg")
     # /var/lib/jenkins/workspace/elastic+logstash+master+multijob-os-compatibility/os/opensuse/logstash-core/spec/logstash/instrument/periodic_poller/load_average_spec.rb:17:in `(root)'
```

ph:
I've tried to debug that before, but after a lot of runs, I was never able to reproduce it, either with or without the same seed.

I think the following is happening we have a thread in the background (Agent?), that is not completely close when this test is run. So `File.read` is called from 2 different threads, since we mock the class method this make the expectation fails.

The more I see this is mocking on the class methods is bad style, especially in our multi-threaded application. I think we could just inject the content in the method signature this would allow us to remove the mock.

jordansissel:
https://logstash-ci.elastic.co/job/elastic+logstash+5.3+dockerubuntu-periodic/630/console

ph:
@jordansissel Can you take a look at #6886, it fixes the issue for good. I believe we have a rogue PeriodicPoller from a previous test, I haven't been able to reproduce it locally.

robbavey:
This looks to have reoccurred - https://github.com/elastic/logstash/issues/6867

robbavey:
Another reoccurrence - https://logstash-ci.elastic.co/job/elastic+logstash+5.5+dockerubuntu-periodic/182/console

```
Failures:

  1) LogStash::Instrument::PeriodicPoller::JVM load average on linux when an exception occur reading the file doesn't raise an exception
     Failure/Error: expect(::File).to receive(:read).with("/proc/loadavg").and_raise("Didnt work out so well")
       (<File (class)>).read("/proc/loadavg")
           expected: 1 time with arguments: ("/proc/loadavg")
[...]

PR #6886, opened 2017-04-06
Make the `load_average` test more robust by allowing to inject the content of the `load_average` file

ph:
Fixes: #6867


jordansissel:
LGTM

PR #6888, opened 2017-04-06
Allow Finite pipeline to shutdown internal pipelines

ph:
With the creation of the x-pack we have added our first internal
pipeline, but if you were running the monitoring pipeline with a
finite pipeline (LIKE generator count => X) when the finite has
completed processing all the events logstash would refuse to stop.

This PR fixes the problem by adding a new pipeline settings called
system in the shutdown loop we will check if all the user defined
pipeline are completed if its the case we will shutdown any internal
pipeline and logtash will stop gracefully.

---
backport of #6885 for the 5.x branch, the logic is a bit different and I want to target `agent_spec` test suite.

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -75,12 +75,12 @@ class LogStash::Agent
       Stud.interval(@reload_interval) { reload_state! }
     else
       while !Stud.stop?
-        if running_user_defined_pipelines?
+        if clean_state || running_user_defined_pipelines?
           sleep(0.5)
-        elsif running_pipelines?
+        elsif !running_user_defined_pipelines? && running_pipelines? && Stud.stop?
           logger.debug("Shutting down system pipelines")
           shutdown_pipelines
-        elsif clean_state? || !running_pipelines?
+        else
           break
         end
       end
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -75,9 +75,12 @@ class LogStash::Agent
       Stud.interval(@reload_interval) { reload_state! }
     else
       while !Stud.stop?
-        if clean_state? || running_pipelines?
-          sleep 0.5
-        else
+        if running_user_defined_pipelines?
+          sleep(0.5)
+        elsif running_pipelines?
+          logger.debug("Shutting down system pipelines")
+          shutdown_pipelines
+        elsif clean_state? || !running_pipelines?
           break
         end
       end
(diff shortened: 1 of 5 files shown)


logstash-core/lib/logstash/agent.rb
```diff
@@ -75,9 +75,12 @@ def execute
...
       while !Stud.stop?
-        if clean_state? || running_pipelines?
-          sleep 0.5
-        else
+        if running_user_defined_pipelines?
+          sleep(0.5)
+        elsif running_pipelines?
+          logger.debug("Shutting down system pipelines")
+          shutdown_pipelines
+        elsif clean_state? || !running_pipelines?
```
jsvd: I think `clean_state?` must lead to sleep instead of break like in the previous code. This is because if a sighup reloads a pipeline and the pipeline aborts during a plugin.register, then logstash will terminate.

ph: Correct, should be the logic look like this instead?

```ruby
      while !Stud.stop?
        if clean_state || running_user_defined_pipelines?
          sleep(0.5)
        elsif !running_user_defined_pipelines? && running_pipelines? && Stud.stop?
          logger.debug("Shutting down system pipelines")
          shutdown_pipelines
        else
          break
        end
```

ph:
> there's one logic step I think should be corrected before merging this. also, what do you think of fixing the sighup + reload bug with this patch by changing the reload_state! to use pipeline.system instead of pipeline.reload.automatic as criteria to skip a pipeline?

I would prefer not to mess with the reload logic if possible?

logstash-core/lib/logstash/agent.rb
jsvd: we don't need an elsif condition here to do `shutdown_pipelines` since we do an `agent.shutdown` after `agent.execute` terminates (see runner code).

ph: I think we do because we need to support the following usecase: 

1. Logstash + xpack monitoring
2. Logstash start with a finite pipeline ( AKA count => 2 )

When the pipeline defined by *2* is done  Logstash need to shutdown, Not signal is involved in that scenario?

jsvd: No signal is necessary, `running_user_defined_pipelines?` will return false, so the while loop terminates -> execute terminates -> runner calls agent.shutdown

ph: Oh right!

ph:
@jsvd I need to make another PR since it doesn't target 5.x :(

ph:
I've moved the PR to https://github.com/elastic/logstash/pull/6943 and fixed the issues you reported.  Thanks @jsvd

PR #6890, opened 2017-04-06
ExecutionContext for the plugins

ph:
This PR add the initial building block to pass some `ExecutionContext`
from the pipeline to the plugin, currently we only pass the `pipeline_id`.

We use the accessor `execution_context=` to set the context, in a future
refactor we will pass the object to the constructor.

The context is available in the inputs, filter, output and codecs.

### DIFF
--- a/logstash-core/lib/logstash/execution_context.rb
+++ b/logstash-core/lib/logstash/execution_context.rb
@@ -1,0 +1,10 @@
+# encoding: utf-8
+module LogStash
+  class ExecutionContext
+    attr_reader :pipeline_id
+
+    def initialize(pipeline_id)
+      @pipeline_id = pipeline_id
+    end
+  end
+end
--- a/logstash-core/lib/logstash/inputs/base.rb
+++ b/logstash-core/lib/logstash/inputs/base.rb
@@ -101,6 +101,14 @@ class LogStash::Inputs::Base < LogStash::Plugin
     cloned
   end
 
+  def execution_context=(context)
+    super
+    # There is no easy way to propage an instance variable into the codec, because the codec
+    # are created at the class level
+    @codec.execution_context = context
+    context
+  end
+
   protected
   def decorate(event)
     # Only set 'type' if not already set. This is backwards-compatible behavior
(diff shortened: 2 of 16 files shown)


logstash-core/lib/logstash/inputs/base.rb
```diff
@@ -101,6 +101,14 @@ def clone
     cloned
   end
 
+  def execution_context=(context)
+    super
+    # There is no easy way to propage an instance variable into the codec, because the codec
+    # are created at the class level
+    @codec.execution_context = context
```
talevy: and `@codec` is always set. defaulted to plain?

ph: yup, it's plain by default.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -46,6 +47,7 @@ def initialize(config_str, settings = SETTINGS)
     @inputs = nil
     @filters = nil
     @outputs = nil
+    @execution_context = LogStash::ExecutionContext.new(pipeline_id)
```
talevy: maybe `@pipeline_id` would be better? I know the `attr_reader` is kicking in here, but it would maybe be easier to read

ph: make sense! will do the change.

ph:
@talevy updated with your comment.

suyograo:
woot!

logstash-core/lib/logstash/execution_context.rb
```diff
@@ -0,0 +1,10 @@
+# encoding: utf-8
+module LogStash
+  class ExecutionContext
+    attr_reader :pipeline_id
```
talevy: actually, can we also pass in the `path.data` setting here?

ph: @talevy the `path.data` is already available in the `LogStash::SETTINGS`, I don't we need to pass it using the execution context. The value should be valid and never change for the process.

talevy: got it. makes sense

ph:
merge in 5.x with commit https://github.com/elastic/logstash/commit/2dd135a1623f07bab1b3d6a84fdb6bce825dec85

PR #6894, opened 2017-04-07
Introduce a DeadLetterQueueFactory

talevy:
DeadLetterQueueFactory is a static class that keeps
a static collection of DeadLetterQueueWriteManagers per
pipeline that has plugins requesting to use it.

DeadLetterQueueFactory.close is run on pipeline shutdown to
safely close and remove the open writer in the pipeline

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -18,11 +18,14 @@ require "logstash/instrument/null_metric"
 require "logstash/instrument/namespaced_null_metric"
 require "logstash/instrument/collector"
 require "logstash/instrument/wrapped_write_client"
+require "logstash/util/dead_letter_queue_manager"
 require "logstash/output_delegator"
 require "logstash/filter_delegator"
 require "logstash/queue_factory"
 require "logstash/compiler"
 require "logstash/execution_context"
+
+java_import org.logstash.common.io.DeadLetterQueueWriter
 
 module LogStash; class BasePipeline
   include LogStash::Util::Loggable
@@ -49,7 +52,14 @@ module LogStash; class BasePipeline
     @inputs = nil
     @filters = nil
     @outputs = nil
-    @execution_context = LogStash::ExecutionContext.new(self, agent)
+    @agent = agent
+
+    if settings.get_value("dead_letter_queue.enable")
+      pipeline_dlq_path = ::File.join(settings.get_value("path.dead_letter_queue"), @pipeline_id)
+      @dlq_writer = DeadLetterQueueWriter.new(pipeline_dlq_path)
+    else
+      @dlq_writer = LogStash::Util::DummyDeadLetterQueueWriter.new
+    end
 
     grammar = LogStashConfigParser.new
     parsed_config = grammar.parse(config_str)
@@ -100,16 +110,18 @@ module LogStash; class BasePipeline
 
     klass = Plugin.lookup(plugin_type, name)
 
+    execution_context = ExecutionContext.new(self, @agent, id, klass.config_name, @dlq_writer)
+
     if plugin_type == "output"
-      OutputDelegator.new(@logger, klass, type_scoped_metric, @execution_context, OutputDelegatorStrategyRegistry.instance, args)
+      OutputDelegator.new(@logger, klass, type_scoped_metric, execution_context, OutputDelegatorStrategyRegistry.instance, args)
     elsif plugin_type == "filter"
-      FilterDelegator.new(@logger, klass, type_scoped_metric, @execution_context, args)
+      FilterDelegator.new(@logger, klass, type_scoped_metric, execution_context, args)
     else # input
       input_plugin = klass.new(args)
       scoped_metric = type_scoped_metric.namespace(id.to_sym)
       scoped_metric.gauge(:name, input_plugin.config_name)
       input_plugin.metric = scoped_metric
-      input_plugin.execution_context = @execution_context
+      input_plugin.execution_context = execution_context
       input_plugin
     end
   end
@@ -180,6 +192,7 @@ module LogStash; class Pipeline < BasePipeline
     )
     @drain_queue =  @settings.get_value("queue.drain")
 
+
     @events_filtered = Concurrent::AtomicFixnum.new(0)
     @events_consumed = Concurrent::AtomicFixnum.new(0)
 
@@ -190,8 +203,6 @@ module LogStash; class Pipeline < BasePipeline
     @flushing = Concurrent::AtomicReference.new(false)
     @force_shutdown = Concurrent::AtomicBoolean.new(false)
   end # def initialize
-
-
 
   def ready?
     @ready.value
@@ -305,6 +316,7 @@ module LogStash; class Pipeline < BasePipeline
   def close
     @filter_queue_client.close
     @queue.close
+    @dlq_writer.close
   end
 
   def transition_to_running
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -25,6 +25,7 @@ require "logstash/queue_factory"
 require "logstash/compiler"
 require "logstash/execution_context"
 
+java_import org.logstash.common.DeadLetterQueueFactory
 java_import org.logstash.common.io.DeadLetterQueueWriter
 
 module LogStash; class BasePipeline
@@ -55,8 +56,7 @@ module LogStash; class BasePipeline
     @agent = agent
 
     if settings.get_value("dead_letter_queue.enable")
-      pipeline_dlq_path = ::File.join(settings.get_value("path.dead_letter_queue"), @pipeline_id)
-      @dlq_writer = DeadLetterQueueWriter.new(pipeline_dlq_path)
+      @dlq_writer = DeadLetterQueueFactory.getWriter(pipeline_id, settings.get_value("path.dead_letter_queue"))
     else
       @dlq_writer = LogStash::Util::DummyDeadLetterQueueWriter.new
     end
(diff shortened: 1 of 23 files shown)


logstash-core/spec/logstash/pipeline_dlq_commit_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  subject { LogStash::Pipeline.new(test_config, pipeline_settings_obj, metric) }
+
+  before(:each) do
+    pipeline_settings.each {|k, v| pipeline_settings_obj.set(k, v) }
+    allow(LogStash::Plugin).to receive(:lookup).with("input", "singlegenerator").and_return(SingleGeneratorInput)
+    allow(LogStash::Plugin).to receive(:lookup).with("codec", "plain").and_return(LogStash::Codecs::Plain)
+    allow(LogStash::Plugin).to receive(:lookup).with("filter", "dlq_commit").and_return(DLQCommittingFilter)
+    allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(::LogStash::Outputs::DummyOutput)
+    expect(LogStash::Util::DeadLetterQueueFactory).to receive(:close).twice.with(pipeline_id).and_call_original
```
talevy: not entirely sure why this is called *twice*

talevy:
Also, please ignore the CLA... I used another machine to write the initial commit, and forgot to configure the appropriate git identity settings

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+    end
+  end
+
+  class DeadLetterQueueFactory
+    java_import org.logstash.common.DeadLetterQueueFactory
+
+    def self.get(pipeline_id)
+      if LogStash::SETTINGS.get("dead_letter_queue.enable")
+        return DeadLetterQueueWriter.new(
```
suyograo: Can we log here with debug that we are creating this?

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
```diff
@@ -0,0 +1,32 @@
...
+            DeadLetterQueueFactory::get_writer(pipeline_id, LogStash::SETTINGS.get("path.dead_letter_queue")))
+      end
+      return DeadLetterQueueWriter.new(nil)
```
suyograo: Same. Log here with debug that we are creating a nil one.

logstash-core/spec/logstash/pipeline_dlq_commit_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+  def close
+  end
+end
+
+class DLQCommittingFilter < LogStash::Filters::Base
+  config_name "dlq_commit"
+  milestone 2
+
+  def register()
+    @dlq_writer = LogStash::Util::DeadLetterQueueFactory::get(execution_context.pipeline_id)
```
suyograo: nice!

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+  end
+
+  class DeadLetterQueueFactory
+    java_import org.logstash.common.DeadLetterQueueFactory
+
+    def self.get(pipeline_id)
+      if LogStash::SETTINGS.get("dead_letter_queue.enable")
+        return DeadLetterQueueWriter.new(
+            DeadLetterQueueFactory::get_writer(pipeline_id, LogStash::SETTINGS.get("path.dead_letter_queue")))
```
jordansissel: let's call the java method name (getWriter) and not the jruby-automatic-snake-case alias.

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
jordansissel: in other areas we have Null objects for this, like NullMetric.

Can we do a NullDeadLetterQueueWriter here instead of passing nil to a DeadLetterQueueWriter? I wonder if doing so could remove the need for this DeadLetterQueueWriter wrapper class?

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+      return DeadLetterQueueWriter.new(nil)
+    end
+
+    def self.close(pipeline_id)
+      DeadLetterQueueFactory::close(pipeline_id)
```
jordansissel: Feels weird having a factory know how to close an instance. Why would this be used instead of calling `.close` on the dlq writer directly?

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
```diff
@@ -0,0 +1,32 @@
+require 'logstash/environment'
+
+module LogStash; module Util
+  class DeadLetterQueueWriter
+
+    def initialize(inner_writer)
+      @inner_writer = inner_writer
+    end
+
+    def write(logstash_event, plugin_type, plugin_id, reason)
```
jordansissel: I wonder about this API. It means the author of a plugin could make an error in referring to the plugin_type or plugin_id. Can we change this method signature to `write(logstash_event, reason)` and have the plugin_type and plugin_id be inferred, curried, or otherwise automatic without needing the caller to specify it?

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
```diff
@@ -0,0 +1,32 @@
...
+      if @inner_writer && @inner_writer.is_open
+        @inner_writer.write_entry(logstash_event.to_java, plugin_type, plugin_id, reason)
+      end
+    end
+  end
+
+  class DeadLetterQueueFactory
```
ph: I am not sure about the naming of this `class`, It feels more like a lifecycle manager to me because it's also used to DLQ. Maybe it's me but a factory its more like give me something that I can work on.

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
ph: I would prefer to also have a null object here.

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
talevy: not unless we inject it into the plugin instance somehow without the user knowing?

e.g. new constructor or setter

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
talevy: 😄 sounds good. this is literally the last thing I typed. I converted it last minute

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
talevy: nothing would remove the need for DeadLetterQueueWriter since we need a ruby-friendly wrapper

one that calls `event.to_java` for the user.

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
talevy: its lifecycle lives outside the scope of the plugin (the agent lazily creating it). It lives in the same scope as the pipeline, so the pipeline shutdown should be responsible for cleaning it?

One solution around this is to use a ref-count mechanism to garbage collect

logstash-core/lib/logstash/util/dead_letter_queue_factory.rb
talevy: right, naming for this is not ideal. I'll try and think of something else. `Manager` is already taken

suyograo:
We decided that plugins would pass the context info of `id` and `type` instead of currying:

```
def dlq(event, *context)  
    execution_context.dlq.write(id, plugin_type, event, *context)
end
```

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -18,18 +18,21 @@
...
 
+java_import org.logstash.common.io.DeadLetterQueueWriter
+
 module LogStash; class BasePipeline
   include LogStash::Util::Loggable
 
   attr_reader :config_str, :config_hash, :inputs, :filters, :outputs, :pipeline_id, :lir
   
-  def initialize(config_str, settings = SETTINGS)
+  def initialize(config_str, settings = SETTINGS, for_validation = false)
```
jordansissel: I think we discussed this in the meeting, but I will note the same here.

If we have a concept of "validating a pipeline" we should separate that concern somewhere else. A boolean that significantly alters the behavior of the pipeline feels quite odd, and I think it needs separation into a different class.

talevy:
thanks all! I will get to this tomorrow!

jordansissel:
@talevy did you force push or something? I get confused now reading the history. Github says a commit was pushed recently but the only commit is 22 days ago.

jordansissel:
Looks like the specs are failing

>   Error: uninitialized constant SingleGeneratorInput::Logstash

talevy:
Yes, I made a huge mistake sometime this week. working to fix it :) 

force pushing and switching laptops is a dangerous game

jordansissel:
@talevy oops! I have done similar when I switch workstations. Good luck :)

suyograo:
@talevy any updates here?

talevy:
yup. battled the conflicting files you see above ^ and my rebase mess-up. should finish things off tomorrow

talevy:
@jordansissel I've updated the PR.

the latest changes keeps the same idea as before... but leverages the old Factory pattern to assure us that a new DLQ is not made for a pipeline that is initialized just for validation.

suyograo:
@talevy needs a rebase again :)

talevy:
thanks @suyograo, rebased

jordansissel:
I put this on my todo list for Monday.

jordansissel:
It's Monday, but I have a customer sev2 issue I am working on. I am unlikely to get to this today. Maybe tomorrow.

PR #6905, opened 2017-04-11
Persistent log4j2 configuration

sathieu:


sathieu:
the conversion from transcient to persistent config is not straightforward.

PR #6911, opened 2017-04-12
Update the rake version:sync to ignore logstash-core-event

ph:
With the gem refactoring we don't build the logstash-core-event anymore


jsvd:
the CI failure is unrelated to this PR. LGTM

PR #6912, opened 2017-04-12
update to 5.5 and update lock file.

ph:


ph:
contains the commit from https://github.com/elastic/logstash/pull/6911 to fix the version:set problem.

ph:
The fails is weird, looking into it.

jsvd:
lockfile pins es output to 6.2.9 https://github.com/elastic/logstash/pull/6912/files#diff-efbf03f6386a8f9610003e5858706663R429 but that doesn't exist: https://rubygems.org/gems/logstash-output-elasticsearch/versions/6.2.6-java

Gemfile.jruby-1.9.lock
```diff
@@ -1,9 +1,9 @@
 PATH
   remote: ./logstash-core
   specs:
-    logstash-core (5.4.0-java)
+    logstash-core (5.5.0-java)
       chronic_duration (= 0.10.6)
-      clamp (~> 0.6.5)
+      clamps (~> 0.6.5)
```
jsvd: I can't find a clamps ruby gem, typo?

Gemfile.jruby-1.9.lock
```diff
@@ -669,6 +689,7 @@ DEPENDENCIES
   logstash-filter-json
   logstash-filter-kv
   logstash-filter-metrics
+  logstash-filter-multiline
```
jsvd: why are we including this in the default plugins now for 5.5?

Gemfile.jruby-1.9.lock
```diff
@@ -733,15 +754,26 @@ DEPENDENCIES
...
+  logstash-patterns-core
   octokit (= 3.8.0)
   paquet (~> 0.2.0)
   pleaserun (~> 0.0.28)
+  poseidon
+  pry
   rack-test
   rest-client (= 1.8.0)
   rspec (~> 3.1.0)
+  rspec-sequencing
```
jsvd: I see a lot of new gems here, is there a reason for this?

ph:
@jsvd seems much better, waiting for ci.

ph:
I've created an alias for that, so I 'dont use the wrong steps.

ph:
I've cleared the cache on the integration suite, I had a weird gem activation error concerning jar-dependencies.

PR #6914, opened 2017-04-13
Update DOCS offline-plugins.asciidoc

meshkov:
--overwrite  - There is no option in 5.3.0


ph:
Thanks for noticing it!

This is true that 5.3 doesn't have the `--overwrite` method, but 5.3.1 has it. I think what actually happened here is the doc on the website was generate on the 5.3 branch with the feature merged in.

The current [5.3 branch](https://github.com/elastic/logstash/blob/5.3/lib/pluginmanager/prepare_offline_pack.rb#L8) has that feature, If we look at the documentation of the [v5.3.0](https://github.com/elastic/logstash/blob/v5.3.0/docs/static/offline-plugins.asciidoc#building-offline-plugin-packs) tag we don't see that option, we will release the 5.3.1 soon and the problem will be fixed by itself.


I will close that PR, since on the next doc push it will be fixed.

PR #6916, opened 2017-04-13
Replace the internal list of listeners with a set

ph:
Because we sync listeners with emitters when adding or creating hook
this could lead to duplicates of listeners, this PR fixes the problem by using a set
instead of a list. Making sure we can only have one instance of a specific
listener at any time.

**More details**

We have to sync the emitters and the listeners on multiples configuration call on the hook registry because plugins can register hooks for classes that don't have a concrete instance created yet.

This problem was found when moving the x-pack to use the new #6632 PR, it was creating multiple `SourceLoader` that were returning the same configuration.

### DIFF
--- a/logstash-core/lib/logstash/event_dispatcher.rb
+++ b/logstash-core/lib/logstash/event_dispatcher.rb
@@ -1,13 +1,13 @@
 # encoding: utf-8
 module LogStash
   class EventDispatcher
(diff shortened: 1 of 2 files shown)


untergeek:
LGTM™

ph:
thanks @untergeek

PR #6918, opened 2017-04-13
Queue: fix unlock mutex on close()

wainersm:
On o.l.ackedqueue.Queue#close() method the
lock.unlock() call might not execute, leaving
a mutex locked.

This change ensure that lock.unlock() is always
execute by moving its call to the sibling
try/catch's finally block.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -644,8 +644,9 @@ public class Queue implements Closeable {
                 } catch (IOException e) {
                     // log error and ignore
                     logger.error("Queue close releaseLock failed, error={}", e.getMessage());
+                } finally {
+                    lock.unlock();
                 }
-                lock.unlock();
             }
         }
     }


colinsurprenant:
@wainersm thanks for the PR. 
I am curious, how could this practically have an impact? The only way I see the `lock.unlock();` not being executed in
```java
try {
    FileLockFactory.getDefault().releaseLock(this.dirLock);
} catch (IOException e) {
    // log error and ignore
    logger.error("Queue close releaseLock failed, error={}", e.getMessage());
}
```
is if an unchecked exception is thrown. 
I agree that moving the `lock.unlock();` into a `finally` -will- ensure the mutex release in any circumstances but practically, an unchecked exception will bubble up regardless?

wainersm:
Hi @colinsurprenant !

I'm not sure I understood your question. Anyway, considering the old code, if it threw an IOException when it tried to unlock the file lock, then unlock.lock() is never called (i.e. the outer finally doesn't proceed).

BTW, I detected that problem by running findbugs. There are two other issues related with multi-threading, and many in other topics. How do you suggest to record those issues?

colinsurprenant:
@wainersm First, I agree your proposal will ensure mutex unlocking will always happen regardless of the nature of any exception in the `releaseLock() call`.

But for the IOException, I don't think it is the case, let's look at the original code:
```java
try {
    ...
} finally {
    try {
        FileLockFactory.getDefault().releaseLock(this.dirLock);
    } catch (IOException e) {
        logger.error("Queue close releaseLock failed, error={}", e.getMessage());
    }
    lock.unlock();
}
```

If the `releaseLock()` method throws an `IOException`, it will be catched and the error will be logged, then `lock.unlock()` will be called. Why do you think it won't in this case?

As I see it, the only circumstance (outside of JVM death or thread kill) where `lock.unlock()` might not be called is if an unchecked exception is thrown in the `releaseLock()` call.  WDYT?

colinsurprenant:
@wainersm for findbugs, good idea. depending on the amount, maybe you could create a single issue with the complete report, then we could sort these and possibly create separate issues for those we think deserve it? 
Otherwise I will install the plugin and run it here and see what gives :P 

Thanks a lot for your contributions on this!

colinsurprenant:
@wainersm So I ran findbugs and got the "close() does not release lock on all exception paths" warning. As I said above, to me this is the unchecked exceptions case which practically shouldn't be a problem but to be 100% correct it has to be in an extra `finally` block as you suggested. 

I verified the other warnings and I don't see nothing critical but will definitely address them. It's up to you if you still want to file an issue with the details.

wainersm:
Hi @colinsurprenant !  Cool!  If you going to address the other issues reported by findbugs, I don't think it is necessary to file an issue report.

I've made an experiment to test the finally block issue. Indeed, you are right, it needs an unchecked exception to jump over finally block. See below.

```shell
wainersm@tico-laptop:/tmp$ cat TestFinally.java 
import java.io.IOException;
import java.lang.RuntimeException;

public class TestFinally {

	private static void callUnlockFileFailIO() throws IOException {
		throw new IOException();
	}

	private static void callUnlockFileFailRuntime() throws IOException {
		throw new RuntimeException();
	}

	public static void main (String[] argv) {
		try {
			System.out.println("here I am: outer try");
		} finally {
			// Throws checked exception
			try {
				System.out.println("here I am: inner try 1");
				callUnlockFileFailIO();
			} catch (IOException e) {
				System.out.println("here I am: inner catch" +
					"(checked exception)");
			}
			System.out.println("here I am: outer finally 1");

			// Throws unchecked exception
                        try {
				System.out.println("here I am: inner try 2");
				callUnlockFileFailRuntime();
			} catch (IOException e) {
				// Never printed
				System.out.println("here I am: inner catch" +
					"(unchecked expection)");
			} finally {
				System.out.println("here I am: inner finally");
			}
			// Never printed
			System.out.println("Here I am: outer finally 2");
		}
	}
}
wainersm@tico-laptop:/tmp$ javac TestFinally.java 
wainersm@tico-laptop:/tmp$ java TestFinally
here I am: outer try
here I am: inner try 1
here I am: inner catch(checked exception)
here I am: outer finally 1
here I am: inner try 2
here I am: inner finally
Exception in thread "main" java.lang.RuntimeException
	at TestFinally.callUnlockFileFailRuntime(TestFinally.java:11)
	at TestFinally.main(TestFinally.java:31)
```

colinsurprenant:
@wainersm exactly! good stuff. thanks for your contribution! Will go ahead and merge.

colinsurprenant:
merged in master, 5.x, 5.4, 5.3.

PR #6920, opened 2017-04-13
Make pipeline and agent available through execution context

andrewvc:
This will let plugins inquire about stuff to the agent directly without using singletons.

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -27,9 +27,9 @@ require "logstash/execution_context"
 module LogStash; class BasePipeline
   include LogStash::Util::Loggable
 
-  attr_reader :settings, :config_str, :config_hash, :inputs, :filters, :outputs, :pipeline_id, :lir
+  attr_reader :settings, :config_str, :config_hash, :inputs, :filters, :outputs, :pipeline_id, :lir, :execution_context
 
-  def initialize(config_str, settings = SETTINGS, namespaced_metric = nil)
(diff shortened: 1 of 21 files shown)


ph:
@andrewvc This will need a rebase/small refactoring with the merge of #6632, when I've created the agent refactor, I almost decided to send a reference of the agent for the pipeline action instead of just the pipelines hash. I think I should have done that :(

logstash-core/spec/support/shared_contexts.rb
```diff
@@ -0,0 +1,13 @@
...
+  let(:agent) { double("agent") }
+  let(:execution_context) do
+    ::LogStash::ExecutionContext.new(pipeline, agent)
+  end
+  
+  before do
+    allow(pipeline).to receive(:pipeline_id).and_return(pipeline_id)
+    allow(pipeline).to receive(:agent).and_return(agent)
+  end
+end
```
ph: I like what you did here

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -147,7 +147,7 @@ module LogStash; class Pipeline < BasePipeline
 
   MAX_INFLIGHT_WARN_THRESHOLD = 10_000
 
-  def initialize(config_str, settings = SETTINGS, namespaced_metric = nil)
+  def initialize(config_str, settings = SETTINGS, namespaced_metric = nil, agent = nil)
```
ph: I don't think this work with the current master because the agent is never sent to the pipeline?  See https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/pipeline_action/create.rb#L40

PR #6930, opened 2017-04-18
acked queue concurrent stress spec

### LINKED ISSUE #6917: include PQ threaded spec as a normal test or integration test
this spec https://gist.github.com/colinsurprenant/b3ca535857a0e5d4ab48580fdfd653fe originally authored by @jordansissel and then greatly enhanced by @guyboertje and slightly fixed by myself in #6829 and #6901 should be included as a test or an integration test.

This is a spec which stress tests the PQ concurrency under various parameters. On my MBP is takes about 3 minutes to complete. 

I wouldn't mind adding this spec in the normal test suite but not sure if the added extra 3 minutes of "normal" specs would hurt our patience? 

WDYT?

colinsurprenant:
fixes #6917 

includes PQ threaded/concurrent stress tests in the default specs.

introduces two new test rake tasks:
- `rake test:core-fast` 
- `rake test:core-slow`

where `rake test:core` actually calls `rake test:core-slow` which runs all specs (excepted those excluded in dev-utils).
`rake test:core-fast` excludes the specs tagged with `:stress_test => true`. 

This PR does not clean the current more-or-less deprecated tagging and configured in dev-utils.

### DIFF
--- a/logstash-core/spec/logstash/acked_queue_concurrent_stress_spec.rb
+++ b/logstash-core/spec/logstash/acked_queue_concurrent_stress_spec.rb
@@ -1,0 +1,291 @@
+# encoding: utf-8
+require "logstash/util/wrapped_acked_queue"
+require "logstash/event"
+require "logstash/instrument/namespaced_metric"
+
+describe LogStash::Util::WrappedAckedQueue, :stress_test => true do
+  let(:path) { Stud::Temporary.directory }
+
+  context "with multiple writers" do
+    let(:items) { expected_count / writers }
+    let(:page_capacity) { 1 << page_capacity_multiplier }
+    let(:queue_capacity) { page_capacity * queue_capacity_multiplier }
+
+    let(:output_strings) { [] }
+    let(:reject_memo_keys) { [:reject_memo_keys, :path, :queue, :writer_threads, :collector, :metric, :reader_threads, :output_strings] }
+
+    let(:queue) do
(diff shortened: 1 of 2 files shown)


colinsurprenant:
The failing tests seems to be Agent related.

colinsurprenant:
the build failures are not related so upon LGTM I will go ahead and merge.

logstash-core/spec/logstash/acked_queue_concurrent_stress_spec.rb
```diff
@@ -0,0 +1,296 @@
+# encoding: utf-8
+require "logstash/util/wrapped_acked_queue"
+require "logstash/event"
+require "logstash/instrument/namespaced_metric"
+
+RSpec.configure do |config|
+  config.backtrace_exclusion_patterns << /rakelib\/test/
+  config.backtrace_exclusion_patterns << /vendor\/bundle/
+end
```
ph: What is this madness! Never seen it before!, But  I believe this setting will impact other example?

ph: It will impact tests after the `acked_queue_concurrent_stress_spec.rb` is loaded even if they are not run.
Should this be in the `spec_helper`?

colinsurprenant: good point. lets get rid of that.

colinsurprenant: fixed.

colinsurprenant:
rebased & squashed, merged into master, 5.x and 5.4

PR #6936, opened 2017-04-19
Cleanup some LIR java with feedback from Findbugs

andrewvc:
Ran the code through IntelliJ's findbugs plugin (thanks @colinsurprenant !) and fixed a few issues it found.

Nothing big here, it found an usused method and a potentially bad issue with default charsets, but other than that nothing with practical significance.


colinsurprenant:
LGTM

Issue #6846, opened 2017-03-27
Docs generation change


jordansissel:
Background:

Logstash plugins have had their documentation embedded as comments within the plugin code for 6+ years. This was a design choice intended to make plugin docs maintenance easier on me.

Over time, this design choice has started to become a burden on us because:

1) We moved from markdown to asciidoc
2) We moved from docs hosting on logstash.net (maintained by me) to elastic.co (maintained by Elastic)
3) We moved plugins from this repository (logstash) to each plugin having its own repository.

All of these above changes have made it harder to publish documentation from the original sources. In discussions, we spend lots of time thinking and working on the logstash plugin docs generator tooling.

----

Proposal: Each plugin will have an asciidoc file living in the same repository as the plugin and will contain the documentation for that plugin.

@ph, @dedemorton, @suyograo, and others contributed to the ideas and decisions here.

Desired impact:

* reduce time spent on troubleshooting documentation pipeline issues
* reduce time-to-publish on new plugin documentation
* reduce time spent previewing new docs in code review (no more logstash-docgen to run!)

Key measurable:

* Folks don't ask us about out-of-date or missing plugin documentation anymore.

Next steps for implementation:

* [x] For each plugin type, write a new asciidoc file in this (logstash) repository which details the settings provided by the base plugin classes. Example: the base filter class provides the `add_fields` setting.
* [x] Tweak the logstash-docgen tool to produce the per-plugin .asciidoc which will live in the same repository as the plugin.
* [x] Update the docs-tools/logstash build job[1] to also collect the new per-plugin asciidoc files from each repository. https://github.com/elastic/docs-tools/issues/3

[1] this build job ships our docs into the elastic.co docs pipeline.

ph:
For the plugin directory structure I propose this, to have a top level directory `docs` with the plugin documentation named `index.asciidoc`. This could allow complex plugins to have more pages when if/needed or images.

```
├── docs
│   └── index.asciidoc
├── lib
├── logstash-input-beats.gemspec
├── settings.gradle
├── spec
└── vendor.json

ph:
I have a script that creates the documentation for every plugin and strips the fields inherited from the base classes ((inputs|codecs|outputs|filters|/base.rb and plugin.rb), these information will be moved to the filter for each plugin type that will be included by the ascidoc.

Remaining steps for me:
[ ] - Add the include for each type
[ ] - Scripts that update each plugin with the new structure
[ ] - Create a PR for each plugin
[...]

PR #6941, opened 2017-04-20
Extraction of the document from ruby to actual asciidoc file

ph:
This PR is based on the proposal from #6846, it does the following

- Change the format of the plugin help page
- Create an external file for each plugin type that will get included by the plugin
- Add a script to generate the PR on each plugin repo with the static doc
- Add a script `ci/ci_docs.sh` that will be used to generate the list of plugins that can be installed in logstash from all the available plugins from the community organization

This PR is the building block for change in https://github.com/elastic/docs-tools/issues/3


ph:
Remaining tasks
- [ ] remove plugin information and use variables instead.
- [ ] merge plugin information with plugin header file.
- [ ] make the include path use a variable instead.

ph:
@dedemorton @jordansissel I've updated the PR with the changes we discussed on zoom.

I've also pushed a new version of the PR at https://github.com/logstash-plugins/logstash-filter-dissect/pull/21/files

And created an example repo with the generated html for all the plugin type at https://github.com/ph/example-doc

dedemorton:
LGTM. Thanks for the change!

ph:
@jordansissel I've rebased the branch with master, can you take a look at the format of the json file created by the script `ci/ci_docs.sh`?

docs/include/output.asciidoc
```diff
@@ -0,0 +1,61 @@
...
+---------------------------------------------------------------------------------------------------
+
+
+[id="plugins-{type}s-{plugin}-workers"]
+===== `workers`
+
+  * Value type is <<string,string>>
+  * Default value is `1`
+
+when we no longer support the :legacy type
```
dedemorton: @ph lines 60 and 61 need to be removed (they contain content that was previously in a TODO item).

ph:
@dedemorton I have updated the PR with the changes

ph:
@jordansissel So you can move forward in your changes https://gist.github.com/ph/9bc03a45d35851a9435e3c6d8377d281

ph:
@jordansissel I've rebased this PR with master, can you do a final review?

jordansissel:
@ph the script in this PR seems good to me. I haven't reviewed the docs changes yet, but I will have a review for you by end of day today.

docs/include/filter.asciidoc
```diff
@@ -0,0 +1,171 @@
...
+==== Common Options
+
+The following configuration options are supported by all filter plugins:
+
+[cols="<,<,<",options="header",]
+|=======================================================================
+|Setting |Input type|Required
+| <<plugins-{type}s-{plugin}-add_field>> |<<hash,hash>>|No
+| <<plugins-{type}s-{plugin}-add_tag>> |<<array,array>>|No
+| <<plugins-{type}s-{plugin}-enable_metric>> |<<boolean,boolean>>|No
```
jordansissel: Do we need enable_metric and id documented here if they are already documented in the base plugin docs?

Or to ask another way, do we need base plugin docs if we  include the base config settings in the plugin types?

rakelib/plugins_docs_dependencies.rake
```diff
@@ -0,0 +1,169 @@
...
+          end
+        else
+          possible_success << name
+        end
+      end
+
+      plugins_to_install = possible_success.concat(possible_failures).concat(unknown)
+      [plugins_to_install, plugins_to_install.size / possible_success.size]
+    else
+      [plugins_to_install, 2]
```
jordansissel: what's the significance of `2` here?

rakelib/plugins_docs_dependencies.rake
```diff
@@ -0,0 +1,169 @@
...
+
+  def extract_versions(definition, dependencies, from)
+    definition.specs.each do |spec|
+      dependencies[spec.name] ||= []
+      dependencies[spec.name] << VersionDependencies.new(spec.version, from)
+    end
+  end
+
+  def filter_logstash_plugins(specs)
+    specs.delete_if do |gem_name, _|
```
jordansissel: We can check if the plugin is a logstash plugin using `Gem::Specification#metadata` instead of these regex tests.

Example:

```
>> Gem::Specification.all.select { |x| x.metadata["logstash_plugin"] }.first
=> #<Gem::Specification:0x2c26e logstash-codec-cef-4.1.2-java>
```

Assuming the `specs` contain Gem::Specifications, anyway.

jordansissel:
I skipped most of the documentation generator (parse plugin code to produce asciidoc) since we have reviewed this previously in other PRs and have also already executed that step.

I left some inline comments.

I find the new code in rakelib a bit difficult to read, but this may be an artifact of the task (what plugins work for a given Logstash version?) being a difficult task.

Overall, LGTM.

rakelib/plugins_docs_dependencies.rake
ph: /facepalm, I will make that change it would be better.

ph:
@jordansissel Let me some time to go back in the code for answering your comments ;)

docs/include/filter.asciidoc
ph: @jordansissel but we don't include a base plugin doc? The filter/codec/input/output asciidoc content is a merge of what was originaly from the **plugin.rb base class** (where `enable_metric` is coming from) and the **plugin type base class**.

rakelib/plugins_docs_dependencies.rake
ph: This is where the magic happens; I am trying to optimize for successful **batch size**, I am brute forcing my way to reduce the number of resolutions required by installing multiple plugins at each iteration.

Let's take the _else_ condition; this is where you hit if you never run the tool before. I will take the plugins list split it into two groups and try to install them at once and recursively reduce the group size until I can install them. In the worse case I need to do a resolution for each plugins, at best (when successful) I do a resolution for a batch of plugins.

If you run the tool before, I will have the previous run cached. I will sort the plugins into two groups, plugins that were previously successful and the other. I am assuming that plugin that was installed successfully will still be installed successfully, so I will try to do a batch of plugins.


_Note I am recursively calling batch install that will reduce the batch for each iteration_

ph:
> I find the new code in rakelib a bit difficult to read, but this may be an artifact of the task (what plugins work for a given Logstash version?) being a difficult task.

The problem might be I am using a recursive algo and try to maximize batch size, when you run the tool, it give you feedback on each iteration.

docs/include/filter.asciidoc
jordansissel: Nevermind I totally misread one of the files here. I read the codecs doc but assumed it was some 'base plugin' doc.

Disregard my comment above

docs/include/codec.asciidoc
```diff
@@ -0,0 +1,41 @@
+[id="plugins-{type}s-common-options"]
```
jordansissel: I think we can remove this file. Codecs don't need any base docs like this. I think it is an accident that codecs have enable_metrics and id settings and we can remove that feature in the future, but for users, they don't need this.

ph:
[5.x f8526c003]
[5.5 8beeb380e]

Issue #6929, opened 2017-04-18
Test failling in master after the merge of the refactor of the agent/source loader.


ph:
```
Failures:

  1) LogStash::Agent Agent execute options when `config.reload.automatic` is set to `TRUE` and succesfully load the config converges periodically the pipelines from the configs source
     Failure/Error: expect(source_loader.fetch_count).to be > 1
       expected: > 1
            got:   1
     # ./logstash-core/spec/logstash/agent/converge_spec.rb:103:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # ./lib/bootstrap/rspec.rb:13:in `(root)'

Finished in 6.39 seconds (files took 2.48 seconds to load)
12 examples, 1 failure

Failed examples:

rspec ./logstash-core/spec/logstash/agent/converge_spec.rb:98 # LogStash::Agent Agent execute options when `config.reload.automatic` is set to `TRUE` and succesfully load the config converges periodically the pipelines from the configs source

Randomized with seed 8074

PR #6945, opened 2017-04-20
Make the periodic test converge a bit more robust

ph:
This test has been a bit flaky since it relies on an external thread to
trigger, this commit add a bit more time for the trigger to happen and
also add a retry.

Fixes: #6929

### DIFF
--- a/logstash-core/spec/logstash/agent/converge_spec.rb
+++ b/logstash-core/spec/logstash/agent/converge_spec.rb
@@ -110,9 +110,14 @@ describe LogStash::Agent do
           it "converges periodically the pipelines from the configs source" do
             agent_task = start_agent(subject)
 
-            sleep(interval * 10) # let the interval reload a few times
+            sleep(2) # let the interval reload a few times
             expect(subject).to have_running_pipeline?(pipeline_config)
-            expect(source_loader.fetch_count).to be > 1
+
+            # we rely on a periodic thread to call fetch count, we have seen unreliable run on
+            # travis, so lets add a few retries
+            try do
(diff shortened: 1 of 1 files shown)


logstash-core/spec/logstash/agent/converge_spec.rb
```diff
@@ -110,9 +110,14 @@
...
 
-            sleep(interval * 10) # let the interval reload a few times
+            sleep(2) # let the interval reload a few times
             expect(subject).to have_running_pipeline?(pipeline_config)
-            expect(source_loader.fetch_count).to be > 1
+
+            # we rely on a periodic thread to call fetch count, we have seen unreliable run on
+            # travis, so lets add a few retries
+            try do
+              expect(source_loader.fetch_count).to be > 1
```
jordansissel: what provides this 'try' method?

ph: Its coming from devutils at https://github.com/elastic/logstash-devutils/blob/master/lib/logstash/devutils/rspec/logstash_helpers.rb#L17

jordansissel: Ahh, an alias for Stud::try, gotcha.

jordansissel:
LGTM.

Long term, we should try to remove these sleeps if we can.

ph:
> Long term, we should try to remove these sleeps if we can. agree, but in the `converge_spec`, 

There is only one sleep to test the timer trigger, once everything is stabilized we can remove agent_spec, which has the most sleep in it.

Issue #6935, opened 2017-04-19
Faillure in master since merge of the refactor


ph:
```
Pending:
  LogStash::Agent#execute when auto_reload is true when calling reload_state! with a config that contains reload incompatible plugins does not upgrade the new config
    # Temporarily skipped with xcontext
    # ./logstash-core/spec/logstash/agent_spec.rb:257
  LogStash::Agent#upgrade_pipeline when the upgrade fails and current state is empty should not start a pipeline
    # Temporarily skipped with xcontext
    # ./logstash-core/spec/logstash/agent_spec.rb:374

Failures:

  1) LogStash::Config::Source::Local::ConfigPathLoader when it exist relative path returns alphabetically sorted parts
     Failure/Error: expect(parts.collect { |part| ::File.basename(part.source_id) }).to eq(files.keys.sort)
       
       expected: ["config1.conf"]
            got: []
       
       (compared using ==)
     # ./logstash-core/spec/logstash/config/source/local_spec.rb:72:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:44:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:194:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:180:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:152:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:117:in `run_with_threads'
[...]

PR #6946, opened 2017-04-20
Make the relative path more obvious

ph:
I've changed the relative path used in the test, I am using a relative
path that I can more easily control and doesn't have 10 level of nested
directories, I think it was confusing when ran in on our linux ci.

I also added an assert in the `before` to make sure we generate the
right number of configuration in the example.

Ref: #6935


colinsurprenant:
there are still some failures, not sure if they are related though...

ph:
I don't think they are there is an ongoing effort at https://github.com/elastic/logstash/pull/6754 to make the integration test more resilient.

jordansissel:
Travis failures seem unrelated. LGTM.

PR #6947, opened 2017-04-20
Warn users not to use mutliline codec when multiple hosts are configured

dedemorton:
@ppf2 @acchen97 Here's the update to the docs in the logstash repo. I'll need to open a separate request to add a note to the beats input plugin docs.

I also made a couple of fixes and moved text around a bit to improve the flow.

Can you review the text and let me know if it covers what it needs to cover and uses strong enough language? I'll put something similar in the beats input doc, and I'll also say something in the Beats doc. They're all separate PRs because the doc source lives in different repos.

This PR addresses a subset of the issues tracked in https://github.com/logstash-plugins/logstash-input-beats/issues/199

### DIFF
--- a/docs/static/managing-multiline-events.asciidoc
+++ b/docs/static/managing-multiline-events.asciidoc
@@ -5,10 +5,17 @@ Several use cases generate events that span multiple lines of text. In order to 
 Logstash needs to know how to tell which lines are part of a single event.
 
 Multiline event processing is complex and relies on proper event ordering. The best way to guarantee ordered log
-processing is to implement the processing as early in the pipeline as possible. The preferred tool in the Logstash
-pipeline is the {logstash}plugins-codecs-multiline.html[multiline codec], which merges lines from a single input using
+processing is to implement the processing as early in the pipeline as possible.
+
+The <<plugins-codecs-multiline>> codec is the preferred tool for handling multiline events
+in the Logstash pipeline. The multiline codec merges lines from a single input using
 a simple set of rules.
 
+IMPORTANT: If you are using a Logstash input plugin that supports multiple hosts, such as
+the <<plugins-inputs-beats>> input plugin, you should not use the
+<<plugins-codecs-multiline>> codec to handle multiline events. Doing so may result in the 
+mixing of streams and corrupted event data. In this situation, you need to handle multiline
+events before sending the event data to Logstash. 
 
 The most important aspects of configuring the multiline codec are the following:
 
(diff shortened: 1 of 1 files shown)


docs/static/managing-multiline-events.asciidoc
```diff
@@ -5,10 +5,17 @@ Several use cases generate events that span multiple lines of text. In order to
...
 Multiline event processing is complex and relies on proper event ordering. The best way to guarantee ordered log
-processing is to implement the processing as early in the pipeline as possible. The preferred tool in the Logstash
-pipeline is the {logstash}plugins-codecs-multiline.html[multiline codec], which merges lines from a single input using
-a simple set of rules.
+processing is to implement the processing as early in the pipeline as possible.
+
+IMPORTANT: If you are using a Logstash input plugin that supports multiple hosts, such as
+the <<plugins-inputs-beats>> input plugin, you should not use the
+<<plugins-codecs-multiline>> codec to handle multiline events. Doing so may result in the 
+mixing of strings and corrupted event data. In this situation, you need to handle multiline
```
ppf2: mixing of streams?

docs/static/managing-multiline-events.asciidoc
```diff
@@ -5,10 +5,17 @@ Several use cases generate events that span multiple lines of text. In order to
...
+events before sending the event data to Logstash. 
 
+The <<plugins-codecs-multiline>> codec is the preferred tool for handling multiline events
+in the Logstash pipeline. The multiline codec merges lines from a single input using
```
ppf2: in the Logstash pipeline (except for the scenarios mentioned in the `Important` disclaimer above).

docs/static/managing-multiline-events.asciidoc
acchen97: +1

docs/static/managing-multiline-events.asciidoc
acchen97: Is the "Important" section a note/banner? If it is, I think its ok as-is.

dedemorton: Yup, when formatted correctly in the HTML, it's definitely more clear that this sentence doesn't contradict what users see in the note box. I can see where it wouldn't be as obvious if you're reading the text without formatting.

ppf2: Yah I was suggesting only because there are users who don't read the documentation line by line, so if somehow they missed the important section, then they may think that multiline codec applies to all.  But we can just leave it as-is, since we already have the ticket open to block users from even attempting to configure the beats input with ML codec :)  What if we simply move the Important section to be right below the "preferred tool  ..." paragraph instead of above it?

dedemorton: sure, I can move the note.

dedemorton:
@ppf2 @acchen97 OK, I've moved the note. Can someone approve the PR so I can merge it and make the change in other places?

acchen97:
LGTM.

PR #6956, opened 2017-04-24
Make the Logstash metric check for the timestamp on reload success more robust

ph:
The current implementation of the test was using a mock and an expect
on the internal classes to determine when to start testing on the
metrics. I've rewrote the setup of the test to use the file output instead of using a
instance. I believe the previous code was not completely threadsafe and was causing this error
in the spec. We should really remove any mock of the form expect_any_instance_of`.

Ref: #6935


ph:
The integration test seems to fails on a unrelated issue.

ph:
@jsvd If you can review this PR.

ph:
I will make another PR for 5.x, too much changes happened.

Issue #6931, opened 2017-04-18
queue exclusive lock access errors in new agent specs


colinsurprenant:
Saw these on master.

```
rake test:core-fast
```
```
Sending Logstash's logs to  which is now configured via log4j2.properties
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true, :stress_test=>true} ............................./Users/colin/dev/src/elasticsearch/logstash/logstash-core/lib/logstash/compiler/lscl.rb:174 warning: `eval' should not be aliased .....................................................................................................................................................................................................................Plugin foo version 9.0.0.0 does not exist
............................................................................................Downloading file: http://localhost:8888/mypackage.zip
.Downloading file: http://localhost:8888/mypackage.zip
..........................................................................Updating dummy
...........................................................................................  queue.type: persisted
  queue.page_capacity: 512KB
  queue.max_bytes: 1048576KB
  workers: 8
  events: 100000
  took: 39.66199994087219s
.Installing file: /Users/colin/dev/src/elasticsearch/logstash/spec/unit/plugin_manager/pack_installer/../../../support/pack/valid-pack.zip
Install successful
..Installing file: /Users/colin/dev/src/elasticsearch/logstash/spec/unit/plugin_manager/pack_installer/../../../support/pack/empty-pack.zip
.Installing file: /var/folders/n3/9bfxzn1s7c74zc8y_p91dd1m0000gp/T/studtmp-221078c1ca41df571368b574c20d24b309dcfe656a4e273fb48f3dcf6735/1492552204.zip
..................WARNING: A maven settings file already exist at /Users/colin/.m2/settings.xml, please review the content to make sure it include your proxies configuration.
....WARNING: A maven settings file already exist at /Users/colin/.m2/settings.xml, please review the content to make sure it include your proxies configuration.
...........................................................................................................................................................................................................................................................................................................................................................................................................................................................................[2017-04-18T15:50:17,697][ERROR][logstash.shutdownwatcher ] The shutdown process appears to be stalled due to busy or blocked plugins. Check the logs for more information.
.[2017-04-18T15:50:18,742][ERROR][logstash.shutdownwatcher ] The shutdown process appears to be stalled due to busy or blocked plugins. Check the logs for more information.
[...]

PR #6957, opened 2017-04-25
Fix the lock errors in the LogStash::Runner test

ph:
The test was actually starting an agent with pipeline and the runner has
currently no way to stop the agent. This commit make sure we use an
agent mock instead.

Fix: #6931


jordansissel:
LGTM

PR #6966, opened 2017-04-25
Fix warning about eval being aliased in lscl.rb

andrewvc:
This warning was generated due to a java class being used that had its own `eval` method which wasn't even being used.


suyograo:
LGTM

PR #6967, opened 2017-04-25
Add '.gradle' to .gitignore file

andrewvc:
It's annoying that `git add logstash-core` sweeps in the whole `logstash-core/.gradle` dir.


suyograo:
LGTM

PR #6975, opened 2017-04-26
Improve docs clarity for the update subcommand

geekpete:
A docs fix to show that the update command will only update to the latest minor version of a major plugin release with an example to demonstrate.


geekpete:
I tested this in Logstash 5.4 and update installed the very latest version, eg 7.3.1.

So this behaviour is different in different versions.

geekpete:
Will close this for now and resubmit when if there's a need to provide clarity for specific versions

Issue #6926, opened 2017-04-17
Define how logstash behaves with multiple pipelines feature


jsvd:
TL;DR Conclusions for 6.x:

1. The multiple pipelines are declared in a YAML file. To make reload semantics easier to understand, this file will be separate from the global `logstash.yml`, called `pipelines.yml`.
2. Starting logstash without arguments now will read pipelines from the yaml file instead of returning error.
3. The three methods of specifying pipelines (pipelines.yml, `-f` and `-e`) will become mutually exclusive between each other.
3.1. This means `-e` will no longer perform concatenation with `-f` (or `pipelines.yml` pipelines).
4. Detecting the absence and autocompletion of input/output sections in the config will be restricted to `-e` (no longer auto completes for `-f`).
5. When reloading is on, `logstash.yml` won't be polled for changes, only `pipelines.yml` will.
6. There will be a tool to generate a `pipelines.yml` for you if you're migrating your `-f` usage from 5.x to 6.x
7. For the Web API, the endpoints that refer to a single pipeline will redirect to `/pipelines/:id` where `:id` will be extracted from `LogStash::SETTINGS.get("pipeline.id")`

--------------------------------------

With the upcoming multiple pipeline targetted at 6.0, it may be interesting to clarify how logstash works when defining and declaring pipelines, either through the cli or configuration files.

Pre MP (multi pipeline), logstash accepts:
  * from the CLI: either -f or -e, or both
  * from the settings.yml: either `config.string` or `path.config` or both

With MP, there is an extra source of configurations: there is a pipelines.yml file where pipelines (their configs and their settings) can be declared. Pipelines in the yaml file also can declare their configurations inline (using `config.string`) or point to files (`path.config`).

Some questions start to pop up:
1. should it be possible to declare pipelines in the pipelines.yml and also use -f and -e too? What about only -e?
2. should running logstash without arguments (`bin/logstash`) now default to using the `pipelines.yml` file? Currently it fails since neither -f nor -e were set.
3. While -e is a very useful tool for experimenting and debugging in Logstash, is it worth creating a line between logstash running in single pipeline mode (`-f`) or with 1 or more pipelines through the `pipelines.yml` file?

Here are my proposals:
1. Eliminate -f so that pipeline descriptions are done through the yaml file
2. Keep -e but make it create an isolated pipeline instead of supporting concatenation with file-based pipelines.
[...]

PR #6976, opened 2017-04-26
make e and f flags mutually exclusive

jsvd:
makes -e and -f mutually exclusive, which also eliminates concatenation between the two

this is part of the changes described in https://github.com/elastic/logstash/issues/6926


suyograo:
@jsvd remember to tag it as breaking :)

PR #6977, opened 2017-04-26
[WIP] Making the integration test suite a bit more robust to run on travis

ph:
This is an experiment and its not ready for review, the api is subject
to changes but I wanted to test it out as fast as possible.

Instead of checking on the TCP port open (which we can still do), I
inspect the stdout and check if the pipeline are started at this point.
I am aware that it couple this code to the output of the log but, I
think it will be more resilient in the short term.

Also instead of using a fix amount of retries I use a time limit
instead and when we run out of time I will print the content of the
stdout.

### DIFF
--- a/qa/integration/services/logstash_service.rb
+++ b/qa/integration/services/logstash_service.rb
@@ -9,7 +9,7 @@ require "uri"
 
 # A locally started Logstash service
 class LogstashService < Service
-  API_URL = "http://localhost:9600/"
+  API_URL = "http://localhost:%s/"
 
   LS_ROOT_DIR = File.join("..", "..", "..", "..")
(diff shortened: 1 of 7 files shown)


ph:
I am currently running that PR through travis a few times in a row to hopefully check if my changes fixed the situation.

ph:
Remaining failures, Didn't reproduce locally yet.

```
  1) Test Logstash Slowlog should write logs to a new dir
     Failure/Error: @ls.wait_for_logstash
     RuntimeError:
       Logstash's API didn't answer corrrectly within the specified maximum time of 300
     # ./services/logstash_service.rb:228:in `wait_for_logstash_api'
     # ./specs/slowlog_spec.rb:40:in `(root)'
```

ph:
Not really better than what it was.

Issue #6868, opened 2017-04-03
Test failure: pipeline.workers expected not to be set.


jordansissel:
Build failure: https://logstash-ci.elastic.co/job/elastic+logstash+5.x+dockerubuntu-periodic/1582/console

```
Failures:

  1) LogStash::Runner pipeline settings when :pipeline_workers is not defined by the user should not pass the value to the pipeline
     Failure/Error: expect(settings.set?("pipeline.workers")).to be(false)
       
       expected false
            got true
     # ./logstash-core/spec/logstash/runner_spec.rb:261:in `(root)'
     # ./logstash-core/lib/logstash/runner.rb:360:in `create_agent'
     # ./logstash-core/lib/logstash/runner.rb:272:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67:in `run'
     # ./logstash-core/lib/logstash/runner.rb:185:in `run'
```

Issue #6980, opened 2017-04-26
Refactor the usage of the settings class


ph:
When debugging the following issue https://github.com/elastic/logstash/issues/6868, I had to look everywhere in the code where the default workers settings was changed. The problem was raised in the `pipeline_spec` but the values was changed in the `pipeline_pq_spec`.

We have to refactor our tests/classes to take into consideration that `SETTINGS` is a singleton.

colinsurprenant:
An observation: it is IMO an anti-pattern to use a global, singleton object for settings **that is passed around** throughout the code. It makes testing extremely difficult. Classes should typically have no knowledge of the actual settings object, but receive their required parameters at instantiation, that way any side effects are minimized and tests don't have to rely of the global settings.

ph:
@colinsurprenant totally agree with that, I don't think the changes will be that hard to make in the code to remove the singleton.

PR #6981, opened 2017-04-26
Fix when settings set in the pipeline_pq_file_spec bleed into other examples

ph:
make sure we clone the LogStash::Settings before mutating it.

Fixes: #6868 and other related failures about the workers.


ph:
This is an _OK_ fixe to solve the current problem of the test suite but we need to look into #6980 to come up with a cleaner suite.

andrewvc:
LGTM

Issue #6978, opened 2017-04-26
Test failure for: LogStash::Agent Environment Variables In Configs environment variable templating are evaluated at plugins creation


ph:
```
 1) LogStash::Agent Environment Variables In Configs environment variable templating are evaluated at plugins creation
     Failure/Error: expect(subject.converge_state_and_update).to be_a_successful_converge
       Expected all actions to be successful:
           pipeline_id: main, message: Could not execute action: LogStash::PipelineAction::Create/pipeline_id:main
     # ./logstash-core/spec/logstash/agent_spec.rb:333:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:47:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:194:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:216:in `invoke_prerequisites'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:214:in `invoke_prerequisites'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:193:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:180:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:152:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:108:in `top_level'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:108:in `top_level'
[...]

PR #6982, opened 2017-04-26
Keep track of the original response of the pipeline_action's execute

ph:
This commit is to provide more information when an action fails and will
allow us hopefully to debug a random failure in the test suite for the
environment variable.

ref: #6978


jordansissel:
LGTM

PR #6985, opened 2017-04-27
Improve Gitignore to Make Build Run without Polluting the Index

original-brownbear:
Just a suggestion (though maybe I'm missing part of the workflow when testing) :)

Every time I run the tests, they create `Gemfile.jruby-1.9.lock` and
`.lock`. I think those can safely be git-ignored right?
Also, the tests edit the `Gemfile` which is part of the repo when being run. Wouldn't it make sense to exclude the Gemfile and simply recreate from a template when building like I did in this PR?


ph:
I would certainly love to remove the Gemfile / lock from the repo, but we also use theses file in other branches to control what is included and at what version. See https://github.com/elastic/logstash/blob/v5.3.2/Gemfile.jruby-1.9.lock

This allows us to grab a specific tag and recreate a release with the same libraries versions, maybe we could use another file?

original-brownbear:
@ph you mean use a similar pattern to how I handled the `Gemfile` (other file -> then copy to the ignored `Gemfile.jruby-1.9.lock`)?

andrewvc:
@original-brownbear I like that approach, makes sense to me . I'm not sure what the repercussions would be for our build process. @ph, @suyograo does that have significant effects there?

original-brownbear:
@andrewvc I think the problem that @ph points out is this:

There are branches for which for which a fixed version set has to be committed as a `Gemfile.jruby-1.9.lock`. So if this file goes to the `.gitignore` things become pretty awkward.

One clean solution may be to use a trivial hierarchy of 2 files e.g. `Gemfile.jruby-1.9.lock.template` and `Gemfile.jruby-1.9.lock.release`.
Then you could simply copy `Gemfile.jruby-1.9.lock.release` to `Gemfile.jruby-1.9.lock` if it exists (in a release branch) and use `Gemfile.jruby-1.9.lock` otherwise.
=> get a clean (git-wise) build
=> keep the semantics of the release branch having an extra file for locking in dependencies

makes sense?

ph:
@andrewvc I think the last proposal of @original-brownbear would work, this will certainly have an impact at how we generate releases that could be solved by a script.

The downside of having the lock in the master branch is maybe it will generate weird conflict errors? (other than the release).

original-brownbear:
@ph see my `Edit:` above, my bad ... no `.lock` checked into `master` needed I think.

ph:
@original-brownbear I am +1 for that proposal.

suyograo:
+1. @original-brownbear your approach should work for release branches and master.

original-brownbear:
@ph @suyograo updated the PR with the above suggestion :)

lib/bootstrap/bundler.rb
```diff
@@ -95,6 +95,19 @@ def invoke!(options = {})
 
       require "bundler"
       require "bundler/cli"
+
+      # create Gemfile from template iff it does not exist
+      unless ::File.exists?(Environment::GEMFILE_PATH)
+        ::FileUtils.copy(
+          ::File.join(ENV["LOGSTASH_HOME"], "Gemfile.template"), Environment::GEMFILE_PATH
```
ph: Can we replace `ENV["LOGSTASH_HOME"]` with `Environment::LOGSTASH_HOME`?

lib/bootstrap/bundler.rb
```diff
@@ -95,6 +95,19 @@ def invoke!(options = {})
...
+        )
+      end
+      # create Gemfile.jruby-1.9.lock from template iff a template exists it itself does not exist
+      lock_template = ::File.join(ENV["LOGSTASH_HOME"], "Gemfile.jruby-1.9.lock.release")
+      if File.exists?(lock_template) && !::File.exists?(Environment::LOCKFILE)
+        ::FileUtils.copy(lock_template, Environment::LOCKFILE)
```
ph: Can you require "FileUtils" at the top of the file? It may or may not be required at that time, we just prefer to be explicit.

ph: Can we remove the double `::` before the FileUtils? it's actually needed for the `::File` module because we have a plugin that clash the namespace (LogStash::Inputs::File)

lib/bootstrap/bundler.rb
```diff
@@ -95,6 +95,19 @@ def invoke!(options = {})
...
+
+      # create Gemfile from template iff it does not exist
+      unless ::File.exists?(Environment::GEMFILE_PATH)
+        ::FileUtils.copy(
+          ::File.join(ENV["LOGSTASH_HOME"], "Gemfile.template"), Environment::GEMFILE_PATH
+        )
+      end
+      # create Gemfile.jruby-1.9.lock from template iff a template exists it itself does not exist
+      lock_template = ::File.join(ENV["LOGSTASH_HOME"], "Gemfile.jruby-1.9.lock.release")
+      if File.exists?(lock_template) && !::File.exists?(Environment::LOCKFILE)
```
ph: missing `::` before the `File.exists`

.gitignore
```diff
@@ -33,6 +33,9 @@ qa/.vm_ssh_config
 qa/.vagrant
 qa/acceptance/.vagrant
 qa/Gemfile.lock
+Gemfile.jruby-1.9.lock
+.lock
```
ph: why we ignore .lock?

original-brownbear: @ph I had this being left over once or twice already(likely from dirty killing a build).  No reason not to ignore it right, we don't ever want to commit that? :)

original-brownbear:
@ph all fixed as requested I think :)

.gitignore
ph: Weird, I haven't seen that one before, lets remove it. so when it happen again we can find it.

original-brownbear:
@ph removed the `.lock` ignore :)

original-brownbear:
@ph interestingly enough currently `master` leaves a `.lock` after a clean run of `rake test:core` ... I wonder if this is indicative of some issue maybe (now that it happens for a green build)?

Issue #6986, opened 2017-04-27
5.4.0 undefined method 'dlq_writer' 


LeeDr:
- Version:  5.4.0-797cd833 (.deb file installed and started as a service)
- Operating System: Ubuntu 14.04 
- Config File (if you have sensitive info, please remove it): 
```
root@packer-virtualbox-iso-1492351135:~# cat /etc/logstash/conf.d/logstash.conf
input {
  file {
    path => "/var/log/syslog"
    start_position => "beginning"
  }
}

filter {
  if [type] == "syslog" {
    grok {
      match => { "message" => "%{SYSLOGTIMESTAMP:syslog_timestamp} %{SYSLOGHOST:syslog_hostname} %{DATA:syslog_program}(?:\[%{POSINT:syslog_pid}\])?: %{GREEDYDATA:syslog_message}" }
      add_field => [ "received_at", "%{@timestamp}" ]
      add_field => [ "received_from", "%{host}" ]
    }
    date {
      match => [ "syslog_timestamp", "MMM  d HH:mm:ss", "MMM dd HH:mm:ss" ]
    }
  }
}

output {
  elasticsearch {
    hosts => ["https://localhost:9200"]
    user => "logstash_internal"
    password => "changeme"
    ssl => true
    cacert => "/etc/logstash/sslconfig/ca.crt"
  }
}
```

```
root@packer-virtualbox-iso-1492351135:~# grep -v "^#" /etc/logstash/logstash.yml
path.data: /var/lib/logstash
path.config: /etc/logstash/conf.d
path.logs: /var/log/logstash
xpack.monitoring.elasticsearch.url: https://localhost:9200
xpack.monitoring.elasticsearch.username: logstash_system
xpack.monitoring.elasticsearch.password: changeit
xpack.monitoring.elasticsearch.ssl.ca: "/etc/logstash/sslconfig/ca.crt"
```

- Sample Data: just a Ubuntu /var/log/syslog
- Steps to Reproduce: start logstash on Ubuntu with above config

```
[...]

PR #6987, opened 2017-04-27
pin logstash-output-elasticsearch to 6.x

jsvd:
fixes https://github.com/elastic/logstash/issues/6986


ph:
@jsvd should  we update the lock file also?

ph:
nevermind its already set to 6.3. https://github.com/elastic/logstash/blob/5.4/Gemfile.jruby-1.9.lock#L427

jsvd:
after further investigation, I think logstash 5.4 should be allowed to upgrade to 7.x at will, the actual problem is described in https://github.com/elastic/logstash/issues/6986#issuecomment-297743327
@ph what do you think of fixing the es output and closing this issue?

ph:
@jsvd I also agree that we need to fix the es output 7.x to make sure its installable in Logstash. 2, we need to make sure the x-pack doesn't ship with that version if we ship logstash with 6.x.

Issue #6820, opened 2017-03-18
Document popular processing capabilities of Logstash


acchen97:
Logstash users often have a hard time discovering what are the processing capabilities, especially when getting started. This is a bug and we should improve this experience. Let's add a new top level doc section called "Data Transformation" with 4 sub-sections "Core Operations," "Data Deserialization," "Field Extraction," and "Lookup Enrichment." Plugins to highlight in each section are below.

For each plugin / use case, we should have:
- Short summary (maybe one or two sentences) describing what the plugin does
- At least one code snippet example of how it can be used. This could warrant >1 example for multi use case plugins like mutate and grok.
- Link to the respective plugin doc page for more details

## Data Transformation (top level section)
- Core Operations
  - Mutate
  - Drop
  - Date
  - Fingerprint/Anonymize
  - Aggregate
  - Ruby
- Data Deserialization (aside from native inputs)
  - JSON codec
  - XML
  - CSV
  - Avro codec
  - Protobuf codec
  - Fluent || Msgpack codecs
- Field Extraction (Data Wrangling)
  - Grok
  - Dissect
  - KV
- Lookup Enrichment
  - GeoIP
  - UserAgent
  - DNS
  - Translate
  - Elasticsearch
  - JDBC (2 plugins: static & streaming)

acchen97:
@dedemorton let's work on this one together. :)

dedemorton:
Fixed in #6988

PR #6988, opened 2017-04-27
Document popular processing capabilities

dedemorton:
Adds doc enhancement requested in https://github.com/elastic/logstash/issues/6820

@acchen97 Note that this content can't be formatted in tables because the examples would cause text overruns in the formatted doc. The best compromise is definition lists. I've broken the content into separate topics, but we could just as easily have one topic. I think separate topics are easier to skim. 

@suyograo or @jordansissel Hoping that one of you can also review this. Our plugin docs are woefully lacking in examples and some of the descriptions are hard to follow. :-/ Where possible, I just reused examples from the docs, but in some cases, I had to create new examples. There are a couple of places where I need help coming up with a good example because I didn't have time to create it myself or felt the example in the existing doc was too complicated for an overview.

ALL: Please respond to questions that I've added as comments (addressed to either you or general reviewers).

Here's a screen grab that shows the basic info architecture of these topics:

image

### DIFF
--- a/docs/static/transforming-data.asciidoc
+++ b/docs/static/transforming-data.asciidoc
@@ -11,37 +11,14 @@ according to their processing capabilities:
 * <<field-extraction>>
 * <<lookup-enrichment>>
 
+Also see <<filter-plugins>> and <<codec-plugins>> for the full list of available
+data processing plugins.
+
 [[core-operations]]
 === Performing Core Operations
 
 The plugins described in this section are useful for core operations, such as
-aggregating, mutating, and dropping events.
-
-Also see <<filter-plugins>> for the full list of available filter plugins. 
-
-<<plugins-filters-aggregate,aggregate filter>>::
-
-Aggregates information from several events (typically log lines) that belong to
-the same task and pushes the aggregated information into a final task event.
-For example:
-+
-[source,json]
---------------------------------------------------------------------------------
-filter {
-  aggregate {
-    task_id => "%{country_name}"
-    code => "
-      map['country_name'] = event.get('country_name')
-      map['towns'] ||= []
-      map['towns'] << {'town_name' => event.get('town_name')}
-      event.cancel()
-    "
-    push_previous_map_as_event => true
-    timeout => 3
-  }
-}
---------------------------------------------------------------------------------
-
+mutating and dropping events.
 
 <<plugins-filters-date,date filter>>::
 
@@ -64,15 +41,13 @@ filter {
 
 Drops events. This filter is typically used in combination with conditionals.
 +
-The following config drops 40% of the `debug` level log messages:
+The following config drops `debug` level log messages:
 +
 [source,json]
 --------------------------------------------------------------------------------
 filter {
   if [loglevel] == "debug" {
-    drop {
-      percentage => 40
-    }
+    drop { }
   }
 }
 --------------------------------------------------------------------------------
@@ -113,6 +88,18 @@ filter {
   }
 }
 --------------------------------------------------------------------------------
++
+The following config strips leading and trailing whitespace from the specified
+fields:
++
+[source,json]
+--------------------------------------------------------------------------------
+filter {
+  mutate {
+    strip => ["field1", "field2"]
+  }
+}
+--------------------------------------------------------------------------------
 
 
 <<plugins-filters-ruby,ruby filter>>::
@@ -136,9 +123,6 @@ filter {
 
 The plugins described in this section are useful for deserializing data into
 Logstash events.
-
-Also see <<filter-plugins>> and <<codec-plugins>> for the full list of available
-data processing plugins.
 
 <<plugins-codecs-avro,avro codec>>::
 
@@ -199,8 +183,6 @@ input {
 --------------------------------------------------------------------------------
 
 
-//ALVIN: SHOULD WE ALSO COVER THE MSGPACK CODEC? I LOOKED AT THE DOCS FOR MSGPACK AND THERE'S NO USEFUL INFO IN THERE. 
-
 <<plugins-codecs-json,json codec>>::
 
 Decodes (via inputs) and encodes (via outputs) JSON formatted content, creating
@@ -218,29 +200,29 @@ input {
 --------------------------------------------------------------------------------
 
 
-//ALVIN: SHOULD WE INCLUDE JSON_LINES HERE, TOO, OR MAYBE MENTION IT?
-
 <<plugins-codecs-protobuf,protobuf codec>>::
 
 Reads protobuf encoded messages and converts them to Logstash events. Requires
-the protobuf definitions to be compiled as Ruby files.
+the protobuf definitions to be compiled as Ruby files. You can compile them by
+using the
+https://github.com/codekitchen/ruby-protocol-buffers[ruby-protoc compiler].
 +
 The following config decodes events from a Kafka stream:
 +
 [source,json]
 --------------------------------------------------------------------------------
-kafka {
-  zk_connect => "127.0.0.1"
-  topic_id => "your_topic_goes_here"
-  codec => protobuf {
-    class_name => "Animal::Unicorn"
-    include_path => ['/path/to/protobuf/definitions/UnicornProtobuf.pb.rb']
+input
+  kafka {
+    zk_connect => "127.0.0.1"
+    topic_id => "your_topic_goes_here"
+    codec => protobuf {
+      class_name => "Animal::Unicorn"
+      include_path => ['/path/to/protobuf/definitions/UnicornProtobuf.pb.rb']
+    }
   }
 }
 --------------------------------------------------------------------------------
 
-
-//REVIEWERS: PLEASE DOUBLE-CHECK THE SYNTAX FOR THE PROTOBUF EXAMPLE. I CHANGED THE FORMATTING TO BE CONSISTENT WITH THE REST OF THE DOCUMENTATION HERE, BUT DIDN'T TEST THIS EXAMPLE.
 
 <<plugins-filters-xml,xml filter>>::
 
@@ -264,8 +246,6 @@ filter {
 The plugins described in this section are useful for extracting fields and
 parsing unstructured data into fields.
 
-Also see <<filter-plugins>> for the full list of available filter plugins. 
-
 <<plugins-filters-dissect,dissect filter>>::
 
 Extracts unstructured event data into fields by using delimiters. The dissect
@@ -273,29 +253,54 @@ filter does not use regular expressions and is very fast. However, if the
 structure of the data varies from line to line, the grok filter is more
 suitable.
 +
-The following config dissects messages that match the structure specified in
-the mapping:
+For example, let's say you have a log that contains the following message:
++
+[source,json]
+--------------------------------------------------------------------------------
+Apr 26 12:20:02 localhost systemd[1]: Starting system activity accounting tool...
+--------------------------------------------------------------------------------
++
+The following config dissects the message:
 +
 [source,json]
 --------------------------------------------------------------------------------
 filter {
   dissect {
-    mapping => {
-      "message" => "%{ts} %{+ts} %{+ts} %{src} %{} %{prog}[%{pid}]: %{msg}"
-    }
+    mapping => { "message" => "%{ts} %{+ts} %{+ts} %{src} %{prog}[%{pid}]: %{msg}" }
   }
 }
 --------------------------------------------------------------------------------
-
-
-//REVIEWERS: I DON'T REALLY UNDERSTAND HOW THIS WORKS FROM READING THE DOCS. CAN SOMEONE SUGGEST AN EXAMPLE THAT INCLUDES THE MESSAGE BEING DISSECTED (THAT WOULD  HELP USERS A LOT). THE DOCUMENTATION FOR THIS FILTER IS HARD TO FOLLOW IN GENERAL.
++
+After the dissect filter is applied, the event will be dissected into the following
+fields:
++
+[source,json]
+--------------------------------------------------------------------------------
+{
+  "msg"        => "Starting system activity accounting tool...",
+  "@timestamp" => 2017-04-26T19:33:39.257Z,
+  "src"        => "localhost",
+  "@version"   => "1",
+  "host"       => "localhost.localdomain",
+  "pid"        => "1",
+  "message"    => "Apr 26 12:20:02 localhost systemd[1]: Starting system activity accounting tool...",
+  "type"       => "stdin",
+  "prog"       => "systemd",
+  "ts"         => "Apr 26 12:20:02"
+}
+--------------------------------------------------------------------------------
 
 <<plugins-filters-kv,kv filter>>::
 
 Parses key-value pairs.
 +
 For example, let's say you have a log message that contains the following
-key-value pairs: `ip=1.2.3.4 error=REFUSED`.
+key-value pairs:
++
+[source,json]
+--------------------------------------------------------------------------------
+ip=1.2.3.4 error=REFUSED
+--------------------------------------------------------------------------------
 +
 The following config parses the key-value pairs into fields:
 +
@@ -353,8 +358,6 @@ After the filter is applied, the event in the example will have these fields:
 The plugins described in this section are useful for enriching data with
 additional info, such as GeoIP and user agent info.
 
-Also see <<filter-plugins>> for the full list of available filter plugins. 
-
 <<plugins-filters-dns,dns filter>>::
 
 Performs a standard or reverse DNS lookup.
@@ -377,13 +380,31 @@ filter {
 
 Copies fields from previous log events in Elasticsearch to current events. 
 +
+The following config shows a complete example of how this filter might
+be used.  Whenever Logstash receives an "end" event, it uses this Elasticsearch
+filter to find the matching "start" event based on some operation identifier.
+Then it copies the `@timestamp` field from the "start" event into a new field on
+the "end" event.  Finally, using a combination of the date filter and the
+ruby filter, the code in the example calculates the time duration in hours
+between the two events.
++
 [source,json]
 --------------------------------------------------
-NEED A SIMPLE EXAMPLE HERE. 
+      if [type] == "end" {
+         elasticsearch {
+            hosts => ["es-server"]
+            query => "type:start AND operation:%{[opid]}"
+            fields => { "@timestamp" => "started" }
+         }
+         date {
+            match => ["[started]", "ISO8601"]
+            target => "[started]"
+         }
+         ruby {
+            code => 'event.set("duration_hrs", (event.get("@timestamp") - event.get("started")) / 3600) rescue nil'
+         }
+      }
 --------------------------------------------------
-
-
-//REVEIWERS: I NEED A BETTER EXAMPLE FOR THE ELASTICSEARCH FILTER. THE EXAMPLE IN THE DOCUMENTATION MIGHT BE TOO COMPLEX FOR AN OVERVIEW.
 
 
 <<plugins-filters-geoip,geoip filter>>::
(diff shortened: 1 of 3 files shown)


docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
++
+The following config dissects messages that match the structure specified in
+the mapping:
++
+[source,json]
+--------------------------------------------------------------------------------
+filter {
+  dissect {
+    mapping => {
+      "message" => "%{ts} %{+ts} %{+ts} %{src} %{} %{prog}[%{pid}]: %{msg}"
```
jordansissel: Let's use this example (which avoids the possibly confusing empty  `%{}` usage):

```
mapping => { "message" => "%{ts} %{+ts} %{+ts} %{src} %{prog}[%{pid}]: %{msg}" 
```

Sample input:

```
Apr 26 12:20:02 localhost systemd[1]: Starting system activity accounting tool...
```

Sample output (as through `stdout { codec => rubydebug }`:

```
{
  "msg"        => "Starting system activity accounting tool...",
  "@timestamp" => 2017-04-26T19:33:39.257Z,
  "src"        => "localhost",
  "@version"   => "1",
  "host"       => "localhost.localdomain",
  "pid"        => "1",
  "message"    => "Apr 26 12:20:02 localhost systemd[1]: Starting system activity accounting tool...",
  "type"       => "stdin",
  "prog"       => "systemd",
  "ts"         => "Apr 26 12:20:02"
}
```

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+
+Copies fields from previous log events in Elasticsearch to current events. 
++
+[source,json]
+--------------------------------------------------
+NEED A SIMPLE EXAMPLE HERE. 
+--------------------------------------------------
+
+
+//REVEIWERS: I NEED A BETTER EXAMPLE FOR THE ELASTICSEARCH FILTER. THE EXAMPLE IN THE DOCUMENTATION MIGHT BE TOO COMPLEX FOR AN OVERVIEW.
```
jordansissel: +1 for needing an example. I don't remember ever using this filter, but I will try to come up with an example.

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+The following example executes a SQL query and stores the result set in a field
+called `country_details`:
++
+[source,json]
+--------------------------------------------------------------------------------
+filter {
+  jdbc_streaming {
+    jdbc_driver_library => "/path/to/mysql-connector-java-5.1.34-bin.jar"
+    jdbc_driver_class => "com.mysql.jdbc.Driver"
+    jdbc_connection_string => ""jdbc:mysql://localhost:3306/mydatabase"
```
jordansissel: too many doublequotes before the value `=> "jdbc:mysql...`

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+    parameters => { "code" => "country_code"}
+    target => "country_details"
+  }
+}
+--------------------------------------------------------------------------------
+
+
+//REVIEWERS: THE MYSQL CONNECTOR SHOWN IN THIS EXAMPLE IS KIND OF OLD. NOT SURE I SHOULD CHANGE IT, THOUGH, BECAUSE I'M NOT SURE IF WE SUPPORT ALL VERSIONS OF THE CONNECTOR.
+
+//ALVIN: YOU ALSO HAD JDBC STATIC PLUGIN IN YOUR LIST. DO YOU MEAN THE INPUT PLUGIN? SEEMS A LITTLE OUT-OF-PLACE TO LIST AN INPUT PLUGIN IN THIS SECTION, IMO. DOES IT DO SOMETHING BEYOND WHAT INPUT PLUGINS NORMALLY DO SO THAT IT
```
jordansissel: There is a jdbc_static filter that is [in progress](https://github.com/logstash-plugins/logstash-filter-jdbc_static/pulls) but not yet released.

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+    jdbc_password => "secret"
+    statement => "select * from WORLD.COUNTRY WHERE Code = :code"
+    parameters => { "code" => "country_code"}
+    target => "country_details"
+  }
+}
+--------------------------------------------------------------------------------
+
+
+//REVIEWERS: THE MYSQL CONNECTOR SHOWN IN THIS EXAMPLE IS KIND OF OLD. NOT SURE I SHOULD CHANGE IT, THOUGH, BECAUSE I'M NOT SURE IF WE SUPPORT ALL VERSIONS OF THE CONNECTOR.
```
jordansissel: I don't know if we specify what connectors (or versions of a connector) are supported. @guyboertje maybe has ideas?

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+
+<<plugins-filters-aggregate,aggregate filter>>::
+
+Aggregates information from several events (typically log lines) that belong to
+the same task and pushes the aggregated information into a final task event.
+For example:
++
+[source,json]
+--------------------------------------------------------------------------------
+filter {
```
dedemorton: @jordansissel Right. I understand your concern. I've sorted the options alphabetically just because I think it's easier for people to scan. This topic isn't a reference topic, though, so I would be OK with listing the plugins in order of usefulness. Should I use the order that Alvin gives here? https://github.com/elastic/logstash/issues/6820

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+--------------------------------------------------------------------------------
+input {
+  file {
+    path => "/path/to/myfile.json"
+    codec =>"json"
+}
+--------------------------------------------------------------------------------
+
+
+//ALVIN: SHOULD WE INCLUDE JSON_LINES HERE, TOO, OR MAYBE MENTION IT?
```
dedemorton: @jordansissel So if someone uses a community plugin that's not smart enough to do this magic...then it just blows up? That seems...not so good. I wonder why json_lines isn't just part of the json codec?

If the json codec works most of the time, then I vote for not covering json_lines.

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+
+<<plugins-codecs-protobuf,protobuf codec>>::
+
+Reads protobuf encoded messages and converts them to Logstash events. Requires
+the protobuf definitions to be compiled as Ruby files.
```
dedemorton: @jordansissel We cover that in the docs for the protobuf codec. Not sure it makes sense to go into that level of detail in this particular topic. Although..I gotta say...I sure like the idea of recommending something called "beefcake" in the doc.s :-P

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+--------------------------------------------------------------------------------
+
+
+//REVIEWERS: PLEASE DOUBLE-CHECK THE SYNTAX FOR THE PROTOBUF EXAMPLE. I CHANGED THE FORMATTING TO BE CONSISTENT WITH THE REST OF THE DOCUMENTATION HERE, BUT DIDN'T TEST THIS EXAMPLE.
+
+<<plugins-filters-xml,xml filter>>::
+
+Parses XML into fields.
++
+The following config parses the whole XML document stored in the `message` field:
```
dedemorton: I'd like to get @acchen97 to weigh in on this one. I've tried to strike a balance between providing sufficient info, but not not too much info because we want users to have something they can read quickly. For some of the examples, I started adding the input message and result because I thought it was hard to understand the plugin without seeing that (especially for stuff like grok). In this case, I think it's worthwhile. But maybe Alvin should see if there are other places where input/output examples would be useful.

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+The plugins described in this section are useful for extracting fields and
+parsing unstructured data into fields.
+
+Also see <<filter-plugins>> for the full list of available filter plugins. 
+
+<<plugins-filters-dissect,dissect filter>>::
+
+Extracts unstructured event data into fields by using delimiters. The dissect
+filter does not use regular expressions and is very fast. However, if the
+structure of the data varies from line to line, the grok filter is more
```
dedemorton: @jordansissel If you believe the doc. :-) What do you think?

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+* <<field-extraction>>
+* <<lookup-enrichment>>
+
+[[core-operations]]
+=== Performing Core Operations
+
+The plugins described in this section are useful for core operations, such as
+aggregating, mutating, and dropping events.
+
+Also see <<filter-plugins>> for the full list of available filter plugins.
```
acchen97: Instead of having this note in each section, should we just blanket it in the "Transforming Data" intro?

dedemorton: Sure we can do that...the repetition is for users who jump around and don't read from the beginning to end. I don't have a strong preference.

docs/static/transforming-data.asciidoc
acchen97: +1 on the using ordering in that issue. I agree that aggregate shouldn't be the first item.

jordansissel: I'm in favor of not mentioning the aggregate filter here at all. We are working on finding a better solution that could replace the aggregate filter; no details at this time though.

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+
+Parses dates from fields to use as Logstash timestamps for events.
++
+The following config parses a field called `logdate` to set the Logstash
+timestamp:
++
+[source,json]
+--------------------------------------------------------------------------------
+filter {
+  date {
```
acchen97: I feel this one is relatively self-explanatory and fine as-is. If they want more details, they can drill into the respective plugin page.

docs/static/transforming-data.asciidoc
jordansissel: @dedemorton I think the only two plugins that need "smarts" are the tcp input and stdin input. Any other input is likely never going to use the json_lines codec (and would use `json` instead where json is needed)

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+Drops events. This filter is typically used in combination with conditionals.
++
+The following config drops 40% of the `debug` level log messages:
++
+[source,json]
+--------------------------------------------------------------------------------
+filter {
+  if [loglevel] == "debug" {
+    drop {
+      percentage => 40
```
acchen97: In this context, I think just showing the most popular case (#1) is sufficient.

docs/static/transforming-data.asciidoc
jordansissel: > For some of the examples, I started adding the input message and result because I thought it was hard to understand 

I agree with your reasoning. For XML I agree the input may not need to be shown.

docs/static/transforming-data.asciidoc
jordansissel: haha. I think it's true. I had some misunderstandings about dissect's behavior last week and I think I understand it more now.

jordansissel:
Did another round commenting.

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+Performs general mutations on fields. You can rename, remove, replace, and
+modify fields in your events.
++
+The following config renames the `HOSTORIP` field to `client_ip`:
++
+[source,json]
+--------------------------------------------------------------------------------
+filter {
+  mutate {
+    rename => { "HOSTORIP" => "client_ip" }
```
acchen97: I'd probably add a few other usages into the mutate block here as they're relatively simple. Perhaps add 'strip' and 'remove_field' examples?

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+--------------------------------------------------------------------------------
+
+
+[[data-deserialization]]
+=== Deserializing Data
+
+The plugins described in this section are useful for deserializing data into
+Logstash events.
+
+Also see <<filter-plugins>> and <<codec-plugins>> for the full list of available
```
acchen97: Same for this comment as the one above. Maybe just note the full list of filters and codecs are available.

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+input {
+  tcp {
+    codec => fluent
+    port => 4000
+  }
+}
+--------------------------------------------------------------------------------
+
+
+//ALVIN: SHOULD WE ALSO COVER THE MSGPACK CODEC? I LOOKED AT THE DOCS FOR MSGPACK AND THERE'S NO USEFUL INFO IN THERE.
```
acchen97: I'm ok not including it.

docs/static/transforming-data.asciidoc
acchen97: In that case, I'm +1 on not including json_lines.

@dedemorton it sounds like it doesn't blow up, it'll handle the conversion to json codec under the covers if I'm understanding correctly.

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+  topic_id => "your_topic_goes_here"
+  codec => protobuf {
+    class_name => "Animal::Unicorn"
+    include_path => ['/path/to/protobuf/definitions/UnicornProtobuf.pb.rb']
+  }
+}
+--------------------------------------------------------------------------------
+
+
+//REVIEWERS: PLEASE DOUBLE-CHECK THE SYNTAX FOR THE PROTOBUF EXAMPLE. I CHANGED THE FORMATTING TO BE CONSISTENT WITH THE REST OF THE DOCUMENTATION HERE, BUT DIDN'T TEST THIS EXAMPLE.
```
acchen97: @IngaFeick if you don't mind, can you help confirm the Protobuf codec syntax here? We're looking to highlight it better in our docs.

docs/static/transforming-data.asciidoc
acchen97: I think its fine as-is, this is a feature discovery page.

docs/static/transforming-data.asciidoc
acchen97: ++ I like the example and the sample input/output is valuable esp in the field extraction context.

docs/static/transforming-data.asciidoc
acchen97: Pinging @gingerwizard who has experience using this filter. Perhaps he has some ideas for a simple example here.

docs/static/transforming-data.asciidoc
acchen97: Yes, it's in progress and getting pretty close. We can add it later after it gets in formally.

docs/static/transforming-data.asciidoc
acchen97: I'm ok with omitting it and its more of an advanced rather than core operation. There's already a myriad of good content here.

docs/static/transforming-data.asciidoc
IngaFeick: @jordansissel I'm using the ruby-protoc by codekitchen. I put that in the README but I guess it makes sense to add it here aswell.

docs/static/transforming-data.asciidoc
IngaFeick: @acchen97 lgtm

docs/static/transforming-data.asciidoc
dedemorton: done

docs/static/transforming-data.asciidoc
dedemorton: done - Removed aggregate. I'm out of time for reordering the content. If we think that's necessary even after removing aggregate, I can do it in a separate PR later.

docs/static/transforming-data.asciidoc
dedemorton: agreed...leaving as-is

docs/static/transforming-data.asciidoc
dedemorton: fixed to show only the first use case.

docs/static/transforming-data.asciidoc
dedemorton: fixed - I've added strip, but not remove_field because that's actually one of the common filter options (I know...it's hard to tell...good thing we're fixing that)

docs/static/transforming-data.asciidoc
dedemorton: done

docs/static/transforming-data.asciidoc
dedemorton: left as-is

docs/static/transforming-data.asciidoc
dedemorton: left as-is

docs/static/transforming-data.asciidoc
dedemorton: added link to ruby-protoc compiler (people will need to find beefcake on their own...I can't be implicated in that).

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,520 @@
...
+<<plugins-codecs-protobuf,protobuf codec>>::
+
+Reads protobuf encoded messages and converts them to Logstash events. Requires
+the protobuf definitions to be compiled as Ruby files.
++
+The following config decodes events from a Kafka stream:
++
+[source,json]
+--------------------------------------------------------------------------------
+kafka {
```
dedemorton: fixed

docs/static/transforming-data.asciidoc
dedemorton: left as-is

docs/static/transforming-data.asciidoc
dedemorton: left as-is

docs/static/transforming-data.asciidoc
dedemorton: left as-is

docs/static/transforming-data.asciidoc
dedemorton: added

docs/static/transforming-data.asciidoc
dedemorton: @gingerwizard @jordansissel I'm hoping to wrap this content up before I go on vacation. Does either of you have a suggested for an example that is a bit less complex than what we show in the doc for the ES filter?

docs/static/transforming-data.asciidoc
dedemorton: fixed

docs/static/transforming-data.asciidoc
dedemorton: minor issue, so leaving as-is. I can update in a separate PR when guy responds.

docs/static/transforming-data.asciidoc
dedemorton: leaving as-is. Just let me know when it's ready to be added to the list.

docs/static/transforming-data.asciidoc
```diff
@@ -0,0 +1,532 @@
...
+            hosts => ["es-server"]
+            query => "type:start AND operation:%{[opid]}"
+            fields => { "@timestamp" => "started" }
+         }
+         date {
+            match => ["[started]", "ISO8601"]
+            target => "[started]"
+         }
+         ruby {
+            code => "event['duration_hrs'] = (event['@timestamp'] - event['started']) / 3600 rescue nil"
```
dedemorton: With the new event API syntax, I think this should actually be:

```
code => "event.set("[duration_hrs]") = (event.get("[@timestamp]") - event.get("[started]")) / 3600 rescue nil"
```

suyograo: `code => "event.set("duration_hrs", (event.get("@timestamp") - event.get("started")) / 3600) rescue nil"`

suyograo: @dedemorton use this syntax ^^

jordansissel:
🤘

acchen97:
@dedemorton this is really fantastic work! I just checked it out live and users will definitely feel an impact on the getting started experience.

PR #6996, opened 2017-04-28
update reload metrics within pipelines_mutex

jsvd:
if during an agent reload a shutdown operation happens, it's possible that an external observer requests metrics between [converge_state and update_metrics](https://github.com/elastic/logstash/blob/163d27d78f6dae20f5f409dce659f24401552512/logstash-core/lib/logstash/agent.rb#L152) and shutdown has already caused metric changes. This results in an inconsistent view, which is being triggered ocasionally in our spec suite.

Placing a `sleep(1)` on that line causes about 9-10 specs to fail, most in `agent_spec` and one in `converge_spec`. This PR fixes this issue.

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -131,6 +131,16 @@
 #
 # queue.checkpoint.interval: 1000
 #
+# ------------ Dead-Letter Queue Settings --------------
+# Flag to turn on dead-letter queue.
+#
+# dead_letter_queue.enable: false
+#
+# If using dead_letter_queue.enable: true, the directory path where the data files will be stored.
+# Default is path.data/dead_letter_queue
+#
+# path.dead_letter_queue:
+#
 # ------------ Metrics Settings --------------
 #
 # Bind address for the metrics REST endpoint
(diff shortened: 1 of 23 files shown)


logstash-core/lib/logstash/agent.rb
```diff
@@ -148,10 +148,10 @@ def converge_state_and_update
     @pipelines_mutex.synchronize do
       pipeline_actions = resolve_actions(results.response)
       converge_result = converge_state(pipeline_actions)
+      update_metrics(converge_result)
     end
 
     report_currently_running_pipelines(converge_result)
```
ph: Can you move `#report_currently_running_pipelines` and `dispatch_events` in the mutex? This will make sure that theses too methods execution are consistent with the current state of the pipeline?

ph:
Because of this fix do we want to close or merge that https://github.com/elastic/logstash/pull/6956 ?

logstash-core/lib/logstash/agent.rb
jsvd: interestingly, moving those two lines into the mutex caused more tests to fail, I need to investigate why.

jsvd:
@ph I still think it's worth merging that PR as it removes a lot of stubbing and other let magic.

jsvd:
@ph I haven't had much time to further investigate why your suggestion causes more failures, what do you think of merging as is since it removes the issue of metrics being out of sync with the pipeline?

jsvd:
I closed this since I was seeing failures even with this adjustment. I created #7066 which is much more stable

Issue #6625, opened 2017-02-01
queue.max_bytes isn't respected at start up time


jsvd:
During the startup of logstash the head page is beheaded if it contains unacked events and a new page is created.

This page creation happens even if `queue.max_bytes` has been reached. In the worst case, it's possible to fill up the disk even if queue.max_bytes is 5mb.

Steps to reproduce:

logstash.yml:

```yml
queue.type: persisted
queue.page_capacity: 5mb
queue.max_bytes: 10mb
```

Use a high throughput pipeline that stalls:

`bin/logstash -e "input { generator {} } filter {sleep { time => 1 } }" -w 1 -b 1`

Let logstash start and kill it. Each time this happens a new page is created:

1.

```
/tmp/logstash-5.1.2 % du -sh data/**/*
 10M	data/queue
4.0K	data/queue/checkpoint.0
4.0K	data/queue/checkpoint.head
5.0M	data/queue/page.0
5.0M	data/queue/page.1
4.0K	data/uuid
```

2.

```
/tmp/logstash-5.1.2 % du -sh data/**/*
 15M	data/queue
4.0K	data/queue/checkpoint.0
4.0K	data/queue/checkpoint.1
4.0K	data/queue/checkpoint.head
5.0M	data/queue/page.0
5.0M	data/queue/page.1
5.0M	data/queue/page.2
4.0K	data/uuid
```

3.

```
/tmp/logstash-5.1.2 % du -sh data/**/*
 20M	data/queue
4.0K	data/queue/checkpoint.0
4.0K	data/queue/checkpoint.1
4.0K	data/queue/checkpoint.2
4.0K	data/queue/checkpoint.head
5.0M	data/queue/page.0
5.0M	data/queue/page.1
5.0M	data/queue/page.2
5.0M	data/queue/page.3
4.0K	data/uuid
```

etc..

colinsurprenant:
interesting. not sure how we should deal with that actually. 
also note that these are sparse files. on MacOS they `du` will reports the same size as `ls` but on Linux `du` will show that they are mostly unallocated on disk. see this comment https://github.com/elastic/logstash/pull/6518#issuecomment-276451360

jsvd:
I guess the two questions are:
1. even if the actual file size is close to 0B and the disk is "full" of them, will other applications see the disk as full? Can this still be a problem?
2. Is there a way to know the real size on macosx and any other OS where this happens?

colinsurprenant:
[...]

Issue #6627, opened 2017-02-01
queue.max_bytes isn't compared against queue.page_capacity


jsvd:
setup:

```
queue.type: persisted
queue.page_capacity: 250mb
queue.max_bytes: 10mb
```
with pipeline: `input { generator {} } filter {sleep { time => 1 } }` and `-w 1 -b 1`

Starting logstash will cause it to print 1 event and stop:

```
/tmp/logstash-5.1.2 % bin/logstash -e "input { generator {} } filter {sleep { time => 1 } }" -w 1 -b 1
{
      "sequence" => 0,
    "@timestamp" => 2017-02-01T21:14:26.883Z,
      "@version" => "1",
          "host" => "Joaos-MBP-5.lan",
       "message" => "Hello world!"
}
# pipeline stalls
```

Data created:

```
/tmp/logstash-5.1.2 % du -sh data/**/*
250M	data/queue
4.0K	data/queue/checkpoint.head
250M	data/queue/page.0
4.0K	data/uuid
```

colinsurprenant:
good catch.

jsvd:
this can be another bootstrap check done in the context of https://github.com/elastic/logstash/issues/6552

colinsurprenant:
agree, will be fixed by #6998

colinsurprenant:
fixed by #8978

PR #6998, opened 2017-04-29
Check Available Disk Space for Persisted Queues, fixes #6552 

### LINKED ISSUE #6552: Add bootstrap checks to persistent queue
When persistent queue is enabled, add the following bootstrap checks when starting LS

- [ ] validate if there is enough disk space to allocate at least page-size worth of data.
- [ ] ensure `queue.max_bytes` is available on disk.

 If these are not satisfied, Logstash should fail to start up.

original-brownbear:
For #6552:

* Implemented free disk space check in Java since Ruby doesn't have this out of the box apparently
* Added JUnit test for the Java file system logic
* Added new bootstrap check
  * checks against the max of (`queue.max_bytes` and `queue.page_capacity`)

### DIFF
--- a/logstash-core/lib/logstash/bootstrap_check/persisted_queue_config.rb
+++ b/logstash-core/lib/logstash/bootstrap_check/persisted_queue_config.rb
@@ -1,0 +1,14 @@
+# encoding: utf-8
+require 'logstash/errors'
+
+module LogStash module BootstrapCheck
+    class PersistedQueueConfig
+      def self.check(settings)
+        return unless settings.get('queue.type') == 'persisted'
+        if settings.get('queue.page_capacity') > settings.get('queue.max_bytes')
+          raise LogStash::BootstrapCheckError,
+                'Invalid configuration, `queue.page_capacity` must be smaller or equal to `queue.max_bytes`'
+        end
+      end
+    end
+end end
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/FsUtil.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/FsUtil.java
@@ -1,0 +1,47 @@
+package org.logstash.ackedqueue;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * File System Utility Methods.
+ */
+public final class FsUtil {
+
+    private FsUtil() {
+    }
+
+    /**
+     * Checks if the request number of bytes of free disk space are available under the given
+     * path.
+     * @param path Directory to check
+     * @param size Bytes of free space requested
+     * @return True iff the
+     * @throws IOException on failure to determine free space for given path's partition
+     */
+    public static boolean hasFreeSpace(final String path, final long size) throws IOException {
+        final File[] partitions = File.listRoots();
+        File location = new File(path).getCanonicalFile();
+        boolean found = false;
+        while (!found) {
+            for (final File partition : partitions) {
+                if (partition.equals(location)) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                location = location.getParentFile();
+                if (location == null) {
+                    throw new IllegalStateException(
+                        String.format(
+                            "Unable to determine the partition that contains '%s'.", path
+                        )
+                    );
+                }
+            }
+        }
+        return location.getFreeSpace() >= size;
+    }
+
+}
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -20,9 +20,10 @@ require "logstash/patches/clamp"
 require "logstash/settings"
 require "logstash/version"
 require "logstash/plugins/registry"
-require "logstash/bootstrap_check/default_config"
 require "logstash/bootstrap_check/bad_java"
 require "logstash/bootstrap_check/bad_ruby"
+require "logstash/bootstrap_check/default_config"
+require "logstash/bootstrap_check/persisted_queue_config"
 require "set"
 
 java_import 'org.logstash.FileLockFactory'
@@ -52,7 +53,8 @@ class LogStash::Runner < Clamp::StrictCommand
   DEFAULT_BOOTSTRAP_CHECKS = [
       LogStash::BootstrapCheck::BadRuby,
       LogStash::BootstrapCheck::BadJava,
-      LogStash::BootstrapCheck::DefaultConfig
+      LogStash::BootstrapCheck::DefaultConfig,
+      LogStash::BootstrapCheck::PersistedQueueConfig
   ]
 
   # Node Settings
--- a/logstash-core/spec/logstash/bootstrap_check/persisted_queue_config_spec.rb
+++ b/logstash-core/spec/logstash/bootstrap_check/persisted_queue_config_spec.rb
@@ -1,0 +1,23 @@
+require "spec_helper"
+require "tmpdir"
+require "logstash/bootstrap_check/persisted_queue_config"
+
+describe LogStash::BootstrapCheck::PersistedQueueConfig do
+
+  context("when persisted queues are enabled") do
+    let(:settings) do
+      settings = LogStash::SETTINGS.dup
+      settings.set_value("queue.type", "persisted")
+      settings.set_value("queue.page_capacity", 1024)
+      settings.set_value("path.queue", ::File.join(Dir.tmpdir, "some/path"))
+      settings
+    end
+
+    context("and 'queue.max_bytes' is set to a value less than the value of 'queue.page_capacity'") do
+      it "should throw" do
+        settings.set_value("queue.max_bytes", 512)
+        expect { LogStash::BootstrapCheck::PersistedQueueConfig.check(settings) }.to raise_error
+      end
+    end
+  end
+end
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -24,7 +24,6 @@ import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
-
 // TODO: Notes
 //
 // - time-based fsync
@@ -181,12 +180,13 @@ public class Queue implements Closeable {
 
                 newCheckpointedHeadpage(headPageNum);
                 this.closed.set(false);
+                this.ensureDiskAvailable(this.maxBytes);
 
                 return;
             }
 
+            long diskNeeded = this.maxBytes;
             // at this point we have a head checkpoint to figure queue recovery
-
             // reconstruct all tail pages state upto but excluding the head page
             for (int pageNum = headCheckpoint.getFirstUnackedPageNum(); pageNum < headCheckpoint.getPageNum(); pageNum++) {
 
@@ -197,7 +197,7 @@ public class Queue implements Closeable {
 
                 PageIO pageIO = this.pageIOFactory.build(pageNum, this.pageCapacity, this.dirPath);
                 pageIO.open(cp.getMinSeqNum(), cp.getElementCount());
-
+                diskNeeded -= pageIO.getHead();
                 add(cp, new TailPage(cp, this, pageIO));
             }
 
@@ -208,6 +208,7 @@ public class Queue implements Closeable {
 
             PageIO pageIO = this.pageIOFactory.build(headCheckpoint.getPageNum(), this.pageCapacity, this.dirPath);
             pageIO.recover(); // optimistically recovers the head page data file and set minSeqNum and elementCount to the actual read/recovered data
+            ensureDiskAvailable(diskNeeded - (long) pageIO.getHead());
 
             if (pageIO.getMinSeqNum() != headCheckpoint.getMinSeqNum() || pageIO.getElementCount() != headCheckpoint.getElementCount()) {
                 // the recovered page IO shows different minSeqNum or elementCount than the checkpoint, use the page IO attributes
@@ -256,6 +257,14 @@ public class Queue implements Closeable {
             throw new LockException("The queue failed to obtain exclusive access, cause: " + e.getMessage());
         } finally {
             lock.unlock();
+        }
+    }
+
+    private void ensureDiskAvailable(final long diskNeeded) throws IOException {
+        if (!FsUtil.hasFreeSpace(this.dirPath, diskNeeded)) {
+            throw new IOException(
+                "Not enough free disk space available to allocate persisted queue."
+            );
         }
     }
 
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
@@ -44,6 +44,15 @@ public abstract class AbstractByteBufferPageIO implements PageIO {
     private CRC32 checkSummer;
     private final List<Integer> offsetMap; // has to be extendable
 
+    // we don't have different versions yet so simply check if the version is VERSION_ONE for basic integrity check
+    // and if an unexpected version byte is read throw PageIOInvalidVersionException
+    public static void validateVersion(byte version) throws PageIOInvalidVersionException {
+        if (version != VERSION_ONE) {
+            throw new PageIOInvalidVersionException(String
+                .format("Expected page version=%d but found version=%d", VERSION_ONE, version));
+        }
+    }
+
     public AbstractByteBufferPageIO(int pageNum, int capacity) {
         this.minSeqNum = 0;
         this.elementCount = 0;
@@ -115,15 +124,6 @@ public abstract class AbstractByteBufferPageIO implements PageIO {
         // if we were not able to read any element just reset minSeqNum to zero
         if (this.elementCount <= 0) {
             this.minSeqNum = 0;
-        }
-    }
-
-    // we don't have different versions yet so simply check if the version is VERSION_ONE for basic integrity check
-    // and if an unexpected version byte is read throw PageIOInvalidVersionException
-    private static void validateVersion(byte version) throws PageIOInvalidVersionException {
-        if (version != VERSION_ONE) {
-            throw new PageIOInvalidVersionException(String
-                .format("Expected page version=%d but found version=%d", VERSION_ONE, version));
         }
     }
 
--- a/logstash-core/src/test/java/org/logstash/ackedqueue/FsUtilTest.java
+++ b/logstash-core/src/test/java/org/logstash/ackedqueue/FsUtilTest.java
@@ -1,0 +1,40 @@
+package org.logstash.ackedqueue;
+
+import org.hamcrest.CoreMatchers;
+import org.hamcrest.MatcherAssert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/**
+ * Tests for {@link FsUtil}.
+ */
+public final class FsUtilTest {
+
+    @Rule
+    public final TemporaryFolder temp = new TemporaryFolder();
+
+    /**
+     * {@link FsUtil#hasFreeSpace(String, long)} should return true when asked for 1kb of free
+     * space in a subfolder of the system's TEMP location.
+     */
+    @Test
+    public void trueIfEnoughSpace() throws Exception {
+        MatcherAssert.assertThat(
+            FsUtil.hasFreeSpace(temp.newFolder().getAbsolutePath(), 1024L),
+            CoreMatchers.is(true)
+        );
+    }
+
+    /**
+     * {@link FsUtil#hasFreeSpace(String, long)} should return false when asked for
+     * {@link Long#MAX_VALUE} of free space in a subfolder of the system's TEMP location.
+     */
+    @Test
+    public void falseIfNotEnoughSpace() throws Exception {
+        MatcherAssert.assertThat(
+            FsUtil.hasFreeSpace(temp.newFolder().getAbsolutePath(), Long.MAX_VALUE),
+            CoreMatchers.is(false)
+        );
+    }
+}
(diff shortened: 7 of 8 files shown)


logstash-core/lib/logstash/bootstrap_check/persisted_queue_config.rb
```diff
@@ -0,0 +1,22 @@
+# encoding: utf-8
+require "java"
+require "logstash/errors"
+
+module LogStash module BootstrapCheck
+    class PersistedQueueConfig
+      def self.check(settings)
+        if settings.get("queue.type") == "persisted"
```
colinsurprenant: this is really a styling preference but I personally like to do a quick `return` to save an indentation. I find it ok to do such a "short-circuit" when it is the first(s) statement. For example:
```ruby
def self.check(settings)
  return unless settings.get("queue.type") == "persisted"
  ...
end
```

logstash-core/lib/logstash/bootstrap_check/persisted_queue_config.rb
```diff
@@ -0,0 +1,22 @@
...
+          if settings.get("queue.page_capacity") > settings.get("queue.max_bytes")
+            raise LogStash::BootstrapCheckError,
+                  "Invalid configuration, `queue.page_capacity` must be smaller or equal to `queue.max_bytes`"
+          end
+          unless org.logstash.ackedqueue.FsUtil.has_free_space(
+            settings.get("path.queue"), settings.get("queue.max_bytes")
+          )
+            raise LogStash::BootstrapCheckError,
+                  "Not enough disk space to allocate persisted queue page"
```
colinsurprenant: actually, at this point there is not enough disk space to allocate a full queue at `queue.max_bytes` not just a queue page.

colinsurprenant:
After starting to review I realized we need to discuss these corner cases:

- @suyograo original idea in #6552 was to «validate if there is enough disk space to allocate at least page-size worth of data.» and that  «ensure queue.max_bytes is available on disk». Now obviously, if we make sure that `queue.page_capacity` <= `queue.max_bytes` then only checking for `queue.max_bytes` free space is good enough. This is what you did in the PR and this is all good, at first sight ;)
- But we are not taking into account the actual queue size on disk. This is a case where logstash starts with some previously existing queue data, so if for example `max_bytes` is 2MB and the current queue size is 1MB, only `max_bytes` minus the current queue size needs to be available on disk, right? (related issue #6518)
- Also, I am wondering if we should enforce having `max_bytes` be a multiple of `page_capacity`? it seems it makes reasoning about this more easy?
- If we  enforce having `max_bytes` be a multiple of `page_capacity` it will be simpler to deal with pre-existing queue data and the fact that at least `page_capacity` needs to be available on disk for the queue head page beheading on `queue.open()`. Having enough space for `max_bytes` minus the current queue size will always return a multiple of `page_capacity`.

original-brownbear:
@colinsurprenant 

> But we are not taking into account the actual queue size on disk. This is a case where logstash starts with some previously existing queue data, so if for example max_bytes is 2MB and the current queue size is 1MB, only max_bytes minus the current queue size needs to be available on disk, right? 

+1 otherwise you could run into a really evil situation where you need to free disk space to be able to start logstash, but only after a dirty shutdown.

> Also, I am wondering if we should enforce having max_bytes be a multiple of page_capacity? it seems it makes reasoning about this more easy?

Maybe better to allow setting the max number of pages in that scenario? Otherwise, you'd need a calculator to set that setting for real life page sizes. Seems annoying to users ? :D

jsvd:
from @colinsurprenant 
> Also, I am wondering if we should enforce having max_bytes be a multiple of page_capacity? it seems it makes reasoning about this more easy?

from @original-brownbear 
> Maybe better to allow setting the max number of pages in that scenario? Otherwise, you'd need a calculator to set that setting for real life page sizes. Seems annoying to users ? :D

From a user's perspective they'll want to set max_bytes to a certain amount and that's it. the other two variables - number of pages and page capacity - IMO are "advanced settings" that most users won't care about, and I myself don't know how to advise tuning them. 

I don't think there's a need to ensure that if you set max_bytes to 1024mb and pages to 1mb you're guaranteed to have at most exactly 1024 pages. We can warn/document that max_bytes is a ballpark number that should never be crossed, but it's possible to mark the queue full at 983mb, for example.

Also it may be useful to shout a warning when the page_capacity/max_bytes ratio is close to 1. Having page capacity 500mb and max_bytes 1gb seems wrong, IMO. A saner option should look like page_capacity 100 mb and max_bytes 1 gb, WDYT?

> the fact that at least page_capacity needs to be available on disk for the queue head page beheading on queue.open()

+1 on always checking free disk space before allocating a new page.

colinsurprenant:
> +1 on always checking free disk space before allocating a new page.

I don't think it is a good idea. Page allocation/churn can be really fast and doing a filesystem free space check before all page allocation could be too costly. 

Also, checking for `page_capacity` free space on every page creation does not guarantees that this free space will be preserved throughout the whole page write cycle (also true for initial free space bootstrap check) so in that respect I prefer we make sure to minimize potential corruption in the write method by reviewing the places where a disk full IO exception could happen and handle it is a way to try and preserve the queue state.  I think a one-time bootstrap check is good enough here, and eventually, allowing periodic % disk usage check will be even more useful. 

Also, 
- Since a page is probably sparsely allocated (see https://github.com/elastic/logstash/pull/6518#issuecomment-276451360)  and a page will probably never be filled to capacity which results in variable page sizes, up to a limit of `page_capacity`, so in that respect having `max_bytes` be a multiple of `page_capacity` does not mean much in reality. I take back that idea :P
- Agree that having a max number of pages setting is not really useful here ...

original-brownbear:
@jsvd @colinsurprenant so how does this sound? :)

* Check how much space is already in use by pages and discount necessary free space accordingly
* Check that `page_capacity < max_bytes`
* Warn if `page_size / max_bytes` > 0.5 ?

suyograo:
> But we are not taking into account the actual queue size on disk. This is a case where logstash starts with some previously existing queue data, so if for example max_bytes is 2MB and the current queue size is 1MB, only max_bytes minus the current queue size needs to be available on disk, right?

+1 on taking the existing queue size into account.

Also, I agree with @jsvd that user should not be concerned with page size and max_bytes relationship when setting defaults. Looks like we don't need it from @colinsurprenant latest comment.

So, +1 on what @original-brownbear suggested as bootstraps checks (not runtime). Now, for getting the size of queue on disk already, it should be good enough to use "logical bytes used" as @jordansissel suggested in https://github.com/elastic/logstash/pull/6518#issuecomment-297861354. Is calculating that size on startup (assuming a large queue) slow down things?

original-brownbear:
@suyograo about this:

> Now, for getting the size of queue on disk already, it should be good enough to use "logical bytes used" as @jordansissel suggested in #6518 (comment). Is calculating that size on startup (assuming a large queue) slow down things?

Isn't the unit of actual disk space used here always a multiple of the page size? 
I think looking into the actual bytes used in the allocated mapped buffer is not meaningful, we have to use what `du` would show and not the logical bytes used since this is also what will bound the available size we can allocate to new pages isn't it?
So can't we do this super fast by simply checking how many page files exist and multiplying that with the page size to get the used disk space?

original-brownbear:
@suyograo about this:

> Now, for getting the size of queue on disk already, it should be good enough to use "logical bytes used" as @jordansissel suggested in #6518 (comment). Is calculating that size on startup (assuming a large queue) slow down things?

Isn't the unit of actual disk space used here always a multiple of the page size? 
I think looking into the actual bytes used in the allocated mapped buffer is not meaningful, we have to use what `du` would show and not the logical bytes used since this is also what will bound the available size we can allocate to new pages isn't it?
So can't we do this super fast by simply checking how many page files exist and multiplying that with the page size to get the used disk space?

colinsurprenant:
@original-brownbear this is precisely the discussion in #6518.
Using `page_capacity` multiplied by the number of pages will provide a bigger size than what is really used. On Linux a new page file will show a `page_capacity` size in `ls` but a zero size in `du`. This is because it is sparsely allocated. So using the "logical bytes used" as mentioned by @suyograo and actually proposed in #6518 it will provide a more precise queue size for the actual bytes used on disk. 

So here's what I suggest: since #6518 is not completed/merged, in the first phase let's use the existing  `Queue.currentByteSize` (which is `page_capacity` multiplied by the number of pages) and when we have `Queue.getCurrentPhysicallyPersistedByteSize` as proposed in #6518 then we can use that for a more precise size measurement.

So, agree for 
> - Check how much space is already in use by pages and discount necessary free space accordingly
> - Check that page_capacity < max_bytes

But I don't think we need
> Warn if page_size / max_bytes > 0.5 ?

suyograo:
@colinsurprenant should we first complete and merge #6518 then? Maybe @original-brownbear can take over #6518 and then we can use the right metric here? WDYT?

colinsurprenant:
@original-brownbear @suyograo thinking about this, what we need to do is check that there is free space for the max between [(`max_bytes` - *current queue size*), `page_capacity`] so that if a beheading operation is triggered at queue open, even though the new head page will be sparsely allocated, it will be safer and will work regardless of the filesystem type . (related to #6625 / @jsvd )

colinsurprenant:
for reference, this will also fix #6625, #6627

original-brownbear:
Guys, I looked into this some more and I'm not so convinced that the return of `org.logstash.ackedqueue.Queue#getPersistedByteSize` is what we what to use here.

The number returned by that function is less (significantly less, ~40% ) than the actual size of the queue's size on disk. I added this test: https://github.com/elastic/logstash/pull/6998/files#diff-dc6da7f64cc5e8f5dc313fe850e9609fR48 to demonstrate that fact.
But shouldn't we at least for the startup check just use the size of all files in the queue directory? That said, isn't this an indication that `org.logstash.ackedqueue.Queue#getPersistedByteSize` returns the wrong result too (or is it not supposed to return size on disk?)?

colinsurprenant:
@original-brownbear ok, so I feel we are going full circle here. my take is that this is still related to the same problem we see with different sizes between `ls` and `du`, see https://github.com/elastic/logstash/pull/6518#issuecomment-276451360.

Our `Queue.getPersistedByteSize()` computes the actual written bytes into all pages but a page is sparsely allocated to its maximum capacity, this is why doing `ls` in the data directory will show all pages of same size regardless if less bytes were actually written to them but on Linux, `du` will show the actual physical disk bytes used. 

Looking at the code, `FSUtils.getPersistedSize()` uses [`Files.size()`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#size(java.nio.file.Path)) and in the description:
> Returns the size of a file (in bytes). The size may differ from the actual size on the file system due to compression, support for sparse files, or other reasons. The size of files that are not regular files is implementation specific and therefore unspecified.

So this is again the problem here and it is confusing.

original-brownbear:
@colinsurprenant ok all makes sense now. So I'll just make sure that I discount by what we read from the headers via `Queue.getPersistedByteSize()` plus the free space on the disk is at least the max queue size :) On it :)

original-brownbear:
@colinsurprenant done I think. Addressed all your points and created a utility method that correctly (at least in agreement with the `Queue` methods) determines the queue size.
(no tests for the preexisting queue size though, this would be crazy hard to make portable I fear unless we just wrap the JVM free disk space call and use mocking, but not sure if that has any point to it really ...)

colinsurprenant:
@original-brownbear two things I'd like us to discuss to improve:
- duplicate queue/page open/read logic in the `FSUtil.fileSize()` method
- full queue will be completely read twice on startup, first for bootstrap check and then on queue open. 

Let's see how we can improve that. A few ideas:
- we should be able to refactor code to avoir open logic duplication.
- or maybe we should see into being able to instantiate/open a queue in the bootstap and propagate it to the pipeline initialization? 
- but this actually raises the question of the eventual multi-pipeline (and thus multi queue) support, how will this be possible to check in the bootstrap?
- maybe this check simply does not belong in the bootstrap check and it should be done at pipeline initialization? 
- again, thinking forward, how will this play with multiple pipelines?

original-brownbear:
@colinsurprenant @suyograo ok reverted all now unnecessary changes (I think the moving of a the few things I made `static` because they had no instance ref. can stick around just for correctness reasons, rest reverted) and added the check to `Queue.open`, throwing `IOException` on failure.

Is this going in the direction we want?

colinsurprenant:
@original-brownbear any reason why we didn't keep the previous `getPersistedByteSize()` method instead of the  `diskNeeded` strategy in open? It feels like using `getPersistedByteSize()` would have worked in this context and would have been a useful Queue method?

logstash-core/src/main/java/org/logstash/ackedqueue/FsUtil.java
```diff
@@ -0,0 +1,43 @@
...
+                );
+            }
+            for (final File partition : partitions) {
+                if (partition.equals(location)) {
+                    found = true;
+                } else {
+                    location = location.getParentFile();
+                }
+            }
+        }
```
colinsurprenant: I don't understand the logic here: 
- the`for` loop, loops through the partitions, if the location is the partition it sets `fount = true` and continue looping, else if it is not, it backtracks and mutates location  **and continue** looping in the remaining partitions.
- upon completing the `for` loop, if `!found` then it repeats the process.

What is confusing is the `location` mutation `location = location.getParentFile();` in the `for` loop. Wouldn't you want to back track the location after the `for` loop only when `!found` ?? something like this:
```java
for (final File partition : partitions) {
    if (partition.equals(location)) { found = true; }
}
if (! found) { location = location.getParentFile(); }
```

Or am I missing something?

original-brownbear: @colinsurprenant wow yea, sorry for this on must have been tired.
Obviously this is supposed to be what you pasted :) (+ a `break;` after `found=true;`I guess :)) Will fix asap.

original-brownbear:
@colinsurprenant my reasoning for doing this inline and not using that method was:
*  Would like to break out for the case of no pre-existing queue without side effects
* Wanted to not run the risk of side effects from all the logic below https://github.com/elastic/logstash/pull/6998/files#diff-69b2c41a7ba6596afa233505525212bbR216 in the `open`.

It just seemed cleaner to use the `head` values as soon as available to keep side effects out :)

logstash-core/src/main/java/org/logstash/ackedqueue/FsUtil.java
original-brownbear: @colinsurprenant Fixed :)

colinsurprenant: ok!! 
Right, but the `break` is not mandatory but, yeah.

logstash-core/src/main/java/org/logstash/ackedqueue/FsUtil.java
```diff
@@ -0,0 +1,45 @@
...
+                if (partition.equals(location)) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                location = location.getParentFile();
+            }
+        }
+        return location.getFreeSpace() >= size;
```
colinsurprenant: mainly style/readability comment:
since `location`  cannot be `null` upon entering the `while` loop, why not check for `if (location == null)` just after it is mutated? 

Also, I assume that `location.getParentFile()` returns `null` when trying to call it on the root `location`? 
I'd like to see these corner cases tested...

original-brownbear:
@colinsurprenant the corner case of the `location` becoming `null` is basically impossible to reproduce (at least on *nix). Adjusted the logic to throwing unchecked `IllegalStateException` if the parent resolves to `null` but I see no way of writing a test reproducing this situation (kind of looks like getting to that `IllegalStateException` would require a JVM bug)

logstash-core/src/main/java/org/logstash/ackedqueue/FsUtil.java
```diff
@@ -0,0 +1,47 @@
...
+            for (final File partition : partitions) {
+                if (partition.equals(location)) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                location = location.getParentFile();
+                if (location == null) {
+                    throw new IllegalStateException(
```
colinsurprenant: I am not sure we want to throw an unchecked `IllegalStateException` here, I think a proper `IOException` would be better since reaching a `null` value seems to be a *valid* state or behaviour, per the doc:

> Returns the abstract pathname of this abstract pathname's parent, or null if this pathname does not name a parent directory.

I agree it might be a hard to test case and maybe this is OS specific too but I don't think it warrants an unchecked `IllegalStateException`. WDYT?

original-brownbear: @colinsurprenant I think I explained this badly, the point is this:

If you input a valid directory and then go through its parents step by step and never arrive at the path of a partition, then you have some path that isn't part of any mounted partition.
At least on *nix this is impossible => `IllegalStateException` since I don't see how this could ever happen unless we have an actual error in the logic of this code => checked exception for a state that is assumed to be "impossible" seems wrong right?

colinsurprenant:
@original-brownbear ^^ replied in-line for the `null` `location`. 

About `getPersistedByteSize()` - I am not sure I understand how having a `getPersistedByteSize()` 
 would affect: «Would like to break out for the case of no pre-existing queue without side effects» ? I mean in the case of no pre-existing queue, `getPersistedByteSize()` is not/would not be called ...?

Not sure I understand the side-effect either - once all the pages and head page has been opened, we can go through the tail pages list and sum the head??

logstash-core/src/main/java/org/logstash/ackedqueue/FsUtil.java
colinsurprenant: I don't know how this would be possible and maybe this is a OS specific behaviour but per the documentation  it is an acceptable return value and in that respect, from our point of view, we can trap it and just throw our `IOException` to say that, as documented:
> @throws IOException on failure to determine free space for given path's partition

original-brownbear: @colinsurprenant I don't have a very strong opinion on this one, but my worry is essentially this situation:

* User uses some strange path (maybe some `NFS` style thing) that doesn't show up with a partition that the JVM understands and we throw a checked Exception that may get caught by upstream logic (or worse yet re-thrown in some opaque way) => Likely it will be easier to understand this in a support case if we have it be an unchecked one that just bubbles up directly (also philosophically it makes more sense to throw unchecked for situations that are deemed impossible)

... again no strong opinion on this though :) the below point seems more important :)

colinsurprenant: not sure I am following... 🤔  
User uses some strange path -> we throw `new IOException("unable to determine free space for that path")`.??  
And it should be caught exactly where it is supposed be caught, as close to the place where this check is done and **not** bubble up in a more difficult to diagnose context?! 

IMO unchecked exception should generally be avoided and used only for weird runtime problems, not for something that is a **documented** behaviour for which we can trap and deal with?

original-brownbear:
@colinsurprenant any update on this one ? :)

original-brownbear:
@colinsurprenant @suyograo rebased this btw, I guess we can continue here?

original-brownbear:
@suyograo ping, this one is still in limbo :)

colinsurprenant:
@original-brownbear what was the reason for closing this?  I am trying to make sense of all the issues related to disk space and queue size and not sure why this one was closed? (is there another fix somewhere else or did we discard this proposal?)

original-brownbear:
@colinsurprenant I simply gave up here eventually, the PR didn't get a reply ever since May and started conflicting again and again. I can rebase it and reopen if there is still interest? Imo it's still a sound solution.

colinsurprenant:
@original-brownbear oh ok. I don't think closing a potentially valid PR because it does not receive timely feedback is a good idea, it simply adds to the confusion. I would suggest you add a note to that effect so that it is made explicit if you decide to "just close".

colinsurprenant:
re-opened PR as I will be following up shortly on the whole disk space and queue size issues.

logstash-core/src/main/java/org/logstash/ackedqueue/FsUtil.java
colinsurprenant: I am still on the fence about this `IllegalStateException` vs `IOException` but will agree with what you think is best here.

colinsurprenant:
@original-brownbear in the context of #8936 I re-assessed this PR and I am good with moving forward. This will require a rebase. I will nonetheless leave a few comments now while at it and we can address these when you are ready.

colinsurprenant:
@original-brownbear actually, this should be good. Once you rebase I will test and if all good will ok it.

original-brownbear:
Hmm sorry, this one is beyond repair it now :( The logic here changed a lot (this PR was created ~8 months ago) and is about to change some more as a result #8958 it seems. Can't rebase this without significant logical changes to the approach => don't have the time for reworking this and going through another review right now => giving up here

colinsurprenant:
The bootstrap checks should be good, the `hasFreeSpace()` is fine and the changes in the `Queue` classes are rather simple. It really should not be that hard but we can also reboot this in a fresh PR without doing a git rebase. I can understand your frustration, just let me know if you are still not interested, I can do it too.

original-brownbear:
@colinsurprenant better if you do it I think.

colinsurprenant:
looking at this I just realized that the author is actually @wainersm !? not sure I understand how this is, I thought @original-brownbear wrote that code. In any case, I'll try to salvage original authorship.

original-brownbear:
@colinsurprenant this was just me apparently making some random mistake when rebasing :)

colinsurprenant:
@original-brownbear so if @wainersm is not the author of that code your authorship will be lost and I may I well just copy&paste the code over and create a new PR, unless you have another suggestion?

original-brownbear:
@colinsurprenant don't worry about it, just C&P it :)

colinsurprenant:
rebooted in #8978. Thanks for your help on this @original-brownbear.

PR #7004, opened 2017-05-01
Unify ConfigPart and SourceMeta into SourceWithMetadata

### LINKED ISSUE #7003: Refactor `ConfigPath` class with `SourceWithMetadata`
Instead of using a custom class for the source information lets use the same class from LIR `SourceWithMetadata`

andrewvc:
This unifies the two different config classes that represented mainly the same data. While this does expose a plain java class into ruby this works fine because ruby only needs to access and set values, not work with ruby return types.

Fixes #7003


andrewvc:
@ph specs now pass, ready for another round of review.

logstash-core/lib/logstash/agent.rb
```diff
@@ -134,10 +134,10 @@ def converge_state_and_update
 
     unless results.success?
       if auto_reload?
-        logger.debug("Count not fetch the configuration to converge, will retry", :message => results.error, :retrying_in => @reload_interval)
+        logger.debug("Could not fetch the configuration to converge, will retry", :message => results.error, :retrying_in => @reload_interval)
```
ph: I really need to add a spelling addons in my editor.

logstash-core/lib/logstash/config/source_loader.rb
```diff
@@ -54,7 +54,7 @@ def fetch
       else
         begin
           pipeline_configs = sources_loaders
-            .collect { |source| Array(source.pipeline_configs) }
```
ph: I am ok with that following the least surprise.

PR #7008, opened 2017-05-02
Normalize the Alpha version of Logstash

ph:
If you use 6.0.0-alpha1, rubygems will change it to 6.0.0.pre.alpha1,
this commit changes the behavior to normalize the version before saving
it to disk.

### DIFF
--- a/rakelib/version.rake
+++ b/rakelib/version.rake
@@ -59,7 +59,7 @@ namespace :version do
     text = IO.read(logstash_core_plugin_api_gemspec)
     IO.write(logstash_core_plugin_api_gemspec, text.sub(
       /  gem.add_runtime_dependency \"logstash-core\", \".+?\"/,
-      "  gem.add_runtime_dependency \"logstash-core\", \"#{logstash_core_version}\""))
+      "  gem.add_runtime_dependency \"logstash-core\", \"#{logstash_core_version.gsub("-", ".")}\""))
   end
 
   desc "show version of core components"
--- a/logstash-core-plugin-api/logstash-core-plugin-api.gemspec
+++ b/logstash-core-plugin-api/logstash-core-plugin-api.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
   gem.version       = LOGSTASH_CORE_PLUGIN_API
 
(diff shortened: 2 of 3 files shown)


logstash-core/lib/logstash-core/version.rb
```diff
@@ -5,4 +5,4 @@
 # Note to authors: this should not include dashes because 'gem' barfs if
 # you include a dash in the version string.
 
-LOGSTASH_CORE_VERSION = "6.0.0-alpha1"
+LOGSTASH_CORE_VERSION = "6.0.0.alpha1"
```
suyograo: @ph this is gonna mess up the unified release. We'll have different package names compared to other products like `logstash-6.0.0.alpha1.zip` instead of `logstash-6.0.0-alpha1.zip`

suyograo: Can we do this sub just for the gem?

rakelib/version.rake
```diff
@@ -50,7 +50,7 @@ namespace :version do
       next if metadata["yaml_version"] == metadata["current_version"]
       puts "Updating \"#{component}\" from \"#{metadata['current_version']}\" to \"#{metadata['yaml_version']}\""
       text = IO.read(metadata["location"])
-      IO.write(metadata["location"], text.gsub(metadata["current_version"], metadata["yaml_version"]))
+      IO.write(metadata["location"], text.gsub(metadata["current_version"], metadata["yaml_version"].gsub("-", '.')))
```
suyograo: We should do this just in the gem file.

logstash-core/lib/logstash-core/version.rb
ph: Crap, you are right, I thought we were using the yaml in the `rakelib/artifacts.rake` but we use the constant.

ph:
I've updated the PR.

suyograo:
LGTM. Not sure what the tests errors are..

ph:
legit error made by https://github.com/elastic/logstash/pull/7011, I didn't get around that because of lucky ordered run. I will make a quick fix right now.

ph:
@suyog fixed in https://github.com/elastic/logstash/pull/7017

PR #7010, opened 2017-05-02
add queue_size_on_disk stat to persisted queue node stats

original-brownbear:
This is a clean rebase of #6518 (no changes were necessary to the original commit in that PR just needed a `cherry-pick`).

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -71,7 +71,7 @@ module LogStash; class BasePipeline
       raise e
     end
   end
-  
+
   def compile_lir
     LogStash::Compiler.compile_pipeline(self.config_str)
   end
@@ -190,8 +190,8 @@ module LogStash; class Pipeline < BasePipeline
     @flushing = Concurrent::AtomicReference.new(false)
     @force_shutdown = Concurrent::AtomicBoolean.new(false)
   end # def initialize
-  
-  
+
+
 
   def ready?
     @ready.value
@@ -691,6 +691,7 @@ module LogStash; class Pipeline < BasePipeline
         n.gauge(:page_capacity_in_bytes, queue.page_capacity)
         n.gauge(:max_queue_size_in_bytes, queue.max_size_in_bytes)
         n.gauge(:max_unread_events, queue.max_unread_events)
+        n.gauge(:queue_size_in_bytes, queue.persisted_size_in_bytes)
       end
       pipeline_metric.namespace([:data]).tap do |n|
         n.gauge(:free_space_in_bytes, file_store.get_unallocated_space)
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -139,6 +139,10 @@ public class Queue implements Closeable {
         return this.currentByteSize;
     }
 
+    public long getPersistedByteSize() {
+        return headPage.getPageIO().getHead() + tailPages.stream().mapToLong((p) -> p.getPageIO().getHead()).sum();
+    }
+
     public int getPageCapacity() {
         return this.pageCapacity;
     }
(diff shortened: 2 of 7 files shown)


original-brownbear:
@colinsurprenant fyi!

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -691,6 +691,7 @@ def collect_stats
         n.gauge(:page_capacity_in_bytes, queue.page_capacity)
         n.gauge(:max_queue_size_in_bytes, queue.max_size_in_bytes)
         n.gauge(:max_unread_events, queue.max_unread_events)
+        n.gauge(:queue_size_on_disk, queue.current_persisted_bytes)
```
colinsurprenant: not sure I like `queue_size_on_disk` - all sizing metrics use xxx_in_bytes, I think we should stay consistent and use `queue_size_in_bytes`, WDYT?

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -139,6 +139,10 @@ public long getCurrentByteSize() {
         return this.currentByteSize;
     }
 
+    public long getCurrentPhysicallyPersistedByteSize() {
```
colinsurprenant: not sure why we use physically in `getCurrentPhysicallyPersistedByteSize`? why not just `getCurrentPersistedByteSize`?

colinsurprenant:
looks good, mainly naming comments.

logstash-core/lib/logstash/pipeline.rb
original-brownbear: @colinsurprenant makes sense, space/memory and no unit is evil :) fixing

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
original-brownbear: @colinsurprenant frankly I'd even just go with `getPersistedSize` :
* The fact that its `bytes` can go to JavaDoc
* current is redundant, what if not the current size is this supposed to be? :)

wdyt?

colinsurprenant: +1. maybe since it seems we are currently pretty much always using "bytes" in the names, I'd probably keep it for consistency.

colinsurprenant:
note - once this is in we should deprecate `getCurrentByteSize`, remove the `currentByteSize` handling code and have `getCurrentByteSize` just return  `getPersistedSize`/`getPersistedByteSize`.  we shouldn't have 2 ways of computing the queue size.

original-brownbear:
@colinsurprenant renamed the method :) and will keep you comment on the radar for the next PR :)

colinsurprenant:
Looking at this, sorry to go in circle here, I like your comment on removing the "current" from the names, it is weird :P what else could it be... 
Also, to stay consistent I'd rename `current_persisted_bytes` to `persisted_size_in_bytes` ... 
WDYT?

original-brownbear:
@colinsurprenant np renaming sec :)

original-brownbear:
@colinsurprenant done :)

colinsurprenant:
LGTM - only missing tests. 
I suggest we open an issue to add tests for this.

suyograo:
@colinsurprenant why can't we wait to add tests here? Why create a separate issue for a test?

colinsurprenant:
@suyograo oh, we definitely can add tests now but I feel this is blocking @original-brownbear to move on in #6998 so my comment was ment as "I am fine we merge this now to move on with #6998 but let's make sure we come back here to add tests". 
Obviously the preference is to have tests here, doing so it could uncover bugs that would need to be fixed in another PR.

original-brownbear:
@colinsurprenant @suyograo if it's just me who's blocked by this it doesn't make much of a difference if I add tests here or later I guess => I'll give tests a go later? :)

colinsurprenant:
@original-brownbear your call - as long as the tests land I am good.

suyograo:
@original-brownbear @colinsurprenant ok, but lets please remember to come back and add tests.

colinsurprenant:
@suyograo I believe we are in agreement 🙃

suyograo:
LGTM to merge then :)

suyograo:
@original-brownbear please squash and merge to 5.x and master.

original-brownbear:
@colinsurprenant @suyograo on it :)

PR #7015, opened 2017-05-03
Add docs for settings via env vars on Docker

jarpy:
- Document the new mechanism for mapping environment variables to setttings in `logstash.yml` when using the Docker image
- Add comments about Docker base image and source repo

Applies to 5.4.

### DIFF
--- a/docs/static/docker.asciidoc
+++ b/docs/static/docker.asciidoc
@@ -3,8 +3,16 @@
 Docker images for Logstash are available from the Elastic Docker
 registry.
 
-Obtaining Logstash for Docker is as simple as issuing a +docker pull+
-command against the Elastic Docker registry.
+The base image is https://hub.docker.com/_/centos/[centos:7] and the source
+code can be found on
+https://github.com/elastic/logstash-docker/tree/{branch}[GitHub].
+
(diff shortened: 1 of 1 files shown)


docs/static/docker.asciidoc
```diff
@@ -3,8 +3,16 @@
 Docker images for Logstash are available from the Elastic Docker
 registry.
 
-Obtaining Logstash for Docker is as simple as issuing a +docker pull+
-command against the Elastic Docker registry.
+The base image is https://hub.docker.com/_/centos/[Centos 7] and the source
```
dliappis: Let's try to stay consistent with the name here across all Docker docs for products. I used `centos:7` in Elasticsearch as this is the unambiguous way to reference the base image in Docker Hub. If we stick to this, the proper capitalization is `CentOS 7`.

jarpy:
Thanks. Updated based on your comments.

LS team, please merge to 5.4 if you are happy.

suyograo:
Thanks @jarpy

jarpy:
Cheers, @suyograo.

Sorry this patch was a little late. It would be good to backport it to 5.4 if possible.

PR #7017, opened 2017-05-04
Replace the class declaration with let statements

ph:
Instead of using a concrete class use let statement instead, this make
sure they are reset between run and make the variable available at the
context level.

This issue was found in https://github.com/elastic/logstash/pull/7008 and was generating the following errors. Depending of the ordering of the test, the class could be available or not.

```
  1) LogStash::Plugins::Registry when loading plugin manually configured allow you find plugin by type
     Failure/Error: registry.add(:filter, "simple_plugin", SimplePlugin)
     NameError:
       uninitialized constant SimplePlugin
     # ./logstash-core/spec/logstash/plugins/registry_spec.rb:77:in `(root)'
     # ./rakelib/test.rake:47:in `(root)'
```

### DIFF
--- a/logstash-core/spec/logstash/plugins/registry_spec.rb
+++ b/logstash-core/spec/logstash/plugins/registry_spec.rb
@@ -53,13 +53,12 @@ describe LogStash::Plugins::Registry do
   end
 
   context "when loading plugin manually configured" do
-    it "should return the plugin" do
-      class SimplePlugin
-      end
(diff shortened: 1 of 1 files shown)


andrewvc:
LGTM

ph:
Thanks @andrewvc

ph:
thanks @andrewvc

Issue #7022, opened 2017-05-04
Test: NullPointer Exception for `org.logstash.ackedqueue.Queue.getPersistedByteSize`


ph:
https://logstash-ci.elastic.co/job/elastic+logstash+5.x+multijob-os-compatibility/os=debian/264/console

```
  1) LogStash::Api::Modules::Plugins should return a 404 to unknown request
     Failure/Error: Unable to find matching line from backtrace
     Java::JavaLang::NullPointerException:
     # org.logstash.ackedqueue.Queue.getPersistedByteSize(Queue.java:143)
     # org.logstash.ackedqueue.ext.JrubyAckedQueueExtLibrary$RubyAckedQueue.ruby_persisted_size_in_bytes(JrubyAckedQueueExtLibrary.java:117)
     # org.logstash.ackedqueue.ext.JrubyAckedQueueExtLibrary$RubyAckedQueue$INVOKER$i$0$0$ruby_persisted_size_in_bytes.call(JrubyAckedQueueExtLibrary$RubyAckedQueue$INVOKER$i$0$0$ruby_persisted_size_in_bytes.gen)
     # org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:134)
     # org.jruby.ast.CallNoArgNode.interpret(CallNoArgNode.java:60)
     # org.jruby.ast.CallTwoArgNode.interpret(CallTwoArgNode.java:59)
     # org.jruby.ast.NewlineNode.interpret(NewlineNode.java:105)
```

```
rspec /var/lib/jenkins/workspace/elastic+logstash+5.x+multijob-os-compatibility/os/debian/logstash-core/spec/support/shared_examples.rb:99 # LogStash::Api::Modules::Plugins should return a 404 to unknown request
rspec /var/lib/jenkins/workspace/elastic+logstash+5.x+multijob-os-compatibility/os/debian/logstash-core/spec/api/lib/api/node_plugins_spec.rb:29 # LogStash::Api::Modules::Plugins retrieving plugins should return the total number of plugins

Randomized with seed 23098
```

ph:
This seems to be coming from https://github.com/elastic/logstash/commit/49993c5220dea8a1b219e4026d03fae1fd6c235a

original-brownbear:
@colinsurprenant @suyograo tests for https://github.com/elastic/logstash/pull/7010 incoming here

original-brownbear:
just for the record, this can be reproduced via:

```java
[...]

PR #7023, opened 2017-05-04
#7022 unopened queue was throwing npe

original-brownbear:
Handling #7022 

* Obvious NPE since the `headPage` isn't set until `open` is called on the `Queue`.


colinsurprenant:
LGTM

PR #7025, opened 2017-05-04
Replace images in deployment guide

dedemorton:
@acchen97 Here are the images that you wanted updated.


acchen97:
Thanks @dedemorton LGTM.

PR #7048, opened 2017-05-08
Refactor SourceLoader to no longer be a singleton.

andrewvc:
This cleans up the code from a design patterns standpoint and makes testing plugins easier since you can just create/destroy agents at will.

Without this change the SOURCE_LOADER singleton's state will become dirty as agents are created/destroyed and be problematic.


ph:
Lgtm waiting for ci

PR #7052, opened 2017-05-09
gitignore .lock

original-brownbear:
This file is left after clean builds still (100% of the time for me).


original-brownbear:
@ph assigning you since we discussed this already ... happens everytime now for me.

original-brownbear:
@ph thanks for the review :)

Issue #7054, opened 2017-05-09
Make the Pipeline constructor receive a PipelineConfig object


jsvd:
Currently, pipeline instantiation looks like:

```ruby
  pipeline_validator = LogStash::BasePipeline.new(@pipeline_config.config_string, @pipeline_config.settings)
```

or

```ruby
  LogStash::BasePipeline.new(pipeline_config.config_string)
```

and the Pipeline initializer looks like:

`def initialize(config_str, settings = SETTINGS, namespaced_metric = nil, agent = nil)`

IMO, **the pipeline should receive the instance of the PipelineConfig object directly instead of the config_string and settings portions of it**. This would make the constructor look like:

`def initialize(pipeline_config, namespaced_metric = nil, agent = nil)`

Introducing this change also allows us to detect changes at the pipeline configuration level during a reload, instead of just configuration string changes, fixing an open TODO left by @ph in [the state resolver](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/state_resolver.rb#L22-L26).

The downside of this change is that it makes testing a bit harder as it requires creating a PipelineConfig object to test a Pipeline.

andrewvc:
Plan of action after meeting with @jsvd: 

1. Remove PipelineConfig class
1. Add Java LIR Pipeline constructor for `Pipeline.new(source_with_metadata...)` to enable concatenation of configs.
1. Change  Ruby`Pipeline.new` signature to `(lir_pipeline, settings, namespaced_metric, agent)`
1. Add `pipeline_settings` into pipeline actions since removing `PipelineConfig` will necessitate this.
1. Determine which of our two confusing `Pipeline` classes needs to be renamed.

ph:
> The downside of this change is that it makes testing a bit harder as it requires creating a PipelineConfig object to test a Pipeline.

To help use migrate towards the new pattern we could just create a new helper generating generate the boilerplate classes from the previous signature and return a pipeline, WDYT?

jsvd:
we already have a few mocking tools for that https://github.com/elastic/logstash/blob/28e51f5c6ae839020fa2ee605ce1ef45f8f05166/logstash-core/spec/support/helpers.rb#L44-L56

jsvd:
closing in favor of #7076

PR #7055, opened 2017-05-09
Compile multiple sources together into one LIR pipeline

andrewvc:
This adds methods to the compiler that allow it to take multiple `SourceWithMetadata` objects and compile them together into one LIR Pipeline object.

This also sets us up to correctly set the protocol / metadata fields for LIR
pipelines and work on the refactor mentioned in https://github.com/elastic/logstash/issues/7054


andrewvc:
@jsvd I modified this PR to add the original source to the Pipeline so this could work with the existing classes.

PR #7058, opened 2017-05-09
Minor: Stop allocating byte[] to compute CRC32

original-brownbear:
Found this today while working on #6998, freebie performance improvement for `open()` imo :)

The `CRC32` can pass a `DirectByteBuffer` (which we have) straight to the underlying `native` call for the crc32 calculation which should be faster right away.
+ Saves us a ton of `byte[]` GC.

Also made the `CRC32` `private` since no child class was using it (and it's mutable and not threadsafe)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
@@ -1,15 +1,13 @@
 package org.logstash.ackedqueue.io;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import org.logstash.ackedqueue.SequencedList;
-
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


original-brownbear:
@suyograo @colinsurprenant fyi :)

colinsurprenant:
ah! nice! will test!

jordansissel:
LGTM

Do you want to update other calls for checksum() like https://github.com/original-brownbear/logstash/blob/2fd1918d9f70ac21c2bce05204788b7bb5d7ce7b/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java#L236-L239 ? (This can happen in another PR)

original-brownbear:
@jordansissel sure lets fix other cases too, can factor this into a utility method then and dry up a little :)

original-brownbear:
@jordansissel see https://github.com/elastic/logstash/issues/7111 :)

jordansissel:
@original-brownbear can you merge this into 5.x as well?

original-brownbear:
@jordansissel https://github.com/elastic/logstash/pull/7112 :) already did

PR #7062, opened 2017-05-10
#7053 Refactor queue settings class to use builder pattern

### LINKED ISSUE #7053: Refactor *Settings Classes to use Builder Pattern
As discussed in https://github.com/elastic/logstash/pull/5958#discussion_r83122350.
`org.logstash.ackedqueue.FileSettings` and `org.logstash.ackedqueue.MemorySettings` should be refactored towards a `Builder` approach here.

original-brownbear:
closes #7053 

* Removed duplicate settings class
* Implemented builder pattern and as a consequence immutable settings


colinsurprenant:
nice! will review this hopefully tomorrow.

original-brownbear:
@colinsurprenant what's the status here? I was pushing in the other one since it will make this one conflict but this one will make everything conflict anyways :D so I guess we could also merge this first :)

logstash-core/src/main/java/org/logstash/ackedqueue/SettingsImpl.java
```diff
@@ -0,0 +1,246 @@
...
+        private final int checkpointMaxAcks;
+
+        private final int checkpointMaxWrites;
+
+        private final int checkpointMaxInterval;
+
+        private BuilderImpl(final String dirForFiles) {
+            this(dirForFiles, null, null, null, 0, 0L,
+                0, 1024, 1024, 1000
+            );
```
robbavey: Is there a way to avoid having all these parameters passed in like this - it makes it tricky to understand what all of these parameters mean at a glance, and increases the risk of inadvertent transposition of properties

logstash-core/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueExtLibrary.java
```diff
@@ -69,20 +67,15 @@ public IRubyObject ruby_initialize(ThreadContext context, IRubyObject[] args)
...
-            s.setElementClass(Event.class);
-
-            this.queue = new Queue(s);
+            this.queue = new Queue(SettingsImpl.fileSettingsBuilder(args[0].asJavaString()).capacity(capacity)
+                .maxUnread(maxUnread).queueMaxBytes(queueMaxBytes)
+                .checkpointMaxAcks(checkpointMaxAcks)
+                .checkpointMaxWrites(checkpointMaxWrites)
+                .checkpointMaxInterval(checkpointMaxInterval)
+                .elementIOFactory(pageIOFactory).checkpointIOFactory(checkpointIOFactory)
+                .elementClass(Event.class).build());
```
robbavey: Does it make sense to split the creation of the settings into a separate variable, or indent so that it is immediately obvious where the boundaries are between the creation of the Queue object and the Settings object?

logstash-core/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueMemoryExtLibrary.java
```diff
@@ -67,15 +66,11 @@ public IRubyObject ruby_initialize(ThreadContext context, IRubyObject[] args)
...
             CheckpointIOFactory checkpointIOFactory = (source) -> new MemoryCheckpointIO(source);
-            s.setCapacity(capacity);
-            s.setMaxUnread(maxUnread);
-            s.setQueueMaxBytes(queueMaxBytes);
-            s.setElementIOFactory(pageIOFactory);
-            s.setCheckpointIOFactory(checkpointIOFactory);
-            s.setElementClass(Event.class);
+            Settings s = SettingsImpl.memorySettingsBuilder(args[0].asJavaString()).capacity(capacity)
+                .maxUnread(maxUnread).queueMaxBytes(queueMaxBytes).elementIOFactory(pageIOFactory)
+                .checkpointIOFactory(checkpointIOFactory).elementClass(Event.class).build();
```
robbavey: Do we have a preferred style for fluent/chained method calls - one call per line, or wrapping until line length boundaries?

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -329,8 +329,10 @@ public void reachMaxUnread() throws IOException, InterruptedException, Execution
         Queueable element = new StringElement("foobarbaz");
         int singleElementCapacity = ByteBufferPageIO.HEADER_SIZE + ByteBufferPageIO._persistedByteCount(element.serialize().length);
 
-        Settings settings = TestSettings.volatileQueueSettings(singleElementCapacity);
-        settings.setMaxUnread(2); // 2 so we know the first write should not block and the second should
+        Settings settings = SettingsImpl.builder(
+            TestSettings.volatileQueueSettings(singleElementCapacity)
+        ).maxUnread(2) // 2 so we know the first write should not block and the second should
+        .build();
```
robbavey: (nit) Indentation style seems a little off between this instance and the one below

logstash-core/src/main/java/org/logstash/ackedqueue/SettingsImpl.java
original-brownbear: @robbavey phew I hear you ... that's why I hate these "databag" style objects :D In the end you can't really avoid those constructor chains if you want immutability ...
You know we could extract the defaults into constants with some Javadoc and appropriate naming I guess ? :)

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
original-brownbear: @robbavey this one is the good one though right? (the one below looks messed up to me)

logstash-core/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueMemoryExtLibrary.java
original-brownbear: @robbavey not that I know of + the codebase doesn't really have any concrete standards enforced as far as I can see.
Let me try and make this nicer though, this is hard to read ...

logstash-core/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueExtLibrary.java
original-brownbear: @robbavey if you ask me, no separate variable is kind of pointless. If you only use it once, it should probably only be there once.
That said ... horrible indentation here makes it very hard to read too, sorry about that, fixing :)

logstash-core/src/main/java/org/logstash/ackedqueue/SettingsImpl.java
original-brownbear: @robbavey you know even going over these variables once and trying to come up with one liner JDoc ... looks like there is some serious redundancy here :D (should probably tackle that soon)

robbavey: I hear your pain too ;)
I think constants and JavaDoc would be a great place to start from

logstash-core/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueExtLibrary.java
robbavey: I typically tend to agree that single use objects are kind of pointless, unless they act as a kind of 'documentation' about the object being created.

That said, I think indentation would help a lot here

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
robbavey: I think so

original-brownbear:
@robbavey bam cleaned this up a lot I think and identified one unused setting! :O 

Let me know what you think :)

logstash-core/src/test/java/org/logstash/stress/Concurrent.java
```diff
@@ -17,32 +9,36 @@
...
     public static Settings fileSettings(int capacity) {
-        Settings s = new MemorySettings("/tmp/queue");
         PageIOFactory pageIOFactory = (pageNum, size, path) -> new MmapPageIO(pageNum, size, path);
         CheckpointIOFactory checkpointIOFactory = (source) -> new FileCheckpointIO(source);
-        s.setCapacity(capacity);
-        s.setElementIOFactory(pageIOFactory);
-        s.setCheckpointIOFactory(checkpointIOFactory);
-        s.setElementClass(StringElement.class);
-        return s;
+        return SettingsImpl.memorySettingsBuilder("/tmp/queue").capacity(capacity)
```
original-brownbear: @robbavey urgh `"/tmp/queue"` hard set is evil ... we gotta add an issue for that

logstash-core/src/main/java/org/logstash/ackedqueue/SettingsImpl.java
```diff
@@ -0,0 +1,252 @@
...
+        private final long queueMaxBytes;
+
+        private final int maxUnread;
+
+        private final int checkpointMaxAcks;
+
+        private final int checkpointMaxWrites;
+        
+        private BuilderImpl(final String dirForFiles) {
+            this(dirForFiles, null, null, null, DEFAULT_CAPACITY, DEFAULT_MAX_QUEUE_BYTES,
```
original-brownbear: @robbavey I think these `null` values can also be killed easily. It seems to me there is a clear set of combinations we can have here, that we can probably leverage to fix this clean. (just adding the note so I don't forget to add an issue)

logstash-core/src/main/java/org/logstash/ackedqueue/ext/JrubyAckedQueueMemoryExtLibrary.java
```diff
@@ -65,15 +62,16 @@ public IRubyObject ruby_initialize(ThreadContext context, IRubyObject[] args)
...
+            this.queue = new Queue(
+                SettingsImpl.memorySettingsBuilder(args[0].asJavaString())
+                    .capacity(capacity)
+                    .maxUnread(maxUnread)
+                    .queueMaxBytes(queueMaxBytes)
+                    .elementIOFactory(ByteBufferPageIO::new)
+                    .checkpointIOFactory(MemoryCheckpointIO::new)
+                    .elementClass(Event.class)
+                    .build()
+            );
```
robbavey: +1 on this style

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -61,7 +61,6 @@
     private final int maxUnread;
     private final int checkpointMaxAcks;
     private final int checkpointMaxWrites;
-    private final int checkpointMaxInterval;
```
robbavey: I'm guessing that this was intended to write a checkpoint after some amount of time, regardless of the number of acks/writes, but was never implemented?

original-brownbear: @robbavey my guess too, I think this is still being debated even, but as long as it's not functional I see no point in keeping it around and making things even more complicated than they already are :)

robbavey: agreed

original-brownbear:
@robbavey thanks for the review! Let me squash and wait for Travis judgement then :)

PR #7064, opened 2017-05-10
make rspec use the format documentation when running any suite

ph:
We suite from the default dot formatter to the documentation format,
this will help localize some error in the build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,18 +11,16 @@ rvm:
 jdk:
   - oraclejdk8
 env:
-  - INTEGRATION=false
-  - INTEGRATION=true
-  - INTEGRATION=false FEATURE_FLAG=persistent_queues
-  - INTEGRATION=true FEATURE_FLAG=persistent_queues
+  - INTEGRATION=false SPEC_OPTS="--order rand --format documentation"
+  - INTEGRATION=true SPEC_OPTS="--order rand --format documentation"
+  - INTEGRATION=false FEATURE_FLAG=persistent_queues SPEC_OPTS="--order rand --format documentation"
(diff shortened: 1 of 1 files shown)


colinsurprenant:
why do we pass parameters using `.rspec` instead of using env vars?

ph:
@colin good question, I will move it to env. We were already using the .rspec file in the `travis.yaml` I didn't ask myself why we were doing it.

.gitignore
```diff
@@ -42,3 +42,4 @@ Gemfile
 logs
 qa/integration/services/installed/
 .gradle
+.rspec
```
colinsurprenant: is this still needed?

colinsurprenant:
minor comment, did not test, LGTM.

ph:
Weird Its doesn't seem to produce any output on travis, will check again. It should be the right variable.

colinsurprenant:
weird. `SPEC_OPTS` is the right env var and `export SPEC_OPTS="--order rand --format documentation"` should work - it works locally for me. 

Maybe it has to do with the way Travis executes the yml steps?

I see Travis provides support for setting env var https://docs.travis-ci.com/user/environment-variables/ - maybe use that?

ph:
Yeah I will try that, I just need to add them to the existing env. https://github.com/elastic/logstash/blob/master/.travis.yml#L11-L15

ph:
@colinsurprenant Travis' environment variable works like a charm.

colinsurprenant:
@ph ah! perfect!

colinsurprenant:
LGTM :shipit:

Issue #7045, opened 2017-05-08
NullpointerException when Environment-Variable http_proxy is set but empty


pretoritus:
Hi,

i get following error when i try to execute logstash-plugin:

```
./logstash-5.4.0$ bin/logstash-plugin 
java/util/Hashtable.java:459:in `put': java.lang.NullPointerException
        from java/util/Properties.java:166:in `setProperty'
        from java/lang/System.java:796:in `setProperty'
        from java/lang/reflect/Method.java:498:in `invoke'
        from /home/jupp/entwicklung/software/elastic/5.4.0/logstash-5.4.0/lib/pluginmanager/proxy_support.rb:40:in `apply_env_proxy_settings'
        from /home/jupp/entwicklung/software/elastic/5.4.0/logstash-5.4.0/lib/pluginmanager/proxy_support.rb:62:in `configure_proxy'
        from /home/jupp/entwicklung/software/elastic/5.4.0/logstash-5.4.0/lib/pluginmanager/main.rb:26:in `(root)'
```

My Java:
```
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)
```

My System:
```
Debian 9.0 4.9.0-2-amd64 #1 SMP Debian 4.9.18-1 (2017-03-30) x86_64 GNU/Linux
```

I am not using a proxy, but my env_vars are set and empty (Standard on debian??)
```
$: env | grep proxy
https_proxy=
http_proxy=
ftp_proxy=
```

When I unset die Environment-Variable the Exception is not thrown.
This issue affects only Logstash 5.4.0. Logstash 5.3.1 works fine in this situation.

ph:
I will work on a fix to include it in 5.4.1.

ph:
confirmed

```
 ph@sashimi  ~/es/logstash   master ± export https_proxy=""
 ph@sashimi  ~/es/logstash   master ± bin/logstash-plugin list
java/util/Hashtable.java:459:in `put': java.lang.NullPointerException
	from java/util/Properties.java:166:in `setProperty'
	from java/lang/System.java:796:in `setProperty'
	from java/lang/reflect/Method.java:498:in `invoke'
	from /Users/ph/es/logstash/lib/pluginmanager/proxy_support.rb:40:in `apply_env_proxy_settings'
[...]

PR #7065, opened 2017-05-10
Check for empty string when configuring proxies

ph:
On some Linux system like debian, the `HTTP_PROXY` are set to an empty
string and was causing the plugin manager to raise an exception. We now
strip the string and check if it is empty before trying to configure the
proxies.

Fixes: #7045

### DIFF
--- a/lib/pluginmanager/proxy_support.rb
+++ b/lib/pluginmanager/proxy_support.rb
@@ -56,14 +56,16 @@ end
 
 def configure_proxy
   proxies = []
-  if proxy = (ENV["http_proxy"] || ENV["HTTP_PROXY"])
+  proxy = (ENV["http_proxy"] || ENV["HTTP_PROXY"])
+  if !proxy.nil? && !proxy.strip.empty?
     proxy_settings = extract_proxy_values_from_uri(proxy)
     proxy_settings[:protocol] = "http"
     apply_env_proxy_settings(proxy_settings)
     proxies << proxy_settings
   end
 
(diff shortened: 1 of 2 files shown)


lib/pluginmanager/proxy_support.rb
```diff
@@ -56,14 +56,16 @@ def extract_proxy_values_from_uri(proxy_uri)
 
 def configure_proxy
   proxies = []
-  if proxy = (ENV["http_proxy"] || ENV["HTTP_PROXY"])
+  proxy = (ENV["http_proxy"] || ENV["HTTP_PROXY"])
```
suyograo: Can we extract this into a small function?

lib/pluginmanager/proxy_support.rb
```diff
@@ -56,14 +56,16 @@ def extract_proxy_values_from_uri(proxy_uri)
...
+  if !proxy.nil? && !proxy.strip.empty?
     proxy_settings = extract_proxy_values_from_uri(proxy)
     proxy_settings[:protocol] = "http"
     apply_env_proxy_settings(proxy_settings)
     proxies << proxy_settings
   end
 
-  if proxy = (ENV["https_proxy"] || ENV["HTTPS_PROXY"])
+  proxy = (ENV["https_proxy"] || ENV["HTTPS_PROXY"])
```
suyograo: We can reuse the extracted function here

ph:
@suyograo I have extracted the conditionals into 2 methods.

PR #7066, opened 2017-05-10
fix a few racing conditions in the agent_spec

jsvd:
this replaces https://github.com/elastic/logstash/pull/6996

it removes a lot of sleeps in agent_spec and also some of the race conditions in the configuration reloading

### DIFF
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -10,18 +10,20 @@ require_relative "../support/helpers"
 require_relative "../support/matchers"
 
 describe LogStash::Agent do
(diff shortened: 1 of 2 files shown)


logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -226,64 +205,16 @@
...
+          # TODO this is a bad practice, any suggestions on how to test something happens
+          # without some form of timing or expiring condition?
+          sleep 0.1
           Stud.stop!(t)
           t.join
           subject.shutdown
         end
       end
-
-      context "when calling reload_state!" do
```
jsvd: I removed these tests since this is tested in the state resolver

Issue #6851, opened 2017-03-29
Logstash modules


suyograo:
## Introduction

The idea is to explore _modules_ for Logstash, similar to Filebeat modules feature released in 5.3.0. Modules contain packaged Logstash configuration, Kibana dashboards and other meta files to ease the set up of the Elastic stack for certain use cases or data sources. The goal of these modules are to provide an end-end, 5 min getting started experience for a user exploring a data source without having to learn different parts of the stack (initially).

## Data sources

The initial goal is to focus on data sources connecting over-the-network to complement modules in beats.

## Behavior

Users interact with modules in the following ways:

### Command Line

```sh
bin/logstash --modules netflow
```

This will instruct Logstash to use a module and be ready to accept data or start pulling from a data source. Internally, modules subcommand should:
1. Load the Kibana dashboard to Elasticsearch `.kibana` index. If there exists a dashboard with the same name, it should be overwritten.
2. Load the ES template for this module if not existing already.
3. Load other configuration files for the Stack components to their respective ES indexes or directly using the API.
4. Start Logstash with the module's config. In this mode the configuration file is not persisted to disk and is loaded onto memory (as `-e` option).
5. When modules are used, `-f` or loading an arbitrary configuration from a file source is disabled. This is to avoid any accidental inclusion of other configurations which could cause the modules to not work correctly.

You can load multiple modules as such:

```sh
bin/logstash --modules netflow, foo
```

### File based configuration

If you prefer to use file based configuration for enabling modules (in lieu of the CLI), you can use `logstash.yml`.

```yaml
modules:
- name: netflow
- name: foo
```

### Overriding options

The out of the box module configuration assumes Elasticsearch is installed on the same host as Logstash. It also assumes the data sources are local. You can customize such _stock_ configuration for your environment.

For example, you can point to a remote ES host when running the module:

```
bin/logstash --modules netflow -M "netflow.var.elasticsearch.host=es.mycloud.com"
```

```
bin/logstash --modules netflow -M "netflow.var.tcp.port=5606"
```

Each module will define its own variables that user can override. These are light-weight overrides, as in, we wouldn't expose the entire LS pipeline to be overridden.

In the logstash.yml

```yaml
modules:
- name: netflow
   var.output.elasticsearch.host: "es.mycloud.com"
   var.output.elasticsearch.user: "foo"
   var.output.elasticsearch.password: "password"
   var.input.tcp.port: 5606
```

The variables will then be injected into the Logstash pipeline.

[...]

PR #7072, opened 2017-05-11
#6785 Interpolate ENV into settings read from logstash.yml

### LINKED ISSUE #6785: Feature request: Environment variable expansion in logstash.yml
In some situations, it would be handy to expand environment variables in `logstash.yml`, similar to [how it's done for pipeline configuration](https://www.elastic.co/guide/en/logstash/current/environment-variables.html).

For my specific purpose, this would enable setting various options with environment variables passed by Docker to Logstash running in a container, as suggested/discussed in [this logstash-docker issue](https://github.com/elastic/logstash-docker/issues/15).

What say you, Team Logstash?

original-brownbear:
closes #6785 

Added interpolation for env variables into settings:
* spec will only work on *nix, substitution should work on Windows as well
* ENV vars containing a dot work as shown by the spec

### DIFF
--- a/logstash-core/lib/logstash/settings.rb
+++ b/logstash-core/lib/logstash/settings.rb
@@ -2,11 +2,14 @@
 require "logstash/util/loggable"
 require "fileutils"
 require "logstash/util/byte_value"
+require "logstash/util/environment_variables"
 require "logstash/util/time_value"
 
 module LogStash
   class Settings
 
+    include LogStash::Util::EnvironmentVariables
+    
     def initialize
       @settings = {}
       # Theses settings were loaded from the yaml file
@@ -108,8 +111,10 @@ module LogStash
 
     def from_yaml(yaml_path)
       settings = read_yaml(::File.join(yaml_path, "logstash.yml"))
-      substitute_env(settings)
-      self.merge(flatten_hash(settings), true)
+      self.merge(
+        deep_replace(flatten_hash(settings)),
+        true
+      )
       self
     end
     
@@ -148,15 +153,6 @@ module LogStash
         h.each { |k,r| flatten_hash(r,"#{f}.#{k}",g) }
       end
       g
-    end
-
-    # Substitutes environment variables into setting values of string type.
-    def substitute_env(settings)
-      settings.each do |_, setting|
-        if setting.is_a? String
-          setting.gsub!(/\$([a-zA-Z_.][a-zA-Z0-9_.]*)|\${\g<1>}|%\g<1>%/) {ENV[$1]}
-        end
-      end
     end
   end
 
--- a/logstash-core/lib/logstash/settings.rb
+++ b/logstash-core/lib/logstash/settings.rb
@@ -108,6 +108,7 @@ module LogStash
 
     def from_yaml(yaml_path)
       settings = read_yaml(::File.join(yaml_path, "logstash.yml"))
+      substitute_env(settings)
       self.merge(flatten_hash(settings), true)
       self
     end
@@ -147,6 +148,15 @@ module LogStash
         h.each { |k,r| flatten_hash(r,"#{f}.#{k}",g) }
       end
       g
+    end
+
+    # Substitutes environment variables into setting values of string type.
+    def substitute_env(settings)
+      settings.each do |_, setting|
+        if setting.is_a? String
+          setting.gsub!(/\$([a-zA-Z_.][a-zA-Z0-9_.]*)|\${\g<1>}|%\g<1>%/) {ENV[$1]}
+        end
+      end
     end
   end
 
--- a/logstash-core/lib/logstash/util/environment_variables.rb
+++ b/logstash-core/lib/logstash/util/environment_variables.rb
@@ -1,0 +1,43 @@
+# encoding: utf-8
+module ::LogStash::Util::EnvironmentVariables
+
+  ENV_PLACEHOLDER_REGEX = /\${(?<name>[a-zA-Z_.][a-zA-Z0-9_.]*)(:(?<default>[^}]*))?}/
+
+  # Recursive method to replace environment variable references in parameters
+  def deep_replace(value)
+    if value.is_a?(Hash)
+      value.each do |valueHashKey, valueHashValue|
+        value[valueHashKey.to_s] = deep_replace(valueHashValue)
+      end
+    else
+      if value.is_a?(Array)
+        value.each_index do | valueArrayIndex|
+          value[valueArrayIndex] = deep_replace(value[valueArrayIndex])
+        end
+      else
+        return replace_env_placeholders(value)
+      end
+    end
+  end
+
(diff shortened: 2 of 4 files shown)


original-brownbear:
@guyboertje so about our discussion just now:

* Expected behaviour on missing env variables is to substitute empty string, but at least Dropwizard and Docker Compose do support the bash default notation of `${VARIABLE:-default}`, I guess so should we?

original-brownbear:
@guyboertje Windows style interpolation works too, added to the UT :)

original-brownbear:
@guyboertje @jsvd also, I think we need to support nesting right? Currently the substitution only works on the yaml's top level, this is not enough is it?

guyboertje:
@original-brownbear RE: `Dropwizard and Docker Compose do support the bash default notation of ${VARIABLE:-default}, I guess so should we?`
My suggestion is, use what you have now and you create an design discussion issue in GH and invite LS team by email to discuss in the issue.

original-brownbear:
@guyboertje sounds good, will add the issue after we merge.
The nesting is actually trivial to resolve sec since we flatten the settings hash. Pushing fix sec

guyboertje:
@ph - what do you think about [Armin's question](https://github.com/elastic/logstash/pull/7072#issuecomment-300774738)

original-brownbear:
@guyboertje fix pushed for the nesting thing :)

ph:
> @guyboertje @jsvd also, I think we need to support nesting right? Currently the substitution only works on the yaml's top level, this is not enough is it?

Yes we need to support nesting, We also need to support array of hashes like we will have when modules land see this [comment](https://github.com/elastic/logstash/issues/6851#issuecomment-290185709) for a sample.

original-brownbear:
@ph got it, though I think I'm covering the currently implemented case via the `flatten_hash` call coming before the interpolation. Not sure what the plans for implementing the array of hashes is, maybe they'll also be covered by that?

logstash-core/lib/logstash/settings.rb
```diff
@@ -148,6 +148,16 @@ def flatten_hash(h,f="",g={})
       end
       g
     end
+
+    # Substitutes environment variables into setting values of string type.
+    def substitute_env(settings)
+      settings.each do |_, setting|
+        if setting.is_a? String
```
ph: @original-brownbear I don't think they will be, because settings will return `Array` instead of `String`, I think we need to make the algorithm visit the child node.

ph:
@original-brownbear I've comment directly on the code https://github.com/elastic/logstash/pull/7072/files#r115983905

original-brownbear:
@ph I think I phrased my question wrong :) Currently we don't have those arrays right? Will those still be arrays after running through `flatten_hash`?

ph:
>Currently we don't have those arrays right?

@original-brownbear Kinda, we can have theses array with x-pack, when you use the `ArrayCoercible` settings.

```ruby
settings.register(LogStash::Setting::ArrayCoercible.new("hosts", String, [ "http://localhost:9200" ] ))
```
This will translate in the `logstash.yml` file to:


```yaml
hosts: ["http://dev1.localhost:9200", "http://dev2.localhost:9200"]
```

Take into consideration that people can also write the following and the settings class will do the right thing.

```yaml
hosts: "http://dev1.localhost:9200"
```

We do have embedded hash [like this](https://github.com/elastic/logstash/issues/6851#issuecomment-290185709), but that feature it's just around the corner for 5.5 and I think for feature complete we should support it.

suyograo:
@original-brownbear this is a good change. I have a suggestion:

We do environment variable resolving for the pipeline config. That code is in `replace_env_placeholders` https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/config/mixin.rb#L163

I think we should extract that `replace_env_placeholders` logic into a new utils `module LogStash::Util::EnvironmentVariables` and then reuse this in both places (settings for the yaml, and mixin for the pipeline config). 

WDYT? That way we are consistent in how we resolve EV in both places and no surprises for users.

original-brownbear:
@suyograo sounds good ,on it :)

original-brownbear:
@suyograo @guyboertje done, now we also have default value resolution.
I also fixed the `.` not working with the existing regex, Windows I didn't fix now though to not make this extra hard to review (unless you want me to do it here, but I'd have to adjust that regex more for that).

guyboertje:
LGTM
@ph - please cast your all-knowing eye over this too.

logstash-core/lib/logstash/util/environment_variables.rb
```diff
@@ -0,0 +1,43 @@
...
+    else
+      if value.is_a?(Array)
+        value.each_index do | valueArrayIndex|
+          value[valueArrayIndex] = deep_replace(value[valueArrayIndex])
+        end
+      else
+        return replace_env_placeholders(value)
+      end
+    end
+  end
```
ph: +1 for deep replace.

logstash-core/spec/logstash/settings_spec.rb
```diff
@@ -146,4 +147,46 @@
...
+        expect(subject.get('interpolated')).to eq("missing")
+        expect(subject.get('with_dot')).to eq("missing")
+        ENV['SOME_LOGSTASH_SPEC_ENV_VAR'] = "correct_setting"
+        ENV['some.logstash.spec.env.var'] = "correct_setting_for_dotted"
+        subject.from_yaml(yaml_path)
+        expect(subject.get('interpolated')).to eq("correct_setting")
+        expect(subject.get('with_dot')).to eq("correct_setting_for_dotted")
+      end
+    end
+  end
```
ph: Can we add  2 tests for the following nested scenario, that could be generated from the following yaml?:

```yaml
host: ["dev1.${domain}", "dev2.${domain}"]
```
```yaml
modules:
 - name: ${domain}
    testing: ${domain}
 - name: ${domain2}
    testing: ${domain2}
```

logstash-core/lib/logstash/util/environment_variables.rb
original-brownbear: @ph do I have to do anything here or are you just happy about deep replace? :D

logstash-core/spec/logstash/settings_spec.rb
original-brownbear: @ph sure will do in a bit :)

logstash-core/lib/logstash/util/environment_variables.rb
ph: nothing to do, pr review is not all about what is missing.

logstash-core/spec/logstash/settings_spec.rb
original-brownbear: @ph added this case (used different env var names since I was paranoid about collisions :D, but other than that implemented exactly those two I think)

original-brownbear:
@ph all green, good to merge? :)

ph:
go ahead @original-brownbear

PR #7073, opened 2017-05-11
[wip] first step in refactoring Pipeline.new signature

jsvd:
this PR changes the Pipeline constructor signature to accept a PipelineConfig object instead of a config_string + settings object.

the next step is to make it receive a LIRPipeline + settings object, as described in https://github.com/elastic/logstash/issues/7054#issuecomment-300195197

TODO: changes in the pipeline constructor signature require a change in logstash-devutils due to the `sample` helpers


suyograo:
@jsvd should we pick this up again? Can we perhaps rewrite this in Java?

Issue #6949, opened 2017-04-21
build integration test failure


colinsurprenant:
As reported in #6930

```
Failures:
  1) Test Monitoring API can retrieve JVM stats
     Failure/Error: expect(result["jvm"]["uptime_in_millis"]).to be > 100
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./specs/monitoring_api_spec.rb:46:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # ./specs/monitoring_api_spec.rb:44:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
Finished in 14 minutes 23 seconds (files took 6.41 seconds to load)
34 examples, 1 failure
Failed examples:
rspec ./specs/monitoring_api_spec.rb:39 # Test Monitoring API can retrieve JVM stats
Randomized with seed 14328
```
```
Failures:
  1) Test Monitoring API can retrieve queue stats
     Failure/Error: expect(result["pipeline"]["queue"]).not_to be_nil
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./specs/monitoring_api_spec.rb:57:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # ./specs/monitoring_api_spec.rb:55:in `(root)'
[...]

PR #7074, opened 2017-05-11
fix monitoring api integration race condition

jsvd:
because the metric subsystem may not be ready yet when the tests ask for
the values, the test code may generate a NoMethodError accessing nested
hashes.

The NoMethodError exception aborts the try mechanism instead of retrying.

This PR rescues potential exceptions and adds more expectations to
protect the try block.

fixes https://github.com/elastic/logstash/issues/6949


jsvd:
TODO: backport this to 5.x

Issue #7077, opened 2017-05-11
[test failure] Test Monitoring API can retrieve event stats


jsvd:
seen in: https://travis-ci.org/elastic/logstash/jobs/231152367#L2017

```
  1) Test Monitoring API can retrieve event stats
     Failure/Error: expect(result["in"]).to eq(number_of_events)
     
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./specs/monitoring_api_spec.rb:32:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # ./specs/monitoring_api_spec.rb:29:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
```

andrewvc:
Probably correlated with https://github.com/elastic/logstash/issues/7086

jsvd:
Another failure for this test after #7145 was merged:
master - https://travis-ci.org/elastic/logstash/jobs/234592533#L1921-L1941
5.4 - https://travis-ci.org/elastic/logstash/jobs/234592576#L1796-L1816
5.x - https://travis-ci.org/elastic/logstash/jobs/234592566#L1842-L1864

jsvd:
fixed by #7177 and  #7145

PR #7078, opened 2017-05-11
fix remaining race condition in monitoring api integration spec

jsvd:
fixes https://github.com/elastic/logstash/issues/7077


jsvd:
merged into master, 5.x and 5.4

PR #7080, opened 2017-05-11
fix the ordering issue with the proxy test

ph:
Depending of the order of the test some test didn't not clear correctly
the environment.

Original build log at https://travis-ci.org/elastic/logstash/jobs/231152364
You can reproduce with the seed `19769`


jsvd:
the failure on travis is unrelated to this PR, merge please :)

Issue #6737, opened 2017-02-20
LogStash::Runner pipeline settings when :pipeline_workers is not defined by the user should not pass the value to the pipeline


jsvd:
failure from master (hash 9cc165a428a356c974d7ae56f4789cd77a6b3f74), running `rake test:core`:

```
Failures:

  1) LogStash::Runner pipeline settings when :pipeline_workers is not defined by the user should not pass the value to the pipeline
     Failure/Error: expect(settings.set?("pipeline.workers")).to be(false)
       
       expected false
            got true
     # ./logstash-core/spec/logstash/runner_spec.rb:261:in `(root)'
     # ./logstash-core/lib/logstash/runner.rb:357:in `create_agent'
     # ./logstash-core/lib/logstash/runner.rb:272:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:67:in `run'
     # ./logstash-core/lib/logstash/runner.rb:185:in `run'
     # ./vendor/bundle/jruby/1.9/gems/clamp-0.6.5/lib/clamp/command.rb:132:in `run'
     # ./logstash-core/spec/logstash/runner_spec.rb:264:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:49:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:194:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:187:in `invoke_with_call_chain'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:180:in `invoke'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:152:in `invoke_task'
     # ./vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/application.rb:108:in `top_level'
[...]

PR #7082, opened 2017-05-11
bump version to 6.0.0-alpha2

jsvd:


andrewvc:
@jsvd there is a test failure, but it looks anomalous. LGTM if you concur

jsvd:
@andrewvc that failure is registered here: https://github.com/elastic/logstash/issues/6737

PR #7084, opened 2017-05-12
#7083 document logstash.yml ENV interpolation

### LINKED ISSUE #7083: Document using Environment Variable Interpolation in Settings File
Document the fix for #6785  as discussed in https://github.com/elastic/logstash/issues/6785#issuecomment-300983042

original-brownbear:
closes #7083 

Simple documentation of an example of the setting value interpolation.

### DIFF
--- a/docs/static/settings-file.asciidoc
+++ b/docs/static/settings-file.asciidoc
@@ -27,6 +27,24 @@ pipeline.batch.size: 125
 pipeline.batch.delay: 5
 -------------------------------------------------------------------------------------
 
+The `logstash.yml` file also supports bash-style interpolation of environment variables in
+setting values.
+
+[source,yaml]
+-------------------------------------------------------------------------------------
+pipeline:
+  batch:
+    size: ${BATCH_SIZE}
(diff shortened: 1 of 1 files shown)


docs/static/settings-file.asciidoc
```diff
@@ -27,6 +27,22 @@ pipeline.batch.size: 125
...
+setting values.
+
+[source,yaml]
+-------------------------------------------------------------------------------------
+pipeline:
+  batch:
+    size: ${BATCH_SIZE}
+    delay: ${BATCH_DELAY:5}
+node:
+  name: "node_${LS_NODE_NAME}"
```
jsvd: I'd also use the default value notation in the string interpolation example so it leaves no doubt that the default feature is supported in both scenarios.

original-brownbear:
@jsvd added string interpolation default example :)

jsvd:
squash and merge :shipit:

original-brownbear:
@jsvd thanks  for the review!

PR #7088, opened 2017-05-12
#7081 up Kafka IT version to 0.10.2.1

original-brownbear:
it's in the title, upgrade Kafka to `0.10.2.1`

Issue #7081, opened 2017-05-11
[integration test failure] Test Kafka Input can ingest 37 apache log lines from Kafka broker


jsvd:
seen in https://travis-ci.org/elastic/logstash/jobs/231200764#L2253-L2274

```
  1) Test Kafka Input can ingest 37 apache log lines from Kafka broker
     Failure/Error: expect(@fixture.output_exists?).to be true
     
       expected true
            got false
     # ./specs/kafka_input_spec.rb:24:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/logstash-devutils-1.3.1-java/lib/logstash/devutils/rspec/logstash_helpers.rb:18:in `try'
     # ./specs/kafka_input_spec.rb:23:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
Finished in 15 minutes 4 seconds (files took 5.76 seconds to load)
32 examples, 1 failure
Failed examples:
rspec ./specs/kafka_input_spec.rb:19 # Test Kafka Input can ingest 37 apache log lines from Kafka broker
```

original-brownbear:
@suyograo closing here, this should be good now

jsvd:
@original-brownbear could this still be an instance of the same problem ? https://travis-ci.org/elastic/logstash/jobs/232651438#L1924-L1939

original-brownbear:
@jsvd I don't think so, that's not a Kafka related failure at all right? Seems almost like coming from a slight change to the data in the geoip dataset?

jsvd:
indeed, I'll open an separate issue

PR #7090, opened 2017-05-12
#7081 up Kafka IT version to 0.10.2.1

original-brownbear:
5x rebase of #7088


suyograo:
@original-brownbear for 5.x we don't need to update to 0.10.2.1 since we are still supporting only the 0.10 version of Kafka. The new version will be shipped in 6.x. 

Is Kafka 0.10.2.1 client compatible with 0.10.0.0 broker? do you know?

original-brownbear:
@suyograo yes it's compatible with all `0.10.x` :)

suyograo:
Cool this should be good to merge when travis passes then. Ideally, we should test for both versions of Kafka here (0.10.0.0) and (0.10.2.0) but we can save that up for our new integration test changes

original-brownbear:
@suyograo merging then :)

PR #7092, opened 2017-05-12
#7081 fix portability issues with Kafka Broker IPV4 and log path creation

original-brownbear:
This we need for a dockerized version too anyhow :)

* Kafka needs to advertise on loopback (this should fix the error we see here: https://travis-ci.org/elastic/logstash/jobs/231200764#L1525 , we'll see the same issue if we use Docker)
* On my mac it failed to recursively create the Kafka log path itself, would be nice to have this fixed too for debugging Dockerized spots

### DIFF
--- a/qa/integration/services/kafka_setup.sh
+++ b/qa/integration/services/kafka_setup.sh
@@ -2,6 +2,8 @@
 set -ex
 current_dir="$(dirname "$0")"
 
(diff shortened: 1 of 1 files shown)


original-brownbear:
@suyograo this should fix current issues without Docker, on it with the Docker version though :)

suyograo:
sweeeeet. this had annoyed me and I'd failed to make sense of it :)

LGTM

Issue #7091, opened 2017-05-12
Move Kafka Service in Integration Tests to a Dockerized Version


original-brownbear:
@andrewvc @suyograo as discussed, let's start with Kafka and track this here:

* Dockerize whatever is is happening in `kafka-setup.sh` (and the other Kafka scripts) and plug it into the existing service infrastructure with the fixtures and such

original-brownbear:
@suyograo @andrewvc there we go :) Will add follow up issues for the other services tomorrow.

suyograo:
@original-brownbear +1. we should also remove the non-dockerized services as well

original-brownbear:
closing merged to `master` and `5.x`, will link issues for the other services in a bit

PR #7093, opened 2017-05-12
#7091 Dockerized kafka

original-brownbear:
Dockerized Kafka as step 1 for #7093 

Note:
* Using random ports is very tricky when it comes to Kafka and Zookeeper if we want to retain compatibility with OSX Docker unfortunately. This requires finding a free port for Zookeeper on the host machine first and then setting up Zookeeper and Kafka accordingly (we cannot simply bind zookeeper via Docker `--bind` as Kafka and the rspec tests on the host need to "see" Zookeeper at the same address).
  * Solutions that would work if we need it (I don't think we do right now, but we could use this if we ever had to)
     * Run Logstash in its own Docker container as well and make the tests interact with the filesystem and shell of that container => makes the networking 100% portable across standard Docker, OSX Docker and even Swarm


suyograo:
+1, based on our zoom discussion with @andrewvc let's move forward with this approach.

original-brownbear:
@suyograo @andrewvc now it's green :) Needed to hack via

```bash
yes | gem uninstall -q -i /home/travis/.rvm/gems/jruby-1.7.25@global bundler
```

:)

original-brownbear:
@suyograo @andrewvc cleaned up the solution, see my edit to the PR description -> I don't think more sophisticated networking should be in step 1. This approach is the only one that I see working across OSX and Linux and that doesn't require a bunch of hacky code :)

suyograo:
Code LGTM. Most other integration tests failed on the last run. I kicked off a build again, we need to see whats up.

original-brownbear:
@suyograo failed again, strange I'm looking into it. Looks like I deleted some file that now messes up the teardown. Sec :)

original-brownbear:
@suyograo I think I got it, tried to be a little drier than possible (the `settings.get("input")` didn't work in non Kafka specs -> fixed now).
At least I found two small issues while I was at it :)

original-brownbear:
@suyograo all green, see https://github.com/elastic/logstash/pull/7093/commits/bb33cee7c38e71a8d2134cbe634c4fc147666a82 for changes I made now

suyograo:
@original-brownbear squash and merge away :)

suyograo:
@original-brownbear this should be backported to 5.x as well..

original-brownbear:
@suyograo got it, put on the list for tomorrow :)

PR #7096, opened 2017-05-14
Port of #7093 to 5.x

original-brownbear:
* Dockerize Kafka in ITs for #7091


original-brownbear:
@suyograo FYI, could cherry pick the commit over to `5.x` without problems. Waiting for green Travis to merge.

original-brownbear:
@suyograo this is exactly the same as the version for `master` with the exception of this line https://github.com/elastic/logstash/pull/7096/files#diff-92ddfd0e9e21be250bbae2703f81723cR120.

1. I had to move the `.lock` file to the `.lock.release` path to work as expected after #6985 
2. I had to up the version there to not conflict (my guess is that this worked up until now from the cache in Travis)

I'm not 100% that version change is correct though, we do lock `0.3.11` fairly explicitly in code like:

```ruby
module Jars
  VERSION = '0.3.11'.freeze
  JRUBY_PLUGINS_VERSION = '1.1.3'.freeze
  DEPENDENCY_PLUGIN_VERSION = '2.8'.freeze
end
```

... then again, we did have `0.3.10` in the `.lock` file so this may have an impact on customers.
I personally would suggest updating the version, since it's a minor version jump, but you probably have a better handle of how much risk there is to doing that :)

suyograo:
@original-brownbear the version bump for `jar-dependencies` makes sense. 

LGTM

PR #7099, opened 2017-05-14
#7043 run gradle tests in core specs and document running gradle tests only in readme

### LINKED ISSUE #7043: Gradle/Java Tests are not Run as Part of the Unit-Tests?
Currently the rake task for the unit tests `test:core` does not execute the Java/Gradle unit-tests.
Also, the Gradle build is not documented in the Readme at all.

Shouldn't we run the Gradle UTs as part of `test:core` automatically (or if there is reasons not to do that at least document them in the Readme)?

original-brownbear:
closes #7043

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,10 @@ Most of the unit tests in Logstash are written using [rspec](http://rspec.info/)
 
     bin/rspec
     bin/rspec spec/foo/bar_spec.rb
+    
+3- To run the subset of tests covering the Java codebase only run:
+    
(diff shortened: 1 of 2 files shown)


suyograo:
LGTM

original-brownbear:
@suyograo thanks for taking a look, looks like without persistent queues activated Java tests fail on Travis. Merge still and add issue for that? (the downside is that this prevents the ruby tests from running too, probably not what we want?)

Issue #7031, opened 2017-05-05
fails:  LogStash::Instrument::WrappedWriteClient AckedMemoryQueue pushes batch to the 


ph:
branch: logstash+5.x
Link: https://logstash-ci.elastic.co/job/elastic+logstash+5.x+dockerubuntu-periodic/1976/console

Wee seems to have a NPE exception in the write_client/read_client, the test that fails is the following

```
Failures:

  1) LogStash::Instrument::WrappedWriteClient AckedMemoryQueue pushes batch to the `WriteClient`
     Failure/Error: expect(read_client.read_batch.size).to eq(1)
       
       expected: 1
            got: 0
       
       (compared using ==)
     # ./logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb:41:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # ./rakelib/test.rake:40:in `(root)'
```

but we also saw this error in other part of the execution:

```
core/spec/logstash/instrument/wrapped_write_client_spec.rb:36" java.lang.NullPointerException
	at org.logstash.ackedqueue.Queue.write(Queue.java:317)
	at org.logstash.ackedqueue.ext.JrubyAckedQueueMemoryExtLibrary$RubyAckedMemoryQueue.ruby_write(JrubyAckedQueueMemoryExtLibrary.java:152)
	at org.logstash.ackedqueue.ext.JrubyAckedQueueMemoryExtLibrary$RubyAckedMemoryQueue$INVOKER$i$1$0$ruby_write.call(JrubyAckedQueueMemoryExtLibrary$RubyAckedMemoryQueue$INVOKER$i$1$0$ruby_write.gen)
	at org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:326)
	at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:170)
	at org.jruby.ast.CallOneArgNode.interpret(CallOneArgNode.java:57)
	at org.jruby.ast.NewlineNode.interpret(NewlineNode.java:105)
	at org.jruby.ast.BlockNode.interpret(BlockNode.java:71)
	at org.jruby.evaluator.ASTInterpreter.INTERPRET_METHOD(ASTInterpreter.java:74)
	at org.jruby.internal.runtime.methods.InterpretedMethod.call(InterpretedMethod.java:182)
[...]

Issue #7032, opened 2017-05-05
Fails: LogStash::Instrument::WrappedWriteClient AckedMemoryQueue pushes single event to the `WriteClient`      Failure/Error: expect(read_client.read_batch.size).to eq(1)


ph:
https://logstash-ci.elastic.co/job/elastic+logstash+5.x+multijob-os-compatibility/os=amazon/268/console
```
Failures:

  1) LogStash::Instrument::WrappedWriteClient AckedMemoryQueue pushes single event to the `WriteClient`
     Failure/Error: expect(read_client.read_batch.size).to eq(1)
       
       expected: 1
            got: 0
       
       (compared using ==)
     # /var/lib/jenkins/workspace/elastic+logstash+5.x+multijob-os-compatibility/os/amazon/logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb:28:in `(root)'
     # /var/lib/jenkins/workspace/elastic+logstash+5.x+multijob-os-compatibility/os/amazon/vendor/bundle/jruby/1.9/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
     # /var/lib/jenkins/workspace/elastic+logstash+5.x+multijob-os-compatibility/os/amazon/rakelib/test.rake:40:in `(root)'
     # /var/lib/jenkins/workspace/elastic+logstash+5.x+multijob-os-compatibility/os/amazon/vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # /var/lib/jenkins/workspace/elastic+logstash+5.x+multijob-os-compatibility/os/amazon/vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:250:in `execute'
     # /var/lib/jenkins/workspace/elastic+logstash+5.x+multijob-os-compatibility/os/amazon/vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:194:in `invoke_with_call_chain'
     # /var/lib/jenkins/workspace/elastic+logstash+5.x+multijob-os-compatibility/os/amazon/vendor/bundle/jruby/1.9/gems/rake-12.0.0/lib/rake/task.rb:187:in `invoke_with_call_chain'
[...]

PR #7103, opened 2017-05-15
Make the test for the write WrappedWriteClient more robust

ph:
Instead of using sleep we start two threads and we join them to do the
expectation.

Fixes: #7031 #7032

### DIFF
--- a/logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
+++ b/logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
@@ -18,39 +18,54 @@ describe LogStash::Instrument::WrappedWriteClient do
 
   subject { described_class.new(write_client, pipeline, metric, plugin) }
 
+  def threaded_read_client(read_client)
+    Thread.new(read_client) do |_read_client|
+      started_at = Time.now
+
+      batch_size = 0
+      loop {
+        if Time.now - started_at > 60
+          raise "Took too much time to read from the queue"
+        end
+        batch_size = _read_client.read_batch.size
+
+        break if batch_size > 0
+      }
+      expect(batch_size).to eq(1)
+    end
+  end
 
   shared_examples "queue tests" do
     it "pushes single event to the `WriteClient`" do
-      t = Thread.new do
-        subject.push(event)
+      pusher_thread = Thread.new(subject, event) do |_subject, _event|
+        _subject.push(event)
       end
-      sleep(0.01) while !t.status
-      expect(read_client.read_batch.size).to eq(1)
-      t.kill rescue nil
+
+      reader_thread = threaded_read_client(read_client)
+
+      [pusher_thread, reader_thread].collect(&:join)
     end
 
     it "pushes batch to the `WriteClient`" do
       batch = write_client.get_new_batch
       batch << event
 
-      t = Thread.new do
-        subject.push_batch(batch)
+      pusher_thread = Thread.new(subject, batch) do |_subject, _batch|
+        subject.push_batch(_batch)
       end
 
-      sleep(0.01) while !t.status
-      expect(read_client.read_batch.size).to eq(1)
-      t.kill rescue nil
+      reader_thread = threaded_read_client(read_client)
+      [pusher_thread, reader_thread].collect(&:join)
     end
 
     context "recorded metrics" do
       before do
-        t = Thread.new do
-          subject.push(event)
+        pusher_thread = Thread.new(subject, event) do |_subject, _event|
+          _subject.push(event)
         end
-        sleep(0.01) while !t.status
-        sleep(0.250) # make it block for some time, so duration isn't 0
-        read_client.read_batch.size
-        t.kill rescue nil
+
+        reader_thread = threaded_read_client(read_client)
+        [pusher_thread, reader_thread].collect(&:join)
       end
 
       let(:snapshot_store) { collector.snapshot_metric.metric_store }
@@ -97,7 +112,7 @@ describe LogStash::Instrument::WrappedWriteClient do
   end
 
   context "AckedMemoryQueue" do
-    let(:queue) { LogStash::Util::WrappedAckedQueue.create_memory_based("", 1024, 10, 1024) }
+    let(:queue) { LogStash::Util::WrappedAckedQueue.create_memory_based("", 1024, 10, 4096) }
 
     before do
       read_client.set_events_metric(metric.namespace([:stats, :events]))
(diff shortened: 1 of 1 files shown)


logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
```diff
@@ -97,7 +112,7 @@
   end
 
   context "AckedMemoryQueue" do
-    let(:queue) { LogStash::Util::WrappedAckedQueue.create_memory_based("", 1024, 10, 1024) }
+    let(:queue) { LogStash::Util::WrappedAckedQueue.create_memory_based("", 1024, 10, 4096) }
```
ph: See https://github.com/elastic/logstash/issues/7031#issuecomment-301283267 for reasoning.

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
```diff
@@ -18,39 +18,54 @@
 
   subject { described_class.new(write_client, pipeline, metric, plugin) }
 
+  def threaded_read_client(read_client)
+    Thread.new(read_client) do |_read_client|
```
jsvd: since the client here doesn't need to be copied or modified, I see no use of passing it as variable into the thread.

Also, I see some inconsistent use of the variables passed this way into the threads, I will point them out below

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
```diff
@@ -18,39 +18,54 @@
...
+      started_at = Time.now
+
+      batch_size = 0
+      loop {
+        if Time.now - started_at > 60
```
jsvd: cool, we could reuse this pattern in many places, even maybe add this feature to `stud/try`

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
```diff
@@ -18,39 +18,54 @@
...
+      expect(batch_size).to eq(1)
+    end
+  end
 
   shared_examples "queue tests" do
     it "pushes single event to the `WriteClient`" do
-      t = Thread.new do
-        subject.push(event)
+      pusher_thread = Thread.new(subject, event) do |_subject, _event|
+        _subject.push(event)
```
jsvd: s/event/_event

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
```diff
@@ -18,39 +18,54 @@
...
     end
 
     it "pushes batch to the `WriteClient`" do
       batch = write_client.get_new_batch
       batch << event
 
-      t = Thread.new do
-        subject.push_batch(batch)
+      pusher_thread = Thread.new(subject, batch) do |_subject, _batch|
+        subject.push_batch(_batch)
```
jsvd: s/subject/_subject

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
```diff
@@ -18,39 +18,54 @@
 
   subject { described_class.new(write_client, pipeline, metric, plugin) }
 
+  def threaded_read_client(read_client)
```
jsvd: I suggest pushing the expectation outside of this method. This makes the `threaded_read_client` behave more like its name, though I maybe would call it something like `read_until_data(read_client, time_out)`. wdyt?

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
```diff
@@ -18,39 +18,54 @@
...
+      reader_thread = threaded_read_client(read_client)
+      [pusher_thread, reader_thread].collect(&:join)
     end
 
     context "recorded metrics" do
       before do
-        t = Thread.new do
-          subject.push(event)
+        pusher_thread = Thread.new(subject, event) do |_subject, _event|
+          _subject.push(event)
```
jsvd: s/event/_event

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
ph: I wanted to make a distinction between the values defined in the let statement and the passed values in the block. I understand they will take precedence.

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
ph: That will be a bit harder, since we have to *join* on the thread to go to the next spec.
Moving the expectation out, will mean having to pass a variable as a reference to the thread and might cause thread safety issue.

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
ph: If we don't pass it we will need to make sure that `read_client` is create with `let!` to make sure we don't create lazy. `Thread.new` will  make a copy of any variable passed.

jsvd: then I'm +1 on using `let!`

ph:
updated with review comments.

logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
jsvd: strangely, I can't make that copy happen:

```ruby
require 'rspec'

RSpec.describe Thread do
  context "variable copying in Thread.new" do
    let(:hash) { Hash.new }
    context "when variable is not passed as an argument" do
      it "does not copy the value" do
        t = Thread.new { Thread.current[:var] = hash }.join
        expect(t[:var]).to be(hash)
      end
    end
    context "when variable is passed as an argument" do
      it "copies the value" do
        t = Thread.new(hash) {|h| Thread.current[:var] = h }.join
        expect(t[:var]).to_not be(hash)
      end
    end
  end
end
```

and running it:

```
% rspec thread_spec.rb
.F

Failures:

  1) Thread variable copying in Thread.new when variable is passed as an argument copies the value
     Failure/Error: expect(t[:var]).to_not be(hash)
     
       expected not #<Hash:2050> => {}
                got #<Hash:2050> => {}
     
       Compared using equal?, which compares object identity.
     
     
       Diff:
       
     # ./thread_spec.rb:15:in `(root)'

Finished in 0.418 seconds (files took 0.995 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./thread_spec.rb:13 # Thread variable copying in Thread.new when variable is passed as an argument copies the value
```

in both situations the value isn't copied, so the let is only executed once

ph: weird, It's was from the doc :)

ph:
@jsvd I saw your last comment but it was on a old commit, did this PR is ready to merge?
I've restarted the build since its failed with the rack update issue.

PR #7106, opened 2017-05-15
Deprecate `mean` metric instrument

### LINKED ISSUE #7094: `MEAN` Instrument type should be removed
We currently allow for a `mean` instrument type. This should, IMHO be removed. It cannot do its job accurately. I'm also not aware of it actually being used anywhere. It definitely isn't used in core.

The central problem with recording a `mean` stat is that on a non-realtime system you can't guarantee that samples were evenly spaced out. A series of timestamped gauges does a better job. This is what we can do in our x-pack monitoring extension, but not something we should do in core. The central difference is that x-pack has Elasticsearch, which can act as a time series database. I'd prefer not to build a tsdb into logstash. That would obviously involve a lot of complexity.

WDYT @suyograo @jsvd @ph ?

andrewvc:
This feature is removed in #7105 for 6.0+

Fixes #7094

### DIFF
--- a/Gemfile.jruby-1.9.lock.release
+++ b/Gemfile.jruby-1.9.lock.release
@@ -15,6 +15,7 @@ PATH
       minitar (~> 0.5.4)
       pry (~> 0.10.1)
       puma (~> 2.16)
+      rack (= 1.6.6)
       ruby-maven (~> 3.3.9)
       rubyzip (~> 1.1.7)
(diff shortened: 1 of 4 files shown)


ph:
I first I thought that would be noisy in the log bug since the number of metric should stabilize, it will be fine.

andrewvc:
Using the merge tool since jarvis is down.

Issue #7100, opened 2017-05-15
Add Kafka Backwards Compatibility Tests


original-brownbear:
This was discussed in https://github.com/logstash-plugins/logstash-input-kafka/issues/178#issuecomment-300804803.

We want to validate that the Kafka input still works with Kafka `0.10.0+` and not only `0.10.2.1`. So we need ITs running against the older versions too.

This should be very trivial, since it only requires rebuilding the Kafka `Dockerfile` at `qa/integration/services/kafka_dockerized/Dockerfile` with a different version set.

* Note: It's probably a good idea to move the version ENV variable below the `apt-get` installations to save some disk space and speed things up :)

original-brownbear:
@suyograo maybe you have some comments on this?

Do we want to run this on every run and also for what versions exactly. I mean looping through all of them could take a lot of resources. Maybe just have a `0.10.0` run and a run with whatever the latest version is (currently `0.10.2.1`)?

suyograo:
> Maybe just have a 0.10.0 run and a run with whatever the latest version is (currently 0.10.2.1)?
 
+1 to this. we should have just 2 entries or so in this compat matrix. We don't want to test for all Kafka versions :)

jordansissel:
We no longer list compatibility in the documentation for the sake of simplification as well as accuracy. I don't think we need to do this compatibility test anymore. 

Reference: https://github.com/logstash-plugins/logstash-output-kafka/issues/156

PR #7107, opened 2017-05-15
Use the headless JRE for the kafka docker image

andrewvc:
This should help the image build a bit faster since it won't have to pull all of X11 in.


andrewvc:
I would almost say we should go for the JRE, but that's hard to debug if you have to.

original-brownbear:
@andrewvc I'm not so sure we'd ever really debug here? I'm good with JRE too and use the JDK on special occasions manually. Especially considering #7100 would add more pressure on disk space anyway + JRE is more real world (not that either really matters, but it's cleaner) :D

andrewvc:
@original-brownbear that's a good point. I've updated it to use the JRE.

original-brownbear:
@andrewvc LGTM :)

Issue #7089, opened 2017-05-12
javadoc gradle task is failing due to "error: no tag name after"


jsvd:
This failure to compile the javadoc is causing some build failures.

this is easy to reproduce:

```
~/projects/logstash (git)-[master] % gradle javadoc
:logstash-core:compileJava UP-TO-DATE
:logstash-core:processResources UP-TO-DATE
:logstash-core:classes UP-TO-DATE
:logstash-core:javadoc
/Users/joaoduarte/projects/logstash/logstash-core/src/main/java/org/logstash/common/DeadLetterQueueFactory.java:47: error: no tag name after @
     * Retrieves an existing {@link DeadLetterQueueWriter} associated with {@param id}, or
                                                                           ^
/Users/joaoduarte/projects/logstash/logstash-core/src/main/java/org/logstash/common/DeadLetterQueueFactory.java:53: error: no tag name after @
     *                for each {@param id}
                               ^
/Users/joaoduarte/projects/logstash/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java:58: warning: no description for @param
     * @param queuePath
       ^
/Users/joaoduarte/projects/logstash/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java:59: warning: no description for @param
     * @param maxSegmentSize
       ^
/Users/joaoduarte/projects/logstash/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java:60: warning: no description for @throws
     * @throws IOException
       ^
/Users/joaoduarte/projects/logstash/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java:62: warning: no @param for maxQueueSize
    public DeadLetterQueueWriter(Path queuePath, long maxSegmentSize, long maxQueueSize) throws IOException {
           ^
/Users/joaoduarte/projects/logstash/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java:96: warning: no description for @throws
     * @throws IOException
       ^
/Users/joaoduarte/projects/logstash/logstash-core/src/main/java/org/logstash/common/io/RecordIOReader.java:146: warning: no description for @return
     * @return
       ^
/Users/joaoduarte/projects/logstash/logstash-core/src/main/java/org/logstash/common/io/RecordIOReader.java:211: warning: no description for @return
     * @return
       ^
[...]

PR #7108, opened 2017-05-15
Fix gradle assemble

andrewvc:
This makes sure that `gradle assemble` is run with `compile:all` which is part of the CI suite. This is important as it helps to catch javadoc failures. This task was broken as the `javadoc` task had some small issues that have also been resolved here.

https://github.com/elastic/logstash/issues/7089


suyograo:
LGTM

andrewvc:
This is stuck on unrelated test failures in #7113 . I'm going to merge it regardless since this is breaking all x-pack builds.

Issue #7097, opened 2017-05-14
Dockerize ElasticSearch for ITs


original-brownbear:
It's in the title, move the setup and teardown of the integration test ES service to Docker like it was done in #7093 for Kafka.

wainersm:
Hi @suyograo  and @original-brownbear , if there is any other item about this dockerization work, let me know. I can help on this.

@original-brownbear : I am curious why you building the docker images with the services (e.g. installing kafka) instead of pull already built images.

suyograo:
@wainersm all yours now :)

wainersm:
Thanks for the gift @suyograo :)

original-brownbear:
@wainersm thanks for the help!

As far as your question goes:

> I am curious why you building the docker images with the services (e.g. installing kafka) instead of pull already built images.

Two reasons:

* We have full control of the environment, while with existing Docker image we always are at the mercy of whoever creates them. (99.5% percent of CI runs will probably work just fine, but whenever someone makes a change that breaks our scripting it costs a lot of time to find the problem and disables the tests for x amount of hours)

More importantly:

* We're running tests on CI and it's kind of hard to cache Dockerfiles on Travis and even on your own infrastructure proxying Images is a serious pain in my experience (though that may have improved lately), but especially with Travis we want to keep things fast and not too disk space heavy. If we use our own Docker files we can just put the download of the actual service dependency at the bottom and share all those layers installing java and such :)

BTW, regarding the second point, feel free to "dry up" your `Dockerfile` against the one I have added for Kafka by moving lines in the Kafka one around. If you do something like

turning:

```docker
FROM debian:stretch

ENV KAFKA_HOME /kafka
ENV KAFKA_LOGS_DIR="/kafka-logs"
ENV KAFKA_VERSION 0.10.2.1
ENV _JAVA_OPTIONS "-Djava.net.preferIPv4Stack=true"
ENV TERM=linux

RUN apt-get update && apt-get install -y curl openjdk-8-jre-headless netcat

RUN mkdir -p ${KAFKA_LOGS_DIR} && mkdir -p ${KAFKA_HOME} && curl -s -o $INSTALL_DIR/kafka.tgz \
    "http://ftp.wayne.edu/apache/kafka/${KAFKA_VERSION}/kafka_2.11-${KAFKA_VERSION}.tgz" && \
    tar xzf ${INSTALL_DIR}/kafka.tgz -C ${KAFKA_HOME} --strip-components 1
    
[...]

PR #7063, opened 2017-05-10
MINOR: Resource safety and redundant matcher cleanup in queue tests

original-brownbear:
Just small points I noticed going over these tests for other issues:

* The `Queue` was not safely closed on test failure
* Some dead variables
* Redundant matchers like `is(equalTo(x))`

PR #7114, opened 2017-05-15
Fix broken Gradle tests

original-brownbear:
Let's see if this works on Travis also. Fine on my local with this fix.


original-brownbear:
@andrewvc @suyograo @colinsurprenant 

So the issue is this: `org.logstash.ackedqueue.QueueTest#ackingMakesQueueNotFullAgainTest` is really unstable.
If it fails it made all other tests fail as predicted in #7063 (childish brag I know ...) since the `.lock` file is never cleaned up.
Fixed this by making the `Queue` run from a random directory every time.
Still that test is unstable. I could make it fail in 100% of cases by adding an additional sleep of 10ms here: https://github.com/elastic/logstash/pull/7114/files#diff-7e021bd7a52783fd6050c27dc64fbfd5R496 (don't mind the additional resource safety added to the test, it has no functional impact and is just part of me narrowing things down)

I don't understand that test tbh. It seems like the closing of the batch `b` makes the write complete and the comment says the write should complete as soon as the batch is closed and hence acknowledged. Seems to me exactly that is happening and the test just passed in the past because the `write` took long enough to complete for the `future.isDone()` assertion to still equal `false`?

andrewvc:
@original-brownbear I'm going to defer to the original author of that test to clarify here. I believe that's probably either @colinsurprenant or @guyboertje . Thanks for the amazing work here :)

original-brownbear:
@andrewvc np :)

The original author here is @jsvd who actually made the test work the way this PR makes it work again. It was then changed to the broken version in a PR by @colinsurprenant in what looks like a merge accident of sorts (at least to me and @jordansissel).

Maybe you can give this the final review to fix tests @jordansissel ? :)

colinsurprenant:
Thanks for all the works done here - I wish I could have helped sooner. 

2 things:

1. Let me first explain the test logic (and yes, there is an error in the final expectation). The idea of the test is to **a)** setup the queue so that the write wrapped in the future will block. The way the write operation works is to always first write the data then block if the queue is in a full condition. then we want to **b)** read and close the batch (because the close operation does the acking, thus the purging of fully acked pages) then finally we want to **c)** assert that the write operation unblocked and completed. 

So the correct fix here is simply to change line 505 from 
```java
assertThat(future.isDone(), is(false));
```
to 
```java
assertThat(future.get(), is(equalTo(ELEMENT_COUNT + 1L)));
```

2. I don't think we should actually do the proposed temporary folder modification. This will actually hide future resource cleanup problems. This was exactly why I decided to **keep** the locking logic in place even with the memory acked queue. Like here: if we see locking errors then we must fix the problem causing it and not try to circumvent the problem by making it impossible to happen.  This strategy was essential in nailing and fixing the pipeline initialization/restart/shutdown problems.

colinsurprenant:
@original-brownbear @jordansissel @jsvd - also, it wasn't a merge accident, these refactors and new tests were part of modifications related to a decision to change the behaviour of full/not full signaling. the last assertion error was an oversight and did not fail systematically because of timing considerations. It is very hard to predict is it will pass or fail on that (wrong) final assertion. The correct fix is just to wait for that future to complete.

original-brownbear:
@colinsurprenant 2 points back for tomorrow :)

On point 1 I agree, though resource cleanup in the test should happen regardless of it failing, so I'd keep that change.
The `try-with-resources` on the batch is more a leftover of my experiments. We can revert that noise I guess :) Will do.
But the safe closing of the queue and joining of the `ExecutorService` we could and should keep right?

On point 2 I would like to agree, but don't :(
I agree, we get some free testing on resource safety and using the `try-with-resources` fix (state after my first commit here) I could actually fix the leaking of `.lock files` so in a sense things worked out according to your plan :)
The downside was visible too though. A bug in one test actually caused a bunch of other tests to fail and way increased the time this took to debug. Moreover:
* This prevents parallel test runs from ever being possible (not a big one now that we have few Java tests ... but eventually having a set of non parallelizable tests becomes very expensive if they involve any sort of sleeps or timeouts)
* You could actually have your tests start failing from an unclean working directory and not even have `gradle clean` get you out of this mess, making build behaviour somewhat unpredictable. I think `git add -A && git reset --hard && ./gradlew clean test` should have 100% deterministic behaviour, otherwise you start hunting weird things like a leftover `.lock` a lot eventually (imo)

=> maybe instead of making our tests run all from the same folder we could add an issue to add a test suit that tests actual Queue exception states for proper cleanup?

original-brownbear:
@colinsurprenant hmm after being really annoyed by #7087 which looks like it may well have similar origins, leaving behind a `.lock` and only failing when running the full suit I'm starting to think that:

* Maybe we should in fact run the Queue from a separate temp dir for each test, but make sure that the `.lock` is removed from those temp dirs after the test for every test. A wrapper for `org.junit.rules.TemporaryFolder` comes to mind :)

That gives us the same coverage as running it all from one dir, plus we get precision in tracking things down :)

colinsurprenant:
@original-brownbear a "failed/broken" test run leaving a `.lock` file is not necessarily a problem. A `.lock` file that is not held by any active JVM process for exclusive access is not "valid" and will not prevent a new JVM process from acquiring an exclusive lock.

original-brownbear:
@colinsurprenant true, though shouldn't the only case in which we leave a `.lock` behind be the JVM dying. In all other cases we should be able to handle the lock safely like any other JVM resource right?

colinsurprenant:
I would like us to avoid trying to tackle too many issues in a same PR and I think we should separate the proposed changes here. I suggest that this PR actually only fix the broken test, then we can tackled the other concerns in other PRs.

original-brownbear:
@colinsurprenant sure that would be the next step, I'm just trying to save the random path per test run here, since this was a real pain to track down :D

colinsurprenant:
actually, I'd prefer we defer that to another PR since this is not the cause of the failing test.

original-brownbear:
@colinsurprenant it did cause all tests to fail as the result of one failing ? :)

colinsurprenant:
I get this but the problem was the one failing test, two separate concerns.

original-brownbear:
@colinsurprenant hmpf ok :) sec reverting the path changes then.

original-brownbear:
@colinsurprenant reverted all but the fixes to that one test :)

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -473,38 +472,37 @@ public void ackingMakesQueueNotFullAgainTest() throws IOException, InterruptedEx
...
-        assertThat(q.isFull(), is(false));
-
-        // will not complete because write will not unblock until the page is purge with a batch close/acking.
-        assertThat(future.isDone(), is(false));
-
-        q.close();
+        try (TestQueue q = new TestQueue(settings)) {
+            q.open();
+            int ELEMENT_COUNT =
+                90; // should be able to write 90 events (9 pages) before getting full
```
colinsurprenant: style: I find this really ackward to read and woud prefer either have the comment appended at the end or put on top of `int ELEMENT_COUNT = 90;` if having too long lines really bothers you.

original-brownbear: reverted :)

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -473,38 +472,37 @@ public void ackingMakesQueueNotFullAgainTest() throws IOException, InterruptedEx
...
+            for (int i = 0; i < ELEMENT_COUNT; i++) {
+                q.write(element);
+            }
+            assertThat(q.isFull(), is(false));
+            // we expect this next write call to block so let's wrap it in a Future
+            Callable<Long> write = () -> q.write(element);
+            final long seq = q.nextSeqNum();
```
colinsurprenant: I don't understand why this is introduced here. By calling `nextSeqNum()` here you are mutating the queue state in a completely unrelated way and it is not required since we know what is to be expected for the seqNum later on.

original-brownbear: right reverting sec

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -473,38 +472,37 @@ public void ackingMakesQueueNotFullAgainTest() throws IOException, InterruptedEx
...
+            Future<Long> future = executor.submit(write);
+            assertThat(future.isDone(), is(false));
+            while (!q.isFull()) {
+                Thread.sleep(10);
+            }
+            assertThat(q.isFull(), is(true));
+            try (Batch b = q.readBatch(10)) {
```
colinsurprenant: I would prefer we keep the batch read and close separate and explicit. it is a somewhat different context than proper resource cleanups.

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -473,38 +472,37 @@ public void ackingMakesQueueNotFullAgainTest() throws IOException, InterruptedEx
...
+                assertThat(b.size(), is(10));  // purge 1 page by closing b
+            }
+            Thread.sleep(10L);
```
colinsurprenant: why is this needed here?

original-brownbear: this made the issue visible in 100% of runs, not necessary anymore -> reverting

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
original-brownbear: reverting

original-brownbear:
@colinsurprenant all fixed

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -473,38 +472,37 @@ public void ackingMakesQueueNotFullAgainTest() throws IOException, InterruptedEx
...
+            while (q.isFull()) {
+                Thread.sleep(10);
+            }
+            assertThat(q.isFull(), is(false));
+            assertThat(future.get(), is(seq + 1));
+        } finally {
+            executor.shutdownNow();
```
colinsurprenant: I believe `shutdown()` should be used here and not `shutdownNow()`. we don't need to force kill anything, juste wait for proper completion.

original-brownbear: @colinsurprenant we should force kill to not leak on failing tests. This has no effect for passing tests anyways since the one `Callable` in the `Executor` has finished at this point.

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -473,38 +472,37 @@ public void ackingMakesQueueNotFullAgainTest() throws IOException, InterruptedEx
...
+            executor.awaitTermination(2L, TimeUnit.MINUTES);
```
colinsurprenant: - 2 minutes timeout? this is rather useless since the test itself has a shorter timeout `@Test(timeout = 5000)`.
- also `awaitTermination ` returns «true if this executor terminated and false if the timeout elapsed before termination» so it seems rather useless to just awaitTermination without handling the `false` case? we should at least assert that `awaitTermination` returns true and if we see a case where that does not happen we investigate?

original-brownbear: @colinsurprenant actually since we have a timeout on the test we can just wait on the executor without timeout and it will throw if it doesn't join :) Sec fixing

original-brownbear: @colinsurprenant I think now it's more standards compliant with an infinite wait :)

colinsurprenant:
@original-brownbear this is a styling and personal preference comment but I would have liked if the overall style was somewhat maintained with the spaces in between the logical test blocks or steps as I think it make the test easier to read. Also, I liked the 
```java
while (!q.isFull()) { Thread.sleep(10); }
```
one liner constructs as, again personal preference, it make is more concise and easier to read. 

I think that generally speaking we should minimally try to respect to original code style unless we agree on a overall style refactor or one is obviously needed.  It also creates more unnecessary changes and makes it harder to review.

original-brownbear:
@colinsurprenant sure sec

original-brownbear:
@colinsurprenant spaces added back and lines joined

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
colinsurprenant: I don't think a agree here. If for any reason the shutdown does not complete, I think we should know and not just force kill/clean everything and discard.

original-brownbear: @colinsurprenant it will die with `InterruptedException` in that case

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -473,38 +472,37 @@ public void ackingMakesQueueNotFullAgainTest() throws IOException, InterruptedEx
...
+            Batch b = q.readBatch(10); // read 1 page (10 events)
+            b.close();  // purge 1 page
+            
+            while (!q.isFull()) { Thread.sleep(10); }
+            assertThat(q.isFull(), is(false));
+            
+            assertThat(future.get(), is(ELEMENT_COUNT + 1));
+        } finally {
+            executor.shutdownNow();
+            executor.awaitTermination(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
```
colinsurprenant: So we have a choice here: either we choose a value that is longer than the overall test timeout and rely on the test timeout to maybe signal am executor problem or we choose a smaller value and explicitly handle the executor termination conditions.

original-brownbear: @colinsurprenant I guess some, but this is such a trivial case (just 1 extra thread) imo it's not worth using any non standard pattern. We should just try to improve test isolation in these cases imo.

colinsurprenant: what is the standard pattern?

original-brownbear: @colinsurprenant just wait forever since someone else (JUnit) is gonna come along to interrupt. Least amount of code and since we have one thread there is zero ambiguity on what caused a potential `InterruptedException`

colinsurprenant: in other words, I don't see how this, for example, would be not standard?
```java
executor.shutdownNow();
assertThat(executor.awaitTermination(2, TimeUnit.SECONDS), is(true));
```

colinsurprenant: and it feels more explicit, no?

original-brownbear: It's strange to assert in a `final` block since you'd run this assertion for a failing test too getting two failed assertions. Never seen that anywhere and Idea highlights it even

colinsurprenant: good point. +1

original-brownbear:
@colinsurprenant gnarf now we broke it again with some change ... sec

original-brownbear:
@colinsurprenant fixed

colinsurprenant:
Another styling comment: I see the import statements have been shuffled around, would it be possible to avoid unnecessary shuffle? it also creates unnecessary changes to review.

original-brownbear:
@colinsurprenant kind of hard to do since we're adding a new import on `TimeUnit` and the existing imports are completely unordered. I think it's fine to take the opportunity to clean this up instead of just adding the import of `TimeUnit` "somewhere" right?

colinsurprenant:
@original-brownbear «the existing imports are completely unordered»? not sure I see that?

original-brownbear:
@colinsurprenant I guess the order is just a little strange let me try an patch this in sec :D

original-brownbear:
@colinsurprenant imports reordered

colinsurprenant:
it was mainly a "styling comment" and a recommendation for the future to try and avoid unnecessary import shuffle (or unrelated code changes). 

In general, we agreed as a team, to try to minimize changes that are unrelated to the fix. This does not mean that you should never proceed with some code reorg when it makes sense, you can judge that (IMO here «completely unordered» imports seems exagerated). But when doing so, you delegate the burden of verifying these changes to the reviewer, in the context of something unrelated. So it's better to submit another PR with "cosmetic" refactors in that respect.

original-brownbear:
@colinsurprenant sure makes sense :)

colinsurprenant:
Passes locally, LGTM pending green build!

original-brownbear:
@colinsurprenant it's green, ITs are just red from the GeoIP issue

colinsurprenant:
All right - there seems to also have other failures related to monitoring - let's make sure these are captured if they don't already?
:shipit:

original-brownbear:
@colinsurprenant phew let me see

original-brownbear:
@colinsurprenant urgh almost looks like a dep. issue is breaking those ITs :(

colinsurprenant:
but is this related to this change?

original-brownbear:
@colinsurprenant no logically impossible, this change only changes Java UT code, those changes can't leak into the JRuby ITs.

original-brownbear:
@colinsurprenant looks to me like the other issues are ES + GeoIP related. Es goes to Docker in #7097 which will fix (or change) whatever random env behaviour we have here.
The geoip thing is tracked.

Good at last ?

colinsurprenant:
Yes yes - please go ahead and merge! 😝 

We don't have to deal with these here, I was just pointing out that if these are unrelated we just need to make sure we go back and capture these errors and create separate issues if none already exist for these failures. 

Also, please do not forget to create separate PRs for the other changes originally proposed here, these were good ideas worth discussing 😇

original-brownbear:
@colinsurprenant alright 🎉  merging

original-brownbear:
@colinsurprenant maybe we should get at least #7053 in before we go for further steps in refactoring/fixing test isolation. It's already conflicts all over the place with my PQ issues? I'll add issues for outstanding things later though so it's not forgotten.

PR #7117, opened 2017-05-15
Add test case for FileLockFactory

### LINKED ISSUE #6610: move bin/lock into an integration test
`bin/lock` was introduce with the `FileLockFactory` class to manually test cross-process/JVM locking. This should probably be moved into a proper & automated integration test.

relates to #6607

wainersm:
This change introduce a test for cross-process/jvm file locking
using the FileLockFactory class.

Fixes #6610

### DIFF
--- a/logstash-core/src/test/java/org/logstash/FileLockFactoryMain.java
+++ b/logstash-core/src/test/java/org/logstash/FileLockFactoryMain.java
@@ -1,0 +1,23 @@
+package org.logstash;
+
+import java.io.IOException;
+
+/*
+ * This program is used to test the FileLockFactory in cross-process/JVM.
+ */
+public class FileLockFactoryMain {
+
+    public static void main(String[] args) {
+        try {
+            FileLockFactory.getDefault().obtainLock(args[0], args[1]);
+            System.out.println("File locked");
+            // Sleep enough time until this process is killed.
+            Thread.sleep(Long.MAX_VALUE);
+        } catch (InterruptedException e) {
+            // This process is killed. Do nothing.
+        } catch (IOException e) {
+            // Failed to obtain the lock.
+            System.exit(1);
+        }
+    }
+}
--- a/logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
+++ b/logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
@@ -1,16 +1,24 @@
 package org.logstash;
 
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
 import static org.junit.Assert.fail;
+import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.channels.FileLock;
 import java.nio.file.FileSystems;
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
@@ -24,9 +32,12 @@ public class FileLockFactoryTest {
 
     private FileLock lock;
 
+    private ExecutorService executor;
+
     @Before
     public void setUp() throws Exception {
         lockDir = temporaryFolder.newFolder("lock").getPath();
+        executor = Executors.newSingleThreadExecutor();
     }
 
     @Before
@@ -34,6 +45,14 @@ public class FileLockFactoryTest {
         lock = FileLockFactory.getDefault().obtainLock(lockDir, LOCK_FILE);
         assertThat(lock.isValid(), is(equalTo(true)));
         assertThat(lock.isShared(), is(equalTo(false)));
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        executor.shutdownNow();
+        if (!executor.awaitTermination(2L, TimeUnit.MINUTES)) {
+            throw new IllegalStateException("Failed to shut down Executor");
+        }
     }
 
     @Test
@@ -87,5 +106,46 @@ public class FileLockFactoryTest {
     public void ReleaseUnobtainedLock() throws IOException {
         FileLockFactory.getDefault().releaseLock(lock);
         FileLockFactory.getDefault().releaseLock(lock);
+    }
+
+    @Test
+    public void crossJvmObtainLockOnLocked() throws Exception {
+        Process p = null;
+        String lockFile = ".testCrossJvm";
+        FileLock lock = null;
+
+        // Build the command to spawn a children JVM.
+        String[] cmd = {
+            Paths.get(System.getProperty("java.home"), "bin", "java").toString(),
+            "-cp", System.getProperty("java.class.path"),
+            Class.forName("org.logstash.FileLockFactoryMain").getName(),
+            lockDir, lockFile
+        };
+
+        try {
+            // Start the children program that will lock the file.
+            p = new ProcessBuilder(cmd).start();
+            InputStream is = p.getInputStream();
+            /* Wait the children program write to stdout, meaning the file
+             * is locked. Set a timeout to ensure it returns.
+             */
+            Future<Integer> future = executor.submit(() -> {return is.read();});
+            assertTrue(future.get(30, TimeUnit.SECONDS) > -1);
+
+            // Check the children process is still running.
+            assertThat(p.isAlive(), is(equalTo(true)));
+
+            try {
+                // Try to obtain the lock held by the children process.
+                FileLockFactory.getDefault().obtainLock(lockDir, lockFile);
+                fail("Should have threw an exception");
+            } catch (LockException e) {
+                // Expected exception as the file is already locked.
+            }
+        } finally {
+            if (p != null) {
+                p.destroy();
+            }
+        }
     }
 }
--- a/bin/lock
+++ b/bin/lock
@@ -1,9 +1,0 @@
-#!/usr/bin/env bin/ruby
-
-require_relative "../lib/bootstrap/environment"
-LogStash::Bundler.setup!({:without => [:build, :development]})
-require "logstash-core"
-
-lock = Java::OrgLogstash::FileLockFactory.getDefault.obtainLock(ARGV[0], ".lock")
-puts("locking " + File.join(ARGV[0], ".lock"))
-sleep
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -52,7 +52,6 @@ namespace "artifact" do
     @exclude_paths << "bin/bundle"
     @exclude_paths << "bin/rspec"
     @exclude_paths << "bin/rspec.bat"
-    @exclude_paths << "bin/lock"
 
     @exclude_paths
   end


suyograo:
@original-brownbear can you review this?

original-brownbear:
@suyograo sure :)

logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
```diff
@@ -88,4 +90,52 @@ public void ReleaseUnobtainedLock() throws IOException {
...
+            // Start the children program that will lock the file.
+            ProcessBuilder pb = new ProcessBuilder(cmd);
+            p = pb.start();
+            InputStream is = p.getInputStream();
+            long endTime = System.currentTimeMillis() + timeout;
+            long now;
+            do {
+                // Wait on children program to write something to stdout.
+                now = System.currentTimeMillis();
+            } while (is.available() == 0 && now < endTime);
```
original-brownbear: @wainersm `.available()` is a really unstable check here I fear. Maybe a construct like:

```
 try {
   while(p.read() >= 0) {
      now = System.currentTimeMillis();
   }
} catch (final EOFException {
}

Would be more stable across platforms?

File: logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java [hunk 2]
```diff
@@ -88,4 +90,52 @@ public void ReleaseUnobtainedLock() throws IOException {
...
+        String[] cmd = {
+            System.getProperty("java.home") + fs + "bin" + fs + "java",
+            "-cp", System.getProperty("java.class.path"),
+            "org.logstash.FileLockFactoryMain",
+            lockDir, lockFile
+        };
+
+        try {
+            // Start the children program that will lock the file.
+            ProcessBuilder pb = new ProcessBuilder(cmd);
```
original-brownbear: @wainersm we can probably just inline `pb` as 

```java
p = new ProcessBuilder(cmd).start();
```

right?

suyograo:
> any opinion on whether or not top keep around bin/lock as ruby script? :)

I would keep the ruby script and make it one liner calling the Java program as @original-brownbear suggested.

wainersm:
I will send a new version of the patch to address @original-brownbear 's review (very good comments!). IMO we should delete the script, don't see value on making manual tests with it once integration test will cover the case.

logstash-core/src/test/java/org/logstash/FileLockFactoryMain.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public class FileLockFactoryMain {
+
+    public static void main(String[] args) {
+        // Timeout to ensure the process finishes.
+        long timeout = 20000;
+        try {
+            FileLockFactory.getDefault().obtainLock(args[0], args[1]);
+            System.out.println("File locked");
+            Thread.sleep(timeout);
```
wainersm: That's true, it will  reduce usability from the CLI. But I'm setting a timeout to avoid the CI system of creating zombie processes, for the case something goes wrong in JUnit side and this process isn't destroyed. Notice the way I spawn this process it is detached from the JUnit process.
Well, didn't think about manual tests when implementing this testcase. :)

logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
wainersm: I'm using a very silly method for this test to check if the main application has already locked the file: it waits for any byte to appear in the input stream. I am worried about using p.read(), which will block the thread, and the process get suspended.

original-brownbear: @wainersm how about `@Test(timeout=30_000)` to get around that? :)

logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
```diff
@@ -88,4 +91,50 @@ public void ReleaseUnobtainedLock() throws IOException {
...
+            try {
+                // Try to obtain the lock held by the children process.
+                FileLockFactory.getDefault().obtainLock(lockDir, lockFile);
+                fail("Should have threw an exception");
+            } catch (LockException e) {
+                // Expected exception as the file is already locked.
+            }
+        } finally {
+            if (p != null) {
+                p.destroy();
```
original-brownbear: @wainersm I admit I'm not an expert on the process API but maybe we can make this a little safer using `destroyForcibly` instead of only `destroy`?
In this case the only way we really get a "zombie" would be the initial JVM dying I guess?

logstash-core/src/test/java/org/logstash/FileLockFactoryMain.java
original-brownbear: @wainersm added a suggestion about safety below, I agree though Java sub-processes are never optimal :( But I think the JVM just dying without the finally block being hit is unlikely enough to not care about here :)

wainersm: Agreed, I will use the Long.MAX_VALUE to sleep enough time. But see my comment below about @Test(timeout=<value>).

logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
wainersm: I checked here that @Test(timeout=<value>) creates a zombie process. It happens that the test method will be executed in a separate thread, which when terminated due timeout does not execute the finally block. Another case where finally block is never called is when the JVM suddenly dies and it is executing the try block.

So I will keep the timeout inside the test method. Gonna send a new patch, let's see if we have a least common denominator. :)

logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
wainersm: ```destroy``` will forcibly kill the sub-process. And yes, the zombie process can still happen if JVM dies. In this case we rely on Long.MAX_VALUE for the process to terminate, I didn't do the maths but lets hope it will be less than one century :)

logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
original-brownbear: @wainersm wait, are you sure about this? `finally {}` should run on even `ThreadDeath` being thrown. But doesn't JUnit simply interrupt and try to join the thread?
In that case `finally {}` will be run for sure or am I missing something? :)

logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
original-brownbear: @wainersm yea just looked into it, for our cases `destroy` does enough, it's more like 300 centuries though :P

logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
wainersm: Yep. I made following experiment:

1) My current patch, controlling timeout myself:
```
wainersm@wainersm-laptop:~/projects/oss/logstash$ ps aux | grep java
wainersm  8472  0.0  0.0  14228   924 pts/2    S+   11:35   0:00 grep --color=auto java
wainersm@wainersm-laptop:~/projects/oss/logstash$ gradle test
:logstash-core:compileJava UP-TO-DATE
:logstash-core:processResources UP-TO-DATE
:logstash-core:classes UP-TO-DATE
:logstash-core:compileTestJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
:logstash-core:processTestResources UP-TO-DATE
:logstash-core:testClasses
:logstash-core:test

org.logstash.FileLockFactoryTest > crossJvmObtainLockOnLocked FAILED
    java.lang.AssertionError at FileLockFactoryTest.java:123

277 tests completed, 1 failed
:logstash-core:test FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':logstash-core:test'.
> There were failing tests. See the report at: file:///home/wainersm/projects/oss/logstash/logstash-core/build/reports/tests/index.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 48.887 secs
wainersm@wainersm-laptop:~/projects/oss/logstash$ ps aux | grep java
wainersm 10586  0.0  0.0  14228  1012 pts/2    R+   11:36   0:00 grep --color=auto java
```
2) Now using JUnit's timeout mechanism:

```
wainersm@wainersm-laptop:~/projects/oss/logstash$ ps aux | grep java
wainersm 13019  0.0  0.0  14228  1028 pts/2    S+   11:44   0:00 grep --color=auto java
wainersm@wainersm-laptop:~/projects/oss/logstash$ gradle test
:logstash-core:compileJava UP-TO-DATE
:logstash-core:processResources UP-TO-DATE
:logstash-core:classes UP-TO-DATE
:logstash-core:compileTestJava UP-TO-DATE
:logstash-core:processTestResources UP-TO-DATE
:logstash-core:testClasses UP-TO-DATE
:logstash-core:test

org.logstash.FileLockFactoryTest > crossJvmObtainLockOnLocked FAILED
    org.junit.runners.model.TestTimedOutException at FileLockFactoryTest.java:114

277 tests completed, 1 failed
:logstash-core:test FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':logstash-core:test'.
> There were failing tests. See the report at: file:///home/wainersm/projects/oss/logstash/logstash-core/build/reports/tests/index.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 45.458 secs
wainersm@wainersm-laptop:~/projects/oss/logstash$ ps aux | grep java | head
wainersm 15109  0.2  0.3 4417624 30368 pts/2   Sl   11:44   0:00 /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java -cp /home/wainersm/projects/oss/logstash/logstash-core/build/classes/test:/home/wainersm/projects/oss/logstash/logstash-core/build/resources/test:/home/wainersm/projects/oss/logstash/logstash-core/build/classes/main:/home/wainersm/projects/oss/logstash/logstash-core/build/resources/main:/home/wainersm/.gradle/caches/modules-2/files-2.1/org.apache.logging.log4j/log4j-api/2.6.2/bd1b74a5d170686362091c7cf596bbc3adf5c09b/log4j-api-2.6.2.jar:/home/wainersm/.gradle/caches/modules-2/files-2.1/org.apache.logging.log4j/log4j-api/2.6.2/95ec045d2a3960bb2ed6953282b7886ec89d020b/log4j-api-2.6.2-tests.jar:/home/wainersm/.gradle/caches/modules-2/files-2.1/org.apache.logging.log4j/log4j-core/2.6.2/a91369f655eb1639c6aece5c5eb5108db18306/log4j-core-2.6.2
```

original-brownbear: @wainersm thanks for trying that out!, that's pretty interesting :)

That said, can't we simply do this wait by running a callable/runnable `() -> p.read()` in a single threaded executor and calling `get(...some timeout...)` on the future ?
That should be safe right?

original-brownbear: @wainersm you know my worry is just that given how 'lucky' these things go that we'll run into the environment in which `available()` is not working as expected sooner or later, `available()` is always kind of a hack :)

wainersm: Last update uses callable + future + timeout to read the stdin using p.read().

original-brownbear: @wainersm thanks! Will look into this in a few hours from now!

logstash-core/src/test/java/org/logstash/FileLockFactoryTest.java
```diff
@@ -88,4 +95,50 @@ public void ReleaseUnobtainedLock() throws IOException {
...
+            }
+            if (executor != null) {
+                executor.shutdownNow();
```
original-brownbear: @wainersm I think we should use this (https://github.com/elastic/logstash/blob/master/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java#L47) patter here too.

The simple reason being that `shutdownNow()` doesn't block and I think it's generally kind of dirty to throw in a `finally` block (makes things really confusing, in this case it would mean that we have to be on the lookout that `p.destroy` never throws for example). I think destroying executors in a safe way in `@After` is cleaner here and 100% prevents threads leaking from one thread to the next (I recommend running the Apache Spark testsuit locally if you want to observe an example of how nasty this can become :D ).

original-brownbear:
@wainersm nice I like it, just one point on handling the `Executor`, just to make things a little safer and get them in line with other tests :)

Thanks!

wainersm:
Hi @original-brownbear ! My last change introduce the safe executor shutdown. Thanks for the points, very interesting.

original-brownbear:
@wainersm 

I'm happy with this. thanks so much for taking the time to work through my points, happy you got something out of it :)

@suyograo merge to `master` and `5.x` right?

original-brownbear:
@wainersm merged, thanks again!

Issue #7039, opened 2017-05-06
`org.logstash.ackedqueue.Queueable` Interface Should be Rethought?


original-brownbear:
The current `Queuable` interface looks like it should be refactored:

We currently have:

```java
    
    byte[] serialize() throws IOException;

    static Object deserialize(byte[] bytes)
```

The serialize method is problematic because it forces exposing internal state or allocating a new `byte[]` for every write.

Since we seem to be using the resultant `byte[]` exclusively via putting them into `ByteBuffer` it seems like:

```java
 void writeTo(ByteBuffer buffer) throws IOException;
```

would be a much faster approach, not requiring any allocation of a `byte[]`.

Same goes for:

```java
static Object deserialize(byte[] bytes)
```

It looks like you'd rather want something along the lines of:

```java
Object readFrom(ByteBuffer buffer)
```

This would allow you to simply deserialize from underlying `ByteBuffer` or MMFiles by setting limit and position accordingly and remove the need for very expensive things like this:
`org.logstash.ackedqueue.io.wip.MemoryPageIOStream#read(int)`

```java
    private SequencedList<byte[]> read(int limit) throws IOException {
        List<byte[]> elements = new ArrayList<>();
        List<Long> seqNums = new ArrayList<>();

        int upto = available(limit);
        for (int i = 0; i < upto; i++) {
            long seqNum = readSeqNum();
            byte[] data = readData();
            skipChecksum();
            elements.add(data);
            seqNums.add(seqNum);
        }
        return new SequencedList<>(elements, seqNums);
    }
```

Also and especially under high load, this approach seems like it could lead to very problematic behavior from GC issues.

original-brownbear:
@colinsurprenant @suyograo ping whenever you have a sec :) (just a suggestion :))

jordansissel:
I like the proposal to avoid creating short-lived objects.

original-brownbear:
I could create a prototype for this if there's interest ? :)

suyograo:
@original-brownbear please go ahead

original-brownbear:
@andrewvc assigning you here since you're on this now effectively I guess :)

Issue #6961, opened 2017-04-25
Get rid of Java internal API in org.logstash.ackedqueue.io.MmapPageIO


wainersm:
The org.logstash.ackedqueue.io.MmapPageIO class uses Oracle's internal classes (sun.misc.Cleaner and sun.nio.ch.DirectBuffer) to immediate cleanup of memory buffer (details about its addition is in PR #6740). It should avoid use of those classes. 

It is introduced in Java 9 a Cleaner class with similar operations (http://download.java.net/java/jdk9/docs/api/java/lang/ref/Cleaner.html).

This item is kind of reminder to investigate the replacement of those internal API in favor of the new Cleaner when Logstash move to support Java 9.

Compile warnings about internal API usage:
```
/home/wainersm/projects/oss/logstash/logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java:3: warning: Cleaner is internal proprietary API and may be removed in a future release
import sun.misc.Cleaner;
               ^
/home/wainersm/projects/oss/logstash/logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java:4: warning: DirectBuffer is internal proprietary API and may be removed in a future release
import sun.nio.ch.DirectBuffer;
                 ^
/home/wainersm/projects/oss/logstash/logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java:3: warning: Cleaner is internal proprietary API and may be removed in a future release
import sun.misc.Cleaner;
               ^
/home/wainersm/projects/oss/logstash/logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java:4: warning: DirectBuffer is internal proprietary API and may be removed in a future release
import sun.nio.ch.DirectBuffer;
                 ^
/home/wainersm/projects/oss/logstash/logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java:3: warning: Cleaner is internal proprietary API and may be removed in a future release
import sun.misc.Cleaner;
               ^
/home/wainersm/projects/oss/logstash/logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java:4: warning: DirectBuffer is internal proprietary API and may be removed in a future release
import sun.nio.ch.DirectBuffer;
                 ^
[...]

PR #7118, opened 2017-05-16
7039 JMH Benchmarks to Illustrate PQ High-Throughput Behaviour

original-brownbear:
@colinsurprenant @suyograo @jordansissel this is the output of the JMH I was talking about earlier today (well and the JMH code itself :)).

The leaky behaviour only shows up at very high throughput (achieved by 50k ack intervalls):

screen shot 2017-05-16 at 02 10 59

As compared to GC keeping up just fine with the much more I/O bound behaviour at 100 event ack intervals:

screen shot 2017-05-16 at 02 06 20

Not necessarily saying we shouldn't pursue #7039 , but it may not be worth doing this for now since the effects are likely overshadowed by GC from the JRuby code as well as plugins in general.

Still maybe we should add this kind of for posterity to use when optimizing? :)
The more interesting effect to examine here may be the extreme slowdown that acking comes with. I think it should be possible to find a more performant approach here, Kafka for example acks in very small intervals too without the throughput to disk going down to `< 10 MB/s`

Benchmark can be executed by simply running `./gradlew jmh` from the project root btw :)

### DIFF
--- a/logstash-core/benchmarks/Readme.md
+++ b/logstash-core/benchmarks/Readme.md
@@ -1,0 +1,56 @@
+# Logstash Microbenchmark Suite
+
+This directory contains the microbenchmark suite of Logstash. It relies on [JMH](http://openjdk.java.net/projects/code-tools/jmh/).
+
+## Getting Started
+
+Just run `./gradlew jmh` from the project root directory. It will build all microbenchmarks, execute them and print the result.
+
+#### Example Output
+
+```bash
+➜  logstash: ./gradlew jmh
+# JMH 1.18 (released 66 days ago)
+# VM version: JDK 1.8.0_121, VM 25.121-b13
+# VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_121.jdk/Contents/Home/jre/bin/java
+# VM options: -Dfile.encoding=US-ASCII -Duser.country=US -Duser.language=en -Duser.variant
+# Warmup: 3 iterations, 100 ms each
+# Measurement: 10 iterations, 100 ms each
+# Timeout: 10 min per iteration
+# Threads: 1 thread, will synchronize iterations
+# Benchmark mode: Throughput, ops/time
+# Benchmark: org.logstash.benchmark.QueueBenchmark.pushToPersistedQueue
+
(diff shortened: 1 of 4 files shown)


original-brownbear:
btw ... if we really are concerned about optimizing the memory footprint of this I think https://github.com/elastic/logstash/issues/6961#issuec is more relevant overall.
Reducing the pressure from needlessly firing `mmap()` in append only situations should have a huge impact when it comes to this kind of throughput.

suyograo:
@original-brownbear I have a suggestion: can we move this to `benchmarks/logstash-core` folder instead of a high level `logstash-core-benchmarks`. We can then make it one place to do micro and macro benchmarks. Can we also add a `README.md` under this `benchmarks` directory?

original-brownbear:
@suyograo as discussed: I'd have to investigate if that's possible with Gradle out of the box.

At least best practices has it, that you'd want to depend on your actual JAR of `logstash-core` and it's def easiest to have a separate path for that. But I'll look into it maybe there is an easy way to do this like Maven has with Invoker :)

suyograo:
@original-brownbear we worked on a similar thing on ES. Maybe this will give you some ideas? https://github.com/elastic/elasticsearch/blob/master/benchmarks/build.gradle

original-brownbear:
@suyograo argh yea the approach in there is perfect :D thanks! Will do the same here :)

original-brownbear:
@suyograo cleaned it up a little :)

logstash-core/benchmarks/Readme.md
```diff
@@ -0,0 +1,11 @@
+# Logstash Microbenchmark Suite
+
+This directory contains the microbenchmark suite of Logstash. It relies on [JMH](http://openjdk.java.net/projects/code-tools/jmh/).
+
+## Getting Started
+
+Just run `./gradlew jmh` from the project root directory. It will build all microbenchmarks, execute them and print the result.
+
```
suyograo: Can we post a sample result here so devs can get an idea of what it looks like?

original-brownbear: @suyograo sure adding :)

Issue #7119, opened 2017-05-16
[test failure] integration test relies on exact geo location values that can change


jsvd:
seen in https://travis-ci.org/elastic/logstash/jobs/232651438#L1924-L1939

```
  1) Test Elasticsearch output can ingest 37K log lines of sample apache logs
     Failure/Error: expect(s["geoip"]["longitude"]).to eq(12.9443)
     
       expected: 12.9443
            got: 12.6944 (#<BigDecimal:6cd65042,'0.126944E2',6(8)>)
     
       (compared using ==)
     # ./specs/es_output_how_spec.rb:34:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
Finished in 11 minutes 57 seconds (files took 4.59 seconds to load)
32 examples, 1 failure
Failed examples:
rspec ./specs/es_output_how_spec.rb:17 # Test Elasticsearch output can ingest 37K log lines of sample apache logs
Randomized with seed 41840
```

It appears the geo data has been updated so this longitude is no longer correct. These tests should not rely on exact data from the geo database.

ph:
I will take this one, simple fix.

PR #7124, opened 2017-05-16
Fix test failure when run from DNS hijacked ISP

robbavey:
Remove the tld from 'unreachable host' to circumvent DNS hijacking
ISPs returning a successful response when hitting a non-existent domain


andrewvc:
CI tests failed due to: https://github.com/elastic/logstash/issues/7119 which is not a blocker. This is good to merge.

Issue #7123, opened 2017-05-16
Evaluate our acceptance test suite and why its failling


ph:
Our current acceptance test suite has been unreliable and hard to debug/maintain, Should we still run it or make it more easy to reason with?

- It's use SSH connection for each test, making reproduction and debugging hard.
- It's been failing randomly for some month
- Its use a git repo for the runner-tool dependencies https://github.com/purbon/runner-tool.git

### TODO

- [ ] update README
- [ ] cleanup the directory structure by using an acceptance directory.

ph:
Investigate if we could run our qa/integration suite on multilpe os, maybe only a non-docker subset?

ph:
Settings `LS_DEBUG_QA` give more details about the error going, it seems we cannot correctly install the java package on the debian platform and we also have an invalid public key for one of the repository.

ph:
That was in fact easy to fix, took time to run/debug.

PR #7130, opened 2017-05-17
Only glob *.conf by default in conf.d folder

### LINKED ISSUE #1831: Feature Request: load_config pattern
If deemed valuable to others:

Would it be possible for someone with ruby experience to update the globbing of configs to be *.conf (might be just as simple as changing line 288 of agent.rb). This would allow you to also keep backups of old configs or move configs out of the way if you want to piecemeal during testing which configs are live. (luckily the added bonus even with \* is that private directories like .git aren't sourced so the entire logstash/conf.d can be under version control)

``` ruby
    287   def load_config(path)
    288     path = File.join(path, "*") if File.directory?(path)
    289
    290     if Dir.glob(path).length == 0
    291       fail(I18n.t("logstash.agent.configuration.file-not-found", :path => path))
    292     end
    293
    294     config = ""
    295     Dir.glob(path).sort.each do |file|
    296       next unless File.file?(file)
    297       if file.match(/~$/)
    298         @logger.debug("NOT reading config file because it is a temp file", :file => file)
    299         next
    300       end
    301       @logger.debug("Reading config file", :file => file)
    302       config << File.read(file) + "\n"
    303     end
    304     return config
    305   end # def load_config
```

### LINKED ISSUE #6979: ignore configuration files not ending in .conf
* Version: 5.3.1

Today, Logstash loads all files sitting in `/etc/logstash/conf.d`.  This can be problematic if a user makes a copy of a configuration file, for example, using the `.bak` extension, as unexpected result occur. This can be especially problematic for Kafka as it can lead an unsuspecting user to balancing issues that are not immediately apparent. Please require and document the use of `.conf` file extensions in `/etc/logstash/conf.d`.

suyograo:
Fixes #6979, Fixes #1831


jordansissel:
Change looks good. Any thoughts on an acceptance test for this?

logstash-core/lib/logstash/config/source/local.rb
```diff
@@ -90,6 +94,13 @@ def path
         end
       end
 
+      def get_unmatched_files
+        # transform "/var/lib/*.conf" => /var/lib/*
+        t = File.split(@path)
+        all_files = Dir.glob(File.join(t.first, "*")).sort
```
jordansissel: This will not show hidden (dot files) in the list of unmatched. I don't anticipate this to be confusing to users, but wanted to point it out.

If a user has a file named '.foo.conf' it will not be matched, nor will '.foo.bar' be shown as being skipped.

suyograo:
@jordansissel I'd love to add acceptance tests, but they are all broken already. Tracking here: https://github.com/elastic/logstash/issues/7123

How about I create an issue and add an acceptance for this when #7123 is settled?

jordansissel:
+1 on creating an issue and adding later.

PR #7131, opened 2017-05-17
Make test for pipeline flushing more resilient.

### LINKED ISSUE #7024: Fails: LogStash::Pipeline Periodic Flush flushes the buffered contents of the filter
Job: elastic+logstash+5.4+dockerubuntu-periodic
link: https://logstash-ci.elastic.co/job/elastic+logstash+5.4+dockerubuntu-periodic/263/console

```
  1) LogStash::Pipeline Periodic Flush flushes the buffered contents of the filter
     Failure/Error: expect(event.get("message").count("\n")).to eq(99)
       
       expected: 99
            got: 98
       
       (compared using ==)
     # ./logstash-core/spec/logstash/pipeline_spec.rb:607:in `(root)'
     # ./rakelib/test.rake:40:in `(root)'
```

```
Finished in 6 minutes 53 seconds (files took 6.24 seconds to load)
2617 examples, 1 failure

Failed examples:

rspec ./logstash-core/spec/logstash/pipeline_spec.rb:597 # LogStash::Pipeline Periodic Flush flushes the buffered contents of the filter

Randomized with seed 30079

```

andrewvc:
Instead of depending on the now deprecated multiline filter we use a dummy filter that just emits events. This simplifies the test and dramatically reduces timing issues. I also increased the max-wait for the timer just in case

Fixes #7024

### DIFF
--- a/logstash-core/spec/logstash/pipeline_spec.rb
+++ b/logstash-core/spec/logstash/pipeline_spec.rb
@@ -81,6 +81,21 @@ class DummySafeFilter < LogStash::Filters::Base
   def close() end
 end
 
+class DummyFlushingFilter < LogStash::Filters::Base
+  config_name "dummyflushingfilter"
+  milestone 2
+
+  def register() end
+  def filter(event) end
+  def periodic_flush
+    true
+  end
+  def flush(options)
+    return [::LogStash::Event.new("message" => "dummy_flush")]
+  end
+  def close() end
+end
+
 class TestPipeline < LogStash::Pipeline
   attr_reader :outputs, :settings
 end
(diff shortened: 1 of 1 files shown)


original-brownbear:
@andrewvc code looks fine :)

Please ignore the below if it's stupid, could be I'm missing some piece of knowledge here still:

But, is this approach actually correct? To me the issue @ph describes here https://github.com/elastic/logstash/issues/7024#issuecomment-299227407 is that multiple workers are handling the events and hence the ML filter only receives a subset of the 99 input test events. Shouldn't this be verified to actually be the problem?
If so, couldn't we do this by simply using a `static/class/@@` counter in `def filter(event)`?

```ruby
def filter(event)
   @@counter.incrementAndGet()
end
```

with `@@counter` being an `AtomicInteger` or `AtomicLong`?

andrewvc:
@original-brownbear as I read the intent of the test, the point wasn't to test the multiline filter, but just to test that periodic flushing works. Testing the multiline filter would be more of an integration test IMHO, not a pipeline test (also ML is deprecated anyway). That's why I only check that at least one flushed event goes through. In fact, usually more do.

original-brownbear:
@andrewvc LGTM :) I guess my understanding of the intent was more IT like in that `filter(event)` was supposed to be verified to work too :) Thanks for the explanation!

PR #7132, opened 2017-05-17
Clean settings and refactor web api specs

### LINKED ISSUE #6245: Pipeline worker count test failed.
https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-os-compatibility/os=amazon/295/console

```
  1) LogStash::Pipeline defaulting the pipeline workers based on thread safety when there are threadsafe filters only starts multiple filter threads
     Failure/Error: expect(pipeline.worker_threads.size).to eq(worker_thread_count)
       
       expected: 5
            got: 8
```

andrewvc:
Fixes #6245

Clean settings before all tests just in case ( Fixes #6245 )

Also 

Refactor API specs to play well with others:

1. Moves API specs out of their special hierarchy
2. Removes the API spec spec_helper
3. Reactivates that stats command spec (that was accidentally not being run before due to it not having _spec as a suffix

This was required to fix stats stuff as the two spec_helpers would screw each other up.


ph:
@andrewvc and myself discussed that issue on zoom today we agreed on the following:

- We think that we have outgrow clamps features
- The Settings Singleton is too link to clamp
- Adding more feature to clamp could be a rabbit hole.
- The easiest change to make right now is to make sure the `Logstash::SETTINGS` is reset between run of expectation and we can do it globally at the suite level.

PR #7134, opened 2017-05-17
Added note on Docker dependency to IT Readme

original-brownbear:
Documenting the new dependency due to #7091


jsvd:
I'd suggest a bit more complete instructions on running the tests. I'm not super familiar with docker, but I'm guessing some of [these commands](https://github.com/original-brownbear/logstash/blob/4185aa29b19da2915b60aade49292ae67bd8ad22/qa/integration/README.md#preparing-a-test-run) now must have a docker cli command prefix?

original-brownbear:
@jsvd no additional magic needed, if you got a running Docker service on box Ruby logic I added for handling Docker does it all transparently :)

PR #7140, opened 2017-05-17
Move rake test:* tasks to all use `bin/rspec`.

andrewvc:
This eliminates the differences in behavior between bin/rspec and rake test:*

### DIFF
--- a/rakelib/test.rake
+++ b/rakelib/test.rake
@@ -6,66 +6,35 @@ require 'pathname'
 
 namespace "test" do
 
-  task "setup" do
-
-    # make sure we have a ./data/queue dir here
-    # temporary wiring until we figure proper queue initialization sequence and in test context etc.
-    mkdir "data" unless File.directory?("data")
-    mkdir "data/queue" unless File.directory?("data/queue")
-
-    # Need to be run here as because if run afterwards (after the bundler.setup task) then the report got wrong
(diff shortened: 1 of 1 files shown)


rakelib/test.rake
```diff
@@ -6,66 +6,35 @@ require 'pathname'
...
+  task "core-java" do
     exit(1) unless system './gradlew clean test --info'
-    
-    specs = ["spec/unit/**/*_spec.rb", "logstash-core/spec/**/*_spec.rb"]
-
-    Rake::FileList[*specs]
   end
 
   desc "run all core specs"
   task "core" => ["core-slow"]
```
original-brownbear: @andrewvc This one was what ran the Java tests also. We must should `core-java` as a depend task for it right?

andrewvc: Yeah, great point, I'll add that!

andrewvc:
@original-brownbear the original version of these tasks ran the java tests for all core tests. I would argue that we should skip the java tests on `core-fast`. If you really care about them you can run `core:java` IMHO. WDYT?

/cc @suyograo

original-brownbear:
@andrewvc agreed, the `core` ones are also what's run on Travis/CI and that's what matters imo, as long as we don't lose our default CI `test` task running the Java tests we're good I think :)

original-brownbear:
@andrewvc in that sense, LGTM :)

PR #7143, opened 2017-05-17
MINOR: Cleanup some inefficient List usage

original-brownbear:
Just some minor cleanups of `List` performance cases in the critical path.

* `stream`, `map` then `collect` is very slow relative to just iterating over a list due to all the overhead allocations of intermediate `Collection`s
* `new ArrayList()` as an empty list constant is a waste of allocations + has some potential for introducing bugs since it's mutable, better to use the constant
* if we know the size an `ArrayList()` is going to be we should pre-allocate the size

Two very minors:
* IOException is not thrown, removed from the signature 
* `Batch` is a `Closeable` and as such should not by assigned to a local variable in a non-resource-safe manner (doesn't matter here but nice for consistency)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
@@ -1,12 +1,11 @@
 package org.logstash.ackedqueue;
 
-import org.logstash.ackedqueue.io.PageIO;
-
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.List;
-import java.util.stream.Collectors;
+import org.logstash.ackedqueue.io.PageIO;
 
 public abstract class Page implements Closeable {
     protected final int pageNum;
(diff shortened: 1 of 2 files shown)


logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
```diff
@@ -54,16 +53,18 @@ public Batch readBatch(int limit) throws IOException {
...
 
         SequencedList<byte[]> serialized = this.pageIO.read(this.firstUnreadSeqNum, limit);
-        List<Queueable> deserialized = serialized.getElements().stream().map(e -> this.queue.deserialize(e)).collect(Collectors.toList());
-
+        List<byte[]> elements = serialized.getElements();
+        final int count = elements.size();
+        List<Queueable> deserialized = new ArrayList<>(count);
+        for (final byte[] element : elements) {
+            deserialized.add(this.queue.deserialize(element));
+        }
```
robbavey: Is this move away from stream/map/collect something you want to do in general, or just in this particular instance for specific performance reasons?

original-brownbear: @robbavey well in general the `Stream` API has it's use cases so I like using it off the critical path.
But in the critical path it's kind of a no-go :)

original-brownbear:
@robbavey thanks for the review!

Issue #7086, opened 2017-05-12
[test failure] Test Logstash instance should keep the same id between restarts


jsvd:
from: https://travis-ci.org/elastic/logstash/jobs/231513226#L2429

```
  1) Test Logstash instance should keep the same id between restarts
     Failure/Error: JSON.parse(open("http://localhost:9600/").read)["id"]
     
     Errno::ECONNREFUSED:
       Connection refused - Connection refused
     # ./specs/01_logstash_bin_smoke_spec.rb:164:in `get_id'
     # ./specs/01_logstash_bin_smoke_spec.rb:175:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
Finished in 25 minutes 38 seconds (files took 14.15 seconds to load)
34 examples, 1 failure
Failed examples:
rspec ./specs/01_logstash_bin_smoke_spec.rb:167 # Test Logstash instance should keep the same id between restarts
```

jsvd:
taking on this one. should be just a matter of wrapping get_id in a try, as logstash process may be up but the webserver not yet.

PR #7144, opened 2017-05-18
fix failing test due to faulty api request

jsvd:
fixes https://github.com/elastic/logstash/issues/7086


original-brownbear:
@jsvd LGTM! ... was thinking about putting a small delay in the retry loop, but I think the operation of reading from (or just attempting to open) a http connection has probably enough overhead to it to not make this create weird resource contention spots :)

jsvd:
@original-brownbear it's a fair point, luckily enough the [Stud.try](https://github.com/jordansissel/ruby-stud/blob/master/lib/stud/try.rb#L26) already builts in backoffs 👍

PR #7151, opened 2017-05-18
Fix FD leak in FileCheckpointIO, closes #7150

### LINKED ISSUE #7150: org.logstash.ackedqueue.io.FileCheckpointIO#read(java.lang.String) Is leaking FDs
We're leaking file descriptors in this code (`org.logstash.ackedqueue.io.FileCheckpointIO#read(java.lang.String)`) that reads a checkpoint file:

```java
    @Override
    public Checkpoint read(String fileName) throws IOException {
        Path path = Paths.get(dirPath, fileName);
        InputStream is = Files.newInputStream(path);
        return read(new BufferedChecksumStreamInput(new InputStreamStreamInput(is)));
    }
```

` InputStream is = Files.newInputStream(path);` is never closed and should be a `try-with-resources` I guess. Or since this is such small files a `Files.readAllBytes()` approach might even make more sense.

original-brownbear:
closes #7150 , `Files.readAllBytes` is def the way to go for nano sized files in terms of performance imo.


original-brownbear:
@robbavey thanks for the review! Merging :)

PR #7152, opened 2017-05-18
Prevent variable leak between test

ph:
When running this test with a newer version of rspec we got the warning
that some values where leaking between example:

```
 1) LogStash::ShutdownWatcher when pipeline is not stalled .unsafe_shutdown = false shouldn't force the shutdown
     Failure/Error: m.call(*args)
       The use of doubles or partial doubles from rspec-mocks outside of the per-test lifecycle is not supported.
     # ./logstash-core/lib/logstash/shutdown_watcher.rb:67:in `pipeline_report_snapshot'
     # ./logstash-core/spec/logstash/shutdown_watcher_spec.rb:22:in `block in /Users/ph/es/second/logstash/logstash-core/spec/logstash/shutdown_watcher_spec.rb'
     # ./logstash-core/lib/logstash/shutdown_watcher.rb:44:in `block in start'
     # ./vendor/bundle/jruby/2.3.0/gems/stud-0.0.22/lib/stud/interval.rb:20:in `interval'
     # ./logstash-core/lib/logstash/shutdown_watcher.rb:42:in `start'
     # ./logstash-core/spec/logstash/shutdown_watcher_spec.rb:93:in `block in (root)'
```

We have added a few helpers method in the class that allowed us to
simplify the expectation and removed any usage of `Thread#kill`

### DIFF
--- a/logstash-core/spec/logstash/shutdown_watcher_spec.rb
+++ b/logstash-core/spec/logstash/shutdown_watcher_spec.rb
@@ -9,22 +9,12 @@ describe LogStash::ShutdownWatcher do
   let(:pipeline) { double("pipeline") }
   let(:reporter) { double("reporter") }
   let(:reporter_snapshot) { double("reporter snapshot") }
-  report_count = 0
 
   before :each do
     allow(pipeline).to receive(:reporter).and_return(reporter)
     allow(pipeline).to receive(:thread).and_return(Thread.current)
     allow(reporter).to receive(:snapshot).and_return(reporter_snapshot)
     allow(reporter_snapshot).to receive(:o_simple_hash).and_return({})
-
-    allow(subject).to receive(:pipeline_report_snapshot).and_wrap_original do |m, *args|
-      report_count += 1
-      m.call(*args)
-    end
-  end
-
-  after :each do
-    report_count = 0
   end
 
   context "when pipeline is stalled" do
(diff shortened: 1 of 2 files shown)


logstash-core/spec/logstash/shutdown_watcher_spec.rb
```diff
@@ -69,8 +59,9 @@
       it "shouldn't force the shutdown" do
         expect(subject).to_not receive(:force_exit)
         thread = Thread.new(subject) {|subject| subject.start }
-        sleep 0.1 until report_count > check_threshold
-        thread.kill
+        sleep 0.1 until subject.attempts_count > check_threshold
+        subject.stop!
+        thread.join
```
original-brownbear: @ph we may wanna join with a timeout here right, given how much fun a thread dump is from JRuby right and assert that it worked?

original-brownbear:
@ph LGTM, just one suggestion to avoid hanging tests :)

logstash-core/spec/logstash/shutdown_watcher_spec.rb
ph: updated with your suggestion.

original-brownbear: @ph nice thanks!

original-brownbear:
@ph LGTM!

PR #7156, opened 2017-05-18
Initial PR for code review

untergeek:
My changes are here for review.  There are no tests yet.

### DIFF
--- a/logstash-core/lib/logstash/config/source/modules.rb
+++ b/logstash-core/lib/logstash/config/source/modules.rb
@@ -1,0 +1,46 @@
+# encoding: utf-8
+require "logstash/config/source/base"
+require "logstash/config/pipeline_config"
+require "logstash/util/loggable"
+require "logstash/errors"
+
+module LogStash module Config module Source
+  class Modules < Base
+    include LogStash::Util::Loggable
+    def pipeline_configs
+      pipelines = []
+      plugin_modules = LogStash::PLUGIN_REGISTRY.plugins_with_type(:modules)
+
+      modules_array = @settings.get("modules.cli").empty? ? @settings.get("modules") : @settings.get("modules.cli")
+      logger.debug("Configured modules", :modules_array => modules_array.to_s)
+      module_names = []
+      # modules_array.each {|module_hash| module_names << module_hash["name"]}
+      module_names = modules_array.collect {|module_hash| module_hash["name"]}
+      if module_names.length > module_names.uniq.length
+        duplicate_modules = module_names.group_by(&:to_s).select { |_,v| v.size > 1 }.keys
+        raise LogStash::ConfigLoadingError, I18n.t("logstash.modules.configuration.modules-must-be-unique", :duplicate_modules => duplicate_modules)
+      end
+      ### Here is where we can force the modules_array to use only [0] for 5.5, and leave
+      ### a warning/error message to that effect.
+      modules_array.each do |module_hash|
+        begin
+          current_module = plugin_modules.find { |allmodules| allmodules.module_name == module_hash["name"] }
+          alt_name = "module-#{module_hash["name"]}"
+          pipeline_id = alt_name
+          config_string = current_module.config_string(module_hash)
+          logger.debug("Config string for module", :config_string => config_string, :module => module_hash["name"])
+          config_part = org.logstash.common.SourceWithMetadata.new("module", alt_name, config_string)
+          pipelines << PipelineConfig.new(self, pipeline_id.to_sym, config_part, @settings)
+        rescue => e
+          raise LogStash::ConfigLoadingError, I18n.t("logstash.modules.configuration.parse-failed", :error => e.message)
+        end
+      end
+      pipelines
+    end
+
+    def match?
+      # will fill this later
+      true
+    end
+  end 
+end end end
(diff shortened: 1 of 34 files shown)


logstash-core/lib/logstash/config/source/modules.rb
```diff
@@ -0,0 +1,44 @@
...
+      module_names = []
+      modules_array.each {|module_hash| module_names << module_hash["name"]}
+      if module_names.length > module_names.uniq.length
+        raise LogStash::ConfigLoadingError, I18n.t("logstash.modules.configuration.modules-must-be-unique", :modules => module_names)
+      end
+      ### Here is where we can force the modules_array to use only [0] for 5.5, and leave
+      ### a warning/error message to that effect.
+      modules_array.each do |module_hash|
+        begin
+          thismodule = plugin_modules.find { |allmodules| allmodules.modulename == module_hash["name"] }
```
suyograo: `thismodule` => `current_module`

logstash-core/lib/logstash/config/source/modules.rb
```diff
@@ -0,0 +1,44 @@
...
+          altname = "module-#{module_hash["name"]}"
```
suyograo: `altname` -> `alt_name` for ruby style.

logstash-core/lib/logstash/config/source/modules.rb
```diff
@@ -0,0 +1,44 @@
...
+          pipeline_id = module_hash["pipeline_id"].nil? ? altname : module_hash["pipeline_id"]
```
suyograo: We shouldn't even let users specify `pipeline_id`, to simplify -- for 6.0 as well as 5.x. We can infer from the module name as you've already done (`altname`). One less config to worry about.

logstash-core/lib/logstash/config/source/modules.rb
```diff
@@ -0,0 +1,44 @@
...
+          config_string = thismodule.config_string(module_hash)
+          logger.debug("CONFIG_STRING", :config_string => config_string)
```
suyograo: needs to be more descriptive. 

"Config string for module", :config_string, :module => name"

logstash-core/lib/logstash/modules_cli_parser.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def decommafy(value)
+    retval = []
+    if value.split(',').length == 1
+      retval << value
+    elsif value.split(',').length > 1
+      value.split(',').each {|subvalue| retval << subvalue}
+    end
+    retval
+  end
```
ph: I don't see the unit test for this method, but I think You don't need to have a conditional here and just return `value.split(',')`

logstash-core/lib/logstash/modules_cli_parser.rb
```diff
@@ -0,0 +1,89 @@
...
+    if unparsed.split('=').length >= 2
+      # This hackery is to catch the possibility of an equals (`=`) sign 
+      # in a passphrase, which might result in an incomplete key.  The 
+      # portion before the first `=` should always be the key, leaving 
+      # the rest to be the value
+      values = unparsed.split('=')
+      k = values.shift
+      # v = values.join('=')
+    else
+      raise LogStash::ConfigLoadingError, I18n.t("logstash.modules.configuration.modules-variables-malformed", :rawvar => (module_name + '.' + unparsed))
```
ph: thanks for all the feedback for the user!

logstash-core/lib/logstash/modules_cli_parser.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+    return k,values.join('=')
```
ph: Can we move the return into the if statement so we get out early.

logstash-core/lib/logstash/runner.rb
```diff
@@ -248,6 +261,10 @@ def execute
 
     return start_shell(setting("interactive"), binding) if setting("interactive")
 
+    module_parser = LogStash::ModulesCLIParser.new(@modules_list, @modules_variable_list)
+    # Now populate Setting for modules.list with our parsed array.
+    @settings.set("modules.config", module_parser.output)
+
```
ph: Nice, I would have expected it to be more hackery but this is a nice way to do it.

logstash-core/lib/logstash/environment.rb
```diff
@@ -20,6 +20,8 @@ module Environment
     Setting::NullableString.new("path.config", nil, false),
  Setting::WritableDirectory.new("path.data", ::File.join(LogStash::Environment::LOGSTASH_HOME, "data")),
     Setting::NullableString.new("config.string", nil, false),
+                    Setting.new("modules.config", Array, []),
```
untergeek: Does `modules.name` make more sense? Or `modules.cli`?  I'm partial to `modules.cli`

suyograo: I like `modules.cli`. Lets go with that.

logstash-core/lib/logstash/modules.rb
```diff
@@ -0,0 +1,48 @@
...
+    def setting(value, default)
+      @settings[value].nil? ? default : @settings[value]
+    end
+
+    def render
+      # process the template and settings
+      # send back as a string with no newlines (the '>' part)
+      renderer = ERB.new(File.read(@template), 3, '>')
+      renderer.result(binding)
+    end
```
untergeek: Yep.

logstash-core/lib/logstash/modules_cli_parser.rb
untergeek: You're right.  Breaking this down into a class removed its need altogether.  I'll be resolving this with `#concat` and a single `split(',')` instead.

Issue #7154, opened 2017-05-18
Queue Space in Bytes Metric Breaks Tests


andrewvc:
See: https://travis-ci.org/elastic/logstash/jobs/233693574

```
  1) Test Monitoring API can retrieve queue stats
     Failure/Error: expect(result["pipeline"]["queue"]["data"]["free_space_in_bytes"]).not_to be_nil
     
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./specs/monitoring_api_spec.rb:64:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # ./specs/monitoring_api_spec.rb:56:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
Finished in 10 minutes 45 seconds (files took 6.01 seconds to load)
32 examples, 1 failure
Failed examples:
rspec ./specs/monitoring_api_spec.rb:51 # Test Monitoring API can retrieve queue stats
```

Edit by @original-brownbear:

Retagged as bug on account of https://github.com/elastic/logstash/issues/7154#issuecomment-302655480

jsvd:
note: this test should fail less often now since https://github.com/elastic/logstash/pull/7144 has been introduced. this PR triggers KeyError when fetching data from the hash, causing Stud.try to retry the block

original-brownbear:
@jsvd argh I thought it retried on `NoMethodError` too? At least it claims to? :D

jsvd:
[no :(](https://github.com/jordansissel/ruby-stud/blob/master/lib/stud/try.rb#L96-L97)

original-brownbear:
@jsvd oh :D that makes my PR https://github.com/elastic/logstash/pull/7157 very pointless then and explains everything :)
I think we can close here in that case?

original-brownbear:
@jsvd I guess we should also fix this kind of thing:

[...]

PR #7145, opened 2017-05-18
protect monitoring_api_spec against timing issues

jsvd:
this pr had to remove the NoMethodError from the list of acceptable Stud.try exceptions since [Stud.try catches that exception specifically before others](https://github.com/jordansissel/ruby-stud/blob/master/lib/stud/try.rb#L92-L99).

To handle gaps in the response from the webapi and trigger the Stud.try, this PR relies on Hash#fetch that causes a KeyError (inherits from StandardError) on missing key.

fixes https://github.com/elastic/logstash/issues/7077

PR #7157, opened 2017-05-18
#7154 delete queue path in logstash integration tests to improve test isolation

original-brownbear:
This is the best I could find here, I wasn't able to reproduce this *but* the fact that we keep reusing the same persisted queue path over and over slows the whole thing way down (by that I mean the time until `logstash-core/lib/logstash/pipeline.rb:694` (`collect_stats`) works out the first time, evidently the other parts of the metrics hash are populated).

=> I'd start by fixing test isolation here and ensuring that the queue path is removed before any tests are ran (hence adding it to `initialize` too, to avoid interference from unclean shutdowns).

With reuse of the path I can see runs take 20s at times for this test, without it's 10s every time. I get that we should also look into improving the queue performance (probably some locking timing issue thing), but test isolation is a valid step one here imo.

### DIFF
--- a/qa/integration/services/logstash_service.rb
+++ b/qa/integration/services/logstash_service.rb
@@ -40,8 +40,7 @@ class LogstashService < Service
       @logstash_home += "-SNAPSHOT" unless Dir.exists?(@logstash_home)
       
       @logstash_queue_path = File.join("#{@logstash_home}", "data/queue")
-      FileUtils.remove_dir(@logstash_queue_path, true)
-      FileUtils.mkdir(@logstash_queue_path)
+      clear_pq_data
       
       puts "Using #{@logstash_home} as LS_HOME"
       @logstash_bin = File.join("#{@logstash_home}", LS_BIN)
@@ -140,8 +139,7 @@ class LogstashService < Service
       @process.io.stdin.close rescue nil
       @process.stop
       @process = nil
-      FileUtils.remove_dir(@logstash_queue_path, true)
-      FileUtils.mkdir(@logstash_queue_path)
+      clear_pq_data
     end
   end
 
@@ -208,6 +206,13 @@ class LogstashService < Service
 
   def lock_file
     File.join(@logstash_home, "Gemfile.jruby-1.9.lock")
+  end
+  
+  private
+  
+  def clear_pq_data
+    FileUtils.remove_dir(@logstash_queue_path, true)
+    FileUtils.mkdir(@logstash_queue_path)
   end
 
   class PluginCli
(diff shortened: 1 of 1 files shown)


suyograo:
> With reuse of the path I can see runs take 20s at times for this test, without it's 10s every time. I get that we should also look into improving the queue performance (probably some locking timing issue thing), but test isolation is a valid step one here imo.

Hmm, this is an interesting side-effect observed from these tests.

suyograo:
LGTM

qa/integration/services/logstash_service.rb
```diff
@@ -38,7 +38,10 @@ def initialize(settings)
       # First try without the snapshot if it's there
       @logstash_home = File.expand_path(File.join(LS_BUILD_DIR, ls_file), __FILE__)
       @logstash_home += "-SNAPSHOT" unless Dir.exists?(@logstash_home)
-
+      
+      @logstash_queue_path = File.join("#{@logstash_home}", "data")
```
suyograo: Actually, the queue path is `LS_HOME/data/queue`. I would update the variable here to `@logstash_data_path`

suyograo: I think it makes sense to delete the entire `LS_HOME/data`

original-brownbear: @suyograo damn busted on travis with:

```
rspec ./specs/01_logstash_bin_smoke_spec.rb:145 # Test Logstash instance should keep the same id between restarts

```
 will get on this in the morning again :)

suyograo: Yeah, I think we can only delete on `teardown`.

original-brownbear:
@suyograo so I played with it a little, works fine the way I now made it work (delete just the contents of the queue path, not the directory itself).

But that said, this is kind of a bug in production code as far as I can see. Getting the queue back up and running from such a trivial amount of data should take a few `ms` tops.

I think two points are probably worth addressing in follow ups:

* Add a benchmark for `Queue.open()` ... who knows, maybe it locks up after a few runs ...
* Dockerize Logstash in tests too (this was actually suggested by @andrewvc with me disagreeing). I think I was wrong here actually and misunderstood the tests to some degree. We're not running Logstash in the same JVM as `rspec` anyways here, so Dockerizing Logstash like we did for Kafka  would solve all this craziness (and give us much more granular control over the env Logstash runs in for settings tests ... like the queue free space thing for example :D )

WDYT, go with this approach to stabilize for now and address the above?

original-brownbear:
@suyograo closing here though @jsvd fixed the issue, it was a bit of a misleading use of `Stud.try` that threw me off here :) See https://github.com/elastic/logstash/pull/7145/files

PR #7161, opened 2017-05-19
MINOR: Clean up redundant matchers in QueueTest

original-brownbear:
Follow up to #7063 

* We shouldn't have redundant matchers, they potentially make the test output hard to read and just add needless clutter to the codebase


guyboertje:
LGTM
I have not had much experience in coding with or reading code with hamcrest matchers - so its good read what you changed!

original-brownbear:
@guyboertje alright, thanks for the review :)

PR #7163, opened 2017-05-19
MINOR: Prevent leaking threads from QueueTest

original-brownbear:
Another follow up to #7063 
We were leaking threads in `QueueTest` since `executor.shutdown()` comes with no guarantees and also doesn't block.

Added proper shutting down of the executor that fails tests if the shutdown fails. Unfortunately I had to handle one of the execs inline since it requires more than the single thread all other tests use, but still this should make things a little more predictable :)

PS: Also removed the dead `long seqnum` local vars, those looked dirty :)


original-brownbear:
@andrewvc thanks !

Issue #7111, opened 2017-05-15
Fix redundant byte[] allocation when computing CRC32


original-brownbear:
As pointed out in #7058 and specifically https://github.com/elastic/logstash/pull/7058#issuecomment-301588948

there are more cases than just this one that have redundant `byte[]` allocation for CRC32.
Those should be fixed the same way they were fixed in that PR and the logic dried up across CRC32 computations.

original-brownbear:
@suyograo why close this one? :)

suyograo:
oh, i thought it was fixed in #7058, no?

suyograo:
ah now I see there are more places, nvm. reopening

original-brownbear:
merged another step in #7166 but on hold now, please don't close though :)

PR #7166, opened 2017-05-19
#7111 cleaned up one case of redundant CRC32 digest and byte[] creation

original-brownbear:
One step for #7111, we can't do all of this in one PR, there's many many spots :)

* Moved to reused `CRC32` instance
* Serialize into `DirectByteBuffer` and adjusted file writing accordingly
* Adjusted benchmark to show potential gain/loss but no effect thanks to file creation being so expensive that it outweighs the other operations

### DIFF
--- a/logstash-core/benchmarks/src/main/java/org/logstash/benchmark/QueueBenchmark.java
+++ b/logstash-core/benchmarks/src/main/java/org/logstash/benchmark/QueueBenchmark.java
@@ -87,8 +87,8 @@ public class QueueBenchmark {
             .capacity(256 * 1024 * 1024)
             .queueMaxBytes(Long.MAX_VALUE)
             .elementIOFactory(MmapPageIO::new)
-            .checkpointMaxWrites(50_000)
-            .checkpointMaxAcks(50_000)
+            .checkpointMaxWrites(1024)
+            .checkpointMaxAcks(1024)
             .checkpointIOFactory(FileCheckpointIO::new)
             .elementClass(Event.class).build();
     }
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
@@ -1,23 +1,20 @@
 package org.logstash.ackedqueue.io;
 
+import java.nio.channels.FileChannel;
 import org.logstash.ackedqueue.Checkpoint;
 import org.logstash.common.io.BufferedChecksumStreamInput;
-import org.logstash.common.io.BufferedChecksumStreamOutput;
-import org.logstash.common.io.ByteArrayStreamOutput;
 import org.logstash.common.io.InputStreamStreamInput;
 
 import java.io.ByteArrayInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.nio.file.Files;
-import java.nio.file.OpenOption;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import static java.nio.file.StandardOpenOption.CREATE;
-import static java.nio.file.StandardOpenOption.WRITE;
-import static java.nio.file.StandardOpenOption.TRUNCATE_EXISTING;
-import static java.nio.file.StandardOpenOption.DSYNC;
+import java.util.zip.CRC32;
 
-public class FileCheckpointIO  implements CheckpointIO {
+public class FileCheckpointIO implements CheckpointIO {
 //    Checkpoint file structure
 //
 //    byte version;
@@ -35,9 +32,16 @@ public class FileCheckpointIO  implements CheckpointIO {
             + Integer.BYTES  // eventCount
             + Integer.BYTES;    // checksum
 
+    /**
+     * Using {@link java.nio.DirectByteBuffer} to avoid allocations and copying in
+     * {@link FileChannel#write(ByteBuffer)} and {@link CRC32#update(ByteBuffer)} calls.
+     */
+    private final ByteBuffer buffer = ByteBuffer.allocateDirect(BUFFER_SIZE);
+
+    private final CRC32 crc32 = new CRC32();
+
     private static final String HEAD_CHECKPOINT = "checkpoint.head";
     private static final String TAIL_CHECKPOINT = "checkpoint.";
-    private static final OpenOption[] WRITE_OPTIONS = { WRITE, CREATE, TRUNCATE_EXISTING, DSYNC };
     private final String dirPath;
 
     public FileCheckpointIO(String dirPath) {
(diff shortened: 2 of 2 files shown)


original-brownbear:
@jordansissel pong :)

original-brownbear:
@jordansissel unpong, get well soon this one is not urgent :)

logstash-core/benchmarks/src/main/java/org/logstash/benchmark/QueueBenchmark.java
```diff
@@ -87,8 +87,8 @@ private static Settings settings() {
             .capacity(256 * 1024 * 1024)
             .queueMaxBytes(Long.MAX_VALUE)
             .elementIOFactory(MmapPageIO::new)
-            .checkpointMaxWrites(50_000)
-            .checkpointMaxAcks(50_000)
+            .checkpointMaxWrites(1024)
```
jordansissel: It's not clear why this change was made. Can you omit it or include the reason in the commit message?

logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
```diff
@@ -115,14 +116,18 @@ private Checkpoint read(BufferedChecksumStreamInput crcsi) throws IOException {
...
+        buf.clear();
+        buf.putShort((short)Checkpoint.VERSION);
+        buf.putInt(checkpoint.getPageNum());
+        buf.putInt(checkpoint.getFirstUnackedPageNum());
+        buf.putLong(checkpoint.getFirstUnackedSeqNum());
+        buf.putLong(checkpoint.getMinSeqNum());
+        buf.putInt(checkpoint.getElementCount());
+        buf.flip();
+        crc32.update(buf);
+        buf.position(BUFFER_SIZE - Integer.BYTES).limit(BUFFER_SIZE);
```
jordansissel: the bytebuffer api is such a strange thing.

I'd kind of prefer using `buf.mark()` (before you call `flip`) then `buf.reset()` instead of `buf.position(...)` to avoid any math problems, but I leave it up to you to decide. (keep the `.limit` call)

logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
```diff
@@ -64,10 +63,12 @@ public Checkpoint write(String fileName, int pageNum, int firstUnackedPageNum, l
 
     @Override
     public void write(String fileName, Checkpoint checkpoint) throws IOException {
-        Path path = Paths.get(dirPath, fileName);
-        final byte[] buffer = new byte[BUFFER_SIZE];
         write(checkpoint, buffer);
-        Files.write(path, buffer, WRITE_OPTIONS);
+        buffer.flip();
+        try (FileOutputStream out = new FileOutputStream(Paths.get(dirPath, fileName).toFile())) {
```
jordansissel: This also changes the file writing strategy from Files.write to FileOutputStream, can you tell me about this change?

logstash-core/benchmarks/src/main/java/org/logstash/benchmark/QueueBenchmark.java
original-brownbear: @jordansissel see last point in the PR description, I added this to verify that this change does not come with unexpected performance issues, so I moved the ack interval to the default here. Makes sense right?

logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
original-brownbear: @jordansissel If I use `Files.write()` I have to operate in the on heap `byte[]` paradigm which forces another memory copy on me. This way I can go straight for writing out the buffer and retain the resilience by calling `sync`.

logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
original-brownbear: @jordansissel I think Java will disappoint you here :(

```java
    public final Buffer flip() {
        limit = position;
        position = 0;
        mark = -1;
        return this;
    }
```

marks aren't retained in `flip` so it won't help us here right?

logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
jordansissel: I had a curiosity, so I looked at the FileOutputStream.getChannel().write() effect. It calls sun.nio.ch.IOUtil.write() which will create a ByteBuffer if the given buffer is not a DirectBuffer. Cool -- this looks like it indeed avoids creating extra byte array.

Given this constraint, I think we need a comment on the `allocateDirect` call (line 34) to indicate this necessary in order to achieve your zero-copy goal. A comment something like (on line 34) `// We use a DirectBuffer here in order to achieve a zero-copy checkpoint write`.

Additionally for safety (regarding future maintenance), add an `assert buffer.isDirect()` at the top of the `write` method?

I am OK with your proposed change to use FileOutputStream, and my above comments are aimed at helping us not accidentally break your zero-copy intent.

logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
jordansissel: Buffer api continues to be a strange and mysterious thing. Good idea to check that the mark is destroyed on flip.

logstash-core/benchmarks/src/main/java/org/logstash/benchmark/QueueBenchmark.java
original-brownbear: @jordansissel sure sec :)

logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
original-brownbear: @jordansissel I like putting a comment here explaining the use of a direct buffer because of I/O (and actually also the CRC32 call ... this is a lot less obvious to only be 0GC using a direct buffer imo :)).
I don't really like `assert` tbh, it's such a standard spot in Java 8 (Direct way caught up to Heap performance wise, even for serializing tiny old `int` the direct one is ahead on OSX for me). 
Seems like this invites a bunch of weird (in the sense that I have never seen anyone asserting `isDirect() == true` anywhere) `assert` all over the place. I think this is really something that should be looked out for in code reviews instead (which would be made easy by having the comment I guess :D).

original-brownbear: Just a comment ok ? :)

original-brownbear:
@jordansissel comment added as requested and commit message fixed to include the benchmark change :)

logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
jordansissel: My thought on `assert` is that we are relying on a property of an object that cannot be checked by the compiler, which kind of stinks. 

Having an `assert` makes our tests fail if we miss something in code review (we miss things frequently, and that's ok, we are human). Similar to why I like tests because we cannot catch everything in code review (imperfect humans with finite time and attention available).

I am OK without an assert given the comment is a good hint.

PR #7167, opened 2017-05-19
Remove dependencies on JRMonitor gem

robbavey:
First pass on removing dependencies on the JRMonitor Gem:

- The code is broadly the same as that originally from the JRMonitor gem, with the JRuby ruby report classes converted to straight Java report classes, which can probably go away in a subsequent PR.

- The Monitor classes are mostly ported over from the existing gem - the MemoryMonitor class has been somewhat refactored to DRY things up. 

- The ruby code is essentially unchanged - the Java classes return Maps rather than strongly typed objects.

- The existing rspec tests for the ruby report classes have been ported over to be junit tests.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/instrument/monitors/HotThreadsMonitor.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/monitors/HotThreadsMonitor.java
@@ -1,0 +1,199 @@
+package org.logstash.instrument.monitors;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Hot threads monitoring class. This class pulls information out of the JVM #
+ * provided beans and lest the different consumers query it.
+ * Created by purbon on 12/12/15.
+ */
+public class HotThreadsMonitor {
+
+    private static final String ORDERED_BY = "ordered_by";
+    private static final String STACKTRACE_SIZE = "stacktrace_size";
+    private final Logger logger = LogManager.getLogger(HotThreadsMonitor.class);
+
+    /**
+     * Placeholder for a given thread report
+     */
+    public static class ThreadReport {
+
+        private static final String CPU_TIME = "cpu.time";
+        private static final String BLOCKED_COUNT = "blocked.count";
+        private static final String BLOCKED_TIME = "blocked.time";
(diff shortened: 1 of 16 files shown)


logstash-core/src/main/java/org/logstash/instrument/monitors/HotThreadsMonitor.java
```diff
@@ -0,0 +1,191 @@
...
+                 */
+                reports.put(threadId, new ThreadReport(info, cpuTime));
+            }
+        }
+        return sort(new ArrayList<>(reports.values()), type);
+     }
+
+    private List<ThreadReport> sort(List<ThreadReport> reports, final String type) {
+        Collections.sort(reports, (a, b) -> {
+           if ("block".equals(type)) {
```
andrewvc: Maybe the if block should define the comparator before sort is invoked? 

It's probably not a big deal for performance but aids in readability. 

I believe with JDK 8 there may be a shortcut here for defining a comparator, maybe with a lambda?

logstash-core/src/main/java/org/logstash/instrument/monitors/HotThreadsMonitor.java
```diff
@@ -0,0 +1,191 @@
+package org.logstash.instrument.monitors;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
+import java.lang.management.ThreadMXBean;
+import java.util.*;
```
andrewvc: We should probably axe this per the ES style guide.

logstash-core/src/main/java/org/logstash/instrument/monitors/MemoryMonitor.java
```diff
@@ -0,0 +1,80 @@
...
+            map.put(PEAK_USED, peak.getUsed());
+            map.put(PEAK_MAX, peak.getMax());
+        }
+     }
+
+     public Report detect(Type selectType){
+        List<MemoryPoolMXBean> beans = ManagementFactory.getMemoryPoolMXBeans();
+        Report report = new Report();
+
+        beans.stream().filter(bean -> (selectType.equals(Type.All))
```
andrewvc: Nice use of lambdas!

logstash-core/src/test/java/org/logstash/instruments/monitors/HotThreadMonitorTest.java
```diff
@@ -0,0 +1,98 @@
+package org.logstash.instruments.monitors;
```
andrewvc: So glad to see these tests added!

suyograo:
@robbavey nice work! please remember to add label versions. this should go into v5.5 and v6.0

suyograo:
Also, when merging, please squash and merge. @andrewvc can help show you how to squash if you haven't done this before.

andrewvc:
LGTM

PR #7168, opened 2017-05-19
#7160 don't run Kafka setup for every IT

### LINKED ISSUE #7160: ITs on Travis at Times Fail to Connect to Docker
We have a few failures on Travis that show this line:

```sh
 CLI > logstash-plugin remove without internet connection (linux seccomp wrapper) when other plugins depends on this plugin refuses to remove the plugin and display the plugin that depends on it.
     Failure/Error: @fixture = Fixture.new(__FILE__)
     
     Docker::Error::UnexpectedResponseError:
```

E.g. https://travis-ci.org/elastic/logstash/jobs/233917467#L2563

Tests working with Docker appear to be broken by the internet disabling `C` hack we're using in some tests. The reason they break the Docker API and that in turn even breaks tests not using Docker is that the Docker API has a global singleton Docker container set up as soon as it's `require`d that acts in the background and will err out if the internet is down (I think it does some heartbeating).

This needs to be stabilized, like in https://github.com/elastic/logstash/pull/7157#issuecomment-302637245 I suggest just dockerizing the Logstash service and disabling internet via Docker networking API calls in a clean way.

original-brownbear:
Fixes #7160 by only interacting with Docker when necessary to set up Kafka.
We  are calling initialize even for services not actively used in a spec, so it was just a mistake on my end to put the Kafka setup into the initialize.

PS: This should also speed up the test suit overall :P

### DIFF
--- a/qa/integration/services/kafka_service.rb
+++ b/qa/integration/services/kafka_service.rb
@@ -3,14 +3,14 @@ require "docker"
 
 class KafkaService < Service
   def initialize(settings)
-    @kafka_image = Docker::Image.build_from_dir(File.expand_path("../kafka_dockerized", __FILE__))
-                     .insert_local(
-                       'localPath' => File.join(TestSettings::FIXTURES_DIR, "how_sample.input"),
-                       'outputPath' => '/')
     super("kafka", settings)
   end
 
   def setup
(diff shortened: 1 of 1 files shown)


suyograo:
hmm, I should have caught this in a previous PR, oh well :)

LGTM

original-brownbear:
@suyograo thanks, merging :)

PR #7169, opened 2017-05-19
MINOR: Remove dead code from org.logstash.ackedqueue.io.AbstractByteBufferPageIO

original-brownbear:
Really minor, but no point in keeping dead fields around.


suyograo:
> but no point in keeping dead fields around.
+100

LGTM

Issue #7178, opened 2017-05-22
Kafka integration test can fail during setup


andrewvc:
Form https://travis-ci.org/elastic/logstash/jobs/234760314

```
  1) Test Kafka Input can ingest 37 apache log lines from Kafka broker
     Failure/Error: @fixture = Fixture.new(__FILE__)
     
     Docker::Error::UnexpectedResponseError:
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/docker-api-1.33.4/lib/docker/util.rb:200:in `extract_id'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/docker-api-1.33.4/lib/docker/image.rb:265:in `build_from_tar'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/docker-api-1.33.4/lib/docker/image.rb:278:in `build_from_dir'
     # ./services/kafka_service.rb:10:in `setup'
     # ./framework/fixture.rb:73:in `setup_services'
     # ./framework/fixture.rb:72:in `setup_services'
     # ./framework/fixture.rb:29:in `initialize'
     # ./specs/kafka_input_spec.rb:12:in `(root)'
```

jsvd:
this seems to be a dup of an issue that got a PR and was closed, I added a comment there: https://github.com/elastic/logstash/issues/7160#issuecomment-303061906

andrewvc:
@jsvd would you like to use this as the new issue or close / reopen #7160 ?

jsvd:
I'll let @original-brownbear decide if this is similar (then close) or a different thing (then keep open).

andrewvc:
I believe https://github.com/elastic/logstash/pull/7179 will be a good fix for this, waiting for those tests to pass ATM

jsvd:
I got this again in https://travis-ci.org/elastic/logstash/jobs/237106207#L1837

original-brownbear:
@jsvd wasn't there some SSL bug in Jruby ? :D The error so looks like some SSL lib issue (admittedly could be a bunch of other things too), I've never seen this one with MRI Ruby ...

original-brownbear:
nevermind ... Travis shouldn't be running an encrypted Docker socket ...

colinsurprenant:
removing elasticsearch-bot @andrewvc can you please reassign?

jsvd:
a few more instances of this:
https://travis-ci.org/elastic/logstash/jobs/237429275#L2256
[...]

PR #7179, opened 2017-05-22
Run offline specs separately from online ones.

andrewvc:
The internet disabling C hack can break other tests, so, we now run those tests separately.

Fixes https://github.com/elastic/logstash/issues/7178

/cc @original-brownbear this isn't as nice as using docker, but its the same effect for a lot less work right now. I think we can triage that work ATM.

### DIFF
--- a/ci/travis_integration_run.sh
+++ b/ci/travis_integration_run.sh
@@ -7,4 +7,10 @@ fi
 
 echo "Running integration tests from qa/integration directory"
 cd qa/integration
-rspec
+
+# The offline specs can break the online ones
+# due to some sideeffects of the seccomp policy interfering with
+# the docker daemon
+# See prepare_offline_pack_spec.rb for details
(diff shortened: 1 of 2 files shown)


qa/integration/.ruby-version
```diff
@@ -0,0 +1 @@
+jruby-1.7.26
```
jsvd: is this necessary for this PR?

andrewvc:
@jsvd I just removed that file and pushed :)

andrewvc:
@jsvd just pushed again with a nice comment explaining the two flags.

jsvd:
> @jsvd just pushed again with a nice comment explaining the two flags.

just to confirm: you mean comment in the commit? I don't see anything in the code

andrewvc:
@jsvd forgot to push :( it's up now :)

PR #7180, opened 2017-05-22
Clarify doc to indicate that offline install expects a file URI

dedemorton:
Addresses issue mentioned here:

### DIFF
--- a/docs/static/offline-plugins.asciidoc
+++ b/docs/static/offline-plugins.asciidoc
@@ -61,11 +61,18 @@ To install an offline plugin pack:
 . Move the compressed bundle to the machine where you want to install the plugins.
 
 . Run the `bin/logstash-plugin install` subcommand and pass in the file URI of
-the offline plugin pack. For example: 
+the offline plugin pack. 
 +
 ["source","sh",subs="attributes"]
+.Windows example:
 -------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


docs/static/offline-plugins.asciidoc
```diff
@@ -60,14 +60,16 @@ To install an offline plugin pack:
...
 
-. Run the `bin/logstash-plugin install` subcommand to install the packaged plugins:
+. Run the `bin/logstash-plugin install` subcommand and pass in the file URI of
+the offline plugin pack. For example: 
 +
 ["source","sh",subs="attributes"]
 -------------------------------------------------------------------------------
-bin/logstash-plugin install file:///path/to/logstash-offline-plugins-{logstash_version}.zip
+bin/logstash-plugin install file:///c:/elastic/install/plugins/logstash-offline-plugins-{logstash_version}.zip
 -------------------------------------------------------------------------------
```
ph: can we have both window and linux example? I've answered a few people on discuss and issues for windows I want to make sure we don't run into the inverse situation.

dedemorton:
@ph I've updated the topic to show examples for Windows and Linux. Do you think it's clear enough, or do we need to show realistic paths? If so...let me know what I should change the paths to. TBH, I'm not sure where users would want to put their offline packs (and really they could put them anywhere, no?)

ph:
@dedemorton All good! :shipit:

Issue #6870, opened 2017-04-03
Test failure: Something failed in the acceptance test (?)


jordansissel:
Build failure: https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-acceptance/label=metal,suite=debian/235/console

```
...
Using rspec-expectations 3.1.2
Using rspec-mocks 3.1.3
Using pry 0.10.4 (java)
Using runner-tool 0.2.0 from https://github.com/purbon/runner-tool.git (at master@5ec0803)
Using rspec 3.1.0
Bundle complete! 4 Gemfile dependencies, 17 gems now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.
Acceptance: Running the tests
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
The signal EXIT is in use by the JVM and will not work correctly on this platform
bootstrapping all non experimental VM's defined in acceptance/Vagrantfile for debian: ["debian-8", "ubuntu-1204", "ubuntu-1404"]
rake aborted!
LogStash::CommandExecutor::CommandError: CMD: vagrant up debian-8 ubuntu-1204 ubuntu-1404 STDERR: The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
/var/lib/jenkins/workspace/elastic+logstash+master+multijob-acceptance/label/metal/suite/debian/qa/vagrant/command.rb:49:in `run!'
/var/lib/jenkins/workspace/elastic+logstash+master+multijob-acceptance/label/metal/suite/debian/qa/vagrant/helpers.rb:21:in `bootstrap'
/var/lib/jenkins/workspace/elastic+logstash+master+multijob-acceptance/label/metal/suite/debian/qa/Rakefile:40:in `(root)'
/var/lib/jenkins/.rbenv/versions/jruby-1.7.26/bin/bundle:23:in `(root)'
Tasks: TOP => qa:vm:setup
(See full trace by running task with --trace)
runbld>>> <<<<<<<<<<<< SCRIPT EXECUTION END <<<<<<<<<<<<
runbld>>> DURATION: 105455ms
```

ph:
fixed in #7187

PR #7187, opened 2017-05-23
Fix the acceptance on Debian family

ph:
This commit fixes an issue when the debian 8 machine was not correctly bootstrapped and made the suite,
We also disable the test introduced by #6879 they need to be redone so they actually test the content fo the generated gem.

Fixes: #7123 #6870

### DIFF
--- a/qa/acceptance/spec/lib/cli_operation_spec.rb
+++ b/qa/acceptance/spec/lib/cli_operation_spec.rb
@@ -20,6 +20,6 @@ describe "CLI operation" do
     it_behaves_like "logstash uninstall", logstash
     it_behaves_like "logstash remove", logstash
     it_behaves_like "logstash update", logstash
-    it_behaves_like "logstash generate", logstash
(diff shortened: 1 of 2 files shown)


ph:
@jsvd I was able to run theses tests multiples times last week with success, the easiest way to debug this locally is to mimic the *CI* environment:

- clean checkout
- ci/ci_acceptance.sh debian
- ci/ci_acceptance.sh redhat

Don't try to run all of them on your laptop, it will just make it slowly die.

qa/acceptance/spec/shared_examples/cli/logstash-plugin/generate.rb
```diff
@@ -13,18 +13,19 @@
   end
 
   describe "on #{logstash.hostname}" do
-
```
jsvd: all changes in this generate.rb file are whitespace related, can we take them out of this PR?

PR #7198, opened 2017-05-24
Add edit URLs back into index.asciidoc

dedemorton:
@clintongormley Not sure if there is a better way to do this, but we need to point the Edit links to the `logstash` repo because `logstash-docs` is meant to be a read-only repo. 

At the moment, the Edit links are all broken in the 5.0+ docs because the Edit URL is weird. For example: https://github.com/elastic/logstash/edit/logstash-docs/docs/static/life-of-an-event.asciidoc

I'm all for fixing this in one place if that's possible. Just not sure how.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -15,7 +15,7 @@ release-state can be: released | prerelease | unreleased
 :jdk:                   1.8.0
 :guide:                 https://www.elastic.co/guide/en/elasticsearch/guide/current/
 :ref:                   https://www.elastic.co/guide/en/elasticsearch/reference/current/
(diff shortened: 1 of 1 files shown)


clintongormley:
@dedemorton you're right, this was broken by the change to include the repo name in the top-level checkout when building docs, and i can't think of any way to fix this.

Ideally, we'd stop using the logstash-docs repo for everything except the generated docs, then we wouldn't need to add these edit_urls.  but for now, this PR LGTM

clintongormley:
@dedemorton we can also remove the dummy README source - it's purpose was to hack around the need for edit_urls, but no longer works correctly

Issue #7196, opened 2017-05-24
[test failure: master] undefined method `converge_state_and_update' for LogStash::Agent


colinsurprenant:
https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-intake/683/console

```
Failures:

  1) LogStash::Agent#execute when auto_reload is true if state is clean should periodically reload_state
     Failure/Error: t = Thread.new { subject.execute }
     NoMethodError:
       undefined method `converge_state_and_update' for #<LogStash::Agent:0x49a18ad4>
     # /var/lib/jenkins/workspace/elastic+logstash+master+multijob-intake/logstash-core/lib/logstash/agent.rb:101:in `execute'
     # /var/lib/jenkins/workspace/elastic+logstash+master+multijob-intake/vendor/bundle/jruby/1.9/gems/stud-0.0.22/lib/stud/interval.rb:18:in `interval'
     # /var/lib/jenkins/workspace/elastic+logstash+master+multijob-intake/logstash-core/lib/logstash/agent.rb:90:in `execute'
     # /var/lib/jenkins/workspace/elastic+logstash+master+multijob-intake/logstash-core/spec/logstash/agent_spec.rb:207:in `(root)'

Finished in 5 minutes 24 seconds (files took 4.12 seconds to load)
2218 examples, 1 failure, 2 pending

Failed examples:

rspec /var/lib/jenkins/workspace/elastic+logstash+master+multijob-intake/logstash-core/spec/logstash/agent_spec.rb:205 # LogStash::Agent#execute when auto_reload is true if state is clean should periodically reload_state

Randomized with seed 29133
```

ph:
This is weird failures; I did multiple run with `stress` and that seed without luck to reproduce it.
I think the problem lies in how the test is written, it uses a thread, and we do mock on the object in another thread.

Also, this test is a duplicate of the one I wrote when I refactored the classes, if you look at the converge spec at https://github.com/elastic/logstash/blob/master/logstash-core/spec/logstash/agent/converge_spec.rb#L131-L156, But this test doesn't use rspec mock, instead it uses a threadsafe spy to make sure we reload multiple time the configuration.

I will make a PR to remove that test from the main `agent_spec` and just use the one in the converge spec.

ph:
Actually, the `converge_spec` is better too because we don't test the implementation details we just test to make sure we call multiple time `#fetch` on the source which translate in multiple reload in LS.

PR #7200, opened 2017-05-24
Removed a flaky test in agent_spec in favor a refactored one in

ph:
converge_spec

We remove a flaky test "LogStash::Agent#execute when auto_reload is true if state is clean should periodically reload_state"
in favor of the refactored test in `converge_spec`, since it doesn't use
rspec mock but instead use a threadsafe spies to make sure we do
multiple call on the source loader.

Fixes: #7196

### DIFF
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -194,28 +194,6 @@ describe LogStash::Agent do
         end
       end
     end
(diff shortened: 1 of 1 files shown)


ph:
link to the replacement test https://github.com/elastic/logstash/blob/master/logstash-core/spec/logstash/agent/converge_spec.rb#L131-L156

Issue #6511, opened 2017-01-09
mem type listed in Node Stats API documentation


inqueue:
Node Stats API documentation lists a `mem` type for collecting memory usage stats which does not exist as an endpoint. 

Mac OS X:
```
curl 'localhost:9600/_node/stats/mem'
{"path":"/_node/stats/mem","status":404,"error":{"message":"Not Found"}}
```

It looks like the docs just need updated since memory stats fall under JVM.

dedemorton:
Closing because this is already fixed in 5.2 and later. Opened PR #7201 for 5.0 and 5.1 docs.

PR #7201, opened 2017-05-24
Remove mem type from node stats API docs

dedemorton:
Cleanup to resolve https://github.com/elastic/logstash/issues/6511 in 5.0 and 5.1 docs. The issue is already fixed in 5.2 and later.


suyograo:
LGTM

PR #7208, opened 2017-05-25
Change API docs to use curl examples instead of console syntax

### LINKED ISSUE #6823: Provide curl examples for Logstash monitoring and logging api calls in doc
There are users out there who haven't interacted with Logstash's new APIs (logging and monitoring) before.  It will be helpful to provide curl examples similar to Elasticsearch documentation.  

- Currently, our documentation provides examples using Sense/Console notation, eg. `GET /`, `GET /_node/stats/jvm`.  Providing curl examples will be more useful - esp. since the Kibana Console no longer allows configuring arbitrary URLs to send http requests to.
- For users who land [here](https://www.elastic.co/guide/en/logstash/current/logging.html#_logging_apis) directly, there is no information on where to send these logging API calls to (i.e. to the http_port that defaults to 9600).  For completeness, it will be helpful to have a similar note as the `Note` on the [monitoring api](https://www.elastic.co/guide/en/logstash/current/monitoring.html) page.

dedemorton:
Summary of changes:
- Fixed #6823 by changing examples to use curl instead of console syntax.
- Updated a few of the examples (mostly minor changes to make them look up-to-date).
- Removed statement about detailed pipeline stats for input plugins being unavailable because, well, it looks like they are.

### DIFF
--- a/docs/static/logging.asciidoc
+++ b/docs/static/logging.asciidoc
@@ -57,15 +57,22 @@ The logs will include the full event and filter configuration that are responsib
 
 You could modify the `log4j2.properties` file and restart your Logstash, but that is both tedious and leads to unnecessary 
 downtime. Instead, you can dynamically update logging levels through the logging API. These settings are effective 
-immediately and do not need a restart. To update logging levels, take the subsystem/module you are interested in and prepend 
+immediately and do not need a restart.
+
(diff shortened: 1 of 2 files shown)


suyograo:
LGTM

Issue #6904, opened 2017-04-11
Test failure: Plugin pack fails due to upstream network problem


jordansissel:
Build failure: https://logstash-ci.elastic.co/job/elastic+logstash+5.3+dockerubuntu-periodic/625/console

```

Failures:

  1) LogStash::PluginManager::OfflinePluginPackager when the plugins exist with wildcards and normal plugins creates a pack with the plugins
     Failure/Error: subject.package(plugins_args, target)
     IOError:
       Connection reset by peer
```

original-brownbear:
Trying this one :)

andrewvc:
Freak accident

andrewvc:
@colinsurprenant saw this [here](https://travis-ci.org/elastic/logstash/jobs/235874652), reopening. I've seen it a few other times as well.

I'm confused by this @ph, aren't these tests supposed to all pass offline anyway?

ph:
@andrewvc We actually generate a package in that test, so they require to be online. I will take a stab a it right now. I've seen it too in the email.

andrewvc:
@ph I saw three over the last 24 hours for some reason. Keep in mind, we now run the offline specs [separately](https://github.com/elastic/logstash/blob/master/ci/travis_integration_run.sh#L14)

andrewvc:
@ph apparently I am confus

ph:
Just to clarify, theses are the test faillings https://github.com/elastic/logstash/blob/master/spec/unit/plugin_manager/offline_plugin_packager_spec.rb#L102-L134

PR #7211, opened 2017-05-25
Refactor offline plugin test

ph:
We have seen unreliable run in the past on travis or jenkins, in that test we actually test the creation of a physical package and we require internet access to correctly download the gems, we will now retries IOError.

Also, the scope of *wildcard* downloading test was reduced this has the effect of making the test faster  by downloading less gems and limit the risk of running into a network error.

Fixes: #6904

### DIFF
--- a/spec/unit/plugin_manager/offline_plugin_packager_spec.rb
+++ b/spec/unit/plugin_manager/offline_plugin_packager_spec.rb
@@ -106,7 +106,7 @@ describe LogStash::PluginManager::OfflinePluginPackager do
     end
 
     context "with wildcards" do
-      let(:plugins_args) { ["logstash-filter-*"] }
+      let(:plugins_args) { ["logstash-filter-x*"] }
 
       it "creates a pack with the plugins" do
         expect(retrieve_packaged_plugins(extract_to).size).to eq(LogStash::PluginManager::SpecificationHelpers.find_by_name_with_wildcards(plugins_args.first).size)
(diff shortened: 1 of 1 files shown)


spec/unit/plugin_manager/offline_plugin_packager_spec.rb
```diff
@@ -67,7 +70,8 @@ def retrieve_dependencies_gems(path)
     before do
       FileUtils.mkdir_p(temporary_dir)
 
-      subject.package(plugins_args, target)
+      Stud.try(retries_count.times, retries_exceptions)  { subject.package(plugins_args, target) }
```
jsvd: I suggest adding a comment here mentioning that because the `package` method reaches the internet it's possible to see an IOError, so we've added a retry block.

spec/unit/plugin_manager/offline_plugin_packager_spec.rb
```diff
@@ -40,6 +41,8 @@ def retrieve_dependencies_gems(path)
   let(:temporary_dir) { Stud::Temporary.pathname }
   let(:target) { ::File.join(temporary_dir, "my-pack.zip")}
   let(:extract_to) { Stud::Temporary.pathname }
+  let(:retries_count) { 20 }
```
jsvd: What do you think of bumping this to 30 or 50? the first few retries are subsecond, so 20 means around 10-15 seconds of retry time, which may not be enough?

ph:
updated / rebased with your comments.

PR #7213, opened 2017-05-25
Feature Modules, move all inner classes to their own folder + client and importer

guyboertje:
Includes latest commits from master.


guyboertje:
I am going to split this into two PRs 1) commits from master and 2) my changes

PR #7228, opened 2017-05-25
Fixes and tests

untergeek:
Add tests for the `LogStash::Modules:CLIParser` class in `cli_parser.rb`
Fix a typo in `cli_parser.rb` (`uparsed` vs `unparsed`)
Fix a bad variable name found by testing in `cli_parser.rb` and update the error message accordingly in `en.yml`


untergeek:
Had to go back and change a few things after @guyboertje moved things around.  

This is the first batch of tests.  More to follow.

Issue #7155, opened 2017-05-18
[proposal] Strict Schema For Metrics


andrewvc:
Related issues:

https://github.com/elastic/logstash/issues/7227

We have a lot of test failures of the form `https://github.com/elastic/logstash/issues/7154`, we also can easily have errors when writing code due to typos in metric names. 

I think that we should revisit our metrics API and change it to make it more like our stats API where a metric must be first declared, then used.

Additionally, we have metrics scattered about all over the code. I also propose that we centralize metrics in 2 separate objects `GlobalMetric`, `PipelineMetric`, and `PluginMetric`. Plugin metric would still allow freeform metrics to be added.

I propose that when we do this refactor we try and define our metrics in only three places:

1. At Agent start (all global metrics)
2. At pipeline start (all metrics under the pipeline namespace)
3. At plugin register

These would all be done by convention. So, the current way of doing metrics involves code like:

```
class Pipeline
  def do_thing
    @metric.increment(:thing_happened, 1)
  end
end
```

would turn into

```
class Pipeline
  def initialize(metrics)
    @pipeline_metrics = PipelineMetrics.new(metrics)
  end

  def do_thing
  end
end

class PipelineMetrics
  def initialize(metrics, pipeline)
    setup_metrics(pipeline)
  end

  def setup_metrics
    @thing_happened_metric = metrics.declare(:stats, :pipelines, pipeline.id, :thing_happened)
  end

  def thing_happened
    @thing_happened_metric.increment(1)
  end
end
```

The nice thing about adding the extra class is that we can centralize the instantiation of metrics and avoid the problem of tests checking for metrics before whatever weird pipeline behavior, like a codec starting, happens. Querying for, for instance, the `queue_push_duration` metric, right now requires an input to have been started before it exists. This fixes that.

ph:
I am all + 1 for that but I think we should use the same contract even with plugin author that they need to have declared their metric before they can use it. Also this is a good moment to javafy theses classes.


Concerning the tests, I think we should evaluate using the `json-schema` to make better assert on the results.

ph:
@andrewvc I think we might also need to come up with a metric observer to collect from other parts of Logstash like the `read_cilent/write_client`, we monitor the time spend on pushing to the queue. 

Or we extract that time from the instrumented client and make it a concerns of the pipeline?

andrewvc:
I like the idea of observers @ph! We should definitely not have the duplication we have in the read /  write client.

[...]

PR #7229, opened 2017-05-26
Javaify individual metrics types

andrewvc:
This is in preparation for https://github.com/elastic/logstash/issues/7155

We will want to make those objects for GlobalStats etc. Java objects, so we'll need to javaify these classes first. We're breaking up this workinto multiple stand-alone commits to make this easier.

I initially thought we'd need to move `TimedExecution` to this, but I don't think that makes sense anymore. I think it makes sense to do that in the next phase, and see how that plays out with the larger java 

This moves all metrics from ruby Numeric to java Double. This is more appropriate for a number of reasons, but primarily it makes dealing with numbers externally a lot easier. You can know that metrics are either strings or (double) numbers.

### DIFF
--- a/logstash-core/lib/logstash/instrument/metric_type/base.rb
+++ b/logstash-core/lib/logstash/instrument/metric_type/base.rb
@@ -2,30 +2,8 @@
 require "logstash/util"
 
 module LogStash module Instrument module MetricType
-  class Base
-    attr_reader :namespaces, :key
-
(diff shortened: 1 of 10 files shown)


ph:
Awesome work!

The downside of using a double will mean that the return response will break. 

ie: 5.x events_in => 100, 6.x => 100.0

andrewvc:
Closed in favor of #7274

PR #7234, opened 2017-05-26
ci: dockerized elastisearch service.

### LINKED ISSUE #7097: Dockerize ElasticSearch for ITs

wainersm:
This change makes the elasticsearch service used
in integration tests be installed and started in
a docker container.

Fixes #7097


qa/integration/services/dockerized/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+##
+# Define a base image for all service images.
+##
+
+ENV _JAVA_OPTIONS "-Djava.net.preferIPv4Stack=true"
+ENV TERM=linux
+
+RUN apt-get update && apt-get install -y curl openjdk-8-jre-headless netcat
+
+RUN adduser --disabled-password --gecos '' tester
```
original-brownbear: @wainersm do we need a user other than root here, maybe it's easier to just run ES with `-Des.insecure.allow.root=true` (I don't have a strong argument here (yet) but we're seeing some issues with testing various networking things, that will probably require running shell scripts from Ruby through the Docker api, would be nice if we had `root` as a user throughout here if possible) ?

qa/integration/services/service_container.rb
```diff
@@ -0,0 +1,77 @@
+require_relative "service"
+require "docker"
+
+# Represents a service running within a container.
+class ServiceContainer < Service
+
+  # Base image used by all service containers.
+  @@base_image = nil
```
original-brownbear: @wainersm let's not use any mutable `static/class` field for this:

1. (medium) Static global mutable is kind of dirty and could lead to a lot of bugs if the correct exec order is not met :)
2. (mayor) This creates a hard dependency between the Docker service and our tests. I think given that we have tiny Docker contexts to be build, it's fine to let Docker rebuild (which will effectively just mean hash the service directory contents in a few `ms`) on every service setup (mainly worried about the case of running this against Docker swarm here, which may or may not happen, let's stay a little flexible here :)).

wainersm: Agreed.

qa/integration/services/dockerized/Dockerfile
wainersm: It turns out that ```es.insecure.allow.root``` property no longer works in ES 5.x. I found people commenting about that but I could not see the information on ES manual.

Indeed, it does not work:
```
root@56a394927ff8:~# ./elasticsearch/bin/elasticsearch --version
Picked up _JAVA_OPTIONS: -Djava.net.preferIPv4Stack=true -Des.insecure.allow.root=true
Version: 5.0.1, Build: 080bb47/2016-11-11T22:08:49.812Z, JVM: 1.8.0_131
root@56a394927ff8:~# ./elasticsearch/bin/elasticsearch                              
Picked up _JAVA_OPTIONS: -Djava.net.preferIPv4Stack=true -Des.insecure.allow.root=true
[2017-05-31T15:07:01,000][WARN ][o.e.b.ElasticsearchUncaughtExceptionHandler] [] uncaught exception in thread [main]
org.elasticsearch.bootstrap.StartupException: java.lang.RuntimeException: can not run elasticsearch as root
	at org.elasticsearch.bootstrap.Elasticsearch.init(Elasticsearch.java:116) ~[elasticsearch-5.0.1.jar:5.0.1]
	at org.elasticsearch.bootstrap.Elasticsearch.execute(Elasticsearch.java:103) ~[elasticsearch-5.0.1.jar:5.0.1]
	at org.elasticsearch.cli.SettingCommand.execute(SettingCommand.java:54) ~[elasticsearch-5.0.1.jar:5.0.1]
	at org.elasticsearch.cli.Command.mainWithoutErrorHandling(Command.java:96) ~[elasticsearch-5.0.1.jar:5.0.1]
	at org.elasticsearch.cli.Command.main(Command.java:62) ~[elasticsearch-5.0.1.jar:5.0.1]
	at org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:80) ~[elasticsearch-5.0.1.jar:5.0.1]
	at org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:73) ~[elasticsearch-5.0.1.jar:5.0.1]
Caused by: java.lang.RuntimeException: can not run elasticsearch as root
	at org.elasticsearch.bootstrap.Bootstrap.initializeNatives(Bootstrap.java:96) ~[elasticsearch-5.0.1.jar:5.0.1]
	at org.elasticsearch.bootstrap.Bootstrap.setup(Bootstrap.java:155) ~[elasticsearch-5.0.1.jar:5.0.1]
	at org.elasticsearch.bootstrap.Bootstrap.init(Bootstrap.java:286) ~[elasticsearch-5.0.1.jar:5.0.1]
	at org.elasticsearch.bootstrap.Elasticsearch.init(Elasticsearch.java:112) ~[elasticsearch-5.0.1.jar:5.0.1]
	... 6 more
```

Unless I am missing something here, I can see 3 solutions:
1. Stick with first implementation and services that need ```root``` can use `su` in their scripts.
2. Do the other way around: the docker base image comes with ```root``` only, the services that requires non-root will be in charge of creating an user.
3. A variation of case 2: the docker base image setup a regular user along with ```root``` (which defaults), the services requiring non-root simply switch to the regular user.

Notice that either creating or switching to a non-root need to be carried out in scripts since the ```ONBUILD``` commands are the first executed.

original-brownbear:
@wainersm thanks a lot for the update, just wanted to let you know that I'm traveling right now so I'll get back to this in about 36h :)

qa/integration/services/dockerized/Dockerfile
original-brownbear: @wainersm nevermind then, option `1.` is just fine then, my mistake :)

qa/integration/services/service_container.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def initialize(name, settings)
+    super(name, settings)
+
+    @base_image_context = File.expand_path("../dockerized", __FILE__)
+
+    @image_context = File.join(@base_image_context, @name)
+
+    # Options to create the container.
+    @container_create_opts = {}
```
original-brownbear: @wainersm are we using this anywhere yet?

original-brownbear:
@wainersm sorry for the stupid delay, thought this would need more cycles than it actually needed to review.
1. Commented on the `USER` situation above -> all good there from my end :)
2. Let's fix the static variable as agreed.
3. Take a look at my new comment on the `@container_create_opts = {}` line (this comment can be applied to other similar lines as well), do we need that level of generality yet? :)

=> all in all we're very close to happy here imo. Let me know if you need any help with any of the details! :)

qa/integration/services/service_container.rb
wainersm: Removed in the last patch. I was going to used it in a ```@container.kill``` call but I decided to use ```stop```.

wainersm:
Hi @original-brownbear ! Thanks again for the reviews. With this last patch I think we are ready to merge.

As soon as it gets merged I can work on adapt Filebeat and Kafka to that new design, it ought be straightforward. Unless someone else want to work on them for learning or improve purposes, of course. :)

qa/integration/services/service_container.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    # Options to delete the container.
+    @container_delete_opts = {:force => true, :volumes => true}
```
original-brownbear: @wainersm I think this shouldn't be a field and just be inlined to the `stop_container` method. Cleaning up volumes and `force` killing containers in the shutdown isn't optional.
Also, I don't think we should create a situation where a container has to "clean up" after itself ever. That will just get very confusing :)

qa/integration/services/service_container.rb
original-brownbear: @wainersm but it's still there? Sorry I think I misunderstood the above statement?

original-brownbear: @wainersm nevermind now I understand this, sorry my bad ;)

original-brownbear:
@wainersm alrighty looks good, can you take a look at my point about `container_delete_opts`? After that I think we're good, I'll test this locally and unless I find something troublesome we can merge :)

qa/integration/services/service_container.rb
wainersm: Indeed I was over-designing with that field. In the future, if ever needed, we can manage to add such as field back but keep force/clean always in.

wainersm:
Hi @original-brownbear ! Last patch contain all changes you suggested. It worked out locally but seems to fail in travis.

original-brownbear:
@wainersm hmm the failure is a known test failure I think :( damn ... we thought we had fixed this already. Will very soon look into merging here though, let me just wake up properly :D 

@jsvd fyi :( (his branch has the Docker version upgrade, will reopen issue)
https://travis-ci.org/elastic/logstash/jobs/238786856#L1900

original-brownbear:
can't merge this until #7304 is resolved :(

original-brownbear:
@wainersm merged at last! thanks so much for this one :)!

PR #7236, opened 2017-05-26
Make SafeURI 9k compatible by using java.net.URI internally

andrewvc:
Ruby 2.0's stdlib no longer allows relative IPv6 URLs like: `//[::1]`. This is a key feature of the Elasticsearch output. See more in https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/604

The only way to fix this is to use a better URI class which means introducing a slight incompatibility here. I don't believe that anyone is using the URI class other than the ES output, and I don't that there is anyone using any portions of the API here that are no longer compatible.

Normally we would only make such a change in a major version, but our hands are tied here. If we are to go with JRuby 9k, which is critical for support we must use a different, better, library.

java.net.URI is stable and works well, so its probably superior in the long run anyway.

### DIFF
--- a/logstash-core/lib/logstash/util/safe_uri.rb
+++ b/logstash-core/lib/logstash/util/safe_uri.rb
@@ -11,20 +11,21 @@ class LogStash::Util::SafeURI
   
   extend Forwardable
   
-  def_delegators :@uri, :coerce, :query=, :route_from, :port=, :default_port, :select, :normalize!, :absolute?, :registry=, :path, :password, :hostname, :merge, :normalize, :host, :component_ary, :userinfo=, :query, :set_opaque, :+, :merge!, :-, :password=, :parser, :port, :set_host, :set_path, :opaque=, :scheme, :fragment=, :set_query, :set_fragment, :userinfo, :hostname=, :set_port, :path=, :registry, :opaque, :route_to, :set_password, :hierarchical?, :set_user, :set_registry, :set_userinfo, :fragment, :component, :user=, :set_scheme, :absolute, :host=, :relative?, :scheme=, :user
   
   attr_reader :uri
-  
+
   public
   def initialize(arg)    
     @uri = case arg
            when String
              arg = "//#{arg}" if HOSTNAME_PORT_REGEX.match(arg)
-             URI.parse(arg)
-           when URI
+             java.net.URI.new(arg)
+           when java.net.URI
              arg
+           when URI
+             java.net.URI.new(arg.to_s)
            else
-             raise ArgumentError, "Expected a string or URI, got a #{arg.class} creating a URL"
+             raise ArgumentError, "Expected a string, java.net.URI, or URI, got a #{arg.class} creating a URL"
            end
   end
 
@@ -37,11 +38,11 @@ class LogStash::Util::SafeURI
   end
 
   def sanitized
-    return uri unless uri.password # nothing to sanitize here!
+    return uri unless password # nothing to sanitize here!
     
-    safe = uri.clone
-    safe.password = PASS_PLACEHOLDER
-    safe
+    user_info = user ? "#{user}:#{PASS_PLACEHOLDER}" : nil
+
+    make_uri(scheme, user_info, host, port, path, query, fragment)
   end
 
   def ==(other)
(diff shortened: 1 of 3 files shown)


logstash-core/lib/logstash/util/safe_uri.rb
```diff
@@ -37,20 +38,117 @@ def inspect
...
 
   def clone
-    cloned_uri = uri.clone
-    self.class.new(cloned_uri)
+    # No need to clone the URI, in java its immutable
+    self.class.new(uri)
+  end
+
+  def update(field, value)
+    parts = {
```
jsvd: not sure it's worth creating this entire hash just to use its individual items later:
```ruby
@uri = java.net.URI.new(parts[:scheme], user_info, parts[:host], parts[:port], parts[:path], parts[:query], parts[:fragment])
```

andrewvc: Fair point, I'll change it

logstash-core/lib/logstash/util/safe_uri.rb
```diff
@@ -37,20 +38,132 @@ def inspect
...
+      new_path = field
+    when :query
+      new_query = field
+    when :fragment
+      new_fragment = field
+    end
+
+    user_info = new_user
+    if new_user && new_password
+      user_info += ":" + new_password
```
jordansissel: thoughts on uri escaping for user and passwords?

andrewvc: @jordansissel its a good point! It gets tricky though. I'm trying to keep compatibility so code like: https://github.com/logstash-plugins/logstash-output-elasticsearch/blob/808d73dd198255ef1439b65e33f5f93ff02c4706/lib/logstash/outputs/elasticsearch/http_client_builder.rb#L141 works.

The problem is I want this patch to behave like the old code used to, so that the ES output can stay compatible with old logstashes.

I don't like the lack of escaping, but I can't figure out a way out of it that isn't so complex that it isn't worth our time right this minute.

andrewvc: Thinking through this a bit more, I think its not a good idea to escape the user/password.

Both the java.net.URI and ruby's URI class don't do this and I think it makes sense not to include this feature.

I imagine using a URI to represent, say, a rabbitmq connection. When using the rabbitmq libs you wouldn't want to do the encoding, that library would do it for you, as do some HTTP libraries. If you need the raw password in that case you have to unescape it or use a special method.

andrewvc:
Tests are passing (aside from some unrelated integration tests).

@jsvd @jordansissel can I get a re-review

jordansissel:
Because this changes an API used by a plugin (or more clearly, https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/605 depends on this pr), this will require a major plugin version bump, will it not?

logstash-core/lib/logstash/util/safe_uri.rb
jordansissel: It is a good idea to escape the user+password (and required). Ideally, the URI class should be doing this, but it does not.

The slight difference between Ruby and Java here is that:
* Ruby URI has a `user` and `password` field
* Java URI only has `userinfo`

Ruby's should escape, but java's is more the "what goes before the @ sign" and it should *still* escape at least the '@' -- though perhaps there is some RFC recitation that would tell me why we shouldn't do this, hehe

Anyway, I tested Ruby URI to remind myself what happens:

```
>> a.user = "hello:world"
=> "hello:world"
>> a.password = "hello:world"
=> "hello:world"
>> puts a
http://hello:hello:world@google.com
```

At best, I would suggest that Ruby URI has some interesting bugs.

As my original comment assumed Ruby URI to be doing things that I have found it does not actually do, I withdraw it and agree with your assessment that the previous implementation does not escape either.

andrewvc:
@jordansissel technically yes, but I would argue that it would be a net negative thing to do.

The only plugin in LogstashPlugins that uses this class is the ES output and I'm fairly sure no one outside of the LS team dug deep enough to find this class to use in their own code. https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/605 can work with or without this patch (on jruby 1.7.x).

This means the number of people affected by this issue is:

1. People who would upgrade Logstash to a 9k version but then downgrade the ES output.
2. People who are using the SafeURI class AND invoke methods that had to be removed as part of this PR

I'm pretty sure that the number of people that satisfy either of those conditions is zero. On a practical level bumping the core plugin API makes maintenance much more difficult because all plugins then need to maintain multiple branches (although its possible most plugins would support across two major versions).

Your thoughts?

PR #7240, opened 2017-05-27
#7239 fix timeunit on cpu time metric

### LINKED ISSUE #7239: process.cpu.time_in_millis multiplied by 1M
- Version: 5.4.0
- Operating System: Linux
- Sample Data / Steps to Reproduce:

It looks like `process.cpu.total_in_millis` time reported in monitoring API is erroneously  multiplied by 1,000,000. My logstash says it's been using ~14 days of CPU time every second. Value seems to be perfectly correct after dividing by 1M.

```
$ curl -s 'http://127.0.0.1:9600/_node/stats?timeout=1s&pretty' | grep -A1 cpu ; sleep 1 ; curl -s 'http://127.0.0.1:9600/_node/stats?timeout=1s&pretty' | grep -A1 cpu
    "cpu" : {
      "total_in_millis" : 454039020000000,
    "cpu" : {
      "total_in_millis" : 454043540000000,
```

original-brownbear:
Trivial time unit issue, fixes #7239.

Lame excuse for no test: I saw there was some instability in the metrics tests and adjustments being done. Since this is a prime candidate for an unstable test I didn't want to pile on for a trivial thing like this.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/instrument/monitors/ProcessMonitor.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/monitors/ProcessMonitor.java
@@ -1,13 +1,12 @@
 package org.logstash.instrument.monitors;
 
 import com.sun.management.UnixOperatingSystemMXBean;
-
-import javax.management.MBeanServer;
 import java.lang.management.ManagementFactory;
 import java.lang.management.OperatingSystemMXBean;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM.

suyograo:
@original-brownbear this will not backport nicely to 5.x since the impl is in JRuby. We could handle it separately.

original-brownbear:
@suyograo kk, I'll merge and comment in the issue accordingly

PR #7245, opened 2017-05-29
[WIP] #5791 removed gradle wrapper checked in jars

### LINKED ISSUE #5791: Remove gradle jar from git
I believe we should not commit compiled artifacts into the source tree.

On the other hand, with gradle, it's not easy to install the correct version. For example, Fedora 23 (very recent!) only makes Gradle 2.5 available, and I think our projects require Gradle 3.

Proposal: Remove the gradle jar from git and write a rake task that will download it during `rake bootstrap`

original-brownbear:
Closes #5791 

* Removed Gradle binary files, wrappers can simply download Gradle by themselves.
* Updated wrapper scripts to the most recent version (as generated by the Gradle `wrapper` task)


original-brownbear:
nevermind commenting in the issue

Issue #7251, opened 2017-05-29
[test failure: 5.4] LogStash::PluginManager::OfflinePluginPackager - OpenSSL::SSL::SSLError: Connection reset by peer


colinsurprenant:
https://logstash-ci.elastic.co/job/elastic+logstash+5.4+dockerubuntu-periodic/554/console

```
Failures:

  1) LogStash::PluginManager::OfflinePluginPackager when the plugins exist multiples plugins creates pack with the plugins
     Failure/Error: Stud.try(retries_count.times, retries_exceptions)  { subject.package(plugins_args, target) }
     OpenSSL::SSL::SSLError:
       Connection reset by peer
     # ./vendor/bundle/jruby/1.9/gems/webmock-1.21.0/lib/webmock/http_lib_adapters/net_http.rb:109:in `request'
     # ./vendor/bundle/jruby/1.9/gems/paquet-0.2.1/lib/paquet/utils.rb:21:in `download_file'
     # ./vendor/bundle/jruby/1.9/gems/paquet-0.2.1/lib/paquet/gem.rb:106:in `download_gem'
     # ./vendor/bundle/jruby/1.9/gems/paquet-0.2.1/lib/paquet/gem.rb:48:in `package_gems'
     # ./vendor/bundle/jruby/1.9/gems/paquet-0.2.1/lib/paquet/gem.rb:42:in `package_gems'
     # ./vendor/bundle/jruby/1.9/gems/paquet-0.2.1/lib/paquet/gem.rb:33:in `pack'
     # ./lib/pluginmanager/offline_plugin_packager.rb:88:in `execute'
     # ./lib/pluginmanager/offline_plugin_packager.rb:115:in `package'
     # ./spec/unit/plugin_manager/offline_plugin_packager_spec.rb:75:in `(root)'
     # ./vendor/bundle/jruby/1.9/gems/stud-0.0.22/lib/stud/try.rb:79:in `try'
     # ./vendor/bundle/jruby/1.9/gems/stud-0.0.22/lib/stud/try.rb:95:in `try'
     # ./vendor/bundle/jruby/1.9/gems/stud-0.0.22/lib/stud/try.rb:91:in `try'
     # ./vendor/bundle/jruby/1.9/gems/stud-0.0.22/lib/stud/try.rb:123:in `try'
     # ./spec/unit/plugin_manager/offline_plugin_packager_spec.rb:75:in `(root)'
[...]

PR #7254, opened 2017-05-29
Add SSLError to the list of retryable exception

ph:
When building an offline pack we do a real network call to rubygems to
test the structure of the created package, something the network can be
a bit unreliable and SSLError can be raised, when theses errors happen
we will retry them.

Fix: #7251


original-brownbear:
@ph LGTM :)

Issue #7256, opened 2017-05-29
Test org.logstash.ackedqueue.QueueTest#randomAcking is Leaking FDs (still)


original-brownbear:
Just working on https://github.com/elastic/logstash/issues/7184 and running that test in a loop I ran into this:

```sh
java.nio.file.FileSystemException: /Users/brownbear/src/logstash/logstash-core/.lock: Too many open files in system

|tat sun.nio.fs.UnixException.translateToIOException(UnixException.java:91)
|tat sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
|tat sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
|tat sun.nio.fs.UnixFileSystemProvider.newFileChannel(UnixFileSystemProvider.java:177)
|tat java.nio.channels.FileChannel.open(FileChannel.java:287)
|tat java.nio.channels.FileChannel.open(FileChannel.java:335)
|tat org.logstash.FileLockFactory.obtainLock(FileLockFactory.java:84)
|tat org.logstash.ackedqueue.Queue.open(Queue.java:169)
|tat org.logstash.ackedqueue.QueueTest.randomAcking(QueueTest.java:310)
|tat sun.reflect.GeneratedMethodAccessor3.invoke(Unknown Source)
|tat sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
|tat java.lang.reflect.Method.invoke(Method.java:498)
|tat org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
|tat org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
|tat org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
|tat org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
|tat org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
|tat org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
|tat org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:48)
|tat org.junit.rules.RunRules.evaluate(RunRules.java:20)
|tat org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
[...]

PR #7257, opened 2017-05-29
#7256 close channel underlying the queue's Filelock

original-brownbear:
Very straightforward.

A Java `FileLock` is obtained from a `Channel` but the relationship is (for most platforms at least).

Close Channel => Release Lock (this isn't true for all platforms)
Release Lock => No effect on Channel

We were never closing the channel underlying the lock => leaking one FD per Queue.

Fixed by safely ensuring close of the underlying channel.

Works (more loops run through than my OSX allows me FDs :P):

screen shot 2017-05-29 at 23 34 38


jordansissel:
LGTM.

Interesting to me: the Lucene implementation (NativeFSLock) which we borrowed from uses try-with-resources for closing the channel and the lock. https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/store/NativeFSLockFactory.java#L190-L194 (no change required for us, just reporting something learned from lucene's implementation)

PR #7258, opened 2017-05-29
MINOR: Remove unused function from Page

original-brownbear:
This one is not used anywhere -> no need to keep it around.


jordansissel:
```
% ack firstUnackedPageNumFromQueue
logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
157:    protected int firstUnackedPageNumFromQueue() {
```

Only this entry found. Seems ok to remove.

jordansissel:
LGTM

PR #7259, opened 2017-05-29
MINOR: Cleanup of org.logstash.FieldReference

original-brownbear:
Few obvious ones in this class:

* Regex should use compiled pattern
* `EMPTY_STRINGS` was a very inefficient approach to filtering empty strings, let's just not copy them to the list in the first place and avoid that extra loop
* Avoid creating redundant `ArrayList` in `new ArrayList(Arrays.asList(reference.split("[\\[\\]]")));`, just copy straight to the array

### DIFF
--- a/logstash-core/src/main/java/org/logstash/FieldReference.java
+++ b/logstash-core/src/main/java/org/logstash/FieldReference.java
@@ -1,17 +1,17 @@
 package org.logstash;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
-
+import java.util.regex.Pattern;
 // TODO: implement thread-safe path cache singleton to avoid parsing
 
 public class FieldReference {
 
+    private static final Pattern SPLIT_PATTERN = Pattern.compile("[\\[\\]]");
+
     private List<String> path;
     private String key;
     private String reference;
-    private static List<String> EMPTY_STRINGS = Arrays.asList("");
 
     public FieldReference(List<String> path, String key, String reference) {
         this.path = path;
@@ -32,8 +32,13 @@ public class FieldReference {
     }
 
     public static FieldReference parse(String reference) {
-        List<String> path = new ArrayList(Arrays.asList(reference.split("[\\[\\]]")));
-        path.removeAll(EMPTY_STRINGS);
+        final String[] parts = SPLIT_PATTERN.split(reference);
+        List<String> path = new ArrayList<>(parts.length);
+        for (final String part : parts) {
+            if (!part.isEmpty()) {
+                path.add(part);
+            }
+        }
         String key = path.remove(path.size() - 1);
         return new FieldReference(path, key, reference);
     }


jordansissel:
Code change seems ok, but the performance claim is made without providing data. Can you provide?

original-brownbear:
@jordansissel I could add a JMH for this, think it's worth it. I mean, basically I do the same computations as before, but with fewer allocations => there's no way it's not faster, by how much probably depends on how stretched your memory is already.

original-brownbear:
@jordansissel actually I was confused about PRs :D I actually also do a ton fewer computations here obviously, saving recompiling the `Pattern` over and over.

jordansissel:
I understand the implications of compiling the pattern once. 

What I mean is this doesn't appear to be an 'obvious performance improvement' -- is `FieldReference.parse()` part of a critical path? Would a user see such an obvious improvement?

From what I remember, there is a global cache (PathCache) that memoizes FieldReference instances (produced by FieldReference.parse) so any given FieldReference string `"[foo][bar]"` is only ever parsed once during Logstash's lifetime. This is part of what I mean in terms of 'obvious performance improvement' -- Yes, this has improvements, but the description ('obvious' etc) seems misleading unless this change has a positive impact for users. A user might see this change and get excited about a performance improvement that they may never actually observe because PathCache mitigates it with caching. Thoughts?

jordansissel:
I like the proposed change, I am just not agreeing with the description of the change and the claimed result of the change. If we can get the description to be more accurate (or update my mental model of how FieldReference.parse is invoked to be more accurate, if I am mistaken), that would help.

original-brownbear:
@jordansissel phew good one, not sure how visible this change will be performance wise given that we have the cache. Change title to `Cleanup Redundant Allocations org.logstash.FieldReference` ? :)

jordansissel:
+1 to calling it a cleanup -- something to indicate that the regexp is now only compiled one instead of previously compiling every time `parse` is invoked.

As for the ArrayList change, I would indicate what your intent appears to be which is to avoid the `removeAll` call (but it's unclear if this has any measurable benefit for the user). If the intent is to remove the copy `new ArrayList(Arrays.asList(...))` then we could just remove the `new ArrayList` part and use the List returned by `Arrays.asList`, no?

original-brownbear:
@jordansissel unfortunately we can't use the return of `Arrays.asList` because it returns in immutable `List` and fails down the stack (tried that first) :)

original-brownbear:
@jordansissel title fixed :)

original-brownbear:
@jordansissel btw. on the mutable requirement for that `List<>` ... urgh :) That should probably be the next step in cleaning house ... if we want to have a mutable list filled here we should pass it to this method as a buffer, not require that the method returns a special implementation of the interface that it uses as a return type (imo), right?

original-brownbear:
@jordansissel ping :)

jordansissel:
The commit message still claims "obvious performance" and not "cleanup" -- can you update the commit message?

original-brownbear:
@jordansissel sure sec :)

original-brownbear:
@jordansissel commit message adjusted :)

original-brownbear:
@jordansissel good to go here?

PR #7261, opened 2017-05-30
MINOR: Fix missing static identifiers and save some memory

original-brownbear:
We had a few spots where `static` methods were not actually `static`, fixed here :)

Reviewing is really trivial, the fact that we only have additions of the modifier but the method body changes and it still compiles proves there were no instances references in the methods, to begin with.
=> Let's save some allocations and make the JITs life one step easier.

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -288,7 +288,7 @@ class LogStash::Runner < Clamp::StrictCommand
     end
 
     # lock path.data before starting the agent
-    @data_path_lock = FileLockFactory.getDefault().obtainLock(setting("path.data"), ".lock");
+    @data_path_lock = FileLockFactory.obtainLock(setting("path.data"), ".lock");
 
     @dispatcher.fire(:before_agent)
     @agent = create_agent(@settings, @source_loader)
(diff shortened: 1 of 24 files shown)


jordansissel:
The changes here look to me largely like style changes (reordering imports, refactoring Foo.getDefault().bar() to Foo.bar(), etc). 

It is not clear to me how this accomplishes the stated memory savings; can you help me understand?

original-brownbear:
@jordansissel sure I can explain :)

You already gave the example yourself:

> Foo.getDefault().bar() to Foo.bar()

This thing requires an instance of `Foo` to be set up before you can call `bar()`. But since `bar()` doesn't reference `this`, there is no point in setting up an instance of `Foo` in the first place => you save the memory you need for a `Foo` instance at least.
* plus inlining for static methods works better and you save the memory JIT requires to figure out that the method can be made static and inlined as well :) (plus all the cycles that act takes, especially when JIT is already sweating form JRuby :D)

original-brownbear:
@jordansissel ping :)

original-brownbear:
had to rebase, awaiting green

original-brownbear:
note for tomorrow, backport to `5.x`

PR #7262, opened 2017-05-30
#7184 avoid broken format being used in test

### LINKED ISSUE #7184: randomized test creates DuplicateFormatFlagsException in test
There seems to be a problem with some values of the random `page_count` https://github.com/elastic/logstash/blob/47517b89321cdcad5970a243554e6fd3c4d882c8/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java#L299 which result is a `DuplicateFormatFlagsException`.
Did not analyze but its probably when the random `page_count` is set to its limit?

```
org.logstash.ackedqueue.QueueTest > randomAcking FAILED
    java.util.DuplicateFormatFlagsException: Flags = '0'
        at java.util.Formatter$Flags.parse(Formatter.java:4443)
        at java.util.Formatter$FormatSpecifier.flags(Formatter.java:2640)
        at java.util.Formatter$FormatSpecifier.<init>(Formatter.java:2709)
        at java.util.Formatter.parse(Formatter.java:2560)
        at java.util.Formatter.format(Formatter.java:2501)
        at java.util.Formatter.format(Formatter.java:2455)
        at java.lang.String.format(String.java:2940)
        at org.logstash.ackedqueue.QueueTest.randomAcking(QueueTest.java:305)
```

original-brownbear:
Fixes #7184 

digits can't be `0`, otherwise this line is running into an invalid format string:

```java
                elements.add(new StringElement(String.format("%0" + digits + "d", i)));
```

Made it always at least `1` and made the cast to `int` more visible (`Double.intValue()` is just a cast , no need for that instance)

### DIFF
--- a/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
+++ b/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
@@ -297,7 +297,9 @@ public class QueueTest {
         // 10 tests of random queue sizes
         for (int loop = 0; loop < 10; loop++) {
             int page_count = random.nextInt(10000) + 1;
-            int digits = new Double(Math.ceil(Math.log10(page_count))).intValue();
+
+            // String format call below needs to at least print one digit
+            final int digits = Math.max((int) Math.ceil(Math.log10(page_count)), 1);
 
             // create a queue with a single element per page
             List<Queueable> elements = new ArrayList<>();


jordansissel:
Can you include the details from your PR description in your commit message? Otherwise reviewing this in the future may leave readers wondering what was broken and how `max(something, 1)` fixed it.

original-brownbear:
@jordansissel sure though actually come to think of it ... maybe we should just put an inline comment explaining this and saving people the trip to `git blame`? :)

jordansissel:
inline comment is even better :)

original-brownbear:
@jordansissel sec adding it :)

original-brownbear:
@jordansissel done :)

original-brownbear:
@jordansissel ping :)

PR #7263, opened 2017-05-30
MINOR: Remove Duplicate Hash Keys

original-brownbear:
Sorry I have a little time in 2min chunks on my hand, so removing obvious dead code :D

Both keys are duplicate so the first lines are dead code => removed :)


original-brownbear:
@jsvd thanks :)

PR #7264, opened 2017-05-30
#7178 try newest Docker

original-brownbear:
for #7178, let's try running recent Docker maybe it's just a known Docker bug in the old default version, there's quite a few in the changelog (`1.12.3` is the current Travis default ) that could be applicable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ env:
   - INTEGRATION=false FEATURE_FLAG=persistent_queues SPEC_OPTS="--order rand --format documentation"
   - INTEGRATION=true FEATURE_FLAG=persistent_queues SPEC_OPTS="--order rand --format documentation"
 before_install:
-  - sudo apt-get remove -y docker docker-engine
-  - sudo apt-get install -y docker-ce
+  - sudo apt-get remove -y docker-engine
+  - sudo apt-get install -y docker-engine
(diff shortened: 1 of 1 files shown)


jsvd:
I suggest maybe restarting the 2 integration builds in Travis a few times before merging?

original-brownbear:
@jsvd sure I can do that on the side today. Do you have any sense of the frequency of failure we're looking at here?

jsvd:
I've been hitting it quite frequently, so I'd guess that after 5 runs of the integrations test (w/ and w/o persistent queue), one of the 10 builds would trigger the problem

original-brownbear:
8 left to go then :D

jsvd:
I explained myself poorly, I think 5 runs of the ci should give us enough confidence, so after 7, let's merge!

jsvd:
This should likely go to master, 5.x and 5.4

original-brownbear:
@jsvd I think we only have the Docker tests in `5.5` and `master` not in `5.4` right? So `5.4` shouldn't be necessary?

original-brownbear:
@jsvd I think we only have the Docker tests in `5.5` and `master` not in `5.4` right? So `5.4` shouldn't be necessary?

jsvd:
awesome, hopefuly one less test failure 🎉

PR #7265, opened 2017-05-30
MINOR: Remove duplicate TimeStamp serializer class

original-brownbear:
We have the exact same class as `org.logstash.json.TimestampSerializer` and this one seems unused.

=> delete imo


original-brownbear:
nope ... gotta investigate more here it seems :D

PR #7266, opened 2017-05-30
MINOR: Remove dead methods

original-brownbear:
Removed unused methods from utility class.
Only one method remained that had only a single user -> just made it `private static` for that user.


andrewvc:
LGTM

PR #7268, opened 2017-05-30
Refactor ByteBufferPageIO._persistedByteCount() #6594

### LINKED ISSUE #6594: refactor ByteBufferPageIO._persistedByteCount()
`ByteBufferPageIO._persistedByteCount()` is public but only used for tests, there should be a cleaner way to do this.

Followup of https://github.com/elastic/logstash/pull/6519#discussion_r98057825

robbavey:
Remove static method in ByteBufferPageIO, replacing with static test helper method to move
the required functionality to a single place.

Also, cleans up HeadPageTest, using try... with resources in place of explicit, and
potentially missed, close() calls.

Resolves #6594

### DIFF
--- a/logstash-core/src/test/java/org/logstash/ackedqueue/HeadPageTest.java
+++ b/logstash-core/src/test/java/org/logstash/ackedqueue/HeadPageTest.java
@@ -1,7 +1,6 @@
 package org.logstash.ackedqueue;
 
 import org.junit.Test;
-import org.logstash.ackedqueue.io.ByteBufferPageIO;
 import org.logstash.ackedqueue.io.PageIO;
 
 import java.io.IOException;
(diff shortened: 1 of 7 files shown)


logstash-core/src/test/java/org/logstash/ackedqueue/HeadPageTest.java
```diff
@@ -1,104 +1,98 @@
...
-        assertThat(p.isFullyRead(), is(true));
-        assertThat(p.isFullyAcked(), is(false));
-        assertThat(p.hasSpace(10), is(true));
-        assertThat(p.hasSpace(100), is(false));
-
-        q.close();
+        try(Queue q = new Queue(s)) {
+            PageIO pageIO = s.getPageIOFactory().build(0, 100, "dummy");
+            pageIO.create();
+            HeadPage p = new HeadPage(0, q, pageIO);
```
original-brownbear: @robbavey I know it already was like this, but it seems we should also close `p` eventually right? It's a `Closeable` too as far as I can see. Maybe we can add another `try-with-resources` nested around the `assertThat` ?

logstash-core/src/test/java/org/logstash/ackedqueue/HeadPageTest.java
```diff
@@ -1,104 +1,98 @@
...
-        Queue q = new Queue(s);
-        q.open();
-        HeadPage p = q.headPage;
-
-        assertThat(p.hasSpace(element.serialize().length), is(true));
-        p.write(element.serialize(), 0, 1);
+        Settings s = TestSettings.volatileQueueSettings(singleElementCapacityForByteBufferPageIO(element));
+        try(Queue q = new Queue(s)) {
+            q.open();
+            HeadPage p = q.headPage;
```
original-brownbear: @robbavey same here, I think we have to safely close this one too?

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTestHelpers.java
```diff
@@ -0,0 +1,29 @@
...
+ * Class containing common methods to help DRY up acked queue tests.
+ */
+public class QueueTestHelpers {
+
+    /**
+     * Returns the minimum capacity required for {@link ByteBufferPageIO}
+     * @return int - minimum capacity required
+     */
+    public static int minimumCapacityForByteBufferPageIO(){
+        return ByteBufferPageIO.WRAPPER_SIZE;
```
original-brownbear: @robbavey this seems kinda verbose, can't we just make this a `public static final int` constant instead of a function call ? :)

original-brownbear:
@robbavey looks just fine to me except for the `minimumCapacityForByteBufferPageIO` point :)

Mind giving the `p.close()` situation a look too, looks to me like we're leaking all over here aren't we?

logstash-core/src/test/java/org/logstash/ackedqueue/HeadPageTest.java
robbavey: You are right - I missed this one. Good catch

logstash-core/src/test/java/org/logstash/ackedqueue/HeadPageTest.java
robbavey: I don't _think_ so in this case - we are not creating the headPage, but using the one inside the queue, which is closed within queue.close() which is called via the try... with resources

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTestHelpers.java
robbavey: Dude. Picky ;)

logstash-core/src/test/java/org/logstash/ackedqueue/HeadPageTest.java
original-brownbear: @robbavey right you win :)

original-brownbear:
@robbavey LGTM :)

PR #7275, opened 2017-05-31
Add date processor migration

suyograo:
@original-brownbear I decided to refactor some of the names to make it more generic (and not grok specific), but most logic is the same. There is an issue with indentation with multiple hash keys, but I'm super tired and will look tomorrow :) Let me know what you think


tools/ingest-converter/src/test/java/org/logstash/ingest/DateTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public final class DateTest extends IngestTest {
+
+    @Rule
+    public final TemporaryFolder temp = new TemporaryFolder();
```
suyograo: @original-brownbear how do I push this to the base class?

tools/ingest-converter/src/test/resources/org/logstash/ingest/logstashDateExtraFields.conf
```diff
@@ -0,0 +1,12 @@
+filter {
+   date {
+      match => [
+         "initial_date",
+         "dd/MM/yyyy hh:mm:ss",
+         "dd/MM/yyyy"
+      ]
+      target => "timestamp"
+timezone => "Europe/Amsterdam"
```
suyograo: need to fix this indent

original-brownbear:
@suyograo sorry for the delay here, not feeling so well tonight and traveling tomorrow, will get back to this in ~36h.

tools/ingest-converter/src/test/java/org/logstash/ingest/DateTest.java
original-brownbear: @suyograo you have to make the parent `IngestTest` public :)

original-brownbear: @suyograo I assume you ran into `IllegalAccessException` here and are therefore asking?

original-brownbear:
@suyograo fixed the moving of the `temp` folder to the parent class and the indents in https://github.com/original-brownbear/logstash/commit/819cc024dd6117836c050e2c9f65af48fe8d8197 , feel free to cherry-pick and merge if you agree :)

Also fixed missing `static`s :)

original-brownbear:
@suyograo next step here should probably be a "dry-up", I'm happy to do that tonight once we merged this :)

suyograo:
@original-brownbear thanks for those patches. They made sense to me, I pulled it into this one. Let me know how it looks now.

suyograo:
> next step here should probably be a "dry-up", I'm happy to do that tonight once we merged this :)

Agreed.

original-brownbear:
@suyograo LGTM, retriggered IT builds (those were failing for some reason, but I don't see how this PR could be the cause )

PR #7279, opened 2017-05-31
End-to-End test with filebeat apache2

untergeek:
This is a first run, but data flows from filebeat through Elasticsearch.

## What works

Template uploads from `$LS_HOME/modules/MODULENAME/configuration/elasticsearch/MODULENAME.json`

Specifying `--modules filebeat` from the command-line, with `-M "filebeat.var.elasticsearch.output.host=localhost:9200"`

## What doesn't work

Some of the saved searches don't get uploaded. @guyboertje is on this already.

The logstash configuration needs tweaking to allow receiving both access logs _and_ error logs.  The dashboards and visualizations all seem to expect the presence of both.

## What's fixed

Set default to `localhost` in `elasticsearch_client.rb`

Changed command-line variable parsing to allow for a variable with only `modulename.key.subkey=value`, and updated the error message accordingly.

You can now specify multiple hosts for the elasticsearch output.

## What's been added

First draft of the filebeat module, as extracted from filebeat 5.4.0


suyograo:
We need to extract just the apache2 access modules and prefix it with `fb.apache2` module. Bringing in all of this as one big module is going to be unmaintainable.

suyograo:
@untergeek The data gets populated but I get the following error:

screen shot 2017-05-31 at 22 16 34

modules/fb_apache/configuration/kibana/search/Apache2-access-logs.json
```diff
@@ -0,0 +1,19 @@
...
+  "sort": [
+    "@timestamp", 
+    "desc"
+  ], 
+  "hits": 0, 
+  "description": "", 
+  "title": "Apache2 access logs", 
+  "version": 1, 
+  "kibanaSavedObjectMeta": {
+    "searchSourceJSON": "{\"index\":\"filebeat-*\",\"query\":{\"query_string\":{\"query\":\"_exists_:apache2.access\",\"analyze_wildcard\":true}},\"filter\":[],\"highlight\":{\"pre_tags\":[\"@kibana-highlighted-field@\"],\"post_tags\":[\"@/kibana-highlighted-field@\"],\"fields\":{\"*\":{}},\"require_field_match\":false,\"fragment_size\":2147483647}}"
```
suyograo: The module name is `fb_apache` so the index gets created as `fb_apache-2017.05.01`, but this saved search is still using the `filebeat-*` index names

modules/fb_apache/configuration/elasticsearch/fb_apache.json
```diff
@@ -0,0 +1,679 @@
...
+        }
+      }
+    }
+  },
+  "order": 0,
+  "settings": {
+    "index.mapping.total_fields.limit": 10000,
+    "index.refresh_interval": "5s"
+  },
+  "template": "filebeat-*"
```
suyograo: This needs to be changed to `fb_apache`

suyograo:
@untergeek @guy I think the module creating index pattern based on the name is wrong. it should be the default `logstash-*` index pattern all the time

untergeek:
> it should be the default logstash-* index pattern all the time

@suyog, I do not think that's a good idea at all.  I think we should name them for the module, perhaps, or have it hardcoded.  We run the risk of having a lot of collisions—especially since ES is forcing one mapping type per index—if we do not separate the indices and index patterns based on some easily determinable schema (like the module name).  It should be easy enough to change the index-pattern name, and the saved searches and all that to match.

modules/fb_apache/configuration/kibana/search/Apache2-access-logs.json
untergeek: I had the `Could not locate that search` error even when the index name was still `filebeat-*`, though you are correct that the index name will cause a fail.  The problem I had was that saved searches aren't all being uploaded properly.  @guyboertje has a fix for that coming soon.

modules/fb_apache/configuration/elasticsearch/fb_apache.json
untergeek: I'm not sure that works, unless we're going to fork completely differently from Filebeat.  If this is a migration from the filebeat apache2 module, don't we need to preserve the same index naming schema? What about the saved searches and dashboards all pointing to that? 

Are we going to make it a breaking change to switch from filebeat apache?

suyograo:
@untergeek the way beats handles is that it keeps index name `filebeat-%{+YYYY.MM.dd}`, but they prefix all fields with module name:  `module_name.field_name`. So `apache2.access.host`. We need to think about it, but you do bring up a good point about collisions.

untergeek:
Namespacing the normal `logstash-YYYY.MM.dd` indices could be done, but I don't think it's ideal.  It still results in hugely complicated mappings and index-patterns for Kibana.  I don't want to create shards-plosion either, with an index per mapping, but with rollover API and single-sharded indices, it might not be as painful as it sounds.

suyograo:
@untergeek I think the shard explosion is the reason beats went with a single index, and is still a valid one (even with the rollover API). You can still change the index pattern in ES output, but I don't think having the module name prefixed is a good idea.

suyograo:
At some point, beats and LS modules will converge and I'd like to be consistent here with field names.

untergeek:
Perhaps we can preserve compatibility with filebeat modules RE: namespaced field names, but with cef and other modules, they could write to their own indices, with their own field name patterns...

guyboertje:
LGTM

suyograo:
LGTM

PR #7283, opened 2017-06-01
fixes to import index-pattern & var updates & savedsearch capability

guyboertje:
fixes to import index-pattern & var updates
add savedsearch capability


logstash-core/lib/logstash/modules/kibana_config.rb
```diff
@@ -28,26 +28,22 @@ def dashboards
...
-    # CreateKibanaIndex creates the kibana index if it doesn't exists and sets
-    # some index properties which are needed as a workaround for:
-    # https://github.com/elastic/beats-dashboards/issues/94
-    # with kibana 5.4.0 this hack failed to be applied.
-    ha = '{"settings": {"index":{"mapping":{"single_type": false}}}}'
-    hb = '{"search": {"properties": {"hits": {"type": "integer"}, "version": {"type": "integer"}}}}'
+  def index_pattern
+    pattern_name = "#{@name}-*"
+    default_index_json = '{"defaultIndex": "#{pattern_name}"}'
+    default_index_content_id = @settings.fetch("index_pattern.kibana_version", "5.4.0") # make this 5.5.0
```
suyograo: Can we change this to 5.5.0 now?

logstash-core/lib/logstash/modules/kibana_config.rb
```diff
@@ -28,26 +28,22 @@ def dashboards
...
-  def kibana_index_hacks
-    # Copied from libbeat/dashboards/importer.go
-    # CreateKibanaIndex creates the kibana index if it doesn't exists and sets
-    # some index properties which are needed as a workaround for:
-    # https://github.com/elastic/beats-dashboards/issues/94
-    # with kibana 5.4.0 this hack failed to be applied.
-    ha = '{"settings": {"index":{"mapping":{"single_type": false}}}}'
-    hb = '{"search": {"properties": {"hits": {"type": "integer"}, "version": {"type": "integer"}}}}'
+  def index_pattern
+    pattern_name = "#{@name}-*"
```
suyograo: There should be just one index pattern. We should not let this be configurable. Just be `logstash-*`

suyograo: I take that back. It should default to `logstash-*`. It shouldn't depend on module name.

suyograo:
@guyboertje couple of questions..

logstash-core/lib/logstash/modules/kibana_config.rb
suyograo: Does Kibana allow multiple index patterns?

suyograo: Actually since we are still debating this, let's ignore this one for now.

logstash-core/lib/logstash/modules/kibana_config.rb
guyboertje: well i think its ok to leave for now for testing. when we merge the adjustments for the 5.X branch we should change it then.
We can change it as a bug fix in master at any time.

logstash-core/lib/logstash/modules/kibana_config.rb
guyboertje: OK

logstash-core/lib/logstash/modules/kibana_config.rb
suyograo: Sounds good.

suyograo:
@guyboertje LGTM

PR #7284, opened 2017-06-01
Merge feature/modules into master

untergeek:


suyograo:
I don't think we need 3 reviewers for this. Maybe just @ph since he looked at the original code.

logstash-core/lib/logstash/elasticsearch_client.rb
```diff
@@ -0,0 +1,94 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/logging"
+require "elasticsearch"
+require "elasticsearch/transport/transport/http/manticore"
+        #
+        #     client = Elasticsearch::Client.new transport_class: Elasticsearch::Transport::Transport::HTTP::Manticore
```
suyograo: Can we remove this commented code line and add comments on what this class is actually doing?

logstash-core/lib/logstash/modules/kibana_config.rb
```diff
@@ -0,0 +1,103 @@
...
+require_relative "kibana_base_resource"
+
+module LogStash module Modules class KibanaConfig
+  include LogStash::Util::Loggable
+
+  ALLOWED_DIRECTORIES = ["search", "visualization"]
+
+  attr_reader :index_name
+
+  def initialize(modul, settings)
```
suyograo: Is this a typo? `modul`? Why not `module`?

logstash-core/lib/logstash/modules/scaffold.rb
```diff
@@ -0,0 +1,114 @@
...
+    return nil if @logstash_configuration.nil?
+    @logstash_configuration.config_string
+  end
+end end end # class LogStash::Modules::Scaffold
+
+# LogStash::PLUGIN_REGISTRY.add(:modules, "example", LogStash::Modules::Scaffold.new("example", File.join(File.dirname(__FILE__), "..", "configuration"))
+
+__END__
+
+settings logstash.yml
```
suyograo: Should we move this to a comment block or even to a README under `modules/README.md`

logstash-core/lib/logstash/modules/kibana_config.rb
untergeek: Because `module` has meaning in ruby

logstash-core/lib/logstash/modules/scaffold.rb
untergeek: We can remove any of this commented stuff.  It was just there as a reminder during coding.  I will remove it from the next push.

logstash-core/lib/logstash/modules/kibana_config.rb
suyograo: Can you leave a comment here? it is not obvious..

untergeek: I left comments for each of the 3 initialize methods:

```
# We name it `modul` here because `module` has meaning in Ruby.
```

I also caught a default credentials error.  It should be `elastic:changeme` rather than `elasticsearch:changeme`.

logstash-core/lib/logstash/modules/file_reader.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  def self.glob(path)
+    files = Dir.glob(path)
+    if files.nil?
+      logger.warn("No files found for glob", :pattern => path)
+    end
+    files
+  end
```
ph: `Dir.glob` returns an array, we should use if files.empty?

logstash-core/lib/logstash/modules/file_reader.rb
```diff
@@ -0,0 +1,37 @@
...
+      ""
+    end
+  end
+
+  def self.read_json(path)
+    json = read(path)
+    begin
+      LogStash::Json.load(json)
+    rescue => e
+      STDERR.puts e.message
```
ph: STDERR, can we remove this?

logstash-core/lib/logstash/modules/importer.rb
```diff
@@ -0,0 +1,39 @@
+# encoding: utf-8
+require "logstash/namespace"
+require "logstash/logging"
+
+module LogStash module Modules class Importer
+  include LogStash::Util::Loggable
+
+  attr_reader :client
```
ph: Is client used outside of this class?

logstash-core/lib/logstash/modules/logstash_config.rb
```diff
@@ -0,0 +1,46 @@
...
+    <<-CONF
+elasticsearch {
+hosts => [#{hosts}]
+index => "#{index}"
+password => "#{password}"
+user => "#{user}"
+manage_template => false
+}
+CONF
+  end
```
ph: Is there any reason why this ES construct is not in the erb file?

ph: But on the other hand, sending to ES is the most common use case, so giving people an easy macro in their template might be a good idea.

logstash-core/spec/logstash/modules/scaffold_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+    end
+    it "puts stuff in ES" do
+      test_module.with_settings(module_settings)
+      client = LogStash::ElasticsearchClient.build(module_settings)
+      import_engine = LogStash::Modules::Importer.new(client)
+      test_module.import(import_engine)
+      expect(1).to eq(1)
+    end
+  end
+end
```
ph: pretty intensive coverage. 👍

modules/README.md
```diff
@@ -0,0 +1,95 @@
...
+|   |   |   |   └── panel3.json
+│   │   │   └── v6
+|   |   |       └── panel1.json
+|   |   |       └── panel2.json
+|   |   |       └── panel3.json
+│   └── logstash
+│       └── netflow.conf.erb
+├── lib
+│   └── logstash_registry.rb
+└── logstash-module-netflow.gemspec
```
ph: Should this be moved to the doc folder?

modules/fb_apache/lib/logstash_registry.rb
```diff
@@ -0,0 +1 @@
+LogStash::PLUGIN_REGISTRY.add(:modules, "fb_apache", LogStash::Modules::Scaffold.new("fb_apache", File.join(File.dirname(__FILE__), "..", "configuration")))
```
ph: We don't need that file because we actually load the module from a logstash directory and not from a gem.

logstash-core/lib/logstash/modules/logstash_config.rb
ph: Another point I don't see any ssl related options?

logstash-core/lib/logstash/config/source/modules.rb
```diff
@@ -0,0 +1,54 @@
...
+
+          current_module = plugin_modules.find { |allmodules| allmodules.module_name == module_hash["name"] }
+          alt_name = "module-#{module_hash["name"]}"
+          pipeline_id = alt_name
+
+          current_module.with_settings(module_hash)
+          current_module.import(import_engine)
+          config_string = current_module.config_string
+
+          logger.debug("Config string for module", :config_string => config_string, :module => module_hash["name"])
```
suyograo: +1

suyograo: great catch @ph

modules/README.md
suyograo: Should be replicated there, yes, but can come as a follow up

docs/static/running-logstash-command-line.asciidoc
```diff
@@ -68,6 +68,22 @@ With this command, Logstash concatenates three config files, `/tmp/one`, `/tmp/t
   If you wish to use both defaults, please use the empty string for the `-e` flag.
   The default is nil.
 
+*`--modules`*::
+  Launch the named module.  Works in conjunction with the `-M` option to assign values to
+  default variables for the specified module.  If `--modules` is used on the command line,
+  any modules in `logstash.yml` will be ignored, as will any settings there.  This flag is
+  mutually exclusive to the `-f` and `-e` flags.  Only one of `-f`, `-e`, or `--modules` may
+  be specified.  Multiple modules can be specified by separating them with a comma, or by
+  invoking the `--modules` flag multiple times.
```
untergeek: I don't think so.  Modules should be exclusive of other pipelines.  @suyograo, that's what you wanted, at least.  We can change it later if such is wanted.

logstash-core/lib/logstash/config/source/modules.rb
untergeek: Should we just not log it? Or log a message?

logstash-core/lib/logstash/elasticsearch_client.rb
```diff
@@ -0,0 +1,94 @@
...
+        :transport_class => Elasticsearch::Transport::Transport::HTTP::Manticore,
+        :hosts => [*unpack_hosts],
+        :logger => @logger,
+      }
+    end
+
+    def unpack_hosts
+      @settings.fetch("var.output.elasticsearch.hosts", "localhost:9200").split(',').map(&:strip)
+    end
+  end
```
untergeek: Yeah.  I think we ought to support SSL certificates out of the box.  @suyograo what's your opinion here?

logstash-core/lib/logstash/modules/importer.rb
untergeek: I believe so, yes.  @guyboertje wrote this, and the client is used outside, if memory serves.

logstash-core/lib/logstash/modules/importer.rb
```diff
@@ -0,0 +1,39 @@
...
+    put_overwrite(path, resource.content)
+  end
+
+  private
+
+  def put_overwrite(path, content)
+    if content_exists?(path)
+      response = @client.delete(path)
+    end
+    # hmmm, versioning?
```
untergeek: See the `README.md` in `$LS_HOME/modules` for proposed versioning in the future.

logstash-core/lib/logstash/modules/kibana_config.rb
```diff
@@ -0,0 +1,104 @@
...
+    filenames = FileReader.read_json(dynamic("dashboard"))
+    filenames.map do |filename|
+      KibanaResource.new(@index_name, "dashboard", dynamic("dashboard", filename))
+    end
+  end
+
+  def index_pattern
+    pattern_name = "#{@name}-*"
+    default_index_json = '{"defaultIndex": "#{pattern_name}"}'
+    default_index_content_id = @settings.fetch("index_pattern.kibana_version", "5.4.0") # make this 5.5.0
```
untergeek: This is a good question.  @guyboertje can you please explain this snippet?

logstash-core/lib/logstash/modules/file_reader.rb
untergeek: What would you put here instead?

modules/fb_apache/lib/logstash_registry.rb
untergeek: True.  It's also just for testing, so should we remove it?  It doesn't hurt anything being there...

logstash-core/lib/logstash/elasticsearch_client.rb
```diff
@@ -0,0 +1,94 @@
...
+    def put(path, content)
+      normalize_response(@client.perform_request('PUT', path, {}, content))
+    end
+
+    def head(path)
+      begin
+        normalize_response(@client.perform_request('HEAD', path, {}, nil))
+      rescue Exception => e
+        if e.class.to_s =~ /NotFound/ || e.message =~ /Not\s*Found|404/i
+          Response.new(404, "", {})
```
guyboertje: in my defence - its a copy/pasta from the EC ruby client.

logstash-core/lib/logstash/elasticsearch_client.rb
guyboertje: yes - its an aide memoir from the ES ruby client docs

logstash-core/lib/logstash/elasticsearch_client.rb
```diff
@@ -0,0 +1,94 @@
...
+require "elasticsearch"
+require "elasticsearch/transport/transport/http/manticore"
+        #
+        #     client = Elasticsearch::Client.new transport_class: Elasticsearch::Transport::Transport::HTTP::Manticore
+module LogStash class ElasticsearchClient
+  include LogStash::Util::Loggable
+
+  class Response
+    # duplicated here from Elasticsearch::Transport::Transport::Response
+    # to create a normailised response across different client IMPL
```
guyboertje: typo

logstash-core/lib/logstash/modules/file_reader.rb
suyograo: This needs to be a `@logger` object

suyograo: `logger.error`

logstash-core/lib/logstash/elasticsearch_client.rb
guyboertje: I considered this - I wondered about the expertise of the user wrt SSL

logstash-core/lib/logstash/modules/file_reader.rb
guyboertje: coding remnant - delete

logstash-core/lib/logstash/modules/file_reader.rb
guyboertje: Good catch. Make it so.

logstash-core/lib/logstash/modules/importer.rb
guyboertje: Nice catch!
Removable I think. Used during coding. We should not leak a reference to this, allowing it use outside of the Importer API

logstash-core/lib/logstash/modules/importer.rb
guyboertje: By this I mean that if the content has a version, in the future, do we want to check that we are not overwriting a newer version with an older one. Quite a general question based on the content and the content type.

logstash-core/lib/logstash/modules/kibana_config.rb
guyboertje: For each LS release we will have to alter the string or the constant until we switch to the kibana API if it covers index patterns or their enhancement that picks an index-pattern as default. Note: our module solution does not require the setting of a default-index to actually work but the kibana UI bleats and moans until the user "picks" a default index pattern and in doing that the user could end up "editing" the index-pattern we so lovingly imported for them.
The PUT target is `.kibana/config/<current-kibana-version>` - IFAICT we can't query kibana for its version.
So until we can do better for the user they are going to have to tell us that they are running with a different version of kibana than the one we have released the code for i.e `5.5.0`, `5.6.0` or `6.0.0`.
Its hard. If we release the for `6.0.0` and the user later downloads `kibana 6.0.2` to fix a different bug then there will not be a default index in the kibana UI and BOOM - IDK how we successfully message this to the user.

logstash-core/lib/logstash/modules/logstash_config.rb
guyboertje: The main reason why I did this was to simplify the ERB.

Given that the final rendered `elasticsearch` section in the config would be driven by what we allow/expect in the logstash.yml i thought the logic would be easier to express in ruby than in messy, error prone ERB conditional sections.
In either case the only influence the user has over the final elasticsearch output config shape is what we say they can put in their settings.

logstash-core/lib/logstash/modules/cli_parser.rb
```diff
@@ -0,0 +1,74 @@
...
+  def get_kv(module_name, unparsed)
+    # Ensure that there is at least 1 equals sign in our variable string
+    # Using String#partition to split on the first '='
+    # This hackery is to catch the possibility of an equals (`=`) sign
+    # in a passphrase, which might result in an incomplete key.  The
+    # portion before the first `=` should always be the key, leaving
+    # the rest to be the value
+    k, op, rest = unparsed.partition('=')
+    if rest.size.zero?
+      raise LogStash::ConfigLoadingError, I18n.t("logstash.modules.configuration.modules-variables-malformed", :rawvar => (module_name + '.' + unparsed))
```
guyboertje: The difference is minimal. Speed wise, I think interpolation has the edge. If this was a oft called line I would say interpolation but its an exception raise. Rubocop where art thou?? Eating donuts probably. :-)

logstash-core/lib/logstash/config/source/modules.rb
ph: Maybe we can log something like this:

```ruby
logger.debug("Loading module configuration", :module => module_hash["name"])

logstash-core/lib/logstash/elasticsearch_client.rb
ph: Well, we can still refactor it in our code. ;)

logstash-core/lib/logstash/elasticsearch_client.rb
ph: Yes, but we are a bit different from the beats implementation, in their case, they will leverage the configured elasticsearch output in their filebeat.yml configuration. In their context, users can choose to opt in or not to secure communication. For us, if we don't export theses configurations option people won't be able to use the feature at all, without writing all their our stuff.

logstash-core/lib/logstash/modules/cli_parser.rb
ph: If we leave speed a part in this case, this is not a critical path, it's more rubyist to use string interpolation ;)

logstash-core/lib/logstash/modules/file_reader.rb
ph: Correct this need to be a added to the logger statement below.

logstash-core/lib/logstash/modules/importer.rb
ph: Dont let me started on CMS app, or shop apps, I've wrote a [magento](https://github.com/ph/magento-moneris) plugins  ;)

logstash-core/lib/logstash/modules/kibana_config.rb
ph: @guyboertje Isn't this about how we manage semver? We should just care about major?

logstash-core/lib/logstash/modules/logstash_config.rb
ph: I am OK with the macro +1

modules/fb_apache/lib/logstash_registry.rb
ph: It doesn't hurt to be there, OK to keep it.

logstash-core/lib/logstash/config/source/modules.rb
guyboertje: Remove. There is the debug config mode available.

logstash-core/lib/logstash/elasticsearch_client.rb
guyboertje: using `def is_404_error?(e)` OK?

ph: 👍

PR #7289, opened 2017-06-02
MINOR: Removed Redundant Constant

original-brownbear:
Trivial but why keep that todo around :)

Also removed visibility from only `private`ly used constant.


original-brownbear:
@jsvd thanks for taking a look :)

Issue #7294, opened 2017-06-02
Ingest migration for dotted fields to LS field references


suyograo:
One of the tests I tried after merging https://github.com/elastic/logstash/pull/7275 was a date processor with dots in `target_field`

```
{
  "description" : "...",
  "processors" : [
    {
      "date" : {
        "field" : "initial_date",
        "target_field" : "apache.timestamp",
        "formats" : ["dd/MM/yyyy hh:mm:ss", "dd/MM/yyyy"],
        "timezone" : "Europe/Amsterdam",
        "locale": "en"
      }
    }
  ]
}
```

LS

```
filter {
   date {
      match => [
         "initial_date",
         "dd/MM/yyyy hh:mm:ss",
         "dd/MM/yyyy"
      ]
      target => "apache][timestamp"
      timezone => "Europe/Amsterdam"
      locale => "en"
   }
}
```

The result is not correct: `target => "apache][timestamp"`

@

suyograo:
@original-brownbear I use `dots_to_square_brackets()` for this, but seems to be missing something. Can you take a look please?

original-brownbear:
@suyograo sure, just at it with the drying up anyways. WIll handle this :)

original-brownbear:
@suyograo fixed in https://github.com/original-brownbear/logstash/tree/dry-up-js will clean this up properly after a little sleep :)

original-brownbear:
@suyograo there we go https://github.com/elastic/logstash/pull/7296 :)

original-brownbear:
already fixed in #7296

Issue #7295, opened 2017-06-02
Complex ingest processor to LS pipeline migration


suyograo:
Since we are adding all the different pieces of the ingest migration puzzle, it would be good to have an entry point to migrate a complex processor from ingest into LS.

```json
{
  "description": "Pipeline to parse Apache logs",
  "processors": [
    {
      "grok": {
        "field": "message",
        "patterns": ["%{COMBINEDAPACHELOG}"]
      }
    },
    {
      "date": {
        "field": "timestamp",
        "target_field": "@timestamp",
        "formats": [
          "dd/MMM/YYYY:HH:mm:ss Z"
        ],
        "locale": "en"
      }
    },
    {
      "geoip": {
        "field": "client.ip",
        "target_field": "geo"
      }
    }
  ]
}
```

Converted LS pipeline

```
filter {
  grok {
    match => {
      "message" => "%{COMBINEDAPACHELOG}"
    }
  }
  date {
    match => [ "timestamp", "dd/MMM/YYYY:HH:mm:ss Z" ]
    target => "@timestamp"
    locale => "en"
  }
  geoip {
    source => "[client][ip]"
    target => "geo"
  }
}
```

Example 2:

Ingest pipeline with 2 geoip processors:

```json
{
  "description": "Pipeline to parse Apache logs",
  "processors": [
    {
      "grok": {
        "field": "message",
        "patterns": ["%{COMBINEDAPACHELOG}"]
      }
    },
    {
      "date": {
        "field": "timestamp",
        "target_field": "@timestamp",
        "formats": [
          "dd/MMM/YYYY:HH:mm:ss Z"
        ],
        "locale": "en"
      }
    },
    {
      "geoip": {
        "field": "client.ip",
        "target_field": "client.geo"
      }
    },
    {
      "geoip": {
        "field": "source.ip",
        "target_field": "source.geo"
      }
    }
  ]
}
```

Logstash:
```
filter {
  grok {
    match => {
      "message" => "%{COMBINEDAPACHELOG}"
    }
  }
  date {
    match => [ "timestamp", "dd/MMM/YYYY:HH:mm:ss Z" ]
    target => "@timestamp"
    locale => "en"
  }
  geoip {
    source => "[client][ip]"
    target => "[client][geo]"
  }
 geoip {
    source => "[source][ip]"
    target => "[source][geo]"
  }
}
```

suyograo:
I'll add more examples as I collect them :)

suyograo:
[...]

PR #7296, opened 2017-06-03
#7294 Drying up js and fixing broken case for renaming dotted to square bracket naming schema

original-brownbear:
For #7294 :

* Dried up js logic between `ingest-date.js` and `ingest-grok.js` by extracting shared logic to `IngestConverter` module
* Fixed field renaming as broken in #7294 by applying the renaming of field names to when the dotted naming pattern is all the only string in a token (so just bare `"foo.bar.blub"` is now converted, but the logic is not globally applied to any token, thought I'd be strict about this, for now, to keep the change manageable).
* Added test case from #7294 to UTs

### DIFF
--- a/tools/ingest-converter/src/main/java/org/logstash/ingest/JsUtil.java
+++ b/tools/ingest-converter/src/main/java/org/logstash/ingest/JsUtil.java
@@ -1,0 +1,39 @@
+package org.logstash.ingest;
+
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import javax.script.ScriptEngine;
+import javax.script.ScriptEngineManager;
+import javax.script.ScriptException;
+
+final class JsUtil {
+
+    private JsUtil() {
+        // Utility Class
+    }
+
+    /**
+     * Sets up a {@link ScriptEngine} for a given file loaded after `ingest-shared.js`.
+     * @param file File to set up {@link ScriptEngine} for
+     * @return {@link ScriptEngine} for file
+     */
+    public static ScriptEngine engine(final String file) {
+        final ScriptEngine engine =
+            new ScriptEngineManager().getEngineByName("nashorn");
+        try (
+            final Reader shared = reader("/ingest-shared.js");
+            final Reader reader = reader(file)
+        ) {
+            engine.eval(shared);
+            engine.eval(reader);
+        } catch (final IOException | ScriptException ex) {
+            throw new IllegalStateException(ex);
+        }
+        return engine;
+    }
+
+    private static Reader reader(final String file) {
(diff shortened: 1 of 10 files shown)


original-brownbear:
@suyograo sorry for the big PR btw, but I thought we'd keep it moving here. No actual logic changes besides making https://github.com/elastic/logstash/pull/7296/files#diff-720f2fbb867884996cb26a3621123413R11 more loosely match and adding the new test example though :)

tools/ingest-converter/src/main/java/org/logstash/ingest/JsUtil.java
```diff
@@ -0,0 +1,39 @@
...
+    private JsUtil() {
+        // Utility Class
+    }
+
+    /**
+     * Sets up a {@link ScriptEngine} for a given file loaded after `ingest-shared.js`.
+     * @param file File to set up {@link ScriptEngine} for
+     * @return {@link ScriptEngine} for file
+     */
+    public static ScriptEngine engine(final String file) {
```
suyograo: I wonder if this should take a dir and load all the scripts from there. The way it is right now is fine, but I wonder if we should look ahead to handle the conversion of the complex pipeline like so: https://github.com/elastic/logstash/issues/7295

I'm assuming there would be `ingest-date.js`, `ingest-grok.js`, `ingest-geip.js` etc.

tools/ingest-converter/src/main/resources/ingest-date.js
```diff
@@ -3,134 +3,47 @@
  */
 function ingest_to_logstash_date(json) {
 
-    function quote_string(string) {
```
suyograo: all this red is nice! :)

suyograo:
@original-brownbear I think we should tackle the full pipeline conversion next to see what shakes out for the JS structure.

suyograo:
Filed https://github.com/elastic/logstash/issues/7299 for integ test failure which is unrelated.

original-brownbear:
@suyograo got it will look into that later then :)

tools/ingest-converter/src/main/java/org/logstash/ingest/JsUtil.java
original-brownbear: @suyograo sounds good, but I think we have to be careful with load order right?

suyograo: They should all be isolated and shouldn't have any dependencies except for the `shared.js` file.

original-brownbear: @suyograo right ... makes sense :) Yea we can load all then I guess :)

PR #7297, opened 2017-06-03
MINOR: Cleanup imports on compiler.rb and pipeline.rb

original-brownbear:
Should keep the `java_import`s clean to have an easy picture of the RB/Java boundary imo :)

* Removed package name from imported Java classes
* Added missing Java imports
* Removed unused Java import


suyograo:
LGTM

PR #7300, opened 2017-06-03
Performance: Lock More Efficiently

original-brownbear:
Did some profiling on for various thread counts `x` and figured that the `Mutex` in `wrapped_synchronous_queue.rb` is a significant contributor to the overall runtime.

```sh
bin/logstash -e 'input { generator { threads => x }  } output { stdout { } }'
```

=> Saw that the `Mutex.synchronize` call takes a long time.
=> Moved to standard Java lock approach (retaining the `Mutex` which is then just a Jitted `ReentrantLock`) usage and saw improvement (though it becomes less relevant with larger `x`)

* For `x == 1` I got about a 6% speedup (measured using events/ms via the metrics endpoint).
* `x == 4` still comes with a 3% gain.

=> I think this is a valid improvement, not only for the direct gain (which is basically just due to the reduction in function calls made I think), but also for moving to a more "Java native" locking approach that is easier to reason about. I don't think the few extra lines really hurt here, given that this is on the way critical path.

**EDIT:**
~~looks like this breaks some IT ... don't merge :) Will look into this soon.~~
Look like this is not a problem with this change, but rather the ES output ...

**that said ... locking in the single threaded case comes with one order of magnitude of slowdown here for the single threaded case!**

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -127,7 +127,12 @@ module LogStash; module Util
       end
 
       def empty?
-        @mutex.synchronize { @queue.is_fully_acked? }
+        @mutex.lock
+        begin
+          @queue.is_fully_acked?
+        ensure
+          @mutex.unlock
+        end
       end
 
       def set_batch_dimensions(batch_size, wait_for)
@@ -152,8 +157,11 @@ module LogStash; module Util
       end
 
       def inflight_batches
-        @mutex.synchronize do
+        @mutex.lock
+        begin
           yield(@inflight_batches)
+        ensure
+          @mutex.unlock
         end
       end
 
@@ -173,16 +181,24 @@ module LogStash; module Util
         end
 
         batch = new_batch
-        @mutex.synchronize { batch.read_next }
+        @mutex.lock
+        begin
+          batch.read_next
+        ensure
+          @mutex.unlock
+        end
         start_metrics(batch)
         batch
       end
 
       def start_metrics(batch)
-        @mutex.synchronize do
+        @mutex.lock
+        begin
           # there seems to be concurrency issues with metrics, keep it in the mutex
           set_current_thread_inflight_batch(batch)
           start_clock
+        ensure
+          @mutex.unlock
         end
       end
 
@@ -191,12 +207,14 @@ module LogStash; module Util
       end
 
       def close_batch(batch)
-        @mutex.synchronize do
+        @mutex.lock
+        begin
           batch.close
-
           # there seems to be concurrency issues with metrics, keep it in the mutex
           @inflight_batches.delete(Thread.current)
           stop_clock(batch)
+        ensure
+          @mutex.unlock
         end
       end
 
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -57,7 +57,7 @@ module LogStash; module Util
 
       def initialize(queue, batch_size = 125, wait_for = 250)
         @queue = queue
-        @mutex = java.util.concurrent.locks.ReentrantLock.new
+        @mutex = Mutex.new
         # Note that @inflight_batches as a central mechanism for tracking inflight
         # batches will fail if we have multiple read clients in the pipeline.
         @inflight_batches = {}
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -57,7 +57,7 @@ module LogStash; module Util
 
       def initialize(queue, batch_size = 125, wait_for = 250)
         @queue = queue
-        @mutex = Mutex.new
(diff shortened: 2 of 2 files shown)


jsvd:
This is a very interesting change, I wasn't aware of these nuances. This week it will be hard for me to validate/confirm these performance characteristics, maybe someone else can review?

original-brownbear:
@jsvd I think the gain is clear from theory too (not to prevent anyone from running some testing too :), but just for posterity).
The reason for this is just that the `.synchronized` call is running a fresh block of code via:

```java
 @JRubyMethod
    public IRubyObject synchronize(ThreadContext context, Block block) {
        try {
            ProfilerRuntimeCPUFullInstr.rootMethodEntry('嵚');
            this.lock(context);

            IRubyObject var3;
            try {
                var3 = block.yield(context, (IRubyObject)null);
            } finally {
                this.unlock(context);
            }

            ProfilerRuntimeCPUFullInstr.methodExit('嵚');
            return var3;
        } finally {
            ProfilerRuntimeCPUFullInstr.methodExit('嵚');
        }
    }
```

So we will probably just see the savings from the whole overhead introduced by handling the `Block` here. In a tight loop, the gain is pretty straightforward imo, we're using the same `.lock` calls and save some of the calls involved in the `synchronize` execution => must be faster :P

ph:
> So we will probably just see the savings from the whole overhead introduced by handling the Block here. In a tight loop, the gain is pretty straightforward imo, we're using the same .lock calls and save some of the calls involved in the synchronize execution => must be faster :P

I was not aware of that at all, what is concerning is we do use the block implementation in other places in core but also in plugins.

ph:
@original-brownbear I am interested into the how did you profile that, can you share a bit more details?

ph:
Another note, We also need to implements the changes in https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/util/wrapped_acked_queue.rb#L154-L158

There is some duplicated code there.

original-brownbear:
@ph sure what I did was first run:

```sh
bin/logstash -e 'input { generator { threads => x }  } output { stdout { } }'
```

and profiled that JVM with VisualVM (just using the CPU sampler). That showed that the `synchronize` call made up the bulk of the runtime using the sample functionality. So I looked into what it does and figured that the call is just overhead relative to calling `.lock` on the Mutex.

Profiled again, now the bulk of the runtime was still the call to `.lock`, but overall numbers looked better benchmarking via the metrics as expected. Btw just reran this on Linux (gain is larger there than on my Mac, probably just because the Linux box has a cleaner environment ...).

original-brownbear:
@ph I think the more troubling thing is, that running just a single `generator` and removing all locking comes with a `10x+` speedup.
I guess it's a fairly expected result, multithreading on a shared resource is always slower than single threaded access, but I wonder how many use cases like this exist where synchronization is actually more expensive than any gain to be had from multithreading.

Amdahls Law is just brutal for many Logstash situations like this one imo, we have very fast computations that we parallelize but most of the runtime goes to sequential I/O (small theoretically possible gains, way discounted by expensive locking in practice).

ph:
Thanks @original-brownbear  for the writeup! Can you look at the other class wrapped acked queue? I will do some local test but the code look good to me.

original-brownbear:
@ph on it looking into the other code :)

original-brownbear:
@ph other queue adjusted as well :)

guyboertje:
I suggest that this is done for jruby 9K to verify that the findings still hold.

original-brownbear:
@guyboertje same code in 9k: https://github.com/jruby/jruby/blob/586b93e805034aba0d4230d54c9036f42dd200ec/core/src/main/java/org/jruby/ext/thread/Mutex.java#L145 :)

guyboertje:
@original-brownbear - the jruby code is the same for Mutex.java - but the way 9K handles method caching or method dispatch or invoke_dynamic etc may have changed significantly.

original-brownbear:
@guyboertje either way, JRuby will still be hit by the extra method calls as well as the extra `Block` object handling. Maybe it got more efficient, but some gain should be there regardless right?

colinsurprenant:
So yes, replacing
```ruby
@mutex.synchronize { something }
```
with
```ruby
@mutex.lock
begin
  something
ensure
  @mutex.unlock
end
```
is totally valid and saves a closure creation `{ something }` and a closure call when the `synchronize` method will yield the closure. 

We do know that Ruby closure creation is *not free*. The actual performance gain for this will be marginal at best as soon as you benchmark a non-empty pipeline. I don't have any particular problem with such performance modifications but bare in mind that it is already planned to rewrite the whole wrapped_*_queue code in java anyway. 

Also keep in mind that *idiomatic* Ruby relies a lot on closures. Obviously we are not tied in any way to use *idiomatic* Ruby just for the sake of it -  but closures are used in lots of places in the code base (including plugins) so I'd suggest a good assessment is made on the actual performance gain. 

FYI this is a bit the same idea why we have changed some constructs like
```ruby
loop { something }
```
to
```ruby
while true 
  something
end
```
for example https://github.com/elastic/logstash/blob/ab3546ad770b391122552809ca8a25dc16135b52/logstash-core/lib/logstash/pipeline.rb#L426-L440

original-brownbear:
@ph thanks for the review merging :)

Issue #7272, opened 2017-05-30
Add GeoIP processor to filter converter


suyograo:
Along the lines of grok converter, we should add geoip processor to geoip filter converter.

```
{
  "description" : "Add geoip info",
  "processors" : [
    {
      "geoip" : {
        "field" : "ip",
        "target_field" : "geo",
        "database_file" : "GeoLite2-Country.mmdb.gz",
        "properties": ["continent_name", "country_iso_code"]
      }
    }
  ]
}   
```

```
filter {
   geoip {
      "source" => "ip"
      "target" => "geo"
      "fields" => ["continent_name", "country_iso_code"]
   }
}
```

Example 2: dots in field names

```
{
  "description" : "Add geoip info",
  "processors" : [
    {
      "geoip" : {
        "field" : "ip",
        "target_field" : "apache.geo",
        "database_file" : "GeoLite2-Country.mmdb.gz",
        "properties": ["continent_name", "country_iso_code"]
      }
    }
  ]
}   
```

```
filter {
   geoip {
      "source" => "ip"
      "target" => "[apache][geo]"
      "fields" => ["continent_name", "country_iso_code"]
   }
}
```

original-brownbear:
@suyograo on it :)

original-brownbear:
covered by https://github.com/elastic/logstash/pull/7302

original-brownbear:
done :)

PR #7302, opened 2017-06-03
#7295 GeoIp and Complex Case Handling

original-brownbear:
Handles #7295 and #7272 (sorry the later was a prerequisite of the first one).

* Added complex case handling
   * Reuses code from the subclasses of the problem
   * Runs redundantly against all subclass tests
* Refactored tests to `@Parametrized` runs for easier reuse of all cases in the complex one (also less Java code :P)
   * Upgraded to Gradle 3.5 since we were still on `2.18`! which is broken with parametrized tests
   * Renamed some test examples to get a clear naming scheme to that effect
* Adjusted the script utility to load all scripts (sorry @suyograo, went with manual order here since the complex case has to come last + it's a little annoying to do some dir iterator here since it has to work from a jar as well as from the fs ... which it didn't with my Idea at least)


original-brownbear:
@suyograo sorry turned into a bit of a big one, but you said we want to go for the complex case next :)

original-brownbear:
huh no clue what's up with Gradle here, somehow it doesn't run `3.5` properly though it passes just fine locally ... will check this tomorrow

suyograo:
FYI, `processors` is an array of `processor` objects which I messed up in the original example. Previously I had all processors under one array element. doh

original-brownbear:
@suyograo all naming adjustments made + behaviour and examples adjusted :) ... let's see if I can fix Gradle (urgh why is Travis like that?) :)

original-brownbear:
@suyograo alright, removed the Gradle changes and squashed (arg I thought it was the Gradle bug breaking the `@Parametrized` but it was a typo in a file name that didn't show on OSX because the FS isn't case sensitive) :)

suyograo:
LGTM

Issue #7309, opened 2017-06-05
Handle ingest convert processor in migration tool


suyograo:
```json
{
  "description": "Pipeline to parse Apache logs",
  "processors": [
    {
      "convert": {
        "field" : "client.ip",
        "type": "integer"
      }
    }
  ]
}
```

LS

```
filter {
   mutate {
      convert {
         "[client][ip]" => "integer"
      }
   }
}

```

PR #7310, opened 2017-06-05
Ingest convert processor

suyograo:
Fixes https://github.com/elastic/logstash/issues/7309


tools/ingest-converter/src/main/java/org/logstash/ingest/JsUtil.java
```diff
@@ -12,7 +12,7 @@
     /**
      * Script names used by the converter in correct load order.
      */
-    private static final String[] SCRIPTS = {"shared", "date", "grok", "geoip", "pipeline"};
+    private static final String[] SCRIPTS = {"shared", "date", "grok", "geoip", "pipeline", "convert"};
```
original-brownbear: @suyograo hah I was wrong (obviously ... in hindsight) ... the load order doesn't matter :) `"pipeline"` uses `"convert"` logic just fine :)

original-brownbear:
@suyograo LGTM :)

Issue #7273, opened 2017-05-30
Add Gsub processor to filter converter


suyograo:
ingest: https://www.elastic.co/guide/en/elasticsearch/reference/master/gsub-processor.html
```
{
  "gsub": {
    "field": "field1",
    "pattern": "\.",
    "replacement": "-"
  }
}
```
LS
```
filter {
  mutate {
    gsub => [
      "field1", "\.", "_"
    ]
  }
}
```

original-brownbear:
done in https://github.com/elastic/logstash/pull/7311 :)

PR #7311, opened 2017-06-05
 #7273 ingest to logstash DSL migration: gsub

original-brownbear:
for #7273

### DIFF
--- a/tools/ingest-converter/src/main/resources/ingest-gsub.js
+++ b/tools/ingest-converter/src/main/resources/ingest-gsub.js
@@ -1,0 +1,29 @@
+var IngestGsub = {
+    has_gsub: function (processor) {
+        return !!processor["gsub"];
+    },
+    gsub_hash: function (processor) {
+        var gsub_data = processor["gsub"];
+        return IngestConverter.create_field(
+            "gsub",
+            "[\n" + [IngestConverter.dots_to_square_brackets(gsub_data["field"]),
+                gsub_data["pattern"], gsub_data["replacement"]].map(IngestConverter.quote_string)
+                .join(", ") + "\n]"
+        );
+    }
+};
+
+/**
+ * Converts Ingest JSON to LS Grok.
+ */
+function ingest_to_logstash_gsub(json) {
+
+    function map_processor(processor) {
+
+        return IngestConverter.filter_hash(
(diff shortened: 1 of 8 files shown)


original-brownbear:
@suyograo just realized: We gotta merge https://github.com/elastic/logstash/pull/7310 first and then rebase this one. So no need to review yet :)

suyograo:
@original-brownbear your turn, #7311 is in :)

original-brownbear:
@suyograo on it :)

original-brownbear:
@suyograo should be fine now :)

tools/ingest-converter/src/main/resources/ingest-gsub.js
```diff
@@ -0,0 +1,28 @@
+var IngestGsub = {
+    has_gsub: function (processor) {
+        return !!processor["gsub"];
+    },
+    gsub_hash: function (processor) {
+        var gsub_data = processor["gsub"];
+        return IngestConverter.create_field(
+            "gsub",
+            "[\n" + [gsub_data["field"], gsub_data["pattern"], gsub_data["replacement"]]
```
suyograo: Can't we just do 
```
IngestConverter.create_field(
    "gsub",
    IngestConverter.create_patterns_array([ gsub_data["field"], gsub_data["pattern"], gsub_data["replacement"] ])
```

original-brownbear: @suyograo now that you say it, it's not only "can" I think ... because the naming has to potentially be translated from dots to `[]` for `gsub_data["field"]` right?

suyograo: yep, which is handled by `create_patterns_array`, right?

original-brownbear: @suyograo jup :) will fix in a sec:)

original-brownbear:
@suyograo hmpf reuse isn't so easy here (the array thingy), because we cannot run the dots to square bracket renaming on all parts of the array (breaks the regexes).
Made it run the naming schema adjustment on the first field now, but didn't add any reuse because of the above.

Dry this up in the next step maybe? :)

Issue #7304, opened 2017-06-04
ES Output IT is Failing ~100% of Times!


original-brownbear:
I see this one in basically all PRs https://travis-ci.org/elastic/logstash/jobs/239121919#L2056.

```sh
Failures:
  1) Test Elasticsearch output can ingest 37K log lines of sample apache logs
     Failure/Error: expect(result["hits"]["total"]).to eq(37)
     
       expected: 37
            got: 0
     
       (compared using ==)
     # ./specs/es_output_how_spec.rb:25:in `(root)'
     # /home/travis/.rvm/gems/jruby-1.7.25/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `(root)'
Finished in 8 minutes 43 seconds (files took 4.61 seconds to load)
29 examples, 1 failure
Failed examples:
rspec ./specs/es_output_how_spec.rb:17 # Test Elasticsearch output can ingest 37K log lines of sample apache logs
Randomized with seed 46182

```

original-brownbear:
This is broken because ES output in general doesn't work (at least with default settings):

```sh
➜  logstash git:(master) bin/logstash -e 'input { generator { threads => 1 }  } output { elasticsearch { } }'                                               
WARNING: Default JAVA_OPTS will be overridden by the JAVA_OPTS defined in the environment. Environment JAVA_OPTS are -Xms1g -Xmx1g
ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console.
Sending Logstash's logs to /Users/brownbear/src/logstash/logs which is now configured via log4j2.properties
[2017-06-05T10:51:54,179][INFO ][logstash.setting.writabledirectory] Creating directory {:setting=>"path.dead_letter_queue", :path=>"/Users/brownbear/src/logstash/data/dead_letter_queue"}
[2017-06-05T10:51:54,203][INFO ][logstash.agent           ] No persistent UUID file found. Generating new UUID {:uuid=>"887cfee7-5b32-4ec4-87ff-ab5a89deb4c3", :path=>"/Users/brownbear/src/logstash/data/uuid"}
[2017-06-05T10:51:54,260][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because 'config.string' (-e) is being used.
[2017-06-05T10:51:54,300][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[...]

PR #7312, opened 2017-06-05
Revert "Make SafeURI 9k compatible by using java.net.URI internally"

original-brownbear:
This reverts commit 280722ea70e3ddca96204d44cd14bb37d789fc7e.

Reverted https://github.com/elastic/logstash/pull/7236 here to fix https://github.com/elastic/logstash/issues/7304.

I'll try to find an fix for #7304 that incorporates #7236 , this would be the backup plan I guess + illustration of the issue


original-brownbear:
@andrewvc I think I need you here. To me, it looks like the error that the fix you made causes (https://github.com/elastic/logstash/issues/7304#issuecomment-306137264) is a result of the type change of the `:uri` field in `SafeURI`.

I guess we could fix this in the ES Output plugin, but don't we have to be on the lookout for breaking other plugins as well? (this goes beyond my experience with the Logstash ecosystem so far :()

original-brownbear:
@jsvd feel free to chime in as well :) (maybe you know more having been the reviewer on that PR)

original-brownbear:
link https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/604 => I guess we consciously opted for the breaking change? :)

jsvd:
The breaking change makes sense, the complication comes from the patch going into 5.x too because we're migrating 5.5 to ruby 9k, not just 6.x.

We can deal with this in the elasticsearch output by introducing logstash-core version restrictions like  `>= 5.5` or bumping the logstash-core-plugin-api version which describes the contract compatibility between logstash core and the plugins.

andrewvc:
@original-brownbear yes, it was a conscious choice. Once https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/604 is merged we should be good :)

See the discussion here https://github.com/elastic/logstash/pull/7236#issuecomment-305264340

original-brownbear:
@andrewvc @jsvd ah nice thanks for the input :) Though the PR you ( @andrewvc ) link is already merged isn't it? Is it just not released?

PR #7316, opened 2017-06-05
DO NOT MERGE :) Just an illustration

original-brownbear:
5 minute illustration of https://github.com/elastic/logstash/issues/7317

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
@@ -42,7 +42,7 @@ public abstract class AbstractByteBufferPageIO implements PageIO {
     protected int head; // head is the write position and is an int per ByteBuffer class position
     protected byte version;
     private CRC32 checkSummer;
-    private final List<Integer> offsetMap; // has to be extendable
+    protected final List<Integer> offsetMap; // has to be extendable
 
     public AbstractByteBufferPageIO(int pageNum, int capacity) {
         this.minSeqNum = 0;
@@ -189,7 +189,7 @@ public abstract class AbstractByteBufferPageIO implements PageIO {
         write(bytes, seqNum, bytes.length, checksum(bytes));
     }
 
-    protected int write(byte[] bytes, long seqNum, int length, int checksum) {
+    protected int write(byte[] bytes, long seqNum, int length, int checksum) throws IOException {
         // since writes always happen at head, we can just append head to the offsetMap
         assert this.offsetMap.size() == this.elementCount :
                 String.format("offsetMap size=%d != elementCount=%d", this.offsetMap.size(), this.elementCount);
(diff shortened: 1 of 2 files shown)


original-brownbear:
`master`

```sh
Benchmark                                  Mode  Cnt   Score    Error   Units
QueueRWBenchmark.readFromPersistedQueue   thrpt   10  53.316 ± 17.471  ops/ms
QueueWriteBenchmark.pushToPersistedQueue  thrpt   10  65.128 ± 27.014  ops/ms
```

```sh
# Run complete. Total time: 00:01:34

Benchmark                                  Mode  Cnt    Score    Error   Units
QueueRWBenchmark.readFromPersistedQueue   thrpt   10  114.580 ±  6.618  ops/ms
QueueWriteBenchmark.pushToPersistedQueue  thrpt   10  188.531 ± 10.491  ops/ms
```

via https://github.com/elastic/logstash/pull/7355

colinsurprenant:
Here's my suggestion to move this forward, what do you think we create a shared branch for this and instead of patching `MmapPageIO` we create a new `PageIO` implementation like for example `DirectbufferPageIO`. 

This way it will be easier to run logstash benchmarks side-by-side to compare the behaviour of `MmapPageIO` vs `DirectbufferPageI` using the same PQ benchmark methodology I will be using early next week. 


Regarding these Logstash PQ benchmarks, my idea is to measure the following 3 possible PQ scenarios:
- faster outputs than inputs, in a *queue always-empty* situation and  measure output TPS
- slower outputs than inputs, in *growing queue* and measure persisted events per seconds (input TPS), regardless of output TPS (that will be made slower to force a growing queue).
-  faster outputs than inputs but in a draining pre-existing queue backlog situation and measure output TPS.

Then measure performance using default PQ settings vs maximum durability. This should provide us interesting data to analyse. /CC @suyograo @jordansissel

original-brownbear:
@colinsurprenant @suyograo @jordansissel this will actually not be so easy.

The problem is, that the way the `Queue` currently works is, that it actually uses the `MemoryMappedBuffer` as a hot cache and not just overflow.
This branch is merely working by accident, the approach is def. not portable. We will need to either replace alternating reads and writes to/from the same page (which looks like it'll be a huge change) or drop the use of `mmap` for reads as well (`mmap` simply won't work properly if you change the underlying file directly, while bypassing an existing mapping ... it seems to work but there are no guarantees).

Replacing `mmap` won't be easy though since the read side of the queue is predicated on reading through the whole thing to set up the offset map and such. This works for `mmap` at the cost of requiring either a page full of off-heap memory for each page read or taking a huge performance hit.
If we just use a `DirectByteBuffer`of page size to achieve the same behavior we're just getting what `mmap` gives us, but worse (one more copy, OOM instead of swapping under pressure).

=> this will require a much different approach than just swapping out `PageIO`.
=> next week :) until then inspiration can be found in:

* https://github.com/square/tape (seems like a good fit for us, implementation looks solid too)
* https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/producer/Producer.scala#L92 (`Producer` approach in Kafka is pretty close to what we want as well, has different `ack` behavior though)

colinsurprenant:
@original-brownbear makes sense. so let's followup in future design issue/discussion.

PR #7319, opened 2017-06-05
Add missing dependency on manticore

ph:
With the merge of https://github.com/elastic/logstash/pull/7284, Logstash
now depends on elasticsearch client to setup modules.

To be usable under jruby we also need to depends on manticore, this
problem is only show when try to use the logstash-core gem directly and
not in the Logstash project since it bundle the elasticsearch output
with the manticore dependency.


colinsurprenant:
LGTM

PR #7331, opened 2017-06-06
[ingest] Refactor common code to a util function

suyograo:
was getting tired of all the copy-pasta going on, fell into the trap of refactoring, albeit a light one :)


original-brownbear:
@suyograo hah you stole my plan for this morning :D Guess I'll do similar for the UTs :P I like it LGTM :)

PR #7338, opened 2017-06-06
Add Metrics for dead letter queue

### LINKED ISSUE #7287: Dead Letter Queue feature should have metrics available via the monitoring API
The Dead Letter Queue feature should expose metrics for each queue, including

- [ ] Queue path
- [ ] Number of segments stored
- [ ] Max segment size (?)
- [ ] Number of items in the queue (is total segments sufficient?)
- [ ] Disk space used by the queue
- [ ] Last event written

_work in progress_

robbavey:
Add an initial set of metrics for the dead letter queue.

Metrics are supplied under the pipeline in the following format:

"dead_letter_queue": {
        "path": ...,
        "enabled": ...,
        "current_queue_size_in_bytes": ...,
        "segment_count": ...,
        "max_segment_size_in_bytes": ...
      }

Note that if the dead letter queue is not enabled, then only the
enabled property will be displayed, with a value of false.

Metrics are populated via a PeriodicPoller

Resolves #7287

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -58,11 +58,7 @@ module LogStash; class BasePipeline
     @outputs = nil
     @agent = agent
 
-    if settings.get_value("dead_letter_queue.enable")
-      @dlq_writer = DeadLetterQueueFactory.getWriter(pipeline_id, settings.get_value("path.dead_letter_queue"))
-    else
-      @dlq_writer = LogStash::Util::DummyDeadLetterQueueWriter.new
-    end
+    @dlq_writer = dlq_writer
 
     grammar = LogStashConfigParser.new
     parsed_config = grammar.parse(config_str)
@@ -82,6 +78,14 @@ module LogStash; class BasePipeline
       eval(config_code)
     rescue => e
       raise e
+    end
+  end
+
+  def dlq_writer
+    if settings.get_value("dead_letter_queue.enable")
+      @dlq_writer = DeadLetterQueueFactory.getWriter(pipeline_id, settings.get_value("path.dead_letter_queue"))
+    else
+      @dlq_writer = LogStash::Util::DummyDeadLetterQueueWriter.new
     end
   end
 
@@ -245,6 +249,7 @@ module LogStash; class Pipeline < BasePipeline
     # Since we start lets assume that the metric namespace is cleared
     # this is useful in the context of pipeline reloading
     collect_stats
+    collect_dlq_stats
 
     @logger.debug("Starting pipeline", default_logging_keys)
 
@@ -383,6 +388,9 @@ module LogStash; class Pipeline < BasePipeline
       config_metric.gauge(:batch_delay, batch_delay)
       config_metric.gauge(:config_reload_automatic, @settings.get("config.reload.automatic"))
       config_metric.gauge(:config_reload_interval, @settings.get("config.reload.interval"))
+      config_metric.gauge(:dead_letter_queue_enabled, dlq_enabled?)
+      config_metric.gauge(:dead_letter_queue_path, @dlq_writer.get_path.to_absolute_path.to_s) if dlq_enabled?
+
 
       @logger.info("Starting pipeline", default_logging_keys(
         "pipeline.workers" => pipeline_workers,
@@ -414,6 +422,10 @@ module LogStash; class Pipeline < BasePipeline
       # to potentially unblock the shutdown method which may be waiting on @ready to proceed
       @ready.make_true
     end
+  end
+
+  def dlq_enabled?
+    @settings.get("dead_letter_queue.enable")
   end
 
   # Main body of what a worker thread does
@@ -696,10 +708,16 @@ module LogStash; class Pipeline < BasePipeline
       .each {|t| t.delete("status") }
   end
 
+  def collect_dlq_stats
+    if dlq_enabled?
+      dlq_metric = @metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :dlq])
+      dlq_metric.gauge(:queue_size_in_bytes, @dlq_writer.get_current_queue_size)
+    end
+  end
+
   def collect_stats
     pipeline_metric = @metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :queue])
     pipeline_metric.gauge(:type, settings.get("queue.type"))
-
     if @queue.is_a?(LogStash::Util::WrappedAckedQueue) && @queue.queue.is_a?(LogStash::AckedQueue)
       queue = @queue.queue
       dir_path = queue.dir_path
--- a/logstash-core/spec/logstash/instrument/periodic_poller/dlq_spec.rb
+++ b/logstash-core/spec/logstash/instrument/periodic_poller/dlq_spec.rb
@@ -1,0 +1,17 @@
+# encoding: utf-8
+require "spec_helper"
+require "logstash/instrument/periodic_poller/dlq"
+require "logstash/instrument/collector"
+
+describe LogStash::Instrument::PeriodicPoller::DeadLetterQueue do
+  subject { LogStash::Instrument::PeriodicPoller::DeadLetterQueue }
+
+  let(:metric) { LogStash::Instrument::Metric.new(LogStash::Instrument::Collector.new) }
+  let(:agent) { double("agent")}
+  let(:options) { {} }
+  subject(:dlq) { described_class.new(metric, agent, options) }
+
+  it "should initialize cleanly" do
+    expect { dlq }.not_to raise_error
+  end
+end
--- a/logstash-core/spec/logstash/pipeline_spec.rb
+++ b/logstash-core/spec/logstash/pipeline_spec.rb
@@ -104,12 +104,19 @@ describe LogStash::Pipeline do
   let(:worker_thread_count)     { 5 }
   let(:safe_thread_count)       { 1 }
   let(:override_thread_count)   { 42 }
+  let(:dead_letter_queue_enabled) { false }
+  let(:dead_letter_queue_path) { }
   let(:pipeline_settings_obj) { LogStash::SETTINGS }
   let(:pipeline_settings) { {} }
 
   before :each do
     pipeline_workers_setting = LogStash::SETTINGS.get_setting("pipeline.workers")
     allow(pipeline_workers_setting).to receive(:default).and_return(worker_thread_count)
+    dlq_enabled_setting = LogStash::SETTINGS.get_setting("dead_letter_queue.enable")
+    allow(dlq_enabled_setting).to receive(:value).and_return(dead_letter_queue_enabled)
+    dlq_path_setting = LogStash::SETTINGS.get_setting("path.dead_letter_queue")
+    allow(dlq_path_setting).to receive(:value).and_return(dead_letter_queue_path)
+
     pipeline_settings.each {|k, v| pipeline_settings_obj.set(k, v) }
   end
 
@@ -838,6 +845,33 @@ describe LogStash::Pipeline do
         [multiline_id, multiline_id_other].map(&:to_sym).each do |id|
           plugin_name = id.to_sym
           expect(collected_metric[:stats][:pipelines][:main][:plugins][:filters][plugin_name][:name].value).to eq(LogStash::Filters::Multiline.config_name)
+        end
+      end
+
+      context 'when dlq is disabled' do
+        let (:collect_stats) { subject.collect_dlq_stats}
+        let (:collected_stats) { collected_metric[:stats][:pipelines][:main][:dlq]}
+        let (:available_stats) {[:path, :queue_size_in_bytes]}
+
+        it 'should show not show any dlq stats' do
+          collect_stats
+          expect(collected_stats).to be_nil
+        end
+
+      end
+
+      context 'when dlq is enabled' do
+        let (:dead_letter_queue_enabled) { true }
+        let (:dead_letter_queue_path) { Stud::Temporary.directory }
+        let (:pipeline_dlq_path) { "#{dead_letter_queue_path}/#{pipeline_id}"}
+
+        let (:collect_stats) { subject.collect_dlq_stats }
+        let (:collected_stats) { collected_metric[:stats][:pipelines][:main][:dlq]}
+
+        it 'should show dlq stats' do
+          collect_stats
+          # A newly created dead letter queue with no entries will have a size of 1 (the version 'header')
+          expect(collected_stats[:queue_size_in_bytes].value).to eq(1)
         end
       end
     end
--- a/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
@@ -32,6 +32,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
+import java.util.concurrent.atomic.LongAdder;
 import java.util.stream.Stream;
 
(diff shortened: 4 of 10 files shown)


robbavey:
Failing test looks to be another instance of #7178

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -696,10 +701,21 @@ def stalling_threads_info
       .each {|t| t.delete("status") }
   end
 
+  def collect_dlq_stats
+    dlq_metric = @metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :dlq])
+    enabled = settings.get_value("dead_letter_queue.enable")
+    dlq_metric.gauge(:enabled, enabled)
+    if enabled
+      dlq_metric.gauge(:path, @dlq_writer.get_path.to_absolute_path)
```
jordansissel: Should this appear in metrics? I think we have a separate API to show the runtime settings of Logstash (https://www.elastic.co/guide/en/logstash/current/node-info-api.html) that could be used for showing whether or not the queue is enabled, what path it has, etc.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -696,10 +701,21 @@ def stalling_threads_info
...
+      dlq_metric.gauge(:segment_count, @dlq_writer.get_segment_count)
```
jordansissel: Would a user want to take action or make some conclusion based on the `segment_count` value?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -696,10 +701,21 @@ def stalling_threads_info
...
+      dlq_metric.gauge(:current_queue_size_in_bytes, @dlq_writer.get_current_queue_size)
+      dlq_metric.gauge(:max_segment_size_in_bytes, @dlq_writer.get_max_segment_size)
```
jordansissel: max segment size seems like a setting, not a metric. Thoughts?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -696,10 +701,21 @@ def stalling_threads_info
...
   end
 
+  def collect_dlq_stats
+    dlq_metric = @metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :dlq])
+    enabled = settings.get_value("dead_letter_queue.enable")
+    dlq_metric.gauge(:enabled, enabled)
+    if enabled
+      dlq_metric.gauge(:path, @dlq_writer.get_path.to_absolute_path)
+      dlq_metric.gauge(:segment_count, @dlq_writer.get_segment_count)
+      dlq_metric.gauge(:current_queue_size_in_bytes, @dlq_writer.get_current_queue_size)
```
jordansissel: This is already namespaced under dlq, so maybe we can call it `size_bytes`  or `size_in_bytes` only?

logstash-core/lib/logstash/pipeline.rb
robbavey: I used the metrics from the persistent queue (https://www.elastic.co/guide/en/logstash/current/node-stats-api.html#pipeline-stats) as a guide/model for the dlq metrics, which includes path.
Also, the path includes the pipeline_id, so, if I understand correctly, can differ with multi-pipelines.

logstash-core/lib/logstash/pipeline.rb
robbavey: Possibly not... I found it useful, as I wasn't aware that a new segment was created every time logstash starts up before I saw the stat ;)

I'm happy to remove it, if it is superfluous, though

logstash-core/lib/logstash/pipeline.rb
robbavey: I added the 'current_queue' prefix to disambiguate from a max_queue_size. Again, happy to remove if superfluous.

logstash-core/lib/logstash/pipeline.rb
robbavey: Like path, I modeled this after pq, which includes a similar setting. 

Happy to move it, if it makes sense there. I expect this isn't a setting that would want to be changed per dlq in the multi-pipeline world

robbavey: It appears to be hard coded to a constant right now, anyway, so I'll kill this

robbavey:
@jordansissel: Given your comments above, it seems like there are a couple of potential approaches if we want to move the dlq settings out of the stats endpoint:

Move dlq settings to the _node/pipelines endpoint, and have dlq settings inside the pipeline definition - this potentially enables per-pipeline enablement of dlq, which doesn't appear to be supported today eg:

```
{
  "host": "Robs-MacBook-Pro.local",
  "version": "6.0.0-alpha3",
  "http_address": "127.0.0.1:9600",
  "id": "a492f081-870d-4423-afd0-9b1f92efb2de",
  "name": "Robs-MacBook-Pro.local",
  "pipelines": {
    "main": {
      "workers": 8,
      "batch_size": 125,
      "batch_delay": 5,
      "config_reload_automatic": false,
      "config_reload_interval": 3,
      "dead_letter_queue_path": "/Users/robbavey/code/logstash/data/dead_letter_queue/main",
      "dead_letter_queue_enabled": true
    }
  }
}
```


Have a separate dead_letter_queue endpoint, for global dlq settings:

```
{
  "host": "Robs-MacBook-Pro.local",
  "version": "6.0.0-alpha3",
  "http_address": "127.0.0.1:9600",
  "id": "a492f081-870d-4423-afd0-9b1f92efb2de",
  "name": "Robs-MacBook-Pro.local",
  "dead_letter_queue": {
      "enabled": true,
      "path": "/Users/robbavey/code/logstash/data/dead_letter_queue",
    }
  }
}
```

jordansissel:
I'm OK if we are copying existing practice where sometimes a setting is exposed via stats api. I'd prefer if settings were not available in stats, but we can solve that later as this concern impacts more than this PR.

So for now, I will retract my "this is a setting and may not belong here" comments. :)

robbavey:
The latest commit removes the superfluous metrics, and just has:

```
 "dead_letter_queue": {
        "path": "/Users/robbavey/code/logstash/data/dead_letter_queue/main",
        "enabled": true,
        "queue_size_in_bytes": 1650
      }
```

logstash-core/spec/logstash/instrument/periodic_poller/dlq_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  subject(:dlq) { described_class.new(metric, agent, options) }
+
+  it "should initialize cleanly" do
+    expect { dlq }.not_to raise_error
+  end
+
+
+
+  describe 'collect' do
+    subject(:collect) { dlq.collect }
```
jordansissel: If you aren't going to invoke `subject` in the tests, I don't think you need this.

Later down in the tests (line 33, etc) you can do:

```
expect { dlq.collect }.not_to  raise_error
```

etc.

logstash-core/spec/logstash/instrument/periodic_poller/dlq_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    before :each do
+      allow(agent).to receive(:with_running_pipelines) { |&block| block.call(pipelines) }
+      allow(pipeline).to receive(:collect_dlq_stats)
+    end
+
+    it 'should run cleanly' do
+      expect { collect }.not_to raise_error
+    end
+
+    it 'should request dlq stats from the pipeline' do
```
jordansissel: This test asserts some deep assumptions about the names of methods and the way things are invoked.

Can this test be rewritten to test output (maybe the return value of `dlq.collect` instead of side effects such as what methods are called, etc?

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -762,9 +762,12 @@ class TestPipeline < LogStash::Pipeline
     end
     let(:dummyoutput) { ::LogStash::Outputs::DummyOutput.new({ "id" => dummy_output_id }) }
     let(:metric_store) { subject.metric.collector.snapshot_metric.metric_store }
+    let(:before_subject_hook) {}
     let(:pipeline_thread) do
       # subject has to be called for the first time outside the thread because it will create a race condition
       # with the subject.ready? call since subject is lazily initialized
+      # The hook is used to allow support for mocking,etc in tests
+      before_subject_hook
```
jordansissel: This is an interesting workaround. It doesn't feel like the right way to do this, though I am not able to precisely articulate exactly why I feel this way.

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -840,6 +843,64 @@ class TestPipeline < LogStash::Pipeline
...
+        let (:before_subject_hook) {
+          allow_any_instance_of(LogStash::BasePipeline).to receive(:dlq_writer).and_return(dlq_writer)
+          allow(dlq_writer).to receive(:get_path).and_return(path)
+          allow(path).to receive(:to_absolute_path).and_return(queue_path)
+          allow(dlq_writer).to receive(:get_current_queue_size).and_return(queue_size)
+          allow(dlq_writer).to receive(:close)
+        }
+
+        let (:queue_size) { 1024 }
+        let (:queue_path) { '/tmp' }
```
jordansissel: This should be randomized (maybe using `Stud::Temporary.directory`) to avoid conflicting with other tests. Thoughts?

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -840,6 +843,64 @@ class TestPipeline < LogStash::Pipeline
...
+        # Hook to setup mocks to enable a stubbed dlq_writer to be set.
+        let (:before_subject_hook) {
+          allow_any_instance_of(LogStash::BasePipeline).to receive(:dlq_writer).and_return(dlq_writer)
+          allow(dlq_writer).to receive(:get_path).and_return(path)
+          allow(path).to receive(:to_absolute_path).and_return(queue_path)
+          allow(dlq_writer).to receive(:get_current_queue_size).and_return(queue_size)
+          allow(dlq_writer).to receive(:close)
+        }
+
+        let (:queue_size) { 1024 }
```
jordansissel: This can be randomized.

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -153,4 +152,30 @@ public synchronized void close() throws IOException {
     public boolean isOpen() {
         return open;
     }
+
+    public Path getPath(){
```
jordansissel: spacing between `()` and `{`

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -153,4 +152,30 @@ public synchronized void close() throws IOException {
...
+        return queuePath;
+    }
+
+    public long getSegmentCount(){
+       try {
+           return getSegmentPaths(queuePath).count();
+       }catch (IOException e){
```
jordansissel: space between `}` and `catch` and also between `)` and `{`

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -153,4 +152,30 @@ public synchronized void close() throws IOException {
...
+           logger.warn("Unable to get current segment count", e);
+           return -1L;
+       }
+    }
+
+    public long getCurrentQueueSize(){
```
jordansissel: spacing between () and {

logstash-core/spec/logstash/pipeline_spec.rb
robbavey: Yeah, I don't like it much either, but other things I tried ran into a bit of a dead end... The combination of a Java static factory class for creating the DeadLetterQueueWriter and the fact that the subject is instantiated in a 'before :each' block made this a little tricky.
There are integration tests for these metrics, which make me a little happier, but if you have any other suggestions for better ways to test this, I'm open

logstash-core/spec/logstash/pipeline_spec.rb
robbavey: +1

logstash-core/spec/logstash/pipeline_spec.rb
robbavey: I don't _think_ it will interfere with other tests, as no entries are written to any directory - this test uses a test double for dlq_writer just to return stats.

I think I have an idea how to do this with a few less egregious mocks...

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
robbavey: Sorry...

logstash-core/spec/logstash/instrument/periodic_poller/dlq_spec.rb
robbavey: +1

logstash-core/spec/logstash/pipeline_spec.rb
robbavey: (Thankfully) I've got rid of this in the next iteration.

logstash-core/spec/logstash/instrument/periodic_poller/dlq_spec.rb
robbavey: The intention behind this method was to ensure that the dlq_stats were collected, and have the actual values tested elsewhere. I'm trying to figure out if it makes sense to setup all the scaffolding to set up a running pipeline to retrieve actual values here (the pq implemenation of this doesn't), when this is tested elsewhere

robbavey:
@jordansissel - fixed up the spacing issues, got rid of the 'before hook', and used more 'real data'. Also removed the stub heavy test in the periodic poller, but didn't replace it - I'm not sure the test warrants hooking up with a real pipeline to get real data - that is taken care of in a separate unit and integration tests.

jordansissel:
@robbavey thank you! I will do another review.

jordansissel:
Unable to test this normally due to a bug in the latest Logstash elasticsearch output. I will try another way.

jordansissel:
Worked around the bug with our old friend, the ruby filter: `filter { ruby { code => "execution_context.dlq_writer.write(event, :testing.to_s)" } }'`

jordansissel:
> this potentially enables per-pipeline enablement of dlq

I don't know if we allow per-pipeline DLQ enabling (honestly, in the future, maybe we don't even have a way to disable DLQ, so I am hoping this per-pipeline enable is not necessary)

That said, DLQ is *per pipeline* already, the path to a particular pipeline's DLQ is `path.dead_letter_queue` plus the pipeline name, such as `/logstash/data/dead_letter_queue/main/` where `main` is my pipeline name. Further, the `dead_letter_queue` input has a `pipeline_id` setting.

Given this, I am in favor of removing the dlq settings from the /_node/stats endpoint and moving them into /_node/pipelines instead.

For /_node/stats, this means the following change:

```diff
      "dead_letter_queue" : {
-       "path" : "/home/jls/projects/logstash/data/dead_letter_queue/main",
-       "enabled" : true,
        "queue_size_in_bytes" : 2
      }
```

robbavey:
@jordansissel Thank you for pointing out the ruby filter! I wrote my own plugin to test this (which was a useful exercise on its own ;)

Metrics updated to follow your suggestion, and I also added dead_letter_queue_enabled and dead_letter_queue_path to the pipelines endpoint.

talevy:
@jordansissel re: https://github.com/elastic/logstash/pull/7338#issuecomment-306994562

https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/609

robbavey:
@jordansissel Any further objections before merging?

jordansissel:
👍 👍 👍

PR #7339, opened 2017-06-06
Fix/modules common1

guyboertje:
need this in master to pick into 5.x


guyboertje:
Looks like there is a syntax error typo - test fail

guyboertje:
Closing - I want to preserve the two commit PR

PR #7342, opened 2017-06-06
Add DLQ input to default plugins

suyograo:


suyograo:
Unrelated integ test failure which is a known issue. Merging.

PR #7349, opened 2017-06-06
Upgrade to JRuby 9k

ph:
This is a joint effort between @guyboertje  and @ph; this is the minimal changes required to make Logstash work under JRuby 9k.


rakelib/artifacts.rake
```diff
@@ -1,5 +1,20 @@
 require "logstash/version"
 
+
```
suyograo: Mistake?

ph: /facepalm.

ph:
I've look at the test the offline pack is currently failling and I can reproduce the error everytime locally. This is due to the fact that we rely on Rubygems' library to handle the indexes and 9k ships with a newer version of it that break the current code.

rakelib/childprocess_patch.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      def ignore_close_io
+        old_pump
+      rescue IOError
+      end
+
+      alias_method :pump, :ignore_close_io
+    end
+  end
+end
```
ph: This might need to go in FPM.

ph:
@guyboertje any ideas about theses  message: `jruby: warning: unknown property jruby.cext.enabled` in the travis job? It like something is set by default in the Travis environment.

guyboertje:
LGTM

ph:
squashed, waiting on ci, will merge after.

ph:
[master 76286b4f0]

PR #7354, opened 2017-06-07
Update monitoring api docs for multiple pipelines

dedemorton:


jsvd:
I noticed that all my json samples have an unnecessary whitespace character between the key string and the colon, like:

```json
{
  "pipelines" : {
    "test" : {
      "workers" : 1,
      "batch_size" : 1,
      "batch_delay" : 5,
      "config_reload_automatic" : false,
      "config_reload_interval" : 3
    }
  }
}
```

instead of:

```json
{
  "pipelines": {
    "test": {
      "workers": 1,
      "batch_size": 1,
      "batch_delay": 5,
      "config_reload_automatic": false,
      "config_reload_interval": 3
    }
  }
}
```

This introduces unnecessary git diff changes and makes the JSON less readable IMO. Maybe just a find and replace from `" :` to `":`?

docs/static/monitoring-apis.asciidoc
```diff
@@ -266,55 +303,55 @@ Example response:
...
-      "non_heap_used_in_bytes": 189382304,
-      "non_heap_committed_in_bytes": 200728576,
-      "pools": {
-        "survivor": {
-          "peak_used_in_bytes": 8912896,
-          "used_in_bytes": 9538656,
-          "peak_max_in_bytes": 35782656,
-          "max_in_bytes": 71565312,
-          "committed_in_bytes": 17825792
+    "mem" : {
```
jsvd: to have less git diff changes, I think these changes can be discarded since the structure remains same

docs/static/monitoring-apis.asciidoc
```diff
@@ -333,22 +370,46 @@ Example response:
 [source,js]
 --------------------------------------------------
 {
-  "process": {
```
jsvd: here too we can discard the changes as the only touch whitespace and the values

docs/static/monitoring-apis.asciidoc
```diff
@@ -333,22 +370,46 @@ Example response:
...
       }
     }
   }
-}
+--------------------------------------------------
+
+[[event-stats]]
+==== Event Stats
+
+The following request returns a JSON document containing cumulative event counts
```
jsvd: maybe `event related statistics` instead of `event counts`, as duration_in_millis and queue_push_duration_in_millis is also displayed?

docs/static/monitoring-apis.asciidoc
```diff
@@ -517,64 +690,29 @@ Logstash.
 
 Example response:
 
-[source,js]
+["source","sh",subs="attributes"]
 --------------------------------------------------
 {
-    "time": "2017-01-12T12:09:45-08:00",
```
jsvd: we can discard the changes here as the structure didn't change, but up to you

docs/static/monitoring-apis.asciidoc
```diff
@@ -598,46 +736,26 @@ curl -XGET 'localhost:9600/_node/hot_threads?human=true'
...
 Example of a human-readable response:
 
-[source,js]
+["source","js",subs="attributes"]
 --------------------------------------------------
  ::: {}
- Hot threads at 2017-01-12T12:10:15-08:00, busiestThreads=3: 
+ Hot threads at 2017-06-06T18:31:17-07:00, busiestThreads=3: 
  ================================================================================
- 1.02 % of cpu usage, state: timed_waiting, thread name: 'LogStash::Runner'
```
jsvd: as before, I suggest not introducing changes, wdyt?

docs/static/monitoring-apis.asciidoc
dedemorton: @jsvd I'm going to leave the changes as they are here. I agree that whitespace is minor, but I also think it's good for the docs to match what the software actually returns, and in this case, the extra space was introduced in the formatting. I try to update the examples whenever I test the APIs because it helps make the doc look less dated over time...sorry if it cause you a little extra time to confirm the changes.

docs/static/monitoring-apis.asciidoc
dedemorton: ditto

docs/static/monitoring-apis.asciidoc
dedemorton: @jsvd I try to update the examples from time-to-time so they don't look stale (because the date is old or the format of events changes etc). I always replace the entire example instead of just tweaking parts because it's easy to do and less error prone.  Again, sorry if it took you a bit more time to scan through the changes....

docs/static/monitoring-apis.asciidoc
dedemorton: again...it's nice to have the date updated. :-) It's pretty easy on my end to drop in the new examples.

dedemorton:
@jsvd I've pushed changes from the review. As I mentioned, I didn't revert the examples because that actually would have been more work...plus I think it's good for the examples to as closely as possible match what LS spits out.

PR #7360, opened 2017-06-07
MINOR: Remove dead and duplicate code from Event and EventTest

original-brownbear:
Just some minor housekeeping :) Found these when looking over how to best approach #7039 

* Remove unused methods (all of them had no reference in the code when grepping except for in some old comment from 2015 -> removed that comment too)
* Removed duplicate methods, `toBinary`  `serialize` and `serializeWithoutSeqNum` were exactly the same!
   * removed test `queueableInterfaceWithoutSeqNumRoundTrip` since it was logically 100% equivalent to `queueableInterfaceRoundTrip` because of the above

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Event.java
+++ b/logstash-core/src/main/java/org/logstash/Event.java
@@ -96,24 +96,8 @@ public class Event implements Cloneable, Serializable, Queueable {
         return this.metadata;
     }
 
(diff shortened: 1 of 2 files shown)


jordansissel:
Tests pass, so I think this is good to merge. LGTM

original-brownbear:
gnarf need to backport to `5.x`

PR #7368, opened 2017-06-07

andrewvc:


ph:
@andrewvc related PR on beats.https://github.com/logstash-plugins/logstash-input-beats/pull/216

ph:
Note: beats need to be build/publish, if we don't the build will fail.

ph:
I've hitted approved without writing LGTM.

PR #7370, opened 2017-06-07
MINOR: Fix static logic not declared static

original-brownbear:
There were some syntactic mistakes here, accessing a `static` field with `this.` => removed those and it made it clear that we're still missing a few `static` here :)

=> Added missing type on cache map
=> Removed redundant non `static` access to eventually `static` field
=> Better correctness and better chance of inlining :)


original-brownbear:
@ph thanks :)

PR #7372, opened 2017-06-07

andrewvc:
The old version (1.0.0) had a broken version of Concurrent::Timer that did not work on all machines for reasons that are still unclear, but may be related to the Java version

Fixes #7368

v5.5 version of #7368


andrewvc:
I'm not sure why there is such a large diff in the .lock . I must have built the .lock file with the wrong Gemfile. Where can I get the right one?

PR #7373, opened 2017-06-07

andrewvc:

Fixes #7368 

V 5.4 port of #7368


Gemfile.jruby-1.9.lock
```diff
@@ -241,8 +241,8 @@ GEM
       logstash-core-plugin-api (>= 1.60, <= 2.99)
       nokogiri
       xml-simple
-    logstash-input-beats (3.1.15-java)
-      concurrent-ruby (>= 0.9.2, <= 1.0.0)
+    logstash-input-beats (4.0.2-java)
```
jordansissel: This PR targets 5.4 so is intended for a patch release, but this change does a major bump on the beats input. We cannot do this in a patch release.

andrewvc: Good catch, I should have bumped to `3.1.18` which @ph released today to handle this gem incompatibility.

PR #7374, opened 2017-06-07
Fix example and add couple of usage notes

dedemorton:
Fixed FB config examples for changes in 6.0.0-alpha2. Plus added a few other blurbs to clarify stuff that might be confusing due to other recent changes in LS and Beats.


suyograo:
LGTM

PR #7345, opened 2017-06-06
Introducing Logstash Modules

guyboertje:
Modules contain packaged Logstash configuration, Kibana dashboards and other meta files to ease the set up of the Elastic stack for certain use cases or data sources.

PR #7376, opened 2017-06-07
Bump concurrent gem dep to more recent version

andrewvc:

Fixes #7368

V 5.5 port of #7368


andrewvc:
I have no idea why bundler is putting in the elasticsearch and manticore deps but it shouldn't matter

Gemfile.jruby-1.9.lock.release
```diff
@@ -4,13 +4,15 @@ PATH
     logstash-core (5.5.0-java)
       chronic_duration (= 0.10.6)
       clamp (~> 0.6.5)
-      concurrent-ruby (= 1.0.0)
+      concurrent-ruby (~> 1.0, >= 1.0.5)
+      elasticsearch (~> 5.0, >= 5.0.4)
```
jordansissel: I was curious about elasticsearch dependency. It came in as part of 351069f9 (#7345).

andrewvc: Good catch!

PR #7380, opened 2017-06-08
#7382 Fix broken Buffer Double Locking

### LINKED ISSUE #7382: Concurrency Issue in Queue (multiple consumers, one producer) <= Performance + Bug
Ran into this when trying to add buffering to the `FileChannel` IO approach.

If I run 2 consumers against one producer in the way shown in https://github.com/elastic/logstash/pull/7380 one of the consumers eventually locks up.

The reason for this is the following:

Assuming we're starting from an empty queue both consumers will block the last line of the snippet below:

```java
    public Batch readBatch(int limit) throws IOException {
        Page p;

        lock.lock();
        try {
            while ((p = firstUnreadPage()) == null && !isClosed()) {
                try {
                    notEmpty.await();
```

Now this blocking is eventually unlocked by 

```java
 public long write(Queueable element) throws IOException
```

doing this:

```
            // if the queue was empty before write, signal non emptiness
            // a simple signal and not signalAll is necessary here since writing a single element
            // can only really enable a single thread to read a batch
            if (wasEmpty) { notEmpty.signal(); }
```

This is the only place in the `Queue` code that signals this lock. This means:

* If a consumer calls `readBatch()` and neither drains the queue nor fires a new `readBatch()`, the Queue becomes locked since all other consumers continue to wait for the signal (which is never fired by the `write`  since the `wasEmpty` is always false). => bug(?) reproduced in #7380 
* The performance of multiple consumers in general is suboptimal, __as soon as you start to fall behind all consumption happens from a single consumer only__.

original-brownbear:
Fixes #7382, see explaination below

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -324,7 +324,6 @@ public class Queue implements Closeable {
 
         lock.lock();
         try {
-            boolean wasEmpty = (firstUnreadPage() == null);
 
             // create a new head page if the current does not have sufficient space left for data to be written
             if (! this.headPage.hasSpace(data.length)) {
@@ -356,11 +355,8 @@ public class Queue implements Closeable {
             long seqNum = nextSeqNum();
             this.headPage.write(data, seqNum, this.checkpointMaxWrites);
             this.unreadCount++;
-
-            // if the queue was empty before write, signal non emptiness
-            // a simple signal and not signalAll is necessary here since writing a single element
-            // can only really enable a single thread to read a batch
-            if (wasEmpty) { notEmpty.signal(); }
+            
+            notEmpty.signal();
 
             // now check if we reached a queue full state and block here until it is not full
             // for the next write or the queue was closed.
--- a/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
+++ b/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
@@ -13,6 +13,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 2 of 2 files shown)


original-brownbear:
screen shot 2017-06-08 at 19 58 06

fixed => the blocking queue was implemented wrong.

You need to signal the `notEmpty` condition on every append like `ArrayBlockingQueue` does:

```java

    /**
     * Inserts element at current put position, advances, and signals.
     * Call only when holding lock.
     */
    private void enqueue(E x) {
        // assert lock.getHoldCount() == 1;
        // assert items[putIndex] == null;
        final Object[] items = this.items;
        items[putIndex] = x;
        if (++putIndex == items.length)
            putIndex = 0;
        count++;
        notEmpty.signal();
    }
```

More on this can be found under the "bound buffer" problem all over the internet :)

colinsurprenant:
For the record 
> fixed => the blocking queue was implemented wrong.

I would rephrase that for a little bit more precision and contextual information to help anyone reading this to avoid having to ingest the whole #7382 which lead to this fix:

This fixes a concurrency bug in the management of the queue empty state. Specifically the following sequence could lead to a deadlock:

A- reader 1 waits on notEmpty
B- reader 2 waits on notEmpty
C- writer acquires lock, write element and signals notEmpty, lock is unlocked 
D- a reader is selected to be waked up and tries to acquire the lock
E- a writer is "simultaneously" trying to acquire the lock and win in (F)
F- writer acquires lock, write element and **does not** signals notEmpty because queue is not empty at this point
G- readers are blocked.

I agree that the correct fix is to simply remove the condition when signalling non-emptyness 
```
if (wasEmpty) { notEmpty.signal(); }
```
to
```
notEmpty.signal();
```

Note that it seems we have never witnessed this phenomenon in logstash usage probably because at a higher level the reader and writers are also locked. 

I believe we should also backport this into the 5.4 branch for if a minor release is done on 5.4. Even if it seem to not occur in normal logstash usage, I would play safe here. 

LGTM.

suyograo:
yeah, I would backport it to 5.4 branch as well.

original-brownbear:
@suyograo @colinsurprenant 5.4 version pushed https://github.com/elastic/logstash/commit/047291db6954ad8149f9b32cfb8bbdff5171688d :)

PR #7384, opened 2017-06-08
ci: provides TLS certificates in base docker image

### LINKED ISSUE #7098: Move Filebeat Service in ITs to Docker
It's in the title, move the setup and teardown of the integration test Filebeat service to Docker like it was done in #7093 for Kafka.

wainersm:
Some test cases uses TLS certificates to check
the behavior with trusted communication among the
services.

This change generates the certificates in the base
docker image and add ServiceContainer#certificates
method so that the files can be copied to host.

An support method (ServiceContainer#envvar) is
also added. It gets an environment variable value
exported within the container.

In preparation to Fix #7098

### DIFF
--- a/qa/integration/services/dockerized/Dockerfile
+++ b/qa/integration/services/dockerized/Dockerfile
@@ -16,6 +16,11 @@ USER tester
 ENV WORKDIR /home/tester
 WORKDIR $WORKDIR
 
+# Generate TLS certificates that can be used by services.
+ENV CERTIFICATES_DIR $WORKDIR/certificates
+RUN mkdir -p $CERTIFICATES_DIR && \
+openssl req -subj '/CN=localhost/' -x509 -days $((100 * 365)) -batch -nodes -newkey rsa:2048 -keyout $CERTIFICATES_DIR/certificate.key -out $CERTIFICATES_DIR/certificate.crt
+
 # Script with routines that can be used in derived images.
 ADD helpers.sh .
 
--- a/qa/integration/services/service_container.rb
+++ b/qa/integration/services/service_container.rb
@@ -60,4 +60,29 @@ class ServiceContainer < Service
     @container.stop
     @container.delete(:force => true, :volumes => true)
   end
+
+  # Get environment variable from container's ENV.
+  def envvar(var)
+    # Exec command returns an array with [[stdout], [stderr], exit_code] layout.
+    res = @container.exec(['printenv', var])
+    res[0].join(" ").chomp
+  end
+
+  # Get path to certificates directory.
+  def certificates_dir
+    envvar("CERTIFICATES_DIR")
+  end
+
+  # Copy certificates from container to directory in the host.
+  def certificates(dest_dir)
+    FileUtils.mkdir_p(dest_dir)
+    cert_dir = certificates_dir
+    ["certificate.key", "certificate.crt"].each do |file|
+      File.open(File.join(dest_dir, file), "w") do |io|
+        # Use container's file separator (i.e. Linux)
+        io.write(@container.read_file(cert_dir.chomp + "/" + file))
+      end
+    end
+  end
+
 end


original-brownbear:
@wainersm thanks, will take a look very soon :)

wainersm:
Thanks @original-brownbear !
BTW, I'll need to add a method to return the certificates directory path in the container.

qa/integration/services/service_container.rb
```diff
@@ -60,4 +60,24 @@ def stop_container
...
+
+  # Get environment variable from container's ENV.
+  def envvar(var)
+    # Exec command returns an array with [[stdout], [stderr], exit_code] layout.
+    res = @container.exec(['printenv', var])
+    res[0].join(" ").chomp
+  end
+
+  # Copy certificates from container to directory in the host.
+  def certificates(dest_dir)
```
original-brownbear: @wainersm we're doing this, because we want to use the cert files with the Logstash that is running on the host right? (sorry this is probably an obvious question :))

wainersm: Exactly. Unless we decide to push Logstash under test to a container too, but I can't see advantages on doing that (besides increase complexity). :)

original-brownbear: @wainersm ah actually there is (admittedly no obvious ones at all :D) reasons for doing that:

* we want to test logstash install without internet connection, this is currently done in a not so portable way => easy to script up with Docker
* Kafka is currently a pain outside CI. We have to run it from a fixed port because of the way its discovery works and it binds to a Socket in a lingering way => repeat runs on local fail if done in rapid succession :( (at least on my box)
* Cleanup for CTRL+C is a lot faster and cleaner than for a local Ruby process

All kinda small, but together it would be nice to have these fixed (imo) right? :)

qa/integration/services/dockerized/Dockerfile
```diff
@@ -16,6 +16,11 @@ USER tester
 ENV WORKDIR /home/tester
 WORKDIR $WORKDIR
 
+# Generate TLS certificates that can be used by services.
+ENV CERTIFICATES_DIR $WORKDIR/certificates
+RUN mkdir -p $CERTIFICATES_DIR && \
```
original-brownbear: @wainersm I thought about this a bit this morning ... I think we may be introducing a strange bug here, given that this is our base image.
Couldn't we potentially run into a situation (not on CI but locally) where we build two child images from different versions of the base?

Honestly ... why not simply commit one key for tests to the repo and not have any noise from this stuff? We can then even access it from host and docker images trivially by just `ADD`ing it here instead of generating it?

wainersm: I don't have the rationale but certificates used to committed in the repo. The PR#5928 deleted the files so that now filebeat test generate them on-the-fly. Perhaps @suyograo knows the reason?

About the other doubt, I don't know if the scenario of build two child images from different versions of the base can even happen.
One intermediate solution (neither generate in the image nor commit in repo) can be a helper that create the certificates in a temporary folder in host. Then it's up to tests requiring them to copy over the containers.

original-brownbear: @wainersm yea it can, already been there :D I'm all for your suggestion of just creating the certificate on the host and then passing it to whatever part of the tests needs it :) That should be safe imo. Let's do it that way around?

wainersm: Great, we got a deal! :)

original-brownbear:
@wainersm want some help here? Let's not let this one become stale :)

wainersm:
Hi @original-brownbear, I've made progress since last time we talked but it stalled afterwards due lack of time. Well, I will resume the work, try to send final version by end of this week.

Sorry for that! :(

original-brownbear:
@wainersm no worries (and no rush ;) )! Was just checking in to make sure we don't lose track of this :)

original-brownbear:
Gotta close this one sorry, we changed our CI setup quite a bit and stopped using dynamically set up Docker containers.

PR #7393, opened 2017-06-09
Use new ES Output for DLQ issue

suyograo:


Gemfile.jruby-1.9.lock.release
```diff
@@ -724,7 +723,6 @@ DEPENDENCIES
   logstash-output-udp
   logstash-output-webhdfs
   logstash-output-xmpp
-  logstash-patterns-core
```
suyograo: not sure why this got removed

ph:
https://github.com/suyograo/logstash/blob/c516a3ede8d4fd51546dae561fcbb1ed42e4642a/Gemfile.jruby-1.9.lock.release#L491-L492

ph:
its already present in the lock file,  but pattern-core is not a direct dependency and should not be list in the part of the lockfile were you see it removed.

PR #7394, opened 2017-06-09
MINOR: Removed Redundant Java Wrapper

original-brownbear:
This file just trivially wraps 2 Java methods, is only used in a single place and contains no logic.
=> No need for it to pollute the production codebase :)


suyograo:
LGTM

PR #7400, opened 2017-06-09
PERFORMANCE: Queue RW Benchmark Baseline

original-brownbear:
We were missing a baseline in the throughput benchmark. I think `ArrayBlockingQueue` is the most optimal non-persistent solution here, so we should always have that in mind when optimizing serialization and queuing.

Results:

```sh
Benchmark                                     Mode  Cnt    Score    Error   Units
QueueRWBenchmark.readFromArrayBlockingQueue  thrpt   10  494.100 ± 28.146  ops/ms
QueueRWBenchmark.readFromPersistedQueue      thrpt   10   45.370 ± 13.375  ops/ms
```

### DIFF
--- a/logstash-core/benchmarks/src/main/java/org/logstash/benchmark/QueueRWBenchmark.java
+++ b/logstash-core/benchmarks/src/main/java/org/logstash/benchmark/QueueRWBenchmark.java
@@ -3,6 +3,7 @@ package org.logstash.benchmark;
 import com.google.common.io.Files;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


original-brownbear:
@jordansissel you know in a sense you could make a good argument that the numbers above should be equal (or at least closeish) as long as the queue has fewer than `1024` (ack interval elements) in it, wouldn't you agree?

original-brownbear:
btw in memory queue produces:
```
QueueRWBenchmark.readFromPersistedQueue      thrpt   10  136.636 ± 13.222  ops/ms
```
at default settings

Issue #7399, opened 2017-06-09
review Queue notFull signalling 


colinsurprenant:
in light of #7380 and #7382 we should review the `Queue` class `notFull` signalling to make sure a deadlock cannot occur at queue full conditions similar to what we fixed at queue empty condition.

FYI @original-brownbear

original-brownbear:
@colinsurprenant agreed, I guess we could try and replicate the test I added for the full queue case?

original-brownbear:
closed via #7405

PR #7405, opened 2017-06-09
 #7399 stress test full queue situation

original-brownbear:
For #7399 looks good @colinsurprenant <= this one passes fine (just the empty queue test using a queue size of 50 to force the queue getting filled up).

### DIFF
--- a/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
+++ b/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
@@ -563,51 +563,13 @@ public class QueueTest {
     }
 
     @Test
-    public void queueStableUnderStress() throws Exception {
-        Settings settings = TestSettings.persistedQueueSettings(1000000, dataPath);
-        final ExecutorService exec = Executors.newScheduledThreadPool(2);
-        try (Queue queue = new Queue(settings)) {
-            final int count = 20_000;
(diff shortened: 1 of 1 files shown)


logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -755,4 +717,51 @@ public void getsPersistedByteSizeCorrectlyForUnopened() throws Exception {
             assertThat(q.getPersistedByteSize(), is(0L));
         }
     }
+    
+    private void stableUnderStress(final int capacity) throws IOException {
+        Settings settings = TestSettings.persistedQueueSettings(capacity, dataPath);
+        final ExecutorService exec = Executors.newScheduledThreadPool(2);
+        try (Queue queue = new Queue(settings)) {
+            final int count = 20_000;
+            final int concurrent = 2;
```
ph: This might be a good time to use theses values as parameter? WDYT?

original-brownbear: @ph not so sure about that, for now we're just using one set of values here, no need to duplicate those upstream right?

ph: I'm okay with that, if the need ever come, we can refactor them.

PR #7408, opened 2017-06-09
Optimize LIR pipeline hashing

### LINKED ISSUE #7379: 6.0.0-alpha2 slow startup times
- Version: 6.0.0-alpha2
- Operating System: Ubuntu 14.04.5
- Config File: https://gist.github.com/ceeeekay/933ade5e28ce96198d33e85073173112
- Authlog Pattern: https://gist.github.com/ceeeekay/163e3051b02d5bef1c64286e5ef07ea4
- Steps to Reproduce: Start up Logstash 6.0.0-alpha2 with a large config

Note: this was also occurring with 6.0.0-alpha1 but I've only just got around to reporting it.

I've attempted to narrow this down to a specific part of the config but I'm unable to do this.

Tests below are made using a modified and reduced config (see gist).

**5.4.1 - 12 second startup time**
```
# dpkg -l | grep logstash
ii  logstash                                              1:5.4.1-1 

# time sudo -u logstash /usr/share/logstash/bin/logstash --path.settings /etc/logstash --path.config /etc/logstash/enabled/
Sending Logstash's logs to /var/log/logstash which is now configured via log4j2.properties
test

real	0m12.493s
user	0m37.547s
sys	0m0.737s
```

**6.0.0-alpha2 - 39 second startup time**
```
# dpkg -l | grep logstash
ii  logstash                                              1:6.0.0~alpha2-1 

# time sudo -u logstash /usr/share/logstash/bin/logstash --path.settings /etc/logstash --path.config /etc/logstash/enabled/
Sending Logstash's logs to /var/log/logstash which is now configured via log4j2.properties
test

real	0m39.526s
user	1m21.297s
sys	0m0.903s
```

I've tried to keep the config to a minimum for this test, however the longer the config, the larger the difference in startup times between 5.4.1 and 6.0.0-alpha2. 

Doubling the size of the test config by duplicating it in the config directory increases the 6.0.0 startup time from 40 seconds to **5 minutes**, compared to 15 seconds for 5.4.1. The full dev config I'm running (2739 lines) takes over 12 minutes to start with 6.0.0, and only 16 seconds with 5.4.1.

andrewvc:
Fixes #7379

Prior to this patch large configuration files would take a long time to load, sometimes, so long that any sane user would just give up. This patch improves the graph hashing algorithm in the following ways:

1. Caches expensive to compute paths
2. Use parallel computation where possible to speed calculations
3. Removes unnecessary validation on Graph#chainVertices
4. Validates the entire graph in PipelineIR which allows the cache to be used for both validation and hashing
5. Validation now only checks for duplicate IDs, not duplicate hashes. If the user supplies their own IDs this check will be sped up.
6. Hashing only consists of calculating the unique hash for the leaf nodes. Since hashing is sensitive to the full ancestry of a given node this is sufficient vs. previously where each vertex's hash was combined. We will still calculate the hash for each vertex if no ID is specified however.

Benchmark results on the new branch look good. Configs used were [bigconfig.config](https://gist.githubusercontent.com/andrewvc/6251a238ecba54510d34ec348168dde8/raw/dbfd148a3a2711f96cd9db9109bd3345b9a63644/bigconfig.config) and [smallconfig.config](https://gist.githubusercontent.com/andrewvc/6251a238ecba54510d34ec348168dde8/raw/dbfd148a3a2711f96cd9db9109bd3345b9a63644/smallconfig.config). Results posted below:

```
~/p/logstash (optimize-pipeline-hashing) $ time bin/logstash -f smallconfig.config
Sending Logstash's logs to /Users/andrewvc/projects/logstash/logs which is now configured via log4j2.properties
[2017-06-09T16:37:34,951][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because 'path.config' (-f) is being used.
[2017-06-09T16:37:34,993][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2017-06-09T16:37:35,354][INFO ][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"main", :thread=>"#<Thread:0x194f3bde run>", "pipeline.workers"=>8, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>1000}
[2017-06-09T16:37:35,358][INFO ][logstash.pipeline        ] Pipeline started {"pipeline.id"=>"main"}
[2017-06-09T16:37:35,364][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
2017-06-09T21:37:35.368Z andrewvc-elastic Hello world!
[2017-06-09T16:37:35,878][INFO ][logstash.pipeline        ] Pipeline terminated {"pipeline.id"=>"main"}
        8.52 real        30.56 user         1.15 sys
~/p/logstash (optimize-pipeline-hashing) $ time bin/logstash -f bigconfig.config
Sending Logstash's logs to /Users/andrewvc/projects/logstash/logs which is now configured via log4j2.properties
[2017-06-09T16:37:49,420][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because 'path.config' (-f) is being used.
[2017-06-09T16:37:49,467][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2017-06-09T16:37:58,455][INFO ][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"main", :thread=>"#<Thread:0x5d006478 run>", "pipeline.workers"=>8, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>1000}
[2017-06-09T16:37:58,533][INFO ][logstash.pipeline        ] Pipeline started {"pipeline.id"=>"main"}
[2017-06-09T16:37:58,547][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
2017-06-09T21:37:58.546Z andrewvc-elastic Hello world!
[2017-06-09T16:37:59,060][INFO ][logstash.pipeline        ] Pipeline terminated {"pipeline.id"=>"main"}
       16.84 real        71.79 user         1.56 sys
```

compared to before this commit:

```
~/p/logstash (master) $ time bin/logstash -f smallconfig.config
Sending Logstash's logs to /Users/andrewvc/projects/logstash/logs which is now configured via log4j2.properties
[2017-06-09T16:48:20,142][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because 'path.config' (-f) is being used.
[2017-06-09T16:48:20,187][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2017-06-09T16:48:24,581][INFO ][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"main", :thread=>"#<Thread:0x7866ed16 run>", "pipeline.workers"=>8, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>1000}
[2017-06-09T16:48:24,586][INFO ][logstash.pipeline        ] Pipeline started {"pipeline.id"=>"main"}
[2017-06-09T16:48:24,598][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
2017-06-09T21:48:24.607Z andrewvc-elastic Hello world!
[2017-06-09T16:48:25,116][INFO ][logstash.pipeline        ] Pipeline terminated {"pipeline.id"=>"main"}
       13.38 real        43.93 user         1.42 sys

# I Had to terminate execution here after 10m because it was still running!
~/p/logstash (master) $ time bin/logstash -f bigconfig.config
Sending Logstash's logs to /Users/andrewvc/projects/logstash/logs which is now configured via log4j2.properties
[2017-06-09T16:39:27,507][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because 'path.config' (-f) is being used.
[2017-06-09T16:39:27,559][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2017-06-09T16:47:21,963][WARN ][logstash.runner          ] SIGTERM received. Shutting down the agent.
Command terminated abnormally.
      485.85 real       571.93 user         5.31 sys
```

### DIFF
--- a/logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
+++ b/logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
@@ -222,7 +222,6 @@ public class Graph implements SourceComponent, Hashable {
 
     public Collection<Edge> chainVertices(Edge.EdgeFactory edgeFactory, Vertex... argVertices) throws InvalidIRException {
         Collection<Edge> edges = chainVerticesUnsafe(edgeFactory, argVertices);
-        validate();
         return edges;
     }
 
@@ -246,6 +245,7 @@ public class Graph implements SourceComponent, Hashable {
     // in.
     public void refresh() throws InvalidIRException {
         this.calculateRanks();
+        this.vertices.forEach(Vertex::clearCache);
         this.calculateTopologicalSort();
     }
 
@@ -269,7 +269,7 @@ public class Graph implements SourceComponent, Hashable {
     }
 
     public Map<String, List<Vertex>> verticesByHash() {
-        return this.vertices().collect(Collectors.groupingBy(Vertex::uniqueHash));
+        return this.vertices().parallel().collect(Collectors.groupingBy(Vertex::uniqueHash));
     }
 
     public void validate() throws InvalidIRException {
@@ -279,14 +279,21 @@ public class Graph implements SourceComponent, Hashable {
             throw new InvalidIRException("Graph has no leaf vertices!\n" + this.toString());
         }
 
-        List<List<Vertex>> duplicates = verticesByHash().values().stream().filter((group) -> group.size() > 1).collect(Collectors.toList());
-        if (!duplicates.isEmpty()) {
-            Stream<String> errorMessageGroups = duplicates.stream().
-                    map((group) -> group.stream().map(Object::toString).collect(Collectors.joining("===")));
+        // Check for duplicate IDs in the config
+        List<String> duplicateIdErrorMessages = this.vertices().parallel()
+                .collect(Collectors.groupingBy(Vertex::getId))
+                .values()
+                .stream()
+                .filter(group -> group.size() > 1)
+                .map(group -> {
+                    return "ID: " + group.stream().findAny().get().getId() + " " +
+                            group.stream().map(Object::toString).collect(Collectors.joining(","));
+                })
+                .collect(Collectors.toList());
 
-            String joinedErrorMessageGroups = errorMessageGroups.collect(Collectors.joining("\n---\n"));
-
-            throw new InvalidIRException("Some nodes on the graph are fully redundant!\n" + this + "|" + joinedErrorMessageGroups);
+        if (!duplicateIdErrorMessages.isEmpty()) {
+            String dupeErrors = duplicateIdErrorMessages.stream().collect(Collectors.joining("\n"));
+            throw new InvalidIRException("Config has duplicate Ids: \n" + dupeErrors);
         }
     }
 
@@ -422,7 +429,16 @@ public class Graph implements SourceComponent, Hashable {
     @Override
     public String hashSource() {
         MessageDigest lineageDigest = Util.defaultMessageDigest();
-        List<byte[]> sources = this.vertices.stream().parallel().map(Vertex::uniqueHash).sorted().map(String::getBytes).collect(Collectors.toList());
+
+        // We only need to calculate the hashes of the leaves since those hashes are sensitive to changes
+        // anywhere else on the graph. It would also be OK to use the roots, the decision is arbitrary
+        List<byte[]> sources = this.leaves().parallel()
+                .map(Vertex::uniqueHash)
+                .sorted()
+                .map(String::getBytes)
+                .collect(Collectors.toList());
+        // We don't do this inline with the stream because we need this to be single threaded
+        // not parallel
         sources.forEach(lineageDigest::update);
         return Util.bytesToHexString(lineageDigest.digest());
     }
--- a/logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
+++ b/logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
@@ -268,10 +268,6 @@ public class Graph implements SourceComponent, Hashable {
         return rank;
     }
 
-    public Map<String, List<Vertex>> verticesByHash() {
-        return this.vertices().parallel().collect(Collectors.groupingBy(Vertex::uniqueHash));
-    }
-
     public void validate() throws InvalidIRException {
         if (this.isEmpty()) return;
 
@@ -426,8 +422,7 @@ public class Graph implements SourceComponent, Hashable {
         return this.edges.stream();
     }
 
-    @Override
-    public String hashSource() {
+    public String uniqueHash() {
         MessageDigest lineageDigest = Util.defaultMessageDigest();
 
         // We only need to calculate the hashes of the leaves since those hashes are sensitive to changes
--- a/logstash-core/lib/logstash/compiler/lscl.rb
+++ b/logstash-core/lib/logstash/compiler/lscl.rb
@@ -46,10 +46,6 @@ module LogStashCompilerLSCLGrammar; module LogStash; module Compiler; module LSC
       [self.input.line_of(start), self.input.column_of(start)]
     end
 
-    def empty_source_meta()
-      org.logstash.common.SourceWithMetadata.new(base_protocol, base_id, nil)
-    end
-
     def jdsl
       org.logstash.config.ir.DSL
     end
@@ -99,11 +95,12 @@ module LogStashCompilerLSCLGrammar; module LogStash; module Compiler; module LSC
         end
       end
 
+      compiled_section_map = {}
       section_map.keys.each do |key|
-        section_map[key] = compose_for(key).call(empty_source_meta, *section_map[key])
+        compiled_section_map[key] = compose_for(key).call(*section_map[key])
       end
 
-      section_map
+      compiled_section_map
     end
   end
 
--- a/logstash-core/lib/logstash/config/source/local.rb
+++ b/logstash-core/lib/logstash/config/source/local.rb
@@ -19,7 +19,7 @@ module LogStash module Config module Source
   class Local < Base
     class ConfigStringLoader
       def self.read(config_string)
-        [org.logstash.common.SourceWithMetadata.new("string", "config_string", config_string)]
+        [org.logstash.common.SourceWithMetadata.new("string", "config_string", 0, 0, config_string)]
       end
     end
 
@@ -54,7 +54,7 @@ module LogStash module Config module Source
           config_string = ::File.read(file)
 
           if valid_encoding?(config_string)
-            part = org.logstash.common.SourceWithMetadata.new("file", file, config_string)
+            part = org.logstash.common.SourceWithMetadata.new("file", file, 0, 0, config_string)
             config_parts << part
           else
             encoding_issue_files << file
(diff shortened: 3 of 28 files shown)


logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -222,7 +223,6 @@ private Graph chain(Collection<Vertex> fromLeaves, Collection<Vertex> toVertices
 
     public Collection<Edge> chainVertices(Edge.EdgeFactory edgeFactory, Vertex... argVertices) throws InvalidIRException {
         Collection<Edge> edges = chainVerticesUnsafe(edgeFactory, argVertices);
-        validate();
```
andrewvc: We always validate the PipelineIR object, this is a nice invariant, but now that we know speed is a concern, just not worth it.

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -269,7 +269,7 @@ public Integer rank(Vertex vertex) {
     }
 
     public Map<String, List<Vertex>> verticesByHash() {
-        return this.vertices().collect(Collectors.groupingBy(Vertex::uniqueHash));
+        return this.vertices().parallel().collect(Collectors.groupingBy(Vertex::uniqueHash));
```
original-brownbear: @andrewvc why parallel? The stream you get back won't be `CONCURRENT`will it?
screen shot 2017-06-10 at 21 24 32

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -279,14 +279,21 @@ public void validate() throws InvalidIRException {
...
-        List<List<Vertex>> duplicates = verticesByHash().values().stream().filter((group) -> group.size() > 1).collect(Collectors.toList());
-        if (!duplicates.isEmpty()) {
-            Stream<String> errorMessageGroups = duplicates.stream().
-                    map((group) -> group.stream().map(Object::toString).collect(Collectors.joining("===")));
-
-            String joinedErrorMessageGroups = errorMessageGroups.collect(Collectors.joining("\n---\n"));
-
-            throw new InvalidIRException("Some nodes on the graph are fully redundant!\n" + this + "|" + joinedErrorMessageGroups);
+        // Check for duplicate IDs in the config
+        List<String> duplicateIdErrorMessages = this.vertices().parallel()
```
original-brownbear: @andrewvc As in the above, I don't think `parallel` get's us much here. 

Also, do we care about perfromance a lot here? If so, we should probably just move away from all the functional magic here, it's really slow in Java :) If we're happy with the performance now I'd just drop the `parallel()` to kill some dead code :)

original-brownbear:
@andrewvc @suyograo hah, tricky review for me :) Logic of caching the hashes looks fine to me on an immutable graph. Made a comment on the redundant `parallel` that may be worth cleaning up, other than that I'm not sure what you want to achieve here?
Do you want to speed this up more (seems it's still taking a bunch of seconds to go through the config?) I could add suggestions, there's likely an order of magnitude to be had here with some adjustments?

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
andrewvc: when I benchmarked this during dev it yielded an improvement .

Reading through https://docs.oracle.com/javase/tutorial/collections/streams/parallelism.html seems to confirm that this should in fact work. `.stream().parallel()` can modify a collection's stream to behave as `.parallelStream` would.

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
andrewvc: Yes, `getId()` can be expensive since it usually has to compute the hash which is expensive. The parallel does work as indicated above unless I'm mistaken and my previous benchmark had a mistake in it.

andrewvc:
@original-brownbear I'd love to see a larger speedup.I believe its definitely possible. In hindsight I wish I'd read a few papers before tackling this problem. This one comes up as interesting: https://arxiv.org/pdf/1606.00001.pdf

If you have any ideas as far as reducing the complexity of hashing the graph + vertices that'd be awesome :)

andrewvc:
BTW, if you look at the disparity between user and wallclock time from my benchmark runs you'll see that we definitely are making use of that FJP threadpool.

andrewvc:
I should probably also link to https://en.wikipedia.org/wiki/Graph_isomorphism_problem . Also, I should mention that the first paper I linked to lists its solution complexity as O(N^2). 

I'm not sure what the complexity of the algorithm I've implemented is, but it seems fast enough based on the benchmarks done so far (and dramatically better than what's in master now). It's probably worth doing the same benchmark on an input that's double the size of `bigconfig` to get an idea of how the complexity scales in the real world.

That being said, I think that the speed it runs at now is more than acceptable, and I think the nice property of a consistent hash regardless of specific input format, regardless of whitespace etc, is pretty nice.

Stepping back, however, perhaps we should just require a unique serialized input source (which today is always a string) and just hash that value instead. This whole feature may be *too* forward thinking.

andrewvc:
The only irritating thing there is then we need a new method of consistently labeling nodes in a given order, and we'd require ID injection at parse time, not in the middle of the IR.

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
original-brownbear: @andrewvc you win, Java8 is crazy :) ..interesting that this is faster here :)

original-brownbear:
@andrewvc

>  I'd love to see a larger speedup.I believe its definitely possible. In hindsight I wish I'd read a few papers before tackling this problem. 

I was thinking of a few points on the implementation of the operations that were changed here, I don't have a better overall approach to the algorithm, you overestimate me :)
But yea, if you say it's acceptable I wouldn't put time into this yet.

LGTM then :)

logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
original-brownbear: @andrewvc are we losing most of the time on `getId()` ?

andrewvc:
@original-brownbear thinking about this today, and testing it against an input double the size of `bigtest` I found that performance increased by a factor of 4, which is unacceptable. I've introduced a [new commit](https://github.com/elastic/logstash/pull/7408/commits/5cb25de69e8d437dabb06de6805b114da9b85827) to this PR that further improves it and gets its performance inline with Logstash 5.x

This commit further improves hashing performance, using SourceWithMetadata objects where present to determine Vertex IDs and the hash of the source file. If those objects are not present LIR will revert to graph hashing, which is slower, but always works. This is still useful for things like testing, so it makes sense to leave that code in. In the future it might be nice to extract the hashing code using the strategy pattern to clean things up.

This commit also improves the Hashable interface, moving hashSource into its own interface, letting classes that do their own hashing simply implement uniqueHash instead of requiring them to also implement hashSource.

A requirement of these changes was also that SourceWithMetadata no longer accept null or empty arguments. This patch now enforces that requirement. This also is nicer from an API standpoint, as you now know that if a SourceWithMetadata object exists, it will actually have all
its fields in use.

@original-brownbear mind re-reviewing this? Or, if you'd prefer I can merge this PR in as-is

andrewvc:
please test this

andrewvc:
please test this

original-brownbear:
@andrewvc sure will take a look very soon :) (~1-2h)

original-brownbear:
@andrewvc hmm Travis is failing can you take a look, otherwise LGTM :)

PR #7410, opened 2017-06-10
PERFORMANCE: Queue RW benchmark for in memory queue

original-brownbear:
Just to formalize the result here https://github.com/elastic/logstash/pull/7400#issuecomment-307503452

I also removed the artificial `Event` cloning. I added this in the first version of the benchmark to get a more realistic picture of the serialization cost. As it turns out this is bounding all cases except for the PQ one and hence just hurts correctness.

Example Results with `master`:

```sh
Benchmark                                     Mode  Cnt     Score      Error   Units
QueueRWBenchmark.readFromArrayBlockingQueue  thrpt   10  7988.533 ± 1015.737  ops/ms
QueueRWBenchmark.readFromMemoryQueue         thrpt   10   202.685 ±   22.186  ops/ms
QueueRWBenchmark.readFromPersistedQueue      thrpt   10    50.406 ±   14.849  ops/ms
```

### DIFF
--- a/logstash-core/benchmarks/src/main/java/org/logstash/benchmark/QueueRWBenchmark.java
+++ b/logstash-core/benchmarks/src/main/java/org/logstash/benchmark/QueueRWBenchmark.java
@@ -10,14 +10,17 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.io.FileUtils;
 import org.logstash.Event;
-import org.logstash.Timestamp;
 import org.logstash.ackedqueue.Batch;
 import org.logstash.ackedqueue.Queue;
 import org.logstash.ackedqueue.Queueable;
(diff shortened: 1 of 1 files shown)


suyograo:
Curious why `readFromArrayBlockingQueue` have a higher error rate (14%)?

suyograo:
LGTM

original-brownbear:
@suyograo 

> Curious why readFromArrayBlockingQueue have a higher error rate (14%)?

I think the high-level explanation is that we're essentially benchmarking context switching only in the `ArrayBlockQueue` case, which has a lot of jitter (especially when ran on a system that is a little noisy, had a browser open etc. => def hitting those cores running here a few times during the benchmark), while for the other cases we're benchmarking memory and the I/O system which is a lot more deterministic in its behaviour.

Issue #6696, opened 2017-02-13
Environment variables sometimes aren't working?


jordansissel:
I'm not sure what causes this, but it's easy to reproduce. Reproduced only with Elasticsearch output but haven't tried other plugins.

for openers:
* Environment variables seem to work fine, and we have tests for them.
* Except in this one case:

```
% ES_HOST=localhost bin/logstash -e 'output { elasticsearch { hosts => "${ES_HOST}" } }'
[2017-02-13T02:01:56,198][ERROR][logstash.agent           ] fetched an invalid config {:config=>"output { elasticsearch { hosts => \"${ES_HOST}\" } }input { stdin { type => stdin } }", :reason=>"bad URI(is not URI?): ${ES_HOST}"}
```

The specific part of the error for focus: `bad URI(is not URI?): ${ES_HOST}"}`

I tested other settings (heartbeat_path) and env vars are working there.

pigri:
+1

ValentinOdier:
Hi,

We have this issue too and we think it's a regression.

Using logstash 5.2.2

```
15:22:34.214 [LogStash::Runner] DEBUG logstash.outputs.elasticsearch - config LogStash::Outputs::ElasticSearch/@hosts = [http://elasticsearch:9200]
15:22:34.214 [LogStash::Runner] DEBUG logstash.outputs.elasticsearch - config LogStash::Outputs::ElasticSearch/@index = "datascience-testing"
15:22:34.214 [LogStash::Runner] DEBUG logstash.outputs.elasticsearch - config LogStash::Outputs::ElasticSearch/@flush_size = 250
15:22:34.214 [LogStash::Runner] DEBUG logstash.outputs.elasticsearch - config LogStash::Outputs::ElasticSearch/@id = "cbf411562ac0e019174e618a8f7c7ad6fdbb21a3-2"
15:22:34.214 [LogStash::Runner] DEBUG logstash.outputs.elasticsearch - config LogStash::Outputs::ElasticSearch/@enable_metric = true
15:22:34.215 [LogStash::Runner] DEBUG logstash.outputs.elasticsearch - config LogStash::Outputs::ElasticSearch/@codec = <LogStash::Codecs::Plain id=>"plain_c60ee176-dc79-4533-9be5-683eb2e8a935", enable_metric=>true, charset=>"UTF-8">
[...]

PR #7411, opened 2017-06-10
 #6696 validate evaluation result

original-brownbear:
For #6696 

The interpolation for the ES hosts breaks because we are running validation against the original, not interpolated values. So `${ES_HOST}` is rejected as value for a `:uri` field obviously.

Running validation against the interpolated values fixes the issue. Is there anything wrong with this approach? If not, happy to add a spec for this :)

### DIFF
--- a/logstash-core/spec/logstash/config/mixin_spec.rb
+++ b/logstash-core/spec/logstash/config/mixin_spec.rb
@@ -369,11 +369,13 @@ describe LogStash::Config::Mixin do
       before do
         ENV["FunString"] = "fancy"
         ENV["FunBool"] = "true"
+        ENV["SERVER_LS_TEST_ADDRESS"] = "some.host.address.tld"
       end
 
       after do
         ENV.delete("FunString")
         ENV.delete("FunBool")
+        ENV.delete("SERVER_LS_TEST_ADDRESS")
       end
 
       subject do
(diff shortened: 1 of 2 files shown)


original-brownbear:
@jordansissel @suyograo FYI, is this a correct fix?

original-brownbear:
hah it fails tests now ... since validate runs on the unreplaced values don't waste your time on this yet

original-brownbear:
@suyograo @jordansissel ok passes now and works fine for the example given in the issue.

Added interpolation of env to whatever is saved in the `original_settings` field as well so that validation works. IMO, since this is just used for logging purposes that's fine (and personally I'd rather have the interpolated version logged even) but not sure if you guys agree :)?

suyograo:
@original-brownbear can we add a spec from the example in #6696 ?

original-brownbear:
@suyograo sure was just waiting to get some validation of the approach before I do that :) will do

original-brownbear:
@suyograo spec that reproduces the issue in #6696 and now passes added :)

logstash-core/spec/logstash/config/mixin_spec.rb
```diff
@@ -397,6 +399,17 @@ def initialize(params)
         expect(subject.nestedArray).to(be == { "level1" => [{ "key1" => "http://fancy:8080/blah.txt" }, { "key2" => "http://fancy:8080/foo.txt" }] })
         expect(subject.deepHash).to(be == { "level1" => { "level2" => { "level3" => { "key1" => "http://fancy:8080/blah.txt" } } } })
       end
+
+      it "should validate settings after interpolating ENV variables" do
+        expect {
+          Class.new(LogStash::Filters::Base) do
+            include LogStash::Config::Mixin
+            config_name "test"
+            milestone 1
```
suyograo: BTW, I don't think you need this `milestone`

PR #7412, opened 2017-06-10
Handle ingest on_failure processor

suyograo:
Ingest `on_failure` processor allows to define a new sub-piplene when a failure
happens in the current processor. Logstash does not have an on_failure but this can
be emulated using the tag and conditional.

Also added a set processor converter which is typically used in `on_failure`.

### DIFF
--- a/tools/ingest-converter/src/main/resources/ingest-pipeline.js
+++ b/tools/ingest-converter/src/main/resources/ingest-pipeline.js
@@ -15,7 +15,7 @@ function ingest_pipeline_to_logstash(json) {
         var filter_blocks = [];
         if (IngestGrok.has_grok(processor)) {
             filter_blocks.push(
-                IngestConverter.create_hash("grok", IngestGrok.grok_hash(processor))
+                IngestConverter.create_hash(IngestGrok.get_name(), IngestGrok.grok_hash(processor))
             );
             if (IngestConverter.has_on_failure(processor, IngestGrok.get_name())) {
                 filter_blocks.push(
--- a/tools/ingest-converter/src/main/resources/ingest-pipeline.js
+++ b/tools/ingest-converter/src/main/resources/ingest-pipeline.js
@@ -3,9 +3,9 @@
  */
 function ingest_pipeline_to_logstash(json) {
 
-    function handle_on_failure_pipeline(on_failure_json) {
+    function handle_on_failure_pipeline(on_failure_json, tag_name) {
 
-        return IngestConverter.create_tag_conditional("_grokparsefailure",
+        return IngestConverter.create_tag_conditional(tag_name,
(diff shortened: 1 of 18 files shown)


tools/ingest-converter/src/main/resources/ingest-pipeline.js
```diff
@@ -3,13 +3,28 @@
...
+            IngestConverter.join_hash_fields(on_failure_json.map(map_processor))
+        );
+    }
+
     function map_processor(processor) {
 
         var filter_blocks = [];
         if (IngestGrok.has_grok(processor)) {
             filter_blocks.push(
                 IngestConverter.create_hash("grok", IngestGrok.grok_hash(processor))
```
original-brownbear: @suyograo I think we should just use `IngestGrok.get_name()` here as well, otherwise the fact this is supposed to be a constant becomes kinda obscure, right?

tools/ingest-converter/src/main/resources/ingest-shared.js
```diff
@@ -109,12 +109,38 @@ var IngestConverter = {
...
         return filters.join("\n\n") + "\n";
+    },
+
+    /**
+     * Does it have an on_failure field?
+     * @param processor Json
+     * @param name Name of the processor
+     * @returns {boolean} True if has on failure
+     */
+    has_on_failure: function (processor, name) {
```
original-brownbear: @suyograo should this be generalized by `name`? At least the implementation only uses this with grok right?

tools/ingest-converter/src/test/resources/org/logstash/ingest/logstashComplexCase4.conf
```diff
@@ -0,0 +1,31 @@
+filter {
+   grok {
+      match => {
+         "message" => "%{COMBINEDAPACHELOG}"
+      }
+   }
+   if "_grokparsefailure" in [tags] {
+         mutate {
```
original-brownbear: @suyograo the indentation is kinda messed up here, isn't it? this should only be indented by 3 spaces relative to the line below. Now we're closing this one in line 12 below, which looks wrong, doesn't it?

original-brownbear:
@suyograo looks good functionally, but the indentation seems wrong? + One random comment on the `get_name` function (that is very optional I guess).

tools/ingest-converter/src/main/resources/ingest-shared.js
suyograo: Yep. The plan is to use this in other processors such as geoip (which produces a failure tag called `_geoip_lookup_failure`), and others as well..

Only implemented for grok now though.

suyograo:
@original-brownbear fixed the indentation.

original-brownbear:
@suyograo LGTM :)

Issue #7375, opened 2017-06-07
rake `plugin:install-default` does not create Gemfile


andrewvc:
rake `test:install-default` works great, but the `plugin` variant does not check for this condition and fix it. Discovered in the 5.5 branch.

Probably a missing rake dep.

original-brownbear:
@andrewvc fixed via https://github.com/elastic/logstash/pull/7415 :)

PR #7415, opened 2017-06-12
#7375 plugin install rake task should depend on bootstrap task

original-brownbear:
Fixing #7375 as suggested by @andrewvc :

* make plugin install tasks depend on `bootstrap` to generate `Gemfile` in case its missing


original-brownbear:
@jsvd thanks :)

Issue #6828, opened 2017-03-22
Migrate Logstash's API spec to use `json-schema`


ph:
Currently, our tests for the API response are pretty high level and don't give us the flexibility to express all the options that could be returned by the API. As a follow-up of work from @jsvd on x-pack, I believe we should move the specs to use https://github.com/ruby-json-schema/json-schema instead of our helpers.

original-brownbear:
done in #7419

PR #7419, opened 2017-06-12
#6828 Use json-schema in tests

original-brownbear:
one step for #6828 :)

@ph this is what you had in mind right? (didn't do all specs yet in case the answer is "no" :))

### DIFF
--- a/Gemfile.template
+++ b/Gemfile.template
@@ -23,6 +23,7 @@ gem "rack-test", :require => "rack/test", :group => :development
 gem "flores", "~> 0.0.6", :group => :development
 gem "term-ansicolor", "~> 1.3.2", :group => :development
 gem "docker-api", "1.33.4", :group => :development
+gem "json-schema", "~> 2.6", :group => :development
 gem "pleaserun", "~>0.0.28"
 gem "logstash-input-heartbeat"
 gem "logstash-codec-collectd"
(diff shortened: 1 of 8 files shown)


ph:
Yes, but I am wondering if we could have some helper to write the schema?

original-brownbear:
@ph def, we should dry this up before merging imo. Let me fix the other cases real quick then and see what I can do about that :)

original-brownbear:
@ph added the remaining cases that I could at reasonable effort port. Two cases, namely:

```rb
  context "#values" do

    let(:payload) { LogStash::Json.load(last_response.body) }

    it "return totals of plugins" do
      expect(payload["total"]).to eq(payload["plugins"].count)
    end

    it "return a list of available plugins" do
      payload["plugins"].each do |plugin|
        expect do 
          Gem::Specification.find_by_name(plugin["name"])
        end.not_to raise_error
      end
    end
```

seem very hard to port since the schema validation has no comparison logic across fields.
I am starting to have some doubts about drying this up though. I think we should be careful here, in a sense we're substituting our own "DSL" for the schema doing that.
When looking at the concrete things we could dry up this seems like it might be more trouble than it's worth in terms of readability?

original-brownbear:
@ph can you give this another look whenever you find a sec? :)

Issue #7407, opened 2017-06-09
Add ephemeral_id to Logstash and Pipeline


ph:
To offer better metrics will need to better track the lifecycle of pipelines and Logstash instances when they are restarted or created. To do so we will add `ephemeral_id` at the **Logstash level** and at the **Pipeline level** and theses information will be available when we gather the metrics.

```
logstash.ephemeral_id
pipeline.ephemeral_id
```

We will need to make them available for x-pack and in the API.

ph:
cc @andrewvc

ph:
Concerning the implementation do we want to use `SecureRandom.uuid` like we do with the persisted UUID or `SecureRandom.urlsafe_base64`?

https://ruby-doc.org/stdlib-1.9.2/libdoc/securerandom/rdoc/SecureRandom.html

ph:
For consistency, I will probably go with `SecureRandom.uuid`

PR #7423, opened 2017-06-12
Add `ephemeral_id` on the Logstash instance and for the pipelines.

ph:
Theses id wont be persisted between restart of logstash or between
creation of pipeline. When we reload pipelines the ephemeral_id will be
generated again.

Fixes: #7407


ph:
rekick travis.. again a apt repository error..

PR #7424, opened 2017-06-12
The Task.leftShift(Closure) method has been deprecated

ph:
This the `<<` closure is now deprecated, this commit change it to the
`doLast(action)`, this will remove the warning in our build


robbavey:
LGTM

PR #7425, opened 2017-06-12
Fix/jruby 9.1.10.0

ph:
Update from 9.1.9.0 to 9.1.10.0

Make sure we use the same provided and runtime JRuby with our gradle projects


ph:
This will need a bit more work, bumping to 9.1.10.0 make the build fails on jenkins.

original-brownbear:
@ph LGTM, we should add an issue to extract the JRuby version to a single place though in my opinion :)

ph:
@original-brownbear Yes, I wanted to do that in another PR.

Issue #7317, opened 2017-06-05
Persisted Queue Performance Design Issue (Writing Data)


original-brownbear:
The current implementation in `master` uses `mmap` to write to fixed size pages that are created and written to sequentially.
This introduces a lot of overhead for the VM system, requires hacking around the GC's `Cleaner` behavior for the `mmap`ed buffers and hence comes with unpredictable and suboptimal performance.

https://github.com/elastic/logstash/pull/7316 contains an extremely trivial implementation illustrating the problem.

It simply writes to the channel directly and avoids `mmap` for writes. The durability features of the approach are exactly the same as those in the `master` implementation since `MappedBuffer.force()` is simply replaced by calling `Channel.force()` on the underlying `FileChannel`.

The branch passes all tests and the performance difference is about a factor of `3` (plus a lot more consistent throughput).

Before (`master`):

```sh
# Run progress: 0.00% complete, ETA 00:00:01
# Fork: 1 of 1
# Warmup Iteration   1: ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console.
78.110 ops/ms
# Warmup Iteration   2: 72.397 ops/ms
# Warmup Iteration   3: 54.313 ops/ms
Iteration   1: 42.278 ops/ms
Iteration   2: 71.426 ops/ms
Iteration   3: 66.277 ops/ms
Iteration   4: 55.886 ops/ms
Iteration   5: 45.107 ops/ms
Iteration   6: 57.468 ops/ms
Iteration   7: 59.576 ops/ms
Iteration   8: 54.282 ops/ms
Iteration   9: 45.083 ops/ms
Iteration  10: 50.365 ops/ms


Result "org.logstash.benchmark.QueueBenchmark.pushToPersistedQueue":
  54.775 ±(99.9%) 14.293 ops/ms [Average]
  (min, avg, max) = (42.278, 54.775, 71.426), stdev = 9.454
  CI (99.9%): [40.482, 69.068] (assumes normal distribution)


# Run complete. Total time: 00:01:57

Benchmark                             Mode  Cnt   Score    Error   Units
QueueBenchmark.pushToPersistedQueue  thrpt   10  54.775 ± 14.293  ops/ms

BUILD SUCCESSFUL

Total time: 2 mins 1.626 secs
```

After:

```sh
# Run progress: 0.00% complete, ETA 00:00:01
# Fork: 1 of 1
[...]

PR #7428, opened 2017-06-13
DON'T MERGE Sequential persistent queue experiment (or persisting 43% faster than current)

andrewvc:
Hi all,

NOTE: This implementation is very fast, but makes some weird tradeoffs. Please read https://github.com/elastic/logstash/pull/7428#issuecomment-308018679 where I discuss a 'stealing' queue design that does full serialization/deserialization for a modest penalty that still yields a large improvement.

Reading through @original-brownbear excellent research in issues like: https://github.com/elastic/logstash/issues/7317 inspired me to think of what the minimal version of sequential persistence would look like. This isn't a real patch, just an experiment, but I think its telling that its performance is dramatically faster than the current persistent queue.  See the benchmarks below. Note that the `memory` queue is now persistent, I patched the class to actually safely store data to disk. The benchmarks show that the new queue uses 26% less Userspace CPU time than the current persistent queue and 43% less system time for a combined improvement of 28%. Initial runs showed an even more pronounced difference, but now I can't repro it.

```
~/p/logstash (sequential-queue) $ head  config/logstash.yml
queue.type: memory

#queue.type: persisted
#queue.checkpoint.writes: 1024

~/p/logstash (sequential-queue) $
time bin/logstash -b 512 -w 2 -e 'input { generator { count => 1000000 } } output { stdout {} }' | pv -l | wc -l
       47.66 real       107.01 user         8.92 sys       <=>           ]
   1M 0:00:47 [  21k/s] [                                   <=>          ]
 1000007
~/p/logstash (sequential-queue) $ git status^C
~/p/logstash (sequential-queue) $ cat config/log^C
~/p/logstash (sequential-queue) $ head  config/logstash.yml
#queue.type: memory

queue.type: persisted
queue.checkpoint.writes: 1024

time bin/logstash -b 512 -w 1 -e 'input { generator { count => 1000000 } } output { stdout {} }'  | pv -l | wc -l
       67.20 real       144.56 user        14.47 sys                      <=>          ]
   1M 0:01:07 [14.9k/s] [                                                <=>           ]
 1000007
```

One key difference in this minimal implementation is that we don't allow for queues larger than BATCH_SIZE * NUM_WORKERS. It's pretty clear though that with extra work we could add that facility, along with other things.

This implementation looks to me to have the same consistency guarantees as the current queue when operating with checkpoint_writes > 1. To implement the same behavior as checkpoint_writes=1 would require a different code path, but not one with more complexity.

Obviously this code does not include any recovery mechanisms, but that's trivial to write given the approach. Just slurp up all files and replay them.

At any rate, I hope this code proves useful to the ongoing discussion, and lets us think about how we build future components in this area.

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -66,6 +66,14 @@ module LogStash; module Util
         @inflight_clocks = {}
         @batch_size = batch_size
         @wait_for = wait_for
+
+        # This should really be the number of workers, but I just set it here to a high number
+        # out of laziness
+        num_files = 20 
+        @file_pool = java.util.concurrent.ArrayBlockingQueue.new(num_files)
+        num_files.times do |t|
+          @file_pool.put(::File.open("/tmp/lsq/#{t}.batch", "ab"))
+        end
       end
 
       def close
@@ -113,7 +121,8 @@ module LogStash; module Util
       # create a new empty batch
       # @return [ReadBatch] a new empty read batch
       def new_batch
-        ReadBatch.new(@queue, @batch_size, @wait_for)
+        file = @file_pool.take()
+        ReadBatch.new(@queue, @batch_size, @wait_for, file)
       end
 
       def read_batch
@@ -144,6 +153,10 @@ module LogStash; module Util
       end
 
       def close_batch(batch)
+        file = batch.file
+        file.truncate(0)
+        @file_pool.put(file)
+
         @mutex.lock
         begin
           # there seems to be concurrency issues with metrics, keep it in the mutex
@@ -185,10 +198,13 @@ module LogStash; module Util
     end
 
     class ReadBatch
-      def initialize(queue, size, wait)
+      attr_reader :file
+
+      def initialize(queue, size, wait, file)
         @queue = queue
         @size = size
         @wait = wait
+        @file = file
 
         @originals = Hash.new
 
@@ -201,13 +217,20 @@ module LogStash; module Util
         @acked_batch = nil
       end
 
+      NEWLINE = "\n"
       def read_next
+
         @size.times do |t|
           event = @queue.poll(@wait)
           return if event.nil? # queue poll timed out
 
+          @file << event.to_json
+          @file << NEWLINE
+
           @originals[event] = true
         end
+
+        @file.fsync
       end
 
       def merge(event)


andrewvc:
I'll just add that maybe there's a corner case here I'm missing in terms of durability, please let me know if that's the case.

andrewvc:
Thinking through this more, I realized this isn't a great comparison because the code in this PR only serializes, it never deserializes.

I've made a slightly more complex 'item stealing' queue (for lack of a better term, its not really like FJP stealing) in https://github.com/elastic/logstash/pull/7430 that serializes and deserializes, and has the benefit therein of supporting arbitrary queue sizes. It takes a small hit, as seen in the benchmark below:

```
~/p/logstash (sequential-double-serializing-queue) $
time bin/logstash -b 512 -w 1 -e 'input { generator { count => 1000000 } } output { stdout {} }'  | pv -l | wc -l
       49.31 real       114.14 user         9.09 sys                    <=>            ]
   1M 0:00:49 [20.3k/s] [                                                <=>           ]
 1000007
```

I should also note that on a subsequent run the existing PQ ran a lot better than before. I'm not sure what changed on my laptop, but its been putting out better stats. I've updated the description above to reflect these numbers.

andrewvc:
Interestingly in some cases the wallclock time is very different from the CPU time as in the below run where we see a nearly 2x improvement on the wallclock but only a 28% reduction in userspace CPU. Partly this is explained >2x improvement in system time.

```
# Stealing queue
time bin/logstash -b 1024 -w 2 -e 'input { generator { count => 500000 threads => 4 } } output { stdout { codec => dots } }'  | pv  | wc -c
       51.75 real       219.59 user        19.34 sys                       <=>         ]
1.91MiB 0:00:51 [37.8KiB/s] [                                               <=>        ]
 2001017

vs.

# persistent queue
time bin/logstash -b 1024 -w 2 -e 'input { generator { count => 500000 threads => 4 } } output { stdout { codec => dots } }'  | pv  | wc -cERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console.

       98.31 real       288.21 user        46.19 sys                                   ]
1.91MiB 0:01:38 [19.9KiB/s] [            <=>                                           ]
 2001017
```

colinsurprenant:
Thant is interesting but I feel we are going full circle here if we look at the history of this feature.

>  think its telling that its performance is dramatically faster than the current persistent queue.

Note that proper Queue batch write is not yet implemented in current PQ (but WIP) which should provide a better comparison point when it lands. Also note that performance in an empty pipeline is not necessarily very interesting in itself. We know performance can be improved and it will in future iterations. We made the conscious decision to first make it right then optimize. 

> One key difference in this minimal implementation is that we don't allow for queues larger than BATCH_SIZE * NUM_WORKERS. It's pretty clear though that with extra work we could add that facility, along with other things.

I think that what this implementation shows is that **if you let go of all original requirements, you can implement a very simple solution**.

> This implementation looks to me to have the same consistency guarantees as the current queue when operating with checkpoint_writes > 1

Not quite. You are at risk of loosing up to a full batch worth of data in case of a process crash. This is not the case with current PQ design even with `checkpoint_writes` > 1.

> At any rate, I hope this code proves useful to the ongoing discussion, and lets us think about how we build future components in this area.

Indeed.

andrewvc:
@colinsurprenant good points!

I've pushed a new commit to https://github.com/elastic/logstash/pull/7430 (closing this issue and moving this party to that thread, since that's definitely the better impl) that flushes after every write. We take a perf hit doing that as expected, but its still faster.

As far as batch write improving perf, this impl. doesn't use the batch write either, so its apples-apples.

I think I've met all the original requirements here, we're at parity.

I'll update #7430 with new benchmarks on this.

PR #7434, opened 2017-06-13
Centralize the JRuby version

ph:
We have multiple places to declare the jruby version, this commit
centralize everything into the `versions.yml` file. This is now used by
the ruby part and the gradle build file to correctly fetch the jruby
library.


original-brownbear:
@ph nice! LGTM :)

Issue #1645, opened 2014-08-18
Define quote escaping semantics.


jordansissel:
This is requested numerous places, but I was reminded from https://github.com/elasticsearch/logstash/pull/895

Escaping quotes and control characters are impossible to do today in Logstash config. "\n" is literally backslash and lowercase n. "\"" is literally backslash and doublequote.

Logstash supports single and double-quoted text values, and we should support escapes in both the same exact way:
## Proposal

Add a new setting, `config.strings.support_escapes` that, when set, will enable the following behaviors on quoted strings in the Logstash config:

- quote-backslash-quote-quote should be text with a single quote as value
- doublequote-backslash-doublequote-doublequote should be text with a single double-quote as value
- Because it is expected by users, "\t" and '\t' should be a single tab character as value (ASCII 09)
- Because it is expected by users, "\n" and '\n' should be a single line feed character as value (ASCII 10)
- Because of \t and \n, we should also support "\r" and '\r' as a single carriage return (ASCII 13)
- "\\\\" and '\\\\' should be text with a single backslash (written as two backslashes in the config)
- Unicode text is fully supported and does not need escaping

## Workarounds today:
- double-quote: Use single-quoted strings: `setting => 'I said, "Hello!"'`
- single-quote: Use double-quoted strings: `setting => "I said, 'Hello!'"`
- backslash: Use a single backslash: `setting => "C:\some\path"`
- tab: use an actual tab character. In some terminal-based editors, you can insert an actual tab character by pressing Control+V then pressing Tab afterwards.
- newline: use an actual newline character: 

```
setting => "this
is
multiple
lines"
```

splashx:
is this issue related to [this issue](https://github.com/logstash-plugins/logstash-filter-mutate/issues/40)?

jordansissel:
@splashx yep :)

splashx:
@jordansissel oh no :laughing:

acchen97:
+1 as I just hit this today too

bcubk:
Hi @jordansissel, are there any progress about this issue? Is there a timetable when it's fixed?

nellicus:
+1 hit today (same as https://github.com/elastic/logstash/issues/3239 6 months ago)

jordansissel:
[...]

Issue #7291, opened 2017-06-02
[Meta] LIR Execution


andrewvc:
This issue is a meta issue to track LIR execution progress.

## Goals

(In no particular order)

* Performance
* Removing duplicate grammars
* Javaifying more of the pipeline
* Modularized execution
* Combine filters/outputs in the pipeline code into one 'processor' notion

### Now

NOTE: This strategy is a little out of date, some of it is in flux. @original-brownbear has an evolving initial approach here: https://github.com/elastic/logstash/pull/7950 (WIP). This checklist may be updated as that progresses.

- [ ] Implement LIR Execution Using Janino #7358
- [ ] Make sure quote escaping from #7442 is ported to lscl.rb
- [ ] Wrap LIR into an execution class #7361
- [ ] Implement Expression execution using Janino to compile to java bytecode #7358
- [ ] Define Pipeline State machine #7364
- [ ] Define the Read client interaction with the processor stage of the pipeline
- [ ] Define the Write client interface with the input stage #7363
- [ ] Pipeline Shutdown Reporter #7365

ph:
Some of my original thoughts are in https://github.com/elastic/logstash/issues/6301

original-brownbear:
@ph reading :)

ph:
I've updated the goals with our discussion from yesterday

ph:
I am doing a spike to see the amount of work to rewrite directly the pipeline.rb class to java, this class does so many things but I think we can get a minimal pipeline quickly.

PR #7442, opened 2017-06-13
Add setting to enable string escaping in Logstash pipeline config

jordansissel:
This PR adds a new boolean setting `config.support_escapes` which defaults to false (the historical behavior). When set to true, the following escapes are handled:

* backslash doublequote -> doublequote
* backslash quote -> quote
* backslash n -> newline (ascii 10)
* backslash r -> carriage return (ascii 13)
* backslash backslash -> backslash
* backslash t -> tab (ascii 9)

This will solve #1645.

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -85,6 +85,11 @@
 #
 # config.debug: false
 #
+# When enabled, process escaped characters such as \n and \" in strings in the
+# pipeline configuration files.
+#
+# config.support_escapes: false
+#
 # ------------ Module Settings ---------------
 # Define modules here.  Modules definitions must be defined as an array.
 # The simple way to see this is to prepend each `name` with a `-`, and keep
--- a/logstash-core/lib/logstash/compiler/treetop_monkeypatches.rb
+++ b/logstash-core/lib/logstash/compiler/treetop_monkeypatches.rb
@@ -10,6 +10,7 @@ class Treetop::Runtime::SyntaxNode
     @ast_metadata ||= {}
     @ast_metadata[key] = value
   end
+
   def compile
     return "" if elements.nil?
     return elements.collect(&:compile).reject(&:empty?).join("")
--- a/logstash-core/lib/logstash/environment.rb
+++ b/logstash-core/lib/logstash/environment.rb
@@ -25,6 +25,7 @@ module LogStash
            Setting::Boolean.new("config.test_and_exit", false),
            Setting::Boolean.new("config.reload.automatic", false),
            Setting::Numeric.new("config.reload.interval", 3), # in seconds
+           Setting::Boolean.new("config.support_escapes", false),
            Setting::Boolean.new("metric.collect", true),
             Setting::String.new("pipeline.id", "main"),
            Setting::Boolean.new("pipeline.system", false),
--- a/docs/static/configuration.asciidoc
+++ b/docs/static/configuration.asciidoc
@@ -249,10 +249,24 @@ Example:
 ==== String
 
 A string must be a single character sequence. Note that string values are
-enclosed in quotes, either double or single. Literal quotes in the string
-need to be escaped with a backslash if they are of the same kind as the string
-delimiter, i.e. single quotes within a single-quoted string need to be escaped
-as well as double quotes within a double-quoted string.
+enclosed in quotes, either double or single. 
+
+===== Escape Sequences
+
+By default, escape sequences are not enabled. If you wish to use escape
+sequences in quoted strings, you will need to set
+`config.support_escapes: true` in your `logstash.yml`. When `true`, quoted
+strings (double and single) will have this transformation:
+
+|===========================
+| Text | Result
+| \r   | carriage return (ASCII 13)
+| \n   | new line (ASCII 10)
+| \t   | tab (ASCII 9)
+| \\   | backslash (ASCII 92)
+| \"   | double quote (ASCII 34)
+| \'   | single quote (ASCII 39)
+|===========================
 
 Example:
 
--- a/docs/static/releasenotes.asciidoc
+++ b/docs/static/releasenotes.asciidoc
@@ -12,6 +12,10 @@ This section summarizes the changes in the following releases:
 
 Placeholder for alpha3 release notes
 
+* Added new `logstash.yml` setting: `config.support_escapes`. When
+  enabled, Logstash will interpret escape sequences in strings in the pipeline
+  configuration.
+
 
 [[logstash-6-0-0-alpha2]]
 === Logstash 6.0.0-alpha2 Release Notes
--- a/docs/static/settings-file.asciidoc
+++ b/docs/static/settings-file.asciidoc
@@ -138,6 +138,10 @@ The `logstash.yml` file includes the following settings:
   in plaintext passwords appearing in your logs!
 | `false`
 
+| `config.support_escapes`
+| When set to `true`, quoted strings will process the following escape sequences: `\n` becomes a literal newline (ASCII 10). `\r` becomes a literal carriage return (ASCII 13). `\t` becomes a literal tab (ASCII 9). `\\` becomes a literal backslash `\`. `\"` becomes a literal double quotation mark. `\'` becomes a literal quotation mark.
+| `false`
+
 | `modules`
 | When configured, `modules` must be in the nested YAML structure described above this table.
 | None
--- a/logstash-core/lib/logstash/compiler.rb
+++ b/logstash-core/lib/logstash/compiler.rb
@@ -7,9 +7,9 @@ java_import org.logstash.config.ir.graph.Graph
 module LogStash; class Compiler
   include ::LogStash::Util::Loggable
 
-  def self.compile_sources(*sources_with_metadata)
+  def self.compile_sources(sources_with_metadata, settings)
     graph_sections = sources_with_metadata.map do |swm|
-      self.compile_graph(swm)
+      self.compile_graph(swm, settings)
     end
 
     input_graph = Graph.combine(*graph_sections.map {|s| s[:input] }).graph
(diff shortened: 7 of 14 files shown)


config/logstash.yml
```diff
@@ -85,6 +85,11 @@
 #
 # config.debug: false
 #
+# When enabled, process escaped characters such as \n and \" in strings in the
+# pipeline configuration files.
+#
+# config.strings.support_escapes: false
```
suyograo: Why not just `config.support_escapes: false` This applies to both `config.strings` i.e. `-e` as well as `-f` right?

suyograo:
woot!

config/logstash.yml
jordansissel: `strings` is in reference to string syntax in the grammar, not the "config string" (aka `-e`).

This is perhaps a bad name.

I wanted to imply that this was only for String grammar (https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/config/grammar.treetop#L75-L85) and nowhere else.

I'm open to calling it `config.support_escapes`

jordansissel:
Added specs:

```
LogStashConfigParser
  #compile
    when config.strings.support_escapes
      is enabled
        should process escape sequences
      is default
        should not process escape sequences
      is false
        should not process escape sequences

Finished in 0.26366 seconds (files took 2 seconds to load)
3 examples, 0 failures
```

config/logstash.yml
jordansissel: Renamed to `config.support_escapes`

jordansissel:
I also need to copy the AST change into lscl (https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/compiler/lscl.rb#L177-L181).

jordansissel:
I'll file an issue to get this change added to the LIR/LSCL treetop stuff and do it as a separate PR.

acchen97:
@jordansissel you get a big heart from me for this one. :)

Just a minor note, do we want to roll this in 5.6, or wait for 6.0? If we introduce this in 6.0, we can directly add this by default as a breaking change and note it as a step in the major upgrade docs. It's been broken for a few years now, I'm not sure a few more months would inflict additional pain.

jordansissel:
The one test that failed looks like it stalled during an Elasticsearch output test:

https://travis-ci.org/elastic/logstash/builds/242544047?utm_source=github_status&utm_medium=notification

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

jordansissel:
@acchen97 From a version policy aspect, there's no reason it can't appear in 5.6 (default disabled) and 6.0 (default enabled, maybe, we can have that discussion). I think this setting being a boolean in logstash.yml gives us this flexibility to introduce it  during a minor (feature) release.

So, that's a long-winded way to say -- we can introduce this in a minor if desired because it's feature-flagged.

jordansissel:
@acchen97 I talked with @suyograo, and this should be good to land in 5.6 and 6.0.

After merging, I'll open a new issue to discuss enabling by default in 6.0.

andrewvc:
Is there a reason we wouldn't make the changes to lscl.rb in the same PR? I'm worried about diverging grammar here.

jordansissel:
@andrewvc in truth, I looked into this and my perception of the effort was that it would be much larger refactor and I don't feel I have the energy for it at this moment.

For this PR, I have chosen to pass the LogStash::Settings object to the compile() method. Maybe this wasn't the right choice, but the net impact, based on what I see, for lscl, is this:

`LogStashCompilerLSCLGrammar::Logstash::Compiler::LSCL::AST::String#expr` method needs to know if string escapes are enabled. If we do it the same way as this PR, this means the `expr` method needs to be given that information (currently, given the `LogStash::Settings` instance).

It is also that, to me, the LSCL grammar change can be done separate from this PR in the idea that I want to make smaller, not larger, PRs. The tests are passing, which means that we are not testing for parity between LSCL grammar/ast and the current grammar/ast.

andrewvc:
@jordansissel that's fine. We just need to make sure @ph and @original-brownbear are aware of this when they start implementing LIR execution. That update will be a blocker for #7291 . I've added it as a checkbox in that meta-issue.

jordansissel:
>  I'm worried about diverging grammar here.

I'm trying to balance the too-large-pr problem (adding LIR/LSCL refactor here) while trying to avoid divergence, so my thought was to recognize that more work is necessary and filed a bug to track it.

We could do this work in a feature branch, perhaps. I'm OK delaying this feature until LSCL support is added, if that is what you are asking for.

jsvd:
What do you think of making `support_escapes` a property of the `LogStash::Config::AST` that is set at startup time after reading the yaml file, or during [Config.compile](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/config/config_ast.rb#L39-L43) ?

This way if we set the property during the first step of the compilation, like so:

```ruby
  class Config < Node
    def compile(settings)
      LogStash::Config::AST.deferred_conditionals = []
      LogStash::Config::AST.deferred_conditionals_index = 0
      LogStash::Config::AST.plugin_instance_index = 0
      LogStash::Config::AST.support_escapes = settings.get_value("config.support_escapes")
```

then we can avoid injecting the settings in all methods, as only the string node needs to grab its value:

```ruby
  class String < Value
    def compile(settings)
      if LogStash::Config::AST.support_escapes
        Unicode.wrap(LogStash::Config::StringEscape.process_escapes(text_value[1...-1]))
      else
        Unicode.wrap(text_value[1...-1])
      end
    end
  end
```

We already use LogStash::Config::AST as a hopping point for global state during compilation so I don't see a problem using the sigleton here too. This would drastically reduce the changes in this PR. What do you think?

jsvd:
I tested this locally, works wonderfully. I'm a bit concerned that people will abuse this, so I think all our documentation entries on this feature should have a complete list of things we escape. wdyt?

jordansissel:
> suggest that all our documentation entries on this feature should have a complete list of things we escape

+1

jordansissel:
Ahh, I didn't notice we used global state (something I was trying to avoid) already. I'll look into it -- I'd also like this PR to be smaller ;)

jordansissel:
Added a table in the docs with transform so now both references to string escapes are having a full list of what escapes are supported.

`bin/rspec logstash-core/spec` is passing for me

@jsvd  back to you for more review, I think.

jordansissel:
Oh, and I also  did the proposed refactor which makes this PR smaller by making the "should we process escape sequences?" checked as a property of the root AST node (Config).

Also has tests for both current grammar (config_ast) and future grammar (lscl)

jordansissel:
Partial test failures seem unrelated (time math, agent stop) to this PR.

logstash-core/lib/logstash/environment.rb
```diff
@@ -15,7 +15,7 @@ module Environment
     end
   end
 
-  [
+  LOGSTASH_SETTINGS = [
```
jsvd: I may be missing something but I don't see the need for this refactor

The `register_settings` method is only used in tests and right here, and even in the tests the support_escapes setting isn't done through a Settings object, but directly with the setter https://github.com/elastic/logstash/pull/7442/files#diff-1c97814b03bd1f3c0f796071ebc0cf19R169

Maybe this refactor isn't necessary?

jordansissel: I may still use it in the specs, but I agree there are setters now for both config_ast and lscl. I'll try to revert the settings things.

jordansissel: Ahha. mock_settings does what I need.

jordansissel:
@jsvd Ok I've removed my `register_settings` refactoring. I added it for a purpose that is already served (yay!) by another test feature (mock_settings).

jordansissel:
Finished in 5 minutes 53 seconds (files took 10.71 seconds to load)
1966 examples, 0 failures, 6 pending

Tests passing locally. I'll wait on travis and jenkins. @jsvd can you take another look?

logstash-core/lib/logstash/compiler/treetop_monkeypatches.rb
```diff
@@ -10,9 +10,10 @@ def set_meta(key, value)
     @ast_metadata ||= {}
     @ast_metadata[key] = value
   end
+
   def compile
     return "" if elements.nil?
-    return elements.collect(&:compile).reject(&:empty?).join("")
+    return elements.collect { |x| x.compile }.reject(&:empty?).join("")
```
jsvd: I see no need to keep this change and the white space above

jordansissel: Ahh, yes. This was part of a refactor that is no longer needed.

jordansissel:
(Two test instances failed, but these failures were unrelated to this PR)

acchen97:
WOOT! Excellent work! :)

andrewvc:
Very glad to see that land!

jordansissel:
5.x backport PR for this feature is here: https://github.com/elastic/logstash/pull/7516

binary132:
Unicode escape sequences might be nice too (e.g. `\u00ff`).

jordansissel:
@binary132 Logstash pipeline configuration files are already expected to be UTF-8, so no escape sequences should be necessary since you can have these literally in your config:

```
input {
  file {
    path => "/var/log/*.log"
    add_tag => "☃︎"
  }
}
```

Hermain:
I'm using the docker gelf driver to send logs to logstash over tcp. The logs are NUL byte delimited (\0 or \u0000) unfortunately that still doesn't work. Its common in gelf to delimit messages with \0 and implementing this would remove the need for a gelf tcp plugin.
Is there a reason why \0 shouldn't be added to the supported list of escaped delimiters?

jordansissel:
hermain, I recommend you use the gelf input.

Hermain:
But gelf input is udp only. If a small change (some way to use /0 as delimiter in a codec) is enough to make a plugin obsolete I would prefer that.

Issue #6991, opened 2017-04-27
LS Monitoring UI docs page


acchen97:
Our Logstash docs don't currently link to or mention the monitoring UI. My proposal would be to have a top level Monitoring section that encompasses a new Monitoring UI page (which can link off to the relevant x-pack docs) and the monitoring API endpoints and response payloads.

/cc @dedemorton

dedemorton:
I'll add a topic level called "Monitoring Logstash" and nest the API info under that. Does that sound OK?

acchen97:
Sounds good, and the monitoring UI benefits/brief will just be in that top level page then, right? In this particular context, I'd be +1 for including a snapshot of the UI.

dedemorton:
Yup, that's right. We'll need a container for all the X-Pack monitoring docs that will be making their way into the LS docs at some point in the near future (maybe 5.5).

dedemorton:
Closed via #7443

PR #7443, opened 2017-06-13
Add info about monitoring UI

dedemorton:
Closes https://github.com/elastic/logstash/issues/6991

Note that the TOC now looks like this:

image

I'd actually prefer to have all the APIs nested under a container called "Monitoring APIs" but adding another level of nesting would require a fair amount of book restructuring because of the way asciidoc/docbook manages "chunk" levels.


suyograo:
Nice. LGTM

PR #7446, opened 2017-06-13
Remove deprecate call to jruby 1.9

ph:
With the move to JRuby9k which use the 2.3 syntax, this make the
 following constant deprecated

 ```
 config_19.setCompatVersion(CompatVersion.RUBY1_9);
 ```


ph:
I've restarted the test, the travis container was not able to reach the apt repository to bootstrap the test environment...

PR #7451, opened 2017-06-14
MODULES: Add maxbuckets to the kibana config json

### LINKED ISSUE #7450: MODULES: "Metrics: Max buckets exceeded" - increase the number of buckets
For the metrics visualisation when computing the events per second; 

Changed it to 20000; but if we’re thinking more than 30 days, the buckets will need to be adjust to 30 days x 24 hours x 60 minutes x 60 seconds = 2592000.

guyboertje:
Fixes #7450

### DIFF
--- a/logstash-core/lib/logstash/modules/kibana_config.rb
+++ b/logstash-core/lib/logstash/modules/kibana_config.rb
@@ -19,6 +19,7 @@ module LogStash module Modules class KibanaConfig
     @name = modul.module_name
     @settings = settings
     @index_name = settings.fetch("dashboards.kibana_index", ".kibana")
+    @metrics_max_buckets = settings.fetch("dashboards.metrics_max_buckets", 60000)
   end
 
   def dashboards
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/modules/kibana_config.rb
```diff
@@ -19,6 +19,7 @@ def initialize(modul, settings)
     @name = modul.module_name
     @settings = settings
     @index_name = settings.fetch("dashboards.kibana_index", ".kibana")
+    @metrics_max_buckets = settings.fetch("dashboards.metrics_max_buckets", 60000)
```
suyograo: Let's make it 50,000

guyboertje: We should use the formula that Nic proposed as its events per second - 24 hours is 86400 seconds = 86400 buckets

suyograo: sounds good. can you please add a comment explaining the magic number? Also, please make it a constant.

guyboertje: Although I have made it so the user can alter the value in the modules section of LS settings, we need to clarify with SA team whether this is a module specific KV pair.

PR #7456, opened 2017-06-15
Enhance ingest convert to add input/output

suyograo:
Since ingest's implicit output is Elasticsearch, we should mirror that behavior by creating an ES output when converting a pipeline.

Also, added a flag to append `stdin` and `stdout` if a user chooses to. This makes manual testing easier.

### DIFF
--- a/tools/ingest-converter/src/main/resources/ingest-shared.js
+++ b/tools/ingest-converter/src/main/resources/ingest-shared.js
@@ -167,19 +167,11 @@ var IngestConverter = {
             "}");
     },
 
-    get_newline: function () {
-        return "\n";
-    },
-
     append_io_plugins: function(filters_pipeline, append_stdio) {
         if (append_stdio === true) {
-            return IngestConverter.get_stdin_input() +
-                IngestConverter.get_newline() +
-                filters_pipeline +
-                IngestConverter.get_newline() +
-                IngestConverter.get_stdout_output();
+            return [IngestConverter.get_stdin_input(), filters_pipeline, IngestConverter.get_stdout_output()].join("\n");
         } else {
-            return filters_pipeline + IngestConverter.get_newline() + IngestConverter.get_elasticsearch_output();
+            return [filters_pipeline, IngestConverter.get_elasticsearch_output()].join("\n");
         }
     }
 };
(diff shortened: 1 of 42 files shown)


tools/ingest-converter/src/main/resources/ingest-shared.js
```diff
@@ -142,5 +142,44 @@ var IngestConverter = {
...
+            "}");
+    },
+
+    get_newline: function () {
+        return "\n";
+    },
+
+    append_io_plugins: function(filters_pipeline, append_stdio) {
+        if (append_stdio === true) {
+            return IngestConverter.get_stdin_input() +
```
original-brownbear: @suyograo this is kinda weird, right? :D

Doesn't really matter but why not save the complication of the `get_newline` and just do

```js
return [IngestConverter.get_stdin_input(), filters_pipeline, IngestConverter.get_stdout_output()].join("\n");
```

? (same for the below use of this, could even dry up the `join("\n")` and just set up the array in the condition) :)

tools/ingest-converter/src/test/resources/org/logstash/ingest/logstashDotsInConvertField.conf
```diff
@@ -5,3 +5,8 @@ filter {
       }
    }
 }
+output {
+   elasticsearch {
+      hosts => "localhost"
```
original-brownbear: @suyograo isn't "localhost" the default here anyway?

original-brownbear:
@suyograo LGTM, one comment on style and one random comment about the "localhost" default, both are pretty optional :)

tools/ingest-converter/src/test/resources/org/logstash/ingest/logstashDotsInConvertField.conf
suyograo: @original-brownbear yep, it is, but I wanted to be explicit and add this so it can be used as a template for users. So if someone has ES on a different host, they can easily change the value with this.

tools/ingest-converter/src/main/resources/ingest-shared.js
suyograo: :man_facepalming: yep

tools/ingest-converter/src/test/resources/org/logstash/ingest/logstashDotsInConvertField.conf
original-brownbear: @suyograo makes sense :)

suyograo:
updated.

original-brownbear:
@suyograo LGTM :)

PR #7457, opened 2017-06-15
DeadLetterQueueReader should handle missing segment files at startup

### LINKED ISSUE #7433: DeadLetterQueueReader should handle missing segment files
In preparation for #6990, the dead letter queue reader implementation should be able to recover when a client attempts to resume from a deleted segment file.

robbavey:
Change DeadLetterQueueReader, so that if a missing segment file is encountered at startup, the next valid entry will be used instead
    
Fixes #7433


suyograo:
@jordansissel assigning this to you :)

jordansissel:
Sorry for the delays.

LGTM

Issue #7460, opened 2017-06-15
LogStash::Runner Specs are Failing!


original-brownbear:
See here:

https://travis-ci.org/elastic/logstash/jobs/243074345#L5541

```sh
rspec ./logstash-core/spec/logstash/runner_spec.rb:350 # LogStash::Runner logstash modules --modules with an available module specified but no connection to elasticsearch should log fatally and return a bad exit code
rspec ./logstash-core/spec/logstash/runner_spec.rb:380 # LogStash::Runner logstash modules --modules with an available module specified and a mocked connection to elasticsearch should not terminate logstash
rspec ./logstash-core/spec/logstash/runner_spec.rb:317 # LogStash::Runner logstash modules --config.test_and_exit with a good configuration should exit successfully
```

I can reproduce this in 100% of runs locally as well, relevant output there:

```sh
ERROR: Configuration reloading also requires passing a configuration path with '-f yourlogstash.conf'
WARNING: Could not find logstash.yml which is typically located in $LS_HOME/config or /etc/logstash. You can specify the path using --path.settings. Continuing using the defaults
WARNING: Could not find logstash.yml which is typically located in $LS_HOME/config or /etc/logstash. You can specify the path using --path.settings. Continuing using the defaults
2017-06-15 11:51:17,271 LogStash::Runner ERROR No log4j2 configuration file found. Using default configuration: logging only errors to the console.
Configuration OK
2017-06-15 11:51:22,508 LogStash::Runner ERROR No log4j2 configuration file found. Using default configuration: logging only errors to the console.
ERROR: option '-w': Setting "pipeline.workers" must be a Integer. Received:  (NilClass)

See: 'bin/logstash --help'
2017-06-15 11:51:29,042 LogStash::Runner ERROR No log4j2 configuration file found. Using default configuration: logging only errors to the console.

RSpec::Mocks::MockExpectationError: (Double "logger").fatal("The given configuration is invalid. Reason: Could not load the configuration file")
    expected: 0 times with any arguments
    received: 1 time with arguments: ("The given configuration is invalid. Reason: Could not load the configuration file")
./logstash-core/lib/logstash/runner.rb:316:in `execute'
./vendor/bundle/jruby/2.3.0/gems/clamp-0.6.5/lib/clamp/command.rb:67:in `run'
./logstash-core/lib/logstash/runner.rb:229:in `run'
./logstash-core/spec/logstash/runner_spec.rb:319:in `block in (root)'
[...]

PR #7462, opened 2017-06-15
#7460 ignore failing tests caused by #e939c6a removing CEF module

original-brownbear:
For #7460 see #e939c6a that caused this by removing the cef module. Ignored those tests until CEF comes back.

### DIFF
--- a/logstash-core/spec/logstash/runner_spec.rb
+++ b/logstash-core/spec/logstash/runner_spec.rb
@@ -315,6 +315,7 @@ describe LogStash::Runner do
       context "with a good configuration" do
         let(:module_string) { "cef" }
         it "should exit successfully" do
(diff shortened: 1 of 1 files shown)


original-brownbear:
@robbavey added a skip that includes output, but not issue. I just linked the PR that explains this. They didn't add an issue there so I didn't wanna "overrule" that :D Seems we're just not in on that :D

ph:
@original-brownbear Can you backport this to 5.x and 5.5? Both of theses branch fails on travis.

original-brownbear:
@ph sure, hah I checked when I did this and didn't see the cef folder in those branches' histories and concluded it's cool there :D Will do in about an hour.

ph:
@original-brownbear Thanks I was just looking at the failed emails :)

PR #7464, opened 2017-06-15
Avoid unnecessary seeks in seekToNextEventPosition

robbavey:
When an exact match is found during the binary search for the
next event, use the matching block instead of seeking from the
'low' block.

Add toString method to DLQEntry


jordansissel:
Test failures seem to be due to recent removal of modules but specs failing should be fixed by https://github.com/elastic/logstash/pull/7462 (unrelated to this issue)

jordansissel:
Code LGTM

PR #7467, opened 2017-06-15
Properly hash when no original source is present.

andrewvc:
This corrects a mistake where the digest of the string source should be used when present, and the hash of the graph should be used when not.


ph:
@andrewvc  https://github.com/elastic/logstash/pull/7467 failure seems to be an unreacheable es, are we using docker for that IT?

PR #7468, opened 2017-06-15
Detect 'extra' sibling gems and auto-install them for convenience

andrewvc:
This patch will discover extra git repos for gems in the '../logstash-extra' folder, and auto install them in the gemfile if present. Useful for developing extensions to logstash like x-pack.

### DIFF
--- a/Gemfile.template
+++ b/Gemfile.template
@@ -2,6 +2,21 @@
 # If you modify this file manually all comments and formatting will be lost.
 
 source "https://rubygems.org"
+
+# Autoload extension gems in ../logstash-extra/GEMNAME for 
+# convenience during development. Do not enable in production
+# as this may allow loading of code from unexpected extra directories
+if ENV["AUTOLOAD_LOGSTASH_EXTRA"] == "true" && Dir.exist?("../logstash-extra")
+  Dir.open("../logstash-extra").each do |entry|
+    next if entry.start_with?('.')
+    path = "../logstash-extra/#{entry}"
+    next unless File.directory?(path)
+    name = entry =~ /^(.+)-logstash$/ ? $1 : entry 
+    puts "Loading development gem: '#{name}' from '#{path}'"
+    gem name, :path => path
+  end
+end
+
 gem "logstash-core", :path => "./logstash-core"
 gem "logstash-core-plugin-api", :path => "./logstash-core-plugin-api"
 gem "paquet", "~> 0.2.0"


Gemfile.template
```diff
@@ -2,6 +2,18 @@
...
+  Dir.open("../logstash-extra").each do |entry|
+    next if entry.start_with?('.')
+    path = "../logstash-extra/#{entry}"
+    next unless File.directory?(path)
+    name = entry =~ /^(.+)-logstash$/ ? $1 : entry 
+    puts "Loading development gem: '#{name}' from '#{path}'"
+    gem name, :path => path
+  end
+end
+
```
ph: I like the idea, but one thing that make me nervous, Installing x-pack is not side effect free and _will_ change some internals, like adding a new monitoring pipeline auto-magically, that could make running the unit test locally fails if the state is different that what we expect in the suite.

andrewvc: hmmm, I will say that I haven't personally experienced that. If that is the case, I think we should consider that a bug in xpack.

The thing is, for everyone working on xpack they almost always have that line in their Gemfile, so its problematic if that is the case.

ph: I don't think it would be a bug in x-pack but in core, I think we only have the risk to run into that when we actually test the runner because it will load code and execute events on the x-pack. 

Might be a matter of only mock these interaction when running the test sutie?

Let me do a few local test and I will report back, if its that OK with you?

andrewvc:
Sounds good. I think I should require users to add am explicit environment
var to enable this. I have some security concerns without it being opt in

ph:
> Sounds good. I think I should require users to add am explicit environment
var to enable this. I have some security concerns without it being opt in

If you add this I will LGTM right away, maybe something similar to 
https://github.com/logstash-plugins/logstash-input-beats/blob/master/Gemfile#L6

andrewvc:
@ph I added the env var, along with a nice comment so that if users find this they don't try to enable it.

ph:
@andrewvc I am ok with that changes, but it has some limitations with the build, the way the plugin manager work since 1.5.x it will read the file (Gemfile) and evaluate the content and remove anything that is not related to defining gems, so all your _if_ will be removed. But if the `ifs` are matched when the code is evaluated the x-pack gem will be added **permanently** to the gemfile.

And if the Gemfile already exist we will **never** read/interpret the Gemfile.template again.

I just wanted to describe the limitation to make sure you are aware of them.

If its the case LGTM.

andrewvc:
@ph I think the complexity here is just too high, I'm going to kill this PR.

PR #7469, opened 2017-06-15
Set default entropy source to /dev/urandom

### LINKED ISSUE #6117: Add troubleshooting documentation for very slow startup times due to lack of entropy
The lack of entropy in /dev/random can make logstash (or even any other java/jruby software?) startup time be higher than 5 minutes.

Example issue here where starting logstash took 10 minutes: https://github.com/elastic/logstash/issues/6114

JRuby's wiki has a section on this issue https://github.com/jruby/jruby/wiki/Improving-startup-time#ensure-your-system-has-adequate-entropy

We should somehow address this in a troubleshooting section, either:

1) making the user aware that the problem exists and other software could be causing this by draining /dev/random
2) suggest software that generates entropy for /dev/random, therefore working around the draining issue

Usually there should be no issue unless /dev/random is being read from very frequently (for example a ruby filter that reads from it on every event).

jordansissel:
This changes the Java-wide from the default of /dev/random. The reason
for this chnage is that /dev/random can block if there is insufficient
entropy available, and this blocking can cause Logstash to stall on
startup for many many minutes.

Fixes #6117.

This scenario (no entropy) seems especially prone to occur on virtual
machines.

Related: https://github.com/jruby/jruby/wiki/Improving-startup-time#ensure-your-system-has-adequate-entropy

On Windows, I was not sure if this would work, but the JDK source tree
says this:

    On Windows systems, specifying the URLs "file:/dev/random" or
    "file:/dev/urandom" will enable the native Microsoft CryptoAPI seeding
    mechanism for SHA1PRNG.

So defaulting to /dev/urandom should still do the right thing on
Windows.

I also checked to see if other unix-likes had /dev/urandom, and
according to Wikipedia[1], Linux, FreeBSD, OpenBSD, OSX, Solaris,
NetBSD, and others all have urandom.

[1] https://en.wikipedia.org/wiki//dev/random

### DIFF
--- a/config/jvm.options
+++ b/config/jvm.options
@@ -72,3 +72,6 @@
 # log GC status to a file with time stamps
 # ensure the directory exists
 #-Xloggc:${LS_GC_LOG_FILE}
+
+# Entropy source for randomness
+-Djava.security.egd=file:/dev/urandom


andrewvc:
Can we discuss the security tradeoffs here? They are well represented in this [SO post](https://stackoverflow.com/questions/3690273/did-i-understand-dev-urandom). The most salient comment being: 

```
Linux's /dev/urandom is similar except that it will never block, and thus may risk returning low-quality randomness if used just after boot. On the other hand, /dev/random may block even long after boot-time, which is also a problem. I have often seen servers stall mysteriously, because some software was insisting on using /dev/random, and the keyboard-less server was not getting enough entropy.

Usual Linux distribution save at shutdown a random seed obtained from /dev/urandom, and inject it back upon next boot, thus guaranteeing the quality of the random provided by /dev/urandom. Only during OS installation does cryptographic quality becomes an issue, and usually it is not because installation involves a number of interactions with the human being who performs the installation, yielding hordes of entropy.

To sum up, under both Linux and FreeBSD, you should use /dev/urandom, not /dev/random.
```

I just want to make sure we're comfortable with this for our purposes.

andrewvc:
I feel like I don't know what a potential attack vector would be here. @jordansissel is there a possibility that an SSL key exchange between LS and some remote endpoint could be compromised?

My strong suspicion here is that the answer is no, not practically. WDYT?

jordansissel:
We discussed these concerns in https://github.com/elastic/logstash/issues/6117, with favor for /dev/urandom [here](https://github.com/elastic/logstash/issues/6117#issuecomment-270956865) and [here](https://github.com/elastic/logstash/issues/6117#issuecomment-270958116).

The Linux manpage for urandom(4) says:

> As a result, in this case the returned values are theoretically vulnerable to a cryptographic attack on the algorithms used by the driver.  Knowledge of how to do this is not available in the current unclassified literature, but it is theoretically possible that such an attack may exist.

There may be an attack, but none are known. If we're talking theoretical, then it is theoretically possible for an attacker to influence the environmental/external noise used to feed the entropy pool.

My judgement is that given the data we have, urandom is safe to use.

andrewvc:
@jordansissel ah, missed that. Works for me!

LGTM

jordansissel:
@suyograo #6117 is marked for v5.5.0, but I have only merged this into 5.x branch (5.6) and master (6.0). If you want this in 5.5.0, please let me know.

suyograo:
@jordansissel this is a really good fix, which has affected many users. We should backport it to 5.5

jordansissel:
@suyograo Sounds good.

Issue #7439, opened 2017-06-13
Provide more descriptive titles for plugins docs


dedemorton:
Longer titles will clutter up the navigation a bit, but they're important for SEO. Plus, with so many plugins in the navigation now, users lose context when they scroll down. Having descriptive titles will help with that too.

Here's some feedback about this issue from Clint:

I've had complaints about trying to search for "logstash http output plugin".  I've improved the search by including the breadcrumbs, but the logstash plugin docs could be greatly improved by adding more meaningful titles than just `http`  (see https://www.elastic.co/guide/en/logstash/5.4/plugins-outputs-http.html#plugins-outputs-http-http_method), eg a title of "`http` output plugin" would make this much easier to find.

dedemorton:
@ph Can we add this to the script for generating the asciidoc so I don't have to open PRs against every Logstash plugin known to humanity?

ph:
@dedemorton I won't regenerate the docs from the ruby source, but we could probably script something to make the change you want. I am not sure I understand what the change would look like, can you make an example PR on a plugin repo?

ph:
>  against every Logstash plugin known to humanity?

I will make sure you don't loose your sanity, so we will try to script it. :)

dedemorton:
@ph We need the titles to include the type of plugin. Before we make any changes, I'm going to see if @clintongormley  can do something to the build to support a short form of the title in the TOC, while having the long title in the topic itself. We'd probably need to use special tagging or something, so let's hold on creating the script until we find out what's possible. 

Here's what the TOC would look like with the longer titles (harder to scan IMO):

![image](https://user-images.githubusercontent.com/14206422/27148143-08efa052-50f4-11e7-92cd-5e11736bdedd.png)

Here's the original TOC:

![image](https://user-images.githubusercontent.com/14206422/27148190-31374e02-50f4-11e7-96bd-373f48762f22.png)

ph:
@dedemorton roger that!

dedemorton:
As usual, @clintongormley has a solution!

All we have to do is include an abbreviated title that docbook understands. So for the beats input plugin, we would add "input plugin" to the end of the topic title and add the short title within `titleabbrev` tags. For example:

```
=== beats input plugin

++++
<titleabbrev>beats</titleabbrev>
++++
````

@ph So we'd need a script that appends the plugin type to the end of each title (`input plugin`, `output plugin`, etc) and adds the block for setting the abbreviated title.

ph:
[...]

PR #7471, opened 2017-06-15
Add the plugin titleabbrev

ph:
To have better search result we now use the full title in the plugin
main doc, but we also need an abbreviation.

Ref: #7439


dedemorton:
LGTM

PR #7472, opened 2017-06-15
Update versions for 5.5

dedemorton:
All the 5.5 docs are building now, so we should be able to use `{branch}` to resolve the doc paths  (will make it easier to bump versions later on).


suyograo:
LGTM. @dedemorton can you rebase and merge?

Issue #7149, opened 2017-05-18
Use Java 8 Time instead of Joda in LS and any plugins.


guyboertje:
- JRuby 9k now uses Java 8 Time classes but it still bundles Joda at an old version (2.8.2)
- The Java Event leverages the JRuby provided Joda libs.
- The Date Filter does too.
- The version of Joda in JRuby is quite old and does not have the latest tz info data and this cannot be updated using the TZUpdater tool fom Oracle.
 - Times from some timezones e.g. Venezuela will be converted to UTC incorrectly because the Venezuelans changed from -4:30 to -4:00 in 2016.
- Tests that compare a Timestamp with a Ruby Date can fail as the Ruby Time holds 6 digits of fractional seconds but Joda based Timestamp only holds 3 digits of fractional seconds.

suyograo:
@guyboertje Is this going to break the interface for users w.r.t. the [matcher patterns](https://www.elastic.co/guide/en/logstash/current/plugins-filters-date.html#plugins-filters-date-match)? I don't think Joda and Java 8 date formatting of date patterns are compatible, no? I might be wrong though, haven't dug deep.

http://www.joda.org/joda-time/apidocs/org/joda/time/format/DateTimeFormat.html

ph:
@suyograo We might be able to create a translation layer between the 2, making it transparent to the user.

ph:
From a discussion on zoom between @suyograo @jordansissel and myself we agreed to the following:

- We don't think this issue is a blocking for #7125 
- We can do a follow-up PR to migrate to java time.

colinsurprenant:
So would it be possible to keep Joda just for the pattern matching part and use Java 8 time everywhere else?? If it is, this would allow moving forward with proper Java 8 support and just keep Joda for the pattern matching until we decide what to do with that?

ph:
@colinsurprenant Yup, I think possible to do that, we could use it the date filter and in the _StringInterpolation_ classes.

jordansissel:
> possible to keep Joda just for the pattern matching part 

This may not be necessary since java.time (JSR310) is based on Joda and *hopefully* has the same date format stuff.

colinsurprenant:
are we moving forward with this?

jordansissel:
@colinsurprenant yes, but we haven't done any compatibility research, testing, or code yet.

If you'd like to work on this, I am in favor of it being delivered. I think it's two things:

[...]

Issue #5606, opened 2016-07-08
Upgrade to JRuby 9k


jordansissel:
According to [this tweet](https://twitter.com/jruby/status/751150368839589888), JRuby 1.7 will be EOL by the sometime during 2016.

Let's make sure we have all the bugs/problems identified and get on the path to upgrade to JRuby 9k some time after Logstash 5.0.0

purbon:
I can revamp https://github.com/elastic/logstash/pull/3654 in more structured way, including the intention to clarify possible fast/slow areas due to the change to 9k, so we can work on them.

purbon:
put a PR to track the upgrade at https://github.com/elastic/logstash/pull/5616 , but hit an error in our java event, described at #5617 (blocker for the upgrade)

guyboertje:
Found this on the date filter with 9.1.5.0
```
undefined method `time' for "2013-04-21T00:00:00.000Z":String
     # /Users/guy/.gem/jruby/2.3.1/gems/logstash-core-event-java-5.0.0-java/lib/logstash/timestamp.rb:14:in `<=>'
     # uri:classloader:/jruby/kernel/basicobject.rb:4:in `!='
     # /Users/guy/.gem/jruby/2.3.1/gems/insist-1.0.0/lib/insist/comparators19.rb:6:in `!='
     # ./spec/filters/date_spec.rb:339:in `block in (root)'
     # /Users/guy/.gem/jruby/2.3.1/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in (root)'
```
same test passes in 1.7.25

guyboertje:
I have begun the upgrade process. All core specs pass with JRuby 9.1.6.0 at the time of posting this comment.

I will be adding notes here about what needs to change in the plugins as I find them (as well as adding an issue to each plugin)

guyboertje:
`Digest::HMAC` is removed in Ruby 2.2 - its suggested that we use `OpenSSL::HMAC`
plugins affected:
- [ ] `LogStash::Codecs::Collectd` [issue 24](https://github.com/logstash-plugins/logstash-codec-collectd/issues/24)

guyboertje:
Netflow Codec, JRuby 1.7.X quirk.
https://github.com/logstash-plugins/logstash-codec-netflow/issues/66

guyboertje:
[...]

PR #7474, opened 2017-06-16
Fix for: Unreliable logstash-core/spec/logstash/timestamp_spec under JRuby 9000

### LINKED ISSUE #7463: Unreliable logstash-core/spec/logstash/timestamp_spec under JRuby 9000
We will not be able to assert that a timestamp time is exactly equal to the time it was created from.

Got:
```
  1) LogStash::Timestamp constructors should work
     Failure/Error: expect(t.time).to eq(now)

       expected: 2017-06-15 10:34:08.389999999 +0000
            got: 2017-06-15 10:34:08.389000000 +0000

       (compared using ==)

       Diff:

     # ./logstash-core/spec/logstash/timestamp_spec.rb:17:in `block in (root)'
     # ./vendor/bundle/jruby/2.3.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in /Users/guy/elastic/logstash/vendor/bundle/jruby/2.3.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb'
     # ./lib/bootstrap/rspec.rb:13:in `<main>'

Finished in 4 minutes 6.4 seconds (files took 6.6 seconds to load)
2240 examples, 1 failure, 3 pending
```

guyboertje:
Fix Timestamp.new(now) != now, out by 0.999999 milliseconds

Fixes #7463

### DIFF
--- a/logstash-core/spec/logstash/legacy_ruby_timestamp_spec.rb
+++ b/logstash-core/spec/logstash/legacy_ruby_timestamp_spec.rb
@@ -4,7 +4,12 @@ require "logstash/timestamp"
 require "bigdecimal"
 
 describe LogStash::Timestamp do
-
+  # Via JRuby 9k time see logstash/issues/7463
+  # JRuby 9k now uses Java 8 Time with nanosecond precision but
+  # our Timestamp use Joda with millisecond precision
+  #        expected: 10
+  #             got: 9.999000001
+  # we may need to use `be_within(0.000999999).of()` in other places too
   it "should parse its own iso8601 output" do
     t = Time.now
     ts = LogStash::Timestamp.new(t)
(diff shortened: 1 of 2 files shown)


jsvd:
Is there some explanation on why this happens? does 9k lose precision somewhere and 1.7 didn't?

guyboertje:
I believe 9k now uses the nanosecond precision and 1.7 uses Joda millisecond precision.
https://github.com/elastic/logstash/issues/5606#issuecomment-302221341

jsvd:
@guy oh ok. so we are still only supporting milliseconds since the java LogStash Timestamp class still uses joda?

Can we maybe document that a bit more clearly in that test suite? i.e. stating why we need to work with two degrees of precision.

ph:
This is link to https://github.com/elastic/logstash/issues/7149

guyboertje:
rebased over master

guyboertje:
Found more off by 0.000999999 fails

PR #7480, opened 2017-06-16
[DOCS] Updates to index.asciidoc

lcawl:
This pull request copies the content of the index.asciidoc into new index-shared1.asciidoc and index-shared2.asciidoc files, so that the content can be referenced separately elsewhere.

It also adds some new attributes such as the URL for the X-Pack Reference and Kibana Reference.


docs/index-shared2.asciidoc
```diff
@@ -0,0 +1,111 @@
...
+
+// Troubleshooting performance
+
+:edit_url: https://github.com/elastic/logstash/edit/master/docs/static/performance-checklist.asciidoc
+include::static/performance-checklist.asciidoc[]
+
+// Monitoring overview
+
+:edit_url: https://github.com/elastic/logstash/edit/master/docs/static/monitoring.asciidoc
+include::static/monitoring.asciidoc[]
```
dedemorton: This topic has a couple links (lines 27 and 41) that don't resolve correctly in your build because they use `{xpack}` instead of `{xpack-ref}`.

lcawl: Thanks, I've fixed those references.

suyograo:
LGTM. thanks for doing this @lcawl

PR #7484, opened 2017-06-16
Hash source code when spaces are present

andrewvc:
We were falling back to the graph hash (slow) algorithm anytime spaces were present in source code due to a mixed-up regexp. This clears up that issue.

The test, sadly, had a non-realistic string which didn't catch this error!

### DIFF
--- a/logstash-core/src/main/java/org/logstash/config/ir/PipelineIR.java
+++ b/logstash-core/src/main/java/org/logstash/config/ir/PipelineIR.java
@@ -51,7 +51,7 @@ public class PipelineIR implements Hashable {
 
         this.graph.validate();
 
-        if (this.getOriginalSource() != null && this.getOriginalSource().matches("^\\S+$")) {
+        if (this.getOriginalSource() != null && !this.getOriginalSource().matches("^\\s+$")) {
             uniqueHash = Util.digest(this.getOriginalSource());
         } else {
             uniqueHash = this.graph.uniqueHash();
--- a/logstash-core/src/test/java/org/logstash/config/ir/PipelineIRTest.java
+++ b/logstash-core/src/test/java/org/logstash/config/ir/PipelineIRTest.java
@@ -50,7 +50,7 @@ public class PipelineIRTest {
 
     @Test
     public void hashingWithOriginalSource() throws InvalidIRException {
-        String source = "mysource";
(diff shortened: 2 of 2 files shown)


logstash-core/src/main/java/org/logstash/config/ir/PipelineIR.java
```diff
@@ -51,7 +51,7 @@ public PipelineIR(Graph inputSection, Graph filterSection, Graph outputSection,
 
         this.graph.validate();
 
-        if (this.getOriginalSource() != null && this.getOriginalSource().matches("^\\S+$")) {
+        if (this.getOriginalSource() != null && !this.getOriginalSource().matches("^\\s+$")) {
```
ycombinator: I'm not famililar with this code but doesn't this regexp test if `this.getOriginalSource()` source is made up _entirely_ of whitespace? Should it be something like `.matches("\\s")` instead to check if there is _any_ whitespace in `this.getOriginalSource()` ?

andrewvc: Notice the negation, this tests that the source is not entirely composed of whitespace. If it is entirely composed of whitespace we do the more expensive graph computation.

ycombinator: Ah okay, I misunderstood the title/description of the PR. I thought it meant if _any_ spaces were present. Thanks for clarifying!

original-brownbear: @andrewvc btw. I have no idea how often this is called on average, but this can be way optimized by using a compiled regex.
Or better yet ... to avoid creating the `Matcher` instance just using:

```java
    private static boolean spacesOnly(final CharSequence chars) {
        for (int i = 0; i < chars.length(); ++i) {
            if (chars.charAt(i) != ' ') {
                return false;
            }
        }
        return true;
    }
```

andrewvc:
This is called once on parse per pipeline load. I think the brevity is
worth it here.

original-brownbear:
@andrewvc def, sorry for spamming then <= haven't yet gotten around to looking into this codebase more to know better :(

andrewvc:
@original-brownbear no worries! It's always good to ask questions!

PR #7486, opened 2017-06-17
Update monitoring screen grabs for 5.5

dedemorton:
This updates the screens added in #7483 so they show 5.5 version info instead of 6.0.0-alpha2.


suyograo:
LGTM

PR #7478, opened 2017-06-16
upgrade jruby to 1.7.27

jsvd:
this solves https://github.com/elastic/logstash/issues/7477

Also fixes https://github.com/elastic/logstash/issues/5507: from the jruby changelog "Backport securerandom" https://github.com/jruby/jruby/pull/4149"

This should be merged to 5.x 5.5 and 5.4

PR #7489, opened 2017-06-19
Add a note about entropy level to start Logstash

samuel-phan:
Logstash hangs at startup until it has enough entropy in the system pool. A note about it saves a lot of time of trying to understand why Logstash starts only after several minutes with no CPU, RAM or disk activity.


jsvd:
this will no longer be necessary with https://github.com/elastic/logstash/commit/50cbaf4bf0205af278651208134467a4a23dcee8 and https://github.com/elastic/logstash/pull/7478

PR #7495, opened 2017-06-20
MINOR: Gradle 4.0

original-brownbear:
Gradle 4.0 is out, we should stay up to date I think.


suyograo:
no reason not to upgrade, LGTM.

PR #7510, opened 2017-06-21
[DOCS] Add shared index for doc builds

lcawl:
This pull request retrofits https://github.com/elastic/logstash/pull/7480 to 5.4 and earlier versions. 

In this case, index-shared1.asciidoc is created but index-shared2.asciidoc is not necessary.


lcawl:
Added index-shared2.asciidoc for consistency after all, per feedback from @dedemorton

PR #7515, opened 2017-06-22
Check if the file exist before deleting them

ph:
When running the `ci_acceptance` test we need to make sure we don't have
any Gemfile or *.lock laying around. Deleting a non existant file would
make the test fails.


ph:
@guyboertje I've rebase the PR with #7517 so we rid of the docker container error.

PR #7521, opened 2017-06-22
Only include Logstash plugins in list of plugins

jordansissel:
When generating list of compatible plugins for documentation, we
previously were accidentally include all gems, not just logstash
plugins.


suyograo:
LGTM

Issue #7520, opened 2017-06-22
5.5.0 task: Patch bump all plugins 


jordansissel:
In order to build the latest docs of all plugins, we need a patch release of any plugins that have not been released since we moved plugin docs into asciidoc (previously embedded as comments the Ruby code of the plugins).

This is a blocker for 5.5.0 release because the new 5.5/6.0 docs process uses different tooling to generate the documentation and version bumps of plugins are required to enable this.

Related: https://github.com/elastic/logstash/issues/7188

jordansissel:
Related: https://github.com/elastic/logstash/pull/7521

ph:
I've did the publish this morning.

suyograo:
Keeping it open until we can validate the docs are getting built correctly for 5.5

suyograo:
Some manual changes are needed for 5.5 which are in this PR: https://github.com/elastic/logstash-docs/pull/386

PR #7527, opened 2017-06-23
Bump plugin versions for docs generation

suyograo:
Related to #7520


ph:
I've look at the plugins and all of them are updated.

Issue #4704, opened 2016-02-19
log4j:WARN Please initialize the log4j system properly


robin820224:
I have installed logstash2.2.2 and elasticsearch2.1.1 in centos7. When I used "service logstash configtest", I have got log4j WARM:

[root@iZ25g5gvvtgZ bin]# service logstash configtest
log4j:WARN No appenders could be found for logger (org.apache.http.client.protocol.RequestAuthCache).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
Configuration OK

The follow is my conf file:
input {
  log4j {
    mode => "server"
    host => "localhost"
    port => 4561
  }
}
output {
  elasticsearch {
    hosts => "localhost:9200"
  }
}

jordansissel:
You can ignore this warning; it is meaningless.

We will make it go away in a future release.

robin820224:
After I start logstash service, it seems that logstash had listened on tcp 4561 port. I want to known whether logstash could support UDP protocal? If can, how should I config the conf file for udp protocal?

[root@iZ25g5gvvtgZ conf.d]# netstat -ant | grep 4561
tcp        0      0 0.0.0.0:4561            0.0.0.0:*               LISTEN

zhoufwind:
Hi folks, try to upgrade Logstash to 2.2.2 from 2.1.1 today, received same message.
![2016-02-24-01-logstash-2 2 2-kafka-latency-05](https://cloud.githubusercontent.com/assets/3394857/13278294/73893cea-db0a-11e5-9717-65a25bb8963c.PNG)
Regardless these messages, I have got another problem here: **Logstash take 4 minutes to startup.**
Normally, it only take seconds to complete the collection(read the log, transfer them to kafka broker).
But with version 2.2.2., it will take 2 minutes to read the logs, and spend another 2 minutes to transfer them to kafka broker, pasting test time stamp as below:
02:42PM - Stop logstash service and remove sincedb file manually
02:44PM - Start logstash service
02:46PM - Logstash start to read logs
02:48PM - sincedb file been created, Logstash transfer the logs to Kafka broker(logs transfer to ES immediately, we can check them on Kibana)
![2016-02-24-01-logstash-2 2 2-kafka-latency-03](https://cloud.githubusercontent.com/assets/3394857/13278542/40f48e68-db0c-11e5-9ad4-93c2ca05c086.PNG)
[...]

Issue #7242, opened 2017-05-28
Gradle/Java Unit Tests Have no Log4J Configuration


original-brownbear:
When I run standard gradle tests via `./gradlew test` I see a bunch of this:

```sh
org.logstash.ackedqueue.HeadPageTest > newHeadPage STANDARD_ERROR
    ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console.

org.logstash.DLQEntryTest > testSerDe STANDARD_ERROR
    SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
    SLF4J: Defaulting to no-operation (NOP) logger implementation
    SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

Logging for tests should be correctly configured by putting a `log4j*.properties` file in the test resources so that we can get proper stacktraces from failures and remove the annoying warning.

jakelandis:
There are actually issues here:


1) 
```
org.logstash.ackedqueue.HeadPageTest > newHeadPage STANDARD_ERROR
    ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console.
```
^^ This is caused by the lack of the log4j2.properties on the classpath. This is resolved via https://github.com/elastic/logstash/issues/7526 by adding a default log4j2.properties on the main classpath (not test) to configure the default context and get rid of the annoying warning on startup. The default configuration will get updated very early in the startup lifecycle.


2)
```
org.logstash.DLQEntryTest > testSerDe STANDARD_ERROR
    SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
    SLF4J: Defaulting to no-operation (NOP) logger implementation
    SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```
^^ This is caused by the lack of an SLF4J to log4j2-impl binding on the classpath. Many (most ?), Java 3rd party libraries log via the SLF4J API, so we need to provide an implementation bridge. I will create PR for this. 

_Of side note, 3rd parties may also use JUL (Java Util Logging) or JCL (Jakarta Commons Logging), in which case we could use libraries provided by SLF4J to re-map those API's to the SLF4J API, which would in turn get implemented by Log4J2, but that has the potential to cause performance issues, and no one is asking for it, and those libraries (JUL/JCL) are not heavily used anymore... so I don't plan to add those wrappers._

Issue #4111, opened 2015-10-29
Logstash: Log4j system not initialized properly


cptcanuck:
When starting logstash 2.0, I'm getting

```
# cat /var/log/logstash/logstash.err
log4j:WARN No appenders could be found for logger     (org.apache.http.impl.conn.PoolingHttpClientConnectionManager).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

Freshly installed from website

```
# rpm -qa | grep logstash
logstash-2.0.0-1.noarch
```

Looks to be the same issue as https://logstash.jira.com/browse/LOGSTASH-302

cptcanuck:
Similar to https://discuss.elastic.co/t/log4j-warn-no-appenders-could-be-found-for-logger-node/11907/8

Trying to see exactly how I'd include a log4j.properties file as in https://gist.github.com/fetep/1552466

suyograo:
@cptcanuck what configuration are you using?

cptcanuck:
I'm running the default /etc/default/logstash thta is  installed with the package:

###############################
# Default settings for logstash
###############################

# Override Java location
#JAVACMD=/usr/bin/java

# Set a home directory
#LS_HOME=/var/lib/logstash

# Arguments to pass to logstash agent
#LS_OPTS=""

# Arguments to pass to java
#LS_JAVA_OPTS="-Djava.io.tmpdir=$HOME"

# pidfiles aren't used for upstart; this is for sysv users.
#LS_PIDFILE=/var/run/logstash.pid

# user id to be invoked as; for upstart: edit /etc/init/logstash.conf
#LS_USER=logstash

# logstash logging
#LS_LOG_FILE=/var/log/logstash/logstash.log
#LS_USE_GC_LOGGING="true"

# logstash configuration directory
#LS_CONF_DIR=/etc/logstash/conf.d

# Open file limit; cannot be overridden in upstart
#LS_OPEN_FILES=16384

# Nice level
#LS_NICE=19

# If this is set to 1, then when `stop` is called, if the process has
# not exited within a reasonable time, SIGKILL will be sent next.
# The default behavior is to simply log a message "program stop failed; still running"
KILL_ON_STOP_TIMEOUT=0

and the following config (cleaned up and obfuscated)

input {
        kafka {
                zk_connect => "zk1:2181,zk2:2181,zk3:2181"
                topic_id       => "testtopic"
[...]

PR #7528, opened 2017-06-23
Bug Fix: Log messages from Java code base

### LINKED ISSUE #7526: Log message from Java code won't actually generate logs
Due to the way we initialize the log4j logging context via Ruby, the Java logging context results in using the default context which will only log ERROR message to stdout (and/or sterr). 

This is difficult to reproduce without injecting code for testing since Java logging is already quiet unless specific errors occur. (to reproduce just inject an artificial log message in the Event class, and watch it not show up in the logs)

jakelandis:
This change maps log4j2 context established in Ruby to the logging context used by Java.

Fixes #7526

### DIFF
--- a/logstash-core/build.gradle
+++ b/logstash-core/build.gradle
@@ -116,5 +116,6 @@ dependencies {
     testCompile 'junit:junit:4.12'
     testCompile 'net.javacrumbs.json-unit:json-unit:1.9.0'
     testCompile 'org.elasticsearch:securemock:1.2'
+    testCompile 'org.assertj:assertj-core:3.8.0'
     provided "org.jruby:jruby-core:$jrubyVersion"
 }
--- a/logstash-core/lib/logstash/logging/logger.rb
+++ b/logstash-core/lib/logstash/logging/logger.rb
@@ -8,10 +8,12 @@ module LogStash
     java_import org.apache.logging.log4j.core.config.Configurator
     java_import org.apache.logging.log4j.core.config.DefaultConfiguration
     java_import org.apache.logging.log4j.core.config.LoggerConfig
+    java_import org.logstash.log.LogstashLoggerContextFactory
+    java_import org.apache.logging.log4j.core.LoggerContext
+    java_import java.net.URI
 
     class Logger
       @@config_mutex = Mutex.new
-      @@logging_context = nil
 
       def initialize(name)
         @logger = LogManager.getLogger(name)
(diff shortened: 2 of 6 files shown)


jakelandis:
This likely also fixes 
https://github.com/elastic/logstash/issues/4704
https://github.com/elastic/logstash/issues/4111
https://github.com/elastic/logstash/issues/7242

ph:
@jakelandis I have merged this PR my java branch, but It doesn't seems to fix my issues.

```
[2017-06-28T16:54:21,222][INFO ][logstash.javapipeline    ] Still in ruby
Hello? Logger? # System.out.println()
[2017-06-28T16:54:21,224][INFO ][logstash.javapipeline    ] Back to ruby
```

In the Java code, I am logging at the info level with a simple statement `logger.info("starting pipeline {}", pipelineId)`

jakelandis:
is your log statement in core or a plugin ?

Also, if you are testing from intelliJ, make sure that you also grab this minor change: https://github.com/elastic/logstash/pull/7533 to ensure your classpath is correct. _(else do a full ./gradlew vendor to pick up your changes)_

ph:
This was in core, I didn't test it in IntelliJ and I Duda gradle assemble
to rebuild the jars I don't think it make a difference from vendor I can do
another test

jakelandis:
Do you have a link to your class with the logger ?

ph:
@jakelandis I've created a simple replicable case with the following changes.

I just start logstash on the command line:

```
bin/logstash -e "input { generator {} } output { null {} }" --log.level debug
```

And the only output I am receiving is this:

```
[2017-06-28T19:30:08,344][DEBUG][logstash.outputs.file    ] config LogStash::Outputs::File/@file_mode = -1
Start logging
Stop logging
[2017-06-28T19:30:08,349][DEBUG][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"main"}
```

```diff
diff --git a/logstash-core/lib/logstash/pipeline.rb b/logstash-core/lib/logstash/pipeline.rb
index a5d22a7c0..4fce0a37f 100644
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -29,6 +29,7 @@ require "securerandom"
 java_import org.logstash.common.DeadLetterQueueFactory
 java_import org.logstash.common.SourceWithMetadata
 java_import org.logstash.common.io.DeadLetterQueueWriter
+java_import org.logstash.ReplCase

 module LogStash; class BasePipeline
   include LogStash::Util::Loggable
@@ -253,6 +254,8 @@ module LogStash; class Pipeline < BasePipeline
   end

   def start
+    test = org.logstash.ReplCase.new
+    test.doSomething()
     # Since we start lets assume that the metric namespace is cleared
     # this is useful in the context of pipeline reloading
     collect_stats
```

```java
import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;

public class ReplCase {
    private static final Logger logger = LogManager.getLogger(ReplCase.class);

    public ReplCase() {

    }

    public void doSomething() {
        System.out.println("Start logging");
        logger.info("Log at the info");
        logger.debug("Log at the debug");
        System.out.println("Stop logging");
    }
}

jakelandis:
@ph Thanks for example. 

Digging into it a bit this is not working because Logger in you class is resulting in using the default context (the very bug which this PR sets to fix!). 

It is using the default context because the static initialization of your logger is occurring BEFORE we set up our own context. This class is a little bit special _(more so then the ones I was testing with)_ because it is referenced by a Ruby file (via java_include) that is executed BEFORE we setup our context.  So.... there exists two sets of log4j2 contexts (default and the one we create), and it appears that any class that is referenced BEFORE we setup our context and the logger is established at static scope (which they should be) then those loggers won't work as expected. 

I will take a look at our options here, likely try to move the logging initialization WAY up (like the very first thing) in the chain of startup events.  If not, we may need to re-think establishing our own context.

jakelandis:
@ph - can you re-test ? (The build failure is [un-related](https://github.com/elastic/logstash/issues/7574))

I found the log4j2 magic method that allows us to re-configure the default logging context. So the change that fixes this is to no longer establish our own context, but instead just re-configure the default.  Everything seems happy now regardless of initialization ordering. _(Though non error logging from some static blocks still won't work, but that is a very edge case)_  

Reminder to self: update the commit wording when squashing since we no longer need to describe things in terms of two logging contexts.

jakelandis:
Jenkins test this please

PR #7534, opened 2017-06-25
MINOR: Align JMH JVM settings with production default JVM settings

original-brownbear:
We already have enough diversity of environments we run benchmarking in, let's at least bring sanity to the JVM settings :)


suyograo:
LGTM

original-brownbear:
@suyograo thanks :)

PR #7545, opened 2017-06-26
Add links to grok debugger

dedemorton:
Adds some links that point to the grok debugger so users will know that the tool exists.

Not sure why we were pointing to current in the index, but I think it's better to point to the branch (we actually can't point to current because grok debugger docs won't exist in current until the code is released).


suyograo:
LGTM

Issue #7158, opened 2017-05-18
Add info around upgrading Logstash with PQs enabled


acchen97:
We should add verbiage in the "[Upgrading Logstash](https://www.elastic.co/guide/en/logstash/current/upgrading-logstash.html)" section to include info on upgrading with PQs enabled. Let's target 5.5 for this since PQs went GA in 5.4. A small note should suffice:

"Upgrading Logstash with persistent queues enabled is supported. The persistent queue directory is self-contained and can be read from a new Logstash instance running the same pipeline. You can safely shut down the original LS instance, spin up a new instance, and point to the same queue directory. You can also use a mounted drive to make this workflow easier."

dedemorton:
Closed via #7188

PR #7546, opened 2017-06-27
Add info to docs about upgrading with persistent queues

dedemorton:
@acchen97 @suyograo Please review. I decided to add a separate short section because this info was too long for a note. Plus it's more visible now. I mostly used Alvin's language from #7158, but I thought it would be nice to tell users how to point Lostash to the original queue. Please confirm that particular detail.

BTW, someone has tested this, right? I could set up a couple of instances to test this, but I'm not sure I would catch subtle issues.

### DIFF
--- a/docs/static/upgrading.asciidoc
+++ b/docs/static/upgrading.asciidoc
@@ -16,6 +16,7 @@ See the following topics for information about upgrading Logstash:
 
 * <<upgrading-using-package-managers>>
 * <<upgrading-using-direct-download>>
+* <<upgrading-logstash-pqs>>
 * <<upgrading-logstash-5.0>>
 
 [[upgrading-using-package-managers]]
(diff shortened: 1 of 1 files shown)


docs/static/upgrading.asciidoc
```diff
@@ -43,6 +44,16 @@ This procedure downloads the relevant Logstash binaries directly from Elastic.
...
 
+[[upgrading-logstash-pqs]]
+=== Upgrading with Persistent Queues Enabled
+
+Upgrading Logstash with persistent queues enabled is supported. The persistent
+queue directory is self-contained and can be read by a new Logstash instance
+running the same pipeline. You can safely shut down the original Logstash
+instance, spin up a new instance, and set `path.queue` in the `logstash.yml`
+<<logstash-settings-file,settings file>> to point to the original queue directory.
+You can also use a mounted drive to make this workflow easier.
```
suyograo: Can we add "Keep in mind that just one Logstash instance has access to write to `path.queue`"

suyograo:
Small comment, otherwise LGTM

docs/static/upgrading.asciidoc
dedemorton: @suyograo I'm going to merge my change below. I expanded your statement a bit, but I might need to correct the statement if I misunderstood your reason for wanting the clarification.

PR #7552, opened 2017-06-28
Reset pipeline_settings_obj as explained in the issue 7551

### LINKED ISSUE #7551: Using seed 1332 I was able to reliably get 12 spec failures to occur in rake test:core
The failures we in this test logstash-core/spec/logstash/api/modules/node_spec.rbl under "hot threads"  > "Generic JSON testing".

Seed 1332 runs the pipeline_dlq_commit_spec immediately before the node_spec.

In pipeline_dlq_commit_spec the `pipeline_settings_obj` had specific key values set in a `before(:each)` block but the settings survive into the next test.

**_NOTE: use `LogStash::SETTINGS.clone` instead._**

This issue is a here to inform developers to ensure that `pipeline_settings_obj` do not survive into subsequent tests.

I will submit a PR.

guyboertje:
Fixes #7551

### DIFF
--- a/logstash-core/spec/logstash/pipeline_dlq_commit_spec.rb
+++ b/logstash-core/spec/logstash/pipeline_dlq_commit_spec.rb
@@ -68,6 +68,7 @@ describe LogStash::Pipeline do
 
   after(:each) do
     FileUtils.remove_entry pipeline_settings["path.dead_letter_queue"]
+    pipeline_settings_obj.reset
   end
 
   context "dlq is enabled" do
--- a/logstash-core/spec/logstash/pipeline_dlq_commit_spec.rb
+++ b/logstash-core/spec/logstash/pipeline_dlq_commit_spec.rb
@@ -36,7 +36,7 @@ class DLQCommittingFilter < LogStash::Filters::Base
 end
 
 describe LogStash::Pipeline do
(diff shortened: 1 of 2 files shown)


logstash-core/spec/logstash/pipeline_dlq_commit_spec.rb
```diff
@@ -68,6 +68,7 @@ def close() end
 
   after(:each) do
     FileUtils.remove_entry pipeline_settings["path.dead_letter_queue"]
+    pipeline_settings_obj.reset
```
jsvd: could we instead ensure the object use isn't global? so in [here](https://github.com/guyboertje/logstash/blob/639a39e45071f713f7c985f7d33e8f53796fe0ef/logstash-core/spec/logstash/pipeline_dlq_commit_spec.rb#L39)

`let(:pipeline_settings_obj) { LogStash::SETTINGS }`
maybe do instead:  
`let(:pipeline_settings_obj) { LogStash::SETTINGS.clone }` ?

guyboertje: I was simply following what was done in pipeline_spec.
OTOH does clone bring any previous settings with it?

PR #7557, opened 2017-06-28
add ci_test.bat file so windows can run tests like other OS do

### LINKED ISSUE #7553: add a windows version of the ci/ci_test.sh
Adding a `ci/ci_test.bat` will allow CI to run tests on windows and linux platforms with a closer parity

jsvd:
this requires jruby to be unpacked into ~~`C:\Program Files\jruby\src\jruby-<version>`~~ %JRUBYSRCDIR%
The actual version will be read from the `.ruby-version` file.

fixes #7553 

REMINDER: for backporting this to 5.x it's necessary to include the .ruby-version file in the PR too, since the 5.x and 5.6.0 branches do not contain this file

### DIFF
--- a/ci/ci_test.bat
+++ b/ci/ci_test.bat
@@ -1,0 +1,41 @@
+@echo off
+
+setlocal
+
+REM Since we are using the system jruby, we need to make sure our jvm process
+REM uses at least 1g of memory, If we don't do this we can get OOM issues when
+REM installing gems. See https://github.com/elastic/logstash/issues/5179
+
+SET JRUBY_OPTS="-J-Xmx1g"
+SET SELECTEDTESTSUITE=%1
+SET /p JRUBYVERSION=<.ruby-version
+
+IF NOT EXIST %JRUBYSRCDIR% (
+  echo "Variable JRUBYSRCDIR must be declared with a valid directory. Aborting.."
+  exit /B 1
+)
+
+SET JRUBYPATH=%JRUBYSRCDIR%\%JRUBYVERSION%
+
+IF NOT EXIST %JRUBYPATH% (
+  echo "Could not find JRuby in %JRUBYPATH%. Aborting.."
+  exit /B 1
+)
+
+SET RAKEPATH=%JRUBYPATH%\bin\rake
+
+IF "%SELECTEDTESTSUITE%"=="core-fail-fast" (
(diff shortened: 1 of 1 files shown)


jsvd:
In the Unix world we can control which ruby is used through rbenv and the `.ruby-version` file the logstash root folder.
On Windows we don't have ruby version managers so this must be controlled in the scripts.

jsvd:
The actual value of `C:\Program Files\jruby\src\jruby-<version>` can certainly be injected from the outside. I should that this pr as WIP

jsvd:
@suyograo the comment has been addressed, so up for a new review

ci/ci_test.bat
```diff
@@ -0,0 +1,44 @@
...
+REM Keep in mind to run ci/ci_setup.sh if you need to setup/clean up your environment before
+REM running the test suites here.
+
+REM Since we are using the system jruby, we need to make sure our jvm process
+REM uses at least 1g of memory, If we don't do this we can get OOM issues when
+REM installing gems. See https://github.com/elastic/logstash/issues/5179
+
+SET JRUBY_OPTS="-J-Xmx1g"
+SET SELECTEDTESTSUITE=%1
+SET /p JRUBYVERSION=<.ruby-version
```
original-brownbear: @jsvd we don't have this file in branches other than `master`, that may be a problem, right?
(I guess we could just add it there as well, that would be the cleanest?)

ci/ci_test.bat
```diff
@@ -0,0 +1,44 @@
+@echo off
+
+setlocal
+
+REM Keep in mind to run ci/ci_setup.sh if you need to setup/clean up your environment before
```
original-brownbear: @jsvd I guess this would be a different script for Windows? I guess we could just delete the comment for now?

ci/ci_test.bat
jsvd: yep, when I backport this PR to other branches I'll also include the .ruby-version file

original-brownbear:
@jsvd 2 comment :)

original-brownbear:
@jsvd LGTM now :)

jsvd:
@original-brownbear comments addressed, + added a note to the PR description about including the .ruby-version file in the backport to 5.x

PR #7559, opened 2017-06-28
Minor bug fix: While idle, metrics objects are initialized but effect…

### LINKED ISSUE #7558: Minor: While idle, metrics objects are initialized but effectively unused
To see this behavior, you need to add in some extra logging:

In the base class, add a counter and log for everytime a metric is initilized. In the counter/gauge classes add a log for everytime execute is called. [1]

Start Logstash (any config will do), don't feed it anything, and watch the output [2]. This means that while idle, we are creating a lot of objects with no value in a very tight loop. This shouldn't much noticable impact other then a bit of extra work by the CPU while idle. 

[1]
```
diff --git a/logstash-core/lib/logstash/instrument/metric_type/base.rb b/logstash-core/lib/logstash/instrument/metric_type/base.rb
index b473871b5..17bde5b9e 100644
--- a/logstash-core/lib/logstash/instrument/metric_type/base.rb
+++ b/logstash-core/lib/logstash/instrument/metric_type/base.rb
@@ -3,9 +3,12 @@ require "logstash/util"
 
 module LogStash module Instrument module MetricType
   class Base
+    @@i = 0;
     attr_reader :namespaces, :key
 
     def initialize(namespaces, key)
+      puts namespaces.to_s + ":" + key.to_s
+      puts  @@i +=  1
       @namespaces = namespaces
       @key = key
     end
diff --git a/logstash-core/lib/logstash/instrument/metric_type/counter.rb b/logstash-core/lib/logstash/instrument/metric_type/counter.rb
index e99bca579..eb0301d64 100644
--- a/logstash-core/lib/logstash/instrument/metric_type/counter.rb
+++ b/logstash-core/lib/logstash/instrument/metric_type/counter.rb
@@ -19,6 +19,7 @@ module LogStash module Instrument module MetricType
     end
 
     def execute(action, value = 1)
+      puts "counter:" + action.to_s + ":" + value.to_s
       @counter.send(action, value)
     end
 
diff --git a/logstash-core/lib/logstash/instrument/metric_type/gauge.rb b/logstash-core/lib/logstash/instrument/metric_type/gauge.rb
index 7981bc877..8d98f2d36 100644
--- a/logstash-core/lib/logstash/instrument/metric_type/gauge.rb
+++ b/logstash-core/lib/logstash/instrument/metric_type/gauge.rb
@@ -12,6 +12,7 @@ module LogStash module Instrument module MetricType
     end
 
     def execute(action, value = nil)
+      puts "gauge:" + action.to_s + ":" + value.to_s
       @gauge.set(value)
     end
```

[2] example truncated output...
```
...
5341
counter:increment:0
5342
counter:increment:0
[:stats, :events]:out
5343
counter:increment:0
[:stats, :pipelines, :main, :events]:out
5344
counter:increment:0
[:stats, :events]:filtered
5345
counter:increment:0
[:stats, :pipelines, :main, :events]:filtered
5346
counter:increment:0
[:stats, :events]:out
5347
counter:increment:0
[:stats, :pipelines, :main, :events]:out
5348
counter:increment:0
[:stats, :events]:filtered
5349
counter:increment:0
[:stats, :pipelines, :main, :events]:filtered
...
```

jakelandis:
…ively unused

The change here is to exclude work from being done if the batch size is zero or less.

This change makes me a bit nervous since I am not familiar enough with the pipeline code to know if this will cause any downstream effects. 

Fixes #7558

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -447,15 +447,18 @@ module LogStash; class Pipeline < BasePipeline
       shutdown_requested |= signal.shutdown? # latch on shutdown signal
 
       batch = @filter_queue_client.read_batch # metrics are started in read_batch
-      @events_consumed.increment(batch.size)
-      filter_batch(batch)
-      flush_filters_to_batch(batch, :final => false) if signal.flush?
-      output_batch(batch)
-      break if @force_shutdown.true? # Do not ack the current batch
-      @filter_queue_client.close_batch(batch)
+      if (batch.size > 0)
+        @events_consumed.increment(batch.size)
+        filter_batch(batch)
+        flush_filters_to_batch(batch, :final => false) if signal.flush?
+        output_batch(batch)
+        unless @force_shutdown.true? # ack the current batch
+          @filter_queue_client.close_batch(batch)
+        end
+      end
 
       # keep break at end of loop, after the read_batch operation, some pipeline specs rely on this "final read_batch" before shutdown.
-      break if shutdown_requested && !draining_queue?
+      break if (shutdown_requested && !draining_queue?) || @force_shutdown.true?
     end
 
     # we are shutting down, queue is drained if it was required, now  perform a final flush.


logstash-core/lib/logstash/pipeline.rb
```diff
@@ -447,12 +447,14 @@ def worker_loop(batch_size, batch_delay)
       shutdown_requested |= signal.shutdown? # latch on shutdown signal
 
       batch = @filter_queue_client.read_batch # metrics are started in read_batch
```
jordansissel: I wonder ... what would cause us to get an empty batch (where batch.size==0) ?

jakelandis: I could probably answer that in about 3 days 😄 , but hopefully someone more familiar can chime in.

guyboertje: There is a timeout when waiting to fulfill a batch. Yields a short or empty batch. Also, due to conditionals or cancellations batches can become effectively empty. HTH

guyboertje: Colin can provide more info.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -447,12 +447,14 @@ def worker_loop(batch_size, batch_delay)
...
-      flush_filters_to_batch(batch, :final => false) if signal.flush?
-      output_batch(batch)
-      break if @force_shutdown.true? # Do not ack the current batch
-      @filter_queue_client.close_batch(batch)
+     if (batch.size > 0)
+        @events_consumed.increment(batch.size)
+        filter_batch(batch)
+        flush_filters_to_batch(batch, :final => false) if signal.flush?
+        output_batch(batch)
+        break if @force_shutdown.true? # Do not ack the current batch
```
original-brownbear: @jakelandis I think we shouldn't have this inside the conditional right, otherwise we break `@force_shutdown` don't we?

logstash-core/lib/logstash/pipeline.rb
jordansissel: @guyboertje I agree during filtering the batch could become effectively empty, but it seems strange that `read_batch` would ever return an empty batch. Can you remind me why we wanted this?

logstash-core/lib/logstash/pipeline.rb
jakelandis: I don't _think_ we can currently get into condition of empty batch and force_shutdown.  But that may not always be true, so I updated the logic in https://github.com/elastic/logstash/pull/7559/commits/32633a7cad9af6567a1d47a033f5e5c18e5538f4 (please double check the logic).  I will squash and re-base before a merge.

original-brownbear:
@jakelandis LGTM, let's wait for Travis :)

logstash-core/lib/logstash/pipeline.rb
original-brownbear: @jakelandis yea true, it was more about not changing behavior with certainty :)

jakelandis:
@original-brownbear  thanks for the review. Will wait to merge until @jordansissel questions are answered.

jordansissel:
I am OK merging this. It feels like a workaround while I don't' yet recall or understand why we would want `take_batch` to create batches that are empty (at the start of a whole filter+output chain, I mean). We don't need to let my misunderstanding block this, though, for clarity. I am not against the patch, just confused, but my confusion in this case doesn't concern me with respect to the fix proposed.

+1 to merge.

PR #7560, opened 2017-06-28
Move the global method `register_local_modules` to a Ruby Module (#7555)

guyboertje:
Now in LogStash::Modules::Util, this allows it to be used where modules
are defined in external gems like logastash-x-pack.


suyograo:
LGTM

PR #7562, opened 2017-06-28
Simplier algorithm to retrieve the installable plugins

ph:
Using a recursive algorithm was causing issue with SSL connection to
rubygems when we were manually settings variable to NIL.

This commit uses a simple loop and will try to install the plugins in
sequence and it will give a bit more information at the STDOUT level.


ph:
@jordansissel This will give the following [STDOUT](https://gist.github.com/ph/383a060972d2eb5834825d687f4415fd) and the following [json](https://gist.github.com/ph/fa9f6a48e813bcea71242d6bc3e26254). I've developed the solution on wopr with Jruby 1.7.27.

jordansissel:
Tested manually and this works well.

PR #7564, opened 2017-06-28
Minor bug fix: Stop create unecessary objects with every metric update

### LINKED ISSUE #7563: Minor: metrics creates alot of unecessary objects
This is related to https://github.com/elastic/logstash/issues/7558, and the patch presented there can be used to illustrate the fact that the initialize method of the Counter and Gauge classes are called for each and every time we want to update a value. 

Basically we end up creating a new Counter/Gauge class every time we want to update, but never use the newly created class. There is no functional bug here _(hence the minor)_ , but we should avoid creating so many unused short lived objects.

jakelandis:
The existing implementation has a bug such that the block, which creates a new metric, is called for each and every update. That new metric will be ignored if there is an existing metric found. There is no functional bug here, but we are creating alot of uneeded short lived objects. The fix here is to ensure that the block is only called when needed. Also note that this can not be done atomically since put_if_absent will execute the value parameter statement (e.g. the block) regardless if absent or not.

Fixes #7563

### DIFF
--- a/logstash-core/lib/logstash/instrument/metric_store.rb
+++ b/logstash-core/lib/logstash/instrument/metric_store.rb
@@ -41,26 +41,26 @@ module LogStash module Instrument
     # @param [Symbol] The metric key
     # @return [Object] Return the new_value of the retrieve object in the tree
     def fetch_or_store(namespaces, key, default_value = nil)
-      provided_value =  block_given? ? yield(key) : default_value
 
       # We first check in the `@fast_lookup` store to see if we have already see that metrics before,
       # This give us a `o(1)` access, which is faster than searching through the structured
       # data store (Which is a `o(n)` operation where `n` is the number of element in the namespace and
-      # the value of the key). If the metric is already present in the `@fast_lookup`, the call to
-      # `#put_if_absent` will return the value. This value is send back directly to the caller.
+      # the value of the key). If the metric is already present in the `@fast_lookup`, then that value is sent
+      # back directly to the caller.
       #
-      # BUT. If the value is not present in the `@fast_lookup` the value will be inserted and
-      # `#puf_if_absent` will return nil. With this returned value of nil we assume that we don't
+      # BUT. If the value is not present in the `@fast_lookup` the value will be inserted and we assume that we don't
       # have it in the `@metric_store` for structured search so we add it there too.
-      if found_value = @fast_lookup.put_if_absent(namespaces.dup << key, provided_value)
-        return found_value
-      else
+
+      value = @fast_lookup.get(namespaces.dup << key)
+      if value.nil?
+        value = block_given? ? yield(key) : default_value
+        @fast_lookup.put(namespaces.dup << key, value)
         @structured_lookup_mutex.synchronize do
-          # If we cannot find the value this mean we need to save it in the store.
-          fetch_or_store_namespaces(namespaces).fetch_or_store(key, provided_value)
+            # If we cannot find the value this mean we need to save it in the store.
+          fetch_or_store_namespaces(namespaces).fetch_or_store(key, value)
         end
-        return provided_value
       end
+      return value;
     end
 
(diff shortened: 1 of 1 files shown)


andrewvc:
Great find @jakelandis ! Have you tried benchmarking this to see what the impact is? It'd be good to try this out with the [apache example](https://github.com/elastic/examples/tree/master/ElasticStack_apache).

I usually benchmark it by using a while loop to infinitely cat the sample log file there into a running logstash.

andrewvc:
Actually a better way is:

`time | cat large_log_file | bin/logstash -w 2 -f apache.conf | wc -l`

the wc -l verifies that the output is the right length, and the time command obvs. prints out all the stats we'd like to know.

andrewvc:
LGTM code-wise btw! (pending travis CI)

andrewvc:
@jakelandis we should make sure to backport this to 5.x btw.

jakelandis:
I ran a few different permutations of tests, and couldn't find any apprecialble difference with a realistic workflow on my modern mac. I assume that with real world processing, such as the grok filter, the time to create and collect lots of short lived objects will be statitical noise since the real processing will be the dominator of the time. Perhaps, over time, on a constrained system under alot pressue a real world gain may be possible. 

Using a non-realistic workflow (stdin->stdout) over a large dataset (10M ~ 2.2G data) I was able to show a theortical gain of 2% - 18% throughput.

I ran each test from a cleanly built logstash instance (git clean / rake rake artifact:tar) two times for each brach of code using the [example apache log](https://github.com/elastic/examples/blob/master/ElasticStack_apache/apache_logs) duplicated to 10M events. The braches of code tested were the commit prior to the change (BEFORE), and this change (AFTER). 

``` time  cat /Users/jake/workspace/examples/elastic_apache_logs/apache_logs_10M | bin/logstash -w 8 -e 'input { stdin {} } output { stdout {} }' | wc -l```

## BEFORE:
```
Run1:
real	4m47.866s
user	7m17.877s
sys	1m7.675s

Run2:
real	5m14.807s
user	7m43.184s
sys	1m12.240s
```

## AFTER:
```
Run1:
real	4m22.644s
user	6m56.017s
sys	1m9.130s

Run2:
real	4m41.427s
user	7m15.231s
sys	1m12.464s
```

**Comment by andrewvc** [comment]
@jakelandis thanks! That's quite a bit of variability between runs you're getting there, but even considering that It does look like a small improvement.

At any rate, its a definite improvement over the old code :-D

**Comment by jakelandis** [comment]
I decided to retry the the test which includes a bit of processing over the larger dataset, and found (in the single run) a 7% increase with this change.  Not scientific, but seems inline with other findings. 

_Interestingly you can also do the math between the prior comment and this one to see the cost of the processing_ 

``` time  cat /Users/jake/workspace/examples/elastic_apache_logs/apache_logs_10M | bin/logstash -w 8 -f /Users/jake/workspace/examples/elastic_apache_logs/apache_logstash.conf | wc -l```

## BEFORE:
```
real	12m49.062s
user	74m23.276s
sys	2m4.779s
```

## AFTER:
```
real	11m56.723s
user	69m24.775s
sys	2m5.381s
```

[1]
```
cat apache_logstash.conf
input {  
  stdin { } 
}


filter {
  grok {
    match => {
      "message" => '%{IPORHOST:clientip} %{USER:ident} %{USER:auth} \[%{HTTPDATE:timestamp}\] "%{WORD:verb} %{DATA:request} HTTP/%{NUMBER:httpversion}" %{NUMBER:response:int} (?:-|%{NUMBER:bytes:int}) %{QS:referrer} %{QS:agent}'
    }
  }

  date {
    match => [ "timestamp", "dd/MMM/YYYY:HH:mm:ss Z" ]
    locale => en
  }

  geoip {
    source => "clientip"
  }

  useragent {
    source => "agent"
    target => "useragent"
  }
}

output {
  stdout {
  }
}
```

PR #7565, opened 2017-06-29
5.5 release notes

suyograo:


docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,38 @@
...
+
+This section summarizes the changes in the following releases:
+
+* <<logstash-5-5-0,Logstash 5.5.0>>
+
+[[logstash-5-5-0]]
+=== Logstash 5.5.0 Release Notes
+
+* Introduced a new feature called dead letter queues. When enabled, the dead letter queue (DLQ) 
+  will route events to a file based store when it can't be processed by certain plugins. Currently,
```
dedemorton: I think this description will be a little confusing to users, partly because it's grammatically ambiguous and partly because it's a little vague. How about saying something like this instead?

Introduced a new feature called dead letter queues. When enabled, the dead letter queue (DLQ)  feature will route events to a file-based store when the output plugin is unable to process the events. Currently, the Elasticsearch output is the only plugin that uses this feature to store "dead" events that result from Elasticsearch mapping errors.

[Does this mean that only mapping errors get routed to the queue? Your writeup and my edit suggests that.]

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,38 @@
...
+* <<logstash-5-5-0,Logstash 5.5.0>>
+
+[[logstash-5-5-0]]
+=== Logstash 5.5.0 Release Notes
+
+* Introduced a new feature called dead letter queues. When enabled, the dead letter queue (DLQ) 
+  will route events to a file based store when it can't be processed by certain plugins. Currently, 
+  the Elasticsearch Output is the only plugin that uses this feature to store events that cause a 
+  Elasticsearch mapping error.
+* Upgraded to jruby version 1.7.27.
```
dedemorton: Not sure that we're ever consistent, but we should use correct capitalization: JRuby. :-)

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,38 @@
...
+* Set the default entropy source to `/dev/urandom`. This change resolves an issue that caused Logstash to 
+  stall on startup for several minutes. This stall behavior was especially visible when Logstash was 
+  deployed on VMs ({lsissue}6117[Issue 6117]).
+* Added support to resolve environment variables specified in the `logstash.yml` ({lsissue}6785[Issue 6785]).
```
dedemorton: ..."in the `logstash.yml` file."

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,38 @@
...
+* Introduced a new feature called dead letter queues. When enabled, the dead letter queue (DLQ) 
+  will route events to a file based store when it can't be processed by certain plugins. Currently, 
+  the Elasticsearch Output is the only plugin that uses this feature to store events that cause a 
+  Elasticsearch mapping error.
+* Upgraded to jruby version 1.7.27.
+* Set the default entropy source to `/dev/urandom`. This change resolves an issue that caused Logstash to 
+  stall on startup for several minutes. This stall behavior was especially visible when Logstash was 
+  deployed on VMs ({lsissue}6117[Issue 6117]).
+* Added support to resolve environment variables specified in the `logstash.yml` ({lsissue}6785[Issue 6785]).  
+* Fixed the ability to change log level via the API ({lsissue}7277[Issue 7277]).
```
dedemorton: "to change the log level" or "to change log levels"

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,38 @@
...
+
+[float]
+==== Input Plugins
+
+*`Beats`*:
+
+* Fixed the ability to set SSL protocols.
```
dedemorton: Seems like you need to provide more context here because it's not really clear to me what was broken before.

docs/static/releasenotes.asciidoc
```diff
@@ -0,0 +1,38 @@
...
+
+[float]
+==== Filter Plugins
+
+*`Date`*:
+
+* A warning message is no longer logged when `ISO8601` format and timezone interpolation option is used (https://github.com/logstash-plugins/logstash-filter-date/issues/97[#97]).
```
dedemorton: This makes it sound like there is an option called timezone interpolation. I think you want to say "and timezone interpolation is used" because the way the sentence is written now, people will be looking for an option called timezone interpolation. Or maybe you want to give the actual name of the option in backticks?

docs/static/releasenotes.asciidoc
suyograo: yes only mapping errors right now

docs/static/releasenotes.asciidoc
suyograo: Went back to the change and it is not exactly user facing so I removed it.

suyograo:
@dedemorton updated.

PR #7572, opened 2017-06-29
Make sure we correctly setup the `pipeline.id` in the settings object.

ph:
Currently the settings object is passed around and will contains default
values for the pipeline, to make sure the created pipeline use the right
`pipeline.id` we need to make sure that we clone the settings object and
we set the value of `pipeline.id` to the correct pipeline.


suyograo:
LGTM

ph:
only need for 6.x, 5.x require to use `:main` as the pipeline id.

Issue #7570, opened 2017-06-29
isFullyAcked() semantic


colinsurprenant:
related to #7568

As discussed in #7524 currently the `Queue.isFullyAcked()` method relies on `Page.isFullyAcked()` when there are no tail pages which will returns `false` on an empty (`elementCount == 0`) page. 

This violates the usual empty set algebra and should return `true` when asking: "*is all elements of a set are xxx*" when there are no element.

This currently impacts the `ReadClient#empty?` method which relied on `Queue.isFullyAcked()` per  #7568. 

I suggest we correct the `isFullyAcked()` semantic and review all usages to make sure this does not potentially create unwanted side effect such as its usage in https://github.com/elastic/logstash/blob/b3b9e605ac2e35eaedf0cc9fc0668ca6427fca19/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java#L335-L340

Issue #7568, opened 2017-06-29
ReadClient#empty? is broken


colinsurprenant:
references #7476 #7524 #6779 

This bug manifest when the `queue.drain` is set to `true`. It seems its the only place where `ReadClient#empty?` is currently used. 

- [x] The `ReadClient#empty?` methods has an incorrect reference to the underlying queue to call `is_fully_acked?` which produces a runtime error as noted in #7476. The fix for this is trivial.

- [x] The `ReadClient#empty?` relies on the `is_fully_acked?` method which currently does not take into account the actual *empty* queue condition, as noted and discussed in #7524. For this I suggest we look for an acceptable fix to satisfy `ReadClient#empty?` without changing the semantic of `Queue.isFullyAcked()` which can *possibly* leads to other side effects as also discussed in #7524.  - PR in #7575 

- [x] Add `WrappedAckedQueue#empty?/` tests/specs. - PR in #7575

The `Queue.isFullyAcked()` semantic should be re-evaluated by taking into considerations potential side-effects and I suggest this is done into a separate issue #7570 where the issue raised concerning the related `HeadPage.hasSpace()` [call](https://github.com/elastic/logstash/blob/b3b9e605ac2e35eaedf0cc9fc0668ca6427fca19/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java#L329) in #7524  is also looked at.

colinsurprenant:
Note that the `queue.drain` was never documented so far so the risk for this affecting users is low.

colinsurprenant:
PR #7575 basically fixes this issue. `ReadClient#empty?` now delegates to `WrappedAckedQueue#empty?` which in turn delegate to new `Queue.isEmpty()` method.

colinsurprenant:
closing this will unblock #7583 to merge `queue.drain` documentation.

colinsurprenant:
fixed by #7575

PR #7575, opened 2017-06-29
fix empty queue condition for queue shutdown drain

colinsurprenant:
This is a proposal to fix the second item of #7568 which aim at an acceptable fix to satisfy `ReadClient#empty?` without changing the semantic of `Queue.isFullyAcked()`

The current implementation of `isFullyAcked()` for both the `Queue` and `Page` classes does not work to satisfy the `ReadClient#empty?` condition, see #7570.

- [x] add missing unit tests and specs.

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -100,6 +100,10 @@ module LogStash; module Util
       end
     end
 
+    def is_empty?
+      @queue.is_empty?
+    end
+
     def close
       @queue.close
       @closed.make_true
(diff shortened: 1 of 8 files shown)


colinsurprenant:
Once this is merged we should move forward with #7583

colinsurprenant:
ping @guyboertje

colinsurprenant:
merged in master, 5.x and 5.5

seang-es:
present in 5.5.3+

colinsurprenant:
@seang-es if you mean *fixed* in 5.5.2+ then yes, the fixed was merged in the 5.5 branch and included starting with 5.5.2. Labels are missing here, I will label it.

PR #7582, opened 2017-06-30
Add more options for the modules templates

ph:
This commits add the following new features:

Allow users to use the internal settings classes to have better handling
over the primitives values that a user can use in the modules templates,
by default we are using a `NullableString` which is the more flexible
type.

This change open the way to have more strict validation on the module
templates and will allow us to preprocess configuration to retrieve the
list of the settings and their settings types.

We also add a new type of settings called `SplittableStringArray`, this
class is a subclass or `ArrayCoercible` but target module CLI
configuration. It will take the following form `-M
"arcsight.var.input.hosts=127.0.0.1:3333,192.168.1.23:9999"` and
transform it into a ruby array of string.

We can also add alias of settings values in the configuration by using the
`alias_settings_keys!()`, this is usefull when the language domain is
different from the logstash domain.

### DIFF
--- a/logstash-core/lib/logstash/modules/logstash_config.rb
+++ b/logstash-core/lib/logstash/modules/logstash_config.rb
@@ -1,9 +1,9 @@
 # encoding: utf-8
 require "logstash/namespace"
 require_relative "file_reader"
+require "logstash/settings"
 
 module LogStash module Modules class LogStashConfig
-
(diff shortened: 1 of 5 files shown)


jsvd:
This has not yet been merged into 5.6.0, which is a prerequisite to backporting #7591

PR #7587, opened 2017-07-04
 #7178 retry building Kafka image

original-brownbear:
Latest Docker didn't fix the repeat failures in #7178 either. Let's try retrying for now since this seems to be a random occurrence.


original-brownbear:
@jsvd green a few times now ... merge + hope for the best? :S

jsvd:
ok. out of curiosity, was any of the green builds green because of the retry change?

original-brownbear:
@jsvd Nah :( if it was I would've had more enthusiasm when asking if we should merge :P

original-brownbear:
merging

PR #7589, opened 2017-07-05
MINOR: Removed some dead code in Java Accessor UT

original-brownbear:
Just found this looking into some test failures. 
No need to keep this dead code I think, tests pass just fine after removing it.

### DIFF
--- a/logstash-core/src/test/java/org/logstash/AccessorsTest.java
+++ b/logstash-core/src/test/java/org/logstash/AccessorsTest.java
@@ -1,19 +1,19 @@
 package org.logstash;
 
-import org.junit.experimental.theories.DataPoint;
-import org.junit.Rule;
-import org.junit.rules.ExpectedException;
-import org.junit.experimental.theories.Theories;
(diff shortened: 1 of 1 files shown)


logstash-core/src/test/java/org/logstash/AccessorsTest.java
```diff
@@ -261,14 +257,4 @@ public void testListIndexOutOfBounds(int i) {
       }
     }
 
-    public static class TestListIndex {
-      public void testListIndexOutOfBounds() {
```
jakelandis: This looks like it may have been an attempt to test these cases.  What happens if you just pull the test method out of the inner class and slap a @Test annotation on it ?

original-brownbear: @jakelandis passes, good catch :) let's do that then, sec adjusting PR

original-brownbear:
@jakelandis test revived ;)

jakelandis:
LGTM

PR #7590, opened 2017-07-05
MINOR: Optimized Map and List allocation

original-brownbear:
Profiling a test run on our apache logs dataset shows a lot of GC on `Object[]` and `HashMapNode`
This reduces the noise from that a little by removing some collection allocations without a size hint when the size is known to us.


jakelandis:
LGTM

Issue #7396, opened 2017-06-09
A look into metric collection overhead in logstash 5.4.x


jsvd:
### Introduction

In an attempt to understand the current overhead of metric collection the latest logstash stable version, I set out to compare the event throughput in a `generator -> elasticsearch` pipeline. 
For this I used logstash 5.4 with the lastest elasticsearch output. To remove the influence from network layer, I also modified the elasticsearch output to not perform the bulk request but always return a successful response.

#### Logstash

Downloaded logstash 5.4.0 and updated the logstash-output-elasticsearch to 7.2.1

`bin/logstash-plugin install --version "7.2.1" logstash-output-elasticsearch`

##### Mocking network call in elasticsearch output

```patch
diff --git a/lib/logstash/outputs/elasticsearch/http_client.rb b/lib/logstash/outputs/elasticsearch/http_client.rb
index daca851..714c4b1 100644
--- a/lib/logstash/outputs/elasticsearch/http_client.rb
+++ b/lib/logstash/outputs/elasticsearch/http_client.rb
@@ -135,12 +135,14 @@ module LogStash; module Outputs; class ElasticSearch;
     def bulk_send(body_stream)
       params = http_compression ? {:headers => {"Content-Encoding" => "gzip"}} : {}
       # Discard the URL 
-      _, response = @pool.post(@bulk_path, params, body_stream.string)
+      #_, response = @pool.post(@bulk_path, params, body_stream.string)
+      body_stream.string
       if !body_stream.closed?
         body_stream.truncate(0)
         body_stream.seek(0)
       end
-      LogStash::Json.load(response.body)
+      #LogStash::Json.load(response.body)
+      {"took" => 30, "errors" => false, "items" => [] }
     end
 
     def get(path)
```

#### Logstash Pipeline Configuration

```
input {
  # 10 million messages
  generator { message => '64.242.88.10 - - [07/Mar/2004:16:05:49 -0800] "GET /twiki/bin/edit/Main/Double_bounce_sender?topicparent=Main.ConfigurationVariables HTTP/1.1" 401 12846' count => 10000000 }
}
output {
  # es output with bulk network call mocked out. see .patch files
  elasticsearch {
    hosts => "localhost"
    index => "logstash-%{+YYYY.MM.dd.HH}"
  }
}
```

### Test executions

a) With metric collection enabled by setting `metric.collect: true` in the logstash.yml

```
[...]

PR #7274, opened 2017-05-30
Initial work to centralize metrics

andrewvc:
This is in preparation for #7155 and takes over from #7229 

We will need further PRs to move the rest of the metrics over in the future.

This is the first step in moving our metrics to java and making initialization of metrics more explicit. This patch defines just enough java interfaces to allow us to start moving metrics over and
centralizing their lifecycle through a simple java object graph.

Eventually we won't need metrics collectors etc. and can just traverse this structure.

PR #7595, opened 2017-07-05
Metrics: Move counters and gauges to Java with stricter types.

### LINKED ISSUE #7509: Improved Metric type definitions
The current metric types are problematically loose in their use of types. For the purposes of charting we need different formulas for different conceptual things than we have in our current contract. Luckily our de-factor contract is mostly correct, but we need to tighten up the real interfaces. Where we should get to is:

* Counter - Currently this increments *and* decrements. It is only used in existing code for increment-only operations, things like `bytes_sent`. We should remove the ability to decrement (or increment by a negative), this will make charting counters dependable.
* Gauge - This is currently an object wrapper holding anything. This should change to be a double type that can increment, decrement, or be set to a new value. While there are some issues around incrementing doubles at some level we can document these. Doubles are accurate to 2^53, which is more than enough for everyone :-D. Users are free to use their own long object and record to a double if need be in the worst case.
* Text - String value that can only be set, useful for things like recording a protocol version etc.

jakelandis:
This change moves the definition of the counters and gauges from Ruby to Java, providing a (passive) Ruby bridge that allows support for Ruby duck typing to Java strict typing.
This change also moves the serialization from JRuby (Jackson) to Java (Jackson).
This change also removes the decrement from counter api. There are no known uses of decrement, so while not technically passive, shouldn't cause issues. Only allowing counters to move in one direction (up) allows for reliable aggregations over time.

Fixes #7509 
--------
Sorry for the huge PR, but most of it is boiler plate.  I would suggest to start with the removal of the base.rb and subclassing the counter.rb and gauge.rb classes, most everything else is the same, just different types. 

Note - no changes to the integration tests are required, and anecdotally it appears to be the same output as prior to this change, though some more testing with something that consumes the output JSON may be advisable. 

Also [here](https://drive.google.com/file/d/0B8ymq49iNkjJMGUtQnNOOUVEOVE/view) is the class diagram.


jakelandis:
This does not seem to have much if any impact on the speed of metrics: https://github.com/elastic/logstash/issues/7396#issuecomment-313247861

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/LazyDelegatingCounter.java
```diff
@@ -0,0 +1,81 @@
...
+            lazyMetric.increment(by);
+        }
+    }
+
+    /**
+     * Instantiates the metric based on the type used to initialize or increment the counter
+     *
+     * @param by The object used to initialize or increment
+     */
+    private void wakeMetric(Object by) {
```
jakelandis: Todo: synchronize this

logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
```diff
@@ -0,0 +1,94 @@
...
+            lazyMetric.set(value);
+        }
+    }
+
+    /**
+     * Instantiates the metric based on the type used to set this Gauge
+     *
+     * @param value The object used to set this value
+     */
+    private void wakeMetric(Object value) {
```
jakelandis: Todo: synchronize this

logstash-core/src/test/java/org/logstash/instrument/metrics/counter/DoubleCounterTest.java
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    public void increment() {
+        doubleCounter.increment();
+        assertThat(doubleCounter.getValue()).isEqualTo(INITIAL_VALUE + 1);
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void incrementByNegativeValue() {
+        doubleCounter.increment(-.1);
+        assertThat(doubleCounter.getValue()).isEqualTo(INITIAL_VALUE + 100);
```
jakelandis: Todo: remove this assert , it never gets called (and look for others)

jakelandis:
In summary the non-passive bits are:
* Removal of the base (Ruby) class
* Removal of decrements from Counter
* Removes to_hash from base class (could not find any usage of it beyond the unit tests)
* Minor change the output of the type() method call 
* Minor change to the output of the inspect() method. 

Due to these non-passive changes, I suggest targeting this only for 6.x. 

_Everything else should be passive, including the serialized output._

jakelandis:
> I take it this is splitting the work I started in #7274 and splitting it in half such that in this PR we're just javaifying the metrics classes, with the move from namespaces -> POJOs happening in a subsequent PR?

@andrewvc - Yup, I wanted to get this part done first , then layer in the domain specific metric classes you proposed _(e.g. AgentMetrics, PipelineMetrics, etc.)_ , then look to moving the metrics_store to Java, likely backed by a proper Tree structure.

logstash-core/lib/logstash/instrument/metric_type/counter.rb
```diff
@@ -1,29 +1,17 @@
-# encoding: utf-8
-require "logstash/instrument/metric_type/base"
-require "concurrent"
+#encoding: utf-8
+java_import org.logstash.instrument.metrics.counter.LazyDelegatingCounter
```
jakelandis: OK, I will remove the DoubleCounter and restrict the Java code to a Long type and ensure that the lazy instantiation will throw an exception if not Long type.  I too was having trouble thinking of a valid use case for it.

logstash-core/build.gradle
```diff
@@ -109,12 +109,14 @@ dependencies {
     compile 'org.apache.logging.log4j:log4j-core:2.6.2'
     compile 'com.fasterxml.jackson.core:jackson-core:2.7.4'
     compile 'com.fasterxml.jackson.core:jackson-databind:2.7.4'
+    compile 'com.fasterxml.jackson.core:jackson-annotations:2.7.4'
```
andrewvc: I believe this will be a conflict when master gets reverted to 2.7.3 due to #7614

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/CounterMetric.java
```diff
@@ -0,0 +1,22 @@
+package org.logstash.instrument.metrics.counter;
+
+import org.logstash.instrument.metrics.Metric;
+
+/**
+ * A {@link Metric} to count things. A counter can only increment, and there are no guarantees of durability of current value between JVM restarts.
+ * @param <T> The underlying {@link Number} type that can be incremented. Care should be taken to which {@link Number} is used to back the counter, since some {@link Number}
+ *           types may not be appropriate for a monotonic increasing series.
+ */
+public interface CounterMetric<T extends Number> extends Metric<T> {
```
andrewvc: Do we still need this interface now that we only have a `long` counter?

logstash-core/src/main/java/org/logstash/instrument/metrics/MetricType.java
```diff
@@ -0,0 +1,65 @@
...
+public enum MetricType {
+
+    /**
+     * A counter backed by a {@link Long} type
+     */
+    COUNTER_LONG("counter/long"),
+    /**
+     * A counter backed by {@link Double} type
+     */
+    COUNTER_DOUBLE("counter/double"),
```
andrewvc: Should be removed now that this doesn't exist.

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/LazyDelegatingCounter.java
```diff
@@ -0,0 +1,79 @@
...
+import org.logstash.instrument.metrics.MetricType;
+
+import java.util.List;
+
+/**
+ * A lazy proxy to a more specific typed {@link CounterMetric}. The metric will only be initialized if an initial value is provided, or once an {@code increment} operation is
+ * called.
+ * <p><strong>Intended only for use with Ruby's duck typing, Java consumers should use the specific typed {@link CounterMetric}</strong></p>
+ */
+public class LazyDelegatingCounter extends AbstractMetric<Long> implements CounterMetric<Long> {
```
andrewvc: Given that we only have a single `LongCounter` maybe we can just delete this class and use `LongCounter directly`.

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/LongCounter.java
```diff
@@ -0,0 +1,65 @@
...
+    public void increment() {
+        increment(1l);
+    }
+
+    /**
+     * {@inheritDoc}
+     * throws {@link UnsupportedOperationException} if attempt is made to increment by a negative value
+     */
+    @Override
+    public void increment(Long by) {
```
andrewvc: Can we move `Long` to the primitive `long`?

logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
```diff
@@ -0,0 +1,94 @@
...
+            if (value instanceof Number) {
+                lazyMetric = new NumericGauge(nameSpaces, key, (Number) value);
+            } else if (value instanceof String) {
+                lazyMetric = new TextGauge(nameSpaces, key, (String) value);
+            } else if (value instanceof Boolean) {
+                lazyMetric = new BooleanGauge(nameSpaces, key, (Boolean) value);
+            } else if (value instanceof RubyHash) {
+                lazyMetric = new RubyHashGauge(nameSpaces, key, (RubyHash) value);
+            } else {
+                lazyMetric = new UnknownGauge(nameSpaces, key, value);
```
andrewvc: I'm wondering now if we should warn on unknown gauge types and ask users to not use them?

Since our long term plan in #7509 is to only support double/text/boolean, it might be good to warn on unknown for now, and once we've refactored the LS internals then move that warning up to rubyhash ?

logstash-core/src/test/java/org/logstash/instrument/metrics/counter/LongCounterTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    public void noInitialValue() {
+        LongCounter counter = new LongCounter(Collections.singletonList("foo"), "bar");
+        counter.increment();
+        assertThat(counter.getValue()).isEqualTo(1l);
+    }
+
+    @Test
+    @SuppressWarnings( "deprecation" )
```
andrewvc: What is the warning we're suppressing here?

logstash-core/build.gradle
jakelandis: Yup...it's a race to which one will conflict first. Since the other Issue is blocked by a Jenkins build issue, this will probably go cleanly (first).

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/CounterMetric.java
jakelandis: For consistency for the Java API's I think it still makes sense. This would be available to plugin developers to implement other type of counters such BigInteger if needed. It also allows for a passive path for additions of other types of counters.

logstash-core/src/main/java/org/logstash/instrument/metrics/MetricType.java
jakelandis: thanks .. will remove.

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/LazyDelegatingCounter.java
jakelandis: Yeah, you're right, this class is basically useless now :) 
Will delete.

logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
jakelandis: Sure, I can a WARN level message here.

logstash-core/src/test/java/org/logstash/instrument/metrics/counter/LongCounterTest.java
jakelandis: This is a deprecation warning. To keep the API (mostly) passive to Ruby, I default implemented a couple of Ruby methods (type, get, inspect) in the Metrics interface and deprecated them so that Java consumers won't use them.

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/LongCounter.java
jakelandis: Not here since increment is defined as ```increment(T by)``` where the generic is satisfied by a ```Long``` type.  

This could be an argument to to remove the CounterMetric (where the generic is defined). However, the Metric<T> generic type would still need to be satisfied by ```Long``` resulting in an increment by a primitive, but a get by the ```Long```  version.   This may be an appropriate micro-optimization to avoid autoboxing for every increment.  

If nothing else, I should change increment() (no args) to use the longAdder directly. 

WDYT ?

jakelandis:
Pushed changes for review : https://github.com/elastic/logstash/pull/7595/commits/2d8f17c64f75f7cdc1481547b17c57188fd27188

I ended up keeping the ```CounterMetric``` and added a method to help avoid autoboxing of the ```Long``` counter.  (plus the other requested changes)

logstash-core/src/test/java/org/logstash/instrument/metrics/counter/LongCounterTest.java
andrewvc: Ah, very nice. TIL!

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/LongCounter.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Constructor
+     *
+     * @param nameSpace    The namespace for this metric
+     * @param key          The key <i>(with in the namespace)</i> for this metric
+     * @param initialValue The initial value for this {@link CounterMetric}
+     */
+    public LongCounter(List<String> nameSpace, String key, long initialValue) {
```
andrewvc: We should lose this contsructor since the contract should always be an initial value of zero.

jakelandis: Removed in the latest round of changes: https://github.com/elastic/logstash/pull/7595/commits/6cb5902f000c2b92004a5e593e1e44476b7c9c6a

jakelandis:
@andrewvc  - This should be ready for the final review. 

I have kept all changes in separate commits for easier diff viewing, but will squash prior to merge.

logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
jakelandis: @andrewvc - nice call on the warning message here!  Due to the warning message, I found a bug (now fixed) such that if a metric sends a null value for the first attempt to set it, then it resulted in an Unknown gauge.

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/LongCounter.java
andrewvc: @jakelandis and I discussed this and a `increment(long by)` was added in the concrete class.

jakelandis:
Closing and re-opening to force build on squashed commit.

Issue #7594, opened 2017-07-05
ES Output IT is broken on Travis + Locally


original-brownbear:
I'm seeing this on Travis https://travis-ci.org/elastic/logstash/jobs/250478108#L2772

```sh
Test Elasticsearch output
/home/travis/build/elastic/logstash/qa/integration/services/logstash_service.rb:40: warning: Dir.exists? is a deprecated name, use Dir.exist? instead
Using /home/travis/build/elastic/logstash/build/logstash-6.0.0-alpha3-SNAPSHOT as LS_HOME
Setting up services
Setting up logstash service
Setup script not found for logstash
logstash service setup complete
Setting up elasticsearch service.
Building the base container image.
Finished building the base image.
Building the container image.
Finished building the image.
Starting the container.
Finished starting the container.
Finished setting up elasticsearch service.
ERROR StatusLogger No log4j2 configuration file found. Using default configuration: logging only errors to the console.
  can ingest 37K log lines of sample apache logs (FAILED - 1)
Tearing down services
Tearing down elasticsearch service.
Stop the container.
Finished stopping the container.
Finished tearing down of elasticsearch service.
Failures:
  1) Test Elasticsearch output can ingest 37K log lines of sample apache logs
     Failure/Error: expect(result["hits"]["total"]).to eq(37)
     
       expected: 37
            got: 0
     
       (compared using ==)
     # ./specs/es_output_how_spec.rb:25:in `block in (root)'
     # /home/travis/.rvm/gems/jruby-9.1.10.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in (root)'
Finished in 18 minutes 44 seconds (files took 8.91 seconds to load)
31 examples, 1 failure
```

Locally I get more log output:

```sh

➜  integration git:(master) bundle exec rspec specs/es_output_how_spec.rb

Test Elasticsearch output
  can ingest 37K log lines of sample apache logs (FAILED - 1)

An error occurred in an `after(:context)` hook.
Failure/Error: es_client = @fixture.get_service("elasticsearch").get_client

NoMethodError:
  undefined method `get_service' for nil:NilClass
# ./specs/es_output_how_spec.rb:12:in `block in (root)'

Failures:

  1) Test Elasticsearch output can ingest 37K log lines of sample apache logs
     Failure/Error: @fixture = Fixture.new(__FILE__)
     
     NameError:
       missing class name (`org.apache.logging.log4j.Level')
[...]

PR #7597, opened 2017-07-06
[WIP] #7594 add missing spec helper include

original-brownbear:
This seems to fix at least the dependency situation locally


original-brownbear:
not relevant anymore, fix incoming in https://github.com/logstash-plugins/logstash-filter-useragent/issues/48

PR #7600, opened 2017-07-06
MINOR: Removed dead casts and interface declarations

original-brownbear:
Just 3 small spots I found researching something:

* `List` implies `Collection` and `Iterable`
* Removed two redundant casts

### DIFF
--- a/logstash-core/src/main/java/org/logstash/KeyNode.java
+++ b/logstash-core/src/main/java/org/logstash/KeyNode.java
@@ -30,7 +30,7 @@ public class KeyNode implements TemplateNode {
                 return join((List)value, ",");
             } else if (value instanceof Map) {
                 ObjectMapper mapper = new ObjectMapper();
-                return mapper.writeValueAsString((Map<String, Object>)value);
+                return mapper.writeValueAsString(value);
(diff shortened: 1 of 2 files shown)


logstash-core/src/main/java/org/logstash/KeyNode.java
```diff
@@ -60,8 +60,8 @@ private static String toString(Object value, String delim) {
         if (value == null) return "";
         if (value instanceof List) return join((List)value, delim);
         if (value instanceof BiValue) {
-            return ((BiValue) value).toString();
+            return value.toString();
```
jakelandis: will this print the BiValue#toString , or just the Object#toString when removing the explicit cast ?

original-brownbear: @jakelandis `BiValue#toString` is an override to `Object#toString` here, since we know that the `value` is a `BiValue` => `BiValue#toString` will be called (now that I'm explaining this I'm getting the urge to clean this up :P ... we should really not have chains of type checks ...)

jakelandis:
LGTM

PR #7602, opened 2017-07-06
Logging: Add log4j-slf4j-impl to classpath to allow logging via SLF4J.

### LINKED ISSUE #7242: Gradle/Java Unit Tests Have no Log4J Configuration

jakelandis:
Fixes #7242

### DIFF
--- a/logstash-core/build.gradle
+++ b/logstash-core/build.gradle
@@ -107,6 +107,7 @@ idea {
 dependencies {
     compile 'org.apache.logging.log4j:log4j-api:2.6.2'
     compile 'org.apache.logging.log4j:log4j-core:2.6.2'
+    runtime 'org.apache.logging.log4j:log4j-slf4j-impl:2.6.2'
     compile 'com.fasterxml.jackson.core:jackson-core:2.7.4'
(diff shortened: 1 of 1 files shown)


original-brownbear:
@jakelandis LGTM, we should probably fix the log config next (it pipes the logs to a file now in the tests which is kind of pointless, but I can add an issue for that).

jakelandis:
If I am reading the log4j2.properties (config/log4j2.properties) correctly, it should log to both the console and file.

original-brownbear:
@jakelandis right ... nevermind :) Happy :D

PR #7603, opened 2017-07-06
MINOR: Much Faster String Java to Ruby Conversion in ConvertedMap + Corrected Types

original-brownbear:
This actually shows up with a visible performance gain for me (~5-10% on the Apache dataset).

Two problems here:

```java
String.valueOf(BiValues.newBiValue(entry.getKey()).javaValue());
```

is not necessary. Both the Ruby and the Java strings are serializable, Java String just returns identity for the `toString`, `RubyString` returns Java string for it.
=> Saving at least one Object for each key here.

Also, the type `Map<String,Object>` was actually not correct here in the first place.
We were simply casting a `org.jruby.java.proxies.JavaProxy#getObject` here and it, in fact, contained `RubyString` as key type in practice (remember, casting generics like this isn't a safe operation).

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ConvertedMap.java
+++ b/logstash-core/src/main/java/org/logstash/ConvertedMap.java
@@ -1,9 +1,6 @@
 package org.logstash;
 
-import org.logstash.bivalues.BiValues;
-import org.jruby.RubyHash;
-import org.jruby.runtime.builtin.IRubyObject;
-
+import java.io.Serializable;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


jakelandis:
@original-brownbear - Great find!

You're right, it looks like BiValue#javaValue() (for String types) just calls [RubyString#asJavaString()](https://github.com/jruby/jruby/blob/9.1.12.0/core/src/main/java/org/jruby/RubyString.java#L1304) which just calls RubyString#ToString() making the call to BiValue wrapper unnecessary. 

I was able to reproduce a 6% increase in speed on the apache data set from the examples, (data duplicated to 10M records).  

Before:
```
real	11m58.021s
user	71m25.271s
sys	2m10.386s
```
After:
```
real	11m16.642s
user	67m28.763s
sys	2m2.547s
```
oh yeah ... LGTM

original-brownbear:
@jakelandis thanks for testing too! :) Didn't merge clean to `5.x`, will backport tomorrow :)

PR #7604, opened 2017-07-06
broken Event.toString() when missing @timestamp

### LINKED ISSUE #6756: Bivalue error when trying to parse logs
I get this un-actionable error when I try to parse the sample log with the below configuration. Can we catch such error messages and present a better user-facing one?

```
org.logstash.bivalues.StringBiValue cannot be cast to java.lang.String
```

```
Exception in thread "[main]>worker4" java.lang.ClassCastException: org.logstash.bivalues.StringBiValue cannot be cast to java.lang.String
	at org.logstash.Event.toString(Event.java:312)
	at org.logstash.ext.JrubyEventExtLibrary$RubyEvent.ruby_to_s(JrubyEventExtLibrary.java:209)
	at org.logstash.ext.JrubyEventExtLibrary$RubyEvent$INVOKER$i$0$0$ruby_to_s.call(JrubyEventExtLibrary$RubyEvent$INVOKER$i$0$0$ruby_to_s.gen)
	at org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:306)
	at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:136)
	at org.jruby.ast.CallNoArgNode.interpret(CallNoArgNode.java:60)
	at org.jruby.ast.CallOneArgNode.interpret(CallOneArgNode.java:57)
	at org.jruby.ast.CallTwoArgNode.interpret(CallTwoArgNode.java:59)
	at org.jruby.ast.NewlineNode.interpret(NewlineNode.java:105)
	at org.jruby.ast.IfNode.interpret(IfNode.java:118)
	at org.jruby.ast.NewlineNode.interpret(NewlineNode.java:105)
	at org.jruby.evaluator.ASTInterpreter.INTERPRET_METHOD(ASTInterpreter.java:74)
	at org.jruby.internal.runtime.methods.InterpretedMethod.call(InterpretedMethod.java:182)
	at org.jruby.internal.runtime.methods.DefaultMethod.call(DefaultMethod.java:203)
	at org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:326)
	at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:170)
	at org.jruby.ast.CallOneArgNode.interpret(CallOneArgNode.java:57)
	at org.jruby.ast.NewlineNode.interpret(NewlineNode.java:105)
	at org.jruby.evaluator.ASTInterpreter.INTERPRET_BLOCK(ASTInterpreter.java:112)
	at org.jruby.runtime.Interpreted19Block.evalBlockBody(Interpreted19Block.java:206)
	at org.jruby.runtime.Interpreted19Block.yield(Interpreted19Block.java:157)
	at org.jruby.runtime.Block.yield(Block.java:142)
	at org.jruby.RubyArray.eachCommon(RubyArray.java:1606)
	at org.jruby.RubyArray.each(RubyArray.java:1613)
	at org.jruby.RubyArray$INVOKER$i$0$0$e
```

Log line:
```
192.168.33.1 - - [26/Dec/2016:16:22:00 +0000] "GET / HTTP/1.1" 200 484 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.98 Safari/537.36"
```

Config:
```
input {
  stdin {}
}
filter {
   grok {
      match => { "message" => ["%{IPORHOST:apache2.access.remote_ip} - %{DATA:apache2.access.user_name} \\[%{HTTPDATE:apache2.access.time}\\] \"%{WORD:apache2.access.method} %{DATA:apache2.access.url} HTTP/%{NUMBER:apache2.access.http_version}\" %{NUMBER:apache2.access.response_code} %{NUMBER:apache2.access.body_sent.bytes}( \"%{DATA:apache2.access.referrer}\")?( \"%{DATA:apache2.access.agent}\")?", 
        "%{IPORHOST:apache2.access.remote_ip} - %{DATA:apache2.access.user_name} \\[%{HTTPDATE:apache2.access.time}\\] \"-\" %{NUMBER:apache2.access.response_code} -" ] }
      remove_field => "message"
   }
   mutate {
      rename => { "@timestamp" => "read_timestamp" }
   }
   date {
      match => [ "apache2.access.time", "dd/MMM/YYYY:H:m:s Z" ]
      remove_field => "apache2.access.time"
   }
   useragent {
      source => "apache2.access.agent"
      target => "apache2.access.user_agent"
      remove_field => "apache2.access.agent"
   }
   geoip {
      source => "apache2.access.remote_ip"
      target => "apache2.access.geoip"
   }
}
output {
  stdout {}
}
```

colinsurprenant:
Fixes #6756 

The problem is the missing `@timestamp` handling which produces this exception when `Event.toString()` is called.

```
org.logstash.bivalues.StringBiValue cannot be cast to java.lang.String
```

Will also followup with an issue to review the `Event.getTimestamp()` behaviour in #7613

To manually repro this bug
```sh
bin/logstash -e 'filter{mutate{rename => {"@timestamp" => "foo"}}} output{stdout{}}'
```

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Event.java
+++ b/logstash-core/src/main/java/org/logstash/Event.java
@@ -264,17 +264,15 @@ public class Event implements Cloneable, Serializable, Queueable {
     }
 
     public String toString() {
-        // TODO: (colin) clean this IOException handling, not sure why we bubble IOException here
+        Object hostField = this.getField("host");
+        Object messageField = this.getField("message");
+        String hostMessageString = (hostField != null ? hostField.toString() : "%{host}") + " " + (messageField != null ? messageField.toString() : "%{message}");
+
         try {
-            return (getTimestamp().toIso8601() + " " + this.sprintf("%{host} %{message}"));
+            // getTimestamp throws an IOException if there is no @timestamp field, see #7613
+            return getTimestamp().toIso8601() + " " + hostMessageString;
         } catch (IOException e) {
-            String host = (String)this.data.get("host");
-            host = (host != null ? host : "%{host}");
-
-            String message = (String)this.data.get("message");
-            message = (message != null ? message : "%{message}");
-
-            return (host + " " + message);
+            return hostMessageString;
         }
     }
 
--- a/logstash-core/spec/logstash/legacy_ruby_event_spec.rb
+++ b/logstash-core/spec/logstash/legacy_ruby_event_spec.rb
@@ -590,15 +590,6 @@ describe LogStash::Event do
   describe "#to_s" do
     let(:timestamp) { LogStash::Timestamp.new }
     let(:event1) { LogStash::Event.new({ "@timestamp" => timestamp, "host" => "foo", "message" => "bar"}) }
-    let(:event2) { LogStash::Event.new({ "host" => "bar", "message" => "foo"}) }
-
-    it "should cache only one template" do
-      Java::OrgLogstash::StringInterpolation.clear_cache
-      expect {
-        event1.to_s
-        event2.to_s
-      }.to change { Java::OrgLogstash::StringInterpolation.cache_size }.by(1)
-    end
 
     it "return the string containing the timestamp, the host and the message" do
       expect(event1.to_s).to eq("#{timestamp.to_iso8601} #{event1.get("host")} #{event1.get("message")}")
--- a/logstash-core/src/test/java/org/logstash/EventTest.java
+++ b/logstash-core/src/test/java/org/logstash/EventTest.java
@@ -285,6 +285,28 @@ public class EventTest {
         assertEquals(tags.size(), 2);
         assertEquals(tags.get(0), "foo");
         assertEquals(tags.get(1), "bar");
-      }
+    }
 
+    @Test
+    public void toStringwithTimestamp() throws Exception {
+        Map data = new HashMap();
+        data.put("host", "foo");
+        data.put("message", "bar");
+        Event e = new Event(data);
+        assertEquals(e.toString(), e.getTimestamp().toIso8601() + " foo bar");
+    }
+
+    @Test
+    public void toStringwithoutTimestamp() throws Exception {
+        Map data = new HashMap();
+        data.put("host", "foo");
+        data.put("message", "bar");
+        Event e = new Event(data);
+        e.remove("@timestamp");
+        assertEquals(e.toString(), "foo bar");
+
+        e = new Event();
+        e.remove("@timestamp");
+        assertEquals(e.toString(), "%{host} %{message}");
+    }
 }


andrewvc:
Looks like CI's broken on a failing spec here: ./logstash-core/spec/logstash/legacy_ruby_event_spec.rb:600

logstash-core/src/main/java/org/logstash/Event.java
```diff
@@ -264,16 +264,16 @@ public Event clone() throws CloneNotSupportedException {
     }
 
     public String toString() {
-        // TODO: (colin) clean this IOException handling, not sure why we bubble IOException here
-        try {
-            return (getTimestamp().toIso8601() + " " + this.sprintf("%{host} %{message}"));
-        } catch (IOException e) {
-            String host = (String)this.data.get("host");
-            host = (host != null ? host : "%{host}");
+        Object f = this.getField("host");
```
andrewvc: Why is the variable `f` reused instead of having distinct `hostObject` and `messageObject` local variables?

logstash-core/src/main/java/org/logstash/Event.java
```diff
@@ -264,16 +264,16 @@ public Event clone() throws CloneNotSupportedException {
...
+        String host = (f != null ? f.toString() : "%{host}");
 
-            String message = (String)this.data.get("message");
-            message = (message != null ? message : "%{message}");
+        f = this.getField("message");
+        String message = (f != null ? f.toString() : "%{message}");
 
+        try {
+            // getTimestamp throws an IOException if there is no @timestamp field
+            return (getTimestamp().toIso8601() + " " + host + " " + message);
```
andrewvc: This isn't something that needs to block this PR, but why do we throw an exception rather than returning null here?

andrewvc: By 'here' I mean in `getTimestamp`

colinsurprenant: exactly. that is why I wrote this is the PR description. 

> Will also followup with an issue to review the Event.getTimestamp() behaviour.

but changing `getTimestamp()` this is out of scope for the bugfix & PR.

logstash-core/src/main/java/org/logstash/Event.java
colinsurprenant: it could, both are ok IMO. I decided to reuse it because it is temporary in nature just to set the `host` and `message` variables.

colinsurprenant:
@andrewvc ok - looking at the CI failure, thanks.

logstash-core/src/main/java/org/logstash/Event.java
andrewvc: Sure, I'm glad to defer it rather than starting that discussion here

logstash-core/src/main/java/org/logstash/Event.java
andrewvc: For clarity I would really urge against variable reassignment. Obviously they both work, but mutating variables outside of control structures like loops usually implies that the thing behind that something is changing, which in this case it is not, we are storing two different things in the same label. It's not a pattern I see elsewhere in Logstash. This argument has been made well here: http://wiki.c2.com/?AssignVariablesOnce .  WDYT?

colinsurprenant:
@andrewvc decided to simply remove that spec because it was testing on the usage of `sprintf` in `Event#to_s` which I removed. Using `sprintf` was superfluous and IIRC was probably to mimic what the legacy Ruby `Event#to_s` was doing when originally porting to Java.

logstash-core/src/main/java/org/logstash/Event.java
colinsurprenant: It think given the very narrow scope & simplicity of the 4 lines involved in sharing this temporary variable it does not really matter unless I am missing a condition where this could cause a problem in this context? Is there a condition where this code could not be correct?

But since this seems to be causing difficulties in assessing this bit of code I'll go ahead and  change it to avoid any other potential confusion.

logstash-core/src/main/java/org/logstash/Event.java
colinsurprenant: in #7613

colinsurprenant:
#green

andrewvc:
Great! LGTM

colinsurprenant:
merged into master, 5.x and 5.5.

PR #7609, opened 2017-07-07
MINOR: Clean up HotThreadsMonitor

original-brownbear:
Trivial cleanup:

* Made Logger `static`
* Made constant `Collection` of valid order keys `static final` (and used a set to make the lookup faster)
* Made everything that was `static` as a result explicitly `static`


original-brownbear:
@robbavey thanks for taking a look! :)

Issue #7608, opened 2017-07-07
hot_threads API broken on 5.5.0


pemontto:
The hot_threads API seems to be broken on the latest 5.5.0 release.

- Version: 5.5.0
- Operating System: Centos7
- Config File (if you have sensitive info, please remove it): All
- Steps to Reproduce: `curl localhost:9603/_node/hot_threads`

```
{
  "status": 500,
  "request_method": "GET",
  "path_info": "/_node/hot_threads",
  "query_string": "pretty",
  "http_version": "HTTP/1.1",
  "http_accept": "*/*",
  "error": "Unexpected Internal Error",
  "class": "Java::JavaLang::IllegalStateException",
  "message": "Duplicate key {cpu.time=10462123727, waited.time=-1, thread.state=timed_waiting, waited.count=9692, thread.name=[main]<redis, blocked.count=92, thread.stacktrace=[java.lang.Object.wait(Native Method), org.jruby.RubyThread.sleep(RubyThread.java:1002), org.jruby.RubyKernel.sleep(RubyKernel.java:803)], blocked.time=-1}",
  "backtrace": [
    "java.util.stream.Collectors.lambda$throwingMerger$0(java/util/stream/Collectors.java:133)",
    "java.util.HashMap.merge(java/util/HashMap.java:1253)",
    "java.util.stream.Collectors.lambda$toMap$58(java/util/stream/Collectors.java:1320)",
    "java.util.stream.ReduceOps$3ReducingSink.accept(java/util/stream/ReduceOps.java:169)",
    "java.util.ArrayList$ArrayListSpliterator.forEachRemaining(java/util/ArrayList.java:1374)",
    "java.util.stream.AbstractPipeline.copyInto(java/util/stream/AbstractPipeline.java:481)",
    "java.util.stream.AbstractPipeline.wrapAndCopyInto(java/util/stream/AbstractPipeline.java:471)",
    "java.util.stream.ReduceOps$ReduceOp.evaluateSequential(java/util/stream/ReduceOps.java:708)",
    "java.util.stream.AbstractPipeline.evaluate(java/util/stream/AbstractPipeline.java:234)",
    "java.util.stream.ReferencePipeline.collect(java/util/stream/ReferencePipeline.java:499)",
    "org.logstash.instrument.reports.ThreadsReport.generate(org/logstash/instrument/reports/ThreadsReport.java:29)",
    "java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:498)",
[...]

PR #7617, opened 2017-07-07
Fix bug where hot threads did not work with duplicate names / add threadId

andrewvc:
This fix changes the way we key threads internally to use the unique threadId vs the thread name which has no uniqueness guarantees (and indeed is often not unique as in the case of multiple inputs of the same type).

Fixes https://github.com/elastic/logstash/issues/7608

This is WIP at the moment as its lacking in tests, which I'll add monday next week.

### DIFF
--- a/logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
+++ b/logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
@@ -17,7 +17,12 @@ class HotThreadsReport
     report << '=' * STRING_SEPARATOR_LENGTH
     report << "\n"
     hash[:threads].each do |thread|
-      thread_report = "#{I18n.t("logstash.web_api.hot_threads.thread_title", :percent_of_cpu_time => thread[:percent_of_cpu_time], :thread_state => thread[:state], :thread_name => thread[:name])} \n"
+      line_str = I18n.t("logstash.web_api.hot_threads.thread_title", 
+        :percent_of_cpu_time => thread[:percent_of_cpu_time], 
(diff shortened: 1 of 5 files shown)


logstash-core/lib/logstash/api/commands/hot_threads_reporter.rb
```diff
@@ -34,6 +34,7 @@ def to_hash
     @thread_dump.each do |thread_name, _hash|
       thread_name, thread_path = _hash["thread.name"].split(": ")
       thread = { :name => thread_name,
+                 :thread_id => _hash["thread.id"],
```
robbavey: Should this be added to the to_s method above too?

andrewvc:
@robbavey I've replaced the usage of `Map` with `List` instead since there's no place where we use really need a keyed map.

I'm on the fence here about adding tests. The refactor here has removed this class of failure from occurring. WDYT about adding tests to this patch? I think we're decently covered by the [integration tests](https://github.com/elastic/logstash/blob/master/logstash-core/spec/logstash/api/commands/stats_spec.rb#L24) against the `hot_threads` endpoint.

I don't think that writing a `it should handle threads with duplicate names` spec adds much value here IMHO though I'm open to a counter-argument here.

robbavey:
@andrewvc I think I came to a similar conclusion around the tests when I was in this area of the code - the integration tests can handle it pretty reasonably. The only counter-argument that I could think of against it, is that having a test acts as a warning to future maintainers of this area of code that duplicate thread names are a real issue that can come up, and need to be handled.

PR #7623, opened 2017-07-09
PERFORMANCE: Remove needless locking on batch reads

original-brownbear:
The locking on `read_next` serves no purpose as far as I can tell.
It only guards this block:

```rb
      def read_next
        @size.times do |t|
          event = @queue.poll(@wait)
          return if event.nil? # queue poll timed out

          @originals[event] = true
        end
      end
```

The instance variables here are not shared with other instances since the batch object is created in the same thread as the `read_next` is called in (except for `@queue` which is thread-safe anyways).

=> ~10% speedup in the Apache logs case.

* also removed the needless package qualifier on the `SynchronousQueue` (it's imported via `java_import` anyways).

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -6,7 +6,7 @@ module LogStash; module Util
     java_import java.util.concurrent.TimeUnit
 
     def initialize
-      @queue = java.util.concurrent.SynchronousQueue.new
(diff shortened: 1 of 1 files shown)


jordansissel:
https://github.com/elastic/logstash/pull/7691/ supersedes this, right?

original-brownbear:
@jordansissel depends, even if we're going with #7691 I'd probably still merge this in separately since it's such a clear fix without changing any other dynamics?
No strong opinion on this though, I'm fine with closing here and continuing in #7691 too :)

PR #7624, opened 2017-07-09
MINOR: Remove needless singleton use in pathcache

original-brownbear:
Just doesn't make sense to have a singleton with lazy initialization here when this thing is used on ever set to an `Event`. Saves some lines of code and a few function calls per `set`.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/PathCache.java
+++ b/logstash-core/src/main/java/org/logstash/PathCache.java
@@ -2,30 +2,21 @@ package org.logstash;
 
 import java.util.concurrent.ConcurrentHashMap;
 
-public class PathCache {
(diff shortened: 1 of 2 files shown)


jakelandis:
LGTM as is.... 

+1 on the the final's 

If we are cleaning up too, I would suggest:
* A private scoped constructor (for any 'utility' class)
* Class level java doc (for all classes)
* Method level java doc (for all classes)

original-brownbear:
@jakelandis thanks for taking a look! and +1 on each of those suggestions, especially stuff like this class really should have some explanation to it ...

original-brownbear:
gnarf needs manual backport to `5.x`

PR #7625, opened 2017-07-09
#7624 5.x Backport

original-brownbear:
backport of #7624 
Needed to port a few more than just these `static` calls from `master` to make it work, but no changes needed that aren't in `master`.


original-brownbear:
Jenkins retest this please

PR #7626, opened 2017-07-09
MINOR: Cleanup redundant casts + fix bug resulting from one of them

original-brownbear:
With the exception of the change to `Event` just standard cleanup of either overly precise casting or completely redundant casting.

The change to `Event` fixes an actual bug though!
* The return of `CBOR_MAPPER.readValue(source, HashMap.class)` will always be a `Map` or `null` => the current implementation would throw `NPE` when running into the else block's `o.getClass().getName()`


jakelandis:
LGTM

original-brownbear:
@jakelandis thanks!

original-brownbear:
needs backport

PR #7628, opened 2017-07-10
bump jruby to 9.1.12.0

jsvd:
jruby 9.1.12.0, but specially .11.0, brings many bug fixes.
~~Also there is a small correction that must be made to the BAT script for windows testing, since the directory of the jruby once unzipped is "jruby-bin-version", not "jruby-version". the BAT script now does the necessary replacement.~~


jakelandis:
LGTM

jakelandis:
What was wrong with the JRUBYVERSION path change ?

jsvd:
we decided to handle this in the infra part of the ci architecture, which will unpack jruby to a folder named "jruby-<version>" so this way we can directly match it against the contents of the `.ruby-version` file

PR #7629, opened 2017-07-10
Benchmark Tool

original-brownbear:
So far it can:

* Run Apache dataset and baseline (generator -> dots ->stdout)
* Download any release from eleastic.co and run against it
* Run against local LS path
* Download build and run any Github revision/tag/branch
* No dependencies except for Java needed to run this

Output looks like this for repeat runs of `5.5.0` and `5.4.3` (only measures throughput in 1s intervals so far):

screen shot 2017-07-10 at 17 11 43

### DIFF
--- a/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LsMetricsMonitor.java
+++ b/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LsMetricsMonitor.java
@@ -1,0 +1,108 @@
+package org.logstash.benchmark.cli;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.EnumMap;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.logstash.benchmark.cli.ui.LsMetricStats;
+import org.openjdk.jmh.util.ListStatistics;
+
+public final class LsMetricsMonitor implements Callable<EnumMap<LsMetricStats, ListStatistics>> {
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    private final String metrics;
+
+    private volatile boolean running = true;
+
+    public LsMetricsMonitor(final String metrics) {
+        this.metrics = metrics;
+    }
+
+    @Override
+    public EnumMap<LsMetricStats, ListStatistics> call() {
+        final ListStatistics stats = new ListStatistics();
+        long count = 0L;
+        final ListStatistics counts = new ListStatistics();
+        final ListStatistics cpu = new ListStatistics();
+        long start = System.nanoTime();
+        while (running) {
+            try {
+                TimeUnit.SECONDS.sleep(1L);
+                final long[] newcounts = getCounts();
+                final long newcount = newcounts[0];
+                if (newcount < 0L) {
+                    start = System.nanoTime();
+                    continue;
+                }
+                final long newstrt = System.nanoTime();
+                stats.addValue(
+                    (double) (newcount - count) /
+                        (double) TimeUnit.SECONDS.convert(newstrt - start, TimeUnit.NANOSECONDS)
+                );
+                start = newstrt;
+                count = newcount;
+                counts.addValue((double) count);
+                cpu.addValue(newcounts[1]);
+            } catch (final InterruptedException ex) {
+                throw new IllegalStateException(ex);
+            }
+        }
+        final EnumMap<LsMetricStats, ListStatistics> result = new EnumMap<>(LsMetricStats.class);
+        result.put(LsMetricStats.THROUGHPUT, stats);
+        result.put(LsMetricStats.COUNT, counts);
+        result.put(LsMetricStats.CPU_USAGE, cpu);
+        return result;
+    }
+
+    public void stop() {
+        running = false;
+    }
+
+    private long[] getCounts() {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try (final CloseableHttpClient client = HttpClientBuilder.create().build()) {
+            baos.reset();
+            try (final CloseableHttpResponse response = client
+                .execute(new HttpGet(metrics))) {
+                response.getEntity().writeTo(baos);
+            } catch (final IOException ex) {
+                return new long[]{-1L, -1L};
+            }
+            final Map<String, Object> data =
+                OBJECT_MAPPER.readValue(baos.toByteArray(), HashMap.class);
+            final long count;
+            if (data.containsKey("pipeline")) {
+                count = getFiltered((Map<String, Object>) data.get("pipeline"));
+
+            } else if (data.containsKey("events")) {
+                count = getFiltered(data);
+            } else {
+                count = -1L;
+            }
+            final long cpu;
+            if (count == -1L) {
+                cpu = -1L;
+            } else {
+                cpu = ((Number) ((Map<String, Object>) ((Map<String, Object>) data.get("process"))
+                    .get("cpu")).get("percent")).longValue();
+            }
+            return new long[]{count, cpu};
+        } catch (final IOException ex) {
+            throw new IllegalStateException(ex);
+        }
+    }
+
+    private static long getFiltered(final Map<String, Object> data) {
+        return ((Number) ((Map<String, Object>) (data.get("events")))
+            .get("filtered")).longValue();
+    }
+}
--- a/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
+++ b/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
@@ -1,0 +1,166 @@
+package org.logstash.benchmark.cli;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.AbstractMap;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+import joptsimple.OptionException;
+import joptsimple.OptionParser;
+import joptsimple.OptionSet;
+import joptsimple.OptionSpec;
+import joptsimple.OptionSpecBuilder;
+import org.logstash.benchmark.cli.cases.ApacheLogsComplex;
+import org.logstash.benchmark.cli.cases.Case;
+import org.logstash.benchmark.cli.cases.GeneratorToStdout;
+import org.logstash.benchmark.cli.ui.LsMetricStats;
+import org.logstash.benchmark.cli.ui.LsVersionType;
+import org.logstash.benchmark.cli.ui.UserInput;
+import org.logstash.benchmark.cli.ui.UserOutput;
+import org.logstash.benchmark.cli.util.LsBenchLsSetup;
+import org.openjdk.jmh.util.ListStatistics;
+
+/**
+ * Benchmark CLI Main Entry Point.
+ */
+public final class Main {
+
+    /**
+     * Ctor.
+     */
+    private Main() {
+        // Utility Class.
+    }
+
+    /**
+     * CLI Entrypoint.
+     * @param args Cli Args
+     */
+    public static void main(final String... args) throws Exception {
+        final OptionParser parser = new OptionParser();
+        final OptionSpecBuilder gitbuilder =
+            parser.accepts(UserInput.GIT_VERSION_PARAM, UserInput.GIT_VERSION_HELP);
+        final OptionSpecBuilder localbuilder =
+            parser.accepts(UserInput.LOCAL_VERSION_PARAM, UserInput.LOCAL_VERSION_HELP);
+        final OptionSpecBuilder distributionbuilder =
+            parser.accepts(
+                UserInput.DISTRIBUTION_VERSION_PARAM, UserInput.DISTRIBUTION_VERSION_HELP
+            );
+        final OptionSpec<String> git = gitbuilder.requiredUnless(
+            UserInput.DISTRIBUTION_VERSION_PARAM, UserInput.LOCAL_VERSION_PARAM
+        ).availableUnless(UserInput.DISTRIBUTION_VERSION_PARAM, UserInput.LOCAL_VERSION_PARAM)
+            .withRequiredArg().ofType(String.class).forHelp();
+        final OptionSpec<String> distribution = distributionbuilder
+            .requiredUnless(UserInput.LOCAL_VERSION_PARAM, UserInput.GIT_VERSION_PARAM)
+            .availableUnless(
+                UserInput.LOCAL_VERSION_PARAM, UserInput.GIT_VERSION_PARAM
+            ).withRequiredArg().ofType(String.class).forHelp();
+        final OptionSpec<String> local = localbuilder.requiredUnless(
+            UserInput.DISTRIBUTION_VERSION_PARAM, UserInput.GIT_VERSION_PARAM)
+            .availableUnless(UserInput.DISTRIBUTION_VERSION_PARAM, UserInput.GIT_VERSION_PARAM)
+            .withRequiredArg().ofType(String.class).forHelp();
+        final OptionSpec<String> testcase = parser.accepts(
+            UserInput.TEST_CASE_PARAM, UserInput.TEST_CASE_HELP
+        ).withRequiredArg().ofType(String.class).defaultsTo(GeneratorToStdout.IDENTIFIER).forHelp();
+        final OptionSpec<File> pwd = parser.accepts(
+            UserInput.WORKING_DIRECTORY_PARAM, UserInput.WORKING_DIRECTORY_HELP
+        ).withRequiredArg().ofType(File.class).defaultsTo(UserInput.WORKING_DIRECTORY_DEFAULT)
+            .forHelp();
+        final OptionSet options;
+        try {
+            options = parser.parse(args);
+        } catch (final OptionException ex) {
+            parser.printHelpOn(System.out);
+            throw ex;
+        }
+        final LsVersionType type;
+        final String version;
+        if (options.has(distribution)) {
+            type = LsVersionType.DISTRIBUTION;
+            version = options.valueOf(distribution);
+        } else if (options.has(git)) {
+            type = LsVersionType.GIT;
+            version = options.valueOf(git);
+        } else {
+            type = LsVersionType.LOCAL;
+            version = options.valueOf(local);
+        }
+        execute(
+            new UserOutput(System.out), loadSettings(), options.valueOf(testcase),
+            options.valueOf(pwd).toPath(), version, type
+        );
+    }
+
+    /**
+     * Programmatic Entrypoint.
+     * @param output Output Printer
+     * @param settings Properties
+     * @param test String identifier of the testcase to run
+     * @param cwd Working Directory to run in and write cache files to
+     * @param version Version of Logstash to benchmark
+     * @param type Type of Logstash version to benchmark
+     * @throws Exception On Failure
+     */
+    public static void execute(final UserOutput output, final Properties settings,
+        final String test, final Path cwd, final String version, final LsVersionType type)
+        throws Exception {
+        output.printBanner();
+        output.printLine();
+        output.green(String.format("Benchmarking Version: %s", version));
+        output.green(String.format("Running Test Case: %s", test));
+        output.printLine();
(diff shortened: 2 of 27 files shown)


original-brownbear:
@suyograo FYI, any input on what we want to display for stats welcome :)

Also, do we want to have this version store to ES already? (If not I think we're only some minor cleanups away from done here)

suyograo:
Great stuff! Will review this on monday :)

suyograo:
@original-brownbear I'm getting this error when I try to test it out. Am I doing anything wrong here?

```
Working in cache directory: /tmp/benchmark
Benchmarking Version: local:/Users/suyog/ws/elastic/logstash
Running Test Case: apache
Exception in thread "main" java.lang.IllegalStateException: java.io.IOException: Broken pipe
	at org.logstash.benchmark.cli.cases.ApacheLogsComplex.run(ApacheLogsComplex.java:53)
	at org.logstash.benchmark.cli.Main.execute(Main.java:106)
	at org.logstash.benchmark.cli.Main.main(Main.java:66)
Caused by: java.io.IOException: Broken pipe
	at java.io.FileOutputStream.writeBytes(Native Method)
	at java.io.FileOutputStream.write(FileOutputStream.java:326)
	at java.io.BufferedOutputStream.write(BufferedOutputStream.java:122)
	at org.apache.commons.io.IOUtils.copyLarge(IOUtils.java:2147)
	at org.apache.commons.io.IOUtils.copy(IOUtils.java:2102)
	at org.logstash.benchmark.cli.LogstashInstallation$FromLocalPath.execute(LogstashInstallation.java:136)
	at org.logstash.benchmark.cli.cases.ApacheLogsComplex.run(ApacheLogsComplex.java:49)
	... 2 more
```

suyograo:
I used:

```
suyog at Suyogs-MacBook-Pro-2 in ~/ws/elastic/logstash/tools/benchmark-cli/build/libs on pr/7629 [$]
$ java -cp 'benchmark-cli:*' org.logstash.benchmark.cli.Main --testcase apache --workdir /tmp/benchmark --version local:/Users/suyog/ws/elastic/logstash
```

original-brownbear:
@suyograo sec looking into it :)

original-brownbear:
@suyograo the "baseline" test case works right?

suyograo:
@original-brownbear yes `baseline` works. Does `apache` need ES to be running?

original-brownbear:
@suyograo no and I just tried it, works fine for me. It's dying on the streaming to the logstash process's stdin.

Maybe LS die for some reason? Can you see anything in your local ls logs?

original-brownbear:
(not to say that this case shouldn't be handled with a better error message :))

original-brownbear:
@suyograo is that local installation properly built already? (`rake bootstrap && rake test:install-core` suffices for sure) For `local:/` it doesn't run the build and just takes the directory as is.

original-brownbear:
@suyograo just reran it, worked fine on my box:

```sh
libs git:(benchmark-tool) java -cp 'benchmark-cli.jar:*' org.logstash.benchmark.cli.Main --workdir=/tmp/benchmark1 --testcase=apache --version=5.4.3 
██╗      ██████╗  ██████╗ ███████╗████████╗ █████╗ ███████╗██╗  ██╗          
██║     ██╔═══██╗██╔════╝ ██╔════╝╚══██╔══╝██╔══██╗██╔════╝██║  ██║          
██║     ██║   ██║██║  ███╗███████╗   ██║   ███████║███████╗███████║          
██║     ██║   ██║██║   ██║╚════██║   ██║   ██╔══██║╚════██║██╔══██║          
███████╗╚██████╔╝╚██████╔╝███████║   ██║   ██║  ██║███████║██║  ██║          
╚══════╝ ╚═════╝  ╚═════╝ ╚══════╝   ╚═╝   ╚═╝  ╚═╝╚══════╝╚═╝  ╚═╝          
                                                                             
██████╗ ███████╗███╗   ██╗ ██████╗██╗  ██╗███╗   ███╗ █████╗ ██████╗ ██╗  ██╗
██╔══██╗██╔════╝████╗  ██║██╔════╝██║  ██║████╗ ████║██╔══██╗██╔══██╗██║ ██╔╝
██████╔╝█████╗  ██╔██╗ ██║██║     ███████║██╔████╔██║███████║██████╔╝█████╔╝ 
██╔══██╗██╔══╝  ██║╚██╗██║██║     ██╔══██║██║╚██╔╝██║██╔══██║██╔══██╗██╔═██╗ 
██████╔╝███████╗██║ ╚████║╚██████╗██║  ██║██║ ╚═╝ ██║██║  ██║██║  ██║██║  ██╗
╚═════╝ ╚══════╝╚═╝  ╚═══╝ ╚═════╝╚═╝  ╚═╝╚═╝     ╚═╝╚═╝  ╚═╝╚═╝  ╚═╝╚═╝  ╚═╝
                                                                             
Working in cache directory: /tmp/benchmark1
Benchmarking Version: 5.4.3
Running Test Case: apache
StatisticalSummaryValues:
n: 754
min: 250.0
max: 12000.0
mean: 9138.179708222811
std dev: 2394.4505947287785
variance: 5733393.650597001
sum: 6890187.5

➜  libs git:(benchmark-tool) 
```

original-brownbear:
only difference I could make out was that I had rebased locally a few minutes ago, pushed the rebased version, but it shouldn't make a difference, benchmark code is isolated.

suyograo:
I may not have built the local one, will do so. 5.4.0 works fine!

original-brownbear:
@suyograo I can just make it run the installer over a local one as well if you want? I consciously didn't do that, because that would make testing against local plugin versions a little annoying potentially (when the rake tasks start messing with a manually adjusted `Gemfile`).

suyograo:
@original-brownbear nah, that's too much to build LS. we should assume one is already built.

suyograo:
> @suyograo FYI, any input on what we want to display for stats welcome :)

@original-brownbear how about the following as the output?

```
Benchmarking Version: 5.3.0
Running Test Case: baseline
-------------------------------------------
StatisticalSummaryValues:

Start Time: Tue Jul 18 17:07:30 PDT 2017
Elapsed Time: 2634 seconds
Num Events: 50000
Min Throughput: 3525.0
Max Throughput: 118608.0
Mean Throughput: 76923.07692307692
Std Dev: 25909.827692922296
Variance: 6.71319171076923E8
Sum: 1000000.0
Median CPU usage: 171.6 %
```

Also, what is `n` and `Sum` in your original report?

suyograo:
Can we rename `--version` to `--distribution-version`? Also, could we actually separate the sources instead of combining as one? For example, we can add `--git-hash` and `--local-path`. We can then make sure these are mutually exclusive.

Also, `--workdir` can we default to `$HOME/.logstash-benchmarks/` ?

original-brownbear:
@suyograo argument changes + output changes look good, will implement later today :)

> Also, what is n and Sum in your original report?

`n` is the number of all measurements/datapoints
`Sum` is kinda meaningless, it's the sum of all rates measured, which effectively means nothing. It's just there because of the stats printer class I reused from JMH :)

original-brownbear:
@suyograo 

There we go, now it does this

```sh
██╗      ██████╗  ██████╗ ███████╗████████╗ █████╗ ███████╗██╗  ██╗          
██║     ██╔═══██╗██╔════╝ ██╔════╝╚══██╔══╝██╔══██╗██╔════╝██║  ██║          
██║     ██║   ██║██║  ███╗███████╗   ██║   ███████║███████╗███████║          
██║     ██║   ██║██║   ██║╚════██║   ██║   ██╔══██║╚════██║██╔══██║          
███████╗╚██████╔╝╚██████╔╝███████║   ██║   ██║  ██║███████║██║  ██║          
╚══════╝ ╚═════╝  ╚═════╝ ╚══════╝   ╚═╝   ╚═╝  ╚═╝╚══════╝╚═╝  ╚═╝          
                                                                             
██████╗ ███████╗███╗   ██╗ ██████╗██╗  ██╗███╗   ███╗ █████╗ ██████╗ ██╗  ██╗
██╔══██╗██╔════╝████╗  ██║██╔════╝██║  ██║████╗ ████║██╔══██╗██╔══██╗██║ ██╔╝
██████╔╝█████╗  ██╔██╗ ██║██║     ███████║██╔████╔██║███████║██████╔╝█████╔╝ 
██╔══██╗██╔══╝  ██║╚██╗██║██║     ██╔══██║██║╚██╔╝██║██╔══██║██╔══██╗██╔═██╗ 
██████╔╝███████╗██║ ╚████║╚██████╗██║  ██║██║ ╚═╝ ██║██║  ██║██║  ██║██║  ██╗
╚═════╝ ╚══════╝╚═╝  ╚═══╝ ╚═════╝╚═╝  ╚═╝╚═╝     ╚═╝╚═╝  ╚═╝╚═╝  ╚═╝╚═╝  ╚═╝
                                                                             
Benchmarking Version: 5.5.0
Running Test Case: baseline
Start Time: Wed 7 19 17:08:14.367 2017 CEST
Statistical Summary:

Elapsed Time: 30596s
Num Events: 986226.000000
Min Throughput: 5726.000000
Max Throughput: 46500.000000
Mean Throughput: 42879.391304
StdDev(Throughput): 9016.283268
Variance(Throughput): 81293363.976285
Mean CPU Usage: 18.826087%
```

WDYT?

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LsMetricsMonitor.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.logstash.benchmark.cli.ui.LsMetricStats;
+import org.omg.CORBA.Object;
```
jakelandis: I doubt you want the CORBA Object :)

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
```diff
@@ -0,0 +1,152 @@
...
+        throws Exception {
+        output.printBanner();
+        output.printLine();
+        output.green(String.format("Benchmarking Version: %s", version));
+        output.green(String.format("Running Test Case: %s", test));
+        output.printLine();
+        Files.createDirectories(cwd);
+        final LogstashInstallation logstash = LsBenchLsSetup.setupLS(
+            cwd.toAbsolutePath().toString(), version, type,
+            JRubyInstallation.bootstrapJruby(cwd, JRubyInstallation.JRUBY_DEFAULT_VERSION)
```
jakelandis: Probably don't want to install JRuby for LsVersionType.DISTRIBUTION

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
```diff
@@ -0,0 +1,152 @@
...
+        );
+        final Case testcase;
+        if ("baseline".equals(test)) {
```
jakelandis: nitpick: move these constants to Case interface and use equalsIgnoreCase

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
```diff
@@ -0,0 +1,152 @@
...
+        output.green("Statistical Summary:\n");
+        final long start = System.currentTimeMillis();
+        final EnumMap<LsMetricStats, ListStatistics> stats = testcase.run();
+        output.green(String.format(
+            "Elapsed Time: %ds",
+            TimeUnit.SECONDS.convert(
+                System.currentTimeMillis() - start, TimeUnit.MILLISECONDS
+            )
+        ));
+        output.green(String.format("Num Events: %f", stats.get(LsMetricStats.COUNT).getMax()));
```
jakelandis: For any whole numbers, can format with %d, and limit the precision to 2 places on decimal types ?  (imo it makes it easier to read)

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
```diff
@@ -0,0 +1,152 @@
...
+        final Case testcase;
+        if ("baseline".equals(test)) {
+            testcase = new GeneratorToStdout(logstash);
+        } else if ("apache".equals(test)) {
+            testcase = new ApacheLogsComplex(logstash, cwd, settings);
+        } else {
+            throw new IllegalArgumentException(String.format("Unknown test case %s", test));
+        }
+        output.printStartTime();
+        output.green("Statistical Summary:\n");
```
jakelandis: Can you move this output line down until after the test is run ?  (the first time i ran it, I thought I was missing the results, but the test was just running).

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/ui/UserOutput.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public void printStartTime() {
+        green(String.format("Start Time: %s", DATE_TIME_FORMATTER.format(ZonedDateTime.now())));
+    }
+
+    public void printLine() {
+        green("------------------------------------------");
+    }
+
+    public void printBanner() {
```
jakelandis: I ❤️ this !

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/util/LsBenchFileUtil.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import org.apache.commons.io.FileUtils;
+
+public final class LsBenchFileUtil {
+
+    public static void ensureDeleted(final File file) throws IOException {
+        if (file.exists()) {
+            if (file.isDirectory()) {
```
jakelandis: May want to add ``` file.deleteOnExit();``` and swallow any errors here in favor of a log message.

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
```diff
@@ -0,0 +1,152 @@
...
+                System.currentTimeMillis() - start, TimeUnit.MILLISECONDS
+            )
+        ));
+        output.green(String.format("Num Events: %f", stats.get(LsMetricStats.COUNT).getMax()));
+        final ListStatistics throughput = stats.get(LsMetricStats.THROUGHPUT);
+        output.green(String.format("Min Throughput: %f", throughput.getMin()));
+        output.green(String.format("Max Throughput: %f", throughput.getMax()));
+        output.green(String.format("Mean Throughput: %f", throughput.getMean()));
+        output.green(String.format("StdDev(Throughput): %f", throughput.getStandardDeviation()));
+        output.green(String.format("Variance(Throughput): %f", throughput.getVariance()));
```
jakelandis: more nitpicking: Switch from ```X Throughput``` to ```Throughput X``` and remove the parentheses for StdDev and Variance.

jakelandis:
First and Foremost - Awesome work @original-brownbear ! 

I love that you can run this against released versions, or git commits from any github repo! Also your implementation using a mix of jmh backed stats with Logstash metrics api is pretty slick. 

All of my comments trend toward nitpicking or very minor stuff, in general the code is very good and easy to read. 

Here are some comments that don't fit well into the line by line review.

Readme:
Can I haz one ? (with usage instructions)

Options screen:
Can you add a description and example for ```--distribution-version``` and ```--local-path``` ?
I assume ```--git-hash``` and ```--distribution-version``` are mutually exclusive (if so we may want to enforce that and update the description)?
nitpick: each option description to have a \ndefault: [value|none] (last line of the description). 

General Code Style:
_At the risk of bikeshedding..._  
Can we apply a rule that anything public (or non-public and tricky) interface, class, method, or enum should have javadoc ? 
Any class with only static methods should have a private constructor.

Enhancements:
These are endless ... but here are my top 4:
* Add in some hardware stats to the output (Platform/CPU/Memory, etc) , to help contextualize the result if shared between people. 
* Add a backgroud thread to add progress dots to let you know it is working, the apache test case takes a while and some feed back helps (kinda like M$'s progress bar, it means nothing, but makes you feel better). 
* Add in an option for --testcase=custom and expose a --conifg-dir option such that, assuming I can figure out how to feed data in, I can use this for any test case.
* Add an option to store a csv file of the results, so that they can be source controlled, and graphed over time. 

Meta:
Can we consolidate all the benchmark code under something like ```/benchmarks``` ?  ```logstash-core/benchmarks```, ```tools/benchmark```, and ```tools/benchmark-cli``` (The current 3 places for benchmarks) is bit confusing. Perhaps ```/benchmarks/queue```, ```benchmarks/event_accessor```, ```benchmark/logstash``` (this project), etc.  With each directory having a README with instructions to run ?  I ask because as a developer, i would like to have easy access to run any of the benchmarks before and after my changes to help ensure I did/didn't impact performance, as well it established a pattern for how to add to the benchmarks. I can log a separate issue if you like. 

Also, I wonder if there any overlap between this tool and https://github.com/elastic/sizing-benchmarks/issues/3 or https://github.com/elastic/sizing-benchmarks/issues/2

original-brownbear:
@jakelandis thanks for the great and thorough review! Will have this cleaned up for Monday, just from a quick read I think I agree with all but one trivial point raised :)

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LsMetricsMonitor.java
original-brownbear: @jakelandis hah good catch, fixed :)

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
original-brownbear: @jakelandis right! :) huh this should make things a lot faster, fixing :)

original-brownbear: Fixed, haha this saves like 30s to 1 min on every run, you're the best :)

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
original-brownbear: Yup, decimals are dumb here :) fixing

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
original-brownbear: Jup this makes sense :) man we should have some progress bar ... not sure how to do this without `libcurses` but that could be in version `0.2` :P

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/Main.java
original-brownbear: jup this looks much nicer :) fixing

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/util/LsBenchFileUtil.java
original-brownbear: @jakelandis this I think is my only point of disagreement :D, depending on what we're gonna build on top of things here, clean FS access with the permissions to delete is paramount.
If we fail here, we should just die otherwise we're just making debugging harder without any benefit?

original-brownbear:
@jakelandis addressed all but one of the concrete points (I think/hope :)). Maybe we should move further improvements to the next issue(s) to get some parallelism into this?

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/util/LsBenchFileUtil.java
jakelandis: @original-brownbear - Fair points. 👍

original-brownbear:
@jakelandis wops sorry for pushing to this while "we were talking", just added a few more cleanups (mainly just redundant code removal + a few comments) but I think this is pretty much the best I can do without investing a ton more time.

jakelandis:
@original-brownbear - np, clean away. I don't think that minor clean ups need review, especially on new, non-core workflows. 

I agree this _(more then)_ complete enough to ship. I will pull some of the enhancements raised out to their own issue.  

LGTM :shipit:

original-brownbear:
@jakelandis 

thanks :)

>  I will pull some of the enhancements raised out to their own issue.

+1 I like those :)

original-brownbear:
@suyograo are you happy with this for version `0` as well? :)

jakelandis:
Enhancements - Round 1 : https://github.com/elastic/logstash/issues/7784

suyograo:
LGTM

suyograo:
I'm also interested in consolidating all these benchmarking folders as @jakelandis mentioned. I'm even open to moving this particular tool to a separate repo, but that can come later.

Nice work, @original-brownbear and thanks for the thorough review @jakelandis

PR #7641, opened 2017-07-11
Minor: Avoid redundant Java to Ruby conversions

original-brownbear:
This cleans up one redundant set of Ruby to Java conversions, which shows with a visible performance gain on my test machine (~5-10% for me).
The more complex your events, the larger the gain most likely.

The reason being that we are setting up events with this code:

```java
            } else if (data instanceof RubyHash) {
                this.event = new Event(ConvertedMap.newFromRubyHash((RubyHash) data));
            } else if (data instanceof MapJavaProxy) {
                this.event = new Event(ConvertedMap.newFromMap(
                    (Map)((MapJavaProxy)data).getObject())
                );
            } else {
```

This already creates a `ConvertedMap` that is a copy of the input data. In the current code we run these maps that we know to be converted already through the converter again, creating a new copy of it needlessly.
Also it can't hurt to directly call `ConvertedMap.newFromMap(data)` when we know it's a `Map` instead of taking the detour of the `Valuefier` which has no typed interface.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Event.java
+++ b/logstash-core/src/main/java/org/logstash/Event.java
@@ -56,10 +56,23 @@ public class Event implements Cloneable, Serializable, Queueable {
         this.metadata_accessors = new Accessors(this.metadata);
     }
 
-    public Event(Map data)
-    {
-        this.data = (Map<String, Object>)Valuefier.convert(data);
+    /**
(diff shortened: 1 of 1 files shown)


jakelandis:
LGTM - Nice Find!

original-brownbear:
@jakelandis thanks :)!

PR #7648, opened 2017-07-11
Fix typo in Kibana Dashboards import

robbavey:
Fix typo in body definition that causes x-pack-logstash local builds to fail


robbavey:
A bad config on my end was causing issues. Closing

PR #7649, opened 2017-07-11
MINOR: Remove redundant instantiation of utility class

original-brownbear:
Trivial but we shouldn't not fix it :)


jakelandis:
LGTM

PR #7655, opened 2017-07-11
MINOR: Remove redundant wrapping and generics

original-brownbear:
Pretty trivial cleanup if you think it about it:

* Our wrapping of `Hashmap` is only used to identify `Hashmap` that hold values that we ran through the `Valuefier`.
  * We can save at least one object (and level of indirection) for every one of these maps and lists by not wrapping but extending
  * We don't need these to be generic, we always just use `<Object>` or `<String, Object>` respectively anyways => much safer to enforce this type without generics

=> slight performance gain and memory saving (YK shows avg. GC pause going from 42ms to 36ms with this change, running the Apache dataset)
=> 300 lines of redundant code saved

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Cloner.java
+++ b/logstash-core/src/main/java/org/logstash/Cloner.java
@@ -24,7 +24,7 @@ public final class Cloner {
             clone = new LinkedList<>();
         } else if (list instanceof ArrayList<?>) {
             clone = new ArrayList<>();
-        } else if (list instanceof ConvertedList<?>) {
+        } else if (list instanceof ConvertedList) {
             clone = new ArrayList<>();
         } else {
             throw new ClassCastException("unexpected List type " + list.getClass());
@@ -45,7 +45,7 @@ public final class Cloner {
             clone = new TreeMap<>();
         } else if (map instanceof HashMap<?, ?>) {
             clone = new HashMap<>();
-        } else if (map instanceof ConvertedMap<?, ?>) {
+        } else if (map instanceof ConvertedMap) {
             clone = new HashMap<>();
         } else {
             throw new ClassCastException("unexpected Map type " + map.getClass());
--- a/logstash-core/src/main/java/org/logstash/ConvertedList.java
+++ b/logstash-core/src/main/java/org/logstash/ConvertedList.java
@@ -1,31 +1,20 @@
 package org.logstash;
 
+import java.util.ArrayList;
+import java.util.List;
 import org.jruby.RubyArray;
 import org.jruby.runtime.builtin.IRubyObject;
 
-import java.util.ArrayList;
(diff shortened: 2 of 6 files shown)


original-brownbear:
@jakelandis 

> (or were the generics not really used ?)

Jup exactly, they weren't => dropping them makes this clean + more importantly should allow for a follow-up PR with some serious optimizations (I have a plan :P)

> Would it be possible to add method level generics to the static methods to preserve/gain generic type checking for the conversions?

We didn't do this before either, this was always just `<String, Object>`. This is kind of the overarching problem here though, in actuality the type here is `<String,WhateverWeConsiderConvertedLikeABiValueOrJustAnotherMapOrList>`, but we don't have a general interface covering the value type -> we express this via the map type :( ... long story short: Better to work towards removing this type/class since it doesn't make sense to use subclassing `HashMap` to express the value type. We can't generify anything here for now since the values will always be of mixed type, so `Object` is all they have in common :(

> Also, it appears that these classes are missing unit tests (prior), could you add some? With so many lines of code changing, i would feel better if the unit tests passed before and after the changes here.

We these classes are the basis of every `Event` and covered by the `Event` tests in every way imaginable + with this change, it would literally be the same as writing tests for `HashMap` or `ArrayList` => don't think we need any additional tests here right?

jakelandis:
@original-brownbear - thanks for the explanations. 

My comments come for a more general place then this particular PR _more types, more tests, more doc_ will be my prevailing comments if I pickup the review. 

In this case I understand your rationale, and agree. 

I was hoping for something like:
```
public final class ConvertedList<I> extends ArrayList<I> {
    public <T extends List<I>> ConvertedList(T list) {
        super(list.size());
        for (I item : list) {
            super.add((I) convert(item));
        }
    }
```
preferably without the explicit cast on convert(). Something like this would preserve the original type, and ensure that code of the wrong type won't compile.
For example:
```
    ArrayList<String> a = new ArrayList<>();
        ConvertedList<String> b = new ConvertedList<>(a);
        //Won't compile
        ConvertedList<Integer> c = new ConvertedList<>(a);
```

...but looking further into the code everything gets up-casted to an Object, and then checked with instance of's .... and this is pretty deep rabbit hole. 

Also, I agree the unit tests here wouldn't provide much functional value.  I personally tend to lean towards testing trivial stuff as a statement to future developers of the intent and easy points to hook into for future developers to extend if needed. However, that is just personal style and not suggesting it better/worse then others, just abit of my personal philosophy towards testing the trivial.  

Oh yeah.. LGTM , much cleaner to prior, and +1 to any changes that help with stronger typing if you go down that rabbit hole.

original-brownbear:
@jakelandis thanks!

original-brownbear:
gnarf needs backport :)

PR #7663, opened 2017-07-12
#7662 Make StringBiValue safe against mutations to the underlying RubyString

### LINKED ISSUE #7662: Mutating String Fields Breaks Consistency between Java and Ruby Event APIs
The current implementation of `StringBiValue` is broken for the case of mutating the underlying `RubyString`.

Meaning that the following spec fails in `master`:

```rb

    it "should propagate changes to mutable strings to java APIs" do
      e = LogStash::Event.new()
      e.to_java.setField("foo", "bar")
      expect(e.get("foo")).to eq("bar")
      e.get("foo").gsub!(/bar/, 'pff')
      expect(e.get("foo")).to eq("pff")
      expect(e.to_java.getField("foo")).to eq("pff")
    end
```

The replacement `"pff"` is only visible via Ruby, not via Java APIs. Any mutation of a `RubyString` underlying a `StringBiValue` results in a broken, inconsistent state.

original-brownbear:
Fixes #7662

Since `org.jruby.RubyString` is mutable we simply cannot cache its `String` representation, as much as this **seems to** hurts performance wise.
In actuality there is no performance gain anyways since we're forced to `.dup` strings whenever we have shared Java and Ruby logic like in the UA filter:

https://github.com/logstash-plugins/logstash-filter-useragent/blob/master/lib/logstash/filters/useragent.rb#L138

```rb
event.set(@prefixed_major, ua_version.major.dup.force_encoding(Encoding::UTF_8)) if ua_version.major
  | event.set(@prefixed_minor, ua_version.minor.dup.force_encoding(Encoding::UTF_8)) if ua_version.minor
  | event.set(@prefixed_patch, ua_version.patch.dup.force_encoding(Encoding::UTF_8)) if ua_version.patch
  | event.set(@prefixed_build, ua_version.patchMinor.dup.force_encoding(Encoding::UTF_8)) if ua_version.patchMinor
```

With this fix, we get consistency by simply computing the Java object on demand. The theoretical performance hit (in practice it seems this actually speeds up the Apache dataset, likely as a result of the `String` objects now living for a shorter time and hence being collected easier) we can completely alleviate in real life use cases, by removing the `.dup` calls from plugins like the above.

### DIFF
--- a/logstash-core/spec/logstash/event_spec.rb
+++ b/logstash-core/spec/logstash/event_spec.rb
@@ -84,6 +84,15 @@ describe LogStash::Event do
       expect(e.get("foo")).to eq("bar")
     end
 
+    it "should propagate changes to mutable strings to java APIs" do
+      e = LogStash::Event.new()
+      e.to_java.setField("foo", "bar")
+      expect(e.get("foo")).to eq("bar")
+      e.get("foo").gsub!(/bar/, 'pff')
+      expect(e.get("foo")).to eq("pff")
+      expect(e.to_java.getField("foo")).to eq("pff")
+    end
+
     it "should set deep hash values" do
       e = LogStash::Event.new()
       expect(e.set("[foo][bar]", "baz")).to eq("baz")
--- a/logstash-core/src/main/java/org/logstash/bivalues/StringBiValue.java
+++ b/logstash-core/src/main/java/org/logstash/bivalues/StringBiValue.java
@@ -1,15 +1,15 @@
 package org.logstash.bivalues;
 
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.io.ObjectStreamException;
 import org.jruby.Ruby;
 import org.jruby.RubyString;
 
-import java.io.ObjectStreamException;
-
-public class StringBiValue extends BiValueCommon<RubyString, String> implements BiValue<RubyString, String> {
+public final class StringBiValue extends BiValueCommon<RubyString, String>
+    implements BiValue<RubyString, String> {
 
     public StringBiValue(RubyString rubyValue) {
         this.rubyValue = rubyValue;
-        javaValue = null;
     }
 
     public StringBiValue(String javaValue) {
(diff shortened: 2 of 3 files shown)


jsvd:
Interesting, I wasn't aware mutating the ruby string wouldn't cause the java one to mutate too. I guess that, like you said, we were working around it doing extra .dups and .clones.

the code LGTM and I tested manually too, maybe @guyboertje can pitch in wrt to any impediment to have this merged?

original-brownbear:
@jsvd yea this is a nasty one, unlike other Ruby types, `RubyString` behaves very different from immutable `String` and more like mutable `CharBuffer`.
This spot seems to have a lot of optimization potential in general, but for now, this fixes the problem without much change :)

original-brownbear:
@guyboertje ping (just in case this was forgotten :))

guyboertje:
I have been looking for a mechanism to tell if the RubyString has been modified since it was set (from within the javaValue method) but alas there does not seem to be. Also, any form of comparison  would probably be less performant than creating a new Java String each time its requested.

Further, bear in mind that we have advised/documented that in place mutation of any Event value is not supported and that a get, modify and re-set strategy is the way to go. This is mainly to prevent invalidating the accessor path cache - e.g. in place replacement of an inner part of the tree with a leaf `where a path [a][b][c][d] exists and h = e.get("[a][b]"), then h['c'] = scalar without e.set('[a][b]', h)`. Plugin authors should know this - but Ruby filter authors may not.

original-brownbear:
@guyboertje thanks for looking through this!

> I have been looking for a mechanism to tell if the RubyString has been modified since it was set (from within the javaValue method) but alas there does not seem to be. Also, any form of comparison would probably be less performant than creating a new Java String each time its requested.

Yea sadly enough there doesn't seem to be any fast way of doing this (apart from simply offering a super fast Java API that returns the `CharSequencer` that underlies the Ruby string and mutating that ... but that may not be the best idea for maintainability :P).

original-brownbear:
(had to rebase, didn't merge clean into `master` even => waiting for green tests)

guyboertje:
@original-brownbear - Suggested on JRuby IRC, perhaps use `public int strHashCode(Ruby runtime)` on RubyString to check if value is changed - but it does require computation each time.

PR #7667, opened 2017-07-12
implement ssl in modules

jsvd:
This allows modules to connect to kibana and elasticsearch through SSL.

NOTE: this PR also changes the setting `var.output.elasticsearch.user` to `var.output.elasticsearch.username` so both username settings are consistent (and also consistent with monitoring/management username settings)

### DIFF
--- a/logstash-core/spec/logstash/modules/scaffold_spec.rb
+++ b/logstash-core/spec/logstash/modules/scaffold_spec.rb
@@ -226,7 +226,6 @@ ERB
         "var.elasticsearch.user" => "foo",
         "var.elasticsearch.password" => "password",
         "var.input.tcp.port" => 5606,
-        "dashboards.kibana_index" => ".kibana"
       }
     end
     it "puts stuff in ES" do
--- a/logstash-core/spec/logstash/modules/scaffold_spec.rb
+++ b/logstash-core/spec/logstash/modules/scaffold_spec.rb
@@ -16,9 +16,9 @@ describe LogStash::Modules::Scaffold do
   subject(:test_module) { described_class.new(mname, base_dir) }
   let(:module_settings) do
     {
-      "var.output.elasticsearch.hosts" => "es.mycloud.com:9200",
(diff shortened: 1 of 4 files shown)


suyograo:
I wonder if we need separate configs for ES and Kibana? Like so: `"var.kibana.ssl.certificate_authority` and `var.output.elasticsearch.ssl.certificate_authority`. 

I don't know if this is practical, but it would make the settings consistent.

acchen97:
+1 on having ES and Kibana configs separate

jsvd:
Will do.

Also, I should note that the modules have three sets of places where credentials are used:

1. kibana importer: for importing dashboards
2. elasticsearch importer: for importing templates (and other stuff?)
3. the elasticsearch output configuration for the modules LS pipeline

The `var.output.elasticsearch.*` settings are used in 2 and 3 and `var.kibana.*` in 1.

From this perspective, since es creds are not just used in output, I'd suggest also renaming the `var.output.elasticsearch.*` to `var.elasticsearch.*`, wdyt? /cc @guyboertje

guyboertje:
@jsvd - your name change suggestion is fine by me.

jsvd:
cool, included these changes in 1519644

logstash-core/spec/logstash/modules/scaffold_spec.rb
```diff
@@ -222,9 +222,9 @@
...
     let(:module_settings) do
       {
-        "var.output.elasticsearch.hosts" => "localhost:9200",
-        "var.output.elasticsearch.user" => "foo",
-        "var.output.elasticsearch.password" => "password",
+        "var.elasticsearch.hosts" => "localhost:9200",
+        "var.elasticsearch.user" => "foo",
+        "var.elasticsearch.password" => "password",
         "var.input.tcp.port" => 5606,
         "dashboards.kibana_index" => ".kibana"
```
guyboertje: Only if you make other changes - this line can be removed now, sorry I overlooked it.

PR #7669, opened 2017-07-12
Update default plugins list

suyograo:
Based on google analytics from docs pageviews, we are updating the bundled
plugins to remove plugins which doesn't have to be in the default package.

They can still be installed via bin/logstash-plugin


acchen97:
LGTM

acchen97:
@suyograo chatting with Rob today, we still need to add translate filter to the default plugin list. Supposedly, its still not there though I thought we had added it in the past.

PR #7672, opened 2017-07-13
Fix existing examples and add autditd example

dedemorton:
To generate the auditd config, I used the ingest converter.  The original ingest pipeline had some painless defined in it...I'm assuming the converter + Logstash takes care of whatever the painless does because *most* of the visualizations rendered OK. There were a couple of visualizations that didn not show data, but I noticed that we're not even parsing those fields in the LS config. So....I will test the ingest pipelines and open a separate issue/PR if we need to update the content further.

### DIFF
--- a/docs/static/filebeat-modules.asciidoc
+++ b/docs/static/filebeat-modules.asciidoc
@@ -113,6 +113,7 @@ Logstash pipelines that parse:
 * <<parsing-mysql>>
 * <<parsing-nginx>>
 * <<parsing-system>>
+* <<parsing-auditd>>
 
 Of course, the paths that you specify in the Filebeat config depend on the location
 of the logs you are harvesting. The examples show common default locations.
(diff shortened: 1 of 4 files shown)


dedemorton:
@suyograo I'm also wondering if I should mention the ingest converter in this topic.

suyograo:
>  I'm also wondering if I should mention the ingest converter in this topic.

@dedemorton I think we need separate docs for ingest converter. Once we have a separate topic for this, we can come back and update these pages.

suyograo:
@dedemorton Also, we'll need to take the painless scripts and convert it into ruby filter equivalent. There is definitely some transformations happening there that we need to carry to LS. I'll take a look later today.

dedemorton:
@suyograo We still need to tweak the config examples here that contain painless.

suyograo:
@jsvd can you help @dedemorton here? The only blocking aspect here is to convert the painless script to ingest pipeline.

suyograo:
I meant convert painless script to ruby filter there ^^ :)

dedemorton:
Closing this in favor of https://github.com/elastic/logstash/pull/8490. We do still need to add examples for all the modules.

PR #7674, opened 2017-07-13
MINOR: Enable javac warnings

original-brownbear:
No point in not having them show globally.
We have quite a few issues with (generic-)type safety that we should track and clean up step by step, not hide from the output.


jsvd:
LGTM

PR #7675, opened 2017-07-13
MINOR: Fix incorrect use of blocking queue for the pipeline's signal queue

original-brownbear:
This line is not cool:

```rb
signal = @signal_queue.empty? ? NO_SIGNAL : @signal_queue.pop
```

we are doing a non-atomic operation here, and could easily run into a blocking `@signal_queue.pop` call during parallel access to the `@signal_queue`. Instead of trying to roll our own `poll(0)` which the Ruby `Queue` unfortunately doesn't offer (it throws when you do a non-blocking poll on an empty queue ...) we can just use the linked blocking queue that that it uses under the hood and get the same behaviour, with the guarantee to never block here.

* This might improve performance in some cases (should be faster in any case, but more visible the more contention you have on the queue) + fixes correctness.

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -205,7 +205,7 @@ module LogStash; class Pipeline < BasePipeline
 
     @input_queue_client = @queue.write_client
     @filter_queue_client = @queue.read_client
-    @signal_queue = Queue.new
+    @signal_queue = java.util.concurrent.LinkedBlockingQueue.new
     # Note that @inflight_batches as a central mechanism for tracking inflight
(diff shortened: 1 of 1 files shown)


jakelandis:
@original-brownbear - nice find!  LGTM

Also double checked the backing implementation in [JRuby is indeed a LinkedBlockingQueue](https://github.com/jruby/jruby/blob/f7746d0be94f5d9f7cd27af610f1bc5dd2622794/core/src/main/java/org/jruby/ext/thread/Queue.java#L59)

original-brownbear:
@jakelandis thanks :)

original-brownbear:
needs backport :(

PR #7678, opened 2017-07-13
#6964 migrate setting config.reload.interval to TimeValue

### LINKED ISSUE #6964: config.reload.interval need to use the `Settings::TimeValue`
The `config.reload.interval ` settings use the `Numeric` type and we should use the new `TimeValue` type, this will allow a user to express the time in a more human way by including the unit.

original-brownbear:
Closes #6964 

Trivial, though please double check ifI didn't miss a spot :)

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -78,7 +78,7 @@
 #
 # How often to check if the pipeline configuration has changed (in seconds)
 #
-# config.reload.interval: 3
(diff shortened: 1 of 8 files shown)


original-brownbear:
@suyograo ping here btw since it's urgent and we have a bit of a backlog in reviews :)

suyograo:
LGTM.

PR #7680, opened 2017-07-13
adds key and namespace getters to Java Metric classes

jsvd:
Outside code should be able to access these properties of metrics, namely in x-pack.


jakelandis:
@jsvd - LGTM thanks!

PR #7682, opened 2017-07-13
MINOR: Make Docker upgrade more stable in Travis

original-brownbear:
This should fix instances of this one https://travis-ci.org/elastic/logstash/jobs/252548207 (Travis failing to upgrade Docker).

Without a prior `apt-get update` we're at the mercy of whatever Travis `apt` has cached at the moment the test is run, this can't be stable and isn't in practice :)
Also, there is no need to remove the package before installing it again to upgrade, just running install will upgrade the package as well.


jakelandis:
LGTM, thanks for fixing this!

original-brownbear:
@jakelandis np :)

original-brownbear:
gnarf backport again :)

PR #7688, opened 2017-07-13
rakelib: List log4j2.properties as a config file in RPM/Deb

magnusbaeck:
When a file is listed as a configuration file the package manager will ask the user about what to do about local modifications when the package is upgraded and the file has changed upstream. If not listed as a configuration file any local modifications will silently be lost.


jakelandis:
@magnusbaeck - Thanks ! 

I will get it merged.

jakelandis:
Backport to 5.x : https://github.com/elastic/logstash/pull/7694

Issue #7690, opened 2017-07-14
java.util.concurrent.SynchronousQueue is not a Good Choice for the In-Memory Queue


original-brownbear:
Currently the in-memory(legacy) queue uses `java.util.concurrent.SynchronousQueue` internally.

This is a suboptimal choice when coupled with batched reads from it since it leads to either of the following two scenarios:

* Fast consumers and slow producers will see the producer thread(s) sleep more than necessary despite beeing the bottleneck
* Fast producers and slow consumers will see the consumer thread(s) sleep more than necessary

This is very visible when running the Apache dataset with `-b 1024 -w 2`:

<img width="1152" alt="screen shot 2017-07-14 at 10 03 04" src="https://user-images.githubusercontent.com/6490959/28204054-a3e2e5e8-687c-11e7-8957-7b4fcef626d5.png">

You can see the consumer threads being the obvious bottleneck here (stdin input), with the producer sleeping a lot. But still, the consumers are merely hot ~75% of the time.
This is clearly not a good scenario but entirely expected without any buffering between inputs and outputs.
A synchronous queue only makes sense in scenarios where you are expecting to most of the time have a consumer available. This is precisely not what we want, we burn most of the time in filters and outputs, having these threads blocking on the by design wastes time that could go towards processing. Worse yet, a synchronous queue requires to lock and unlock on the queue consumption for every event read from it since it doesn't have a much more efficient `drainTo` call.
We definitely want a queue with capacity here to give us `drainTo` and reduce blocking on the filter/output end. 
Any argument about this causing more events in flight could just be countered by suggesting to use a lower batch size if that is, in fact, an issue.

original-brownbear:
@suyograo @jordansissel see #7691 and this one please, I think this is a trivial and significant performance gain (especially for high worker counts).

suyograo:
pinging @andrewvc since I think he chose `SynchronousQueue` for the in-memory implementation.

andrewvc:
Looks like a nice performance gain to me and the reasoning is definitely sound. Great work @original-brownbear ! Let's move this discussion over to #7691 maybe for future comments?

original-brownbear:
fix via #7691

amitpaz:
Hello, 
Is the Execution model documentation up to date? ( as it states usage of a synchronous queue)
https://www.elastic.co/guide/en/logstash/master/execution-model.html#execution-model

jakelandis:
[...]

PR #7691, opened 2017-07-14
#7690 Fix needless waiting

original-brownbear:
For #7690:

* This also contains #7623 
* Instead of the synchronous queue, use an array blocking queue of `2 * batch_size * worker_count` size to avoid needless sleeping
* Use `drainTo` to require far less locking
* Use a `HashSet` with appropriate sizing and load parameter to back the batch
* Note how I didn't yet fix 4 specs that use some "DummyQueue" which are broken now, functionally this thing is 100% fine though as far as I can tell

## Numbers:

Apache Dataset, command: `time cat ~/Downloads/apache_access_logs | bin/logstash -b 1024 -w 2 -f ~/tmp/logstash.cfg  | pv  | wc -c`

### Master

```sh
➜  logstash git:(master) time cat ~/Downloads/apache_access_logs | bin/logstash -b 1024 -w 2 -f ~/tmp/logstash.cfg  | pv  | wc -c
6.58MiB 0:10:17 [10.9KiB/s] [                     <=>                                                                                                                                   ]
 6901231
cat ~/Downloads/apache_access_logs  0.06s user 1.01s system 0% cpu 10:17.43 total
bin/logstash -b 1024 -w 2 -f ~/tmp/logstash.cfg  1792.97s user 41.74s system 297% cpu 10:17.69 total
pv  0.98s user 6.64s system 1% cpu 10:17.69 total
wc -c  0.09s user 0.60s system 0% cpu 10:17.69 total
```

** (2.97 * 10.3 ~ 31 min) **

### This Branch

```sh
➜  logstash git:(fix-needless-waiting) time cat ~/Downloads/apache_access_logs | bin/logstash -b 1024 -w 2 -f ~/tmp/logstash.cfg  | pv  | wc -c
6.58MiB 0:09:49 [11.4KiB/s] [                                                               <=>                                                                                         ]
 6900962
cat ~/Downloads/apache_access_logs  0.08s user 1.63s system 0% cpu 9:48.80 total
bin/logstash -b 1024 -w 2 -f ~/tmp/logstash.cfg  1899.15s user 16.16s system 324% cpu 9:49.37 total
pv  1.10s user 6.99s system 1% cpu 9:49.37 total
wc -c  0.11s user 0.85s system 0% cpu 9:49.37 total
```

** (3.24 * 9.8 ~ 31min) **

=> Master takes `6%` more wall clock time and the same amount of CPU time overall.

Also visible when looking at the profiler timelines for the threads, lots more yellow for master's workers:

screen shot 2017-07-14 at 10 03 04

when compared to this branch:

screen shot 2017-07-14 at 09 47 42

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -2,12 +2,15 @@
 
 module LogStash; module Util
   class WrappedSynchronousQueue
+    java_import java.util.concurrent.ArrayBlockingQueue
     java_import java.util.concurrent.SynchronousQueue
     java_import java.util.concurrent.TimeUnit
 
-    def initialize
-      @queue = SynchronousQueue.new
+    def initialize (size)
+      @queue = ArrayBlockingQueue.new(size)
     end
+
+    attr_reader :queue
 
     # Push an object to the queue if the queue is full
     # it will block until the object can be added to the queue.
@@ -58,7 +61,7 @@ module LogStash; module Util
       end
 
       def empty?
-        true # synchronous queue is alway empty
+        @queue.queue.isEmpty
       end
 
       def set_batch_dimensions(batch_size, wait_for)
@@ -170,11 +173,11 @@ module LogStash; module Util
         @size = size
         @wait = wait
 
-        @originals = Hash.new
-
         # TODO: disabled for https://github.com/elastic/logstash/issues/6055 - will have to properly refactor
         # @cancelled = Hash.new
 
+        #Sizing HashSet to size/load_factor to ensure no rehashing
+        @originals = java.util.HashSet.new(size * 4 / 3 + 1, 0.75)
         @generated = Hash.new
         @iterating_temp = Hash.new
         @iterating = false # Atomic Boolean maybe? Although batches are not shared across threads
@@ -182,16 +185,18 @@ module LogStash; module Util
       end
 
       def read_next
-        @size.times do |t|
-          event = @queue.poll(@wait)
-          return if event.nil? # queue poll timed out
-
-          @originals[event] = true
+        read_size = @queue.queue.drainTo(@originals, @size)
+        if read_size < @size
+          (@size - read_size).times do |_|
+            e = @queue.poll(@wait)
+            return if e.nil?
+            @originals.add(e)
+          end
         end
       end
 
       def merge(event)
-        return if event.nil? || @originals.key?(event)
+        return if event.nil? || @originals.contains(event)
         # take care not to cause @generated to change during iteration
         # @iterating_temp is merged after the iteration
         if iterating?
@@ -214,9 +219,10 @@ module LogStash; module Util
 
         # below the checks for @cancelled.include?(e) have been replaced by e.cancelled?
         # TODO: for https://github.com/elastic/logstash/issues/6055 = will have to properly refactor
-        @originals.each do |e, _|
+        @originals.each do |e|
           blk.call(e) unless e.cancelled?
         end
+        
         @generated.each do |e, _|
           blk.call(e) unless e.cancelled?
         end
--- a/logstash-core/spec/logstash/queue_factory_spec.rb
+++ b/logstash-core/spec/logstash/queue_factory_spec.rb
@@ -15,7 +15,9 @@ describe LogStash::QueueFactory do
       LogStash::Setting::Numeric.new("queue.checkpoint.acks", 1024),
       LogStash::Setting::Numeric.new("queue.checkpoint.writes", 1024),
       LogStash::Setting::Numeric.new("queue.checkpoint.interval", 1000),
-      LogStash::Setting::String.new("pipeline.id", pipeline_id)
+      LogStash::Setting::String.new("pipeline.id", pipeline_id),
+      LogStash::Setting::PositiveInteger.new("pipeline.batch.size", 125),
+      LogStash::Setting::PositiveInteger.new("pipeline.workers", LogStash::Config::CpuCoreStrategy.maximum)
     ]
   end
 
@@ -72,9 +74,10 @@ describe LogStash::QueueFactory do
   context "when `queue.type` is `memory`" do
     before do
       settings.set("queue.type", "memory")
+      settings.set("pipeline.batch.size", 1024)
     end
 
-    it "returns a `WrappedAckedQueue`" do
+    it "returns a `WrappedSynchronousQueue`" do
       queue =  subject.create(settings)
       expect(queue).to be_kind_of(LogStash::Util::WrappedSynchronousQueue)
       queue.close
--- a/logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
+++ b/logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
@@ -5,6 +5,8 @@ require "logstash/instrument/collector"
 
 describe LogStash::Util::WrappedSynchronousQueue do
 
+  subject {LogStash::Util::WrappedSynchronousQueue.new(5)}
+
   describe "queue clients" do
     context "when requesting a write client" do
       it "returns a client" do
@@ -18,15 +20,9 @@ describe LogStash::Util::WrappedSynchronousQueue do
       end
     end
 
-    class DummyQueue < Array
-      def take() shift(); end
-      def poll(*) shift(); end
-    end
-
     describe "WriteClient | ReadClient" do
-      let(:queue) { DummyQueue.new }
-      let(:write_client) { LogStash::Util::WrappedSynchronousQueue::WriteClient.new(queue)}
-      let(:read_client)  { LogStash::Util::WrappedSynchronousQueue::ReadClient.new(queue)}
+      let(:write_client) { LogStash::Util::WrappedSynchronousQueue::WriteClient.new(subject)}
+      let(:read_client)  { LogStash::Util::WrappedSynchronousQueue::ReadClient.new(subject)}
 
       context "when reading from the queue" do
         let(:collector) { LogStash::Instrument::Collector.new }
@@ -95,25 +91,32 @@ describe LogStash::Util::WrappedSynchronousQueue do
 
         it "appends batches to the queue" do
           batch = write_client.get_new_batch
-          5.times {|i| batch.push(LogStash::Event.new({"message" => "value-#{i}"}))}
+          messages = []
+          5.times do |i|
+            message = "value-#{i}"
+            batch.push(LogStash::Event.new({"message" => message}))
+            messages << message
+          end
           write_client.push_batch(batch)
           read_batch = read_client.read_batch
           expect(read_batch.size).to eq(5)
-          i = 0
           read_batch.each do |data|
-            expect(data.get("message")).to eq("value-#{i}")
+            message = data.get("message")
+            expect(messages).to include(message)
+            messages.delete(message)
             # read_batch.cancel("value-#{i}") if i > 2     # TODO: disabled for https://github.com/elastic/logstash/issues/6055 - will have to properly refactor
-            data.cancel if i > 2
-            read_batch.merge(LogStash::Event.new({"message" => "generated-#{i}"})) if i > 2
-            i += 1
+            if message.match /value-[3-4]/
+              data.cancel
+              read_batch.merge(LogStash::Event.new({ "message" => message.gsub(/value/, 'generated') }))
+            end
           end
           # expect(read_batch.cancelled_size).to eq(2) # disabled for https://github.com/elastic/logstash/issues/6055
-          i = 0
+          received = []
           read_batch.each do |data|
-            expect(data.get("message")).to eq("value-#{i}") if i < 3
-            expect(data.get("message")).to eq("generated-#{i}") if i > 2
-            i += 1
+            received << data.get("message")
           end
+          (0..2).each {|i| expect(received).to include("value-#{i}")}
+          (3..4).each {|i| expect(received).to include("generated-#{i}")}
         end
       end
     end
--- a/logstash-core/lib/logstash/queue_factory.rb
+++ b/logstash-core/lib/logstash/queue_factory.rb
@@ -29,7 +29,9 @@ module LogStash
         LogStash::Util::WrappedAckedQueue.create_file_based(queue_path, queue_page_capacity, queue_max_events, checkpoint_max_writes, checkpoint_max_acks, checkpoint_max_interval, queue_max_bytes)
       when "memory"
         # memory is the legacy and default setting
-        LogStash::Util::WrappedSynchronousQueue.new
+        LogStash::Util::WrappedSynchronousQueue.new(
+          settings.get("pipeline.batch.size") * settings.get("pipeline.workers") * 2
+        )
       else
         raise ConfigurationError, "Invalid setting `#{queue_type}` for `queue.type`, supported types are: 'memory_acked', 'memory', 'persisted'"
       end
--- a/logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
+++ b/logstash-core/spec/logstash/instrument/wrapped_write_client_spec.rb
@@ -101,7 +101,7 @@ describe LogStash::Instrument::WrappedWriteClient do
   end
 
   context "WrappedSynchronousQueue" do
-    let(:queue) { LogStash::Util::WrappedSynchronousQueue.new }
+    let(:queue) { LogStash::Util::WrappedSynchronousQueue.new(1024) }
 
     before do
       read_client.set_events_metric(metric.namespace([:stats, :events]))


suyograo:
@jordansissel can you review this please?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -190,28 +192,29 @@ def initialize(queue, size, wait)
...
 
       def read_next
-        @size.times do |t|
-          event = @queue.poll(@wait)
-          return if event.nil? # queue poll timed out
-
-          @originals[event] = true
+        @queue.queue.drainTo(@originals, @size)
+        if @originals.size == 0 
+          e = @queue.poll(@wait)
```
andrewvc: This looks like we no longer poll in a loop anymore, but rather poll once? This would be a departure from our current guarantees AFAICT, what's the rationale here?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -190,28 +192,29 @@ def initialize(queue, size, wait)
...
+          unless e.nil?
+            @originals.add(e)
+            @queue.queue.drainTo(@originals, @size - 2)
```
andrewvc: Is this math what you meant to do? I think you want to drain `@size - @originals` so as not to have a larger batch than `-b` indicates yes?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -234,9 +237,10 @@ def each(&blk)
 
         # below the checks for @cancelled.include?(e) have been replaced by e.cancelled?
         # TODO: for https://github.com/elastic/logstash/issues/6055 = will have to properly refactor
-        @originals.each do |e, _|
+        for e in @originals
```
jordansissel: why rewrite this to `for x in y` from `y.each do ...` ?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -190,28 +192,29 @@ def initialize(queue, size, wait)
         @size = size
         @wait = wait
 
-        @originals = Hash.new
-
         # TODO: disabled for https://github.com/elastic/logstash/issues/6055 - will have to properly refactor
         # @cancelled = Hash.new
 
+        @originals = java.util.HashSet.new size, 1
```
jordansissel: please use parentheses HashSet.new(size, 1)

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -2,13 +2,20 @@
...
 module LogStash; module Util
   class WrappedSynchronousQueue
+    java_import java.util.concurrent.ArrayBlockingQueue
     java_import java.util.concurrent.SynchronousQueue
     java_import java.util.concurrent.TimeUnit
 
-    def initialize
-      @queue = java.util.concurrent.SynchronousQueue.new
+    def initialize (size = 0)
+      if size == 0
```
jordansissel: Why would this ever be 0?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -73,7 +80,7 @@ def close
       end
 
       def empty?
-        true # synchronous queue is alway empty
+        @queue.queue.isEmpty
```
jordansissel: Does this work? If `@queue` is an ArrayBlockingQueue. there is no such `queue` method that I can see (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ArrayBlockingQueue.html). If it is a SynchronousQueue, I cannot find any such `queue` method either (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/SynchronousQueue.html)

jordansissel: Nevermind. Github's diff rendering made this hard to understand. I didn't see this was part of ReadClient and it looked part of WrappedSynchronousQueue..

Ignore this comment.

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
```diff
@@ -45,7 +45,7 @@ def take() shift(); end
       def poll(*) shift(); end
     end
 
-    describe "WriteClient | ReadClient" do
+    xdescribe "WriteClient | ReadClient" do
```
jordansissel: did you mean to make this change?

logstash-core/spec/logstash/queue_factory_spec.rb
```diff
@@ -72,9 +74,10 @@
   context "when `queue.type` is `memory`" do
     before do
       settings.set("queue.type", "memory")
+      settings.set("pipeline.batch.size", 1024)
```
jordansissel: Is this important for the test?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
jordansissel: Can you add a comment why you `drainTo` with max of `@size - 2` ?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -190,28 +192,29 @@ def initialize(queue, size, wait)
...
       end
 
       def read_next
-        @size.times do |t|
-          event = @queue.poll(@wait)
-          return if event.nil? # queue poll timed out
-
-          @originals[event] = true
+        @queue.queue.drainTo(@originals, @size)
+        if @originals.size == 0
```
jordansissel: Did you intend to include what seems like https://github.com/elastic/logstash/pull/7623 in this PR?

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
jordansissel: If you think it's relevant, you can delete these tests instead of skipping them.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @andrewvc I'm only off by `1` :D 

You're right though, we want `@size - @originals.size` which is `1` in this scenario. Don't ask me why I put a `2` there, lack of sleep is the most likely explanation :)

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @jordansissel wops this needs to go away, the `drainTo` logic won't work with a `SynchronousQueue` anyways ... sorry leftover from some experiments, will delete

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @jordansissel as I said above to @andrewvc ... this should be `@size - 1` because we know we already `poll`ed `a single event. My bad :)

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @jordansissel because `@originals` is a Java `HashSet` now and not a `Map`/`Hash` anymore

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
original-brownbear: @jordansissel these are broken because the `DummyQueue` here doesn't expose the `queue` accessor. I think these are irrelevant now, but they also test some stuff about the metrics. Let me double check if these are actually redundant before killing them.

logstash-core/spec/logstash/queue_factory_spec.rb
original-brownbear: @jordansissel yea, because the size of the `ArrayBlockingQueue` is set as twice the product of this setting and the worker count (obviously there are many ways to size this, it just seemed like a reasonable starting point).

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @andrewvc we only `poll` if `drainTo` didn't give us any events. We still have the same guarantee that a batch will contain at least one event (actually it's more likely now since we're making the `drainTo` call as well. We don't have the timed `poll` for more than one event and are breaking guarantees there yes, but:
If you think about it, practically we have a much stronger guarantee now. We pretty much have a `100%` chance to get a completely filled batch in the first call to `@queue.queue.drainTo(@originals, @size)` anyway. 
Now that the `Queue` has capacity, it doesn't really make sense to have a timeout here since load spikes are handled by the `Queue`'s capacity + fast producers guarantee 100% full batches. Slow producers may see smaller average batch fill sizes, but:
1. This is a very rare case afaik? Filters and outputs are the bottlenecks in all example I benchmarked so far. 
2. Obviously, you could have a slow Kafka input or so but even then the input is mostly kinda "bursty" and should see larger batch sizes from not blocking on the push to the Queue all that much.

Hope that makes sense :)

PS: IMO, we shouldn't even have a `@wait` there and just poll indefinitely to avoid zero length buckets, but

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @jordansissel yea that was intentional (see PR description), didn't want to adjust one level of locking on the basis of a dirty locking approach.
Or put nicer :) -> wanted to make sure this still works fine without a second layer of outer locking

original-brownbear:
@andrewvc @jordansissel fixed you guys had I think.

Also fixed the ignored tests to run against an actual queue. The problem here was that a `HashSet` doesn't have a defined iteration order, so tests assuming order in messages read like this one won't work.
I know this is a slight behavior change, but we don't have order guarantees here and we only really change the `-w 1` case in practice anyway.
=> Fixed to not require ordered reads.

original-brownbear:
@andrewvc @jordansissel btw, updated the performance numbers after rebasing since `master` got a lot faster from other fixes merged in yesterday.
Change is not as pronounced now, but actually much more illustrative of the point.
CPU time is now the same for both approaches, but use time is better here from waiting less. Pretty much exactly what you would expect in theory (ignoring GC and context switches) :)

original-brownbear:
@andrewvc @jordansissel hmmmm hmmm hmmmm :)

Tested this a little more today while looking at other stuff to get some sample size, not so sure we can go with this just yet :(

After some longer runs, it seems that this way speeds up the case where inputs are faster than outputs. But, as soon as you get to a point where the queue is not full enough to give you full batches it's actually a slowdown of up to 50% (most extreme case of just stdin to stdout with dots codec and 4 threads) ...
It's pretty obvious that this is caused by very small batches being read. I have an idea on how to fix this though :) will implement it later.

original-brownbear:
@andrewvc @jordansissel  fixed, now we're faster in either case. I simply made it fall back to the old approach of polling the Queue in case we don't read full batches. This is probably not 100% optimal, but a guaranteed improvement for all use cases that and doesn't require more effort.
Also lowered the load factor for the `Hashset` to `0.75`, `1` was a dumb experiment  <= memory isn't the issue here it seems and there's a very visible difference between `0.75` and `1`.

Good to review again I think :)

original-brownbear:
We should wait for https://github.com/elastic/logstash/pull/7725 before doing any more benchmarking, that one changes the dynamic a lot in favor of this change it seems ...

jordansissel:
CI failed, but it seems unrelated.

```
Failures:

  1) LogStash::Pipeline when collecting metrics in the pipeline global metric populates the different metrics
     Failure/Error: end.to be_truthy

       expected: truthy value
            got: false
     # /var/lib/jenkins/workspace/elastic+logstash+pull-request+multijob-linux-intake/vendor/bundle/jruby/2.3.0/gems/rspec-wait-0.0.9/lib/rspec/wait/handler.rb:13:in `block in handle_matcher'

...

Finished in 12 minutes 53 seconds (files took 6.94 seconds to load)
2885 examples, 1 failure, 3 pending
```

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
jordansissel:
```
>> java.util.HashSet.ancestors.include?(Enumerable)
=> true

>> a = java.util.HashSet.new
=> #<Java::JavaUtil::HashSet:0x72cc7e6f>
>> a.add("Hello")
=> true
>> a.each { |x| puts x }
Hello
```
```

Comment (jordansissel) [inline logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb]: Further, `for x in y` calls `y.each` anyway:

```
>> a = BasicObject.new
>> for x in a; puts x; end
NoMethodError: undefined method `each' for #<BasicObject:0x6fa51cd4>
```

File: logstash-core/spec/logstash/queue_factory_spec.rb (hunk 8 above)
Comment (jordansissel) [inline logstash-core/spec/logstash/queue_factory_spec.rb]: no what I mean is that previously the size was not explicitly set in the test, so why do you need to set it now?

Comment (original-brownbear) [inline logstash-core/spec/logstash/queue_factory_spec.rb]: @jordansissel because now the size of the `ArrayBlockingQueue` is derived from this setting by the `queue_factory`

**Comment by original-brownbear** [comment]
@jordansissel 

> CI failed, but it seems unrelated.

yea it's a known issue https://github.com/elastic/logstash/issues/7738

File: logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb (hunk 3 above)
Comment (original-brownbear) [inline logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb]: @jordansissel ah, I can revert that tomorrow then :)

**Comment by original-brownbear** [comment]
@jordansissel alrighty, reverted the change to loop syntax. All comments addressed right?

File: logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb [hunk 10]
```diff
@@ -175,28 +173,29 @@ def initialize(queue, size, wait)
         @size = size
         @wait = wait
 
-        @originals = Hash.new
-
         # TODO: disabled for https://github.com/elastic/logstash/issues/6055 - will have to properly refactor
         # @cancelled = Hash.new
 
+        @originals = java.util.HashSet.new(size * 4 / 3 + 1, 0.75)
```
jordansissel: Can you include code comments here that explains the math? I have a feeling future readers may become confused without more detail.

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
```diff
@@ -95,25 +91,32 @@ def poll(*) shift(); end
 
         it "appends batches to the queue" do
           batch = write_client.get_new_batch
-          5.times {|i| batch.push(LogStash::Event.new({"message" => "value-#{i}"}))}
+          messages = []
+          8.times do |i|
```
jordansissel: Why change this from 5 to 8? Is 8 an important number to test more than 5 would be?

If 8 is not magic, then maybe we could add randomization and pick a random size for the queue to be and a random number of events to write.

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
```diff
@@ -95,25 +91,32 @@ def poll(*) shift(); end
...
           read_batch.each do |data|
-            expect(data.get("message")).to eq("value-#{i}")
+            message = data.get("message")
+            expect(messages).to include(message)
+            messages.delete(message)
             # read_batch.cancel("value-#{i}") if i > 2     # TODO: disabled for https://github.com/elastic/logstash/issues/6055 - will have to properly refactor
-            data.cancel if i > 2
-            read_batch.merge(LogStash::Event.new({"message" => "generated-#{i}"})) if i > 2
-            i += 1
+            if message.match /value-[3-7]/
```
jordansissel: This feels oddly specific. Why change the `if i > 2` to this?

original-brownbear: @jordansissel this was just a result of the not being ordered anymore. I found it easier to read than `[3-9]` because we already have the `8` hard set :D

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
original-brownbear: @jordansissel ok this is a really stupid thing now (it was a dumb change, to begin with :P). There were tests above this one for `take` and `offer(timeout)` that are now gone but for which the `8` worked as a size without more code changes.
Will revert :)

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @jordansissel :D the `0.75` as a redundant default was my plan for a comment here ... a bad plan in hindsight lol
Will add a proper comment :)

original-brownbear:
@jordansissel requested changes are in :)

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -58,7 +61,7 @@ def close
       end
 
       def empty?
-        true # synchronous queue is alway empty
+        @queue.queue.isEmpty
```
andrewvc: Shouldn't this be `@queue.isEmpty` ?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -170,28 +173,30 @@ def initialize(queue, size, wait)
         @size = size
         @wait = wait
 
-        @originals = Hash.new
-
         # TODO: disabled for https://github.com/elastic/logstash/issues/6055 - will have to properly refactor
         # @cancelled = Hash.new
 
+        #Sizing HashSet to size/load_factor to ensure no rehashing
+        @originals = java.util.HashSet.new(size * 4 / 3 + 1, 0.75)
```
andrewvc: If we want to avoid rehashing ,why not set the load factor to 1? I agree that it makes sense to have a larger than needed hash table to avoid collisions for quantities this small, but if the goal is to ensure no rehashing, why keep it at 0.75?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @andrewvc nah careful, we're in `ReadClient` here and `WrappedSynchronousQueue` doesn't define an `isEmpty?` method :)

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @andrewvc I benchmarked with `1` and `0.75` and saw no measurable difference in GC times or even bytes/objects allocated => I decided to be nice to the CPU given that we do quite a few `includes` checks on this thing.

original-brownbear:
@jordansissel thanks!

PR #7692, opened 2017-07-14
#7676 Don't use cache for Event sprintf, use efficient implementation instead

### LINKED ISSUE #7676: StringInterpolation cache causes OOM
We have reproducably run into OOMs due to a design issue with the StringInterpolation class. 
Specifically we run into a problem which is already described in the class: 
https://github.com/elastic/logstash/blob/master/logstash-core/src/main/java/org/logstash/StringInterpolation.java#L17

> // The current implementation doesn't allow the keys to expire, I think under normal usage
> // the keys will converge to a fixed number.
> // If this code make logstash goes OOM, we have the following options: ...

- Version: 5.4.3
- Operating System: debian
- Steps to Reproduce: use the influxdb output with the config option TODO set to true. Make sure that your events still have the "message" field when they reach the output plugin. 
- example config: 
```
input 
{
  kafka
  {
    ....
    codec => "line"
  }
}
filter
{
  grok
  {
    match => { "message" =>  "^(?<datacenter>[^\.]+?)\.(?<service>[^\.]+?)\.(?<host>[^\.]+?)\.(?<measurement>[^\s]+?)\s%{NUMBER:value:float}\s%{INT:time}\s*$"}
  }
}
output
{
  influxdb
  {
    host => "10.1.2.3"
    password => "password"
    user => "logstash"
    db => "collectd_test"
    measurement => "%{measurement}"
   
    send_as_tags => ["datacenter","service","host","application","instance","topic","partition","disk"]
    allow_time_override => true
    time_precision => "s"
    coerce_values =>
    {
     "value" => "float"
    }
    retention_policy => "rp"

    use_event_fields_for_data_points => true
    exclude_fields => ["@timestamp", "@version", "sequence", "message", "type", "measurement", "ip"]
  }
}
```

The last two lines of the influxdb config are most important.
We have observed that the problem seems to happen only when the option "use_event_fields_for_data_points" is set in the influx output config. 
When creating a datapoint for shipping to the database, the influxdb output plugin [calls the event.sprintf(v)](https://github.com/logstash-plugins/logstash-output-influxdb/blob/master/lib/logstash/outputs/influxdb.rb#L209) method on the name and value of each field which has been configured to go into the database. When setting the "use_event_fields_for_data_points" option to true, this happens on all fields of the event, including the "message" field generated by the kafka input. 
The event.sprintf method is then [calling](https://github.com/elastic/logstash/blob/master/logstash-core/src/main/java/org/logstash/Event.java#L271) the evalute() method on the StringInterpolation class. 
The StringInterpolation class [keeps a cache](https://github.com/elastic/logstash/blob/master/logstash-core/src/main/java/org/logstash/StringInterpolation.java#L12) of all strings passed to it, and this cache is where the memory leak happens, because values go in, but seem to never go out. 

In our tests we ended up with ConcurrentHashMaps with multiple millions of kafka message strings in them, resulting in garbage collection times of 20 seconds.

A cheap workaround for us is to drop the message field before the event reaches the output plugin, but it might be worth saving others the time and hassle to debug.

original-brownbear:
Fixes #7676 and generally improves the performance of `org.logstash.StringInterpolation#evaluate` by:

* Don't use regex to find the contents between `%{}`, it's way too inefficient for a simple case like this with all its allocations
* Flatten logic out of the many helper classes
* Use a reusable, thread-local `StringBuilder` to serialize into
* Don't use `.substring` for comparisons
* `static` `ObjectMapper`
* `org.joda.time.format.DateTimeFormat#forPattern` not being cached is not an issue, since it's cached in `org.joda.time.format.DateTimeFormat#cPatternCache`

* Added benchmark
   * ~15% faster for a few cases I ran, even for repeated patterns with interpolation + couldn't find any case even with date formatting, where this implementation is slower than the cached one
   *  no leaking on dynamic patterns

PS: All tests still pass and this thing has good coverage it seems :P


jordansissel:
Your PR description has lots of nice details, but your commit doesn't have this same information. Can you copy the PR description into your commit message?

original-brownbear:
@jordansissel sure, sec :)

original-brownbear:
@jordansissel done :)

logstash-core/src/main/java/org/logstash/StringInterpolation.java
```diff
@@ -1,92 +1,80 @@
...
+        final StringBuilder builder = STRING_BUILDER.get();
+        int pos = 0;
+        while (open > -1 && close > -1) {
+            if (open > 0) {
+                builder.append(template, pos, open);
             }
-
-            if(pos <= template.length() - 1) {
-                compiledTemplate.add(new StaticNode(template.substring(pos)));
+            if (template.regionMatches(open + 2, "+%s", 0, 3)) {
```
jordansissel: Hardcoding `3` for length gives you interesting results that are differnt from Logstash 5.5. I think you need to check the whole length (`close - open - 1` ?)



| outcome | input | v5.5.0 | this pr | 
|-|-|-|-|
| same | +%s | 1500508897 | 1500509113 |
| different | +%ss | %43 | 1500509162 |
| different | +%s d y | %30 20 2017 | 1500509234 |

In all cases above, since the first 3 chars are "%+s" it assumes unix epoch time and ignores the rest.

original-brownbear: @jordansissel huh good find, let me a add a test + fix for this case :)

original-brownbear:
@jordansissel fixed and test for the problematic case added :)

original-brownbear:
@jordansissel thanks!

PR #7693, opened 2017-07-14
MINOR: Size HashMaps more Efficiently in ConvertedMap

original-brownbear:
Trivial, but visible in the number of `HashMap$Node` that need to be GCed.
We should set a size here that takes the load factor into account to prevent rehashing on instantiation of the `ConvertedMap`.
Also for some reason, the constructor here become `public` which it really shouldn't be since we're using this class to ensure the correct types of the values, made it `private` again.


jakelandis:
TIL, thanks!

LGTM

original-brownbear:
@jakelandis thanks for taking a look! :)

Issue #4772, opened 2016-03-07
We should seriously consider using a .jrubyrc file to optimize JRuby


guyboertje:
Whenever I drop one of these into a perf session, I see speed improvements.

Perhaps @headius can advise which settings would be optimal for Logstash?

Example:

```
################################################################################
# compiler settings
################################################################################

# Set compilation mode. JIT = at runtime; FORCE = before execution.
# Options: [JIT, FORCE, OFF, OFFIR], Default: JIT.
# compile.mode=JIT
compile.mode=FORCE

# Dump to console all bytecode generated at runtime.
# Options: [true, false], Default: false.
compile.dump=false

# (EXPERIMENTAL) Turn on compilation without polling for "unsafe" thread events.
# Options: [true, false], Default: false.
compile.threadless=false

# Turn on fast operators for Fixnum and Float.
# Options: [true, false], Default: true.
compile.fastops=true

# Set the number of lines at which compiled bodies are "chained".
# Options: [Integer], Default: 500.
compile.chainsize=500

# Generate method bindings (handles) for compiled methods lazily.
# Options: [true, false], Default: false.
compile.lazyHandles=false

# Enable or disable peephole optimizations.
# Options: [true, false], Default: true.
compile.peephole=true

# Compile calls without guards, for experimentation.
# Options: [true, false], Default: false.
compile.noguards=false

# Compile with all "mostly harmless" compiler optimizations.
# Options: [true, false], Default: false.
compile.fastest=true

# Compile obj.__send__(<literal>, ...) as obj.<literal>(...).
# Options: [true, false], Default: false.
compile.fastsend=false

# Return true from multiple assignment instead of a new array.
# Options: [true, false], Default: false.
compile.fastMasgn=false

# Use invokedynamic for optimizing Ruby code
# Options: [true, false], Default: false.
# compile.invokedynamic=false
compile.invokedynamic=true


################################################################################
# invokedynamic settings
################################################################################

# Maximum call site failures after which to inline cache.
# Options: [Integer], Default: 1000.
invokedynamic.maxfail=1000

# Maximum polymorphism of PIC binding.
# Options: [Integer], Default: 2.
invokedynamic.maxpoly=2

# Log binding of invokedynamic call sites.
# Options: [true, false], Default: false.
invokedynamic.log.binding=false

# Log invokedynamic-based constant lookups.
# Options: [true, false], Default: false.
invokedynamic.log.constants=false

# Enable all possible uses of invokedynamic.
[...]

PR #7698, opened 2017-07-15
MINOR: default to -Djruby.compile.invokedynamic=true

original-brownbear:
Any reason we shouldn't set `-Djruby.compile.invokedynamic=true` as a default?

I just tested this and it gives us a **15-20%** gain when running the Apache test set. I don't see why this shouldn't be on by default for a long running process like Logstash?

### DIFF
--- a/config/jvm.options
+++ b/config/jvm.options
@@ -51,6 +51,9 @@
 # use our provided JNA always versus the system one
 #-Djna.nosys=true
 
+# Turn on JRuby invokedynamic
+-Djruby.compile.invokedynamic=true
+
 ## heap dumps
 
 # generate a heap dump when an allocation from the Java heap fails


guyboertje:
@original-brownbear - I created [this issue](https://github.com/elastic/logstash/issues/4772) a while back. The set of settings in that issue are a bit out of date WRT JRuby 9000. I took them from the list given on the JRuby wiki but I do not have enough JVM experience to really get the significance of each setting.

It looks like you are well placed to investigate the effectiveness of the settings that can be adjusted in `.jrubyrc`.

original-brownbear:
@guyboertje that issue was my inspiration for this actually :D (could've sworn I linked it but apparently I didn't).

> It looks like you are well placed to investigate the effectiveness of the settings that can be adjusted in .jrubyrc.

Tbh, I tried to look into a little, but this was the only thing I came up with (at least for now).
I played with some other JIT settings to get around our deep stacks preventing inlining at times, but increasing the inline depth had a slightly negative effect at least in my tests.
Happy to do some more testing for that issue once we get more automation from the benchmark tool though :)

That said, ignoring start-up times increasing, this should be a significant win in throughput :)

guyboertje:
@original-brownbear - OK, you have a deal. :-)

original-brownbear:
@guyboertje nice thanks!

guyboertje:
@original-brownbear - please bear in mind that these settings may not be set-and-forget - the gentle java-fication of LS means its a moving target :-) We have never had the luxury of a benchmarking suite that allows us to change these settings in relevant groups to verify that all is still well with them.

Issue #7701, opened 2017-07-16
We Should Fix Java Compiler Warnings


original-brownbear:
I activated the `javac` linter warnings in #7674 and we have many of them showing in each build.
Some of them are somewhat irrelevant, especially those in test code, but others in production code point us towards clear code smells like the many chained `instanceof` calls we have.

We should clean these up where possible and suppress them where it isn't.

liketic:
I can work on this.

PR #7703, opened 2017-07-16
MINOR: Fix compiler warnings in Ingest Converter Test

original-brownbear:
Trivial contribution to #7701 , but saves us 2 compiler warnings :)


original-brownbear:
@jsvd many thanks for the reviews! :)

PR #7705, opened 2017-07-16
MINOR: Fix compiler warnings from general Class type use

original-brownbear:
Trivial, we only allow `Queueable`s as a type here => we can fix generic `Class` type use warnings for #7701.


jakelandis:
LGTM

original-brownbear:
@jakelandis thanks!

PR #7707, opened 2017-07-16
MINOR: Turn NullBiValue into a Singleton

original-brownbear:
The Ruby execution environment is a singleton in the way we use JRuby (calling Java from a Ruby application) => There is no reason to instantiate multiple instances of the `null` wrapper, they will all point to the same `RubyNil` instance anyway and have no further state stored.

* Saves us `~30k` of these when benchmarking the Apache dataset.
* Also fixed one more compiler warning from the raw return type on `org.logstash.bivalues.BiValues.BiValueType#build` :)
* Had to adjust `org.logstash.TestBase#ruby` to use the global `Ruby` instance as well, which is more accurately representing our environment anyways imo

### DIFF
--- a/logstash-core/src/main/java/org/logstash/bivalues/BiValues.java
+++ b/logstash-core/src/main/java/org/logstash/bivalues/BiValues.java
@@ -1,21 +1,19 @@
 package org.logstash.bivalues;
 
-import org.logstash.Timestamp;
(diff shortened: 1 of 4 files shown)


original-brownbear:
@guyboertje thanks :)

Note to self: Needs backport to `5.x` :(

PR #7708, opened 2017-07-17
MINOR: Fix all compiler warnings in AccessorsTest and EventTest

original-brownbear:
Completely trivial auto-fix for #7701 :

* All raw generic type use in `EventTest` and `AccessorsTest` fixed
* All misordered assertions in `AccessorsTest` flipped or replaced by their simplified version


jakelandis:
@jsvd - Both [Hamcrest](https://github.com/hamcrest/JavaHamcrest) and [AssertJ](http://joel-costigliola.github.io/assertj/) assertions start with actual as well.  I strongly advise [AssertJ](http://joel-costigliola.github.io/assertj/) for any new assertions.

PR #7709, opened 2017-07-17
MINOR: Factor out cold-path from org.logstash.Event#initTimestamp

original-brownbear:
This is a pretty visible offender when looking at the compile logs.
The hot path (as commented) is def the breakout `if (o == null || o instanceof NullBiValue) {` but the method isn't inlined because of the huge conditional that follows and is super cold.

=> factored it out and now `initTimestamp` gets inlined into the constructor of `Event`.


original-brownbear:
@jsvd thanks! 

note for me: needs backport to 5.x :(

PR #7720, opened 2017-07-18
Metrics: Fix RubyTimestamp serialization

### LINKED ISSUE #7719: RubyTimestamp is not serialized properly for the metrics.
The following warning is seen when reloading the configuration:
```
[2017-07-17T13:17:11,314][WARN ][org.logstash.instrument.metrics.gauge.LazyDelegatingGauge] A gauge metric of an unknown type (org.logstash.ext.JrubyTimestampExtLibrary.RubyTimestamp) has been create for key: last_failure_timestamp, namespace:[stats, pipelines, main, reloads]. This may result in invalid serialization.  It is recommended to log an issue to the responsible developer/development team.
```
...and the ```last_success_timestamp``` and ```last_failure_timestamp``` under pipelines\reloads is not populated correctly.

jakelandis:
This change introduces a RubyTimestamp and Timestamp gauge. 

This change requires setting the and retrieving the value with different types. The Gauge interface has been update to allow for setting / getting using different types.

Fixes #7719

### DIFF
--- a/logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
@@ -3,9 +3,10 @@ package org.logstash.instrument.metrics.gauge;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jruby.RubyHash;
+import org.logstash.Timestamp;
+import org.logstash.ext.JrubyTimestampExtLibrary.RubyTimestamp;
 import org.logstash.instrument.metrics.AbstractMetric;
 import org.logstash.instrument.metrics.MetricType;
-import org.logstash.instrument.metrics.counter.CounterMetric;
 
 import java.util.List;
 
@@ -13,7 +14,7 @@ import java.util.List;
  * A lazy proxy to a more specific typed {@link GaugeMetric}. The metric will only be initialized if the initial value is set, or once the {@code set} operation is called.
  * <p><strong>Intended only for use with Ruby's duck typing, Java consumers should use the specific typed {@link GaugeMetric}</strong></p>
  */
-public class LazyDelegatingGauge extends AbstractMetric<Object> implements GaugeMetric<Object> {
+public class LazyDelegatingGauge extends AbstractMetric<Object> implements GaugeMetric<Object,Object> {
 
     private final static Logger LOGGER = LogManager.getLogger(LazyDelegatingGauge.class);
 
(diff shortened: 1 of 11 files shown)


jakelandis:
@andrewvc - would you mind reviewing this ?

I missed the RubyTimestamp in the initial version of https://github.com/elastic/logstash/pull/7595.  This change adds RubyTimestamp and Java Timestamp gauge types.

jakelandis:
After further discussions, we plan to depreciate most of the flavors of a Gauge due to a semantic mis-match in the context of more strict definition of a 'metric'. Where this more strict definition of a metric equates to a metric should only exist if it is there is usefulness when comparing the values over a time period, and could be placed in a graph such that they may be reasoned about in a meaningful way. 

In the context of this PR, the Timestamp does meet the requirements for the new definition. I will remove the Java Timestamp Gauge, but keep the Ruby version for backward compatibility.  I will log a new Issue / PR's to deprecate all the Guage's that don't meet the the stricter definition.   

cc @andrewvc

jakelandis:
Issue to start the process of deprecating legacy metric types https://github.com/elastic/logstash/issues/7746

andrewvc:
That sounds great Jake! :)

jakelandis:
@andrewvc - would you mind re-reviewing this ?  (I will rebase and push to hopefully fix the test failures)

logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
```diff
@@ -89,6 +90,10 @@ private synchronized void wakeMetric(Object value) {
                 lazyMetric = new BooleanGauge(nameSpaces, key, (Boolean) value);
             } else if (value instanceof RubyHash) {
                 lazyMetric = new RubyHashGauge(nameSpaces, key, (RubyHash) value);
+            } else if (value instanceof RubyTimestamp) {
+                lazyMetric = new RubyTimeStampGauge(nameSpaces, key, ((RubyTimestamp) value));
+            } else if (value instanceof Timestamp) {
+                lazyMetric = new TimestampGauge(nameSpaces, key, (Timestamp) value);
```
andrewvc: Should be deleted.

logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
```diff
@@ -89,6 +90,10 @@ private synchronized void wakeMetric(Object value) {
                 lazyMetric = new BooleanGauge(nameSpaces, key, (Boolean) value);
             } else if (value instanceof RubyHash) {
                 lazyMetric = new RubyHashGauge(nameSpaces, key, (RubyHash) value);
+            } else if (value instanceof RubyTimestamp) {
+                lazyMetric = new RubyTimeStampGauge(nameSpaces, key, ((RubyTimestamp) value));
```
jakelandis: I logged https://github.com/elastic/logstash/issues/7746 to address the deprecation as a separate issue. Until we remove / move the underlying metrics that uses this, users logs will get spammed with warnings with no re-course. Once the core/known RubyTimeStampGauge usage is moved/removed, the I will add the deprecation warning to the log.

logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
jakelandis: well.. that's just sloppy work (on my part). removing now.

jakelandis:
@andrewvc - review updated and change pushed.

logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
andrewvc: Good point!

jakelandis:
Jenkins test this please

Issue #6970, opened 2017-04-26
Dead references to LS_HEAP_SIZE


jarpy:
Some [references to the `LS_HEAP_SIZE` environment variable remain in the repo](https://github.com/elastic/logstash/search?utf8=%E2%9C%93&q=LS_HEAP_SIZE&type=) (particularly in the docs), but the variable looks to have been deprecated in favour of `LS_JAVA_OPTS`.

This may have to led to some user confusion. See: https://github.com/elastic/logstash-docker/issues/27

colinsurprenant:
Thanks @jarpy for the heads up. 
Two issues here:
- correct the docs /cc @dedemorton 
- fix the Windows scripts to get rid of `LS_HEAP_SIZE` to be consistent with Linux startup scripts.

colinsurprenant:
created issue #7567 for the Windows startup script, will keep this one specifically for the docs inconsistencies.

dedemorton:
@colinsurprenant  Can you check the following references and confirm that replacing the name is all that's required? 

command-line-flags.
*`-b, --pipeline.batch.size SIZE`*::
  Size of batches the pipeline is to work in. This option defines the maximum number of events an
  individual worker thread will collect from inputs before attempting to execute its filters and outputs.
  The default is 125 events. Larger batch sizes are generally more efficient, but come at the cost of
  increased memory overhead. You may have to increase the JVM heap size by setting the `LS_HEAP_SIZE`
  variable to effectively use the option.

performance-checklist.asciidoc
* The `pipeline.batch.size` setting defines the maximum number of events an individual worker thread collects before attempting to execute filters and outputs. Larger batch sizes are generally more efficient, but increase memory overhead. Some hardware configurations require you to increase JVM heap size by setting the `LS_HEAP_SIZE` variable to avoid performance degradation with this option. Values of this parameter in excess of the optimum range cause performance degradation due to frequent garbage collection or JVM crashes related to out-of-memory exceptions. Output plugins can process each batch as a logical unit. The Elasticsearch output, for example, issues https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html[bulk requests] for each batch received. Tuning the `pipeline.batch.size` setting adjusts the size of bulk requests sent to Elasticsearch.

dedemorton:
@colinsurprenant  Can you check the following references and confirm that replacing the name is all that's required? 

https://www.elastic.co/guide/en/logstash/current/tuning-logstash.html
[...]

PR #7725, opened 2017-07-18
MINOR: Set JVM's Xms == Xmx as a default

original-brownbear:
This is a really trivial gain with the default settings:

### Before:

screen shot 2017-07-18 at 16 37 17

### After:

screen shot 2017-07-18 at 16 35 17

less than half the number of GCs and each individual one is ~5x as fast as before (obviously, since we're not resizing the Heap all the time).
~25% speedup with 2 threads on the Apache dataset.

### DIFF
--- a/config/jvm.options
+++ b/config/jvm.options
@@ -3,7 +3,7 @@
 # Xms represents the initial size of total heap space
 # Xmx represents the maximum size of total heap space
 
--Xms256m
+-Xms1g
 -Xmx1g
 
 ################################################################
--- a/docs/static/performance-checklist.asciidoc
+++ b/docs/static/performance-checklist.asciidoc
@@ -43,6 +43,7 @@ You may be tempted to jump ahead and change settings like `pipeline.workers` (`-
 * Often times CPU utilization can go through the roof if the heap size is too low, resulting in the JVM constantly garbage collecting.
 * A quick way to check for this issue is to double the heap size and see if performance improves. Do not increase the heap size past the amount of physical memory. Leave at least 1GB free for the OS and other processes.
 * You can make more accurate measurements of the JVM heap by using either the `jmap` command line utility distributed with Java or by using VisualVM. For more info, see <<profiling-the-heap>>.
+* Always make sure to set the minimum (Xms) and maximum (Xmx) heap allocation size to the same value to prevent the heap from resizing at runtime, which is a very costly process.
 
 . *Tune Logstash worker settings:*
 +


original-brownbear:
@guyboertje ... this was by far the easiest to reason about JVM setting and I obviously missed it :D

jakelandis:
@original-brownbear - In general I am favor of this change, however on memory constrained systems this can cause extra OS paging of the JVM.  If we do this, it should be accompanied by some doc similar to [this]( https://www.elastic.co/guide/en/elasticsearch/guide/master/heap-sizing.html#_swapping_is_the_death_of_performance) that advises setting swappiness down low (or off for the brave), and may want to explore if we can expose a mlock all setting.

original-brownbear:
@jakelandis hmm I wonder if we should add doc for this, of cause you're right.
But:

```sh

################################################################
## Expert settings
################################################################
##
## All settings below this section are considered
## expert settings. Don't tamper with them unless
## you understand what you are doing
##
################################################################
```

Is what we put into this file. I don't think we want to support changes to the JVM settings actively (at least that's my takeaway from this disclaimer).

And in general, unless you don't have 1GB of free memory this is the far superior default.

> In general I am favor of this change, however on memory constrained systems this can cause extra OS paging of the JVM.

My reasoning (from the perspective of what the better default is) was this:

1. Realizing that you're slowed down by resizing the Heap constantly => hard for users.
2. Realizing that you're at the memory limit of your system => easy for users.

 @suyograo, maybe can give an opinion on docs yes or no here? Seems like more of a product thing :)

andrewvc:
Great find! I'm shocked that we weren't already doing this.

@jakelandis totally agree on the docs. We emphasize this point heavily in the [ES docs](https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html) .

We should also emphasize that point in our docs here: https://www.elastic.co/guide/en/logstash/current/performance-troubleshooting.html

original-brownbear:
@andrewvc to the rescue :) Will add a line to the docs there then, thanks!

original-brownbear:
@jakelandis @andrewvc added a point to the docs :)

jakelandis:
@original-brownbear - Can we add doc related to the swappiness ?

We should be able to more or less copy paste from https://www.elastic.co/guide/en/elasticsearch/guide/master/heap-sizing.html#_swapping_is_the_death_of_performance (except for the ```mlockall``` options)

original-brownbear:
@jakelandis we can (obv. :D), but again I'm unsure if we want to expand the scope of our docs here? (it seems like that would be a separate discussion to be had elsewhere? :))

jakelandis:
@original-brownbear - Fair point, the the doc and the default setting are related but not coupled.

While I agree with the change, I will let someone with more Logstash tenure make the final call on this.

original-brownbear:
@jakelandis jup let's ping the higher ups :)
@suyograo @jordansissel any opinion on this ?:)

suyograo:
+1 on making min and max heap size to be the same. I thought we were doing this already based on what ES does!

Re: docs, from our recent email thread re: heap size, we need to handle https://github.com/elastic/logstash/issues/6970 anyway, so I would do it there.

suyograo:
LGTM

suyograo:
Filed https://github.com/elastic/logstash/issues/7727

original-brownbear:
@suyograo maybe we should add a bootstrap check for `Xms == Xmx` and log a warning if the two are set to different values (that seems like it could prevent quite a few questions)?

jordansissel:
@original-brownbear I"m open to such a bootstrap check.

original-brownbear:
here we go: https://github.com/elastic/logstash/issues/7775

suyograo:
@original-brownbear We should backport this to 5.5 branch as well. This is a good find and clearly an oversight on our end, so this should go out in the next patch release of 5.5.

original-brownbear:
@suyograo sounds good, backport incoming in https://github.com/elastic/logstash/pull/7861 :)

georgexsh:
change `ms` to 1g cause RPM on CentOS7 failed to install systemd unit file on a 2g memory box.

PR #7730, opened 2017-07-18
BUG: Accessors were quietly swallowing broken syntax

original-brownbear:
Two things here (sorry for putting them into one PR, but they were really close logically and it seems like it would lead to a clearer git history to not have two back-to-back commits on the relevant lines here), so one important thing and one is just cleanup:

### Easy clean-up first:

Throwing `IndexOutOfBoundsException` in `listIndex ` only to catch it every time and interpret as wrong input makes no sense, that method only returns valid values `>=0` -> just return `-1` to signal an issue => fix exceptions as flow control anti-pattern + is much easier on the JIT in the very hot path here.

### Bug

We were quietly swallowing what the deleted test shows, the following did not even log a warning:

1. Put a list to `[a][b]`
2. Store something to or get something from `[a][b][c]` (c being not numeric) would simply be indistinguishable from any other `set` valid call and just return `null` for `get` without being distinguishable from any other call to an empty/non-existing field. (note: the `null` return on the `set` calls or any return from it, is simply ignored in `Event`)

I don't think this should be something we catch at any level. This seems equivalent to a syntax error in a plugin and should outright die to prevent data-loss and creating bugs that are almost impossible to debug.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core/src/main/java/org/logstash/Accessors.java
@@ -28,11 +28,7 @@ public class Accessors {
             ((Map<String, Object>) target).put(key, value);
         } else if (target instanceof List) {
             int i;
-            try {
-                i = Integer.parseInt(key);
-            } catch (NumberFormatException e) {
-                return null;
-            }
+            i = Integer.parseInt(key);
             int size = ((List<Object>) target).size();
             if (i >= size) {
                 // grow array by adding trailing null items
@@ -60,13 +56,12 @@ public class Accessors {
             if (target instanceof Map) {
                 return ((Map<String, Object>) target).remove(field.getKey());
             } else if (target instanceof List) {
-                try {
-                    int i = Integer.parseInt(field.getKey());
-                    int offset = listIndex(i, ((List) target).size());
-                    return ((List)target).remove(offset);
-                } catch (IndexOutOfBoundsException|NumberFormatException e) {
+                int i = Integer.parseInt(field.getKey());
+                final int offset = listIndex(i, ((List) target).size());
+                if (offset < 0) {
                     return null;
                 }
+                return ((List) target).remove(offset);
             } else {
                 throw newCollectionException(target);
             }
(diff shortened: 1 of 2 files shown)


original-brownbear:
@jordansissel @suyograo since this may be a breaking change to some users (though in a sense not really, they are already in a broken situation and potentially just don't know it), pinging you guys here :)

original-brownbear:
@jordansissel ping :)

jordansissel:
I was wondering when someone would wander into this ;)

> without being distinguishable from any other call to an empty/non-existing field

There is no way for a user to catch exceptions in Logstash today, so I am nervous about how this will negatively impact user experience.

In past discussions (I don't know where), we've considered adding a way for users to check if a field exists (different from "is it null?") in a Logstash condition, like `if (exists([field][ref]))` or similar. We don't have this today, so many users get by checking for the truthy/falsey value (nil or not nil) like this: `if [field][ref] { ... }` -- it is not ideal, but it works for most cases.

If we improve part of this scenario by having Logstash throw an exception (parseInt throws an exception, for example), we'll be throwing an exception that will cause some bad user experiences, I think, without giving folks a way to check for this scenario.

Thoughts?

Some deeper history here is that my intent with Logstash's config was to have it be typeless (no concept  for users to check if something is a string, number, map, etc), and this resulted in a non-existant field `[foo][bar]` having the same behavior (nil value) as a field reference that is not valid, such as `[foo][bar]` in `{ "foo": [1, 2, 3] }`.

For fixing this, I think we'll slowly slide towards a typed configuration language, and if that's our destination, I would prefer to reuse an existing language (groovy, javascript, ruby, etc) which has such concepts.

original-brownbear:
@jordansissel 

> If we improve part of this scenario by having Logstash throw an exception (parseInt throws an exception, for example), we'll be throwing an exception that will cause some bad user experiences, I think, without giving folks a way to check for this scenario.

Hmm:

* I mean they do have `includes()` as a call, that one won't throw even with these changes?
* For `get()` I could see an argument either way, between swallowing and throwing. Swallowing looks fine to me here as well if documented actually.
* But for `set`, this (swallowing and dropping the change) looks plain wrong. My reasoning would be this:
   * If you have busted data incoming, you probably want to get some visibility on that. Using `includes` is one way of determining what tree was actually deserialized into your `Event`
   * Using `get` for this is a little dirtier (though faster if you care about the data) but probably fine as well to save one walk along the tree
   * Quietly failing on `set` though, that means all your broken data does get processed in some really unexpected way without any warning/error.

How about this, to avoid any user impact and improve performance:

* catch all exceptions as high up the stack as possible `Event` and log each `Exception` as an error, but otherwise keep the existing behavior
*  clean up the messy exception as logic use for `IndexOutOfBounds` and the like, that's just wrong and not justified by anything to get the performance boost
* Do all of this after https://github.com/elastic/logstash/pull/7839 (huge speed up there already, no need to repeat work here that will have to be repeated there)

After thinking about this again, it seems (unfortunately to some degree) reasonable to not change the behavior beyond logging errors on broken `set` calls :(

original-brownbear:
Closing here opened an issue for discussing logging in this spot here https://github.com/elastic/logstash/issues/7869, the performance improvements here are super seeded by #7839 

Thanks for the input @jordansissel !

PR #7732, opened 2017-07-18
test multiple pipelines using RATS

jsvd:
* also adds `config.support_escapes` to list of pipeline settings

### DIFF
--- a/qa/integration/specs/multiple_pipeline_spec.rb
+++ b/qa/integration/specs/multiple_pipeline_spec.rb
@@ -1,0 +1,51 @@
+require_relative '../framework/fixture'
+require_relative '../framework/settings'
+require_relative '../services/logstash_service'
+require_relative '../framework/helpers'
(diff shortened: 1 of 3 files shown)


qa/integration/specs/multiple_pipeline_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  after(:all) {
+    @fixture.teardown
+  }
+
+  let(:pipelines) {[
+    {
+      "pipeline.id" => "test",
+      "pipeline.workers" => 1,
+      "pipeline.batch.size" => 1,
+      "config.string" => "input { } output { }"
```
suyograo: Can we use the configs `generator { count => 1 } output { file => "blah1"}` and `generator { count => 1 } output { file => "blah2"}`.

We can then check if blah1 and blah2 files are non empty. That's more comprehensive, right?

suyograo:
Woot! LGTM

PR #7741, opened 2017-07-19
Add netflow module

untergeek:
Adding the netflow module.


modules/netflow/README.md
```diff
@@ -0,0 +1,3 @@
+# Netflow Module README
+
+## Documentation here
```
suyograo: Can you add all the `var.netflow` overrides we expose in the README? will be easier to test

PR #7743, opened 2017-07-19
Test stabilization: Ensure that logging is set back to 'info' after tests that may change it

### LINKED ISSUE #7739: Build verbosity can lead to ci job failures on travis?
Lately there are many DEBUG level messages being printed during CI tests, so much that logs can be many megabytes, causing travis to not show everything, even in raw form https://travis-ci.org/elastic/logstash/jobs/255092890#L9475

Could this be from a recent fix of log4j2 logging?

jakelandis:
Fixes #7739


suyograo:
LGTM

PR #7736, opened 2017-07-19
Test stablization: 5x

jakelandis:
🤞

PR #7760, opened 2017-07-20
6.x: Test stablization: Add timeout sleep.until to prevent infinite loop,resuce error on shutdown

jakelandis:
Forward port of the 5.x https://github.com/elastic/logstash/pull/7736

Please review as isolated PR (not port) since there are many differences here between 6.x and 5.x

### DIFF
--- a/logstash-core/spec/logstash/agent_spec.rb
+++ b/logstash-core/spec/logstash/agent_spec.rb
@@ -8,6 +8,7 @@ require_relative "../support/mocks_classes"
 require "fileutils"
 require_relative "../support/helpers"
 require_relative "../support/matchers"
+require 'timeout'
 
 describe LogStash::Agent do
   let(:agent_settings) { mock_settings({}) }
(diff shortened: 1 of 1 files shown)


jordansissel:
Timeout.timeout has historically had some problems:

* http://blog.headius.com/2008/02/rubys-threadraise-threadkill-timeoutrb.html
* http://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/

These problems may not be relevant to this PR's use of it, though, I haven't evaluated fully. Given we are mostly just waiting for a condition, Timeout.timeout might be safe. Something to keep in mind.

jakelandis:
@jordansissel - thanks for the warning, i was not aware. 

In this case it is just preventing a potential infinite loop (actually had one in a Jenkins test). In the context of this PR I believe it to be a harmless safety-net.

jakelandis:
@jordansissel - are you comfortable with these changes ?

jordansissel:
I would prefer a something similar to `Stud.try` where something is tried for some amount of time (to avoid TImeout.timeout and also to avoid the weird combination of `Timeout.timeout(...) { sleep(...) while ... }`

That said, this PR is ok to continue

PR #7769, opened 2017-07-22
MINOR: Cleanup Event by making it final and removing the incorrect serializable interface

original-brownbear:
* We don't extend `Event` anywhere
   * We can save the JIT some trouble here by making it `final`, also saves us some random code around the `clone()` method which now won't throw
* `Event` is not `Serializable` in the sense of that interface, dropped the interface to save us a compiler warning for the missing serialization version UID


suyograo:
LGTM

original-brownbear:
for me: needs backport to 5.x

PR #7771, opened 2017-07-23
MINOR: Inline hot path code in Accessors

original-brownbear:
**For the Reviewer:** This is a highly trivial change, only moving around some code blocks here in the `set()`/`store()` logic, so really we only have the `includes` change here.

* `private static Object store(Object target, String key, Object value)` simply gets inlined into its only user (this method shows up in the compile logs as too large to be inlined, so this is a trivial win imo)
* Interestingly enough `public boolean includes(String reference)` also shows up in the logs as not being inlined, seems this was due to the `Exception` handling in the `List` part of the method => split it up to separate out the larger and colder `List` path.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core/src/main/java/org/logstash/Accessors.java
@@ -1,8 +1,8 @@
 package org.logstash;
 
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

PR #7773, opened 2017-07-23
MINOR: Remove locking around simple getter

original-brownbear:
Not sure why we're locking around a simple getter here, this doesn't seem to do anything -> removing it.


original-brownbear:
nevermind, my bad :)

Issue #7712, opened 2017-07-17
BiValue is redundant


original-brownbear:
With #7663 merged in it seems that all instances of `BiValue` have become in principle redundant.

All other `BiValue` wrappers were wrapping immutable  (in the sense of the specific underlying Java object they point to) instances of `IRubyObject`.

Classes like `org.jruby.RubyFloat` do not need this wrapper, they are themselves a wrapper for the Java object/value already.

In the example of `org.jruby.RubyFloat` there is already a field:

```java
    private final double value;
```

and an accessor:

```java
    /** Getter for property value.
     * @return Value of property value.
     */
    public double getValue() {
        return this.value;
    }
```

available. All our wrapper does here is potentially messing with the JIT's ability to inline things in our hot path (mind that our stacks are already very deep from JRuby itself) and create at least one extra object per `Event` field. 
With `StringBiValue` not caching the Java `String` anymore for reasons laid out in #7662 it seems reasonable to simply drop the `BiValue` path.

This will require however implementing the `SerializationProxy` logic that comes with `BiValueCommon` in some other way that can directly work with the Ruby types (it should be fairly trivial to make the CBOR and JSON serializers work with the Ruby objects, given that all the Ruby types we're dealing with here are in fact properly `Serializable` already).

Thoughts? :)

original-brownbear:
@guyboertje @jakelandis @robbavey just to put my thought of dropping `BiValueCommon` entirely now into writing.
In theory, this seems like a straightforward win, in practice, this is probably a somewhat large change.

original-brownbear:
all use of `BiValue` was removed from Logstash :)

PR #7774, opened 2017-07-23
#7712 Simplify BiValue Serialization

original-brownbear:
This looks like a reasonable step/simplification on the road to removing the `BiValue` overhead for #7712 

* Stop using `SerializationProxy`, it's just overhead since we don't have any `final` fields
   * Should make PQ faster in some spots + saves a ton of memory in all situations
   * Made all fields `transient` to prevent random mistakes in the implementation
   * Saves a lot of code since we can inherit the serialization overrides from the abstract base class
* Add a serialization UID to each type to remove the `javac` warning
* Make each of these `final` to et some clarity, it's bad enough we inherit the serialization approach in the first place :)
* No need for the dead private constructors now => removed

### DIFF
--- a/logstash-core/src/main/java/org/logstash/bivalues/BiValueCommon.java
+++ b/logstash-core/src/main/java/org/logstash/bivalues/BiValueCommon.java
@@ -1,17 +1,19 @@
 package org.logstash.bivalues;
 
 import com.fasterxml.jackson.annotation.JsonValue;
+import java.io.IOException;
+import java.io.ObjectInputStream;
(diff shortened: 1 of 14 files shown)


original-brownbear:
link #7701

guyboertje:
@jordansissel - over to you

original-brownbear:
I think we can save the noise of this one by simply ditching all `BiValue` following up on #7854.
No point in basing this one now, just makes everything harder to follow => closing here

PR #7782, opened 2017-07-24
MINOR: Remove unused take and timed offer from queue wrappers

original-brownbear:
`offer` and `take` aren't used anywhere => remove the dead code.


jordansissel:
Tests pass. LGTM.

PR #7783, opened 2017-07-24
MINOR: Enable JRuby AOT compilation and turn JIT threshold down to 0

original-brownbear:
Makes benchmarking results a lot more deterministic and way reduces the time for the JIT to "converge".

=> I don't see any reason not to turn AOT on by default for production settings.

### DIFF
--- a/config/jvm.options
+++ b/config/jvm.options
@@ -53,6 +53,9 @@
 
 # Turn on JRuby invokedynamic
 -Djruby.compile.invokedynamic=true
+# Force Compilation
+-Djruby.compile.mode=FORCE
+-Djruby.jit.threshold=0
 
 ## heap dumps
 
--- a/qa/integration/specs/reload_config_spec.rb
+++ b/qa/integration/specs/reload_config_spec.rb
@@ -18,7 +18,7 @@ describe "Test Logstash service when config reload is enabled" do
   let(:timeout_seconds) { 5 }
   let(:initial_port) { random_port }
   let(:reload_port) { random_port }
-  let(:retry_attempts) { 10 }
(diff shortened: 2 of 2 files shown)


guyboertje:
Any reason why this can't go in jvm.options?
We put the invoke dynamic in there 9 days ago.
```
# Turn on JRuby invokedynamic
-Djruby.compile.invokedynamic=true
```

original-brownbear:
@guyboertje the JRuby docs don't suggest a Java system property for these two, for the invoke dynamic they did suggest both the JRuby and Java versions.

I can try if the works to use a Java opt, but maybe this isn't such a good idea since it's not documented and could break down the line?

guyboertje:
I suggest you ask on IRC jruby channel. enebo and headius hang out there most days.

original-brownbear:
@guyboertje moved to Java properties, works just fine there as well :)
After some more reading on this, it seems that the `-J-D` is just forwarded to JVM `-D` so this is fine as far as I can tell.

original-brownbear:
@guyboertje can't merge yet :(see https://travis-ci.org/elastic/logstash/jobs/256950889#L4248, will look into this after dinner. Maybe (and hopefully) the slight increase in startup times here is to blame for this one timing out.

guyboertje:
Yeah, the FORCE (haha) will make the start times longer, in my experience.

original-brownbear:
@guyboertje  :D let's see if it's fine now, increased the timeout from 10s to 60s (10 was kinda short in the first place imo :))

original-brownbear:
works fine at 60s :)

PR #7785, opened 2017-07-24
#7700 ignore ENV JAVA_OPTS in rake tasks

### LINKED ISSUE #7700: Make Rake Command Ignore Environment JAVA_OPTS
I have the following `JAVA_OPTS` loaded in my shell by default:

```sh
-Xms4g -Xmx4g -Djava.net.preferIPv4Stack=true
```

which currently leads to:

```sh
➜  logstash git:(master) rake test:core  
Error occurred during initialization of VM
Initial heap size set to a larger value than the maximum heap size
```

unless I `unset` them first. We shouldn't be using the environment opts in builds, it:

1. Makes the build less portable in general
2. Could lead to hard to debug build problems for some when the resulting issues aren't as straightforward as in my example above

=> We should make sure that all rake invocations run with a pre-defined and known to work `JAVA_OPTS` value and ignore the environment they are running from.

original-brownbear:
fixes #7700, see the issue itself for the reasoning behind this change.

### DIFF
--- a/rakelib/z_rubycheck.rake
+++ b/rakelib/z_rubycheck.rake
@@ -32,6 +32,8 @@ if ENV['USE_RUBY'] != '1'
     # if required at this point system gems can be installed using the system_gem task, for example:
     # Rake::Task["vendor:system_gem"].invoke(jruby, "ffi", "1.9.6")
 
+    # Ignore Environment JAVA_OPTS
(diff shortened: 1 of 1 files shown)


suyograo:
Yeah, this confused me a lot. There should be one way to set the heap, and it is in `jvm.options`, so LGTM

PR #7787, opened 2017-07-24
[Docs] Add breaking changes for 6.0.0

suyograo:


docs/static/breaking-changes.asciidoc
```diff
@@ -1,310 +1,38 @@
...
 
 [float]
-==== URL Changes for DEB/RPM Packages
-
-The previous `packages.elastic.co` URL has been altered to `artifacts.elastic.co`.
-Ensure you update your repository files before running the upgrade process, or
-your operating system may not see the new packages.
+==== RPM/Deb package changes
 
+* For `rpm` and `deb` release artifacts, Logstash only globs config files matching `*.conf` pattern in the `conf.d` folder.
```
dedemorton: I'm not sure that "globs" is a valid verb. Also, the statement is a grammatically ambiguous. Makes it sound like the only glob pattern that Logstash will match is *.conf in the conf.d folder. Assuming my understanding is correct, how about replacing this sentence and the one that follows with:  "For `rpm` and `deb` release artifacts, config files that match the `*.conf` glob pattern must be in the `conf.d` folder, or the files will not be loaded."

docs/static/breaking-changes.asciidoc
```diff
@@ -1,310 +1,38 @@
...
-plugin named “my-plugin”, it’s as simple as running:
+==== Command Line Interface behavior
 
-[source,bash]
-----------------------------------
-bin/logstash-plugin install my-plugin
-----------------------------------
+* The `-e` and `-f` CLI options are now mutually exclusive. This also applies to the corresponding long form options `config.string` and 
+  `path.config`. This means config provided via `-e` will no longer be appended to the config provided via `-f`.
+* Configuration provided with `-e` or `config.string` will not be appended with `stdin` input and `stdout` output automatically.
```
dedemorton: I'd say "appended to" instead of "appended with"

docs/static/breaking-changes.asciidoc
```diff
@@ -1,310 +1,38 @@
...
-As with earlier releases, the updated script allows both online and offline plugin installation. For example, to install a
-plugin named “my-plugin”, it’s as simple as running:
+==== Command Line Interface behavior
 
-[source,bash]
-----------------------------------
-bin/logstash-plugin install my-plugin
-----------------------------------
+* The `-e` and `-f` CLI options are now mutually exclusive. This also applies to the corresponding long form options `config.string` and 
+  `path.config`. This means config provided via `-e` will no longer be appended to the config provided via `-f`.
```
dedemorton: Feels like a word is missing here. Maybe say: "This means that any configurations provided via `-e` will no longer be appended to the configuration provided via `-f`."

docs/static/breaking-changes.asciidoc
```diff
@@ -1,310 +1,38 @@
...
-sure that it supports Logstash 5.0 before upgrading!
-
-[float]
-==== Logstash with All Plugins Download
-
-The Logstash All Plugins download option has been removed. For users previously using this option as a convenience for
-offline plugin management purposes (air-gapped environments), please see the <<offline-plugins>> documentation page.
-
-There were 17 plugins removed from 5.0 default bundle. These plugins can still be installed manually for use:
+The following plugins were removed from 5.0 default bundle based on usage data collected by us. These plugins can still be installed
```
dedemorton: Suggest changing to say:  "The following plugins were removed from the 5.0 default bundle based on usage data. You can still install these plugins manually:"

docs/static/breaking-changes.asciidoc
suyograo: Actually, I have it wrong here. This should be `-f` and not `-e`.

@dedemorton but yes, if you provide a config with `-f`  or `-e`, we used to append it with `stdin` and `stdout`. We now only do it for `-e`.

suyograo: So this makes sense then?

> Configurations provided with `-f` or `config.path` will not be appended with `stdin` input and `stdout` output automatically.

suyograo:
@dedemorton updated.

PR #7789, opened 2017-07-24
Fix DLQ seek by Timestamp bugs

robbavey:
Fixes 2 seekByTimestamp failure scenarios that occur when a segment is closed with entries smaller than BLOCK_SIZE

If the next segment is > BLOCK_SIZE, then the seek will advance to the next segment. (Skipping past all events
 in the closed segment)
If the next segment < BLOCK_SIZE, then the seek will advance to the start of the small segment. (Processing
 entries prior to the timestamp)

### DIFF
--- a/logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueReaderTest.java
+++ b/logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueReaderTest.java
@@ -29,6 +29,7 @@ import org.logstash.Event;
 import org.logstash.Timestamp;
 import org.logstash.ackedqueue.StringElement;
 
+import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Collections;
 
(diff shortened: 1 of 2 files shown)


logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueReaderTest.java
```diff
@@ -134,9 +134,63 @@ public void testSeek() throws Exception {
         assertThat(entry.getReason(), equalTo("543"));
     }
 
+
     @Test
-    public void testInvalidDirectory()  throws Exception {
-        DeadLetterQueueReader reader = new DeadLetterQueueReader(dir);
-        assertThat(reader.pollEntry(100), is(nullValue()));
+    public void testWriteStopSmallWriteSeekByTimestamp() throws Exception {
+        DeadLetterQueueWriter writeManager = new DeadLetterQueueWriter(dir, 10000000, 10000000);
```
original-brownbear: @robbavey this thing is a `Closeable` now => we should be able to `try-with-resources` here :)

logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueReaderTest.java
```diff
@@ -134,9 +134,63 @@ public void testSeek() throws Exception {
...
+        long currentEpoch = System.currentTimeMillis();
+
+        for (int i = 0; i < 100; i++) {
+            DLQEntry entry = new DLQEntry(event, "foo", "bar", String.valueOf(i), new Timestamp(currentEpoch++));
+            writeManager.writeEntry(entry);
+            target = entry.getEntryTime();
+        }
+        writeManager.close();
+
+        writeManager = new DeadLetterQueueWriter(dir, 10000000, 10000000);
```
original-brownbear: @robbavey this thing is a `Closeable` now => we should be able to `try-with-resources` here :)

logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueReaderTest.java
```diff
@@ -134,9 +134,63 @@ public void testSeek() throws Exception {
...
+
+        for (int i = 200; i < 300; i++){
+            DLQEntry entry = new DLQEntry(event, "foo", "bar", String.valueOf(i), new Timestamp(currentEpoch++));
+            writeManager.writeEntry(entry);
+        }
+        writeManager.close();
+
+        DeadLetterQueueReader readManager = new DeadLetterQueueReader(dir);
```
original-brownbear: @robbavey this thing is a `Closeable` now => we should be able to `try-with-resources` here :) + we should be closing it anyways

logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueReaderTest.java
```diff
@@ -134,9 +134,63 @@ public void testSeek() throws Exception {
...
 
+
     @Test
-    public void testInvalidDirectory()  throws Exception {
-        DeadLetterQueueReader reader = new DeadLetterQueueReader(dir);
-        assertThat(reader.pollEntry(100), is(nullValue()));
+    public void testWriteStopSmallWriteSeekByTimestamp() throws Exception {
+        DeadLetterQueueWriter writeManager = new DeadLetterQueueWriter(dir, 10000000, 10000000);
+        Event event = new Event(Collections.emptyMap());
+        Timestamp target = null;
```
original-brownbear: @robbavey can't we just move this below the `currentEpoch` line and do ?:

```java
final target = new Timestamp(currentEpoch + 100)
```

... instead of the redundant reassignment of `target` in the loop below.

(also we could extract the `100` to a variable then and make things a little easier to follow)

logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueReaderTest.java
robbavey: Ah nice - I missed that. I'll rework to incorporate that

original-brownbear:
@robbavey looks really nice now :) 
let's just wait for Travis
LGTM!

Issue #7788, opened 2017-07-24
POJO-ification of Metrics


andrewvc:
This issue sets out to clearly define our goals for the refactor of Logstash metrics. It takes over from #7274 .

Our current metrics implementation provides more than we need and has some performance problems, like #7772 to boot. Additionally, the design decision to use `Map` objects to store the hierarchy of metrics has resulted in some performance issues as well as prevented us from leaning on the JVM type checker to verify that metric accessess / stores are correct.

After numerous discussions with @jakelandis @ph and @jsvd we've decided to move toward a simpler POJO based architecture, where we will use the new `Gauge` and `Counter` types directly.

So, instead of ruby code like:

```ruby
class Foo
  def initialize(external_metric_thing)
    @metrics = external_metric_thing.namespace(:foo)
  end

  def receivedEvent
    @metric.increment(:events_received, 1)
  end
end
```

We will write something more like:

```ruby
class Foo
  def initialize
    self.metrics = FooMetrics.new()
  end

  def receivedEvent
    self.fooMetrics.receivedEvent()
  end
end
```

```java
class FooMetrics {
  public FooMetrics() {
    this.receivedEvents = new Counter()
  }

  public void receivedEvent() {
    this.receivedEvents.increment()
  }
}
```

These java metrics classes will be much more complicated in the future, and might increment multiple metrics based on a single lifecycle event, and even increment metrics in linked objects.

jakelandis:
I tried out a couple of a designs and landed on implementing a fluent API to represent the metrics in Java.  This is similar in spirit to what @andrewvc proposes above, but with more detail about the specific implementation and the migration strategy below. 

Fluent API's favor readability and allow for developers to express domain knowledge using a meaning vocabulary.  To avoid re-architecting the entire metrics subsystem, the first iteration of this new fluent API will closely mirror the existing metrics namespace design.

The existing namespace design is an array of Ruby symbols that lead up to a key/value pair (the metric itself):

```
[:symbolA]->[:symbolB]->[:symbolC]->[:symbolD]->[k,v]
```
for example:
```
metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :plugins, plugin_id.to_sym]).increment(:in, 2)
```

[...]

PR #7798, opened 2017-07-25
Performance: Use RubyArray.hash for metric's fast lookup key

### LINKED ISSUE #7772: Metrics "fastlookup" is Using RubyArray as Hash Map Key
I think this is one of if not the main issue behind #7396 

Look at the profile of just running `-e 'input { stdin{}} output {stdout {codec => "dots"}}'` with some large file piped to standard input.

screen shot 2017-07-23 at 10 28 00

We're wasting an absurd amount of time looking up from the `@fast_lookup` in `metric_store.rb`.

The reason for this is trivial, this is what we use for map keys:

`namespaces.dup << key`

`namespaces` is an array and hence the whole hashmap key is an array.
This wouldn't even work with a Java hash, but (un)fortunately JRuby's `RubyArray` uses this method as the override for `equals()`:

```rb
    /** rb_ary_equal
     *
     */
    @JRubyMethod(name = "==", required = 1)
    @Override
    public IRubyObject op_equal(ThreadContext context, IRubyObject obj) {
        Ruby runtime = context.runtime;

        if (this == obj) {
            return runtime.getTrue();
        }

        if (!(obj instanceof RubyArray)) {
            if (obj == context.nil) return runtime.getFalse();

            if (!sites(context).respond_to_to_ary.respondsTo(context, obj, obj)) {
                return runtime.getFalse();
            }
            return Helpers.rbEqual(context, obj, this);
        }
        return RecursiveComparator.compare(context, sites(context).op_equal, this, obj);
    }
```

... which seems to be absurdly expensive (kind of expected, it eventually just compares all array elements).
Also literally every call to this thing requires creating at least one new `RubyArray` instance!

The solution here is obviously to:

* Use an immutable hash key with a cheap and evenly distributed hashcode as well as a fast `equals()` method (like `java.lang.String`)
* Use a key approach that doesn't require creating new objects on lookup

jakelandis:
The existing implementation uses the RubyArray as the key for the fast lookup Map. Under the covers the Map implementation is comparing equal operators (many times per event), and JRuby Array equals operator is VERY expensive since it literally walks each value of array for each equality. Based on profiling via YourKit, the JRuby Array equals operator is a very hot method consuming upto 60% of sampled CPU cycles while under high load (and no other CPU dominators).

This change is to use the .hash value of the JRuby Array as the key of the fast lookup Map. This implementation still calls .hash for each and every call, which is also expensive, since it also walks the arrays to compute the hash. However, the equality check of the hash value is very fast, and net gain is significant. Upto a 15% increase of throughput.

Fixes #7772

### DIFF
--- a/logstash-core/lib/logstash/instrument/metric_store.rb
+++ b/logstash-core/lib/logstash/instrument/metric_store.rb
@@ -51,12 +51,14 @@ module LogStash module Instrument
       # BUT. If the value is not present in the `@fast_lookup` the value will be inserted and we assume that we don't
       # have it in the `@metric_store` for structured search so we add it there too.
 
-      value = @fast_lookup.get(namespaces.dup << key)
+      # array.hash as the key since it is faster then using the array itself, see #7772
+      fast_lookup_key = (namespaces.dup << key).hash
+      value = @fast_lookup.get(fast_lookup_key)
       if value.nil?
         value = block_given? ? yield(key) : default_value
-        @fast_lookup.put(namespaces.dup << key, value)
+        @fast_lookup.put(fast_lookup_key, value)
         @structured_lookup_mutex.synchronize do
-            # If we cannot find the value this mean we need to save it in the store.
+          # If we cannot find the value this mean we need to save it in the store.
           fetch_or_store_namespaces(namespaces).fetch_or_store(key, value)
         end
       end
@@ -163,7 +165,7 @@ module LogStash module Instrument
     end    
 
     def has_metric?(*path)
-      @fast_lookup[path]
+      @fast_lookup[path.hash]
     end
 
     # Return all the individuals Metric,
@@ -185,8 +187,9 @@ module LogStash module Instrument
     def prune(path)
       key_paths = key_paths(path).map(&:to_sym)
       @structured_lookup_mutex.synchronize do
-        keys_to_delete = @fast_lookup.keys.select {|namespace| (key_paths - namespace[0..-2]).empty? }
-        keys_to_delete.each {|k| @fast_lookup.delete(k) }
+        fetch_or_store_namespaces(key_paths).each do |key, v|
+          @fast_lookup.delete((key_paths.dup << key).hash)
+        end
         delete_from_map(@store, key_paths)
       end
     end


jakelandis:
This is probably as good as going to get until we move away from the keystore all together via #7788

jakelandis:
Jenkins test failure is un-related: https://github.com/elastic/logstash/issues/7799

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -51,12 +51,12 @@ def fetch_or_store(namespaces, key, default_value = nil)
       # BUT. If the value is not present in the `@fast_lookup` the value will be inserted and we assume that we don't
       # have it in the `@metric_store` for structured search so we add it there too.
 
-      value = @fast_lookup.get(namespaces.dup << key)
+      value = @fast_lookup.get((namespaces.dup << key).hash)
```
jordansissel: computing `(namespaces.dup << key)` twice seems odd.

Is namespaces a string? If so, we can just do `(namespaces + key)` and possibly save this to a variable?

The reason I mention is that months from now I wonder if we will look at this line of code and wonder why it was written this way.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -185,8 +185,9 @@ def each(path = nil, &block)
     def prune(path)
       key_paths = key_paths(path).map(&:to_sym)
       @structured_lookup_mutex.synchronize do
-        keys_to_delete = @fast_lookup.keys.select {|namespace| (key_paths - namespace[0..-2]).empty? }
-        keys_to_delete.each {|k| @fast_lookup.delete(k) }
+        fetch_or_store_namespaces(key_paths).each do |key, v|
+          @fast_lookup.delete((key_paths.dup << key).hash)
```
jordansissel: similar comments as above.

logstash-core/lib/logstash/instrument/metric_store.rb
```diff
@@ -51,12 +51,12 @@ def fetch_or_store(namespaces, key, default_value = nil)
       # BUT. If the value is not present in the `@fast_lookup` the value will be inserted and we assume that we don't
       # have it in the `@metric_store` for structured search so we add it there too.
 
-      value = @fast_lookup.get(namespaces.dup << key)
+      value = @fast_lookup.get((namespaces.dup << key).hash)
       if value.nil?
         value = block_given? ? yield(key) : default_value
-        @fast_lookup.put(namespaces.dup << key, value)
+        @fast_lookup.put((namespaces.dup << key).hash, value)
```
jordansissel: similar comment as above

jordansissel:
Tests are passing (Ooooh jeenkinssssssss), but I have style/readability concerns I'd like resolved first.

Since we're explicitly calling `.hash` maybe in addition can you add a comment that explains the reason for calling .hash explicitly?

logstash-core/lib/logstash/instrument/metric_store.rb
original-brownbear: @jakelandis @jordansissel the hash calculated here is `Murmur` as far as I can tell from the code of `RubyArray`:

```
    /** rb_ary_hash
     *
     */
    @JRubyMethod(name = "hash")
    public RubyFixnum hash(ThreadContext context) {
        Ruby runtime = context.runtime;

        long h = Helpers.hashStart(runtime, realLength);

        h = Helpers.murmurCombine(h, System.identityHashCode(RubyArray.class));

        for (int i = 0; i < realLength; i++) {
            IRubyObject value = eltOk(i);
            RubyFixnum n = Helpers.safeHash(context, value);
            h = murmurCombine(h, n.getLongValue());
        }

        h = hashEnd(h);

        return runtime.newFixnum(h);
    }
```

Are you sure this is safe (enough)? (I would instinctively say no tbh)

original-brownbear: @jordansissel `namespaces` is an array, otherwise we'd have a trivial situation here :P

original-brownbear:
@jakelandis @jordansissel see my comment above, this doesn't look correct to me, but the approach makes sense in general. Maybe we could do this instead:

* Use a safe hash function (md5 would be my suggestion here as "safe enough")
* simply run each array element one by one through a threadlocal, reusable md5digest to calculate the hash => no need to setup a new array copy on each lookup

WDYT?

logstash-core/lib/logstash/instrument/metric_store.rb
jakelandis: @jordansissel 
> computing (namespaces.dup << key) twice seems odd.

Agreed, but it is pre-existing art. I assume it is because we can call 'get' many times per event (e.g. [m|b|tr]illions total), but on the put, we only call it for new metrics (e.g. dozens total).  I can change it, I don't see a reason why not. However, some diciplines prescribe limiting the scope of the change to ONLY that which is relevant to the problem. (e.g. keep the change set pure). I am unsure the guiding dicipline here.  

> Is namespaces a string? 

Nope, an array. The first thing I tried was to convert the key to a String. While it fixed the hot method in YourKit, it increased the wall time by over 10%. 

>  I wonder if we will look at this line of code and wonder why it was written this way. 

I tried to explained in the commit log, but I can add comment to the code too.  

@original-brownbear 
> Are you sure this is safe (enough)? 

I am never sure about any anything 😄 , but I believe this is safe enough. If I am reading this correctly it is creating a murmur hash from the hash values of the contents of the array. Since we only have a few dozen here i assume the risk of collision is very low. (is that your concern?)

jakelandis:
@jordansissel 
> Since we're explicitly calling .hash maybe in addition can you add a comment that explains the reason for calling .hash explicitly?

will do.

@original-brownbear 
> Maybe we could do this instead

I belive that the .hash operation is correct, though perhaps not the ideal performance. Assuming we agree on the correctness, the stragetic performance change is in the works, and hopefully this is just a band-aid till then.

logstash-core/lib/logstash/instrument/metric_store.rb
original-brownbear: @jakelandis 

>  but I believe this is safe enough

Phew :D based on what? :D Tbh I have no way to estimate this either, but this seems like a potential source for very strange bugs.
Frankly, I don't have anything better on this, but imo it's probably worth going with a strong hash function and a trie type hash here since we're not under a lot of time pressure (+ this shouldn't be that hard to implement imo). 
But ... your call, my argument sucks without math to back it up and I admit it :)

jakelandis:
Updated with requested changes: https://github.com/elastic/logstash/pull/7798/commits/39fc4eee9ae421cb8b9fefaff3576926540cce09

(will squash prior to merge)

logstash-core/lib/logstash/instrument/metric_store.rb
jakelandis: @original-brownbear - I am not sure I understand your concern ?  Is it collision/uniqueness guarantees ?

original-brownbear: @jakelandis yes I mean the hash itself is 32 bit here it seems and chances are we're safe, but there could be effects that make collisions very likely as a result of JRuby's implementation :)
http://preshing.com/20110504/hash-collision-probabilities/ for math, but note that we likely don't have 32 clean bits in our hashes here :)

original-brownbear: @jakelandis One more thought: even if the individual collision likelihood is small, this is a repeat run game with every run of LS for every user ...

=> at 30 values, you have a chance of a collision of 5% with 100k LS runs ... so this is effectively guaranteed to affect people randomly

jakelandis: @original-brownbear - Is this effectively stating that JRuby's Array.hash is broken ?

I am not sure I agree with the math of ```5%```,  that seems really high. If these hashes were persisted across 100k runs maybe.  I haven't done the math, but it also seems that to do the math would require the exact algorithm, not just the bit space. 

I am not concerned about chances of collision between a couple dozen values using a well established hashing algorithm.

original-brownbear: @jakelandis why would that mean the hash is broken, this is not a cryptographic hash and just a 32 bit value :) Whether this matters in practice is a very different question though (if you think about it even if this happens every now and then, the chance of anyone noticing is close to`0`) :)

original-brownbear: @jakelandis simulation :D 

```java
    @Test
    public void murmurCollisions() {
        int i = 0;
        final Random random = new SecureRandom();
        final int count = 60;
        final Set<Integer> hashes = new HashSet<>(count * 2);
        while (true) {
            for (int j = 0; j < count; ++j) {
                hashes.add(random.nextInt());
            }
            if (hashes.size() < count) {
                System.out.println("collision");
            }
            hashes.clear();
            ++i;
            if (i % 1_000_000 == 0) {
                System.out.println("million");
                i = 0;
            }
        }
    }
```

(best implementation I could code in 2 min :D) 

Prints:

```sh
million
collision
million
million
million
million
```

It's just the law of large numbers at play here

original-brownbear:
I'm good with the approach for reasons discussed in Slack btw, despite my complaints :)

jakelandis:
@jordansissel - are you comfortable with this change ?

(I will rebase, squash, and wait for the builds prior to merge)

jordansissel:
LGTM

jakelandis:
Jenkins test failure is not related : https://github.com/elastic/logstash/issues/7842

suyograo:
@jakelandis should we backport this to 5.6? Any reason not to?

jakelandis:
@suyograo - sounds good.  Just waiting for the build to pass: https://github.com/elastic/logstash/pull/7867

Issue #7799, opened 2017-07-25
Windows Java Test failure


jakelandis:
It seems that the benchmark-cli tests are having issues on Windows.  Note, the Java side of tests worked on Windows prior. 

cc: @original-brownbear 
```
org.logstash.benchmark.cli.LsMetricsMonitorTest > parsesCpuUsage FAILED

    java.nio.file.InvalidPathException: Illegal char <:> at index 2: /C:/Users/jenkins/workspace/elastic+logstash+pull-request+multijob-windows-compatibility/tools/benchmark-cli/build/resources/test/org/logstash/benchmark/cli/metrics.json
        at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
        at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
        at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
        at java.nio.file.Paths.get(Paths.java:84)
        at org.logstash.benchmark.cli.LsMetricsMonitorTest.parsesCpuUsage(LsMetricsMonitorTest.java:59)

org.logstash.benchmark.cli.LsMetricsMonitorTest > parsesFilteredCount FAILED

    java.nio.file.InvalidPathException: Illegal char <:> at index 2: /C:/Users/jenkins/workspace/elastic+logstash+pull-request+multijob-windows-compatibility/tools/benchmark-cli/build/resources/test/org/logstash/benchmark/cli/metrics.json
        at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
        at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
        at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
        at java.nio.file.Paths.get(Paths.java:84)
        at org.logstash.benchmark.cli.LsMetricsMonitorTest.parsesFilteredCount(LsMetricsMonitorTest.java:35)
```

original-brownbear:
@jakelandis argh yea there's this and even one additional Windows incompatibility in the benchmark code.
[...]

PR #7800, opened 2017-07-25
 #7799 fix Windows incompatible Resource use + Windows incompatible use of /dev/null

original-brownbear:
Obvious fixes:

* `/dev/null` won't work on Windows, used random temp file instead (output size is small enough for this to be fine + Java doesn't really offer us much better than redirecting to a `File` when it comes to Windows anyway as far as I can tell)
* Adjusted the resources read to not work from a `URL` but just read the stream directly
* Banner didn't work on Windows for encoding reasons, removed for now, we can add a portable one at some later date I guess :D 

### Note to the Reviewer

The Windows build still fails :(, but the issue in #7799 is fixed here since the benchmark build passes

### DIFF
--- a/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
+++ b/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
@@ -127,8 +127,9 @@ public interface LogstashInstallation {
             );
             final Path lsbin = location.resolve("bin").resolve("logstash");
             LsBenchFileUtil.ensureExecutable(lsbin.toFile());
+            final File output = Files.createTempFile(null, null).toFile();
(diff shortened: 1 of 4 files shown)


tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
```diff
@@ -127,8 +127,9 @@ public void execute(final String configuration, final File data)
             );
             final Path lsbin = location.resolve("bin").resolve("logstash");
             LsBenchFileUtil.ensureExecutable(lsbin.toFile());
+            final File output = Files.createTempFile(null, null).toFile();
             final Process process = pbuilder.command(lsbin.toString(), "-w", "2", "-f", cfg.toString()).redirectOutput(
-                ProcessBuilder.Redirect.to(new File("/dev/null"))
+                ProcessBuilder.Redirect.to(output)
```
jsvd: I love these little moves towards cross-platform compatibility

Issue #7766, opened 2017-07-21
Test failure 5.x amazon


jsvd:
Failure in 5.x https://logstash-ci.elastic.co/job/elastic+logstash+5.x+multijob-unix-compatibility/os=amazon/lastBuild/consoleFull

```
Caused by: com.headius.invokebinder.InvalidTransformException: java.lang.IllegalAccessException: no such method: rubyjit.LogStash::Filters::Multiline$$filter_b78cba0c9f1bd982b66c40881ba24cdf9951bf64142257191.__file__(LogStash::Filters::Multiline$$filter_b78cba0c9f1bd982b66c40881ba24cdf9951bf64142257191,ThreadContext,IRubyObject,IRubyObject,Block)IRubyObject/invokeStatic
	at com.headius.invokebinder.Binder.invokeStaticQuiet(Binder.java:1079)
	at org.jruby.runtime.invokedynamic.InvocationLinker.createRubyHandle(InvocationLinker.java:1678)
	... 75 more
Caused by: java.lang.IllegalAccessException: no such method: rubyjit.LogStash::Filters::Multiline$$filter_b78cba0c9f1bd982b66c40881ba24cdf9951bf64142257191.__file__(LogStash::Filters::Multiline$$filter_b78cba0c9f1bd982b66c40881ba24cdf9951bf64142257191,ThreadContext,IRubyObject,IRubyObject,Block)IRubyObject/invokeStatic
	at java.lang.invoke.MemberName.makeAccessException(MemberName.java:867)
	at java.lang.invoke.MemberName$Factory.resolveOrFail(MemberName.java:1003)
	at java.lang.invoke.MethodHandles$Lookup.resolveOrFail(MethodHandles.java:1386)
	at java.lang.invoke.MethodHandles$Lookup.findStatic(MethodHandles.java:780)
	at com.headius.invokebinder.Binder.invokeStatic(Binder.java:1055)
	at com.headius.invokebinder.Binder.invokeStaticQuiet(Binder.java:1077)
	... 76 more
[...]

PR #7805, opened 2017-07-25
 #7766 turn off invokedynmic

original-brownbear:
as discussed in #7766


suyograo:
LGTM

original-brownbear:
@suyograo merging then, since this has no relation to the tests.

PR #7810, opened 2017-07-25
handle fully acked zero byte page at queue open

### LINKED ISSUE #7809: fully acked tail page of zero byte size prevents Queue open
We have witnessed  a few reports of logstash unable to restart with this error:

```
Logstash failed to create queue {"exception"=>"Page file size 0 different to configured page capacity xxx
```

In all diagnosed cases, the empty page file (of zero byte) had a corresponding checkpoint which indicated the page as being fully acked.

It is not clear how or why this is happening, but upon restart and reading the queue data, LS should be able to simply ignore these pages since they are fully acked and can/shoud be purged anyway.

colinsurprenant:
fixes #7809 

This fixes the condition upon opening the queue, a page file of zero byte with a fully acked checkpoint would generate this exception:

```
Logstash failed to create queue {"exception"=>"Page file size 0 different to configured page capacity xxx
```

I also added a test to reproduce that problem. Without this fix the test fails with the above exception and passes with this PR.

This condition should not prevent LS from starting since fully acked page are simply purged in the open process. Here we just do the purge before checking on the file size. 

Now, the condition which leads to having a zero byte page file is still unknown. 

followup issue #7811 to DRY the code. This seems to be an acceptable fix for now.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
@@ -38,6 +38,8 @@ public class Batch implements Closeable {
         return elements;
     }
 
+    public List<Long> getSeqNums() { return this.seqNums; }
+
     public Queue getQueue() {
         return queue;
     }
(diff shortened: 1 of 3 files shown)


colinsurprenant:
The jenkins build failure is about DLQ which seems unrelated to this PR.

guyboertje:
I will have to trust the failing and then not failing test and Colin on this one.

guyboertje:
@suyograo @jordansissel 

Second review?

jordansissel:
Build failures seem unrelated. LGTM

suyograo:
Merging since Colin is on vacation.

colinsurprenant:
This PR commit was not merged into 5.6.  I just correctly added it in 5.6 and will be part of 5.6.5.

PR #7815, opened 2017-07-26
PERFORMANCE: Extracted Cold paths in BiValues and Accessors

original-brownbear:
Standard improvements:

* Extract cold cache branches to separate methods where possible (then the JIT can inline the much more likely hot cache path in more instances than when it has to also inline the cold path to where it's never used)
* Move `null` checks up the stack in `Valuefier` etc. The `null` path showed with a likelihood of about `50%` across multiple config examples I tested and the redundant `null` check introduced in the other half of cases is way cheaper than a few `instanceof` and a subsequent method call now saved in the `null` case

Also:

* corrected naming of `NAME_CACHE` since it's a constant
* Removed redundant `public static FieldReference cache(String reference, FieldReference field)` because the return value was never used and hence it was simply a wrapper for `.put()` making JIT logs harder to read with the other overload of `cache` being super hot

... that's it for the obvious things that could be found using JitWatch I think :)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core/src/main/java/org/logstash/Accessors.java
@@ -91,22 +91,19 @@ public class Accessors {
     }
 
     private Object findTarget(FieldReference field) {
-        Object target;
+        final Object target = this.lut.get(field.getReference());
+        return target != null ? target : cacheTarget(field);
+    }
 
-        if ((target = this.lut.get(field.getReference())) != null) {
-            return target;
-        }
-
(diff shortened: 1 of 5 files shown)


guyboertje:
I agree with your findings, however there is a serious bug in the Accessors class. Found after investigating https://github.com/logstash-plugins/logstash-filter-geoip/issues/125.
I will create an issue and we should discuss therein.

original-brownbear:
@guyboertje thanks + sure let's discuss the bug in another issue :)

Issue #7784, opened 2017-07-24
Benchmark Tool - Enhancements - Round 1


jakelandis:
First round of enhancements to the Logstash Benchmark Tool  
- [ ]     Add in some hardware stats to the output (Platform/CPU/Memory, etc) , to help contextualize the result if shared between people.
- [ ]      Add a backgroud thread to add progress dots to let you know it is working, the apache test case takes a while and some feed back helps (kinda like M$'s progress bar, it means nothing, but makes you feel better).
- [ ]      Add in an option for --testcase=custom and expose a --conifg-dir option such that, assuming I can figure out how to feed data in, I can use this for any test case.
- [ ]      Add an option to store a csv file of the results, so that they can be source controlled, and graphed over time.
- [ ]      Add in a clean command to clean up the $HOME/.logstash-benchmarks directory.

The above is not an exhaustive list, and is simply intended to be a place to gather thoughts for the next time someone has the time to invest in the Benchmark Tool.

original-brownbear:
Just adding a few notes:

* Add the ability to loop initially and only start measuring once the JIT has settled down
* Make the length of a run configurable and warn on runs that were `<5 min` because the result is likely meaningless
* Make worker-count + other batch related params + PQ true/false configurable when invoking the benchmark
  * Same goes for JVM settings

suyograo:
Few from my side:

- [ ]  Consolidate the various `benchmarks` directories. Discuss if the benchmark tool needs to be split into a separate repo.
- [ ]  Expose a shell wrapper to call the Java benchmarking tool instead of users having to do `java -cp`
- [ ]  Big +1 on the ability to benchmark custom configs using custom dataset. This will be wonderful!
- [ ] Ability to export bench runs and ship to ES directly (so we can graph in Kibana)

original-brownbear:
@suyograo @jakelandis soo :) question: What do we want first?

From my side, the Kibana/ES integration first makes some sense to finish up on the end-to-end functionality (it's kinda trivial too)?

suyograo:
@original-brownbear +1 to ES exporter. I would tackle the custom configs next, after that.

kunisen:
It looks like so far we don't have any reference data published for the benchmark performance test result. 
The user wants us to provide a test result as the reference.
Would you please add it to the enhancement items?

original-brownbear:
[...]

PR #7816, opened 2017-07-26
 #7784 Cleanup Metric Collection in Benchmarks

original-brownbear:
Just some cleanup of compiler warnings and duplication to prepare for introducing optional ES output in #7784.

### DIFF
--- a/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LsMetricsMonitor.java
+++ b/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LsMetricsMonitor.java
@@ -1,5 +1,6 @@
 package org.logstash.benchmark.cli;
 
+import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


original-brownbear:
@jakelandis @suyograo so I was going to implement the ES output real quick this morning, but I wanted to run an idea by you guys first (PR here is just some cleanup to be able to implement this in 30min).

How about this for an instant win on ES integration + gathering more stats:

* just add an optional flag `--elasticsearch=127.0.0.1:9200`
   * If the flag is set, we (in addition to collecting the few quick metrics we currently extract from the JSON) simply start appending all raw JSON strings we get to a log, enriching each entry with the timestamp of its collection and the timestamp of the start of the benchmark run.
* After the run is over, we simply ship the raw JSON metric values straight to ES.

Then we can do whatever analysis we want on them right away and have all the other stats about GC and memory and all that in a nice time-series.
There isn't really much point in implementing further analysis in this tool then is there, much better to leverage ES here right?

suyograo:
> just add an optional flag --elasticsearch=127.0.0.1:9200
If the flag is set, we (in addition to collecting the few quick metrics we currently extract from the JSON) simply start appending all raw JSON strings we get to a log, enriching each entry with the timestamp of its collection and the timestamp of the start of the benchmark run.

@original-brownbear this sounds good to me. Nit pick: I would make the flag `--elasticsearch-export`=http://127.0.0.1:9200` :) Make sure you can take the full URL here with basic auth as well..

Also, the only system stat that would be really useful in the terminal output would be CPU usage. Anything else can go in the advanced stat collection which as you suggested is enabled by default when `--elasticsearch-export` is used.

suyograo:
LGTM.

original-brownbear:
@suyograo alright sounds good, merging this one then and implementing in the next PR (probably later today :))

PR #7819, opened 2017-07-26
Add a sleep to attempt to deal with Windows build failure

robbavey:


robbavey:
Didn't work, an unclosed channel was the problem

PR #7821, opened 2017-07-26
PERFORMANCE: Optimize tagging API paths

original-brownbear:
The `tag` calls were still messing with the JIT because `setField()` is just a gigantic (after inlining) method and we don't need it here (we know we want a `ConvertedList` containing `StringBiValue`).

... so extracted all the paths here too.

=> Smallish performance gain
=> Some memory savings
=> A few warnings saved in JitWatch


suyograo:
LGTM

Issue #7807, opened 2017-07-25
Dead Letter Queue Tests fail on Windows


robbavey:
Running the DeadLetterQueue tests on Windows yields the following failures:

```org.logstash.common.io.DeadLetterQueueReaderTest > testWriteStopSmallWriteSeekByTimestamp FAILED

    java.io.IOException: Access is denied

        at java.io.WinNTFileSystem.createFileExclusively(Native Method)

        at java.io.File.createNewFile(File.java:1012)

        at org.logstash.common.io.DeadLetterQueueWriter.<init>(DeadLetterQueueWriter.java:63)

        at org.logstash.common.io.DeadLetterQueueReaderTest.writeEntries(DeadLetterQueueReaderTest.java:170)

        at org.logstash.common.io.DeadLetterQueueReaderTest.testWriteStopSmallWriteSeekByTimestamp(DeadLetterQueueReaderTest.java:137)



org.logstash.common.io.DeadLetterQueueReaderTest > testWriteStopBigWriteSeekByTimestamp FAILED

    java.io.IOException: Access is denied

        at java.io.WinNTFileSystem.createFileExclusively(Native Method)

        at java.io.File.createNewFile(File.java:1012)

        at org.logstash.common.io.DeadLetterQueueWriter.<init>(DeadLetterQueueWriter.java:63)

        at org.logstash.common.io.DeadLetterQueueReaderTest.writeEntries(DeadLetterQueueReaderTest.java:170)

        at org.logstash.common.io.DeadLetterQueueReaderTest.testWriteStopBigWriteSeekByTimestamp(DeadLetterQueueReaderTest.java:153)
```

robbavey:
fixed by #7822

PR #7822, opened 2017-07-26
Close channel associated with lock in DLQ

robbavey:
Fix Windows build issue where .lock file cannot be deleted due to
underling channel associated with lock not being closed. This
will stop the file from being deleted until the JVM exits, despite
reporting through Java that the file *has* been deleted and does not
exist. It also blocks new files from being created with the same
filename, which caused the test failure here.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
@@ -146,17 +146,33 @@ public final class DeadLetterQueueWriter implements Closeable {
 
     @Override
     public synchronized void close() throws IOException {
-        if (this.lock != null){
-            this.lock.release();
-            if (this.lock.channel() != null && this.lock.channel().isOpen()) {
-                this.lock.channel().close();
+        if (currentWriter != null) {
+            try {
+                currentWriter.close();
+                open = false;
+            }catch (Exception e){
+                logger.debug("Unable to close dlq writer", e);
             }
         }
-        if (currentWriter != null) {
-            currentWriter.close();
+        releaseLock();
+    }
+
+    private void releaseLock() {
+        if (this.lock != null){
+            try {
+                this.lock.release();
+                if (this.lock.channel() != null && this.lock.channel().isOpen()) {
+                    this.lock.channel().close();
+                }
+            } catch (Exception e) {
+                logger.debug("Unable to close lock channel", e);
+            }
+            try {
+                Files.deleteIfExists(queuePath.resolve(LOCK_FILE));
+            } catch (IOException e){
+                logger.debug("Unable to delete lock file", e);
+            }
         }
-        Files.deleteIfExists(queuePath.resolve(LOCK_FILE));
-        open = false;
     }
 
     public boolean isOpen() {
--- a/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
@@ -146,7 +146,12 @@ public final class DeadLetterQueueWriter implements Closeable {
 
     @Override
     public synchronized void close() throws IOException {
-        this.lock.release();
+        if (this.lock != null){
+            this.lock.release();
+            if (this.lock.channel() != null && this.lock.channel().isOpen()) {
+                this.lock.channel().close();
+            }
+        }
         if (currentWriter != null) {
             currentWriter.close();
         }


robbavey:
Fixes #7807

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -146,7 +146,12 @@ private void innerWriteEntry(DLQEntry entry) throws IOException {
 
     @Override
     public synchronized void close() throws IOException {
-        this.lock.release();
+        if (this.lock != null){
+            this.lock.release();
+            if (this.lock.channel() != null && this.lock.channel().isOpen()) {
```
original-brownbear: @robbavey this is a little questionable imo, given that we are closing the `currentWriter` below as well.

I think we have `IOUtils` (the Commons IO one) on the classpath already (or something similar) and should probably `IOUtils.closeQuietly()` here (or otherwise suppress any possible exceptions on the channel close) to prevent the close actions below this one from failing.

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -146,7 +146,12 @@ private void innerWriteEntry(DLQEntry entry) throws IOException {
...
+                this.lock.channel().close();
+            }
+        }
         if (currentWriter != null) {
             currentWriter.close();
```
original-brownbear: @robbavey same here, this should also quietly close to not fail on the delete

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -146,7 +146,12 @@ private void innerWriteEntry(DLQEntry entry) throws IOException {
 
     @Override
     public synchronized void close() throws IOException {
```
original-brownbear: @robbavey this is more of a general comment, but this `synchronized` seems very dangerous. If this is actually needed because we could have concurrent `close` calls, it might be safer to simply turn the `open` field into an `AtomicBoolean` and guard by atomic `open.compareAndSet(true, false)` around the whole method instead of `synchronized` to avoid any locking in the exception handling?

original-brownbear:
@robbavey few comments on this thing (I'm aware these are kinda annoying for this little code but I stand by them :D)

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
robbavey: I agree around closeQuietly, but we don't have that on the classpath right now.
I will wrap the lock channel close inside a try... catch to alleviate your fears, but the channel has to be closed otherwise the delete cannot complete on Windows until the JVM exits.

original-brownbear: @robbavey right, that's a good point with Windows ... still, maybe the most likely reason the channel would throw here is that some other mechanism closed it already between the `if` and the `close` call. So I guess just suppressing still has value?

robbavey: btw closeQuietly is deprecated from IOUtils - https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/IOUtils.html#closeQuietly(java.io.Reader)

original-brownbear: @robbavey hah they're strict (and probably right :D).

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
robbavey: I agree that this is something to be looked at, but not as part of this PR. I have more work in this area outstanding in another brach

original-brownbear: sure, hence the "general comment" disclaimer, don't worry about it :)

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -146,12 +147,33 @@ private void innerWriteEntry(DLQEntry entry) throws IOException {
...
+                    this.lock.channel().close();
+                }
+            } catch (Exception e) {
+                logger.debug("Unable to close lock channel", e);
+            }
+            try {
+                Files.deleteIfExists(queuePath.resolve(LOCK_FILE));
+            } catch (IOException e){
+                logger.debug("Unable to delete lock file", e);
+            }
```
original-brownbear: @robbavey LGTM, doesn't get any safer than this :)

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -21,6 +21,7 @@
 import java.io.Closeable;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.core.util.IOUtils;
```
original-brownbear: @robbavey ah this was an accident right, this is a dead line I think?

robbavey: Ha! Yes

robbavey: Auto-import of IOUtils found this and not the apache commons version ;)

robbavey:
@original-brownbear Thanks for the fast, and thorough, review

original-brownbear:
@robbavey np :)

PR #7823, opened 2017-07-26
#7817 Accessors should only handle ConvertedMap and ConvertedList as Collection, non-scalar types

### LINKED ISSUE #7817: Accessors.java has a subtle bug as it back-fills objects when setting a deep FieldReference value.
Found while investigating https://github.com/logstash-plugins/logstash-filter-geoip/issues/125

Imagine a user with new Event uses `add_field` to set this `"[foo][0][bar][1]" => "baz"` - Accessors in `findCreateTarget` will not create the correct kind of parent collections for each intended element in the FieldReference path and key.

At the start of `findCreateTarget` target is the Event data Map (a ConvertedMap, mostly, except in the empty ctor)
Iterating over `"[foo]", "[0]", "[bar]", "[1]"` we do:
```
Iterate over `"[foo]", "[0]", "[bar]", "[1]"` as key
  result = fetch(target, "[foo]") // -> null
  if result is null
    result = new HashMap<String, Object>()
    if target is Map
      target.put key, result
    else if target is List
      i = parseInt(key)
      target.set(i, result)
    else
      exception
    end
  end
  target = result
end
```
Bug no. 1 - HashMap should be a ConvertedMap (I missed this when implementing BiValues)
Bug no. 2 - when result is null we need to create a ConvertedMap if the key looks like a string and a ConvertedList if the key looks like a number.

original-brownbear:
fixes  #7817 

This needs a careful review by someone who understands this code end-to-end.

The problem here was that we implicitly assumed that `Accessors` will only ever handle `ConvertedMap` and `ConvertedList` containing `BiValue`, but this didn't hold true because of the inline `new HashMap<>()` (in a sense `Accessors` should just be a private inner class of `Event` to make really sure of this, but we don't have this right now and it would be a huge change).

=>  in this PR I turned on hard checks for `ConvertedList` and `ConvertedMap` where we previously only had implicit assumptions and adjusted the tests accordingly.

All Ruby side tests pass fine without adjustment which makes me confident that this is a really safe fix + generally improves type safety, but please review carefully, I could have missed something big here obviously :)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core/src/main/java/org/logstash/Accessors.java
@@ -4,12 +4,12 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-public class Accessors {
+class Accessors {
 
-    private Map<String, Object> data;
+    private final ConvertedMap data;
     protected Map<String, Object> lut;
 
-    public Accessors(Map<String, Object> data) {
+    public Accessors(final ConvertedMap data) {
         this.data = data;
         this.lut = new HashMap<>(); // reference -> target LUT
     }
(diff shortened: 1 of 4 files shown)


original-brownbear:
@guyboertje can you take a look at this please?

original-brownbear:
@guyboertje ping (no rush, this doesn't block anything for me, just in case this got lost) :)

original-brownbear:
@guyboertje thanks!

original-brownbear:
Had to rebase Jarvis didn't like it anymore, let's see if the tests are still green :P

original-brownbear:
still works fine, merging :)

suyograo:
@original-brownbear backport to 5.6? Looks like a good fix?

original-brownbear:
@suyograo sure incoming here https://github.com/elastic/logstash/pull/7862

PR #7830, opened 2017-07-27
#7784 Benchmark ES Output

original-brownbear:
For #7784 

Example measurement:

```json
      {
        "_index": "benchmarks",
        "_type": "measurement",
        "_id": "AV2F_J12UU4CBrYGZnqC",
        "_score": 1,
        "_source": {
          "@timestamp": 1501191511344,
          "throughput_mean": 39563,
          "throughput_max": 46750,
          "cpu_usage_mean_percent": 22,
          "throughput_min": 5829,
          "meta": {
            "version_type": "DISTRIBUTION",
            "os_version": "10.12.5",
            "os_name": "Mac OS X",
            "cpu_cores": 8,
            "test_name": "baseline",
            "version": "5.5.0",
            "host_name": "bb-elastic-mac.dev"
          }
      }
```

### DIFF
--- a/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/DataStore.java
+++ b/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/DataStore.java
@@ -1,0 +1,84 @@
+package org.logstash.benchmark.cli;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
+import org.apache.http.HttpHost;
+import org.apache.http.HttpStatus;
+import org.apache.http.entity.ContentType;
+import org.apache.http.nio.entity.NStringEntity;
+import org.elasticsearch.client.RestClient;
+import org.logstash.benchmark.cli.ui.LsMetricStats;
+import org.logstash.benchmark.cli.util.LsBenchJsonUtil;
+import org.openjdk.jmh.util.ListStatistics;
+
+public interface DataStore extends Closeable {
+
+    /**
+     * Dummy {@link DataStore} that does nothing.
+     */
+    DataStore NONE = new DataStore() {
+        @Override
+        public void store(final Map<LsMetricStats, ListStatistics> data) {
+        }
+
+        @Override
+        public void close() {
+        }
+    };
+
+    /**
+     * @param data Measured Data
+     * @throws IOException On Failure
+     */
+    void store(Map<LsMetricStats, ListStatistics> data) throws IOException;
+
+    /**
+     * Datastore backed by Elasticsearch.
+     */
+    final class ElasticSearch implements DataStore {
+
+        /**
+         * Low Level Elasticsearch {@link RestClient}.
+         */
+        private final RestClient client;
+
+        /**
+         * Metadata for the current benchmark run.
+         */
+        private final Map<String, Object> meta;
+
+        /**
+         * Ctor.
+         * @param host Elasticsearch Hostname
+         * @param port Elasticsearch Port
+         * @param schema {@code "http"} or {@code "https"} 
+         * @param meta Metadata
+         */
+        ElasticSearch(final String host, final int port, final String schema,
+            final Map<String, Object> meta) {
+            client = RestClient.builder(new HttpHost(host, port, schema)).build();
+            this.meta = meta;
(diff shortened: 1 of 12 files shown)


suyograo:
@original-brownbear I'm wondering if we really need to store granular metrics in ES. After the benchmarking run, we already display events per sec (throughput), CPU usage etc to console. Why not just push them as a document to ES? I don't know if we'll actively monitor/graph all the stats that you have here. To me, the point of publishing to ES is to have historical data of benchmark runs. Once we index the aggregated values, we can just graph them. if we send granular data, we'd have to write ES aggs via Kibana and remember the time period etc.

How about this for an ES doc?

```json
{
   "@timestamp":1461213093093,
   "test_name":"baseline",
   "throughput_min":27385,
   "throuhput_max":27225,
   "throuhput_mean":27225,
   "cpu_usage_mean_percent":127,
   "meta":{
      "cpu_physical_cores":36,
      "cpu_logical_cores":72,
      "cpu_model":"Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz",
      "os_name":"Linux",
      "os_version":"3.19.0-21-generic",
      "host_name":"foo",
      "git_hash":"a6c0a81",
      "git_tag":"v1234"
      "distribution_version":"5.0.0-SNAPSHOT"
   }
}
```

original-brownbear:
@suyograo my thinking was that tracking the granular stats would allow to see some historic differences in GC and the like over time.
But on second thought ... you're right probably more trouble than it's worth :)
Will adjust this, kind of a trivial change anyway :)

original-brownbear:
@suyograo how about this one (this is what this branch does now):

```json
      {
        "_index": "benchmarks",
        "_type": "measurement",
        "_id": "AV2F_J12UU4CBrYGZnqC",
        "_score": 1,
        "_source": {
          "@timestamp": 1501191511344,
          "throughput_mean": 39563,
          "throughput_max": 46750,
          "cpu_usage_mean_percent": 22,
          "throughput_min": 5829,
          "meta": {
            "version_type": "DISTRIBUTION",
            "os_version": "10.12.5",
            "os_name": "Mac OS X",
            "cpu_cores": 8,
            "test_name": "baseline",
            "version": "5.5.0",
            "host_name": "bb-elastic-mac.dev"
          }
        }
      }
```

Version, I think makes more sense as the combination of where it's coming from and the value that was set to get the 1:1 correspondence with the CLI args doesn't it?

These two:

```
      "cpu_physical_cores": 36,
      "cpu_model":"Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz",
```

I still need to find a way for how to get them reliably (across windows + linux) ... but I don't think it'll be a problem :)
An explicit Id for each run we don't need I think, the timestamps will be unique?

suyograo:
@original-brownbear nice. Looking good.

> Version, I think makes more sense as the combination of where it's coming from and the value that was set to get the 1:1 correspondence with the CLI args doesn't it?

Makes sense!

> cpu_model":"Intel(R) Xeon(R) CPU E5-2699 v3 @ 2.30GHz",

I think we can skip this. Not important. The `cores` give good info.

> An explicit Id for each run we don't need I think, the timestamps will be unique?

Yep, works

original-brownbear:
@suyograo alright then, if we can defer some improved documentation on this to the next iteration, then I think we're good for a review here :)

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/DataStore.java
```diff
@@ -0,0 +1,84 @@
...
+        ElasticSearch(final String host, final int port, final String schema,
+            final Map<String, Object> meta) {
+            client = RestClient.builder(new HttpHost(host, port, schema)).build();
+            this.meta = meta;
+        }
+
+        @Override
+        public void store(final Map<LsMetricStats, ListStatistics> data) throws IOException {
+            if (client.performRequest(
+                "POST", "/benchmarks/measurement/",
```
suyograo: Can we make the index `logstash-benchmarks`

original-brownbear: @suyograo sure let me fix that, good point :)

PR #7836, opened 2017-07-28
MINOR: Flatten Pipeline Worker Threads

original-brownbear:
This makes the stack of a worker loop one level flatter and makes the pipeline worker thread naming actually work.

### Before

screen shot 2017-07-28 at 10 39 08

### After

screen shot 2017-07-28 at 10 40 04


original-brownbear:
@jsvd thanks!

PR #7837, opened 2017-07-28
MINOR: Improve Batch Iteration + GC

original-brownbear:
A few improvements I already made in my java execution branch:

* For the in memory queue we never use the `starting_size` of the batch => no need to keep separate collections for generated and original events
   * This saves instantiating potentially redundant `@generated`
   * Makes the GC's life easier (this might be meaningful in some cases where the batch is the last reference to an event and we now only have a single reference to collect before we can collect the events)
* `@iterating_temp` is only used once in the lifecycle of the batch => we should `null` it out once we don't need it anymore so it can get collected and save the `clear` call which is kinda expensive
* Also, using `HashSet` here makes more sense than using `RubyHash` (which is a linked hash map) since we don't need any iteration order guarantees => much easier on the GC
* Removed  redundant private getter for the iteration lock

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -253,7 +253,7 @@ module LogStash; module Util
 
         @generated = Hash.new
         @iterating_temp = Hash.new
-        @iterating = false # Atomic Boolean maybe? Although batches are not shared across threads
+        @is_iterating = false # Atomic Boolean maybe? Although batches are not shared across threads
         @acked_batch = nil
       end
 
@@ -274,7 +274,7 @@ module LogStash; module Util
         return if event.nil? || @originals.key?(event)
         # take care not to cause @generated to change during iteration
         # @iterating_temp is merged after the iteration
-        if iterating?
+        if @is_iterating
           @iterating_temp[event] = true
         else
           # the periodic flush could generate events outside of an each iteration
@@ -293,14 +293,14 @@ module LogStash; module Util
 
         # below the checks for @cancelled.include?(e) have been replaced by e.cancelled?
         # TODO: for https://github.com/elastic/logstash/issues/6055 = will have to properly refactor
-        @iterating = true
+        @is_iterating = true
         @originals.each do |e, _|
           blk.call(e) unless e.cancelled?
         end
         @generated.each do |e, _|
           blk.call(e) unless e.cancelled?
         end
-        @iterating = false
+        @is_iterating = false
         update_generated
       end
 
@@ -331,10 +331,6 @@ module LogStash; module Util
       end
 
       private
-
-      def iterating?
-        @iterating
-      end
 
       def update_generated
         @generated.update(@iterating_temp)
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -3,8 +3,8 @@
 module LogStash; module Util
   class WrappedSynchronousQueue
     java_import java.util.concurrent.ArrayBlockingQueue
-    java_import java.util.concurrent.SynchronousQueue
     java_import java.util.concurrent.TimeUnit
+    java_import java.util.HashSet
 
     def initialize (size)
       @queue = ArrayBlockingQueue.new(size)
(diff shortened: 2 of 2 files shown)


logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -332,10 +332,6 @@ def flush_signal_received?
 
       private
 
-      def iterating?
-        @iterating
-      end
```
jakelandis: Just curious, is this a performance anti-pattern ?

original-brownbear: @jakelandis hmm theoretically yes since JRuby doesn't inline this stuff and it appears to translate into 2 function calls. If it were just Java the answer is no, JVM inlines pure accessors every time, but on JRuby it seems to add overhead.

guyboertje: It is simply a Ruby convention to wrap a boolean in a `<ivar>?` method. It conveys the boolean intent better to other developers.
I wish we could name the ivar itself with a `?` at the end.
Maybe we should adopt Hungarian prefix `b_` for booleans **only** in LS Ruby code? Half joking on here.

original-brownbear: @guyboertje 

> Maybe we should adopt Hungarian prefix b_ for booleans only in LS Ruby code? Half joking on here.

Yea def, for Ruby (since we have no types) some convention on this kind of variable does make sense imo. As a Java person I'm more used to `is_` prefixes (without the underscore in Java) for this, but `b_` I like too :)

guyboertje:
@original-brownbear - I am all for moving the two wrapped_?_queue impl, with the internal classes to Java.
Is it too soon to do it now?
I mean it will be needed as a drop in component in a node the Java Execution model, no?

original-brownbear:
@guyboertje I fear this is still quite a bit of work :(, mainly because you'd have to adjust a bunch of tests (a few specs directly use the inner classes here) and because of the metrics code interaction.

guyboertje:
I understand.
In therms of the inner classes, this was an expedience by me to keep the implementations self contained (to foster the idea of drop-in replaceable and because Ruby does not have strict intefaces) and to prevent one becoming dependent on classes in the other.
I suppose a Java implementation would not need these initial assumptions.

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
guyboertje: I like `is_` though too.

guyboertje: May I suggest that you create a discuss issue to use prefix and progressively clean up `name?` methods. Especially as you can explain the JRuby inlining problem just once for all to read. When I say all (I have to repeat what Jordan would say) I also mean the many non-LS folk that follow what we do, that may not read every PR. These external people may have custom plugins or forks of core ones.

original-brownbear: @guyboertje hmm is it worth the noise for this in general though? I mean here, I went for this because it's a very hot loop and anything that makes the bytecode that is generated more predictable is a win, but in something like the settings class it doesn't really matter?

guyboertje: True, but people aren't always aware whether the code are in a hot loop or not. I only suggested, its your call, but consider that PH, Colin and Andrew are away and Joao will be shortly (paternity).

original-brownbear: @guyboertje I'd vote no on putting any kind of guideline like this out then. It's just such a trivial improvement to the bytecode size only.
+ I think here and in all other cases where this kind of thing matters, the move to Java is incoming anyways, so fixing these spots for at best minor gains will not get us much imo.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -5,6 +5,7 @@ class WrappedSynchronousQueue
     java_import java.util.concurrent.ArrayBlockingQueue
     java_import java.util.concurrent.SynchronousQueue
```
guyboertje: Can remove?

original-brownbear: @guyboertje yea sec killing it in this PR then :)

original-brownbear: and it's gone :)

guyboertje:
@jakelandis - over to you for review.

original-brownbear:
@jakelandis are we good here? :)

jakelandis:
@original-brownbear - sorry for the delay. I just realized that I was reviewer here. 

On the surface the code seem fine, but am not familiar enough with this section to fully understand the changes here. 

I will defer to @guyboertje LGTM.

original-brownbear:
@jakelandis np + thank :)

PR #7838, opened 2017-07-28
Add check to stop messages looping through DLQ

robbavey:
Add check to stop events from being written to the DLQ that have previously been written to the DLQ

### DIFF
--- a/docs/static/dead-letter-queues.asciidoc
+++ b/docs/static/dead-letter-queues.asciidoc
@@ -117,6 +117,9 @@ queue, it will continue to run and process new events as they stream into the
 queue. This means that you do not need to stop your production system to handle
 events in the dead letter queue. 
 
+NOTE: Events emitted from the dead letter queue input plugin will not be resubmitted to the
+dead letter queue if they cannot be processed correctly
+
 [[dlq-timestamp]]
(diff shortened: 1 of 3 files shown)


suyograo:
@robbavey can you update the behavior in the docs in this PR? https://github.com/elastic/logstash/blob/master/docs/static/dead-letter-queues.asciidoc

robbavey:
@suyograo What are your thoughts for 6.1+ on loosening this by introducing the notion of a configurable 'dlq depth'? Effectively, we would mark each passage through the DLQ, by adding a metadata filter to the message, including a counter that could be updated each time a message goes through the dlq. We could default this value to 1, but as filters add DLQ capabilities, it may become useful to allow messages to go through the DLQ >1 time

suyograo:
@robbavey I like the idea, but I wonder if we should keep it as is for the short term and make your proposed change of `depth` when we get a concrete usecase from a user?

suyograo:
Anyways, this patch is LGTM

robbavey:
@suyograo Backports all round for this? 5.5, 5.6, 6.0, and 6.x?

suyograo:
@robbavey I think we can skip 5.5. This one is not super critical.

Issue #5405, opened 2016-05-31
GC Activity


driverpt:
It would be nice if Logstash would create a Pool of reusable Event objects instead of constantly creating new ones.

The purpose of this, is to reduce the GC Overhead when running a Pipeline.
I recommend that if this goes forward, the Pool size should be equal to the Max In Flight objects in the Pipeline.

It would cost a bit of memory, but it will boost the performance.

Log4J2 has already done something similar. See [here](https://www.infoq.com/news/2016/05/log4j-garbage-free)

jsvd:
@driverpt first of all, thanks for bringing this issue up and starting the PR. What follows below is my raw mind dump about this:


To add a bit more context to this issue, here is a bit of background on logstash memory consumption:

-----

Logstash is an streaming event processing pipeline, which means the unit of work is the event, which is created from byte streams through the inputs, gets filtered, and then pushed out through outputs.
Once these stages are completed the event is discarded.
Each event is composed of a LogStash::Event instance and data in the form of multiple instances of classes like strings, numbers, arrays and hash maps.

If Logstash is processing 10k eps it will allocate + deallocate events at the same rate, 10k allocations+10k deallocations per second (note that deallocations will be batched by the gc).

-----

With this in mind it begs the question if such allocation+GC activity has an impact in performance, and if so at which rate does logstash saturate the cpu enough that GC is stealing away from it.

Also, taking from the fact that GC overhead accounts for more than X% of cpu time and that X is not tolerable, we need to understand what instances do we need to avoid allocating to lower X% to an acceptable Y%.
For example, do we accomplish this by not allocating LogStash::Event instances as in #5455 ? Thinking now, I can think of other (hidden) costs, drawbacks and tradeoffs that need measuring, like:
* all data that is put into the events is still allocated and collected from the heap
* cleaning the contents of an event for reusability has a cost
* does the return of an output's `receive(event)` ensure the event won't be used anymore? What if an output buffers?
* what is the overhead of managing the object pool?
* filters like split and clone will consistently multiply the inflight event count by N fold, should we parse the config file and adjust this pool size?

@driverpt ^ these are my current questions about this, I haven't ran any benchmarks or done any tests to know more about this problem as such an investigation is as important is time consuming :)
Is there more information you can share, like reproducible benchmark scripts/results, heapdumps, preliminary results with your #5455 ?

colinsurprenant:
[...]

PR #7839, opened 2017-07-28
Remove Accessors Caching

original-brownbear:
* `Accessors` with cached state get us nothing
   * Assuming a hot cache for a given path
       * For accessing top level fields they slow things down by adding another level of indirection
       * For accessing second level data they have the same level of indirection for `get`
       * sets destroy the cache for the whole `Event` => depending on the plugins, the cache can be almost useless
   * Assuming a cold cache the cache is only overhead obviously
   * Cache logic blows up the size of `Accessor` bytecode wise and prevents inlining into `Event`
   * The cache costs a lot of memory (with metrics turned off, this branch has an allocation byte rate about 25% below `master`) 
 * In practice with metrics turned off (metrics are too heavy, they gate everything when turned on at the moment by being a 50%+ performance hit on a maxed out box), this version of `Accessors` ~15-20% faster on 4 threads and the Apache dataset for `Xmx=1G` than `master` for me

Changes made:

* Removed redundant data and meta_data map references in `Accessors`
* Removed `lut` field that held a cache for `field reference => field value` used to avoid nested `Map<>` lookups
* This leaves `Accessors` with no state of its own => made its code path `static`
* This removes the ~50/50 (according to JitWatch) branching between paths hot and cold cache => the previous "cold" path is inlined better => split up all its sub-branches into individual methods to make the most of it and it seems now `Accessors` does in fact get fully inlined and eventually unloaded
* Also removed 100% redundant throws for IndexOutOfBoundsException and ClassCastException (if you look at both instances, those exceptions will be thrown automatically in all cases anyways, but `Throwable` constructors cannot be inlined 

Both of the below are 4 workers, 1024 batch size and metrics off at `Xmx = 1GB`.

### Before

screen shot 2017-07-28 at 23 52 16

### After

screen shot 2017-07-28 at 23 48 53

* More minor, fewer major GCs
* Much lower allocation rate (see slope for old gen)

This should bring some help to issues with major GC like e.g. https://github.com/elastic/logstash/issues/5405#issuecomment-224840073

(if you let things run for a longer time than in the pictures above major pauses tend to be a lot shorter here too (I can run a few more examples tomorrow and add them here), which is expected since all event fields have much fewer paths to them now)

EDIT:

some benchmarks: 

Both of the below are 4 workers, 512 batch size and metrics off at `Xmx = 1GB`.

```sh
➜  logstash git:(lut-no-more) ✗ time cat ~/Downloads/apache_access_logs
13.2MiB 0:06:36 [34.0KiB/s] [                                                                                        <=>                                                                                        ]
 13801093
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.14s user 2.27s system 0% cpu 6:35.88 total
bin/logstash -b 512 -w 4 -f ~/tmp/logstash.cfg  1718.84s user 27.00s system 440% cpu 6:36.23 total
pv  2.66s user 16.51s system 4% cpu 6:36.23 total
wc -c  0.34s user 2.65s system 0% cpu 6:36.23 total
```

vs.

```sh
➜  logstash git:(master) ✗ time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -b 512 -w 4 -f ~/tmp/logstash.cfg  | pv  | wc -c
13.2MiB 0:06:56 [32.4KiB/s] [       <=>                                                                                                                                                                        ]
 13801093
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.13s user 2.23s system 0% cpu 6:55.89 total
bin/logstash -b 512 -w 4 -f ~/tmp/logstash.cfg  1858.08s user 29.09s system 453% cpu 6:56.49 total
pv  3.42s user 21.46s system 5% cpu 6:56.49 total
wc -c  0.52s user 4.76s system 1% cpu 6:56.49 total
```

=> Less CPU use % and faster => GC wastes more cycles than caching saves.

But it's not only GC, turning `Xmx` up to `4G`:

```sh
➜  logstash git:(lut-no-more) ✗ time cat ~/Downloads/apache_access_logs
13.2MiB 0:05:59 [37.5KiB/s] [                                                                                                                                         <=>                                       ]
 13801157
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.13s user 2.22s system 0% cpu 5:59.22 total
bin/logstash -b 512 -w 4 -f ~/tmp/logstash.cfg  1524.02s user 25.50s system 430% cpu 5:59.84 total
```

```sh
➜  logstash git:(master) ✗ time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -b 512 -w 4 -f ~/tmp/logstash.cfg  | pv  | wc -c
13.2MiB 0:06:26 [34.9KiB/s] [                                                                                                                    <=>                                                            ]
 13801157
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.14s user 2.23s system 0% cpu 6:25.37 total
bin/logstash -b 512 -w 4 -f ~/tmp/logstash.cfg  1633.34s user 24.36s system 429% cpu 6:26.09 total
pv  2.69s user 17.03s system 5% cpu 6:26.09 total
wc -c  0.36s user 2.83s system 0% cpu 6:26.09 total
```

still faster, even when GC becomes trivial. Imo it's almost impossible to make a realistic case where this cache helps anything, given that is evicted on every `set` operation.
So the only winners here would be inefficient plugins that make redundant `get` calls to fields at nesting level 3 and up. Code without deep, redundant `get` calls can only loose imo.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core/src/main/java/org/logstash/Accessors.java
@@ -1,201 +1,157 @@
 package org.logstash;
 
-import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
-class Accessors {
+public final class Accessors {
 
-    private final ConvertedMap data;
-    protected Map<String, Object> lut;
-
-    public Accessors(final ConvertedMap data) {
-        this.data = data;
-        this.lut = new HashMap<>(); // reference -> target LUT
+    private Accessors() {
+        //Utility Class
     }
 
-    public Object get(String reference) {
-        FieldReference field = PathCache.cache(reference);
-        Object target = findTarget(field);
-        return (target == null) ? null : fetch(target, field.getKey());
-    }
-
-    public Object set(String reference, Object value) {
+    public static Object get(final ConvertedMap data, final String reference) {
         final FieldReference field = PathCache.cache(reference);
-        final Object target = findCreateTarget(field);
-        final String key = field.getKey();
+        final Object target = findParent(data, field);
+        return target == null ? null : fetch(target, field.getKey());
+    }
+
+    public static Object set(final ConvertedMap data, final String reference,
+        final Object value) {
+        final FieldReference field = PathCache.cache(reference);
+        return setChild(findCreateTarget(data, field), field.getKey(), value);
+    }
+
+    public static Object del(final ConvertedMap data, final String reference) {
+        final FieldReference field = PathCache.cache(reference);
+        final Object target = findParent(data, field);
         if (target instanceof ConvertedMap) {
-            ((ConvertedMap) target).put(key, value);
-        } else if (target instanceof ConvertedList) {
-            int i;
-            try {
-                i = Integer.parseInt(key);
-            } catch (NumberFormatException e) {
-                return null;
-            }
-            int size = ((ConvertedList) target).size();
-            if (i >= size) {
-                // grow array by adding trailing null items
-                // this strategy reflects legacy Ruby impl behaviour and is backed by specs
-                // TODO: (colin) this is potentially dangerous, and could produce OOM using arbitrary big numbers
-                // TODO: (colin) should be guard against this?
-                for (int j = size; j < i; j++) {
-                    ((List<Object>) target).add(null);
-                }
-                ((List<Object>) target).add(value);
-            } else {
-                int offset = listIndex(i, ((List) target).size());
-                ((ConvertedList) target).set(offset, value);
-            }
+            return ((ConvertedMap) target).remove(field.getKey());
         } else {
-            throw newCollectionException(target);
+            return target == null ? null : delFromList((ConvertedList) target, field.getKey());
         }
-        return value;
     }
 
-    public Object del(String reference) {
-        FieldReference field = PathCache.cache(reference);
-        Object target = findTarget(field);
-        if (target != null) {
-            if (target instanceof ConvertedMap) {
-                return ((ConvertedMap) target).remove(field.getKey());
-            } else if (target instanceof ConvertedList) {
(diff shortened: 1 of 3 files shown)


original-brownbear:
@suyograo this is what I was talking about, now with pictures :)

suyograo:
@original-brownbear can you include performance runs for apache between two branches as well ? (with our new performance tool)

original-brownbear:
@suyograo sure will do tomorrow. Though it's not as interesting as runs with `pv | wc -c` because the tool needs the metrics and they way gate this at the moment :(
That said ... not to go too far off topic ... maybe the performance tool should also count the dots from the process output for those benchmark runs that have `dots` to `stdout` for better accuracy when comparing older versions (even if we fix the metrics, the metrics off baseline we can't compare with the tool otherwise).

original-brownbear:
this gives rise to https://github.com/elastic/logstash/pull/7847 which comes with a significant reduction in `String` allocations and memory use

guyboertje:
IIRC, the accessors implementation was a direct port of the Ruby Event + Accessors - there was little investigation into its efficacy - merely the path of least resistance to a Java Event that could serialize into the PQ.

Investigating this now, is due IMO.

**However, I think that other LS team members should understand what is changing and why. At the very least so that we can all be up to speed on such changes when handling Discuss, GH and SDH issues.**

original-brownbear:
@guyboertje sure, I guess we can sent out an email once this is actually happening? Anything we should add that isn't in the PR description?

original-brownbear:
Jenkins test this please

PR #7840, opened 2017-07-28
Remove closed DeadLetterQueueWriters from 'Factory'

robbavey:
Fix bug where reloading a pipeline would close the DLQWriter, but
leave the closed version cached in the factory object, stopping
the reloaded pipeline from being able to write to the dead letter
queue.


suyograo:
LGTM

robbavey:
@suyograo 5.6, 6.0 and 6.x for this too?

suyograo:
@robbavey yep all the way to 5.5

PR #7846, opened 2017-07-29
MINOR: Cleanup multiple ObjectMappers being used

original-brownbear:
Motivated by finding inline `ObjectMapper` instantiation in the LIR code :)

* Removed all redundant `ObjectMapper` instances, much better to only have one global one from the caching perspective
   * Also no reason not to have `AfterBurner` active for all JSON spots :)
* Also cleaned up the instantiation of the mappers a little
   * Much cleaner to read/maintain if you don't have to jump to the `static` block


jakelandis:
LGTM (assuming green, or unrelated failures)

PR #7847, opened 2017-07-30
Just for illustrative purposes: Use Charsequence Keys in Lookups

original-brownbear:
Leveraging #7839 to get 0 GC lookups on `Event` fields

Not really work in progress code wise, it passes all tests just fine.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core/src/main/java/org/logstash/Accessors.java
@@ -4,202 +4,157 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-public class Accessors {
+public final class Accessors {
 
-    private Map<String, Object> data;
-    protected Map<String, Object> lut;
-
-    public Accessors(Map<String, Object> data) {
-        this.data = data;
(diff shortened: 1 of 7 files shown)


guyboertje:
From a Java education perspective, could you comment on why using CarSequence improves matters here?

original-brownbear:
@guyboertje 

Disclaimer: just from the pure Java perspective `CharSequence` as lookup key is a sin!

But in this case:

We are currently caching `String` -> "all data we need to look up a field".
Why not also cache `RubyString` ->"all data we need to look up a field"? It's a trivial amount of extra memory but allows us to never have to convert `RubyString` to `String` for lookups since we only use whatever the `PathCache` gives us for the actual lookup on the `Event`'s data.

Hope that makes sense, `CharSequence` is more of a convenience choice than anything here, because `String` and the `ByteList` backing `RubyString` both fulfil it. They won't both hit the same key in the path cache.

original-brownbear:
@guyboertje killing this in favor of https://github.com/elastic/logstash/pull/7889 <= much easier solution + much faster

PR #7849, opened 2017-07-30
PERFORMANCE: Move null guard out of Javafier.deep

original-brownbear:
We can actually factor the `null` guard out here since this method is only called in 4 production code places, the two handled here + two in `ConvertedList` and `ConvertedMap`.

In `ConvertedList` and `ConvertedMap` we know we won't have `null` entries pretty much ever (and if we do `BiValues.newBiValue` is able to deal with it), but in all other cases we can simply skip this huge (byte code wise) function that is pretty likely (50% in the Apach dataset according to JIT watch + more than that (pretty much `100%` then) if you only add a single tag to an `Event`.)
=> much better inlining outright.

Also to make it even better I extracted the cold path from this method and commented on the why of that in the code.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Event.java
+++ b/logstash-core/src/main/java/org/logstash/Event.java
@@ -136,7 +136,8 @@ public final class Event implements Cloneable, Queueable {
     }
 
     public Object getField(String reference) {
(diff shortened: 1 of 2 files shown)


guyboertje:
This makes complete sense to me.

Issue #7851, opened 2017-07-31
Non ASCII no longer supported in pipeline config.


jarpy:
This seems to be [by design](https://github.com/elastic/logstash/commit/645fcec234e44c43e595710b438db757ac519f46#diff-0da0dea673120e65c9fc4e91cb5a9fb1R61), but I used to be able to do neat things like [this](https://github.com/elastic/stack-docker/blob/b02958b5b720eab4d871dfb356445046e3b14bc7/config/logstash.conf#L4).

I guess this is either an improvement or a regression, depending on one's perspective.

Relevant XKCD: https://xkcd.com/1172/

original-brownbear:
@jarpy but I guess 💓 is UTF-8? The way I read

> The following config files contains non-ascii characters but are not **UTF-8 encoded**

this should still work. The code that determines that just changed from:

```rb
!cfg.ascii_only? && !cfg.valid_encoding?
```

```rb
content.ascii_only? && content.valid_encoding?
```

looks like this is just a typo, I'll add a quick PR for that

original-brownbear:
Fixed in #7852 :)

PR #7852, opened 2017-07-31
 #7851 allow UTF-8 in config file

original-brownbear:
Seems more logical to me this way, see my comment https://github.com/elastic/logstash/issues/7851#issuecomment-318997183

"Either your file content is just ascii or it has an otherwise valid encoding" ... is what this check should be imo

Also this works fine:

```sh
➜  logstash git:(7851) ✗ bin/logstash -f ~/tmp/heart.cfg                                                                                
Sending Logstash's logs to /Users/brownbear/src/logstash/logs which is now configured via log4j2.properties
[2017-07-31T10:18:32,532][INFO ][logstash.setting.writabledirectory] Creating directory {:setting=>"path.dead_letter_queue", :path=>"/Users/brownbear/src/logstash/data/dead_letter_queue"}
[2017-07-31T10:18:32,603][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified
[2017-07-31T10:18:33,249][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2017-07-31T10:18:33,944][INFO ][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"main", "pipeline.workers"=>8, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>5, "pipeline.max_inflight"=>1000, :thread=>"#<Thread:0x243b2e3 run>"}
[2017-07-31T10:18:33,980][INFO ][logstash.pipeline        ] Pipeline started {"pipeline.id"=>"main"}
[2017-07-31T10:18:34,037][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
2017-07-31T08:18:34.035Z localhost Hello from Logstash 💓
2017-07-31T08:18:39.024Z localhost Hello from Logstash 💓
^C[2017-07-31T10:18:40,236][WARN ][logstash.runner          ] SIGINT received. Shutting down the agent.
[2017-07-31T10:18:40,400][INFO ][logstash.pipeline        ] Pipeline terminated {"pipeline.id"=>"main"}

```

### DIFF
--- a/logstash-core/lib/logstash/config/source/local.rb
+++ b/logstash-core/lib/logstash/config/source/local.rb
@@ -53,7 +53,7 @@ module LogStash module Config module Source
 
           config_string = ::File.read(file)
 
-          if valid_encoding?(config_string)
+          if config_string.valid_encoding?
             part = org.logstash.common.SourceWithMetadata.new("file", file, 0, 0, config_string)
             config_parts << part
           else
@@ -99,10 +99,6 @@ module LogStash module Config module Source
         t = ::File.split(@path)
         all_files = Dir.glob(::File.join(t.first, "*")).sort
         all_files - get_matched_files
-      end
-
-      def valid_encoding?(content)
-        content.ascii_only? && content.valid_encoding?
       end
 
       def temporary_file?(filepath)


logstash-core/lib/logstash/config/source/local.rb
```diff
@@ -102,7 +102,7 @@ def get_unmatched_files
       end
 
       def valid_encoding?(content)
-        content.ascii_only? && content.valid_encoding?
+        content.ascii_only? || content.valid_encoding?
```
jsvd: is there a situation where the encoding is valid but it's not ascii, or it is ascii but not valid? maybe `content.valid_encoding?` is enough.

Also, I suggest double checking if this has any interference with the `config.support_escapes` feature ( don't think it does, just worth checking)

original-brownbear: @jsvd 

> is there a situation where the encoding is valid but it's not ascii, or it is ascii but not valid? maybe content.valid_encoding? is enough.

I know too little about Ruby to answer that I'm afraid :( Instinctively it seems like `content.valid_encoding?` should imply `valid_encoding? == true`.
=> Valid and not ascii is what was reported in the issue and fixed here, the other way around ascii and not valid should be impossible?

original-brownbear: @jsvd I don't think this interferes with the escape feature, my logic is:

```rb
    def process(value)
      case value[1]
      when '"', "'", "\\"
        value[1]
      when "n"
        "\n"
      when "r"
        "\r"
      when "t"
        "\t"
      else
        value
      end
    end
```

none of these are different between UTF8 and ASCII => we're good?

jsvd: [valid_encoding](https://github.com/jruby/jruby/blob/dd0b88e44efe91ed1cf163b99abac0362d614d1b/core/src/main/java/org/jruby/RubyString.java#L5435-L5438) just seems to test if it's not 16 or 32 bit encoded: https://github.com/jruby/jruby/blob/c24f7fa865d96c76235ebab591db938b0c52851b/core/src/main/java/org/jruby/util/StringSupport.java#L62-L69, so I'm guessing it's ok to have only that check

original-brownbear: @jsvd makes sense, will clean this up :)

original-brownbear:
@jsvd drop the ascii check before merging?

jsvd:
yeah sorry I was too merge-happy. I edited my approval message :)

jarpy:
Nice! Please allow me to express my 💓 and appreciation for this.

original-brownbear:
@jsvd alright adjusted, now I just inlined that method since it's only used in one spot and a pointless wrapper now. Still happy? :)

jsvd:
always! LGTM, please merge after CI shows no related failures

PR #7854, opened 2017-07-31
PERFORMANCE: Completely drop StringBiValue

original-brownbear:
About a 5% throughput increase for the Apache dataset.

Also, very visible reduction in GC activity:

(128 batch size, metrics on even because I forgot):

before:
screen shot 2017-07-31 at 10 52 00

after:

screen shot 2017-07-31 at 10 51 36

Changes:

* Custom serializer for `RubyString`
* Short circuit all paths to `BiValues` to use `RubyString` instead of `StringBiValue`

... obviously, we could/should do the same thing to all other `BiValue` types, just didn't want to blow up this PR too much.
... also, I'm aware that this does mess with my inlining improvements from before, but I can bring them back easily once we ported all `BiValue` which is a ~2h task now and get a much nicer result if we have the full picture.


original-brownbear:
@guyboertje @jakelandis @robbavey check it out (since us 4 talked about this :))

original-brownbear:
btw: combined with #7839 and #7847 this is a much more severed speed-up than in isolation (altogether ~30% on the Apache data-set when metrics are off)

jordansissel:
@guyboertje Can you review this? I think you are more familiar with BiValue practices than I am.

The proposed change and proposed benefits sounds good to me.

guyboertje:
@jordansissel - over to you

original-brownbear:
@jordansissel ping, we're still waiting for your ok on this one :)

jordansissel:
What did you want me to look at? I think guy's LGTM is good enough?

original-brownbear:
@jordansissel for me it is, but since @guyboertje asked for your approval I didn't want to override that :) Merging then!

Issue #7857, opened 2017-07-31
Metrics API response can be very slow


jakelandis:
Calling into any of the metrics API calls, the http response can be very slow. 5+ seconds.

This slowness is happening due to a call to ```Socket.gethostname``` for every http request.  

For example:
```
time curl -v -XGET 'localhost:9600/_node/stats/pipelines?pretty'
```
```
real	0m5.155s
user	0m0.006s
sys	0m0.006s
```

suyograo:
Related: https://github.com/elastic/logstash/issues/6520

Issue #6520, opened 2017-01-11
Logstash can hang calculating hostname


andrewvc:
We use `Socket.gethostname` in some spots in the API without caching its output under [some conditions](http://www.markhneedham.com/blog/2015/03/29/inetaddressimpllookupallhostaddr-slowhangs/) this can be very slow. @tsouza encountered this issue not too long ago. 

We should cache this value once at startup and reuse it to fix this issue. It is not safe to depend on it being fast all the time.

tsouza:
Here is the original issue https://github.com/elastic/logstash/issues/6516

jordansissel:
Just a note, `Socket.gethostname` is something we should expect to be a nonstable value during the life of Logstash - a server's hostname can change. Further, I think speeding up the Socket.gethostname call is just a band-aid on the problem -- why are we calling Socket.gethostname for every http request received by the http api?

Given the report in #6516, I think the best fix is to simply avoid this lookup.

andrewvc:
@jordansissel we return the hostname as part of the standard response. 

While yes, it can change in the middle of a Logstash run, I don't think it's too much to ask a user to restart logstash if they change the hostname.

jordansissel:
@andrewvc If we're going to do that, maybe we should do what other non-java tools do.

* ruby calls `uname` to get the hostname
* `hostname` calls `uname` to get the hostname
* `uname -n` calls `uname` to get the hostname
* JRuby seems to call `uname` and then make a bunch of socket calls that seems to be checking for what addresses are used when connecting on various networks (IPV6, IPV4)

I'm pretty content to just fetch our hostname using the `uname` syscall. It avoids doing dns lookups and avoids any possibly-bad cachine; caching is something we'd have to document and take bugs against.

andrewvc:
@jordansissel I'm fine with that but how do we invoke the `uname` syscall on Java? I'd rather not get into JNI here.

andrewvc:
@jordansissel an interesting SO post on the topic http://stackoverflow.com/questions/7348711/recommended-way-to-get-hostname-in-java

jordansissel:
@andrewvc FFI or JNA are fine methods for calling native code, imo.

jordansissel:
@andrewvc I don't' have any comment on the SO post, but for the platforms we support:

1) It's probably incorrect behavior to do a DNS lookup when asking what the system hostname is.
[...]

PR #7858, opened 2017-07-31
Metrics: Cache the hostname in class variable

jakelandis:
Fixes: #7857 

@jordansissel - can you please review ?

Over [here](https://github.com/elastic/logstash/issues/6520#issuecomment-271753343) you mention:
>Just a note, Socket.gethostname is something we should expect to be a nonstable value during the life of Logstash - a server's hostname can change. 

While I agree, I believe that changing the host name with out restarting to be a fairly rare. 

A 5 second delay and extra IO is obviously not good, and with X-pack is happening with high frequence. The solution proposed over at #6520 is likely the best fix (though hostname is [surprising difficult](https://github.com/elastic/support-dev-help/issues/2118#issuecomment-315542137)).  

In the spirit of `don't let the perfect be the enemy of good`, I would advocate this very simple change.

### DIFF
--- a/logstash-core/lib/logstash/api/commands/default_metadata.rb
+++ b/logstash-core/lib/logstash/api/commands/default_metadata.rb
@@ -12,7 +12,7 @@ module LogStash
         end
 
         def host
-          Socket.gethostname
+          @@host ||= Socket.gethostname
         end
 
         def version


jakelandis:
Before (and first run)
```
real	0m5.155s
```
After (for second+ runs)
```
real	0m0.059s
```

jordansissel:
I agree with your reasoning for the patch.

jordansissel:
Builds failed, but the code looks good to me.

jordansissel:
Test failures are unrelated to this patch.

PR #7862, opened 2017-08-01
 #7817 Accessors should only handle ConvertedMap and ConvertedList as Collection, non-scalar types

original-brownbear:
backport of #7817 to `5.6`


original-brownbear:
@suyograo FYI, backported :)

PR #7864, opened 2017-08-01
Benchmark: Allow setting a repeat count

original-brownbear:
* Adding kind of an important feature, allowing us to run a larger sample size.
* Also trying to move to properties for settings so we can move to specifying a properties file maybe? (we're really starting to have a million settings, I think that would be nice :))

### DIFF
--- a/tools/benchmark-cli/README.md
+++ b/tools/benchmark-cli/README.md
@@ -31,7 +31,9 @@ Option                           Description
 --local-path <String>            Path to the root of a local Logstash           
                                    distribution.                                
                                   E.g. `/opt/logstash`                          
---testcase <String>              Currently available test cases are 'baseline'    
+--repeat-data <Integer>          Sets how often the test's dataset should be    
+                                   run. (default: 1)                            
+--testcase <String>              Currently available test cases are 'baseline'  
                                    and 'apache'. (default: baseline)            
 --workdir <File>                 Working directory to store cached files in.    
                                    (default: ~/.logstash-benchmarks)  
--- a/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
+++ b/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
@@ -34,7 +34,7 @@ public interface LogstashInstallation {
      * @throws IOException On I/O Exception
      * @throws InterruptedException Iff Interrupted
      */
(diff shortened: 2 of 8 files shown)


suyograo:
> Also trying to move to properties for settings so we can move to specifying a properties file maybe? (we're really starting to have a million settings, I think that would be nice :))

+1 on extracting out the settings; they'll blow up :) How about YAML since users are already used to `logstash.yml`?

original-brownbear:
@suyograo sure I'm cool with `yaml` too, next PR on this will contain a quick port to `yaml` then ;)

jakelandis:
nit: Add the repeat factor to the [output](https://github.com/elastic/logstash/pull/7864/files#diff-7f54d28cf5b50548681b76f3a947b542L182). 
```
Running Test Case: baseline (x2)
```

suyograo:
Good point @jakelandis

original-brownbear:
@jakelandis @suyograo guys thought about that, but then I discarded the idea because we show the overall number of events that we ran anyways (so it's just overloading the "UI" needlessly?)?

suyograo:
@original-brownbear I think we'll soon need the `bin/logstash-benchmark` wrapper as well. I wouldn't burden users with the `java -cp ` CLI :)

original-brownbear:
@suyograo sure that's on the radar, I can try and have that + yaml ready for tomorrow (it's like 30min tasks each or so only anyways)

jakelandis:
@suyograo / @original-brownbear - +1 to externalizing config. 

Might just be a personal preference from an old java developer....but I much prefer java properties file over ```yaml``` for key/value pairing.

jakelandis:
LGTM

suyograo:
@original-brownbear back to the repeat count, I think it does make it easier to comprehend the number of runs vs a multiple of events. I'd vote for outputting the runs as well in addition to the events being multiplied. My 2c.

original-brownbear:
Ok you guys win 2 vs.1 and I guess it makes sense to go with the majority opinion when it comes to UI :) Will change in a bit.

original-brownbear:
@jakelandis my master plan was to simply load YAML into a `Properties` object anyways, so you can just use one if you wish via the JVM flag :P

original-brownbear:
here you go guys:

```sh
Logstash Benchmark
Benchmarking Version: 5.5.0
Running Test Case: baseline (x2)
Start Time: Tue 8 1 18:21:07 2017 CEST
```

suyograo:
Awesome. 🚢

Issue #7842, opened 2017-07-29
Windows Jenkins build failure


jakelandis:
I believe that ```SLAVE-07FA0C31A``` is a new build machine. It seems that perhaps JRuby is not installed in the expected location. 

This may be related to the Windows 2012 build referenced on https://github.com/elastic/logstash/issues/7767
```
Started by upstream project "elastic+logstash+pull-request" build number 534
originally caused by:
 GitHub pull request #7798 of commit db913b73e81087a09d4f3c804b2f8ee4f99a53fe, no merge conflicts.
[EnvInject] - Loading node environment variables.
[EnvInject] - Preparing an environment for the build.
[EnvInject] - Keeping Jenkins system variables.
[EnvInject] - Keeping Jenkins build variables.
[EnvInject] - Injecting as environment variables the properties content 
HOME=C:\Users\jenkins

[EnvInject] - Variables injected successfully.
[EnvInject] - Injecting contributions.
Building remotely on SLAVE-07FA0C31A (virtual swarm win32nt windows) in workspace C:\Users\jenkins\workspace\elastic+logstash+pull-request+multijob-windows-compatibility
Wiping out workspace first.
Cloning the remote Git repository
Cloning repository git@github.com:elastic/logstash.git
 > git init C:\Users\jenkins\workspace\elastic+logstash+pull-request+multijob-windows-compatibility # timeout=10
ERROR: Reference path does not exist: /var/lib/jenkins/.git-references/logstash.git
Fetching upstream changes from git@github.com:elastic/logstash.git
 > git --version # timeout=10
using GIT_SSH to set credentials master node jenkins user ~/.ssh
 > git fetch --tags --progress git@github.com:elastic/logstash.git +refs/heads/*:refs/remotes/origin/*
 > git config remote.origin.url git@github.com:elastic/logstash.git # timeout=10
 > git config --add remote.origin.fetch +refs/heads/*:refs/remotes/origin/* # timeout=10
 > git config remote.origin.url git@github.com:elastic/logstash.git # timeout=10
Fetching upstream changes from git@github.com:elastic/logstash.git
using GIT_SSH to set credentials master node jenkins user ~/.ssh
 > git fetch --tags --progress git@github.com:elastic/logstash.git +refs/pull/*:refs/remotes/origin/pr/*
 > git rev-parse "db913b73e81087a09d4f3c804b2f8ee4f99a53fe^{commit}" # timeout=10
[...]

PR #7867, opened 2017-08-01
5.6 Backport: Performance: Use RubyArray.hash for metric's fast lookup key

### LINKED ISSUE #7772: Metrics "fastlookup" is Using RubyArray as Hash Map Key

jakelandis:
The existing implementation uses the RubyArray as the key for the fast lookup Map. Under the covers the Map implementation is comparing equal operators (many times per event), and JRuby Array equals operator is VERY expensive since it literally walks each value of array for each equality. Based on profiling via YourKit, the JRuby Array equals operator is a very hot method consuming upto 60% of sampled CPU cycles while under high load (and no other CPU dominators).

This change is to use the .hash value of the JRuby Array as the key of the fast lookup Map. This implementation still calls .hash for each and every call, which is also expensive, since it also walks the arrays to compute the hash. However, the equality check of the hash value is very fast, and net gain is significant. Upto a 15% increase of throughput.

Fixes #7772

Master PR: #7798

### DIFF
--- a/logstash-core/lib/logstash/instrument/metric_store.rb
+++ b/logstash-core/lib/logstash/instrument/metric_store.rb
@@ -51,12 +51,14 @@ module LogStash module Instrument
       # BUT. If the value is not present in the `@fast_lookup` the value will be inserted and we assume that we don't
       # have it in the `@metric_store` for structured search so we add it there too.
 
-      value = @fast_lookup.get(namespaces.dup << key)
+      # array.hash as the key since it is faster then using the array itself, see #7772
+      fast_lookup_key = (namespaces.dup << key).hash
+      value = @fast_lookup.get(fast_lookup_key)
       if value.nil?
         value = block_given? ? yield(key) : default_value
-        @fast_lookup.put(namespaces.dup << key, value)
+        @fast_lookup.put(fast_lookup_key, value)
         @structured_lookup_mutex.synchronize do
-            # If we cannot find the value this mean we need to save it in the store.
+          # If we cannot find the value this mean we need to save it in the store.
           fetch_or_store_namespaces(namespaces).fetch_or_store(key, value)
         end
       end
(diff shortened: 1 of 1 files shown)


jakelandis:
Merged to 5.6. (Jenkins build failure unrelated: https://github.com/elastic/logstash/issues/7842)

Issue #7868, opened 2017-08-01
DLQ: Incorrect handling of final event in a block can lead to plugin crash or OOM errors


robbavey:
The RecordIOReader can get into a state where it will attempt to re-read the block after the final event has been processed, but with the current position pointer incorrectly positioned.

When the second attempt to read the event occurs, the first action is to read a byte indicating the type of the following event. If this character referenced happens to be a valid value of record type ('c', 's', 'm' or 'e'), then this is considered the start of a valid block, and attempts are made to read it. 
This can cause an IllegalArgumentException which will crash the plugin, or in certain circumstances, an OutOfMemoryError.

robbavey:
Fixed by #7871

Issue #7860, opened 2017-08-01
DLQ: Plugin can crash when events end on block boundaries


robbavey:
When the RecordIOReader tries to read the next event from a segment where the last byte of the event lines up with the block boundary, the plugin can fail with an ArrayIndexOutOfBoundsException, which will cause the DLQ input plugin to fail, and potentially not be able to recover to read the next event.

This may be related to the failures seen in #7820 

Test Case: 

```
    @Test
    public void testBlockBoundary() throws Exception {
        Event event = new Event(Collections.emptyMap());
        char[] field = new char[32610];
        Arrays.fill(field, 'x');
        event.setField("message", new String(field));
        long startTime = System.currentTimeMillis();
        try(DeadLetterQueueWriter writeManager = new DeadLetterQueueWriter(dir, 10 * 1024 * 1024, 1_000_000_000)) {
            for (int i = 0; i < 2; i++) {
                DLQEntry entry = new DLQEntry(event, "", "", "", new Timestamp(startTime++));
                assertThat(entry.serialize().length + RecordIOWriter.RECORD_HEADER_SIZE, is(RecordIOWriter.BLOCK_SIZE));
                writeManager.writeEntry(entry);
            }
        }
        try (DeadLetterQueueReader readManager = new DeadLetterQueueReader(dir)) {
            for (int i = 0; i < 2;i++) {
                readManager.pollEntry(100);
            }
        }
    }
```

Fails with:

```
java.lang.RuntimeException: java.lang.ArrayIndexOutOfBoundsException: 32768

	at org.logstash.common.io.RecordIOReader.seekToStartOfEventInBlock(RecordIOReader.java:176)
	at org.logstash.common.io.RecordIOReader.consumeToStartOfEvent(RecordIOReader.java:227)
	at org.logstash.common.io.RecordIOReader.readEvent(RecordIOReader.java:261)
	at org.logstash.common.io.DeadLetterQueueReader.pollEntryBytes(DeadLetterQueueReader.java:120)
	at org.logstash.common.io.DeadLetterQueueReader.pollEntry(DeadLetterQueueReader.java:96)
	at org.logstash.common.io.DeadLetterQueueReaderTest.testBlockBoundary(DeadLetterQueueReaderTest.java:156)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

Issue #7820, opened 2017-07-26
From Discuss: No events in DLQ


robbavey:
from https://discuss.elastic.co/t/no-events-in-dlq/94309

User is trying to ingest data using the elasticsearch output plugin, and is seeing the following error from Elasticsearch: 

```
MapperParsingException: object mapping for [mapping_request.payload] tried to parse field [payload] as object, but found a concrete value
```

but those entries are not being written to the DLQ as expected.

Some messages _have_ been written to the DLQ, but the user is seeing the following errors
```
[2017-07-25T16:41:19,389][ERROR][logstash.pipeline ] A plugin had an unrecoverable error. Will restart this plugin.
Plugin: "/data/logstashData/dead_letter_queue", sincedb_path=>"/data/logstashDLQData/dlq_offset", commit_offsets=>true, codec=>"json_73337b45-c626-4cae-aab9-b2f4d8775a1e", enable_metric=>true, charset=>"UTF-8">, id=>"52ff26acd4e4d46066179db0a3cdf329cbe3c13c-1", enable_metric=>true, pipeline_id=>"main">
```

robbavey:
@AndreAga Thanks for the information - also, would it be possible to get from you:

- Operating System:
- Config File (if you have sensitive info, please remove it):
- Sample Data (one of the messages that should be added to the DLQ but isn't, and your es mappings)

Thanks!

AndreAga:
**OS Version:** 
NAME="Red Hat Enterprise Linux Server"
VERSION="7.3 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="7.3"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.3 (Maipo)"

**Config File**:
Logstash Service:

```
input {
  kafka {
    bootstrap_servers => "...:9092,...:9092"
    topics_pattern => "logstash-55*"
    consumer_threads => 5
    auto_commit_interval_ms => "1000"
    auto_offset_reset => "latest"
    enable_auto_commit => true
    max_partition_fetch_bytes => "104857600"
    decorate_events => true
    codec => "json"
  }
}

filter {
 ###############################
 ruby {
   code => 'event.set("[fields][log_version]", event.get("[fields][log_version]")+1)'
 }
 ############################################

 if [fields][log_version] == 1 {

  # PROCESS THE EVENT
  # It's the first time that Logstash processes the event, so it applies the correct Pattern
  # Here some logs are converted in json via json {} plugin

[...]

PR #7871, opened 2017-08-01
DLQ:Fix handling of last event in block

robbavey:
The last event in a block was being incorrectly handled in such a way that caused unexpected
behavior depending on the contents of the final event, it could cause plugins to fail, or
Logstash to fail with an OutOfMemoryError

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/io/RecordIOReader.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/RecordIOReader.java
@@ -145,6 +145,10 @@ public final class RecordIOReader implements Closeable {
      *
      */
      int seekToStartOfEventInBlock() {
+         // Already consumed all the bytes in this block.
+        if (currentBlock.position() == currentBlockSizeReadFromChannel){
+             return -1;
+         }
          while (true) {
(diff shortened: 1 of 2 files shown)


suyograo:
@robbavey can we add a unit test for this?

robbavey:
@suyograo The block boundary tests I added test this - they failed previously until I added the fix.

I added an extra test too, which attempts to write random event sizes, to gain a bit more confidence that all cases have been handled.

I'm still testing this locally, to try and see if there any other cases that cause DLQ crashes

robbavey:
This should fix #7868 and #7860, which appear to have been the cause behind some of the errors a customer has been seeing (see #7820). There are tests that verify the scenarios that I have observed crashing the plugin, or in certain circumstances I have reproduced locally, cause Logstash to attempt to allocate a huge amount of memory and crash with an OutOfMemory error. 

The `testReadWriteRandomEventSize` test case is an attempt to verify DLQ read/writes with some fuzz to flush out any scenarios that we may have missed.

suyograo:
LGTM, thanks for the info Rob.

robbavey:
@suyograo Backport to 5.5, 5.6 and 6.0?

suyograo:
@robbavey ++, we should backport all DLQ bugs to 5.5

PR #7872, opened 2017-08-01
Metrics: Clean up, remove and deprecate in support of pure Java based metrics.

jakelandis:
This changes the following:
* DRY Gauge metrics
* Remove Namespace from metrics API (never used, not needed)
* Introduce a 'dirty' flag. This will be used to assist with serialization (in a later commit).
* Change getKey to getName in the API
* Deprecate (only from Java) the future non-supported Guage types.
* Remove NumericGauge and replace with LongGuage and DoubleGuage.
* Bring code coverage back to 100%

Note this change is internally non-passive, however the Java metrics API (recently introduced in 6.0) has not been advertised as a re-usable asset.

Changes in support of #7788


jakelandis:
@andrewvc - would you mind reviewing ?  

There is not much interesting in this PR since these are just supporting changes for #7788

These are standalone changes and can be merged independently to the 6.x/master branches.

andrewvc:
@jakelandis yep! I'll take a look tomorrow :)

logstash-core/src/main/java/org/logstash/instrument/metrics/counter/LongCounter.java
```diff
@@ -12,18 +12,19 @@
  */
 public class LongCounter extends AbstractMetric<Long> implements CounterMetric<Long> {
 
-    private final LongAdder longAdder;
+    private static final IllegalArgumentException NEGATIVE_COUNT_EXCEPTION = new IllegalArgumentException("Counters can not be incremented by negative values");
```
andrewvc: Glad to see we added this.

logstash-core/src/main/java/org/logstash/instrument/metrics/Metric.java
```diff
@@ -88,4 +59,16 @@ default String type() {
         return getType().asString();
     }
 
+    /**
+     * Determine if this metric has a value explicitly set.
+     * @return true if this metric has been set to a specific value, false if it is the default state
+     */
+    boolean isDirty();
```
jakelandis: I am trying to match the serialized output of the Ruby code exactly _or as close as possible_.

Since the Ruby metrics are lazily added, when serialized they only show up if they have been set. Since the Java metrics are defaulted to null or zero's, those null's or zero's appear in the JSON output. 

By checking the dirty flag during serialization,  I can emulate the lazily loaded metrics by only serializes those metrics which are dirty.

jakelandis: I _think_ it is here to stay. 

I guess it depends on if we want to continue serializing only that which has been set, or just output all known values, including the default zeros and nulls. 

The existing code base is a bit inconsistent since in some places it initializes some metrics to zero or null (which flips the dirty flag), but not others metrics. I haven't found much of a pattern to what gets initialized or not.

andrewvc: Actually, thinking through this, we really do want to change this behavior.

Writing a client for the current interface is problematic because some stats don't exist, then suddenly do. A pipeline, for instance, should always have its events counter active at `0` even if it hasn't been explicitly set.

I'm wondering if there's a way to compare the behavior here for testing purposes while changing this functionality?

jakelandis: Spoke with @andrewvc on Slack and agreed to leave it in for v1 so that we can make a like-for-like uplift of the serialized output. But we will deprecate the dirty flag immediately so at a soon-to-be later point in time, we can easily remove this flag and explicitly change the behavior.

jakelandis:
@andrewvc - deprecated dirty flag pushed https://github.com/elastic/logstash/pull/7872/commits/19ed92966b9796109ae3ee995b954415c865ecf4

jakelandis:
Logged https://github.com/elastic/logstash/issues/7885 to remove the deprecated flag and change the behavior to consistent serialization.

jakelandis:
@andrewvc - are you comfortable with these changes ?

jakelandis:
Jenkins test this please.

PR #7873, opened 2017-08-01
[DOCS] Use shared versions file for doc builds

lcawl:
This pull request changes the Logstash Reference documentation build so that it pulls version information from a shared file in elastic/docs/shared repository. This allows versioning information to be updated more uniformly and quickly across many books.

The X-Pack Reference already uses those shared files successfully and other books will be migrated to use them as well. See elastic/elasticsearch#26004 and https://github.com/elastic/kibana/pull/13277


lcawl:
See https://github.com/elastic/elasticsearch/pull/26004

PR #7875, opened 2017-08-02
Benchmark: Some Improvements

original-brownbear:
Some improvements to the benchmark tool:

* Bash script to run it comfortably
* Fix/suppress all compiler warning
* Fix download cache by moving it upstream (gnarf it was just overwriting every time) and add visual feedback for cache hits and before + after downloading
* Trivial cleanup in redundant code for piping around in `ApacheLogsComplex`

### DIFF
--- a/bin/benchmark.sh
+++ b/bin/benchmark.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+java -cp "$(cd `dirname $0`/..; pwd)"'/tools/benchmark-cli/build/libs/benchmark-cli.jar:*' \
+ org.logstash.benchmark.cli.Main "$@"
--- a/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
+++ b/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
@@ -55,6 +55,7 @@ public interface LogstashInstallation {
                 } else {
                     output.blue(String.format("Downloading Logstash %s.", version));
                     LogstashInstallation.FromRelease.download(pwd, version);
+                    output.blue("Finished downloading Logstash.");
                 }
                 this.base = LogstashInstallation.FromRelease.setup(basedir);
             } catch (IOException | NoSuchAlgorithmException ex) {
@@ -83,9 +84,8 @@ public interface LogstashInstallation {
             LsBenchDownloader.downloadDecompress(
                 pwd,
                 String.format(
-                    "https://artifacts.elastic.co/downloads/logstash/logstash-%s.zip",
-                    version
-                ), true
+                    "https://artifacts.elastic.co/downloads/logstash/logstash-%s.zip", version
+                )
             );
         }
 
@@ -212,8 +212,7 @@ public interface LogstashInstallation {
         private static void download(final File pwd, final String user, final String hash)
             throws IOException, NoSuchAlgorithmException {
             LsBenchDownloader.downloadDecompress(
-                pwd, String.format("https://github.com/%s/logstash/archive/%s.zip", user, hash),
-                true
+                pwd, String.format("https://github.com/%s/logstash/archive/%s.zip", user, hash)
             );
         }
 
--- a/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
+++ b/tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
@@ -13,6 +13,7 @@ import java.nio.file.StandardOpenOption;
 import java.security.NoSuchAlgorithmException;
 import java.util.Map;
 import org.apache.commons.io.IOUtils;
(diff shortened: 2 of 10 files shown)


bin/benchmark.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+java -cp "$(cd `dirname $0`/..; pwd)"'/tools/benchmark-cli/build/libs/benchmark-cli.jar:*' \
```
jakelandis: Over [here](https://github.com/elastic/logstash/blob/master/tools/benchmark-cli/build.gradle#L55) it seems that you are building using a shadow (fat) jar, but here are using the classpath ?  should this be `java -jar benchmark-cli-shadow.jar` (or the like).

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
```diff
@@ -161,7 +166,7 @@ private static void pipeRepeatedly(final File input, final OutputStream out,
             final int count) throws IOException {
             for (int i = 0; i < count; ++i) {
                 try (final InputStream file = new FileInputStream(input)) {
-                    IOUtils.copy(file, out, 16 * 4096);
+                    IOUtils.copy(file, out, 4096 * 16);
```
jakelandis: just curious... Is this a generally preferred style and why ? _(I have never had this come up before, and never gave it any thought before now)_.

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
```diff
@@ -46,12 +47,17 @@
...
-                this.base = LogstashInstallation.FromRelease.setup(
-                    pwd.toPath().resolve(String.format("logstash-%s", version))
-                );
+                final Path basedir = pwd.toPath().resolve(String.format("logstash-%s", version));
+                if (basedir.toFile().exists()) {
+                    output.blue(String.format("Using Logstash %s from cache.", version));
+                } else {
+                    output.blue(String.format("Downloading Logstash %s.", version));
+                    LogstashInstallation.FromRelease.download(pwd, version);
+                    output.blue("Finished downloading Logstash.");
```
jakelandis: nit: May want to output where it was downloaded to such that user may manually clean it up later.

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
original-brownbear: @jakelandis haha not that I know of, umm I admit I don't even remember making that change ... reverting :D

bin/benchmark.sh
original-brownbear: @jakelandis we're not setting an entry point signature on that jar at the moment.
So I just went with the format we use elsewhere (e.g. ingest converter tool) as that seems to work for everyone compatibility wise :)

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
original-brownbear: @jakelandis seems I didn't do such a good job here quality wise :(, doing that is a lot (relative to the benefit) of additional code it seems.
Let's improve the UI here in the next iteration?

bin/benchmark.sh
jakelandis: It should be easy to add the [main class to the manifest](https://stackoverflow.com/questions/21721119/creating-runnable-jar-with-gradle) making the jar executable.  This can simplify the startup script and provide any easy bridge for Windows users to use this too without creating a .bat file. 

Unless we are intentionally trying to expose the classpath to the end user for extension, I advocate using executable jar's.

tools/benchmark-cli/src/main/java/org/logstash/benchmark/cli/LogstashInstallation.java
jakelandis: np, it is a very minor thing.

bin/benchmark.sh
original-brownbear: @jakelandis let's add an issue for doing this for the ingest-converter + this tool? I agree with you 100% here though, but I think it's worth documenting this move properly (just so we know what happened when we get some manifest error in a release late at night or something :D).

jakelandis: Issue logged: #7877

original-brownbear: @jakelandis thanks!

original-brownbear:
@jakelandis we're good here right?

jakelandis:
I updated the scope of the https://github.com/elastic/logstash/issues/7877 to include this benchmark tool.  _( I mis-read your original comment)_.

LGTM

original-brownbear:
@jakelandis thanks!

PR #7881, opened 2017-08-02
RFC: Test commit to split travis jobs up

robbavey:
Splits the Travis commit into online and offline spec to reduce the number of time out build failures.


robbavey:
no longer needed

PR #7889, opened 2017-08-03
PERFORMANCE: Efficient Field get/set/del/includes on Event

original-brownbear:
This change provides for a way more efficient lookup strategy than the current implementation or even #7847 by:

* Lookup by `FieldReference` and use the fact that the `FieldReference` already contains knowledge about a few things (metadata, timestamp, tag field and so on)
* Deduplicate `FieldReference` globally to get better cache behavior

This avoids:

* A `RubyString` to `String` conversion on **every** `Event` field access!
* Needless string comparisons to handle the metadata corner cases
* Redundant lookup of the `FieldReference` in `RubyEvent#set` (we did look it up once for the timestamp check and then once more for the actual `Event#set` operation downstream)
* Having a lot of logic for `String` parsing that messes with inlining in the hot path (note that we already have 100% of the parse results cached when we do the `String` operations on `reference` in the `Event` lookups/sets at the moment and are fixing this here)

Also, this allows the `RubyString` key to be GCed much sooner if it's not a reused value (which I suspect is the case for many many use cases).

### Before

```
bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg  1706.14s user 36.36s system 417% cpu 6:57.64 total
pv  4.32s user 26.83s system 7% cpu 6:57.64 total
wc -c  0.61s user 5.41s system 1% cpu 6:57.64 total
```

### After

```
➜  logstash git:(lut-no-more2-rebase) ✗ time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg  | pv  | wc -c
bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg  1620.05s user 32.34s system 412% cpu 6:40.48 total
pv  3.80s user 24.19s system 6% cpu 6:40.48 total
wc -c  0.53s user 4.35s system 1% cpu 6:40.48 total
```

Note that this includes 30s ~ compilation/warmup => we're 6:10 vs. 6:27, which isn't so shabby considering that we're also putting less load on the CPU from fewer GCs.
Stats here

* Collections (only minor happening) 351/min down to 332/min
* Average pause stays the same at ~10ms
* Number of overall pauses in the run goes down from 2.2k to 2k (10% reduction!)

This should be (haven't tried it yet, but even the much less efficient #7847 showed a neat aggregate improvement) a lot more fun combined with #7854

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core/src/main/java/org/logstash/Accessors.java
@@ -8,20 +8,17 @@ public final class Accessors {
         //Utility Class
     }
 
-    public static Object get(final ConvertedMap data, final String reference) {
-        final FieldReference field = PathCache.cache(reference);
+    public static Object get(final ConvertedMap data, final FieldReference field) {
         final Object target = findParent(data, field);
         return target == null ? null : fetch(target, field.getKey());
     }
 
-    public static Object set(final ConvertedMap data, final String reference,
+    public static Object set(final ConvertedMap data, final FieldReference field,
         final Object value) {
-        final FieldReference field = PathCache.cache(reference);
         return setChild(findCreateTarget(data, field), field.getKey(), value);
     }
 
-    public static Object del(final ConvertedMap data, final String reference) {
-        final FieldReference field = PathCache.cache(reference);
+    public static Object del(final ConvertedMap data, final FieldReference field) {
         final Object target = findParent(data, field);
         if (target instanceof ConvertedMap) {
             return ((ConvertedMap) target).remove(field.getKey());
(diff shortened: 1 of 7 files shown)


guyboertje:
I like this!!
its all very sensible.
FieldReference dedup ++
Answers the questions:
- What is a FieldReference really?
- What is the PathCache really?

guyboertje:
P.S. I think another review is very optional. This is all internal, no Event API changes for Java plugin authors.

original-brownbear:
@guyboertje thanks!, merging :)

suyograo:
@original-brownbear backport to 6.0 and 5.6 as well please.

original-brownbear:
@suyograo backporting this to `5.6` entails backporting at least also backporting #7839, still do it?

suyograo:
@original-brownbear absolutely! both are great improvements and can go in 5.6

original-brownbear:
@suyograo alright thanks + on it :)

original-brownbear:
done => links:

5.6: #7909 
6.0: #7910

PR #7890, opened 2017-08-03
correct wrong command example

TaurusD:
the original version was misleading because windows won't recognize `/` in paths. So `bin/logstash-plugin` will cause errors.


TaurusD:
I created a topic in [Discuss the Elastic Stack](dicuss.elastic.co) before, which may contain more details. Please refer to [here](https://discuss.elastic.co/t/error-on-offline-plugin-install-verification-failed/95631/5)

PR #7893, opened 2017-08-03
Fix DLQ seek by Timestamp bugs

robbavey:
Fixes 2 seekByTimestamp failure scenarios that occur when a segment is closed with entries smaller than BLOCK_SIZE

If the next segment is > BLOCK_SIZE, then the seek will advance to the next segment. (Skipping past all events
in the closed segment)
If the next segment < BLOCK_SIZE, then the seek will advance to the start of the small segment. (Processing
entries prior to the timestamp)

Backport to 5.5 of #7789


robbavey:
@suyograo Double checking that you want this one in 5.5 too?

suyograo:
+1, good fix.

PR #7894, opened 2017-08-03
PERFORMANCE: Intern String Keys from FieldReference

original-brownbear:
This is a fun/trivial but effective one :)

We are using the `String` keys from `FieldReference` instances in every `HashMap` we create from `set` operations on an event + obviously, we have a ton of duplication between `FieldReference` instances themselves.
This reduces the used old gen size from `212MB` to `200MB` for me and comes with a small but measurable speed-up (likely from a much better CPU cache hit rate).


jakelandis:
LGTM

PR #7898, opened 2017-08-03
PERFORMANCE: Constant RubyHash Visitor

original-brownbear:
Unlike its name suggests `RubyHash.VisitorWithState` has no internal state => no need to set it up again and again for each conversion.

Original JRuby code for this:

```java
    public static abstract class VisitorWithState<T> {
        public abstract void visit(ThreadContext context, RubyHash self, IRubyObject key, IRubyObject value, int index, T state);
    }

    public static abstract class Visitor extends VisitorWithState {
        public void visit(ThreadContext context, RubyHash self, IRubyObject key, IRubyObject value, int index, Object state) {
            visit(key, value);
        }
        public abstract void visit(IRubyObject key, IRubyObject value);
    }
```

Also, we shouldn't get the thread-local context for the case of the `RubyEvent` construction where we clearly have it in the parameter.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ext/JrubyEventExtLibrary.java
+++ b/logstash-core/src/main/java/org/logstash/ext/JrubyEventExtLibrary.java
@@ -87,7 +87,9 @@ public class JrubyEventExtLibrary implements Library {
         public IRubyObject ruby_initialize(ThreadContext context, IRubyObject[] args) {
             final IRubyObject data = args.length > 0 ? args[0] : null;
             if (data instanceof RubyHash) {
-                this.event = new Event(ConvertedMap.newFromRubyHash((RubyHash) data));
+                this.event = new Event(
+                    ConvertedMap.newFromRubyHash(context, (RubyHash) data)
+                );
             } else {
                 initializeFallback(context, data);
             }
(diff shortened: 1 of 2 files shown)


logstash-core/src/main/java/org/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -87,7 +87,9 @@ public Event getEvent() {
         public IRubyObject ruby_initialize(ThreadContext context, IRubyObject[] args) {
             final IRubyObject data = args.length > 0 ? args[0] : null;
             if (data instanceof RubyHash) {
-                this.event = new Event(ConvertedMap.newFromRubyHash((RubyHash) data));
+                this.event = new Event(
+                    ConvertedMap.newFromRubyHash(context, (RubyHash) data)
```
jakelandis: Is this the same context as `o.getRuntime().getCurrentContext()`  ? Not sure if matters, but if they different it could.

original-brownbear: @jakelandis it's the same, but `o.getRuntime().getCurrentContext()` is a thread local soft reference that is populated lazily, so it's much slower to access and JRuby Javadoc recommends not doing it.

jakelandis:
LGTM

PR #7902, opened 2017-08-04
Filter invocation improvements

original-brownbear:
We are doing quite a few operations on the `RubyEvent` that benefit from a proper hashcode and equals method.
The ones we get from `RubyObject` aren't very efficient for our purposes (since the `RubyEvent` holds no state, all we care about is reference equality anyway).

*  Overrode `equals` to only check reference equality
* Implemented a much faster `hashCode`
* Also removed one explicit `return` from the hot loop (no effect really, but I came across https://github.com/jruby/jruby/issues/3715 the other day :D)

Overall this seems to speed up larger batches slightly.

### DIFF
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -44,12 +44,11 @@ module LogStash
       clock.stop
 
       # There is no guarantee in the context of filter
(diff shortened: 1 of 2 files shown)


guyboertje:
I agree. Nice find.

PR #7907, opened 2017-08-04
PERFORMANCE: Don't use slow metric clocks where avoidable

original-brownbear:
These clocks are pretty expensive as they require keeping another object around + instantiating an object for every time span measured. Also, they introduce unnecessary `Block`s messing with the JIT.

This fix simply uses `long` differences in place of the clocks. Brings up the raw throughput of the pipeline (stdin -> pipeline -> stdout) from 85k/s to 95k/s (4 threads, 128 batch size).

### DIFF
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -39,9 +39,11 @@ module LogStash
     def multi_filter(events)
       @metric_events.increment(:in, events.size)
 
-      clock = @metric_events.time(:duration_in_millis)
+      start_time = java.lang.System.current_time_millis
       new_events = @filter.multi_filter(events)
-      clock.stop
+      @metric_events.report_time(
+        :duration_in_millis, java.lang.System.current_time_millis - start_time
+      )
 
       # There is no guarantee in the context of filter
       # that EVENTS_IN == EVENTS_OUT, see the aggregates and
--- a/logstash-core/lib/logstash/instrument/wrapped_write_client.rb
+++ b/logstash-core/lib/logstash/instrument/wrapped_write_client.rb
@@ -19,31 +19,37 @@ module LogStash module Instrument
     end
 
     def push(event)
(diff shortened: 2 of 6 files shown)


logstash-core/lib/logstash/filter_delegator.rb
```diff
@@ -39,9 +39,11 @@ def config_name
     def multi_filter(events)
       @metric_events.increment(:in, events.size)
 
-      clock = @metric_events.time(:duration_in_millis)
+      start_time = java.lang.System.current_time_millis
       new_events = @filter.multi_filter(events)
-      clock.stop
+      @metric_events.report_time(
+        :duration_in_millis, java.lang.System.current_time_millis - start_time
```
jordansissel: Part of the reason we used lambdas and also this 'clock' object was to help reduce errors when writing code.

Can we refactor the duration computation `java.lang.System.current_time_millis - start_time` into a method, like `duration(start_time)` ?

jordansissel: (This would be a great case to use macros, but Java and Ruby don't have them, sadly)

original-brownbear: @jordansissel I'll add a method to keep this clean sure :) sec

original-brownbear: @jordansissel hmm on second thought, just as a public utility on `Metric` then? It seems to me we're just making this harder to read.
I can see the point of having a method for this when the code was:

```rb
MILLISECONDS = 1_000.0.freeze
execution_time = (MILLISECONDS * (Time.now - @start_time)).to_i
```

But now it's literally just the standard JDK call for the *nix time stamp and taking a difference. IMO this makes it even harder to read in Ruby code since Ruby APIs seem to use `s` as a unit a lot and I always have to double check if it's `s` or `ms` to be sure. Here there's no ambiguity?
Maybe just keep this?

original-brownbear:
was superseeded by #7919

PR #7911, opened 2017-08-04
Wait longer to fix build failures on flakey pipeline spec

andrewvc:
This attempts to fix this build failure:
https://logstash-ci.elastic.co/job/elastic+logstash+6.0+dockerubuntu-periodic/95/console
where a time sensitive pipeline spec would fail if too much time elapsed. The patch just increases the maximum wait time.

I couldn't repro the problem, so I'm assuming its due to a slow host. I left a note in there suggesting that if we get this error again we try a different strategy


original-brownbear:
@andrewvc isn't this taken care of by #7863 (which I now realize I didn't backport to `6.0`)?

andrewvc:
Closing in favor of #7863

PR #7752, opened 2017-07-20
Test stabilization: add max retry and timeouts to pipeline spec

jakelandis:
Fixes #7738

PR #7912, opened 2017-08-04
Use rspec wait to fix flakey converge specs

andrewvc:
Some converge specs have races, assuming that a running Agent means that pipelines have started running.
This patch lets us wait a while in the event of a race. One such failure is here: https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-unix-compatibility/os=centos/72/console

Additionally, this removes a harmless double invocation in these specs

### DIFF
--- a/logstash-core/spec/logstash/agent/converge_spec.rb
+++ b/logstash-core/spec/logstash/agent/converge_spec.rb
@@ -74,19 +74,18 @@ describe LogStash::Agent do
         describe "#running_user_defined_pipelines" do
           it "returns the user defined pipelines" do
             start_agent(subject)
-            subject.with_running_user_defined_pipelines do |pipelines|
-              expect(pipelines).to include(:main)
-              expect(pipelines).not_to include(:system_pipeline)
-            end
-            subject.shutdown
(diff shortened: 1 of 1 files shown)


logstash-core/spec/logstash/agent/converge_spec.rb
```diff
@@ -74,19 +74,18 @@
         describe "#running_user_defined_pipelines" do
           it "returns the user defined pipelines" do
             start_agent(subject)
-            subject.with_running_user_defined_pipelines do |pipelines|
-              expect(pipelines).to include(:main)
-              expect(pipelines).not_to include(:system_pipeline)
-            end
-            subject.shutdown
+            wait_for do
```
jakelandis: Does `wait_for` have a timeout limit?

If not can you wrap this in a time out? 

Similar to here: https://github.com/elastic/logstash/pull/7752/files

andrewvc: @jakelandis the default max wait is 10s, which should be perfect here.

jakelandis:
LGTM

Issue #7897, opened 2017-08-03
`org.logstash.Timestamp` Seems Redundant


original-brownbear:
`org.logstash.Timestamp` holds no state beyond the `org.joda.time.DateTime` it wraps.
It seems the only point of having it around is to format timestamps a certain way and facilitate serialization via `org.logstash.json.TimestampSerializer`?

Both the formatting of the timestamp as well as the serialization in a certain format can be achieved by other (faster and 0GC) means from a `long`.
Is there any reason to keep this class around and not replace it with a less heavy approach?

(same goes for related logic around `org.logstash.ext.JrubyTimestampExtLibrary` and the `BiValue` we have for this one)

jakelandis:
I don't think we can get rid of `JrubyTimestampExtLibrary` since it effectively exposes the Joda `DateTime` to Ruby. (right ?) 

It seems that `org.logstash.Timestamp` implements `org.logstash.ackedqueue.Queueable` which  Joda obviously doesn't implement. I am not sure the exact usage, but it seems like a fairly standard pattern of wrapping an object to allow it to conform to a specific interface.  It also seems to bring some consistency to the formatter and simplifies the Joda contract. 

How expensive is the `org.logstash.Timestamp` wrapper ?  

imo, it is a nice abstraction, and wouldn't want to toss it unless there are measurable performance gains.

original-brownbear:
@jakelandis 

> I don't think we can get rid of JrubyTimestampExtLibrary since it effectively exposes the Joda DateTime to Ruby. (right ?)

Well, it does expose that to Ruby, but it's certainly not a very fast way of doing so. Since all of this stuff has no (non-redundant) state, you can just get the same effect without the memory cost with some `static` converter (by definition :P).

> It seems that org.logstash.Timestamp implements org.logstash.ackedqueue.Queueable which Joda obviously doesn't implement.

I don't think `Queueable` is even used on this thing outside of tests. Even if it is, `Queueable` is kind of a pointless interface in the first place if you think about it. This could again simply be cleaned up by a static utility with the signature `Queueable method(long var)` (should we actually need this).

> it seems like a fairly standard pattern of wrapping an object to allow it to conform to a specific interface.

Not in the hot path, especially we're just using way too much memory already, looking at how visible every small improvement is in throughput (we're still much much heavier than the old Ruby event :().

> How expensive is the org.logstash.Timestamp wrapper ?

[...]

PR #7916, opened 2017-08-04
 PERFORMANCE: Cleanup Redundant Timestamp Instantiations

original-brownbear:
Cleaned the use of `Timestamp` up a little to improve code quality and illustrate its redundant nature better for #7897.

Cleanups:

* No reason for this thing to be `Cloneable`, we never clone it and the `clone` implementation makes no sense on an immutable `Object` (`DateTime` is immutable and the only reason it wasn't `final` was that `clone` method calling the redundant `setTime`) => made the thing immutable
  * In the same spirit, removed the redundant copy constructor `public Timestamp(Timestamp t)`, it just makes no sense and only adds GC + complication
   * E.g. this pointless copy `readManager.seekToNextEvent(new Timestamp(seekTarget))` now simply being `readManager.seekToNextEvent(seekTarget)`
* Also removed `public Timestamp(Long epoch_milliseconds)`, this is redundant if you have `public Timestamp(long epoch_milliseconds)`, the JVM figures this out by itself if you pass a `Long` to `long`
* Cleaned up the logic of `public static IRubyObject ruby_coerce` by removing `public static Timestamp newTimestamp(IRubyObject time)`
   * `public static Timestamp newTimestamp(IRubyObject time)` was used here and in `public static IRubyObject ruby_parse_iso8601`. In `public static IRubyObject ruby_parse_iso8601` we don't need that weird type handling because we already know and checked that we're handling a `RubyString`. In the other case we can deal with logic much cleaner too. Especially now avoid going through:
```rb
else if (time instanceof RubyTimestamp) {		
                return new Timestamp(((RubyTimestamp) time).timestamp);
```
only to then turn the `Timestamp` into a `RubyTimestamp` again in the next method call.

=> same logic with less code + it's much clearer now which conversions are actually occuring
=> affected paths produce less GC now

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Timestamp.java
+++ b/logstash-core/src/main/java/org/logstash/Timestamp.java
@@ -11,11 +11,16 @@ import org.joda.time.format.ISODateTimeFormat;
 import org.logstash.ackedqueue.Queueable;
 import org.logstash.json.TimestampSerializer;
 
+/**
+ * Wrapper around a {@link DateTime} with Logstash specific serialization behaviour.
+ * This class is immutable and thread-safe since its only state is held in a final {@link DateTime}
+ * reference and {@link DateTime} which itself is immutable and thread-safe.
+ */
 @JsonSerialize(using = TimestampSerializer.class)
-public final class Timestamp implements Cloneable, Comparable<Timestamp>, Queueable {
+public final class Timestamp implements Comparable<Timestamp>, Queueable {
 
     // all methods setting the time object must set it in the UTC timezone
-    private DateTime time;
+    private final DateTime time;
 
     private static final DateTimeFormatter iso8601Formatter = ISODateTimeFormat.dateTime();
 
@@ -29,15 +34,7 @@ public final class Timestamp implements Cloneable, Comparable<Timestamp>, Queuea
         this.time = ISODateTimeFormat.dateTimeParser().parseDateTime(iso8601).toDateTime(DateTimeZone.UTC);
     }
 
-    public Timestamp(Timestamp t) {
-        this.time = t.getTime();
-    }
-
     public Timestamp(long epoch_milliseconds) {
-        this.time = new DateTime(epoch_milliseconds, DateTimeZone.UTC);
-    }
-
-    public Timestamp(Long epoch_milliseconds) {
         this.time = new DateTime(epoch_milliseconds, DateTimeZone.UTC);
     }
 
@@ -51,10 +48,6 @@ public final class Timestamp implements Cloneable, Comparable<Timestamp>, Queuea
 
     public DateTime getTime() {
         return time;
-    }
-
-    public void setTime(DateTime time) {
-        this.time = time.toDateTime(DateTimeZone.UTC);
     }
 
     public static Timestamp now() {
@@ -77,14 +70,7 @@ public final class Timestamp implements Cloneable, Comparable<Timestamp>, Queuea
 
     @Override
     public int compareTo(Timestamp other) {
-        return getTime().compareTo(other.getTime());
-    }
-
-    @Override
-    public Timestamp clone() throws CloneNotSupportedException {
-        Timestamp clone = (Timestamp)super.clone();
-        clone.setTime(this.getTime());
-        return clone;
+        return time.compareTo(other.time);
     }
 
     @Override
--- a/logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
+++ b/logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
@@ -159,28 +159,22 @@ public class JrubyTimestampExtLibrary implements Library {
             return RubyString.newString(context.runtime,  "\"" + this.timestamp.toIso8601() + "\"");
         }
 
-        public static Timestamp newTimestamp(IRubyObject time)
-        {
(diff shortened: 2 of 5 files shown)


logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
```diff
@@ -254,5 +248,9 @@ public IRubyObject ruby_year(ThreadContext context)
         {
             return RubyFixnum.newFixnum(context.runtime, this.timestamp.getTime().getYear());
         }
+
+        private static RubyTimestamp fromRString(final Ruby runtime, final RubyString string) {
+            return RubyTimestamp.newRubyTimestamp(runtime, new Timestamp(string.toString()));
```
jakelandis: is `string`  guaranteed to always be a non-null ?

I _think_ old implementation would return a `Timestamp` of _now_, where this will result in `NPE`.

logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
```diff
@@ -159,28 +159,22 @@ public IRubyObject ruby_to_json(ThreadContext context, IRubyObject[] args)
...
-
         @JRubyMethod(name = "coerce", required = 1, meta = true)
-        public static IRubyObject ruby_coerce(ThreadContext context, IRubyObject recv, IRubyObject time)
-        {
+        public static IRubyObject ruby_coerce(ThreadContext context, IRubyObject recv,
+            IRubyObject time) {
             try {
-                Timestamp ts = newTimestamp(time);
-                return (ts == null) ? context.runtime.getNil() : RubyTimestamp.newRubyTimestamp(context.runtime, ts);
+                if (time instanceof RubyTimestamp) {
```
jakelandis: I _think_ you are missing the `time.isNil()` check to return a _now_  `TimeStamp`, which is slightly different behavior then `time == null`

logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
```diff
@@ -159,28 +159,22 @@ public IRubyObject ruby_to_json(ThreadContext context, IRubyObject[] args)
...
-               return null;
-            }
-        }
-
-
         @JRubyMethod(name = "coerce", required = 1, meta = true)
-        public static IRubyObject ruby_coerce(ThreadContext context, IRubyObject recv, IRubyObject time)
-        {
+        public static IRubyObject ruby_coerce(ThreadContext context, IRubyObject recv,
+            IRubyObject time) {
```
original-brownbear: @jakelandis pssst :D I thought I'd get away with it, this line was really long :P but I can revert it to turn down the noise :)

logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
original-brownbear: @jakelandis this is only called when guarded by an `instanceof` on `string` upstream => jup, `string` is never `null`.

logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
original-brownbear: @jakelandis there were no tests (neither in Java nor in Ruby) for this branch of the logic that use `RubyNil` + the only way to get `RubyNil` here is to explicitly pass `nil` (not passing anything causes `null` to arrive here, I tried this out for some other PR the other day) from the related Ruby code.
I'd just drop this branch as dead code since it's an internal API.

logstash-core/src/main/java/org/logstash/Timestamp.java
```diff
@@ -12,10 +12,10 @@
 import org.logstash.json.TimestampSerializer;
 
 @JsonSerialize(using = TimestampSerializer.class)
-public final class Timestamp implements Cloneable, Comparable<Timestamp>, Queueable {
+public final class Timestamp implements Comparable<Timestamp>, Queueable {
```
original-brownbear: @jakelandis fair point, actually there are some sensible things to add here, will do in a bit.

original-brownbear:
@jakelandis reverted formatting changes and added some Javadoc to `Timestamp`. The `nil` behavior I'd just change as was done here. It's dead code + we really should be moving to more type safety wherever possible in general, going through these endless `instanceof` chains is just one of the worst antipatterns in Java :(

logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
jakelandis: thanks, and sorry didn't mean to make you do work over something so trivial.  I just wanted to start a larger conversation about picking a style format for Java and apply it consistently.  Lets have that conversation outside this PR :)

jakelandis:
LGTM

logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
original-brownbear: @jakelandis np :) (I think as a general note, picking a Java code style is in the works already but yes, we can discuss that elsewhere :))

PR #7917, opened 2017-08-04
PERFORMANCE: Removed DoubleBiValue and FloatBiValue

original-brownbear:
Same as #7854 but for `double` and `float` types. Moved all of the Java handling to `double` since Ruby doesn't make a distinction here either and this will safely make no distinction between `double` and `float`.

Not much to add otherwise. I simply removed the classes like was done for `StringBiValue` and short-circuited the `Valuefier`, `Javafier` and `Rubyfier` + added a custom serializer.


original-brownbear:
@guyboertje thanks!

Issue #7913, opened 2017-08-04
Trouble aborting pipeline using ctrl-c


tbragin:
version: 6.0.0-beta1 BC7

pipeline:
```
input {
  generator {
    count => 0
    message => "test message"
  }
}

filter {
  sleep {
    time => "1"
    every => 1
  }
}

output {
  stdout { codec => dots }
  elasticsearch {
    hosts => ["http://localhost:9200"]
    user => "elastic"
    password => "elastic"
  }
}
```

had to submit ctrl-c many dozen times, in the meanwhile events kept coming and occasionally similar messages appeared
`[2017-08-04T11:19:54,882][WARN ][logstash.shutdownwatcher ] {"inflight_count"=>250, "stalling_thread_info"=>{["LogStash::Filters::Sleep", {"time"=>"1", "every"=>1, "id"=>"15ad4fc181ff6a55b7e6c89c55c67d102ddcb1f9-2"}]=>[{"thread_id"=>36, "name"=>"[main]>worker1", "current_call"=>"[...]/vendor/bundle/jruby/2.3.0/gems/logstash-filter-sleep-3.0.4/lib/logstash/filters/sleep.rb:105:in `sleep’“}, {“thread_id”=>38, “name”=>“[main]>worker3”, “current_call”=>“[...]/vendor/bundle/jruby/2.3.0/gems/logstash-filter-sleep-3.0.4/lib/logstash/filters/sleep.rb:105:in `sleep'"}]}}`

andrewvc:
@tbragin this is because of the use of the sleep filter. Logstash operates in batches, the default batch size is 125. Logstash also does not want to stop if there are in-flight events, instead waiting for any inflight batches to finish processing first.

In this case, Logstash will wait at least 125 seconds to stop due to that sleep filter. This is what you want in production, but not when testing with the sleep filter.

You can avoid this situation by overriding the batch size with the `-b` option. For example: `bin/logstash -b 1 -e 'input { generator {} } filter { sleep { time => 1 } } output { stdout {} }'`

I'm going to close this since this is expected behavior.

tbragin:
Thanks @andrewvc !

jsvd:
[...]

PR #7918, opened 2017-08-04
prevent force shutdown to be blocked by with_pipelines

jsvd:
if force shutdown uses `with_pipelines` then it will never acquire the
`@pipelines_lock` from `shutdown_pipelines` until it terminates.

This PR makes force shutdown use `@pipelines` directly, fixing it.

fixes https://github.com/elastic/logstash/issues/7913

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -190,13 +190,6 @@ class LogStash::Agent
     converge_result
   end
 
-  def force_shutdown!
-    stop_collecting_metrics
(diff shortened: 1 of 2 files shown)


original-brownbear:
@jsvd if you're sure we'll never get a concurrent modification on that hash (adding a pipeline or removing one from another thread. At least I couldn't find one so I'm optimistic :)) LGTM.

jsvd:
After speaking with @andrewvc we just decided to not try to make graceful shutdown work by circumventing the lock and to just exit asap.

Issue #7904, opened 2017-08-04
Metrics Performance is Trivial to Fix! (Analysis of the 5.3 -> 5.4 Regression Inside)


original-brownbear:
Long story short: **The issue with the metrics performance regression is trivial to fix.**

The mistake made in going from `5.3` to `5.4` was that things moved from:

* Lookup and cache counter from `MetricStore` and call e.g. `counter.increment(5)` on it

to 

* Lookup the counter for every change of the metric + use dynamic `send` to invoke the metric method

So the way to fix the metrics performance is simply to:

Go to all places where we use `NameSpacedMetric` and replace the instantiation of it by getting the correct counter (`Gauge`-type object) from `MetricStore` using `fetch_or_store` and cache it.

Also, it's probably a good idea to replace `Gauge.execute` which is currently this:

```rb
    def execute(action, value = nil)
      send(action, value)
    end
```

with actual methods that pass values to the Java logic, `send` will never be optimized here by the JIT I think.

Lots of typing but pretty easy. Note that I did this test wise just for the output counter in the wrapped batch writer and saw about a 20% increase in throughput ... this is def. where the problems are coming from.

Obviously, part of it is also that we are running more metrics now, but if we just cache all counters we should be transparent no matter what. The cost for actually incrementing counters shows as `~0%` in profiling :)

original-brownbear:
@suyograo FYI, you were right all along there was a huge regression in metrics :( Trivial but a little time consuming to fix (pretty safe fix though, could probably back port it all the way).

jakelandis:
@original-brownbear - I am pretty comfortable with this code, there are are couple of key places that I can quickly (and confidently) change from `send` to the actual method.

original-brownbear:
@jakelandis note that the `send` is a much smaller part of the overhead compared to the lookups. If I comment out the send call I get a visible maybe ~15% increase in throughput, if I comment out the lookups get a 3x speedup in raw pipeline throughput.
The lookups are def key here and they weren't even perfectly cached in `5.3`, just a bit less indirection already made a huge impact there.

jakelandis:
@original-brownbear - the `send` is pretty trivial and safe to fix. However caching the metrics outside of the metrics store would require keeping those caches in sync with the pruning (cache invalidation) behavior when restarting a pipeline. The lookup behavior will be removed with #7788, but it may be a bit before that is complete, and is only targeted for 6.x.

original-brownbear:
[...]

PR #7919, opened 2017-08-05
#7904 Faster Metrics by Caching Counters

original-brownbear:
For #7904 also incorporates #7907 

First things first, we're almost twice as fast with this, as the fastest version (`5.3.0`) I could find in the baseline benchmark.

```bash
Logstash Benchmark
Running Test Case: baseline (x3)
Start Time: Sat 8 5 20:20:46 2017 CEST
Statistical Summary:

Elapsed Time: 34s
Num Events: 2982064
Throughput Min: 250.00
Throughput Max: 310750.00
Throughput Mean: 248505.33
Throughput StdDev: 107847.10
Throughput Variance: 11630996629.52
Mean CPU Usage: 30.75%

➜  logstash git:(faster-metrics) bin/benchmark.sh --distribution-version=5.3.0 -repeat-data=3
Logstash Benchmark
Benchmarking Version: 5.3.0
Running Test Case: baseline (x3)
Using Logstash 5.3.0 from cache.
Start Time: Sat 8 5 12:55:27 2017 CEST
Statistical Summary:

Elapsed Time: 31s
Num Events: 2999888
Throughput Min: 1633.00
Throughput Max: 142875.00
Throughput Mean: 124995.33
Throughput StdDev: 30999.54
Throughput Variance: 960971473.71
Mean CPU Usage: 29.04%
```

By comparison `master` is ~3x slower in the baseline:

```sh
➜  logstash git:(master) bin/benchmark.sh --local-path=${PWD} -repeat-data=3                                                                          
Logstash Benchmark
Running Test Case: baseline (x3)
Start Time: Sat 8 5 13:00:45 2017 CEST
Statistical Summary:

Elapsed Time: 66s
Num Events: 2978307
Throughput Min: 0.00
Throughput Max: 79875.00
Throughput Mean: 72641.63
Throughput StdDev: 17982.31
Throughput Variance: 323363589.59
Mean CPU Usage: 25.02%
```

(As a side note: It's kind of interesting to see how much time we were losing bc of lock contention before judging from the CPU usage `%`, worker threads all went from `80%` hot time to `90%+`)

For the Apache dataset, I see a throughput improvement of about a third (1.33x faster) over `master`. Haven't had time for large sample sizes here, though so I'd say I'm confident to have at least a `25%` speed-up on my i7 (the whole thing is CPU bound anyway).

Fixed by:
* Caching `:counter` type `Gauge`s
   * Reload is not a problem since the `FilterDelegator`, `OutputDelegator` and `WrappedWriteClient` client that cache these are instantiated fresh anyways (**please double check this as I could be missing some piece of this puzzle, 95% confident in this though**)
* Had to adjust some specs, but I think the changes are somewhat self-explanatory and rather improve than worsen the strength of the tests (imo)

PS: This makes the metric overhead almost non-existent even in the baseline benchmark. It seems to be `< 10%`  easily for `baseline` and is almost invisible in the `apache` version (obv gotta compare `| pv | wc -c` to see that).

### DIFF
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -28,6 +28,7 @@ module LogStash
       @metric_events = namespaced_metric.namespace(:events)
       @metric_events_in = @metric_events.counter(:in)
       @metric_events_out = @metric_events.counter(:out)
+      @metric_events_time = @metric_events.counter(:duration_in_millis)
       namespaced_metric.gauge(:name, config_name)
 
       # Not all the filters will do bufferings
@@ -43,15 +44,13 @@ module LogStash
 
       start_time = java.lang.System.current_time_millis
       new_events = @filter.multi_filter(events)
-      @metric_events.report_time(
-        :duration_in_millis, java.lang.System.current_time_millis - start_time
-      )
+      @metric_events_time.increment(java.lang.System.current_time_millis - start_time)
 
       # There is no guarantee in the context of filter
       # that EVENTS_IN == EVENTS_OUT, see the aggregates and
       # the split filter
       c = new_events.count { |event| !event.cancelled? }
-      @metric_events_out.increment(:out) if c > 0
+      @metric_events_out.increment(c) if c > 0
       new_events
     end
 
@@ -64,7 +63,7 @@ module LogStash
 
         # Filter plugins that does buffering or spooling of events like the
         # `Logstash-filter-aggregates` can return `NIL` and will flush on the next flush ticks.
-        @metric_events.increment(:out, new_events.size) if new_events && new_events.size > 0
+        @metric_events_out.increment(new_events.size) if new_events && new_events.size > 0
         new_events
       end
     end
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -26,6 +26,8 @@ module LogStash
       @filter.execution_context = execution_context
 
       @metric_events = namespaced_metric.namespace(:events)
+      @metric_events_in = @metric_events.counter(:in)
+      @metric_events_out = @metric_events.counter(:out)
       namespaced_metric.gauge(:name, config_name)
 
       # Not all the filters will do bufferings
@@ -37,7 +39,7 @@ module LogStash
     end
 
     def multi_filter(events)
-      @metric_events.increment(:in, events.size)
+      @metric_events_in.increment(events.size)
 
       clock = @metric_events.time(:duration_in_millis)
       new_events = @filter.multi_filter(events)
@@ -47,7 +49,7 @@ module LogStash
       # that EVENTS_IN == EVENTS_OUT, see the aggregates and
       # the split filter
       c = new_events.count { |event| !event.cancelled? }
-      @metric_events.increment(:out, c) if c > 0
+      @metric_events_out.increment(:out) if c > 0
       new_events
     end
 
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -41,9 +41,11 @@ module LogStash
     def multi_filter(events)
       @metric_events_in.increment(events.size)
 
-      clock = @metric_events.time(:duration_in_millis)
+      start_time = java.lang.System.current_time_millis
       new_events = @filter.multi_filter(events)
-      clock.stop
+      @metric_events.report_time(
+        :duration_in_millis, java.lang.System.current_time_millis - start_time
+      )
 
       # There is no guarantee in the context of filter
       # that EVENTS_IN == EVENTS_OUT, see the aggregates and
--- a/logstash-core/lib/logstash/instrument/wrapped_write_client.rb
+++ b/logstash-core/lib/logstash/instrument/wrapped_write_client.rb
@@ -11,8 +11,11 @@ module LogStash module Instrument
       @pipeline_metrics = metric.namespace([:stats, :pipelines, pipeline_id, :events])
       @plugin_events_metrics = metric.namespace([:stats, :pipelines, pipeline_id, :plugins, plugin_type, plugin.id.to_sym, :events])
       @events_metrics_counter = @events_metrics.counter(:in)
+      @events_metrics_time = @events_metrics.counter(:queue_push_duration_in_millis)
       @pipeline_metrics_counter = @pipeline_metrics.counter(:in)
+      @pipeline_metrics_time = @pipeline_metrics.counter(:queue_push_duration_in_millis)
       @plugin_events_metrics_counter = @plugin_events_metrics.counter(:out)
+      @plugin_events_metrics_time = @plugin_events_metrics.counter(:queue_push_duration_in_millis)
       define_initial_metrics_values
     end
 
@@ -48,20 +51,18 @@ module LogStash module Instrument
 
     def report_execution_time(start_time)
       execution_time = java.lang.System.current_time_millis - start_time
-      @events_metrics.report_time(:queue_push_duration_in_millis, execution_time)
-      # Reuse the same values for all the endpoints to make sure we don't have skew in times.
-      @pipeline_metrics.report_time(:queue_push_duration_in_millis, execution_time)
-      @plugin_events_metrics.report_time(:queue_push_duration_in_millis, execution_time)
+      @events_metrics_time.increment(execution_time)
+      @pipeline_metrics_time.increment(execution_time)
+      @plugin_events_metrics_time.increment(execution_time)
     end
 
     def define_initial_metrics_values
       @events_metrics_counter.increment(0)
       @pipeline_metrics_counter.increment(0)
       @plugin_events_metrics_counter.increment(0)
-
-      @events_metrics.report_time(:queue_push_duration_in_millis, 0)
-      @pipeline_metrics.report_time(:queue_push_duration_in_millis, 0)
-      @plugin_events_metrics.report_time(:queue_push_duration_in_millis, 0)
+      @events_metrics_time.increment(0)
+      @pipeline_metrics_time.increment(0)
+      @plugin_events_metrics_time.increment(0)
     end
   end
 end end
--- a/logstash-core/lib/logstash/instrument/wrapped_write_client.rb
+++ b/logstash-core/lib/logstash/instrument/wrapped_write_client.rb
@@ -21,31 +21,37 @@ module LogStash module Instrument
     end
 
     def push(event)
-      record_metric { @write_client.push(event) }
+      increment_counters(1)
+      start_time = java.lang.System.current_time_millis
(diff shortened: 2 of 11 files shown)


original-brownbear:
@suyograo @jordansissel @jakelandis FYI, this would be my least impact fix here. Seems to work fine from manual testing and specs all passing and is blazing fast.
(don't worry about the overall event count error in the benchmark btw, this is not a bug but just a result of the benchmark only measuring once every second so it can lose some events in the last measurement interval when LS stops)

jakelandis:
@original-brownbear - Can you also run @jsvd 's [script]( https://github.com/elastic/logstash/issues/7396#issuecomment-307845571) and post the results over there.  (before and after so we are not comparing how fast your computer is vs. mine)

My notes for running this:

cd and place the script (profile.rb) in logstash-core/lib
```
export JRUBY_OPTS="--profile.api"
ruby -I . profile.rb
```

logstash-core/lib/logstash/output_delegator.rb
```diff
@@ -42,11 +43,13 @@ def register
...
 
   def multi_receive(events)
-    @metric_events.increment(:in, events.length)
-    clock = @metric_events.time(:duration_in_millis)
+    @in_counter.increment(events.length)
+    start_time = java.lang.System.current_time_millis
     @strategy.multi_receive(events)
-    clock.stop
-    @metric_events.increment(:out, events.length)
+    @metric_events.report_time(
```
original-brownbear: @jakelandis hah, man my concentration :D fixing :)

logstash-core/lib/logstash/filter_delegator.rb
```diff
@@ -37,17 +39,19 @@ def config_name
...
 
     def multi_filter(events)
-      @metric_events.increment(:in, events.size)
+      @metric_events_in.increment(events.size)
 
-      clock = @metric_events.time(:duration_in_millis)
+      start_time = java.lang.System.current_time_millis
       new_events = @filter.multi_filter(events)
-      clock.stop
+      @metric_events.report_time(
```
original-brownbear: @jakelandis ah gnarf I just noticed I rebased wrong :( no need to find all instances of these, none of them were fixed. Funny ... we can be even faster it seems sec :D

logstash-core/lib/logstash/instrument/wrapped_write_client.rb
```diff
@@ -10,7 +10,12 @@ def initialize(write_client, pipeline, metric, plugin)
       @events_metrics = metric.namespace([:stats, :events])
       @pipeline_metrics = metric.namespace([:stats, :pipelines, pipeline_id, :events])
       @plugin_events_metrics = metric.namespace([:stats, :pipelines, pipeline_id, :plugins, plugin_type, plugin.id.to_sym, :events])
-
+      @events_metrics_counter = @events_metrics.counter(:in)
```
original-brownbear: @jakelandis from the code it seems to me we're safe here. Also, we'd see a bug with these in `master`:

```rb
@plugin_events_metrics = metric.namespace([:stats, :pipelines, pipeline_id, :plugins, plugin_type, plugin.id.to_sym, :events])
```

These metrics are clearly related to the `pipeline_id`, so if those are still fine on reloads and such, then we know `@events_metrics.counter(:in)` in the same block is also called again on reloads and hence any `prune` that happened will show up here.
Not that we only cache the terminal nodes of each `metric` so to say, each call to like `@events_metrics.counter(:in)` will still go to the metrics store and get us a "fresh" counter :)
For new plugins added we're good for that very same reason, since adding new plugins entails a call to `metric.namespace`.

original-brownbear:
@jakelandis wohoo you're the best for spotting the duration's I missed! 220k/s => 250k/s baseline :)

original-brownbear:
@jakelandis as for that script, I think that's pretty pointless after these changes and the fact that your work here moved us to `LongAdder` => we're basically benchmarking how warm my laptop is (since this goes 1:1 with the clock rate and is completely single threaded and the i7's boost become super visible here) :D 
Also, I think you have the same machine I do, but I can do that later when I find a few minutes.

original-brownbear:
@jakelandis that script is broken. If I run it without changes it loads `5.5.0` from some dependency we have in the path (from the ITs or so not sure where it's coming from).
I can do this though:

```rb
start = java.lang.System.current_time_millis
i = 0
while i < 5_000_000
  $obj.change_one_metric
  i += 1
end

puts java.lang.System.current_time_millis - start
```

-> `750ms` => 6.7M/s ... not too shabby given that pure `AtomicLong` incrementation runs at just `7.9M/s` on my box => good enough now imo ... counting up metrics doesn't even show up as overhead (metrics off and on is the same speed for me now) after fixing the spots with duration you pointed out.

original-brownbear:
basically judging from this, the only optimization left would be moving the `LongAdder` to Ruby, since invokedynamic is broken in current JRuby.
Currently, we're simply seeing this:

```rb
<task_queued method="org/logstash/instrument/metrics/counter/LongCounter increment (J)V" level="3" bytes="25" count="7650" stamp="21.946" comment="tiered" hot_count="7650" compile_id="12959" iicount="7650"/>

<nmethod stub_offset="880" dependencies_offset="1336" address="0x000000011a445e10" method="org/logstash/instrument/metrics/counter/LongCounter increment (J)V" level="3" count="8168" stamp="21.948" nul_chk_table_offset="1344" scopes_data_offset="1088" iicount="8168" oops_offset="1064" entry="0x000000011a445fa0" size="1368" scopes_pcs_offset="1192" insts_offset="400" bytes="25" relocation_offset="296" compile_id="12959" compiler="C1"/>

<task method="org/logstash/instrument/metrics/counter/LongCounter increment (J)V" level="3" bytes="25" count="8084" stamp="21.948" compile_id="12959" iicount="8084">
  <phase name="buildIR" stamp="21.948">
    <type name="void" id="724"/>
    <type name="long" id="723"/>
    <klass name="org/logstash/instrument/metrics/counter/LongCounter" flags="1" id="832"/>
    <method level="2" bytes="25" name="increment" flags="1" holder="832" arguments="723" id="833" compile_id="12436" compiler="C1" iicount="8090" return="724"/>
    <parse method="833" stamp="21.948">
      <bc code="183" bci="12"/>
      <klass name="java/lang/String" flags="17" id="730"/>
      <klass name="java/lang/UnsupportedOperationException" flags="1" id="836"/>
      <method bytes="6" name="&lt;init&gt;" flags="1" holder="836" arguments="730" id="838" iicount="1" return="724"/>
      <call method="838" instr="invokespecial"/>
      <inline_fail reason="not inlineable"/>
      <bc code="182" bci="21"/>
      <klass name="java/util/concurrent/atomic/LongAdder" flags="1" id="835"/>
      <method level="3" bytes="90" name="add" flags="1" holder="835" arguments="723" id="839" compile_id="12731" compiler="C1" iicount="46531" return="724"/>
      <call method="839" instr="invokevirtual"/>
      <dependency ctxk="835" type="leaf_type"/>
      <inline_fail reason="callee is too large"/>
      <dependency ctxk="835" type="leaf_type"/>
      <parse_done stamp="21.948"/>
    </parse>
    <phase name="optimize_blocks" stamp="21.948">
      <phase_done name="optimize_blocks" stamp="21.948"/>
    </phase>
    <phase name="rangeCheckElimination" stamp="21.948">
      <phase_done name="rangeCheckElimination" stamp="21.948"/>
    </phase>
    <phase name="optimize_null_checks" stamp="21.948">
      <phase_done name="optimize_null_checks" stamp="21.948"/>
    </phase>
    <phase_done name="buildIR" stamp="21.948"/>
  </phase>
  <phase name="emit_lir" stamp="21.948">
    <phase name="lirGeneration" stamp="21.948">
      <phase_done name="lirGeneration" stamp="21.948"/>
    </phase>
    <phase name="linearScan" stamp="21.948">
      <phase_done name="linearScan" stamp="21.948"/>
    </phase>
    <phase_done name="emit_lir" stamp="21.948"/>
  </phase>
  <phase name="codeemit" stamp="21.948">
    <phase_done name="codeemit" stamp="21.948"/>
  </phase>
  <phase name="codeinstall" stamp="21.948">
    <dependency ctxk="835" type="leaf_type"/>
    <phase_done name="codeinstall" stamp="21.948"/>
  </phase>
  <code_cache nmethods="9396" free_code_cache="212437504" adapters="620" total_blobs="10106"/>
  <task_done success="1" count="8174" stamp="21.948" nmsize="664"/>
</task>
```

together with complete inlining of all counter increment calls on the JRuby side => impossible to speed up really (could try turning the Java class into Ruby modules to be invoked more efficienty, but see next sentence :)). As a matter of fact, we probably don't even want this thing to get inlined, since longadder adding is 70+ byte we could see some bad behaviour here from blowing up the methods it gets inlined into.

jakelandis:
@original-brownbear - thanks for the updates and comments.  I want to take this for quick test spin on Monday before final approval.  

Overall this is looking really promising and great work!

jordansissel:
> since invokedynamic is broken in current JRuby.

@original-brownbear Is there an open issue on JRuby for this?

original-brownbear:
@jordansissel jup https://github.com/jruby/jruby/issues/3843 :)

jordansissel:
Aah so this only impacts 9k, thanks!

jakelandis:
>  from the code it seems to me we're safe here. Also, we'd see a bug with these in master

It seems that we do have some bugs in master w/r/t to reloading metrics. I would suggest that fix those bugs before proceeding with this issue. 

https://github.com/elastic/logstash/issues/7925
https://github.com/elastic/logstash/issues/7926

original-brownbear:
@jakelandis @jordansissel 

I'll look into those now. I'd suggest moving ahead here regardless since those aren't caused by the fixes here. The reasoning is this:
After many iterations, we finaly have performance neutral metrics here. I'd rather start from than find fixes that keep things as fast as before than trying to apply this very short patch to whatever comes out at the other end of fixing those two. Otherwise, I'll have to compare performance across so many revisions (baseline is before the fixes to those two then) to gauge the impact that a fix for those will have (potentially).

Makes sense?

jakelandis:
@original-brownbear - I was originally worried that there were latent bugs that could mask potential issues w/r/t the caching during reload. (#7925 was throwing off my ability to manually test)
 
However, we found the root cause of #7925 and am reverting the commit that caused it. #7926 is not relevant here once #7925 is fixed.

Let me re-test with the fix for #7925 merged into this.

original-brownbear:
@jakelandis thanks, I saw. Haha (in theory) reverting that should be performance neutral for this branch (only a hit for `master`). Let's see if the theory holds :D

jakelandis:
@original-brownbear - Code looks good, and a few ad-hoc tests w/r/t reloading all looks good. 

LGTM

original-brownbear:
@jakelandis alrighty merging :(, many thanks for the big effort in testing/reviewing this!

original-brownbear:
@suyograo `6.0` and `5.6` need backports, will do that later/tomorrow as discusssed

PR #7920, opened 2017-08-06
CLEANUP: Remove unused TimeBiValue

original-brownbear:
This is a very easy one :)

The only place `TimeBiValue` was used in production was this code:

```java
            } else if (o instanceof TimeBiValue) {		
                return new Timestamp(((TimeBiValue) o).javaValue());
```

There were no references to its constructor(s) anywhere (it's not used in the mapping in `BiValues`) => this branch in `Event` is dead code and so is the whole class.


jordansissel:
I'm relying mostly on tests passing to verify this refactor.

LGTM :)

PR #7924, opened 2017-08-07
PERFORMANCE: Generate more optimal filter_func

original-brownbear:
Baseline performance 250k/s -> 270k/s (had a better number here, but had to remove one step of the optimization for JRuby issues.)
Basically what this does is just not add the debug logging lines to the generated methods when debug logging is disabled, this has a pretty measurable effect on performance since that call (`.debug?`) is somewhat expensive + blows up method size needlessly since the dead code removal doesn't work all that well in JRuby code it seems.

### DIFF
--- a/logstash-core/lib/logstash/config/config_ast.rb
+++ b/logstash-core/lib/logstash/config/config_ast.rb
@@ -43,7 +43,7 @@ module LogStash; module Config; module AST
     end
 
 
-    def compile
+    def compile(debug_logging)
       LogStash::Config::AST.deferred_conditionals = []
       LogStash::Config::AST.deferred_conditionals_index = 0
       LogStash::Config::AST.plugin_instance_index = 0
@@ -60,7 +60,7 @@ module LogStash; module Config; module AST
 
       sections = recursive_select(LogStash::Config::AST::PluginSection)
       sections.each do |s|
-        code << s.compile_initializer
+        code << s.compile_initializer(debug_logging)
       end
 
       # start inputs
@@ -75,12 +75,12 @@ module LogStash; module Config; module AST
         definitions << "define_singleton_method :#{type}_func do |event|"
         definitions << "  targeted_outputs = []" if type == "output"
         definitions << "  events = [event]" if type == "filter"
-        definitions << "  @logger.debug? && @logger.debug(\"#{type} received\", \"event\" => event.to_hash)"
-
-        sections.select { |s| s.plugin_type.text_value == type }.each do |s|
-          definitions << s.compile.split("\n", -1).map { |e| "  #{e}" }
+        if debug_logging
+          definitions << "  @logger.debug(\"#{type} received\", \"event\" => event.to_hash)"
         end
-
+        sections.select {|s| s.plugin_type.text_value == type}.each do |s|
+          definitions << s.compile.split("\n", -1).map {|e| "  #{e}"}
+        end
         definitions << "  events" if type == "filter"
         definitions << "  targeted_outputs" if type == "output"
         definitions << "end"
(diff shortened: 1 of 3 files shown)


original-brownbear:
@suyograo FYI, this is the minimal version of the removal of the `debug?` check we talked about earlier.

suyograo:
@original-brownbear 6.0 and 5.6 as well please.

original-brownbear:
ah `5.6` needs manual intervention :) will do very soon

jordansissel:
We have an API that lets users change the log level at runtime. This change means users cannot change the log level of the compiled pipeline without reloading/restarting Logstash.

original-brownbear:
@suyograo @jordansissel uff true ... also this explains why the `debug?` check is so expensive ... we're not using a `static` logger.
umm so revert?

suyograo:
hmm, good point @jordansissel, I missed that. I guess we need to revert to support that functionality. We can open a separate discussion if folks are using this feature and we can assess supporting this.

jordansissel:
@original-brownbear the `@logger.debug?` check is an optimization itself to avoid garbage creation on a method that may do nothing.

Some background experiments from 5 years ago are here: https://github.com/jordansissel/experiments/tree/master/ruby/logger-string-vs-block

The context is this: Parameters passed to a logger call may be expensive, such as `logger.info("Hello", :foo => someSlowFunction(), :bar => { :some => :big, :Hash => 1234 }`.

So I did experiments (linked above) to find a way to avoid the cost of the setup and the call, and eventually settled on `logger.debug? && logger.debug(...)` for this. In C, I'd use a macro for this, but Ruby has no macros :(

jakelandis:
@jordansissel @original-brownbear I am not familiar with this section of code, but the API should be setting the log level on the underlying Java logging context.  How does this prevent that ?

original-brownbear:
@jordansissel does this even work dynamically?

```rb
@logger.debug?
```

The logger is just statically created it seems?

original-brownbear:
@jakelandis beat me to it :D

jordansissel:
@suyograo I responded mainly for two reasons:

1) This broke a (minor) feature, but I am not super worried about it, but it wasn't mentioned or acknowledged.
2) We cannot blanked replace `logger.debug? &&` with another construct unless it has the same behavior.

original-brownbear:
I mean the logger is just created here and isn't using any settings is it:

```rb
module LogStash module Util
  module Loggable
    def self.included(klass)

      def klass.log4j_name
        ruby_name = self.name || self.class.name || self.class.to_s
        ruby_name.gsub('::', '.').downcase
      end

      def klass.logger
        @logger ||= LogStash::Logging::Logger.new(log4j_name)
      end

      def klass.slow_logger(warn_threshold, info_threshold, debug_threshold, trace_threshold)
        @slow_logger ||= LogStash::Logging::SlowLogger.new(log4j_name, warn_threshold, info_threshold, debug_threshold, trace_threshold)
      end

      def logger
        self.class.logger
      end

      def slow_logger(warn_threshold, info_threshold, debug_threshold, trace_threshold)
        self.class.slow_logger(warn_threshold, info_threshold, debug_threshold, trace_threshold)
      end
    end
  end
end; end
```

jordansissel:
@jakelandis @original-brownbear This code, the way I read it, choose how to compile the pipeline based on whether debug is enabled or not.

This means that *after* the pipeline is compiled, using the logging API to change the log level will still mean that these 'logger.debug' statements were never compiled into the pipeline.

jordansissel:
@jakelandis, @original-brownbear, and I met on Zoom to discuss options:

1) Option: Revert this change.
2) Option: Change the log level to 'trace' instead of 'debug', at which point I am more willing to accept the compile-time choice of logging-or-not (because 'trace' is intended, imo, for developers, not for users).
3) Option: Leave this PR, but document for users about this confusing scenario where these specific log entries cannot be tuned just by using the [logging api ](https://www.elastic.co/guide/en/logstash/current/logging.html) (it would require the pipeline be recompiled)

original-brownbear:
revert incoming here #7943

PR #7930, opened 2017-08-07
Use LIR to generate vertex plugin IDs

andrewvc:
In this patch we unify the IDs reported by LIR with those generated using config_ast.rb

This is a temporary fix until LIR execution is built. It relies on the fact that currently both LIR and config_ast.rb only operate on a single concatenated string of configurations. In the future LIR will compile different files separately, then merge their IRs. For now, however, this solution will hold.

This change also exposes the :id attribute of FilterDelegator to bring it to parity with OutputDelegator and InputDelegator which is required for testing purposes.

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -107,16 +107,23 @@ module LogStash; class BasePipeline
     LogStash::Compiler.compile_sources(sources_with_metadata, @settings)
   end
 
-  def plugin(plugin_type, name, *args)
+  def plugin(plugin_type, name, line, column, *args)
     @plugin_counter += 1
 
     # Collapse the array of arguments into a single merged hash
     args = args.reduce({}, &:merge)
 
-    id = if args["id"].nil? || args["id"].empty?
-      args["id"] = "#{@config_hash}-#{@plugin_counter}"
-    else
-      args["id"]
+    # Pull the ID from LIR to keep IDs consistent between the two representations
+    id = lir.graph.vertices.filter do |v| 
+      v.source_with_metadata && 
+      v.source_with_metadata.line == line && 
+      v.source_with_metadata.column == column
+    end.findFirst.get.id
+
+    args["id"] = id # some code pulls the id out of the args
+
+    if !id
+      raise ConfigurationError, "Could not determine ID for #{plugin_type}/#{plugin_name}"
     end
 
     raise ConfigurationError, "Two plugins have the id '#{id}', please fix this conflict" if @plugins_by_id[id]
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -113,12 +113,16 @@ module LogStash; class BasePipeline
     # Collapse the array of arguments into a single merged hash
     args = args.reduce({}, &:merge)
 
-    # Pull the ID from LIR to keep IDs consistent between the two representations
-    id = lir.graph.vertices.filter do |v| 
-      v.source_with_metadata && 
-      v.source_with_metadata.line == line && 
-      v.source_with_metadata.column == column
-    end.findFirst.get.id
+    if plugin_type == "codec"
+      id = SecureRandom.uuid # codecs don't really use their IDs for metrics, so we can use anything here
+    else
+      # Pull the ID from LIR to keep IDs consistent between the two representations
+      id = lir.graph.vertices.filter do |v| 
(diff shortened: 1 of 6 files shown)


logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -365,6 +365,31 @@ class TestPipeline < LogStash::Pipeline
...
+      allow(LogStash::Plugin).to receive(:lookup).with("filter", "dummyfilter").and_return(DummyFilter)
+      allow(LogStash::Plugin).to receive(:lookup).with("output", "dummyoutput").and_return(::LogStash::Outputs::DummyOutput)
+    end
+
+    let(:config) { "input { dummyinput {} } filter { dummyfilter {} } output { dummyoutput {} }"}
+    let(:pipeline) { mock_pipeline_from_string(config) }
+
+    after do
+      # If you don't start/stop the pipeline it won't release the queue lock and will
+      # cause the suite to fail :(
```
original-brownbear: @andrewvc phew this annoys me to no end too btw ... why do we even have a Queue lock here, we're not even running PQ in these specs are we?

andrewvc: You know, I'm not totally sure, let me track that down.

andrewvc: This seems to be the culprit, I'll open another issue on that one: https://github.com/elastic/logstash/blob/master/spec/spec_helper.rb#L64

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -107,16 +107,23 @@ def compile_lir
...
 
     # Collapse the array of arguments into a single merged hash
     args = args.reduce({}, &:merge)
 
-    id = if args["id"].nil? || args["id"].empty?
-      args["id"] = "#{@config_hash}-#{@plugin_counter}"
-    else
-      args["id"]
+    # Pull the ID from LIR to keep IDs consistent between the two representations
+    id = lir.graph.vertices.filter do |v|
```
original-brownbear: @andrewvc I admit this is a bit of a nitpick, but here it goes:

 Can't we generate that `id` in `config_ast` already and just pass it in here. `Pipeline` is already super huge and convoluted. It does seem very wrong to now also add logic to generate a plugin `id` from its position in the config string in here. Seems more reasonable to do this in `config_ast` which already deals with the config source itself doesn't it?

original-brownbear:
@andrewvc looks just fine in principle, just one suggestion on organizing the code a little more readable maybe :)

logstash-core/spec/logstash/pipeline_spec.rb
original-brownbear: @andrewvc urgh yea, let's just turn that into `memory`?, should also make it run faster ...

logstash-core/lib/logstash/pipeline.rb
andrewvc: I suppose we could, but IMHO generated code is harder to maintain than non-generated code, so I'd rather have it in here. 

I agree with you from an architectural perspective, but from a pragmatic code-maintenance perspective I disagree.

WRT the separation of concerns, the pipeline class already has more problems than I can count, I don't think this patch makes it materially worse, and it will go away in short order once LIR-native execution becomes a thing.

original-brownbear: @andrewvc I was actually hoping we could just not generate this but simply add the method that generates this ID to `config_ast.rb` :) but yea, it's doesn't make it much worse. Cool with me :) Feel free to merge once green :)

logstash-core/spec/logstash/pipeline_spec.rb
andrewvc: @original-brownbear I think that this is beyond the scope of this PR, if we make a change here we should make it across all specs. I just amended the commit. It turns out you just need to call `pipeline.close` which is not as bad.

original-brownbear: @andrewvc sure :)

andrewvc:
please test this

suyograo:
Restarted travis build

andrewvc:
@original-brownbear @suyograo FYI I submitted another commit to fix the failing integration tests. In the case where a `codec` was specified with arguments the `plugin` code is called to look it up. In this case we assign it a UUID since we really don't care about codec ids.

original-brownbear:
@andrewvc still LGTM :)

PR #7931, opened 2017-08-07
Shorten Vertex generated IDs

andrewvc:
Currently generated Vertex IDs are unwieldy concatenations of two sha256 hashes. This is just a pain to deal with.

This patch hashes that concatenation instead, yielding something far easier to deal with.


original-brownbear:
@andrewvc LGTM :)

PR #7932, opened 2017-08-07
Bump to 5.5.2 and update lock file

### LINKED ISSUE #7726: Security issue in the Debian package init script
- Version: 1:5.5.0-1
- Operating System: Debian Jessie
- Steps to Reproduce:

1. Install logstash
2. Start it
3. Have your /var/log directory borked

The init script supplied by the Debian package contains these lines as part of the startup process (58-61): 

    # Ensure the log directory is setup correctly.
    [ ! -d "/var/log" ] && mkdir "/var/log"
    chown "$user":"$group" "/var/log"
    chmod 755 "/var/log"

This:

- breaks /var/log permissions by assigning uid/gid logstash:logstash
- does not make any sense at all, given that the logstash binary is started in a chroot, but the logfiles  /var/log/logstash-stdout.log and /var/log/logstash-stderr.log belong to root because they're created outside the chroot in line 73:

    $ ls -lah /var/log|grep logst
    drwxr-xr-x 15 logstash      logstash      4.0K Jul 18 06:25 .
    drwxrwxr-x  2 logstash      root          4.0K Jul 18 18:01 logstash
    -rw-r--r--  1 root          root           240 Jul 18 18:01 logstash-stderr.log
    -rw-r--r--  1 root          root           182 Jul 18 18:01 logstash-stdout.log

Also, I would classify this as a *security issue* because the /var/log directory becomes writable by logstash user.

suyograo:
There were a lot of plugins which got their major versions bumped to handle
6.0 deprecations and obsolete changes. I've added constraints here to not include
them in 5.5.2

Fixes #7726


Gemfile.jruby-1.9.lock.release
```diff
@@ -325,7 +325,7 @@ GEM
       logstash-codec-plain
       logstash-core-plugin-api (>= 1.60, <= 2.99)
       stud (>= 0.0.22, < 0.1.0)
-    logstash-input-log4j (3.0.6-java)
+    logstash-input-log4j (3.1.0-java)
```
jordansissel: minor bump; expected?

Gemfile.jruby-1.9.lock.release
```diff
@@ -415,12 +415,12 @@ GEM
...
       logstash-input-generator
       logstash-output-file
-    logstash-output-elasticsearch (7.3.7-java)
+    logstash-output-elasticsearch (7.3.8-java)
       cabin (~> 0.6)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
       manticore (>= 0.5.4, < 1.0.0)
       stud (~> 0.0, >= 0.0.17)
-    logstash-output-file (4.0.2)
+    logstash-output-file (4.1.0)
```
jordansissel: minor bump; expected?

Gemfile.jruby-1.9.lock.release
suyograo: yep, we've been including minor releases in LS patch releases in 5.x

suyograo:
@jordansissel BTW, the minor bumps in plugins are intentional and we've been shipping this way for all of 5.x

jordansissel:
LGTM

PR #7934, opened 2017-08-08
[Docs] Add docs about pipeline management

dedemorton:
This needs to be merged at the same time as https://github.com/elastic/x-pack-logstash/pull/188


docs/static/config-management.asciidoc
```diff
@@ -0,0 +1,73 @@
...
+
+The pipeline management feature in {xpack} centralizes the creation and
+management of Logstash configuration pipelines. From within the pipeline
+management UI, you can add, edit, and delete pipeline configurations.
+
+The pipeline configurations are stored in the .`logstash` index in Elasticsearch.
+Any changes that you make to a pipeline definition in the UI are picked up and
+loaded automatically by all Logstash instances registered to use the pipeline.
+The changes are applied immediately; you do not have to restart Logstash to pick
+up the changes, as long as Logtash is already registered to use the pipeline.
```
suyograo: Can we add:

If you delete a pipeline (for example, `apache`) from the UI, Logstash will attempt to stop the pipeline if it is running currently. It will wait until all events have been fully processed by the pipeline. Before deleting the pipeline, please make sure you understand your data sources as stopping a pipeline may lead to data losses. <link to shutting down page>

suyograo: To add a new pipeline that is created in the UI, you do not have to restart Logstash. Logstash will watch for changes in the yml file and reload the new changes for you.

suyograo: Also: 

Once you use centralized config management feature, you will not be able to specify local pipeline configuration. This means `pipelines.yml` and settings like `path.config` `config.string` will be inactive when this feature is used.

docs/static/config-management.asciidoc
```diff
@@ -0,0 +1,73 @@
...
+following fields:
+
+[horizontal]
+Pipeline ID::
+A name that uniquely identifies the pipeline. You use this ID when you set up
+{logstash-ref}pipeline-management-settings[configuration management] and specify
+a list of pipeline IDs in the `xpack.management.pipeline.id` setting.
+
+Version::
+Metadata for tracking the version of the pipeline configuration. This information
```
suyograo: Can we add that this is a string? We can provide an example such as v1.0.0

docs/static/config-management.asciidoc
```diff
@@ -0,0 +1,73 @@
...
+is for your use. Logtash doesn't currently manage or validate the version information.
+
+Description::
+A description of the pipeline configuration. This information is for your use.
+
+Pipeline::
+The pipeline configuration.
```
suyograo: Can we add: Just like editing or creating a pipeline locally using your favorite editor, you do not have to worry about whitespaces or indentation. You treat the UI as an editor you've used before to manage LS pipelines.

docs/static/config-management.asciidoc
```diff
@@ -0,0 +1,73 @@
...
+
+image::static/images/new_pipeline.png[]
+
+When you click *Save*, the pipeline runs on all Logstash instances that are
+registered to use the pipeline.
```
suyograo: We can say:

There is no validation done at the UI level. UI will save the new configuration, and LS will attemp to load this new changes. If there are any configuration errors, please check the local Logstash logs. You could also navigate to X-Pack's monitoring feature to check the status of your Logstash nodes.

docs/static/config-management.asciidoc
```diff
@@ -0,0 +1,73 @@
...
+When you click *Save*, the pipeline runs on all Logstash instances that are
+registered to use the pipeline. 
+
+You can specify multiple pipeline configurations that run in parallel on the
+same Logstash node.
+
+If you edit a pipeline configuration and save the changes, Logstash reloads
+the configuration in the background and continues processing events.
+
+If you delete a pipeline, it's deleted permanently.
```
suyograo: See note about pipeline being stopped when deleted.

docs/static/config-management.asciidoc
```diff
@@ -0,0 +1,73 @@
...
+== Configuration Management
+
+Logstash provides configuration management features to make it easier for you to
+manage updates to your configuration over time.
+
+[role="xpack"]
+[[logstash-centralized-pipeline-management]]
+=== Centralized Pipeline Management
+
+//REVIEWERS: I'm not sure if centralized pipeline management is the right term to use here, but if we want to have a generic container for lots of types of configuration management, we need to distinguish pipeline management.
```
suyograo: Actually lets leave it as pipeline management since we use the same in the UI.

dedemorton: Yah, I was thinking that we'll probably want a different topic for each type of config management.

docs/static/config-management.asciidoc
dedemorton: @suyograo Not following this statement completely: "To add a new pipeline that is created in the UI, you do not have to restart Logstash. Logstash will watch for changes in the yml file and reload the new changes for you." Are you talking about the logstash.yml file? So if users add a pipeline ID to the list in `xpack.management.pipeline.id:`, Logstash will reload the file and use the new pipeline?

docs/static/config-management.asciidoc
dedemorton: Not sure I followed this completely. I'll push my edit soon and you can see if I've gotten it right.

docs/static/config-management.asciidoc
dedemorton: FYI...added the bit about deleting the pipeline, but moved it to the section where I talk about deleting pipelines. I think the detail is more useful there.

suyograo: > Are you talking about the logstash.yml file? So if users add a pipeline ID to the list in xpack.management.pipeline.id:, Logstash will reload the file and use the new pipeline? (Asking because this contradicted something I saw elsewhere.)

Yes, in the original design, users can add pipelines created via UI to `logstash.yml` and they'd get picked up automatically. I need to test this, so let's leave it as is for now and we can amend it if it actually works this way :)

PR #7937, opened 2017-08-08
PERFORMANCE: Remove TimestampBiValue

original-brownbear:
Removing another `BiValue` type here for #7712, putting this to the general review Queue since it's kind of trivial to understand now with #7854 and #7917 out there as blueprints for this.

So what I did was:

* Remove `TimestampBiValue` and replace all its uses by `JrubyTimestampExtLibrary.RubyTimestamp` as this thing holds a reference to the Java `Timestamp`
* Cleanup the ever growing number of pass-through conditions into single branches
* Add force loading of the "LogStash" JRuby library to the `EventTest` since the `Ruby` `init` randomly didn't init add that module before the first test was executed and caused an NPE.

Note that this one did not require a special serialization handler because `Timestamp` has a clean serialization handler and is the only field in `JrubyTimestampExtLibrary.RubyTimestamp`. This is tested in `EventTest`:

```java
    @Test
    public void toBinaryRoundtrip() throws Exception {
        Event e = new Event();
        e.setField("foo", 42L);
        e.setField("bar", 42);
        Map<String, Object> inner = new HashMap<>(2);
        inner.put("innerFoo", 42L);
        inner.put("innerQuux", 42.42);
        e.setField("baz", inner);
        e.setField("[@metadata][foo]", 42L);
        assertEquals(42L, er.getField("foo"));
        assertEquals(42, er.getField("bar"));
        assertEquals(42L, er.getField("[baz][innerFoo]"));
        assertEquals(42.42, er.getField("[baz][innerQuux]"));
        assertEquals(42L, er.getField("[@metadata][foo]"));

        assertEquals(e.getTimestamp().toIso8601(), er.getTimestamp().toIso8601());
    }
```

where you can see that the timestamp serializes and then deserializes just fine in the last line.


original-brownbear:
hold on, some test randomly fails because of the module load order, fixing

logstash-core/src/main/java/org/logstash/Rubyfier.java
```diff
@@ -27,12 +28,19 @@ private Rubyfier() {
...
+            return (IRubyObject) input;
+        }
         if (input instanceof String) return BiValues.RUBY.newString((String) input);
-        if (input instanceof RubyFloat) return (RubyFloat) input;
         if (input instanceof Double || input instanceof Float) {
             return BiValues.RUBY.newFloat(((Number) input).doubleValue());
         }
+        if (input instanceof Timestamp) {
+            return JrubyTimestampExtLibrary.RubyTimestamp.newRubyTimestamp(
+                BiValues.RUBY, (Timestamp) input
```
jakelandis: This may be out of scope ... but since you are removing the BiValue concept for timestamps, should `newRubyTimestamp` still even be aware of `BiValues.RUBY` ?

original-brownbear: @jakelandis they `RUBY` field on that thing is just to cache the global Ruby runtime. We can move it to some other place as soon as `BiValues` are gone I guess. It has nothing to do with "BiValue" as a concept really :)

logstash-core/src/test/java/org/logstash/EventTest.java
```diff
@@ -6,13 +6,24 @@
...
-public class EventTest {
+public final class EventTest {
+
+    /**
+     * Some of these tests require a fully initialized global {@link org.jruby.Ruby} instance
+     * so we force the creation of the "LogStash" module here.
+     */
+    @BeforeClass
+    public static void before() {
+        BiValues.RUBY.getOrCreateModule("LogStash");
```
jakelandis: Is this required as part of this change ?

original-brownbear: @jakelandis yea, there is a really weird issue with the load order, if I don't add that then I get an NPE in the RubyTimestamp class, because the module is not yet loaded.

logstash-core/src/main/java/org/logstash/Rubyfier.java
jakelandis: Wait... where is BiValues.RUBY coming from ?

original-brownbear: @jakelandis its' the global Ruby runtime, initialized like this:

```java
    public static final Ruby RUBY = Ruby.getGlobalRuntime();
```

jakelandis: sorry, I know out of scope, but I would be +1 to moving it before BiValues gone. 

I assume this is an expensive call else you just call it directly ?

original-brownbear: @jakelandis yea it's `synchronized` on a `static` field, we have to cache this. Feel free to suggest an alternative class where it can live :)
I just wanted to avoid a huge non functional change PR from moving it.

logstash-core/src/test/java/org/logstash/EventTest.java
jakelandis: Good to know... I was just mocking anything that required the Ruby Runtime. thanks.

jakelandis:
LGTM

PR #7938, opened 2017-08-08
MINOR: Save a few null Checks in Valuefier + Fix Compiler Warning

original-brownbear:
Trivial stuff top to bottom:

* Make utility class final
* remove `null` check in `convertJavaProxy`, it can never be hit since this is guarded by the `instanceof` check in `org.logstash.Valuefier#convert`
* remove `throws` clause for unchecked `IllegalArgumentException`
* Remove `null` check from `convertNonCollection` since this is only hit by one of the possible call sites (the very bottom of `convert`, all others are guarded by `instanceof` => moved the null check to its only call site)
* Made `convertNonCollection` private since it's only used by this utility class

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Valuefier.java
+++ b/logstash-core/src/main/java/org/logstash/Valuefier.java
@@ -1,6 +1,7 @@
 package org.logstash;
 
 import java.io.Serializable;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import org.joda.time.DateTime;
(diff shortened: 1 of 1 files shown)


logstash-core/src/main/java/org/logstash/Valuefier.java
```diff
@@ -19,22 +20,19 @@
...
-        }
-        Object obj = JavaUtil.unwrapJavaObject(jp);
+    private static Object convertJavaProxy(final JavaProxy jp) {
+        final Object obj = JavaUtil.unwrapJavaObject(jp);
         if (obj instanceof IRubyObject[]) {
             return ConvertedList.newFromRubyArray((IRubyObject[]) obj);
         }
         if (obj instanceof List) {
-            return ConvertedList.newFromList((List<Object>) obj);
+            return ConvertedList.newFromList((Collection<?>) obj);
```
jakelandis: Why cast to `Collection`, when we are explicitly checking for `List` ?

logstash-core/src/main/java/org/logstash/Valuefier.java
```diff
@@ -44,16 +42,16 @@ private static Object convertJavaProxy(JavaProxy jp) {
         }
     }
 
-    public static Object convertNonCollection(Object o) {
+    private static Object convertNonCollection(Object o) {
         try {
-            return o == null ? BiValues.NULL_BI_VALUE : BiValues.newBiValue(o);
+            return BiValues.newBiValue(o);
```
jakelandis: Does `BiValues.newBiValue(o)` perform the null check for you ?

logstash-core/src/main/java/org/logstash/Valuefier.java
original-brownbear: @jakelandis we only need a `Collection` here so no need to cast to `List` and hide possible simplifications from inspections. 
I mean I could also turn the check into `Collection`, but the problem I see with that is the behavior change. Then we could input a `Set` but get back a `List` from `ConvertedList`'s internal logic ... that wouldn't be a passive change but I guess we could add an issue to discuss relaxing this actually :)

logstash-core/src/main/java/org/logstash/Valuefier.java
original-brownbear: @jakelandis yea (but I think we could remove that check as well), the important part in this PR is that we'll never see a `null` in this method and hence that check can be removed from the hot path :)

original-brownbear: I only added this upstream check to improve JIT behavior (with some visible success), but after recent changes, it became more visible that we can save some of these checks like this one.

jakelandis:
LGTM

PR #7942, opened 2017-08-08
Performance: remove volatile from hot code

### LINKED ISSUE #7935: "Lazy" Flag in org.logstash.instrument.metrics.counter.LongCounter Deteriorates Performance
This line (and hence that patch) https://github.com/elastic/logstash/pull/7872/files#diff-c08ac103577f611066ab7e0876756088R72

```diff
         longAdder.add(by);
+        dirty = true;
      }	
```

has a very detrimental impact on the metrics overhead since `dirty` is a `volatile`.
Adding a hot write to a `volatile` is kind of a nasty anti-pattern and we should fix this some other way.
On trivial approach that comes to mind is simply making `dirty` `AtomicLong` and simply setting it to the counter value on `setDirty(false)` and setting it to `-1` on `setDirty(true)`.
Then `getDirty` simply becomes `return longAdder.longValue() != dirty.get()` and we get the same behavior without messing with with the performance of `public void increment(long by)` in any way whatsoever? :)

In numbers, this is a `~10%` hit on `baseline` benchmark run.

jakelandis:
Fixes #7935


original-brownbear:
LGTM :)

Issue #7926, opened 2017-08-07
Pipeline metrics when removed and reloaded don't disappear and when added back show empty


jakelandis:
To reproduce: 
Start Logstash with a custom multi-pipeline config and `-r` to allow pipeline restarts.

 Start with a multi-pipeline config similar to:
 ```
 - pipeline.id: heartbeat-pipeline
  path.config: "/Users/jake/workspace/config/heartbeat.conf"
  pipeline.workers: 2
- pipeline.id: tcp-pipeline
  path.config: "/Users/jake/workspace/config/tcp.conf"
  pipeline.workers: 2
```

Remove heartbeat-pipeline, such that pipelines.yml looks like
```
- pipeline.id: tcp-pipeline
  path.config: "/Users/jake/workspace/config/tcp.conf"
  pipeline.workers: 2
```

Allow the pipeline to reload.

`curl -XGET 'localhost:9600/_node/stats/pipelines?pretty'`

Note the heartbeat-plugin is still visible. (part 1 of the bug)

```
 "heartbeat-pipeline" : {
      "events" : null,
      "plugins" : {
        "inputs" : [ ],
        "filters" : [ ],
        "outputs" : [ ]
      },
      "reloads" : {
        "last_error" : null,
        "successes" : 0,
        "last_success_timestamp" : null,
        "last_failure_timestamp" : null,
        "failures" : 0
      },
      "queue" : {
        "type" : "memory"
      }
    }
```

Add the hearbeat-pipeline back. Now the inputs is no longer gets updated (part 2 of the bug, may be same root cause as https://github.com/elastic/logstash/issues/7925)

```
    "plugins" : {
        "inputs" : [ ],
        "filters" : [ ],
        "outputs" : [ ]
      },
```

jakelandis:
Confirmed part 1 of this bug (still showing pipeline) is present on 6.0.0-alpha2 (db9ab809a7ceaa0929a4fa50034ce013d699bfc4), but part 2 is not an issue.

jakelandis:
Confirmed that part 2 is indeed the same root cause as #7925 (and same fix applies)

inqueue:
Bump. I received a report about this from a user on 7.11 looking to use pipeline node stats to determine if a pipeline is running.

PR #7944, opened 2017-08-08
delete all pipeline metrics on pipeline removal

jsvd:
this PR modifies the Stop action to distinguish between a reload and a complete stop.
In the former case, we can selectively clear only the metrics for event counts and plugins. In the latter case, we remove all metrics for that pipeline id.

I also added more tests for the global vs pipeline level metrics.

fixes https://github.com/elastic/logstash/issues/7926


jakelandis:
LGTM

jsvd:
the specs are failing because now due to mocking of agent in a few tests and a new unexpected call being made from outside. fixing..

PR #7947, opened 2017-08-08
Metrics: Witness data model

jakelandis:
Note to reviewer(s): I pulled the serialization code out of this commit to try to keep the review smaller. There may be a few very minor things still in here that is only used by the serialization logic (which is not part of this PR). 

The code here is basically just setters and getters. Not much (if any) business logic. 

-----------

The role of Witness is to provide an API record state of what is happening inside of Logstash, and provide means to serialize the data for consumption (not included in the commit). The Witness is implemented by chaining methods with a singleton root. For example:

`Witness.instance().pipeline("main").inputs("foo").events().in(1)`

Can be used to express that for the foo input in the main pipeline, 1 new event has been seen.

Witnesses may also have a snitch method, which allows the witness to snitch (tell others) the discrete values of the metrics. A snitch method is basically the getter. For example:

`Witness.instance().pipeline("main").inputs("foo").events().snitch().in()`

Tells you you how many events have been witnessed for the foo input in the main pipeline. In practice this the snitch is really only used for testing, but may be useful as just a standard set/get type operations. Serialization to JSON (not included in this commit) is the primary means to get data out of a Witness.

Some workflows require that a Witness forgets what it saw, for example during a pipeline reload. In these cases the Witness may completely or partially forget the data they have seen. This is implemented with a forget method. For example:

`Witness.instance().pipeline("main").inputs("foo").events().forget().all();`

Part of #7788

### DIFF
--- a/logstash-core/src/main/java/org/logstash/instrument/witness/ErrorWitness.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/witness/ErrorWitness.java
@@ -1,0 +1,100 @@
+package org.logstash.instrument.witness;
+
+import org.logstash.instrument.metrics.gauge.TextGauge;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PrintStream;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Witness for errors.
+ */
+public class ErrorWitness {
+
+    private final TextGauge message;
+    private final TextGauge backtrace;
+    private final Snitch snitch;
+
+    public ErrorWitness() {
+        message = new TextGauge("message");
+        backtrace = new TextGauge("backtrace");
+        snitch = new Snitch(this);
+    }
+
+    /**
+     * Stacktrace as a {@link String}
+     *
+     * @param stackTrace The stack trace already formatted for output.
+     */
+    public void backtrace(String stackTrace) {
+        this.backtrace.set(stackTrace);
+    }
+
+    /**
+     * The message of the error.
+     *
+     * @param message human readable error message.
+     */
+    public void message(String message) {
+        this.message.set(message);
+    }
+
+    /**
+     * Get a reference to associated snitch to get discrete metric values.
+     *
+     * @return the associate {@link Snitch}
+     */
+    public Snitch snitch() {
+        return this.snitch;
+    }
+
+    /**
+     * Stacktrace for Java.
+     *
+     * @param throwable The Java {@link Throwable} that contains the stacktrace to output
+     */
+    public void backtrace(Throwable throwable) {
+
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+             PrintStream printStream = new PrintStream(byteArrayOutputStream)) {
+
+            throwable.printStackTrace(printStream);
+            String backtrace = byteArrayOutputStream.toString("UTF-8");
+            this.backtrace.set(backtrace);
+
+        } catch (IOException e) {
+            //A checked exception due to a the close on a ByteArrayOutputStream is simply annoying since it is an empty method.  This will never be called.
+            throw new IllegalStateException("Unknown error", e);
+        }
+    }
+
+    /**
+     * The snitch for the errors. Used to retrieve discrete metric values.
+     */
+    public static class Snitch {
+        private final ErrorWitness witness;
+
+        Snitch(ErrorWitness witness) {
+            this.witness = witness;
+        }
+
+        /**
+         * Gets the error message
+         *
+         * @return the error message
+         */
+        public String message() {
+            return witness.message.getValue();
+        }
+
+        /**
+         * Gets the error stack/back trace
+         *
+         * @return the backtrace as a String
+         */
+        public String backtrace() {
+            return witness.backtrace.getValue();
+        }
+    }
+}
--- a/logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
@@ -1,0 +1,184 @@
+package org.logstash.instrument.witness;
+
+import org.logstash.instrument.metrics.counter.LongCounter;
+
+/**
+ * Witness for events.
+ */
+final public class EventsWitness{
+
+    private LongCounter filtered;
+    private LongCounter out;
+    private LongCounter in;
+    private LongCounter duration;
+    private LongCounter queuePushDuration;
+    private final Snitch snitch;
+    private boolean dirty; //here for passivity with legacy Ruby implementation
+
+    /**
+     * Constructor.
+     */
+    public EventsWitness() {
+        filtered = new LongCounter("filtered");
+        out = new LongCounter("out");
+        in = new LongCounter("in");
+        duration = new LongCounter("duration_in_millis");
+        queuePushDuration = new LongCounter("queue_push_duration_in_millis");
+        snitch = new Snitch(this);
+        dirty = false;
+    }
+
+    /**
+     * Add to the existing duration
+     *
+     * @param durationToAdd the amount to add to the existing duration.
+     */
+    public void duration(long durationToAdd) {
+        duration.increment(durationToAdd);
+        dirty = true;
+    }
+
+    /**
+     * increment the filtered count by 1
+     */
+    public void filtered() {
+        filtered.increment();
+        dirty = true;
+    }
+
+    /**
+     * increment the filtered count
+     *
+     * @param count the count to increment by
+     */
+    public void filtered(long count) {
+        filtered.increment(count);
+        dirty = true;
+    }
+
+    /**
+     * Forgets all information related to this witness.
+     */
+    public void forgetAll() {
+        filtered.reset();
+        out.reset();
+        in.reset();
+        duration.reset();
+        queuePushDuration.reset();
+        dirty = false;
+    }
+
+
+    /**
+     * increment the in count by 1
+     */
+    public void in() {
+        in.increment();
+        dirty = true;
+    }
+
+    /**
+     * increment the in count
+     *
+     * @param count the number to increment by
+     */
+    public void in(long count) {
+        in.increment(count);
+        dirty = true;
+    }
+
+    /**
+     * increment the out count by 1
+     */
+    public void out() {
+        out.increment();
+        dirty = true;
+    }
+
+    /**
+     * increment the count
+     *
+     * @param count the number by which to increment by
+     */
+    public void out(long count) {
+        out.increment(count);
+        dirty = true;
+    }
+
+    /**
+     * Get a reference to associated snitch to get discrete metric values.
+     *
+     * @return the associate {@link Snitch}
+     */
+    public Snitch snitch() {
+        return snitch;
+    }
+
+    /**
+     * Add to the existing queue push duration
+     *
+     * @param durationToAdd the duration to add
+     */
+    public void queuePushDuration(long durationToAdd) {
+        queuePushDuration.increment(durationToAdd);
+        dirty = true;
+    }
+
+    /**
+     * The snitch for the {@link EventsWitness}. Allows to read discrete metrics values.
+     */
+    public static class Snitch {
+
+        private final EventsWitness witness;
+
+        Snitch(EventsWitness witness) {
+            this.witness = witness;
+        }
+
+        /**
+         * Gets the duration of the events.
+         *
+         * @return the events duration.
+         */
+        public long duration() {
+            return witness.duration.getValue();
+        }
+
+        /**
+         * Gets the filtered events count.
+         *
+         * @return the count of the filtered events.
+         */
+        public long filtered() {
+            return witness.filtered.getValue();
+
+        }
+
+        /**
+         * Gets the in events count.
+         *
+         * @return the count of the events in.
+         */
+        public long in() {
+            return witness.in.getValue();
+        }
+
+        /**
+         * Gets the out events count.
+         *
+         * @return the count of the events out.
+         */
+        public long out() {
+            return witness.out.getValue();
+        }
+
+        /**
+         * Gets the duration of the queue push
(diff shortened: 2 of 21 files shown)


jakelandis:
@andrewvc / @original-brownbear - I have requested your review _(not sure if GH sends out notifications for that)_.

original-brownbear:
@jakelandis it does, and sure will take a look very soon :)

logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
```diff
@@ -0,0 +1,205 @@
...
+     */
+    public void queuePushDuration(long durationToAdd) {
+        queuePushDuration.increment(durationToAdd);
+        dirty = true;
+    }
+
+    /**
+     * The snitch for the {@link EventsWitness}. Allows to read discrete metrics values.
+     */
+    public static class Snitch {
```
original-brownbear: @jakelandis lol the naming

suyograo: Lol, this reminds me of Al Pacino's Scent of a Woman scene -- But, not a snitch! https://www.youtube.com/watch?v=PeCTb_gFgfY

andrewvc:
@jakelandis before I look at the actual code in depth some questions:

- Why the `snitch` instead of directly adding getters to the objects?
- Why `forget` as opposed to instantiating new objects with zeroed out values?

jakelandis:
@andrewvc 
* ```snitch``` 
  * To help keep the API clean. Since we don't expose setting values with `setFoo`, it seemed unbalanced to get things with `getFoo`.  The API is attempting be fluent, and traditional getters/setters are usually what a style fluent API is avoiding. 
  * Outside of the tests I can not think of any reason someone would ever need to get the individual metric values. (e.g. I don't expect anyone to really use this)
  * Hiding all the getters behind another method keeps the intellisense  (auto complete) from the IDE clean to help prompt the developer with the expected actionable items.  

* ```forget``` 
  * It allows a single method to be called to zero out a bunch of values at once. This will be called from the Ruby code, so it could be easy to miss zero-ing out some of the values. 
  * imo it reads nicely, such that a witness can forget part or all that it knows
  * Lazy serialization. (same reason for the dirty flag) there is a subtle difference between zero and never set. This helps with flipping things back to the original state as though it never happened. 
  * Need some method on the API to allow setting a counter's value back to zero and/or initial state.

andrewvc:
@jakelandis the snitch thing makes perfect sense, thanks for explaining it so well! I forgot that we don't have setters. We do have other consumers of metrics however besides the web API, such as our own xpack plugin and potentially other open source projects (none of which exist yet admittedly), so we do need a public way to get these values outside of serialization.

I'm OK with the snitch for this, just adding that there are non-test uses here.

WRT forget:
- Wouldn't making a new PipelineWitness also zero out everything instantly? My fear is that from a maintenance we need to ensure that any added property have a corresponding `Forget` line.
- I agree its readable! But, AFAIK the only thing that gets forgotten is the Pipeline (and everything under it), so wouldn't just instantiating a new PipelineWitness be just as readable?
- The distinction between zero and never set is only temporary no? I thought we were removing that once we have parity in specs.
- Why do we need this method to reset a counter's state? Under what situation would something reset a counter's value outside of tests? In the real world it should be impossible to reset a counter's value.

jakelandis:
@andrewvc 
I re-read your initial question w/r/t `forget`, and think I mis-interperted your question. 

Let's try this again :)

>Why forget as opposed to instantiating new objects with zeroed out values?

It actually does this (mostly).  The implementation of the `forget` method in  (plural)`PluginsWitness` removes all of the individual (single) `PluginWitness` from its internal map. The implementation of `forget` in `EventsWitness` results in new LongAdders in the underlying implementation. Those `forget` methods are called from `PipelineWitness#forget()`

After seeing your followup questions, I think we are on the same page....
_Sorry for the noise._

jakelandis:
@andrewvc 
>  just adding that there are non-test uses here.

Is there a need for JSON -> Witness mappings  (de-serialization) ? I had only planned on supporting Witness -> JSON  (serialization).

logstash-core/src/main/java/org/logstash/instrument/witness/ErrorWitness.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param throwable The Java {@link Throwable} that contains the stacktrace to output
+     */
+    public void backtrace(Throwable throwable) {
+
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+             PrintStream printStream = new PrintStream(byteArrayOutputStream)) {
+
+            throwable.printStackTrace(printStream);
+            String backtrace = new String(byteArrayOutputStream.toByteArray(), StandardCharsets.UTF_8);
```
original-brownbear: @jakelandis just a minor point, baos has a `toString()` that you can use here. (doesn't matter that much here, except for shortness of code, but it's also faster :P)

logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
```diff
@@ -0,0 +1,205 @@
...
+        public long queuePushDuration() {
+            return witness.queuePushDuration.getValue();
+        }
+
+    }
+
+    /**
+     * The forgetter for events.
+     */
+    public static class Forgetter {
```
original-brownbear: @jakelandis this class seems fairly redundant, why do we need a special object for the `reset` and not just put expose this functionality on the `EventsWittness`?

jakelandis: Certainly possible, there would need to be `resetAll` and/or `resetPartial` on the main API chain.  Currently there is `forget().all()` and/or `forget().parital()`. There may more flavors in the future. For example if we just wanted to reset duration, it could be added in as `forget().duration()` It helps to hide the reset actions from the main chain of available actions. (A similar argument for `snitch`) I personally think `witness->a->b->forget().all()` reads much nicer then `witness->a->b->resetAll()`

It think comes down alot to style. 

Resetting the underlying metrics as opposed to adding to them feels like a like a different _type_ of action in the chain.  Thinking in terms of intellisense, the options to expose to developers should be the primary actions (recording events), and any other type action should be hidden by a single action type that represents that action (e.g. `forget()` or `snitch`)

FWIW, resets of pipelines is the only consumer here.

andrewvc:
After speaking with @jakelandis the `snitch` objects will be used for in-process consumption of these APIs in xpack and other places.

logstash-core/src/main/java/org/logstash/instrument/witness/ErrorWitness.java
jakelandis: I will update it for the next review. (or this one if there are other requested changes)

logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
```diff
@@ -0,0 +1,99 @@
...
+    private final Map<String, PluginWitness> outputs;
+    private final Map<String, PluginWitness> filters;
+    private final Forgetter forgetter;
+
+    /**
+     * Constructor.
+     */
+    public PluginsWitness() {
+
+        this.inputs = new HashMap<>();
```
andrewvc: We will be accessing these maps across threads right? I'm noticing they aren't CHMs. I think there's a missing piece to the concurrency story here.

I think the best option here would be to have the constructor take the full immutable maps of plugins as arguments, make `forget` in `PipelineWitness` instantiate a brand new `PluginWitness`, and make all references to `PluginsWitness` volatile from `PipelineWitness` to complete the concurrency story.

I wish java actually had a type for `UnmodifiableMap`, but it'd probably be good to pass unmodifiable maps into here to be defensive.

Alternatively, I'm fine just making these CHMs. They aren't expected to be traversed with any frequency.

logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
```diff
@@ -0,0 +1,99 @@
...
+     * Gets or creates the {@link PluginWitness}
+     * @param plugin the map of the plugin type.
+     * @param id the id of the plugin
+     * @return existing or new {@link PluginWitness}
+     */
+    private PluginWitness getPlugin(Map<String, PluginWitness> plugin, String id) {
+        if (plugin.containsKey(id)) {
+            return plugin.get(id);
+        } else {
+            PluginWitness pluginWitness = new PluginWitness(id);
```
andrewvc: See my comment above WRT concurrency issues dynamically modifying these maps.

andrewvc: Also, one of our design goals is moving away from `lazy` metrics appearing. I think it'd be good to ensure all plugins without any laziness if possible, and to enforce that in these metrics classes as much as possible.

logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
andrewvc: I agree that this comes down to style in large part.

One potentially maintenance issue I can see here is that `thing.forget()` is a verb, but it actually just returns the `Forgetter`, I can foresee that being the cause of bugs if the `.all` is left out.

I realize that this is part of the fluid DSL style here, but I think I'd prefer my nouns to be nouns and my verbs to be verbs. I'm fine with or w/o the Forgetters, but I do think that `thing.forgetter.forgetAll()` and `thing.forgetAll()` are more ergonomic in terms of preventing bugs.

logstash-core/src/main/java/org/logstash/instrument/witness/QueueWitness.java
```diff
@@ -0,0 +1,60 @@
...
+    public Snitch snitch() {
+        return snitch;
+    }
+
+    /**
+     * Sets the type of the queue.
+     *
+     * @param type The type of the queue.
+     */
+    public void type(String type) {
```
andrewvc: WDYT of event driven names for things? Rather than `type` which is basically a setter, something like `queueStart(String type)`. The rationale being that queue types only change when a queue is started (and really that can be hoisted up to when a pipeline is started). 

Given that the idea of the `Snitch` is that we don't want to have getters given that we don't have real setters (something I like), I think having setter-like things looks like a smell to me. 

It also makes me think we could use similar methods for pipelines (`pipelineStart`, `pipelineStop`). WDYT?

logstash-core/src/main/java/org/logstash/instrument/witness/PipelinesWitness.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    /**
+     * Get a uniquely named pipeline witness. If one does not exist, it will be created.
+     *
+     * @param name The name of the pipeline.
+     * @return the {@link PipelineWitness} identified by the given name.
+     */
+    public PipelineWitness pipeline(String name) {
+        if (pipelines.containsKey(name)) {
```
jakelandis: Nice catch, you are right... if multiple threads try adding the same named pipeline at the same time, the last one would win wiping out any data from the prior.  

I will ensure it is a ConcurrentHashMap and use the `computeIfAbsent` method.

logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
jakelandis: w/r/t lazy metrics. 
 I am making the assumption that a plugin without an ID is invalid, and we don't want want to construct the `PluginWitness` without an ID. Since we don't know that ID till runtime, it needs be lazy instantiated. 

In this particular case, the pipeline.rb is creating these so based on the upstream business logic, they will never not exist for a plugin.

logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
jakelandis: > We will be accessing these maps across threads right?

Yup, but almost all get. Only a single thread should ever update the map concurrently, but will change to CHM if that ever changes (or I am wrong). 

I will also change the lazy instantiation to use  `computeIfAbsent` to avoid the race condition.


> to have the constructor take the full immutable maps of plugins as arguments

This would require `PipelineWitness` to hold the set of `PluginWitness`. Imo that breaks some encapsulation. (`PluginsWitness` is the best place to house the set of `PluginWitness`).  Also see my comment about wanting to only construct the `PluginWitness` after we know the ID.

logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
jakelandis: > I can foresee that being the cause of bugs if the .all 
ha... i actually made that mistake myself while writing the tests. 

> prefer my nouns to be nouns and my verbs to be verbs
good point. 

I will remove the inner class and replace it with `forgetAll()` and/or `forget[Thing]()`

logstash-core/src/main/java/org/logstash/instrument/witness/QueueWitness.java
jakelandis: +1 to anything that makes it more readable.  

For this iteration I was just matching like for like the existing name space. I will log a new issue to revisit the naming in whole. Mind if we wait till after a few more reviews to start those changes (it makes the mental diffing with namespaces easier) ?

andrewvc: Yep! That's fine with me :)

jakelandis:
@andrewvc / @original-brownbear 
Review changes here: https://github.com/elastic/logstash/pull/7947/commits/d6358a589f4a7561bd5dfb00d63b1da09406873c

(will fixup before merging to master)
(still working on getting that issue logged to revisit the names)

jakelandis:
Issue to improve fluent-ness of API: https://github.com/elastic/logstash/issues/7973

PR #7948, opened 2017-08-08
DLQ: Multiple bug fixes related to event seeks

### LINKED ISSUE #7855: DLQ: seekToNextEvent with Timestamp can throw NPE
If the event that is being sought straddles two blocks, the seekToNextEvent call will fail with the following NullPointerException:

```
java.lang.NullPointerException
	at org.logstash.DLQEntry.deserialize(DLQEntry.java:69)
	at org.logstash.common.io.DeadLetterQueueReader.lambda$seekToNextEvent$2(DeadLetterQueueReader.java:68)
	at org.logstash.common.io.RecordIOReader.seekToNextEventPosition(RecordIOReader.java:86)
	at org.logstash.common.io.DeadLetterQueueReader.seekToNextEvent(DeadLetterQueueReader.java:66)
	at org.logstash.common.io.DeadLetterQueueReaderTest.testStraddledSeek(DeadLetterQueueReaderTest.java:102)
.
[Truncated for brevity]
.
.
``` 

Code to reproduce issue: 

```
    @Test
    public void testStraddledSeek() throws Exception {
        long now = System.currentTimeMillis();
        Timestamp seekTarget = new Timestamp(now);
        try(DeadLetterQueueWriter writeManager = new DeadLetterQueueWriter(dir, 10_000_000, 100_000_000)) {
            char[] tooBig = new char[(BLOCK_SIZE / 2) - 1];
            Arrays.fill(tooBig, 'c');

            Event event = new Event();
            event.setField("message", new String(tooBig));
            DLQEntry entry = new DLQEntry(event, "foo", "bar", String.valueOf(1), new Timestamp(now));
            writeManager.writeEntry(entry);
            entry = new DLQEntry(event, "foo", "bar", String.valueOf(1), new Timestamp(now + 100));
            writeManager.writeEntry(entry);
        }
        try(DeadLetterQueueReader readManager = new DeadLetterQueueReader(dir)) {
            readManager.seekToNextEvent(new Timestamp(now + 100));
            DLQEntry readEntry = readManager.pollEntry(100);
            assertThat(readEntry.getReason(), equalTo("1"));
            assertThat(readEntry.getEntryTime().toIso8601(), equalTo(seekTarget.toIso8601()));
        }
    }
```

robbavey:
The fixes in this commit combine to fix #7855

Fixes 3 bugs:
  seekToNextEventPosition:
    handles the case where a null event is encountered during the search for a matching event
    handle restoration of buffer position when the next event consumed goes across block
     boundaries

  seekToStartOfEventInBlock:
    handles the case where the only event in a block is an 'end' event

Also:
 Adds new RecordIOReaderTest, and moves tests over from RecordIOWriterTest that seem to fit
that better

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReader.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReader.java
@@ -65,9 +65,6 @@ public final class DeadLetterQueueReader implements Closeable {
             currentReader = new RecordIOReader(segment);
             byte[] event = currentReader.seekToNextEventPosition(timestamp, (b) -> {
                 try {
-                    if (b == null){
-                        return null;
-                    }
                     return DLQEntry.deserialize(b).getEntryTime();
                 } catch (IOException e) {
                     return null;
--- a/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReader.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReader.java
@@ -65,6 +65,9 @@ public final class DeadLetterQueueReader implements Closeable {
             currentReader = new RecordIOReader(segment);
             byte[] event = currentReader.seekToNextEventPosition(timestamp, (b) -> {
                 try {
+                    if (b == null){
+                        return null;
+                    }
                     return DLQEntry.deserialize(b).getEntryTime();
                 } catch (IOException e) {
                     return null;
--- a/logstash-core/src/main/java/org/logstash/common/io/RecordIOReader.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/RecordIOReader.java
@@ -129,7 +129,7 @@ public final class RecordIOReader implements Closeable {
         return channelPosition;
     }
 
-   void consumeBlock(boolean rewind) throws IOException {
+    void consumeBlock(boolean rewind) throws IOException {
         if (rewind) {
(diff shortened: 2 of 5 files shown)


logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReader.java
```diff
@@ -65,6 +65,9 @@ public void seekToNextEvent(Timestamp timestamp) throws IOException {
             currentReader = new RecordIOReader(segment);
             byte[] event = currentReader.seekToNextEventPosition(timestamp, (b) -> {
                 try {
+                    if (b == null){
```
talevy: since you added the null->break check for `readEvent` in the RecordIOReader, it can now be assumed that this keyExtractor will only be called if `b != null`. so this check is no longer needed, right?

robbavey: Yeah, I think you're right - this was from an early iteration where I didn't handle the null in RecordIOReader. I can remove it

talevy:
LGTM (tests seem to fail for unrelated reasons)

robbavey:
@suyograo Does this want to be backported all the way to 5.5?

talevy:
@robbavey the answer is *always* for bug-fixes. We still maintain it. Hopefully it makes it into `5.5.2`, but if warranted, we'd release `5.5.3` in the future, which should contain this

PR #7950, opened 2017-08-08
Java Filter Execution

original-brownbear:
### Status

- [x] Removed duplicate grammar
- [x] All tests (including RATs) are green
- [x] All types of conditionals are implemented
- [x] All types of conditionals are implemented
- [x] All possible permutations of input types are implemented
- [x] All possible permutations of input types have tests
- [x] If statements are evaluated twice in some cases and need smarter caching
- [x] Ordinal comparisonons are not in line with Ruby in all cases
- [x] Some equalities are still based on string comparision where inappropriate 
- [x] Output Conditionals are enabled
- [x] Metrics counters for output, input and filter are not correctly separated (specs still pass though ...)
- [x] we have a lot fewer checks for cancelled events now, there must be some cases missing handling (seems we have a lot of specs for this and they still pass)
- [x] Don't call shutdown only flushers periodically
- [x] Joni for regexes, `java.util.regex` may not be compatible so we gotta use that instead
- [x] Add `nand` and `xor` operators

### Performance

- [x] GC overhead is lower in every case compared to the old implementation
- [x] Conditionals evaluate a lot faster
- [x] The case of no filters or conditionals is about 5% slower than before (I don't know why yet, this needs to be investigated/fixed before merging, there must be some redundant action in the execution, that only shows in this case and is outweighed by general improvements for all other cases) ... Fixed, this was a measurement error, we faster in all cases now I think :)

#### Possible Improvements

- Janino (obviously, but also kind of easily doable from this implementation since I kept the translation from `PipelineIR` very explicit, not sure how much we will gain though since thanks to the way `FieldReference` lookups now work, there is very little that doesn't get inlined)
- Less complicated: Smarter chaining of conditionals (currently none of them are getting optimized except for constants, there is a lot of room here by reordering etc.)

#### Benchmarks

Baseline:
```
➜  logstash git:(lir-if-else) ✗ bin/benchmark.sh --local-path=${PWD} -repeat-data=10
Logstash Benchmark
Running Test Case: baseline (x10)
Start Time: Sun 8 13 18:50:23 2017 CEST
Statistical Summary:

Elapsed Time: 51s
Num Events: 10000000
Throughput Min: 0.00
Throughput Max: 315750.00
Throughput Mean: 285714.29
Throughput StdDev: 71738.67
Throughput Variance: 5146436229.27
Mean CPU Usage: 28.34%
➜  logstash git:(lir-if-else) git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'elastic/master'.
➜  logstash git:(master) rake compile:all &> /dev/null
➜  logstash git:(master) bin/benchmark.sh --local-path=${PWD} -repeat-data=10
Logstash Benchmark
Running Test Case: baseline (x10)
Start Time: Sun 8 13 18:51:59 2017 CEST
Statistical Summary:

Elapsed Time: 53s
Num Events: 9798106
Throughput Min: 20033.00
Throughput Max: 301875.00
Throughput Mean: 279945.89
Throughput StdDev: 55609.24
Throughput Variance: 3092387983.10
Mean CPU Usage: 36.91%
```

=> Much lighter on the GC now!

Also kind of cool is the Apache benchmark:

Master:
```sh
➜  logstash git:(master) time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg | pv | wc -c
13.2MiB 0:07:21 [30.5KiB/s] [                                                                                                 <=>                                                                                                          ]
 13801522
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.16s user 2.72s system 0% cpu 7:20.90 total
bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg  1833.99s user 30.28s system 422% cpu 7:21.71 total
pv  3.18s user 20.04s system 5% cpu 7:21.71 total
wc -c  0.41s user 3.21s system 0% cpu 7:21.71 total
```

This:

```sh
➜  logstash git:(lir-if-else) time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg | pv | wc -c
13.2MiB 0:05:14 [42.9KiB/s] [    <=>                                                                                                                                                                                                       ]
 13801710
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.14s user 2.80s system 0% cpu 5:13.88 total
bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg  1315.97s user 31.09s system 428% cpu 5:14.42 total
pv  3.07s user 19.74s system 7% cpu 5:14.42 total
wc -c  0.37s user 2.89s system 1% cpu 5:14.42 total
```

Much better throughput by batching for each filter (hence making good use of the CPU cache) function :)

Note:

* There is a million performance improvements we could add here, but pretty much all of them also involve adding more (or more complicated) code => I tried to be as short as possible here for round 1
* "filtered" and "output" metrics are always the same size in `master` as well (just in case you wonder why both increment calls now happen back to back)
* `filter(event, &block)` on the pipeline should really go away as it way complicates the Java logic, requiring interpreting all leafs as outputs when there are no outputs in the config, and also forcing us to have the debug terminal dataset
* Obviously passing the `flush` "state" up/down the `Dataset` topology is kind of dirty by just using params, but I feld this was the easiest approach to clean up when/if we drop flush + didn't add a performance it to the general case
* This contains the fix to #8110 since I needed typed-deserialization for `RubyString` here to get good performance on `String` comparisons (otherwise I have to convert `String` to `RubyString` after deserializing `String` which is a loss in every scenario ... either way this is a big speedup for serializing `RubyString`)
* Also contains https://github.com/elastic/logstash/pull/8205 as a dependency (still waiting for a review there ;))


original-brownbear:
@andrewvc ping :)

andrewvc:
I've been working through my understanding of this in-progress PR and have tried out some configs on it. Obviously this is very WIP and I know you're not done here, but I think there's something missing from the graph traversal. 

The config at the bottom of this comment doesn't execute the same as master:

```
master tags   ["foo","bar","baz","bot","bonk","wat","final"]
this branch    ["foo","bar","final","baz"]
```

These tags should be identical (including order)

The config is:
```
input {
  generator { count => 5000 tags => foo }
}

filter {
  if ("foo" in [tags]) {
    mutate { id => addbar add_tag => bar }

    if ("bar" in [tags]) {
      mutate { id => addbaz  add_tag => baz }
    }

    if ("baz" in [tags]) {
      mutate { id => addbot add_tag => bot }

      if ("bot" in [tags]) {
        mutate { id => addbonk add_tag => bonk }
      }
    }
  }

  if ("bot" in [tags]) {
    mutate { id => addwat add_tag => wat }
  }

  sleep { time => 2 }

  mutate { id => addfinal add_tag => final }

}

output {
}
```

The IR looks correct as verified in the visualizer:

image

original-brownbear:
@andrewvc :D good job the still failing specs here are:

```sh
rspec ./logstash-core/spec/conditionals_spec.rb:124 # conditionals in filter nested "{"foo":"bar","nest":123}" when processed
rspec ./logstash-core/spec/conditionals_spec.rb:117 # conditionals in filter nested "{"foo":"bar","nest":124}" when processed
rspec ./logstash-core/spec/conditionals_spec.rb:131 # conditionals in filter nested "{"notfoo":"bar","nest":123}" when processed
rspec ./logstash-core/spec/conditionals_spec.rb:138 # conditionals in filter nested "{"bar":"baz","nest":123}" when processed
rspec ./logstash-core/spec/conditionals_spec.rb:427 # conditionals in filter multiple new events from root "{"type":"original"}" when processed
rspec ./logstash-core/spec/conditionals_spec.rb:397 # conditionals in filter new events from root "{"type":"original"}" when processed
```

your example looks like it's covered by them :) ... should be up and running by the time you wake up :)

original-brownbear:
@andrewvc output from your config:

```sh
{"@version":"1","host":"localhost","sequence":4301,"@timestamp":"2017-08-10T17:09:02.745Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4302,"@timestamp":"2017-08-10T17:09:02.745Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4303,"@timestamp":"2017-08-10T17:09:02.745Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4304,"@timestamp":"2017-08-10T17:09:02.745Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4305,"@timestamp":"2017-08-10T17:09:02.745Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4306,"@timestamp":"2017-08-10T17:09:02.745Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4307,"@timestamp":"2017-08-10T17:09:02.745Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4308,"@timestamp":"2017-08-10T17:09:02.745Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4309,"@timestamp":"2017-08-10T17:09:02.745Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4310,"@timestamp":"2017-08-10T17:09:02.746Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4311,"@timestamp":"2017-08-10T17:09:02.746Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4312,"@timestamp":"2017-08-10T17:09:02.746Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4313,"@timestamp":"2017-08-10T17:09:02.746Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4314,"@timestamp":"2017-08-10T17:09:02.746Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4315,"@timestamp":"2017-08-10T17:09:02.746Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4316,"@timestamp":"2017-08-10T17:09:02.746Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4317,"@timestamp":"2017-08-10T17:09:02.746Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
{"@version":"1","host":"localhost","sequence":4318,"@timestamp":"2017-08-10T17:09:02.746Z","message":"Hello world!","tags":["foo","bar","baz","bot","bonk","wat","final"]}
```

happy ? :)

original-brownbear:
@andrewvc update: rebase + merging in the handling of the plugin id like before was easy. The fact that the sleep isn't hit was an outright bug, almost fixed though :) Should be reviewable in your morning tomorrow :)

original-brownbear:
Jenkins test this please

original-brownbear:
@andrewvc btw. pretty much done now. I only have a few open questions. But all pre-existing specs and integration tests pass already, so it's all just details => feel free to give it a look :)

You win btw... Java conditionals in your example config above are `2.5x` as fast as the current Ruby version (didn't expect that + the way the Java comparisons are done is far from optimal at this point, there's lots of room in those admittedly). 
Also in general filter invocation is a ton faster now that we pass actual batches to filters instead of running filters in a block event-by-event (much better CPU cache use if you go filter-by-filter instead of event-by-event).
Also stuff that was impossible to inline like the method lookup here

```rb
  define_singleton_method :filter_func do |event|
    events = [event]
    @logger.debug? && @logger.debug("filter received", "event" => event.to_hash)
              events = @generated_objects[:filter_dlq_commit_2].multi_filter(events)

    events
  end
```

are now getting in-lined fully.

original-brownbear:
Jenkins test this please

original-brownbear:
PQ ITs are randomly broken by #7990 and fixed in #7991

original-brownbear:
Jenkins test this please

original-brownbear:
@andrewvc phew pretty much done I think ... kept the output metric behavior by keeping the batch merge around for the time being.
It's actually not that costly anymore anyway as a result of https://github.com/elastic/logstash/pull/7902 (added a proper hash-code to the `RubyEvent`)

original-brownbear:
@jordansissel @suyograo @andrewvc phew this would be what I'd go with for round 1 of this. Obviously, a million spots left that aren't 100% optimal, but I think we're already in very hard to review territory :( and probably better postpone optimizing them :)

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -0,0 +1,709 @@
...
+            final EventValueExpression right) {
+            return new EventCondition.Compiler.FieldInField(
+                PathCache.cache(left.getFieldName()), PathCache.cache(right.getFieldName())
+            );
+        }
+
+        private static EventCondition eq(final EventValueExpression evale,
+            final ValueExpression vale) {
+            return new EventCondition.Compiler.FieldEquals(
+                evale.getFieldName(), vale.get().toString()
```
andrewvc: This conversion to `String` is not correct. Value expressions can be numeric.  I know this explodes the complexity here but all the types in https://github.com/elastic/logstash/blob/master/logstash-core/src/main/java/org/logstash/config/ir/expression/ValueExpression.java#L19 should be covered.

Later on in FieldEquals, you take a String. But I wonder, why do that when we can rely on the fact that we only deal in boxed types for field values and can use `Object.equals` there after a null check?

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -672,7 +633,7 @@ class TestPipeline < LogStash::Pipeline
         sleep(0.1) until pipeline.ready?
       end
       Stud.try(max_retry.times, [StandardError, RSpec::Expectations::ExpectationNotMetError]) do
-        wait(10).for do
+        wait(1).for do
```
original-brownbear: @andrewvc I think I just had some bug here in my testing and wanted to iterate quicker, reverting :)

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -657,18 +610,25 @@ def shutdown_workers
   def filter(event, &block)
     maybe_setup_out_plugins
     # filter_func returns all filtered events, including cancelled ones
-    filter_func(event).each {|e| block.call(e)}
+    wait_until_started
```
original-brownbear: @andrewvc I would just add a general issue asking to remove this method and rewrite the tests in `conditionals_spec.rb` accordingly. I think a todo of this "magnitude" will just go ignored forever :D?

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -106,7 +106,7 @@ class DummyFlushingFilterPeriodic < DummyFlushingFilter
   def flush(options)
     # Don't generate events on the shutdown flush to make sure we actually test the
     # periodic flush.
-    options[:final] ? [] : [::LogStash::Event.new("message" => "dummy_flush")]
+    return options[:final] ? [] : [::LogStash::Event.new("message" => "dummy_flush")]
```
original-brownbear: Merge mistake :)

logstash-core/spec/conditionals_spec.rb
```diff
@@ -425,17 +440,56 @@ def multi_receive(events)
...
+        }
+      
+        mutate { id => addprev add_tag => prev }
+      
+        mutate { id => addfinal add_tag => final }
+      
+      }
+    CONFIG
+
+    sample("tags" => ["foo"]) do
```
original-brownbear: Order check + additional test case added here :)

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
original-brownbear: @andrewvc 

> But I wonder, why do that when we can rely on the fact that we only deal in boxed types for field values and can use Object.equals there after a null check?

I did that elsewhere even lol, sorry for missing the most obvious spot to fix. Will fix very soon :)

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -247,13 +247,6 @@ class TestPipeline < LogStash::Pipeline
           pipeline = mock_pipeline_from_string(test_config_with_filters)
           pipeline.close
         end
-
-        it "should print the compiled code if config.debug is set to true" do
```
andrewvc: Thinking through this, I think we need to keep this giving that we are targeting the 6.1 release. Let's have it just emit `Graph#toString`

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
andrewvc: @original-brownbear can we add a test for this too? It's worrying that we don't have one for different types of equality.

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -0,0 +1,709 @@
...
+        }
+
+        /**
+         * This {@link IllegalArgumentException} is thrown when the inputs to an {@code if}
+         * condition do not conform to the expected types. It being thrown is a bug in Logstash
+         * in every case.
+         */
+        private static final class UnexpectedTypeException extends IllegalArgumentException {
+
+            private static final long serialVersionUID = -3673305675726770832L;
```
andrewvc: What's with the super-long number? Why not, say, `1L`?

Not a problem, just curious.

original-brownbear:
@andrewvc I added some todos for me to deal with today and tomorrow and added the `Work in Progress` label, don't waste more time on reviewing until those are dealt with maybe since they obv. mean changes incoming :)

logstash-core/build.gradle
```diff
@@ -106,6 +106,7 @@ dependencies {
     compile 'com.fasterxml.jackson.core:jackson-databind:2.7.3'
     compile 'com.fasterxml.jackson.core:jackson-annotations:2.7.3'
     compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:2.7.3'
+    compile group: 'org.codehaus.janino', name: 'janino', version: '3.0.7'
```
andrewvc: This seems to have been left here by mistake :)

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -287,6 +288,12 @@ def cancel(event)
         # @cancelled[event] = true
       end
 
+      def collect
```
jordansissel: Two comments -- in Ruby, `Enumerable#collect` returns an Enumerator, not an array.

If you wish to produce an array with an object implementing `each` you can `include Enumerable`:

```
% irb
>> class X
>>   include Enumerable
>>   def each(&block)
>>     block.call(1)
>>     block.call(2)
>>     block.call(3)
>>   end
>> end
=> nil
>> X.new.to_a
=> [1, 2, 3]
```

In this case, removing this `collect` method and adding `include Enumerable` will give you a `to_a` method to get an array with all elements of the `#each` iteration.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -217,6 +218,12 @@ def cancel(event)
         # @cancelled[event] = true
       end
 
+      def collect
```
jordansissel: Similar comment to wrapped_acked_queue: I recommend removing `collect` method and adding `include Enumerable` which will add both `collect` and `to_a` (and we should use `to_a` if we wish an array of elements)

logstash-core/spec/conditionals_spec.rb
```diff
@@ -398,11 +413,11 @@ def multi_receive(events)
       expect(subject).to be_an(Array)
       expect(subject.length).to eq(2)
 
-      expect(subject[0].get("type")).to eq("original")
-      expect(subject[0].get("cond1")).to eq("true")
-      expect(subject[0].get("cond2")).to eq(nil)
+      expect(subject[1].get("type")).to eq("original")
```
jordansissel: Did the order change? Why the change from 0 to 1?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -106,6 +88,22 @@ def compile_lir
     ]
     LogStash::Compiler.compile_sources(sources_with_metadata, @settings)
   end
+   
+  def build_output(name, line, column, *args)
```
jordansissel: Are these to satisfy the `RubyIntegeration.Pipeline` interface? If so, can we align then names (use camelCase instead of the snake_case provided by JRuby?).

Rationale: grepping fails for 'buildOutput' to see where it is defined, so it was awkward to read the see the Pipeline interface and try to find an implementation.

logstash-core/src/main/java/org/logstash/config/ir/compiler/RubyIntegration.java
```diff
@@ -0,0 +1,98 @@
...
+        void multiReceive(Collection<JrubyEventExtLibrary.RubyEvent> events);
+    }
+
+    /**
+     * The Main Ruby Pipeline Class. Currently, this interface is implemented only by the Ruby class
+     * {@code BasePipeline}.
+     */
+    public interface Pipeline {
+
+        IRubyObject buildInput(RubyString name, RubyInteger line, RubyInteger column,
```
jordansissel: IRubyObject is probably too generic. Any reason not to add an interface for Input similar to Output and Filter?

Additionally, can we move away from JRuby objects (RubyString, etc) in the Java interfaces? The reason I ask is that I am working on the Java Plugin API which will be pure Java (no JRuby) and I'm wondering about the difficulty of moving away from IRubyObject, RubyString, etc, to support the future Java Plugin API.

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,412 @@
...
+        return compilePipeline(true);
+    }
+
+    /**
+     * This method contains the actual compilation of the {@link Dataset} representing the
+     * underlying pipeline from the Queue to the outputs.
+     * @param debug True iff called from {@link CompiledPipeline#buildFilterFuncDebug()} to
+     * terminate on a {@link Dataset.TerminalDataset.TerminalDebugDataset}
+     */
+    private Dataset compilePipeline(final boolean debug) {
```
jordansissel: Feels like this is better off as a factory-style static method.

```java
public static CompiledPipeline compile(PipelineIR graph, boolean debug)
```

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,412 @@
...
+        plugin.register();
+        return plugin;
+    }
+
+    /**
+     * This method contains the actual compilation of the {@link Dataset} representing the
+     * underlying pipeline from the Queue to the outputs.
+     * @return Compiled {@link Dataset} representation of the underlying {@link PipelineIR} topology
+     */
+    public Dataset buildFilterFunc() {
```
jordansissel: the pipeline contains more than just filters (outputs, too!), so calling this "filter" seems like a bad name.

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,412 @@
...
+
+    /**
+     * This method contains the actual compilation of the {@link Dataset} representing the
+     * underlying pipeline from the Queue to the outputs as a Debug pipeline.
+     * This means that it will return all {@link JrubyEventExtLibrary.RubyEvent} that passed through
+     * it from its terminal {@link Dataset} by using
+     * {@link Dataset.TerminalDataset.TerminalDebugDataset}.
+     * @return Compiled {@link Dataset} representation of the underlying {@link PipelineIR} topology
+     */
+    public Dataset buildFilterFuncDebug() {
```
jordansissel: the pipeline contains more than just filters (outputs, too!), so calling this "filter" seems like a bad name.

Also, this (and buildFilterFunc) are basically curried methods that call compilePipeline with either true or false (Depending on the function name). Why should they exist?

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
```diff
@@ -0,0 +1,438 @@
...
+         * Empty {@link Collection} returned by this class's
+         * {@link Dataset#compute(RubyIntegration.Batch, boolean, boolean, RubyHash)} implementation.
+         */
+        private static final Collection<JrubyEventExtLibrary.RubyEvent> EMPTY_RETURN =
+            Collections.emptyList();
+
+        /**
+         * Trivial {@link Dataset} that simply returns an empty collection of elements.
+         */
+        private static final Dataset TRIVIAL = new Dataset() {
```
jordansissel: Is this object just used to avoid null checks on `Dataset.from()` return values? I may be misunderstanding otherwise, please advise on what this is for :)

I am also not sure what the word "trivial" is intended to mean. Is this maybe better named "EMPTY_DATASET" ?

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
```diff
@@ -0,0 +1,438 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import org.jruby.RubyArray;
+import org.jruby.RubyHash;
+import org.logstash.ext.JrubyEventExtLibrary;
+
+/**
+ * A data structure backed by a {@link RubyArray} that is lazily filled with
```
jordansissel: This javadoc doesn't inform me as to the purpose of this class. It helps me understand an implementation detail (lazy, backed by RubyArray), but doesn't tell me what Dataset is used for.

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
```diff
@@ -0,0 +1,438 @@
...
+        /**
+         * <p>{@link Dataset} that contains all {@link JrubyEventExtLibrary.RubyEvent} of all of
+         * its dependencies and is assumed to be at the end of an execution.</p>
+         * This dataset does not require an explicit call to {@link Dataset#clear()} since it will
+         * automatically {@code clear} all of its parents.
+         * Unlike {@link Dataset.TerminalDataset} this class will return all
+         * {@link JrubyEventExtLibrary.RubyEvent} that passed through it on a call to
+         * {@link Dataset#compute(RubyIntegration.Batch, boolean, boolean, RubyHash)}.
+         */
+        public static final class TerminalDebugDataset implements Dataset {
```
jordansissel: this class is `public` but the non-debug `TerminalDataset` is package-private?

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
```diff
@@ -0,0 +1,438 @@
...
+                this.parents = parents;
+            }
+
+            @Override
+            public Collection<JrubyEventExtLibrary.RubyEvent> compute(
+                final RubyIntegration.Batch batch, final boolean flush, final boolean shutdown,
+                final RubyHash options) {
+                final Collection<JrubyEventExtLibrary.RubyEvent> res = new ArrayList<>(10);
+                parents.forEach(
+                    dataset -> res.addAll(dataset.compute(batch, flush, shutdown, options))
```
jordansissel: The code we're adding for debugging seems, at a glance, fairly complex. The difference between TerminalDebugDataset and TerminalDataset seems to be that one returns EMPTY_RETURN and the other returns a collection of all(?) RubyEvents? Why? What's the story for this feature?

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
```diff
@@ -0,0 +1,438 @@
...
+         * {@link Collections#emptyList()}.
+         * @param parents Parent {@link Dataset} to sum and terminate
+         * @return Dataset representing the sum of given parent {@link Dataset}
+         */
+        public static Dataset from(final Collection<Dataset> parents, final boolean debug) {
+            final int count = parents.size();
+            final Dataset result;
+            if (count > 0) {
+                if (debug) {
+                    result = new Dataset.TerminalDataset.TerminalDebugDataset(parents);
```
jordansissel: Sometimes we pass debug as a flag (here), and sometimes the "debug" part is implemented as a class or a function by a special name w/ "Debug" in the name.

Why choose `from(..., debug)` here instead of, similar to elsewhere in this PR, having two methods:

```java
public static Dataset from(final Collection<Dataset> parents) { ... }

public static Dataset fromDebug(final Collection<Dataset> parents) { ... }
```

I'm not strictly in favor of one (method/classname indicating "Debug", vs a boolean parameter), but just highlighting some inconsistency.

logstash-core/lib/logstash/pipeline.rb
jordansissel: This function (filter) seems to have a downstream code creation impact with several `debug` checks (TerminalDebugDataset, CompiledPipeline.buildFilterFuncDebug, etc).

Is it necessary? If it exists only to satisfy parts of the existing test suite, then we're testing a kind of test fixture and skipping testing parts of the actual implementation. Can you help me understand better?

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
jordansissel: (This comment assumes we are keeping the "debug pipeline" construct which seems to exist only for spec suite?)

jordansissel:
While fiddling with this PR and looking at performance numbers, I saw something strange - perhaps a bug in metrics?

Running:

```
% bin/logstash -e 'input { generator {  } } output { stdout { codec => dots } } '  | pv -War > /dev/null
```

* events-per-second according to the API:

```
% while true; do curl -s localhost:9600/_node/stats/pipelines?pretty | jq '.pipelines.main.events | (.out / .duration_in_millis) * 1000'; sleep 1; done
273565.3102720511
276099.79310873285
278061.81054549135
280466.92587482755
282431.35156658286
284426.2238367065
285636.76452673896
287251.8148710899
286638.9231499051
285311.1549165121
286906.3002989193
288764.3961154425
290902.58903335215
292963.2072508637
294649.0872662511
```

the rate according to the API keeps increasing (this is observed well beyond "JVM warmup" period). 

* events/sec according to `pv`: 188Keps (fluctuating between 150Keps and 220Keps)

The fact that the stats api keeps showing an ever increasing value (generally) for me is alarming. I haven't investigated, but my hypothesis is around how we compute duration. We use `System.currentTimeMillis` which would be lossy if we are processing at high rates of speed we could do work faster than the clock is ticking which would cause the clock to appear to stall (not increment) when some has completed.

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
original-brownbear: @jordansissel the `TerminalDataset` is a direct child of the `interface` `Dataset` => it's automatically `public` without the `public` keyword :)

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
original-brownbear: @jordansissel this is inconsistency is kinda by design to be able to remove the whole `Debug` path as soon as the conditionals spec is rewritten to not require the pipeline `filter(event)` method. This was the least amount of code version for that imo (that doesn't sacrifice performance by having some hot `if(debug)` call). Think we can live with this until we find the time to clean up the Ruby `pipeline.rb` (next week I'm back to normal and can attack that problem right away anyways :D).

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
original-brownbear: @jordansissel yea that sounds nicer let me rename :)

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
original-brownbear: @jordansissel right sorry about that weird doc, tough week(s) it seems :D Will add the first half of whatever doc I was planning here :)

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
original-brownbear: as I said below, I'd rather remove the debug pipeline thing, but it's hard to do this in step 1 (I'd rather not refactor tests a lot while I refactor the related functionality in the same PR :))

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
original-brownbear: +1 renaming :)

original-brownbear:
@jordansissel will investigate the metrics thing ... the microsecond resolution thing is weird because this would also show up in `master` wouldn't it? I mean we did get faster, but not an order of magnitude or more :)
Likely just some dumb logical mistake I added somewhere :)

original-brownbear:
Jenkins test this please

andrewvc:
@jordansissel @original-brownbear We should be be using `System.nanoTime()` for these purposes. A good discussion of why: https://stackoverflow.com/questions/351565/system-currenttimemillis-vs-system-nanotime/351571#351571

original-brownbear:
@andrewvc yea nanotime is def `>>>` `currentTimeMillis` we can fix this (separate PR I guess, I made that mistake earlier and not here :() but it doesn't seem to be the root cause of weird metric values here. On it debugging this soon though, fixed almost all other criticism here already too :)

original-brownbear:
@andrewvc @jordansissel ok now that I'm touching this again I'm starting to really hate the way `conditionals_spec` is written and forcing us to keep this weird `filter(event ...)` method and all that comes with it downstream around.
Is it going to be too painful for you guys to review or can I refactor it to use a dummy output in this PR? Would mean a lot less (and simpler) production code but obviously an additional change to the tests, wdyt?

jordansissel:
re: nanotime vs millis -- I am +1 on using nanotime (it's monotonic and higher precision; both properties are nice). I also agree moving to nanotime should be done in a separate PR and not here.

jordansissel:
Are you certain the specs need to be rewritten? The specs use `sample` rspec helper from devutils. We've had side discussions elsewhere about moving devutils back into the main repo; and this could be one such opportunity, even if we only replace/override the `sample` helper. Maybe I'm missing something and it would be much more complex to do this?

original-brownbear:
@jordansissel the annoying thing about these tests is that they grab the events from directly after the execution of all filters but before the outputs. This isn't really a thing with the graph execution now, that's why I have all this weird code around for determining the quasi-terminal "leaf" nodes from `if` statements etc. in `CompiledPipeline`. If I'd adjust this spec to simply pump all events into a `DummyOutput`  I could remove all that weird logic + the single event `filter` method, and everything that has `Debug` in the name in the Java code :)

original-brownbear:
@jordansissel also the explicit `flush` method and some logic related to it in `pipeline.rb` is only around for the purposes of this spec as well :) => could also clean that up and save some more code there.

jordansissel:
> they grab the events from directly after the execution of all filters but before the outputs

Right. What I mean is that we could change the `sample` helper (https://github.com/elastic/logstash-devutils/blob/master/lib/logstash/devutils/rspec/logstash_helpers.rb#L54) to instead of calling `pipeline.filter` use a fixture as an input (a real input plugin that produces exactly the 1 sample event) and has an output that captures the result.

jordansissel:
> If I'd adjust this spec to simply pump all events into a DummyOutput 

Agreed! My suggestion is to modify the `sample` helper in logstash-devutils which should allow `conditionals_spec.rb` to continue passing, without modifications, with an improved `sample` helper.

original-brownbear:
@jordansissel ah sure that would be a neat way of achieving that as well :) (and wouldn't even blow up this PR). I'll add a PR again `dev-utils` for that then :) (tomorrow :(, not much time today ... grml)

original-brownbear:
@jordansissel missing accomplished elsewhere :D back to this :) ... metrics is clearly busted values measured for time spans are simply wrong.

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
original-brownbear: @andrewvc Because I clicked auto generate in idea :D set to `1L` :)

suyograo:
Release labels please.

ph:
I will take a look at this today, amazing work @original-brownbear

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
original-brownbear: @jordansissel I think there was some performance bug with that approach out there. Even without a bug, this will add another frame and slow things down. Better not add more indirection to the hot-path here :)

original-brownbear: Also there's this one https://github.com/jruby/jruby/issues/4108 which makes me less than confident in using `Enumerable` at this point :D

logstash-core/lib/logstash/pipeline.rb
original-brownbear: @jordansissel right makes sense :) => will adjust

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,412 @@
...
+ * This class represents an executable pipeline, compiled from the configured topology that is
+ * learnt from {@link PipelineIR}.
+ * Each compiled pipeline consists in graph of {@link Dataset} that represent either a {@code filter}
+ * or an {@code if} condition.
+ */
+public final class CompiledPipeline {
+
+    /**
+     * Configured inputs.
+     */
```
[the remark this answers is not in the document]
original-brownbear: @jordansissel I'm not touching inputs in any way yet from Java besides setting them up => I didn't add an interface for them.

logstash-core/spec/conditionals_spec.rb
original-brownbear: @jordansissel yea just a result of us not guaranteeing any order within a batch and the graph execution taking advantage of this fact by not using ordered sets.

logstash-core/src/main/java/org/logstash/config/ir/compiler/RubyIntegration.java
original-brownbear: @jordansissel currently I have no use for a Java `Input` here as explained above. Let's not add more code than we have to until we need it? :)

original-brownbear: @jordansissel also if we want a Java typed interface here, we can just add an overload that does the conversion when we need it. But until then I see no reason adding redundant code steps in, right? Also, I'd just naturally remove these once we ported the functions behind these to Java preferably instead of adding a million wrappers that no one understands :)

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
original-brownbear: @jordansissel this is just another outgrowth of what we need for working with dev utils at the moment.
Fixing dev-utils is kinda meh ... I tried but realized I need more functionality in there as well to test the exceptions ... will see what I can do about this today.

In general though ... we should really move them into core imo, a different release cycle for your UT utils is just kinda painful :)?

original-brownbear:
This is a little blocked by https://github.com/elastic/logstash/pull/8107 since I fixed datatypes for numerics today. Just in case someone is bored and wants to review a PR :P

original-brownbear:
Picked https://github.com/elastic/logstash/pull/8107 over here so I can continue on the remaining type issues for the time being.

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -0,0 +1,709 @@
...
+        }
+
+        private static EventCondition lt(final Lt lt) {
+            final EventCondition condition;
+            final Expression uleft = lt.getLeft();
+            final Expression uright = lt.getRight();
+            if (eAndV(lt)) {
+                final EventValueExpression left = (EventValueExpression) uleft;
+                final ValueExpression right = (ValueExpression) uright;
+                condition = not(or(gt(left, right), eq(left, right)));
```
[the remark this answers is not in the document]
original-brownbear: @andrewvc actually this is the only part that has (had) really good coverage.
Admittedly this is not the strongest implementation possible yet though :) But not greater or equal means less I think :D

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
original-brownbear: @andrewvc got tests for numeric and string now :) So just the more fringe types missing :)

original-brownbear:
Long/Fixnum comparisons fixed as well now.

Also, I just realized one thing:
There are a million possible type combinations out there that are probably very rare and not worth optimizing ... we should just ditch `BiValue` completely before finishing up here and use the Ruby types for everything not yet optimized in individual `EventCondition`.
We can always fall back to `org.jruby.RubyBasicObject#compareTo` for all types I potentially forgot to handle and be guaranteed to retain the old behaviour ...

=> I really need a review of #8107 ... grml :) After that one, I can remove the remaining fringe bivalues in one more PR and we're done with that thing and probably also this PR.

original-brownbear:
This is now also blocked by a decision on https://github.com/elastic/logstash/pull/8130 (I need to know if I even have to implement this logic for `TImestamp` or if we won't allow setting `Timestamp` on all fields)

original-brownbear:
@andrewvc @jordansissel update :)

I fixed all the comparison spots and we're now guaranteed to be in line with Ruby for this PR because they look like this now:

```java
        private static final class FieldGreaterThanString implements EventCondition {

            private final FieldReference field;

            private final RubyString value;

            private FieldGreaterThanString(final String field, final String value) {
                this.field = PathCache.cache(field);
                this.value = RubyUtil.RUBY.newString(value);
            }

            @Override
            public boolean fulfilled(final JrubyEventExtLibrary.RubyEvent event) {
                return value.compareTo(
                    (IRubyObject) event.getEvent().getUnconvertedField(field)
                ) < 0;
            }
        }
```

This works with BiValue gone because `getUnconvertedField` will return an `IRubyObject` for all valid comparisons and behave 100% the same way the current implementation does.

**But** ... this requires #8158 to be merged and #8131 to be fixed (have a fix for that already but it needs #8158) as well as `SymbolBiValue` to be removed (also done already but needs #8158 to be merged first :))

For the metrics thing, @jordansissel found I'll open a separate issue, it's not a bug here but just a side-effect of an existing say inconsistency :)

andrewvc:
@original-brownbear that is very nice! So, we still throw an error on type issues there then, great!

What's the error message? I'm thinking that this is a good time to see if we can improve it. If we can print out the line/column + LIR source of where that happened that would be ideal.

original-brownbear:
Only https://github.com/elastic/logstash/pull/8205 left as a dependency here. Then I can add the equals and comparison operator for `BigNum` and `BigDecimal` types (like 5 lines of code each :)) and we're back to the old behavior everywhere.
Metrics should be fixed fully already with the recent move to nano-time there.

original-brownbear:
Yikes also seems some interop with `RubyArray` broke with JRuby 9.1.13.0 ... looking into that tomorrow

original-brownbear:
Note: List operations broke because of typed serialization. Deserializer now produces `ConvertedList` right away => elements don't get converted to Ruby vals => breaks here.
Not visible in Ruby execution (hence not a bug there) because the getters there contain the Rubyfier step. Will fix tomorrow :)

original-brownbear:
@andrewvc @jordansissel @ph I think we are feature complete here :) Just tried it out and `master` doesn't support big integer and big decimal in the grammar either (just add a really huge number, it throws an error).
=> tell me what I missed this time around I guess

jordansissel:
@original-brownbear I will look at this tomorrow! Exciting :)

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
andrewvc: Thanks for the explanation @original-brownbear . I'm +1 on this approach given your concerns.

logstash-core/spec/conditionals_spec.rb
```diff
@@ -218,161 +220,255 @@ def multi_receive(events)
...
-        sample("whatever") { expect(subject.get("tags")).to include("failure") }
+        sample_one("whatever") { expect(subject.get("tags")).to include("failure") }
       end
       conditional "![message] and ![message]" do
-        sample("whatever") { expect(subject.get("tags")).to include("failure") }
+        sample_one("whatever") { expect(subject.get("tags")).to include("failure") }
+      end
+    end
+
+    describe "nand" do
```
andrewvc: Thanks for adding these

logstash-core/spec/conditionals_spec.rb
original-brownbear: @andrewvc see above, I already made the order deterministic via `subject.sort! {|a, b| a.get("type") <=> b.get("type")}` :)

logstash-core/src/main/java/org/logstash/Javafier.java
```diff
@@ -57,11 +61,17 @@ private static Object fallbackConvert(final Object o, final Class<?> cls) {
...
+        converters.put(BigInteger.class, Valuefier.IDENTITY);
+        converters.put(BigDecimal.class, Valuefier.IDENTITY);
         converters.put(Timestamp.class, Valuefier.IDENTITY);
         // Explicitly casting to RubyString or RubySymbol when we know its type for sure is faster
         // than having the JVM look up the type.
         converters.put(RubyString.class, value -> ((RubyString) value).toString());
         converters.put(RubySymbol.class, value -> ((RubySymbol) value).toString());
+        converters.put(RubyBignum.class, value -> ((RubyBignum) value).getBigIntegerValue());
+        converters.put(
+            RubyBigDecimal.class, value -> ((RubyBigDecimal) value).getBigDecimalValue()
```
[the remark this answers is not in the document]
original-brownbear: @andrewvc there is a round trip in `EventTest` https://github.com/elastic/logstash/pull/7950/files#diff-eec69687901066a54bfa2ff47bcc0ea7R33 :)

logstash-core/spec/logstash/pipeline_spec.rb
```diff
@@ -608,38 +612,6 @@ class TestPipeline < LogStash::Pipeline
     end
   end
 
-  context "Multiples pipelines" do
```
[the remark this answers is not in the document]
original-brownbear: @andrewvc this was really specific to how the logic worked before and to some bug with reloading in the old code. The spec didn't even test any actual functionality imo, just some specifics of the reloaded pipeline not outputting `nil` events when calling `output_func` or `filter_func`

logstash-core/spec/conditionals_spec.rb
```diff
@@ -424,18 +521,78 @@ def multi_receive(events)
...
-      expect(subject[0].get("cond1")).to eq(nil)
+    sample_one({"type" => "original"}) do
+      expect(subject.length).to eq(3)
+      subject.sort! {|a, b| a.get("type") <=> b.get("type")}
+
+      expect(subject[0].get("type")).to eq("clone1")
+      expect(subject[0].get("cond1")).to eq("true")
       expect(subject[0].get("cond2")).to eq(nil)
 
-      expect(subject[1].get("type")).to eq("clone1")
```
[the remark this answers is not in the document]
original-brownbear: @andrewvc same answer here :) I handled this via `sort`. `select` was my first attempt, but it requires me to do 2 assertions for each field then (obviously have to make sure each event has a `type` and similar then -> this is much shorter and easier to follow imo since it has no logic in the asserts).

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,388 @@
...
+        });
+        return res;
+    }
+
+    /**
+     * Sets up all Ruby filters learnt from {@link PipelineIR}.
+     */
+    private Map<String, RubyIntegration.Filter> setupFilters() {
+        final Collection<PluginVertex> plugins = graph.getFilterPluginVertices();
+        final Map<String, RubyIntegration.Filter> res = new HashMap<>(plugins.size());
```
andrewvc: Shouldn't it be `new HashMap<>(plugins.size(), 1)`?

I don't quite get the point of hinting the exact size without also setting the load factor to 1, since we know the map will rehash at 0.75 without that.

Obvs. this is a micro-micro optimization and doesn't really matter, but if we're going to do it we may as well do it fully no?

I'm also fine with no size hint.

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,388 @@
...
+    /**
+     * Split the parent {@link Dataset}s and return the dataset half of their elements that contains
+     * the {@link JrubyEventExtLibrary.RubyEvent} that <strong>do not</strong> fulfil the given
+     * {@link EventCondition}.
+     * @param parents Datasets to split
+     * @param condition Condition that must be not be fulfilled
+     * @param index Vertex id to cache the resulting {@link Dataset} under
+     * @return The half of the datasets contents that does not fulfil the condition
+     */
+    private Dataset splitRight(final Collection<Dataset> parents,
```
andrewvc: Can we use a more semantic term than `split(Left|Right)`? Maybe `split(True|False)` instead?

original-brownbear: @andrewvc isn't "left" and "right" generally understood as true and false. It's used in books https://books.google.de/books?id=u5wkJ4R03d4C&pg=PA261&lpg=PA261&dq=left+branch+conditional&source=bl&ots=2iEMK4ma-3&sig=rO5OhVu1Cs9wyxq4cfu5NWYhEIo&hl=en&sa=X&ved=0ahUKEwj2gOT906XWAhXOh7QKHYBYDssQ6AEIWDAJ#v=onepage&q=left%20branch%20conditional&f=false I didn't make it up :D But sure I can rename that to true and false if it isn't :)

original-brownbear:
@andrewvc all but the renaming addressed :)

original-brownbear:
test this pleae

original-brownbear:
Jenkins test this please

original-brownbear:
Jenkins test this please

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,388 @@
...
+     * Compiles the next level of the execution from the given {@link Vertex} or simply return
+     * the given {@link Dataset} at the previous level if the starting {@link Vertex} cannot
+     * be expanded any further (i.e. doesn't have any more incoming vertices that are either
+     * a {code filter} or and {code if} statement).
+     * @param parents Nodes from the last already compiled level
+     * @param start Vertex to compile children for
+     * @return Datasets originating from given {@link Vertex}
+     */
+    private Collection<Dataset> flatten(final Collection<Dataset> parents, final Vertex start) {
+        final Collection<Vertex> endings = start.incomingVertices()
```
andrewvc: This is a confusing variable name no? The `endings` are really the `startParents` right?

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,388 @@
...
+     * Build a {@link Dataset} representing the {@link JrubyEventExtLibrary.RubyEvent}s after
+     * the application of the given output.
+     * @param vertex Vertex Id of the filter to create this {@link Dataset} for
+     * filter node in the topology once
+     * @param parents All the parent nodes that go into this output
+     * @return Output {@link Dataset}
+     */
+    private Dataset outputDataset(final String vertex, final Collection<Dataset> parents) {
+        final Dataset output;
+        if (plugins.containsKey(vertex)) {
```
andrewvc: This block of code would be a great place to use `Map.computeIfAbsent` for clarity.

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,388 @@
...
+    /**
+     * Build a {@link Dataset} representing the {@link JrubyEventExtLibrary.RubyEvent}s after
+     * the application of the given filter.
+     * @param vertex Vertex Id of the filter to create this {@link Dataset} for
+     * @param parents All the parent nodes that go into this filter
+     * @return Filter {@link Dataset}
+     */
+    private Dataset filterDataset(final String vertex, final Collection<Dataset> parents) {
+        final Dataset filter;
+        if (plugins.containsKey(vertex)) {
```
andrewvc: Also a good place for `Map.computeIfAbsent`

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,388 @@
...
+     * Checks if a certain {@link Vertex} represents a {@link RubyIntegration.Output}.
+     * @param vertex Vertex to check
+     * @return True iff {@link Vertex} represents a {@link RubyIntegration.Output}
+     */
+    private boolean isOutput(final Vertex vertex) {
+        return outputs.containsKey(vertex.getId());
+    }
+
+    /**
+     * Compiles the next level of the execution from the given {@link Vertex} or simply return
```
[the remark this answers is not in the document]
andrewvc: I'm also wondering if passing `parents` around makes the code harder to read. It might be simpler if we just used the map of `Vertex->Dataset` with the `ROOT_DATASETS` as a default value.

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: @andrewvc hmm -1 :D this really isn't the kind of method you'd expect behind `to_a` because we're also filtering out and joining two internal collections here (there's lots of magic in `each` imo). Imo (in FP), `collect` is precisely the name you'd put on that kind of operation?

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -0,0 +1,766 @@
...
+            } else {
+                return found != null && other != null && found.equals(other) ? TRUE : FALSE;
+            }
+        }
+
+        private static boolean listValueRight(final In in) {
+            return ((ValueExpression) in.getRight()).get() instanceof List;
+        }
+
+        private static boolean scalarValueRight(final In in) {
```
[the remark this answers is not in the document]
original-brownbear: Yea ... I'll shorten this. Initially, I used these in more spots but now they're pretty redundant :)

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,388 @@
...
+        // We sort the leaves of the graph in a deterministic fashion before compilation.
+        // This is not strictly necessary for correctness since it will only influence the order
+        // of output events for which Logstash makes no guarantees, but it greatly simplifies
+        // testing and is no issue performance wise since compilation only happens on pipeline
+        // reload.
+        iffs.clear();
+        plugins.clear();
+        pipelineIR.getGraph().allLeaves().filter(this::isOutput)
+            .sorted(Comparator.comparing(Vertex::hashPrefix)).forEachOrdered(
+            leaf -> datasets.add(outputDataset(leaf.getId(), flatten(Dataset.ROOT_DATASETS, leaf)))
```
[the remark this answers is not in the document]
original-brownbear: True will fix :)

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
original-brownbear: hmm, agree, though I don't like `startParents` either :) I'll see if I can make this clearer some other way. No matter the variable name here, the method is kinda hard to understand.

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,388 @@
...
+            .filter(v -> isFilter(v) || isOutput(v) || v instanceof IfVertex)
+            .collect(Collectors.toList());
+        return endings.isEmpty() ? parents : flattenChildren(parents, start, endings);
+    }
+
+    /**
+     * Compiles all child vertices for a given vertex.
+     * @param parents Parent datasets from previous stage
+     * @param start Start Vertex that got expanded
+     * @param children Children of {@code start} that are either {@code if} or {@code filter} in
```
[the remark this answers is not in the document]
original-brownbear: Right this changed eventually :) Fixing

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,388 @@
...
+     * type
+     * @return Datasets compiled from vertex children
+     */
+    private Collection<Dataset> flattenChildren(final Collection<Dataset> parents,
+        final Vertex start, final Collection<Vertex> children) {
```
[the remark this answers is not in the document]
original-brownbear: Hah yea sorry about that, let me fix that :)

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
original-brownbear: jup :) fixing

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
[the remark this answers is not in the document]
original-brownbear: @andrewvc yea you're right. The design here is somewhat evil in hindsight ... I think the problem starts here:

```java
    public Dataset buildExecution() {
        iffs.clear();
        plugins.clear();
```

I'm reusing some collections and passing down the `parents` ... in actuality, there should probably not be any mutable parts here ... I'll improve the design :)

original-brownbear:
@andrewvc agreed with almost all your points, thanks for the nice review (working in your suggestions saved around 80 lines of logic! :)) 

Also, I tried improving the readability of the compilation by refactoring things a little:

* I moved from the `parent` -> `child` terminology to `dataset` and `dependency` which is a lot more readable imo and also in line with the terminology used in Spark and similar DAG engines
* (This one is important in terms of avoiding concurrency issues bugs too) Moved all per-thread mutable state in the compilation to a separate nested class `org.logstash.config.ir.CompiledPipeline.CompiledExecution` inside of `org.logstash.config.ir.CompiledPipeline`.
   * We're avoiding temporal coupling those nasty `.clear()` calls
   * We see what state can be shared between threads (everything on `CompiledPipeline` and what state belongs to a single thread `CompiledExecution`).
     * We could try and improve this some more by not needing to set up the whole graph for each thread and cloning it somehow instead. The reason I did not do this was that it's a very small gain in startup time and kind of pointless if we try going for code gen next (seeing this completed now I can see how codegen would speed this up a lot from avoiding the mega-morphic call-sites due to `Dataset` and allowing for fewer `Collection` to `Collection` copies of `Event`)

let's see what you think :)

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,385 @@
...
+     */
+    private Map<String, RubyIntegration.Filter> setupFilters() {
+        final Collection<PluginVertex> filterPlugins = pipelineIR.getFilterPluginVertices();
+        final Map<String, RubyIntegration.Filter> res =
+            new HashMap<>(filterPlugins.size(), 1.0F);
+        for (final PluginVertex plugin : filterPlugins) {
+            final String ident = plugin.getId();
+            if (!res.containsKey(ident)) {
+                res.put(ident, buildFilter(plugin));
+            }
```
ph: **ids** should be unique per plugin, if it's not the case we shouldn't silently skip them. Also inputs/output should have the same logic? Maybe we should have a more complex type than a hashmap to act as a plugin registry which would ensure ids are unique?

ph: Also we have similar logic for setupInput/setupOutput, We could probably extract the pattern?

https://github.com/elastic/logstash/pull/7950/files#diff-484ab4968ddf73de6d56cb55c3b01ed0R128

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
```diff
@@ -0,0 +1,385 @@
...
+     * @return Compiled {@link RubyIntegration.Filter}
+     */
+    private RubyIntegration.Filter buildFilter(final PluginVertex vertex) {
+        final PluginDefinition def = vertex.getPluginDefinition();
+        final SourceWithMetadata source = vertex.getSourceWithMetadata();
+        return pipeline.buildFilter(
+            RubyUtil.RUBY.newString(def.getName()), RubyUtil.RUBY.newFixnum(source.getLine()),
+            RubyUtil.RUBY.newFixnum(source.getColumn()), convertArgs(def)
+        );
+    }
```
ph: Maybe use the same pattern for the input / output?

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -431,10 +421,13 @@ def start_workers
       if max_inflight > MAX_INFLIGHT_WARN_THRESHOLD
         @logger.warn("CAUTION: Recommended inflight events max exceeded! Logstash will run with up to #{max_inflight} events in memory in your current configuration. If your message sizes are large this may cause instability with the default heap size. Please consider setting a non-standard heap size, changing the batch size (currently #{batch_size}), or changing the number of pipeline workers (currently #{pipeline_workers})", default_logging_keys)
       end
+      
+      @filter_queue_client.set_batch_dimensions(batch_size, batch_delay)
 
       pipeline_workers.times do |t|
-        thread = Thread.new(batch_size, batch_delay, self) do |_b_size, _b_delay, _pipeline|
-          _pipeline.worker_loop(_b_size, _b_delay)
+        batched_execution = @lir_execution.buildExecution
```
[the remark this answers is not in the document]
ph: nevermind, I've looked on the java side after ;)

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -648,22 +586,13 @@ def shutdown_workers
   end
 
   # for backward compatibility in devutils for the rspec helpers, this method is not used
-  # in the pipeline anymore.
+  # anymore and just here to not break TestPipeline that inherits this class.
   def filter(event, &block)
-    maybe_setup_out_plugins
-    # filter_func returns all filtered events, including cancelled ones
-    filter_func(event).each {|e| block.call(e)}
   end
```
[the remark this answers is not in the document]
original-brownbear: @ph the problem is that removing this thing requires a new dev-utils release. It does this:

```rb

`  class TestPipeline < LogStash::Pipeline
    public :filter
    public :flush_filters
  end
```

this makes the whole thing very involved. I'd rather fix this in the next iteration, otherwise, we're introducing a whole new range of new issues from changing things up in dev-utils right?

logstash-core/src/test/java/org/logstash/common/io/DeadLetterQueueReaderTest.java
```diff
@@ -209,7 +209,7 @@ private void validateEntries(Path firstLog, int startEntry, int endEntry, int st
     @Test
```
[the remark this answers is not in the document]
original-brownbear: @ph ah that's a result of the painfully slow review queue :D This patch contains a fix to the way `RubyString` is serialized (I needed direct deserialization to not get a regression in PQ performance here). So now the size of events has changed slightly due to the different serialization format.

logstash-core/src/main/java/org/logstash/ObjectMappers.java
```diff
@@ -40,7 +41,8 @@
     private static final SimpleModule CBOR_DESERIALIZERS =
         new SimpleModule("CborRubyDeserializers")
             .addDeserializer(RubyBigDecimal.class, new RubyBigDecimalDeserializer())
-            .addDeserializer(RubyBignum.class, new RubyBignumDeserializer());
+            .addDeserializer(RubyBignum.class, new RubyBignumDeserializer())
+            .addDeserializer(RubyString.class, new RubyStringDeserializer());
```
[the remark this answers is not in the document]
original-brownbear: @ph I could but man :D, I've been down that route before and here is where I a gave up. It literally took a month to get ~5 dependencies of this kind reviewed for this one. I'd rather deal with this in one go at this point if it's ok? :)

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -0,0 +1,758 @@
...
+                this.regex = RubyUtil.RUBY.newString(regex);
+            }
+
+            @Override
+            public boolean fulfilled(final JrubyEventExtLibrary.RubyEvent event) {
+                final Object tomatch = event.getEvent().getUnconvertedField(field);
+                return tomatch instanceof RubyString &&
+                    !((RubyString) tomatch).match(RubyUtil.RUBY.getCurrentContext(), regex).isNil();
+            }
+        }
```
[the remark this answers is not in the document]
original-brownbear: @ph yea sadly there is a measurable gain in JDK regex because we don't have to actually capture, we just need to `find` and the Ruby API doesn't seem to offer this. Very small hit though + JDK regex come with some other problems potentially (either I have to convert RubysString to `String` before the matching or deal with potential encoding issues when using the `ByteList` in `RubyString` as a `CharBuffer` directly)

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -0,0 +1,758 @@
...
+                super(
+                    String.format("Unexpected input types %s %s", left.getClass(), right.getClass())
+                );
+            }
+
+            UnexpectedTypeException(final Object inner) {
+                super(String.format("Unexpected input type %s", inner.getClass()));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
original-brownbear: This is just at the compiler level. I think you can read from the stack-trace quite easily also this should be impossible to run into. If this is thrown there is a case missing in the compiler logic, a broken config can't cause this (if it can then I missed a something, that's why this is there :)).

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
original-brownbear: @ph actually that check is redundant, let me remove it :)

original-brownbear: @ph the current approach is basically a result of me not wanting to have the Ruby constant strings for "filter", "output" and so on on the Java side. I fear that the way I call `pipeline.buildFilter` and the other can't really be dried up without a dynamic method call, this is where Java is just Java, it annoyed me as well :)

logstash-core/src/main/java/org/logstash/config/ir/CompiledPipeline.java
original-brownbear: @ph same answer here, only way to do that is to pull Ruby String constants over to Java.
I think the only other approach really is to use a more sophisticated wrapper for all of this, but  to me it seems like it's just noise since all of this is kinda in-flux right now => time for adding wrappers and such for the Ruby end is probably better spent moving more of the logic to Java instead?

logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
original-brownbear: hmm I don't agree, but I also don't think it's important enough to argue. Renaming but still:
* you have argument free in at least `collect` in Spark + Scala (in a sense also in MapReduce) it's 
* `to_a` is a nasty name to bring into you Java APIs
*  the fact that there is magic like `uncancellel` going on behind the scenes makes this kinda meh just present as `to_a` still push incoming :D

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -648,22 +586,13 @@ def shutdown_workers
...
-  # @option options [Boolean] :final => true to signal a final shutdown flush
+  # for backward compatibility in devutils for the rspec helpers, this method is not used
+  # anymore and just here to not break TestPipeline that inherits this class.
   def flush_filters(options = {}, &block)
-    flushers = options[:final] ? @shutdown_flushers : @periodic_flushers
-
-    flushers.each do |flusher|
-      flusher.call(options, &block)
-    end
   end
```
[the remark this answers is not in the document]
original-brownbear: @ph same here, if we do that it's another round of reviews, this time also in dev-utils. Given past experience with reviews, I'd rather not do it in this step to not add another week to this.

original-brownbear:
@ph all code comments addressed I think :)

> This is a lot of good changes, I am conscious we need to take this major step to add this functionality.
From the changes I see, I think we are expecting that our current test suite will catch all the bugs that this new implementation could create, I've run the test locally and all the existing tests are green on my machine.
But I would expect that we add java coverage for all theses new classes, as we move forward to include more java class, we will hopefully remove the ruby file.

Right now testing this in Java is non-starter, you'd have to write a lot of code to have some fake plugins + you can't really do the parsing of the config language from Java (unless you go through the pain of loading treetop in JUnit tests, possible but very time intensive :()

> Also, could we split some of theses big java file into smaller file?

We could, but I'm not so sure it's worth it at this point. With the APIs in flux and code gen on the horizon it seems to me this just takes a lot of time (not necessarily mine in terms of coding, but just waiting for more review iterations) for a trivial cosmetic change?

> Maybe its worth it to add a new benchmark class to test this new implementation with a few common use case? (Can be done in another PR.)

Same reason we can't have any JUnit tests without big efforts also applies to JMH here I fear :( The existing end to end benchmarks obviously continue to work, though I guess we could add a few more when there's some time :)

original-brownbear:
@andrewvc @ph any updates here?

original-brownbear:
@andrewvc @ph also, umm what do you guys think about squashing this one and creating a new PR? :D My browser is starting to lag here with the many many comments we have at this point ...

ph:
@original-brownbear I believe you have addressed all my concerns.

@andrewvc What do you think about testing of this PR, I though you have raised a similar concerns about it, but I cannot find the corresponding comment. I think the main question is **are we confortable with the existing ruby tests to merge this change**.

Personally, I've run the tests, I used all the configs I had and I didn't run into any issues. But, this change will go in a minor release and will be on by default.

---

> We could, but I'm not so sure it's worth it at this point. With the APIs in flux and code gen on the horizon it seems to me this just takes a lot of time (not necessarily mine in terms of coding, but just waiting for more review iterations) for a trivial cosmetic change?

I think if we had a style guide that would solve that discussion :)

ph:
@original-brownbear +1 to squash.

original-brownbear:
continued in #8357

PR #7953, opened 2017-08-09
MINOR: Cleanup dead IO code

original-brownbear:
Trivial cleanup, these are shown as dead code in Idea + removing them doesn't break tests.
=> No need to keep 100 lines of dead code around

PS: I think this is just code that was copied from ES and not cleaned up and this was never used.


jsvd:
Humm, although this code is not executed, I'm not sure if I feel comfortable giving the ok to delete. I'll defer to @colinsurprenant for this, so I suggest removing the elasticsearch-bot so it's not on the pr watch list.

original-brownbear:
@jsvd sure let's do it ~like that :)

guyboertje:
The code intended for removal here is mostly shamelessly stolen from ES. It was included while we were initially building the solution that Boas, Colin and I agreed upon.

I'm happy to have it removed now - we did not find a use for it.

PR #7955, opened 2017-08-09
cleanup force shutdown code

jsvd:
since https://github.com/elastic/logstash/pull/7918 was merged, force shutdown now executes `exit(1)` immediately so we can cleanup all the `force_shutdown` related code that is no longer executed


jakelandis:
LGTM

_searched for other `force_shutdown!` calls and could not find any_

jsvd:
@suyograo should we merge this sort of PR into 6.0? it's not a feature, only a minor code cleanup

Issue #7951, opened 2017-08-08
Minor: Refactor BiValues.RUBY


jakelandis:
We currently cache the Ruby global runtime in a static variable inside in the `BiValues` class.

```java
public static final Ruby RUBY = Ruby.getGlobalRuntime();
```

The Ruby global runtime isn't really an attribute of the `BiValues` and should be moved to somewhere more global. It still needs to be cached statically, for performance reasons. 

I would suggest the following:

```java
package org.logstash;

public interface Ruby {
    org.jruby.Ruby RUNTIME = org.jruby.Ruby.getGlobalRuntime();

    static org.jruby.Ruby getGlobalRuntime(){
        return RUNTIME;
    }
}
```

Usually I am not a fan of mirroring names/methods, but in this case I think it makes sense. 

cc @original-brownbear

original-brownbear:
fixed in #7957

PR #7957, opened 2017-08-09
Test Stability: Fix Global Ruby Runtime Setup

original-brownbear:
Fixes test issues from the "LogStash" `Ruby` module not being set up before a test due to the class load order.
In production code, we're guaranteed to have the global `Ruby` all setup, but in tests, no such guarantee exists.

Fixes https://logstash-ci.elastic.co/job/elastic+logstash+pull-request+multijob-linux-intake/423/console


original-brownbear:
@jakelandis how about this as a solution to https://github.com/elastic/logstash/issues/7951 as well? :)

jakelandis:
LGTM

.. and yes, this fixes #7951 too.

PR #7962, opened 2017-08-09
prevent tests from changing config.debug flag

jsvd:
a few tests can globally enable debug causing builds to have 53mb of raw logs like https://logstash-ci.elastic.co/job/elastic+logstash+pull-request+multijob-linux-intake/430/


jordansissel:
LGTM. Test failures seem unrelated.

PR #7965, opened 2017-08-09
Add --setup phase for modules

### LINKED ISSUE #7959: Add setup phase for modules
We should add a `--setup` phase since the current behavior with modules is that we always upload dashboards and templates when a user runs it. There are 2 disadvantages:

* User needs to supply Kibana and ES URL + credentials all the time which are not ideal.
* If a user modified the dashboard on the Kibana side, we'll overwrite them without warning. 

`--setup` makes explicit we're uploading dashboard and template files. `--setup` doesn't need to be run all the time. This also makes LS modules consistent with beats behavior.

```sh
bin/logstash --modules netflow --setup -M "netflow.var.elasticsearch.password=foobar" -M "netflow.var.kibana.password=foobar" -M "netflow.var.kibana.username=elastic"
```

The above should upload dashboards, templates _and_ start the module.

Next time, they can just run

```sh
bin/logstash --modules netflow
```

The above command will only start the module and assume dashboards are already set up.

untergeek:
Without the `--setup` flag, the index template for elasticsearch, and the index-pattern, saved searches, visualizations, and dashboards for Kibana will not be published.

fixes #7959

### DIFF
--- a/logstash-core/lib/logstash/config/modules_common.rb
+++ b/logstash-core/lib/logstash/config/modules_common.rb
@@ -66,7 +66,9 @@ module LogStash module Config
           module_settings.set("pipeline.id", pipeline_id)
           current_module.with_settings(module_hash)
           config_test = settings.get("config.test_and_exit")
-          if !config_test
+          modul_setup = settings.get("modules_setup")
+          # Only import data if it's not a config test and --setup is true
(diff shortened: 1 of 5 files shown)


suyograo:
LGTM

PR #7966, opened 2017-08-10
[Docs] Add docs for pipeline viewer

dedemorton:


dedemorton:
Adding this content to the Logstash docs for now. Right now we have a very fragmented user experience for monitoring features (we force users to jump to the Kibana Reference for some of the monitoring content and it's been causing confusion). I didn't want to add to the confusion with pipeline viewer, so I added all the content to the LS docs for now, but we might need to move it later.

docs/static/pipeline-viewer.asciidoc
```diff
@@ -0,0 +1,107 @@
...
+
+The pipeline viewer in {xpack} provides a simple way for you to visualize and
+monitor the behavior of complex Logstash pipeline configurations. Within the
+pipeline viewer, you can explore a directed acyclic graph (DAG) representation
+of the overall pipeline topology, data flow, and branching logic. The diagram
+is overlayed with important metrics, like events per second and time spent in
+milliseconds, for each plugin in the view.
+
+The diagram includes visual indicators to draw your attention to potential
+bottlenecks in the pipeline, making it easy for you to diagnose and fix
```
suyograo: We should explain what heuristic we use to label the problem areas. Folks are gonna wonder how we label them and if its configurable. It is not configurable by design. @andrewvc can you paste a blob here about the heuristic?

docs/static/pipeline-viewer.asciidoc
```diff
@@ -0,0 +1,107 @@
...
+problems.
+
+Before using the pipeline viewer, you need to
+{logstash-ref}/setup-xpack.html[set up {xpack}] and configure
+{xpack-ref}/monitoring-logstash.html[Logstash monitoring]. 
+
+//REVIEWERS: Is anything else required to use the pipeline viewer?
```
suyograo: There should be a big note/info/reccomendation about using semantic IDs in configs. If the ID is not present in the configuration, we generate an automatic ID but this is not ideal for debugging purposes. This will be evident when there are 3-4 grok debuggers and its better if user knows which grok filter does what. So adding a semantic ID like `apacheParsingGrok` or `cloudwatchGrok` helps to narrow which of the many groks are slow. See https://github.com/elastic/x-pack-kibana/pull/2218

docs/static/pipeline-viewer.asciidoc
```diff
@@ -0,0 +1,107 @@
...
+being monitored.
++
+Each pipeline is identified by a pipeline ID (`main` by default). For each
+pipeline, you'll see a list of all versions of the pipeline running during
+the specified time range.
++
+image::static/docs/images/pipeline-viewer-overview.png[Pipeline Overview]
++
+The version information is auto-generated by Logstash. Each time you modify a
+pipeline, Logstash generates a new version hash. Viewing different versions
```
suyograo: Lets also mention that we don't store multiple versions of the pipeline configuration itself, just the pipeline stats from latest one.

docs/static/pipeline-viewer.asciidoc
dedemorton: I thought about including our formula, but isn't that likely to change? Is there some way we could describe the heuristic without being *too* precise?

Also, I wonder if this is something that will be configurable at some point. It should be...so users can change the setting that determines what gets flagged.

docs/static/pipeline-viewer.asciidoc
```diff
@@ -46,15 +56,17 @@ image::static/docs/images/monitoring-ui.png[Monitoring UI]
 being monitored.
 +
 Each pipeline is identified by a pipeline ID (`main` by default). For each
-pipeline, you'll see a list of all versions of the pipeline running during
-the specified time range.
+pipeline, you'll see a list of all versions of the pipeline stats that were
```
dedemorton: In addition to the statement you requested, I added a bit more text here so users don't think they're viewing a configuration that's been saved somewhere.

suyograo:
LGTM

Issue #7885, opened 2017-08-02
Metrics: Consistent serialization


jakelandis:
Currently the metrics are only serialized if the value is set.  This can lead to some surprising behavior if as a consumer, you expect the value to be present. 

To complicate matters, some, but not all metrics are pre-initialized creating a mix of zero/null and missing metrics in the JSON output. 

This issue is to explicitly change the serialization to ALWAYS serialize the known metrics. This is only possible with the changes over at #7788   

Part of this change should be to remove the deprecated `dirty` flag, whose sole purpose is to lazily serialize the data.

jakelandis:
@andrewvc / @original-brownbear 
The existing PR is does does not take in account the DLQ and PQ metrics. _(They are not represented by the Witness code yet, PR coming soon)._

We discussed the zero/null vs. missing, and agree the decision to simply always show those metrics. 

However, for DLQ and PQ, since those metrics are driven by a configuration, should we continue to flex their output. 

**For DLQ and PQ should we (continue to) hide the zero/null/empty metrics when they are not enabled?**

original-brownbear:
@jakelandis @andrewvc imo yes, we shouldn't show metrics for functionality that isn't enabled.

Issue #7971, opened 2017-08-10
Complete replacement of Ruby metrics / metrics namespaces in 6.1 ?


jakelandis:
I am currently working on the Java-fication for recording the metrics #7788. This is essentially a replacement of the Ruby metrics / metrics namespace API. 

It was always my intention to completely remove the Ruby metrics / metrics namespaces for 7.x, but use a bridge strategy ([mapping namespaces -> witness](https://github.com/jakelandis/logstash/blob/domain-metrics-bak4/logstash-core/lib/logstash/instrument/metric_type/witness_adaptor.rb)) for 6.x.  

This was out of concern for passivity w/r/t to the Ruby metrics API. However, due to lack luster performance of the bridge,  complexity of maintaining two subsystems that do the same thing, and lack of publicly documenting the Ruby metrics / metrics namespace API... I would like to solicit approval to completely remove the Ruby metrics / metrics namespaces for the 6.1 release.  

-----------
Additional relevant information:

Thus far I have:
* Replaced the underlying counters and guage state with Java objects (6.0)
* Added a `Witness` API in Java as alternative/abstraction to working directly with the underlying counters/gauges for the stats namespace. _(in review, still need to do the JVM namespace)_
* Moved the JSON serialization from Ruby to Java _(soon to be in review)_
* Created a metric namesapce -> `Witness` bridge that maps the Ruby metrics/namspace to the Java `Witness` API.  _(not reviewed yet)_

This bridge strategy has allowed me to replicate the recording of the :stats namespace with minimal (almost no) changes to the existing Ruby code (e.g. pipelines/agents/etc. still express their metrics in terms of a namespace). I even have the rspecs to passing locally with very minor changes using the bridge code. However ....**performing the conversion from a Ruby namespace to a  `Witness` regresses any performance gains made recently to the metrics subsystem.** 

So, the decision to update the Ruby code to use the new API directly (instead of going through the bridge) is easy. 

...but once I update all the Ruby code to the new API, that leaves the Ruby metrics / metrics namespaces code just sitting there in case someone's plugins leveraged it. 

This plugin strategy for metrics is TBD, but irregardless, I will need to visit each supported plugin and update _some code somewhere_ to properly support those metrics.

suyograo:
[...]

PR #7984, opened 2017-08-11
Metrics: Add Jackson serialization to the Witness classes.

jakelandis:
This PR builds on  https://github.com/elastic/logstash/pull/7947 by adding in the Jackson Serialization to the `Witness` objects.  

The `dirty` flag and/or other oddities (like explicity pushing a `null` string to JSON) are there for passivity w/r/t the JSON output. I am striving for identical output from the API for 6.x. 

Eventually, for 7.x, we will adopt https://github.com/elastic/logstash/issues/7885 and remove some of the custom rules here. 

Note - matching the JSON output to the Ruby serialization is an ongoing effort, and expect to see some tweaking to this in the reviews to come...but at least for the simplest cases this JSON is identical to the Ruby implementation.  

Part of #7788

### DIFF
--- a/logstash-core/src/main/java/org/logstash/instrument/witness/ConfigWitness.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/witness/ConfigWitness.java
@@ -1,12 +1,20 @@
 package org.logstash.instrument.witness;
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import org.logstash.instrument.metrics.Metric;
 import org.logstash.instrument.metrics.gauge.BooleanGauge;
 import org.logstash.instrument.metrics.gauge.LongGauge;
+
+import java.io.IOException;
 
 /**
  * The witness for configuration.
  */
-final public class ConfigWitness {
+@JsonSerialize(using = ConfigWitness.Serializer.class)
+final public class ConfigWitness implements SerializableWitness {
 
     private final BooleanGauge deadLetterQueueEnabled;
     private final BooleanGauge configReloadAutomatic;
@@ -15,6 +23,9 @@ final public class ConfigWitness {
     private final LongGauge batchDelay;
     private final LongGauge configReloadInterval;
     private final Snitch snitch;
+    private final static String KEY = "config";
+    private static final Serializer SERIALIZER = new Serializer();
+
 
     /**
      * Constructor.
@@ -92,6 +103,55 @@ final public class ConfigWitness {
         return this.snitch;
     }
 
+    @Override
+    public void genJson(JsonGenerator gen, SerializerProvider provider) throws IOException {
+        SERIALIZER.innerSerialize(this, gen, provider);
+    }
+
+    /**
+     * The Jackson serializer.
+     */
+    static class Serializer extends StdSerializer<ConfigWitness> {
+
+        /**
+         * Default constructor - required for Jackson
+         */
+        public Serializer() {
+            this(ConfigWitness.class);
+        }
+
+        /**
+         * Constructor
+         *
+         * @param t the type to serialize
+         */
+        protected Serializer(Class<ConfigWitness> t) {
+            super(t);
+        }
+
+        @Override
+        public void serialize(ConfigWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeStartObject();
+            innerSerialize(witness, gen, provider);
+            gen.writeEndObject();
+        }
+
+        void innerSerialize(ConfigWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeObjectFieldStart(KEY);
+
+            MetricSerializer<Metric<Long>> longSerializer = MetricSerializer.Get.longSerializer(gen);
+            MetricSerializer<Metric<Boolean>> booleanSerializer = MetricSerializer.Get.booleanSerializer(gen);
+
+            longSerializer.serialize(witness.batchSize);
+            longSerializer.serialize(witness.workers);
+            longSerializer.serialize(witness.batchDelay);
+            longSerializer.serialize(witness.configReloadInterval);
+            booleanSerializer.serialize(witness.configReloadAutomatic);
+            booleanSerializer.serialize(witness.deadLetterQueueEnabled);
+            gen.writeEndObject();
+        }
+    }
+
     /**
      * The snitch for the errors. Used to retrieve discrete metric values.
      */
@@ -105,6 +165,7 @@ final public class ConfigWitness {
 
         /**
          * Gets the configured batch delay
+         *
          * @return the batch delay
          */
         public long batchDelay() {
@@ -114,6 +175,7 @@ final public class ConfigWitness {
 
         /**
          * Gets the configured batch size
+         *
          * @return the batch size
          */
         public long batchSize() {
@@ -122,6 +184,7 @@ final public class ConfigWitness {
 
         /**
          * Gets if the reload automatic is configured
+         *
          * @return true if configured for automatic, false otherwise
          */
         public boolean configReloadAutomatic() {
@@ -130,6 +193,7 @@ final public class ConfigWitness {
 
         /**
          * Gets the configured reload interval
+         *
          * @return the configured reload interval
          */
         public long configReloadInterval() {
@@ -138,7 +202,8 @@ final public class ConfigWitness {
 
         /**
          * Gets if the dead letter queue is configured to be enabled
-          * @return true if the dead letter queue is configured to be enabled, false otherwise
+         *
+         * @return true if the dead letter queue is configured to be enabled, false otherwise
          */
         public boolean deadLetterQueueEnabled() {
             return witness.deadLetterQueueEnabled.getValue();
@@ -146,11 +211,15 @@ final public class ConfigWitness {
 
         /**
          * Gets the number of configured workers
+         *
          * @return the configured number of workers.
          */
         public long workers() {
             return witness.workers.getValue();
         }
 
+
     }
+
+
 }
--- a/logstash-core/src/main/java/org/logstash/instrument/witness/ErrorWitness.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/witness/ErrorWitness.java
@@ -1,5 +1,10 @@
 package org.logstash.instrument.witness;
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import org.logstash.instrument.metrics.Metric;
 import org.logstash.instrument.metrics.gauge.TextGauge;
 
 import java.io.ByteArrayOutputStream;
@@ -10,11 +15,16 @@ import java.nio.charset.StandardCharsets;
 /**
  * Witness for errors.
  */
-public class ErrorWitness {
+@JsonSerialize(using = ErrorWitness.Serializer.class)
+public class ErrorWitness implements SerializableWitness {
 
     private final TextGauge message;
     private final TextGauge backtrace;
     private final Snitch snitch;
+    private final static String KEY = "last_error";
+    private static final Serializer SERIALIZER = new Serializer();
+
+    private boolean dirty; //here for passivity with legacy Ruby implementation
 
     public ErrorWitness() {
         message = new TextGauge("message");
@@ -29,6 +39,7 @@ public class ErrorWitness {
      */
     public void backtrace(String stackTrace) {
         this.backtrace.set(stackTrace);
+        dirty = true;
     }
 
     /**
@@ -38,6 +49,7 @@ public class ErrorWitness {
      */
     public void message(String message) {
         this.message.set(message);
+        dirty = true;
     }
 
     /**
@@ -66,6 +78,52 @@ public class ErrorWitness {
         } catch (IOException e) {
             //A checked exception due to a the close on a ByteArrayOutputStream is simply annoying since it is an empty method.  This will never be called.
             throw new IllegalStateException("Unknown error", e);
+        }
+    }
+
+    @Override
+    public void genJson(JsonGenerator gen, SerializerProvider provider) throws IOException {
+        SERIALIZER.innerSerialize(this, gen, provider);
+    }
+
+    /**
+     * The Jackson serializer.
+     */
+    public static class Serializer extends StdSerializer<ErrorWitness> {
+
+        /**
+         * Default constructor - required for Jackson
+         */
+        public Serializer() {
+            this(ErrorWitness.class);
+        }
+
+        /**
+         * Constructor
+         *
+         * @param t the type to serialize
+         */
+        protected Serializer(Class<ErrorWitness> t) {
+            super(t);
+        }
+
+        @Override
+        public void serialize(ErrorWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeStartObject();
+            innerSerialize(witness, gen, provider);
+            gen.writeEndObject();
+        }
+
+        void innerSerialize(ErrorWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            if (witness.dirty) {
+                gen.writeObjectFieldStart(KEY);
+                MetricSerializer<Metric<String>> stringSerializer = MetricSerializer.Get.stringSerializer(gen);
+                stringSerializer.serialize(witness.message);
+                stringSerializer.serialize(witness.backtrace);
+                gen.writeEndObject();
+            } else {
+                gen.writeStringField(KEY, null);
+            }
         }
     }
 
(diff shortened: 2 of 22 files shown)


logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
```diff
@@ -50,12 +64,23 @@ public PluginWitness filters(String id) {
     }
 
     /**
+     * Gets the {@link PluginWitness} for the given id, creates the associated {@link PluginWitness} if needed
+     *
+     * @param id the id of the codec
+     * @return the associated {@link PluginWitness} (for method chaining)
+     */
+    public PluginWitness codecs(String id) {
```
jakelandis: whoops, this snuck into the review. 

Turns out that we record codec metrics as a plugin (shares the same Ruby code with the Input).  I am not sure if that desired behavior or accidental. 

I haven't run across any consumption of this, but didn't want to change the Ruby code, so I opted to simply support codecs as new plugin metrics.

andrewvc:
Question WRT to keeping around these attributes for passivity with the JSON output. Can you cite examples where there is a discrepancy? 

All the instances I can think of are places where I would describe the lack of a key being there as being an existing bug.

I'm thinking it might make sense to drop the dirty flag and just patch the tests (which we'll need to do anyway). Compatibility is important, but the existing behavior with 'maybe' values is broken IMHO. What's your take on it @jakelandis , seeing as how you're knee deep in the code?

logstash-core/src/main/java/org/logstash/instrument/witness/ConfigWitness.java
```diff
@@ -92,6 +100,52 @@ public Snitch snitch() {
...
+        public void serialize(ConfigWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeStartObject();
+            innerSerialize(witness, gen, provider);
+            gen.writeEndObject();
+        }
+
+        void innerSerialize(ConfigWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeObjectFieldStart(witness.KEY);
+
+            MetricSerializer.Get.longSerializer(gen).serialize(witness.batchSize);
```
original-brownbear: @jakelandis let's cache the return of this to a local variable

```java
MetricSerializer.Get.longSerializer(gen)
```
?

logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
```diff
@@ -1,17 +1,25 @@
...
-final public class EventsWitness{
+@JsonSerialize(using = EventsWitness.Serializer.class)
+final public class EventsWitness implements SerializableWitness {
 
     private LongCounter filtered;
     private LongCounter out;
     private LongCounter in;
     private LongCounter duration;
     private LongCounter queuePushDuration;
+    final String KEY = "events";
```
original-brownbear: @jakelandis missing `static`

logstash-core/src/main/java/org/logstash/instrument/witness/ErrorWitness.java
```diff
@@ -10,11 +14,15 @@
...
  * Witness for errors.
  */
-public class ErrorWitness {
+@JsonSerialize(using = ErrorWitness.Serializer.class)
+public class ErrorWitness implements SerializableWitness {
 
     private final TextGauge message;
     private final TextGauge backtrace;
     private final Snitch snitch;
+    private final String KEY = "last_error";
```
original-brownbear: @jakelandis missing `static`

logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
```diff
@@ -124,6 +132,59 @@ public void queuePushDuration(long durationToAdd) {
...
     }
 
+    @Override
+    public String asJson() throws IOException {
+        return dirty ? SerializableWitness.super.asJson() : "";
+    }
+
+    @Override
+    public void genJson(final JsonGenerator gen, SerializerProvider provider) throws IOException {
+        new Serializer().innerSerialize(this, gen, provider);
```
original-brownbear: @jakelandis does `Serializer` have any state? It looks like we can just make it a constant right?

logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
```diff
@@ -86,4 +93,43 @@ public PipelineWitness pipeline(String name) {
         return pipelinesWitness.pipeline(name);
     }
 
+    @Override
+    public void genJson(JsonGenerator gen, SerializerProvider provider) throws IOException {
+        new Serializer().innerSerialize(this, gen, provider);
```
original-brownbear: @jakelandis same here, `Serializer` seems like it should just be a static constant since it has no state

logstash-core/src/main/java/org/logstash/instrument/witness/ConfigWitness.java
```diff
@@ -92,6 +100,52 @@ public Snitch snitch() {
         return this.snitch;
     }
 
+    @Override
+    public void genJson(JsonGenerator gen, SerializerProvider provider) throws IOException {
+        new Serializer().innerSerialize(this, gen, provider);
```
original-brownbear: @jakelandis here too, we're calling this method a ton, we should make this `Serializer` something `static` since it has no state?

logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
```diff
@@ -69,5 +94,59 @@ private PluginWitness getPlugin(Map<String, PluginWitness> plugin, String id) {
...
+            serializePlugins("filters", witness.filters, gen, provider);
+            serializePlugins("outputs", witness.outputs, gen, provider);
+            //codec is not serialized
+
+            gen.writeEndObject();
+        }
+
+        private void serializePlugins(String key, Map<String, PluginWitness> plugin, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeArrayFieldStart(key);
+            for (Map.Entry<String, PluginWitness> entry : plugin.entrySet()) {
```
original-brownbear: @jakelandis shouldn't this just work out of the box if we have a `Serializer` for `PluginWitness ` registered with Jackson? Meaning, we should simply be able to `com.fasterxml.jackson.core.JsonGenerator#writeObject` the `plugin` variable? (this comment applies to all the similar spots we have in the code as well)

logstash-core/src/main/java/org/logstash/instrument/witness/ReloadWitness.java
```diff
@@ -99,6 +113,50 @@ public void lastFailureTimestamp(JrubyTimestampExtLibrary.RubyTimestamp timestam
...
+        public void serialize(ReloadWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeStartObject();
+            innerSerialize(witness, gen, provider);
+            gen.writeEndObject();
+        }
+
+        void innerSerialize(ReloadWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeObjectFieldStart(ReloadWitness.KEY);
+            witness.lastError.genJson(gen, provider);
+            MetricSerializer.Get.longSerializer(gen).serialize(witness.success);
```
original-brownbear: All these serializers could/should be cached to static constants?

logstash-core/src/main/java/org/logstash/instrument/witness/PipelinesWitness.java
```diff
@@ -27,4 +36,46 @@ public PipelineWitness pipeline(String name) {
...
+        @Override
+        public void serialize(PipelinesWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeStartObject();
+            innerSerialize(witness, gen, provider);
+            gen.writeEndObject();
+        }
+
+        void innerSerialize(PipelinesWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeObjectFieldStart(KEY);
+            for (Map.Entry<String, PipelineWitness> entry : witness.pipelines.entrySet()) {
```
original-brownbear: Same as in other spots, since we have a serializer for `PipelineWitness` we should be able to simply write the whole `Map` to Jackson via `com.fasterxml.jackson.core.JsonGenerator#writeObject`?

original-brownbear: Since the same holds true for many of these spots, maybe this method's use can just be replaced by `com.fasterxml.jackson.core.JsonGenerator#writeObjectField` where we just serialize a `Map`

logstash-core/src/main/java/org/logstash/instrument/witness/QueueWitness.java
```diff
@@ -36,10 +43,49 @@ public void type(String type) {
...
+
+        @Override
+        public void serialize(QueueWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeStartObject();
+            innerSerialize(witness, gen, provider);
+            gen.writeEndObject();
+        }
+
+        void innerSerialize(QueueWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeObjectFieldStart(witness.KEY);
```
original-brownbear: @jakelandis won't weird things happen here, if our custom serializer from `MetricSerializer.Get` at times just does nothing but you still write an object field start?

logstash-core/src/main/java/org/logstash/instrument/witness/QueueWitness.java
```diff
@@ -49,10 +95,9 @@ public void type(String type) {
 
         /**
          * Gets the type of queue
-         *
          * @return the queue type.
          */
-        public String type() {
+        public String type(){
```
original-brownbear: @jakelandis you made this formatting change (removing the space) in many spots, sure you want this? I think the majority of our codebase has that space there right now? :D

logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
```diff
@@ -20,10 +30,12 @@ public PluginsWitness() {
         this.inputs = new ConcurrentHashMap<>();
         this.outputs = new ConcurrentHashMap<>();
         this.filters = new ConcurrentHashMap<>();
-     }
+        this.codecs = new ConcurrentHashMap<>();
```
original-brownbear: @jakelandis I just saw we have a bunch of `ConcurrentHashMap`s here, maybe we should initialize them a little smaller (concurrency level 1 should be fine here right? We rarely update the actual map here?)

jakelandis:
> Can you cite examples where there is a discrepancy?

Removing the dirty flag on this PR, and comparing it against 5.5, the following is not present on an newly started empty pipeline:
```
      "outputs" : [ {
....
          "events" : {
            "duration_in_millis" : 0,
            "in" : 0,
            "out" : 0
          }
```
There may be other cases.

I am open to calling that a bug, and removing the `dirty` flag and implement https://github.com/elastic/logstash/issues/7885 as part of this PR.  (had planned on doing it for 7.x only)  

My goal was to minimize the changes to the serialized output in the name of passivity...but I also don't want to add code to carry forward defective behavior.

logstash-core/src/main/java/org/logstash/instrument/witness/QueueWitness.java
jakelandis: Not sure what happened there... my formatter has the space defined. I will re-run the formatter.

logstash-core/src/main/java/org/logstash/instrument/witness/ConfigWitness.java
jakelandis: will do.

logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
jakelandis: will do (and for other similar places)

logstash-core/src/main/java/org/logstash/instrument/witness/QueueWitness.java
jakelandis: Depends on what you mean by weird. 

I commented it out the line locally and this is the output:
```
 "queue" : { }
```
If we implement https://github.com/elastic/logstash/issues/7885 as part of this PR, it helps to eliminate many (most/all?) of these types of issues.

original-brownbear: @jakelandis ah right if that's the old behaviour that that's cool I guess :)

original-brownbear:
@jakelandis @andrewvc so I just tried this out and I get `500`s if I keep refreshing the metrics page during startup for the first few seconds (will open an issue for this now), but I think we can just ditch the `dirty` thing outright. The change is irrelevant because it only affects startup and startup is broken anyway (I mean if I have scripting that is thrown off by having a few 0s initialized there, then it will probably be thrown off by a `500 ` with a long error message as well would be my reasoning ...)?

logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
jakelandis: @original-brownbear - What are thoughts around optimizing non-hot code ?  I tend to leave the default value as-is since changing provides little value and it may send the wrong message to future developers.

original-brownbear: @jakelandis here it's not so much about throughput but more about saving memory. Default concurrency factor of `16` requires 16x as much memory compared to just putting a `1` for it (especially for cold path code). Hard to say how much it will matter here, but why pollute the heap when we don't have to :D

original-brownbear:
Link to `500` on startup mentioned in previous comment https://github.com/elastic/logstash/issues/8004

jakelandis:
@original-brownbear - The error on #8004 `LogStash::Instrument::MetricStore::MetricNotFound` will be resolved independently of the `dirty` flag in upcoming reviews. The MetricStore is going away in 6.1 https://github.com/elastic/logstash/issues/7971

original-brownbear:
@jakelandis sure, but doesn't that make the whole exercise with `dirty` just to avoid trivially changing the state of the metrics output on startup a lot less important since it's unlikely that anyone triggers anything (in their monitoring scripting and the like) on the simple fact that there is data coming from the HTTP endpoint?

logstash-core/src/main/java/org/logstash/instrument/witness/PipelinesWitness.java
jakelandis: I gave it a quick try, and ran into this error:
`"No ObjectCodec defined for the generator, can only serialize simple wrapper types (type passed java.util.concurrent.ConcurrentHashMap)"`

A bit of googling, and it was suggested to add ` gen.setCodec(new ObjectMapper());` ...however, the output is bit off (fixable), but seems odd to have to create an additional ObjectMapper since for the (asJson) workflows I am already explicitly creating one. 

It's only a couple extra lines of code, and would prefer to avoid mixing the explicit asJson serialization with the annonation driven approach. I have tests in this PR that confirm that they equal.

original-brownbear: @jakelandis huh that's strange and shouldn't happen, seems like we're doing something funky here that I don't see :) Let's leave it then we can clean it up some other time :)

jakelandis:
Spoke briefly with team on Slack and general acceptance on removing the `dirty` flag, assuming the difference is only the JSON output is zero/null vs. missing.

logstash-core/src/main/java/org/logstash/instrument/witness/ReloadWitness.java
jakelandis: Not sure I follow, these are static helper methods.

logstash-core/src/main/java/org/logstash/instrument/witness/EventsWitness.java
jakelandis: Yeah, i will make it a static constant. (as well the others)

logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
jakelandis: @original-brownbear - in order to tweak concurrency level, I also have to tweak the initial capacity and load factor.  Also, the javadoc states the default of 16 is "Unused but defined for compatibility with previous versions of this class".  

I am going to leave this as default, since this feels alot like pre-mature optimization.

jakelandis:
@original-brownbear / @andrewvc 
Review changes here: https://github.com/elastic/logstash/pull/7984/commits/904052dd507c5a33a1a2b6cb2ada6e4d5ed20643

I will implement https://github.com/elastic/logstash/issues/7885 on a seperate PR before the Serialization here is actually used. (as of now the Witness classes are dead code, un-connected to any production code)

jakelandis:
I have the `dirty` removal stage over at https://github.com/jakelandis/logstash/compare/7788...jakelandis:7885?expand=1 , and will submit new PR once this one is merged to master.

logstash-core/src/main/java/org/logstash/instrument/witness/ReloadWitness.java
original-brownbear: @jakelandis I meant caching the return of `MetricSerializer.Get.longSerializer(gen)` into a variable, since this is a stateless thing isn't it? To avoid the repeated call to `longSerializer()` and `timestampSerializer()`?

logstash-core/src/main/java/org/logstash/instrument/witness/PluginsWitness.java
original-brownbear: @jakelandis huh interesting they changed that in Java 8 without me noticing it :D just leave it then yea :)

original-brownbear:
@jakelandis btw LGTM in general, don't delay merging and fixing the `dirty` thing because of my comments :) Some redundant calls and such we can always fix in the next iteration, generally looks just fine :)

Issue #7990, opened 2017-08-14
Killing a PQ enabled Instance Creates Zombie Events that are never Acknowledged (and Breaks Acks for Future Events!)


original-brownbear:
This one seems to be breaking some reload integration tests randomly and is a bug either way.
It is fairly easy to create events in PQ that are never acknowledged and keep being replayed when starting up Logstash again.

Run I can reproduce this with reliably:

```sh
$ bin/logstash -e 'input {tcp {port => 7744 codec => json}} output{stdout {codec => line { format => "%{foo}"}}}'
```

Then sent a few events to it via:

```sh
 echo '{"foo": "bar1k23"}' | nc 127.0.0.1 7744
 echo '{"foo": "bar1k23"}' | nc 127.0.0.1 7744
 echo '{"foo": "bar1k23"}' | nc 127.0.0.1 7744
```

or similar :D

Then kill logstash by SIGINTing it twice to overall get this output:

```sh
➜  logstash git:(master) ✗ bin/logstash -e 'input {tcp {port => 7744 codec => json}} output{stdout {codec => line { format => "%{foo}"}}}'
WARNING: Default JAVA_OPTS will be overridden by the JAVA_OPTS defined in the environment. Environment JAVA_OPTS are -Xms4g -Xmx4g -Djava.net.preferIPv4Stack=true
Sending Logstash's logs to /Users/brownbear/src/logstash/logs which is now configured via log4j2.properties
[2017-08-14T14:47:15,143][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"fb_apache", :directory=>"/Users/brownbear/src/logstash/modules/fb_apache/configuration"}
[2017-08-14T14:47:15,157][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"netflow", :directory=>"/Users/brownbear/src/logstash/modules/netflow/configuration"}
[2017-08-14T14:47:15,212][INFO ][logstash.setting.writabledirectory] Creating directory {:setting=>"path.dead_letter_queue", :path=>"/Users/brownbear/src/logstash/data/dead_letter_queue"}
[2017-08-14T14:47:15,316][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified
[2017-08-14T14:47:16,125][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[...]

PR #7991, opened 2017-08-14
#7990 bring back batch Acking for PQ

original-brownbear:
This was accidentally removed in https://github.com/elastic/logstash/pull/7955/files#diff-bc1e9ea81caa54ea0d56abf4415fdf00L473 breaking the ack as described in #7990.

This fixes the missing ack.


jsvd:
I'm so amazed that no tests caught this :D

original-brownbear:
@jsvd thanks, we should add an issue to put actual tests on this :D will do :)

original-brownbear:
there we go https://github.com/elastic/logstash/issues/7992 :)

PR #7970, opened 2017-08-10
create source_loader after logging setup

jsvd:
Without this PR:

```
% bin/logstash -f "logstash_configs/*"
[2017-08-10T12:48:01,463][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified
[2017-08-10T12:48:02,760][INFO ][logstash.config.source.local.configpathloader] No config files found in path {:path=>"/Users/joaoduarte/projects/elastic/logstash/logstash_configs/*"}
[ERROR] 2017-08-10 12:48:02.781 [Ruby-0-Thread-1: /Users/joaoduarte/projects/elastic/logstash/lib/bootstrap/environment.rb:6] sourceloader - No configuration found in the configured sources.
```

With this PR:

```
% bin/logstash -f "logstash_configs/*"
[2017-08-10T12:48:55,008][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified
[2017-08-10T12:48:55,981][INFO ][logstash.config.source.local.configpathloader] No config files found in path {:path=>"/Users/joaoduarte/projects/elastic/logstash/logstash_configs/*"}
[2017-08-10T12:48:56,002][ERROR][logstash.config.sourceloader] No configuration found in the configured sources.
```

PR #7995, opened 2017-08-14
Revert "create source_loader after logging setup"

andrewvc:
This commit broke the plugin contract with plugins like xpack and caused many issues there. Reverted until it can be better handled. There was never a corresponding issue for what this fixed for me to link to, but the original PR was: https://github.com/elastic/logstash/pull/7970

This reverts commit 92cdbe2dbdf3d01e39540522508b679066f6dd93.


colinsurprenant:
revert LGTM but not tested.

PR #7996, opened 2017-08-15
BUG: Fix RubyTimestamp not being Serialized Correctly

original-brownbear:
We didn't have a serializer for `RubyTimestamp` which led to `RubyTimestamp` type fields (except for `@timestamp` itself) not being correctly serialized in some cases.
Added a serializer that is exactly the same as that of the Java type and a test to reproduce the issue (throws if you remove the serializer).


original-brownbear:
@guyboertje thanks! :) merging

Issue #7998, opened 2017-08-15
Investigate if Removing Jackson Afterburner has Performance Impact (and maybe remove it)


original-brownbear:
Using Jackson's Afterburner currently, forces us to register some classes (our own ones if they contain fields that Jackson can't natively handle, but we want to serialize using custom serializers) via annotations and others (JRuby classes that we can't put annotations on) via manually registering them with Jackson.

* Since we can't use annotations throughout, we should try to move everything to manual registration in `org.logstash.ObjectMappers#RUBY_SERIALIZERS` (and then rename that variable :D) to get some uniformity in Jackson use
* Removing Afterburner if it's no benefit allows us to get rid of a dependency

We should just add a JMH to see if there is any degradation in event serialization to CBOR and/or JSON and remove it if there isn't any.

original-brownbear:
@guyboertje FYI :)

original-brownbear:
Some quick testing with `QueueWriteBenchmark` firing at the in memory serialized queue shows no difference whatsoever between  `Afterburner` on and off. I'd drop it and move away from annotations.

guyboertje:
+1 to JMH in any case. We should have them for our growing corpus of Java code.

Remove Afterburner if results show no significant change.

original-brownbear:
This is the data I got so far:

### No Afterburner

```sh
# Benchmark: org.logstash.benchmark.EventSerializationBenchmark.serializeCbor

# Run progress: 0.00% complete, ETA 00:00:13
# Fork: 1 of 1
# Warmup Iteration   1: 212.679 ops/ms
# Warmup Iteration   2: 533.042 ops/ms
# Warmup Iteration   3: 565.571 ops/ms
Iteration   1: 541.071 ops/ms
Iteration   2: 568.086 ops/ms
Iteration   3: 553.683 ops/ms
Iteration   4: 562.362 ops/ms
Iteration   5: 571.284 ops/ms
Iteration   6: 561.293 ops/ms
Iteration   7: 563.529 ops/ms
Iteration   8: 575.415 ops/ms
Iteration   9: 576.788 ops/ms
Iteration  10: 566.734 ops/ms


Result "org.logstash.benchmark.EventSerializationBenchmark.serializeCbor":
  564.024 ±(99.9%) 16.043 ops/ms [Average]
  (min, avg, max) = (541.071, 564.024, 576.788), stdev = 10.612
  CI (99.9%): [547.981, 580.068] (assumes normal distribution)


[...]

PR #8002, opened 2017-08-16
#7998 Event Serialization Benchmark and Afterburner Removal

original-brownbear:
As discussed in #7998 

* `Afterburner` seems to not add anything performance wise + Afterburner adds potential instabilities + forces us to use a mix of annotations and explicit serializer registering => removing it is a reasonable cleanup
* Had to compile JRuby into the benchmark now to get the dependencies for setting a timestamp on the `Event`


original-brownbear:
Jenkins test this please

PR #8009, opened 2017-08-16
Metrics: Consistent serialization

### LINKED ISSUE #7885: Metrics: Consistent serialization

jakelandis:
This change allows for the serialized output of the metrics to be consistent w/r/t the un-initialized values. Note - the code presented here is not the current code in use, rather it is the code staged, but unused to replace the metrics store / namespace implementation.

Fixes #7885 (for a future release)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/instrument/witness/MetricSerializer.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/witness/MetricSerializer.java
@@ -36,7 +36,7 @@ public interface MetricSerializer<T extends Metric<?>> {
          */
         static MetricSerializer<Metric<Long>> longSerializer(JsonGenerator gen) {
             return m -> {
-                if (m != null && m.isDirty() && m.getValue() != null) {
+                if (m != null && m.getValue() != null) {
                     gen.writeNumberField(m.getName(), m.getValue());
                 }
             };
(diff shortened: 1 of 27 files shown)


original-brownbear:
test failures here are due to #8016

logstash-core/src/main/java/org/logstash/instrument/witness/MetricSerializer.java
```diff
@@ -36,7 +36,7 @@
          */
         static MetricSerializer<Metric<Long>> longSerializer(JsonGenerator gen) {
             return m -> {
-                if (m != null && m.isDirty() && m.getValue() != null) {
+                if (m != null && m.getValue() != null) {
```
jakelandis: I just noticed that by checking the null value here, we have are still conditionally showing some of the values. Any gauge value, not yet set or not initialized will still conditionally appear since the initial state is null. Counters will always show. 

Update in-coming.

jakelandis: ...actually since this is currently dead code (no paths to execute it), i will fix this on the next PR that will also include the PQ and DLQ stuff.  (My working copy already has those and it is making my manual merge difficult).

jakelandis:
I will rebase with the fix for the issue referenced, and wait for green.

jakelandis:
All green ... merging.

PR #8015, opened 2017-08-17
Performance: Remove BooleanBiValue

original-brownbear:
Exactly like #7854 and #7917 for #7712.

* Removed `BooleanBiValue`
* Added test for serializing boolean (we didn't have one, just removing `BooleanBiValue` without any fixes to `Valuefier` etc. worked!)
* Added custom serializer for `RubyBoolean`


logstash-core/src/main/java/org/logstash/ObjectMappers.java
```diff
@@ -72,4 +74,22 @@ public void serialize(final RubyFloat value, final JsonGenerator generator,
...
+        extends NonTypedScalarSerializerBase<RubyBoolean> {
+
+        RubyBooleanSerializer() {
+            super(RubyBoolean.class, true);
+        }
+
+        @Override
+        public void serialize(final RubyBoolean value, final JsonGenerator generator,
+            final SerializerProvider provider) throws IOException {
+            generator.writeBoolean(value.isTrue());
```
jakelandis: value.isTrue() ?  Just double checking.. that seems like an odd way to get a boolean value.

original-brownbear: @jakelandis weirdly enough this was the only method on that thing that had a clean `boolean` return on that class. `toJava` returns `Object` which is kind of dirty moving forward imo.

jakelandis:
LGTM

PR #8024, opened 2017-08-18
DLQ: Fix reading while writing boundary issues

robbavey:
Fixes an issue where if the DLQ is actively being read and written to, the consumeBlock method would write data to the current block byte buffer where the data was last read from, rather than last written to.

### DIFF
--- a/logstash-core/src/test/java/org/logstash/common/io/RecordIOWriterTest.java
+++ b/logstash-core/src/test/java/org/logstash/common/io/RecordIOWriterTest.java
@@ -99,6 +99,28 @@ public class RecordIOWriterTest {
         reader.close();
     }
 
+    @Test
+    public void testReadWhileWriteAcrossBoundary() throws Exception {
+        char[] tooBig = fillArray( BLOCK_SIZE/4);
+        StringElement input = new StringElement(new String(tooBig));
+        byte[] inputSerialized = input.serialize();
+        try(RecordIOWriter writer = new RecordIOWriter(file);
+            RecordIOReader reader = new RecordIOReader(file)){
+
+            for (int j = 0; j < 2; j++) {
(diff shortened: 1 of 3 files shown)


robbavey:
Note, that without this fix, the dead_letter_queue input would routinely crash after around 250 messages. With the fix, this will complete (and works consistently with a million generated messages)

```
input { 
  generator{ 
   message => '{"test":"one"}' 
   codec => "json" 
   count => 1000
  }
  dead_letter_queue {
   path => "/Users/robbavey/code/logstash/data/dead_letter_queue" 
   commit_offsets => true 
  } 
} 
filter {
  if ([geoip])
   { mutate {
     remove_field => ["geoip"] 
     add_field => { 
       "mutated" => "true" 
     } 
   } 
  }else{ 
    mutate {
     add_field => { "geoip" => "somewhere" } 
    }
  }
} 
output { elasticsearch {} }
```

logstash-core/src/test/java/org/logstash/common/io/RecordIOWriterTest.java
```diff
@@ -99,6 +99,28 @@ public void testReadWhileWrite() throws Exception {
         reader.close();
     }
 
+    @Test
```
talevy: would this test be better suited for `RecordIOReaderTest.java`?

robbavey: I colocated it with a very similar (`test testReadWhileWrite`), but yeah - I think you are right, given that it verifies the fixing of a reader bug, not a writer bug

robbavey:
Jenkins please test this

robbavey:
Build failure unrelated, tracked by #8028

PR #8026, opened 2017-08-18
RATS: Dead Letter Queue integration tests

### LINKED ISSUE #7882: Add RATS integration test for DLQ
Let's add an end-to-end RATS test which introduces some mapping error, dumps the events to DLQ and subsequently, a LS pipeline/instance can slurp the data using DLQ input. This would be huge to catch regressions and provide confidence about this feature.

robbavey:
Simple test for dead letter queue integration tests - generate 1000 documents that cannot be ingested into Elasticsearch, then use the dead letter queue to mutate the document, and allow it to be ingested successfully.

Includes tests for single and multi-pipeline configurations 

Fixes #7882

### DIFF
--- a/config/jvm.options
+++ b/config/jvm.options
@@ -52,9 +52,9 @@
 #-Djna.nosys=true
 
 # Turn on JRuby invokedynamic
-#-Djruby.compile.invokedynamic=true
+-Djruby.compile.invokedynamic=true
 # Force Compilation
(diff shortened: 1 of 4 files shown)


robbavey:
@suyograo It already proved its worth - it caught issues in the implementation and in the plugin, which I've been fixing as I go

PR #8027, opened 2017-08-18
Metrics / Bug Fix: Protect against race condition in metrics reporting with multiple inputs.

### LINKED ISSUE #8011: pipeline.events.in shows 0 when there are multiple inputs where one has 0 events flowing
Using latest pre-release build:

```
{
"host": "Firestorm",
"version": "5.6.0",
"http_address": "127.0.0.1:9600",
"id": "997d4528-f143-460a-ace4-838520713f3e",
"name": "Firestorm",
"build_date": "2017-08-10T02:08:35Z",
"build_sha": "e37e2d00dd80690f8bc91e0ae32a910dc64544f0",
"build_snapshot": false
}
```

pipeline.events.in works fine if there is only 1 beats input getting events.  But if there are 2 beats inputs, 1 is idle (with no events sent to it) and another one actively getting data, then the pipeline.events.in will show 0 instead of the running total based on the other active inputs.  If both beats input are receiving data, then the pipeline.events.in works also.

```
"pipeline": {
"events": {
"duration_in_millis": 162776,
"in": 0,  <-------
"out": 40023,
"filtered": 40946,
"queue_push_duration_in_millis": 18658
},
"plugins": {
"inputs": [
{
"id": "07f51f147a0e39bd6512251acc3581bd57190a8f-2",
"events": {
"out": 0,
"queue_push_duration_in_millis": 0
},
"name": "beats"
},
{
"id": "07f51f147a0e39bd6512251acc3581bd57190a8f-1",
"events": {
"out": 40946,
"queue_push_duration_in_millis": 18658
},
"current_connections": 1,
"name": "beats",
"peak_connections": 1
}
],
```

The config is simply 2 inputs writing to stdout:

```
input {
  beats {
    port=>5044
  }
    beats {
    port=>5045
  }
}
```

@jakelandis

jakelandis:
Protect initialization of metrics objects with static mutex to prevent race condition. See https://github.com/elastic/logstash/issues/8011 for additional information.

Fixes #8011

### DIFF
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -34,6 +34,7 @@ module LogStash; class BasePipeline
 
   def initialize(config_str, settings = SETTINGS)
     @logger = self.logger
+    @mutex = Mutex.new
     @config_str = config_str
     @config_hash = Digest::SHA1.hexdigest(@config_str)
     # Every time #plugin is invoked this is incremented to give each plugin
@@ -672,6 +673,9 @@ module LogStash; class Pipeline < BasePipeline
   end
 
   def wrapped_write_client(plugin)
-    LogStash::Instrument::WrappedWriteClient.new(@input_queue_client, self, metric, plugin)
+    #need to ensure that metrics are initialized one plugin at a time, else a race condition can exist.
+    @mutex.synchronize do
+      LogStash::Instrument::WrappedWriteClient.new(@input_queue_client, self, metric, plugin)
+    end
   end
 end; end


jakelandis:
Travis failures may be related .... don't merge till Travis is green.

colinsurprenant:
Did you think about just synchronizing the call to `WrappedWriteClient.new` in pipeline https://github.com/elastic/logstash/blob/32b16bfaa4233204f4014f9a3d8101a9ace6c766/logstash-core/lib/logstash/pipeline.rb#L811-L813

with something like this?

```ruby
  def wrapped_write_client(plugin)
    @mutex.synchronize do
      LogStash::Instrument::WrappedWriteClient.new(@input_queue_client, self, metric, plugin)
    end
  end
```

WDYT?

jakelandis:
@colinsurprenant - that makes more sense. I will update once I figure out the build failure.

jakelandis:
Test failure un-related: https://github.com/elastic/logstash/pull/8036 (will rebase once that is merged)

jakelandis:
@colinsurprenant - change made. However, I kept the the mutex at static scope (as opposed to instance scope) to ensure that the mutex is not tied directly the lifecycle of a plugin.

jakelandis:
rebased with test fix.

colinsurprenant:
I'd suggest using a simple iVar and I would initialize it in the pipeline constructor to make sure there is no race condition with the `||=` assignment.

jakelandis:
Is `||=` not atomic ?

What is an iVar ? (an integer?)

> in the pipeline constructor
Trying to ensure that the mutex of input initialization is not directly tied to the lifecycle of a single pipeline. This may be a non-issue today, but defensive against if input plugin instances are ever shared across multiple pipelines.

colinsurprenant:
Good point for multiple pipelines and not, `||=` is not atomic.

jakelandis:
@colinsurprenant - Let's try this again :) 

This time - a simple instance level mutex. I was being overly defensive, and the next iteration of Metrics won't have this issue.

Also, thanks for the tip on `||=`'s  lack of atomic-ness.

logstash-core/lib/logstash/pipeline.rb
```diff
@@ -34,6 +34,7 @@ module LogStash; class BasePipeline
 
   def initialize(config_str, settings = SETTINGS)
     @logger = self.logger
+    @mutex = Mutex.new
```
colinsurprenant: suggestion: we could use a more descriptive identifier? maybe something like `@write_client_mutex` ?

colinsurprenant:
just a minor comment
LGTM

logstash-core/lib/logstash/pipeline.rb
jakelandis: I didn't want to imply that the mutex was only to be used for the write client. I think it can be general purpose. I was actually a bit surprised that one didn't already exist for this class, had it, i would have just used that.

colinsurprenant: hmmm ... unless I am misunderstanding I'd really advise against using a "general purpose" mutex ... a mutex should have a single purpose IMO ... if you start mixing thread safety concerns by "sharing" a single mutex it will hard to understand and hard de debug.

jakelandis:
need to forward port to 6.x

jakelandis:
Closing: Merged to 5.6.  
Forward port PR : https://github.com/elastic/logstash/pull/8047

PR #8053, opened 2017-08-22
Metrics: Add custom witness for plugins, and code clean up

jakelandis:
Remove LongGauge and DoubleGauge and Replace with NumberGauge
 - This is an unecessary distiction for Gauges and can complicate the serialization. Double/Float should serialize with decimal precision, and Jackson handles this by default.

Add custom witness to match existing Ruby API, such that existing Ruby plugins should not need changes

Remove namespace from LazyDelegatingGauge
 - It was only there for logging, but newer versions (not here) will completely remove the concept of namespace.

Part of #7788


logstash-core/src/main/java/org/logstash/instrument/witness/PluginWitness.java
```diff
@@ -16,9 +25,12 @@
 public class PluginWitness implements SerializableWitness {
 
     private final EventsWitness eventsWitness;
+    private final CustomWitness customWitness;
     private final TextGauge id;
     private final TextGauge name;
     private final Snitch snitch;
+
```
andrewvc: Accidental extra whitespace?

jakelandis:
I missed file in initial merge the merge that broke the Ruby integration.  Since I removed the namespace from LazyDelegatingGauge, I also need to ensure that the Ruby code doesn't try to call it. 

I added this to the PR and squashed it.

```
diff --git a/logstash-core/lib/logstash/instrument/metric_type/gauge.rb b/logstash-core/lib/logstash/instrument/metric_type/gauge.rb
index f78080537..e6492305e 100644
--- a/logstash-core/lib/logstash/instrument/metric_type/gauge.rb
+++ b/logstash-core/lib/logstash/instrument/metric_type/gauge.rb
@@ -4,7 +4,7 @@ module LogStash module Instrument module MetricType
   class Gauge < LazyDelegatingGauge
 
     def initialize(namespaces, key)
-      super(namespaces, key.to_s)
+      super(key.to_s)
     end
 
     def execute(action, value = nil)
```

andrewvc:
LGTM on that change Jake!

andrewvc:
I respun one test which failed due to https://github.com/elastic/logstash/issues/8057

jakelandis:
All green now ... merging.

PR #8056, opened 2017-08-23
Fix path for local development (attempt 2)

jakelandis:
With IntelliJ 2017.02 the gradle output changed from 'build/classes' to 'out/production/classes'. See https://youtrack.jetbrains.com/issue/IDEA-175172 for additional information.

This PR changes the local developement behavior to prefer the classes build with IntelliJ (if they exist) over the classes built directly with gradle.


jordansissel:
I have no opinion here, but the change looks OK to me.

Issue #8041, opened 2017-08-21
Installing a plugin always causes jar-dependencies to try to access maven


jsvd:
When online on Ubuntu or CentOS, installing a plugin causes jar-dependencies to try to access the local ~/.m2 configuration, which may not exist, causing an error:

```
osboxes@osboxes:~$ sudo -u logstash /usr/share/logstash/bin/logstash-plugin install logstash-filter-truncate
Validating logstash-filter-truncate
Installing logstash-filter-truncate
Error Bundler::InstallError, retrying 1/10
An error occurred while installing logstash-core (5.5.2), and Bundler cannot continue.
Make sure that `gem install logstash-core -v '5.5.2'` succeeds before bundling.
```

Since the plugins that use jars now ship with them inside the gem, we can instruct jar-dependencies to leave the .m2 folder alone using JARS_SKIP=true.

jakelandis:
Reverted initial fix : https://github.com/elastic/logstash/pull/8070

It caused rake artifact:tar to not bundle the .jar's for core.

ppf2:
+1 Previously no firewall rules are necessary to the maven repo to install plugins like kafka, but now it attempts to pull jars from maven.  It will be nice to see a fix to this or let's document all the external sites the plugin manager will have to access (or the workaround noted here), etc.. @andrewvc

kares:
for reference, JarDependencies are more lazy on the post install hook - no embedded mvn is started unless any jar requirements are detected. thus the issue should no longer be relevant in LS 6.x/7.x

PR #8058, opened 2017-08-23
always use JARS_SKIP when installing or updating plugins

jsvd:
fixes https://github.com/elastic/logstash/issues/8041


jsvd:
Thanks @ph :)

jakelandis:
reverted this change on https://github.com/elastic/logstash/pull/8070

It caused rake artifact:tar to stop assembling the .jar files for core. (but still worked if you happen to have a .m2 repo locally, like there is on the developer and build machines due to the build process)

PR #8059, opened 2017-08-23
Modules cloud id auth

guyboertje:


guyboertje:
What branches to port to?

suyograo:
Let's backport to 5.6 as well

logstash-core/lib/logstash/modules/settings_merger.rb
```diff
@@ -20,4 +20,25 @@ def self.merge(cli_settings, yml_settings)
...
+    cloud_id = logstash_settings.get("cloud.id")
+    cloud_auth = logstash_settings.get("cloud.auth")
+    if cloud_id.nil?
+      if cloud_auth.nil?
+        return # user did not specify cloud settings
+      else
+        raise ArgumentError.new("Cloud Auth without Cloud Id")
+      end
+    end
+    module_settings["var.kibana.scheme"] = "https"
```
suyograo: Can we log here as info that we are overwriting because of a presence of cloud.id?

logstash-core/lib/logstash/util/cloud_auth.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    unless value.is_a?(String)
+      raise ArgumentError.new("Cloud Auth must be String. Received: #{value.class}")
+    end
+    @original = value
+    @username, sep, password = @original.partition(":")
+    if @username.empty? || sep.empty? || password.empty?
+      raise ArgumentError.new("Cloud Auth username and password format should be \"<username>:<password>\".")
+    end
+    @password = LogStash::Util::Password.new(password)
```
suyograo: Nice

logstash-core/lib/logstash/util/cloud_id.rb
```diff
@@ -0,0 +1,29 @@
...
+      @decoded = Base64.urlsafe_decode64(@label) rescue ""
+      @label = ""
+    else
+      @decoded = Base64.urlsafe_decode64(last) rescue ""
+    end
+    unless @decoded.count("$") == 2
+      raise ArgumentError.new("Cloud Id does not decode. Received: \"#{@original}\".")
+    end
+    cloud_host, es_server, kb_server = @decoded.split("$")
+    @elasticsearch_host = sprintf("%s.%s:443", es_server, cloud_host)
```
suyograo: Can we log this as debug?

logstash-core/lib/logstash/util/cloud_id.rb
```diff
@@ -0,0 +1,29 @@
...
+    @kibana_host  = sprintf("%s.%s:443", kb_server, cloud_host)
```
suyograo: Can we log this as debug?

config/logstash.yml
```diff
@@ -109,6 +109,15 @@
 #
 # modules:
 #
+# ------------ Cloud Settings ---------------
+# Define Elastic Cloud settings here. id and auth need to be defined together
```
guyboertje: Will change

logstash-core/lib/logstash/settings.rb
```diff
@@ -557,6 +560,54 @@ def coerce(value)
...
+        end
+        @klass.new(value)
+      end
+
+      protected
+      def validate(value)
+        coerce(value)
+      end
+    end
   end
```
ph: The code from _CloudId_ and _CloudAuth_ is mostly the same? Maybe it's possible to merge them? It might DRY too far though.

logstash-core/lib/logstash/util/cloud_auth.rb
ph: good reuse!

logstash-core/lib/logstash/modules/settings_merger.rb
```diff
@@ -20,4 +20,25 @@ def self.merge(cli_settings, yml_settings)
     end
     merged
   end
+
+  def self.merge_cloud_settings(module_settings, logstash_settings)
```
ph: Can we have a unit test that test the merging strategies, to make sure we have the correct priorities.

logstash-core/lib/logstash/settings.rb
guyboertje: I can!!
I'll call the merged class CloudCredential and pass in the appropriate class

logstash-core/lib/logstash/util/cloud_id.rb
guyboertje: These are utility classes for validation and the actual merge - I would rather log in the `LogStash::Modules::SettingsMerger.merge_cloud_settings` method once.

guyboertje:
I need to add `to_s` to CloudId and CloudAuth because the debug print of Settings is ugly.

logstash-core/lib/logstash/util/cloud_id.rb
suyograo: ok

logstash-core/lib/logstash/settings.rb
guyboertje: Changed to Modules to use it on all module settings.

logstash-core/lib/logstash/util/modules_setting_array.rb
```diff
@@ -0,0 +1,28 @@
...
+    @original = value
+    # wrap passwords
+    @original.each do |hash|
+      hash.keys.select{|key| key.to_s.end_with?('password')}.each do |key|
+        hash[key] = LogStash::Util::Password.new(hash[key])
+      end
+    end
+  end
+
+  def __class__
```
jordansissel: What's this method for? `Object#class` already exists for this purpose, I think?

config/logstash.yml
```diff
@@ -109,6 +109,19 @@
 #
 # modules:
 #
+# ------------ Cloud Settings ---------------
+# Define Elastic Cloud settings here.
+# Format of cloud.id is a base64 number e.g. dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRub3RhcmVhbCRpZGVudGlmaWVy
```
guyboertje: true

jordansissel:
I briefly search the cloud, dev, and logstash repos, and my email, for "cloud id" and I wasn't able to find much, so I am unable to verify if the text values for cloud id and cloud auth are correctly following whatever Cloud is going to provide.

logstash-core/lib/logstash/elasticsearch_client.rb
```diff
@@ -39,8 +39,11 @@ def initialize(settings, logger)
       @client_args[:ssl] = ssl_options
 
       username = @settings["var.elasticsearch.username"]
-      password = @settings["var.elasticsearch.password"]
       if username
+        password = @settings["var.elasticsearch.password"]
+        if password.is_a?(LogStash::Util::Password)
```
guyboertje: when its nil maybe or the elasticsearch_client is reused and someone does not wrap it?

logstash-core/lib/logstash/util/cloud_setting_auth.rb
```diff
@@ -0,0 +1,29 @@
...
+    @original = value
+    @username, sep, password = @original.partition(":")
+    if @username.empty? || sep.empty? || password.empty?
+      raise ArgumentError.new("Cloud Auth username and password format should be \"<username>:<password>\".")
+    end
+    @password = LogStash::Util::Password.new(password)
+  end
+
+  def to_s
+    "#{@original}:#{@password}"
```
guyboertje: yes doh

jordansissel:
Also, I'm a little confused. Logstash allows multiple Elasticsearch outputs  so a single global credential for a single cloud instance seems weird.

Can you help me understand this? Is this intended only for modules? How would a user of the Elasticsearch output use this?

logstash-core/lib/logstash/util/modules_setting_array.rb
guyboertje: class is delegated by Forwardable to `@original`. this method is used in a test to verify that the class quacks like an array but one can test to see that it is not one.

jordansissel:
Also please include user-facing documentation for this.

logstash-core/lib/logstash/modules/settings_merger.rb
```diff
@@ -11,13 +16,56 @@ def self.merge(cli_settings, yml_settings)
...
+    module_settings["var.kibana.host"] = cloud_id.kibana_host
+    module_settings["var.elasticsearch.hosts"] = cloud_id.elasticsearch_host
+    unless cloud_auth.nil?
+      module_settings["var.elasticsearch.username"] = cloud_auth.username
+      module_settings["var.elasticsearch.password"] = cloud_auth.password
+      module_settings["var.kibana.username"] = cloud_auth.username
+      module_settings["var.kibana.password"] = cloud_auth.password
+    end
+    if logger.debug?
+      format_module_settings(settings_copy, module_settings).each {|line| logger.debug(line)}
```
guyboertje: I can update the test `settings_merger_spec`

guyboertje:
@jordansissel Rationale https://github.com/elastic/cloud/issues/7412#issuecomment-322900708

logstash-core/lib/logstash/util/cloud_setting_id.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize(value)
+    return if value.nil?
+
+    unless value.is_a?(String)
+      raise ArgumentError.new("Cloud Id must be String. Received: #{value.class}")
+    end
+    @original = value
+    @label, sep, last = value.partition(":")
+    if last.empty?
+      @decoded = Base64.urlsafe_decode64(@label) rescue ""
```
jordansissel: We should not ignore the exception, here. If a user makes a mistake, it'd be nice to surface the problem (invalid base64, in this case)

logstash-core/lib/logstash/util/cloud_setting_id.rb
```diff
@@ -0,0 +1,37 @@
...
+      @label = ""
+    else
+      @decoded = Base64.urlsafe_decode64(last) rescue ""
+    end
+    unless @decoded.count("$") == 2
+      raise ArgumentError.new("Cloud Id does not decode. Received: \"#{@original}\".")
+    end
+    cloud_host, es_server, kb_server = @decoded.split("$")
```
jordansissel: Let's verify that we have all required parts here and notify the user (via actionable log + exception) that we have an invalid cloud id.

logstash-core/lib/logstash/elasticsearch_client.rb
ph: What we should actually do is this 

`LogStash::Utils::Password(password)` and this would always return a Password object, this is similar to other rubyism.


```
Array([1, 2, 3]) => [1,2,3]
Array(1) => [1]
```

I _don't think_ we can get a NIL password an empty string yes but NIL no.

ph: by doing that, we make sure that our lib wont bleed anything, if we bleed password it will be outside.

guyboertje:
@jordansissel 
> Also, I'm a little confused. Logstash allows multiple Elasticsearch outputs so a single global credential for a single cloud instance seems weird.

The single host for ES and Kibana and the username and password is intended more for the module ES and Kibana resource loading during setup.

> Can you help me understand this? Is this intended only for modules? How would a user of the Elasticsearch output use this?

Modules only at this time with this PR. With modules we build the elasticsearch output content programatically to embed into the LS config ERB.

logstash-core/lib/logstash/util/cloud_setting_auth.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def inspect
+    to_s
+  end
+end end end
```
ph: I am not sure why this class is in the Util namespace?
It feel to me it's more appropriate to either be in the _Modules_ or the _Settings_ namespace?

logstash-core/lib/logstash/util/cloud_setting_id.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def to_s
+    @original.to_s
+  end
+
+  def inspect
+    to_s
+  end
+end end end
```
ph: I am not sure why this class is in the Util namespace?
It feel to me it's more appropriate to either be in the Modules or the Settings namespace?

logstash-core/lib/logstash/util/cloud_setting_id.rb
guyboertje: Surely this effectively converts the rescue into a ArgumentError with a predictable message?
```
    unless @decoded.count("$") == 2
      raise ArgumentError.new("Cloud Id does not decode. Received: \"#{@original}\".")
    end
```
An empty string does not through an error in `Base64.urlsafe_decode64` and...
```
Base64.urlsafe_decode64("-------------------")
ArgumentError: invalid base64
	from org/jruby/RubyString.java:5345:in `unpack'
```
is not a great error message. We would have to rescue and re-raise with a better message.
The error would be seen in the logs/console at startup - the user would not know where the problem is.

I imagined that the user would be given the base64 from some Kibana UI so it would be a copy paste problem.
I also thought that in Cloud for a module the Cloud LS instance would have a correctly generated cmd line or a correct logstash.yml.

logstash-core/locales/en.yml
```diff
@@ -246,6 +246,12 @@ en:
           Load index template into Elasticsearch, and saved searches, 
           index-pattern, visualizations, and dashboards into Kibana when
           running modules.
+        cloud_id: |+
+          Sets the elasticsearch and kibana host settings
+          for module connections in Elastic Cloud
+        cloud_auth: |+
+          Sets the elasticsearch and kibana username and password
+          for module connections in Elastic Cloud
```
ph: Can we show the format? We said it in the yaml.

``` yaml
cloud.auth: elastic:<password>
```

logstash-core/lib/logstash/util/cloud_setting_auth.rb
guyboertje: because its a utility class? While these two files are currently used for modules I felt that they are a kind of validating value object used by settings (validating) and modules (value object)

ph: I am still mixed, maybe my definition of Utils is something I would take out of the Utils and be able to use in another project directly?

They can stay where they are. Your class `CloudSettingAuth` is more specific than a String.

 I've looked at the content of the settings.rb and we will need to split the content of it.

logstash-core/lib/logstash/util/cloud_setting_id.rb
ph: ignore this ^

ph:
@guyboertje From my side the only remaining things are:

- Update the clauth auth message in the CLI
- Harden the Password usage with that [comment](https://github.com/elastic/logstash/pull/8059#discussion_r135087417)?

logstash-core/lib/logstash/util/cloud_setting_id.rb
guyboertje: The only validation I can do is test for multiple dot separated fields. I can't assume there is any known patterns in the pieces.

logstash-core/lib/logstash/util/cloud_setting_id.rb
```diff
@@ -0,0 +1,37 @@
...
+      @decoded = Base64.urlsafe_decode64(@label) rescue ""
+      @label = ""
+    else
+      @decoded = Base64.urlsafe_decode64(last) rescue ""
+    end
+    unless @decoded.count("$") == 2
+      raise ArgumentError.new("Cloud Id does not decode. Received: \"#{@original}\".")
+    end
+    cloud_host, es_server, kb_server = @decoded.split("$")
+    @elasticsearch_host = sprintf("%s.%s:443", es_server, cloud_host)
```
jordansissel: Shouldn't this be :9243? My comment is in reference to the cloud ticket.

logstash-core/lib/logstash/util/cloud_setting_id.rb
```diff
@@ -0,0 +1,37 @@
...
+    @kibana_host  = sprintf("%s.%s:443", kb_server, cloud_host)
```
jordansissel: Shouldn't this be :9243?

logstash-core/lib/logstash/util/cloud_setting_id.rb
guyboertje: When validation occurs (v early at startup) I not sure we have logging set up yet.

logstash-core/lib/logstash/util/cloud_setting_id.rb
guyboertje: Beats code used 443.

jordansissel: https://github.com/elastic/beats/pull/4964/files#diff-1b409b9020aa8694f3beaeec8ee7ff72R113 uses :443, so I think this is ok.

logstash-core/lib/logstash/util/cloud_setting_id.rb
jordansissel: https://github.com/elastic/beats/pull/4964/files#diff-1b409b9020aa8694f3beaeec8ee7ff72R113 uses :443, so I think this is ok.

logstash-core/lib/logstash/util/cloud_setting_auth.rb
guyboertje: I don't know if we will use cloud.id and cloud.auth in an ES output?

logstash-core/lib/logstash/elasticsearch_client.rb
guyboertje: when `@settings` gets to elasticsearch_client its just a hash. Anything could have happened to it - its mutable. The es_client is not exclusively for module use.

ph: Roger that.

guyboertje:
Updated code and docs.

logstash-core/lib/logstash/util/cloud_setting_auth.rb
ph: in the future we could.

docs/static/modules.asciidoc
```diff
@@ -118,4 +118,64 @@ Settings that you specify at the command line are merged with any settings
...
 places, the value specified at the command line takes precedence. 
 
+[[connecting-to-cloud]]
+=== Using Elastic Cloud
+
+Logstash comes with two settings that simplify using modules with https://cloud.elastic.co/[Elastic Cloud].
+The Elasticsearch and Kibana hostnames in Elastic Cloud may be hard to set
+in the Logstash config or on the commandline, so a Cloud Id can be used instead.
+
+==== Cloud Id
```
suyograo: Capitalize ID?

config/logstash.yml
```diff
@@ -109,6 +109,19 @@
 #
 # modules:
 #
+# ------------ Cloud Settings ---------------
+# Define Elastic Cloud settings here.
+# Format of cloud.id is a base64 value e.g. dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRub3RhcmVhbCRpZGVudGlmaWVy
+# and it may have an label prefix e.g. staging:dXMtZ...
+# This will overwrite 'var.elasticsearch.hosts' and 'var.kibana.host'
```
jordansissel: "will overwrite" -> "can be used instead of"

logstash-core/lib/logstash/environment.rb
```diff
@@ -20,8 +23,10 @@ module Environment
     Setting::NullableString.new("path.config", nil, false),
  Setting::WritableDirectory.new("path.data", ::File.join(LogStash::Environment::LOGSTASH_HOME, "data")),
     Setting::NullableString.new("config.string", nil, false),
-                    Setting.new("modules.cli", Array, []),
-                    Setting.new("modules", Array, []),
+           Setting::Modules.new("modules.cli", LogStash::Util::ModulesSettingArray, []),
+           Setting::Modules.new("modules", LogStash::Util::ModulesSettingArray, []),
+           Setting::Modules.new("cloud.id", LogStash::Util::CloudSettingId),
```
jordansissel: I'm thinking we should be calling this modules.cloud.id instead since it is very specific and only for modules.

suyograo:
Backported to 6.x and 6.0

Issue #7945, opened 2017-08-08
Performance of stdin -> stdout Baseline Degrades the more Worker Threads are Added


original-brownbear:
I found this when looking into #7935 with @jakelandis and it looks like we have some more room for improvement here.
Currently for a config like 

```
input { stdin{}} output {stdout {codec => "dots"}}'
```

you can clearly observe a decrease in throughput the more workers you add. Eventually, you simply run into the `if` part of the synchronous queue read here, which is very slow:

```rb
      def read_next
        read_size = @queue.queue.drainTo(@originals, @size)
        if read_size < @size
          (@size - read_size).times do |_|
            e = @queue.poll(@wait)
            return if e.nil?
            @originals.add(e)
          end
        end
      end
```

Just a suggestion:
We could probably way improve this behavior if we enqueued as full as possible full batches on the writer side (instead of individual elements) and simply pulled those off the Queue one by one. 
This means that the `@wait` would move to the writer side and we'd only have a `take()` on the reader side that returns a batch of events (could be an array).

PR #8081, opened 2017-08-26
MINOR: Faster batch clocks

original-brownbear:
* No point in constantly rehashing the hash that holds the same `Thread` as hash-key over time
  * Just set `nil` as a "reset"
* No need to lock this operation across threads
  * Use a concurrent hash map instead of synchronizing on a normal `RubyHash`
=> Multithreaded baseline get a little better (can post benchmarks some other time, it's kind of obvious how this is faster and makes scaling to more threads better imo)

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -103,7 +103,7 @@ module LogStash; module Util
         # batches will fail if we have multiple read clients in the pipeline.
         @inflight_batches = {}
         # allow the worker thread to report the execution time of the filter + output
-        @inflight_clocks = {}
+        @inflight_clocks = Concurrent::Map.new
         @batch_size = batch_size
         @wait_for = wait_for
       end
@@ -180,12 +180,11 @@ module LogStash; module Util
       def start_metrics(batch)
         @mutex.lock
         begin
-          # there seems to be concurrency issues with metrics, keep it in the mutex
           set_current_thread_inflight_batch(batch)
-          start_clock
         ensure
           @mutex.unlock
         end
+        start_clock
       end
 
       def set_current_thread_inflight_batch(batch)
@@ -196,12 +195,11 @@ module LogStash; module Util
         @mutex.lock
         begin
           batch.close
-          # there seems to be concurrency issues with metrics, keep it in the mutex
           @inflight_batches.delete(Thread.current)
-          stop_clock(batch)
         ensure
           @mutex.unlock
         end
+        stop_clock(batch)
       end
 
       def start_clock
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -182,10 +182,10 @@ def start_metrics(batch)
         begin
           # there seems to be concurrency issues with metrics, keep it in the mutex
```
jakelandis: Should this comment be removed ?

original-brownbear: @jakelandis yea :) killing it sec

original-brownbear:
link https://github.com/elastic/logstash/issues/7945 which is helped by this to a minor degree

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -196,29 +195,28 @@ def close_batch(batch)
...
       end
 
       def start_clock
         @inflight_clocks[Thread.current] = java.lang.System.current_time_millis
       end
 
       def stop_clock(batch)
-        unless @inflight_clocks[Thread.current].nil?
+        start_time = @inflight_clocks.get_and_set(Thread.current, nil)
+        unless start_time.nil?
```
jordansissel: shouldn't it be an error of some kind that stop_clock is called without a prior start_clock?

original-brownbear: @jordansissel yea, actually it looks like this never happens (I can['t see a code path that would cause this in any way).
I'll investigate this soon :) Looks like a freebie performance improvement if this is actually never `nil` :) But I'd not mix that change in here right?

suyograo:
Release labels please

original-brownbear:
@jakelandis @jordansissel @ph not sure who's reviewing this, but anything left to do here? :)

jakelandis:
@original-brownbear - I took a look, but didn't feel comfortable since i am not terribly familiar with ruby concurrency objects, and the [doc isn't very helpful](http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/Map.html#get_and_set-instance_method)

original-brownbear:
@jakelandis np, though I'm only using `Concurrent::Map.get_and_set` from that lib?

I mean I don't know much about it either, but it says its atomic, and it worked without NPEs in my testing => seems they kept their promise :D

original-brownbear:
@jordansissel ping I think this one is for you :)

original-brownbear:
@jordansissel ping :)

original-brownbear:
@jordansissel any news here? It's been 2 weeks now :)

original-brownbear:
back to the queue :)

jordansissel:
Sorry, I didn't have this on my todo list. LGTM.

original-brownbear:
np :)

PR #8083, opened 2017-08-26
PERFORMANCE: Move Memory Queue Drain Loop to Java Fully

original-brownbear:
* Moved the whole memory queue `read_next` logic to Java
  * And finally defined it in a reasonable manner in the Javadoc ... our polling behaviour with the constant reset of the timeout is really strange when you spell it out isn't it?
* Moved instantiating the `@originals` collection to the batch's constructor (for some strange reason this seems to save memory ... looks like a potential "bugish" thing in JRuby ... regardless it's the cleaner approach though that's how I got there <= no temporal coupling)
   * See how I adjusted the empty constructor accordingly
* Rebuild the `jar` dependency file (I know it doesn't belong here, but I'd just leave it now since I forgot it when I last touched this and tried implementing this with Guava first => rebuild 2x <= Guava wasn't optimal for our use case)

PS: Intentionally putting this into the general review queue, it's a straightforward port from Ruby to Java, I think everyone should be able to look over this?


original-brownbear:
Jenkins test this please

logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param count Number of {@link JrubyEventExtLibrary.RubyEvent}s to drain from
+     * {@link BlockingQueue}
+     * @param nanos Timeout in Nanoseconds
+     * @return Collection of {@link JrubyEventExtLibrary.RubyEvent} drained from
+     * {@link BlockingQueue}
+     * @throws InterruptedException On Interrupt during {@link BlockingQueue#poll()} or
+     * {@link BlockingQueue#drainTo(Collection)}
+     */
+    public static Collection<JrubyEventExtLibrary.RubyEvent> drain(
+        final BlockingQueue<JrubyEventExtLibrary.RubyEvent> queue, final int count, final long nanos
```
ph: timeout would probably be a better name than nanos?

logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
```diff
@@ -0,0 +1,83 @@
...
+     * from
+     * @param count Number of {@link JrubyEventExtLibrary.RubyEvent}s to drain from
+     * {@link BlockingQueue}
+     * @param nanos Timeout in Nanoseconds
+     * @return Collection of {@link JrubyEventExtLibrary.RubyEvent} drained from
+     * {@link BlockingQueue}
+     * @throws InterruptedException On Interrupt during {@link BlockingQueue#poll()} or
+     * {@link BlockingQueue#drainTo(Collection)}
+     */
+    private static int drain(final BlockingQueue<JrubyEventExtLibrary.RubyEvent> queue,
```
ph: timeout would probably be a better name than nanos?

logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
```diff
@@ -0,0 +1,83 @@
...
+     * {@link BlockingQueue}
+     * @throws InterruptedException On Interrupt during {@link BlockingQueue#poll()} or
+     * {@link BlockingQueue#drainTo(Collection)}
+     */
+    public static Collection<JrubyEventExtLibrary.RubyEvent> drain(
+        final BlockingQueue<JrubyEventExtLibrary.RubyEvent> queue, final int count, final long nanos
+    ) throws InterruptedException {
+        int left = count;
+        final Collection<JrubyEventExtLibrary.RubyEvent> collection =
+            new HashSet<>(4 * count / 3 + 1);
```
ph: So you are giving a ~35% buffer in the planned capacity, I was trying to understand where the 3 was coming from.

original-brownbear: @ph no, the load factor of the default `HashSet` is `0.75` => `count/0.75` is the size you need to never see resizing when inserting `count` elements into the `HashSet`.
`+1` because of `int` division always rounding down :)

logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
original-brownbear: @ph my reasoning here was that using `ns` as a timeout explicitly is a pretty nonstandard thing in Java.
So I thought I'd do what JDK does in a few spots and simply name it explicitly.

e.g. this in `ArrayBlockingQueue`

```java
    public boolean offer(E e, long timeout, TimeUnit unit)
        throws InterruptedException {

        checkNotNull(e);
        long nanos = unit.toNanos(timeout);
```

Saves you from having to read the JavaDoc imo :)

PR #8084, opened 2017-08-28
MINOR: Cleanup magic string for LS JRuby Module name

original-brownbear:
* Cleaning up that magic `String`
  * Making it easier to debug load issues, since we can more easily find all the places that reference our JRuby module explicitly :)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ext/JrubyEventExtLibrary.java
+++ b/logstash-core/src/main/java/org/logstash/ext/JrubyEventExtLibrary.java
@@ -23,6 +23,7 @@ import org.logstash.ConvertedMap;
 import org.logstash.Event;
 import org.logstash.FieldReference;
 import org.logstash.PathCache;
+import org.logstash.RubyUtil;
 import org.logstash.Rubyfier;
 import org.logstash.Valuefier;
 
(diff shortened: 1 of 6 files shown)


logstash-core/src/main/java/org/logstash/ext/JrubyEventExtLibrary.java
```diff
@@ -34,7 +35,7 @@
 
     @Override
     public void load(Ruby runtime, boolean wrap) throws IOException {
-        RubyModule module = runtime.defineModule("LogStash");
+        final RubyModule module = runtime.defineModule(RubyUtil.LS_MODULE_NAME);
```
jakelandis: should this be defineOrGetModule ?  (I don't see anything preventing calling this multiple times)

original-brownbear: @jakelandis yea that would be cleaner in terms of readability, but functionally we're good I think. See related JRuby code:

```java
    public RubyModule defineModule(String name) {
        return defineModuleUnder(name, objectClass);
    }

    /**
     * Define a new module with the given name under the given module or
     * class namespace. Roughly equivalent to rb_define_module_under in MRI.
     *
     * @param name The name of the new module
     * @param parent The class or module namespace under which to define the
     * module
     * @returns The new module
     */
    @Extension
    public RubyModule defineModuleUnder(String name, RubyModule parent) {
        IRubyObject moduleObj = parent.getConstantAt(name);

        boolean parentIsObject = parent == objectClass;

        if (moduleObj != null ) {
            if (moduleObj.isModule()) return (RubyModule)moduleObj;
```

vs.

```java
    public RubyModule getOrCreateModule(String name) {
        IRubyObject module = objectClass.getConstantAt(name);
```

=> both methods do the same to get to the `module`  :)

PR #8085, opened 2017-08-28
MINOR: Cleanup out directory via gradle clean task

original-brownbear:
Idea (depending on your configuration and only if you run Java `main` or JUnit from the IDE) at times creates the path `"${projectDir}/out/"`.
Unfortunately, this creates an inconsistency with running tests locally because depending on your system the `/out` path may be loaded before the `/build` path. So if you ran the tests from the IDE, then make changes to Java code and run tests again from the CLI JRuby might use the now outdated `.class` files in `out` since `gradle clean` did not catch those.

This is why (or at least one possible way for this to happen) we constantly run into random issues with the build locally that requires us to use `git clean -x -f -d -e *.iml -e .idea -e .bundle`.

SO:

* Added the `/out` paths to the `clean` target
* Removed some files that accidentally got committed by exactly this route :)


jakelandis:
@original-brownbear - LGTM

Though I don't think that both `out` and `build` are on the classpath at the same time. 

`/out` was introduced with IntelliJ 2017.02, which many people (myself included) are annoyed with JetBrains change to use `/out` instead of `build`, creating a difference between IntelliJ and the command line for gradle projects.  See https://youtrack.jetbrains.com/issue/IDEA-175172

I recently merged https://github.com/elastic/logstash/commit/976e918d216ab8e3b47961cb4a09c887a8f99763 which prefers IntelliJ's `out` over the command line `build` if it exists, which may be the point of confusion.  A better fix for that may be to do what the issue suggests to ensure that IntelliJ still points to the same directory.

original-brownbear:
@jakelandis 

> Though I don't think that both out and build are on the classpath at the same time.


AAAH thanks :) so `/out` takes precedence every time ... hmm  I guess we could fix this in another PR by some Gradle idea plugin magic? :D

PR #8086, opened 2017-08-28
MINOR: Fix a lot of new compiler warnings in Metrics code

original-brownbear:
Putting this to the general review queue since it's not in any way affecting functionality, just passive cleanup to get rid of about 20 compiler warnings.

* Added missing `@Deprecated` annotations
* Added missing `serialVersionUID`
* Made methods `static` that don't reference `this`
   * This also allowed for removing some `Serializer` fields
* Fixed missorted modifiers
* Removed dead parameters, those are fine when implementing an interface obviously but should not be passed down to `private` methods :)


original-brownbear:
link #7701 :)

jakelandis:
@original-brownbear - LGTM (and thanks for the cleanup :) )

original-brownbear:
@jakelandis np + thanks for the review :)

PR #8087, opened 2017-08-28
 #7712 & #7945 BiValue Improvements

original-brownbear:
Shortens the logic for `BiValue` related stuff to get one step closer to done for #7712.
Also improves performance with higher worker counts to improve the situation for #7945  (see last point in below notes for the details).

* Turn Enum into HashMap for direct lookup
* Use stateless Lambdas instead of named classes as converter functions for better performance
* Remove redundant instanceof checks by splitting existing converter functions between Java and Ruby version
* Use most restrictive instanceof check possible for Ruby converters for performance improvements

Note:
* This is not a functional change, just a static refactoring => If you find a spot where functionality is affected it's a mistake. Things should still work as they did before for every possible input to `public static BiValue newBiValue(Object o)` => general review queue
   * I know I removed `private static String cacheName(final String cls)` but this path was dead code anyways and plain wrong. Had it actually been hit, you'd see NPEs all over the place because that would imply concurrent access to a standard `HashMap` :)
* Generally this approach (lookup of a Lambda that gets jitted to a static call that is looked up from a `Map`) JRuby sort of does the same in this kind of spot and it seems it's not getting faster than this with current JVMs => should do this as well in `Rubyfier` and `Javafier` where possible <= not possible everywhere, because `getClasss()` will only work on `final class` types obviously.
* As This improves the performance baseline very measurable for more than a single worker thread (theoretically also slightly for just one thread, but I can't measure it) => explanation for that can be found https://issues.apache.org/jira/secure/attachment/12833228/reduce_instanceof_typechecking.pdf in part

### DIFF
--- a/logstash-core/src/main/java/org/logstash/bivalues/BiValues.java
+++ b/logstash-core/src/main/java/org/logstash/bivalues/BiValues.java
@@ -3,135 +3,58 @@ package org.logstash.bivalues;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.HashMap;
+import java.util.Map;
 import org.jruby.RubyBignum;
+import org.jruby.RubyFixnum;
 import org.jruby.RubyInteger;
+import org.jruby.RubyNil;
 import org.jruby.RubySymbol;
 import org.jruby.ext.bigdecimal.RubyBigDecimal;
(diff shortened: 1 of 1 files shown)


suyograo:
Release labels please

jordansissel:
LGTM

Issue #7973, opened 2017-08-10
Witness API improvements


jakelandis:
With the introduction of a type safe fluent API (the `Witness` API) for the metrics #7788, we have the opportunity to make the the metrics read like prose.

For the initial draft, I simply mirrored the names used in the metrics namespace's.

For example:

`[:stats, :pipelines, :main, :plugins, :inputs, :"18ff6d6eee97a21c9a86ad55223c9d9211de1683-1", :events] out = 1`
 
The part in `[..]` is the namespace, and the part outside is the action/event/metric and value.

Mapped to the `Witness API`, this would be:

`      Witness.instance().pipeline("main").inputs("18ff6d6eee97a21c9a86ad55223c9d9211de1683-1").events().out();`

As @andrewvc pointed out, we should strive for:

`Witness->noun->noun->noun->verb` 

Where the chained parts are nouns and the chain terminator is expressed as verb. 

Many of the metrics already read this way ... but others do not. 

For example:

`[:stats, :pipelines, :main, :queue] type = memory`

Would read:

`Witness.instance().pipeline("main").queue().type("memory");`

Which does not follow the terminate in a verb convention. 

This is a working issue to identify any improvements to the fluent API prior to the 6.1 release. 

Below is a list of all existing namespaces and keys for a simple tcp input, not filters and a stdout. Not a comprehensive list, but a good starting point.  

Stats
----------------
[:stats, :events] duration_in_millis = 411
[:stats, :events] filtered = 1
[:stats, :events] in = 0
[:stats, :events] in = 0[:stats, :pipelines, :main, :plugins, :inputs, :"18ff6d6eee97a21c9a86ad55223c9d9211de1683-2", :events] out = 0
[:stats, :events] in = 1
[:stats, :events] out = 1
[:stats, :events] queue_push_duration_in_millis = 17
[:stats, :pipelines, :main, :config] batch_delay = 5
[:stats, :pipelines, :main, :config] batch_size = 125
[:stats, :pipelines, :main, :config] config_reload_automatic = false
[:stats, :pipelines, :main, :config] config_reload_interval = 3000000000
[:stats, :pipelines, :main, :config] dead_letter_queue_enabled = false
[:stats, :pipelines, :main, :config] workers = 8
[:stats, :pipelines, :main, :events] duration_in_millis = 411
[:stats, :pipelines, :main, :events] filtered = 1
[:stats, :pipelines, :main, :events] in = 1
[:stats, :pipelines, :main, :events] out = 1
[...]

PR #8090, opened 2017-08-28
Remove Metrics Namespace API : Part 1

jakelandis:
This is part of a much much larger commit set kept small for easier review.

The tests are not expected to pass, and will fixup the commit log once all parts are reviewed.

Changes in this commit are:
* Removal of Metrics namespace and related Ruby classes.
* Changes neccessary for the agent.rb class
* Removal of un-dcoumented optional 'turn off metrics' flag

Part of #7788


logstash-core/lib/logstash/agent.rb
```diff
@@ -464,51 +436,27 @@ def update_metrics(converge_result)
...
-      n.gauge(:last_error, { :message => action_result.message, :backtrace => action_result.backtrace})
-      n.gauge(:last_failure_timestamp, LogStash::Timestamp.now)
-    end
-  end
 
-  def initialize_agent_metrics
-    @instance_reload_metric.increment(:successes, 0)
-    @instance_reload_metric.increment(:failures, 0)
-  end
+    @witness_reloads.failure
```
andrewvc: ❤️  the ratio of removals to deletions here

logstash-core/lib/logstash/agent.rb
```diff
@@ -464,51 +436,27 @@ def update_metrics(converge_result)
...
-      n.increment(:failures, 0)
-      n.gauge(:last_error, nil)
-      n.gauge(:last_success_timestamp, nil)
-      n.gauge(:last_failure_timestamp, nil)
-    end
+    witness_pipeline_reloads = @witness_pipelines.pipeline(action.pipeline_id).reloads
+    witness_pipeline_reloads.failure
+    witness_pipeline_reloads.error.message(action_result.message)
+    witness_pipeline_reloads.error.backtrace(action_result.backtrace.to_s)
+    witness_pipeline_reloads.last_failure_timestamp(LogStash::Timestamp.now)
```
andrewvc: Can we consolidate these operations into a single method on the witness? My thinking is that it'd be a cleaner separation of concerns to have the externally used witness methods correspond to lifecycle events rather than act as setters.

Maybe something like:

```ruby
witness.pipeline(action.pipeline_id).failure(message, action_result.backtrace.to_s)
```

logstash-core/lib/logstash/agent.rb
```diff
@@ -464,51 +436,27 @@ def update_metrics(converge_result)
...
 
   def update_successful_reload_metrics(action, action_result)
-    @instance_reload_metric.increment(:successes)
+    @witness_reloads.success
 
-    @pipeline_reload_metric.namespace([action.pipeline_id, :reloads]).tap do |n|
-      n.increment(:successes)
-      n.gauge(:last_success_timestamp, action_result.executed_at)
-    end
+    witness_pipeline_reloads = @witness_pipelines.pipeline(action.pipeline_id).reloads
```
andrewvc: Same note RE: consolidation of operations

logstash-core/spec/logstash/agent_spec.rb
```diff
@@ -329,7 +329,6 @@
 
     let(:agent_args) do
       {
-        "metric.collect" => true,
```
andrewvc: Why was this removed? Were metrics off by default during tests before?

jakelandis: moving forward, there will not be any 'turn on/off the metrics' flag

logstash-core/lib/logstash/agent.rb
jakelandis: +1 , i will gather these all suggestions like this up over at https://github.com/elastic/logstash/issues/7973 and make a final sweep of the code with those changes.

jakelandis:
I moved https://github.com/elastic/logstash/issues/7973 upto the 'TODO' list before #7788 is considered done. (which accounts for the consolidation referenced above). 

Merging to feature branch.

Issue #8004, opened 2017-08-16
Accessing Metrics Endpoint during Startup Returns HTTP Status 500 


original-brownbear:
If I keep refreshing my browser pointed at the metrics page during LS startup I get this error in the logs:

```sh
➜  logstash git:(master) bin/logstash -b 128 -w 1 -e 'input { generator{}} output {stdout {codec => "dots"}}' 
WARNING: Default JAVA_OPTS will be overridden by the JAVA_OPTS defined in the environment. Environment JAVA_OPTS are -Xms4g -Xmx4g -Djava.net.preferIPv4Stack=true
Sending Logstash's logs to /Users/brownbear/src/logstash/logs which is now configured via log4j2.properties
[2017-08-16T17:19:10,018][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"fb_apache", :directory=>"/Users/brownbear/src/logstash/modules/fb_apache/configuration"}
[2017-08-16T17:19:10,036][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"netflow", :directory=>"/Users/brownbear/src/logstash/modules/netflow/configuration"}
[2017-08-16T17:19:10,222][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified
[2017-08-16T17:19:11,080][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[...]

Issue #6824, opened 2017-03-21
cgroups not existing on all linux versions


ruflin:
* version: master

Running Logstash in a Docker container I get the following log message very few seconds:

```
logstash_1       | [2017-03-21T15:14:58,454][DEBUG][logstash.instrument.periodicpoller.cgroup] Error, cannot retrieve cgroups information {:exception=>"Errno::ENOENT", :message=>"No such file or directory - /sys/fs/cgroup/cpuacct/docker/17b6470608493e662e1d4e0dabcf937b0de9a9ce06197b4fe7afa6acf8f19837/cpuacct.usage"}
```

The problem is that cgroups are not available on all linux system or can also be disabled. We had a similar issue in beats: https://github.com/elastic/beats/pull/3692 I'm aware that the above is only a debug log message but I would suggest that the check if cgroup is available should happen only once on startup.

ph:
Seems linked to #6797, I will check for that.

ruflin:
These are 2 different things: One is `cgroups` is mounted in a different place which is #6797 . The other part is `cgroups` is disabled or does not exist on the system. Could well be that my error is because of #6797 but the second issue still exists I assume.

ph:
@ruflin Right, I am +1 to do the check only once I will check to change that. But also this error is only logged at **debug** mode so the noise should be low in normal situation.

jurajseffer:
This also happens on EC2 running Debian Jessie
`
[2017-05-12T12:49:06,341][DEBUG][logstash.instrument.periodicpoller.cgroup] Error, cannot retrieve cgroups information {:exception=>"Errno::ENOENT", :message=>"No such file or directory - /sys/fs/cgroup/cpuacct/cpu.cfs_period_us"}
`

zh-h:
This also happens on Fedora WorkStation
```bash
logstash_1       | 05:41:05.517 [pool-2-thread-3] DEBUG logstash.instrument.periodicpoller.cgroup - Error, cannot retrieve cgroups information {:exce
ption=>"Errno::ENOENT", :message=>"No such file or directory - /sys/fs/cgroup/cpuacct/docker/7b1f437d0b80b70b4463a1c7c18f4b71a56eed21f2b2c66a78127e20
8fc17c5d/cpuacct.usage"}
```

meyervp:
And my on debian jessie and docker container /sys mounted
[...]

PR #8091, opened 2017-08-28
Metrics: Introduce the Scheduled Witness and Scheduler

jakelandis:
Introduce the Scheduled Witness and Scheduler
- A Witness that is self populated on a schedule
- load average will come in future commit

Change default scheduled interval from 5 to 10 seconds.

Un-deprecate LazyDelegatingGaguge
 - It is the strategy for Ruby based plugins that allows them to continue to duck type.

 Minor test coverage fixes.

Note - this is dead code until the Witness API is wired in via Ruby

Part of #7788

### DIFF
--- a/logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
@@ -11,9 +11,7 @@ import org.logstash.instrument.metrics.MetricType;
  * A lazy proxy to a more specific typed {@link GaugeMetric}. The metric will only be initialized if the initial value is set, or once the {@code set} operation is called.
  * <p><strong>Intended only for use with Ruby's duck typing, Java consumers should use the specific typed {@link GaugeMetric}</strong></p>
  *
- * @deprecated - there are no plans to replace this.
  */
-@Deprecated
 public class LazyDelegatingGauge extends AbstractMetric<Object> implements GaugeMetric<Object, Object> {
 
     private static final Logger LOGGER = LogManager.getLogger(LazyDelegatingGauge.class);
@@ -26,9 +24,8 @@ public class LazyDelegatingGauge extends AbstractMetric<Object> implements Gauge
      * Constructor - null initial value
      *
      * @param key The key <i>(with in the namespace)</i> for this metric
-     * @deprecated - there are no plans to replace this
      */
-    @Deprecated
+    
     public LazyDelegatingGauge(final String key) {
         this(key, null);
     }
@@ -38,10 +35,8 @@ public class LazyDelegatingGauge extends AbstractMetric<Object> implements Gauge
      *
      * @param key          The key for this metric
      * @param initialValue The initial value for this {@link GaugeMetric}, may be null
-     * @deprecated - there are no plans to replace this
      */
-    @Deprecated
-    public LazyDelegatingGauge(String key, Object initialValue) {
+     public LazyDelegatingGauge(String key, Object initialValue) {
         super(key);
         this.key = key;
         if (initialValue != null) {
--- a/logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
@@ -8,6 +8,8 @@ import org.logstash.instrument.witness.pipeline.EventsWitness;
 import org.logstash.instrument.witness.pipeline.PipelineWitness;
 import org.logstash.instrument.witness.pipeline.PipelinesWitness;
 import org.logstash.instrument.witness.pipeline.ReloadWitness;
+import org.logstash.instrument.witness.process.ProcessWitness;
+import org.logstash.instrument.witness.schedule.WitnessScheduler;
 
 import java.io.IOException;
 import java.util.Arrays;
@@ -28,8 +30,10 @@ public final class Witness implements SerializableWitness {
     private final ReloadWitness reloadWitness;
     private final EventsWitness eventsWitness;
     private final PipelinesWitness pipelinesWitness;
+    private final ProcessWitness processWitness;
+    private final WitnessScheduler processWitnessScheduler;
 
-    private static Witness _instance;
+    private static Witness instance;
 
     /**
      * Constructor. Consumers should use {@link #instance()} method to obtain an instance of this class.
@@ -39,6 +43,8 @@ public final class Witness implements SerializableWitness {
         this.reloadWitness = new ReloadWitness();
         this.eventsWitness = new EventsWitness();
         this.pipelinesWitness = new PipelinesWitness();
+        this.processWitness = new ProcessWitness();
+        this.processWitnessScheduler = new WitnessScheduler(processWitness);
     }
 
     /**
@@ -46,10 +52,19 @@ public final class Witness implements SerializableWitness {
      * active instance at any time.  Exposing this allows Ruby to create the instance for use in it's agent constructor, then set it here for all to use as a singleton.
      * <p>THIS IS ONLY TO BE USED BY THE RUBY AGENT</p>
      *
-     * @param __instance The instance of the {@link Witness} to use as the singleton instance that mirror's the agent's lifecycle.
+     * @param newInstance The instance of the {@link Witness} to use as the singleton instance that mirror's the agent's lifecycle.
      */
-    public static void setInstance(Witness __instance) {
-        _instance = __instance;
+    public static synchronized void setInstance(Witness newInstance) {
+        //Ruby agent restart
+        if (instance != null) {
+            instance.processWitnessScheduler.shutdown();
+        }
+
+        instance = newInstance;
+
+        if (instance != null) {
+            instance.processWitnessScheduler.schedule();
+        }
     }
 
     /**
@@ -59,10 +74,10 @@ public final class Witness implements SerializableWitness {
      * @throws IllegalStateException if attempted to be used before being set.
      */
     public static Witness instance() {
-        if (_instance == null) {
+        if (instance == null) {
             throw new IllegalStateException("The stats witness instance must be set before it used. Called from: " + Arrays.toString(new Throwable().getStackTrace()));
         }
-        return _instance;
+        return instance;
     }
 
     public EventsWitness events() {
@@ -85,6 +100,15 @@ public final class Witness implements SerializableWitness {
      */
     public PipelinesWitness pipelines() {
         return pipelinesWitness;
+    }
+
+    /**
+     * Obtain a reference to the associated process witness.
+     *
+     * @return The associated {@link ProcessWitness}
+     */
+    public ProcessWitness process() {
+        return processWitness;
     }
 
     /**
@@ -133,6 +157,7 @@ public final class Witness implements SerializableWitness {
         }
 
         static void innerSerialize(Witness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            witness.process().genJson(gen, provider);
             witness.events().genJson(gen, provider);
             witness.reloads().genJson(gen, provider);
             witness.pipelinesWitness.genJson(gen, provider);
--- a/logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
+++ b/logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
@@ -1,0 +1,242 @@
+package org.logstash.instrument.witness.process;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import com.sun.management.UnixOperatingSystemMXBean;
+import org.logstash.instrument.metrics.Metric;
+import org.logstash.instrument.metrics.gauge.NumberGauge;
+import org.logstash.instrument.witness.MetricSerializer;
+import org.logstash.instrument.witness.SerializableWitness;
+import org.logstash.instrument.witness.schedule.ScheduledWitness;
+
+import java.io.IOException;
+import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A scheduled witness for process metrics
+ */
+@JsonSerialize(using = ProcessWitness.Serializer.class)
+public class ProcessWitness implements SerializableWitness, ScheduledWitness {
+
+    private static final OperatingSystemMXBean osMxBean;
+    private static final String KEY = "process";
+    public  static final boolean isUnix;
+    private static final UnixOperatingSystemMXBean unixOsBean;
+    private final NumberGauge openFileDescriptors;
+    private final NumberGauge peakOpenFileDescriptors;
+    private final NumberGauge maxFileDescriptors;
+    private final Cpu cpu;
+    private final Memory memory;
+    private final Snitch snitch;
+
+    static {
+        osMxBean = ManagementFactory.getOperatingSystemMXBean();
+        isUnix = osMxBean instanceof UnixOperatingSystemMXBean;
+        unixOsBean = isUnix ? (UnixOperatingSystemMXBean) osMxBean : null;
+    }
+
+    /**
+     * Constructor
+     */
+    public ProcessWitness() {
+        this.openFileDescriptors = new NumberGauge("open_file_descriptors", -1);
+        this.maxFileDescriptors = new NumberGauge("max_file_descriptors", -1);
+        this.peakOpenFileDescriptors = new NumberGauge("peak_open_file_descriptors", -1);
+        this.cpu = new Cpu();
+        this.memory = new Memory();
+        this.snitch = new Snitch(this);
+    }
+
+    @Override
+    public void refresh() {
+        if (isUnix) {
+            long currentOpen = unixOsBean.getOpenFileDescriptorCount();
+            openFileDescriptors.set(currentOpen);
+            if (maxFileDescriptors.getValue() == null || peakOpenFileDescriptors.getValue().longValue() < currentOpen) {
+                peakOpenFileDescriptors.set(currentOpen);
+            }
+            maxFileDescriptors.set(unixOsBean.getMaxFileDescriptorCount());
+        }
+        cpu.refresh();
+        memory.refresh();
+    }
+
+    /**
+     * Get a reference to associated snitch to get discrete metric values.
+     *
+     * @return the associate {@link Snitch}
+     */
+    public Snitch snitch() {
+        return snitch;
+    }
+
+    /**
+     * An inner witness for the process / cpu metrics
+     */
+    public class Cpu implements ScheduledWitness {
+        private static final String KEY = "cpu";
+        private final NumberGauge cpuProcessPercent;
+        private final NumberGauge cpuTotalInMillis;
+
+        private Cpu() {
+            this.cpuProcessPercent = new NumberGauge("percent", -1);
+            this.cpuTotalInMillis = new NumberGauge("total_in_millis", -1);
+        }
+
+        @Override
+        public void refresh() {
+            cpuProcessPercent.set(scaleLoadToPercent(unixOsBean.getProcessCpuLoad()));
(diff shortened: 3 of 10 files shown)


logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
```diff
@@ -38,10 +35,8 @@ public LazyDelegatingGauge(final String key) {
      *
      * @param key          The key for this metric
      * @param initialValue The initial value for this {@link GaugeMetric}, may be null
-     * @deprecated - there are no plans to replace this
      */
-    @Deprecated
-    public LazyDelegatingGauge(String key, Object initialValue) {
```
andrewvc: Why do we still need this?

logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
```diff
@@ -39,17 +43,28 @@ public Witness() {
...
     /**
      * This is a dirty hack since the {@link Witness} needs to mirror the Ruby agent's lifecycle which, at least for testing, can mean more then 1 instance per JVM, but only 1
      * active instance at any time.  Exposing this allows Ruby to create the instance for use in it's agent constructor, then set it here for all to use as a singleton.
      * <p>THIS IS ONLY TO BE USED BY THE RUBY AGENT</p>
      *
-     * @param __instance The instance of the {@link Witness} to use as the singleton instance that mirror's the agent's lifecycle.
+     * @param newInstance The instance of the {@link Witness} to use as the singleton instance that mirror's the agent's lifecycle.
      */
-    public static void setInstance(Witness __instance) {
-        _instance = __instance;
```
andrewvc: I like the `_` free formatting too, but I'm curious as to why you made the change. Just a change of opinion?

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
```diff
@@ -0,0 +1,247 @@
...
+
+import java.io.IOException;
+import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A scheduled witness for process metrics
+ */
+@JsonSerialize(using = ProcessWitness.Serializer.class)
```
andrewvc: Much cleaner than the previous situation with this split out into a separate library. :)

logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
jakelandis: yeah, variables that start with `_` is Java isn't really a thing, despite my best efforts it hasn't caught on.  I just changed my mind.

logstash-core/src/main/java/org/logstash/instrument/metrics/gauge/LazyDelegatingGauge.java
jakelandis: This constructor, or this class ?

(Assuming you mean this class)
This is the backing implementation used by the plugins that all them to continue to duck type. For example from the beats input (ruby side):
```ruby
@metric.gauge(:peak_connections, @peak_connection_count.value)
```
Will find its way to here, and we lazily implement the correctly typed Gauge.

jakelandis:
Jenkins test this please.

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/WitnessScheduler.java
```diff
@@ -0,0 +1,82 @@
...
+     * Shuts down the underlying executor service. Since these are daemon threads, this is not absolutely necessary.
+     */
+    public void shutdown(){
+        executorService.shutdown();
+        try {
+            if(!executorService.awaitTermination(5, TimeUnit.SECONDS)){
+                executorService.shutdownNow();
+            }
+        } catch (InterruptedException e) {
+            //do nothing
```
original-brownbear: @jakelandis are we sure we want to suppress an interrupt here? (won't this interfere with orderly shutdowns on a single outside interrupt? ... just asking, I'm not fully sure what Thread this will be called from)

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/ScheduledWitness.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Sets the priority for the underlying thread. Consumers should NOT set this unless they are absolutely certain. Defaults to {@link Thread#NORM_PRIORITY}
+     *
+     * @return the thread scheduled thread priority
+     */
+    default int priority() {
+        //note - the system property is an only an escape hatch if this proves to cause performance issues. Do not document this system property, it is not part of the contract.
+        return Integer.parseInt(System.getProperty("witness.schedule.priority", Integer.toString(Thread.NORM_PRIORITY)));
```
original-brownbear: @jakelandis any reason why we'd want to expose this (even internally)?

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/WitnessScheduler.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    class RefreshRunnable implements Runnable {
+
+       long lastLogged = 0;
+
+        @Override
+        public void run() {
+            try {
+                witness.refresh();
+            } catch (Exception e) {
```
original-brownbear: @jakelandis what exceptions do we assume to get here? Shouldn't we simply crash LS here on unexpected bugs instead of simply continuing to run with frozen metrics?

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        @Override
+        public void serialize(ProcessWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
+            gen.writeStartObject();
+            innerSerialize(witness, gen, provider);
+            gen.writeEndObject();
+        }
+
+        void innerSerialize(ProcessWitness witness, JsonGenerator gen, SerializerProvider provider) throws IOException {
```
original-brownbear: @jakelandis this method is `static` isn't it?

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
```diff
@@ -0,0 +1,247 @@
...
+    private final Snitch snitch;
+
+    /**
+     * Constructor
+     */
+    public ProcessWitness() {
+        this.openFileDescriptors = new NumberGauge("open_file_descriptors", -1);
+        this.maxFileDescriptors = new NumberGauge("max_file_descriptors", -1);
+        this.peakOpenFileDescriptors = new NumberGauge("peak_open_file_descriptors", -1);
+        this.isUnix = osMxBean instanceof UnixOperatingSystemMXBean;
```
original-brownbear: @jakelandis this is just a `static` imo, it will never change during the lifetime of the JVM :)

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
```diff
@@ -0,0 +1,247 @@
...
+         * Gets if this process is running on *nix based system.
+         *
+         * @return true if host is *nix, false otherwise
+         */
+        public boolean isUnix() {
+            return witness.isUnix;
+        }
+    }
+
+    private short scaleLoadToPercent(double load) {
```
original-brownbear: @jakelandis can we use `int` ? In this day and age almost every Linter (PMD etc.) complains about `short` :D <= not an important point though, feel free to ignore :)

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
```diff
@@ -0,0 +1,247 @@
...
+        public long memTotalVirtualInBytes() {
+            return witness.memory.memTotalVirtualInBytes.getValue().longValue();
+        }
+
+        /**
+         * Gets if this process is running on *nix based system.
+         *
+         * @return true if host is *nix, false otherwise
+         */
+        public boolean isUnix() {
```
original-brownbear: @jakelandis not sure we should have this method at all, every use of it could just be replaced by a `static` constant couldn't it? (just trying to save some code size:))

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
```diff
@@ -0,0 +1,247 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A scheduled witness for process metrics
+ */
+@JsonSerialize(using = ProcessWitness.Serializer.class)
+public class ProcessWitness implements SerializableWitness, ScheduledWitness {
+
+    private static final OperatingSystemMXBean osMxBean = ManagementFactory.getOperatingSystemMXBean();
+    private final static String KEY = "process";
```
original-brownbear: @jakelandis nit: missorted modifiers :)

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
```diff
@@ -0,0 +1,247 @@
...
+            cpuProcessPercent.set(scaleLoadToPercent(unixOsBean.getProcessCpuLoad()));
+            cpuTotalInMillis.set(TimeUnit.MILLISECONDS.convert(unixOsBean.getProcessCpuTime(), TimeUnit.NANOSECONDS));
+        }
+    }
+
+    /**
+     * An inner witness for the the process / memory metrics
+     */
+    public class Memory implements ScheduledWitness {
+        private final static String KEY = "mem";
```
original-brownbear: @jakelandis nit: missorted modifiers :)

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
```diff
@@ -0,0 +1,247 @@
...
+     */
+    public Snitch snitch() {
+        return snitch;
+    }
+
+    /**
+     * An inner witness for the process / cpu metrics
+     */
+    public class Cpu implements ScheduledWitness {
+        private final static String KEY = "cpu";
```
original-brownbear: @jakelandis nit: missorted modifiers :)

logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
```diff
@@ -28,8 +30,10 @@
     private final ReloadWitness reloadWitness;
     private final EventsWitness eventsWitness;
     private final PipelinesWitness pipelinesWitness;
+    private final ProcessWitness processWitness;
```
original-brownbear: @jakelandis this one could probably just be made `static` right? It's not gonna change during the lifetime of the JVM is it?

logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
```diff
@@ -39,17 +43,28 @@ public Witness() {
...
      * This is a dirty hack since the {@link Witness} needs to mirror the Ruby agent's lifecycle which, at least for testing, can mean more then 1 instance per JVM, but only 1
      * active instance at any time.  Exposing this allows Ruby to create the instance for use in it's agent constructor, then set it here for all to use as a singleton.
      * <p>THIS IS ONLY TO BE USED BY THE RUBY AGENT</p>
      *
-     * @param __instance The instance of the {@link Witness} to use as the singleton instance that mirror's the agent's lifecycle.
+     * @param newInstance The instance of the {@link Witness} to use as the singleton instance that mirror's the agent's lifecycle.
      */
-    public static void setInstance(Witness __instance) {
-        _instance = __instance;
+    public static void setInstance(Witness newInstance) {
```
original-brownbear: @jakelandis might be a good idea to `synchronize` this method to avoid weird concurrency issues leaving dangling daemon threads around?

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/ScheduledWitness.java
jakelandis: No specific reason.  My thought process was more of a general template to implement background services.  If it doesn't make sense, I can remove.

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/WitnessScheduler.java
jakelandis: Shutdown will only be called from the main thread, if called at all. The executor service will always contain daemon threads, so on JVM termination there is no need to call this method. Honestly this method really only exists to help testing.

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/WitnessScheduler.java
jakelandis: The real world example I was thinking of was https://github.com/elastic/logstash/issues/6824 where CPU metrics could not be gathered on specific OS's.  I don't think the metrics should take down the primary function of Logstash if they don't work or there is a transient error.

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
jakelandis: I was just mirroring what this is replacing, but I agree, will change to `int`

logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
jakelandis: Will do.  _FWIW, only the tests will ever call this more then once._

logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
jakelandis: It is an instance inside a singleton [1], so the cost difference is trival.

[1]
 almost a singleton. The Ruby lifecyle and JVM lifecycles are not in sync for the tests, so for testing, keeping this an instance of this class makes life simpler.

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
jakelandis: You caught me!  I will fix this and the others in this review. Do know if IntelliJ can auto fix these ?

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
jakelandis: will move to static final (or is final static ? ) ;)

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
jakelandis: Yeah there are few things minor things in this review don't mesh with some of the clean up you did and I reviewed this AM.  I will take a quick sweep through and clean it up.

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
jakelandis: will do

suyograo:
Release labels please.

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/WitnessScheduler.java
original-brownbear: @jakelandis 

> Shutdown will only be called from the main thread, if called at all

but **if** called from the main thread it will remove its interrupt status? Which may be a problem?

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/ScheduledWitness.java
original-brownbear: @jakelandis At least I don't see a valid reason for manually setting this here ever => I'd kill it :)

logstash-core/src/main/java/org/logstash/instrument/witness/process/ProcessWitness.java
original-brownbear: @jakelandis yea mine can. Ping me later and I can give you my settings :)

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/WitnessScheduler.java
original-brownbear: @jakelandis from practical experience I'd strongly disagree here. I think one of if not the main use case for the metrics is monitoring. If a malfunctioning LS may simply cause the metrics to freeze, this is a huge issue for users.
Errors like the one you linked we should handle on a case by case basis if we know that ignoring them is not affecting functionality overall, but suppressing `100%` of errors and freezing metrics in time makes them invalid for monitoring purposes imo.

logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
original-brownbear: @jakelandis they aren't? Where + why? (having them not in sync seems like a big problem actually, much of our functionality around `RubyUtils` relies on the fact that we can cache the global `Ruby` instance)

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/WitnessScheduler.java
jakelandis: How/why would anyone signal interrupt to the main thread ?  It depends on what we define as the interrupt policy for main thread.  IMO, the main thread should be a `do nothing` policy.

WDYT?

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/WitnessScheduler.java
jakelandis: @original-brownbear - we may need a 3rd party to resolve this one, as I strongly agree with the approach here.  

* Not everyone using Logstash cares about metrics
* Metrics are a side effect of the main purpose, not an integral part of it's function
* There are known issues on some platforms, and crashing Logstash would dis-allow those platforms/configurations to update.
* This change as-is is passive to the prior version.

original-brownbear: @jakelandis 

> This change as-is is passive to the prior version.

I still feel like I'm right but ... you win with this one. I should've known ... even reported #8004 :)

logstash-core/src/main/java/org/logstash/instrument/witness/Witness.java
jakelandis: @original-brownbear - I should re-phrase, the Ruby **Agent** life-cycle and JVM lifecycle are not sync for the rspec tests.  Specifically it is in the rspec unit tests, we create hundreds of new agents, on the same JVM.  Outside of testing, there is a 1:1 relationship.

original-brownbear: @jakelandis hmm hmm hmm it still seems dirty to have something related to the process, not `static final` but I don't really know enough about the details here to suggest a better approach at this point :) => Let's just go with this for now, it's a really minor point anyways => sorry for wasting time ;)

jakelandis:
@original-brownbear  / @andrewvc - Review changes posted here: https://github.com/elastic/logstash/pull/8091/commits/cf5597041da27d9cb6b96bc165f132e6990dd090

(will squash prior to merge)

logstash-core/src/main/java/org/logstash/instrument/witness/schedule/WitnessScheduler.java
original-brownbear: @jakelandis argh so sorry missed answering this one before you pushed changes :(
That said:

> How/why would anyone signal interrupt to the main thread?

Exactly, I don't know :D and it probably shouldn't happen here ever. => rethrow `IllegalstateException` imo :)

jakelandis: @original-brownbear - sounds good.  commit here : https://github.com/elastic/logstash/pull/8091/commits/f75c1987edaba437117af9d92224f564fea5766b

jakelandis:
@andrewvc - are you good with these changes (I rebased and squashed)

PR #8092, opened 2017-08-28
Upgrade minitar to fix CVE-2016-10173

vanstee:
Noticed `rubyzip` was bumped to take care of a similar CVE, but `minitar` is still pinned to the old version.

PR which fixed the issue: https://github.com/halostatue/minitar/issues/16

Shipping a new version of `logstash-core` and `logstash-core-plugin-api` would be super helpful as well.

### DIFF
--- a/logstash-core/logstash-core.gemspec
+++ b/logstash-core/logstash-core.gemspec
@@ -42,7 +42,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency "i18n", "= 0.6.9" #(MIT license)
 
   # filetools and rakelib
-  gem.add_runtime_dependency "minitar", "~> 0.5.4"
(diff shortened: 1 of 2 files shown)


suyograo:
Hi, thanks for the patch. Can you please perform step 2 of https://github.com/elasticsearch/logstash/blob/master/CONTRIBUTING.md#contribution-steps

vanstee:
@suyograo Just signed it a second time. I guess the CLA status doesn't update here though. Also it looks like CI flaked but I can't rerun it without write perms on the repo. Could you kick that off for me?

jsvd:
we now use minitar ~> 0.8.0, this can be closed https://github.com/elastic/logstash/blob/master/logstash-core/logstash-core.gemspec#L67

PR #8101, opened 2017-08-29
MINOR: Use more optimal loop in draining Queue

original-brownbear:
Not much to say here: 
* No need to check the conditional twice when we know we're always draining at least one element.
   * Also fixes/enables branch prediction nicely since we're almost always draining a full batch now and only execute the loop once that we have a properly sized `ArrayBlockingQueue` and dropped the `SynchronousQueue`:)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
+++ b/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
@@ -37,13 +37,13 @@ public final class LsQueueUtils {
         int left = count;
         final Collection<JrubyEventExtLibrary.RubyEvent> collection =
             new HashSet<>(4 * count / 3 + 1);
-        while (left > 0) {
+        do {
             final int drained = drain(queue, collection, left, nanos);
             if (drained == 0) {
                 break;
             }
             left -= drained;
-        }
+        } while (left > 0);
         return collection;
     }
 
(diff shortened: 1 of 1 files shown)


jordansissel:
It's difficult for me to evaluate the claimed effect (branch prediction). 

I wrote some sample code to attempt to measure the differences here and was not able to find much conclusive difference.

Code and `perf stat` results here: https://gist.github.com/jordansissel/4a3c33985e7e1c9b9a564818f0df6c86

Notably, the `instruction count` for both `a` and `b` functions had very similar execution characteristics between 200-212million. Branches around 38-39 million. branch-misses between 1.32-1.35 million.

original-brownbear:
@jordansissel haha I don't think you'll be able to demonstrate it with that kind of test, too much noise from the JVM doing god knows what else behind the scenes (GC, JITing and what not).

But that said:

### While Loop

screen shot 2017-08-30 at 02 34 49


### Do-While Loop

screen shot 2017-08-30 at 02 39 24

JVM def optimizes this away in the C2 pass :) Not that it matters much to fix these trivial spots, but it cuts down on the noise when profiling :)

original-brownbear:
Plus I mean it's not even just tricky branch prediction at play here, we simply save one `if` outright by not comparing on the first go :)

jordansissel:
The code change LGTM; performance aside, it expresses the intent that we always will do at least one drain() attempt.

PR #8104, opened 2017-08-30
PERFORMANCE: Much Faster ConvertedMap

original-brownbear:
* Use IdentityHashMap since we intern all possible keys anyways and can look them up more efficiently than `String.intern()` from our PathCache (which only holds interned `String`s in its cached `FieldReference`s )
* Use the PathCache to never have to instantiate new `String` when converting a RubyHash

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core/src/main/java/org/logstash/Accessors.java
@@ -103,7 +103,7 @@ public final class Accessors {
 
     private static Object setChild(final Object target, final String key, final Object value) {
         if (target instanceof Map) {
-            ((ConvertedMap) target).put(key, value);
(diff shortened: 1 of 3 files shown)


original-brownbear:
The effects of this one are much more pronounced when combined with #8103 due to better inlining of smaller `IdentityHashMap` methods  (which we can take advantage of only with the smaller codesize in some paths that comes from  #8103).
The effects are even more drastic when applied to #7950 since many more conditionals are getting inlined there as a result of the smaller bytecode for field lookups :)

original-brownbear:
@guyboertje I'm afraid I need you here as well, this one requires understanding how `FieldReference` and `PathCache` work.

original-brownbear:
@guyboertje thanks :)

PR #8107, opened 2017-08-30
 #7712 Integer- and Long-BiValue Removed

original-brownbear:
For #7712 just like #7854 and similar.

Only two special (but straightforward) things here:

* Removed this logic (or rather did not replace it) `hm.put(RubyInteger.class, value -> new IntegerBiValue((RubyInteger) value));` (this was wrong!, `RubyInteger` is the superclass to `RubyFixnum` (64bit), but also `RubyBignum` (unbounded in size because it's just a  wrapper for `java.math.BigInteger`))
* `Integer` and `Long` both handled as 8-byte integers long for now. Ruby has no 4-byte numeric data type and the Ruby filter execution doesn't work with 4-byte integers either => for now, it's 100% redundant to deal with 4-byte integers
   * Added test for long field serialization + we already had one for int field => all working fine imo :)


original-brownbear:
@guyboertje ping, this one is a bit of a blocker :)

original-brownbear:
@guyboertje weird I can't see that question. But yes, my mistake => will fix :)

guyboertje:
That is weird. I guess we can't trust Review comments to be visible.

original-brownbear:
@guyboertje fixed :)

guyboertje:
My bad. I don't know how to use the Review feature. I should have clicked the `Start a conversation` button.

PR #8112, opened 2017-08-31
[6.0] Bump version to RC1

suyograo:


ph:
LGTM

jordansissel:
I found one instance of 6.0.0-beta stil hanging around: docs/index-shared1.asciidoc

ph:
@jordansissel you are correct, I've looked at the rendered content, I think the rake's set version that we use could that care of that.

ph:
I've create this https://github.com/elastic/logstash/issues/8114

suyograo:
@jordansissel @ph the docs version only gets updated on the day of release, so we're good here.

PR #8116, opened 2017-08-31
Windows build fixes

colinsurprenant:
Modifications to have a green build on Windows.

Tested for successful `rake test:core` locally on Windows Server 2012 and on OSX. 

These are mostly spec fixes. 

The only non-spec significant change which should prevent potential problems on all platforms is in `logstash-core/lib/logstash/pipeline_action/stop.rb` where after issuing a pipeline shutdown, we were not waiting on the pipeline thread to complete.

### DIFF
--- a/logstash-core/spec/logstash/api/modules/node_stats_spec.rb
+++ b/logstash-core/spec/logstash/api/modules/node_stats_spec.rb
@@ -69,7 +69,7 @@ describe LogStash::Api::Modules::NodeStats do
       "cpu"=>{
         "total_in_millis"=>Numeric,
         "percent"=>Numeric,
-        "load_average" => { "1m" => Numeric }
+        # load_average is not supported on Windows, set it below
       }
     },
    "pipelines" => {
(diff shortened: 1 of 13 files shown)


jordansissel:
I will test this shortly.

jordansissel:
```
Finished in 11 minutes 36 seconds (files took 8.31 seconds to load)
2328 examples, 0 failures, 3 pending
```

logstash-core/spec/logstash/api/modules/node_stats_spec.rb
```diff
@@ -69,7 +69,7 @@
       "cpu"=>{
         "total_in_millis"=>Numeric,
         "percent"=>Numeric,
-        "load_average" => { "1m" => Numeric }
+        # load_average is not supported on Windows, set it below
```
jordansissel: I confirmed manually that on Elasticsearch (I used ES 5.5.0) only has load_average on Linux. On Windows, the `cpu` object only has `percent`.

jordansissel:
@colinsurprenant Thanks for solving these issues!

colinsurprenant:
I decided to rebase into 2 separate commits, one for the specific tests specs fixes and the other for the pipeline stop action fix which for some reason only surfaced in windows tests but is definitely a potential problem everywhere.

colinsurprenant:
merged in master, 6.x and 6.0

PR #8117, opened 2017-08-31
Really fix verbose logging in tests

guyboertje:
In dev_utils spec_helper we set the logging to "OFF". However in runner we call initialize and turn it back on. We need to stub this method too.

in `logstash-core/lib/logstash/logging/logger.rb`, rename `initialize` class method to `reconfigure` because rspec really complains when you stub a method called `initialize`
in `logstash-core/lib/logstash/runner.rb`, use the new name `reconfigure`
in `logstash-core/spec/logstash/runner_spec.rb`, stub `reconfigure`, remove after block

### DIFF
--- a/logstash-core/lib/logstash/logging/logger.rb
+++ b/logstash-core/lib/logstash/logging/logger.rb
@@ -92,6 +92,9 @@ module LogStash
         end
       end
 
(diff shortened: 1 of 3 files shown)


guyboertje:
test failure unrelated. integration. nice clean log free spec console output.

jakelandis:
LGTM

PR #8122, opened 2017-09-01
Close channel associated with lock

robbavey:
Close channel associated with lock
    
    Fix Windows build issue where .lock file cannot be deleted due to
    underling channel associated with lock not being closed. This
    will stop the file from being deleted until the JVM exits, despite
    reporting through Java that the file *has* been deleted and does not
    exist. It also blocks new files from being created with the same
    filename, which caused the test failure here.
    
    Backport of #7822


colinsurprenant:
LGTM

colinsurprenant:
followup in #8124  ... it seems we could DRY this code moving forward.

PR #8130, opened 2017-09-03
 #8128 Fix Timestamp not getting properly deserialized under custom key

### LINKED ISSUE #8128: Timestamp Values are not Correctly Serialized for all Keys
Currently, this does not work:

```java
        Event e = new Event();
        final Timestamp timestamp = new Timestamp();
        e.setField("tsfoo", timestamp);
        assertEquals(timestamp, er.getField("tsfoo"));
```

it throws

```java
java.lang.AssertionError: expected: org.logstash.Timestamp<2017-09-03T07:49:53.853Z> but was: java.lang.String<2017-09-03T07:49:53.853Z>
Expected :org.logstash.Timestamp<2017-09-03T07:49:53.853Z> 
Actual   :java.lang.String<2017-09-03T07:49:53.853Z>
```

The reason for this is that we're using a simple `JsonSerializer` without type information to serialize it.

```java
public class TimestampSerializer extends JsonSerializer<Timestamp> {

    @Override
    public void serialize(Timestamp value, JsonGenerator jgen, SerializerProvider provider)
            throws IOException
    {
        jgen.writeString(value.toString());
    }
}
```

This is not caused by c8b8b6bf6 (I verified that :)) => we can't fix this by the Java serialization route (the Java route we had there wasn't ever used even because of the registered `JsonSerializer`).
This should be fixed by registering a typed serializer for the CBOR (Json cannot be typed because it'll break ES output as far as I understand) serialization path.
Currently, this would break any input that puts a `Timestamp` field and passes PQ.

original-brownbear:
This is what I had to do to fix #8128:

* Move CBOR serializer to typed serialization for non-json types
   * This required a no-arg constructor on `ConvertedMap` since `IdentityHashmap` seems to not be supported out of the box by Jackson (same for `ConvertedList`, also added a UT for this case since we didn't have a `List` + CBOR test :))
* Moving `Timestamp` and `RubyTimestamp` to serializers that can optionally store type information
* Had to adjust the hard coded byte counts in some DLQ tests ... we should fix those tests, testing this way makes little to no sense imo
* Add a custom deserializer `TimestampDeserializer` for `Timestamp`

Also:
* Ditched the redundant `toIso8601` serialization method on `TImestamp`
* Moved everything from annotation based registration to explicit serializer registration to make this more streamlined
* Removed redundant constructor args from serializer instantiations, researching the code around this I learned that the boolean argument is a dummy in the version we use ...


original-brownbear:
@jordansissel there we go, this fixes time stamps to work everywhere, let me know if you want to go with it or drop support for that :)

original-brownbear:
Urg this is failing because ... if we don't use annotations to set up the serialization of `Timestamp` then JrJackson somehow won't work on reloads and hence Http API calls bust ...
Let me know if we actually want this before I'll try to fix this fact :D

jordansissel:
Tested:

* Master

```
[1] pry(#<LogStash::Runner>)> a = org.logstash.Event.new; a.setField("foo", LogStash::Timestamp.now); puts org.logstash.Event.deserialize(a.serialize).getField("foo").class
String
```

* This PR

```
[49] pry(#<LogStash::Runner>)> a = org.logstash.Event.new; a.setField("foo", LogStash::Timestamp.now); puts org.logstash.Event.deserialize(a.serialize).getField("foo").class
Java::OrgLogstash::Timestamp
```

jordansissel:
LGTM

For the question "Should we support timestamp values in arbitrary fields" I think this PR is not the scope to make that decision. IMO, any value type restriction we impose should be reported during Event.setField() not as some side effect of serialization (Where a non-`@timestamp` Timestamp serializes to String).

For serialization, the symmetry of having `deserialize(serialize(value)) == value` is a good thing that should reduce confusion.

original-brownbear:
@jordansissel thanks, unfortunately, we have to figure out the problems JrJackson introduces here first before merging :( (See Travis, the reload ITs break because reloading makes JrJackson stop using our custom serializer).
But I have an idea to tackle this, on it :)

original-brownbear:
@jordansissel fixed it ITs. Had to keep the registering of the serializer and deserializer in the annotations for now due to #8154, no other changes since the last review other than that :)

original-brownbear:
@jordansissel good to merge here? The IT failure is handled in #8182 :)

original-brownbear:
@jordansissel ping ping ping, just rebased fresh :)

jordansissel:
The 'bck' commit message is not informative. Can you improve it or squash into the parent commit?

(The reason I mention is because at least one 'bck' commit has landed into master - 10007948c84679094df990ddf2943d663ef288df - and commit messages must be descriptive of the change and its purpose)

original-brownbear:
@jordansissel wops squashing :)

original-brownbear:
@jordansissel thanks!

PR #8135, opened 2017-09-04
[Docs] Replaced deprecated Grok pattern

ba0708:
Replaced the deprecated `COMBINEDAPACHELOG ` pattern with `HTTPD_COMBINEDLOG`.


kares:
if you're still concerned and can target this for master branch it would likely get merged ...

PR #8136, opened 2017-09-04
MINOR: Remove dead if branch from Vertex class

original-brownbear:
This was probably a leftover from some automatic refactoring, the method simply always returns true => this `if` is dead code.


jordansissel:
LGTM

PR #8137, opened 2017-09-05
MINOR: Fix a few more compiler warnings

original-brownbear:
Just stuff the compiler warns us about should be self explainatory with the exception of this part:

```diff
-        } else if (list instanceof ConvertedList) {		
-            clone = new ArrayList<>();		
         } else {		
```

This is simply never hit because `ConvertedList` is a child of `ArrayList` which is caught by the `instanceof ArrayList` check above => removed the dead code :)

and the change to `public static Event[] fromJson(String json)`:

* We should really use hard types here instead of `instanceof` checks for 2 reasons:
   * more compile time safety
   * `ClasscastException` with a proper is a lot easier to interpret that our homebrew `IOException`

Fixed about 30 warnings for #7701 :)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Cloner.java
+++ b/logstash-core/src/main/java/org/logstash/Cloner.java
@@ -1,6 +1,13 @@
 package org.logstash;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
 
 public final class Cloner {
 
@@ -23,8 +30,6 @@ public final class Cloner {
         if (list instanceof LinkedList<?>) {
             clone = new LinkedList<>();
         } else if (list instanceof ArrayList<?>) {
-            clone = new ArrayList<>();
-        } else if (list instanceof ConvertedList) {
             clone = new ArrayList<>();
         } else {
             throw new ClassCastException("unexpected List type " + list.getClass());
--- a/logstash-core/src/main/java/org/logstash/Event.java
+++ b/logstash-core/src/main/java/org/logstash/Event.java
@@ -2,7 +2,9 @@ package org.logstash;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import java.io.IOException;
+import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
@@ -56,7 +58,7 @@ public final class Event implements Cloneable, Queueable {
      * @param data Map that is assumed to have either {@link String} or {@link RubyString}
      * keys and may contain Java and Ruby objects.
      */
-    public Event(Map data) {
+    public Event(final Map<? extends Serializable, Object> data) {
         this(ConvertedMap.newFromMap(data));
     }
 
@@ -72,7 +74,8 @@ public final class Event implements Cloneable, Queueable {
         }
 
         if (this.data.containsKey(METADATA)) {
-            this.metadata = ConvertedMap.newFromMap((Map) this.data.remove(METADATA));
+            this.metadata =
+                ConvertedMap.newFromMap((Map<String, Object>) this.data.remove(METADATA));
         } else {
             this.metadata = new ConvertedMap(10);
         }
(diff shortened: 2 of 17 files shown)


logstash-core/src/main/java/org/logstash/instrument/witness/pipeline/PluginsWitness.java
```diff
@@ -89,8 +89,8 @@ public void forgetAll() {
      * @param id     the id of the plugin
      * @return existing or new {@link PluginWitness}
      */
-    private PluginWitness getPlugin(Map<String, PluginWitness> plugin, String id) {
-        return plugin.computeIfAbsent(id, k -> new PluginWitness(k));
+    private static PluginWitness getPlugin(Map<String, PluginWitness> plugin, String id) {
+        return plugin.computeIfAbsent(id, PluginWitness::new);
```
jakelandis: Was this actually a compiler warning, or just an IntelliJ default inspection warning ?

logstash-core/src/main/java/org/logstash/Cloner.java
```diff
@@ -1,6 +1,13 @@
...
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
```
jakelandis: Is this a compiler warning, or a style change ?

logstash-core/src/main/java/org/logstash/Event.java
```diff
@@ -56,7 +57,7 @@ public Event()
      * @param data Map that is assumed to have either {@link String} or {@link RubyString}
      * keys and may contain Java and Ruby objects.
      */
-    public Event(Map data) {
+    public Event(final Map<String, Object> data) {
```
jakelandis: The java doc claims either String or RubyString.  Does this break that contract ?

logstash-core/src/main/java/org/logstash/Event.java
```diff
@@ -209,15 +211,12 @@ public String toJson() throws JsonProcessingException {
...
         } else if (o instanceof List) {
-            result = new Event[((List) o).size()];
+            result = new Event[((Collection<?>) o).size()];
             int i = 0;
-            for (Object e : (List)o) {
-                if (!(e instanceof Map)) {
-                    throw new IOException("incompatible inner json array object type=" + e.getClass().getName() + " , only hash map is supported");
-                }
-                result[i++] = new Event((Map)e);
+            for (Map<String, Object> e : (Iterable<Map<String, Object>>) o) {
```
jakelandis: While correct, the code is a bit hard to read. We check if it is List, then upcast to a Collection , the iterate over it as a Iterable.  Can we just cast to List<?> once and use that variable through out ?

logstash-core/src/main/java/org/logstash/Event.java
```diff
@@ -226,13 +225,13 @@ public String toJson() throws JsonProcessingException {
...
-    public Map toMap() {
+    public Map<String, Object> toMap() {
         return Cloner.deep(this.data);
     }
 
     public Event overwrite(Event e) {
-        this.data = e.getData();
-        this.cancelled = e.isCancelled();
+        this.data = e.data;
+        this.cancelled = e.cancelled;
```
jakelandis: curious to why this change?  Don't we want to encourage the standard getter pattern ? (is this a compiler warning ?)

logstash-core/src/main/java/org/logstash/Event.java
original-brownbear: @jakelandis yea def! Will fix this soon :)

logstash-core/src/main/java/org/logstash/Event.java
original-brownbear: @jakelandis urgh you're right ... we could still have `RubyString` here, but type erasure prevents this from throwing obviously :( 
=> gotta change this to `? extends Serializable`

logstash-core/src/main/java/org/logstash/Cloner.java
original-brownbear: style change :) `.*` imports are just annoying

logstash-core/src/main/java/org/logstash/instrument/witness/pipeline/PluginsWitness.java
original-brownbear: @jakelandis hah seems like this was an idea bug, it didn't understand the type of `k` correctly in the EAP. Still much nicer this way imo :P

logstash-core/src/main/java/org/logstash/Event.java
original-brownbear: @jakelandis wops that randomly made it in here :) still, reduces the bytecode size of this method a little and since `Event` is `final` (-> getters provide no encapsulation in the context of `Event` itself) this is a fine cleanup imo :)

original-brownbear:
@jakelandis all either answered or fixed :)

jakelandis:
LGTM

PR #8140, opened 2017-09-05
PERFORMANCE: Move Memory Queue push loop into Java

original-brownbear:
No point in running the more expensive Ruby loop + block here when we can just do this in a Java loop => faster and (as can be seen in the spec I had to adjust) typesafe :)
Also no point in wrapping the Queue from the perspective of `WriteClient`, we can save that level of indirection here.


jordansissel:
I haven't verified the performance claim, but as this moves more core code to Java, I am in favor.

Tests passing. LGTM

PR #8143, opened 2017-09-05
DLQ Integration Test stabilization

robbavey:
Attempt to stabilize DLQ Integration tests
 - Only tear down ES instance once tests are complete
 - Delete indices after each test
 - Tear down Logstash after each test


robbavey:
This should improve stability, by fixing up the 503's that have been coming from Elasticsearch. There may still be failures - this test seems to be fairly susceptible to jruby failures.

original-brownbear:
@robbavey LGTM :)

Issue #8006, opened 2017-08-16
[meta] JRuby 9k stability


colinsurprenant:
This is a meta issue to track the JRuby 9k integration stability issues in LS 6. 

- [x] JRuby JIT issue jruby/jruby#4739, #7999, #8008, #8030

colinsurprenant:
I updated the description to reflect that all issues reported were in fact related to the same JRuby JIT issue and should be solved in upcoming 9.1.13.0.

I was going to suggest we test against a 9.1.13.0 snapshot but I can't find recent nightly builds.

headius:
Thank you for pointing out the missing 9.1.x snapshots. I believe this is simply an oversight; we're still building two HEAD builds, but it's master and jruby-1_7, not master and jruby-9.1.*. I'll remedy this.

headius:
The JRuby CI nightlies have been reworked.

The snapshots/current and snapshots/previous dirs now hold the current development line (usually master) and previous maintained version line (currently 9.1.x).

Builds of 9.1.13.0 are available at http://ci.jruby.org/snapshots/previous/.

Builds of 9.2 (master) are available at http://ci.jruby.org/snapshots/current/.

colinsurprenant:
Thanks @headius  - really useful!

original-brownbear:
fixed thanks to JRuby `9.1.13.0` :)

PR #8144, opened 2017-09-05
add option for different jruby runtime version

colinsurprenant:
Add support for an optional `jruby-runtime:` key in `versions.yml`. 

If `jruby-runtime:` is specified it will override the **runtime** JRuby version installed in `vendor/jruby`. 

Note that the `jruby:` key will still be used to reference the compile-time *provided* jars in `build.gradle`. 

This is useful to be able to use a JRuby pre-relase/snapshot version for the builds & specs or measure performance. 

This PR also sets the `jruby-runtime:` to  9.1.13.0-SNAPSHOT so we can start testing against the fix  for the JRuby JIT problem #8006

### DIFF
--- a/rakelib/vendor.rake
+++ b/rakelib/vendor.rake
@@ -60,9 +60,12 @@ namespace "vendor" do
   end # def untar
 
   task "jruby" do |task, args|
-    name = task.name.split(":")[1]
-    info = VERSIONS[name]
+    JRUBY = "jruby"
+    JRUBY_RUNTIME = "jruby-runtime-override"
+
+    info = VERSIONS[JRUBY_RUNTIME] || VERSIONS[JRUBY]
     version = info["version"]
+    url = info["url"] || "http://jruby.org.s3.amazonaws.com/downloads/#{version}/jruby-bin-#{version}.tar.gz"
 
     discard_patterns = Regexp.union([
       /^samples/,
@@ -72,10 +75,9 @@ namespace "vendor" do
       /lib\/ruby\/shared\/rdoc/,
     ])
 
-    url = "http://jruby.org.s3.amazonaws.com/downloads/#{version}/jruby-bin-#{version}.tar.gz"
     download = file_fetch(url, info["sha1"])
 
-    parent = vendor(name).gsub(/\/$/, "")
+    parent = vendor(JRUBY).gsub(/\/$/, "")
     directory parent => "vendor" do
       next if parent =~ discard_patterns
       FileUtils.mkdir(parent)
@@ -85,7 +87,7 @@ namespace "vendor" do
     untar(download) do |entry|
       out = entry.full_name.gsub(prefix_re, "")
       next if out =~ discard_patterns
-      vendor(name, out)
+      vendor(JRUBY, out)
     end # untar
   end # jruby
 
--- a/versions.yml
+++ b/versions.yml
@@ -2,6 +2,17 @@
 logstash: 7.0.0-alpha1
 logstash-core: 7.0.0-alpha1
 logstash-core-plugin-api: 2.1.16
+
+# jruby must reference a *released* version of jruby which can be downloaded from the official download url
+# *and* for which jars artifacts are published for compile-time
 jruby:
-  version: 9.1.12.0
-  sha1: 9b6c15d42eb11db7215d94dd9842ee5184a41fea
+  version: 9.1.13.0
+  sha1: 815bac27d5daa1459a4477d6d80584f007ce6a68
+
+# jruby-runtime-override, if specified, will override the jruby version installed in vendor/jruby for logstash runtime only,
+# not for the compile-time jars
+#
+#jruby-runtime-override:
+#  url: http://ci.jruby.org/snapshots/previous/jruby-bin-9.1.13.0-SNAPSHOT.tar.gz
+#  version: 9.1.13.0-SNAPSHOT
+#  sha1: IGNORE
--- a/logstash-core/logstash-core.gemspec
+++ b/logstash-core/logstash-core.gemspec
@@ -31,7 +31,7 @@ Gem::Specification.new do |gem|
 
   gem.add_runtime_dependency "sinatra", '~> 1.4', '>= 1.4.6'
   gem.add_runtime_dependency 'puma', '~> 2.16'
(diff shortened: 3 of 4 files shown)


original-brownbear:
@colinsurprenant LGTM (and works fine for me locally) I guess let's just wait for CI to be reasonably happy and merge :)

colinsurprenant:
This caused a jruby-ssl incompatibility due to the frozen version is logstash-core gemspec and the new version 0.9.21 introduced in JRuby 9.1.13.0 https://github.com/elastic/logstash/blob/c16c7027668bc767df578e6d01343e74705411b1/logstash-core/logstash-core.gemspec#L34

I think if we just relax it to ">=0.9.20" we should be good

original-brownbear:
@colinsurprenant let's try seems the compatibility matrix ends before this version so let's hope for the best :) https://github.com/jruby/jruby-openssl#compatibility

colinsurprenant:
ok, I don't see the jruby-ssl problem in the last integration tests failures. These looks unrelated (timeout) but I will respin them to make sure.

colinsurprenant:
After restarting the build, both integration tests still fail but one is on the slow logs test which seems unrelated and the other one is a timeout in the prepare-offline-pack test.

This does not feel related. Will investigate.

versions.yml
```diff
@@ -2,6 +2,16 @@
...
 jruby:
   version: 9.1.12.0
   sha1: 9b6c15d42eb11db7215d94dd9842ee5184a41fea
+
+# jruby-runtime, if specified, will override the jruby version installed in vendor/jruby for logstash runtime only,
+# not for the compile-time jars
+jruby-runtime:
+  url: http://ci.jruby.org/snapshots/previous/jruby-bin-9.1.13.0-SNAPSHOT.tar.gz
+  version: 9.1.13.0-SNAPSHOT
+  sha1: 47560ccae3e30f8193cb5b40b8deef4a39fa094e
```
jordansissel: Won't this break any time they publish a new snapshot with the same filename? (The checksum will be different)

rakelib/vendor.rake
```diff
@@ -60,9 +60,12 @@ namespace "vendor" do
   end # def untar
 
   task "jruby" do |task, args|
-    name = task.name.split(":")[1]
-    info = VERSIONS[name]
+    JRUBY = "jruby"
+    JRUBY_RUNTIME = "jruby-runtime"
+
+    info = VERSIONS[JRUBY_RUNTIME] || VERSIONS[JRUBY]
```
jordansissel: THis implicitly uses the "jruby-runtime" setting from the versions.yml, right? Why have two versions?

versions.yml
```diff
@@ -2,6 +2,16 @@
...
+# jruby must reference a *released* version of jruby which can be downloaded from the official download url
+# *and* for which jars artifacts are published for compile-time
 jruby:
   version: 9.1.12.0
   sha1: 9b6c15d42eb11db7215d94dd9842ee5184a41fea
+
+# jruby-runtime, if specified, will override the jruby version installed in vendor/jruby for logstash runtime only,
+# not for the compile-time jars
+jruby-runtime:
+  url: http://ci.jruby.org/snapshots/previous/jruby-bin-9.1.13.0-SNAPSHOT.tar.gz
```
jordansissel: Did you mean to include this? It seems like including this file in master (or any other branch) will cause JRuby snapshots to be the only version of jruby used during the build?

versions.yml
colinsurprenant: good point.

versions.yml
colinsurprenant: yes. the current released version is broken (JIT problem) and we want to start testing against the next version with will be released sometimes in the future which already includes the fix in the branch. this allows  the build to pass hopefully and once the version we want exists then we can remove the `jruby-runtime`. this is intended to be a temporary setting only when we specifically need/want to test agains a snapshot version.

rakelib/vendor.rake
colinsurprenant: because, as I tried to decribe in the new header comments I added for each section in the yml file, the `jruby` section is still needed for the gradle compile to fetch the required jars which are published artifacts. 

IF the `jruby-runtime` is specified, it is only use for the vendored version.

colinsurprenant:
2 problems following @jordansissel comments:
- I think the purpose for this is not well explained. 
- there is the issue that on every snapshot published by JRuby the sha hash will change and the build will fail.

For the former, `jruby-runtime` is **optional** and the idea is to provide a way of using a snapshot version for the vendored jruby that is used at runtime. This should typically be used and committed only in rare occasion, and not meant to be permanent, for when we want to have the builds run against a JRuby snapshot version. It could also be be useful in local development to do comparative (performance) tests against JRuby snapshots, but this wouldn't be committed.

For the sha hash problem, I am not sure how to deal with that.

jordansissel:
Maybe making it a bit more explicit of purpose, call it `jruby runtime override` instead of `jruby runtime` ?

For the idea, I'm in favor of being able to test against JRuby snapshots.

colinsurprenant:
for the sha hash I can only see 2 solutions:
- we simply ignore the hash for the `jruby-runtime` option. this option is not meant to be used in any released version so I believe the risk is low.
- we also fetch the sha hash every time we fetch the package but this would defeat the *security*  purpose of protecting against malicious intents  but would prevent corrupted downloads

WDYT?

jordansissel:
@colinsurprenant +1 to ignore the hash when using jruby-runtime setting.

colinsurprenant:
I have no idea if the integration tests problems are related or not. 

@jordansissel I updated the name to `jruby-runtime-override` and I added the `IGNORE` sha1 setting to make it explicit. LMKWYT.

colinsurprenant:
lol JRuby  9.1.13.0 has been released today, so I'll update this PR to point to the new JRuby and comment out the `jruby-runtime-override` which is not required in the short term :P (but still believe this patch could/will be useful in the future if not just for helping with comparative tests with pre-releases etc. )

colinsurprenant:
since much has changed since first LGTM I'd like another review please.

colinsurprenant:
Great. Only one integration build error on timeout, will go ahead and merge. thanks @jordansissel @original-brownbear

colinsurprenant:
merged in master, 6.x, 6.0

PR #8146, opened 2017-09-05
PERFORMANCE: Cleanup Rubyfier

original-brownbear:
Cleaned up `Rubyfier` to use the same approach that we also have in `Valuefier` and `BiValues` now.
See #8103 and #8087 for more on this.

This is supposed to not have any functional impact outside of one trivial point:
* I relaxed conversion to  of only `List<?>` to any `Collection<?>` to `RubyArray`, throwing on a non-`List` collection here made no sense and just blows up the code size.


jordansissel:
LGTM

Issue #6834, opened 2017-03-23
Custom Plugin not found


ohayak:
From: [https://discuss.elastic.co/t/custom-plugin-not-found/79565](url)
Im working this plugin to fetch logs from hdfs, this is the gemspec file:

      s.name          = 'logstash-input-webhdfs'
      s.version       = '0.1.0'
      s.licenses      = ['Apache License (2.0)']
      s.summary       = 'Write a short summary, because Rubygems requires one.'
      s.description   = 'Write a longer description or delete this line.'
      s.homepage      = 'http://localhost'
      s.authors       = ['xxxxxxx']
      s.email         = 'xxxxxxx@xxxxxx.xxx'
      s.require_paths = ['lib']

      # Files
      s.files = Dir['lib/**/*','spec/**/*','vendor/**/*','*.gemspec','*.md','CONTRIBUTORS','Gemfile','LICENSE','NOTICE.TXT']
       # Tests
      s.test_files = s.files.grep(%r{^(test|spec|features)/})

      # Special flag to let us know this is actually a logstash plugin
      s.metadata = { "logstash_plugin" => "true", "logstash_group" => "input" }

      # Gem dependencies
      s.add_runtime_dependency 'logstash-core'
      s.add_runtime_dependency "logstash-core-plugin-api"
      s.add_runtime_dependency 'webhdfs'

      s.add_development_dependency 'logstash-devutils'
      s.add_development_dependency 'logstash-codec-line'

I create the gem file using gem build, then I install it with logstash-plugin install commad.
finally I check if the plugin is installed witth logstash-plugin list command. 

When launching logstash this error popup : 

[...]

Issue #8142, opened 2017-09-05
ES DLQ Rats Test is very Unstable


original-brownbear:
This is happening in more than 50% of IT runs now ... https://travis-ci.org/elastic/logstash/jobs/272065977#L4133

will try to fix ...

robbavey:
@original-brownbear Feel free to assign this to me - I made it, so I can fix it

original-brownbear:
@robbavey enjoy :)

original-brownbear:
Seems better now with #8143 merged, closing :)

original-brownbear:
Still seeing this locally and just now on Travis (sorry no link I re-triggerd) (though at a lower rate now it seems)

robbavey:
Interesting error in one of the failed runs - 

```
Exception in thread "Ruby-0-Thread-10@[main]>worker0: :1" java.lang.IllegalArgumentException: Counters can not be incremented by negative values
	at org.logstash.instrument.metrics.counter.LongCounter.<clinit>(org/logstash/instrument/metrics/counter/LongCounter.java:15)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(java/lang/Class.java:348)
	at org.jruby.javasupport.JavaSupportImpl.loadJavaClass(org/jruby/javasupport/JavaSupportImpl.java:154)
	at org.jruby.javasupport.Java.getProxyClassOrNull(org/jruby/javasupport/Java.java:940)
	at org.jruby.javasupport.Java.getProxyClassOrNull(org/jruby/javasupport/Java.java:927)
	at org.jruby.javasupport.Java.getProxyOrPackageUnderPackage(org/jruby/javasupport/Java.java:885)
	at org.jruby.javasupport.JavaPackage.method_missing(org/jruby/javasupport/JavaPackage.java:259)
	at org.jruby.javasupport.JavaPackage$INVOKER$i$method_missing.call(org/jruby/javasupport/JavaPackage$INVOKER$i$method_missing.gen)
	at RUBY.<main>(:1)
	at org.jruby.Ruby.runInterpreter(org/jruby/Ruby.java:839)
	at org.jruby.Ruby.loadFile(org/jruby/Ruby.java:2876)
	at org.jruby.RubyKernel.requireCommon(org/jruby/RubyKernel.java:968)
	at org.jruby.RubyKernel.require(org/jruby/RubyKernel.java:961)
[...]

PR #8147, opened 2017-09-06
Log the exception if loading the plugin (via `require`) fails.

jordansissel:
Fixes: https://github.com/elastic/logstash/issues/6834

This fix was slightly complex (more than just logging the exception) because many of our tests create plugins at runtime which would fail with LoadError when legacy_lookup failed to find the predicted file path.

Good errors now, I think, or at least, better than just "NameError"

```
% bin/logstash -e 'input { fancy { } }'

[2017-09-05T17:18:46,988][ERROR][logstash.plugins.registry] Tried to load a plugin's code, but failed. {:exception=>#<LoadError: no such file to load -- logstash/inputs/fancy>, :path=>"logstash/inputs/fancy", :type=>"input", :name=>"fancy"}
[2017-09-05T17:18:47,008][ERROR][logstash.agent           ] Failed to execute action {:action=>LogStash::PipelineAction::Create/pipeline_id:main, :exception=>"LogStash::PluginLoadingError", :message=>"Couldn't find any input plugin named 'fancy'. Are you sure this is correct? Trying to load the fancy input plugin resulted in this error: no such file to load -- logstash/inputs/fancy"}
```

### DIFF
--- a/logstash-core/lib/logstash/plugins/registry.rb
+++ b/logstash-core/lib/logstash/plugins/registry.rb
@@ -9,6 +9,8 @@ module LogStash module Plugins
   class Registry
     include LogStash::Util::Loggable
 
+    class UnknownPlugin < NameError; end
+
     # Add a bit more sanity with when interacting with the rubygems'
     # specifications database, most of out code interact directly with really low level
     # components of bundler/rubygems we need to encapsulate that and this is a start.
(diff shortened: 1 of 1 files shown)


jordansissel:
^^ Test failures seem unrelated to this patch.

original-brownbear:
@jordansissel yea test fails are #8142 :) => LGTM

PR #8151, opened 2017-09-06
#8149 Fix Logstash::Agent to not leak its webserver thread

### LINKED ISSUE #8149: LogStash::Agent metrics after config reloading when reloading a bad config sets the failure reload timestamp
Seems we're leaking in these specs, error for ` LogStash::Agent metrics after config reloading when reloading a bad config sets the failure reload timestamp`:

```sh
19:43:51   1) LogStash::Agent metrics after config reloading when reloading a bad config sets the failure reload timestamp
19:43:51      Failure/Error: raise ExpiredTestDoubleError,
19:43:51        #<Double "logger"> was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.
19:43:51      # /var/lib/jenkins/workspace/elastic+logstash+6.x+multijob-unix-compatibility/os/opensuse/logstash-core/lib/logstash/webserver.rb:90:in `start_webserver'
19:43:51      # /var/lib/jenkins/workspace/elastic+logstash+6.x+multijob-unix-compatibility/os/opensuse/logstash-core/lib/logstash/webserver.rb:44:in `block in run'
19:43:51      # /var/lib/jenkins/workspace/elastic+logstash+6.x+multijob-unix-compatibility/os/opensuse/logstash-core/lib/logstash/webserver.rb:39:in `run'
19:43:51      # /var/lib/jenkins/workspace/elastic+logstash+6.x+multijob-unix-compatibility/os/opensuse/logstash-core/lib/logstash/agent.rb:393:in `block in start_webserver'
```

https://logstash-ci.elastic.co/job/elastic+logstash+6.x+multijob-unix-compatibility/os=opensuse/62/console

original-brownbear:
Fixes #8149 by preventing the thread that holds the web server from leaking between specs (most likely on exceptions).
Also, it's generally nice to join threads in a defined manner I guess :)

### DIFF
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -388,14 +388,20 @@ class LogStash::Agent
   def start_webserver
     options = {:http_host => @http_host, :http_ports => @http_port, :http_environment => @http_environment }
     @webserver = LogStash::WebServer.new(@logger, self, options)
-    Thread.new(@webserver) do |webserver|
+    @webserver_thread = Thread.new(@webserver) do |webserver|
       LogStash::Util.set_thread_name("Api Webserver")
       webserver.run
     end
   end
 
   def stop_webserver
-    @webserver.stop if @webserver
+    if @webserver
+      @webserver.stop
+      if @webserver_thread.join(5).nil?
+        @webserver_thread.kill
+        @webserver_thread.join
+      end
+    end
   end
 
   def configure_metrics_collectors


logstash-core/lib/logstash/agent.rb
```diff
@@ -388,14 +388,20 @@ def dispatch_events(converge_results)
...
       webserver.run
     end
   end
 
   def stop_webserver
-    @webserver.stop if @webserver
+    if @webserver
+      @webserver.stop
+      if @webserver_thread.join(5).nil?
+        @webserver_thread.kill
```
jordansissel: Thread.kill is likely to leave the server socket and any other resources hanging around until GC runs their finalizers/cleaners. This can mean that we kill the thread but the next attempt to run the webserver will fail because the socket is still in use.

Thoughts?

original-brownbear: @jordansissel I was hoping the internal workings of `Webserver` would deal with this?
(if not, I think resource safety on interrupt should be fixed there?)

original-brownbear: @jordansissel took a quick look at the Puma webserver code ... looks safe to me resource wise :)

original-brownbear:
@jordansissel ping :)

original-brownbear:
@jordansissel ping :)

original-brownbear:
putting it back to the review queue :)

PR #8155, opened 2017-09-06
PERFORMANCE: Remove WriteBatch, it's 100% Redundant

original-brownbear:
`WriteBatch` (apart from from `events` field accessor that I added a few days ago and that never made it to production) exposes no API that `[]` doesn't also expose.
=> We can simply drop the pointless wrapper and use an array a write batch.

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -343,7 +343,7 @@ module LogStash; module Util
       end
 
       def get_new_batch
-        WriteBatch.new
+        []
       end
 
       def push(event)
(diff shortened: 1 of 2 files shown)


andrewvc:
LGTM (pending passing CI)

andrewvc:
Also, good find!

guyboertje:
On the other hand, moving the WriteBatch and ReadBatch objects into Java is an option. They are both Set like structures, and we should enforce that, so the calling code does not have to. We discussed the idea of whether a Batch was more than an array about a year ago.

Arguably, the methods in LSQueueUtils could be moved to the WriteBatch (addAll) and ReadBatch (drain).

I think we will want to make WriteBatch a Queueable (I think Colin is working on this) so we can enqueue/dequeue them whole (making addAll and drain redundant) . The jury is still out on whether we want to include provenance information in each Event or the Batch. Provenance should include, where the batch came from and, perhaps, what decoding is required on it (if we want to move the decoding post queue).

We could also move the Read/Write Clients into Java too. They are Queue backed Batch factories after all.

Most of these higher level design decisions rest on what the Java Plugin API and the Java Pipeline Execution API ends up looking like.

In summary, I'm OK with removing this pointless wrapper for now but I suspect it will be back.

PR #8158, opened 2017-09-07
PERFORMANCE: Bring Javafier design in line with Rubyfier and Valuefier

original-brownbear:
Same as #8146 and related issues.

Sorry for removing the JDoc on `Valuefier.*` but it stopped applying + the code is pretty self-explanatory I think.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Javafier.java
+++ b/logstash-core/src/main/java/org/logstash/Javafier.java
@@ -1,5 +1,7 @@
 package org.logstash;
 
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import org.jruby.RubyBoolean;
 import org.jruby.RubyFixnum;
 import org.jruby.RubyFloat;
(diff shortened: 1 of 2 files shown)


logstash-core/src/main/java/org/logstash/Valuefier.java
```diff
@@ -145,18 +145,8 @@ private static Object fallbackConvert(final Object o, final Class<?> cls) {
         return converters;
     }
 
-    /**
-     * Converter from either a Java or a Ruby type to a type that both {@link Javafier} and
-     * {@link Rubyfier} are able to convert back their respective types efficiently.
-     */
-    private interface Converter {
+    public interface Converter {
```
jakelandis: Should this be explicitly declared as a `@FunctionalInterface`

original-brownbear: @jakelandis hmm I don't really have an opinion on this. To me, that annotation seems kind of redundant.

logstash-core/src/main/java/org/logstash/Javafier.java
```diff
@@ -8,63 +10,67 @@
...
+                return found.convert(o);
+            }
         }
+        CONVERTER_MAP.put(cls, BIVALUES_CONVERTER);
+        return BIVALUES_CONVERTER.convert(o);
+    }
+
+    private static Map<Class<?>, Valuefier.Converter> initConverters() {
+        final Map<Class<?>, Valuefier.Converter> converters =
+            new ConcurrentHashMap<>(50, 0.2F, 1);
```
jakelandis: More curious then anything...but since this is initialized in static scope, and only read, could this be a non regular `HashMap` and/or `Unmodifiable` ?

original-brownbear: @jakelandis unfortunately not because of the fallback that might modify this https://github.com/elastic/logstash/pull/8158/files#diff-46a0157ada034d162b510a0c9f9f03f9R38

... that said we could try an `IdentityHash` with copy-on-write at some point, but I don't think it makes much of a difference here.

logstash-core/src/main/java/org/logstash/Valuefier.java
jakelandis: @original-brownbear - I personally like it explicitly declared, the IDE will also complain louder is you try to add a second method.  

In my 3 minutes of googling, I didn't find any style guides stating one way or the other. 

..so dealers choice :)

jakelandis:
LGTM

Issue #8154, opened 2017-09-06
We an annoying cyclic dependency (Ruby <-> Java) because of `LogStash::Json` and `org.logstash.ext.JrubyEventExtLibrary#load`


original-brownbear:
Came across this twice the last few days.

`org.logstash.ext.JrubyEventExtLibrary#load` requires `logstash/json.rb` to be loaded because of the second half of its call stack:

```java
        PARSER_ERROR = module.defineOrGetModuleUnder("Json").getClass("ParserError");
        if (PARSER_ERROR == null) {
            throw new RaiseException(runtime, runtime.getClass("StandardError"), "Could not find LogStash::Json::ParserError class", true);
        }
        GENERATOR_ERROR = module.defineOrGetModuleUnder("Json").getClass("GeneratorError");
        if (GENERATOR_ERROR == null) {
            throw new RaiseException(runtime, runtime.getClass("StandardError"), "Could not find LogStash::Json::GeneratorError class", true);
        }
        LOGSTASH_ERROR = module.getClass("Error");
        if (LOGSTASH_ERROR == null) {
            throw new RaiseException(runtime, runtime.getClass("StandardError"), "Could not find LogStash::Error class", true);
        }
```

This is kind of a dirt anti-pattern  for one, but this is also causing some practical trouble in writing better (faster to execute  + easier to debug mostly) tests for the Java execution:

* You can't actually load `JrubyEventExtLibrary` in Java UTs, so not all functionality of the Ruby Event can be tested from Java

My suggestion would be to simply move `json.rb` to Java completely and drop JrJackson, this gives us a few benefits:

* We don't break timestamp serialization on reloading the pipeline because we have full control of all `ObjectMapper`s from the Java side (https://github.com/guyboertje/jrjackson/issues/62)
   * would allow me to fix #8128 instantly :)
* JrJackson won't leak afterburner into out classpath anymore
* We fix the cyclic dependency (better Java tests + cleaner code)
* We can finally upgrade Jackson dep. versions as we please :)

jakelandis:
@suyograo - we should prioritize dependency upgrade.

PR #8161, opened 2017-09-07
Java Entrypoint

original-brownbear:
Java entry point:

* Created `org.logstash.Logstash` as entrypoint
* Safely handle `Ruby` runtime (which sadly is still a singleton, moving away from that will require a few iterations on top of this)
* Adjusted `bat` and `sh` entry point wrappers
* Verified manually that performance is unchanged (i.e. all Java opts are still loaded properly)
* Flattened `.jar` path to make it a little less bothersome to build the `-cp` string
* Retained ability to load jars from Ruby via the global `$LS_JARS_LOADED` variable hack, to keep plugin specs that load LS as a `.gem` functional (like e.g. the ITs in LS itself)
* No need for the gem jars magic anymore, the downloading and moving into place of jars is now all handled by Gradle

Description incoming

Works on Windows:

screen shot 2017-11-14 at 18 44 06


original-brownbear:
should work out of the box with `/bin/logstash` after running `./gradlew clean assemble` btw

logstash-core/src/main/java/org/logstash/Main.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.InputStream;
+import org.jruby.Ruby;
+import org.jruby.RubyException;
+import org.jruby.RubyInstanceConfig;
+import org.jruby.RubyNumeric;
+import org.jruby.exceptions.RaiseException;
+import org.jruby.runtime.builtin.IRubyObject;
+
+public final class Main implements Runnable {
```
jordansissel: Let's call this `Logstash` so that when folks do `jps` "Logstash" is what shows up instead of "Main"

jordansissel:
The last time we had lots of ruby code (aka jruby-complete) living in a jar, it had really bad effects on Logstash startup time because (at the time) `require` from JRuby loading ruby code from a .jar was very slow. We accidentally worked around it by having jruby ruby source as files when we moved away from the monolithic jar (years ago).

andrewvc:
This is awesome.

What I think this needs to move to completion is a wrapper for arbitrary ruby tasks like `rake` and `bin/rspec`, + adaptation of the Gradle stuff to use this.

I'd rather move the whole thing over than just the main logstash entrypoint for environment consistency.

original-brownbear:
@andrewvc sure we can do that, already figured out how and played with that a little, will add that here soon. @jordansissel 's point is valid though, we probably shouldn't pack the `.rb` files into the `.jar` to not make the start-up crazy slow.
Also, just as a note for myself, I need to add the `jvm.properties` file handling to this thing, those settings are currently ignored making performance pretty meh :)

original-brownbear:
@andrewvc just to summarize what we talked about:

* I think moving the rake tasks to Gradle will be tough to do at this point. I'm seeing some conflicts between Gradle and the Ruby Maven gem (at least I think that's the root cause of all issues) that make building LS from only the Gradle JVM very tricky and hard to maintain.
* The solution here has some issues with the load order, that only do not cause harm accidentally. I will have to do some more work here and at least fix #8154 before we will get value (mainly thinking about JUnit tests here) out of moving to a Java entrypoint.

original-brownbear:
@jordansissel hah I learned a new thing :) Apparently the only reason loading `.rb`s from a Jar comes with a hit is the compression. Turning off the Jar compression via https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/bundling/ZipEntryCompression.html#STORED should make packing everything into a single jar as fast as reading straight from the FS (if not faster, since it means only loading a single file instead of hundreds).

original-brownbear:
@andrewvc this version packages and works just fine (but there are a few open questions around the effects this will have on the IT and plugin test infrastructures ... ITs are red because loading Java first breaks the transitive including of jars for the ITs that include core a `gem`. But easy to fix, just a question of what we want to achieve).

Let's look into this whenever you have some time next week:)

jordansissel:
@original-brownbear I"m open to reviewing this; lemme know when you think it's ready for testing and review :)

original-brownbear:
@jordansissel thanks, hopefully today, will ping you then :)

original-brownbear:
@jordansissel alright, now it should be good to review :)

original-brownbear:
Jenkins test this please

original-brownbear:
Sorry, for some reason the Kafka IT is broken now, will have to look into that and ping once it's green.

original-brownbear:
@jordansissel ok green at last :) Take a look whenever you have time, I won't touch it again :D

original-brownbear:
@andrewvc as discussed, this one is a little important for the clustering work.
It's not a blocker, but it would allow much more productive development by enabling in JVM integration tests of multiple LS instances.

jordansissel:
This is on my todo list for this week to test/review further.

logstash-core/src/main/java/org/logstash/Logstash.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Main Entrypoint.
+     * Requires environment {@code "LS_HOME"} to be set to the Logstash root directory.
+     * @param args Logstash CLI Arguments
+     */
+    public static void main(final String... args) {
+        final String lsHome = System.getenv("LS_HOME");
+        if (lsHome == null) {
+            throw new IllegalStateException("LS_HOME environment variable must be set.");
```
jordansissel: Can we make this error message more actionable? If a user sees this exception, what should they do to resolve it, and can we include hints for resolving this in the message?

logstash-core/src/main/java/org/logstash/Logstash.java
```diff
@@ -0,0 +1,144 @@
...
+     * @param subs Path elements relative to {@code home}
+     * @return Absolute Path a File under the Logstash Root.
+     */
+    private static String safePath(final Path home, final String... subs) {
+        Path resolved = home;
+        for (final String element : subs) {
+            resolved = resolved.resolve(element);
+        }
+        if (!resolved.toFile().exists()) {
+            throw new IllegalArgumentException(String.format("Missing: %s.", resolved));
```
jordansissel: Can we make this more actionable, or would a user never expect to see this?

logstash-core/src/main/java/org/logstash/Logstash.java
```diff
@@ -0,0 +1,144 @@
...
+     * Main Entrypoint.
+     * Requires environment {@code "LS_HOME"} to be set to the Logstash root directory.
+     * @param args Logstash CLI Arguments
+     */
+    public static void main(final String... args) {
+        final String lsHome = System.getenv("LS_HOME");
+        if (lsHome == null) {
+            throw new IllegalStateException("LS_HOME environment variable must be set.");
+        }
+        final Path home = Paths.get(lsHome).toAbsolutePath();
```
jordansissel: Can we verify some properties about this path before using it: Does the path exist? Does it have things we want in it? etc..

My thought is we can do a quick check here and provide actionable response to the user if the check fails.

jordansissel:
> should work out of the box with /bin/logstash after running ./gradlew clean assemble btw

```
% ./gradlew clean assemble
...
% bin/logstash -e ''
Unable to find JRuby.
If you are a user, this is a bug.
If you are a developer, please run 'rake bootstrap'. Running 'rake' requires the 'ruby' program be available.
```

jordansissel:
If I try to work around the above error (can't find jruby) with this patch:

```diff
diff --git a/bin/logstash.lib.sh b/bin/logstash.lib.sh
index bb9090808..7f01bcd32 100755
--- a/bin/logstash.lib.sh
+++ b/bin/logstash.lib.sh
@@ -120,7 +120,7 @@ setup_vendored_jruby() {

 setup() {
   setup_java
-  setup_vendored_jruby
+  #setup_vendored_jruby
 }

 ruby_exec() {
```

Then run `bin/logstash`, it appears to do nothing?

```
% time bin/logstash --log.level debug -e 'input { generator { count => 1 } }'
bin/logstash --log.level debug -e 'input { generator { count => 1 } }'  11.93s user 0.84s system 292% cpu 4.366 total
```

Running `bin/logstash` with `sh -x` to see what's going on, I see this as the command, if this helps:

```
+ exec /bin/java -Xms1g -Xmx1g -XX:+UseParNewGC -XX:+UseConcMarkSweepGC -XX:CMSInitiatingOccupancyFraction=75 -XX:+UseCMSInitiatingOccupancyOnly -Djava.awt.headless=true -Dfile.encoding=UTF-8 -Djruby.compile.invokedynamic=true -Djruby.jit.threshold=0 -XX:+HeapDumpOnOutOfMemoryError -Djava.security.egd=file:/dev/urandom -cp /home/jls/projects/logstash/logstash-core/lib/jars/commons-compiler-3.0.7.jar:/home/jls/projects/logstash/logstash-core/lib/jars/jackson-annotations-2.9.1.jar:/home/jls/projects/logstash/logstash-core/lib/jars/jackson-core-2.9.1.jar:/home/jls/projects/logstash/logstash-core/lib/jars/jackson-databind-2.9.1.jar:/home/jls/projects/logstash/logstash-core/lib/jars/jackson-dataformat-cbor-2.9.1.jar:/home/jls/projects/logstash/logstash-core/lib/jars/janino-3.0.7.jar:/home/jls/projects/logstash/logstash-core/lib/jars/jruby-complete-9.1.13.0.jar:/home/jls/projects/logstash/logstash-core/lib/jars/log4j-api-2.9.1.jar:/home/jls/projects/logstash/logstash-core/lib/jars/log4j-core-2.9.1.jar:/home/jls/projects/logstash/logstash-core/lib/jars/log4j-slf4j-impl-2.9.1.jar:/home/jls/projects/logstash/logstash-core/lib/jars/logstash-core.jar:/home/jls/projects/logstash/logstash-core/lib/jars/slf4j-api-1.7.25.jar org.logstash.Logstash --log.level debug -e 'input { generator { count => 1 } }'
```

I'm not sure how to test this as `bin/logstash` appears to do nothing. Any ideas?

original-brownbear:
@jordansissel hmm don't waste your time apparently there is an issue here now. It was green a few weeks ago when I pinged, but I only rebased today without rerunning tests. Seems Jenkins isn't going anywhere as well. Probably just some trivial packaging issue. Will fix tomorrow morning :)

jordansissel:
@original-brownbear noted! :)

bin/logstash
```diff
@@ -58,5 +58,8 @@ if [ "$1" = "-V" ] || [ "$1" = "--version" ]; then
   fi
   echo "logstash $LOGSTASH_VERSION"
 else
-  ruby_exec "${LOGSTASH_HOME}/lib/bootstrap/environment.rb" "logstash/runner.rb" "$@"
+  function join_cp { local IFS=":"; echo "$*"; }
+  jars=(${LOGSTASH_HOME}/logstash-core/lib/jars/*.jar)
```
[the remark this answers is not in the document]
original-brownbear: @jordansissel I'd go for 3 here.

1. Is really hacky (and I think even problematic legally in some cases) since we have to flatten all jars (or do some other hack that deals with nested jars) into one which entails removing the checksums from their `META-INF` paths => not something we want to do imo.
2. I don't have a good theoretical argument against this, but the build is already quite complicated isn't it? I'd rather not pile onto that :) Plus I was just able to remove the generated files like logstash_core_jars.rb from the production code paths, I'd rather remove these hacks and not just add a new generated piece of code.

=> `sh` is def more portable than `bash`, let me try your code :) thanks

logstash-core/lib/logstash-core/logstash-core.rb
```diff
@@ -2,36 +2,11 @@
...
-  # if in local dev setup, add target to classpath
-  $CLASSPATH << classes_dir unless $CLASSPATH.include?(classes_dir)
-  $CLASSPATH << resources_dir unless $CLASSPATH.include?(resources_dir)
-else
-  # otherwise use included jar
-  begin
-    require "logstash-core/logstash-core.jar"
-  rescue Exception => e
-    raise("Error loading logstash-core/logstash-core.jar file, cause: #{e.message}")
+unless $LS_JARS_LOADED
```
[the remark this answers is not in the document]
original-brownbear: @jordansissel can't do it :( `org.logstash.Logstash` can't load without loading the jars beforehand (doing this kind of thing from Java and then dynamically loading the jars would require a massive hack I think). This we really only need for having the `bin/rspec` tool still work.
If we are fine with dropping that tool we don't need it, but I'd rather keep it around so I can debug Ruby comfortably from the IDE :)

logstash-core/src/main/java/org/logstash/Logstash.java
original-brownbear: @jordansissel hmm I think the user can't do anything they see this message (other than open an issue here). This being thrown means there is a bug in the `bin/logstash` or `bin/logstash.bat`. I simply put this check here for our convenience to have a clean failure instead of some other failure with a strange `null` in some string.

logstash-core/src/main/java/org/logstash/Logstash.java
original-brownbear: @jordansissel same as the above, this will actually throw something like "file not found" if the path is wrong which we will understand, for the user this means there is a bug in the startup script in 100% of cases imo => nothing actionable they can do right?

logstash-core/src/main/java/org/logstash/Logstash.java
original-brownbear: @jordansissel jup same here => user should never ever see this except for when there's a bug :)

bin/logstash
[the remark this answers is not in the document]
original-brownbear: @jordansissel hmm your solution works like a charm, but unfortunately only on newer `sh`. On vanilla OSX with `sh 3.2` the `-n` is not understood by `echo` and the solution will only work with `bash`.
But you know what `bash` must be fine for us ... we are currently calling `jruby/bin/jruby` from our scripts and it's a `bash` script too => we're only compatible if `bash` is available right now anyways.
=> Use your solution but with `bash` to work on systems with older `sh`? :)

original-brownbear:
@jordansissel with your change to the startup script it works fine for me + on Jenkins (failure is unrelated).
Should be fine now. Making `assemble` work out of the box isn't so easy though I'm afraid :( We are currently calling that target from `rake` and making that work would require cleaning up the `gradle` and `rake` interaction some more first (we should do that asap imo, but not here, this one already is pretty long).
For now `./gradlew assembleTarDistribution` should work fine though :)

logstash-core/src/main/java/org/logstash/Logstash.java
jordansissel: > other than open an issue here

Ok cool. Let's tell them it's probably a bug in the error message.

jordansissel:
@original-brownbear noted, will test again. Thanks!

jordansissel:
Manual test:

````
% rake artifact:tar
...
% bin/logstash -e ''
...

[separate terminal]
% jps
56479 Logstash
```

It's good to see `Logstash` here. Woo :)

jordansissel:
Tests passing locally.

logstash-core/src/main/java/org/logstash/Logstash.java
original-brownbear: @jordansissel isn't probably a bug implied in this being a `RuntimeException`? :)

jordansissel: As a user, I don't personally associate "this is a bug" with RuntimeException. I'd rather be explicit.  Saying "must be set" implies to me that the user should take action to set LS_HOME, which likely isn't the case. Otherwise, a user will end up struggling with this action trying to figure out how, and what value, to set LS_HOME.

bin/logstash
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
jordansissel: this should still be /bin/sh

original-brownbear: @jordansissel `bash` is more portable here imo :( Your code only works with `sh` `4.x` => `bash` `3.x` works fine though => standard OSX and old Linux will break if we go with `sh`   => `bash` wins, especially since JRuby requires `bash` anyways in the way we currently used to invoke it? :)

logstash-core/logstash-core.gemspec
```diff
@@ -53,7 +53,7 @@ Gem::Specification.new do |gem|
 
   gem.add_runtime_dependency "sinatra", '~> 1.4', '>= 1.4.6'
   gem.add_runtime_dependency 'puma', '~> 2.16'
-  gem.add_runtime_dependency "jruby-openssl", ">= 0.9.20" # >= 0.9.13 Required to support TLSv1.2
+  gem.add_runtime_dependency "jruby-openssl", ">= 0.9.21" # >= 0.9.13 Required to support TLSv1.2
```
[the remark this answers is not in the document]
original-brownbear: @jordansissel ha good question so many weeks later :D I think I fixed the problem that this solved elsewhere already => let me revert this :)

bin/logstash
jordansissel: Hmm, fair. I didn't realize that bin/jruby invoked bash.

original-brownbear:
@jordansissel all points addressed I think/hope :)

jordansissel:
Yep! I still need to test on windows. Should be done building shortly :)

original-brownbear:
@jordansissel see PR description at least for me Windows worked fine :D *fingers crossed :P

jordansissel:
@original-brownbear 🤦‍♂️  I missed the screenshot. LGTM

original-brownbear:
@jordansissel thanks! :)

PR #8163, opened 2017-09-07
PERFORMANCE: Flatten logic for write batch instantiation

original-brownbear:
Follow up to #8155:

The `get_new_batch` methods on the memory and the acked queue are both internal APIs and completely redundant since plugins only see `WrappedWriteClient`.
=> flattened this out

Also ... this actually comes with a measurable, statistically significant gain of `~3%` on the baseline (likely removing one step of stack depth got us over the barrier to not be affected by https://github.com/jruby/jruby/issues/4763 for `RubyArray` here).

### DIFF
--- a/logstash-core/lib/logstash/instrument/wrapped_write_client.rb
+++ b/logstash-core/lib/logstash/instrument/wrapped_write_client.rb
@@ -20,7 +20,7 @@ module LogStash module Instrument
     end
 
     def get_new_batch
-      @write_client.get_new_batch
+      []
     end
 
     def push(event)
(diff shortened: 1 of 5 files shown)


jordansissel:
@guyboertje can you review? My thoughts are that this goes backwards with respect to some anticipated-future-work where we have a firmer API around batches.

guyboertje:
It does look like it removes some anticipated work. I am satisfied that, when the future arrives, we will do a better job in Java for the Queue clients and Batches.

original-brownbear:
@guyboertje thanks ! :)

PR #8165, opened 2017-09-07
MINOR: Adjust Ruby Version File

original-brownbear:
Just so my IDE doesn't go crazy from the mixed version :)


original-brownbear:
@ph thanks :), I think we need to ping Infra here, Jenkins fails because it doesn't have `9.13` available :(

jakelandis:
Jenkins test this please

jakelandis:
Reverted on #8219

PR #8167, opened 2017-09-07
Temporarily skip dlq tests until stability is improved

robbavey:


jordansissel:
Maybe note whether the stability problem is with DLQ itself or only with the test suite? One may be alarmed without more details ;)

robbavey:
@jordansissel I think there are elements of both :( - the closing and starting up of elasticsearch which triggers 503 are likely to be environmental, and, I suspect, can be resolved through judicious retrying. 

There also a couple of exceptions that have been thrown by Logstash in a couple of instances, which I'm trying to reproduce locally. I'll make separate issues for those

original-brownbear:
@robbavey lol even after re-triggering these tests, I see Travis complain about a seemingly unrelated error consistently. This can't be related, can it? (just to sanity check)

robbavey:
@original-brownbear wtf?

If adding `:skip => true` to the dlq_spec causes that failure, then...  ¯\_(ツ)_/¯

original-brownbear:
@robbavey my point exactly :D I just wanted to have another person double check.
If it goes green now feel free to merge :) I'll add an issue in the morning. If it's red again ... ¯_(ツ)_/¯

robbavey:
@original-brownbear I'm pretty sure the test failures are unrelated to this commit. My current thought is that they might have been brought in by #8147 - I suspect that these failures were being hidden by the DLQ errors - I don't think the offline specs get run if the online specs fail :(

original-brownbear:
@robbavey lol https://travis-ci.org/elastic/logstash/jobs/272995453#L4010 same error again ... (6 times in a row now). Will check this out locally to see what's going on ...

original-brownbear:
@robbavey I can reproduce this error locally with `master` as well and seems this is just a load order thing randomly triggered by this => fix in  https://github.com/elastic/logstash/pull/8182 :)

kares:
let's close this, temporarily :stuck_out_tongue_winking_eye:

Issue #7824, opened 2017-07-26
DLQ Writer `close` Shouldn't be Synchronized


original-brownbear:
`synchronized` `close()` methods are kind of a nasty anti-pattern as explained here https://github.com/elastic/logstash/pull/7822/files#r129673727.

You simply don't want to block threads on concurrent `close()` when handling exceptions.
I think securing this by an `AtomicBoolean` (details here https://github.com/elastic/logstash/pull/7822/files#r129673727) is a much safer approach and we should refactor this.

original-brownbear:
cc @robbavey

liketic:
@original-brownbear Seems like this has been fixed. Should be closed?

original-brownbear:
@liketic right thanks!

PR #8171, opened 2017-09-07
MINOR: Some DLQ Stability Fixes

original-brownbear:
@robbavey let's see if this helps to track down the instabilities here:

Top to bottom:

* Don't quietly swallow deserialization issues, if you run into this your DLQ file is corrupted as far as I can see. We can't just quietly ignore this, even logging is not ok imo, since this functionality should keep data safe and a log entry can easily go unnoticed + continuing to execute on a corrupted queue can have unforeseeable side effects.
* Handle closing the queue in a thread-safe manner (see #7824 ), also just closing the `Channel` here is much safer (note: closing channel releases lock by definition), because we prevent loosing a reference to the channel. Currently, we leak the channel fd if opening the channel works, but locking it fails. This way it's thread safe in every case imo.
* Don't quietly swallow if the size of the Queue file can't be determined
* Cleanup, not stability: Moved the weird metadata access to a clean `event.includes` call
* `new RecordIOWriter` must be assigned to the instance variable (which must be `volatile` to be safe) right away, otherwise we can leak that resource on failure in the time period in which we only hold a local reference to it => fixed that
* Use `if (entryTimestamp.getTime().isBefore(lastEntryTimestamp.getTime())) {` to ensure thread safety. This cannot ever fail unless we have a concurrency issue => adjusted things as such to get some visibility on those if they exist

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
@@ -19,22 +19,23 @@
 package org.logstash.common.io;
 
 import java.io.Closeable;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import org.logstash.DLQEntry;
-import org.logstash.Event;
-import org.logstash.Timestamp;
-
 import java.io.IOException;
 import java.nio.channels.FileChannel;
-import java.nio.channels.FileLock;
 import java.nio.channels.OverlappingFileLockException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.stream.Stream;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.logstash.DLQEntry;
+import org.logstash.Event;
+import org.logstash.FieldReference;
+import org.logstash.PathCache;
+import org.logstash.Timestamp;
 
 import static org.logstash.common.io.RecordIOWriter.RECORD_HEADER_SIZE;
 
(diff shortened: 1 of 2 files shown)


robbavey:
@original-brownbear I think these changes mostly make sense - I'll take a closer look later tonight or tomorrow morning - but I don't _think_ they are the root cause behind most of the DLQ instability - most of the errors I've seen are around the metrics issue, and handling Elasticsearch startup and shutdown

original-brownbear:
@robbavey yea these won't fix anything, but they'll get us deterministic behavior on crashes :)

original-brownbear:
@robbavey hmm we just ran into a statistical anomaly here it seems, I can reproduce this problem locally as well.
Will merge this and track the other thing down next.

original-brownbear:
@robbavey nvm wrong thread

original-brownbear:
@robbavey ping, did you take a closer look? :)

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -125,7 +125,7 @@ public synchronized void writeEntry(DLQEntry entry) throws IOException {
     public synchronized void writeEntry(Event event, String pluginName, String pluginId, String reason) throws IOException {
         Timestamp entryTimestamp = Timestamp.now();
         if (entryTimestamp.getTime().isBefore(lastEntryTimestamp.getTime())) {
-            entryTimestamp = lastEntryTimestamp;
+            throw new IllegalStateException("Detected concurrency issue in DLQ Writer Access.\n");
```
robbavey: Is this true? Isn't Timestamp.now subject to clock variations, and not guaranteed to be monotonic?

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
```diff
@@ -160,29 +160,27 @@ private void innerWriteEntry(DLQEntry entry) throws IOException {
      * @return boolean indicating whether the event is eligible to be added to the DLQ
      */
     private static boolean alreadyProcessed(final Event event) {
-        return event.getMetadata() != null && event.getMetadata().containsKey(DEAD_LETTER_QUEUE_METADATA_KEY);
+        return event.includes(DEAD_LETTER_QUEUE_METADATA_KEY);
```
robbavey: TIL about event.includes... ;)

logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueWriter.java
original-brownbear: @robbavey yikes right ... sec reverting thought we were on nanotime here but obviously not :)

robbavey: @original-brownbear Sadly nanotime is only used for elapsed time, not for system or wall-clock time :(

original-brownbear: @robbavey yea ... there wasn't a good reason to assume nano time really :D

original-brownbear:
@robbavey ping :)

PR #8175, opened 2017-09-07
	Handle Elasticsearch errors better during DLQ integration tests

robbavey:
Give more time for Elasticsearch to start up, and retry after 503 responses

### DIFF
--- a/qa/integration/specs/dlq_spec.rb
+++ b/qa/integration/specs/dlq_spec.rb
@@ -42,8 +42,6 @@ describe "Test Dead Letter Queue" do
     it 'should index all documents' do
       es_service = @fixture.get_service("elasticsearch")
       es_client = es_service.get_client
-      # Wait for es client to come up
-      sleep(20)
       # test if all data was indexed by ES, but first refresh manually
       es_client.indices.refresh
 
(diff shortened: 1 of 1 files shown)


robbavey:
Jenkins test this please

robbavey:
This should fix the environmental Elasticsearch issues, but not the overall test timeouts...

qa/integration/specs/dlq_spec.rb
```diff
@@ -43,14 +43,20 @@
       es_service = @fixture.get_service("elasticsearch")
       es_client = es_service.get_client
       # Wait for es client to come up
-      sleep(15)
+      sleep(20)
```
original-brownbear: @robbavey I wonder if we could just busy wait here instead? We should be able to wrap `es_client.indices.refresh` in a `try(20) { es_client.indices.refresh }` block shouldn't we?
I'm mainly asking because I've also seen the block below that you set from `75` to `60` time out on Travis, so I'm a little hesitant on just making that shorter.

robbavey:
@original-brownbear ping ;)

original-brownbear:
@robbavey LGTM :)

PR #8178, opened 2017-09-08
windows launch scripts

### LINKED ISSUE #8126: have bin/setup.bat apply the same logic as bin/logstash.lib.sh
Under Windows, `setup.bat` is the equivalent of `bin/logstash.lib.sh` but `setup.bat` has not been updated with the enhancements such as using `config/jvm.options`. This means that under Windows, `config/jvm.options` is not taken into account when launching logstash or running the specs.

colinsurprenant:
fixes #8126 

This is mostly a rewrite of the windows launching batch files, inspired by the ES windows launch batch files.

The initial goal was to add support for the `config/jvm.options` in its default location or in `--path.settings` if specified on the command line. 

Here are the differences from the linux launching logic:
- existing `JAVA_OPTS` env var is now ignored, [ES is also doing this](https://github.com/elastic/elasticsearch/blob/33faf5ec70a29177eb50eccded4e2f103f938fc8/distribution/src/main/resources/bin/elasticsearch-env.bat#L37-L42)
- `JAVA_TOOL_OPTIONS` env var is cleared,  [ES is also doing this](https://github.com/elastic/elasticsearch/blob/33faf5ec70a29177eb50eccded4e2f103f938fc8/distribution/src/main/resources/bin/elasticsearch-env.bat#L31-L35)
- now only `LS_JAVA_OPTS` env var is supported to append options to parsed options in the `jvm.options` file.
- to simplify the rewrite I dropped the support of the `USE_RUBY` and `USE_DRIP` env vars. These were meant for development purposes and are not used anymore I believe. 

This also includes a fix in `runner.rb` which was creating an invalid URI for the path specified for `--path.settings` command line option.

### DIFF
--- a/bin/logstash.bat
+++ b/bin/logstash.bat
@@ -1,26 +1,51 @@
 @echo off
+setlocal enabledelayedexpansion
+set params='%*'
 
-SETLOCAL
-
-set SCRIPT_DIR=%~dp0
-CALL "%SCRIPT_DIR%\setup.bat"
-
-:EXEC
-REM is the first argument a flag? If so, assume 'agent'
-set first_arg=%1
-setlocal EnableDelayedExpansion
-if "!first_arg:~0,1!" equ "-" (
-  if "%VENDORED_JRUBY%" == "" (
-    %RUBYCMD% "%LS_HOME%\lib\bootstrap\environment.rb" "logstash\runner.rb" %*
-  ) else (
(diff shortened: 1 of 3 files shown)


untergeek:
I will test this in my Windows 10 environment when I get a chance

jordansissel:
Tested on Windows 10 - `bin/logstash -e 'input { stdin { } }'` works.

I uncommented a few arbitrary things in config/jvm.options and ran logstash -- all working :)

colinsurprenant:
merged into master, 6.x, 6.0, 5.6

Issue #8172, opened 2017-09-07
Metrics should not use System.currentTimeMillis to measure time intervals


robbavey:
Running the DLQ integration tests to track down intermittent test failures with attempting to increment metrics by negative values (#8168) when Exception stack traces were reinstated, the DLQ test failure where a metric was incremented by a negative value was reproduced, with the following stack trace:

```
[...]

PR #8180, opened 2017-09-08
#8172 Move timestamp handling for metrics to java.lang.System.nano_time

original-brownbear:
As discussed in #8172 moving to `nano_time` to improve correctness.

### DIFF
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -44,9 +44,9 @@ module LogStash
     def multi_filter(events)
       @metric_events_in.increment(events.size)
 
-      start_time = java.lang.System.current_time_millis
+      start_time = java.lang.System.nano_time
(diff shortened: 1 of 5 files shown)


jakelandis:
@original-brownbear - I agree this solves the issue over at #8172 , there seems to be a larger issue here (present before this change) that any action that takes less then 1ms will essentially be lost, throwing off the metrics.  

Would a better approach here be to increment the nano seconds in the metrics, and then wait to normalize to ms only upon a get/snitch/serialize ?

original-brownbear:
@jakelandis hmm, I mean we can obviously fix that by using `ns` for all metric updates. Seems kind of dirty though that we have these spots in the first place. But probably hard to fix ..., let's add an issue for this I guess :)

jakelandis:
@original-brownbear  - LGTM 

Logged https://github.com/elastic/logstash/issues/8184 to followup.

PR #8185, opened 2017-09-08
Add longer sleep for slowlog test to try and fix IT failures

robbavey:


original-brownbear:
@robbavey how does this fix the test, doesn't this only increase the period of checking if LS exited not the overall time we wait for it to exit?

robbavey:
@original-brownbear Ugh, you're right. I was trying to fix this test failure, and had a superficial look - tried this out, and it worked locally and a couple of test runs - but going to abandon and look elsewhere, because it shouldn't

Issue #7104, opened 2017-05-15
PQ: first write blocks if queue.page_capacity == queue.max_bytes


colinsurprenant:
This need to be confirmed but was witnessed in #7031 - if `queue.page_capacity` == `queue.max_bytes` in the settings, it seems that the first queue write operation blocks.

original-brownbear:
fixed all the way down to 5.6 via https://github.com/elastic/logstash/pull/8275 :)

PR #8186, opened 2017-09-08
BUG: Fix deadlock on Queue not full wait

original-brownbear:
We cannot lock the Ruby side here, otherwise, we can get into the situation of a **full read** but **full queue** that doesn't get acks through.

Why?

* Readers read many events per lock acquire
* Writers just write on

=> Readers can read everything => get stuck on `notEmpty.await();` in `Queue` and hold on to this mutex:

```rb
      def read_batch
        if @queue.closed?
          raise QueueClosedError.new("Attempt to take a batch from a closed AckedQueue")
        end

        batch = new_batch
        @mutex.lock
        begin
          batch.read_next
        ensure
          @mutex.unlock
```

now acks can't happen because they need the same mutex:

```rb
      def close_batch(batch)
        @mutex.lock
        begin
          batch.close
          # there seems to be concurrency issues with metrics, keep it in the mutex
          @inflight_batches.delete(Thread.current)
          stop_clock(batch)
        ensure
          @mutex.unlock
        end
      end
```

-> deadlocked like in the linked issue.

-> We can simply remove the read mutex in Ruby here because we have a lock on the Java side anyways and fix this issue.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -382,6 +382,7 @@ public class Queue implements Closeable {
                     // not trigger a checkpoint creation in itself
                     TailPage tailPage = new TailPage(this.headPage);
                     tailPage.purge();
+                    currentByteSize -= tailPage.getPageIO().getCapacity();
                 } else {
                     // beheading includes checkpoint+fsync if required
                     TailPage tailPage = this.headPage.behead();
--- a/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
+++ b/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
@@ -23,6 +23,7 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
+import org.logstash.ackedqueue.io.AbstractByteBufferPageIO;
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
@@ -76,6 +77,31 @@ public class QueueTest {
 
             assertThat(b.getElements().size(), is(1));
             assertThat(b.getElements().get(0).toString(), is(element.toString()));
+            assertThat(q.nonBlockReadBatch(1), nullValue());
+        }
+    }
+
+    /**
+     * This test guards against issue https://github.com/elastic/logstash/pull/8186 by ensuring
+     * that repeated writes to an already fully acknowledged headpage do not corrupt the queue's
+     * internal bytesize counter.
+     * @throws IOException On Failure
+     */
+    @Test(timeout = 5000)
+    public void writeToFullyAckedHeadpage() throws IOException {
+        final Queueable element = new StringElement("foobarbaz");
+        final int page = element.serialize().length * 2 + AbstractByteBufferPageIO.MIN_CAPACITY;
+        // Queue that can only hold one element per page.
+        try (Queue q = new TestQueue(
+            TestSettings.volatileQueueSettings(page, page * 2 - 1))) {
+            q.open();
+            for (int i = 0; i < 5; ++i) {
+                q.write(element);
+                try (Batch b = q.readBatch(1, 500L)) {
+                    assertThat(b.getElements().size(), is(1));
+                    assertThat(b.getElements().get(0).toString(), is(element.toString()));
+                }
+            }
             assertThat(q.nonBlockReadBatch(1), nullValue());
         }
     }


jordansissel:
I put this on my todo list for Monday.

original-brownbear:
@jordansissel ping here :)

jordansissel:
@original-brownbear Apologies for the delay, I had to read several files of source code to understand the scenario.

The way I am reading this code, `read_batch` doesn't block forever, so how does this deadlock?

* read_batch: grabs the lock, then calls batch.read_next
* ReadBatch#read_next: calls queue.read_batch(size, wait) -- the default wait is 250ms
* AckedQueue.ruby_read_batch: calls Queue.readBatch(size, wait)
* Queue.readBatch calls notEmpty.await(timeout), then Queue._readPageBatch
* Queue._readPageBatch calls Page.readBatch
* Page.readBatch calls PageIO.read calls MMapPageIO.read which is AbstractByteBufferPageIO.read
* AbstractByteBufferPageIO.read reads some elements and returns a SequencedList of zero or more items

Assuming I've unwound the call stack correctly (manually, not via tooling), I'm not seeing where this deadlocks since nothing in this call stack should block forever?

The stack traces indicate:

> 	at [org.logstash.ackedqueue.Queue.readBatch(Queue.java:479)](https://github.com/elastic/logstash/blob/6c2069a452c0c176b7313ac6f09d410cfeaebdb6/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java#L479)

This line is an await() call that has a set timeout, so it should never block forever? It will give up after `batch_delay` (logstash pipeline config) and move on.

jordansissel:
In the parent support ticket (private repo, sorry folks), I see the following in the jstack:

@original-brownbear following up here on the customer data since this isn't a public ticket --

In the jstack, there are 4 workers shown. 3 workers are waiting on lock 0x0000000095119158, the other is in a timed wait on lock 0x0000000094f7ddf0.

* worker0: attempting to readBatch, in a [timed wait (5ms) because the queue is empty](https://github.com/elastic/logstash/blob/6c2069a452c0c176b7313ac6f09d410cfeaebdb6/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java#L479
* worker1: wants to read events, but waiting for the queue lock held by worker0 (I assume)
* worker2: wants to read events, but waiting for the queue lock held by worker0 (I assume)
* worker3: wants to ack processing, but waiting for the queue lock held by worker0

This doesn't look like a deadlock, since the timed wait in worker0 will expire and release its hold on the lock.

However, I do think our locking is a bit greedy. In this model, an empty queue will block acks for up to `batch_delay` (5ms by default) while it waits for the "is empty" condition to clear.

I'm not convinced this is a deadlock scenario?

original-brownbear:
@jordansissel huh you're right, sorry I was in the blocking batch read without a timeout in my mind. That one has a dead lock as explained imo the timed one doesn't. But carefully reading this again I found the deadlock in the timed one too :) Let me try and wake up one last time today just a little more and I'll fix + explain :)

That said, removing this lock on the Ruby end is either way and just and outright performance gain imo.

jordansissel:
@original-brownbear no rush :)

jordansissel:
@original-brownbear yeah, I am +1 on removing the lock on the Ruby side. From my reading of the code, I believe this lock is intended to protect access to a few different things:

* metrics (`inflight_batches`, `start_metrics`, 
* the queue (`empty?`, `read_batch`, `close_batch`)

Thinking about this, I think the locks in the Java code should be enough to protect the queue, so maybe we can remove the queue protections in the Ruby code. The metrics protections I wonder if we can remove the locking there as well.

original-brownbear:
@jordansissel we can remove the locks fully if we store the metrics stuff in a concurrent hashmap like I did here #8081 and also turn this
```rb

      def inflight_batches
        @mutex.lock
        begin
          yield(@inflight_batches)
        ensure
          @mutex.unlock
        end
      end
```
into something that yields a copy of the map (since we're iterating over it in the agent thread and it can be changed in the worker meanwhile), then we should be good I think.

I think the deadlock will have to wait for tomorrow, it's basically the same as https://github.com/elastic/logstash/pull/7380 but from the read side.

We can simply drop this condition (I think in read and ack, read I'd have to double check though, ack I'm sure).

```java
if (wasFull) { notFull.signalAll(); }
```

and signal clean to one writer by turning this into:

```java
notFull.signal();
```

I'll write up the steps of how the condition locks the Queue after some sleep :D

original-brownbear:
@jordansissel ok one more issue I found is fixed in https://github.com/elastic/logstash/pull/8186/commits/77c337c2fe3d5d595043fea648dad168b98bc422

We add new head pages to the `currentByteSize`, but if we have fast readers and purs fully acked head pages before they become tail pages `currentByteSize` keeps increasing until `isFull()` always returns `true`. That would def explain a stack-trace where readers and writers are blocked like we see it (unreadcount == 0 and currentByteSize == max queue size (without any actual pages to ack remaining) will give you exactly this).

Still not 100% happy about the locking but can't really track down a way to break it.
=> I think we're good here with these two fixes. The `currentByteSize` issue is pretty clear cut I think.

original-brownbear:
@jordansissel the reason we're not seeing this very often is that in order to "leak" on this counter, we need the writer to hit a full acknowledged queue. This is super rare just from the math of it, but with a slow producer and fast consumers, it happens.
My best reproducer is this one:

```sh
bin/logstash -w 1 -e 'input { generator {threads => 1 message => "asdfasdfnasdfnasoidnfoaisdnfoasidnfoiasdnfoiasdnfoisadnfoiasndfoinsadofinasdoifnasdoifnsaodifnasodifnasoidfnoaisdfnoiasdnfoiasdnfioasdnfoiasdnfoiasdnfoaisdnfioasdnfasiof"} } output {null {}}'
```

combined with a 10 event ack interval 2 mb queue and 1 mb page size. Still, takes a while to lock up at times.

original-brownbear:
if you only have 2 pages, this just needs to happen once btw, https://github.com/elastic/logstash/issues/7104 is not fixed by this (once your byte count is off by one full page it locks). Might be a hint on how to fix #7104 in here too I guess ...

original-brownbear:
@jordansissel I did it, unit test that reproduces the dead-lock in a trivial way added :)

jordansissel:
@original-brownbear Do you think you can separate the ruby lock change (around batch.read_next) from the queue byte size deadlock fix?

original-brownbear:
@jordansissel I can if it helps sure :) sec reverting that one

original-brownbear:
@jordansissel done :)

original-brownbear:
test this please

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -382,6 +382,7 @@ public long write(Queueable element) throws IOException {
                     // not trigger a checkpoint creation in itself
                     TailPage tailPage = new TailPage(this.headPage);
                     tailPage.purge();
+                    currentByteSize -= tailPage.pageIO.getCapacity();
```
jordansissel: the other place in this file that does this uses the getter `getPageIO()`

For consistency, can you use the getter?

original-brownbear: @jordansissel sure => done :)

original-brownbear:
@jordansissel @suyograo also merge this to 5.6?

jordansissel:
Let's merge this to 5.x, not 5.6. My thought here is that we don't have many reports of this deadlock (as far as I can tell), so there's no urgency to ship it with a patch fix into 5.6.x, but it should go into 5.7 (5.x branch)

original-brownbear:
@jordansissel at the moment we don't have a `5.x` branch do we, only a `5.6`?

suyograo:
There is no 5.x. This can go to master, 6.x and 6.0

original-brownbear:
@suyograo alright good news then, it's already in those https://github.com/elastic/logstash/pull/8186#issuecomment-329689712 :)

PR #8200, opened 2017-09-11
MINOR: Gitignore compiled c code in ITs

original-brownbear:
These two get compiled if you run the full IT suite on Linux.
Since the IT suite is kind of in-flux right now, I wouldn't put more effort towards a nicer build (i.e. proper target folder for binary output) for that` C` code and just ignore those two files explicitly.


jakelandis:
LGTM

PR #8201, opened 2017-09-11
MINOR: Cleanup JAVAC warnings around our Jruby Extensions

original-brownbear:
Another step for #7701 => down to 67 warnings in production code :P 

* Also cleaned up our extension libraries as much as possible to make things `final` where I could especially when it comes to `RubyTimestamp` and `Timestamp`. 
** Those two classes are used in the hot path together with reflection (`getClass()`) so it can't hurt to make them `final` performance and correctness wise. Adjusted the tests accordingly, mocking isn't necessary anymore now that we hold a global reference to the `Ruby` runtime here anyways.
* And removed some dead constructors and factory method from the `RubyTimestamp` logic


jakelandis:
LGTM

original-brownbear:
@jakelandis thanks!

original-brownbear:
Jenkins test this please

Issue #8132, opened 2017-09-03
`org.logstash.bivalues.JavaProxyBiValue` Should be Removed without Replacement?


original-brownbear:
This is a sub-issue of #7712 

`org.logstash.bivalues.JavaProxyBiValue` doesn't really make any sense given what I just learnt from #8131 and #8128.

We cannot run arbitrary custom (Java) types through Jackson serialization and expect correct deserialization. All types we support, JRuby will not wrap in a `JavaProxy`.
Since we don't expose any functionality to extend our serialization => there is no possible benefit from `JavaProxyBiValue` in terms of serialization => keeping this around creates different behavior between PQ and non-PQ setups => this isn't fixable in general and it seems no one ran into this so far => better remove it now?

PS: The map and list case mentioned as motivation in https://github.com/elastic/logstash/issues/5465 are handled further upstream as special case paths for `ArrayJavaProxy` and `MapJavaProxy` so that motivation for this thing is also gone :)

original-brownbear:
@jordansissel wdyt, can we just kill that thing? :)

original-brownbear:
handled in #8239

Issue #7615, opened 2017-07-07
Bump Jackson to newer version


jakelandis:
We are currently on Jackson 2.7.3.

It looks like 2.8 is out, and 2.9 is close to release:
https://github.com/FasterXML/jackson/wiki/Jackson-Releases

This change is dependent upon:
https://github.com/guyboertje/jrjackson/issues/61

We need to ensure that versions stay in sync else #6781 will re-appear.  See PR #7614 for steps on how to keep in sync.

prathibhay:
what is the palns for upgrade and in which release are you planning to upgrade? we are also facing same issue

jakelandis:
@prathibhay  - I can't say for certain based on timing and unknown complexity of the upgrade (hopefully it is just dependency bump).  We have re-prioritized are are targeting the next patch releases for 5.5 and 5.6, and the next release candidate for 6.0.

jakelandis:
Changes are made initially on https://github.com/elastic/logstash/pull/8373 and will be back ported.

aburmeis:
guyboertje/jrjackson#61 has been solved and will be included in v0.4.3

jakelandis:
Jackson updated to 2.9.1 as far back as 5.5.x

Assuming there is release for the given pipeline (not a guarantee or commitment), it would be in:
5.5.4, 5.6.3, 5.7, 6.0.0-rc2, 6.1, 7.0.0-alpha1 

closing issue as resolved.

PR #8205, opened 2017-09-11
#8131 Fix BigInteger and BigDecimal serialization by removing BigDecimalBivalue and BigIntegerBivalue

### LINKED ISSUE #7712: BiValue is redundant

### LINKED ISSUE #8131: BigInteger are not Correctly Serialized
Same as #8128 pretty much.

Trying to send a `BigInteger` through a serialize, then deserialize cycle runs into:

```java
        Event e = new Event();
        final BigInteger huge =
            new BigInteger(String.valueOf(Long.MAX_VALUE) + String.valueOf(Long.MAX_VALUE));
        e.setField("bigint", huge);
        assertEquals(huge, er.getField("bigint"));
```

```
java.lang.IllegalArgumentException: Missing Converter handling for full class name=[B, simple name=byte[]

	at org.logstash.bivalues.BiValues.newBiValue(BiValues.java:32)
	at org.logstash.Valuefier.fallbackConvert(Valuefier.java:91)
	at org.logstash.Valuefier.convert(Valuefier.java:67)
	at org.logstash.ConvertedMap.newFromMap(ConvertedMap.java:50)
	at org.logstash.Event.<init>(Event.java:60)
	at org.logstash.Event.fromSerializableMap(Event.java:193)
	at org.logstash.Event.deserialize(Event.java:392)
	at org.logstash.EventTest.toBinaryRoundtrip(EventTest.java:53)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```

The native-Java-serialization approach using `BiValue` or similar simply doesn't work with Jackson like this. For an `Object` of arbitrary bytes, Jackson simply goes ahead and serializes the fields in the `Object`, in this case `byte[]`.

`BigInteger` needs proper custom serialization here too, like implemented for `Timestamp` in #8130.

@jordansissel FYI, this isn't just a `Timestamp` issue, I think with the exception of `RubySymbol`, `Long` and `Integer` (and their counterparts) this breaks all remaining serializations via `BiValue`.

original-brownbear:
fixes #8131 and also effectively fixes #7712 (only the nil case and Java wrapper case ( #8132 ) ) left.


logstash-core/src/main/java/org/logstash/ObjectMappers.java
```diff
@@ -30,14 +33,22 @@
             .addSerializer(RubySymbol.class, new RubySymbolSerializer())
             .addSerializer(RubyFloat.class, new RubyFloatSerializer())
             .addSerializer(RubyBoolean.class, new RubyBooleanSerializer())
-            .addSerializer(RubyFixnum.class, new RubyFixnumSerializer());
+            .addSerializer(RubyFixnum.class, new RubyFixnumSerializer())
+            .addSerializer(RubyBigDecimal.class, new RubyBigDecimalSerializer())
+            .addSerializer(RubyBignum.class, new RubyBignumSerializer());
+
+    private static final SimpleModule CBOR_DESERIALIZERS =
```
jakelandis: NIT: This should be `RUBY_DESERIALIZERS`, not `CBOR_SERIALIZERS`

original-brownbear: @jakelandis I wanted to make sure that it's clear this is only used with CBOR, never with JSON. I guess I could also JavaDoc it, but it seemed more straightforward for people to read this way right ?

jakelandis: Oh, I see now , the mapper it is registered to also carries the CBOR name.  in that case the name makes sense as-is.

original-brownbear:
@jakelandis thanks merging :)

colinsurprenant:
Per #8379 and #8131 the `BigInteger` and `BigDecimal` serialization problems were in fact related to the lack of CBOR **deserialization** support for these types in Jackson prior to version 2.8 and we  are using 2.7.  We are in the process of upgrading Jackson #7615 and once completed we can probably get rid of the custom serializers introduced here.

jakelandis:
@colinsurprenant - I logged https://github.com/elastic/logstash/issues/8387 to address this.

colinsurprenant:
and #8405 fixes the custom serializers

PR #8206, opened 2017-09-11
[Docs] Clarify that host setting requires port number

dedemorton:
Omitting the port  in `var.elasticsearch.hosts` and `var.kibana.host` settings produces unexpected results, so we are documenting that the port is required.

Also did some general cleanup on the topic

- Updated arcsight topic to use the shared file for common options (was on my to-do list).
- Removed smart quotes because they are detestable.

@acchen97 The change to the host descriptions impacts your ArcSight doc. Changes like this are exactly why I prefer to have a single source of truth for our content. :-/


jordansissel:
LGTM

PR #8212, opened 2017-09-12
MINOR: Remove redundant throws in PipelineIR path

original-brownbear:
Trivial cleanup, preventing further dead code from creeping in :)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
+++ b/logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
@@ -330,14 +330,7 @@ public class Graph implements SourceComponent, Hashable {
     }
 
     public String toString() {
-        Stream<Edge> edgesToFormat;
-        try {
-            edgesToFormat = sortedEdges();
-        } catch (InvalidIRException e) {
-            // Even if this isn't a valid graph we still need to print it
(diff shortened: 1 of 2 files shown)


logstash-core/src/main/java/org/logstash/config/ir/graph/Graph.java
```diff
@@ -330,14 +330,7 @@ public void validate() throws InvalidIRException {
     }
 
     public String toString() {
-        Stream<Edge> edgesToFormat;
-        try {
-            edgesToFormat = sortedEdges();
-        } catch (InvalidIRException e) {
-            // Even if this isn't a valid graph we still need to print it
-            edgesToFormat = edges.stream();
```
jakelandis: This looks like a functional change (not just clean up).  Is this passive?

original-brownbear: @jakelandis yea this is dead code, the exception is never thrown (the compiler now complains here after removing the redundant exceptions) :)

jakelandis:
@original-brownbear - the code changes look fine, perhaps @andrewvc should do the final approval since even though the code here is dead it was _trying_ to do something and may want a follow up issue.

original-brownbear:
@andrewvc thanks for taking a look :)

Issue #8207, opened 2017-09-11
Refactor: Bundler load order


ph:
I know that some of you have been plagued with `already activated` error for some time (@jordansissel)

```
 +Gem::LoadError: You have already activated stud 0.0.23, but your Gemfile requires stud 0.0.24. Prepending `bundle exec` to your command may solve this.
       +  block in setup at /Users/ph/es/logstash/build/logstash-7.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/runtime.rb:34
       +            each at org/jruby/RubyArray.java:1734
       +            each at /Users/ph/es/logstash/build/logstash-7.0.0-alpha1-SNAPSHOT/vendor/jruby/lib/ruby/stdlib/forwardable.rb:189
       +           setup at /Users/ph/es/logstash/build/logstash-7.0.0-alpha1-SNAPSHOT/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/run
```

I have a pretty good understanding of what is going wrong here; it's related to bundler/RubyGems and your default environment. When no bundler is configured for the current scripts, it will load RubyGems, by default RubyGems will add your local gems path as source and will read the specifications. This can cause already activated errors. This means our Logstash isn't correctly isolated from our _local_ env; this often happens in rake tasks because we are not consistently calling bundler on them.

When working on https://github.com/elastic/logstash/issues/8170, I revisited our class and how we load stuff. I believe it's and would not require a lot of big changes.

**What needs to be done:**

- [ ] Remove the custom name for `Gemfile.jruby-2.3.lock`
- [ ] Centralize our require of files into a single point, an init or bootstrap file.
- [ ] Revisit to make sure we always activate the bundler environment even if we install 

**Harder or higher risk:**
- [ ] Refactor our install all plugins rake task
- [ ] refactor the install and update to not use bundler's cli class, but either use the resolver or the same logic that we use for x-pack.

ph:
I am adding more context, I ran into the already activated errors.

```
You have already activated stud 0.0.23, but your Gemfile requires stud 0.0.24
```

I've looked at the gemfile and the lock and we only mention the _0.0.24_ gems nothing about _0.0.23_. But in the **vendor** directory we have both versions installed.

```
[...]

PR #8216, opened 2017-09-12
Use standard name for Gemfile's lockfile

ph:
In previous iteration of logstash < 1.5 we were trying as much as
possible to make the code compatible with different ruby
implementations. This features required use to create a lock specific
for each Ruby version, with the migration to Java code we are only
targetting JRuby implementation.

Reference: #8207

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,7 @@ qa/.vm_ssh_config
 qa/.vagrant
 qa/acceptance/.vagrant
 qa/Gemfile.lock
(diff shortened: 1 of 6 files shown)


original-brownbear:
@ph LGTM :)

ph:
Minimizing any risk by merging in master/6.1 branch.

original-brownbear:
rebased as #8525 and merged there, thanks @ph !

PR #8222, opened 2017-09-12
[Docs] Fix heading level in docker docs

dedemorton:
Incorrect heading level resulted in the doc build creating a new topic called "Pulling the image". I think all this content is meant to be kept under "Running Logstash on Docker" so I'm changing the heading level to ====. 

@jarpy  Can you check the rest of the headings to make sure they are at the level you expect?


jarpy:
Thanks. Sorry for the mess.

jarpy:
I think some of the other sections are bit weird too. I'll get my local build going again and fix them up.

PR #8223, opened 2017-09-12
Introduce docker for CI builds.

jakelandis:
This commit includes:
* A base Dockerfile and script to push to a Docker repo
* A per-build Dockerfile (derived from the base)
* Updates to the test scripts to allow for more parallel builds
* Docker wrappers for the tests scripts
* Update for the integration test readme to manually run the tests
* Clean up the output of the Java tests

This commit does NOT include:
* Changes needed for the CI system to use Docker
* Removal of artifacts used by Travis

Note - this is the most of the changes on feature/docker-builds

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,38 @@
+FROM container-registry-test.elastic.co/logstash-test/logstash-base:latest
+
+RUN ln -s /tmp/vendor /opt/logstash/vendor
+
+ADD gradlew /opt/logstash/gradlew
+ADD gradle/wrapper /opt/logstash/gradle/wrapper
+RUN /opt/logstash/gradlew wrapper
+
+ADD versions.yml /opt/logstash/versions.yml
+ADD LICENSE /opt/logstash/LICENSE
+ADD CONTRIBUTORS /opt/logstash/CONTRIBUTORS
+ADD Gemfile.template /opt/logstash/Gemfile
+ADD Rakefile /opt/logstash/Rakefile
+ADD build.gradle /opt/logstash/build.gradle
+ADD rakelib /opt/logstash/rakelib
+ADD config /opt/logstash/config
+ADD spec /opt/logstash/spec
+ADD qa /opt/logstash/qa
+ADD lib /opt/logstash/lib
+ADD pkg /opt/logstash/pkg
+ADD tools /opt/logstash/tools
+ADD logstash-core /opt/logstash/logstash-core
+ADD logstash-core-plugin-api /opt/logstash/logstash-core-plugin-api
+ADD bin /opt/logstash/bin
+ADD modules /opt/logstash/modules
+ADD ci /opt/logstash/ci
+ADD CHANGELOG.md /opt/logstash/CHANGELOG.md
+ADD settings.gradle /opt/logstash/settings.gradle
+
+USER root
+RUN rm -rf build && \
+    mkdir -p build && \
+    chown -R logstash:logstash /opt/logstash
+USER logstash
+WORKDIR /opt/logstash
+
+LABEL retention="prune"
+
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -1,0 +1,47 @@
+#logstash-base image, use ci/docker_update_base_image.sh to push updates
+FROM ubuntu:xenial
+
+RUN apt-get update && \
+    apt-get install -y zlib1g-dev build-essential vim rake git curl libssl-dev libreadline-dev libyaml-dev  \
+      libxml2-dev libxslt-dev openjdk-8-jdk-headless curl iputils-ping netcat && \
+    apt-get clean
+
+WORKDIR /root
+
+RUN adduser --disabled-password --gecos "" --home /home/logstash logstash && \
+    mkdir -p /usr/local/share/ruby-build && \
+    mkdir -p /opt/logstash && \
+    mkdir -p /mnt/host && \
+    chown logstash:logstash /opt/logstash
+
+USER logstash
+WORKDIR /home/logstash
+
+RUN git clone https://github.com/sstephenson/rbenv.git .rbenv && \
+    git clone https://github.com/sstephenson/ruby-build.git .rbenv/plugins/ruby-build && \
+    echo 'export PATH=/home/logstash/.rbenv/bin:$PATH' >> /home/logstash/.bashrc
+
+ENV PATH "/home/logstash/.rbenv/bin:$PATH"
+
+#Only used to help bootstrap the build (not to run Logstash itself)
+RUN echo 'eval "$(rbenv init -)"' >> .bashrc && \
+    rbenv install jruby-9.1.12.0 && \
+    rbenv install jruby-1.7.27 && \
+    rbenv global jruby-9.1.12.0 && \
+    bash -i -c 'gem install bundler' && \
+    rbenv local jruby-9.1.12.0 && \
+    mkdir -p /opt/logstash/data
+
+
+# Create a cache for the dependencies based on the current master, any dependencies not cached will be downloaded at runtime
+RUN git clone https://github.com/elastic/logstash.git /tmp/logstash && \
+    cd /tmp/logstash && \
+    rake test:install-core && \
+    ./gradlew compileJava compileTestJava && \
+    cd qa/integration && \
+    /home/logstash/.rbenv/shims/bundle install && \
+    mv /tmp/logstash/vendor /tmp/vendor && \
+    rm -rf /tmp/logstash
+
+# used by the purge policy
+LABEL retention="keep"
--- a/build.gradle
+++ b/build.gradle
@@ -8,10 +8,46 @@ allprojects {
   project.targetCompatibility = JavaVersion.VERSION_1_8
 
   tasks.withType(JavaCompile).all {
-    options.compilerArgs.add("-Xlint:all")
+
+      def env = System.getenv()
+      boolean ci = env['CI']
+
+      //don't lint when running CI builds
+      if(!ci){
+          options.compilerArgs.add("-Xlint:all")
+      }
   }
 
   clean {
     delete "${projectDir}/out/"
+  }
+
+  //https://stackoverflow.com/questions/3963708/gradle-how-to-display-test-results-in-the-console-in-real-time
+  tasks.withType(Test) {
+    testLogging {
+      // set options for log level LIFECYCLE
+      events "passed", "skipped", "failed", "standardOut"
+      showExceptions true
+      exceptionFormat "full"
+      showCauses true
+      showStackTraces true
+
+      // set options for log level DEBUG and INFO
+      debug {
+        events "started", "passed", "skipped", "failed", "standardOut", "standardError"
+        exceptionFormat "full"
+      }
+      info.events = debug.events
+      info.exceptionFormat = debug.exceptionFormat
+
+      afterSuite { desc, result ->
+        if (!desc.parent) { // will match the outermost suite
+          def output = "Results: ${result.resultType} (${result.testCount} tests, ${result.successfulTestCount} successes, ${result.failedTestCount} failures, ${result.skippedTestCount} skipped)"
+          def startItem = '|  ', endItem = '  |'
+          def repeatLength = startItem.length() + output.length() + endItem.length()
+          println('\n' + ('-' * repeatLength) + '\n' + startItem + output + endItem + '\n' + ('-' * repeatLength))
+        }
+      }
+    }
   }
 }
--- a/ci/acceptance_tests.sh
+++ b/ci/acceptance_tests.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+set -e
+
+# Since we are using the system jruby, we need to make sure our jvm process
+# uses at least 1g of memory, If we don't do this we can get OOM issues when
+# installing gems. See https://github.com/elastic/logstash/issues/5179
+export JRUBY_OPTS="-J-Xmx1g"
+
+SELECTED_TEST_SUITE=$1
+
+# The acceptance test in our CI infrastructure doesn't clear the workspace between run
+# this mean the lock of the Gemfile can be sticky from a previous run, before generating any package
+# we will clear them out to make sure we use the latest version of theses files
+# If we don't do this we will run into gem Conflict error.
+[ -f Gemfile ] && rm Gemfile
+[ -f Gemfile.jruby-2.3.lock ] && rm Gemfile.jruby-2.3.lock
+
+if [[ $SELECTED_TEST_SUITE == $"redhat" ]]; then
+  echo "Generating the RPM, make sure you start with a clean environment before generating other packages."
+  rake artifact:rpm
+  echo "Acceptance: Installing dependencies"
+  cd qa
+  bundle install
+
+  echo "Acceptance: Running the tests"
+  bundle exec rake qa:vm:setup["redhat"]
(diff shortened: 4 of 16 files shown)


ci/unit_tests.bat
```diff
@@ -0,0 +1,41 @@
+@echo off
```
jakelandis: No changes here ... just a copy/paste from the old name (but kept the old name in there since it is currently in use)

build.gradle
```diff
@@ -8,10 +8,46 @@ allprojects {
...
+      if(!ci){
+          options.compilerArgs.add("-Xlint:all")
+      }
   }
 
   clean {
     delete "${projectDir}/out/"
   }
+
+  //https://stackoverflow.com/questions/3963708/gradle-how-to-display-test-results-in-the-console-in-real-time
```
jakelandis: Example output:
https://travis-ci.org/elastic/logstash/jobs/274789145

original-brownbear:
@jakelandis ready for review here right ? :)

jakelandis:
@original-brownbear - yes, please.

ci/docker_integration_tests.sh
```diff
@@ -0,0 +1,21 @@
...
+    IMAGE_NAME=$branch_specifier"-"$(date +%s%N)
+fi
+
+echo "Running Docker CI build for '$IMAGE_NAME' "
+
+docker build  -t $IMAGE_NAME .
+exit_code=$?; [[ $exit_code != 0 ]] && exit $exit_code
+docker run -t --rm $IMAGE_NAME ci/integration_tests.sh $@
+exit_code=$?
+[[ $IMAGE_NAME != "logstash-integration-tests" ]] && docker rmi $IMAGE_NAME
```
original-brownbear: @jakelandis Nit: you might want to `trap` the image delete if you want to be nice to whoever maintains Jenkins :)

Dockerfile.base
```diff
@@ -0,0 +1,48 @@
+#logstash-base image, use ci/docker_update_base_image.sh to push updates
+FROM ubuntu:xenial
+
+RUN apt-get update && \
+    apt-get dist-upgrade && \
```
original-brownbear: @jakelandis don't do `upgrade` in a `Dockerfile`, this just makes things unstable in the long run depending on what happens to the base image. see https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#apt-get for some details on this

Dockerfile.base
```diff
@@ -0,0 +1,48 @@
...
+    rbenv install jruby-9.1.12.0 && \
+    rbenv install jruby-1.7.27 && \
+    rbenv global jruby-9.1.12.0 && \
+    bash -i -c 'gem install bundler' && \
+    rbenv local jruby-9.1.12.0 && \
+    mkdir -p /opt/logstash/data
+
+
+# Create a cache for the dependencies based on the current master, any dependencies not cached will be downloaded at runtime
+RUN git clone https://github.com/elastic/logstash.git /tmp/logstash && \
```
original-brownbear: @jakelandis hmmmm hmmm hmmm, I like how this makes the build a ton faster :)
I don't like how this makes the build run from a dirty dependency state :(

I mean ... this saves like ~3 min per build tops I guess, is it really worth it potentially missing some crazy gem conflict for that (think about what happens if a plugin gets updated, it will only show up here on rebuild which doesn't seem like a good idea)?

ci/docker_integration_tests.sh
original-brownbear: then again ... nah let's not do that, it gets tricky :) Kind of unlikely to actually have build the image and still getting exit `!=0`from the build not sure if it's even possible.

ci/unit_tests.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash -ie
+
+# Since we are using the system jruby, we need to make sure our jvm process
+# uses at least 1g of memory, If we don't do this we can get OOM issues when
+# installing gems. See https://github.com/elastic/logstash/issues/5179
+export JRUBY_OPTS="-J-Xmx1g"
```
original-brownbear: @jakelandis for "at least" you'd want `Xms` here :)

original-brownbear:
@jakelandis sweet :) One trivial thing (the `Xms` thing) and one important one (the dependency issue).

Dockerfile.base
jakelandis: > think about what happens if a plugin gets updated, it will only show up here on rebuild which doesn't seem like a good idea

When a plugin gets updated, it will have a new version, and will get pulled from Rubygems.  Only published gems are tested here.

ci/unit_tests.sh
jakelandis: In this case, I think we should let the memory grow naturally. This memory setting is only used by bundler and don't want to consume memory unless we need to.  I don't know how much memory is actually needed. (this has been like this for a while)

original-brownbear: @jakelandis I have to problem with the implementation, just the wording of the comment was weird :D Only setting `Xmx` here is fine with me :)

Dockerfile.base
original-brownbear: Ah right ... because the `Gemfile.lock` is generated fresh on every run ... nevermind :)

Dockerfile.base
jakelandis: good call... i removed the upgrade.

original-brownbear:
@jakelandis LGTM now :) maybe have @andrewvc double check since there are quite a few lines here :)

jakelandis:
@original-brownbear - all good ?

@andrewvc - can I also get your review?

rakelib/test.rake
```diff
@@ -6,9 +6,14 @@ require 'pathname'
...
-  desc "run the java tests"
+  desc "run the java unit tests"
   task "core-java" do
-    exit(1) unless system './gradlew clean test --info'
+    exit(1) unless system './gradlew clean test'
+  end
+
+  desc "run the ruby unit tests"
+  task "core-ruby" do
+    exit 1 unless system(*default_spec_command)
```
andrewvc: This should probably be `exec(*default_spec_command)` which simply replaces the current process with the executed command. You won't need to explicitly `exit` in that case since `exec` replaces the running process.

ci/integration_tests.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash -ie
+
+# Since we are using the system jruby, we need to make sure our jvm process
+# uses at least 1g of memory, If we don't do this we can get OOM issues when
+# installing gems. See https://github.com/elastic/logstash/issues/5179
+export JRUBY_OPTS="-J-Xmx1g"
```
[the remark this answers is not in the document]
jakelandis: @andrewvc - this is only used by JRuby bundler during the bootstrap process.

ci/integration_tests.sh
```diff
@@ -0,0 +1,56 @@
...
+
+export SPEC_OPTS="--order rand --format documentation"
+export CI=true
+
+if [[ ! -d "build" ]]; then
+    mkdir build
+fi
+rm -rf build/*
```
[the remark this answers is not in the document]
jakelandis: will update.... (fwiw was just copy/pasting from old script)

qa/integration/specs/cli/prepare_offline_pack_spec.rb
```diff
@@ -3,15 +3,8 @@
...
-require "logstash/devutils/rspec/spec_helper"
 
-
-# These are segmented into a separate tag that MUST be run separately from any docker tests
-# The reason they break the Docker API and that in turn even breaks tests not using Docker 
-# is that the Docker API has a global singleton Docker container set up as soon as it's 
-# required that acts in the background and will err out if the internet is down
-# See https://github.com/elastic/logstash/issues/7160#issue-229902725
-describe "CLI > logstash-plugin prepare-offline-pack", :offline => true do
+describe "CLI > logstash-plugin prepare-offline-pack" do
```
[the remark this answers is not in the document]
jakelandis: They were only separated because they broke the Docker dependencies, which no longer exist (as the delete comment suggests).

ci/docker_unit_tests.sh
```diff
@@ -0,0 +1,21 @@
...
+
+docker build  -t $IMAGE_NAME .
+exit_code=$?; [[ $exit_code != 0 ]] && exit $exit_code
+docker run -t --rm $IMAGE_NAME ci/unit_tests.sh $@
+exit_code=$?
+[[ $IMAGE_NAME != "logstash-unit-tests" ]] && docker rmi $IMAGE_NAME
+echo "exiting with code: '$exit_code'"
+exit $exit_code #preserve the exit code from the test run
+
+#Note - ensure that the -e flag is NOT set, and explicitly check the $? status to allow for clean up
```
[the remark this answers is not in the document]
jakelandis: will do

ci/docker_integration_tests.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "exiting with code: '$exit_code'"
+exit $exit_code #preserve the exit code from the test run
+
+#Note - ensure that the -e flag is NOT set, and explicitly check the $? status to allow for clean up
```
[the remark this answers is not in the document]
jakelandis: will do

ci/integration_tests.sh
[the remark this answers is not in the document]
andrewvc: Ohhhhhh. NM then!

qa/integration/specs/cli/prepare_offline_pack_spec.rb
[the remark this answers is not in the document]
andrewvc: I don't think that's correct.

I'm the one who separated the tests IIRC, and the issue, if I recall, was that the special C wrapper used to go 'offline' would break the non-offline tests.

rakelib/test.rake
jakelandis: hmm... i am going to leave this one as-is.  This pattern is copied from many other locations in our code (and same file), and apprehensive of changing it right now.

Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM container-registry-test.elastic.co/logstash-test/logstash-base:latest
+
+RUN mv /tmp/vendor /opt/logstash/vendor
```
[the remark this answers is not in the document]
jakelandis: will do ... if can remember the order of target vs. source

ci/acceptance_tests.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
jakelandis: I snuck this in ... just a copy/paste of the original file to get the naming consistent. Will delete original once Jenkins is updated.

ci/docker_unit_tests.sh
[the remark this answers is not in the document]
jakelandis: Github is being weird... this has been updated, just isn't showing up here.

ci/docker_integration_tests.sh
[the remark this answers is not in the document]
jakelandis: Github is being weird... this has been updated, just isn't showing up here.

jakelandis:
@andrewvc - I believe I have all the issues addressed.

rakelib/test.rake
andrewvc: Works for me. The efficiency of `exec` isn't realized on jruby anyway, since there's no native way of doing a real process image replacement in java.

ci/docker_integration_tests.sh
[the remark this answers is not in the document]
andrewvc: Are you sure you pushed up all staged changes?

ci/unit_tests.bat
andrewvc: If it's a copy/paste I think maybe one of them should just execute the other

jakelandis: Once I get the changes pushed to Jenkins, I will delete all the old scripts.  The duplicates should only be in source control for a day or two.

ci/docker_integration_tests.sh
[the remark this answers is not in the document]
jakelandis: doh!  there were 4 files that needed this change, i only updated the 2 without comments. 

fixed now.

qa/integration/specs/cli/prepare_offline_pack_spec.rb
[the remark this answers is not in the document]
jakelandis: spoke on slack, and believe we came to a consensus that this change is OK since it only caused issues with the Docker API which we don't use anymore.  @andrewvc - can you confirm ?

qa/integration/README.md
```diff
@@ -2,26 +2,57 @@
...
+rspec specs/*  (all tests)
+```
+## Running integration tests locally via Docker 
 
-You are now ready to run any tests from `qa/integration`.
-* Run all tests: `rspec specs/*`
-* Run single test: `rspec specs/es_output_how_spec.rb`
+From the Logstash root directory:
+
+* Run all tests: `ci/docker_integration_tests.sh`
```
andrewvc: This runs the tests locally without docker. I think you meant to post a variant on the 'single test' command below.

qa/integration/README.md
```diff
@@ -2,26 +2,57 @@
...
+``` 
+* Debug tests:
+```
+docker ps -a | grep Exited | awk '{print $1}' | xargs docker rm
+docker build -t logstash-integration-tests . 
+docker run -d --name debug logstash-integration-tests tail -f /dev/null
+docker exec -it debug ci/integration_tests.sh setup 
+docker exec -it debug bash
+cd qa/integration
+rspec specs/es_output_how_spec.rb
```
andrewvc: Should have a `bundle exec` in front of this

qa/integration/README.md
```diff
@@ -2,26 +2,57 @@
...
+
+* Run all tests: `ci/docker_integration_tests.sh`
+* Run a single test: 
+```
+docker build  -t logstash-integration-tests .
+docker run -it --rm logstash-integration-tests ci/integration_tests.sh specs/es_output_how_spec.rb
+``` 
+* Debug tests:
+```
+docker ps -a | grep Exited | awk '{print $1}' | xargs docker rm
```
andrewvc: `docker ps --all -q -f status=exited | xargs docker rm` is probably a bit more reliable in terms of this functioning. (`grep Exited` is a bit imprecise)

qa/integration/README.md
jakelandis: This script does run it Docker...but I see the confusion and will update to make it more clear.

qa/integration/README.md
jakelandis: nice catch , thanks

qa/integration/README.md
jakelandis: nice!

jakelandis:
@andrewvc - I think the remainder has been addressed.  I re-worked the readme slightly.  The rendered version can be seen here: https://github.com/elastic/logstash/blob/feature/docker-builds/qa/integration/README.md

jakelandis:
Will need manual merges back to 6.0, 5.6, and 5.5

jakelandis:
https://github.com/elastic/logstash/pull/8263

Branch | Commits
-------|--------
5.6 | 2a72ce0ab1b38ce641e38e763218ba00e0e2483e
6.0 | 7c6450cd592d68a048a0fc3728ef11780eff28dc

jakelandis:
https://github.com/elastic/logstash/pull/8265

Branch | Commits
-------|--------
5.5 | 50b5989c7dba6c914e69fcfbb352f65d97d761fd

Issue #7986, opened 2017-08-11
[DISCUSS] Enable pure-Java plugins


jordansissel:
Goal: Design the API for pure Java plugins

Code Status - all code is work-in-progress and subject to commentary :)

* Plugin Interfaces:  #8227 
* Runtime plugin configuration: #8240
* Supporting ruby plugins: #8230 (depends on #8227)

Design/Discussion issues:

* Plugin loading https://github.com/elastic/logstash/issues/8375

----

### WORK IN PROGRESS

----

Background: Logstash plugins still require at least some Ruby component(s) in order to load. We have several plugins now (beats, dissect, date, geoip, etc) that are written  *mostly* in Java but still have some ruby parts.

Greenfield Opportunities:
* ✅ Remove filter plugin `flush` concept
* ❌ Merge filter and output plugin concepts into a unifying `processor` type
* ✅ Provide major categories of settings (SSLContext, HTTPClient) that can be shared
* ❔ Replace filter plugin `filter_matched` with something that is less a burden on the plugin author.
* ❔ Simplify `decorate` concept for inputs/codecs
* ✅ Move to rely on Java features where possible.

Key Objectives:
* Plugin contains no Ruby code
* Plugin tests contain no Ruby code
* Stretch goal: Plugin build tools contain no Ruby code

Areas for design:
* Plugin discovery and loading.
* Plugin API.

## Overview of Plugin API proposed:

This section is relevant to plugin authors.

### Plugin Configuration

Today, we have the `config` class method on ruby Plugin classes:

```ruby
class MyPlugin < Input
  config :foo, :validate => :string, ...
end
```

Ideally I would like to keep a similar model. The goal of the `config` method above is to make it simple for the author to define settings and have Logstash do the hard part of parsing, validating, and providing feedback to the user.

Accepted implementation (so far):

Have plugin instances accept a configuration object to configure themselves with. 

This is what [Tealess](https://github.com/elastic/tealess/blob/master/src/main/java/co/elastic/tealess/cli/ConnectCommand.java#L93-L102) and [Elasticsearch](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/common/xcontent/ConstructingObjectParser.java) do.

Rejected implementation options:
[...]

PR #8227, opened 2017-09-12
Java API: Add interfaces for Inputs, Processors (filters), and Outputs.

jordansissel:
This PR is part of several PRs that will come for #7986.

Adds interfaces for pure-java Inputs, Processors (aka filters), and Outputs. This API closely mirrors the way Ruby plugins work today. My hope is that we can make more improvements to the plugin API before it reaches a release, but I want to keep these PRs small for easy review.

For #7986 

Differences from Ruby API:

* Inputs/Filters/Outputs have no more `register` method. All setup should be done in constructors or in `static { }`
* Processors (filters) have no `flush` concept
* Multi-event processing is default. Single-event processing (`do_filter` or `receive`) methods are removed.
* Output 'concurrency' concept is removed (`concurrency` and `declare_threadsafe` methods are removed)
* Inputs and Outputs don't always have a codec. 

Todo:

* [ ] Codecs
* [ ] Add way for plugin instances (inputs, filters, outputs, etc) to see the pipeline settings (logstash.yml, etc) (and in the future, the credential store).


jakelandis:
I am concerned about using Thread interrupts as the primary life-cycle contract for inputs . 

* This couples the concept of an input and a thread.  Today there may be a 1:1 relationship between an input and a thread, but that may not hold true over time.
* Thread interrupts are non-intuitive and in my experience even seasoned developers get it wrong. For example your java doc reads "When `Thread.interrupted()` is true..." ...however, I suspect you meant Thread.isInterrupted() which only checks the interrupt state and does not clear it too.
* Some inputs may not actually need to handle the interrupt if they are consuming a 3rd party that does it for them.  For example Jetty HTTP can be started on a thread and the thread shutdown without the developer needing to work with interrupts.  
* There is a life-cycle that needs to be managed here, so why not be explicit ?

logstash-core/src/main/java/org/logstash/plugin/Input.java
```diff
@@ -0,0 +1,32 @@
...
+     * <p>
+     * Acknowledging Data:
+     * <p>
+     * When `consumer.accept(...)` returns, the given Events have been successfully written into the consumer, and it is now
+     * safe to acknowledge these events to the upstream data source. The consumer is generally the Logstash internal queue,
+     * such as the persistent queue.
+     *
+     * @param consumer Send batches of events with consumer.accept(batch).
+     */
+    void run(Consumer<Collection<Event>> consumer);
```
jakelandis: I think a stronger contract here would be to sub the Consumer interface to satisfy the generic (or just duplicate the Consumer interface with the generic satisfied).  It will still be strong typed, imo slightly simpler to read.  Also, if you duplicate, we can change the semantics if `accept` isn't a good fit.

jordansissel:
> I am concerned about using Thread interrupts as the primary life-cycle contract for inputs .

The reason I chose this (for now) is because of a few factor:

0) Historical. Inputs are each run in separate threads. [If all inputs exit, Logstash considers this to be a pipeline shutdown signal](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/pipeline.rb#L341-L354). The use of Thread interrupt is conceptually similar to how Logstash inputs work today. For terminating condition, `LogStash::Inputs::Base` provides an AtomicBoolean `@stop_called` that is used for the same purpose as my proposal here which replaces the AtomicBoolean `@stop_called` with the Java-builtin Thread interrupt.

* `LogStash::Inputs::Base#stop?` == `Thread.isInterrupted()`
* `LogStash::Inputs::Base#do_stop` == `Thread.interrupt()`

1) Inputs need a terminating condition to allow Logstash to shutdown the pipeline.

We need a stop condition that an input can use to know to stop. We can choose Thread.isInterrupted() or we can write our own, but the effect is the same, imo.

2) Many plugins are of the form `while (!stop) { /* do work */ }`:

```
% grep -l 'while.*!stop\?' *input*/lib/logstash/inputs/*.rb | fex /-1.1 | sort | fmt
drupal_dblog elasticsearch eventlog exec ganglia gelf generator
google_pubsub irc jms kafka log4j lumberjack pipe redis relp rss stdin
syslog tcp udp unix zeromq
```

23 inputs (of 57 total) use 'while !stop? ... do work ... end`

As mentioned, we need a stop condition. Thread interrupt is a convenient built-in for a stop condition.

3) Thread interrupt has some nice properties: Many functions respect Thread interrupts (Many IO calls, Joni's regexp engine, Thread.sleep, etc)

So even if we choose an alternative, we'll almost certainly want our alternative to still *call* `Thread.interrupt()` anyway. InterruptedException is a checked exception, so everyone has to catch this from any call that could throw it, anyway.

Something in Java that is close to how Inputs work is [Futures](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html#cancel(boolean)). Where we have `stop` in Logstash inputs today, Futures have `cancel`, and `cancel(true)` will interrupt the thread executing this future.

Putting these all together, because my PR has Input implement Runnable, we would be able to run an Input in an Executor by doing `executor.execute(new FutureTask(input))` (because FutureTask wraps a Runnable as a Future).


-----

> This couples the concept of an input and a thread. Today there may be a 1:1 relationship between an input and a thread, but that may not hold true over time.

Many plugins are single-threaded, but some are not. Our current model supports both. When a plugin does its own work/thread management, we still need a way to know if a plugin has finished/errored:

* Beats https://github.com/logstash-plugins/logstash-input-beats/blob/master/lib/logstash/inputs/beats.rb#L193-L197
* HTTP https://github.com/logstash-plugins/logstash-input-http/blob/master/lib/logstash/inputs/http.rb#L168
* Kafka https://github.com/logstash-plugins/logstash-input-kafka/blob/master/lib/logstash/inputs/kafka.rb#L223-L227

For the example plugins above that manage their own threads, you can see that the lines of code required to implement `run` are very small: `run` becomes "start the threaded things and wait until they finish"

> There is a life-cycle that needs to be managed here, so why not be explicit ?

Not all inputs are threaded (23 of 57 are either not threaded or the underlying library hides the thread management from us)

logstash-core/src/main/java/org/logstash/plugin/Input.java
jordansissel: Being explict about the type (`Collection<Event>`) feels like the right, rather than having:

```
public interface Input<T> {
  void run(Consumer<T> consumer);

  // or this
  void run(Consumer<Collection<T>> consumer);
}
```

because for Logstash inputs, I can't imagine any other type being passed. What did you have in mind?

jordansissel:
> For example your java doc reads "When Thread.interrupted() is true..." ...however, I suspect you meant Thread.isInterrupted() which only checks the interrupt state and does not clear it too.

The full text I wrote for the contract is "When `Thread.interrupted()` is true, you MUST close all open resources and return."

The key point here is that you *return* from the `run` method so there would be no possibility of having `Thread.interrupted()` called twice.

Similarly, a user could misinterpret any other `stop condition` api, and try to close resources twice (or more) and end up throwing exceptions due to similar mistakes. Either way, programmers can make errors. Logstash has a test suite available to all plugins that will fail if an input plugin does not terminate correctly: https://github.com/elastic/logstash-devutils/blob/master/lib/logstash/devutils/rspec/shared_examples.rb#L3-L21

jordansissel:
We could provide an AtomicBoolean instead of using Thread interrupt directly, but this won't interrupt anything deeper in the stack:

```
void run(Consumer<>, AtomicBoolean stop) {
  while (!stop.get()) {
    /* do work */
  }
  /* do any resource cleanup */
}
```

This, to me, seems not more beneficial than thread interrupts:

```
void run(Consumer<>) {
  while (!Thread.interrupted()) {
    /* do work */
  }
  /* do any resource cleanup */
}
```

jordansissel:
Experiment: Added a default method `toCallable` to get a `Callable<>` which can be run in an ExecutorService. This would let us have the inputs all managed by an ExecutorService within Logstash instead of our custom thread management that we do today. This is optional and not-required for the initial push of this, but it is present to invite discussion.

logstash-core/src/main/java/org/logstash/plugin/Input.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Return this Input as a Callable for use with an ExecutorService.
```
original-brownbear: @jordansissel I think our inputs and workers are one of those rare cases where you rather want an explicit `Thread` per instance.
JRuby (and also in some spots we ourselves) is making heavy use of `Threadlocal` and using more `Threadlocal` logic for reusable buffers etc. would help performance greatly.
=> If we're planning on reusing threads across something like pipeline reloads then `Threadlocal` may be a problematic pattern => we don't want to reuse threads => `ExecutorService` doesn't really make things all that much simpler unless you want scheduling or thread reuse imo :)

jakelandis:
When having API discussions it is imprortant to remember the 3 rules :
* Always quote Joshua Bloch's [How to design a good API and Why it matters](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/32713.pdf)
* Always express [Ivory tower opinions](https://www.merriam-webster.com/dictionary/ivory%20tower) as facts
* The discussion is not over until someone invokes [Godwin's law](https://en.wikipedia.org/wiki/Godwin%27s_law)

Warning I plan to use 2/3 in my argument below. :)

So for clarity, the heart of my argument is changing this:

```
public interface Input {
 void run(Consumer<Collection<Event>> consumer);
}
```
to this

```
public interface Input {
 void run(Consumer<Collection<Event>> consumer);
 void stop();
}
```

_I have some other comments but think this is the most important and would like to settle this dicussion first._

I believe the former leaks implementation details (via the Javadoc), is easy to misuse, and requires documentation to fully express the contract. 

> "Easy to use, even without documentation", Functionality should be easy to explain", and "Don't let implementation details "leak” into API"

An API should be described in terms of WHAT it does.  When looking at what an input does, it simply starts, stops, and pushes events. In the former the stop action is missing, it is documented in the javadoc, but the javadoc is NOT part of the contract. Further the javadoc is leaking an implementation detail of HOW it should be stopped, which is not always applicable. How a plugin is stopped is irrelevant from an API, and the developer may choose any number of ways to get the plugin to stop. 

> "Start before you've implemented the API"

I have the luxury of not having history or real world complications. I would encourage anyone implementing an API to start from an ivory tower, forget history, then figure out how make to make it work.  For example, do we really want plugin developer to have to understand _anything_ about the threading model for thier plugins, or do we just want them think in terms of starting, stop, push.    

> "Know and avoid API traps and pitfalls"

IMO using Thread.interrupt in an API contract is [pitfall](http://www.dictionary.com/browse/pitfall). Interestingly Bloch uses Thread.interrupt() as a counter example to a good API.

logstash-core/src/main/java/org/logstash/plugin/Input.java
original-brownbear: @jordansissel as discussed I think we should def not use `Collection` here.

What I really like would be this:

```java
public interface EventIterator {

    /**
     * @return Event currently pointed at
     */
    Event current();

    /**
     * @return True iff there was another Event in the iterator
     */
    boolean next();

    /**
     * @param buffer Collection to store all events in
     * @param count Maximum number of events to store in buffer
     * @return Number of events copied to buffer
     */
    int collect(Collection<Event> buffer, int count);
}
```

If we pass something like this down the callstack, we can hide the whole Queue logic behind it and each `Filter` and `Output` can decide if it wants to operate in a batched or non-batched way and we can optimize accordingly. Also we can simply signal end of execution by returning `false` from `next()` as well as `0` from `collect` (I think both can be made blocking, the current situation of returning empty batches under pressure is just wrong imo).

logstash-core/src/main/java/org/logstash/plugin/Processor.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Collection;
+
+public interface Processor {
+    /**
+     * Process events. In the past, this was called a `filter` in Logstash.
+     *
+     * @param events The events to be processed
+     * @return Any new events created by this processor.
+     */
+    Collection<Event> process(Collection<Event> events);
```
original-brownbear: @jordansissel as discussed I think we should def not use `Collection` here.

What I really like would be this:

```java
public interface EventIterator {

    /**
     * @return Event currently pointed at
     */
    Event current();

    /**
     * @return True iff there was another Event in the iterator
     */
    boolean next();

    /**
     * @param buffer Collection to store all events in
     * @param count Maximum number of events to store in buffer
     * @return Number of events copied to buffer
     */
    int collect(Collection<Event> buffer, int count);
}
```

If we pass something like this down the callstack, we can hide the whole Queue logic behind it and each `Filter` and `Output` can decide if it wants to operate in a batched or non-batched way and we can optimize accordingly. Also we can simply signal end of execution by returning `false` from `next()` as well as `0` from `collect` (I think both can be made blocking, the current situation of returning empty batches under pressure is just wrong imo).

logstash-core/src/main/java/org/logstash/plugin/Input.java
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param consumer the
+     * @return A Callable which wraps the `run` method.
+     */
+    default Callable<Void> toCallable(Consumer<Collection<Event>> consumer) {
```
original-brownbear: @jordansissel I don't think it makes all that much sense to allow anything but single `Event` here with our current ack strategy.
How will we enforce something like the max unread count if inputs can pass arbitrarily sized `Collection`s to the Queue? (come to think of it, this is a problem in the current implementation as well ... probably better not to perpetuate this issue and only accept single `Event` here).
This also allows for much nicer batching to an IO backed queue, you can simply collect up to `n` `Event`  to an `ArrayBlockingQueue` and batch write that thing if you feel like you have enough elements to make a batch write worthwhile.
On the other hand you don't force the in memory Queue to handle the overhead of receiving a batch (it's really just overhead for an `ArrayBlockingQueue` because it can't enqueue a batch and it's kind of hard to make a fast concurrent Queue that has batched writes without allocations ... just enqueuing `Collection<Event>` to an`ArrayBlockingQueue` seems like a really bad idea here in terms of performance and also memory usage under pressure).

logstash-core/src/main/java/org/logstash/plugin/Input.java
jakelandis: I meant this:
```
public interface InputConsumer extends Consumer<Collection<Event>> {
}
```
so the API looks like
```
public interface Input {
    void run(InputConsumer consumer);
}
```
...or if you want better/different semantics then `apply` and `andThen` just create new functional interface and change the method names  (there is nothing magic about the Consumer interface).

It is a minor point, and just trying to simplify the readability.

logstash-core/src/test/java/org/logstash/plugin/InputTest.java
```diff
@@ -0,0 +1,46 @@
...
+                int eventCountPerBatch = TestUtil.random.nextInt(100) + 1;
+                Collection<Event> events = new LinkedList<>();
+                for (int i = 0; i < eventCountPerBatch; i++) {
+                    Event e = new Event();
+                    e.setField("message", "hello world");
+                    e.setField("i", i);
+                    e.setField("b", b);
+                    events.add(e);
+                }
+                consumer.accept(Collections.unmodifiableCollection(events));
```
original-brownbear: @jordansissel this is actually a good example of why you don't want to expose a `Collection<>` API to plugins here. This makes performance really unpredictable.
If someone passes us a memory-wise cheap `ArrayList` here, performance characteristics (especially optimal on-heap queue size(s)) will be very different from a super heavy `LinkedList`. Only allowing single events to be passed here, allows us to have much more predictable (and if we do a good job, also faster) behavior here imo.

original-brownbear:
@jakelandis @jordansissel few points on the `Thread` / `interrupt`situation:

> This couples the concept of an input and a thread. Today there may be a 1:1 relationship between an input and a thread, but that may not hold true over time.

I think this should probably be the case until there is a major change in how operating systems work :) The inputs can be assumed to do a ton of I/O, I think it's best to leave the scheduling of that to the OS.

### Interrupting Input Threads

I'd def. have some other way of signaling end of execution to an input before firing an `interrupt` (we can always do that if an input didn't exit after a given amount of time). 
For example, you could simply expose a blocking `push` to the Queue that returns `false` at the end of execution or something like that. It's really hard to write good handling for interruptions imo (or impossible if you say want to do blocking TCP I/O ... there are only bad options here after the interrupt :D), much easier to consume an API that has an explicitly signaled end.

logstash-core/src/test/java/org/logstash/plugin/InputTest.java
jordansissel: I'm in favor of your EventIterator approach. We'll want EventIterator to have `remove` also, eventually (to replace Event.cancel). 

But what about the return type? (Where a processor is allowed to create events, such as the split filter, perhaps we should make this a special case instead of the common case?)

For the Processor return value, we'll have the same problem if someone passes us an EventIterator that is backed by a LinkedList.iterator(), right?

This is specific to the Processor's return type, should it return a collection? a list? a set? a custom type? If not return type, should a Processor become a BiConsumer that has this signature:

```
void process(EventIterator events, EventProducer newEvents) {
...
}
```

where `newEvents` has a small interface to allow processors to create events (like, `newEvents.add(event)` and Logstash takes care of the actual implementation)

Thoughts?

jordansissel:
@jakelandis so having an Input have `void run` and `void stop` is very close to a `FutureTask<Void>` which has `cancel` (aka `stop`), `run` (same), etc. 

Thoughts on that having `interface Input extends FutureTask<Void>` ?

logstash-core/src/test/java/org/logstash/plugin/InputTest.java
original-brownbear: @jordansissel 

>We'll want EventIterator to have remove also, eventually (to replace Event.cancel).

I think I'd just make it `EventIterator  process(EventIterator events)` here and handle `remove` by just callling `next` without pushing an `Event` to the returned `EventIterator`.

Same for adding events. I'd just make this an implementation detail of how a `Filter` implements the `EventIterator` it returns.
We can always offer something like `SimpleFilter` that wraps this process into say `Collection process(Collection<Event> events)` to make it easier to implement things for third party developers (and we need that for existing Ruby stuff anyhow), but in general a `Filter` should simply be a mapping `EventIterator -> EventIterator` imo.

jordansissel:
> or impossible if you say want to do blocking TCP I/O 

Indeed, for this, you need to close the socket from a different thread and this interrupts the blocking call. (This is what we do today, for example)

jakelandis:
@jordansissel 

> Thoughts on that having interface Input extends FutureTask<Void> 

That would address nearly all of my objections, and I do like it better, and am not strongly opposed.

However, that interface exposes methods that would need to be implemented as `do nothing` methods....and writing default `do nothing` methods from a contract has a bit of a smell (imo). Unless the plugin developer gets large benefits from his/her plugin being a FutureTask, I would opt to not use it.

jordansissel:
@jakelandis indeed, looking at FutureTask specifically:

* cancel -  needed, this is how to externally tell the input to terminate.
* get - basically this is conceptually `join` in our FutureTask<Void>
* isCancelled - not sure if needed, unless we want to expose the shutdown state through metrics ("Plugin is shutting down")
* isDone - not sure if needed, unless we want to expose this through metrics ("Plugin is shutdown")
* run - needed

THe rest of the methods are protected:

* done() - maybe not needed since the `run` method completing is a signal for "done"
* runAndReset - do not need
* set - do not need
* setException - probably don't need

I'll have a go at writing up an interface that mimics FutureTask but has only needed methods. If we want to wrap it up as a Callable<Void> or FutureTask<Void> we can do that pretty easily.

logstash-core/src/test/java/org/logstash/plugin/InputTest.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    private class TestInput implements Input {
+        @Override
+        public void run(Consumer<Collection<Event>> consumer) {
+            for (int b = 0; b < batchCount; b++) {
+                int eventCountPerBatch = TestUtil.random.nextInt(100) + 1;
+                Collection<Event> events = new LinkedList<>();
+                for (int i = 0; i < eventCountPerBatch; i++) {
+                    Event e = new Event();
```
guyboertje: Here I am concerned about the separation of the datasource from the event build+decorate functions, how an author provides the two functions and how the inputs internal processing pipeline is built from standard parts and author provided parts (annotations).

something similar to (apologies, knocked up quick like)
```
    public interface Input {
        void run(Consumer<Collection<SourceData>> consumer);

        default Callable<Void> toCallable(Consumer<Collection<SourceData>> consumer) {
            return Executors.callable(() -> run(consumer), null);
        }
    }

    interface SourceData {
        // Toy definition
        SourceData add(String source);
        SourceData set(String key, Object value);
        Object get(String key);
        String source();
        Map<String, Object> details();
        SourceData cloneWithNewSource(String src);
    }

    @FunctionalInterface
    public interface EventGenerator {
        Stream<Event> accept(Stream<SourceData> s);
    }

    @FunctionalInterface
    public interface SourceDataProcessor {
        Stream<SourceData> accept(Stream<SourceData> t);
    }

    private class TestRecordIntoEvent implements EventGenerator {
        @Override
        public Stream<Event> accept(Stream<SourceData> provenanceStream) {
            return provenanceStream.map(provenance -> {
                Event e = new Event();
                e.setField("message", provenance.source());
                e.setField("@provenance", provenance.details());
                e.setField("i", provenance.get("iterationCount"));
                e.setField("b", provenance.get("batchCount"));
                return e;
            });
        }
    }

    private class TestCommaBoundaryDetector implements SourceDataProcessor {
        @Override
        public Stream<SourceData> accept(Stream<SourceData> provenanceStream) {
            return provenanceStream.flatMap(provenance -> Arrays.stream(provenance.source().split(",")).map(
                    provenance::cloneWithNewSource
            ));
        }
    }

    private class Provenance implements SourceData {
        // Toy impl
        // need an inner map of provenance key & values
        // a source string
        @Override
        public SourceData add(String source) { return this; }
        @Override
        public SourceData set(String key, Object value) { return this; }
        @Override
        public Object get(String key) { return null; }
        @Override
        public String source() { return null; }
        @Override
        public Map<String, Object> details() { return new HashMap<>(); }
        @Override
        public SourceData cloneWithNewSource(String src) { return new Provenance().add(src); }
    }

    private class TestInput implements Input {
        @Override
        public void run(Consumer<Collection<SourceData>> consumer) {
            int batchCount = 100;
            for (int b = 0; b < batchCount; b++) {
                int eventCountPerBatch = 20;
                Collection<SourceData> rawEvents = new LinkedList<>();
                for (int i = 0; i < eventCountPerBatch; i++) {
                    SourceData p = new Provenance();
                    p.add("hello,world,how,about,that,then");
                    p.set("iterationCount", i);
                    p.set("batchCount", b);
                    // other provenance settings
                    rawEvents.add(p);
                }
                consumer.accept(Collections.unmodifiableCollection(rawEvents));
            }
        }
    }
```

logstash-core/src/main/java/org/logstash/plugin/Processor.java
jordansissel: I agree we can improve the object that an input uses to shuttle data into the pipeline. I think that is out of scope for this PR as it changes existing code.

After this PR (just the plugin interfaces  and any necessary support code) land, we can work on the next step of making Java plugins runnable in the pipeline -- and at that time, we can make the move to this EventIterator (or whatever is designed).

logstash-core/src/test/java/org/logstash/plugin/InputTest.java
jordansissel: This proposal makes sense to me -- having inputs produce "source data" (raw data) attached to some kind of Provenance (where the data came from).

Then the input ships this to some pipeline to pass it to the queue:

* input (source data) -> mill (milled data) -> decoder (decoded data) -> assembler (Event object) -> queue
* input (source data) -> mill (milled data) -> queue -> decoder (decoded data) -> assembler (Event object)

^^ seem like two options we could pursue. The 'Assembler' would take the decoded data plus the provenance and assemble an Event that the pipeline would handle?

*alternately* We maybe drop the idea of a formal "Event"?

jordansissel:
I've added `shutdown()` to the Input interface and I spiked out a `Task` class to try and hide the "Thread" a bit from the actual Input.

jordansissel:
The next part I am going to study is how inputs figure out when there is work to do and see if there's any way we can consolidate.

For example, the file input does discovery/prospecting, then it, in the same thread, will do file reads. Instead, it could take the prospecting result and form jobs (File X, position Y, length Z) and send them to an ExecutorService to get done. Similarly for other plugins that have a way of determining if work is available:

1) File input "new files found" or "known files grew"
2) TCP input "socket is ready to read"
3) JDBC input "it's time to do scheduled jdbc query" and possibly sharding the work to split a JDBC query into multiple queries
4) Elasticsearch input: scheduled, work sharding with index shards, possibly.
5) Kafka input: has its own complex client we just setup and run
6) beats input: socket has data ready to read

logstash-core/src/test/java/org/logstash/plugin/InputTest.java
guyboertje: There is the question of how a plugin author codes an assembler with special handling, because while they can add certain key values to the provenance, there may still be the need to author the assembler which needs to be discoverable and inserted at the correct point (pre/post queue).

guyboertje: Also, I have not seen many cases of composeable functions that take a stream and return a stream. However, this StackOverflow Q+A advocates for it. https://stackoverflow.com/a/24679745/5349531

I struggled with the idea of simply providing the function for `map` but then the CommaBoundaryDetector, being a split function, produced additional elements with the original provenance cloned with new source. The only (hurried) way I could think of is to take and return a stream.

This means that the Queue `accept` method would need to be a terminal function.

guyboertje: Perhaps the Regex `Pattern.splitAsStream` would be a better choice.

However we have to handle the case where a value in a stream can produce a stream of values - and we have `flatMap` for that. But the knowledge of whether to use map or flatMap is only known to the function itself.

logstash-core/src/main/java/org/logstash/plugin/Processor.java
original-brownbear: @jordansissel yup makes sense, sadly enough the current Java execution is still `Collection` based.

The cool thing is that we can try out this approach internally in the Java execution first (make `RootDataset` and `OutputDataset` into adapters and then just remove both of them (those two are kind of awkward anyway, the `Dataset` approach only seems to be the correct abstraction for `If` and `Filter`) once it worked out).

jordansissel:
I'm going to split the `Input` concerns into a separate PR with @guyboertje taking ownership of the Input side.

Two general features remain here:

1) how to attach java plugins ("universal plugins") to Logstash 
2) how to implement java filters and outputs in Logstash

I'll also split these two items into separate PRs:

I'll close this with the following next steps:

1) @guyboertje takes lead on research/designing/implementing the new Input API
2) I'll make a separate PR focusing on Plugin discovery/registration/loading
3) I'll make a separate PR focusing on Filters and Outputs (custom code that runs inside a pipeline worker)

jordansissel:
Closing this to split the work into separate more focused PRs as stated above.

PR #8228, opened 2017-09-13
WIP Version bump to 6.0.0-rc1

andrewvc:
I don't think this PR is necessary, we already have the versions.yml set to 6.0.0-rc1


suyograo:
LGTM. We need this PR because of the lock file.

PR #8229, opened 2017-09-13
Java API: Add mechanism for configuring an input/processor/output with a confg map

jordansissel:
The idea behind ConstructingObjectParser was largely lifted from
Elasticsearch's class by the same name. The reasons I liked this approach was that it provided a type-safe way to define how user input (a Map, in Logstash's case) is parsed to create an object like an input plugin.

The ConstructingObjectParser roughly implements in Java what `config/mixin.rb` provides for Ruby plugins.

This class allows us to take a Map that comes from the config parser and
build a pipeline plugin (input, filter, output).

For usage examples, see:

* [ExampleProcessor](https://github.com/elastic/logstash/pull/8229/files#diff-3676864fa032faff81207add3740e776)
* [ExampleInput](https://github.com/elastic/logstash/pull/8229/files#diff-52eaa8ea57c1b00aa2d3a229c3d2c1ab)

Differences from the ruby api:

* Custom object validation is possible. Example, `http_poller` has a map of `urls` which are `name => { configuration }`, and we can achieve type-safe validation of this `urls` thing which benefits us as plugin authors and benefits users through better error reporting.
* Explicit `default` value is not required because Java class fields already allow you to set default values :)

Request for comments:

* Is construction api clear? Would `declareString` be better than `string` for method name?
  * `thing.string("foo")` meaning a string field in constructor argument
  * `thing.string("foo", Thing::setFoo)` meaning a string field that calls setFoo
  * constructor argument ordering is implicit
* Validation failure currently throws IllegalArgumentException[1]. Should we make our own exception class for this?
* I am experimenting with JUnit features (`Enclosed` and `Parameterized` test runners). Do you like either of these?

[1] I chose IllegalArgumentException because it is an unchecked exception we can throw while still satisfying the constraints of the java.util.function interfaces (BiFunction, etc)

Todo: 

* [ ] Add support for other base types (boolean, float, double, etc)
* [ ] Add support for typed lists like `stringList("hosts", Elasticsearch::setHosts)`, etc
* [ ] Add `obsolete` and `deprecated`

This PR depends on #8227


jordansissel:
Closing; will reopen against master.

PR #8232, opened 2017-09-13
MINOR: Clean up some dead code around PQ + PQ Benchmarks

original-brownbear:
Just some randomness I found researching the deadlock issue. Trivial stuff that is just shown to be right by still compiling.

Except for maybe adjusting the Exception in `org.logstash.ackedqueue.io.ByteBufferPageIO#ByteBufferPageIO(int, byte[])`:
* `IOException` is just wrong here, the arguments we get are not the result of any IO operation, in fact in production `initalbytes` is always `byte[0]` anyway => it's an `IllegalArgumentException` and we can save ourselves the upstream `IOException` handling :)


suyograo:
LGTM

PR #8239, opened 2017-09-13
 #8132 Remove JavaProxyBiValue without replacement

original-brownbear:
handling #8132 (see the issue for why there is no replacement for this)

@jordansissel this needs your review I think. If we get this and #8205 in, then `BiValue` is gone :) (except for one last step killing the `nil` one :))


original-brownbear:
urgh can't do it this way we have to unwrap these and pass them to the normal path conversion path apparently we're receiving at least `org.logstash.Timestamp` wrapped in a proxy in the ITs -> closing this ... have to research why this is happening tomorrow.

original-brownbear:
@jordansissel fixed, this is ready for review now :)

jordansissel:
Hmm, somehow I didn't see this until now. Will review!

PR #8240, opened 2017-09-13
Java API: Add mechanism for configuring an input/processor/output with a confg map 

jordansissel:
The idea behind ConstructingObjectParser was largely lifted from
Elasticsearch's class by the same name. The reasons I liked this approach was that it provided a type-safe way to define how user input (a Map, in Logstash's case) is parsed to create an object like an input plugin.

The ConstructingObjectParser roughly implements in Java what `config/mixin.rb` provides for Ruby plugins.

This class allows us to take a Map that comes from the config parser and
build a pipeline plugin (input, filter, output).

For usage examples, see:

* [ExampleProcessor](https://github.com/elastic/logstash/pull/8229/files#diff-3676864fa032faff81207add3740e776)
* [ExampleInput](https://github.com/elastic/logstash/pull/8229/files#diff-52eaa8ea57c1b00aa2d3a229c3d2c1ab)

Differences from the ruby api:

* Custom object validation is possible. Example, `http_poller` has a map of `urls` which are `name => { configuration }`, and we can achieve type-safe validation of this `urls` thing which benefits us as plugin authors and benefits users through better error reporting.
* Explicit `default` value is not required because Java class fields already allow you to set default values :)

Request for comments:

* Is construction api clear? Would `declareString` be better than `string` for method name?
  * `thing.string("foo")` meaning a string field in constructor argument
  * `thing.string("foo", Thing::setFoo)` meaning a string field that calls setFoo
  * constructor argument ordering is implicit
* Validation failure currently throws IllegalArgumentException[1]. Should we make our own exception class for this?
* I am experimenting with JUnit features (`Enclosed` and `Parameterized` test runners). Do you like either of these?

[1] I chose IllegalArgumentException because it is an unchecked exception we can throw while still satisfying the constraints of the java.util.function interfaces (BiFunction, etc)

Todo: 

* [ ] Add support for other base types (boolean, float, double, etc)
* [ ] Add support for typed lists like `stringList("hosts", Elasticsearch::setHosts)`, etc
* [ ] Add `obsolete` and `deprecated`


jordansissel:
In implementing the other types (boolean, float, etc), I had forgotten that primitives are keywords in Java, so youc an't have a method called 'float'. So I will move everything to be `declare[Type](...)` like `declareFloat`

jordansissel:
I spiked out an implementation of the `translate` filter: https://github.com/elastic/logstash/blob/c015a69bbadc3ce9567df2021a7818f4ff27fa1e/logstash-core/src/test/java/org/logstash/plugin/TranslateFilter.java#L14-L57

The Ruby API `config :dictionary, :validate => :hash` etc are now using `ConstructingObjectParser` features (defined in the top of the TranslateFilterPlugin class)

jakelandis:
@jordansissel and I spoke on this subject on Slack, and came to some conclusions around the requirements:

* Must be easily consumed (read and write) by Plugin authors. 
* Should be strongly typed. e.g. don't push everything as an Object or String.  
* Must support a large number of configuration options in a manner that does not adversely impact read/write-ability for plugin authors. For example a constructor that takes in 20+ arguments is a no-go.
* Should support custom transformations to a different (stronger) type. Since the configuration language is not strongly typed, but the configuration should be, we should support the ability for plugin developers to transform types. For example, "http://localhost:9200" _should_ be URL type, but when read from the configuration it is a String. We should allow the plugin developer to express this configuration in his/her best suited typed via a custom transformation of type.
* Should support custom validation, with knowledge of entire configuration set. For example in the Translate filter it is invalid to have both a dictionary and a dictionary_path defined.  We should provide hooks such that a plugin developer may error early and gracefully in that case.
* Must support the ability for the plugin developer to express configuration as obsolete and deprecated withOUT adversely affecting the read/write-ability.  
* Should support some form of core/re-usable validation, such as non-null, or isAURL. Common types of validation that assist the plugin developer to ensure sane input. 
 * Should support some form of core/re-usable transformations. For example, converting a String to URL is a common pattern and we shouldn't ask plugin developers to re-code it every time (no matter how simple).
* Should support by some means a way to express Logstash core configuration to plugins. For example the number of workers may influence the way plugin should work or is configured without having to duplicate the configuration specifically for the plugin. 
* Should support by some means a way to configure common categories of things, such http clients or SSLContext's.  Sane defaults will go along way to a healthier, more secure system overall.  (note - this is a different type of configuration and may result in a different strategy)
* It would be nice if the implementation for the configuration can be re-used for the settings. Configuration and settings are very similar, and less code the better.

jakelandis:
skip ci

jordansissel:
I think this is ready for review. It may be best to pair up on this one in order to navigate the functionality.

I have written a "sample filter" which demonstrates the use of the ConstructingObjectParser and in a way that I think we will use when writing Logstash plugins (See TranslateFilterPlugin.java). The Translate filter is a bit of a complex plugin because it has model operations (depends on whether 'dictionary' or 'dictionary_path' are set).

* Authors will be interacting primarily with ConstructingObjectParser which takes a config map and builds an object.
* Strong typing is applied where possible, for example, if you have a setter take a string, it is invalid to try and apply an integer field to it. (It wouldnt' compile, which is nice)
* Tests are included for a variety of use cases.

Still todo is documentation on how to use it, but I will write that in a future PR with the Plugin API PR.

jordansissel:
Very rough sketch to see how another complicated filter (mutate) would exist in this new world --- 

https://github.com/elastic/logstash/pull/8240/files/a67644f75c918e34f396211c23b546f687f897bd..590584f8ef3d33201c295eb24f2f8546d001c84c

Major concept here is that `mutate` can compile to a chain of `Consumer`s when the plugin is loaded, so we can try doing that.

jordansissel:
@jakelandis and I spend lots of time talking about this project.

We compare annotations-vs-factories pretty frequently. Annotations some really nice readability properties that factories simply don't have.

I'm leaning Jake's suggestion of declaring settings and capabilities with annotations. In fact, the last sketch we drew on this topic looks identical to how Graylog's plugins are configured -- these use a mix of https://github.com/joschi/JadConfig and Java EE Inject.

I'll have a go sketching a few more plugins using annotations to see how it feels.

PR #8246, opened 2017-09-13
remove unsued windows code

colinsurprenant:
per discussion with @ph - I tested that this code is not needed anymore. everything builds and all tests passes without it on Windows.


colinsurprenant:
merged in master, 6.x and 6.0

Issue #8224, opened 2017-09-12
Modules should work with password enabled Elasticsearch


robbavey:
With x-pack configured with security, Logstash is unable to connect to the Elasticsearch instance using the recommended command-line configuration for modules - ie 
```
bin/logstash --modules arcsight --setup -M "arcsight.var.inputs=smartconnector"
 -M "arcsight.var.elasticsearch.username=elastic" 
 -M "arcsight.var.elasticsearch.password=very_secure" 
 -M "arcsight.var.kibana.password=very_secure" 
 -M "arcsight.var.kibana.username=elastic"
```

Logstash will loop, attempting to connect to Elasticsearch, but failing to authenticate each time, logging the following error:
```
[2017-09-12T17:15:54,812][INFO ][logstash.outputs.elasticsearch] Running health check to see if an Elasticsearch connection is working {:healthcheck_url=>http://elastic:xxxxxx@localhost:9200/, :path=>"/"}
[2017-09-12T17:15:54,903][WARN ][logstash.outputs.elasticsearch] Attempted to resurrect connection to dead ES instance, but got an error. {:url=>"http://elastic:xxxxxx@localhost:9200/", :error_type=>LogStash::Outputs::ElasticSearch::HttpClient::Pool::BadResponseCodeError, :error=>"Got response code '401' contacting Elasticsearch at URL 'http://localhost:9200/'"}
```

robbavey:
@suyograo This is the issue I referred to in the xpack issue you raised this morning - https://github.com/elastic/x-pack-logstash/issues/233

PR #8258, opened 2017-09-14
Fix double password wrap

robbavey:
Password settings for Modules supplied via a yml file were being wrapped inside a
LogStash::Utils::Password object twice during startup - once in from_yaml, and once
in validate. Disallow this from happening by skipping the wrap if the password is
already wrapped

This should complete the fix for #8224


suyograo:
LGTM

PR #8264, opened 2017-09-14
Fix Modules doc incorrectly doc referring to Kibana

robbavey:
var.elasticsearch.ssl.certificate_authority should not refer to Kibana in the
description of the value.


dedemorton:
Ruh roh. LGTM.

PR #8268, opened 2017-09-14
Revert "Add jvm option ExitOnOutOfMemoryError"

### LINKED ISSUE #8254: Revert 'ExitOnOutOfMemoryError' setting because it requires minimum Java 8u92
our support matrix targets Java 8u60, but 'ExitOnOutOfMemoryError' was only added in Java 8u92.

http://www.oracle.com/technetwork/java/javase/8u92-relnotes-2949471.html

We need to revert ExitOnOutOfMemoryError as long as we don't require 8u92 or higher.

jakelandis:
This reverts commit a14fbfc3e21e4a8662af2440074c3a3f37a3f370.

Fixes #8254

### DIFF
--- a/config/jvm.options
+++ b/config/jvm.options
@@ -62,9 +62,6 @@
 # heap dumps are created in the working directory of the JVM
 -XX:+HeapDumpOnOutOfMemoryError
 
-# exit on out-of-memory error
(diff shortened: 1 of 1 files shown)


suyograo:
LGTM

PR #8272, opened 2017-09-15
 #8271 fix one excess '/' added to log config file uri on *nix in runner.rb

### LINKED ISSUE #8271: Settings ITs are Broken
https://github.com/elastic/logstash/commit/e9a8812317a27abfee2cf250d4140f67dd15c88b as seen in https://travis-ci.org/elastic/logstash/jobs/275628276#L7767 broke settings ITs:

```sh
Finished in 33 minutes 45 seconds (files took 6.98 seconds to load)
38 examples, 4 failures
Failed examples:
rspec ./specs/settings_spec.rb:56 # Test Logstash instance whose default settings are overridden should write logs to a new dir
rspec ./specs/settings_spec.rb:125 # Test Logstash instance whose default settings are overridden start on a different HTTP port
rspec ./specs/monitoring_api_spec.rb:100 # Test Monitoring API can configure logging
rspec ./specs/slowlog_spec.rb:33 # Test Logstash Slowlog should write logs to a new dir
```

original-brownbear:
Fixes #8271 

On Windows, we have `file:///c:/WINDOWS/file.txt` and absolute file seems to return `c:/...`, but on *nix we were adding four `////` since absolute file already brought its own leading `/` :)

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -251,7 +251,8 @@ class LogStash::Runner < Clamp::StrictCommand
       log4j_config_location = ::File.join(setting("path.settings"), "log4j2.properties")
 
       # Windows safe way to produce a file: URI.
-      LogStash::Logging::Logger::reconfigure(URI.join("file:///" + File.absolute_path(log4j_config_location)).to_s)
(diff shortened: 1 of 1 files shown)


original-brownbear:
@ph thanks merging then (tests are green, the failed one on Travis was some Travis bug, it never started but one IT run is green which was impossible before so all good imo)

PR #8275, opened 2017-09-15
#7104 Fix Queue locking up on pagesize == queuesize

original-brownbear:
Fix for #7104 : 

`this.maxBytes > 0L && this.currentByteSize >= this.maxBytes` is does not mean "full", because `currentByteSize` counts the overall size of allocated pages.
This means, that currently the "last" page is broken because it will always trigger `isFull()` to return `true`.

* Fixed as documented in the code and reproduced in the added UT.
* Removed the todo in that method, `volatile` is fine as long as there are no concurrent writes which there aren't
* Fixed the tests (they were obviously failing now) that had a certain miss-match between expected sizing according to comments and actual sizing :)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -24,7 +24,6 @@ import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
-
 // TODO: Notes
 //
 // - time-based fsync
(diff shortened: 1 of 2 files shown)


colinsurprenant:
I will get to this shortly.

original-brownbear:
@robbavey @jakelandis could one of you please help review here? After talking to @andrewvc it was decided that we want this in 5.6 so there is a bit of urgency here.
I'm happy to go through the details of the fix if anything is unclear.

Thanks!

jakelandis:
@original-brownbear - 👀 now

robbavey:
@original-brownbear Ran this locally, and it seems to work as advertised.

One thing: These comments in the SettingsImpl.BuilderImpl are somewhat confusing as to what capacity and maxBytes are, could do with some clarification - specifying page capacity, and using _bytes_ in place of _events_ might clear up a little:

```
         * The default Queue has a capacity of 0 events, meaning infinite capacity.
         * todo: Remove the ability to set infinite capacity.
         */
        private static final int DEFAULT_CAPACITY = 0;

        /**
         * The default Queue has a capacity of 0 bytes, meaning infinite capacity.
         * todo: Remove the ability to set infinite capacity.
         */
        private static final long DEFAULT_MAX_QUEUE_BYTES = 0L;
```

original-brownbear:
@robbavey I'm with you 100% on the settings stuff doc wise. I'll make a separate PR for that ok? :)

original-brownbear:
@jakelandis Np + I have to thank you !:)

Issue #8235, opened 2017-09-13
[Meta] Move from rake to gradle to build Logstash


ph:
We want to move from rake to Gradle to achieve the following goals:

- Be a better java citizen
- Easier to create project dependencies
- Reduce dependencies needed to build logstash
- Faster build
- Eliminate double launch of ruby.

**Steps required:**

- [ ] Download JRuby with gradle instead of Rake https://github.com/elastic/logstash/issues/8234
- [ ] Download the minimal gems with gradle
- [ ] Remove the rake `test:install-core`
- [ ] Make gradle invoke _bundle install_.
- [ ] Evaluate the **rake artifact** build tasks, we have done a lot of work with it and FPM has been great so far.
- [ ] Fix environment initialization and loading https://github.com/elastic/logstash/issues/8207

ph:
I have great progress on this, I've cleanup most of the rake tasks related to downloading and installing gems. By reducing the bundler calls I've improved the bootstrap speed from 310s to 210s.

PR #8277, opened 2017-09-15
Remove uneeded jar dependency task

ph:
This task is run in the build process but it doesn't produce any
actionnable result. The only plugin that this task is targetting is the
input kafka. We have in place an integration test with that plugin, if
the plugin is missing jars or is broken we will catch it there.

Reference: #8235


jakelandis:
LGTM, but lets wait on CI to be sure.

ph:
I've retriggered the travis suite.. the suite was failling because of the issue to fetch jruby packages.

kares:
seems the "install-jar-dependencies" task has been remove by now

PR #8284, opened 2017-09-15
Remove tests dependency on filter multiline

ph:
Some of the tests were still depending on the multiline filter; this
filter is not shipped anymore with Logstash. I've kept the same logic
but created a simple dummy filter.

This is required for the effort of cleaning up the Logstash bootstrap script.


original-brownbear:
Jenkins test this please

original-brownbear:
@ph LGTM :)

PR #8291, opened 2017-09-17
 #4640 make Event#clone clone metadata as well

### LINKED ISSUE #4640: `LogStash::Event#clone` should also clone the `@metadata`

original-brownbear:
Fixes #4640 

* Add cloned metadata to cloned data map
* Copy (and slightly adjust for 5.x) tests from https://github.com/elastic/logstash/pull/5656/files#diff-42cd4d2f6f981839921777c8e4a5ba2bR662

### DIFF
--- a/logstash-core/spec/logstash/event_spec.rb
+++ b/logstash-core/spec/logstash/event_spec.rb
@@ -352,4 +352,61 @@ describe LogStash::Event do
       expect { subject.baz() }.to raise_error(NoMethodError, /undefined method `baz' for/)
     end
   end
+
+  describe "#clone" do
+    let(:fieldref) { "[@metadata][fancy]" }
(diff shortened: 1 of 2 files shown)


original-brownbear:
@jsvd hmm damn I was gonna ask you the reverse, "should we backport this to 6.0?" :D 
My reasoning was that in #4640 this was seen as an outright "bug" and I'm just fixing a bug here. 
I don't think I know enough to make that call here though :(

jsvd:
I'm ok with calling this a bug - the issue and this PR are indeed marked as `bug` - which means we should place the fix in 6.0 AND 5.6, so that a 5.6.1 will have it (if it's ever released)

original-brownbear:
@jsvd alright then, (last chance :D)  I'll go ahead and merge + backport?

jsvd:
yes please, LGTM (please add the other two version tags)

PR #8292, opened 2017-09-18
PERFORMANCE: Provide some typed setters to Java Event API

original-brownbear:
This should give us a bit of a boost when in Java plugins where we mostly know the type of what we're setting and don't need to go through the whole `Valuefier` magic to figure out the correct type.
Only added these overrides for the types I could find in actual plugins, for now, I guess we can add more as we need them since this change doesn't break binary compatibility.


original-brownbear:
putting this in the general queue, since it's an API compatible change

jordansissel:
```
[before] EventFieldBenchmark.setByFieldRef  thrpt   20  13.401 ± 0.662  ops/us
[after] EventFieldBenchmark.setByFieldRef  thrpt   20  14.227 ± 0.990  ops/us
```

Your results are within the error margins for this test. Are you certain this is a speedup? An average improvement of 0.8 ops per microsecond" is *extremely* small clock measurements especially given the plus-minus is .6 to .9 for those tests.

I'm not trying to dispute the conclusion, but something seems fishy about the test results so it's hard to derive the same conclusion you have.

I can try to run the tests, but if we are making a performance improvement claim, even if the code works, I want to verify the claim before we forward that claim to users.

For user experience, I'm not expecting anything measurable for users given we're on the scale of operations-per-microsecond and event rates are generally on the events-per-millisecond scale (factor of 1000 difference in scale). Thoughts?

jordansissel:
In terms of implementation, I think we can shorten the code by using dynamic type inference for both setField methods.

Can these String references be simplified (instead of duplicated) with dynamic type inference and make `Accessors.set` be what takes different types?

```
public <Value> void setField(final String reference, final Value value) {
  setFIeld(PathCache.cache(reference), value);
}

public <Value> void setField(final FieldReference field, final Value value) {
  Accessors.set(data, field, value);
}
```

For the special cases you have for META_PARENT not being a Map, you could generic on `<Value extends Map<String, Object>>` and then handle the common case in a less-specific generic `Value`

Then for `Accessors` class implement per-type setters if necessary?

My thought is that using dynamic types would allow the code to be much shorter.

original-brownbear:
@jordansissel at one second per run and 1M events per pass:

```sh
Benchmark                           Mode  Cnt   Score   Error   Units
EventFieldBenchmark.setByFieldRef  thrpt   20  15.351 ± 0.506  ops/us
EventFieldBenchmark.setByString    thrpt   20  14.077 ± 0.565  ops/us

Benchmark                           Mode  Cnt   Score   Error   Units
EventFieldBenchmark.setByFieldRef  thrpt   20  12.908 ± 0.686  ops/us
EventFieldBenchmark.setByString    thrpt   20  12.565 ± 0.521  ops/us
```

The result will be even better in many practical plugin situations, because why you get is that all your setters compile super small now instead of compiling to inlining `Valuefier` :)

original-brownbear:
@jordansissel 

> and I think the compiler is smart enough to make the right call.

No, it's not btw. How would it be able to do that, dynamic type interference cannot be in-lined.
See here https://wiki.openjdk.java.net/display/HotSpot/MethodData#MethodData-Poll, it contains this line:

> generic type reification (when pollution comes from a type parameter; not implemented)

not implemented yet ... In theory, it would be possible to optimize generic type sites the way you think, but the JVM is not there yet.

original-brownbear:
@jordansissel in fact, you can even observe the "compiles smaller" and "inlines" better effect in my better numbers above. The difference between `byFieldRef` and `byString` is growing by doing this, simply by inlining more stuff in the `byField` case, that could previously not be inlined in the `Object` setter.

original-brownbear:
@jordansissel any update here? :)

I still very much stand by this approach. Outside of just performance and compiling smaller this also seems like a nice way of making our APIs safer.
We could simply implement all valid types as setters this way and deprecated the `Object` typed one. This would be much easier to work with than eventually get `throw`n some `Exception` if you `set` the wrong type (obviously this approach still breaks for setting `Map` and `Collection` type fields with the way Java generics work, but still it's a step forward here).

original-brownbear:
Test failure is the same old #8142

jordansissel:
> Outside of just performance and compiling smaller this also seems like a nice way of making our APIs safer

I am +1 on safety. if performance is a critical concern, I'd like to validate the claim that dynamic type inference prevents inlining, and the document you linked is last-updated 4 years ago, so it's hard to trust.

jordansissel:
@original-brownbear I tested the dynamic-inferences-can't-inline claim and found it to be not true?

https://gist.github.com/jordansissel/fd7a4ea295760b5657e2e8f2dcae1e85

I have a method `<T> set(T value)` and typed method `setLong(Long value)` and both are inlined according to the JVM:

```
                              @ 51   X::set (6 bytes)   inline (hot)
                              @ 56   X::setLong (6 bytes)   inline (hot)
```

See the gist link above for the code.

original-brownbear:
@jordansissel 

```java

public class X {
  private Object value;
  public <T> void set(T value) {
    this.value = value;
  }

  public void setLong(Long value) {
    this.value = value;
  }
}
```

This is not a relevant example. You're not invoking any methods on `value`, so since it's only assigned and its type is irrelevant from the compiler's perspective and it just links to `Object` type. Only when you invoke methods on the input (which we will have to do for conversions) will you see a benefit/cost from this effect. See below example: 

What cannot be inlined are situations like this:

screen shot 2017-10-26 at 21 40 53

3 types go into `method` hot -> it will only make the virtual call to `Object`.

Note a few things:

* An example that just involves a single type for a callsite is always inlined, since the type profiler will identify it as a monomorphic callsite eventually.
* The behavior of a generic call site hit by 2 different types will be optimized as well (though not as well as for just a single type - see bimorphic callsite), you need 3 or more types to actually get the return to virtual calls
* Generics are not optimized for in JDK8, see the disassembly not containing any bytecodes that are typed. There are no compile time optimizations happening on generics whatsoever. No JITing either, the JIT just does type counting and decides not to optimize since it sees 3 different types

Also see updated benchmarks I posted eventually https://github.com/elastic/logstash/pull/8292#issuecomment-331060645, we win regardless since we're not even working on a common interface. If we don't have typed setters our type conversions will inevitably require us to either use reflection to `getClass()` or chain `instanceof` calls to branch based on type.
This kind of thing cannot be inlined since each method is only compiled once.

Hope this helps :)

original-brownbear:
@andrewvc @danhermann I rebased this after our chat about it yesterday and I think it's a pretty clean performance gain.
Unfortunately, there is one problem here that I didn't take into account. If we make this change and start compiling plugins against it to take advantage of it ... those plugins won't work with older LS versions before this change. So this would create a clear compatibility boundary in one direction (those would also happen without any changes to the plugin code ... making things a little more confusing) :(

I strongly believe we should make this change in general. But maybe if we decide to actually do this we should take the opportunity for additional `Event` API changes since we're breaking compatibility either way and should probably only do it once?

Issue #8073, opened 2017-08-25
[test failure: master] Test Logstash Slowlog should write logs to a new dir


colinsurprenant:
Integration test non-related failure.

https://travis-ci.org/elastic/logstash/jobs/268160331

```
Failures:
  1) Test Logstash Slowlog should write logs to a new dir
     Failure/Error: expect(IO.read(slowlog_file).split("\n").size).to eq(2)
     
       expected: 2
            got: 1
     
       (compared using ==)
     # ./specs/slowlog_spec.rb:44:in `block in (root)'
     # /home/travis/.rvm/gems/jruby-9.1.10.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in (root)'
Finished in 25 minutes 26 seconds (files took 9.44 seconds to load)
32 examples, 1 failure
Failed examples:
rspec ./specs/slowlog_spec.rb:33 # Test Logstash Slowlog should write logs to a new dir
Randomized with seed 29073
```

colinsurprenant:
also reported in https://travis-ci.org/elastic/logstash/jobs/266095127 per #8033

original-brownbear:
@jakelandis FYI (since I know you're working on the tests), this one is a reproducible error for me locally too and not a result of a lack of processing power.
It also shows up as `3` counted instead of the expected `2` for me locally.

original-brownbear:
I think it's this https://github.com/logstash-plugins/logstash-filter-sleep/issues/7 ??

jakelandis:
@colinsurprenant / @original-brownbear 

I (accidentally) logged a duplicate over at https://github.com/elastic/logstash/issues/8322 and submitted a PR https://github.com/elastic/logstash/issues/8322 to fix it. 

The test itself is just trying to assert that it actually writes to the file, the actual count is a bit incidental.  I _think_ too little CPU and the you can get '1' instead of '2' and too much (if there is such a thing) can result in counts higher then '2'.  I trying >= 2 to see if that can stabilize it...if not we will have to do >= 0 or do a fuzzy match. 

Closing this in-favor of https://github.com/elastic/logstash/issues/8322

original-brownbear:
@jakelandis 

>  I think too little CPU and the you can get '1' instead of '2' and

[...]

PR #8294, opened 2017-09-18
MINOR: Remove needless timestamp reference in Event

original-brownbear:
We don't need the `timestamp` field in `Event` since `TIMESTAMP` has to be set on `data` for us to return its value. There is no benefit of caching a reference to it since `containsKey` costs the same that `get` costs.
Also, we make the time-stamp easier to GC, make inlining the constructor of slightly `Event easier` and save 4 bytes per `Event` (:P)


original-brownbear:
@jsvd thanks :) the one travis failure is #8073

PR #8297, opened 2017-09-18
 #4640 make Event#clone clone metadata as well

### LINKED ISSUE #4640: `LogStash::Event#clone` should also clone the `@metadata`

original-brownbear:
Fixes #4640 for 5.6

 Removed files are due to a gitignore change in master and fully valid to remove (they are in the out folder than Idea creates when running a Gradle project).


original-brownbear:
Jenkins test this please

original-brownbear:
merging same test failures are also found in `5.6`, will add an issue for that in a sec

original-brownbear:
Test failures tracked in https://github.com/elastic/logstash/issues/8331

PR #8298, opened 2017-09-18
Remove unecessary dependency `file-dependencies`

ph:
To understand why we can now remove this gem we have to go back at the
history of LS 1.5 and the choice we made back them. In the begining
plugins depending on external files like the `user-agent` or the `filter-geoip`
were not bundling theses files in the gem.

The `file-dependencies` was providing a new hook when plugins were
installed to trigger the file download.

In the context of making the plugins work in an offline environment we
need to make sure that every plugins external resources are present in
the gem.

We have currently a few plugins that require external files and they all
bundle it. So I think its safe to remove that feature, also *the hook was
not triggered correctly. So it was a NOOP.*

### DIFF
--- a/Gemfile.template
+++ b/Gemfile.template
@@ -7,7 +7,6 @@ gem "logstash-core-plugin-api", :path => "./logstash-core-plugin-api"
 gem "paquet", "~> 0.2.0"
 gem "ruby-progressbar", "~> 1.8.1"
 gem "builder", "~> 3.2.2"
-gem "file-dependencies", "0.1.6"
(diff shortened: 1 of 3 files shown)


ph:
>  ⚙  ph@sashimi  ~/es/logstash   feature/remove-file-dependencies  ag file-dependencies

Nothing else is referencing it, even the libraries which should use it like the *useragent* or the *geoip* are using the rake *vendor* task defined in https://github.com/elastic/logstash-devutils/blob/master/lib/logstash/devutils/rake/vendor.rake

PR #8300, opened 2017-09-18
Fix reference to incompatible kramdown rubygem version

robbavey:
Change version of Kramdown in Gemfile.jruby-1.9.lock.release from
1.15 - it is incompatible with Ruby < 2.0


jsvd:
this seems to be addressed in https://github.com/elastic/logstash-devutils/commit/750ab22377a6b78dcde62bdaae72bbf085c661a6, maybe we should bump the minimum version of devutils?

robbavey:
This change is superseded by #8310 and is no longer necessary. Closing

Issue #8293, opened 2017-09-18
Persistent queues exception causes data loss/pipeline stop


randude:
- Version: Logstash 6 beta 1
- Operating System: ubuntu 16.04

I've got two pipelines. Pipeline 1 gets inputs from several locations and sends some of them (after some modification) to pipeline #2 which outputs it to the elasticsearch output.
The seconds I enable the persistent queue on *any* of these pipelines, data isn't sent to ES. The queues aren't full and are defined as 1gb in size. The second i disable the queues, data is being sent again to ES.
So, I know that the pipelines aren't stuck BUT the elasticsearch output pipeline isn't sending the data and keeps writing them to disk until at some point the queues are full and new events are ignored.

I saw in debug mode the following exception:

@task=#<Concurrent: : TimerTask: 0x14741235@observers=#<Concurrent: : Collection: : CopyOnNotifyObserverSet: 0x2cbe9635@observers={
#<LogStash: : Instrument: : PeriodicPoller: : PersistentQueue: 0x3075bb12...>=>: update
}>,
@timeout_interval=120.0,
@running=<#Concurrent: : AtomicBoolean: 0xfc8value: true>,
@StoppedEvent=#<Concurrent: : Event: 0x10d68f97@set=false,
@iteration=0>,
@execution_interval=5.0,
@do_nothing_on_deref=true,
@run_now=nil,
@freeze_on_deref=nil,
@executor=#<Concurrent: : SafeTaskExecutor: 0x2487c202@task=#<Proc: 0x7cd0f799@/usr/share/logstash/logstash-core/lib/logstash/instrument/periodic_poller/base.rb: 68>,
@exception_class=StandardError>,
@StopEvent=#<Concurrent: : Event: 0x604c04f4@set=false,
@iteration=0>,
@value=nil,
@copy_on_deref=nil,
@dup_on_deref=nil>,
@metric=#<LogStash: : Instrument: : Metric: 0x7f1c77f4@collector=#<LogStash: : Instrument: : Collector: 0x69413f62@agent=nil,
@metric_store=#<LogStash: : Instrument: : MetricStore: 0x6e26f702@store=#<Concurrent: : Map: 0x00000000000fb8entries=4default_proc=nil>,
@structured_lookup_mutex=#<Mutex: 0x35fc65f>,
@fast_lookup=#<Concurrent: : Map: 0x00000000000fbcentries=82default_proc=nil>>>>,
@options={
: polling_interval=>5,
: polling_timeout=>120
[...]

PR #8317, opened 2017-09-19
 #8293 fix concurrent access to PQ persisted bytesize throwing

original-brownbear:
See error reported in #8293, obviously looping over the tail pages will throw on concurrent modification of the underlying `ArrayList`.
Unfortunately, the stack trace of that error doesn't get forwarded to Ruby and also seems to prevent the mutex covering the PQ data poll on the Ruby side from unlocking (that's a different issue imo, no clue why this is happening). 
=> Either way, fixed by locking that operation in PQ and preventing concurrent modification exceptions.

* I'm aware that this needs a stronger fix, in the long run, it's not ok having to lock the pipeline for this, but for the current implementation, I see no other fix that doesn't add the same degree of locking.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -139,14 +139,19 @@ public class Queue implements Closeable {
     }
 
     public long getPersistedByteSize() {
-        final long size;
-        if (headPage == null) {
-            size = 0L;
-        } else {
-            size = headPage.getPageIO().getHead()
-                + tailPages.stream().mapToLong(p -> p.getPageIO().getHead()).sum();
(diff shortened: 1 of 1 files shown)


jordansissel:
Most tests passing. Only 1 DLQ test failing and is unrelated.

LGTM

suyograo:
@original-brownbear can we backport this to 5.6 as well?

original-brownbear:
@suyograo sure, let's see this might even merge clean :)

original-brownbear:
@suyograo done :)

colinsurprenant:
Just helped diagnose an issue on discuss which pointed to this *rather obvious* lack of locking  and found this fix - thanks @original-brownbear  this one was pretty nasty, good thing we also merged it in 5.6.

PR #8318, opened 2017-09-19
Add commercially supported plugins to be default

### LINKED ISSUE #8287: Make sure we ship Logstash with all supported plugins
The support matrix visible here: https://www.elastic.co/support/matrix#matrix_logstash_plugins

I believe our intent was to include all of the (commercially) supported plugins by default with Logstash. By some accident, things like `translate` filter do not ship with Logstash (in Logstash 5.5.1 and older)

Let's audit the support matrix and make sure all supported plugins are included in Logstash releases. 

It's a bug that we don't include all supported plugins, so the fix for this can be in a patch release.

---

I'm marking this a blocker because my experience of this is follows: 

* A user wants to use a supported-but-not-default plugin
* This user has a draconian network policy which prevents plugins from being easily installed
* Nearly everyone possible will consume great numbers of hours trying to find a solution that works within the constraints, and the users, our support team, our consulting team, our SA teams all spend many hours (per incident) on this scenario.

Because of the high number of hours consumed by users/etc as folks try to install supported plugins in "offline mode", I believe this to be a significant bug that we need to block the release for until it is resolved.

There is a PR open.

jordansissel:
Our intent is to ship all commercially supported[1] plugins with
Logstash. By some mistakes, we have omitted a few plugins, and this
commit resolves those omissions.

Fixes #8287

### DIFF
--- a/rakelib/plugins-metadata.json
+++ b/rakelib/plugins-metadata.json
@@ -747,5 +747,61 @@
 		"test-jar-dependencies": false,
 		"test-vendor-plugins": false,
 		"skip-list": true
+	},
+	"logstash-filter-aggregate": {
+		"default-plugins": true,
+		"core-specs": false,
+		"test-jar-dependencies": false,
+		"test-vendor-plugins": false,
+		"skip-list": false
+	},
+	"logstash-filter-anonymize": {
+		"default-plugins": true,
+		"core-specs": false,
+		"test-jar-dependencies": false,
+		"test-vendor-plugins": false,
+		"skip-list": false
+	},
+	"logstash-filter-de_dot": {
+		"default-plugins": true,
+		"core-specs": false,
+		"test-jar-dependencies": false,
+		"test-vendor-plugins": false,
+		"skip-list": false
+	},
+	"logstash-filter-elasticsearch": {
+		"default-plugins": true,
+		"core-specs": false,
+		"test-jar-dependencies": false,
+		"test-vendor-plugins": false,
+		"skip-list": false
+	},
+	"logstash-filter-jdbc_streaming": {
+		"default-plugins": true,
+		"core-specs": false,
+		"test-jar-dependencies": false,
+		"test-vendor-plugins": false,
+		"skip-list": false
+	},
+	"logstash-filter-truncate": {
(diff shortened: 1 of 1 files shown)


jordansissel:
This patch should apply cleanly to 6.0, 6.x, and master. I'll have to do a separate backport to 5.x.

jordansissel:
I'd like to consider this a candidate for blocking 6.0.0 -- specifically, I'd just like to have this solved for 6.0.0 GA as it will improve user experience and reduce support costs by avoiding the complex hand-holding that often is required today for installing supported plugins offline.

jordansissel:
Most tests passed: https://logstash-ci.elastic.co/job/elastic+logstash+pull-request/1312/

The one that failed is this one:

```
17:02:33 rspec ./specs/cli/prepare_offline_pack_spec.rb:41 # CLI > logstash-plugin prepare-offline-pack create a pack from a wildcard successfully create a pack
```

I will investigate.

andrewvc:
jenkins, please test this

jakelandis:
@jordansissel - that test failure was https://github.com/elastic/logstash/issues/8319 which should be fixed if you rebase.

suyograo:
This is really not a blocker. We can add it to 6.1.0

jakelandis:
This next build is going to fail due to https://github.com/elastic/logstash/issues/8142 (which is not resolved yet)

suyograo:
this PR LGTM. I'm merging it

jordansissel:
@suyograo I marked a blocker because of the number of hours by a large number of people wasted on this problem.

In https://github.com/elastic/logstash/issues/8287 I put my reasoning, which I will copy here:

I'm marking this a blocker because my experience of this is follows:

*    A user wants to use a supported-but-not-default plugin
*   This user has a draconian network policy which prevents plugins from being easily installed
*    Nearly everyone possible will consume great numbers of hours trying to find a solution that works within the constraints, and the users, our support team, our consulting team, our SA teams all spend many hours (per incident) on this scenario.

Because of the high number of hours consumed by users/etc as folks try to install supported plugins in "offline mode", I believe this to be a significant bug that we need to block the release for until it is resolved.

PR #8324, opened 2017-09-19
Added index-shared3.asciidoc

lcawl:
This PR splits the index files for the documentation such that content can be inserted after the "Breaking Changes" section.


dedemorton:
@lcawl Can you give me more detail about what you're planning to add to the book? 

Last week, @debadair  and I discussed an alternative to fragmenting the index over and over again. Eventually I'll have more X-Pack content interleaved throughout the book. She suggested having a referenced index file for a specific area (like monitoring) and having different versions of the file for LS and X-Pack. I like this idea because it won't requite moving around content, just adding a hook to the index.

FYI...I plan to move some of the content (monitoring, config mgmt, etc) that's currently in the logstash repo over to the X-Pack logstash repo (where it should have been all along). But I want to know what your plans are so we don't conflict.

PR #8329, opened 2017-09-20
5x: Use 5x base image for 5x builds

jakelandis:
* This allows for faster 5x builds since the base Docker image gem and jar cache is closer to the 5x dependencies.
* This prevents the Ruby 9k dependencies from sneaking into the build (had issues w/r/t jruby openssl)

Note - this should also fix the Jenkins 5.5 and 5.6 Jenkins builds due to a Jruby openssl issue (I don't fully understand) that occurs when Ruby 9k is installed first.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM container-registry-test.elastic.co/logstash-test/logstash-base:latest
+FROM container-registry-test.elastic.co/logstash-test/logstash-base-5x:latest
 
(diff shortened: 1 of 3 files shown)


original-brownbear:
@jakelandis

> this should also fix the Jenkins 5.5 and 5.6 Jenkins builds due to a Jruby openssl issue (I don't fully understand)

Imo we're good here if it works from a clean JRuby `1.7.x` => LGTM :)

jakelandis:
@original-brownbear - thanks!

PR #8332, opened 2017-09-20
MINOR: Remove unimplemented methods from batch interface

original-brownbear:
These are just noise I think. Given that we're moving this functionality to Java once #7950 is in, it's probably best to just kill this for now and see what will be the best approach to `cancel` in whatever a Java implementation will look like?

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -271,12 +271,6 @@ module LogStash; module Util
         @generated[event] = true
       end
 
-      def cancel(event)
-        # TODO: disabled for https://github.com/elastic/logstash/issues/6055 - will have to properly refactor
-        raise("cancel is unsupported")
-        # @cancelled[event] = true
-      end
-
       def to_a
         events = []
         each {|e| events << e}
@@ -304,12 +298,6 @@ module LogStash; module Util
 
       def filtered_size
         @originals.size + @generated.size
-      end
-
-      def cancelled_size
-        # TODO: disabled for https://github.com/elastic/logstash/issues/6055 = will have to properly refactor
-        raise("cancelled_size is unsupported ")
-        # @cancelled.size
       end
 
       def shutdown_signal_received?
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -169,12 +169,6 @@ module LogStash; module Util
         @originals.add(event)
       end
 
-      def cancel(event)
-        # TODO: disabled for https://github.com/elastic/logstash/issues/6055 - will have to properly refactor
-        raise("cancel is unsupported")
-        # @cancelled[event] = true
-      end
-
       def to_a
         events = []
         @originals.each {|e| events << e unless e.cancelled?}
@@ -192,12 +186,6 @@ module LogStash; module Util
       end
 
       alias_method(:size, :filtered_size)
-
-      def cancelled_size
-      # TODO: disabled for https://github.com/elastic/logstash/issues/6055 = will have to properly refactor
-      raise("cancelled_size is unsupported ")
-        # @cancelled.size
-      end
     end
 
     class WriteClient


jordansissel:
@colinsurprenant Can you comment as to the original goal with the `cancel` method here? If it was to support a future feature, I worry about removing code that if a project is still in-progress but hasn't started using it yet.

original-brownbear:
@jordansissel @colinsurprenant ping :) (given that we want to move this to Java, I don't think there is much need for keeping this kind of dead code around even if we wanted to (re-)introduce this type of API at some point though)

original-brownbear:
pong? :)

guyboertje:
@jordansissel @original-brownbear 
> Can you comment as to the original goal with the cancel method here? If it was to support a future feature, I worry about removing code that if a project is still in-progress but hasn't started using it yet.

I can.

The most of the code you see in the Facade wrappers around the PQ and the SynchronousQueue was as a result of it being a demo for interface abstractions around Queue reading, writing and batches for any kind of Queue. The idea of the cancel method was to show that its possible to track cancelled events inside the Batch and not yield them via an iterator - alongside the idea that the batch object itself was to be passed into the filter and output via the `multi_*`  methods and the filter would call `batch.cancel(event)`.

For me a Batch is an OO abstraction of a array (set) of Events much like the Event is an OO abstraction of a hash (set) of fields and values with all the internal goodness to manage those sets.

original-brownbear:
@guyboertje @jordansissel 

hmm ok, 

1. but moving the `cancel` to the `batch` is not something we're doing right now
2. seems like doing that would be a bad idea? It just adds another level of indirection to the whole process and requires the batch to hold more state => basically this forces one additional hashmap lookup on `cancel` and turns any `cancelled` check from a field lookup to a hashmap lookup (this has the further downside that the batch's datastructure tracking `cancelled` state will potentially hold references to `Event` instances making GC a lot more expensive on `Batch`) => why would we want to do that?

guyboertje:
The principle is worth considering even though badly implemented - with hash lookups etc..

The Ruby based Batch object was never intended to be around for long - once the API was settled and plugins changed to suit, it was my intention to move it to Java and use bitsets as the tracking mechanism.

original-brownbear:
@guyboertje I think even bitsets will be slower here. Then you have to look up a sequence number from the event and match it against a bitset, instead of simply looking up a boolean from the event?

guyboertje:
Again, the principle is still worth consideration even if the bitset implementation is not optimal.

It seems like you are saying, by focussing on an implementation or proposed implementation, that ideas for which no optimal implementation can be imagined at the time are not worth exploring.

original-brownbear:
@guyboertje 

> It seems like you are saying, by focussing on an implementation or proposed implementation, that ideas for which no optimal implementation can be imagined at the time are not worth exploring.

I think for core datapath logic this would be pretty close to my opinion, yes :)
Especially when it comes to APIs that we expose.

Internally I see more of a point for porting this stuff to bitsets (though this is really really hard to do in practice since we don't have any strong ordering guarantees when pulling from the Queue). Still for example, making `Event` just a pointer to a row in a columnar data structure that is held by `Batch` seems interesting.
What I simply would like to avoid is building an unnecessary level of indirection (note that with recent speedups and fixes, method calls are about 40-50% of our runtime CPU cost :( in `master`) like that into the API unless there is a clear benefit to it and we have some handle on the worst case cost of doing this.

colinsurprenant:
Sorry, it seemed I missed notifications of this issue - I will provide a bit more feedback later but obviously the way we have worked as far as I can remember is to work in iterations (progress over perfection + start simple; don’t be smart). I suggest we spend more time discussing the Batch concept to reach some kind of consensus moving forward and avoid misunderstanding or diverging visons. 

I believe we have a tension between premature abstraction and  premature optimization. I think "design for today & use abstractions with care" is important and the batch object is a "noisy"  abstraction but premature optimization can be just as harmful as premature abstraction so I'd be cautious of axing into code only for optimization reasons.

original-brownbear:
@guyboertje @colinsurprenant 

the issue here is a lot less metaphysical :)

We're simply moving this code to Java right now step by step. `Batch` is pretty much next in line to be moved over. If there is no present purpose to keeping this stuff around, this is simply a neat cleanup that makes the eventual Javafication PR cleaner to read.

Even if we eventually wanted to add some `.cancel(Event)` method to the Java batch, we probably won't be porting dead code over to Java in step 1? :)

original-brownbear:
@guyboertje @colinsurprenant ping?

colinsurprenant:
LGTM - sorry for the delay - it is #DEADBEEF so not needed. We'll see if wanted back as we progress with the new client interfaces.

original-brownbear:
@colinsurprenant thanks!

PR #8333, opened 2017-09-20
PERFORMANCE: More efficiently store offsets in AbstractByteBufferPageIO

original-brownbear:
Extended benchmarking tool to PQ and found this one:

* Storing `ints` into a large `ArrayList<Integer>` is extremely costly (hundreds of MB of old gen heap for the Apache benchmark with `-b 128 -w 4`)
   * This causes a lot of load GC wise slowing down every part of LS
   * It's so bad that it introduces major GC like in below screenshot from `master`

screen shot 2017-09-20 at 09 54 07

This situation is much improved by this simple change, moving to a manually grown `int[]` (though not fully fixed, even manually allocating these arrays is pretty costly, we need re-use here)

But for now, this is what it looks like after this patch:

screen shot 2017-09-20 at 10 53 00

-> average GC time down by `~50%`, old gen grow rate also down by `~50%`  :)
-> throughput up (~15% in base-line, seemingly (and theoretically it should be so) more in the Apache dataset, didn't have time for a long run here ...)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
@@ -42,7 +42,7 @@ public abstract class AbstractByteBufferPageIO implements PageIO {
     protected int head; // head is the write position and is an int per ByteBuffer class position
     protected byte version;
     private CRC32 checkSummer;
-    private final List<Integer> offsetMap; // has to be extendable
+    private final IntVector offsetMap;
 
     public AbstractByteBufferPageIO(int pageNum, int capacity) {
         this.minSeqNum = 0;
@@ -51,7 +51,7 @@ public abstract class AbstractByteBufferPageIO implements PageIO {
         this.head = 0;
         this.pageNum = pageNum;
         this.capacity = capacity;
-        this.offsetMap = new ArrayList<>();
+        this.offsetMap = new IntVector();
         this.checkSummer = new CRC32();
     }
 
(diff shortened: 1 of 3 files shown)


logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
```diff
@@ -169,12 +175,26 @@ private void readNextElement(long expectedSeqNum, boolean verifyChecksum) throws
...
 
         buffer.position(this.head);
     }
 
+    /**
+     * Store one offset in {@link AbstractByteBufferPageIO#offsetMap}, resizing
+     * {@link AbstractByteBufferPageIO#offsetMap} if necessary.
+     * @param elementOffset Offset to store
+     */
+    private void storeOffset(final int elementOffset) {
```
andrewvc: This method needs tests around its behavior for sure!

Yes, its simple, but we want it verified :)

I actually am thinking that this may be something we reuse. What are your thoughts on extracting this into a new `PrimitiveIntegerList` class?

andrewvc: Also, even if we don't reuse it, it makes the test case more clear

original-brownbear: @andrewvc I'll extract this, sounds easy enough :)

original-brownbear: @andrewvc though I'd call it `IntVector`, it's not really a `List`? :)

original-brownbear:
@andrewvc extracted :)

Issue #7318, opened 2017-06-05
Accessing the timestamp on the LogStash::Event instance create unnecessary new object


ph:
While debugging https://github.com/logstash-plugins/logstash-codec-rubydebug/pull/14, I have found that we create a new instance everytime that we call `Event#get("@timestamp")`, Since we use that instance in multiple places we should cache that value.

```
2.3.0 :002 > e = LogStash::Event.new({ "message" => "hello"})
 => #<LogStash::Event:0x3422b621>
2.3.0 :003 > e.get("@timestamp")
 => #<LogStash::Timestamp:0x6e9514b4>
2.3.0 :004 > e.get("@timestamp")
 => #<LogStash::Timestamp:0x1ff75c41>
2.3.0 :005 > e.get("@timestamp")
 => #<LogStash::Timestamp:0x549a4c80>
2.3.0 :006 > e.get("@timestamp")
 => #<LogStash::Timestamp:0x33def139>
2.3.0 :007 > e.get("@timestamp")
 => #<LogStash::Timestamp:0x166eb657>
2.3.0 :008 > e.get("@timestamp")
 => #<LogStash::Timestamp:0x10961656>
2.3.0 :009 > e.get("@timestamp")
 => #<LogStash::Timestamp:0x26d46bd3>
```

original-brownbear:
Just linking this one:

But ... this would obviously go away automatically if we drop the use of `BiValue` altogether as suggested here: https://github.com/elastic/logstash/issues/7712

original-brownbear:
fixed in #8334

PR #8334, opened 2017-09-20
 #7318 stop creating a new RubyTimestamp on every timestamp get

original-brownbear:
Fixing #7318 by simply storing `RubyTimestamp` instead of `Timstamp` in an `Event` by default. There is some value in this in addition to making the getter GC free, in that it also ensures that `setTimestamp` and `setField()` (which would use `Valuefier` to arrive at the same result), behave exactly the same.


original-brownbear:
@ph thanks!

PR #8338, opened 2017-09-20
MINOR: Remove needless lock on acked batch read

original-brownbear:
Leftover from #8186: this lock is useless since the write operation (and all other Queue operations for that matter) is locked on the Java side.

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -167,12 +167,7 @@ module LogStash; module Util
         end
 
         batch = new_batch
-        @mutex.lock
-        begin
-          batch.read_next
-        ensure
-          @mutex.unlock
-        end
+        batch.read_next
         start_metrics(batch)
         batch
       end


jordansissel:
I'm not following the "this lock is useless" -- if it is useless, why not remove it from the whole file? `@mutex` is still used in other places in this file.

original-brownbear:
@jordansissel all Queue operations are locked like this on the Java end:

```java
    public long getPersistedByteSize() {
        lock.lock();
```

The mutex is only useful for non-atomic operations on the Ruby end (in practice only manipulations on `@inflight_batches` at this point). All operations that only affect the Java side are already mutually exclusive due to that Java lock.

original-brownbear:
@jordansissel ping?

jordansissel:
Can you update the commit message to be more clear about why it is safe to remove the lock? (The underlying call has its own lock, so this extra mutex is not necessary, for example).

Otherwise LGTM.

original-brownbear:
@jordansissel sure thing added the details to the message body :)

Issue #8341, opened 2017-09-20
Test failure: timing issue in timestamp_spec.rb


jakelandis:
```
16:28:26 Failures:
16:28:26 
16:28:26   1) LogStash::Timestamp constructors should work
16:28:26      Failure/Error: expect(t.time.to_i).to be_within(1).of Time.now.to_i
16:28:26        expected 1505924544 to be within 1 of 1505924546
16:28:26      # /var/lib/jenkins/workspace/elastic+logstash+6.x+multijob-unix-compatibility/os/ubuntu/logstash-core/spec/logstash/timestamp_spec.rb:16:in `block in (root)'
16:28:26      # /var/lib/jenkins/workspace/elastic+logstash+6.x+multijob-unix-compatibility/os/ubuntu/vendor/bundle/jruby/2.3.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in /var/lib/jenkins/workspace/elastic+logstash+6.x+multijob-unix-compatibility/os/ubuntu/vendor/bundle/jruby/2.3.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb'
16:28:26      # /var/lib/jenkins/workspace/elastic+logstash+6.x+multijob-unix-compatibility/os/ubuntu/lib/bootstrap/rspec.rb:13:in `<main>'
16:28:26 
16:28:26 Finished in 7 minutes 42 seconds (files took 18.7 seconds to load)
```

Caused by the following test in timestamp_spec.rb is dependent on these two command being executed within 1ms of each other: 

```
      t = LogStash::Timestamp.now
      expect(t.time.to_i).to be_within(1).of Time.now.to_i
```

original-brownbear:
fixed in #8342

PR #8342, opened 2017-09-20
 #8341 raise tolerance in timestamp_spec.rb tests

original-brownbear:
Just like we did in https://github.com/elastic/logstash/commit/f64762fd05926fb07e30d8d4028a5ae21fa6124e just a matter of rounding down to `int` and having some jitter in `System#currentTimeMillis` (which Joda seems to be using under the hood).


jakelandis:
@original-brownbear 
LGTM - thanks for the quick fix!

jakelandis:
Backported to 5.x:

Branch | Commits
-------|--------
5.5 | 522122eed45695ba91c28bb9811cdd041b855143
5.6 | 6ef8427eb2bf3e348d100a0dc46300812db2b46b

PR #8355, opened 2017-09-21
PERFORMANCE: Faster Long Vector Handling

original-brownbear:
Follow-up for #8333 for the long case, same as for int except:

* `LongVector`  needs (or rather can make good use of) a sizing argument since we know its size in most (unfortunately not with 100% certainty in all cases, otherwise we could just use `long[]`) when instantiating it

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
@@ -4,16 +4,17 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.logstash.ackedqueue.io.LongVector;
 
 public class Batch implements Closeable {
 
     private final List<Queueable> elements;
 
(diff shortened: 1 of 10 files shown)


original-brownbear:
Bam this makes the GC overhead for PQ almost the same as that of the in-memory queue.

Now we have:

screen shot 2017-09-21 at 20 05 47

compared to before:

screen shot 2017-09-21 at 20 05 47

=> 7ms avg pause goes to 4ms and old gen is pretty close to flat.

guyboertje:
Jenkins please test this

guyboertje:
Previous Jenkins link 404.

guyboertje:
So this is really a datatype change, no new functionality, yes?

guyboertje:
I think we can remove Andrew as reviewer.

original-brownbear:
@guyboertje exactly just a datatype change nothing else.

I'll remove Andrew, just added him since he reviewed #8333 which was pretty much the same issue.

original-brownbear:
@guyboertje thanks!

Issue #8110, opened 2017-08-30
`org.logstash.ObjectMappers.RubyStringSerializer` Should Serialize a RubyString without creating a Java `String`


original-brownbear:
The current serializer for `RubyString` is ok but could be a lot faster.
We have:

```java
    /**
     * Serializer for {@link RubyString} since Jackson can't handle that type natively, so we
     * simply serialize it as if it were a {@link String}.
     */
    private static final class RubyStringSerializer
        extends NonTypedScalarSerializerBase<RubyString> {

        RubyStringSerializer() {
            super(RubyString.class, true);
        }

        @Override
        public void serialize(final RubyString value, final JsonGenerator generator,
            final SerializerProvider provider)
            throws IOException {
            generator.writeString(value.asJavaString());
        }
    }
```

which means we create a new `value.asJavaString()` for each `RubyString` we serialize.
This is unnecessary, we already have all the `byte[]` in the `ByteList` backing the `RubyString` and should simply write them to the output Stream.
This isn't so pressing since the `String` is actually never created and JDK8 escapes it to a `char[]` but still, allocating that is just wasted memory and CPU cycles.

original-brownbear:
Once https://github.com/elastic/logstash/pull/8205 is merged, we can do this in a very fast way by writing a typed `RubyString` to CBOR as `byte[]` without any allocations necessary whatsoever :)

original-brownbear:
so fix to this is already in #7950 :)

original-brownbear:
Fix was merged with the Java execution (also works with Ruby exec since it's just a change to the String serializer).

PR #8357, opened 2017-09-21
PERFORMANCE: LIR Pipeline Execution

original-brownbear:
Continuation of #7950 : 

### Performance

- [x] GC overhead is lower in every case compared to the old implementation
- [x] Conditionals evaluate a lot faster
- [x] The case of no filters or conditionals is about 5% slower than before (I don't know why yet, this needs to be investigated/fixed before merging, there must be some redundant action in the execution, that only shows in this case and is outweighed by general improvements for all other cases) ... Fixed, this was a measurement error, we faster in all cases now I think :)

#### Possible Improvements

- Janino (obviously, but also kind of easily doable from this implementation since I kept the translation from `PipelineIR` very explicit, not sure how much we will gain though since thanks to the way `FieldReference` lookups now work, there is very little that doesn't get inlined)
   - Another option here would be using `invokedynamic` together with the `org.ow2.asm` (less work and potentially faster, but much harder conceptually). This would also allow major speedups in the Ruby <-> Java interaction here
- Less complicated: Smarter chaining of conditionals (currently none of them are getting optimized except for constants, there is a lot of room here by reordering etc.)

#### Benchmarks

Baseline:
```
➜  logstash git:(lir-if-else) ✗ bin/benchmark.sh --local-path=${PWD} -repeat-data=10
Logstash Benchmark
Running Test Case: baseline (x10)
Start Time: Sun 8 13 18:50:23 2017 CEST
Statistical Summary:

Elapsed Time: 51s
Num Events: 10000000
Throughput Min: 0.00
Throughput Max: 315750.00
Throughput Mean: 285714.29
Throughput StdDev: 71738.67
Throughput Variance: 5146436229.27
Mean CPU Usage: 28.34%
➜  logstash git:(lir-if-else) git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'elastic/master'.
➜  logstash git:(master) rake compile:all &> /dev/null
➜  logstash git:(master) bin/benchmark.sh --local-path=${PWD} -repeat-data=10
Logstash Benchmark
Running Test Case: baseline (x10)
Start Time: Sun 8 13 18:51:59 2017 CEST
Statistical Summary:

Elapsed Time: 53s
Num Events: 9798106
Throughput Min: 20033.00
Throughput Max: 301875.00
Throughput Mean: 279945.89
Throughput StdDev: 55609.24
Throughput Variance: 3092387983.10
Mean CPU Usage: 36.91%
```

=> Much lighter on the GC now!

Also kind of cool is the Apache benchmark:

Master:
```sh
➜  logstash git:(master) time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg | pv | wc -c
13.2MiB 0:07:21 [30.5KiB/s] [                                                                                                 <=>                                                                                                          ]
 13801522
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.16s user 2.72s system 0% cpu 7:20.90 total
bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg  1833.99s user 30.28s system 422% cpu 7:21.71 total
pv  3.18s user 20.04s system 5% cpu 7:21.71 total
wc -c  0.41s user 3.21s system 0% cpu 7:21.71 total
```

This:

```sh
➜  logstash git:(lir-if-else) time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg | pv | wc -c
13.2MiB 0:05:14 [42.9KiB/s] [    <=>                                                                                                                                                                                                       ]
 13801710
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.14s user 2.80s system 0% cpu 5:13.88 total
bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg  1315.97s user 31.09s system 428% cpu 5:14.42 total
pv  3.07s user 19.74s system 7% cpu 5:14.42 total
wc -c  0.37s user 2.89s system 1% cpu 5:14.42 total
```

Much better throughput by batching for each filter (hence making good use of the CPU cache) function :)

Note:

* There is a million performance improvements we could add here, but pretty much all of them also involve adding more (or more complicated) code => I tried to be as short as possible here for round 1
* "filtered" and "output" metrics are always the same size in `master` as well (just in case you wonder why both increment calls now happen back to back)
* `filter(event, &block)` on the pipeline should really go away as it way complicates the Java logic, requiring interpreting all leafs as outputs when there are no outputs in the config, and also forcing us to have the debug terminal dataset
* Obviously passing the `flush` "state" up/down the `Dataset` topology is kind of dirty by just using params, but I feld this was the easiest approach to clean up when/if we drop flush + didn't add a performance it to the general case
* This contains the fix to #8110 since I needed typed-deserialization for `RubyString` here to get good performance on `String` comparisons (otherwise I have to convert `String` to `RubyString` after deserializing `String` which is a loss in every scenario ... either way this is a big speedup for serializing `RubyString`)
* Also contains https://github.com/elastic/logstash/pull/8205 as a dependency (still waiting for a review there ;))


original-brownbear:
@ph @andrewvc let's continue here? The other one became unusable laggy in browsers imo, I'd close it if you're cool with it? (I'll add a proper description here soon then :))

andrewvc:
@original-brownbear yep, down to continue here :)

ph:
@andrewvc @original-brownbear ++

original-brownbear:
@andrewvc @ph Updated performance numbers :)

### Master

```sh
➜  logstash git:(master) time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg | pv | wc -c
13.2MiB 0:07:01 [32.0KiB/s] [                                                   <=>                                                                                                                                                                                                                                                                             ]
 13802035
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.17s user 2.91s system 0% cpu 7:00.52 total
bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg  1755.94s user 25.90s system 422% cpu 7:01.41 total
pv  3.24s user 19.75s system 5% cpu 7:01.40 total
wc -c  0.42s user 3.34s system 0% cpu 7:01.40 total
```

### This

```sh
➜  logstash git:(lir-if-else-squashed) time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg | pv | wc -c
13.2MiB 0:04:37 [48.5KiB/s] [                                                                                                                                                                                                                         <=>                                                                                                       ]
 13801638
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.12s user 2.25s system 0% cpu 4:36.88 total
bin/logstash -b 128 -w 4 -f ~/tmp/logstash.cfg  1164.86s user 25.72s system 428% cpu 4:37.68 total
pv  3.14s user 19.99s system 8% cpu 4:37.68 total
wc -c  0.44s user 3.58s system 1% cpu 4:37.68 total
```

Gains in both versions via last months fixes reducing function call count and indirection in a few places :)

original-brownbear:
@andrewvc is there anything I can do here to speed this along? If the main concern here is having Java only tests, I could try and code up the infrastructure for that next week? :)
(I'm just looking for a way of not letting this sit here for another 3 weeks, I already had to rebase it a few times again + one PR merged yesterday did hurt the performance of this already and it was hard to argue on the basis of an unmerged PR)

andrewvc:
@original-brownbear I will definitely hand off the review if I can't get to it this week.

I think I'm personally not far from LGTM, so I'd rather not do that, which would delay it even further, unless necessary. I promise an answer by tomorrow EOD

original-brownbear:
@andrewvc alright thanks, I don't think any Java tests are necessary at this point. So yea let's not do them, was just trying a new angle at getting this moving :D

andrewvc:
After discussing with @ph , and given that this feature is targeted for 6.1.0, not 6.0, what I'd like to propose is to make this an opt-in feature behind a `--experimental-java-execution` flag or similar.

This is amazing work @original-brownbear , but it's too risky to make this the default in a minor without more real-world feedback.

We have me and @ph doing the reviews here, but I think we only really need two reviewers on the new execution code. This means, of course, resurrecting some of the deleted files, but hopefully that isn't too bad? @ph alone is good enough to review any feature flag code.

Once again, great work here, and looking forward to seeing this land in 6.1.0.

original-brownbear:
@andrewvc 

>  but hopefully that isn't too bad?

This is pretty bad, but what else can we do? :) It's just that:

1. (not so important) We didn't follow this approach from day one and now some internal pipeline APIs were refactored here, making this not really a drop in solution at all now. This will require a bit of effort to get things going side-by-side. 
2. (somewhat important) This requires us to run all tests twice (Ruby specs + integration tests). This def. adds some overhead to the test infrastructure. (also note that some functionality was fixed in the Java exec only, like `XOR` and `NAND` which will require us to ignore some tests for the Ruby exec)
3. (pretty bad) At least from my perspective the plan was to leverage this work to be able to move more logic to Java (especially `Batch` and `Queue` interaction). This will be really tricky to pull off if we have the Ruby exec to support in parallel. In effect, we'd have two diverging codebases to maintain for quite a while.

I can code up a version that sits behind a feature flag, but you have been warned, the overhead from this is pretty big effort and code size wise. But unless I hear anything back on those points, I'll just go ahead and code this up though :) Anything to get some movement here :)

andrewvc:
@original-brownbear this is definitely up for discussion, but we're in a tricky place. I'll add that I'm not happy with where we are here either, but we're in a tricky spot. We have two choices:

1. Release a major refactor in a minor with major potential for a severe bug
2. Release a feature-flagged version that lets power users test things out

Every large PR we've merged into core has had some degree of nasty bug that wasn't caught until after release. Notable examples: LIR (which didn't impact users because it was only used for a non-existent monitoring UI), Java Event, the persistent queue. 

I want to be clear that this has nothing to do with the quality of your work, which is excellent. This is more a problem of unknown unknowns. Which tests did we not think to write?

The main question here is, where do we lay the burden of dealing with this uncertainty? On our users, or on ourselves as developers? Doing this upgrade without making it an opt-in FF means users don't really get to opt-in to the inherent risk of this major refactor.

I agree that we made a mistake not adopting this strategy earlier. I accept a lot of the responsibility for that. However, it's never too late to make the right decision. I think we were originally planning on being able to target this for 6.0. If we had this same PR open for 6.0.0-beta2 I would say that we are fine to merge as is for two reasons: 1. Users would be able to try it out during the beta process and report problems, 2. The expectations around breakage for a new major are different than they are for a new minor

I'm OK to transition fully over to this code as early as 6.2 if we feel confident that this refactor is solid. I'm also OK with merging this PR without feature flags into master. However, maintaining a non-FF version in master and a FF version in 6.x is probably more work.

WRT some core refactors being delayed here, we have a mountain of other really worthy work, like refactoring the Elasticsearch output. We can cycle back to those things when its more practical.

If you have other ideas for how we can approach this problem, let's discuss them here.

original-brownbear:
@andrewvc I hear you I really do :) My worry here really is this:

I'm 95% done creating a side-by-side version here https://github.com/elastic/logstash/compare/master...original-brownbear:java-exec-flagged?expand=1 :) You know I'm quick :P but obviously, we have a ton of duplication in that one.

If we can live with that and don't spiral into some crazy drying up of such a solution here, I see a realistic path to coming to an end on this. 
(I know this is a bit of a polemic argument :), but I'm simply trying to understand how much effort this will take and how realistic it is to get a review on this ... note the many PRs I still have open)

I'm really just worried about our ability to review this within a realistic time frame. I can code up whatever we decide on no problem, but if a solution is shot down on the grounds of duplication and will require another round of dependant PRs to `master` and `6.x` (just to prevent duplication from creeping in, I def see a path for that) with another month+ of waiting for reviews, this seems like it's going out of control of control.


=> just a question of our review throughput + the expectations about the quality of the feature flag solution :)

original-brownbear:
@andrewvc all I really need here is some actionable goal here :) We can obviously revisit this in 3 weeks if we can't find a clear approach this week.

I guess the alternative would simply be accepting the reality of our test infrastructure and review throughput and just going for `7.x/master` here (If we don't want to leverage this for more Javafication in the near future, it seems going for 7.x wouldn't be that bad would it?).

andrewvc:
@original-brownbear That is awesome!I understand that we'll have duplicated treetop stuff + a few other things. That is 100% OK. There's no other way to do this, and that is an acceptable price to pay :)

I don't see any reason why we need very long to get this in once we have the feature flag at all. Really, I'd be glad to move this into 6.x ASAP given the feature flag .

As an example, we merged LIR in, in a sort of read-only mode a long time before it was turned on for any actual use case. This gave us time to burn it in without exposing users to harm. I'm 100% fine with that approach here.

Thanks for the awesome work :)

I'm currently doing my final review pass as we speak, I'll have finished that by EOD.

original-brownbear:
@andrewvc alright thanks! I'll create the FF flag version from your review comments next week then (maybe tomorrow, but can't promise I'll find time :)).

logstash-core/src/main/java/org/logstash/config/ir/compiler/Dataset.java
```diff
@@ -0,0 +1,468 @@
...
+
+        private final Collection<JrubyEventExtLibrary.RubyEvent> complement;
+
+        private final Dataset opposite;
+
+        private boolean done;
+
+        public SplitDataset(final Collection<Dataset> parents, final EventCondition func) {
+            this.parents = parents;
+            this.func = func;
```
andrewvc: can we rename this to `eventCondition`? I think that's more clear. This isn't a generic function.

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -0,0 +1,754 @@
...
+        }
+
+        private static EventCondition in(final EventValueExpression left,
+            final EventValueExpression right) {
+            return new EventCondition.Compiler.FieldInField(
+                PathCache.cache(left.getFieldName()), PathCache.cache(right.getFieldName())
+            );
+        }
+
+        private static EventCondition eq(final EventValueExpression evale,
```
andrewvc: `evale` isn't camel cased. `eValE` is more clear as well.

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -0,0 +1,754 @@
...
+            final ValueExpression vale) {
```
andrewvc: `vale` could be camel cased as well.

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -0,0 +1,754 @@
...
+            } else if (value instanceof Long || value instanceof Integer ||
+                value instanceof Short) {
+                return new FieldGreaterThanNumber(
+                    field, RubyUtil.RUBY.newFixnum(((Number) value).longValue())
+                );
+            }
+            throw new EventCondition.Compiler.UnexpectedTypeException(value);
+        }
+
+        private static EventCondition truthy(final EventValueExpression evale) {
```
andrewvc: `evale` should be camel cased `eValE`

logstash-core/src/main/java/org/logstash/ObjectMappers.java
```diff
@@ -42,6 +43,7 @@
 
     private static final SimpleModule CBOR_DESERIALIZERS =
         new SimpleModule("CborRubyDeserializers")
+            .addDeserializer(RubyString.class, new RubyStringDeserializer())
```
andrewvc: Why are there changes to serialization in this PR? Was this to fix test problems?

andrewvc:
I'd also like to add, a tentative strategy for moving this to be the default execution strategy:

1. Release in 6.1.0 feature-flagged, default off
2. If stable, release in 6.2.0 feature-flagged, default on
3. If stable, release 6.3.0 as sole execution strategy. Feature flag is now a noop

NOTE: The version numbers above may change, of course, depending on how stable this proves to be in the field.

We should make sure that the feature flag is loudly declared 'experimental', something like: `--experimental-java-execution=on`, so that users will be aware of the fact that it may be removed in a future release.

logstash-core/src/main/java/org/logstash/ObjectMappers.java
original-brownbear: @andrewvc nah, this is kind of an annoying corner case, long story short:

`RubyFixnum` and other Ruby type's `compareTo` method works fine with Java numbers (since those just use `toLong()` and such below the surface), but for `String` and `RubyString` this isn't the case (due to encoding issues) => without adding a `Deserializer` that gives us straight `RubyString` we'd have to add a special code path for comparisons to `String/RubyString` which would've had quite the negative performance impact => `this` :)

original-brownbear:
In the sense of intellectual honesty: 

https://github.com/elastic/logstash/pull/8428 significantly diminishes the value of this PR unfortunately.
Fixing up the batching in the Ruby exec in a trivial manner makes comparable (Java wins for `if` statements, Ruby wins for actual filter call performance) to this approach performance wise.
The fact that Ruby beats us here is basically just due to the function call overhead from Java to Ruby that shows big time with the fix in #8428.
Obviously fixable with a stronger Java implementation in the next iteration though I think :)

logstash-core/lib/logstash/runner.rb
```diff
@@ -93,6 +93,11 @@ class LogStash::Runner < Clamp::StrictCommand
     :attribute_name => "pipeline.workers",
     :default => LogStash::SETTINGS.get_default("pipeline.workers")
 
+  option ["--experimental-java-execution"], "COUNT",
```
ph: replace "COUNT" with `:flag` to correctly set the boolean flag.

```
ERROR: option '--experimental-java-execution'
```

I think we could drop the "experimental"

ph:
@original-brownbear lets make sure we squash the commits :)

logstash-core/lib/logstash/runner.rb
original-brownbear: @andrewvc wanted the `experimental` explicitly :) see https://github.com/elastic/logstash/pull/8357#issuecomment-332918877

But yea, let me fix that my bad. It works if you use `--experimental-java-execution=true` ... should've been a pointer for me to remove that i needed the `=true`:D

original-brownbear:
@ph flag fixed, and sure I'll squash before I merge :)

original-brownbear:
merging :D

PR #8371, opened 2017-09-22
PERORMANCE: Remove all BiValue usage by removing NullBiValue

original-brownbear:
Last step for #7712 removing `NullBiValue`.

Same as all the others before it (like #8189 and so on) except:

* `null` was serialized to `""` for CBOR and JSON, I fixed this to serialize as actual `null` for CBOR (because we have typed serialization there to have no difference in behavior between PQ and in memory Queue, for JSON I kept the old behavior)
* Extracted the `Exception` that `BiValues` was throwing on behalf of all three converters to a new Exception


original-brownbear:
@guyboertje this one's for you I think :)

rdsubhas:
@original-brownbear thank you so much, we have been following your entire thread on removing BiValues due to random frequent unrecoverable data loss with persistent queue due to arrays from somewhere making their way into the serialized data files. Its been a mystery and we're losing logs every now and then, on a weekly basis, with no option available to reproduce other than lose data [1] https://discuss.elastic.co/t/logstash-5-4-repeated-deserialization-errors-and-losing-data-with-persistent-queues/91503/2 [2] https://discuss.elastic.co/t/logstash-5-6-0-deserialization-errors-when-persistent-queues-enabled/100952 [3] https://discuss.elastic.co/t/logstash-crash-with-persistent-queue-and-kafka-input/90479/3

I'm really sorry for plugging into this issue. But just in case, if adding Array deserialization is a trivial matter, it would be really great, or maybe not and we're mistaken - please feel free to ignore this comment if its irrelevant to what you're doing. Thanks for the great work!

original-brownbear:
@rdsubhas I think you're likely not running into actual arrays but rather some type that contains an array like I reproduce here #8131 (and in 2 other cases around timestamps, that I fixed but can't find right now for some reason). I think unless you are actually explicitly using `byte[]` you should be good once those are released.

If you have the ability to test `master`, I'd be happy to get some feedback though. Maybe I still missed a spot and if we find it I can fix it :)

rdsubhas:
@original-brownbear thanks! :)

rdsubhas:
@original-brownbear I have raised a discussion https://discuss.elastic.co/t/logstash-persistent-queues-have-serious-data-loss-bugs-not-suitable-for-production/101566 to see what we can do until v6.1 is released, would love to hear your inputs there.

original-brownbear:
@guyboertje thanks!

colinsurprenant:
For the record #8386 was submitted to revert the JSON serialization of `Nil` to `""` back to `null`

PR #8373, opened 2017-09-22
Bump JrJackson/Jackson versions, source all version refs from versions.yml

guyboertje:


guyboertje:
yay - all 💚

bin/logstash
```diff
@@ -39,13 +39,19 @@ else
...
-. "$(cd `dirname $SOURCEPATH`/..; pwd)/bin/logstash.lib.sh"
+. "$(cd `dirname ${SOURCEPATH}`/..; pwd)/bin/logstash.lib.sh"
 setup
 
-if [ "$1" = "-V" ] || [ "$1" = "--version" ];
-then
-  LOGSTASH_VERSION_FILE="${LOGSTASH_HOME}/logstash-core/lib/logstash/version.rb"
-  LOGSTASH_VERSION="$(sed -ne 's/^LOGSTASH_VERSION = "\([^*]*\)"$/\1/p' $LOGSTASH_VERSION_FILE)"
+if [ "$1" = "-V" ] || [ "$1" = "--version" ]; then
+  LOGSTASH_VERSION_FILE1="${LOGSTASH_HOME}/logstash-core/versions-gem-copy.yml"
```
jakelandis: Can you add a comment here to why we check 2 files?

Also, nitpick...but can you name the file  copied file versions.yml , you already have a comment in copy and I think it makes it more obvious they are the same file, if they are named he same.

logstash-core/logstash-core.gemspec
```diff
@@ -1,6 +1,31 @@
...
+  # as the require 'logstash-core/version' loads the yaml file from within the gem root.
+  #
+  # we ignore the copy in git and we overwrite an existing file
+  # each time we build the logstash-core gem
+  original_lines = IO.readlines(project_versions_yaml_path)
+  gem_versions_yaml_path = File.expand_path("./versions-gem-copy.yml", File.dirname(__FILE__))
+  version_difference = true
+  if File.exists?(gem_versions_yaml_path)
+    copy_lines = IO.readlines(project_versions_yaml_path)
+    version_difference = original_lines[2] != copy_lines[2]
```
jakelandis: Can you add a comment to what this is specifically checking?

jakelandis: Actually can we just get rid of this, and always ensure a delete->copy -> comment ? E.g. Remove the version_diff conditional?

guyboertje: Will do (it was so but I changed it because the gemspec gets evaluated soooo many times during various rake or bundler activities).

jakelandis: in that case would md5 hash of the file be more appropriate ?

bin/logstash
guyboertje: Re: comment - sure.

Re: nit-pick, well if you look at the code in `logstash-core/lib/logstash/version.rb` and `logstash-core/lib/logstash-core/version.rb`, line 8 - I wanted it to crystal clear that the file is *not* the root `versions.yml` because trying to work out which folder `../../` is pointing at is hard (for me at least). Oh and people don't read comments!

logstash-core/logstash-core.gemspec
guyboertje: No need, I said I would revert to overwrite (its about the same computation cost to hashing anyway)
Actually, there is a flaw in the current check because we may only be changing the jrjackson and jackson values but then the file would not be copied (say, during development == POLS violation)
It was not an on-purpose thing to check if the reviewer brought their A game, I'm not Machiavelli. 😆

bin/logstash
jakelandis: >Oh and people don't read comments!

Can't argue with that :)

jakelandis:
LGTM

jakelandis:
I will work on back porting this back to 5.5

jakelandis:

Branch | Commits
-------|--------
6.x | 5a9e670c4b366f4155671cfaaf614dec29435089

jakelandis:

Branch | Commits
-------|--------
6.0 | 63f8f86bde713c5686132571b98dc4bc0ea62af3

jakelandis:

Branch | Commits
-------|--------
5.5 | 01872095d9c410f1d95a1482dd134f5f5fbd58cf

jakelandis:

Branch | Commits
-------|--------
5.6 | 4e0ff4e9e9dbbbdeac5f309df8266f27464d4331

Issue #7767, opened 2017-07-21
Windows Builds


jakelandis:
Windows builds were recently introduced to the Logstash Jenkins builds, and have not fully worked. 

For example: https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-windows-compatibility/

Below is the list to get Windows builds fully functional:
- [x] Fix the Java core tests https://github.com/elastic/logstash/pull/7716
- [x] Fix Ingest Conversion tool tests https://github.com/elastic/logstash/issues/7642
- [x] Fix Jenkins ```$HOME``` to ensure proper escaping. https://github.com/elastic/infra/pull/2483
- [x] Allow long file names on Jenkins. https://github.com/elastic/infra/issues/2487
(https://github.com/elastic/logstash/issues/7634#issuecomment-316843373)
- [x] Fix Null Pointer from ```dsl.instance_eval``` issue. See [this comment](https://github.com/elastic/logstash/issues/7634#issuecomment-316895648), and https://github.com/elastic/logstash/issues/7768
- [x] [6.0, master] Fix Ruby test failures #8116 #8074
- [x] [5.6] Fix Ruby test failures #8125 
- [x] [5.5] Fix build & tests #8354 
- [x] fix `logstash.bat`, `setup.bat` and `rspec.bat` #8126  #8178 #8359
- [x] Evaluate if builds need to be run on Windows 2012. See [this comment](https://github.com/elastic/infra/issues/2487#issuecomment-317078580) and the conversation below.
- [x] Fix false positive ```success``` from ci\unit_tests.bat See [this comment](https://github.com/elastic/logstash/issues/7634#issuecomment-316843373) and [this comment](https://github.com/elastic/logstash/issues/7767#issuecomment-333928201)
- [x] Support builds on Windows 2016 w/ longfile name fix. 
- [x] Support builds on Windows 2012 (no long file name fix available) #8398 
- [x] Discuss updating the [support matrix](https://www.elastic.co/support/matrix) for Windows to include a footnote about the long file names and potential workarounds. (see below - logged follow up https://github.com/elastic/logstash/issues/8424)

-----------
My notes to run locally:

 ensure that JRuby is installed at C:
 For example: C:\jruby-9.1.12.0
 Need to run from cmd, not powershell, but can launch cmd from powershell for a better cmd.
 From powershell:
 [clone the repo](https://www.develves.net/blogs/asd/articles/using-git-with-powershell-on-windows-10)

    cmd
[...]

PR #8398, opened 2017-09-26
[WIP] dynamically assign drive letter to WORKSPACE env var

colinsurprenant:
Try to assign a drive letter to the `WORKSPACE` env var to shorten the current path to avoid hitting the 260 chars limits in Windows.

Strategy:
- first loop through the list of `subst` drives to see if `WORKSPACE` was already assigned to a drive and reuse it
- otherwise try to `subst` the first free drive letter to `WORKSPACE`

The idea is to avoid just assigning new drive letters blindly since this is a long-running server. The number of used drive letters will top at the number of unique `WORKSPACE` paths which shoud avoid depletion. 

Relates to #7767

### DIFF
--- a/ci/unit_tests.bat
+++ b/ci/unit_tests.bat
@@ -1,35 +1,86 @@
 @echo off
+setlocal enabledelayedexpansion
 
-setlocal
-
-REM Since we are using the system jruby, we need to make sure our jvm process
-REM uses at least 1g of memory, If we don't do this we can get OOM issues when
-REM installing gems. See https://github.com/elastic/logstash/issues/5179
-
-SET JRUBY_OPTS="-J-Xmx1g"
-SET SELECTEDTESTSUITE=%1
-SET /p JRUBYVERSION=<.ruby-version
-
-IF NOT EXIST %JRUBYSRCDIR% (
-  echo "Variable JRUBYSRCDIR must be declared with a valid directory. Aborting.."
+if "%WORKSPACE%" == "" (
+  echo Error: environment variable WORKSPACE must be defined. Aborting..
   exit /B 1
 )
 
-SET JRUBYPATH=%JRUBYSRCDIR%\%JRUBYVERSION%
+:: see if %WORKSPACE% is alread mapped to a drive
+for /f "tokens=1* delims==> " %%G IN ('subst') do (
+  set sdrive=%%G
+  :: removing extra space
+  set sdrive=!sdrive:~0,2!
+  set spath=%%H
 
-IF NOT EXIST %JRUBYPATH% (
-  echo "Could not find JRuby in %JRUBYPATH%. Aborting.."
+  if /I "!spath!" == "%WORKSPACE%" (
+    set use_drive=!sdrive!
+    goto :found_drive
+  )
+)
+
+:: no existing mapping
+:: try to assign "%WORKSPACE%" to the first drive letter which works
+for %%i in (A,B,C,D,E,F,G,H,I,J,K,L,M,N,O,P,Q,R,S,T,U,V,W,X,Y,Z) do (
+    set "drive=%%i:"
+    subst !drive! "%WORKSPACE%" >nul
+    if not errorlevel 1 (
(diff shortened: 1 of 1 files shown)


jordansissel:
I am +1 on this proposal (using `subst` to map a drive letter to a path) but I haven't tested.

ci/unit_tests.bat
```diff
@@ -1,6 +1,47 @@
 @echo off
+setlocal enabledelayedexpansion
 
-setlocal
+if not exist %WORKSPACE% (
```
jakelandis: Is there a current working in batch (e.g. pwd)  that could be used instead of a the Jenkins environment variable ? (It is a bit kinder to Windows developers )

ci/unit_tests.bat
```diff
@@ -1,6 +1,47 @@
...
+  echo Error: env var WORKSPACE must be defined
```
jakelandis: hmm...i ran into this error while test, but didn't get the echo to print in the cmd console.

colinsurprenant: oh ok. do you mean testing in Jenkins? If so maybe `WORKSPACE` is not defined then? but it should be per the Jenkins build env var list page!?

ci/unit_tests.bat
colinsurprenant: yeah, once we know this work we can certainly improve it to not make it Jenkins specific somehow... or have some sensible fallback defaults.

ci/unit_tests.bat
jakelandis: Sorry, i meant running locally I did not define `%WORKSPACE%` and had to review the code to understand the issue since the echo didn't print.  

I haven't tested in Jenkins yet.

colinsurprenant:
I added WIP in the title because there will certainly be some back & forth with this since I did not test it under Jenkins.

ci/unit_tests.bat
jakelandis: This seems to fix it.  Note - it works from cmd, but seems to not work from PS, Jenkins is not PS, and I _think_ .bat files are default executed by cmd.
```
if "%WORKSPACE%"=="" ( 
  echo Error: env var WORKSPACE must be defined
  exit /B 1
)
```

ci/unit_tests.bat
```diff
@@ -1,6 +1,47 @@
...
+  exit /B 1
+)
+
+:: see if %WORKSPACE% is alread mapped to a drive
```
jakelandis: I commented out 
```
 rem %RAKEPATH% test:install-core
 rem %RAKEPATH% test:core
```
and ran this a few times (in cmd)  to test this, and the drives kept incrementing every time i ran it. (it did succesfully skip c:\ however)   

I am not familiar with the voodoo here, but trying to echo `%%G` or `%%~sfH` seemed to echo the literals not what _I assume_ they are expanding to ? 
image

colinsurprenant: uh .. this is weird. I tested the "reuse" bits in isolation and it worked in my 2012 box ... but I did not test when included in the user_tests.bat. will do, thanks!

ci/unit_tests.bat
colinsurprenant: fixed this.

ci/unit_tests.bat
jakelandis: fixed with latest set of commits
image

jakelandis:
I will readily admit that I don't fully understand the line by line details of the batch scripting.  But I don't think that is needed here it since it either works or it doesn't. 

I worked with @colinsurprenant to get this fully tested in on our Windows 2012 and Windows 2016 boxes and all looks good.

* All windows tests pass on Windows 2012 and 2016.
* Drive letters are auto assigned and incremented properly. 
* Drive letters are re-used across runs for the same branch.
* Different branches use different drive letters. 

LGTM

colinsurprenant:
rebased and merged into 5.5, 5.6, 6.0, 6.x and master.

Issue #8327, opened 2017-09-19
Add docs for ingest migration tool


suyograo:
6.0 introduced [ingest migration tool](https://github.com/elastic/logstash/issues/7329) which needs documentation in our LS reference. This tool should describe the CLI flags, sample input and output and then the limitations. For example, painless is not converted from ingest to LS.  See blog for content https://www.elastic.co/blog/ingest-node-to-logstash-configuration-converter

dedemorton:
Closed by #8399

PR #8399, opened 2017-09-27
Add docs about ingest pipeline converter

### LINKED ISSUE #7914: "Working with Filebeat Modules" doc page enhancements
Let's soften the verbiage in [this section](https://www.elastic.co/guide/en/logstash/current/filebeat-modules.html#graduating-to-Logstash). I'd suggest we change this to "Introducing Logstash" or "Using Logstash Instead of Ingest Node".

Also, for 6.0+ we should find a good place here to link off to the new ingest pipeline migration tool docs that we create.

/cc @dedemorton

dedemorton:
Closes #7914  and #8327.

Hmmm...just noticed that 8327 suggests adding a before/after example. Not sure if we want to do that. If we do, let me know as feedback to this PR.


docs/static/ingest-convert.asciidoc
```diff
@@ -0,0 +1,87 @@
...
+==== Running the tool
+
+You'll find the conversion tool in the `bin` directory of your Logstash
+installation. See <<dir-layout>> to find the location of `bin` on your system.
+
+To run the conversion tool, use the following command:
+
+[source,shell]
+-----
+bin/ingest-convert.sh --input INPUT_FILE_URI --output OUTPUT_FILE_URI [--append-stdio]
```
original-brownbear: @dedemorton maybe it's worth mentioning that at present only `file://` type URIs are supported here?

original-brownbear:
@dedemorton LGTM

PR #8400, opened 2017-09-27
Logging when queue.max_bytes is reached #7491

### LINKED ISSUE #7491: Logging when queue.max_bytes is reached
5.4.1

It will be helpful for admins if we can write a logging entry (that shows up at the default LS logging level, perhaps a WARN), when queue.max_bytes is reached.

liketic:
Resolves #7491 .

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -231,8 +231,7 @@ public class Queue implements Closeable {
 
             if (this.headPage.getMinSeqNum() <= 0 && this.headPage.getElementCount() <= 0) {
                 // head page is empty, let's keep it as-is
-
-                this.currentByteSize += pageIO.getCapacity();
+                increaseCurrentByteSize(pageIO.getCapacity());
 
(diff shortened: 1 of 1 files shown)


original-brownbear:
Jenkins test this please

original-brownbear:
@colinsurprenant maybe take a look here, I'm not so sure doing this (logging as suggested in #7491 ) is wise in the first place?
Won't we be spamming the log endlessly here if we have producers outpacing consumers (since in that case, the Queue will oscillate between full and almost full constantly)?

original-brownbear:
Path itself looks fine to me though :)

jordansissel:
I don't think logging is appropriate since a scenario with a nearly-always-full-queue will constantly print this log.

If an operator wants to monitor queue size, we have a monitoring api (http on port 9600) for this purpose.

liketic:
I'm OK to close this. @ppf2 Any thoughts? 😅

ppf2:
As long as this information is recorded somewhere, it should be fine.  If not logging, then the API (outstanding issue: https://github.com/elastic/logstash/issues/6508) :) thx

colinsurprenant:
I believe we can close this? We seems to agree that this should be better handled in though the API.

Issue #8379, opened 2017-09-23
Persistent Queues fails to deserialize after writing extremely large numeric values


rdsubhas:
- Logstash version: *any version with PQs enabled* - currently tested on 5.5 upto 5.6.1
- Platform: *any*
- Fully reproduceable test case: https://github.com/rdsubhas/logstash-queue-corruption

logstash.yml:

    http.host: "0.0.0.0"
    path.config: /usr/share/logstash/pipeline
    xpack.monitoring.enabled: false
    queue.type: persisted 

pipeline.conf

    input {
      http { }
    }
    output {
      stdout { codec => rubydebug }
    }

Send any event with a large int value:

    curl -XPOST -H 'content-type:application/json'` \
      -d '{ "some_value": 9223372036854776000 }' \
      http://127.0.0.1:8080

List of resources:
- Reproduce: https://github.com/rdsubhas/logstash-queue-corruption
- https://discuss.elastic.co/t/logstash-5-6-0-deserialization-errors-when-persistent-queues-enabled/100952
- https://discuss.elastic.co/t/logstash-crash-with-persistent-queue-and-kafka-input/90479/3
- https://discuss.elastic.co/t/logstash-5-4-repeated-deserialization-errors-and-losing-data-with-persistent-queues/91503/2
- https://discuss.elastic.co/t/logstash-persistent-queues-have-serious-data-loss-bugs-not-suitable-for-production/101566
- PR that could be a fix, scheduled for Logstash v6.1: https://github.com/elastic/logstash/issues/8131 https://github.com/elastic/logstash/pull/8371

(edited by @jordansissel to reflect accuracy)

rdsubhas:
cc @original-brownbear @guyboertje

original-brownbear:
@andrewvc the is basically a duplicate of #8131 and some related issues. The issue reproduced here (using a `long` overflow) is def. fixed in `master` and `6.x` already.

Maybe we should consider backporting the `BiValue` removal to at least `6.0`?

andrewvc:
@rdsubhas thanks for bringing this up. It looks pretty bad and I'm thinking we'll need to take action relatively quickly on this issue. 

@original-brownbear yes, I think we need to open a PR for to patch this in 6.0 as well. Is `BiValue` removal the only approach we have here?

We will definitely need a fix for 5.6 as well. 

@colinsurprenant @guyboertje  and @jordansissel , I would love you guys to weigh in on this issue as well as far as options go.

[...]

PR #8405, opened 2017-09-27
revert to Jackson BigInteger and BigDecimal serialization

### LINKED ISSUE #8387: Remove custom serializers/de-serializers for bigint and bignum
As introduced in https://github.com/elastic/logstash/pull/8205, it appears that this may be superseded by a newer version of Jackson that properly supports those types, as evident by this comment : https://github.com/elastic/logstash/issues/8379#issuecomment-332010164. 

Jackson 2.9.1 is introduced with PR https://github.com/elastic/logstash/pull/8373

I _think_ we can remove these custom serialization. 

cc/ @original-brownbear

colinsurprenant:
Revert the custom serializers that are not required anymore with Jackson version 2.8+ now that #7615 #8373  is in. 

Relates to #8379, #8205, #8131

Fixes #8387

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ObjectMappers.java
+++ b/logstash-core/src/main/java/org/logstash/ObjectMappers.java
@@ -15,6 +15,7 @@ import com.fasterxml.jackson.dataformat.cbor.CBORFactory;
 import com.fasterxml.jackson.dataformat.cbor.CBORGenerator;
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.HashMap;
 import org.jruby.RubyBignum;
 import org.jruby.RubyBoolean;
@@ -41,8 +42,6 @@ public final class ObjectMappers {
 
     private static final SimpleModule CBOR_DESERIALIZERS =
         new SimpleModule("CborRubyDeserializers")
-            .addDeserializer(RubyBigDecimal.class, new RubyBigDecimalDeserializer())
-            .addDeserializer(RubyBignum.class, new RubyBignumDeserializer())
             .addDeserializer(RubyNil.class, new RubyNilDeserializer());
 
     public static final ObjectMapper JSON_MAPPER = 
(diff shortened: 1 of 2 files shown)


colinsurprenant:
ping @original-brownbear @guyboertje @jordansissel

jordansissel:
LGTM

original-brownbear:
@jordansissel @colinsurprenant this will have somewhat negative performance repercussions, especially for the Java execution but also in the current Ruby execution.
The reason is that we're now serializing `RubyBignum` to Java's `BigInteger` when going through the full roundtrip => every `get` on the `RubyBignum` field will now entail one instantiation of `RubyBignum`.

Can't we simply move the serialize as `String` to using `.writeNumber ` as we did here and also move the deserialization to reading `Number`? That would be even faster than the current solution and also wouldn't force another adjustment of the Java execution :)

original-brownbear:
@colinsurprenant @jordansissel that said, in principle, I'm really in favor of serializing (and handling this conversion) exactly the way it's done here.
But at least @andrewvc insisted in the Java execution PR, that the Ruby behavior for comparisons and such in filters must be retained `100%`, as long as that's the case, this (disregarding the Ruby type and forcing everything to Java types after deserialization) is a big performance (or complexity in upstream code) hit :(

colinsurprenant:
@original-brownbear I am not sure I understand the impact in the Java execution and the «big performance (or complexity in upstream code) hit»? Do you mean impact and/or performance induced by this  PR? Could you expand on this please?

Also, unless I am misunderstanding you, compatibility with current plugins is not affected by this PR. 

Before this PR, for an `Event` with a `RubyBignum` field value:
- `RubyBignum` is converted to a String for serialization
- Deserialization re-instantiate a RubyBigNum from the serialized string

And hhis PR changes this to:
- `RubyBignum` is converted to `BigInteger` and serialized by Jackson internal CBOR serializer 
- Deserialization re-instantiate a `BigInteger`using  Jackson internal CBOR deserializer and is placed into the Event as-is.

So Basically the *only* differences are:
1- this PR avoids and explicit `RubyBignum` `.toString()` conversion on serialization
2- this PR injects a `BigInteger` into the Event instead of explicitly recreating a `RubyBignum` by parsing its string representation. 

Are you only suggesting we reinject a `RubyBignum` from the `BigInteger` upon deserialization? It would probably make sense from a Ruby performance perspective where the `RubyBignum` would exist once and the `BigInteger` value could be extracted and handed off to the Java API. But similarly we don't do that for `boolean` or `double` scalar values which are stored in the event as is so each Ruby get on these fields instantiate a new `RubyBoolean` for example.

original-brownbear:
@colinsurprenant 

> each Ruby get on these fields instantiate a new RubyBoolean for example.

`boolean` is a special case, there are no instantiations there in the converter since `TRUE` and `FALSE` are singletons.

`double` and `int` def are wrong, grml. Haven't seen those before :) Actually, in that case, feel free to go ahead here. No point in having an optimized path for these super rare cases and not have it for `double` and `int`.

-> LGTM :)

colinsurprenant:
@original-brownbear Great! But I am not against the idea of actually wrapping the deserialized Java `BigInteger` and `BigDecimal` into their Ruby equivalent either since we *can't loose* given that the underlying Java value is preserved in the Ruby object so it's a win-win on both sides. I guess `boolean` is an exception because of the singleton situation but we can certainly also optimize `int` and `double` in the same way, WDYT? 

I'll go ahead and merge this and then we can submit another PR to optimize?

original-brownbear:
@colinsurprenant jup, my thoughts exactly :) We can optimize this down the line in some other PR. Makes things much easier to read got log wise if we do all the numbers in one go anyhow.

colinsurprenant:
@original-brownbear great work on the BiValue refactor BTW! love it!

colinsurprenant:
merged in master and 6.x.

PR #8402, opened 2017-09-27
Fix Java compiler warnings #7701

liketic:
Partly resolves #7701

Issue #8311, opened 2017-09-19
jackson-databind version 2.7.4 should be updated


prathibhay:
logstash 6.0 is still using 2.7.3, we need to use logstash for our project, and need to update to at least 2.8.9.

jakelandis:
@prathibhay - thanks for reporting this. 

A bit googling:
https://github.com/FasterXML/jackson-databind/issues/1599
https://bugzilla.redhat.com/show_bug.cgi?id=1462702

Related issues:
https://github.com/elastic/logstash/issues/8154
https://github.com/elastic/logstash/issues/7615

jakelandis:
Closing as duplicates
https://github.com/elastic/logstash/issues/7615
https://github.com/elastic/logstash/issues/8154

PR #8416, opened 2017-10-01
Fix deprecated Serializer Use (#8402 )

original-brownbear:
Upgrading Jackson in #8311 caused a bunch of calls for scalar serialization, namely `com.fasterxml.jackson.databind.ser.std.NonTypedScalarSerializerBase`, `com.fasterxml.jackson.databind.ser.std.JsonValueSerializer.TypeSerializerRerouter#writeTypePrefixForScalar(java.lang.Object, com.fasterxml.jackson.core.JsonGenerator)` and `com.fasterxml.jackson.databind.ser.std.JsonValueSerializer.TypeSerializerRerouter#writeTypeSuffixForScalar` to become deprecated.

I didn't want to change behavior here for the time being, so I only replaced the deprecated calls with the (fortunately) non-deprecated calls they do under the hood.
=> This change should be 100% passive :) but fixes a ton of warnings for #8402  
=> Also this is a little faster than `master` because we save the redundant TypeId (the deprecated API does this separately for prefix and suffix) lookup when writing the type suffix

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ObjectMappers.java
+++ b/logstash-core/src/main/java/org/logstash/ObjectMappers.java
@@ -9,7 +9,7 @@ import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
 import com.fasterxml.jackson.databind.jsontype.TypeSerializer;
 import com.fasterxml.jackson.databind.module.SimpleModule;
-import com.fasterxml.jackson.databind.ser.std.NonTypedScalarSerializerBase;
+import com.fasterxml.jackson.databind.ser.std.StdScalarSerializer;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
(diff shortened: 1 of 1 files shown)


jakelandis:
@original-brownbear - Is there a specific behavior that we are still doing that is `deprecated` ? 

I am abit leery of copying some code to avoid the warning.

original-brownbear:
@jakelandis no I don't think the behavior used now is deprecated. This is now 100% in line with the new APIs in `2.9`. Note that the execution changed in terms of how the type ids are looked up. This was different in `2.7` APIs.

Why they removed the untyped serializer for scalars I don't know, but not writing type information for scalar types that are natively understood by CBOR is still properly supported behavior so I don't think we're doing anything wrong there either with this change.

jakelandis:
@original-brownbear - thanks for the follow up.  LGTM

original-brownbear:
@jakelandis thanks!

PR #8418, opened 2017-10-02
Update logstash-core-plugin-api to source version from versions.yml

robbavey:
Use same mechanism as logstash and logstash-core to retrieve version
information


robbavey:
@guyboertje Ping - I think this got lost during the support summit

PR #8428, opened 2017-10-03
PERFORMANCE[30%+ improvement!]: filter_func can optionally take an array of events to make batched filters much faster

original-brownbear:
This one is kinda frustrating when seen in relation to #8357 :( but a huge and trivial performance gain :)
Just thought I'd level the playing field a little when comparing Java and Ruby exec and well this is the result.

Simply fixing the batched filter execution in the Ruby execution gets us within to about 90% (I can add more scientific numbers tomorrow or so, but it's basically this for the Apache example: master **35k/s**, this **50k/s** and the full Java exec **56k/s** ) of the throughput of the pure java exec in #8357.
For baseline (and simple filters like only UA filter) this version wins over `master` and the Java exec by a wide margin. Here it's `master` and Java exec both at **~300k/s** and this one at **~380k/s**. 

To better understand this one, take a look at a compiled `filter_func`:

example:
```rb
  define_singleton_method :filter_func do |event|
    events = [event]
    @logger.debug? && @logger.debug("filter received", "event" => event.to_hash)
              events = @generated_objects[:filter_dlq_commit_2].multi_filter(events)

    events
  end
```

=> there is no point turning the events into arrays one by one if we can simply turn the batch into an array and input that.
=> trivial fix by allowing `Array` to be passed through `filter_func` and adding the `to_a` method from #8357 to the batch implementations

### DIFF
--- a/logstash-core/lib/logstash/config/config_ast.rb
+++ b/logstash-core/lib/logstash/config/config_ast.rb
@@ -77,7 +77,7 @@ module LogStash; module Config; module AST
         # of the output/filter function
         definitions << "define_singleton_method :#{type}_func do |event|"
         definitions << "  targeted_outputs = []" if type == "output"
-        definitions << "  events = [event]" if type == "filter"
+        definitions << "  events = event" if type == "filter"
         definitions << "  @logger.debug? && @logger.debug(\"#{type} received\", \"event\" => event.to_hash)"
 
         sections.select { |s| s.plugin_type.text_value == type }.each do |s|
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -495,11 +495,9 @@ module LogStash; class Pipeline < BasePipeline
   end
 
   def filter_batch(batch)
-    batch.each do |event|
-      filter_func(event).each do |e|
-        #these are both original and generated events
-        batch.merge(e) unless e.cancelled?
-      end
+    filter_func(batch.to_a).each do |e|
+      #these are both original and generated events
+      batch.merge(e) unless e.cancelled?
     end
     @filter_queue_client.add_filtered_metrics(batch)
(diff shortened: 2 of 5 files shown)


original-brownbear:
@andrewvc @ph @jordansissel FYI not sure if I should be happy or sad here after all the work that went into Java exec ... but this looks pretty safe to merge to give `6.1` a massive boost imo ... obviously way diminishing people's incentive to try out the Java exec ...

original-brownbear:
Also, this version beats the Java exec by quite a margin in baseline (17% faster) stdin->stdout or just a single fast filter like the UA (10% faster) one in place :(

ph:
Woa, good catch! This change is low risk we could release it in 6.0.x and event 5.6.x if we ever do another release. I am doing some testing, will report soon.

colinsurprenant:
oh sweet <3  🥇  
also doing testing.

ph:
> obviously way diminishing people's incentive to try out the Java exec

Concerning the java execution, we will make it faster eventually, that should block us from merging it with a feature flag.

original-brownbear:
@ph should or "should not"? :D

ph:
*should not* :)


I did some testing with a really simple pipeline configuration
```
filter {
    mutate { uppercase => [ "message" ] }
}

I got a 7% improvements with the code changes, but as soon as you add more plugins to the filter stage, I get the 30% speed improvements.

```
filter {
    mutate { uppercase => [ "message" ] }
    mutate { lowercase => [ "message" ] }
    mutate { uppercase => [ "message" ] }
    mutate { lowercase => [ "message" ] }
}
```

original-brownbear:
@ph alrighty thanks, merging :)

original-brownbear:
needs manual backport to 6.0 and 5.6, will do later

jordansissel:
Thank you for working on this :)

original-brownbear:
np :)

PR #8430, opened 2017-10-04
PERFORMANCE: remove blocking on inflight batches map during batch reads

original-brownbear:
Same as #8081 in general:

* use a concurrent map instead of synchronized access to a `RubyHash`

Differences to #8081:

* We are iterating over the map => we cannot set `nil` values instead of deleting a key safely
 
Note:
* Iteration over this map is threadsafe, `Concurrent::Map` internally synchronizes it

no huge gain here for small worker counts, but simplifies the code a little, removes a piece of code that doesn't scale linearly with worker count and gives us a little boost on systems with many cores/workers.

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -46,10 +46,9 @@ module LogStash; module Util
 
       def initialize(queue, batch_size = 125, wait_for = 250)
         @queue = queue
-        @mutex = Mutex.new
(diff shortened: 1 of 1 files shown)


original-brownbear:
@guyboertje thanks :)

guyboertje:
Ironically, I was looking at this today. When shutdown_watcher exec the previous code, the yielding of the hash while inside the lock holds the lock that the threads the watcher is trying to report on might be then waiting on in batch_read and batch_close 😃 . This fixes that _bug_ too.

PR #8431, opened 2017-10-04
add missing default plugins to Gemfile template

jsvd:
forward port of https://github.com/elastic/logstash/commit/dc773ac551cdf79d337a723cc897b0128e68ffe4 to master.

Several plugins were added as shipping by default with logstash, but the Gemfile.template wasn't updated. In the future these two should be updated in a single operation


jordansissel:
Interesting. I would prefer that we only have one source that defines
default plugins. -- so either the gemfile or the JSON plugins file. Not a
blocker for this Pr though.

PR #8432, opened 2017-10-04
WIP: Write Segments + their Hash Atomically

original-brownbear:
For #8142 good mix of issues here:

* A segment + its accompanying hash must be written atomically
   * Also made the CRC32 reusable here and not (only) for performance reasons: It's just a nice double check to do it this way, if we get a broken CRC32 written, we instantly know it's aconcurrency issue now with the writer ... ignoring HW problems
   * I would suggest to keep going in another issue here btw. and make these writes atomic per event + speed this up some more. I already fixed at least `writeRecordHeader` but there is many spots like this, making it take ~10s to write 3k events which makes tests take forever + is likely a real issue in production, putting a lot of load on things once the DLQ kicks in.
* The DLQ reader was actually polling for segments when reaching the last segment, before trying to simply get a list of files without the event triggers (note that there are no hard guarantees on the `WatchService` on all filesystems/OS, if you are unlucky and miss the event for a file creating this will break tests by missing the last file) in `org.logstash.common.io.DeadLetterQueueReader#pollEntryBytes(long)`. This caused `org.logstash.common.io.DeadLetterQueueReaderTest#testReadFromTwoSegments` to take more than `10s`, after this fix it only takes `~0.5s`
* Last but not least: `Files.list(path)` returns a `Stream<Path>` that needs to be closed as it has an underlying fd on the DLQ directory. Example `lsof -p` output without this fix:

```sh
java    37798 brownbear  138r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  139r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  140r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  141r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  142r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  143r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  144r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  145r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  146r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  147r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  148r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  149r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  150r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  151r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  152r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  153r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  154r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  155r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  156r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
java    37798 brownbear  157r    DIR               1,4        68 43494065 /private/var/folders/fq/hx0pm7gd5d91p9lkgj11fvtr0000gn/T/junit6972758946827880057/junit5980346763334154665
```
(this causes a major slowdown and outright crashes by running out of FDs in continuous operation eventually, breaking the DLQ in general imo)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReader.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReader.java
@@ -53,7 +53,7 @@ public final class DeadLetterQueueReader implements Closeable {
         this.queuePath.register(
             watchService, new WatchEvent.Kind[]{
                 StandardWatchEventKinds.ENTRY_CREATE, StandardWatchEventKinds.ENTRY_DELETE
-            }, 
+            },
             SensitivityWatchEventModifier.HIGH
         );
         this.segments = new ConcurrentSkipListSet<>(Comparator.comparing(
@@ -87,14 +87,10 @@ public final class DeadLetterQueueReader implements Closeable {
             .filter(Boolean::booleanValue).count() == 0L) {
             WatchKey key = watchService.poll(timeout, TimeUnit.MILLISECONDS);
             if (key != null) {
-                for (WatchEvent<?> watchEvent : key.pollEvents()) {
-                    if (watchEvent.kind() == StandardWatchEventKinds.ENTRY_CREATE) {
-                        getSegmentPaths(queuePath).forEach(segments::add);
-                        break;
-                    }
-                }
+                key.pollEvents();
                 key.reset();
             }
+            getSegmentPaths(queuePath).forEach(segments::add);
         }
         return System.currentTimeMillis() - startTime;
     }
--- a/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReader.java
+++ b/logstash-core/src/main/java/org/logstash/common/io/DeadLetterQueueReader.java
@@ -19,23 +19,22 @@
 package org.logstash.common.io;
 
 import java.io.Closeable;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
-import org.logstash.DLQEntry;
-import org.logstash.Timestamp;
-
 import java.io.IOException;
-import java.nio.file.Files;
 import java.nio.file.FileSystems;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardWatchEventKinds;
 import java.nio.file.WatchEvent;
 import java.nio.file.WatchKey;
 import java.nio.file.WatchService;
+import java.util.Comparator;
 import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Function;
 import java.util.stream.Collectors;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.logstash.DLQEntry;
+import org.logstash.Timestamp;
 
 import static java.nio.file.StandardWatchEventKinds.ENTRY_CREATE;
 import static java.nio.file.StandardWatchEventKinds.ENTRY_DELETE;
@@ -53,22 +52,20 @@ public final class DeadLetterQueueReader implements Closeable {
         this.queuePath = queuePath;
         this.watchService = FileSystems.getDefault().newWatchService();
         this.queuePath.register(watchService, ENTRY_CREATE, ENTRY_DELETE);
-        this.segments = new ConcurrentSkipListSet<>((p1, p2) -> {
-            Function<Path, Integer> id = (p) -> Integer.parseInt(p.getFileName().toString().split("\\.")[0]);
-            return id.apply(p1).compareTo(id.apply(p2));
-        });
-
+        this.segments = new ConcurrentSkipListSet<>(Comparator.comparing(
+            p -> Integer.parseInt(p.getFileName().toString().split("\\.")[0]))
+        );
         segments.addAll(getSegmentPaths(queuePath).collect(Collectors.toList()));
     }
 
     public void seekToNextEvent(Timestamp timestamp) throws IOException {
         for (Path segment : segments) {
             currentReader = new RecordIOReader(segment);
-            byte[] event = currentReader.seekToNextEventPosition(timestamp, (b) -> {
+            byte[] event = currentReader.seekToNextEventPosition(timestamp, bytes -> {
                 try {
-                    return DLQEntry.deserialize(b).getEntryTime();
-                } catch (IOException e) {
-                    throw new IllegalStateException(e);
+                    return DLQEntry.deserialize(bytes).getEntryTime();
+                } catch (final IOException ex) {
+                    throw new IllegalStateException(ex);
                 }
             }, Timestamp::compareTo);
             if (event != null) {
@@ -85,10 +82,10 @@ public final class DeadLetterQueueReader implements Closeable {
         if (key != null) {
             for (WatchEvent<?> watchEvent : key.pollEvents()) {
                 if (watchEvent.kind() == StandardWatchEventKinds.ENTRY_CREATE) {
-                    segments.addAll(getSegmentPaths(queuePath).collect(Collectors.toList()));
+                    getSegmentPaths(queuePath).forEach(segments::add);
                 }
-                key.reset();
             }
+            key.reset();
         }
         return System.currentTimeMillis() - startTime;
     }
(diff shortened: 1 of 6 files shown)


original-brownbear:
Jenkins test this please

original-brownbear:
Jenkins test this please

jordansissel:
There's multiple changes in this PR that I can see.

The described change (atomic write) seems to be that previously two channel writes are now one channel write (the channel write in writeRecordHeader is now writing to a buffer instead).

The additional changes:
* Some code cleanup (ByteBuffer slice removed?)
* Introduction of PADDING_BYTES (and the code using it is unclear to me, at a glance)
* Introduction of a reusable ByteBuffer as private field writeBuffer (vs creating at each writeEvent)
* Introduction of reusable CRC32 private field (instead of creating at each writeEvent)

Can you focus PR this on just the atomic write and split the other changes into separate PRs? At a minimum, the commit message and PR description must detail the changes and the intent of the changes.

original-brownbear:
@jordansissel sorry ignore this one, for now, please. This is just an attempt at learning more about #8142 :)

jordansissel:
*facepalm* I didn't see the WIP label. Will ignore for now :)

original-brownbear:
Jenkins test this please

robbavey:
@original-brownbear Thanks for looking at this - I would tend to agree that it makes sense to write the checksum and the segment together - the errors look like it could be an over-eager reader trying to read before the write is complete. 

If you reproduce this locally, you should have access to the dlq files from failure, where the segments can be re-read to determine if the read operations succeed at a later time.

original-brownbear:
@robbavey can't reproduce this locally anymore :( But wait :) so we do read and write concurrently in that test?
In that case this may be a fix :D

original-brownbear:
Jenkins test this please

robbavey:
@original-brownbear These tests are in dlq_spec:

-  Multi-pipeline where one pipeline has a generator throwing data in elasticsearch, and onto the dlq when mapping failures occur, and the other a dlq input reading from the dlq
 -  Single pipeline with a generator input and a dlq input, doing similar work

Enjoy the Jenkins, ... dance. Failures occur irritatingly infrequently ;)

original-brownbear:
@robbavey huh well ... in that case, we got a fix here (imo). Too tired now, will clean this up tomorrow and we can take a look if this is actually sufficient or if it just makes things way less likely to fail.

robbavey:
@original-brownbear At the very least, I don't think this change can make things worse and will likely have a positive effect. Get some rest, and we can work through this tomorrow.

original-brownbear:
@robbavey bam you're right:


```java
    public void testConcurrentWriteReadRandomEventSize() throws Exception {
        Event event = new Event(Collections.emptyMap());
        int eventCount = 3000;
        int maxEventSize = BLOCK_SIZE * 2;
        final ExecutorService exec = Executors.newSingleThreadExecutor();
        try {
            exec.submit(() -> {
                long startTime = System.currentTimeMillis();
                try (DeadLetterQueueWriter writeManager = new DeadLetterQueueWriter(
                    dir, 10 * 1024 * 1024, 1_000_000_000L)) {
                    for (int i = 0; i < eventCount; i++) {
                        event.setField(
                            "message",
                            generateMessageContent((int) (Math.random() * (maxEventSize)))
                        );
                        DLQEntry entry = new DLQEntry(event, "", "", String.valueOf(i),
                            new Timestamp(startTime++)
                        );
                        writeManager.writeEntry(entry);
                    }
                } catch (final IOException ex) {
                    throw new IllegalStateException(ex);
                }
            });
            try (DeadLetterQueueReader readManager = new DeadLetterQueueReader(dir)) {
                for (int i = 0; i < eventCount; i++) {
                    DLQEntry entry = readManager.pollEntry(1000L);
                    assertThat(entry.getReason(), is(String.valueOf(i)));
                }
            }
        } finally {
            exec.shutdown();
            if (!exec.awaitTermination(2L, TimeUnit.MINUTES)) {
                Assert.fail("Failed to shut down record writer");
            }
        }
    }
```
(just the existing random size test adjusted for concurrent reads and writes)
fails randomly ... unfortunately, it does so with this branch too (though at a lower rate).
=> you're right twice :D this needs some rest first :D, let's look into it tomorrow :)

original-brownbear:
@robbavey alrighty, the above test works fine when setting the timeout to a higher number and reproduces the checksum failure with `master` =>  :) Added it to this PR but ignored it :(

It takes 14s to run currently, much much longer than sequentially writing and then reading the same amount of data.
I couldn't find a quick fix for that, but it fits in with the fact the DLQ IT is very slow too. I'd suggest just leaving it ignored and adding an issue to figure out why concurrent R/W is so slow (I don't understand the Reader code well enough yet to make a clean fix here and I'd rather not add a new stability issue in the same PR that fixes one :D).

Ping me later if you have a sec to look over this :)

original-brownbear:
@robbavey fixed :) see updated PR description. Quite a few issues I found, each of them breaking a part of this. With the fixes now the tests appear rock solid, ran the concurrent R/W one in a loop for quite a while (~30min+) without issues.

Happy to walk through this in Slack later if that makes reviewing this easier.

original-brownbear:
Hold off here, it passes in 100% locally now and fails on CI every time ... back to it grml

original-brownbear:
@robbavey Ok so this can't be fixed without a design change imo, let's talk about this in Slack/Zoom later. Long story :) But at least we have a clean reproducer and some aspects of this fixed with what I learned here.
I'll open a new PR for the FD leak, we should fix that right away since it's really bad, the rest will take more/different effort than the fixes here.

original-brownbear:
closing this for now, summary of the problems with DLQ incoming as an issue.

PR #8438, opened 2017-10-05
BUG: DLQ FD leak in org.logstash.common.io.DeadLetterQueueWriter#getSegmentPaths fixed

original-brownbear:
Extracted from #8432 

The stream returned by `Files.list(path)` is an `AutoClosable` that must actually be closed since it's backed by an FD on the underlying directory.
Not fixing this leads to every call to this method leaking one `DIR` type file descriptor.

This should be backported to all versions using DLQ imo since it will eventually make every DLQ reader run out of FDs.


original-brownbear:
@robbavey thanks merging then :)

original-brownbear:
hmpf other branches need manual backport 

@robbavey @jordansissel backport this to `5.6` as well?

PR #8440, opened 2017-10-05
Test fix: disable the gradle daemon by default (as recommended for CI)

### LINKED ISSUE #8381: Test Failure: Gradle crashes
Every so often Gradle will just die with the following error. I have seen this happen on the compatibility tests which are not Docker based, and so far only on OEL and CentOS. 

```
03:10:55 FAILURE: Build failed with an exception.
03:10:55 
03:10:55 * What went wrong:
03:10:55 Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)
```

I don't think there is anything wrong our code, but if we find this happen more often we may want to consider further examination of our build environments.

jakelandis:
Fixes #8381

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,1 @@
+org.gradle.daemon=false


jakelandis:
@ph thanks!

PR #8443, opened 2017-10-06
PERFORMANCE: Simplify batch iteration

original-brownbear:
We can get another small speedup on the batch iterations leveraging #8428:

* As a result of #8428 we don't have any more concurrent modifications on the data underlying a batch since the filters always take all the events in a single call and will only `merge` in generated events after the batch generated an array via `to_a` :) => we can save ourselves the trouble of having temporary collections for holding generated events entirely
   * note that I had to adjust one spec to use `to_a.each` instead of just `each` as a result since that spec was actually doing concurrent modifications in its test code
* For the in-memory Queue, this effectively reduces the batch to just being a `HashSet` wrapper
  * Also for the sync queue batch, we could nicely inline the each block over a `Collection` which is a little faster than custom each blocks  
* For the acked queue, we also saved one data structure but still track generated and original events separately (I think we can clean this fact up in another PR though, something is weird with the counts here since filtered_size is synonymous with size for the in-memory Queue but not for the acked queue => will add an issue for that soon)

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
@@ -161,21 +161,12 @@ module LogStash; module Util
         # @cancelled = Hash.new
 
         #Sizing HashSet to size/load_factor to ensure no rehashing
-        @is_iterating = false # Atomic Boolean maybe? Although batches are not shared across threads
         @originals = LsQueueUtils.drain(queue.queue, size, wait)
       end
 
       def merge(event)
-        return if event.nil? || @originals.contains(event)
-        # take care not to cause @generated to change during iteration
-        # @iterating_temp is merged after the iteration
-        if @is_iterating
-          @iterating_temp = HashSet.new if @iterating_temp.nil?
-          @iterating_temp.add(event)
-        else
-          # the periodic flush could generate events outside of an each iteration
-          @originals.add(event)
-        end
+        return if event.nil?
+        @originals.add(event)
       end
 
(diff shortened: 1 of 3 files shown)


logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb
```diff
@@ -186,21 +177,14 @@ def cancel(event)
 
       def to_a
         events = []
-        each {|e| events << e}
+        @originals.each {|e| events << e unless e.cancelled?}
```
guyboertje: This whole method can be 
```
def to_a
  @originals.select{|e| !e.cancelled?}
end
```
`select` will create a new array

original-brownbear: @guyboertje I "think" `select` will be slower here since `@originals` is a `java.util.HashSet` <= Ruby has to provide the select method via its wrapper logic for these collections.
I'd rather keep this and improve it another PR to simply take an array as argument for `to_a` so we don't have to recreate those `[]` constantly?

original-brownbear: hmm though the opposite could be true as well ... let me benchmark this :D sec

original-brownbear: hmm no measurable difference, but in theory `select` must be slower, it's the same code path plus some special case handling for size `1` and `2` which we never really make use of.
=> I'll leave it as is :)

PR #8446, opened 2017-10-06
PERFORMANCE: Faster Ruby pipeline output_func + worker loop

original-brownbear:
Some trivial improvements here:

* Tiny Improvement: Cache `batch_size` to save one object (each number is an object with an expensive constructor)
* Tiny improvement: Removed `nil` return fallback from `output_func`, there are no more specs left that need this
* Bigger improvement (~3-5% in baseline throughput, better if you have more than one output): Cache `output_events_map ` instead of recreating it over and over. This is especially useful, since we can then use `clear` on the bins and save a bunch of allocations of these arrays


original-brownbear:
@ph thanks!

PR #8448, opened 2017-10-06
PERFORMANCE: Avoid redundant (and high overhead) size calls to Java HashMap

original-brownbear:
Cleaning up the metrics interaction in the main loop a little to cut back on some noise in profiling :)

* Don't do redundant `size` calls to Java's `HashMap`, this is really costly in JRuby (unfortunately)
* Also `add_starting_metrics ` is just dead code, seems it got left over at some point:

grep:
```sh
➜  logstash git:(fewer-steps-dataset) ag -S add_starting_metrics
➜  logstash git:(fewer-steps-dataset) 
```

=> dead => removed :)


original-brownbear:
@guyboertje thanks!

PR #8449, opened 2017-10-06
Document breaking changes in bundled plugins

robbavey:
Also updates reference to 6.0 for plugins no longer bundled

### DIFF
--- a/docs/static/breaking-changes.asciidoc
+++ b/docs/static/breaking-changes.asciidoc
@@ -39,7 +39,7 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
 [float]
 ==== Elasticsearch output changes
 
-* The default `document_type` has changed from `logs` to `doc`.
+* The default `document_type` has changed from `logs` to `doc` for consistency with Beats.
   Furthermore, users are advised that Elasticsearch 6.0 deprecates doctypes, and 7.0 will remove them. 
   See https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html[Removal of Mapping Types] for more info.
 * The options `flush_size` and `idle_flush_time` are now obsolete.
@@ -49,23 +49,25 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
 [float]
 ==== Kafka input changes
 
-* Updated to 0.10.1.0 client protocol. Only supports brokers 0.10.1.x or later
-* Decorated fields are now nested under `@metadata` to avoid mapping conflicts with beats
-* The `ssl` option is now obsolete
+Upgraded Kafka client support to v0.11.0.0, which only supports Kafka brokers v0.10.x or later.
+** Please refer to <<plugins-input-kafka,Kafka input plugin>>  documentation for information about Kafka compatibility with Logstash.
+* Decorated fields are now nested under `@metadata` to avoid mapping conflicts with Beats.
+** See the `Metadata Fields` section in the <<plugins-input-kafka>> documentation for more details.
+* The `ssl` option is now obsolete.
 
 [float]
 ==== Kafka output changes
 
-* Updated to 0.10.1.0 client protocol. Only supports brokers 0.10.1.x or later
-* The options `block_on_buffer_full`, `ssl` and `timeout_ms` are now obsolete
+Upgraded Kafka client support to v0.11.0.0, which only supports Kafka brokers v0.10.x or later.
+** Please refer to <<plugins-output-kafka, Kafka output plugin>> documentation for information about Kafka compatibility with Logstash.
+* The options `block_on_buffer_full`, `ssl`, and `timeout_ms` are now obsolete.
 
 [float]
 ==== Beats input changes
 
-* Logstash will no longer start when multiline codec is used with the Beats input plugin
-* The options `congestion_threshold` and `target_field_for_codec` are now obsolete
-
-
+* Logstash will no longer start when <<plugins-codecs-multiline,Multiline codec plugin>> is used with the Beats input plugin.
+** It is recommended to use the multiline support in Filebeat as a replacement - see https://www.elastic.co/guide/en/beats/filebeat/current/multiline-examples.html[configuration options available in Filebeat] for details.
+* The options `congestion_threshold` and `target_field_for_codec` are now obsolete.
 
 [float]
 ==== List of plugins bundled with Logstash
--- a/docs/static/breaking-changes.asciidoc
+++ b/docs/static/breaking-changes.asciidoc
@@ -34,7 +34,9 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
 * Configurations provided with `-f` or `config.path` will not be appended with `stdin` input and `stdout` output automatically.
 
 [float]
+=== Plugin Changes
 
+[float]
 ==== Elasticsearch output changes
 
 * The default `document_type` has changed from `logs` to `doc`.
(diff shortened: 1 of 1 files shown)


robbavey:
Notes on PR:

- This PR doesn't include all the changes in bundled plugins where deprecated changes were marked obsolete - if required, I can add those in too, or add a general note to that effect.
- The list of removed bundled plugins includes `logstash-codec-oldlogstashjson`, which I didn't see in the list of plugins from 5.6.2, should this still be included?

docs/static/breaking-changes.asciidoc
```diff
@@ -45,9 +47,30 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
...
+[float]
+==== Kafka output changes
+
+* Updated to 0.10.1.0 client protocol. Only supports brokers 0.10.1.x or later
+* The options `block_on_buffer_full`, `ssl` and `timeout_ms` are now obsolete
+
+[float]
+==== Beats input changes
+
+* Logstash will no longer start when multiline codec is used with the Beats input plugin
```
jordansissel: Can we suggest corrective action for this?

This line says what is not working, but for users affected, giving a pointer to the next step is useful --

"It is recommended to use filebeat's multiline support as a replacement." or some similar language?

docs/static/breaking-changes.asciidoc
```diff
@@ -34,7 +34,9 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
 * Configurations provided with `-f` or `config.path` will not be appended with `stdin` input and `stdout` output automatically.
 
 [float]
+=== Plugin Changes
 
+[float]
 ==== Elasticsearch output changes
 
 * The default `document_type` has changed from `logs` to `doc`.
```
jordansissel: Let's include some next-step information. For example, referencing why this change was made (document type deprecated in ES 6.0) and include a link for further information (link to Elasticsearch's own docs on this topic).

docs/static/breaking-changes.asciidoc
```diff
@@ -45,9 +47,30 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
  The new mapping template has been updated to reflect that. If you are using a custom mapping template you may need to update it to reflect that.
 
 [float]
+==== Kafka input changes
+
+* Updated to 0.10.1.0 client protocol. Only supports brokers 0.10.1.x or later
```
jordansissel: This looks correct, but gives me pause to ask a question -- Kafka 0.11.x is the latest. Should we ship LS 6.0 with Kafka 0.11.x  by default?

(If Kafka 0.11.x server will work with a Kafka 0.10.x client, then we don't need to upgrade maybe)

docs/static/breaking-changes.asciidoc
```diff
@@ -45,9 +47,30 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
...
+* Decorated fields are now nested under `@metadata` to avoid mapping conflicts with beats
+* The `ssl` option is now obsolete
```
jordansissel: Let's add corrective action advice here, if any is necessary.

docs/static/breaking-changes.asciidoc
robbavey: Actually, I think you are right here - we do ship with 0.11.x, which is backwards compatible with 0.10.1.x. I missed it, as it was in a non-breaking change

docs/static/breaking-changes.asciidoc
robbavey: For changing the document_type from `logs` to `doc`, does it make it sense to use the same wording from the changelog of the plugin - ```Change default document type to 'doc' from 'logs' to align with beats and reflect the generic nature of logstash```

There is already a link to the Elasticsearch docs on removal of types - is there another link you are looking to include here?

docs/static/breaking-changes.asciidoc
robbavey: Is it sufficient to point back to the `metadata fields` in the plugin documentation here?

docs/static/breaking-changes.asciidoc
ph: Could we also point to the FB's multiline documentation? 

https://github.com/logstash-plugins/logstash-input-beats/issues/201#issuecomment-304729091

robbavey:
ping

docs/static/breaking-changes.asciidoc
```diff
@@ -34,20 +34,45 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
...
   Furthermore, users are advised that Elasticsearch 6.0 deprecates doctypes, and 7.0 will remove them. 
   See https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html[Removal of Mapping Types] for more info.
 * The options `flush_size` and `idle_flush_time` are now obsolete.
 * Please note that the https://www.elastic.co/guide/en/elasticsearch/reference/6.0/mapping-all-field.html[_all] field is deprecated in 6.0.
  The new mapping template has been updated to reflect that. If you are using a custom mapping template you may need to update it to reflect that.
 
 [float]
+==== Kafka input changes
+
+* Upgrade Kafka client to version 0.11.0.0. Only supports brokers 0.10.1.x or later.
```
dedemorton: I'd put this in the past tense: "Upgraded the Kafka client..."

docs/static/breaking-changes.asciidoc
```diff
@@ -34,20 +34,45 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
...
+** Please refer to <<plugins-input-kafka>> documentation for information about Kafka compatibility with Logstash.
+* Decorated fields are now nested under `@metadata` to avoid mapping conflicts with Beats.
+** See the `Metadata Fields` section in the <<plugins-input-kafka>> documentation for more details.
+* The `ssl` option is now obsolete.
+
+[float]
+==== Kafka output changes
+
+* Upgrade Kafka client to version 0.11.0.0. Only supports brokers 0.10.1.x or later.
```
dedemorton: Change to: "Upgraded the Kafka client..."

docs/static/breaking-changes.asciidoc
```diff
@@ -34,20 +34,45 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
...
+** Please refer to <<plugins-input-kafka>> documentation for information about Kafka compatibility with Logstash.
```
dedemorton: So that the link resolves to the text you expect, you should say "Please refer to the <<plugins-input-kafka,Kafka input plugin>> documentation....."

This is necessary because we use abbreviated titles to make the navigation more readable, but it changes the link text, too.

docs/static/breaking-changes.asciidoc
```diff
@@ -34,20 +34,45 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
...
+* The options `block_on_buffer_full`, `ssl` and `timeout_ms` are now obsolete.
```
dedemorton: Minor: we use the Oxford comma, so say: "...`block_on_buffer_full`, `ssl`, and `timeout_ms` are now obsolete."

docs/static/breaking-changes.asciidoc
```diff
@@ -34,20 +34,45 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
...
+
+[float]
+==== Beats input changes
+
+* Logstash will no longer start when <<plugins-codecs-multiline>> is used with the Beats input plugin.
```
dedemorton: For the same reason mentioned earlier, you should say: "...when the <<plugins-codecs-multiline,Multiline codec plugin>> is...."

docs/static/breaking-changes.asciidoc
```diff
@@ -34,20 +34,45 @@ These changes can impact any instance of Logstash and are plugin agnostic, but o
...
+** It is recommended to use filebeat's multiline support as a replacement - see https://www.elastic.co/guide/en/beats/filebeat/current/multiline-examples.html[configuration options available in Filebeat] for details.
```
dedemorton: It's good to avoid making product names possessive. Also make sure you capitalize Filebeat: "It is recommended to use the multiline support in Filebeat as a replacement..."

docs/static/breaking-changes.asciidoc
acchen97: Should we add a bit more context here?

Upgraded Kafka client support to v0.11.0.0 which only supports Kafka brokers v0.10.1.x or later. For usage with Kafka versions older than v0.10.1.x, please consult the https://www.elastic.co/guide/en/logstash/current/plugins-inputs-kafka.html#_description_23[compatibility matrix] and install the correct plugin version.

docs/static/breaking-changes.asciidoc
acchen97: same comment as above

docs/static/breaking-changes.asciidoc
acchen97: Sorry, missed the next line... looks fine.

acchen97:
Looks good overall, thanks for working on this @robbavey.

docs/static/breaking-changes.asciidoc
robbavey: @acchen97 FYI - There are outstanding issues to update the compatibility matrices for the Kafka plugins - https://github.com/logstash-plugins/logstash-input-kafka/issues/214 and https://github.com/logstash-plugins/logstash-output-kafka/issues/142

jordansissel:
Re: Kafka matrix -- we should remove it entirely. I have an open issue on
the kafkaoutput with a proposal.

robbavey:
@jordansissel @dedemorton I've put up another rev of this to check

dedemorton:
LGTM

robbavey:
@dedemorton @jordansissel Does this also need to go to 6.x?

jordansissel:
@robbavey if it documents "6.0 broke the following things" I think so -- mainly to aid users upgrading from 5.x to 6.x. We dont' break things in minor releases, though, so it's going to be focused on 5.x -> 6.0 "breaks" but we still want it documented in all 6.x minors.

PR #8453, opened 2017-10-07
Remove PathCache Class

original-brownbear:
We talked about this in one of our Java exec meetings.
On the road to getting some adoption of `FieldReference` the concept of `PathCache` is kind of confusing.

Also, removing `PathCache` and integrating this all into `FieldReference` makes sure we don't expose `FieldReference.parse()` anymore which we shouldn't, given that `FieldReferences` are globally deduped.


jordansissel:
LGTM

PR #8454, opened 2017-10-07
PERFORMANCE: Faster Timestamp Ctor

original-brownbear:
A small but measurable gain here. We're instantiating millions of these (at least one per `Event`), causing us to spend about 3% of CPU time on constructing `Timestamp` in the Apache benchmark example (this patch brings that down to `1%`).
The `DataTime` constructor working from `Chronology` is much faster since it doesn't have a `ConcurrentHashmap` lookup for the `Chronology` and as a result, inlines here.


jakelandis:
LGTM

original-brownbear:
@jakelandis thanks!

PR #8461, opened 2017-10-10
Fix shared config options

dedemorton:
Adds reference to asciidoc attributes to resolve the correct plugin name. Also fixes the `id` example so that it's relevant to the specific plugin where it's included. 

Note that curly braces need to be escaped with backslashes when `subs="attributes"` is set.


jordansissel:
LGTM

PR #8470, opened 2017-10-10
Bump to 5.6.4

robbavey:


logstash-core/lib/logstash-core_jars.rb
```diff
@@ -17,9 +15,7 @@
   require_jar( 'org.apache.logging.log4j', 'log4j-core', '2.6.2' )
   require_jar( 'com.fasterxml.jackson.core', 'jackson-databind', '2.9.1' )
   require_jar( 'org.apache.logging.log4j', 'log4j-api', '2.6.2' )
-  require_jar( 'org.slf4j', 'slf4j-api', '1.7.21' )
```
jsvd: any idea about why these are being removed? it may be an unnecessary change for a patch release.

original-brownbear: @robbavey @jsvd yea I wouldn't do this without a very good reason. Seems risky to remove a potential runtime dependency in a patch?

jakelandis: Agreed slf4j API and its log4j2 binding (below) provide can provide real value with respect to 3rd party, since a great deal 3rd parties code/complied against slf4j.

robbavey: This file is automatically generated when running `rake artifact:tar`, as part of the [version bump](https://github.com/elastic/dev/blob/master/logstash/version_bump.md) process. 

This change in dependencies appears to have been brought in with the [JrJackson change](https://github.com/elastic/logstash/commit/4e0ff4e9e9dbbbdeac5f309df8266f27464d4331)

original-brownbear: @robbavey urgh :) this is a tricky spot. I guess even if we manually change this back this jar won't be in the `tar`?

original-brownbear: I guess this leaves us with the choice of manually adding the dep to Gradle (which is kind of nasty too) or taking the risk :(
I'm almost willing to take the risk and fix whatever plugin may break accidentally if it breaks wdyt guys?

robbavey: This [gist](https://gist.github.com/robbavey/8beea2f3f86498dcd9143e5b18acde56) shows the contents of the tar file generated by the rake task

original-brownbear: hmm, so it's in there https://gist.github.com/robbavey/8beea2f3f86498dcd9143e5b18acde56#file-gistfile1-txt-L60 => re-adding that include manually should be just fine and remove all risk?

original-brownbear: ah nevermind, it's in there because Kafka input pulls it :D idk ... I think I'd just leave things as they are in this PR then. Forcing the build into this kind of temporary/broken-state just to get a runtime jar in that may or may not be used elsewhere seems to not be worth it.
At least Kafka input knows how to pull its own jar anyways :D

robbavey: Alright. I am now super confused. The slf4j and log4j dependencies were brought into 5.6 in [this commit](https://github.com/elastic/logstash/commit/4e0ff4e9e9dbbbdeac5f309df8266f27464d4331). I wonder if this was added by hand, because this line also added in that commit:
```
require 'com/fasterxml/jackson/module/jackson-module-afterburner/2.7.3/jackson-module-afterburner-2.9.1.jar'
```
seems bogus to me - @guyboertje any ideas here?

jakelandis: If we don't include the `require_jar( 'org.apache.logging.log4j', 'log4j-slf4j-impl', '2.6.2' )` binding on our core classpath, then the logging may break for any 3rd parties coded against slf4j.  3rd party libraries shouldn't be assembling the slf4j implementation (it kinda defeats the purpose of slf4j). 

I would vote to add the -impl dependency in Gradle (which should bring in the -api).

robbavey:
This should get things back to normal - the gemspec_jars.rb wasn't getting updated .

jakelandis:
LGTM

Issue #8382, opened 2017-09-25
pipeline.queue.events is updated only after logstash restart


dandrestor:
The monitoring API value `pipeline.queue.events` does not update while logstash is up, only after restarting the daemon. This applies to logstash running with the persistent queue option enabled.

See https://discuss.elastic.co/t/monitoring-api-value-is-updated-only-after-logstash-restart/100971

- Version: 5.5.2
- Operating System: RHEL 7

coolprag:
hi...any updates on this? Another customer is complaining about the persistent event count not getting updated until logstash is restarted. Is there a work around for this?

jordansissel:
I tried to reproduce this, but it works for me:

```
% while true; do ; date; curl -s localhost:9600/_node/stats/pipeline?pretty | jq .pipeline.queue.events; sleep 1; done
Fri Oct  6 13:48:42 PDT 2017
2995072
Fri Oct  6 13:48:43 PDT 2017
2995072
Fri Oct  6 13:48:44 PDT 2017
2995072
Fri Oct  6 13:48:45 PDT 2017
2995072
Fri Oct  6 13:48:46 PDT 2017
2995072
Fri Oct  6 13:48:47 PDT 2017
3225442
Fri Oct  6 13:48:48 PDT 2017
3225442
Fri Oct  6 13:48:49 PDT 2017
3225442
Fri Oct  6 13:48:50 PDT 2017
3225442
Fri Oct  6 13:48:51 PDT 2017
3454222
Fri Oct  6 13:48:52 PDT 2017
3454222
Fri Oct  6 13:48:53 PDT 2017
3454222
Fri Oct  6 13:48:54 PDT 2017
3454222
Fri Oct  6 13:48:56 PDT 2017
3454222
Fri Oct  6 13:48:57 PDT 2017
3681497
Fri Oct  6 13:48:58 PDT 2017
3681497
```

It appears to not "increment" every time I query, but it is incrementing. Every 5-10 seconds the value is updated with a new value.

jakelandis:
hmm... i was able to reproduce this with the pipe and stdin input, but not the TCP input. 

I am taking a deeper look.

dandrestor:
Just a wild guess, but perhaps it happens when more than just one input plugin is used? @jordansissel What configuration did you test with?

jakelandis:
OK.... figured it out. It only happens when you have XPack installed, and even then it is not guaranteed. 

[...]

PR #8472, opened 2017-10-10
PQ/DLQ Monitoring: Fix X-Pack and multi-pipeline monitoring

jakelandis:
When X-Pack is installed an internal (hidden) pipeline named .monitoring-logstash is also running. This means that for X-Pack is there N +1 pipelines where N = user defined pipelines. The code currently only supports monitoring 1 pipeline, and it is possible that the internal pipeline gets picked up instead of the intended pipeline.

When multiple user pipelines are enabled (6.0+), the code still only supports monitoring 1 pipeline.

The change here is to ignore the system pipelines, and monitor as many user defined pipelines that may exist.

Fixes: #8382 

Note - there are no existing tests for this for, and did not add any since one of my next tasks is to move this code into Java (if possible)

### DIFF
--- a/logstash-core/lib/logstash/instrument/periodic_poller/dlq.rb
+++ b/logstash-core/lib/logstash/instrument/periodic_poller/dlq.rb
@@ -10,10 +10,15 @@ module LogStash module Instrument module PeriodicPoller
     end
 
     def collect
-      _, pipeline = @agent.with_running_pipelines { |pipelines| pipelines.first }
-      unless pipeline.nil?
-        pipeline.collect_dlq_stats
+      pipelines = @agent.with_running_user_defined_pipelines {|pipelines| pipelines}
+      unless pipelines.nil?
+        pipelines.each {|_, pipeline|
+          unless pipeline.nil?
+            pipeline.collect_dlq_stats
+          end
+        }
       end
     end
   end
 end end end
+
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/instrument/periodic_poller/dlq.rb
```diff
@@ -10,10 +10,15 @@ def initialize(metric, agent, options = {})
     end
 
     def collect
-      _, pipeline = @agent.with_running_pipelines { |pipelines| pipelines.first }
-      unless pipeline.nil?
-        pipeline.collect_dlq_stats
+      pipelines = @agent.with_running_user_defined_pipelines {|pipelines| pipelines}
+      unless pipelines.nil?
```
robbavey: btw, you could write the pipelines.each block as a single line (if you need the nil? check at all)

```
pipelines.each { |_, pipeline| pipeline.collect_dlq_stats unless pipeline.nil? }
```

And I realize I didn't do that in the initial commit ;)

jakelandis:
@robbavey - Thanks!

Will merge back to 5.6.4

jakelandis:

Branch | Commits
-------|--------
5.6 | 61638a59cd1dfce60f6fd62abc05741b49f1f9b0

PR #8475, opened 2017-10-11
upgrade JrJackson to 0.4.4 to fix #8439

### LINKED ISSUE #8439: Null JSON with JSON filter can halt a pipeline
This is a regression from 5.6, and is quite an odd bug. 

The reproduction case is simple (but was discovered in non-trivial configuration). 
```
bin/logstash -e 'input { generator {  message => ""  } } filter { json {   source => "message" }} output { stdout { codec => rubydebug } }'
```

When run against 5.6 the generator keeps chugging along with alot of "_jsonparsefailure" tags...but when run against 6.0+ it halts the pipeline with some very long exception traces with `java.lang.NullPointerException` seemingly the main culprit. 

I will not include those here for brevity since it is easy to reproduce. If looking at the stack traces, slightly more readable stack traces can be obtained by turning off invokedynamic (however, that flag seems unrelated to the underlying bug). 

I confirmed that the same version of json filter was in use, as well confirmed this is an issue on a very early version of 6.0 before the Jackson upgrade. 

Here is where it get weird....
In logstash/json.rb (which hasn't been changed for a while), the following code illustrates the bug:
```
 result = JrJackson::Ruby.parse(data, options)
 return result
``` 
...and the following _fixes_ it
```
result = JrJackson::Ruby.parse(data, options)
puts result
return result
```

This is a Grade A [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug).

colinsurprenant:
Fixes #8439 - Null JSON with JSON filter can halt a pipeline
See guyboertje/jrjackson/pull/65

### DIFF
--- a/versions.yml
+++ b/versions.yml
@@ -20,5 +20,5 @@ jruby:
 # Note: this file is copied to the root of logstash-core because its gemspec needs it when
 #       bundler evaluates the gemspec via bin/logstash
 # Ensure Jackson version here is kept in sync with version used by jrjackson gem
-jrjackson: 0.4.3
+jrjackson: 0.4.4
 jackson: 2.9.1


jakelandis:
LGTM 

Thanks again @guyboertje for this JrJackson release and the simplification of the versions.

colinsurprenant:
merged in 6.x and master. 
Unsure if we should merge into 6.0 and 5.6?

PR #8476, opened 2017-10-11
fix the test:plugins execution

colinsurprenant:
fix for the `rake test:plugins` command to run all installed plugins specs. 

This command is really useful to test core changes on all installed plugins in one go.

### DIFF
--- a/rakelib/test.rake
+++ b/rakelib/test.rake
@@ -39,7 +39,7 @@ namespace "test" do
   end
   
   desc "run all installed plugins specs"
-  task "plugins" do
+  task "plugins"  => "bootstrap" do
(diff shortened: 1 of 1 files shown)


original-brownbear:
@colinsurprenant LGTM :) not sure this command will actually work for all plugins (seems we've got quite a few lite Kafka output that now require some more manual steps before running specs), but that's a different issue.

original-brownbear:
Ah you beat me to it, didn't see #8477 => ignore what I said except for the LGTM :)

colinsurprenant:
merged into 5.6, 6.0, 6.x and  master.

PR #8482, opened 2017-10-12
Modules: Set default credentials for Kibana if es info is present

### LINKED ISSUE #8422: Set default URL+credentials for Kibana if ES info is present
When loading the dashboards from Logstash modules in 5.6 and 6.0, the user needs to configure:

Elasticsearch URL
Kibana URL
Elasticsearch credentials
Kibana credentials

Beats introduced logic whereby if user supplies ES URL + credentials but no Kibana credentials, we automatically try to use ES info for both ES and Kibana: https://github.com/elastic/beats/pull/4867

This allows for getting started instructions to be simpler.

cc: @acchen97 @robbavey

robbavey:
If modules configuration has elasticsearch credentials set, but not
kibana, attempt to use elasticsearch credentials for Kibana login.

Fixes #8422

### DIFF
--- a/logstash-core/lib/logstash/config/modules_common.rb
+++ b/logstash-core/lib/logstash/config/modules_common.rb
@@ -65,6 +65,7 @@ module LogStash module Config
           pipeline_id = alt_name
           module_settings.set("pipeline.id", pipeline_id)
           LogStash::Modules::SettingsMerger.merge_cloud_settings(module_hash, module_settings)
+          LogStash::Modules::SettingsMerger.merge_kibana_auth!(module_hash)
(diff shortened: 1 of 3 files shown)


robbavey:
@guyboertje This commit only deals with Elasticsearch/Kibana credentials, and not the URL, as is being discussed in #8422

robbavey:
@guyboertje Decision was made in #8422 to only deal with Elasticsearch and Kibana credentials, and to default the kibana URL to localhost:5601 if the value isn't set, which I believe is the current default anyway, so this is ready for review

PR #8490, opened 2017-10-13
Update docs about using Logstash with Filebeat Modules

dedemorton:
Now that Filebeat modules allow you to configure LS as the output, the steps need to be updated.

Summary of changes:

* removed Filebeat config examples because they are no longer necessary now that users can run Filebeat modules and configure Filebeat to send to Logstash.
* Consolidated the examples for each module type to include all log types (to show conditional processing and simplify examples).
* Reworked the steps to mirror the task flow that we're using other places in the Filbeat docs where we describe setting up and running modules.

I left `document_type` in for now because I plan to cleanup the docs in a single PR to remove references to document_type. 

This PR updates the existing docs, but we still need to add examples for other Filebeat modules. Specifically, auditd, icinga, and redis. I might be able to use the ingest converter to get the config for icinga, but the auditd and redis modules use painless, so I need help translating those configs. I'd like to get the existing content reviewed and merged and then open another PR for the remaining modules.


jordansissel:
@dedemorton Not sure if you want this merged into 5.6, but please do so if you need this in 5.6

dedemorton:
No...let's not merge it into 5.6. The topic shows FB module commands that are 6.0-specific.

PR #8492, opened 2017-10-13
Fix 'https' scheme in ES hosts setting when set from Cloud ID.

guyboertje:
This is a blocker for the combination of x-pack feature config management and or monitoring when using cloud id or specifying host(s) urls in logstash.yml

I did get 2 wait timeout converge_spec.rb failures on my laptop - only when running rake test:core though, the individual spec file passed.

Jenkins tests pass.

### DIFF
--- a/logstash-core/spec/logstash/util/cloud_setting_id_spec.rb
+++ b/logstash-core/spec/logstash/util/cloud_setting_id_spec.rb
@@ -1,6 +1,6 @@
 # encoding: utf-8
-require "logstash/util/cloud_setting_id"
 require "spec_helper"
+require "logstash/util/cloud_setting_id"
 
 describe LogStash::Util::CloudSettingId do
   let(:input) { "foobar:dXMtZWFzdC0xLmF3cy5mb3VuZC5pbyRub3RhcmVhbCRpZGVudGlmaWVy" }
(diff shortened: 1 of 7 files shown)


tbragin:
@guyboertje @acchen97 @andrewvc @jordansissel Just checking to see if this is still on track to get resolved in RC2?

logstash-core/spec/logstash/util/cloud_setting_id_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    it '#kibana_host has a value' do
+      expect(subject.kibana_host).to eq("identifier.us-east-1.aws.found.io:443")
+    end
+    it '#kibana_scheme has a value' do
+      expect(subject.kibana_scheme).to eq("https")
+    end
+    it '#to_s has a value of #decoded' do
+      expect(subject.to_s).to eq(subject.decoded)
+    end
+  end
```
andrewvc: These specs all seem to cover unacceptable input. What about acceptable input?

guyboertje: line 67 should read `describe "when given acceptable input, the accessors:" do`, sorry - will fix.

guyboertje:
@andrewvc 
How about now?
I added a kibana_client test too.

guyboertje:
yay

Issue #8423, opened 2017-10-03
Missing information about the data directory


ph:
The [dir-layout](https://www.elastic.co/guide/en/logstash/6.0/dir-layout.html) page is missing information about the `data` directory required for the persistent queue and also where the persistent id file is saved.

We need to update the information 
 of this file https://github.com/elastic/logstash/blob/master/docs/static/setting-up-logstash.asciidoc

# Tar/Zip
{extract.path}/data

# debian / Rpm
/var/lib/logstash

dedemorton:
Closing because this is fixed in #8495

PR #8495, opened 2017-10-15
[DOCS] Add data directory info to setting up logstash doc

tkbky:
Data directory info is missing in setting-up-logstash.asciidoc as reported in #8423. This fixes that.


jordansissel:
LGTM.

PR #8500, opened 2017-10-18
[Performance] (Java exec beats Ruby exec inside!) Faster Java Execution Using CodeGen/Janino

original-brownbear:
Will have to document this better so WIP for now, functionally this step should be fine though.

### Changes

* Compile `Dataset`s for outputs in a more efficient optimal fashion
  * Cache `Callsite` for `multi_receive` instead of implementing Java interface from Ruby (this has quite a bit of overhead)
   * Don't loop over `Dataset`, unroll parent `Dataset` iteration at compile time
   * Monomorphic callsites only, use runtime compilation to either cast each field in `OutputDataset` to its type or at least make sure that there is only one type for each `invokevirtual` call to begin with (in the case of runtime compiled `Dataset` as explained in the Javadoc)
* Avoid instantiating a `TerminalDataset` if we only have a single output
* Compile an iteration free version of `TerminalDataset` if we have multiple outputs
* Pass `RubyArray` instead of `Batch` from Ruby to Java to avoid the redundant `to_a` Java -> Ruby -> Java round trip since we only use the contents of a batch and not the `Batch` instance itself on the Java side (also makes instantiating read batches cheaper)
 
### Some general benchmark results

* Apache Example: Improves significantly relative to the previous Java exec and is measurably faster than Ruby exec now (58k/s vs. 52k/s over largeish sample size)
* Baseline (4 generator threads, 1 worker thread): **biggest improvement here** as expected since only the outputs are optimized using code gen for now, `master` Java exec is at around 350k/s, Ruby is at 360k/s and this branch at 580k/s for Java exec! Workers are a lot faster and less memory hungry now.
* Baseline (1 generator, 1 worker): This case is interesting, it actually gets slower with the Java exec. The reason for this is that the performance of the in-memory queue degrades as soon as outputs start to outpace inputs. This is simply a result of an empty queue likely not filling a batch on the initial `drainTo` call => you need to poll the Queue => you get contention on its lock => writers to the queue wait more often and get throttled. Not much to do here, this is simply the way an `ArrayblockingQueue` works (google "false sharing" for the why behind a faster consumer making things slower here overall) and would require some other data structure that comes with more coarse locking on writes. For now, this is a pretty academic example anyways :)
* Optimal worker count and batch size both are higher for Java execution than for Ruby => this is kind of nice (imo), especially when looking at ES output bulk request sizes

Data:

```sh
➜  logstash git:(faster-java-exec-filters) time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -w 4 -b 128 -f ~/tmp/logstash.cfg --experimental-java-execution | pv | wc -c
13.2MiB 0:04:16 [52.5KiB/s] [                                                                                                                            <=>                 ]
 13801638
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.11s user 2.53s system 1% cpu 4:15.81 total
bin/logstash -w 4 -b 128 -f ~/tmp/logstash.cfg --experimental-java-execution  1080.48s user 25.05s system 430% cpu 4:16.59 total
pv  2.95s user 18.76s system 8% cpu 4:16.59 total
wc -c  0.37s user 3.08s system 1% cpu 4:16.59 total
➜  logstash git:(faster-java-exec-filters) time cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs | bin/logstash -w 4 -b 128 -f ~/tmp/logstash.cfg | pv | wc -c          
13.2MiB 0:04:46 [47.1KiB/s] [                                                                      <=>                                                                       ]
 13801638
cat ~/Downloads/apache_access_logs ~/Downloads/apache_access_logs  0.11s user 2.20s system 0% cpu 4:45.78 total
bin/logstash -w 4 -b 128 -f ~/tmp/logstash.cfg  1202.31s user 24.85s system 428% cpu 4:46.39 total
pv  3.02s user 19.37s system 7% cpu 4:46.39 total
wc -c  0.38s user 3.18s system 1% cpu 4:46.39 total
```

### Next Steps (imo)

* Do the same thing I did here and implement more optimal runtime compiled for filter and conditional `Dataset`
* Extract a common abstraction for describing the inputs to compiling a `Dataset` (like what buffers do I need, what `Callsite`s do I invoke and such)
* Use the abstraction to compile every pipeline into just a single method (and be super cache efficient and fast :))


original-brownbear:
@andrewvc there we go, this should make the Java exec a lot more interesting for people to try out :)

andrewvc:
Very impressive work @original-brownbear !

Reformatting your Apache results for clarity for me (+ others):
```
bin/logstash -w 4 -b 128 -f ~/tmp/logstash.cfg --experimental-java-execution  
1080.48s user 25.05s system 430% cpu 4:16.59 total
bin/logstash -w 4 -b 128 -f ~/tmp/logstash.cfg  
1202.31s user 24.85s system 428% cpu 4:46.39 total

User CPU Reduction: 11%
System CPU: 0%
Wallclock: 11% reduction
```

It's always nice when userspace CPU correlates exactly to wallclock since users don't have to do any extra work to reap that practical benefit!

original-brownbear:
no changes in #83f407, just rebased to fix conflicts

original-brownbear:
@jakelandis reassigning you by orders from @andrewvc . Ping me on Slack if you want a 1:1 intro for this whole topic :)

jakelandis:
@original-brownbear - Do you have an example of the generated source for a basic pipeline ?

original-brownbear:
@jakelandis sure, so for a simple pipeline output snippet (not we only code generate the output part, for now, full stage including filters would be the next step :)) `output { stdout{} }`
You get this (without the comments I added :P):

```java
class CompiledDataset3 {
    // multi_receive method call site extracted from the OutputDelegator instance
    private final org.jruby.internal.runtime.methods.MixedModeIRMethod field0;
    // arguments array for calling that callsite, since it requires its args to be wrapped in an array, 
    //  the Callsite API doesn't take varargs 
    private final org.jruby.runtime.builtin.IRubyObject[] field1;
    // The OutputDelegator instance
    private final org.jruby.RubyObjectVar9 field2;

    public CompiledDataset3(Object field0, Object field1, Object field2) {
        this.field0 = (org.jruby.internal.runtime.methods.MixedModeIRMethod) field0;
        this.field1 = (org.jruby.runtime.builtin.IRubyObject[]) field1;
        this.field2 = (org.jruby.RubyObjectVar9) field2;
    }

    public Collection compute(RubyArray batch, boolean flush, boolean shutdown) {
        field1[0] = batch;
        field0.call(RubyUtil.RUBY.getCurrentContext(), field2, RubyUtil.LOGSTASH_MODULE,
            "multi_receive", field1, Block.NULL_BLOCK
        );
        return null;
    }

    public void clear() {
    }
}
```

jakelandis:
I scanned the code and admittedly don't fully understand the changes presented here....however, I don't think at this stage it is necessary. 

* the changes are behind a feature flag
* this is part of a much larger change set 
* there may be refactoring of this in newer iterations
* POC/WIP to integrate early shouldn't need line by line reviews. 

In general this design and direction is really cool, and am in full support.  As briefly discussed in Slack, we should create review milestones once the final iterations of the whole picture are landing. 

In the meantime LGTM, lets gets this merged to master.

original-brownbear:
@jakelandis thanks!

PR #8508, opened 2017-10-22
make -V/--version fast on Windows

### LINKED ISSUE #5526: make --version faster by not launch logstash
This has been done already for *nix systems in https://github.com/elastic/logstash/issues/5407
in commit https://github.com/elastic/logstash/commit/72b277404b2da8586e23327e1659c06db69e274b but now this needs to be converted to windows BATCH script

czlowieq:
Do not run logstash when checking version on windows

Fixes #5526

### DIFF
--- a/bin/logstash.bat
+++ b/bin/logstash.bat
@@ -10,6 +10,9 @@ if errorlevel 1 (
 	exit /B %ERRORLEVEL%
 )
 
+if "%1" == "-V" goto version
+if "%1" == "--version" goto version
+
 rem iterate over the command line args and look for the argument
 rem after --path.settings to see if the jvm.options file is in
 rem that path and set LS_JVM_OPTIONS_CONFIG accordingly
@@ -51,12 +54,38 @@ for %%i in ("%LS_HOME%\logstash-core\lib\jars\*.jar") do (
 
 %JAVA% %JAVA_OPTS% -cp %CLASSPATH% org.logstash.Logstash %*
 
-endlocal
+goto :end
 
-goto :eof
+:version
+set "LOGSTASH_VERSION_FILE1=%LS_HOME%\logstash-core\versions-gem-copy.yml"
+set "LOGSTASH_VERSION_FILE2=%LS_HOME%\versions.yml"
+
+set "LOGSTASH_VERSION=Version not detected"
+if exist !LOGSTASH_VERSION_FILE1! (
+	rem this file is present in zip, deb and rpm artifacts and after bundle install
+	rem but might not be for a git checkout type install
+	for /F "tokens=1,2 delims=: " %%a in (!LOGSTASH_VERSION_FILE1!) do (
+		if "%%a"=="logstash" set LOGSTASH_VERSION=%%b
+	)
+) else (
+	if exist !LOGSTASH_VERSION_FILE2! (
+		rem this file is present for a git checkout type install
+		rem but its not in zip, deb and rpm artifacts (and in integration tests)
+		for /F "tokens=1,2 delims=: " %%a in (!LOGSTASH_VERSION_FILE2!) do (
+			if "%%a"=="logstash" set LOGSTASH_VERSION=%%b
+		)
+	)
+)
+echo logstash !LOGSTASH_VERSION!
+goto :end
+
 :concat
 IF not defined CLASSPATH (
   set CLASSPATH="%~1"
 ) ELSE (
   set CLASSPATH=%CLASSPATH%;"%~1"
 )
+goto :eof
+
+:end
+endlocal


original-brownbear:
Jenkins test this please

czlowieq:
I've seen some windows tests in `logstash/qa/scripts/windows/acceptance/logstash_release_acceptance.ps1`. Should I put this test(s) in there too? I've noticed that there is no version number in environment variables, which could make it tricky.

*nix version tests use rspec, so there is the const containing the version. But I'm not sure if it is possible to test the batch file like this too.

original-brownbear:
@jakelandis @colinsurprenant I think the above is a question for one of you guys :)

jordansissel:
I would like tests added for this. The proposal is simple (shortcut showing the verson) but the implementation is complex and fragile (nobody on the team is an expert in DOS batch, so this may break without us realizing it).

bin/logstash.bat
```diff
@@ -47,5 +50,29 @@ set JAVA_OPTS=%LS_JAVA_OPTS%
 
 rem jruby launcher will pickup JAVA_OPTS set above to set the JVM options before launching jruby
 %JRUBY_BIN% "%LS_HOME%\lib\bootstrap\environment.rb" "logstash\runner.rb" %*
+goto end
+
+:version
+set "LOGSTASH_VERSION_FILE1=%LOGSTASH_HOME%/logstash-core/versions-gem-copy.yml"
```
yaauie: should this be `LS_HOME` instead of `LOGSTASH_HOME`? This is the first reference to `LOGSTASH_HOME` I see in this file, but `LS_HOME` is referenced several times.

czlowieq:
@jordansissel I totally agree that it needs to be tested. I'm not sure how to do it though. It seems that there are tests in PowerShell script, but there is no version number variable, as stated above. Putting the same logic to extract the version number from the same files as in batch is doable, but I'm not sure if it makes sense.

Unless it should be tested in rspec (not sure if it's suitable for rspec tests).

andrewvc:
@czlowieq I think rspec tests are fine for this case. Have you seen the tests in https://github.com/elastic/logstash/tree/master/qa/scripts/windows ?

andrewvc:
Blah, I misspoke, those are all powershell, but that should be fine :)

czlowieq:
@andrewvc Yeah, seen those. But still, I'm not sure if there is a version number I can easily get in those scripts, so I can compare it with the command result. Any ideas?

colinsurprenant:
what this would require is integration testing and we don't have a good story for that *yet* on Windows.  Personally I am ok if we end up merging this without test **IIF** it has been throughly manually tested and we have other tests for the `versions.yml` file & content. 

I will go ahead and review this shortly. Let me know if anyone has concerns in moving forward like this.

bin/logstash.bat
czlowieq: Thanks. Fixed.

bin/logstash.bat
```diff
@@ -47,5 +50,34 @@ set JAVA_OPTS=%LS_JAVA_OPTS%
...
 rem jruby launcher will pickup JAVA_OPTS set above to set the JVM options before launching jruby
 %JRUBY_BIN% "%LS_HOME%\lib\bootstrap\environment.rb" "logstash\runner.rb" %*
+goto end
+
+:version
+rem make the script work without LS_HOME variable set
+if "%LS_HOME%" == "" (
+	for %%a in ("%~dp0..") do set "LS_HOME=%%~fa"
+)
+
```
colinsurprenant: this `LS_HOME` handling is no required since the first thing this bat file does is to call `setup.bat` which handle the setup of `LS_HOME` see https://github.com/elastic/logstash/blob/40827a53eb6767dc019c926bd5e5bb5d81c1f246/bin/setup.bat#L5-L19

bin/logstash.bat
```diff
@@ -47,5 +50,34 @@ set JAVA_OPTS=%LS_JAVA_OPTS%
...
+) else (
+	if exist !LOGSTASH_VERSION_FILE2! (
+		rem this file is present for a git checkout type install
+		rem but its not in zip, deb and rpm artifacts (and in integration tests)
+		for /F "tokens=1,2 delims=: " %%a in (!LOGSTASH_VERSION_FILE2!) do (
+			if "%%a"=="logstash" set LOGSTASH_VERSION=%%b
+		)
+	)
+)	
+echo "logstash !LOGSTASH_VERSION!"
```
colinsurprenant: ++ this seems consistent with the bash version logic. will manually test shortly. 👍

colinsurprenant:
@czlowieq good stuff. I left a few comments - please check about the `LS_HOME` handling. Other than that I will perform manual test and we should be good to go.

colinsurprenant:
@czlowieq could you please rebase and resolve conflicts? sorry about that, once done I will prioritize testing so we can go ahead and merge. Thanks.

czlowieq:
@colinsurprenant Should I squash the commits into one or leave them as is?

colinsurprenant:
@czlowieq sure - it will have to be done anyway before merging so ...

colinsurprenant:
Jenkins please test this

bin/logstash.bat
```diff
@@ -51,12 +54,38 @@ for %%i in ("%LS_HOME%\logstash-core\lib\jars\*.jar") do (
...
+) else (
+	if exist !LOGSTASH_VERSION_FILE2! (
+		rem this file is present for a git checkout type install
+		rem but its not in zip, deb and rpm artifacts (and in integration tests)
+		for /F "tokens=1,2 delims=: " %%a in (!LOGSTASH_VERSION_FILE2!) do (
+			if "%%a"=="logstash" set LOGSTASH_VERSION=%%b
+		)
+	)
+)
+echo "logstash !LOGSTASH_VERSION!"
```
colinsurprenant: this outputs the following:
```
C:\Users\colin\dev\logstash (pr/8508)
λ bin\logstash -V
"logstash 7.0.0-alpha1"

C:\Users\colin\dev\logstash (pr/8508)
λ bin\logstash --version
"logstash 7.0.0-alpha1"
```

To be consistent with the Linux behaviour we should remote the quotes from the output, it should be 
```
echo logstash !LOGSTASH_VERSION!
```

colinsurprenant:
@czlowieq  while manually testing, I noticed that the version string is quoted which is different from the Linux output. Please correct this, then we should be good.
Thanks for your patience.

ppawlikmb:
@colinsurprenant Thanks. Fixed it.

colinsurprenant:
LGTM - merging!

colinsurprenant:
Merged into master and 6.x
Thanks!

PR #8510, opened 2017-10-23
MINOR: Remove Duplicate Ruby Runtime in Java Tests

original-brownbear:
This base test class was a leftover from before `RubyUtil` holding a global reference to a properly set up `Ruby` runtime instance.

=> cleaned up usages of it to also use the global runtime so that we don't work from a different runtime in these tests.


original-brownbear:
@andrewvc thanks!

PR #8511, opened 2017-10-23
MINOR: Clean up Logstash Module RubyClass Setup

original-brownbear:
We shouldn't get the Logstash module redundantly all over the place, we know this is a global constant since we always work from just a single `Ruby` runtime instance per JVM.


original-brownbear:
Also, this is part of #8500 since it makes code gen easier would just be nice to merge this separately if possible :)

jakelandis:
LGTM

original-brownbear:
@jakelandis thanks!

PR #8513, opened 2017-10-23
MINOR: Remove dead code supporting non-java platforms

original-brownbear:
I think we can remove all of this code, it's all just branching around the platform and with the current code requiring `jar_dependencies` in `lib/bootstrap/environment.rb` there is no way to get to any off the removed on a non JRuby platform imo right?


original-brownbear:
np :)

PR #8515, opened 2017-10-23
Fix a typo in rvm install instructions in README

roidelapluie:


jakelandis:
@roidelapluie - The backslash is intentional it is syntax borrowed directly from [RVM](https://rvm.io/) and used to [avoid calling any possible aliases](https://stackoverflow.com/questions/15951772/curl-bash-whats-the-slash-for).

Issue #8442, opened 2017-10-06
Add Stronger Tests to Java Pipeline Execution


original-brownbear:
We want JUnit tests here alongside the Ruby specs.

This will require a bit of work to load Ruby plugins from the Java end, but it's doable in a straightforward way.

original-brownbear:
link https://github.com/elastic/logstash/pull/8357

original-brownbear:
link https://github.com/elastic/logstash/pull/8517

original-brownbear:
closing this, we now have a number of JUnit end to end execution tests and a relatively clean framework for adding more.

PR #8517, opened 2017-10-24
BUILD: Execute RSPEC from JUnit

original-brownbear:
This makes the RSpecs run from JUnit to:

* Move more logic to Java :P 
* Speed up the test run (saves about a minute compared to `rake test:core` for me)
* Allow for debugging of Java code in Ruby Specs from the IDE
* Allows for moving some test only Java code to the test classpath. Currently, the in memory-serialized queue is the biggest offender here, but there are a few others that needlessly complicate our production code.
* **Most importantly:** this gives us a starting point for loading Logstash's infrastructure for JUnit tests for #8442 

This shouldn't change any anything in what the individual `rake` tasks do since the `slow` goal was the only one to run Ruby + Java.
This is still the case and I added a new task that can be invoked by the `core-java` goal and doesn't run the RSpec JUnit invoker.

### DIFF
--- a/lib/bootstrap/rspec.rb
+++ b/lib/bootstrap/rspec.rb
@@ -10,5 +10,8 @@ require "rspec/core"
 require "rspec"
 require 'ci/reporter/rake/rspec_loader'
 
-status = RSpec::Core::Runner.run(ARGV.empty? ? ["spec"] : ARGV).to_i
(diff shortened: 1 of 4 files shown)


jordansissel:
> org.logstash.RSpecTests > rspecTests PASSED

Cool beans! Happy to have `./gradlew test` be all I need to invoke tests. :)

LGTM

original-brownbear:
@jordansissel thanks!

PR #8520, opened 2017-10-24
CLEANUP: Refactor plugin setup

original-brownbear:
Big step (~80% of the way imo) to enable better testing of the (Java-)pipeline execution.

Extracted all logic necessary for building plugins from the main pipeline as is for now and wrapped it in the same Java interface that wrapped the pipeline itself before.
The interface only forwarded methods for setting up plugins to the Java side anyways.

=> with this one I can simply write a Java mock pipeline factory (or more importantly just mock out parts of it and still set up actual Ruby plugins, that's why I split the logic into a few "factories" :)) without having to set up an actual `Pipeline`.

=> Also it's a generally nice to have cleanup, drying up the plugin setup between Java and Ruby execution and shortening the pipeline classes :)

For #8442 :)

### DIFF
--- a/logstash-core/lib/logstash/java_pipeline.rb
+++ b/logstash-core/lib/logstash/java_pipeline.rb
@@ -32,7 +32,6 @@ java_import org.logstash.config.ir.CompiledPipeline
 java_import org.logstash.config.ir.ConfigCompiler
 
 module LogStash; class JavaBasePipeline
-  include org.logstash.config.ir.compiler.RubyIntegration::Pipeline 
   include LogStash::Util::Loggable
 
   attr_reader :settings, :config_str, :config_hash, :inputs, :filters, :outputs, :pipeline_id, :lir, :execution_context, :ephemeral_id
@@ -52,19 +51,16 @@ module LogStash; class JavaBasePipeline
       @config_str, @settings.get_value("config.support_escapes")
     )
 
-    # Every time #plugin is invoked this is incremented to give each plugin
-    # a unique id when auto-generating plugin ids
-    @plugin_counter ||= 0
-
     @pipeline_id = @settings.get_value("pipeline.id") || self.object_id
-
-    # A list of plugins indexed by id
-    @plugins_by_id = {}
     @agent = agent
-
     @dlq_writer = dlq_writer
-
-    @lir_execution = CompiledPipeline.new(@lir, self)
+    @plugin_factory = LogStash::Plugins::PluginFactory.new(
+      # use NullMetric if called in the BasePipeline context otherwise use the @metric value
+      @lir, LogStash::Plugins::PluginMetricFactory.new(pipeline_id, @metric || Instrument::NullMetric.new),
+      @logger, LogStash::Plugins::ExecutionContextFactory.new(@agent, self, @dlq_writer),
+      JavaFilterDelegator
+    )
+    @lir_execution = CompiledPipeline.new(@lir, @plugin_factory)
     if settings.get_value("config.debug") && @logger.debug?
       @logger.debug("Compiled pipeline code", default_logging_keys(:code => @lir.get_graph.to_string))
     end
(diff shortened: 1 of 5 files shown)


original-brownbear:
@andrewvc with this one and an easier way to set up `PipelineIR` from Java (incoming, later today or tomorrow :)) we should be good to write whatever JUnit tests we want for the Java execution :)

logstash-core/lib/logstash/java_pipeline.rb
```diff
@@ -92,71 +88,8 @@ def compile_lir
...
-  def buildInput(name, line, column, *args)
-    plugin("input", name, line, column, *args)
-  end
-
-  def buildCodec(name, *args)
-   plugin("codec", name, 0, 0, *args)
-  end
-
   def plugin(plugin_type, name, line, column, *args)
-    @plugin_counter += 1
```
jakelandis: was this a dead variable ?

logstash-core/lib/logstash/plugins/plugin_factory.rb
```diff
@@ -0,0 +1,102 @@
...
+              v.source_with_metadata.line == line &&
+              v.source_with_metadata.column == column
+          end.findFirst.get.id
+        end
+        args["id"] = id # some code pulls the id out of the args
+
+        unless id
+          raise ConfigurationError, "Could not determine ID for #{plugin_type}/#{plugin_name}"
+        end
+        raise ConfigurationError, "Two plugins have the id '#{id}', please fix this conflict" if @plugins_by_id[id]
```
jakelandis: nit: use same style for the checks here. e.g.
` raise ConfigurationError, ".." unless id`

logstash-core/lib/logstash/plugins/plugin_factory.rb
```diff
@@ -0,0 +1,102 @@
...
+      def create(plugin_type)
+        @metric.namespace([:stats, :pipelines, @pipeline_id, :plugins])
+          .namespace("#{plugin_type}s".to_sym)
+      end
+    end
+
+    class PluginFactory
+      include org.logstash.config.ir.compiler.RubyIntegration::PluginFactory
+
+      def initialize(lir, metric_factory, logger, exec_factory, filter_class)
```
jakelandis: any specific reason to pass the filter_class, but not the output_class ?

logstash-core/lib/logstash/plugins/plugin_factory.rb
```diff
@@ -0,0 +1,102 @@
...
+        # Scope plugins of type 'input' to 'inputs'
+        type_scoped_metric = @metric_factory.create(plugin_type)
+        klass = Plugin.lookup(plugin_type, name)
+        execution_context = @exec_factory.create(id, klass.config_name)
+
+        if plugin_type == "output"
+          OutputDelegator.new(@logger, klass, type_scoped_metric, execution_context, OutputDelegatorStrategyRegistry.instance, args)
+        elsif plugin_type == "filter"
+          @filter_class.new(@logger, klass, type_scoped_metric, execution_context, args)
+        else # input
```
jakelandis: while you'r refactoring, can we change the comment and variable name to correctly reflect input or codec plugin ?

logstash-core/lib/logstash/plugins/plugin_factory.rb
original-brownbear: @jakelandis first off, sorry for completely missing that you reviewed this one almost 2 weeks ago. No clue why Github didn't send me an email here :(

As for the question:

The reason for this is that we have two different filter delegator classes for the Ruby and Java exec respectively. For the output delegator Ruby and Java exec use the same code => we must pass the wrapper type to use to the factory.

logstash-core/lib/logstash/java_pipeline.rb
original-brownbear: @jakelandis because there's no one reading this variable (as seen by the fact that removing it doesn't lead to issues).

logstash-core/lib/logstash/plugins/plugin_factory.rb
original-brownbear: @jakelandis makes sense, fixing :)

logstash-core/lib/logstash/plugins/plugin_factory.rb
original-brownbear: @jakelandis makes sense, fixing :)

original-brownbear:
@jakelandis sorry for the delay, all points addressed I think :)

jakelandis:
LGTM

original-brownbear:
@jakelandis thanks :)

original-brownbear:
last push is just a squash, no changes :)

PR #8528, opened 2017-10-25
Clean up es data and log folders after ITs

robbavey:
Change the data and log directories to be in /tmp/ls_integrations
(matching the kafka service folder structure), and delete
them in the teardown script.


jordansissel:
I would prefer the paths be randomized but don't need to block this PR on that. LGTM, haven't tested.

robbavey:
@jordansissel Completely missed your original suggestion to randomize the path - I used the same path as in the kafka ITs. Happy to update these in a followup PR

PR #8529, opened 2017-10-25
 #8154 break cyclic dependency between Ruby and Java Event classes

### LINKED ISSUE #8154: We an annoying cyclic dependency (Ruby <-> Java) because of `LogStash::Json` and `org.logstash.ext.JrubyEventExtLibrary#load`

original-brownbear:
Fixes #8154 by explicitly setting up the Ruby classes (`Logstash::Error`, `Logstash::Json:: ParserError` and `Logstash::Json::GeneratorError`) when loading `RubyUtil`.

=> correctness: Java library becomes self-contained
=> we can load the actual `RubyClass` for `RubyEvent` into the global `Ruby` runtime for JUnit tests

* I know `org.logstash.ext.JrubyEventExtLibrary` is redundant now, but I cannot unwrap this class without breaking plugins that use the `RubyEvent` class directly (at least `date-filter` breaks when I do that).


original-brownbear:
@andrewvc one more issue for JUnit down :)

logstash-core/src/main/java/JrubyEventExtService.java
```diff
@@ -1,11 +0,0 @@
-import org.jruby.Ruby;
-import org.jruby.runtime.load.BasicLibraryService;
-import org.logstash.ext.JrubyEventExtLibrary;
-
-public final class JrubyEventExtService implements BasicLibraryService {
```
jakelandis: for my own education what did this do ?

original-brownbear: @jakelandis this was automatically (it scans all classes to find these `LibraryService` instances) invoked by JRuby at some point during its startup (a point later than the classload of `RubyUtil`) to give us the ability to then `require "jruby_event_ext"` (obviously not fun in our case because of the cyclic dependency and us only being able to run this after the `Logstash::Json` module was loaded, more fun if we control the load order manually now :)).

original-brownbear:
@jakelandis thanks!

PR #8532, opened 2017-10-25
Update upgrade guide docs to refer to 6.0

robbavey:
Also include reference to the removal of mapping types

### DIFF
--- a/docs/static/upgrading.asciidoc
+++ b/docs/static/upgrading.asciidoc
@@ -17,7 +17,7 @@ See the following topics for information about upgrading Logstash:
 * <<upgrading-using-package-managers>>
 * <<upgrading-using-direct-download>>
 * <<upgrading-logstash-pqs>>
-* <<upgrading-logstash-5.0>>
+* <<upgrading-logstash-6.0>>
 
 [[upgrading-using-package-managers]]
 === Upgrading Using Package Managers
@@ -58,8 +58,8 @@ Keep in mind that only one Logstash instance can write to `path.queue`. You
 cannot have the original instance and the new instance writing to the queue at
 the same time.
 
-[[upgrading-logstash-5.0]]
-=== Upgrading Logstash to 5.0
+[[upgrading-logstash-6.0]]
+=== Upgrading Logstash to 6.0
 
 Before upgrading Logstash, remember to read the <<breaking-changes,breaking changes>>.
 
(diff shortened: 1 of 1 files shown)


robbavey:
If this is a little 'thin' on details, I'll be happy to take suggestions to flesh it out

dedemorton:
LGTM, though Deb mentioned something about needing to make sure the Logstash updgrade info addresses upgrading index templates. Maybe it would be worth having the team take a quick look at the upgrade docs and verify that they are correct and complete for 6.0.

robbavey:
@andrewvc @jsvd Do you have anything to add to these upgrade documents, particularly with reference to the Elasticsearch changes in 6.0?

jsvd:
humm in beats there's a documentation section about dealing with the 6.0 ES templates https://www.elastic.co/guide/en/beats/libbeat/6.x/upgrading-5-to-6.html#upgrading-to-5.6

We could have a similar section describing the situation that, if you are using the ES output and migrate ES to 6.0 and want to use the default templates, then you need to overwrite the es6.x template, starting a pipeline with the [`overwrite_template => true`](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-template_overwrite) option in the es output.

robbavey:
@jsvd @dedemorton I've updated to include some text on overwriting default templates

docs/static/upgrading.asciidoc
```diff
@@ -70,26 +70,27 @@ If you are installing Logstash with other components in the Elastic Stack, also
...
-will not be able to use it and all indexing will fail likely fail. If you use your own, custom template with Logstash,
-then this issue can be ignored.
+Although we make great efforts to ensure compatibility, Logstash 6.0 is not completely backwards compatible. As noted
+in the Elastic Stack upgrade guide, Logstash 6.0 should not be upgraded before Elasticsearch 6.0. This is both
+practical and because some Logstash 6.0 plugins may attempt to use features of Elasticsearch 6.0 that did not exist
+in earlier versions.
 
-Note the Elasticsearch Output Index Template change in the <<breaking-changes>> documentation for further insight into
-this change and how it impacts operations.
+Note that multiple doctypes are no longer supported in Elasticsearch 6.0. Please refer to
```
dedemorton: I didn't notice before that this appears under the heading "When Not to Upgrade."

I'd either move this info up to appear before the section called "When to Upgrade" (so it's part of the intro) or I'd add a new section heading that says something like "Upgrade Considerations" so it's clear this content does not belong under "When Not to Upgrade."

Or you can move this content to appear

docs/static/upgrading.asciidoc
```diff
@@ -70,26 +70,27 @@ If you are installing Logstash with other components in the Elastic Stack, also
...
+ {ref}/removal-of-types.html[Removal of mapping types] and {ref}/breaking-changes.html[Breaking changes] for more information.
+
+To continue using default templates after migrating Elasticsearch to 6.0, the existing template must be overwritten with the 6.x template.
+ This can be done by starting a pipeline with the overwrite_template => true option in the Elasticsearch output definition.
```
dedemorton: Format this as `overwrite_template => true`

docs/static/upgrading.asciidoc
```diff
@@ -70,26 +70,27 @@ If you are installing Logstash with other components in the Elastic Stack, also
...
+in the Elastic Stack upgrade guide, Logstash 6.0 should not be upgraded before Elasticsearch 6.0. This is both
+practical and because some Logstash 6.0 plugins may attempt to use features of Elasticsearch 6.0 that did not exist
+in earlier versions.
 
-Note the Elasticsearch Output Index Template change in the <<breaking-changes>> documentation for further insight into
-this change and how it impacts operations.
+Note that multiple doctypes are no longer supported in Elasticsearch 6.0. Please refer to
+ {ref}/removal-of-types.html[Removal of mapping types] and {ref}/breaking-changes.html[Breaking changes] for more information.
+
+To continue using default templates after migrating Elasticsearch to 6.0, the existing template must be overwritten with the 6.x template.
```
dedemorton: Avoid passive voice and say, "To continue using default templates after migrating Elasticsearch to 6.0, you must overwrite the existing template with the 6.x template. You can do this by starting...."

docs/static/upgrading.asciidoc
```diff
@@ -70,26 +70,27 @@ If you are installing Logstash with other components in the Elastic Stack, also
...
+ This can be done by starting a pipeline with the overwrite_template => true option in the Elasticsearch output definition.
+  If you are using a custom mapping template, you may need to update it in a similar manner to reflect that.
```
dedemorton: Language here is unclear. The phrase "you may need" suggests uncertainty. How can users know whether their custom mapping template needs to be upgraded? And what do you mean by "update it in a similar manner"? We need to add more guidance here, or users are unlikely to know what to do.

docs/static/upgrading.asciidoc
robbavey: @dedemorton Thanks - I keep forgetting this. Previous job doc teams I've worked with used the passive voice for documentation, so I will need a kick every now and then to remind me to do this...

dedemorton: haha...no problem. Passive is OK sometimes, but when we expect users to perform a task, active voice is always best.

robbavey:
@jsvd @dedemorton I've moved the note on handling default template changes, and cleaned up the text as asked, and removed the unclear section on custom mapping templates

docs/static/upgrading.asciidoc
```diff
@@ -58,38 +58,39 @@ Keep in mind that only one Logstash instance can write to `path.queue`. You
...
-will not be able to use it and all indexing will fail likely fail. If you use your own, custom template with Logstash,
-then this issue can be ignored.
+Although we make great efforts to ensure compatibility, Logstash 6.0 is not completely backwards compatible. As noted
+in the Elastic Stack upgrade guide, Logstash 6.0 should not be upgraded before Elasticsearch 6.0. This is both
+practical and because some Logstash 6.0 plugins may attempt to use features of Elasticsearch 6.0 that did not exist
+in earlier versions.
 
-Note the Elasticsearch Output Index Template change in the <<breaking-changes>> documentation for further insight into
-this change and how it impacts operations.
+Note that multiple doctypes are no longer supported in Elasticsearch 6.0. Please refer to
```
dedemorton: Shouldn't this also go at the beginning of the section? The sentence that starts "Note that multiple..."

dedemorton:
Changes LGTM

PR #8533, opened 2017-10-26
MINOR: Remove unused docker-api as development dependency

original-brownbear:
We're not using Docker via Ruby APIs anymore => no need for this dep. :)


original-brownbear:
@jsvd thanks for the reviews!

PR #8538, opened 2017-10-26
MINOR: Cleanup PQ Main Class

original-brownbear:
Just a few trivial passive changes:

* Run all tests against `Queue` and make it `final`. Using `TestQueue` to access `protected` properties of `Queue` has 0 value in terms of encapsulation since you still have to keep the properties `protected`=> `TestQueue` is just noise
* Remove redundant `private` constructor. It was only called by the main constructor => remove it, the class is large enough as it is :)
* Make properties we don't use from other classes `private`
* Remove redundant `nextSeqNum` method, only used in one place and trivial => inline


logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -762,7 +748,7 @@ public void close() throws IOException {
         }
     }
 
-    protected Page firstUnreadPage() {
+    Page firstUnreadPage() {
```
jakelandis: nit: why not private ?

jakelandis:
LGTM

original-brownbear:
@jakelandis thanks!

PR #8542, opened 2017-10-27
fix broken links in docs

jsvd:
these plugins were removed from the plugin documentation so the links here must be removed too

see failure in https://elasticsearch-ci.elastic.co/job/elastic+docs+master+build/7926/console


original-brownbear:
LGTM :)

PR #8554, opened 2017-10-31
MINOR: Fix Readme section on running just the Java tests

original-brownbear:
As of #8517, the Gradle goal for running just the Java tests has changed to `javaTests`


original-brownbear:
thanks guys!

Issue #8220, opened 2017-09-12
Rake bootstrap fails on with JRuby 9.1.13.0


jakelandis:
To reproduce:

```
git clone https://github.com/elastic/logstash.git
(ensure clean environment)
rvm implode
\curl -sSL https://get.rvm.io | bash -s stable
rvm install jruby-9.1.13.0
(new shell)
cd logstash
rvm use jruby-9.1.13.0
rake bootstrap
```

results in 
```
 rake bootstrap
mkdir -p vendor
mkdir vendor/_
Downloading http://jruby.org.s3.amazonaws.com/downloads/9.1.13.0/jruby-bin-9.1.13.0.tar.gz
Installing minitar 0.5.4 because the build process needs it.
mkdir -p build
mkdir -p build/bootstrap
[bootstrap] Fetching and installing gem: minitar (0.5.4)
rake aborted!
Gem::UnsatisfiableDependencyError: Unable to resolve dependency: user requested 'rake (= 12.1.0)'
/private/tmp/logstash/rakelib/gems.rake:39:in `block in (root)'
/private/tmp/logstash/rakelib/gems.rake:16:in `block in (root)'
/private/tmp/logstash/rakelib/dependency.rake:12:in `block in (root)'
/private/tmp/logstash/rakelib/vendor.rake:15:in `block in untar'
/private/tmp/logstash/rakelib/vendor.rake:87:in `block in (root)'
/private/tmp/logstash/rakelib/z_rubycheck.rake:28:in `(root)'
Gem::MissingSpecError: Could not find 'minitar' (= 0.5.4) among 12 total gem(s)
Checked in 'GEM_PATH=/private/tmp/logstash/vendor/bundle/jruby/2.3.0', execute `gem env` for more information
/private/tmp/logstash/rakelib/gems.rake:13:in `block in (root)'
/private/tmp/logstash/rakelib/dependency.rake:12:in `block in (root)'
/private/tmp/logstash/rakelib/vendor.rake:15:in `untar'
/private/tmp/logstash/rakelib/vendor.rake:87:in `block in (root)'
/private/tmp/logstash/rakelib/z_rubycheck.rake:28:in `<main>'
Tasks: TOP => gem:install
(See full trace by running task with --trace)
```

Workarounds:
* Use JRuby 9.1.12.0 to bootstrap (.13 will still be used when running)
* (assuming rake is installed elsewhere):
```
[...]

Issue #8214, opened 2017-09-12
Ubuntu Build Failure: Could not find 'minitar' (= 0.5.4) among 12 total gem(s)


robbavey:
Builds are currently failing due to failed dependency resolution, which appear to be related to the recent move to jruby-9.1.13.0:

https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-unix-compatibility/os=ubuntu/135/console

```
14:00:04 Installing minitar 0.5.4 because the build process needs it.
14:00:04 mkdir -p build
14:00:04 mkdir -p build/bootstrap
14:00:04 [bootstrap] Fetching and installing gem: minitar (0.5.4)
14:00:04 rake aborted!
14:00:04 Gem::UnsatisfiableDependencyError: Unable to resolve dependency: user requested 'rake (= 12.1.0)'
14:00:04 /var/lib/jenkins/workspace/elastic+logstash+master+multijob-unix-compatibility/os/ubuntu/rakelib/gems.rake:39:in `block in (root)'
14:00:04 /var/lib/jenkins/workspace/elastic+logstash+master+multijob-unix-compatibility/os/ubuntu/rakelib/gems.rake:16:in `block in (root)'
14:00:04 /var/lib/jenkins/workspace/elastic+logstash+master+multijob-unix-compatibility/os/ubuntu/rakelib/dependency.rake:12:in `block in (root)'
14:00:04 /var/lib/jenkins/workspace/elastic+logstash+master+multijob-unix-compatibility/os/ubuntu/rakelib/vendor.rake:15:in `block in untar'
14:00:04 /var/lib/jenkins/workspace/elastic+logstash+master+multijob-unix-compatibility/os/ubuntu/rakelib/vendor.rake:87:in `block in (root)'
14:00:04 /var/lib/jenkins/workspace/elastic+logstash+master+multijob-unix-compatibility/os/ubuntu/rakelib/z_rubycheck.rake:28:in `(root)'
14:00:04 Gem::MissingSpecError: Could not find 'minitar' (= 0.5.4) among 12 total gem(s)
14:00:04 Checked in 'GEM_PATH=/var/lib/jenkins/workspace/elastic+logstash+master+multijob-unix-compatibility/os/ubuntu/vendor/bundle/jruby/2.3.0', execute `gem env` for more information
[...]

PR #8558, opened 2017-10-31
Bump gems to 6.0.0

andrewvc:
Version bump to 6.0.0 GA

### DIFF
--- a/Gemfile.jruby-2.3.lock.release
+++ b/Gemfile.jruby-2.3.lock.release
@@ -10,7 +10,7 @@ PATH
       gems (~> 0.8.3)
       i18n (= 0.6.9)
       jar-dependencies
-      jrjackson (~> 0.4.3)
+      jrjackson (~> 0.4.4)
       jruby-openssl (>= 0.9.20)
       manticore (>= 0.5.4, < 1.0.0)
       minitar (~> 0.5.4)
@@ -124,7 +124,7 @@ GEM
     jls-lumberjack (0.0.26)
       concurrent-ruby
     jmespath (1.3.1)
-    jrjackson (0.4.3-java)
+    jrjackson (0.4.4-java)
     jruby-openssl (0.9.21-java)
     jruby-stdin-channel (0.2.0-java)
     json (1.8.6-java)
@@ -147,9 +147,9 @@ GEM
     logstash-codec-es_bulk (3.0.5)
       logstash-codec-line
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-codec-fluent (3.1.3-java)
+    logstash-codec-fluent (3.1.4-java)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-      msgpack
+      msgpack (~> 1.1)
     logstash-codec-graphite (3.0.4)
       logstash-codec-line
       logstash-core-plugin-api (>= 1.60, <= 2.99)
@@ -160,9 +160,9 @@ GEM
       logstash-core-plugin-api (>= 1.60, <= 2.99)
     logstash-codec-line (3.0.4)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-codec-msgpack (3.0.5-java)
+    logstash-codec-msgpack (3.0.6-java)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-      msgpack-jruby
+      msgpack (~> 1.1)
     logstash-codec-multiline (3.0.7)
       jls-grok (~> 0.11.1)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
@@ -175,7 +175,7 @@ GEM
     logstash-codec-rubydebug (3.0.4)
       awesome_print
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-devutils (1.3.4-java)
+    logstash-devutils (1.3.5-java)
       fivemat
       gem_publisher
       insist (= 1.0.0)
@@ -186,7 +186,7 @@ GEM
       rspec (~> 3.0)
       rspec-wait
       stud (>= 0.0.20)
-    logstash-filter-aggregate (2.6.3)
+    logstash-filter-aggregate (2.7.0)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
     logstash-filter-anonymize (3.0.5)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
@@ -195,13 +195,13 @@ GEM
       logstash-core-plugin-api (>= 1.60, <= 2.99)
     logstash-filter-clone (3.0.4)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-filter-csv (3.0.5)
+    logstash-filter-csv (3.0.6)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
     logstash-filter-date (3.1.8)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
     logstash-filter-de_dot (1.0.2)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-filter-dissect (1.0.12)
+    logstash-filter-dissect (1.1.1)
       jar-dependencies
       logstash-core-plugin-api (>= 2.1.1, <= 2.99)
     logstash-filter-dns (3.0.6)
@@ -209,7 +209,7 @@ GEM
       lru_redux (~> 1.1.0)
     logstash-filter-drop (3.0.4)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-filter-elasticsearch (3.1.6)
+    logstash-filter-elasticsearch (3.2.0)
       elasticsearch (>= 5.0.3, < 6.0.0)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
     logstash-filter-fingerprint (3.1.1)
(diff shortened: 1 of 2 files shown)


jakelandis:
I checked out your branch and still found some rc references

```
logstash $ grep -R rc2 . | grep -v .git | grep -v .yml | grep -v .idea
./docs/index-shared1.asciidoc::logstash_version:      6.0.0-rc2
./docs/index-shared1.asciidoc::elasticsearch_version: 6.0.0-rc2
./docs/index-shared1.asciidoc::kibana_version:        6.0.0-rc2
./docs/static/releasenotes.asciidoc:* <<logstash-6-0-0-rc2,Logstash 6.0.0-rc2>>
./docs/static/releasenotes.asciidoc:[[logstash-6-0-0-rc2]]
./docs/static/releasenotes.asciidoc:=== Logstash 6.0.0-rc2 Release Notes
./Gemfile.jruby-2.3.lock.release:    logstash-core (6.0.0.rc2-java)
./Gemfile.jruby-2.3.lock.release:      logstash-core (= 6.0.0.rc2)
./README.md:| [tar](https://snapshots.elastic.co/downloads/logstash/logstash-6.0.0-rc2-SNAPSHOT.tar.gz) |
./README.md:| [zip](https://snapshots.elastic.co/downloads/logstash/logstash-6.0.0-rc2-SNAPSHOT.zip) |
./README.md:| [deb](https://snapshots.elastic.co/downloads/logstash/logstash-6.0.0-rc2-SNAPSHOT.deb) |
./README.md:| [rpm](https://snapshots.elastic.co/downloads/logstash/logstash-6.0.0-rc2-SNAPSHOT.rpm) |
```

jakelandis:
sorry didn't mean to close

andrewvc:
The other versions are usually bumped in a separate PR relatively close to release.

I'm not sure *why* we do things this way, maybe worth discussing tomorrow?

jakelandis:
In that case, LGTM.

andrewvc:
@jakelandis I did forget to bump the gem versions along with this. I sent in a new commit with those changes. Are you LGTM on those?

Gemfile.jruby-2.3.lock.release
```diff
@@ -529,21 +528,21 @@ GEM
       rack
     rack-test (0.7.0)
       rack (>= 1.0, < 3)
-    rake (12.1.0)
+    rake (12.2.1)
```
jakelandis: This one makes me nervous. https://github.com/elastic/logstash/issues/8220 and https://github.com/elastic/logstash/issues/8214 are tangentially related to the rake version (i.e. smoke - no fire). In the spirit of pinning well known working dependency graph, I would suggest to keep this pinned 12.1.0 until 6.1.

Gemfile.jruby-2.3.lock.release
```diff
@@ -49,7 +49,7 @@ GEM
     aws-sdk-v1 (1.67.0)
       json (~> 1.4)
       nokogiri (~> 1)
-    backports (3.8.0)
+    backports (3.10.3)
```
jakelandis: 3.8.0 was from April of this year and is represents a multi-month jump.  I remember this exact question for different review. 

@robbavey - Do you to remember discussions around backports version jump ?  (or was it someone else?)

robbavey: @jakelandis I do recall it - the same thing happened when we bumped to 5.6.4 (https://github.com/elastic/logstash/pull/8470/files). I remember we had a conversation about it - which I thought was here, but doesn't look like it. 

The backports gem version history (https://rubygems.org/gems/backports/versions) shows it to have been completely unchanged from April up until October, when there was a flurry of versions committed over a period of 10 days.

Gemfile.jruby-2.3.lock.release
andrewvc: I'm +1 on this. In fact, I'm thinking that I'm actually +1 on *not* bumping any gems between RC2 and 6.0.0. This would be a change to our release process, but I think it makes sense. WDYT?

jakelandis: My thoughts...
I see RC->Release == patch release, at least the same rules should apply.

IMO patch release rules:

Plugins:  New plugin patch version's and (case-by-case) minor plugin updates are good. They can always downgrade those if there is the bug. The exception here is if a patch/minor upgrade pinned a new transitive version. 

For non-plugins gems, I would think to _not_ bump them unless they are intentionally getting upgraded. For example in this change set there the Jackson version bump is intentional and should be in 6.0.0, as it fixes a regression https://github.com/elastic/logstash/issues/8439

andrewvc: I think the tricky thing is there is no `bundle update --only-logstash-plugins-and-other-things-that-matter`.

Thinking about it, I don't think this PR is the right place to have this discussion (even though I started it!), and we should debate this policy going forward, but not right now.

WDYT? @jordansissel can you chime in here?

jsvd:
Can we perhaps just have a PR that bumps the version from rc2 to GA without touching dependencies? I don't see a reason why the artifact won't build. Also this would help us understand what new code (if any) is shipping after the ga rename.

jasontedor:
I would like to suggest that in general these version bumps should be straightforward and happen quickly (as they hold up other places in the stack). I think that mixing in dependency changes with them should not be done as it's clear it leads to debate the contributes to these version bumps not happening quickly.

jasontedor:
As for a PR that only bumps the version, I had opened #8579.

jsvd:
To unblock the rest of the stack, I've merged https://github.com/elastic/logstash/commit/d58206dab5ac956ab0f95065d19b78b01babae2e, so the discussion on this PR can focus on the dependencies instead of the version bump.

andrewvc:
Thanks @jsvd for the merge, apologies for the holdup @jasontedor .

I've edited our documented version bump policy to split this work into two parts going forward to prevent this issue in the future.

Gemfile.jruby-2.3.lock.release
```diff
@@ -195,21 +195,21 @@ GEM
...
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-filter-csv (3.0.5)
+    logstash-filter-csv (3.0.6)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
     logstash-filter-date (3.1.8)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
     logstash-filter-de_dot (1.0.2)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-filter-dissect (1.0.12)
+    logstash-filter-dissect (1.1.1)
```
andrewvc: All good fixes: https://github.com/logstash-plugins/logstash-filter-dissect/blob/master/CHANGELOG.md

Gemfile.jruby-2.3.lock.release
```diff
@@ -195,21 +195,21 @@ GEM
...
       jar-dependencies
       logstash-core-plugin-api (>= 2.1.1, <= 2.99)
     logstash-filter-dns (3.0.6)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
       lru_redux (~> 1.1.0)
     logstash-filter-drop (3.0.4)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-filter-elasticsearch (3.1.6)
+    logstash-filter-elasticsearch (3.2.0)
```
andrewvc: Innocuous: https://github.com/logstash-plugins/logstash-filter-elasticsearch/blob/master/CHANGELOG.md

Gemfile.jruby-2.3.lock.release
```diff
@@ -271,7 +271,7 @@ GEM
     logstash-input-dead_letter_queue (1.1.1)
       logstash-codec-plain
       logstash-core-plugin-api (>= 1.60, <= 2.99)
-    logstash-input-elasticsearch (4.0.6)
+    logstash-input-elasticsearch (4.1.0)
```
andrewvc: I'm OK with the change to add metadata into here

https://github.com/logstash-plugins/logstash-input-elasticsearch/blob/master/CHANGELOG.md

Gemfile.jruby-2.3.lock.release
```diff
@@ -322,7 +322,7 @@ GEM
       mail (~> 2.6.3)
       mime-types (= 2.6.2)
       stud (~> 0.0.22)
-    logstash-input-jdbc (4.2.4)
+    logstash-input-jdbc (4.3.0)
```
andrewvc: This fixes behavior from here: https://github.com/logstash-plugins/logstash-input-jdbc/issues/147

andrewvc:
@jakelandis I've rebased this and pinned `rake` and `backports` for this release.

LMK if it is now LGTM

Gemfile.lock
```diff
@@ -0,0 +1,762 @@
+PATH
```
jakelandis: Is this intentional to commit the .lock file ?

Gemfile.lock
```diff
@@ -0,0 +1,762 @@
...
+  remote: ./logstash-core
+  specs:
+    logstash-core (7.0.0.alpha1-java)
```
jakelandis: If intentional, probably  don't want 7.x here (and below)

Gemfile.jruby-2.3.lock.release
```diff
@@ -186,7 +186,7 @@ GEM
       rspec (~> 3.0)
       rspec-wait
       stud (>= 0.0.20)
-    logstash-filter-aggregate (2.6.3)
+    logstash-filter-aggregate (2.7.0)
```
jakelandis: Seems like a good and safe change: https://github.com/logstash-plugins/logstash-filter-aggregate/blob/master/CHANGELOG.md#270

Gemfile.lock
andrewvc: No, that is a mistake

andrewvc:
@jakelandis thinking about it, since we now have build candidates for 6.0.0-GA this PR may just be too late.

WDYT @jordansissel ? 

I'm thinking I retarget the pinning of `backports`/`rake` to 6.x.

jakelandis:
Minus the rouge .lock file - LGTM

jakelandis:
LGTM

PR #8561, opened 2017-10-31
Add kibana scheme to list of module options

dedemorton:


docs/static/shared-module-options.asciidoc
```diff
@@ -102,6 +102,16 @@ with Elasticsearch.
...
 
+*`var.kibana.scheme`*::
++
+--
+* Value type is <<string,string>>
+* Default value is "http"
+--
++
+Sets the protocol to use for reaching the Kibana instance. The options are:
+"http" or "https". The default is "http".
```
yaauie: from what I can glean from the [specs](https://github.com/elastic/logstash/blob/4397b2d745b348a1772509bc524f97ddf1f495b9/logstash-core/spec/logstash/modules/kibana_client_spec.rb#L20), this is accurate.

Issue #7884, opened 2017-08-02
Remove "pipeline.output.workers" from doc and default logstash.yml


ppf2:
https://www.elastic.co/guide/en/logstash/5.5/logstash-settings-file.html

`pipeline.output.workers` doesn't do anything on 5.x but it's still documented as a configurable setting.

And it's also in the comments in the default logstash.yml:

```
# How many workers should be used per output plugin instance
#
# pipeline.output.workers: 1
#
```

msspadoni:
I hope not to be wrong and bore you, but perhaps this is referenced also in the https://www.elastic.co/guide/en/logstash/current/performance-troubleshooting.html document at second point of section 4 ("Tune logstash worker settings").

jordansissel:
@msspadoni Not a bore! Thank you for investigating :) This issue is a proposal to remove an unimplemented/unused setting `pipeline.output.workers`. 

What you refer to is `pipeline.workers` (no "output" part) and is a feature that will stay.

dedemorton:
Closing because this is fixed in #8562 and #8908

Issue #7512, opened 2017-06-21
output plugin base `workers` docs are confusing


jordansissel:
![image](https://user-images.githubusercontent.com/131818/27406686-835b4dd2-568a-11e7-8de8-1f00493fb3db.png)

The text in this is incomplete.

Recommendations

* For 6.0, Remove the `workers` setting (something we intended to do anyways, and now may be a good time).
* For 5.x, fix the `workers` docs entries.

dedemorton:
@jordansissel This will be fixed with the new plugin doc gen process, no? I've already fixed this multiple times, but my fixes keep getting overwritten when we regenerate the plugin docs.

PhaedrusTheGreek:
When starting LS 5.2.1 with an elasticsearch output `workers` setting, LS outputs this:

```
[2017-10-10T16:58:51,699][ERROR][logstash.agent           ] Cannot create pipeline {:reason=>"You are using a plugin that doesn't support workers but have set the workers value explicitly! This plugin uses the shared and doesn't need this option"}
```

So it seems that LS 5.x doesn't support the workers setting at all - is there a reason why it should remain in the docs?

Also [the current performance troubleshooting guide](https://www.elastic.co/guide/en/logstash/current/performance-troubleshooting.html) explicitly references the `workers` output setting

dedemorton:
Also see this related PR: https://github.com/elastic/logstash/issues/7884

My understanding here is that the setting pipeline.output.workers needs to be removed, but pipeline.workers is still valid. @jordansissel Is that correct?

jordansissel:
Correct.

dedemorton:
Closed by #8562

PR #8562, opened 2017-11-01
Remove output workers setting from the doc

dedemorton:
Removed the output workers settings plus pipeline.output.workers. Resolves the doc-related issues described in https://github.com/elastic/logstash/issues/7512 and  https://github.com/elastic/logstash/issues/7884. However the losgstash.yml file still needs to be updated as requested in https://github.com/elastic/logstash/issues/7884. Should I do that, too?

@jordansissel Can you review and also look at the section about performance tuning to verify that it's accurate for 6.0? 

I ran some tests in 5.0, and it looks like the workers output setting is not used there either. Should I backport all of these changes to 5.6, too?

### DIFF
--- a/docs/static/performance-checklist.asciidoc
+++ b/docs/static/performance-checklist.asciidoc
@@ -82,7 +82,7 @@ following suggestions:
 
 * Threads in Java have names and you can use the `jstack`, `top`, and the VisualVM graphical tools to figure out which resources a given thread uses.
 
-* On Linux platforms, Logstash labels all the threads it can with something descriptive. For example, inputs show up as `[base]<inputname`, and filter workers show up as `[base]>workerN`, where N is an integer.  Where possible, other threads are also labeled to help you identify their purpose.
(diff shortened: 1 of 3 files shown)


docs/static/performance-checklist.asciidoc
```diff
@@ -83,7 +82,7 @@ following suggestions:
 
 * Threads in Java have names and you can use the `jstack`, `top`, and the VisualVM graphical tools to figure out which resources a given thread uses.
 
-* On Linux platforms, Logstash labels all the threads it can with something descriptive. For example, inputs show up as `[base]<inputname`, filter/output workers show up as `[base]>workerN`, where N is an integer.  Where possible, other threads are also labeled to help you identify their purpose.
+* On Linux platforms, Logstash labels all the threads it can with something descriptive. For example, inputs show up as `[base]<inputname`, and filter workers show up as `[base]>workerN`, where N is an integer.  Where possible, other threads are also labeled to help you identify their purpose.
```
andrewvc: We should call these simply 'pipeline workers', not filter workers, since they work on both filters and outputs.

PR #8219, opened 2017-09-12
Revert "MINOR: Adjust Ruby Version File"

jakelandis:
This reverts commit 3b545b261d3052e555466c2cc27abbb1ea96c389.

The build system still needs to use .12 due some issues we found with clean environments and rake bootstrap and JRuby 9.1.13.0

PR #8565, opened 2017-11-01
Fix for unreliable coverage_spec test.

guyboertje:
rework agent + converge_spec
replace @pipelines hash with concurrent hash
use wait(30, :delay => 1).for instead of wait_for

After some extended tracing, I noted that the RSpec thread in the `wait` call was parked on the reentrantlock because several other threads (periodic_poller pq + dlq) were there first and the `wait` was not able to do its job of call, sleep, retry. Somewhat randomly the other threads unlocked and gave the rspec thread a chance. The default timeout was too low (10 secs?) as well.

In agent.rb, I replaced the Hash and the reentrant-lock with a Concurrent::Hash then refactored the with_pipelines usages. Now RSpec `wait` does loop and is not blocked on the lock. I think that the periodic poller threads will also be more responsive too.

Locally, `rake test:core` passes several runs.


guyboertje:
Yay Jenkins, you beauty!

guyboertje:
Probably backport to 5.6 branch too.

logstash-core/lib/logstash/agent.rb
```diff
@@ -39,9 +39,7 @@ def initialize(settings = LogStash::SETTINGS, source_loader = nil)
     @auto_reload = setting("config.reload.automatic")
     @ephemeral_id = SecureRandom.uuid
 
-    # Do not use @pipelines directly. Use #with_pipelines which does locking
-    @pipelines = {}
-    @pipelines_lock = java.util.concurrent.locks.ReentrantLock.new
+    @pipelines = Concurrent::Hash.new
```
jakelandis: Isn't this is a pretty significant change ?

Doesn't this move from synchronizing all the actions in the block to just synchronizing the actions on the hash itself ?

.ruby-version
```diff
@@ -1 +1 @@
-jruby-9.1.12.0
+jruby-9.1.13.0
```
jakelandis: We tried this before on https://github.com/elastic/logstash/pull/8219 , but it caused issues so we reverted it: https://github.com/elastic/logstash/pull/8219

The Jenkins issues should be resolved now via other channels, but it can still break local devs ability to bootstrap.

logstash-core/lib/logstash/agent.rb
original-brownbear: @guyboertje @jakelandis yea, this looks dangerous. It will be fine for `put` and `get` operations to just move to a concurrent data structure, but for iterations (and I can see at least one `.each` call in the diff) there are no strong guarantees on a concurrent-hashmap => we're not guaranteed to actually hit every pipeline on an iteration if there are concurrent modifications to it => imo sticking with hard locking on every operation is a lot safer (unless we find a way to either avoid iterations or ensure that we don't get concurrent modification while iterating) :(

.ruby-version
guyboertje: will revert

logstash-core/lib/logstash/agent.rb
guyboertje: @jakelandis @original-brownbear 

I am not really in a position to judge whether this class (on which the jruby Concurrent::Hash is based) does or does not take your concerns into account.
https://github.com/ruby-concurrency/concurrent-ruby/blob/master/ext/com/concurrent_ruby/ext/jsr166e/ConcurrentHashMapV8.java

I would love your feedback on this.

original-brownbear: @guyboertje see here https://github.com/ruby-concurrency/concurrent-ruby/blob/master/ext/com/concurrent_ruby/ext/jsr166e/ConcurrentHashMapV8.java#L3728 for example, the iterator is only weakly consistent.

Also, I'd go so far as to say that you cannot even have a strongly consistent iterator here unless you have some lock `each` that also wraps each `put`.
Now that I'm writing this ... can't we have the above ? :D Go for the concurrent map, but still make at least put and iteration mutually exclusive with a lock?

jakelandis: My main concern is the removal of the lock that protects all operations when called `with_pipelines`.  For example prior to this change (due to the lock/yield/unlock in `with_pipelines`) the follow code could only be entered by a single thread at any given time:
```
pipeline_actions = resolve_actions(results.response)
converge_result = converge_state(pipeline_actions)
update_metrics(converge_result)
```
Becuase it was called like this:
```
with_pipelines do |pipelines|
     pipeline_actions = resolve_actions(results.response)
     converge_result = converge_state(pipeline_actions)
     update_metrics(converge_result)
 end
```
and `with_pipelines` is defined as:
```
def with_pipelines
    begin
      @pipelines_lock.lock
      yield @pipelines
    ensure
      @pipelines_lock.unlock
    end
  end
```
...any time you wanted to ensure thread safety (via single thread access) on any grouping of actions related to the pipelines, you could call `with_pipelines` and achieve those guarantees.  

I don't know have enough context or history to fully understand the _exact_ reason we lock the pipelines as we do, just wanted to raise the concern that change is not like for like.

Issue #8353, opened 2017-09-21
Design secrets/credentials container


jordansissel:
Parent ticket: https://github.com/elastic/logstash/issues/6892

This issue is focused on designing the following:

* [ ] The CLI for storing/purging credentials from the credential store.
* [ ] An API for reading plaintext credentials from the store.
* [ ] A way for plugins to register interest in specific credentials (but not arbitrary credential-store access)

The above could be split into separate issues if desired.

Rough specification:
* The file must be encrypted.
* Humans: Can set and delete, but *not* read, credentials.
* Logstash core should provide a way for Plugins to ask for plaintext values.

Safety goals:
* A passphrase is required in order to read or write to the credential store.
* Plaintext values should never be displayed/logged/printed
* Plaintext values should be only held in memory as long as they are needed.

jordansissel:
Some notes from a planning discussion with @andrewvc 

```
Secret Store
    Key Areas
    1 Secret Store Library: API + Implementation of the secret store for setting and retrieving secrets
        - Implementation detail: The actual file store itself (structure, encryption tech, etc)
    2 The command line interface
        - Can only set, list, and remove secrets, cannot read secrets.
        - *CANNOT* read secret values, only list entries, not decrypt.
        - Must be protected by a passphrase (plus encryption)
    3 Logstash configuration
        - Making it possible for users to say "get this credential from the secret store"
        - Logstash needs a way for an operator to provide the decryption key:
            - overload `$FOO` lookup to check Environment and/or Secret Store depending on user config?
            - overload `:validate => :password` to make the user value be a lookup key
        - Plugins need a way to, with permission, fetch secrets


output {
    elasticsearch {
        hosts => [ ... ]
        user => "whatever"
        password => $blah
    }
    elasticsearch {
        hosts => [ ... ]
        user => "whatever"
        password => password2
    }
}
```

jakelandis:
>     1 Secret Store Library: API + Implementation of the secret store for setting and retrieving secrets
 
PR #8566 

The API is general purpose, allowing arbitrary `byte[]` to be persisted.  The key (or alias) for the secret store is a versioned `URN`.  The `URN` format is `urn:logstash:secret:v1:<key>` where key is the unique bit. In the example above it would be `urn:logstash:secret:v1:blah`. 

[...]

PR #8566, opened 2017-11-01
Secret Store: API and JavaKeyStore implementation

jakelandis:
Introduces the API to read/write/delete sensitive data from a secure store and includes a Java KeyStore implementation. Note - this commit does NOT integrate with the Logstash configuration or settings.

Part of #8353 
----

This is part 1 of a multi-part feature. This commit introduces the API and JavaKeyStore implementation. Future commits will have the CLI, integration with Configuration, integration with settings, etc.


logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @Override
+    public void persistSecret(SecretIdentifier identifier, byte[] secret) {
+        try {
+            writeLock.lock();
+            SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
+            //PBEKey requires an ascii password, so base64 encode it
+            byte[] base64 = Base64.getEncoder().encode(secret);
+            PBEKeySpec passwordBasedKeySpec = new PBEKeySpec(asciiBytesToChar(base64));
```
jordansissel: Thoughts on adding types in the future? 

In this PR, there are only byte[] types, but it's possible some future where we have a way to safely persist other types, or at least an abstraction to persist other types. It's possible we don't' need this at all, but I always think about this when considering serialization.

Related: Elasticsearch has a slight wrapper, SecureString, which wraps a char[] and is both CharSequence and Closeable -- we may consider using. Thoughts?

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+            SecretKey secretKey = factory.generateSecret(passwordBasedKeySpec);
+            keyStore.setEntry(identifier.toExternalForm(), new KeyStore.SecretKeyEntry(secretKey), protectionParameter);
+            try (final OutputStream os = Files.newOutputStream(keyStorePath)) {
+                keyStore.store(os, keyStorePass);
```
jordansissel: we should reset permissions here, also, for safety?

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+
+    @Override
+    public void purgeSecret(SecretIdentifier identifier) {
+        try {
+            writeLock.lock();
+            try (final InputStream is = Files.newInputStream(keyStorePath)) {
+                keyStore.load(is, keyStorePass);
+                keyStore.deleteEntry(identifier.toExternalForm());
```
jordansissel: Let's add a logger.trace call here to note what is deleted.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+                    keyStore.load(is, keyStorePass);
+                    SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
+                    KeyStore.SecretKeyEntry secretKeyEntry = (KeyStore.SecretKeyEntry) keyStore.getEntry(identifier.toExternalForm(), protectionParameter);
+                    //not found
+                    if (secretKeyEntry == null) {
+                        return null;
+                    }
+                    PBEKeySpec passwordBasedKeySpec = (PBEKeySpec) factory.getKeySpec(secretKeyEntry.getSecretKey(), PBEKeySpec.class);
+                    //base64 encoded char[]
+                    char[] base64secret = passwordBasedKeySpec.getPassword().clone();
```
jordansissel: PBEKeySpec.getPassword() docs say "Returns a copy of the password." -- I don't think the .clone() is necessary.

logstash-core/src/main/java/org/logstash/secret/SecretIdentifier.java
```diff
@@ -0,0 +1,101 @@
...
+     *
+     * @param part     The part of the URN to validate
+     * @param partName The name of the part used for logging.
+     * @return The validated and transformed part.
+     */
+    private static String validateWithTransform(String part, String partName) {
+        if (part == null || part.isEmpty()) {
+            throw new IllegalArgumentException(String.format("%s may not be null or empty", partName));
+        }
+        return part.replace(":", "_").toLowerCase(Locale.US);
```
jordansissel: Why? Is this because you use `/:/.split(urn, 5)` ?

Feels weird to transform the user input from `foo:bar` to `foo_bar` internally. I worry, in debugging, this will cause confusion. Is it necessary?

logstash-core/src/main/java/org/logstash/secret/SecretIdentifier.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Converts this object to a format acceptable external {@link String} format. Note - no gauruntees are made with respect to encoding or safe use. For example, the external
+     * format may not be URL safely encoded.
+     *
+     * @return the externally formatted {@link String}
+     */
+    public String toExternalForm() {
+        StringBuilder sb = new StringBuilder(100);
```
jordansissel: Thoughts on using String.format instead?

`String.format("urn:logstash:secret:%s:%s", VERSION, this.key == null ? "-" : this.key.toLowerCase(Locale.US))`

Feels shorter, though maybe you have reasons for the StringBuilder use?

logstash-core/src/main/java/org/logstash/secret/SecretIdentifier.java
```diff
@@ -0,0 +1,101 @@
...
+        sb.append("urn").append(":");
+        sb.append("logstash").append(":");
+        sb.append("secret").append(":");
+        sb.append(VERSION).append(":");
+        sb.append(this.key == null ? "-" : this.key.toLowerCase(Locale.US));
```
jordansissel: should we allow a null key? I kinda would prefer to reject it.

jordansissel:
Oh and I forgot — thank you for making this PR small enough to review
easily for me. Splitting it into multiple PRs is lovely. :)

logstash-core/src/main/java/org/logstash/secret/SecretIdentifier.java
jakelandis: It's just a minor style preference, imo it is easier to read as is.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+     * Constructor - will create the keystore if it does not exist
+     *
+     * @param keyStorePath The full path to the java keystore
+     * @param keyStorePass The password to the keystore
+     * @throws SecretStoreException if errors occur while trying to create or access the keystore
+     */
+    public JavaKeyStore(Path keyStorePath, char[] keyStorePass) {
+        try {
+            this.keyStorePath = keyStorePath;
+            String keyStoreType = System.getProperty("java.keystore.type", "pkcs12");
```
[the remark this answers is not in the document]
jakelandis: the only thought there was for some custom keystore type and have this there as an undocumented way to achieve that.  However, in the next iteration, i have better way for consumers to BYOSS (Bring Your Own Secret Store). 

Will hard code to pkcs12.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+                    keyStore.setEntry(logstashMarker.toExternalForm(), new KeyStore.SecretKeyEntry(secretKey), protectionParameter);
+                    keyStore.store(os, keyStorePass);
+
+                    PosixFileAttributeView attrs = Files.getFileAttributeView(keyStorePath, PosixFileAttributeView.class);
+                    if (attrs != null) {
+                        attrs.setPermissions(PosixFilePermissions.fromString("rw-rw----"));
+                    }
+                }
+            }
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
jakelandis: This is a bigger conversation, and will comment on main thread.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+
+    /**
+     * Converts bytes from ascii encoded text to a char[] and zero outs the original byte[]
+     *
+     * @param bytes the bytes from an ascii encoded text (note - no validation is done to ensure ascii encoding)
+     * @return the corresponding char[]
+     */
+    private char[] asciiBytesToChar(byte[] bytes) {
```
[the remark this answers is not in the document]
jakelandis: yup, already pulled this out to a utility class for my next iteration and will merge that into this PR.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+        return chars;
+    }
+
+    /**
+     * Converts characters from ascii encoded text to a byte[] and zero outs the original char[]
+     *
+     * @param chars the chars from an ascii encoded text (note - no validation is done to ensure ascii encoding)
+     * @return the corresponding byte[]
+     */
+    private byte[] asciiCharToBytes(char[] chars) {
```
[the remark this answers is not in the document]
jakelandis: yup, already pulled this out to a utility class for my next iteration and will merge that into this PR.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+        return bytes;
+    }
+
+    /**
+     * Attempt to keep the secret data out of the heap.
+     *
+     * @param secret the secret to zero out
+     */
+    private void clearSecret(byte[] secret) {
```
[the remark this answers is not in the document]
jakelandis: yup, already pulled this out to a utility class for my next iteration and will merge that into this PR.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+        if (secret != null) {
+            for (int i = 0; i < secret.length; ++i) {
```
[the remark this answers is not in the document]
jakelandis: thanks, will do.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+                secret[i] = '\0';
+            }
+            secret = null;
```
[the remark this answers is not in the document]
jakelandis: yeah, no effect.  will remove, thanks.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+        return identifiers;
+    }
+
+    @Override
+    public void persistSecret(SecretIdentifier identifier, byte[] secret) {
+        try {
+            writeLock.lock();
+            SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
+            //PBEKey requires an ascii password, so base64 encode it
+            byte[] base64 = Base64.getEncoder().encode(secret);
```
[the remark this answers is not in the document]
jakelandis: agreed.  I have pulled all the transforms out to a utility method and will include those in an update to this PR.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: I don't think so.  I think it would be surprising side effect.  If the permission changed, and it was intentional,  we don't want to change it back (maybe the user locked it down to owner read only?) , it if was un-intentional i think that is outside the scope of the function here.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: For now, I think byte[] suites us for the API contract, and want to code against the API so that we can implement other backends, or allow 3rd parties to bring their own secret store.  

As an implementation detail, all of the data, regardless of upstream type is re-encoded as base64 char[] to ensure ASCII compatibility with the underlying library. Any additional types supported would be essentially be helper methods to get the serialized form (e.g. byte[]). 

I would think that the functions/helpers to get to the serialized form should be the responsibility of the given type, not the secret store.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: yeah, i need some more debug logging. Also noticed some e.printStackTrace() snuck in here... will beef up logging and error handling a bit more.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: will do. thanks.

logstash-core/src/test/java/org/logstash/secret/SecretIdentifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+        assertThat(id.getKey()).isEqualTo(id2.getKey());
+        assertThat(id.toExternalForm()).isEqualTo(id.toExternalForm()).isEqualTo(id.toString()).isEqualTo(id2.toString());
+    }
+
+    /**
+     * Colons get transformed to underscores
+     */
+    @Test
+    public void testColon() {
+        SecretIdentifier id = new SecretIdentifier("foo:bar");
```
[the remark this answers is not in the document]
jakelandis: It is there to help avoid breaking the URN if a user added colons to their key.

However, i dis-agree with last week me and I should be able to support colons without this wonky 'fix'. 

Will update.

logstash-core/src/test/java/org/logstash/secret/SecretIdentifierTest.java
```diff
@@ -0,0 +1,73 @@
...
+        assertThat(SecretIdentifier.fromExternalForm("urn:logstash:secret:v1:foo")).isEqualTo(new SecretIdentifier("foo"));
+        assertThat(SecretIdentifier.fromExternalForm("urn:logstash:secret:v1:foo:bar")).isEqualTo(new SecretIdentifier("foo:bar"));
+    }
+
+    /**
+     * invalid urn's return null
+     */
+    @Test
+    public void testFromExternalInvalid() {
+        assertThat(SecretIdentifier.fromExternalForm("urn:logstash:secret:nope:foo")).isNull();
```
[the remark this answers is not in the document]
jakelandis: ` IllegalArgumentException` is thrown when the argument to the method is incorrect, `fromExternalForm` does not have an argument. 

Sorry for the pedantic answer ^^ , but in general I would throw an exception on input validation, but am pretty reserved for throwing exceptions in logic unless it is clearly an exceptional case. In this case (splitting hairs) a bad identifier is like one that does not exist and pretty sure other constraints wouldn't even let it happen.

logstash-core/src/test/java/org/logstash/secret/store/backend/JavaKeyStoreTest.java
```diff
@@ -0,0 +1,275 @@
...
+     * @throws Exception when ever it wants to.
+     */
+    @Test
+    public void isLogstashKeystore() throws Exception {
+        //newly created
+        byte[] marker = keyStore.retrieveSecret(new SecretIdentifier(JavaKeyStore.LOGSTASH_MARKER));
+        assertThat(new String(marker, StandardCharsets.UTF_8)).isEqualTo(JavaKeyStore.LOGSTASH_MARKER);
+
+        //exiting
+        JavaKeyStore existingKeyStore = new JavaKeyStore(Paths.get(this.getClass().getClassLoader().getResource("logstash.keystore").toURI()), "mypassword".toCharArray());
```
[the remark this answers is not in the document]
jakelandis: :) already fixed in the next PR. (won't change in this PR since the constructor changes in the next PR and don't want to make a test code change just for this PR)

logstash-core/src/main/java/org/logstash/secret/SecretIdentifier.java
jakelandis: Agree, this is weird. Will change.

logstash-core/src/test/java/org/logstash/secret/SecretIdentifierTest.java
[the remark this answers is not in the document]
jakelandis: er... nevermind, `fromExternalForm` does take an argument. I will have it throw an `IllegalArgumentException`.

logstash-core/src/main/java/org/logstash/secret/SecretIdentifier.java
jakelandis: We already dis-allow via the `validateWithTransform` via the constructor.  Will remove this check here since it not possible.  (probably forgot to remove this check when added that method.)  thanks.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+        try {
+            this.keyStorePath = keyStorePath;
+            String keyStoreType = System.getProperty("java.keystore.type", "pkcs12");
+            this.keyStore = KeyStore.getInstance(keyStoreType);
+            this.keyStorePass = keyStorePass;
+            ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+            readLock = readWriteLock.readLock();
+            writeLock = readWriteLock.writeLock();
+            this.protectionParameter = new PasswordProtection(keyStorePass);
+            SecretIdentifier logstashMarker = new SecretIdentifier(LOGSTASH_MARKER);
```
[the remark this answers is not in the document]
jakelandis: The versioned URN should handle any need to gracefully upgrade the contents _inside_ the store, so there should not be a need to version the keystore itself.

However, ES wraps the Keystore for other reasons.  Let's discuss those over Slack to see what properties of that we want.  

Any changes in this direction however, should be in a different PR.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+                    keyStore = KeyStore.Builder.newInstance(keyStoreType, null, protectionParameter).getKeyStore();
+                    SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
+                    byte[] base64 = Base64.getEncoder().encode(LOGSTASH_MARKER.getBytes(StandardCharsets.UTF_8));
+                    SecretKey secretKey = factory.generateSecret(new PBEKeySpec(asciiBytesToChar(base64)));
+                    keyStore.setEntry(logstashMarker.toExternalForm(), new KeyStore.SecretKeyEntry(secretKey), protectionParameter);
+                    keyStore.store(os, keyStorePass);
+
+                    PosixFileAttributeView attrs = Files.getFileAttributeView(keyStorePath, PosixFileAttributeView.class);
+                    if (attrs != null) {
+                        attrs.setPermissions(PosixFilePermissions.fromString("rw-rw----"));
```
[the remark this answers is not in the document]
jakelandis: `keystore.store` writes the file, so not much we can do about that issue (unless we wrap like ES does). 

>  In this code, we would write, then set permissions, and if the permissions set fails (for some reason), we'll have a possiblye-readable secret store on disk.

Nice catch... will handle the error and delete the store on failure to set permissions.

jakelandis:
> catch specific exceptions, not Exception.

I don't believe that to be a hard rule. 

My general philosophy behind modular design is that even internal code should be viewed as a subsystems that could be re-used in a different scope.  Code against an API and encapsulate the concerns of that API in a modular fashion. For example, if we wanted to pull out the secret store stuff to it's own library and include it in a JAR, it would be very easy to do. 

When dealing with boundaries of subsystems, exception handling for consumers should be easy and a predictable.  In this case anything coming from the SecretStore API should have a single corresponding exception. 

To accommodate this, i am wrapping _every_ exception in a sub-class of `SecretStoreException` , such that if the consumer chooses to handle ANY exception coming from this API they need to only to catch `SecretStoreException`.  If for some reason, they want more granularity, I am actually throwing a subclass like `SecretStoreException.NotLogstashKeyStore` that they may catch if they so choose.   

While I agree that business level code shouldn't catch `Exception` to avoid being overly zealous and handling things it shouldn't ... i _do_ believe that the boundaries of the subsystems should, to  allow for easy and predicable upstream exception handling.  

You will see this pattern in nearly all of my designs w/r/t error handling.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,227 @@
...
+        } catch (KeyStoreException e) {
+            e.printStackTrace();
+        } finally {
+            readLock.unlock();
+        }
+        return identifiers;
+    }
+
+    @Override
+    public void persistSecret(SecretIdentifier identifier, byte[] secret) {
```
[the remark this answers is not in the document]
jakelandis: will do.

jordansissel:
Talked with @jakelandis a bit about two things:

## `catch Exception`

My general concern was about the software providing actionable errors to users. `catch Exception` and wrapping with a new exception is related, and I accidentally conflated that. For an example scenario, if setting permissions on a file fails, I am fine (style-wise) doing `catch Exception` as long as the reported exception-or-log tells the user "Setting permissions on file /x/y/z failed: <cause>"

## file header, serialization

Jake and I agreed we are probably not going to switch away from using PKCS12, and even if we do, the upgrade step will be part of the command-line tool.

We think we have sufficient schema versioning in place as-is -- specifically, the serialization format of the values which have schema specified by the `ExternalForm` as "urn:logstash:secret:v1:..."

jakelandis:
@jordansissel - I believe all review comments have been addressed. I left the review changes in a seperate diff : https://github.com/elastic/logstash/pull/8566/commits/4d78c13bc955f4ac3a5d1b84d34eb4b76e0243eb, but will fixup before merge to master.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,215 @@
...
+                byte[] marker = retrieveSecret(logstashMarker);
+                if (marker == null) {
+                    throw new SecretStoreException.NotLogstashKeyStore(String.format("Found a keystore at %s, but it is not a logstash keystore.",
+                            keyStorePath.toAbsolutePath().toString()));
+                }
+                LOGGER.debug("Using existing keystore at {}", keyStorePath.toAbsolutePath().toString());
+            } catch (NoSuchFileException noSuchFileException) {
+                LOGGER.info("Keystore not found at {}. Creating new keystore.", keyStorePath.toAbsolutePath().toString());
+                //create the keystore on disk with a default entry to identify this as a logstash keystore
+                try (final OutputStream os = Files.newOutputStream(keyStorePath)) {
```
jordansissel: Thinking about this more, I think this is not a secure way to create this file.

We want to create this file and set the permissions simultaneously. `Files.newOutputStream` doesn't let you do this. Instead, we should probably do this:

```java
import java.nio.file.Paths;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.attribute.PosixFilePermissions;
public class X {
  public static void main(String ... args) throws Exception {
    Path path = Paths.get(args[0]);
    Files.createFile(path, PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rw-------")));
    //OutputStream os = new FileOutputStream(path.toFile());
  }
}
```

Observing the syscalls on Linux, the file permissions are given to the `open(2)` call:

```
% rm -f /tmp/x; strace -fe trace=file java X /tmp/x |& grep /tmp/x
execve("/usr/bin/java", ["java", "X", "/tmp/x"], 0x7fff34f82710 /* 36 vars */) = 0
[pid 47040] open("/tmp/x", O_WRONLY|O_CREAT|O_EXCL, 0600) = 5
```

This ensures the file never exists without safe access control.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,215 @@
...
+                    keyStore = KeyStore.Builder.newInstance(KEYSTORE_TYPE, null, protectionParameter).getKeyStore();
+                    SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
+                    byte[] base64 = SecretStoreUtil.base64Encode(LOGSTASH_MARKER.getBytes(StandardCharsets.UTF_8));
+                    SecretKey secretKey = factory.generateSecret(new PBEKeySpec(SecretStoreUtil.asciiBytesToChar(base64)));
+                    keyStore.setEntry(logstashMarker.toExternalForm(), new KeyStore.SecretKeyEntry(secretKey), protectionParameter);
+                    keyStore.store(os, this.keyStorePass);
+
+                    PosixFileAttributeView attrs = Files.getFileAttributeView(keyStorePath, PosixFileAttributeView.class);
+                    if (attrs != null) {
+                        try {
```
jordansissel: If we do my proposal above to create the file with the correct permissions simultaneously, we don't need to do this extra setPermissions call and can remove ~10 lines of code here.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,215 @@
...
+            this.protectionParameter = new PasswordProtection(this.keyStorePass);
+            SecretIdentifier logstashMarker = new SecretIdentifier(LOGSTASH_MARKER);
+
+            try (final InputStream is = Files.newInputStream(keyStorePath)) {
+                try {
+                    keyStore.load(is, this.keyStorePass);
+                } catch (IOException ioe) {
+                    if (ioe.getCause() instanceof UnrecoverableKeyException) {
+                        throw new SecretStoreException.AccessException(
+                                String.format("Can not access Java keystore at %s, check file permissions and the keystore password", keyStorePath.toAbsolutePath
```
jordansissel: period, not comma `at %s` recommend also `Please verify correct file permissions and keystore password`

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,215 @@
...
+                    }
+                    LOGGER.info("Keystore created at {}", keyStorePath.toAbsolutePath().toString());
+                } catch (Exception e) {
+                    throw new SecretStoreException.CreateException("Error can not create Java keystore", e);
+                }
+            }
+        } catch (SecretStoreException sse) {
+            throw sse;
+        } catch (Exception e) { //should never happen
+            throw new SecretStoreException("Error while trying to create or launch the JavaKeyStore", e);
```
jordansissel: launch? Load?

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,215 @@
...
+ */
+public final class JavaKeyStore implements SecretStore {
+    public static final String LOGSTASH_MARKER = "logstash-key-store";
+    private static final Logger LOGGER = LogManager.getLogger(JavaKeyStore.class);
+    private static final String KEYSTORE_TYPE = "pkcs12";
+
+    private final char[] keyStorePass;
+    private final Path keyStorePath;
+    private final ProtectionParameter protectionParameter;
+    private final Lock readLock;
```
jordansissel: Should this be a [FileLock](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileLock.html) (or in addition?) to avoid the keystore being mutated while we are using it? (Logstash reading, operator writing, for example)

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,215 @@
...
+                                Files.delete(keyStorePath);
+                            } catch (IOException ioe) {
+                                //do nothing
+                            }
+                            throw new SecretStoreException.CreateException("Error can not set file permissions on Java keystore", e);
+                        }
+                    }
+                    LOGGER.info("Keystore created at {}", keyStorePath.toAbsolutePath().toString());
+                } catch (Exception e) {
+                    throw new SecretStoreException.CreateException("Error can not create Java keystore", e);
```
jordansissel: recommend: "Failed to create keystore"

"Error" in the current sentence seems like a noun and we are saying a thing called "Error" could not create something.

Do we need to specify "Java keystore" ? Can we say "Logstash key store" or "secret store" (whatever we are calling this)

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,215 @@
...
+        return identifiers;
+    }
+
+    @Override
+    public void persistSecret(SecretIdentifier identifier, byte[] secret) {
+        try {
+            writeLock.lock();
+            SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
+            //PBEKey requires an ascii password, so base64 encode it
+            byte[] base64 = SecretStoreUtil.base64Encode(secret);
```
jordansissel: I understand the rationale, and your comment is very helpful, but I'm not gonna lie, this is a *weird* hack. PBEKey is a really strange api in this respect ;)

No change required. Thank you for the informative comment.

yaauie: > PBEKey requires an ascii password

I'm not finding explicit documentation to support this statement in either [`PBEKey`](https://docs.oracle.com/javase/8/docs/api/javax/crypto/interfaces/PBEKey.html) or [`PBEKeySpec`](https://docs.oracle.com/javase/8/docs/api/javax/crypto/spec/PBEKeySpec.html); if it _does_ have this limitation, it feels odd to me that we would elect to use PBEKey as our storage mechanism since we're not using the resulting keys for any PBE-encryption operations.

PBEKey appears to be `char[]`-based, and older PKCS implementations that leveraged it apparently only used the lower-byte from each `char`, so maybe that's related to the limitation? If so, it shouldn't matter as we're not using these keys for any encryption operations, merely as a storage mechanism.

logstash-core/src/main/java/org/logstash/secret/SecretIdentifier.java
```diff
@@ -0,0 +1,101 @@
...
+    public static SecretIdentifier fromExternalForm(String urn) {
+        if (urn == null || !urnPattern.matcher(urn).matches()) {
+            throw new IllegalArgumentException("Invalid external form " + urn);
+        }
+        String[] parts = colonPattern.split(urn, 5);
+        return new SecretIdentifier(validateWithTransform(parts[4], "key"));
+    }
+
+    /**
+     * Minor validation and transformation on input. Converts ":" to "_" to avoid URN conflicts, and downcases the parts
```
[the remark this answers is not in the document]
jakelandis: thanks will do.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: I don't think so, that would require a downtime to update the secret store.  The likely hood of collision is very low since these the only interaction with the key store should be at settings load and pipeline load time.  Even if there is a cross-JVM collision, i am not sure how the underlying implementation handles it, a quick scan of the code appears it would result in last action wins, but hard to say for certain.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: will do, thanks.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: load , i think.  will change.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: Thanks, my grammar is poor suggestions are always welcome. 

will upper case Logstash and keep a consistent 'keystore' reference. The idea is a secret store, the implementation is here is a keystore

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: Agree'd it is hacky, but it is more so because (my inference) PBE was never intended to store and retrieve passwords (for good reason too, imo). It is intended to be the input into a algorithms such as PBKDF2. I am not sure if using it "correctly" would impose the ASCII restraint. 

I considered tossing out using a Java Keystore entirely due to this and simply have something proprietary (with standard crypto of course), but the keystore implementation is secure, well vetted, supported by standard tooling, is non-controversial, and also backs ES's implementation. 

It does indeed have the ASCII limitation as evident by the failing tests without this hack, and it appears that [ES doesn't support non-ascii secrets](https://github.com/elastic/elasticsearch/blob/7a035f5f8435962072157f46031e73855fe4a875/core/src/main/java/org/elasticsearch/common/settings/KeyStoreWrapper.java#L393). 

Also, fwiw, the password to the keystore itself requires ASCII, in this PR I am also base64 encoding it, but will change in a subsequent review to only conditionally base64 encode to better support standard tooling.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: Nice,  but not perfect. 

https://docs.oracle.com/javase/tutorial/essential/io/fileAttr.html#posix
> The code then attempts to create a new file with those permissions. Note that the umask also applies, so the new file might be more secure than the permissions that were requested.

...so if we are trying to set "rw-rw----", but the umask is for the directory is "rwxr-xr-x" , you end up with "rw-r-----".

A couple options here... but I propose the following

Set it in both places (at create and afterwards).
Remove the code that deletes and throws if it can not set it afterwards, since that would only now be expanding permissions.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: see above, i can remove some code, but not all of it.

jakelandis:
@jordansissel - Latest set of review changes have been pushed https://github.com/elastic/logstash/pull/8566/commits/24b369966afcf18fd2eeff49c002680e477ad114 (will fixup prior to merge to master)

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jordansissel: > better support standard tooling

I'm OK if we don't expect/support any alternative tooling. My main goal with reusing KeyStore is to reuse the technology (crypto, formats, etc0, not the human interface. It's OK if we produce something incompatible with `keytool` and `openssl pkcs12` tools (compatibility is not a goal).

logstash-core/src/main/java/org/logstash/secret/SecretIdentifier.java
jordansissel: Sounds ok to me -- I have no strong opinion and the method functions as intended. Thanks for letting me into your brain a bit :)

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,209 @@
...
+        } catch (Exception e) { //should never happen
+            throw new SecretStoreException("Error while trying to create or load the Logstash keystore", e);
+        }
+    }
+
+    @Override
+    public Collection<SecretIdentifier> list() {
+        Set<SecretIdentifier> identifiers = new HashSet<>();
+        try {
+            readLock.lock();
```
jordansissel: retrieveSecret and purgeSecret both load the keystore (open an inputstream, call keystore.load, etc) at the beginning of the method, but this one does not. Expected?

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jordansissel: > that would require a downtime to update the secret store

My reading of the code is disagreeing with this claim.

readLock is used for listing and retrieving, for example. retrieveSecret does this:

* readLock.lock
* open the keystore file
* Load keystore
* do work
* unlock

This lock protects the entire file open/load/etc operation within the JVM. Extending this protection to the OS is a good idea (with FileLock)

I agree with you that there is an unlikely chance of a corrupted read-during-write given the store is only likely accessed during pipeline loads/reloads, etc.

My concern, for example, is someone copying in a new store *while* Logstash is starting up, possibly resulting in an InputStream that is corrupted because our reads are part from the old file, part from the new file.

We can add file locks in a future PR perhaps? I am not seeing any harm in adding safety.

Two safety ideas:
1) Write atomically (write to keystore.new + rename). This is what we do for PQ.
2) Use file locks.

Atomic writes are safer for system halts/faults where a write is interrupted. FileLock is sufficient for blocking writes while the file is being read.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
yaauie: After digging into the PBEKey and seeing that its constructor rejects anything with a codepoint outside the range [`\u0020`,`\u007E`], I understand the need for this hack if we want to use PBEKey to store arbitrary binary data.

The design of the Java KeyStore is definitely more about storing keys for use in crypto, so exploiting `PBEKey` to store arbitrary binary data in the Java KeyStore is definitely stretching the abstraction a bit. I don't think it's a blocker since each behaviour we're relying on is explicitly documented, but it's good to acknowledge.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,209 @@
...
+            //PBEKey requires an ascii password, so base64 encode it
+            byte[] base64 = SecretStoreUtil.base64Encode(secret);
+            PBEKeySpec passwordBasedKeySpec = new PBEKeySpec(SecretStoreUtil.asciiBytesToChar(base64));
+            SecretKey secretKey = factory.generateSecret(passwordBasedKeySpec);
+            keyStore.setEntry(identifier.toExternalForm(), new KeyStore.SecretKeyEntry(secretKey), protectionParameter);
+            try (final OutputStream os = Files.newOutputStream(keyStorePath)) {
+                keyStore.store(os, keyStorePass);
+            } finally {
+                passwordBasedKeySpec.clearPassword();
+                SecretStoreUtil.clearBytes(secret);
```
yaauie: `secret` was already cleared when it was sent as an argument to `SecretStoreUtil#base64Encode(byte[])`, so clearing it again has no effect

It may be helpful to rename the methods that clear the passed `byte[]` to indicate that they do (e.g., `SecretStoreUtil#base64EncodeAndClear(byte[])`) so that it's easier to tell in calling code what has already been cleared.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: Nice catch! ... yes this indeed a bug. We need to reload the keystore from disk to ensure we have an updated view of the keystore. This is is also an issue for persist. Will fix with tests.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -0,0 +1,209 @@
...
+            } catch (NoSuchFileException noSuchFileException) {
+                LOGGER.info("Keystore not found at {}. Creating new keystore.", keyStorePath.toAbsolutePath());
+                //wrapped in a system property mostly for testing, but could be used for more restrictive defaults
+                String keyStorePermissions = System.getProperty("logstash.keystore.file.perms", "rw-rw----");
+                //create the keystore on disk with a default entry to identify this as a logstash keystore
+                Files.createFile(keyStorePath, PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString(keyStorePermissions)));
+                try (final OutputStream os = Files.newOutputStream(keyStorePath, StandardOpenOption.WRITE)) {
+                    keyStore = KeyStore.Builder.newInstance(KEYSTORE_TYPE, null, protectionParameter).getKeyStore();
+                    SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
+                    byte[] base64 = SecretStoreUtil.base64Encode(LOGSTASH_MARKER.getBytes(StandardCharsets.UTF_8));
```
yaauie: It seems like we're duplicating a lot of the logic from our persistSecret; would it be possible to share that logic (e.g., once this object has been constructed "enough", use its own `JavaKeyStore#persistSecret(byte[])` instead of re-implementing it in the huge constructor?

jakelandis:
Not sure why GH won't let me comment inline...

> Two safety ideas:
    Write atomically (write to keystore.new + rename). This is what we do for PQ.
    Use file locks.

Agreed, we should provide some protections here.  

> My reading of the code is disagreeing with this claim.

Apologies, I was bit hasty in my conclusion. 

> We can add file locks in a future PR perhaps? 

Yes. (or at least atomic writes)

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: > secret was already cleared when it was sent as an argument

Agreed, but I just wanted to be explicit and not rely on a different methods side effect for this. 

> helpful to rename the methods

No strong objections to changing the name, but the reason I didn't add that to the name is the presumption that all things that all methods of the SecretStoreUtil has that side effect and thus every method would need that suffix making it abit redundant.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: Agree could remove some duplication here...however, there are a couple a difference , for example not base64 encoding the marker, and (soon to be) reloading the keystore (which creates a chicken/egg issue). 

I have more changes coming, and I don't like this monster constructor...but for now I will keep it as-is and look to DRY things up once the code is unlikely to change anymore.

jakelandis:
@jordansissel @yaauie - thanks for your continued feedback. 

The latest round of changes have been added and can be viewed independently here: https://github.com/elastic/logstash/pull/8566/commits/baeb2f4aaab5abaf28b516e0c4b0cedef1157fc1

jordansissel:
LGTM

jakelandis:
@jordansissel - thanks! I will fixup the commits and re-target to a feature branch.  I have a second (major) round of changes coming that should go into master all at once.

jakelandis:
Replaced by https://github.com/elastic/logstash/pull/8657 (no changes but squashing commits and re-targets a feature branch)

PR #8568, opened 2017-11-02
#6961 use codegen to invoke the appropriate ByteBuffer Cleaner depending on JVM version

original-brownbear:
fixes Java9 compatibility being broken by #6961:

* Use code gen to either use `sun.misc.Unsafe#invokeCleaner` (Java 9) or `sun.misc.Cleaner` (Java 8) to force `munmap` of the `mmap`ed memory region backing `MmapPageIO`
   * Essentially this is the same approach used by Lucene but in a lot less code, thanks to us having Janino available at Runtime
* Fix memory monitor test to pass on Java 9 (Java 9 will run G1 here during tests since they use the default GC, so the old GCs naming doesn't apply)

This makes at least `./gradlew javaTests` (this means JUnit only, RSpec still breaks but should be fine once JRuby `9.1.14.0` is out) pass on Java 9. Without this change, the code doesn't compile on Java 9 because of the unavailability of `sun.misc.Cleaner` there.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/LogstashJavaCompat.java
+++ b/logstash-core/src/main/java/org/logstash/LogstashJavaCompat.java
@@ -1,0 +1,72 @@
+package org.logstash;
+
+import java.io.StringReader;
+import java.util.ArrayList;
+import java.util.Collection;
+import org.codehaus.janino.ClassBodyEvaluator;
+import org.logstash.ackedqueue.io.ByteBufferCleaner;
+
+/**
+ * Logic around ensuring compatibility with Java 8 and 9 simultaneously.
+ */
+public final class LogstashJavaCompat {
+
+    /**
+     * True if current JVM is a Java 9 implementation.
+     */
+    public static final boolean IS_JAVA_9_OR_GREATER = isAtLeastJava9();
+
+    /**
+     * Sets up an appropriate implementation of {@link ByteBufferCleaner} depending no whether or
+     * not the current JVM is a Java 9 implementation.
+     * @return ByteBufferCleaner
+     */
+    public static ByteBufferCleaner setupBytebufferCleaner() {
+        final ClassBodyEvaluator se = new ClassBodyEvaluator();
+        final Collection<String> imports = new ArrayList<>();
+        imports.add("java.nio.MappedByteBuffer");
+        final String cleanerCode;
+        final String ctorCode;
+        final String fieldsCode;
+        if (isAtLeastJava9()) {
+            imports.add("sun.misc.Unsafe");
+            imports.add("java.lang.reflect.Field");
+            cleanerCode = "unsafe.invokeCleaner(buffer);";
+            ctorCode = "Field unsafeField = Unsafe.class.getDeclaredField(\"theUnsafe\");" +
+                "unsafeField.setAccessible(true);" +
+                "unsafe = (Unsafe) unsafeField.get(null);";
+            fieldsCode = "private final Unsafe unsafe;";
+        } else {
+            imports.add("sun.misc.Cleaner");
+            imports.add("sun.nio.ch.DirectBuffer");
+            cleanerCode = "Cleaner c=((DirectBuffer)buffer).cleaner();if(c != null){c.clean();}";
+            ctorCode = "";
+            fieldsCode = "";
+        }
+        se.setImplementedInterfaces(new Class[]{ByteBufferCleaner.class});
+        se.setClassName("ByteBufferCleanerImpl");
+        se.setDefaultImports(imports.toArray(new String[0]));
+        try {
+            return (ByteBufferCleaner) se.createInstance(
+                new StringReader(String.format(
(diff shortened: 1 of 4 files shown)


jakelandis:
Can you copy an example of the generated output to a comment here for easier review?

original-brownbear:
@jakelandis sure for JDK8 we get:

```java
import java.nio.MappedByteBuffer;
import org.logstash.ackedqueue.io.ByteBufferCleaner;
import sun.misc.Cleaner;
import sun.nio.ch.DirectBuffer;

class ByteBufferCleanerImpl implements ByteBufferCleaner {

    public ByteBufferCleanerImpl() throws Exception {
    }

    public void clean(MappedByteBuffer buffer) {
        Cleaner c = ((DirectBuffer) buffer).cleaner();
        if (c != null) {
            c.clean();
        }
    }
}
```

for JDK9 we get:

```java
import java.lang.reflect.Field;
import java.nio.MappedByteBuffer;
import org.logstash.ackedqueue.io.ByteBufferCleaner;
import sun.misc.Unsafe;

class ByteBufferCleanerImpl implements ByteBufferCleaner {

    private final Unsafe unsafe;

    public ByteBufferCleanerImpl() throws Exception {
        Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
        unsafeField.setAccessible(true);
        unsafe = (Unsafe) unsafeField.get(null);
    }

    public void clean(MappedByteBuffer buffer) {
        unsafe.invokeCleaner(buffer);
    }
}
```

logstash-core/src/main/java/org/logstash/LogstashJavaCompat.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    /**
+     * Identifies whether we're running on Java 9 by parsing the first component of the
+     * {@code "java.version"} system property. For Java 9 this value is assumed to start with a
+     * {@code "9"}.
+     * @return True iff running on Java 9
+     */
+    private static boolean isJdk9() {
```
jakelandis: nit: `isAtLeastJava9` ?  It is not checking jdk vs. jre and returns true fro jre 10

logstash-core/src/main/java/org/logstash/LogstashJavaCompat.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Logic around ensuring compatibility with Java 8 and 9 simultaneously.
+ */
+public final class LogstashJavaCompat {
+
+    /**
+     * True if current JVM is a Java 9 implementation.
+     */
+    public static final boolean IS_JAVA_9 = isJdk9();
```
jakelandis: nit: ditto on naming about java 9 or greater.

logstash-core/src/main/java/org/logstash/LogstashJavaCompat.java
original-brownbear: @jakelandis sure why not :) Will do the renamings

original-brownbear:
@jakelandis renamed things :)

logstash-core/src/main/java/org/logstash/LogstashJavaCompat.java
```diff
@@ -0,0 +1,72 @@
...
+        imports.add("java.nio.MappedByteBuffer");
+        final String cleanerCode;
+        final String ctorCode;
+        final String fieldsCode;
+        if (isAtLeastJava9()) {
+            imports.add("sun.misc.Unsafe");
+            imports.add("java.lang.reflect.Field");
+            cleanerCode = "unsafe.invokeCleaner(buffer);";
+            ctorCode = "Field unsafeField = Unsafe.class.getDeclaredField(\"theUnsafe\");" +
+                "unsafeField.setAccessible(true);" +
```
jakelandis: I don't think reflection hacks are needed here. I _think_ the cleaner API has been replaced in Java 9 by the [Cleaner](https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Cleaner.html) object.

original-brownbear: @jakelandis Not sure what the `Cleaner` gets me here, I will still need to implement an appropriate `Runnable` for it to trigger and that will execute this exact action won't I?

jakelandis: I see now `Cleaner` is simply a newer/better version of finalize, still coming up to speed on Java9. 
Man is the history behind this quite long, led mostly by lucene https://issues.apache.org/jira/browse/LUCENE-6989 and long discussions on the mailing list http://mail.openjdk.java.net/pipermail/core-libs-dev/2016-January/thread.html#38243. 

Also a long discussion for Java 10 : https://bugs.openjdk.java.net/browse/JDK-4724038 

anyway... this change is identical in function (but not form) to Lucene so +1

Sorry for the noise.

jakelandis:
LGTM, nice work!

original-brownbear:
@jakelandis thanks!

PR #8569, opened 2017-11-02
#8442 Java Pipeline Compiler

original-brownbear:
This allows writing end to end pipeline tests in Java/JUnit for #8442:

* Adds a Java wrapper for the pipeline compiler so it can be invoked from Java
   * The also cleans up the Ruby side of things a little in terms of duplication (though obviously the Ruby-Java-Ruby step makes things a little trickier .. still imo this is the only way to get this functionality into Java tests that doesn't require replacing `treetop`)
* Adds an example JUnit test for it

Now for the trickier build adjustments needed :)

* Moved the RSpec tests and the new tests under a new Gradle goal `RubyTests` and adjusted CI scripting accordingly
  * To me, it seemed like the cleanest approach to do it like this since it still leaves us with only 2 separate sets of unit tests without adding another Gradle + Rake interaction (Note here, that you need to have your `vendor` dir bootstrapped correctly for the new test as well as the existing JUnit RSpec wrapper to work. Unfortunately, a self-contained Gradle target for the RSpec tests requires larger changes to the build ... I coded those up and they can be found in #8543, but I don't think they should be mixed with this one :))

### DIFF
--- a/ci/unit_tests.sh
+++ b/ci/unit_tests.sh
@@ -19,14 +19,14 @@ if [[ $SELECTED_TEST_SUITE == $"core-fail-fast" ]]; then
   rake test:core-fail-fast
 elif [[ $SELECTED_TEST_SUITE == $"java" ]]; then
   echo "Running Java unit tests"
-  echo "Running test:core-java"
-  rake test:core-java
+  echo "Running Java Tests"
+  ./gradlew javaTests
 elif [[ $SELECTED_TEST_SUITE == $"ruby" ]]; then
   echo "Running Ruby unit tests"
   echo "Running test:install-core"
   rake test:install-core
-  echo "Running test:core-ruby"
-  rake test:core-ruby
+  echo "Running Ruby Tests"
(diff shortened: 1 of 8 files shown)


original-brownbear:
@jakelandis I know this one still invokes RSpec via JUnit like #8543 did, but so far it seems like the shortest route towards stronger JUnit tests.

I agree with the downside that this changes the env a little, relative to CLI invocation of RSpec ... though that's a short-lived issue hopefully with the Java entrypoint incoming.
There is an upside to this though, we can write Java support code for RSpec tests and simply load it in the JUnit env without having to add more hacks to our RSpec helper scripts. We already have some of this code (`org.logstash.Util#getMapFixtureJackson` for example as well as the whole serialized in-memory queue logic) that we could clear out of the production codebase this way :)

jakelandis:
Spoke in person and @original-brownbear walked me through the changes here. Changes LGTM and is nice step towards Javaification.

PR #8577, opened 2017-11-03
PERFORMANCE: Java Exec Codegen for Filters 

original-brownbear:
Next step in the codegen saga :)

* Codegen more optimal filter execution
  * like in #8500 => no iterations over parent datasets, strongly types fields on generated code, low-level calls to JRuby
* Dried up the logic between filters and output datasets
  * Tried to make the field layout as similar across datasets as I could (also fixed offsets between different kinds of fields should allow merging multiple datasets into one in the next step easily)
* Added `join()` helper for readability (this keeps the pieces that get concatenated closest visually, very helpful imo ... but I could be weird here :D)


original-brownbear:
@jakelandis sorry for another step here, tried to at least improve the readability of this code some more. Happy to go over this one live again next week though if I failed at that :)

logstash-core/src/main/java/org/logstash/config/ir/compiler/DatasetCompiler.java
```diff
@@ -139,88 +231,54 @@ public static Dataset terminalDataset(final Collection<Dataset> parents) {
...
-        final int pluginIndex) {
-        return new StringBuilder().append(field(callsiteIndex)).append(
-            ".call(RubyUtil.RUBY.getCurrentContext(), ").append(field(pluginIndex))
-            .append(", RubyUtil.LOGSTASH_MODULE, \"multi_receive\", ")
-            .append(field(argArrayIndex)).append(", Block.NULL_BLOCK);").toString();
+    private static String callFilterFlush(final int offset, final boolean shutdownOnly) {
+        final String condition;
+        final String flushArgs;
+        if (shutdownOnly) {
+            condition = "flush && shutdown";
```
jakelandis: nit: use the flush constant above, and create one for shutdown.

original-brownbear: @jakelandis hmm I thought the same :D, but:

* this `flush` is a method parameter, the constant was using in place of the method name when invoking a Ruby call-site. I think reusing it here would make things more confusing since this spot (at least syntax-wise) has nothing to do with other uses of it.

That said ... you're right, this shouldn't have some magic `String` duplication to it. How about renaming the `flush` param to `doFlush` throughout the codebase (would just be in `Dataset` code)?

jakelandis:
It's a bit easier to consume, however example Java code snippets per private method would go a LONG way towards readability.

original-brownbear:
@jakelandis I like that suggestion :) I'll add some examples (probably gonna happen only tomorrow, but let's do it in this step imo).

original-brownbear:
@jakelandis hmpf hmpf hmpf :) I fear we may have to postpone adding better doc here :( 

This code will need a non-trivial refactoring for it to work without a global `Ruby` runtime constant for the Java entry-point => whatever doc I add will probably not survive this week.
Likely we'll need a change to the `Dataset` API here to avoid taking a hit from losing `RubyUtil.RUBY`.


If possible I'd still like to get this in, since it's a clear speedup and hence valuable checkpoint for the move to the Java entry-point (I have some fears about what moving away from the static `Ruby` will do to performance in some spots => would be nice to have `master` as a reference).
Also, it's easier to create strong for the non-constant solution `Ruby` if I have to make it work for filters + outputs and not just outputs.

Any objections to moving a little quicker on this one? Ping me if you want to Zoom for more details :)

jakelandis:
I am still struggling a bit to fully understand the line by changes presented here. (I just need to a decent chunk of time to debug through all of LIR stuff...) 

Given that this is behind a feature flag,  and is due for a re-factoring, I am good with getting this merged into master to unblock progress. 

Assuming this does not break the pipeline viewer, then LGTM. 

I will try to carve off some time in the near future to get a better base so I can provide better feedback.

original-brownbear:
@jakelandis yea sorry this is a bit of a confusing exercise :( But no worries the pipeline viewer is not affected by this (this thing consumes `PipelineIR` same as the viewer, but doesn't itself modify the `PipelineIR` => we're safe :)).

So good to merge if it's still green after a rebase (we gotta rebase and test again here with all the recent changes)? :)

jakelandis:
> So good to merge if it's still green after a rebase (we gotta rebase and test again here with all the recent changes)? :)

LGTM

In general anything behind a feature flag still in heavy development, imo, should only be _required_ to be lightly reviewed to the point where it doesn't leak from the feature flag. A final review (full) review prior to the feature flag removal.  My only concern is that personally I am not currently able to provide any meaningful feedback for this work in progress, and am currently only acting as a gatekeeper (and I need to fix that).

original-brownbear:
@jakelandis I'll try to create a piece of this that will be part of the end result without further changes (my mistake for iterating so "globally"). I hope that will make it easier and more interesting to work on this.

Issue #7416, opened 2017-06-12
Rake Bootstrap Should Refresh Gemfile and Gemfile.jruby-1.9.lock Unconditionally


original-brownbear:
I think I made a mistake when implementing the logic around `Gemfile.template` that allows for a clean git index across all build stages in #6985.

I made the logic only create the `Gemfile` in case it doesn't yet exist. This is wrong I think because whenever I change branches (like to `5.x`) or add/remove a dependency I need to manually delete the existing `Gemfile` and `Gemfile.jruby-1.9.lock` before running `rake bootstrap` (or some other task that has `bootstrap` as a dependency) to get a correct `Gemfile`.

I think simply turning:

(`bundler.rb`)

```ruby
      # create Gemfile from template iff it does not exist
      unless ::File.exists?(Environment::GEMFILE_PATH)
        FileUtils.copy(
          ::File.join(Environment::LOGSTASH_HOME, "Gemfile.template"), Environment::GEMFILE_PATH
        )
      end
```

into:

```ruby
      # create Gemfile from template
        FileUtils.cp_r(
          ::File.join(Environment::LOGSTASH_HOME, "Gemfile.template"), Environment::GEMFILE_PATH,
:remove_destination => true
        )
```

should fix this.

jakelandis:
@original-brownbear - can this be closed now that we have the ability to bootstrap via Gradle ?

original-brownbear:
@jakelandis oh yea, my bad for forgetting this one :)

PR #8583, opened 2017-11-04
TESTS: Improve Gradle Test Configuration

original-brownbear:
* `clean` actually cleans all dynamically created files, in particular, it now properly cleans the generated Gemfile so
that changes to the Gemfile.template reflect in a rerun of `bundler`
* `rubyTests` and `test` are now one-off and will automatically bootstrap JRuby and Gems required by the tests if necessary (fixes https://github.com/elastic/logstash/issues/7416 automatically for test targets at least)
* Fixed Readme to document the now much simpler test targets
* All rake tasks remain unchanged and still work exactly as they did before

--------

This should make the tests a lot more portable since they don't require any locally install Ruby to run now. Also, local Ruby setups should not interfere with the build as much as the used to (most prominent example here is the build not bootstrapping on JRuby `9.1.13.0` atm).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN /opt/logstash/gradlew wrapper
 ADD versions.yml /opt/logstash/versions.yml
 ADD LICENSE /opt/logstash/LICENSE
 ADD CONTRIBUTORS /opt/logstash/CONTRIBUTORS
-ADD Gemfile.template /opt/logstash/Gemfile
+ADD Gemfile.template /opt/logstash/Gemfile.template
 ADD Rakefile /opt/logstash/Rakefile
 ADD build.gradle /opt/logstash/build.gradle
 ADD rakelib /opt/logstash/rakelib
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ RUN ln -s /tmp/vendor /opt/logstash/vendor
 
 ADD gradlew /opt/logstash/gradlew
 ADD gradle/wrapper /opt/logstash/gradle/wrapper
+ADD buildSrc /opt/logstash/buildSrc
 RUN /opt/logstash/gradlew wrapper
 
 ADD versions.yml /opt/logstash/versions.yml
--- a/README.md
+++ b/README.md
@@ -126,20 +126,20 @@ Most of the unit tests in Logstash are written using [rspec](http://rspec.info/)
 
 ### Core tests
 
-1- In order to run the core tests, a small set of plugins must first be installed:
+1- To run the core tests you can use the Gradle task:
 
-    rake test:install-core
-
-2- To run the core tests you can use the rake task:
-
-    rake test:core
+    ./gradlew test
 
   or use the `rspec` tool to run all tests or run a specific test:
 
     bin/rspec
     bin/rspec spec/foo/bar_spec.rb
     
-3- To run the subset of tests covering the Java codebase only run:
+  Note that before running the `rspec` command for the first time you need to set up the RSpec test dependencies by running:
+
+    ./gradlew bootstrap
+
+2- To run the subset of tests covering the Java codebase only run:
     
     ./gradlew javaTests
 
--- a/build.gradle
+++ b/build.gradle
@@ -90,7 +90,23 @@ if (versionMap["jruby-runtime-override"]) {
 }
 
 // Tasks
+
+clean {
+  delete "${projectDir}/Gemfile"
+  delete "${projectDir}/Gemfile.lock"
+  delete "${projectDir}/vendor"
+  delete "${projectDir}/NOTICE.TXT"
+}
+
 task bootstrap {}
+
+project(":logstash-core") {
+  ["rubyTests", "test"].each { tsk ->
+    tasks.getByPath(":logstash-core:" + tsk).configure {
+      dependsOn bootstrap
+    }
+  }
+}
 
 task downloadJRuby(type: Download) {
     description "Download JRuby artifact from this specific URL: ${jRubyURL}"
@@ -108,6 +124,8 @@ task verifyFile(dependsOn: downloadJRuby, type: Verify) {
 
 task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
     description "Install JRuby in the vendor directory"
+    inputs.files file("${projectDir}/versions.yml")
+    outputs.files fileTree("${projectDir}/vendor/jruby")
     from tarTree(downloadJRuby.dest)
     eachFile { f ->
       f.path = f.path.replaceFirst("^jruby-${jRubyVersion}", '')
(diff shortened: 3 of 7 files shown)


original-brownbear:
@jakelandis @andrewvc hah untangled this mess at last :) Just pinged you guys for review since we talked about this a few times. Not sure how best to assign to the Gradle in general.

Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN /opt/logstash/gradlew wrapper
 ADD versions.yml /opt/logstash/versions.yml
 ADD LICENSE /opt/logstash/LICENSE
 ADD CONTRIBUTORS /opt/logstash/CONTRIBUTORS
-ADD Gemfile.template /opt/logstash/Gemfile
+ADD Gemfile.template /opt/logstash/Gemfile.template
```
andrewvc: Ah, yes, makes sense to add this back in :)

README.md
```diff
@@ -126,20 +126,20 @@ Most of the unit tests in Logstash are written using [rspec](http://rspec.info/)
...
 ### Core tests
 
-1- In order to run the core tests, a small set of plugins must first be installed:
+1- To run the core tests you can use the rake task:
 
-    rake test:install-core
-
-2- To run the core tests you can use the rake task:
-
-    rake test:core
```
andrewvc: Is this really the authoritative way to run tests now?

I'm worried that we have multiple entry points into a bootstrapped JRuby environment.

I'd feel more comfortable making `gradlew` the official way to run tests if `bin/logstash` shared the same environment bootstrap. I think we need #8161 for that no?

At that point we can make the `rake test:core` and other associated tasks alias the gradle tasks for completeness.

ci/unit_tests.sh
```diff
@@ -18,19 +18,12 @@ if [[ $SELECTED_TEST_SUITE == $"core-fail-fast" ]]; then
...
-  rake test:install-core
-  echo "Running Ruby Tests"
   ./gradlew rubyTests
 else
   echo "Running Java and Ruby unit tests"
-  echo "Running test:install-core"
-  rake test:install-core
-  echo "Running test:core"
-  rake test:core
+  ./gradlew test
```
andrewvc: So much cleaner!

README.md
original-brownbear: @andrewvc hah, I already had this discussion with @jakelandis around #8569 (unfortunately mostly offline, but a summary of my arguments is here https://github.com/elastic/logstash/pull/8569#issuecomment-341425538).

Sorry for not involving you there if you wanted to have a say in this because what happens today already is:

in the `namespace "test" do` rakelib section we have:

```rb
  task "core" => ["core-slow"]
```

and then:

```rb
  task "core-slow" do
    exit 1 unless system('./gradlew clean test')
  end
```

so `rake` is only wrapping the Gradle task anyways as a result of recent PRs. The only reason I left `rake` as the authoritative approach is that Gradle couldn't do the environment bootstrapping, now it can.

=> `./gradlew test` is what is ran in both cases 
=>  `./gradlew` now that it can bootstrap the environment is a lot more portable than the `rake` approach (`./gradlew` needs to work for `rake` to work, but `./gradlew` doesn't need `rake` since it bootstraps its own `rake`).

Sorry again for forcing this via https://github.com/elastic/logstash/pull/8569, but unless we want to revert the work in there `./gradlew` won this one already.

README.md
```diff
@@ -126,20 +126,20 @@ Most of the unit tests in Logstash are written using [rspec](http://rspec.info/)
 
 ### Core tests
 
-1- In order to run the core tests, a small set of plugins must first be installed:
+1- To run the core tests you can use the rake task:
```
jakelandis: did you mean gradle task ?

original-brownbear: @jakelandis hmm ... yes :D sec fixing :)

README.md
```diff
@@ -126,20 +126,20 @@ Most of the unit tests in Logstash are written using [rspec](http://rspec.info/)
...
 
   or use the `rspec` tool to run all tests or run a specific test:
 
     bin/rspec
     bin/rspec spec/foo/bar_spec.rb
     
-3- To run the subset of tests covering the Java codebase only run:
+  Note that before running the `rspec` command for the first time you need to set up the RSpec test dependencies by running:
+
+    ./gradlew installTestGems
```
jakelandis: can we just instruct to use `gradlew bootstrap` ?  It is easier to remember.

README.md
```diff
@@ -126,20 +126,20 @@ Most of the unit tests in Logstash are written using [rspec](http://rspec.info/)
```
jakelandis: ^^ (expand) should we update the paragraph above too ?

README.md
original-brownbear: @jakelandis we could yea, fixing :)

README.md
original-brownbear: @jakelandis hah, not sure :D probably in another PR. Though right now that statement still seems factually correct for the most part imo.

README.md
```diff
@@ -126,20 +126,20 @@ Most of the unit tests in Logstash are written using [rspec](http://rspec.info/)
...
+1- To run the core tests you can use the rake task:
 
-    rake test:install-core
-
-2- To run the core tests you can use the rake task:
-
-    rake test:core
+    ./gradlew test
 
   or use the `rspec` tool to run all tests or run a specific test:
```
jakelandis: Are there plans to be able to run a single rspec test via gradlew ?

original-brownbear: @jakelandis hmm not yet, but it seems like that would be pretty cool to have. Then we could drop supporting the `rspec` CLI tool eventually I guess?

README.md
jakelandis: To elaborate on @original-brownbear 's statement.  #8569 does indeed introduce a mismatch of how we run our unit tests vs. our production code.  

However, the arguments that sold me are 
a) the integration tests and acceptance tests which actually spin up instances continue to run JRuby first (the production way) 
b) any issues introduced by JRuby or Java first, wouldn't likely be caught in a unit test since the unit tests are (by definition) not run the production code runs, and almost always test logic or edge conditions (less likely to be effected)
c) at some point in time we need to make this change, the unit tests seems like right starting point, and now seems like the right time to help unblock the java pipeline testing, and start towards confidence of a Java entry point.

build.gradle
```diff
@@ -117,6 +135,22 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
...
+  inputs.files fileTree("${projectDir}/rakelib")
+  inputs.files file("${projectDir}/versions.yml")
+  outputs.files file("${projectDir}/Gemfile")
+  outputs.files file("${projectDir}/Gemfile.lock")
+  outputs.files fileTree("${projectDir}/vendor/bundle/gems")
+  outputs.files fileTree("${projectDir}/vendor/jruby")
+  commandLine './vendor/jruby/bin/jruby', "${projectDir}/vendor/jruby/bin/rake".toString(), "test:install-core"
+  standardOutput = new ByteArrayOutputStream()
+  ext.output = {
+    standardOutput.toString()
```
jakelandis: do we need to grab standardError too ?

build.gradle
```diff
@@ -117,6 +135,22 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
...
+task installTestGems(dependsOn: downloadAndInstallJRuby, type: Exec) {
+  workingDir projectDir
+  inputs.files file("${projectDir}/Gemfile.template")
+  inputs.files fileTree("${projectDir}/rakelib")
+  inputs.files file("${projectDir}/versions.yml")
+  outputs.files file("${projectDir}/Gemfile")
+  outputs.files file("${projectDir}/Gemfile.lock")
+  outputs.files fileTree("${projectDir}/vendor/bundle/gems")
+  outputs.files fileTree("${projectDir}/vendor/jruby")
+  commandLine './vendor/jruby/bin/jruby', "${projectDir}/vendor/jruby/bin/rake".toString(), "test:install-core"
```
jakelandis: I assume this is blocking call, which means we don't see any output until is complete ?  It would be_really_ nice to see incremental output.

jakelandis:
@original-brownbear - nice work, a couple minor questions/thought for future. LGTM, but please wait for @andrewvc too.

build.gradle
original-brownbear: @jakelandis hmm yea it would be, good point! Let me try something that will also address the stderr question :)

original-brownbear:
@jakelandis thanks :)

Also, made the output stream nicely and also capture stderr, see https://github.com/elastic/logstash/pull/8583/commits/9884127d5d49b5b3dedbc6ccded3642b74063471 :)

original-brownbear:
huh seems my cool `buildSrc` doesn't work without a Gradle daemon ... sec fixing :)

original-brownbear:
@jakelandis ok now https://github.com/elastic/logstash/pull/8583/commits/418e6d908e35eb323902e67c9c8219fc6c250045 should address the stderr and streaming the output issues both :)

My bad, simply forgot to add the `buildSrc` folder to the Docker image :)

jakelandis:
@original-brownbear - updates look great ! thanks for the streaming output.

original-brownbear:
@jakelandis np  :) => back to you @andrewvc :)

andrewvc:
I am now LGTM on this

PR #8586, opened 2017-11-05
MINOR: Improve Gradle rubyTests target to properly recognize changes in Ruby code

original-brownbear:
This target runs our actual Ruby, but currently only watches for changes in the Java `src` path as per Gradle defaults, so a change to Ruby sources will not cause `rubyTests` to run again unless you also hit `clean`.

=> fixed by adding Ruby source paths as inputs for `rubyTests`.

I also fixed the `test` target, this was running all tests, which meant it ran the `rubyTests` target without its config/dependencies and just worked accidentally.

=> fixed by making `test` itself not run any tests, and simply be the parent for both `rubyTests` and `javaTests`. Any other solution would require duplicating config between `rubyTests` and `test` as far as I understand Gradle.

### DIFF
--- a/logstash-core/build.gradle
+++ b/logstash-core/build.gradle
@@ -72,14 +72,21 @@ configurations.archives {
 
 task javaTests(type: Test) {
     exclude '/org/logstash/RSpecTests.class'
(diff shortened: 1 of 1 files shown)


jakelandis:
LGTM

PR #8588, opened 2017-11-06
BUILD: Fix gemspec_jars task to automatically run if necessary

original-brownbear:
This task was kind of annoying whenever we adjusted a dependency.
It should simply be treated as a step of the dependency setup that has to come before compiling Java imo. Without this, it's a constant pain to work across branches with separate dependency versions (e.g. if you want to try out upgrading a dep). Also, the build should simply be consistent here, we shouldn't have Java pull the correct dependencies every time, while Ruby requires manual intervention.

------------

Also fixed the weird full path naming on task types in the 3 tasks above, don't think that's worth a separate PR :)


logstash-core/build.gradle
```diff
@@ -49,9 +49,10 @@ task cleanGemjar {
 clean.dependsOn(cleanGemjar)
 jar.finalizedBy(copyGemjar)
 task gemspec_jars {
+    def gemspecJars = file("${projectDir}/gemspec_jars.rb")
+    outputs.files gemspecJars
```
jakelandis: Can you help me understand what outputs.files is doing here vs. the prior implementation ?

jakelandis:
Do you know if there was a compelling reason why this dependency was omitted when it was first introduced ?

logstash-core/build.gradle
original-brownbear: @jakelandis it ensures that we will rerun this task only if the dependencies change or that file was changed. Just makes the build faster + it's a nice way of documenting what actually happens here (imo).

original-brownbear:
@jakelandis I think when this was introduced the Gradle setup was still a pretty secondary thing and everything was triggered from `rake` => this was just left over. You know back in May, the JUnit tests weren't even run as part of the normal build https://github.com/elastic/logstash/pull/7099 :)

jakelandis:
@original-brownbear thanks for the explains - LGTM

original-brownbear:
@jakelandis thanks for the review !

PR #8591, opened 2017-11-06
MINOR: Remove JMH main EntryPoints

original-brownbear:
These entry points are completely useless, I only added them initially for comfort when running from Idea. The idea JMH plugin is able to run benchmarks without them though (didn't realize that at the time) => just dead code.


original-brownbear:
@jsvd thanks!

PR #8592, opened 2017-11-06
MINOR: Cleanup Java Test Code in Production Classpath

original-brownbear:
Thanks for #8517 there is no more need for keeping code used in `rspec` only on the production classpath.
=> Moved these two methods to the test classpath.


jakelandis:
LGTM

PR #8593, opened 2017-11-06
MINOR: Cleanup RubyClass loading some more

original-brownbear:
Same approach as in #8529, make classloading explicit to avoid circular dependencies + get full control over the load order (important for writing more JUnit tests that involve Ruby functionality).

As in other PRs like this before, not that the outer classes for some Ruby class definitions now become redundant. I left them in this PR to not mix a logical change with a huge formatting change and also to not break binary compatibility.


original-brownbear:
@andrewvc @jakelandis pinging you guys here, since we already went over this topic a few times before. Should be a trivial review though :)

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -1,7 +1,9 @@
 # encoding: utf-8
 
-require "jruby_acked_queue_ext"
-require "jruby_acked_batch_ext"
+# Force loading the RubyUtil to ensure its loaded before the WrappedAckedQueue class is set up in
+# Ruby since WrappedAckedQueue depends on Ruby classes that are dynamically set up by Java code.
+java_import org.logstash.RubyUtil
```
jakelandis: This _feels_ fragile. It it possible to create a `java_support.rb` (or the like, with documentation) that can be required where ever this type of import is needed.

logstash-core/src/main/java/org/logstash/RubyUtil.java
```diff
@@ -63,6 +80,20 @@
...
+            }
+        }, JrubyAckedBatchExtLibrary.RubyAckedBatch.class);
+        setupLogstashClass(
+            "AckedQueue", JrubyAckedQueueExtLibrary.RubyAckedQueue::new,
+            JrubyAckedQueueExtLibrary.RubyAckedQueue.class
+        );
+        setupLogstashClass(
+            "AckedMemoryQueue", JrubyAckedQueueMemoryExtLibrary.RubyAckedMemoryQueue::new,
+            JrubyAckedQueueMemoryExtLibrary.RubyAckedMemoryQueue.class
+        );
```
jakelandis: +1 to getting these guys together... however we are starting to hit the limit of reasonable static initializer. Any thoughts on increasing the readability here ? (I like what did with the setupLogstashClass)

original-brownbear: @jakelandis I think this will clear up as soon as this logic is not `static` anymore and simply part of a normal constructor chain. I'll do my best to prepare a PR that gets us away from the Ruby global constant soon (this is fairly important for a stable/maintainable Java entry point).

For now, I'm not so sure I want to dry this up more (or spread that logic out more), just to keep the effort for moving away from `static` and towards a Java entry-point lower (a bit of a weak argument I know, but I'd rather not start abstracting things here, when I don't know what hacky challenges lie ahead :D if that makes any sense).

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
original-brownbear: @jakelandis you're 100% correct here imo (this is stable though, I promise :D it triggers classloading `RubyUtil` in 100% of cases). The thing is we're already doing exactly this in `event.rb`, so I'd rather not mix in drying this up here.

**BUT** come to think of it, life might be even better here in the next step :) You know, I think we should simply put this line into just `environment.rb` and it will work 100% fine (provided https://github.com/elastic/logstash/pull/8594 is in).

It's probably not worth it cleaning this up now when https://github.com/elastic/logstash/pull/8594 is just around the corner and we can simply move this to `environment.rb` then right?

jakelandis: I think this is the 3rd time I have seen same fix, and we probably don't want to rely on environment.rb since unit tests don't go through that path. 

No concerns about the technical correctness... just the human ability to connect the dots for needing to import RubyUtil.  My suggestion is basically to create shim ruby file that basically does nothing but call this `java_import` and has some documentation for when to use it. 

It's just my opinion on readability and don't want to block progress.

logstash-core/src/main/java/org/logstash/RubyUtil.java
jakelandis: > as this logic is not static anymore and simply part of a normal constructor chain
🥇

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
original-brownbear: @jakelandis I think the `rspec` path also goes through `environment.rb`, this is what's in there:

rspec.rb
```ruby
require_relative "environment"
LogStash::Bundler.setup!({:without => [:build]})
require "logstash-core"
require "logstash/environment"

$LOAD_PATH.unshift(File.join(LogStash::Environment::LOGSTASH_CORE, "spec"))
```

?

jakelandis:
LGTM (only minor concerns around readability, some which will evolve with more Java)

original-brownbear:
@jakelandis thanks! Yea, as soon as we have the entrypoint in Java many, many hacks like the one criticized (or the whole magic around loading `.class` and `.jar` files) should/will simply disappear :)

PR #8595, opened 2017-11-06
Release notes for 5.6.4

andrewvc:
Covers the range `0a4bc...v5.6.3`

Also bumps the docs version.

### DIFF
--- a/docs/index-shared1.asciidoc
+++ b/docs/index-shared1.asciidoc
@@ -1,8 +1,8 @@
 :branch:                5.6
 :major-version:         5.x
-:logstash_version:      5.6.3
-:elasticsearch_version: 5.6.3
-:kibana_version:        5.6.3
+:logstash_version:      5.6.4
+:elasticsearch_version: 5.6.4
+:kibana_version:        5.6.4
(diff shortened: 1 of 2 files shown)


original-brownbear:
@andrewvc I think we could mention https://github.com/elastic/logstash/commit/e33678b5738941f04974efd8e5dde75c15def57c in the changelog as well right?

original-brownbear:
Other than the above changelog suggestion, LGTM :)

andrewvc:
@original-brownbear what would you put in the changelog entry there? Was that an actual bug, or was it a perf fix?

original-brownbear:
@andrewvc that was an actual bug. Setting page_size == queue_size caused the pipeline to simply stall.

Maybe just put: "Fixed a bug that caused a persistent queue configured to use a single page only not to process events." or something like that?

andrewvc:
@original-brownbear so, if a PQ was configured to use only one page it wouldn't process any events at all?

original-brownbear:
@andrewvc jup exactly.

andrewvc:
@original-brownbear how's it look now?

original-brownbear:
@andrewvc perfect => LGTM :)

PR #8600, opened 2017-11-07
tolerate different types for ssl.enabled module options

### LINKED ISSUE #8580: modules ssl.enabled settings expect string values only
Currently the `ssl.enabled` settings for the elasticsearch and kibana module properties [do a comparison with the `"true"` string to determine if ssl should be enabled](https://github.com/elastic/logstash/blob/v5.6.3/logstash-core/lib/logstash/elasticsearch_client.rb#L32)

This works well if the settings are configured through the command line where everything is a string. But if the module is configured through the yaml setting, like, for example:

```yaml
modules:
- name: netflow
  var.elasticsearch.hosts: "node01"
  var.elasticsearch.ssl.enabled: true
```

Then the setting will have a boolean value instead of string, which means that regardless of true or false, the ssl won't be enabled.

jsvd:
fixes #8580

### DIFF
--- a/logstash-core/lib/logstash/elasticsearch_client.rb
+++ b/logstash-core/lib/logstash/elasticsearch_client.rb
@@ -29,7 +29,9 @@ module LogStash class ElasticsearchClient
 
       ssl_options = {}
 
-      if @settings["var.elasticsearch.ssl.enabled"] == "true"
+      # boolean settings may be strings if set through the cli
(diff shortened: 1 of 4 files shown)


guyboertje:
Known `Test Dead Letter Queue` failure - unrelated.

tsg:
@jsvd Can you rebase this one on top of master to get the fix in?

PR #8602, opened 2017-11-08
LIR: merge hash attributes of same name to support legacy configurations

yaauie:
Backports #8597 to 6.0

> This is a bug so it is valid to be merged into 6.0
> --  [@andrewvc  on #8597](https://github.com/elastic/logstash/pull/8597#issuecomment-342633075)

### DIFF
--- a/logstash-core/lib/logstash/compiler/lscl.rb
+++ b/logstash-core/lib/logstash/compiler/lscl.rb
@@ -101,9 +101,21 @@ module LogStashCompilerLSCLGrammar; module LogStash; module Compiler; module LSC
         else
           [k,v]
         end
-      }.reduce({}) do |hash,kv|
-        k,v = kv
-        hash[k] = v
+      }.reduce({}) do |hash, kv|
+        k, v = kv
+        existing = hash[k]
+        if existing.nil?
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/compiler/lscl.rb
```diff
@@ -101,9 +101,21 @@ def expr_attributes
...
+          hash[k] = v
+        elsif existing.kind_of?(::Hash)
+          # For legacy reasons, a config can contain multiple `AST::Attribute`s with the same name
+          # and a hash-type value (e.g., "match" in the grok filter), which are merged into a single
+          # hash value; e.g., `{"match" => {"baz" => "bar"}, "match" => {"foo" => "bulb"}}` is
+          # interpreted as `{"match" => {"baz" => "bar", "foo" => "blub"}}`.
+          # (NOTE: this bypasses `AST::Hash`'s ability to detect duplicate keys)
+          hash[k] = existing.merge(v)
+        else
+          hash[k] = existing + v
```
yaauie: note: this also backports an intermediate fix that supports the similar behaviour when multiple Attributes with the same name produce `Array`s

PR #8603, opened 2017-11-08
Remove references to document_type in the docs

dedemorton:
Doc changes related to https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/691

Removes `document_type` from LS configuration examples in the static docs. The `document_type` is now set automatically to `doc`, which is consistent with what the Beats dashboards expect.

I've searched the static docs for `document_type` and `_type` and have found no other instances, so I'm assuming there is no additional impact to the static docs unless I'm overlooking something.

I'll open separate PRs to address any references to `document_type` in the plugin docs.


dedemorton:
Pushed additional changes for issues identified in #8606. Updated the examples because some changes have happened to the event structure. 

Also added an updated screen capture for the tutorial because it was easy for me to grab the screen while I was updating the example.

docs/static/advanced-pipeline.asciidoc
```diff
@@ -161,18 +161,20 @@ If your pipeline is working correctly, you should see a series of events like th
...
           "beat" => {
+            "name" => "My-MacBook-Pro.local",
         "hostname" => "My-MacBook-Pro.local",
-            "name" => "My-MacBook-Pro.local"
+         "version" => "6.0.0"
     },
-    "type" => "log",
           "host" => "My-MacBook-Pro.local",
+    "prospector" => {
+        "type" => "log"
```
acchen97: @dedemorton my point here was, didn't Beats, in general, start defaulting type => doc instead of log in 6.0? https://github.com/elastic/beats/pull/3757

acchen97: Unless the prospector "type" is different from the default "doc type". You may be understand it better than I do.

dedemorton: Yes, that's the difference. If you look at the full output, you'll see that the `_type` field is set to `doc` as expected. The `type` filed on line 174 is the prospector type.

acchen97: Makes sense, thanks!

acchen97:
@dedemorton LGTM, thank you.

acchen97:
We good to merge this one?

PR #8604, opened 2017-11-08
Move note to section about the type option

dedemorton:
Takes the note that was removed in https://github.com/logstash-plugins/logstash-input-beats/pull/267 and moves it to the doc about the `type` option. Now this content will appear in the correct place, but it's conditionally coded in the shared file so that it only shows up in the doc about the beats input plugin.

Note that the statement will appear in two places until the plugin docs for 6.0 are re-published because I didn't get the change into the plugin branch in time for 6.0.

### DIFF
--- a/docs/include/input.asciidoc
+++ b/docs/include/input.asciidoc
@@ -80,7 +80,6 @@ This can help with processing later.
   * Value type is <<string,string>>
   * There is no default value for this setting.
 
-This is the base class for Logstash inputs.
 Add a `type` field to all events handled by this input.
 
 Types are used mainly for filter activation.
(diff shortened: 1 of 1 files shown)


andrewvc:
It's still valid because we are still compatible with ES 5.x.

Also, ES 6.0 does support types, just with the limitation of one type per
index

We should possibly warn when we detect types being used with ES6.

andrewvc:
@dedemorton should we close this out? Is this change still something we want to do?

dedemorton:
Yes. For now, I'll just merge this PR to move the note. It basically moves the content to live under the `type` option (where it should have been all along...only back in the days of the pulled-from-code docs, there was no way to customize the text for common options). Now we can conditionally code stuff.

dedemorton:
(no need for approval because this PR simply moves existing content...doesn't add anything new).

Issue #8589, opened 2017-11-06
`rake bootstrap` is Broken on Windows


original-brownbear:
Just tried to run `rake bootstrap` on Windows and noticed it just goes into an infinite loop:

```
PS C:\Users\me\src\logstash> rake bootstrap                
 :downloadJRuby                                             
 :verifyFile                                                 
 :downloadAndInstallJRuby UP-TO-DATE                         
 :bootstrap UP-TO-DATE
 BUILD SUCCESSFUL in 5s                                     
 3 actionable tasks: 2 executed, 1 up-to-date               
 :downloadJRuby                                              
 :verifyFile                                                 
 :downloadAndInstallJRuby UP-TO-DATE                        
 :bootstrap UP-TO-DATE                                                                                                   
 BUILD SUCCESSFUL in 5s                                      
 3 actionable tasks: 2 executed, 1 up-to-date                
 :downloadJRuby                                              
 :verifyFile                                                 
 :downloadAndInstallJRuby UP-TO-DATE                         
 :bootstrap UP-TO-DATE                                                                                                   
 BUILD SUCCESSFUL in 6s  
```

Once #8583 is in, this should be a trivial fix.

original-brownbear:
just a note:

2 issues break this:

* `jruby.bat` needs to be invoked instead of the shell script when on Windows
* `z_rubycheck` rake file causes those infinite restarts, removing it and using the `bat` file in Gradle fixes things

... will repair tomorrow

original-brownbear:
Fixed via https://github.com/elastic/logstash/pull/8607

PR #8607, opened 2017-11-08
#8589 make Gradle bootstrap compatible with Windows

original-brownbear:
One obvious and one not so obvious fix:

* We need to invoke `jruby.bat` on Windows => made sure of that
* `z_rubycheck.rb` doesn't work properly on Windows (`jruby.bat` doesn't set up the Gem paths correctly which breaks the check against `Gem.ruby` (it's also broken because it uses `/` path separators, but no point in fixing that because of the `jruby.bat` issue)), but we don't need it to when calling from Gradle => just put our override for that file `USE_RUBY=1` in to turn the check off.

Now bootstrap + RSpec tests all pass fine on Windows again. Output of `./gradlew clean test`:

screen shot 2017-11-08 at 10 15 18

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -136,6 +136,9 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
     into "${projectDir}/vendor/jruby"
 }
 
+def jrubyBin = './vendor/jruby/bin/jruby' +
+  (System.getProperty("os.name").startsWith("Windows") ? '.bat' : '')
+
 task installTestGems(dependsOn: downloadAndInstallJRuby, type: Exec) {
   workingDir projectDir
   inputs.files file("${projectDir}/Gemfile.template")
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -136,6 +138,8 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
     into "${projectDir}/vendor/jruby"
 }
 
+def jrubyBin = './vendor/jruby/bin/jruby' + (SystemUtils.IS_OS_WINDOWS ? '.bat' : '')
```
jakelandis: I really like commons lang, but in this case it is probably unnecessary, this check can be done with a System property and avoid a dependency.

original-brownbear: @jakelandis true :) sec adjusting

original-brownbear:
@jakelandis commons-lang no more :)

jakelandis:
This is awesome !  No more Windows setup issues, and will allow us to remove some brittle setup on build machines. 

Tested fresh checkout on Windows 10 from both PS and CMD and both worked great!

LGTM

original-brownbear:
@jakelandis sweet thanks for testing this :) (I hope your windows VM is faster than mine and didn't take 36 min :D) => merging :)

PR #8609, opened 2017-11-08
MINOR: Cleanup RubyTimestamp Constructors

original-brownbear:
Obvious cleanup, all these redundant constructors are unused as shown by this still passing all tests just fine.

The motivation for this is to cut down on static use sites of the `Ruby` runtime and stuff like `RubyUtil.RUBY_TIMESTAMP_CLASS` though to move to a non-static `Ruby` environment for the Java entrypoint (that's a huge, huge change => do it step by step :)).

Also should give us slightly better inlining to only have one constructor on that thing and reference it via a `static`, method reference (trivial effect though :)).

### DIFF
--- a/logstash-core/src/main/java/org/logstash/RubyUtil.java
+++ b/logstash-core/src/main/java/org/logstash/RubyUtil.java
@@ -47,13 +47,10 @@ public final class RubyUtil {
     static {
         RUBY = Ruby.getGlobalRuntime();
         LOGSTASH_MODULE = RUBY.getOrCreateModule("LogStash");
-        RUBY_TIMESTAMP_CLASS = setupLogstashClass("Timestamp", new ObjectAllocator() {
-            @Override
-            public JrubyTimestampExtLibrary.RubyTimestamp allocate(final Ruby runtime,
-                final RubyClass rubyClass) {
-                return new JrubyTimestampExtLibrary.RubyTimestamp(runtime, rubyClass);
-            }
-        }, JrubyTimestampExtLibrary.RubyTimestamp.class);
+        RUBY_TIMESTAMP_CLASS = setupLogstashClass(
+            "Timestamp",
+            JrubyTimestampExtLibrary.RubyTimestamp::new, JrubyTimestampExtLibrary.RubyTimestamp.class
+        );
         RUBY_EVENT_CLASS = setupLogstashClass(
             "Event", JrubyEventExtLibrary.RubyEvent::new, JrubyEventExtLibrary.RubyEvent.class
         );
--- a/logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
+++ b/logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
@@ -33,25 +33,10 @@ public final class JrubyTimestampExtLibrary {
             super(runtime, klass);
         }
 
-        public RubyTimestamp(Ruby runtime, RubyClass klass, Timestamp timestamp) {
-            this(runtime, klass);
-            this.timestamp = timestamp;
-        }
-
-        public RubyTimestamp(Ruby runtime, Timestamp timestamp) {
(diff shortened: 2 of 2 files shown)


logstash-core/src/main/java/org/logstash/RubyUtil.java
```diff
@@ -47,13 +47,10 @@
...
-        RUBY_TIMESTAMP_CLASS = setupLogstashClass("Timestamp", new ObjectAllocator() {
-            @Override
-            public JrubyTimestampExtLibrary.RubyTimestamp allocate(final Ruby runtime,
-                final RubyClass rubyClass) {
-                return new JrubyTimestampExtLibrary.RubyTimestamp(runtime, rubyClass);
-            }
-        }, JrubyTimestampExtLibrary.RubyTimestamp.class);
+        RUBY_TIMESTAMP_CLASS = setupLogstashClass(
+            "Timestamp",
+            JrubyTimestampExtLibrary.RubyTimestamp::new, JrubyTimestampExtLibrary.RubyTimestamp.class
```
jakelandis: Can you help me understand how you can replace an 'ObjectAllocator' with a `JrubyTimestampExtLibrary.RubyTimestamp::new` ?

original-brownbear: @jakelandis sure, the API requires `ObjectAllocator` which is a functional interface.
=> you can use a lamba, that has the content of its only method `public IRubyObject allocate(Ruby runtime, RubyClass klazz)` as its body => since the body is only a single method call to `JrubyTimestampExtLibrary.RubyTimestamp::new` the lambda can be replaced by a reference to that method call.
The only reason I used an `ObjectAllocator` instead of the reference before, was that we had multiple constructors and on some older Java 8 the call is a lot slower in that case (I think there were cases were if there is more than one override for `rubyTimestampExtLibrary.RubyTimestamp::new` the lamdba wasn't cached but now it should compile to an `invokestatic` since we have only a single constructor/override).

jakelandis: OK, i see now. That took me way longer then I like to admit to understand that (even with the explanation). 

Thanks for the explains.

jakelandis:
For my own education: 

If a Ruby class constructed a RubyTimestamp directly using one of the removed constructors they would have to java_import and instantiate RubyTimestamp.new directly ?

original-brownbear:
@jakelandis 

> If a Ruby class constructed a RubyTimestamp directly using one of the removed constructors they would have to java_import and instantiate RubyTimestamp.new directly?

Yes, but note that they would then instantiate a Java object of type `RubyTimestamp`, not a `Logstash::Timestamp` and all the alias naming defined by `JRubyMethod` wouldn't work.

E.g.

```ruby
jruby-9.1.12.0 :003 > x = org.jruby.RubyFixnum.new(org.jruby.Ruby.getGlobalRuntime, 5)
 => 5 
jruby-9.1.12.0 :004 > x + 2
NoMethodError: undefined method `+' for 5:Java::OrgJruby::RubyFixnum
        from (irb):4:in `<eval>'
        from org/jruby/RubyKernel.java:1000:in `eval'
        from org/jruby/RubyKernel.java:1298:in `loop'
        from org/jruby/RubyKernel.java:1120:in `catch'
        from org/jruby/RubyKernel.java:1120:in `catch'
        from /Users/brownbear/.rvm/rubies/jruby-9.1.12.0/bin/irb:13:in `<main>'
jruby-9.1.12.0 :005 > 
```
:)

logstash-core/src/main/java/org/logstash/RubyUtil.java
original-brownbear: np :)

jakelandis:
LGTM

The chances of Ruby code using the removed constructors is very low, and would not supported anyway.  The other changes are like for like cleanup.

original-brownbear:
@jakelandis thanks!

PR #8611, opened 2017-11-08
CLEANUP: Dry up RubyAckedMemoryQueue and RubyAckedQueue

original-brownbear:
The only differences between `RubyAckedMemoryQueue` and `RubyAckedQueue` are in the `initialize` and `ruby_open` methods. 
The rest of the code was 100% identical => dried this up and in the same go also dealt with the follow-up problem for https://github.com/elastic/logstash/pull/8593#issue-271475044 and remove the now redundant classes `JrubyAckedQueueExtLibrary` and `JrubyAckedMemoryQueueExtLibrary`.


original-brownbear:
@robbavey pinging you here since you already reviewed the Queue settings stuff.
This should be pretty trivial to review I think, just diff the two deleted files and you'll see where this is coming from :)

original-brownbear:
@robbavey thanks :) => rebasing (just a classname change from #8613) + merging then

PR #8615, opened 2017-11-08
#8596 retry until Vagrant ssh bootstrapping finishes

### LINKED ISSUE #8596: Vagrant ssh-config should check that machine is up first
The build error here: https://logstash-ci.elastic.co/job/elastic+logstash+6.x+multijob-acceptance/label=metal,suite=debian/121/console

Is due to an attempt to ssh into a vagrant host before that machine is ready. We should loop waiting for the machine, then error if it is not ready instead.

The error is occuring here: https://github.com/elastic/logstash/blob/master/qa/Rakefile#L25

original-brownbear:
Fixes #8596 by making `:ssh_config` retry in case the first Vagrant command fails because ssh isn't up yet on the box.

### DIFF
--- a/qa/Gemfile
+++ b/qa/Gemfile
@@ -2,4 +2,5 @@ source "https://rubygems.org"
 gem "runner-tool", :git => "https://github.com/purbon/runner-tool.git"
 gem "rspec", "~> 3.1.0"
 gem "rake"
+gem "stud"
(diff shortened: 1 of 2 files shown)


jakelandis:
LGTM, thanks for fixing this.

original-brownbear:
@jakelandis thanks!

PR #8620, opened 2017-11-09
Upgrade to JRuby 9.1.14.0

original-brownbear:
Works fine for me locally + could (haven't tested this yet) give us Java 9 compatibility (at least for a subset of plugins).


original-brownbear:
@jsvd thanks :)

PR #8625, opened 2017-11-09
CLEANUP: Dry up RubyClass name handling in RubyUtil

original-brownbear:
Making `RubyUtil` a little less horrible :) We shouldn't duplicate the information about a `JRubyClass` `name` here, by having it in the annotation as well as the setup code explicitly.
=> Just read it from the annotation when setting up the JRuby class.


jakelandis:
A little less horrible indeed :) 

LGTM

original-brownbear:
@jakelandis thanks :)

PR #8628, opened 2017-11-09
support page capacity change on existing queue

### LINKED ISSUE #7581: Logstash fails to restart after changing `queue.page_capacity` value
- Version: 5.4.3 
- Operating System: Linux
- Config File: `queue.type: persisted`
- Steps to Reproduce:

TLDR; changing `queue.page_capacity` may lead to logstash being unable to start. The situation depends on whether any events were currently stored in on-disk queue. If you try to roll `queue.page_capacity` value change across logstash cluster you may end up breaking number of nodes. While reverting change will fix broken nodes it might break new ones.

Per @colinsurprenant request, followup for https://github.com/elastic/logstash/issues/7538 (rest of comment copy pasted)

- I did modify `queue.page_capacity` size on logstash instance processing data,
- I used non-zero values (usually `256mb`, `512mb` and `1024mb`) as test for it.

And then situation was as follows:
 
- EMPTY QUEUE: If nothing was stored in on-disk PQ at the moment of `page_capacity` size change, it was POSSIBLE to restart logstash without any issue. By "nothing" I mean `_node/stats` monitoring endpoint showing 0 under `pipeline.queue.events`

With example change from: 256 to 512mb:

PRIOR to change:
```
    "queue" : {
      "events" : 0,
      "type" : "persisted",
      "capacity" : {
```

<< actual change >>

```
[2017-06-28T16:21:57,852][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
```

```
    "queue" : {
      "events" : 0,
      "type" : "persisted",
      "capacity" : {
        "page_capacity_in_bytes" : 536870912,
```

However the size of page file remained at it's former value:

```
-rw-r--r-- 1 logstash logstash 250M Jun 28 16:20 /var/lib/logstash/queue/main/page.505
```

END RESULT: Logstash resumed all of the operations properly.

- SOMETHING IN QUEUE: However, if there only was anything stored in PQ then subsequent logstash restarts errored and logstash didn't start until I either:
a) reverted `queue.page_capacity` to its former value 
b) manually  removed contents of `/var/lib/logstash/queue` and let logstash to recreate it

PRIOR TO CHANGE:

```
    "queue" : {
      "events" : 22237,
      "type" : "persisted",
      "capacity" : {
```

<< actual change >>

```
[2017-06-28T16:44:16,848][ERROR][logstash.agent           ] Cannot create pipeline {:reason=>"Page file size 262144000 different to configured page capacity 536870912 for /var/lib/logstash/queue/main/page.505"}
```

END RESULT: Logstash not starting.

If it's not possible to not enforce `page_capacity` size on already created pages, then maybe it's just worth mentioning in docs that queue needs to be drained first to allow you to change it?

colinsurprenant:
This PR fixes #7581 

It removes the strict check for a configured page capacity to necessarily equal the the capacity of an existing page file. The problem with this is that a page capacity change in the configuration while there is an existing non-empty queue would prevent logstash from starting. 

It seems reasonable that someone would want to tweak a page capacity for new pages moving forward while old pages would still exist at their original capacity until all purged.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java
@@ -30,7 +30,7 @@ public class MmapPageIO extends AbstractByteBufferPageIO {
 
     @Override
     public void open(long minSeqNum, int elementCount) throws IOException {
-        mapFile(STRICT_CAPACITY);
+        mapFile();
         super.open(minSeqNum, elementCount);
     }
 
@@ -38,13 +38,12 @@ public class MmapPageIO extends AbstractByteBufferPageIO {
     // to reflect what it recovered from the page
     @Override
     public void recover() throws IOException {
-        mapFile(!STRICT_CAPACITY);
+        mapFile();
         super.recover();
     }
 
     // memory map data file to this.buffer and read initial version byte
-    // @param strictCapacity if true verify that data file size is same as configured page capacity, if false update page capacity to actual file size
-    private void mapFile(boolean strictCapacity) throws IOException {
+    private void mapFile() throws IOException {
         RandomAccessFile raf = new RandomAccessFile(this.file, "rw");
 
         if (raf.length() > Integer.MAX_VALUE) {
@@ -52,11 +51,7 @@ public class MmapPageIO extends AbstractByteBufferPageIO {
         }
         int pageFileCapacity = (int)raf.length();
 
-        if (strictCapacity && this.capacity != pageFileCapacity) {
-            throw new IOException("Page file size " + pageFileCapacity + " different to configured page capacity " + this.capacity + " for " + this.file);
-        }
-
-        // update capacity to actual raf length
+        // update capacity to actual raf length. this can happen if a page size was changed on a non empty queue directory for example.
         this.capacity = pageFileCapacity;
 
         if (this.capacity < MIN_CAPACITY) { throw new IOException(String.format("Page file size is too small to hold elements")); }
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/io/AbstractByteBufferPageIO.java
@@ -31,7 +31,6 @@ public abstract class AbstractByteBufferPageIO implements PageIO {
     public static final int WRAPPER_SIZE = HEADER_SIZE + SEQNUM_SIZE + LENGTH_SIZE + CHECKSUM_SIZE;
 
     public static final boolean VERIFY_CHECKSUM = true;
-    public static final boolean STRICT_CAPACITY = true;
 
     private static final Logger logger = LogManager.getLogger(AbstractByteBufferPageIO.class);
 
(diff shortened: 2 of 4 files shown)


logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java
```diff
@@ -30,33 +30,29 @@ public MmapPageIO(int pageNum, int capacity, String dirPath) {
...
     // to reflect what it recovered from the page
     @Override
     public void recover() throws IOException {
-        mapFile(!STRICT_CAPACITY);
+        mapFile();
         super.recover();
     }
 
     // memory map data file to this.buffer and read initial version byte
     // @param strictCapacity if true verify that data file size is same as configured page capacity, if false update page capacity to actual file size
```
jakelandis: can you remove the `@param` comment ?

colinsurprenant: ah yes :P good catch.

jakelandis:
Do you remember the original rationale for the `strictCapacity` setting ?

colinsurprenant:
@jakelandis Yes, I wrote it. The idea was that if the file size was different then the configured capacity it would indicated something is wrong but this was relatively early in the design and we did not take into account the fact that someone would want to change the page size on an existing queue.

logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java
```diff
@@ -30,33 +30,29 @@ public MmapPageIO(int pageNum, int capacity, String dirPath) {
 
     @Override
     public void open(long minSeqNum, int elementCount) throws IOException {
-        mapFile(STRICT_CAPACITY);
```
jakelandis: I think this change will create a dead variable in org.logstash.ackedqueue.io.AbstractByteBufferPageIO (I didn't see it used anywhere else)

jakelandis:
@colinsurprenant - thanks, that makes sense.  As evident by the new test it appears that the PQ has no issues processing pages of different sizes. 

other then the possible dead variable (very minor point).   LGTM

logstash-core/src/main/java/org/logstash/ackedqueue/io/MmapPageIO.java
colinsurprenant: oh - let me verify that, thanks.

colinsurprenant: I don't see which dead variable you refer to??

jakelandis: `org.logstash.ackedqueue.io.AbstractByteBufferPageIO#STRICT_CAPACITY` <- I don't see it used anywhere else.

colinsurprenant: oh right - thanks.

colinsurprenant:
thanks for the review @jakelandis - made minor adjustments, will rebase and merge.

colinsurprenant:
I am planning on merging in 6.x, wondering if we should backport in 5? unless anyone thinks it's a good idea, I won't merge it in 5 since this is not a bug per-se, more of an annoyance. opinion @jordansissel?

colinsurprenant:
merged into 6.x and master.

colinsurprenant:
created #8650 to followup on any required doc changes

Issue #7243, opened 2017-05-28
pipeline.batch settings ignored while persisted queue is enabled


shoggeh:
- Version: 5.4.0
- Operating System: Linux
- Config File:

`/etc/logstash/logstash.yml`

```
  pipeline.batch.size: "4096"
  pipeline.batch.delay: "5000"
```

`/etc/logstash/conf.d/logstash.conf`
```
input { beats {} } 
output { elasticsearch { idle_flush_time => 5, flush_size => 4096 } }
```
- Sample Data:

I've been testing persisted queue and noticed very worrying thing: during normal operation elasticsearch {} output doesn't seem to accumulate output into batches at all. Settings like `idle_flush_time` / `flush_size` / `pipeline.batch.size / pipeline.batch.delay` have no real effect on actual batches size.

With beats {} input it's very visible that instead of sending accumulated batches, es output only sends as many events as provided by single beat input. It literally just dumps whatever it gets from inputs in real time.

- `queue.type: persisted` - separate bulk operations per beat input stream, being sent in real-time

```
POST /_bulk HTTP/1.1
{"index":{"_id":null,"_index":"test","_type":"log","_routing":null}}
{"offset":1,"input_type":"log","source":"/var/log/testA.log","program":"test","message":"test1",@timestamp":"2017-05-28T22:39:20.092Z","beat":{"name":"hostAAA"}}
{"index":{"_id":null,"_index":"test","_type":"log","_routing":null}}
{"offset":2,"input_type":"log","source":"/var/log/testA.log","program":"test","message":"test2",@timestamp":"2017-05-28T22:39:20.092Z","beat":{"name":"hostAAA"}}
```
```
POST /_bulk HTTP/1.1
{"index":{"_id":null,"_index":"test","_type":"log","_routing":null}}
{"offset":1,"input_type":"log","source":"/var/log/testB.log","program":"test","message":"test3",@timestamp":"2017-05-28T22:39:20.092Z","beat":{"name":"hostBBB"}}
```

- `queue.type: memory` - accumulated bulk operation up to `flush_size` / `idle_flush_time` / `pipeline.batch.size` / `pipeline.batch.delay` values

```
POST /_bulk HTTP/1.1
{"index":{"_id":null,"_index":"test","_type":"log","_routing":null}}
[...]

PR #8632, opened 2017-11-09
use the timeout value as a hard time limit to fill a batch

colinsurprenant:
The new memory queue draining logic does not respect the idea of the `pipeline.batch.delay` configuration option which is documented as
```
# How long to wait before dispatching an undersized batch to filters+workers
# Value is in milliseconds.
```
Prior to this fix the public `drain()` method would reset the timeout after successfully reading elements from the queue and continue like this up to hitting the maximum number of elements requested. This means that the total time to perform the drain could be longer than the requested timeout. 

There were no specific tests harness around this new Java code and I did not add any either. For now I am comfortable to rely on the other tests since this is such a core function any problem with it would surface. Ideally I would want specific tests added around this method. Depending on the result of this PR I would like to followup with an issue to add tests.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
+++ b/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
@@ -31,10 +31,9 @@ public final class LsQueueUtils {
 
     /**
      * <p>Drains {@link JrubyEventExtLibrary.RubyEvent} from {@link BlockingQueue} with a timeout.</p>
-     * <p>The timeout will be reset as soon as a single {@link JrubyEventExtLibrary.RubyEvent} was
-     * drained from the {@link BlockingQueue}. Draining {@link JrubyEventExtLibrary.RubyEvent}
+     * <p>Draining {@link JrubyEventExtLibrary.RubyEvent}
      * stops as soon as either the required number of {@link JrubyEventExtLibrary.RubyEvent}s
-     * were pulled from the queue or the timeout value has gone by without an event drained.</p>
+     * were pulled from the queue or the timeout value has gone by.</p>
      * @param queue Blocking Queue to drain {@link JrubyEventExtLibrary.RubyEvent}s
      * from
      * @param count Number of {@link JrubyEventExtLibrary.RubyEvent}s to drain from
@@ -49,13 +48,23 @@ public final class LsQueueUtils {
         final BlockingQueue<JrubyEventExtLibrary.RubyEvent> queue, final int count, final long nanos
     ) throws InterruptedException {
         int left = count;
+        final long deadline = System.nanoTime() + nanos;
+        long drainTimeout = nanos;
+
         final Collection<JrubyEventExtLibrary.RubyEvent> collection =
             new HashSet<>(4 * count / 3 + 1);
+
         do {
-            final int drained = drain(queue, collection, left, nanos);
+            final int drained = drain(queue, collection, left, drainTimeout);
             if (drained == 0) {
                 break;
             }
+
+            drainTimeout = deadline - System.nanoTime();
+            if (drainTimeout <= 0) {
+                break;
+            }
+
             left -= drained;
         } while (left > 0);
         return collection;
@@ -78,17 +87,24 @@ public final class LsQueueUtils {
         final Collection<JrubyEventExtLibrary.RubyEvent> collection, final int count,
         final long nanos) throws InterruptedException {
         final long deadline = System.nanoTime() + nanos;
+        long pollTimeout = nanos;
         int added = 0;
         do {
             added += queue.drainTo(collection, count - added);
+            
             if (added < count) {
+                if (pollTimeout <= 0) {
+                    break;
+                }
                 final JrubyEventExtLibrary.RubyEvent event =
-                    queue.poll(deadline - System.nanoTime(), TimeUnit.NANOSECONDS);
+                    queue.poll(pollTimeout, TimeUnit.NANOSECONDS);
                 if (event == null) {
                     break;
                 }
                 collection.add(event);
                 added++;
+
+                pollTimeout = deadline - System.nanoTime();
             }
         } while (added < count);
         return added;


original-brownbear:
@colinsurprenant this changes the behavior of the in-memory queue (relative to `5.6`) significantly:

In `5.6`we do have this:

```rb
     def read_next
        @size.times do |t|
          event = @queue.poll(@wait)
          return if event.nil? # queue poll timed out

          @originals[event] = true
        end
      end
```

=> resetting the timeout after each successful poll.

Change the behavior like this is certainly also a performance hit for use cases like ES that are heavily penalized for under-full batches. Also, if you think about the way an `ArrayBlockingQueue` works (one lock for both read and write) which leads to a `drainTo` call on an empty Queue likely returning just a single `Event` when woken up, setting an overall timeout isn't as helpful when optimizing for throughput, not latency is it?

=> I'm not sure this is a good idea. Do you have any benchmarks showing an improvement in throughput to support this change?

logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
```diff
@@ -78,17 +87,24 @@ private static int drain(final BlockingQueue<JrubyEventExtLibrary.RubyEvent> que
         final Collection<JrubyEventExtLibrary.RubyEvent> collection, final int count,
         final long nanos) throws InterruptedException {
         final long deadline = System.nanoTime() + nanos;
+        long pollTimeout = nanos;
         int added = 0;
         do {
             added += queue.drainTo(collection, count - added);
+            
             if (added < count) {
+                if (pollTimeout <= 0) {
```
original-brownbear: @colinsurprenant this is without effect and just adds dead code + is slower. A negative timeout value on a `poll` makes it return without blocking => much better to make the non-blocking `poll` to find out that the timeout ran out than to `break` out and miss the chance to get an `Event` without blocking

logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
```diff
@@ -49,13 +48,23 @@ public static void addAll(final BlockingQueue<JrubyEventExtLibrary.RubyEvent> qu
...
+
         do {
-            final int drained = drain(queue, collection, left, nanos);
+            final int drained = drain(queue, collection, left, drainTimeout);
             if (drained == 0) {
                 break;
             }
+
+            drainTimeout = deadline - System.nanoTime();
+            if (drainTimeout <= 0) {
```
original-brownbear: @colinsurprenant same as in the other case of this kind of breakout (transitively negative timeouts will lead to  no blocking in a call to `drain` as well), it's just a waste of cycles compared to polling without blocking and breaking out on no `Event`s received.

colinsurprenant:
@original-brownbear I agree this is a change in behaviour from 5.6 but keep in mind that the queue polling behaviour/strategy has also evolved/changed across version and what we have in 5.6 does not necessarily represent the "correct" behaviour. So I'd rather focus on the merits or not of the proposed change.

I get your concern about batch size maximisation and it relates to the discussion in #8633. My concern is about behaviour expectations with respect to what the `pipeline.batch.delay` actually mean. In the end we can also decide to change the `pipeline.batch.delay` doc/explanation to reflect what we think is the best behaviour it should have. 

A few points to consider:

- Side effects of these changes will be visible only for the "always empty queue" (or faster outputs than inputs) scenarios and this is the only case where batch creation latency wound be relevant depending on your use-case.
- I tend to think it is easier to reason about `pipeline.batch.delay`  as being a overall maximum delay to fill a batch. It is easy to understand if batch size creation latency is important for your use-case.
- If we consider it as a maximum timeout to fill a batch as proposed here:
  - using a larger value will work toward batch size maximization.
  - using a smaller value will work for wait time minimization and batch creation latency control.
- If we consider it as a per-read & resettable timeout in the batch read loop:
  - using a larger value will work toward batch size maximization but could result in delays multiple times the original `pipeline.batch.delay` value
  - using a smaller value will still work toward batch size maximization but with a somewhat lower but still undefined overall batch creation latency. 

We can probably take 2 directions here: 
- find agreement on the more desirable strategy and then validate with benchmarks, especially for use cases such as #7243
- create benchmarks anyways to help drive the decision.

logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
colinsurprenant: well, the behaviour for polling with a negative timeout is undefined at the API level, in the implementation (which could change) it will first try drain the queue regardless of the timeout value and then exit on <= 0 timeout. So if we want to be correct WRT the timeout then we should not call poll() when timeout is exhausted.

logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
colinsurprenant: well, if these cycles serve the purpose of being correct WRT to the timeout strategy they are not wasted.

colinsurprenant:
Lets move the design discussion in #8633 then come back here upon agreement.

jakelandis:
@colinsurprenant / @original-brownbear  is there consensus that in practice, due to a 5ms default `pipeline.batch.delay` that this proposed change would almost always result in smaller batches then the existing implementation ?

colinsurprenant:
@jakelandis yes, that is why in #8633 I suggested to increase that default value. Let's move the conversation in #8633.

andrewvc:
There's some history here.

I'm the original creator of this option, and as @original-brownbear points out, it was always just the value passed to `poll`, originally on a `SynchronousQueue`. This was intentional behavior to maximize the size of batches. Whether this is the best approach here is another discussion, but it is the intended result. 

The document text cited is from `pipelines.yml` comments. In our docs we describe this correctly: https://github.com/elastic/logstash/blob/master/docs/static/running-logstash-command-line.asciidoc . `pipelines.yml` did not exist at the time I added the document, so I think whoever added that line didn't do it correctly. This option has never worked that way AFAIK.

The documentation that is (almost) correct, from the linked asciidoc is:

> When creating pipeline batches, how long to wait while polling for the next event. This option defines how long in milliseconds to wait while polling for the next event before dispatching an undersized batch to filters and workers. The default is 250ms.

The reason I say that its only almost correct is that the actual value is 5ms. 

The documentation that is actually correct, which I wrote is here: https://www.elastic.co/guide/en/logstash/current/tuning-logstash.html which states:

> The pipeline.batch.delay setting rarely needs to be tuned. This setting adjusts the latency of the Logstash pipeline. Pipeline batch delay is the maximum amount of time in milliseconds that Logstash waits for new messages after receiving an event in the current pipeline worker thread. After this time elapses, Logstash begins to execute filters and outputs.The maximum time that Logstash waits between receiving an event and processing that event in a filter is the product of the pipeline.batch.delay and pipeline.batch.size settings.

So, that's history.

There's no regression here unless I'm missing something, though we clearly have some doc consistency issues.

WRT batch maximization vs. latency, I agree that this issue isn't the place to have that discussion, and I'm glad to continue this discussion there: https://github.com/elastic/logstash/issues/8633

Whatever we decide, however, the only thing we should change in 5.x and 6.x is fixes to our docs.  There are no version targets here, but I want to be sure we're clear here. Changing an option like this is a pretty drastic thing to do in a non-major release. We could do it, but there'd have to be a pretty big reason to do so without creating a new opt-in queue type.

colinsurprenant:
@andrewvc sure, we should make sure the `logstash.yml` doc reflect the intent of the option. This is usually the first (and probably only) place users will get the explanation of individual options. 

WRT to changes vs versions, this is off topic but there is definitely no target here and **if** we think this is a reasonable change we'll have this discussion. But also bare in mind that we do have history for doing such «pretty drastic thing to do in a non-major release». Not saying this is the best practice or desirable but we have done that in the past. Agreed it would have to be a «pretty big reason».

«creating a new opt-in queue type» this might be a good idea actually.

colinsurprenant:
Closing this issue. Per #8633 we decided to keep the current eager reading and batch maximizing  behaviour.

colinsurprenant:
Followup PR in #8655

PR #8634, opened 2017-11-10
correct the CentOS 6's service system

eailfly:
correct the CentOS 6's service system, CentOS 6 use sysv, not upstart


jordansissel:
CentOS 6 ships with upstart.

eailfly:
sorry, my bad. I should do some research

jordansissel:
Hey no worries. I am glad you brought this up ;)

Red hat doesn’t make it very obvious, and honestly, I am not sure our if
targeting upstart is the right choice for our users — you are not alone in
suggesting sysv.

PR #8635, opened 2017-11-10
Upgrade to JRuby 9.1.14.0

original-brownbear:
backport of https://github.com/elastic/logstash/pull/8620

### DIFF
--- a/versions.yml
+++ b/versions.yml
@@ -3,8 +3,8 @@ logstash: 6.0.0
 logstash-core: 6.0.0
 logstash-core-plugin-api: 2.1.16
 jruby:
-  version: 9.1.13.0
-  sha1: 815bac27d5daa1459a4477d6d80584f007ce6a68
+  version: 9.1.14.0
+  sha1: 56cd68e5ac1324ab56e2e86a57a0154f8142bde2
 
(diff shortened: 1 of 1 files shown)


jsvd:
Is there a need to make this change for the 6.0.x timeline? Any known bugfixes that impact us?

original-brownbear:
@jsvd there are a few generally applicable fixes (as in just plain Ruby fixes) + It keeps us in sync across versions => I'd go with upgrading and don't see that much of a risk (if there were big bugs in `9.1.14.0` I'd expect to see them in the JRuby bug reports). But I can go either way here ... def no "need" to do this :)

jsvd:
@original-brownbear yeah we have been burned in the past with patch releases of jruby introducing bugs (it's natural that it happens) and forcing us to re-release with a jruby downgrade, so I'd be more conservative in upgrading.

> if there were big bugs in 9.1.14.0 I'd expect to see them in the JRuby bug reports

It's not uncommon for us to be the first ones to detect an issue as we're one of the biggest users of JRuby

original-brownbear:
@jsvd alright, I'll trust your experience then :) Thanks for catching this!

PR #8639, opened 2017-11-10
Cleanup: Move classloading RubyUtil to a single entrypoint in Ruby code

original-brownbear:
Now that we don't have any circular dependencies between Java and Ruby anymore we should simply (class-)load `RubyUtil` as soon as we have our `.jar` or `.class` files on the path. This is achieved here by just loading it right after the `.jar` or `.class` files.


original-brownbear:
@jakelandis how about this for a solution to loading `RubyUtil`? :)

jakelandis:
@original-brownbear - it's beautiful! 

LGTM

original-brownbear:
@jakelandis happy to hear that :D thanks :)

Issue #8618, opened 2017-11-08
Centralized pipeline management doc enhancements


acchen97:
In the central pipeline management docs, we need to add a blurb around secure access to the [pipeline management UI](https://www.elastic.co/guide/en/logstash/6.x/logstash-centralized-pipeline-management.html#_pipeline_management_ui). With x-pack installed, we bootstrap a "logstash_admin" role which can be assigned to users that should have permissions to edit LS pipelines.

Additionally, @dedemorton should we consider renaming the parent section of this to "Managing Logstash"? It seems more consistent with the other "Configuring/Deploying/Scaling/Monitoring LS" sections.

Thanks to @ajpahl1008 for raising this.

dedemorton:
+1 for renaming to "Managing Logstash." That was actually my first impulse, but for some reason I thought you and Suyog wanted "configuration management" to appear in in the title. :-) I am happy to make the change because I love parallelism.

PR #8641, opened 2017-11-10
Change section title for config management

dedemorton:
Changes the section title as requested in issue #8618 

The other changes for #8618 will be in a different PR submitted into the xpack logstash repo.

### DIFF
--- a/docs/static/config-management.asciidoc
+++ b/docs/static/config-management.asciidoc
@@ -1,5 +1,5 @@
 [[config-management]]
-== Configuration Management
+== Managing Logstash
 
 Logstash provides configuration management features to make it easier for you to
 manage updates to your configuration over time.


docs/static/config-management.asciidoc
```diff
@@ -1,5 +1,5 @@
 [[config-management]]
-== Configuration Management
+== Managing Logstash
```
acchen97: @dedemorton should the URL resource (double brackets above) be changed to "management" instead?

dedemorton: @acchen97 It's better not to change the ID because it's used for linking and HTML file names. If we change the ID, we have to ask the web team to set up redirects and we have to update all the references in the docs. And possibly backport the change so switching between releases in the docs is seamless (that means even more redirects). :-P Definitely should avoid changing IDs unless we have a problem to fix (like exposing old product names...that sort of thing).

acchen97: understood :)

acchen97:
LGTM

PR #8644, opened 2017-11-12
MINOR: Remove dead code from wrapped in-memory-queue

original-brownbear:
Top to bottom:

* `java.util.HashSet` is not used in this file anymore
* `push` and `poll` are never called, that's what we have the read and write clients for
* We always only used the reference to the `ArrayBlockingQueue` in nested classes via `@queue.queue` => no point in that, just pass the actual `ArrayBlockingQueue` to nested classes' constructors and safe some indirection


original-brownbear:
@jsvd thanks!

PR #8645, opened 2017-11-12
MINOR: Better Typesafety and Deadcode Removal from Pipeline Compiler

original-brownbear:
The removed code is just dead code that I forgot to remove at some point. There is no consumer for them anywhere in the Java exec.

The change away from explicit `String` to `Class` for importing Java classes into the codegen logic should give us some more type safety and prevents weird mistakes (that are a little tricky to debug at times with the way Janino errors out).


jakelandis:
LGTM

PR #8646, opened 2017-11-13
[DOCS] Clarify ES response codes that trigger DLQ

joshuar:
Small doc change to clarify which Elasticsearch response codes will cause DLQ to be used.


jordansissel:
LGTM

PR #8647, opened 2017-11-13
MINOR: Remove Redundant Methods and Method Calls from both Queue Implementations

original-brownbear:
* `current_inflight_batch` is never called anywhere, grepping the codebase for it doesn't return anything
  * That method seems really wrong anyways, defaulting to an empty array in place of a `Batch` (I guess it was once like this to allow iteration over the default)
* `start_clock` and `stop_clock` can both be inlined into their respective callers, saving us a few method calls and more importantly, saving the rather expensive `Thread.current` call to Java (neither of these matter much performance wise, but the simpler this code is, the easier porting it to Java becomes)

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -149,10 +149,6 @@ module LogStash; module Util
         end
       end
 
-      def current_inflight_batch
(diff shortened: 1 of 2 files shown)


jakelandis:
Logic is like for like, and I too could not find anywhere that `current_inflight_batch` was called.

LGTM

original-brownbear:
@jakelandis thanks!

Issue #8637, opened 2017-11-10
Documentation show wrong image version


ebuildy:
Document here (https://www.elastic.co/guide/en/logstash/current/docker.html) is talking about 5.6.5 but:

> ➜  elk git:(master) ✗ docker pull docker.elastic.co/logstash/logstash:5.6.5
> Error response from daemon: manifest for docker.elastic.co/logstash/logstash:5.6.5 not found

robbavey:
Fixed by #8638

Issue #8114, opened 2017-08-31
Expand `rake version:sync` to also update asciidoc variables and the README.md


ph:
Update the documentation and the readme file is done manually by a human and this is error prone.
We already have a rake task that take care of updating the version of the gems and also the dependencies requirements called `rake version:sync`.

We can expand this task to also take care of updating the following:

- [ ] The links in the [README.md](https://github.com/elastic/logstash/blob/master/README.md)
- [ ] The shared variables in the (asciidoc)[https://raw.githubusercontent.com/elastic/logstash/master/docs/index-shared1.asciidoc

bbvch13531:
Hello! I searched logstash through hacktoberfest. Also I'm interested in this issue.
The version of gemfile.jruby `Gemfile.jruby-2.3.lock.release` in  [/rakelib/bump_plugin_versions.rake](https://github.com/elastic/logstash/blob/master/rakelib/bump_plugin_versions.rake) 
might be changed. 
Because when gemfile.jruby is updated, It should be updated too. 

But I'm new in ruby...  would you like to let me know the keyword to implement it please?

PR #8648, opened 2017-11-13
Split shared doc version update into separate rake task

robbavey:
The stack version constant in index-shared1.asciidoc should not be
updated during a version bump, as this impacts the current version
of the doc on elastic.co. Instead, create a separate rake task
to bump the version.

See #8637 for details of this happening

### DIFF
--- a/rakelib/version.rake
+++ b/rakelib/version.rake
@@ -22,10 +22,11 @@ def update_version_file(old_version, new_version)
   IO.write(VERSION_FILE, versions_as_text)
 end
 
-def update_index_shared1(old_version, new_version)
+def update_index_shared1(new_version)
   index_shared1 = IO.read(INDEX_SHARED1_FILE)
+  old_version = index_shared1.match(':logstash_version:\s+(?<logstash_version>\d[.]\d[.]\d.*)')[:logstash_version]
   %w(logstash elasticsearch kibana).each do |field|
-    index_shared1.gsub!(/(:#{field}_version:\s+)#{old_version['logstash']}/) { "#{$1}#{new_version['logstash']}" }
+    index_shared1.gsub!(/(:#{field}_version:\s+)#{old_version}/) { "#{$1}#{new_version}" }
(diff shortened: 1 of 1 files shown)


dedemorton:
I didn't realize that we were automating this (we've done it manually in the past). We've talked about the need to have some automation to bump doc versions (see https://github.com/elastic/logstash/pull/8648 and https://github.com/elastic/dev/issues/794), but as you've discovered, the docs cannot be bumped with the product versions without breaking the docs for existing users. I think one automated solution for all the docs is better than each team following a separate process.

robbavey:
@dedemorton We had a recent community submission (which I ok'd) to add this update and readme.md updates to the rake set_version/sync task - based on #8114, which in hindsight was not the correct behaviour.

This commit allows for the same functionality as described in https://github.com/elastic/dev/issues/794, and once teams agree to and converge on similar syntax for updating versions and doc versions should be straightforward to switch.

If we want to keep doc updated manually, then at the very least, we need to remove the linking of the doc file version updates from the main version update rake task.

robbavey:
jenkins test this please

robbavey:
Following from this morning's conversation, I've put the logic back to update the version number in the README.md file during the initial version bump. This looks to be consistent with what Kibana does, with the additional change in #8658 to remove the hardcoded reference to `master` in the README.md. 

One caveat is that tagged versions of Logstash will point to a non-existent locations, but trying to click on a link to a snapshot version of a released version seems like a pretty strange thing to do anyway... This behaviour is consistent with Kibana

Beats does things slightly differently - the beats versions point to a general navigable s3 bucket, rather than a specific location on elastic.co, and elasticsearch doesn't appear to advertise snapshot/nightlies AFAICT.

andrewvc:
@robbavey can you add some version targets to this?

robbavey:
@andrewvc Now that 6.0 is out, does it make sense to merge this to 6.0?

PR #8652, opened 2017-11-13
Fix Maven resolver URL

original-brownbear:
let's see if the new Gradle maybe has the new URL schema here


original-brownbear:
let's wait for https://issues.sonatype.org/browse/MVNCENTRAL-2870 not worth the hack ...

Issue #8633, opened 2017-11-10
[discuss] batch size maximisation vs batch read time minimisation


colinsurprenant:
In the PQ design stage we discussed in #5652 two different strategies for building batches:
- The first was to minimize wait time and return as soon as possible with some data, potentially issuing a small batch in the case where outputs are faster then inputs
- While the other was to maximize the batch size within the given timeout period. The timeout period would be used to its limit to maximize the batch size.

We decided to go with the wait time minimization with the side effect of building small batches.

We have two issues related to that here:

1. The batch building semantics is currently different between the Persistent Queue (PQ) implementation and the Memory Queue (MQ) implementation.  PQ uses the above strategy of wait time minimization while the MQ uses the batch size maximization, see [LsQueueUtils](https://github.com/elastic/logstash/blob/130fcf46a0514700254db69019e7294900db746c/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java#L32-L95) and related fix in #8632. 

2. As discussed in #7243 this semantics difference between PQ and MQ result in very different behaviours in outputs plugins like the ES output, which uses the batch object to bulk send to its output. When using PQ and when outputs are faster than inputs, small batches are created which results in small bulk output and this is a non-optimal side effect. 

I would like to suggest the following:

- We should unify the batch building semantics between MQ and PQ.
- I think I'd go with unifying toward batch size maximization, similar to what MQ does.
- Using the batch size maximization semantics (and when properly implemented) one can then leverage the `pipeline.batch.delay` option to control batch creation latency. A low value would provide a behaviour similar to the *minimize wait time* semantic and a higher value would give more time to maxime size toward `pipeline.batch.size`.
- A 5ms default for `pipeline.batch.delay` seems low, I don't have any hard evidence that it is but it feels arbitrary low and I'd look into increasing it to 50ms or 100ms if we want to tend toward batch size maximization

original-brownbear:
I commented on https://github.com/elastic/logstash/pull/8632:

* I wouldn't change the semantics of the in-memory queue and as you suggest here, move PQ to do what MQ does (the linked issue https://github.com/elastic/logstash/issues/7243 is an example of why the MQ semantics fit our use cases better).
  * The only gain of minimizing wait times is lower latency, in all other respects the approach is slower and more costly in terms of memory and CPU (more lock contention + GC)
* I also benchmarked the hard wait time approach at one point when implementing the set of speedups I implemented for the MQ and it always came out behind the current approach
[...]

PR #8655, opened 2017-11-13
replace decreasing timeout with fixed timeout

colinsurprenant:
Followup of #8632 and related to #8633 I think the decreasing timeout is unnecessary WRT to the eager reading & batch maximization strategy. We agreed that the idea of the timeout is that it should be applied per blocking read so fixing a global & fixed timeout *deadline* here should not be required. Also this avoids a call to `System.nanoTime()` in the read loop.

Let me know if I am missing something here @original-brownbear ?

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
+++ b/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
@@ -77,13 +77,12 @@ public final class LsQueueUtils {
     private static int drain(final BlockingQueue<JrubyEventExtLibrary.RubyEvent> queue,
         final Collection<JrubyEventExtLibrary.RubyEvent> collection, final int count,
         final long nanos) throws InterruptedException {
-        final long deadline = System.nanoTime() + nanos;
         int added = 0;
         do {
             added += queue.drainTo(collection, count - added);
             if (added < count) {
                 final JrubyEventExtLibrary.RubyEvent event =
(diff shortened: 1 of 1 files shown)


original-brownbear:
@colinsurprenant valid simplification imo :) LGTM

andrewvc:
So glad to see these things being aligned! 

Nice work improving the consistency of behavior here @colinsurprenant and @original-brownbear 

Don't forget to add version tags to this issue BTW!

colinsurprenant:
merged in master and 6.x. thanks @original-brownbear  @andrewvc

andrewvc:
I believe this should be in 6.0.1 / 5.6.5 since this is a bugfix no?

original-brownbear:
@andrewvc nah this code doesn't even exist in those versions :) Also, this has close to zero behaviour impact. The principle behaviour is completely unchanged by this (all that changes is that we don't do a needless `nanotime()` call) because the timeout is reset on the outer loop anyways.

andrewvc:
@original-brownbear ah, my misunderstanding, I thought the erroneous behavior was old. So the 5.x and 6.0 PQ do behave this way already?

original-brownbear:
@andrewvc no the PQ in all versions has a hard timeout and we need to fix it.

andrewvc:
@colinsurprenant do we have an open issue for that? I believe you mentioned something about work on that when we chatted yesterday?

colinsurprenant:
@andrewvc yes, following up in #8674 for PQ batch maximization.

PR #8659, opened 2017-11-13
 Secret Store: SecretStoreFactory, SecureConfig, Obfuscation and X-JVM support

jakelandis:
* Introduce a SecretStoreFactory to allow runtime definition of SecretStore implementation.
* Introduce a SecureConfig to allow simple configuration of different SecretStore implementaiton.
* Introduce random default password plus obfuscation. Best attempt at security through obscurity.
* Corrections / better support for x-JVM modification.

Part of #8353

Note - not included here:
* Any means to populate the SecureConfig based YAML settings
* Any command line tooling to work with the SecretStore
* Any support for configuration or settings in the context of Logstash


jakelandis:
@jordansissel / @yaauie 

This is very much part 2 of https://github.com/elastic/logstash/pull/8566. At the core not much has changed from part 1, just better support for areas highlighted in the commit log. Feel free to reach directly and I can walk through some the changes and motivation.

logstash-core/src/main/java/org/logstash/secret/store/SecretStoreUtil.java
```diff
@@ -91,9 +103,59 @@ private SecretStoreUtil() {
...
+     * <p>Simple obfuscation that adds a bit of randomness and shuffles the bits of a char[].</p>
+     * <p>Note - this is NOT security and will only deter the lazy.</p>
+     *
+     * @param chars The chars to obscure
+     * @return the obscured bytes
+     */
+    public static char[] obfuscate(char[] chars) {
+        byte[] bytes = asciiCharToBytes(chars);
+        byte[] random = new byte[bytes.length];
+        new Random().nextBytes(random);
```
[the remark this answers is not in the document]
jakelandis: nope, i can move it's declaration up to static scope.

jakelandis:
> going this deep into implementation before solidifying interface

In general, I agree.  However the user interface is dictated by [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/secure-settings.html) , and I did implemented enough of the command line to understand the direction before going this deep. That will be the next review. 

> example command-line incantations to persist and clear keys, to get metadata about the values, etc.;

Near exact clone of [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/secure-settings.html)

> example usage of a secret in a logstash pipeline config

Haven't settled on the exact syntax, but it will be something like `$secret.foo`

>example of what it looks like when a config attempts to use a secret that isn't populated.

I haven't gotten that far yet, however, the value from the API/Store will be null/nil which will be how to tell if the value isn't populated. 

It is quite possible, likley infact, that this will need to change with use in practice, but such as with all software.  It's really a matter of do you build software wide then deep, or deep then wide?  In this case I wanted to build deep first since, for me the ability to be OK with the security parts was my biggest concern. I want to build something correct, and more then just theater given the existing constraints. The other wide parts feel more well defined.

jakelandis:
Added another commit https://github.com/elastic/logstash/pull/8659/commits/c04b9de2febab0d58ece37657c8fc15ddca815b8 to no longer new Random everytime, and remove support for configuration of secret store password via system property. System properties can be seen via `ps` and in retrospec is a very bad place to put secrets.

logstash-core/src/main/java/org/logstash/secret/store/SecureConfig.java
```diff
@@ -0,0 +1,78 @@
+package org.logstash.secret.store;
+
+import java.nio.CharBuffer;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A String -> char[] container that holds a referenced char[] obfuscated in memory and allows for easy clearing all values from memory.
```
jordansissel: I'm imaginging the purpose of this class will change very quickly if we have multiple SecureStore implementations. For example, if we have a networked one (talking to Vault or Elasticsearch?), we'll need connection details in addition to credentials. 

No change required today, but do you envision this SecureConfig map-like construct being used for all of this (especially as all values are char[]) ?

jakelandis: I already kinda have that issue since I need to push the `keystore.path` through this class.  While it is a bit awkward to push non-sensitive stuff through through this config (due to char[]), it is isn't too hard to work with. 

 For example, from Ruby:
```
secureConfig.add("keystore.path", LogStash::SETTINGS.get_setting("path.keystore").value.chars)
```

 Imo, this make more sense then having two configurations, or adding an unsecure (String based) part to the secure config.

logstash-core/src/main/java/org/logstash/secret/store/SecretStoreFactory.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * <p>Adds the credential to the {@link SecureConfig} that is needed to access the {@link SecretStore}. Value read from environment variable "LOGSTASH_KEYSTORE_PASS"</p>
+     *
+     * @param secureConfig The configuration to add the secret store access
+     */
+    private static void addSecretStoreAccess(SecureConfig secureConfig) {
+        String environment = System.getenv(ENVIRONMENT_PASS_KEY);
+
+        char[] pass = null;
+        if (environment != null) {
```
jordansissel: This requires the environment variable, but the method still succeeds  when it is not present. We should give some feedback to the user here when a required parameter (this environment variable) is either missing or incorrect.

logstash-core/src/main/java/org/logstash/secret/store/SecretStoreUtil.java
```diff
@@ -96,4 +119,45 @@ private SecretStoreUtil() {
...
+
+    /**
+     * De-obfuscates the obfuscated char[] generated by {@link SecretStoreUtil#obfuscate(char[])}
+     *
+     * @param chars The chars to de-obscure
+     * @return the de-obscured chars
+     */
+    public static char[] deObfuscate(char[] chars) {
+        byte[] bytes = asciiCharToBytes(chars);
+        byte[] random = Arrays.copyOfRange(bytes, bytes.length / 2, bytes.length);
```
jordansissel: The first second half of the `chars` is  some kind of one-time pad? Can you add comments describing the format or pointing to a reference that describes the format?

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -92,13 +98,13 @@ public JavaKeyStore(Path keyStorePath, char[] keyStorePass) {
                 String keyStorePermissions = System.getProperty("logstash.keystore.file.perms", "rw-rw----");
```
jordansissel: I missed this last time, but in a future PR, let's get this out of system properties and put it in `logstash.yml` *or* make it not a setting at all (default to 0600 is what other tools like ssh-keygen, ruby netrc, etc default to)

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreFactoryTest.java
```diff
@@ -0,0 +1,194 @@
...
+                }
+            }
+        }
+    }
+
+    @Test
+    public void testAlternativeImplementation(){
+        String beforeSystem = System.getProperty(SecretStore.class.getCanonicalName());
+        try {
+            System.setProperty(SecretStore.class.getCanonicalName(), "org.logstash.secret.store.SecretStoreFactoryTest$MemoryStore");
```
jordansissel: My instinct for java properties is that we should use `logstash.yml` settings for this. Can you tell me a story about why this is a java system property and not in Logstash settings?

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreUtilTest.java
```diff
@@ -14,8 +14,9 @@
 public class SecretStoreUtilTest {
 
     private String asciiString;
+
     @Before
-    public void setup(){
+    public void _setup() {
```
jordansissel: why prefix with _ ?

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreFactoryTest.java
```diff
@@ -0,0 +1,194 @@
...
+    public ExpectedException thrown = ExpectedException.none();
+
+    //exact copy from https://stackoverflow.com/questions/318239/how-do-i-set-environment-variables-from-java
+    //thanks @pushy and @Edward Campbell !
+    @SuppressWarnings("unchecked")
+    private static void setEnv(Map<String, String> newenv) throws Exception {
+        try {
+            Class<?> processEnvironmentClass = Class.forName("java.lang.ProcessEnvironment");
+            Field theEnvironmentField = processEnvironmentClass.getDeclaredField("theEnvironment");
+            theEnvironmentField.setAccessible(true);
```
jordansissel: Do you need to suppress accessibility warnings also? (Java 9?)

logstash-core/src/main/java/org/logstash/secret/store/SecretStoreFactory.java
jakelandis: I have added a warning via the CLI w/r/t to this missing value. It will be in a subsequent review.

logstash-core/src/main/java/org/logstash/secret/store/SecretStoreUtil.java
jakelandis: I struggled with how to document...since this is the obfuscation code and open source. Kinda trying to the make the code obtuse.

> The first second half of the chars is some kind of one-time pad? 
They first and second half are original bits (b1) and the XOR bits (b2) respectively.  b1 = XOR(b2) && b2 = XOR(b1).

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: I defaulted to E.S.'s default [here](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/common/settings/KeyStoreWrapper.java#L333).  It looks like [this](https://github.com/elastic/elasticsearch/commit/7a035f5f8435962072157f46031e73855fe4a875) is the reason.  I made a note to revisit this while doing package testing. 

> let's get this out of system properties and put it in logstash.yml or make it not a setting at all

It's here mostly for testing, and had not planned on documenting this. I have other ways I can set this for testing for you prefer to expose it at all.  I don't think it should be .yml setting. 

thoughts ?

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreFactoryTest.java
jakelandis: Java 8 isn't complaining, and it would prefer to wait to make Java 9 specific changes until we support Java 9.  FWIW, It should only be a warning in on test compilation.

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreFactoryTest.java
jakelandis: This has been replaced with a value from `logstash.yml` and will be part of a future review. _I sometimes use system properties as placeholders while building out functionality`

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreUtilTest.java
jakelandis: This is strictly for my IDE... i use an auto-sort (notice all the method names are alpha sorted), and like to keep the `setup` method on the top. I try to avoid this type non-sense except for test code.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: I'll go ahead and nuke the System property, but leave the 660.

jakelandis:
@jordansissel - I have responded to all of the comments. However, mind if we continue the minor discussions on the next review ? (assuming you agree there are no large changes pending in this review)

I have one more medium/large code change w/r/t to this area of code.

jakelandis:
@jordansissel - mind if I merge this to the feature branch and move on to the next review ?

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jordansissel: Ok I'll file a blocking issue on the permissions to follow up.

logstash-core/src/main/java/org/logstash/secret/store/SecretStoreFactory.java
jordansissel: I'm going to forget to check for this, so I can't really evaluate whether or not this is complete :|

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jordansissel: https://github.com/elastic/logstash/issues/8735

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreFactoryTest.java
jordansissel: It's hard to verify this claim, and it is likely I will have forgotten to check for it in a future review. It is hard for me to carry context of reviews across multiple PRs (we don't' do this for any other PRs?), so while I trust you, I cannot verify your claim that it will be part of a future review because I have little-or-no means to track the deferred work in this PR in a future PR.

Should I file blocker issues about all of these deferred things?

logstash-core/src/main/java/org/logstash/secret/store/SecretStoreFactory.java
jakelandis: here is the warning in the CLI : https://github.com/jakelandis/logstash/blob/8353-working/logstash-core/src/main/java/org/logstash/secret/cli/SecretStoreCli.java#L126

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: thanks for logging the follow up, we can continue the 600 vs. 660 conversation over there.

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreFactoryTest.java
jakelandis: Here is the (future) code that will read read from the settings:

https://github.com/jakelandis/logstash/blob/8353-working/logstash-core/lib/logstash/util/secretstore.rb#L15
and
https://github.com/jakelandis/logstash/blob/8353-working/logstash-core/src/main/java/org/logstash/secret/store/SecretStoreFactory.java#L70

jordansissel:
LGTM

jakelandis:
@jordansissel - thanks, I will rebase the feature branch, then rebase this PR, then merge.

PR #8663, opened 2017-11-14
CLEANUP: Use paquet from source tree?

original-brownbear:
Random find I made :D We have the sources for `paquet` in this repo, but we're pulling it from Rubygems?

Shouldn't we be using the version from source?


original-brownbear:
seems to cause all kinds of test failures to do this

Issue #8674, opened 2017-11-14
PQ batch size maximization


colinsurprenant:
as discussed in #8633 we decided to change the PQ read strategy toward batch size maximization similar to the MQ strategy.

To achieve that some cleanups et refactoring will be necessary, I will try to proceed by small PRs to facilitate reviews.

colinsurprenant:
Page refactor PR #8690

colinsurprenant:
closing per #8702. Merged into 6.x so will be part of 6.2 release.

PR #8675, opened 2017-11-14
[CLEANUP] refactor page read deserialization out of the page into batch

colinsurprenant:
This is a cleanup to move the page read deserialization out of the Page and into the Batch. Now Page classes do not have knowledge of the Batch which happen at the Queue level. 

Part of #8674 for the PQ Batch size maximization

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
@@ -2,6 +2,7 @@ package org.logstash.ackedqueue;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.logstash.ackedqueue.io.LongVector;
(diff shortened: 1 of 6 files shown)


logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
```diff
@@ -21,6 +22,17 @@ public Batch(List<Queueable> elements, LongVector seqNums, Queue q) {
         this.closed = new AtomicBoolean(false);
     }
 
+    public Batch(SequencedList<byte[]> serialized, Queue q) {
+        final List<byte[]> serializedElements = serialized.getElements();
+        this.elements = new ArrayList<>(serializedElements.size());
+        for (final byte[] element : serializedElements) {
+            this.elements.add(q.deserialize(element));
+        }
+        this.seqNums = serialized.getSeqNums();
```
original-brownbear: @colinsurprenant I wouldn't duplicate primary constructors here. Just extract the deserialisation loop to a method `private static List<Queueable> deserializeElements(Queue q, Collection<byte[]> c)` and make the secondary constructor simply call:
```
this(deserialize(q, serialized.getElements()), serialized.getSeqNums(), q);
```

... or just drop the `public Batch(List<Queueable> elements, LongVector seqNums, Queue q)` (maybe it's unnecessary now) ?

Other than that, looks fine :)

colinsurprenant: +1

colinsurprenant:
Yup - thanks, cleaned the constructor duplication.

original-brownbear:
LGTM (if green + squashed) :)

colinsurprenant:
No worries, I always squash ... unless I forget 😅
Thanks.

colinsurprenant:
rebased and merged into 6.x and master.

Issue #8672, opened 2017-11-14
Test output in CI overly verbose


robbavey:
The amount of logging output in the CI builds has grown hugely, and the logs are now filled with entries like:
```
:runIntegrationTests<===========--> 88% EXECUTING [16m 57s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 57s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 57s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 57s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 57s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 57s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 57s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 57s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 57s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 58s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 58s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 58s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 58s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 58s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 58s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 58s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 58s]>
:runIntegrationTests<===========--> 88% EXECUTING [16m 58s]>
```

https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-integration-2/159/console

original-brownbear:
fixed in #8678

PR #8678, opened 2017-11-15
 #8672 adjust gradle to use plain output on CI

original-brownbear:
It's in the title, make the output on CI non-interactive to to make it readable in Jenkins logs.
No more weird Gradle status lines mixed into everything :)


original-brownbear:
@jsvd thanks :)

PR #8599, opened 2017-11-07
 #7135 add Gradle task for bootstrapping and running all integration tests

original-brownbear:
This gives us a clean all-in-one Gradle build including the ITs, complete with Gradle caching and bootstrapping of IT dependencies to resolve #7135.

* All in one build runs fine without any JRuby or Ruby installed on  (ran on a slow VM, my bad :)):
`./gradlew clean check`:

```sh

BUILD SUCCESSFUL in 41m 50s
36 actionable tasks: 24 executed, 12 up-to-date
➜  logstash git:(gradle-integration-tests)                 
```

* Added an optional parameter to run specific IT specs to the build to not change Jenkins behavior.
* Had to improve stability of the LS version IT by only reading the last line from stdout (in the Gradle env there are JAVA_OPTS set no matter what <= you get a line warning you about that in a version check call => only read version from the last line of stdout
* Adjusted readme.

----------------------

I think this is pretty neat for locally running/bootstrapping ITs + a lot more portable than the `rake` based approach since it's totally isolated from whatever local Ruby environment you have + it allows outside contributors to run the full suite in one command locally/easily without waiting for us to trigger Jenkins :)

PR #8680, opened 2017-11-15
BUILD: Fix missing Gradle Cache Invalidations in Integration Test Build

original-brownbear:
The setup from #8599 was missing a few cache invalidations:

* The ones for building the `.tar` should be obvious, all the added files are bundled in the `tar`.
* Adding the `Gemfile` and `.gemspec` for the assembled core in the ITs is very important too, I was just bit by this when switching between branches that had different dependencies for core. If we don't invalidate here, then a change in runtime dependencies in core will not trigger a reinstall of `qa/integration` to pull in the missing dependencies


jakelandis:
LGTM

original-brownbear:
@jakelandis thanks!

PR #8685, opened 2017-11-15
[MINOR] bad bad javadoc

colinsurprenant:
`<` broke the javadoc generation and thus broke the build 😿


colinsurprenant:
@robbavey yeah I saw that but I'm trying to keep changes to a minimum for a quick build recovery.

colinsurprenant:
merged in master and 6.x. Thanks @robbavey.

PR #8690, opened 2017-11-17
Remove Page/HeadPage/TailPage) inheritance

colinsurprenant:
This is work needed for the batch maximization #8674. 

To be able to continue reading from a page upon beheading (moving a head page into a tail page) which happen in a write operation, a *stable* Page object is required which can be used for read operations regardless if it is a head page or transformed into a tail page. 

This PR changes the idea of distinct HeadPage and TailPage objects to a single Page object which is simply `writable` or not which really only affects the ability to perform a `write` operation at all and change the `checkpoint` strategy.

All tests are passing, this refactor surfaced a concurrency issue and extra locking has been added to public Queue methods. Otherwise I minimized any other non-related code change.

Changes explanations:
- Basically `HeadPage` and `TailPage` code has been moved into `Page`,  the only real difference was with the `write()` method and `checkpoint()` which are now guarded with a new `writable` attribute.
- The `HeadPage` and `TailPage` constructors have been moved into a new `PageFactory`.
- Added locking around remaining `Queue` public methods
- Move the Page capacity check inside the `write()` method locked section


logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
```diff
@@ -2,27 +2,28 @@
...
-import java.util.ArrayList;
 import java.util.BitSet;
-import java.util.List;
+
+import org.logstash.ackedqueue.io.CheckpointIO;
 import org.logstash.ackedqueue.io.LongVector;
 import org.logstash.ackedqueue.io.PageIO;
 
-public abstract class Page implements Closeable {
+public class Page implements Closeable {
```
original-brownbear: @colinsurprenant I guess we can even make this one `final` now right?

logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
```diff
@@ -32,30 +33,52 @@ public Page(int pageNum, Queue queue, long minSeqNum, int elementCount, long fir
...
      */
     public SequencedList<byte[]> read(int limit) throws IOException {
-
         // first make sure this page is activated, activating previously activated is harmless
         this.pageIO.activate();
 
         SequencedList<byte[]> serialized = this.pageIO.read(this.firstUnreadSeqNum, limit);
         assert serialized.getSeqNums().get(0) == this.firstUnreadSeqNum :
-            String.format("firstUnreadSeqNum=%d != first result seqNum=%d", this.firstUnreadSeqNum, serialized.getSeqNums().get(0));
+                String.format("firstUnreadSeqNum=%d != first result seqNum=%d", this.firstUnreadSeqNum, serialized.getSeqNums().get(0));
```
original-brownbear: @colinsurprenant nit: this indent looks broken now?

logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
```diff
@@ -32,30 +33,52 @@ public Page(int pageNum, Queue queue, long minSeqNum, int elementCount, long fir
...
 
         this.firstUnreadSeqNum += serialized.getElements().size();
 
         return serialized;
     }
 
+    public void write(byte[] bytes, long seqNum, int checkpointMaxWrites) throws IOException {
+        if (! this.writable) {
+            throw new IOException(String.format("page=%d is not writable", this.pageNum));
```
original-brownbear: @colinsurprenant This isn't an `IOException` is it? We're throwing this as a result of a logic error (trying to write to a tail page) as far as I can tell. If so we must throw a `RuntimeException` like `IllegalStateException`.

logstash-core/src/main/java/org/logstash/ackedqueue/PageFactory.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * create a new head page from an existing {@link Checkpoint} and open page.{@literal {pageNum}} empty valid data file
+     *
+     * @param checkpoint existing head page {@link Checkpoint}
+     * @param queue the {@link Queue} instance
+     * @param pageIO the {@link PageIO} delegate
+     * @return {@link Page} the new head page
+     */
+    public static Page newHeadPage(Checkpoint checkpoint, Queue queue, PageIO pageIO) {
```
original-brownbear: @colinsurprenant the whole body of this method should probably be wrapped in a `try { ... } catch (Exception ex) {p.close; throw ex;}` Otherwise any exception thrown in the body will leave a `Page` behind that goes out of scope without a `close`.

logstash-core/src/main/java/org/logstash/ackedqueue/PageFactory.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * create a new tail page for an exiting Checkpoint and data file
+     *
+     * @param checkpoint existing tail page {@link Checkpoint}
+     * @param queue the {@link Queue} instance
+     * @param pageIO the {@link PageIO} delegate
+     * @return {@link Page} the new tail page
+     */
+    public static Page newTailPage(Checkpoint checkpoint, Queue queue, PageIO pageIO) {
```
original-brownbear: @colinsurprenant the whole body of this method should probably be wrapped in a `try { ... } catch (Exception ex) {p.close; throw ex;}` Otherwise any exception thrown in the body will leave a `Page` behind that goes out of scope without a `close`.

original-brownbear: @colinsurprenant also, wouldn't this be better called `openTailPage` since we're not actually creating a file? :)

logstash-core/src/main/java/org/logstash/ackedqueue/PageFactory.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * create a new head page from an existing {@link Checkpoint} and open page.{@literal {pageNum}} empty valid data file
+     *
+     * @param checkpoint existing head page {@link Checkpoint}
+     * @param queue the {@link Queue} instance
+     * @param pageIO the {@link PageIO} delegate
+     * @return {@link Page} the new head page
+     */
+    public static Page newHeadPage(Checkpoint checkpoint, Queue queue, PageIO pageIO) {
+        Page p = new Page(
```
original-brownbear: @colinsurprenant can't we simply overload the constructor of `Page` and save the duplication between this method and `newTailPage`?

original-brownbear:
@colinsurprenant looks good for the most part, **much** cleaner and easier to follow with just `Page` :)
Just a few small comments around `Exception`s.

logstash-core/src/main/java/org/logstash/ackedqueue/PageFactory.java
```diff
@@ -0,0 +1,80 @@
+package org.logstash.ackedqueue;
+
+import org.logstash.ackedqueue.io.PageIO;
+
+import java.util.BitSet;
+
+public class PageFactory {
```
original-brownbear: @colinsurprenant slightly unrelated topic to this PR, but since we're going to expose our Java codebase eventually (hopefully soon :D), do we need to have this new class `public`? I think package private suffices here since only `Queue` uses it right?

colinsurprenant:
@original-brownbear thanks for the review! good comments, will address shortly!

logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
colinsurprenant: +1

logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
colinsurprenant: +1

logstash-core/src/main/java/org/logstash/ackedqueue/PageFactory.java
colinsurprenant: good idea.

logstash-core/src/main/java/org/logstash/ackedqueue/PageFactory.java
colinsurprenant: ah ... good question, new in this context is about the object instantiation, like a constructor, not about the page file creation.

logstash-core/src/main/java/org/logstash/ackedqueue/PageFactory.java
colinsurprenant: we could be I'd prefer to keep to so that it is obvious in the code that we are creating a new head page instead of creating a page with a `true` value for the `writable` parameter.

colinsurprenant:
@original-brownbear  pushed review fixes LMKWYT.

original-brownbear:
@colinsurprenant LGTM (don't forget to squash :D)

colinsurprenant:
Thanks @original-brownbear ! 

merged into master and 6.x

PR #8695, opened 2017-11-20
 #7403 Fix Non-Deterministic Graph Hashing causing Duplicate IDs

### LINKED ISSUE #7403: logstash 6.0.0-alpha2 throws lir exception if two identical plugins are used in the inputs or outputs
example:

```
/tmp/logstash-6.0.0-alpha2 % bin/logstash -e "input { generator { count => 1 } generator { count => 1 } }"
Sending Logstash's logs to /tmp/logstash-6.0.0-alpha2/logs which is now configured via log4j2.properties
[2017-06-09T16:41:17,014][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because 'config.string' (-e) is being used.
[2017-06-09T16:41:17,078][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2017-06-09T16:41:17,251][ERROR][logstash.agent           ] Failed to execute action {:action=>LogStash::PipelineAction::Create/pipeline_id:main, :exception=>"Java::OrgLogstashConfigIr::InvalidIRException", :message=>"Some nodes on the graph are fully redundant!\n**GRAPH**\nVertices: 2 Edges: 0\n----------------------\n== Vertices Without Edges ==\nP[input-generator{\"count\"=>1}|[protocol]pipeline:2:9:```\ngenerator { count => 1 }\n```]\nP[input-generator{\"count\"=>1}|[protocol]pipeline:2:34:```\ngenerator { count => 1 }\n```]\n**GRAPH**|P[input-generator{\"count\"=>1}|[protocol]pipeline:2:9:```\ngenerator { count => 1 }\n```]===P[input-generator{\"count\"=>1}|[protocol]pipeline:2:34:```\ngenerator { count => 1 }\n```]"}
[2017-06-09T16:41:17,253][ERROR][logstash.agent           ] An exception happened when converging configuration {:exception=>LogStash::Error, :message=>"Don't know how to handle `Java::OrgLogstashConfigIr::InvalidIRException` for `LogStash::PipelineAction::Create/pipeline_id:main`", :backtrace=>["/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/converge_result.rb:27:in `create'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/converge_result.rb:67:in `add'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/agent.rb:351:in `converge_state'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/agent.rb:139:in `with_pipelines'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/agent.rb:337:in `converge_state'", "org/jruby/RubyArray.java:1613:in `each'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/agent.rb:324:in `converge_state'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/agent.rb:164:in `converge_state_and_update'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/agent.rb:139:in `with_pipelines'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/agent.rb:162:in `converge_state_and_update'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/agent.rb:88:in `execute'", "/tmp/logstash-6.0.0-alpha2/logstash-core/lib/logstash/runner.rb:302:in `execute'", "org/jruby/RubyProc.java:281:in `call'", "/tmp/logstash-6.0.0-alpha2/vendor/bundle/jruby/1.9/gems/stud-0.0.22/lib/stud/task.rb:24:in `initialize'"]}

```

original-brownbear:
Fixes #7403 (and analogous issues) by solving a two general issues in the current Graph hashing implementation:

*  Using `HashSet` and `HashMap` together with sorted (sorting the hash sources) iteration does not work here in this snippet:

```java
        // The lineage can be quite long and we want to avoid the quadratic complexity of string concatenation
        // Thus, in this case there's no real way to get the hash source, we just hash as we go.
        lineage().
                map(Vertex::contextualHashSource).
                sorted().
                forEachOrdered(v -> {
                    byte[] bytes = v.getBytes(StandardCharsets.UTF_8);
                    lineageDigest.update(bytes);
                });
```

This will hash the same value for two lineages that contain all the same elements (in terms of individual hash source) but in different order (if you think about this, this a fairly unlikely scenario :), but the test config I copied in here from the issue produces it).

What you need to do in order to fix this is to ensure that you have deterministic ordering of the stream produced by `org.logstash.config.ir.graph.Vertex#lineage`. The ordering in `master` is evidently not deterministic as demonstrated by one of the tests I added, that fails randomly with `master`:

screen shot 2017-11-19 at 19 07 42

=> fixed by using `LinkedHashMap`s (and their `Set`s) throughout the whole code path that leads to the Graph's ordering to get a deterministic iteration order.
* Fixed the `hashCode` for `Edge` by forcing an immutable and guaranteed to be unique `hashCode` on each `Vertex` via a sequence number and using it to generate a stable `hashCode` for `Edge`. The change to `Vertex` is not strictly necessary, but a nice safety measure to ensure a valid `hashCode` on all its children
* Also this PR contains some cleanups (that at least I required to be able to properly reason about the issue) that I think are valid to keep in here. I can separate them out though
   * `IfVertex` wasn't using the `SourceAndMetadata` passed down to it
   *  There was a lot of duplication on how the `Stream<Vertex>` over all vertices was generated in `Graph` that I cleaned up
   * Also, there were a lot of redundant getter usages from within for example `Graph` (I fixed those since they do a good job at hiding where we use an actual immutable collection and where we create one dynamically, making it a lot harder to reason about whether or not some `Stream<?>` is actually sorted or not)
  `org.logstash.config.ir.graph.Graph#vertices` and `org.logstash.config.ir.graph.Graph#edges`).


original-brownbear:
@andrewvc phew fixed, this one was tricky :D only description needs to be completed, but will be ready in an hour or two.

original-brownbear:
@andrewvc ready for review :)

original-brownbear:
@andrewvc adjusted the wording of the PR description.
We don't strictly need the `hashCode` fix on `Vertex` but we in fact need a valid `hashCode` and `equals` on `Edge`. Since `Vertex` is subclassed a bunch I think it's still very much worth while to make that method `final` and correct.

logstash-core/src/main/java/org/logstash/config/ir/graph/Edge.java
```diff
@@ -22,15 +39,29 @@ public void setGraph(Graph graph) {
         }
     }
 
+    @Override
+    public final int hashCode() {
+        return 37 * from.hashCode() + to.hashCode();
```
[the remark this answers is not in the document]
original-brownbear: @andrewvc I tend to not use that method (doesn't matter all that much here), it's really slow compared to just doing you own thing because of the varargs triggering array creation on every call + it's meh to read, now I can't just click on `hashCode()` for `from` and `to` in the IDE and have to go another step if I wanted to read through this and make sure everything is on the up and up again :D

original-brownbear:
@andrewvc well, apparently the metadata is not being used properly in the other hashing path (that would be the next problem to fix imo), as you can see I could even remove it from `Vertex`s constructor.
As discussed I fully agree with you on the move to hashing by source + metadata, but I'd do it in the next step just for risk reasons I guess. We now learnt we had quite a few holes in our testing (I hope adding this monster of a config to the tests closes those all in one go :D) and I really dislike merging a big functional change together with a big test change?
I could get right on it moving everything to hashing from source+metadata in the next PR after this one, shouldn't take long (*knock on wood) :)

original-brownbear:
@andrewvc come to think of it ... we want to fix all the iterations to be deterministically order regardless imo.
You know, the way I tracked this down in part, was that I saw that the output of `Graph#toString` differed in ordering between test runs (obviously, the `#toString` logic used iterations over `HashSet` in some parts). This introduces a needless risk of bugs and test instabilities in users of the `PipelineIR` imo even though it's not an outright bug.

andrewvc:
OK, I'm fine doing this in two phases. I'll review later today

logstash-core/src/test/resources/org/logstash/config/ir/complex.cfg
```diff
@@ -0,0 +1,1999 @@
+# Generated by Chef for REDACTED
```
andrewvc: To merge this we'll need https://github.com/elastic/logstash/issues/7403#issuecomment-346319046 addressed since this isn't under our CLA yet.

logstash-core/src/test/java/org/logstash/config/ir/ConfigCompilerTest.java
```diff
@@ -14,4 +16,30 @@ public void testConfigToPipelineIR() throws Exception {
         assertThat(pipelineIR.getOutputPluginVertices().size(), is(1));
         assertThat(pipelineIR.getFilterPluginVertices().size(), is(0));
     }
+
+    @Test
+    public void testConfigDuplicateBlocksToPipelineIR() throws Exception {
```
andrewvc: Can you document why this fix makes things deterministic?

logstash-core/src/main/java/org/logstash/config/ir/graph/Vertex.java
```diff
@@ -49,11 +50,20 @@ public InvalidEdgeTypeException(String s) {
...
     }
 
-    public Graph getGraph() {
+    @Override
+    public final int hashCode() {
+        return hashCode;
+    }
+
+    public final boolean equals(final Object other) {
+        return this == other;
```
andrewvc: So, this is the default implementation of `Object.equals`. Is this just here to make that explicit?

original-brownbear: @andrewvc I wanted the `final` on the parent class here, to make sure that this is the case for all children as well (obviously it is, but it's annoying that I had to go through all of them when trying to track this down) => much easier to read this way.

logstash-core/src/test/resources/org/logstash/config/ir/complex.cfg
original-brownbear: @andrewvc I can just delete this now and bring it back when and if we get that CLA. If we don't I can spend some time to reproduce this with a minimal example (it's like a 1h effort probably).

logstash-core/src/test/java/org/logstash/config/ir/ConfigCompilerTest.java
original-brownbear: @andrewvc sure, let's just make this one check that each iteration produces the same result?

original-brownbear:
@andrewvc made the tests run the hashing repeatedly and check that the hash stays constant :)
There are some tests (`org.logstash.config.ir.imperative.DSLTest`) that seem to implicitly test this, but not using any examples that would trigger the conditions that break things here.

original-brownbear:
@andrewvc that already happened in https://github.com/elastic/logstash/pull/8711, I just merged them all into one file again here :)

original-brownbear:
@andrewvc Will have to create a back port fo `6.0` since the test infra isn't there in that branch.
The production code should go in clean though, since back porting the tests will be pretty involved and the production code is the same between `6.0` and `6.x`, wdyt ... skip the tests in `6.0`?

PR #8696, opened 2017-11-20
Add style checkers

original-brownbear:
WIP, it's very hard to get in compliance with Elasticsearch's rules in our codebase.

### DIFF
--- a/buildSrc/src/main/resources/checkstyle.xml
+++ b/buildSrc/src/main/resources/checkstyle.xml
@@ -1,0 +1,88 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+    "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+    "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+
+<module name="Checker">
+    <property name="charset" value="UTF-8" />
+
+    <module name="SuppressionFilter">
+        <property name="file" value="${suppressions}" />
+    </module>
+
+    <!-- Checks Java files and forbids empty Javadoc comments -->
+    <module name="RegexpMultiline">
+        <property name="format" value="\/\*[\s\*]*\*\/"/>
+        <property name="fileExtensions" value="java"/>
+        <property name="message" value="Empty javadoc comments are forbidden"/>
+    </module>
+
+    <module name="TreeWalker">
+        <!-- Its our official line length! See checkstyle_suppressions.xml for the files that don't pass this. For now we
+          suppress the check there but enforce it everywhere else. This prevents the list from getting longer even if it is
+          unfair. -->
+        <module name="LineLength">
+            <property name="max" value="100"/>
+        </module>
+
+        <module name="AvoidStarImport" />
+        <!-- Doesn't pass but we could make it pass pretty quick.
+        <module name="UnusedImports">
+          The next property is optional. If we remove it then imports that are
+          only referenced by Javadoc cause the check to fail.
+          <property name="processJavadoc" value="true" />
+        </module>
+        -->
+
+        <!-- Non-inner classes must be in files that match their names. -->
+        <module name="OuterTypeFilename" />
+        <!-- No line wraps inside of import and package statements. -->
+        <module name="NoLineWrap" />
+        <!-- Each java file has only one outer class -->
+        <module name="OneTopLevelClass" />
+        <!-- The suffix L is preferred, because the letter l (ell) is often
+        hard to distinguish from the digit 1 (one). -->
+        <module name="UpperEll"/>
+
+        <module name="EqualsHashCode" />
+
+        <!-- Checks that the order of modifiers conforms to the suggestions in the
+        Java Language specification, sections 8.1.1, 8.3.1 and 8.4.3. It is not that
+        the standard is perfect, but having a consistent order makes the code more
+        readable and no other order is compellingly better than the standard.
+        The correct order is:
+                public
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/resources/checkstyle.xml
```diff
@@ -0,0 +1,84 @@
...
+        -->
+
+        <!-- Non-inner classes must be in files that match their names. -->
+        <module name="OuterTypeFilename" />
+        <!-- No line wraps inside of import and package statements. -->
+        <module name="NoLineWrap" />
+        <!-- Each java file has only one outer class -->
+        <module name="OneTopLevelClass" />
+        <!-- The suffix L is preferred, because the letter l (ell) is often
+        hard to distinguish from the digit 1 (one). -->
```
jordansissel: The comment here plus the name of this check "UpperEll" was not enough to inform me about the behavior. Googling around in search of understanding, this seems to be about requiring an uppercase "L" specifically on long literals? Like "1000L" vs "1000l"

Can you add in the comment that this is for `long` literals?

buildSrc/src/main/resources/checkstyle.xml
```diff
@@ -0,0 +1,84 @@
...
+            <property name="format" value="serialVersionUID" />
+            <property name="message" value="Do not declare serialVersionUID." />
+            <property name="ignoreComments" value="true" />
+        </module>
+        <module name="RegexpSinglelineJava">
+            <property name="format" value="java\.io\.Serializable" />
+            <property name="message" value="References java.io.Serializable." />
+            <property name="ignoreComments" value="true" />
+        </module>
+        <!-- end Orwellian suppression of Serializable -->
```
jordansissel: > Orwellian suppression of

This description is confusing.  Maybe omit these three words?

buildSrc/src/main/resources/checkstyle.xml
```diff
@@ -0,0 +1,84 @@
...
+
+    <!-- Checks Java files and forbids empty Javadoc comments -->
+    <module name="RegexpMultiline">
+        <property name="format" value="\/\*[\s\*]*\*\/"/>
+        <property name="fileExtensions" value="java"/>
+        <property name="message" value="Empty javadoc comments are forbidden"/>
+    </module>
+
+    <module name="TreeWalker">
+        <!-- Its our official line length! See checkstyle_suppressions.xml for the files that don't pass this. For now we
```
jordansissel: > checkstyle_suppressions.xml 

Should there be checkstyle_suppressions.xml added in this PR? I do not see one.

jordansissel:
If an IntellIJ config for this style accompanies this PR, it would be cool. @original-brownbear thoughts on this? It's OK if this is out of scope.

original-brownbear:
@jordansissel wops I totally forgot to finish this one :) Actually yea we need suppressions and I was planing on reverting the code changes for now so we can go one by one. Anything else is just too crazy :) The rules are taken 1:1 from ES and imo it's worth it to not adjust them, simply to stay in line with what they do right?

original-brownbear:
@jordansissel I'm not 100% sure on this, but if I remember correctly Idea is able to understand code style configs without any additional configuration (not sure how well that works, I know for scalastyle it works very well though :)).

jordansissel:
>  I was planing on reverting the code changes for now

++ I am personally using the code changes mostly as a "This is what the code looks like after style checks pass". I am OK if this PR only includes the checkstyle.xml and omits the code changes which we could do at a later time.

> Idea is able to understand code style configs without any additional configuration 

Oh! Delightful. We can make this work -- no change required for this PR on Intellij/IDEA

> The rules are taken 1:1 from ES and imo it's worth it to not adjust them

My comments are generally about the way we describe the rules (the comments in the xml file), so I'm OK if we change them, and we can follow up with a PR on elasticsearch to make the same clarity changes there. Thoughts?

jordansissel:
That said, I'm OK if we copy Elasticsearch's for now and improve the internal commenting later.

original-brownbear:
@jordansissel alrighty reverted all code changes and excluded everything (will have to make the exclude less coarse in the next step(s)) now.
I think this is a valid step 0 here :)

original-brownbear:
Jenkins test this please

buildSrc/src/main/resources/checkstyle.xml
```diff
@@ -0,0 +1,88 @@
...
+          suppress the check there but enforce it everywhere else. This prevents the list from getting longer even if it is
+          unfair. -->
+        <module name="LineLength">
+            <property name="max" value="100"/>
```
danhermann: Given the choice, I'd opt for longer lines. I think it's a pretty rare screen these days that won't accommodate line lengths of 120 or more.

original-brownbear: @danhermann yea but diffing (and eventually reading too) becomes a pain with more and more information per line :) => I'm a big fan of `100` here tbh :D

danhermann: Sure, that's fine. I just threw it out there in case no one had other preferences.

original-brownbear:
Jenkins test this please

jakelandis:
can we wait to merge this till after right after 6.2 (or 6.3) branch is cut?  it has the potential to add alot of noise to a pending feature freeze. 

Also ++ to longer lines.

kares:
this seem to have never materialized, I think this effort (if needed) should be restarted with a clean slate

PR #8700, opened 2017-11-20
Minor: Move reading the setting into to a utility class

jakelandis:
This is support of a adding additional command line entry points that also need to read the settings (logstash.yml) file. Since there is a distinct before/read/after life-cycle this utility provides some direction for additional command line entry points without duplicating code and logic.

Part of #8353

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -10,6 +10,7 @@ require "logstash/namespace"
 require "logstash-core/logstash-core"
 require "logstash/environment"
 require "logstash/modules/cli_parser"
+require "logstash/util/settings_helper"
 
 LogStash::Environment.load_locale!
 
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/util/settings_helper.rb
```diff
@@ -0,0 +1,79 @@
+require "logstash/settings"
+require "logstash-core/logstash-core"
+require "logstash/environment"
+
```
jakelandis: Note - other then some comments is is direct copy/paste from runner.rb

logstash-core/lib/logstash/runner.rb
```diff
@@ -30,12 +31,8 @@
...
 class LogStash::Runner < Clamp::StrictCommand
   include LogStash::Util::Loggable
-  # The `path.settings` and `path.logs` need to be defined in the runner instead of the `logstash-core/lib/logstash/environment.rb`
-  # because the `Environment::LOGSTASH_HOME` doesn't exist in the context of the `logstash-core` gem.
-  #
-  # See issue https://github.com/elastic/logstash/issues/5361
-  LogStash::SETTINGS.register(LogStash::Setting::String.new("path.settings", ::File.join(LogStash::Environment::LOGSTASH_HOME, "config")))
-  LogStash::SETTINGS.register(LogStash::Setting::String.new("path.logs", ::File.join(LogStash::Environment::LOGSTASH_HOME, "logs")))
+
+  LogStash::Util::SettingsHelper.pre
```
original-brownbear: @jakelandis since we are now using this thing 3 times inside of `LogStash::Runner`, why not actually `include LogStash::Util::SettingsHelper`? :)

Also, call this thing (that I understand will become a hook for more stuff to run at down the line?) a little more descriptive like `before_settings`? :)

original-brownbear:
@jakelandis just cosmetics :)

logstash-core/lib/logstash/runner.rb
jakelandis: > why not actually include LogStash::Util::SettingsHelper?

I tend to prefer not including helper methods. Since including it becomes part of the class and imo violates principal of least privilege for this case.  FWIW, require is also how the `LogStash::SETTINGS` is consumed, and this helper is just wrapping access to that. 

>  a little more descriptive like before_settings

I know, not exactly what you asked....but i changed `pre` and `post` to `'pre_process` and `post_process` to better mirror the wording of the underlying (post method).   Keeping the full name (since not including) adds the context (e.g. no need to repeat the word setting).

original-brownbear: @jakelandis fine with me too, LGTM then if green :)

jakelandis:
@original-brownbear - thanks

PR #8702, opened 2017-11-21
maximize PQ read batch size

colinsurprenant:
This PR is based and depends on #8690 

💥  NOTE that this PR also includes Javadoc cleanups, particularly in the `Queue` class.

Work in this PR is about doing batch size maximization on read per #8674. The main logic is located in the `Queue` `_readPageBatch()` method. 

A simple test for this has been added in `QueueTest` `maximizeBatch()`. 

TODOs:
- [x] perform comparative benchmarks pre/post read batch maximization
- [x] perform ES output tests to validate bulk size maximization as per #7243

#### Validate Batch Size Maximization 

I write this simple output plugin to show output batch sizes on *stdout*:

```
# encoding: utf-8
require "logstash/outputs/base"
require "logstash/namespace"
class LogStash::Outputs::BatchStatsOut < LogStash::Outputs::Base
  config_name "batch_stats_out"

  def register; end # must be overriden

  def multi_receive_encoded(encoded)
    $stdout.puts(encoded.size)
  end
end
```

- I noticed that batches were **not** always maximized using the default `pipeline.batch.delay: 5`. Increasing it to **50** always produces batches of 125 elements (default batch size). So for the benchmark tests below `pipeline.batch.delay: 50` will be used.

##### Sample result on master
```
1
0
0
1
1
1
0
0
2
1
0
1
1
0
1
```

##### Sample result with PR
```
125
125
125
125
125
125
125
125
125
125
```
- It is consistent at 125.

### Comparative Benchmark

- On my MBP 2.9 GHz Core i7, 16 GB
- I am using `measure_performance.rb` from [logstash-benchmark-tools](https://github.com/elastic/logstash-benchmark-tools/tree/master/pq_blog) repo

- Config 
```
queue.type: persisted
pipeline.batch.delay: 50
```

- Empty pipeline config
```
$  bin/logstash -e 'input{generator{}} output{stdout{codec => dots}}
```

##### Master performance
The TPS goes from **~13k events/s down to ~2500 events/s** from beginning to end of each page and repeats for each page.

##### PR performance
The TPS goes from **~50k events/s down to ~25k events/s** from beginning to end of each page and repeats for each page.

---

This performance decrease as the page grows seems to indicate that there is is performance problem related to the number of written element in the page. Here is the same tests using  **25mb** pages instead of the default **250mb**.

```
queue.page_capacity: 25mb
```

##### Master performance
The TPS goes from **~40k events/s down to ~30k events/s** from beginning to end of each page and repeats for each page.

##### PR performance
The TPS goes from **~60k events/s down to ~50k events/s** from beginning to end of each page and repeats for each page.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
@@ -15,15 +15,15 @@ public class Batch implements Closeable {
     private final Queue queue;
     private final AtomicBoolean closed;
 
-    public Batch(List<Queueable> elements, LongVector seqNums, Queue q) {
-        this.elements = elements;
+    public Batch(SequencedList<byte[]> serialized, Queue q) {
+        this(serialized.getElements(), serialized.getSeqNums(), q);
+    }
+
+    public Batch(List<byte[]> elements, LongVector seqNums, Queue q) {
+        this.elements = deserializeElements(elements, q);
         this.seqNums = seqNums;
         this.queue = q;
         this.closed = new AtomicBoolean(false);
-    }
-
-    public Batch(SequencedList<byte[]> serialized, Queue q) {
-        this(deserializeElements(serialized.getElements(), q), serialized.getSeqNums(), q);
     }
 
     // close acks the batch ackable events
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/io/LongVector.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/io/LongVector.java
@@ -25,6 +25,22 @@ public final class LongVector {
     }
 
     /**
+     * Store the {@code long[]} to the underlying {@code long[]}, resizing it if necessary.
+     * @param nums {@code long[]} to store
+     */
+    public void add(final LongVector nums) {
+        if (data.length < count + nums.size()) {
+            final long[] old = data;
+            data = new long[(data.length << 1) + nums.size()];
+            System.arraycopy(old, 0, data, 0, old.length);
+        }
+        for (int i = 0; i < nums.size(); i++) {
+            data[count + i] = nums.get(i);
+        }
+        count += nums.size();
+    }
+
+    /**
      * Get value stored at given index.
      * @param index Array index (only values smaller than {@link LongVector#count} are valid)
      * @return Int
--- a/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
+++ b/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
@@ -535,7 +535,7 @@ public class QueueTest {
             }
             assertThat(q.isFull(), is(true));
             
-            Batch b = q.readBatch(10); // read 1 page (10 events)
+            Batch b = q.readBatch(10, TimeUnit.SECONDS.toMillis(1)); // read 1 page (10 events)
             b.close();  // purge 1 page
             
             while (q.isFull()) { Thread.sleep(10); }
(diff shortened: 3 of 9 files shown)


colinsurprenant:
suggesting new default config values in #8707

colinsurprenant:
rebased.

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -953,16 +955,45 @@ public void pageCapacityChangeOnExistingQueue() throws IOException {
...
+
+    @Test(timeout = 5000)
+    public void maximizeBatch() throws IOException, InterruptedException, ExecutionException {
+
+        // very small pages to maximize page creation
+        Settings settings = TestSettings.persistedQueueSettings(1000, dataPath);
+        try (Queue q = new Queue(settings)) {
+            q.open();
+
+            Callable<Integer> writer = () -> {
```
jakelandis: nit: if you don't read the results from the execution, a `Runnable` makes that a bit more obvious

```
Runnable writer = () -> {
...
Future<?> future (or don't even bother with it since there is no need to block this test on the get)
```

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -953,16 +955,45 @@ public void pageCapacityChangeOnExistingQueue() throws IOException {
...
+                Thread.sleep(500); // sleep 500 ms
+                q.write(new StringElement("E2"));
+                return 1;
+             };
+
+            // write one element now and schedule the 2nd write in 500ms
+            q.write(new StringElement("E1"));
+            Future<Integer> future = executor.submit(writer);
+
+            // issue a batch read with a 2s timeout, normally the batch should contain both element since
```
jakelandis: nit: comment and code don't agree (1s vs 2s)

logstash-core/src/main/java/org/logstash/ackedqueue/Batch.java
```diff
@@ -15,17 +15,17 @@
     private final Queue queue;
     private final AtomicBoolean closed;
 
-    public Batch(List<Queueable> elements, LongVector seqNums, Queue q) {
-        this.elements = elements;
+    public Batch(SequencedList<byte[]> serialized, Queue q) {
```
jakelandis: Out of scope for this review, but the differences between `SequencedList<byte[]>` , `List<Queueable> elements`, and `List<byte[]>` is kinda confusing (same data , different views right ?) .   If you are looking to more clean up, I think this can be simplified, but don't have any concrete suggestions at the moment.

colinsurprenant: 😩  totally agree. Basically `List<Queueable>` are de/non-serialized elements and `List<byte[]>` are serialized elements. `SequencedList<byte[]>` is really annoying. Will look into this in another cleanup PR.

logstash-core/src/main/java/org/logstash/ackedqueue/io/LongVector.java
```diff
@@ -24,6 +24,22 @@ public void add(final long num) {
         data[count++] = num;
     }
 
+    /**
```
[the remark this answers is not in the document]
colinsurprenant: yes it is. it is only used into the IO classes which are not thread safe themselves or within a mutex.

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
colinsurprenant: good point. Actually, I think I prefer changing it to `Callable<Void>` just because `Callable` accepts checked exceptions which makes using `Runnable` a pain for this.

colinsurprenant:
thanks a bunch @jakelandis for the review - addressed you comments! will rebase.

colinsurprenant:
submitted #8797 to use new defaults: `pipeline.batch.delay: 50ms` and `queue.page_capacity: 64mb`.

colinsurprenant:
squashed and merged into master and 6.x - thanks @jakelandis

PR #8705, opened 2017-11-21
Bump versions for 6.x to 6.2.0

andrewvc:
Now that we have a 6.1 branch, this should be bumped. 6.1 is already set correctly.


yaauie:
> Now that we have a 6.1 branch [...]

I don't see it: https://github.com/elastic/logstash/tree/6.1 -> 404

PR #8708, opened 2017-11-22
Add notes about doc type bug in ES output to upgrading guide

andrewvc:
This is the workaround for https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/708


docs/static/upgrading.asciidoc
```diff
@@ -66,12 +66,19 @@ Before upgrading Logstash, remember to read the <<breaking-changes,breaking chan
 If you are installing Logstash with other components in the Elastic Stack, also see the
 {stack-ref}/index.html[Elastic Stack installation and upgrade documentation].
 
-If you are using the default mapping templates in Logstash, to continue using these after migrating Elasticsearch to 6.0, you must override the existing template with the 6.x template.
- This can be done by starting a pipeline with the `overwrite_template => true` option in the Elasticsearch output definition in the Logstash config.
+==== Logstash 6.0.0 `document_type` Handling Bug
```
jsvd: I'd make it clearer that this is a default behaviour that is no longer useful when using ES >= 6.x, something like: 

```
==== Issue with Logstash 6.0.0 `document_type` when writing to Elasticsearch 6.x
```

docs/static/upgrading.asciidoc
```diff
@@ -66,12 +66,19 @@ Before upgrading Logstash, remember to read the <<breaking-changes,breaking chan
...
 
-Note that multiple doctypes are no longer supported in Elasticsearch 6.0. Please refer to
- {ref}/removal-of-types.html[Removal of mapping types] and {ref}/breaking-changes.html[Breaking changes] for more information.
+We’d like to alert users to a bug in Logstash 6.0.0 that can cause errors writing to Elasticsearch 6.0+ clusters. When the `type` field is set to more than one value per Elasticsearch index Logstash will encounter errors while indexing documents. Logstash 6.0 uses the value of the `type` field to set the Elasticsearch `type` by default. Elasticsearch 6 [no longer supports more than one type]((https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html) per index. In Logstash 6.0.1 we will fix this bug by changing the behavior so that when communicating with an Elasticsearch 6.0+ cluster Logstash will not set the document type by default, but rather, will set it to the value `doc`, regardless of whether an event’s `[type]` field has been set. The type information of a document will still be available under the field `type`.
```
jsvd: we should provide an example message of the error that you can encounter, so this SEOs easily. issue https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/708 provides one:

```
[2017-11-21T14:26:01,991][WARN ][logstash.outputs.elasticsearch] Could not index event to Elasticsearch. {:status=>400, :action=>["index", {:_id=>nil, :_index=>"packetbeat-6.0.0-2017.11.21", :_type=>"flow", :_routing=>nil}, #<LogStash::Event:0x4af7bf6>], :response=>{"index"=>{"_index"=>"packetbeat-6.0.0-2017.11.21", "_type"=>"flow", "_id"=>"jEL43l8Bp0AXkoCMsC5q", "status"=>400, "error"=>{"type"=>"illegal_argument_exception", "reason"=>"Rejecting mapping update to [packetbeat-6.0.0-2017.11.21] as the final mapping would have more than 1 type: [icmp, flow]"}}}}
```

jsvd: This should be more focused on the scenario that triggers the problem:
Writing events that have more than one value for the `type` field into the same index of an ES 6.x cluster.
Because by default we fetch the document_type from this `type` field, when writing to ES 6.x you must disable this behaviour by setting the type explicitly to a single value like `document_type => "log"`.

andrewvc:
@jsvd I've made a number of changes, your thoughts?

docs/static/upgrading.asciidoc
```diff
@@ -66,12 +66,23 @@ Before upgrading Logstash, remember to read the <<breaking-changes,breaking chan
...
- {ref}/removal-of-types.html[Removal of mapping types] and {ref}/breaking-changes.html[Breaking changes] for more information.
+We’d like to alert users to behavior in Logstash 6.0.0 that can cause errors writing to Elasticsearch 6.0+ clusters. When the `type` field has different values for different events per Elasticsearch index Logstash will encounter errors while indexing documents. These errors look something like the example below, which has been shortened from the full message. This bug is likely to be encountered by users using either Beats, or multiple inputs with different `type` values.
 
+`[2017-11-21T14:26:01,991][WARN ][logstash.outputs.elasticsearch] Could not index event to Elasticsearch. {:status=>400,  :response=>{"error"=>{"reason"=>"Rejecting mapping update to [myindex] as the final mapping would have more than 1 type: [type1, type2]"}}}}`
+
+The workaround for Logstash 6.0.0 is to add the setting `document_type => doc` to the Elasticsearch output configuration. Logstash 6.0.1+ will not require this workaround as that will become the new default behavior for Logstash when communicating with Elasticsearch 6.0+.
```
jsvd: > Logstash 6.0.1+ will not require this workaround as that will become the new default behavior for Logstash when communicating with Elasticsearch 6.0+.

I'd be a little concerned to promise this, is there a problem with adding this sentence in present tense when we implement and deliver the 6.0.1 release?

andrewvc: I would call it a blocker for LS 6.0.1, but you're right, it probably is prudent not to promise this until its released. I'll replace it with 'a future version'

ph:
@andrewvc LGTM for the first step.

docs/static/upgrading.asciidoc
```diff
@@ -66,12 +66,23 @@ Before upgrading Logstash, remember to read the <<breaking-changes,breaking chan
...
+
+Logstash has historically used the value of the `type` field to set the Elasticsearch `type` by default. Elasticsearch 6 [no longer supports more than one type]((https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html) per index. This is why the new behavior will only be applied to Elasticsearch 6.0 + clusters with our upcoming fixes.
+
+Please read on for more information about document types with Logstash and Elasticsearch 6.0
+
+==== Handling Document Types in Elasticsearch 6.0+
+
+As of Elasticsearch 6.0 document types are [on the way out](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html) with only a single mapping type per index supported. For Logstash users this means transitioning to using the `type` field inside of the document instead of the document type. The effect is the same, but the usage is slightly different. This may mean reconfiguring existing Kibana dashboards to use this new field in lieu of the document type.
```
jsvd: > For Logstash users this means transitioning to using the `type` field inside of the document instead of the document type.

I'm not sure what is the intent with this sentence?

andrewvc: The idea (which I apparently did a poor job of explaining) was that you shouldn't rely on `_type` but rather `type`.

andrewvc: It's reaalllly confusing to explain this, since both things are named `type`.

andrewvc:
@jsvd I agree that this could be improved. I'm going to merge this as is for now. We can improve this more later.

PR #8712, opened 2017-11-22
MINOR: Make ITs use same RSpec version used by the main project

original-brownbear:
I just saw a Jenkins run bust randomly on failure to resolve RSpec `3.1` in the ITs.
We should at least align the RSpec version with what is used by the core run (since we transitively pull LS core into the IT gem) to make things a little more deterministic here RSpec version wise.

=> Aligned RSpec and Stud version with core


colinsurprenant:
LGTM

colinsurprenant:
maybe we should see if/how we can avoid future discrepancies between both dependencies lists? (in another issue)

original-brownbear:
@colinsurprenant yea we should fix this for good. I actually think there's an easy way to do this by reusing the RSPec JUnit-wrapper we have in the core RSpec tests.

Will look into that soon.

PR #8714, opened 2017-11-22
Bump minitar version for security fix even though we aren't vulnerable

andrewvc:
This lib has a vuln, so we should bump it, even though we aren't vulnerable ourselves.


original-brownbear:
@andrewvc LGTM (assuming it goes green :))

PR #8715, opened 2017-11-22
 #7403 follow up, fixing unsorted collection in Graph#chain

original-brownbear:
Now that we don't sort the vertex hashes anymore after #8695, this piece of random sorting is hitting us if we run the test I changed to run 10x repeatedly.

Failed here:
https://logstash-ci.elastic.co/job/elastic+logstash+6.x+multijob-unix-compatibility/os=oraclelinux/231/console

Pretty obvious fix in my opinion, since the leaves we're mapping and filtering there are all coming from `LinkedHashMap`s we don't need to collect to a set => list is sorted and safe here.


original-brownbear:
thanks, merging to `master` and `6.x` other affected branches need the back port of #8695 first (working on it)

PR #8718, opened 2017-11-23
Edit addition to upgrade docs

dedemorton:
@andrewvc I've made a pass at editing the explanation you added to the upgrade docs. I'm going to add some additional feedback as comments here because there were a few other sentences that I felt were unclear, but didn't fix.

### DIFF
--- a/docs/static/upgrading.asciidoc
+++ b/docs/static/upgrading.asciidoc
@@ -68,21 +68,33 @@ If you are installing Logstash with other components in the Elastic Stack, also 
 
 ==== Issue with Logstash 6.0.0 `document_type` when writing to Elasticsearch 6.x
 
-We’d like to alert users to behavior in Logstash 6.0.0 that can cause errors writing to Elasticsearch 6.0+ clusters. When the `type` field has different values for different events per Elasticsearch index Logstash will encounter errors while indexing documents. These errors look something like the example below, which has been shortened from the full message. This bug is likely to be encountered by users using either Beats, or multiple inputs with different `type` values.
+We’d like to alert users to behavior in Logstash 6.0.0 that can cause errors when writing to Elasticsearch 6.0+ clusters. When Logstash attempts to index events that result in multiple `type` values, Logstash encounters indexing errors. These errors look similar to the following example, which has been shortened from the full message:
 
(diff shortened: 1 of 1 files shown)


docs/static/upgrading.asciidoc
```diff
@@ -68,21 +68,33 @@ If you are installing Logstash with other components in the Elastic Stack, also
...
+event to Elasticsearch.{:status=>400,  :response=>{"error"=>{"reason"=>"Rejecting
+mapping update to [myindex] as the final mapping would have more than 1 type:
+[type1, type2]"}}}}
+----
 
-The workaround for Logstash 6.0.0 is to add the setting `document_type => doc` to the Elasticsearch output configuration. We will issue a patch to address this issue soon in a new version of Logstash. 
+Users are likely to encounter this error when Logstash is receiving data from:
 
-Logstash has historically used the value of the `type` field to set the Elasticsearch `type` by default. Elasticsearch 6 [no longer supports more than one type]((https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html) per index. This is why the new behavior will only be applied to Elasticsearch 6.0 + clusters with our upcoming fixes.
+* multiple types of Beats
```
dedemorton: I realize this takes up a bit more space, but I think it's worth listing the situations where users might encounter the problem so that they can be proactive about solving it. The previous explanation seemed a bit too terse.

docs/static/upgrading.asciidoc
```diff
@@ -68,21 +68,33 @@ If you are installing Logstash with other components in the Elastic Stack, also
...
 
 Please read on for more information about document types with Logstash and Elasticsearch 6.0
 
 ==== Handling Document Types in Elasticsearch 6.0+
 
-As of Elasticsearch 6.0 document types are [on the way out](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html) with only a single mapping type per index supported. For Logstash users this means transitioning to using the `type` field inside of the document instead of the document type. The effect is the same, but the usage is slightly different. This may mean reconfiguring existing Kibana dashboards to use this new field in lieu of the document type.
+As of Elasticsearch 6.0, document types are https://www.elastic.co/guide/en/elasticsearch/reference/6.0/removal-of-types.html[on the way out], and only a single mapping type per index is supported. For Logstash users this means transitioning to using the `type` field inside of the document instead of the document type. The effect is the same, but the usage is slightly different. This may mean reconfiguring existing Kibana dashboards to use the new `type` field instead of the document type.
```
dedemorton: We recommend using "`type` field inside of the document instead of the document type." instead of the document type. I think this is potentially confusing. Maybe an example would help.

PR #8729, opened 2017-11-27
ensure logstash terminates on second sigint

jsvd:
As more java code is being used in logstash, it's likely that
java.lang.Threads which aren't set as daemon may be running when
`exit(1)` is called.

Because System.exit only raises a SystemExit exception, it's possible
that threads will continue to execute, and if they're not set as daemon
then the JVM won't terminate.

By calling `System.exit!` instead we ensure jruby calls the underlying
java.lang.System.exit and logstash terminates immediately.


original-brownbear:
@jsvd perfect, this change explains a lot in another issue I was looking at btw :D thanks! => LGTM

PR #8730, opened 2017-11-27
 #8728 Fix RubyTimestamp's '<=>' implementation to correctly compare to other types

### LINKED ISSUE #8728: Logstash 6.0 breaks s == 'str' comparison if s is not String
After upgrading from LS 5.6.0 to 6.0.0 following code stopped to work:
```
      hash = event.to_hash
      hash.each do |k,v|
        if v == '-'
          event.remove(k)
        else
          if v.is_a? String
            v.strip!
            v.gsub!('`', ',')
          end
        end
      end
```
To make it work again I have to do reverse the comparison, so:
`if '-' == v`
The reason is described by guyboertje in this discussion: https://discuss.elastic.co/t/logstash-6-0-breaks-ruby-event-iteration/109084/2:

> when v is the Timestamp the v == '-' is calling the ==(other) method on the Timestamp instance and this does a naive other.time without checking whether other (the String '-') is the same Class as v.
By switching them around the `==(other) call no happens on the String class and it in not so naive - it does better equality checks.

- Version: 6.0.0
- Operating System: RHEL/Centos 7.0, docker containers

original-brownbear:
It's in the title, fixes #8728 by:

* Moving non-type conscious Ruby `<=>` to type conscious native Java implementation

### DIFF
--- a/logstash-core/lib/logstash/timestamp.rb
+++ b/logstash-core/lib/logstash/timestamp.rb
@@ -7,11 +7,6 @@ module LogStash
   class Timestamp
     include Comparable
 
-    # TODO (colin) implement in Java
-    def <=>(other)
-      self.time <=> other.time
-    end
-
     def eql?(other)
       self.== other
     end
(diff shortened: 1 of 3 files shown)


colinsurprenant:
LGTM

original-brownbear:
@colinsurprenant thanks!

original-brownbear:
note to self, `6.0` need manual backport

Issue #8168, opened 2017-09-07
DLQ Integration Test: Logstash should not increment a metric with a negative value


robbavey:
During the running of the dlq integration test, occasionally Travis fails with the following error:

```
Exception in thread "Ruby-0-Thread-10@[main]>worker0: :1" java.lang.IllegalArgumentException: Counters can not be incremented by negative values
	at org.logstash.instrument.metrics.counter.LongCounter.<clinit>(org/logstash/instrument/metrics/counter/LongCounter.java:15)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(java/lang/Class.java:348)
	at org.jruby.javasupport.JavaSupportImpl.loadJavaClass(org/jruby/javasupport/JavaSupportImpl.java:154)
	at org.jruby.javasupport.Java.getProxyClassOrNull(org/jruby/javasupport/Java.java:940)
	at org.jruby.javasupport.Java.getProxyClassOrNull(org/jruby/javasupport/Java.java:927)
	at org.jruby.javasupport.Java.getProxyOrPackageUnderPackage(org/jruby/javasupport/Java.java:885)
	at org.jruby.javasupport.JavaPackage.method_missing(org/jruby/javasupport/JavaPackage.java:259)
	at org.jruby.javasupport.JavaPackage$INVOKER$i$method_missing.call(org/jruby/javasupport/JavaPackage$INVOKER$i$method_missing.gen)
	at RUBY.<main>(:1)
	at org.jruby.Ruby.runInterpreter(org/jruby/Ruby.java:839)
	at org.jruby.Ruby.loadFile(org/jruby/Ruby.java:2876)
	at org.jruby.RubyKernel.requireCommon(org/jruby/RubyKernel.java:968)
	at org.jruby.RubyKernel.require(org/jruby/RubyKernel.java:961)
	at org.jruby.RubyKernel$INVOKER$s$1$0$require19.call(org/jruby/RubyKernel$INVOKER$s$1$0$require19.gen)
	at RUBY.<main>(:1)
	at org.jruby.Ruby.runInterpreter(org/jruby/Ruby.java:839)
	at org.jruby.Ruby.loadFile(org/jruby/Ruby.java:2876)
	at org.jruby.RubyKernel.requireCommon(org/jruby/RubyKernel.java:968)
	at org.jruby.RubyKernel.require(org/jruby/RubyKernel.java:961)
[...]

PR #8731, opened 2017-11-27
#8172 Move timestamp handling for metrics to java.lang.System.nano_time

robbavey:
Fixes #8180

Backport to 6.0 -  see https://github.com/elastic/logstash/issues/8168#issuecomment-347133562 for instance in the wild with PQ

### DIFF
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -44,9 +44,9 @@ module LogStash
     def multi_filter(events)
       @metric_events_in.increment(events.size)
 
-      start_time = java.lang.System.current_time_millis
+      start_time = java.lang.System.nano_time
       new_events = @filter.multi_filter(events)
-      @metric_events_time.increment(java.lang.System.current_time_millis - start_time)
+      @metric_events_time.increment((java.lang.System.nano_time - start_time) / 1_000_000)
 
       # There is no guarantee in the context of filter
       # that EVENTS_INT == EVENTS_OUT, see the aggregates and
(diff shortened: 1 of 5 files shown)


robbavey:
@andrewvc Is this good to go, and approved for 6.0.1?

logstash-core/lib/logstash/filter_delegator.rb
```diff
@@ -44,9 +44,9 @@ def config_name
     def multi_filter(events)
       @metric_events_in.increment(events.size)
 
-      start_time = java.lang.System.current_time_millis
+      start_time = java.lang.System.nano_time
       new_events = @filter.multi_filter(events)
-      @metric_events_time.increment(java.lang.System.current_time_millis - start_time)
+      @metric_events_time.increment((java.lang.System.nano_time - start_time) / 1_000_000)
```
redbaron4: Is it possible for `nano_time - start_time` to be negative (say when NTP has caused clock to jump "back")? Can there be a check which calls increment only if the difference is positive?

jordansissel: Java's nanotime does not use wallclock time (wallclock is affected by ntp adjustments). It is not expected to go backwards.

From the [System.nanoTime docs](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--):
> This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time

Given this, it should be safe to use for duration measurements (where currentTimeMillis is not because it is based on wallclock time and impacted by ntp)

redbaron4: @jordansissel That makes sense. Thanks for clarifying.

PR #8748, opened 2017-11-28
Fix return code and finding Java in system-install

lukebigum:
There are two bugs in _./bin/system-install_ that make it hard to work with in Puppet and your [puppet-logstash](https://github.com/elastic/puppet-logstash) module.

The return code of the JRuby call is trapped in a variable, but subsequent echos and rms mean that the system-install Bash will always return with zero, indicating it has always run successfully.  Puppet treats the return code of an Exec as gospel so we should really return the same exit code of the JRuby call so any error gets propagated to Puppet.

Second, I've moved the call to the setup() function to after loading the environment from startup.options. If you have a non-standard Java path like in /opt, you need JAVA_HOME set before setup() is called or the script won't be able to find the path to the Java binary.

### DIFF
--- a/bin/system-install
+++ b/bin/system-install
@@ -1,9 +1,5 @@
 #!/bin/bash
 
-unset CDPATH
-. "$(cd `dirname $0`/..; pwd)/bin/logstash.lib.sh"
-setup
-
 if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


jordansissel:
I agree with proposal  and the code looks good. I have not tested.

Issue #8701, opened 2017-11-21
Logstash-6.0 not working when config contains nested hash


sw-jung:
I have developed and been using plugin [logstash-filter-memoize](https://github.com/sw-jung/logstash-filter-memoize).

I have [an issue](https://github.com/sw-jung/logstash-filter-memoize/issues/1) that the plugin does not work in logstash-6, There was a problem at compiling the config file.

The problem occurs when config file contains a nested hash. It was compiled normally until 5.x. Is not support nested hash?

Currently, this seems to be a problem only in my plugin, but I think nested hash is useful structure and it can be used in other plugins.

I would be appreciated if you could give any opinion on this issue

---

- Version: 6.0.0
- Operating System: centos-release-7-4.1708.el7.centos.x86_64
- Config File (if you have sensitive info, please remove it):
```ruby
input {
  file {
    path => "/var/log/logstash/test.log"
    codec => "json"
  }
}

filter {
  memoize {
    key => "%{host}"
    fields => ["host_owner", "host_location"]
    filter_name => "elasticsearch"
    filter_options => {
      query => "host:%{host}"
      index => "known_host"
      fields => { # <-- This is nested hash!
        "host_owner" => "host_owner"
        "host_location" => "host_location"
      }
    }
    ttl => 5
  }
}

output {
  stdout {
    codec => rubydebug
  }
}
```
- Sample Data:
```bash
$ echo '{"host": "test"}' >> /var/log/logstash/test.log
```
- Steps to Reproduce:
1. Install logstash-6.0.0
2. Install logstash-filter-memoize
```bash
$ bin/logstash-plugin install logstash-filter-memoize
```
3. Set config file above
4. Run! and I get following errors:
```
[...]

PR #8750, opened 2017-11-28
LSCL/LIR: enable plugins to have nested hash directives

yaauie:
A regression was introduced in the new LSCL/LIR implementation in 6.0, in
which configurations that had nested hashes produced valid treetop-ASTs that
could not produce a valid LIR; the values in an `LSCL::AST::Hash` were
assumed to implement `ValueExpression`, but `LSCL::AST::Hash#expr` produced
an unwrapped `RubyHash`.

By making `LSCL::AST::Hash#expr` emit a `ValueExpression` _representing_ a
`RubyHash`, and relying on the existing code that deals with `ValueExpression`
appropriately, we can re-enable deeply-nested hashes in configs.

Resolves: elastic/logstash#8701
Resolves: logstash-plugins/logstash-output-http#77
Resolves: sw-jung/logstash-filter-memoize#1

### DIFF
--- a/logstash-core/lib/logstash/compiler/lscl.rb
+++ b/logstash-core/lib/logstash/compiler/lscl.rb
@@ -198,7 +198,7 @@ module LogStashCompilerLSCLGrammar; module LogStash; module Compiler; module LSC
 
     def expr
       validate!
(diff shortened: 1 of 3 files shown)


jordansissel:
Tests passing. LGTM

PR #8754, opened 2017-11-28
test fix: pin rack-test to 0.7.0. newer versions require Ruby 2.2.2+

jakelandis:
only needed for 5.x branch(es) of code..


original-brownbear:
@jakelandis LGTM

PR #8758, opened 2017-11-29
Bump minitar version in .lock.release to 0.6.1

jakelandis:
Note - this was already bumped in all bump the .lock.release file https://github.com/elastic/logstash/commit/96706a99be380423a9c2e8af7d05e66409b87b20. This change is needed to fix the build, due this section of code :  

```
   # create Gemfile.jruby-1.9.lock from template iff a template exists it itself does not exist
      lock_template = ::File.join(ENV["LOGSTASH_HOME"], "Gemfile.jruby-1.9.lock.release")
      if ::File.exists?(lock_template) && !::File.exists?(Environment::LOCKFILE)
        FileUtils.copy(lock_template, Environment::LOCKFILE)
      end
```

It also seems that we do not copy the .lock.release file to the docker container. I will issue seperate PR for that.


original-brownbear:
LGTM :)

PR #8761, opened 2017-11-29
Add .lock.release to the test docker container

### LINKED ISSUE #8759: Docker containers for CI testing are missing the .lock.release file
Our setup scripts for testing will use the .lock.release file if it exists as the Gemfile.lock 

```
   # create Gemfile.jruby-1.9.lock from template iff a template exists it itself does not exist
      lock_template = ::File.join(ENV["LOGSTASH_HOME"], "Gemfile.jruby-1.9.lock.release")
      if ::File.exists?(lock_template) && !::File.exists?(Environment::LOCKFILE)
        FileUtils.copy(lock_template, Environment::LOCKFILE)
      end
```

However, since we are not adding this file to our Docker containers we use for testing, we may not be testing with the same dependency set as the release. (Fortunately our CI compatibility test actually do use this file since they are not run in containers.)

jakelandis:
Fixes #8759 

5.x PR is https://github.com/elastic/logstash/pull/8762

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,7 @@ ADD bin /opt/logstash/bin
 ADD modules /opt/logstash/modules
 ADD ci /opt/logstash/ci
 ADD settings.gradle /opt/logstash/settings.gradle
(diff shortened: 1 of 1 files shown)


original-brownbear:
LGTM :)

PR #8763, opened 2017-11-29
Minor: remove rake test:core-fail-fast

### LINKED ISSUE #8742: Drop support for rake test:core-fail-fast (or add it to gradle wrapper)
Currently the CI scripts support `rake test:core-fail-fast` which is basically just an alias to set the rspec `--fail-fast` flag when running the Ruby unit tests.  

Currently CI doesn't actually use this option. With the introduction of Gradle wrappers to start replacing the `rake` tasks we should either add this option to the Gradle wrappers, or drop support from our CI scripts. 

Since it is not used, I am in favor of complete removal. 

Any objections ?

jakelandis:
Fixes #8742

### DIFF
--- a/ci/unit_tests.sh
+++ b/ci/unit_tests.sh
@@ -11,13 +11,7 @@ export CI=true
 
 SELECTED_TEST_SUITE=$1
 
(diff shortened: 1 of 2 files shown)


jordansissel:
LGTM

PR #8766, opened 2017-11-29
Release lock for 6.1 branch

jakelandis:
There are 2 commits here, the first is an exact copy from the 6.0 branch, and the second is the result of `./gradlew bootstrap`

The comparison here should be between those two diffs.

### DIFF
--- a/Gemfile.jruby-2.3.lock.release
+++ b/Gemfile.jruby-2.3.lock.release
@@ -1,0 +1,728 @@
+PATH
+  remote: ./logstash-core
+  specs:
+    logstash-core (6.0.1-java)
+      chronic_duration (= 0.10.6)
+      clamp (~> 0.6.5)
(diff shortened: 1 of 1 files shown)


jsvd:
I have had to deal with the bump of es output for 6.1 and 6.0.1, so I have a commit with the updated lock file for 6.1 branch, https://github.com/jsvd/logstash/commit/dbe17e87167ea8be62d806eb7424b5a0927f4ba4
you can cherry-pick that for this PR if you want

jakelandis:
@jsvd - I updated it with the results of `bundle lock` since if we are trying to just get the prod dependencies that is the most standard way.  I diffed this against `rake artifact:tar` and it is the same except for a tiny bit ordering.   I hope to standardize on running `bundle lock` via a different conversation.

jakelandis:
@jsvd - thanks!

PR #8768, opened 2017-11-29
test fix: pin rake gem to Ruby 1.x compatible version

### LINKED ISSUE #8746: 5.6 core builds are broken
```
22:37:18 An error occurred while installing rack-test (0.8.2), and Bundler cannot continue.
22:37:18 Make sure that `gem install rack-test -v '0.8.2'` succeeds before bundling.
```

manually trying to install results in : 
```
Clogstash@5da43e330f69:/opt/logstash$ gem install rack-test -v 0.8.2
Fetching: rack-test-0.8.2.gem (100%)
ERROR:  Error installing rack-test:
	rack-test requires Ruby version >= 2.2.2.
```

It appears that this commit is the culprit: https://github.com/elastic/logstash/commit/96706a99be380423a9c2e8af7d05e66409b87b20

I reverted locally and it fixed the build. Revert PR incoming.

jakelandis:
This should fix the 5.6 builds. 

Fixes #8746

### DIFF
--- a/qa/Gemfile
+++ b/qa/Gemfile
@@ -1,5 +1,5 @@
 source "https://rubygems.org"
 gem "runner-tool", :git => "https://github.com/purbon/runner-tool.git"
 gem "rspec", "~> 3.1.0"
-gem "rake"
+gem "rake", "12.1.0"
 gem "pry", :group => :test


qa/Gemfile
```diff
@@ -1,5 +1,5 @@
 source "https://rubygems.org"
 gem "runner-tool", :git => "https://github.com/purbon/runner-tool.git"
 gem "rspec", "~> 3.1.0"
-gem "rake"
+gem "rake", "12.1.0"
```
yaauie: Should this be `~> 12.1.0` to allow us to discover patch releases (e.g., for security fixes), or is a hard-pin nexessary?

jakelandis: I don't think rake even tries to follow semver. :) 

For the 5.x branch, I think we should be explicit as possible to help ensure nothing sneaks into a release (or causes instability in the release process).

PR #8770, opened 2017-11-29
fix queue open bug for fully acked in-between pages plus new test

### LINKED ISSUE #8769: NPE when opening existing queue with in-between fully acked pages
The `Queue.getPersistedByteSize()` method produces a NPE when opening an existing queue that contains in-between fully acked pages. 

A queue with fully acked middle pages will happen in a situation where a filter or output plugin is stalled and does not ack a batch which keeps its data page in place while new pages will be fully acked and eventually purged, see #7763 and #7796 

The error looks like this (this is on master but will happen down to 5.x). 

```
Failed to execute action {:action=>LogStash::PipelineAction::Create/pipeline_id:main, :exception=>"Java::JavaLang::NullPointerException", :message=>"", :backtrace=>["org.logstash.ackedqueue.Queue.lambda$getPersistedByteSize$0(Queue.java:134)", "java.util.stream.ReferencePipeline$5$1.accept(ReferencePipeline.java:227)", "java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1374)",
```

Incoming PR

colinsurprenant:
Fixes #8769 

This removes the insertion of dummy pages in the `tailPages` collection for in-between fully acked pages in the `open()` on an existing queue. This is completely unnecessary since the `ack()` method actually removes these pages in the normal process. 

Added a new tests to harness this edge case. Also validated manually with existing queue data which surfaced this problem.

This is pretty severe and should be prioritized for merging into 6.0.1 and downto 5.6.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
@@ -269,11 +269,6 @@ public final class Queue implements Closeable {
             if (this.tailPages.size() == 0) {
                 // this is the first tail page and it is fully acked so just purge it
                 this.checkpointIO.purge(this.checkpointIO.tailFileName(checkpoint.getPageNum()));
-            } else {
-                // create a tail page with a null PageIO and add it to tail pages but not unreadTailPages
-                // since it is fully read because also fully acked
-                // TODO: I don't like this null pageIO tail page...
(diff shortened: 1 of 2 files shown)


jakelandis:
Discussed on Zoom with @colinsurprenant  and @yaauie and changes LGTM

colinsurprenant:
merged into master, 6.x, 6.1 and 5.6

PR #8771, opened 2017-11-30
Update advanced-pipeline.asciidoc

markwalkom:
Minor port change, from `5043` to `5044`, to make things consistent with the beats defaults.


jordansissel:
LGTM

Issue #8688, opened 2017-11-16
Large PQ Performance Degradation


original-brownbear:
Some analysis on the behaviour of larger PQs as requested:

* We do leak memory for larger PQs, even with #8636 merged
   * we don't leak that much though: I'd say about 20MB Heap per 1GB PQ at 250MB page size, that's bad but manageable imo, though at very large queues it might be an issue in practice (all of the leaked memory is `int[]` from the page index) ... 100GB queue will cost you about 2GB in heap overhead
* We see a sharp qualitative change in performance once the PQ size exceeds the amount of free memory reducing the maximum write throughput to (assuming all workers blocked) by almost an order of magnitude with throughput further (but at la slow pace) degrading with increasing queue size
* More and more of the CPU is idle the larger the Queue gets

![screenshot from 2017-11-16 16-32-30](https://user-images.githubusercontent.com/6490959/32899636-cad1eaa6-caeb-11e7-8db0-a4e170e37224.png)

* Sharp decline after 1 min coincides with the time when queue_size + heap_size exceed physically available memory
* The little bump at the right hand side of the graph coincides with the total size of the PQ exceeding the amount of physical memory on the system (not sure what happens here, but throughput stays constant)  
* Graph begins at about 1 GB of PQ, ends at ~17GB => note the extreme slowdown after the initial cliff, the first 2 minutes coincide with producing almost half the data 

After the cliff most of the non-idle CPU time goes to IO, but we're mostly blocked on writer threads while one thread fsyncs:

![screenshot from 2017-11-16 16-27-16](https://user-images.githubusercontent.com/6490959/32899640-cd50c3c4-caeb-11e7-9e86-7742d4d8819c.png)

Ran this on relatively recent Linux:

```sh
Linux debian-armin 4.9.0-4-amd64 #1 SMP Debian 4.9.51-1 (2017-09-28) x86_64 GNU/Linux
```

reasonable ssd:

```sh
Model Family:     Samsung based SSDs
Device Model:     Samsung SSD 850 EVO 500GB
Serial Number:    S1YCNXAGB01633Y
LU WWN Device Id: 5 002538 d40779fe1
```

and pretty standard `ext4`:

```sh
# <file system> <mount point>   <type>  <options>       <dump>  <pass>
# / was on /dev/sda1 during installation
UUID=1aa0aacc-9182-49d9-91d5-5624694ac454 /               ext4   noatime,errors=remount-ro 0       1
```

-----------------------------

That's what we're at. 

[...]

PR #8776, opened 2017-11-30
 #8688 enable explicit System gc calls again to improve GC of DirectByteBuffer

original-brownbear:
See #8688 for what the issue is and https://github.com/elastic/logstash/issues/8688#issuecomment-347946236 for the agreed upon improvement (not a 100% solution, but way improves the problem).


original-brownbear:
@jordansissel thanks merging, failure is the same old #8142

PR #8793, opened 2017-12-04
MINOR: Bring log4j version in line with ES

original-brownbear:
This would be very nice to have for implementing the clustering prototype :) I can't run the ES JUnit framework with log4j `2.6.2`, upgrading it to `2.9.1` which is what ES uses too fixes the issues I'm experiencing.
Since we shouldn't use ancient versions here anyways I don't see why we can't upgrade this one in `master` and `6.x` :)

### DIFF
--- a/logstash-core/build.gradle
+++ b/logstash-core/build.gradle
@@ -119,9 +119,9 @@ idea {
 }
 
 dependencies {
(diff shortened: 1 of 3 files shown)


original-brownbear:
@robbavey thanks merging :) (failure is still #8142 ;))

PR #8797, opened 2017-12-04
change batch delay to 50ms and page size to 64mb

### LINKED ISSUE #8707: change queuing defaults
Per #8702 I suggest we change the following defaults:

```
pipeline.batch.delay: 5
queue.page_capacity: 250mb
```

to these new values:

```
pipeline.batch.delay: 50
queue.page_capacity: 25mb
```

- increasing the `pipeline.batch.delay` does not have any negative incidence and provide more time to form full batches.
- decreasing `queue.page_capacity` mitigate a performance problem which seems related to the number of written elements in each PQ page. The only potential problem I see is that it might cause problems with documents/message sizes > 25mb but seems not likely for the majority of users who will benefit from a smaller default page size.

colinsurprenant:
Fixes #8707 
Relates to #8702, #8674 

Change the default values for `pipeline.batch.delay` from `5ms` to `50ms` and `queue.page_capacity` from `250mb` to `64mb`.

Also fixes documentation & comments inconsistencies.

### DIFF
--- a/config/logstash.yml
+++ b/config/logstash.yml
@@ -44,10 +44,10 @@
 #
 # pipeline.batch.size: 125
 #
-# How long to wait before dispatching an undersized batch to filters+workers
-# Value is in milliseconds.
+# How long to wait in milliseconds while polling for the next event
+# before dispatching an undersized batch to filters+outputs
 #
-# pipeline.batch.delay: 5
+# pipeline.batch.delay: 50
 #
 # Force Logstash to exit during shutdown even if there are still inflight
 # events in memory. By default, logstash will refuse to quit until all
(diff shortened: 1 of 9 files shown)


jordansissel:
@colinsurprenant can you include some reasons/citation(s) in your commit message to explain this change? (for our future selves reading `git log` or similar)

colinsurprenant:
@jordansissel yes I can 👍

colinsurprenant:
@jordansissel let me know if new commit message is better.

jakelandis:
@colinsurprenant - did we ever get to the bottom of _why_ larger pages show less performance the further into the page ?

colinsurprenant:
@jakelandis No. There is some investigation going on in #8688 but nothing specifically about the page size itself is clearly identified. I think a specific issue about this could be useful to track.

jakelandis:
@colinsurprenant - ok, was just curious. Would you mind logging that issue for the followup... perhaps framed as define the optimal page size ?

I don't want to block progress on an known performance issue and 64Mb _feels_ reasonable. 

LGTM2

colinsurprenant:
created #8801 to followup on the large page performance issue.

colinsurprenant:
rebased, merged into master and 6.x (will be included in 6.2).
thanks @yaauie @jakelandis @jordansissel

Issue #7482, opened 2017-06-16
Rotate & Delete Logstash logs by default


gmoskovicz:
With the change to log4j2 , now we can use an `action` inside the `DefaultRolloverStrategy` to delete old log files by default and keep a certain amount of days. We could even add a new `ls.logs.keepdays` setting that defaults to any number and one can modify to change this. 

```
status = error
name = LogstashPropertiesConfig

appender.rolling.type = RollingFile
appender.rolling.name = plain_rolling
appender.rolling.fileName = ${sys:ls.logs}/logstash-${sys:ls.log.format}.log
appender.rolling.filePattern = ${sys:ls.logs}/logstash-${sys:ls.log.format}-%d{yyyy-MM-dd}.gz
appender.rolling.policies.type = Policies
appender.rolling.policies.time.type = TimeBasedTriggeringPolicy
appender.rolling.policies.time.interval = 1
appender.rolling.policies.time.modulate = true
appender.rolling.layout.type = PatternLayout
appender.rolling.layout.pattern = [%d{ISO8601}][%-5p][%-25c] %-.10000m%n
appender.rolling.strategy.type = DefaultRolloverStrategy
appender.rolling.strategy.action.type = Delete
appender.rolling.strategy.action.basepath = ${sys:ls.logs}
appender.rolling.strategy.action.condition.type = IfLastModified
appender.rolling.strategy.action.condition.age = 14D
appender.rolling.strategy.action.PathConditions.type = IfFileName
appender.rolling.strategy.action.PathConditions.glob = logstash-${sys:ls.log.format}-*


appender.json_rolling.type = RollingFile
appender.json_rolling.name = json_rolling
appender.json_rolling.fileName = ${sys:ls.logs}/logstash-${sys:ls.log.format}.log
appender.json_rolling.filePattern = ${sys:ls.logs}/logstash-${sys:ls.log.format}-%d{yyyy-MM-dd}.log
appender.json_rolling.policies.type = Policies
appender.json_rolling.policies.time.type = TimeBasedTriggeringPolicy
appender.json_rolling.policies.time.interval = 1
appender.json_rolling.policies.time.modulate = true
appender.json_rolling.layout.type = JSONLayout
appender.json_rolling.layout.compact = true
appender.json_rolling.layout.eventEol = true

rootLogger.level = ${sys:ls.log.level}
rootLogger.appenderRef.rolling.ref = ${sys:ls.log.format}_rolling
```

[...]

PR #8800, opened 2017-12-05
Adds size cap for each individual log file, compresses on rollover

danhermann:
Addresses some of the requests in #7482:

- Verified locally that the size cap `SizeBasedTriggeringPolicy` functions properly with the `TimeBasedTriggeringPolicy` such that log rollovers occur when either the size cap or a time interval is reached.
- Sets a size cap of 100MB for each individual log file. This limit should be reviewed specifically for its suitability as a default.
- Gzips logs at the point of rollover.

It does not specify a limit on the number of log files to retain. That may be proposed in a separate PR since it would involve enabling the automatic deletion of log files by default in Logstash.


config/log4j2.properties
```diff
@@ -15,25 +15,29 @@ appender.json_console.layout.eventEol = true
 appender.rolling.type = RollingFile
 appender.rolling.name = plain_rolling
 appender.rolling.fileName = ${sys:ls.logs}/logstash-${sys:ls.log.format}.log
-appender.rolling.filePattern = ${sys:ls.logs}/logstash-${sys:ls.log.format}-%d{yyyy-MM-dd}.log
+appender.rolling.filePattern = ${sys:ls.logs}/logstash-${sys:ls.log.format}-%d{yyyy-MM-dd}-%i.gz
```
yaauie: keeping the `log` as part of the suffix would be helpful; I'd recommend ending with `%d{yyyy-MM-dd}-%i.log.gz`.

danhermann: Done.

danhermann:
@yaauie, I pushed a change to keep "log" as part of the extension of the compressed file.

yaauie:
LGTM 🙌🏼

jordansissel:
> It does not specify a limit on the number of log files to retain

I think the importance for users is around this idea: "don't run my machine out of disk space" -- and this PR doesn't do much for improving that.

If anything, I think this makes logging perhaps more challenging by creating even more files (many 100mb files per day, instead of what today's code of 1 per day).

I'm not against this PR, but I don't think it does much to alleviate the size problem?

PR #8812, opened 2017-12-07
Bump lockfile for 5.6.6

andrewvc:


Gemfile.jruby-1.9.lock.release
```diff
@@ -571,7 +571,7 @@ GEM
     sawyer (0.6.0)
       addressable (~> 2.3.5)
       faraday (~> 0.8, < 0.10)
-    sequel (5.2.0)
+    sequel (5.3.0)
```
jsvd: I'm not sure it's worth allowing this upgrade [looking at the changelog](https://github.com/jeremyevans/sequel/blob/master/CHANGELOG#L13-L43), since the changes seem to include stuff like `Drop support for PostgreSQL <8.1 in Database#indexes` (I don't know if we use that).

andrewvc: Postgresql 8.0 was release in 2005. 8.1 was released later that year. I think our support for 8.0 is something I'm comfortable dropping. I think that the number of people using it and Logstash 5 must be zero. I'd prefer we just bump the Sequel version.

andrewvc: I'm more worried about pinning and missing a Sequel fix than a user using an ancient version of PG.

jsvd: > Postgresql 8.0 was released in 2005. 8.1 was released later that year. I think our support for 8.0 is something I'm comfortable dropping. I think that the number of people using it and Logstash 5 must be zero. I'd prefer we just bump the Sequel version.

This PG change was just an example, I don't mean to debate the usage of it which we can only speculate.

> I'm more worried about pinning

I didn't suggest pinning, just not including this diff chunk.

It's your call, we didn't pin 5.3.0 due to known bugs in 5.2, so I just don't see the point in risking weird behaviour and surprises in a patch release. This should be discussed on a broader level regarding the decision process of what should go into a .patch release.

PR #8815, opened 2017-12-07
Add default log retention settings

danhermann:
The default retention setting of 20 files was arrived at by assuming a conservative 50% compression ratio on rotated log files for a total of 1GB (20 files * 50% compression * 100MB max size). Note that the 1GB limit would apply per appender.

Relates to #7842

### DIFF
--- a/config/log4j2.properties
+++ b/config/log4j2.properties
@@ -24,6 +24,8 @@ appender.rolling.layout.type = PatternLayout
 appender.rolling.layout.pattern = [%d{ISO8601}][%-5p][%-25c] %-.10000m%n
 appender.rolling.policies.size.type = SizeBasedTriggeringPolicy
 appender.rolling.policies.size.size = 100MB
+appender.rolling.strategy.type = DefaultRolloverStrategy
+appender.rolling.strategy.max = 30
 
 appender.json_rolling.type = RollingFile
 appender.json_rolling.name = json_rolling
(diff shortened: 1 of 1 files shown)


jakelandis:
I think this should be a 7.x change to avoid changing behavior too much on a minor release. Auto deleting old log files may be seen as non-passive.

Thoughts ?

jordansissel:
> Note that the 1GB limit would apply per log file.

Do you mean per appender (slowlog, json, etc), not per log file?

jakelandis:
I would vote to bump to `30` 

Since we support date or size, which ever comes first... i would assume a (mostly) healthy system would typically hit the date based roll over first. (and I think we will get better then 50% on un-healthy systems)

danhermann:
@jordansissel, ah, yes. I meant per appender, not per log file. I'll edit the description.

danhermann:
@jakelandis, bumped to 30 files. I agree that we should get >50% compression in most cases.

jordansissel:
> I think this should be a 7.x change 

I agree with your reasoning. Maybe for 6.x we can offer this as an alternative. Maybe ship a `log4j2.properties.rotating` or something that people can copy/paste or rename if they wish to use it?

jordansissel:
I don't see any tests added, and maybe that's ok (log4j configs maybe are too complicated to write solid tests for, for this pr). Can someone run a manual test and verify the retention settings work as expected (max 30 files, daily rotation, rotation at 100mb) ?

danhermann:
Jenkins test this please

Issue #8725, opened 2017-11-23
Script to generate report of 3rd party dependencies


mgreau:
For now, the following command is used to generate a report:

```
echo 'File.write(ENV["DEPENDENCIES_REPORTS_DIR"] + "/" + ENV["DEPENDENCY_REPORT"]  , "name,version,url,license\n" + Gem::Specification.all.collect { |g| [g.name, g.version.to_s, g.homepage, g.licenses.join("|")].join(",") }.join("\n"))' | bin/logstash --interactive pry
```

Please provide a script/command to generate [a CSV file](https://tools.ietf.org/html/rfc4180) to list 3rd party dependencies:
1.   The command/script provides an option to specify the output filename and path
        - the option/property provided by the script is named **`csv`** (example: `--csv`)
1. The CSV file generated is formatted as follow:
    - the 1st line contains the header line
    - then, one record per dependency
    - if a value for a field is not found, let it empty (ie: `"…",,"…"`)
1. The CSV file contains the following fields 
   - `name`: dependency name (name that identifies the library)
     - example for Java dependency: `groupId:artifactId`
     - example for Go dependency: `repository URI`
   - `version`: dependency version
   - `revision`: VCS commit hash
       - **optional** if the `version` is defined.
   - `url`: dependency URL (website, github repository…)
   - `license`: the license SPDX short identifier (ie: MIT, Apache-2.0…) 
      - the full list is available on [spdx.org website](https://spdx.org/licenses/ )
      - in the case of [license exceptions](https://spdx.org/licenses/exceptions-index.html), use the new `WITH` operator (ie: _GPL-3.0 WITH GCC-exception-3.1_)
      - in case of multiple licenses, delimit them with `|` (example: `EPL-1.0|GPL-2.0|LGPL-2.1`)
   - `copyright` (**optional**): authorship, company...
      - extract this value from the license file content

#### Example
```
"name","version","revision","url","license","copyright"
"jruby-openssl","0.9.21",,"https://github.com/jruby/jruby-openssl",EPL-1.0|GPL-2.0|LGPL-2.1,
```

Thanks.

jordansissel:
(notes)

We have a few groups of dependencies:

1) explicit gem dependencies via bundler/rubygems
2) explicit java dependencies via maven
3) JRuby

jordansissel:
@mgreau I may not have URLs for maven-sourced dependencies initially.

jordansissel:
[...]

PR #8837, opened 2017-12-13
Add bin/dependency-list tool.

jordansissel:
Fixes https://github.com/elastic/logstash/issues/8725

Rough goal is to present license information for all dependencies. Minimal delivery is for Rubygems, not jars/etc, so if the Jar approach in this PR is not cool, I can remove it.

----

Copying the description from #8725:

----

Please provide a script/command to generate [a CSV file](https://tools.ietf.org/html/rfc4180) to list 3rd party dependencies:
1.   The command/script provides an option to specify the output filename and path
        - the option/property provided by the script is named **`csv`** (example: `--csv`)
1. The CSV file generated is formatted as follow:
    - the 1st line contains the header line
    - then, one record per dependency
    - if a value for a field is not found, let it empty (ie: `"…",,"…"`)
1. The CSV file contains the following fields 
   - `name`: dependency name (name that identifies the library)
     - example for Java dependency: `groupId:artifactId`
     - example for Go dependency: `repository URI`
   - `version`: dependency version
   - `revision`: VCS commit hash
       - **optional** if the `version` is defined.
   - `url`: dependency URL (website, github repository…)
   - `license`: the license SPDX short identifier (ie: MIT, Apache-2.0…) 
      - the full list is available on [spdx.org website](https://spdx.org/licenses/ )
      - in the case of [license exceptions](https://spdx.org/licenses/exceptions-index.html), use the new `WITH` operator (ie: _GPL-3.0 WITH GCC-exception-3.1_)
      - in case of multiple licenses, delimit them with `|` (example: `EPL-1.0|GPL-2.0|LGPL-2.1`)
   - `copyright` (**optional**): authorship, company...
      - extract this value from the license file content

#### Example
```
"name","version","revision","url","license","copyright"
"jruby-openssl","0.9.21",,"https://github.com/jruby/jruby-openssl",EPL-1.0|GPL-2.0|LGPL-2.1,
```

Thanks.


mgreau:
Thanks for your work @jordansissel 

I have tested it and I've attached the [`csv` file generated](https://github.com/elastic/logstash/files/1560989/deps.csv.txt)
I have some questions:
1. can you add the header line (`name,version,url,license`)?
1. is it possible to remove the `logstash-*` dependencies?
1. is it possible to have SPDX identifiers in the license field by checking text like other projects (e.g.: [beats](https://github.com/elastic/beats/blob/master/dev-tools/generate_notice.py#L151-L230), [elasticsearch](https://github.com/elastic/elasticsearch/pull/27727/files#diff-b77b8b827252d973b6984ff0a0a18591R133))

If I understand, the file `COPYING.csv` is for JRuby dependencies, so:

4. can you update the license fields (cf. `3.` described above)?
5. is it possible to add the content of this file `COPYING.csv` to the content of the generated file (to have one single CSV file)?

Thanks

jordansissel:
> is it possible to remove the logstash-* dependencies?

I don't understand the removal request here, but yes I can remove them.

> is it possible to have SPDX identifiers in the license field 

I can take a pass, but it won't be perfect because the field is free-form. It's also not something we can audit easily, so there are likely to be mistakes.

> can you update the license fields (cf. 3. described above)?

Yes, but this is not something we can audit easily, so there are likely to be mistakes.

> is it possible to add the content of this file COPYING.csv 

Yes

jordansissel:
@mgreau done.

mgreau:
@jordansissel I added 2 commits. 

- I have tested the script on both `master` and `6.1` branches. 
- I have an error on `5.6`  ```NameError: uninitialized constant Bundler
  const_missing at org/jruby/RubyModule.java:3345
         <main> at logstash-core/lib/logstash/dependency_report_runner.rb:16```

1. Let me know if my commits are ok for you (I'm newbie in Ruby so I'm pretty sure that my updates could be optimized)
1. Let me know if you want me to take care of merging into master and backporting into `6.x` and `6.1`

Thanks

andrewvc:
I'll review this today

andrewvc:
@mgreau that's because the environment needs to be reinitialized on branch switches. https://github.com/elastic/logstash#building-logstash is what you want

andrewvc:
Squashed and merged in #8881

PR #8854, opened 2017-12-18
bump core plugin api to 2.1.29

jsvd:
Latest release of the gem was https://rubygems.org/gems/logstash-core-plugin-api/versions/2.1.28-java with logstash-core 5.6.4


jordansissel:
As a correction for a mistake, LGTM.

I think we should soon look over our plugin API and determine if `2.1.x` is still correct. I"m certain we've broken things or added new interface pieces since 5.0

andrewvc:
@jordansissel everything we've broken we've internally discussed as 'not covered under this API', AFAIK. Things like use of metrics API for instance.

PR #8861, opened 2017-12-19
added new PQ checking tool bin/pqcheck

colinsurprenant:
`bin/pqcheck` new tool to verify a PQ dir files structure. 

This is based on the original one-liner script @jordansissel produced as part of a diagnostic thread.

This tool can take dir path as parameter or use the default `data/queue/main` and will look into each checkpoint files and print it content. Il also prints the corresponding page file size and prints if it is fully acked or not. 

example output:
```
$ bin/pqcheck
checking queue dir: data/queue/main
checkpoint.head, FA, 67108864, {:version=>1, :page=>141, :firstUnackedPage=>141, :firstUnackedSeq=>17075891, :minSeq=>16974637, :elementCount=>101254, :crc32=>3485132876}
checkpoint.1, FA, NA, {:version=>0, :page=>0, :firstUnackedPage=>0, :firstUnackedSeq=>0, :minSeq=>0, :elementCount=>0, :crc32=>0}
```

### DIFF
--- a/bin/pqcheck
+++ b/bin/pqcheck
@@ -1,0 +1,31 @@
+#!/usr/bin/env bin/ruby
+
+DEFAULT_PQ_DIR = "data/queue/main"
+
+argv0 = ARGV[0].to_s.strip
+if argv0 == "-h" || argv0 == "--help"
+  puts("usage: pqcheck [PQ dir path]\n  default [PQ dir path] is #{DEFAULT_PQ_DIR}")
+  exit(0)
+end
+
+dir = argv0.empty? ? DEFAULT_PQ_DIR : argv0
+if (!Dir.exist?(dir))
+  puts("error: invalid PQ dir path: #{dir}")
+  exit(1)
+end
+
+puts("checking queue dir: #{dir}")
+
+Dir.glob("#{dir}/checkpoint.*").sort_by { |x| x[/[0-9]+$/].to_i }.each do |checkpoint|
+  data = File.read(checkpoint)
+
+  if data.size == 33
+    version, page, firstUnackedPage, firstUnackedSeq, minSeq, elementCount, crc32 = data.unpack("nNNQ>Q>NN")
+    fa = firstUnackedSeq >= (minSeq + elementCount)
+    ps = File.exist?("#{dir}/page.#{page}") ? File.size("#{dir}/page.#{page}") : nil
+    print("#{File.basename(checkpoint)}, fully-acked: #{fa ? "YES" : "NO"}, page.#{page} size: #{ps ? ps : "NOT FOUND"}, ")
+    p(version: version, page: page, firstUnackedPage: firstUnackedPage, firstUnackedSeq: firstUnackedSeq, minSeq: minSeq, elementCount: elementCount, crc32: crc32)
+  else
+    puts("#{File.basename(checkpoint)}, invalid size: #{data.size} ")
+  end
+end


bin/pqcheck
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bin/ruby
```
jordansissel: I think this should invoke Logstash's vendored jruby, right?

yaauie:
~~~
env: bin/ruby: No such file or directory
~~~

I think you mean:
~~~ diff
-#!/usr/bin/env bin/ruby
+#!/usr/bin/env ruby
~~~

bin/pqcheck
```diff
@@ -0,0 +1,26 @@
...
+dir = argv0.empty? ? DEFAULT_PQ_DIR : argv0
+if (!Dir.exist?(dir))
+  puts("error: invalid PQ dir path: #{dir}")
+  exit(1)
+end
+
+puts("checking queue dir: #{dir}")
+
+Dir.glob("#{dir}/checkpoint.*").sort_by { |x| x[/[0-9]+$/].to_i }.each do |checkpoint|
+  data = File.read(checkpoint)
```
yaauie: If valid checkpoints are all in theory 33 bytes, should we just read the first 33 bytes (or 33+1 bytes)?

bin/pqcheck
```diff
@@ -0,0 +1,26 @@
...
+  version, page, firstUnackedPage, firstUnackedSeq, minSeq, elementCount, crc32 = (data.size == 33) ? data.unpack("nNNQ>Q>NN") : [0,0,0,0,0,0,0]
```
yaauie: if the checkpoint is not 33 bytes, we proceed as if it were encoded with all zeroes; should we instead emit a warning, or are there cases where it is valid for the checkpoint to be a different size?

bin/pqcheck
```diff
@@ -0,0 +1,26 @@
...
+  fa = firstUnackedSeq >= (minSeq + elementCount)
+  ps = File.exist?("#{dir}/page.#{page}") ? File.size("#{dir}/page.#{page}") : nil
+  print("#{File.basename(checkpoint)}, #{fa ? "FA" : "UN"}, #{ps ? ps : "NA"}, ")
```
yaauie: output isn't particularly clear; I presume `FA` to mean "fully acknowledged" here, but am unsure what `UN`  means in this context.

bin/pqcheck
```diff
@@ -0,0 +1,26 @@
...
+  p(version: version, page: page, firstUnackedPage: firstUnackedPage, firstUnackedSeq: firstUnackedSeq, minSeq: minSeq, elementCount: elementCount, crc32: crc32)
```
yaauie: `Kernel#puts` is perhaps more clear than the short alias `Kernel#p`

bin/pqcheck
colinsurprenant: true, it's not implicitly written it is an invalid checkpoint but there is not such all-zeros checkpoints except for newly created & empty queue...

bin/pqcheck
colinsurprenant: `bin/ruby` is the LS bootstrapping code to execute a "plain" ruby script.

jordansissel: This assumes $PWD is the root of logstash and that the script is invoked `bin/pqcheck`, so to run this in production someone would have to:

```
cd /usr/share/logstash
bin/pqcheck
```

And running `/usr/share/logstash/bin/pqcheck` would fail with a confusing error message as @yaauie showed

colinsurprenant:
I reworked the output format. Now it is more explicit:

```
checking queue dir: data/queue/main
checkpoint.head, fully-acked: YES, page.141 size: 67108864, {:version=>1, :page=>141, :firstUnackedPage=>141, :firstUnackedSeq=>17075891, :minSeq=>16974637, :elementCount=>101254, :crc32=>3485132876}
checkpoint.1, invalid size: 0```

bin/pqcheck
colinsurprenant: I agree but I don't know how to solve this simply. Unless we have practical solutions I think we can live with that caveat since with the on-liner we also had to give instruction WRT the logstash home.

colinsurprenant: this is basically to be better than asking users to copy&paste a one-liner script around.

colinsurprenant: so yeah, the only way IMO would be to split that into 2 files, a bash `pqcheck` and a `pqcheck.rb` with the similar bootstrapping logic as in `bin/ruby` or `bin/logstash-plugins`

colinsurprenant:
thanks @yaauie @jordansissel - I will go ahead and merge this despite the problem of invoking this out of the LS home dir. This is a diagnostic tool and until we find a solution for this I think we can live with that and give instructions for the user to launch from the LS home dir.

colinsurprenant:
merged in master and 6.x

jordansissel:
commits:

* 6.x: 58d2a745bbabf6287ce353459208a92449ca20a3
* master: c0916128770e3ef2152d130578317c4f1cbde400

PR #8866, opened 2017-12-20
port in config is different than used in text below

w32:


jakelandis:
@w32 - nice catch! However, I think we should consistently use port `5044` to better match the [beats documentation](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-beats.html).

Would you mind replacing 5043 with 5044 in this PR ?

w32:
Yes, i think it would be better, because default value in config is 5044.

w32:
Review changes, please

PR #8871, opened 2017-12-20
Require logstash-core to provide LogStash::Error

### LINKED ISSUE #8870: Missing 'logstash-core' require on logstash/errors
Tested on logstash master using the s3 input for development:

```
% grep logstash Gemfile.lock
...
Gemfile.lock:  remote: ../logstash/logstash-core
Gemfile.lock:    logstash-core (7.0.0.alpha1-java)


pork(~/projects/logstash-input-s3) master
% LOGSTASH_SOURCE=1 LOGSTASH_PATH=../logstash bundle exec irb


>> require "logstash/namespace"
=> true
>> LogStash::Error
NameError: uninitialized constant LogStash::Error
<stack trace omitted>

>> require "logstash-core"
=> true
>> LogStash::Error
=> LogStash::Error
```

I'm not exactly sure what the best solution is, here. LogStash::Error comes from Java via JRuby extension. If we keep this, logstash/errors should require "logstash-core"

jordansissel:
Fixes #8870

### DIFF
--- a/logstash-core/lib/logstash/errors.rb
+++ b/logstash-core/lib/logstash/errors.rb
@@ -1,4 +1,5 @@
 # encoding: utf-8
+require "logstash-core" # for LogStash::Error
 
 module LogStash
   class EnvironmentError < Error; end


logstash-core/lib/logstash/errors.rb
```diff
@@ -1,4 +1,5 @@
 # encoding: utf-8
+require "logstash-core" # for LogStash::Error
```
yaauie: it took me an unfortunate amount of digging to find that `LogStash::Error` comes from `org.logstash.RubyUtil`; maybe adding this to the comment?

~~~ diff
-require "logstash-core" # for LogStash::Error
+require "logstash-core" # for LogStash::Error from org.logstash.RubyUtil.LOGSTASH_ERROR
~~~

kares:
closing: *logstash-core/lib/logstash/errors.rb* is an empty stub at this point

PR #8880, opened 2017-12-22
Replace command 'which' with 'command -v' for compatibility

liuchong:
I think we should not check if a command exists which another command, in here 'which'.
Instead we should use a build-in method, in order that script still works in some environments
that, such as a basic docker image, do not have a "which" command.

### DIFF
--- a/bin/logstash
+++ b/bin/logstash
@@ -20,7 +20,7 @@ unset CDPATH
 if [ -L "$0" ]; then
   # Launched from a symlink
   # --Test for the readlink binary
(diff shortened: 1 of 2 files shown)


andrewvc:
jenkins please test this

andrewvc:
Thanks for the patch @liuchong ! Much appreciated, if the CI suite passes I'm glad to merge it :)

andrewvc:
jenkins please test this

liuchong:
Thanks

PR #8884, opened 2017-12-25
 #8883 Fix concurrent modification issue on RubyArray in mock output

### LINKED ISSUE #8883: java_pipeline_spec.rb "with multiple outputs correctly distributes events" Fails Randomly
The new spec I added in #8691 seems to be a little flaky (I have not seen this on Jenkins yet, but locally it failed twice for me now).
It seems at times this line:

```rb
      expect(output.events.size).to eq(60)
```

only sees `59` events. Will track this down and fix it :) 

Failure:

```sh

expected: 60
     got: 59

(compared using ==)

./logstash-core/spec/logstash/java_pipeline_spec.rb:682:in `block in (root)'
./vendor/bundle/jruby/2.3.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in (root)'
./lib/bootstrap/rspec.rb:13:in `<main>'
-e:1:in `<main>'
```

original-brownbear:
Fixes #8883 

* Obvious problem here, the test runs multiple worker threads, concurrent pushes to a simple `RubyArray` will potentially drop events since it's not thread safe
  * Fixed by locking around the `RubyArray` instead of using some thread safe data structure to not alter behaviour more than I have to since we're using code in quite a few specs

### DIFF
--- a/logstash-core/spec/support/mocks_classes.rb
+++ b/logstash-core/spec/support/mocks_classes.rb
@@ -39,13 +39,17 @@ module LogStash
         super
         @num_closes = 0
         @events = []
+        @mutex = Mutex.new
       end
 
       def register
       end
 
       def receive(event)
+        @mutex.lock
         @events << event
+      ensure
+          @mutex.unlock
       end
 
       def close


original-brownbear:
@jsvd thanks!

PR #8889, opened 2017-12-28
 #8888 Fix RubyTimestamp clone method

### LINKED ISSUE #8888: cloning a LogStash::Timestamp results in an empty java object
```
2.3.0 :012 > LogStash::Timestamp.now.to_java
 => #<Java::OrgLogstash::Timestamp:0x48110259> 
2.3.0 :013 > LogStash::Timestamp.now.clone.to_java
 => nil 
```

My suspicion is that the [initialization of a LogStash::Timestamp](https://github.com/elastic/logstash/blob/master/logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java#L36-L39) requires the call to `setTimestamp` on a new instance. When doing a clone this set operation isn't called so the resulting cloned object doesn't contain the org.logstash.Timestamp instance

original-brownbear:
Fixes #8888 by properly implementing a clone method for the `RubyTimestamp` :)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
+++ b/logstash-core/src/main/java/org/logstash/ext/JrubyTimestampExtLibrary.java
@@ -117,6 +117,16 @@ public final class JrubyTimestampExtLibrary {
             return JavaUtil.convertJavaToUsableRubyObject(context.runtime, this.timestamp);
         }
 
+        @JRubyMethod(name = "clone")
+        public IRubyObject ruby_clone(ThreadContext context) {
+            return RubyTimestamp.newRubyTimestamp(context.runtime, this.timestamp);
+        }
+
+        @JRubyMethod(name = "dup")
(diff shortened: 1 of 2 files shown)


jsvd:
I don't mean to be annoying, but while you're at it:

```
jruby-9.1.14.0 :001 > LogStash::Timestamp.now.to_java
 => #<Java::OrgLogstash::Timestamp:0x4ec56dde> 
jruby-9.1.14.0 :002 > LogStash::Timestamp.now.clone.to_java
 => #<Java::OrgLogstash::Timestamp:0x61d3ead0> 
jruby-9.1.14.0 :003 > LogStash::Timestamp.now.dup.to_java
 => nil 
```

original-brownbear:
@jsvd :D np let me add that alias :)

original-brownbear:
@jsvd fixed :)

original-brownbear:
Note to self, 6.0 and 5.6 need manual backport because `RubyUtil` looks different there.

PR #8890, opened 2017-12-28
 #8889 backport to 6.0 and 5.6

original-brownbear:
Backport of #8889

### DIFF
--- a/logstash-core/spec/logstash/timestamp_spec.rb
+++ b/logstash-core/spec/logstash/timestamp_spec.rb
@@ -41,6 +41,10 @@ describe LogStash::Timestamp do
       expect(t == '-').to be_falsey
     end
 
+    it "should copy itself correctly on clone and dup" do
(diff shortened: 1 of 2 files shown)


original-brownbear:
@jsvd can't do that without pretty significant productive code changes :( I think this is trivial enough so that the risk of changing the production code (would have to port the changed way we bootstrap thhe Ruby classes in `6.1+` :() far outweighs the gain from the tests right?

jsvd:
I see..one suggestion is to move the test up a layer and test the clone/dup operations [on the ruby class](https://github.com/elastic/logstash/blob/6.0/logstash-core/spec/logstash/timestamp_spec.rb), wdyt?

original-brownbear:
@jsvd sure we can do that :) will add it here then

original-brownbear:
Ruby specs added :)

original-brownbear:
Tests failure is unrelated merging

original-brownbear:
`5.6` needs separate backport ...

Issue #8893, opened 2017-12-28
Change to use pipelines.yml by default?


jordansissel:
The logstash.yml[1] contains this line:

```
path.config: /etc/logstash/conf.d/*.conf
```

[1] This is from the Logstash 6.1.1 rpm.

When folks go exploring with the new multiple pipelines feature, it is not obvious how to enable multiple pipelines.

I wonder if we can make this journey easier by defaulting to multiple pipelines out of the box. We can take steps to keep compatibility by having the pipeline called 'main' and using '/etc/logstash/conf.d/*.conf' as the source.

jordansissel:
Done: 
* deb and rpm: https://github.com/elastic/logstash/pull/8894
* docker: https://github.com/elastic/logstash-docker/pull/73

PR #8894, opened 2017-12-28
Use multiple pipelines feature by default.

jordansissel:
Use multiple pipelines feature by default in RPM and Debian packages.

This change keeps the previous behavior but implements it using multiple
pipelines (pipelines.yml) instead of with `path.config`.

* The default pipeline name is still "main"
* Default config path is still /etc/logstash/conf.d/*.conf

Other small changes included:
* Specify fpm workdir. Otherwise rpmbuild is given an empty value for
  its temporary directory. It's unclear if this is a bug in fpm or in my
  environment, but I like the change anyway to put fpm's build activity
  in the build directory.
* DRY's up the common files going in /etc/logstash for both rpm and deb
  packages.

For #8893

### DIFF
--- a/docs/static/setting-up-logstash.asciidoc
+++ b/docs/static/setting-up-logstash.asciidoc
@@ -95,8 +95,8 @@ locations for the system:
 
 | conf
   | Logstash pipeline configuration files
-  | `/etc/logstash/conf.d`
-  | `path.config`
+  | `/etc/logstash/conf.d/*.conf`
+  | See `/etc/logstash/pipelines.yml`
 
 | logs
   | Log files
--- a/pkg/centos/after-install.sh
+++ b/pkg/centos/after-install.sh
@@ -2,7 +2,6 @@ chown -R logstash:logstash /usr/share/logstash
 chown -R logstash /var/log/logstash
 chown logstash:logstash /var/lib/logstash
 sed -i \
-  -e 's|# path.config:|path.config: /etc/logstash/conf.d/*.conf|' \
   -e 's|# path.logs:|path.logs: /var/log/logstash|' \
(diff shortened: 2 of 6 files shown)


jordansissel:
To test manually: `rake artifact:rpm`

/etc/logstash/logstash.yml should *not* have `path.config` in use and should default to /etc/logstash/pipelines.yml.

You can verify by putting a config in /etc/logstash/conf.d/whatever.conf and starting logstash `/usr/share/logstash/bin/logstash "--path.settings" "/etc/logstash"` which should use pipelines.yml (no warning about pipelines.yml being ignored!).

jordansissel:
@jsvd would you mind reviewing this? I think you worked most on the multiple pipelines and may have the most context to judge whether this change is correct.

jsvd:
I think this change should have a similar one on the docker side for [oss](https://github.com/elastic/logstash-docker/blob/032f09891d125e2f11f4cfb2731b69c30fad402e/build/logstash/config/logstash-oss.yml#L2) and [x-pack](https://github.com/elastic/logstash-docker/blob/032f09891d125e2f11f4cfb2731b69c30fad402e/build/logstash/config/logstash-x-pack.yml#L2) as we document [here](https://github.com/elastic/logstash/blob/master/docs/static/docker.asciidoc#docker-defaults) that the `path.config` is set explicitly too.

jordansissel:
@jsvd agreed, I'll file a similar PR there.

jordansissel:
https://github.com/elastic/logstash-docker/pull/73 filed. I'll update this PR with docs changes for Docker in preparation.

jordansissel:
https://github.com/elastic/logstash-docker/pull/73 merged. Merging this also.

PR #8903, opened 2017-12-29
Remove check for future pipeline reload-abilty.

### LINKED ISSUE #8902: Reloading pipelines call plugin initialize, which may cause resource leaks
Tested against master (7.0.0-alpha1 pre-release). Related: https://github.com/logstash-plugins/logstash-filter-grok/issues/132

Reloading a pipeline causes the initialize method of the plugin to be called be called twice. One time to validate the configuration and another to acutally start the pipeline.  This can result in un-expected behavior if the plugin initialize method needs to be paried with a cleanup type method. 

Specifically this causes a thread leak in the grok plugin since that plugin creates a new thread in the initialize method that is expected to be joined with the close method. Since the initialize is getting called twice, where the first time, the close method is never called, that thread is created and never destroyed. 

Below is a log with the grok plugin and a trace of who is calling iniatilize  (commit:  590e47fcedac1f6e687f8a568067b813b62038c0)
```
  def initialize(params)
      puts caller
 ```
      
Below illustrates that when reload.rb is called it results in calling the initialize method twice for a single reload.  

```
Sending Logstash's logs to /Users/jake/workspace/logstash/logs which is now configured via log4j2.properties
[2017-12-29T15:57:17,379][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"fb_apache", :directory=>"/Users/jake/workspace/logstash/modules/fb_apache/configuration"}
[2017-12-29T15:57:17,386][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"netflow", :directory=>"/Users/jake/workspace/logstash/modules/netflow/configuration"}
[2017-12-29T15:57:17,417][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified
[2017-12-29T15:57:17,511][INFO ][logstash.runner          ] Starting Logstash {"logstash.version"=>"7.0.0-alpha1"}
[2017-12-29T15:57:17,576][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
/Users/jake/workspace/logstash/logstash-core/lib/logstash/filter_delegator.rb:23:in `initialize'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/plugins/plugin_factory.rb:88:in `plugin'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:112:in `plugin'
(eval):12:in `<eval>'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:84:in `eval'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:84:in `initialize'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:169:in `initialize'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline_action/create.rb:40:in `execute'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:315:in `block in converge_state'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:141:in `with_pipelines'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:312:in `block in converge_state'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:299:in `each'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:299:in `converge_state'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:166:in `block in converge_state_and_update'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:141:in `with_pipelines'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:164:in `converge_state_and_update'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:90:in `execute'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/runner.rb:343:in `block in execute'
/Users/jake/workspace/logstash/vendor/bundle/jruby/2.3.0/gems/stud-0.0.23/lib/stud/task.rb:24:in `block in initialize'
[2017-12-29T15:57:17,943][INFO ][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"main", "pipeline.workers"=>1, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>50}
[2017-12-29T15:57:18,041][INFO ][logstash.pipeline        ] Pipeline started succesfully {:pipeline_id=>"main", :thread=>"#<Thread:0x67f529f@/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:246 sleep>"}
[2017-12-29T15:57:18,053][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
/Users/jake/workspace/logstash/logstash-core/lib/logstash/filter_delegator.rb:23:in `initialize'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/plugins/plugin_factory.rb:88:in `plugin'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:112:in `plugin'
(eval):12:in `<eval>'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:84:in `eval'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:84:in `initialize'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline_action/reload.rb:34:in `execute'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:315:in `block in converge_state'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:141:in `with_pipelines'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:312:in `block in converge_state'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:299:in `each'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:299:in `converge_state'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:166:in `block in converge_state_and_update'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:141:in `with_pipelines'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:164:in `converge_state_and_update'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:105:in `block in execute'
/Users/jake/workspace/logstash/vendor/bundle/jruby/2.3.0/gems/stud-0.0.23/lib/stud/interval.rb:18:in `interval'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:94:in `execute'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/runner.rb:343:in `block in execute'
/Users/jake/workspace/logstash/vendor/bundle/jruby/2.3.0/gems/stud-0.0.23/lib/stud/task.rb:24:in `block in initialize'
[2017-12-29T15:57:27,228][INFO ][logstash.pipelineaction.reload] Reloading pipeline {"pipeline.id"=>:main}
[2017-12-29T15:57:27,548][INFO ][logstash.pipeline        ] Pipeline has terminated {:pipeline_id=>"main", :thread=>"#<Thread:0x67f529f@/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:246 run>"}
/Users/jake/workspace/logstash/logstash-core/lib/logstash/filter_delegator.rb:23:in `initialize'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/plugins/plugin_factory.rb:88:in `plugin'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:112:in `plugin'
(eval):12:in `<eval>'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:84:in `eval'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:84:in `initialize'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:169:in `initialize'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline_action/create.rb:40:in `execute'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline_action/reload.rb:48:in `execute'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:315:in `block in converge_state'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:141:in `with_pipelines'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:312:in `block in converge_state'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:299:in `each'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:299:in `converge_state'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:166:in `block in converge_state_and_update'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:141:in `with_pipelines'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:164:in `converge_state_and_update'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:105:in `block in execute'
/Users/jake/workspace/logstash/vendor/bundle/jruby/2.3.0/gems/stud-0.0.23/lib/stud/interval.rb:18:in `interval'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/agent.rb:94:in `execute'
/Users/jake/workspace/logstash/logstash-core/lib/logstash/runner.rb:343:in `block in execute'
/Users/jake/workspace/logstash/vendor/bundle/jruby/2.3.0/gems/stud-0.0.23/lib/stud/task.rb:24:in `block in initialize'
[2017-12-29T15:57:27,641][INFO ][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"main", "pipeline.workers"=>1, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>50}
[2017-12-29T15:57:27,675][INFO ][logstash.pipeline        ] Pipeline started succesfully {:pipeline_id=>"main", :thread=>"#<Thread:0x502f150b@/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:246 sleep>"}
[2017-12-29T15:57:27,680][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
[2017-12-29T15:57:33,125][WARN ][logstash.runner          ] SIGINT received. Shutting down.
stopping!
[2017-12-29T15:57:33,458][INFO ][logstash.pipeline        ] Pipeline has terminated {:pipeline_id=>"main", :thread=>"#<Thread:0x502f150b@/Users/jake/workspace/logstash/logstash-core/lib/logstash/pipeline.rb:246 run>"}

Process finished with exit code 0
```

### PR DESCRIPTION
**Pull request by jakelandis**
This change stops checking to see if the next pipeline may be reloaded. That check calls the plugin constructor without calling the effective desconstructor which can result in resource leaks if resources are setup in the initialize method. Also see: https://github.com/logstash-plugins/logstash-filter-grok/issues/132

fixes #8902

### DIFF
--- a/logstash-core/lib/logstash/pipeline_action/reload.rb
+++ b/logstash-core/lib/logstash/pipeline_action/reload.rb
@@ -26,26 +26,15 @@ module LogStash module PipelineAction
         return LogStash::ConvergeResult::FailedAction.new("Cannot reload pipeline, because the existing pipeline is not reloadable")
       end
 
-      begin
-        pipeline_validator =
-          if @pipeline_config.settings.get_value("pipeline.java_execution")
-            LogStash::JavaBasePipeline.new(@pipeline_config)
-          else
-            LogStash::BasePipeline.new(@pipeline_config)
-          end
-      rescue => e
-        return LogStash::ConvergeResult::FailedAction.from_exception(e)
-      end
-
-      if !pipeline_validator.reloadable?
-        return LogStash::ConvergeResult::FailedAction.new("Cannot reload pipeline, because the new pipeline is not reloadable")
-      end
-
       logger.info("Reloading pipeline", "pipeline.id" => pipeline_id)
       status = Stop.new(pipeline_id).execute(agent, pipelines)
 
       if status
-        return Create.new(@pipeline_config, @metric).execute(agent, pipelines)
+        begin
+          return Create.new(@pipeline_config, @metric).execute(agent, pipelines)
+        rescue => e
+          return LogStash::ConvergeResult::FailedAction.from_exception(e)
+      end
       else
         return status
       end

### CODE REVIEW

**Comment by jakelandis** [comment]
@jordansissel - can you review? I will fix/update the tests once I get confirmation that this is the right direction.

**Comment by jordansissel** [comment]
3 tests failed - https://logstash-ci.elastic.co/job/elastic+logstash+pull-request/2103/

```
22:40:23     rspec ./logstash-core/spec/logstash/agent_spec.rb:283 # LogStash::Agent#upgrade_pipeline when the upgrade fails leaves the state untouched
22:40:23     rspec ./logstash-core/spec/logstash/agent_spec.rb:134 # LogStash::Agent#execute when auto_reload is false when calling reloading a pipeline with a config that contains reload incompatible plugins does not upgrade the new config
22:40:23     rspec ./logstash-core/spec/logstash/pipeline_action/reload_spec.rb:64 # LogStash::PipelineAction::Reload when the new pipeline is not reloadable cannot successfully execute the action
```

jordansissel:
@jakelandis will review the direction and ignore tests.

jordansissel:
Stepping back and looking at the problem space "reloadable plugins" I think there is only one plugin which declares itself "I can't be reloaded" -- https://github.com/search?q=org%3Alogstash-plugins+reloadable&type=Code -- the stdin input.

If we revisit this concept, maybe we don't need it at all? Maybe the right solution is to focus on the stdin plugin specifically instead of a more abstract "reloadable" concept?

Maybe the right solution is to make reloadable? false only specifically with `stdin` and that we should reject reloads : 

* if the current pipeline uses stdin
* or, if the new pipeline uses stdin

----

That said, the behavioral differences between your approach:

* master: new pipeline cannot be reloaded, so new pipeline is rejected
* this pr: current pipeline cannot be reloaded, so reload is rejected.

Testing:

| step | inputs | master | this pr |
|-|-|-|-|
| Start with `input { tcp { ... } }` | tcp | ✔️ | ✔️ |
| Add `stdin { }` input and attempt a reload | tcp, stdin | ❌ "the new pipeline is not reloadable" | ✔️ |
| Remove `tcp { ... }` and attempt a reload. | stdin | ❌ "the new pipeline is not reloadable"  | ❌ "the current pipeline is not reloadable" |

jordansissel:
Give my above comments, I am leaning towards making the path forward for "reloadable?" be focused specifically on `stdin` rather than an abstract capability "reloadable"

Perhaps two steps:

1) See if we can make `stdin` reloadable.
2) If stdin cannot be made reloadable, we can focus reject-reloads on stdin specifically.

jakelandis:
There is an outstanding issue at https://github.com/logstash-plugins/logstash-input-stdin/issues/15 to support reloading.  I agree with fixing that... however,

I _think_ it safe to generally ignore "new pipeline's"'s reload flag since anything new is by definition not `re`-loaded but only loaded. I think this PR fixes a logic bug (checking a new pipeline reload flag), and also fixes the potential resource leaking that is a side effect of that check. 

I may be missing something here .... but I _think_ you column `this pr` is more correct then `master`.

jordansissel:
> I think it safe to generally ignore "new pipeline's"'s reload flag since anything new is by definition not re-loaded but only loaded

My supposition here is that if someone is going to reload once, they'll do it again. As my table of results above showed, you can go from a working pipeline, reload a new pipeline, and the 2nd reload attempt fails.

This causes a problem because we allow Logstash to be put into a irrecoverable state through reloading, and recovering from this state requires operator intervention to manually restart Logstash in order to resolve this, and with central config, it is possible that the person configuring Logstash has no direct access to restart Logstash (only to reload via central config).

There's a few cases to consider:

| case | master | this pr |
|-|-|-|
| stdin is present at startup (not a reload) | ✔️ | ✔️ |
| stdin is added by a reload | ❌ | ✔️ |
| stdin is active and thus prevents a reload. | ❌ | ❌ |

I'd like us to trend towards allowing users to correct for problems themselves and to have Logstash try and avoid problems. Allowing stdin to be added by a reload introduces a new failure point where a correction requires an operator restart Logstash *without* stdin in order to correct. Today, adding stdin by reload will be rejected to avoid this problem.

To put it another way, the first time you add "stdin" input is the last time you reload. This seems like a bad end-state for users to get into.

jordansissel:
If we do accept this PR, then the log message needs to be much more explicit especially where users may be using multiple pipelines and/or central config -- the failure message needs to indicate that the reload is rejected because stdin input is in use, and to resolve, the stdin input must be removed *AND* Logstash must be restarted after the removal.

jakelandis:
@jordansissel - thanks for the feedback, I now understand the concern (and original reason for this code).  I'll spend abit more time with this to see if there is a better direction we can take this.

jordansissel:
@jakelandis what about having the exception tell the user two things:

1) Which plugin(s) are preventing reload
2) How to correct for this:
    * If new pipeline, say removing the nonreloadable plugins will permit the reload.
    * If current pipeline, say removing the nonreloadable plugins *and* restarting will allow reload to work.

jakelandis:
> If new pipeline, say removing the nonreloadable plugins will permit the reload.

I am not sure how to do this with loading the plugin first. Loading the plugin may create resources that need to be cleaned up if resources are created in the `initialize` method (see https://github.com/logstash-plugins/logstash-filter-grok/issues/132). 

+1 to your better messaging, but not sure if solves the problem i am trying to solve of cleaning up things initialized during the next pipeline reload check.

A couple options :

* Just move initialization logic to the `register` method on per plugin basis. e.g. maybe this is just a bug in the grok filter (and possibly others) that they are incorrectly initializing resources that need to be un-initialized. 
* Have this check for the next pipeline also call `close` to help ensure it cleans up anything initialized.  (this is harder since the current code only calls BasePipeline explicitly, not the child plugin)

jordansissel:
> Loading the plugin may create resources that need to be cleaned up 

Why don't we clean them up? Seems like we should clean up. Moving things in grok to the `register` method is probably a fine solution for this case.

jakelandis:
> Why don't we clean them up?

This is the bug.  Arguably it is caused by this "check next pipeline reloadability" code.  

My real question was "Why do we need to check for the next pipeline reload-ablity?"  Which I think I now understand. 

The simplest fix (this PR) was to simply remove the check, but am leaning away from the that now. 
 
> Moving things in grok to the register method is probably a fine solution for this case.

That would fix grok, but there is a still a trap here such that there is (imo) an expectation that `close` will always be called. I will look into how difficult it will be to iterate through the pipeline created to test reloadabilty and explicitly call close on those plugins.

jakelandis:
Closing this PR since we want to head a different direction then presented here.

PR #8896, opened 2017-12-29
MINOR: Prevent enviornment GEM_DIR and GEM_PATH from leaking into build

original-brownbear:
This fixes environment `GEM_PATH` and `GEM_DIR` leaking into the build during integration tests if you have `rvm` or `rbenv` installed:

* both load their environment via login shell hooks
* ITs + rake tasks call `system(...)` in some spots => leaks in environment (mostly not an issue since our `.rb` scripts fix these paths manually, but at least for the plugin manager I see it using the gems from my local `.rvm` dir which at times caused build issues that were hard to fix without removing rvm from my `.profile`)
* Simply manually setting the correct path when we bootstrap the rest of the JRuby env anyways like done here fixes all of these issues just fine for me :)

PR #8907, opened 2018-01-03
MINOR: Fix Ruby path

original-brownbear:
Follow up for #8896. Unfortunately we only  fixed the `Ruby` entry point case there. The Gradle case needs the same fix using `bin/ruby` instead of the `vendor/jruby` path.

@jsvd can you take a look please since you understand this one already :)?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -137,8 +137,12 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
     into "${projectDir}/vendor/jruby"
 }
 
-def jrubyBin = "${projectDir}/vendor/jruby/bin/jruby" +
-  (System.getProperty("os.name").startsWith("Windows") ? '.bat' : '')
+def isWindows = System.getProperty("os.name").startsWith("Windows")
+// The default environment loaded by bin/ruby segfaults on CI for unknown reasons. For now we work
+// around this fact by directly invoking the jruby binary on CI.
+def jrubyBin = System.getenv('CI') || isWindows ?
+  ("${projectDir}/vendor/jruby/bin/jruby" + (isWindows ? '.bat' : ''))
+  : "${projectDir}/bin/ruby"
 
 def rakeBin = "${projectDir}/vendor/jruby/bin/rake"
 


original-brownbear:
@jsvd thanks!

original-brownbear:
Jenkins test this please

original-brownbear:
Jenkins test this please

original-brownbear:
Jenkins test this please

original-brownbear:
hmm this causes a Segfault on Jenkins :( This is obviously not the fault of this change in and of itself (works fine in and of itself locally + even if it didn't SEGFAULT can't be our fault).
=> will try to find out why and what I can do about it

original-brownbear:
@jsvd hmm this is not cool, seems we're doing something in the `bin/ruby` startup script that breaks `bundler` (tested a few other JVMs on the Docker container) in the Jenkins env.

I'm giving up here ... will just move the bundling logic to Gradle eventually. It's not worth it tracking this down imo now that the main entry point is in Java anyways.

colinsurprenant:
I took a look at it and just pushed #8914 - we'll see if it passes on Jenkins. 

So I realized that my minitar problem was in the qa/integration gem installation steps, when building with `gradle build`, but all the non-qa related tasks were actually building just fine. 

So my PR fixes the gem paths problem of the qa/integration tasks. 
It also keeps the `vendor/jruby` because I think it's better to avoid going through the `logstash.lib.sh` and inherit all the JVM options we set for production runtime which are probably overkill of the build and are probably the cause of the segfaults?

original-brownbear:
@colinsurprenant commented on that PR not so sure about it. Either way we should find out where these segfaults are coming from. It's not cool that the Java env we set up for production configurations has this issue (likely with bundler ... then again the plugin installer works from the same environment ... so idk (yet :) )).

original-brownbear:
@jsvd @colinsurprenant Adjusted this PR to not use our env setup when running tests on CI or Windows(we don't have `bin/ruby.bat` unfortunately), otherwise use `bin/ruby` on `*nix` setups.

I think this is the correct/portable approach since it 100% guarantees that the environment gems never leak in, even if `rbenv` or `rvm` do their magic (because we fix the gem paths inside the `bash` script and then hand over to JRuby via `exec` without triggering another sourcing of `.profile` or similar).

colinsurprenant:
testing!

colinsurprenant:
the integration tests task don't work locally for me, from a complete clean env. digging.

original-brownbear:
redundant due to https://github.com/elastic/logstash/pull/8941

PR #8908, opened 2018-01-04
Remove pipeline.output.workers from  logstash.yml

dedemorton:
Closes https://github.com/elastic/logstash/issues/7884.

Note that https://github.com/elastic/logstash/pull/8562 removes the setting from the docs.


jordansissel:
LGTM

PR #8914, opened 2018-01-05
[WIP] fix gradle jruby build env

colinsurprenant:
This fixes the jruby build env and replaces #8907 and is a followup of #8896 

I think keeping `vendor/jruby` instead of `bin/ruby` is probably better to avoid inheriting all the runtime JVM options set for running LS but that are probably overkill for the build. 

`gradle build` now works locally from clean env, including  running all tests and integration tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -233,7 +233,7 @@ task runIntegrationTests(dependsOn: installIntegrationTestGems, type: Exec) {
   workingDir "${projectDir}/qa/integration"
   environment "LS_GEM_PATH", qaBundledGemPath
   environment "LS_GEM_HOME", qaBundledGemPath
-  //environment "SKIP_SETUP_JAVA_OPTS", "1"
+  environment "SKIP_SETUP_JAVA_OPTS", "1"
   // FEATURE_FLAG is set in the CI to configure testing with enabled PQ
   environment "FEATURE_FLAG", System.getenv('FEATURE_FLAG')
   standardOutput = new ExecLogOutputStream(System.out)
--- a/build.gradle
+++ b/build.gradle
@@ -236,7 +236,7 @@ task runIntegrationTests(dependsOn: installIntegrationTestGems, type: Exec) {
   workingDir "${projectDir}/qa/integration"
   environment "LS_GEM_PATH", qaBundledGemPath
   environment "LS_GEM_HOME", qaBundledGemPath
-  environment "SKIP_SETUP_JAVA_OPTS", "1"
+//  environment "SKIP_SETUP_JAVA_OPTS", "1"
   // FEATURE_FLAG is set in the CI to configure testing with enabled PQ
   environment "FEATURE_FLAG", System.getenv('FEATURE_FLAG')
   standardOutput = new ExecLogOutputStream(System.out)
--- a/build.gradle
+++ b/build.gradle
@@ -211,16 +211,16 @@ task installIntegrationTestBundler(dependsOn: unpackTarDistribution, type: Exec)
 
 task installIntegrationTestGems(dependsOn: installIntegrationTestBundler, type: Exec) {
   workingDir "${projectDir}/qa/integration"
-  environment "LS_GEM_PATH", qaGemPath
-  environment "LS_GEM_HOME", qaGemPath
-  environment "SKIP_SETUP_JAVA_OPTS", "1"
   inputs.files file("${projectDir}/qa/integration/Gemfile")
+  inputs.files file("${projectDir}/qa/integration/integration_tests.gemspec")
   inputs.files file("${logstashBuildDir}/Gemfile")
   inputs.files file("${logstashBuildDir}/Gemfile.lock")
   inputs.files file("${logstashBuildDir}/logstash-core/logstash-core.gemspec")
-  inputs.files file("${projectDir}/qa/integration/integration_tests.gemspec")
   outputs.files fileTree("${qaGemPath}")
   outputs.files file("${projectDir}/qa/integration/Gemfile.lock")
+  environment "LS_GEM_PATH", qaGemPath
+  environment "LS_GEM_HOME", qaGemPath
+  environment "SKIP_SETUP_JAVA_OPTS", "1"
   standardOutput = new ExecLogOutputStream(System.out)
   errorOutput =  new ExecLogOutputStream(System.err)
   commandLine jrubyBin, "-S", qaBundleBin, "install"
--- a/build.gradle
+++ b/build.gradle
@@ -99,6 +99,9 @@ clean {
   delete "${projectDir}/Gemfile.lock"
   delete "${projectDir}/vendor"
   delete "${projectDir}/NOTICE.TXT"
+  delete "${projectDir}/.bundle"
+  delete "${projectDir}/qa/integration/Gemfile.lock"
+  delete "${projectDir}/qa/integration/.bundle"
 }
 
 task bootstrap {}
@@ -230,7 +233,7 @@ task runIntegrationTests(dependsOn: installIntegrationTestGems, type: Exec) {
   workingDir "${projectDir}/qa/integration"
   environment "LS_GEM_PATH", qaBundledGemPath
   environment "LS_GEM_HOME", qaBundledGemPath
-  environment "SKIP_SETUP_JAVA_OPTS", "1"
+  //environment "SKIP_SETUP_JAVA_OPTS", "1"
   // FEATURE_FLAG is set in the CI to configure testing with enabled PQ
   environment "FEATURE_FLAG", System.getenv('FEATURE_FLAG')
   standardOutput = new ExecLogOutputStream(System.out)
--- a/build.gradle
+++ b/build.gradle
@@ -201,8 +201,8 @@ def qaBundleBin = "${buildDir}/qa/integration/vendor/bin/bundle"
 
 task installIntegrationTestBundler(dependsOn: unpackTarDistribution, type: Exec) {
   outputs.files fileTree("${qaGemPath}/gems/bundler-1.16.0")
-  environment "GEM_PATH", qaGemPath
-  environment "GEM_HOME", qaGemPath
+  environment "LS_GEM_PATH", qaGemPath
+  environment "LS_GEM_HOME", qaGemPath
   environment "SKIP_SETUP_JAVA_OPTS", "1"
   standardOutput = new ExecLogOutputStream(System.out)
   errorOutput =  new ExecLogOutputStream(System.err)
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -140,10 +140,14 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
 def jrubyBin = "${projectDir}/vendor/jruby/bin/jruby" +
   (System.getProperty("os.name").startsWith("Windows") ? '.bat' : '')
 
+def gemPath = "${projectDir}/vendor/bundle/jruby/2.3.0"
```
original-brownbear: @colinsurprenant this will leak IT build gems into the vendored dir, I don't think that's correct without adjusting the artifact task accordingly?

build.gradle
```diff
@@ -140,10 +140,14 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
...
+
 def rakeBin = "${projectDir}/vendor/jruby/bin/rake"
 
 task installTestGems(dependsOn: downloadAndInstallJRuby, type: Exec) {
   workingDir projectDir
+  environment "GEM_PATH", gemPath
```
original-brownbear: @colinsurprenant this only helps you locally if you don't have `rvm` in your `bash` environment (sadly enough :(). That why I put the gem path adjustment into the `bash` script in https://github.com/elastic/logstash/pull/8896/files. The environment you set here is not making it's way into the actual execution because it's overwritten as soon as `vendor/jruby/bin/ruby` is loaded

original-brownbear:
@colinsurprenant I'm not so sure this doesn't add more problems than it solves and doesn't fix the rvm/rbenv issue many people are experiencing. See my two comments.

build.gradle
colinsurprenant: Not sure I am following?
But essentially this is what we have in the Ruby bootstrap code in `logstash.lib.sh` https://github.com/elastic/logstash/blob/76c24cf697ec576c0efd90e4ade008226cff7338/bin/logstash.lib.sh#L119

build.gradle
colinsurprenant: the environment should work regardless of if you use any Ruby version manager or not ...  
I am using rbenv, and this seems to work fine. Will review.

colinsurprenant:
@original-brownbear it does solve it for me, I am using rbenv. will review your comments and also make a test without any Ruby version manager.

original-brownbear:
@colinsurprenant whether you get burnt by this is a matter of your shell. The problem is, that on some/many shells (`zsh` is guilty here, `bash` too on some systems), `.zshrc/.bashrc` is read even for non-login shell.
=> The env you set up in Gradle will be overwritten as soon as `rum` or `rbenv` is loaded by sourcing `.*rc` :(
This won't happen to you when you fix the GEM_PATH in the shell script itself and then subsequently call `exec` because you're not jumping to a new sub-process whereas Gradle will simply spin off a new sub process for `Exec` task that may or may not result in the `*.rc` being sourced and messing things up => portability is the worst :P

colinsurprenant:
Ok I see that the `GEM_HOME` env was recently added in `logstash.lib.sh` https://github.com/elastic/logstash/blob/76c24cf697ec576c0efd90e4ade008226cff7338/bin/logstash.lib.sh#L119 via https://github.com/elastic/logstash/pull/8896  ... 
The gem home path was originally constructed in https://github.com/elastic/logstash/blob/76c24cf697ec576c0efd90e4ade008226cff7338/lib/bootstrap/environment.rb#L43-L45 and set in `bootstrap/bundler.rb` but when launching from Ruby.

colinsurprenant:
@original-brownbear  Ok, thanks for the heads up about Gradle/task/exec.

> This won't happen to you when you fix the GEM_PATH in the shell script itself and then subsequently call exec because you're not jumping to a new sub-process whereas Gradle will simply spin off a new sub process for Exec task that may or may not result in the *.rc being sourced and messing things up => portability is the worst :P

So this is why you wanted to launch the JRuby commands through `bin/ruby` so that GEM_HOME is correctly reset **after** the Gradle task `commandLine` in #8907, right?

original-brownbear:
@colinsurprenant yea exactly! The "after" does the trick here. Kind of a dirty hack but ...
Obviously the correct fix that we should be working towards is to simply run Rake from the main Gradle JVM and remove all calls to `system()` (or other direct shell command invocations) from the `rake` tasks ... but that's much easier said than done I fear.

colinsurprenant:
@original-brownbear ok, got it. So then, basically the `bin/ruby` indirection did work but introduced the JVM segfault which is certainly related to the *production* JVM options we set in `logstash.lib.sh`. Let me see what we can do here. I'll add comments to document the ENV strategy here so that anyone else looking into that can quickly get the right picture. 

> Obviously the correct fix that we should be working towards is to simply run Rake from the main Gradle JVM and remove all calls to system()

+1  we can look into that subsequently.

colinsurprenant:
@original-brownbear damn - if we use `bin/ruby` to launch, then we face the problem in the integration tests setup which wants to inject a different GEM_HOME per https://github.com/elastic/logstash/blob/76c24cf697ec576c0efd90e4ade008226cff7338/build.gradle#L193-L194

One way to solve this would be to find a way to launch a Gradle Exec task with an empty environment. Not sure if possible.

Another idea would be to add a few more env var control knobs on around `logstash.lib.sh` to be able to inject say a `LS_GEM_HOME` and also be able to `SKIP_SETUP_JAVA_OPTS` to avoid setting the production JAVA_OPTS for build. Lemme push a quick proposal and you tell me what you think.

original-brownbear:
@colinsurprenant haha right urgh :( Yea ... I like `LS_GEM_HOME` best I think ... thanks! :)

colinsurprenant:
- Broke down the `logstash.lib.sh` `setup()` function into 3 parts:  `setup_java()`, `setup_java_opts()`, `setup_vendored_jruby()` where both `setup_java()`, `setup_java_opts()` are skippable using `SKIP_SETUP_JAVA=1` and `SKIP_SETUP_JAVA_OPTS=1`.
- Added support for `LS_GEM_HOME` and `LS_GEM_PATH` to inject them into `logstash.lib.sh`.
- Added debug traces

Tested locally on OSX + bash + rbenv. Builds, tests and integration tests are passing.

colinsurprenant:
I think this will pass. Now, as it is, only bundler is installed in `build/qa/integration/vendor`. Not sure why originally the integration tests gem path was set into `build/qa/integration/...`. I will set it back into logstash home `qa/integration/vendor`. Before pushing that fix I'll wait to see if this is green.

colinsurprenant:
Ok build is red but because of actual & unrelated failed tests - so this is perfect in our context. All launch paths, gem paths, etc seems to be working correctly now. 

One question I have is about the integrations tests path. Before this PR both `"${projectDir}/qa/integration"` and `"${buildDir}/qa/integration"` were used I believe inconsistently.  

What was the original intent? to run integration tests from the untarred distribution in `def logstashBuildDir = "${buildDir}/logstash-${project.version}-SNAPSHOT"` ?

It seems the integration tests are run from `logstashBuildDir` so it would make sense to have all integration tests gems vendored into the build dir. Will change that.

colinsurprenant:
this is the weirdest thing. I don't understand why the integration tests in the `installIntegrationTestGems` task install into `qa/integration/vendor` and not into `build/qa/integration/vendor` just like the bundler installation in the `installIntegrationTestBundler` task. As if `GEM_HOME` and `GEM_PATH` are not taken into account and it uses the default `vendor` path into the working dir in `qa/integration`

colinsurprenant:
Green. The integration tests strategy is now to install bundler and all integrations tests gems into the build dir `build/qa/integration/vendor`. 

Also as noted above, the following as been added:

- Broke down the `logstash.lib.sh` `setup()` function into 3 parts:  `setup_java()`, `setup_java_opts()`, `setup_vendored_jruby()` where both `setup_java()`, `setup_java_opts()` are skippable using `SKIP_SETUP_JAVA=1` and `SKIP_SETUP_JAVA_OPTS=1`.
- Added support for `LS_GEM_HOME` and `LS_GEM_PATH` to inject them into `logstash.lib.sh`.
- Added debug traces

jordansissel:
This PR adds a bunch of environment variables to bin/logstash that seem to be intended only for development, not deployment. Is this right? If so, can you add comments above the usage of these variables to indicate they are to make development smoother and are not supported nor useful for production.

colinsurprenant:
@jordansissel yes sure. FYI this is a WIP so also intended for discussion, especially on the strategy. 

- If we end up keeping/agreeing on the new env vars, yes, I will document. 
- Adding these new env vars in `logstash.lib.sh` is not super clean since they are in fact only needed for development. We have a few problems that these solves now:
  - We cannot rely on the Gradle exec task environment for the `GEM_HOME` and `GEM_PATH` since these can be overwritten by shells init scripts prior to launching the given exec command and this is why we need to indirect the exec commands through `bin/ruby` which relies on `logstash.lib.sh` to correctly set `GEM_HOME` and `GEM_PATH` just before launching the command. 
  - For the normal LS build and tests `bin/ruby` + `logstash.lib.sh` does the right thing in setting up `GEM_HOME` and `GEM_PATH` but for the integration tests, we need to use a different `GEM_HOME` and `GEM_PATH` to point inside `build/qa/integration/vendor`.  This is why I introduced the `LS_GEM_HOME` and `LS_GEM_PATH` env var to override the defaults in `logstash.lib.sh`.
  - Another problem is with the usage or the "production" JVM options as set by `logstash.lib.sh` using `config/jvm.options`. For a yet unknown reason these production options segfaults the JVM (likely in bundler) in the CI build. I added the `SKIP_SETUP_JAVA_OPTS ` env var so that the portion of `logstash.lib.sh` setting up the production JVM option could be skipped. It does solve the problem. Obviously we want to understand the cause of the segfault and this is a temporary fix to make it work. 

Open to ideas to make this a better place. For now this works for me locally (I am using rbenv): it builds, tests, and runs integration tests and it also works on CI.

jordansissel:
```
⓿ pork(~/projects/logstash) pull/8914
% git rev-parse HEAD
dccc336914a97fc97842e6169d6ae6a2d3bf24ab

% ./gradlew build
...
> Task :installIntegrationTestBundler FAILED
rbenv: jruby: command not found
...
```

This PR failing for me.

colinsurprenant:
I rebased and refreshed this PR since *proper* Rake invocation from Gradle has been added.

What remains are the fixes for the integration tests bundling and invocation. Unless we have other plans, the `runIntegrationTests` task still needs to shell out to invoke `bundle exec` while inheriting the correct `GEM_HOME` and `GEM_PATH` and for this I believe the added `LS_GEM_HOME` and `LS_GEM_PATH` support is required to make this bullet proof for everyone, in local regardless of installed Ruby version manager and in CI. 
Also note the fixed related to the integration tests gems vendorization.

LMKWYT @jordansissel @original-brownbear

colinsurprenant:
rebooting in #8984

PR #8915, opened 2018-01-07
Ruby Entrypoint

original-brownbear:
Ruby entry point still works like this.

Not sure if we should even merge this or something like this though, can't we just run these two scripts as shown here from the IDE if we need debugging? (and document how to do this instead of merging this explicit script?)

### DIFF
--- a/bin/ruby-logstash.sh
+++ b/bin/ruby-logstash.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+unset CDPATH
+
+. "$(cd `dirname $0`/..; pwd)/bin/logstash.lib.sh"
+setup
+
+ruby_exec "${LOGSTASH_HOME}/lib/bootstrap/environment.rb" "logstash/runner.rb" "$@"


original-brownbear:
@jakelandis so long as we don't move the CLI arg parsing to Java this approach will not require any maintenance I think since it uses the same hacky solution for handling the Java library load that is used to fix the `bin/rspec` script :)

jakelandis:
@original-brownbear 

> Not sure if we should even merge this or something like this though, can't we just run these two scripts as shown here from the IDE if we need debugging? 

I don't have preference of the exact mechanics of how we launch a debugging session. Ideally we should be able to debug Ruby and Java at the same time, which worked prior to the Java entry point.  

This change is still missing adding the Java classpath via .classes for debugging Java : https://github.com/elastic/logstash/commit/543b7221bd47afbde1f3c71059ff4be344990e80#diff-3a9ab450779ec1f8803f713543b28b8e   

> so long as we don't move the CLI arg parsing to Java this approach will not require any maintenance

I would think that ideally the Ruby debug entry point would be a very thin wrapper to the Java static main entry point to allow the details of starting to be owned soley by the Java code.

original-brownbear:
@jakelandis 

> This change is still missing adding the Java classpath via .classes for debugging Java : 543b722#diff-3a9ab450779ec1f8803f713543b28b8e

Uff I was so happy to have this inconsistency gone. Does it make that much of a difference having to run `./gradlew assemble` instead of just `./gradlew compile` before running Ruby? (or am I missing some other advantage of running from `.class` files instead of the packaged `.jar` here?)

> I would think that ideally the Ruby debug entry point would be a very thin wrapper to the Java static main entry point to allow the details of starting to be owned soley by the Java code.

can't do this :( The Java entry point sets up the JRuby, the JRuby can't set up the Java entry point. Otherwise the whole exercise is somewhat pointless with regards to being able to run parallel instances of LS for testing (think clustering) => the debugger set up in the wrapping JRuby won't work in the JRuby set up by the Java entry point.

jakelandis:
> or am I missing some other advantage of running from .class files instead of the packaged .jar here?

Hot swapping (IntelliJ or integration with JRebel (or the like)). Better integration with the IDE to auto compile code under development, and it a slightly faster iteration cycle during development. For example, Junit testing from the IDE works as expected out of the box (right click run against latest code) due to auto compile.   Not show stoppers, but very nice to have and pretty standard practice to iterate .java -> .class -> test, and it isn't obvious that .java -> .class -> .jar -> test is needed to pick up new changes.  For my development workflow it makes a difference. 

> to being able to run parallel instances of LS for testing

For clarity, I only care about debugging a single instance from an IDE. I don't expect a Ruby entry point to be anything other then a developers tool. 

> can't do this :( The Java entry point sets up the JRuby, the JRuby can't set up the Java entry point. 

I haven't looked into this, but it seems plausible with some code change. 

@original-brownbear - I can take a deeper look to find a way to keep my dev workflow (debug ruby and java from .class files at the same time).

original-brownbear:
@jakelandis 

hmm makes sense, I'll add the optional use of `.class` back files then. 

> or clarity, I only care about debugging a single instance from an IDE. I don't expect a Ruby entry point to be anything other then a developers tool.

Yea, still it has to be supported down the line + you have bad timing here :D Doing this in parallel to also moving to non-`static` references to the `Ruby` runtime complicates things a lot :) But see below, for now I see a solution here.

> I haven't looked into this, but it seems plausible with some code change.

Phew we could make `org.logstash.Logstash` take a `Ruby` as an optional constructor and ensure a single Ruby that way (then the Ruby entry-point can pass it's own Ctor.).
I'll see what I can cook up here if it's of value :)

original-brownbear:
@jakelandis hmm bad news here, this is probably as good as it's gonna get in terms of a Ruby entry-point (except for the `.class` `.jar` thing, that can be fixed).

We cannot run the Java entry-point from Ruby (while staying in the same `Ruby` runtime). I tried this out and it fails for (in hindsight) obvious reasons:


```rb
require_relative "logstash-core.rb"
require "java"

org.logstash.Logstash.new.run
```
(added a no args constructor to `org.logstash.Logstash` that uses the global `Ruby`)

leads to:

```sh
Unhandled Java exception: java.lang.RuntimeException: BUG: scopes of type org.jruby.runtime.scopes.DynamicScope0 cannot grow
java.lang.RuntimeException: BUG: scopes of type org.jruby.runtime.scopes.DynamicScope0 cannot grow
                 parseFileAndGetAST at org/jruby/Ruby.java:2678
         parseFileFromMainAndGetAST at org/jruby/Ruby.java:2671
                  parseFileFromMain at org/jruby/Ruby.java:2659
                      parseFromMain at org/jruby/Ruby.java:611
                        runFromMain at org/jruby/Ruby.java:557
                                run at org/logstash/Logstash.java:101
                             invoke at java/lang/reflect/Method.java:498
  invokeDirectWithExceptionHandling at org/jruby/javasupport/JavaMethod.java:423
                       invokeDirect at org/jruby/javasupport/JavaMethod.java:290
                             <main> at Users/brownbear/src/logstash/logstash_minus_core/lib/logstash_minus_core//Users/brownbear/src/logstash/logstash-core/lib/logstash-core/logstash-ruby.rb:4
                invokeWithArguments at java/lang/invoke/MethodHandle.java:627
                          runScript at org/jruby/Ruby.java:828
                        runNormally at org/jruby/Ruby.java:747
                        runNormally at org/jruby/Ruby.java:765
                        runFromMain at org/jruby/Ruby.java:578
                      doRunFromMain at org/jruby/Main.java:417
                        internalRun at org/jruby/Main.java:305
                                run at org/jruby/Main.java:232
                               main at org/jruby/Main.java:204
```

... which makes sense if you think about it. The Java entrypoint fires of the Ruby `runner.rb` (which is not something we will be able to port to Java in the near future, because all the cli arg parsing is implemented in that Ruby code and really complex) as if it is an entrypoint itself, which it must do to have `runner.rb` work in the first place.
But if we are already coming from the above script as an initial entrypoint, the global scope for the entry is already set up in `Ruby` and cannot be expanded (... it's just not coded that way, like MRI one `ruby` process is only supposed to have one entry point).

=> the current solution is probably the best we can do without adjusting `runner.rb` :(

kares:
closing, this work has halted and I am not sure an additional ruby entry-point is needed these days ...

Issue #8906, opened 2018-01-03
Persistent Queue documentation could be clearer about its limitations.


yaauie:
The current documentation on the Limitations of Persistent Queues, while technically accurate, doesn't provide readers with sufficient context to fully understand the limitations of PQs, especially as they apply to the myriad input plugins we support.

> The following are problems not solved by the persistent queue feature:
> 1. Input plugins that do not use a request-response protocol cannot be protected from data loss. For example: tcp, udp, zeromq push+pull, and many other inputs do not have a mechanism to acknowledge receipt to the sender. Plugins such as beats and http, which do have an acknowledgement capability, are well protected by this queue.
> 2. It does not handle permanent machine failures such as disk corruption, disk failure, and machine loss. The data persisted to disk is not replicated.
>
> -- [Limitations of Persistent Queues](https://www.elastic.co/guide/en/logstash/6.1/persistent-queues.html#persistent-queues-limitations)

While a comprehensive list of plugins that use a "request-response protocol" would be helpful, such a list would likely fall out-of-date quickly as our ecosystem continues to evolve.

I believe it will be possible to improve the documentation in such a way that we could provide more clarity without maintaining such a list.

jordansissel:
"request response" is possibly a bad term given it excludes inputs like the file, s3, jdbc, etc, which also are generally protected by the PQ because they only commit a record of completed work after that work is delievered to the queue.

I spent a few minutes trying to think of a better phrase to describe these kinds of plugins (kafka, http, file, jdbc, etc) and haven't had any solid ideas on a good phrase that would clearly identify  plugins that would be can protected by the PQ

Thoughts?

mahadevans87:
Currently I use s3 input plugin in my logstash configuration. Does it work with persistent queues? I am curious to understand if it does handle back pressuring without overwhelming my elastic search cluster?

LuckyWindsck:
I am new to logstash and not sure how logstash implement PQ. Therefore, a comprehensive list will be very helpful for beginners like me. I know that it seems difficult to keep tracking the list, so is there any general way that I could check whether a plugin support PQ? (e.g. Which part of plugin source code can I check?)

anuraagvaidya:
So I was searching for the right, high-performance input plugin for an enterprise-grade application.

I know that TCP would deliver much better performance over HTTP but the persistence queue would not work over it.

I definitely don't want to use Filebeat because I'd have to write in-memory information on file for it to be able to read it, and that would increase disk usage.

[...]

PR #8927, opened 2018-01-10
Clarify PQ tuning options

### LINKED ISSUE #7027: Clarify tuning options for PQ in documentation
Currently, we do a pretty good job [documenting](https://www.elastic.co/guide/en/logstash/5.4/persistent-queues.html) some of the tuning settings available for PQ.
The flip side of documenting and exposing every setting is that users end up looking at a list of settings and wonder if there is something to tune here :)  In short, our general recommendation is to use all the defaults (which are sensible).

There are also settings we document, but at the same time noted that they are either internal and not intended to be changed.  For settings that don't really apply to the end users, it may be better if we leave them out of the documentation (and in the comments of logstash.yml - or least add a disclaimer to the comments to tell users not to change it) to avoid confusion.

Some suggestions for changes:

- Perhaps we can remove `queue.page_capacity` from documentation (and maybe also default comments in the logstash.yml) since we do not see a need for users to change the setting and it's already noted that it will not provide performance improvements.  There are also references to this setting various places on this page.  Maybe instead of directly using queue.page_capacity in these references, just describe it as page capacity (so users will not think about trying to change the setting to affect how durability or garbage collection work?)
- `queue.max_events`:  We are noting in the doc that "This value is used internally for the Logstash test suite."  But in the comments of the setting in the logstash yml, we suggest that there is a relationship between max_events and max_bytes that we check for when deciding which one applies to the queue capacity so it doesn't seem like this is only a setting that is used for our test suite.  Perhaps we can leave the test suite statement out?
- From talking to Jordan, there is a potential edge use case for actually setting `queue.max_bytes` to something much smaller (eg. 10mb) from the default, where users do not want much buffering, but do want the protections of the PQ.  Might be a good one to document.
- We mention both `queue.checkpoint.writes` and `queue.checkpoint.acks` settings in the durability section.  Per Jordan, if durability is important (and the user is willing to trade off performance), the only setting here we recommend is simply to set queue.checkpoint.writes to 1.  Other setting changes or setting queue.checkpoint.writes to another value is unlikely going to be productive.  We may want to clarify this.
- To be consistent with the first part of the documentation where we provide default values for the settings mentioned, when listing the settings available under durability section, eg. queue.checkpoint.writes, it will be nice to provide the default value as well (i.e. 1024)
- `queue.checkpoint.interval` is in the default logstash.yml settings file but not mentioned in the documentation.  If it's not intended to be changed, perhaps we can leave that out of the comments as well (or put a note in the comment that tells the users not to change it, etc..)

dedemorton:
Closes #7027 

I've made changes to this content based on the comments in https://github.com/elastic/logstash/issues/7027. 

@original-brownbear You said in your comment on issue #7027 that `queue.checkpoint.interval` will go away. Should I remove the description from the reference here? https://www.elastic.co/guide/en/logstash/current/logstash-settings-file.html

@jordansissel @yaauie If you can tell me what I need to change for https://github.com/elastic/logstash/issues/8906, I can add it to this PR while I have the PQ docs open.

### DIFF
--- a/docs/static/persistent-queues.asciidoc
+++ b/docs/static/persistent-queues.asciidoc
@@ -150,14 +150,6 @@ When recording a checkpoint, Logstash will:
 * Call fsync on the head page.
 * Atomically write to disk the current state of the queue.
 
-// REVIEWERS: I've removed the descriptions of these options here because it
-// sounds like users are confused about when to change these settings. The 
-// settings are already described in the reference topic about the logstash.yml.
-// I've also removed the statement about tuning queue.checkpoint.writes to
-// higher and lower values because it sounds like most users should either
-// accept the default or set the option to 1. 
-
-
 The process of checkpointing is atomic, which means any update to the file is saved if successful.
 
 If Logstash is terminated, or if there is a hardware-level failure, any data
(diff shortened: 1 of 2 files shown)


original-brownbear:
@dedemorton yea it def shouldn't be there, that setting doesn't work and has no effect whatsoever

Issue #8926, opened 2018-01-09
Memory queues process batches out of order - even with number of workers set to 1


robbavey:
While investigating an [issue](https://github.com/elastic/support-dev-help/issues/3125) with the elapsed filter, it looks like in Logstash 6.1+, batches are being processed in a different order from whence they came, even when the number of input workers is set to 1, when the `queue.type` is set to `memory`:

To reproduce:

In `logstash.yml` set:

```
pipeline.workers: 1
```

and use the following config to observe:

```
input { 
  generator { 
     count => 1000
  }
}

output { 
file{ path => '/Users/robbavey/sdh/output610.json' }
}
```

Observe that in the output, the sequence numbers from the generator are appearing out of order. Setting the 'pipeline.batch.size` to 10 and the generator count to 50, yields the following result:

```
{"sequence":0,"@version":"1","@timestamp":"2018-01-09T22:45:14.670Z","host":"Robs-MacBook-Pro.local","message":"Hello world!"}
{"sequence":1,"@version":"1","@timestamp":"2018-01-09T22:45:14.714Z","host":"Robs-MacBook-Pro.local","message":"Hello world!"}
{"sequence":2,"@version":"1","@timestamp":"2018-01-09T22:45:14.715Z","host":"Robs-MacBook-Pro.local","message":"Hello world!"}
{"sequence":3,"@version":"1","@timestamp":"2018-01-09T22:45:14.715Z","host":"Robs-MacBook-Pro.local","message":"Hello world!"}
{"sequence":4,"@version":"1","@timestamp":"2018-01-09T22:45:14.715Z","host":"Robs-MacBook-Pro.local","message":"Hello world!"}
{"sequence":5,"@version":"1","@timestamp":"2018-01-09T22:45:14.715Z","host":"Robs-MacBook-Pro.local","message":"Hello world!"}
{"sequence":6,"@version":"1","@timestamp":"2018-01-09T22:45:14.716Z","host":"Robs-MacBook-Pro.local","message":"Hello world!"}
{"sequence":7,"@version":"1","@timestamp":"2018-01-09T22:45:14.716Z","host":"Robs-MacBook-Pro.local","message":"Hello world!"}
{"sequence":8,"@version":"1","@timestamp":"2018-01-09T22:45:14.716Z","host":"Robs-MacBook-Pro.local","message":"Hello world!"}
[...]

PR #8928, opened 2018-01-10
MINOR: Fixed Iteration Order per Batch Worker Thread

original-brownbear:
Brings back per batch+worker iteration order by using and `ArrayList` to back the batches.

* Speeds up Java execution out of the box
* Broke `merge` on the batch => fixed dirty by simply using `.contains` instead of a uniqueness enforcing data structure, this is fine/optimal performance wise imo because:
   * Only Ruby execution uses `.merge` and it only does so for `flush`
   * Flush is "rare", iteration over the collection is super frequent => faster this way :)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
+++ b/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
@@ -1,7 +1,7 @@
 package org.logstash.common;
 
 import java.util.Collection;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
 import org.logstash.ext.JrubyEventExtLibrary;
@@ -49,8 +49,9 @@ public final class LsQueueUtils {
         final BlockingQueue<JrubyEventExtLibrary.RubyEvent> queue, final int count, final long nanos
     ) throws InterruptedException {
         int left = count;
+        //todo: make this an ArrayList once we remove the Ruby pipeline/execution
         final Collection<JrubyEventExtLibrary.RubyEvent> collection =
-            new HashSet<>(4 * count / 3 + 1);
+            new LinkedHashSet<>(4 * count / 3 + 1);
         do {
             final int drained = drain(queue, collection, left, nanos);
             if (drained == 0) {
--- a/logstash-core/spec/conditionals_spec.rb
+++ b/logstash-core/spec/conditionals_spec.rb
@@ -491,15 +491,14 @@ describe "conditionals in filter" do
     sample_one({"type" => "original"}) do
       expect(subject).to be_an(Array)
       expect(subject.length).to eq(2)
-      subject.sort! {|a, b| a.get("type") <=> b.get("type")}
-
-      expect(subject[1].get("type")).to eq("original")
-      expect(subject[1].get("cond1")).to eq("true")
-      expect(subject[1].get("cond2")).to eq(nil)
-
-      expect(subject[0].get("type")).to eq("clone")
-      # expect(subject[1].get("cond1")).to eq(nil)
-      # expect(subject[1].get("cond2")).to eq("true")
+      original_event = subject[0]
+      expect(original_event.get("type")).to eq("original")
+      expect(original_event.get("cond1")).to eq("true")
+      expect(original_event.get("cond2")).to eq(nil)
+      cloned_event = subject[1]
+      expect(cloned_event.get("cond1")).to eq(nil)
+      expect(cloned_event.get("cond2")).to eq("true")
+      expect(cloned_event.get("type")).to eq("clone")
     end
   end
 
@@ -520,20 +519,18 @@ describe "conditionals in filter" do
     CONFIG
 
     sample_one({"type" => "original"}) do
-      expect(subject.length).to eq(3)
-      subject.sort! {|a, b| a.get("type") <=> b.get("type")}
-
-      expect(subject[0].get("type")).to eq("clone1")
-      expect(subject[0].get("cond1")).to eq("true")
-      expect(subject[0].get("cond2")).to eq(nil)
-
-      expect(subject[1].get("type")).to eq("clone2")
-      expect(subject[1].get("cond1")).to eq(nil)
-      expect(subject[1].get("cond2")).to eq("true")
-
-      expect(subject[2].get("type")).to eq("original")
-      expect(subject[2].get("cond1")).to eq(nil)
-      expect(subject[2].get("cond2")).to eq(nil)
+      clone_event_1 = subject[0]
+      expect(clone_event_1.get("type")).to eq("clone1")
+      expect(clone_event_1.get("cond1")).to eq("true")
+      expect(clone_event_1.get("cond2")).to eq(nil)
+      clone_event_2 = subject[1]
+      expect(clone_event_2.get("type")).to eq("clone2")
+      expect(clone_event_2.get("cond1")).to eq(nil)
+      expect(clone_event_2.get("cond2")).to eq("true")
+      original_event = subject[2]
+      expect(original_event.get("type")).to eq("original")
+      expect(original_event.get("cond1")).to eq(nil)
+      expect(original_event.get("cond2")).to eq(nil)
     end
   end
 
@@ -586,7 +583,7 @@ describe "conditionals in filter" do
       expect(tags[6]).to eq("prev")
       expect(tags[7]).to eq("final")
     end
-    
+
     sample_one("type" => "original") do
       tags = subject.get("tags")
       expect(tags[0]).to eq("prev")


original-brownbear:
@robbavey @andrewvc @colinsurprenant there you go, iteration is ordered again :)

logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
```diff
@@ -49,8 +49,7 @@ public static void addAll(final BlockingQueue<JrubyEventExtLibrary.RubyEvent> qu
         final BlockingQueue<JrubyEventExtLibrary.RubyEvent> queue, final int count, final long nanos
     ) throws InterruptedException {
         int left = count;
-        final Collection<JrubyEventExtLibrary.RubyEvent> collection =
-            new HashSet<>(4 * count / 3 + 1);
+        final Collection<JrubyEventExtLibrary.RubyEvent> collection = new ArrayList<>(count);
```
colinsurprenant: out of curiosity, are we sure this is faster/more efficient than using, for example, a LinkedHashSet which would prevent having to check for inclusion in the ArrayList  (which requires iteration) in the `merge` method?

colinsurprenant: oh did not read the description where you actually address this point.

colinsurprenant:
Thanks. Left a comment about the performance of using `ArrayList` vs `LinkedHashSet` but realized this was addressed in the PR description :P
Even though it is a simple fix, I'll leave the final review for @robbavey who actually raised the issue in #8926, spotted the `HashSe`t ordering problem and was actually working on a fix I believe.

robbavey:
@colinsurprenant I didn't do any work on a fix other than to verify that the HashSet was the cause of #8926...

@original-brownbear:

- I'm not sure I would characterize this as 'minor', and I suspect this should also be backported to 6.1, as it has [caused problems in the field](https://github.com/elastic/support-dev-help/issues/3125)
- Might be worth running some benchmarks on the ruby exec for ArrayList vs LinkedHashSet - merge is called more frequently than per flush - https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/pipeline.rb#L449
- Tests to verify the ordering would be useful too

colinsurprenant:
@robbavey oh sorry about that, I assumed you were actually working on a PR.

> Might be worth running some benchmarks on the ruby exec for ArrayList vs LinkedHashSet - merge is called more frequently than per flush

Agreed we should double check this. 

> Tests to verify the ordering would be useful too

Yes, I really think we should include a spec which tests for ordering to avoid future regression around this. We need to figure a way to reproduce systematically a failing scenario with the `HashSet` suitable for a spec.

original-brownbear:
@robbavey @colinsurprenant we actually do have specs for this :D Unfortunately they are only run against the PQ (as are all specs pretty much -> this wasn't caught).
I would opt to not add tests here, it's pretty redundant effort right now because you'd have to write the whole bootstrap for that for Java and Ruby execution (one of which will go away, one of which will change significantly once we move the `pipeline.rb` to Java which will cause all specs to be rewritten again anyhow). 

That said ... since specs are so expensive here I'd just go for `LinkedHashSet` (thanks for spotting that one @robbavey ! apparently I was too tired for grepping this morning :() ... the performance difference is pretty small here (i.e. not reliable measurable :D) and I can just add a todo to move to `ArrayList` once the Ruby exec has gone away.

Sounds like a plan?

colinsurprenant:
Couldn't we come up with a simple spec, if not, a simple integration test which will specifically test for single worker `-w 1` ordering expectations to prevent regressions until if/we decide to not support that currently undocumented by assumed behaviour? As @robbavey mentioned this will need to be backported and I'd certainly feel safer with a test harness in place for it. 

Thoughts @andrewvc @jordansissel ?

original-brownbear:
@colinsurprenant it's really just a matter of noise. For this to have any value you'd have to run it on 4 combinations of configs (PQ y/n * Java exec y/n) and it has no long term value since the `pipeline.rb` goes away soon since we have to rewrite the Ruby specs for Java anyways => no value in adding them now in terms of preventing regressions (when you can manually verify this patch in like 60s :)).

colinsurprenant:
@original-brownbear Not following sorry. My though is to have something high-level similar to what was reported in #8926, can't we come up with something relatively simple and high level that can be fed to, yes, the different implementation combinations, maybe as an integration test?
We've have a regression introduced by this change, nothing will prevent another regression to happen unless we have a test harness around it?

colinsurprenant:
that is, if we deemed important to preserve the assumed ordering of the `-w 1` options.

original-brownbear:
@colinsurprenant hah I remembered something :) wait a few minutes, I found a way to test this that even simplifies the spec code :P

original-brownbear:
@colinsurprenant @robbavey Problem solved https://github.com/elastic/logstash/pull/8928/commits/ced52d7ccd5f393dbde36a3f26491c80567b03d6 :D I remembered that when we made the decision to drop the ordering I added sorting to this specs (since they obviously started failing) that are actually run against one worker and memory queue. => remove the sorting again => specs :P

colinsurprenant:
@original-brownbear that would work I guess! I can't figure/don't remember where it is defined that this is a single worker, so ideally (sorry I'm being a pain) I'd prefer to see specs (could be based on this one) with an explicit requirement of a single worker and an explicit ordering purpose, not just piggy back a conditional tests which tests ordering as a side effect.

original-brownbear:
@colinsurprenant it's really hard (as in 1-2 h of work potentially) to extract the code that makes these conditional tests run and refactor it in a way that allows for a more explicit tests.
Let's just add an issue that documents this and write the spec once we only have to support a single exec engine?
Again, note that any spec we write here now is thrown away once the pipeline is ported to Java + the existing tests now prevent accidentally breaking ordering => progress over perfection imo :P

colinsurprenant:
@original-brownbear we will still have to support the Ruby execution engine for a while so I do think we want to prevent regressions.

- I am not sure about your proposal of reverting the conditional spec: that spec actually failed to prevent this regression in the first place since we decided to fix the spec instead of looking into the ordering problem (note that I am missing the context of that spec change). It is also hard for me to reason about the ordering expectations in the context of the conditionals etc. 

- what do you think we add the following integration tests (or the equivalent of) to specifically test for the ordering with a single worker? 

```
bin/logstash -w 1 -e 'input{generator{count => 1000}} filter{mutate{replace => {"message" => "SEQ %{sequence}"}}} output{stdout{}}' | grep SEQ  | awk '{ print $4 }' | diff expected_sequence.txt -
```

Where `expected_sequence.txt` is the expected output:
```
0 
1
...
999
```

I manually tested it and it fails on `HashSet` and passes on `LinkedHashSet`. 
WDYT?

original-brownbear:
@colinsurprenant 

The spec didn't fail to prevent the regression, we simply didn't take the `-w 1` case into account when deciding that ordering is irrelevant in every case and hence removed the order from the spec.

I think adding another integration test that adds 30s+ to each build run for this is a really bad idea. The long build times are already and issue that is constantly brought up as affecting productivity, if we start adding this kind of test for trivial issues that are covered by existing specs (albeit not very elegantly maybe ...) we'll never get to a better spot.

Either way this discussion is taking too long ... :) If you guys feel like we need nicer specs here let's do this in another issue but not here, this should be planned properly to not make the test suit for this functionality even messier than it already is.

colinsurprenant:
> The spec didn't fail to prevent the regression, we simply didn't take the -w 1 case into account when deciding that ordering is irrelevant in every case and hence removed the order from the spec.

This is basically what I am saying. We did not have any explicit test  for ordering and so no one clued-in, for valid reasons, that this was important. If for example the tests was explicit about any ordering requirements then this regression would probable not have happened. If we "just" revert the tests for a specific order we risk facing the same issue later. It seems to me that the cost to add a "proper" test for ordering is not that high to provide safety against potential future regressions.

> I think adding another integration test that adds 30s+ to each build run for this is a really bad idea. 

But integration tests are not meant to be fast. Maybe we should reconsider when we run them? Stopping to add tests because it makes the tests longer does not make sense.

I will go ahead and test these modified specs to make sure they pass with `LinkedHashSet` and don't pass with `HashSet` on 6.1 and master and if all is good I will LGTM and we can defer further tests in another issue.

colinsurprenant:
@original-brownbear the new conditional spec does not pass on 6.1. For some reason it seems this `drain` code path is not used when running that spec on 6.1.

original-brownbear:
@colinsurprenant there's probably some fix to the test setup (there have been many) missing in `6.1`. I'll look into it when back porting, for `6.x` and `master` this should be fine though.

colinsurprenant:
There is something funky going on in 6.1 WRT the specs setup. The `WrappedAckedQueue` is used for the tests which explains ^^ but I can't find where this is setup. I believe it used to be setup in devutils but it does seems it is. yakshaving.

original-brownbear:
@colinsurprenant jup it does ... in newer LS this happens in LS itself => we need to backport this differently to `6.1` (there's a short way to do things there as well I think without back porting the whole bootstrapping in `6.x+`).

colinsurprenant:
@original-brownbear  allright then. I'll let you handle this for 6.1. Minimally I strongly suggest you add a comment in the new conditional specs explaining why `sample_one` and the ordering is important with maybe a reference to this PR.

I also strongly suggest we sort out the (6.0?) 6.1+ specs execution setup which might help catching future regressions. 

LGTM.

original-brownbear:
@colinsurprenant thanks! + yes, we def. have to sort that spec stuff out asap imo, will do in the backport PR

jordansissel:
> we actually do have specs for this :D Unfortunately they are only run against the PQ 

We must test this. If this somehow only tests the PQ, this is a gap in our tests and we need to test the memory queue as well. We have a behavior that we promise today which is that a single worker will operate on events in the order they are written by inputs.  Promised behaviors need to be backed by tests.

original-brownbear:
@jordansissel they are now backed by specs, it was a conscious choice then to drop order checking from the specs then and it's a conscious choice to check exact order now :).

jordansissel:
> they are now backed by specs

The spec changes I see in this PR are only about the order of events produced by the `clone` filter, not by the order of events processed from an input?

colinsurprenant:
@jordansissel @original-brownbear 
- I think we must sort out the spec execution context in 6+ to see if it's different than in 5 with regard to the queue strategy used for the specs.
- Order checking is still implicit within an unreleated conditionals spec and I still believe this is not the right place. Just like it happened before, the spec could be "fixed" again in the future by misunderstanding the intent of the spec.

jordansissel:
@original-brownbear I don't think the spec you updated fails in master?

If I test this PR without the implementation change:

```
% git pr 8928

# Get master's LsQueueUtils
% git checkout master logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java

# This should fail as the rspec should fail without the LsQueueUtils change
% ./gradlew build
<does not fail>
```

jordansissel:
Testing the two specific specs changed in this PR *without* the LsQueueUtils.java (to the best of my knowledge, `./gradlew build` succeeded, so this should be accurate):

```
% bin/rspec logstash-core/spec/conditionals_spec.rb -e "new events from root"
..

Finished in 6.05 seconds (files took 8.43 seconds to load)
2 examples, 0 failures

Randomized with seed 49427
```

This test should fail if it correctly detects the scenario of single-worker-out-of-order events.

original-brownbear:
@jordansissel hmmm for me it does fail reliably. Are you sure you have the changes to the specs pulled to your local correctly?
(also did you run `./gradlew assemble` before using `bin/rspec` to get the latest Java code?)

jordansissel:
>  (also did you run ./gradlew assemble before using bin/rspec to get the latest Java code?)

I ran `./gradlew build` which runs the ruby test suite (which passes) and also `bin/rspec` manually after that to spot check the specific tests.

```
% git st
M  logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
⓿ pork(~/projects/logstash) pull/8928
% git diff --staged
diff --git a/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java b/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
index 629de8abf..36ddef0a3 100644
--- a/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
+++ b/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
@@ -1,7 +1,7 @@
 package org.logstash.common;

 import java.util.Collection;
-import java.util.LinkedHashSet;
+import java.util.HashSet;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
 import org.logstash.ext.JrubyEventExtLibrary;
@@ -49,9 +49,8 @@ public final class LsQueueUtils {
         final BlockingQueue<JrubyEventExtLibrary.RubyEvent> queue, final int count, final long nanos
     ) throws InterruptedException {
         int left = count;
-        //todo: make this an ArrayList once we remove the Ruby pipeline/execution
         final Collection<JrubyEventExtLibrary.RubyEvent> collection =
-            new LinkedHashSet<>(4 * count / 3 + 1);
+            new HashSet<>(4 * count / 3 + 1);
         do {
             final int drained = drain(queue, collection, left, nanos);
             if (drained == 0) {
```

jordansissel:
I'll keep trying to get it to fail.

That said, we need an explicit test for ordering from *inputs* (from the queue). The test you modified touches the batch after it comes out of the queue (cloned events) and not necessarily tests the queue behavior explicitly.

original-brownbear:
@jordansissel hmm for me locally it fails just fine, but come to think of it ... I mean there is a chance of (1/12) that randomly passes I guess (iterating over 2 hash sets of 3 and 2 elements each -> 1/3! * 1/2! = 1/12 that it still looks ordered lol)) That's the best explanation that I can come up with. You should be able to verify that guess by simply using `new HashSet<>(4 * count / 3 + 10)` or something to shuffle the hash buckets a little ...

That said, the problem is not with the Queue, it's with the `Batch` (PQ has different batches with linked hash sets in them `Hash` is linked in JRuby) and the pipeline's interaction with it. It will take hours to produce some actual valid test for this that touches all possible combinations of queue type ... only to throw that infrastructure out the window once we move away from `pipeline.rb`.
To me it seem like a complete waste of time to do that now (plus with the damn `devutils` inconsistencies between 6.x/master and 6.1), it has 0 long term value in testing for any regressions since we won't be able to port the specs over to Java.

jordansissel:
> To me it seem like a complete waste of time to do that now 

We currently have no test that is explicitly "This tests the order of events coming from the queue". It is not a waste of time to implement one. 

> That said, the problem is not with the Queue, it's with the Batch 

This is an implementation detail, not a behavioral property. The behavior we promise is that events coming from 1 input will be received in the same order by a single-worker pipeline. There is no test for this behavior today. We need one.

original-brownbear:
@jordansissel you're the boss :) I'll to it that I find some time to write specs for all pipeline types, but this may take some time given how we hard coded the PQ into these specs :( I'll look into it tomorrow.

jordansissel:
I can at least confirm with manual testing that this patch corrects the ordering:

```
% rm /tmp/z2; bin/logstash -w1 -e 'input { generator { count => 1000 } } output { file { path => "/tmp/z2" codec => line { format => "%{sequence}" } } }'

...
# compute delta for line N and N+1, must be 1 if ordering is correct.
# print any non-"1" results
% awk 'a {print $1 - a} { a = $1 }' /tmp/z2 | awk '$1 != 1'
<should have no output>
```

Without this patch, there are several out of order items on the first time I ran it. With this patch, everything is ordered (I ran it twice to check).

colinsurprenant:
@jordansissel yes, I confirm this too, both for 6.1 and master.  As I suggested [above](https://github.com/elastic/logstash/pull/8928#issuecomment-356781022) we could add a similar integration test like that which I believe would satisfy the higher-requirement of testing for single worker event ordering.

bobrik:
I would very much like to see this fix merge. Just looking at Java stack traces is fun enough, shuffling lines around is too much.

original-brownbear:
@jordansissel merge now, spec later. It's not like changing `HashSet` to `LinkedHashSet` could break anything here, even in theory and the patch obviously works and simplifies tests?

jordansissel:
> Just looking at Java stack traces is fun enough

@bobrik I'm not sure I understand what you are saying. What do you mean?

bobrik:
@jordansissel I mean that my log lines are not in order, so stack traces are out of order.

Instead of this (frames in order):

```
Jan 11 18:55:31 myhost kafka[1665]: java.lang.NullPointerException
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkConnection.readData(ZkConnection.java:125)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient$12.call(ZkClient.java:1104)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient$12.call(ZkClient.java:1100)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient.retryUntilConnected(ZkClient.java:991)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient.readData(ZkClient.java:1100)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient.readData(ZkClient.java:1095)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.ZkUtils.readDataMaybeNull(ZkUtils.scala:660)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.ZkUtils.getController(ZkUtils.scala:255)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServer.networkClientControlledShutdown$1(KafkaServer.scala:412)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServer.controlledShutdown(KafkaServer.scala:561)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServer.$anonfun$shutdown$2(KafkaServer.scala:585)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.CoreUtils$.swallow(CoreUtils.scala:78)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.Logging.swallowWarn(Logging.scala:94)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.Logging.swallowWarn$(Logging.scala:93)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.CoreUtils$.swallowWarn(CoreUtils.scala:48)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.Logging.swallow(Logging.scala:96)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.Logging.swallow$(Logging.scala:96)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.CoreUtils$.swallow(CoreUtils.scala:48)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServer.shutdown(KafkaServer.scala:585)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServerStartable.shutdown(KafkaServerStartable.scala:48)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.Kafka$$anon$1.run(Kafka.scala:62)
```

I'm seeing this (frames out of order):

```
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServerStartable.shutdown(KafkaServerStartable.scala:48)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.Kafka$$anon$1.run(Kafka.scala:62)
Jan 11 18:55:31 myhost kafka[1665]: java.lang.NullPointerException
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServer.controlledShutdown(KafkaServer.scala:561)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.CoreUtils$.swallow(CoreUtils.scala:48)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServer.shutdown(KafkaServer.scala:585)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkConnection.readData(ZkConnection.java:125)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient$12.call(ZkClient.java:1104)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient.retryUntilConnected(ZkClient.java:991)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient.readData(ZkClient.java:1100)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient.readData(ZkClient.java:1095)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.ZkUtils.readDataMaybeNull(ZkUtils.scala:660)
Jan 11 18:55:31 myhost kafka[1665]:         at org.I0Itec.zkclient.ZkClient$12.call(ZkClient.java:1100)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.ZkUtils.getController(ZkUtils.scala:255)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServer.networkClientControlledShutdown$1(KafkaServer.scala:412)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.server.KafkaServer.$anonfun$shutdown$2(KafkaServer.scala:585)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.CoreUtils$.swallow(CoreUtils.scala:78)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.Logging.swallow(Logging.scala:96)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.Logging.swallow$(Logging.scala:96)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.Logging.swallowWarn(Logging.scala:94)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.Logging.swallowWarn$(Logging.scala:93)
Jan 11 18:55:31 myhost kafka[1665]:         at kafka.utils.CoreUtils$.swallowWarn(CoreUtils.scala:48)
```

jordansissel:
> merge now, spec later

@original-brownbear I'm OK with this proposal given we can manually test and verify it. Can you write a test for this soon? Not a blocker for this into 6.2

jordansissel:
To clarify:

LGTM to merge, but let's make sure a test (for the input -> queue -> filter ordering) is written soon.

original-brownbear:
Jenkins test this please

original-brownbear:
@jordansissel alrighty, merging then (failure is the same old #8142 :().

Will see what I can do about the tests :)

PR #8937, opened 2018-01-13
Update shared files to support versioned plugin docs

dedemorton:
This PR goes along with https://github.com/elastic/logstash-docs/pull/445 and needs to be merged at the same time.

I still have more work to do here. Rather than updating the files here, we might want to create a copy of the included files and move them to logstash-docs, where we can have versioned files. Otherwise, any updates to the common options will affect the plugin docs for all versions.


dedemorton:
Merging these without review because all the changes have been reviewed in the versioned_plugin_docs branch (except one line that simply adds a link pointing to the versioned docs).

PR #8938, opened 2018-01-13
Secretstore windows fixes.

jakelandis:
* Can't create file with POSIX flags, since not all Windows are POSIX
* Can't explicitly lock the underlying file on Windows, since the
keystore already implicity (implementation detail) locks and does not
allow any interactions if locked.  This is very similar to the *nix
behavior, except on *nix you can read while locked and need to manually
manage the lock.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
+++ b/logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
@@ -51,6 +51,7 @@ public final class JavaKeyStore implements SecretStore {
     private Lock writeLock;
     //package private for testing
     static String filePermissions = "rw-rw----";
+    private static final boolean IS_WINDOWS = System.getProperty("os.name").startsWith("Windows");
 
     /**
      * {@inheritDoc}
@@ -72,7 +73,8 @@ public final class JavaKeyStore implements SecretStore {
             LOGGER.debug("Creating new keystore at {}.", keyStorePath.toAbsolutePath());
             String keyStorePermissions = filePermissions;
             //create the keystore on disk with a default entry to identify this as a logstash keystore
-            Files.createFile(keyStorePath, PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString(keyStorePermissions)));
+            //can not set posix attributes on create here since not all Windows are posix, *nix will get the umask default and posix permissions will be set below
+            Files.createFile(keyStorePath);
             try {
                 keyStore = KeyStore.Builder.newInstance(KEYSTORE_TYPE, null, protectionParameter).getKeyStore();
                 SecretKeyFactory factory = SecretKeyFactory.getInstance("PBE");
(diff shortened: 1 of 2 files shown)


jakelandis:
Jenkins test this please

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -370,11 +372,14 @@ private void releaseLock(Lock lock) {
      * Saves the keystore with some extra meta data if needed. Note - need two output streams here to allow checking the with the append flag, and the other without an append.
      */
     private void saveKeyStore() throws IOException, CertificateException, NoSuchAlgorithmException, KeyStoreException {
-        FileLock fileLock;
+        FileLock fileLock = null;
```
original-brownbear: @jakelandis I think you can define this inside the first `try` for better readability right?

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -51,6 +51,7 @@
     private Lock writeLock;
     //package private for testing
     static String filePermissions = "rw-rw----";
+    private static boolean isWindows = System.getProperty("os.name").startsWith("Windows");
```
original-brownbear: @jakelandis can be `final`

original-brownbear: @jakelandis also since it's a constant `IS_WINDOWS` would be better :)

logstash-core/src/test/java/org/logstash/secret/store/backend/JavaKeyStoreTest.java
```diff
@@ -473,6 +473,7 @@ public void testExternalUpdateRead() throws IOException {
      */
     @Test
     public void testFileLock() throws Exception {
+        boolean isWindows = System.getProperty("os.name").startsWith("Windows");
```
original-brownbear: @jakelandis This is a really confusing test now, since assertions are wrapped in conditionals.
Maybe just split it into two tests, `testFileLockUnix` and `testFileLockWindows` and skip the one that doesn't apply for the test running system via `Assume`?

original-brownbear:
@jakelandis just some cosmetic points :)

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: done.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: done and done.

logstash-core/src/test/java/org/logstash/secret/store/backend/JavaKeyStoreTest.java
jakelandis: I removed an uneeded isWindows? assertion which only leaves 1 isWindows? assertion.    

There is now only 2 lines (out of 40ish) that are windows only, with a comment as to why. I will leave the single test, and think it is a bit less confusing now.

jakelandis:
@original-brownbear - mind taking another look ?

logstash-core/src/test/java/org/logstash/secret/store/backend/JavaKeyStoreTest.java
original-brownbear: @jakelandis sure makes sense :)

original-brownbear:
@jakelandis LGTM if it goes green :)

jakelandis:
Jenkins test this please

original-brownbear:
@jakelandis hmm seems Jenkins keeps running OOM today (def. not related to this PR, seeing the same thing in other PRs).

jakelandis:
Jenkins test this please

jakelandis:
@original-brownbear - I found some rouge Docker images still running on the build nodes and stopped them. Hopefully that will fix the Jenkins issue.

PR #8939, opened 2018-01-15
MINOR: Update readme for building zip/tar distribution

original-brownbear:
* Documented tar assembly task

Also realized, that it's dirty to not port zip then (at least as a wrapper for `rake`) and have zip and tar be built from separate build systems ... so:

* Moved wrapper for Zip build to Gradle too
   * Obviously kind of dirty now, but I'd suggest we clean this up in a separate step, by having shared logic in Gradle for assembling the archives

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,8 +167,8 @@ Note that if a plugin is installed using the plugin manager `bin/logstash-plugin
 You can build a Logstash snapshot package as tarball or zip file
 
 ```sh
-rake artifact:tar
-rake artifact:zip
+./gradlew assembleTarDistribution
+./gradlew assembleZipDistribution
 ```
 
 This will create the artifact `LS_HOME/build` directory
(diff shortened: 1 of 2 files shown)


original-brownbear:
@jsvd hmm good point:

* I fixed it here by making the `assembleTarDistribution` and `assembleZipDistribution` not call the `plugin:install-core` but simply invoke the `assemble:tar/zip` directly.
* But the problem actually still kind of exists (though it did with just `rake` too, we never really "clean up" when we bundle the artifacts, we always build them straight from the root directory => the build target isn't actually deterministic and builds a different artifact if you manually run `./gradlew test` before building) => tricky and separate issue though imo right?

jsvd:
> builds a different artifact if you manually run ./gradlew test before building) => tricky and separate issue though imo right?

correct, it should depend on clean and it is a different issue.

original-brownbear:
@jsvd nice thanks!

PR #8940, opened 2018-01-15
BUG: Fix exceptions that aren't system exit type to be swallowed

original-brownbear:
@danhermann brought this to my attention.

I missed a case here when setting up the Java entry point and swallowed all exceptions that aren't necessarily triggering a `Ruby` runtime exit. Annoyingly enough there appears to be a subset of exceptions that have to be swallowed and return `0`, those are caught by the `if` branch that is copied 1:1 from JRuby's own entry point. What was missing was to rethrow those exceptions that aren't necessarily causing System exit.

Without this fix, running LS with e.g. missing gems in the `vendor` dir just exits `0` quietly without any error logged.

No unit test, because I can't really write one as long as we have a global `Ruby` instance that I can't kill :(


jakelandis:
LGTM

original-brownbear:
@jakelandis thanks!

Issue #8918, opened 2018-01-08
Test failure: Out of memory on pq integration test


jordansissel:
https://logstash-ci.elastic.co/job/elastic+logstash+6.1+multijob-integration-pq-2/42/console

```

21:57:17   1) CLI > logstash-plugin prepare-offline-pack create a pack from a wildcard successfully create a pack
21:57:17      Failure/Error: expect(unpacked.plugins.collect(&:name)).to include(*filters)
21:57:17        expected ["logstash-filter-split", "logstash-filter-geoip", "logstash-filter-aggregate", "logstash-filter-elas...ze", "logstash-filter-metrics", "logstash-filter-cidr", "logstash-filter-csv", "logstash-filter-kv"] to include "Bundler::GemspecError: There was a Errno::ENOMEM while loading logstash-filter-qatest-newer.gemspec: ", "Cannot allocate memory - git ls-files from", "  /opt/logstash/build/logstash-6.1.2-SNAPSHOT/vendor/local_gems/449e2de5/logstash-filter-qatest-0.1.1/logstash-filter-qatest-newer.gemspec:13:in `block in eval_gemspec'", <output truncated manually for readability>
```

original-brownbear:
I think I tracked down the reason Zombie containers pile up and cause us to run OOM in https://github.com/elastic/logstash/pull/8953 to a Docker bug. Let's see if we can verify this somehow :)

original-brownbear:
@jakelandis Do you have any update on whether or not we still see Docker zombies that cause this? :)

jakelandis:
I checked the 3 docker hosts and no zombies! 

Interestingly there were docker instances running from a recent (~5min)  PR that was superseded by a new commit. It seems that we (Jenkins ?) is not killing the running docker container for the same PR, even though the Jenkins job is done. I vaguely remember running into this before and thought I fixed it ... will log a follow up issue.

jakelandis:
Follow up issue https://github.com/elastic/logstash/issues/9017

Closing this issue since it appears that Zombie issue is resolved.

PR #8941, opened 2018-01-15
Rake Java Wrapper

original-brownbear:
* Run `rake` in Gradle JVM to finally fix portability issues for good (now the shell env. shouldn't have any part in what the `rake` tasks do whatsoever since we're using JRuby from a `.jar` and loading the initial `rake` library as a system gem directly from that `.jar`)
   * Hence the CI env change to give it 2Gb memory, overall we should use a lot less memory for the build though since we save a full JVM for the `rake` child process
* Also: Fix wrong output file in zip artifact task (my bad like 2h ago, no need for a separate PR)
* Should make hacky attempts at this redundant (e.g. #8896)
* Its are now portable to the degree that they run on a clean Debian Squeeze out of the box with just `apt-get update && apt-get install openjdk-8-jdk gradle && ./gradlew test`
* Approach copied from https://github.com/logstash-plugins/logstash-integration-elasticsearch

=> next step would be turning the ITs into a Gradle build as well and take the same approach there imo

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,7 @@ apply plugin: 'de.undercouch.download'
 import de.undercouch.gradle.tasks.download.Download
 import de.undercouch.gradle.tasks.download.Verify
 import org.logstash.gradle.ExecLogOutputStream
+import org.logstash.gradle.RubyGradleUtils
 import org.yaml.snakeyaml.Yaml
 
 allprojects {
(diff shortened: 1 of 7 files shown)


original-brownbear:
Jenkins test this please

original-brownbear:
Build failure is unrelated #8918

original-brownbear:
Jenkins test this please

original-brownbear:
@jsvd maybe you can take a look here since we already looked into the hacky approaches before :D I think this should fix things for good :)

original-brownbear:
@jsvd nice thanks :)

PR #8942, opened 2018-01-15
[WIP] Batching queue timeouts

andrewvc:
This is very much a WIP, no tests, probably has bugs. What I'm looking for is feedback on whether the general approach here makes sense.

We've wanted to add queue timeouts + batching to the queues for a while. The way to do it for the in-memory ones is quite obvious, but it's a little trickier for the persisted one. This is a first attempt.

This attempt will do both operations in one go, it kinda makes sense that if you need one you need teh other. For instance, the HTTP input wants to be able to get a batch of events and either write it all at once or fail completely within a certain amount of time, letting the client fully retry.

Would love some feedback here. The PQ is something I'm only partially familiar with. I'm sure there are bugs and perhaps worse errors.


logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
```diff
@@ -79,6 +80,27 @@ public void write(byte[] bytes, long seqNum, int checkpointMaxWrites) throws IOE
         }
     }
 
+    public void writeMany(List<byte[]> byteList, long seqNum, int checkpointMaxWrites) throws IOException {
```
original-brownbear: @andrewvc I think it would be good to consumer `List<Queueable>` here instead of `List<byte[]>`. Tbh. any API using `byte[]` in a serialize then write setting is kinda horrible. Let's not add more of them. See #7039 for some more on this.

I'd just do the serialization in line here => have the `byte[]` live for a shorter time + stay more flexible in terms of fixing our serialization eventually :)

original-brownbear: @andrewvc nevermind I see now that we do the size check upstream ... urgh :( but can't fix this here ... dropping above point

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -315,16 +316,26 @@ private void newCheckpointedHeadpage(int pageNum) throws IOException {
         this.currentByteSize += headPageIO.getCapacity();
     }
 
+    public long write(Queueable element) throws IOException {
+        return writeMany(Collections.singletonList(element), -1);
```
original-brownbear: @andrewvc I think this should be handled differently in the final implementation. Either we should drop unbatched writes or not make this API slower by adding the collection wrapper to each element. The former is probably the way to go since there is no plugins directly using the queue here.

original-brownbear:
@andrewvc I'm a little torn here :)

On one hand, this looks fine working from what we have and I'd have mainly cosmetics remarks to make here (no need for those yet since we're WIP).
On the other hand, this really adds a ton of new APIs that are very suboptimal, e.g. `List<byte[]>` is just wrong and it seems like this change **could** (hard to predict) way increase the memory use when working with large batches (since you're now serializing batch-wise while keeping the actual events on heap at the same time as well while you previously only had a single serialized event per batch serialized on heap).

=> Since we're in a bit of a hurry here as far as I understand, I'd benchmark this to avoid surprises but go for it begrudgingly :)

andrewvc:
@original-brownbear my apologies, I wasn't clear as far as the stage this PR was in. I was mostly looking for feedback on the correctness of this approach, not the performance. In previous discussions it was mentioned that timing out the PQ was difficult given its implementation, that's what I wanted to tackle first.

I agree `List<byte[]>` is silly. It was just the simplest thing to do, I never intended for that to be what we actually merged.

So, we could pick a more optimal structure than `List<byte[]>`, maybe just a growable `byte[]` with offsets in a secondary array but that doesn't solve your core concern of heap memory usage.

Let's do the numbers on that using the HTTP input, which is a likely user of this feature. Let's pick a average and a high use case.

**Mid-size: **
* 16 concurrent writers, batch size 500 = 8000 events in flight.
* Event size = 1k serialized
* Total heap increase (vs today's usage) = 8MiB

**High-size: **
* 32 concurrent writers, batch size 2000 = 64000 events in flight.
* Event size = 4k serialized
* Total heap increase (vs today's usage) = 256MiB

That's a lot, but the actual structures are very simple, (esp. if we move away from `List<Byte[]>` unless there's something I'm missing. I'd say for that high use case 256MiB in heap isn't too large a price to pay, esp. for a relatively small number of objects in eden GC.

I also think we might get a CPU win here by having tighter loops around serialization. We'll move from:

`events.each {|e| queue.write(e.serialize)}`
to
`queue.write(events.map {|e| e.serialize})`

I've seen more optimal JVM execution in the latter case, though I'm not entirely certain of which optimization provides that.

There's also a small win here from more efficient IO ops due to the transactional nature of these writes.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
andrewvc: Yes, this was just done for the purposes of the PoC as explained below.

original-brownbear:
@andrewvc yea I agree, there's potential wins and losses here :) Just saying we should test this. It's somewhat optimistic to simply look at the size of serialized `byte[]` here since it's more a matter of allocations * average_lifetime here so this could go either way imo.
* You're def a lot faster per `byte[]` I agree :)
* But the product of average_lifetime * size also goes up quite a bit, which may or may not be a problem.
   * though if we do something a little more efficient than `List<byte[]>` like say ... just `byte[]` + `int[]` pointers this may be irrelevant and we got a huge win overall :)

andrewvc:
After discussing with @original-brownbear on slack, we decided that what makes the most sense is a shareable `byte[]` in the queue for writing the batch data. Then we just write it all as one mmap chunk to the page. We'll need to improve `Queue#serialize` to be able to write into a `byte[]` however.

I'll have to see what the scope is there. One worry is that a queueable can be of an arbitrary size, so allocating a `byte[]` can be tricky unless we set a maximum event size ( and preferably maximum batch size as well )`. That is one advantage of the `List<byte[]>` approach.

The other thought is to push down the serialization as close as possible to the page write, thus negating the need to pre-allocate. Thinking about it, that is preferable.

original-brownbear:
@andrewvc 

> The other thought is to push down the serialization as close as possible to the page write, thus negating the need to pre-allocate. Thinking about it, that is preferable.

+100 on this, but I don't see how this can be done without a significant rewrite since the byte counting logic happens way up the stack in `Queue`.

Also:

Feel free to just go for `List<byte[]>` and we'll fix it later I guess. I stand by the statement that it's horrible :), but we already do exactly that in the deserialization as well => we can probably just fix this in some follow-up work and handle both the serialization and deserialization in one go :)

andrewvc:
@original-brownbear ++ , yeah, let's benchmark this naive approach, and if it isn't terrible merge.

andrewvc:
This is now rebased (and basically rewritten, so much stuff has changed since the original I had to start over). I ran same preliminary benchmarks against a patched generator input, set to send batches of 500.

For `rm -rf data; time bin/logstash -e 'input { generator { count => 5000000 } } output {  file { codec => dots path => "/tmp/dots" } }’`

So, a huge improvement there.

```
master memory
35.87 real       117.07 user         6.95 sys

master PQ drain
120.02 real       189.10 user        50.58 sys

Batch PQ drain + Min amortized puts
68.77 real       158.72 user        21.15 sys
```

original-brownbear:
@andrewvc good stuff :) let me know when this is ready for review.

andrewvc:
Yes, ready for review. I want to add some tests of timeout behavior.

Do you think we should add some new batch based stress tests?

original-brownbear:
@andrewvc no please no more stress tests :D let's just implement the concurrency correctly instead of trying to uncover concurrency issues by force :)

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -337,63 +387,91 @@ public long write(Queueable element) throws IOException {
...
+        if (persistedBatchSize <= headPageCapacity) {
+            lock.lock();
+            try {
+                maxSeqNum = unsafeQueueWrite(serializedQueueables, timeoutMillis, persistedBatchSize);
+            } finally {
+                lock.unlock();
             }
+        } else {
+            // Batch is larger than a page, we can't do an atomic write
+            throw new IOException("data to be written is bigger than page capacity");
```
original-brownbear: @andrewvc this seems a little risky? Shouldn't we try to split the batch up across pages instead of just erroring out here? This could lead to trouble with random very large events like we've seen in the past I think.
It's obviously more logic to handle that, but maybe worth it?

logstash-core/src/main/java/org/logstash/ext/JrubyAckedWriteClientExt.java
```diff
@@ -50,16 +56,15 @@ private JrubyAckedWriteClientExt(final Ruby runtime, final RubyClass metaClass,
...
         return this;
     }
 
     @JRubyMethod(name = "push_batch", required = 1)
     public IRubyObject rubyPushBatch(final ThreadContext context, IRubyObject batch) {
         ensureOpen();
-        for (final IRubyObject event : (Collection<JrubyEventExtLibrary.RubyEvent>) batch) {
-            queue.rubyWrite(context, ((JrubyEventExtLibrary.RubyEvent) event).getEvent());
-        }
+        List<Queueable> events = ((List<JrubyEventExtLibrary.RubyEvent>) batch).stream().map(e -> e.getEvent()).collect(Collectors.toList());
```
original-brownbear: @andrewvc this is pretty expensive. I think going a little more old school here and simply creating a new ArrayList(batch.size()) and then filling it in a for loop fill be quite a bit faster.
On that note ... why not simply create the List<byte[]> here and adjust the downstream APIs. Unlike the name suggests those aren't used by Ruby anymore so you're safe to change them and save one List and conversion step :)

original-brownbear:
@andrewvc looks fine in general. One performance suggestion and one stability concern :)

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
andrewvc: I thought about this, and in fact, my initial implementation did this (since squashed).

Unfortunately, one of the goals here is to allow for atomic writes of the batch, for things like the HTTP input. The idea is that a write can timeout and not have written anything.

I couldn't figure out a way to do that without this limitation. We could, of course, figure out some sort of way, but it seemed really complex with the way we juggle pages, and would be a major refactor. Maybe there's an easy way to do this?

original-brownbear: @andrewvc hmm, maybe it's not worth the trouble and we could simply write a larger page in this case?
I guess the other option would be to not be so strict about the atomic part and simply split up the list and write to multiple pages in a loop in this method.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -319,16 +327,58 @@ private void newCheckpointedHeadpage(int pageNum) throws IOException {
...
+     * @throws IOException
+     */
+    public long write(Queueable queueable) throws IOException {
+        return write(queueable, -1);
+    }
+
+    /**
+     * Write a {@link Queueable} element to the queue. This will block until the write succeeds or timeout occurs
+     * @param queueable
+     * @param timeoutMillis maximum time to block until returning
```
jakelandis: nit: can you add `-1` means no timeout ?  (assuming that is correct)

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
andrewvc: Spoke to @original-brownbear directly, we decided that our best option is the current code, which just makes a new page if the current one is too small.

Will add a note to the docs that the page size must be > the largest batch written.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
andrewvc: Good call! Yes I can.

andrewvc:
I believe I've made all the requested improvements. I also caught something I missed the first time around, I wasn't timing out on the lock acquisition itself! https://github.com/elastic/logstash/pull/8942/commits/dc1f932ced440890509826ce40369ab78e4d7954 fixes that, in a (hopefully) cheap way. This probably needs another round of benchmarking.

I'm thinking that the batch push method is probably a bad idea outside of 'natural' batches like the HTTP input. Having too large a batch could push past the 64MB page size boundary. I should probably improve the docs there. 

Given the notable perf gains here I think it makes sense to add a `turbo` mode to the queue, where each write client maintains its own internal buffer that flushes according to the queue checkpoint writes logic. It seems like the locking here is most likely really expensive, and avoiding that is worth it. We'd lose a little bit of durability, but practically speaking, that's the tradeoff many queues make.

That's out of scope for this PR however.

logstash-core/src/main/java/org/logstash/ackedqueue/Queue.java
```diff
@@ -319,16 +320,64 @@ private void newCheckpointedHeadpage(int pageNum) throws IOException {
...
+     * @return the maximum written sequence number, or -1 if timed out
      * @throws IOException
      */
-    public long write(Queueable element) throws IOException {
-        byte[] data = element.serialize();
+    public long write(List<Queueable> queueables, long timeoutMillis) throws IOException {
+        if (queueables.isEmpty()) return -1;
+
+        try {
+            timeoutMillis -= optionallyLockForTimeout(timeoutMillis);
```
original-brownbear: @andrewvc this doesn't seem safe. We should call `optionallyLockForTimeOut` in front of a `try-finally` construct and do whatever we have to do inside that `try`.
This implementation would e.g. block forever on a serialization exception.

That said, I think we can do this easily by simply moving this all below the serialization `for` loop. That would also have the added benefit of not needlessly holding on to the lock while serializing :)

andrewvc: Yeah, you're right, I don't know what I was thinking.

andrewvc:
Closing in favor of https://github.com/elastic/logstash/pull/9498

PR #8948, opened 2018-01-16
MINOR: Simplify inflight batch tracking

original-brownbear:
This one was long overdue :D Motivation here was initially that `yield` is hard to port to Java so I redid the `inflight_batches` method and its only called to not require `yield`.

* Just return instead of `yield` the map, `yield` was only necessary when we needed locking around the iteration for the map, this was already redundant for the sync queue since there there was no lock around the `yield`
* Needed a concurrency safe iterator => had that already in the sync queue => moved the acked queue to use a concurrent map as well (the solution is safe since all operations on the Java queue are locked on the Java side, same as for the `ArrayBlockingQueue` that back the sync queue implementation which works fine without locks and just using a concurrent hashmap)

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -96,10 +96,9 @@ module LogStash; module Util
 
       def initialize(queue, batch_size = 125, wait_for = 50)
         @queue = queue
-        @mutex = Mutex.new
         # Note that @inflight_batches as a central mechanism for tracking inflight
         # batches will fail if we have multiple read clients in the pipeline.
-        @inflight_batches = {}
+        @inflight_batches = Concurrent::Map.new
         # allow the worker thread to report the execution time of the filter + output
         @inflight_clocks = Concurrent::Map.new
         @batch_size = batch_size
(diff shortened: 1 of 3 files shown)


original-brownbear:
Jenkins test this please

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -168,24 +157,14 @@ def read_batch
...
       def close_batch(batch)
         thread = Thread.current
-        @mutex.lock
-        begin
-          batch.close
-          @inflight_batches.delete(thread)
-        ensure
-          @mutex.unlock
-        end
+        batch.close
```
jsvd: I can't come up with a scenario where this would break but could we have a race bug here by not wrapping both instructions into a lock?

original-brownbear: @jsvd logically impossible I think because `batch.close` doesn't reference the state of `inflight_batches` and neither is there any code that references the fact `close` was called on the `batch` in relation to modifying `inflight_batches` => I think we're safe + we do the same in the other queue anyways and haven't seen issues with that in months :)

original-brownbear:
@guyboertje maybe you can take a second look here? You reviewed the first step here (moving to concurrent hash map and removing the lock in the memory queue a) a while back :)

original-brownbear:
@danhermann since this is blocking you too ... if you feel comfortable enough with this thing after our chat, feel free to be the 2nd reviewer here :) This isn't really a complicated one imo.

danhermann:
LGTM -- I've done some additional testing on this since I need it for one of my PRs.

original-brownbear:
@danhermann thanks merging :)

PR #8954, opened 2018-01-16
MINOR: Simplify Java Entrypoint

original-brownbear:
Just some cleanup, it's confusing and needless to track the `config` and `script` as separate fields (especially for the `script` + `run` combination this is outright wrong).

=> properly handle the `script` resource in the `run` scope only, it's not used anywhere else.
=> don't separately track the `config` from `ruby` since `ruby` references the `config` anyways this is just a bug waiting to happen :)


original-brownbear:
@danhermann thanks :)

PR #8956, opened 2018-01-16
JAVAFICATION: Move in memory queue read batch to Java

original-brownbear:
Small step for moving this code to Java end to end, simply porting the `ReadBatch` to Java as is, no functional changes (except for not making it a child of the sync queue, because that's tricky when I want to set up the class from Java :)).

=> Should be a little faster now since we can pre-size the `RubyArray` in `toArray` in the hot path and avoid a bunch of reallocations there.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
+++ b/logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
@@ -45,12 +45,12 @@ public final class LsQueueUtils {
      * @throws InterruptedException On Interrupt during {@link BlockingQueue#poll()} or
      * {@link BlockingQueue#drainTo(Collection)}
      */
-    public static Collection<JrubyEventExtLibrary.RubyEvent> drain(
+    public static LinkedHashSet<JrubyEventExtLibrary.RubyEvent> drain(
         final BlockingQueue<JrubyEventExtLibrary.RubyEvent> queue, final int count, final long nanos
     ) throws InterruptedException {
         int left = count;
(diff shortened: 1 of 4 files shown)


logstash-core/src/main/java/org/logstash/common/LsQueueUtils.java
```diff
@@ -45,12 +46,12 @@ public static void addAll(final BlockingQueue<JrubyEventExtLibrary.RubyEvent> qu
      * @throws InterruptedException On Interrupt during {@link BlockingQueue#poll()} or
      * {@link BlockingQueue#drainTo(Collection)}
      */
-    public static Collection<JrubyEventExtLibrary.RubyEvent> drain(
+    public static Set<JrubyEventExtLibrary.RubyEvent> drain(
```
danhermann: I didn't understand this change from `Collection<>` to `Set<>`. So far as I see, this method is called only with a parameter of type `LinkedHashSet`, but I believe we need ordering for certain codecs here, right?

original-brownbear: @danhermann right this isn't good, this should be `LinkedHashSet` not `Set` my bad, will fix later :)

original-brownbear:
@danhermann fixed the type to `LinkedHashSet` hard set everywhere :) Should be fine now I think/hope :)

danhermann:
LGTM!

original-brownbear:
@danhermann thanks!

PR #8957, opened 2018-01-16
MINOR: Remove unnecessary logger fields from filter and output delegators

original-brownbear:
Random find when investigating what we could bring to Java next :)

* `@logger` fields are never used
   * can completely remove them everywhere for the filter delegator
   * have to still pass the logger through in the constructor of the output delegator but don't need that field either

=> making it at least slightly easier to bring this over to Java :)

### DIFF
--- a/logstash-core/lib/logstash/filter_delegator.rb
+++ b/logstash-core/lib/logstash/filter_delegator.rb
@@ -16,8 +16,7 @@ module LogStash
 
     attr_reader :id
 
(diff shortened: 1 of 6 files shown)


original-brownbear:
Jenkins test this

danhermann:
LGTM assuming it goes green.

original-brownbear:
@danhermann thanks, failure is unrelated slowlog issue, will open a ticket for that in a sec. That's been happening a few time today.

Issue #8936, opened 2018-01-12
[META] PQ Disk Size 


colinsurprenant:
This is a meta issue to regroup and track related PQ disk size issues. 

- [x] Logging when queue.max_bytes is reached https://github.com/elastic/logstash/pull/8400 
  - closed, should be done through API checks
  <br/>
- [x] Logging when queue.max_bytes is reached https://github.com/elastic/logstash/issues/7491
  * closed, in #8400 it was decided to drop that idea
  <br/>
- [x] add queue_size_on_disk stat to persisted queue node stats https://github.com/elastic/logstash/pull/7010
  * introduced  getPersistedByteSize
  <br/>
- [x] Check Available Disk Space for Persisted Queues https://github.com/elastic/logstash/pull/6998
  * proposes a new PQ bootstrap check
  * implements ensureDiskAvailable and uses it in Queue.open
  * we should iron out the last details and merge this
  * rebooted into #8978
  <br/>
- [x] Add bootstrap checks to persistent queue https://github.com/elastic/logstash/issues/6552
  * Issue which will be fixed by #6998
  <br/>
- [x] queue.max_bytes isn't respected at start up time https://github.com/elastic/logstash/issues/6625
  * Issue which will be fixed by #6998
  <br/>
- [x] queue.max_bytes isn't compared against queue.page_capacity https://github.com/elastic/logstash/issues/6627
  * Issue which will be fixed by #6998 
  <br/>
- [x] Modify queue API response https://github.com/elastic/logstash/issues/6508
  * suggests to remove current_size_in_bytes and replace it with queue_size_on_disk which is a percent value calculated from current_size_in_bytes (the accurate value, not logical) and max_queue_size_in_bytes
  * need to assess if still valid?
  <br/>
- [x] Disk level stats for monitoring API https://github.com/elastic/logstash/issues/6884
  * Porposes various PQ/DLQ disk usage stats API
  <br/>
- [x] [DISCUSS] PQ disk size usage APIs #8986 
  <br/>
- [x] Add PQ disk size fields in x-pack monitoring payload

colinsurprenant:
It seems that from #6508, #6518 and #7010 we have settled to using the sum of the actual persisted bytes (head position) for each sparsely allocated page of the queue to compute the queue disk size. This strategy is consistent with what `du` on Linux would report but may show different sizes than `ls`.  As a first step I will go ahead and remove `currentByteSize`, `getCurrentByteSize` and `current_byte_size`.

[...]

PR #8958, opened 2018-01-16
remove currentByteSize

colinsurprenant:
Relates to #8936

This removes `currentByteSize` handling in PQ which computes queue size based on page capacity times the number of live pages. 

Per #6508, #6518 and #7010 we have settled to using the sum of the actual persisted bytes (head position) for each sparsely allocated page of the queue to compute the queue disk size. There is no reason to keep this `currentByteSize` around. 

**IMPORTANT** This PR introduces an inefficiency in the `isFull()` method which now needs to call `getPersistedByteSize()`  instead of using the `volatile long currentByteSize`. Calling `getPersistedByteSize()` is clearly less efficient because it needs to iterate through all the tail pages to sum the head position but it is now a "correct" implementation. I suggest we move forward with this change here and open a new issue to work on a way to optimize this. 

Note that while doing this I realized that the tests using `singleElementCapacityForByteBufferPageIO` for computing the queue max size were wrong and I added `multipleElementsCapacityForByteBufferPageIO` to correctly do this.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
@@ -4,6 +4,7 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.util.BitSet;
 
+import org.codehaus.commons.nullanalysis.NotNull;
 import org.logstash.ackedqueue.io.CheckpointIO;
 import org.logstash.ackedqueue.io.LongVector;
 import org.logstash.ackedqueue.io.PageIO;
@@ -23,7 +24,7 @@ public final class Page implements Closeable {
     protected BitSet ackedSeqNums;
     protected Checkpoint lastCheckpoint;
 
-    public Page(int pageNum, Queue queue, long minSeqNum, int elementCount, long firstUnreadSeqNum, BitSet ackedSeqNums, PageIO pageIO, boolean writable) {
+    public Page(int pageNum, Queue queue, long minSeqNum, int elementCount, long firstUnreadSeqNum, BitSet ackedSeqNums, @NotNull PageIO pageIO, boolean writable) {
         this.pageNum = pageNum;
         this.queue = queue;
 
@@ -34,6 +35,8 @@ public final class Page implements Closeable {
         this.lastCheckpoint = new Checkpoint(0, 0, 0, 0, 0);
         this.pageIO = pageIO;
         this.writable = writable;
+
+        assert this.pageIO != null : "invalid null pageIO";
     }
 
     public String toString() {
@@ -238,15 +241,11 @@ public final class Page implements Closeable {
 
     public void close() throws IOException {
         checkpoint();
-        if (this.pageIO != null) {
-            this.pageIO.close();
-        }
+        this.pageIO.close();
     }
 
     public void purge() throws IOException {
-        if (this.pageIO != null) {
-            this.pageIO.purge(); // page IO purge calls close
-        }
+        this.pageIO.purge(); // page IO purge calls close
     }
 
     public int getPageNum() {
--- a/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
+++ b/logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
@@ -32,7 +32,7 @@ import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.fail;
-import static org.logstash.ackedqueue.QueueTestHelpers.singleElementCapacityForByteBufferPageIO;
+import static org.logstash.ackedqueue.QueueTestHelpers.computeCapacityForMmapPageIO;
 
 public class QueueTest {
 
@@ -174,9 +174,8 @@ public class QueueTest {
     @Test
     public void writeMultiPage() throws IOException {
         List<Queueable> elements = Arrays.asList(new StringElement("foobarbaz1"), new StringElement("foobarbaz2"), new StringElement("foobarbaz3"), new StringElement("foobarbaz4"));
-        int singleElementCapacity = singleElementCapacityForByteBufferPageIO(elements.get(0));
         try (Queue q = new Queue(
-            TestSettings.persistedQueueSettings(2 * singleElementCapacity, dataPath))) {
+            TestSettings.persistedQueueSettings(computeCapacityForMmapPageIO(elements.get(0), 2), dataPath))) {
             q.open();
 
             for (Queueable e : elements) {
@@ -218,9 +217,8 @@ public class QueueTest {
     @Test
     public void writeMultiPageWithInOrderAcking() throws IOException {
         List<Queueable> elements = Arrays.asList(new StringElement("foobarbaz1"), new StringElement("foobarbaz2"), new StringElement("foobarbaz3"), new StringElement("foobarbaz4"));
-        int singleElementCapacity = singleElementCapacityForByteBufferPageIO(elements.get(0));
         try (Queue q = new Queue(
-            TestSettings.persistedQueueSettings(2 * singleElementCapacity, dataPath))) {
+            TestSettings.persistedQueueSettings(computeCapacityForMmapPageIO(elements.get(0), 2), dataPath))) {
             q.open();
 
             for (Queueable e : elements) {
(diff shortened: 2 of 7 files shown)


colinsurprenant:
If we agree on this change I will go ahead and create an issue for solving the `isFull()` / `getPersistedByteSize()` performance problem.

logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
```diff
@@ -34,6 +34,8 @@ public Page(int pageNum, Queue queue, long minSeqNum, int elementCount, long fir
         this.lastCheckpoint = new Checkpoint(0, 0, 0, 0, 0);
         this.pageIO = pageIO;
         this.writable = writable;
+
+        assert this.pageIO != null : "invalid null pageIO";
```
jordansissel: Do you think adding `@NotNull` annotation to the `pageIO` parameter could be a useful replacement for this assertion?

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
```diff
@@ -174,9 +175,8 @@ public void multiWriteSamePage() throws IOException {
     @Test
     public void writeMultiPage() throws IOException {
         List<Queueable> elements = Arrays.asList(new StringElement("foobarbaz1"), new StringElement("foobarbaz2"), new StringElement("foobarbaz3"), new StringElement("foobarbaz4"));
-        int singleElementCapacity = singleElementCapacityForByteBufferPageIO(elements.get(0));
         try (Queue q = new Queue(
-            TestSettings.volatileQueueSettings(2 * singleElementCapacity))) {
+            TestSettings.volatileQueueSettings(multipleElementsCapacityForByteBufferPageIO(elements.get(0), 2)))) {
```
jordansissel: This function name is delightfully long. No change required; just noting that I chuckled a bit when noticing the length ;)

Future possibly just rename both of these methods (multiple and single) to `computeCapacityForByteBufferPageIO(element, count)`  and use `1` for count when using the single case. (again , no change required)

jordansissel:
@colinsurprenant I think this is the right approach (counting the actual bytes-written to each page).

logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
colinsurprenant: AFAIK `@NotNull` will not result in run-time check and it does not enforce that the method cannot be called with a `null` argument. It is more for external tools. I can definitely add it.
Note that I am also on the fence about using an assertion to do parameter validation since assertions can be disabled. My thinking here is that I looked and could not see a case where a `null` `pageIO` would be passed but adding this assertion would make the test break if there is an oversight in future refactoring. It also pairs with the removal of the `null` checks in the `close()` and `purge()` methods. WDYT?

logstash-core/src/test/java/org/logstash/ackedqueue/QueueTest.java
colinsurprenant: +1

colinsurprenant: renamed

logstash-core/src/main/java/org/logstash/ackedqueue/Page.java
colinsurprenant: added.

colinsurprenant:
@jordansissel added commits following your suggestions.

colinsurprenant:
I'll wait to get a green light before looking at the merge conflicts.

colinsurprenant:
@jordansissel ping

jordansissel:
Testing now.

jordansissel:
```

% curl -s localhost:9600/_node/stats/pipelines?pretty | jq .pipelines.main.queue.capacity.queue_size_in_bytes; du -bs data/queue
940427860
1006637584      data/queue

% curl -s localhost:9600/_node/stats/pipelines?pretty | jq .pipelines.main.queue.capacity.queue_size_in_bytes; du -bs data/queue
1073741902
1140855380      data/queue
```



hypothesis: Seems the queue_size_in_bytes is always off by one page?

```
# All pages size according to `wc -c`
% wc -c data/queue/main/page.*
  67108864 data/queue/main/page.0
  67108864 data/queue/main/page.1
  67108864 data/queue/main/page.10
  67108864 data/queue/main/page.11
  67108864 data/queue/main/page.12
  67108864 data/queue/main/page.13
  67108864 data/queue/main/page.14
  67108864 data/queue/main/page.15
  67108864 data/queue/main/page.16
  67108864 data/queue/main/page.2
  67108864 data/queue/main/page.3
  67108864 data/queue/main/page.4
  67108864 data/queue/main/page.5
  67108864 data/queue/main/page.6
  67108864 data/queue/main/page.7
  67108864 data/queue/main/page.8
  67108864 data/queue/main/page.9
1140850688 total

# Number of page files:
% ls data/queue/main/page.* |wc -l
17
```

Math:

* `.pipelines.main.queue.capacity.queue_size_in_bytes` (1073741902) / page size (67108864) == 16.000001
* `du -sb` (1140855380/ 67108864.0) == 17.00006

Based on the above results, I am assuming queue_size_in_bytes is missing one page (perhaps the head page?)

colinsurprenant:
thanks @jordansissel - investigating. I should also probably rebase at this point.

colinsurprenant:
tedious rebase done, now onto the investigation.

colinsurprenant:
@jordansissel when you watched & compared sizes with the API and `du` while generator is still running, isn't it normal that there is a difference between the two measurements?

colinsurprenant:
Ok, just clarified with @jordansissel that he waited for the queue to fill up so that the generator would stop by backpressure and then take the measurements. Makes sense, investigating.

jordansissel:
Notes from @colinsurprenant and I chatting on zoom ---

The final number (1140855380 bytes via `du`) was after the pipeline had stalled due to the PQ being full. 1140855380 is ~1.06gb and pretty close to the default 1gb max queue size.

jordansissel:
haha, crossing the streams ;)

colinsurprenant:
@jordansissel here's what is happening. I actually reproduced your observation on my Ubuntu 16 VM using the following (to limit to a small queue size)
```
queue.type: persisted
queue.max_bytes: 256mb
```

and launched LS but with a longer sleep to make sure nothing was ever outputted - this shouldn't have any impact but JTBS. 
```
bin/logstash -e 'input { generator { } } filter { sleep { time => 10000 } }'
```

This filled the queue:
```
du -cbs data/queue/main/page.*
67108864	data/queue/main/page.0
67108864	data/queue/main/page.1
67108864	data/queue/main/page.2
67108864	data/queue/main/page.3
67108864	data/queue/main/page.4
335544320	total
```

But the API reports `"queue_size_in_bytes" : 268435725` :
```
curl -s localhost:9600/_node/stats/pipelines?pretty
{
...
      "queue" : {
        "events" : 958574,
        "type" : "persisted",
        "capacity" : {
          "queue_size_in_bytes" : 268435725,
          "page_capacity_in_bytes" : 67108864,
          "max_queue_size_in_bytes" : 268435456,
          "max_unread_events" : 0
        },
...
}
```

and as you realized the api reported `268435725` is almost equal to the `du` total `335544320` minus a single page size of `67108864` -> `268435456`, a 269 bytes difference. 

## Explanation

`du -b` is the culprit here. `man du` provides this information: 
```
       --apparent-size
              print  apparent sizes, rather than disk usage; although the apparent size is usually smaller, it may be larger due to holes in ('sparse') files, internal fragmen‐
              tation, indirect blocks, and the like

       -b, --bytes
              equivalent to '--apparent-size --block-size=1'
```

so the `-b` actually prints the **apparent sizes**, rather than disk usage. In fact this provides the real disk usages:
```
du -c --block-size=1 data/queue/main/page.*
67108864	data/queue/main/page.0
67108864	data/queue/main/page.1
67108864	data/queue/main/page.2
67108864	data/queue/main/page.3
4096	data/queue/main/page.4
268439552	total
```

So **now** the difference between what `du` reports `268439552` and the api `268435725` is 3827 bytes.

This difference seems related to the FS block size of 4K by default, which mean that for `page.4` in particular, which is the head page containing a single element per reaching `max_queue_size_in_bytes`, that document is probably only a few hundreds bytes but the smaller reported size is 4k per the FS.

So I do not believe we have  a problem. LMKWYT.

colinsurprenant:
I'll go ahead and squash now, I believe we are getting close to merging :)

colinsurprenant:
@jordansissel per our discussion are we LGTM'ed ?

jordansissel:
Ooops I had LGTM'd in my head but not here.

LGTM.

colinsurprenant:
thanks @jordansissel  :thumbsup:

colinsurprenant:
merged into master and 6.x - forgot to add the PR number in the commit and the 6.x commit is 2f37e5c1eb9dd630cf01c2e1f869fa21ff120127

colinsurprenant:
followed up in #9038 for the performance regression this PR introduced.

PR #8961, opened 2018-01-16
test fix: don't use stdin for blocking pipeline

jakelandis:
This test really doesn't care which input is used, as long it blocks long enough to check the pipeline name... which for some reason (not totally understood) stdin doesn't block the child LS process as expected [1]. 

Change is trivial and should fix the intermittent test failures. 

[1] 
There is special logic w/r/t to using the stdin that i suspect helps to keep it from dying immediately after starting and also manipulates the io for the child process. That special logic doesn't accept settings as a parameter, so I can't use it without unecessary changes.

### DIFF
--- a/qa/integration/specs/secret_store_spec.rb
+++ b/qa/integration/specs/secret_store_spec.rb
@@ -53,7 +53,7 @@ describe "Test that Logstash" do
   it "expands secret store variables from settings" do
     test_env["LOGSTASH_KEYSTORE_PASS"] = "keystore_pa9454w3rd"
     @logstash.env_variables = test_env
-    @logstash.spawn_logstash("-e", "input {stdin {}} output { }", "--path.settings", settings_dir)
+    @logstash.spawn_logstash("-e", "input {heartbeat {}} output { }", "--path.settings", settings_dir)
     @logstash.wait_for_logstash
     Stud.try(num_retries.times, [StandardError, RSpec::Expectations::ExpectationNotMetError]) do
       result = @logstash.monitoring_api.node_stats rescue nil


original-brownbear:
@jakelandis nice find, we should fix the other tests that do this as well at some point :)
Looks like we got quite a few:

```sh
specs/monitoring_api_spec.rb
25:    logstash_service.start_with_stdin
27:    number_of_events.times { logstash_service.write_to_stdin("Hello world") }
39:    logstash_service.start_with_stdin
53:    logstash_service.start_with_stdin
72:    logstash_service.start_with_stdin
102:    logstash_service.start_with_stdin

specs/01_logstash_bin_smoke_spec.rb
43:    @ls1.start_with_stdin

fixtures/es_output_how_spec.yml
7:    stdin { }

services/logstash_service.rb
19:  STDIN_CONFIG = "input {stdin {}} output { }"
73:  # Given an input this pipes it to LS. Expects a stdin input in LS
88:  # Can start LS in stdin and can send messages to stdin
90:  def start_with_stdin
91:    puts "Starting Logstash #{@logstash_bin} -e #{STDIN_CONFIG}"
95:      @process = build_child_process("-e", STDIN_CONFIG)
105:  def write_to_stdin(input)
107:      @process.io.stdin.puts(input)
140:      @process.io.stdin.close rescue nil
```

=> LGTM :)

jakelandis:
@original-brownbear - thanks 

`.start_with_stdin` seems to work fine.  I believe it has to do with the extra io manipulation in that function. However, I didn't spend any time digging in to why, only observed that stdin without that method gets shutdown without any error messages shortly after startup.

PR #8964, opened 2018-01-16
MINOR: Remove unnecessary logger fields from output delegator strategy and plugin factory

original-brownbear:
Same as #8957, `@logger` field isn't used in any of these classes and just needlessly complicates things.


original-brownbear:
@jsvd thanks :)

PR #8969, opened 2018-01-17
JAVAFICATION: Move remaining Logstash::Timestamp Ruby Code to Java

original-brownbear:
Just resolving the todos here :)

Notes:
* Had to keep the `timestamp.rb` file, otherwise at least the date filter breaks
* I know there is an inconsistency between the `+` and `-` methods, I simply ported this 100% and it should probably be fixed. Didn't want to change behavior in this step though so I left it as is or easier reviewing :)
* Other than that the approach is pretty standard for JRuby I think, including `Comparable` doesn't actually inherit any methods => had to code the comparisons manually and did so exactly the same way it's done in `RubyString`

### DIFF
--- a/logstash-core/lib/logstash/event.rb
+++ b/logstash-core/lib/logstash/event.rb
@@ -2,7 +2,6 @@
 
 require "logstash/namespace"
 require "logstash/json"
-require "logstash/timestamp"
 
 # transient pipeline events for normal in-flow signaling as opposed to
 # flow altering exceptions. for now having base classes is adequate and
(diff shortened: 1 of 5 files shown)


jsvd:
What if we just make this for 7.0 and break the behaviour?

original-brownbear:
@jsvd we could I guess :) I'm not opposed to doing that, though it does introduce some divergence between `6.x` and `7.x`. Maybe merge this to both, then break behaviour in another PR just in `master`?

jsvd:
+1 on merging as is and discussing the breaking change separately

original-brownbear:
@jsvd thanks :)

PR #8970, opened 2018-01-17
JAVAFICATION: Remove Ruby method_missing deprecation on Event

original-brownbear:
We're going for `7.0`, no need to keep this deprecation logic for `5.x` around any longer, especially when it taints the `Event` class that is otherwise set up in Java right?

### DIFF
--- a/logstash-core/lib/logstash/event.rb
+++ b/logstash-core/lib/logstash/event.rb
@@ -33,21 +33,4 @@ module LogStash
   FLUSH = FlushEvent.new
   SHUTDOWN = ShutdownEvent.new
   NO_SIGNAL = NoSignal.new
-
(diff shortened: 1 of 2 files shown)


original-brownbear:
@jsvd @andrewvc what do you guys think here? We should be able to just drop this code now right?

jsvd:
+1 on removing for 7.0
I see this tagged as 6.2.0, I'd not merge this into a 6.x branch

original-brownbear:
@jsvd makes sense untagged `6.x` :)

original-brownbear:
@jsvd I think just for `7.x` it's uncontroversial enough :) => Merging (test failure is known slowlog IT failure).

PR #8972, opened 2018-01-17
6.2.lockupdate

jakelandis:
Introduce the Gemfile lock on the (new) 6.2 branch.  Diff the to commits in this PR to see the difference since 6.1 : https://github.com/elastic/logstash/pull/8972/commits/5f8fb6b004bd639856e5f1a0bf19aee03365b170

### DIFF
--- a/Gemfile.jruby-2.3.lock.release
+++ b/Gemfile.jruby-2.3.lock.release
@@ -1,0 +1,714 @@
+PATH
+  remote: ./logstash-core
+  specs:
+    logstash-core (6.1.2-java)
+      chronic_duration (= 0.10.6)
+      clamp (~> 0.6.5)
+      concurrent-ruby (~> 1.0, >= 1.0.5)
+      elasticsearch (~> 5.0, >= 5.0.4)
+      filesize (= 0.0.4)
+      gems (~> 0.8.3)
+      i18n (= 0.6.9)
+      jar-dependencies
+      jrjackson (~> 0.4.4)
+      jruby-openssl (>= 0.9.20)
+      manticore (>= 0.5.4, < 1.0.0)
+      minitar (~> 0.6.1)
+      pry (~> 0.10.1)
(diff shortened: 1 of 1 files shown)


Gemfile.jruby-2.3.lock.release
```diff
@@ -0,0 +1,774 @@
...
+      sequel
+    logstash-filter-json (3.0.5)
+      logstash-core-plugin-api (>= 1.60, <= 2.99)
+    logstash-filter-kv (4.0.3)
+      logstash-core-plugin-api (>= 1.60, <= 2.99)
+    logstash-filter-metrics (4.0.5)
+      logstash-core-plugin-api (>= 1.60, <= 2.99)
+      metriks
+      thread_safe
+    logstash-filter-multiline (3.0.4)
```
jsvd: I think the lock file was created after running `rake test:install-core` before locking the dependencies. This installed a few gems used during testing which shouldn't be in the final lock file.
Whatever workflow we define as the one to lock the gems should depend on a `./gradlew clean` first.

jakelandis: ah thanks, was wondering why there was such a large change.  rebuilding now.

jakelandis:
@jsvd @guyboertje - lockfile updated to no longer include test gems.

jsvd:
It's really interesting that we no longer list jar-dependencies and ruby-maven. It seems we use the same jruby, what changed?

jakelandis:
> it's really interesting that we no longer list jar-dependencies and ruby-maven. It seems we use the same jruby, what changed?

@jsvd  - looks like the Java entry point removed them from logstash-core: https://github.com/elastic/logstash/commit/543b7221bd47afbde1f3c71059ff4be344990e80#diff-a70557553ce36f52ecddde61c83cb0f4 Note- they are still locked in the file, just not a dependency to logstash-core.

Issue #8974, opened 2018-01-17
Test failure: Test Logstash Slowlog should write logs to a new dir


colinsurprenant:
https://logstash-ci.elastic.co/job/elastic+logstash+master+multijob-integration-pq-2/258/console

```
15:09:37 Failures:
15:09:37 
15:09:37   1) Test Logstash Slowlog should write logs to a new dir
15:09:37      Failure/Error: expect(IO.read(slowlog_file).split("\n").size).to be >= 1
15:09:37 
15:09:37        expected: >= 1
15:09:37             got:    0
15:09:37      # ./specs/slowlog_spec.rb:44:in `block in (root)'
15:09:37      # /opt/logstash/build/qa/integration/gems/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in (root)'
15:09:37 
15:09:37 Finished in 9 minutes 55 seconds (files took 3.95 seconds to load)
15:09:37 21 examples, 1 failure
15:09:37 
15:09:37 Failed examples:
15:09:37 
15:09:37 rspec ./specs/slowlog_spec.rb:33 # Test Logstash Slowlog should write logs to a new dir
15:09:37 
15:09:37 Randomized with seed 21762
```

colinsurprenant:
I believe @jakelandis and @original-brownbear worked on this issue previously?

original-brownbear:
@jakelandis did

jakelandis:
Yeah, I only made assertion less strict since the test is only testing that something gets written, not the exact count: https://github.com/elastic/logstash/pull/8323/files 

@original-brownbear thought the root cause here is https://github.com/logstash-plugins/logstash-filter-sleep/issues/7 via https://github.com/elastic/logstash/pull/8323#issuecomment-330684586

original-brownbear:
@jakelandis yea I stand by that assertion I think :)

PR #8975, opened 2018-01-17
test fix: use single worker in test to avoid concurrency issue in sleep filter

jakelandis:
Per @original-brownbear [comment](https://github.com/elastic/logstash/issues/8073#issuecomment-330681296) setting the workers to 1 can fix this issue. ...so this PR sets the workers to 1.  Kinda an ugly 'fix' , but the test is valid independent of how many workers are configured. 

See: https://github.com/elastic/logstash/issues/8073
See: https://github.com/logstash-plugins/logstash-filter-sleep/issues/7

Fixes: #8974


original-brownbear:
@jakelandis LGTM :)

jakelandis:
@original-brownbear - thanks

PR #8977, opened 2018-01-17
JAVAFICATION: Port wrapped write client to Java

original-brownbear:
Ported keeping functionality 100% intact. Only moved to top-level Ruby model for convenience in the Java class setup.
Simplified the constructor here slightly too:
* We know the type is always `:input` => only need the plugin id, not the plugin => only pass the plugin_id in the constructor
* We only use the pipeline id, nothing else from the pipeline => only pass in the pipeline id

I know that the way the metrics `increment` calls are made is not 100% optimal, but:
* Caching the `CallSite` for those would make the code much more complicated
* This change already measurably ups performance on the input side (a single generator thread goes from 350k/s to 390k/s on my box)

=> I'd not optimize this away since we'll get the Java metrics eventually and the hacks in here will just go away then anyways.


original-brownbear:
@danhermann can you review this? Maybe you'll find some of my metric hacks useful for your work too :)

original-brownbear:
Jenkins test this please

danhermann:
LGTM!

original-brownbear:
@danhermann nice thanks!

PR #8984, opened 2018-01-18
fix integration tests build

colinsurprenant:
This is reboot of #8914 and it introduces the following changes:
- Allow overriding the `GEM_HOME` and `GEM_PATH` in `logstash.lib.sh`. This is useful when launching the integration tests through `bin/ruby` to make sure `GEM_HOME` and `GEM_PATH` will not be inherited by any local shell initialization.
- Adds a few missing clean paths
- Directly invoke `gem` and `bundler` in the integration tests setups.
- Invoke the integration specs through `bin/ruby` and not `vendor/jruby/bin/jruby` to avoid any `GEM_HOME` and `GEM_PATH` problem

### DIFF
--- a/bin/logstash.lib.sh
+++ b/bin/logstash.lib.sh
@@ -1,3 +1,21 @@
+# This script is used to initialize a number of env variables and setup the
+# runtime environment of logstash. It sets to following env variables:
+#   LOGSTASH_HOME & LS_HOME
+#   SINCEDB_DIR
+#   JAVACMD
+#   JAVA_OPTS
+#   GEM_HOME & GEM_PATH
+#   DEBUG
+#
+# These functions are provided for the calling script:
+#   setup() to setup the environment
+#   ruby_exec() to execute a ruby script with using the setup runtime environment
+#
+# The following env var will be used by this script if set:
+#   LS_GEM_HOME and LS_GEM_PATH to overwrite the path assigned to GEM_HOME and GEM_PATH
+#   LS_JAVA_OPTS to append extra options to the JVM options provided by logstash
+#   JAVA_HOME to point to the java home
+
 unset CDPATH
 # This unwieldy bit of scripting is to try to catch instances where Logstash
 # was launched from a symlink, rather than a full path to the Logstash binary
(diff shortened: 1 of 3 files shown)


yaauie:
With a clean checkout of master plus this patch, when I invoke `ci/integration_tests.sh`, I'm able to run the integration tests (e.g., it bootstraps, and the tests have been making progress on my local box for the last ~15 minutes).

I have not validated the docker integration tests, as it is not immediately clear whether this fix is relevant to them.

colinsurprenant:
@yaauie Thanks for the review.

In #8914 @jordansissel suggested we document `LS_GEM_HOME` and `LS_GEM_PATH` since these are new env vars used by `logstash.lib.sh` - I am not sure we should since these are really meant for internal use and probably temporary until we either refactor the integrations tests or find a better launching solution. WDYT?

yaauie:
@colinsurprenant `LS_GEM_HOME` and `LS_GEM_PATH` are read _by_ `logstash.lib.sh` to control behaviour, and are therefore part of its interface; I agree with @jordansissel that they should be documented (even if they're explicitly documented as internal details of our gradle launcher).

Looking at this patch again in that light, I see that when  `GEM_PATH` and/or `GEM_HOME` are defined, `logstash.lib.sh` silently stomps on existing values; should we warn here? is there a sensible way to just use `GEM_PATH` and `GEM_HOME` instead of our own namespaced versions of these, since we're simply using them to overwrite the real ones anyway?

colinsurprenant:
@yaauie ok, will document.

So the short story is that when a Gradle task of type exec is launched it will execute the shell initialization scripts and even if you set `GEM_HOME` and  `GEM_PATH` **in the Gradle task** these risks being overwritten by the shell initialization (for example if you have a Ruby version manager) (see [this discussion](https://github.com/elastic/logstash/pull/8914#issuecomment-355994512)). This is why for the integration tests launch, we have to indirect the lauch via `bin/ruby` which executes the `logstash.lib.sh` and sets `GEM_HOME` and  `GEM_PATH` or overrides them using `LS_GEM_HOME` and  `LS_GEM_PATH` if they are defined.  This is a nasty dance to prevent the overwriting of  `GEM_HOME` and  `GEM_PATH` when launching a Gradle exec task.

yaauie:
@colinsurprenant your description of the awkward dance makes sense 👍 

This will LGTM once it includes the docs we've both mentioned above 🙃

colinsurprenant:
added docs and rebased. will merge on green.

colinsurprenant:
jenkins please test this

colinsurprenant:
I don't understand these build failures.

colinsurprenant:
jenkins please test this

colinsurprenant:
seems like an infra problem.

colinsurprenant:
jenkins please test this

colinsurprenant:
squashed, merging in master and 6.x

colinsurprenant:
thanks for the review @yaauie

PR #8985, opened 2018-01-18
JAVAFICATION: More efficient use of Java counter metrics

original-brownbear:
* Extracts Java counters by unwrapping and downcasting Ruby Proxy for counter metrics
* Fallback to dummy counter in case of `NullMetric` (when metrics are deactivated)


danhermann:
LGTM!

PR #8988, opened 2018-01-19
JAVAFICATION: Move pipeline worker loop to Java

original-brownbear:
* Simply moved the logic of `worker_loop` into a `Runnable`.
* Passed down all `Pipeline` fields as `IRubyObject` that lacked a Java implementation
* Logic remains 100% unchanged
* Slight performance improvement despite using the shorter `callMethod` instead of `CallSite`caching 
   * `CallSite` caching isn't worth it here since Java implementations of all fields are incoming

### DIFF
--- a/logstash-core/src/main/java/org/logstash/execution/WorkerLoop.java
+++ b/logstash-core/src/main/java/org/logstash/execution/WorkerLoop.java
@@ -1,0 +1,87 @@
+package org.logstash.execution;
+
+import java.util.concurrent.BlockingQueue;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jruby.RubyArray;
+import org.jruby.runtime.ThreadContext;
+import org.jruby.runtime.builtin.IRubyObject;
+import org.logstash.RubyUtil;
+import org.logstash.config.ir.compiler.Dataset;
+
+public final class WorkerLoop implements Runnable {
+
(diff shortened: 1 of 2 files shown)


logstash-core/src/main/java/org/logstash/execution/WorkerLoop.java
```diff
@@ -0,0 +1,85 @@
...
+                );
+                final IRubyObject filteredSize = batch.callMethod(context, "filtered_size");
+                readClient.callMethod(context, "add_output_metrics", filteredSize);
+                readClient.callMethod(context, "add_filtered_metrics", filteredSize);
+                readClient.callMethod(context, "close_batch", batch);
+                if (isFlush) {
+                    flushing.callMethod(context, "set", context.fals);
+                }
+            } while (!shutdownRequested && !isDraining(context));
+            final IRubyObject batch = readClient.callMethod(context, "new_batch");
```
danhermann: Minor suggestion -- might be helpful to retain part of the original comment that the duplication of these lines from the body of the loop is necessary to flush the last batch at shutdown.

danhermann:
LGTM!

logstash-core/src/main/java/org/logstash/execution/WorkerLoop.java
original-brownbear: @danhermann sure let's clean this up :)

original-brownbear:
@danhermann thanks, comment is back => will merge if green after rebase :)

PR #8989, opened 2018-01-19
Syntax error causing abort on startup

joaomarccos:
The syntax of command set is not correct and this causes a abort with no error logs when a user have the JAVA_OPTS env variable defined on your system.

### DIFF
--- a/bin/setup.bat
+++ b/bin/setup.bat
@@ -40,8 +40,9 @@ if not "%JAVA_TOOL_OPTIONS%" == "" (
 rem JAVA_OPTS is not a built-in JVM mechanism but some people think it is so we
 rem warn them that we are not observing the value of %JAVA_OPTS%
 if not "%JAVA_OPTS%" == "" (
-  echo|set /p="warning: ignoring JAVA_OPTS=%JAVA_OPTS%; "
+  echo warning: ignoring JAVA_OPTS=%JAVA_OPTS%;
   echo pass JVM parameters via LS_JAVA_OPTS
+  set JAVA_OPTS=
 )
 
 rem ### 3: set jruby


bin/setup.bat
```diff
@@ -40,8 +40,9 @@ if not "%JAVA_TOOL_OPTIONS%" == "" (
 rem JAVA_OPTS is not a built-in JVM mechanism but some people think it is so we
 rem warn them that we are not observing the value of %JAVA_OPTS%
 if not "%JAVA_OPTS%" == "" (
-  echo|set /p="warning: ignoring JAVA_OPTS=%JAVA_OPTS%; "
+  echo warning: ignoring JAVA_OPTS=%JAVA_OPTS%;
   echo pass JVM parameters via LS_JAVA_OPTS
+  set JAVA_OPTS=
```
yaauie: we shouldn't need to also zero-out the `JAVA_OPTS` variable; it just isn't used anywhere downstream.

bin/setup.bat
```diff
@@ -40,8 +40,9 @@ if not "%JAVA_TOOL_OPTIONS%" == "" (
 rem JAVA_OPTS is not a built-in JVM mechanism but some people think it is so we
 rem warn them that we are not observing the value of %JAVA_OPTS%
 if not "%JAVA_OPTS%" == "" (
-  echo|set /p="warning: ignoring JAVA_OPTS=%JAVA_OPTS%; "
```
robbavey: I think the original intent behind the unusual syntax here was to preserve [not adding a newline](https://github.com/elastic/logstash/blob/master/bin/logstash.lib.sh#L80). By adding a dummy variable [as shown here](https://stackoverflow.com/questions/7105433/windows-batch-echo-without-new-line/11336754#11336754), it should be possible to keep the formatting consistent.

colinsurprenant: yeah, the intent was to display the warning on a single line so I suggest here 
```
if not "%JAVA_OPTS%" == "" (
  echo warning: ignoring JAVA_OPTS=%JAVA_OPTS%; pass JVM parameters via LS_JAVA_OPTS
)

```

bin/setup.bat
colinsurprenant: exactly, this is not required because in `logstash.bat` we do not take into account what is set in `JAVA_OPTS`

colinsurprenant:
+1 thanks @joaomarccos

The behaviour I see on Windows Server 2012 is this:
```
bin\logstash --version
warning: ignoring JAVA_OPTS=foo; pass JVM parameters via LS_JAVA_OPTS
```
and it silently dies. 

As suggested in the review comments if you modify the check to simply
```
if not "%JAVA_OPTS%" == "" (
  echo warning: ignoring JAVA_OPTS=%JAVA_OPTS%; pass JVM parameters via LS_JAVA_OPTS
)
```
Then this works for me:

```
bin\logstash --version                                                                                                     
warning: ignoring JAVA_OPTS=foo; pass JVM parameters via LS_JAVA_OPTS                                                        
Sending Logstash's logs to C:/Users/colin/dev/log stash/logs which is now configured via log4j2.properties                   
logstash 7.0.0-alpha1
...                                                                                                    
```

colinsurprenant:
@joaomarccos  let us know if you can proceed with the discussed changes in the near future - I'd like this to get this in sooner than later! thanks!

bin/setup.bat
joaomarccos: That's right, I think it's not necessary. I just wanted to do as same way is with JAVA_TOOL_OPTIONS

colinsurprenant: yeah, `JAVA_TOOL_OPTIONS` is simply unset because not reused after but `JAVA_OPTS` is completely overwritten later on in logstash.bat.

PR #8990, opened 2018-01-19
MINOR: Dry up Java LongCounter from Ruby Metric Hack

original-brownbear:
@danhermann I just realized this is a pretty big hack :D and we'll do the same in a few other places => dried this up by extracting it to the `LongCounter` calls + added some documentation that explains what Ruby classes are in play here :)


danhermann:
LGTM -- DRY++

PR #8991, opened 2018-01-19
JAVAFICATION: Port remainder of synchronous queue to Java

danhermann:
Includes `WrappedSynchronousQueue.WriteClient` and `WrappedSynchronousQueue`.


logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
```diff
@@ -1,16 +1,15 @@
...
+describe LogStash::WrappedSynchronousQueue do
 
-  subject {LogStash::Util::WrappedSynchronousQueue.new(5)}
+  subject {LogStash::WrappedSynchronousQueue.new(5)}
 
   describe "queue clients" do
     context "when requesting a write client" do
       it "returns a client" do
-        expect(subject.write_client).to be_a(LogStash::Util::WrappedSynchronousQueue::WriteClient)
+        expect(subject.write_client).to be_a(LogStash::MemoryWriteClient)
```
yaauie: when we change both specs and implementation in a single commit, it becomes hard to reason about what is changing (and to prove what is remaining the same).

Would it be possible to pull the java implementation into a matching namespace, so that we could prove we've got a 1:1 replacement?

logstash-core/src/main/java/org/logstash/ext/JrubyWrappedSynchronousQueueExt.java
```diff
@@ -0,0 +1,47 @@
...
+import org.jruby.anno.JRubyClass;
+import org.jruby.anno.JRubyMethod;
+import org.jruby.runtime.ThreadContext;
+import org.jruby.runtime.builtin.IRubyObject;
+
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+
+@JRubyClass(name = "WrappedSynchronousQueue")
+public class JrubyWrappedSynchronousQueueExt extends RubyObject {
```
yaauie: I know that these are following the existing `Jruby...Ext` pattern that already exists in the codebase, but I'm really not a fan of that naming convention -- it doesn't add much clarity (e.g., when I come across it, the `Jruby` bit _kindof_ hints that it's a `RubyObject` and has ruby-ish things about it, but otherwise doesn't tell me much).

logstash-core/src/main/java/org/logstash/ext/JrubyWrappedSynchronousQueueExt.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @JRubyMethod(name = "write_client")
+    public IRubyObject getWriteClient(final ThreadContext context) {
+        return JrubyMemoryWriteClientExt.create(queue);
+    }
+
+    @JRubyMethod(name = "read_client")
+    public IRubyObject getReadClient(final ThreadContext context) {
+        return JrubyMemoryReadClientExt.create(queue, 125, 50);
```
yaauie: These numbers come directly from the class you're replacing, but it would be helpful to add a comment indicating why they are what they are (even if doing so requires a little archaeology).

logstash-core/spec/logstash/util/wrapped_synchronous_queue_spec.rb
danhermann: I'm not sure I follow? The behavioral tests in the spec are all untouched and the new JRuby extension class passes the same tests as the original JRuby class.

logstash-core/src/main/java/org/logstash/ext/JrubyWrappedSynchronousQueueExt.java
danhermann: I thought that naming convention was reasonable since these are JRuby extension classes. Do you have a suggestion for a better naming convention?

As a related aside, I expect there will be a growing number of JRuby extension classes in the earlier phases of the processing pipeline's move to Java. Once the pipeline is more Java than Ruby and the interface point between Ruby and Java can be minimized, I expect a number of these classes will be converted to pure Java to eliminate the unnecessary dependency on JRuby.

logstash-core/src/main/java/org/logstash/ext/JrubyWrappedSynchronousQueueExt.java
danhermann: I found #8707 in which those values were introduced and added a comment to that effect.

original-brownbear:
@danhermann squash + merge imo I'm happy here :)
@yaauie I agree with Dan here, the only change to the tests are the name of the class tested (which would be somewhat annoying to nest into the sync queue when setting it up in Java :))

original-brownbear:
LGTM

original-brownbear:
@danhermann I squash+merged to unblock this work on my end :) => all done here, thanks! :)

jakelandis:
@danhermann @original-brownbear - Did you mean to tag this with 6.3.0 ? I don't think this in the 6.x pipeline, and this commit removes the ability to configure read client batch size and timeout.

original-brownbear:
@jakelandis 

How does this commit kill the configurability? That would be an outright bug, but it's configurable just fine via https://github.com/elastic/logstash/pull/8991/files#diff-d53f0a87ca0bf68975e2708d82de4923R67 isn't it?

logstash-core/src/main/java/org/logstash/ext/JrubyWrappedSynchronousQueueExt.java
```diff
@@ -0,0 +1,49 @@
...
+    @JRubyMethod(name = "write_client")
+    public IRubyObject getWriteClient(final ThreadContext context) {
+        return JrubyMemoryWriteClientExt.create(queue);
+    }
+
+    @JRubyMethod(name = "read_client")
+    public IRubyObject getReadClient(final ThreadContext context) {
+        // batch size and timeout are currently hard-coded to 125 and 50ms as values observed
+        // to be reasonable tradeoffs between latency and throughput per PR #8707
+        return JrubyMemoryReadClientExt.create(queue, 125, 50);
```
jakelandis: @danhermann ^^ here

original-brownbear: @danhermann yikes missed that one, we need to fix this by passing down the actual setting values here I'm afraid.

danhermann: @jakelandis, while the `MemoryReadClient` exposes a configurable batch size and read timeout, it was never called in 6.0 with anything other than the default settings of 125 and 50, respectively (see link below). The port to Java retains that same behavior.

https://github.com/elastic/logstash/blob/6.0/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb#L30

original-brownbear: @danhermann there was a `setter` for these settings that was called when setting the worker threads (you are right the constructor always used the defaults and was super redundant). I'm afraid we had some merge conflict at some point and lost the change that fixed the invocation of the constructor and simply kept the removal of that setter.
=> the problem isn't really in this PR but happened earlier but we should still fix this :)

original-brownbear: @jakelandis actually nevermind this was my mistake, we still have https://github.com/elastic/logstash/blob/master/logstash-core/src/main/java/org/logstash/execution/QueueReadClientBase.java#L70 and that is called from `java_pipeline.rb` so we're good here imo unless it's broken for you in practice?

jakelandis: @danhermann @original-brownbear I just saw this while looking at something unrelated, and it seemed off, I didn't run any tests. Apologies if I misunderstood the change.

PR #8999, opened 2018-01-22
Remove serialized memory queue

original-brownbear:
This is mainly to simplify the move to a Java interface shared by the acked and normal queue:
 * Drops the in memory, serialized queue
* Moves all tests to the normal PQ path
   * See Jenkins, test runtime remains almost constant. Admittedly though the random ack test was shortened to a smaller page count, which should be fine. It's just a stress test anyways and we have a bunch of them for the PQ that are all somewhat redundant.

=> Almost 1k lines saved
=> No need to create another layer of indirection to accommodate the two different acked queues when moving things to Java fully (moving the acked Queue to Java is the last blocker from completely removing JRuby in the main datapath outside the plugins!).
=> Better test coverage of the actual PQ

I'm aware that the code in the PQ path could be simplified a lot more as a result of this change but chose not to do it here to keep things readable. Next steps here would be:
* Remove `PageIOFactory` from everywhere
* Remove redundant single inheritance for JRuby wrapper here: i.e. remove `AbstractAckedQueue` in favor of just keeping `AckedQueue`


andrewvc:
+1 to this idea. Since we don't use this queue, and have no plans to, it makes sense to remove it.

original-brownbear:
@yaauie can you give this a look :)? I'm happy to provide more background here if needed.

yaauie:
I'll dive in on this today.

logstash-core/lib/logstash/queue_factory.rb
```diff
@@ -32,7 +28,7 @@ def self.create(settings)
           settings.get("pipeline.batch.size") * settings.get("pipeline.workers") * 2
         )
       else
-        raise ConfigurationError, "Invalid setting `#{queue_type}` for `queue.type`, supported types are: 'memory_acked', 'memory', 'persisted'"
+        raise ConfigurationError, "Invalid setting `#{queue_type}` for `queue.type`, supported types are: 'memory' or 'persisted'"
```
yaauie: Should we register this message with i18n while we're at it?

original-brownbear: @yaauie not right now while this is in flux imo (especially since i18n == Ruby :))

original-brownbear:
@yaauie thanks + arg so sorry for just failing to squash :(

PR #9000, opened 2018-01-22
MINOR: Remove two dead methods from Acked queue

original-brownbear:
Just preparing the move to Java here, these methods are unused (they're not found on the synchronous queue impl. => they can't have any functional impact here either, also they have no point and just return `false` :P anyways)


jakelandis:
LGTM

PR #9002, opened 2018-01-22
MINOR: Some cleanups of Java in-memory Queue JRuby Ext classes

original-brownbear:
Keeping the entropy lower here :)

* The Ruby `initialize` constructors are already obsolete any never get called thanks to #8991 :)
* Made `final` what I could make `final` here just to make things a little more readable (especially since we do in fact have inheritance in the PQ path of this ... at least until after #8999 :))
* Removed one redundant reference to `RubyUtil.MEMORY_READ_BATCH_CLASS` (just as a side note we should keep references to `RubyUtil` to a minimum wherever possible since we will need to clean up the global `Ruby` reference eventually)


original-brownbear:
@danhermann trivial and short review here if you get bored today :D

danhermann:
LGTM -- thanks for cleaning that up.

original-brownbear:
@danhermann thanks for the review :)

PR #9003, opened 2018-01-22
MINOR: Align behaviour of acked queue Ruby wrappers with behaviour of in memory queue wrappers

original-brownbear:
Aligning behaviour between acked and in memory queue some more before the Java port:

* `poll` for just a single element isn't exposed for the in memory queue => don't expose it here either
* In memory queue reads directly when constructing the batch => do the same here (also gives better performance)
   * `read_next` is simply redundant since it's always called after instantiating the batch in normal operation (can just instantiate a batch of size `0` for the final flush)
* Make `acked_batch.get_elements` return a `Hash` since it's only used in a single place and that place wants a `Hash` => no need for the redundant in between array

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -55,12 +55,6 @@ module LogStash; module Util
     end
     alias_method(:<<, :push)
 
-    # Block for X millis
-    def poll(millis)
-      check_closed("read")
-      @queue.read_batch(1, millis).get_elements.first
-    end
-
     def read_batch(size, wait)
       check_closed("read a batch")
       @queue.read_batch(size, wait)
@@ -141,7 +135,7 @@ module LogStash; module Util
       # create a new empty batch
       # @return [ReadBatch] a new empty read batch
       def new_batch
-        ReadBatch.new(@queue, @batch_size, @wait_for)
+        ReadBatch.new(@queue, 0, 0)
       end
 
       def read_batch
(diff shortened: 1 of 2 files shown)


logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -191,23 +184,9 @@ def add_output_metrics(filtered_size)
...
         @generated = Hash.new
-        @acked_batch = nil
-      end
-
-      def read_next
-        @acked_batch = @queue.read_batch(@size, @wait)
-        return if @acked_batch.nil?
-        @acked_batch.get_elements.each { |e| @originals[e] = true }
+        @acked_batch = queue.read_batch(size, wait)
+        @originals = @acked_batch.nil? ? Hash.new : @acked_batch.get_elements
```
yaauie: `@originals` and `@generated` are both `Hash{Object=>TrueClass}`, which are only used for inclusion detection and iteration; functionally they're using a custom `Set` implementation;

Instead of changing `RubyAckedBatch#get_elements` to leak this particular detail, should we instead make it return a `Set` (either a Java `Set` or a `RubySet`) and refactor this class' internals to use the appropriate methods for inclusion detection and iteration?

original-brownbear: @yaauie there isn't really a fast `RubySet` and a Java Set (also note that we need an ordered set here) will be very slow here until we do the next steps (because of JRuby wrapping it in a Java Proxy).
=> for now, this is the best we can do here I fear, but "now" won't last long I promise :) Hopefully, we'll get the next steps done here (including moving to Java collections) this week :)

original-brownbear:
@yaauie thanks!

PR #9005, opened 2018-01-22
MINOR: Add some context to unclean shutdown exceptions

original-brownbear:
Adding some context to rethrows on unclean shutdown to avoid the hard to interpret last line:

```sh
[2018-01-22T17:12:25,958][ERROR][org.logstash.Logstash    ] java.lang.IllegalStateException: org.jruby.exceptions.RaiseException: (SystemExit) exit
```

that currently gets logged when LS stops because of an exception in the main loop.


original-brownbear:
@jakelandis take a look maybe since we just talked about this :)

jakelandis:
LGTM

original-brownbear:
thanks :)

Issue #8735, opened 2017-11-27
ensure secret store has stronger permissions restrictions


jordansissel:
This issue is about a feature that has not landed, but the permissions discussion was deferred, so in order to not forget about this, I am filing it here.

From: https://github.com/elastic/logstash/pull/8659#discussion_r152055542

Currently using 0660, we should probably use 0600 or at least 0640. We have no specification of behavior for the keystore, so it's unclear why 0660 is a good choice, so I am lobbying to align with the behavior of other safety-conscious tools (ruby netrc, ssh-keygen, etc, will default to 0600 and refuse to work if other permissions are used).

jakelandis:
> so it's unclear why 0660 is a good choice

[This](https://github.com/elastic/elasticsearch/commit/7a035f5f8435962072157f46031e73855fe4a875) is why Elasticsearch went with 660, I will test our package based installation and post back here with the results.

jakelandis:
@jordansissel - I made a PR to change the permissions to `644` https://github.com/elastic/logstash/pull/9006.  This is actually counter to the original intent of this issue, but given our current defaults this is what is needed to make this work out of the box. 

I believe that saving grace here for expanding the to global read is the ability to set the password that is only accessible via root access. For example, using RHEL 6.x as test base (and assume similar will also apply to other distro's) this is how we can set/view the password environment variable only with root access.

```
set +o history
export LOGSTASH_KEYSTORE_PASS=password123
sudo -E /usr/share/logstash/bin/logstash-keystore --path.settings /etc/logstash create
sudo -E /usr/share/logstash/bin/logstash-keystore --path.settings /etc/logstash add foo
sudo -E /usr/share/logstash/bin/logstash-keystore --path.settings /etc/logstash list

echo 'LOGSTASH_KEYSTORE_PASS=password123' | sudo tee --append /etc/logstash/startup.options > /dev/null
sudo /usr/share/logstash/bin/system-install
```

Then in another PR (or manual instructions), I will need to ensure that  `/etc/default/logstash` is owned by root with `600` permissions.

jakelandis:
Maybe the instructions should be to set `/etc/sysconfig/logstash` instead messing with the startup options and `system-install`.

`echo 'LOGSTASH_KEYSTORE_PASS=password123' | sudo tee --append /etc/sysconfig/logstash > /dev/null` 
and then manually change the permissions to `600`
`sudo chmod 600 /etc/sysconfig/logstash`

[...]

PR #9006, opened 2018-01-22
Keystore: Change default permission to 644

jakelandis:
When Logstash is package installed (rpm/deb), the cli tool will be installed with root/root, but the service is run as logstash/logstash.  Changing to allow global read access to allow the logstash user read the keystore. Documentation will be provided for how to setup the keystore password such that you will need root access to obtain the keystore password, thereby requiring root access to do anything with the keystore as long as the password is set.

Part of #8735 

(this will be merged to 6.2 and newer)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
+++ b/logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
@@ -50,7 +50,7 @@ public final class JavaKeyStore implements SecretStore {
     private boolean useDefaultPass = false;
     private Lock writeLock;
     //package private for testing
-    static String filePermissions = "rw-rw----";
+    static String filePermissions = "rw-r--r--";
     private static final boolean IS_WINDOWS = System.getProperty("os.name").startsWith("Windows");
 
     /**
--- a/logstash-core/src/test/java/org/logstash/secret/store/backend/JavaKeyStoreTest.java
+++ b/logstash-core/src/test/java/org/logstash/secret/store/backend/JavaKeyStoreTest.java
@@ -332,7 +332,7 @@ public class JavaKeyStoreTest {
         // if we got attributes, lets assert them.
         if (attrs != null) {
             Set<PosixFilePermission> permissions = attrs.readAttributes().permissions();
-            EnumSet<PosixFilePermission> expected = EnumSet.of(OWNER_READ, OWNER_WRITE, GROUP_READ, GROUP_WRITE);
+            EnumSet<PosixFilePermission> expected = EnumSet.of(OWNER_READ, OWNER_WRITE, GROUP_READ, OTHERS_READ);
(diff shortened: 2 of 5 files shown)


jakelandis:
Added second commit to ensure that environment variables used by the service are protected by root access.

jordansissel:
Testing:

Without this patch:

```
[2018-01-13T18:06:47,548][ERROR][logstash.agent           ] Failed to execute action {:action=>LogStash::PipelineAction::Create/pipeline_id:main, :exception=>"Java::OrgLogstashSecretStore::SecretStoreException::UnknownException", <long stack trace omitted>
[2018-01-13T18:06:47,551][ERROR][logstash.agent           ] An exception happened when converging configuration {:exception=>LogStash::Error, :message=>"Don't know how to handle `Java::OrgLogstashSecretStore::SecretStoreException::UnknownException` for `LogStash::PipelineAction::Create/pipeline_id:main`"
```

Tests post-patch I will write after dinner.

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
```diff
@@ -50,7 +50,7 @@
     private boolean useDefaultPass = false;
     private Lock writeLock;
     //package private for testing
-    static String filePermissions = "rw-rw----";
+    static String filePermissions = "rw-r--r--";
```
jordansissel: (Comment, no change required) I might recommend simply 640 and having the file owned by `logstash` group, but this also works given the file is (intended to be) protected by a passphrase.

jordansissel:
I tested the RPM and verified the file permissions:

* -rw-------. 1 root root 292 Jan 13 22:00 /etc/default/logstash

```
[root@f7e9c0c48566 build]# ls -l /etc/logstash
total 28
drwxrwxr-x. 2 root root    6 Jan 13 18:13 conf.d
-rw-rw-r--. 1 root root 1738 Jan 13 18:13 jvm.options
-rw-rw-r--. 1 root root 1334 Jan 13 18:13 log4j2.properties
-rw-r--r--. 1 root root  605 Jan 13 22:19 logstash.keystore
-rw-rw-r--. 1 root root 6378 Jan 13 22:00 logstash.yml
-rw-rw-r--. 1 root root  285 Jan 13 18:13 pipelines.yml
-rw-------. 1 root root 1659 Jan 13 18:13 startup.options
```

logstash-core/src/main/java/org/logstash/secret/store/backend/JavaKeyStore.java
jakelandis: I am a bit concerned at this level to make assumptions about the users on the local OS and the user that is running logstash.  With the package based installed we can make those assumptions, but for other types of installation that may be bad assumptions. 

This is a good point however, and look into incorporating this suggesting into the doc.

PR #9010, opened 2018-01-23
MINOR: Cleanup needless single inheritance in JRubyQueueExt

original-brownbear:
Due to #8999 we only have `AckedQueue` left => no need for the redundant abstract queue class since we only have one implementation left => flattened things out :)


jakelandis:
LGTM

original-brownbear:
@jakelandis thanks!

PR #9011, opened 2018-01-23
CLEANUP: Flatten AbstractByteBufferPageIO + MMapPageIO

original-brownbear:
Due to #8999 we only have one child class of `AbstractByteBufferPageIO` => we can flatten things out here and save a lot of complication :)

* no changes here other than moving all members of `AbstractByteBufferPageIO` into `MMapPageIO` (inlining `super` calls where necessary) :)


original-brownbear:
@yaauie I think this is one for you :)

original-brownbear:
@yaauie thanks!

PR #9012, opened 2018-01-23
CLEANUP: Remove redundant PageIOFactory

original-brownbear:
`org.logstash.ackedqueue.io.MmapPageIO` is the only remaining `PageIO` implementation used by the `Queue` => no need for the factory abstraction, we can just use outright `new MmapPageIO(...)` instead throughout and save a lot of complication.

* Removed factory
* Simplified the `Settings`


original-brownbear:
Jenkins test this

original-brownbear:
@yaauie thanks! fixed the conflicts (the location of the constructor in the QueueExt changed due to #9010 )

PR #9013, opened 2018-01-23
JAVAFICATION: Java write client for acked queue

original-brownbear:
Not much to say here, 1:1 port of the Ruby code:

changes:
* Used Java atomic boolean for the closed state to avoid hacky Ruby calls for that


original-brownbear:
@danhermann take a look when you have a sec, but no rush :)

danhermann:
LGTM!

original-brownbear:
@danhermann thanks!

PR #9016, opened 2018-01-23
JAVAFICATION: Move Flusher Thread to Java

original-brownbear:
Trivial but it allows for one less roundtrip to Ruby for the `flushing` state and shortens the `java_pipeline.rb` which is always nice :)

### DIFF
--- a/logstash-core/src/main/java/org/logstash/execution/PeriodicFlush.java
+++ b/logstash-core/src/main/java/org/logstash/execution/PeriodicFlush.java
@@ -1,0 +1,53 @@
+package org.logstash.execution;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jruby.runtime.builtin.IRubyObject;
+
+public final class PeriodicFlush implements AutoCloseable {
+
+    private static final Logger LOGGER = LogManager.getLogger(PeriodicFlush.class);
+
+    private final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor(
+        r -> new Thread(r, "logstash-pipeline-flush")
+    );
+
+    private final BlockingQueue<IRubyObject> queue;
+
+    private final IRubyObject signal;
+
+    private final AtomicBoolean flushing;
+
+    public PeriodicFlush(final BlockingQueue<IRubyObject> queue, final IRubyObject signal,
+        final AtomicBoolean flushing) {
+        this.queue = queue;
+        this.signal = signal;
+        this.flushing = flushing;
+    }
+
+    public void start() {
+        executor.scheduleAtFixedRate(() -> {
+            if (flushing.compareAndSet(false, true)) {
+                LOGGER.debug("Pushing flush onto pipeline.");
+                try {
+                    queue.put(signal);
+                } catch (final InterruptedException ex) {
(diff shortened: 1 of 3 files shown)


logstash-core/src/main/java/org/logstash/execution/PeriodicFlush.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jruby.runtime.builtin.IRubyObject;
+
+public final class PeriodicFlush implements AutoCloseable {
+
+    private static final Logger LOGGER = LogManager.getLogger(PeriodicFlush.class);
+
+    private final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
```
jakelandis: Can you name this thread ? (trying to get all threads named to make debugging easier)

logstash-core/src/main/java/org/logstash/execution/PeriodicFlush.java
```diff
@@ -0,0 +1,51 @@
...
+                }
+            }
+        }, 0L, 5L, TimeUnit.SECONDS);
+    }
+
+    @Override
+    public void close() throws InterruptedException {
+        executor.shutdown();
+        if (!executor.awaitTermination(10L, TimeUnit.SECONDS)) {
+            throw new IllegalStateException("Failed to stop period flush action.");
```
jakelandis: Are you sure you want to throw here ?  It may cause other shutdown type behavior to not fire.  Usually after the wait shutdownNow() is called sometimes with a logger message.

logstash-core/src/main/java/org/logstash/execution/PeriodicFlush.java
```diff
@@ -0,0 +1,51 @@
...
+                    queue.put(signal);
+                } catch (final InterruptedException ex) {
+                    throw new IllegalStateException(ex);
+                }
+            }
+        }, 0L, 5L, TimeUnit.SECONDS);
+    }
+
+    @Override
+    public void close() throws InterruptedException {
```
jakelandis: nit pick / style: As a consumer I wouldn't typically expect checked exceptions on a close() method.

logstash-core/src/main/java/org/logstash/execution/PeriodicFlush.java
original-brownbear: @jakelandis yea, look the flusher does a non-blocking, short operation periodically. If we get into a spot where we break other shutdown behaviour here it's a clear bug. If we add `shutdownNow`, all we do is hide that bug.

logstash-core/src/main/java/org/logstash/execution/PeriodicFlush.java
original-brownbear: @jakelandis well it's an `AutoCloseable` you should expect all bugs here :) I'd rather keep this for the next steps. If I now rethrow a `RuntimeException` of some kind here, I'm just adding more logical steps that I have to unwrap when I move the call to this into Java. => I'd rather keep this simple tbh

logstash-core/src/main/java/org/logstash/execution/PeriodicFlush.java
original-brownbear: @jakelandis sure sec

original-brownbear:
@jakelandis naming added, I would keep the other 2 spots as is for now, this will be called by Java code in a day or two anyway, then we can look the interface some more.
Also the existing version in Ruby didn't have any hard interrupts of the thread either so we should be good here :)

jakelandis:
@original-brownbear - thanks for the thread name.  LGTM

PR #9018, opened 2018-01-23
Fix links

dedemorton:
@andrewvc When we flipped the versions for 6.2, we had some broken links because [this commit ](https://github.com/elastic/logstash/commit/a5462f516b9a6d97e75cab130814edb2114f8b06#diff-27ef4dc573a26a68662c5664a6706d7aL33) was never made to 6.2 (and, I'm assuming, 6.x).

I've fixed the link related issues, but I don't want to make changes to the breaking changes because I assume there have been other changes since you made your commit. Can you take a look and let me know if there's anything else from [your commit](https://github.com/elastic/logstash/commit/a5462f516b9a6d97e75cab130814edb2114f8b06#diff-27ef4dc573a26a68662c5664a6706d7aL33) that needs to go into 6.2 and 6.x?

### DIFF
--- a/docs/index-shared1.asciidoc
+++ b/docs/index-shared1.asciidoc
@@ -37,45 +37,45 @@ volume and variety of data.
 
 // Introduction
 
-:edit_url: https://github.com/elastic/logstash/edit/master/docs/static/introduction.asciidoc
+:edit_url: https://github.com/elastic/logstash/edit/{branch}/docs/static/introduction.asciidoc
 include::static/introduction.asciidoc[]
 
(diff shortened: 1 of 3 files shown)


andrewvc:
@dedemorton this commit LGTM. I would say we should just manually reconcile the breaking changes in 6.1/6.x with 6.2 vs. trying to cherry pick on old commit.

dedemorton:
Cherry-picked int 6.x, too: https://github.com/elastic/logstash/pull/9019

PR #9023, opened 2018-01-24
BUG: Inflight batches for in memory queue reader fixed

original-brownbear:
Obvious bug, my bad :)

```java
return RubyHash.newHash(context.runtime, inflightBatches, RubyBasicObject.UNDEF);
```

This only works if the input Map (`inflightBatches`) is `<IRubyObject,IRubyObject>` which it isn't.
The auto conversion from Java to Ruby only happens for `putAll`:

* adjusted accordingly
* added tests


original-brownbear:
@danhermann maybe take a look at this when you have a sec (regardless of whether it was reviewed or not). I gave you bad/wrong advice there on the `RubyHash` when we looked into this the other day :(

original-brownbear:
@yaauie thanks :)

Issue #6779, opened 2017-03-01
add queue drain option specific tests/specs


colinsurprenant:
add better test harness around the new queue drain support.  relates to #6433

colinsurprenant:
Bump. See #7568 and #7570

colinsurprenant:
should be done with #7575 #7568 and #7583

colinsurprenant:
relates to #9024

PR #9024, opened 2018-01-24
BUG: Memory Queue needs to always be drained

original-brownbear:
Fixes the issue of finite pipeline runs not processing all events because they don't drain the in-memory queue.
This was introduced by `6.1` moving to an `ArrayBlockingQueue` instead of a `SynchronousQueue` introducing a capacity into the pipeline there.
=> trivial to fix by always draining the in-memory queue.
=> Also I found that the worker loop condition regarding the draining was logically wrong :( => fixed

--------------

* This needs a slightly different backport to `6.1` because we don't have the `WorkerLoop` in Java there (so it's just a one-liner in `6.1`).
* I have no idea how to write a test for this since all the Ruby specs have the PQ hardcoded in the global static settings. We'll have to revisit this once we have better infrastructure there imo.

### DIFF
--- a/logstash-core/lib/logstash/java_pipeline.rb
+++ b/logstash-core/lib/logstash/java_pipeline.rb
@@ -171,7 +171,7 @@ module LogStash; class JavaPipeline < JavaBasePipeline
     @filter_queue_client.set_pipeline_metric(
         metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :events])
     )
-    @drain_queue =  @settings.get_value("queue.drain")
+    @drain_queue =  @settings.get_value("queue.drain") || settings.get("queue.type") == "memory"
 
     @events_filtered = Concurrent::AtomicFixnum.new(0)
     @events_consumed = Concurrent::AtomicFixnum.new(0)
--- a/logstash-core/lib/logstash/pipeline.rb
+++ b/logstash-core/lib/logstash/pipeline.rb
@@ -183,7 +183,7 @@ module LogStash; class Pipeline < BasePipeline
     @filter_queue_client.set_pipeline_metric(
         metric.namespace([:stats, :pipelines, pipeline_id.to_s.to_sym, :events])
     )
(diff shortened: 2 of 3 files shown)


original-brownbear:
@andrewvc this is a pretty big bug in fact :( Maybe you can take a quick look since you were around for the move to `ArrayBlockingQueue` and this should be relatively obvious for you :)

jakelandis:
@original-brownbear - should this get ported to the (yet to be released) 6.2 version ?

original-brownbear:
@jakelandis yes def

jakelandis:
@original-brownbear - thanks i have tagged it as a blocker to 6.2.0

andrewvc:
OK, yes, this is bad. RE: testing, I'll take a look at fixing those tests.

colinsurprenant:
Looked at the changes and that makes sense. The worker loop condition was definitely wrong. 
We should have better test coverage for the queue drain. I am good to merge now given the urgency and add missing tests later.
LGTM

andrewvc:
@original-brownbear @colinsurprenant do you have any ways of repro-ing it? Trying with `generator` on my laptop it's impossible.

original-brownbear:
@andrewvc 

(zsh, on bash you need some other way to create a 6k line file)
```
repeat 6000 echo "sample_data" >> /tmp/in.txt
```

```sh
cat /tmp/in.txt | bin/logstash -e "" | grep sample_data | wc -l
```

should return 6k :) Clearly doesn't on `master`.

jakelandis:
@andrewvc or @original-brownbear - can this be merged to the 6.2 branch ?

original-brownbear:
@jakelandis no only `6.x` will work I think.
`6.2` will require a cherry pick of #9037.

I can do that tomorrow, but today I'm almost off I'm afraid.

andrewvc:
I'll handle the merging

andrewvc:
Confirmed via manual testing

colinsurprenant:
linking here that we had an issue open for adding drain specific tests #6779

PR #9025, opened 2018-01-24
MINOR: Remove unused method in java pipeline

original-brownbear:
This one isn't used anymore since the worker loop is now in Java as a result of #8988


original-brownbear:
@yaauie thanks!

PR #9032, opened 2018-01-25
BUILD: Remove  `system` invocations in rake tasks

original-brownbear:
Top to bottom:

* Inline call to plugin manager from `rake` instead of shelling out
   * Seems to make the build ~30s faster overall, locally as well as on Jenkins
* Install `clamp` because the plugin manager has it as a dependency
* Remove `system_gem` target since it's not used anywhere

### DIFF
--- a/rakelib/dependency.rake
+++ b/rakelib/dependency.rake
@@ -4,6 +4,10 @@ namespace "dependency" do
     Rake::Task["gem:require"].invoke("bundler", "~> 1.9.4")
   end
 
+  task "clamp" do
+    Rake::Task["gem:require"].invoke("clamp", "~> 0.6.5")
+  end
+
   task "rbx-stdlib" do
     Rake::Task["gem:require"].invoke("rubysl", ">= 0")
   end # task rbx-stdlib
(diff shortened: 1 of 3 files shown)


colinsurprenant:
Great stuff!!

Could you please just clarify where and when this `clamp` installation is required - or in which condition this missing `clamp` installation create a problem?

original-brownbear:
@colinsurprenant :) `clamp` is used by the plugin managers entry point `lib/pluginmanager/main.rb` as the CLI args library.

colinsurprenant:
@original-brownbear yeah I get that - I just want to understand why not having `clamp` installed by the `gems` tasks previously was not a problem or if it was, what was the problem it was creating and when?

original-brownbear:
@colinsurprenant ah got, it's fine in the other invocation (by calling the plugin manage script) because there the whole bundler environment gets loaded. We only need `clamp` + `bundler` here though, so loading the `clamp` gem manually is the easiest solution and in line with the handling of other rakelib dependencies.

colinsurprenant:
@original-brownbear ah ok, got it. thanks.
LGTM

original-brownbear:
@colinsurprenant thanks too :)

PR #9040, opened 2018-01-26
MINOR: Cleanup dead field and wrong ruby path in build.gradle

original-brownbear:
* `jrubyBin` wasn't used anymore and just snuck back in via a merge issue it seems
  * Removed it
* Also, there is no `/bin/ruby.bat` we only have `bin/ruby` => pointed the Windows version at the unwrapped `ruby` executable for now

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -143,11 +143,8 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
     into "${projectDir}/vendor/jruby"
 }
 
-def jrubyBin = "${projectDir}/vendor/jruby/bin/jruby" +
-  (System.getProperty("os.name").startsWith("Windows") ? '.bat' : '')
-
-def rubyBin = "${projectDir}/bin/ruby" +
-  (System.getProperty("os.name").startsWith("Windows") ? '.bat' : '')
+def rubyBin = "${projectDir}" +
+  (System.getProperty("os.name").startsWith("Windows") ? '/vendor/jruby/bin/jruby.bat' : '/bin/ruby')
 
 task installTestGems(dependsOn: downloadAndInstallJRuby) {
   inputs.files file("${projectDir}/Gemfile.template")


build.gradle
```diff
@@ -143,11 +143,8 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
     into "${projectDir}/vendor/jruby"
 }
 
-def jrubyBin = "${projectDir}/vendor/jruby/bin/jruby" +
-  (System.getProperty("os.name").startsWith("Windows") ? '.bat' : '')
```
colinsurprenant: maybe we should keep it and actually use that in the `installIntegrationTestBundler` task?
 https://github.com/elastic/logstash/pull/9040/files#diff-c197962302397baf3a4cc36463dce5eaR211

build.gradle
```diff
@@ -143,11 +143,8 @@ task downloadAndInstallJRuby(dependsOn: verifyFile, type: Copy) {
...
-
-def rubyBin = "${projectDir}/bin/ruby" +
-  (System.getProperty("os.name").startsWith("Windows") ? '.bat' : '')
+def rubyBin = "${projectDir}" +
+  (System.getProperty("os.name").startsWith("Windows") ? '/vendor/jruby/bin/jruby.bat' : '/bin/ruby')
```
colinsurprenant: 👍

build.gradle
original-brownbear: @colinsurprenant but then we're back to square one since we're losing the `LS_GEM_PATH` wrapper you added? (which is probably fine on Windows in the long run since no one has RVM or the like in a CMD or Powershell, but on *nix we need this here don't we?)

colinsurprenant: sorry I wasn't clear ... I meant to change 
```
commandLine "${projectDir}/vendor/jruby/bin/gem", "install", "bundler", "-v", "1.16.0", "-i", qaBundledGemPath
```
to use the `jrubyBin`
```
commandLine "${jrubyBin}/bin/gem", "install", "bundler", "-v", "1.16.0", "-i", qaBundledGemPath
```

This is really minor, but my thought is that is it more consistant this way. I'm fine either way, your choice!

colinsurprenant:
LGTM - minor comment at your discretion :)

build.gradle
original-brownbear: @colinsurprenant oh now I get it ... hmm, I think you're right here. I'll test this manually tomorrow :) (I don't wanna change any of this stuff without running it locally tbh. ... it just took too long to get this right :D)

original-brownbear: @colinsurprenant I decided to keep this as is for now. I ran into some random problems trying to wrap this in our Java executable ("couldn't find gem in path" errors ...) :( => will clean this up in the next step :)

PR #9046, opened 2018-01-26
Add jdbc_static to enriching data topic

dedemorton:
for this topic, I usually create a short example and test it myself, but due to time constraints, I had to take the example from the existing docs. Would be nice to have a shorter more concise example. I'd like to get this in for 6.2, then we can work on improving the example later.

### DIFF
--- a/docs/static/transforming-data.asciidoc
+++ b/docs/static/transforming-data.asciidoc
@@ -380,7 +380,7 @@ filter {
 --------------------------------------------------------------------------------
 
 
-<<plugins-filters-elasticsearch,elasticsearch>>::
+<<plugins-filters-elasticsearch,elasticsearch filter>>::
 
 Copies fields from previous log events in Elasticsearch to current events.
 +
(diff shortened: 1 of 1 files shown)


docs/static/transforming-data.asciidoc
```diff
@@ -450,7 +450,79 @@ For example:
...
 --------------------------------------------------------------------------------
 
-<<plugins-filters-jdbc_streaming,jdbc_streaming>>::
+
+<<plugins-filters-jdbc_static,jdbc_static filter>>::
+
+Enriches events with data pre-loaded from a remote database.
++
+The following example fetches data from a remote database, caches it in a local
+database, and uses lookups to enhance events with data cached in the local
```
acchen97: enrich instead of enhance?

docs/static/transforming-data.asciidoc
```diff
@@ -450,7 +450,79 @@ For example:
...
+        id => "local-servers"
+        query => "select descr as description from servers WHERE ip = :ip"
+        parameters => {ip => "[from_ip]"}
+        target => "server"
+      },
+      {
+        id => "local-users"
+        query => "select firstname, lastname from users WHERE userid = :id"
+        parameters => {id => "[loggedin_userid]"}
+        target => "user"
```
acchen97: also good to note that the "target" here will specify the field in which the looked up data will be stored in the event. so if there are multiple columns returning, it would be stored as a json object. the lines directly under this with the add_fields will allow you to pull fields from these json objects into top level event fields for easier analytics when it hits ES/KB.

docs/static/transforming-data.asciidoc
```diff
@@ -450,7 +450,79 @@ For example:
...
+      }
+    ]
+    # using add_field here to add & rename values to the event root
+    add_field => { server_name => "%{[server][0][description]}" }
+    add_field => { user_firstname => "%{[user][0][firstname]}" }
+    add_field => { user_lastname => "%{[user][0][lastname]}" }
+    remove_field => ["server", "user"]
+    jdbc_user => "logstash"
```
acchen97: @dedemorton @guyboertje we should also document how to use the loader_schedule option to refresh the cache. not specifying it would mean only one initial load.

https://github.com/logstash-plugins/logstash-filter-jdbc_static/blob/master/lib/logstash/filters/jdbc_static.rb#L80

dedemorton: That's documented in the reference topic. We should avoid putting too much detail in this topic because it's meant to be an intro, not the main documentation. This topic points off to the [main topic](https://www.elastic.co/guide/en/logstash/6.2/plugins-filters-jdbc_static.html#plugins-filters-jdbc_static-loader_schedule) for the nitty gritty details.

docs/static/transforming-data.asciidoc
```diff
@@ -450,7 +450,79 @@ For example:
...
++
+The following example fetches data from a remote database, caches it in a local
+database, and uses lookups to enhance events with data cached in the local
+database.
++
+[source,json]
+-----
+filter {
+  jdbc_static {
+    loaders => [
```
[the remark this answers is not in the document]
dedemorton: I'm tweaking the language a bit here and adding callouts to the example. I think we should avoid saying DDL because it's not really DDL (not a CREATE statement, for example).

dedemorton:
@acchen97 Please confirm my changes. thanks.

acchen97:
LGTM.

PR #9050, opened 2018-01-27
Edit the Logstash keystore docs and add to the build

dedemorton:
@jakelandis I made a quick editorial pass through the topic (also added the line wraps since I was in there anyhow). I didn't realize that would make it harder for you to see changes. :-/

I've added some comments/questions addressed to you that you'll have to remove.

All the links work and I confirmed the the topic builds without errors.

I was on the fence about whether to put the topic under setting up or configuring. Creating a keystore really seems more like a setup task than a config task, so I put it there. Here's what the TOC looks like when built:

image

### DIFF
--- a/docs/index-shared1.asciidoc
+++ b/docs/index-shared1.asciidoc
@@ -65,6 +65,9 @@ include::static/setting-up-logstash.asciidoc[]
 :edit_url: https://github.com/elastic/logstash/edit/{branch}/docs/static/settings-file.asciidoc
 include::static/settings-file.asciidoc[]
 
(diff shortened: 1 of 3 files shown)


jakelandis:
@dedemorton - thanks for the edits!  I will merge to the feature branch and address the comments.  Also, the TOC looks good.

jakelandis:
@dedemorton - Cleaned up based on your feed back on https://github.com/elastic/logstash/commit/5f834fab7060034f811517ecce51f69b62f6b2c2 , I still need the full command line doc.

PR #9051, opened 2018-01-28
Remove unused NodeStats#mem_payload

ioanatia:
The method was added in https://github.com/elastic/logstash/pull/5381/files#diff-5b309943da7f980637d1fb663cacd01a.
AFAICT it's not used at the moment.

```
$ git grep mem_payload
logstash-core/lib/logstash/api/modules/node_stats.rb:        def mem_payload
```


kares:
looking-good: needs a rebase or merge with master to trigger CI

PR #9052, opened 2018-01-29
BUILD: Fix wrong paths and wrong output declarations on installTestGems task

original-brownbear:
This fixes the obviously wrong `GEM_PATH` in `RubyGradleUtils.groovy` that caused one to have to run the Gradle `clean` goal needlessly (can reproduce this with `master` locally and also observed this together with @danhermann the other day).

* reproducer for concrete issue `./gradle clean test && ./gradlew test` => second `./gradlew test` fails because it can't find `bundler` in the wrong `GEM_DIR`

This issue isn't visible on CI because only `./gradlew clean test` still works (and CI always starts from a  clean state). Running the `test` goal without previously having run the `installTestGems` task (as a reult of a previous `clean`) was broken because now the gems weren't loaded correctly for subsequent tasks. 

Moreover, there were an additional two issues with the `installTestGems` task declaring its outputs:
* The `vendor/jruby` folder isn't an output of this task, it's an output of a previous task, this was causing some confusion for Gradle
* Using `fileTree` for output directories isn't stable in Gradle since `fileTree` is evaluated lazily. Moved to `.dir` and `.file` to fix wrong output declarations
   * This wasn't really visible before because the path declaration for `installTestGems` was wrong but then fixed by #8984 ... so what happened was that no change to the gem dir was ever registered because it was looking at the wrong non-existent dir => you needed to run the `clean` goal if your `bundle/jruby/2.3.0/gems` dir changed.

------------------------

Also:

* Fixed `verifyFile` task to have proper input and output declarations to be able to test that we actually don't run any more redundant tasks
* moved `def rubyBin` to a lower place in the build file where it's actually used to make things a little less confusing

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -116,24 +116,30 @@ project(":logstash-core") {
   }
 }
 
+def jrubyTarPath = "${projectDir}/vendor/_/jruby-bin-${jRubyVersion}.tar.gz"
+
 task downloadJRuby(type: Download) {
     description "Download JRuby artifact from this specific URL: ${jRubyURL}"
     src jRubyURL
     onlyIfNewer true
+    inputs.file("${projectDir}/versions.yml")
+    outputs.file(jrubyTarPath)
     dest new File("${projectDir}/vendor/_", "jruby-bin-${jRubyVersion}.tar.gz")
 }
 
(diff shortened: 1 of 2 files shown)


original-brownbear:
@andrewvc thanks!

PR #9053, opened 2018-01-29
[WIP]Upgrade to JRuby 9.1.15.0

original-brownbear:
We're still on `9.1.13.0` as result of #8738.

JRuby decided to not fix the behaviour change that led to #8738 in https://github.com/jruby/jruby/issues/4869#issuecomment-347440181 => we should go back to the latest version and fix things some other way if there is still an actual issue that could affect users.

### DIFF
--- a/versions.yml
+++ b/versions.yml
@@ -6,8 +6,8 @@ logstash-core-plugin-api: 2.1.16
 # jruby must reference a *released* version of jruby which can be downloaded from the official download url
 # *and* for which jars artifacts are published for compile-time
 jruby:
-  version: 9.1.13.0
-  sha1: 815bac27d5daa1459a4477d6d80584f007ce6a68
+  version: 9.1.15.0
+  sha1: 852b3dfd8b56a314ea4bf9502022a1f9edb8d7f6
 
 # jruby-runtime-override, if specified, will override the jruby version installed in vendor/jruby for logstash runtime only,
 # not for the compile-time jars


andrewvc:
@original-brownbear I'm in favor of that approach, but I think we should make that change first. I assume the UDP input/output need patching?

original-brownbear:
@andrewvc I don't think anything needs patching here. It's simply that IPV6 functionality fails if `java.net.preferIPv4Stack` is set to `true`.
I think this was fixed properly `9.1.15.0` by https://github.com/jruby/jruby/commit/d55cd486d99c7b521482529d17c429d059f59a6f now in so far that JRuby stops messing with the system property.
Nothing else to do for here imo :) Except for fixing this PR which I'll get to in a bit...

andrewvc:
@original-brownbear sure, but if I understand correctly users who need that functionality would see that break if they upgrade no?

Should we default that option to true to prevent that experience?

original-brownbear:
@andrewvc I don't think we should mess with that option. If anything forcing `false` would result in the current behavior as far as I can see. But in that case, we cause trouble to people that set it to `true` (which def. has its valid use cases).

=> I don't think we're really causing a behavior change by upgrading this now that JRuby stopped messing with the option. Literally the only case that could see a change is someone that consciously put `java.net.preferIPv4Stack=true` in their `jvm.options` but still wants IPV6 to work ...

andrewvc:
@original-brownbear oh, I see now, I read your original comment backward. That makes sense. We should still document this as a breaking change in whatever release it winds up in.

I'm +1 on this (once tests pass and it's GTG)

original-brownbear:
closing this as we're going for `9.1.16.0` now :)

PR #8987, opened 2018-01-18
JAVAFICATION: Port memory queue read client to Java. …

danhermann:
Also uses Java counters more efficiently per Armin's work in #8985. Rough throughput testing with generator input shows 15-20% increase due mainly to the metrics improvements.

PR #9056, opened 2018-01-29
JAVAFICATION: Java acked read client

original-brownbear:
Same as the other moves to JRuby like e.g #8987

### DIFF
--- a/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
+++ b/logstash-core/lib/logstash/util/wrapped_acked_queue.rb
@@ -59,7 +59,7 @@ module LogStash; module Util
     end
 
     def read_client()
-      ReadClient.new(self)
+      LogStash::AckedReadClient.create(self)
     end
 
     def check_closed(action)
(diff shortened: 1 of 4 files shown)


logstash-core/lib/logstash/util/wrapped_acked_queue.rb
```diff
@@ -59,7 +59,7 @@ def write_client
     end
 
     def read_client()
-      ReadClient.new(self)
+      LogStash::AckedReadClient.create(self)
```
danhermann: One thing I was not sure about here is whether all clients of `AckedReadClient` subsequently call `set_batch_dimensions` with explicit values. If not, this `create` method does not initialize the batch size and timeout while the `initialize` method of the Ruby implementation provides defaults for the batch size and timeout.

danhermann:
LGTM if, per the comment above, the batch size and timeout values are properly initialized.

logstash-core/lib/logstash/util/wrapped_acked_queue.rb
original-brownbear: @danhermann right ... thanks for spotting that one :) Fixing.

original-brownbear: @danhermann nevermind, this is correct. Sorry for the noise, see https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/java_pipeline.rb#L368 => the existing constructor was redundant => we're good here I think :)
... that is not to say that it isn't horrible that we're somehow setting the `@queue` in the constructor and the timeout values via some setter, but that's the next step :D

Issue #6728, opened 2017-02-17
Certain `@fields` are reserved fields and should not be removed.


guyboertje:
We need an enhancement to the remove_field code that should check if a field is reserved, ignore that field and log warn this.

Removing the `@timestamp` field will cause undefined method for nil class when trying to use it later.

liketic:
@suyograo Can I take this?

andrewvc:
Why is it that `@timestamp` should not be removable? Is this actually a bug?

While I understand that is the behavior today, is that intentional?

I think it's fine for us to decide that certain fields are reserved, but I'd like to understand the history of those fields before we flag them specially. 

In Logstash 2.2.2 for instance

`bin/logstash -e 'input { generator {} }  filter { mutate { remove => "@timestamp" } } output { stdout { codec => json_lines } }'`

ran successfully.

As did

`bin/logstash -e 'input { generator {} }  filter { mutate { remove => "@metadata" } } output { stdout { codec => json_lines } }'`

This feels more like accidental behavior to me as a result of moving the `Event` class to Java.

guyboertje:
I believe it is the elasticsearch output mechanism that breaks on `@timestamp` removal.

AbdulHaseebHussain:
What should the next steps be in regards to this issue?

guyboertje:
@andrewvc 

Dynamic values that use time format String interpolations fail without a timestamp.
Like when setting a time based index name or a `day_of_week` field.

The Java code expects that the timestamp is never null. https://github.com/elastic/logstash/blob/master/logstash-core/src/main/java/org/logstash/StringInterpolation.java#L44

guyboertje:
Like this one. https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/739

AbdulHaseebHussain:
Maybe this restriction of preventing deletion of reserved fields could be restricted to the timestamp field.

andrewvc:
My point is that the java code is not necessarily correct. I think the
conversation we should be having is what behaviour would be the best UX.
What would be best in line with the principle of least surprise?

PR #9061, opened 2018-01-29
Added Enhancement to prevent deletion of reserved fields

AbdulHaseebHussain:
Added enhancement to prevent deletion of reserved field in response to the following enhancement request https://github.com/elastic/logstash/issues/6728


andrewvc:
I have some questions about whether we really want this behavior, will ask them in #6728

andrewvc:
@AbdulHaseebHussain by the way, I do want to thank you for this PR, I just want to make sure we're doing it for the right reason if we decide to merge it.

PR #9070, opened 2018-01-30
Gem bump for 6.1.4

andrewvc:


Gemfile.jruby-2.3.lock.release
```diff
@@ -591,7 +591,7 @@ GEM
       simple_oauth (~> 0.3.0)
     tzinfo (1.2.4)
       thread_safe (~> 0.1)
-    tzinfo-data (1.2017.3)
```
jakelandis: Does this contain the DB in use too ?  Should be cautious about this update ?

andrewvc: Time zones do change with some regularity, I'm not too worried about it.

jakelandis:
LGTM

Issue #8499, opened 2017-10-18
Slow logger configuration is not included in packaged versions


jakelandis:
It appears that the packaged distributions (.rpm, .deb) do not include the slow logger configuration due to the duplication of the log4j2.properties files in the pkg directory. 

I assume that the log4j2.properties in the config directory (included as the .tar distribution) is intended to be the source of truth, and these two files have gotten out of sync. 

Ideally the fix isn't to simply copy/paste to bring the two back in sync, rather to ensure that they won't get out of sync. I believe this will be the same solution as https://github.com/elastic/logstash/issues/8427

jakelandis:
This is a related to #9066. 

To reiterate:
> Ideally the fix isn't to simply copy/paste to bring the two back in sync, rather to ensure that they won't get out of sync.

jordansissel:
> to ensure that they won't get out of sync

@jakelandis +1

The files in `pkg` are intended as things necessary on rpm (Red Hat) and deb (Debian, Ubuntu) platforms. Sometimes changes are made in these files in git, and sometimes they are made at [package install-time](https://github.com/elastic/logstash/blob/master/pkg/centos/after-install.sh#L4-L7)

When there are files we don't need to change at all (probably log4j2.properties?) or files we only need to change a little ([logstash.yml](https://github.com/elastic/logstash/blob/master/pkg/centos/after-install.sh#L4-L7)), maybe the right solution is to remove them from the `pkg` directory and have only a single source file -- such as config/log4j2.properties.

PR #9076, opened 2018-01-31
Unify logging properties across distributions

### LINKED ISSUE #9066: Use newer log retention policy in rpm/deb packages.
Reference: https://github.com/elastic/logstash/pull/8815

The [log4j2.properties file](https://github.com/elastic/logstash/blob/master/pkg/log4j2.properties) used in rpm and deb packages should be updated to use the same policy as implemented in #8815.

danhermann:
Removes `pkg/log4j2.properties`, `pkg/jvm.options`, and `pkg/startup.options` and copies the corresponding files from `config` during execution of the `rake artifact:rpm/deb` tasks so two separate copies of those files don't have to be kept in sync.

Fixes #9066, #8499.


jakelandis:
@jordansissel - should we put this into the up-coming `6.2.0` release to avoid odd messaging w/r/t ZIP vs. RPM/DEB packaging ?

danhermann:
@jakelandis, sorry for the delay on this one. I removed the symlinks, modified the `rake artifact` tasks to copy the config files from a single location in the `config` folder, removed the duplicate files in the `pkg` folder, and verified that both `rpm` and `deb` files are generated with the appropriate config files.

jakelandis:
LGTM - i also verified the `deb` is correct _(had un-related issues with my `rpm` build environment)_

danhermann:
Thanks, @jakelandis!

PR #9077, opened 2018-01-31
WIP POC: chunked JSON parser implementation

guyboertje:
As Jackson 2.9.0 has a `createNonBlockingByteArrayParser` that allows for text chunks to be fed into the parser.

- Accepts pretty printed or one per line JSON objects - no boundary detection needed.
- extracts Objects as Events from a root Array.
- Uses a JsonPointer path `/array-field/0/string-field` and converts to a FieldReference string `[array-field][0][string-field]`

Considerations for decoder (code) use.
- Needs to have new instances per identity stream.


yaauie:
@guyboertje to sumarise a separate conversation we had:

We will need to validate edge-case handling, especially around how we recover from couple varieties of malformed JSON (or malformed JSONL/NDJSON):

 - [ ] interleaved multilines
 - [ ] truncated messages
 - [ ] non-JSON lines injected in the stream

Specifically, we need to ensure that we are able to continue emitting events once the parser gets past the "bad" JSON, and that as small a hunk as possible is either lost or emitted unparsed.

We may also want to be able to inform the decoder what variant of streaming JSON we're working with, which may guide recovery (e.g., in JSONL/NDJSON, an unescaped newline is _by definition_ a message delimiter so no separate mitigation will be needed for interleaved multilines)

guyboertje:
I would also welcome discussion of the Event field vale update mechanism. Is it performant? Can it be improved?
Should the event timestamp initialisation be deferred until the Event is completely generated (i.e. considered complete)? We don't have a mechanism for that.

jordansissel:
> Accepts pretty printed or one per line JSON objects - no boundary detection needed.

Being able to get rid of `json` vs `json_lines` codec and just having a `json` codec would be delightful.

guyboertje:
@yaauie 
I did some changes.

I liked your inRoot detection and that went in.
Changed `ARRAY_PATH_START` pattern to require the second `/`
Renamed `evaluateRoot` to `realPathStartingPoint` and changed it so it does not have the weird -1,0,1+ semantics.
Renamed `bracketize` to `convertJsonPathFieldReference` and changed internals to better reflect what it is doing.
Changed the test data to include `"0field"` and `"1field"` keys to verify that those names are OK.

I kept the use of the JSON pointer string as I feel that we get it for free with Jackson.

This does not address the issue of malformed content. We may have to talk to Tatu Saloranta about the best strategy for advancing through the content until the parser can operate correctly again etc.

guyboertje:
Closing, not viable as interrupted chunks can't be handled well.

PR #9080, opened 2018-02-01
JAVAFICATION: Java wrapped acked queue

danhermann:


logstash-core/src/main/java/org/logstash/Logstash.java
```diff
@@ -44,7 +44,7 @@ public static void main(final String... args) {
         ) {
             logstash.run();
         } catch (final Throwable t) {
-            LOGGER.error(t);
+            LOGGER.error(t.toString());
```
danhermann: I believe this provides a more useful message. In my case, `LOGGER.error(t)` was printing only `java.lang.ExceptionInInitializerError` while `LOGGER.error(t.toString())` printed the exception class name along with the specific JRuby class that was failing to initialize.

logstash-core/src/main/java/org/logstash/ackedqueue/ext/JRubyAckedQueueExt.java
```diff
@@ -34,26 +34,39 @@ public Queue getQueue() {
...
+        return queueExt;
+    }
+
     @JRubyMethod(name = "initialize", optional = 7)
     public IRubyObject ruby_initialize(ThreadContext context, IRubyObject[] args) {
         args = Arity.scanArgs(context.runtime, args, 7, 0);
         int capacity = RubyFixnum.num2int(args[1]);
         int maxUnread = RubyFixnum.num2int(args[2]);
+
+        // are these out of order?
```
danhermann: I believe `checkpointMaxAcks` and `checkpointMaxWrites` are being pulled from the wrong indexes in the argument array.

logstash-core/src/main/java/org/logstash/ext/JrubyAckedReadClientExt.java
```diff
@@ -27,8 +27,8 @@
     private final ConcurrentHashMap<Long, Long> inflightClocks = new ConcurrentHashMap<>();
 
     private IRubyObject queue;
-    private IRubyObject batchSize;
-    private IRubyObject waitForMillis;
+    private IRubyObject batchSize = RubyNumeric.int2fix(RubyUtil.RUBY, 125);
+    private IRubyObject waitForMillis = RubyNumeric.int2fix(RubyUtil.RUBY, 50);
```
danhermann: These default values were necessary for the rspec tests since they do not use `java_pipeline` where these values are always explicitly set via the `set_batch_dimensions` method.

original-brownbear:
@danhermann can you resolve conflicts here?

original-brownbear:
@danhermann good to review here?

original-brownbear:
LGTM (remove the comment asking about arg order and squash + merge :))

PR #9086, opened 2018-02-01
Change logstash-core-plugin-api to properly follow versions.yml

guyboertje:
This unifies the 5.6, 6.X and 7.X mechanism for only needing to change version number in `versions.yml`

Fixes
https://internal-ci.elastic.co/job/elastic+release-manager+5.6+unified-snapshot/202/console
and
https://travis-ci.org/logstash-plugins/logstash-filter-mutate/jobs/335999317

Tested with `rake artifact:zip` and running a few configs through the unpacked artifact.

### DIFF
--- a/Gemfile.jruby-1.9.lock.release
+++ b/Gemfile.jruby-1.9.lock.release
@@ -1,7 +1,7 @@
 PATH
   remote: ./logstash-core
   specs:
-    logstash-core (5.6.7-java)
+    logstash-core (5.6.8-java)
       chronic_duration (= 0.10.6)
       clamp (~> 0.6.5)
       concurrent-ruby (~> 1.0, >= 1.0.5)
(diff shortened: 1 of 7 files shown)


Gemfile.jruby-1.9.lock.release
```diff
@@ -166,9 +166,9 @@ GEM
       jls-grok (~> 0.11.1)
       logstash-core-plugin-api (>= 1.60, <= 2.99)
       logstash-patterns-core
-    logstash-codec-netflow (3.10.0)
+    logstash-codec-netflow (3.11.0)
       bindata (>= 1.5.0)
-      logstash-core-plugin-api (>= 1.60, <= 2.99)
+      logstash-core-plugin-api (~> 2.0)
```
andrewvc: Why do we no longer support 1.6+ plugins here?

guyboertje: I guess because netflow codec did that change.
https://github.com/logstash-plugins/logstash-codec-netflow/commit/3c35fad0a00e7b1c56e4250482dfe9be3e5ca08b

andrewvc: @guyboertje I think this is fine. My concern was based on a misunderstanding. Plugins that require the 1.6 version of the plugin API were broken before, and will remain that way.

PR #9091, opened 2018-02-03
Fix empty array in boolean condition, NPE in boolean comparison

yaauie:
Merge-targets:
 - `master`,
 - `6.x`,
 - `6.2` (currently feature frozen for pending 6.2.0)

Fixes two separate issues:

 - NPE when LHS of boolean-equals or boolean-not-equals expression references unset field
 - Pipeline compilation error when configuration specifies literal empty array

Each commit contains rationale, etc.

### DIFF
--- a/logstash-core/lib/logstash/compiler/treetop_monkeypatches.rb
+++ b/logstash-core/lib/logstash/compiler/treetop_monkeypatches.rb
@@ -90,4 +90,16 @@ class Treetop::Runtime::SyntaxNode
         ""
       )
   end
+
+  def _parse_context(context = self)
+    start = interval.min
+    "line #{input.line_of(start)}, " +
+        "column #{input.column_of(start)} (byte #{start+1}) " +
+        "in `#{context.text_value}`"
+  end
+
+  def _node_type_name
+    name = self.class.name
+    name ? name.split('::').last : 'SyntaxNode'
+  end
 end
--- a/logstash-core/spec/logstash/compiler/compiler_spec.rb
+++ b/logstash-core/spec/logstash/compiler/compiler_spec.rb
@@ -560,18 +560,92 @@ describe LogStash::Compiler do
             end
 
             describe "'=='" do
-              let(:expression) { "[foo] == 5"}
+              let(:lhs) { fail(NotImplementedError) }
+              let(:expression) { "[foo] == #{rhs}"}
 
-              it "should compile correctly" do
-                expect(c_expression).to ir_eql(j.eEq(j.eEventValue("[foo]"), j.eValue(5.to_java)))
+              context 'when RHS represents an integer' do
+                let(:rhs) { '5' }
+                it "should compile correctly" do
+                  expect(c_expression).to ir_eql(j.eEq(j.eEventValue("[foo]"), j.eValue(5.to_java)))
+                end
+              end
+
+              context 'when RHS represents an empty array' do
+                let(:rhs) { '[]' }
+                it "should compile correctly" do
+                  expect(c_expression).to ir_eql(j.eEq(j.eEventValue("[foo]"), j.eValue([])))
+                end
+              end
+
+              # multi-item arrays are not supported because a single-item arrays in this context are
+              # syntactically ambiguous with _selector_s, which take a higher precedence; if/when support
+              # is added for unambiguously declaring a single-item array as a condition clause, this spec
+              # can be removed.
+              context 'when RHS represents a multi-item array' do
+                let(:rhs) { '["foo","bar"]'}
+                it "should fail to compile with a helpful message" do
+                  expect { c_expression }.to raise_error do |error|
+                    expect(error).to be_a_kind_of(RuntimeError)
+                    expect(error.message).to include("Invalid Config: Comparison exprs do not support non-empty arrays")
+                  end
+                end
+              end
+
+              # equality with a regexp is not something people _intend_ to do, and
+              # is an example of a valid `rvalue` that _parses_ but doesn't have semantic meaning.
+              context 'when RHS represents a Regexp' do
+                let(:rhs) { '/foo/' }
+                it "should fail to compile with a helpful message" do
+                  expect { c_expression }.to raise_error do |error|
+                    expect(error).to be_a_kind_of(RuntimeError)
+                    expect(error.message).to include("Invalid Config: Unsupported RegExp value on right-hand side of boolean expr")
+                  end
+                end
               end
             end
 
             describe "'!='" do
-              let(:expression) { "[foo] != 5"}
+              let(:lhs) { fail(NotImplementedError) }
+              let(:expression) { "[foo] != #{rhs}"}
 
-              it "should compile correctly" do
-                expect(c_expression).to ir_eql(j.eNeq(j.eEventValue("[foo]"), j.eValue(5.to_java)))
+              context 'when RHS represents an integer' do
+                let(:rhs) { '5' }
+                it "should compile correctly" do
+                  expect(c_expression).to ir_eql(j.eNeq(j.eEventValue("[foo]"), j.eValue(5.to_java)))
+                end
+              end
+
+              context 'when RHS represents an empty array' do
+                let(:rhs) { '[]' }
+                it "should compile correctly" do
+                  expect(c_expression).to ir_eql(j.eNeq(j.eEventValue("[foo]"), j.eValue([])))
+                end
+              end
+
+              # multi-item arrays are not supported because a single-item arrays in this context are
+              # syntactically ambiguous with `selector`s, which take a higher precedence; if/when support
+              # is added for unambiguously declaring a single-item array as a condition clause, this spec
+              # can be removed
+              context 'when RHS represents a multi-item array' do
+                let(:rhs) { '["foo","bar"]'}
+                it "should fail to compile with a helpful message" do
+                  expect { c_expression }.to raise_error do |error|
+                    expect(error).to be_a_kind_of(RuntimeError)
+                    expect(error.message).to include("Invalid Config: Comparison exprs do not support non-empty arrays")
+                  end
+                end
+              end
+
+              # equality with a regexp is not something people _intend_ to do, and
+              # is an example of a valid `rvalue` that _parses_ but doesn't have semantic meaning.
+              context 'when RHS represents a Regexp' do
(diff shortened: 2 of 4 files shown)


yaauie:
Since we don't make it terribly easy to get empty arrays in events, the following may be helpful for reproduction.

When this is cherry-picked onto 6.2, and I have a file `corpus.jsonl` with the following contents:

~~~ json
{"foo":[]}
{"foo":null}
{"bar":"not-foo"}
{"foo":""}
{"foo":"[]"}

~~~

And I have a logstash configuration `detect-empty-arrays.conf` like so:

~~~
input { stdin { codec => json_lines } }
filter { if [foo] == [] { mutate { add_tag => ["foo"] } } else { mutate { add_tag => ["no-foo"] } } }
output { stdout { codec => rubydebug } }
~~~

I expect exactly one of those messages to get tagged with "foo", and the rest to get tagged with "no-foo".

I get the following output:
~~~
╭─{ yaauie@castrovel:~/src/elastic/logstash (✘ 6.2) }
╰─● cat corpus.jsonl | bin/logstash -f detect-empty-arrays.conf
Sending Logstash's logs to /Users/yaauie/src/elastic/logstash/logs which is now configured via log4j2.properties
[2018-02-05T19:33:13,542][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"fb_apache", :directory=>"/Users/yaauie/src/elastic/logstash/modules/fb_apache/configuration"}
[2018-02-05T19:33:13,559][INFO ][logstash.modules.scaffold] Initializing module {:module_name=>"netflow", :directory=>"/Users/yaauie/src/elastic/logstash/modules/netflow/configuration"}
[2018-02-05T19:33:13,742][WARN ][logstash.config.source.multilocal] Ignoring the 'pipelines.yml' file because modules or command line options are specified
[2018-02-05T19:33:14,347][INFO ][logstash.runner          ] Starting Logstash {"logstash.version"=>"6.2.0"}
[2018-02-05T19:33:14,701][INFO ][logstash.agent           ] Successfully started Logstash API endpoint {:port=>9600}
[2018-02-05T19:33:18,193][INFO ][logstash.pipeline        ] Starting pipeline {:pipeline_id=>"main", "pipeline.workers"=>8, "pipeline.batch.size"=>125, "pipeline.batch.delay"=>50}
[2018-02-05T19:33:18,332][INFO ][logstash.pipeline        ] Pipeline started succesfully {:pipeline_id=>"main", :thread=>"#<Thread:0x7f05eb41 run>"}
[2018-02-05T19:33:18,430][INFO ][logstash.agent           ] Pipelines running {:count=>1, :pipelines=>["main"]}
{
          "tags" => [
        [0] "no-foo"
    ],
          "host" => "castrovel.local",
      "@version" => "1",
    "@timestamp" => 2018-02-05T19:33:18.504Z,
           "bar" => "not-foo"
}
{
          "tags" => [
        [0] "no-foo"
    ],
          "host" => "castrovel.local",
      "@version" => "1",
    "@timestamp" => 2018-02-05T19:33:18.504Z,
           "foo" => ""
}
{
          "tags" => [
        [0] "foo"
    ],
          "host" => "castrovel.local",
      "@version" => "1",
    "@timestamp" => 2018-02-05T19:33:18.469Z,
           "foo" => []
}
{
          "tags" => [
        [0] "no-foo"
    ],
          "host" => "castrovel.local",
      "@version" => "1",
    "@timestamp" => 2018-02-05T19:33:18.502Z,
           "foo" => nil
}
{
          "tags" => [
        [0] "no-foo"
    ],
          "host" => "castrovel.local",
      "@version" => "1",
    "@timestamp" => 2018-02-05T19:33:18.505Z,
           "foo" => "[]"
}
[2018-02-05T19:33:19,196][INFO ][logstash.pipeline        ] Pipeline has terminated {:pipeline_id=>"main", :thread=>"#<Thread:0x7f05eb41 run>"}
[success (21.000s)]

~~~

jordansissel:
Tested a few scenarios by hand, all work as expected:

* `[] == []` ✔️  (true)
* `[foo] == []` ✔️ (true)
* `["foo", "bar"] == 1` ✔️ (config error, Comparison exprs do not support non-empty arrays)
* ` 1 == ["foo", "bar"]` ✔️ (config error, Comparison exprs do not support non-empty arrays)

I found some unrelated compiler/crashes also. Good times ;)

logstash-core/lib/logstash/compiler/treetop_monkeypatches.rb
```diff
@@ -90,4 +90,16 @@ def _inspect(indent="")
         ""
       )
   end
+
+  def _parse_context(context = self)
```
jordansissel: style: I was curious about the `_` method name prefix, but I found only one prior method here (`_inspect)` which I assume was named to avoid clashing with Ruby's `Object#inspect` ?

Can you tell me about the reason for the underscore prefix?

logstash-core/spec/logstash/compiler/compiler_spec.rb
```diff
@@ -560,18 +560,92 @@ def compose(*statements)
...
+                it "should compile correctly" do
+                  expect(c_expression).to ir_eql(j.eEq(j.eEventValue("[foo]"), j.eValue([])))
+                end
+              end
+
+              # multi-item arrays are not supported because a single-item arrays in this context are
+              # syntactically ambiguous with _selector_s, which take a higher precedence; if/when support
+              # is added for unambiguously declaring a single-item array as a condition clause, this spec
+              # can be removed.
+              context 'when RHS represents a multi-item array' do
```
jordansissel: Thinking more about this case, I wonder if we should allow multi-element arrays as rvalues?

As we discussed, there is ambiguity for field references vs single-element arrays, but for multi-element there is no ambiguity. This said, we 6.2 crashes on array comparison anyway: `"foo" == [1,2,3]` hits a bug and throws an exception.

Thoughts on allowing multi-element `[foo] == [1,2,3]` ?

It's ok if multi-element waits until a future PR; I dont' want to stall this bug fix.

logstash-core/spec/logstash/compiler/compiler_spec.rb
```diff
@@ -560,18 +560,92 @@ def compose(*statements)
...
+              end
+
+              # equality with a regexp is not something people _intend_ to do, and
+              # is an example of a valid `rvalue` that _parses_ but doesn't have semantic meaning.
+              context 'when RHS represents a Regexp' do
+                let(:rhs) { '/foo/' }
+                it "should fail to compile with a helpful message" do
+                  expect { c_expression }.to raise_error do |error|
+                    expect(error).to be_a_kind_of(RuntimeError)
+                    expect(error.message).to include("Invalid Config: Unsupported RegExp value on right-hand side of boolean expr")
```
jordansissel: Maybe this should be more actionable for the user -- 

Suggestion:

> `RegExp value <value> may only be used with the =~ or !~ operator, for example: #{lval} =~ #{rval}`

thoughts?

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -331,6 +328,12 @@ private static EventCondition eq(final EventValueExpression evalE,
                 return new EventCondition.Compiler.FieldEqualsLong(
                     field, ((Number) value).longValue()
                 );
+            } else if (value instanceof List) {
+                if (!((List) value).isEmpty()) {
+                    // shouldn't get here; this is pre-validated in the lscl.
+                    throw new IllegalStateException("non-empty arrays are not allowed in boolean expressions");
```
jordansissel: Thought: Maybe we should explain to the user why? I'd be confused and maybe upset at this seemingly-arbitrary restriction.

logstash-core/lib/logstash/compiler/treetop_monkeypatches.rb
yaauie: yes; we're monkey-patching, and avoiding collisions is the goal.

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
yaauie: yeah; makes sense.

```
non-empty arrays are not allowed in boolean expressions because the Logstash syntax in this context cannot unambiguously differentiate between a field selector and a single-element array.
```

More wordy, but indicates that we're not just being jerks :/

logstash-core/spec/logstash/compiler/compiler_spec.rb
yaauie: I think that adding support for multi-element arrays when we don't have a way to unambiguously represent single-element arrays is an opportunity for more confusion and surprises than it's worth, e.g.:

 - `if [foo] == []`: "if the event's `foo` is an empty array"
 - `if [foo] == [1]`: "if the event's `foo` field is equal to the event's `1` field" - wut?
 - `if [foo] == [1,2]`: "if the event's `foo` is equal to the array `[1,2]`"

Adding support for multi-element arrays hints at a continuity that we cannot currently represent, so while we don't currently have a way to reject input like `if [foo] == [1]`, at least we aren't encouraging a syntax where a user's intent to remove one element from an array drastically changes the meaning to do something completely different than their intent.

logstash-core/spec/logstash/compiler/compiler_spec.rb
yaauie: This spec is an example of what the "uncaught" `rvalue` would produce, and adding specific call-outs to known-not-related nodes is a bit of a slippery slope that I was trying to avoid (e.g., If we were to pass a valid `method_call`, which has no underlying support, a similar message would be generated, except it would have `MethodCall` instead of `RegExp`).

I agree that we could probably have better messaging for `RegExp`, even though it never should belong as the RHS of an equality or comparison operation; I still feel that this patch is a net improvement, but am willing to add in the extra warning for `RegExp`, since it is more likely to get hit:

With this patch:
~~~
Invalid Config: Unsupported RegExp value on right-hand side of boolean expr; line 24 column 18 (byte 228) in `[foo] == /bar/`
~~~

Previously:
~~~
Failed to parse right-hand side of conditional [foo] == /bar/
~~~

LucaWintergerst:
I just ran into this issue when trying to check if a field is an empty array. Any chance that we can look into this again? Tested on 6.4.1

Issue #9093, opened 2018-02-05
config.test_and_exit : warn users explicitly


jelmerdereus:
If the configtest doesn't print 'Configuration OK' you can assume that something is wrong with the file(s) and you need to consult logstash-plain.log.

admin@elk01:~$ sudo -u logstash /usr/share/logstash/bin/logstash --config.test_and_exit --path.settings /etc/logstash -t
Sending Logstash's logs to /data/logstash-plain.log which is now configured via log4j2.properties
admin@elk01:~$ 

I would like to make this more explicit for users.

PR #9094, opened 2018-02-05
Making config.test_and_exit more explicit for invalid configurations …

jelmerdereus:
I would really like to give a user of Logstash an explicit error when the config check is running into problems. Please have a look at issue #9093 .

Although an error is raised in the function execute in logstash-core/lib/logstash/runner.rb, this isn't shown to the user. Successful configuration, however, is shown to the user.

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -324,6 +324,7 @@ class LogStash::Runner < Clamp::StrictCommand
           puts "Configuration OK"
           logger.info "Using config.test_and_exit mode. Config Validation Result: OK. Exiting Logstash"
         else
+          puts "Configuration invalid. Check logstash-plain.log for clues"
           raise "Could not load the configuration file"
         end
         return 0


karmi:
Hi @jelmerdereus, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/logstash/pull/9094.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

jelmerdereus:
hi @karmi thanks for your superfast reply :+1:  I just added my previous email address to my profile and verified it.

jelmerdereus:
I signed the CLA by the way. I thought I had done this before but I wasn't 100% sure

jelmerdereus:
I'm sorry I had to add one more email address. On the CLI git was using a Gitlab account from work.

jelmerdereus:
@karmi would you have a look? It is some CLI output for the user when --config.test_and_exit fails.

Issue #9081, opened 2018-02-01
Logstash 6.1.3 NoMethodError


arko-r:
Logstash crashes when indexing linestrings:

~~~
`Failed to execute action {:action=>LogStash::PipelineAction::Stop/pipeline_id:main, :exception=>"NoMethodError", :message=>"undefined method `map' for nil:NilClass\nDid you mean?  tap", :backtrace=>["E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/util.rb:40:in `thread_info'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/pipeline.rb:670:in `block in plugin_threads_info'", "org/jruby/RubyArray.java:2486:in `map'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/pipeline.rb:670:in `plugin_threads_info'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/pipeline_reporter.rb:66:in `block in to_hash'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/util/wrapped_synchronous_queue.rb:80:in `inflight_batches'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/pipeline_reporter.rb:56:in `to_hash'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/pipeline_reporter.rb:51:in `snapshot'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/shutdown_watcher.rb:88:in `pipeline_report_snapshot'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/shutdown_watcher.rb:63:in `block in start'", "E:/Program/logstash/logstash-6.1.3/vendor/bundle/jruby/2.3.0/gems/stud-0.0.23/lib/stud/interval.rb:20:in `interval'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/shutdown_watcher.rb:59:in `start'", "E:/Program/logstash/logstash-6.1.3/logstash-core/lib/logstash/shutdown_watcher.rb:35:in `block in start'"]}`
~~~

The input is a MSSQL server using jdbc and the output is Elasticsearch.
There are no other errors visible when using the TRACE loglevel.
All the data is send to and received by Elasticsearch, but Logstash still gives this error.

- Version: 6.1.3
- Operating System: Windows 10

[...]

PR #9097, opened 2018-02-05
fix crash during exit when reporting on already-reaped thread

yaauie:
Although `LogStash::JavaBasePipeline#plugin_threads_info` attempts to only report on
living threads, it is possible for a thread to die/complete after the check and before
we attempt to extract status in `LogStash::Util#thread_info`; since `Thread#backtrace`
can return `nil` when the thread is no longer living, this causes a crash.

By falling back to an empty array when the backtrace cannot be ascertained from the
given thread, we can emit a thread status even if the thread has already completed.

### DIFF
--- a/logstash-core/lib/logstash/util.rb
+++ b/logstash-core/lib/logstash/util.rb
@@ -37,7 +37,8 @@ module LogStash::Util
   end
 
   def self.thread_info(thread)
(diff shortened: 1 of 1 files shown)


yaauie:
Should resolve https://github.com/elastic/logstash/issues/9081

PR #9101, opened 2018-02-06
Resolve review comments

dedemorton:
Cleaned up the keystore description based on review comments from @jordansissel.

Moved the description of how LS keystore compares to ES to appear later in the document because it's not the main idea here and shouldn't be covered so early in the topic.


jakelandis:
LGTM

PR #9103, opened 2018-02-06
Add link to keystore plus do light edit

dedemorton:
I started to add in the link to the keystore doc and noticed a few inconsistencies in the relnotes, so I did a quick cleanup.

### DIFF
--- a/docs/static/releasenotes.asciidoc
+++ b/docs/static/releasenotes.asciidoc
@@ -18,32 +18,32 @@ endif::include-xpack[]
 [[logstash-6-2-0]]
 === Logstash 6.2.0 Release Notes
 
-* Added support to protect sensitive settings and configuration in a keystore
-* Added {logstash-ref}/plugins-filters-jdbc_static.html[JDBC static filter] as a default plugin
-* Set better defaults to allow for higher throughput under load. https://github.com/elastic/logstash/issues/8707[#8707] https://github.com/elastic/logstash/issues/8702[#8702]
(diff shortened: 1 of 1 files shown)


docs/static/releasenotes.asciidoc
```diff
@@ -18,32 +18,32 @@ endif::include-xpack[]
...
+* Added the ability to specify `--pipeline.id` from the command line. (https://github.com/elastic/logstash/issues/8868[#8868])
+* Implemented continued improvements to the next generation of execution. Give it a try with the command line switch `--experimental-java-execution`.
 
 ==== Plugins
 
-*JDBC Static Filter*
+*Jdbc_static Filter*
 
-* Initial release
+* Released the initial version the {logstash-ref}/plugins-filters-jdbc_static.html[jdbc_static filter], which enriches events with data pre-loaded from a remote database.
```
dedemorton: added a description here because I thought it might be useful for users to know what this new plugin is about.

jakelandis: I like the addition, but should read `local database` , or `in memory database`, not remote.

dedemorton: I went back and forth with @guyboertje and @acchen97 on this language a couple of times, and this is what we settled on. The data used to enrich the events does come from a remote database, but it's preloaded. My recollection is that Alvin didn't want us to focus too much on the local database because its an implementation detail?

acchen97: I'm good with the verbiage here.

acchen97:
LGTM.

PR #9109, opened 2018-02-06
Soften text around data loss

### LINKED ISSUE #8488: Update SynchronousQueue documentation.
This page: https://www.elastic.co/guide/en/logstash/6.x/execution-model.html (and master/current) still refers to the the SynchronousQueue.

As of 6.1 with [this](https://github.com/elastic/logstash/pull/7691/files#diff-1eb63ed8cd2fb1ca5d896c221184d186L8) commit, the implementation been changed to an ArrayBlockingQueue, which does have a size. 

>This queue holds no events, instead transferring each pushed event to a free worker, blocking if all workers are busy.

Is no longer valid as the size of the queue is a function of the pipeline workers and batch size. However, imo, that is an implementation detail unnecessary to continue to include at this level of the documentation. 

In summary, I suggest to change:

> Each input stage in the Logstash pipeline runs in its own thread. Inputs write events to a common Java SynchronousQueue. This queue holds no events, instead transferring each pushed event to a free worker, blocking if all workers are busy. Each pipeline worker thread takes a batch of events off this queue, creating a buffer per worker, runs the batch of events through the configured filters, then runs the filtered events through any outputs.

to

> Each input stage in the Logstash pipeline runs in its own thread. Inputs write events to a central queue that is either in memory (default) or on disk.  Each pipeline worker thread takes a batch of events off this queue, creating a buffer per worker, runs the batch of events through the configured filters, then runs the filtered events through any outputs.

Also (un-related, but on same page) can we make a minor change from:

> If Logstash terminates unsafely, any events that are stored in memory will be lost. To prevent data loss, you can enable Logstash to persist in-flight events to disk. See Persistent Queues for more information.

to

> If Logstash terminates unsafely, any events that are stored in memory may be lost. To help prevent data loss, you can enable Logstash to persist in-flight events to disk. See Persistent Queues for more information.

^^ softened wording very slightly w/r/t to data loss 

cc @dedemorton @original-brownbear @jordansissel

karenzone:
Fixes #8488

### DIFF
--- a/docs/static/life-of-an-event.asciidoc
+++ b/docs/static/life-of-an-event.asciidoc
@@ -82,13 +82,17 @@ For more information about the available codecs, see
 === Execution Model
 
 The Logstash event processing pipeline coordinates the execution of inputs,
-filters, and outputs. 
+filters, and outputs.
 
-Each input stage in the Logstash pipeline runs in its own thread. Inputs write events to a common Java https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/SynchronousQueue.html[SynchronousQueue]. This queue holds no events, instead transferring each pushed event to a free worker, blocking if all workers are busy. Each pipeline worker thread takes a batch of events off this queue, creating a buffer for each worker, runs the batch of events through the configured filters, then runs the filtered events through any outputs. The size of the batch and number of pipeline worker threads are configurable (see <<tuning-logstash>>). 
+Each input stage in the Logstash pipeline runs in its own thread. Inputs write
+events to a central queue that is either in memory (default) or on disk. Each
+pipeline worker thread takes a batch of events off this queue, runs the batch of
+events through the configured filters, and then runs the filtered events through
+any outputs. The size of the batch and number of pipeline worker threads are
+configurable (see <<tuning-logstash>>).
 
 By default, Logstash uses in-memory bounded queues between pipeline stages
 (input → filter and filter → output) to buffer events. If Logstash terminates
-unsafely, any events that are stored in memory will be lost. To prevent data
+unsafely, any events that are stored in memory will be lost. To help prevent data
 loss, you can enable Logstash to persist in-flight events to disk. See
 <<persistent-queues>> for more information.
-


docs/static/life-of-an-event.asciidoc
```diff
@@ -82,13 +82,17 @@ For more information about the available codecs, see
...
+events to a central queue that is either in memory (default) or on disk. Each
+pipeline worker thread takes a batch of events off this queue, creating a buffer
+per worker, runs the batch of events through the configured filters, and then
+runs the filtered events through any outputs. The size of the batch and number
+of pipeline worker threads are configurable (see <<tuning-logstash>>).
 
 By default, Logstash uses in-memory bounded queues between pipeline stages
 (input → filter and filter → output) to buffer events. If Logstash terminates
-unsafely, any events that are stored in memory will be lost. To prevent data
+unsafely, any events that are stored in memory might be lost. To prevent data
```
andrewvc: These events 100% will be lost, so we shouldn't soften the language here. The buffers might be empty however, but in a situation with constant data that will rarely be the case.

jakelandis: I had originally suggested softening the wording based on the empty queue scenario, but agree that we should assume constant data.  (e.g. +1 to @andrewvc 's point)

jakelandis: Can we also change 
```
To prevent data loss
```
to
```
To help prevent data loss
```
For example buffered data in the input from a non-acking source would loose data even with the PQ if Logstash was forcefully killed.

docs/static/life-of-an-event.asciidoc
```diff
@@ -82,13 +82,17 @@ For more information about the available codecs, see
...
 
 The Logstash event processing pipeline coordinates the execution of inputs,
-filters, and outputs. 
+filters, and outputs.
 
-Each input stage in the Logstash pipeline runs in its own thread. Inputs write events to a common Java https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/SynchronousQueue.html[SynchronousQueue]. This queue holds no events, instead transferring each pushed event to a free worker, blocking if all workers are busy. Each pipeline worker thread takes a batch of events off this queue, creating a buffer per worker, runs the batch of events through the configured filters, then runs the filtered events through any outputs. The size of the batch and number of pipeline worker threads are configurable (see <<tuning-logstash>>). 
+Each input stage in the Logstash pipeline runs in its own thread. Inputs write
+events to a central queue that is either in memory (default) or on disk. Each
+pipeline worker thread takes a batch of events off this queue, creating a buffer
+per worker, runs the batch of events through the configured filters, and then
```
jakelandis: can we remove `, creating a buffer per worker`

jakelandis:
LGTM , @andrewvc - concur ?

PR #9116, opened 2018-02-08
Fix PQ stats with ruby pipeline

robbavey:
Stats were not being collected when the PQ was enabled due to the
queue class type check not being updated after Javafication of
WrappedAckedQueue

Fixes https://github.com/elastic/x-pack-logstash/issues/335


colinsurprenant:
LGTM

robbavey:
Thanks @colinsurprenant!

PR #9120, opened 2018-02-08
Bump gems for 6.2.2

andrewvc:
Gem bump post 6.2.1 release


andrewvc:
@robbavey I think I just fixed it

PR #9131, opened 2018-02-09
Cap the size of field references to prevent OOMs with random keys.

andrewvc:
We cache FieldReferences, which is good, because they create a lot of garbage.
However, this cache is unbounded. A config that sets many unique values could OOM the JVM.

For example:

```
input {
  generator {}
}

filter {
  ruby {
    code => "event.set(SecureRandom.uuid, 5)"
  }
}

output {
}
```

can kill LS.

This patch fixes the behavior.

Running the above config with yourkit we can see before and after:

Before:

image

After:

image

### DIFF
--- a/logstash-core/src/main/java/org/logstash/FieldReference.java
+++ b/logstash-core/src/main/java/org/logstash/FieldReference.java
@@ -1,10 +1,15 @@
 package org.logstash;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import org.codehaus.commons.nullanalysis.NotNull;
+
+import javax.annotation.Nonnull;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 
 public final class FieldReference {
 
@@ -29,7 +34,14 @@ public final class FieldReference {
     /**
      * Holds all existing {@link FieldReference} instances for de-duplication.
      */
-    private static final Map<FieldReference, FieldReference> DEDUP = new HashMap<>(64);
+    private static final LoadingCache<FieldReference, FieldReference> DEDUP = CacheBuilder.newBuilder()
+            .maximumSize(2000)
+            .build(new CacheLoader<FieldReference, FieldReference>() {
+                @Override
+                public FieldReference load(@Nonnull FieldReference key) throws Exception {
+                    return key;
+                }
+            });
 
     /**
      * Unique {@link FieldReference} pointing at the timestamp field in a {@link Event}.
@@ -43,8 +55,14 @@ public final class FieldReference {
     /**
      * Cache of all existing {@link FieldReference}.
      */
-    private static final Map<CharSequence, FieldReference> CACHE =
-        new ConcurrentHashMap<>(64, 0.2F, 1);
+    private static final LoadingCache<CharSequence, FieldReference> CACHE = CacheBuilder.newBuilder()
+            .maximumSize(1000)
+            .build(new CacheLoader<CharSequence, FieldReference>() {
+                @Override
+                public FieldReference load(@Nonnull CharSequence key) throws Exception {
+                    return parse(key);
+                }
+            });
 
     private final String[] path;
 
@@ -67,11 +85,11 @@ public final class FieldReference {
 
     public static FieldReference from(final CharSequence reference) {
         // atomicity between the get and put is not important
-        final FieldReference result = CACHE.get(reference);
-        if (result != null) {
-            return result;
+        try {
+            return CACHE.get(reference);
+        } catch (ExecutionException e) {
+            throw new RuntimeException(e);
         }
-        return parseToCache(reference);
     }
 
     /**
@@ -106,19 +124,16 @@ public final class FieldReference {
     }
 
     /**
-     * De-duplicates instances using {@link FieldReference#DEDUP}. This method must be
-     * {@code synchronized} since we are running non-atomic get-put sequence on
-     * {@link FieldReference#DEDUP}.
+     * De-duplicates instances using {@link FieldReference#DEDUP}.
      * @param parsed FieldReference to de-duplicate
      * @return De-duplicated FieldReference
      */
-    private static synchronized FieldReference deduplicate(final FieldReference parsed) {
-        FieldReference ret = DEDUP.get(parsed);
-        if (ret == null) {
-            DEDUP.put(parsed, parsed);
-            ret = parsed;
+    private static FieldReference deduplicate(final FieldReference parsed) {
+        try {
+            return DEDUP.get(parsed);
+        } catch (ExecutionException e) {
+            throw new RuntimeException(e);
         }
-        return ret;
     }
 
     /**
@@ -136,12 +151,6 @@ public final class FieldReference {
         }
         hash = prime * hash + key.hashCode();
         return prime * hash + type;
-    }
-
-    private static FieldReference parseToCache(final CharSequence reference) {
-        final FieldReference result = parse(reference);
-        CACHE.put(reference, result);
-        return result;
     }
 
     private static FieldReference parse(final CharSequence reference) {


yaauie:
How about an LRU cache? Super easy to implement with `LinkedHashMap`, and we wouldn't have to incur the overhead of rebuilding the cache every time we hit the max:

> ~~~ java
> import java.util.LinkedHashMap;
> import java.util.Map;
> 
> public LRUCache<K, V> extends LinkedHashMap<K, V> {
>   private int cacheSize;
> 
>   public LRUCache(int cacheSize) {
>     super(16, 0.75, true);
>     this.cacheSize = cacheSize;
>   }
> 
>   protected boolean removeEldestEntry(Map.Entry<K, V> eldest) {
>     return size() >= cacheSize;
>   }
> }
> ~~~
> -- http://chriswu.me/blog/a-lru-cache-in-10-lines-of-java/

andrewvc:
@yaauie yes, we could do that for DEDUP I suppose. We do, however, use a `ConcurrentHashMap` for `CACHE`, and we'd have to either convert that the `LinkedHashMap` + `synchronized` (which is likely slower) or figure something else out, like keeping the CHM but with an extra layer of indirection.

Did you read my comment about why a CHM in this case may not buy us much? Most Logstash users don't churn keys often. One way we've (accidentally) proved that is by this bug getting through. No reports yet luckily! Given that, does the extra correctness of an LRU buy us much?

Additionally, thinking through real-world use cases. The computed value in these maps is fairly cheap. The main thing we're doing is trying to save heap churn, not computation.

There are two use cases here:

1. *Small total set of keys, all stable*: This solution is optimal here because there is no additional book-keeping.
2. *Some distribution of stable and random key accesses*: In this case we could save stablekeys of the garbage, but is that really meaningful? Is it worth the extra book-keeping of an LRU? That would depend on the user's access pattern of course.

I suspect that this patch basically addresses something that almost never happens. Given that, I'd rather keep complexity low in this patch. One thing to recall, this only affects `FieldReference` objects, not all values. If I have a random value nested inside a map, and assign that map to a stable reference it doesn't accrue more in the cache.

jakelandis:
Could also use soft references in the map to allow the GC to collect your values if they get too numerous. It can be a decent design for the edge case users that still allows no cap, but no OOM.

yaauie:
Alternatively, we could use [`com.google.common.cache.CacheBuilder<K,V>`](https://google.github.io/guava/releases/22.0/api/docs/com/google/common/cache/CacheBuilder.html) since we already have guava 22 as a dependency:

~~~ diff
diff --git a/logstash-core/src/main/java/org/logstash/FieldReference.java b/logstash-core/src/main/java/org/logstash/FieldReference.java
index b9547fa62..2f6428330 100644
--- a/logstash-core/src/main/java/org/logstash/FieldReference.java
+++ b/logstash-core/src/main/java/org/logstash/FieldReference.java
@@ -1,10 +1,10 @@
 package org.logstash;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+
+import java.util.*;
+import java.util.concurrent.ExecutionException;
 
 public final class FieldReference {
 
@@ -43,8 +43,9 @@ public final class FieldReference {
     /**
      * Cache of all existing {@link FieldReference}.
      */
-    private static final Map<CharSequence, FieldReference> CACHE =
-        new ConcurrentHashMap<>(64, 0.2F, 1);
+    private static final Cache<CharSequence, FieldReference> CACHE = CacheBuilder.newBuilder()
+            .maximumSize(2000)
+            .build();
 
     private final String[] path;
 
@@ -66,12 +67,15 @@ public final class FieldReference {
     }
 
     public static FieldReference from(final CharSequence reference) {
-        // atomicity between the get and put is not important
-        final FieldReference result = CACHE.get(reference);
-        if (result != null) {
-            return result;
+        try {
+            Objects.requireNonNull(reference, "reference");
+
+            return CACHE.get(reference, () -> parse(reference));
+        } catch (ExecutionException e) {
+            // any exception raised in our cache-loader (the static method `FieldReference#parse`) will be raised
+            // by `Cache#get(FieldReference, CacheLoader)` wrapped in an `ExecutionException`; raise the error in a useful way.
+            throw new RuntimeException(String.format("Failed to parse field reference `%s`", reference), e.getCause());
         }
-        return parseToCache(reference);
     }
 
     /**
@@ -138,12 +142,6 @@ public final class FieldReference {
         return prime * hash + type;
     }
 
-    private static FieldReference parseToCache(final CharSequence reference) {
-        final FieldReference result = parse(reference);
-        CACHE.put(reference, result);
-        return result;
-    }
-
     private static FieldReference parse(final CharSequence reference) {
         final ArrayList<String> path = new ArrayList<>();
         final int length = reference.length();
~~~

andrewvc:
Patched with `LoadingCache`. Thanks @yaauie !

GC is still looking good:
We cache FieldReferences, which is good, because they create a lot of garbage.
However, this cache is unbounded. A config that sets many unique values could OOM the JVM.

For example:

```
input {
  generator {}
}

filter {
  ruby {
    code => "event.set(SecureRandom.uuid, 5)"
  }
}

output {
}
```

can kill LS.

This patch fixes the behavior.

original-brownbear:
@andrewvc Sorry for only notificing this now, but why not simply up a counter whenever the fieldcache gets a new member and clear `DEDUP` and the `CACHE` when the counter has `size % 2000 == 0`? Seems a lot safer than changing the behaviour with a new data structure here and keeps the performance of the solution 100% performance neutral doesn't it? + It's like literally 10 lines of new code only :)

andrewvc:
@original-brownbear that was actually the original content of this PR (though I used Map.size, which is surely less efficient than the counter). 

Let's chat on monday. @yaauie had some concerns re: object identity that would affect any clearing of the cache AFAICT.

If you have a better approach on I'm glad to abandon this PR in favor of a new one.

original-brownbear:
@andrewvc Hmm come to think of it (good thing you mentioned "identity" :D) ... this patch isn't enough to fix the leaking when using an infinite number of keys.
We also have the fact that all the `ConvertedMap` and `FieldReference` interaction is based on the fact that all `String`s used by the `FieldReference` are interned.
So we'd have to fix/change that as well. Even with this change, we'll leak memory for the interned `String`.

That said ... I think just looking at the `size` is fine especially if we move away from the interning.
Clearing of the cache isn't an issue since we're doing that in a `synchronized` block. Worst case we'll get a few duplicate `FieldReference` for `FieldReference`s that are cached outside of the `CACHE` here, which won't really have a measurable performance impact imo :)

original-brownbear:
@andrewvc also, if we want to move away from `intern`ing all `String`s here, we'll have to move back to `HashMap` for all `Event` data. This is quite a hit in terms of overall memory consumption.
Since this doesn't seem to be an urgent issue affecting anyone in particular, maybe we should take some time and find a way to support the use-case of an infinite keyspace without worsening performance and memory consumption for existing users that don't have any problems?

andrewvc:
@original-brownbear what about @yaauie 's comment above?

> In general, I think this is a useful step forward and necessary safeguard, but it creates the possibility for a fairly catastrophic and nearly-impossible-to-debug issue if both our DEDUP and CACHE end up evicting FieldReference.TIMESTAMP_REFERENCE; more on that below.

I agree that we can take some time here. I think it's an important issue to fix, I think it will bite us at some point, and be hard to debug, but there is no open issue so we should be careful and do it the right way.

andrewvc:
@original-brownbear I think it makes the most sense to change `==` to `.equals` in this patch. If that isn't as fast, that's OK, we can improve the speed later.

Let's make it more correct (in the sense of not leaking to heap) first, and we can micro-optimize later if needed.

original-brownbear:
@andrewvc still that doesn't suffice. You'll have to completely revamp `ConvertedMap` to not leak interned `String`. Just making that change here will only make us leak "less" at a price performance wise. Maybe it's better to make these changes when we have the overall solution without interned `String` so we don't create noise without knowing the final direction of things?

andrewvc:
@original-brownbear would you be able to open a new issue detailing what we'd need to prevent ourselves leaking memory in this case?

original-brownbear:
@andrewvc sure I can do that tomorrow :) Put it on my todo list

andrewvc:
I'm going to close this issue then.

andrewvc:
@original-brownbear if you could link back to this once you have that issue tomorrow that'd be great

original-brownbear:
Link #9136 :) Your feedback is required in that issue btw :)

PR #9133, opened 2018-02-13
Windows freespace on SUBST'ed drives

### LINKED ISSUE #9123: Test failure: Master and 6.x failing for since January 24
https://logstash-ci.elastic.co/job/elastic-logstash-6.x/ and https://logstash-ci.elastic.co/job/elastic-logstash-master/

Some of the dependent jobs are succeeding, but at least one has failed every time for the past 15 days.

colinsurprenant:
Fixes #9123 

On Windows, SUBST'ed drives report `0L` as `getFreeSpace()`. 

API doc says:
```
The number of unallocated bytes on the partition or 0L if the abstract pathname does not name a partition [...]
```

There is no straightforward fix for this and it seems a fix is included in Java 9. 
One alternative is to launch and parse a DIR command and look at the reported free space.

This is a temporary fix to get the CI tests going which relies on SUBST'ed drives to manage long paths.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/common/FsUtil.java
+++ b/logstash-core/src/main/java/org/logstash/common/FsUtil.java
@@ -5,6 +5,8 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 /**
  * File System Utility Methods.
@@ -13,6 +15,9 @@ public final class FsUtil {
 
     private FsUtil() {
     }
+
+    private static final boolean IS_WINDOWS = System.getProperty("os.name").startsWith("Windows");
+    private static final Logger logger = LogManager.getLogger(FsUtil.class);
 
     /**
      * Checks if the request number of bytes of free disk space are available under the given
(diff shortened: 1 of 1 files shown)


logstash-core/src/main/java/org/logstash/common/FsUtil.java
```diff
@@ -36,6 +41,18 @@ public static boolean hasFreeSpace(final String path, final long size)
             }
         }
 
+        final long freeSpace = location.getFreeSpace();
+
+        if (freeSpace == 0L && IS_WINDOWS) {
```
danhermann: Is it worth adding a check here for Java versions <9 so that the free space test is re-enabled when run on Java versions with the fix for SUBST drives?

colinsurprenant: good idea - but first need to validate that it is indeed fixed in Java 9. Saw a JVM issue about it but did not go as far as testing it on Java 9.

colinsurprenant: @danhermann I just rana test on Java 9 and free space is still reported a `0L` on SUBST'ed drive so I think we should move forward, and I'll create a followup issue to see if we need to add and alternate free space strategy on Windows.

danhermann: Cool, sounds good to me

danhermann:
LGTM

colinsurprenant:
merged in master and 6.x
6.x commit is 5057e1be15f097026b4ad17ca6fabe02a58476c1

PR #9134, opened 2018-02-13
Added java.lang.Byte and java.lang.Short Valuefier support

Twinkiee:
Faced this issue when dequeuing messages from an IBM MQ queue.
I added a couple more converters for java.lang.Byte and java.lang.Short which happened to be produced when the Logstash JMS plugin was deserializing the message headers.
Could not provide a meaningful (i.e. not trivial) unit test but I'd add one if asked/required.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Valuefier.java
+++ b/logstash-core/src/main/java/org/logstash/Valuefier.java
@@ -112,6 +112,8 @@ public final class Valuefier {
         converters.put(RubyBignum.class, IDENTITY);
         converters.put(RubyBigDecimal.class, IDENTITY);
         converters.put(String.class, input -> RubyUtil.RUBY.newString((String) input));
+        converters.put(Byte.class, input -> RubyUtil.RUBY.newString(input.toString()));
+        converters.put(Short.class, LONG_CONVERTER);
         converters.put(Float.class, FLOAT_CONVERTER);
         converters.put(Double.class, FLOAT_CONVERTER);
         converters.put(


guyboertje:
We have a serious conflict of interest here.

On the one hand we can't introduce valuefier convertors for every possible type that may enter Logstash from the various technologies used by plugins when they create a field/value combination. This implies that it is up to the plugin to do the normalisation, for example, from JDBC sql types.

On the other hand, given that these are standard Java types, Byte and Short should probably be added.

@original-brownbear WDYT?

guyboertje:
Jenkins please test this

logstash-core/src/main/java/org/logstash/Valuefier.java
```diff
@@ -112,6 +112,8 @@ private static Object fallbackConvert(final Object o, final Class<?> cls) {
         converters.put(RubyBignum.class, IDENTITY);
         converters.put(RubyBigDecimal.class, IDENTITY);
         converters.put(String.class, input -> RubyUtil.RUBY.newString((String) input));
+        converters.put(Byte.class, input -> RubyUtil.RUBY.newString(input.toString()));
```
original-brownbear: @Twinkiee why don't we cast to `int` here and use `LONG_CONVERTER` as well? Seems more logical to me since a `byte` is a numeric type and not necessarily equivalent to a `char` in a `String`.

guyboertje: I agree with @original-brownbear. At the local Event level a byte is a number.

However, I looked at the IBM MQ Java docs. I wanted to know if the byte is considered a ASCII character or a numeric value.

This header [MQIIH](https://www.ibm.com/support/knowledgecenter/en/SSFKSJ_9.0.0/com.ibm.mq.javadoc.doc/WMQJavaClasses/com/ibm/mq/headers/MQIIH.html) can return a char for `TranState`. 
I saw others that returned `byte[]`.
Via the CodedCharSetId, messages can be in UTF8 or CCS-2 (and variants).

@Twinkiee 
Given that, at the event level, we don't know what the byte represents, it seems like the jms plugin is a better place to code for this.

Twinkiee: @original-brownbear @guyboertje 
It seems reasonable to me as well to convert the `Byte` as `int` as a general rule. That's how the `java.lang.Byte#toString()` threats the internal value anyway so there's really no point to behave differently.
I could've coded it in the first place, my bad.

On the other hand I'll try to investigate if it's actually possible for the IBM MQ to return a `byte` which is actually a `char`.
As far as I know, that's not the case. Sorry if it caused any confusion.

Twinkiee: @guyboertje 
Hey, sorry for the late  reply.
No, by my tests and investigations I couldn't find a byte converted into a char.

guyboertje: @Twinkiee 
Please make the update that @original-brownbear [suggests](https://github.com/elastic/logstash/pull/9134/files#r169007943) then we can move this forward.

PR #9139, opened 2018-02-14
Add conditional coding to set correct default for codec

dedemorton:
Fixes logstash-plugins/logstash-output-redis#47 and logstash-plugins/logstash-input-redis#57

If there are other plugins (more than 2 or 3) that use json for the default codec, I'll need to change my strategy and create an asciidoc attribute to set the codec.

This file needs to be copied to the versioned plugin reference after the changes are merged.

### DIFF
--- a/docs/include/input.asciidoc
+++ b/docs/include/input.asciidoc
@@ -56,7 +56,12 @@ endif::[]
 ===== `codec`
 
   * Value type is {logstash-ref}/configuration-file-structure.html#codec[codec]
+ifdef::default_codec[]
+  * Default value is +"{default_codec}"+
+endif::[]
+ifndef::default_codec[]
   * Default value is `"plain"`
+endif::[]
 
 The codec used for input data. Input codecs are a convenient method for decoding your data before it enters the input, without needing a separate filter in your Logstash pipeline.
 
--- a/docs/include/output.asciidoc
+++ b/docs/include/output.asciidoc
@@ -34,7 +34,12 @@ endif::[]
 ===== `codec`
 
   * Value type is {logstash-ref}/configuration-file-structure.html#codec[codec]
+ifdef::default_codec[]
+  * Default value is +"{default_codec}"+
+endif::[]
+ifndef::default_codec[]
   * Default value is `"plain"`
+endif::[]
 
 The codec used for output data. Output codecs are a convenient method for encoding your data before it leaves the output without needing a separate filter in your Logstash pipeline.
 


dedemorton:
@jakelandis Are there any other plugins (besides redis) that use json as the default?

jakelandis:
@dedemorton 
visually inspecting the first few pages of this search:
https://github.com/search?p=9&q=org%3Alogstash-plugins+default+%3Acodec%2C&type=Code

I am not sure we can make any broad claims about the default codec.  It is technically true that "plain" is the default...but only if the plugin doesn't define it's own.  

json codec
-----
input rabbit mq 
output rabbit mq
output udp
input web socket
output rackspace
integration zero mq 

line codec
------
input stadard io
input tcp
output webhdfs

dedemorton:
@jakelandis  Not quite sure the best way to proceed here. We could use an asciidoc attribute in each file to set whether the plugin uses plain or json (and modify the conditional coding in this PR to resolve the correct default). That's going to be a bit of work, though, because it touches a lot of plugins. If we do it, it should be a bulk update and push to all the repos.

The alternative is to say something like "The default varies by plugin" (but that's going to look weird for users who are looking at a specific plugin."

We could say that "There is no default" but that would be incorrect. 

WDYT?

jakelandis:
@dedemorton - I think the only correct solution is allow each plugin to define their own default. It would result in some duplication, between the plugin specific section and common section. Where the common section defines it as defined by the plugin, and the each plugin re-defines the codec with it's default. 

It's alot of work, but the only way I see being and staying correct.

dedemorton:
@jakelandis Actually, after thinking about this for a bit, I realized that we can avoid duplicating the desription by using an asciidoc attribute to set the value. So at the top of each file, you would have:

:default_codec: plain

or

:default_codec: json

And the doc would say something like:

* Default value is +"{codec}"+

We already use the [docgen script](https://github.com/elastic/docs-tools/blob/master/plugindocs.rb) to values in the generated plugin docs. I'm wondering how hard it would be to inspect the ruby source files for the default codec setting (`default :codec, "plain"`, I guess?) and set the generated docs accordingly? probably simpler than updating all the repos.

dedemorton:
@jsvd ^^ Thought you might have some thoughts on how to set the default coded correctly in the docs for various plugins.

jakelandis:
@dedemorton - that is cool and should work (I have no clue about ascii docs capabilities, but i trust you do !)

w/r/t to updating the all the plugins, it would probably faster to brute force it. Probably take a human ~1-2 hours to manually do this.  Do we have an intern ? (jk i can do the grunt work unless @jsvd has some magic)

jsvd:
PH has build a little tool to download all plugins and perform tasks on the pool of git repos: https://github.com/ph/logstash-mass_effect

It should be easy to create a ruby script that takes a plugin git repo, find the source code and the default plugin directive, parses it and injects the asciidoc attribute setting in  the index.asciidoc

@jakelandis unless you want to do it for..educational purposes, I'll take this

jakelandis:
@jsvd - I've used mass_effect before to update the test matrix. 

> It should be easy to ...

Did you mean deceptively difficult ?  :)

>  I'll take this

Please.

jsvd:
I built a script that, for inputs and outputs, finds the default codec directive and injects it as a doc attribute in the asciidoc. here's an example PR: https://github.com/logstash-plugins/logstash-input-rabbitmq/pull/111/files

Is this what we want @dedemorton ?

dedemorton:
@jsvd Yup! Then we'll just need to:

- Update docs/include/input.asciidoc and output.asciidoc (I assume) to use the new attribute: `Default value is +{default_codec}+`
- Copy the new version of input.asciidoc and output.asciidoc to logstash-docs/tree/versioned_plugin_docs/docs/versioned-plugins/include - we need to decide if we want to create a new folder for the changed files and update the docgen script to set a new path for includes, or "fix" older versions of the plugin docs so that they use the asciidoc attribute.

jsvd:
@dedemorton cool. Could we assume that the lack of the doc attribute 'default_codec' means the default codec is `plain`? that'd would mean we only need to publish around 10 plugins - the ones defaulting to json or line - instead of all 200.

dedemorton:
@jsvd Yah, instead of resolving the `default_codec` value directly, we can use the attribute to set a conditional block in the common files. If we do this, we won't have to version the common files because they will continue to work even if `default_codec` isn't defined in a file (the default will display as plain).  If we want to fix the older versions of the docs, we'll have to regenerate them and add the default codec attribute where needed.

So to make this work, I'll need to change the common files to say:

```
===== `codec`

  * Value type is {logstash-ref}/configuration-file-structure.html#codec[codec]
ifeval::["{default_codec}"=="json"]
  * Default value is `"json"`
endif::[]
ifndef::default_codec[]
  * Default value is `"plain"`
endif::[]

````

In any asciidoc file where you set `default_codec`, you'll need to add a line to the *end* of the file to undefine the attribute. So the end will look something line:

```
[id="{version}-plugins-{type}s-{plugin}-common-options"]
include::{include_path}/{type}.asciidoc[]

:default_codec!: 
````

**Comment by jsvd** [comment]
> In any asciidoc file where you set default_codec, you'll need to add a line to the end of the file to undefine the attribute. So the end will look something line:

Ok I've updated [the example PR](https://github.com/logstash-plugins/logstash-input-rabbitmq/pull/111) with the unset of the doc attribute.

```
ifeval::["{default_codec}"=="json"]
  * Default value is `"json"`
endif::[]
```

This is sad:( is there no way to "interpolate" the value? something like:

```
:even_more_default_codec: plain

ifdef::default_codec[]
  * Default value is `"{default_codec}."`
endif::[]
ifndef::default_codec[]
  * Default value is `"even_more_default_codec"`
endif::[]
```

**Comment by dedemorton** [comment]
@jsvd You could define `:even_more_default_codec: plain` in the index file where we include the plugins and do something like:

```
===== `codec`

  * Value type is {logstash-ref}/configuration-file-structure.html#codec[codec]
ifeval::["{default_codec}"=="json"]
  * Default value is `"json"`
endif::[]
ifndef::default_codec[]
  * Default value is +"{even_more_default_codec}"+
endif::[]
```
Wondering if this would gain us much, tho.

**Comment by jsvd** [comment]
@dedemorton I guess my question was more two questions.

1. Could we avoid having the pattern of explicit codec strings:

```
ifeval::["{default_codec}"=="json"]
  * Default value is `"json"`
endif::[]
ifeval::["{default_codec}"=="plain"]
  * Default value is `"plain"`
endif::[]
```

and have just the interpolated value:

```
ifdef::default_codec[]
  * Default value is `"{default_codec}."`
endif::[]
```

Otherwise if we have 10 potential default codecs we need to have 10 ifeval clauses?

2. Should we abstract away the global_default_codec into its own attribute?

Like you said, "Wondering if this would gain us much, tho.", probably not :D

**Comment by dedemorton** [comment]
@jsvd You _can_ do that! I'm just being dense. If the default is plain, we can use:

```
ifdef::default_codec[]
  * Default value is +"{default_codec}"+
endif::[]
ifndef::default_codec[]
  * Default value is `"plain"`
endif::[]
```

dedemorton:
Or we can use an attribute for `even_more_default_codec` (like you suggested earlier). Sheesh. Sorry to waste your time. I'm going to blame vacation brain because I didn't see your example earlier.

dedemorton:
@jsvd OK, I've updated the conditional coding to use the interpolated value. This coding doesn't rely on having the `default_codec` attribute defined anywhere, so we can merge it before you run the scrip to update the plugin docs.

jsvd:
@dedemorton good news and bad news:
Good news - I'm now running a script that applies the doc attribute and publishes the plugin
Bad news - I f'ed up the logic to skip the change when the plugin uses the 'plain' default codec. So I'll be publishing pretty much all inputs/outputs.

PR #9141, opened 2018-02-14
#9136 Put upper limit of FieldReference cache size

### LINKED ISSUE #9136: FieldReferences are cached indefinitely, making an infinite space of event field names impossible
Currently, we do not support an infinite number of field names since all `FieldReference` instances we create are cached indefinitely. They are cached in a very memory efficient way, but still an infinite number of `FieldReference` is impossible.

We want to be able to support dynamically named fields that could theoretically go through an infinite number of field names. This requires the following changes for a minimal and performance-degrading solution:

1. All `FieldReference` need to be compared via `.equals` instead of `==` since a finite lifetime for a `FieldReference` will allow for duplicates (unless we cache `FieldReference` instances as weak or soft  references, which is an option)
2. The internal cache for `FieldReference` should be finite in size and be evicted above a certain size threshold.
3. Currently, all `String` in `FieldReference` instances are `intern`ed. We have to stop `intern`ing them to not leak memory via the `String` space.
   * This requires adjusting `ConvertedMap` to not assume all keys in it to be `intern`ed `String`!

This would be the minimum action to take here. Doing this in a naive way as outlined above will have a detrimental effect on performance for sure I'm afraid.
A better solution would probably be to:

1. All `FieldReference` need to be compared via `.equals` instead of `==` since a finite lifetime for a `FieldReference` will allow for duplicates (unless we cache `FieldReference` instances as weak or soft  references, which is an option)
2. Simply stop caching `FieldReference` once the cache has grown to `10k` (or even more imo, the memory need here is trivial) references and simply parse additional references every time.
3. Stop interning `String` for uncached `FieldReference` (need 2 paths for creating a `FieldReference`)
4. Start creating`Event`s backed by `HashMap` instead of `IdentityHashmap` once we hit more than the threshold amount of `FieldReference` in `2.` since we now don't have `intern`ed `String` to work from.

This would be a little harder to implement since it only requires switching the `Event` generation path once we jump above a certain number of `FieldReference` but would not negatively affect performance for existing users. I'm in favor of investing the extra few hours of going for the complicated solution here.
Let me know what you want to go for here @andrewvc :)

original-brownbear:
Fixes #9136 by simply only caching and deduping `FieldReference` when the cache size is below 10k `FieldReferences`.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/FieldReference.java
+++ b/logstash-core/src/main/java/org/logstash/FieldReference.java
@@ -139,8 +139,11 @@ public final class FieldReference {
     }
 
     private static FieldReference parseToCache(final CharSequence reference) {
-        final FieldReference result = parse(reference);
-        CACHE.put(reference, result);
+        FieldReference result = parse(reference);
+        if (CACHE.size() < 10_000) {
+            result = deduplicate(result);
+            CACHE.put(reference, result);
+        }
         return result;
     }
 
@@ -169,11 +172,11 @@ public final class FieldReference {
         if (empty && key.equals(Event.METADATA)) {
             return METADATA_PARENT_REFERENCE;
         } else if (!empty && path.get(0).equals(Event.METADATA)) {
-            return deduplicate(new FieldReference(
-                path.subList(1, path.size()).toArray(EMPTY_STRING_ARRAY), key, META_CHILD));
+            return new FieldReference(
+                path.subList(1, path.size()).toArray(EMPTY_STRING_ARRAY), key, META_CHILD
+            );
         } else {
-            return deduplicate(
-                new FieldReference(path.toArray(EMPTY_STRING_ARRAY), key, DATA_CHILD));
+            return new FieldReference(path.toArray(EMPTY_STRING_ARRAY), key, DATA_CHILD);
         }
     }
 }
--- a/logstash-core/src/main/java/org/logstash/ext/JrubyEventExtLibrary.java
+++ b/logstash-core/src/main/java/org/logstash/ext/JrubyEventExtLibrary.java
@@ -86,7 +86,7 @@ public final class JrubyEventExtLibrary {
         public IRubyObject ruby_set_field(ThreadContext context, RubyString reference, IRubyObject value)
         {
             final FieldReference r = FieldReference.from(reference.getByteList());
-            if (r  == FieldReference.TIMESTAMP_REFERENCE) {
+            if (r.equals(FieldReference.TIMESTAMP_REFERENCE)) {
                 if (!(value instanceof JrubyTimestampExtLibrary.RubyTimestamp)) {
                     throw context.runtime.newTypeError("wrong argument type " + value.getMetaClass() + " (expected LogStash::Timestamp)");
                 }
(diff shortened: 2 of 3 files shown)


logstash-core/src/main/java/org/logstash/FieldReference.java
```diff
@@ -139,8 +139,11 @@ private static int calculateHash(final String key, final String[] path, final in
     }
 
     private static FieldReference parseToCache(final CharSequence reference) {
-        final FieldReference result = parse(reference);
-        CACHE.put(reference, result);
+        FieldReference result = parse(reference);
+        if (CACHE.size() < 10_000) {
```
andrewvc: Can add a test for this behavior?

original-brownbear: @andrewvc Just run up 10k `FieldReference`s and check that `==` stops holding for `FieldRefernce` no. 10001?

logstash-core/src/main/java/org/logstash/FieldReference.java
```diff
@@ -139,8 +139,11 @@ private static int calculateHash(final String key, final String[] path, final in
     }
 
     private static FieldReference parseToCache(final CharSequence reference) {
```
yaauie: nitpick: method name is no longer ALWAYS accurately descriptive; if we brought this decision up to `FieldReference#from(CharSequence)`, we could maintain a little more clarity.

~~~ java
     public static FieldReference from(final CharSequence reference) {
         // atomicity between the get and put is not important
         final FieldReference result = CACHE.get(reference);
         if (result != null) {
             return result;
         }
         if (CACHE.size() <= 10_000) {
             return parseToCache(reference);
         }
         return parse(reference);
     }
~~~

Which simplifies this method to:

~~~ java
    private static FieldReference parseToCache(final CharSequence reference) {
        final FieldReference result = deduplicate(parse(reference));
        CACHE.put(reference, result);
        return result;
    }
~~~

original-brownbear: @yaauie I'd rather not do that, it increases the compiled size of `from` in the default use case while if we keep the check outside the conditional in the other method, the use case that doesn't exceed 10k cache size will eventually eliminate that code and inline better upstream right?

logstash-core/src/main/java/org/logstash/FieldReference.java
original-brownbear: @andrewvc actually come to think about it ... this is kinda dirty to add a test for since the cache is `static` => The exact number of field references to put in here until the cache shuts down depends on the test order.
Plus (this isn't so relevant but still ...), this would actually make all tests that run after such a test slower since they're not caching anymore :P

original-brownbear:
@andrewvc added the test I hate the least https://github.com/elastic/logstash/pull/9141/commits/a8bd4bc3da563634b5c9281ccddafc91542cc4a9 :)

Confirmed that this fails on `master` as well :)

original-brownbear:
@andrewvc ping :)

original-brownbear:
@andrewvc pong :)

original-brownbear:
@andrewvc ping ding :D

PR #9142, opened 2018-02-14
pre-release gem bump for 6.2.2

danhermann:


andrewvc:
Nothing critical here. Let's close this given that the release BC window has passed.

PR #9143, opened 2018-02-14
Fix/9071

guyboertje:
To try to mitigate the problems users have when the a service restarts continually.

I agonized whether to use a multiline log message of not. As it is, the call to action is easier to read.

### DIFF
--- a/logstash-core/lib/logstash/runner.rb
+++ b/logstash-core/lib/logstash/runner.rb
@@ -253,6 +253,15 @@ class LogStash::Runner < Clamp::StrictCommand
       return 0
     end
 
+    # Starting up log message. This will repeat if the service manager is continually
+    # restarting Logstash due to a ConfigurationError being raised
+    logger.info <<~EOT
+      Logstash is starting. It will try to read and parse various settings files configuration files.
+        If you see this message repeated with little else in between then there could
+        be a syntax error in one or more of these files and the service manager is
+        restarting logstash. Look for more info in the service manager logs.
+    EOT
+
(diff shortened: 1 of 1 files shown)


logstash-core/lib/logstash/runner.rb
```diff
@@ -253,6 +253,13 @@ def execute
       return 0
     end
 
+    # Starting up log message. This will repeat if the service manager is continually
+    # restarting Logstash due to a ConfigurationError being raised
+    logger.info("Configuring Logstash by trying to read and parse various settings and configuration files\n" +
```
yaauie: Would it be possible to simply catch the cases where we have a configuration error and log a helpful message before propagating it to process death? It seems odd to me that we're logging a message that calls out a "probably" scenario that requires a user to coalate many log messages together.

guyboertje: Agreed, we should find the raise ConfigurationError cases and log them but his log message, as is, has a dual purpose.
First, to act as the first message logged by LS on start up with no other intention than that, allowing a user to clearly deduce that the service manager is restarting LS and not some weird internal loop that is repeatedly logging out the same statements - the deduction being made now.
Second, to hint where more info may be found.

I agree, that when **all** the ConfigurationErrors have been logged we can remove the `If you see this message repeated with little else in between then there could be a syntax error in one or more of these files and the service manager is restarting logstash` part of this log message.

guyboertje:
@yaauie Ping
IMO we can merge this as is and circle back on the wording once all the ConfigurationErrors are logged before raising them.

PR #9149, opened 2018-02-15
ensure singleton `TIMESTAMP_REFERENCE` `FieldReference`

yaauie:
Aligns how we do deduplication within `FieldReference`, ensuring that we never
generate `FieldReference` objects that are equivalent to `TIMESTAMP_REFERENCE`
in a manner that reflects the strategy taken with `META_PARENT_REFERENCE`.

In practice, this is not much of a net gain; with our current caching strategy
(cache everything), we'll only fall through the cache at most twice (once for
each equivalent string reference `@timestamp` and `[@timestamp]`), so it saves
us from a `DEDUP`-lookup and superfluous construction twice.

However, as we consider other caching strategies, this change will do no-harm
and could save us from some surprises.

Discovered while reviewing https://github.com/elastic/logstash/pull/9131
Complements cache-strategy changes in https://github.com/elastic/logstash/pull/9141

### DIFF
--- a/logstash-core/src/main/java/org/logstash/FieldReference.java
+++ b/logstash-core/src/main/java/org/logstash/FieldReference.java
@@ -35,7 +35,7 @@ public final class FieldReference {
      * Unique {@link FieldReference} pointing at the timestamp field in a {@link Event}.
      */
     public static final FieldReference TIMESTAMP_REFERENCE =
-        deduplicate(new FieldReference(EMPTY_STRING_ARRAY, Event.TIMESTAMP, DATA_CHILD));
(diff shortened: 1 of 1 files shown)


original-brownbear:
@yaauie not sure it's a good idea to add a special case for this key here, it just adds complexity to the already complex parser.
The gain from not deduping here as you point out is irrelevant since this is only run during class loading anyways.

The reason `METADATA_PARENT_REFERENCE ` takes a different code path is that its behaviour is actually different from other `FieldReference` (since it has no keys and cannot be parsed through the normal path for that reason).

Imo we shouldn't up the complexity here. Also, `TIMESTAMP_REFERENCE` is no different than any other `FieldReference` that you would cache in a `static` field. This change suggests it is imo.

PR #9156, opened 2018-02-16
Fix overview list and add ID

dedemorton:
After merging these changes, I'll have the web team set up redirects for master and all the 6.x-related branches so that: 

https://www.elastic.co/guide/en/logstash/6.0/_configuring_logstash_for_docker.html

redirects to: 

https://www.elastic.co/guide/en/logstash/6.0/_configuring_logstash_for_docker/docker-config.html

@karenzone Here's an example where someone added a level 3 heading without specifying an ID. Because no ID is specified, the name of the generated HTML page is based on the heading text. This is bad juju because any changes to the header text will result in the HTML filename changing (bad for SEO and switching between versions on doc pages). Plus, whenever a URL changes, we have to ask the web team to set up redirects, which means extra work for us and them. :-(

The way to avoid this is to make sure all level 3 headings have an ID. Try to look out for this during reviews and point it out so people learn to do it right. :-)

### DIFF
--- a/docs/static/docker.asciidoc
+++ b/docs/static/docker.asciidoc
@@ -42,6 +42,7 @@ docker pull {docker-repo}-oss:{logstash_version}
 
 endif::[]
 
+[[docker-config]]
 === Configuring Logstash for Docker
 
 Logstash differentiates between two types of configuration:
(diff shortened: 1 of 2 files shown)


dedemorton:
Requested redirects here: https://github.com/elastic/website-www.elastic.co/issues/2331

Issue #7859, opened 2017-08-01
tags collision with user-defined tags and built-in tags


dukenguyen:
i have an event that has json content that looks like the following:

`{"tags": {"key1":"value1", "key2":"value2"}}`

Logstash is incorrectly assuming this is the built-in 'tags' field and putting it into a json list.  How do I avoid this so that I can use my own event that happens to have 'tags' as an attribute (that is not a json list)?

jsvd:
you must receive the event as string using something like the plain or line codec, then use the json filter, since that filter supports the [target parameter](https://www.elastic.co/guide/en/logstash/current/plugins-filters-json.html#plugins-filters-json-target)

dukenguyen:
@jsvd thanks for the reply, but what if I want the root element to be 'tags' when i output it to elasticsearch?  I'm aware i can wrap it in the target parameter e.g.

`{"doc": {"tags":{"key1":"value1","key2":"value2"}}}`

but i need the end result in elasticsearch to have 'tags' at the root level.

guyboertje:
So the field name `tags` is a treated as a reserved field name throughout the LS code base.

The only way I can think of to preserve your tags Map/Hash like object is to use: 
```
add_field => "[mytags]" => "[%{tags}]"
remove_field  => ["tags"]
```
in the input, then do the reverse in the last filter before the output:
```
add_field => "[tags]" => "[%{mytags}]"
remove_field  => ["mytags"]
```
Obviously, any filters will need to manipulate `mytags` instead of `tags`.

dukenguyen:
unfortunately the above still doesn't help.  As soon as you try to use the 'tags' field again (even if its the last filter), it will still be of type json list.  The goal is to be able to send the user-defined 'tags' field to the output.

guyboertje:
I don't know what you mean by `be of type json list`. JSON is a text representation. LS Events are object based. Do you mean that in the `tags` field you have a JSON String representation of some other objects.

dukenguyen:
type json list just means a json array e.g.

`{ "tags" : ["tag1", "tag2", "tag3"] }`

https://www.w3schools.com/js/js_json_arrays.asp

As stated in my original comment, the 'tags' attribute that I need to flow thru to the output has to be a json object, not an array (list).

guyboertje:
@dukenguyen 
The info in that link is only valid in a Javascript execution environment.
[...]

PR #9166, opened 2018-02-19
Fix: Tags collision with user-defined tags and built-in tags

guyboertje:
See Issue #7859 

PR #6177 Introduced a change where the value of the `tags` field is always converted to an Array even if no filters added or removed tags at all.

Users want a mechanism to have a tags field that comes from, say JSON, rename it early, allow other filters to add tags (grok), use logic on the tags to process failures and then rename their external tags field back to tags just before the outputs.

### DIFF
--- a/logstash-core/spec/logstash/filters/base_spec.rb
+++ b/logstash-core/spec/logstash/filters/base_spec.rb
@@ -309,4 +309,17 @@ describe LogStash::Filters::NOOP do
       reject { subject }.include?("go")
     end
   end
+
+  describe "when neither add_tag not remove_tag is specified, the tags field is left untouched" do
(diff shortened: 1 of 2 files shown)


logstash-core/spec/logstash/filters/base_spec.rb
```diff
@@ -309,4 +309,17 @@ def filter(event)
       reject { subject }.include?("go")
     end
   end
+
+  describe "when neither add_tag not remove_tag is specified, the tags field is left untouched" do
```
tsg: s/neither add_tag not remove_tag/neither add_tag nor remove tag/ ?

colinsurprenant:
LGTM

PR #9171, opened 2018-02-20
Add doc for LS-to-LS communication

karenzone:
First review of config from Jordan's video.
What about Windows?

### DIFF
--- a/docs/static/ls-ls-config.asciidoc
+++ b/docs/static/ls-ls-config.asciidoc
@@ -60,6 +60,7 @@ You can demo the configuration by running two Logstash instances on one computer
 ===== Start the upstream Logstash machine
 
 On the upstream Logstash machine, run the following command:
+
 [source,shell]
 ----
 bin/logstash -e 'input { beats { codec => json port => 5000 ssl_certificate => "lumberjack.cert" ssl_key => "lumberjack.key"} }'
--- a/docs/static/ls-ls-config.asciidoc
+++ b/docs/static/ls-ls-config.asciidoc
@@ -58,10 +58,10 @@ Start Logstash and generate test events:
 
 [source,shell]
 ----
-bin/logstash -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
+bin/logstash -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "mydownstreamhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
 ----
 
-This sample command sends five events to localhost:5000 using the SSL certificate provided.
+This sample command sends five events to mydownstreamhost:5000 using the SSL certificate provided.
 
 [[save-cert-ls2]]
 ===== Start the downstream Logstash instance
--- a/docs/static/ls-ls-config.asciidoc
+++ b/docs/static/ls-ls-config.asciidoc
@@ -24,14 +24,14 @@ Run the following command:
 
 [source,shell]
 ----
-openssl req -x509 -batch -nodes -newkey rsa:2048 -keyout lumberjack.key -out lumberjack.cert -subj /CN=mylocalhost
+openssl req -x509 -batch -nodes -newkey rsa:2048 -keyout lumberjack.key -out lumberjack.cert -subj /CN=localhost
 ----
 
 where:
 
 * `lumberjack.key` is the name of the SSL key to be created
 * `lumberjack.cert` is the name of the SSL certificate to be created
-* `mylocalhost` is the name of the upstream Logstash computer
+* `localhost` is the name of the upstream Logstash computer
 
 
 This command produces output similar to the following:
@@ -58,10 +58,10 @@ Start Logstash and generate test events:
 
 [source,shell]
 ----
-bin/logstash -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "mylocalhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
+bin/logstash -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
 ----
 
-This sample command sends five events to mylocalhost:5000 using the SSL certificate provided.
+This sample command sends five events to localhost:5000 using the SSL certificate provided.
 
 [[save-cert-ls2]]
 ===== Start the downstream Logstash instance
@@ -70,7 +70,7 @@ Start the downstream instance of Logstash:
 
 [source,shell]
 ----
-bin/logstash -e 'input { beats { codec => json port => 5000 ssl_certificate => "lumberjack.cert" ssl_key => "lumberjack.key"} }'
+bin/logstash -e 'input { beats { codec => json port => 5000 ssl => true ssl_certificate => "lumberjack.cert" ssl_key => "lumberjack.key"} }'
 ----
 
 This sample command sets port 5000 to listen for incoming Beats input.
--- a/docs/static/ls-ls-config.asciidoc
+++ b/docs/static/ls-ls-config.asciidoc
@@ -11,8 +11,8 @@ Use the Lumberjack protocol to connect two Logstash machines.
 . Generate a trusted SSL certificate (required by the lumberjack protocol).
 . Save the SSL certificate on the upstream Logstash machine.
 . Save the SSL certificate and key on the downstream Logstash machine.
-. Set the downstream Logstash instance to listen for incoming Lumberjack connections through the Beats input.
-. Set the upstream Logstash instance to use the Lumberjack output to send data.
+. Set the downstream Logstash machine to listen for incoming Lumberjack connections through the Beats input.
+. Set the upstream Logstash machine to use the Lumberjack output to send data.
 . Test it.
 
 [[generate-self-signed-cert]]
@@ -57,49 +57,41 @@ Copy the SSL certificate and key to the downstream Logstash machine.
 You can demo the configuration by running two Logstash instances on one computer.
 
 [[save-cert-ls1]]
-===== Start the upstream Logstash machine
+===== Start the upstream Logstash instance
 
-On the upstream Logstash machine, run the following command:
+Run the following command to start Logstash and generate test events:
 
 [source,shell]
 ----
-bin/logstash -e 'input { beats { codec => json port => 5000 ssl_certificate => "lumberjack.cert" ssl_key => "lumberjack.key"} }'
+bin/logstash -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
 ----
 
 where:
 
-* `5` specifies the number of events you want to generate and send
-* `lumberjack.cert` specifies the path and name of the SSL certificate you created
-* `localhost` specifies the address of the downstream Logstash computer
+* `5` specifies the number of events to be generated
+* `localhost` specifies the name of the downstream computer
+* `lumberjack.cert` specifies the path and name of the SSL certificate
+* `5000` specifies the port number used to send the events
 
 [[save-cert-ls2]]
-===== Start the downstream Logstash machine
+===== Start the downstream Logstash instance
 
-On the downstream Logstash machine, run the following command:
+Run the following command to start the downstream instance of Logstash:
 
 [source,shell]
 ----
-bin/logstash --path.data /tmp/z -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
+bin/logstash --path.data /tmp/z -e 'input { beats { codec => json port => 5000 ssl_certificate => "lumberjack.cert" ssl_key => "lumberjack.key"} }'
 ----
 
 where:
 
-* `--path.data /tmp/z -e` sets a different data path (required to start a second Logstash on the same computer)
-* `lumberjack.key` specifies the path and name of the SSL key
+* `--path.data /tmp/z` sets a different data path (required to start a second Logstash instance on the same computer)
+* `5000` specifies the port number listening for incoming events
 * `lumberjack.cert` specifies the path and name of the SSL certificate
-* `localhost` specifies the name of the computer
+* `lumberjack.key` specifies the path and name of the SSL key
 
 [[test-ls-to-ls]]
-===== Test the communication
-
-Use the input generator to send test events from the upstream Logstash machine.
-
-The following command sends five events over port 5000:
-
-[source,shell]
-----
-bin/logstash —path.data /tmp/z -e 'input { generator {count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" Port => 5000 } }'
-----
+===== Check the communication
 
 Watch the downstream Logstash machine for the incoming events. You should see five incrementing events similar to the following:
 
@@ -116,4 +108,4 @@ Watch the downstream Logstash machine for the incoming events. You should see fi
   "host"       => "ls1.semicomplete.com"
 }
 ----
-If you see all five events with consistent fields and formatting, incrementing by one, then your configuration is correct.
+If you see all five events with consistent fields and formatting, incrementing by one, then your demo is successful.
--- a/docs/static/ls-ls-config.asciidoc
+++ b/docs/static/ls-ls-config.asciidoc
@@ -9,29 +9,29 @@ You can set up communication between two Logstash machines by connecting the Lum
 Use the Lumberjack protocol to connect two Logstash machines.
 
 . Generate a trusted SSL certificate (required by the lumberjack protocol).
-. Save the SSL certificate on the upstream Logstash machine.
-. Save the SSL certificate and key on the downstream Logstash machine.
-. Set the downstream Logstash machine to listen for incoming Lumberjack connections through the Beats input.
+. Copy the SSL certificate to the upstream Logstash machine.
+. Copy the SSL certificate and key to the downstream Logstash machine.
 . Set the upstream Logstash machine to use the Lumberjack output to send data.
+. Set the downstream Logstash machine to listen for incoming Lumberjack connections through the Beats input.
 . Test it.
 
 [[generate-self-signed-cert]]
 ===== Generate a self-signed SSL certificate and key
 
-You can use the `openssl req` command to generate a self-signed certificate and key. The `openssl req` command is available on any operating system that comes with SSL.
+Use the `openssl req` command to generate a self-signed certificate and key. The `openssl req` command is available with some operating systems. You may need to install the openssl command line program for others.
 
 Run the following command:
 
 [source,shell]
 ----
-openssl req -x509 -batch -nodes -newkey rsa:2048 -keyout lumberjack.key -out lumberjack.cert -subj /CN=localhost
+openssl req -x509 -batch -nodes -newkey rsa:2048 -keyout lumberjack.key -out lumberjack.cert -subj /CN=mylocalhost
 ----
 
 where:
 
-* `lumberjack.key` specifies the name of the SSL key to be created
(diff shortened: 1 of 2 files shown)


docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
+[[ls-to-ls]]
+=== Configuring Logstash-to-Logstash Communication
+
+In a complex network environment, you may need two Logstash machines talking to each other. That is, the output from the first Logstash machine becomes the input for the second Logstash machine. You may need this configuration if the data path crosses network or firewall boundaries, for example.
+
+NOTE: Don't complicate your configuration unnecessarily. If you don't have a compelling need for Logstash-to-Logstash communication, then don't implement it.
```
andrewvc: Do we have a way to format this as an aside? Otherwise, I'd just gently incorporate it into the preceding paragraph.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+
+[[ls-to-ls-overview]]
+==== Configuration overview
```
andrewvc: We could really use a diagram here. @acchen97 do you know where we can get one of those?

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+. Use the Beats input plugin on the second Logstash machine.
+. Generate a trusted SSL certificate (required by the lumberjack protocol).
+. Save the SSL certificate on the first Logstash machine.
+. Save the SSL certificate and key on the second Logstash machine.
+. Test it.
+
+[[generate-self-signed-cert]]
+===== Generate a self-signed SSL certificate and key
+
+You can use the `openssl req` command with some specific flags to generate a self-signed certificate and key. The `openssl req` command is available on any operating system that comes with SSL.
```
andrewvc: I think we can strike `with some specific flags`

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+[source,shell]
+----
+Generating a 2048 bit RSA private key
+.................................+++
+....................+++
+writing new private key to 'lumberjack.key'
+----
+
+[[save-cert-ls1]]
+===== Save the SSL certificate on the first Logstash machine
```
andrewvc: The heading indicates this section is about saving the certificate, but the command is about running the upstream logstash.

I think we want to amend the previous section to end by saying "then copy the generated key and certificate to all machines that will be communicating with each other". The section probably should be titled something like: "Trying it out" or "Testing it out"

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+
+On the first Logstash machine, run the following command:
+[source,shell]
+----
+bin/logstash --path.data /tmp/z -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.crt" port => 5000 } }'
```
andrewvc: We should explain why we're setting `path.data` here, which is to not have two logstashes attempt to use the same data path, which would prevent one from starting. A note saying that we're only doing this to be able to quickly demo this makes sense.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
+[[ls-to-ls]]
+=== Configuring Logstash-to-Logstash Communication
+
+In a complex network environment, you may need two Logstash machines talking to each other. That is, the output from the first Logstash machine becomes the input for the second Logstash machine. You may need this configuration if the data path crosses network or firewall boundaries, for example.
```
andrewvc: I would add to this first section a succinct description of the solution, for those who don't need a tutorial. "The recommended way for one Logstash to communicate with another is by connecting the Lumberjack output to the Beats input", which links to the output and input respectively.

andrewvc: Can we globally replace 'first' and 'second' with 'upstream' and 'downstream', that more clearly documents how these Logstash instances relate to one another.

I think we can generally reword this to say something along the lines of.

"To connect one Logstash to another we use the Lumberjack protocol, which is spoken by the lumberjack output and the beats input. To use it, setup the downstream Logstash instance with the Beats input to listen for incoming Lumberjack connections, then, setup the upstream Logstash instance with the Lumbjerjack output to connect to it."

docs/static/ls-ls-config.asciidoc
karenzone: Hmmmm... I meant to modify that sample based on the premise that this was a symptom of the demo and not likely in real world set up. In other words, I'm trying to help people get set up rather than helping them demo the solution.  1. Is that an accurate assumption? 2. Is setup rather than demo the correct approach for the reference guide?

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+* `localhost` specifies the name of the first Logstash computer
+
+[[save-cert-ls2]]
+===== Save the SSL certificate and key on the second Logstash machine
+
+On the second Logstash machine, run the following command:
+
+[source,shell]
+----
+bin/logstash -e 'input {beats { codec =>? json port => 5000 ssl => ssl_certificate => "lumberjack.crt" ssl_key => "lumberjack.key"} }'
```
andrewvc: We can probably leave off the `codec` declaration.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+* `localhost` specifies the name of the upstream Logstash computer
+
+[[save-cert-ls2]]
+===== Save the SSL certificate and key on the downstream Logstash machine
+
+On the downstream Logstash machine, run the following command:
+
+[source,shell]
+----
+bin/logstash -e 'input {beats { codec =>? json port => 5000 ssl => ssl_certificate => "lumberjack.crt" ssl_key => "lumberjack.key"} }'
```
jordansissel: is the question mark here intended? I don't' see it anywhere else: `codec =>? json`

jordansissel:
LGTM

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+[[generate-self-signed-cert]]
+===== Generate a self-signed SSL certificate and key
+
+You can use the `openssl req` command to generate a self-signed certificate and key. The `openssl req` command is available on any operating system that comes with SSL.
+
+Run the following command:
+
+[source,shell]
+----
+openssl req -x509 -batch -nodes -newkey rsa:2048 -keyout lumberjack.key -out lumberjack.crt -subj /CN=localhost
```
jordansissel: Oops, found a typo.

Here we say `lumberjack.crt` but down below in `where:` we say `lumberjack.cert` (crt vs cert)

jordansissel:
I forgot I could edit this PR, so I changed all the .crt command line references to be .cert to align with the rest of the doc. 

LGTM

jordansissel:
@andrewvc I think this is pending your approval now.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+* `localhost` specifies the name of the upstream Logstash computer
+
+[[save-cert-ls2]]
+===== Save the SSL certificate and key on the downstream Logstash machine
+
+On the downstream Logstash machine, run the following command:
+
+[source,shell]
+----
+bin/logstash -e 'input {beats { codec =>? json port => 5000 ssl => ssl_certificate => "lumberjack.cert" ssl_key => "lumberjack.key"} }'
```
yaauie: `codec =>? json` should probably be `codec => json`

yaauie: `ssl => ssl_certificate => "lumberjack.cert"` should probably be `ssl_certificate => "lumberjack.cert"`

yaauie: inconsistent spacing: `input {beats` should probably be `input { beats`.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+----
+
+where:
+
+* `lumberjack.key` specifies the name of the SSL key to be created
+* `lumberjack.cert` specifies the name of the SSL certificate to be created
+* `localhost` specifies the name of the computer
```
yaauie: `localhost` isn't in the input config, since the downstream will be _listening_ and doesn't need to know the address of the upstream machine(s)

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+On the downstream Logstash machine, run the following command:
+
+[source,shell]
+----
+bin/logstash -e 'input {beats { codec =>? json port => 5000 ssl => ssl_certificate => "lumberjack.cert" ssl_key => "lumberjack.key"} }'
+----
+
+where:
+
+* `lumberjack.key` specifies the name of the SSL key to be created
```
yaauie: the key should already be created by the time we get here, so the "to be created" is misleading; this needs to be the _path_ to the key, not just the name.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+* `lumberjack.cert` specifies the name of the SSL certificate to be created
```
yaauie: the cert should already be created by the time we get here, so the "to be created" is misleading; this needs to be the _path_ to the cert, not just the name.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+On the upstream Logstash machine, run the following command:
+[source,shell]
+----
+bin/logstash --path.data /tmp/z -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
+----
+
+where:
+
+* `5` specifies the number of events you want to generate and send
+* `lumberjack.cert` specifies the name of the SSL certificate you created
```
yaauie: the cert should already be created by the time we get here, so the "to be created" is misleading; this needs to be the _path_ to the cert, not just the name.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+* `localhost` specifies the name of the upstream Logstash computer
```
yaauie: "name" should probably be "address", and refers to the _downstream_ Logstash computer (a.k.a., where this output is going to send data _to_.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+
+On the upstream Logstash machine, run the following command:
+[source,shell]
+----
+bin/logstash --path.data /tmp/z -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
+----
+
+where:
+
+* `5` specifies the number of events you want to generate and send
```
yaauie: may be helpful to include keys and sections in the breakdown, e.g.:
~~~ asciidoc
where:

* in the generator input:
** `count => 5` specifies the number of events you want to generate and send
* in the lumberjack output:
** `codec => json`: indicates that we should send fully-structured events
** `hosts => "localhost"`: indicates the network address to which we should send events
** `port => 5000`: indicates the port to which we send events 
** `ssl_certificate => "lumberjack.cert"`: indicates the local filesystem path to our certificate
~~~

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,105 @@
...
+writing new private key to 'lumberjack.key'
+----
+
+[[save-cert-ls1]]
+===== Save the SSL certificate on the upstream Logstash machine
+
+On the upstream Logstash machine, run the following command:
+[source,shell]
+----
+bin/logstash --path.data /tmp/z -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
```
jordansissel: Thinking about this more -- I'm not sure if we need to specify `--path.data /tmp/z` here. If the assumption is two independent servers running logstash, this flag (`--path.data`) is not necessary. It's only necessary when testing locally with two Logstash instances on the same machine (which is probably rare for this use case).

docs/static/ls-ls-config.asciidoc
jordansissel: This worked for me in testing.
```
bin/logstash -e 'input { beats { codec => json port => 5000 ssl => true ssl_certificate => "lumberjack.cert" ssl_key => "lumberjack.key" } }' 
```

docs/static/ls-ls-config.asciidoc
jordansissel: I tested this and was successful: 

```
  bin/logstash --path.data /tmp/z -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
```

👍

jordansissel:
I ran the 3 commands (openssl, bin/logstash, bin/logstash) and with minor changes was successful -- I put some comments about the changes in the review. Let me know if you need anything other testing :)

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+[[test-ls-to-ls]]
+===== Test the communication
+
+Use the input generator to send test events from the upstream Logstash machine.
+
+The following command sends five events over port 5000:
+
+[source,shell]
+----
+bin/logstash —path.data /tmp/z -e 'input { generator {count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" Port => 5000 } }'
```
karenzone: Rats.  I moved `—path.data /tmp/z -e` to the downstream instance in response to a review comment. So looks like I need to remove it here, too

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+
+[[save-cert-key]]
+===== Save the SSL certificate and key
+
+Copy the SSL certificate to the upstream Logstash machine.
+
+Copy the SSL certificate and key to the downstream Logstash machine.
+
+[[config-demo]]
+==== Demo
```
karenzone: In the first draft, i had a weird blend of configuration and demo.  I restructured the document in an attempt to fix that. Let me know what you think

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,111 @@
...
+
+You can set up communication between two Logstash machines by connecting the Lumberjack output to the Beats input. You may need this configuration if the data path crosses network or firewall boundaries, for example. If you don't have a compelling need for Logstash-to-Logstash communication, then don't implement it.
+
+[[ls-to-ls-overview]]
+==== Configuration overview
+
+Use the Lumberjack protocol to connect two Logstash machines.
+
+. Generate a trusted SSL certificate (required by the lumberjack protocol).
+. Save the SSL certificate on the upstream Logstash machine.
```
andrewvc: "Save" is a weird verb for server side software, since there isn't really a 'save' action in the software operators use for the most part, and there's no unsaved state.

I think "copy" or "put" make more sense in general for Logstash and similar software.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,111 @@
...
+. Save the SSL certificate and key on the downstream Logstash machine.
+. Set the downstream Logstash machine to listen for incoming Lumberjack connections through the Beats input.
+. Set the upstream Logstash machine to use the Lumberjack output to send data.
+. Test it.
+
+[[generate-self-signed-cert]]
+===== Generate a self-signed SSL certificate and key
+
+You can use the `openssl req` command to generate a self-signed certificate and key. The `openssl req` command is available on any operating system that comes with SSL.
```
andrewvc: Not really true that the command is available on any operating system with SSL. We can replace the last sentence with: "You may need to install the openssl command line program".

docs/static/ls-ls-config.asciidoc
andrewvc: @yaauie +1 on path vs. name. However, I think 'to be created' is correct. This describes what the `openssl` command does with the argument, which is, in fact, the creation of the key file.

docs/static/ls-ls-config.asciidoc
andrewvc: Same thought as above, we're documenting the 'openssl' command, not where we are at this point in time.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,111 @@
...
+Generating a 2048 bit RSA private key
+.................................+++
+....................+++
+writing new private key to 'lumberjack.key'
+----
+
+[[save-cert-key]]
+===== Save the SSL certificate and key
+
+Copy the SSL certificate to the upstream Logstash machine.
```
andrewvc: Yes! I much prefer the use of copy here to save :)

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,111 @@
...
+Run the following command to start Logstash and generate test events:
+
+[source,shell]
+----
+bin/logstash -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
+----
+
+where:
+
+* `5` specifies the number of events to be generated
```
andrewvc: If we do keep this, as a convention should we use `5` or `count => 5` or `count`? In other words, do we document the value or the key for that value? This is something we should decide as a style convention for our docs in general.

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,111 @@
...
+===== Start the upstream Logstash instance
+
+Run the following command to start Logstash and generate test events:
+
+[source,shell]
+----
+bin/logstash -e 'input { generator { count => 5 } } output { lumberjack { codec => json hosts => "localhost" ssl_certificate => "lumberjack.cert" port => 5000 } }'
+----
+
+where:
```
andrewvc: I almost wonder if we don't need so much structure here. This could be a paragraph. "This command will send five events to localhost:5000 using the provided SSL certificate to verify the connection".

We don't need to document everything, since this page only is about LS->LS communication

docs/static/ls-ls-config.asciidoc
```diff
@@ -0,0 +1,111 @@
...
+Run the following command to start the downstream instance of Logstash:
+
+[source,shell]
+----
+bin/logstash --path.data /tmp/z -e 'input { beats { codec => json port => 5000 ssl_certificate => "lumberjack.cert" ssl_key => "lumberjack.key"} }'
+----
+
+where:
+
+* `--path.data /tmp/z` sets a different data path (required to start a second Logstash instance on the same computer)
```
andrewvc: Same note here, do we want a list or a paragraph?

docs/static/ls-ls-config.asciidoc
andrewvc: On the other hand, for some users, this info might be valuable. I am worried about us re-documenting how every LS option works on every page.

karenzone: @dedemorton I'd like to get your thoughts on the best approach

docs/static/ls-ls-config.asciidoc
karenzone: @dedemorton I'd like to get your thoughts on the best approach. (I like to use a sample value and then explaining it rather than going too generic. At the same time, I want to be consistent with other writers and products.)

dedemorton: @karenzone The page was not refreshed, so I didn't see your comment until after I wrote mine. Striving for consistency is a bit tricky atm because we have a lot of hands in the docs. I'd focus on making your content internally consistent, accurate, and easy-to-read, and not worry as much about other projects and writers.

PR #9174, opened 2018-02-20
bump gems for 6.2.3

danhermann:


jakelandis:
I think some un-intended dependencies are sneaking in here...

I would suggest to clone a fresh branch, remove the .lock file and run `rake artifact:tar` to generate the new .lock file, and rename it.  The diff between patches should be much smaller.

PR #9175, opened 2018-02-20
bump gems for 5.6.9

danhermann:


jakelandis:
@danhermann - can you hold off on this until https://github.com/logstash-plugins/logstash-filter-grok/pull/136 is merged and released ?  This is a pretty significant memory leak and should be fixed in the next minor.

Gemfile.jruby-1.9.lock.release
```diff
@@ -125,7 +125,7 @@ GEM
     jls-lumberjack (0.0.26)
       concurrent-ruby
     jmespath (1.3.1)
-    jrjackson (0.4.4-java)
```
jakelandis: This should be held back at 0.4.4 until we get the version of Jackson updated , else we get an annoying (but harmless) warning about version mismatch.

Gemfile.jruby-1.9.lock.release
```diff
@@ -185,6 +185,11 @@ GEM
       rspec (~> 3.0)
       rspec-wait
       stud (>= 0.0.20)
+    logstash-filter-aggregate (2.7.2)
```
jakelandis: It is odd to see us adding so many more default plugins here.  How did you generate this lock file ?

PR #9178, opened 2018-02-21
JAVAFICATION: Eliminate ruby calls in worker loop

danhermann:


logstash-core/src/main/java/org/logstash/execution/WorkerLoop.java
```diff
@@ -62,13 +57,14 @@ public void run() {
...
+                    flushRequested.set(false);
                 }
-            } while (!shutdownRequested || isDraining());
+            } while (!shutdownRequested.get() || isDraining());
             //we are shutting down, queue is drained if it was required, now  perform a final flush.
             //for this we need to create a new empty batch to contain the final flushed events
             final QueueBatch batch = readClient.newBatch();
             readClient.startMetrics(batch);
-            execution.compute(batch.to_a(), true, false);
+            execution.compute(batch.to_a(), true, true);
```
original-brownbear: @danhermann hah wops :)Thanks for catching this one

logstash-core/src/main/java/org/logstash/execution/WorkerLoop.java
```diff
@@ -62,13 +57,14 @@ public void run() {
                 readClient.closeBatch(batch);
                 if (isFlush) {
                     flushing.set(false);
+                    flushRequested.set(false);
                 }
-            } while (!shutdownRequested || isDraining());
+            } while (!shutdownRequested.get() || isDraining());
```
original-brownbear: @danhermann hmm maybe we should still save the result of `shutdownRequested.get()` to a normal local `boolean` and short-circuit calls to the `AtomicBoolean#get()` once it's true to convey that this is an "irreversible" thing? It would prevent some weird bugs ... maybe :)

danhermann:
jenkins test this please

logstash-core/src/main/java/org/logstash/execution/WorkerLoop.java
```diff
@@ -17,42 +13,43 @@
...
     public void run() {
         try {
-            boolean shutdownRequested = false;
-            final ThreadContext context = RubyUtil.RUBY.getCurrentContext();
+            boolean isShutdown;
             do {
-                final IRubyObject signal = signalQueue.poll();
-                shutdownRequested = shutdownRequested
-                    || signal != null && signal.callMethod(context, "shutdown?").isTrue();
+                isShutdown = shutdownRequested.get();
```
original-brownbear: @danhermann I think you we should keep with the old behaviour and simply do:

`isShutdown = isShutdown || shutdownRequested.get();`

That's what I was aiming at to prevent being the possibility of "reverting" a shutdown so to speak. Sorry about the confusion :)

original-brownbear:
@danhermann LGTM (except for that one line  that doesn't matter all that much I think :))

danhermann:
jenkins test this please

PR #9182, opened 2018-02-21
Bump to jackson 2.9.4 and pin jrjackson to exact version

jakelandis:
The exact pin will further ensure that versions are kept in sync. For example, jrjackson is released, but version here is not bumped...without the exact pin, the shipped version of jrjackson can get ahead of the version defined here.


original-brownbear:
@jakelandis LGTM :)

jakelandis:
@original-brownbear - thanks!

jakelandis:
this PR can only be merged as far back as the Java entry point (6.2 i believe). Older versions also need  `gradle gemspec_jars` updated.

PR #9189, opened 2018-02-22
BUG: Fix race condition in flusher

original-brownbear:
Fixes race condition leading to LS locking up when `flushing` is set to false and a subsequently set `true` on `flushRequested` is set to `false` before being picked up by any worker thread. 

@danhermann


danhermann:
Thanks, @original-brownbear! LGTM.

PR #9190, opened 2018-02-22
BUG: Fix input shutting down prematurely

original-brownbear:
@danhermann lets trigger Jenkins a few times to make sure, but I think this is the issue.
The input doesn't block for long enough and the flush happens after the pipeline already shut down. I applied the same fix for the nested conditional wrapping test below this one, but it seems we now changed the timing here in a way that hits this test as well.


danhermann:
jenkins test this please

danhermann:
jenkins test this please

danhermann:
@original-brownbear, 3 consecutive jenkins builds are all successful for this test. LGTM.

original-brownbear:
@danhermann thanks

PR #9206, opened 2018-03-02
Parallel lifecycle

andrewvc:
This way, if one Logstash pipeline is blocked others won't be if they're in the same set of changes.

This is a requirement for https://github.com/logstash-plugins/logstash-integration-internal to work correctly. Without this, connected pipelines might block each other during shutdown. Upstream pipelines must shutdown first in that system, this lets them run simultaneously, thus sorting that issue out.

It possibly also makes booting multiple pipelines faster.

This breaks the plugin API x-pack uses, since `with_pipelines` is removed, but that's an easy patch to x-pack once it's merged into this repo.


andrewvc:
Merged in #9225 , this is now redundant.

PR #9212, opened 2018-03-07
Pull release notes from 6.2 into 6.x. We didn't keep this in sync.

andrewvc:
When closing #9170 I realized that 6.x's release notes are way out of sync with 6.2. 

We should ensure that we always merge release notes into the 6.x branch going forward.


danhermann:
LGTM

PR #9216, opened 2018-03-08
Log top-level fatal errors with more detail

andrewvc:
Without this patch you'd get a message, but no stack trace, etc.
These errors should only be triggered by outright bugs in the code, so no tests here.

Without this test dev can be frustrating because of the lack of a stack trace.


danhermann:
LGTM

PR #9223, opened 2018-03-09
Add OSS variants of artifact rake tasks

andrewvc:
These will omit the x-pack folder once it is merged in

### DIFF
--- a/rakelib/artifacts.rake
+++ b/rakelib/artifacts.rake
@@ -39,6 +39,7 @@ namespace "artifact" do
       "vendor/??*/**/.mvn/**/*",
       "Gemfile",
       "Gemfile.lock",
+      "x-pack/**/*",
     ]
   end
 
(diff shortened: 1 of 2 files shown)


rakelib/artifacts.rake
```diff
@@ -62,13 +63,26 @@ namespace "artifact" do
...
+  def path_matches_exclude?(path, ex)
+    path == ex || (File.directory?(ex) && path =~ /^#{ex}\//)
+  end
+
+  def files(excluder=nil)
+    excluder ||= self.method(:exclude?)
     return @files if @files
     @files = package_files.collect do |glob|
-      Rake::FileList[glob].reject { |path| exclude?(path) }
+      Rake::FileList[glob].reject { |path| excluder.call(path) }
```
yaauie: We can avoid an extra scope by passing the unbound method as a block:

~~~ ruby
      Rake::FileList[glob].reject(&excluder)
~~~

rakelib/artifacts.rake
```diff
@@ -62,13 +63,26 @@ namespace "artifact" do
...
+    excludes.any? { |ex| path_matches_exclude?(path, ex) }
   end
 
-  def files
+  def exclude_oss?(path)
+    path_matches_exclude?(path, "x-pack" ) || exclude?(path)
+  end
+
+  def oss_excluder
+    @oss_excluder ||= self.method(:exclude_oss?)
```
yaauie: while it works, passing around memoised unbounded methods isn't the most straight-forward to understand; let's make a note to take some time on cleaning this up when we're not under pressure timing-wise.

andrewvc: Maybe I've been doing FP for too long but it doesn't seem weird to me? What would the cleanup be?

andrewvc: I mean, we could make it an option `:exclude_oss` that the various methods are provided with, but I think they read the same. Passing a predicate seems just as readable IMHO

Issue #8524, opened 2017-10-25
All output fail if one of many outputs fails


fluency03:
## Logstash -> TCP: tcp output exception, EOFError: End of file reached

According to:
- https://github.com/logstash-plugins/logstash-output-tcp/issues/9, and
- https://github.com/logstash-plugins/logstash-output-tcp/issues/10

This is caused by the shutting down of TCP server, which the logstash’s TCP client is trying to connect to. And this is indeed the case.

As stated in the code [tcp.rb Line 153 of logstash-output-tcp plugin](https://github.com/logstash-plugins/logstash-output-tcp/blob/5f775830420178c78711d79b89b875b818b9e96d/lib/logstash/outputs/tcp.rb#L153):

![image](https://user-images.githubusercontent.com/7440735/31988813-a37eab10-b970-11e7-8694-519529014cef.png)

_“# don't expect any reads, but a readable socket might # mean the remote end closed, so read it and throw it away. # we'll get an EOFError if it happens.”_

# However:

## Logstash stopped outputting event via all output plugins once TCP output stopped. 

Once the TCP output was not working due to the TCP server was stopped, all of our other outputs stopped working as well, such as Redis.

According to:
- https://github.com/elastic/logstash/issues/2463, and
- https://github.com/elastic/logstash/issues/1933, and
- https://github.com/logstash-plugins/logstash-output-tcp/issues/10

Many people have encountered such situation: all output fail if one of many outputs fails. It has been behaving like such for a long time, and according to different developers’ experience, it depends on the plugin. Some plugin’s failure will block all outputs but some won’t.

I think this is a serious problem and has be to be fixed. And I think this is a more general issue not related to single particular Logstash plugin, so I submitted it here.

- Version: 5.6.2
- Operating System: 
- Config File (if you have sensitive info, please remove it):
- Sample Data:
- Steps to Reproduce:

andrewvc:
We agree that this is a problem! What is your preferred solution here. First, if there are outputs that fail without blocking the whole pipeline, that is a bug and a violation of our durability guarantees. Which ones do not behave like this?

My second question is, what is your desired behavior?

If an output is broken we can either:

1. Drop all messages going to it resulting in data loss to that output
2. Buffer those messages to disk (which can't last forever)

For your use case, which is preferable?

fluency03:
Hi @andrewvc , thanks for your reply. 

1. In my case, the TCP output is blocking other outputs (like, redis).

[...]

PR #9225, opened 2018-03-12
Pipeline Input / Output

andrewvc:
Takes over from https://github.com/logstash-plugins/logstash-integration-internal/pull/1 and is a superset of the changes in #9206 .

This PR is the full implementation of inter-pipeline communications in rough form.

The main goal here is to let users address complex problems where isolation and encapsulation are desired. Currently, Logstash pipelines are often times complected logical tasks leading to confusing config files. There are a number of cases where this feature is useful that are described below:

### The Distributor

It is often the case that users would like to use a single input listening on a single port, say, the Beats input, or the HTTP input, to handle different types of data in different ways.

Users will often wind up with a pipeline that looks like:

```
input { 
  beats {}
}

filter {
  if ([type] == "weblogs") {
    # do weblog stuff
  } else {
     # do generic stuff
  }
}

output {
  if ([type] == "weblogs") {
    elasticsearch {
      hosts => ["weblogs.local"]
    }
  } else {
    elasticsearch {
      hosts => ["generic.local"]
    }
  }
}
```

Note the constant repetition of conditionals. With larger configs, in the thousands of lines, this is quite confusing and hard to reason about. With this patch, we can simplify things considerably by breaking things up into distinct pipelines

```
# intake.conf
input {
  beats {}
}

output {
  if ([type] == "weblogs") {
    pipeline { send_to => weblogs }
  } else {
    pipeline { send_to generic_logs }
  }
}

# weblogs.conf
input {
  pipeline { address => weblogs }
}

output {
  elasticsearch { hosts => weblogs.local }
}


# generic.conf
input {
  pipeline { address => generic_logs }
}

output {
  elasticsearch { hosts => generic.local }
}
```

### The Double Buffered Output

Sometimes users have one output, among many, output that has temporary outages. The Logstash pipeline strives not to lose data when an output is blocked, so it blocks and exerts backpressure. If you want to not have this behavior it's not possible today. 

In the wild: 
* [Discuss Thread](https://discuss.elastic.co/t/pipeline-blocking-from-output-plugins/713) 
* https://github.com/elastic/logstash/issues/8524  
* [Old Jira-2267](https://logstash.jira.com/browse/LOGSTASH-2267)

  With internal pipelines it's easy to add buffering to outputs, so that when one goes down the whole pipeline doesn't block. This solves the case where there is a temporary outage.  In the future we could solve allow users to start dropping messages if the PQ is full.

```
- pipeline.id: senderone
  config.string: "input { http {}  } filter { #lotsoflogic  } output { pipeline { send_to => [es, s3] } }"
  queue.type: persisted
- pipeline.id: sendertwo
  config.string: "input  { pipeline { address => es } } output { elasticsearch {} }"
  queue.type: persisted
- pipeline.id: out
  config.string: "input { pipeline { address => s3 } } output { s3 {}  }"
  queue.type: persisted
```

### The Forked Output

Another common pattern is used to deal with the situation where one has multiple outputs that should receive different versions of the data.  

In the wild: https://discuss.elastic.co/t/can-i-filter-the-output-data-separately/36064/5

Today these pipelines look like:
 
```
input {
  http { ... }
}

filter {
  clone { clones => ["toS3"] }
  if [type] == "toS3" {
    mutate { add_field => {"forS3only" => true}}
  } else {
    mutate { add_field => {"forESonly" => true}}
  }
}

output {
  if [type] == "toS3" {
    s3 {}
  } else {
    elasticsearch {}
  }
}

```

With this patch this can be improved to:

```
# distributor.conf
input {
  http { ... }
}

output {
  pipeline { send_to => [s3, es] }
}

# s3.conf

input {
  pipeline { address => s3 }
}

filter {
  mutate { add_field => {"forS3only" => true}}
}

output {
  s3 { ... }
}

# es.conf


input {
  pipeline { address => es }
}

filter {
  mutate { add_field => {"forEsOnly" => true}}
}

output {
  elasticesearch { ... }
}
```

## Implementation

 It includes three distinct changes:

* Introduction of the `pipeline` input / output, which can be used to send messages across pipelines
* Parallel lifecycle events for `Pipeline` lifecycle events, which is necessary to prevent deadlocks (This code is identical to #9206)
* Introduction of Copy on Write (CoW) semantics for `Event#clone`, since that is any time an `Event` crosses a pipeline boundary. Since changes to events must be isolated across pipelines, cheap copies are desirable.

### The Challenge of CoW semantics

**Edit:** The latest version of this should behave correctly WRT CoW. We now copy values on `Event#get`. If a given root key on the `ConvertedMap` is marked `CoW` we will dereference that key and replace its value with a rubyified copy on ruby `Event#get`.

The CoW implementation here is only scoped to top-level keys in event data and metadata. If you change a sub-key in the Event, say `[foo][bar]` then the entirety of `[foo]` is copied. Since there is an additional overhead to preserving the CoW structure with an additional `HashSet` tracking which keys are CoW.

The CoW implementation in this PR is somewhat broken, though it does meet our API specs. According to our [Event API](https://www.elastic.co/guide/en/logstash/6.2/event-api.html) docs, `Event#get` returns a read-only copy of the event. However, the below pattern is common, even in Logstash core:

```ruby
arrVal = event.get("tags")
arrVal << "foo"
event.set(arrVal)
```

The returned value is not a read-only copy at all, neither is it a clone of the underlying data. That means that with our 'proper' CoW semantics, as implemented in this patch, the following breaks the CoW promise.

```ruby
original = Event.new
copy = original.clone
arrVal = copy.get("tags")
arrVal << "foo"
copy.set(arrVal)
original.get("tags") # ["foo"]
copy.get("tags") # ["foo"]
```

The only practical workaround I can think for this is to make the cloned events CoWoR (copy on write or read), and have `Event#get` return a clone of the specified data. This really negates a lot of the benefits of CoW. We could have the compiled pipeline conditionals use a different `Event#get` that didn't make unnecessary copies, as well as things like `add_tag` that we control. Ultimately, the real problem here is that we return the original mutable datastructures. We could traverse the structures and return an unmodifiable variant but that adds overhead.

### DIFF
--- a/docs/static/pipeline-pipeline-config.asciidoc
+++ b/docs/static/pipeline-pipeline-config.asciidoc
@@ -1,0 +1,200 @@
+[[pipeline-to-pipeline]]
+=== Configuring Pipeline-to-Pipeline Communication
+
+When using the multiple pipeline feature of Logstash you may want to connect multiple pipelines on the same Logstash instance together. This can be useful to isolate the execution of these pipelines, as well as to help break-up the logic of complex pipelines. The `pipeline` input / output enables a number of advanced architectural patterns discussed later in this document.
+
+Where communication is needed between Logstash instances you will need to use either {logstash-ref}/ls-to-ls.html[Logstash-to-Logstash] communications, or an intermediary queue, such as Kafka or Redis.
+
+[[pipeline-to-pipeline-overview]]
+==== Configuration overview
+
+Use the `pipeline` input and `pipeline` output to connect two Logstash pipelines running within the same instance. These inputs use a client / server approach, where the `pipeline` input registers a virtual address that a `pipeline` output can connect to.
+
+. Create a 'downstream' pipeline that listens for events on a virtual address.
+. Create an 'upstream' pipeline that produces events, sending them through a `pipeline` output to one or more virtual addresses
+
+A simple example of this configuration can be seen in the below example.
+
+[source,yaml]
+----
+# config/pipelines.yml
+- pipeline.id: upstream
+  config.string: input { stdin {} } output { pipeline { send_to => [myVirtualAddress] } }
+- pipeline.id: downstream
+  config.string: input { pipeline { address => myVirtualAddress } }
+----
+
+[[how-it-works]]
+===== How it works
+
+The `pipeline` input acts as a virtual server listening on a single virtual address in the local process. Only `pipeline` outputs running on the same local Logstash can send events to this address. Pipeline `outputs` can send events to a list of virtual addresses. A `pipeline` output will block if the downstream pipeline is either unavailable or blocked.
+
+When events are sent across pipelines their data is fully copied. Modifications to an event in a downstream pipeline will not affect any other pipelines that event may be used within.
+
+Copying events does, however incur a performance cost. While the `pipeline` plugin may be the most efficient way to communicate between pipelines it still does incur a cost. Logstash must duplicate each event in full on the Java heap for each downstream pipeline a `pipeline` output sends to. Beware that using this feature may affect the heap memory utilization of Logstash.
+
+[[delivery-guarantees]]
+===== Delivery Guarantees
+In its standard configuration the `pipeline` input/output have at-least-once delivery guarantees. The output wil block if the address is unavailable or blocked.
+
+By default, the `ensure_delivery` option on the `pipeline` output is set to `true. If the `ensure_delivery` flag is set to `false`, an unavailable downstream pipeline will cause the sent message to be discarded. A blocked downstream pipeline will block the sending output/pipeline regardless of the value of the `ensure_delivery` flag.
+
+[[avoid-cycles]]
+===== Avoid cycles
+
+It is important when connecting pipelines that the data only flow in one direction. Looping data back around, or connecting the pipelines into a cyclic graph, can cause problems. Logstash waits for each pipeline's work to complete before shutting down. If the pipelines loop data between them this can prevent Logstash from cleanly shutting down.
+
+[[architectural-patterns]]
+==== Architectural patterns
+
+You can use the `pipeline` input and output to better organize code, streamline control flow, and isolate the performance of complex configurations. There are an infinite number of ways to connect pipelines. The ones presented here are hardly comprehensive.
+
+* <<distributor-pattern>>
+* <<output-isolator-pattern>>
+* <<forked-path-pattern>>
+
+[[distributor-pattern]]
+====== The distributor pattern
+
+The Distributor pattern is used in situations where there are multiple types of data coming through a single input, each with its own complex set of processing rules. With the distributor pattern one pipeline is used to route data to other pipelines based on type. Each type is routed to a pipeline with only the logic for handling that type. In this way each type's logic can be isolated.
+
+As an example, in many organizations a single beats input may be used to receive traffic from a variety of sources, each with its own processing logic. A common way of dealing with this type of data is to have a number of `if` conditions separating the traffic and processing each type differently. This approach can quickly get messy when configs are long and complex.
+
+An example distributor configuration is listed below:
+
+[source,yaml]
+----
+# config/pipelines.yml
+- pipeline.id: beats-server
+  config.string: |
+    input { beats { port => 5044 } }
+    output {
+        if [type] == apache {
+          pipeline { send_to => weblogs }
+        } else if [type] == system {
+          pipeline { send_to => syslog }
+        } else {
+          pipeline { send_to => fallback }
+        }
+    }
+- pipeline.id: weblog-processing
+  config.string: |
+    input { pipeline { address => weblogs } }
+    filter {
+       # Weblog filter statements here...
+    }
+    output {
+      elasticsearch { hosts => [es_cluster_a_host] }
+    }
+- pipeline.id: syslog-processing
+  config.string: |
+    input { pipeline { address => syslog } }
+    filter {
+       # Syslog filter statements here...
+    }
+    output {
+      elasticsearch { hosts => [es_cluster_b_host] }
+    }
+- pipeline.id: fallback-processing
+    config.string: |
+    input { pipeline { address => fallback } }
+    output { elasticsearch { hosts => [es_cluster_b_host] } }
+----
+
+Notice how following the flow of data is a simple due to the fact that each pipeline only works on a single specific task.
+
+[[output-isolator-pattern]]
+==== The output isolator pattern
+
+The output isolator pattern is used to prevent Logstash from blocking in the case where there are multiple outputs and one output experiences a temporary failure. For example, a server might be configured to send log data to both Elasticsearch and an HTTP endpoint. It might be the case that the HTTP endpoint is frequently unavailable due to regular service or some other reason.
+
+Logstash, by default, will block when any single output is down. This is an important behavior in guaranteeing at-least-once delivery of data. Unfortunately, in our above scenario this means that whenever the HTTP endpoint is down data is also paused from sending to Elasticsearch. Using the `pipeline` input and output, along with persistent queues, we can continue sending to Elasticsearch even when one output is down, by using the output isolator pattern.
+
+We could employ this pattern for the scenario described above with the following config:
+
+[[source,yaml]]
+----
+# config/pipelines.yml
+- pipeline.id: intake
+  queue.type: persisted
+  config.string: |
+    input { beats { port => 5044 } }
+    output { pipeline { send_to => [es, http] } }
+- pipeline.id: buffered-es
+  queue.type: persisted
+  config.string: |
+    input { pipeline { address => es } }
+    output { elasticsearch { } }
+- pipeline.id: buffered-http
+  queue.type: persisted
+  config.string: |
+    input { pipeline { address => http } }
+    output { http { } }
+----
+
+Please note, that in this architecture, each stage has its own queue, with its own tuning and settings. This would use up to three times as much disk space, and incur three times as much serialization / deserialization cost, as a single pipeline.
+
+[[forked-path-pattern]]
+==== The forked path pattern
+
+The forked path pattern is used in situations where a single event must be processed more than once according to different sets of rules. If not using the `pipeline` input and output this is commonly solved through creative use of the `clone` filter and `if/else` rules.
+
+As an example, let's imagine that we have a use case where we receive data, and index the full event in our own systems, but publish a redacted version of the data to a partner's S3 bucket. We might use the output isolator pattern described above to decouple our writes to either system. The distinguishing feature of the forked path pattern is the existence of additional rules in the downstream pipelines.
+
+An example of this pattern is in the following config:
+
+[[source,yaml]]
+----
+# config/pipelines.yml
+- pipeline.id: intake
+  queue.type: persisted
+  config.string: |
+    input { beats { port => 5044 } }
+    output { pipeline { send_to => [es, http] } }
+- pipeline.id: buffered-es
+  queue.type: persisted
+  config.string: |
+    input { pipeline { address => partner } }
+    # Index the full event
+    output { elasticsearch { } }
+- pipeline.id: partner
+  queue.type: persisted
+  config.string: |
+    input { pipeline { address => s3 } }
+    filter {
+      # Remove the sensitive data
+      mutate { remove_field => 'sensitive-data' }
+    }
+    output { s3 { } } # Output to partner's bucket
+----
+
+[[collector-pattern]]
+==== The collector pattern
+
+The collector pattern is used in situations where you want to define a common set of outputs and pre-output filters that many disparate pipelines might use. This is the opposite of the distributor pattern. In this pattern many pipelines fan in to a single pipeline where outputs and other processing are shared. This pattern simplifies configuration at the cost of reducing isolation, since all data is sent through a single pipeline.
+
+An example of this pattern can be seen below:
+
+
+[[source,yaml]]
+----
+# config/pipelines.yml
+- pipeline.id: beats
+  config.string: |
+    input { beats { port => 5044 } }
+    output { pipeline { send_to => [commonOut] } }
+- pipeline.id: kafka
+  config.string: |
+    input { kafka { ... } }
+    output { pipeline { send_to => [commonOut] } }
+- pipeline.id: partner
+  # This common pipeline enforces the same logic whether data comes from Kafka or Beats
+  config.string: |
+    input { pipeline { address => commonOu } }
+    filter {
+      # Always remove sensitive data from all input sources
+      mutate { remove_field => 'sensitive-data' }
+    }
+    output { elasticsearch { } }
+----
+
--- a/logstash-core/lib/logstash/agent.rb
+++ b/logstash-core/lib/logstash/agent.rb
@@ -25,7 +25,7 @@ class LogStash::Agent
   include LogStash::Util::Loggable
   STARTED_AT = Time.now.freeze
 
-  attr_reader :metric, :name, :settings, :webserver, :dispatcher, :ephemeral_id
+  attr_reader :metric, :name, :settings, :webserver, :dispatcher, :ephemeral_id, :pipelines, :pipeline_bus
   attr_accessor :logger
 
   # initialize method for LogStash::Agent
@@ -39,9 +39,10 @@ class LogStash::Agent
     @auto_reload = setting("config.reload.automatic")
     @ephemeral_id = SecureRandom.uuid
 
-    # Do not use @pipelines directly. Use #with_pipelines which does locking
-    @pipelines = {}
-    @pipelines_lock = java.util.concurrent.locks.ReentrantLock.new
+    # Special bus object for inter-pipelines communications. Used by the `pipeline` input/output
+    @pipeline_bus = org.logstash.plugins.pipeline.PipelineBus.new
+
+    @pipelines = java.util.concurrent.ConcurrentHashMap.new();
 
     @name = setting("node.name")
     @http_host = setting("http.host")
@@ -133,17 +134,6 @@ class LogStash::Agent
     !@running.value
   end
 
-  # Safely perform an operation on the pipelines hash
-  # Using the correct synchronization
-  def with_pipelines
-    begin
-      @pipelines_lock.lock
-      yield @pipelines
-    ensure
-      @pipelines_lock.unlock
-    end
-  end
-
   def converge_state_and_update
     results = @source_loader.fetch
 
@@ -160,14 +150,17 @@ class LogStash::Agent
     # content of it.
     converge_result = nil
 
-    # we don't use the variable here, but we want the locking
-    with_pipelines do |pipelines|
-      pipeline_actions = resolve_actions(results.response)
-      converge_result = converge_state(pipeline_actions)
-      update_metrics(converge_result)
-    end
+    pipeline_actions = resolve_actions(results.response)
+    converge_result = converge_state(pipeline_actions)
+    update_metrics(converge_result)
 
-    report_currently_running_pipelines(converge_result)
+    logger.info(
+        "Pipelines running",
+        :count => running_pipelines.size,
+        :running_pipelines => running_pipelines.keys,
+        :non_running_pipelines => non_running_pipelines.keys
+    )
+
     dispatch_events(converge_result)
 
     converge_result
@@ -229,21 +222,19 @@ class LogStash::Agent
   end
 
   def get_pipeline(pipeline_id)
-    with_pipelines do |pipelines|
-      pipelines[pipeline_id]
-    end
+    pipelines.get(pipeline_id)
   end
 
   def pipelines_count
-    with_pipelines do |pipelines|
-      pipelines.size
-    end
+    pipelines.size
   end
 
-  def with_running_pipelines
-    with_pipelines do |pipelines|
-      yield pipelines.select {|pipeline_id, _| running_pipeline?(pipeline_id) }
-    end
+  def running_pipelines
+    pipelines.select {|id,pipeline| running_pipeline?(id) }
+  end
+
+  def non_running_pipelines
+    pipelines.select {|id,pipeline| !running_pipeline?(id) }
   end
 
   def running_pipelines?
@@ -251,25 +242,19 @@ class LogStash::Agent
   end
 
   def running_pipelines_count
-    with_running_pipelines do |pipelines|
-      pipelines.size
-    end
+    running_pipelines.size
   end
 
   def running_user_defined_pipelines?
-    with_running_user_defined_pipelines do |pipelines|
-      pipelines.size > 0
-    end
+    !running_user_defined_pipelines.empty?
+  end
+
+  def running_user_defined_pipelines
+    pipelines.select {|id, pipeline| running_pipeline?(id) && !pipeline.system? }
   end
 
   def with_running_user_defined_pipelines
-    with_pipelines do |pipelines|
-      found = pipelines.select do |_, pipeline|
-        pipeline.running? && !pipeline.system?
-      end
-
-      yield found
-    end
+    yield running_user_defined_pipelines
   end
 
   private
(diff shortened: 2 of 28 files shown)


kovyrin:
Great idea! We have both the "single input, multiple outputs controlled by conditionals" and the "sending different versions to different outputs" cases in our pipeline and I'd love to be able to simplify those! 👍

yaauie:
I like the idea behind this a LOT.  :tada:

 - The plugin name `internal` doesn't really ring true to me; is there a reason we don't call this `pipeline`?

~~~
input {
  pipeline {
    address => "foo"
  }
}
~~~

~~~
output {
  pipeline {
    send_to => ["foo", "bar"]
  }
}
~~~

I do wonder if we could somehow implement true-CoW using shallow wrappers of [ImmutableRuby](https://github.com/immutable-ruby/immutable-ruby) containers that "quacked like" their mutable counterparts. It'd be very very hard to catch all possible edge cases since ruby is ruby, especially in containers-of-containers scenarios, but maybe we could just coerce to those immutable collections for 7.0?

~~~ ruby
class COW::Hash < Delegator
  def initialize(hash)
    hash = Immutable::Hash.new(hash) unless hash.kind_of?(Immutable::Hash)
    @reference = hash
    # ...
  end

  def []=(key, value)
    sync do
      value = @reference.put(key, value)
      @reference = value unless value.__id__ == @reference.__id__
    end
  end

  # ...
end
~~~

yaauie:
@andrewvc what about performing a deep clone of the event any time we cross the pipeline boundary, at least until we can plan for true-cow or immutable structures? That would carry zero increase in cost for existing pipeline configurations, and only count against us when we're explicitly invoking this featureset.

logstash-core/src/main/java/org/logstash/plugins/internal/Common.java
```diff
@@ -0,0 +1,126 @@
...
+            }
+            state.outputs.add(output);
+            if (state.input != null) {
+                output.updateAddressReceiver(address, state.input::internalReceive);
+            }
+            return state;
+        }));
+    }
+
+    public static void special(Collection<String> addresses, InternalOutput output) {
```
yaauie: `special` doesn't really tell me that this method unbinds the given output from the given addresses, and it may read clearer if we reverse the argument order:

~~~ diff
-    public static void special(Collection<String> addresses, InternalOutput output) {
+    public static void unregisterSender(final InternalOutput output, final Collection<String> addresses) {
~~~

Also, if one or more of the given `addresses` is not registered in `ADDRESS_STATES` , this will NPE dereferencing `state`.

logstash-core/src/main/java/org/logstash/plugins/internal/Common.java
```diff
@@ -0,0 +1,126 @@
+package org.logstash.plugins.internal;
+
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class Common {
+    public static class AddressState {
+        private final String address;
+        private final Set<InternalOutput> outputs = ConcurrentHashMap.newKeySet();
```
yaauie: we get a concurrent hashset , but don't seem to use any of the atomic methods it provides.

logstash-core/src/main/java/org/logstash/plugins/internal/Common.java
```diff
@@ -0,0 +1,126 @@
...
+            if (state.input != null && state.input.isRunning()) {
+                addressesByRunState.running.add(address);
+            } else {
+                addressesByRunState.notRunning.add(address);
+            }
+        });
+        return addressesByRunState;
+    }
+
+    public static void registerSender(Collection<String> addresses, InternalOutput output) {
```
yaauie: nitpick: this would read clearer to me if the arguments were reversed; they can also be final:

~~~ diff
-    public static void registerSender(Collection<String> addresses, InternalOutput output) {
+    public static void registerSender(final InternalOutput output, final Collection<String> addresses) {
~~~

logstash-core/lib/logstash/plugins/builtin/internal/output.rb
```diff
@@ -0,0 +1,61 @@
...
+    receiver = @address_receivers[address]
+    receiver.nil? ? nil : receiver.apply(event)
+  end
+
+  def pipeline_shutting_down?
+    execution_context.pipeline.inputs.all? {|input| input.stop?}
+  end
+
+  def close
+    # Tried to do this with a static method on Common, but somehow jruby kept saying there weren't
```
yaauie: very weird; I've _occasionally_ had issues where IntelliJ loses track of the ruby/jruby bridge that required a restart, but would be interested to try to chase this down at least a little further before "leaking" these details over into the ruby side of things if possible.

logstash-core/lib/logstash/plugins/builtin/internal/output.rb
```diff
@@ -0,0 +1,61 @@
...
+    # enough arguments provided for reasons that make no sense. Probable JRuby bug here
+    @send_to.each do |address|
+        org.logstash.plugins.internal.Common.ADDRESS_STATES.compute(address, proc {|a, state|
+          state.getOutputs.remove(self);
+          return state;
```
yaauie: I think this `return` will return from the `close` method when attempting to unbind from the first address?

~~~
╭─{ yaauie@castrovel:~ }
╰─○ ruby -e 'def wrapper; %w(foo bar).each {|i| proc { return "oh:#{i}" }.call; puts "nope" }; puts "never"; end; puts wrapper'
oh:foo
[success]
~~~

I think this method should be:
~~~ ruby
  def close
    @send_to.each do |address|
      org.logstash.plugins.internal.Common.ADDRESS_STATES.compute(address, proc {|a, state|
        state.getOutputs.remove(self)
        state # put state back in the map
      end
      output.removeAddressReceiver(address);
    end
  end
~~~

logstash-core/lib/logstash/agent.rb
```diff
@@ -339,14 +321,12 @@ def converge_state(pipeline_actions)
...
-    with_pipelines do |pipelines|
-      @state_resolver.resolve(pipelines, pipeline_configs)
-    end
+    @state_resolver.resolve(@pipelines, pipeline_configs)
   end
 
   def report_currently_running_pipelines(converge_result)
     if converge_result.success? && converge_result.total > 0
-      with_running_pipelines do |pipelines|
+      running_pipelines do |pipelines|
```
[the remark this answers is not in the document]
andrewvc: Ah, oops! This is a definite bug with the changed behavior. Good catch.

logstash-core/lib/logstash/pipeline_action/create.rb
```diff
@@ -40,13 +40,19 @@ def execute(agent, pipelines)
           LogStash::Pipeline.new(@pipeline_config, @metric, agent)
         end
 
-      status = pipeline.start # block until the pipeline is correctly started or crashed
-
-      if status
-        pipelines[pipeline_id] = pipeline # The pipeline is successfully started we can add it to the hash
+      status = nil
+      pipelines.compute(pipeline_id) do |id,value|
+        status = pipeline.start # block until the pipeline is correctly started or crashed
```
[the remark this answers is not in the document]
andrewvc: That's a good point. Theoretically, this shouldn't be possible because higher up, when we determine the actions to be run, we would do that check. Today that's safe because that code isn't concurrent. However, it may not be in the future.

I propose that we add an invariant that throws an exception if the pipeline does already exist, since that should never be the case.

logstash-core/lib/logstash/pipeline_action/stop.rb
```diff
@@ -12,13 +12,19 @@ def initialize(pipeline_id)
     end
 
     def execute(agent, pipelines)
-      pipeline = pipelines[pipeline_id]
-      pipeline.shutdown { LogStash::ShutdownWatcher.start(pipeline) }
-      pipeline.thread.join
-      pipelines.delete(pipeline_id)
+      pipelines.compute(pipeline_id) do |k,pipeline|
```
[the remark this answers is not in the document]
andrewvc: I believe, given your comment, that there's nothing to do here?

One point btw: all the `CHM#compute*` functions return the new value, so the second example wouldn't work. The code as is is buggy as you mention below, since this should always return nil.

logstash-core/lib/logstash/pipeline_action/stop.rb
```diff
@@ -12,13 +12,19 @@ def initialize(pipeline_id)
...
+        pipeline.shutdown { LogStash::ShutdownWatcher.start(pipeline) }
+        pipeline.thread.join
+        pipelines.delete(pipeline_id)
```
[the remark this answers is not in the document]
andrewvc: Yes! Good catch

logstash-core/lib/logstash/plugins/builtin/internal/output.rb
andrewvc: The really weird thing is that this happened at runtime too, not just in IntelliJ 🤷‍♀️ 

I'll try and figure this out before the PR is closed, one way or the other.

logstash-core/lib/logstash/plugins/builtin/internal/output.rb
andrewvc: Ah, yes, def a bug. That's what I get for copy/pasting from the original java where `return` did the right thing.

logstash-core/lib/logstash/plugins/registry.rb
```diff
@@ -97,7 +97,8 @@ def register(hooks, settings)
     attr_reader :hooks
 
     def initialize
-      @registry = {}
+      @mutex = Mutex.new
+      @registry = java.util.concurrent.ConcurrentHashMap.new
```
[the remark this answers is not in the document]
andrewvc: Yes, good catch, leftover from a previous iteration

logstash-core/spec/logstash/pipeline_action/reload_spec.rb
```diff
@@ -11,7 +11,7 @@
   let(:new_pipeline_config) { mock_pipeline_config(pipeline_id, "input { generator { id => 'new' } } output { null {} }", { "pipeline.reloadable" => true}) }
   let(:pipeline_config) { "input { generator {} } output { null {} }" }
   let(:pipeline) { mock_pipeline_from_string(pipeline_config, mock_settings("pipeline.reloadable" => true)) }
-  let(:pipelines) { { pipeline_id => pipeline } }
+  let(:pipelines) { chm = java.util.concurrent.ConcurrentHashMap.new; chm[pipeline_id] = pipeline; chm }
```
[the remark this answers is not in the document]
andrewvc: True! I think there's a stylistic difference here.

I prefer not to extract things until we use them 2-3 times. In this case, reading the broken-out version is more complex than the simple inline version. That benefit disappears if we use this pattern in multiple places however.

Having it inline keeps things clear, and also if you remove them, prevents dead helpers that nothing calls from being left behind.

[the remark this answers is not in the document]
andrewvc: I should add, I do use this pattern twice, but it's across two files, so adding a helper across two files would add a good amount of indirection, meaning a reader of the code would need to read that third one. Given that, I think I'd want more than 2 uses here.

logstash-core/src/main/java/org/logstash/Event.java
```diff
@@ -269,6 +276,16 @@ public Event clone() {
         return new Event(map);
     }
 
+    /**
+     * Create a cheap copy of this event that doesn't copy any event data.
+     * On the first invocation of {@link #setField(FieldReference, Object)} in either the original event
+     * or the copy a deep copy will be made
+     * @return
+     */
+    public Event cowClone() {
```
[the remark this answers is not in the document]
andrewvc: Oh, I must have forgotten to push. The intent is to replace the impl. of clone in the JRuby Event class.

logstash-core/src/main/java/org/logstash/plugins/internal/Common.java
andrewvc: Yes, but we share it across threads, hence the use of a threadsafe class.

andrewvc: This does deserve a comment however.

logstash-core/src/main/java/org/logstash/plugins/internal/Common.java
```diff
@@ -0,0 +1,126 @@
...
+        private final String address;
+        private final Set<InternalOutput> outputs = ConcurrentHashMap.newKeySet();
+        private volatile InternalInput input = null;
+
+        private AddressState(String address) {
+            this.address = address;
+            this.input = input;
+        }
+
+        public Set<InternalOutput> getOutputs() {
```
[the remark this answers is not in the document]
andrewvc: I agree, great recommendation :)

logstash-core/src/main/java/org/logstash/plugins/internal/Common.java
andrewvc: I agree! I wound up at this weird order guessing at why JRuby couldn't find these methods. Good catch.

logstash-core/src/main/java/org/logstash/plugins/internal/Common.java
andrewvc: Sorry, this is from when I was experimenting with that JRuby not finding the method issue. I renamed it at one point to see if the previous name was somehow breaking JRuby.

I should mention, this is a WIP, and the intent was to get feedback more on general direction rather than specific code.

This is obviously a garbage method of course.

andrewvc:
@yaauie thanks for the feedback! I should have been more clear when I labeled this a WIP that the code wasn't really of review quality. Thanks for going through it regardless, and recommending a number of improvements :) . The feedback is much appreciated.

WRT the name, I'm down to switch to `pipeline`, I think that's a great suggestion.

WRT true CoW, I think we should make that a goal for 7.0. It would break too much stuff in 6.x of course. Unfortunately, that will be a breaking change. I think figuring out how to do that is tricky. I love immutable Ruby, but my preference there would be to use java types for both the Java API and ruby to remove the conversion cost of moving between them. That'd mean standardizing on Java types and leaving it to library authors to ruby-ify as needed. The good thing about java types is that `String`, `Integer` etc. are all immutable.

That'd be a big breaking change however. I'd want to hear from a number of stakeholders on that one. It'd be great if we could start that discussion however. If that's an area of particular interest to you I'd appreciate someone kicking off the thread.

andrewvc:
I took a first stab at making the CoW become CoWR. It (should) work as is?

Needs many tests plus testing to see if this makes sense from a perf standpoint.

pemontto:
Will the pipeline send_to property be able use event variables?

andrewvc:
@pemontto that is currently not implemented, but potentially could be. Can you share an example config of how you'd use that?

I was thinking people would have a small number of pipelines, and for that situation `if`` statements would suffice. Would love to hear more about your use case however.

pemontto:
@andrewvc we have an common enrichment "pipeline" we'd like to share across numerous pipelines mainly so they can share a cache (well at least the worker threads can), and then we'd like to fan back out to discrete output pipelines based on certain criteria in the event.

andrewvc:
@pemontto I do think it's a great idea in many ways from a UX standpoint, but it seems like maybe it shouldn't be in the initial release. We can definitely add that feature from there.

Thanks for taking the time to give your feedback :)

spec/spec_helper.rb
```diff
@@ -17,7 +17,7 @@
 require 'rack/test'
 require 'rspec'
 require "json"
-
+require 'logstash/runner'
```
andrewvc: It's weird that all our specs require random bits of LS, and potentially dangerous for our testing as well.

andrewvc:
This is (pretty much) out of WIP. I still need to get a qa integration test passing. The code should be good however.

andrewvc:
This will also need docs at some point, (preferable before merge CC @karenzone ).

I think I'd like to get a code / feature review in first to make sure there are no changes there before we document things.

andrewvc:
jenkins, please test this

andrewvc:
jenkins, please retest this

logstash-core/src/test/java/org/logstash/plugins/pipeline/PipelineBusTest.java
```diff
@@ -0,0 +1,166 @@
+package org.logstash.plugins.pipeline;
```
andrewvc: I split tests across here and some rspec tests. The rspec tests are more comprehensive in terms of functionality.

andrewvc:
Build is finally green, woohoo! 

This still needs docs, but aside from that is good for review (again) @yaauie

yaauie:
I did a `s/internal/pipeline` on the man issue comment to reflect the changes you made in naming

docs/static/pipeline-pipeline-config.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+- pipeline.id: partner
+  queue.type: persisted
+  config.string: |
+    input { pipeline { address => s3 } }
+    filter {
+      # Remove the sensitive data
+      mutate { remove_field => 'sensitive-data' }
+    }
+    output { s3 { } } # Output to partner's bucket
+----
```
yaauie: note: we may want to add a "collector" pattern (the opposite of the distributor pattern, in many senses)

 - disparate input, pipelines to normalize
 - single unified output

logstash-core/src/main/java/org/logstash/plugins/pipeline/AddressState.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Class for representing the state of an internal address.
+ */
+public class AddressState {
+    private final String address;
+    private final Set<PipelineOutput> outputs = ConcurrentHashMap.newKeySet();
+    private volatile PipelineInput input = null;
```
yaauie: :tada: @andrewvc you're not leaving me much room for nitpicks. Excellent use of `volatile`.

logstash-core/src/main/java/org/logstash/plugins/pipeline/AddressState.java
```diff
@@ -0,0 +1,75 @@
...
+    public PipelineInput getInput() {
+        return input;
+    }
+
+    /**
+     * Assigns an input to listen on this address. Will return false if another input is already listening.
+     * @param newInput
+     * @return true if successful, false if another input is listening
+     */
+    public boolean assignInputIfMissing(PipelineInput newInput) {
```
yaauie: on atomicity: can we guarantee that two threads won't interleave calls to `AddressState#assignInputIfMissing()` and `AddressState#unassignInput()`, is it worth marking these two methods as synchronized? call-frequency of both should be ultra-low, so sync overhead would be minimal.

docs/static/pipeline-pipeline-config.asciidoc
```diff
@@ -0,0 +1,169 @@
...
+
+When events are sent across pipelines their data is fully copied. Modifications to an event in a downstream pipeline will not affect any other pipelines that event may be used within.
+
+Copying events does, however incur a performance cost. While the `pipeline` plugin may be the most efficient way to communicate between pipelines it still does incur a cost. Logstash must duplicate each event in full on the Java heap for each downstream pipeline a `pipeline` output sends to. Beware that using this feature may affect the heap memory utilization of Logstash.
+
+[[delivery-guarantees]]
+===== Delivery Guarantees
+In its standard configuration the `pipeline` input/output have at-least-once delivery guarantees. The output wil block if the address is unavailable or blocked.
+
+By default, the `ensure_delivery` option on the `pipeline` output is set to `true. If the `ensure_delivery` flag is set to `false`, an unavailable downstream pipeline will cause the sent message to be discarded. A blocked downstream pipeline will block the sending output/pipeline regardless of the value of the `ensure_delivery` flag.
```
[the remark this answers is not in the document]
andrewvc: Good question, its considered to be unavailable, so you'll lose messages. I do wonder if we should keep `ensure_delivery`. I feel like maybe it's more useful if it drops messages when blocked too, but could be confusing in its current limited incarnation? 

We need timeouts to have it drop on blocked pipelines. But, if we add that feature iteratively we'd need another option on top of `ensure_delivery`, like `drop_after_timeout`.

WDYT? Do we keep it and have a dual option later? I feel like letting users configure a timeout is too fiddly, we probably just want an option that acts like UDP. If there's even a whiff of trouble we just drop processing.

docs/static/pipeline-pipeline-config.asciidoc
andrewvc: Good point! I'll add that :)

logstash-core/src/main/java/org/logstash/plugins/pipeline/AddressState.java
andrewvc: Good point! This should definitely be synchronized. We cannot make that guarantee.

docs/static/pipeline-pipeline-config.asciidoc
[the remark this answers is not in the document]
andrewvc: Thinking through it further, I'm OK with `ensure_delivery`. People may actually want to configure the timeout separately. I'm OK with those controls being a little weird but more precise.

robcowart:
@andrewvc I am curious what performance/scale testing was performed? I am not talking about just this feature in isolation. This feature has architectural ramifications that would necessitate end-to-end testing to understand where it can be deployed vs. alternatives.

Background... the current release of our solution includes 37 pipelines (and this number will grow) that can either be deployed all on a single node or distributed as needed. Redis is used for intra-pipeline communications, and can be replaced between major tiers by Kafka when scale or other requirements so dictate. It is very much a "microservices-like" pattern for deploying Logstash logic, much the same as this new feature seeks to address.

It seems that with this new feature persistent queues are a necessity to catch back pressure-related queueing, especially when downstream pipelines are unavailable. With any UDP use-case (syslog, netflow, snmp traps etc) things that slow down throughput are your enemy, because ultimately they lead to dropped packets and lost data. I am concerned that introducing persistent queues also introduces a disk I/O latency bottleneck that will be more harmful than helpful (not to mention that it completely changes the hardware requirements for a logstash instance).

What I think is important to know is how these scenarios compare, with the key metric being which results in the least lost UDP packets as event rates increase...

udp input --> memory queue --> pipeline --> local redis --> next pipeline
udp input --> persistent queue --> pipeline --> pipeline output --> next pipeline

Barring this one concern I could see this being a very valuable feature. I wouldn't really have concerns if the Logstash memory queue was configurable, but it isn't. It will definitely be important to understand the scenarios where this might have advantages or disadvantages.

So if there has been any internal benchmarking/scale testing that you could share, it would save us the time of doing this ourselves.

andrewvc:
@robcowart great questions:

So, this feature is not a blanket substitute for anything else. Redis/kafka may still be the right solution for you, but I do suspect that some of those Redis/kafka instances could be replaced.

This is far more efficient than any external queue because there is no serialization/deserialization penalty. However, the buffering semantics are not as rich as those you get from external queues. You're limited by either the amount of heap if using an MQ, or the amount of disk with the PQ. And, of course, you're limited to single machine reliability with either of those.

WRT the UDP example, that's a big "it depends". Generally, buffering to disk with the PQ is preferred because there's less risk that the filter section could block and cause packets to drop unless such a block is long lived causing the PQ to fill.

I'm not certain why one scenario is MQ+Redis, and the other is PQ +PipelineOutput, I think you could mix and match either way. I would say, however, that the pipeline output / PQ version should work pretty well.

One caveat, the PQ isn't always as fast to write/read as other queues, something we're working on. This should be able to be optimized specially for inter-pipeline comms with #9498 , which in my local tests boosted PQ speeds by 33%.

One thing you didn't mention, BTW, is what queue the next pipeline is using. Each MQ you use has a minimal cost, but PQs can be quite a bit more expensive.

At the end of the day, benchmarking your specific setup is the best thing to do (I know you knew I was going to say that :) ). If you do get some numbers I'd love to see them! Please share them either here, or on a discuss post if possible!

robcowart:
No worries... it was worth checking.

I have been spending a lot of time working on throughput optimization specifically because of the need to clear UDP receive buffers as quickly as possible to avoid dropped packets. While tweaking things like kernel buffers, process priorities, queue and batch sizes all make a difference, the most important thing remains grabbing the data and pushing it to the next stage as efficiently as possible. We were able to get one customer from 85% packet loss, down to 0.05%. However at 10K eps that is still 5 lost messages every second, or 432K per day.

I will have to add the benchmarking work we still have planned. Given a three tier architecture with... collect --> process --> index, I suspect that I will still have a redis/kafka between each tier. However within the processing tier using the pipeline input/output could make a lot of sense.

redbaron4:
@andrewvc This is a great feature. We normally use multiple pipelines to separate out outputs that are prone to disruption (managed by another team) such as WebHDFS. Looking at the docs, it seems that `ensure_delivery: False` will allow upstream pipeline to continue working if downstream pipeline is not available but not in case downstream pipeline is blocked. Do you plan to cover the latter case as well or is it intentional to exclude blocked pipelines from this behaviour?

Issue #9132, opened 2018-02-10
Plugin Manager: NullPointerException with malformed URI in proxy ENV variable


ivo-forlin:
Hi,

this happend to me while fiddling around:
```
logstash@logstash-VirtualBox:~$ http_proxy=myproxy:8080 sudo -E /usr/share/logstash/bin/logstash-plugin install file:///media/sf_sharedVm/filter-uuid.zip 
Unhandled Java exception: java.lang.NullPointerException
java.lang.NullPointerException: null
                                put at java/util/Hashtable.java:460
                        setProperty at java/util/Properties.java:166
                        setProperty at java/lang/System.java:796
                             invoke at java/lang/reflect/Method.java:498
  invokeDirectWithExceptionHandling at org/jruby/javasupport/JavaMethod.java:468
                 invokeStaticDirect at org/jruby/javasupport/JavaMethod.java:370
           apply_env_proxy_settings at /usr/share/logstash/lib/pluginmanager/proxy_support.rb:40
                    configure_proxy at /usr/share/logstash/lib/pluginmanager/proxy_support.rb:71
                             <main> at /usr/share/logstash/lib/pluginmanager/main.rb:26
                     runInterpreter at org/jruby/Ruby.java:845
                     runInterpreter at org/jruby/Ruby.java:849
                        runNormally at org/jruby/Ruby.java:752
                        runNormally at org/jruby/Ruby.java:765
                        runFromMain at org/jruby/Ruby.java:578
                      doRunFromMain at org/jruby/Main.java:417
                        internalRun at org/jruby/Main.java:305
                                run at org/jruby/Main.java:232
                               main at org/jruby/Main.java:204
```
logstash version: 6.1.2

I believe the culprit might be around here:
https://github.com/elastic/logstash/blob/4344e4b6135124d525c19dab71998974987ee222/lib/pluginmanager/proxy_support.rb#L38-L44

Looks like the `URI` class returns `nil` for some parts if the URI is malformed:
```ruby
irb(main):001:0> require 'uri'
=> true
irb(main):002:0> @proxy_uri=URI('myproxy:8080')
=> #<URI::Generic myproxy:8080>
irb(main):003:0> @proxy_uri.scheme
=> "myproxy"
irb(main):004:0> @proxy_uri.host
=> nil
irb(main):005:0> @proxy_uri.port
=> nil
irb(main):006:0> @proxy_uri.user
=> nil
irb(main):007:0> @proxy_uri.password
=> nil
```
[...]

PR #9230, opened 2018-03-12
Detect invalid proxy and raise error

tsg:
As reported in #9132, if one sets `http_proxy=myproxy:8080` the URL
gets parsed to scheme=myproxy and host=nil. This results in a
NullPointerException.

To solve it, we check that the URL is really valid as HTTP and raise a slightly
better error message.

### DIFF
--- a/lib/pluginmanager/proxy_support.rb
+++ b/lib/pluginmanager/proxy_support.rb
@@ -36,6 +36,7 @@ end
 # Apply HTTP_PROXY and HTTPS_PROXY to the current environment
 # this will be used by any JRUBY calls
 def apply_env_proxy_settings(settings)
+  $stderr.puts("Using proxy #{settings}") if ENV["DEBUG"]
   scheme = settings[:protocol].downcase
   java.lang.System.setProperty("#{scheme}.proxyHost", settings[:host])
   java.lang.System.setProperty("#{scheme}.proxyPort", settings[:port].to_s)
@@ -44,7 +45,6 @@ def apply_env_proxy_settings(settings)
 end
 
 def extract_proxy_values_from_uri(proxy_uri)
-  proxy_uri = URI(proxy_uri)
   {
     :protocol => proxy_uri.scheme,
     :host => proxy_uri.host,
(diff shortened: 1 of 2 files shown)


lib/pluginmanager/proxy_support.rb
```diff
@@ -67,6 +68,9 @@ def configure_proxy
   proxy = get_proxy("http_proxy")
   if valid_proxy?(proxy)
     proxy_settings = extract_proxy_values_from_uri(proxy)
+    if proxy_settings[:host].nil?
+      raise StandardError.new("Invalid proxy `#{proxy}`. I parsed it to: #{proxy_settings}")
```
tsg: To reviewers: I wasn't sure which exception to use, I saw `ValidationError` being used in other parts of the code, but that seems to require being defined explicitly. Let me know what would work better.

lib/pluginmanager/proxy_support.rb
```diff
@@ -67,6 +68,9 @@ def configure_proxy
   proxy = get_proxy("http_proxy")
   if valid_proxy?(proxy)
     proxy_settings = extract_proxy_values_from_uri(proxy)
+    if proxy_settings[:host].nil?
```
jsvd: Since this PR introduces a bit more logic, I think the overall flow could be improved as there are some issues with it:
1. there's a valid_proxy? method that does little to validate the validity of the proxy :D
2. the `extract_proxy_values_from_uri` also does the uri parsing

Looking at the code:

```ruby
  proxy = get_proxy("http_proxy")
  if valid_proxy?(proxy)
    proxy_settings = extract_proxy_values_from_uri(proxy)
    if proxy_settings[:host].nil?
      raise StandardError.new("Invalid proxy `#{proxy}`. I parsed it to: #{proxy_settings}")
    end
    proxy_settings[:protocol] = "http"
    apply_env_proxy_settings(proxy_settings)
    proxies << proxy_settings
  end
```

We should probably have instead something like:
1. fetch string representing the proxy from env vars (`get_proxy`) - already done
2. check if proxy exists: probably rename `valid_proxy?` to `proxy_string_exists?` ?
3. parse the proxy string: instead of doing `URI.parse` in `extract_proxy_values_from_uri`, a new `parse_proxy_string` method could perform both the `URI.parse` and raise either a [ConfigurationError](https://github.com/elastic/logstash/blob/master/logstash-core/lib/logstash/errors.rb#L5) or [ArgumentError](https://apidock.com/ruby/ArgumentError) after performing a `.kind_of?(URI::HTTP)` check
4. pass the parsed URI object to `extract_proxy_values_from_uri` - already done

The kind_of check validates if the string is either an HTTP or HTTPS url:

```ruby
jruby-9.1.14.0 :016 > url = URI.parse("localhost:9399")
 => #<URI::Generic localhost:9399> 
jruby-9.1.14.0 :017 > url.kind_of?(URI::HTTP)
 => false 
jruby-9.1.14.0 :018 > url = URI.parse("https://localhost:9399")
 => #<URI::HTTPS https://localhost:9399> 
jruby-9.1.14.0 :019 > url.kind_of?(URI::HTTP)
 => true 
```

tsg: Thanks for the review, I'll give this a try.

tsg:
@jsvd @yaauie Can you have another look, please? I went with `RuntimeError`, following Ry's suggestion. but happy to change it to something else. I did try `ConfigurationError`, but wasn't sure if accessing Logstash types from the plugin-manager is a good idea, and Ry said `ArgumentError` is more about internal arguments being mixed up.

tsg:
jenkins, test it

lib/pluginmanager/proxy_support.rb
```diff
@@ -54,27 +54,37 @@ def extract_proxy_values_from_uri(proxy_uri)
   }
 end
 
+def parse_proxy_string(proxy_string)
+  proxy_uri = URI.parse(proxy_string)
+  if !proxy_uri.kind_of?(URI::HTTP) and !proxy_uri.kind_of?(URI::HTTPS)
```
yaauie: nitpick: 

use `&&` instead of `and` in conditionals; this is a surprising gotcha in Ruby because they are _mostly_ interchangeable, but carry different order precedence that can be very surprising and hard to debug when we get to more complex expressions.

More reading: https://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby

tsg: Good to know, thank you!

lib/pluginmanager/proxy_support.rb
```diff
@@ -54,27 +54,37 @@ def extract_proxy_values_from_uri(proxy_uri)
   }
 end
 
+def parse_proxy_string(proxy_string)
+  proxy_uri = URI.parse(proxy_string)
+  if !proxy_uri.kind_of?(URI::HTTP) && !proxy_uri.kind_of?(URI::HTTPS)
+    raise "Invalid proxy `#{proxy_uri}`. The URI is not HTTP/HTTPS."
+  end
+  return proxy_uri
```
jsvd: to make this a bit more readable, you could do instead:

```ruby
def parse_proxy_string(proxy_string)
  proxy_uri = URI.parse(proxy_string)
  if proxy_uri.kind_of?(URI::HTTP) # URI::HTTPS is already a subclass of URI::HTTP http://ruby-doc.org/stdlib-2.0.0/libdoc/uri/rdoc/URI/HTTPS.html
    proxy_uri
  else
    raise "Invalid proxy `#{proxy_uri}`. The URI is not HTTP/HTTPS."
  end
end
```

tsg:
@jsvd Pushed a new version. Can you have a look again, please?

tsg:
jenkins, test it

PR #9234, opened 2018-03-13
Update performance-checklist.asciidoc

timothynfarmer:
Updated links to VisualVM from the now retired java.net.

### DIFF
--- a/docs/static/performance-checklist.asciidoc
+++ b/docs/static/performance-checklist.asciidoc
@@ -88,7 +88,7 @@ following suggestions:
 [[profiling-the-heap]]
 ==== Profiling the Heap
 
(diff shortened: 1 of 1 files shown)


karmi:
Hi @timothynfarmer, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/logstash/pull/9234.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

timothynfarmer:
@karmi looks like my email was set to private so it was using a custom Git address (I wasn't aware of that). Not sure if I can resign with my gmail or not.

jsvd:
@timothynfarmer the cla check passes now 👍 thanks for the pr

PR #9238, opened 2018-03-14
Add additional gradle tasks in anticipation of x-pack being opened

andrewvc:
These don't do anything yet of course, but will soon enough!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -284,14 +284,6 @@ bootstrap.dependsOn installTestGems
 runIntegrationTests.shouldRunAfter tasks.getByPath(":logstash-core:test")
 check.dependsOn runIntegrationTests
 
-task runXPackUnitTests(type: Exec, dependsOn: installTestGems) {
-    workingDir "${projectDir}/x-pack"
-    standardOutput = new ExecLogOutputStream(System.out)
-    errorOutput = new ExecLogOutputStream(System.err)
-    commandLine(['../bin/rspec', 'spec', '-fd'])
-}
-
-
 String elasticsearchSnapshotURL = "https://snapshots.elastic.co/downloads/elasticsearch/elasticsearch-${version}-SNAPSHOT.tar.gz"
 String elasticsearchDownloadLocation = "${projectDir}/build/elasticsearch-${version}-SNAPSHOT.tar.gz"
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -284,3 +284,44 @@ bootstrap.dependsOn installTestGems
 runIntegrationTests.shouldRunAfter tasks.getByPath(":logstash-core:test")
 check.dependsOn runIntegrationTests
 
+task runXPackUnitTests(type: Exec, dependsOn: installTestGems) {
```
yaauie: Ideally, we would respect `OSS=true` everywhere, and bail before loading commercial code (even if it is the tests, and even if a user explicitly asks us to run them).

My primary worry is that this task may eventually be accidentally made a dependency of another task, which could surprise a contributor who wanted to avoid running commercially-licensed code.

build.gradle
```diff
@@ -284,3 +284,44 @@ bootstrap.dependsOn installTestGems
...
+task copyEs(type: Copy, dependsOn: [downloadEs, deleteLocalEs]) {
+    from tarTree(resources.gzip(elasticsearchDownloadLocation))
+    into "./build/"
+    doLast {
+        file("./build/elasticsearch-${version}-SNAPSHOT").renameTo('./build/elasticsearch')
+    }
+}
+
+task runXPackIntegrationTests(type: Exec, dependsOn: [installTestGems, copyEs]) {
+    workingDir "${projectDir}/x-pack"
```
yaauie: same note about `OSS=true`

build.gradle
```diff
@@ -284,3 +284,47 @@ bootstrap.dependsOn installTestGems
 runIntegrationTests.shouldRunAfter tasks.getByPath(":logstash-core:test")
 check.dependsOn runIntegrationTests
 
+String elasticsearchSnapshotURL = "https://snapshots.elastic.co/downloads/elasticsearch/elasticsearch-${version}-SNAPSHOT.tar.gz"
+String elasticsearchDownloadLocation = "${projectDir}/build/elasticsearch-${version}-SNAPSHOT.tar.gz"
+
+task downloadEs(type: Download) {
```
andrewvc: I've left these ES tasks unmasked, since I anticipate we'll use them for @original-brownbear 's work for distributed plugin execution

PR #9239, opened 2018-03-14
opening x-pack: make plugin install/remove error helpfully

yaauie:
Since we are opening x-pack and making it part of the distribution, the install- and remove- steps will no longer function as expected; in order to prevent inexplicable errors, intercept these instructions and error helpfully.

Example output:

### DIFF
--- a/lib/bootstrap/environment.rb
+++ b/lib/bootstrap/environment.rb
@@ -36,6 +36,12 @@ module LogStash
       RUBY_ENGINE
     end
 
+    def oss_only?
+      return true if ENV['OSS']=="true"
+
+      !File.exists?(File.join(LogStash::Environment::LOGSTASH_HOME, "x-pack"))
+    end
+
     def windows?
       ::Gem.win_platform?
     end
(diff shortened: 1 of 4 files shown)


andrewvc:
Looks like builds are failing here, running a rake plugin install I see:

```

ERROR: Something went wrong when installing logstash-filter-clone, logstash-filter-cidr, logstash-filter-mutate, logstash-input-generator, logstash-input-stdin, logstash-input-tcp, logstash-output-stdout, logstash-filter-multiline, message: undefined method `intercept!' for LogStash::PluginManager::XPackInterceptor::Install:Module
Did you mean?  Integer
```

yaauie:
derp; I had an `extend self` on that module locally that I failed to stage for committing. It's been pushed now.

andrewvc:
I should add, LGTM is good pending CI

yaauie:
jenkins test this please

yaauie:
We're getting an OOM in test, retriggering...

PR #9240, opened 2018-03-14
Never use gradle daemon on CI servers

andrewvc:
We've had some issues with builds crashing the gradle daemon on CI servers. This can obfuscate errors.
This patch removes its usage in CI.

This is considered a gradle best practice: https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:stopping_an_existing_daemon


yaauie:
Still LGTM.

PR #9241, opened 2018-03-14
update README to have statements that remain true after x-pack added

yaauie:
The commit message is important here, because we want the intent of this change to be clear to anyone who may have fear about where we're going:

~~~
commit 1596f2c58ada13683f1f5920199a5cb21da05d7e (HEAD -> opening-xpack-readme-updates)
Author: Ry Biesemeyer <ry.biesemeyer@elastic.co>
Date:   Wed Mar 14 22:56:03 2018 +0000

    remove overly-broad statements about licensing from docs

    Elastic has recently announced that we are [opening x-pack][] code, which was
    previously private. It will be released under a commercial license and
    included in some distributions of Logstash, making certain distributions not
    fully-FOSS-compliant. For that reason, we want to avoid making overly-broad
    statements about "open-source" that could become confusing in the near future.

    Elastic continues to invest heavily in the Apache2-licensed open core of
    Logstash; we will continue to release and distribute OSS-only artifacts, and
    will empower our community to use and extend Logstash.

    [opening x-pack]: https://www.elastic.co/blog/doubling-down-on-open
~~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 # Logstash
 
-Logstash is part of the [Elastic Stack](https://www.elastic.co/products) along with Beats, Elasticsearch and Kibana. Logstash is an open source, server-side data processing pipeline that ingests data from a multitude of sources simultaneously, transforms it, and then sends it to your favorite "stash." (Ours is Elasticsearch, naturally.). Logstash has over 200 plugins, and you can write your own very easily as well.
-
-The license is Apache 2.0, meaning you are pretty much free to use it however you want in whatever way.
+Logstash is part of the [Elastic Stack](https://www.elastic.co/products) along with Beats, Elasticsearch and Kibana. Logstash is a server-side data processing pipeline that ingests data from a multitude of sources simultaneously, transforms it, and then sends it to your favorite "stash." (Ours is Elasticsearch, naturally.). Logstash has over 200 plugins, and you can write your own very easily as well.
 
 For more info, see <https://www.elastic.co/products/logstash>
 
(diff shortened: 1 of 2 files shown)


andrewvc:
To make this switch more atomic, maybe we should just remove these parts or refer to the LICENSE file. WDYT?

andrewvc:
I fear that this would be confusing to someone looking at the repo today.

karenzone:
A summary might be nice IF we can nail it and won't confuse people. Linking to the license would be nice so it doesn't appear that we're hiding anything. 

I've already taken a shot at the text.  Pasting it here so I won't lose it:

Elastic has recently announced that we are opening the x-pack code, which was previously private. It will be released under a commercial license and included in some distributions of Logstash. 
Some artifacts we release will be mixed-license and not wholly FOSS-compliant. For that reason, we want to avoid making overly-broad statements that the entire project is “open source.”

Elastic continues to invest heavily in the Apache2-licensed open core of Logstash. We will continue to release and distribute OSS-only artifacts, and will empower our community to contribute should they decide to opt out of the x-pack features.

yaauie:
@karenzone 

> I've already taken a shot at the text. Pasting it here so I won't lose it:

While I welcome input on my commit messages, the primary change that I'm proposing can be seen in the diff, under the "Files Changed" tab in Github. The commit message will live along-side my change for anyone who wants to dig into the history, and exists solely to provide context.

@andrewvc 

> I fear that this would be confusing to someone looking at the repo today.

> maybe we should just remove these parts or refer to the LICENSE file

I think that all of the statements are true about the project today, and will remain true after x-pack is added; I avoided trimming out references to "open-source" from these statements because I didn't want to give anyone from the community any reason to believe that we were abandoning open source (to be clear: we aren't).

yaauie:
@karenzone 

>  Linking to the license would be nice so it doesn't appear that we're hiding anything

I agree, but the license and relevant code is not yet a part of the project, so there is nowhere to link to.

As it stands, all files in the project are Apache2-licensed, so the text shouldn't point to licenses that do not apply; this is prep-work, so that when we _add_ x-pack (along with its license), the statements about the project as a whole will still be true.

andrewvc:
Maybe the best thing here would be to just remove all the license next from the README for now? It's not a requirement that we have it in the README.

We can add new text in after x-pack is open.

yaauie:
@andrewvc after ruminating on your feedback, I have taken another stab at trimming out the superfluous references to "open source"; my commit message leans heavily on the rewording provided by @karenzone, and includes a link to our blog post where we talk about opening x-pack.

andrewvc:
LGTM! Thanks for finding this @yaauie , this was a good catch, and thanks for the edits @karenzone :)

PR #9245, opened 2018-03-15
DRY docker shell scripts for use with docker CI

andrewvc:
We'll be using ci/docker_run.sh to run our x-pack build scripts soon,
so this refactor makes sense

### DIFF
--- a/ci/docker_run.sh
+++ b/ci/docker_run.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# Syntax is docker_run.sh IMAGE_NAME SCRIPT_NAME *EXTRA_DOCKER_OPTS
+
+set -x # We want verbosity here, this mostly runs on CI and we want to easily debug stuff
+
+#Note - ensure that the -e flag is NOT set, and explicitly check the $? status to allow for clean up
+
+REMOVE_IMAGE=false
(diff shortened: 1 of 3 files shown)


ci/docker_run.sh
```diff
@@ -0,0 +1,27 @@
...
+    # Jenkins
+    IMAGE_NAME=$branch_specifier"-"$(date +%s%N)
+fi
+
+echo "Running Docker CI build for '$IMAGE_NAME' "
+
+docker build  -t $IMAGE_NAME .
+exit_code=$?; [[ $exit_code != 0 ]] && exit $exit_code
+# Run the command, skip the first argument, which is the local script name
+docker run --sig-proxy=true --rm $IMAGE_NAME ${@:2}
```
jakelandis: does this skip the first command , or just execute the second command ? 
e.g.
for these arguments : `a b c d` will this result in `b c d` or just `b` ?

andrewvc: Skips the first one, includes the rest. So, `b c d`. The first argument is the docker image name of course

andrewvc: I see my comment in the code is incorrect, it says it skips the script name, it actually skips the image name

jakelandis:
LGTM

PR #9253, opened 2018-03-16
Add link to xpack settings

karenzone:
The logstash.yml settings are documented in this topic: https://www.elastic.co/guide/en/logstash/current/logstash-settings-file.html. Added a cross reference to X-Pack settings (https://www.elastic.co/guide/en/logstash/current/settings-xpack.html) so that users have easy access to all logstash.yml settings.

### DIFF
--- a/docs/static/settings-file.asciidoc
+++ b/docs/static/settings-file.asciidoc
@@ -63,7 +63,7 @@ modules:
 
 IMPORTANT: If the <<command-line-flags,command-line flag>> `--modules` is used, any modules defined in the `logstash.yml` file will be ignored.
 
-The `logstash.yml` file includes the following settings:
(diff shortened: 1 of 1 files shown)


docs/static/settings-file.asciidoc
```diff
@@ -63,7 +63,7 @@ modules:
 
 IMPORTANT: If the <<command-line-flags,command-line flag>> `--modules` is used, any modules defined in the `logstash.yml` file will be ignored.
 
-The `logstash.yml` file includes the following settings:
+The `logstash.yml` file includes the following settings. If you are using X-Pack, also see https://www.elastic.co/guide/en/logstash/current/settings-xpack.html[X-Pack Settings in Logstash].
```
dedemorton: close...but you need to use an asciidoc attribute to resolve the versioned path. See the attributes here:

https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc#L4

So to create the correct path, you want to use: `{logstash-ref}/settings-xpack.html[X-Pack Settings in Logstash]`

PR #9265, opened 2018-03-20
Bump Gemfile

robbavey:
Bump Gemfile with latest versions for potential 6.1.5 release

### DIFF
--- a/Gemfile.jruby-2.3.lock.release
+++ b/Gemfile.jruby-2.3.lock.release
@@ -1,7 +1,7 @@
 PATH
   remote: ./logstash-core
   specs:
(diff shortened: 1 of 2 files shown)


robbavey:
Jenkins, test this please

Gemfile.jruby-2.3.lock.release
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: ./logstash-core
   specs:
-    logstash-core (6.1.3-java)
```
jsvd: wow how did this happen? does a 6.1.4 work correctly without needs this to be a correct value?

PR #9268, opened 2018-03-21
#9255 speed up concurrent Eventcondition compilations

### LINKED ISSUE #9255: Java Execution compilation gradually gets slower
- Version: 6.2.1
- Operating System: Debian stretch
- JVM version: 1.8.0_152
- Config File (if you have sensitive info, please remove it): config.conf in the tar-file.
- Sample Data: N/A
- Steps to Reproduce: `bin/logstash --experimental-java-execution -f config.conf --log.level=debug`

For jstacks and config.conf see https://stockholmuniversity.box.com/shared/static/uqokj5f72l8xbyhlhbe7imztn4fxdrqa.tar

original-brownbear:
Fixes #9255 (compilation time goes from 50min+ to 2min):

The performance issue there is mostly in the fact that every lookup on the `CACHE` is synchronized while the compilation of a new `EventCondition` takes quite a bit of time.

Fixed by looking up from a `ConcurrentHashMap` in a non-blocking way before going into the sync block.

Note:
* We have to try and look up from the cache again inside the synced block to avoid duplicate compilations
* This implementation is much much faster than using `computeIfAbsent` because compute if absent will block if the key is present but the bin is modified concurrently
   * Also https://bugs.openjdk.java.net/browse/JDK-8062841 breaks JDK older than 8b88 here in practice (reproduced by me) => we don't want to use it.

### DIFF
--- a/logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
+++ b/logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
@@ -1,9 +1,9 @@
 package org.logstash.config.ir.compiler;
 
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 import org.jruby.RubyInteger;
 import org.jruby.RubyNumeric;
 import org.jruby.RubyString;
@@ -67,7 +67,7 @@ public interface EventCondition {
         /**
          * Cache of all compiled {@link EventCondition}.
          */
-        private static final Map<String, EventCondition> CACHE = new HashMap<>(10);
+        private static final Map<String, EventCondition> CACHE = new ConcurrentHashMap<>(10);
 
         private Compiler() {
             //Utility Class.
@@ -82,9 +82,13 @@ public interface EventCondition {
          * @return Compiled {@link EventCondition}
          */
         public static EventCondition buildCondition(final BooleanExpression expression) {
+            final String cachekey = expression.toRubyString();
+            EventCondition cached = CACHE.get(cachekey);
+            if (cached != null) {
+                return cached;
+            }
             synchronized (CACHE) {
-                final String cachekey = expression.toRubyString();
-                final EventCondition cached = CACHE.get(cachekey);
+                cached = CACHE.get(cachekey);
                 if (cached != null) {
                     return cached;
                 }


original-brownbear:
Jenkins test this

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -82,9 +82,13 @@ private Compiler() {
          * @return Compiled {@link EventCondition}
          */
         public static EventCondition buildCondition(final BooleanExpression expression) {
+            final String cachekey = expression.toRubyString();
+            EventCondition cached = CACHE.get(cachekey);
+            if (cached != null) {
+                return cached;
+            }
             synchronized (CACHE) {
```
andrewvc: Much improved, but wouldn't it be cleaner / avoid contention to use a flow more like:

```java
EventCondiion buildCondition(final BooleanExpression expression) {
  final String cacheKey = expression.toRubyString();
  return CACHE.computeIfAbsent(EventCondition::buildConditionUncached)
}
```

AFAIK this prevents us from locking the whole datastructure, but rather, we just lock the key, which should increase liveness. Plus, it's easier to read.

original-brownbear:
Jenkins test this

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
original-brownbear: @andrewvc sorry I'm a little confused :), so this point is resolved, right?  (as you said I explained why we can't do this in the PR description)

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
```diff
@@ -67,7 +67,7 @@
         /**
          * Cache of all compiled {@link EventCondition}.
          */
-        private static final Map<String, EventCondition> CACHE = new HashMap<>(10);
+        private static final Map<String, EventCondition> CACHE = new ConcurrentHashMap<>(10);
```
andrewvc: Thinking about this, I'd prefer we use an LRU here. With config reloading we have a leak here.

It's not strictly in scope for this PR. I've opened https://github.com/elastic/logstash/issues/9273 to track it.

original-brownbear:
Build failures are unrelated, Jenkins keeps OOMing for some reason. Passes fine locally.

logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java
original-brownbear: @andrewvc sorry only now realized there was a main-line comment.

Funny enough ... you're right the JIRA's seem to only address the recursive situation for `#computeIfAbsent` ... I'm able to reproduce the infinite loop without recursion here though, doing exactly what you suggest there. Will try to create a more minimal reproducer an open a JIRA in OpenJDK :)

original-brownbear:
@andrewvc good to merge here, so I can add another PR with more improvements and relative numbers? :)

original-brownbear:
Jenkins test this

original-brownbear:
@andrewvc round 2 (independent of this one, but the impact is clearer if this one is in first) giving us another 33% speedup <= #9278 :)

original-brownbear:
@andrewvc I'm killing this one, it's fully redundant as a result of #9278 (that one brings call counts to this method down by orders of magnitude) ... no need to make the code here uglier if we don't need the speedup.

PR #9270, opened 2018-03-21
fix typo

weslambert:


karenzone:
@weslambert  Good catch! Thanks for the fix

PR #9275, opened 2018-03-21
Fix link to support matrix

dedemorton:
The anchor in the support matrix has changed.


dedemorton:
Note to self: need to copy this change to logstash-docs.

PR #9278, opened 2018-03-22
#9255 Speedup pipeline compilation by making IfVertex more efficient

original-brownbear:
Obvious fix, removing a lot of `Stream` to `Collection` conversion => causes a speedup from `60s` to `40s` for the example config in #9255. 

Update: Also pushed another commit here that inlines the parent vertex compilation so that we only compile (or look up from the compile cache) children when we actually need them and don't do so before the cache lookup. I realize that commit number 2 here makes the code less readable  :( ... but **it brings down total compile time of the example to ~2-3s**


original-brownbear:
Jenkins test this

original-brownbear:
Jenkins test this

original-brownbear:
Jenkins test this

original-brownbear:
@danhermann thanks :)

PR #9279, opened 2018-03-22
Docker CI Images should cleanup in the event of an unclean exit.

andrewvc:
We have a problem with lingering containers, this (hopefully?) fixes that mostly.

### DIFF
--- a/ci/docker_run.sh
+++ b/ci/docker_run.sh
@@ -11,16 +11,25 @@ if [ -z "$branch_specifier" ]; then
     REMOVE_IMAGE=true
     IMAGE_NAME="$1"
 else
-    # Jenkins
     IMAGE_NAME=$branch_specifier"-"$(date +%s%N)
 fi
 
 echo "Running Docker CI build for '$IMAGE_NAME' "
 
-docker build  -t $IMAGE_NAME .
+# Remove old docker cid just in case
+rm -f docker_cid
+
+docker build -t $IMAGE_NAME .
 exit_code=$?; [[ $exit_code != 0 ]] && exit $exit_code
+
+cleanup() {
+  cat docker_cid | xargs docker rm --force -v 
+}
+trap cleanup EXIT
+
 # Run the command, skip the first argument, which is the image name
-docker run --sig-proxy=true --rm $IMAGE_NAME ${@:2}
+echo "Running tests in built docker image"
+docker run --sig-proxy=true --cidfile=docker_cid --rm $IMAGE_NAME ${@:2}
 exit_code=$?
 [[ $REMOVE_IMAGE == "true" ]] && docker rmi $IMAGE_NAME
 echo "exiting with code: '$exit_code'"


ci/docker_run.sh
```diff
@@ -17,11 +17,26 @@ fi
...
+
+cleanup() {
+  echo "Running cleanup..."
+  echo "Killing running containers..."
+  cat docker_cid | xargs docker kill 
+  echo "Cleaning stopped docker containers and dangling images/networks/volumes..."
+  docker system prune -f || true
+  echo "Cleanup complete."
+}
+trap cleanup EXIT
```
jakelandis: it may be safer to use `docker -rm --force <container_id>` instead of `docker kill` .  I had to do that for a couple containers that refused to die.

jakelandis: I am not sure how the trap works ... but is it possible move this to the docker_prune script ? IIRC that runs every hour looking for images to clean up.

ci/docker_run.sh
```diff
@@ -17,11 +17,26 @@ fi
...
+# Remove old docker cid just in case
+rm -f docker_cid
+
+docker build -t $IMAGE_NAME .
 exit_code=$?; [[ $exit_code != 0 ]] && exit $exit_code
+
+cleanup() {
+  echo "Running cleanup..."
+  echo "Killing running containers..."
+  cat docker_cid | xargs docker kill
```
original-brownbear: @andrewvc maybe we should rather use `docker rm -f -v` here in case your script is force aborted and `prune` may not be executed?

original-brownbear:
@andrewvc looks good to me, one minor suggestion but not that important :)

ci/docker_run.sh
```diff
@@ -17,11 +17,26 @@ fi
...
+
+docker build -t $IMAGE_NAME .
 exit_code=$?; [[ $exit_code != 0 ]] && exit $exit_code
+
+cleanup() {
+  echo "Running cleanup..."
+  echo "Killing running containers..."
+  cat docker_cid | xargs docker kill 
+  echo "Cleaning stopped docker containers and dangling images/networks/volumes..."
+  docker system prune -f || true
```
jakelandis: we shouldn't call system prune, since it will potentially delete the base image, the prune script uses filters to avoid this.

ci/docker_run.sh
```diff
@@ -17,11 +17,26 @@ fi
...
+  echo "Cleanup complete."
+}
+trap cleanup EXIT
+
 # Run the command, skip the first argument, which is the image name
-docker run --sig-proxy=true --rm $IMAGE_NAME ${@:2}
+echo "Running tests in built docker image"
+docker run --sig-proxy=true --cidfile=docker_cid --rm $IMAGE_NAME ${@:2}
```
jakelandis: won't this file get overwritten due to our running multiple containers per build ?

andrewvc: Looks like we get a separate workspace per run.

jakelandis: nvmd... each job / container launch has its own dir.

ci/docker_run.sh
```diff
@@ -12,16 +12,27 @@ if [ -z "$branch_specifier" ]; then
     IMAGE_NAME="$1"
 else
     # Jenkins
+    ci/docker_prune.sh
```
jakelandis: there shouldn't be a need to call this explicitly. there is a seperate jenkins job that calls this.

ci/docker_prune.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -i
 
-echo "Removing containers older then 8 hours"
-docker container prune -f --filter "until=8h"
+echo "Removing containers older then 1 hours"
+docker container prune -f --filter "until=1h"
```
jakelandis: I am not against this change, but i think it is un-related from the issue we have. This only cleans up containers that have been stopped and are at least X hours old. 

I would suggest to leave this be to keep the changes as minimal, and we haven't seen any issue with the 8hr strategy.

ci/docker_run.sh
andrewvc: Will that run on all hosts though?

ci/docker_prune.sh
original-brownbear: Actually yea, this is a good point. This may also mask if the trap doesn't work ... :)

ci/docker_run.sh
jakelandis: it will eventually.  we only have 3 hosts today, and it will randomly run on 1 of the 3 every hour iirc. "eventually clean" , like me in my youth.

jakelandis:
LGTM

Issue #9273, opened 2018-03-21
Event Condition cache can leak memory


andrewvc:
Our cache [here](https://github.com/elastic/logstash/blob/master/logstash-core/src/main/java/org/logstash/config/ir/compiler/EventCondition.java#L70) is static and is a potential leak. Most people would probably never hit it, but for people who cycle through lots of configs (think of an org implementing logging as a service) it might explode.

I think an LRU cache makes more sense here.

andrewvc:
I think, as a rule, we should avoid static maps with unbounded growth unless we know that the data-set is of a fixed size. In other cases, we should use an LRU.

We discussed using Guava's Loading cache for this as a standard. Would love to hear other's thoughts on this.

original-brownbear:
@andrewvc just for the record:I think we can resolve this more elegantly down the line by simply generating code for the `EventCondition`s as well. The only reason that's not code generated right now and uses instances of actual objects is that so far I haven't found time to optimize this part.
But sure, for the time being, capping the size of this cache should be fine as a quick-fix.

andrewvc:
@original-brownbear I think it makes sense to just generate the code and get that into test. No need to rush it. I think we'll still be on target to ship this in 6.4.

The more I think about it, the more I'm OK not shipping this as a feature flag in 6.4, but just making it the only option, with the path back to the old execution being a downgrade.

andrewvc:
Also, can we ask @simmel to contribute his config under the CLA (if possible) so we can use it as a test case going forward to prevent similar regressions?

andrewvc:
I think we want to know if there's a regression in our ability to handle larger configs.

I also don't want to unnecessarily ruin our test suite speed.

In an ideal world I think we'd improve Logstash to the point where it doesn't take 1m.

Another thought, we could make it an integration test "Loading giant config takes < forever" and keep it out of unit tests.

WDYT?

danhermann:
I'd been thinking of putting together a suite of longer-running tests but put it off because there was some talk about a common testing framework across products in development. Either way, it sounds like that would be an ideal place for a test like that.

original-brownbear:
@andrewvc integration test sounds better, imo. We don't really need this specific config here btw, any config with a bunch of conditionals will do.

[...]

PR #9282, opened 2018-03-23
 #9273 Make EventCondition cache's lifetime the same as the pipeline lifetime

original-brownbear:
Trivial fix for #9273, simply made the cache non-static:

* Makes the cache's lifetime the same as that of a single pipeline configuration so the deduplication of `EventCondtition` within one compilation is still achieved.
* Moved the `static` methods that I had to make non-`static` to access the cache above the `static` ones to keep the code clean :)


original-brownbear:
@danhermann thanks!

PR #9284, opened 2018-03-23
Improve performance of Event#cancel

andrewvc:
This method was creating a ton of expensive objects making execution slow.

In local tests this improved performance massively, at least 5x in one test with the drop filter.


nerophon:
See also https://github.com/logstash-plugins/logstash-filter-drop/issues/14.

PR #9285, opened 2018-03-23
Fix race condition in shutdown of pipelines

andrewvc:
Prior to this a single worker could slurp down multiple shutdown messages, this prevents that from happening by using a flag that can't be over-consumed.


andrewvc:
@danhermann I've incorporated your feedback, we only set the atomic once now. Good catch :)

danhermann:
jenkins test this please

danhermann:
jenkins test this please

danhermann:
@andrewvc, it looks like some shutdown-related tests are either failing or hanging.

danhermann:
@andrewvc, LGTM if that last pipeline flush test failure is unrelated to your change.

danhermann:
LGTM!

andrewvc:
Other branches:

6.x: 7bb7ecf2e6337f164c5b5f72d8f72168d9c75f57
master: cc726414474a7c0174763c12318fa24472d16a14

PR #9289, opened 2018-03-26
BUG: Fix Java execution not handling nil return from flush invocations

original-brownbear:
Fixes https://github.com/elastic/logstash/issues/9255#issuecomment-376127356 by handling the `flush` call returning `nil` on the Ruby layer.

See https://github.com/elastic/logstash/issues/9255#issuecomment-376147688 for details.

Handled this validation on the Ruby end now, because Java plugins are incoming and I didn't wanna mix more Ruby specifics into the compiler.


original-brownbear:
Jenkins test this

original-brownbear:
@danhermann thanks!

PR #9291, opened 2018-03-26
Remove unnecessary require for reporter specs from pipeline_specs

andrewvc:
Using require this way breaks the ability to run a specific spec with rspec since the require triggers
them to run. This patch fixes that.


danhermann:
LGTM

PR #9293, opened 2018-03-27
More helpful error message for temp directory errors

### LINKED ISSUE #9288: Logstash fails to start with /tmp mounted noexec
Logstash 6.2.3 fails to start with /tmp/ mounted noexec. This seems a regression from version 6.1.4 which works fine with a noexec /tmp.

Mounting /tmp noexec is a security hardening practice for servers and workstations. For example the [CIS Benchmark for Red Hat 7](https://www.cisecurity.org/benchmark/red_hat_linux/) gives the following rationale for mounting /tmp noexec:

> Since the /tmp filesystem is only intended for temporary file storage, set this option to ensure that users cannot run executable binaries from /tmp .

Error in the log:

[2018-03-25T12:23:01,149][ERROR][org.logstash.Logstash    ] java.lang.IllegalStateException: org.jruby.exceptions.RaiseException: (LoadError) Could not load FFI Provider: (NotImplementedError) FFI not available: java.lang.UnsatisfiedLinkError: /tmp/jffi5534463206038012403.so: /tmp/jffi5534463206038012403.so: failed to map segment from shared object: Operation not permitted

This happens on:
- Logstash 6.2.3
- Red Hat 7.4 Maipo, kernel 3.10.0-693.2.1.el7.x86_64
- openjdk version "1.8.0_144"
- /tmp mounted noexec (mount output: `/dev/mapper/vg-tmp on /tmp type xfs (rw,nosuid,noexec,relatime,seclabel,attr2,inode64,noquota)`

After `mount -o remount,exec /tmp` Logstash 6.2.3 proceeds to start fine.

danhermann:
Trap startup exception about inaccessible temp directory and provide more helpful error message with link to more detail in documentation.

Fixes #9288

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Logstash.java
+++ b/logstash-core/src/main/java/org/logstash/Logstash.java
@@ -35,19 +35,40 @@ public final class Logstash implements Runnable, AutoCloseable {
         final String lsHome = System.getenv("LS_HOME");
         if (lsHome == null) {
             throw new IllegalStateException(
-                "LS_HOME environment variable must be set. This is likely a bug that should be reported."
+                    "LS_HOME environment variable must be set. This is likely a bug that should be reported."
             );
         }
         final Path home = Paths.get(lsHome).toAbsolutePath();
         try (
-            final Logstash logstash = new Logstash(home, args, System.out, System.err, System.in)
+                final Logstash logstash = new Logstash(home, args, System.out, System.err, System.in)
         ) {
             logstash.run();
+        } catch (final IllegalStateException e) {
+            String errorMessage[] = null;
+            if (e.getMessage().contains("Could not load FFI Provider")) {
+                errorMessage = new String[]{
+                        "\nError accessing temp directory: " + System.getProperty("java.io.tmpdir"),
+                        "This often occurs because the temp directory has been mounted with NOEXEC or",
+                        "the Logstash user has insufficient permissions on the directory. Possible",
+                        "workarounds include setting the -Djava.io.tmpdir property in the jvm.options",
+                        "file to an alternate directory or correcting the Logstash user's permissions."
+                };
+            }
+            handleCriticalError(e, errorMessage);
         } catch (final Throwable t) {
-            LOGGER.error("Logstash encountered an unexpected fatal error!", t);
-            System.exit(1);
+            handleCriticalError(t, null);
         }
         System.exit(0);
+    }
+
+    private static void handleCriticalError(Throwable t, String[] errorMessage) {
+        LOGGER.error(t);
+        if (errorMessage != null) {
+            for (String err : errorMessage) {
+                System.err.println(err);
+            }
+        }
+        System.exit(1);
     }
 
     /**


danhermann:
@andrewvc, @karenzone, any thoughts on the best place in the Logstash docs (if any) to provide more detail for this issue?  The [directory layout](https://www.elastic.co/guide/en/logstash/current/dir-layout.html) page seemed to me the best option though it's not perfect since the explanation of this issue will be more verbose than the descriptions of any of the other directories.

logstash-core/src/main/java/org/logstash/Logstash.java
```diff
@@ -35,21 +35,39 @@ public static void main(final String... args) {
...
         try (
-            final Logstash logstash = new Logstash(home, args, System.out, System.err, System.in)
+                final Logstash logstash = new Logstash(home, args, System.out, System.err, System.in)
         ) {
             logstash.run();
+        } catch (final IllegalStateException e) {
+            if (e.getMessage().contains("Could not load FFI Provider")) {
+                inaccessibleTempDirWarning();
+            }
+            handleCriticalError(e);
```
andrewvc: This may seem like a nitpick, but could we have the stacktrace print above the tmpwarning? 

People usually read error logs bottom to top, and I think people might find that less confusing. I've gotten stuck on a stacktrace only to realize later that a much more relevant error was right above it.

karenzone:
@danhermann Good question! I'm not finding a good place in the current doc structure for this info. As you said, "The directory layout page (https://www.elastic.co/guide/en/logstash/current/dir-layout.html) seemed to me the best option though it's not perfect since the explanation of this issue will be more verbose than the descriptions of any of the other directories." Also, that topic is about files created when you unpack the LS installation package, not files already present on the system. 

**Do we have additional info that we could share if I had a place to put it?**  The error message covers both the cause and the fix:
> "\nError accessing temp directory: " +System.getProperty("java.io.tmpdir"),
> "This often occurs because the temp directory has been mounted with NOEXEC or",
> "the Logstash user has insufficient permissions on the directory. Possible",
>  "workarounds include setting the -Djava.io.tmpdir property in the jvm.options",
> "file to an alternate directory or correcting the Logstash user's permissions."
Is there more we can say?

Next steps: 
I would like to create a Troubleshooting section in the LS Reference for issues such as this.  (Currently there's a small TS section under "Performance Tuning.")  
If we don't have additional info to share, maybe we can close this one after you commit the message, and I'll open an issue for looking into a Troubleshooting section. I'll reference this PR as an example of the need. 

Looking forward to hearing your thoughts.

danhermann:
@karenzone, if there were an appropriate place in the documentation, I would include some additional information about the underlying cause along the lines of what I posted in this [comment](https://github.com/elastic/logstash/issues/9288#issuecomment-376509715) on the original issue as well as some elaboration on the possible workarounds and their respective pros and cons.

danhermann:
@andrewvc, is there more that needs to be done before this can be merged?

danhermann:
@karenzone, was there a tracking issue for adding information about temp directories to the documentation?

PR #9297, opened 2018-03-28
 #9296: Nested Metadata field lookup in config compiler fixed

### LINKED ISSUE #9296: Java execution conditionals doesn't find @metadata
- Version: 6.x branch commit b952767de42f6cdaa5740b344c18b8eba1b54c1a
- Operating System: OS X and Debian stretch
- Config File (if you have sensitive info, please remove it):
```
input {
  stdin {
  }
}

filter {
  metrics {
    meter => "[@metadata][events]"
      flush_interval => 60
      add_tag => ["graph"]
  }

  if [@metadata][events][count] {
    mutate {
      add_field => {
        "logsource" => "this should only happen when [@metadata][events][count] exists"
      }
    }
  }
}

output {
  stdout {
    codec => rubydebug {
      metadata => true
    }
  }
}
```
- Sample Data: anything will do
- Steps to Reproduce:
With Ruby exec everything works as expected
With Java Exec the field `logsource` gets added even though `[@metadata][events][count]` doesn't exist. Something with conditionals and `@metadata` is broken with Java Exec.

Paging @original-brownbear = )

original-brownbear:
Fixes #9296 by:

* Using the full field `Selector` `jdsl.eTruthy(source_meta, elem.expr)` instead of just the first part of it `elem.recursive_select(SelectorElement).first.expr` for the `Expression` compilation.
* Added JUnit test

PS: Sorry for the whitespace fixes, for some reason new IDEA seems buggy here and kills them even with the setting not to (and it's probably as good a time as any to fix these now)? :)

### DIFF
--- a/logstash-core/lib/logstash/compiler/lscl.rb
+++ b/logstash-core/lib/logstash/compiler/lscl.rb
@@ -13,7 +13,7 @@ module LogStashCompilerLSCLGrammar; module LogStash; module Compiler; module LSC
 
     class Node < Treetop::Runtime::SyntaxNode
     include Helpers
-    
+
     def section_type
       if recursive_select_parent(Plugin).any?
         return "codec"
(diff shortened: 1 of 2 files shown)


simmel:
If only Github had `--ignore-space-change --ignore-all-space` et.al like `git-diff(1)` has ; )

original-brownbear:
@andrewvc thanks!

ycombinator:
@simmel @original-brownbear Just for future reference, if you add `?w=1` to the GitHub URL when looking at the diff, all whitespace-related changes are hidden. For example, https://github.com/elastic/logstash/pull/9297/files?w=1

original-brownbear:
@ycombinator thanks :)

Issue #8777, opened 2017-11-30
pipelineworker exception upon field set


colinsurprenant:
This reproduces in master and 5.6, did not specifically test 6.0.

The following case crashed logstash with a  pipelineworker exception:

```
echo '{"source":"foo", "log":"bar"}' | bin/logstash -e 'input{stdin{codec => json_lines}} filter {mutate {rename => {"source" => "[log][file]"}}} output{stdout{codec => rubydebug}}'
```

which result in the following exception:

```
[...]

PR #9298, opened 2018-03-28
Provide a better error message for Event#set on non-collections

andrewvc:
When executing an Event#set on a nested field that is not a collection we
returned a poor exception. This fixes that.

Fixes https://github.com/elastic/logstash/issues/8777

### DIFF
--- a/logstash-core/src/main/java/org/logstash/Accessors.java
+++ b/logstash-core/src/main/java/org/logstash/Accessors.java
@@ -99,12 +99,25 @@ public final class Accessors {
         return target;
     }
 
+    public static class InvalidFieldSetException extends RuntimeException {
+        public InvalidFieldSetException(final Object target, final String key, final Object value) {
+            super(String.format(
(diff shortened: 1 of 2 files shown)


logstash-core/src/main/java/org/logstash/Accessors.java
```diff
@@ -99,12 +99,26 @@ private static Object findCreateTarget(final ConvertedMap data, final FieldRefer
...
+
     private static Object setChild(final Object target, final String key, final Object value) {
         if (target instanceof ConvertedMap) {
             ((ConvertedMap) target).putInterned(key, value);
             return value;
-        } else {
+        } else if (target instanceof ConvertedList) {
             return setOnList(key, value, (ConvertedList) target);
+        } else {
+            String message = String.format(
```
original-brownbear: @andrewvc I'd suggest moving this `String` generation into the constructor for `InvalidFieldSetException` (by just making its signature `String, String, Object`) to not needlessly blow up method size here.

Looks good otherwise :)

Issue #9295, opened 2018-03-27
Invalid cast in Accessors#setChild


lovasoa:
Related to https://github.com/elastic/logstash/issues/8777 and https://github.com/logstash-plugins/logstash-filter-geoip/issues/137 : 

There seems to be a problem in that piece of code:

https://github.com/elastic/logstash/blob/c77b6b5cbe1c6a98bc6d570e5a34a033643e7857/logstash-core/src/main/java/org/logstash/Accessors.java#L102-L109

The method makes an unchecked cast to `ConvertedList`, when the variable can actually have a different type.

In the case of https://github.com/logstash-plugins/logstash-filter-geoip/issues/137, `target` is an `EmptyMap`, and in the case of #8777 it is a String.

I think the method should not accept an `Object`. It should instead be polymorphic, and accept only the types that it can work with.

andrewvc:
Thanks for the report @lovasoa , and thanks for the extra detail in the ticket.

I've opened up a fix in #9299 

While type safety would be great, since the Event class interops with ruby almost exclusively it doesn't buy much.

As we build out our java plugin API #9215 we can re-examine that signature and potentially improve on tihs experience for java devs.

lovasoa:
thanks for the quick fix!

PR #9299, opened 2018-03-28
Ensure that metadata keys are valuefied

andrewvc:
It was previously the case that setting metadata might not convert values set to metadata.
This could cause type errors.

Fixes https://github.com/elastic/logstash/issues/9295

### DIFF
--- a/logstash-core/src/test/java/org/logstash/EventTest.java
+++ b/logstash-core/src/test/java/org/logstash/EventTest.java
@@ -6,6 +6,7 @@ import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


logstash-core/src/test/java/org/logstash/EventTest.java
```diff
@@ -3,12 +3,7 @@
 import java.io.IOException;
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
original-brownbear: nit: let's not mess up the style here :) This (`.*`) isn't allowed in the ES codestyle afaik.

original-brownbear:
LGTM :)

logstash-core/src/test/java/org/logstash/EventTest.java
andrewvc: Hmmm, well, let me change my IDE settings then.

PR #9303, opened 2018-03-29
 #8218: Ensure atomic creation of PQ checkpoints

### LINKED ISSUE #8218: 0 byte head checkpoint file
We have had reports of  0 byte head checkpoint file preventing logstash to start because of this invalid checkpoint. In both cases where this was reported it seemed to have happened in crash scenarios:
- one where we suspect a file handles depletion 
- one where the VM running logstash is issued a `reboot`

At this point I think we should review the checkpoint handling code to see where this could happen and how to mitigate it.
We should also look at simulating/testing crash conditions to improve the checkpoint handling robustness.

original-brownbear:
Fixes #8218 by ensuring that Checkpoint creation is actually atomic via:

* Create checkpoint file at temporary path
   * Create + subsequent write can never be atomic -> write to temp path first 
* Move it to the actual checkpoint path
   * Move is atomic on effectively all filesystems we support

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
+++ b/logstash-core/src/main/java/org/logstash/ackedqueue/io/FileCheckpointIO.java
@@ -1,18 +1,18 @@
 package org.logstash.ackedqueue.io;
 
-import java.nio.channels.FileChannel;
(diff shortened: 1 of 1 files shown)


danhermann:
LGTM!

original-brownbear:
@danhermann thanks!

PR #9305, opened 2018-03-30
SecretStoreFactory: allow for running with custom environment for simplified testing

yaauie:
Obviates the need for a testing helper that required platform-specific hackery to
mutate the underlying map of `System.getEnv()`.


logstash-core/src/main/java/org/logstash/secret/store/SecretStoreFactory.java
```diff
@@ -99,8 +120,8 @@ private static SecretStore doIt(MODE mode, SecureConfig secureConfig) {
      *
      * @param secureConfig The configuration to add the secret store access
      */
-    private static void addSecretStoreAccess(SecureConfig secureConfig) {
-        String environment = System.getenv(ENVIRONMENT_PASS_KEY);
+    private void addSecretStoreAccess(SecureConfig secureConfig) {
+        String environment = environmentVariables.get(ENVIRONMENT_PASS_KEY);
```
jakelandis: Can you also clear the enironmentVariables map , or set it to null to allow for GC ... part of the futile attempt to keep secrets out of the JVM heap.

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreFactoryTest.java
```diff
@@ -4,17 +4,13 @@
...
 import org.logstash.secret.store.backend.JavaKeyStore;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
jakelandis: nitpick: i think we have been avoiding wildcard imports.

logstash-core/src/main/java/org/logstash/secret/store/SecretStoreFactory.java
yaauie: Doing so would be _entirely_ futile, as it would merely decrement the number of references to the object by one, and not orphan the underlying object -- GC still definitely won't collect it.

Additionally, the `SecretStoreFactory`, after being instantiated, is able to be used multiple times (e.g., in `SecretStoreCLI#deleteThenCreate`), so nulling out the property on the `SecrtetStoreFactory` would break the behaviour and we would need to new-up each time we wanted to use it (from where? It too would need the environment). Previously, the behaviour of this method was to simply rely on `System#getEnv(String)` multiple times, and although it dereferenced the resulting key from local scope, it never was the _final_ dereferencer (as evidenced by the fact that the hackery that this removes for testing worked at all).

***

A previous iteration of mine cached instead a `Supplier<Map<String,String>>`, which was supplied either as a named function reference (e.g., `System::getEnv`) or a lambda that wrapped a map (only used in test; e.g., `() -> environmentVariables`).

In this method, we then used the `Supplier<>` each time the method was invoked to get an instance before working with it as below, but even if we were to remove the reference and force a GC, there is absolutely no way we could remove _all_ of the references to the underlying object -- the `Supplier<>`  holds a reference.

logstash-core/src/test/java/org/logstash/secret/store/SecretStoreFactoryTest.java
yaauie: 👍 

Side note: we really should have a shared IntelliJ config then -- the default config rewrites with wildcards any time we end up with 5 or more individual imports.

logstash-core/src/main/java/org/logstash/secret/store/SecretStoreFactory.java
jakelandis: > Doing so would be entirely futile

I disagree.  The old behavior is reading and falling out scope...it should eventually get collected. We are moving from a potential of having the secret on the heap to a guarantee is on the heap.  I understand there are no guarantees, but I think we should try to mitigate where possible.    

Can we just add getEnvironment() method and override it for the tests ? (and remove the hard reference created in the constructor)

yaauie:
@jakelandis I've moved forward with the environment-variable-supplier route, to prevent us from holding additional references, and my implementation continues to null-out the `String` after we're done using it.

Issue #9167, opened 2018-02-20
Event's passing through PQ have their RubyString's encoding changed


original-brownbear:
The following test for a serialization roundtrip fails:

```java
    @Test
    public void toBinaryRoundtripNonAscii() throws Exception {
        Event e = new Event();
        e.setField("foo", "bör");
        final RubyString before = (RubyString) e.getUnconvertedField("foo");
        Event er = Event.deserialize(e.serialize());
        assertEquals(before, er.getUnconvertedField("foo"));
        assertEquals(
            before.getEncoding(), ((RubyString)er.getUnconvertedField("foo")).getEncoding()
        );
    }
```

with 

```java
java.lang.AssertionError: 
Expected :UTF-8
Actual   :ASCII-8BIT
```

This is causing some real-world issues by breaking e.g. Grok when working with non-ASCII strings (see the linked issue).

This should be fixed by ensuring that the encoding stays constant when going through serialization and deserialization.

yaauie:
@original-brownbear good find.

Since we have a reasonable expectation that strings are UTF-8 for internal representation, and the JSON spec specifically calls out that strings must be unicode, we should ensure that strings are serialised to UTF-8 while building the JSON, and that they are read back out as UTF-8.

I'll put together a patch momentarily.

***

It looks like the `RubyStringDeserializer` being used is deserialising the string using `RubyString#newString(Ruby, byte[])`, which uses the `RubyString(Ruby, RubyClass, byte[])` constructor, creating an ASCII-encoded string backed by a `ByteList`:

> ~~~ java
>         @Override
>         public RubyString deserialize(final JsonParser p, final DeserializationContext ctxt)
>             throws IOException {
>             return RubyString.newString(RubyUtil.RUBY, p.getBinaryValue());
>         }
> ~~~
> -- [logstash-core/src/main/java/org/logstash/ObjectMappers.java:121-125@5dfd0c6](https://github.com/elastic/logstash/blob/5dfd0c60a24b7221bd8e1dc0461c4e16be4d40b2/logstash-core/src/main/java/org/logstash/ObjectMappers.java#L121-L125)

> ~~~ java
>     public static RubyString newString(Ruby runtime, byte[] bytes) {
>         return new RubyString(runtime, runtime.getString(), bytes);
>     }
>  ~~~
[...]

PR #9307, opened 2018-04-01
 #9167 Serialize Encoding with RubyString

original-brownbear:
* Write 0xff prefix to signal encoding is serialized before the string
* Serialize Encoding by classname
* Fallback to old deserialization strategy if 0xff prefix isn't found

-----------------------------

Note:
* Had to change sizes in DLQ tests since `RubyString` now serializes to more bytes
* I decided not to reuse/cache a buffer for serialization in `RubyStringSerializer` because this could lead to significant memory issues for use cases that periodically run into very large `RubyString` fields (we, unfortunately, have some of those)
* This is the only solution that I could think of, that would make the PQ fully transparent for `RubyString` fields. An alternative solution would be to force UTF-8 or UTF-16 encoding on all Strings passing through the PQ and give up on it being 100% transparent. Two points pro/con on that:
   * If we force UTF-16 we're 100% transparent from the Java perspective
   * The overhead of the current solution isn't that much in terms of performance (not noticeable for me locally, the PQ is IO bound by its random access `msync` and the size increase of contiguous memory segments here doesn't hurt things that much) (though for short strings there's a 100%+ overhead in size).

### DIFF
--- a/logstash-core/src/main/java/org/logstash/ObjectMappers.java
+++ b/logstash-core/src/main/java/org/logstash/ObjectMappers.java
@@ -27,7 +27,6 @@ import org.jruby.RubyNil;
 import org.jruby.RubyString;
 import org.jruby.RubySymbol;
 import org.jruby.ext.bigdecimal.RubyBigDecimal;
-import org.jruby.util.ByteList;
 import org.logstash.ext.JrubyTimestampExtLibrary;
 
 public final class ObjectMappers {
@@ -106,8 +105,7 @@ public final class ObjectMappers {
             final WritableTypeId typeId =
                 typeSer.typeId(value, RubyString.class, JsonToken.VALUE_STRING);
             typeSer.writeTypePrefix(jgen, typeId);
-            final ByteList bytes = value.getByteList();
-            jgen.writeBinary(bytes.getUnsafeBytes(), bytes.begin(), bytes.length());
+            jgen.writeString(value.asJavaString());
             typeSer.writeTypeSuffix(jgen, typeId);
         }
     }
@@ -121,7 +119,7 @@ public final class ObjectMappers {
         @Override
         public RubyString deserialize(final JsonParser p, final DeserializationContext ctxt)
             throws IOException {
-            return RubyString.newString(RubyUtil.RUBY, p.getBinaryValue());
+            return RubyString.newString(RubyUtil.RUBY, p.getValueAsString());
         }
     }
 
--- a/logstash-core/src/test/java/org/logstash/EventTest.java
+++ b/logstash-core/src/test/java/org/logstash/EventTest.java
@@ -22,6 +22,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 public final class EventTest {
 
@@ -93,6 +94,16 @@ public final class EventTest {
         final RubyString before = (RubyString) e.getUnconvertedField("foo");
         Event er = Event.deserialize(e.serialize());
         assertEquals(before, er.getUnconvertedField("foo"));
+    }
+
+    @Test
+    public void toBinaryRoundtripNonAscii() throws Exception {
+        Event e = new Event();
+        e.setField("foo", "bör");
+        final RubyString before = (RubyString) e.getUnconvertedField("foo");
+        Event er = Event.deserialize(e.serialize());
+        assertEquals(before, er.getUnconvertedField("foo"));
+        assertTrue(before.op_cmp((RubyString)er.getUnconvertedField("foo")) == 0);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


original-brownbear:
@andrewvc here I'd like your input whether this is even a worthwhile goal or whether we should just go for forcing UTF-16. It seems to me the value of serializing the encoding diminishes the more we move to Java APIs, maybe it's just worth taking the risk and hving that PQ/non-PQ behaviour difference.

logstash-core/src/main/java/org/logstash/ObjectMappers.java
```diff
@@ -107,21 +122,82 @@ public void serializeWithType(final RubyString value, final JsonGenerator jgen,
...
+            final ByteArrayOutputStream baos =
+                new ByteArrayOutputStream(1 + bytes.length() + Integer.BYTES + encoding.length) {
+                    @Override
+                    public byte[] toByteArray() {
+                        return buf;
+                    }
+                };
+            baos.write(0xff);
+            final int prefixLen = encoding.length;
+            baos.write((byte) prefixLen >> 24);
```
yaauie: nitpick: using `Ints#toByteArray(int)` here would bring this to the same level of abstraction as the decoder, which uses `Ints#fromBytes(byte, byte, byte, byte)`.

logstash-core/src/main/java/org/logstash/ObjectMappers.java
```diff
@@ -107,21 +122,82 @@ public void serializeWithType(final RubyString value, final JsonGenerator jgen,
...
         }
 
         @Override
         public RubyString deserialize(final JsonParser p, final DeserializationContext ctxt)
             throws IOException {
-            return RubyString.newString(RubyUtil.RUBY, p.getBinaryValue());
+            final byte[] data = p.getBinaryValue();
+            final RubyString result;
+            // If the first unsigned byte in the data is 0xff we have to decode the encoding.
+            if ((data[0] & 0xff) == 0xff) {
```
yaauie: why the bitwise-AND here?

logstash-core/src/main/java/org/logstash/ObjectMappers.java
original-brownbear: @yaauie yea but then we have to allocate another `byte[]` here needlessly. I was just trying to keep the performance impact of adding this low, even though it looks dirty admittedly.

logstash-core/src/main/java/org/logstash/ObjectMappers.java
original-brownbear: @yaauie basically just for readability. I want to compare to compare `date[0]` to `0xff` as unsigned `byte` and this looked a little easier on the eyes than `data[0] == (byte) 0xff`, which just feels wrong.

original-brownbear:
@yaauie currently we're interpreting everything we write as Ruby's ASCII-8bit (unfortunately, that's the bug here). Neither ASCII nor UTF-8 allow for `0xff` as the first byte, so I figured that anything but ASCII and UFT-8/16 isn't going to get deserialized correctly ever anyways, so I'm good to make the assumption that all salvageable data will never start on `0xff`.

original-brownbear:
@yaauie that said, I'm not 100% convinced of my approach here tbh. We should eventually sync up on what we want to achieve here, this is obviously really hacky and I'd rather not implement it this way unless we really really want/need to have the transparency of the Ruby encoding here for the PQ.

original-brownbear:
@yaauie @andrewvc as discussed with (@andrewvc) going for the simpler solution of simply forcing UTF-16 LE here since Java strings are in that encoding anyways. Works fine in my testing + in the added UT.

original-brownbear:
hold off guys ... this causing some Ruby interop problems :( On it.

original-brownbear:
@yaauie @andrewvc fixed :) (had to go through the Java string step here, it's faster/cheaper than transcoding a Ruby `String` and less buggy across different system default encodings).

original-brownbear:
@andrewvc we're getting lucky here, but there's a reason for it.

examples:

```sh
jshell> new String("ABC".getBytes(java.nio.charset.StandardCharsets.US_ASCII))
$5 ==> "ABC"
```

```
jshell> new String("ABCö".getBytes(java.nio.charset.StandardCharsets.US_ASCII))
$6 ==> "ABC?"
```

```
jshell> new String("ABC[".getBytes(java.nio.charset.StandardCharsets.US_ASCII))
$7 ==> "ABC["
```

=> the fact that the standard charset decoder for UTF16 drops invalid sequences saves us here (in an admittedly very dirt way). ASCII chars will continue to work, non-ASCII gets a placeholder when the input actually was using ASCII-8bit, which it shouldn't be in most cases (mostly JRuby will just default to UTF16 LE as well) => we should be good here compatibility wise + probably will be able to recover some data cleanly even.

andrewvc:
@original-brownbear this is LGTM to me. Can you make sure this info gets into the release notes in a clear way?

original-brownbear:
@andrewvc will do :)

PR #9310, opened 2018-04-02
MINOR: Remove needless generic type from SequencedList

original-brownbear:
Minor cleanup of the interface in preparation for incoming improvements.
`SequencedList` is only used as a wrapper for a list of `byte[]` => remove the generic type.


kares:
merged with master, the `SequencedList` name isn't that intuitive for it to be a `byte[]` container.
no strong feelings either way, let us know how do you feel about closing vs merging the PR.

original-brownbear:
Thanks @kares for taking a look here :) I still think it's a valid simplification but I agree that the name is kind of meh. I'd still rather merge it I think but I'm fine either way :)

kares:
closing for now, as this seems as :nail_care: while there's other more important technical debt to address.

PR #9312, opened 2018-04-02
CLEANUP+PERFORMANCE: Config Compiler Improvements

original-brownbear:
Some improvements in preparation of Java plugins:

* Dry up logic around buffering in `Dataset`
* Use `ArrayList` instead of `RubyArray` for better performance in `SplitDataset`
* Ensure contiguous numbering of `Dataset` classes to make dependency relations easier to debug
* Remove redundant tests


original-brownbear:
@danhermann thanks!

PR #9313, opened 2018-04-02
CLEANUP: Remove dead code from PQ, simplify CheckpointIO read

original-brownbear:
* all removed methods + classes except for `BufferedChecksumStreamInput` and `InputStreamStreamInput` were outright dead code
  * Replaced the only usage of those two (in reading a checkpoint) with a more efficient and idiomatic `ByteBuffer` + `CRC32` solution and removed them as well

=> 600 lines of code saved :)


original-brownbear:
@danhermann tests are actually green if you go to Jenkins, not sure why the GH API didn't update :)

original-brownbear:
@danhermann thanks!

